Issue #220, opened 2018-09-12
Subclassing a #[pyclass] in Python causes segfault during interpreter finalization


joar:
## 🐛 Bug Reports

When reporting a bug, please provide the following information. If this is not a bug report you can just discard this template.

### 🌍 Environment

 - Your operating system and version: Ubuntu 18.04
 - Your python version: Python 3.7
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?:
   - Python installed through the "deadsnakes" PPA
   - I use python's `venv`
 - Your rust version (`rustc --version`):  
`rustc 1.30.0-nightly (2d4e34ca8 2018-09-09)`

### 💥 Reproducing

Check out the branch at https://github.com/joar/pyo3/tree/bug/subclass-causes-sigsegv/examples/word-count

```console
$ cd examples/word-count

$ python setup.py develop

$ cat ../../target/debug/build/pyo3-4833970e0e0d17d0/output
cargo:rustc-cfg=Py_3_5
cargo:rustc-cfg=Py_3_6
cargo:rustc-cfg=Py_3_7
cargo:rustc-cfg=Py_3
cargo:rustc-cfg=py_sys_config="WITH_THREAD"
cargo:python_flags=FLAG_WITH_THREAD=1,CFG_Py_3_5,CFG_Py_3_6,CFG_Py_3_7
cargo:rerun-if-env-changed=LD_LIBRARY_PATH
cargo:rerun-if-env-changed=PATH
cargo:rerun-if-env-changed=PYTHON_SYS_EXECUTABLE
cargo:rerun-if-env-changed=LIB

$ python minimal-test.py
===> creating WordCounter
===> created WordCounter
===> deleting wc
===> deleted wc
fish: “python minimal-test.py” terminated by signal SIGSEGV (Address boundary error)
```

I believe the issue occurs during python's final GC.

konstin:
Thanks for the report!

I've minimized the example:

```rust
#[pyclass(subclass)]
struct SomeClass {}

#[pymethods]
impl SomeClass {
    #[new]
    fn __new__(obj: &PyRawObject) -> PyResult<()> {
        obj.init(|_| SomeClass {})
    }
}

#[pymodinit]
fn word_count(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<SomeClass>()?;
    Ok(())
}
```

```python
from word_count import SomeClass


class SomeSubClass(SomeClass):
    pass


[...]

PR #322, opened 2019-01-09
Pass -c instead of --workdir to "examples/*" tox invocation

joar:
The `-c example-path/tox.init` parameter will read the tox.ini from the
provided location and also set `workdir` and `toxinidir` to the
directory of the passed tox.ini

Previously, when only --workdir was passed, tox read the root `tox.ini`,
with the effect that tox was never run on the projects in `examples/`.

### DIFF
--- a/ci/travis/test.sh
+++ b/ci/travis/test.sh
@@ -1,12 +1,10 @@
 #!/bin/bash
-# Note: Avoid using "-e" globally, its behaviour may sometimes be convoluted.
-# For more details, see http://mywiki.wooledge.org/BashFAQ/105
-set -x
+set -ex
 
-cargo fmt --all -- --check || exit 1
-cargo test --features "$FEATURES" || exit 1
-cargo clippy --features "$FEATURES" || exit 1
+cargo fmt --all -- --check
+cargo test --features "$FEATURES"
+cargo clippy --features "$FEATURES"
 
 for example_dir in examples/*; do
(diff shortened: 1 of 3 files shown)


joar:
What are the odds? 

The [AppVeyor test](https://ci.appveyor.com/project/fafhrd91/pyo3/builds/21504161/job/qf66tw4inv2a8rb1) for my unrelated, bash-script-only commit https://github.com/PyO3/pyo3/pull/322/commits/2bd5931a04eb6c824f8a1ff81bc3806686c2f79 [1] crashed with what according to the internet (https://github.com/Alexpux/MINGW-packages/issues/2520) looks like a segfault.

[1]: the commit itself however fixes an issues found when debugging & reproducing a segfault issue on linux: https://github.com/PyO3/pyo3/issues/220#issuecomment-452822240. I'm not it's the computers that are wrong anymore. It's starting to look like I'm the one to blame for all these segfaults.

joar:
@konstin I added back `-e`

joar:
Test failure seems to be caused by broken rust/cargo nightly build. I can't install the `nightly` toolchain via rustup locally at the moment either.

athre0z:
Given that the nightly toolchain's recent tendency to be broken more often than it is working, it might be worth considering to pin it to a dated version and update it only every once in a while (e.g. `nightly-2019-01-08` appears to be working fine for me).

konstin:
Yes, I've already done that for pyo3-pack and I think this could be useful for pyo3 as well.

PR #323, opened 2019-01-09
Avoid race condition in gc tests

### LINKED ISSUE #198: Sporadic segfault on test_gc
I have the below error when running tests without `--test-threads=1`, e.g., the default or  `RUST_BACKTRACE=full cargo test --features python3 --no-default-features -- --test-threads=8`. It doesn't happen when using a single thread. It happens maybe one in twenty runs.

```
Running target/debug/deps/test_gc-a5a51037f354cb8a
running 7 tests
test class_with_freelist ... ok
test create_pointers_in_drop ... ok
test inheritance_with_new_methods_with_drop ... ok
test data_is_dropped ... ok
error: process didn't exit successfully: `pyo3/target/debug/deps/test_gc-a5a51037f354cb8a` (signal: 11, SIGSEGV: invalid memory reference
```

Alexander-N:
To avoid a segfault when the object is collected, disable garbage
collection in `gc_integration2`.

Closes #198

### DIFF
--- a/tests/test_gc.rs
+++ b/tests/test_gc.rs
@@ -186,8 +186,11 @@ struct GCIntegration2 {}
 fn gc_integration2() {
     let gil = Python::acquire_gil();
     let py = gil.python();
+    // Temporarily disable pythons garbage collector to avoid a race condition
+    py.run("import gc; gc.disable()", None, None).unwrap();
     let inst = Py::new_ref(py, || GCIntegration2 {}).unwrap();
-    py_run!(py, inst, "import gc; assert inst in gc.get_objects()");
+    py_run!(py, inst, "assert inst in gc.get_objects()");
+    py.run("gc.enable()", None, None).unwrap();
 }
 
 #[pyclass(weakref)]


pganssle:
Sorry, @Alexander-N would you mind explaining why this works? I am not sure why disabling garbage collection during this function call solves the problem.

I don't totally understand what this is intended to test, and I'm a bit worried that by disabling the gc during the test, we're papering over an actual threading bug in PyO3.

Alexander-N:
Of course. While I'm confident that this solves the race condition in the tests I can't assess if this is needs further considerations to avoid that this happens to users since I'm not familiar with this project.

With the gc parameter in `#[pyclass(gc)]` the new object `inst` is tracked by the python garbage collector, which is verified with ```assert inst in gc.get_objects()```. But if `inst` is collected between creation and the assertion it will result in a segfault. To avoid this, the garbage collector is disabled until the test is done.

I could reproduce the segfault every time with
```
for run in {1..20}; do cargo test --test test_gc; done
```

pganssle:
What are `gc_integration` and `gc_integration2` intended to test? They obviously test something related to the garbage collector, I am not sure that disabling the garbage collector during the test is the right thing to do, but without understanding what this tests I can't be sure.

pganssle:
Looking at this slightly more closely, it seems like the original version of `gc_integration` is supposed to ensure that the garbage collector is tracking `inst`, right?

```rust
    let inst = Py::new_ref(py, || GCIntegration2 {}).unwrap();
    py_run!(py, inst, "assert inst in gc.get_objects()");
```

Why would the garbage collector collect `inst` before `py` relinquishes its reference to the object?

PR #324, opened 2019-01-13
Add $crate:: prefix in create_exception!

athre0z:
This allows for selective macro includes in Rust 2018 projects. Surrounding code already makes heavy use of the `$crate` construct, so I suspect it was simply forgotten in these two places.


konstin:
Thank you!

PR #327, opened 2019-01-24
Implement Basic Cross Compile Support

mtp401:
This PR implements basic cross compile support. When cross compiling, the target python interpreter cannot be invoked since it may be built for an incompatible architecture (e.g. an armv7 target on an x86_64 host). Instead, attempt to parse the target interpreters compilation options out of it's headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,16 +27,6 @@ E.g. if you are on Ubuntu18.04, please run
 sudo apt install python3-dev python-dev
 ```
 
-### Cross Compiling
-
-Cross compiling Pyo3 modules requires setting the following environment
-variables:
-
-- `PYO3_CROSS_INCLUDE_DIR`: This variable must be set to the directory
-   containing the headers for the target's python interpreter.
-- `PYO3_CROSS_LIB_DIR`: This variable must be set to the directory
-   containing the target's libpython DSO.
-
 ## Using rust from python
 
 Pyo3 can be used to generate a native python module.
--- a/README.md
+++ b/README.md
@@ -32,11 +32,9 @@ sudo apt install python3-dev python-dev
 Cross compiling Pyo3 modules requires setting the following environment
 variables:
 
-- `PYO3_XC`: This variable must be set to indicate a cross compilation
-   environment. The value does not matter.
-- `PYO3_XC_PYTHON_INCLUDE_DIR`: This variable must be set to the directory
+- `PYO3_CROSS_INCLUDE_DIR`: This variable must be set to the directory
    containing the headers for the target's python interpreter.
-- `PYO3_XC_PYTHON_LIB_DIR`: This variable must be set to the directory
+- `PYO3_CROSS_LIB_DIR`: This variable must be set to the directory
    containing the target's libpython DSO.
 
 ## Using rust from python
--- a/README.md
+++ b/README.md
@@ -27,6 +27,18 @@ E.g. if you are on Ubuntu18.04, please run
 sudo apt install python3-dev python-dev
 ```
 
+### Cross Compiling
+
+Cross compiling Pyo3 modules requires setting the following environment
+variables:
+
+- `PYO3_XC`: This variable must be set to indicate a cross compilation
+   environment. The value does not matter.
+- `PYO3_XC_PYTHON_INCLUDE_DIR`: This variable must be set to the directory
+   containing the headers for the target's python interpreter.
+- `PYO3_XC_PYTHON_LIB_DIR`: This variable must be set to the directory
+   containing the target's libpython DSO.
+
 ## Using rust from python
 
 Pyo3 can be used to generate a native python module.
--- a/build.rs
+++ b/build.rs
@@ -124,8 +124,7 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
 
 fn load_cross_compile_info() -> Result<(PythonVersion, HashMap<String, String>, Vec<String>), String>
 {
-    let python_include_dir =
-        env::var("PYO3_XC_PYTHON_INCLUDE_DIR").expect("PYO3_XC_PYTHON_INCLUDE_DIR must be defined");
+    let python_include_dir = env::var("PYO3_CROSS_INCLUDE_DIR").unwrap();
     let python_include_dir = Path::new(&python_include_dir);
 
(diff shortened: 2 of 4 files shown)


kngwyu:
Thanks!
But I'm sorry I don't know what the `patchlevel.h` is and what `parse_header_defines` does.
Could you please explain them more?

mtp401:
I think the comment for `parse_header_defines` sums up the behavior of that function. It takes a path to a C header and parses every single line `#define X Y` statement into a `HashMap` entry that maps `X` to `Y`. `patchlevel.h` is where cpython keeps all of the versioning information associated with the python interpreter. You can take a look at `patchlevel.h` for [Python 3.8.0 as an example](https://github.com/python/cpython/blob/master/Include/patchlevel.h).

README.md
```diff
@@ -27,6 +27,18 @@ E.g. if you are on Ubuntu18.04, please run
 sudo apt install python3-dev python-dev
 ```
 
+### Cross Compiling
+
+Cross compiling Pyo3 modules requires setting the following environment
+variables:
+
+- `PYO3_XC`: This variable must be set to indicate a cross compilation
```
kngwyu: Do we really need this variable?
I think just
```rust
env::var("PYO3_XC_PYTHON_INCLUDE_DIR").is_ok() && env::var("PYO3_XC_PYTHON_LIB_DIR")
```
or so is enough.

kngwyu:
Thanks!
I still don't like the situation we need the header file only to know the python configs, but other than that, I'm :+1:  for this PR.

build.rs
```diff
@@ -74,11 +78,99 @@ static SYSCONFIG_VALUES: [&'static str; 1] = [
     "Py_UNICODE_SIZE", // note - not present on python 3.3+, which is always wide
 ];
 
+/// Attempts to parse the header at the given path, returning a map of definitions to their values.
+/// Each entry in the map directly corresponds to a `#define` in the given header.
+fn parse_header_defines<P: AsRef<Path>>(
+    pyconfig_path: P,
+) -> Result<HashMap<String, String>, String> {
+    let define_regex =
+        Regex::new(r"^\s*#define\s+(?P<ident>[a-zA-Z0-9_]+)\s+(?P<value>.+)\s*$").unwrap();
```
konstin: This should have an example line from the header, otherwise it's hard to see what the regex does.

build.rs
```diff
@@ -74,11 +78,99 @@ static SYSCONFIG_VALUES: [&'static str; 1] = [
...
+    let header_reader = BufReader::new(&header_file);
+
+    let definitions = header_reader
+        .lines()
+        .filter_map(|maybe_line| {
+            let line = if let Ok(l) = maybe_line {
+                l
+            } else {
+                return None;
+            };
```
konstin: You're silently discarding an error here; Please handle the error in some way (`.expect()` is fine or iirc there's some way to collect from `Iterator<Result<...>>`  to `Result<Iterator<...>, ...>`

src/types/exceptions.rs
```diff
@@ -342,7 +342,7 @@ impl UnicodeDecodeError {
         reason: &CStr,
     ) -> PyResult<&'p PyObjectRef> {
         unsafe {
-            let input: &[c_char] = &*(input as *const [u8] as *const [i8]);
+            let input: &[c_char] = std::mem::transmute::<&[u8], &[c_char]>(input);
```
konstin: Please do not use transmute if not strictly required. See https://doc.rust-lang.org/nightly/std/mem/fn.transmute.html:

> transmute is incredibly unsafe. There are a vast number of ways to cause undefined behavior with this function. transmute should be the absolute last resort.

konstin:
Thank you for your pull request! 

Please add some documentation about how to use cross compiling with pyo3, ideally with instructions on installing the required software. Adding a chapter to the book seems like the best place for that.

---

As far as I can see, the only mandatory information from the header files is the major and the minor version. Would it be possible to pass the major and minor version through environment variables or features, so you wouldn't even need to have the python version installed to build a native module? This is nothing I expect you to implement, but something that could be very useful even for same-platform compilation.

README.md
```diff
@@ -27,6 +27,18 @@ E.g. if you are on Ubuntu18.04, please run
...
+   environment. The value does not matter.
+- `PYO3_XC_PYTHON_INCLUDE_DIR`: This variable must be set to the directory
+   containing the headers for the target's python interpreter.
+- `PYO3_XC_PYTHON_LIB_DIR`: This variable must be set to the directory
+   containing the target's libpython DSO.
```
konstin: What do you think about `PYO3_CROSS_INCLUDE_DIR` and `PYO3_CROSS_LIB_DIR` instead of `PYO3_XC_INCLUDE_DIR` and `PYO3_XC_PYTHON_LIB_DIR`? This should be clearer.

src/types/exceptions.rs
mtp401: I'm open to a different way of doing this. The fundamental issue is that `char` is unsigned by default on some platforms (like ARM) and signed by default on others (such as x86). So, this code compiles fine on x86 but results in a compilation error when compiling for ARM. I thought about using a `cfg(target_arch = xxx)` for this but that has the downside of constantly playing catch up as new platforms are added.

konstin: Huh I didn't know about that! Would `let input: &[c_char] = &*(input as *const [u8] as *const [c_char]);` work? I've checked it works x86

mtp401:
As for:

```
As far as I can see, the only mandatory information from the header files is the major and the minor version. Would it be possible to pass the major and minor version through environment variables or features, so you wouldn't even need to have the python version installed to build a native module? This is nothing I expect you to implement, but something that could be very useful even for same-platform compilation.
```

There are other important pieces of information that are parsed out of `pyconfig.h` that are dependent not only on the version of Python itself, but also how the interpreter was compiled (`Py_UNICODE_SIZE` is a good example of this). Additionally, wouldn't you still needed Python installed as you need to link against the interpreter eventually anyway?

src/types/exceptions.rs
mtp401: I'm incorporating your other suggestions now and will try this.

mtp401:
@konstin this should be good to go. Tested fine on my end.

konstin:
Thank you!

PR #330, opened 2019-01-28
Add $crate:: prefix in import_exception!

Alexander-N:
Also change doctest to use Rust 2018 macro includes.


konstin:
Thank you!

Issue #210, opened 2018-08-25
Get rid of specialization


konstin:
~The last step for #5 is getting rid of specialization. I've remove most uses in 7c0379b1 and subsequent commits, except for the following ~17~ ~10~ ~9~ ~7~ 4:~ https://github.com/rust-lang/rust/pull/64564 changed the rules so there are again many `default fn`.

```rust 
src/class/descr.rs
76:    default fn tp_descr_get() -> Option<ffi::descrgetfunc> {
97:    default fn tp_descr_set() -> Option<ffi::descrsetfunc> {
130:    default fn tp_as_descr(_type_object: &mut ffi::PyTypeObject) {}

src/class/buffer.rs
46:    default fn tp_as_buffer() -> Option<ffi::PyBufferProcs> {
74:    default fn cb_bf_getbuffer() -> Option<ffi::getbufferproc> {
110:    default fn cb_bf_releasebuffer() -> Option<ffi::releasebufferproc> {

src/class/gc.rs
27:    default fn update_type_object(_type_object: &mut ffi::PyTypeObject) {}
72:    default fn tp_traverse() -> Option<ffi::traverseproc> {
124:    default fn tp_clear() -> Option<ffi::inquiry> {

src/class/iter.rs
49:    default fn tp_as_iter(_typeob: &mut ffi::PyTypeObject) {}
71:    default fn tp_iter() -> Option<ffi::getiterfunc> {
94:    default fn tp_iternext() -> Option<ffi::iternextfunc> {

src/class/basic.rs
152:    default fn tp_as_object(_type_object: &mut ffi::PyTypeObject) {}
153:    default fn nb_bool_fn() -> Option<ffi::inquiry> {
183:    default fn tp_getattro() -> Option<ffi::binaryfunc> {
252:        default fn set_attr() -> Option<ffi::setattrofunc> {
274:        default fn del_attr() -> Option<ffi::setattrofunc> {
296:        default fn set_del_attr() -> Option<ffi::setattrofunc> {
324:    default fn tp_str() -> Option<ffi::unaryfunc> {
344:    default fn tp_repr() -> Option<ffi::unaryfunc> {
364:    default fn tp_hash() -> Option<ffi::hashfunc> {
389:    default fn nb_bool() -> Option<ffi::inquiry> {
409:    default fn tp_richcompare() -> Option<ffi::richcmpfunc> {

src/class/mapping.rs
[...]

PR #332, opened 2019-01-30
 Allow 0..n pymethod blocks without specialization 

### LINKED ISSUE #302: Allow multiple #[pymethods] impl blocks
Currently, having more than one impl block annotated with `#[pymethods]` results in a compiler error (`conflicting implementation for 'A'`). This appears to be the case due to

```rust
impl ::pyo3::class::methods::PyMethodsProtocolImpl for A {
    fn py_methods() -> &'static [::pyo3::class::PyMethodDefType] {
        static METHODS: &'static [::pyo3::class::PyMethodDefType] = &[];
        METHODS
    }
}
```

being emit on every invocation. While most certainly not a high priority issue, it'd be nice to be able to have multiple such impl blocks in order to allow macros to implement methods by opening up their own impl block instead of having to do so manually (e.g. for cases where you invoke them on types that otherwise wouldn't have any non-macro impl blocks).

Attached you'll find a minimal example of the issue.
[pyo3-muliple-impl.tar.gz](https://github.com/PyO3/pyo3/files/2656464/pyo3-muliple-impl.tar.gz)

konstin:
Surprisingly it's possible and even easy to support multiple impl blocks and even get rid of specialization! This all works thanks to [inventory](https://github.com/dtolnay/inventory), which uses the ctor binary section to have life-before-main code collect a bunch of instances.

This pull requests uses the `doc(hidden)` trait `inventory::Collect`, for which I've opened https://github.com/dtolnay/inventory/issues/5

As a bonus, we also get rid of `PyPropMethodsProtocolImpl`.

Fixes #302 
Relevant for #210

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,21 +20,21 @@ appveyor = { repository = "fafhrd91/pyo3" }
 codecov = { repository = "PyO3/pyo3", branch = "master", service = "github" }
 
 [dependencies]
-libc = "0.2.43"
+libc = "0.2.48"
 spin = "0.5.0"
 num-traits = "0.2.6"
 pyo3cls = { path = "pyo3cls", version = "=0.6.0-alpha.2" }
 mashup = "0.1.9"
 num-complex = { version = "0.2.1", optional = true }
-inventory = "0.1.2"
+inventory = "0.1.3"
 
 [dev-dependencies]
-assert_approx_eq = "1.0.0"
+assert_approx_eq = "1.1.0"
 docmatic = "0.1.2"
 indoc = "0.3.1"
 
 [build-dependencies]
-regex = "1.0.5"
+regex = "1.1.0"
 version_check = "0.1.5"
 
 [features]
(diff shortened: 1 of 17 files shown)


discosultan:
Does this mean pyo3 will be supported on stable Rust?

konstin:
Not yet, but that was a major blocker. You can see the remaining blocker by commenting out `#![feature(specialization)]` in `src/lib.rs` and running `cargo check`. And we're of course happy about pull requests fixing some of those errors ;)

kngwyu:
Really great PR, thanks.
Looks the outputs of `repr` changed in examples, e.g. `rustapi_module.datetime.TzClass` changed to `TzClass`, which makes travis fail.
Is this intended?

konstin:
I've removed the module name because currently the module name is whatever module initializes the pyclass, which imho bad because it's not really deterministic. E.g. Adding a module to your module that also adds the class make the class in both modules end up with a different module name.

konstin:
Eventually, I'd like to have a module name again, but it should be something less volatile.

kngwyu:
@konstin 
> I've removed the module name because currently the module name is whatever module initializes the pyclass, which imho bad because it's not really deterministic.

I see, :+1: for this.

Issue #320, opened 2019-01-05
Segfault adding class object to module.


derekdreery:
### 🌍 Environment

 - Your operating system and version: Linux: 4.18.16-arch1-1-ARCH
 - Your python version: 3.7.1
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: arch-linux (no venv)
 - Your rust version (`rustc --version`):
 - Are you using the latest pyo3 version?: git-master

### 💥 Reproducing

```rust
#![feature(specialization)]

use pyo3::prelude::*;

/// Main entry point - set up modules, functions, etc.
#[pymodule]
fn libbug(_: Python, m: &PyModule) -> PyResult<()> {
    m.add("Foo", A { kind: 0 })?;
    Ok(())
}

#[pyclass]
#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
pub struct A {
    kind: u32,
}
```

Import the .so -> segfault (somewhat intermittent).

kngwyu:
Thanks for reporting!
It looks design mistake...
[This implementation](https://github.com/PyO3/pyo3/blob/master/src/conversion.rs#L44) is used when `m.add(A {})` is called, but **we can't use `as_ptr` before we call `into_object`**.
So to fix this bug is relatively easy, but I think we have to rethink `pyclass` design.
1. We can't know a struct is in a python heap(and has `ob_refcnt` and `ob_type`), or not. Is it proper design? Ideally we should use different type for python-allocated items, though it's far from current design.
2. For `pyclass`, currently `IntoPyObject` actually allocates in python heap and copies its data. So I think `Into` prefix is not suitable.

athre0z:
I think you're on the right track there @kngwyu. This design issue is probably also closely related to #308. The [`ToPyObject` implementation](https://github.com/PyO3/pyo3/blob/master/pyo3-derive-backend/src/py_class.rs#L174) generated by the proc macro has the same issue.

derekdreery:
Do you have any ideas of how to fix this? I'm happy to help implement if possible.

konstin:
I fear it's a bit complex to fix. Basically, we need to properly differentiate between instances of pyclass structs that can be wrapped in a python object, but aren't, and types that are python objects (i.e. they are/have a `*mut PyObject`). I think the bug is we assume that our instance is the latter while it is the former in the the pyclass implementation:

[...]

PR #335, opened 2019-02-01
Remove ToPyPointer and so on from pyclass

### LINKED ISSUE #308: Auto-implementation of `ToPyObject` allows PyObject to outlive object
### 🌍 Environment

 - Your operating system and version: macOS 14.2
 - Your python version: Python 3.6
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: macports, yes
 - Your rust version (`rustc --version`): rustc 1.32.0-nightly (14997d56a 2018-12-05)
 - Are you using the latest pyo3 version? -- Yes, I'm tested this on `master`

### 💥 Reproducing

I PRed two tests that reliably reproduce the issue in #307.

kngwyu:
And introduce `PyRef` and `PyRefMut`.
See #320 for the reason.

Currently I don't have good idea for [`PyIterProtocol` test](https://github.com/PyO3/pyo3/blob/ee4fc3b21d05236df2a3fd2addf83b0aab8fcdf6/tests/test_dunder.rs#L56) and it still fails to be compiled.

Fixes #308

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -17,11 +17,15 @@ struct MyClass {
 The above example generates implementations for `PyTypeInfo` and `PyTypeObject` for `MyClass`.
 
 ## Get Python objects from `pyclass`
-In rust side, we can get a Python object which includes `pyclass` by 3 methods.
+You can use `pyclass`es like normal rust structs.
+
+However, if instantiate noramlly, you can't treat `pyclass`es as Python objects.
+
+To get a Python object which includes `pyclass`, we have to use some special methods.
 
 ### `PyRef`
-`PyRef` is a special reference, which ensures the reffrered struct is a part of
-a Python object.
+`PyRef` is a special reference, which ensures that the reffrered struct is a part of
+a Python object, and you are also holding the GIL.
 
 You can get an instance of `PyRef` by `PyRef::new`, which does 3 things:
 1. Allocate a Python object in the Python heap
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -16,6 +16,53 @@ struct MyClass {
 
 The above example generates implementations for `PyTypeInfo` and `PyTypeObject` for `MyClass`.
 
+## Get Python objects from `pyclass`
+In rust side, we can get a Python object which includes `pyclass` by 3 methods.
+
+### `PyRef`
+`PyRef` is a special reference, which ensures the reffrered struct is a part of
+a Python object.
+
+You can get an instance of `PyRef` by `PyRef::new`, which does 3 things:
+1. Allocate a Python object in the Python heap
+2. Copies the rust struct into the Python object
+3. Returns a reference of it
+
+You can use `PyRef` just like `&T`, because it implements `Deref<Target=T>`.
+```rust
+# use pyo3::prelude::*;
+#[pyclass]
+struct MyClass {
+   num: i32,
+   debug: bool,
+}
+let gil = Python::acquire_gil();
+let py = gil.python();
+let obj = PyRef::new(py, || MyClass { num: 3, debug: true }).unwrap();
+assert_eq!(obj.num, 3);
+let dict = PyDict::new();
+// You can treat a `PyRef` as a Python object
+dict.set_item("obj", obj)).unwrap();
+```
+
+### `PyRefMut`
+`PyRefMut` is a mutable version of `PyRef`.
+```rust
+# use pyo3::prelude::*;
+#[pyclass]
+struct MyClass {
+   num: i32,
+   debug: bool,
+}
+let gil = Python::acquire_gil();
+let py = gil.python();
+let mut obj = PyRefMut::new(py, || MyClass { num: 3, debug: true }).unwrap();
+obj.num = 5;
+```
+
+### `Py`
+`Py` is a object wrapper which stores an object longer than the GIL lifetime.
+TODO: Write a good example
 
 ## Customizing the class
 
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -62,7 +62,22 @@ obj.num = 5;
 
 ### `Py`
 `Py` is a object wrapper which stores an object longer than the GIL lifetime.
-TODO: Write a good example
+
+You can use it to avoid lifetime problems.
+```rust
+# use pyo3::prelude::*;
+#[pyclass]
+struct MyClass {
+   num: i32,
+}
+fn return_myclass() -> Py<MyClass> {
+    let gil = Python::acquire_gil();
+    Py::new(|| MyClass { num: 1 })
+}
+let gil = Python::acquire_gil();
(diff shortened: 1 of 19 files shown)


konstin:
Thank you, the two types are great! 

Both structs should get a docstring with an example, a note in the guide and a conversion from PyObject to that struct, so that it can be used in arguments.

For `PyIterProtocol`, I think that the current signature is wrong. There might be a better solution with `arbitrary_self_types`, but for now passing `self` as not-self parameter to `__iter__` might be the best solution.

kngwyu:
> There might be a better solution with arbitrary_self_types

You mean `self: Py<Self>` or so?
It seems really good except it's unstable :sad:

> for now passing self as not-self parameter to __iter__ might be the best solution

Thanks, I'll try this.

kngwyu:
Ready to be reviewed now.
To avoid confusion, I modified `AsPyRef` to return `PyRef`.
By this change, now `PyObject::as_ref` returns `PyRef<PyObjectRef>`.
It's not desirable, but I think it's not so much confusing because I also modified `PyTryFrom` so that it can take `PyRef` and `PyRefMut`.

src/instance.rs
```diff
@@ -25,18 +25,130 @@ pub trait PyObjectWithGIL: Sized {
...
+
+impl<'a, T> PyRef<'a, T> {
+    pub(crate) fn new(t: &'a T) -> Self {
+        PyRef { inner: t }
+    }
+}
+
+impl<'a, T: PyTypeInfo> ToPyPointer for PyRef<'a, T> {
+    fn as_ptr(&self) -> *mut ffi::PyObject {
+        unsafe { (self.inner as *const _ as *mut u8).offset(-T::OFFSET) as *mut _ }
```
konstin: Clippy doesn't like this; I think we just need to clippy::allow that.

src/instance.rs
```diff
@@ -25,18 +25,130 @@ pub trait PyObjectWithGIL: Sized {
...
+
+impl<'a, T> PyRefMut<'a, T> {
+    pub(crate) fn new(t: &'a mut T) -> Self {
+        PyRefMut { inner: t }
+    }
+}
+
+impl<'a, T: PyTypeInfo> ToPyPointer for PyRefMut<'a, T> {
+    fn as_ptr(&self) -> *mut ffi::PyObject {
+        unsafe { (self.inner as *const _ as *mut u8).offset(-T::OFFSET) as *mut _ }
```
konstin: Here also clippy allow

src/instance.rs
```diff
@@ -25,18 +25,130 @@ pub trait PyObjectWithGIL: Sized {
 #[doc(hidden)]
 pub trait PyNativeType: PyObjectWithGIL {}
 
+/// A reference of type `T` which is in Python heap.
+///
+/// We can't implement `ToPyPointer` or `ToPyObject` for `pyclass`es, because they're not python
+/// objects untill copied to Python heap. So, instead of treating `&pyclass` as a Python object, we
```
konstin:
```suggestion
/// objects until copied to the Python heap. So, instead of treating `&pyclass` as a Python object, we
```

src/instance.rs
```diff
@@ -25,18 +25,130 @@ pub trait PyObjectWithGIL: Sized {
 #[doc(hidden)]
 pub trait PyNativeType: PyObjectWithGIL {}
 
+/// A reference of type `T` which is in Python heap.
```
konstin: The docs should mention that `T` is part of a python object if it's a PyRef[Mut].

src/instance.rs
```diff
@@ -25,18 +25,130 @@ pub trait PyObjectWithGIL: Sized {
...
+///
+/// We can't implement `ToPyPointer` or `ToPyObject` for `pyclass`es, because they're not python
+/// objects untill copied to Python heap. So, instead of treating `&pyclass` as a Python object, we
+/// need to use special reference types `PyRef` and `PyRefMut`.
+/// ```
```
konstin: This should be separated by an `# Example` header like in the std docs.

src/instance.rs
```diff
@@ -25,18 +25,130 @@ pub trait PyObjectWithGIL: Sized {
...
+/// # extern crate pyo3; fn main() {
```
konstin: iirc you don't need this line

src/instance.rs
```diff
@@ -25,18 +25,130 @@ pub trait PyObjectWithGIL: Sized {
...
+}
+
+impl<'a, T> Deref for PyRef<'a, T> {
+    type Target = T;
+    fn deref(&self) -> &T {
+        self.inner
+    }
+}
+
+impl<'a, T> Into<&'a PyObjectRef> for PyRef<'a, T>
```
konstin: https://doc.rust-lang.org/nightly/std/convert/trait.Into.html:

> Library authors should not directly implement this trait, but should prefer implementing the From trait, which offers greater flexibility and provides an equivalent Into implementation for free, thanks to a blanket implementation in the standard library.

src/instance.rs
kngwyu: :+1:

src/instance.rs
kngwyu: Thanks

src/instance.rs
kngwyu: Thanks, I didn't know that

src/instance.rs
kngwyu: :+1:

src/instance.rs
kngwyu: Thanks

kngwyu:
Thanks for your review!
I refactored codes based on your comments, though still I haven't completed the guide.
> Are there Ts for PyRef that do not implement PyTypeInfo?

I think there aren't so added a bound to PyRef itself.

kngwyu:
Write a draft of document, and removed `Py::new_ref` in favor of `PyRef::new`.
I'm sorry this change may conflicts with #344, but I think we need this, because `Py::new_ref` is not related to `Py`.

guide/src/class.md
```diff
@@ -16,6 +16,68 @@ struct MyClass {
 
 The above example generates implementations for `PyTypeInfo` and `PyTypeObject` for `MyClass`.
 
+## Get Python objects from `pyclass`
+In rust side, we can get a Python object which includes `pyclass` by 3 methods.
```
konstin: This needs a short explanation that if you instantiate a rust struct, it doesn't need to be part of a python object a python object. PyRef however guarantees that the struct instance is part of a python object and that you're also holding the GIL, which allows you to modify the struct. 

Everything else looks good!

konstin:
I'll rebase #344 onto this once it is merged to master, it shouldn't be much more than a bit of search an replace.

guide/src/class.md
kngwyu: :+1:

PR #337, opened 2019-02-03
Fix some examples

Alexander-N:
Adapt some imports which changed in https://github.com/PyO3/pyo3/commit/302c099a766a2a3759933b9c5da9c7ca90c91507


konstin:
Thank you!

PR #347, opened 2019-02-10
[CI] Install clippy from git when it's missing

kngwyu:
See https://github.com/rust-lang/rust-clippy#travis-ci


konstin:
Great find! Does this mean we can unpin the nightly?

kngwyu:
Oh I didn't find now we use pinned nighlty in travis.
I hope so, though still it can be broken sometimes.

PR #349, opened 2019-02-11
Check minimum rustc version on travis

konstin:


konstin:
So i figured out that you can't pin the nightly if you have a `rust-toolchain` file because than rustup decides to download the exact toolchain in the file which was in our case the latest nightly. I also removed the long broken kcov clutter

PR #351, opened 2019-02-12
Make the init methods use a value instead of a function

konstin:
Cherry picked due to some trait error on the other branch

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -52,7 +52,7 @@ pub trait PyNativeType: PyObjectWithGIL {}
 /// }
 /// let gil = Python::acquire_gil();
 /// let py = gil.python();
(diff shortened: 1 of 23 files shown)


kngwyu:
Looks `cargo fmt` fails

src/instance.rs
```diff
@@ -350,17 +346,40 @@ where
...
-        ob.init(f)?;
+        ob.init(value)?;
 
         let ob = unsafe { Py::from_owned_ptr(ob.into_ptr()) };
         Ok(ob)
     }
+
+    /// Create new instance of `T` and move it under python management.
+    /// Returns references to `T`
+    pub fn new_ref(py: Python, value: T) -> PyResult<&T>
```
kngwyu: We don't need these two functions, since now `PyRef::new` and `PyRefMut::new` exist

PR #359, opened 2019-02-16
Add support for creating `PyList` from any `IntoIterator` for known s…

chr1sj0nes:
…ize.


kngwyu:
Thanks!
I fear it doesn't accept `&[PyRef]` or so, but it's much better.

PR #363, opened 2019-02-18
Run rustfmt and clippy only at the minimum nightly

kngwyu:
They are too often missing...


konstin:
Thanks, running rustfmt and clippy on one build only is much better,

PR #364, opened 2019-02-18
Big proc macro refactoring

konstin:
* Removed a lot of clutter, unified some code
 * Started using syn::parse::Parse for pyfunction attributes 
 * No more newlines between imports
 * Renamed `#[prop(get, set)]` to `#[pyo3(get, set)]`
 * `#[pyfunction]` now supports the same arguments as `#[pyfn()]`
 * Some macros now emit proper spanned errors instead of panics.
 * Fixed clippy lints

Examples for the spanned errors:

```
error: Unsupported parameter
  --> examples/word-count/src/lib.rs:11:21
   |
11 | #[pyclass(subclass, invalid)]
   |                     ^^^^^^^
```

grafik


pyo3-derive-backend/src/func.rs
```diff
@@ -81,274 +80,221 @@ pub fn impl_method_proto(
...
+            }
+            let p: syn::Path = syn::parse_str(proto).unwrap();
+            let arg1_name = syn::Ident::new(arg1, Span::call_site());
+            let arg1_ty = get_arg_ty(sig, 0);
+            let arg2_name = syn::Ident::new(arg2, Span::call_site());
+            let arg2_ty = get_arg_ty(sig, 1);
+            let (ty, succ) = get_res_success(ty);
+
+            // rewrite ty
+            let tmp = extract_decl(syn::parse_quote! {fn test(
```
kngwyu: I feel this hard to read

pyo3-derive-backend/src/func.rs
```diff
@@ -81,274 +80,221 @@ pub fn impl_method_proto(
...
+            }
+            let p: syn::Path = syn::parse_str(proto).unwrap();
+            let arg1_name = syn::Ident::new(arg1, Span::call_site());
+            let arg1_ty = get_arg_ty(sig, 1);
+            let arg2_name = syn::Ident::new(arg2, Span::call_site());
+            let arg2_ty = get_arg_ty(sig, 2);
+            let (ty, succ) = get_res_success(ty);
+
+            // rewrite ty
+            let tmp = extract_decl(syn::parse_quote! {fn test(
```
kngwyu: Same as above, I want indent here.

pyo3-derive-backend/src/pyfunction.rs
```diff
@@ -0,0 +1,261 @@
...
+                        return Err(syn::Error::new_spanned(item, "*(var args) is defined"));
+                    }
+                    self.has_varargs = true;
+                    self.arguments.push(Argument::VarArgs(nv.ident.clone()));
+                } else if litstr.value() == "**" {
+                    // kwargs="**"
+                    if self.has_kwargs {
+                        return Err(syn::Error::new_spanned(
+                            item,
+                            "self.arguments already define ** (kw args)",
```
kngwyu: I don't understand why you mention `self.arguments` here.
Do users really need to know about this?

src/conversion.rs
```diff
@@ -339,29 +339,24 @@ where
...
 ///
-/// This is necessaty because proc macros run before typechecking and can't decide
+/// This is necessary because proc macros run before typechecking and can't decide
 /// whether a return type is a (possibly aliased) PyResult or not. It is also quite handy because
 /// the codegen is currently built on the assumption that all functions return a PyResult.
-pub trait ReturnTypeIntoPyResult {
-    type Inner;
-
-    fn return_type_into_py_result(self) -> PyResult<Self::Inner>;
+pub trait IntoPyResult<T> {
```
kngwyu: :+1:  for this change

kngwyu:
Great work, thanks!
Though I couldn't check all sentences in detail, I left some comments mainly about style and error messages.

konstin:
The code you commented wasn't written or changed by and is only in the diff because I reduced the indent. That part is for another refactoring.

konstin:
Btw @pganssle I think I found the reason why pyo3 is so slow for function calls. Every function call generates a call to `parse_fn_args` with a slice to options which is than filled by that function. I've removed the slice-to-iter-than-convert-in-nested-blocks part, but the argument conversion is still awkward.

https://github.com/PyO3/pyo3/blob/d02f7c3aa54bde2c03e358b539b96583aa5bf95b/pyo3-derive-backend/src/pymethod.rs#L396-L420

PR #367, opened 2019-02-23
Some api improvements

konstin:


kngwyu:
Is `PyManagedRef` meant to enable users to avoid selecting `as_ptr` and `to_object` manually?
If so, it seems a too complex solution to me:thinking:

konstin:
You can't select that manually because you can't tell if the object even has a `as_ptr` and always calling `to_object` is imho worse. The trait is essentially java's `instanceof` for a specific, and I fear it's not possible to do this more straightforward without specialization.

PR #368, opened 2019-02-24
Rename ToPyPointer with AsPyPointer

kngwyu:
Why is it called `ToPyPointer`?


konstin:
Oh yeah that's definitely a better name

PR #369, opened 2019-02-24
Make PyNativeType unsafe

kngwyu:
Somehow it's now used to zero cost conversion from pointer to reference, so it's dangerous to expose it.


konstin:
There are some legitimate cases to implement PyNativeType in a downstream crate, e.g. in rust-numpy with PyArray. (Afaik this the trait is currently implicitly added through `pyobject_native_type_named!`.) What about marking it as an unsafe trait instead?

kngwyu:
:+1:

konstin:
Thank you!

PR #370, opened 2019-02-25
Test with PyPy and PyPy3

thedrow:
PyPy has a compatibility layer called cpyext which allows us to use CPython extensions.
I think it's worth our time to test PyO3 with it. I have marked the PyPy jobs as allowed failures for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,18 @@ matrix:
       python: "3.7"
       # Keep this synced up with build.rs
       env: FEATURES=python3 TRAVIS_RUST_VERSION=nightly-2019-02-07
+    - name: PyPy2.7 6.0
+      python: "pypy2.7-6.0"
+      env: FEATURES=python2
+    - name: PyPy3.5 6.0
+      python: "pypy3.5-6.0"
+      env: FEATURES=python3
   allow_failures:
     - python: "3.8-dev"
+      env: FEATURES=python3
+    - python: pypy2.7-6.0
+      env: FEATURES=python2
+    - python: pypy3.5-6.0
       env: FEATURES=python3
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,6 @@ matrix:
   allow_failures:
     - python: "3.8-dev"
       env: FEATURES=python3
-    - python: pypy2.7-6.0
-      env: FEATURES=python2
     - python: pypy3.5-6.0
       env: FEATURES=python3
 


thedrow:
PyPy does not support embedding Python through cpyext.
Is there a way we can disable it on PyPy builds?

kngwyu:
Looks [compile fails](https://travis-ci.org/PyO3/pyo3/jobs/498106953).
And, since we call `PyEval_ThreadsInitialized` from both extension and rust application now, it looks we need large refactoring to make pyo3 work with PyPy :thinking:

thedrow:
Why do you need `PyEval_ThreadsInitialized` for extensions as well?

kngwyu:
I'm not the author but I guess it's just because to reuse the same code.

konstin:
pypy support would be super cool to have! I didn't even know that cpyext was already part of a release. I think the build failure is only due to pypy not setting `WITH_THREAD`. It should be fixable with adding or removing some `cfg` guards.

> PyPy does not support embedding Python through cpyext.

This would be a showstopper since all of our tests work that way. The only option would be to only test the examples in the ci job.

Regarding pypy2: Python 2 support will be removed, so I don't think we should spend a ci run on pypy2.

omerbenamram:
Just a random comment, I actually got PyPy3 to work a while ago against an older version of PyO3. 
We have a production parsing system running PyPy, and we wanted to port some parsers to rust, and have them run side-by-side for now.
PyPy + rust via PyO3 is a bit slower than CPython because of the emulation overhead, but it is still beneficial since it's significantly faster than even the jitted python code.

So a lot of declarations look like this:
image

And I've also needed to fork PyObject to accout for PyPy's refcounting semantics in cpyext:
image

It actually works quite nicely, it did require to mess with the build script, and add a feature which renames link names for some functions, since pypy emulates the CPython API via a different convention of API names.

My little fork never got around to passing the all of the tests, a lot of them were segfaulting probably because of initialization code. I've also never ported it to compile against recent PyO3, but I think this may be worth an effort in the future :)

konstin:
That's some really great groundwork @omerbenamram! I'd love to use the work that went into [that branch](https://github.com/PyO3/pyo3/compare/master...omerbenamram:feature/add-build-pypy-support). Getting pyo3 to compile with pypy would be a huge benefit, even if the tests don't pass. We could simply the build script a bit by using e.g. `platform.python_implementation()` and adding a field to python version, but the approach remains.

> It actually works quite nicely, it did require to mess with the build script, and add a feature which renames link names for some functions, since pypy emulates the CPython API via a different convention of API names.

Do you have some further reading on that? The official pypy docs don't give any details about cpyext and searching the link names on google returns 2 to 5 results.

omerbenamram:
@konstin So the way cpyext work is that that you're supposed to link your cpython extension against pypy's version of `Python.h`, where it has a bunch of if-defs that reroute the cpython api calls to PyPy's implementation of the c-api layer.

konstin:
Replaced by #393

PR #371, opened 2019-02-26
Added an example of a Rust function returning a Python type.

thedrow:
I had to struggle for about 10 minutes to figure this out myself and I did not find such an example in one of your example code bases.

This PR includes an example of a Rust function returning a Python tuple.

### DIFF
--- a/guide/src/function.md
+++ b/guide/src/function.md
@@ -48,6 +48,21 @@ fn module_with_functions(py: Python, m: &PyModule) -> PyResult<()> {
 # fn main() {}
 ```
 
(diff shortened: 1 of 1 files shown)


konstin:
Thank you for the pull requests!

Two notes: I think you wanted to import `pyo3::PyTuple` and not `pyo3::Py`. pyo3 also can return types without wrapping them in a `PyResult` now, so it would be cool if the example would return a `Py<PyTuple>` instead of a `PyResult<Py<PyTuple>>`.

kngwyu:
I think only few methods returns `Py<PyType>`

konstin:
Closing because method signatures have changed in the meantime

PR #377, opened 2019-02-27
Change time_zone_info argument type for PyDateTime::from_timestamp() due to broken design

thedrow:
Since Rust has no inheritance, specifying `Option<&PyTzInfo>` as the argument type of time_zone_info prevents us entirely from specifying time zones since no time zone is of the type `PyTzInfo`.
The time zone struct may inherit from it (using Python inheritance) or it may be imported from Python.

It is possible to change the API of `from_timestamp()` to expect a `TzInfo` trait but that will require more extensive refactoring.

### DIFF
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -144,12 +144,12 @@ impl PyDateTime {
     pub fn from_timestamp(
         py: Python,
         timestamp: f64,
-        time_zone_info: Option<&PyTzInfo>,
+        time_zone_info: Option<&PyObject>,
     ) -> PyResult<Py<PyDateTime>> {
         let timestamp: PyObject = timestamp.to_object(py);
 
         let time_zone_info: PyObject = match time_zone_info {
-            Some(time_zone_info) => time_zone_info.to_object(py),
+            Some(time_zone_info) => time_zone_info,
             None => py.None(),
         };
 
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -149,8 +149,8 @@ impl PyDateTime {
         let timestamp: PyObject = timestamp.to_object(py);
 
         let time_zone_info = match time_zone_info {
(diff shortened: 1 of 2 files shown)


kngwyu:
We can use `extract` here, like
```rust
fn test_datetime_from_timestamp_utc() {
    use pyo3::types::{PyDateTime, PyTzInfo, IntoPyDict};

    let py = gil.python();
    let datetime = py.import("datetime").map_err(|e| e.print(py)).unwrap();
    let timezone = datetime.get("timezone").unwrap();
    let utc: &PyTzInfo = timezone.getattr("utc").unwrap().extract().unwrap();
    let dt = PyDateTime::from_timestamp(py, 1551279638., Some(&utc)).unwrap();
    let locals = [("dt", dt)].into_py_dict(py);
    let offset: f32 = py
        .eval("dt.utcoffset().total_seconds()", None, Some(locals))
        .unwrap()
        .extract()
        .unwrap();
    assert_eq!(offset, 0f32);
}
```
So if we can assume that `time_zone_info` is always an instance of a child class of `PyTzInfo`, there's no problem.
cc: @pganssle 
Do you think we need modification here?

thedrow:
What happens if we define the TzInfo class in Rust?

konstin:
`PyDateTime::from_timestamp` will fail because the rust `TzInfo` isn't python's `TzInfo`

pganssle:
> Do you think we need modification here? Is `time_zone_info`'s class always a child class of `PyTzInfo`?

The Python C API has a runtime check that the Python Object has type `tzinfo`. I would not want to *panic* if that's not true, but it's an error case for sure.

I think it's mildly preferable to have the safe Rust signature use `&PyTzInfo`, but I wasn't sure that it would be possible to use `tzinfo` objects from Python in Rust with that requirement. If it is possible, then I think we should change `PyDateTime::new` back to using `&PyTzInfo` as well.

PR #382, opened 2019-03-03
Remove unrequired encoding header

Alisa-lisa:
This is not required anymore for Python 3.

### DIFF
--- a/examples/word-count/word_count/__init__.py
+++ b/examples/word-count/word_count/__init__.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 from .word_count import WordCounter, count_line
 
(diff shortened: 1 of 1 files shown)


pganssle:
This library still supports Python 2 for at least a little while, though, I think.

Also, while not required I don't think it hurts to explicitly declare the encoding of the file even if it's the same as the default.

kngwyu:
Let's leave it until python2 support ends :)

konstin:
Yep, but I think we can keep this open until then.

milesgranger:
I'd just add that while it's the _default_ (utf-8) for Python3, so far as Python3 is concerned, it makes no difference, but IDEs will still make use of this declaration.

PR #387, opened 2019-03-04
Test examples in user guide with travis

### LINKED ISSUE #385: Reactivate doctests for guide
To avoid that the examples in the guide are in a broken state, I'd like to reactivate the tests which were ignored in https://github.com/PyO3/pyo3/commit/f9ff7cd465b53fde996d8a795e9700e144fe3cee. I think https://github.com/assert-rs/docmatic/issues/4 can be mitigated by changing .travis.yml and appveyor.yml so that `cargo clean` is executed before the tests.

Unfortunately, it seems that `extern crate pyo3;` has to be added to the examples in order to make the doctests pass (https://github.com/rust-lang/rust/issues/58875). At least we could hide them from the output with `#`. This would also fix the "Run this code" button in the guide which does not work at the moment.

If this sounds ok I could start working on this.

Alexander-N:
#385

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ features = ["extension-module"]
 **`src/lib.rs`**
 
 ```rust
+// Not required when using Rust 2018
+extern crate pyo3;
+
 use pyo3::prelude::*;
 use pyo3::wrap_pyfunction;
 
@@ -95,6 +98,9 @@ pyo3 = "0.6.0-alpha.4"
 Example program displaying the value of `sys.version`:
 
 ```rust
+// Not required when using Rust 2018
+extern crate pyo3;
+
 use pyo3::prelude::*;
 use pyo3::types::PyDict;
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
       env: FEATURES=python2
     - name: Python 3.5
       python: "3.5"
-      env: FEATURES=python3
+      env: FEATURES="python3 test-doc"
     - name: Python 3.6
       python: "3.6"
       env: FEATURES=python3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,9 @@ extension-module = []
 # are welcome.
 # abi3 = []
 
+# Use this feature to test the examples in the user guide
+test-doc = []
+
 [workspace]
 members = [
     "pyo3cls",
--- a/ci/travis/test.sh
+++ b/ci/travis/test.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -ex
 
+cargo clean
 cargo test --features "$FEATURES num-complex"
 if [ $TRAVIS_JOB_NAME = 'Minimum nightly' ]; then
     cargo fmt --all -- --check
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -6,6 +6,7 @@ To define a custom python class, a rust struct needs to be annotated with the
 `#[pyclass]` attribute.
 
 ```rust
+# extern crate pyo3;
 # use pyo3::prelude::*;
 
 #[pyclass]
@@ -35,6 +36,7 @@ You can get an instance of `PyRef` by `PyRef::new`, which does 3 things:
 
 You can use `PyRef` just like `&T`, because it implements `Deref<Target=T>`.
 ```rust
+# extern crate pyo3;
 # use pyo3::prelude::*;
 # use pyo3::types::PyDict;
 #[pyclass]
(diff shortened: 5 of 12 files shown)


README.md
```diff
@@ -51,6 +51,8 @@ features = ["extension-module"]
 **`src/lib.rs`**
 
 ```rust
+extern crate pyo3;
```
konstin: Please add a note in the readme that `extern crate pyo3;` is not required when using the 2018 edition

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -182,7 +182,7 @@ fn impl_wrap_init(cls: &syn::Type, name: &syn::Ident, spec: &FnSpec<'_>) -> Toke
         unsafe extern "C" fn __wrap(
             _slf: *mut ::pyo3::ffi::PyObject,
             _args: *mut ::pyo3::ffi::PyObject,
-            _kwargs: *mut ::pyo3::ffi::PyObject) -> libc::c_int
+            _kwargs: *mut ::pyo3::ffi::PyObject) -> ::std::os::raw::c_int
```
konstin: Is this required? I would really like to avoid having to stick with the 2015 edition for the derive only for the doctest. (We should be able pass `--extern=2018` to rustdoc to avoid this if it doesn't work otherwise)

konstin:
Thank you! I've left two comments, otherwise it's a that's a big improvement!

pyo3-derive-backend/src/pymethod.rs
Alexander-N: Even when passing the edition directly to rustdoc, I get
```
---- guide/src/class.md - Python_Class::Object_properties (line 232) stdout ----
error[E0433]: failed to resolve: use of undeclared type or module `libc`
  --> guide/src/class.md:240:1
   |
10 | #[pymethods]
   | ^^^^^^^^^^^^ use of undeclared type or module `libc`

thread 'guide/src/class.md - Python_Class::Object_properties (line 232)' panicked at 'couldn't compile the test', src/librustdoc/test.rs:354:13
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- guide/src/class.md - Python_Class::Object_properties (line 261) stdout ----
error[E0433]: failed to resolve: use of undeclared type or module `libc`
  --> guide/src/class.md:269:1
   |
10 | #[pymethods]
   | ^^^^^^^^^^^^ use of undeclared type or module `libc`

thread 'guide/src/class.md - Python_Class::Object_properties (line 261)' panicked at 'couldn't compile the test', src/librustdoc/test.rs:354:13

---- guide/src/class.md - Python_Class::Object_properties (line 289) stdout ----
error[E0433]: failed to resolve: use of undeclared type or module `libc`
 --> guide/src/class.md:292:1
  |
5 | #[pyclass]
  | ^^^^^^^^^^ use of undeclared type or module `libc`

thread 'guide/src/class.md - Python_Class::Object_properties (line 289)' panicked at 'couldn't compile the test', src/librustdoc/test.rs:354:13
```

README.md
Alexander-N: Done

Alexander-N:
On travis I'm getting weird linking errors like `/usr/bin/ld: cannot find -lpython3.6m`. Any help would be great.

konstin:
I don't know the exact problem or the solution, but I can give some context: To run the tests, we need to build binaries, which needs to link `libpython` in the version specified by the build script. Normally the build script also tells the linker the path of `libpyhon` through [rustc-link-search](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script):

https://github.com/PyO3/pyo3/blob/d63fac8d1cbdc3dcf4eefbbb871b3a6eb031a1b3/build.rs#L444-L448

This seems to fail here, presumably because the we're calling rustdoc here directly (through docmatic) and not cargo.

It's also possible to add directories to the linker path using environment variables, but that should have already been done by the script that travis runs:

https://github.com/PyO3/pyo3/blob/d63fac8d1cbdc3dcf4eefbbb871b3a6eb031a1b3/ci/travis/setup.sh#L16-L20 

The logs even tell that this path has been set: https://travis-ci.org/PyO3/pyo3/jobs/502189479#L322

I am not really sure what to do here because debugging travis is a tedious task. The quick and dirty solution would be to make the minimum nightly use python 3.5 (which apparently works) and then only run those test in the minimum nightly builds, just like rustfmt and clippy.

Alexander-N:
Ok, then I would suggest we activate the tests only for Python 3.5 for now. This way we have at least some mechanism in place which prevents the examples to drift from the current API. We could still leave #385 open and look for a better solution.

konstin:
I think that eventually, we want to either use a fix version of skeptic/docmatic or (which is more likely) write our own small doc tester. But for now this is a really good solution for moving forward with the docs.

Issue #356, opened 2019-02-15
Confusing APIs around PyObject


vadimcn:
I started learning PyO3 a couple of days ago, and for the longest time I was confused by the fact that `PyObject` represents a "free-floating" reference to a Python object, whereas the test of Py\<thing\>'s represent concrete Python types that are only accessible with `'py` lifetime.  

It might be too late now to change all that, but... it would have been much easier to grok PyO3's API structure if `PyObject` and `PyObjectRef` meanings were swapped.   Furthermore, I would suggest moving `PyObject`'s convenience methods to `Py` and making `PyObjectRef` just an alias for `Py<PyObject>`.

At the very least, all this should have been explained somewhere close to the top of the user guide.

kngwyu:
Thanks for reporting!
I think you misunderstand `Py`.
Please see [the guide](https://pyo3.rs/master/class.html#py) about it.

However, I completely agree with things around `PyObject`, `PyObjectRef` and so on are **really** confusing(in addition, now we have `PyRef` types!).
So anyway I think we should simplify our APIs and to make `PyObjectRef` just an alias sounds good.

kngwyu:
I changed your issue title but please feel free to change it again if you dislike it.

konstin:
I agree that `PyObject` is very confusing. Unfortunately it's difficult to at least rename `PyObject` to something more reasonable because there's `ffi::PyObject` (which make the situation even more confusing).

Afaik we could replace `PyObject` with `Py<PyObjectRef>` and maybe add alias like `PyStored` for convenience, but that would be a huge refactoring.

konstin:
I think we can avoid much of the confusion without changing the api by explaining the types properly in the guide.

vadimcn:
Well, the crate is still pre-1.0, but I agree, renaming PyObject now would break everybody's code.  
Still, I would suggest to at least consider renaming `PyObjectRef` to something like `PyBaseObject`.

Regarding `PyRef`: the distinction between built-in and user-defined Python objects seems like another wart in crate API.  Why not put native objects behind `PyRef` as well?   Also, calling it `PyBox` would be more intuitive, IMO.

jothan:
With the latest 0.6.0-alpha.4, I'm having trouble with storing references to other Python instances from within a pyclass. What is the proper type to use inside structs now ? With the current situation, I'm having trouble upgrading &self or &mut self to Py<Self>. Does the new API account for this ?

[...]

PR #388, opened 2019-03-04
Rename PyObjectRef with PyAny

kngwyu:
See https://github.com/PyO3/pyo3/issues/356#issuecomment-467724825 and around comments for the motivation.
Some related ideas:
- Maybe should we remove `FromPyObject` rather than renaming it? This trait [remains unchanged](https://github.com/dgrunwald/rust-cpython/blob/master/src/conversion.rs#L99) and doesn't fit to our semantics
- We should also rename `PyTryFrom` with `PyDowncast`?


pganssle:
As someone who doesn't really understand what `PyObjectRef` does at the moment, I will say that changing this name to `PyAny` doesn't help me much, if at all.

I wonder if it might be a good idea to *start* by adding clear documentation about the difference between `PyObject` and `PyObjectRef`, and *then* propose a change to the name?

konstin:
@pganssle I'm convinced that this is a better name and, in combination with more documentation, it will make more sense than `PyObjectRef`.

For short, the new name means that 

```python
from typing import List, Any


def foo(x: List, y: Any):
    ...
```

translates into

```rust
fn foo(x: &PyList, y: &PyAny) {
    unimplemented!()
}
```

konstin:
> Maybe should we remove FromPyObject rather than renaming it? This trait remains unchanged and doesn't fit to our semantics

It should be replaced by `FromPy` at some point.

> We should also rename PyTryFrom with PyDowncast?

I like the idea, but that should be discussed separately.

konstin:
I'd say we wait a bit before merging so everyone has time to add comments.

Oh, and @kngwyu could you please add a changelog entry?

pganssle:
@konstin 
> For short, the new name means that
> ...
> translates into
> ...

Yes, this is is a nice parallel, but this is very specific to the Python `typing` module. The equivalent for things *other than annotations* would be "derived from `PyObject`. I think you don't want to start using "mixed metaphors", where the Rust type system parallels type annotations in some respects, and parallels the C API / inheritance model in other respects. It will prevent newcomers from forming a useful mental model of how to use `pyo3` types.

I haven't looked into this closely, but it may be worth considering [other parts of the type system](https://docs.python.org/3/library/typing.html) to see if the parallels break down. or if they hold up. What about `Callable`? `Sequence`?

konstin:
As I've said in the issue thread, `PyObject` would indeed be the ideal name. Unfortunately that is already taken twice (for the ffi type and for our wrapper) and `PyAny` was second best I could come up with. If you have a better name, I'm happy to discuss that. 

`Callable` isn't implemented yet, `PySequence` exists. The [types module doc page](https://docs.rs/pyo3/0.6.0-alpha.4/pyo3/types/index.html#structs) has a good overview, and I'd say it's kinda consistent. 

> I think in the end you should only adopt this Any if the idea is that PyO3's type system will generally try to have a fairly consistent relationship to the typing type system, and that you want to abandon the idea of inheritance.

I don't want to base pyo3 on typing, I just want to use the resemblance to make it easier understandable (I'd also say that "any" makes sense even without `typing`). I don't see how this related to inheritance, especially given that pyo3 can only emulate inheritance.

kngwyu:
@pganssle 
I'm not the author, but I think the reason why both `PyObject` and `PyObjectRef` exist is mostly historical one.
`PyObject` remains unchanged since [rust-cpython](https://github.com/dgrunwald/rust-cpython/blob/master/src/objects/object.rs#L46) age and `PyObjectRef` is added after this project started.
`PyObjectRef` behaves like other python native types like `PyDict` and mostly we use it as `&'gil PyObjectRef`. It has a life time same as GIL, which is forced by our object stack.
Other than that, `PyObject` behaves like `Py<PyObjectRef>`. It's used as `PyObject` and not has a lifetime same as GIL, though it also does `refcnt -= 1` when the current GILGuard drops.

So while the idea that `PyObject` as the name of the current `PyObjectRef` is good, it needs a lots of refactoring, especially around `ToPyObject` and `IntoPyObject`.
But I think it's still possible and worth trying, though how to change `ToPyObject` is unclear.

PR #390, opened 2019-03-05
Fix flaky tests

### LINKED ISSUE #386: GIL tests are flaky
The following tests fail randomly, for example in https://travis-ci.org/PyO3/pyo3/jobs/501186502:
* `gil::test::test_owned`
* `gil::test::test_owned_nested`
* `gil::test::test_pyobject_drop`

The reason seems to be that they all take references of the empty Python tuple and make assertions for the expected reference count. This can fail if the tests are executed in parallel.

Alexander-N:
The tests were flaky because they took references to the empty tuple
which is a singleton. Use a unique object for each test to make them
independent.

Closes #386

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -334,14 +334,21 @@ mod test {
     use crate::ToPyObject;
     use crate::{ffi, gil};
 
-    fn get_object() -> PyObject {
+    fn get_object() -> (PyObject, *mut ffi::PyObject) {
         // Convenience function for getting a single unique object
         let gil = Python::acquire_gil();
         let py = gil.python();
 
-        let obj = py.eval("object()", None, None).unwrap();
+        let obj = py.eval("object()", None, None).unwrap().to_object(py);
+        let obj_ptr = obj.as_ptr();
 
-        obj.to_object(py)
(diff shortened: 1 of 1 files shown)


kngwyu:
I don't like using redundant `INCREF` here.
How about using an object which is not singleton, instead of `(,)`?

Alexander-N:
I don't follow, using separate objects instead of a singleton was my approach here. I don't like the `INCREF` either, could you explain your alternative a bit?

konstin:
Thank you for the pull request! Those flaky tests have been a long-lingering problem and it's good to finally have them fixed.

---

The `get_object()` function should return either the pointer or the PyObject, since the are identical (incidentally they are even identical in memory). I'd return only the PyObject and use `.as_ptr()` where you need the pointer.

The incref usage is fine, even though after a quick glance it looks like it is only actually required in one place. However you always need to call decref in the end, otherwise your leaking memory. Those `cnt` variables also shouldn't be required any more since we can just compare with 1 and 2. Having constants will also make sure we're not accidentally using a singleton again.

konstin:
That clone_ref solution is much better that INCREF'ing :+1:

Issue #276, opened 2018-11-15
Rewrite build.rs


konstin:
The current is a bit of a mess and should be rewritten.

 - [x] Currently, setting `PYTHON_SYS_EXECUTABLE` to python 2 requires the python2 feature to be additionally set, while this should be required.
 - [x] Currently, flags and options/values are parsed separately, there are special cases for different operating systems which we don't need and the manual string building and parsing is bad. We can do this much more simple and resilient using a single python script dumping json which is then read by serde. See [pyo3-pack](https://github.com/PyO3/pyo3-pack/blob/master/src/python_interpreter.rs) for an example. It also contains [multiple sysconfigs](https://github.com/PyO3/pyo3-pack/tree/master/sysconfig) showing the os specifics.
 - [ ] `PYTHON_SYS_EXECUTABLE` should be renamed into something more explicit like `PYO3_INTERPRETER`.
- [ ] Support static linking, at least on linux. `cargo:rustc-link-search=native` must be set to the value of  `LIBPL` in the python sysconfig and `cargo:rustc-link-lib=static=` must be used instead of `cargo:rustc-link-lib`. It should be possible to enable this with a feature flag.

The first point should be doable without prior knowledge by some basic `println`-style debugging. The second should be doable if you already know serde and what compile settings are by using the same structure as pyo3-pack. Having a second os installed is helpful for testing. The third should be done last and is a simple `s/PYTHON_SYS_EXECUTABLE/PYO3_INTERPRETER/g` with some extra notes in the documentation. The fourth point is optional and can still be finished later.

If you want to work on this issue and have questions, feel free to post them here.

kngwyu:
> We can do this much more simple and resilient using a single python script dumping json which is then read by serde. See pyo3-pack for an example. 

Should we extract related codes as library from pyo3-pack?

konstin:
No, I don't think there's enough shared code to build a library from it.

kngwyu:
> In the first point do we want to print some debugging line during build?

You mean the lines like
```rust
            println!(
                "cargo:warning={}",
                "Build may fail due to incompatible rustc version."
            );
```
?
They're just messages to cargo, not debug print.
Please see [this page](https://doc.rust-lang.org/cargo/reference/build-scripts.html#case-study-building-some-native-code).

Or, you're talking about actually printing some messages to stdout or else, while executing the build script? 
If so, I don't know how to do it.

eriol:
[...]

PR #393, opened 2019-03-09
Support rust extensions for PyPy via cpyext

omerbenamram:
Hi,
Following the discussion on https://github.com/PyO3/pyo3/pull/370, I've taken some time to dust off my work on support for PyPy via cpyext.

It runs nicely on my machine (osx) for the several test cases I've been using (raising errors, moving objects back and forth and so on).

I've managed to run the `word_count` example successfully, altough for this usecase pypy beats rust in terms of performance (overhead is greater than benefit).

~~Unfortunely, it seems the `rustapi_module` example is still crashing, because of 2 outstanding issues:~~ 
**EDIT: Both solved!**

~~1. The datetime API wrapper.~~
~~2. The way `initialize_type` behaves (causes SIGSEGV on attempt to register a new class)
My guess is that there is still some outstanding difference in the implementation of the `PyTypeObject` struct that causes it to crash~~
 
I'd like to outline some changes I've done, and why I think there is definitely some more work needed on this before it's ready for prime time.

1. cpyext does not offer complete support for the entirety of the cpython API. So I've had (mostly by dumping symbols out of the pypy interpreter) manually add the parts of it that it implements with annotations like this:
`#[cfg_attr(PyPy, link_name = "PyPyErr_SetInterrupt")]`

I couldn't find any automatic way to set this up, since it's not for all the symbols, so it's quite verbose (I've done this with a regexes), but it makes it very easy to spot changes when the cpyext api layer changes and add them.

2. I've never succeeded in getting this to work with python2, and since I've read in the PR that it's likely to be discontinued, I didn't invest more time with it.

3. I've had to manually shim some functions that pypy expects to find when compiling against their version of `Python.h`, like `PyUnicodeDecodeError_Create`, these are mostly utility functions which they never bothered implementing, so it looks something like this:
image

4. I've modified the `build.rs` script originally in my implementation, since I needed it to detect and correctly emit configuration variables that are suitable for PyPy, it's a slight drift from the current model (single `build.rs` file), into a more structured internal crate that I've used. I didn't want to spend time porting all the changes back into the big `build.rs` file without hearing your thoughts on it first.

5. As the title suggests, this only supports python -> rust, and not embedding pypy (due to pypy not behaving nicely with embedding), though I don't think embedding pypy is a significant use case, as it's mostly used to squeeze performance out of existing python codebases.

**EDIT**: Functionallity that is still missing from cpyext that is used by PyO3 internally:
These functions will need to be implemented in rust.
- complex numbers `_Py_c_*` functions
- `PyDict_MergeFromSeq2`
- `_PyLong_FromByteArray`, `_PyLong_AsByteArray`
- `PySequence_Count`
- `Py_CompileStringExFlags` (EDIT: PyPy exports Py_CompileStringFlags, so this is actually OK)

I'm still certain there are some semantics that I've missed, but i'd love to get another pair of eyes looking over some of this stuff.

Would love to hear your thoughts :) Cheers 🐍 

P.S I'm away for next week, but I will be able to address suggestions the week after.


kngwyu:
Really cool stuff, thanks!
Could you please enable PyPy3 build on travis?

omerbenamram:
@kngwyu added :) (though the tests should fail)
Best bet is to run the python examples, they are a pretty strong indicator that we are able to correctly compile and run rust extensions with pypy

src/gil.rs
```diff
@@ -53,7 +53,11 @@ pub fn prepare_freethreaded_python() {
             // Signal handling depends on the notion of a 'main thread', which doesn't exist in this case.
             // Note that the 'main thread' notion in Python isn't documented properly;
             // and running Python without one is not officially supported.
+
+            // PyPy does not support the embedding API
+            #[cfg(not(PyPy))]
             ffi::Py_InitializeEx(0);
+            #[cfg(not(Py_3_7))]
```
kngwyu: ?

kngwyu: I'm not sure but I think we have to call this function to create GIL.

omerbenamram: It seems that in `src/derive_utils.rs` there is this comment:
image

Since we're initializing here I assumed this shouldn't have been here.

kngwyu: I checked cpython code and it seems true, thanks.

Cargo.toml
```diff
@@ -26,16 +26,16 @@ pyo3cls = { path = "pyo3cls", version = "=0.6.0-alpha.4" }
 mashup = "0.1.9"
 num-complex = { version = "0.2.1", optional = true }
 inventory = "0.1.3"
-
+pretty_assertions = "0.5.1"
```
konstin: Is pretty_assertions used as a regular dependency?

src/ffi3/object.rs
```diff
@@ -594,7 +634,7 @@ pub struct PyType_Slot {
 
 impl Default for PyType_Slot {
     fn default() -> PyType_Slot {
-        unsafe { mem::zeroed() }
+        unsafe { ::std::mem::zeroed() }
```
konstin: Is this required?

src/gil.rs
konstin: I think we should copy that comment

src/macros.rs
```diff
@@ -0,0 +1,9 @@
+
+#[macro_export]
+/// Constructs a `&'static CStr` literal.
+macro_rules! cstr {
+    ($s: tt) => {
+        // TODO: verify that $s is a string literal without nuls
+        unsafe { ::std::ffi::CStr::from_ptr(concat!($s, "\0").as_ptr() as *const _) }
```
konstin: Sidenote: I'd prefer `CString::new("...").unwrap().as_ptr()` to that macro.

konstin:
Thank you, that's really cool!

> It runs nicely on my machine (osx) for the several test cases I've been using (raising errors, moving objects back and forth and so on).

If those cases are self-contained, it would be great to make them part of the examples.

> cpyext does not offer complete support for the entirety of the cpython API. So I've had (mostly by dumping symbols out of the pypy interpreter) manually add the parts of it that it implements with annotations like this:
> #[cfg_attr(PyPy, link_name = "PyPyErr_SetInterrupt")]

Seems like all of those have been fixed.

> I've modified the build.rs script originally in my implementation, since I needed it to detect and correctly emit configuration variables that are suitable for PyPy, it's a slight drift from the current model (single build.rs file), into a more structured internal crate that I've used. I didn't want to spend time porting all the changes back into the big build.rs file without hearing your thoughts on it first.

I've written down my long term wishes for the build script in #276. For the sake of this pull request, I think just adding a field to `PythonVersion` to the existing build.rs by querying `import platform; platform.python_implementation()` is the best solution. We also don't need a pypy feature, the cfg is enough.

Re travis ci: We can just add a special script that only test what we now is working with pypy, i.e. `cargo build` for the pyo3 crate and a tox/pytest run for word-count.

src/ffi3/object.rs
omerbenamram: nope, removed :) 
I think this was a hiccup from some old version.

Cargo.toml
omerbenamram: nope, removed.

src/macros.rs
omerbenamram: wouldn't `CString::new("...").unwrap().as_ptr()` deallocate at the end of the block due to itself being dropped too early?

src/gil.rs
omerbenamram: copied.

src/macros.rs
konstin: Oh your're right, I meant `CString::new("...").unwrap().into_raw()` ("Consumes the CString and transfers ownership of the string to a C caller.").

konstin: I've removed that macro in 09883574, which results in a minimal merge conflict.

omerbenamram: Nice! Creating `CStr` directly is much nicer, I'll merge those changes :)

omerbenamram: by the way, it seems adding a nul terminator is needed for those strings (otherwise it will error)

omerbenamram:
@konstin just letting you know `word_count` tests are passing locally against PyPy3.7-7.1, they are failing on travis because they are using an older release (PyPy3.5-6.0) which is missing some symbols.

I'm not sure it's worth putting effort into supporting older releases of PyPy3.5, since the whole thing is considered beta anyways. I'm trying to see if there's any way I could get travis to run the newer version, their version names are pretty confusing https://travis-ci.community/t/confusing-pypy-versions/2505

konstin:
I just realized I mixed up CStr and CString and broke stuff. I'll give this a proper fix and fix the resulting conflict with this branch.

omerbenamram:
@konstin the `rustapi_module` tests are now successfullly passing aswell 🎉 !
There was a pretty terrible bug in PyPy where passing a `\0` string as a docstring (`tp_doc`) crashes the interpreter.

I've now fixed this and all the tests seem to pass.
I still have 2 little things left:
- [ ] The integration test suite (`word_count`, `rustapi_module`) crashes only in tox (exits with error code -11), after finishing running all the tests successfully, but not with my installation (simply running `pytest`). Do you know anything about why this might happen?
- [ ] The `build.rs` stuff.

omerbenamram:
@konstin @kngwyu CI is finally green.
I'd love if you could take a look and see if everything seems OK 😄 

There are still the things that I've written above that aren't implemented in cpyext that I've posted above (complex numbers etc..), but I could handle those in a separate PR. 

I think this is enough to consider basic support for now.

examples/rustapi_module/tests/test_datetime.py
```diff
@@ -80,7 +83,8 @@ def test_invalid_date_fails():
         rdt.make_date(2017, 2, 30)
 
 
-@given(d=dates())
+# Feeding this tests dates from too early will cause `get_timestamp` to raise.
```
kngwyu: What does it mean?

src/ffi/datetime.rs
```diff
@@ -94,8 +105,47 @@ pub struct PyDateTime_CAPI {
     ) -> *mut PyObject,
 }
 
-// Type struct wrappers
+#[cfg(PyPy)]
+extern "C" {
+    #[cfg_attr(PyPy, link_name = "_PyPyDateTime_Import")]
```
kngwyu: Maybe just `link_name` is enough?

ci/travis/setup.sh
```diff
@@ -33,3 +33,15 @@ if [ ! -f "$HOME/.cargo/bin/kcov" ]; then
     install src/kcov $HOME/.cargo/bin/kcov
     cd $TRAVIS_BUILD_DIR
 fi
+
+
+### Setup PyPy ################################################################
+
+wget --quiet https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
```
kngwyu: Maybe we should use `if [[ $FEATURES == *"pypy"* ]]` here?

kngwyu:
Thanks for lots of works, @omerbenamram.
LGTM except a few minor things, though we still need @konstin's review

src/ffi/datetime.rs
omerbenamram: 👍

ci/travis/setup.sh
omerbenamram: 👍

examples/rustapi_module/tests/test_datetime.py
omerbenamram: I was getting crashes in the python test when hypothesis was feeding really early dates which are completely unrelated to the rust end of things. It might be a bug in PyPy's interpreter, since In cpython it just returns a large negative number.

I guess it's because trying to make dates from before the epoch (from timestamp) doesn't make a lot of sense.

image

kngwyu: I see, thanks.

ijl:
v7.1 is out: https://morepypy.blogspot.com/2019/03/pypy-v71-released-now-uses-utf-8.html.

It breaks the datetime module and needs different `cfg` or reimplementations of some functions.

examples/rustapi_module/tests/test_datetime.py
```diff
@@ -80,7 +83,8 @@ def test_invalid_date_fails():
...
+@given(d=st.dates(min_value=pdt.date(1900, 1, 1)))
```
pganssle: If this only happens on PyPy, can you make it conditional on that? Same with below.

Even better would be to make this test `xfail` on PyPy:

```python
# get_timestamp fails on Pypy if the dates are too early
if PYPY and dt < pdt.date(1900, 1, 1):
    pytest.xfail("get_timestamp will raise on PyPy with dates before 1900")
```

This seems like a bug in Pypy, and we should not encode these restrictions into our code for all platforms.

src/ffi/datetime.rs
```diff
@@ -16,6 +16,11 @@ use std::os::raw::{c_char, c_int, c_uchar};
 use std::ptr;
 use std::sync::Once;
 
+// A note regarding cpyext support:
```
pganssle: Should this be an actual docstring?

src/ffi/datetime.rs
```diff
@@ -62,11 +70,14 @@ pub struct PyDateTime_CAPI {
     #[cfg(Py_3_7)]
     pub TimeZone_FromTimeZone:
         unsafe extern "C" fn(offset: *mut PyObject, name: *mut PyObject) -> *mut PyObject,
+
+    // Defined for PyPy as a separate symbol
```
pganssle: These comments are confusing. Separate from what?

src/ffi/datetime.rs
```diff
@@ -201,13 +252,21 @@ impl Deref for PyDateTimeAPI {
 /// such as if you do not want the first call to a datetime function to be
 /// slightly slower than subsequent calls.
 pub unsafe fn PyDateTime_IMPORT() -> &'static PyDateTime_CAPI {
-    // PyDateTime_CAPSULE_NAME is a macro in C
-    let PyDateTime_CAPSULE_NAME = CString::new("datetime.datetime_CAPI").unwrap();
+    // PyPy excepts the C-API to be initialized via PyDateTime_Import, so trying to use
```
pganssle: s/excepts/expects

Also, I think this comment is *slightly* confusing, I would reword as:

```rust
// PyPy uses a different mechanism to initialize the datetime C API
// and so must call PyDateTime_Import
```

I can give perhaps better suggestion for the wording if you can explain what `PyDateTime_Import` is? Is that a PyPy-specific function that imports the datetime capsule?

src/ffi/datetime.rs
```diff
@@ -288,6 +347,7 @@ macro_rules! _access_field {
 
 // Accessor functions for PyDateTime_Date and PyDateTime_DateTime
 #[inline]
+#[cfg(not(PyPy))]
```
pganssle: For the accessor functions, might it make sense to do this in two blocks like this:

```rust
#[cfg(not(PyPy))]
{
...  // Current code
}

#[cfg(PyPy)] {
 #[link_name = "PyPyDateTime_DATE_GET_HOUR"]
    pub fn PyDateTime_DATE_GET_HOUR(o: *mut PyObject) -> c_int;
    #[link_name = "PyPyDateTime_DATE_GET_MICROSECOND"]
    pub fn PyDateTime_DATE_GET_MICROSECOND(o: *mut PyObject) -> c_int;
    #[link_name = "PyPyDateTime_DATE_GET_MINUTE"]
    ...
}

Seems like something that would be better than individually annotating every function.

File: src/types/datetime.rs [hunk 13]
```diff
@@ -83,11 +90,17 @@ impl PyDate {
...
     /// This is equivalent to `datetime.date.fromtimestamp`
     pub fn from_timestamp(py: Python, timestamp: i64) -> PyResult<Py<PyDate>> {
-        let args = PyTuple::new(py, &[timestamp]);
+        let time_tuple = PyTuple::new(py, &[timestamp]);
 
         unsafe {
-            let ptr = (PyDateTimeAPI.Date_FromTimestamp)(PyDateTimeAPI.DateType, args.as_ptr());
-            Py::from_owned_ptr_or_err(py, ptr)
+            #[cfg(PyPy)]
+            let ptr = PyDate_FromTimestamp(time_tuple.as_ptr());
```
pganssle: Can you clarify why this is necessary? Does this mean that on PyPy the `PyDateTimeAPI` struct doesn't have a FromTimestamp method? Why is this?

examples/rustapi_module/tests/test_subclassing.py
```diff
@@ -1,9 +1,13 @@
...
 
+PYPY = platform.python_implementation() == 'PyPy'
 
 class SomeSubClass(Subclassable):
     pass
 
 
-a = SomeSubClass()
-_b = str(a) + repr(a)
+if not PYPY:
```
pganssle: This should be moved into a test function, like:

```python
def test_subclass_repr():
    a = SomeSubclass()
    _b = str(a) + repr(a)
```

Or maybe into several tests, since it seems like it's testing more than one thing here.

When it's a separate test, it should be marked `xfail` on PyPy unless it's causing a segfault, in which case it should be skipped conditionally on PyPy.

src/types/datetime.rs
omerbenamram: this is what PyPy has for (as time of writing this 7.0) for datetime:
image

cpython has indeed some extra methods:
image

So I've had to go around the current cpyext API and call those methods directly.

src/ffi/datetime.rs
omerbenamram: So from what I understand cpython expects users to import the datetime C-API via `PyDateTime_IMPORT`, which is a macro that imports the capsule.

PyPy only tries to emulate the API, so it exposes the name of the macro as a seperate symbol, which it expects to be called.
So importing the capsule directly (which has no effect in cpython since this is a pure macro) does not work. 

cpython:
image

helpful comment:
image

If you have better suggestions for wording to make this clearer I'll be happy to head 😄

examples/rustapi_module/tests/test_datetime.py
omerbenamram: 👍

examples/rustapi_module/tests/test_subclassing.py
omerbenamram: Putting this in a test function is what I've originally done (being a good boy-scout), but this actually causes cpython to segfault, not pypy https://github.com/PyO3/pyo3/issues/407.
So I suggest handling this separately.

omerbenamram:
@ijl I'll give it a spin, thanks!

src/ffi/datetime.rs
omerbenamram: If we wanted to do this, we would need something like `cfg-if` (since we cant just arbitrarly `#[cfg]` blocks on this scope).

image

Do you think it's worth adding for this case?

src/ffi/datetime.rs
omerbenamram: changed it to better reflect how to access these methods on PyPy

src/ffi/datetime.rs
pganssle: But the capsule *is* the C API. The macros aren't even available outside of C (which is why we have to define them here ourselves).

Does PyPy not have a mechanism to emulate capsule imports as well?

src/types/datetime.rs
pganssle: @omerbenamram Can you use actual blocks of text instead of screen shots from now on?  You can still get code highlighting with blocks of 3 backticks.

As for why this is different, I'm a bit stumped. Maybe we can make a PR to PyPy.

The thing is that the capsule object *is* part of the public interface for PyO3, so ideally we want that to be a seamless experience for users who are using the `ffi` bindings directly.

src/ffi/datetime.rs
omerbenamram: It does export the symbol `PyCapsule_Import`, but when I've tried it the module was not initialized.

It seems they are doing some extra work when "PyDateTime_IMPORT" is called via an _PyDateTime_IMPORT internal symbol.

(source for the internal initialization code)
image

Also, this is the test testing cpyext support for datetime, notice how what is defined as a macro in cpython, is considered a whole different beast here (when linking against their version of Python.h)
image

src/ffi/datetime.rs
pganssle: For something like this I think either `cfg-if` or a custom macro would be worth it, but I'll defer to whatever @konstin thinks.

Another possibility is that I think we don't actually *need* to call the PyPy's implementation of these macros, since in this case we're not actually making calls to the C API, we're just *replicating* the C API in Rust. I'm pretty sure we can just leave all of this stuff alone.

src/types/datetime.rs
omerbenamram: I'm not sure I get your last point about the capsule object being a public interface, from what I've understood (as a user of the library) - the public API is `PyDateTime` (https://github.com/PyO3/pyo3/blob/master/src/types/datetime.rs), which was where I've felt that I could bridge the differences.

This is an internal decision by PyPy, I'm not entirely sure why they chose to implement it this way 😕 But I don't really see a way to rewire these functions to behave the same.

src/ffi/datetime.rs
pganssle: Let me look into this a bit more closely. My understanding is that even though `datetime` is one of the only standard library modules to actually use them, capsules are the preferred mechanism for handling C API imports.

This particular conditional compilation doesn't need to be a blocker or anything, but it might be worth trying to support capsule imports upstream in PyPy, if only because it makes `cpyext` a better compatibility layer for C extensions.

src/ffi/datetime.rs
omerbenamram: This is why I've ignored all the other macros (`PyDateTime_CheckExact` for example) in many other places, because they're shallow enough so they behave the same.

But `PyDateTime_GET_YEARS` for example does seem to access some internals of structs, so I wasn't sure that it's completely OK to assume it will behave the same.

src/ffi/datetime.rs
omerbenamram: awesome, let me know if there's anything you find out about this! 

This may be a specific case with `datetime` they implemented specific behavior for, but I guess if they intend to support other c-extensions they would need to make it behave the same.

examples/rustapi_module/tests/test_datetime.py
```diff
@@ -80,8 +83,11 @@ def test_invalid_date_fails():
         rdt.make_date(2017, 2, 30)
 
 
-@given(d=dates())
+@given(d=st.dates())
 def test_date_from_timestamp(d):
+    if PYPY and d < pdt.date(1900, 1, 1):
+        pytest.xfail("get_timestamp will raise on PyPy with dates before 1900")
```
pganssle: I'm not sure if this is hypothesis messing with things, but I'm not actually seeing any `xfail`s in the test suite.

I looked into this to see if it was a bug in PyPy, and basically all the PyPy C API stuff is thin wrappers around calls to Python itself, and at least on Linux, PyPy *does* support large negative values in` fromtimestamp`.

I think this may actually be platform dependent and may not have anything to do with PyPy. @omerbenamram what platform are you seeing the failures for? Windows? Mac?

omerbenamram: Mac

omerbenamram:
```
latform darwin -- Python 3.5.3[pypy-7.1.0-alpha], pytest-4.3.0, py-1.8.0, pluggy-0.9.0
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=True warmup_iterations=100000)
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/omerba/Workspace/pyo3/examples/rustapi_module/.hypothesis/examples')
rootdir: /Users/omerba/Workspace/pyo3/examples/rustapi_module, inifile:
plugins: faulthandler-1.5.0, benchmark-3.2.2, hypothesis-4.9.0
collected 60 items

tests/test_datetime.py ..XFalsifying example: test_date_from_timestamp(d=datetime.date(1899, 12, 31))
x.ssxxxxXXXXXXX........sX.Falsifying example: test_datetime_from_timestamp(dt=datetime.datetime(1899, 1, 1, 0, 0))
x.......................
tests/test_dict_iter.py    1/  64 iterations:0=>0
```

src/types/datetime.rs
pganssle: @omerbenamram I mean that the unsafe bindings are also part of the public interface, not just the safe Rust ones. If people are calling the `ffi::datetime::PyDateTimeAPI` functions directly, they'll have to do the compatibility shims for PyPy themselves, which is not good.

I think ideally we can make a PR to PyPy to add these to the C API. We may also be able to do something like populate our `PyDateTimeAPI` object with function pointers to the relevant `cpyext` functions, if necessary, though I think that might make the ABI less stable than it already is (not sure what guarantees are made about ABI compatibility here).

build.rs
```diff
@@ -450,27 +493,28 @@ fn configure(interpreter_version: &PythonVersion, lines: Vec<String>) -> Result<
...
 
+    if interpreter_version.implementation == PythonInterpreterKind::PyPy {
+        println!("cargo:rustc-cfg=PyPy");
+        flags += format!("CFG_PyPy").as_ref();
+    };
+
     if let PythonVersion {
         major: 3,
         minor: some_minor,
+        implementation: implementation,
```
pganssle: Looks like there's a warning about this being unused? https://travis-ci.org/PyO3/pyo3/jobs/511140863#L649

I'm seeing a lot of other warnings in that build as well, not sure how many of them are new with this PR.

examples/rustapi_module/tests/test_datetime.py
omerbenamram: However on cpython (3.7)
```
platform darwin -- Python 3.7.1, pytest-4.3.1, py-1.8.0, pluggy-0.9.0
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/omerba/Workspace/pyo3/examples/rustapi_module/.hypothesis/examples')
rootdir: /Users/omerba/Workspace/pyo3/examples/rustapi_module, inifile:
plugins: benchmark-3.2.2, hypothesis-4.14.0
collected 60 items

tests/test_datetime.py .......xxxx..........................................
```

build.rs
omerbenamram: it's some `if let` patterns where I've simply added this is (but isn't actually needed inside). I'll make a round and clean up these warnings :)

src/ffi/datetime.rs
pganssle: @omerbenamram Hm.. That's a good point. I am not 100% clear about this, but I *think* you already are making these assumptions because the structures [are defined in Rust in `ffi/datetime.rs`](https://github.com/PyO3/pyo3/blob/27fb0f3fd0c685d7a72a52239d605e9fbeacf5e2/src/ffi/datetime.rs#L155), and the "macros" are operating on these internal structures.

It seems like there would be many other problems if the C representations of these objects is different depending on whether you constructed it in Rust or in PyPy.

I think it's probably safe to remove the PyPy-specific stuff here, but I'm willing to be convinced otherwise.

examples/rustapi_module/tests/test_datetime.py
pganssle: @omerbenamram The ones with the big X are xpasses, so evidently we haven't quite got it right. I'm not seeing either xpasses *or* xfails on PyPy on Travis, though, which is very puzzling to me.

build.rs
omerbenamram: there are still a few imports (CStr), that rust complains that are unused, but if deleted cause a compile error (about missing CStr). This is strage (maybe it's my nightly version?)

examples/rustapi_module/tests/test_datetime.py
omerbenamram: you can see it's printing `Falsifying example` which I believe it's hypothesis' way of telling the test failed?

src/ffi/datetime.rs
omerbenamram: It does break the tests (good thing that there are tests :)!

This is with removing the PyPy (`PyDateTime_GET_*`) macros.

```
=================================================================================================================================== FAILURES ====================================================================================================================================
______________________________________________________________________________________________________________________________ test_date_accessors ______________________________________________________________________________________________________________________________

    @given(d=st.dates())
>   def test_date_accessors(d):

tests/test_datetime.py:73:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

d = datetime.date(2000, 1, 1)

    @given(d=st.dates())
    def test_date_accessors(d):
        act = rdt.get_date_tuple(d)
        exp = (d.year, d.month, d.day)

>       assert act == exp
E       assert (0, 0, 0) == (2000, 1, 1)
E         At index 0 diff: 0 != 2000
E         Use -v to get the full diff

tests/test_datetime.py:77: AssertionError
---------------------------------------------------------------------------------------------------------------------------------- Hypothesis -----------------------------------------------------------------------------------------------------------------------------------
Falsifying example: test_date_accessors(d=datetime.date(2000, 1, 1))
___________________________________________________________________________________________________________________________________ test_time ___________________________________________________________________________________________________________________________________

    @given(t=st.times())
>   def test_time(t):

tests/test_datetime.py:112:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

t = datetime.time(0, 0, 0, 1)

    @given(t=st.times())
    def test_time(t):
        act = rdt.get_time_tuple(t)
        exp = (t.hour, t.minute, t.second, t.microsecond)

>       assert act == exp
E       assert (0, 0, 0, 0) == (0, 0, 0, 1)
E         At index 3 diff: 0 != 1
E         Use -v to get the full diff

tests/test_datetime.py:116: AssertionError
---------------------------------------------------------------------------------------------------------------------------------- Hypothesis -----------------------------------------------------------------------------------------------------------------------------------
Falsifying example: test_time(t=datetime.time(0, 0, 0, 1))
______________________________________________________________________________________________________________________________ test_datetime_tuple ______________________________________________________________________________________________________________________________

    @given(dt=st.datetimes())
>   def test_datetime_tuple(dt):

tests/test_datetime.py:194:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

dt = datetime.datetime(2000, 1, 1, 0, 0)

    @given(dt=st.datetimes())
    def test_datetime_tuple(dt):
        act = rdt.get_datetime_tuple(dt)
        exp = dt.timetuple()[0:6] + (dt.microsecond,)

>       assert act == exp
E       assert (0, 0, 0, 0, 0, 0, ...) == (2000, 1, 1, 0, 0, 0, ...)
E         At index 0 diff: 0 != 2000
E         Use -v to get the full diff

tests/test_datetime.py:198: AssertionError
---------------------------------------------------------------------------------------------------------------------------------- Hypothesis -----------------------------------------------------------------------------------------------------------------------------------
Falsifying example: test_datetime_tuple(dt=datetime.datetime(2000, 1, 1, 0, 0))
_____________________________________________________________________________________________________________________________ test_delta_accessors ______________________________________________________________________________________________________________________________

    @given(td=st.timedeltas())
>   def test_delta_accessors(td):

tests/test_datetime.py:264:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

td = datetime.timedelta(0, 0, 1)

    @given(td=st.timedeltas())
    def test_delta_accessors(td):
        act = rdt.get_delta_tuple(td)
        exp = (td.days, td.seconds, td.microseconds)

>       assert act == exp
E       assert (1, 0, 0) == (0, 0, 1)
E         At index 0 diff: 1 != 0
E         Use -v to get the full diff

tests/test_datetime.py:268: AssertionError
---------------------------------------------------------------------------------------------------------------------------------- Hypothesis -----------------------------------------------------------------------------------------------------------------------------------
Falsifying example: test_delta_accessors(td=datetime.timedelta(0, 0, 1))
===================================================================================================== 4 failed, 38 passed, 3 skipped, 6 xfailed, 9 xpassed in 2.35 seconds ======================================================================================================
```

omerbenamram:
@pganssle also from looking at the official docs: https://docs.python.org/3/c-api/datetime.html
It seems that the `PyDateTime_CAPI` is strictly an implementation detail, it does not appear in any form as documented public API for cpython, which is why I believe pypy took the liberty with the internals.

kngwyu:
Still we have a warning
image

ijl:
pyo3-pack support in https://github.com/PyO3/pyo3-pack/pull/105

src/ffi/datetime.rs
pganssle: Hm.. Interesting. Does this mean that the various datetime structs we define are unused in PyPy?

In any case, if it's true that we need these alternate macro definitions, I think maybe something like `[cfg-if]` will be nicer and easier to maintain.

@konstin Thoughts?

pganssle:
> @pganssle also from looking at the official docs: https://docs.python.org/3/c-api/datetime.html
It seems that the PyDateTime_CAPI is strictly an implementation detail

The C API is, generally speaking, a bunch of implementation details that have been exposed, and it's a bit of a mess but everything documented in the "C API" is actually a bunch of macros defined in `Python.h`. If we restricted ourselves to that, we'd need a C compatibility layer that wraps those in symbols that actually get exported, or we'd need to do something like PyPy does and fake the C API by just calling out to Python.

In reality, everyone uses the `PyDateTime_CAPI` capsule, and it's effectively part of the public interface. I can probably get it officially documented in CPython fairly easily, but I think for now we should consider it public and make use of it as we can.

By the way, thank you for all the work you are putting into this and for responding to my sometimes off-the-mark comments. I apologize if I'm creating more work for you here, and your responses have been very instructive to me.

omerbenamram:
@pganssle I understand, thanks again for your time reviewing this!

I'm not sure contributing to cpyext itself is something I'm quite capable of doing, but if there's anything you think we can do with the rust end of things to provide a better API for users I'll be happy to try and get it though.

konstin:
> In reality, everyone uses the PyDateTime_CAPI capsule, and it's effectively part of the public interface. I can probably get it officially documented in CPython fairly easily, but I think for now we should consider it public and make use of it as we can.

It would be great if you could get it to be part of the docs

build.rs
```diff
@@ -501,6 +545,12 @@ fn version_from_env() -> Option<PythonVersion> {
     // sort env::vars so we get more explicit version specifiers first
     // so if the user passes e.g. the python-3 feature and the python-3-5
     // feature, python-3-5 takes priority.
+    let interpreter_kind = if cfg!(feature = "pypy") {
```
konstin: Does this pypy feature still exist?

src/ffi/datetime.rs
konstin: I don't think cfg-if works in that scope, and I don't have a problem with annotating functions individually. However it would be possible to group those functions into modules (that are re-exported from other modules) and then annotate the modules, though I don't know if that's worth it.

src/ffi3/ceval.rs
```diff
@@ -20,30 +21,39 @@ pub unsafe fn PyEval_CallObject(func: *mut PyObject, arg: *mut PyObject) -> *mut
...
     pub fn Py_MakePendingCalls() -> c_int;
+    #[cfg_attr(PyPy, link_name = "PyPy_SetRecursionLimit")]
     pub fn Py_SetRecursionLimit(arg1: c_int) -> ();
+    #[cfg_attr(PyPy, link_name = "PyPy_GetRecursionLimit")]
     pub fn Py_GetRecursionLimit() -> c_int;
     fn _Py_CheckRecursiveCall(_where: *mut c_char) -> c_int;
     static mut _Py_CheckRecursionLimit: c_int;
 }
 
+//#[cfg_attr(PyPy, link_name = "PyPy_EnterRecursiveCall")]
```
konstin:
```suggestion

```

src/ffi3/object.rs
```diff
@@ -43,12 +73,23 @@ pub unsafe fn Py_REFCNT(ob: *mut PyObject) -> Py_ssize_t {
...
+    );
+}
+
+#[inline(always)]
 pub unsafe fn Py_TYPE(ob: *mut PyObject) -> *mut PyTypeObject {
     (*ob).ob_type
 }
 
-#[inline]
+#[inline(always)]
```
konstin: Is there a reason for that change?

src/ffi3/object.rs
```diff
@@ -584,6 +699,7 @@ mod typeobject {
 
 // The exported types depend on whether Py_LIMITED_API is set
 pub use self::typeobject::*;
+use std::ffi::CStr;
```
konstin: Please move that import to the top

pganssle:
@omerbenamram
> I'm not sure contributing to cpyext itself is something I'm quite capable of doing, but if there's anything you think we can do with the rust end of things to provide a better API for users I'll be happy to try and get it though.

No worries, I can do that part.

@konstin 
> It would be great if you could get it to be part of the docs

Yes, I will raise an issue about this in a few weeks when I have some time.

build.rs
omerbenamram: nope, only as CFG flags, removing this.

src/ffi3/object.rs
omerbenamram: no, reverted.

src/ffi3/object.rs
omerbenamram: 👍

omerbenamram:
@konstin I've addressed your comments, and I've added a small section to the guide about installation instructions.

what do you think?

guide/src/pypy.md
```diff
@@ -0,0 +1,11 @@
+# PyPy Support
+
+Using PyPy is supported via cpyext.
+
+Support is only provided for building rust extension for code running under PyPy. This means PyPy **cannot** be called from rust via cpyext.
+ 
+This is a limitation of cpyext and supported for embedding cpyext is not planned.
+
+Compilation against PyPy is done by exporting the `PYTHON_SYS_EXECUTABLE` to a pypy binary.
```
konstin:
```suggestion
Compilation against PyPy is done by exporting the `PYTHON_SYS_EXECUTABLE` to a pypy binary or by compiling in a PyPy virtualenv.
```

guide/src/pypy.md
```diff
@@ -0,0 +1,11 @@
+# PyPy Support
+
+Using PyPy is supported via cpyext.
+
+Support is only provided for building rust extension for code running under PyPy. This means PyPy **cannot** be called from rust via cpyext.
```
konstin:
```suggestion
Support is only provided for building rust extension for code running under PyPy. This means PyPy **cannot** be called from rust via cpyext. Note that there some differences in the ffi module between pypy and cpython. 
```

konstin:
Docs look good, thank you!

There were some warnings and compilation failed for pypy3.6. I've fixed both and exported them as gist, which should be easily `git apply`-able: https://gist.github.com/konstin/ee00d9aaae44117dbae6f4f5eb2998eb (I think I can also push to this branch if that doesn't work).

pganssle:
I dunno what the plan is with respect to the merge strategy, but would it be possible to squash this, or otherwise clean up the history into related changes? Right now there are 128 commits, which would be something like 10% of the commit history if merged without squashing.

omerbenamram:
@pganssle I think squashing should be probably a good idea.

omerbenamram:
@pganssle @konstin BTW it seems that they haven't added support for date fold for the 3.6 release, so I'll apply the patch and also change the test to make it pass

```
 !  ~/W/pyo3   master-pypy *$-  examples/rustapi_module  nm -gU /tmp/pypy3.6-v7.1.0-osx64/bin/libpypy3-c.dylib | rg -i date                                                                                                                         Thu Mar 28 22:17:29 2019
000000000233f650 S _PyPyDateTimeAPI
00000000005adb20 T _PyPyDateTime_Check
00000000005adb40 T _PyPyDateTime_CheckExact
00000000005addc0 T _PyPyDateTime_DATE_GET_HOUR
00000000005ade20 T _PyPyDateTime_DATE_GET_MICROSECOND
00000000005adde0 T _PyPyDateTime_DATE_GET_MINUTE
00000000005ade00 T _PyPyDateTime_DATE_GET_SECOND
00000000005adec0 T _PyPyDateTime_DELTA_GET_DAYS
00000000005adf00 T _PyPyDateTime_DELTA_GET_MICROSECONDS
00000000005adee0 T _PyPyDateTime_DELTA_GET_SECONDS
00000000005add00 T _PyPyDateTime_FromTimestamp
00000000005adda0 T _PyPyDateTime_GET_DAY
00000000005add80 T _PyPyDateTime_GET_MONTH
00000000005add60 T _PyPyDateTime_GET_YEAR
00000000005ade40 T _PyPyDateTime_TIME_GET_HOUR
00000000005adea0 T _PyPyDateTime_TIME_GET_MICROSECOND
00000000005ade60 T _PyPyDateTime_TIME_GET_MINUTE
00000000005ade80 T _PyPyDateTime_TIME_GET_SECOND
00000000005adb60 T _PyPyDate_Check
00000000005adb80 T _PyPyDate_CheckExact
00000000005add20 T _PyPyDate_FromTimestamp
00000000005ac070 T _PyPyDict_Update
00000000005adcb0 T __PyPyDateTime_FromDateAndTime
00000000005adb10 T __PyPyDateTime_Import
00000000005adc60 T __PyPyDate_FromDate
```

konstin:
From my point of view, we can finally merge this :sparkler: 

@pganssle Anything outstanding issues with the datetime part?

Unless there are objections or somebody wants to rebase & squash the commits, I'd use github's "Squash and merge".

pganssle:
@konstin I'll probably need 1 more day to look over the datetime parts. I think the most critical remaining question is the fact that as it stands now, the `ffi` API is different between PyPy and CPython, which I'm not amazingly comfortable with. I'm hoping there might be a decent hack that can give the end user the same `ffi` interface, like possibly creating wrapper functions in Rust and assigning them to the relevant struct entries as part of the import function in PyPy.

ci/travis/cover.sh
```diff
@@ -2,6 +2,12 @@
 
 set -ex
 
+### PyPy does not run the test suite ###########################################
+
+if ! [[ $FEATURES == *"pypy"* ]]; then
```
pganssle: Did I miss something? Why isn't the test suite run in Pypy?

guide/src/pypy.md
```diff
@@ -0,0 +1,11 @@
+# PyPy Support
+
+Using PyPy is supported via cpyext.
+
+Support is only provided for building rust extension for code running under PyPy. This means PyPy **cannot** be called from rust via cpyext. Note that there some differences in the ffi module between pypy and cpython. 
+ 
+This is a limitation of cpyext and supported for embedding cpyext is not planned.
+
+Compilation against PyPy is done by exporting the `PYTHON_SYS_EXECUTABLE` to a pypy binary or by compiling in a PyPy virtualenv.
```
pganssle: When I try `cargo build` after setting `export PYTHON_SYS_EXECUTABLE=$(which pypy3.5)`, I am seeing failures:

>    Compiling pyo3cls v0.6.0 (.../pyo3/pyo3cls)
error: could not find native static library `python3.5`, perhaps an -L flag is missing?

Similar results for `pypy3.6`. Is this something I'm doing wrong?

omerbenamram: This looks like a regression in the build script. It is supposed to look for `libpypy3-c` instead of `libpython3.x`, I'll look into it later today, thanks for the catch!

ci/travis/cover.sh
omerbenamram: It is currently impossible to initialize pypy from rust, and call cpyext (PyPy doesn't  expose the `PyInitialize` symbol).
If you'll try to run a test binary it will immediately segfault.

However if we could expand the test suite that uses rust from PyPy we could hit more code paths, and that is a good start :)

guide/src/pypy.md
pganssle: Can we add at least a run with `cargo build` to the run, to prevent regressions like this in the future? If we can't say that the tests will pass, we should at least be able to check that the project builds.

ci/travis/cover.sh
pganssle: Hm.. I guess it doesn't have to block things, but I am personally uncomfortable with the idea of skipping the *entire* test suite on PyPy, we've already seen one regression, and the `rustapi_module` tests only cover a small fraction of the library.

ijl: Simple tests in Python can be run using pytest. I would appreciate that being done, because when I tried this branch on a project, the build succeeded but Python tests segfaulted.

guide/src/pypy.md
omerbenamram: added, good idea

ci/travis/cover.sh
omerbenamram: @ijl do you know what failed? can we add it to the test suite?

guide/src/pypy.md
omerbenamram: @pganssle could you please check that it's working on your machine?

pganssle: Still broken for me with your latest commits, plus now I'm seeing some warnings about unused imports and such:

> error: could not find native static library `pypy3-c`, perhaps an -L flag is missing?

If this is not some issue on my local environment, then it also seems like the tests are not effectively testing the `cargo build` command on PyPy. I'm using [pyenv](https://github.com/pyenv/pyenv) with `pypy3.5-7.0.0` and using `export PYTHON_SYS_EXECTUABLE=$(which pypy3.5)`. Perhaps the test suite needs to also export the `PYTHON_SYS_EXECUTABLE` variable?

omerbenamram: I've set cargo build  explicitly the env.
The examples tests are using `--cfg=PyPy` inside the setup.py.

Btw, what platform are you using? I cannot reproduce it locally.
It seems it's looking for the right dylib this time.
Where is it located in your environment? it should be named `libpypy3-c.dylib` or `libpypy3-c.so`.

omerbenamram: @pganssle it should be really fixed this time, and actually testing the build in the CI.
Thanks for lending a hand in testing this again, and sorry about it!

src/ffi/datetime.rs
kngwyu: I think it would be good to expose this document after adding some more explaination about `PyDateTime_CAPI` itself

src/ffi3/object.rs
```diff
@@ -466,6 +520,69 @@ mod typeobject {
         pub tp_next: *mut PyTypeObject,
     }
 
+    #[cfg(PyPy)]
```
kngwyu: On train, I came up with a simpler way.
```rust
    macro_rules! _type_object_init {
        ({$($head:tt)*} $tp_as_async:ident, $($tail:tt)*) => {
            as_expr! {
                PyTypeObject {
                    $($head)*
                    tp_name: ptr::null(),
                    tp_basicsize: 0,
                    tp_itemsize: 0,
                    tp_dealloc: None,
                    tp_print: None,
                    tp_getattr: None,
                    tp_setattr: None,
                    $tp_as_async: ptr::null_mut(),
                    tp_repr: None,
                    tp_as_number: ptr::null_mut(),
                    tp_as_sequence: ptr::null_mut(),
                    tp_as_mapping: ptr::null_mut(),
                    tp_hash: None,
                    tp_call: None,
                    tp_str: None,
                    tp_getattro: None,
                    tp_setattro: None,
                    tp_as_buffer: ptr::null_mut(),
                    tp_flags: ffi3::object::Py_TPFLAGS_DEFAULT,
                    tp_doc: ptr::null(),
                    tp_traverse: None,
                    tp_clear: None,
                    tp_richcompare: None,
                    tp_weaklistoffset: 0,
                    tp_iter: None,
                    tp_iternext: None,
                    tp_methods: ptr::null_mut(),
                    tp_members: ptr::null_mut(),
                    tp_getset: ptr::null_mut(),
                    tp_base: ptr::null_mut(),
                    tp_dict: ptr::null_mut(),
                    tp_descr_get: None,
                    tp_descr_set: None,
                    tp_dictoffset: 0,
                    tp_init: None,
                    tp_alloc: None,
                    tp_new: None,
                    tp_free: None,
                    tp_is_gc: None,
                    tp_bases: ptr::null_mut(),
                    tp_mro: ptr::null_mut(),
                    tp_cache: ptr::null_mut(),
                    tp_subclasses: ptr::null_mut(),
                    tp_weaklist: ptr::null_mut(),
                    tp_del: None,
                    tp_version_tag: 0,
                    $($tail)*
                }
            }
        }
    }

    #[cfg(PyPy)]
    macro_rules! py_type_object_init {
        ($tp_as_async:ident, $($tail:tt)*) => {
            _type_object_init!({
                    ob_refcnt: 1,
                    ob_pypy_link: 0,
                    ob_type: ptr::null_mut(),
                    ob_size: 0,
                }
                $tp_as_async,
                tp_pypy_flags: 0,
                $($tail)*
            )
        }
    }

    #[cfg(not(PyPy))]
    macro_rules! py_type_object_init {
        ($tp_as_async:ident, $($tail:tt)*) => {
            _type_object_init!({
                ob_base: ffi3::object::PyVarObject {
                    ob_base: ffi3::object::PyObject_HEAD_INIT,
                    ob_size: 0
                },}
                $tp_as_async,
                $($tail)*
            )
        }
    }
```

src/types/datetime.rs
konstin: >  If people are calling the ffi::datetime::PyDateTimeAPI functions directly, they'll have to do the compatibility shims for PyPy themselves, which is not good.

While I agree that it would be ideal if cpython and pypy exposed the same functions in their c api, I think that it's ok that users have to care for compatibility themselves when using `ffi` directly, as pyo3 is providing a safe and uniform abstraction.

pganssle: > While I agree that it would be ideal if cpython and pypy exposed the same functions in their c api, I think that it's ok that users have to care for compatibility themselves when using ffi directly, as pyo3 is providing a safe and uniform abstraction.

Yeah, that's probably true, but if we *can* provide a better interface, I think we should. One of the reasons `cpyext` even exists is so that pre-existing extension modules will just work without having to explicitly target pypy. If you're explicitly targeting pypy, my understanding is that a `cffi`-style approach is better than a C extension approach. This is why I think it would be best to abstract away the differences as much as possible.

pganssle:
I think the unification of the `ffi` API is more likely to cause problems than not, I'm just going to try to fix this on the PyPy side, so this can go ahead.

src/ffi3/object.rs
omerbenamram: Cool, I think this highlights the differences in the head struct a bit better :)
Updated.

omerbenamram:
@pganssle Cool!

Then if everything looks good you could go ahead and merge it.

Thanks!

omerbenamram:
fixed the issues with tox.
Everything is now checked against the pypy version as expected.

OK with me to merge :)

omerbenamram:
@konstin anything else needed?

kngwyu:
I'll merge this PR waiting after @konstin's reaction 1 day.

pganssle:
@kngwyu Please don't forget to squash merge!

src/ffi/datetime.rs
konstin:
```suggestion
    // PyPy expects the C-API to be initialized via PyDateTime_Import, so trying to use
```

konstin:
> I think the unification of the ffi API is more likely to cause problems than not, I'm just going to try to fix this on the PyPy side, so this can go ahead.

Big :+1: for fixing this in pypy

> @konstin anything else needed?

It's ready to merge from my point of view. 

I went through the unresolved conversations and found two minor things: 

 * @pganssle Is https://github.com/PyO3/pyo3/pull/393/files#r268843941 resolved?
 * @omerbenamram https://github.com/PyO3/pyo3/pull/393/files#r272995675 wouldn't hurt to make that a docstring

omerbenamram:
@konstin do you think it may be better to move https://github.com/PyO3/pyo3/pull/393/files#r272995675 to the module level documentation? I could also beef it up a little bit and let @pganssle  go over it.

konstin:
Yes, that sounds good

pganssle:
> @pganssle Is https://github.com/PyO3/pyo3/pull/393/files#r268843941 resolved?

It's not *resolved* per se, but I don't see it as a blocking issue here. The skipping is sufficiently well-scoped that I think it doesn't undermine the effectiveness of the test, and we can just open a new issue about why that's happening (it may not have anything to do with PyPy, just happens to manifest in PyPy).

> @konstin do you think it may be better to move https://github.com/PyO3/pyo3/pull/393/files#r272995675 to the module level documentation? I could also beef it up a little bit and let @pganssle go over it.

Happy to look it over, but hopefully it will be fairly ephemeral, as I've [already fixed this issue on PyPy for 2.7 and 3.6](https://bitbucket.org/pypy/pypy/commits/d5b44bb7b84d799cb7bc2f7dd530502c40619229?at=py3.6). Depending on how far back you want to support versions of PyPy, you can drop at least this part of the difference in the API as soon as the next PyPy release.

omerbenamram:
@pganssle https://bitbucket.org/pypy/pypy/commits/d5b44bb7b84d799cb7bc2f7dd530502c40619229?at=py3.6 looks awesome!
I just pushed some notes to the module level.

Feel free to revise them as you see fit, since you clearly have a broader outlook on the subject, and when 3.6 support will be added, we could simply use this version.

kngwyu:
> Happy to look it over, but hopefully it will be fairly ephemeral, as I've already fixed this issue on PyPy for 2.7 and 3.6. 

Wow it's great

src/ffi/datetime.rs
```diff
@@ -5,16 +5,29 @@
...
+//! PyPy's internal representation of dates differs from that of cpython.
+//! It expects to initialize some internal rpython machinery when the symbol `PyDateTime_Import` is called.
+//! After initialization, the emulation layer expects that all calls to the api will be limited to those defined in
+//! `https://docs.python.org/3/c-api/datetime.html`.
+//!
+//! For example, macros such as `PyDate_FromTimestamp` and `PyDateTime_FromTimestamp` are exported as symbols in `libpypy`.
+//! This means we cannot rely on emulating those macros in rust, since it doesn't ensure ABI comparability with the c structs.
+//!
+//! All functions that belong to `PyDateTime_CAPI`, are exported as private symbols in PyPy (ex. `_PyPyDate_FromDate`).
+//! This means that their behavior is subjected to change in upcoming PyPy versions.
```
pganssle:
```suggestion
//! This means that their behavior is subject to change in upcoming PyPy versions.
```

src/ffi/datetime.rs
```diff
@@ -5,16 +5,29 @@
 //! This is the unsafe thin  wrapper around the [CPython C API](https://docs.python.org/3/c-api/datetime.html),
 //! and covers the various date and time related objects in the Python `datetime`
 //! standard library module.
+//!
+//! (PyPy 7.0.0) - A note regarding PyPy (cpyext) support:
+//!
+//! PyPy's internal representation of dates differs from that of cpython.
```
pganssle: I'm not sure I understand this section or the action that a user is intended to take from it. Using the C API struct *is* supported for PyPy, right? It's just that not all of the functions are available, and you've added one additional variable in its place.

I think the stuff about the import logic and the stuff about the macros in the C API documentation are not relevant here, just keep it simple and say that there is limited function for some of the PyDatetimeCAPI functions in earlier versions of PyPy, and specifically `DateTime_FromTimestamp` and `Date_FromTimestamp` are not currently supported (we can specify the version that they are supported for later, when it comes out).

I'm torn about the `PyDate_FromTimestamp` and `PyDateTime_FromTimestamp` methods. I think don't mention them in this note. Ideally they would go away entirely.

src/types/datetime.rs
```diff
@@ -156,11 +165,18 @@ impl PyDateTime {
         let args = PyTuple::new(py, &[timestamp, time_zone_info]);
 
         unsafe {
-            let ptr = (PyDateTimeAPI.DateTime_FromTimestamp)(
-                PyDateTimeAPI.DateTimeType,
-                args.as_ptr(),
-                ptr::null_mut(),
-            );
+            #[cfg(PyPy)]
+            let ptr = PyDateTime_FromTimestamp(args.as_ptr());
```
pganssle: It occurs to me that this may lead to a panic or segfault.

I believe the current PyPy implementation of `PyDateTime_FromTimestamp` *requires* that you call `PyDatetime_Import()` before calling this function. All the other functions don't have this requirement because it's done once at global scope in the implementation of `PyDatetimeAPI`, so dereferencing that object at all guarantees that Import() has been called.

I think maybe the easiest short-term solution for this is to create a wrapper function around `PyDateTime_FromTimestamp` in the `ffi::datetime` module that just calls import (or dereferences `PyDateTimeAPI`) and then calls the function from PyPy.

Obviously this is under-tested as well, but I think because it involves global state it might be a bigger undertaking to add proper testing to all these functions.

I'm having trouble building the project again (`cannot find -lpypy3-c`), so I'm having trouble testing whether or not this causes problems.

src/ffi/datetime.rs
omerbenamram: I see your point regarding what users would potentially care about.
The original comment was more of a note for of a code comment to help me reason about why the rational of this api.

I've removed most of it and mentioned only `DateTime_FromTimestamp` and `Date_FromTimestamp`.

src/types/datetime.rs
konstin: > I'm having trouble building the project again (cannot find -lpypy3-c), so I'm having trouble testing whether or not this causes problems.

Have you tried building in a fresh pypy venv? It sometimes seems that a venv works better than `PYTHON_SYS_EXECUTABLE`.

pganssle: @konstin I think it was an issue with conflicts in my `pyenv` setup. I do a lot of library development so my `pyenv` global has a bunch of different things in it. When I set `pyenv shell` to just the pypy version I want it will build, then I just have to adjust `LD_LIBRARY_PATH` to get it to *run*.

In any case, here's an MWE to reproduce the issue, I can confirm it produces a segfault on pypy3.5:

```rust
extern crate pyo3;

use pyo3::prelude::*;
use pyo3::types::PyDateTime;

fn main() {
    let py = gil.python();
    let x = PyDateTime::from_timestamp(py, 123456.0, None);

}
```

It does not segfault on Python 3.7.2.

pganssle: Ah, my mistake, I can't reproduce the segfault using extension modules, I guess these particular symbols don't rely on the `_PyDateTime_Import` being called.

omerbenamram: It seems PyPy will import datetime internally when this is called.
```
@cpython_api([PyObject], PyObject)
def PyDateTime_FromTimestamp(space, w_args):
    """Create and return a new datetime.datetime object given an argument tuple
    suitable for passing to datetime.datetime.fromtimestamp().
    """
    w_datetime = PyImport_Import(space, space.newtext("datetime"))
    w_type = space.getattr(w_datetime, space.newtext("datetime"))
    w_method = space.getattr(w_type, space.newtext("fromtimestamp"))
    return space.call(w_method, w_args)
```

pganssle: Yeah, I think `PyDatetime_Import` is needed to initialize the CAPI object, but not when you use the macro directly (but only for PyPy), because of the way it's implemented.

I remember I had some segfaults when I didn't do the import when I was implementing the feature for pypy, which is why I was worried, but I realize now that was because I was dereferencing `PyDatetimeCAPI`, not because I was calling the `fromtimestamp` function.

konstin:
Big thanks to everyone involved!

:sparkles: :shipit:

PR #396, opened 2019-03-13
Fix some typos in user guide

Alexander-N:
Fix some typos and refer to PyO3 in a consistent way instead of alternating between different spellings.


konstin:
I'm not a fan of the capitalization, but it's better if it's consistently that way.

PR #403, opened 2019-03-17
Complete __dict__ support

### LINKED ISSUE #123: #[pyclass(dict)] makes objects without __dict__
I tried to mitigate problem of storing &PyObject in pyo3 class struct by going back to Python, using `__dict__` instance storage. Documentation mentions single-token attribute to py::class macros. I have studied code that uses this attribute and found it is sophisticated machinery, using special Python C-API construct in PyType header. Respect for special handling of `__dict__`, AFAIK you could get away with heap allocated PyDict and provide access via getter.

I guess it's some minor scholar mistake again, please point in right direction.

```rust
#[py::class(dict)]
struct Hub {
    token: PyToken,
}

#[py::methods]
impl Hub {
    #[new]
    fn __new__(obj: &PyRawObject, py: Python) -> PyResult<()> {
        obj.init(|token| Hub { token: token })
    }
}
```

```
% ../venv-27/bin/python -c 'import ext ; print(vars(ext.Hub()))
TypeError: vars() argument must have __dict__ attribute

Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64
rustc 1.25.0-nightly (27a046e93 2018-02-18)
Python 2.7.10
setuptools-rust 0.8.3
```

Alexander-N:
Add `__dict__` slot to tp_getset of type_object. See
https://bugs.python.org/issue16272 for why this is necessary.

Closes #123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
  * A soudness hole where every instances of a `#[pyclass]` struct was considered to be part of a python object, even though you can create instances that are not part of the python heap. This was fixed through `PyRef` and `PyRefMut`. 
+ * Fix kwargs support in [#328].
+ * Add full support for `__dict__` in [#403].
 
 ## [0.5.3] - 2019-01-04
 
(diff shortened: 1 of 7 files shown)


konstin:
Thank you!

PR #408, opened 2019-03-20
Update tests & docs with IntoPyDict::into_py_dict(py)

### LINKED ISSUE #378: Use IntoPyDict in examples and tests if possible
In examples and tests, we have many codes like
```rust
    let d = PyDict::new(py);
    d.set_item("obj", obj).unwrap();
    py.run("# some python codes", None, Some(d)).unwrap();
```
But now we can write this shortly, like
```rust
    use pyo3::types::IntoPyDict;
    let d = [("obj", obj)].into_py_dict(py);
    py.run("# some python codes", None, Some(d)).unwrap();
```
I think we should recommend users to use `slice.into_py_dict` pattern, when the dict is relatively small(with 1~2 keys). It's easier to read than `set_item`.
So first I think we should rewrite our examples, though it's a bit tough :roll_eyes:

milesgranger:
Will close #378

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,15 +103,14 @@ Example program displaying the value of `sys.version`:
 extern crate pyo3;
 
 use pyo3::prelude::*;
-use pyo3::types::PyDict;
+use pyo3::types::IntoPyDict;
 
 fn main() -> PyResult<()> {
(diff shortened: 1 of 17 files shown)


konstin:
Thank you

PR #419, opened 2019-03-28
Allow slf: PyRef<Self>/PyRefMut<Self> in pymethods

kngwyu:
Looks like 
```rust
#[pymethods]
impl Reader {
    fn get_iter(slf: PyRef<Self>, keys: Py<PyBytes>) -> PyResult<Iter> {
        Ok(Iter {
            reader: slf.into(),
            keys: keys,
            idx: 0,
        })
    }
}
```
and can be called like usual methods from python.
Resolves https://github.com/PyO3/pyo3/issues/356#issuecomment-465742221
Currently only `slf: PyRef` is tested.
Another concern is I observed an undesirable segfault when using `arg: PyRef<PyBytes>` in pymethod(though I'm not sure it's related to this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
  * Have `PyModule` generate an index of its members (`__all__` list).
+ * Allow `slf: PyRef<T>` for pyclass(#419)
 
 ### Changed
 
--- a/pyo3-derive-backend/src/method.rs
+++ b/pyo3-derive-backend/src/method.rs
@@ -27,6 +27,35 @@ pub enum FnType {
     FnCall,
     FnClass,
     FnStatic,
+    PySelf(PySelfType),
+}
+
+/// For fn(slf: &PyRef<Self>) support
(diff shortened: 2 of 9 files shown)


konstin:
In general for the self type I think it's better if we had a trait (or maybe reuse extract) that is implemented for all three types instead of trying to guess the right method based on the string-representation of the type

kngwyu:
>  I think it's better if we had a trait 

Experimentally I introduced a new trait named `FromPyPointer`, which registers the pointer to GIL pool and returns the reference of it.
If you come up with a better design, please let me know.

konstin:
I'll have to check if we already have the traits that can do what FromPyPointer does or with we can replace some other code with the new FromPyPointer, but otherwise it's looking good

kngwyu:
@konstin 
I want this PR move forward.
Though `FromPyPointer` is not implemented for `Py` and `PyObject`, we should fix this after #446.
It remove lots of `Py::from_owned_ptr*`s.

konstin:
Merging master should make travis ci pass again (or at least only fail for minimum nightly)

kngwyu:
Now CI passed, thanks

PR #424, opened 2019-04-01
[WIP] Add PyTypeObject impl to PyRef

konstin:
As @DeflatedPickle noticed on gitter, there was no way to get the `PyType` from the associated type of a `PyRef`. 

Open question: Do we want to add the following?

```rust
impl<T> AsPyRef<T> for T
where
    T: PyTypeInfo,
{
    #[inline]
    fn as_ref(&self, _py: Python) -> PyRef<T> {
        PyRef::from_ref(self)
    }
    #[inline]
    fn as_mut(&mut self, _py: Python) -> PyRefMut<T> {
        PyRefMut::from_mut(self)
    }
}
```

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -1,4 +1,3 @@
-// Copyright (c) 2017-present PyO3 Project and Contributors
 use crate::err::{PyErr, PyResult};
 use crate::ffi;
 use crate::gil;
@@ -69,8 +68,8 @@ where
 }
 
 impl<'a, T: PyTypeInfo> PyRef<'a, T> {
-    pub(crate) fn from_ref(r: &'a T) -> Self {
-        PyRef(r, PhantomData)
+    pub fn from_ref(reference: &'a T) -> Self {
+        PyRef(reference, PhantomData)
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/instance.rs
```diff
@@ -69,8 +68,8 @@ where
 }
 
 impl<'a, T: PyTypeInfo> PyRef<'a, T> {
-    pub(crate) fn from_ref(r: &'a T) -> Self {
-        PyRef(r, PhantomData)
+    pub fn from_ref(reference: &'a T) -> Self {
```
kngwyu: I'm not for this change.
PyRef exists because we cannot always get `*mut ffi::PyObject` from `&T whre T: PyTypeInfo`.
So this change allows `pyclass` to use `as_ptr` without copying it to Python heap

kngwyu:
> Open question: Do we want to add the following?

I **strongly** disagree. It's unsafe and misleading.

I think `get_type_from_object` is OK.
But I don't understand what other parts of this PR(`pub fn from_ref` or so) is for and need more explanation.

konstin:
I need to rethink this with a proper approach

PR #425, opened 2019-04-02
Add `FromPyObject` and `Clone` impls for every `PyNativeType`

althonos:
This PR adds `FromPyObject` and `Clone` to structs implementing `PyNativeType`:

* `FromPyObject`  is only as complicated as checking the `PyAny` pointer can be downcasted and then wrapping the pointer if successful
* `Clone` only requires to call `clone_ref` using the internal token.

I am still not super at ease with the whole holding-the-GIL contract, so I'm not sure if I'm actually breaking everything or just providing the right implementation. 

I had to change the `$name: ty` to `$name: ident` in several macros because Rust does not accept `$name(...thing...)` when expanding even if it is valid for tuple types.

## Side effect 

Structs with native types in their fields implemented with `#[pyclass]` trying to use `#[pyo3(get, set)] should now compile properly:
```rust
#[pyclass]
pub struct DateWrapper {
    #[pyo3(get, set)]
    date: PyDateTime,
}
```

## Maintenance

 - Run `cargo fmt` (This is checked by travis ci) :white_check_mark: 
 - Run `cargo clippy` and check there are no hard errors (There are a bunch of existing warnings; This is also checked by travis) :white_check_mark: 
 - If applicable, add tests for all new or fixed functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-## Fixed
+### Added
+
+ * Added `FromPyObject` and `Clone` implementations for Python native types.
+
+### Fixed
 
  * `type_object::PyTypeObject` has been marked unsafe because breaking the contract `type_object::PyTypeObject::init_type` can lead to UB.
  * Fixed automatic derive of `PySequenceProtocol` implementation in [#423](https://github.com/PyO3/pyo3/pull/423).
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -86,6 +86,17 @@ macro_rules! pyobject_native_type (
                 unsafe{&*(ob as *const $name as *const $crate::types::PyAny)}
             }
         }
+
+        impl Clone for $name {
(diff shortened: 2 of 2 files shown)


pganssle:
I think this needs tests, at least in Rust. The side effect probably can also be tested in rustapi_module, but I'm not sure what was broken about the pyclass struct before this PR.

althonos:
@pganssle : The derived `setter` did not work because `FromPyObject` was not available, and the derived getter did not work because `Clone` was not available.

kngwyu:
How about using `Py<PyDateTime>`?
All Py~ types except PyObject are designed to use as `&Py~`, so it's invalid to convert PyObject to Py~ types without registering it to GILPool(I'm from phone now and will add more explanation later).

althonos:
@kngwyu : yes I understood that while tinkering, Python objects should stay behind `Py` references. The only problem is that `#[pyo3(get, set)]` still doesn't work because `Py` does not implement `Clone` (although it could implicitly acquire the GIL, and then clone).

althonos:
Closing since this is not at all a good solution. I will open a PR later and try to fix that on the `pyo3-derive-backend` level.

PR #432, opened 2019-04-13
Fix documentation-related bugs

althonos:
Hi,

After trying to get PyO3 working with Sphinx, I noticed the following bugs:

* [x] Having unescaped double quotes in a doc comment would cause the derive macros to fail, because
  of how the doc comment were handled in `pyo3-derive-backend`
* [x] The `__doc__` attribute was not properly set for `#[getter]` methods.
* [ ] The `__doc__` attribute is not properly set for `#[new]` methods.

This PR fixes the two first issues and adds tests for the second bugfix.

## Maintenance

 - Run `cargo fmt` :white_check_mark: 
 - Run `cargo clippy` and check there are no hard errors :white_check_mark: 
 - If applicable, add an entry in the changelog :white_check_mark: 


kngwyu:
> Having unescaped double quotes in a doc comment would cause the derive macros to fail

I think we also should have a test case for this problem, but otherwise looks good.

althonos:
@kgnwyu: sure, will add one.

PR #434, opened 2019-04-14
PyFloat_AS_DOUBLE()

ijl:
This is a reimplementation of https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Include/floatobject.h#L50-L52.

### DIFF
--- a/src/ffi3/floatobject.rs
+++ b/src/ffi3/floatobject.rs
@@ -1,6 +1,13 @@
 use crate::ffi3::object::*;
 use std::os::raw::{c_double, c_int};
 
+#[cfg(not(Py_LIMITED_API))]
+#[repr(C)]
+pub struct PyFloatObject {
+    pub ob_base: PyObject,
+    pub ob_fval: c_double,
+}
+
 #[cfg_attr(windows, link(name = "pythonXY"))]
 extern "C" {
     pub static mut PyFloat_Type: PyTypeObject;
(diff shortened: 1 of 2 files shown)


src/ffi3/floatobject.rs
```diff
@@ -1,6 +1,12 @@
 use crate::ffi3::object::*;
 use std::os::raw::{c_double, c_int};
 
+#[repr(C)]
```
kngwyu: We also need `#[cfg(not(Py_LIMITED_API))]` here

src/types/floatob.rs
```diff
@@ -103,4 +103,14 @@ mod test {
...
+
+    #[test]
+    fn test_as_double_macro() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let v = 1.23f64;
+        let obj = v.to_object(py);
+        assert_eq!(v, unsafe { PyFloat_AS_DOUBLE(obj.as_ptr()) });
+    }
+
```
kngwyu: Just a nit but please remove this empty line.

ijl:
@kngwyu ok, resolved.

ijl:
The last two pushes have had failures on "Collecting coverage data". master is affected in the same way.

althonos:
@ijl : yes, my PR was successful but the same commit when merged in master was not because of this. I tried deleting the cache but it didn't change anything. Will try to debug that.

ijl:
Ok, well can this merge to master?

kngwyu:
Thanks.
I think we should fix the coverage problem, but I don't think we should wait for the fix, since this PR is not related to it.

PR #438, opened 2019-04-17
Make tons of small fixes in the guide

birkenfeld:
- spelling/grammar
- consistent casing of names (Rust/Python)
- update docs.rs links and Cargo.toml examples to 0.6.0
- fix a few factual mistakes I found in the process

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -1,8 +1,8 @@
-# Python Class
+# Python Classes
 
 ## Define new class
 
-To define a custom python class, a rust struct needs to be annotated with the
+To define a custom Python class, a Rust struct needs to be annotated with the
 `#[pyclass]` attribute.
 
 ```rust
@@ -18,6 +18,7 @@ struct MyClass {
 The above example generates implementations for `PyTypeInfo` and `PyTypeObject` for `MyClass`.
 
 ## Get Python objects from `pyclass`
+
 You can use `pyclass`es like normal rust structs.
 
 However, if instantiated normally, you can't treat `pyclass`es as Python objects.
@@ -25,13 +26,14 @@ However, if instantiated normally, you can't treat `pyclass`es as Python objects
 To get a Python object which includes `pyclass`, we have to use some special methods.
 
 ### `PyRef`
+
 `PyRef` is a special reference, which ensures that the referred struct is a part of
 a Python object, and you are also holding the GIL.
 
 You can get an instance of `PyRef` by `PyRef::new`, which does 3 things:
-1. Allocate a Python object in the Python heap
-2. Copies the rust struct into the Python object
-3. Returns a reference of it
+1. Allocates a Python object in the Python heap
+2. Copies the Rust struct into the Python object
+3. Returns a reference to it
 
 You can use `PyRef` just like `&T`, because it implements `Deref<Target=T>`.
 ```rust
@@ -52,6 +54,7 @@ dict.set_item("obj", obj).unwrap();
 ```
 
 ### `PyRefMut`
+
 `PyRefMut` is a mutable version of `PyRef`.
 ```rust
 # use pyo3::prelude::*;
@@ -67,7 +70,8 @@ obj.num = 5;
 ```
 
 ### `Py`
-`Py` is a object wrapper which stores an object longer than the GIL lifetime.
+
+`Py` is an object wrapper which stores an object longer than the GIL lifetime.
 
 You can use it to avoid lifetime problems.
 ```rust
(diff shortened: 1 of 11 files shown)


kngwyu:
Thanks.
I'm not good at English so personally I feel this kind of PR really helpful.

guide/src/class.md
```diff
@@ -222,12 +224,12 @@ impl MyClass {
...
-ways how to override name.
+A getter or setter's function name is used as the property name by default. There are several
+ways how to override the name.
 
-If function name starts with `get_` or `set_` for getter or setter respectively.
-Descriptor name becomes function name with prefix removed. This is useful in case of
-rust's special keywords like `type`.
+If a function name starts with `get_` or `set_` for getter or setter respectively,
+the descriptor name becomes the function name with this prefix removed. This is also useful in case of
+Rust keywords like `type` (raw identifiers can be used since Rust 2018).
```
konstin:
```suggestion
Rust keywords like `type` ([raw identifiers](https://doc.rust-lang.org/nightly/edition-guide/rust-2018/module-system/raw-identifiers.html) can be used since Rust 2018).
```

guide/src/building-and-distribution.md
```diff
@@ -32,11 +32,11 @@ my_main_crate = { path = "..", default-features = false }
...
 
 ## Distribution
 
-There are two ways to distribute your module as python package: The old [setuptools-rust](https://github.com/PyO3/setuptools-rust) and the new [pyo3-pack](https://github.com/pyo3/pyo3-pack). setuptools-rust needs some configuration files (`setup.py`,  `MANIFEST.in`, `build-wheels.sh`, etc.) and external tools (docker, twine). pyo3-pack doesn't need any configuration files. It can not yet build sdist though ([pyo3/pyo3-pack#2](https://github.com/PyO3/pyo3-pack/issues/2)).
+There are two ways to distribute your module as a Python package: The old is [setuptools-rust](https://github.com/PyO3/setuptools-rust) and the new is [pyo3-pack](https://github.com/pyo3/pyo3-pack). setuptools-rust needs some configuration files (`setup.py`, `MANIFEST.in`, `build-wheels.sh`, etc.) and external tools (docker, twine). pyo3-pack doesn't need any configuration files. It can not yet build sdist though ([pyo3/pyo3-pack#2](https://github.com/PyO3/pyo3-pack/issues/2)).
```
pganssle: I think adding "is" here is a grammatical mistake, it sounds awkward. In the original wording "old" and "new" are simple adjectives describing the two packages, now there is an elided "way" in both that I don't think sounds great.

Maybe we can split the difference by adding commas?

"The old, setuptools-rust, and the new, pyo3-pack."

guide/src/class.md
```diff
@@ -19,19 +19,21 @@ struct MyClass {
...
+
 `PyRef` is a special reference, which ensures that the referred struct is a part of
 a Python object, and you are also holding the GIL.
 
 You can get an instance of `PyRef` by `PyRef::new`, which does 3 things:
-1. Allocate a Python object in the Python heap
-2. Copies the rust struct into the Python object
+1. Allocates a Python object in the Python heap
+2. Copies the Rust struct into the Python object
 3. Returns a reference of it
```
pganssle:
```suggestion
3. Returns a reference to it
```

guide/src/class.md
```diff
@@ -92,25 +96,24 @@ assert_eq!(obj.as_ref(gil.python()).num, 1);
 
 ## Customizing the class
 
-The `#[pyclass]` macro accepts following parameters:
+The `#[pyclass]` macro accepts the following parameters:
 
-* `name=XXX` - Set the class name shown in python code. By default struct name is used as a class name.
-* `freelist=XXX` - `freelist` parameter add support of free allocation list to custom class.
+* `name=XXX` - Set the class name shown in Python code. By default, the struct name is used as the class name.
+* `freelist=XXX` - `freelist` parameter adds support of free allocation list to custom class.
```
pganssle:
```suggestion
* `freelist=XXX` - The `freelist` parameter adds support of free allocation list to custom class.
```

Not required, but I do think this sounds better.

guide/src/class.md
```diff
@@ -140,19 +143,19 @@ Rules for the `new` method:
...
 * The first parameter is the raw object and the custom `new` method must initialize the object
-  with an instance of the struct using `init` method. The type of the object may be the type object of
+  with an instance of the struct using the `init` method. The type of the object may be the type object of
   a derived class declared in Python.
-* The first parameter implicitly has type `&PyRawObject`.
-* For details on `parameter-list`, see the documentation of `Method arguments` section.
-* The return type must be `PyResult<T>` for some `T` that implements `IntoPyObject`. Usually, `T` will be `MyType`.
+* The first parameter must have type `&PyRawObject`.
+* For details on the parameter list, see the `Method arguments` section below.
+* The return value must be `T` or `PyResult<T>` where `T` is ignored, so it can
```
pganssle: This changes the actual meaning of the sentence. What does it mean that `T` is ignored?

Are you saying that nothing uses the return value of this function unless it's a `PyResult<T>`, which is used to throw exceptions, but not to return a value?

guide/src/class.md
```diff
@@ -194,15 +197,14 @@ impl SubClass {
 }
 ```
 
-`ObjectProtocol` trait provides `get_base()` method. It returns reference to instance of
-base class.
+The `ObjectProtocol` trait provides a `get_base()` method, which returns a reference
+to the instance of the base struct.
```
pganssle:
```suggestion
to an instance of the base struct.
```

Unless I'm misunderstanding this sentence - is there a specific instance this refers back to?

guide/src/class.md
birkenfeld: I'll put in the non-nightly link.

guide/src/building-and-distribution.md
birkenfeld: I'll defer to your judgment :)

guide/src/class.md
```diff
@@ -435,22 +439,22 @@ impl MyClass {
...
+By default, PyO3 uses function signatures to determine which arguments are required. Then it scans
+the incoming `args` and `kwargs` parameters. If it can not find all required
 parameters, it raises a `TypeError` exception. It is possible to override the default behavior
-with `#[args(...)]` attribute. `args` attribute accepts a comma separated list of parameters in
-form of `attr_name="default value"`. Each parameter has to match the method parameter by name.
+with the `#[args(...)]` attribute. This attribute accepts a comma separated list of parameters in
+the form of `attr_name="default value"`. Each parameter has to match the method parameter by name.
 
-Each parameter could be one of following type:
+Each parameter can be one of following type:
```
pganssle:
```suggestion
Each parameter can be one of following types:
```

guide/src/class.md
```diff
@@ -435,22 +439,22 @@ impl MyClass {
...
 
- * "\*": var arguments separator, each parameter defined after "*" is keyword only parameters.
-   corresponds to python's `def meth(*, arg1.., arg2=..)`
- * args="\*": "args" is var args, corresponds to python's `def meth(*args)`. Type of `args`
+ * `"*"`: var arguments separator, each parameter defined after "*" is keyword only parameters.
```
pganssle:
```suggestion
 * `"*"`: var arguments separator, each parameter defined after "*" is a keyword-only parameter.
```

guide/src/class.md
birkenfeld: As I understand it, it is the specific instance that forms part of the derived object.

guide/src/class.md
birkenfeld: Also adding a "the" here...

guide/src/class.md
pganssle: Ah, withdrawn.

guide/src/class.md
kngwyu: > What does it mean that T is ignored?

Agreed

pganssle:
I will not have time to finish reviewing this until after PyCon (mid-May), but from what I did review, the average quality is quite high, and I think this is a net positive for the project, regardless of minor errors.

Not sure if @konstin and @kngwyu are waiting on me to merge this, but if you are, feel free to go ahead and merge. This changes so many things that it's going to have constant conflicts, and it's better to merge early and have later cleanup PRs than to have to keep fixing conflicts.

@birkenfeld This does need a rebase, though, looks like several conflicts have already emerged. ☹

birkenfeld:
I'll rebase, but then I'd appreciate a speedy merge.

guide/src/class.md
birkenfeld: > Are you saying that nothing uses the return value of this function unless it's a PyResult<T>, which is used to throw exceptions, but not to return a value?

Yes, exactly. It's the same as Python's `__init__` in that regard.

See https://github.com/PyO3/pyo3/blob/master/pyo3-derive-backend/src/pymethod.rs#L192 where the result is matched as `Ok(_)`.

The old text probably refers to an earlier version's behavior.

konstin:
Thank you!

PR #444, opened 2019-04-17
Make Py(Frozen)Set::new return PyResult<&Py(Frozen)Set>

### LINKED ISSUE #437: Should PySet::new return a PyResult?
Currently `PySet::new` can panic easily, since not all objects are hashable. I think it would make sense to return a PyResult here.

(For example, `PyDict::from_sequence` or even `PyDict::contains`, where the same restriction applies, return results already.)

On a related note, I'm not sure why `PySet::new` returns a `Py<PySet>` while `PyDict::new` returns a `&PyDict`...

birkenfeld:
fixes #437


kngwyu:
```rust
        let v = vec![1];
        assert!(PySet::new(py, &[v]).is_err());
```
Why does this returns an error?

birkenfeld:
Lists (converted from the vector) can't be put into sets, since they are not hashable.

kngwyu:
Ah I see, thanks.

konstin:
Thank you!

Issue #392, opened 2019-03-07
Make `PyDateTime::new` take a `&PyTzInfo`


konstin:
Change the argument of `PyDateTime::new` to take a `Option<&PyTzInfo>` instead of a `Option<&PyObject>` as per discussion in https://github.com/PyO3/pyo3/pull/377#issuecomment-468279772.

pganssle:
@konstin One thing we may want to take into account would be the possible performance implications of this, I think. I'm not sure if `extract` is a zero-cost abstraction. I think right now `&PyObject` is used because it can be used with `PyTzInfo` objects created entirely in Rust *and* with `tzinfo` objects passed through Python, and it's safe-ish in the sense that, while not enforced at compile-time, it will not panic if something other than `tzinfo` is passed, it will raise a Python exception.

Unfortunately with all the procedural macros (and the fact that the overhead has not been optimized too much), it's hard for me to tell how much overhead we have on the existing functions anyway, and how much this will change it.

konstin:
Oh, that's a good point. The conversion is not zero cost, because we need to make one call to cpython's type check method. But I don't think we should care about the performance impact as we don't have benchmarks nor too many optimizations.

pganssle:
Thinking about this more, the requirement that something be a `PyTzInfo` is unnecessary and doesn't match the spirit of the C/Python API. I think a better approach might be to switch this over to:

```rust
    pub fn new<TZ: IntoPyTzInfo, 'p> (
        py: Python<'p>,
        year: i32,
        ...
        tzinfo: Option<TZ>,
    ) -> PyResult<&'p PyDateTime>
{
```

The C API enforces that whatever's there is a `tzinfo` object, but not for any structural reason - `tzinfo` is an abstract base class that defines an interface. In rust `TzInfo` is best expressed as a trait.

pganssle:
I just realized the downside of the above is that it will require an explicit reference to `PyTzInfo` or something whenever you pass `None`, so:

```rust
PyDateTime::new(2019, 1, 1, 0, 0, 0, 0, None)
```
will have to become:

```rust
PyDateTime::new::<PyTzInfo>(2019, 1, 1, 0, 0, 0, None)
```

which I don't love. Maybe the best option is to switch it over to `PyAny` and let CPython enforce the relevant conditions on the input.

konstin:
Using a trait sounds great! We might be able to solve the  `PyDateTime::new::<PyTzInfo>` with a default for the generic .

[...]

PR #446, opened 2019-04-17
Make more constructors return &PyX instead of Py<PyX>

### LINKED ISSUE #405: Safe datetime constructors should return PyErr<&T>
After some discussion in the gitter, it seems like we're converging on the standard that the `new` constructor for the safe rust bindings for Python classes should return either `&'p T` where `'p` is the lifetime of the `Python` object, or `PyErr<&'p T>`, but when I first wrote the `datetime` bindings, I had them return `PyErr<Py<T>>`.

I would like to switch those over for the 0.6.0 release (see #404).

I have a local branch where I've already done this for `PyDate`, shouldn't be too much longer converting the others.

birkenfeld:
fixes #405

**Remaining** one is PyTuple, where I'm running into a test failure. Will provide as a separate PR.

### DIFF
--- a/examples/rustapi_module/src/datetime.rs
+++ b/examples/rustapi_module/src/datetime.rs
@@ -6,7 +6,7 @@ use pyo3::types::{
 use pyo3::wrap_pyfunction;
 
 #[pyfunction]
-fn make_date(py: Python<'_>, year: i32, month: u8, day: u8) -> PyResult<Py<PyDate>> {
+fn make_date<'p>(py: Python<'p>, year: i32, month: u8, day: u8) -> PyResult<&'p PyDate> {
     PyDate::new(py, year, month, day)
 }
 
@@ -19,19 +19,19 @@ fn get_date_tuple(py: Python<'_>, d: &PyDate) -> Py<PyTuple> {
 }
 
 #[pyfunction]
-fn date_from_timestamp(py: Python<'_>, timestamp: i64) -> PyResult<Py<PyDate>> {
+fn date_from_timestamp<'p>(py: Python<'p>, timestamp: i64) -> PyResult<&'p PyDate> {
     PyDate::from_timestamp(py, timestamp)
 }
 
 #[pyfunction]
-fn make_time(
-    py: Python<'_>,
+fn make_time<'p>(
+    py: Python<'p>,
     hour: u8,
     minute: u8,
     second: u8,
     microsecond: u32,
     tzinfo: Option<&PyTzInfo>,
-) -> PyResult<Py<PyTime>> {
+) -> PyResult<&'p PyTime> {
     PyTime::new(
         py,
         hour,
@@ -44,15 +44,15 @@ fn make_time(
 
 #[cfg(Py_3_6)]
 #[pyfunction]
-fn time_with_fold(
-    py: Python,
+fn time_with_fold<'p>(
+    py: Python<'p>,
     hour: u8,
     minute: u8,
     second: u8,
     microsecond: u32,
     tzinfo: Option<&PyTzInfo>,
     fold: bool,
-) -> PyResult<Py<PyTime>> {
+) -> PyResult<&'p PyTime> {
     PyTime::new_with_fold(
         py,
         hour,
@@ -93,7 +93,12 @@ fn get_time_tuple_fold(py: Python, dt: &PyTime) -> Py<PyTuple> {
 }
 
 #[pyfunction]
-fn make_delta(py: Python<'_>, days: i32, seconds: i32, microseconds: i32) -> PyResult<Py<PyDelta>> {
+fn make_delta<'p>(
+    py: Python<'p>,
+    days: i32,
+    seconds: i32,
+    microseconds: i32,
+) -> PyResult<&'p PyDelta> {
     PyDelta::new(py, days, seconds, microseconds, true)
 }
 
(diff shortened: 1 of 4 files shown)


birkenfeld:
Missing changelog entry; I'd like to combine this with the one for PySet when both are merged.

konstin:
CC @pganssle for the datetime constructors

pganssle:
The datetime constructors broke a bunch of tests. I have a half-completed branch somewhere that fixes the tests, but IIRC it was semi-blocked on #392.

@birkenfeld If you don't want to be bothered with fixing the `datetime` changes, I can handle them later.

birkenfeld:
@pganssle I've found them now :smile: 

Do you know how to change the `utcoffset` method? Trying to make it return `PyResult<&'p PyDelta>` makes the proc-macro complain:
```
190 | #[pymethods]
    | ^^^^^^^^^^^^
    |
    = help: message: python method can not be generic: Ident { ident: "utcoffset", span: #0 bytes(4224..4233) }
```
The pyfunctions seem to be fine with it. Although, trying to import that module makes python3 segfault...

kngwyu:
> Do you know how to change the utcoffset method? Trying to make it return PyResult<&'p PyDelta> makes the proc-macro complain

Changing it to return `Py<PyDelta>` works but I think it's better to separate datetime fixes into an other PR.

birkenfeld:
Sorry, can you be more specific what you want to separate? Without the datetime changes there isn't much left here...

birkenfeld:
Also it's unclear to me *how* to get the Py<PyDelta> from the reference :smile:

kngwyu:
> Without the datetime changes there isn't much left here...

Yeah but there's many examples for datetime module and I thought it's tough stuff to fix all of them.
But spending after some time fixing this, I found it not too much difficult.

> Also it's unclear to me how to get the Py from the reference

```rust
use pyo3::AsPyPointer;
#[pyfunction]
fn make_date(py: Python<'_>, year: i32, month: u8, day: u8) -> PyResult<Py<PyDate>> {
    PyDate::new(py, year, month, day).map(|d| unsafe { Py::from_borrowed_ptr(d.as_ptr()) })
}
```
It works but maybe we need an easier way.

birkenfeld:
> ```rust
> use pyo3::AsPyPointer;
> #[pyfunction]
> fn make_date(py: Python<'_>, year: i32, month: u8, day: u8) -> PyResult<Py<PyDate>> {
>     PyDate::new(py, year, month, day).map(|d| unsafe { Py::from_borrowed_ptr(d.as_ptr()) })
> }
> ```
> 
> This works but maybe we need an easier way.

Uh oh. Should unsafe really be required here?

konstin:
I think we need a generic impl of that snippet:

https://github.com/PyO3/pyo3/blob/874d8a0835ff0d0d8239108d01bb2d1628918e42/src/types/tuple.rs#L137-L141

This would also avoid unsafe in downstream code

konstin:
This seems to work:

```rust
impl<'a, T: AsPyPointer> FromPy<&'a T> for Py<T> {
    fn from_py(tuple: &'a T, _py: Python) -> Py<T> {
        unsafe { Py::from_borrowed_ptr(tuple.as_ptr()) }
    }
}
```

Might be good idea to investigate this for soundness and make another pull request of that.

kngwyu:
@konstin 
Yeah I agree.

I noticed we can return `PyResult<&T>` for some pyfunctions in the datetime example.
E.g.,
```rust
#[pyfunction]
fn make_date(py: Python, year: i32, month: u8, day: u8) -> PyResult<&PyDate> {
    PyDate::new(py, year, month, day)
}
```
But when we have `&PyTzInfo` as an argument, it doesn't work for the lifetime reason. We have to return `Py<T>` in `datetime_from_timestamp` and `make_datetime`.
```rust
#[pyfunction]
fn datetime_from_timestamp(py: Python, ts: f64, tz: Option<&PyTzInfo>) -> PyResult<&PyDateTime> {
    PyDateTime::from_timestamp(py, ts, tz)
}
```
I think the latter part can be resolved by proc-macro hack, but not sure.

pganssle:
@kngwyu Why can't we use:

```python
fn datetime_from_timestamp(py: Python<'p>, ts: f64, tz: Option<&'p PyTzInfo>) -> PyResult<&'p PyDateTime> {
```

The `PyTzInfo` just has to live as long as the `PyDateTime` or longer, so the annotation seems correct.

kngwyu:
@pganssle 
We can't use the lifetime specifier `py` there. It's the limitation of our proc macro codes.
So the compiler can't know `python` and `tz` has the same lifetime.

kngwyu:
@pganssle 
Hoops, sorry I noticed this works now.
```rust
#[pyfunction]
fn datetime_from_timestamp<'py>(
    py: Python<'py>,
    ts: f64,
    tz: Option<&'py PyTzInfo>,
) -> PyResult<&'py PyDateTime> {
    PyDateTime::from_timestamp(py, ts, tz)
}
```

But now methods can't take lifetime specifiers so we can't compile
```rust
    fn utcoffset<'py>(&self, py: Python<'py>, _dt: &'py PyDateTime) -> PyResult<&'py PyDelta> {
        PyDelta::new(py, 0, 3600, 0, true)
    }
```

birkenfeld:
Whoops, rustfmt fixed.

konstin:
Thank you!

PR #447, opened 2019-04-17
Make PyTuple constructors return &PyTuple

birkenfeld:
This has a test failure in test_gc:
```
---- create_pointers_in_drop stdout ----
thread 'create_pointers_in_drop' panicked at 'assertion failed: `(left == right)`
  left: `2127`,
 right: `2126`', tests/test_gc.rs:123:9
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Also I'm not sure about the `FromPy` impl...

### DIFF
--- a/tests/test_gc.rs
+++ b/tests/test_gc.rs
@@ -93,8 +93,9 @@ impl Drop for ClassWithDrop {
         unsafe {
             let py = Python::assume_gil_acquired();
 
-            let _empty1 = PyTuple::empty(py);
-            let _empty2: PyObject = PyTuple::empty(py).into();
+            let _empty1: Py<PyTuple> = FromPy::from_py(PyTuple::empty(py), py);
+            let _empty2: Py<PyTuple> = FromPy::from_py(PyTuple::empty(py), py);
+            let _empty2: PyObject = _empty2.into();
             let _empty3: &PyAny = py.from_owned_ptr(ffi::PyTuple_New(0));
         }
     }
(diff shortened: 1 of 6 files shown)


konstin:
I assume the test calls `PyTuple::empty` which had returned an owned reference (`Py<PyTuble`) while `&PyTuple` isn't owned. Creating an owned reference increases the reference count by 1, which that test expects to happen. 

https://github.com/PyO3/pyo3/blob/874d8a0835ff0d0d8239108d01bb2d1628918e42/tests/test_gc.rs#L96-L98

birkenfeld:
Tried to make this equivalent to before using `from_py`, but still failing...

tests/test_gc.rs
```diff
@@ -110,7 +111,7 @@ fn create_pointers_in_drop() {
     {
         let gil = Python::acquire_gil();
         let py = gil.python();
-        let empty = PyTuple::empty(py);
+        let empty: Py<PyTuple> = FromPy::from_py(PyTuple::empty(py), py);
         ptr = empty.as_ptr();
         cnt = empty.get_refcnt() - 1;
         let inst = Py::new(py, ClassWithDrop {}).unwrap();
```
Alexander-N: After calling `drop(inst)` the refcount of empty is increased by 4. Only 2 of those references survive the end of the scope.

tests/test_gc.rs
```diff
@@ -110,7 +111,7 @@ fn create_pointers_in_drop() {
     {
         let gil = Python::acquire_gil();
         let py = gil.python();
-        let empty = PyTuple::empty(py);
+        let empty: Py<PyTuple> = FromPy::from_py(PyTuple::empty(py), py);
         ptr = empty.as_ptr();
         cnt = empty.get_refcnt() - 1;
```
Alexander-N: `PyTuple::empty(py)` increases the refcount of the empty tuple by 1 and `FromPy::from_py(PyTuple::empty(py), py)` by 2. So if `Py<PyTuple>` is used here for `empty` (not necessary) then `cnt` has to be corrected with
```rust
cnt = empty.get_refcnt() - 2;
```

kngwyu:
I fixed the tests and examples.
@birkenfeld Thanks for your PR!
@Alexander-N Thanks for your advice on the failing test!

PR #449, opened 2019-04-17
Allow omitting PyResult wrapping in property get/set methods

### LINKED ISSUE #443: Property #[getter] and #[setter] methods don't support non-PyResult return types
This is probably easy to fix, but would be good for consistency - if no PyErr can occur, accept `T` and `()` for getters and setters, respectively.

birkenfeld:
fixes #443

### DIFF
--- a/pyo3-derive-backend/src/pymethod.rs
+++ b/pyo3-derive-backend/src/pymethod.rs
@@ -273,7 +273,9 @@ pub(crate) fn impl_wrap_getter(cls: &syn::Type, name: &syn::Ident) -> TokenStrea
             let _py = pyo3::Python::assume_gil_acquired();
             let _slf = _py.mut_from_borrowed_ptr::<#cls>(_slf);
 
-            match _slf.#name() {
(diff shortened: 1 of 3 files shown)


pyo3-derive-backend/src/pymethod.rs
```diff
@@ -273,7 +273,9 @@ pub(crate) fn impl_wrap_getter(cls: &syn::Type, name: &syn::Ident) -> TokenStrea
             let _py = pyo3::Python::assume_gil_acquired();
             let _slf = _py.mut_from_borrowed_ptr::<#cls>(_slf);
 
-            match _slf.#name() {
+            let _result = pyo3::derive_utils::IntoPyResult::into_py_result(_slf.#name());
```
konstin: `_result` is guaranteed to be used, so it can just be called `result`.

konstin: (The existing code is a bad example here)

kngwyu:
Thanks :+1:

PR #450, opened 2019-04-17
Adjust the varargs/kwds objects to remove arguments consumed by parameters

birkenfeld:
There should be a *lot* more tests for argument parsing.  I think keyword-only param handling is still broken.

If the code changes here are looking good, I'll see if I can copy and adapt some tests from the CPython test suite.

### DIFF
--- a/guide/src/function.md
+++ b/guide/src/function.md
@@ -46,6 +46,34 @@ fn module_with_functions(py: Python, m: &PyModule) -> PyResult<()> {
 # fn main() {}
 ```
 
+## Argument parsing
+
+Both the `#[pyfunction]` and `#[pyfn]` attributes support specifying details of
+argument parsing.  The details are given in the section "Method arguments" in
+the [Classes](class.md) chapter.  Here is an example for a function that accepts
+arbitrary keyword arguments (`**kwargs` in Python syntax) and returns the number
+that was passed:
+
+```rust
+# extern crate pyo3;
+use pyo3::prelude::*;
+use pyo3::wrap_pyfunction;
+use pyo3::types::PyDict;
+
+#[pyfunction(kwds="**")]
+fn num_kwds(kwds: Option<&PyDict>) -> usize {
+    kwds.map_or(0, |dict| dict.len())
+}
+
+#[pymodule]
+fn module_with_functions(py: Python, m: &PyModule) -> PyResult<()> {
+    m.add_wrapped(wrap_pyfunction!(num_kwds)).unwrap();
+    Ok(())
+}
+
+# fn main() {}
+```
+
 ### Making the function signature available to Python
 
 In order to make the function signature available to Python to be retrieved via
--- a/src/derive_utils.rs
+++ b/src/derive_utils.rs
@@ -36,12 +36,12 @@ pub fn parse_fn_args<'p>(
     py: Python<'p>,
     fname: Option<&str>,
     params: &[ParamDescription],
-    args: &mut &'p PyTuple,
-    kwargs: &mut Option<&'p PyDict>,
+    args: &'p PyTuple,
+    kwargs: Option<&'p PyDict>,
     accept_args: bool,
     accept_kwargs: bool,
     output: &mut [Option<&'p PyAny>],
-) -> PyResult<()> {
+) -> PyResult<(&'p PyTuple, Option<&'p PyDict>)> {
     let nargs = args.len();
     let mut used_args = 0;
     macro_rules! raise_error {
@@ -49,6 +49,11 @@ pub fn parse_fn_args<'p>(
             concat!("{} ", $s), fname.unwrap_or("function") $(,$arg)*
         ))))
     }
+    // Copy kwargs not to modify it
+    let kwargs = match kwargs {
+        Some(k) => Some(k.copy()?),
+        None => None,
+    };
     // Iterate through the parameters and assign values to output:
     for (i, (p, out)) in params.iter().zip(output).enumerate() {
         *out = match kwargs.and_then(|d| d.get_item(p.name)) {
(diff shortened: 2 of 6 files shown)


kngwyu:
Sorry for my ignorance, but what's the wrong with the current code and what does this PR fix?

birkenfeld:
Well, a Python function that takes arguments like `a, *b` and is called with three positional arguments gets the first one as `a` and two in a tuple in `b`. Without this change, for the equivalent in Rust you would get the first argument and then all the again in the varargs tuple

The same applies to `**kwargs` and the resulting dictionary.

birkenfeld:
Also just see the linked bug, it has a nice example.

kngwyu:
Thanks for your explanation, now I think I understand.

src/derive_utils.rs
```diff
@@ -34,76 +34,88 @@ pub struct ParamDescription {
...
+            )));
+        }
+    }
+    // check for extraneous positional arguments
+    if !accept_args && used_args < nargs {
+        return Err(TypeError::py_err(format!(
+            "{} takes at most {} positional argument{} ({} given)",
+            fname.unwrap_or("function"),
+            used_args,
+            if used_args == 1 { "" } else { "s" },
```
kngwyu: I feel this too strict :confused: ...

src/derive_utils.rs
```diff
@@ -34,76 +34,88 @@ pub struct ParamDescription {
...
+            nargs
+        )));
+    }
+    // adjust the remaining args
+    if accept_args {
```
kngwyu: How about `accept_args && used_args < nargs`?
It would be slightly faster.

src/derive_utils.rs
```diff
@@ -34,76 +34,88 @@ pub struct ParamDescription {
...
-            let item = <PyTuple as PyTryFrom>::try_from(item)?;
-            let key = <PyString as PyTryFrom>::try_from(item.get_item(0))?.to_string()?;
-            if !params.iter().any(|p| p.name == key) {
-                return Err(TypeError::py_err(format!(
-                    "'{}' is an invalid keyword argument for this function",
-                    key
-                )));
-            }
+    // check for extraneous keyword arguments
+    if !accept_kwargs && !kwargs.as_ref().map_or(true, |d| d.is_empty()) {
```
kngwyu: Since `d.is_empty()` is called later again, how about defining `let kwargs_is_empty = kwargs.as_ref().map_or(false, |d| d.is_empty());`?

src/derive_utils.rs
```diff
@@ -34,76 +34,88 @@ pub struct ParamDescription {
...
+        for (key, _) in kwargs.unwrap().iter() {
```
kngwyu: Looks just `let key = kwargs.unwrap().iter().next().unwrap()` is sufficient.

guide/src/function.md
```diff
@@ -50,6 +50,34 @@ fn module_with_functions(py: Python, m: &PyModule) -> PyResult<()> {
 # fn main() {}
 ```
 
+## Argument parsing
```
kngwyu: Good document :+1:

kngwyu:
@birkenfeld 
I noticed this test
```py
@pytest.mark.parametrize(
    "args, kwargs",
    ((2017, 9, 1, 12, 45, 30, 0), {"tzinfo": UTC}),
)
def test_datetime(args, kwargs):
    act = rdt.make_datetime(*args, **kwargs)
    exp = pdt.datetime(*args, **kwargs)

    assert act == exp
    assert act.tzinfo is exp.tzinfo
```
fails because kwargs becomes `{}` after `rdt.make_datetime` is called.
So it's desirable not to delete the element of kwargs.

src/derive_utils.rs
```diff
@@ -34,77 +34,74 @@ pub struct ParamDescription {
...
+            if used_args == 1 { "" } else { "s" },
+            nargs
+        )
+    }
+    // Adjust the remaining args
+    if accept_args {
+        let slice = args
+            .slice(used_args as isize, nargs as isize)
+            .into_object(py);
+        *args = py.checked_cast_as(slice).unwrap();
```
kngwyu: Why do we have to truncate the args?

birkenfeld:
> @birkenfeld
> I noticed this test (...)
> fails because kwargs becomes `{}` after `rdt.make_datetime` is called.
> So it's desirable not to delete the element of kwargs.

I agree; this change is not supposed not affect anything in Python-land. If it does, that needs to be fixed.

src/derive_utils.rs
birkenfeld: Well, this PR assumes that the semantics on the Rust side is supposed to mimic the semantics on the Python side.

I.e. when you declare a parameter as `*args`, it only gets those positional arguments that are not matched by other parameters.

Same for keywords; the dictionary Python gives you for `**kwds` parameters does not contain keywords that are already matched elsewhere.  Python obviously creates fresh dictionaries for this purpose, e.g.

```
>>> def f(**kwds):
...  kwds.clear()
... 
>>> a = {'a': 1}
>>> f(**a)
>>> a
{'a': 1}
```

andreycizov:
Waiting for this to be merged.

Just to be clear, there's a good example of argument unwrapping in inspect.Signature._bind(): https://github.com/python/cpython/blob/f35c51d2eadd297bcf06d4f7c536bd1d8682b724/Lib/inspect.py#L2889

konstin:
@kngwyu Is there anything besides the ci that needs to be changed before merging this pull request?

kngwyu:
2 points:
- args and kwargs shouldn't be changed by this operation
- But I want to avoid copying

birkenfeld:
Both at the same time seems impossible. But copying only involves some PyObject pointers, right?

konstin:
Imho it's more important at this point to have correct and readable code than to make it fast

kngwyu:
Finally, I didn't came up with a nicer implementation than just copying kwargs

kngwyu:
Hoops clippy says `parse_fn_args` has too many arguments... :sweat:

konstin:
Thank you!

PR #451, opened 2019-04-18
Add advice for test and test-doc to Contributing.md

birkenfeld:


birkenfeld:
Yes, I had it happen. Can't say if the "wrong" one was from a previous Rust version or something else...

konstin:
Too bad. Thanks anyway!

PR #452, opened 2019-04-18
WIP: Fix code coverage

pganssle:
It seems like this is stalling out during `kcov`, so maybe we should pin to a specific release of `kcov` for now rather than using the master branch.


konstin:
I'm very much in favour of pinning a branch, but I fear that's not the cause for those build failures since the last commit to master was on 2019-03-17.

PR #453, opened 2019-04-18
Fix inaccurate comment

pganssle:
The previous commit was merged somewhat prematurely, and fixed the coverage on Travis by limiting `kcov` to a single process.

The previous commit also pins to a specific branch, which was not necessary to fix this issue, but is a good idea anyway to prevent bad commits on kcov's master branch from breaking PyO3's build.

This is cleanup for #452.

### DIFF
--- a/ci/travis/cover.sh
+++ b/ci/travis/cover.sh
@@ -2,12 +2,14 @@
 
 set -ex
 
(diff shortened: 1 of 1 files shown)


kngwyu:
Thanks!

PR #455, opened 2019-04-21
Fix some clippy lints

Alexander-N:


src/types/list.rs
```diff
@@ -20,10 +20,7 @@ pyobject_native_type!(PyList, ffi::PyList_Type, ffi::PyList_Check);
 
 impl PyList {
     /// Construct a new list with the given elements.
-    pub fn new<'p, T, U>(
-        py: Python<'p>,
-        elements: impl IntoIterator<Item = T, IntoIter = U>,
-    ) -> &'p PyList
+    pub fn new<T, U>(py: Python<'_>, elements: impl IntoIterator<Item = T, IntoIter = U>) -> &PyList
```
kngwyu: Is this a new grammar?
Great.

src/types/sequence.rs
```diff
@@ -27,6 +27,12 @@ impl PySequence {
         }
     }
 
+    #[inline]
+    pub fn is_empty(&self) -> PyResult<bool> {
+        let len = self.len()?;
```
kngwyu: I prefer to `self.len().map(|l| l == 0)`.

src/types/sequence.rs
```diff
@@ -656,4 +662,18 @@ mod test {
...
+        let py = gil.python();
+        let list = vec![1].to_object(py);
+        let seq = list.cast_as::<PySequence>(py).unwrap();
+        assert_eq!(seq.is_empty().unwrap(), false);
+        let vec: Vec<u32> = Vec::new();
+        let empty_list = vec.to_object(py);
+        let empty_seq = empty_list.cast_as::<PySequence>(py).unwrap();
+        assert_eq!(empty_seq.is_empty().unwrap(), true);
+    }
+
```
kngwyu: Just a nit but I don't like this empty line.

kngwyu:
Thanks, I lest some comments but mostly looks good.

Alexander-N:
Ok, done.

src/types/list.rs
kngwyu: Oh it was stabilized a year ago https://github.com/rust-lang/rust/pull/49458.

kngwyu:
Thanks again

src/types/list.rs
konstin: The [edition guide](https://doc.rust-lang.org/stable/edition-guide/) and the official rust blog (esp. the release notes) are really good resources for that

PR #460, opened 2019-04-23
Add PyMarshal_* functions to ffi interface.

de-vri-es:
This PR adds the `PyMarshal_*` functions [1] to the ffi interface.

This can be used to serialize/deserialize a limited set of python objects. Most interestingly for me, it can serialize code objects. That makes it possible to split compiling and running of python code between different processes, which can be very useful.

I haven't added any tests, mainly because it looks like there are no tests for the ffi modules.

[1] https://docs.python.org/3/c-api/marshal.html


pganssle:
@de-vri-es I think generally no tests are added because the ffi stuff is used by the safe rust layer, and that is tested.

I'm not sure what the policy is on expanding the ffi without a corresponding change in the safe Rust layer. I think it would be better to have a safe rust layer, but if it doesn't make sense to add one, then maybe just test the ffi layer.

kngwyu:
Though I don't against adding the marshal module, I have not idea what we can use it for :thinking:

de-vri-es:
> Python 2.7 is no longer supported, so I think the marshal.rs module can be added directly into the ffi/ folder, not the ffi3/ folder.

Moved :)

> I'm not sure what the policy is on expanding the ffi without a corresponding change in the safe Rust layer. I think it would be better to have a safe rust layer, but if it doesn't make sense to add one, then maybe just test the ffi layer.

I added a safe wrapper for `Read/WriteObjectToString` with a test that checks if a dict round-trips. I pretty much copied the interface of from the python [marshal module](https://docs.python.org/3/library/marshal.html).

The rest of the ffi functions take a `*mut FILE`. That makes it difficult to safely wrap them, and to test them. I considered not even adding those functions, but that seemed like half work. However, if tests are a rquirement, maybe we should just remove these? They don't really play well with Rust.

> Though I'm not against adding the marshal module, I have no idea what we can use it for thinking

@kngwyu: The use case I have in mind is a proc macro that compiles python code and serializes it. Then at runtime, the compiled code is deserialized and executed (currently in a PR at https://github.com/dronesforwork/inline-python/pull/5).

src/marshal.rs
```diff
@@ -0,0 +1,68 @@
...
+use std::os::raw::c_int;
+
+/// The current version of the marshal binary format.
+pub const VERSION: i32 = 4;
+
+/// Serialize an object to bytes using the Python built-in marshal module.
+///
+/// The built-in marshalling only supports a limited range of object.
+/// See the [python documentation](https://docs.python.org/3/library/marshal.html) for more details.
+pub fn dumps(py: Python, object: &PyObject, version: i32) -> PyResult<Vec<u8>> {
```
kngwyu: How about returning `PyResult<&PyBytes>`?
E.g.,
```rust
use crate::conversion::FromPyPointer;
...
FromPyPointer::from_owned_ptr_or_err(py, bytes)
```
would work.

src/marshal.rs
```diff
@@ -0,0 +1,68 @@
...
+    unsafe {
+        ffi::PyBytes_AsStringAndSize(bytes, &mut data, &mut size);
+        let data = Vec::from(std::slice::from_raw_parts(data as *const u8, size as usize));
+        ffi::Py_DecRef(bytes);
+        Ok(data)
+    }
+}
+
+/// Deserialize an object from bytes using the Python built-in marshal module.
+pub fn loads(py: Python, data: &impl AsRef<[u8]>) -> PyResult<PyObject> {
```
kngwyu: It's better to return `PyResult<&PyAny>`, since it has `downcast*` methods.
`FromPyPointer::from_owned_ptr_or_err` would work here too.

src/marshal.rs
```diff
@@ -0,0 +1,68 @@
...
+use crate::{PyErr, PyObject, PyResult, Python};
+use std::os::raw::c_int;
+
+/// The current version of the marshal binary format.
+pub const VERSION: i32 = 4;
+
+/// Serialize an object to bytes using the Python built-in marshal module.
+///
+/// The built-in marshalling only supports a limited range of object.
+/// See the [python documentation](https://docs.python.org/3/library/marshal.html) for more details.
```
kngwyu: I want a `#Example` section here.
It's difficult to guess how to specify the `version` without example.

kngwyu:
Thanks, I left some comments but your implementation mostly looks good.
`PyObject` and `&PyAny` are really confusing objects.
In most cases we prefer to `&PyAny`, since it has a lifetime same as GILGuard.
But if you need to store objects jumping over `GILGuard`s, you need to use `PyObject`.

> However, if tests are a rquirement, maybe we should just remove these? 

Currently we don't test all of FFI functions, so I don't think tests are absolutely necessary.
However, I'm for removing these functions if you don't need them.

> The use case I have in mind is a proc macro that compiles python code and serializes it.

Wow it's exciting :smiley:

de-vri-es:
Thanks for the feedback! I believe I processed the requested changes.

`FromPyPointer::from_owned_ptr_or_err` also simplified the wrapper to just a few lines. Very nice :)

As input the `dumps` function now takes a `&impl AsPyPointer` instead of a `&PyObject`. That seemed like the right bound.

src/marshal.rs
de-vri-es: Ah, that avoids the possibly pointless copy into a `Vec`. Nice :)

src/marshal.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// Serialize an object to bytes using the Python built-in marshal module.
+///
+/// The built-in marshalling only supports a limited range of object.
+/// The exact types supported depend on the version argument.
+/// The [`VERSION`] constant holds the highest version currently supported.
+///
+/// See the [python documentation](https://docs.python.org/3/library/marshal.html) for more details.
+///
+/// # Example:
```
kngwyu: :+1:

kngwyu:
Thanks for the change.
I approved this PR but since it adds a new module, it's better if @konstin would take a glance on it.

kngwyu:
@de-vri-es 
It would make CI green.

de-vri-es:
Rebased on master. Though it first had a DNS failure in the appveyor environment, but a re-trigger got it past that :)

PR #463, opened 2019-04-24
Fix PEP 517 build issues in test extensions

### LINKED ISSUE #462: CI Builds are failing due to editable install
Currently the CI builds are all red because `rustapi_module` has a `pyproject.toml`, and something is invoking `pip install -e` (see [error here](https://travis-ci.org/PyO3/pyo3/jobs/523877773#L880)).

This is related to pypa/pip#6434. I think pinning `pip` to an older version (I think `pip<19.0`) should temporarily fix the issue while we work out what to do. I'm not entirely sure why anything is doing an editable install, though.

pganssle:
I'm not sure why we had `usedevelop` on in the first place, but with `pip>19.1`, it has started hard-failing.

FIxes #462

### DIFF
--- a/examples/rustapi_module/setup.py
+++ b/examples/rustapi_module/setup.py
@@ -1,8 +1,10 @@
+import os
 import sys
 import platform
 
 from setuptools import setup
 from setuptools.command.test import test as TestCommand
+from setuptools.command.sdist import sdist as SdistCommand
 from setuptools_rust import RustExtension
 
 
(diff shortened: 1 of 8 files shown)


pganssle:
Ah, I think `usedevelop` was the only thing saving us from pypa/pip#6276. Fixing.

pganssle:
I just duplicated the `CargoModifiedSdist` code across both modules for now. If this remains the long-term solution to the problem or if we have a proliferation of these modules, we can maybe do some `sys.path` hacking to avoid the copy-pasted code.

examples/rustapi_module/setup.py
```diff
@@ -18,6 +20,41 @@ def run(self):
         raise SystemExit(errno)
 
 
+class CargoModifiedSdist(SdistCommand):
+    """Modifies Cargo to use an absolute rather than a relative path
```
kngwyu: `Cargo` is the name of the CLI tool.
Please use `Cargo.toml` here.

kngwyu:
Thanks for investigating the problem!
But now I can't understand why we need to use `usedevelop = True` or `isolated_build = True` :thinking: 
Why the tox fails to install PyO3 without one of these flags?

kngwyu:
@konstin 
Do you think seuptools-rust should have this kind of feature(rewrite a relative path to an absolute path in Cargo.toml)?

pganssle:
> Do you think seuptools-rust should have this kind of feature(rewrite a relative path to an absolute path in Cargo.toml)?

No, this is a *very* unusual situation and really only makes sense to do within the test suite of pyo3. An `sdist` is normally a deployable artifact, so deploying it with absolute paths for a specific system is dangerous. The only reason it makes sense here is that this is a test suite and so the `sdist` is only ever used as an ephemeral artifact from which to make test builds.

> Thanks for investigating the problem!
> But now I can't understand why we need to use usedevelop = True or isolated_build = True thinking
> Why the tox fails to install PyO3 without one of these flags?

First off, I'm turning *off* `usedevelop`. I'm turning *on* `isolated_build` because we have build-time dependencies (specified in pep 517),j and that's how you get `tox` to install them. I think we basically had `usedevelop` on before out of laziness, because it was doing an "editable installation", and I think that opted us out of doing some things like creating a `MANIFEST.in`.

The reason it started failing is that we have build-time dependencies (which are declared in `pyproject.toml`), and `pip` has been updated to start throwing an error if you have a `pyproject.toml` because a ` pyproject.toml` opts you in to PEP 517, and PEP 517 is incompatible with editable installs.

kngwyu:
Thanks for your explanation.
Now I think I understand the problem.

PR #464, opened 2019-04-24
Small fixes for tests

Alexander-N:
In 3a6e0a272 an assert was erroneously changed and the test was ignored.


kngwyu:
Nice catch, thanks.

PR #466, opened 2019-04-26
Make doctest dev-dependency

kngwyu:
Maybe we don't need it when `#[cfg(not(test))]`.
cc: @Alexander-N


Alexander-N:
I don't see the tests in the travis output, are they still getting executed?

kngwyu:
Oh it looks it doesn't work, thanks.

kngwyu:
Since doc-comment is a very small crate, now I think it's not so problematic to have it in our install dependencies.

PR #469, opened 2019-04-28
Remove doc-comment

konstin:
When I tried to make doc-comment a dev-dependency, I realized it actually only creates a mod with the file as doc string and can be replaced by two small macros.

CC @kngwyu @Alexander-N


kngwyu:
Good catch, thanks.
I think ideally we should make it dev-deps to let the author know the crate is used.
But it's also important to make the install time shorter for users.

Issue #100, opened 2017-12-27
Pickle Support


SerialDev:
As of right now its not possible to pickle classes created by PyO3.  
This feature would be invaluable for situations where some form of persistence would be desireable.

rth:
Would this mean implementing `__getstate__` and `__setstate__` methods (cf https://docs.python.org/3/library/pickle.html#pickling-class-instances)?

For instance, the way pickling works for,
 - [Boost.Python C++ library](https://www.boost.org/doc/libs/1_58_0/libs/python/doc/v2/pickle.html)
 - [Cython extensions](https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#controlling-pickling)

might provide some examples.

rth:
For instance, if we take the documentation example for  `MyClass`,

<details>

```rust
# use pyo3::prelude::*;
# use pyo3::PyRawObject;
#[pyclass]
struct MyClass {
   num: i32,
}

#[pymethods]
impl MyClass {

     #[new]
     fn new(obj: &PyRawObject, num: i32) {
         obj.init({
             MyClass {
                 num,
             }
         });
     }
}
```
</details>

by default we get the following error when pickling this class,
```python
        obj = MyClass()
    
>       pickle.dumps(obj)
E       TypeError: can't pickle MyClass objects
```

If we now add the `__getstate__`/ `__setstate__` methods,
```rust
    fn __getstate__(&self) -> PyResult<(i32)> {
        Ok(self.num)
    }

    fn __setstate__(&mut self, state: i32) -> PyResult<()> {
        self.num = state;
        Ok(())
    }
```
we get another exception,
```
_pickle.PicklingError: Can't pickle <class 'MyClass'>: attribute lookup MyClass on builtins failed
```
There some step I must be missing here.

althonos:
@rth : this may be related to the fact that PyO3 exposes all classes as part of the `builtins` module, because the import mechanism has not been properly implemented, so pickle tries to use `builtins.MyClass` and fails with the error you reported.

rth:
Thanks @althonos ! Opened a separate issue about it in #474

rth:
So by subclassing , to set the `__module__` correctly as suggested in https://github.com/PyO3/pyo3/issues/474#issuecomment-489521285, pickling seems to work. 

Though, I get a segfault occasionally (i.e. it does seem to be random) at exit. For instance when running a pytest session where one test checks pickling,
```
[...]

PR #477, opened 2019-05-07
Enable setting the module name of a class

### LINKED ISSUE #474: Generated classes are part of the builtin module
Related to the pickle support #100 , as mentioned by @althonos ,
> PyO3 exposes all classes as part of the builtins module, because the import mechanism has not been properly implemented, so pickle tries to use builtins.MyClass and fails with the error you reported.

This is also discussed in the following test, 

https://github.com/PyO3/pyo3/blob/9f2afbc3cc6cceca8d9bc2fdfd3236032e4b9577/tests/test_class_basics.rs#L70-L73

Once `__module__` it is set to `builtin`, it cannot be changed in an obvious way (at least I have not found one). Using [the example class](https://github.com/PyO3/pyo3/blob/master/guide/src/class.md#constructor),
```py
       cls = MyClass()
       cls.__module__ = "correct_name"
E   TypeError: can't set attributes of built-in/extension type 'MyClass'
```

it would be nice if there way some way to manually set the module name.

Also setuptools-rust does have the module name provided by the user for each `RustExtension` , but doesn't seem to be assigning it?

Alexander-N:
Didn't implement setting it automatically to avoid unexpected behavior. Pickling is possible but very awkward.

#100
Closes #474

### DIFF
--- a/src/type_object.rs
+++ b/src/type_object.rs
@@ -256,7 +256,7 @@ where
             let gil = Python::acquire_gil();
             let py = gil.python();
 
-            initialize_type::<Self>(py).unwrap_or_else(|_| {
+            initialize_type::<Self>(py, None).unwrap_or_else(|_| {
                 panic!("An error occurred while initializing class {}", Self::NAME)
             });
         }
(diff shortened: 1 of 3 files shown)


tests/test_various.rs
```diff
@@ -117,3 +118,58 @@ fn pytuple_pyclass_iter() {
...
+}
+
+fn add_module(py: Python, module: &PyModule) {
+    py.import("sys")
+        .expect("failed to import python sys module")
+        .dict()
+        .get_item("modules")
+        .expect("failed to get python modules dictionary")
+        .downcast_mut::<PyDict>()
+        .expect("failed to turn sys.modules into a PyDict")
```
kngwyu: Here we call `sys.modules` and ... can it fail?
I prefer to `unwrap` if the operation doesn't fail.

kngwyu:
Thank you for investigating the problem!
Looks it's the correct solution to me.

src/type_object.rs
```diff
@@ -315,6 +309,13 @@ where
 
     type_object.tp_base = base_type_object;
 
+    let name = match module_name {
+        Some(module_name) => format!("{}.{}", module_name, T::NAME),
+        None => T::NAME.to_string(),
+    };
+    let name = CString::new(name).expect("Module name/type name must not contain NUL byte");
```
lycantropos: *NULL ?

althonos: [NUL is valid in that context](https://en.wikipedia.org/wiki/Null_character) :wink:

PR #479, opened 2019-05-09
Guard against PyUnicode_AsUTF8AndSize returning null in PyString::as_bytes

goffrie:
Noticed that this is only a debug_assert when this case is actually possible.

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -2,7 +2,7 @@
 
 use crate::conversion::FromPyObject;
 use crate::conversion::{IntoPyObject, PyTryFrom, ToPyObject};
-use crate::err::{PyErr, PyResult};
+use crate::err::{self, PyErr, PyResult};
 use crate::exceptions;
 use crate::ffi;
 use crate::instance::PyNativeType;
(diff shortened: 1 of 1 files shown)


kngwyu:
https://github.com/python/cpython/blob/3.7/Objects/unicodeobject.c#L3825

So you mean `PyUnicode_UTF8(unicode)` can be NULL?
Hmm...

ijl:
Python unicode can contain surrogates such as "\ud800" that will cause a null pointer to be returned here. as_bytes() could return a PyResult.

kngwyu:
@ijl 
Thanks for your advice.
So @goffrie could you please change `as_bytes` to return `PyResult`?
The error case is very rare but still it can happen.

Alexander-N:
@ijl Could you provide an example how surrogates would cause problems here? I can't come up with a test case and fail to see why `PyUnicode_AsUTF8AndSize` would return NULL in this case. Maybe I'm misunderstanding the issue, but for example this works:
```rust
#[test]
fn test_as_bytes_surrogate() {
    let py = gil.python();
    let s = "���";
    let obj: PyObject = PyString::new(py, s).into();
    let py_string = <PyString as PyTryFrom>::try_from(obj.as_ref(py)).unwrap();
    assert_eq!(s.as_bytes(), py_string.as_bytes());
}
```

ijl:
@Alexander-N � is UTF-8 still, but `\ud800`, which is a UTF-16 surrogate, causes `PyUnicode_AsUTF8AndSize` to return null.

Alexander-N:
I got it now, thank you! Strange that these can be valid Python strings...

kngwyu:
Closed in favor of #639

PR #483, opened 2019-05-12
Fix example

thedrow:
Just added `extern crate pyo3;` since it was missing.

### DIFF
--- a/guide/src/get_started.md
+++ b/guide/src/get_started.md
@@ -45,6 +45,8 @@ features = ["extension-module"]
 **`src/lib.rs`**
 
 ```rust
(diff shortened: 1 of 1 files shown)


kngwyu:
I think we don't need it in Rust2018

thedrow:
> I think we don't need it in Rust2018

But we're still on nightly no?

konstin:
nightly/stable and the editions are independent. More details are in the [edition guide](https://doc.rust-lang.org/nightly/edition-guide/editions/index.html)

thedrow:
In that case, nevermind.

PR #489, opened 2019-05-15
build.rs: Handle `python` being Python 2

Alexander-N:
If `python` refers to Python 2 (as recommended in [1]) the build fails
with: `AttributeError: 'module' object has no attribute 'base_prefix'`

Since there is no `base_prefix` in Python 2, simply use `exec_prefix` if
the AttributeError is raised.

[1] https://www.python.org/dev/peps/pep-0394/


kngwyu:
Good catch, thanks!

PR #491, opened 2019-05-16
build.rs configuration as struct instead of strings

konstin:
Changes the configuration system from passing around a Vec<String> where each entry was a value for a different part of the configuration (with different types, all wrapped in strings) with a `InterpreterConfig` struct.

Also removes some outdated code.

This should either be merged after 0.7 or we should make another prerelease to ensure this doesn't break any platform.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ travis-ci = { repository = "PyO3/pyo3", branch = "master" }
 appveyor = { repository = "fafhrd91/pyo3" }
 
 [dependencies]
-libc = "0.2.53"
+libc = "0.2.54"
 spin = "0.5.0"
 num-traits = "0.2.6"
 pyo3cls = { path = "pyo3cls", version = "=0.7.0-alpha.1" }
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -413,204 +410,130 @@ fn get_interpreter_version(line: &str, implementation: &str) -> Result<PythonVer
...
-print(sysconfig.get_config_var('LDVERSION') or sysconfig.get_config_var('py_version_short'))
 try:
-    print(sys.base_prefix)
+    base_prefix = sys.base_prefix
 except AttributeError:
-    print(sys.exec_prefix)
-print(platform.python_implementation())
+    base_prefix = sys.exec_prefix
+
+print(json.dumps({
```
kngwyu: :+1:

Cargo.toml
```diff
@@ -34,6 +34,8 @@ indoc = "0.3.3"
 [build-dependencies]
 regex = "1.1.6"
 version_check = "0.1.5"
+serde = { version = "1.0.91", features = ["derive"] }
+serde_json = "1.0.39"
```
kngwyu: Do we really need the specific version of serde?

kngwyu:
Looks good refactoring for me, though I'm skeptical about specifying the minor version of serde.

Cargo.toml
konstin: Currently no, but it's only a build dependency and it avoids problems when using features from newer versions without checking `-Z minimal-versions` for compatibility

PR #499, opened 2019-05-31
Allow setting the module name for a pyclass

althonos:
Following #474, this PR adds an argument for `#[pyclass]` that lets the user specify the name of the module the class is part of. I also added the right module for all builtin types (either `builtins` or `datetime`), with a few tests.

###### Maintenance

 - Run `cargo fmt` (This is checked by travis ci) :white_check_mark: 
 - If applicable, add documentation to all new items and extend the guide. :white_check_mark: 

### DIFF
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -98,26 +98,19 @@ impl PyClassArgs {
             },
 
             "module" => match *assign.right {
-                syn::Expr::Lit(ref exp) => {
-                    match exp.lit {
-                        syn::Lit::Str(ref lit) => {
-                            self.module = Some(lit.clone());
-                        }
-                        _ => {
-                            return Err(syn::Error::new_spanned(
-                                *assign.right.clone(),
-                                "Wrong format for module",
-                            ));
-                        }
-                    }
-                },
+                syn::Expr::Lit(syn::ExprLit {
+                    lit: syn::Lit::Str(ref lit),
+                    ..
+                }) => {
+                    self.module = Some(lit.clone());
+                }
                 _ => {
                     return Err(syn::Error::new_spanned(
                         *assign.right.clone(),
                         "Wrong format for module",
                     ));
                 }
-            }
+            },
 
             _ => {
                 return Err(syn::Error::new_spanned(
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -96,7 +96,6 @@ impl PyClassArgs {
                     ));
                 }
             },
-
             "module" => match *assign.right {
                 syn::Expr::Lit(syn::ExprLit {
                     lit: syn::Lit::Str(ref lit),
@@ -111,7 +110,6 @@ impl PyClassArgs {
                     ));
                 }
             },
-
             _ => {
                 return Err(syn::Error::new_spanned(
                     *assign.left.clone(),
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -15,6 +15,7 @@ pub struct PyClassArgs {
     pub name: Option<syn::Expr>,
     pub flags: Vec<syn::Expr>,
     pub base: syn::TypePath,
+    pub module: Option<syn::LitStr>,
 }
 
 impl Parse for PyClassArgs {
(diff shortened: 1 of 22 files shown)


src/types/mod.rs
```diff
@@ -79,9 +79,9 @@ macro_rules! pyobject_native_type_named (
 
 #[macro_export]
 macro_rules! pyobject_native_type (
```
kngwyu: Since most native types has `builtin` as their module name, I suggest to make the argument `$module` optional.
E.g., you can do it by 
```rust
#[macro_export]
macro_rules! pyobject_native_type (
    ($name: ty, $typeobject: expr, $module: expr, $checkfunction: path $(,$type_param: ident)*) => {
        ...
    };
    ($name: ty, $typeobject: expr, $checkfunction: path $(,$type_param: ident)*) => {
        pyobject_native_type!{$name, $typeobject, Some("builtin"), $checkfunction $(,$type_param)*}
    };
);
```

src/types/mod.rs
```diff
@@ -93,12 +93,13 @@ macro_rules! pyobject_native_type (
 
 #[macro_export]
 macro_rules! pyobject_native_type_convert(
-    ($name: ty, $typeobject: expr, $checkfunction: path $(,$type_param: ident)*) => {
+    ($name: ty, $typeobject: expr, $module: expr, $checkfunction: path $(,$type_param: ident)*) => {
```
kngwyu: Same as above.
I want `$module`  optional.

kngwyu:
Thanks, I think it's a really meaningful enhancement for us, but left comments about the coding.
In addition, I want a test in which we check that the module name is correctly set.

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -94,6 +96,29 @@ impl PyClassArgs {
                     ));
                 }
             },
+
+            "module" => match *assign.right {
```
kngwyu: I prefer to
```rust
           "module" => match *assign.right {
                syn::Expr::Lit(syn::ExprLit {
                    lit: syn::Lit::Str(ref lit),
                    ..
                }) => {
                    self.module = Some(lit.clone());
                }
                _ => {
                    return Err(syn::Error::new_spanned(
                        *assign.right.clone(),
                        "Wrong format for module",
                    ));
                }
```

althonos:
I forgot to push the latest commits with changelog expansion and tests for the module name, will do that when I can.

src/types/mod.rs
althonos: I tried to do that at first (mostly to save me from editing all `pyobject_native_type` invocations) but it doesn't work since the resulting macro is ambiguous: since a path is also a valid expression, 
```rust
pyobject_native_type!(a, b, c, d)
```
could be
```rust
pyobject_native_type!($name = a, $typeobject = b, $module = c, $checkfunction = d)
```
or 
```rust
pyobject_native_type!($name = a, $typeobject = b, $checkfunction = c, $type_param = d)
```

kngwyu: I tried both
```rust


#[macro_export]
macro_rules! pyobject_native_type (
    ($name: ty, $typeobject: expr, $checkfunction: path $(,$type_param: ident)*) => {
        pyobject_native_type!{$name, $typeobject, Some("builtin"), $checkfunction $(,$type_param)*}
    };
    ($name: ty, $typeobject: expr, $module: expr, $checkfunction: path $(,$type_param: ident)*) => {
        ...
    };
);
```
and
```rust
    ($name: ty, $typeobject: expr, $checkfunction: path $(,$type_param: ident)*) => {
        pyobject_native_type!{$name, $typeobject, Some("builtin"), $checkfunction $(,$type_param)*}
    };
```
, and only the latter worked(macro invocation is rellay difficult in Rust... :confused:

althonos: More like not very deterministic :disappointed: 
I made `$module` optional in both, but made a missing module be set as `None` instead of `Some("builtins")`, and left `Some("builtins")` in macro invocation of current native types. Omitting the module for type `x` would name the type `x` as well while calling the macro with `Some("builtins")` would name it `builtins.x`; I have no idea what this changes for the Python interpreter since I'm not that familiar with the CPython code.

kngwyu: Hmm I think it's OK to make `Some("builtin")` default, since now this macro is used only by PyO3 and rust-numpy as far as I know.

kngwyu:
Thanks! :tada:

PR #505, opened 2019-06-05
Use existing fields and methods before calling custom __getattr__

konstin:
Previously, defining `__getattr__` would override all existing fields and methods. This changes it to behave like a `__getattr__` method defined in python, i.e. the custom method is only called if there isn't a field or method of that name.


kngwyu:
Rebased and added a test.

konstin:
Thank for adding the test!

PR #507, opened 2019-06-05
Added fastuuid as an example

thedrow:
I just completed the first version of fastuuid and released it.
Let's include it as an example.


konstin:
Thank you!

BTW I think it would be useful add a link to pypi; I didn't even realize at first that the project was published there.

PR #509, opened 2019-06-08
Ignore return code of coverage script

Alexander-N:
As suggested in https://github.com/PyO3/pyo3/issues/439#issuecomment-498853201, a failure in the coverage script should not cause the whole build to
fail.

#439


kngwyu:
Thank you!

PR #511, opened 2019-06-13
Refactor according to #[warn(bare_trait_objects)]

kngwyu:
The latest nightly compiler raises warnings for `Box<Trait>` patterns.
So I fixed them mostly by adding `dyn`, except `add_wrapped` that I modified to take `&impl Fn` as its argument.


kngwyu:
Closed because this PR is duplication of #506

PR #512, opened 2019-06-13
Expose py_run! macro

kngwyu:
Currently, we have a macro named `py_run` in tests/common.rs.
I think this macro is useful for users (e.g. when testing libraries).

When exposing this macro, I made a modification.
When string literal is passed, we use [indoc](https://github.com/dtolnay/indoc) to generate an indented string in compile time.
In other cases, we do string conversion in runtime.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -213,14 +213,13 @@ macro_rules! wrap_pymodule {
     }};
 }
 
-/// A convinient macro to execute a Python code snippet, with some local variables set.
+/// A convenient macro to execute a Python code snippet, with some local variables set.
 ///
 /// # Example
 /// ```
 /// use pyo3::{prelude::*, py_run, types::PyList};
 /// let gil = Python::acquire_gil();
 /// let py = gil.python();
-/// let obj = pyo3::types::PyDict::new(py);
 /// let list = PyList::new(py, &[1, 2, 3]);
 /// py_run!(py, list, "assert list == [1, 2, 3]");
 /// ```
@@ -229,8 +228,9 @@ macro_rules! wrap_pymodule {
 ///
 /// # Example
 /// ```
-/// use pyo3::{prelude::*, PyRawObject, py_run};
+/// use pyo3::{prelude::*, py_run};
 /// #[pyclass]
+/// #[derive(Debug)]
 /// struct Time {
 ///     hour: u32,
 ///     minute: u32,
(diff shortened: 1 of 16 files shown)


src/lib.rs
```diff
@@ -210,6 +213,96 @@ macro_rules! wrap_pymodule {
     }};
 }
 
+/// A convinient macro to execute a Python code snippet, with some local variables set.
```
birkenfeld: typo: convenient

src/lib.rs
```diff
@@ -210,6 +213,96 @@ macro_rules! wrap_pymodule {
...
+///     }
+/// }
+/// let gil = Python::acquire_gil();
+/// let py = gil.python();
+/// let time = PyRef::new(py, Time {hour: 8, minute: 43, second: 16}).unwrap();
+/// py_run!(py, time, r#"
+/// assert time.hour == 8
+/// assert time.repl_japanese() == "8時43分16秒"
+/// "#);
+/// ```
```
birkenfeld: I think there should be an example with more than one variable, or it should be explained better what the general syntax is.

src/lib.rs
```diff
@@ -210,6 +213,96 @@ macro_rules! wrap_pymodule {
...
+///     minute: u32,
+///     second: u32,
+/// }
+/// #[pymethods]
+/// impl Time {
+///     fn repl_japanese(&self) -> String {
+///         format!("{}時{}分{}秒", self.hour, self.minute, self.second)
+///     }
+///     #[getter]
+///     fn hour(&self) -> PyResult<u32> {
```
birkenfeld: (PyResult isn't necessary here)

src/lib.rs
```diff
@@ -210,6 +213,96 @@ macro_rules! wrap_pymodule {
...
+/// let obj = pyo3::types::PyDict::new(py);
+/// let list = PyList::new(py, &[1, 2, 3]);
+/// py_run!(py, list, "assert list == [1, 2, 3]");
+/// ```
+///
+/// You can use this macro to test pyfunctions or pyclasses quickly.
+///
+/// # Example
+/// ```
+/// use pyo3::{prelude::*, PyRawObject, py_run};
```
birkenfeld: why is PyRawObject imported?

src/lib.rs
```diff
@@ -210,6 +213,96 @@ macro_rules! wrap_pymodule {
...
 
+/// A convinient macro to execute a Python code snippet, with some local variables set.
+///
+/// # Example
+/// ```
+/// use pyo3::{prelude::*, py_run, types::PyList};
+/// let gil = Python::acquire_gil();
+/// let py = gil.python();
+/// let obj = pyo3::types::PyDict::new(py);
+/// let list = PyList::new(py, &[1, 2, 3]);
```
birkenfeld: nit: it's kind of inconsistent to import `PyList` but use `PyDict` fully-qualified...

src/lib.rs
```diff
@@ -210,6 +213,96 @@ macro_rules! wrap_pymodule {
...
+        $py.run($code, None, Some(d))
+            .map_err(|e| {
+                e.print($py);
+                // So when this c api function the last line called printed the error to stderr,
+                // the output is only written into a buffer which is never flushed because we
+                // panic before flushing. This is where this hack comes into place
+                $py.run("import sys; sys.stderr.flush()", None, None)
+                    .unwrap();
+            })
+            .expect($code)
```
birkenfeld: To be generally useful, I think this should not panic on exceptions, or reflect this in the name?

src/lib.rs
```diff
@@ -210,6 +213,96 @@ macro_rules! wrap_pymodule {
...
+    }};
+}
+
+/// Removes indentation from multiline strings in pyrun commands
+#[doc(hidden)]
+pub fn _indoc_runtime(commands: &str) -> String {
+    let indent;
+    if let Some(second) = commands.lines().nth(1) {
```
birkenfeld: better: `let indent = if ...`

src/lib.rs
```diff
@@ -210,6 +213,96 @@ macro_rules! wrap_pymodule {
...
+        indent = second
```
birkenfeld: `Some('\n').iter().chain(chars...)` gets rid of the additional allocation `"\n".to_string()` below.

kngwyu:
@birkenfeld 
Thanks for your review!

> To be generally useful, I think this should not panic on exceptions, or reflect this in the name?

Since this macro is mainly for testing, I think unwrapping is OK and more convenient.
In the latest commit I added a note about it in the document.

kngwyu:
And I rewrote `_indoc_runtime`(former called `indoc`) so that it behaves the same as `indoc!` macro.

src/lib.rs
konstin: We should optimize the examples for legibility, not for performance

src/lib.rs
```diff
@@ -210,6 +213,113 @@ macro_rules! wrap_pymodule {
...
+/// [Python::run] returns `Err` internally.
+/// If you need to handle failures, please use [Python::run] directly.
+///
+#[macro_export]
+macro_rules! py_run {
+    ($py:expr, $($val:ident)+, $code:literal) => {{
+        pyo3::py_run_impl!($py, $($val)+, pyo3::indoc::indoc!($code))
+    }};
+    ($py:expr, $($val:ident)+, $code:expr) => {{
+        pyo3::py_run_impl!($py, $($val)+, &pyo3::_indoc_runtime($code))
```
konstin: Could you explain why we're sometimes using indoc and sometimes don't?

kngwyu: indoc only accepts string literals so it doesn't work for the cases like
```rust
let code = generates_code();
py_run!(py, obj, code)
```

konstin: Then you can use [https://crates.io/crates/unindent](unindent), which is what indoc uses internally.

kngwyu: OK, I'll try that

Issue #514, opened 2019-06-20
Default arguments on methods causing crashes


sebpuetz:
### 🌍 Environment

 - Your rust version (`rustc --version`): nightly 1.36
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` Tested with both 0.6 and 0.7

### 💥 Reproducing

Compilation (both through pyo3-pack and `cargo build` fails with an error when defining default string args to pymethods. `bool` and `usize` args are parsed correctly.

```Rust
#[pyclass]
struct Foo{
    inner: String
}

#[pymethods]
impl Foo {
    #[args(bar="test", limit = 10)]
    fn foo(&mut self, bar: &str, limit: usize, py: Python) -> PyResult<()> {
        self.inner = bar.to_string();
        Ok(())
    }
}
```

```
error[E0425]: cannot find value `test` in this scope
  --> src/lib.rs:37:1
   |
37 | #[pymethods]
   | ^^^^^^^^^^^^ not found in this scope

error: aborting due to previous error

For more information about this error, try `rustc --explain E0425`.
error: Could not compile `finalfusion`.
```

If the default parameter doesn't match any of the actual parameters, the example compiles:

```Rust
#[pyclass]
struct Foo{
    inner: String
}

#[pymethods]
impl Foo {
    #[args(foo="test", limit = 10)]
    fn foo(&mut self, bar: &str, limit: usize, py: Python) -> PyResult<()> {
        self.inner = bar.to_string();
        Ok(())
    }
}
```

kngwyu:
Thanks for reporting!
Looks it's a bug in our proc macro code :thinking:

sebpuetz:
I was also looking into `Vec` as default argument, judging by the documentation, it should be possible to do something along these lines:
```Rust
    #[args(bar=vec![false, false, false], limit = 10)]
    fn foo(&mut self, bar: Vec<bool>, limit: usize, py: Python) -> PyResult<()> {
        Ok(())
    }
```

This crashes with the same message as above. Is this intended or should this compile if the bug in the macro is fixed?

kngwyu:
@sebpuetz 
I think `bar = "string` should be fixed but `vec!` can't be there because **proc macro backend cannot parse it**, though it should be possible to pass `'vec![]'`.

birkenfeld:
[...]

PR #515, opened 2019-06-21
Add arg=string example in the method arguments section in the guide

kngwyu:
It is not intuitive that we have to use `arg = "\"esaceped str\""` to pass &str as default arguments, though we can use other literals like integer without wrapping by str.
Related to: #514


birkenfeld:
+1

konstin:
Thanks!

Ideally, we should change the parsing code so that nested quotation marks aren't required anymore.

PR #523, opened 2019-07-07
Fix deprecation warnings on sync and mem

ijl:
```
warning: use of deprecated item 'std::sync::ONCE_INIT': the `new` function is now preferred
  --> src/gil.rs:12:28
   |
12 | static START: sync::Once = sync::ONCE_INIT;
   |                            ^^^^^^^^^^^^^^^ help: replace the use of the deprecated item: `Once::new()`
   |
   = note: #[warn(deprecated)] on by default

...

warning: use of deprecated item 'std::mem::uninitialized': use `mem::MaybeUninit` instead
   --> src/gil.rs:304:47
    |
304 |                 self.inner.push_back(unsafe { mem::uninitialized() });
    |                                               ^^^^^^^^^^^^^^^^^^

...
```

### DIFF
--- a/src/types/dict.rs
+++ b/src/types/dict.rs
@@ -10,7 +10,7 @@ use crate::AsPyPointer;
 use crate::IntoPyPointer;
 use crate::Python;
 use crate::{IntoPyObject, ToBorrowedObject, ToPyObject};
-use std::{cmp, collections, hash, mem};
+use std::{cmp, collections, hash};
 
 /// Represents a Python `dict`.
 #[repr(transparent)]
@@ -172,8 +172,8 @@ impl<'py> Iterator for PyDictIterator<'py> {
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
         unsafe {
-            let mut key: *mut ffi::PyObject = mem::uninitialized();
-            let mut value: *mut ffi::PyObject = mem::uninitialized();
+            let mut key: *mut ffi::PyObject = std::ptr::null_mut();
+            let mut value: *mut ffi::PyObject = std::ptr::null_mut();
             if ffi::PyDict_Next(self.dict.as_ptr(), &mut self.pos, &mut key, &mut value) != 0 {
                 let py = self.py;
                 Some((py.from_borrowed_ptr(key), py.from_borrowed_ptr(value)))
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -12,7 +12,7 @@ use crate::AsPyPointer;
 use crate::Python;
 use std::borrow::Cow;
 use std::os::raw::c_char;
-use std::{mem, str};
+use std::str;
 
 /// Represents a Python `string`.
 #[repr(transparent)]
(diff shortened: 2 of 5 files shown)


src/types/dict.rs
```diff
@@ -172,8 +172,8 @@ impl<'py> Iterator for PyDictIterator<'py> {
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
         unsafe {
-            let mut key: *mut ffi::PyObject = mem::uninitialized();
-            let mut value: *mut ffi::PyObject = mem::uninitialized();
+            let mut key: *mut ffi::PyObject = mem::MaybeUninit::uninit().assume_init();
```
kngwyu: I think `ptr::null_mut()` is better.

src/types/string.rs
```diff
@@ -56,7 +56,7 @@ impl PyString {
     #[inline]
     pub fn as_bytes(&self) -> &[u8] {
         unsafe {
-            let mut size: ffi::Py_ssize_t = mem::uninitialized();
+            let mut size: ffi::Py_ssize_t = mem::MaybeUninit::uninit().assume_init();
```
kngwyu: Same as above.
Just `let mut size: ffi::Py_ssize_t = 0;` would work.

kngwyu:
Thank you for the PR, but I think we have some redundant usage of `mem::uninitialized` now, so let's fix them too.

kngwyu:
And it looks we change the version of minimal nightly.

ijl:
Using a null pointer is in both cases a type error. I've just used the preferred APIs to avoid warnings during the build.

ijl:
The non-minimum build is broken due to tox:

```
 File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pluggy/manager.py", line 273, in load_setuptools_entrypoints
    message="Plugin %r could not be loaded: %s!" % (ep.name, e),
pluggy.manager.PluginValidationError: Plugin 'venv' could not be loaded: (pluggy 0.8.0 (/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages), Requirement.parse('pluggy<1,>=0.12.0'), {'tox'})!
```

kngwyu:
> Using a null pointer is in both cases a type error.

But we already have [this kind of usage of null_mut](https://github.com/PyO3/pyo3/blob/9669edbfe387a14912cd1abdb2cbcf506466ab05/src/err.rs#L153).

> The non-minimum build is broken due to tox:

Thanks, I'll check it later :thinking:

ijl:
I see, I changed all but one usage to a null ptr. The minimum rust that works is 2019-06-22, which is the first version after the stabilization of MaybeUninit that also has clippy available.

kngwyu:
Thanks for the fix.
You should also edit [this line](https://github.com/PyO3/pyo3/blob/9669edbfe387a14912cd1abdb2cbcf506466ab05/build.rs#L19) in build.rs.

And, about the error in travis, I'm still investigating.

kngwyu:
Thanks, again.

PR #529, opened 2019-07-12
Allow py: Python as an argument of getter

### LINKED ISSUE #528: `getter`s do not support specification of `py: Python` argument
Regular methods support automagical `py: Python` argument, however if these methods have the `#[getter]` attribute applied to them, this magic disappears and there is no way to obtain reference to `Python`, making getters like these impossible to write:

```rust
#[getter]
fn foo<'p>(&self, py: Python<'p>) -> &'p PyList {
    PyList::new(py, self.inner.foo.iter())
}
```

kngwyu:
Resolves #528.
This PR enables
```rust
#[pymethods]
impl Class {
    #[getter]
    fn get_as_list<'py>(&self, py: Python<'py>) -> &'py PyList {
        PyList::new(py, self.member)
    }
}
```

### DIFF
--- a/pyo3-derive-backend/src/pymethod.rs
+++ b/pyo3-derive-backend/src/pymethod.rs
@@ -9,13 +9,13 @@ pub fn gen_py_method(
     name: &syn::Ident,
     sig: &mut syn::MethodSig,
     meth_attrs: &mut Vec<syn::Attribute>,
-) -> TokenStream {
-    check_generic(name, sig);
+) -> syn::Result<TokenStream> {
+    check_generic(name, sig)?;
 
     let doc = utils::get_doc(&meth_attrs, true);
-    let spec = FnSpec::parse(name, sig, meth_attrs).unwrap();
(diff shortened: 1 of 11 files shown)


pyo3-derive-backend/src/pymethod.rs
```diff
@@ -31,28 +31,43 @@ pub fn gen_py_method(
             impl_py_method_def_static(name, doc, &impl_wrap_static(cls, name, &spec))
         }
         FnType::Getter(ref getter) => {
-            impl_py_getter_def(name, doc, getter, &impl_wrap_getter(cls, name))
+            let mut takes_py = false;
+            for arg in &spec.args {
```
konstin: Could this fail with a function definition like `foo(&self, py: Python, py: Python)`? In that case a slice pattern might be better.

kngwyu: Slice pattern works nicely here, thanks :+1:

Issue #395, opened 2019-03-13
GC integration does not work as expected


Alexander-N:
As noticed by @pganssle in #323 the garbage collection leads to objects being unexpectedly dropped. For example, the following test fails with a segfault:

```rust
#[pyclass(gc)]
struct GCIntegration {}

#[test]
fn gc_integration() {
    let gil = Python::acquire_gil();
    let py = gil.python();
    let inst = Py::new(py, GCIntegration {}).unwrap();
    py.run("import gc; gc.collect()", None, None).unwrap();
    py_run!(py, inst, "print(inst)");
}
```
It passes when `__traverse__` is implemented as nop:
```rust
#[pyproto]
impl PyGCProtocol for GCIntegration {
    fn __traverse__(&self, visit: PyVisit) -> Result<(), PyTraverseError> {
        Ok(())
    }
}
```
As I understand it, the GC should only collect objects involved in [circular references](https://docs.python.org/3/extending/newtypes_tutorial.html#supporting-cyclic-garbage-collection).

kngwyu:
> the GC should only collect objects involved in circular references.

`pyclass(gc)` set a type flag so that the Python interpreter recognize the object as the one which has a reference to other objects.
So in PyO3 it's invalid operation to use `pyclass(gc)` to an object which doesn't have a reference to other objects.
However there still can be a bug around typeobject ... :thinking: 
I'll check it.

Alexander-N:
Ok, but in this test an object which has a reference gets dropped: https://github.com/PyO3/pyo3/blob/6c587662433b9df967ee15a4a22559bfe8abaf1a/tests/test_gc.rs#L134-L177

Alexander-N:
The first example was segfaulting because of the missing `tp_traverse`, this will be resolved by https://github.com/PyO3/pyo3/pull/531. In the second example the object gets collected correctly, since the self-reference produces a cycle.

PR #530, opened 2019-07-14
Implement PyGCProtocol in test class

Alexander-N:
For the garbage collection to work the PyGCProtocol has to be
implemented. Ideally it should not even get compiled if it doesn't since
a missing tp_traverse can produce a segfault.

Issues #323 #395

### DIFF
--- a/tests/test_gc.rs
+++ b/tests/test_gc.rs
@@ -180,15 +180,19 @@ fn gc_integration() {
 #[pyclass(gc)]
 struct GCIntegration2 {}
 
+#[pyproto]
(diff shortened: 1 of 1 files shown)


kngwyu:
Thanks but... isn't this PR overlap #531?

pganssle:
@kngwyu No, the two are complementary (and in fact this one needs to be merged first). PR #531 enforces at compile time that you implement `PyGCProtocol` for anything using the `gc` flag. This PR implements `PyGCProtocol` for the objects in the test suite that are using the `gc` flag.

kngwyu:
@pganssle 
I see, thanks.

PR #531, opened 2019-07-14
Enforce GC contract at compile time

pganssle:
PyGCProtocol must be implemented for anything registered as tracked by the garbage collector. This modifies the `pyclass` macro to enforce this at compile time.

### DIFF
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -296,12 +296,15 @@ fn impl_class(
     // insert space for weak ref
     let mut has_weakref = false;
     let mut has_dict = false;
+    let mut has_gc = false;
     for f in attr.flags.iter() {
         if let syn::Expr::Path(ref epath) = f {
             if epath.path == parse_quote! {pyo3::type_object::PY_TYPE_FLAG_WEAKREF} {
                 has_weakref = true;
             } else if epath.path == parse_quote! {pyo3::type_object::PY_TYPE_FLAG_DICT} {
                 has_dict = true;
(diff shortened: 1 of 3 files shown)


kngwyu:
Good catch, :+1:  for this PR

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -365,6 +380,9 @@ fn impl_class(
         #inventory_impl
 
         #extra
+
+        #gc_impl
```
pganssle: This macro will stop evaluating at the token `;` included from `gc_impl` for some reason. Similarly it will stop evaluating at `let` if you change this to `let x = 3`. You can move either of these statements around in the `quote!` macro to kill it earlier. Not sure why this is happening.

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -321,6 +324,18 @@ fn impl_class(
         quote! { None }
     };
 
+    // Enforce at compile time that PyGCProtocol
+    let gc_impl = if has_gc {
+        quote! {
+            use pyo3::class;
+
+            fn _assert_implements_protocol<T: pyo3::class::PyGCProtocol>() {};
```
kngwyu: This `:` looks redundant

pganssle: Which one? Isn't it `fn func_name<GenericName: TraitBound>() {};`?

kngwyu: Sorry typo of `;`

kngwyu:
Thanks!

PR #532, opened 2019-07-14
Remove default implementation for PyGCProtocol

Alexander-N:
Implementations for these methods should always be provided. See #531

### DIFF
--- a/src/class/gc.rs
+++ b/src/class/gc.rs
@@ -15,13 +15,8 @@ pub struct PyTraverseError(c_int);
 /// GC support
 #[allow(unused_variables)]
 pub trait PyGCProtocol<'p>: PyTypeInfo {
-    fn __traverse__(&'p self, visit: PyVisit) -> Result<(), PyTraverseError> {
-        unimplemented!()
-    }
-
-    fn __clear__(&'p mut self) {
-        unimplemented!()
-    }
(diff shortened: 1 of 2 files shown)


src/class/gc.rs
```diff
@@ -15,13 +15,8 @@ pub struct PyTraverseError(c_int);
...
 #[allow(unused_variables)]
 pub trait PyGCProtocol<'p>: PyTypeInfo {
-    fn __traverse__(&'p self, visit: PyVisit) -> Result<(), PyTraverseError> {
-        unimplemented!()
-    }
-
-    fn __clear__(&'p mut self) {
-        unimplemented!()
-    }
+    fn __traverse__(&'p self, visit: PyVisit) -> Result<(), PyTraverseError>;
```
kngwyu: Let's rename `visit` to `_visit` and remove `#[allow(unused_variables)]`

pganssle: I don't think it needs to be renamed, since there's no implementation.

Alexander-N: I removed the `#[allow(unused_variables)]`. Clippy doesn't complain about `visit`.

src/class/gc.rs
```diff
@@ -13,15 +13,9 @@ use std::os::raw::{c_int, c_void};
...
+    fn __clear__(&'p mut self);
```
kngwyu: What is this method for?

konstin:
Thank you!

PR #533, opened 2019-07-14
Turn off codecov PR comments and remove badge from readme

konstin:
Coverage will still be collected and the coverage change should still be shown in the status checks.

See https://docs.codecov.io/docs/pull-request-comments#section-disable-comment

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+comment: off
+


pganssle:
@konstin I'm definitely +1 on removing the comment, but why remove the badge? That's usually how I determine whether a package could use some test contributions and how I navigate to the code coverage to find tests to add.

kngwyu:
What's the problem with showing comments?

konstin:
The coverage the badge is way too high. The [file list](https://codecov.io/gh/PyO3/pyo3/list/master/) e.g. doesn't include pyo3cls and pyo3-derive-backend and the extern function declarations aren't counted as lines. 

The comments trigger notifications and are distracting while we don't actually use them.

PR #534, opened 2019-07-14
Make CI fail on clippy warnings

Alexander-N:
Fix or ignore all current clippy warnings and let the CI fail if in the future if there are warnings.

### DIFF
--- a/src/class/basic.rs
+++ b/src/class/basic.rs
@@ -221,7 +221,7 @@ where
             // Behave like python's __getattr__ (as opposed to __getattribute__) and check
             // for existing fields and methods first
             let existing = ffi::PyObject_GenericGetAttr(slf, arg);
-            if existing == std::ptr::null_mut() {
+            if existing.is_null() {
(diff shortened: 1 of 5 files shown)


src/class/basic.rs
```diff
@@ -221,7 +221,7 @@ where
             // Behave like python's __getattr__ (as opposed to __getattribute__) and check
             // for existing fields and methods first
             let existing = ffi::PyObject_GenericGetAttr(slf, arg);
-            if existing == std::ptr::null_mut() {
+            if existing.is_null() {
```
kngwyu: :+1:

src/python.rs
```diff
@@ -234,6 +234,7 @@ impl<'p> Python<'p> {
     }
 
     /// Register `ffi::PyObject` pointer in release pool
+    #[allow(clippy::wrong_self_convention)]
```
konstin: I'd like to eventually rename and and move those functions, but for now that's fine.

PR #542, opened 2019-07-16
Implement Index for PyBytes

konstin:


src/types/string.rs
```diff
@@ -120,6 +126,15 @@ impl PyBytes {
     }
 }
 
+/// This is the same way [Vec] is indexed
+impl<I: SliceIndex<[u8]>> Index<I> for PyBytes {
+    type Output = I::Output;
+
+    fn index(&self, index: I) -> &Self::Output {
+        Index::index(self.as_bytes(), index)
```
kngwyu: Doesn't `self.as_bytes()[index]`work?

konstin: Thanks, fixed.

Issue #373, opened 2019-02-26
bytes should translate to &[u8]


pganssle:
Per a discussion in the gitter chat, it seems that currently the proper traits are not implemented to allow for Python's `bytes` objects to be automatically translated into `&[u8]` parameters. Right now, in order to write a function that takes `bytes`, you need to use `&PyBytes`:

```rust
#[pyfunction]
fn print_bytes(_py: Python, x: &PyBytes) {
    println!("{:?}", x.as_bytes())
}
```

If you instead use `&[u8]` directly, there's an error at compile time because the `pyo3::PyTypeInfo` trait is not implemented for `[u8]`. I think it should be possible to pass both a `bytes` and `bytearray` object into a Rust function that takes `&[u8]`.

I am using version 0.6.0-alpha.4, and with this code:

```rust
#[pyfunction]
fn print_bytes(_py: Python, x: &[u8]) {
    println!("{:?}", x.as_bytes())
}
```

I get this error message:

```
error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
  --> src/lib.rs:36:1
   |
36 | #[pyfunction]
   | ^^^^^^^^^^^^^ doesn't have a size known at compile-time
   |
   = help: the trait `std::marker::Sized` is not implemented for `[u8]`
   = note: to learn more, visit <...>
   = note: required because of the requirements on the impl of
           `pyo3::FromPyObject<'_>` for `&[u8]`

error[E0277]: the trait bound `[u8]: pyo3::PyTypeInfo` is not satisfied
  --> src/lib.rs:36:1
   |
36 | #[pyfunction]
   | ^^^^^^^^^^^^^ the trait `pyo3::PyTypeInfo` is not implemented for `[u8]`
   |
   = note: required because of the requirements on the impl
           of `pyo3::PyTryFrom<'_>` for `[u8]`
   = note: required because of the requirements on the impl
           of `pyo3::FromPyObject<'_>` for `&[u8]`

error: aborting due to 2 previous errors
```

CC @thedrow

konstin:
I think the second example was meant to contain a different snipper

pganssle:
@konstin Oops, sorry about that, fixed.

Hywan:
Would be a nice feature :-). Right now, the workaround I've is the following:

```rs
pub fn foo(bytes: &PyAny) -> PyResult<bool> {
    match <PyBytes as PyTryFrom>::try_from(bytes) {
[...]

PR #544, opened 2019-07-17
Handle null in PyIterator::from_object to fix #494

konstin:

PR #540, opened 2019-07-15
Disable segfaulty subclassing by default

konstin:
See https://github.com/PyO3/pyo3/issues/220 and https://github.com/PyO3/pyo3/issues/407 for details on those segfaults

Issue #494, opened 2019-05-26
SIGSEGV when PyIterator is given an integer


althonos:
## 🐛 Bug Reports

`PyIterator` creates a `SIGSEGV` when passed an integer.

### 🌍 Environment

 - Your operating system and version: Arch Linux 5.1.4-arch1-1-ARCH
 - Your python version:  3.7.3
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: `pacman -S python`
 - Your rust version (`rustc --version`): `rustc 1.36.0-nightly (7158ed9cb 2019-05-15)`
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` yes, yes

### 💥 Reproducing

The following program will panic:

```rust
extern crate pyo3;

use pyo3::prelude::*;
use pyo3::types::PyIterator;

fn main() {
    let gil = Python::acquire_gil();
    let py = gil.python();

    let obj = 1.into_object(py);
    assert!(PyIterator::from_object(py, &obj).is_err());
}
```

with `Cargo.toml`:
```toml
[dependencies]
pyo3 = { git = "https://github.com/PyO3/pyo3" }
```

birkenfeld:
This is because the return value from `PyObject_GetIter` in https://github.com/PyO3/pyo3/blob/master/src/types/iterator.rs#L42 is not checked for `NULL`.

I also don't understand why there is an additional `PyIter_Check` below (referring to #71)...

PR #545, opened 2019-07-18
Replace unsound PyByteArray::data with PyByteArray::to_vec

konstin:
Thanks to @ExpHP's comment in #373 pointing out the existing unsoundness of `PyByteArray::data` I've replace the function with `to_vec`, which copies the bytearray to a Vec. In the long run we likely want something like [ReadOnlyCell](https://docs.rs/cpython/0.2.1/cpython/buffer/struct.ReadOnlyCell.html) for an `as_slice`, but this at least fixes the UB for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
  * Implementing the Using the `gc` parameter for `pyclass` (e.g. `#[pyclass(gc)]`) without implementing the `class::PyGCProtocol` trait is now a compile-time error. Failing to implement this trait could lead to segfaults. [#532](https://github.com/PyO3/pyo3/pull/532)
+ * `PyByteArray::data` has been replaced with `PyDataArray::to_vec` because returning a `&[u8]` is unsound. (See [this comment](https://github.com/PyO3/pyo3/issues/373#issuecomment-512332696) for a great write-up for why that was unsound)
 
 ## [0.7.0] - 2018-05-26
 
--- a/src/types/bytearray.rs
+++ b/src/types/bytearray.rs
@@ -18,8 +18,6 @@ pyobject_native_type!(PyByteArray, ffi::PyByteArray_Type, ffi::PyByteArray_Check
 impl PyByteArray {
     /// Creates a new Python bytearray object.
     /// The byte string is initialized by copying the data from the `&[u8]`.
-    ///
-    /// Panics if out of memory.
     pub fn new<'p>(py: Python<'p>, src: &[u8]) -> &'p PyByteArray {
         let ptr = src.as_ptr() as *const c_char;
         let len = src.len() as ffi::Py_ssize_t;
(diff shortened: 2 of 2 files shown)


kngwyu:
I'm not against this PR, but to think about **what kind of safety we can provide under the current design**, we should keep in our mind that **we can do mutate opretaion to all &Py~ types**.
It means we cannot follow [the Rust's covariant rule about &'a types](https://doc.rust-lang.org/nomicon/subtyping.html).
So, to get surface safety, one possible approach we(and rust-cpython tema) can do  is **to use cell/pointer types when exposing data underlying PyObect pointer to Rust world**, which makes mutate operation unsafe.
Yeah, I think this approach is fine, but I also feel it makes our codes more complicated and not offers much benefit.
So, to sammarize, **we should consider carefully what kind of approach we should take, to make PyO3 safer**.

birkenfeld:
In the end, to be taken seriously by the Rust community, and to deserve the name of a safe wrapper, PyO3 cannot leave UB possible in safe code. This is the benefit that outweighs any drawback of more complicated code.

konstin:
For all I can tell pyo3 is in its general design sound. The known exceptions are:
 * `&mut self` and setters without `RefCell`(#342). The solution is relatively straight forward (wrap all rust objects in `RefCell`s before making them part of a PyObject), but the implementation requires changing a lot of code. A workaround would be feature gating `&mut self` and setter with a "you-can-write-unsound-code-this-way" feature. 
 * Panics are propagated through ffi boundaries (#492). The body of all `extern "C" fn` that are implemented by pyo3 needs to be wrapped in `catch_unwind`. I think `panic = "abort"` would also solve this, but `catch_unwind` is the proper solution.
 * Subclassing is segfaulty (#220, #470). I think this isn't general unsoundness but some problem with the gc. This will either get fixed by either #540 or a proper solution.
 * A missing null check (#494). Will get fixed by #544 once figure out why the this one CI job keeps failing.
 * The issue this PR fixes obviously.

I'd be happy about pull requests for those issues and I'm willing to help with questions arising when implementing those.

Given the size of pyo3's codebase and the cpython ffi, there are surely more bugs that can produce UB in the code, but I don't think that there are fundamental problems with pyo3's current design.

>  I'm not against this PR, but to think about what kind of safety we can provide under the current design, we should keep in our mind that we can do mutate operation to all &Py~ types. It means we cannot follow the Rust's covariant rule about &'a types.

Could you explain that a bit more? For my understanding what we're doing is well defined because every modification happens through a call to a c function with a raw pointer, which are both opaque to rustc.

kngwyu:
> Could you explain that a bit more?

Sorry for my poor English and explanation, ...
In Rust, `fn(&A)->&B` is generally safe(so `&~` types are covariant), but in our case, it is sometimes not safe since `&Py~` types are actually mutable(e.g., the case `A` is `PyByteArray` and `B` is `[u8]`).
I mean this fact makes it difficult to define what kind of safety we can provide.
We don't explicitly follow Rust's rule about references.
To provide safe interfaces, one realistic way is to wrap internal data(say, `&[u8]`) by `*Cell` when we expose it Rust types, since `*Cell` types are considered as invariant(≒ we can retrieve `&mut` from `&`) in Rust's manner.

> For my understanding what we're doing is well defined because every modification happens through a call to a c function with a raw pointer, which are both opaque to rustc.

Yeah, I think this is a good interpretation of what kind of *safety* we can provide.
But there still remain unsafety in that we can do mutate opretation via `py.run` without `unsafe` block.

PR #546, opened 2019-07-19
Add empty() to PySet and PyFrozenSet

thedrow:
I needed to initialize an empty set and found out it is currently missing.
This PR adds the `empty()` helper method for PySet and PyFrozenSet.

### DIFF
--- a/src/types/set.rs
+++ b/src/types/set.rs
@@ -8,6 +8,7 @@ use crate::object::PyObject;
 use crate::AsPyPointer;
 use crate::Python;
 use crate::{ToBorrowedObject, ToPyObject};
+use std::ptr;
 use std::{collections, hash};
 
 /// Represents a Python `set`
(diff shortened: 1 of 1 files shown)


thedrow:
I'm not sure why the type annotation is needed.
Any ideas?

kngwyu:
Hmm, does we really need this method?
I think we can do it by, say, 
```rust
let set = PySet::new<&PyAny>(py, &[]).unwrap();
```

thedrow:
We do if we want to be consistent with `PyList`.
I'd rather have it than not have it if possible.

kngwyu:
Anyway you can fix the compile error by 
```rust
let set = PySet::empty::<&crate::types::PyAny>(py).unwrap();
```

thedrow:
Yes but why do we need to specify the type in this case?
It isn't consistent with what we're doing with `PyList::empty()`.

kngwyu:
Ah, I see.
Then let's rewrite the method definition to
```rust
    pub fn empty<'p>(py: Python<'p>) -> PyResult<&'p PySet> {
        unsafe { py.from_owned_ptr_or_err(ffi::PyFrozenSet_New(ptr::null_mut())) }
    }
```

thedrow:
I completely missed that I copied the type argument as well.
Thanks.

When we don't set the `iterable` argument to `NULL` we don't skip this check:

https://github.com/python/cpython/blob/1f11cf9521114447b3e32e2ac88f075ffaa37555/Objects/setobject.c#L1041-L1046

konstin:
Thanks for the pull request. IMO we should add `empty()` mostly because it makes the code more readable. The performance gain for not creating a list first and running through the cpython code linked above is also useful.

thedrow:
@konstin That's even a better point :)

kngwyu:
Thanks!

PR #550, opened 2019-07-21
Attempting to fix #549

### LINKED ISSUE #549: Macro generated code for `__radd__`, `__rsub__`, etc does not compile
## 🐛 Bug Reports

### 🌍 Environment

 - Your operating system and version: OSX 10.12.6
 - Your python version: 3.7
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: homebrew, yes
 - Your rust version (`rustc --version`): rustc 1.38.0-nightly (83e4eed16 2019-07-14)
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` Yes, latest master branch does not work either.

### 💥 Reproducing

Please provide a [minimal working example](https://stackoverflow.com/help/mcve). This means both the rust code and the python.

The complete minimal example is here: https://github.com/hengchu/pyo3-bug, the critical parts are also shown here:
```
use pyo3::prelude::*;
use pyo3::PyNumberProtocol;
use pyo3::types::PyAny;

#[pyclass]
#[derive(Debug)]
struct BinaryArithmetic {}

#[pyproto]
impl PyNumberProtocol for BinaryArithmetic {
    fn __add__(lhs: &PyAny, rhs: &PyAny) -> PyResult<String> {
        Ok(format!("{:?} + {:?}", lhs, rhs))
    }

    fn __radd__(&self, other: &PyAny) -> PyResult<String> {
        Ok(format!("{:?} + {:?}", self, other))
    }
}
```

The above code produces the following compilation error:
```
error[E0603]: trait `PyNumberRAddProtocolImpl` is private
 --> src/lib.rs:9:1
  |
9 | #[pyproto]
  | ^^^^^^^^^^

error[E0119]: conflicting implementations of trait `pyo3::class::number::PyNumberRAddProtocolImpl` for type `BinaryArithmetic`:
 --> src/lib.rs:9:1
  |
9 | #[pyproto]
  | ^^^^^^^^^^
  |
  = note: conflicting implementation in crate `pyo3`:
          - impl<'p, T> pyo3::class::number::PyNumberRAddProtocolImpl for T
            where T: pyo3::class::number::PyNumberProtocol<'p>;

error: aborting due to 2 previous errors

Some errors have detailed explanations: E0119, E0603.
For more information about an error, try `rustc --explain E0119`.
error: Could not compile `pyo3-bug`.
```

hengchu:
Attempting to fix https://github.com/PyO3/pyo3/issues/549

Now code that implements `__radd__`, etc compiles, but when tested in python, they are not getting called... instead it seems some default implementations that return `NotImplemented` is executed. It would be great if maintainers could point me towards the right direction.

Thanks!

### DIFF
--- a/tests/test_arithmetics.rs
+++ b/tests/test_arithmetics.rs
@@ -205,7 +205,9 @@ fn rhs_arithmetic() {
 
     let c = Py::new(py, RhsArithmetic {}).unwrap();
     py_run!(py, c, "assert c.__radd__(1) == '1 + RA'");
-    py_run!(py, c, "assert 1 + c == '1 + RA'");
+    // TODO: commented out for now until reflected arithemtics gets fixed.
+    // see discussion here: https://github.com/PyO3/pyo3/pull/550
+    // py_run!(py, c, "assert 1 + c == '1 + RA'");
 }
 
 #[pyclass]
--- a/tests/test_arithmetics.rs
+++ b/tests/test_arithmetics.rs
@@ -1,3 +1,5 @@
+#![feature(specialization)]
+
 use pyo3::class::basic::CompareOp;
 use pyo3::class::*;
 use pyo3::prelude::*;
@@ -184,6 +186,26 @@ fn binary_arithmetic() {
     py_run!(py, c, "assert 1 ^ c == '1 ^ BA'");
     py_run!(py, c, "assert c | 1 == 'BA | 1'");
     py_run!(py, c, "assert 1 | c == '1 | BA'");
+}
+
+#[pyclass]
+struct RhsArithmetic {}
+
+#[pyproto]
+impl PyNumberProtocol for RhsArithmetic {
+    fn __radd__(&self, other: &PyAny) -> PyResult<String> {
+        Ok(format!("{:?} + RA", other))
+    }
+}
+
+#[test]
+fn rhs_arithmetic() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+
+    let c = Py::new(py, RhsArithmetic {}).unwrap();
+    py_run!(py, c, "assert c.__radd__(1) == '1 + RA'");
+    py_run!(py, c, "assert 1 + c == '1 + RA'");
 }
 
 #[pyclass]
--- a/src/class/number.rs
+++ b/src/class/number.rs
@@ -1271,7 +1271,7 @@ where
     }
 }
 
-trait PyNumberRAddProtocolImpl {
+pub trait PyNumberRAddProtocolImpl {
     fn __radd__() -> Option<PyMethodDef> {
         None
     }
@@ -1279,7 +1279,7 @@ trait PyNumberRAddProtocolImpl {
 
 impl<'p, T> PyNumberRAddProtocolImpl for T where T: PyNumberProtocol<'p> {}
 
-trait PyNumberRSubProtocolImpl {
+pub trait PyNumberRSubProtocolImpl {
     fn __rsub__() -> Option<PyMethodDef> {
         None
     }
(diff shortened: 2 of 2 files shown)


hengchu:
After some more fiddling, it seems that the simplest fix is to just make the RHS arithmetic protocol impl traits public, and require clients to use specialization. This makes the minimal failing example in #549 compile, but it seems that the `__radd__` code itself is still not executed... Would love to get some help on where to look next.

kngwyu:
Thanks for your PR, but I'm sorry now I'm not sure we should fix proc-macro code or trait definition.
Please wait for a few days...

hengchu:
Actually I reverted the changes that modify proc-macro and trait definitions. Now it just adds a `pub` for the rhs-arithmetic protocol traits. This is a much simpler fix :)

konstin:
Thank you for the great bug report and the pull request!

Please mark those traits you made public as `#[doc(hidden)]`. They are not meant to used by other crates, but only by the macros, so we should hide them from the docs. Also please add a simple test referencing the bug report. You can just adapt the snippet from the original bug report and add it to a fitting file in `tests`.

kngwyu:
@konstin 
Then why [these impls](https://github.com/PyO3/pyo3/blob/master/src/class/number.rs#L753) exist?

tests/test_arithmetics.rs
```diff
@@ -186,6 +188,26 @@ fn binary_arithmetic() {
...
+}
+
+#[test]
+fn rhs_arithmetic() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+
+    let c = Py::new(py, RhsArithmetic {}).unwrap();
+    py_run!(py, c, "assert c.__radd__(1) == '1 + RA'");
+    py_run!(py, c, "assert 1 + c == '1 + RA'");
```
hengchu: I think this patch isn't quite complete... The test case on line 207 passes while line 208 fails with this error:
```
TypeError: unsupported operand type(s) for +: 'int' and 'RhsArithmetic'
```

For some reason `1 + c` is not picking up the `__radd__` implementation, even though calling it directly works. I can do some more work on this if someone could give me some pointers on where to look next.

konstin: If I understand the cpython code correctly, there's no `__radd__` in the c api and pyo3's way of implementing `__radd__` is actually fundamentally broken: It would need to share the `tp_as_number->nb_add` slot with `__add__`, with pyo3 dispatching to either `__add__` or `__radd__` depending on whether self or other is the first argument.

Since no one reported this in the two years that `__radd__` is part of pyo3, I think the best is just to comment out that line, merge this pull request and open an issue about reflected operations being broken.

konstin:
>  Then what are these impls for?

I tried to get rid of specialization by replacing all traits with `default fn` with two trait, without realizing that this still uses specialization. So they are essentially some complex way to dispatch to either the custom implementation or the default (which becomes null)

tests/test_arithmetics.rs
hengchu: Now that I think about it, the way currently `__add__` works in `pyo3` is kind of magical: it allows `LHS` and `RHS` to be arbitrary types that may not even be the actual `Self` type of `PyNumberProtocol`'s impl. So I imagine `pyo3` must be doing some sort of smart dispatching to `__add__` already based on the lhs and rhs type information.

Comparing this to a plain python `__add__` implementation, if I understand things correctly, the python interpreter always passes lhs as `self` to `__add__` first, unless `__add__` returns `NotImplemented`, then it tries `__radd__` with the rhs as `self` to that method.

konstin:
Thank you!

PR #552, opened 2019-07-21
Poc dunder without specialization

konstin:
This is a prove of concept to how we can support dunder methods/protocols without specialization using inventory. Currently the only implemented method is `__add__`.

I don't have the time to bring this into a mergeable state or extend it to all dunder methods, so feel free to take this branch as a starting point.

### DIFF
--- a/src/class/number.rs
+++ b/src/class/number.rs
@@ -5,6 +5,7 @@
 
 use crate::callback::PyObjectCallbackConverter;
 use crate::class::basic::PyObjectProtocolImpl;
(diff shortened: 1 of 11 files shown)


src/class/number.rs
```diff
@@ -742,17 +750,17 @@ where
     }
 }
```
gilescope: Doesn't the selection of the PyNumberProtocolImpl impl depend upon specialisation? This branch doesn't compile for me on nightly 2019-11-09.

PR #554, opened 2019-07-23
Try out grcov

konstin:
grcov reported more reasonable coverage data when running locally. Also it's much more straightforward to use on travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,8 @@ install:
 
 script:
   # Options for grcov
-  - export CARGO_INCREMENTAL=0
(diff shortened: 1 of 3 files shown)


konstin:
There are some false positives, but otherwise it looks good: https://codecov.io/gh/PyO3/pyo3/pull/554/tree

konstin:
CC @pganssle as this would kinda supersede #533.

If there's consensus on switching to grcov I'll debug which of the `RUSTFLAGS` makes pypy fail.

pganssle:
I don't really have any opinions on `kcov` vs `grcov`. If it has more reasonable coverage numbers I say go for it.

I definitely think the code comment should still be turned off, so that part of #533 still seems like a good idea to me.

kngwyu:
Thanks, :+1:  for this change.
Looks travis failure is in PyPy datetime tests.
@pganssle 
Could you please check it?

PR #555, opened 2019-07-26
Run clippy on all the codebase

macisamuele:
The main objective of this PR is to ensure that:
 * clippy is executed on all the codebase (test included)
 * address reported warnings

I would like to enable more pedantic checks like [`doc_markdown`](https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown), [`use_self`](https://rust-lang.github.io/rust-clippy/master/index.html#use_self) and [`if_not_else`](https://rust-lang.github.io/rust-clippy/master/index.html#if_not_else) but they would generate a pretty high amount of changes (and so potential merge conflicts), so I'm not publishing a PR for now but limiting on having a clear environment such that restrictions can be added without having un-needed noise.

Side note: I've updated .travis.yml to reflect the min rust nightly version specified in build.rs

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -715,7 +715,7 @@ mod test {
             .unwrap()
             .call_method("array", ("f", (1.0, 1.5, 2.0, 2.5)), None)
             .unwrap();
-        let buffer = PyBuffer::get(py, array.into()).unwrap();
+        let buffer = PyBuffer::get(py, array).unwrap();
         assert_eq!(buffer.dimensions(), 1);
         assert_eq!(buffer.item_count(), 4);
         assert_eq!(buffer.format().to_str().unwrap(), "f");
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -706,6 +706,7 @@ mod test {
         assert_eq!(buffer.to_vec::<u8>(py).unwrap(), b"abcde");
     }
 
+    #[allow(clippy::float_cmp)] // The test wants to ensure that no precision was lost on the Python round-trip
     #[test]
     fn test_array_buffer() {
         let gil = Python::acquire_gil();
--- a/src/types/complex.rs
+++ b/src/types/complex.rs
@@ -281,8 +281,8 @@ mod test {
         let l = PyComplex::from_doubles(py, 3.0, 1.2);
         let r = PyComplex::from_doubles(py, 1.0, 2.6);
         let res = l / r;
-        assert_approx_eq!(res.real(), 0.7886597938144329);
-        assert_approx_eq!(res.imag(), -0.8505154639175257);
+        assert_approx_eq!(res.real(), 0.788_659_793_814_432_9);
+        assert_approx_eq!(res.imag(), -0.850_515_463_917_525_7);
     }
 
     #[cfg(not(Py_LIMITED_API))]
@@ -302,7 +302,7 @@ mod test {
         let gil = Python::acquire_gil();
         let py = gil.python();
         let val = PyComplex::from_doubles(py, 3.0, 1.2);
-        assert_approx_eq!(val.abs(), 3.2310988842807022);
+        assert_approx_eq!(val.abs(), 3.231_098_884_280_702_2);
     }
 
     #[cfg(not(Py_LIMITED_API))]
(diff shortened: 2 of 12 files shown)


rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly
```
konstin: Please remove this file; It seems useful at first, but becomes really confusing when your default toolchain isn't latest nightly any more, e.g. when the latest nightly is broken.

src/buffer.rs
```diff
@@ -708,14 +708,16 @@ mod test {
 
     #[test]
     fn test_array_buffer() {
+        use assert_approx_eq::assert_approx_eq;
```
konstin: In this case we actually want to check for exact equality to ensure that we didn't loose precision from roundtrip to python.

src/types/complex.rs
```diff
@@ -198,12 +198,14 @@ mod complex_conversion {
 
     #[test]
     fn from_complex() {
+        use assert_approx_eq::assert_approx_eq;
```
konstin: Same here.

ci/travis/setup.sh
```diff
@@ -7,7 +7,7 @@ set -e
 curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain=$TRAVIS_RUST_VERSION
 export PATH=$PATH:$HOME/.cargo/bin
 if [ "$TRAVIS_JOB_NAME" = "Minimum nightly" ]; then
-    rustup component add clippy
+    rustup component add clippy || cargo install --git https://github.com/rust-lang/rust-clippy/ --force clippy
```
konstin: The code is fine, I just wanted to note that I only pinned nightlies in .travis.yml that have clippy.

konstin: Correction: I actually think we should revert the change in .travis.yml because the master of clippy is not compatible with that rustc version.

macisamuele: @konstin  I saw the issue and you're right ... we cannot build the latest clippy available with the min rust compiler supported by the project.
The major issue here is that instally clippy on `nightly-2019-06-22` does not work anymore even if it should be present (according to [static.rust-lang.org](https://static.rust-lang.org/dist/2019-06-22/)).
An example of build failure is available [here](https://travis-ci.com/macisamuele/pyo3/jobs/220053898)).

What would be your suggestion in this case?

My current idea would be to modify travis such that `Minimum nightly` task runs only tests and a new task (ie. `Lints`) will be added to uniquely run `cargo fmt` and `cargo clippy`.
Waiting for your feedback before continuing

kngwyu:
I don't think we should check tests/examples codes...

Alexander-N:
> I don't think we should check tests/examples codes...

Why not? Shouldn't the test code also be in good shape so that it can serve as example on how to use PyO3?

kngwyu:
Yeah, but I don't like `#[allow(clippy::new_ret_no_self)]`

Alexander-N:
I see, it's definitely noisy. But as long as the constructor has the `#new` attribute, we could call it however we like. How about we use some other name, maybe `__new__`? I think it would be a good idea to also change this in the guide since users following the conventions they find there also have to deal with this issue.

konstin:
Imo the problem is that the signature that pyo3 requires is wrong; There's no reason why the `new` can't return an instance of the rust object and the proc macro calls `obj.init(instance)` as part of the generated wrapper. I've already made some changes simplifying the constructor syntax, but never got to completely  converting the syntax. I'd be for leaving the clippy annotations until that's fixed.

ci/travis/setup.sh
konstin: My suggestion is to revert the changes in `.travis.yml` and to `setup.sh`. That way ci should pass again. If that clippy version has bugs fixed in newer versions, we could bump the minimum nightly to e.g. `nightly-2019-08-01`, which also has clippy and rustfmt. (See https://rust-lang.github.io/rustup-components-history/)

macisamuele:
I'm going to rebase this PR out off the current master (to keep a cleaner history).
As from the comments looks like there is some friction around `new_ret_no_self` I'll make sure that clippy will ignore them.

macisamuele:
Test compilation issues with latest nightly version are happening on master as well ( https://travis-ci.com/macisamuele/pyo3/jobs/226035909#L789)

kngwyu:
@macisamuele 
Thank you for reporting that!
I'll check it.

kngwyu:
I found the cause https://github.com/dtolnay/proc-macro-hack/issues/40
So maybe it will be fixed in nightly-2019-08-18

konstin:
Thank you!

PR #556, opened 2019-07-27
Remove mention of Python 2

athre0z:
Since Py2 support is gone, I figured that we should remove this mention in the book.


kngwyu:
Thank you :+1:

PR #560, opened 2019-07-29
Use Once::new() instead of ONCE_INIT

kngwyu:
The latest compiler complains about this


kngwyu:
Hoops this looks already resolved

PR #561, opened 2019-07-31
Use to_string() instead of description() for exception messages

Alexander-N:
`description()` is soft-deprecated and can differ from `to_string()`.

For example for NotFound error:
* `description()`: "entity not found"
* `to_string()`: "No such file or directory (os error 2)"

Since io errors provide their error code through the Display trait there
is no need to get it through `raw_os_error()`.


kngwyu:
Good catch :+1: , thanks!

PR #562, opened 2019-08-03
Fix example in guide

Alexander-N:
Also add a test for exceptions with a custom error.


kngwyu:
Looks it's a really good example, thank you!

PR #563, opened 2019-08-04
Add dict-derive to Examples and tooling section in readme

Alexander-N:
@konstin, @kngwyu: @gperinazzo asked on [gitter](https://gitter.im/PyO3/Lobby?at=5cfd7245702b7e5e765982b4) if it's ok to use PyO3 in the name.


konstin:
@gperinazzo Sure you can use the name

gperinazzo:
Thanks for plugging my small crate here! I've been too busy but there's a few more things I want to do with it.
@konstin I think I'll keep the name as it is for now, thanks!

PR #564, opened 2019-08-04
Replace mashup with paste

kngwyu:
https://github.com/dtolnay/paste

It's much simpler and I like it.


kngwyu:
We can't build this by the toolchain older than 2019-07-12 because of https://github.com/dtolnay/paste/issues/9.
I found it was resolved by binary-search and don't know why.

kngwyu:
How about disabling clippy in the minimum nightly build?

PR #573, opened 2019-08-11
Implement IntoPyObject to PyRef

kngwyu:
## TL; DR
This PR enables
```rust
#[pyclass]
#[derive(Clone, Copy)]
struct Test {}

#[pymethods]
impl Test {
    fn test_self_reference(slf: PyRef<Self>, py: Python) -> PyRef<Self> {
        slf
    }
}
```

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -96,6 +96,12 @@ impl<'a, T: PyTypeInfo> ToPyObject for PyRef<'a, T> {
     }
 }
 
+impl<'a, T: PyTypeInfo> IntoPyObject for PyRef<'a, T> {
+    fn into_object(self, py: Python) -> PyObject {
+        self.to_object(py)
+    }
+}
+
 impl<'a, T: PyTypeInfo> Deref for PyRef<'a, T> {
     type Target = T;
     fn deref(&self) -> &T {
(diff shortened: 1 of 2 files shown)


konstin:
Thanks for adding that! IMO this should get an entry in the changelog.

RKennedy9064:
@kngwyu @konstin Quick question about this change. I wanted to test it out, so I pointed to master and am getting this error when I try to use it.

```
#[pyclass]
#[derive(Clone, Copy)]
struct Test {}

#[pymethods]
impl Test {
    fn test_self_reference(slf: PyRef<Self>, py: Python) -> PyRef<Self> {
        slf
    }
}

error[E0106]: missing lifetime specifier
  --> src\lib.rs:23:61
   |
23 |     fn test_self_reference(slf: PyRef<Self>, py: Python) -> PyRef<Self> {
   |                                                             ^^^^^^^^^^^ expected lifetime parameter
   |
   = help: this function's return type contains a borrowed value, but the signature does not say whether it is borrowed from `slf` or `py`
```

Am I missing something in how to use this new feature?

kngwyu:
@konstin 
Oh, sorry I forgot to add it.
@RKennedy9064
Sorry,  it looks we need modification of the internal code to use `Python` here.

kngwyu:
Now we replace `PyRef<Self>` with `PyRef<_>` internally.
So we have to also replace `PyRef<'_, Self>` with `PyRef<_>`, so modification would be straightforward.
But I want to wait #574 merged to avoid confclict.

PR #583, opened 2019-08-24
Remove intoPyObject

konstin:
This replaces `IntoPyObject` with `IntoPy<PyObject>`, meaning one less conversion trait.


kngwyu:
Thank you for this heavy PR.

Issue #311, opened 2018-12-16
Documentaion about when we free memory 


kngwyu:
I got [an issue](https://github.com/rust-numpy/rust-numpy/issues/90) in rust-numpy repo today.
See the below code:
```rust
fn main() {
    let v = vec![0; 100_000];
    let gil = Python::acquire_gil();
    let py = gil.python();
    loop {
        let py_arr = PyArray2::from_vec2(py, &vec![v.clone()]).unwrap();
        ();
        // &PyArray2 drops and nothing happens
    }
    // GILGruard drops and Python objects are freed
}
```
In such a case, I think most users assume that `py_arr` is freed when an iteration of the loop ends, but truly, it's freed after `GILGuard` drops.

So my suggestions are:
- We should have a document that clearly demonstrates objects are freed after `GILGuard` drops
- How about adding a 'freed-after-drop' object wrapper? But I'm not sure we can do it.

inv2004:
I suppose that  'freed-after-drop' can cause slowness. I suppose it would be fine to free mem by calling gil.clear() manually.

konstin:
I agree that we should improve the documentation, we especially need need a proper explanation of our GC interacting types in the guide. 

> How about adding a 'freed-after-drop' object wrapper? But I'm not sure we can do it.

It would be possible with a `PyOwned<T, 'py>` type, which would be very useful to have in generaö. Implementing that in sound way would a huge effort however and requires major changes to both pyo3  and rust-numpy.

`GILGuard::clear` should relatively easy to implement and would quite useful.

kngwyu:
I found 
```rust
```

andersk:
Wait a minute. Since anyone can create and drop a new `GILPool` to clear the shared `ReleasePool` at any time, from any thread, with or without the GIL held, doesn’t that defeat the entire supposed safety purpose of the `ReleasePool`?

kngwyu:
> Since anyone can create and drop a new GILPool to clear the shared ReleasePool at any time, from any thread, with or without the GIL held, doesn’t that defeat the entire supposed safety purpose of the ReleasePool?

I don't think so.
`drop` does nothing when it is used without Python objects.

andersk:
> `GILPool::drop` does nothing when it is used without Python objects.

[...]

PR #585, opened 2019-08-30
Require the GIL to be held in ReleasePool::drain

andersk:
This adds a `Python` marker to `GILPool`, to prevent the caller from misusing it to drain the `ReleasePool` and release Python objects without the GIL held.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  * Implementing the Using the `gc` parameter for `pyclass` (e.g. `#[pyclass(gc)]`) without implementing the `class::PyGCProtocol` trait is now a compile-time error. Failing to implement this trait could lead to segfaults. [#532](https://github.com/PyO3/pyo3/pull/532)
  * `PyByteArray::data` has been replaced with `PyDataArray::to_vec` because returning a `&[u8]` is unsound. (See [this comment](https://github.com/PyO3/pyo3/issues/373#issuecomment-512332696) for a great write-up for why that was unsound)
  * Replace `mashup` with `paste`.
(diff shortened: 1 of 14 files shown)


konstin:
IMHO we don't generally need this as `GILPool` is a `#[doc(hidden)]` internal api, but it makes upholding the invariants easier. Specifically, dropping the `GILPool` calls `ffi::Py_DECREF`, which requires the GIL to outlive the `GILPool`.

For my understanding misusing `GILPool::new` and `GILPool::new_no_pointers` will result in use-after-free, so IMO we should also mark those functions as unsafe. We might also want to inline and remove `impl Default for GILPool`, but that's a separate task.

andersk:
We should never expose functions that can be used to break the invariants of safe Rust without marking them `unsafe`, even in `#[doc(hidden)]`. A user who accidentally (or [intentionally](https://github.com/PyO3/pyo3/issues/311#issuecomment-449915963)!) relies on a `#[doc(hidden)]` API, or an auditor scanning for `unsafe` blocks, will receive no indication that this code might be dangerous.

If the functions can still be misused even with the GIL held, then yeah, we should mark them `unsafe` too. Either way, requiring a `Python` does make them less dangerous.

(The `impl Default` was removed because there’s no way for that to require a `Python`. As far as I can tell, it isn’t used anywhere, and was only added to satisfy a clippy rule that now no longer applies because `new` now takes an argument.)

konstin:
> (The impl Default was removed because there’s no way for that to require a Python. As far as I can tell, it isn’t used anywhere, and was only added to satisfy a clippy rule that now no longer applies because new now takes an argument.)

Oh sorry I had totally missed that you had already done that.

kngwyu:
> IMHO we don't generally need this as GILPool is a #[doc(hidden)] internal api, but it makes upholding the invariants easier. 

OK, I understand.

konstin:
Thank you!

PR #588, opened 2019-09-01
Enable slf: PyRef<'py, Self>

kngwyu:
Tweak of #573


konstin:
Thank you, that's a nice addition

PR #594, opened 2019-09-05
Replace pyo3-pack with maturin

konstin:


kngwyu:
BTW, what does *maturin* mean?

konstin:
It doesn't have any real meaning, I just like the name and thought it isn't as misleading as "pyo3-pack".

birkenfeld:
No reference to Patrick O'Brien, by chance?

PR #595, opened 2019-09-06
Move files from ffi3 into ffi

Alexander-N:
Since Python 2 is no longer supported, one folder should be enough.


kngwyu:
Thank you!

Issue #584, opened 2019-08-29
PyO3 has a build dependency on Python 2


vadixidav:
### 🌍 Environment

```Dockerfile
FROM ubuntu:18.04 as build

# Install dependencies.
RUN set -eux; \
    apt-get update \
    && apt-get install -y --no-install-recommends \
    pkg-config \
    libssl-dev \
    curl \
    ca-certificates \
    gcc \
    libc-dev \
    python3-dev \
    && rm -rf /var/lib/apt/lists/*
ENV LANG en_US.utf8

# Rust
RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain nightly-2019-08-28-x86_64-unknown-linux-gnu -y
ENV PATH "/root/.cargo/bin:$PATH"
```

### 💥 Reproducing

Add this dependency:

`pyo3 = "0.7.0"`

Now build your code. You will see the following:

```
error: failed to run custom build command for `pyo3 v0.7.0`

Caused by:
  process didn't exit successfully: `/web/target/release/build/pyo3-23c79148396499f4/build-script-build` (exit code: 1)
--- stderr
Error: "Could not find any interpreter at python, are you sure you have python installed on your PATH?"
```

### Workaround

I was able to fix this temporarily by adding the following to my Dockerfile:

```Dockerfile
# Force python to be python3.
RUN rm -f /usr/bin/python && ln -s /usr/bin/python3 /usr/bin/python
```

I notice that on the main `README.md` it is stated to run the following on Ubuntu:

```bash
sudo apt install python3-dev python-dev
```

However, I would prefer to only install `python3-dev` to minimize downloads, build time, etc for Docker. I feel like if just linking python to python3 works, it should probably also support discovering python3 and running python3 on the system.

Basically, another way to word this ticket is "Remove build dependency on Python 2".

kngwyu:
Closed via #596

PR #596, opened 2019-09-07
Python3 bin

expobrain:
Explicitly use the `python3` binary instead of lain `python` to target Python 3.x even on systems where Python 2.x is not installed

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -18,6 +18,7 @@ use version_check::{Channel, Date, Version};
 /// But note that this is the rustc version which can be lower than the nightly version
 const MIN_DATE: &'static str = "2019-07-18";
 const MIN_VERSION: &'static str = "1.37.0-nightly";
+const PYTHON_INTERPRETER: &'static str = "python3";
 
 /// Information returned from python interpreter
 #[derive(Deserialize, Debug)]
@@ -357,7 +358,7 @@ elif sysconfig.get_config_var("Py_ENABLE_SHARED"):
 else:
     print("static")
 "#;
-    let out = run_python_script("python", script).unwrap();
+    let out = run_python_script(PYTHON_INTERPRETER, script).unwrap();
     Ok(out.trim_end().to_owned())
 }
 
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -425,22 +426,22 @@ fn find_interpreter_and_get_config() -> Result<(InterpreterConfig, HashMap<Strin
     };
 
     // check default python
-    let interpreter_path = "python";
+    // let interpreter_path = "python3";
```
kngwyu: Could you please remove this line?

kngwyu:
It's a really nice refactoring :+1: , thank you!
And maybe this resolves #584?

build.rs
expobrain: done

expobrain:
@kngwyu yes, it solves exactly that issue

kngwyu:
Ah, sorry, now I noticed the Windows CI fails because of this fix.
Wait for a minute...

expobrain:
Interesting, I'm wondering if make sense to check in order for both `python3` and `python` binaries and use the first found

build.rs
```diff
@@ -18,6 +21,18 @@ use version_check::{Channel, Date, Version};
 /// But note that this is the rustc version which can be lower than the nightly version
 const MIN_DATE: &'static str = "2019-07-18";
 const MIN_VERSION: &'static str = "1.37.0-nightly";
+//const PYTHON_INTERPRETER: &'static str = "python3";
+
+lazy_static! {
+    static ref PYTHON_INTERPRETER: &'static str = {
+        vec!["python3", "python"]
```
kngwyu:
```suggestion
        ["python3", "python"]
            .iter()
            .find_map(|bin| Command::new(bin).spawn().ok().map(|_| bin))
            .expect("Python 3.x interpreter not found")
```

kngwyu:
```suggestion
         ["python3", "python"]
            .iter()
            .find_map(|bin| Command::new(bin).spawn().ok().map(|_| bin))
            .expect("Python 3.x interpreter not found")
```

kngwyu:
Thanks!
I left a comment about coding, but other than that LGTM.

kngwyu:
Thank you!

vadixidav:
Thank you for the fix 👍. Just saw that the issue was closed.

Issue #602, opened 2019-09-14
[More documentation] Loading and using python functions


efugier:
Can someone help me put together a full minimal example ?

I think it would be great to include it in the Readme in the "using python from rust" section which clearly lacks content.

It would create the basis for an app with the logic written in rust which rely on a few python functions. An example use case could be a machine learning (where almost everything is done with python libs) based app with some heavy post treatment that would benefit from being done in rust.

Here's my failed attempt :

`main.rs`
```rust
use std::include_str;
use pyo3::prelude::*;

fn main() -> PyResult<()> {
    let python_functions = include_str!("python-functions.py");

    let gil = Python::acquire_gil();
    let py = gil.python();

    py.eval(python_functions, None, None)?;

    let code = "test()";
    let res : String = py.eval(code, None, None)?.extract()?;


    println!("{}", res);

    Ok(())
}
```
`python_functions.py`
```python
def test():
    return "hello"
```

it gives me `Error: PyErr { type: Py(0x7ff14324ed60, PhantomData) }` which isn't really a workable error.

The next step is to make a **function that takes parameters** and we'll be set !

kngwyu:
I think rust-numpy crate has [a better example](https://github.com/rust-numpy/rust-numpy#execute-a-python-program-from-rust-and-get-results), but yeah, we need good examples in this area.

efugier:
Thanks for pointing this out, it explains how to correctly output errors and that's useful.

However I think something is still off: when the code in the script relies on multiple lines it fails with a syntax error even though there is none (the script is correct and runs with python).

kngwyu:
`eval` only accepts Python's expression(≒ one liner).
So if you want to run long code, you can use [`py_run`](https://pyo3.rs/master/doc/pyo3/macro.py_run.html) or so but it returns `None`.

efugier:
That's exactly the type of thing that the documentation is lacking, explaining basic high level stuff on what is used to do what. Thanks !

I'll make a PR as soon as I get some free time.

kngwyu:
Also, I wrote an example to get an Python object using `Python::run`.
```rust
    use pyo3::types::{IntoPyDict, PyDict, PyList};
    let gil = Python::acquire_gil();
    let py = gil.python();
    let ret_dict = [("ret", 3)].into_py_dict(py);
    py.run(
        r#"
l = [8, 7, 3, 4, 5]
l.sort()
ret = l
"#,
        None,
        Some(ret_dict),
    );
[...]

PR #603, opened 2019-09-15
Update README and document

kngwyu:
Inspired by #602.
- Update README example
- Add an example for `Python::run`

## Updated(9/23)
- Add a section named 'Call Python from Rust` in the guide.
- Add an example usage of `PyModule::from_code`.


README.md
```diff
@@ -113,6 +121,10 @@ fn main() -> PyResult<()> {
 }
 ```
 
+If you want to excute a code with multiple lines, you can use
```
birkenfeld: I don't think "multiple lines" is quite accurate here: the difference is between executing a *single expression* (and getting the return value), and executing *one or more statements*.

For example, although it's a single line, you can `eval("assert True")`.

src/python.rs
```diff
@@ -103,6 +103,26 @@ impl<'p> Python<'p> {
     ///
     /// If `globals` is `None`, it defaults to Python module `__main__`.
     /// If `locals` is `None`, it defaults to the value of `globals`.
+    ///
+    /// # Example:
+    /// ```
+    /// use pyo3::{types::{IntoPyDict, PyBytes, PyDict}, prelude::*};
+    /// let gil = pyo3::Python::acquire_gil();
+    /// let py = gil.python();
+    /// let locals = [("ret", py.None())].into_py_dict(py);
```
birkenfeld: it's probably not needed to initialize "ret" -- the code below would fail anyway on exception since None is not a PyBytes.

src/python.rs
```diff
@@ -103,6 +103,26 @@ impl<'p> Python<'p> {
...
+    /// py.run(
+    ///     r#"
+    /// import base64
+    /// s = 'Hello Rust!'
+    /// ret = base64.b64encode(s.encode('utf-8'))
+    /// "#,
+    ///    None,
+    ///    Some(locals),
+    /// );
```
birkenfeld: The Result from `py.run` should be handled.

README.md
kngwyu: Thanks.

src/python.rs
kngwyu: :+1:

PR #608, opened 2019-09-28
Support conversion between num-bigint <-> Python Long

### LINKED ISSUE #543: python3 int <-> num_bigint support
See also #124

kngwyu:
Resolves #543.

### DIFF
--- a/src/types/num.rs
+++ b/src/types/num.rs
@@ -93,7 +93,7 @@ macro_rules! int_convert_128 {
                     }
                     let buffer: [c_uchar; $byte_size] = [0; $byte_size];
                     let ok = ffi::_PyLong_AsByteArray(
-                        ob.as_ptr() as *mut ffi::PyLongObject,
+                        num as *mut ffi::PyLongObject,
                         buffer.as_ptr() as *const c_uchar,
                         $byte_size,
                         IS_LITTLE_ENDIAN,
@@ -232,13 +232,13 @@ int_convert_u64_or_i64!(
     ffi::PyLong_AsUnsignedLongLong
 );
 
-#[cfg(all(not(Py_LIMITED_API), target_endian = "little"))]
+#[cfg(not(Py_LIMITED_API))]
 int_convert_128!(i128, 16, 1);
 #[cfg(not(Py_LIMITED_API))]
 int_convert_128!(u128, 16, 0);
 
 #[cfg(all(feature = "num-bigint", not(Py_LIMITED_API)))]
-mod bitint_conversion {
+mod bigint_conversion {
     use super::*;
     use num_bigint::{BigInt, BigUint};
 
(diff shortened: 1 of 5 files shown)


src/types/num.rs
```diff
@@ -242,10 +232,175 @@ int_convert_u64_or_i64!(
...
+#[cfg(all(not(Py_LIMITED_API), target_endian = "little"))]
+int_convert_128!(i128, 16, 1);
 #[cfg(not(Py_LIMITED_API))]
-int_convert_bignum!(i128, 16, IS_LITTLE_ENDIAN, 1);
-#[cfg(not(Py_LIMITED_API))]
-int_convert_bignum!(u128, 16, IS_LITTLE_ENDIAN, 0);
+int_convert_128!(u128, 16, 0);
+
+#[cfg(all(feature = "num-bigint", not(Py_LIMITED_API)))]
+mod bitint_conversion {
```
birkenfeld: typo?

src/types/num.rs
```diff
@@ -242,10 +232,175 @@ int_convert_u64_or_i64!(
...
+                }
+            }
+            impl<'source> FromPyObject<'source> for $rust_ty {
+                fn extract(ob: &'source PyAny) -> PyResult<$rust_ty> {
+                    unsafe {
+                        let num = ffi::PyNumber_Index(ob.as_ptr());
+                        if num.is_null() {
+                            return Err(PyErr::fetch(ob.py()));
+                        }
+                        let n_bytes = (ffi::_PyLong_NumBits(ob.as_ptr()) as usize - 1) / 8 + 1;
```
birkenfeld: looks like this should use `num`, not `ob`...

Can `_PyLong_NumBits` return zero? The subtract would wrap in that case.

src/types/num.rs
```diff
@@ -242,10 +232,175 @@ int_convert_u64_or_i64!(
     ffi::PyLong_AsUnsignedLongLong
 );
 
+#[cfg(all(not(Py_LIMITED_API), target_endian = "little"))]
+int_convert_128!(i128, 16, 1);
 #[cfg(not(Py_LIMITED_API))]
-int_convert_bignum!(i128, 16, IS_LITTLE_ENDIAN, 1);
-#[cfg(not(Py_LIMITED_API))]
-int_convert_bignum!(u128, 16, IS_LITTLE_ENDIAN, 0);
+int_convert_128!(u128, 16, 0);
```
programmerjake: `i128` conversions on big-endian appears to have been inadvertently deleted

kngwyu: No, it's at line 236

src/types/num.rs
kngwyu: Thanks.

src/types/num.rs
kngwyu: > looks like this should use num, not ob...

Wow, I didn't know that.

> Can _PyLong_NumBits return zero? The subtract would wrap in that case.

Good catch, thanks.

src/types/num.rs
programmerjake: line 236 is only enabled on little-endian systems, so it can't provide the `i128` conversions on big-endian systems since it is disabled.

kngwyu: Ah, I understand.
Thank you for pointing it out!

PR #598, opened 2019-09-07
Fix memory leak of PyBytes -> Vec conversion

kngwyu:
Resolves #570.
For now, I only added a test.

Issue #570, opened 2019-08-09
Memory leak on extracting Vec<u8>


glasah:
Extracting a Vec of u8s from pyobject leaks memory.

### 🌍 Environment

 - Your operating system and version:

Linux 3bda9a19dfcb 4.15.0-54-generic #58-Ubuntu SMP Mon Jun 24 10:55:24 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

 - Your python version:

Python 3.6.8

 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?:

System install inside docker container, installed via apt

 - Your rust version (`rustc --version`):

rustc 1.37.0-nightly (5eeb567a2 2019-06-06)

 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` 

pyo3 = { version = "0.7.0", features = ["extension-module"] }

### 💥 Reproducing
```rust
use pyo3::prelude::*;

#[pyclass]
struct MyClass {}

#[pymethods]
impl MyClass {
    #[new]
    pub fn __new__(obj: &PyRawObject, _py: Python) {
        obj.init({ MyClass {} });
    }

    pub fn leaks(&mut self, py: Python, py_byte_string: PyObject) -> PyResult<()> {
        let byte_vec: Vec<u8> = py_byte_string.extract(py).unwrap();
        println!("{:?}", byte_vec.len());
        Ok(())
    }

    pub fn doesnt_leak(&mut self, py: Python, py_byte_string: PyObject) -> PyResult<()> {
        let byte_vec: String = py_byte_string.extract(py).unwrap();
        println!("{:?}", byte_vec.len());
        Ok(())
    }
}

#[pymodule]
pub fn my_module(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<MyClass>()?;
    Ok(())
}
```

Running in python 
```python
import os
import psutil
import pyo3_memory_leak

process = psutil.Process(os.getpid())
initial_mem = process.memory_info().rss

my_instance = pyo3_memory_leak.my_module.MyClass()

for i in range(10**6):
    my_instance.doesnt_leak("Hello world")

middle_mem_1 = process.memory_info().rss

for i in range(10**6):
    my_instance.doesnt_leak("Hello world")

middle_mem_2 = process.memory_info().rss

for i in range(10**6):
    my_instance.leaks(b"Hello world")

[...]

PR #612, opened 2019-10-06
Add assertions around memory_diff in test_pybuffer_doesnot_leak_memory

macisamuele:
The goal of this CR is to make `test_pybuffer_doesnot_leak_memory` more accurate.
The test was introduced in #598.

As the test was written it would fail only for "logic" issues on the rust side implementation instead of failing in case of memory leaks (as the test name would suggest).

This is caused by the missing `assert` statements on the tests check, which are going to be to be introduced by this CR.

### DIFF
--- a/examples/rustapi_module/src/buf_and_str.rs
+++ b/examples/rustapi_module/src/buf_and_str.rs
@@ -13,14 +13,14 @@ impl BytesExtractor {
         obj.init({ BytesExtractor {} });
     }
 
-    pub fn to_vec(&mut self, bytes: &PyBytes) -> PyResult<usize> {
+    pub fn from_bytes(&mut self, bytes: &PyBytes) -> PyResult<usize> {
         let byte_vec: Vec<u8> = bytes.extract().unwrap();
         Ok(byte_vec.len())
     }
 
-    pub fn to_str(&mut self, bytes: &PyString) -> PyResult<usize> {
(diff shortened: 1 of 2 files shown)


macisamuele:
NOTE: The test is currently failing, so probably the memory leak issue (#570) was never fixed.

kngwyu:
Thank you, it was my mistake that it lacks `assert` :sweat:

macisamuele:
@kngwyu that's not a major problem, it happens from time to time.
I'm trying to get this test green but I don't think that we can really achieve the result by using this logic due to how garbage collection works on Python (as far as I'm aware you cannot really force the collection and removal of all the objects, but rather hinting the interpreter to do it)

kngwyu:
Hmm, I'm not sure the memory leak is really fixed :thinking: 
Anyway, thank you for fixing the test!

macisamuele:
Before reaching the conclusion with the test in this PR I've tried to interact with the binded functions a variable amount of times (from 1k to 1M times, 1k as step) and the rss was bounded to frew KB most probably due to additional objects creation during the test. To confirm that I've tried to revert the pyo3 changes in #598 and RSS usage was growing with the increase of the number of binded function interaction (so confirming that the memory leak was solved).

The last thought here is that maybe this approach of testing for the memory leak is not ideal, but I would put more effort if you think that it might be worth

Thanks a lot for taking the change in

Issue #619, opened 2019-10-11
WindowsAppStore initially set dummy interpreter


namuyan:
## 🐛 Bug Reports
If users install Anaconda3 or normal python from official site, cannot find correct **python interpreter**. These days Windows10 updated, WindowsAppStore initially make `python.exe` and `python3.exe` but those EXE open shop for recommend install!

We must install Python from Microsoft app store?
caused: https://github.com/PyO3/pyo3/blob/master/build.rs#L28
path: `C:\Users\name\AppData\Local\Microsoft\WindowsApps\python.exe`

### 🌍 Environment

 - Your operating system and version: Windows10 64bit
 - Your python version: python3.6
 - How did you install python?: Anaconda3
 - Your rust version (`rustc --version`): 1.40.0-nightly (898f36c83 2019-10-10)
 - Are you using the latest pyo3 version? 0.8.1 stable

### Solutions
1. dummy exe don't show version by `python --version`. use it?
2. manually setup python path?
good ways?

kngwyu:
Wow :fearful:

namuyan:
merged, reopen if problem occurred.

PR #620, opened 2019-10-12
check python interpreter by --version option

namuyan:
https://github.com/PyO3/pyo3/issues/619
solved on my condition (Windows10 64bit, Python 3.6.7 :: Anaconda custom (64-bit))

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -27,12 +27,14 @@ lazy_static! {
     static ref PYTHON_INTERPRETER: &'static str = {
         ["python", "python3"]
             .iter()
-            .map(|bin| (bin, Command::new(bin).arg("--version").output()))
-            .filter(|(_, r)| r.is_ok())
-            .map(|(bin, r)| (bin, r.unwrap()))
-            // begin with `Python 3.X.X :: additional info`
-            .find(|(_, r)| r.stdout.starts_with(b"Python 3") || r.stderr.starts_with(b"Python 3"))
-            .map(|(bin, _)| bin)
+            .find(|bin| {
+                if let Ok(out) = Command::new(bin).arg("--version").output() {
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -25,10 +25,13 @@ const MIN_VERSION: &'static str = "1.37.0-nightly";
 
 lazy_static! {
     static ref PYTHON_INTERPRETER: &'static str = {
-        ["python3", "python"]
+        ["python", "python3"]
             .iter()
-            .map(|bin| (bin, Command::new(bin).spawn()))
-            .find(|(_, r)| r.is_ok())
+            .map(|bin| (bin, Command::new(bin).arg("--version").output()))
```
kngwyu: How about this:
```rust
            .find(|bin| {
                if let Ok(out) = Command::new(bin).arg("--version").output() {
                    out.stdout.starts_with(b"Python 3")
                } else {
                    false
                }
            })
            .expect("Python 3.x interpreter not found")
```

kngwyu:
Thank you for fixing this!
I left a comment about coding, but if you'd prefer the current code, I merge this PR as is.

build.rs
namuyan: Thank you for comment!
Travis show me one curious things, Windows Anaconda3 Python output version by stderr. Python provided by official output version by stdout, only Anaconda3.
So It looks Anaconda3 bug. No one noticed.

kngwyu: Ah, OK, thank you for investigating that.
My intention is just to shorten codes, though.

namuyan: No problem, I feel method chain is not readable code.
I changed to something like your suggestion.

kngwyu:
Thanks!

Issue #618, opened 2019-10-11
missing FromPy<BigInt> impl for PyObject


programmerjake:
Fails with following message:

```
error[E0277]: the trait bound `pyo3::object::PyObject: pyo3::conversion::FromPy<num_bigint::bigint::BigInt>` is not satisfied
 --> src/lib.rs:8:1
  |
8 | #[pymethods]
  | ^^^^^^^^^^^^ the trait `pyo3::conversion::FromPy<num_bigint::bigint::BigInt>` is not implemented for `pyo3::object::PyObject`
  |
  = help: the following implementations were found:
            <pyo3::object::PyObject as pyo3::conversion::FromPy<&'a T>>
            <pyo3::object::PyObject as pyo3::conversion::FromPy<()>>
            <pyo3::object::PyObject as pyo3::conversion::FromPy<bool>>
            <pyo3::object::PyObject as pyo3::conversion::FromPy<f32>>
          and 3 others
  = note: required because of the requirements on the impl of `pyo3::conversion::IntoPy<pyo3::object::PyObject>` for `num_bigint::bigint::BigInt`
  = note: required because of the requirements on the impl of `pyo3::derive_utils::IntoPyResult<num_bigint::bigint::BigInt>` for `num_bigint::bigint::BigInt`
  = note: required by `pyo3::derive_utils::IntoPyResult::into_py_result`

error: aborting due to previous error
```

### 🌍 Environment

 - Your operating system and version:
Ubuntu 18.04
 - Your python version:
cpython 3.6.8
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?:
apt + virtualenv
 - Your rust version (`rustc --version`):
rustc 1.40.0-nightly (032a53a06 2019-10-03)
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` 
Using commit 89b7c7e3

### 💥 Reproducing

run cargo build with following files:

`Cargo.toml`:
```toml
[package]
name = "pyo3-bigint"
version = "0.1.0"
edition = "2018"
publish = false

[lib]
name = "pyo3_bigint"
crate-type = ["rlib", "cdylib"]

[dependencies]
num-bigint = "0.2"
pyo3 = { git = "https://github.com/PyO3/pyo3", features = ["extension-module", "num-bigint"] }
```

`src/lib.rs`:
```rust
use pyo3::prelude::*;
[...]

PR #622, opened 2019-10-12
Implement IntoPy<PyObject> for BigInts

kngwyu:
For #618

### DIFF
--- a/src/types/num.rs
+++ b/src/types/num.rs
@@ -290,6 +290,11 @@ mod bigint_conversion {
                     }
                 }
             }
+            impl IntoPy<PyObject> for $rust_ty {
+                fn into_py(self, py: Python) -> PyObject {
+                    self.to_object(py)
+                }
+            }
             impl<'source> FromPyObject<'source> for $rust_ty {
                 fn extract(ob: &'source PyAny) -> PyResult<$rust_ty> {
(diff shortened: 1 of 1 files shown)


programmerjake:
Is there a reason `IntoPy` was implemented instead of the more generic `FromPy`?

kngwyu:
No, but I don't think `FromPy` is more generic.

programmerjake:
It is more generic, in the sense that a `FromPy` implementation gives you an implementation for both `FromPy` and `IntoPy`, but implementing `IntoPy` doesn't give you a `FromPy` implementation.

This is entirely analogous to why implementing `From` should be preferred to implementing `Into` whenever possible, as explained in [the docs for `Into`][1]:

> One should only implement [`Into`][1] if a conversion to a type outside the current crate is
> required. Otherwise one should always prefer implementing [`From`][2] over [`Into`][1] because
> implementing [`From`][2] automatically provides one with a implementation of [`Into`][1] thanks to
> the blanket implementation in the standard library. [`From`][2] cannot do these type of
> conversions because of Rust's orphaning rules.

[1]: https://doc.rust-lang.org/std/convert/trait.Into.html
[2]: https://doc.rust-lang.org/std/convert/trait.From.html

PR #625, opened 2019-10-12
Enable to give None as default value for an argument

### LINKED ISSUE #486: When specifying None as a default value for an argument, the attribute panics
## 🐛 Bug Reports

### 🌍 Environment

 - Your operating system and version: Ubuntu 18.04
 - Your python version: 3.7.2
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: pyenv, yes
 - Your rust version (`rustc --version`): rustc 1.36.0-nightly (af98304b9 2019-05-11)
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` I'm using 0.6.0. I haven't tried master.

### 💥 Reproducing

```rust
    #[pyclass]
    struct UUID {
    }

    #[pymethods]
    impl UUID {
        #[new]
        #[args(hex=None, bytes=None, bytes_le=None, fields=None, int=None)]
        fn new(
            obj: &PyRawObject,
            hex: Option<&str>,
            bytes: Option<Py<PyBytes>>,
            bytes_le: Option<Py<PyBytes>>,
            fields: Option<Py<PyTuple>>,
            int: Option<u64>,
        ) {}
}
```

The result is the following error:
```
error: custom attribute panicked --> src/lib.rs:17:5 | 17 | #[pymethods] | ^^^^^^^^^^^^ | = help: message: called `Option::unwrap()` on a `None` value
```

Alexander-N:
Also added some tests which show that it is not strictly necessary to provide `None` explicitly.

Closes #486

### DIFF
--- a/pyo3-derive-backend/src/pymethod.rs
+++ b/pyo3-derive-backend/src/pymethod.rs
@@ -523,11 +523,14 @@ fn impl_arg_param(
         }
     } else if arg.optional.is_some() {
         let default = if let Some(d) = spec.default_value(name) {
-            quote! { Some(#d) }
+            if d.to_string() == "None" {
+                quote! { None }
+            } else {
+                quote! { Some(#d) }
+            }
         } else {
             quote! { None }
         };
-
         quote! {
             let #arg_name = match #arg_value.as_ref() {
                 Some(_obj) => {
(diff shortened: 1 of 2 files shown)


kngwyu:
It looks ergonomic and I like this.
Thank you.

PR #630, opened 2019-10-15
Fix off-by-one error in FromPyObject for BigInt

### LINKED ISSUE #629: OverflowError when converting python int to BigInt
Related: #622 #543 #608 

I think what's happening is that the conversion code is not accounting for the converted bytes needing a sign bit so allocates too few bytes when converting numbers from i8_max to u8_max and all larger multiples of 8 bits

### 🌍 Environment

Ubuntu 18.04
 - Your python version:
Python 3.6.8
apt + virtualenv
rustc 1.40.0-nightly (e413dc36a 2019-10-14
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` 
using commit c7f3cde930dc4480b38549b7c7678e9db7feac7a

### 💥 Reproducing

https://github.com/programmerjake/pyo3_bigint_overflow/tree/7e03b67ef43976e01950811039cc78e1490973ad

Build:
```
maturin develop
```

Test:
```
python -m unittest
```

programmerjake:
Fixes #629

### DIFF
--- a/src/types/num.rs
+++ b/src/types/num.rs
@@ -308,7 +308,7 @@ mod bigint_conversion {
                         } else if n_bits == 0 {
                             0
                         } else {
-                            (n_bits as usize - 1) / 8 + 1
+                            (n_bits as usize - 1 + $is_signed) / 8 + 1
                         };
                         if n_bytes <= 128 {
                             extract_small(ob, n_bytes, $is_signed)
@@ -335,6 +335,8 @@ mod bigint_conversion {
         use crate::types::{PyDict, PyModule};
         use indoc::indoc;
         use num_traits::{One, Zero};
+        use std::any::TypeId;
+        use std::fmt;
 
         fn python_fib(py: Python) -> &PyModule {
             let fib_code = indoc!(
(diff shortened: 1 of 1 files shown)


programmerjake:
tests failing due to unrelated errors caused by pymodule interacting with latest nightly producing an out-of-range array index.

kngwyu:
Thank you for the fix.

src/types/num.rs
```diff
@@ -420,6 +422,49 @@ mod bigint_conversion {
             let zero: BigInt = FromPyObject::extract(fib.call1("fib", (0,)).unwrap()).unwrap();
             assert_eq!(zero, BigInt::from(0));
         }
+
+        /// `OverflowError` on converting python int to BigInt
+        #[test]
+        fn issue_629() {
```
kngwyu: I don't think this name is very understandable.
How about `check_overflow` or so?

src/types/num.rs
```diff
@@ -420,6 +422,49 @@ mod bigint_conversion {
...
+            let gil = Python::acquire_gil();
+            let py = gil.python();
+            fn test<T>(value: T, py: Python)
```
kngwyu: How about change this function into macro?
No one wants to read such a long trait bound in test functions.

src/types/num.rs
```diff
@@ -420,6 +422,49 @@ mod bigint_conversion {
...
+                    assert_eq!(type_id, TypeId::of::<BigUint>());
+                    "BigUint"
+                };
+                println!("{}: {}", type_name, value);
+                let python_value = value.clone().to_object(py);
+                let roundtrip_value = python_value.extract::<T>(py).unwrap();
+                assert_eq!(value, roundtrip_value);
+            }
+            for i in 0..=256usize {
+                test(BigInt::from(i), py);
```
kngwyu: Overflow occurs when `x` in `BifInt::from(x)` consists of 8 * n bits, right?
Then these tests don't look very effective.

But I also did not come up with good test cases, so it's OK if you don't mind.

src/types/num.rs
programmerjake: sounds good, will change in next revision

src/types/num.rs
programmerjake: good idea

src/types/num.rs
programmerjake: The idea is that the tests weren't effective before, so we should test lots of stuff to ensure we didn't miss anything.

programmerjake: everything in `128 <= v && v <= 255` triggers the bug, as well as every 8th value for the tests with `<<`

programmerjake: added comment to that effect

kngwyu:
Thank you!

PR #631, opened 2019-10-15
Fix index error in parameter conversion codes

kngwyu:
Generated codes for argument conversion looks like
```rust
let mut output = [None; N];
let arg0 = match output[0] {
...
};
```
This PR fix this index not to increase for *args and **kwargs.
I hope this PR would fix our build status for the latest nightly.

### DIFF
--- a/pyo3-derive-backend/src/pymethod.rs
+++ b/pyo3-derive-backend/src/pymethod.rs
@@ -394,10 +394,8 @@ pub(crate) fn impl_wrap_setter(
 
 /// This function abstracts away some copied code and can propably be simplified itself
 pub fn get_arg_names(spec: &FnSpec) -> Vec<syn::Ident> {
-    spec.args
-        .iter()
-        .enumerate()
-        .map(|(pos, _)| syn::Ident::new(&format!("arg{}", pos), Span::call_site()))
+    (0..spec.args.len())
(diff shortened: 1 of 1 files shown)


pyo3-derive-backend/src/pymethod.rs
```diff
@@ -394,10 +394,8 @@ pub(crate) fn impl_wrap_setter(
 
 /// This function abstracts away some copied code and can propably be simplified itself
 pub fn get_arg_names(spec: &FnSpec) -> Vec<syn::Ident> {
-    spec.args
-        .iter()
-        .enumerate()
-        .map(|(pos, _)| syn::Ident::new(&format!("arg{}", pos), Span::call_site()))
+    (0..spec.args.len())
```
kngwyu: This is just refactoring, not related to the core of this PR.

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -461,15 +455,15 @@ pub fn impl_arg_params(spec: &FnSpec<'_>, body: TokenStream) -> TokenStream {
 
     let accept_args = bool_to_ident(spec.accept_args());
     let accept_kwargs = bool_to_ident(spec.accept_kwargs());
-
+    let num_normal_params = params.len();
```
kngwyu: This is just refactoring, too.

PR #634, opened 2019-10-18
Pass surrogates in PyString::as_bytes

Alexander-N:
Fixes the handling of surrogate codepoints which would would cause
`PyUnicode_AsUTF8AndSize` to return NULL.

This is my take on #479. Since we have `to_string` and `to_string_lossy` to deal with errors I think `as_bytes` should not return a `PyResult`.

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -11,6 +11,7 @@ use crate::Python;
 use crate::{exceptions, IntoPy};
 use crate::{ffi, FromPy};
 use std::borrow::Cow;
+use std::ffi::CStr;
 use std::ops::Index;
 use std::os::raw::c_char;
 use std::slice::SliceIndex;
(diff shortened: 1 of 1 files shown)


kngwyu:
Thank you, but the memory leak test is failing.
So... maybe `PyUnicode_AsEncodedString` allocates another buffer?
We need to check it, but if so, we should use `AsUTF8AndSize` and return `Result`.

Alexander-N:
> So... maybe PyUnicode_AsEncodedString allocates another buffer?
> We need to check it, but if so, we should use AsUTF8AndSize and return Result.

Yes, that's right. Can you think of some good way to deallocate it after use? I think at least `to_string_lossy` should not return a `PyResult`.

kngwyu:
> Can you think of some good way to deallocate it after use? 

We shouldn't return `&[u8]` type that refers to a deallocated buffer.

> I think at least to_string_lossy should not return a PyResult. The problem I see is that the purpose of to_string_lossy is to replace invalid characters, which it can not do if as_bytes returns an error before.

Yeah, that's a good point.
I'll investigate Python C APIs tomorrow, but have no good idea now.

Alexander-N:
We could let `as_bytes` return a PyResult and use `PyUnicode_AsEncodedString` with errors='replace' in let `to_string_lossy`. If this is an option I could use some hint how to ensure that `Py_DECREF` gets called.

kngwyu:
For now, let's resolve the `as_bytes` problem first.
I noticed that it can return NULL when the internal string is UTF-16 or UTF-32, which seems problematic.

Let's modify `to_string_lossy` in the next PR.

Alexander-N:
> For now, let's resolve the as_bytes problem first.
> I noticed that it can return NULL when the internal string is UTF-16 or UTF-32, which seems problematic.
> 
> Let's modify to_string_lossy in the next PR.

I think it would be preferable to do this in one PR since breaking `to_string_lossy` seems equally problematic. The changes to `to_bytes`  can not be done independently since letting it return `PyResult` means that `to_string_lossy` also has to return a `PyResult` (also every call site has to be changed).

kngwyu:
Hmm...,
> The changes to to_bytes can not be done independently since letting it return PyResult means that to_string_lossy also has to return a PyResult

You can use `unwrap` until the true fix.

But if you want to fix `to_string_lossy` in this PR, it would look like
```rust
    pub fn to_string_lossy(&self) -> Cow<str> {
        match self.as_bytes() {
            Ok(s) => String::from_utf8_lossy(self.as_bytes()),
            Err(_) => {
                // Fallback if the internal string is not a valid UTF-8 one
                let encoded = ffi::PyUnicode_AsEncodedString(...);
                self.py().register_pointer(encoded);  // ensure DECREF will be called
                let data = ffi::PyUnicode_UTF8(encoded);
                String::from_utf8_lossy(data)
            }
        }
    }
```
where `PyUnicode_UTF8` is a macro in `unicodeobject.c` and needs re-implementation.

pganssle:
It may be worth adding hypothesis and/or [quickcheck](https://github.com/BurntSushi/quickcheck) tests for this kind of thing, as they are likely to catch new issues like this earlier in the process.

Alexander-N:
Thanks @kngwyu. I will open separate pull requests for these issues.

PR #635, opened 2019-10-19
Fix minimum and maximum for datetime tests

### LINKED ISSUE #587: Datetime test fails in my local machine
[This line](https://github.com/PyO3/pyo3/blob/master/examples/rustapi_module/tests/test_datetime.py#L57) fails in my local environment(Arch Linux 64bit).
```bash
Python 3.7.4 (default, Jul 16 2019, 07:12:58) 
[GCC 9.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime as pdt
>>> MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(253402300799)
ValueError: year 0 is out of range
```

IMHO, we really need such a strict test?

Alexander-N:
Depending on the local timezone one could get
```python 
ValueError: year 0 is out of range
```
This does fix the tests for me, what do you think @pganssle?

Closes #587

### DIFF
--- a/examples/rustapi_module/tests/test_datetime.py
+++ b/examples/rustapi_module/tests/test_datetime.py
@@ -40,21 +40,8 @@ MIN_DAYS = pdt.timedelta.min // pdt.timedelta(days=1)
 MAX_MICROSECONDS = int(pdt.timedelta.max.total_seconds() * 1e6)
 MIN_MICROSECONDS = int(pdt.timedelta.min.total_seconds() * 1e6)
 
-IS_X86 = platform.architecture()[0] == "32bit"
-IS_WINDOWS = sys.platform == "win32"
-if IS_WINDOWS:
-    MIN_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(86400)
-    if IS_X86:
-        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(32536789199)
-    else:
-        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(32536799999)
-else:
-    if IS_X86:
-        MIN_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(-2147483648)
-        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(2147483647)
-    else:
-        MIN_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(-62135510400)
-        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(253402300799)
+MIN_DATETIME = pdt.datetime.min
+MAX_DATETIME = pdt.datetime.max
 
 PYPY = platform.python_implementation() == "PyPy"
 HAS_FOLD = getattr(pdt.datetime, "fold", False)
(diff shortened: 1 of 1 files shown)


Alexander-N:
Does not work for windows..

Alexander-N:
Works with the reduced datetime range. Not a perfect solution, but I guess it's better than a flaky test.

examples/rustapi_module/tests/test_datetime.py
```diff
@@ -40,21 +40,8 @@ def tzname(self, dt):
...
-    else:
-        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(32536799999)
-else:
-    if IS_X86:
-        MIN_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(-2147483648)
-        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(2147483647)
-    else:
-        MIN_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(-62135510400)
-        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime.fromtimestamp(253402300799)
+MIN_DATETIME = pdt.datetime(1970, 1, 1)
```
pganssle: I agree with using `pdt.datetime` literals, but I disagree with both of these choices, they are reducing the effectiveness of the tests by arbitrarily removing a lot of the interesting edge cases!

I think we should start by switching the current conditional boundaries over to using datetime literals instead of `fromtimestamp`, then possibly (probably in a second PR), add some additional tests that run the datetime-related tests in different time zone locales.

Alexander-N: I agree this would be much better. But I think removing the edge cases until then would still be an improvement since I fear that at the moment any errors these tests will find will be ignored because of its flakiness e.g. https://ci.appveyor.com/project/fafhrd91/pyo3/builds/28239822/job/dnquxyqc6sratqlk. Also at the moment they are causing a bit of inconvenience by permanently failing locally for several developers.

pganssle: Here is the new version (I have tested it, it does not cause the failures):

```python
if IS_WINDOWS:
    MIN_DATETIME_FROM_TIMESTAMP = pdt.datetime(1970, 1, 2, 0, 0)
    if IS_X86:
        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime(3001, 1, 19, 4, 59, 59)
    else:
        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime(3001, 1, 19, 7, 59, 59)
else:
    if IS_X86:
        # TS ±2147483648 (2**31)
        MIN_DATETIME_FROM_TIMESTAMP = pdt.datetime(1901, 12, 13, 20, 45, 52)
        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime(2038, 1, 19, 3, 14, 8)
    else:
        MIN_DATETIME_FROM_TIMESTAMP = pdt.datetime(1, 1, 2, 0, 0)
        MAX_DATETIME_FROM_TIMESTAMP = pdt.datetime(9999, 12, 31, 18, 59, 59)
```

Alexander-N: Thanks, I changed it as suggested.

kngwyu:
@Alexander-N @pganssle 
Thank you for the fix!

PR #637, opened 2019-10-20
Implement conversion traits for PyBytes

### LINKED ISSUE #373: bytes should translate to &[u8]

Alexander-N:
Closes #373

### DIFF
--- a/tests/test_bytes.rs
+++ b/tests/test_bytes.rs
@@ -1,0 +1,40 @@
+use pyo3::prelude::*;
+use pyo3::types::PyBytes;
+use pyo3::wrap_pyfunction;
+
+mod common;
+
+#[pyfunction]
+fn get_pybytes() -> PyObject {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+
+    PyBytes::new(py, b"Hello World").into()
+}
+
+#[pyfunction]
+fn bytes_pybytes_conversion(bytes: &[u8]) -> &[u8] {
+    bytes
+}
+
+#[test]
+fn test_get_pybytes() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+
+    let get_bytes = wrap_pyfunction!(get_pybytes)(py);
+    py_assert!(py, get_bytes, "get_bytes() == b'Hello World'");
+}
+
+#[test]
+fn test_pybytes_bytes_conversion() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+
+    let bytes_pybytes_conversion = wrap_pyfunction!(bytes_pybytes_conversion)(py);
+    py_assert!(
+        py,
+        bytes_pybytes_conversion,
+        "bytes_pybytes_conversion(b'Hello World') == b'Hello World'"
+    );
+}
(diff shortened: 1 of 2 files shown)


tests/test_bytes.rs
```diff
@@ -0,0 +1,40 @@
+use pyo3::prelude::*;
+use pyo3::types::PyBytes;
+use pyo3::wrap_pyfunction;
+
+mod common;
+
+#[pyfunction]
+fn get_pybytes() -> PyObject {
```
kngwyu: Since we recommend users to return `&Py~` types(in this case, `&PyBytes`), I don't think we need this test.

kngwyu:
Thanks.

tests/test_bytes.rs
Alexander-N: Done

Alexander-N: Actually, it is not clear to me how to return `&PyBytes` directly, see https://gitter.im/PyO3/Lobby?at=5d83e8d456677e767a225079

kngwyu:
```rust
#[pyclass]
struct ByteSequence {
    elements: Vec<u8>,
}

#[pymethods]
impl ByteSequence {
    fn bytes<'py>(&'py self, py: Python<'py>) -> &'py PyBytes {
        PyBytes::new(py, &self.elements)
    }
}
```

Alexander-N: I see, thanks. Should this be added as a test?

Alexander-N: Ok, I saw it already is.

PR #639, opened 2019-10-22
Guard against PyUnicode_AsUTF8AndSize returning null in PyString::as_bytes

Alexander-N:
See discussion in #479 and #634.


kngwyu:
Thank you!

PR #641, opened 2019-10-23
FFI for PEP 590 Vectorcall

ijl:
https://www.python.org/dev/peps/pep-0590/

This was tested on 3.7 using _PyCFunctionFast and 3.8
using PyObject_Vectorcall. Extending pyo3-derive-backend
to generate code using vectorcall is not done here.

This exposes PyObject_Vectorcall with a link_name to the
underscored name on 3.8 because it is expected to be stabilized
on 3.9.

This fixes the "fast" objects being new in 3.7, not 3.6.


kngwyu:
Thanks!
Vectorcall looks really nice and I want to use it in PyO3 internally in the future.

PR #642, opened 2019-10-23
Fix handling of invalid utf-8 sequences in PyString::to_string_lossy

Alexander-N:
See discussion in #479 and #634.

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -3,6 +3,7 @@
 use crate::conversion::FromPyObject;
 use crate::conversion::{PyTryFrom, ToPyObject};
 use crate::err::{PyErr, PyResult};
+use crate::gil;
 use crate::instance::PyNativeType;
 use crate::object::PyObject;
 use crate::types::PyAny;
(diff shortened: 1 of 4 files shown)


src/types/string.rs
```diff
@@ -87,10 +88,29 @@ impl PyString {
...
+                        self.0.as_ptr(),
+                        CStr::from_bytes_with_nul(b"utf-8\0").unwrap().as_ptr(),
+                        CStr::from_bytes_with_nul(b"surrogatepass\0")
+                            .unwrap()
+                            .as_ptr(),
+                    );
+                    // Since we have a valid PyString and replace any surrogates, assume success.
+                    debug_assert!(!py_bytes.is_null());
+                    // ensure DECREF will be called
+                    self.py().from_owned_ptr::<PyBytes>(py_bytes);
```
kngwyu: How about `crate::gil::register_pointer`?(My snippet was incorrect, sorry)

kngwyu:
Thanks!

src/types/string.rs
Alexander-N: Ok, done. What's the difference?

kngwyu: 1. Less overhead.
2. Our GIL pool has 3 types of buffer(owned ref, borrowed ref, pointer). 
In this case, we can use both owned and pointer buffer.
But pointer buffer is preferable for the case `GILGuard` drops earlier(I'll add an example but have no time now :sweat: ).

Issue #611, opened 2019-10-03
Implementing iterators in PyMappingProtocol


sebpuetz:
## 🐛 Bug Reports

When reporting a bug, please provide the following information. If this is not a bug report you can just discard this template.

### 🌍 Environment

 - Your operating system and version: Ubuntu 18.04
 - Your python version: 3.73
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: pyenv
 - Your rust version (`rustc --version`): rustc 1.38.0-nightly (311376d30 2019-07-18)
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` yes

### 💥 Reproducing

Implementing `PyIterProtocol` for a type implementing `PyMappingProtocol` seems to overwrite `__iter__` in Python. Invoking an iterator through any kind of statement calls `PyIterProtocol::__iter__`. `PyMappingProtocol::__iter__` is not accessible. Checking whether an element is in `mapping` through the `in` keyword runs an iterator and compares items instead of using `__contains__`. `__contains__` can be invoked explicitly through `PyMapping.__contains__`. 

If the implementation of `PyIterProtocol` is removed for `PyMapping`, statements such as:

~~~Python
import pymapping
mapping = pymapping.Mapping(["a", "b", "c", "d"]) 
for k in mapping:
    print(k)

print("a" in mapping)
~~~

fail by stating that `mapping` is not iteratable. Although, the iterator can be called explicitly through `PyMapping.__iter__` but using `iter()` on a `PyMapping` object fails:

~~~Python
In [1]: import pymapping                                                                                                                                                                             

In [2]: mapping = pymapping.Mapping(["a", "b", "c", "d"])                                                                                                                                            

In [3]: iter(mapping)                                                                                                                                                                                
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-73d010d9f2e9> in <module>
----> 1 iter(mapping)

TypeError: 'Mapping' object is not iterable

In [4]: mapping.__iter__()                                                                                                                                                                           
own_iter
Out[4]: <MappingIter at 0x7f234178de40>
~~~

~~~Rust
#![feature(specialization)]

#[pymodule]
fn pymapping(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<PyMapping>()?;
    Ok(())
}

use pyo3::prelude::*;
[...]

PR #644, opened 2019-10-25
Remove contains and iter from PyMappingProtocol.

sebpuetz:
The methods are not expected by CPython and are only explicitly
callable. To get iteration support, PyIterProtocol should be
implemented and to get support for `x in mapping`,
PySequenceProtocol's __contains__ should be implemented.

https://github.com/PyO3/pyo3/issues/611

 - [x] Run `cargo fmt` (This is checked by travis ci)
 - [x] Run `cargo clippy` and check there are no hard errors (There are a bunch of existing warnings; This is also checked by travis)
 - [x] If applicable, add an entry in the changelog.
 - [x] If applicable, add documentation to all new items and extend the guide.
   - Could be nice to eventually add a Rust implementation of a PyDict under examples.
 - [x] If applicable, add tests for all new or fixed functions


kngwyu:
Thanks!
Looks CI fails because of `rustfmt` check.
Could you please run `cargo fmt` on your local machine?

sebpuetz:
Sure, amended the commit!

 Looks like I forgot to run that after removing the defs from the derive backend.

kngwyu:
Thanks, again.

PR #651, opened 2019-10-26
Fix proc macro for init fn.

sebpuetz:
The if statement evaluated to true if the return type was not `"()"` or `"PyResult<()>"` which was always the case since `"()" != "PyResult<()>"`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 and `PyString::to_string_lossy` [#642](https://github.com/PyO3/pyo3/pull/642).
 * Remove `__contains__` and `__iter__` from PyMappingProtocol. [#644](https://github.com/PyO3/pyo3/pull/644)
 * Fix proc-macro definition of PySetAttrProtocol. [#645](https://github.com/PyO3/pyo3/pull/645)
+* Fix `impl_wrap_init`. [#651](https://github.com/PyO3/pyo3/pull/651)
 
 
 ## [0.8.1]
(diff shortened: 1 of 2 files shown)


sebpuetz:
`test_datetime_check` failed on an earlier build for windows on python 3.6 with an unwrap on Err which must have originated this method:
https://github.com/PyO3/pyo3/blob/da1ab1b2b82334138a6bc8a4fc3c5ff45d602b99/tests/test_datetime.rs#L8-L35

The tests passed after pushing the same code again.

kngwyu:
Thank you, but... is this feature actually used?
This is called for 
```rust
impl Hoge {
    #[init] 
    fn __init__(...) -> {}
}
```
, but I can't find any document about this.
If this feature is not used, we should remove it instead of this fix.

@konstin 
Do you know what's this?

sebpuetz:
It does get called, although I'm not sure about the use of it:

~~~Rust
#[pyclass]
pub struct Foo {
    bar: usize,
}

#[pymethods]
impl Foo {
    #[init]
    fn init(&mut self) -> () {
        println!("{}", self.bar);
        self.bar = 10;
        println!("{}", self.bar);
    }

    #[new]
    fn new(obj: &PyRawObject) -> PyResult<()> {
        obj.init(Self { bar: 0 });
        Ok(())
    }
}
~~~

~~~
>>> Foo()
0
10
~~~

konstin:
>  Do you know what's this?

Unfortunately not

programmerjake:
> > Do you know what's this?
> 
> Unfortunately not

I would guess `#[init]` is the equivalent of Python's `__init__` method, whereas `#[new]` is the equivalent of `__new__`.

Alexander-N:
Relevant: https://github.com/dgrunwald/rust-cpython/issues/11#issuecomment-472638633

kngwyu:
Thank you for comments.
I opened #658 which removes `#[init]` feature.
Please feel free to write your opinion on the thread.

kngwyu:
I'm sorry for closing this PR via #658. 
Thank you for your contribution, though.

Issue #664, opened 2019-11-12
PyRawObject API unsafe


programmerjake:
`PyRawObject` doesn't track initialization:
```Rust
#[pyclass]
struct MyType {
    value: String,
}

#[pymethods]
impl MyType {
    #[new]
    fn new(obj: &PyRawObject) -> PyResult<()> {
        obj.init(get_my_type_or_fail()?); // can fail to initialize, dropping uninitialized String
        obj.init(MyType { value: "abc".into() }); // can initialize twice, leaking memory
        Ok(())
    }
}
fn get_my_type_or_fail() -> PyResult<MyType> {
    unimplemented!()
}
```

`PyRawObject` doesn't check types:
```Rust
#[pyclass]
struct MyType {
    value: String,
}
#[pyclass]
struct OtherType {
    value: f32,
}

#[pymethods]
impl MyType {
    #[new]
    fn new(obj: &PyRawObject) {
        obj.init(OtherType {value:1.0}); // initialize with incorrect type
        let value: &OtherType = obj.as_ref(); // get reference with incorrect type
        println!("{}", value.value); // access using incorrect type
    }
}
```

programmerjake:
I think it would be pretty common to forget to call `init` when handling errors, such as by using the `?` operator. The rest of these problems are mostly theoretical.

kngwyu:
Thank you for reporting.
But since I'm planning a huge design change in this area, I don't want to fix this immediately.

programmerjake:
> Thank you for reporting.
> But since I'm planning a huge API modification in this area, I don't want to fix this immediately.

Would you be open to a pull request that just fixes initialization tracking so it panics, doesn't drop uninitialized memory, and/or doesn't double-initialize memory?

That should be pretty easy to do without a user-facing API change, would catch the most common mistake (forgetting to initialize), and can be replaced with your rewritten code when you're ready.

kngwyu:
OK.
I’ll add assertions for that.

Alexander-N:
Fixed by #683

Issue #679, opened 2019-12-01
Why object creation is slow in PyO3 and how to enhance it


kngwyu:
This is a support issue following #661, where I explain *why* PyO3’s object creation can be slow and *how* we can speed up this.
Since I joined this project a year ago, I’m not sure my understanding is correct, especially about the *historical* aspect of the design.

cc: @ijl @ehiggs

# TL; DR
- We have an overhead around object creation.
- For `PyObject` and `Py<T>`, we cannot remove this overhead.
- For current `&PyAny, &PyDict` or so, we can remove this overhead by changing them to `PyAny<'py>, PyDict<'py>`, and so on.
- But for it to come true, we need a careful design decision about how to handle owned/borrowed pointers.
- Once this would come true, we could replace many usages of `PyObject` with low-cost `PyAny<'py>` types(and then we should rename PyAny with PyObject).

## First, let’s revisit our 2 kinds of object

It is a really [confusing thing](https://github.com/PyO3/pyo3/issues/356), but we have 2 types for representing Python’s object, named `PyObject` and `PyAny` .
So, what’s the difference between them?
The answer is **PyAny is forced to use as &’py PyAny, which has the same lifetime as GILGuard, but PyObject isn’t.**
```rust
    let any: &PyAny = {
        let gil = Python::acquire_gil();
        let dict: &PyDict = PyDict::new(gil.python());
        dict.as_ref()
    };
```
Thus, this snippet causes a compile error, because `&PyAny` has the same lifetime as `GILGuard`.
CPython has a reference-count based GC, which really differs from Rust’s lifetime based memory management system. For integrating these two systems, it’s a natural choice to represent Python objects’ lifetime by GILGuard(= our RAII guard that corresponds to Python’s thread lock).

In contrast, PyObject can exist even after `GILGuard` drops.
```rust
    let obj: PyObject = {
        let gil = Python::acquire_gil();
        let dict: &PyDict = PyDict::new(gil.python());
        dict.to_object()
    };
```
This ‘not bounded’ nature of `PyObject` helps users to write complicated extensions. For example, you can send `PyObject` into another thread, though you cannot send `PyAny`.
However, `&PyAny` is a sufficient and reasonable choice for many use cases.

For other types than `PyAny` like `PyDict`, this ‘not bounded’ types are represented by `Py<T>` wrapper type.

| Bounded by GIL      | Not bounded               |
| ------------------- | ------------------------- |
| `&PyAny`              | `PyObject`                  |
[...]

PR #683, opened 2019-12-07
New #[pyclass] system that is aware of its concrete layout

kngwyu:
This PR introduces a new `#[pyclass]` system that uses `struct PyClassShell` instead of offset access.
I'll add a more detailed explanation later.
cc: @davidhewitt 

### TODO
- [x] Modify derive-backend
- [x] Renew `#[new]` API
- [x] Renew `super().__init()__`
- [x] Implement `weakref` and `dict` support
- [x] Give correct `ConcreteLayout`s for C-FFI types
  - [x] bool
  - [x] bytearray
  - [x] complex
  - [x] datetime
  - [x] dict
  - [x] float
  - [x] int
  - [x] list
  - [x] module
  - [x] set
  - [x] slice
  - [x] string
  - [x] tuple
- [x] Fix tests
- [x] Fix guides
- [x] Fix examples
- [x] Proof read docs
- [x] Check naming issues
- [x] CHANGELOG


kngwyu:
Still incomplete, but here I note the rough sketch.
This PR introduces `PyClassShell`, which represents the concrete layout of `#[pyclass]` in Python heap.
Currently, we use a hacky way to deal with `#[pyclass]`.
Specifically, we use `OFFSET` for getting` &T: PyClass` from `*mut ffi::PyObject`, like this.
```rust
(value.as_ptr() as *const u8).offset(T::OFFSET) as *const T
```
(From https://github.com/PyO3/pyo3/blob/v0.8.4/src/conversion.rs#L399)

By using `PyClassShell`, we can write this operation by (roughly)
```rust
(*(value.as_ptr() as *const PyClassShell<T>).).pyclass
```
Additionally, this PR does
- Replace `PyRef` with `&PyClassShell`
- Replace `PyRefMut` with `&mut PyClassShell`
- Remove `PyTypeCreate` and  `PyObjectAlloc`
- Introduce `PyClassAlloc`, `PyConcreteObject` for `tp_alloc` and `tp_dealloc`
- Remove `PyRawObject`
- Introduce `PyClassInitializer`, which changes  initialization API dramatically

Now `#[new]` doesn't take `PyRawObject`.
It instead is a roughly `FnOnce(*args, **kwargs) -> impl IntoInitializer`.
```rust
#[pymethods]
impl BaseClass {
    #[new]
    fn new() -> Self {
        BaseClass { val1: 10 }
    }
}

#[pyclass(extends=BaseClass)]
struct SubClass {
    #[pyo3(get)]
    val2: usize,
}

#[pymethods]
impl SubClass {
    #[new]
    fn new() -> PyClassInitializer<Self> {
        let mut init = PyClassInitializer::from_value(SubClass { val2: 5 });
        init.get_super().init(BaseClass { val1: 10 });
        init
    }
}
``` 
If there is an uninitialized *Rustic* baseclass, `PyClassInitializer` raise an exception, which resolves #664.
Indeed we need more discussion, I believe this API is much easier to use.

programmerjake:
One thing that should be supported (I didn't check) is the ability to skip initialization and throw away the new object when raising an exception from `#[new]`. The code would need to handle destroying a partially initialized object (some rust base classes could have been initialized when the failure is detected).

It would also be nice, but not required, to support conditionally returning an unrelated object from `#[new]`, rather than the newly allocated object, similar to how Python's `enum.Enum` constructor works (it always returns the selected enumerant object instead of constructing a new object).

src/type_object.rs
```diff
@@ -31,17 +58,14 @@ pub trait PyTypeInfo {
...
-
     /// Type flags (ie PY_TYPE_FLAG_GC, PY_TYPE_FLAG_WEAKREF)
     const FLAGS: usize = 0;
 
     /// Base class
-    type BaseType: PyTypeInfo;
+    type BaseType: PyTypeInfo + PyTypeObject;
+
+    /// Layout
+    type ConcreteLayout: PyConcreteObject<Self>;
```
davidhewitt: Is it necessary for the `ConcreteLayout` to be part of the `TypeInfo` trait? It seems like this is only needed for `#[pyclass]` types. Having it here creates a lot of work to get the right definitions for all the py native types, but I'm not sure if it helps.

src/pyclass.rs
```diff
@@ -0,0 +1,533 @@
...
+
+        unsafe { NonNull::new_unchecked(type_object) }
+    }
+}
+
+/// `PyClassShell` represents the concrete layout of our `#[pyclass]` in the Python heap.
+#[repr(C)]
+pub struct PyClassShell<T: PyClass> {
+    ob_base: <T::BaseType as PyTypeInfo>::ConcreteLayout,
+    pyclass: ManuallyDrop<T>,
```
davidhewitt: This `PyClassShell` idea is nice, and is heading in the direction that `pybind11` goes. There's two main differences between this idea and what `pybind11` does:

1. The `pybind11` equivalent just stores `*mut void` instead of being generic over `T`. This means that the python type allocation for all pybind11 classes is more or less exactly the same. It also means smaller binary sizes, because there's only one copy of the `PyClassShell` methods instead of one for each `T`.
1. `pybind11` allows references to be stored in the `PyClassShell` as well as owned values. In Rust speak, instead of `pyclass: ManuallyDrop<T>`, we'd have something like:

       enum PyClassValue<T> {
           Borrowed(*const T),
           MutBorrowed(*mut T),
           Owned(ManuallyDrop<T>),
       }

The great thing about storing references in `PyClassShell` is that it allows us to do a lot less copying when going Rust -> Python -> Rust.

For example, the code below doesn't currently compile with `pyo3` because the `&Address` return type. If we could store references in the `PyClassShell`, then this should be fairly straightforward to support:

    #[pyclass]
    struct Address {
        value: String
    }

    #[pyclass]
    struct Employee {
        address: Address
    }

    #[pymethods]
    impl Employee {
        pub fn address(&self) -> &Address {
            &self.address
       }
    }

programmerjake:
> * It would be great if we can keep `PyClassShell<T>` out of the `pyo3` api, so that users can just use `&T` / `&mut T`. I don't know how easy this is.

We need to make sure that we never can simultaneously hand out more than one `&mut T` or both `&mut T` and `&T`, so there should be some method of tracking that (like [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html)).

programmerjake:
Also, since python modifies refcounts and stuff, that part needs to be wrapped in `UnsafeCell` so Rust doesn't assume it's immutable.

davidhewitt:
> We need to make sure that we never can simultaneously hand out more than one &mut T or both &mut T and &T, so there should be some method of tracking that (like RefCell).

Agreed.

kngwyu:
> we'd have something like:
```rust
enum PyClassValue<T> {
    Borrowed(*const T),
    MutBorrowed(*mut T),
    Owned(ManuallyDrop<T>),
}
```

We never should do so.
If we are to expose Rust's references to Python world, we should handle the lifetime explicitly, or we'd break Rust's safe rule(=no dangling pointer, no memory leak).

davidhewitt:
> If we are to expose Rust's references to Python world, we should handle the lifetime explicitly, or we'd seriously break Rust's safe rule(=no dangling pointer, no memory leak).

I agree with this in principle, but I don't think it's possible. As far as I know, Rust can't express the lifetime of the borrow in Python.

That's why a keep-alive mechanism is necessary. I can demonstrate a PR for this if you like.

kngwyu:
@davidhewitt 
Thank you for the proposal, but we should discuss that *keep-alive* feature in the other thread.
This PR does not intend to add another features to PyO3, but to clean up some dirty parts.

davidhewitt:
:+1: that makes total sense.

It's for reasons like this that I recommend we keep `PyClassShell` out of the public pyo3 API, if possible.

Then we can consider adding features like reference support to without more breaking changes. I'll write a PR for that another time.

kngwyu:
> It's for reasons like this that I recommend we keep PyClassShell out of the public pyo3 API,

Sorry I cannot understand the relationship between this claim and the above discussions.

src/type_object.rs
kngwyu: If we want to give a fixed size to `PyClassShell`, we need it.

> Having it here creates a lot of work to get the right definitions for all the py native types,

Yeah but I'll do it by script.

kngwyu:
I added some additional features to this PR.
1. Safe inheritance of native types
Currently this struct causes SIGSEGV when we use it as dict.
```rust
#[pyclass(extends=PyDict)]
struct DictWithName {
    #[pyo3(get(name))]
    _name: &'static str,
}
```
This SIGSEGV is fixed by properly using baseclass's `tp_new`.

2. Inhibits inheriting PyVarObject
[PyVarObject](https://docs.python.org/3.8/c-api/structures.html?highlight=pyvarobject#c.PyVarObject) has a variable length so it can be problematic to inherit it with our current *fixed layout* strategy.
E.g., now we cannot do
```rust
#[pyclass(extends=PyTuple)]
struct TupleWithName {
    _name: &'static str,
}
```

src/pyclass.rs
```diff
@@ -0,0 +1,557 @@
...
+                return Err(PyErr::new::<RuntimeError, _>(format!(
+                    "Base class '{}' is not initialized",
+                    $name
+                )));
+            };
+        }
+        let PyClassInitializer { init, super_init } = self;
+        if let Some(value) = init {
+            unsafe { shell.py_init(value) };
+        } else if !T::ConcreteLayout::NEED_INIT {
```
programmerjake: is the condition reversed here, erroring when `NEED_INIT` is `false` seems backwards.

src/pyclass.rs
```diff
@@ -0,0 +1,557 @@
...
+    pub fn new() -> Self {
+        PyClassInitializer {
+            init: None,
+            super_init: None,
+        }
+    }
+
+    #[must_use]
+    #[doc(hiddden)]
+    pub fn init_class(self, shell: &mut T::ConcreteLayout) -> PyResult<()> {
```
programmerjake: shouldn't this be `unsafe`, it seems likely to be UB to invoke twice on the same `shell` object

davidhewitt:
> PyVarObject has a variable length so it can be problematic to inherit it with our current fixed layout strategy.

We would presumably be able to work around this if the layout contained a pointer to the actual tuple object, e.g. `*const PyTupleObject` ?

This is actually how C++'s `pybind11` does all its layout. The `PyClassShell` contains a pointer to the actual struct, as well as a pointer for each Python base.

src/pyclass.rs
kngwyu: > erroring when NEED_INIT is false seems backwards

Sorry I'm not sure what you are saying..., but you mean we should not raise an error for redundant initialization?

src/pyclass.rs
kngwyu: Hmm :thinking: ... , I don't think it's UB.
`init` just do `shell.pyclass = value`.

kngwyu:
> This is actually how C++'s pybind11 does all its layout. 

Really? :thinking: 
It looks it has just `PyObject_HEAD` and value pointer/
https://github.com/pybind/pybind11/blob/v2.4.3/include/pybind11/detail/common.h#L382###

davidhewitt:
> Really? 🤔

Yes:

- The `simple_value_holder` variant of the union is the pointer to the struct plus potentially an `Rc` or a `Box` if pybind11 has responsibility of the lifetime of the pointer.
- The `nonsimple_values_and_holders` is a pointer to a dynamically allocated chunk of memory storing pointers to the struct as well as all its Python bases.

kngwyu:
Thank you for the explanation.

> The nonsimple_values_and_holders is a pointer to a dynamically allocated chunk of memory storing pointers to the struct as well as all its Python bases.

So it can store tuples or lists?
It sounds really weird for me..., but maybe it's better we discuss it in another thread.

davidhewitt:
👍 I'll find some time over the holidays to post a new issue which proposes how we could a similar thing as `nonsimple_values_and_holders`.

guide/src/class.md
```diff
@@ -15,57 +15,111 @@ struct MyClass {
...
 
-You can use `PyRef` just like `&T`, because it implements `Deref<Target=T>`.
+## Get Python objects from `pyclass`
+You sometimes need to convert your `pyclass` into a Python object in Rust code (e.g., for testing it).
+
+You can use `PyClassShell<T>` or `Py<T>` for this purpose.
+
+### `PyClassShell`
+
+`PyClassShell` is a special type made for `PyClass`.
```
Alexander-N:
```suggestion
`PyClassShell` is a special type made for `pyclass`.
```
I'm not sure if this sentence helps, maybe we could drop it? I found the documentation for `PyRef` very clear, does it still apply for `PyClassShell`?

> 
> You can get an instance of `PyRef` by `PyRef::new`, which does 3 things:	
> 1. Allocates a Python object in the Python heap	
> 2. Copies the Rust struct into the Python object	
> 3. Returns a reference to it	
>

guide/src/class.md
```diff
@@ -15,57 +15,111 @@ struct MyClass {
...
+    const NAME: &'static str = "MyClass";
+    const MODULE: Option<&'static str> = None;
+    const DESCRIPTION: &'static str = "This is a demo class";
+    const FLAGS: usize = 0;
+
+    #[inline]
+    unsafe fn type_object() -> &'static mut pyo3::ffi::PyTypeObject {
+        static mut TYPE_OBJECT: pyo3::ffi::PyTypeObject = pyo3::ffi::PyTypeObject_INIT;
+        &mut TYPE_OBJECT
+    }
```
programmerjake: This should be:
```rust
    unsafe fn type_object() -> *mut pyo3::ffi::PyTypeObject {
        static TYPE_OBJECT: UnsafeCell<pyo3::ffi::PyTypeObject> = UnsafeCell::new(pyo3::ffi::PyTypeObject_INIT);
        TYPE_OBJECT.get()
    }
```

since mutable references to `static mut` are nearly always undefined behavior.

See https://users.rust-lang.org/t/why-is-static-mut-bad/32888
and https://github.com/rust-lang/rust/issues/53639

The function still needs to be `unsafe` so users can't safely mis-implement it.

src/pyclass.rs
programmerjake: doesn't this call the CPython init functions for the base classes? Those are likely to not be allowed to be called twice on the same object.

src/pyclass.rs
programmerjake: looks good!

guide/src/class.md
kngwyu: Yeah, but this is not a new part that this PR introduced.
It is copy-pasted from type_object.rs.

Though I completely agree that it is problematic, maybe what I need is a mechanism that inhibits access to uninitialized TYPE_OBJECT. 
Let's do it in another PR.

src/pyclass.rs
kngwyu: No, it's a part of `tp_new` and doesn't call init function.

kngwyu:
Thank you for reviews.

> but from a users perspective PyClassShell is a bit unwieldy. 

Yeah, I also think it is just too long, but haven't come up with a good name.
`PyHeap` sounds good, though a bit confusing since Python has `HEAP_TYPE` type flag.

src/pyclass.rs
programmerjake: ah, ok. Related, if `shell` is partially uninitialized, it would be UB for safe code to read from it, so it should have `MaybeUninitialized` and/or be passed in as `*mut T::ConcreteLayout`. Also, if later code depends on `shell` being initialized instead of being `uninit`, then that's also a good reason for `init_class` to be `unsafe`.

guide/src/class.md
programmerjake: Continued at #699

src/instance.rs
```diff
@@ -275,14 +35,15 @@ unsafe impl<T> Sync for Py<T> {}
 
 impl<T> Py<T> {
     /// Create new instance of T and move it under python management
-    pub fn new(py: Python, value: T) -> PyResult<Py<T>>
+    pub fn new(py: Python, value: impl IntoInitializer<T>) -> PyResult<Py<T>>
```
davidhewitt: It might be possible to use `TryInto<PyClassInitializer<T>, Error = PyErr>` instead of creating a whole new trait?

src/instance.rs
```diff
@@ -357,47 +118,15 @@ impl<T> Py<T> {
...
-{
-    #[inline]
-    fn as_ref(&self, py: Python) -> PyRef<T> {
-        PyRef::from_ref(self.as_ref_dispatch(py))
-    }
-    #[inline]
-    fn as_mut(&mut self, py: Python) -> PyRefMut<T> {
-        PyRefMut::from_mut(self.as_mut_dispatch(py))
+impl<T: PyTypeInfo> AsPyRef<T> for Py<T> {
+    fn as_ref(&self, _py: Python) -> &T {
```
davidhewitt: Given that `Python` is unused here, what about a blanket impl

    impl<T: AsRef<T>> AsPyRef<T> for T 

and then implementing `AsRef<T> for Py<T>`.

src/instance.rs
```diff
@@ -412,8 +141,8 @@ impl<T> IntoPy<PyObject> for Py<T> {
     /// Converts `Py` instance -> PyObject.
     /// Consumes `self` without calling `Py_DECREF()`
     #[inline]
-    fn into_py(self, py: Python) -> PyObject {
-        unsafe { PyObject::from_owned_ptr(py, self.into_ptr()) }
+    fn into_py(self, _py: Python) -> PyObject {
+        unsafe { PyObject::from_not_null(self.into_non_null()) }
```
davidhewitt: Naming is inconsistent: `not_null` and `non_null`. Maybe one can be tweaked so they're consistent?

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+pub struct PyClassShell<T: PyClass> {
+    ob_base: <T::BaseType as PyTypeInfo>::ConcreteLayout,
+    pyclass: ManuallyDrop<T>,
+    dict: T::Dict,
+    weakref: T::WeakRef,
+}
+
+impl<T: PyClass> PyClassShell<T> {
+    /// Make new `PyClassShell` on the Python heap and returns the reference of it.
+    pub fn new_ref(py: Python, value: impl IntoInitializer<T>) -> PyResult<&Self>
```
davidhewitt: I'm trying to get my head around the refcounting here.

The return type is `&PyClassShell<T>`, so presumably this is a borrowed reference? Is the refcount of the allocated python object 1 or 0 at this point?

Alexander-N:
Would it be possible to combine `Py` and `PyClassShell` so that the user can write `Py<T>` instead of `PyClassShell<T>` and `Py` could be read as "managed by Python"?

src/pyclass.rs
kngwyu: We have **really** complicated system.
See #679 for detail.

kngwyu:
> Would it be possible to combine Py and PyClassShell so that the user can write Py<T> instead of PyClassShell<T> and Py could be read as "managed by Python"?

Unfortunately not.
As I explained in #679, `Py<T>` is a reference type that survives over multiple GILs.
They have completely different roles.

> I still have a feeling that if we can do so, we should aim to keep PyClassShell<T> mostly internal, and let users just use &T / &mut T in most places.

Before `PyRef` was introduced, we've done so and we had *really dangerous* `&T -> &PyAny` function, which causes segfault if we used to `&T` on Rust heap/stack.
To fix this, I implement `PyRef` and `PyRefMut`... and now their function ended.

tests/test_text_signature.rs
```diff
@@ -144,7 +144,7 @@ fn test_methods() {
             let _ = a;
         }
         #[text_signature = "($self, b)"]
-        fn pyself_method(_this: PyRef<Self>, b: i32) {
+        fn pyself_method(_this: &PyClassShell<Self>, b: i32) {
```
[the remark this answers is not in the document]
kngwyu: We can do zero cost conversion between `&PyClassShell` and `PyObject`/`PyAny` or so.

src/instance.rs
kngwyu: We cannot implement it for `T: PyClass` because of reflective impl block https://doc.rust-lang.org/std/convert/trait.TryFrom.html#generic-implementations

src/instance.rs
kngwyu: This `py` is necessary since `Py` can exist even after we release GIL.

src/instance.rs
kngwyu: Yeah, we should fix this.

guide/src/class.md
```diff
@@ -15,57 +15,111 @@ struct MyClass {
 }
 ```
 
-The above example generates implementations for `PyTypeInfo` and `PyTypeObject` for `MyClass`.
+The above example generates implementations for `PyTypeInfo`, `PyTypeObject`
+and `PyClass` for `MyClass`.
 
-## Get Python objects from `pyclass`
+Specifically, the following implementation is generated.
```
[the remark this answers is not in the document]
kngwyu: So what do you think is a good way to tell users *what it does*?

guide/src/class.md
kngwyu: Hmm, 'Copies the Rust struct into the Python object' is not correct for pyclass... :thinking: 
But I try to enhance it.

guide/src/class.md
```diff
@@ -15,57 +15,111 @@ struct MyClass {
...
-You can get an instance of `PyRef` by `PyRef::new`, which does 3 things:
-1. Allocates a Python object in the Python heap
-2. Copies the Rust struct into the Python object
-3. Returns a reference to it
 
-You can use `PyRef` just like `&T`, because it implements `Deref<Target=T>`.
+## Get Python objects from `pyclass`
+You sometimes need to convert your pyclass into Python object in Rust code(e.g., for testing it).
+
+You can use `PyClassShell<T>` or `Py<T>` for this purpose.
```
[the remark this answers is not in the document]
kngwyu: Yeah, but... they are really complicated in nature :sweat:

kngwyu:
I think this PR is almost complete... but does anyone have a better name other than `PyClassShell` or `PyHeap`?

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+        })
+    }
+    unsafe fn from_borrowed_ptr_or_opt(py: Python<'p>, ptr: *mut ffi::PyObject) -> Option<Self> {
+        NonNull::new(ptr).map(|p| {
+            &mut *(gil::register_borrowed(py, p).as_ptr() as *const PyClassShell<T> as *mut _)
+        })
+    }
+}
+
+/// A speciall initializer for `PyClassShell<T>`, which enables `super().__init__`
```
pganssle:
```suggestion
/// A special initializer for `PyClassShell<T>`, which enables `super().__init__`
```

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+        <T::BaseType as PyTypeInfo>::ConcreteLayout:
+            crate::type_object::PyObjectSizedLayout<T::BaseType>,
+    {
+        let shell = PyClassShell::new(py)?;
+        self.init_class(&mut *shell)?;
+        Ok(shell)
+    }
+}
+
+/// Represets that we can convert the type to `PyClassInitializer`.
```
pganssle:
```suggestion
/// Represents that we can convert the type to `PyClassInitializer`.
```

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+///
+/// It is mainly used in our proc-macro code.
```
pganssle:
```suggestion
/// This is mainly used in our proc-macro code.
```

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+    }
+}
+
+impl<T: PyClass> IntoInitializer<T> for PyResult<PyClassInitializer<T>> {
+    fn into_initializer(self) -> PyResult<PyClassInitializer<T>> {
+        self
+    }
+}
+
+/// Register new type in python object system.
```
pganssle:
```suggestion
/// Register new type in python's object system.
```

Alternatively you can use "in **the** python object system".

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+///     }
+/// }
+/// let gil = Python::acquire_gil();
+/// let py = gil.python();
+/// let _basetype = py.get_type::<BaseClass>();
+/// let typeobj = py.get_type::<SubClass>();
+/// let inst = typeobj.call((), None).unwrap();
+/// py_run!(py, inst, "assert inst.basename == 'base'; assert inst.subname == 'sub'");
+/// ```
+pub struct PyClassInitializer<T: PyTypeInfo> {
```
pganssle: I see a decent amount of explanation of how to use this, but I'm still not entirely clear on why it is necessary for it to be a wrapper type. Is it possible for this to be a trait instead where the user is expected to implement a function to initialize the base class(es)?

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+        let base_tp_ptr = <T::BaseType as PyTypeInfo>::type_object();
+        if let Some(base_new) = (*base_tp_ptr).tp_new {
+            return base_new(tp_ptr, ptr::null_mut(), ptr::null_mut());
+        }
+    }
+    let alloc = (*tp_ptr).tp_alloc.unwrap_or(ffi::PyType_GenericAlloc);
+    alloc(tp_ptr, 0)
+}
+
+/// A trait that enables custome alloc/dealloc implementations for pyclasses.
```
pganssle:
```suggestion
/// A trait that enables custom alloc/dealloc implementations for pyclasses.
```

guide/src/class.md
```diff
@@ -15,57 +15,110 @@ struct MyClass {
...
-You can get an instance of `PyRef` by `PyRef::new`, which does 3 things:
-1. Allocates a Python object in the Python heap
-2. Copies the Rust struct into the Python object
-3. Returns a reference to it
 
-You can use `PyRef` just like `&T`, because it implements `Deref<Target=T>`.
+## Get Python objects from `pyclass`
+You sometimes need to convert your `pyclass` into a Python object in Rust code (e.g., for testing it).
+
+For getting *GIL-bounded*(i.e., with `'py` lifetime) references of `pyclass`,
```
Alexander-N:
```suggestion
For getting *GIL-bounded* (i.e., with `'py` lifetime) references of `pyclass`,
```

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+    // so we need to call DECREF here:
+    if ffi::PyType_HasFeature(ty, ffi::Py_TPFLAGS_HEAPTYPE) != 0 {
+        ffi::Py_DECREF(ty as *mut ffi::PyObject);
+    }
+}
+
+/// If `PyClass` is implemented for `T`, then we can use `T` in the Python world,
+/// via `PyClassShell`.
+///
+/// `#[pyclass]` attribute automatically implement this trait for your Rust struct,
```
pganssle:
```suggestion
/// The `#[pyclass]` attribute automatically implements this trait for your Rust struct,
```

guide/src/class.md
```diff
@@ -15,57 +15,110 @@ struct MyClass {
...
+you can use `PyClassShell<T>`.
+Or you can use `Py<T>` directly, for *not-GIL-bounded* references.
+
+### `PyClassShell`
+`PyClassShell` represents the actual layout of `pyclass` on the Python heap.
+
+If you want to instantiate `pyclass` in Python and get the the reference,
```
Alexander-N:
```suggestion
If you want to instantiate `pyclass` in Python and get the reference,
```

src/pyclass.rs
```diff
@@ -0,0 +1,642 @@
...
+                e.print(py);
+                panic!("An error occurred while initializing class {}", Self::NAME)
+            });
+        }
+
+        unsafe { NonNull::new_unchecked(type_object) }
+    }
+}
+
+/// `PyClassShell` represents the concrete layout of `PyClass` in the Python heap.
```
pganssle: Is `PyClass` another class, or are you saying that this represents the concrete layout of *a Python class* on the Python heap? If the latter, I would reword it like this:

``` `PyClassShell` represents the concrete layout of a Python class on the Python heap```

(Maybe you can exclude the "Python" before `class`).

File: src/pyclass.rs [hunk 24]
```diff
@@ -0,0 +1,642 @@
...
+///
+/// You can use it for testing your `#[pyclass]` correctly works.
```
pganssle:
```suggestion
/// You can use it to test that your `#[pyclass]` correctly works.
```

programmerjake:
One thing to watch out for: what happens if some user creates a `PyClassShell` type on the Rust heap/stack and tries to convert `&PyClassShell<T>` to `Py<T>` or `PyObject`? I think that may lead to UB.

tests/test_text_signature.rs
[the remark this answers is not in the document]
davidhewitt: Isn't this also true for `Py<Self>` ?

guide/src/class.md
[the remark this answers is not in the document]
davidhewitt: I think if we are exposing two different choices to pyo3 users, we should definitely make a clear description to them what the benefits of each is.

guide/src/class.md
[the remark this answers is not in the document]
davidhewitt: I do also like this. It is helpful for users to see it. We will just need to be careful to keep it up to date in future.

src/instance.rs
davidhewitt: This sounds reasonable but actually now I search the codebase, in both of the implementations of `AsPyRef` we don't actually use the `py` argument. I propose we remove the trait completely, as it's essentially equivalent to `AsRef`.

The two implementations I see are:

https://github.com/kngwyu/pyo3/blob/pyclass-new-layout/src/instance.rs#L127

https://github.com/kngwyu/pyo3/blob/pyclass-new-layout/src/object.rs#L253

davidhewitt:
> Before PyRef was introduced, we've done so and we had really dangerous &T -> &PyAny function, which causes segfault if we used it for &T on Rust heap/stack.

I can see why that indeed would be super dangeous. 😮 

I want to back up Alex's point that having *both* `Py<T>` and `PyClassShell<T>` in the public API seems like two solutions to the same thing from a user perspective. I understand the need for both for our implementation: they serve very different functions internally.

However, to the user, they both represent "a value on the Python heap", so I think that we should aim for having `Py<T>` as public and `PyClassShell<T>` as internal if at all possible.

guide/src/class.md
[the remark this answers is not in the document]
kngwyu: Hmm... :thinking: 
I think it's a rough skecth and just 'compile pass' is suffcient.

kngwyu:
@crepererum 
Thank you for pointing it out. It looks it was changed by my file system was crashed.

kngwyu:
@pganssle 
Thank you for the lots of reviews!

> I do really like the new type signatures for #[pyclass], but could you comment on how this will affect performance? Is it expected to add overhead because of an additional layer of abstraction? Or possibly remove overhead because it understands the Python layout better?

It is not for performance.
Current `OFFSET` based design has some problems
- Corner cases that cause SIGSEGV(e.g., `#[pyclass(extends=PyDict)]`)
- We have to use specialization to distinguish Rust-native type and Python-native type
So this PR mainly resolves these 2 problems, but it removes an `if OFFSET == 0` branch from conversion codes.

src/pyclass.rs
kngwyu: So, if we have something like `PyClassNew` and `PyClassNew::new`, there appears some problems:
- What kind of arguments should it take? We allow `#[new]` to take *any* arguments, but trait cannot allow that.
- Maybe we need specialization.

kngwyu:
@programmerjake 
> One thing to watch out for: what happens if some user creates a PyClassShell type on the Rust heap/stack and tries to convert &PyClassShell<T> to Py<T> or PyObject? 

@davidhewitt 
I think we cannot address this problem in this PR.
Current `Py` system is not sufficient for replacing that.

guide/src/class.md
```diff
@@ -139,24 +186,29 @@ Rules for the `new` method:
...
-* The first parameter is the raw object and the custom `new` method must initialize the object
-  with an instance of the struct using the `init` method. The type of the object may be the type object of
-  a derived class declared in Python.
-* The first parameter must have type `&PyRawObject`.
+* All parameters are from Python.
+* It can return one of these types:
+  - `T`
+  - `PyResult<T>`
+  - `PyClassInitializer<T>`
+  - `PyResult<PyClassInitializer<T>>`
```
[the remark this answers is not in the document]
kngwyu: Thank you :+1: 
This advice made the explanation much clearer.

davidhewitt:
> Nice idea, but,... 🤔, I don't think we cannot do so now.
The current Py system is too limited for replacing that.

Ok. Let's perhaps try to think if we can do this as part of the design thinking for #697 ?

pganssle:
> It is not for performance.
> Current `OFFSET` based design has some problems
> 
> * Corner cases that cause SIGSEGV(e.g., `#[pyclass(extends=PyDict)]`)
> * We have to use specialization to distinguish Rust-native type and Python-native type
>   So this PR mainly resolves these 2 problems, but it removes an `if OFFSET == 0` branch from conversion codes.

It would still be good to know how it affects performance. Does this add overhead, and if so, how much? One thing to consider is that the more overhead PyO3 adds, the less likely it is to be adopted (at least from the Python side), since many people are expicitly looking for a safer way to get speed in their Python files. If PyO3 adds a ton of overhead over C or Cython, people will simply say, "Oh well it would be nice if we could have safety and speed, but I guess we can't."

If this sort of thing is super expensive but necessary to provide safety guarantees mainly in strange edge cases, it may make sense to have a middle-of-the-road "fast but possibly unsafe" interface for people who really care about speed and a "slow but safe" version for people whose bottlenecks are elsewhere.

src/pyclass.rs
pganssle: @kngwyu I will admit I don't have deep Rust expertise, so feel free to ignore me if this is just noise, but I still don't really understand the issue. I am not entirely sure what problem this class is solving, but it seems like the idea is that you need to make sure that all the Rust-based superclasses are initialized when the subclass is initialized.

I would think that you could have a trait like `PySuperclassInitialize<T>`, and you implement `PySuperclassInitialize<BaseClass> for SubClass` for each value for `BaseClass`. The procedural macro `#[pyclass]` would inject calls to the functions that the trait provides, which would enforce at compile-time that the initializers were implemented.

In the short term, you could probably make `PyClassInitializer` an implementation detail (with an eye towards eliminating it entirely if that simplifies things), with a trait like this:

```rust
trait PySuperClassInitializer<T> {
    fn make_super_value(&self) -> T;
}
```

And for a `SubClass` that extends `BaseClass`, you'd have the `#[new]` macro inject this code into the initializer:

```rust
// Here `val` is the return value of the defined __new__ function
let mut init = PyClassInitializer::from_value(val);
init.get_super().init(val.make_super_value::<BaseClass>());
init
```

So then the `SubClass` code looks like this:

```rust
#[pyclass(extends=BaseClass)]
struct SubClass {
    #[pyo3(get)]
    subname: &'static str,
}

#[pymethods]
impl SubClass {
    #[new]
    fn new() -> Self {
        SubClass { subname: "sub" }
    }
}

impl PySuperclassInitializer<BaseClass> for SubClass {
    fn make_super_value(&self) -> BaseClass {
        BaseClass { basename: "base" }
    }
}
```

I don't really love the names for that API, but you get the gist - if I understand this problem correctly, you *should* be able to accomplish this just with proc macros and traits.

kngwyu:
>  Does this add overhead,

No.
This abstraction is zero-cost and does more things at compile time.
E.g., it removes `if OFFSET == 0` branches from all conversion codes.

src/pyclass.rs
kngwyu: > but it seems like the idea is that you need to make sure that all the Rust-based superclasses are initialized when the subclass is initialized

Yes. Exactly.

> fn make_super_value(&self) -> T;

Sorry for repeating the same argument, but the problem is trait methods cannot have arbitrary arguments, unlike Python's `__init__`.
What I want to do here is pass arguments to `super().__init__`, like
```rust
impl SubClass {
    #[new] 
    fn new(basename: String, name: String, id: usize) -> PyClassInitializer<Self> {
        let mut init = PyClassInitializer::from_value(Self::new_impl(name));
        init.get_super().init(BaseClass::new(basename, id));
        init
    }
}
```

So the use of trait does not fit well here.

kngwyu: One of the possible solutions is to **change the type after `super().__init__` called**.
Like,
```rust
let init: PyClassInitalizer<Self, NoInit> = PyClassInitializer::from_value(Self::new_impl(...);
let init: PyClassInitiazlier<Self, BaseClass> = init.super_init(BaseClass::new(...));
init
```

But it's difficult for nested types...

pganssle: > Sorry for repeating the same argument, but the problem is trait methods cannot have arbitrary arguments, unlike Python's `__init__`.

So to be clear, what I was suggesting is that you would call the `make_super_value` with *no arguments*. The "arguments" to the constructor would be in the trait implementation, so in your example you'd do:

```rust
impl PySuperclassInitializer<BaseClass> for SubClass {
    fn make_super_class(&self) -> BaseClass {
        BaseClass::new(basename, id)
    }
}
```

That makes it so that you can construct the base class however you want, and I think it avoids the end user constructing a `PyClassInitializer` *at all*. Maybe there are other scenarios you have in mind where it's not as straightforward as "initialize the subclass, then initialize the subclass" where it wouldn't be as straightforward.

If what I'm saying doesn't make sense, don't worry about it, it's not terribly important. If you don't think it can be done or it's a good idea, we can merge it like this and I'll try to come up with a proof-of-concept of the trait-based mechanism to eliminate `PyClassInitializer` from the public interface (and I'd say there's a 70-80% chance that in doing so I would convince myself that you were right to use the `PyClassInitializer` approach in the first place).

kngwyu:
## Important change for `PyClassInitializer`
I pushed the new `PyClassInitializer` that can prevent uninitialized base class at compile time.
Though this adds another boiler plate to `PyTypeInfo`, the implementation is quite simple and clean.
The interface employs `std::process::Command` like initializer pattern:
```rust
#[pymethods]
impl SubSubClass {
    #[new]
    fn new() -> PyClassInitializer<Self> {
        let base_init = PyClassInitializer::from_value(BaseClass{basename: "base"});
        base_init.add_subclass(SubClass{subname: "sub"})
                 .add_subclass(SubSubClass{subsubname: "subsub"})
    }
}
```
For simple inheritance where `B` inherits `PyAny` and `S` inherits `B`, we allow `(S, B)` as return type of `#[new]`.

```rust
#[pymethods]
impl SubClass {
    #[new]
    fn new() -> (Self, BaseClass) {
        (SubClass { val2: 5 }, BaseClass { val1: 10 })
    }
}
```
cc: @pganssle

src/pyclass.rs
kngwyu: @pganssle 
I employed another pattern to prevent un-initialized base classes.
Please check it!

davidhewitt:
I very much like the new `PyClassInitializer`, though have concerns about `IntoInitializer` as implemented. It's got a strange coupling to `PyResult` which I don't think we need.

I opened a PR against your fork to propose a simpler API for `IntoInitializer`. Further elaboration on my thoughts are also there. https://github.com/kngwyu/pyo3/pull/1

davidhewitt:
Further to my thoughts about `IntoInitializer` and `PyResult`, the new `(S, B)` return type of `#[new]` is also creeping into the API for `Py::new`:

    #[pyclass]
    struct Base { }

    #[pyclass(extends=Base)]
    struct Sub { }

    let sub = pyo3::Py::new(py, (Sub {}, Base {})).unwrap()

This doesn't seem desirable to me. What do others think?

(If we merge my suggested PR, then it would be as simple as changing `Py::new` to take `impl Into<PyClassInitializer<T>>` instead of `impl IntoInitializer<T>`.)

kngwyu:
Following the line of @davidhewitt's simplification, I replaced `IntoInitializer<T>` with `Into<PyClassInitializer<T>>`.

src/prelude.rs
```diff
@@ -12,14 +12,13 @@
...
 pub use crate::gil::GILGuard;
-pub use crate::instance::{AsPyRef, Py, PyRef, PyRefMut};
+pub use crate::instance::{AsPyRef, Py};
 pub use crate::object::PyObject;
 pub use crate::objectprotocol::ObjectProtocol;
 pub use crate::python::Python;
 pub use crate::{FromPy, FromPyObject, IntoPy, IntoPyPointer, PyTryFrom, PyTryInto, ToPyObject};
 // This is only part of the prelude because we need it for the pymodule function
+pub use crate::pyclass_init::PyClassInitializer;
 pub use crate::types::PyModule;
```
davidhewitt: This new `pub use crate::pyclass_init::PyClassInitializer;` has split up a comment from the line beneath it.

src/pyclass_init.rs
```diff
@@ -0,0 +1,122 @@
...
+pub struct PyClassInitializer<T: PyClass> {
+    init: T,
+    super_init: Box<<T::BaseType as PyTypeInfo>::Initializer>,
+}
+
+impl<T: PyClass> PyClassInitializer<T> {
+    pub fn new(init: T, super_init: <T::BaseType as PyTypeInfo>::Initializer) -> Self {
+        Self {
+            init,
+            super_init: Box::new(super_init),
```
davidhewitt: Is it necessary to box `super_init`? This seems to be an allocation we may not need.

src/prelude.rs
kngwyu: :+1:

src/pyclass_init.rs
kngwyu: You're right, thank you.

Issue #682, opened 2019-12-04
PyErr does not implement std::error::Error


nagisa:
PyErr is a very odd structure. It pretends to emulate the 3-tuple returned by `PyErr_Fetch`, but then its `pvalue` is a 3-variant that also allows lazy construction of exception objects.

As thus it fails to implement a couple of traits typically expected of the error type (`std::error::Error`, `std::fmt::Display`, `Sync` and `Send`) and does not integrate very well with the error handling story in Python.

There are a couple of ways to go forward here, but one of the simpler things that could be done I think is:

1. Add a method along the lines of `fn(Python, PyErr) -> Py<BaseException>`;
    * I don’t believe this can be done easily right now without unsafe code.
2. Implement the typical error traits for `Py<ExceptionT>` types.
    * Specifically for `Display` and friends I think it is fine to just do a `Python::acquire_gil()` inside if that is necessary to format the data.

Willing to contribute the necessary code to implement this. Please let me know what you think about this. Perhaps other approaches seem better?

kngwyu:
Thank you for the proposal.
Error handling is indeed a major pain around PyO3 and thus should be enhanced.

> Perhaps other approaches seem better?

I'm not sure they are better, but possible alternatives are
3. Get error string when fetching error (difficult, but I don't think it's impossible)
4. Hold some 'intermidiate' error message (see #652)

> Implement the typical error traits for Py<ExceptionT> types. 

PyErr has already the same memory-management semantics as `Py<T>` so I don't think `Py<ExceptionT>` is appropriate.

nagisa:
> PyErr has already the same memory-management semantics as Py<T> so I don't think Py<ExceptionT> is appropriate.

To make `PyErr` `Send + Sync` it would require a breaking change to the lazy-initialized variants of the `pvalue` field. If those variants did not exist, this problem would be significantly easier and consist of basically making a `Display` implementation, something I already have.

FWIW I feel it is fine to have multiple ways to do the same thing in pyo3. It is a necessity, in fact, due to the fact that some conversions are not easily possible (such as going from `PyErr` to `BaseException`…)
---

Unless I misunderstand them, options 3&4 are insufficient as they lose information: the type, traceback and causation chain (`raise exc from otherexc`) are not preserved. `BaseException` and derived objects store all that information in the object (`type()`, `__traceback__`, etc.)

kngwyu:
So how about separating Rust-originated errors into another struct?
Then we would have(roughly)
```rust
[...]

PR #686, opened 2019-12-09
Proto #1: Exception instances as Py<BaseException>

nagisa:
This prototype implements use of `Py<BaseException>` as the instance to
use for exception instances. These instances integrate reasonably well
with the Rust’s standard error mechanisms by implementing the `Display`
and `Error` traits. These error types can also be stored into e.g.
`failure::Fail`s or other error types as a cause of some greater error.

There are a couple of sketchy points in this PR most notable of which is
the implementation of `std::error::Error::source`. See the included FIXME.

This is also somewhat of an experiment around the ownership-based `Py<>`
that I have outlined in https://github.com/PyO3/pyo3/issues/679 and the issues
implementing `source` appear to stem from exactly the fact that Python’s GIL
and Rust’s ownership are fairly incompatible.

FWIW: I think this might be revealing a soundness issue in our current setup too.

### DIFF
--- a/src/exceptions.rs
+++ b/src/exceptions.rs
@@ -8,6 +8,7 @@ use crate::type_object::PyTypeObject;
 use crate::types::{PyAny, PyTuple};
 use crate::Python;
 use crate::{AsPyPointer, ToPyObject};
+use crate::{AsPyRef, Py, PyDowncastError, PyTryFrom};
 use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::{self, ops};
@@ -208,11 +209,13 @@ macro_rules! impl_native_exception (
                 PyErr::new::<$name, _>(())
             }
         }
+
         impl<T> std::convert::Into<$crate::PyResult<T>> for $name {
             fn into(self) -> $crate::PyResult<T> {
                 PyErr::new::<$name, _>(()).into()
             }
         }
+
         impl $name {
             pub fn py_err<V: ToPyObject + 'static>(args: V) -> PyErr {
                 PyErr::new::<$name, V>(args)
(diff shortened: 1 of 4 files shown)


nagisa:
cc @kngwyu

src/exceptions.rs
```diff
@@ -340,6 +403,72 @@ impl StopIteration {
...
+        } else {
+            write!(f, ": <exception str() failed>")
+        }
+    }
+}
+
+impl std::error::Error for BaseException {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        unsafe {
+            // Returns either `None` or an instance of an exception.
```
nagisa: This needs to obtain the GIL.

src/exceptions.rs
```diff
@@ -221,11 +224,71 @@ macro_rules! impl_native_exception (
...
             }
         }
+
         unsafe impl PyTypeObject for $name {
             fn init_type() -> std::ptr::NonNull<$crate::ffi::PyTypeObject> {
                 unsafe { std::ptr::NonNull::new_unchecked(ffi::$exc_name as *mut _) }
             }
         }
+
+        impl<'v> PyTryFrom<'v> for $name {
```
kngwyu: Upcasting to `PyAny` is indeed a missing part :+1:

src/exceptions.rs
```diff
@@ -340,6 +403,72 @@ impl StopIteration {
     }
 }
 
+impl std::fmt::Debug for BaseException {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        // Sneaky: we don’t really need a GIL lock here as nothing should be able to just mutate
+        // the "type" of an object, right? RIGHT???
```
kngwyu: Sorry I also am not sure :sweat: 
We need tests.

src/exceptions.rs
```diff
@@ -340,6 +403,72 @@ impl StopIteration {
...
+        //
+        // let gil = Python::acquire_gil();
+        // let _py = gil.python();
+        let py_type_name = unsafe { CStr::from_ptr((*(*self.as_ptr()).ob_type).tp_name) };
+        let type_name = py_type_name.to_string_lossy();
+        f.debug_struct(&*type_name)
+            // TODO: print out actual fields!
```
kngwyu: We can access actual field only by type name?
Or we need to get the global `PyExc_~`s here?

src/exceptions.rs
```diff
@@ -340,6 +403,72 @@ impl StopIteration {
...
+                // FIXME: PyException_GetCause returns a new reference to the cause object.
+                //
+                // While we know that `self` is "immutable" (`&self`!) it is also true that between
+                // now and when the return value is actually read the GIL could be unlocked and
+                // then concurrent threads could modify `self` to change its `__cause__`.
+                //
+                // If this was not a possibility, we could just `DECREF` here, instead, now, we
+                // must return a `&Py<BaseException>` instead… but we cannot do that because
+                // nothing is storing such a thing anywhere and thus we cannot take a reference to
+                // that…
```
kngwyu: Sorry I feel it difficult to understand why we cannot use `Py<Exception>` here... :thinking:

kngwyu:
Thank you, this would be a big step for us.

src/exceptions.rs
nagisa: By "actual fields" I meant the fields of the `PyException_HEAD` structure component, which is defined as such:

```
#define PyException_HEAD PyObject_HEAD PyObject *dict;\
             PyObject *args; PyObject *traceback;\
             PyObject *context; PyObject *cause;\
             char suppress_context;
```

At very least values of `args`, `traceback`, `context`, `cause` and `suppress_context` are interesting, but I can imagine it being useful to also print out at least the `refcount` from `PyObject_HEAD` too.

`f.debug_struct(&*type_name)` is just a builder for debug output. See https://doc.rust-lang.org/stable/std/fmt/struct.Formatter.html#method.debug_struct. The final result should look something like

```
f.debug_struct(type_name).field("args", &self.args).field("traceback", &self.traceback)/* etc... */.finish()
```

src/exceptions.rs
```diff
@@ -340,6 +403,72 @@ impl StopIteration {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let py_type_name = unsafe { CStr::from_ptr((*(*self.as_ptr()).ob_type).tp_name) };
+        let type_name = py_type_name.to_string_lossy();
+        write!(f, "{}", type_name)?;
+        let py_self: Py<PyAny> = unsafe { Py::from_borrowed_ptr(self.as_ptr()) };
+
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        if let Ok(s) = crate::ObjectProtocol::str(&*py_self.as_ref(py)) {
+            write!(f, ": {}", &s.to_string_lossy())
```
nagisa: FIXME: if the resulting string is empty we want to not add the `: `.

src/exceptions.rs
nagisa: We cannot return a `Py<Exception>` because the trait contract requires:

```rust
fn source(&self) -> Option<&(dyn std::error::Error + 'static)>
```

The most important detail is that this method is returning a *reference* with a lifetime tied to `&self`. In other words, it requires that you return a reference to something in `Self`. Trying to return a `Py<BaseException>` would result in the typical "value does not live long enough" error, very much like the snippet below would:

```rust
fn bogus(&self) -> &str {
    &String::from("banana") // cannot compile because `String` does not live long enough...
}
```

kngwyu: I see, thanks.

davidhewitt:
@nagisa I'm curious if you plan to continue with this PR? If not, I'd be happy to take it on, rebase it on master and address the changes reviewed here.

nagisa:
@davidhewitt you’re welcome to take over this. I still think it is very worthwhile to do this but I also no longer have much drive to work on this feature as I implemented something simpler, although less elegant, directly in my code.

kngwyu:
I noticed we lack the linnk to the most relevant issue #682

davidhewitt:
I have a PR which is almost ready to open as a draft to replace this one.

I've stuck to the original idea here but gone a bit further: `&BaseException` and `Py<BaseException>` implement `std::error::Error`. 

`&BaseException` turns out to be pretty nice here - if we bound `&BaseException` by the GIL lifetime in the normal way, then the lifetime issue above related to `Error::cause` goes away. (Because the returned reference can be an owned `&BaseException` reference tied to the GIL lifetime.)

davidhewitt:
Superseded by #1024

PR #692, opened 2019-12-17
Add #[name = "foo"] attribute to #[pymethods]

### LINKED ISSUE #663: Allow overriding generated python method names
Allow code like:
```rust
#[pymethods]
impl MyType {
    // public API for Python
    #[name = "my_method"]
    #[args(arg2 = "None")]
    fn my_method_python(&self, arg1: i32, arg2: Option<i32>) {
        match arg2 {
            Some(arg2) => self.my_method_with_arg2(arg1, arg2),
            None => self.my_method(arg1),
        }
    }
}

impl MyType {
    // public API for Rust
    pub fn my_method(&self, arg1: i32) {
        unimplemented!()
    }
    pub fn my_method_with_arg2(&self, arg1: i32, arg2: i32) {
        unimplemented!()
    }
}
```

`my_method` is accessible from Python:
```python
a = MyType()
a.my_method(1)
a.my_method(1, 2)
```

This is useful to allow the Rust and Python APIs for a type to differ somewhat due to different conventions between Python and Rust.

There is already precedent for this kind of renaming: `#[pyclass(name = MyType)]`

davidhewitt:
Closes #663 

To achieve this, I added to `FnSpec`:
- a new field `name` - the name of the Rust function this maps to
- a new field `python_name` - either the optionally-set `#[name]`, or the `.unraw()` rust name

While I was at it I did some refactoring to swap some `panic!`-s to spanned errors, and a bit of cleanup.

TODO:
- [ ] Bikeshed name - `#[pyname]` perhaps?
- [x] Support `#[pyfunction]`?
- [ ] Update documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+* Support for `#[name = "foo"]` attribute in `#[pymethods]`. [#692](https://github.com/PyO3/pyo3/pull/692)
+
 ## [0.8.4]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
-* Support for `#[name = "foo"]` attribute in `#[pymethods]`. [#692](https://github.com/PyO3/pyo3/pull/692)
+* Support for `#[name = "foo"]` attribute for `#[pyfunction]` and in `#[pymethods]`. [#692](https://github.com/PyO3/pyo3/pull/692)
 
 ## [0.8.4]
 
--- a/pyo3-derive-backend/src/method.rs
+++ b/pyo3-derive-backend/src/method.rs
@@ -2,6 +2,7 @@
 
 use crate::pyfunction::Argument;
 use crate::pyfunction::PyFunctionAttr;
+use crate::utils;
 use proc_macro2::TokenStream;
 use quote::quote;
 use quote::ToTokens;
@@ -20,8 +21,8 @@ pub struct FnArg<'a> {
 
 #[derive(Clone, PartialEq, Debug)]
 pub enum FnType {
-    Getter(Option<String>),
-    Setter(Option<String>),
+    Getter,
+    Setter,
     Fn,
     FnNew,
     FnCall,
@@ -33,9 +34,14 @@ pub enum FnType {
 #[derive(Clone, PartialEq, Debug)]
 pub struct FnSpec<'a> {
     pub tp: FnType,
+    // Rust function name
+    pub name: &'a syn::Ident,
+    // Wrapped python name
+    pub python_name: Option<syn::Ident>,
     pub attrs: Vec<Argument>,
     pub args: Vec<FnArg<'a>>,
     pub output: syn::Type,
+    pub doc: syn::LitStr,
 }
 
 pub fn get_return_info(output: &syn::ReturnType) -> syn::Type {
@@ -48,11 +54,29 @@ pub fn get_return_info(output: &syn::ReturnType) -> syn::Type {
 impl<'a> FnSpec<'a> {
     /// Parser function signature and function attributes
     pub fn parse(
-        name: &'a syn::Ident,
         sig: &'a syn::Signature,
         meth_attrs: &mut Vec<syn::Attribute>,
+        allow_custom_name: bool,
     ) -> syn::Result<FnSpec<'a>> {
-        let (mut fn_type, fn_attrs) = parse_attributes(meth_attrs)?;
+        let name = &sig.ident;
+        let (mut fn_type, fn_attrs, mut python_name) =
+            parse_attributes(meth_attrs, allow_custom_name)?;
+
+        // "Tweak" getter / setter names: strip off set_ and get_ if needed
+        if let FnType::Getter | FnType::Setter = &fn_type {
+            if python_name.is_none() {
+                let prefix = match &fn_type {
+                    FnType::Getter => "get_",
+                    FnType::Setter => "set_",
+                    _ => unreachable!(),
+                };
+
+                let ident = sig.ident.to_string();
+                if ident.starts_with(prefix) {
+                    python_name = Some(syn::Ident::new(&ident[prefix.len()..], ident.span()))
+                }
+            }
+        }
 
         let mut has_self = false;
         let mut arguments = Vec::new();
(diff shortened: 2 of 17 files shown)


pyo3cls/src/lib.rs
```diff
@@ -83,8 +83,9 @@ pub fn pyfunction(attr: TokenStream, input: TokenStream) -> TokenStream {
     let mut ast = parse_macro_input!(input as syn::ItemFn);
     let args = parse_macro_input!(attr as PyFunctionAttr);
 
+    // TODO: Support #[name = "..."] attribute?
```
davidhewitt: This: should `#[pyfunction]` also support this attribute? I think it can be done with a little more refactoring.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+* Support for `#[name = "foo"]` attribute in `#[pymethods]`. [#692](https://github.com/PyO3/pyo3/pull/692)
```
davidhewitt: Bikeshed: maybe the attribute should be `#[pyname]` instead of `#[name]`?

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -6,97 +6,29 @@ use quote::quote;
...
+        FnType::Fn => impl_py_method_def(&spec, &impl_wrap(cls, &spec, true)),
+        FnType::PySelf(ref self_ty) => {
+            impl_py_method_def(&spec, &impl_wrap_pyslf(cls, &spec, self_ty, true))
         }
+        FnType::FnNew => impl_py_method_def_new(&spec, &impl_wrap_new(cls, &spec)),
+        FnType::FnCall => impl_py_method_def_call(&spec, &impl_wrap(cls, &spec, false)),
+        FnType::FnClass => impl_py_method_def_class(&spec, &impl_wrap_class(cls, &spec)),
+        FnType::FnStatic => impl_py_method_def_static(&spec, &impl_wrap_static(cls, &spec)),
+        FnType::Getter => impl_py_getter_def(&spec, &impl_wrap_getter(cls, &spec)?),
+        FnType::Setter => impl_py_setter_def(&spec, &impl_wrap_setter(cls, &spec)?),
```
davidhewitt: Now that these function signatures are so similar (and their implementations very similar too) I wonder whether with a bit of refactoring this can be simplified. Perhaps in a separate PR.

pyo3cls/src/lib.rs
programmerjake: I think supporting `#[pyfunction]` would be a good idea!

CHANGELOG.md
programmerjake: `#[pyname]` could work, on the other hand, we already have `#[text_signature]` and not `#[pytext_signature]`, however `#[name]` is much more likely to conflict with something in the future due to being a much more common word than `#[text_signature]`

tests/ui/invalid_pymethod_names.stderr
```diff
@@ -0,0 +1,29 @@
...
+18 |     #[name = "bar"]
+   |       ^^^^
+
+error: name can not be specified with #[new]
+  --> $DIR/invalid_pymethod_names.rs:24:7
+   |
+24 |     #[name = "makenew"]
+   |       ^^^^
+
+error: cannot find attribute `name` in this scope
```
programmerjake: `#[name]` attributes should not make it out of the proc-macro even in error conditions. rustc changing error messages for unknown attributes is what's causing the compile failure.

pyo3-derive-backend/src/method.rs
```diff
@@ -279,10 +344,15 @@ pub fn check_arg_ty_and_optional<'a>(
     }
 }
 
-fn parse_attributes(attrs: &mut Vec<syn::Attribute>) -> syn::Result<(FnType, Vec<Argument>)> {
+fn parse_attributes(
+    attrs: &mut Vec<syn::Attribute>,
+    allow_custom_name: bool,
+) -> syn::Result<(FnType, Vec<Argument>, Option<syn::Ident>)> {
```
programmerjake: The return type should probably be replaced with a named `struct`, something like:

```rust
struct ParsedAttributes {
    fn_type: FnType,
    arguments: Vec<Argument>,
    python_name: Option<syn::Ident>,
}

fn parse_attributes(
    attrs: &mut Vec<syn::Attribute>,
    allow_custom_name: bool,
) -> syn::Result<ParsedAttributes> {
```

tests/ui/invalid_pymethod_names.stderr
davidhewitt: :+1: I was thinking I need to refactor this as part of supporting `#[pyfunction]`. The compile failure gives me a solid reason to do it!

pyo3-derive-backend/src/pyfunction.rs
```diff
@@ -193,6 +197,53 @@ impl PyFunctionAttr {
...
+            _ => {
+                return Err(syn::Error::new(
+                    span,
+                    "Expected string literal for #[name] argument",
+                ))
+            }
+        };
+    }
+
+    Ok(name)
```
kngwyu: I recommend
```rust
    match *name_attrs {
        [] => Ok(None),
        [(lit, span)] => {
            ...
        }
        _ => Err(...)
    }
```

kngwyu:
Thank you!
Though I have a concern that users really need this feature...(isn't it confusing to give different names to Python API?), this PR is generally well written and refactorings are good :+1: 
But, if possible, I want you to avoid refactoring codes unrelated to the PR.
It makes code review more difficult and introduces unnecessary merge conflicts.

davidhewitt:
> Though I have a concern about this feature itself...(isn't it confusing to give different names to Python API?)

I agree most users won't need this. There will be edge cases like the example originally supplied on #663.

Note that the `#[pyfn]` macro actually *requires* a python name as the second argument. So adding this `#[name]` attribute brings the other macros to feature parity without forcing all users to specify names.

It's probably too breaking but I now actually think it would be nice to remove the required name argument from `#[pyfn]` and replace it with this optional attribute.

pyo3-derive-backend/src/pyfunction.rs
davidhewitt: Thanks, that's much nicer

davidhewitt:
I've pushed changes to all the things you suggested.

> But, if possible, I want you to avoid refactoring codes unrelated to the PR.
> It makes code review more difficult and introduces unnecessary merge conflicts.

Sorry. I'll do refactorings as separate PRs next time 😄

programmerjake:
For `#[getter]` and `#[setter]`, maybe it would be worthwhile to have the error message be something like: ```"#[name] can't be used with `#[getter]`, to override the Python name, use `#[getter(python_name)]`"```

alternatively, for consistency, just accept:

```rust
#[getter]
#[name = "python_name"]
fn rust_name(&self) -> i32 {
    0
}
#[getter(python_name2)]
fn rust_name2(&self) -> i32 {
    0
}
```

but not:

```rust
#[getter(other_python_name)]
#[name = "python_name"]
fn rust_name(&self) -> i32 {
    0
}
```

kngwyu:
Thank you!

> alternatively, for consistency, just accept:

I often feel this kind of design decision difficult.
Let's decide if it's good or not after using the current interface for a while.
If some users complain about this, then we should change the interface.

PR #693, opened 2019-12-18
Support `#[pyfunction]` inside `#[pymodule]`

davidhewitt:
It struck me as strange that `#[pyfn]` exists at all when it should be possible to use `#[pyfunction]` inside pymodule.

This PR adds exactly that. Maybe this is a road to deprecating / removing `#[pyfn]` so as to make the overall API simpler? (Just one `#[pyfunction]` attribute instead of two.)

TODO:
- [ ] Update documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+
+* Support for `#[pyfunction]` inside `#[pymodule]`. [#693](https://github.com/PyO3/pyo3/pull/693)
+
 ## [0.8.4]
 
 ### Added
--- a/pyo3-derive-backend/src/module.rs
+++ b/pyo3-derive-backend/src/module.rs
@@ -28,8 +28,10 @@ pub fn py_init(fnname: &Ident, name: &Ident, doc: syn::Lit) -> TokenStream {
 }
 
 /// Finds and takes care of the #[pyfn(...)] in `#[pymodule]`
-pub fn process_functions_in_module(func: &mut syn::ItemFn) {
+pub fn process_functions_in_module(func: &mut syn::ItemFn) -> syn::Result<()> {
     let mut stmts: Vec<syn::Stmt> = Vec::new();
+
+    let module_name = get_module_name(&func.sig)?;
 
     for stmt in func.block.stmts.iter_mut() {
         if let syn::Stmt::Item(syn::Item::Fn(ref mut func)) = stmt {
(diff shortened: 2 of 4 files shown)


programmerjake:
one reason to have separate `#[pyfn]` and `#[pyfunction]` attributes is for when you want a function that's not automatically added to a Python module.

kngwyu:
```rust

#[pymodule]
fn pyfunction_module(_py: Python, module: &PyModule) -> PyResult<()> {
    #[pyfunction]
    fn foobar() -> usize {
        101
    }

    Ok(())
}
```
Hmm... :thinking: 
It's confusing for me that `pyfunction` automatically adds itself to `module`.
I think we need *alternative* syntax if we want to do this kind of automatic module declaration.
E.g.,
```rust
#[pymodule] 
mod pyfunction_module {
    #[pyfunction]
    fn foobar() -> usize {
        101
    }
}
```

davidhewitt:
> one reason to have separate #[pyfn] and #[pyfunction] attributes is for when you want a function that's not automatically added to a Python module.

Is there any case where you would write a `#[pyfunction]` inside a `#[pymodule]` but not want it automatically added to a module?

This PR doesn't change the fact that `#[pyfunction]`s outside of a `#[pymodule]` still need to be manually added.

davidhewitt:
> I think we need alternative syntax if we want to do this kind of automatic module declaration.

I can add `mod` syntax if that's preferable.

However, my original motivation is that we don't seem to need two differently-named attributes `#[pyfn]` and `#[pyfunction]`.

davidhewitt:
@kngwyu I see what you mean by automatic adding being confusing. I realise now this PR is probably a breaking change, as this is already valid code, but ***does not*** add `foobar` to the module:

    #[pymodule]
    fn pyfunction_module(_py: Python, module: &PyModule) -> PyResult<()> {
        #[pyfunction]
        fn foobar() -> usize {
            101
        }

        Ok(())
    }

I think the solution is instead to add an attribute `#[add_to_module(module)]`. I'll rework this PR later today.

kngwyu:
> I see what you mean by automatic adding being confusing. I realise now this PR is probably a breaking change, as this is already valid code, but does not add foobar to the module:

I mean that **few users expect that the generated code uses unrelated local variable `module`**.
Not about the breaking change.

> Does this seem better? 

Hmm... :thinking:
I'm sorry but I feel diffult to understand why it's better.

davidhewitt:
Ok 👍 

Let's pull back on this idea for now. I still think it would be potentially worth having just `#[pyfunction]` so that users don't have to learn two different attribute syntaxes for almost exactly the same thing. I'll leave that in #694 .

Issue #698, opened 2019-12-23
Ensure all APIs have safety notes `#![deny(clippy::missing_safety_doc)]`


kngwyu:
We have many unsafe APIs but they have poorly documented with their *safe* precondition, which makes clippy unhappy.
![image](https://user-images.githubusercontent.com/16046705/71343740-0bbcac80-25a4-11ea-8e9b-95f73d6ba3f8.png)

I don't think we need to address all those warnings, but maybe it is variable to add `#Safety` section to common unsafe APIs!

davidhewitt:
It would be really cool to try and achieve this in `0.13`.

mejrs:
afaik, there are just two of these left:

 - https://github.com/PyO3/pyo3/blob/main/src/pyclass_slots.rs#L17 I'm guessing the safety doc for that is just "pointer must point to a valid PyObject?
- https://github.com/PyO3/pyo3/blob/main/src/type_object.rs#L21 it looks like it's unsafe to implement (which it already is, it's in an unsafe trait), not unsafe  to call - this probably doesn't need to be an unsafe function?

davidhewitt:
Oh nice, I hadn't noticed we were so close!

For 1. I was thinking of tidying up the `pyclass_slots` stuff so maybe I can sort that at the same time.

2. is removed in #1657

PR #703, opened 2019-12-28
Ignore missing `# Safety` section in ffi module

Alexander-N:
I'm all for documenting unsafe functions, but I would suggest excluding the ffi module.

#698


kngwyu:
Thank you!

PR #705, opened 2019-12-28
Trigger windows tests on pull requests

Alexander-N:
Any reason why the github actions workflow should not be triggered by a pull request?


kngwyu:
Thank you!

> Any reason why the github actions workflow should not be triggered by a pull request?

No, just I didn't know *how to use it* well :roll_eyes:

PR #708, opened 2020-01-01
Fixes PyObjectDelAttrProtocol signature

lochbrunner:
This PR fixes the bug when implementing `__delattr__` for in `PyObjectProtocol`, you got the following error message:

> error[E0437]: type `Success` is not a member of trait `pyo3::class::basic::PyObjectDelAttrProtocol`


kngwyu:
Thank you!

PR #714, opened 2020-01-07
guide page on classes; correction and expansion

paddywwoof:
In the guide there seems to be no explanation or example of using PyModule.add_class().

The explanation and example of using `#[args(...)]` is very thin (and missing some `"`s) the use of arg and args as parameters(?) method arguments(?) is quite confusing. Also the example method doesn't do anything so it's almost impossible to discern how the different (very powerful) functionality can be used.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -504,16 +504,16 @@ impl MyClass {
 }
 
 #[pymodule]
-fn rpi3d(_py: Python, m: &PyModule) -> PyResult<()> {
+fn mymodule(_py: Python, m: &PyModule) -> PyResult<()> {
     m.add_class::<MyClass>()?;
     Ok(())
 }
 ```
 N.B. the position of the `"*"` argument (if included) controls the system of handling positional and keyword arguments. In Python:
 ```python
-import rpi3d
+import mymodule
 
-mc = rpi3d.MyClass()
+mc = mymodule.MyClass()
 print(mc.method(44, False, "World", 666, x=44, y=55))
 print(mc.method(num=-1, name="World"))
 print(mc.make_change(44, False))
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -21,6 +21,12 @@ The above example generates implementations for `PyTypeInfo` and `PyTypeObject` 
 Custom Python classes can then be added to a module using `add_class`.
 
 ```rust
+# use pyo3::prelude::*;
+# #[pyclass]
+# struct MyClass {
+#    num: i32,
+#    debug: bool,
+# }
 #[pymodule]
 fn mymodule(_py: Python, m: &PyModule) -> PyResult<()> {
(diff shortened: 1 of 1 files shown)


guide/src/class.md
```diff
@@ -13,7 +13,14 @@ struct MyClass {
    num: i32,
    debug: bool,
 }
+
+#[pymodule]
```
kngwyu: I think this should be in another subsection, e.g., `## Add class to module`

guide/src/class.md
```diff
@@ -460,13 +478,54 @@ use pyo3::types::{PyDict, PyTuple};
...
+    fn new(obj: &PyRawObject, num: i32, debug: bool) {
+        obj.init({
+            MyClass {
+                num,
+                debug,
+            }
+        });
+    }
+    #[args(num="10", debug="true", py_args="*", name="\"Hello\"", py_kwargs="**")]
+    fn method(&mut self, num: i32, debug: bool, name: &str, py_args: &PyTuple,
```
kngwyu: Could you please format this method in rustfmt style?
```rust
    #[args(
        num = "10",
        debug = "true",
        py_args = "*",
        name = "\"Hello\"",
        py_kwargs = "**"
    )]
    fn method(
        &mut self,
        num: i32,
        debug: bool,
        name: &str,
        py_args: &PyTuple,
        py_kwargs: Option<&PyDict>,
    ) -> PyResult<String> {
        ...
    }
```

guide/src/class.md
```diff
@@ -460,13 +478,54 @@ use pyo3::types::{PyDict, PyTuple};
...
 }
-```
 
+#[pymodule]
+fn mymodule(_py: Python, m: &PyModule) -> PyResult<()> {
+    m.add_class::<MyClass>()?;
+    Ok(())
+}
+```
+N.B. the position of the `"*"` argument (if included) controls the system of handling positional and keyword arguments. In Python:
```
kngwyu: This Python code example is really good :+1:

guide/src/class.md
```diff
@@ -460,13 +478,54 @@ use pyo3::types::{PyDict, PyTuple};
...
+    fn make_change(&mut self, num: i32, debug: bool) -> PyResult<String> {
+        self.num = num;
+        self.debug = debug;
+        Ok(format!("num={}, debug={}", self.num, self.debug))
     }
 }
-```
 
+#[pymodule]
+fn mymodule(_py: Python, m: &PyModule) -> PyResult<()> {
```
kngwyu: I think this is unrelated here

kngwyu:
Thank you, they are missing parts for us.

guide/src/class.md
paddywwoof: Hi, all good suggestions. Hopefully I've incorporated them all.

kngwyu:
Thank you!

PR #716, opened 2020-01-08
Implement IntoIterator for PySet and PyFrozenSet

### LINKED ISSUE #715: PySet is not iterable
As per [the docs](https://docs.rs/pyo3/0.8.5/pyo3/types/struct.PySet.html) (and reproduced with pyo3 v0.8.5, python 3.7, ubuntu) `PySet` is not an iterator.

Therefore `for element in my_set {` throws an error.

I tried wrapping the set in `PyList::new` but got `the trait std::iter::ExactSizeIterator is not implemented for &pyo3::types::PySet`.

Is there a good reason to prevent iteration over sets? If not could it be added, also is there a workaround for now?

davidhewitt:
Closes #715

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * The blanket implementations for `FromPyObject` for `&T` and `&mut T` are no longer specializable. Implement `PyTryFrom` for your type to control the behavior of `FromPyObject::extract()` for your types.
 * The implementation for `IntoPy<U> for T` where `U: FromPy<T>` is no longer specializable. Control the behavior of this via the implementation of `FromPy`.
 
+### Added
+
+* Implemented `IntoIterator` for `PySet` and `PyFrozenSet`. [#716](https://github.com/PyO3/pyo3/pull/716)
+
 ## [0.8.5]
 
 * Support for `#[name = "foo"]` attribute for `#[pyfunction]` and in `#[pymethods]`. [#692](https://github.com/PyO3/pyo3/pull/692)
--- a/src/ffi/setobject.rs
+++ b/src/ffi/setobject.rs
@@ -60,4 +60,13 @@ extern "C" {
     pub fn PySet_Add(set: *mut PyObject, key: *mut PyObject) -> c_int;
     #[cfg_attr(PyPy, link_name = "PyPySet_Pop")]
     pub fn PySet_Pop(set: *mut PyObject) -> *mut PyObject;
+
+    #[cfg(not(Py_LIMITED_API))]
+    #[cfg_attr(PyPy, link_name = "_PySet_NextEntry")]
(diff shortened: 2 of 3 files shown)


kngwyu:
Could you please rewrite the implementation using `_PySet_NextEntry`?
1. Add this declaration in `ffi::setobject.rs`
```rust
#[cfg(not(Py_LIMITED_API))]
extern "C" {
    #[cfg_attr(PyPy, link_name = "_PySet_NextEntry")]
    pub fn _PySet_NextEntry(
        set: *mut PyObject,
        pos: *mut Py_ssize_t,
        key: *mut *mut PyObject,
        hash: *mut super::Py_hash_t,
    ) -> c_int;
}
```

2. Then implementation comes like
```rust
use super::PyAny;
pub struct PySetIterator<'py> {
    set: &'py PyAny,
    pos: isize,
}

impl<'py> Iterator for PySetIterator<'py> {
    type Item = &'py PyAny;

    #[inline]
    fn next(&mut self) -> Option<Self::Item> {
        unsafe {
            let mut key: *mut ffi::PyObject = std::ptr::null_mut();
            let mut hash: ffi::Py_hash_t = 0;
            if ffi::_PySet_NextEntry(self.set.as_ptr(), &mut self.pos, &mut key, &mut hash) != 0 {
                Some(self.set.py().from_borrowed_ptr(key))
            } else {
                None
            }
        }
    }
}

impl<'a> std::iter::IntoIterator for &'a PySet {
    type Item = &'a PyAny;
    type IntoIter = PySetIterator<'a>;

    fn into_iter(self) -> Self::IntoIter {
        PySetIterator {
            set: self.as_ref(),
            pos: 0,
        }
    }
}
```

davidhewitt:
👍 Thanks - I missed `_PySet_Next` because I didn't find it at https://docs.python.org/3/c-api/set.html!

kngwyu:
Thank you!

samuelcolvin:
Thanks for fixing this.

My only concern is that `PySet` still doesn't implement `iter()` as an explicit method (i'm not sure how that's being implemented, but I guess it must be as it's working in tests) so it's not included in the documentation.

For a list (ignoring the comment that I guess is left over from copying the method from tuples) we have:

https://github.com/PyO3/pyo3/blob/b3ca27dd1c3817659d6d5f29ffd937880bd949be/src/types/list.rs#L116-L122

Would it help to add `iter()` to sets? It's trivial enough that even I can create a PR.

davidhewitt:
I left `.iter()` off because I didn't see a need for it, given `IntoIterator` is now implemented and that'll be in the docs.

If you still would like to have `iter()`, please do open a PR. Note you'd need to wrap any `iter()` method you add in `#[cfg(not(Py_LIMITED_API))]`, like the other definitions we just merged.

samuelcolvin:
okay, I'll leave it.

kngwyu:
@samuelcolvin 
I implemented it in #717, for the compatibility with `PyDict`.

PR #717, opened 2020-01-08
Implement iter for PySet and PyFrozenSet

kngwyu:
Continuation of #716


samuelcolvin:
Maybe worth correcting the description on `PyList.iter()` at the same time?

kngwyu:
Thanks :+1:

PR #719, opened 2020-01-10
Clear error indicator when the exception is handled on the Rust side

### LINKED ISSUE #688: PySet::pop sets the global exception state
Discovered by running the test suite locally.

The `PySet_Pop` that’s run as part of the `test_set_pop` will return NULL *and* set the global exception state (raise an exception) when the set is empty. We handle the `NULL` case and return an `Option<T>`, but the exception state remains set.

Given that the convention in PyO3 is to convert exceptions to return values, this can be very misleading. In particular, this causes the test suite to fail further along when one of the tests invokes `py.eval` and then converts this exception to a `Result`.

To reproduce try: `cargo test --lib -- --test-threads=1`.

Alexander-N:
Leaving Python's global exception state is misleading, e.g. subsequent
calls of `py.eval` will fail.

Closes #688


kngwyu:
Nice :+1: , thank you!

PR #722, opened 2020-01-12
Remove ignoring of new_ret_no_self in clippy target

Alexander-N:
After the changes in the `#[pyclass]` system in #683, `new` can return
self and there is no reason anymore to ignore this lint.


kngwyu:
Thanks.

PR #724, opened 2020-01-12
Fix coverage

Alexander-N:
There seems to be a problem with the reported coverage, the last commit on https://codecov.io/gh/PyO3/pyo3 is 3 months old.

Edit: A grcov flag was renamed https://github.com/mozilla/grcov/issues/193


kngwyu:
Thank you!

Issue #720, opened 2020-01-11
Flaky datetime test


Alexander-N:
The test `test_datetime_check` fails from time to time with an error like
```
---- test_datetime_check stdout ----
thread 'test_datetime_check' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: Py(0x60d15300, PhantomData) }', src\libcore\result.rs:1192:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
It could be that this occurs only on Windows with Python 3.6: https://github.com/PyO3/pyo3/pull/719/checks?check_run_id=384307686 and https://github.com/PyO3/pyo3/pull/651#issuecomment-546640959.

@pganssle Do you have any idea what could cause this?

davidhewitt:
When developing pyO3 on Windows with 3.8 the test wobbles for me occasionally too.

pganssle:
This is a real head-scratcher. I would expect possible "flakiness" for the hypothesis-based tests based on rare edge-cases, but `test_datetime_check` should be entirely deterministic (and not dependent on platform, either). That test just constructs a few `datetime` subclasses and calls the "check" functions on them, pretty dead simple.

Maybe we should start running the tests with `RUST_BACKTRACE=1`, so we can get a bit more info in hard-to-repro cases like this?

My immediate possible thoughts are some sort of thread-safety issue or possibly another test setting the Python exception state and returning without failing to check it (which would then raise an exception in whatever the next thing to run is that sets the exception state).

One possibility is that this is bleeding over from some issue with [these out-of-bounds tests](https://github.com/PyO3/pyo3/blob/68726f9674d4c79b0437ce559af4964347f82b12/tests/test_datetime.rs#L146), but that shouldn't be Windows-specific.

davidhewitt:
I had this fail on me on CI earlier today. I think it was in a Linux environment?!

https://github.com/davidhewitt/pyo3/commit/43df3299abc3cfdf006914c44d5b7f3d0a987f6c/checks?check_suite_id=414584878

pganssle:
> I had this fail on me on CI earlier today. I think it was in a Linux environment?!

I don't think so - maybe WSL or something, but a lot of the paths use `\` as a separator.

[...]

PR #726, opened 2020-01-12
Set RUST_BACKTRACE=1 in windows tests

Alexander-N:
Setting `RUST_BACKTRACE=1` as suggested in https://github.com/PyO3/pyo3/issues/720#issuecomment-573414444.


kngwyu:
Isn't #727 sufficient? :thinking:

Alexander-N:
> Isn't #727 sufficient?

Yes it is. But I think getting the full backtrace is useful in any case, we do this already in Travis.

kngwyu:
I got it:+1:, thanks.

PR #730, opened 2020-01-13
[WIP] FromPyObject for #[pyclass] with T: Clone

### LINKED ISSUE #706: Automatically implement `Clone` based conversion
`&PyAny -> T where T: Clone` conversion should be implementd for all `T: PyClass + Clone`, though I don't know it's possible...
```rust
#[pyclass]
#[derive(Clone)]
struct Attr {
    inner: u32,
}

impl<'source> pyo3::FromPyObject<'source> for Attr {
    fn extract(ob: &'source pyo3::types::PyAny) -> pyo3::PyResult<Attr> {
        let cls: &Attr = pyo3::PyTryFrom::try_from(ob)?;
        Ok(cls.clone())
    }
}
```

davidhewitt:
I had an idea how to support extracting for pyclasses automatically for `T: Clone`.

It also has really nice error messages if the user attempts to extract a non-clone `#[pyclass]` (see the `ui` test).

Closes #706

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -244,25 +244,80 @@ where
     }
 }
 
-/// Extract reference to instance from `PyObject`
-impl<'a, T> FromPyObject<'a> for &'a T
-where
-    T: PyTryFrom<'a>,
-{
-    #[inline]
-    fn extract(ob: &'a PyAny) -> PyResult<&'a T> {
-        Ok(T::try_from(ob)?)
+#[doc(hidden)]
+pub mod extract_impl {
+    use super::*;
+
+    pub trait ExtractImpl<'a, Target>: Sized {
+        fn extract(source: &'a PyAny) -> PyResult<Target>;
+    }
+
+    pub struct Cloned;
+    pub struct Reference;
+    pub struct MutReference;
+
+    impl<'a, T: 'a> ExtractImpl<'a, T> for Cloned
+    where
+        T: Clone,
+        Reference: ExtractImpl<'a, &'a T>,
+    {
+        fn extract(source: &'a PyAny) -> PyResult<T> {
+            Ok(Reference::extract(source)?.clone())
+        }
+    }
+
+    impl<'a, T> ExtractImpl<'a, &'a T> for Reference
+    where
+        T: PyTryFrom<'a>,
+    {
+        fn extract(source: &'a PyAny) -> PyResult<&'a T> {
+            Ok(T::try_from(source)?)
+        }
+    }
+
+    impl<'a, T> ExtractImpl<'a, &'a mut T> for MutReference
+    where
+        T: PyTryFrom<'a>,
+    {
+        fn extract(source: &'a PyAny) -> PyResult<&'a mut T> {
+            Ok(T::try_from_mut(source)?)
+        }
     }
 }
 
-/// Extract mutable reference to instance from `PyObject`
-impl<'a, T> FromPyObject<'a> for &'a mut T
+use extract_impl::ExtractImpl;
+
+/// This is an internal trait for re-using `FromPyObject` implementations for many pyo3 types.
+///
+/// Users should implement `FromPyObject` directly instead of via this trait.
+pub trait FromPyObjectImpl {
+    // Implement this trait with to specify the implementor of `extract_impl::ExtractImpl` to use for
+    // extracting this type from Python objects.
+    //
+    // Example valid implementations are `extract_impl::Cloned`, `extract_impl::Reference`, and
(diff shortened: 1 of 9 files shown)


kngwyu:
Nice :+1: , thank you.
But please wait for the review for a while, since I want to check we can simplify the implementation or not.

davidhewitt:
Sure thing, take as long as you like!

I tried myself to do some simplifications but couldn't get much further than what's above.

It is possible to add some blanket `impl ExtractPolicy` for `&T` and `&mut T`, for example. But it's not possible to also have a blanket impl for `T` because it conflicts with the other two.

davidhewitt:
I did spot a slight simplification!

Also, I realised that if I work the `FromPyObject` implementation for `Vec<T>` using this new `FromPyObjectImpl` trait, we can get the `buffer::Element` optimization without specialization. See https://github.com/davidhewitt/pyo3/pull/1

It isn't perfect and it's really un-ergonomic to have to implement `FromPyObjectImpl` for everything to get `Vec<T>` to work.

So we might want to think a bit longer about this idea. I think if we can find a nice API for it then we can use it to remove specialization from `impl FromPyObject for Vec<T>`.

kngwyu:
Sorry for the lazy review, but after playing around the code, I'm really convinced that this is almost the minimal change :+1:

src/conversion.rs
```diff
@@ -244,25 +244,77 @@ where
...
+pub mod extract_impl {
+    use super::*;
+
+    pub trait ExtractImpl<'a, Target>: Sized {
+        fn extract(source: &'a PyAny) -> PyResult<Target>;
+    }
+
+    pub struct Cloned;
+    pub struct Reference;
+    pub struct MutReference;
```
kngwyu: Do we need `MutReference`?
We can also do
```rust

    impl<'a, T> ExtractImpl<'a, &'a mut T> for Reference
    where
        T: PyTryFrom<'a>,
    {
        fn extract(source: &'a PyAny) -> PyResult<&'a mut T> {
            Ok(T::try_from_mut(source)?)
        }
    }
```
...
But, it is also good to leave it as is for ergonomics.

src/lib.rs
```diff
@@ -119,8 +119,8 @@
 
 pub use crate::class::*;
 pub use crate::conversion::{
-    AsPyPointer, FromPy, FromPyObject, FromPyPointer, IntoPy, IntoPyPointer, PyTryFrom, PyTryInto,
-    ToBorrowedObject, ToPyObject,
+    extract_impl, AsPyPointer, FromPy, FromPyObject, FromPyObjectImpl, FromPyPointer, IntoPy,
```
kngwyu: Should we expose `FromPyObjectImpl` on the top page of doc? :thinking:

src/conversion.rs
```diff
@@ -244,25 +244,77 @@ where
...
 
-/// Extract reference to instance from `PyObject`
-impl<'a, T> FromPyObject<'a> for &'a T
-where
-    T: PyTryFrom<'a>,
-{
-    #[inline]
-    fn extract(ob: &'a PyAny) -> PyResult<&'a T> {
-        Ok(T::try_from(ob)?)
+pub mod extract_impl {
```
kngwyu: I checked `extract_impl` doesn't appear on the rust doc.
If it is intended, this should be marked as `#[doc(hidden)]`

kngwyu:
Just a note:
I added this PR to 0.9.0 milestone since somehow rust-numpy needs this fix.

davidhewitt:
Ok. I'm closing in on a design I like for the `Vec<T>` specialization solution (see the PR I linked in comment above).

Once that's ready I'll open it as a PR against pyO3 and then we can choose whether to take this simpler change, or the bigger design change from that PR.

src/conversion.rs
davidhewitt: I think it's clearer to have all three, even if we only technically need two.

src/lib.rs
davidhewitt: Good idea. I've taken it off.

src/conversion.rs
davidhewitt: 👍

src/lib.rs
```diff
@@ -152,7 +152,7 @@ pub mod buffer;
 #[doc(hidden)]
 pub mod callback;
 pub mod class;
-mod conversion;
+pub mod conversion;
```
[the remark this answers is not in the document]
kngwyu: :+1:

kngwyu:
Thank you!

PR #733, opened 2020-01-16
Bump version to 0.9.0-alpha.1

kngwyu:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pyo3"
-version = "0.8.5"
+version = "0.9.0-alpha.1"
 description = "Bindings to Python interpreter"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -25,11 +25,9 @@ libc = "0.2.62"
 num-bigint = { version = ">= 0.2", optional = true }
 num-complex = { version = ">= 0.2", optional = true }
 num-traits = "0.2.8"
+parking_lot = { version = "0.10", features = ["nightly"] }
+paste = "0.1.6"
 pyo3cls = { path = "pyo3cls", version = "=0.9.0-alpha.1" }
-num-complex = { version = ">= 0.2", optional = true }
-num-bigint = { version = ">= 0.2", optional = true }
-inventory = "0.1.4"
-indoc = "0.3.4"
```
Alexander-N: I think these should not get removed.

PR #734, opened 2020-01-16
Use parking_lot::Mutex instead of spin::Mutex

### LINKED ISSUE #718: spin is no longer actively maintained
cargo audit warns that the spin dependency of pyo3 is unmaintained. 
https://rustsec.org/advisories/RUSTSEC-2019-0031.html

ijl:
spin is no longer maintained.

Fixes #718.


birkenfeld:
So the obvious question: why didn't the original use the `std` Mutex -- and is `parking_lot` the wrong choice for the same reason?

ijl:
spin and parking_lot are both faster than std. This change is least likely to cause a difference for dependent projects. For anything more, putting energy into #679 would make these locks unnecessary.

kngwyu:
@birkenfeld 
I'm not the original author of this part and don't know the intention.
Both spin and parking_lot's documents say
> No poisoning, the lock is released normally on panic.

So the difference from std is they return Error when panicking or not.
And... I honestly am not sure this feature matters, since I think most of the users don't write code where multiple threads try to get GIL :thinking:

kngwyu:
@ijl 
Oh, I wrote the above comment before reading your comment.
> spin and parking_lot are both faster than std. 

I understand..., thanks.

kngwyu:
Sorry, I want to think about std vs parking_lot problem more.
Wait for a while.

kngwyu:
How we should handle `LockResult` is not obvious, so I decided to use `parking_lot` for now.
However, maybe we should remove it in the near future.

PR #735, opened 2020-01-16
_PyDict_SetItem_KnownHash(), _Py_HashBytes()

ijl:
These are two CPython-specific APIs available since before 3.5. I've used these locally with success.


kngwyu:
LGTM :+1: , thanks.

PR #745, opened 2020-01-27
Fix usage of raw idents with #[pyo3(set)]

### LINKED ISSUE #743: Rust ICE when attribute is escaped like r#type
### 💥 Reproducing

If you have code like

```rust
#[pyclass]
struct Test {
    #[pyo3(get, set)]
    r#type: i64
}
```
Rust ICEs:
```
  Compiling twitter v0.1.0 (C:\Users\ethanhs\abserde\build\abserde\twitter)
thread 'rustc' panicked at '`"set_r#type"` is not a valid identifier', src\librustc_expand\proc_macro_server.rs:333:13

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: rustc 1.42.0-nightly (b5a3341f1 2020-01-20) running on x86_64-pc-windows-msvc

note: compiler flags: -C opt-level=3 -C lto -C codegen-units=1 -C target-cpu=native --crate-type cdylib

note: some of the compiler flags provided by cargo are hidden

error: could not compile `twitter`.

To learn more, run the command again with --verbose.
💥 maturin failed
  Caused by: Failed to build a native library through cargo
  Caused by: Cargo build finished with "exit code: 101": `cargo rustc --message-format json --manifest-path Cargo.toml --lib --release --`

🔗 Found pyo3 bindings
🐍 Found CPython 3.7 at C:\Users\ethanhs\AppData\Local\pypoetry\Cache\virtualenvs\abserde-1LDqkpmO-py3.7\Scripts\python.exe
error: custom attribute panicked
   --> src\lib.rs:484:1
    |
484 | #[pyclass(dict)]
    | ^^^^^^^^^^^^^^^^
    |
    = help: message: `"set_r#type"` is not a valid identifier


```

It seems that it can't handle identifiers like `set_r#type`. I will report the ICE upstream but I think pyo3 should be able to support escaped keywords as attributes.

davidhewitt:
Fixes #743

### DIFF
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -427,7 +427,7 @@ fn impl_descriptors(
         .flat_map(|&(ref field, ref fns)| {
             fns.iter()
                 .map(|desc| {
-                    let name = field.ident.clone().unwrap();
+                    let name = field.ident.as_ref().unwrap();
                     let field_ty = &field.ty;
                     match *desc {
                         FnType::Getter => {
@@ -441,7 +441,7 @@ fn impl_descriptors(
                         }
                         FnType::Setter => {
                             let setter_name =
-                                syn::Ident::new(&format!("set_{}", name), Span::call_site());
+                                syn::Ident::new(&format!("set_{}", name.unraw()), Span::call_site());
                             quote! {
                                 impl #cls {
                                     fn #setter_name(&mut self, value: #field_ty) -> pyo3::PyResult<()> {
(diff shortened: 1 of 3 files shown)


davidhewitt:
CI failure is because of #747

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -441,7 +441,7 @@ fn impl_descriptors(
                         }
                         FnType::Setter => {
                             let setter_name =
-                                syn::Ident::new(&format!("set_{}", name), Span::call_site());
+                                syn::Ident::new(&format!("set_{}", name.unraw()), Span::call_site());
```
kngwyu: :+1:

davidhewitt:
Rebased now CI should be green.

kngwyu:
Thank you!

ethanhs:
Thank you so much @davidhewitt !

PR #747, opened 2020-01-27
Fix CI failures

davidhewitt:
Seems like a recent nightly has added a warning for having unneeded brackets in this position, which caused CI failures on my other PRs this morning.


kngwyu:
Thank you!

PR #751, opened 2020-01-30
Remove static mut from PyTypeInfo implementation

### LINKED ISSUE #699: static mut should not be used
See https://github.com/PyO3/pyo3/pull/683#discussion_r360727301

davidhewitt:
Fixes #699 .

I looked at using `lazy_static!` and a couple other options, in the end I went for `once_cell`.

I modified the signature of `PyTypeInfo::type_object()` to return `*mut` instead of `&mut`. It's a minor thing but handing out `&mut` in this way is definitely breaking Rust's aliasing rules.

Also removed `unsafe` from the function because there's nothing unsafe about calling it - the user is responsible for using the returned  `*mut ffi::PyTypeInfo` pointer safely.

Might need a changelog entry because the trait signature changed?

(As a note, I plan to follow up this PR with a second one which removes the need for `init_type()`. Let's start with this to keep the PR smaller and simpler.)

### DIFF
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -384,9 +384,10 @@ fn impl_class(
             const FLAGS: usize = #(#flags)|* | #extended;
 
             #[inline]
-            unsafe fn type_object() -> &'static mut pyo3::ffi::PyTypeObject {
-                static mut TYPE_OBJECT: pyo3::ffi::PyTypeObject = pyo3::ffi::PyTypeObject_INIT;
-                &mut TYPE_OBJECT
+            fn type_object() -> *mut pyo3::ffi::PyTypeObject {
+                static TYPE_OBJECT: pyo3::derive_utils::LazyTypeObject =
+                    pyo3::derive_utils::LazyTypeObject::new();
+                TYPE_OBJECT.get()
             }
         }
 
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -273,8 +273,6 @@ fn impl_class(
                             if FREELIST.is_null() {
                                 FREELIST = Box::into_raw(Box::new(
                                     pyo3::freelist::FreeList::with_capacity(#freelist)));
-
-                                <#cls as pyo3::type_object::PyTypeObject>::init_type();
                             }
                             &mut *FREELIST
                         }
@@ -372,7 +370,7 @@ fn impl_class(
     };
 
     Ok(quote! {
-        impl pyo3::type_object::PyTypeInfo for #cls {
+        unsafe impl pyo3::type_object::PyTypeInfo for #cls {
             type Type = #cls;
             type BaseType = #base;
             type ConcreteLayout = pyo3::pyclass::PyClassShell<Self>;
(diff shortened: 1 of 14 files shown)


programmerjake:
the `PyTypeInfo` trait should be `unsafe`, because otherwise a user can cause UB by mis-implementing it.

kngwyu:
> I modified the signature of PyTypeInfo::type_object() to return *mut instead of &mut. 

:+1: for this change.

So... I think this PR is good but the problem is our pair of `PyTypeInfo::type_object` and `PyTypeObject` has really complex semantics(I don't remember correctly... but maybe to avoid specialization?).

I'm sorry but I'm going to merge this PR after understanding the problem clearer.

davidhewitt:
No rush. 👍 

I spent some time thinking about `PyTypeObject` this morning. I experimented with merging `PyTypeInfo` and `PyTypeObject`, but it leads to refactorings across the codebase. Probably too big to land in a single PR. I'll share notes about it sometime.

davidhewitt:
I had some ideas to take this a little further:

- Now `type_object()` always returns an **initialized** `PyTypeObject`.
- `type_object()` now returns `NonNull` instead of `*mut`.
- `PyTypeInfo` is now `unsafe trait`
- Also removed use of `static mut` in all the `exceptions.rs` macros.

kngwyu:
Sorry for the delay with the review..., but now I'm considering `fn type_object() -> Pin<&'static ffi::PyTypeObject` for this purpose and experimenting it.

davidhewitt:
Please do experiment with better designs! I'd rather we make these changes once and get them right than break stuff again and again :)

kngwyu:
I tried to remove `Box::new` in `create_type_object` but this attempt just told me that `ffi::PyType_Ready` requires type object has static address :roll_eyes: 
So now I think the currently implementation is sufficiently good, except some minor things.

I also tried `Pin<&'static ffi::PyTypeObject>`, but finally I didn't like it since it requires too many pointer conversion.

kngwyu:
force-push is my failure... don't mind it.

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -384,9 +382,23 @@ fn impl_class(
             const FLAGS: usize = #(#flags)|* | #extended;
 
             #[inline]
-            unsafe fn type_object() -> &'static mut pyo3::ffi::PyTypeObject {
-                static mut TYPE_OBJECT: pyo3::ffi::PyTypeObject = pyo3::ffi::PyTypeObject_INIT;
-                &mut TYPE_OBJECT
+            fn type_object() -> std::ptr::NonNull<pyo3::ffi::PyTypeObject> {
```
kngwyu: How about moving this code to impl block of `LazyTypeObject`?
Then we have 
```rust
impl LazyTypeObject {
...
    pub fn get_pyclass_type<T: PyClass>(&self) -> NonNull<ffi::PyTypeObject> { ... }
}

File: src/pyclass.rs [hunk 2]
```diff
@@ -276,15 +252,17 @@ where
...
+    py: Python,
+    module_name: Option<&str>,
+) -> PyResult<Box<ffi::PyTypeObject>>
 where
     T: PyClass,
 {
-    let type_object: &mut ffi::PyTypeObject = unsafe { T::type_object() };
-    let base_type_object: &mut ffi::PyTypeObject =
-        unsafe { <T::BaseType as PyTypeInfo>::type_object() };
+    let mut boxed = Box::new(ffi::PyTypeObject_INIT);
```
kngwyu: Let's comment why we need `Box` here(i.e., because `PyType_Ready` reuires a constant address).

src/type_object.rs
```diff
@@ -85,32 +91,64 @@ pub trait PyTypeInfo: Sized {
...
+where
+    T: PyTypeInfo,
+{
     fn type_object() -> Py<PyType> {
-        unsafe { Py::from_borrowed_ptr(Self::init_type().as_ptr() as *mut ffi::PyObject) }
+        unsafe { Py::from_borrowed_ptr(<Self as PyTypeInfo>::type_object().as_ptr() as *mut _) }
     }
 }
+
+/// Type used to store type objects
```
kngwyu: I think `#[doc(hidden)]` is prefferable here.

pyo3-derive-backend/src/pyclass.rs
davidhewitt: Thanks, that's a really nice idea! It makes the macro code a lot neater 😄

kngwyu:
Thank you!

Issue #704, opened 2019-12-28
Full Python 3.8 FFI support


Alexander-N:
I think https://github.com/dgrunwald/rust-cpython/pull/187 does provide a good starting point for us.

3.8 should also be added to CI. There was some issue when I tried to replace 3.8-dev with 3.8 in https://github.com/PyO3/pyo3/pull/648

Edit: Removed unrelated issue.

kngwyu:
I wrote some patches on Python3.8 and observed no test failing...
But yeah, now it's time to check FFI compatiblitity and add it to CI.

Alexander-N:
With github actions I get the same error as with travis:
```
AttributeError: module 'os' has no attribute 'link'
```
from https://github.com/PyO3/pyo3/runs/365982476

kngwyu:
> With github actions I get the same error as with travis:
AttributeError: module 'os' has no attribute 'link'

I saw the same error when I set up Python3.8 CI for Mac and Win on Azure, and still don't know the solution :sweat:

davidhewitt:
Now that we have 3.8 on CI, what else are we wanting to add here? (Presumably all new C APIs added in Python 3.8?)

Alexander-N:
> Now that we have 3.8 on CI, what else are we wanting to add here? (Presumably all new C APIs added in Python 3.8?)

This and I would say checking that all [changes in the c-api](https://docs.python.org/3/whatsnew/3.8.html#changes-in-the-c-api) are accounted for. For example in https://github.com/dgrunwald/rust-cpython/pull/187 there was a `#[cfg(all(Py_3_5, not(Py_3_8)))]` added on `PyTypeObject_INIT`.

PR #753, opened 2020-02-01
Add Python 3.8 to CI

Alexander-N:
The problem mentioned in https://github.com/PyO3/pyo3/issues/704#issuecomment-569410933 seems gone, probably fixed by https://github.com/python/cpython/commit/8d0f36940e728989822c3789025b0813a8fe249a.


kngwyu:
Wow.
I'm really happy to see this.

Thank you!

PR #754, opened 2020-02-02
Implement From<&PyNativeType> for Py

kngwyu:
Since rust-numpy has already [this kind of conversion](https://docs.rs/numpy/0.7.0/numpy/array/struct.PyArray.html#method.to_owned), it's reasonable to we have this

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -164,6 +164,17 @@ impl<T> IntoPyPointer for Py<T> {
     }
 }
 
+// Native types `&T` can be converted to `Py<T>`
+impl<'a, T> std::convert::From<&'a T> for Py<T>
+where
+    T: AsPyPointer + PyNativeType,
+{
+    fn from(obj: &'a T) -> Self {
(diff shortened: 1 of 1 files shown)


src/instance.rs
```diff
@@ -164,6 +164,17 @@ impl<T> IntoPyPointer for Py<T> {
     }
 }
 
+// Native types `&T` can be converted to `Py<T>`
+impl<'a, T> std::convert::From<&'a T> for Py<T>
+where
+    T: AsPyPointer + PyNativeType,
```
davidhewitt: The impl for `PyObject` is:

    impl<'a, T> std::convert::From<&'a T> for PyObject
    where
        T: AsPyPointer,

Maybe here we also don't need the `PyNativeType` bound?

`PyObject` also has the same impl for `&'a mut T`, but I'm unsure whether we want that.

kngwyu: This bound is not to allow `PyClassShell<T>` -> `Py<PyClassShell<T>>`

davidhewitt: Ah, seems reasonable :+1:

PR #755, opened 2020-02-03
Fix docstrings generated by `[pyo3(get, set)]`

davidhewitt:
Use the Rust docs from the field for `#[pyo3(get, set)]` property macros.

It was asked on Gitter why it didn't work, and I thought it probably should!


kngwyu:
Nice, thank you!

PR #757, opened 2020-02-06
Make PyTypeInfo::type_object return &'static instead of NonNull

kngwyu:
Here `NonNull` doesn't mean so much and... I think `static` means what we need clearer.


kngwyu:
Closed in favor of #758

PR #760, opened 2020-02-08
Improvements to property macros

davidhewitt:
There are two main fixes that I've added to the property macro code:

- `#[pyo3(get)]` now works with `PyObject` fields
- `#[setter]` functions can now take a `Python` argument

 After implementing the first I'd changed enough of the code such that cleaning it up caused the second to be fixed trivially.

I also bumped the minimum Rust version to `1.42.0-nightly 2020-01-21`, to make use of the recently stabilized `#![feature(slice_patterns)]`.


davidhewitt:
The CI failures are from new clippy lints from the newer rust version. I've added a PR #761 to fix those separately.

src/derive_utils.rs
```diff
@@ -199,3 +199,22 @@ impl<T: PyClass, I: Into<PyClassInitializer<T>>> IntoPyNewResult<T, I> for PyRes
...
+pub trait GetPropertyValue {
+    fn get_property_value(&self, py: Python) -> PyObject;
+}
+
+impl<T> GetPropertyValue for &T
+where
+    T: IntoPy<PyObject> + Clone,
+{
+    fn get_property_value(&self, py: Python) -> PyObject {
+        (*self).clone().into_py(py)
```
kngwyu: Why `Deref` here... ?

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -244,28 +258,48 @@ pub fn impl_wrap_static(cls: &syn::Type, spec: &FnSpec<'_>) -> TokenStream {
...
+    let fncall = if py_arg.is_some() {
         quote! { _slf.#name(_py) }
     } else {
         quote! { _slf.#name() }
     };
 
+    Ok(fncall)
+}
+
+/// Generate functiona wrapper (PyCFunction, PyCFunctionWithKeywords)
```
kngwyu: typo?

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -291,24 +325,48 @@ pub(crate) fn impl_wrap_getter(cls: &syn::Type, spec: &FnSpec) -> syn::Result<To
...
-            ))
+    Ok(fncall)
+}
+
+/// Generate functiona wrapper (PyCFunction, PyCFunctionWithKeywords)
+pub(crate) fn impl_wrap_setter(
+    cls: &syn::Type,
+    property_type: PropertyType,
+) -> syn::Result<TokenStream> {
+    let python_name;
```
kngwyu: I recommend
```rust
let (python_name, setter_impl) = match property_type {
    ...
};
```

src/derive_utils.rs
```diff
@@ -199,3 +199,22 @@ impl<T: PyClass, I: Into<PyClassInitializer<T>>> IntoPyNewResult<T, I> for PyRes
...
+    }
+}
+
+impl GetPropertyValue for PyObject {
+    fn get_property_value(&self, py: Python) -> PyObject {
+        self.to_object(py)
```
kngwyu: `self.clone_ref(py)`

kngwyu:
Thank you!
I left some comments but generally LGTM :+1: 
But it looks you update minimum nightly also in #761

src/derive_utils.rs
davidhewitt: `&self` is `&&T`,  so `self.clone()` makes `&T`. We need to deref like `(*self).clone()` to get a `T`.

pyo3-derive-backend/src/pymethod.rs
davidhewitt: This comment actually looks very out of date - I'll update it.

pyo3-derive-backend/src/pymethod.rs
davidhewitt: 👍 I did consider that style too, but marginally preferred the style I chose. I'll switch to what you propose.

src/derive_utils.rs
```diff
@@ -199,3 +199,22 @@ impl<T: PyClass, I: Into<PyClassInitializer<T>>> IntoPyNewResult<T, I> for PyRes
...
     }
 }
+
+pub trait GetPropertyValue {
+    fn get_property_value(&self, py: Python) -> PyObject;
+}
+
+impl<T> GetPropertyValue for &T
+where
+    T: IntoPy<PyObject> + Clone,
```
davidhewitt: I was wondering whether the bound should really be `Copy`, so that we don't let people accidentally do really expensive properties with `#[pyo3(get)]` (such as `Vec<T>`).

They could always write their own `#[getter]` functions for types which aren't `Copy`.

davidhewitt:
Thanks, I've made some changes to address your comments. I suggest we wait for #761 to merge first. After that I'll fix up the merge conflicts in this PR and then we can potentially merge this.

kngwyu:
Thank you!

davidhewitt:
Thanks for fixing the merge conflicts!

PR #761, opened 2020-02-08
Bump minimum Rust version to 1.42.0-nightly

davidhewitt:
I fixed all the clippy warnings as of present.

To silence the safety warnings for now (#698) I've added an `#![allow(clippy::missing_safety_doc)]`. This should make it easier to check for new clippy errors when people are developing patches while we wait for someone to pick up #698.


kngwyu:
Ooops, `missing_safety_doc` is from 1.39 https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-139, so `minimum_nightly` fails in traivis.

Could you please update minimum nightly, or remove `#![allow(clippy::missing_safety_doc)]`?

davidhewitt:
I've moved the minimum Rust version bump from #760 into this PR.

kngwyu:
Thank you!

Issue #742, opened 2020-01-22
test suite failures and missing symbols on macOS


damaxwell:
## 🐛 Bug Reports

There are failures in the test suite on macOS with `pyenv` installed Python 3.7.6.

```
failures:

---- buffer::test::test_array_buffer stdout ----
thread 'buffer::test::test_array_buffer' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: Py(0x10b9fa4f0, PhantomData) }', src/libcore/result.rs:1189:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- exceptions::test::test_check_exception stdout ----
thread 'exceptions::test::test_check_exception' panicked at 'Can not import module: socket: PyErr { type: Py(0x10b9fa4f0, PhantomData) }', src/libcore/result.rs:1189:5


failures:
    buffer::test::test_array_buffer
    exceptions::test::test_check_exception
```

The issue is caused by missing Python library symbols in the final linked rust executable.  For example, the failure in `test_check_exception` happens on an `import socket`.  The Python `socket` module imports a C extension module `_socket` that references `_PyTime_AsMilliseconds`, which is missing:

```
Traceback (most recent call last):
  File "/Users/david/.pyenv/versions/3.7.6/lib/python3.7/socket.py", line 49, in <module>
    import _socket
ImportError: dlopen(/Users/david/.pyenv/versions/3.7.6/lib/python3.7/lib-dynload/_socket.cpython-37m-darwin.so, 2): Symbol not found: __PyTime_AsMilliseconds
  Referenced from: /Users/david/.pyenv/versions/3.7.6/lib/python3.7/lib-dynload/_socket.cpython-37m-darwin.so
  Expected in: flat namespace
 in /Users/david/.pyenv/versions/3.7.6/lib/python3.7/lib-dynload/_socket.cpython-37m-darwin.so
```
The issue may be related to the fact that the `pyenv` installation only provides a static Python library.  

The issue may be known; it was obliquely mentioned in #704 before being deleted from the original issue report.

### 🌍 Environment

 - Your operating system and version:
macOS 10.15.2

 - Your python version:
3.7.6

 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?:
pyenv 

 - Your rust version (`rustc --version`):
1.41.0-nightly

[...]

PR #763, opened 2020-02-11
Add a note about pyenv in README and the guide

### LINKED ISSUE #762: Importing Python random lead to an error
Hi!

### 🌍 Environment

macOS Mojave 10.14.6
 - Your python version: 

Tested on 3.8.1 and 3.6.10

I use pyenv

rustc 1.43.0-nightly (58b834344 2020-02-05)
 - Are you using the latest pyo3 version? Have you tried using latest master?

0.8.4 and master both have this issue 

### 💥 Reproducing

```rust
fn main(){
    let py = gil.python();
    match py.import("random") {
            Err(e) => e.print(py),
            Ok(x) => {},
        }
}
```

```
  File "/Users/maksim.danilchenko/.pyenv/versions/3.6.10/lib/python3.6/random.py", line 42, in <module>
    from math import log as _log, exp as _exp, pi as _pi, e as _e, ceil as _ceil
ImportError: dlopen(/Users/maksim.danilchenko/.pyenv/versions/3.6.10/lib/python3.6/lib-dynload/math.cpython-36m-darwin.so, 2): Symbol not found: __PyLong_GCD
  Referenced from: /Users/maksim.danilchenko/.pyenv/versions/3.6.10/lib/python3.6/lib-dynload/math.cpython-36m-darwin.so
  Expected in: flat namespace
 in /Users/maksim.danilchenko/.pyenv/versions/3.6.10/lib/python3.6/lib-dynload/math.cpython-36m-darwin.so
```

kngwyu:
Closes #762 and #742 

When using `pyo3` with pyenv, `env PYTHON_CONFIGURE_OPTS="--enable-shared"` is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@
 
 [Rust](http://www.rust-lang.org/) bindings for [Python](https://www.python.org/). This includes running and interacting with Python code from a Rust binary, as well as writing native Python modules.
 
+You can either write a native Python module in Rust, or use Python from a Rust binary.
+
 * User Guide: [stable](https://pyo3.rs) | [master](https://pyo3.rs/master)
 
 * API Documentation: [stable](https://docs.rs/pyo3/) |  [master](https://pyo3.rs/master/doc)
@@ -22,16 +24,29 @@ If you have never used nightly Rust, the official guide has
 [a great section](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html#rustup-and-the-role-of-rust-nightly)
 about installing it.
 
-PyPy is also supported (via cpyext) for Python 3.5 only, targeted PyPy version is 7.0.0.
-Please refer to the guide for installation instruction against PyPy.
+### OS dependent packages
+Basically, `Python` is only what you need.
+However, on some OSs, you need to install some additional packages including Python header files.
 
-You can either write a native Python module in Rust, or use Python from a Rust binary.
-
-However, on some OSs, you need some additional packages. E.g. if you are on *Ubuntu 18.04*, please run
+E.g. if you are on *Ubuntu 18.04*, please run
 
 ```bash
-sudo apt install python3-dev python-dev
+sudo apt install python3-dev
 ```
+
+### pyenv
+[pyenv](https://github.com/pyenv/pyenv) is a major Python version manager.
+
+To use PyO3 with pyenv, please set `--enable--shared` when builiding CPython.
+```bash
+$ env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.7.5
+```
+You can also refer to [pyenv wiki](https://github.com/pyenv/pyenv/wiki#how-to-build-cpython-with---enable-shared).
+
+### PyPy
+
+PyPy is also supported (via cpyext) for Python 3.5 only, targeted PyPy version is 7.0.0.
+Please refer to the guide for installation instruction against PyPy.
 
 ## Using Rust from Python
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ If you have never used nightly Rust, the official guide has
 about installing it.
 
 ### OS dependent packages
-Basically, `Python` is only what you need.
+Basically, you only need `Python`.
 However, on some OSs, you need to install some additional packages including Python header files.
 
 E.g. if you are on *Ubuntu 18.04*, please run
--- a/guide/src/get_started.md
+++ b/guide/src/get_started.md
@@ -19,7 +19,7 @@ If you have never used nightly Rust, the official guide has
 about installing it.
 
 ### OS dependent packages
-Basically, `Python` is only what you need.
+Basically, you only need `Python`.
 However, on some OSs, you need to install some additional packages including Python header files.
 
 E.g. if you are on *Ubuntu 18.04*, please run
--- a/guide/src/get_started.md
+++ b/guide/src/get_started.md
@@ -2,6 +2,8 @@
 
 [Rust](http://www.rust-lang.org/) bindings for [Python](https://www.python.org/). This includes running and interacting with Python code from a Rust binary, as well as writing native Python modules.
 
+You can either write a native Python module in Rust, or use Python from a Rust binary.
+
 * User Guide: [stable](https://pyo3.rs) | [master](https://pyo3.rs/master)
 
 * API Documentation: [master](https://pyo3.rs/master/doc)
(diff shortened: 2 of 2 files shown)


kngwyu:
But... how should we do when we want to use a static link?

kngwyu:
Here's my experiments note(on Arch Linux):
1. With `pyenv install 3.7.6`
I got
```
---- buffer::test::test_array_buffer stdout ----
thread 'buffer::test::test_array_buffer' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: Py(0x564b5afc12e0, PhantomData) }', src/buffer.rs:712:21

---- exceptions::test::test_check_exception stdout ----
thread 'exceptions::test::test_check_exception' panicked at 'Can not import module: socket: PyErr { type: Py(0x564b5afc12e0, PhantomData) }', src/exceptions.rs:360:5


failures:
    buffer::test::test_array_buffer
    **exceptions::test::test_check_exception**
```
2. With `env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.7.6`
I got 
```
/home/yu/Programs/PyO3/target/debug/deps/pyo3-1aaccc8eaf7dd9d3: error while loading shared libraries: libpython3.7m.so.1.0: cannot open shared object file: No such file or directory
error: test failed, to rerun pass '--lib'
```

:thinking:

README.md
```diff
@@ -22,16 +24,29 @@ If you have never used nightly Rust, the official guide has
 [a great section](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html#rustup-and-the-role-of-rust-nightly)
 about installing it.
 
-PyPy is also supported (via cpyext) for Python 3.5 only, targeted PyPy version is 7.0.0.
-Please refer to the guide for installation instruction against PyPy.
+### OS dependent packages
+Basically, `Python` is only what you need.
```
Alexander-N:
```suggestion
Basically, you only need `Python`.
```

guide/src/get_started.md
```diff
@@ -16,16 +18,29 @@ If you have never used nightly Rust, the official guide has
 [a great section](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html#rustup-and-the-role-of-rust-nightly)
 about installing it.
 
-PyPy is also supported (via cpyext) for Python 3.5 only, targeted PyPy version is 7.0.0.
-Please refer to the [pypy section](https://pyo3.rs/master/pypy.html).
+### OS dependent packages
+Basically, `Python` is only what you need.
```
Alexander-N:
```suggestion
Basically, you only need `Python`.
```

kngwyu:
@Alexander-N 
Thank you for suggestions!

README.md
```diff
@@ -22,16 +24,29 @@ If you have never used nightly Rust, the official guide has
...
+
+```bash
+sudo apt install python3-dev
+```
 
-However, on some OSs, you need some additional packages. E.g. if you are on *Ubuntu 18.04*, please run
+### pyenv
+[pyenv](https://github.com/pyenv/pyenv) is a major Python version manager.
 
+To use PyO3 with pyenv, please set `--enable--shared` when builiding CPython.
```
konstin: Is this necessary on all platforms or only on mac? I'm using pyenv on ubuntu and I do not remember setting that option.

kngwyu: I don't have a macOS machine, but weirdly, Ubuntu 18.04 + Python 3.7.6(from pyenv) works but Arch Linux + Python 3.7.6 doesn't work.
There's certainly some OS-specific issue, but I still haven't detect it :cold_sweat: :

konstin: I guess in this it's best to just document this as "you might `--enable--shared` on some linux distributions"

davidhewitt:
FWIW I took a quick look at this today, and I think it might be a general issue when linking against a static Python lib, not just `pyenv`.

As far as I can tell, when you link an executable on unix against a static Python, only the symbols which the executable needs are included. But to run arbitrary python code when embedding the python interpreter in pyo3, all the Python interpreter symbols need to be in the final executable.

Because the unused symbols are stripped out, this leads to the kind of failures seen on #742 and #762 when loading some other `.so` that needs the missing symbol fails.

The correct fix might be we might have to disable certain functionality when linking statically. Or maybe there is a way to force all the symbols from `libpython` to be included in the final executable?

I've asked at https://users.rust-lang.org/t/embed-whole-python-interpreter-using-static-linking/42509 , no response yet...

nfwvogt:
On the topic of this being a general issue when linking against a static Python lib, I think I see the same issue with a conda environment on linux. ("Linux Mint 19.3", conda version : 4.8.2, conda-build version : 3.18.9, python 3.7.4.final.0)
Normal python modules can be imported just fine, but everything with a compiled component panics (including numpy).

I could not get any combination of force_load options to work for static linking.
However as far as I understand this  anaconda [discussion](https://github.com/conda-forge/python-feedstock/issues/222) it is built without --enable-shared but ships with a shared libpython.so anyway.

It seems I could get it to work using that libpython.so.
When overwriting `config.shared` in the pyo3 to true even though  `Py_ENABLE_SHARED == False` a minimal test builds fine and can import numpy after setting the LD_LIBRARY_PATH to the anaconda directory.

Not sure if that workaround is known or obvious, but at least I couldn't find it anywhere.

Hizoul:
> When overwriting `config.shared` in the pyo3 to true even though `Py_ENABLE_SHARED == False` a minimal test builds fine and can import numpy after setting the LD_LIBRARY_PATH to the anaconda directory.

Can confirm this is true. It even loads tensorflow so definitely a "working" solution.

> Not sure if that workaround is known or obvious, but at least I couldn't find it anywhere.

I didn't know it and it's not obvious, so thank you very much for sharing this!
Pyo3 always runs instantly on my usual setups. But I had to setup my code on a system where I don't have root priviliges and outdated python as well as outdated pythonlibs so I was forced to use conda.
Even when running `conda activate myenv` with everything installed I still had to do the following to get everything to work:
1. Create libpython.so in the miniconda/lib directory because it was missing `ln -s libpython3.8.so.1.0 libpython.so`
2. Set environment variables so pyo3 uses condas python stuff and not conflicting already installed outdated system libraries
```sh
export PYTHON_SYS_EXECUTABLE=python3.8
export PYO3_CROSS_INCLUDE_DIR=$CONDA_PREFIX/include
export PYO3_CROSS_LIB_DIR=$CONDA_PREFIX/lib
export PYTHONPATH=$CONDA_PREFIX/lib/python3.8
```
3. apply the build.rs workaround in pyo3/build.rs suggested by @nfwvogt

aviramha:
is this still relevant?

davidhewitt:
I don't think we will merge this PR in this form. However, there is definitely scope to have a new section in the guide about "environment setup" which describes how to install Rust and Python for new users on a range of OSes, and macOS users would benefit from having this information about pyenv in that section.

aviramha:
Should we maybe close this then and create a new issue?

davidhewitt:
Created #2017

PR #764, opened 2020-02-12
Add mocpy to the list of examples using pyo3

bmatthieu3:
MOCpy is an astronomical Python library used for describing arbitrary
coverage regions on the sky unit sphere. It relies on the HEALPix tesselation
scheme and because of this, it is very easy and fast to compute the
intersection/union... between coverage regions!

Please check the documentation of mocpy [here](https://cds-astro.github.io/mocpy/) if you are curious.


kngwyu:
Thank you!
I'm glad to see PyO3 is used for scientific computing :smiley:

PR #767, opened 2020-02-14
(docs update) Focusing on FromPy rather than IntoPy

gilescope:
(as you get IntoPy for free)

Also try and point out that intoPy doesn't mean it's necessarily going to convert
something into a python object.

### DIFF
--- a/guide/src/conversions.md
+++ b/guide/src/conversions.md
@@ -100,9 +100,11 @@ fn main() {
 }
 ```
 
-## `IntoPy<T>`
+## `FromPy<T>` and `IntoPy<T>`
 
-Many conversions in PyO3 can't use `std::convert::Into` because they need a GIL token. That's why the `IntoPy<T>` trait offers an `into_py` method that works just like `into`, except for taking a `Python<'_>` argument.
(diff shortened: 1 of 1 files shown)


gilescope:
(If I've got anything wrong please wade in!) Is there any way we can make things clearer? I just want to make sure the next person doesn't thing IntoPy will definitely return a py object.

davidhewitt:
This documentation looks good to me, thanks. I do agree that the naming of `FromPy` and `IntoPy` is potentially very confusing. I haven't at this point thought too hard about what a better design would be.

(E.g. I sort of wonder whether `FromPy` shouldn't exist, though I see why it does for symmetry with the types from `std`.)

kngwyu:
Thank you!

kngwyu:
> I sort of wonder whether FromPy shouldn't exist, though I see why it does for symmetry with the types from std

We have various conversion traits.
1. Convert Rust object to Python object (Rust<->Python object). `FromPyObject`, `FromPyPointer`, `FromPy`
2. `PyObject` -> `&'py PyAny` conversion. `AsPyRef`
3. Downcast(any to dict, list or so). `PyTryFrom`

I don't think `FromPy` is an ideal name, but I think `PyTryFrom` is more confusing and should fix it as priority.

PR #768, opened 2020-02-14
We need nightly, so make that explicit.

gilescope:
Added a rust-toolchain file so people don't have to type `rustup default nightly` - instead `cargo build` just works.


gilescope:
Ok, sorry to bother. Makes sense.

Issue #749, opened 2020-01-29
Unsoundness: `&T` and `&mut T` to same object


davidhewitt:
It is currently possible with pyo3 to get `&mut T` and `&T` to the same object simultaneously.

This is trivially demonstrated with the following code:

    #[pyclass]
    struct Foo {
        value: i32
    }

    #[pyfunction]
    fn edit_foo(a: &mut Foo, b: &Foo) {
        let start = b.value;
        a.value += 1;
        assert_eq!(b.value, start);
    }

Because of the borrow rules the Rust compiler is safe to assume that `a` and `b` are not borrowing the same object. So it may do any number of optimisations to the `assert_eq!` line using the assumption that `b.value` is never modified.

This test can be written which breaks these rules by using the same python object for both arguments:

    #[test]
    fn test_edit_foo() {
        use pyo3::types::IntoPyDict;

        let gil = Python::acquire_gil();
        let py = gil.python();
        let foo = Py::new(py, Foo { value: 0 }).unwrap().to_object(py);

        let locals = [
            ("foo", foo),
            ("edit_foo", wrap_pyfunction!(edit_foo)(py)),
        ].into_py_dict(py);

        py.run("edit_foo(foo, foo)", None, Some(&locals)).unwrap();
    }

This is clearly UB. With `cargo test`, the `assert_eq!` check panics and the test fails. With `cargo test --release`, the test passes, presumably because some optimisation has been applied.

A solution is to implement #342 to guard against this unsoundness.

If a user attempted to call `edit_foo(foo, foo)` from Python, the only choice I see is to raise a Python exception reading something like "`TypeError: Attempted to access a Foo object mutably while also accessing it immutably`".

I pushed this test on a branch; you can see the test failure here: https://github.com/davidhewitt/pyo3/runs/414604743?check_suite_focus=true

kngwyu:
Now we have `PyClassShell` so we can add a borrow flag to it.
Then `PyClassCell` may a prefferable name, though...

davidhewitt:
What about calling it `PyCell`? Even shorter name but still captures the intent.

konstin:
I think this another case of #342

kngwyu:
After first 1 hour trying to resolve this... 
Using `RefCell`(or our customized borrow checker) is OK, but current `PyTryFrom` doesn't fit this change. Maybe fixing `PyTryFrom` is much more difficult than borrow checking.

davidhewitt:
[...]

Issue #738, opened 2020-01-18
AsPyRef is stripped down, not in changelog


birkenfeld:
The `AsPyRef` trait was stripped down considerably; methods `with`, `with_mut`, `into_py`, `into_mut_py` are gone.

The changelog should mention how to rewrite code using them, or if removal was an oversight they should be reinstated.

kngwyu:
Thank you.
Now I'm not sure *why they exist* so need to check them. 
But anyway I will fix it until the non-alpha release.

birkenfeld:
I was using them to access instances of a Rust object that were behind a `Py<T>` pointer.

If there is better or newer way, let me know :)

birkenfeld:
Any news on this?

davidhewitt:
There's a soundness problem; because `Py<T>` is a shared pointer, we need to be careful about getting `&mut T` from a `Py<T>`.

I think in the we really need to have `RefCell`-like semantics, where `Py<T>` will have `.borrow()`, `.borrow_mut()`, `.try_borrow()` and `.try_borrow_mut()` methods.

kngwyu:
Just a reference for readers: https://docs.rs/pyo3/0.8.5/pyo3/trait.AsPyRef.html
@birkenfeld 
I'm OK to retrieve `with` and `with_mut`. Are they sufficient?
`into_py` has a confusing name and I think it should be under another trait.

@davidhewitt 
`Py<T>` can handle all Python pointers(including `PyDict` or so) and I'd rather enhance `PyClassShell` in that direction.

birkenfeld:
@davidhewitt you're right; this cannot be allowed like it was. A RefCell like functionality with runtime checking would be very nice.

@kngwyu `with` is probably also served by `as_ref`? `with_mut` needs a safer alternative, "somehow" :)

kngwyu:
Now `AsPyRef::as_ref` is changed to return `&PyCell`.
For PyCell, please see [the guide](https://pyo3.rs/master/class.html#pycell-and-interior-mutability).

birkenfeld:
Great, thanks!

Issue #687, opened 2019-12-09
A soundness issue with GIL-bound references


nagisa:
I think @pganssle’s comment in https://github.com/PyO3/pyo3/issues/679 ultimately reveals a soundness issue in PyO3. 

Remember, that Rust has very strict invariants for its references:

* There must ever be one and only one live `&mut` reference to the same object (not too relevant to as, but do note that object here is *not* the python object);
* There may be multiple `&` references to the same object but only the `UnsafeCell` fields (interior mutability) of the object may change while there any live references to the object.

The existence of `&PyNativeType` potentially violates the 2nd invariant as soon as the GIL is unlocked as part of some FFI or `allow_threads(|| ...)` call. Consider this example and potential way things could go wrong (presume some super-optimised implementation of the functions used here)

```rust
let gil = Python::acquire_gil();
let list = PyList::new(gil.python(), [1, 2, 3]);
allow_threads(|| { /*
    Some code here mutates the list object, possibly reallocating or clearing it
    after they obtain their own GIL lock.
*/ } );
// compiler is allowed to reorder this statement to before `allow_threads`
// or constant propagate the statically known length of 3 into here.
//
// After all the invariant does not permit for the underlying data to change
// while a reference to the list is live.
let length = list.len();
// can optimise out the bounds check to use `length` variable
// rather than reading length again.
list[2] // => OOB access
```

We already solved an soundness bug by making these references `!Send` and thus preventing invalid `Python::assume_gil_acquired()` calls, but what we really need is a prohibition to unlock the GIL at all. At least while these references exist.

An alternative, and more pragmatic, option would be to make all these types have interior mutability by adding an `UnsafeCell` into them.

nagisa:
NB: I don’t actually know if `UnsafeCell` + whatever GIL setup we have currently is sufficient and it seems hard to prove too.

programmerjake:
A related possible issue with PyO3 giving out `&mut` references to objects:
I didn't take the time to check if this actually works (should be run with compiler optimizations disabled to avoid having code optimized out), but here goes:

```rust
#[pyclass]
struct MyClass {
    lock: std::Mutex<String>,
}

fn my_fn(value: &PyAny) -> PyResult<()> {
    let mut_ref1: &mut MyClass = FromPyObject::extract(value)?;
    let mut_ref2: &mut MyClass = FromPyObject::extract(value)?;
    assert!(std::ptr::eq(mut_ref1, mut_ref2),
[...]

Issue #381, opened 2019-03-02
Inheritance example does not compile


Alexander-N:
### 🌍 Environment
 - Your operating system and version: Ubuntu 18.10
 - Your python version: 3.6.7
 - How did you install python (e.g. apt or pyenv)? apt
 - Did you use a virtualenv?: no
 - Your rust version (`rustc --version`): rustc 1.34.0-nightly (aadbc459b 2019-02-23)
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` yes

### 💥 Reproducing

The following code from https://pyo3.rs/master/class.html#inheritance

```rust
#![allow(unused_variables)]
fn main() {
   use pyo3::prelude::*;
   use pyo3::PyRawObject;
   #[pyclass]
   struct BaseClass {
      val1: usize,
   }

   #[pymethods]
   impl BaseClass {
      #[new]
      fn new(obj: &PyRawObject) {
         obj.init(BaseClass { val1: 10 });
      }

      pub fn method(&self) -> PyResult<()> {
         Ok(())
      }
   }

   #[pyclass(extends=BaseClass)]
   struct SubClass {
      val2: usize,
   }

   #[pymethods]
   impl SubClass {
      #[new]
      fn new(obj: &PyRawObject) {
         obj.init(SubClass { val2: 10 });
         BaseClass::new(obj);
      }

      fn method2(&self) -> PyResult<()> {
         self.get_base().method()
      }
   }
}
```
gives me
```
error[E0599]: no method named `get_base` found for type `&main::SubClass` in the current scope
  --> src/main.rs:93:15
   |
93 |          self.get_base().method()
   |               ^^^^^^^^
   |
   = note: the method `get_base` exists but the following trait bounds were not satisfied:
           `&main::SubClass : pyo3::objectprotocol::ObjectProtocol`
           `main::SubClass : pyo3::objectprotocol::ObjectProtocol`
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following trait defines an item `get_base`, perhaps you need to implement it:
           candidate #1: `pyo3::objectprotocol::ObjectProtocol`
```

kngwyu:
Looks current implementation is only for `T: PyObjectNativeType + AsPyPointer` so we can't actually use it :thinking:

Alexander-N:
`get_base` was removed in #770.

PR #770, opened 2020-02-16
Rename PyClassShell with `PyCell` and do mutability checking

### LINKED ISSUE #342: Use RefCell to enforce aliasing rules
Currently pyo3 allows to break rust's aliasing guarantees, i.e. you can get two mutable references to the same object. This is not a big issue because you can only have those two references on the same thread, but this still needs to be fixed eventually. See [this capter in wasm bindgen's guide](https://rustwasm.github.io/wasm-bindgen/contributing/design/exporting-rust-struct.html) for a good write-up on the why and how.

kngwyu:
Resolves #342 #749 #738 
## TODO
- [x] Fix documents
- [x] Fix documentation in th guide
- [x] Add documentation about interior mutability.
- [x] Now protocol implementations uses `try_borrow_unguarded` so fix them to use `try_borrow`
- [x]  Add documents to all new APIs
- [x] Rethink `Py::new`
  - I just marked it as soft-duplicated
- [x] Add tests for inheriting classes with weakref/dict slots

# TL; DR
Now we can take multiple *mutable* references of `PyClass`. 
This PR fixes it by runtime reference counting like [RefCell](https://doc.rust-lang.org/std/cell/struct.RefCell.html).
We can use `PyCell` like RefCell. We can get `PyRef` and `PyRefMut` from `&PyCell`, with interior mutability checking.
The implementation is done via overhauling our `PyClass` and `PyClassShell`, resulting in renaming `PyClassShell` with `PyCell`.

## Notation
`BorrowFlag` it the reference counter that `PyCell` has. It can also represent *mutably borrowed*.

## Implementation
So, what's the problem of implementing mutability checking for `PyClass`es?
The answer is **inheritation**.
Our PyClass can inherit other class by  `#[pyclass(extends=OtherClass)]` syntax.
In such cases, we should share `BorrowFlag` between child and parent classes.
So the following object layout is required:
```rust
pub struct PyCell {
    ob_base: ffi::PyObject,
    borrow_flag: BorrowFlag,
    value1: ParentClass,
    value2: ChildClass,
}
```
It is somewhat difficult since `PyCell` is defined recursively using the layout of base object, but I resolved this by using 
```rust
#[repr(C)]
pub struct PyCellBase<T: PyTypeInfo> {
    ob_base: T::Layout,
    borrow_flag: Cell<BorrowFlag>,
}
```
as the farthest baseclasses for all `PyClass`es.
This needs some helper types like `derive_utils::PyBaseTypeUtils`.

## Changed APIs
- `PyTryFrom`: Now returns `&PyCell<T>` instead of `&T` and `&mut T`
- `FromPyPointer`: Same as above.
- `AsPyRef`: Now has an associated type `Target`. Returns `&PyCell<T>` from `Py<T>`.
(TODO: make the complete list)

## New APIs
- `PyRef`
- `PyRefMut`
- (hidden) `PyDowncastImpl`
(TODO: make the complete list)


davidhewitt:
Awesome! There's a lot here and I'll try to give it all a thorough review, though this will take me a few days...

src/conversion.rs
```diff
@@ -467,63 +408,71 @@ impl FromPy<()> for Py<PyTuple> {
 
 /// Raw level conversion between `*mut ffi::PyObject` and PyO3 types.
 pub unsafe trait FromPyPointer<'p>: Sized {
```
davidhewitt: There are so many methods in this trait. Is there a reason why we don't just have 

    unsafe fn from_owned_ptr(py: Python<'p>, ptr: *mut ffi::PyObject) -> PyResult<&'p Self>;
    unsafe fn from_borrowed_ptr(py: Python<'p>, ptr: *mut ffi::PyObject) -> PyResult<&'p Self>;

As the only trait method. And then just let users choose if they want to `.ok()`, `.unwrap()`, or `.expect()` this, instead of all the `_or_opt`, `_or_panic` etc.

This would be quite a big change, so maybe worth doing as a separate PR.

src/derive_utils.rs
```diff
@@ -218,3 +218,18 @@ impl GetPropertyValue for PyObject {
         self.clone_ref(py)
     }
 }
+
+/// Utitlities for basetype
```
davidhewitt:
```suggestion
/// Utilities for basetype
```

src/instance.rs
```diff
@@ -39,11 +39,10 @@ impl<T> Py<T> {
     pub fn new(py: Python, value: impl Into<PyClassInitializer<T>>) -> PyResult<Py<T>>
```
davidhewitt: I wonder if we should remove this method and instead always encourage use of `PyCell::new()`.

Or maybe the argument type should be `impl IntoPy<PyObject>`, so that e.g. `Py<PyTuple>` can also be created with `Py::new()`.

src/instance.rs
```diff
@@ -119,15 +118,17 @@ impl<T> Py<T> {
...
+pub trait AsPyRef: Sized {
+    type Target;
     /// Return reference to object.
-    fn as_ref(&self, py: Python) -> &T;
+    fn as_ref(&self, py: Python<'_>) -> &Self::Target;
 }
 
-impl<T: PyTypeInfo> AsPyRef<T> for Py<T> {
-    fn as_ref(&self, _py: Python) -> &T {
+impl<'p, T: PyClass> AsPyRef for Py<T> {
```
davidhewitt: Maybe also need `impl<'p, T: AsPyPointer> AsPyRef for Py<T>` ?

src/objectprotocol.rs
```diff
@@ -176,15 +171,10 @@ pub trait ObjectProtocol {
...
     where
-        Self: PyTypeInfo;
-
-    /// Gets the Python base object for this object.
-
-    fn get_mut_base(&mut self) -> &mut <Self as PyTypeInfo>::BaseType
-    where
-        Self: PyTypeInfo;
+        Self: PyTypeInfo,
+        <Self as PyTypeInfo>::BaseType: FromPyPointer<'py>;
```
davidhewitt: Might be able to do `for <'py> FromPyPointer<'py>` instead of needing to add `'py` everywhere?

src/pycell.rs
```diff
@@ -0,0 +1,446 @@
...
+use std::cell::{Cell, UnsafeCell};
+use std::fmt;
+use std::mem::ManuallyDrop;
+use std::ops::{Deref, DerefMut};
+
+#[doc(hidden)]
+#[repr(C)]
+pub struct PyCellBase<T: PyTypeInfo> {
+    ob_base: T::Layout,
+    borrow_flag: Cell<BorrowFlag>,
```
davidhewitt: Is it worth making an assertion that this doesn't need to be thread-safe (i.e. not `AtomicXyz`) because the of the GIL?

src/type_object.rs
```diff
@@ -1,38 +1,28 @@
...
+/// E.g., `PyCell` is a concrete representaion of all `pyclass`es, and `ffi::PyObject`
 /// is of `PyAny`.
 ///
 /// This trait is intended to be used internally.
-pub trait PyObjectLayout<T: PyTypeInfo> {
+pub unsafe trait PyObjectLayout<T: PyTypeInfo> {
     const IS_NATIVE_TYPE: bool = true;
-
-    fn get_super_or(&mut self) -> Option<&mut <T::BaseType as PyTypeInfo>::ConcreteLayout> {
+    fn get_super_or(&mut self) -> Option<&mut T::BaseLayout> {
```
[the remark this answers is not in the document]
kngwyu: This method is for initialization and needs `mut`.

src/class/macros.rs
```diff
@@ -415,32 +389,51 @@ macro_rules! py_func_set_del {
...
             }
         }
         Some(wrap::<$generic>)
     }};
 }
+
+// Utitlities for extract arguments + call method for PyCell<T>
+macro_rules! call_ref {
+    ($slf: expr, $fn: ident $(; $args: expr)*) => {
+        match $slf.try_borrow_unguarded() {
```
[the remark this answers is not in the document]
kngwyu: I tried to change this to `try_borrow`, but it's difficult because of a lifetime problem.
In detail:
Now this macro is expanded to
```rust
match arg1.extract() {
    Ok(arg1) => match arg2.extract() => 
       Ok(arg2) => match slf.extract() {... }
       Err(e) => ...,
    Err(e) => ...,
}
```
and compiler claims that slf lives shorter than args.
So re-ordering this to
```rust
match slf.extract() {
    Ok(slf) => match arg1.extract() => { ... },
    Err(e) => ...,
}
```
works for most cases.., but still doesn't work well with `CallbackConverter`.

src/instance.rs
kngwyu: This would cause `conflicting implementation` error.
I placed implementations for native types in `types/mod.rs`.

src/objectprotocol.rs
kngwyu: Good catch, thanks :+1:

kngwyu:
Thank you for the lots of reviews!
Enjoying addressing them :smiley: 

> Can you explain in a bit more detail why this has to be the case? I would think that as the Rust objects are separate, it should be safe to have &mut T::Base and &T simultaneously.

If we allow having `PyRefMut<T>` and `PyRefMut<T::BaseType>` safely, `get_super` APIs should be (roughly) as follows:
```rust
impl<T> PyCell<T> {
    pub fn get_super(&self) -> &PyCell<T::BaseType> { ... }
}
impl<'p, T> PyRef<'p, T>
{
    pub fn get_super(&self) ->  &PyCell<T::BaseType> { ... }
}
...
```
I think this has some downsides:
- Simply not convenient. I feel `PyRefMut<T> -> PyRefMut<T::BaseType>` style API is easier to use.
- If `T::BaseType` have `weakref` or `dict` flag we can't return `PyCell<T::BaseType>`(just an implementation detail, but can be hard to fix)

src/instance.rs
davidhewitt: Ah yes, of course. Then how about 

```
impl<'p, T: PyTypeInfo> AsPyRef for Py<T> {
    type Target = <T as PyTypeInfo>::Layout;
    ...
}
```

kngwyu: Layout is `ffi::PyObject` or so for native types.
It's not convenient to expose them.

davidhewitt: Ah - I thought their layout was `PyAny`, my bad. Ok let's stick with the current design :)

src/pycell.rs
kngwyu: This is already not `!Send` nor `Sync`.
What kind of assertion do you have in mind?

src/instance.rs
kngwyu: :+1:

davidhewitt:
Ah, that makes sense, thanks. Let's see what the gitter thread prefers. I'm warming to your design. It would be strange to try to `.super()` an object and then get a `.PyCell`.

src/types/mod.rs
```diff
@@ -56,11 +56,30 @@ macro_rules! pyobject_native_type_named (
...
 #[macro_export]
 macro_rules! pyobject_native_type {
     ($name: ty, $layout: path, $typeobject: expr, $module: expr, $checkfunction: path $(,$type_param: ident)*) => {
-        impl $crate::type_object::PyObjectLayout<$name> for $layout {}
+        impl_layout!($name, $layout);
         impl $crate::type_object::PyObjectSizedLayout<$name> for $layout {}
+        impl $crate::derive_utils::PyBaseTypeUtils for $name {
+            type Dict = $crate::pyclass_slots::PyClassDummySlot;
+            type WeakRef = $crate::pyclass_slots::PyClassDummySlot;
+            type Layout = $crate::pycell::PyCellBase<$name>;
```
davidhewitt: Do native types need borrow checking? I would have thought that they should just continue to be accessed with `&PyAny`.

davidhewitt:
An idea: what if we had a type `PyBorrowed<'a, T>` ?

Then the API for `get_super` could be something like

```
impl<T> PyBorrowed<`_, T> {
    fn get_super(&self) -> PyBorrowed<T::BaseLayout>
}
```

Along with:
- `PyBorrowed<T>` derefs to `T`
- `PyRef<T>` derefs to `PyBorrowed<T>`
- `PyBorrowed<PyDict>` etc. may also be useful for #679, if we make `PyDict<'py, T>` an owned pointer.

I wrote this in a rush, sorry if it doesn't make much sense, or it has other problems!

src/pycell.rs
davidhewitt: Even if it's not `Send` or `Sync` in Rust, the python interpreter will happily share this data between threads. Perhaps just a comment explaining why `Cell` is enough?

src/types/mod.rs
kngwyu: It's confusing but `<T as PyBaseTypeUtils>::Layout` means *the layout as a baseclass of PyClass`.
Maybe we need a clearer name.

src/pycell.rs
kngwyu: I see.

kngwyu:
> An idea: what if we had a type PyBorrowed<'a, T> ?

Sounds good for the future design, but I don't think we should add it now.

davidhewitt:
Excited to see all the progress on this. I'm hoping to find time to give this another review in the next day or two. :+1:

birkenfeld:
I can't contribute at the moment, but just wanted to say that this is *very* cool. Making the API safe is very much appreciated!

kngwyu:
@Alexander-N 
Could you please take a glance at class.md and documents for PyCell?
Thanks.

src/pycell.rs
```diff
@@ -0,0 +1,678 @@
...
+    }
+    const fn decrement(self) -> Self {
+        Self(self.0 - 1)
+    }
+}
+
+/// An error returned by [`PyCell::try_borrow`](struct.PyCell.html#method.try_borrow).
+pub struct PyBorrowError {
+    _private: (),
+}
```
davidhewitt: I wonder whether we should capture the Python pointer in here, so that we can report errors later:

```suggestion
pub struct PyBorrowError<'a> {
    ptr: &'a PyAny
}
```

Python users might like, for example, an error message more like:

```
Cannot borrow this object of type `Foo`, as it is already mutably borrowed.
```

src/pyclass.rs
```diff
@@ -71,185 +66,20 @@ pub unsafe fn tp_free_fallback(obj: *mut ffi::PyObject) {
...
-        })
-    }
-    unsafe fn from_borrowed_ptr_or_opt(py: Python<'p>, ptr: *mut ffi::PyObject) -> Option<Self> {
-        NonNull::new(ptr).map(|p| {
-            &mut *(gil::register_borrowed(py, p).as_ptr() as *const _ as *mut PyClassShell<T>)
-        })
-    }
+    /// The closest native ancestor. This is `PyAny` by default, and when you declare
+    /// `#[pyclass(extends=PyList)]`, it's `PyList`.
+    type BaseNativeType: PyTypeInfo + PyNativeType;
```
davidhewitt: I thought we can't safely extend from `PyTuple` or `PyList` yet, so might want to change this doc?

src/type_object.rs
```diff
@@ -1,46 +1,49 @@
...
+/// E.g., `PyCell` is a concrete representaion of all `pyclass`es, and `ffi::PyObject`
 /// is of `PyAny`.
 ///
 /// This trait is intended to be used internally.
-pub trait PyObjectLayout<T: PyTypeInfo> {
+pub unsafe trait PyLayout<T: PyTypeInfo> {
     const IS_NATIVE_TYPE: bool = true;
-
-    fn get_super_or(&mut self) -> Option<&mut <T::BaseType as PyTypeInfo>::ConcreteLayout> {
+    fn get_super_or(&mut self) -> Option<&mut T::BaseLayout> {
```
davidhewitt: I'd like to suggest just calling this function `super` - I don't think `get` or `or` add anything here.

src/type_object.rs
```diff
@@ -1,46 +1,49 @@
...
 /// , in addition that `T` is a concrete representaion of `U`.
+pub trait PySizedLayout<T: PyTypeInfo>: PyLayout<T> + Sized {}
+
+/// Marker type indicates that `Self` can be a base layout of `PyClass`.
+///
+/// # Safety
 ///
-/// `pyclass`es need this trait for their base class.
-pub trait PyObjectSizedLayout<T: PyTypeInfo>: PyObjectLayout<T> + Sized {}
+/// Self should be layouted as follows:
```
davidhewitt:
```suggestion
/// Self should be laid out as follows:
```

src/pyclass.rs
kngwyu: Oops, good catch, thanks.

src/type_object.rs
kngwyu: `super` is keyword in Rust so we need some prefix or suffix

src/pycell.rs
kngwyu: Hmm... :thinking: 
I think it's difficult since Error should be `'static`.

I'll rethink it when addressing #682

src/instance.rs
```diff
@@ -119,15 +122,47 @@ impl<T> Py<T> {
...
+/// let obj: PyObject = {
+///     let gil = Python::acquire_gil();
+///     let py = gil.python();
+///     py.eval("[]", None, None).unwrap().to_object(py)
+/// };
+/// let gil = Python::acquire_gil();
+/// let py = gil.python();
+/// assert_eq!(obj.as_ref(py).len().unwrap(), 0);  // PyAny implements ObjectProtocol
+/// ```
+pub trait AsPyRef: Sized {
```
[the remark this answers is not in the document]
kngwyu: :+1:

src/type_object.rs
davidhewitt: Ah shoot, good point. How about just `get_super` then?

kngwyu: :+1:

src/instance.rs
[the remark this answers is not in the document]
kngwyu: Hmm but since
- `AsPyRef` is just for `Py<T>` and `PyObject`(and PyObject is actually Py<PyAny>)
- It requires lots of changes
I decided to leave it as is

kngwyu:
@davidhewitt 
Thank you for lots of reviews :smiley: 

> I wonder if we could simplify the code and make PyCell a wrapper around RefCell

`RefCell<T>` is a pair (T, BorrowFlag).
However, to enable `extends=` feature, `PyCell` can be a triplet (Base, Sub, BorrowFlag) or larger tuple. This nature conceptually makes it difficult to use `RefCell` here.

src/instance.rs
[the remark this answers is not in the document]
davidhewitt: I suppose `AsPyRef` will need further changes after #687 anyway

src/conversion.rs
davidhewitt: Opened as #781

guide/src/class.md
```diff
@@ -261,8 +278,9 @@ impl SubClass {
        (SubClass{ val2: 15}, BaseClass::new())
    }
 
-   fn method2(self_: &PyClassShell<Self>) -> PyResult<usize> {
-      self_.get_super().method().map(|x| x * self_.val2)
+   fn method2(self_: PyRef<Self>) -> PyResult<usize> {
+       let super_ = self_.as_ref();  // Get &BaseClass
```
Alexander-N: For me this is surprising, while `get_super` was very clear.

guide/src/class.md
```diff
@@ -279,29 +297,30 @@ impl SubSubClass {
...
 # let gil = Python::acquire_gil();
 # let py = gil.python();
-# let subsub = pyo3::PyClassShell::new_ref(py, SubSubClass::new()).unwrap();
+# let subsub = pyo3::PyCell::new(py, SubSubClass::new()).unwrap();
 # pyo3::py_run!(py, subsub, "assert subsub.method3() == 3000")
 ```
 
 To access the super class, you can use either of these two ways:
-- Use `self_: &PyClassShell<Self>` instead of `self`, and call `get_super()`
+- Use `self_: &PyCell<Self>` instead of `self`, and call `get_super()`
```
Alexander-N: I find it a bit confusing that only `into_super` was used in the example.

guide/src/class.md
```diff
@@ -279,29 +297,30 @@ impl SubSubClass {
...
 - `ObjectProtocol::get_base`
-We recommend `PyClassShell` here, since it makes the context much clearer.
+We recommend `PyCell` here, since it makes the context much clearer.
```
Alexander-N: Maybe we can just remove `get_base` since I remember there was a problem with this method https://github.com/PyO3/pyo3/issues/381

guide/src/class.md
```diff
@@ -761,16 +780,16 @@ struct GCTracked {} // Fails because it does not implement PyGCProtocol
 Iterators can be defined using the
 [`PyIterProtocol`](https://docs.rs/pyo3/latest/pyo3/class/iter/trait.PyIterProtocol.html) trait.
 It includes two methods `__iter__` and `__next__`:
-  * `fn __iter__(slf: &mut PyClassShell<Self>) -> PyResult<impl IntoPy<PyObject>>`
-  * `fn __next__(slf: &mut PyClassShell<Self>) -> PyResult<Option<impl IntoPy<PyObject>>>`
+  * `fn __iter__(slf: PyRefMut<Self>) -> PyResult<impl IntoPy<PyObject>>`
```
Alexander-N: I liked that PyRefMut was replaced with PyClassShell since I find PyRef/PyRefMut unintuitive for reasons mentioned in https://github.com/PyO3/pyo3/issues/356

pyo3-derive-backend/src/method.rs
```diff
@@ -29,12 +29,16 @@ pub enum FnType {
     FnCall,
     FnClass,
     FnStatic,
-    PySelf(syn::TypeReference),
+    // self_: &PyCell<Self>,
+    PySelfRef(syn::TypeReference),
+    // self_: PyRef<Self> or PyRefMut<Self>
```
Alexander-N: Forgotten comments?

guide/src/class.md
kngwyu: It was initially named `as_super` but changed to use `AsRef` since [wasm-bindgen](https://rustwasm.github.io/wasm-bindgen/web-sys/inheritance.html#accessing-parent-classes-using-asref) does so.
I prefer to `as_super` butt, in the long run, it would be beneficial to use the same API convention with a major library.

guide/src/class.md
kngwyu: This was left unchaged... thanks

guide/src/class.md
kngwyu: :+1:

pyo3-derive-backend/src/method.rs
kngwyu: They are comments for developers.
Rewrote as doc comments.

kngwyu:
@Alexander-N 
Thank you for reviews!

Issue #657, opened 2019-10-30
Implement the buffer protocol safely


jimy-byerley:
Hi ! 
I'm using rust as backend for a python module I'm working on, to get high performances on mesh operations. The data is owned by some rust `Vec`s but I would like to expose its content to python, to make possible extend the module with pure python (or at least to get some data points).

So I'm trying to implement the buffer protocol for my structs. I seen [this page in pyo3](https://pyo3.rs/master/doc/pyo3/class/buffer/trait.PyBufferProtocol.html#method.bf_getbuffer). But it relies on pointers so it's unsafe. 
But the buffer protocol is not incompatible with pyo3's safety since i read [in the python doc](https://docs.python.org/3/c-api/typeobj.html#buffer-structs) that the `memoryview` object kept a hook on the viewed object

Is there a way to safely implement the buffer protocol for rust objects exposed by a `#[pyclass]` statement ?

kngwyu:
Sorry, I'm saying about `class/buffer.rs` where `PyBufferProtocol` is.
Though `PyBuffer` also can be changed a bit...(e.g., it can have lifetime `py`)

davidhewitt:
This is now tracked in #1884 as part of the move to `#[pymethods]`.

PR #772, opened 2020-02-20
Buffer protocol updates to support object references, custom release method

andrewwhitehead:
This PR doesn't cover the buffer refactoring mentioned in #657 for example, but it allows for implementors to avoid having the exporter dropped while Python is holding a memoryview referencing it. Leaving as a draft because I'm not sure what the impact would be on existing implementations.

### DIFF
--- a/tests/test_buffer_protocol.rs
+++ b/tests/test_buffer_protocol.rs
@@ -1,33 +1,43 @@
+use pyo3::buffer::PyBuffer;
 use pyo3::class::PyBufferProtocol;
 use pyo3::exceptions::BufferError;
 use pyo3::ffi;
 use pyo3::prelude::*;
 use pyo3::types::IntoPyDict;
+use pyo3::{AsPyPointer, PyClassShell};
 use std::ffi::CStr;
 use std::os::raw::{c_int, c_void};
 use std::ptr;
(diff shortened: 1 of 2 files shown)


kngwyu:
Thank you.

> it allows for implementors to avoid having the exporter dropped while Python is holding a memoryview referencing it. 

So you mean this PR enables `(*view).obj = slf.as_ptr();`?
It's nice :+1: 

The only concern is the conflict with #770, but I can fix it myself.

tests/test_buffer_protocol.rs
```diff
@@ -58,21 +68,68 @@ impl PyBufferProtocol for TestClass {
...
+        assert_eq!(buf.to_vec::<u8>(py).unwrap(), input);
+        drop(instance);
+        buf
+    };
+
+    assert!(!drop_called.load(Ordering::Relaxed));
+
+    {
+        let _py = Python::acquire_gil().python();
+        drop(buf);
```
kngwyu: This test is nice :+1:

andrewwhitehead:
@kngwyu That's great, yes I was also concerned about the conflict with that other PR.

Incidentally, my use case is a memory buffer with a private key which I want to zero out as soon as there's no more references to it, but not before then.

kngwyu:
Thank you!

Issue #776, opened 2020-02-28
PyBuffer::release calls ffi::PyBuffer_Release twice on the same buffer


oconnor663:
The `PyBuffer::release` method calls `ffi::PyBuffer_Release` internally. Because it takes `self` by value, it then immediately drops `self`, and `PyBuffer::drop` calls `ffi::PyBuffer_Release` again. I think this kind of defeats the purpose of `release`, which is allowing the caller to pass in their `py` argument to avoid re-acquiring the GIL. More importantly, it looks like it violates the requirements described in Python's [Buffer Protocol docs](https://docs.python.org/3/c-api/buffer.html):

> Successful calls to `PyObject_GetBuffer()` must be paired with calls to `PyBuffer_Release()`, similar to `malloc()` and `free()`. Thus, after the consumer is done with the buffer, `PyBuffer_Release()` must be called exactly once.

Should this be fixed with `mem::forget` or similar?

davidhewitt:
Thanks for the catch. Did you come across this in a crash or some other repro we can use to make a test?

I would probably suggest just calling `std::mem::drop(self)` in `PyBuffer::release` to be explicit.

oconnor663:
I only saw it reading the code. As far as I can tell, it does not trigger a crash. Looking at the upstream CPython source, [here's what I see](https://github.com/python/cpython/blob/master/Objects/abstract.c#L726-L738):

```c
void
PyBuffer_Release(Py_buffer *view)
{
    PyObject *obj = view->obj;
    PyBufferProcs *pb;
    if (obj == NULL)
        return;
    pb = Py_TYPE(obj)->tp_as_buffer;
    if (pb && pb->bf_releasebuffer)
        pb->bf_releasebuffer(obj, view);
    view->obj = NULL;
    Py_DECREF(obj);
}
```

So it looks like `view->obj` gets nulled out, which would cause the second call to short-circuit rather than doing anything explosive. The Python docs make it sound like we shouldn't rely on this always being the case, but I guess I'd be surprised if they changed it. I have no idea what PyPy does.

Still, it seems like the purpose of `release` is to avoid the performance hit of `Python::acquire_gil()` in the destructor? If we call `mem::drop`, we'll still be paying that cost. Or is that cheaper than I think it is?

oconnor663:
Just put up an example fix. Not sure how I would test this though.

davidhewitt:
Thanks very much for providing the PR. Have followed up on that thread.

PR #777, opened 2020-02-28
avoid calling ffi::PyBuffer_Release twice in PyBuffer::release

oconnor663:
Fixes https://github.com/PyO3/pyo3/issues/776.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -26,7 +26,7 @@ use crate::Python;
 use std::ffi::CStr;
 use std::os::raw;
 use std::pin::Pin;
-use std::{cell, mem, slice};
+use std::{cell, mem, ptr, slice};
 
 /// Allows access to the underlying buffer used by a python object such as `bytes`, `bytearray` or `array.array`.
 // use Pin<Box> because Python expects that the Py_buffer struct has a stable memory address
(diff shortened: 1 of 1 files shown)


oconnor663:
CI failure looks related. I'm trying to repro it.

oconnor663:
Got a repro. (I've found I need to `export RUSTUP_TOOLCHAIN=nightly`, I think because there are temp directories involved, and `rustup override set nightly` isn't sufficient to guarantee that nightly runs. Other folks who use stable as their default will probably run into the same issue. Maybe the tests could be tweaked to explicitly invoke `cargo +nightly`?)

oconnor663:
Oh I'm just being an idiot and leaking the Box.

davidhewitt:
This is a very elegant solution, though I think in practice it's probably sufficient to just replace the whole function implementation with `std::mem::drop(self)`. The cost of "acquiring" the gil is unlikely to be that bad given one must already hold the GIL to call `PyBuffer::release`

oconnor663:
Gotcha, though in that case, why expose the `release` method at all?

davidhewitt:
That's a great question - all of this code was written well before I got involved with pyo3! 😄

davidhewitt:
Let's see what other maintainers say (cc @kngwyu)

kngwyu:
Good catch and nice PR, thank you @oconnor663 !

src/buffer.rs
```diff
@@ -577,10 +577,19 @@ impl PyBuffer {
...
-            let ptr = &*self.0 as *const ffi::Py_buffer as *mut ffi::Py_buffer;
-            ffi::PyBuffer_Release(ptr)
-        };
+            // Next, make the actual PyBuffer_Release call.
+            ffi::PyBuffer_Release(&mut *mdself.0);
+
+            // Finally, drop the contained Pin<Box<_>> in place, to free the
+            // Box memory.
+            let inner: *mut Pin<Box<ffi::Py_buffer>> = &mut mdself.0;
+            ptr::drop_in_place(inner);
```
kngwyu:
```suggestion
            ptr::drop_in_place(&mut mdself.0);
```

kngwyu:
> Gotcha, though in that case, why provide the release method at all?

This function was written before forking under the name of rust-cpython, so I don't know why too.

src/buffer.rs
oconnor663: I did it this way because I'm a little bit afraid of (future) Deref magic. For example, it looks like there's some Deref magic happening in the line `ffi::PyBuffer_Release(&mut *mdself.0)`. I'm worried that someone else might change this type in the future, and if so I _want_ this line to break if the type changes. It's too easy for `drop_in_place` to accidentally become a no-op, for example if you pass it a raw `*mut T`.

kngwyu: OK, let's leave it as is :+1:

PR #778, opened 2020-02-29
Implement `FromPyObject` and `IntoPy<PyObject>` traits for arrays (up to 32)

### LINKED ISSUE #775: `#[new]` should support arrays
Maybe a missing implementation?

```rust
use pyo3::prelude::*;

#[pyclass]
struct Foo {
  array: [i32; 2]
}

#[pymethods]
impl Foo {
  #[new]
  //pub fn new(array: [i32; 2]) -> Self {
  pub fn new(array: &[i32; 2]) -> Self {
    //Self { array }
    Self { array: *array }
  }
}
```

c410-f3r:
Fixes #775.

`Copy + Default` bounds to initialize memory for `copy_to_slice_impl` and avoid unsafe (`MaybeUninit` and `ptr` friends). `Copy` bound is also necessary due to the lack of a "real" `IntoIterator` implementation for arrays.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * `FromPyObject` is now automatically implemented for `T: Clone` pyclasses. [#730](https://github.com/PyO3/pyo3/pull/730)
 * `#[pyo3(get)]` and `#[pyo3(set)]` will now use the Rust doc-comment from the field for the Python property. [#755](https://github.com/PyO3/pyo3/pull/755)
 * `#[setter]` functions may now take an argument of `Pyo3::Python`. [#760](https://github.com/PyO3/pyo3/pull/760)
+* Implement `FromPyObject` and `IntoPy<PyObject>` traits for arrays (up to 32). [#778](https://github.com/PyO3/pyo3/pull/778)
 
 ### Fixed
 
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -229,7 +229,7 @@ mod test {
     use crate::objectprotocol::ObjectProtocol;
     use crate::types::PyList;
     use crate::Python;
-    use crate::{PyTryFrom, PyObject, IntoPy, ToPyObject};
+    use crate::{IntoPy, PyObject, PyTryFrom, ToPyObject};
 
     #[test]
     fn test_new() {
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -178,6 +178,26 @@ where
     }
 }
 
+macro_rules! array_impls {
+    ($($N:expr),+) => {
+        $(
+            impl<T> IntoPy<PyObject> for [T; $N]
+            where
+                T: ToPyObject
+            {
+                fn into_py(self, py: Python) -> PyObject {
+                    self.as_ref().to_object(py)
+                }
+            }
+        )+
+    }
+}
+
+array_impls!(
+    0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25,
+    26, 27, 28, 29, 30, 31, 32
+);
+
 impl<T> ToPyObject for Vec<T>
 where
     T: ToPyObject,
@@ -209,7 +229,7 @@ mod test {
     use crate::objectprotocol::ObjectProtocol;
     use crate::types::PyList;
     use crate::Python;
-    use crate::{PyTryFrom, ToPyObject};
+    use crate::{PyTryFrom, PyObject, IntoPy, ToPyObject};
 
     #[test]
     fn test_new() {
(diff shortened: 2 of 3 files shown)


src/types/list.rs
```diff
@@ -424,3 +464,10 @@ mod test {
         assert_eq!(2, list.get_item(3).extract::<i32>().unwrap());
     }
 }
+
+    }
+}
+
+    create_tests!(array);
+    create_tests!(vec vec!);
```
c410-f3r: Not sure if this approach is desired. If so, I will also do the same in sequence.rs.

src/types/list.rs
```diff
@@ -178,6 +178,42 @@ where
     }
 }
 
+macro_rules! array_impls {
+    ($($N:expr),+) => {
+        $(
+            impl<T> ToPyObject for [T; $N]
```
kngwyu: As these imples are not faster than one for slices(`impl ~ for &[T]`), I don't think we need this.

src/types/sequence.rs
```diff
@@ -282,6 +325,22 @@ where
     Ok(v)
 }
 
+fn extract_sequence_into_slice<'s, T>(obj: &'s PyAny, slice: &mut [T]) -> PyResult<()>
```
kngwyu: How about `extract_sequence_inplace`?

kngwyu:
Thank you!
`FromPyObject` impl looks good but I don't think we need `IntoPy<PyObject>` and `ToPyObject` impls for arrays.

davidhewitt:
Also thanks from me!

> `FromPyObject` impl looks good but I don't think we need `IntoPy<PyObject>` and `ToPyObject` impls for arrays.

I slightly disagree with this. `ToPyObject` taking `&[T; N]` maybe doesn't offer benefit over `&[T]` but I believe `IntoPy<PyObject>` offers value.

For example, the following function will not compile without the `IntoPy<PyObject>` implementation:

    #[pyfunction]
    fn make_array()-> [i32; 2] {
        [0, 0]
    }

src/types/sequence.rs
davidhewitt: I tend to think `_inplace` as referring to the first argument, so I would vote against that name.

`extract_sequence_into_slice` is fine imo.

src/types/list.rs
davidhewitt: I agree for `ToPyObject` but not for `IntoPy<PyObject>` - see below.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [0.9.0]
 
+* Implement `*Py` traits for arrays (up to 32).
```
davidhewitt: Can you please change the line to read the following and also move it to the `Added` section of the `0.9.0` changelog.

```suggestion
* Implement `FromPyObject` and `IntoPy<PyObject>` traits for arrays (up to 32). [#778](https://github.com/PyO3/pyo3/pull/778)
```

src/types/list.rs
davidhewitt: I think this is a good idea, thanks for adding comprehensive tests!

c410-f3r:
Comments addressed. Thank you guys

kngwyu:
> For example, the following function will not compile without the IntoPy<PyObject> implementation:

That makes sense, thanks :+1:

src/types/list.rs
```diff
@@ -178,6 +178,33 @@ where
...
+        $(
+            impl<T> IntoPy<PyObject> for [T; $N]
+            where
+                T: Copy + IntoPy<PyObject>,
+            {
+                fn into_py(self, py: Python) -> PyObject {
+                    unsafe {
+                        let ptr = ffi::PyList_New(self.len() as Py_ssize_t);
+                        for (i, e) in self.iter().copied().enumerate() {
+                            let obj = e.into_py(py).into_ptr();
```
kngwyu: `let obj = e.to_object(py)`.

src/types/sequence.rs
```diff
@@ -282,6 +325,22 @@ where
...
+where
+    T: FromPyObject<'s>,
+{
+    let seq = <PySequence as PyTryFrom>::try_from(obj)?;
+    if (seq.len()? as usize) != slice.len() {
```
kngwyu:
```suggestion
    if seq.len()? as usize != slice.len() {
```

src/types/sequence.rs
```diff
@@ -330,6 +389,9 @@ impl<'v> PyTryFrom<'v> for PySequence {
 
 #[cfg(test)]
 mod test {
+    macro_rules! create_tests {
```
kngwyu: Here not all tests are for `FromPyObject`.
Do we need to repeat all of them?
I think we really need it array version of `test_extract_bytearray_to_vec`.

src/types/list.rs
```diff
@@ -205,6 +232,9 @@ where
 
 #[cfg(test)]
 mod test {
```
kngwyu: Same as below.
Don't need to repeat all tests.

kngwyu:
Thank you for addressing comments, but as I commented, I think current macro-generated tests seem to make it unclear what is tested here.
For `FromPyObject` implementation, we just need a test that calls `let arr: [u32; N] = obj.extract().unwrap()`. Most of tests like `test_seq_contains` are not related to this PR at all.
We can say the same thing for `IntoPy<PyObject>` implementation.
Most of the tests are not array-specific and use `ToPyObject for &[T]` implementation(e.g., `test_new`).
What we need is simply a test that calls `[1, 2, 3].into_py(py)`.

c410-f3r:
I was waiting for David but looks like these tests are actually unwanted

davidhewitt:
Thanks - sorry about the extra work there. I thought your macro test implementation was nice, but @kngwyu has been a maintainer of pyo3 for much longer than I so I'm still deferring to his opinion when we differ.

I wasn't going to ask you to undo the work though - was thinking about pushing a change myself. Thanks for doing so.

kngwyu:
@c410-f3r 
Thank you for completing the PR!
@davidhewitt 
> so I'm still deferring to his opinion when we differ.

You don't have to do so. We sometimes make mistakes, but fortunately, we can help us with each other. When our opinion differs, it just means we need a discussion.

PR #783, opened 2020-03-02
Fix PyLayout and restrict PyRef::as_ref for non-native types

### LINKED ISSUE #782: Implementation of PyLayout::get_ptr for native types is broken
It's returning the address of the local variable `self`, which then is destroyed when the function returns:

https://github.com/PyO3/pyo3/blob/fee755adbea01504d12cd858e2705608c10163a4/src/pycell.rs#L28-L30

Several of the other implementations appear to be broken in the same way.

Demo code:
```rust
pub struct S(u64);
pub struct PyAny;

impl S {
    pub unsafe fn get_ptr(&self) -> *mut PyAny {
        // body copied from implementations of PyLayout::get_ptr
        (&self) as *const &Self as *const _ as *mut _
    }
}

fn main() {
    let a = [S(0), S(0)];
    for i in &a {
        let i: &S = i; // i is of type `&S`
        // note objects' address is different
        dbg!(i as *const _);
        // totally broken: note same address printed for different objects
        // also, address computed is before the beginning of either object
        dbg!(unsafe { i.get_ptr() });
    }
}
```

([Playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=a3dcd56c0819933eb33f2fc09c4d048b))

Output:

```
   Compiling playground v0.0.1 (/playground)
    Finished release [optimized] target(s) in 0.45s
     Running `target/release/playground`
[src/main.rs:16] i as *const _ = 0x00007ffe69ee9cf0
[src/main.rs:19] unsafe { i.get_ptr() } = 0x00007ffe69ee9ce8
[src/main.rs:16] i as *const _ = 0x00007ffe69ee9cf8
[src/main.rs:19] unsafe { i.get_ptr() } = 0x00007ffe69ee9ce8

```

kngwyu:
Fix #782.
Simply I make `get_ptr` a private method of `PyCellInner`.
This change forbids to get a base object of `#[pyclass(extends=NativeType]` style pyclasses.
Current `&PyAny` style API uses mainly *a pointer of a pointer* as an object handle, and integrates really badly with new `PyCell` :persevere:

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -249,7 +249,6 @@ or by `self_.into_super()` as `PyRef<Self::BaseClass>`.
 
 ```rust
 # use pyo3::prelude::*;
-use pyo3::PyCell;
 
 #[pyclass]
 struct BaseClass {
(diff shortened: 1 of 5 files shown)


kngwyu:
FYI, this is a valid implementation of `get_ptr` of PyAny
```rust
    unsafe fn get_ptr(&self) -> *mut PyAny {
        crate::Python::assume_gil_acquired().from_borrowed_ptr(self as *const _ as *mut _)
            as *const PyAny as *mut _
    }
```
... but I'm not sure we should do so.

davidhewitt:
I've had a read through and looks fine. It's a shame to lose the safe API to access the super class.

I want to think for a bit longer and see if I can figure out another way to solve this.

kngwyu:
> It's a shame to lose the safe API to access the super class.

Yeah, but since native type inheritance was broken in PyO3 < 0.9, it's a not breaking change and not *very* problematic.

davidhewitt:
Agreed. I suggest we merge this PR to fix the safety issue, and I'll create a separate issue for finding a safe API for upcasting.

PR #791, opened 2020-03-04
Make it enable to take &PyClass as arguments as pyfunctions/methods

### LINKED ISSUE #786: Restore support for `&Class` and `&mut Class` as arguments
> We can't implement FromPyObject for them. You can use PyRef<T> and PyRefMut<T> as arguments.

Agreed we can't implement `FromPyObject` for them. It would be nice if we can find another way for the generated wrappers to support `&Class` and `&mut Class` arguments - by calling `.try_borrow()` in the wrapper, perhaps, like we can do for `&self`.

_Originally posted by @davidhewitt in https://github.com/PyO3/pyo3/issues/785#issuecomment-593813961_

kngwyu:
Resolves #786.
For `C: PyClass`, this PR enables the use of `&C`, `&mut C`, `Option<&C>` and `Option<&mut C>` as function/method arguments.
The hack I used is
1. Set `ExtractExt::Target` for `&C`, like
```rust
impl<'a> pyo3::derive_utils::ExtractExt<'a> for &'a C
{
    type Target = pyo3::PyRef<'a, C>;
}
```
2. If the argument is one of `&C`/`&mut C`/`Option<&C>`/`Option<&mut C>`, extract the argument to `<type as ExtractExt::Target>` and insert deref.

But, for object protocols, still we need to use `PyRef<C>` or `PyRefMut<C>` due to its trait requirements.


pyo3-derive-backend/src/method.rs
```diff
@@ -305,55 +304,18 @@ pub fn is_ref(name: &syn::Ident, ty: &syn::Type) -> bool {
...
-                None
-            }
-        }
-        _ => {
-            None
-            //panic!("argument type is not supported by python method: {:?} ({:?})",
-            //name,
-            //ty);
-        }
+pub(crate) fn check_ty_optional<'a>(ty: &'a syn::Type) -> Option<&'a syn::Type> {
```
kngwyu: This is just a refactoring.
Not related to the logic so much.

kngwyu:
Somehow `tests/test_dunder.rs` was mode 100755  so I executed `chmod -x` for that file.

tests/test_methods.rs
```diff
@@ -405,6 +405,62 @@ fn method_with_lifetime() {
...
 }
 
+#[pyclass]
+struct MethodWithPyClassArg {
+    #[pyo3(get)]
+    value: i64,
+}
+
+#[pymethods]
+impl MethodWithPyClassArg {
```
davidhewitt: The test should probably check `Option<&mut C>` and `PyRefMut<C>`, just for completeness.

kngwyu:
Thank you for the review :wink:

1tgr:
FYI, I think the syntax `x: &Self` no longer works for method parameters as a result of this change. A method taking a parameter of type `&Self` did work on pyo3 v0.8.5.

It's because of this usage of the `#tref` token outside of the class's impl block:
https://github.com/kngwyu/pyo3/blob/96115eaaaad6df8039fb0dc07521137a89ce15b4/pyo3-derive-backend/src/pymethod.rs#L567

davidhewitt:
Thanks, have opened #1070 to remember this.

PR #792, opened 2020-03-04
Fix the interpretation of vararg separator

### LINKED ISSUE #790: "*" for keyword-only arguments allows signatures that it shouldn't
Using version `0.9.0-alpha.1`, I tweak the README example as follows, to add a keyword-only `debug` argument:

```rust
#[pyfunction(a, b, "*", debug = "false")]
/// Formats the sum of two numbers as string
fn sum_as_string(a: usize, b: usize, debug: bool) -> PyResult<String> {
    if debug {
        println!("DEBUGGING!");
    }
    Ok((a + b).to_string())
}
```

Now the following works in my Python interpreter, which is good:

```
>>> import string_sum
>>> string_sum.sum_as_string(4, 5)                                                              
'9'
```

But the following also works, when it shouldn't:

```
>>> string_sum.sum_as_string(4, 5, 6, 7, 8, 9, 10)
'9'
```

It looks like the `"*"` is get getting interpreted as "any number of positional arguments", where in this case that's not what it's supposed to mean.

kngwyu:
Resolves #790 
`*` does not mean `accept_args` and we should reject `fn(10)` for `def fn(*, var=10)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Usage of raw identifiers with `#[pyo3(set)]`. [#745](https://github.com/PyO3/pyo3/pull/745)
 * Usage of `PyObject` with `#[pyo3(get)]`. [#760](https://github.com/PyO3/pyo3/pull/760)
 * `#[pymethods]` used in conjunction with `#[cfg]`. #[769](https://github.com/PyO3/pyo3/pull/769)
-* Interpretation of `*`. #[792](https://github.com/PyO3/pyo3/pull/792)
+* Interpretation of `*` and some unreasonable behaviors of `#[args]`. #[792](https://github.com/PyO3/pyo3/pull/792)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Usage of raw identifiers with `#[pyo3(set)]`. [#745](https://github.com/PyO3/pyo3/pull/745)
 * Usage of `PyObject` with `#[pyo3(get)]`. [#760](https://github.com/PyO3/pyo3/pull/760)
 * `#[pymethods]` used in conjunction with `#[cfg]`. #[769](https://github.com/PyO3/pyo3/pull/769)
-* Interpretation of `*` and some unreasonable behaviors of `#[args]`. #[792](https://github.com/PyO3/pyo3/pull/792)
+* `"*"` in a `#[pyfunction()]` argument list incorrectly accepting any number of positional arguments (use `args = "*"` when this behaviour is desired). #[792](https://github.com/PyO3/pyo3/pull/792)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Usage of raw identifiers with `#[pyo3(set)]`. [#745](https://github.com/PyO3/pyo3/pull/745)
 * Usage of `PyObject` with `#[pyo3(get)]`. [#760](https://github.com/PyO3/pyo3/pull/760)
 * `#[pymethods]` used in conjunction with `#[cfg]`. #[769](https://github.com/PyO3/pyo3/pull/769)
+* Interpretation of `*`. #[792](https://github.com/PyO3/pyo3/pull/792)
 
 ### Removed
 
--- a/pyo3-derive-backend/src/pyfunction.rs
+++ b/pyo3-derive-backend/src/pyfunction.rs
@@ -52,34 +52,30 @@ impl PyFunctionAttr {
             NestedMeta::Lit(ref lit) => {
                 self.add_literal(item, lit)?;
             }
-            _ => {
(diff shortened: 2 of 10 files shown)


tests/test_methods.rs
```diff
@@ -231,11 +231,15 @@ impl MethArgs {
         [a.to_object(py), args.into(), kwargs.to_object(py)].to_object(py)
     }
 
+    #[args("*", c = 10)]
```
pganssle: Should you also add a test for `args(a, "*", c=10)`? (or whatever the correct syntax is for a required argument, I'm not sure).

kngwyu: Thanks.
Now this allows `args(a, "*", c=10)`, `args(a, b, "*", c=10)`, and `args("*", c=10)`, which seems confusing :thinking: 
So it may be good to restrict the syntax :thinking:

pganssle: I don't think it's a good idea to restrict things, and in fact we should also add support for positional-only arguments (in a separate change, IMO). The full spec for a python function definition can be found here: https://docs.python.org/3/tutorial/controlflow.html#special-parameters

```
def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
      -----------    ----------     ----------
        |             |                  |
        |        Positional or keyword   |
        |                                - Keyword only
         -- Positional only
```

That doesn't include `*args` and `**kwargs`.

kngwyu: Yeah, but our proc-macro has a somewhat more free syntax like
```rust
    #[args(a, "*", d=10)]
    fn get_pos_arg_kw_sep(&self, a: i32, b: i32, c: i32) -> PyResult<i32>
```
, which is really confusing.

davidhewitt: If we're working on this, should we also change to accept `*` instead of `"*"` ? e.g. `#[args(a, *, d=10)]`

davidhewitt: (Maybe we should note these down each as separate bug issues.)

kngwyu: syn doesn't accept it.

kngwyu:
I pushed a change that raises a compile error `#[pyfunction(a, "*", b)]`.
For problems of ambiguity, I opened another issue #794

pyo3-derive-backend/src/pyfunction.rs
```diff
@@ -94,97 +83,82 @@ impl PyFunctionAttr {
...
+        if self.has_kwargs {
+            return Err(syn::Error::new_spanned(
+                item,
+                "Keyword argument or kwargs(**) is not allowed after kwargs(**)",
+            ));
+        }
+        Ok(())
+    }
+
+    fn add_nv_common(
```
kngwyu: This function is just meant for refactoring.
Seprated the common parts from `add_name_value`.

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -406,15 +406,6 @@ fn impl_call(_cls: &syn::Type, spec: &FnSpec<'_>) -> TokenStream {
     quote! { _slf.#fname(#(#names),*) }
 }
 
-/// Converts a bool to "true" or "false"
-fn bool_to_ident(condition: bool) -> syn::Ident {
```
kngwyu: I removed this function since just `quote! { }` works.

tests/test_methods.rs
davidhewitt: I could be wrong, can't it parse it as `syn::Token![*]` ?

davidhewitt:
Will give this a full review tomorrow 👍

tests/test_methods.rs
kngwyu: https://docs.rs/syn/1.0.16/syn/enum.NestedMeta.html

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Usage of raw identifiers with `#[pyo3(set)]`. [#745](https://github.com/PyO3/pyo3/pull/745)
 * Usage of `PyObject` with `#[pyo3(get)]`. [#760](https://github.com/PyO3/pyo3/pull/760)
 * `#[pymethods]` used in conjunction with `#[cfg]`. #[769](https://github.com/PyO3/pyo3/pull/769)
+* Interpretation of `*` and some unreasonable behaviors of `#[args]`. #[792](https://github.com/PyO3/pyo3/pull/792)
```
davidhewitt: We should probably be more specific here, so that users which experience changed behaviour as a result of this fix can easily find what they need from the changelog.

```suggestion
* `"*"` in a `#[pyfunction()]` argument list incorrectly accepting any number of positional arguments (use `args = "*"` when this behaviour is desired). #[792](https://github.com/PyO3/pyo3/pull/792)
```

pyo3-derive-backend/src/pyfunction.rs
```diff
@@ -52,139 +52,109 @@ impl PyFunctionAttr {
             NestedMeta::Lit(ref lit) => {
                 self.add_literal(item, lit)?;
             }
-            _ => {
-                return Err(syn::Error::new_spanned(item, "Unknown argument"));
+            NestedMeta::Meta(syn::Meta::List(ref list)) => {
+                return Err(syn::Error::new_spanned(
+                    list,
+                    "List is not supported as argument",
+                ));
```
davidhewitt: Nice :)

kngwyu:
@davidhewitt 
Thanks!

PR #795, opened 2020-03-07
Write migration guide 

### LINKED ISSUE #785: Porting to 0.9 after #770
Two points (Using pyo3 master branch from today, Rust 1.43.0 nightly from 2020-02-22), both should be noted in the release notes.

## first

After #770, I've noticed that I can't use `&Class` and `&mut Class` as method arguments anymore, where `Class` is a `#[pyclass]` annotated struct.  Instead, it has to be `&PyCell<Class>` (I assume, at least that works).

## second

Furthermore, now I can't cast `PyAny` and `PyObject` objects referring to a `Class` to Rust references anymore.  Previously I used `PyAny::downcast_ref` (now `downcast`) and `PyObject::cast_as`.  The error is in both cases the same, i.e.
```
error[E0277]: the trait bound `ast::segments::OtherSegment: pyo3::instance::PyNativeType` is not satisfied
   --> src/ast/mod.rs:405:37
    |
405 |         } else if let Ok(seg) = seg.cast_as::<OtherSegment>(py) {
    |                                     ^^^^^^^ the trait `pyo3::instance::PyNativeType` is not implemented for `ast::segments::OtherSegment`
    |
    = note: required because of the requirements on the impl of `pyo3::conversion::PyTryFrom<'_>` for `ast::segments::OtherSegment`
```

One way I found to fix this was to use `seg.extract::<&PyCell<OtherSegment>>.borrow()`.
This is quite cumbersome -- is there/could there be a shortcut?

Of course this should also be noted in the release notes.

Thanks!

kngwyu:
Resolves #785 
I wrote a migration guide in Appendix B of the guide.
Now only pushed the migration guide, but I'm going to use this branch for the release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,20 +19,20 @@ travis-ci = { repository = "PyO3/pyo3", branch = "master" }
 appveyor = { repository = "fafhrd91/pyo3" }
 
 [dependencies]
-indoc = "0.3.4"
-inventory = "0.1.4"
-libc = "0.2.62"
-num-bigint = { version = ">= 0.2", optional = true }
-num-complex = { version = ">= 0.2", optional = true }
-num-traits = "0.2.8"
+indoc = "^0.3.4"
+inventory = "^0.1.4"
+libc = "^0.2.62"
+num-bigint = { version = "0.2", optional = true }
+num-complex = { version = "0.2", optional = true }
+num-traits = "^0.2.8"
 parking_lot = { version = "0.10", features = ["nightly"] }
-paste = "0.1.6"
+paste = "^0.1.6"
 pyo3cls = { path = "pyo3cls", version = "=0.9.0-alpha.1" }
-unindent = "0.1.4"
+unindent = "^0.1.4"
 
 [dev-dependencies]
 assert_approx_eq = "1.1.0"
-trybuild = "1.0.14"
+trybuild = "1.0.23"
 
 [build-dependencies]
 regex = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,16 +19,16 @@ travis-ci = { repository = "PyO3/pyo3", branch = "master" }
 appveyor = { repository = "fafhrd91/pyo3" }
 
 [dependencies]
-indoc = "^0.3.4"
-inventory = "^0.1.4"
-libc = "^0.2.62"
+indoc = "0.3.4"
+inventory = "0.1.4"
+libc = "0.2.62"
 num-bigint = { version = "0.2", optional = true }
 num-complex = { version = "0.2", optional = true }
-num-traits = "^0.2.8"
+num-traits = "0.2.8"
 parking_lot = { version = "0.10", features = ["nightly"] }
-paste = "^0.1.6"
+paste = "0.1.6"
 pyo3cls = { path = "pyo3cls", version = "=0.9.0-alpha.1" }
-unindent = "^0.1.4"
+unindent = "0.1.4"
 
 [dev-dependencies]
 assert_approx_eq = "1.1.0"
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -42,7 +42,7 @@ For more, see [the constructor section](https://pyo3.rs/master/class.html#constr
 PyO3 0.9 introduces [`PyCell`](https://pyo3.rs/master/doc/pyo3/pycell/struct.PyCell.html), which is
 a [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) like object wrapper
 for dynamically ensuring
-[Rust's rules of references](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references).
+[The Rules of References](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references).
 
 For `#[pymethods]` or `#[pyfunction]`s, your existing code should continue to work without any change.
 Python exceptions will automatically be raised when your functions are used in a way which breaks Rust's
@@ -58,13 +58,13 @@ struct Names {
 
 #[pymethods]
 impl Names {
-   #[new]
-   fn new() -> Self {
-       Names { names: vec![] }
-   }
-   fn merge(&mut self, other: &mut Names) {
-       self.names.append(&mut other.names)
-   }
+    #[new]
+    fn new() -> Self {
+        Names { names: vec![] }
+    }
+    fn merge(&mut self, other: &mut Names) {
+        self.names.append(&mut other.names)
+    }
 }
 # let gil = Python::acquire_gil();
 # let py = gil.python();
@@ -78,16 +78,18 @@ impl Names {
 #    isinstance(e, borrow_mut_err)
 # ");
 ```
-`Names` has `merge` method, which takes `&mut self` and `&mut Self`.
+`Names` has a `merge` method, which takes `&mut self` and another argument of type `&mut Self`.
 Given this `#[pyclass]`, calling `names.merge(names)` in Python raises a `PyBorrowMutError` exception,
 since it requires two mutable borrows of `names`.
 
-However, for `#[pyproto]` and some functions, you need to manually fix codes.
+However, for `#[pyproto]` and some functions, you need to manually fix the code.
 
 #### Object creation
-We could use the older `PyRef` and `PyRefMut` for object creation, but now they are just
-reference wrappers for `PyCell`.
-Use `PyCell::new` instead.
+In 0.8 object creation was done with `PyRef::new` and `PyRefMut::new`.
+In 0.9 these have both been removed.
+To upgrade code, please use `PyCell::new` instead.
+If a `PyRef` or `PyRefMut` is needed, just call `.borrow()` or `.borrow_mut()`
+on the newly-created `PyCell`.
 
 Before:
 ```compile_fail
@@ -111,8 +113,8 @@ let obj_ref = obj.borrow();
 ```
 
 #### Object extraction
-`T: PyClass`, `&T` and `&mut T` no longer have `FromPyObject` implementations.
-Instead you should extract `&PyCell`, `PyRef`, and `PyRefMut` respectively.
+For `PyClass` types `T`, `&T` and `&mut T` no longer have `FromPyObject` implementations.
+Instead you should extract `PyRef<T>` or `PyRefMut<T>`, respectively.  You can also extract `&PyCell<T>`.
 
 Before:
 ```ignore
@@ -125,7 +127,7 @@ After:
 ```
 # use pyo3::prelude::*;
 # use pyo3::types::{PyAny, IntoPyDict};
-# #[pyclass] struct MyClass {}
+# #[pyclass] #[derive(Clone)] struct MyClass {}
 # #[pymethods] impl MyClass { #[new]fn new() -> Self { MyClass {} }}
 # let gil = Python::acquire_gil();
 # let py = gil.python();
@@ -134,6 +136,7 @@ After:
 # let create_obj = || py.eval("c()", None, Some(d)).unwrap();
 let obj: &PyAny = create_obj();
 let obj_cell: &PyCell<MyClass> = obj.extract().unwrap();
+let obj_cloned: MyClass = obj.extract().unwrap(); // extracted via Clone
 {
     let obj_ref: PyRef<MyClass> = obj.extract().unwrap();
     // we need to drop obj_ref before we can extract a PyRefMut due to Rust's rules of references
@@ -143,8 +146,8 @@ let obj_ref_mut: PyRefMut<MyClass> = obj.extract().unwrap();
 
 
 #### `#[pyproto]`
-Most of `#[pyproto]` arguments requires [`FromPyObject`] implementation.
-So if your protocol methods take `&T` or `&mut T`(where `T: PyClass`),
+Most of the arguments to methods in `#[pyproto]` impls require a [`FromPyObject`] implementation.
+So if your protocol methods take `&T` or `&mut T` (where `T: PyClass`),
 please use `PyRef` or `PyRefMut` instead.
 
 Before:
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -40,9 +40,10 @@ For more, see [the constructor section](https://pyo3.rs/master/class.html#constr
 
 ### PyCell
 PyO3 0.9 introduces [`PyCell`](https://pyo3.rs/master/doc/pyo3/pycell/struct.PyCell.html), which is
-a [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) like object wrapper
-for dynamically ensuring
-[The Rules of References](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references).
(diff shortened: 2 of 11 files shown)


guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
+# Appendix B: Migration Guides for major version changes
+
```
davidhewitt: It might be worth adding a link to the CHANGELOG.md here too.
```suggestion
This guide can help you upgrade code through breaking changes from one PyO3 version to the next. For a detailed list of all changes, see [CHANGELOG.md](https://github.com/PyO3/pyo3/blob/master/CHANGELOG.md)
```

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
+# Appendix B: Migration Guides for major version changes
```
davidhewitt: Technically `0.8` -> `0.9` is a minor version, so I don't think we need to be so specific here.
```suggestion
# Appendix B: Migrating from older PyO3 versions
```

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+Basically you can return `Self` or `Result<Self>` directly.
+For more, see [the constructor section](https://pyo3.rs/master/class.html#constructor) of this guide.
+
+### PyCell
+PyO3 0.9 introduces [`PyCell`](https://pyo3.rs/master/doc/pyo3/pycell/struct.PyCell.html), which is
+a [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) like object wrapper
+for dynamically ensuring
+[Rust's rule of Reference](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references).
```
davidhewitt:
```suggestion
[Rust's rules of references](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references).
```

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+
+For `#[pymethods]` or `#[pyfunction]`s, `PyCell` works without any change.
+Just throw errors when there happens invalid borrowing.
```
davidhewitt:
```suggestion
For `#[pymethods]` or `#[pyfunction]`s, your existing code should continue to work without any change.
Python exceptions will automatically be raised when your functions are used in a way which breaks Rust's
rules of references.
```

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+# pyo3::py_run!(py, names borrow_mut_err, r"
+# try:
+#    names.merge(names)
+#    assert False, 'Unreachable'
+# except Exception as e:
+#    isinstance(e, borrow_mut_err)
+# ");
+```
+`Names` has `merge` method, which takes `&mut self` and `&mut Self`.
+Given this `#[pyclass]`, calling `names.merge(names)` in Python raises `PyBorrowMutError` exception,
```
davidhewitt:
```suggestion
Given this `#[pyclass]`, calling `names.merge(names)` in Python raises a `PyBorrowMutError` exception,
```

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+#[pymethods]
+impl Names {
+   #[new]
+   fn new() -> Self {
+       Names { names: vec![] }
+   }
+   fn merge(&mut self, other: &mut Names) {
+       self.names.append(&mut other.names)
+   }
+}
```
davidhewitt: It's really nice to see this example. It's mind-blowing how crazy unsound this is before `PyCell`! 😄

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+#    assert False, 'Unreachable'
+# except Exception as e:
+#    isinstance(e, borrow_mut_err)
+# ");
+```
+`Names` has `merge` method, which takes `&mut self` and `&mut Self`.
+Given this `#[pyclass]`, calling `names.merge(names)` in Python raises `PyBorrowMutError` exception,
+since it requires two mutable borrows of `names`,
+
+However, for `#[pyproto]` and some functions, you need to manually fix codes.
```
davidhewitt: We should probably link to the `#[pyproto]` section. Or move the `#[pyproto]` section here.

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+# try:
+#    names.merge(names)
+#    assert False, 'Unreachable'
+# except Exception as e:
+#    isinstance(e, borrow_mut_err)
+# ");
+```
+`Names` has `merge` method, which takes `&mut self` and `&mut Self`.
+Given this `#[pyclass]`, calling `names.merge(names)` in Python raises `PyBorrowMutError` exception,
+since it requires two mutable borrows of `names`,
```
davidhewitt:
```suggestion
since it requires two mutable borrows of `names`.
```

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+
+However, for `#[pyproto]` and some functions, you need to manually fix codes.
+
+#### Object creation
+We could use the older `PyRef` and `PyRefMut` for object creation, but now they are just
+reference wrappers for `PyCell`.
+Use `PyCell::new` instead.
```
davidhewitt: In `0.8` object creation was done with `PyRef::new` and `PyRefMut::new`. In `0.9` these have both been removed.

To upgrade code, `PyCell::new` instead. If a `PyRef` or `PyRefMut` is needed, just call `.borrow()` or `.borrow_mut()` on the newly-created `PyCell`.

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+# let gil = Python::acquire_gil();
+# let py = gil.python();
+# let typeobj = py.get_type::<MyClass>();
+# let d = [("c", typeobj)].into_py_dict(py);
+# let create_obj = || py.eval("c()", None, Some(d)).unwrap();
+let obj: &PyAny = create_obj();
+let obj_cell: &PyCell<MyClass> = obj.extract().unwrap();
+{
+    let obj_ref: PyRef<MyClass> = obj.extract().unwrap();
+    // we need to drop obj_ref before taking RefMut
```
davidhewitt:
```suggestion
    // we need to drop obj_ref before we can extract a PyRefMut due to Rust's rules of references
```

guide/src/migration.md
```diff
@@ -0,0 +1,183 @@
...
+# struct MyClass {}
+let gil = Python::acquire_gil();
+let py = gil.python();
+let obj = PyCell::new(py, MyClass {}).unwrap();
+let obj_ref = obj.borrow();
+```
+
+#### Object extraction
+Now for `T: PyClass`, `&T` and `&mut T` don't have `FromPyObject` implementation.
+Instead, you can use `&PyCell`, `PyRef`, and `PyRefMut` for object extraction.
```
davidhewitt:
```suggestion
`T: PyClass`, `&T` and `&mut T` no longer have `FromPyObject` implementations.
Instead you should extract `&PyCell`, `PyRef`, and `PyRefMut` respectively.
```

kngwyu:
@davidhewitt 
Thank you for the reviews!

guide/src/migration.md
```diff
@@ -42,10 +42,11 @@ For more, see [the constructor section](https://pyo3.rs/master/class.html#constr
 PyO3 0.9 introduces [`PyCell`](https://pyo3.rs/master/doc/pyo3/pycell/struct.PyCell.html), which is
 a [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) like object wrapper
 for dynamically ensuring
-[Rust's rule of Reference](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references).
+[Rust's rules of references](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references).
```
birkenfeld: I would be more explicit and say "rules regarding aliasing of references".

guide/src/migration.md
```diff
@@ -110,8 +111,8 @@ let obj_ref = obj.borrow();
 ```
 
 #### Object extraction
-Now for `T: PyClass`, `&T` and `&mut T` don't have `FromPyObject` implementation.
-Instead, you can use `&PyCell`, `PyRef`, and `PyRefMut` for object extraction.
+`T: PyClass`, `&T` and `&mut T` no longer have `FromPyObject` implementations.
```
birkenfeld:
```suggestion
For `PyClass` types `T`, `&T` and `&mut T` no longer have `FromPyObject` implementations.
```
To avoid confusion.  The current text sounds like `T` cannot be extracted, but I think it still can (as long as it's `Clone`)?

guide/src/migration.md
```diff
@@ -110,8 +111,8 @@ let obj_ref = obj.borrow();
...
+Instead you should extract `&PyCell`, `PyRef`, and `PyRefMut` respectively.
```
birkenfeld:
```suggestion
Instead you should extract `PyRef<T>` or `PyRefMut<T>`, respectively.  You can also extract `&PyCell<T>`.
```
(I don't know when you would want to extract `PyCell` when you can do the latter two.)

guide/src/migration.md
```diff
@@ -0,0 +1,184 @@
...
+
+After:
+```
+# use pyo3::prelude::*;
+#[pyclass]
+struct MyClass {}
+
+#[pymethods]
+impl MyClass {
+   #[new]
```
birkenfeld: nit: 3 spaces indent here

guide/src/migration.md
birkenfeld: Indeed!

guide/src/migration.md
```diff
@@ -0,0 +1,184 @@
...
+# let borrow_mut_err = py.get_type::<pyo3::pycell::PyBorrowMutError>();
+# pyo3::py_run!(py, names borrow_mut_err, r"
+# try:
+#    names.merge(names)
+#    assert False, 'Unreachable'
+# except Exception as e:
+#    isinstance(e, borrow_mut_err)
+# ");
+```
+`Names` has `merge` method, which takes `&mut self` and `&mut Self`.
```
birkenfeld:
```suggestion
`Names` has a `merge` method, which takes `&mut self` and another argument of type `&mut Self`.
```

guide/src/migration.md
```diff
@@ -0,0 +1,184 @@
...
+Given this `#[pyclass]`, calling `names.merge(names)` in Python raises a `PyBorrowMutError` exception,
+since it requires two mutable borrows of `names`.
+
+However, for `#[pyproto]` and some functions, you need to manually fix codes.
```
birkenfeld:
```suggestion
However, for `#[pyproto]` and some functions, you need to manually fix the code.
```

guide/src/migration.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### Object creation
+We could use the older `PyRef` and `PyRefMut` for object creation, but now they are just
```
birkenfeld: I don't understand the first part of this sentence.

guide/src/migration.md
```diff
@@ -0,0 +1,184 @@
...
+{
+    let obj_ref: PyRef<MyClass> = obj.extract().unwrap();
+    // we need to drop obj_ref before we can extract a PyRefMut due to Rust's rules of references
+}
+let obj_ref_mut: PyRefMut<MyClass> = obj.extract().unwrap();
+```
+
+
+#### `#[pyproto]`
+Most of `#[pyproto]` arguments requires [`FromPyObject`] implementation.
```
birkenfeld:
```suggestion
Most of the arguments to methods in `#[pyproto]` impls require a [`FromPyObject`] implementation.
```

guide/src/migration.md
```diff
@@ -0,0 +1,184 @@
...
+So if your protocol methods take `&T` or `&mut T`(where `T: PyClass`),
```
birkenfeld:
```suggestion
So if your protocol methods take `&T` or `&mut T` (where `T: PyClass`),
```

guide/src/migration.md
kngwyu: Hmm I want to use the same expression as the rustbook here

guide/src/migration.md
kngwyu: > he current text sounds like T cannot be extracted, but I think it still can (as long as it's Clone)?

Ah yes, it should be also noted.

guide/src/migration.md
kngwyu: Ooops, thanks.

guide/src/migration.md
kngwyu: Thanks :+1:

guide/src/migration.md
birkenfeld: Sure, I just don't think "the rules of references" is a standing term in the community. I wouldn't immediately know what it refers to until clicking the link.

kngwyu:
Pushed some changes for rust-numpy.

guide/src/migration.md
```diff
@@ -0,0 +1,189 @@
...
+   }
+}
+```
+
+Basically you can return `Self` or `Result<Self>` directly.
+For more, see [the constructor section](https://pyo3.rs/master/class.html#constructor) of this guide.
+
+### PyCell
+PyO3 0.9 introduces [`PyCell`](https://pyo3.rs/master/doc/pyo3/pycell/struct.PyCell.html), which is
+a [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) like object wrapper
```
davidhewitt:
```suggestion
a [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html)-like object wrapper
```

src/instance.rs
```diff
@@ -158,9 +158,12 @@ pub trait AsPyRef: Sized {
     fn as_ref(&self, py: Python<'_>) -> &Self::Target;
 }
 
-impl<'p, T: PyClass> AsPyRef for Py<T> {
-    type Target = PyCell<T>;
-    fn as_ref(&self, _py: Python) -> &PyCell<T> {
+impl<T> AsPyRef for Py<T>
+where
+    T: PyTypeInfo,
```
davidhewitt: 👍

guide/src/migration.md
davidhewitt: Why not have both and write something like 

```
RefCell-like object wrapper for ensuring Rust's rules regarding aliasing of references are upheld.
For more detail, see the 
[Rust Book's section on Rust's rules of references](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references)
```

Cargo.toml
```diff
@@ -19,20 +19,20 @@ travis-ci = { repository = "PyO3/pyo3", branch = "master" }
 appveyor = { repository = "fafhrd91/pyo3" }
 
 [dependencies]
-indoc = "0.3.4"
-inventory = "0.1.4"
-libc = "0.2.62"
-num-bigint = { version = ">= 0.2", optional = true }
-num-complex = { version = ">= 0.2", optional = true }
-num-traits = "0.2.8"
+indoc = "^0.3.4"
```
birkenfeld: `^` is the default operator, this doesn't change anything.

kngwyu: Thanks, I misunderstood that it is equal to "=="

kngwyu:
I decided to merge this PR now and will open a separate PR for the release for other PRs to sync with the master easier.

PR #796, opened 2020-03-08
Refactor build.rs

kngwyu:
- Remove lazy-static from build deps
- Mitigate version restriction for some build deps
- Use `Result<T, Box<dyn Error>>` as a result type instead of `Result<T, String>`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,19 +1,8 @@
-#[macro_use]
-extern crate lazy_static;
-
 use regex::Regex;
 use serde::Deserialize;
-use std::collections::HashMap;
-use std::convert::AsRef;
-use std::env;
-use std::fmt;
-use std::fs::File;
-use std::io;
-use std::io::{BufRead, BufReader};
-use std::path::Path;
-use std::process::exit;
-use std::process::Command;
-use std::process::Stdio;
+use std::io::{self, BufRead, BufReader};
+use std::process::{Command, Stdio};
+use std::{collections::HashMap, convert::AsRef, env, fmt, fs::File, path::Path};
 use version_check::{Channel, Date, Version};
 
 /// Specifies the minimum nightly version needed to compile pyo3.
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -121,35 +100,25 @@ static SYSCONFIG_VALUES: [&str; 1] = [
...
-                None
-            }
-        })
-        .collect();
-
+    let mut definitions = HashMap::new();
+    let tostr = |r: regex::Match<'_>| r.as_str().to_string();
+    for maybe_line in header_reader.lines() {
+        if let Some(captures) = define_regex.captures(&maybe_line?) {
+            match (captures.name("ident"), captures.name("value")) {
```
birkenfeld: if the regex can't match without capturing the groups, you can use `captures["ident"]` etc which already gives you a `&str`.

build.rs
```diff
@@ -198,7 +175,7 @@ fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, Strin
 
     let intepreter_config = InterpreterConfig {
```
birkenfeld: typo in variable name

build.rs
```diff
@@ -485,24 +469,25 @@ print(json.dumps({
 }))
 "#;
     let json = run_python_script(interpreter, script)?;
-    serde_json::from_str(&json).map_err(|e| format!("Deserializing failed: {}", e))
+    Ok(serde_json::from_str(&json)?)
```
birkenfeld: removal of these custom error messages is always a gamble: is the error message still as good afterwards?

build.rs
kngwyu: Oops, thanks.

build.rs
kngwyu: Thanks, changed to use a more specific error message here.

build.rs
kngwyu: Thanks, but looks we need `Option` here.

build.rs
```diff
@@ -421,42 +394,44 @@ fn get_rustc_link_lib(config: &InterpreterConfig) -> Result<String, String> {
...
+
     // check default python
-    let interpreter_config = get_config_from_interpreter(&PYTHON_INTERPRETER)?;
+    let interpreter_config = get_config_from_interpreter(&python_interpreter)?;
     if interpreter_config.version.major == 3 {
-        return Ok((
-            interpreter_config,
-            fix_config_map(get_config_vars(&PYTHON_INTERPRETER)?),
-        ));
+        return Ok((interpreter_config, get_config_vars(&python_interpreter)?));
```
kngwyu: Removed ` fix_config_map` here since I changed to call it in `get_config_vars` for both targets.

Issue #756, opened 2020-02-03
Release pool unsoundness


Alexander-N:
Follow up for #720 where NameErrors and SIGSEGVs were found. The error can be observed with two tests running in parallel (might need a few tries):

```rust
#[test]
fn test_release_race1() {
    let gil = Python::acquire_gil();
    let py = gil.python();
    println!("make ob1");
    let ob1 = py.eval("object()", None, None).unwrap();
    py.import("datetime").unwrap();
    py_run!(py, ob1, "print('ob1' + str(ob1))");
}

#[test]
fn test_release_race2() {
    let gil = Python::acquire_gil();
    let py = gil.python();
    println!("make ob2");
    let ob2 = py.eval("object()", None, None).unwrap();
    py.import("datetime").unwrap();
    py_run!(py, ob2, "print('ob2' + str(ob2))");
}
```
What might be happening is:
- ob1 gets put into the release pool
- the import releases the GIL temporarily
- test_release_race2 makes progress and ob2 gets put into the release pool
- test_release_race1 finishes and with dropping of it's GILGuard ob2 instead of ob1 is popped from the relase pool
- test_release_race2 fails because ob2 has been freed

davidhewitt:
Now that #887 is merged we should have resolved the thread safety issues here :tada:

Issue #653, opened 2019-10-27
Static references to Python objects are allowed, but unsafe


Askannz:
## 🐛 Bug Reports

There is no compile-time check preventing functions wrapped in PyO3 to take `static` references to Python objects. This allows those references to be stored and cause unsafe behavior.

Suggestion: forbid static references at compile-time (if possible), or add some documentation explaining why they shouldn't be used.

### 🌍 Environment

 - Your operating system and version: Archlinux
 - Your python version: 3.7.4
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: Repo package + virtualenv
 - Your rust version (`rustc --version`): rustc 1.39.0-nightly (ca3766e2e 2019-09-14)
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` Yes

### 💥 Reproducing

Rust:
```rust
use pyo3::prelude::*;
use pyo3::types::PyList;

#[pymodule]
fn my_module(_py: Python, m: &PyModule) -> PyResult<()> {

    #[pyclass]
    struct MyClass { list: &'static PyList}

    #[pymethods]
    impl MyClass {

        #[new]
        fn new(obj: &PyRawObject, list: &'static PyList) {
            obj.init({ MyClass { list } });
        }

        fn print(&self) {
            println!("{}", self.list.len()); // KABOOM
        }
    }

    m.add_class::<MyClass>()?;

    Ok(())
}
```

Python :
```
>>> from my_module import MyClass
>>> obj = MyClass([1, 2, 3])
>>> obj.print()
Segmentation fault (core dumped)
```

kngwyu:
It cannot be static since we often load Python interpreter dynamically.
But this sigsegv looks odd 🤔

PR #797, opened 2020-03-08
Add catch_unwind! macro to prevent panics crossing ffi boundaries

### LINKED ISSUE #492: Panics should be caught at the FFI boundary and propagated into python some other way
All Rust stack frames which have FFI stack frames directly under them should be guarded by a `catch_unwind` to ensure that it is impossible to accidentally panic back into Python’s stack frames. It is undefined behaviour to panic-unwind into stack frames of functions written in other languages, which makes any Rust-written python method that may panic for any reason – pyo3 is not exempt – unsound.

Alternatively, users of `pyo3` should be instructed to wrap their code into `catch_unwind` and handle this scenario on their own. In that case the requirement to not panic without `catch_unwind` should be thoroughly documented.

---

For reference, I encountered this issue by experimenting with errors and doing something along the lines of:

```rust
#[pyclass]
struct Exception {};

// In a `PyResult` returning method
return Err(PyErr::new::<Exception, _>("hello"));
```

Where `PyErr` panicked because `Exception` is not a valid `Exception` type.

davidhewitt:
Fixes #492 

This is a first attempt at updating all the wrapping code to use `catch_unwind` to prevent unsoundness of panics inside Rust callback functions.

I implemented this using a `pyo3::catch_unwind` macro, which takes a `Python` token and a body of code which should evaluate to a `PyResult`.

I ran into complications around lifetimes and the callback converters so I ended up simplifying a lot of the callback conversion code. I think a lot of those macros are now in slightly better shape but there's probably more refactoring that could be done.

Please give this a thorough review, I'm prepared to rework this as many times as needed to get it right.

TODO:
  - [X] Custom Exception type from BaseException
    - [ ] Mechanism for users to import this Exception (will do later after #805)
    - [x] Resume unwind when pyO3 gets this Exception in a PyErr
  - [ ] Panic section for the book

### DIFF
--- a/src/callback.rs
+++ b/src/callback.rs
@@ -157,6 +157,8 @@ where
 /// It sets up the GILPool and converts the output into a Python object. It also restores
 /// any python error returned as an Err variant from the body.
 ///
+/// Finally, any panics inside the callback body will be caught and translated into PanicExceptions.
+///
 /// # Safety
 /// This macro assumes the GIL is held. (It makes use of unsafe code, so usage of it is only
 /// possible inside unsafe blocks.)
@@ -204,11 +206,27 @@ macro_rules! callback_body {
 macro_rules! callback_body_without_convert {
     ($py:ident, $body:expr) => {{
         let pool = $crate::GILPool::new();
+        let unwind_safe_py = std::panic::AssertUnwindSafe(pool.python());
+        let result = match std::panic::catch_unwind(move || -> $crate::PyResult<_> {
+            let $py = *unwind_safe_py;
+            $body
+        }) {
+            Ok(result) => result,
+            Err(e) => {
+                // Try to format the error in the same way panic does
+                if let Some(string) = e.downcast_ref::<String>() {
+                    Err($crate::panic::PanicException::py_err((string.clone(),)))
+                } else if let Some(s) = e.downcast_ref::<&str>() {
+                    Err($crate::panic::PanicException::py_err((s.to_string(),)))
+                } else {
+                    Err($crate::panic::PanicException::py_err((
+                        "panic from Rust code",
+                    )))
+                }
+            }
+        };
 
-        let $py = pool.python();
-        let callback = move || -> $crate::PyResult<_> { $body };
-
-        callback().unwrap_or_else(|e| {
+        result.unwrap_or_else(|e| {
             e.restore(pool.python());
             $crate::callback::callback_error()
         })
--- a/src/err.rs
+++ b/src/err.rs
@@ -1,11 +1,12 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
+use crate::panic::PanicException;
 use crate::type_object::PyTypeObject;
 use crate::types::PyType;
 use crate::{exceptions, ffi};
 use crate::{
-    AsPyPointer, FromPy, IntoPy, IntoPyPointer, Py, PyAny, PyObject, Python, ToBorrowedObject,
-    ToPyObject,
+    AsPyPointer, FromPy, FromPyPointer, IntoPy, IntoPyPointer, ObjectProtocol, Py, PyAny, PyObject,
+    Python, ToBorrowedObject, ToPyObject,
 };
 use libc::c_int;
 use std::ffi::CString;
@@ -168,13 +169,33 @@ impl PyErr {
     ///
     /// The error is cleared from the Python interpreter.
     /// If no error is set, returns a `SystemError`.
-    pub fn fetch(_: Python) -> PyErr {
+    ///
+    /// If the error fetched is a `PanicException` (which would have originated from a panic in a
+    /// pyo3 callback) then this function will resume the panic.
+    pub fn fetch(py: Python) -> PyErr {
         unsafe {
             let mut ptype: *mut ffi::PyObject = std::ptr::null_mut();
             let mut pvalue: *mut ffi::PyObject = std::ptr::null_mut();
             let mut ptraceback: *mut ffi::PyObject = std::ptr::null_mut();
             ffi::PyErr_Fetch(&mut ptype, &mut pvalue, &mut ptraceback);
-            PyErr::new_from_ffi_tuple(ptype, pvalue, ptraceback)
+
+            let err = PyErr::new_from_ffi_tuple(ptype, pvalue, ptraceback);
+
+            if ptype == PanicException::type_object().as_ptr() {
+                let msg: String = PyAny::from_borrowed_ptr_or_opt(py, pvalue)
+                    .and_then(|obj| obj.extract().ok())
+                    .unwrap_or_else(|| String::from("Unwrapped panic from Python code"));
+
+                eprintln!(
+                    "--- PyO3 is resuming a panic after fetching a PanicException from Python. ---"
+                );
+                eprintln!("Python stack trace below:");
+                err.print(py);
+
+                std::panic::resume_unwind(Box::new(msg))
+            }
+
+            err
         }
     }
 
@@ -564,6 +585,7 @@ pub fn error_on_minusone(py: Python, result: c_int) -> PyResult<()> {
 #[cfg(test)]
 mod tests {
     use crate::exceptions;
+    use crate::panic::PanicException;
     use crate::{PyErr, Python};
 
     #[test]
@@ -574,5 +596,17 @@ mod tests {
         err.restore(py);
         assert!(PyErr::occurred(py));
         drop(PyErr::fetch(py));
+    }
+
+    #[test]
+    fn fetching_panic_exception_panics() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let err: PyErr = PanicException::py_err("new panic");
+        err.restore(py);
+        assert!(PyErr::occurred(py));
+        let started_unwind =
+            std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| PyErr::fetch(py))).is_err();
+        assert!(started_unwind);
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+* Panics from Rust will now be caught and raised as Python errors. [#797](https://github.com/PyO3/pyo3/pull/797)
 * `PyObject` and `Py<T>` reference counts are now decremented sooner after `drop()`. [#851](https://github.com/PyO3/pyo3/pull/851)
   * When the GIL is held, the refcount is now decreased immediately on drop. (Previously would wait until just before releasing the GIL.)
   * When the GIL is not held, the refcount is now decreased when the GIL is next acquired. (Previously would wait until next time the GIL was released.)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -188,6 +188,7 @@ mod internal_tricks;
 pub mod marshal;
 mod object;
 mod objectprotocol;
+pub mod panic;
 pub mod prelude;
 pub mod pycell;
 pub mod pyclass;
--- a/src/panic.rs
+++ b/src/panic.rs
@@ -1,0 +1,12 @@
+use crate::exceptions::BaseException;
+
+/// The exception raised when Rust code called from Python panics.
+///
+/// Like SystemExit, this exception is derived from BaseException so that
+/// it will typically propagate all the way through the stack and cause the
+/// Python interpreter to exit.
+pub struct PanicException {
+    _private: (),
+}
+
+pyo3_exception!(PanicException, BaseException);


kngwyu:
Awesome! :100: 
I'll leave some comments but generally, the design concept looks good.

src/callback.rs
```diff
@@ -115,50 +117,63 @@ wrapping_cast!(i32, Py_hash_t);
...
-    value: PyResult<T>,
-) -> <PyObjectCallbackConverter<T> as CallbackConverter>::Result {
-    PyObjectCallbackConverter::<T>::convert_result(py, value)
+    value.convert(py)
 }
 
+#[doc(hidden)]
 #[inline]
-pub unsafe fn cb_err<C>(_c: C, py: Python, err: impl Into<crate::PyErr>) -> C::Result
+pub fn callback_error<T>() -> T
```
kngwyu: We need this...?

src/callback.rs
```diff
@@ -115,50 +117,63 @@ wrapping_cast!(i32, Py_hash_t);
...
+
+#[macro_export]
+macro_rules! catch_unwind {
+    ($py:ident, $block:expr) => {{
+        let result = match std::panic::catch_unwind(|| -> PyResult<_> { $block }) {
+            Ok(result) => result,
+            Err(e) => {
+                // Try to format the error in the same way panic does
+                if let Some(string) = e.downcast_ref::<String>() {
+                    Err($crate::exceptions::RuntimeError::py_err((string.clone(),)))
```
kngwyu: Looks good, but isn't it better if we have a special exception type?

src/class/iter.rs
```diff
@@ -104,31 +99,20 @@ where
...
-        match val {
-            Some(val) => val.into_py(py).into_ptr(),
-            None => unsafe {
-                ffi::PyErr_SetNone(ffi::PyExc_StopIteration);
-                ptr::null_mut()
-            },
+    fn convert(self, py: Python) -> PyResult<*mut ffi::PyObject> {
+        match self.0 {
+            Some(val) => Ok(val.into_py(py).into_ptr()),
+            None => Err(crate::exceptions::StopIteration::py_err(())),
```
kngwyu: I checked this works almost the same as `ffi::PyErr_SetNone(ffi::PyExc_StopIteration)`

src/class/macros.rs
```diff
@@ -3,51 +3,46 @@
...
-            let py = $crate::Python::assume_gil_acquired();
-            let _pool = $crate::GILPool::new(py);
-            let slf = py.from_borrowed_ptr::<$crate::PyCell<T>>(slf);
-            $call!(slf, $conv, py, $f)
+            let gil = $crate::assume_gil_acquired();
+            let py = gil.py();
+            $crate::catch_unwind!(py, {
+                let _pool = $crate::GILPool::new(py);
+                let slf = py.from_borrowed_ptr::<$crate::PyCell<T>>(slf);
+                $crate::callback::convert(py, $call!(slf, $f)$(.map($conv))?)
```
kngwyu: Why this `map` works?
Isn't `conv` struct?

src/class/macros.rs
```diff
@@ -56,51 +51,49 @@ macro_rules! py_unary_refmut_func {
...
-        {
-            let py = Python::assume_gil_acquired();
-            let _pool = $crate::GILPool::new(py);
-            let slf = py.from_borrowed_ptr::<$crate::PyCell<T>>(slf);
-            let result = call_ref!(slf, $f);
-            $crate::callback::cb_convert($conv, py, result)
-        }
-        Some(wrap::<$class>)
-    }};
+    ($trait:ident, $class:ident :: $f:ident) => {
```
kngwyu: :+1:

src/class/macros.rs
```diff
@@ -231,23 +220,20 @@ macro_rules! py_ternary_num_func {
...
-                    Err(e) => Err(e.into()),
-                },
-                Err(e) => Err(e.into()),
-            };
-            $crate::callback::cb_convert($conv, py, result)
+            let gil = $crate::assume_gil_acquired();
+            let py = gil.py();
+            $crate::catch_unwind!(py, {
+                let _pool = $crate::GILPool::new(py);
+                let arg1 = py.from_borrowed_ptr::<$crate::types::PyAny>(arg1);
```
kngwyu: How about `let arg1 = py.from_borrowed_ptr::<$crate::types::PyAny>(arg1).extract()?;`

kngwyu:
@davidhewitt 
Now I'm not sure we should include this in the 0.9.0 release. What's your opnion?

davidhewitt:
Thanks for all the reviews! I will answer and make a round of changes maybe tomorrow.

> Now I'm not sure we should include this in the 0.9.0 release. What's your opnion?

If you're ready for 0.9 to be released then please proceed without waiting for this. We can include this in 0.10. (There's technically breaking changes here to a couple of `pub` functions and traits, so I think this will have to go in a breaking semver release.)

src/callback.rs
```diff
@@ -7,86 +7,88 @@ use crate::exceptions::OverflowError;
...
+    fn convert(self, _: Python) -> PyResult<()> {
+        Ok(())
+    }
+}
 
-impl CallbackConverter for UnitCallbackConverter {
-    type Source = ();
-    type Result = c_int;
-    const ERR_VALUE: Self::Result = -1;
+pub struct LenResultConverter(pub usize);
```
kngwyu: Now, this is rather a `Result` than a converter.
How about renaming it `LenResult`. `CbLenResult` or so?

konstin:
I haven't looked thoroughly at the implementation, I rather wanted to ask if we should forcefully exit (e.g. by abort) instead of raising an exception. A panic breaks all normal assumption about control flow in rust and might leave the rust part in a broken state, into which the python can than call later.

See https://doc.rust-lang.org/nomicon/unwinding.html and https://doc.rust-lang.org/nomicon/poisoning.html for some related information.

davidhewitt:
I can see why aborting is an easier choice. However, I don't think we should do this.

I think Python users writing Rust for the first time are likely to write a lot of things which panic (e.g. `unwrap`). Aborting the process is a fairly hard to debug error for such users who aren't familiar with low level code.

From the edition guide, I understand that the use of `catch_unwind` is exactly intended for the kind of scenario we have here: https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/controlling-panics-with-std-panic.html

konstin:
I like the way normal rust binaries currently handle this:

```rust
fn crashy() {
    None.unwrap()
}

fn main() {
    println!("Hello, world!");
    crashy();
    println!("Goodbye, world!");
}
```

This prints 

```
Hello, world!
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/main.rs:2:5
```

when running it with `cargo run`, and a long backtrace if the environment variable is set:

<details> <summary>Long backtrace</summary>

```
Hello, world!
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/main.rs:2:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1053
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1428
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:204
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:224
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:470
  11: rust_begin_unwind
             at src/libstd/panicking.rs:378
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  13: core::panicking::panic
             at src/libcore/panicking.rs:52
  14: core::option::Option<T>::unwrap
             at /rustc/823ff8cf1397a5772b1f6954b60576202bf91836/src/libcore/macros/mod.rs:10
  15: foo_crash::crashy
             at src/main.rs:2
  16: foo_crash::main
             at src/main.rs:7
  17: std::rt::lang_start::{{closure}}
             at /rustc/823ff8cf1397a5772b1f6954b60576202bf91836/src/libstd/rt.rs:67
  18: std::rt::lang_start_internal::{{closure}}
             at src/libstd/rt.rs:52
  19: std::panicking::try::do_call
             at src/libstd/panicking.rs:303
  20: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:86
  21: std::panicking::try
             at src/libstd/panicking.rs:281
  22: std::panic::catch_unwind
             at src/libstd/panic.rs:394
  23: std::rt::lang_start_internal
             at src/libstd/rt.rs:51
  24: std::rt::lang_start
             at /rustc/823ff8cf1397a5772b1f6954b60576202bf91836/src/libstd/rt.rs:67
  25: main
  26: __libc_start_main
  27: _start
```

</details>

pyo3 could have some message like "Your rust code panicked. pyo3 catched this for you. Do <...> to get a backtrace. Note: If you return a PyResult, it will be converted to a python error."

kngwyu:
@davidhewitt 
> We can include this in 0.10

Then let's include this in 0.9.
Releasing 0.10 just a few days 0.9 is not very user friendly.
@konstin 
> I rather wanted to ask if we should forcefully exit (e.g. by abort) instead of raising an exception.

It seems too difficult for me. I opened [a question](
https://users.rust-lang.org/t/can-caught-panics-be-converted-into-python-exception/39254) on the user forum.

konstin:
> @konstin
> 
> > I rather wanted to ask if we should forcefully exit (e.g. by abort) instead of raising an exception.
> 
> It seems too difficult for me. I opened [a question](https://users.rust-lang.org/t/can-caught-panics-be-converted-into-python-exception/39254) on the user forum.

Oh that's a great idea, thank you! Hopefully we'll get some clearer answers there

kngwyu:
Based on the discussion in the forum, now I think we should raise exceptions for panics.
It can be problematic if the thread panics while we have `PyRefMut`, but it's not UB.

src/gil.rs
```diff
@@ -105,6 +105,19 @@ pub struct GILGuard {
     no_send: Unsendable,
 }
 
+/// Marker to use when the GIL is expected to have been acquired
+pub struct AssumeGILAcquired {
```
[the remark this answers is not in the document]
davidhewitt: As the docs for `Python::assume_gil_acquired()` say:

https://github.com/PyO3/pyo3/blob/f2f5877f000b931394c6fbfd2546774184ec7c20/src/python.rs#L54-L59

`assume_gil_acquired()` is *super dangerous* imho because it allows you to get a `&'static` reference to GIL-bound objects, because the GIL lifetime can be inferred as `'static'`. I'm fairly certai this is what the crash in #653 is caused by.

By adding a token `AssumeGILAcquired`, then the GIL lifetime inferred must be bound by the lifetime of that token, preventing `'static` references.

If you want, I can separate this out into a separate PR.

pyo3-derive-backend/src/module.rs
```diff
@@ -220,14 +220,17 @@ fn function_c_wrapper(name: &Ident, spec: &method::FnSpec<'_>) -> TokenStream {
         {
             const _LOCATION: &'static str = concat!(stringify!(#name), "()");
 
-            let _py = pyo3::Python::assume_gil_acquired();
-            let _pool = pyo3::GILPool::new(_py);
-            let _args = _py.from_borrowed_ptr::<pyo3::types::PyTuple>(_args);
-            let _kwargs: Option<&pyo3::types::PyDict> = _py.from_borrowed_ptr_or_opt(_kwargs);
+            let gil = pyo3::assume_gil_acquired();
```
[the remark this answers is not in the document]
davidhewitt: See comment above for why I think we need this. I am happy to change the name, or factor this out into a separate PR, but I am convinced that `Python::assume_gil_acquired()` must die. We need to have some kind of "token" on the stack to which the compiler can tie the GIL lifetime.

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -360,35 +366,29 @@ pub(crate) fn impl_wrap_setter(
...
-                Err(e) => e.restore_and_minus1(_py),
-            }
+            let gil = pyo3::assume_gil_acquired();
+            let _py = gil.py();
+            pyo3::catch_unwind!(_py, {
+                let _pool = pyo3::GILPool::new(_py);
+                let _slf = _py.from_borrowed_ptr::<pyo3::PyCell<#cls>>(_slf);
+                #borrow_self
+                let _value = _py.from_borrowed_ptr::<pyo3::types::PyAny>(_value);
+                let _val = pyo3::FromPyObject::extract(_value)?;
```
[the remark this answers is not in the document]
davidhewitt: Yes I was very convinced that forcing `catch_unwind!` blocks to evaluate to `PyResult` would make our lives a lot easier 😄

src/callback.rs
```diff
@@ -7,86 +7,88 @@ use crate::exceptions::OverflowError;
...
+    const ERR_VALUE: Self = -1;
+}
+
+impl PyCallbackResult for () {
+    const ERR_VALUE: Self = ();
+}
 
-impl<T> CallbackConverter for PyObjectCallbackConverter<T>
+/// Convert the result of callback function into the appropriate return value.
+pub trait IntoPyCallbackResult<Target> {
```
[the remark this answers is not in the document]
davidhewitt: Unfortunately this isn't quite true, for example `libc::c_int` can be converted into a callback result of either `PyObject` or `libc::c_int` depending on the Python callback.

src/class/iter.rs
davidhewitt: 👍

davidhewitt: I made the switch both because it now returns `PyResult` and because I want to reduce the use of `unsafe` as much as possible. In this case I saw we had a safe API to do the same thing, so I chose to use it.

src/callback.rs
davidhewitt: In theory I like a special exception type, but I am not sure it is helpful in practice.

If we make an exception `RustPanicException`, how do we `import` it in Python code to catch it? Without the ability to name the exception in Python, they will be forced to write things such as `catch Exception as e`, which is bad practice because they're then catching all exceptions and not just panics.

I used to use C++ / boost::Python, which defines custom exception types for its `ArgumentError`, which was really frustrating when you then cannot catch this exception.

How we solve the `import pyo3.RustPanicException` gets even more complicated as we consider the possibility that multiple python libraries might be using different `pyo3` versions, and each version might have its own version of `RustPanicException`.

If we can come up with a good solution for this `import` problem, I am in favour of making a custom exception type. Until then, I want to stick to `RuntimeError`.

src/class/macros.rs
davidhewitt: I'm assuming that `conv` is a function which takes one argument and returns something new.

It just so happens that tuple structs with only one parameter can also behave as such functions:
E.g.

```
struct Foo<T>(pub T);

fn wrap(x: Option<i32>) -> Foo<i32> {
     x.map(Foo)
}
```

src/gil.rs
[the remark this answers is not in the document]
kngwyu: Ah I understand, thanks.
But I think the problem is the design of `GILPool`.
Since `GILPool` frees objects when it drops, `Python` has to have the same lifetime as it.

src/callback.rs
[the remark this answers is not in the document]
kngwyu: OK, thanks.

davidhewitt:
> pyo3 could have some message like "Your rust code panicked. pyo3 catched this for you. Do <...> to get a backtrace. Note: If you return a PyResult, it will be converted to a python error."

We could potentially add this as an optional feature `panic_abort`, or maybe even piggy-back off the the `cfg` setting if it's possible. The only problem is I'm not sure where to call `panic::set_hook` (it seems like it's probably a lot of overhead to call it in every wrapper?)

davidhewitt:
> Based on the discussion in the forum, now I think we should raise exceptions for panics.
It can be problematic if the thread panics while we have PyRefMut, but it's not UB.

Cool. I am not sure I see why panic with PyRefMut specifically is a problem. AFAIK drop code still runs during stack unwinding. We should _definitely_ add a test for the PyRefMut case.

src/gil.rs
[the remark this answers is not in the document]
davidhewitt: Ok. For now I will undo this change from this PR, and instead open a separate issue.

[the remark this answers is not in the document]
kngwyu: How about adding `assume_gil_acquired` method for `GILPool`, instead of creating a new type?

src/class/macros.rs
davidhewitt: Good idea, let me try that...

src/gil.rs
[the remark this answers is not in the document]
davidhewitt: `GILPool<'p>` still has a lifetime which may get inferred to `'static` in the same way as the existing `Python<'p>` can. So we can't put it there. Let's figure this out in a separate issue.

kngwyu:
@davidhewitt 
> I am not sure I see why panic with PyRefMut specifically is a problem. 

I said `PyRefMut` is problematic because it can cause *logical* error when panics.
[UnwindSafe documents](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html#who-implements-unwindsafe) says that `&mut T` or `RefCell<T>` is not UnwindSafe since it can break *logical* invariant.

Here's [an example of such a logical invariant](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7723c86f394a3af2e08ee1c095c5390f).
We sometimes hold logical assumptions for mutable data. In the above example, I assume that the elements of `array` are monotonically increasing.
However, this kind of logical assumption can be broken because of panic, like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9e01d77c674266c5264c12c4f033778c).
If we are certain that such kind of logical error can not happen, we can use `AssertUnwindSafe`.

The problem we have is we assume `PyRefMut<T>` is `AssertUnwindSafe` even if users don't intend it. Maybe we should note this *implicit AssertUnwindSafe* in the guide.

src/gil.rs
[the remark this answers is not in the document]
kngwyu: > GILPool<'p> still has a lifetime which may get inferred to 'static in the same way as the existing Python<'p> can.

I mean removing `Python` token from GILPool and implementing
```rust
pub fn assume_gil_acquired<'a>(&'a self) -> Python<'a> { ... }
```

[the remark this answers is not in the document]
davidhewitt: Ah, I see. Yes, that's much better.

But I still don't think that quite works, because the `GILPool` drop may run arbitrary code which could panic. So we need this pool to live only inside of `std::panic::catch_unwind`.

But we need `Python` after `std::panic::catch_unwind` finishes, so we can  restore the exception to Python.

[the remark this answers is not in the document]
kngwyu: > But I still don't think that quite works, because the GILPool drop may run arbitrary code which could panic. So we need this pool to live only inside of std::panic::catch_unwind.

I don't think we have to be too nervous about panic here, but it is indeed problematic that `Python<'p>` has two confusing meanings:
- It means that all references registered by `Python::from_owned/borrowed_pointer` have the lifetime `'p`. This is why `'p` must be the same as the lifetime of `GILGuard` or `GILPool`.
- It means that the current thread has GIL
In this case, `Python` is still necessary to ensure that we have GIL... :thinking:

src/callback.rs
davidhewitt: I don't like using `Result` when it's not an alias of `std::result::Result` so I'm going to call it `LenCallbackOutput`

davidhewitt: And in fact I'm going to rename the `PyCallbackResult` trait to `PyCallbackOutput` in the same way.

src/gil.rs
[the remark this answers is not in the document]
davidhewitt: I agree that `Python<'p>` is confusing for this duality. I think this might be solved with #756 . For now I will back out of changing it with this PR.

davidhewitt:
> The problem we have is we assume PyRefMut<T> is AssertUnwindSafe even if users don't intend it. Maybe we should note this implicit AssertUnwindSafe in the guide.

I'm not hugely keen on "implicit `AssertUnwindSafe`" in the guide, because many pyo3 users may not read it and make mistakes.

I have been thinking that the best solution is probably to "poison" the `PyCell`  if there is a panic during a ` PyRefMut` borrow, in just the same way a `Mutex` can be poisoned. What do you think of this idea?

src/class/basic.rs
```diff
@@ -502,27 +487,17 @@ where
...
-                        Err(e) => Err(e),
-                    };
-                    match res {
-                        Ok(val) => val.into_py(py).into_ptr(),
-                        Err(e) => e.restore_and_null(py),
-                    }
-                }
-                Err(e) => PyErr::from(e).restore_and_null(py),
-            }
+            let py = crate::Python::assume_gil_acquired();
```
konstin: I might have missed this in the thread, but is `assume_gil_acquired` sound without calling `GILPool::new` beforehand? If so, could we document the reason somewhere?

davidhewitt: In all current code, `GILPool::new` is called immediately after, like this:

```rust
let pool = GILPool::new(py);
...
```

What I'm changing it to in this PR is now

```rust
catch_unwind!(py, {
    let pool = GILPool::new(py);
    ...
})
...
```

However you've rightly noticed I have accidentally lost the `GILPool::new` here. I will fix this and also carefully check the rest of the changes I made.

I can't comment on whether this usage of `assume_gil_acquired` is sound, but I have other soundness concerns with it. See #800

src/callback.rs
kngwyu: OK, it's convincing.

kngwyu:
> I have been thinking that the best solution is probably to "poison" the PyCell if there is a panic during a PyRefMut borrow, in just the same way a Mutex can be poisoned. What do you think of this idea?

I'm not for that.
- Simply **We rarely need it**. It can be only problematic when
  - Panics happen when a user is doing some delicate operation with mutable data
  - He or she ignores the error like
    ```python
    try
       pyclass.call()
    except RuntimeError:
        pass
    ```
  So I don't think we should take care of such cases seriously.
-  To handle poisoning, `std::sync::Mutex` has a complex [try_lock](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.try_lock) API, but now parking_lot's Mutex with [simpler try_lock](https://docs.rs/lock_api/0.3.3/lock_api/struct.Mutex.html#method.try_lock), which does not handle poisoning, is also widely used. It seems like this fact implies most of users don't actually need poisoning.

davidhewitt:
Ok. Fwiw `lock_api` is also not `UnwindSafe`: https://github.com/Amanieu/parking_lot/issues/32

I guess we can always change to have poisoning later if it turns out to be a problem.

Sometime in the next few days I'll push some tests and documentation to this branch

programmerjake:
Personally, I only `panic!` when my program should be aborted due to logic errors, so Python code ignoring any resulting exceptions is likely to observe my code behaving in unexpected/unintended ways. That said, I'm fine with `PyCell` not implementing poisoning as long as the specific Python exception thrown is extremely uncommon to catch and ignore.

davidhewitt:
What if we made the exception type derive from `BaseException`, like `SystemExit` ?

https://docs.python.org/3/library/exceptions.html#SystemExit

This would mean that it would be _really_ rare for Python programs to catch this exception.

To do this, I'd just like there to be _some_ way to refer to this exception from Python code, so that in cases where users really do want to catch it, they can. I haven't thought of a good way to do this, so any ideas are welcome.

programmerjake:
> What if we made the exception type derive from `BaseException`, like `SystemExit` ?

Sounds like a good idea.

> https://docs.python.org/3/library/exceptions.html#SystemExit
> 
> This would mean that it would be _really_ rare for Python programs to catch this exception.
> 
> To do this, I'd just like there to be _some_ way to refer to this exception from Python code, so that in cases where users really do want to catch it, they can. I haven't thought of a good way to do this, so any ideas are welcome.

What about injecting a module containing the `PanicException` definition into `sys.modules` if it's not already there so any python code can just `import pyo3_panics.PanicException` to access it? If the module is already there then we use the existing `PanicException` class and abort on `panic!` if it's not found in the existing module.

We could just publish a python package to PyPI to reserve the name with some simple code like:

```python
class PanicException(BaseException):
    pass
```

and it would be constructed by calling `PanicException(panic_message_as_python_string)`

kngwyu:
> To do this, I'd just like there to be some way to refer to this exception from Python code, so that in cases where users really do want to catch it, they can. I

How about recommending users to register our custom exception module to their module?
But maybe we need another PR to do so.

> What about injecting a module containing the PanicException definition into sys.modules

Can it be confusing?

programmerjake:
> > What about injecting a module containing the PanicException definition into sys.modules
> 
> Can it be confusing?

Probably not from the perspective of PyO3 users, but it may be confusing for developers of PyO3 itself. The user would see `pyo3_panics` as a module that can be imported in order to access `PanicException`, which would get raised on `panic!` whether or not the user imported the `pyo3_panics` module, looking just like standard Python where they imported some module that throws an exception imported from some other module that `__main__` didn't import.

The only edge cases are:
* when they have something else named `pyo3_panics` -- PyO3 might be forced to abort on panic if the `pyo3_panics` module is messed up.
* when they try to `import pyo3_panics` before the Rust library is imported and they didn't install the `pyo3_panics` package using `pip` or similar.

programmerjake:
Additionally, when PyO3 is translating an exception to `PyErr`, it might be good to panic again if the exception is a `PanicException` in order to propagate panics.

davidhewitt:
> How about recommending users to register our custom exception module to their module?

Couldn't we do this automatically? E.g. add a `_pyo3` submodule to the top-level module exposed by the extension.

If that can work, I think I prefer it to a globally installed module. Though the globally installed module is a nice idea too; it's just more work for us :D

> Additionally, when PyO3 is translating an exception to PyErr, it might be good to panic again if the exception is a PanicException in order to propagate panics.

Yes I think this is a good idea.

programmerjake:
One other benefit of the global module is it allows multiple independent PyO3 libraries to use the exact same type.

davidhewitt:
> One other benefit of the global module is it allows multiple independent PyO3 libraries to use the exact same type.

True, but it also complicates compatibility if we ever needed to change this module for any reason.

I'm very torn on which is the best approach :smile:

kngwyu:
> Couldn't we do this automatically? E.g. add a _pyo3 submodule to the top-level module exposed by the extension.

My 1st attempt is placed on #805 

> If that can work, I think I prefer it to a globally installed module. 

I don't know how we can do it. Maybe we need one more `.so` file?

kngwyu:
I feel like 'exposing custom exceptions' is too difficult while it does not offer so many benefits.
They would *sometimes* help to debug, though.

Using RuntimeError is not the best, but looks a sufficiently better solution for the next release.

birkenfeld:
If it's instrumental for safety that the exception isn't caught, we can't rely on that anyway.  It is pretty common in Python code to catch anything, log it, and continue in *some* way.

kngwyu:
@davidhewitt 
Thank you for reporting it.
It happens around `__ipow__`.
I'm going to try to fix it, but if I couldn't we should remove `__ipow__` API.

src/class/macros.rs
```diff
@@ -319,17 +313,13 @@ macro_rules! py_dummy_ternary_self_func {
...
-                    ffi::Py_INCREF(slf);
-                    slf
-                }
-                Err(e) => e.restore_and_null(py),
-            }
+            $crate::catch_unwind!(py, {
+                let _pool = $crate::GILPool::new(py);
+                let slf_cell = py.from_borrowed_ptr::<$crate::PyCell<T>>(slf);
+                let arg1 = py.from_borrowed_ptr::<$crate::PyAny>(arg1);
+                call_mut!(slf_cell, $f, arg1)?;
```
kngwyu, line 320:
```suggestion
                call_mut!(slf_cell, $f, arg1)?;
                ffi::Py_INCREF(slf);
```
Here we need `INCREF` not to release the object.

davidhewitt: Thanks, that's fixed it 👍

davidhewitt:
Now that rebase has been done, I plan to (tomorrow) experiment with a couple of designs for RuntimeError / PanicException and see what I like best. Then we should be good to perhaps merge this. Thanks for your patience 😄

davidhewitt:
As promised, I've created a basic `PanicException` which is used when panicking.

For now users can't import this exception type to catch it, so for the moment this will cause the whole stack to unwind and the program to exit. Let's resolve that separately in #805 . Once that's done, we can add some documentation about panicking to the guide.

src/err.rs
```diff
@@ -171,12 +169,20 @@ impl PyErr {
...
     /// If no error is set, returns a `SystemError`.
-    pub fn fetch(_: Python) -> PyErr {
+    pub fn fetch(py: Python) -> PyErr {
         unsafe {
             let mut ptype: *mut ffi::PyObject = std::ptr::null_mut();
             let mut pvalue: *mut ffi::PyObject = std::ptr::null_mut();
             let mut ptraceback: *mut ffi::PyObject = std::ptr::null_mut();
             ffi::PyErr_Fetch(&mut ptype, &mut pvalue, &mut ptraceback);
+
+            if ptype == PanicException::type_object().as_ptr() {
```
kngwyu, line 179: I **really** don't understand why we need this.
Why is it problematic to fetch `PanicException`?
cc: @programmerjake

kngwyu:
>  Let's resolve that separately in #805 . Once that's done, 

Honestly, I'm not sure I can resolve the problem now.
So, instead of exporting the exception, how about recommending users to use this pattern?
```python
    except SomeException as e:
...
    except BaseException as e:
        if e.__class__.__name__ == "PanicException":
              # do something...
        else:
              raise e
```

src/err.rs
davidhewitt: The idea is that unless caught a `panic!` should cause a stack unwind all the way until program exit (like it does without this PR). The changes in this PR are intended to make it so that we unwind safely through the FFI layer, because what we do at the moment is technically UB.

This means we need two main things:
   - When Rust code panics, we start unwinding in python by raising `PanicException`.
   - If this `PanicException` is not caught, we should use `std::panic::resume_unwind()` when we get back into Rust code so that we continue the unwind.

I need the original panic payload to call `resume_unwind()`, but at the moment it's difficult for me to store that on `PanicException`. I expect that will create a lot of work which will extend this already large PR, so I opened #853 instead.

Without the original payload, for now I just call `panic!` instead of `std::panic::resume_unwind()`, which is close enough. It has the downside that the panic handler runs again, so at the moment the `thread '<unnamed>' panicked at xyz` message will be printed multiple times as the program unwinds. (Once for the original `panic!` and once each time we continue unwinding when we fetch `PanicException`)

Perhaps I should add a comment to the code documenting this?

davidhewitt:
> Honestly, I'm not sure I can resolve the problem now.
So, instead of exporting the exception, how about recommending users to use this pattern?

That pattern looks good for now. I wasn't saying that #805 will be resolved soon - but I hope that one day we find a solution 😄 

Shall I add some docs to this PR? Alternatively you can merge this and add an issue to add a panic section to the book, if you want to get access to the `CallbackConverter` changes.

davidhewitt:
Travis failure is a bit strange, I will take a look at that maybe tomorrow.

src/err.rs
kngwyu: > because what we do at the moment is technically UB.

It's not true.
As I explained it can cause just a *logic* error.

> If this PanicException is not caught, we should use std::panic::resume_unwind() when we get back into Rust code so that we continue the unwind.

So I'm not still convinced that PyO3 itself should cause panic for that.
Let's consider a scenario where multiple PyO3 crates, say, c1 and c2 work together.
c1 can fetch c2's PanicException. Do we really want to resume panic in such a situation?
How about adding an API (e.g., `resume_panic`) to PanicException?

kngwyu:
> Shall I add some docs to this PR?

Yeah, we need some documents for panic handling.
But it's ok to do that in another PR if you need to take some time.

src/err.rs
davidhewitt: > because what we do at the moment is technically UB.

By this I mean the current pyo3 behaviour of not using `catch_unwind` is UB - Rust panic through the FFI stack frames isn't well defined. This is why we need to add `catch_unwind`.

I agree with you that catching `PanicException` can cause a logic error. This could occur either when Python catches the exception, or if pyO3 does not continue to unwind after fetching a `PanicException`.

This is why I think the default behaviour should be to keep unwinding all the way to program exit - even if c1 fetches c2's `PanicException`.

I'll put some diagrams together this evening to try to explain better.

davidhewitt:
Ok, new plan. I still can't settle on what design we like best for panic exceptions.

Also I saw this morning that the [ffi unwind group](https://github.com/rust-lang/project-ffi-unwind) is preparing an RFC for a "C unwind" ABI which might make this PR redundant / unnecessary. At the very least it might significantly change this PR's design. 

I have reached out on the ffi-unwind group Zulip to ask their opinion on where they think panic design is going for this kind of library.

So for now I want to wait on merging this PR again, and have put it back into draft. I have cut out almost all of the interesting code refactorings into a new PR #858 . We can then think more slowly on this PR while we wait to see what the language decision is for ffi unwinding.

davidhewitt:
... The guys on Zulip were very helpful in explaining that the "C unwind" ABI is for C++ exceptions and wouldn't be relevant for what we're doing here.

So I think that the only decision we have to make before I finish off this PR is whether we should `resume_unwind` if we fetch a `PanicException`. 

I think the answer is yes, because normally if I write `panic!` in a Rust program, it exits the program with a panic message.

So the default behaviour I think we should be aiming for, is that if someone writes `panic!` anywhere in pyo3 binding code, that panic will keep unwinding all the way until the program exits. To do this, if we get a PanicException from Python, we need to `resume_unwind` with it.

programmerjake:
If it's not too hard, it would be nice to optionally print a Python stack trace when resuming a panic.

kngwyu:
Just for reference:
image

davidhewitt:
> I'm going to make some use cases and test them.

👍 . I have rebased this patch on master and force-pushed; it is now _much_ less smaller so we have plenty of time to decide what we like.

I think now this diff is quite small it makes sense to include docs in this PR again. And perhaps a solution for #853

kngwyu:
I wrote this test code.
```rust
#[pyclass]
struct MyClass {}

#[pymethods]
impl MyClass {
    fn panic(&self) -> usize {
        panic!(">_<")
    }
}

#[test]
fn test_catch_panic() {
    use pyo3::types::IntoPyDict;

    let py = gil.python();
    let my_class = PyCell::new(py, MyClass {}).unwrap();
    let vars = [("c", my_class)].into_py_dict(py);    
    let result = py
        .run("c.panic()", None, Some(vars))
        .map_err(|e| e.print(py))
        .unwrap();
}
```

If we use `resume_unwind`, the error message is
```
---- test_catch_panic stdout ----
thread 'test_catch_panic' panicked at '>_<', tests/test_exceptions.rs:102:9
thread 'test_catch_panic' panicked at 'Unwrapped panic from Python code', src/err.rs:183:17
```
And if we not use, the one is
```
running 1 test
pyo3_runtime.PanicException: >_<
test test_catch_panic ... FAILED

failures:

---- test_catch_panic stdout ----
thread 'test_catch_panic' panicked at '>_<', tests/test_exceptions.rs:102:9
thread 'test_catch_panic' panicked at 'called `Result::unwrap()` on an `Err` value: ()', tests/test_exceptions.rs:114:18
```
So... though I'm not sure which is better, here it's not so meaningful to cause panic in `fetch` since users can easily cause panic by `unwrap`. Maybe it's sufficient to make the error printable(like attempting in #686), instead of continuing panic.

> I agree with you that catching PanicException can cause a logic error. This could occur either when Python catches the exception, or if pyO3 does not continue to unwind after fetching a PanicException.

Yeah, but I don't think PyO3 itself needs to ensure that there's no logic error. Users themselves should deal with this kind of errors.

davidhewitt:
> If we use `resume_unwind`, the error message is
> 
> ```
> ---- test_catch_panic stdout ----
> thread 'test_catch_panic' panicked at '>_<', tests/test_exceptions.rs:102:9
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> thread 'test_catch_panic' panicked at 'Unwrapped panic from Python code', src/err.rs:183:17
> ```

I'm suprised by that - I understood that with `std::panic::resume_unwind` the second panic message (`Unwrapped panic from Python code`) should not run. Did you just use the implementation I have in this PR which uses `panic!` in place of `resume_unwind` because I don't have access to the `Box<dyn Any>` payload?

> So... though I'm not sure which is better, here it's not so meaningful to cause panic in fetch since users can easily cause panic by unwrap. Maybe it's sufficient to make the error printable(like attempting in #686), instead of continuing panic.

I think printing the error is definitely a good idea, whether we resume or not.

So I think we should mark this as `help-wanted` - I think both designs are fine but we seem to prefer different ones (I prefer resuming the unwind, you prefer not to)? Would be interesting to have input from others on whether uncaught `PanicException`s should lead to `resume_unwind`.

kngwyu:
> I'm suprised by that -

Yeah, I'm not sure what happens too :roll_eyes:  Please run the test code yourself.

davidhewitt:
> Please run the test code yourself.

:+1: Give me a few days to play with the printing also and I'll report back

kngwyu:
@davidhewitt 
I'm not still convinced that we should panic when fetching a panic error, but leaving this PR unmerged is worse.
If you still think we need to retrieve panic, then let's merge this PR as is. We can fix this behavior if a user complains about it.

davidhewitt:
@kngwyu if that's how you feel, that's fine with me. I can re-open #853 after to change the re-pancking to use `resume_unwind!` instead of `panic!`.

davidhewitt:
First need to fix travis. I can do that this weekend.

davidhewitt:
FYI haven't forgotten this, will do tomorrow.

davidhewitt:
@kngwyu This is now ready for review and/or merge if you want. I'd still like to improve `PanicException` to carry the payload so that I can use `resume_unwind!` instead of `panic!`, but first I think I might have to finish off #686 to make that possible...

src/err.rs
```diff
@@ -168,12 +169,20 @@ impl PyErr {
     ///
     /// The error is cleared from the Python interpreter.
     /// If no error is set, returns a `SystemError`.
```
kngwyu, line 171: Please add a comment about when it causes panic.

davidhewitt:
I've added a note to that doc.

I've also improved the error message that occurs. Now, with this sample code:

```
use pyo3::prelude::*;
use pyo3::exceptions;
use pyo3::types::{PyDict, IntoPyDict};
use pyo3::{ffi, AsPyPointer, AsPyRef};

#[pyclass]
struct MyClass { }

#[pymethods]
impl MyClass {
  fn panic(&self) {
    panic!("Hello!");
  }
}

fn main() {
  let py = gil.python();

  let c = PyCell::new(py, MyClass { }).unwrap();

  pyo3::py_run!(py, c, "def foo(c): c.panic()\n  \nfoo(c)");
}
```

This is the output:

```
thread 'main' panicked at 'Hello!', src\main.rs:13:5
--- PyO3 is resuming a panic after an uncaught PanicException ---
Python stack trace below:

  File "<string>", line 3, in <module>
  File "<string>", line 1, in foo
pyo3_runtime.PanicException: Hello!
error: process didn't exit successfully: `target\debug\pyo3-scratch.exe` (exit code: 101)
```

src/err.rs
```diff
@@ -168,13 +169,31 @@ impl PyErr {
...
-    pub fn fetch(_: Python) -> PyErr {
+    ///
+    /// If the current error is a `PanicException` (which would have originated in other pyo3 code)
+    /// then this function will resume the uncaught panic.
```
kngwyu, line 174: What does `uncaught` mean here?
We catch the panic and resume it here, right?

davidhewitt: Ah yes, I meant the Python exception specifically was not caught in Python code. Perhaps this can be phrased better?

davidhewitt:
FYI while finishing this branch off I had an idea for a nice refactoring #901. But I wanted to open that as a separate PR, so please review that PR first and then I will rebase this one and finish up the documentation suggestion requested.

davidhewitt:
Rebased on #901 and updated the `PyErr::fetch` documentation. Let me know if you want anything else changed. 👍

kngwyu:
Thank you for lots of work!

PR #801, opened 2020-03-12
guide: add type overview

### LINKED ISSUE #789: Overview of pointer types and related traits
When diving deeper into the PyO3 docs, I'm always flattened by the sheer number of smart pointer types and related traits.  I get that most of them are necessary, and some of them may be remnants that are on the way out, but it's still overwhelming, especially once you start playing tetris with those types, trying to get the result you need.  The guide doesn't help in that regard.

I propose (and volunteer to draft, as far as I understand it) a table-like list of those traits, their approximate meaning, and when to use them.

It can be located in the guide, or in the API docs, as you prefer.  But wherever it is, it should be linked to from every mentioned type's API docs, and the explanations should be made consistent.

birkenfeld:
fixes #789

### DIFF
--- a/guide/src/types.md
+++ b/guide/src/types.md
@@ -1,0 +1,143 @@
+# GIL lifetimes, mutability and Python object types
+
+On first glance, PyO3 provides a huge number of different types that can be used
+to wrap or refer to Python objects.  This page delves into the details and gives
+an overview of their intended meaning, with examples when each type is best
+used.
+
+
+## Mutability and Rust types
+
+Since Python has no concept of ownership, and works solely with boxed objects,
+any Python object can be referenced any number of times, and mutation is allowed
+from any reference.
+
+The situation is helped a little by the Global Interpreter Lock (GIL), which
+ensures that only one thread can use the Python interpreter and its API at the
+same time, while non-Python operations (system calls and extension code) can
+unlock the GIL.  (See [the section on parallelism](parallelism.md) for how to do
+that in PyO3.)
+
+In PyO3, holding the GIL is modeled by acquiring a token of the type
+`Python<'py>`, which serves three purposes:
+
+* It provides some global API for the Python interpreter, such as
+  [`eval`][eval].
+* It can be passed to functions that require a proof of holding the GIL,
+  such as [`PyObject::clone_ref`][clone_ref].
+* Its lifetime can be used to create Rust references that implicitly guarantee
+  holding the GIL, such as [`&'py PyAny`][PyAny].
+
+The latter two points are the reason why some APIs in PyO3 require the `py:
+Python` argument, while others don't.
+
+The PyO3 API for Python objects is written such that instead of requiring a
+mutable Rust reference for mutating operations such as
+[`PyList::append`][PyList_append], a shared reference (which, in turn, can only
+be created through `Python<'_>` with a GIL lifetime) is sufficient.
+
+However, Rust structs wrapped as Python objects (called `pyclass` types) usually
+*do* need `&mut` access.  Due to the GIL, PyO3 *can* guarantee thread-safe acces
+to them, but it cannot statically guarantee uniqueness of `&mut` references once
+an object's ownership has been passed to the Python interpreter, ensuring
+references is done at runtime using `PyCell`, a scheme very similar to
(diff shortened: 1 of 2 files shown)


birkenfeld:
Obviously unfinished, but I'd like feedback if the style is ok...

guide/src/types.md
```diff
@@ -0,0 +1,101 @@
...
+
+Since `PyObject` does not carry a GIL lifetime, using most of its methods
+calling Python API will require a `Python` token.
+
+Can be cloned using Python reference counts with `.clone_ref()`.
+
+**Conversions:**
+
+- To `&PyAny`: `.as_ref(py)`
+- To `Py<ConcreteType>`: ???
```
kngwyu: We can use `obj.as_ref(py).extract::<Py<ConcreteType>>`

guide/src/types.md
```diff
@@ -0,0 +1,101 @@
...
+
+**Conversions:**
+
+- From `PyObject`: `obj.as_ref(py)`
+
+
+### `PyCell<SomeType>`
+
+**Represents:** a reference to a `PyClass` instance which is wrapped in a Python
+object.  The cell part is an analog to Rust's `RefCell` to allow access to
```
kngwyu:
```suggestion
object.  The cell part is an analog to stdlib's [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) to allow access to
```

kngwyu:
Thank you, it's really easy to read.
My concern is it lacks a section that explains how we should use native types(`&PyDict`, `&PyList`, or so).
And it would be good to explain that we allow mutable operations for `&PyDict` or so since it's really hard to track mutability of them

guide/src/types.md
```diff
@@ -0,0 +1,101 @@
...
+
+On first glance, PyO3 provides a huge number of different types that can be used
+to wrap or refer to Python objects.  This page tries to give an overview of
+their meaning, with examples when each type is best used.
+
+* explain: pyclass
+* explain: GIL lifetime
+* explain: mutability?
+
+## Smart pointers
```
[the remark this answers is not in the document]
birkenfeld: Indeed, that was one question I was going to pose to y'all :)

guide/src/types.md
birkenfeld: OK, that's good to know!

birkenfeld:
> Thank you, it's really easy to read.
> My concern is it lacks a section that explains how we should use native types(`&PyDict`, `&PyList`, or so).

What specifically do you have in mind here?

> And it would be good to explain that we allow mutable operations for `&PyDict` or so since it's really hard to track mutability of them

OK, I'll try to find a spot to integrate this.

kngwyu:
> What specifically do you have in mind here?

Just like what you've added as `### PyTuple, PyDict, and many more`.

guide/src/types.md
```diff
@@ -0,0 +1,157 @@
...
+
+**Note:** `PyObject` is semantically equivalent to `Py<PyAny>` and might be
+merged with it in the future.
+
+
+### `PyAny`
+
+**Represents:** a Python object of unspecified type, restricted to a GIL
+lifetime.  Currently, `PyAny` can only ever occur as a reference, e.g. `&PyAny`
+or `PyRef<PyAny>`.
```
kngwyu: `PyRef<PyAny>` is duplicated since 0.9

guide/src/types.md
```diff
@@ -0,0 +1,157 @@
...
+**Represents:** a native Python object of known type, restricted to a GIL
+lifetime just like `PyAny`.
+
+**Used:** Whenever you want to operate with native Python types while holding
+the GIL.  Like `PyAny`, this is the most convenient form to use for function
+arguments and intermediate values.
+
+**Conversions:**
+
+- To `PyAny`: `obj.as_ref()`
```
kngwyu: I recommend adding 'To `Py<T>`:  `Py::from(obj)`

guide/src/types.md
davidhewitt: > duplicated

I think you meant deprecated / removed?

kngwyu:
Thank you!

PR #809, opened 2020-03-16
Fix PyModule::dict

kngwyu:
Here `from_owned_ptr` is used, which decrements refcnt invalidly and can cause segfault.


davidhewitt:
:+1: good catch

PR #812, opened 2020-03-17
Bump version to 0.9.0

### LINKED ISSUE #808: release non-alpha version of 0.9
release non-alpha version of 0.9

kngwyu:
Resolves #808

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * `PyTypeInfo::BaseLayout` and `PyClass::BaseNativeType`. [#770](https://github.com/PyO3/pyo3/pull/770)
 * `PyDowncastImpl`. [#770](https://github.com/PyO3/pyo3/pull/770)
 * Implement `FromPyObject` and `IntoPy<PyObject>` traits for arrays (up to 32). [#778](https://github.com/PyO3/pyo3/pull/778)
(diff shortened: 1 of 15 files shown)


birkenfeld:
@kngwyu I'd like to update the docstrings of pyo3 types to point to the new guide first. I can do it today.

kngwyu:
@birkenfeld 
Thank you!
I'm willing to wait for that.

birkenfeld:
Can we also import PyAny to the main crate namespace?

kngwyu:
@birkenfeld 
I don't have any strong opinions about that.
They are hidden since https://github.com/PyO3/pyo3/commit/302c099a766a2a3759933b9c5da9c7ca90c91507 and left as is.
@konstin 
What kind of intention did you have for not exposing `types::*` in lib.rs?

birkenfeld:
I think it makes sense to put the long list of concrete types there.

But if we want to enhance visibility of `PyAny` and make clear that it's an important type, counterpart to `PyObject`, it should be in the root/prelude.

konstin:
The idea was to prevent the top level from becoming to cluttered with all the builtin types. But imo PyAny can also be the root namespace and/or the prelude.

kngwyu:
@birkenfeld @konstin 
Thank you for the feedback!
I'll create a PR for `PyAny` later.

kngwyu:
Now ready to release :)

birkenfeld:
Thanks! Can you set a tag?

kngwyu:
Sure, I made it now. https://github.com/PyO3/pyo3/releases/tag/v0.9.0

PR #820, opened 2020-03-18
More bytes, dict FFI

ijl:


src/ffi/bytesobject.rs
```diff
@@ -53,4 +53,6 @@ extern "C" {
         s: *mut *mut c_char,
         len: *mut Py_ssize_t,
     ) -> c_int;
+    #[cfg(not(PyPy))]
+    pub fn _PyBytes_Resize(bytes: *mut *mut PyObject, newsize: Py_ssize_t) -> c_int;
```
ijl: https://docs.python.org/3/c-api/bytes.html#c._PyBytes_Resize

src/ffi/dictobject.rs
```diff
@@ -92,6 +92,14 @@ extern "C" {
         key: *mut *mut PyObject,
         value: *mut *mut PyObject,
     ) -> c_int;
+    #[cfg(not(PyPy))]
+    pub fn _PyDict_Next(
```
ijl: https://github.com/python/cpython/blob/a24107b04c1277e3c1105f98aff5bfa3a98b33a0/Include/cpython/dictobject.h#L49 and elsewhere for the others.

kngwyu:
Thank you!

PR #779, opened 2020-02-29
Check python and target arcitecture match in build.rs

davidhewitt:
Fixes #752

PR #830, opened 2020-03-23
use struct.calcsize("P") rather than platform.machine()

oconnor663:
platform.machine() gives the wrong answer if you're running 32-bit
Python on a 64-bit machine. ~sys.maxsize~ struct.calcsize("P") (EDITED based on https://github.com/PyO3/pyo3/pull/830#issuecomment-603906094) gives the answer we want. See
also https://stackoverflow.com/a/1405971/823869.

Also use CARGO_CFG_TARGET_POINTER_WIDTH rather than inferring the
pointer width from CARGO_CFG_TARGET_ARCH.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -32,7 +32,7 @@ struct InterpreterConfig {
     /// Prefix used for determining the directory of libpython
     base_prefix: String,
     executable: String,
-    machine: String,
+    calcsize_pointer: Option<u32>,
 }
 
 #[derive(Deserialize, Debug, Clone, PartialEq)]
@@ -180,7 +180,7 @@ fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, Strin
         ld_version: "".to_string(),
         base_prefix: "".to_string(),
         executable: "".to_string(),
-        machine: "".to_string(),
+        calcsize_pointer: None,
     };
 
     Ok((interpreter_config, fix_config_map(config_map)))
@@ -433,10 +433,11 @@ fn find_interpreter_and_get_config() -> Result<(InterpreterConfig, HashMap<Strin
 /// Extract compilation vars from the specified interpreter.
 fn get_config_from_interpreter(interpreter: &str) -> Result<InterpreterConfig> {
     let script = r#"
+import json
+import platform
+import struct
 import sys
 import sysconfig
-import platform
-import json
 
 PYPY = platform.python_implementation() == "PyPy"
 
@@ -456,7 +457,7 @@ print(json.dumps({
     "base_prefix": base_prefix,
     "shared": PYPY or bool(sysconfig.get_config_var('Py_ENABLE_SHARED')),
     "executable": sys.executable,
-    "machine": platform.machine()
+    "calcsize_pointer": struct.calcsize("P"),
 }))
 "#;
     let json = run_python_script(interpreter, script)?;
@@ -475,7 +476,7 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<String> {
         }
     }
 
-    check_target_architecture(&interpreter_config.machine)?;
+    check_target_architecture(interpreter_config)?;
 
     let is_extension_module = env::var_os("CARGO_FEATURE_EXTENSION_MODULE").is_some();
     if !is_extension_module || cfg!(target_os = "windows") {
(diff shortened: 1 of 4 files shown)


oconnor663:
This should fix the issue described at https://github.com/PyO3/pyo3/pull/779#issuecomment-602717125.

oconnor663:
It would be ideal to have CI test this case for us. Would you guys be open to adding test targets like what I've got in https://github.com/oconnor663/blake3-py/blob/master/.github/workflows/push.yml?

davidhewitt:
Thanks very much for this. I would be in favour of having both 32-bit and 64-bit Python in CI.

build.rs
```diff
@@ -33,6 +33,7 @@ struct InterpreterConfig {
     base_prefix: String,
     executable: String,
     machine: String,
```
kngwyu: Could you please remove `machine`?

kngwyu:
I'm going to prepare x86/x64 tests for macOS and Win.
Please wait for that.

kngwyu:
I enabled x64 + macOS and x86 + Windows tests in #831. 
Please rebase on the master and add a test you want (x86 + macOS?)

oconnor663:
I've tried to add Linux CI testing, but it's failing with this error:

/home/runner/work/pyo3/pyo3/target/debug/deps/pyo3-1c9bfe8599f85944: error while loading shared libraries: libpython3.7m.so.1.0: cannot open shared object file: No such file or directory

Is that a known problem?

kngwyu:
@oconnor663 
As in travis scripts,
```bash
PYTHON_LIB=$($PYTHON_BINARY -c "import sysconfig; print(sysconfig.get_config_var('LIBDIR'))")

export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$PYTHON_LIB:$HOME/rust/lib"
```
would solve the problem.

But I don't think we need Ubuntu CI on Actions unless it can test something different from what we do on travis.

oconnor663:
Oh that makes sense. I missed the Travis runs. I'll just remove the `test.yml` change then.

oconnor663:
Looks like I'm hitting some `OverflowError: timestamp out of range`, possibly related to PyPy. Could those be my fault?

kngwyu:
@oconnor663 
No, this is due to #720

oconnor663:
> The failing test uses platform.architecture()[0] == "32bit" to determine whether the architecture is 32bit or 64bit...

That's consistent with [the admonition here](https://stackoverflow.com/a/1405971/823869), which I mention in the code comments in this PR:

> BTW, you might be tempted to use platform.architecture() for this. Unfortunately, its results are not always reliable, particularly in the case of OS X universal binaries.
>
> ```
> $ arch -x86_64 /usr/bin/python2.6 -c 'import sys,platform; print platform.architecture()[0], sys.maxsize > 2**32'
> 64bit True
> $ arch -i386 /usr/bin/python2.6 -c 'import sys,platform; print platform.architecture()[0], sys.maxsize > 2**32'
> 64bit False
> ```

Let me see if I can switch it over to the same `maxsize` approach.

oconnor663:
Actually I hadn't seen [this comment](https://stackoverflow.com/questions/1405913/how-do-i-determine-if-my-python-shell-is-executing-in-32bit-or-64bit-mode-on-os/1405971#comment6209952_1405971) before. It sounds like `struct.calcsize("P")` might be the One True Way?

oconnor663:
https://github.com/PyO3/pyo3/pull/830/commits/884db67ef55fce69d8a5e59581bd408310f1f6ca doesn't seem to have helped. I'm going to keep staring at it.

oconnor663:
It looks like the test failure is only on macOS, and only on Python 3.5 specifically. I don't know what could've changed here between 3.5 and 3.6.

davidhewitt:
The exception pathway being hit looks like it exists on python 3.5 but was removed from 3.6.

https://github.com/python/cpython/blob/3.5/Modules/_datetimemodule.c#L4936

oconnor663:
Cool, that's definitely what we're seeing. But I don't know enough about this test to interpret whether we want it to fail here or not. Is it that 3.5 is failing for a dumb obsolete reason, or is it that 3.6-3.8 happen to silently succeed when we wish they would fail?

oconnor663:
I'm going to defer to you all on this one. I'll remove my temporary debugging commits. If you want to see the debugging output, it's visible in this run: https://github.com/PyO3/pyo3/pull/830/checks?check_run_id=534169690

build.rs
```diff
@@ -456,7 +457,13 @@ print(json.dumps({
     "base_prefix": base_prefix,
     "shared": PYPY or bool(sysconfig.get_config_var('Py_ENABLE_SHARED')),
     "executable": sys.executable,
-    "machine": platform.machine()
+
+    # The reason we don't use platform.architecture() here is that it's not
```
kngwyu: I don't want to pass this string to Python.
How about moving to Rust side or just removing it(since we have the same one below)?

kngwyu:
@oconnor663 
Thank you for hard debugging!

I think it's better for us to wait for @pganssle, who is the author of our datetime binding.
I don't have a macOS machine so this is hard to debug for me, though I'm going to try using Actions.

konstin:
Is 32-bit Mac OS still a used platform? Even numpy doesn't have 32-bit Mac OS wheels on pypi (https://pypi.org/project/numpy/#files). If I remember correctly there isn't a wheel tag for 32-bit Mac OS.

This is what the [blog post of rust demoting apple 32-bit targets](https://blog.rust-lang.org/2020/01/03/reducing-support-for-32-bit-apple-targets.html#why-are-those-targets-being-demoted):

> ## Why are those targets being demoted?
>
> Apple dropped support for running 32-bit binaries starting from macOS 10.15 and iOS 11. They also prevented all developers from cross-compiling 32-bit programs and apps starting from Xcode 10 (the platform’s IDE, containing the SDKs).
>
> Due to those decisions from Apple, the targets are no longer useful to our users, and their choice to prevent cross-compiling makes it hard for the project to continue supporting the 32-bit platform in the long term.

build.rs
oconnor663: Removed.

oconnor663:
Made the comment change requested above.

I want to emphasize that this bug is causing build failures with the current release version, for example https://github.com/oconnor663/blake3-py/runs/534852943. If you guys can think of a way to reduce the scope of this change, so that we can release a fix quickly without the risk of breaking anything, that might be worth doing.

davidhewitt:
It's all a bit arcane.

I couldn't find anything about this specifically in the cpython changelog. I did find the commit in cpython which reworked the implementation removing this `OverflowError`. It looks like it's a side-effect of changes for PEP 495.

Prior to those changes the implementation used `mktime`. I can't find good documentation on the apple docs about what values should be accepted. The following stack overflow post suggests that for some undocumented reason macOS can't handle `time_t` values (aka `mktime` output) earlier than 1900.

That would at least be consistent with the error we're seeing on that CI. So I suggest you mark that test as xfail for macOS on 3.5.

davidhewitt:
(Sorry for not being more helpful on this one, I haven't had much free time recently!)

oconnor663:
I just xfailed the tests on macOS, but now something else datetime+3.5 related is failing on Windows?That's odd to me, because everything passed on this run: https://github.com/PyO3/pyo3/actions/runs/63179251

kngwyu:
@oconnor663 
It's not related to this PR. Please don't care.

>So I suggest you mark that test as xfail for macOS on 3.5.

I'm not strongly against using xfail, but please let me debug this for a while.

oconnor663:
Gotcha, I've removed the commit.

oconnor663:
Note: I put `fail-fast: false` on one of me earlier temporary commits, and I've since removed it from this branch. But it looks like there might be some unexpected failures getting masked by failures on other platforms here. It might be worth putting `fail-fast: false` in master, or at least in whatever experimental branches you're working on.

kngwyu:
@oconnor663 
Sorry I can't detect the cause at last. Please use xfail for the test.

oconnor663:
I've added back https://github.com/PyO3/pyo3/pull/830/commits/e44e4ce1e9799ab39167e0506b86fc94df08789c.

kngwyu:
Thank you!
I'll push some tweaks to make CI happy.

PR #832, opened 2020-03-24
Need to be able to create structs via macro_rules

gilescope:
(Currently only possible using tt escape type which doesn't do any processing.)

There's quite a few places I think we need to update but using `pyclass` as an example to agree the right way to do things. When you pass the argument via macro_rules unless it's the escape sequence type of `tt` then the Expr::Path seems to get wrapped in an Expr::Group whatever type I choose for the variable I'm passing in: `ident`, `path`, `expr`.

(btw I notice you can't match on box which might make the code a tad nicer.)

### DIFF
--- a/tests/test_macros.rs
+++ b/tests/test_macros.rs
@@ -1,0 +1,12 @@
+use pyo3::prelude::*;
+
+macro_rules! make_struct_using_macro {
+    // Ensure that one doesn't need to fall back on the escape type: tt
+    // in order to macro create pyclass.
+    ($className:path) => {
+        #[pyclass(name=$className)]
+        struct MyClass {}
+    };
+}
+
+make_struct_using_macro!(MyClass);
--- a/tests/test_macros.rs
+++ b/tests/test_macros.rs
@@ -44,6 +44,30 @@ macro_rules! fn_macro {
 
 fn_macro!("(a, b=None, *, c=42)", a, b = "None", c = 42);
 
+macro_rules! property_rename_via_macro {
+    ($prop_name:ident) => {
+        #[pyclass]
+        struct ClassWithProperty {
+            member: u64,
+        }
+
+        #[pymethods]
+        impl ClassWithProperty {
+            #[getter($prop_name)]
+            fn get_member(&self) -> u64 {
+                self.member
+            }
+
+            #[setter($prop_name)]
+            fn set_member(&mut self, member: u64) {
+                self.member = member;
+            }
+        }
+    };
+}
+
+property_rename_via_macro!(my_new_property_name);
+
 #[test]
 fn test_macro_rules_interactions() {
     Python::with_gil(|py| {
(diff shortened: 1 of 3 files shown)


gilescope:
(The test fails to compile without the patch which is why there's no assertions in it)

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -90,6 +90,17 @@ impl PyClassArgs {
...
+                    if let syn::Expr::Path(exp) = *(exp_grp.expr).clone() {
+                        if exp.path.segments.len() == 1 {
+                            self.name = Some(exp.clone().into());
+                        } else {
+                            expected!("Must be Name not Path::Name")
+                        }
+                    } else {
+                        expected!("type name (e.g., Name), found: {:?}", exp_grp)
+                    }
+                }
```
kngwyu: `expected` macro does not accept formatting args. Just `type name` would be sufficient.
And this can be simplified like
```suggestion
                // For macro args(e.g., $name)
                syn::Expr::Group(exp_grp) => match &*exp_grp.expr {
                    syn::Expr::Path(exp) if exp.path.segments.len() == 1 => {
                        self.name = Some(exp.clone().into());
                    }
                    _ => expected!("type name (e.g., Name)"),
                },
```

tests/test_macros.rs
```diff
@@ -0,0 +1,13 @@
+#![feature(specialization)]
```
kngwyu: I don't want to have too many test files.
Could you please move this test to `test_class_new` or so?

kngwyu:
Thanks :+1:

tests/test_macros.rs
gilescope: Hold fire on that thought. I will add in the additional tests and we can agree where to put them once I’ve made sure we can macro_rules create anything with an attribute.

programmerjake:
Encountered this issue myself.

One suggestion: maybe there should be an `unwrap_group` function that unwraps a group or returns the input unchanged:

```rust
// pseudo-code, probably needs some derefs and other misc. cleanup
fn unwrap_group(expr: &syn::Expr) -> &syn::Expr {
    if let syn::Expr::Group(g) = expr {
        &*g.expr
    }
    expr
}
```

that way, the `match` cases can be simplified:
```rust
            "name" => match unwrap_group(&**right) {
                syn::Expr::Path(exp) if exp.path.segments.len() == 1 => {
                    self.name = Some(exp.clone().into());
                }
                _ => expected!("type name (e.g., Name)"),
            },
```

scalexm:
Btw, I would suggest a `while let Group(g) = expr` approach rather than unwrapping a single layer. Not sure if that could actually happen but it’s harmless.

davidhewitt:
👍 also my parsing improvements in #1567 might resolve a lot of the problems here. I plan to retest this once that PR is merged.

davidhewitt:
As this branch is very old, I just rebased it and will merge it if tests are green 🤞

scalexm:
More of these problems started to appear with Rust 1.53 (at first I thought it was a compiler régression, but actually it’s just the compiler adding a few more groups and pyo3 not handling them correctly).

I think we should use `unwrap_group` quite liberally: I have found a few more places where we need it. Once this is merged, I’ll open a follow-up PR.

messense (approved):

PR #835, opened 2020-03-26
Remove GILPool::new-no-pointer

kngwyu:
So what is problematic is this code.

```rust
    unsafe extern "C" fn tp_dealloc_callback<T>(obj: *mut ffi::PyObject)
    where
        T: PyClassAlloc,
    {
        let _pool = gil::GILPool::new_no_pointer(py);
        <T as PyClassAlloc>::dealloc(py, (obj as *mut T::Layout) as _)
    }
```
It means that, **if a rust struct(pyclass) `T` drops, we don't call decref to all `PyObject` and `Py<...>` that `T` has**.

This can cause memory leak when 
- `T` has `PyObject` or `Py<...>`
- `GILGuard` is not required after `T` drops

So here `GILPool::new(py)` should be used.
And since `new_no_pointer` is not used for other usage, I decided to remove this function.

I found this bug via [a question(in Japanese site)](https://teratail.com/questions/249003) by @namuyan.  Thanks for that.


tests/test_gc.rs
```diff
@@ -81,52 +81,6 @@ fn data_is_dropped() {
...
 
-#[pyclass]
-struct ClassWithDrop {}
-
-impl Drop for ClassWithDrop {
-    fn drop(&mut self) {
-        unsafe {
-            let py = Python::assume_gil_acquired();
-
-            let _empty1: Py<PyTuple> = FromPy::from_py(PyTuple::empty(py), py);
```
kngwyu: So what is this test for?
@konstin do you know?

konstin: Afaik they are stored in the pointer list belonging to the GIL wrapper, so we can observe in `create_pointers_in_drop` that they are only dropped when our GIL wrapper cleans up .

PR #839, opened 2020-03-28
Make __r*__ methods work with operators

### LINKED ISSUE #838: Binary operators don't work for `__r*__` methods
The implementation as presented [here](https://github.com/PyO3/pyo3/blob/d7fe7a92a5e473bcfb46414c2ae348d06e6a3600/tests/test_arithmetics.rs#L190-L198):

```rust
#[pyclass]
struct RhsArithmetic {}

#[pyproto]
impl PyNumberProtocol for RhsArithmetic {
    fn __radd__(&self, other: &PyAny) -> PyResult<String> {
        Ok(format!("{:?} + RA", other))
    }
}
```
leads to the following behaviour when called from python:

```python
# rhs is an instance of RhsArithmetic 
1.0 + rhs         # error:  unsupported operand type(s) for +: 'float' and 'RhsArithmetic'
1.0.__add__(rhs)  # error: NotImplemented
rhs.__radd__(1.0) # works as intended
```
while we'd expect that `1.0 + rhs` should fall back to `rhs.__radd__(1.0)` which is not working.

From the disussion in gitter with @kngwyu :
> We have to fill nb_add field in [PyNumberMethods](https://docs.python.org/3/c-api/typeobj.html?highlight=pynumbermethods#c.PyNumberMethods) if the object has __radd__.

kngwyu:
Resolves #838. 
This PR uses `__radd__` methods as reversed fallback methods for `__add__` or so.
This is the same as [what cpython does](https://github.com/python/cpython/blob/v3.8.2/Objects/typeobject.c#L6996) and I'm sure it's a correct strategy, but, apparently, this implementation is dirty and has lots of templates :roll_eyes: 
Need to simplify in the future.


tests/test_arithmetics.rs
```diff
@@ -195,6 +196,38 @@ impl PyNumberProtocol for RhsArithmetic {
...
+
+    fn __rxor__(&self, other: &PyAny) -> PyResult<String> {
+        Ok(format!("{:?} ^ RA", other))
+    }
+
+    fn __ror__(&self, other: &PyAny) -> PyResult<String> {
+        Ok(format!("{:?} | RA", other))
+    }
+
+    fn __rpow__(&self, other: &PyAny, _module: &PyAny) -> PyResult<String> {
```
davidhewitt:
```suggestion
    fn __rpow__(&self, other: &PyAny, _mod: &PyAny) -> PyResult<String> {
```

pyo3-derive-backend/src/defs.rs
```diff
@@ -7,6 +7,21 @@ pub struct Proto {
     pub py_methods: &'static [PyMethod],
 }
 
+impl Proto {
+    pub(crate) fn get_proto<Q>(&self, query: Q) -> Option<&'static MethodProto>
+    where
+        Q: PartialEq<&'static str>,
```
davidhewitt: I think this might work without needing to be a reference?

```suggestion
        Q: PartialEq<str>,
```

src/class/number.rs
```diff
@@ -647,22 +647,22 @@ where
 {
     fn tp_as_number() -> Option<ffi::PyNumberMethods> {
         Some(ffi::PyNumberMethods {
-            nb_add: Self::nb_add(),
-            nb_subtract: Self::nb_subtract(),
-            nb_multiply: Self::nb_multiply(),
+            nb_add: Self::nb_add().or_else(Self::nb_add_fallback),
```
davidhewitt: So I haven't followed the rest of the logic in too much detail, but I have a question here.

If a `#[pyclass]` defines both `__add__` and `__radd__`, it looks like `__add__` wins here. What happens to `__radd__` - can it still be called?

src/class/number.rs
```diff
@@ -1407,14 +1407,72 @@ where
     }
 }
 
+// Fallback trait for nb_add
+trait PyNumberAddFallback {
```
davidhewitt: Why `PyNumberAddFallback` instead of `PyNumberRAdd` ? (Same for the rest of the new traits & functions.)

kngwyu: There's `PyNumberRAddProtocol` and `PyNumberRAddProtocolImpl` so we need a one that is clearly different from them.
We can change the implementation to use only `PyNumberRAddProtocol`, but it would need many changes in our proc-macro code. So I chose to use two traits for this, though it needs many boilerplate codes.
I don't think `RAdd` and `AddFallback` pair is not so bad, but if you come up with another idea, please let me know.

src/class/number.rs
kngwyu: > If a #[pyclass] defines both __add__ and __radd__, it looks like __add__ wins here

Yeah, but this is the same as what CPython does.
```python
In [4]: class C: 
   ...:     def __add__(self, other): 
   ...:         return f"Self + {other}" 
   ...:     def __radd__(self, other): 
   ...:         return f"{other} + Self" 
   ...:     def __repr__(self): 
   ...:         return "C" 
   ...:                                                                         

In [5]: c = C()                                                                 

In [6]: c + c                                                                   
Out[6]: 'Self + C'
```

pyo3-derive-backend/src/defs.rs
kngwyu: It takes `proc_macro2::Ident` as an argument, which implements `PartialEq<AsRef<str>>`.

src/class/number.rs
davidhewitt: That makes sense. What I mean is should we add a test for a class in `test_arithmetic` which defines both `__add__` and `__radd__`, and check everything works as expected? 

E.g. `c + c -> __add__` , `c + 1 -> __add__`, `1 + c -> __radd__` .

pyo3-derive-backend/src/defs.rs
davidhewitt: `str` implements `AsRef<str>` ? Maybe I'm missing something.

kngwyu: `PartialEq::eq` takes `&Other` so at last it needs `&T where T: AsRef<str>`.

davidhewitt: Sorry, yes, I'm having a brain misfire! 🤦‍♂

tests/test_arithmetics.rs
```diff
@@ -256,6 +256,47 @@ fn rhs_arithmetic() {
...
+    }
+}
+
+#[test]
+fn lhs_override_rhs() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+
+    let c = PyCell::new(py, LhsAndRhsArithmetic {}).unwrap();
+    py_run!(py, c, "assert c.__radd__(1) == '1 + BA'");
```
davidhewitt: So if I understand this test, it means that it's never possible to call `fn __radd__` ?

(`fn __radd__` has `RA` in output, this has `BA`.)

kngwyu: Oops, this can be a bug.
I'll check it tommorow.

kngwyu: Fixed by using [METH_COEXIST](https://docs.python.org/3/c-api/structures.html?highlight=coexist#METH_COEXIST)

davidhewitt:
👍 nicely figured out! Looks like `METH_COEXIST` should be a performance improvement too, from the python docs.

kngwyu:
It looks like we can't use COEXIST for `__rpow__` since it can take 3rd argument...

davidhewitt:
> 
> 
> It looks like we can't use COEXIST for `__rpow__` since it can take 3rd argument...

I'm not sure that's true... isn't the issue that `__pow__` and `__rpow__`'s `modulo` argument should be `Option<Self::Modulo>`?

e.g. the supported forms should be

```
# 2-argument forms
pow(c, 1) -> c.__pow__(1)
pow(1, c) ->  c.__rpow__(1)

# 3-argument form
pow(c, 1, mod) -> c.__pow__(1, mod)
```

According to the note [here](https://docs.python.org/3/reference/datamodel.html?highlight=__pow__#object.__rpow__), `pow(1, c, mod)` will not call `__rpow__`.

kngwyu:
> I'm not sure that's true... isn't the issue that __pow__ and __rpow__'s modulo argument should be Option<Self::Modulo>?

:+1: 
Changed to do so.

kngwyu:
Sign, coexist + __round__ cause segfault that I can't reproduce in my local machine... :nauseated_face:

davidhewitt:
> 
> 
> Sign, coexist + **round** cause segfault that I can't reproduce in my local machine... 🤢

😢 I'll try later on mine and let you know if I have any luck!

kngwyu:
@davidhewitt 
Simply I think we shouldn't use METH_COEXIST for `__round__` since the builtin `round` can pass some arguments.

davidhewitt:
> 
> 
> @davidhewitt
> Simply I think we shouldn't use METH_COEXIST for `__round__` since the builtin `round` can pass some arguments.

Ok. I guess also there is no need for `METH_COEXIST` with `__round__` because there is no such thing as `__rround__` 😄

davidhewitt:
Actions CI failure this time was the usual datetime fail, I'm rerunning jobs...

kngwyu:
@davidhewitt
Thank you for fixing clippy warnings!

PR #848, opened 2020-04-06
Reduce build dependencies from regex

ijl:
This avoids users compiling four crates for the "perf" and
"unicode" features. regex is only used in build.rs to parse
some #define lines so it doesn't need to compile support for
thread-local caches etc or parsing by unicode code point.


kngwyu:
Good catch, thanks!

PR #849, opened 2020-04-06
_PyDict_NewPresized()

ijl:
This links https://github.com/python/cpython/blob/master/Objects/dictobject.c#L1323-L1324.


kngwyu:
Thanks!

Issue #800, opened 2020-03-10
Python::assume_gil_acquired() can infer static lifetime.


davidhewitt:
As per [the comment on #797](https://github.com/PyO3/pyo3/pull/797#discussion_r390146736) we should probably remove `Python::assume_gil_acquired()`, because it can give you a `Python<'static>` thanks to the lifetime inference. This is almost certainly unsound.

I think it probably explains this crash #653 

See this playground for a minimal example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=213173005786e15199cf0312079c3d72

A function `assume_gil_acquired()` would still be useful, but I think it needs to return a type without a lifetime, so that there is no risk of incorrect inference. For example, something like this:

```
struct AssumeGILAcquired(Unsendable);

fn assume_gil_acquired() -> AssumeGilAcquired { ... }

impl AssumeGILAcquired {
    fn py(&self) -> Python { ... }
}

/// later on:
let gil = assume_gil_acquired();
let py = gil.py();
```

In the above example, the `Python` lifetime is tied exactly to the lifetime of the `AssumeGILAcquired` token, so we can avoid `'static` lifetime problems.

kngwyu:
Internally, we have lots of codes like
```rust
        let py = Python::assume_gil_acquired();
        let _pool = gil::GILPool::new_no_pointers(py);
```
Here `py` has to have the same lifetime as `_pool` so, a possible fix is 
```rust
struct GILPool { ... }
let _pool = GILPool::new();
let py = _pool.py(); // the existance of GILPool already assumes GIL is hold
```

kngwyu:
For now, could I open a PR that adds `GILPool::py`?
I'm not sure we need `AssumeGILAcquired` or not, but they can coexist.
But I'm afraid the change would conflict with #797.

davidhewitt:
> For now, could I open a PR that adds GILPool::py?

Sure thing, I'd be happy to see the unsoundness fixed sooner rather than later.

Don't worry about #797. I'll fix it up when I get a chance; I've been quite busy recently. When I have some time again I was planning to first help with bugfixes for 0.9 before we begin making breaking changes like #797 for an 0.10 release

kngwyu:
Ooops, because of the lifetime problem, it looks like we need the same trick as #797 to use `GILPool::py`. So maybe it would be better to merge it first.

davidhewitt:
[...]

PR #851, opened 2020-04-07
Improve performance on pointer drop

davidhewitt:
There's a `Mutex` involved each time we drop a `PyObject` or `Py<T>`, so that we can safely push the reference into the shared `Vec`. I have been wondering for a while if we could add a simple check if the GIL is held (as a `thread_local!` variable) and if so, decrease the refcount immediately.

I tried this out - turns out in the simple benchmark I made (`bench_object`) that adding this check makes the benchmark 5x faster.

I think decreasing the reference count sooner is also more predictable from a pyo3 user perspective. It helps with some of the issues around #311. (Though this doesn't improve the situation for references.)

If the GIL is not held, then I made it so that the pointers will be released as soon as the GIL is next acquired.

To track when the GIL is held I have added counters to `GILGuard::new()` and `GILGuard:drop()`.

For this optimization to apply all the time we also need to increment / decrement this counter when we call `assume_gil_acquired()` - which suggests to me that we should wait to finish this off once #800 is solved.

TODO:
 - [x] ~~Also increment / decrement GIL_COUNT when we call `assume_gil_acquired()`~~ (solved by also using `GILPool` for the counter)

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -3,11 +3,32 @@
 //! Interaction with python's global interpreter lock
 
 use crate::{ffi, internal_tricks::Unsendable, PyAny, Python};
+use std::cell::Cell;
 use std::ptr::NonNull;
 use std::{any, sync};
 
 static START: sync::Once = sync::Once::new();
 static START_PYO3: sync::Once = sync::Once::new();
+
+thread_local! {
+    /// This is a internal counter in pyo3 monitoring whether this thread has the GIL.
+    ///
+    /// It will be incremented whenever a GIL-holding RAII struct is created, and decremented
+    /// whenever they are dropped.
+    ///
+    /// As a result, if this thread has the GIL, GIL_COUNT is greater than zero.
+    static GIL_COUNT: Cell<u32> = Cell::new(0);
+}
+
+/// Check whether the GIL is acquired.
+///
+/// Note: This uses pyo3's internal count rather than PyGILState_Check for two reasons:
+///  1) for performance
+///  2) PyGILState_Check always returns 1 if the sub-interpreter APIs have ever been called,
+///     which could lead to incorrect conclusions that the GIL is held.
+fn gil_is_acquired() -> bool {
+    GIL_COUNT.with(|c| c.get() > 0)
+}
 
 /// Prepares the use of Python in a free-threaded context.
 ///
@@ -110,6 +131,8 @@ impl Drop for GILGuard {
             pool.drain(self.python(), self.owned, self.borrowed);
             ffi::PyGILState_Release(self.gstate);
         }
+
+        decrement_gil_count();
     }
 }
 
@@ -177,6 +200,7 @@ pub struct GILPool<'p> {
 impl<'p> GILPool<'p> {
     #[inline]
     pub fn new(py: Python) -> GILPool {
+        increment_gil_count();
         let p: &'static mut ReleasePool = unsafe { &mut *POOL };
         GILPool {
             py,
@@ -193,6 +217,7 @@ impl<'p> Drop for GILPool<'p> {
             let pool: &'static mut ReleasePool = &mut *POOL;
             pool.drain(self.py, self.owned, self.borrowed);
         }
+        decrement_gil_count();
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/gil.rs
```diff
@@ -233,7 +248,12 @@ impl GILGuard {
 
         unsafe {
             let gstate = ffi::PyGILState_Ensure(); // acquire GIL
+            GIL_COUNT.with(|c| c.set(c.get() + 1));
+
+            // Release objects that were dropped since last GIL acquisition
             let pool: &'static mut ReleasePool = &mut *POOL;
+            pool.release_pointers();
```
kngwyu, line 255: Why do we need to release pointers here?
To reduce memory usage?

benches/bench_object.rs
```diff
@@ -0,0 +1,16 @@
+#![feature(test)]
```
kngwyu, line 1: (For the file name) how about bench_pyobject or so?

kngwyu:
Great, thanks!

> For this optimization to apply all the time we also need to increment / decrement this counter when we call assume_gil_acquired()

I'm not sure how the next `assume_gil_acquired` API would be so I think it's sufficient if `GILPool`  increments/decrements the counter for now.

src/gil.rs
davidhewitt: Two reasons:

1. Yes, memory usage - we're now releasing the pointers sooner
2. This (https://github.com/PyO3/pyo3/issues/679#issuecomment-562964486) suggested to me that it's important in the long run we migrate to releasing objects on GIL acquire, not GIL exit.

benches/bench_object.rs
davidhewitt: Done 👍

davidhewitt:
Good idea. I now use both `GILGuard` and `GILPool` for the counter, which should cover all use cases.

src/gil.rs
```diff
@@ -3,12 +3,33 @@
...
+    /// whenever they are dropped.
+    ///
+    /// As a result, if this thread has the GIL, GIL_COUNT is greater than zero.
+    static GIL_COUNT: Cell<u32> = Cell::new(0);
+}
+
+/// Check whether the GIL is acquired.
+///
+/// Note: This uses pyo3's internal count rather than PyGILState_Check for two reasons:
+///  1) for performance
```
davidhewitt, line 26: For reference, I found that checking with `PyGILState_Check` was only 2x faster in the benchmark, versus 5x faster with this internal counter.

davidhewitt:
Formatting error again :facepalm: I'll fix later

src/gil.rs
kngwyu: Ah, I understand. Thank you for the reference.

davidhewitt:
Re-pushed with fixed formatting; should be green in CI now I hope.

src/gil.rs
```diff
@@ -250,6 +284,20 @@ impl GILGuard {
     }
 }
 
+/// Increment pyo3's internal GIL count - to be called whenever GILPool or GILGuard is created.
+fn increment_gil_count() {
```
kngwyu, line 288: I recommend using`#[inline(always)]` for this function.

src/gil.rs
```diff
@@ -250,6 +284,20 @@ impl GILGuard {
...
+    GIL_COUNT.with(|c| c.set(c.get() + 1))
+}
+
+/// Decrement pyo3's internal GIL count - to be called whenever GILPool or GILGuard is dropped.
+fn decrement_gil_count() {
+    GIL_COUNT.with(|c| {
+        let current = c.get();
+        assert!(current > 0);
```
kngwyu, line 296:
```suggestion
        debug_assert!(
            current > 0,
            "Negative GIL count is detected. Please report this error to PyO3 repo as a bug."
        );
```

Or any other nice error message.

src/gil.rs
```diff
@@ -3,12 +3,33 @@
...
+    static GIL_COUNT: Cell<u32> = Cell::new(0);
+}
+
+/// Check whether the GIL is acquired.
+///
+/// Note: This uses pyo3's internal count rather than PyGILState_Check for two reasons:
+///  1) for performance
+///  2) PyGILState_Check always returns 1 if the sub-interpreter APIs have ever been called,
+///     which could lead to incorrect conclusions that the GIL is held.
+pub(crate) fn gil_is_acquired() -> bool {
```
kngwyu, line 29: Why is this function `pub(crate)`?

davidhewitt: I wanted to be conservative about what we add to our public API. 

I thought other modules inside `pyo3` might need it, which is why I chose `pub(crate) fn` rather than just `fn`.

What would you prefer this to be?

kngwyu: > I thought other modules inside pyo3 might need it, which is why I chose pub(crate) fn rather than just fn.
What would you prefer this to be?

I'm sorry but now I cannot judge which is better since I can't imagine other usages of this function.
I was just worried about the case you forgot to remove `pub(crate)`. If not, I want to let you decide.

davidhewitt: Ok for now I will just go for `fn`. 👍

kngwyu:
Thank you!

PR #852, opened 2020-04-07
Fix clippy complaint of empty return type

davidhewitt:
Clippy on recent nightly is telling me that the empty return type is not necessary. (` -> () `)


kngwyu:
Thank you!

Issue #854, opened 2020-04-08
Allow non-mutable reference in PyIterProtocol::__iter__


althonos:
This is both a bug report and an improvement request concerning the `PyIterProtocol` trait.

### 🌍 Environment

 - Your operating system and version: Archlinux, Linux 5.3.8
 - Your python version: Python 3.8
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: `pacman`
 - Your rust version (`rustc --version`): `1.43.0-nightly`
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` using `master`

### 💥 Reproducing

(complete code in [althonos/nanoset.py](https://github.com/althonos/nanoset.py) repo on the `experimental` branch)

I have a Rust struct wrapping a single `PyObject`, which I expose both in the `PyIterProtocol` to call the wrapped object `__iter__`, and in the `PyGCProtocol` so that it's traversed and garbage collected as expected:

```python
#[pyclass(gc)]
pub struct MyWrapper {
    inner: PyObject,
}

#[pyproto]
impl PyIterProtocol for MyWrapper {
    fn __iter__(slf: PyMutRef<Self>) -> PyResult<PyObject> {
        let gil = Python::acquire_gil();
        let py = gil.python();
        slf.inner.call_method0(py, "__iter__")
    }
}

#[pyproto]
impl PyGCProtocol for MyWrapper {
    fn __traverse__(&'p self, visit: PyVisit) -> Result<(), PyTraverseError> {
        visit.call(self.inner)
    }

    fn __clear__(&'p mut self) {
        let gil = Python::acquire_gil();
        gil.python().release(self.inner)
    }
}
```

However, when running this within Python, I get some panics whenever the GC tries to traverse my object and i'm requesting iteration of the object at about the same time. That can happen when iterating on the object from within a function it was passed as an argument of. I get the following backtrace:
```
thread '<unnamed>' panicked at 'Already mutably borrowed: PyBorrowError', /home/althonos/.local/share/cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.9.1/src/pycell.rs:183:9
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::panicking::default_hook::{{closure}}
[...]

PR #855, opened 2020-04-08
Fix potential panics caused by Garbage Collector

althonos:
As discovered in #854, having a mutable borrow at any point on a `PyClass` that also implements the Garbage Collection protocol leads to unexpected panics, since the garbage collector can be run at any time, including when our class is already mutably borrowed.

As a solution, the wrapper for the garbage collector now calls `PyCell::try_borrow` instead of `PyCell::borrow`, and simply returns if it could not acquire the cell. I'm not sure about the expected behaviour here though, feedback is welcome.

## Maintenance 

 - Run `cargo fmt` :white_check_mark: 
 - Run `cargo clippy` and check there are no hard errors (There are a bunch of existing warnings; This is also checked by travis) :white_check_mark: 

### DIFF
--- a/tests/test_gc.rs
+++ b/tests/test_gc.rs
@@ -2,6 +2,7 @@ use pyo3::class::PyGCProtocol;
 use pyo3::class::PyTraverseError;
 use pyo3::class::PyVisit;
 use pyo3::prelude::*;
+use pyo3::type_object::PyTypeObject;
 use pyo3::{py_run, AsPyPointer, PyCell, PyTryInto};
 use std::cell::RefCell;
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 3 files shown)


kngwyu:
Really good catch, thank you!

> simply returns if it could not acquire the cell. 

I think it's OK, though it's really difficult to test. Maybe it's sufficient to check the `else` branch(=in case `try_borrow` fails) correctly?

althonos:
I'm trying to craft something with a method that blocks while borrowing, and a background thread that calls the GC, but I have no luck currently. I can confirm this test works on the codebase I'm using but since the test case (which comes from the CPython test suite) is very cryptic I can't port it as is.

davidhewitt:
Looks good to me, thanks 👍 

I think that doing nothing if already borrowed is probably fine for `tp_traverse` - the worst case is that the GC won't detect a cycle, and will do nothing.

I think we might have to look carefully in the future at `tp_clear`, which I think will only get called once and will need to succeed right away.

althonos:
Okay, I have a test case where I'm manually invoking the `tp_traverse` method as-if I was the GC; once while the instance is not borrowed, once when it is mutably borrowed. Without the fix the test panics, with the fix it doesn't. Both branches are covered according to `tarpaulin`.

tests/test_gc.rs
```diff
@@ -245,3 +246,63 @@ fn inheritance_with_new_methods_with_drop() {
...
+            _arg: *mut core::ffi::c_void,
+        ) -> std::os::raw::c_int {
+            0
+        }
+
+        // get the traverse function
+        let ty = TraversableClass::type_object().as_ref(py).as_type_ptr();
+        let traverse = (*ty).tp_traverse.unwrap();
+
+        // create an object and check that traversing it works normally
```
kngwyu, line 290: Good test :+1:

PR #858, opened 2020-04-08
Refactor CallbackConverter code

davidhewitt:
This is the portion of #797 which is related to `CallbackConverter` changes.

It makes use of a `run_callback`  function, which takes a closure to execute. The closure must return a `PyResult`.

A change of this fashion was originally necessary for #797. However I think it's good independent of #797, because a lot of the code is easier to read after these changes (in my opinion). Many `match` statements have been replaced with `?` as a result.

I factored this out because I think #797 still might need more design before we can settle.


kngwyu:
Thank you for doing all this hard work!

PR #862, opened 2020-04-09
Lock other threads before importing datetime

kngwyu:
Just as #860, I created this PR as a possible solution for #720. 
But not sure it really works.


davidhewitt:
Seems reasonable to me. I've closed #860

PR #863, opened 2020-04-10
Remove PyObject, rename PyAny -> PyObject

### LINKED ISSUE #356: Confusing APIs around PyObject

davidhewitt:
Fixes #356 

This is an experiment to solve the well-known confusion around `PyObject` and `PyAny` by removing `PyObject` and then renaming `PyAny` to `PyObject`, roughly as suggested in #679 

All the tests pass except for doc tests. I think there's a lot of documentation that would need fixing up.

I wanted to open this PR as an experiment; there will be plenty of things we can improve in it and I'm prepared to keep rebasing this and fixing it up against `master` as we go.

Most notable changes that I think might want discussion:
- I had to remove the implementation `FromPy<T> for T` to get conversions `Py<T> -> Py<PyObject>` to compile. We might want to rethink this. Maybe `min_specialization` feature would allow these conflicting impls.
- `ToPyObject::to_object()` trait now returns `&PyObject`. Not sure if maybe `Py<PyObject>` is better.

~~Bikeshed:~~

~~I've been wondering what happens if we renamed `PyObject` to just `Object` - I don't think the `Py` is particularly interesting and typing `Py<Object>` is heck of a lot less annoying than`Py<PyObject>`. This would be a good moment to do such a rename if we're breaking everything anyway.~~ Moved bikeshed to #873 

TODO:
 - [x] Thorough review of consequences of this change
   - [x] `FromPy` change
   - [x] `ToPyObject` change 
 - [x] Update guide
 - [x] Update docstrings


birkenfeld:
In general, I'm delighted with this.

> I've been wondering what happens if we renamed PyObject to just Object - I don't think the Py is particularly interesting and typing Py<Object> is heck of a lot less annoying than Py<PyObject>. This would be a good moment to do such a rename if we're breaking everything anyway.

Although `Py<Object>` looks good, I guess that would make all the concrete types candidates for renaming as well. `PyList` -> `List` etc?  In the reference form, this now loses the information that they're Python objects.

While we are bold, let me throw renaming `Py` in the ring: it's easily confused with the `Python` type, instances of which are usually called `py`.  Also the name doesn't *really* give a hint as to the smart pointer's purpose.

Unfortunately I don't have a good name in mind, let's call it `Own` here (which is bad, I know). So we'd have `Own<PyObject>`, `Own<PyList>` and `Own<MyStruct>` for Rust-based types...

kngwyu:
Thank you for the hard work!

> I've been wondering what happens if we renamed PyObject to just Object - I don't think the Py is particularly interesting and typing Py<Object> is heck of a lot less annoying thanPy<PyObject>. This would be a good moment to do such a rename if we're breaking everything anyway.

> While we are bold, let me throw renaming Py in the ring: it's easily confused with the Python type, instances of which are usually called py. Also the name doesn't really give a hint as to the smart pointer's purpose.

It sounds better to rename `Py` if we can find a good name, though I don't have a good name too.

davidhewitt:
I'm open to renaming `Py` as an alternative.

Given that `Py` is kinda like `Arc`, `Rc`, one name could be `Prc` aka "Python Reference Counted". (I'm reading that out loud as "Parc")

`Prc` does look weird but so did `Arc` and `Rc` when I started learning Rust.

Another named for `Py` could perhaps be `Unscoped` (because it's detaching GIL-scoped references from the GIL lifetime), but that's much longer...

davidhewitt:
Also, I'm kinda hoping in the long run that we don't use `Py` much because most things will work with GIL-scoped references. So `Py` only comes into use when things need to escape the GIL lifetime.

kngwyu:
Added `help-wanted` label since I want to hear more opinions about this naming issue.

althonos:
> I've been wondering what happens if we renamed PyObject to just Object - I don't think the Py is particularly interesting and typing Py<Object>

I'm kinda for keeping `PyObject` against `Object` for consistency with all the other types (`PyDict`, `PyList`, etc.), even with the `Py<PyObject>` situation. An interesting option is what [`rlua`](https://github.com/kyren/rlua) does; types are not `Lua` prefixed in the API, but they are in `rlua::prelude` to avoid confusion.

> Given that Py is kinda like Arc, Rc, one name could be Prc aka "Python Reference Counted". (I'm reading that out loud as "Parc")

Sounds good to make that more explicit.

programmerjake:
how about `PyRc` instead of `Py`? it's short and seems totally obvious.

davidhewitt:
> An interesting option is what rlua does; types are not Lua prefixed in the API, but they are in rlua::prelude to avoid confusion.

I find this option quite interesting too. In code where I'm interacting with Python a lot I feel that typing `Py` (in any form) a lot isn't particularly helpful, as I already know the code is in a Python context.

I think there's a lot of ideas here around naming of `PyObject vs Object`, `Py vs PyRc` etc. To keep this PR more focussed specifically on simplifying the current `PyAny` / `PyObject` situation, perhaps we should move the more general discussion into its own issue?

davidhewitt:
I did some cleanups and rebased. CI might even be green this time! I moved the question of names to #873.

davidhewitt:
> ToPyObject::to_object() trait now returns &PyObject. Not sure if maybe Py<PyObject> is better.

I have been thinking further about this and I think it is the correct decision. Especially in the future when we have `PyObject<'a>`, I think this offers the best ergonomics. (Also I think we want most code to use `&PyObject` / `PyObject<'a>` and only to use the `Py<PyObject>` form when necessary.)

src/instance.rs
```diff
@@ -131,38 +146,154 @@ impl<T> Py<T> {
...
+    // ) -> PyResult<PyObject> {
+    //     self.call_method(py, name, args, None)
+    // }
+
+    // /// Calls a method on the object with no arguments.
+    // ///
+    // /// This is equivalent to the Python expression `self.name()`.
+    // pub fn call_method0(&self, py: Python, name: &str) -> PyResult<PyObject> {
+    //     self.call_method(py, name, (), None)
+    // }
```
davidhewitt, line 278: These methods used to exist on `PyObject`. I'm trying to decide if we should move them to `Py<T>`, or delete them.

src/instance.rs
```diff
@@ -131,38 +146,154 @@ impl<T> Py<T> {
...
 ///     let py = gil.python();
-///     py.eval("[]", None, None).unwrap().to_object(py)
+///     py.eval("[]", None, None).unwrap().to_object(py).into()
 /// };
 /// let gil = Python::acquire_gil();
 /// let py = gil.python();
-/// assert_eq!(obj.as_ref(py).len().unwrap(), 0);  // PyAny implements ObjectProtocol
+/// assert_eq!(obj.as_ref(py).len().unwrap(), 0);  // PyObject implements ObjectProtocol
 /// ```
 pub trait AsPyRef: Sized {
```
davidhewitt, line 298: Now that the old `PyObject`  doesn't exist, I'm wondering if `AsPyRef` trait should be moved to be just a method on `Py<T>`. I don't think there is value in it being a trait any more.

davidhewitt:
@kngwyu this is now looking good to me; would be curious to hear what you have to say about it! 😄

birkenfeld:
I tried this with my (smallish) extension module, and there was no unexpected fallout.

I changed existing `PyObject` to `&PyObject` (in function args) and `Py<PyObject>` (elsewhere, e.g. in trait impls of `IntoPy`).  Then all `PyAny` to `PyObject`.

Finally, I just had remove a bunch of `py` arguments to methods that now didn't need them anymore.

:+1:

**However**, this needs a good step-by-step guide in the release notes, since the errors I got trying to compile the original code were mostly not particularly helpful.

davidhewitt:
> However, this needs a good step-by-step guide in the release notes, since the errors I got trying to compile the original code were mostly not particularly helpful.

I agree. It would be interesting to see if I could write [`rerast`](https://github.com/google/rerast) rules to automate migration. @birkenfeld is your extension module open source?

I also wondered about adding to `migration.md` in the guide.

birkenfeld:
@davidhewitt no, it's not unfortunately. I have a few other projects waiting for PyO3 bindings but can only start them once it compiles on stable...

kngwyu:
@birkenfeld 

> However, this needs a good step-by-step guide in the release notes, since the errors I got trying to compile the original code were mostly not particularly helpful.

Yeah, thank you for trying it out.

@davidhewitt 
I'm sorry but now I don't have enough time to review all changes. Please wait for 3~5 days.

But my main concern is... should we do this **now**?

This change would force users to replace many `PyObject` with `Py<PyObject>` since many of our APIs are built around `PyObject`(e.g., `IntoPy<PyObject>` is used everywhere).`

However, since `PyObject`(or `Py<PyObject>`) would be costlier than `PyAny` after #679 is solved, we should rewrite lots of `PyObject` APIs to use `PyAny`.
So if we do renaming now, users would have to
1. Replace `PyObject` with `Py<PyObject>`
2. Then (if we refactor `Py<PyObject>` APIs) replace `Py<PyObject>` with (new) `PyObject`...
Isn't it confusing?

Indeed, we cannot avoid this kind *curse of renaming* and at last, should replace `PyAny` with `PyObject`. 
But doing it now(= before refactoring the internal system of `PyAny` and conversion traits) doesn't seem so good idea to me.

davidhewitt:
@kngwyu that's a good question. I actually had started on this change by working on #679 in totality (both PyAny ownership change and PyAny rename).

It seemed too much to do in one go so I did this bit first.

I'd be happy to instead try out the `PyAny` ownership change first, if you prefer.

Give me a few days to make a draft PR for that. Don't worry about reviewing this one in the meanwhile :+1:

kngwyu:
> I'd be happy to instead try out the PyAny ownership change first, if you prefer.

Firstly, how about rethinking the usage of `IntoPy<PyObject>` or `ToPyObject`?
These two traits would be major burdens when replacing `PyObject` with `Py<PyObject>`.
But I don't any strong opinions about this. It's just a *proposal* and please feel free to start from other points.

davidhewitt:
> Firstly, how about rethinking the usage of IntoPy<PyObject> or ToPyObject?

:thumbsup: can have a look there. Though I think we might want to get to the "lightweight" `PyAny<'a>` before we try to redesign these traits much.

For example I foresee already that `IntoPy<PyAny<'a>>` is also slightly frustrating to type!

davidhewitt:
I've rebased this PR on master, and converted back to draft.

I think it's still interesting to continue to explore this branch, and I will continue to rebase it as we go.

You're right that we can probably improve the design of `IntoPy` and `ToPyObject` before we merge this. Also I would advise that we don't merge this in `0.10.0`, as that version is getting exciting enough already. Maybe save it for `0.11.0`. 😄

birkenfeld:
On the other hand, large incompatible changes with every version are also tiring...

davidhewitt:
That's fair - maybe we'll save it for `0.12.0`. 😆 

In all serious I'm happy to get this into `0.10.0` if we think we are in favour of moving ahead with this change. I think there are still open questions about what the conversion traits should look like (`IntoPy` and `ToPyObject`) - especially as we chose to merge #887 instead of #885.

kngwyu:
I know `PyAny` is criticized since it's confusing (see #356 and #388), but personally I'm too accustomed to the name. I need more feedbacks from users.

birkenfeld:
Personally, I'm fine with either name, as long as there's only one of them remaining.

(PyAny has the very slight advantage of not clashing with the ffi PyObject, but most users won't ever need the latter.)

davidhewitt:
This draft PR is now massively out-of-date. Also, I think that there are enough breaking changes with this PR that it probably doesn't make sense to push forward as-is with it until we've got a final design for the rest of the conversion traits.

As a result, I'm going to close this. I _do_ still think that the rough objective of this PR is still relevant - it'll shrink PyO3's range of names and types, which is a common criticism of the PyO3 API.

I've got a lot of notes on what I learned from trying this, which I'll write up as an issue later so that we can pick up this and continue to design API improvements.

PR #865, opened 2020-04-10
Add IntoPy<PyObject> for HashSet and BTreeSet

davidhewitt:
Spotted these were missing after a question on Gitter.


kngwyu:
Thank you for the PR!
I think our `get/set` API should also handle `ToPyObject`, though.

Could you please rebase on the master to make travis green?

davidhewitt:
Will fix up in a couple hours.

Regarding `ToPyObject` - I agree this UX is not great right now. I think supporting both with two blanket impls might need specialization, but maybe there's an alternative route.

ArniDagur:
I just wanted to say: Thanks for this PR! I'm really liking PyO3 so far. Keep up all the good work :)

PR #866, opened 2020-04-10
Tighten down locking on datetime tests

davidhewitt:
After the datetime tests failed again on my machine, I realised that `lock!()` needs to go _before_ the `acquire_gil()` call - otherwise released objects can still go into the "wrong" GILPool.

### DIFF
--- a/tests/test_datetime.rs
+++ b/tests/test_datetime.rs
@@ -66,9 +66,9 @@ macro_rules! lock {
 
 #[test]
 fn test_date_check() {
+    lock!();
     let gil = Python::acquire_gil();
     let py = gil.python();
-    lock!();
     let (obj, sub_obj, sub_sub_obj) = _get_subclasses(&py, "date", "2018, 1, 1").unwrap();
 
     assert_check_exact!(PyDate_Check, obj);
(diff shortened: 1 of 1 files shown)


programmerjake:
From what I can see, the `lock!()` macro effectively does nothing since each call creates a new mutex then locks it, every function gets it's own new mutex at every call so none of them are shared. what you probably wanted was to lock a global mutex since that may actually do something.

programmerjake:
The Travis CI failure is caused by parking_lot being updated for the renamed `asm` -> `llvm_asm` macro but the nightly compiler being too old to have the macro renamed yet.

programmerjake:
see https://github.com/rust-lang/rust/pull/68404

davidhewitt:
Yikes, sorry, you're right about the Mutex, I should have caught that in the original PR. I'll fix this PR and also create one to update the minimum nightly in the morning.

kngwyu:
> From what I can see, the lock!() macro effectively does nothing since each call creates a new mutex then locks it, every function gets it's own new mutex at every call so none of them are shared. 

Ah, sorry about that.
I'll open a PR to fix it.

davidhewitt:
Closed in favour of #867

PR #867, opened 2020-04-11
Fix synchronization of datetime tests

kngwyu:
As reported by @programmerjake  by [this comment](https://github.com/PyO3/pyo3/pull/866#issuecomment-612257127), the `lock!` macro used in the datetime tests doesn't work.
This PR
- Fixes the synchronization
and
- Remove `nighly` feature from parking_lot

@davidhewitt 
I'm sorry that this PR conflicts with #866. I wanted to fix my fault myself.

### DIFF
--- a/tests/test_datetime.rs
+++ b/tests/test_datetime.rs
@@ -1,5 +1,6 @@
 #![feature(concat_idents)]
 
+use parking_lot::{lock_api::RawMutex as _RawMutex, RawMutex};
 use pyo3::ffi::*;
 use pyo3::prelude::*;
 use pyo3::types::IntoPyDict;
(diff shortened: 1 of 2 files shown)


kngwyu:
I can still see an odd error :thinking: 
image

tests/test_datetime.rs
```diff
@@ -56,19 +57,14 @@ macro_rules! assert_check_only {
...
 // we need to stop other threads before calling `py.import()`.
-// TODO(kngwyu): Remove this macro
-macro_rules! lock {
-    () => {
-        let _mutex = std::sync::Mutex::new(());
-        let _lock = _mutex.lock().unwrap();
-    };
-}
+// TODO(kngwyu): Remove this variable
+const MUTEX: RawMutex = RawMutex::INIT;
```
davidhewitt, line 61: I _think_ a `const` is actually not a real variable until the location it's used, so this still creates multiple copies of the mutex.

I think that the suggestion to use `static` on Travis is correct, as then there will only be one shared mutex:

```suggestion
static Mutex: RawMutex = RawMutex::INIT;
```

kngwyu: Oops, thank you.

davidhewitt:
Looks like something weird has happened to travis again. I wonder if we should be migrating to github actions to simplify the CI?

kngwyu:
> Looks like something weird has happened to travis again.

No worry, it was because of a bug.

PR #869, opened 2020-04-12
Bound 'py lifetime by GILPool in the internal usecases

### LINKED ISSUE #653: Static references to Python objects are allowed, but unsafe

kngwyu:
Fixes #653. 
By using `GILPool::py` instead of `Python::assume_gil_acquired`, we can inhivit `&'static` reference as an argument.
This PR also pushes some cleanups around `POOL`.


src/gil.rs
```diff
@@ -187,42 +176,65 @@ impl ReleasePool {
...
+    value: UnsafeCell<Option<ReleasePoolImpl>>,
+}
+
+impl ReleasePool {
+    const fn new() -> Self {
+        Self {
+            value: UnsafeCell::new(None),
+        }
+    }
+    unsafe fn get_or_init(&self) -> &mut ReleasePoolImpl {
```
davidhewitt, line 189: It looks like this function isn't thread safe, so I guess you intended for this to be only used alongside the GIL to ensure no race conditions.

```suggestion
    /// # Safety
    /// This function should only be used when the GIL is acquired, to ensure thread safety.
    unsafe fn get_or_init(&self) -> &mut ReleasePoolImpl {
```

src/gil.rs
```diff
@@ -187,42 +176,65 @@ impl ReleasePool {
...
+        if (*self.value.get()).is_none() {
+            *self.value.get() = Some(ReleasePoolImpl::new());
+        }
+        (*self.value.get()).as_mut().unwrap()
```
davidhewitt, line 193:
```suggestion
        (*self.value).get_or_insert_with(ReleasePoolImpl::new)
```

CHANGELOG.md
```diff
@@ -10,8 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 * `PyObject` and `Py<T>` reference counts are now decremented sooner after `drop()`. [#851](https://github.com/PyO3/pyo3/pull/851)
-  * When the GIL is held, the refcount is now decreased immediately on drop. (Previously would wait until just before releasing the GIL.)
-  * When the GIL is not held, the refcount is now decreased when the GIL is next acquired. (Previously would wait until next time the GIL was released.)
```
davidhewitt, line 14: Why did you remove the second bullet?

src/gil.rs
```diff
@@ -187,42 +176,65 @@ impl ReleasePool {
...
+unsafe impl Sync for ReleasePool {}
 
 #[doc(hidden)]
-pub struct GILPool<'p> {
-    py: Python<'p>,
+pub struct GILPool {
     owned: usize,
     borrowed: usize,
     no_send: Unsendable,
 }
```
davidhewitt, line 206: Now that GILPool doesn't have `py`, I think we can change GILGuard to be 

```
struct GILGuard {
    gstate: ffi::PyGILState_State,
    pool: GILPool
}
```

and then we can simplify _a lot_ of the duplication in this file.

src/gil.rs
```diff
@@ -187,42 +176,65 @@ impl ReleasePool {
...
-impl<'p> GILPool<'p> {
+impl GILPool {
+    /// # Safety
+    /// This function requires that GIL is already acquired.
     #[inline]
-    pub fn new(py: Python) -> GILPool {
+    pub unsafe fn new() -> GILPool {
         increment_gil_count();
-        let p: &'static mut ReleasePool = unsafe { &mut *POOL };
+        let pool = POOL.get_or_init();
```
davidhewitt, line 214: Oh, I think I missed a `release_pointers()` call here:

```suggestion
        // Release objects that were dropped since last GIL acquisition
        let pool = POOL.get_or_init();
        pool.release_pointers();
```

Maybe this is why you removed the second bullet from the CHANGELOG?

src/gil.rs
```diff
@@ -187,42 +176,65 @@ impl ReleasePool {
...
     no_send: Unsendable,
 }
 
-impl<'p> GILPool<'p> {
+impl GILPool {
+    /// # Safety
+    /// This function requires that GIL is already acquired.
     #[inline]
-    pub fn new(py: Python) -> GILPool {
+    pub unsafe fn new() -> GILPool {
```
davidhewitt, line 212: 👍 I agree this is unsafe - and nice use of `# Safety`.

tests/ui/static_ref.stderr
```diff
@@ -0,0 +1,11 @@
...
+ --> $DIR/static_ref.rs:9:1
+  |
+9 | #[pymethods]
+  | ^^^^^^^^^^^-
+  | |          |
+  | |          `_pool` dropped here while still borrowed
+  | borrowed value does not live long enough
+  | cast requires that `_pool` is borrowed for `'static`
+  |
+  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
davidhewitt, line 11: Looks like this `note` was added in early feb: https://github.com/rust-lang/rust/commit/26c86a6a2828b1048fa2fd425595ef8fb475249a

So might need to bump minimum rust version slightly to keep travis happy?

davidhewitt:
Should we add a `# Safety` doc to `Python::assume_gil_acquired()` to specifically mention that `'static` lifetimes can easily be inferred but are never safe? Then this PR can probably also resolve #800 for now.

CHANGELOG.md
kngwyu: Ah, sorry, I didn't notice that these two lines point out different things.

src/gil.rs
kngwyu: :+1:

src/gil.rs
kngwyu: Seems nice :+1:

tests/ui/static_ref.stderr
kngwyu: Yeah, but since to find the minimum nightly by binary search is really hard, I just skip the test using `TRAVIS_JOB_NAME`.

kngwyu:
@davidhewitt 
Thank you for comments!
I think they are all addressed now.

CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+* When the GIL is held, the refcount is now decreased immediately on drop. (Previously would wait until just before releasing the GIL.)
+* When the GIL is not held, the refcount is now decreased when the GIL is next acquired. (Previously would wait until next time the GIL was released.)
```
davidhewitt, line 14: I meant these bullet points to be sub-bullets of the one above. But maybe you prefer them flattened?

```suggestion
  * When the GIL is held, the refcount is now decreased immediately on drop. (Previously would wait until just before releasing the GIL.)
  * When the GIL is not held, the refcount is now decreased when the GIL is next acquired. (Previously would wait until next time the GIL was released.)
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 * `__radd__` and other `__r*__` methods now correctly work with operators. [#839](https://github.com/PyO3/pyo3/pull/839)
 * Garbage Collector causing random panics when traversing objects that were mutably borrowed. [#855](https://github.com/PyO3/pyo3/pull/855)
+* `&'static Py~` is not allowed as arguments. [#869](https://github.com/PyO3/pyo3/pull/869)
```
davidhewitt, line 23:
```suggestion
* `&'static Py~` being allowed as arguments. [#869](https://github.com/PyO3/pyo3/pull/869)
```

src/python.rs
```diff
@@ -57,6 +57,10 @@ impl<'p> Python<'p> {
     /// Because the output lifetime `'p` is not connected to any input parameter,
     /// care must be taken that the compiler infers an appropriate lifetime for `'p`
     /// when calling this function.
+    ///
+    /// # Safety
+    /// The lifetime `'p` must be shorter than the period you *assume* that you have GIL.
+    /// I.e., `Python<'static>` is always *really* unsafe.
```
davidhewitt, line 63: 👍 nice. Not sure the comma is necessary.

```suggestion
    /// I.e. `Python<'static>` is always *really* unsafe.
```

CHANGELOG.md
kngwyu: Ah, sorry again, I was not aware of how they are structured.

src/python.rs
kngwyu: I didn't know but it looks a matter of Blitish/American English https://english.stackexchange.com/questions/6894/should-there-be-a-comma-after-i-e

davidhewitt: Heh that makes sense with me being British 😂

PR #871, opened 2020-04-13
Remove myself from the CoC

konstin:
I'm not active on this project anymore, so I shouldn't be the CoC contact.

@kngwyu @davidhewitt Feel free to add yourself instead.


kngwyu:
I'm kind of sad to see this, but @konstin thank you for all your work on PyO3!
I really learned a lot from working with you.

PR #886, opened 2020-05-01
Add dir() to ObjectProtocol.

m-ou-se:

### DIFF
--- a/src/objectprotocol.rs
+++ b/src/objectprotocol.rs
@@ -552,4 +552,22 @@ mod test {
         let obj = py.eval("42", None, None).unwrap();
         assert_eq!(unsafe { obj.get_type().as_type_ptr() }, obj.get_type_ptr())
     }
+
+    #[test]
+    fn test_dir() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let obj = py.eval("42", None, None).unwrap();
+        let dir = py
+            .eval("dir(42)", None, None)
+            .unwrap()
+            .extract::<&PyList>()
+            .unwrap();
+        let a = obj
+            .dir()
+            .into_iter()
+            .map(|x| x.extract::<String>().unwrap());
+        let b = dir.into_iter().map(|x| x.extract::<String>().unwrap());
+        assert!(a.eq(b));
+    }
 }
--- a/src/objectprotocol.rs
+++ b/src/objectprotocol.rs
@@ -3,7 +3,7 @@
 use crate::class::basic::CompareOp;
 use crate::err::{self, PyDowncastError, PyErr, PyResult};
 use crate::exceptions::TypeError;
-use crate::types::{PyAny, PyDict, PyIterator, PyString, PyTuple, PyType};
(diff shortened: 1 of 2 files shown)


kngwyu:
@m-ou-se 
#888 would resolve the travis failure.
In addition, any test for this API is appriciated.

davidhewitt:
A general comment about these kinds of functions - `dir`, `repr` etc.

At the moment they're trait methods, but I personally would prefer them to be free functions like this:

```
fn dir<T>(obj: T) -> &PyList
where T: AsRef<PyAny>
{ ... }

// Use later as 
pyo3::dir(obj);
```

It's not a _strong_ feeling, but I think it is more consistent with how these work in Python, where they're also used as `dir(obj)`, `repr(obj)` etc.

Wonder how you feel about this? 

(If we wanted compatibility we could leave them on `ObjectProtocol` but mark as deprecated?)

kngwyu:
> A general comment about these kinds of functions - dir, repr etc.

A bit off-topic and should be discussed in another thread?

davidhewitt:
> A bit off-topic and should be discussed in another thread?

I thought it was relevant given we're adding `dir` now, but sure I'll move this to another issue.

davidhewitt:
Many thanks @m-ou-se for updating this PR. As I guess you have a need for `dir()`, I'd be interested to hear what you think of the idea proposed in #892 . (We could have `pyo3::dir()` and `o.__dir__()` in the way suggested by that PR.)

m-ou-se:
@davidhewitt Ah, didn't see that one yet. Looks good.

davidhewitt:
> 
> 
> @davidhewitt Ah, didn't see that one yet. Looks good.

I think I pushed that PR about 20 seconds after you updated this branch¬ 😄  I'll wait for some more feedback to see what others think of that PR, but I will definitely make sure to add `dir()` .

src/objectprotocol.rs
```diff
@@ -545,4 +552,15 @@ mod test {
...
         assert_eq!(unsafe { obj.get_type().as_type_ptr() }, obj.get_type_ptr())
     }
+
+    #[test]
+    fn test_dir() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let obj = py.eval("42", None, None).unwrap();
+        let dir = py.eval("dir(42)", None, None).unwrap().extract::<&PyList>().unwrap();
+        let a = obj.dir().into_iter().map(|x| x.extract::<String>().unwrap());
```
kngwyu, line 562: rustfmt complains about this

```suggestion
        let dir = py
            .eval("dir(42)", None, None)
            .unwrap()
            .extract::<&PyList>()
            .unwrap();
        let a = obj
            .dir()
            .into_iter()
            .map(|x| x.extract::<String>().unwrap());
```

davidhewitt:
FWIW please do merge this PR without waiting for #892 - we might wait a little while to decide the right design there.

kngwyu:
Thank you!

Issue #883, opened 2020-04-30
Returning borrowed objects is probably always unsound


davidhewitt:
EDIT (Mistakenly posted the issue template, sorry if it looks weird in email notifications)

While digging around with lifetimes and borrowed objects just now I realised that using `py::from_borrowed_ptr()` for `PyDict::get_item` is fundamentally unsound.

If a user deletes the item from the dict after getting it, the user has easy access in safe code to a dangling PyObject.

This can easily be observed by the following code:

```rust
#[test]
fn test_dict_borrowed_object() {
    use crate::{ffi, AsPyPointer};

    let gil = Python::acquire_gil();
    let py = gil.python();

    let arr = [("a", 1234567)];
    let py_map = arr.into_py_dict(py);

    let item = py_map.get_item("a").unwrap();

    // The value has a ref count of 1, because only the dict owns it still
    assert_eq!(unsafe { ffi::Py_REFCNT(item.as_ptr()) }, 1);

    // Deleting it means the refcount goes to 0 and the borrowed object `item` is now dangling
    py_map.del_item("a").unwrap();

    // I know that Py_REFCNT will never return 0; this is more for illustration.
    // On my machine Py_REFCNT returns 2468790368784 here, but I don't know if it's deterministic.
    assert_eq!(unsafe { ffi::Py_REFCNT(item.as_ptr()) }, 0);

    // User could still use `item` and cause a segfault...
}
```

My feeling from this is that this means that returning borrowed objects from our Py* Rust APIs is probably almost always unsound for similar reasons.

We might want to convert these functions to return owned pointers.

davidhewitt:
Here's another experiment; it's possible to `.clear()` a list during iteration and have undefined behaviour - the borrowed object returned in iteration is pointing to invalid memory after the list is cleared.

I know these are contrived examples but my point is that there are subtle ways bugs may creep into software with borrowed pointers 😢 

```
let gil = Python::acquire_gil();
let py = gil.python();

let owned = vec![(1235443, 213213210), (1, 2)].into_py_dict(py).to_object(py);
drop(gil);

let gil = Python::acquire_gil();
let py = gil.python();
let dict = owned.cast_as::<PyDict>(py).unwrap();

for pair in dict {
  dict.clear();

  // Undefined behaviour! Use after object pointed to by borrowed reference is dropped
  dbg!(pair);
}
```

PR #887, opened 2020-05-01
New Native Types and Lighter GILPool

### LINKED ISSUE #679: Why object creation is slow in PyO3 and how to enhance it

kngwyu:
# TL; DR
This PR changes the internal representation of `PyAny` from `*mut ffi::PyObject` to `UnsafeCell<ffi::PyObject>`.
Thus, `&PyAny` was 'a pointer to a pointer', but now it is simply a pointer.
Related parts (e.g, GILPool) are also changed.
Partly resolves #679.

## What does this PR resolve?
We store almost all objects(except `PyObject` and `Py<T>`) to the internal storage, which costs when we get many objects.
This PR resolves 60% of this problem.
Since now `&PyAny` is a pointer, we don't need to store borrowed pointers to the object storage. So we can simply remove it. 50% of the problem is resolved by this.
However, for owned pointers, we still need the object storage so that we can do `obj.refcnt -= 1` when `gil` drops. But we can make the storage a bit faster than now. Now we use LinkedList for the storage so that the pointer to a pointer(=&PyAny) has a consistent address, but after this PR, we can use faster `Vec` since we don't use that reference. I believe this optimization resolves 10% of the problem.

## What does not this PR resolve?
We still have the object storage after this PR. How should we remove this?
Yeah, we can remove it if all constructors return `Py<T>` instead of `T`, say:
```rust
impl PyDict {
    pub fn new(py: Python) -> Py<PyDict> {}
}
```
`Py` can decrease its own reference count when dropping, so it would be efficient.

## Comparison with #885?

#885 also tries to resolve this problem by introducing `PyAny<'py>`, which is a GIL-bounded 'owned' object type. One thing I'm worried about `PyAny<'py>` is that it requires users to change too many codes.
The pro of this PR is **it does not change almost all API**, though it suggests that we should gradually change constructors to return `Py<T>`.

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -244,16 +244,14 @@ impl Drop for GILPool {
                 // while calling Py_DECREF we may cause other callbacks to run which will need to
                 // register objects into the GILPool.
                 let len = owned_objects.borrow().len();
-                for i in self.owned_objects_start..len {
-                    let ptr = owned_objects.borrow().get_unchecked(i).as_ptr();
-                    ffi::Py_DECREF(ptr);
+                if self.owned_objects_start < len {
+                    let rest = owned_objects
+                        .borrow_mut()
+                        .split_off(self.owned_objects_start);
+                    for obj in rest {
+                        ffi::Py_DECREF(obj.as_ptr());
+                    }
                 }
-                // If this assertion fails, something weird is going on where another GILPool that was
-                // created after this one has not yet been dropped.
-                debug_assert!(owned_objects.borrow().len() == len);
-                owned_objects
-                    .borrow_mut()
-                    .truncate(self.owned_objects_start);
             });
 
             OWNED_ANYS.with(|owned_anys| owned_anys.borrow_mut().truncate(self.owned_anys_start));
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -3,7 +3,8 @@
 //! Interaction with python's global interpreter lock
 
 use crate::{ffi, internal_tricks::Unsendable, Python};
-use std::cell::{Cell, UnsafeCell};
+use parking_lot::Mutex;
+use std::cell::{Cell, RefCell, UnsafeCell};
 use std::{any, mem::ManuallyDrop, ptr::NonNull, sync};
 
 static START: sync::Once = sync::Once::new();
@@ -16,6 +17,13 @@ thread_local! {
     ///
     /// As a result, if this thread has the GIL, GIL_COUNT is greater than zero.
     static GIL_COUNT: Cell<u32> = Cell::new(0);
+
+    /// These are objects owned by the current thread, to be released when the GILPool drops.
+    static OWNED_OBJECTS: RefCell<Vec<NonNull<ffi::PyObject>>> = RefCell::new(Vec::with_capacity(256));
+
+    /// These are non-python objects such as (String) owned by the current thread, to be released
+    /// when the GILPool drops.
+    static OWNED_ANYS: RefCell<Vec<Box<dyn any::Any>>> = RefCell::new(Vec::with_capacity(256));
 }
 
 /// Check whether the GIL is acquired.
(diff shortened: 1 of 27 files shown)


src/instance.rs
```diff
@@ -21,6 +21,15 @@ pub unsafe trait PyNativeType: Sized {
     fn py(&self) -> Python {
         unsafe { Python::assume_gil_acquired() }
     }
+    /// Cast `&PyAny` to `&Self` without no type checking.
+    ///
+    /// # Safety
+    ///
+    /// Unless obj is not an instance of a type corresponding to Self,
```
programmerjake, line 28: You have a accidental double negative.
Did you mean: 
```rust
/// Unless `obj` is an instance of a type corresponding to `Self`,
```

src/gil.rs
```diff
@@ -275,14 +269,9 @@ pub unsafe fn register_pointer(obj: NonNull<ffi::PyObject>) {
...
-pub unsafe fn register_owned(_py: Python, obj: NonNull<ffi::PyObject>) -> &PyAny {
+pub unsafe fn register_owned(_py: Python, obj: NonNull<ffi::PyObject>) {
     let pool = POOL.get_or_init();
-    &*(pool.owned.push_back(obj) as *const _ as *const PyAny)
-}
-
-pub unsafe fn register_borrowed(_py: Python, obj: NonNull<ffi::PyObject>) -> &PyAny {
-    let pool = POOL.get_or_init();
-    &*(pool.borrowed.push_back(obj) as *const _ as *const PyAny)
+    pool.owned.push(obj);
```
davidhewitt, line 274: Idea: could we make the object storage for owned pointers a `thread_local!`, rather than using the `'static` GILPool? Might be even faster and avoid potential threading issues (might even resolve #756).

programmerjake: If we're going to support Python on embedded platforms (not sure), we can't use `thread_local`, however, unless we currently support embedded platforms, I have no issues requiring `thread_local` support.

Some other thing we should think about, how does it interact with futures trying to move between threads (just make them non-`Send`?), also, does it work with scoped threads?

programmerjake: Also, does this force types like `Py<PyAny>` to be non-`Send`able?

davidhewitt: > If we're going to support Python on embedded platforms (not sure), we can't use thread_local, however, unless we currently support embedded platforms, I have no issues requiring thread_local support.

Oh really? Have you some docs on that? I never saw this issue with thread_local before.

Regarding all the `Send` interactions - I think even if `PyAny` is not `Send`, we can explicitly `impl Send for Py<T>`, right?

programmerjake: > > If we're going to support Python on embedded platforms (not sure), we can't use thread_local, however, unless we currently support embedded platforms, I have no issues requiring thread_local support.
> 
> Oh really? Have you some docs on that? I never saw this issue with thread_local before.

I didn't find docs that explicitly say that, however https://github.com/rust-lang/rust/issues/62918 alludes to that:

> Futures currently have a dependency on TLS for storing a pointer to` Context<'_>`.
> https://github.com/rust-lang/rust/blob/master/src/libstd/future.rs#L116
> 
> This makes usage in `no_std` environments quite difficult.

programmerjake: > Regarding all the `Send` interactions - I think even if `PyAny` is not `Send`, we can explicitly `impl Send for Py<T>`, right?

Yeah. I think that may be what we already do, though didn't check.

programmerjake: regarding embedded and TLS, https://github.com/rustasync/team/issues/42 (referenced from previously mentioned issue) is much more clear.

davidhewitt: Oh I see; I think what you're saying is that `thread_local!` isn't available in `no_std` applications? It would be interesting to think about pyO3 support for `no_std`, but I think it's a long way off. Also not sure how many users might want this?

programmerjake: In that case, I think we should just switch to `thread_local` for now, and, whenever we add `no_std` support, we could `cfg` something else in.

src/python.rs
```diff
@@ -3,17 +3,15 @@
...
 use crate::gil::{self, GILGuard};
-use crate::instance::AsPyRef;
-use crate::object::PyObject;
-use crate::type_object::{PyDowncastImpl, PyTypeInfo, PyTypeObject};
+use crate::type_object::{PyTypeInfo, PyTypeObject};
 use crate::types::{PyAny, PyDict, PyModule, PyType};
-use crate::{AsPyPointer, FromPyPointer, IntoPyPointer, PyTryFrom};
+use crate::{
+    ffi, AsPyPointer, AsPyRef, FromPyPointer, IntoPyPointer, PyNativeType, PyObject, PyTryFrom,
+};
```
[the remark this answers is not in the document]
kngwyu: Ah, sorry about that.

> Perhaps we could make guidelines as to what should be imported from crate, and what should be imported from a more specific path?

Hmm ... :thinking:  
I'll try to find a similar guideline in other projects.

src/instance.rs
kngwyu: Thanks!

src/gil.rs
kngwyu: >  could we make the object storage for owned pointers a thread_local!, rather than using the 'static GILPool?

Possibly we can, but I failed to implement it due to some odd errors... :thinking:

davidhewitt: @kngwyu I have just opened https://github.com/kngwyu/pyo3/pull/2 which implements `thread_local`

kngwyu:
Some benchmark results:
PyO3-bench1
PyO3-bench2

davidhewitt:
Nice - so I think from the benchmarks:

`drop_many_objects` is testing `ReleasePool` rather than `GILPool`, so `thread_local` performance there is actually because of the cleanups I did there at the same time as `thread_local`.

`iter_dict` is testing borrowed objects performance, which again `GILPool` doesn't change. But as I say in #883 I think we need to change dict iteration to use borrowed objects. @kngwyu do you think we should make that change and re-run benchmarks?

kngwyu:
> do you think we should make that change and re-run benchmarks?

I think we should do that in another PR.

davidhewitt:
> I think we should do that in another PR.

Ok. I'm prepared to write a follow up PR later to change to owned objects. 👍 

For now I have one further benchmark for you for `iter_dict`:

- Using `borrowed` objects (master): **~3500ms**
- Using `borrowed` objects (current): **~1500ms**
- Using `owned` objects + this PR: **~2350ms**
- Using `owned` objects + `PyAny<'py>`: **~2150ms**

My conclusions:
- this branch is much faster than master even with owned objects, so I really think we should use owned objects for dict iteration :)
- `PyAny<'py>` is not much faster than this branch, even with owned objects where `PyAny<'py>` is supposed to be at it's best. So I think that for now, we should use this branch and forget `PyAny<'py>`? Can always look again at `PyAny<'py>` in the future if we really want to optimise as much as possible...

(aside, if the scale on your chart is really about ~250 ms then your computer is 10x faster than mine - you using a desktop machine? The benchmark is my laptop using battery power, so worst case measurement 😄 )

kngwyu:
> you using a desktop machine? 

I'm using a desktop machine with Core i5-6500 + Arch Linux 5.6.4.

kngwyu:
> PyAny<'py> is not much faster than this branch, even with owned objects where PyAny<'py> is supposed to be at it's best.

Interesting. Thank you for your benchmark.

Since we can succeed to make the owned pool thread-safe, it seems good to keep `&PyAny` APIs for now.

src/object.rs
```diff
@@ -268,7 +262,7 @@ impl PyObject {
 impl AsPyRef for PyObject {
     type Target = PyAny;
     fn as_ref<'p>(&'p self, _py: Python<'p>) -> &'p PyAny {
-        unsafe { &*(self as *const _ as *const PyAny) }
+        unsafe { &*(self.as_ptr() as *const PyAny) }
```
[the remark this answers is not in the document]
kngwyu: Looks we can't do so...? :thinking:

PR #889, opened 2020-05-02
Unify PyMethodsInventoryDispatch and PyMethodsProtocol

kngwyu:
`PyMethodProtocol` is exposed as a public API, but it is actually really difficult to use without proc-macro.
So I decided to unify `PyMethodsInventoryDispatch` and `PyMethodsProtocol`, yielding `PyMethodsImpl`.

Since we need more inventory-related types to remove specialization, I think we need this kind of simplification not to make our code too complex.

### DIFF
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -213,7 +213,7 @@ fn parse_descriptors(item: &mut syn::Field) -> syn::Result<Vec<FnType>> {
 fn impl_inventory(cls: &syn::Ident) -> TokenStream {
     // Try to build a unique type that gives a hint about it's function when
     // it comes up in error messages
-    let name = cls.to_string() + "GeneratedPyo3Inventory";
+    let name = format!("Pyo3MethodsInventoryFor{}", cls);
     let inventory_cls = syn::Ident::new(&name, Span::call_site());
 
     quote! {
@@ -234,8 +234,8 @@ fn impl_inventory(cls: &syn::Ident) -> TokenStream {
             }
         }
 
-        impl pyo3::class::methods::PyMethodsInventoryDispatch for #cls {
-            type InventoryType = #inventory_cls;
+        impl pyo3::class::methods::PyMethodsImpl for #cls {
+            type Methods = #inventory_cls;
         }
 
         pyo3::inventory::collect!(#inventory_cls);
@@ -461,7 +461,7 @@ fn impl_descriptors(
     Ok(quote! {
         pyo3::inventory::submit! {
             #![crate = pyo3] {
-                type ClsInventory = <#cls as pyo3::class::methods::PyMethodsInventoryDispatch>::InventoryType;
+                type ClsInventory = <#cls as pyo3::class::methods::PyMethodsImpl>::Methods;
                 <ClsInventory as pyo3::class::methods::PyMethodsInventory>::new(&[#(#py_methods),*])
             }
         }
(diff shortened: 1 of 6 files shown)


pyo3-derive-backend/src/pyclass.rs
```diff
@@ -213,7 +213,7 @@ fn parse_descriptors(item: &mut syn::Field) -> syn::Result<Vec<FnType>> {
 fn impl_inventory(cls: &syn::Ident) -> TokenStream {
     // Try to build a unique type that gives a hint about it's function when
     // it comes up in error messages
-    let name = cls.to_string() + "GeneratedPyo3Inventory";
+    let name = format!("Pyo3MethodsInventoryFor{}", cls);
```
davidhewitt, line 216:
```suggestion
    let name = format!("PyMethodsOf{}", cls);
```

Longer name is also fine, just wondered if I could find a shorter one which worked.

src/class/methods.rs
```diff
@@ -133,20 +125,15 @@ pub trait PyMethodsInventory: inventory::Collect {
...
-/// The implementation of this trait defines which methods a Python type has.
-///
-/// For pyclass derived structs this is implemented by collecting all impl blocks through inventory
-pub trait PyMethodsProtocol {
-    /// Returns all methods that are defined for a class
-    fn py_methods() -> Vec<&'static PyMethodDefType>;
-}
+#[doc(hidden)] // Only to be used through the proc macros
+/// For pyclass derived structs, this trait collects method from all impl blocks using inventory.
+pub trait PyMethodsImpl {
```
davidhewitt, line 130:
```suggestion
pub trait PyClassMethods {
```

Don't think we have a `PyMethods` type, so `PyMethodsImpl` is slightly confusing imo.

src/class/methods.rs
```diff
@@ -133,20 +125,15 @@ pub trait PyMethodsInventory: inventory::Collect {
...
 
-/// The implementation of this trait defines which methods a Python type has.
-///
-/// For pyclass derived structs this is implemented by collecting all impl blocks through inventory
-pub trait PyMethodsProtocol {
-    /// Returns all methods that are defined for a class
-    fn py_methods() -> Vec<&'static PyMethodDefType>;
-}
+#[doc(hidden)] // Only to be used through the proc macros
+/// For pyclass derived structs, this trait collects method from all impl blocks using inventory.
```
davidhewitt, line 129: Usually I put attributes after doc? Maybe it doesn't work for `#[doc(hidden)]`

```suggestion
/// For pyclass derived structs, this trait collects method from all impl blocks using inventory.
#[doc(hidden)] // Only to be used through the proc macros
```

davidhewitt:
We probably need a `CHANGELOG` entry for removal / hiding of `PyMethodsProtocol`?

pyo3-derive-backend/src/pyclass.rs
kngwyu: Since this is a *generated* class, I think it's better if a user can understand that *this is generated* from the name if he or she uses `cargo expand`.

src/class/methods.rs
kngwyu: Yeah, but it's the implementation detail of `#[pymethods]`.

pyo3-derive-backend/src/pyclass.rs
davidhewitt: That's fair. `PyO3GeneratedMethodsFor{}` ? Your choice of name is also fine.

PR #890, opened 2020-05-02
Remove unsound return of borrowed objects

### LINKED ISSUE #883: Returning borrowed objects is probably always unsound

davidhewitt:
As discussed in #883, I think these are a few cases where we should be converting borrowed objects to owned objects for safety reasons.

Closes #883.

### DIFF
--- a/src/types/tuple.rs
+++ b/src/types/tuple.rs
@@ -72,10 +72,8 @@ impl PyTuple {
     pub fn get_item(&self, index: usize) -> &PyAny {
         assert!(index < self.len());
         unsafe {
-            // PyTuple_GET_ITEM return borrowed ptr; must make owned for safety (see #890).
-            let ptr = ffi::PyTuple_GET_ITEM(self.as_ptr(), index as Py_ssize_t);
-            ffi::Py_INCREF(ptr);
-            self.py().from_owned_ptr(ptr)
+            self.py()
+                .from_borrowed_ptr(ffi::PyTuple_GET_ITEM(self.as_ptr(), index as Py_ssize_t))
         }
     }
 
--- a/src/types/tuple.rs
+++ b/src/types/tuple.rs
@@ -70,12 +70,12 @@ impl PyTuple {
     ///
     /// Panics if the index is out of range.
     pub fn get_item(&self, index: usize) -> &PyAny {
-        // TODO: reconsider whether we should panic
-        // It's quite inconsistent that this method takes `Python` when `len()` does not.
         assert!(index < self.len());
         unsafe {
-            self.py()
-                .from_borrowed_ptr(ffi::PyTuple_GET_ITEM(self.as_ptr(), index as Py_ssize_t))
+            // PyTuple_GET_ITEM return borrowed ptr; must make owned for safety (see #890).
+            let ptr = ffi::PyTuple_GET_ITEM(self.as_ptr(), index as Py_ssize_t);
+            ffi::Py_INCREF(ptr);
+            self.py().from_owned_ptr(ptr)
         }
     }
 
@@ -83,7 +83,6 @@ impl PyTuple {
     pub fn as_slice(&self) -> &[PyObject] {
         // This is safe because PyObject has the same memory layout as *mut ffi::PyObject,
         // and because tuples are immutable.
-        // (We don't even need a Python token, thanks to immutability)
         unsafe {
             let ptr = self.as_ptr() as *mut ffi::PyTupleObject;
             let slice = slice::from_raw_parts((*ptr).ob_item.as_ptr(), self.len());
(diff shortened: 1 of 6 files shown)


davidhewitt:
In case we're worried about performance regressions, I made some benchmarks (from #891). I ran them on `master`, this PR, #887 , and also #887 with this PR cherry-picked on top.

Results: 

  | master | no-borrowed-objects | new-nativetypes | new-nativetypes + no-borrowed-objects
-- | -- | -- | -- | --
dict_get_item | 3160 +- 820 | 3120 +- 750 | 2280 +- 1600 | 2530 +- 1260
iter_dict | 3680 +- 1000 | 3720 +- 700 | 1620 +- 700 | 2330 +- 560
iter_list | 2090 +- 550 | 2340 +- 1300 | 1080 +- 590 | 1610 +- 430
list_get_item | 1080 +- 410 | 1100 +- 400 | 470 +- 230 | 740 +- 320
drop_many_objects | 4.4 +- 1.2 | 4.8 +- 0.4 | 3.8 +- 0.4 | 3.7 +- 0.9
iter_tuple | 780 +- 110 | 760 +- 560 | 790 +- 300 | 820 +- 340
tuple_get_item | 960 +- 250 | 950 +- 100 | 420 +- 200 | 580 +- 330

Conclusions:

- This PR does now slow down vs master much (maybe just a little)
- `new-nativetypes` is much much faster than master (yay!)
- Applying this PR after `new-nativetypes` causes greater slowdown because the difference between borrowed and owned objects is greater after that. But the combination is still faster than master 😄 

So I would be really keen to see this safety gap closed. If we are *really* worried about performance we could add `unsafe fn get_item_unborrowed` to `PyList` etc., but I don't think the performance difference is so great that it's worth it.

src/types/tuple.rs
```diff
@@ -70,20 +70,19 @@ impl PyTuple {
...
     /// Panics if the index is out of range.
     pub fn get_item(&self, index: usize) -> &PyAny {
-        // TODO: reconsider whether we should panic
-        // It's quite inconsistent that this method takes `Python` when `len()` does not.
         assert!(index < self.len());
         unsafe {
-            self.py()
-                .from_borrowed_ptr(ffi::PyTuple_GET_ITEM(self.as_ptr(), index as Py_ssize_t))
+            // PyTuple_GET_ITEM return borrowed ptr; must make owned for safety (see #890).
+            let ptr = ffi::PyTuple_GET_ITEM(self.as_ptr(), index as Py_ssize_t);
```
kngwyu, line 76: Do we need this trick even for tuple? :thinking:

davidhewitt: It's less critical for tuple I think, because it's immutable from Python - but it's possible to mutate a tuple in C / Rust!

I felt for extra peace of mind I would make this owned too. I can undo this one if you prefer.

kngwyu: > I can undo this one if you prefer.

I don't think it's the problem of whether I prefer or do not prefer it.
Since we don't have an API for deleting an object(= `del_item` returns an error), I don't see a clear advantage for using `owned` here.

davidhewitt: Ok. 👍

davidhewitt:
I've undone the change for `PyTuple::get_item`

kngwyu:
Thank you!

PR #891, opened 2020-05-02
New benchmarks

davidhewitt:
I threw some benchmarks together to evaluate #890. Thought it better to submit them in a separate PR.


benches/bench_pyclass.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn __iter__(slf: PyRef<'p, Self>) -> PyResult<Py<Iter>> {
+        let iter = Iter {
+            inner: slf.values.clone().into_iter(),
+        };
+        PyCell::new(slf.py(), iter).map(Into::into)
+    }
+}
+
+#[bench]
+fn create_many_pyclass(b: &mut Bencher) {
```
kngwyu, line 42: What this benchmark is for?

benches/bench_pyclass.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    b.iter(|| {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let obj = PyCell::new(
+            py,
+            MyClass {
+                obj: py.None(),
+                x: 0,
+                values: (0..10_000).into_iter().collect(),
```
kngwyu, line 73: It looks that clippy complains about this.
```suggestion
                values: (0..10_000).collect(),
```

davidhewitt:
👍 for now I've just removed `bench_pyclass`.

I did some benchmarking a while ago in Python using `timeit` where I found that a simple `pyclass` was slower than a simple `dict`. I'd like to add these benchmarks to the pyo3 repo eventually, but really they need to be `.py` benchmarks, so I will add them back another time.

kngwyu:
>  I'd like to add these benchmarks to the pyo3 repo eventually, but really they need to be .py benchmarks, so I will add them back another time.

We have [a benchmark](https://github.com/PyO3/pyo3/tree/master/examples/word-count#benchmark) using pytest. Maybe it's helpful to take a glance at it.

kngwyu:
Thank you!

davidhewitt:
> We have [a benchmark](https://github.com/PyO3/pyo3/tree/master/examples/word-count#benchmark) using pytest. Maybe it's helpful to take a glance at it.

Ah nice, I will be sure to play around with that some time 👍

PR #893, opened 2020-05-03
Close soundness hole with acquire_gil

### LINKED ISSUE #311: Documentaion about when we free memory

### LINKED ISSUE #864: Easy dangling PyObject with GILPool
This sample code produces a dangling `PyObject` after the `GILPool` is dropped:

```
use pyo3::prelude::*;
use pyo3::{ffi, GILPool, AsPyPointer};
use pyo3::types::PyList;

fn main() {
    let py = gil.python();
    let obj;

    {
        let pool = GILPool::new(py);
        obj = py.eval("object()", None, None).unwrap();
        println!("Count before pool drop {}", unsafe { ffi::Py_REFCNT(obj.as_ptr()) });
    }

    println!("Count after pool drop {}", unsafe { ffi::Py_REFCNT(obj.as_ptr()) });
}
```

As seen by the output:

```
Count before pool drop 1
Count after pool drop 1919707401104
```

I think what must be going on is that `obj` has a lifetime tied to `py`, but it is actually only valid for the lifetime of `pool`. Because of the wrong lifetime semantics the Rust compiler lets us create the dangling pointer.

I'm not sure yet exactly what the solution should be. Maybe the change to `GILPool / py` relationship proposed in #800 will mitigate this issue.

I wonder whether this issue combined with #756 is telling us that the right design is really `PyAny<'a>` as proposed in #679

davidhewitt:
After thinking on #864 some more, I realised we now have an easy route to make this API sound.

If a `GILPool` already exists when `Python::acquire_gil()` is called, we should not create another `GILPool`.

This ensures that users cannot in safe code accidentally create a `GILPool` and allow references to outlast it. For those that still really want to clear memory manually, they can still use `GILPool::new()` and drop that pool when they want.

Fixes #864 
Fixes #311

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -12,8 +12,8 @@ static START: sync::Once = sync::Once::new();
 thread_local! {
     /// This is a internal counter in pyo3 monitoring whether this thread has the GIL.
     ///
-    /// It will be incremented whenever a GIL-holding RAII struct is created, and decremented
-    /// whenever they are dropped.
+    /// It will be incremented whenever a GILPool is created, and decremented whenever they are
+    /// dropped.
     ///
     /// As a result, if this thread has the GIL, GIL_COUNT is greater than zero.
     ///
@@ -115,22 +115,35 @@ pub fn prepare_freethreaded_python() {
 #[must_use]
 pub struct GILGuard {
     gstate: ffi::PyGILState_STATE,
-    pool: ManuallyDrop<GILPool>,
+    pool: ManuallyDrop<Option<GILPool>>,
 }
 
 impl GILGuard {
-    /// Acquires the global interpreter lock, which allows access to the Python runtime.
+    /// Acquires the global interpreter lock, which allows access to the Python runtime. This is
+    /// safe to call multiple times without causing a deadlock.
     ///
     /// If the Python runtime is not already initialized, this function will initialize it.
     /// See [prepare_freethreaded_python()](fn.prepare_freethreaded_python.html) for details.
+    ///
+    /// If PyO3 does not yet have a `GILPool` for tracking owned PyObject references, then this
+    /// new `GILGuard` will also contain a `GILPool`.
     pub fn acquire() -> GILGuard {
         prepare_freethreaded_python();
 
         unsafe {
             let gstate = ffi::PyGILState_Ensure(); // acquire GIL
+
+            // If there's already a GILPool, we should not create another or this could lead to
+            // incorrect dangling references in safe code (see #864).
+            let pool = if !gil_is_acquired() {
+                Some(GILPool::new())
+            } else {
+                None
+            };
+
             GILGuard {
                 gstate,
-                pool: ManuallyDrop::new(GILPool::new()),
+                pool: ManuallyDrop::new(pool),
             }
         }
     }
@@ -208,7 +221,7 @@ unsafe impl Sync for ReleasePool {}
 
 static POOL: ReleasePool = ReleasePool::new();
 
-#[doc(hidden)]
+/// A RAII pool which PyO3 uses to store owned Python references.
 pub struct GILPool {
     owned_objects_start: usize,
     owned_anys_start: usize,
@@ -217,8 +230,13 @@ pub struct GILPool {
 }
 
 impl GILPool {
+    /// Create a new `GILPool`. This function should only ever be called with the GIL.
+    ///
+    /// It is recommended not to use this API directly, but instead to use `Python::new_pool`, as
+    /// that guarantees the GIL is held.
+    ///
     /// # Safety
-    /// This function requires that GIL is already acquired.
+    /// As well as requiring the GIL, see the notes on `Python::new_pool`.
     #[inline]
     pub unsafe fn new() -> GILPool {
         increment_gil_count();
@@ -230,8 +248,10 @@ impl GILPool {
             no_send: Unsendable::default(),
         }
     }
-    pub unsafe fn python(&self) -> Python {
-        Python::assume_gil_acquired()
+
+    /// Get the Python token associated with this `GILPool`.
+    pub fn python(&self) -> Python {
+        unsafe { Python::assume_gil_acquired() }
     }
 }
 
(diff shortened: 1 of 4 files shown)


kngwyu:
I have a mixed feeling about this PR.

I can understand we can avoid what you call 'dangling' by this trick.
However, this delays the release of owned objects and thus users cannot control the timing of the release.

I feel this demerit surpasses the merit it offers.
I don't think this situation occurs in rela use cases.
```rust
  let py = gil.python();
  let obj;

  {
    obj = py.eval("object()", None, None).unwrap();
    println!("Count before gil2 drop {}", unsafe { ffi::Py_REFCNT(obj.as_ptr()) });
  }

```

davidhewitt:
> However, this delays the release of owned objects and thus users cannot control the timing of the release.

This is not true - users can still control timing of the release, just they must now use the unsafe `GILPool::new` to control timing of the release.

I think it is very important we make the manual memory control an unsafe API. Because if we continue to tell people to use `acquire_gil` to 
manually manage memory, some users somewhere out there will make this mistake and create segfaults / security holes / UB.

davidhewitt:
Idea: we could perhaps make a safe API similar to `allow_threads` to control memory like this:

```rust
// The below sample would fail to pass the borrow checker because returning a reference to py

let obj: &PyAny = pyo3::with_memory_pool(|py| {
  // py comes from the new GILPool we create, so no objects can accidentally be made
  // with too long lifetime
  py.eval("object()", None, None).unwrap()
});

// The below sample would compile successfully

let obj: PyObject = pyo3::with_memory_pool(|py| {
  py.eval("object()", None, None).unwrap().to_object(py)
});
```

Like `allow_threads` this would use the `Sync` bound to ensure that the user cannot access the wrong `Python` GIL token.

Not sure what I think of the name `with_memory_pool`. Maybe `cleanup_pointers` ? Or another better name?

kngwyu:
> Idea: we could perhaps make a safe API similar to allow_threads to control memory like this:

Hmm :thinking: 
I don't want to our API such complex.

As an experiment, I created #894 witch uses `Vec<Vec<NonNull<ffi::PyObject>>>` for owned storage, but it's 1.5x slower in `list_get_item` or so.

So..., since (now) we cannot find a better solution than PR, now I feel it OK to delay the object relase.
I think it cannot be problematic in most cases.

davidhewitt:
If you like, as part of this PR I can later today add more documentation to address #311

kngwyu:
> I can later today add more documentation to address

How about adding
```rust
impl Python {
    unsafe fn acruiqe_gil_with_memorypool(self) -> GILGuard { 
        GILGuard::acquie_with_gil_pool()  // name your own
    }
}
```
?
Then we can place the document on it.

src/gil.rs
```diff
@@ -220,6 +232,11 @@ pub struct GILPool {
 impl GILPool {
     /// # Safety
     /// This function requires that GIL is already acquired.
+    ///
+    /// Additionally, when this GILPool is dropped, all PyO3 owned references created since this
```
kngwyu, line 236: > all PyO3 owned references created since this

What is problematic is using older `Python` token than this pool and it should be clearly written.

kngwyu:
One thing I dislike about this PR is the lifetime `py` is incorrect if we have nested GILGuards.
So here's another idea: just panics if we detect invalid `Python` is used. #902

But it results in 1.2x slowdown... :cold_sweat:

davidhewitt:
Pushed some more docs, and I want to push some more docs to `advanced.md` for this PR later today.

src/gil.rs
```diff
@@ -231,9 +271,39 @@ impl GILPool {
...
+    /// // This repr would have lifetime tied to `obj` but cannot be safely used after GILPool
+    /// let repr = pool2.reduce_lifetime(obj.repr().unwrap());
+    ///
+    /// drop(pool2);
+    ///
+    /// // Usage of repr is disallowed because pool2 is dead.
+    /// println!("{:?}", repr);
+    /// ```
+    #[inline]
+    pub fn reduce_lifetime<'py, 'a: 'py, T>(&'py self, reference: &'a T) -> &'py T {
```
kngwyu, line 304: I can understand what this function does, but I don't understand why we expose this function as a public API.
We need `GILPool` when we can't get `Python` without it(and he/she is certain that the caller has GIL).
So this kind of situation does not seem realistic to me :thinking:

davidhewitt: 👍 I will write some docs in `advanced.md` that explain the situations I think there are. Then we can review what API we like for each situation.

davidhewitt:
I am unsure if #913 will show a better way to solve this problem? Not sure yet, I haven't thought too hard, feel like there might be a way but haven't seen one yet.

davidhewitt:
Okay, I've had another go at this. I also added some documentation to `advanced.md`.

Based on the feedback you gave, I moved the function to `Python::new_pool()`. This is still `unsafe` because of the reasons we discuss in this PR regarding dangling references, but at least it takes `Python` so guarantees that the GIL is held.

This function still returns `GILPool` though. I prefer this to the proposal `Python::acquire_gil_with_memorypool` because I think if the user needs to acquire GIL they should use the existing `acquire_gil`.

I'm happy to keep tweaking this PR until we find something which feels good though. So let me know what you think of this version 😄

src/python.rs
```diff
@@ -292,6 +292,60 @@ impl<'p> Python<'p> {
...
+    ///   outlive the `GILPool`.
+    /// - Similarly, methods on existing owned references will implicitly refer back to the
+    ///   `Python` token which that reference was originally created with. If the returned values
+    ///   from these methods are owned references they will inherit the same lifetime. As a result,
+    ///   Rust's lifetime rules may allow them to outlive the `GILPool`, even though this is not
+    ///   safe for reasons discussed above. Care must be taken to never access these return values
+    ///   after the `GILPool` is dropped, unless they are converted to `Py<T>` *before* the pool
+    ///   is dropped.
+    #[inline]
+    pub unsafe fn new_pool(self) -> GILPool {
```
kngwyu, line 346: :+1:

kngwyu:
I like the new `Python::new_pool`.
The only concern I have is `GILPool::python` should not be marked `unsafe` to encourage users to use it>

davidhewitt:
> The only concern I have is GILPool::python should not be marked unsafe to encourage users to use it?

Yes I wondered about this too. If we add an extra safety note to `new_pool` saying "the pool must be dropped before the GIL is released", then `GILPool::python` can be deemed safe.

And I guess this requirement is true anyway, else when the `GILPool` drops we will be calling `Py_DECREF` without the GIL.

davidhewitt:
I've made `GILPool::python` safe as discussed 👍

kngwyu:
Thank you!

PR #894, opened 2020-05-04
[EXPERIMENTAL] Indexed owned storages

kngwyu:
To make it safe.
Currently, 1.5x slower when we have many owned objects.


kngwyu:
I like this idea, but I also dislike what is slow :disappointed:

PR #897, opened 2020-05-04
Move macros into separate feature.

m-ou-se:
This makes it possible to opt-out of all the (proc) macros, which reduces the amount of dependencies that need to be downloaded and built.

Together with #895, #896 and #899, this reduces the dependency tree from:

```
pyo3 v0.9.2
├── indoc v0.3.5
│   ├── indoc-impl v0.3.5
│   │   ├── proc-macro-hack v0.5.15
│   │   ├── proc-macro2 v1.0.10
│   │   │   └── unicode-xid v0.2.0
│   │   ├── quote v1.0.3
│   │   │   └── proc-macro2 v1.0.10 (*)
│   │   ├── syn v1.0.17
│   │   │   ├── proc-macro2 v1.0.10 (*)
│   │   │   ├── quote v1.0.3 (*)
│   │   │   └── unicode-xid v0.2.0
│   │   └── unindent v0.1.5
│   └── proc-macro-hack v0.5.15
├── inventory v0.1.6
│   ├── ctor v0.1.14
│   │   ├── quote v1.0.3 (*)
│   │   └── syn v1.0.17 (*)
│   ├── ghost v0.1.1
│   │   ├── proc-macro2 v1.0.10 (*)
│   │   ├── quote v1.0.3 (*)
│   │   └── syn v1.0.17 (*)
│   └── inventory-impl v0.1.6
│       ├── proc-macro2 v1.0.10 (*)
│       ├── quote v1.0.3 (*)
│       └── syn v1.0.17 (*)
├── libc v0.2.69
├── num-traits v0.2.11
│   [build-dependencies]
│   └── autocfg v1.0.0
├── parking_lot v0.10.2
│   ├── lock_api v0.3.4
│   │   └── scopeguard v1.1.0
│   └── parking_lot_core v0.7.2
│       ├── cfg-if v0.1.10
│       ├── libc v0.2.69
│       └── smallvec v1.4.0
├── paste v0.1.11
│   ├── paste-impl v0.1.11
│   │   ├── proc-macro-hack v0.5.15
│   │   ├── proc-macro2 v1.0.10 (*)
│   │   ├── quote v1.0.3 (*)
│   │   └── syn v1.0.17 (*)
│   └── proc-macro-hack v0.5.15
├── pyo3cls v0.9.2 (/home/mara/dev-ext/pyo3/pyo3cls)
│   ├── pyo3-derive-backend v0.9.2 (/home/mara/dev-ext/pyo3/pyo3-derive-backend)
│   │   ├── proc-macro2 v1.0.10 (*)
│   │   ├── quote v1.0.3 (*)
│   │   └── syn v1.0.17 (*)
│   ├── quote v1.0.3 (*)
│   └── syn v1.0.17 (*)
└── unindent v0.1.5
[build-dependencies]
├── regex v1.3.7
│   └── regex-syntax v0.6.17
├── serde v1.0.106
│   └── serde_derive v1.0.106
│       ├── proc-macro2 v1.0.10 (*)
│       ├── quote v1.0.3 (*)
│       └── syn v1.0.17 (*)
├── serde_json v1.0.51
│   ├── itoa v0.4.5
│   ├── ryu v1.0.3
│   └── serde v1.0.106 (*)
└── version_check v0.9.1
```
to:
```
pyo3 v0.9.2
└── libc v0.2.69
[build-dependencies]
└── version_check v0.9.1
```
with `--no-default-features`.

### DIFF
--- a/src/class/methods.rs
+++ b/src/class/methods.rs
@@ -138,6 +138,7 @@ impl PySetterDef {
 /// Allows arbitrary pymethod blocks to submit their methods, which are eventually
 /// collected by pyclass.
 #[doc(hidden)]
+#[cfg(feature = "macros")]
 pub trait PyMethodsInventory: inventory::Collect {
     /// Create a new instance
     fn new(methods: &'static [PyMethodDefType]) -> Self;
@@ -149,6 +150,7 @@ pub trait PyMethodsInventory: inventory::Collect {
 /// Implementation detail. Only to be used through the proc macros.
 /// For pyclass derived structs, this trait collects method from all impl blocks using inventory.
 #[doc(hidden)]
+#[cfg(feature = "macros")]
 pub trait PyMethodsImpl {
     /// Normal methods. Mainly defined by `#[pymethod]`.
     type Methods: PyMethodsInventory;
@@ -159,5 +161,13 @@ pub trait PyMethodsImpl {
             .into_iter()
             .flat_map(PyMethodsInventory::get_methods)
             .collect()
+    }
+}
+
+#[doc(hidden)]
+#[cfg(not(feature = "macros"))]
+pub trait PyMethodsImpl {
+    fn py_methods() -> Vec<&'static PyMethodDefType> {
+        Vec::new()
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,21 +150,18 @@ pub use crate::type_object::{type_flags, PyTypeInfo};
 // Since PyAny is as important as PyObject, we expose it to the top level.
 pub use crate::types::PyAny;
 
-// Re-exported for wrap_function
+#[cfg(feature = "macros")]
 #[doc(hidden)]
-pub use paste;
-// Re-exported for py_run
-#[doc(hidden)]
-pub use indoc;
-// Re-exported for pymethods
-#[doc(hidden)]
-pub use inventory;
+pub use {
+    indoc,     // Re-exported for py_run
+    inventory, // Re-exported for pymethods
+    paste,     // Re-exported for wrap_function
+    unindent,  // Re-exported for py_run
+};
+
 // Re-exported for the `__wrap` functions
 #[doc(hidden)]
 pub use libc;
-// Re-exported for py_run
-#[doc(hidden)]
-pub use unindent;
 
 pub mod buffer;
 #[doc(hidden)]
@@ -197,6 +194,7 @@ pub mod type_object;
 pub mod types;
 
 /// The proc macros, which are also part of the prelude.
+#[cfg(feature = "macros")]
 pub mod proc_macro {
     pub use pyo3cls::pymodule;
     /// The proc macro attributes
@@ -278,6 +276,7 @@ macro_rules! wrap_pymodule {
 /// If you need to handle failures, please use [Python::run] directly.
 ///
 #[macro_export]
+#[cfg(feature = "macros")]
 macro_rules! py_run {
     ($py:expr, $($val:ident)+, $code:literal) => {{
         pyo3::py_run_impl!($py, $($val)+, pyo3::indoc::indoc!($code))
@@ -289,6 +288,7 @@ macro_rules! py_run {
 
 #[macro_export]
 #[doc(hidden)]
+#[cfg(feature = "macros")]
 macro_rules! py_run_impl {
     ($py:expr, $($val:ident)+, $code:expr) => {{
         use pyo3::types::IntoPyDict;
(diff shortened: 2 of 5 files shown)


m-ou-se:
Not sure what's up with the failed checks. They all passed on the same commit here: https://github.com/fusion-engineering-forks/pyo3/actions/runs/95365899

davidhewitt:
Hmmm this is an interesting idea! In principle I am keen to give users an option to have a smaller compile for sure.

I wonder - once the macros are removed then a lot of other code becomes irrelevant I think e.g. `PyCell`, and the entirety of `src/class`. Rather than just drawing the line at macros, should we feature-gate just a little more code? This would get us a long way towards satisfying the recent suggestion in #5 to add a feature to subset pyo3 just to stable code.

m-ou-se:
Yeah, with `feature = "macros"` a lot of the code is unused (or should be), but I didn't manage to find a nice line to cut things off at. I tried putting `PyClass` behind the `feature = "macros"` gate, but that broke a lot of other things. Also gating the things it broke, broke even more things, and so on. So in the end I gave up and just took the easy way out with a dummy implementation for `PyMethodsImpl` in the `not(feature = "macros")` case, and leaving all the `PyClass` stuff in place.

My main goal was to avoid pulling in unnecessary dependencies, so I'm not too worried about parts of the pyo3 crate that stay unused.

It would definitely be nice to split this better (maybe even separate crates?), but that's probably a lot of work.

davidhewitt:
👍 maybe I will have a quick play with this later to see if I can find something which feels good.

I'm in split opinion about separate crates. It helps with some big projects but it does separate the documentation out into multiple places, which can make it a little bit harder for users to find the docs they need...

ijl:
@m-ou-se how much of pyo3 outside of the FFI bindings does your use case exercise if it doesn't use pymodule, pyfunction, etc.? Would moving the FFI bindings into a pyo3ffi crate be useful?

src/class/methods.rs
```diff
@@ -142,3 +144,11 @@ pub trait PyMethodsImpl {
             .collect()
     }
 }
+
+#[doc(hidden)]
+#[cfg(not(feature = "macros"))]
```
kngwyu, line 149: Nice :+1:

kngwyu:
> This would get us a long way towards satisfying the recent suggestion in #5 to add a feature to subset pyo3 just to stable code.

You mean we should separate `PyClass`? If so, please wait untill my next trial to remove specialization fails.

m-ou-se:
> @m-ou-se how much of pyo3 outside of the FFI bindings does your use case exercise if it doesn't use pymodule, pyfunction, etc.? Would moving the FFI bindings into a pyo3ffi crate be useful?

The [`inline-python`](https://docs.rs/inline-python) crate requires `PyDict` and `ToPyObject` and some other stuff, but none of the macros.

Another crate I'm about to publish converts objects to and frorm PyO3's `PyObject` using `serde`. It uses `PyDict`, `PyList`, `PyTuple` and more, but again no macros.

All the macros like `#[pymodule]` are mostly used when using Rust from Python. But when using Python from Rust, they're often not needed (in simple cases).

src/lib.rs
```diff
@@ -280,6 +278,7 @@ macro_rules! wrap_pymodule {
 /// If you need to handle failures, please use [Python::run] directly.
 ///
 #[macro_export]
+#[cfg(feature = "macros")]
```
kngwyu, line 281: I'm sorry for my *lazy* review but could you please add this definition?
```rust
#[macro_export]
#[cfg(not(feature = "macros"))]
macro_rules! py_run {
    ($py:expr, $($val:ident)+, $code:expr) => {{
        pyo3::py_run_impl!($py, $($val)+, &pyo3::unindent::unindent($code))
    }};
}
```

`unindent` is a really small dependency (see [its Cargo.toml](https://github.com/dtolnay/indoc/blob/master/unindent/Cargo.toml)) and I think it's OK.

birkenfeld: then `py_run` would behave differently depending on the feature? that seems confusing...

m-ou-se: That would mean that code like

```rust
let code = "asdf";
py_run!(py, a b, code);
```

Would compile fine with `feature = "macros"` disabled, but will no longer compile when it gets enabled.

kngwyu: > then py_run would behave differently depending on the feature?

No, it behaves the same.
But
1. with `feature="macros"` and
2. string literal is passed like `py_run!(py, a, "print(a)")`
it works faster thanks to [indoc](https://github.com/dtolnay/indoc).

ijl:
I don't have a need to separate the FFI right now. It is simple to do, though. The `build.rs` in each should be duplicated, but this is rarely touched and I think pretty mild for how difficult some build processes can be, so not that bad. So if it helps #904 or something else it's simple to do.

kngwyu:
Github doesn't allow maintainers to edit an organization branch,

kngwyu:
Thanks!

PR #899, opened 2020-05-04
Simplify ReleasePool, remove parking_lot dependency.

m-ou-se:
Replaces the parking_lot Mutex by a simple AtomicBool spinlock, and simplifies the ReleasePool to contain one Vec instead of pointers to two Vecs.

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -3,8 +3,8 @@
 //! Interaction with python's global interpreter lock
 
 use crate::{ffi, internal_tricks::Unsendable, Python};
-use parking_lot::Mutex;
 use std::cell::{Cell, RefCell, UnsafeCell};
+use std::sync::atomic::{spin_loop_hint, AtomicBool, Ordering};
 use std::{any, mem::ManuallyDrop, ptr::NonNull, sync};
 
 static START: sync::Once = sync::Once::new();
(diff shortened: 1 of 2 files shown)


m-ou-se:
@davidhewitt Since you wrote this part, can you review this?

I'm curious if there's any specific reason that I'm missing for why
- the `UnsafeCell`s contained an `*mut Vec` instead of `Vec`,
- the `pointers_being_dropped` was contained within the `ReleasePool`,
- `.set_len(0)` was used instead of `.clear()`.

This PR changes those things, but let me know if i'm missing something here. (In that case we should probably add a comment about it.)

src/gil.rs
```diff
@@ -153,54 +153,56 @@ impl Drop for GILGuard {
...
-        let mut storage = self.pointers_to_drop.lock();
-        if storage.is_null() {
-            *storage = Box::into_raw(Box::new(Vec::with_capacity(256)))
-        }
+        let _lock = Lock::new(&self.locked);
+        let v = self.pointers_to_drop.get();
         unsafe {
-            (**storage).push(obj);
+            if (*v).capacity() == 0 {
+                (*v).reserve(256);
```
kngwyu, line 192: Recently I noticed this kind of reserving doesn't offer so much speed up.
Could you please remove this if sentences?

m-ou-se: Sure. Done.

kngwyu:
Thanks for the update!

programmerjake:
This pull request should have been changed to avoid spinlocks, see #916

PR #901, opened 2020-05-04
Simplify callback code using callback_body! macro

davidhewitt:
While finishing off #797 I had an idea how to remove the `run_callback` function and instead use a macro. This had three benefits:

1. Now `run_callback` no longer appears in the stack trace when panicking
2. The new macro is also able to set up the pool with `GILPool::new()`, `pool.python()`
3. The new macro is able to automatically convert the output with `callback::convert()`

As a result, all callback code now uses this macro instead of `run_callback`, making all callback code shorter and easier to read 😄


kngwyu:
Thank you!

PR #902, opened 2020-05-05
[EXPERIMENTAL] Panics to prevent objects from outliving

kngwyu:
Make `Python` have the gil count when GILPool/GILGuard is created.
And if we detect an older `Python` is used, we cause panic.

1.2x slower than the master :thinking:

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -46,7 +46,7 @@ pub use gil::prepare_freethreaded_python;
 /// To avoid deadlocking, you should release the GIL before trying to lock a mutex, e.g. with
 /// [Python::allow_threads].
 #[derive(Copy, Clone)]
-pub struct Python<'p>(PhantomData<&'p GILGuard>);
+pub struct Python<'p>(u32, PhantomData<&'p GILGuard>);
 
 impl<'p> Python<'p> {
     /// Retrieves a Python instance under the assumption that the GIL is already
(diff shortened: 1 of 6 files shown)


src/python.rs
```diff
@@ -61,7 +61,13 @@ impl<'p> Python<'p> {
     /// I.e., `Python<'static>` is always *really* unsafe.
     #[inline]
     pub unsafe fn assume_gil_acquired() -> Python<'p> {
-        Python(PhantomData)
+        let count = gil::get_gil_count();
+        debug_assert!(count > 0);
+        Python(count, PhantomData)
```
davidhewitt, line 66: I think this isn't quite right, because you could have this:

```rust
let py = gil.python();

let list = PyList::new(py, &[1, 2, 3]);

// Because of assume_gil_acquired(), it is put into gil2's pool.
// But the Rust compiler thinks `item` has the same lifetime as `list`, i.e. `gil1` 
let item = list.get_item(0);          

// Rust thinks this is fine, but item is now dangling.
drop(gil2)
```

I think the only correct way for functions like `list.get_item()` to work is that `item` has to go into the same `GILPool` as `list`, because that's the lifetime that rust sees.

To do this, I think all py native types would have to store `Python`. Maybe it's not so bad, but it doesn't feel great...

kngwyu: Ah, you're right. Thank you

kngwyu:
> Is the change in #893 really so bad?

As I [said](https://github.com/PyO3/pyo3/pull/893#issuecomment-623350214), I don't mean it's so bad. This PR is just for an experiment.

Issue #662, opened 2019-11-12
Support creating class member variables (not instance members)


programmerjake:
This can apparently be done by setting `PyTypeObject.tp_dict` to a dict with the appropriate values before calling `PyType_Ready`.

See https://stackoverflow.com/questions/35954016/defining-an-inner-class-using-the-python-c-api for a use case. Another use case would be creating enum constants (assuming initialization order can be worked out).

I think a useful way to do that would be to have an overridable `static_init` function that returns the `PyDict` used to initialize `tp_dict`, since that also allows non-const code to create the member variables.

davidhewitt:
@programmerjake with #905 and #914 is there any more work to be done here? I think we established that the class attributes (as we decided to call them) can't be mutable, so I believe we've covered everything...

programmerjake:
I think that should be all that's needed, unless we want to add a function (`static_init` from first comment) that can add members algorithmically, though that would have to be unsafe since it could save a copy of the dict somewhere and try to modify it after initialization finished.

davidhewitt:
Let's call this done, as nobody else has asked for more on this front since we added this over a year ago. We can open a new issue if there are real-world use cases for the `static_init` case.

PR #905, opened 2020-05-05
Add `#[classattr]` methods to define Python class attributes

scalexm:
Here is my approach to #662: we add an attribute on methods defined in `#[pymethods]` impl blocks to allow defining class attributes (also named class variables).

A class attribute method cannot take arguments, and must return `impl IntoPy<PyObject>`.

For example, given the following Python code:
```python
class Foo:
    foo = 5
    BAR = 'bar'
```

a translation in Rust with `pyo3` could be given by:
```rust
#[pyclass]
struct Foo { }

#[pymethods]
impl Foo {
    #[classattr]
    fn foo() -> i32 {
        5
    }

    #[classattr]
    #[name = "BAR"]
    fn bar() -> String {
        "bar".to_string()
    }
}
```

### DIFF
--- a/pyo3-derive-backend/src/method.rs
+++ b/pyo3-derive-backend/src/method.rs
@@ -29,6 +29,7 @@ pub enum FnType {
     FnCall,
     FnClass,
     FnStatic,
+    ClassAttribute,
     /// For methods taht have `self_: &PyCell<Self>` instead of self receiver
     PySelfRef(syn::TypeReference),
     /// For methods taht have `self_: PyRef<Self>` or `PyRefMut<Self>` instead of self receiver
@@ -139,6 +140,15 @@ impl<'a> FnSpec<'a> {
             };
         }
 
+        if let FnType::ClassAttribute = &fn_type {
+            if self_.is_some() || !arguments.is_empty() {
+                return Err(syn::Error::new_spanned(
+                    name,
+                    "Class attribute methods cannot take arguments",
+                ));
+            }
+        }
+
         // "Tweak" getter / setter names: strip off set_ and get_ if needed
         if let FnType::Getter | FnType::Setter = &fn_type {
             if python_name.is_none() {
@@ -178,7 +188,7 @@ impl<'a> FnSpec<'a> {
                 "text_signature not allowed on __new__; if you want to add a signature on \
                  __new__, put it on the struct definition instead",
             )?,
-            FnType::FnCall | FnType::Getter | FnType::Setter => {
+            FnType::FnCall | FnType::Getter | FnType::Setter | FnType::ClassAttribute => {
                 parse_erroneous_text_signature("text_signature not allowed with this attribute")?
             }
         };
@@ -331,6 +341,8 @@ fn parse_method_attributes(
                     res = Some(FnType::FnClass)
                 } else if name.is_ident("staticmethod") {
                     res = Some(FnType::FnStatic)
+                } else if name.is_ident("classattr") {
+                    res = Some(FnType::ClassAttribute)
                 } else if name.is_ident("setter") || name.is_ident("getter") {
                     if let syn::AttrStyle::Inner(_) = attr.style {
                         return Err(syn::Error::new_spanned(
--- a/pyo3-derive-backend/src/pymethod.rs
+++ b/pyo3-derive-backend/src/pymethod.rs
@@ -30,6 +30,9 @@ pub fn gen_py_method(
         FnType::FnCall => impl_py_method_def_call(&spec, &impl_wrap(cls, &spec, false)),
         FnType::FnClass => impl_py_method_def_class(&spec, &impl_wrap_class(cls, &spec)),
         FnType::FnStatic => impl_py_method_def_static(&spec, &impl_wrap_static(cls, &spec)),
+        FnType::ClassAttribute => {
+            impl_py_class_attribute(&spec, &impl_wrap_class_attribute(cls, &spec))
+        }
         FnType::Getter => impl_py_getter_def(
             &spec.python_name,
             &spec.doc,
@@ -242,6 +245,19 @@ pub fn impl_wrap_static(cls: &syn::Type, spec: &FnSpec<'_>) -> TokenStream {
 
                 pyo3::callback::convert(_py, #body)
             })
+        }
+    }
+}
+
+/// Generate a wrapper for initialization of a class attribute.
+/// To be called in `pyo3::pyclass::initialize_type_object`.
+pub fn impl_wrap_class_attribute(cls: &syn::Type, spec: &FnSpec<'_>) -> TokenStream {
+    let name = &spec.name;
+    let cb = quote! { #cls::#name() };
+
+    quote! {
+        fn __wrap(py: pyo3::Python<'_>) -> pyo3::PyObject {
+            pyo3::IntoPy::into_py(#cb, py)
         }
     }
 }
@@ -610,6 +626,20 @@ pub fn impl_py_method_def_static(spec: &FnSpec, wrapper: &TokenStream) -> TokenS
                 ml_meth: pyo3::class::PyMethodType::PyCFunctionWithKeywords(__wrap),
                 ml_flags: pyo3::ffi::METH_VARARGS | pyo3::ffi::METH_KEYWORDS | pyo3::ffi::METH_STATIC,
                 ml_doc: #doc,
+            }
+        })
+    }
+}
+
+pub fn impl_py_class_attribute(spec: &FnSpec<'_>, wrapper: &TokenStream) -> TokenStream {
+    let python_name = &spec.python_name;
+    quote! {
+        pyo3::class::PyMethodDefType::ClassAttribute({
+            #wrapper
+
+            pyo3::class::PyClassAttributeDef {
+                name: stringify!(#python_name),
+                meth: __wrap,
             }
         })
     }
--- a/src/class/methods.rs
+++ b/src/class/methods.rs
@@ -1,8 +1,9 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
-use crate::ffi;
+use crate::{ffi, PyObject, Python};
 use libc::c_int;
 use std::ffi::CString;
+use std::fmt;
 
 /// `PyMethodDefType` represents different types of Python callable objects.
 /// It is used by the `#[pymethods]` and `#[pyproto]` annotations.
@@ -18,6 +19,8 @@ pub enum PyMethodDefType {
     Static(PyMethodDef),
     /// Represents normal method
     Method(PyMethodDef),
+    /// Represents class attribute, used by `#[attribute]`
+    ClassAttribute(PyClassAttributeDef),
     /// Represents getter descriptor, used by `#[getter]`
     Getter(PyGetterDef),
     /// Represents setter descriptor, used by `#[setter]`
@@ -38,6 +41,12 @@ pub struct PyMethodDef {
     pub ml_meth: PyMethodType,
     pub ml_flags: c_int,
     pub ml_doc: &'static str,
+}
+
+#[derive(Copy, Clone)]
+pub struct PyClassAttributeDef {
+    pub name: &'static str,
+    pub meth: for<'p> fn(Python<'p>) -> PyObject,
 }
 
 #[derive(Copy, Clone, Debug)]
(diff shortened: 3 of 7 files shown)


davidhewitt:
Thanks for this PR, a way to add class attributes like this would be great. Couple questions (disclaimer, I haven't read the code yet):

> Bike-shedding on the name of the attribute (currently #[attribute]) is welcome. Also, I'm not used to the Python C API so it's possible that my code has memory errors.

Can these things ever be settable? Maybe a pair of names like `classgetter` / `classsetter`? Though maybe these functions run only once during type object creation?

> Note that in the latter case, returning an Err value will crash the Python interpreter

Did you explore mechanisms to prevent returning `PyResult` from these functions? It would be nice to make it impossible to compile code that's guaranteed to cause runtime crash, if that's viable.

Also, I wonder whether we could use syntax like associated constants instead of functions. But I think they're only stable for traits, not structs?

programmerjake:
It would be nice if we could also use associated constants (which are stable, `i8::MAX` (not to be confused with `std::i8::MAX`) is a good example):

```rust
#[pymethods]
impl MyClass {
    #[attribute]
    pub const MY_CONST: &str = "blah";

    #[attribute]
    pub fn my_const2(_py: Python) -> Vec<u8> {
        vec![1, 2, 3]
    }
}

**Comment by kngwyu** [comment]
Interesting, thank you @scalexm!

>  It would be nice if we could also use associated constants 

How about using `const fn` instead of constants?

File: pyo3-derive-backend/src/pymethod.rs [hunk 1]
```diff
@@ -246,6 +249,21 @@ pub fn impl_wrap_static(cls: &syn::Type, spec: &FnSpec<'_>) -> TokenStream {
...
 
+/// Generate a wrapper for initialization of a class attribute.
+/// To be called in `pyo3::pyclass::initialize_type_object`.
+pub fn impl_wrap_class_attribute(cls: &syn::Type, spec: &FnSpec<'_>) -> TokenStream {
+    let name = &spec.name;
+    let cb = quote! { #cls::#name() };
+    let into_result = quote! { pyo3::derive_utils::IntoPyResult::into_py_result };
+
+    quote! {
+        fn __wrap(py: pyo3::Python<'_>) -> pyo3::PyResult<*mut pyo3::ffi::PyObject> {
```
kngwyu, line 260: Since this `__wrap` is not an FFI function(=it is called only in PyO3, not CPython itself), how about using pyo3's objects?
I.e., 
```rust
fn __wrap(py: pyo3::Python<'_>) -> pyo3::PyResult<PyObject> { ... }
```
or so.
Note that you need to use `into_ptr` instead of `as_ptr` when you set the `PyObject` to the tp_dict.

src/pyclass.rs
```diff
@@ -165,13 +165,31 @@ where
...
-    let (new, call, mut methods) = py_class_method_defs::<T>();
     if !methods.is_empty() {
         methods.push(ffi::PyMethodDef_INIT);
         type_object.tp_methods = Box::into_raw(methods.into_boxed_slice()) as *mut _;
     }
 
+    // class attributes
+    if !attrs.is_empty() {
+        unsafe {
+            type_object.tp_dict = ffi::PyDict_New();
```
kngwyu, line 179: Same as above.
You can use `let dict = pyo3::types::PyDict::new();` and `type_object.tp_dict = dict.to_object(py).into_ptr()`.
The current code is also OK, though.

tests/test_class_attributes.rs
```diff
@@ -0,0 +1,29 @@
+use pyo3::prelude::*;
+
+mod common;
+
+#[pyclass]
+struct Foo { }
```
kngwyu, line 6: rustfmt complains about it.
```suggestion
struct Foo {}
```

src/class/methods.rs
```diff
@@ -85,6 +94,22 @@ impl PyMethodDef {
     }
 }
 
+impl PyClassAttributeDef {
+    pub fn name_cstring(&self) -> CString {
```
kngwyu, line 98: Does this need to be `pub`?

pganssle:
Haven't dug into the implementation, but it's not clear - can you modify these at runtime? Either way I would add a test for the expected behavior if you try to assign a value to it.

Also, it's not immediately clear: Is the function evaluated when the class is constructed, or every time it is accessed?

kngwyu:
> Haven't dug into the implementation, but it's not clear - can you modify these at runtime? Either way I would add a test for the expected behavior if you try to assign a value to it.
Also, it's not immediately clear: Is the function evaluated when the class is constructed, or every time it is accessed?

In the current implementation, we just set the variable to `type(PyClass).__dict__` once when the type object is created.

1. About returning an error
Is there any use case where we need to return an error from `attribute`? If not, I think we should restrict the return type to `IntoPy<PyObject>`.

2. About naming
`attribute` is not so bad, but aren't they called [class variable](https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables) in Python?
How about `#[classvar]`?

3. > Can these things ever be settable? 
They are settable in Python.
But since in Rust, we need to use `Cell/RefCell` to share mutable variables and it's not a common pattern. So I don't think we have any strong reason to support `setter` for it.

4. > It would be nice if we could also use associated constants 
Agreed.

scalexm:
Thanks for the reviews.

> How about `#[classvar]`
Yes, that’s probably better. Always called that class attributes but indeed the official Python documentation calls them class variables :)

> Is there any use case where we need to return an error from `attribute`
Python does support it, in which case an exception is raised when the module is imported. It’s indeed not clear how useful it is, so I guess we can disable it for now (only accept `IntoPy<PyObject>`) and think about it later if someone comes up with a use case.

About using associated constants, I can open a follow-up PR.

davidhewitt:
> It would be nice if we could also use associated constants (which are stable, i8::MAX (not to be confused with std::i8::MAX) is a good example):

Nice, I was so sure they were stable but then when looking at documentation to confirm this last night I got bogged down in the docs for traits xD

> They are settable in Python. 

If they're settable in Python, but we don't want to allow setting these in Rust, I suggest we make it so that an exception is raised if a user attempts to set these attributes from Python. Otherwise the original Rust value and current Python value may become out of sync?

pganssle:
For comparison, here's what happens with a Cython `cdef` class defined like so:

```python
cdef class MyClass:
    attr = 3
```

```
>>> from mymodule._backend import MyClass
>>> MyClass.attr
3
>>> MyClass.attr = 4
TypeError: can't set attributes of built-in/extension type 'mymodule._backend.MyClass'
>>> MyClass().attr
3
>>> MyClass().attr = 4
AttributeError: 'mymodule._backend.MyClass' object attribute 'attr' is read-only
```

So it seems like they are called "attributes" and that it is reasonable for them to be read-only in both Python and Rust.

I think `#[attribute]` or `#[classattr]` or `#[classattribute]` are all reasonable spellings. I'm not terribly fond of the idea of `#[classvar]`. Class scoping in Python is super weird and there may be a distinction between "class variable" and "attribute" that I don't understand.

tests/test_class_attributes.rs
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    #[attribute]
+    #[name = "BAR"]
+    fn bar() -> String {
+        "bar".to_string()
+    }
+}
+
+#[test]
```
pganssle, line 22: Mentioned this in my comment, but adding an in-line note so it doesn't get lost: I think it would be a good idea to add tests for the error cases, to ensure that they raise the correct exception.

Could be done here or it could be done in the `examples/rustapi_module` tests, which are written in Python and where you can use `pytest.raises`.

kngwyu:
> I think #[attribute] or #[classattr] or #[classattribute] are all reasonable spellings. I'm not terribly fond of the idea of #[classvar]. Class scoping in Python is super weird and there may be a distinction between "class variable" and "attribute" that I don't understand.

Thank you for the feedback. 

> I suggest we make it so that an exception is raised if a user attempts to set these attributes from Python.

I checked the code that Cython produced and what it does is the same as this PR. They set `attr` to `tp_dict`.
```c
  if (PyDict_SetItem((PyObject *)__pyx_ptype_4demo_MyClass->tp_dict, __pyx_n_s_attr, __pyx_int_3) < 0) __PYX_ERR(1, 2, __pyx_L1_error)
```
But I don't understand how it makes the `attr` read-only :thinking:

programmerjake:
> But I don't understand how it makes the `attr` read-only  ...

Because Python enforces that, for classes defined using the C API that PyO3 uses, type objects can't be changed (I'm assuming since they're shared across all interpreters in the program) and `tp_dict` is considered part of that type object.

davidhewitt:
+1 for `#[classattr]` as my favourite of the names suggested.

> Because Python enforces that, for classes defined using the C API that PyO3 uses, type objects can't be changed (I'm assuming since they're shared across all interpreters in the program) and tp_dict is considered part of that type object.

Nice. I guess should make sure there are tests in this PR to verify that trying to modify a classattr raises an exception as described.

scalexm:
Current status:
* applied @kngwyu comments: it's cleaner that way, thanks
* removed the possibility to return a `PyResult`: this can always be added later if someone has a use case for it
* added a test to showcase that class variables from an extension module are indeed immutable

Next steps:
* decide on the name: I personally don't have a strong opinion
* add a paragraph to the guide
* allow associated constants to define class attributes: I'd prefer to open a separate follow-up PR for that one

tests/test_class_attributes.rs
```diff
@@ -0,0 +1,37 @@
...
+    py_assert!(py, typeobj, "typeobj.foo == 5");
+    py_assert!(py, typeobj, "typeobj.BAR == 'bar'");
+}
+
+#[test]
+fn class_attributes_are_immutable() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+    let typeobj = py.get_type::<Foo>();
+    py_expect_exception!(py, typeobj, "typeobj.foo = 6", TypeError);
```
kngwyu, line 36: Nice :+1:

kngwyu:
Thank you for the update!

> decide on the name: I personally don't have a strong opinion

Me too, so the top candidate is `#[classattr]`?(+1 by @davidhewitt)

> add a paragraph to the guide

Yeah, please place some docs in `class.md`.

> allow associated constants to define class attributes: I'd prefer to open a separate follow-up PR for that one

:+1:

programmerjake:
I'm happy with any of the options other than `#[classvar]` since the attributes act like constants, not variables.

pyo3-derive-backend/src/method.rs
```diff
@@ -139,6 +140,15 @@ impl<'a> FnSpec<'a> {
             };
         }
 
+        if let FnType::ClassAttribute = &fn_type {
+            if self_.is_some() || !arguments.is_empty() {
+                return Err(syn::Error::new_spanned(
+                    name,
+                    "Class attribute methods cannot take arguments",
```
davidhewitt, line 147: It may be helpful to optionally allow one argument, of type `Python`, for those who want to create python types as class attributes. See `pymethod/impl_wrap_getter` as an example of how we allow this for getters.

One caveat I am not sure about though: as we're currently in the middle of creating a type object, is it safe to run arbitrary Python code?

kngwyu: > is it safe to run arbitrary Python code?

Maybe this code can cause SIGSEGV.
```rust
#[pymethods]
impl MyClass {
    #[classattr]
    fn foo() -> MyClass { ... } 
}
```

I'll open a small PR to prevent this.

src/pyclass.rs
```diff
@@ -165,13 +167,23 @@ where
...
+
     // normal methods
-    let (new, call, mut methods) = py_class_method_defs::<T>();
     if !methods.is_empty() {
         methods.push(ffi::PyMethodDef_INIT);
         type_object.tp_methods = Box::into_raw(methods.into_boxed_slice()) as *mut _;
     }
 
+    // class attributes
+    if !attrs.is_empty() {
```
kngwyu, line 179: Considering a recursive case(e.g.,
```rust
#[pymethods]
impl MyClass {
    #[classattr]
    fn foo() -> MyClass { ... } 
}
```
), could you please move this code after `PyType_Ready`?
Then an incomplete type object is never used.

scalexm: That was precisely one of my use cases indeed. But I think it’s still possible to return `MyOtherClass` where `MyOtherClass` is has not yet been initialized?

kngwyu: > But I think it’s still possible to return MyOtherClass where MyOtherClass has not yet been initialized?

Yes, but I'm not sure there's no corner case :thinking:

davidhewitt: Is it safe to modify `type_object` after `PyType_Ready` ?

kngwyu: I realized that it cannot be a problem since `PyCell::new` doesn't use the type object at all.

pyo3-derive-backend/src/method.rs
davidhewitt: @kngwyu is your conclusion that this is safe?

src/pyclass.rs
scalexm: I’ll add a few « recursive » test cases to be sure that we don’t break that assumption in the future.

programmerjake: > Is it safe to modify `type_object` after `PyType_Ready` ?

no, it's not safe, according to the C API docs

kngwyu:
LGTM, thank you for the update!

Issue #903, opened 2020-05-05
Generic lists


skreborn:
## ❓ Support request

_Apologies if this is not the correct forum to post support requests - I couldn't find a more appropriate channel._

### ℹ Setup

I have the following struct defined.

```rust
#[derive(Clone)]
pub struct Array<T: ToPyObject + IntoPy<PyObject>> {
  item_size: Option<u32>,
  items: Vec<T>,
}
```

Now, `#[pyclass]` does not allow for using generics, which is too bad, but I'm sure there's a technical reason behind it. That's quite alright, I suppose - I thought I could simply define another struct that's the Python representation like so.

```rust
#[pyclass(name = Array)]
#[derive(Clone)]
pub struct PyArray {
  #[pyo3(get)]
  item_size: Option<u32>,
  #[pyo3(get)]
  items: ?, 
}
```

Here is my conversion code as of now.

```rust
impl<T: ToPyObject + IntoPy<PyObject>> IntoPy<PyObject> for Array<T> {
  fn into_py(self, py: Python) -> PyObject {
    PyArray {
      item_size: self.item_size,
      items: ?,
    }.to_object(py)
  }
}
```

### ⚠ Problem

What I cannot figure out, however, is how can I actually have `pyo3` convert to and from this representation.

- What should be the type of `PyArray::items`?
- What can I do to convert the generic `Vec<T>` in my `impl` block?

birkenfeld:
I think `items` has to be `Vec<PyObject>`, and you convert by creating a new `Vec` with something like `self.items.map(|v| v.into()).collect()`.

skreborn:
Thank you for getting back to me. Unfortunately, `Vec<PyObject>` does not `impl Clone`, so neither `#[derive(Clone)]`, nor `#[pyclass(name = Array)]` deem the struct acceptable.

`Py<PyList>` is acceptable by `#[pyclass(name = Array)]` as of the latest commit (I believe), but it still isn't `Clone`able. Now, I think I might be able to live without it being so, but I'm still curious if there's a way to work around this issue.

If this is what's currently possible, then all I need is to convert from a `Vec<T>` to `Py<PyList>`, which I might or might not be able to achieve with my limited knowledge on the subject, but it should be the least of my worries.

birkenfeld:
I don't believe `Clone` is required for `#[pyclass]`. At least, I have no problem compiling this:

```
#[pyclass]
struct Test {
    items: Vec<PyObject>
}
```

[...]

Issue #803, opened 2020-03-12
Expose python version for cfg()


ijl:
pyo3 uses `[cfg(Py_3_6)]` etc., which is set in `build.rs` as `println!("cargo:rustc-cfg=Py_3_...`. Could this exposed to projects that use pyo3? There is a note in the file about being able to use flags to read these values, but that's left from cpython-rust some time ago and doesn't seem to work.

pganssle:
IIRC there wasn't an easy way to do this at compile time when I wanted to do it for the test module in Python, so I ended up passing it in the config files [via `setup.py`](https://github.com/PyO3/pyo3/blob/759dbeaed79907b7b9a8a406b5e0f4bcdebc512c/examples/rustapi_module/setup.py#L59-L70).

That was added in #200, but the reasoning isn't obvious in that ticket. I don't know enough about this sort of thing to know if there's a better solution, but when I looked into it, those `cfg` flags are set by the flags passed to `rustc`, and PyO3 generates them for its own invocation in `build.rs`. IIRC I couldn't come up with a good way to expose that information to downstream projects at compile time, but I agree that it would be useful if it's possible.

ijl:
Ok, maybe it'd be appropriate to build into maturin. That invokes rustc (and can set cfg) and already has the information available from its checks on the python environment.

davidhewitt:
... should I move this ticket to maturin?

ijl:
I looked into it and there's no way for another crate to specify this to another crate. It must be done through build.rs in the crate using pyo3. It would be best I think to specify a recipe of ~20 lines rather than try to create an API through env vars as cpython-rust does.

kngwyu:
>  It would be best I think to specify a recipe of ~20 lines rather than try to create an API through env vars as cpython-rust does.

I think it's good to create a separate crate and use it from maturin and pyo3's build.rs

PR #907, opened 2020-05-06
Use ToPyObject instead of GetPropertyValue for #[pyo3(get)]

kngwyu:
Some how it works.
cc: #903


davidhewitt:
I think it might not work for `T: PyClass` ? A test must be missing.

kngwyu:
Ah, you're right. Thank you.

Maybe we need another way to enable `get` without `Clone`

davidhewitt:
See my comment in #803 about `Clone` for `PyObject`... I'm excited to see what you think of it!

PR #908, opened 2020-05-06
Add Clone for PyObject / Py<T>

davidhewitt:
I had this idea suddenly that similar to how we store pending `Py_DECREF`, I believe we can also store pending `Py_INCREF`, as long as all increfs are done before decrefs.

This allows `Py` and `PyObject` to have `Clone` implemented. If the GIL is held the reference count is updated immediately, otherwise the reference count increase is saved for the next time a thread acquires the GIL.

I have been mulling this over in my head and cannot find a way that this produces a dangling `PyObject`. Please do challenge me on this one though - I would prefer to be proved wrong and we don't merge this if it is not safe.

I will add tests later - gotta go to work! Just wanted to share it with you first as I think having `Clone` for these types will open the way for simplifying some things. (e.g. I think we could potentially add blanket impl `ToPyObject` for all `T: Clone`? Haven't thought too hard about that.)

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -166,10 +166,11 @@ impl Drop for GILGuard {
     }
 }
 
-/// Thread-safe storage for objects which were dropped while the GIL was not held.
-struct ReleasePool {
+/// Thread-safe storage for objects which were inc_ref / dec_ref while the GIL was not held.
+struct ReferencePool {
     locked: AtomicBool,
-    pointers_to_drop: UnsafeCell<Vec<NonNull<ffi::PyObject>>>,
+    pointers_to_incref: UnsafeCell<Vec<NonNull<ffi::PyObject>>>,
+    pointers_to_decref: UnsafeCell<Vec<NonNull<ffi::PyObject>>>,
 }
 
 struct Lock<'a> {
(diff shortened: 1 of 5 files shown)


kngwyu:
> I had this idea suddenly that similar to how we store pending Py_DECREF, I believe we can also store pending Py_INCREF, as long as all increfs are done before decrefs.

Wow. Awesome :+1: 

> I have been mulling this over in my head and cannot find a way that this produces a dangling PyObject

The logic seems correct to me, though we need to test it carefully.

skreborn:
> I think we could potentially add blanket impl `ToPyObject` for all `T: Clone`

That sounds excellent! I would love to remove a bunch of those straightforward implementations from my code base.

davidhewitt:
FYI now that we have merged #912 and #913 I should be able to finish writing tests for this tomorrow 👍

davidhewitt:
Pushed some tests - ready for review!

src/gil.rs
```diff
@@ -501,32 +544,23 @@ mod test {
 
     #[test]
     fn test_allow_threads() {
```
davidhewitt, line 546: I had to simplify this test because otherwise it could have race conditions with `test_clone_in_allow_threads`, as both of them are testing the same global state.

davidhewitt: EDIT: I had to remove `test_clone_in_allow_threads` anyway, because it was very brittle checking the global state with potentially many other tests all changing the state. But I still think the simplification I did here makes sense.

src/types/string.rs
```diff
@@ -71,24 +71,17 @@ impl PyString {
...
-                        self.as_ptr(),
-                        CStr::from_bytes_with_nul(b"utf-8\0").unwrap().as_ptr(),
-                        CStr::from_bytes_with_nul(b"surrogatepass\0")
-                            .unwrap()
-                            .as_ptr(),
-                    );
-                    // Since we have a valid PyString and replace any surrogates, assume success.
-                    debug_assert!(!py_bytes.is_null());
-                    // ensure DECREF will be called
-                    gil::register_pointer(NonNull::new(py_bytes).unwrap());
```
davidhewitt, line 85: Usage of `register_pointer` was broken since my change to pointer drop immediately with GIL, suprised this test wasn't failing (probably luck).

I have rewritten this to use usual owned references.

kngwyu: :+1:

kngwyu:
Thank you!

PR #909, opened 2020-05-06
Use -Cpanic=abort for correct coverage

kngwyu:
See the discussion in https://github.com/mozilla/grcov/issues/427 for more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,9 +47,6 @@ install:
   - pip install setuptools-rust pytest pytest-benchmark tox
 
 script:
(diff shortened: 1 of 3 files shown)


davidhewitt:
The test I added for panic (in PyErr) might abort if this flag is set; might need to `cfg` that test out?

kngwyu:
Looks it works correctly: https://codecov.io/github/PyO3/pyo3/commit/0d357bbb55f82345b9c3c904b727637178bfc86f
And now travis test is 4~5 min fast since I moved codecov flags to `cover.sh`.

davidhewitt:
Nice!

Issue #787, opened 2020-03-03
A safe API for upcasting


davidhewitt:
#770 initially implemented `AsRef<Base>` for `PyRef<T>`.

We had to remove that in #783 for native base types due to a safety issue in the implentation.

It would be ideal if `AsRef` could be implemented for `PyRef<T>` for both `T` and all bases of `T` - e.g. `AsRef<T>`, `AsRef<T::Base>`, `AsRef<T::Base::Base>` etc.

Icxolu:
This is solved now by the `Bound|PyRef|PyClassGuard::as|into_super` family of functions.

PR #911, opened 2020-05-07
Remove ObjectProtocol

davidhewitt:
This is a reworked version of #892 which does not include the `builtin_methods` idea.

Summary of changes:
- `ObjectProtocol` methods have been moved to `PyAny`.
- All native types implement `Deref<Target = PyAny>`
- ~~`PyCell<T>` implements `Deref<Target = <T::BaseType as PyTypeInfo>::AsRefType>`. This allows deref to super types, and because `PyAny` is always the last base type, this means `PyCell<T>` can always access the `PyAny` methods too. This actually provides a lot of the solution for #787~~ Changed so that `PyCell<T>` now just implements `Deref<Target = PyAny>`, thanks for feedback.

I wanted to add `AsRef<PyAny>` to `PyRef` and `PyRefMut`, but this is more disruptive change because at the moment `py_ref.as_ref()` always gives the super type. If I add a second `AsRef` impl, then sometimes type inference fails where it did not before.

I think I can improve that by creating `PyRef::as_super`, but still wanted to check what people think before I do that.


kngwyu:
Nice, thank you. I'll give a detailed review later.

> PyCell<T> implements Deref<Target = <T::BaseType as PyTypeInfo>::AsRefType>

Casting `PyCell<Super>` to `PyCell<Base>` is not always safe because of dict and weakref slots.
So I think `Target = PyAny` is sufficient.

> I wanted to add AsRef<PyAny> to PyRef and PyRefMut, but this is more disruptive change because at the moment py_ref.as_ref() always gives the super type. 

Good catch :+1: , but I have another concern about borrow flags.
For example, this code panics because it validates `PyCell`'s borrowing rule.
```rust
#[pyclass] 
struct Class {
    field: usize
}
let ref_: PyRef<Class> = ~;
ref.setattr("field", 5).unwrap();
```
So I think providing safe and good APIs for `PyRef/PyRefMut` conversion is difficult.
And, since it's not so common usage, I don't think we need it.

davidhewitt:
> Casting PyCell<Super> to PyCell<Base> is not always safe because of dict and weakref slots.
So I think Target = PyAny is sufficient.

Oh, can you explain this to me? I thought because `PyCellInner<Super>` has PyCellInner<Base>` as the first field, this upcast is ok.

kngwyu:
> Oh, can you explain this to me? 

When a class with `#[pyclass(dict)]` or `#[pyclass(weakref)]` is extended only the subclass has `dict/weakref` slots. So if we get `PyClass<Base>` from `PyClass<Super>`, it doesn't have dict/weakref slots. But this can't be actually problematic so... please don't mind about this unsafety.
Anyway, I still think that `Deref<Target=PyAny` is better, because we can get no useful methods by `Deref<Target=PyClass<Base>>`.

davidhewitt:
Hmm I think if casting `PyCell<Sub>` to `PyCell<Base>` is not safe, we have bigger problems, as `downcast` already makes this possible:

```rust
let py = gil.python();

let sub_any = PyCell::new(py, (SubClass {}, BaseClass { value: 120 })).unwrap().to_object(py);

let base_cell = sub_any.as_ref(py).downcast::<PyCell<BaseClass>>().unwrap();
println!("base: {:?}", base_cell as *const _);

let sub_cell = sub_any.as_ref(py).downcast::<PyCell<SubClass>>().unwrap();
println!("sub: {:?}", sub_cell as *const _);

// Output is the same address:
// base: 0x1f556db67f0
// sub: 0x1f556db67f0
```

davidhewitt:
> because we can get no useful methods by Deref<Target=PyClass<Base>>.

We get access to `PyClass<Base>::borrow` etc. See [this line in test_pycell_deref](https://github.com/PyO3/pyo3/pull/911/files#diff-99e048cc288af4555563a8fadb228e82R166) where I can borrow `PyRefMut<BaseClass>` from a `PyCell<SubClass>`.

You sure you don't want this? If you're sure, I am happy to change it to just `Deref<Target = PyAny>` 👍

kngwyu:
> We get access to PyClass<Base>::borrow etc

Ah I missed that, sorry. So now we have to ways to get `PyRef<Base>`?
```rust
// 1.
let base = cell.borrow().into_super();
// 2.
let base: PyRef<Base> = PyCell::borrow(cell);
```
Hmm... :thinking: 
I don't think that the later one is completely useless, but isn't it difficult to notice for users?

davidhewitt:
> Hmm... 🤔
I don't think that the later one is completely useless, but isn't it difficult to notice for users?

Yeah, I would be glad to have ideas how to make the documentation better here.

In my opinion this new way to borrow `Base` is more powerful, because you can even borrow a grandparent in the same way:

``` rust
let cell: &PyCell<SubSubClass> = ...;

// Current way to get PyRef<Base>
// As one line: cell.borrow().into_super().as_ref()
let sub_sub_ref: PyRef<SubSubClass> = cell.borrow();
let sub_ref: PyRef<SubClass> = sub_sub_ref.into_super();
let base_ref: &PyRef<Base> = sub_ref.as_ref();

// New way to get PyRef<Base>
let base_ref: PyRef<Base> = PyCell::borrow(cell);
```

kngwyu:
> In my opinion this new way to borrow Base is more powerful, because you can even borrow a grandparent in the same way:

Yeah, but I think *simpleness* is more important here than such kind of *powerfulness*, since it's rare to have grandparents.

davidhewitt:
That's fair 👍 . I will change `PyCell<T>` to `Deref<Target = PyAny>`.

davidhewitt:
I've pushed that change 😄

src/types/any.rs
```diff
@@ -59,10 +66,449 @@ pyobject_native_type_convert!(
...
 
 impl PyAny {
+    /// Convert this PyAny to a concrete Python type.
     pub fn downcast<T>(&self) -> Result<&T, PyDowncastError>
     where
         for<'py> T: PyTryFrom<'py>,
     {
         <T as PyTryFrom>::try_from(self)
     }
+
```
kngwyu, line 76: You removed `fn None`?
Though I'm for that since we can get None by `py.None()`, it should be added to the removed section in the CHANGELOG.

guide/src/types.md
```diff
@@ -108,9 +111,12 @@ lifetime just like `PyAny`.
 the GIL.  Like `PyAny`, this is the most convenient form to use for function
 arguments and intermediate values.
 
+These types all implement `Deref<Target = PyAny>`, so they all expose the same
+methods which can be found on `PyAny`.
+
 **Conversions:**
 
-- To `PyAny`: `obj.as_ref()`
+- To `PyAny`: `obj.as_ref()` or `obj.deref()`
```
kngwyu, line 119: Since [clippy dislike explicit deref](https://rust-lang.github.io/rust-clippy/master/index.html#explicit_deref_methods), `&*obj` is better?

src/types/any.rs
davidhewitt: Yeah I didn't think `ObjectProtocol::None` was ever a nice API 😄 .

I'll add detail to the `Changed` section where I explain the overall change.

guide/src/types.md
davidhewitt: Hmm it'll actually have to be `&**obj` in order to go `&PyList` -> `PyList` (deref) -> ` PyAny`.

Maybe it's nicer to just recommend `.as_ref()` here for now.

birkenfeld: It's good to remind people they can call `PyAny` methods via Deref, which doesn't require any calling or `*`.  But to explicitly go to `PyAny`, `as_ref()` is probably better.

davidhewitt:
I pushed a change to `types.md`:

- Swapped the order around. Previously was `PyObject`, `Py<T>`, `PyAny`, `PyList`. Now is `PyAny`, `PyList`, `PyObject`, `Py<T>`. I think this is better as `PyAny` is the type we want users to use most imo.
- Changed the conversion bullet points into doc-tested examples.

kngwyu:
Thank you!

PR #912, opened 2020-05-07
Make allow_threads safe with panics

davidhewitt:
I came across this crash while writing tests for #908 .

Seemed like an easy fix and very worth doing. Without this fix I get everything from stack overflow to sigsegv if I make a panic inside `allow_threads`.

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -136,9 +136,16 @@ impl<'p> Python<'p> {
         // transferring the `Python` token into the closure.
         unsafe {
             let save = ffi::PyEval_SaveThread();
-            let result = f();
+            // Unwinding right here corrupts the Python interpreter state and leads to weird
+            // crashes such as stack overflows. We will catch the unwind and resume as soon as
+            // we've restored the GIL state.
+            //
+            // Because we will resume unwinding as soon as the GIL state is fixed, we can assert
+            // that the closure is unwind safe.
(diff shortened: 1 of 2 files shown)


src/python.rs
```diff
@@ -136,9 +136,16 @@ impl<'p> Python<'p> {
...
         unsafe {
             let save = ffi::PyEval_SaveThread();
-            let result = f();
+            // Unwinding right here corrupts the Python interpreter state and leads to weird
+            // crashes such as stack overflows. We will catch the unwind and resume as soon as
+            // we've restored the GIL state.
+            //
+            // Because we will resume unwinding as soon as the GIL state is fixed, we can assert
+            // that the closure is unwind safe.
+            let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(f));
```
kngwyu, line 145: How about adding `F: Send + UnwindSafe + FnOnce() -> T` bound?

davidhewitt: I considered this, but it's a breaking change and I don't believe it is necessary here.

As I understand `UnwindSafe`, it is saying "if I `catch_unwind` here and continue my program execution, I will not be in a logically incorrect state".

Because we `resume_unwind` straight away after fixing the Python interpreter state, the program doesn't continue so it's not possible to see logically incorrect state here. So I don't think `UnwindSafe` is necessary.

I'm happy to ask this in users.rust-lang.org, or add `F: UnwindSafe` bound, if you would prefer.

PR #914, opened 2020-05-08
Allow `#[classattr]` on associated constants

scalexm:
```rust
#[pymethods]
impl Foo {
    #[classattr]
    const MY_CONST: &'static str = "foobar";
}
```

### DIFF
--- a/pyo3-derive-backend/src/pymethod.rs
+++ b/pyo3-derive-backend/src/pymethod.rs
@@ -1,4 +1,5 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
+use crate::konst::ConstSpec;
 use crate::method::{FnArg, FnSpec, FnType};
 use crate::utils;
 use proc_macro2::{Span, TokenStream};
@@ -31,7 +32,7 @@ pub fn gen_py_method(
         FnType::FnClass => impl_py_method_def_class(&spec, &impl_wrap_class(cls, &spec)),
         FnType::FnStatic => impl_py_method_def_static(&spec, &impl_wrap_static(cls, &spec)),
         FnType::ClassAttribute => {
-            impl_py_class_attribute(&spec, &impl_wrap_class_attribute(cls, &spec))
+            impl_py_method_class_attribute(&spec, &impl_wrap_class_attribute(cls, &spec))
         }
         FnType::Getter => impl_py_getter_def(
             &spec.python_name,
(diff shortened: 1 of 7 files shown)


pyo3-derive-backend/src/pymethod.rs
```diff
@@ -62,6 +63,23 @@ fn check_generic(sig: &syn::Signature) -> syn::Result<()> {
...
+pub fn gen_py_const(
+    cls: &syn::Type,
+    name: &syn::Ident,
+    attrs: &mut Vec<syn::Attribute>,
+) -> syn::Result<Option<TokenStream>> {
+    let spec = ConstSpec::parse(name, attrs)?;
+    if spec.is_class_attr {
+        let wrapper = quote! {
+            fn __wrap(py: pyo3::Python<'_>) -> pyo3::PyObject {
+                pyo3::IntoPy::into_py(#cls::#name, py)
```
davidhewitt, line 75: I think this will only work for `Copy` types, but I guess that's pretty much everything that could be generated by `const fn` anyway. Might want to document this restriction though?

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -631,8 +650,7 @@ pub fn impl_py_method_def_static(spec: &FnSpec, wrapper: &TokenStream) -> TokenS
     }
 }
 
-pub fn impl_py_class_attribute(spec: &FnSpec<'_>, wrapper: &TokenStream) -> TokenStream {
-    let python_name = &spec.python_name;
```
davidhewitt, line 635: I understand why you did this refactoring but I think I preferred how it was before, to keep symmetry with the the other `impl_py_method_xyz` functions.

scalexm: I hesitated making this change indeed. I’ll write two functions, it’ll duplicate code slightly but will preserve the symmetry.

pyo3-derive-backend/src/pymethod.rs
scalexm: Associated `const`, like any other `const` variables, are basically syntactically replaced by their compile-time evaluated expression everywhere they appear.

As a result, you can do:
```rust
const FOO: String = String::new();

fn foo(x: String) {}

fn main() {
    foo(FOO);
    foo(FOO);
}
```
although `String` is not copy.

davidhewitt: Oh yes, of course, my bad! 😄

davidhewitt:
Also I think the original #905 missed off a `CHANGELOG` entry, so can you add one here?

(Might want to wait to rebase on #915 to make life easier.)

kngwyu:
Thank you!

PR #917, opened 2020-05-09
Use PyMethodsImpl instead of *ProtocolImpl::methods

kngwyu:
To reduce specialization.
We have two kinds of protocol methods.
1. Slot methods. We have to set these methods to the type object. E.g., we have to set `mp_length` to enable `len(obj)`.
2. Normal methods, called `PyMethods` in our proc-macro implementation. 

For the later one, we don't need to use specialization. We can use the same trick as `#[pymethods]`.

#897 should be merged before this PR.
I'll add CHANGELOG later to make rebasing easier.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -17,76 +17,7 @@ struct MyClass {
 The above example generates implementations for [`PyTypeInfo`], [`PyTypeObject`],
 and [`PyClass`] for `MyClass`.
 
-Specifically, the following implementation is generated:
-
-```rust
-use pyo3::prelude::*;
-
-/// Class for demonstration
-struct MyClass {
-    num: i32,
-    debug: bool,
-}
-
-impl pyo3::pyclass::PyClassAlloc for MyClass {}
-
-unsafe impl pyo3::PyTypeInfo for MyClass {
-    type Type = MyClass;
-    type BaseType = PyAny;
-    type BaseLayout = pyo3::pycell::PyCellBase<PyAny>;
-    type Layout = PyCell<Self>;
-    type Initializer = PyClassInitializer<Self>;
-    type AsRefTarget = PyCell<Self>;
-
-    const NAME: &'static str = "MyClass";
-    const MODULE: Option<&'static str> = None;
-    const DESCRIPTION: &'static str = "Class for demonstration";
-    const FLAGS: usize = 0;
-
-    #[inline]
-    fn type_object() -> &'static pyo3::ffi::PyTypeObject {
-        use pyo3::type_object::LazyStaticType;
-        static TYPE_OBJECT: LazyStaticType = LazyStaticType::new();
-        TYPE_OBJECT.get_or_init::<Self>()
-    }
-}
-
-impl pyo3::pyclass::PyClass for MyClass {
-    type Dict = pyo3::pyclass_slots::PyClassDummySlot;
-    type WeakRef = pyo3::pyclass_slots::PyClassDummySlot;
-    type BaseNativeType = PyAny;
-}
-
-impl pyo3::IntoPy<PyObject> for MyClass {
-    fn into_py(self, py: pyo3::Python) -> pyo3::PyObject {
-        pyo3::IntoPy::into_py(pyo3::Py::new(py, self).unwrap(), py)
-    }
-}
-
-pub struct Pyo3MethodsInventoryForMyClass {
-    methods: &'static [pyo3::class::PyMethodDefType],
-}
-
-impl pyo3::class::methods::PyMethodsInventory for Pyo3MethodsInventoryForMyClass {
-    fn new(methods: &'static [pyo3::class::PyMethodDefType]) -> Self {
-        Self { methods }
-    }
-
-    fn get_methods(&self) -> &'static [pyo3::class::PyMethodDefType] {
-        self.methods
-    }
-}
-
-impl pyo3::class::methods::PyMethodsImpl for MyClass {
-    type Methods = Pyo3MethodsInventoryForMyClass;
-}
-
-pyo3::inventory::collect!(Pyo3MethodsInventoryForMyClass);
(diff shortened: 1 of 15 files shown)


guide/src/class.md
```diff
@@ -17,76 +17,7 @@ struct MyClass {
 The above example generates implementations for [`PyTypeInfo`], [`PyTypeObject`],
 and [`PyClass`] for `MyClass`.
 
-Specifically, the following implementation is generated:
```
kngwyu, line 20: I moved this code block to the last of the page. 
It's too long for a user who doesn't have a strong interest in the implementation detail.

davidhewitt: Good idea!

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -208,11 +208,9 @@ fn parse_descriptors(item: &mut syn::Field) -> syn::Result<Vec<FnType>> {
...
 
-/// The orphan rule disallows using a generic inventory struct, so we create the whole boilerplate
-/// once per class
-fn impl_inventory(cls: &syn::Ident) -> TokenStream {
-    // Try to build a unique type that gives a hint about it's function when
-    // it comes up in error messages
+/// To allow multiple #[pymethods] block, we define inventory types.
+fn impl_methods_inventory(cls: &syn::Ident) -> TokenStream {
+    // Try to build a unique type for better error messages
     let name = format!("Pyo3MethodsInventoryFor{}", cls);
```
davidhewitt, line 214: Looking at the comment "The orphan rule disallows using a generic inventory struct..."

It might be possible now to achieve this, as `Rust 1.41.0` relaxed orphan rules slightly: https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html#relaxed-restrictions-when-implementing-traits

So maybe we can have `Pyo3MethodsInventory<MyClass>`

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -57,53 +57,52 @@ fn impl_proto_impl(
     ty: &syn::Type,
     impls: &mut Vec<syn::ImplItem>,
     proto: &defs::Proto,
-) -> TokenStream {
-    let mut tokens = TokenStream::new();
+) -> syn::Result<TokenStream> {
```
davidhewitt, line 60: 👍

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -57,53 +57,52 @@ fn impl_proto_impl(
...
-
-                            Some(pyo3::class::PyMethodDef {
-                                ml_name: stringify!(#name),
-                                ml_meth: pyo3::class::PyMethodType::PyCFunctionWithKeywords(__wrap),
-                                // We need METH_COEXIST here to prevent __add__  from overriding __radd__
-                                ml_flags: pyo3::ffi::METH_VARARGS | pyo3::ffi::METH_KEYWORDS | #coexist,
-                                ml_doc: ""
-                            })
-                        }
+                // TODO(kngwyu): doc
```
davidhewitt, line 79: :)

src/class/methods.rs
```diff
@@ -157,7 +157,34 @@ pub trait PyMethodsImpl {
...
+pub trait PyProtoMethodsInventory: inventory::Collect {
+    /// Create a new instance
+    fn new(methods: &'static [PyMethodDef]) -> Self;
+
+    /// Returns 'non-slot' methods for a `#[pyproto] impl` block
+    fn get(&self) -> &'static [PyMethodDef];
+}
+
+/// Implementation detail. Only to be used through the proc macros.
+/// Similar to `PyProtoMethodsImpl`, but for `#[pyproto]`.
```
davidhewitt, line 177: "Similar to" looks wrong as it's listing the same type?

src/pyclass.rs
```diff
@@ -73,7 +75,7 @@ pub unsafe fn tp_free_fallback(obj: *mut ffi::PyObject) {
 /// The `#[pyclass]` attribute automatically implements this trait for your Rust struct,
 /// so you don't have to use this trait directly.
 pub trait PyClass:
-    PyTypeInfo<Layout = PyCell<Self>> + Sized + PyClassAlloc + PyMethodsImpl
+    PyTypeInfo<Layout = PyCell<Self>> + Sized + PyClassAlloc + PyMethodsImpl + PyProtoMethodsImpl
```
davidhewitt, line 78: I'm trying to understand - why is it necessary to make a new `PyProtoMethodsImpl` rather than using `PyMethodsImpl` for all of these?

kngwyu: Technically we can reuse `PyMethodsImpl` but I felt it easier to understand to have two inventory types corresponding to `#[pymethods]` and `#[pyproto]`.
Not so string reason, though.

pyo3-derive-backend/src/pyclass.rs
kngwyu: What the above change made possible is `impl<T> ForeignTrait<LocalType> for ForeignType<T>`.
`impl inventory::Collect for PyO3MethodsInventory<MyClass>` is still impossible.

src/class/methods.rs
kngwyu: :+1:

pyo3-derive-backend/src/pyclass.rs
davidhewitt: Ah yep you're correct 👍

src/pyclass.rs
davidhewitt: As this isn't a user-facing API, I guess the only people who need to understand it are us. I probably have a preference for just one inventory type so that there's less code overall to understand, but happy if you prefer to keep it as two.

kngwyu: I removed `PyProtoMethodsImpl`.

PR #918, opened 2020-05-10
Clarify pyo3 version on issue template

davidhewitt:
When looking at old bug reports, it's hard to tell from a glance if they are still relevant because they don't list exactly the pyo3 version they relate to. So I have adjusted the list of requested information slightly so this is captured.


kngwyu:
Thank you!

PR #920, opened 2020-05-11
Revert #889

kngwyu:
rust-numpy has a stub implementation of `PyMethodsImpl`, so it should not require inventory storage by default.


davidhewitt:
Interesting... `rust-numpy` is making custom `PyClass`?

kngwyu:
> Interesting... rust-numpy is making custom PyClass?

Yeah, it has a PyClass that wraps `Box<[T]>` to avoid copy entire buffer.

Issue #904, opened 2020-05-05
"extension-module" feature causing problems


m-ou-se:
See https://github.com/fusion-engineering/inline-python/issues/27

Enabling `features = ["extension-module"]` on `pyo3` in a crate that uses `inline-python` gives a linker error, because `"extension-module"` is then also enabled within `inline-python-macros`, which contains the proc macro that will use `pyo3` to compile the Python code to bytecode. That crate is not an extension module, so leaving the symbols undefined will not work, as it won't be loaded by Python, but by Rustc.

This makes me think that `"extension-module"` should not be a feature, as cargo features should be strictly additive. That the end user is making an extension module, doesn't always mean that everything in the dependency tree will end up in an extension module, although this might very well be limited to procedural macros.

However, I don't know of a solution. Even if the `ffi` part would be split off into a separate crate, it'd still need to be compiled with different linker flags for the procedural macro than for the final crate. Having a separate crate just for linking to the Python libraries might work, but will probably cause confusion and other problems.

Any ideas?

m-ou-se:
This is mostly a duplicate of #771, with the added detail that it's not about a `bin` crate here, but about a (`proc_macro`) `lib` crate.

This might make a solution involving `rustc-bin-link-arg` trickier.

Issue #771, opened 2020-02-18
The extension-module feature fails to be additive


nagisa:
In Rust it is the best practice to have features to be strictly additive. `extension-module`, instead, subtracts from the crate by not linking python. This can result in weird and difficult to resolve failure modes in e.g. workspaces.

For instance consider…

### 💥 Reproducing

```toml
[workspace]
members = [
    "extension",
    "some_binary",
]
```

where both `extension` and `some_binary` use as a dependency `pyo3`, but `extension` enables the `extension-module` feature, while `some_binary` explicitly disables the features.

When `cargo build` is run from the workspace root, then both of the binaries will get a `pyo3` with `extension-module` feature enabled and `some_binary` will fail to link.

This behaviour is entirely valid from `cargo`’s standpoint – it dictates that features should be additive and thus there must not be a way to write any single crate in a way that would fail when additional features are enabled in its dependencies.

To be honest, I don’t see a good way to fix it, other than spiting `pyo3` into two separate crates.

konstin:
The underlying problem are two cargo bugs.

One is the behavior of workspaces, as explained in https://github.com/rust-lang/cargo/issues/5928#issuecomment-415522044:

> This is a conflation of a number of preexisting bugs. They're not super well indexed unfortunately. Cargo unions all features which means with `--all` it would union all the features requested by anything, and a different bug means that Cargo never compiles multiple versions of a crate with different sets of features meaning that a crate is only compiled once with one set of features.
> 
> In that sense it's not compatible with features today in Cargo to have a set of binaries that only work with a feature and a set that only work without, and then build them all at once.

This might get solved by https://github.com/rust-lang/cargo/pull/7820.

The other problem is that cargo currently doesn't have target-dependent linker args (https://github.com/rust-lang/cargo/issues/5881). This would be solved by https://github.com/rust-lang/cargo/pull/781, but this seems to be stuck in a stalled RFC (https://github.com/rust-lang/cargo/pull/7811#issuecomment-575926481).

nagisa:
> One is the behavior of workspaces, as explained in ...

However Rust (and Cargo) project was operating presuming features to be additive (see also https://github.com/rust-lang/cargo/issues/4328) and a ton functionality assumes that enabling additional features is valid. A large many of workflows like `cargo test --all-features` and even outside of cargo assume this.

While some of the links do suggest very relevant solutions, given their state I’m not sure why pyo3 shouldn’t be investigating ways to resolve this and similar issues on its end.

[...]

PR #921, opened 2020-05-11
Replace extension-module feature with pyo3::main and pyo3::test

davidhewitt:
This is an experiment to remove `extension-module` feature, with its known linking problems. See #904 #771 #341 

Instead, to enable linking to `libpython` when needed, I introduce two proc macro attributes
`#[pyo3::main]` and `#[pyo3::test]` which force linking using the [macro suggested in #771](https://github.com/PyO3/pyo3/issues/771#issuecomment-589827334)

To sweeten the deal, both of these attributes can automatically acquire the GIL if you make the first argument `py: Python`. e.g.:

```rust
#[pyo3::main]
fn main(py: Python) {
  // use py! :)
}
```

This implementation is not yet complete - I hard-coded `libpython3.8` to test on my laptop. Needs to use proper detection of the python library. It also breaks windows where this hack isn't needed.

I'm happy with `#[pyo3::main]`, though I wonder if adding this link arg to _every_ `#[pyo3::test]` is overkill - an alternative would be to have a `pyo3_test_link_python!()` macro or so which could be included once in each test file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Correct FFI definitions `Py_SetProgramName` and `Py_SetPythonHome` to take `*const` argument instead of `*mut`. [#1021](https://github.com/PyO3/pyo3/pull/1021)
 - Rename `PyString::to_string` to `to_str`, change return type `Cow<str>` to `&str`. [#1023](https://github.com/PyO3/pyo3/pull/1023)
 - Correct FFI definition `_PyLong_AsByteArray` `*mut c_uchar` argument instead of `*const c_uchar`. [#1029](https://github.com/PyO3/pyo3/pull/1029)
+- Replace `extension-module` feature with `PYO3_MAKE_EXTENSION_MODULE` environment variable. See the migration guide for help with upgrading. [#1040](https://github.com/PyO3/pyo3/pull/1040)
 
 ### Removed
 - Remove `PyString::as_bytes`. [#1023](https://github.com/PyO3/pyo3/pull/1023)
 - Remove `Python::register_any`. [#1023](https://github.com/PyO3/pyo3/pull/1023)
+- Remove `extension-module` feature. [#1040](https://github.com/PyO3/pyo3/pull/1040)
 
 ### Fixed
 - Conversion from types with an `__index__` method to Rust BigInts. [#1027](https://github.com/PyO3/pyo3/pull/1027)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,11 +44,6 @@ nightly = []
 # this is no longer needed internally, but setuptools-rust assumes this feature
 python3 = []
 
-# Use this feature when building an extension module.
-# It tells the linker to keep the python symbols unresolved,
-# so that the module can also be used with statically linked python interpreters.
-extension-module = []
-
 # The stable cpython abi as defined in PEP 384. Currently broken with
 # many compilation errors. Pull Requests working towards fixing that
 # are welcome.
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ crate-type = ["cdylib"]
 
 [dependencies.pyo3]
 version = "0.11.1"
-features = ["extension-module"]
 ```
 
 **`src/lib.rs`**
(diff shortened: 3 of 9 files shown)


davidhewitt:
@m-ou-se I'm not quite sure if this solves what you need for `inline-python`. I think it might - users of `inline-python` will just have to add `#[pyo3::main]` where appropriate?

programmerjake:
it might be a good idea to be able to apply `#[pyo3::main]` or similar to a function other than `fn main()`, since there are systems that can't use the default `main` function, e.g. for libSDL on Android where the `main` is essentially provided by libSDL and the Android Java libraries.

kngwyu:
Awesome, thank you!
I'll add detailed comments later, but from a glance,

1. Do we need 
```rust
            #[link(name = "python3.8")]
            extern "C" {}
```
for each test function?

2. We should also allow `#[pyo3::main]fn main() {}`

programmerjake:
another idea is to just merge them both into `#[pyo3::entry]` or `#[pyo3::calls_python]` or some other suitable name

davidhewitt:
> Do we need `[...]` for each test function?

No, we only need this `#[link]` attribute once per executable really. So an alternative I was considering was that somewhere in a test file, users must add:

```
#[cfg(test)]
pyo3::link_python();
```

I think it does make sense for `#[pyo3::main]` to include this `#[link]`, as that should only ever appear once.

> another idea is to just merge them both into #[pyo3::entry] or #[pyo3::calls_python] or some other suitable name

I quite like `#[pyo3::main]` because it's an obvious name that hits 99% of use cases. I think for the niche use cases you describe above I'd rather recommend they just call `pyo3::link_python!()` explicitly.

davidhewitt:
Also, I quite like having `#[pyo3::test]` separate to `#[pyo3::main]` because the test macro can then generate the `#[test]` attribute, as well as acquire the GIL if needed.

But maybe if we prefer just calling `pyo3::link_python!()` once per test file, and remove `#[link]` from `#[pyo3::test]`, I should split it into a separate PR.

Let me know what design you all like best 😊

programmerjake:
honestly, I'm fine with having `#[pyo3::main/test]` as long as the link macro is also exposed for the corner cases.

m-ou-se:
I don't think an attribute like that is the way to go.

Some of my projects use Python somewhere in some module, but not directly from `main`. Attaching the attribute to `main` seems like the wrong place. It automatically locking the GIL is also not always what I'd want.

It also requires the use of a procedural macro, making the recently introduced `macros` feature a bit useless, as you'd always need a separate proc_macro crate with all the dependencies (quote, proc_macro2, syn, ..) again. I'd much prefer a simple `macro_rules!{}` macro like `pyo3::link_python!()`.

But even then, it seems not like a very ergonomic solution. Everybody who forgets this macro (including all existing code) will just get a long linker error with no proper explanation. :(

Maybe this is something to be solved on the side of Cargo/Rustc instead. This is not the only problem when compiling plugins/extensions (whether for Python or other things). There's the problem of `-undefined dynamic_lookup` on MacOS, there's the problem of the name of the compiled library being `libabc.so` instead of `abc.so`, and the problem of only allowing one plugin/extension per package/Cargo.toml, because they are a library and not binaries to Cargo.

Maybe it's time to propose an addition to Cargo, allowing something like
```toml
[[bin]]
crate-type = ["plugin"]
```
and then letting build scripts specify flags for these separately. Or something in that direction.

davidhewitt:
I agree that the error case for forgetting to link is not very ergonomic.

A cargo plugin sounds great, though I guess it could be a long time before the design and implementation of it is complete.

Maybe a better workaround for now is to keep `extension-module` for now and add the `pyo3::link_python!()` macro? That shouldn't break existing code and will let those that run into problem cases have a way to compile.

m-ou-se:
> Maybe a better workaround for now is to keep extension-module and add the pyo3::link_python!() macro? That shouldn't break existing code and will let those that run into problem cases have a way to compile.

@davidhewitt That might be the least surprising option for now. None of the solutions are perfect, but at least this one means that the regular use cases don't change, and the macro only needs to be used in the edge cases.

davidhewitt:
Yikes, accidentally re-used the same branch name for a new PR... Closing this PR. I plan to revisit the linking macro, but otherwise I think this is dead code.

Issue #916, opened 2020-05-08
Avoid spinlocks


programmerjake:
See https://matklad.github.io/2020/01/02/spinlocks-considered-harmful.html for some reasoning.

TLDR: spinlocks cause issues because the OS doesn't know when something is blocked, so can cause awful scheduling issues and big time delays. spinlocks are basically only appropriate for embedded or OS kernel software, and then you have to be really careful.

See also: #899

davidhewitt:
Thanks for that, that's a good article.

I think we should probably return to `parking_lot::Mutex` in this case. @m-ou-se as you contributed the PR which removed the `parking_lot` dependency, I just wanted to check this will not cause you big problems if I put this dependency back?

m-ou-se:
This crate did originally (2017-2020) use a spinlock, using the `spin` crate. The only reason a 'real' mutex (the one from `parking_lot`) was used starting four months ago, was because the `spin` crate was no longer maintained. See https://github.com/PyO3/pyo3/pull/734

@programmerjake Are you actually experiencing any problems with this?

Python isn't really usable with multiple threads anyway, with the global interpreter lock. So yes, you can send `PyObject`s to other threads and have them dropped there, but it seems like an unusual use case. Under normal circumstances, this lock is only used from one thread.

You could simply replace `spin_loop_hint()` inside the `compare_and_swap` loop by a call to `std::thread::yield_now()` if you're worried about the OS not being involved. But note that a regular `Mutex` also spins for a while until it even tries yielding or blocking through the OS, because that's what's usually most efficient.

kngwyu:
I also don't think this is so problematic since `spin_loop_hint` is rarely called.

IvanKuzavkov:
Hello, today we have moved in our project to the latest version 0.10, and it seems that spinlock dead lock is not so rare as you may expected, since we found it straight out after update - python process consumes 100% of CPU. I've attached sample of this process, and after remove spinlock from the project and returning `parking_lot::Mutex` eliminates this dead lock from our tests.

```
2185 pyo3::gil::GILPool::new::h000de5dc7faabf11  (in api.cpython-37m-darwin.so) + 37  [0x1068352f5]  gil.rs:268
2185 pyo3::gil::ReferencePool::update_counts::h939079ca1c222ef9  (in api.cpython-37m-darwin.so) + 471  [0x106ac6e57]  gil.rs:236
[...]

PR #924, opened 2020-05-13
Fix deadlock in update_counts

davidhewitt:
This fixes the deadlock reported in the discussion in #916, which became possible after incomplete test coverage of the change made in #899.

I wrote the test first and so can confirm the deadlock did indeed exist before this bugfix.

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -3,8 +3,8 @@
 //! Interaction with python's global interpreter lock
 
 use crate::{ffi, internal_tricks::Unsendable, Python};
-use std::cell::{Cell, RefCell, UnsafeCell};
-use std::sync::atomic::{spin_loop_hint, AtomicBool, Ordering};
+use parking_lot::{const_mutex, Mutex};
+use std::cell::{Cell, RefCell};
 use std::{any, mem::ManuallyDrop, ptr::NonNull, sync};
 
 static START: sync::Once = sync::Once::new();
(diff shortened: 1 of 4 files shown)


kngwyu:
Looks good, but I feel the combination of `Lock` and `UnsafeCell` redundant here. How about using `parking_lot::Mutex` or `std::sync::Mutex`?
Also, it would be good if @IvanKuzavkov could check this patch works.

IvanKuzavkov:
@kngwyu @davidhewitt I've checked this fix, it works well - thanks.

davidhewitt:
I would be happy to use a mutex to reduce amount of `unsafe` code. For now we'll probably need to use  `parking_lot` because `std::sync::mutex` cannot be easily constructed in a `static` without `lazy_static`. Also I think I read somewhere that `parking_lot` has better performance?

src/gil.rs
```diff
@@ -168,65 +168,39 @@ impl Drop for GILGuard {
...
-        }
-        Self { lock }
-    }
-}
-
-impl<'a> Drop for Lock<'a> {
-    fn drop(&mut self) {
-        self.lock.store(false, Ordering::Release);
-    }
+    pointers_to_incref: Mutex<Option<Vec<NonNull<ffi::PyObject>>>>,
```
kngwyu, line 171: Why Option here?

davidhewitt: Yeah I had thought [`Vec::new`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new) was not `const fn`, but I'm wrong, changing now...

davidhewitt:
Changed to remove `Option`. I gotta head to work; if you want this branch merged today and want any changes to it, please feel free to change as you like 😄

src/gil.rs
```diff
@@ -168,74 +168,56 @@ impl Drop for GILGuard {
...
+        self.pointers_to_decref.lock().push(obj)
     }
 
     fn update_counts(&self, _py: Python) {
-        let _lock = Lock::new(&self.locked);
+        macro_rules! swap_vec_with_lock {
+            // Get vec from one of ReferencePool's mutexes via lock, swap vec if needed, unlock.
+            ($cell:expr) => {{
+                let mut locked = $cell.lock();
+                let mut out = None;
```
kngwyu, line 196: I don't think we need Option here, too. Zero-sized Vec is internally a constant and fast.

davidhewitt: Yep agreed :+1: if you want to change this before I get a chance later tonight, please push to this branch :smile:

kngwyu: Done :+1:

kngwyu:
Thank you!

PR #927, opened 2020-05-15
Add nightly-optimizations feature

davidhewitt:
Add a feature to gate the sequence implementations using buffer + specialization behind feature flag.

Questions: 
 - Should this feature be disabled or enabled by default?
 - How should we configure CI to test this feature?

TODO:
- [ ] I'd like to write some documentation about this feature before merge. Wanted to propose this feature before I spend too long writing docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## Added
+- Gate some optimizations needing specialization in a `nightly-optimizations` feature (disabled by default). [#927](https://github.com/PyO3/pyo3/pull/927)
 
 ## [0.10.0] - 2020-05-14
 ### Fixed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,10 @@ python3 = []
 # so that the module can also be used with statically linked python interpreters.
 extension-module = []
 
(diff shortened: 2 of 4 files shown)


kngwyu:
Thank you, but I don't think we should do this now since we can't test that we can compile the code without `#![feature(specialization)]`.

davidhewitt:
That's ok. I'm happy to leave this PR to wait until we have finished removing specialization from `PyClass`.

Also I have some other ideas I am exploring with `ManagedPyRef` / `ToBorrowedObject` to remove specialization there, which will be the only other stuff left

davidhewitt:
FYI I continued to look at `ManagedPyRef` and `ToBorrowedObject`. I would propose that we remove `ManagedPyRef` for now (see #930) and add `nightly-optimizations` feature to `ToBorrowedObject`'s specialization.

I will leave this here for now until the provider pattern to remove specialization from `PyClass` is implemented. After that, we should be able to test this PR fully - and compile without specialization!

davidhewitt:
Just for completeness - my vision is that once we have made this feature gate, we can rework these optimizations to use `min_specialization`. And then we should eventually be able to get them onto stable Rust and remove the feature flag.

The feature flag will be a bridge which lets us compile on stable before `min_specialization` is stabilized.

davidhewitt:
FYI I will rebase this after #961 is merged - and (fingers crossed) add a CI for stable rust at the same time.

davidhewitt:
I will rebase this tonight unless you want me to close this and we focus on #969 ?

kngwyu:
I'm sorry I close this PR in favor of #961

PR #931, opened 2020-05-15
Add documentation for `PyModule::import`

davidhewitt:
Recently a few gitter questions have come up where `PyModule::import` is the solution, so I wanted to add a note to it in the guide.


kngwyu:
Thank you!

PR #934, opened 2020-05-17
Remove GetPropertyValue & improve property docs

### LINKED ISSUE #903: Generic lists

davidhewitt:
Now that we have `Clone` for `PyObject` and `Py<T>`, we don't need `GetPropertyValue` any more, so this cleans that up.

At the same time I also:
- Added a test for `Vec<PyObject>` with `#[pyo3(get)]`  (which was broken in `0.9.2`, see #903)
- Improved documentation for `#[pyo3(get, set)]`.

This is all internal refactoring so I think this can be merged without blocking us from making another patch release for `0.10.x` if necessary.

Closes #903

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Add FFI definition `PyObject_AsFileDescriptor` [#938](https://github.com/PyO3/pyo3/pull/938)
 
+### Changed
+- Simplify internals of `#[pyo3(get)]` attribute. (Remove the hidden API `GetPropertyValue`.) [#934](https://github.com/PyO3/pyo3/pull/934)
+
 ### Removed
 - Remove `ManagedPyRef` (unused, and needs specialization) [#930](https://github.com/PyO3/pyo3/pull/930)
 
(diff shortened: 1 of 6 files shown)


davidhewitt:
CHANGELOG added 👍

kngwyu:
Thanks!

PR #937, opened 2020-05-17
Add conversion table to guide

### LINKED ISSUE #374: Documentation: Add a table mapping Python types to Rust equivalents
It is often confusing trying to figure out how to write a `#[pyfunction]` signature that will match the Python signature you want (see #373, for example).

It would be great to have two lookup tables for all native types, one mapping the Python -> Rust, for writing the types of the function parameters, and another mapping Rust -> Python, saying the type of the object you'll get if you return a given type. So the first table would look like this:

| Python        | Rust                                          |
| ------------- |:---------------------------------------------:|
| `int`         | Any integer type (`i32`, `u32`, `usize`, etc) |
| `list`        | `Vec<T>`, `&PyList`                           |
| `tuple`       | `Vec<T>`, `&PyTuple`                          |
| `str`         | `String`, `&PyUnicode`                        |

The second one might look like this:

| Rust              | Python      |
| ------------------|:-----------:|
| `Vec<T>`          | `list`      |
| `u8`, `u16`, etc  | `int`       |

I don't think it needs to be comprehensive, but at least for some of the basic types (and especially showing the *preferred* way to do things) would be super helpful.

davidhewitt:
Closes #374 

Tried to keep it clear and short, but cover all relevant detail. Happy to edit any terms, wording or other suggestions you have.

### DIFF
--- a/guide/src/conversions.md
+++ b/guide/src/conversions.md
@@ -1,8 +1,104 @@
 # Type Conversions
 
+In this portion of the guide we'll talk about the mapping of Python types to Rust types offered by PyO3, as well as the traits available to perform conversions between them.
+
+## Mapping of Rust types to Python types
+
+When writing functions callable from Python (such as a `#[pyfunction]` or in a `#[pymethods]` block), the trait `FromPyObject` is required for function arguments, and `IntoPy<PyObject>` is required for function return values.
+
+Consult the tables in the following section to find the Rust types provided by PyO3 which implement these traits.
+
+### Argument Types
+
+When accepting a function argument, it is possible to either use Rust library types or PyO3's Python-native types. (See the next section for discussion on when to use each.)
+
+The table below contains the Python type and the corresponding function argument types that will accept them:
+
+| Python        | Rust                            | Rust (Python-native) |
+| ------------- |:-------------------------------:|:--------------------:|
+| `object`      | -                               | `&PyAny`             |
+| `str`         | `String`, `Cow<str>`, `&str`    | `&PyUnicode`         |
+| `bytes`       | `Vec<u8>`, `&[u8]`              | `&PyBytes`           |
+| `bool`        | `bool`                          | `&PyBool`            |
+| `int`         | Any integer type (`i32`, `u32`, `usize`, etc) | `&PyLong` |
+| `float`       | `f32`, `f64`                    | `&PyFloat`           |
+| `complex`     | `num_complex::Complex`[^1]      | `&PyComplex`         |
+| `list[T]`     | `Vec<T>`                        | `&PyList`            |
+| `dict[K, V]`  | `HashMap<K, V>`, `BTreeMap<K, V>` | `&PyDict`          |
+| `tuple[T, U]` | `(T, U)`, `Vec<T>`              | `&PyTuple`           |
+| `set[T]`      | `HashSet<T>`, `BTreeSet<T>`     | `&PySet`             |
+| `frozenset[T]` | `HashSet<T>`, `BTreeSet<T>`    | `&PyFrozenSet`       |
+| `bytearray`   | `Vec<u8>`                       | `&PyByteArray`       |
+| `slice`       | -                               | `&PySlice`           |
+| `type`        | -                               | `&PyType`            |
+| `module`      | -                               | `&PyModule`          |
+| `datetime.datetime` | -                         | `&PyDateTime`        |
+| `datetime.date` | -                             | `&PyDate`            |
+| `datetime.time` | -                             | `&PyTime`            |
+| `datetime.tzinfo` | -                           | `&PyTzInfo`          |
+| `datetime.timedelta` | -                        | `&PyDelta`           |
+| `typing.Optional[T]` | `Option<T>`              | -                    |
+| `typing.Sequence[T]` | `Vec<T>`                 | `&PySequence`        |
+| `typing.Iterator[Any]` | -                      | `&PyIterator`        |
+
+There are also a few special types related to the GIL and Rust-defined `#[pyclass]`es which may come in useful:
+
+| What          | Description |
+| ------------- | ------------------------------- |
+| `Python`      | A GIL token, used to pass to PyO3 constructors to prove ownership of the GIL |
+| `PyObject`    | A Python object isolated from the GIL lifetime. This can be sent to other threads. To call Python APIs using this object, it must be used with `AsPyRef::as_ref` to get a `&PyAny` reference. |
+| `Py<T>`       | Same as above, for a specific Python type or `#[pyclass]` T. |
+| `&PyCell<T>`  | A `#[pyclass]` value owned by Python. |
+| `PyRef<T>`    | A `#[pyclass]` borrowed immutably. |
+| `PyRefMut<T>` | A `#[pyclass]` borrowed mutably. |
+
+For more detail on accepting `#[pyclass]` values as function arguments, see [the section of this guide on Python Classes](class.md).
+
+#### Using Rust library types vs Python-native types
+
+Using Rust library types as function arguments will incur a conversion cost compared to using the Python-native types. Using the Python-native types is almost zero-cost (they just require a type check similar to the Python builtin function `isinstance()`).
+
+However, once that conversion cost has been paid, the Rust standard library types offer a number of benefits:
+- You can write functionality in native-speed Rust code (free of Python's runtime costs).
+- You get better interoperability with the rest of the Rust ecosystem.
+- You can use `Python::allow_threads` to release the Python GIL and let other Python threads make progress while your Rust code is executing.
+- You also benefit from stricter type checking. For example you can specify `Vec<i32>`, which will only accept a Python `list` containing integers. The Python-native equivalent, `&PyList`, would accept a Python `list` containing Python objects of any type.
+
+For most PyO3 usage the conversion cost is worth paying to get these benefits. As always, if you're not sure it's worth it in your case, benchmark it!
+
+### Returning Rust values to Python
+
+When returning values from functions callable from Python, Python-native types (`&PyAny`, `&PyDict` etc.) can be used with zero cost.
+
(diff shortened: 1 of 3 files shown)


kngwyu:
Thank you!
Generally, I think we should note both class name and mypy type.
E.g., `object` and `Any`, `dict` and `Dict[K, V]`, ...

In addition, it would be great if we have a review from someone with expertise in Python(e.g., @birkenfeld).

guide/src/conversions.md
```diff
@@ -1,8 +1,103 @@
...
+
+The table below contains the Python type and the corresponding Rust library type or PyO3 Python-native type that can be used as a function argument:
+
+| Python        | Rust                            | Rust (Python-native) |
+| ------------- |:-------------------------------:|:--------------------:|
+| `Any`         | -                               | `&PyAny`             |
+| `str`         | `String`, `&str`, `Cow<str>`    | `&PyUnicode`         |
+| `bytes`       | `&[u8]`                         | `&PyBytes`           |
+| `bool`        | `bool`                          | `&PyBool`            |
+| `int`         | Any integer type (`i32`, `u32`, `usize`, etc) | `&PyLong` |
```
birkenfeld, line 21: Should we mark cases where the conversion can fail, like here?

guide/src/conversions.md
```diff
@@ -1,8 +1,103 @@
...
+| `Optional[T]` | `Option<T>`                     | -                    |
+| `List[T]`     | `Vec<T>`                        | `&PyList`            |
+| `Dict[K, V]`  | `HashMap<K, V>`, `BTreeMap<K, V>` | `&PyDict`          |
+| `Tuple[T, U]` | `(T, U)`                        | `&PyTuple`           |
+| `Set[T]`      | `HashSet<T>`, `BTreeSet<T>`     | `&PySet`             |
+| `FrozenSet[T]` | `HashSet<T>`, `BTreeSet<T>`    | `&PyFrozenSet`       |
+| `FrozenSet[T]` | `HashSet<T>`, `BTreeSet<T>`    | `&PyFrozenSet`       |
+| `Sequence[T]` | `Vec<T>`                        | `&PySequence`        |
+| `Iterator[Any]` | -                             | `&PyIterator`        |
+| `bytearray`   | -                               | `&PyByteArray`       |
```
birkenfeld, line 33: Does `Vec<u8>` not work?

guide/src/conversions.md
```diff
@@ -1,8 +1,103 @@
...
+| `int`         | Any integer type (`i32`, `u32`, `usize`, etc) | `&PyLong` |
+| `float`       | `f32`, `f64`                    | `&PyFloat`           |
+| `complex`     | `num_complex::Complex`[^1]      | `&PyComplex`         |
+| `Optional[T]` | `Option<T>`                     | -                    |
+| `List[T]`     | `Vec<T>`                        | `&PyList`            |
+| `Dict[K, V]`  | `HashMap<K, V>`, `BTreeMap<K, V>` | `&PyDict`          |
+| `Tuple[T, U]` | `(T, U)`                        | `&PyTuple`           |
+| `Set[T]`      | `HashSet<T>`, `BTreeSet<T>`     | `&PySet`             |
+| `FrozenSet[T]` | `HashSet<T>`, `BTreeSet<T>`    | `&PyFrozenSet`       |
+| `FrozenSet[T]` | `HashSet<T>`, `BTreeSet<T>`    | `&PyFrozenSet`       |
```
birkenfeld, line 30: Doubled line.

guide/src/conversions.md
```diff
@@ -1,8 +1,103 @@
...
+
+When you accept a function argument, you can choose to either use Rust library types or PyO3's Python-native types. (See the next section for discussion on when to use each.)
+
+The table below contains the Python type and the corresponding Rust library type or PyO3 Python-native type that can be used as a function argument:
+
+| Python        | Rust                            | Rust (Python-native) |
+| ------------- |:-------------------------------:|:--------------------:|
+| `Any`         | -                               | `&PyAny`             |
+| `str`         | `String`, `&str`, `Cow<str>`    | `&PyUnicode`         |
+| `bytes`       | `&[u8]`                         | `&PyBytes`           |
```
birkenfeld, line 19: For example, `Vec<u8>` also works here.

guide/src/conversions.md
```diff
@@ -1,8 +1,103 @@
...
+
+## Mapping of Rust types to Python types
+
+When writing functions callable from Python (such as `#[pyfunction]`s and `#[pymethods]`) consult the tables below to determine the correct Rust types to use for argument and return types.
+
+### Argument Types
+
+When you accept a function argument, you can choose to either use Rust library types or PyO3's Python-native types. (See the next section for discussion on when to use each.)
+
+The table below contains the Python type and the corresponding Rust library type or PyO3 Python-native type that can be used as a function argument:
```
birkenfeld, line 13: This makes it sound like they are the only types; generally more than one can be allowed if `FromPyObject` is implemented.

guide/src/conversions.md
```diff
@@ -1,8 +1,103 @@
...
+- You can write functionality in native-speed Rust code (free of Python's runtime costs).
+- You get better interoperability with the rest of the Rust ecosystem.
+- You can use `Python::allow_threads` to release the Python GIL and let other Python threads make progress while your Rust code is executing.
+- You also benefit from stricter type checking. For example you can specify `Vec<i32>`, which will only accept a Python `list` containing integers. The Python-native equivalent, `&PyList`, would accept a Python `list` containing Python objects of any type.
+
+For most PyO3 usage the conversion cost is worth paying to get these benefits. As always, if you're not sure it's worth it in your case, benchmark it!
+
+### Returning Rust values to Python
+
+When returning values from `#[pyfunction]`s and `#[pymethods]`, Python-native types (`&PyAny`, `&PyDict` etc.) can be used with zero cost.
```
birkenfeld, line 70: Nit: put the `s` consistently either inside or out (#[pyfunction]s vs #[pymethods])

guide/src/conversions.md
```diff
@@ -1,8 +1,103 @@
...
+| What          | Description |
+| ------------- | ------------------------------- |
+| `Python`      | A GIL token, used to pass to PyO3 constructors to prove ownership of the GIL |
+| `PyObject`    | A Python object isolated from the GIL lifetime. This can be sent to other threads. To call Python APIs using this object, it must be used with `AsPyRef::as_ref` to get a `&PyAny` reference. |
+| `Py<T>`       | Same as above, for a specific Python type or `#[pyclass]` T. |
+| `&PyCell<T>`  | A `#[pyclass]` value owned by Python. |
+| `PyRef<T>`    | A `#[pyclass]` borrowed immutably. |
+| `PyRefMut<T>` | A `#[pyclass]` borrowed mutably. |
+
+For more detail on accepting `#[pyclass]` values as function arguments, see [the section of this guide on Python Classes](class.md).
```
birkenfeld, line 54: Quickly mention the required traits for conversion?

guide/src/conversions.md
```diff
@@ -1,8 +1,103 @@
...
+| `str`         | `String`, `&str`, `Cow<str>`    | `&PyUnicode`         |
+| `bytes`       | `&[u8]`                         | `&PyBytes`           |
+| `bool`        | `bool`                          | `&PyBool`            |
+| `int`         | Any integer type (`i32`, `u32`, `usize`, etc) | `&PyLong` |
+| `float`       | `f32`, `f64`                    | `&PyFloat`           |
+| `complex`     | `num_complex::Complex`[^1]      | `&PyComplex`         |
+| `Optional[T]` | `Option<T>`                     | -                    |
+| `List[T]`     | `Vec<T>`                        | `&PyList`            |
+| `Dict[K, V]`  | `HashMap<K, V>`, `BTreeMap<K, V>` | `&PyDict`          |
+| `Tuple[T, U]` | `(T, U)`                        | `&PyTuple`           |
```
birkenfeld, line 27: `Vec<T>` is also works (important for instances in Python where tuples are used as "frozenlist" instead of struct-like tuples).

guide/src/conversions.md
davidhewitt: Not sure. Technically all conversion can fail with a `TypeError` ;)

guide/src/conversions.md
davidhewitt: Good catch.

guide/src/conversions.md
davidhewitt: Mmm I agree - sadly that's what the attributes are called! I'll change the wording here.

guide/src/conversions.md
davidhewitt: Which traits do you mean? They use the same `FromPyObject` trait?

guide/src/conversions.md
davidhewitt: Yes, thanks - added test.

guide/src/conversions.md
davidhewitt: 👍

guide/src/conversions.md
davidhewitt: Thanks - added (and a test).

davidhewitt:
> Thank you!
Generally, I think we should note both class name and mypy type.
E.g., object and Any, dict and Dict[K, V], ...

In the interest of not duplicating things I'm going to switch to use `object`, `dict[K, V]` etc. as that's the syntax that will be official from Python 3.9 (as per [PEP 585](https://www.python.org/dev/peps/pep-0585/))

Thanks for the review @birkenfeld - I've pushed some changes based on your feedback 👍

kngwyu:
@birkenfeld Thank you for your review :slightly_smiling_face:

guide/src/conversions.md
birkenfeld: Ah right, my bad.

guide/src/conversions.md
birkenfeld: Yes, and ToPyObject for returning back from functions.

guide/src/conversions.md
davidhewitt: If you don't have a strong opinion on this, I think for the moment let's keep the doc simpler.

davidhewitt:
Thanks, have addressed the final points of feedback.

PR #942, opened 2020-05-20
Improve comment in iterator.rs

davidhewitt:
This is a totally mundane change to update a comment. 

The original comment made me nervous that we were masking up some unsafety, but on investigation I realised that actually the usage is correct and so I wanted to document it.


kngwyu:
Thank you!

PR #951, opened 2020-06-03
Rename PyMethodsImpl -> PyMethods

kngwyu:
As I'll introduce `PyProtoMethods`, I renamed `PyMethodsImple` for consistent naming.

### DIFF
--- a/src/class/methods.rs
+++ b/src/class/methods.rs
@@ -134,7 +134,16 @@ impl PySetterDef {
     }
 }
 
-/// Implementation detail. Only to be used through the proc macros.
+/// Indicates that the type `T` has some Python methods.
+pub trait PyMethods {
+    /// Returns all methods that are defined for a class.
+    fn py_methods() -> Vec<&'static PyMethodDefType> {
+        Vec::new()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


kngwyu:
Actions says
```

running 1 test
ERROR: failed to read cargo metadata: EOF while parsing a value at line 1 column 0

```
Wow.

davidhewitt:
How about just having one trait maybe called `PyClassMembers` with `py_methods()` and `py_proto_methods()`? Maybe this isn't possible if `PyProtoMethods` won't always be implemented.

I don't have strong opinion though so please continue with two traits if you prefer to keep them separate!

kngwyu:
>  just having one trait maybe called PyClassMembers

I think it's possible but can be confusing since `PyProtoMethods` has many methods like:
```rust
pub trait PyProtoMethods {
    fn basic_methods() -> Option<NonNull<PyObjectMethods>> {
        None
    }
    fn buffer_methods() -> Option<NonNull<PyBufferProcs>> {
        None
    }
    fn descr_methods() -> Option<NonNull<PyDescrMethods>> {
        None
    }
...
}
```

PR #952, opened 2020-06-04
Typed PyBuffer

kngwyu:
`PyBuffer` is now typed as `PyBuffer<T>`.
Since it can only have a single type, it is reasonable to make it typed.

In addition, I implemented `FromPyObject` for `PyBuffer<T>`.
The intention is enabling the use of `PyBuffer` as an argument of `pyfunction`, like:
```rust
#[pyfunction]
fn buffer_inplace_add(py: Python, x: PyBuffer<i32>, y: PyBuffer<i32>) { ... }
```
, which means that we can still use fast(=buffer to vec) conversion for function arguments even after #927 is merged.

### DIFF
--- a/tests/test_pyfunction.rs
+++ b/tests/test_pyfunction.rs
@@ -34,7 +34,6 @@ fn buffer_inplace_add(py: Python, x: PyBuffer<i32>, y: PyBuffer<i32>) {
 
 #[test]
 fn test_buffer_add() {
-    // Regression test for issue #932
     let gil = Python::acquire_gil();
     let py = gil.python();
     let f = wrap_pyfunction!(buffer_inplace_add)(py);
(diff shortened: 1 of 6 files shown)


ethanhs:
This looks really handy! I was just asking on Gitter if
>numpy PyObject -> PyArray -> extract an ArrayView -> to_owned_array to a OwnedArray -> into_raw_vec to Vec

Was the best route, this seems much more straightforward.

tests/test_pyfunction.rs
```diff
@@ -20,3 +21,45 @@ fn test_optional_bool() {
...
+    let y = y.as_slice(py).unwrap();
+    for (xi, yi) in x.iter().zip(y) {
+        let xi_plus_yi = xi.get() + yi.get();
+        xi.set(xi_plus_yi);
+    }
+}
+
+#[test]
+fn test_buffer_add() {
+    // Regression test for issue #932
```
davidhewitt, line 37: Looks like comment from copy-and-paste?

```suggestion
```

kngwyu: Oops, thanks.

PR #955, opened 2020-06-05
Small cleanup

Alexander-N:
- Small cleanup of Makefile to reuse targets
- Also check examples when running make clippy
- Fix or ignore clippy lints


kngwyu:
Thank you!

PR #957, opened 2020-06-05
Make benchmarks more comparable

### LINKED ISSUE #956: Parallel Rust is slower in benchmark than sequential Python
When running the [parallel benchmark from the guide](https://pyo3.rs/v0.10.1/parallelism.html#benchmark) I noticed that sequential Python was running faster than the parallel Rust example. This can also be seen in the CI tests, for example in https://github.com/PyO3/pyo3/runs/741594832?check_suite_focus=true:
```
----------------------------------------------------------------------------------------- benchmark: 3 tests ----------------------------------------------------------------------------------------
Name (time in ms)                         Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_word_count_python_sequential     15.2808 (1.0)      17.3052 (1.0)      15.8203 (1.0)      0.2878 (1.0)      15.8229 (1.0)      0.2674 (1.0)          14;1  63.2101 (1.0)          63           1
test_word_count_rust_parallel         22.2590 (1.46)     32.0573 (1.85)     23.9752 (1.52)     2.2992 (7.99)     22.6933 (1.43)     2.6776 (10.01)         4;3  41.7098 (0.66)         44           1
test_word_count_rust_sequential       83.0824 (5.44)     84.8830 (4.91)     83.6569 (5.29)     0.5247 (1.82)     83.4806 (5.28)     0.5954 (2.23)          4;1  11.9536 (0.19)         12           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Alexander-N:
Closes #956, I'll adapt the parallelism chapter in the guide in another pull request.

### DIFF
--- a/examples/word-count/tests/test_word_count.py
+++ b/examples/word-count/tests/test_word_count.py
@@ -8,8 +8,8 @@ current_dir = os.path.abspath(os.path.dirname(__file__))
 path = os.path.join(current_dir, "zen-of-python.txt")
 
 
-@pytest.fixture(scope="session", autouse=True)
-def textfile():
+@pytest.fixture(scope="session")
+def contents() -> str:
     text = """
 The Zen of Python, by Tim Peters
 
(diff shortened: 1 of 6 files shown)


examples/word-count/tests/test_word_count.py
```diff
@@ -8,8 +8,8 @@
 path = os.path.join(current_dir, "zen-of-python.txt")
 
 
-@pytest.fixture(scope="session", autouse=True)
-def textfile():
+@pytest.fixture(scope="session")
+def contents() -> str:
```
kngwyu, line 12: OMG we still support Python 3.5

kngwyu, line 12: Does it work with Python 3.5?

Alexander-N: Yes https://docs.python.org/3.5/library/typing.html

kngwyu:
Good refactoring :+1: 
I'm now investigating the way to enable `debug=True` only when testing. Please wai for a bit...

kngwyu:
> I'm now investigating the way to enable debug=True only when testing. 

I could not find the solution. So how about specifying `debug=False` in setup.py?

Alexander-N:
Closing in favor of #959

Issue #649, opened 2019-10-26
Python::allow_threads() causes segmentation fault


Askannz:
## 🐛 Bug Reports

Accessing some Python objects from within the closure given to `Python::allow_threads()` causes a segmentation fault. I can reproduce this behavior with `PyList`, `PyTuple`, and `PyArray` from the `numpy` crate (haven't tested with other objects).

### 🌍 Environment

 - Your operating system and version: Archlinux
 - Your python version: 3.7.4
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: Repo package + virtualenv
 - Your rust version (`rustc --version`): rustc 1.39.0-nightly (ca3766e2e 2019-09-14)
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` Yes

### 💥 Reproducing

Rust code:

```rust
use pyo3::prelude::*;
use pyo3::types::PyList;
use pyo3::wrap_pyfunction;

#[pymodule]
fn my_module(_py: Python, m: &PyModule) -> PyResult<()> {

    #[pyfunction]
    fn my_function(py: Python, list: &PyList) {
        py.allow_threads(|| {
            println!("{:?}", list);
        });
    }

    m.add_wrapped(wrap_pyfunction!(my_function))?;

    Ok(())
}
```

Python shell:

```
>>> from my_module import my_function
>>> my_function([1, 2, 3])
Segmentation fault (core dumped)
```

kngwyu:
This function is not intended to execute Python functions in parallel.
Python's GIL stops all threads other than the current interpreter thread, but this function removes this lock.
So this is needed to execute Rust's functions in parallel in Python extension codes.

In your example, you call `PyList::debug` which requires Python' GIL and looks problematic.

But...
> Maybe I'm just using it wrong, but in this case it should be marked unsafe and have some additional documentation.

(from your comment in #640)
Yeah, I agree.
We should do something for that.

sebpuetz:
@kngwyu Could you provide an example where this can be used? 

My example in #640 demonstrated that it's at least not required to get multi-threading going in pure Rust code called from Python.

birkenfeld:
> In your example, you call `PyList::debug`, which requires Python' GIL and looks problematic.

It seems that implementing the format traits for Py structs is a bad idea, since they internally call Python code without ensuring the GIL being held.

[...]

Issue #640, opened 2019-10-23
Use of Python::allow_threads


sebpuetz:
## 🐛 Bug Reports

Not really a bug, but anyways: The guide has a section on parallelism that states that `allow_threads` allows parallelism through e.g. rayon. The code in the example demonstrates

~~~Rust
        let count = py.allow_threads(move || {
            contents
                .par_lines()
                .map(|line| count_line(line, &search))
                .sum()
        });
~~~

is faster than:

~~~Rust
let result = contents.lines().map(|line| count_line(line, &needle)).sum();
~~~

which makes absolute sense. 

The point I'm getting at is, that `allow_threads` is not required in any way to run `.par_lines()` and changing `search_sequential` to the following makes it run as fast as the version with `allow_threads`:

~~~Rust
    /// Searches for a word in a classic sequential fashion
    fn search_sequential(&self, needle: String) -> PyResult<usize> {
        let contents = fs::read_to_string(&self.path)?;

        let result = contents
            .par_lines()
            .map(|line| count_line(line, &needle))
            .sum();

        Ok(result)
    }
~~~

![image](https://user-images.githubusercontent.com/30175181/67432401-825e1d80-f5e6-11e9-85b0-6b339913a789.png)

What is the actual use-case of `allow_threads` then?

programmerjake:
`allow_threads` allows other python code to run concurrently on other threads, since otherwise the GIL would block the other python threads.

sebpuetz:
Might just be my limited knowledge about Python, but do you mean other code as in Python code running in different subprocesses? Or how could the Python code proceed before the ffi call returns?

The following fn renders the process entirely unresponsive, so I don't see such an effect on the currently running process:

~~~Rust
#[pyfunction]
fn never_return() {
    let gil = Python::acquire_gil();
    Python::allow_threads(gil.python(), || {
        loop{}
    });
}
~~~

programmerjake:
> Might just be my limited knowledge about Python, but do you mean other code as in Python code running in different subprocesses?

I mean python code running different threads (yes, python [supports multithreading](https://docs.python.org/3/library/threading.html#thread-objects), but it's not commonly used due to the GIL only allowing one thread at a time to execute python code).

> Or how could the Python code proceed before the ffi call returns?

it doesn't in the current thread, but it can in other threads.

sebpuetz:
Thanks for the explanation!

Askannz:
[...]

PR #959, opened 2020-06-05
User Guide: Rewrite parallelism chapter

### LINKED ISSUE #956: Parallel Rust is slower in benchmark than sequential Python

Alexander-N:
I took a stab at rewriting the parallelism chapter because it seemed to suggest that you need to release the GIL in order to achieve parallelism within your Rust code and that `allow_threads` enables running of Python code in parallel. See #640 and #649 for discussion and thanks @Askannz for the nice example which I included.

Also included the previous PR #957 to make the benchmarks more comparable.

Closes #956

### DIFF
--- a/guide/src/parallelism.md
+++ b/guide/src/parallelism.md
@@ -3,7 +3,7 @@
 CPython has the infamous [Global Interpreter Lock](https://docs.python.org/3/glossary.html#term-global-interpreter-lock), which prevents that several threads can execute Python code in parallel. This makes threading in Python a bad fit for [CPU-bound](https://stackoverflow.com/questions/868568/) tasks and often forces developers to accept the overhead of multiprocessing.
 
 In PyO3 parallelism can be easily achieved in Rust-only code. Let's take a look at our [word-count](https://github.com/PyO3/pyo3/blob/master/examples/word-count/src/lib.rs) example, where we have a `search` function that utilizes the [rayon](https://github.com/nikomatsakis/rayon) crate to count words in parallel.
-```rust
+```rust, ignore
 #[pyfunction]
 fn search(contents: &str, needle: &str) -> usize {
     contents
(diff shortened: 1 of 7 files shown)


guide/src/parallelism.md
```diff
@@ -1,66 +1,75 @@
...
-hold the GIL by default when called from Python, in order to allow manipulation
-of Python objects, you can release the GIL when executing Rust-only code to
-achieve true parallelism.
-
-The [`Python::allow_threads`] method temporarily releases the GIL, thus allowing other Python threads to run.
-
-```rust,ignore
-impl Python {
-    pub fn allow_threads<T, F>(self, f: F) -> T where F: Send + FnOnce() -> T {}
+CPython has the infamous [Global Interpreter Lock](https://docs.python.org/3/glossary.html#term-global-interpreter-lock), which prevents that several threads can execute Python code in parallel. This makes threading in Python a bad fit for [CPU-bound](https://stackoverflow.com/questions/868568/) tasks and often forces developers to accept the overhead of multiprocessing.
```
pganssle, line 3: It's not really grammatical to say "prevents that", "prevents" usually takes the gerund: "CPython has the infamous Global Interpreter Lock which prevents several threads from executing Python code in parallel."

I also think maybe it should be "Python bytecode" instead of "Python code", like in the glossary entry. Possibly the meaning is the same, but out of context and without thinking too much about it "executing Python code" would, to me, include operations during which the GIL is *not* held, like importing a module or opening a file or something.

kngwyu:
Hmm, it looks that we have a problem with travis integration.
https://travis-ci.org/github/PyO3/pyo3/builds/695173811
It looks tests are correctly executed, but Github fails to fetch the result.

Alexander-N:
Closed and reopened to retrigger CI.

kngwyu:
Thanks!

Issue #940, opened 2020-05-19
Cannot use &PyClass as argument type in protocol methods


davidhewitt:
### 🌍 Environment
 - Your PyO3 version: `0.10.1`
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: Yes

### 💥 Reproducing

Create a custom object `#[pyclass]`, let's call it `MyClass`.

Try to use `&MyClass` as a function argument in a `#[pyproto]` block:

```
use pyo3::prelude::*;
use pyo3::basic::{CompareOp, PyObjectProtocol};

#[pyclass]
pub struct MyClass {}

#[pyproto]
impl PyObjectProtocol for MyClass {
    fn __richcmp__(&self, other: &MyClass, op: CompareOp) -> PyResult<()> {
        unimplemented!()
    }
}
```

This will fail to compile:

```
error[E0277]: the trait bound `&MyClass: pyo3::pyclass::PyClass` is not satisfied
   --> src\lib.rs:7:1
    |
7   | #[pyproto]
    | ^^^^^^^^^^ the trait `pyo3::pyclass::PyClass` is not implemented for `&MyClass`
    |
   ::: C:\Users\david\dev\pyo3\src\class\basic.rs:140:17
    |
140 |     type Other: FromPyObject<'p>;
    |                 ---------------- required by this bound in `pyo3::class::basic::PyObjectRichcmpProtocol`
    |
    = help: the following implementations were found:
              <MyClass as pyo3::pyclass::PyClass>
    = note: required because of the requirements on the impl of `pyo3::conversion::FromPyObject<'p>` for `&MyClass`
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

I expect this broke on `0.9.0` when we changed so that `&PyClass` implements `derive_utils::ExtractExt` instead of `FromPyObject`.

cc @kngwyu - I guess that when you refactor protocol methods to remove specialization this problem might automatically be fixed at the same time. It could be interesting to add a test for it then.

kngwyu:
I don't think it's possible.  Users can use `&PyRef` instead.
To enable this, we need a more complex trait that can represent `T: FromPyObject` or `U: FromPyObject, U: Deref<Target = T>`.

>  I guess that when you refactor protocol methods to remove specialization this problem might automatically be fixed at the same time. 

I don't think they're related.

davidhewitt:
[...]

Issue #844, opened 2020-03-30
__radd__ etc don't get called if __add__ is also defined


davidhewitt:
If we make a class like this:

```rust
#[pyclass]
#[derive(Debug)]
struct RhsArithmetic {}

#[pymethods]
impl RhsArithmetic {
    #[new]
    fn new() -> RhsArithmetic {
        RhsArithmetic {}
    }
}

#[pyproto]
impl PyNumberProtocol for RhsArithmetic {
    fn __mul__(lhs: PyRef<'p, Self>, rhs: &PyAny) -> PyResult<String> {
        Ok(format!("MUL"))
    }
    fn __rmul__(&self, other: &PyAny) -> PyResult<String> {
        Ok(format!("RMUL"))
    }
}
```

Then both `1 * rhs` and `rhs * 1` call the function in the `tp_as_number.nb_mul` slot. As of #839 this will be our `__mul__` implementation above (and never `__rmul__`).

And specifically it will call it with the arguments in the same order as invoked. So from above, we will have `__mul__(1, rhs)` and `__mul__(rhs, 1)`.

This is problematic, because `__mul__` as defined above will raise a `TypeError` if the first argument is not `Self`.

We saw this behaviour reported on Gitter. All three of the below will currently be calling `__mul__`, which is why the second will `TypeError`:

```
>>> rhs * 1.0
'MUL'
>>> 1.0 * rhs # should call __rmul__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError
>>> rhs * rhs
'MUL'
```

I am thinking that instead of doing [this](https://github.com/PyO3/pyo3/blob/1aa4ae82847ec418791419ff74f54e5b5f85b30f/src/class/number.rs#L650), if both `__mul__` and `__rmul__` are defined we should be creating a wrapper function which calls the correct function depending on the type of the first argument.

The implementation of [`PyNumber_Add`](https://github.com/python/cpython/blob/224481a8c988fca12f488544edd2f01c0af2a91d/Objects/abstract.c#L999) suggests that we should return `Py_NotImplemented` from this wrapper function if there is a type error, so that `sq_concat` is given a chance to be called also.

programmerjake:
Similar logic in [`PyNumber_Multiply`](https://github.com/python/cpython/blob/224481a8c988fca12f488544edd2f01c0af2a91d/Objects/abstract.c#L1030)

davidhewitt:
[...]

PR #961, opened 2020-06-06
Object protocols without specialization

kngwyu:
For #210.
Now all pyclass has a method registory for protocols:
```rust
impl pyo3::class::proto_methods::HasProtoRegistry for MyClass {
    fn registory() -> &'static pyo3::class::proto_methods::PyProtoRegistry {
        static REGISTRY: PyProtoRegistry = PyProtoRegistry::new();
        &REGISTRY
    }
}
```
, where `PyProtoRegistory` is a set of *method tables*:
```rust
#[doc(hidden)]
pub struct PyProtoRegistry {
    async_methods: AtomicPtr<PyAsyncMethods>,
    basic_methods: AtomicPtr<PyObjectMethods>,
    ...
}
```
.

Using the [`ctor`](https://github.com/mmastrac/rust-ctor) crate, we insert some initialization codes for this registry.
For example, when we write this code:
```rust
#[pyproto]
impl PyObjectProtocol for MyClass {
    fn __hash__(&self) -> PyResult<isize> {
        Ok(self.val as isize)
    }
}
```
, (roughly) the following code is generated:
```rust
#[ctor]
fn __init_Object_2817873() {
    let mut table = pyo3::class::PyObjectMethods::default();
    table.set_hash::<MyClass>();
    <MyClass as pyo3::class::proto_methods::HasProtoRegistry>::registory().set_basic_methods(table);
}
```

One disadvantage of this solution is it needs redundant `Box::new` for some protocols(basic, iter, and so on). However, to keep `PyProtoRegistry` simple and small, I chose this design.

Unfortunately, some protocols are not yet tested.

### TODO
- [x] Add tests for async protocols
  - [x] Maybe we need to change the trait definition to make it work?
- [x] Add tests for descr protocols
- [x] Maybe need some more doc comments?

### DIFF
--- a/pyo3-derive-backend/src/pyproto.rs
+++ b/pyo3-derive-backend/src/pyproto.rs
@@ -71,7 +71,7 @@ fn impl_proto_impl(
                 // Insert the method to the HashSet
                 method_names.insert(met.sig.ident.to_string());
             }
-            // Add non-slot methods to inventory slots
+            // Add non-slot methods to inventory like `#[pymethods]`
             if let Some(m) = proto.get_method(&met.sig.ident) {
                 let name = &met.sig.ident;
                 let fn_spec = FnSpec::parse(&met.sig, &mut met.attrs, false)?;
@@ -82,7 +82,7 @@ fn impl_proto_impl(
                 } else {
                     quote!(0)
                 };
-                // TODO(kngwyu): ml_doc
+                // TODO(kngwyu): Set ml_doc
                 py_methods.push(quote! {
                     pyo3::class::PyMethodDefType::Method({
                         #method
@@ -125,24 +125,23 @@ fn slot_initialization(
     ty: &syn::Type,
     proto: &defs::Proto,
 ) -> syn::Result<TokenStream> {
-    let mut initializers: Vec<TokenStream> = vec![];
     // Some setters cannot coexist.
     // E.g., if we have `__add__`, we need to skip `set_radd`.
-    let mut skipped_setters = HashSet::new();
-    'setter_loop: for m in proto.slot_setters {
-        if skipped_setters.contains(m.set_function) {
+    let mut skipped_setters = Vec::new();
+    // Collect initializers
+    let mut initializers: Vec<TokenStream> = vec![];
+    'outer_loop: for m in proto.slot_setters {
+        if skipped_setters.contains(&m.set_function) {
             continue;
         }
         for name in m.proto_names {
+            // If this `#[pyproto]` block doesn't provide all required methods,
+            // let's skip implementing this method.
             if !method_names.contains(*name) {
-                // If this `#[pyproto]` block doesn't provide all required methods,
-                // let's skip implementing this method.
-                continue 'setter_loop;
+                continue 'outer_loop;
             }
         }
-        for s in m.skipped_setters {
-            skipped_setters.insert(s.to_string());
-        }
+        skipped_setters.extend_from_slice(m.skipped_setters);
         // Add slot methods to PyProtoRegistry
         let set = syn::Ident::new(m.set_function, Span::call_site());
         initializers.push(quote! { table.#set::<#ty>(); });
--- a/pyo3-derive-backend/src/pyproto.rs
+++ b/pyo3-derive-backend/src/pyproto.rs
@@ -4,9 +4,10 @@ use crate::defs;
 use crate::func::impl_method_proto;
 use crate::method::FnSpec;
 use crate::pymethod;
-use proc_macro2::TokenStream;
+use proc_macro2::{Span, TokenStream};
 use quote::quote;
 use quote::ToTokens;
+use std::collections::HashSet;
 
 pub fn build_py_proto(ast: &mut syn::ItemImpl) -> syn::Result<TokenStream> {
     if let Some((_, ref mut path, _)) = ast.trait_ {
@@ -60,12 +61,17 @@ fn impl_proto_impl(
 ) -> syn::Result<TokenStream> {
     let mut trait_impls = TokenStream::new();
     let mut py_methods = Vec::new();
+    let mut method_names = HashSet::new();
 
     for iimpl in impls.iter_mut() {
         if let syn::ImplItem::Method(ref mut met) = iimpl {
+            // impl Py~Protocol<'p> { type = ... }
             if let Some(m) = proto.get_proto(&met.sig.ident) {
                 impl_method_proto(ty, &mut met.sig, m).to_tokens(&mut trait_impls);
+                // Insert the method to the HashSet
+                method_names.insert(met.sig.ident.to_string());
             }
+            // Add non-slot methods to inventory slots
             if let Some(m) = proto.get_method(&met.sig.ident) {
                 let name = &met.sig.ident;
                 let fn_spec = FnSpec::parse(&met.sig, &mut met.attrs, false)?;
@@ -76,7 +82,7 @@ fn impl_proto_impl(
                 } else {
                     quote!(0)
                 };
-                // TODO(kngwyu): doc
+                // TODO(kngwyu): ml_doc
                 py_methods.push(quote! {
                     pyo3::class::PyMethodDefType::Method({
                         #method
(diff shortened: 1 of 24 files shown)


kngwyu:
I forgot to mention that this PR has a breaking change. `nb_bool` is now moved under `PyNumberProtocol` from `PyObjectProtocol`.

src/class/pyasync.rs
```diff
@@ -85,92 +85,29 @@ pub trait PyAsyncAexitProtocol<'p>: PyAsyncProtocol<'p> {
...
-}
-
-impl<'p, T> PyAsyncAiterProtocolImpl for T
-where
-    T: PyAsyncProtocol<'p>,
-{
-    default fn am_aiter() -> Option<ffi::unaryfunc> {
-        None
+impl ffi::PyAsyncMethods {
+    pub fn set_await<T>(&mut self)
```
davidhewitt, line 89: As these are public types, but these methods aren't actually part of Python ffi interface, I wonder if these functions should be `pub(crate)`?

src/instance.rs
```diff
@@ -45,9 +45,8 @@ pub unsafe trait PyNativeType: Sized {
 #[repr(transparent)]
 pub struct Py<T>(NonNull<ffi::PyObject>, PhantomData<T>);
 
-unsafe impl<T> Send for Py<T> {}
-
-unsafe impl<T> Sync for Py<T> {}
+unsafe impl<T: Send> Send for Py<T> {}
+unsafe impl<T: Sync> Sync for Py<T> {}
```
davidhewitt, line 49: This isn't correct. `Py<PyAny>` should be both `Send` and `Sync`, but `PyAny` is neither.

src/pycell.rs
```diff
@@ -372,7 +372,7 @@ impl<T: PyClass> AsPyPointer for PyCell<T> {
     }
 }
 
-impl<T: PyClass> ToPyObject for &PyCell<T> {
+impl<T: PyClass + Send + Sync> ToPyObject for &PyCell<T> {
```
davidhewitt, line 375: As discussed in #948 I don't think we don't need these bounds. Let's keep it as it was?

```suggestion
impl<T: PyClass> ToPyObject for &PyCell<T> {
```

kngwyu:
Thank you for the review!
> Let's fix the Send + Sync bounds in #948 rather than adding in this PR.

I accidentally pushed these changes when I was playing with them. I'll revert them.

> Does it make the implementation much more complicated if we can keep this in the PyObjectProtocol?

I don't have any good solution now, unfortunately.

src/class/proto_methods.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+    fn sequence_methods() -> Option<NonNull<PySequenceMethods>> {
+        NonNull::new(Self::registory().sequence_methods.load(Ordering::SeqCst))
+    }
+}
+
+#[doc(hidden)]
+pub struct PyProtoRegistry {
+    /// Async protocols.
+    async_methods: AtomicPtr<PyAsyncMethods>,
```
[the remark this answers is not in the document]
kngwyu: Just because they are static but mutable.
Another solution(e.g.,  `RefCell<Option<NonNull<Methods>>>`, `fn registry() -> &'static RefCell<PyProtoRegistry>`)is also OK, but I'm not sure there's merit to do so.

src/class/pyasync.rs
kngwyu: They are called from proc-macro code so we can't do so.

src/class/proto_methods.rs
```diff
@@ -0,0 +1,152 @@
...
+            gc_methods: AtomicPtr::new(ptr::null_mut()),
+            mapping_methods: AtomicPtr::new(ptr::null_mut()),
+            number_methods: AtomicPtr::new(ptr::null_mut()),
+            iter_methods: AtomicPtr::new(ptr::null_mut()),
+            sequence_methods: AtomicPtr::new(ptr::null_mut()),
+        }
+    }
+    pub fn set_async_methods(&self, methods: PyAsyncMethods) {
+        self.async_methods
+            .store(Box::into_raw(Box::new(methods)), Ordering::SeqCst)
```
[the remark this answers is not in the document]
kngwyu: Yes, but the current code does the same. See https://github.com/PyO3/pyo3/blob/v0.10.1/src/pyclass.rs#L151.
The difference is we cause memory leak for all protocols(e.g., `iter` and `basic`).
We can avoid `Box::new` for those methods by changing the `pub struct PyProtoRegistry` like the below:
```rust
pub struct PyProtoRegistry {
    /// Protocols that needs specific table types
    async_methods: AtomicPtr<ffi::PyAsyncMethods>,
    buffer_methods: AtomicPtr<ffi::PyBufferProcs>,
    mapping_methods: AtomicPtr<ffi::PyMappingMethods>,
    number_methods: AtomicPtr<ffi::PyNumberMethods>,
    sequence_methods: AtomicPtr<ffi::PySequenceMethods>,
    /// Other methods.
    /// AtomicUsize is used instead of `Atomic<Option<ffi::func>>` or so because
    /// we can't use function pointer in const function.
    tp_str: AtomicUsize,
    tp_repr: AtomicUsize.
    ....
}
```
.

I chose this design first but abandoned it due to its ugliness.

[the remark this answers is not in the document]
davidhewitt: Cool, I think your design choice is fine. Can we maybe just add a docstring for now to warn calling these twice will leak?

davidhewitt:
> I don't have any good solution now, unfortunately.

:+1: ok. I have a half-baked idea but it's hard to write down as I haven't properly thought about it yet. I'll try to find time tomorrow evening to experiment with it whether it can help and explain it better then.

kngwyu:
> I have a half-baked idea but it's hard to write down as I haven't properly thought about it yet. 

Thank you but wait for a bit...
I don't think most users know that `__bool__` is classified as [a part of *Object* protocols](https://docs.python.org/3.8/reference/datamodel.html#object.__bool__).
Thus I don't think it's worth doing if we need more codes to do so.

kngwyu:
At last, I decided to move `nb_bool` under `PyObjectProtocol` again.
I don't think the code is *very* ugly but please check it.

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -91,20 +97,79 @@ fn impl_proto_impl(
...
+
+fn slot_initialization(
+    method_names: HashSet<String>,
+    ty: &syn::Type,
+    proto: &defs::Proto,
+) -> syn::Result<TokenStream> {
+    let mut initializers: Vec<TokenStream> = vec![];
+    // Some setters cannot coexist.
+    // E.g., if we have `__add__`, we need to skip `set_radd`.
+    let mut skipped_setters = HashSet::new();
```
davidhewitt, line 131: Do you think that we still need `skipped_setters` once we implement a solution for #844?

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -76,7 +82,7 @@ fn impl_proto_impl(
                 } else {
                     quote!(0)
                 };
-                // TODO(kngwyu): doc
+                // TODO(kngwyu): ml_doc
```
davidhewitt, line 85: `ml_doc`?

src/class/pyasync.rs
davidhewitt: That's fair enough. Maybe then we need `#[doc(hidden)]` for all these setters?

```suggestion
    #[doc(hidden)]
    pub fn set_await<T>(&mut self)
```

src/class/macros.rs
```diff
@@ -106,7 +106,7 @@ macro_rules! py_binary_num_func {
 
 #[macro_export]
 #[doc(hidden)]
-macro_rules! py_binary_reverse_num_func {
+macro_rules! py_binary_reversed_num_func {
```
davidhewitt, line 109: 👍

src/pyclass.rs
```diff
@@ -135,35 +138,43 @@ where
...
+        nb_bool = unsafe { basic.as_ref() }.nb_bool;
     }
 
     // number methods
-    type_object.tp_as_number = to_ptr(<T as class::number::PyNumberProtocolImpl>::tp_as_number());
+    type_object.tp_as_number = T::number_methods()
+        .map(|mut p| {
+            unsafe { p.as_mut() }.nb_bool = nb_bool;
+            p.as_ptr()
+        })
```
davidhewitt, line 168: Yep I like how you've done this for `nb_bool` - nice 👍 👍 👍

davidhewitt:
It would also be good to add a test for #940 in this PR - if I read your implementation correctly I *think* that should already be solved by what you've done!

kngwyu:
Added a test for `__await__`.
I was really confused by Python's async/await and noticed some things:
- Returning `Future` from Rust is hard. 
- To return some value, we have to make `StopIterException` have the value.
But for now, I think this is sufficient for testing proc-macros.

kngwyu:
OMG...
image
(Windows + Python3.8)

tests/test_dunder.rs
```diff
@@ -552,3 +553,77 @@ fn getattr_doesnt_override_member() {
...
+        Ok(slf.into())
+    }
+}
+
+#[pyproto]
+impl PyIterProtocol for OnceFuture {
+    fn __iter__(slf: PyRef<Self>) -> PyResult<Py<Self>> {
+        Ok(slf.into())
+    }
+    fn __next__(mut slf: PyRefMut<Self>) -> PyResult<Option<PyObject>> {
```
davidhewitt, line 587: Just curious: can this be `fn __next__(&mut self) -> Option<PyObject>`  ?

tests/test_dunder.rs
```diff
@@ -552,3 +553,77 @@ fn getattr_doesnt_override_member() {
...
+    fn __await__(slf: PyRef<Self>) -> PyResult<Py<Self>> {
+        Ok(slf.into())
+    }
+}
+
+#[pyproto]
+impl PyIterProtocol for OnceFuture {
+    fn __iter__(slf: PyRef<Self>) -> PyResult<Py<Self>> {
+        Ok(slf.into())
+    }
```
davidhewitt, line 586: Why `__iter__` / `__next__` instead of `__aiter__` / `__anext__` ?

Also can this method just return `PyRef<Self>` ?

tests/test_dunder.rs
```diff
@@ -552,3 +553,77 @@ fn getattr_doesnt_override_member() {
...
+loop.close()
+"#
+    );
+    let globals = PyModule::import(py, "__main__").unwrap().dict();
+    globals.set_item("Once", once).unwrap();
+    py.run(source, None, Some(globals))
+        .map_err(|e| {
+            e.print(py);
+            py.run("import sys; sys.stderr.flush()", None, None)
+                .unwrap();
```
davidhewitt, line 626: Is the `flush()` still needed now that we have the `Py_Finalize()` call?

kngwyu: Removed it, thanks.

tests/test_dunder.rs
kngwyu: We can't do so since `None` is converted to `exceptions::StopIteration`.
Python C-API allow `yield` some value by wrapping it by `StopIteration` so maybe it's better to return `PyResult<Either<Future, Result>>`. But it would too much for this PR.

tests/test_dunder.rs
kngwyu: Good catch :+1:

kngwyu: > Why __iter__ / __next__ instead of __aiter__ / __anext__ ?

See the [document](https://docs.python.org/3/reference/datamodel.html#object.__await__).
I feel it confusing but `__await__` should return iterator.

tests/test_dunder.rs
davidhewitt: Ah, ok sounds good :+1:

madhavajay:
🍿🤓

kngwyu:
> I would be interested to hear what you think about how we could solve #844 with this new implementation.

I think we can reuse the same trick as `__set/delitem__` for that, but maybe it should be done in another PR :thinking:

davidhewitt:
> I think we can reuse the same trick as __set/delitem__ for that, but maybe it should be done in another PR 🤔

I'm happy with that - this PR is already quite large 😄

kngwyu:
Now I think this PR is ready.
Though I added more comments than usual, please point out what you think is unclear.

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -76,7 +82,7 @@ fn impl_proto_impl(
                 } else {
                     quote!(0)
                 };
-                // TODO(kngwyu): doc
+                // TODO(kngwyu): Set ml_doc
```
davidhewitt, line 85: Let's create an issue for this so that we remember to do it?

konstin:
The `test_await` was failing for me locally (python3.6 from apt in virtualenv), this is the workaround I used to make them pass.

```
  File "<string>", line 12, in <module>
  File "/usr/lib/python3.6/asyncio/events.py", line 694, in get_event_loop
    return get_event_loop_policy().get_event_loop()
  File "/usr/lib/python3.6/asyncio/events.py", line 602, in get_event_loop
    % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'Dummy-1'.
test test_await ... FAILED
```

```diff
diff --git a/tests/test_dunder.rs b/tests/test_dunder.rs
index 67a66fb49..47f22fc91 100644
--- a/tests/test_dunder.rs
+++ b/tests/test_dunder.rs
@@ -612,7 +612,11 @@ async def main():
 # but we see still errors on Github actions...
 if sys.platform == "win32" and sys.version_info >= (3, 8, 0):
     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
-loop = asyncio.get_event_loop()
+try:
+    loop = asyncio.get_event_loop()
+except RuntimeError: # RuntimeError: There is no current event loop in thread 'Dummy-1'.
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
 assert loop.run_until_complete(main()) is None
 loop.close()
 "#
```

kngwyu:
@konstin 
Thanks, I pushed the change based on your diff.

davidhewitt:
🎉

PR #329, opened 2019-01-27
Update Get Started section in user guide

Alexander-N:
Use content of README.md for an updated version of
guide/src/get_started.md

PR #962, opened 2020-06-06
Update README and remove Getting Started section from user guide

Alexander-N:
Updated the README with content from the Getting Started section of the user guide.

I suggest we remove this section completely, since it's content should be exactly the same as in the README, but both files have to be kept in sync constantly (see for example #329 and #476).


kngwyu:
> it's content should be exactly the same as in the README, but both files have to be kept in sync constantly 

Indeed it's annoying, thank you!

README.md
```diff
@@ -61,13 +61,13 @@ use pyo3::prelude::*;
...
 #[pyfunction]
-/// Formats the sum of two numbers as string
+/// Formats the sum of two numbers as string.
 fn sum_as_string(a: usize, b: usize) -> PyResult<String> {
     Ok((a + b).to_string())
 }
 
-/// This module is a python module implemented in Rust.
 #[pymodule]
+/// A Python module implemented in Rust.
```
davidhewitt, line 70: Actually [the rust style guide](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md#doc-comments) suggests "Doc comments should come before attributes".

README.md
```diff
@@ -61,13 +61,13 @@ use pyo3::prelude::*;
 use pyo3::wrap_pyfunction;
 
 #[pyfunction]
-/// Formats the sum of two numbers as string
+/// Formats the sum of two numbers as string.
```
davidhewitt, line 64: Same here, can we move the doc comment first?

PR #964, opened 2020-06-08
Doc trait exposition

Nateckert:
Contribution to the documentation:

- New appendix C in the guide.
- Adding some precision in several sections in the doc and the guide


guide/src/module.md
```diff
@@ -34,6 +34,9 @@ fn sum_as_string(a: i64, b: i64) -> String {
 
 The `#[pymodule]` procedural macro attribute takes care of exporting the initialization function of your module to Python. It can take as an argument the name of your module, which must be the name of the `.so` or `.pyd` file; the default is the Rust function's name.
 
+If the name of the module (the default being the function name) does not match the name of the `.so` or `.pyd` file, you may get an import error in Python with the following message:
+`ImportError: dynamic module does not define module export function (PyInit_name_of_your_module)`
```
davidhewitt, line 38: 👍 Good idea to add this - happens to many beginners! I suggest changing "you may get an import error" to "you **will** get an import error". 😄

guide/src/get_started.md
```diff
@@ -81,6 +81,11 @@ rustflags = [
 
 While developing, you can symlink (or copy) and rename the shared library from the target folder: On MacOS, rename `libstring_sum.dylib` to `string_sum.so`, on Windows `libstring_sum.dll` to `string_sum.pyd`, and on Linux `libstring_sum.so` to `string_sum.so`. Then open a Python shell in the same folder and you'll be able to `import string_sum`.
 
+Adding the `cdylib` arguments in the `Cargo.toml` files changes the way your crate is compiled.
+Other Rust projects using your crate will have to link against the `.so` or `.pyd` file rather than include your library directly as normal.
+In order to make available your crate in the usual way for Rust user, you you might want to consider using both `crate-type = ["cdylib", "rlib"]` so that Rust users can use the `rlib` (the default lib crate type).
+Another possibility is to create a new crate to perform the binding.
```
davidhewitt, line 87: These notes are a good idea, but we have just removed the `get_started` chapter so for now please move these to the same section on the README.

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+  #     let py = gil.python();
+  #     self.model
+  #         .as_ref(py)
+  #         .call_method("compute", (), None)
+  #         .unwrap();
+  # }
+}
+```
+
+However, the prievious code will not compile. The compilation error is the following one:
```
davidhewitt, line 220:
```suggestion
However, the previous code will not compile. The compilation error is the following one:
```

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+`error: #[pymethods] can not be used only with trait impl block`
```
davidhewitt, line 221: Eww didn't know that error was so gammy! Just opened #965 to fix, would be great if you could update this PR after that gets merged.

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+  #
+  # fn compute(&mut self) {
+  #     println!("Rust calling Python to perform the computation");
+  #     let gil = Python::acquire_gil();
+  #     let py = gil.python();
+  #     self.model
+  #         .as_ref(py)
+  #         .call_method("compute", (), None)
+  #         .unwrap();
+  # }
```
davidhewitt, line 216: This code is a lot of hidden code that I don't think we need - can we either remove it or add doctest?

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+`error: #[pymethods] can not be used only with trait impl block`
+
+That's a bummer!
+However, we can write a second wrapper around these functions to call them directly.
+This wrapper will also perform the type conversions between Python and Rust.
+
+```rust
+#[pymethods]
+impl UserModel {
+
```
davidhewitt, line 230:
```suggestion
```

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+    pub fn set_variables_in_rust(&mut self, var: Vec<f64>) -> PyResult<()> {
```
davidhewitt, line 231: I'm not 100% sure, but I think you can actually still call these functions `set_variables` etc. without the `_in_rust` suffix.

(I believe the inherent method won't conflict with the trait one; you can still call the trait ones by `Model::set_variables(self, &var)` etc.)

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+
+At the execution of Python, we get :
+
+```block
+File "main.py", line 15, in <module>
+   my_rust_model.set_variables_in_rust(2)
+TypeError
+```
+
+It is a type error and Python points to it, so it's easy to identify and solve.
```
davidhewitt, line 318: One day I hope to add detail to these `TypeError`s!

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+      let gil = Python::acquire_gil();
+      let py = gil.python();
+      let py_result: &PyAny = self
+          .model
+          .as_ref(py)
+          .call_method("get_results", (), None)
+          .unwrap();
+
+      if py_result.get_type().name() != "list" {
+          panic!("Expected a list for the get_results() method signature, got {}", py_result.get_type().name());
```
davidhewitt, line 382: Instead of panicking, the UserModel trait could be modified to return some form of `Result`. Did you consider this option, and if so, why did you choose against it?

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+      py_result.extract().unwrap()
+  }
+}
+```
+
+By doing so, you catch the result of the Python computation and check its type in order to be able to deliver a better error message before performing the unwrapping.
+
+Of course, it does not cover all the possible wrong outputs:
+the user could return a list of strings instead of a list of floats.
+In this case, a runtime panic would still occur due to PyO3, but with an error message much more difficult to decipher for non-rust user.
```
davidhewitt, line 393: Note that we might improve the panic error messages built into pyo3 in the future too 😄

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,509 @@
...
+    fn set_variables(&mut self, var: &Vec<f64>);
+    fn get_results(&self) -> Vec<f64>;
+    fn compute(&mut self);
+}
+
+pub fn solve<T: Model>(model: &mut T) {
+  println!("Magic solver that mutates the model into a resolved state");
+}
+
+#[pyfunction]
```
davidhewitt, line 424: IIRC I implemented `#[name]` attribute for pyfunctions:

```suggestion
#[pyfunction]
#[name = "solve"]
```

guide/src/trait_bounds.md
Nateckert: Sure, I will edit this PR as soon as it is merged

guide/src/get_started.md
Nateckert: I merge master into my PR and added that note in the README

guide/src/trait_bounds.md
Nateckert: Corrected in https://github.com/PyO3/pyo3/pull/964/commits/97ef5b0daa0c538dd9131cfc44082599b57f69e6

guide/src/trait_bounds.md
Nateckert: I removed it

guide/src/trait_bounds.md
Nateckert: I did not consider it as I assumed the trait could be changed as I was in a scenario of having a code base that we wanted to make available in Python without actually changing it.

guide/src/trait_bounds.md
Nateckert: 👍

guide/src/trait_bounds.md
Nateckert: That would be an improvement! However, I think it is already easy to debug as Python points to the incorrect call.

guide/src/trait_bounds.md
Nateckert: I tried and indeed it is possible, I added it in https://github.com/Nateckert/pyo3/commit/3ae6c472d85aaf975b8ed257a44dcd2cefe09302

Nateckert:
Hi,
Thanks for the suggestion, I added the chapter to doctest ! There is one example that is expected not to compile and I tagged it as such.

guide/src/trait_bounds.md
```diff
@@ -0,0 +1,630 @@
...
+#       println!("Rust calling Python to perform the computation");
+#       let gil = Python::acquire_gil();
+#       let py = gil.python();
+#       self.model
+#           .as_ref(py)
+#           .call_method("compute", (), None)
+#           .unwrap();
+#   }
+}
+```
```
davidhewitt, line 217:
```suggestion
```rust,ignore
#[pymethods]
impl Model for UserModel {
  // the previous trait implementation
}
```
```

kngwyu:
I cannot afford to give a detailed review now but this example looks nice!
Thanks @Nateckert !

PR #965, opened 2020-06-08
Improve error messages in pyimpl.rs

davidhewitt:
Just noticed these and thought I'd tweak them. Not sure if CHANGELOG entry is necessary.


kngwyu:
Thanks!

PR #966, opened 2020-06-08
Require Send for #[pyclass] (no compilefail test)

davidhewitt:
This is the same as #948, rebased on master, without the compile_fail test. I wanted to propose merging this fix without blocking on the `trybuild` issue.

I will leave #948 open, and once my bugfix for `trybuild` is released, I will rebase it and we can merge the test.


kngwyu:
Thanks!

PR #967, opened 2020-06-08
Add accessor methods to `PyByteArray`

### LINKED ISSUE #632: Add fn as_bytes to PyByteArray
I understand the rational for to_vec (https://github.com/PyO3/pyo3/issues/373#issuecomment-512332696) but in my use case it won't never happen (I just need to read data and nothing will modified them).
So it'd be nice to avoid a useless copy here.

davidhewitt:
Closes #632 .

This adds `PyByteArray::data` and `PyByteArray::as_bytes` as proposed in that issue.

I decided to also add `as_bytes_mut`, but if you think it's too far I can happily remove it. (Users can always implement themselves using `std::slice::from_raw_parts_mut(b.data(), b.len())`.)

### DIFF
--- a/src/types/bytearray.rs
+++ b/src/types/bytearray.rs
@@ -42,6 +42,51 @@ impl PyByteArray {
         self.len() == 0
     }
 
+    /// Get the start of the buffer containing the contents of the bytearray.
+    ///
+    /// Note that this bytearray object is both shared and mutable, and the backing buffer may be
+    /// reallocated if the bytearray is resized. This can occur from Python code as well as from
+    /// Rust via [PyByteArray::resize].
+    ///
+    /// As a result, the returned pointer should be dereferenced only if since calling this method
+    /// no Python code has executed, [PyByteArray::resize] has not been called.
+    pub fn data(&self) -> *mut u8 {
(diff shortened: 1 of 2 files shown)


src/types/bytearray.rs
```diff
@@ -42,6 +42,48 @@ impl PyByteArray {
...
+    /// As a result, the returned pointer should be dereferenced only if since calling this method
+    /// no Python code has executed, [PyByteArray::resize] has not been called.
+    pub fn data(&self) -> *mut u8 {
+        unsafe { ffi::PyByteArray_AsString(self.as_ptr()) as *mut u8 }
+    }
+
+    /// Get the contents of this buffer as a slice.
+    ///
+    /// # Safety
+    /// This is equivalent to `std::slice::from_raw_parts(self.data(), self.len())`, and so all
```
kngwyu, line 60: This document is well written, but could you please emphasize that **we should not resize or delete the bytearray while holding the slice**?

davidhewitt: Thanks, tweaked the doc slightly - what do you think of it now?

kngwyu: :+1:

kngwyu:
Thank you!

PR #968, opened 2020-06-09
Tidy up docs for from_borrowed_ptr methods

davidhewitt:
The release pool doesn't exist for borrowed objects any more, so just tidied these docs up.


kngwyu:
Good catch, thank you!

PR #970, opened 2020-06-12
Cleanups to ensure GIL-safety of Py<T> and PyObject methods

### LINKED ISSUE #949: Audit Py<T> methods for GIL safety
As part of the invariant that `Py<T>`'s data can only be accessed when the GIL is locked, all the member functions should either acquire the GIL or take a GIL token as an argument. [`Py::get_refcnt`](https://pyo3.rs/master/doc/pyo3/struct.Py.html#method.get_refcnt) and several of the `unsafe` functions doesn't appear to lock the GIL internally.

Note that it _is_ Undefined Behavior to read the reference count using a plain memory read (what `ffi::Py_REFCNT` does) while a different thread is concurrently modifying it, making `get_refcnt` have safety issues.

_Originally posted by @programmerjake in https://github.com/PyO3/pyo3/pull/948#issuecomment-637353345_

Proposed actions to resolve this:

- Add `Python` token as an argument to all safe function that require the GIL.
- Add `# Safety` documentation to all unsafe functions that require the GIL. (Or alternatively add `Python` token argument to them too.)

davidhewitt:
Closes #949 

This introduces `Python` arguments to `Py<T>` and `PyObject` methods which really should be requiring the GIL.

Doing this higlighted a couple of other places across the codebase which could be tweaked:
- `PyTypeObject::type_object` now returns `&PyType` instead of `Py<PyType>` - should be faster as now no reference count increase needed.
- `PyTuple::slice` and `PyTuple::split_from` now return `&PyTuple`.

### DIFF
--- a/src/err.rs
+++ b/src/err.rs
@@ -1,12 +1,13 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
+use crate::gil::ensure_gil;
 use crate::panic::PanicException;
 use crate::type_object::PyTypeObject;
 use crate::types::PyType;
 use crate::{exceptions, ffi};
 use crate::{
-    AsPyPointer, FromPy, FromPyPointer, IntoPy, IntoPyPointer, Py, PyAny, PyObject, Python,
-    ToBorrowedObject, ToPyObject,
+    AsPyPointer, FromPy, FromPyPointer, IntoPy, IntoPyPointer, Py, PyAny, PyNativeType, PyObject,
+    Python, ToBorrowedObject, ToPyObject,
 };
 use libc::c_int;
 use std::ffi::CString;
@@ -88,11 +89,14 @@ impl PyErr {
         T: PyTypeObject,
         V: ToPyObject + 'static,
     {
-        let ty = T::type_object();
+        let gil = ensure_gil();
+        let py = unsafe { gil.python() };
+
+        let ty = T::type_object(py);
         assert_ne!(unsafe { ffi::PyExceptionClass_Check(ty.as_ptr()) }, 0);
 
         PyErr {
-            ptype: ty,
+            ptype: ty.into(),
             pvalue: PyErrValue::ToObject(Box::new(value)),
             ptraceback: None,
         }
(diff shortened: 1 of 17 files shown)


src/err.rs
```diff
@@ -565,7 +572,7 @@ impl_to_pyerr!(std::string::FromUtf16Error, exceptions::UnicodeDecodeError);
 impl_to_pyerr!(std::char::DecodeUtf16Error, exceptions::UnicodeDecodeError);
 impl_to_pyerr!(std::net::AddrParseError, exceptions::ValueError);
 
-pub fn panic_after_error() -> ! {
+pub fn panic_after_error(_py: Python) -> ! {
```
kngwyu, line 575: :+1:

src/gil.rs
```diff
@@ -370,6 +370,38 @@ extern "C" fn finalize() {
...
+pub(crate) fn acquire_if_needed() -> GILAcquisition {
+    if gil_is_acquired() {
+        GILAcquisition::AlreadyHeld
+    } else {
+        GILAcquisition::JustInTime(GILGuard::acquire())
+    }
+}
+
+/// Enum used internally which avoids acquiring the GIL where it's not necessary.
+pub(crate) enum GILAcquisition {
```
kngwyu, line 384: I like this abstraction since it helps to simplify the implementation  :+1: 
But I don't think we need naming each variant (AlreadyHeld` and `JustInTime`).
To make it simpler, how about  
```rust
pub(crate) struct EnsureGILHeld(Option<GILGuard>);
```
or so?

src/instance.rs
```diff
@@ -121,14 +121,14 @@ impl<T> Py<T> {
 
     /// Gets the reference count of the `ffi::PyObject` pointer.
     #[inline]
-    pub fn get_refcnt(&self) -> isize {
+    pub fn get_refcnt(&self, _py: Python) -> isize {
```
kngwyu, line 124: :+1:

src/gil.rs
davidhewitt: :+1: I think struct is nicer than enum here - I will update PR.

davidhewitt:
Updated to use `EnsureGIL` struct instead of enum 👍

kngwyu:
LGTM, thanks.

PR #974, opened 2020-06-13
Revert soft-duplicated for Py::new

kngwyu:
Since `PyCell` is registered to GILPool, `Py<T>::new` can be more efficient especially when the object is directly returned to Python.
E.g.,
```rust
#[pyfunction] 
fn new_class(py: Python) -> Py<MyClass> {
    Py::new(py, ...)
}
```

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -52,9 +52,8 @@ unsafe impl<T> Sync for Py<T> {}
 impl<T> Py<T> {
     /// Create a new instance `Py<T>`.
     ///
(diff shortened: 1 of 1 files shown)


davidhewitt:
That's a fair point :+1: . Perhaps we should update docs to recommend `Py::new` instead of `PyCell::new` in most cases?

kngwyu:
> Perhaps we should update docs to recommend Py::new instead of PyCell::new in most cases?

Yeah, but I'm not sure where they are

Issue #973, opened 2020-06-13
shared python object access across threads and lazy_static deadlocks


clegaard:
When wrapping Python libraries in Rust it may be useful to share a static instance of an object across threads. For example, python's datetime library has a one-time initialization of an API object. Using lazy_static for accessing the object can cause a deadlock as described below:
https://github.com/rust-lang-nursery/lazy-static.rs/issues/116

@davidhewitt mentioned that he has put some thought into general solutions for internal use.

kngwyu:
Honestly, now I have no good solution for this.
E.g., to avoid this kind of deadlock, rust-numpy locks the thread [after importing the module](https://github.com/PyO3/rust-numpy/blob/v0.9.0/src/npyffi/array.rs#L50), which I really don't like.

Possible solutions in user codes are:
- Do not use lock. E.g., using [once_cell::unsync](https://docs.rs/once_cell/1.4.0/once_cell/unsync/index.html)
- Do not import any module when locking the thread.

davidhewitt:
I'll try to sketch out a draft PR with my idea tomorrow. I think there can be a pretty elegant solution using the `Python` token...

davidhewitt:
See #975

clegaard:
Amazing, thank you. I will try to check the branch and see if it resolves my specific issue.

davidhewitt:
:+1: note that we could probably wrap it with a macro similar to `lazy_static` too!

clegaard:
Hi, I finally had the time to try it out. Using the OnceCell seem to do the trick, my tests no longer freeze when running in parallel :)

Being the rust noob that I tried to instantiate the object in a similar way to the datetime object:

``` Rust
static SLAVE_MANAGER_ONCE_CELL: OnceCell<PyObject> = OnceCell::new();

struct SlaveManagerApi {
}

static SLAVE_MANAGER : SlaveManagerApi = SlaveManagerApi {

};

impl Deref for SlaveManagerApi {

    type Target = PyObject;

    fn deref(&self) -> &'static PyObject {

        let py = unsafe {Python::assume_gil_acquired()};
        

        SLAVE_MANAGER_ONCE_CELL.get_or_init(py, || {
            let cls = || -> PyResult<PyObject> {
                let ctx: pyo3::PyObject = py
                    .import("pyfmu.fmi2.slaveContext")
                    .expect("Unable to import module declaring slave manager. Ensure that PyFMU is installed inside your current envrioment.")
[...]

PR #975, opened 2020-06-14
Add pyo3::once_cell::GILOnceCell

davidhewitt:
This is a draft idea of how to get something similar to `lazy_static` which avoids deadlocks related to the GIL.

It's based on `once_cell` API, but uses GIL to guarantee thread safety instead of its own locks.

I was able to prove it works by replacing both our datetime's `static mut` as well as `LazyHeapType`. I think `LazyStaticType` should be able to be replaced too, but I had some complications around GIL safety which I think #970 will solve - so I'll wait for that to merge first.

### DIFF
--- a/guide/src/faq.md
+++ b/guide/src/faq.md
@@ -1,0 +1,16 @@
+# Frequently Asked Questions / Troubleshooting
+
+## I'm experiencing deadlocks using PyO3 with lazy_static or once_cell!
+
+`lazy_static` and `once_cell::sync` both use locks to ensure that initialization is performed only by a single thread. Because the Python GIL is an additional lock this can lead to deadlocks in the following way:
+
+1. A thread (thread A) which has acquired the Python GIL starts initialization of a `lazy_static` value.
+2. The initialization code calls some Python API which temporarily releases the GIL e.g. `Python::import`.
+3. Another thread (thread B) acquires the Python GIL and attempts to access the same `lazy_static` value.
+4. Thread B is blocked, because it waits for `lazy_static`'s initialization to lock to release.
+5. Thread A is blocked, because it waits to re-aquire the GIL which thread B still holds.
+6. Deadlock.
+
+PyO3 provides a struct [`GILOnceCell`] which works equivalently to `OnceCell` but relies solely on the Python GIL for thread safety. This means it can be used in place of `lazy_static` or `once_cell` where you are experiencing the deadlock described above. See the documentation for [`GILOnceCell`] for an example how to use it.
+
+[`GILOnceCell`]: https://docs.rs/pyo3/latest/pyo3/once_cell/struct.GILOnceCell.html
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -401,10 +401,10 @@ fn impl_class(
             const FLAGS: usize = #(#flags)|* | #extended;
 
             #[inline]
-            fn type_object() -> &'static pyo3::ffi::PyTypeObject {
+            fn type_object_raw(py: pyo3::Python) -> &'static pyo3::ffi::PyTypeObject {
                 use pyo3::type_object::LazyStaticType;
                 static TYPE_OBJECT: LazyStaticType = LazyStaticType::new();
-                TYPE_OBJECT.get_or_init::<Self>()
+                TYPE_OBJECT.get_or_init::<Self>(py)
             }
         }
 
--- a/src/exceptions.rs
+++ b/src/exceptions.rs
@@ -90,28 +90,27 @@ macro_rules! import_exception_type_object {
     ($module: expr, $name: ident) => {
         unsafe impl $crate::type_object::PyTypeObject for $name {
             fn type_object(py: $crate::Python) -> &$crate::types::PyType {
-                use $crate::type_object::LazyHeapType;
-                static TYPE_OBJECT: LazyHeapType = LazyHeapType::new();
+                use $crate::once_cell::GILOnceCell;
+                use $crate::AsPyRef;
+                static TYPE_OBJECT: GILOnceCell<$crate::Py<$crate::types::PyType>> =
+                    GILOnceCell::new();
 
-                let ptr = TYPE_OBJECT.get_or_init(|py| {
-                    let imp = py
-                        .import(stringify!($module))
-                        .expect(concat!("Can not import module: ", stringify!($module)));
-                    let cls = imp.get(stringify!($name)).expect(concat!(
-                        "Can not load exception class: {}.{}",
-                        stringify!($module),
-                        ".",
-                        stringify!($name)
-                    ));
+                TYPE_OBJECT
+                    .get_or_init(py, || {
+                        let imp = py
+                            .import(stringify!($module))
+                            .expect(concat!("Can not import module: ", stringify!($module)));
+                        let cls = imp.get(stringify!($name)).expect(concat!(
+                            "Can not load exception class: {}.{}",
+                            stringify!($module),
+                            ".",
+                            stringify!($name)
+                        ));
 
-                    unsafe {
-                        std::ptr::NonNull::new_unchecked(
-                            $crate::IntoPyPointer::into_ptr(cls) as *mut _
-                        )
-                    }
-                });
-
-                unsafe { py.from_borrowed_ptr(ptr.as_ptr() as *mut $crate::ffi::PyObject) }
+                        cls.extract()
+                            .expect("Imported exception should be a type object")
+                    })
+                    .as_ref(py)
             }
         }
     };
@@ -174,19 +173,25 @@ macro_rules! create_exception_type_object {
     ($module: ident, $name: ident, $base: ty) => {
         unsafe impl $crate::type_object::PyTypeObject for $name {
             fn type_object(py: $crate::Python) -> &$crate::types::PyType {
-                use $crate::type_object::LazyHeapType;
-                static TYPE_OBJECT: LazyHeapType = LazyHeapType::new();
+                use $crate::once_cell::GILOnceCell;
+                use $crate::AsPyRef;
+                static TYPE_OBJECT: GILOnceCell<$crate::Py<$crate::types::PyType>> =
+                    GILOnceCell::new();
 
-                let ptr = TYPE_OBJECT.get_or_init(|py| {
-                    $crate::PyErr::new_type(
-                        py,
-                        concat!(stringify!($module), ".", stringify!($name)),
-                        Some(py.get_type::<$base>()),
-                        None,
-                    )
-                });
-
-                unsafe { py.from_borrowed_ptr(ptr.as_ptr() as *mut $crate::ffi::PyObject) }
+                TYPE_OBJECT
+                    .get_or_init(py, || unsafe {
+                        $crate::Py::from_owned_ptr(
+                            py,
+                            $crate::PyErr::new_type(
+                                py,
+                                concat!(stringify!($module), ".", stringify!($name)),
+                                Some(py.get_type::<$base>()),
+                                None,
+                            )
+                            .as_ptr() as *mut $crate::ffi::PyObject,
+                        )
+                    })
+                    .as_ref(py)
             }
         }
     };
--- a/src/once_cell.rs
+++ b/src/once_cell.rs
@@ -1,0 +1,100 @@
+use crate::Python;
+use std::cell::UnsafeCell;
+
+/// A write-once cell similar to [`once_cell::OnceCell`](https://docs.rs/once_cell/1.4.0/once_cell/).
+///
+/// Unlike `once_cell::sync` which blocks threads to achieve thread safety, this implementation
+/// uses the Python GIL to mediate concurrent access. This helps in cases where `once_sync` or
+/// `lazy_static`'s synchronization strategy can lead to deadlocks when interacting with the Python
+/// GIL. For an example, see [the FAQ section](https://pyo3.rs/master/faq.html) of the guide.
+///
+/// # Example
+///
+/// The following example shows how to use `GILOnceCell` to share a reference to a Python list
+/// between threads:
+///
+/// ```
+/// use pyo3::prelude::*;
+/// use pyo3::types::PyList;
+/// use pyo3::once_cell::GILOnceCell;
+///
+/// static LIST_CELL: GILOnceCell<Py<PyList>> = GILOnceCell::new();
+///
+/// pub fn get_shared_list(py: Python) -> &PyList {
+///     LIST_CELL
+///         .get_or_init(py, || PyList::empty(py).into())
+///         .as_ref(py)
+/// }
+/// # let gil = Python::acquire_gil();
+/// # let py = gil.python();
+/// # assert_eq!(get_shared_list(py).len(), 0 );
+/// ```
+pub struct GILOnceCell<T>(UnsafeCell<Option<T>>);
+
+// T: Send is needed for Sync because the thread which drops the GILOnceCell can be different
+// to the thread which fills it.
+unsafe impl<T: Send + Sync> Sync for GILOnceCell<T> {}
+unsafe impl<T: Send> Send for GILOnceCell<T> {}
+
+impl<T> GILOnceCell<T> {
+    /// Create a `GILOnceCell` which does not yet contain a value.
+    pub const fn new() -> Self {
+        Self(UnsafeCell::new(None))
+    }
+
+    /// Get a reference to the contained value, or `None` if the cell has not yet been written.
+    pub fn get(&self, _py: Python) -> Option<&T> {
+        // Safe because if the cell has not yet been written, None is returned.
+        unsafe { &*self.0.get() }.as_ref()
+    }
+
+    /// Get a reference to the contained value, initializing it if needed using the provided
+    /// closure.
+    ///
+    /// Note that:
+    ///  1) reentrant initialization can cause a stack overflow.
+    ///  2) if f() temporarily releases the GIL (e.g. by calling `Python::import`) then it is
+    ///     possible (and well-defined) that a second thread may also call get_or_init and begin
+    ///     calling `f()`. Even when this happens `GILOnceCell` guarantees that only **one** write
+    ///     to the cell ever occurs - other threads will simply discard the value they compute and
+    ///     return the result of the first complete computation.
+    pub fn get_or_init<F>(&self, py: Python, f: F) -> &T
+    where
+        F: FnOnce() -> T,
+    {
+        let inner = unsafe { &*self.0.get() }.as_ref();
+        if let Some(value) = inner {
+            return value;
+        }
+
+        // Note that f() could temporarily release the GIL, so it's possible that another thread
+        // writes to this GILOnceCell before f() finishes. That's fine; we'll just have to discard
+        // the value computed here and accept a bit of wasted computation.
+        let value = f();
+        let _ = self.set(py, value);
+
+        self.get(py).unwrap()
+    }
+
+    /// Get the contents of the cell mutably. This is only possible if the reference to the cell is
+    /// unique.
+    pub fn get_mut(&mut self) -> Option<&mut T> {
+        // Safe because we have &mut self
+        unsafe { &mut *self.0.get() }.as_mut()
+    }
+
+    /// Set the value in the cell.
+    ///
+    /// If the cell has already been written, `Err(value)` will be returned containing the new
+    /// value which was not written.
+    pub fn set(&self, _py: Python, value: T) -> Result<(), T> {
+        // Safe because GIL is held, so no other thread can be writing to this cell concurrently.
+        let inner = unsafe { &mut *self.0.get() };
+        if inner.is_some() {
+            return Err(value);
+        }
+
+        *inner = Some(value);
+        Ok(())
+    }
+}
--- a/src/type_object.rs
+++ b/src/type_object.rs
@@ -1,13 +1,13 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 //! Python type object information
 
+use crate::once_cell::GILOnceCell;
 use crate::pyclass::{initialize_type_object, PyClass};
 use crate::pyclass_init::PyObjectInit;
 use crate::types::{PyAny, PyType};
-use crate::{ffi, AsPyPointer, Python};
-use std::cell::UnsafeCell;
-use std::ptr::NonNull;
-use std::sync::atomic::{AtomicBool, Ordering};
+use crate::{ffi, AsPyPointer, PyNativeType, Python};
+use parking_lot::{const_mutex, Mutex};
+use std::thread::{self, ThreadId};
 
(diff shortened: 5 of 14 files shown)


src/exceptions.rs
```diff
@@ -90,28 +90,28 @@ macro_rules! import_exception_type_object {
...
-                        )
-                    }
-                });
-
-                unsafe { $crate::Py::from_borrowed_ptr(ptr.as_ptr() as *mut $crate::ffi::PyObject) }
+                use $crate::once_cell::OnceCell;
+                static TYPE_OBJECT: OnceCell<$crate::Py<$crate::types::PyType>> = OnceCell::new();
+
+                let gil = $crate::Python::acquire_gil();
+                let py = gil.python();
```
davidhewitt, line 97: Should use the new `ensure_gil()` API from #970 . (And same later in this file.)

src/once_cell.rs
```diff
@@ -0,0 +1,58 @@
...
+/// Unlike once_cell::sync which uses locks to achieve thread safety, this implementation uses
+/// the Python GIL to mediate concurrent access.
+pub struct OnceCell<T>(UnsafeCell<Option<T>>);
+
+// T: Send is needed for Sync because the thread which drops the OnceCell can be different
+// to the thread which fills it.
+unsafe impl<T: Send + Sync> Sync for OnceCell<T> {}
+unsafe impl<T: Send> Send for OnceCell<T> {}
+
+impl<T> OnceCell<T> {
```
davidhewitt, line 15: Need to add lots of docs / examples to all these methods.

davidhewitt:
I managed to use `OnceCell` for the implementation of `LazyStaticType` too.

This led to an interesting discovery. I think our original assessment about "recursive" class attributes, as discussed here, might be wrong: https://github.com/PyO3/pyo3/pull/905#discussion_r421334210

The assumption there was that `PyCell::new` does not use the type object. But it turns out that it does - `PyCell::new` eventually calls `T::alloc` which uses the type object to create a new instance. 

For "recursive" class attributes I think this shows they are unfortunately not sound, because I don't think we should be using the incomplete type object to create new instances.

As a result I now disable recursive class attributes in this PR.

cc @scalexm - as original implementor of class attributes, what do you think of this?

kngwyu:
I'm sorry I cannot afford to give a detailed review now. I'll give it later.

At a glance, this looks like a really nice refactoring for our internal implementation :+1: 
But I'm not sure this should be a public API or not and how to explain what this is to users :thinking: 

> This led to an interesting discovery. I think our original assessment about "recursive" class attributes, as discussed here, might be wrong: #905 (comment)

Sign... :sweat: sorry for my incorrect comment. But could you please leave it as is for now?
I think it's better if we can set `tp_dict` after initializing the class if it's possible, which could be out of this PR.

scalexm:
First of all, I’ve just read the current implementation (before this PR) of `LazyStaticType`. The public API was unsound, as one caller could set the `initialized` flag, then start calling `initialize_type_object` with a `&mut` ref to the cell interior, then while that call is still ongoing, an other caller can call `get_or_init`, see that `initialized` is already set to true and create a `&` ref to the cell interior, which incorrectly aliases with the former `&mut` reference.

In particular `get_or_init` **was not reentrant**. A correct implementation should have deadlocked or panicked, and we would have caught the « recursive class attribute » problem thanks to the tests I added specifically for that.

This PR now does the correct thing, in that it takes a lock token as an argument to `get_or_init`.

Now, according to what @davidhewitt found, it seems to me that a solution for class attributes needs to happen in two steps: for each type object, we’d have two distinct initialization procedures, each one encapsulated in some kind of `Once`:
* `initialize_type_object`, which does everything except initializing `tp_dict`, and which disallows reentrancy (for example with this PR, I believe it would be deadlocking) to be sure that we don’t introduce other bugs
* `initialize_tp_dict`, which initializes the `tp_dict`, and which resolves reentrancy cycles by simply not calling the function again

I know that @programmerjake expressed concern about changing the `tp_dict` after calling `PyType_ready`, but I think it’s perfectly allowed: see for example https://github.com/python/cpython/blob/25f38d7044a3a47465edd851c4e04f337b2c4b9b/Python/sysmodule.c#L2757

scalexm:
To clarify, I was thinking about something like this:
```rust
struct LazyStaticType {
    value: OnceCell<Box<ffi::PyTypeObject>>,

    // Does not need to be atomic, an `UnsafeCell` would suffice in reality,
    // but it's less convenient.
    init_dict: AtomicBool,
}

impl LazyStaticType {
    ...
    
    pub fn get_or_init<T: PyClass>(&self, py: Python) -> &ffi::PyTypeObject {
        let type_object = self.value
            .get_or_init(py, || {
                let mut type_object = Box::new(ffi::PyTypeObject_INIT);
                // Reentrant call inside `initialize_type_object` will loop indefinitely.
                // We can also make it panic if wanted, as in the PR.
                initialize_type_object::<T>(py, T::MODULE, type_object.as_mut()).unwrap_or_else(
                    |e| {
                        e.print(py);
                        panic!("An error occurred while initializing class {}", T::NAME)

                    },
                );
                type_object
            })
            .as_ref()
        
        // Note: we're still holding the GIL lock here.
        if init_dict.load(Ordering::Relaxed) {
            // Re-entrant call: just return the type object even if `tp_dict` is
            // not yet filled. Since `ffi::pyTypeObject` is not `Send`, there are
            // also no risks of concurrent read/writes to the `tp_dict`.
            return type_object;
        }
        init_dict.store(true, Ordering::Relaxed);
        // Maybe add a RAII guard here for cleanup in case of panic.
        init_tp_dict(type_object.tp_dict);
        type_object
    }
    
    ...
}

fn init_tp_dict(tp_dict: *mut PyObject) {
    // Only use FFI functions to manipulate the dict here.
    ...
}
```

src/once_cell.rs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        let inner = unsafe { &*self.0.get() }.as_ref();
+        if let Some(value) = inner {
+            return value;
+        }
+
+        // Note that f() could release the GIL, so it's possible that another thread writes to this
+        // OnceCell before f() finishes. That's fine; we'll just have to discard the value
+        // computed here and accept a bit of wasted computation.
+        let value = f();
```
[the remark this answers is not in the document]
davidhewitt: Unfortunately `f()` can call things like `py.allow_threads()` to temporarily release the GIL, or it might call Python APIs which may in turn release the GIL.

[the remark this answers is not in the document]
davidhewitt: I'll add the word "temporarily" to the comment. 👍

src/type_object.rs
```diff
@@ -132,78 +136,65 @@ where
...
+        self.value
+            .get_or_init(py, || {
+                {
+                    // Code evaluated at class init time, such as class attributes, might lead to
+                    // recursive initalization of the type object if the class attribute is the same
+                    // type as the class.
+                    //
+                    // That could lead to all sorts of unsafety such as using incomplete type objects
+                    // to initialize class instances, so recursive initialization is prevented.
+                    let thread_not_already_initializing = self
```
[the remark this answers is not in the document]
davidhewitt: I think because of the above issue that `f()` can release the GIL, we can't necessarily prevent multiple threads from attempting to initialize the type object. What the `OnceCell` implementation above can guarantee is that all values returned from `get_or_init` are the same reference - even if `init` runs multiple times...

Trying to work around this by blocking all threads except the first with a lock is exactly how we get to the deadlock scenario in `lazy_static` 😢

davidhewitt:
> I know that @programmerjake expressed concern about changing the tp_dict after calling PyType_ready, but I think it’s perfectly allowed: see for example https://github.com/python/cpython/blob/25f38d7044a3a47465edd851c4e04f337b2c4b9b/Python/sysmodule.c#L2757

I'm not sure exactly how to interpret the docs:
- See the big red warning box here: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_dict
- But also it says just above that box: 
   > Once PyType_Ready() has initialized the type, extra attributes for the type may be added to this dictionary only if they don’t correspond to overloaded operations (like __add__()).`

src/type_object.rs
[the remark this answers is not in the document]
davidhewitt: > Also, without this check, a recursive call will just overflow the stack, so maybe we don’t need the check at all, unless you prefer it for debugging purposes.

I'm torn. I don't expect the check to cause much overhead (will usually be a `HashMap` with just one element and never read.)

The stack overflow is a little harder to debug, but I guess it's fine :)

davidhewitt:
> But I'm not sure this should be a public API and how to explain what this is to users 🤔

It gives users a way to avoid the `lazy_static` deadlock seen in #973, so it would be nice if it is a public API.

src/once_cell.rs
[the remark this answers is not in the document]
scalexm: Oh I see, I didn’t know we could do that. Thanks for the explanation!

src/type_object.rs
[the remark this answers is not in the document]
scalexm: Yes it’s not much a matter of overhead but rather of code simplification :)

also about the boolean flag, in that case we control the `f` that is being called (`initialize_type_object`) so we can assume it won’t release the GIL.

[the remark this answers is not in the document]
davidhewitt: We almost control the `f` being called but I don't know if we can guarantee it won't release the GIL. User could write a custom implementation of just one part (e.g. class attribute initialization) which might easily break this assumption in a very subtle way...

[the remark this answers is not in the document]
scalexm: As long as we don’t give the user access to the `Python` token (which we don’t in class attributes, exactly because we were afraid to do so!), we’re safe. We need to ensure this anyway for my simple two steps solution to work.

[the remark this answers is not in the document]
davidhewitt: > As long as we don’t give the user access to the Python token (which we don’t in class attributes, exactly because we were afraid to do so!), we’re safe.

I believe `Python::acquire_gil().python().allow_threads(|| { ... })` will temporarily release the GIL, even in class attributes. So I _think_ we don't have to give class attribute user the `Python` token for stuff to happen!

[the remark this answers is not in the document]
scalexm: Ok. Thanks for the clarification. Then I understand that the discussion we had on #905 of whether or not to allow passing a `Python` parameter to class attr methods was a bit pointless :)

Also the solution I proposed cannot work as is, as I need to make the assumption that the GIL is held during the whole tp_dict initialization.

Do you know whether we have some kind of guarantees about the tp_dict synchronization? For example, do we guarantee that whenever we read or write to it (either directly, or indirectly through FFI functions), we hold a GIL?

Also, I’m in favor of leaving your check for recursive initialization as is. We can revisit when we sort out this recursive class attribute thing.

scalexm:
I’m in favor of merging this PR as it is (modulo review from @kngwyu), as it fixes an important soundness issue in `LazyStaticType::get_or_init`.

It’s ok to disable recursive class attributes for now, as it seems clear that a solution to this problem is out of scope for this PR.

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -387,10 +387,10 @@ fn impl_class(
             const FLAGS: usize = #(#flags)|* | #extended;
 
             #[inline]
-            fn type_object() -> &'static pyo3::ffi::PyTypeObject {
+            fn type_object_raw(py: pyo3::Python) -> &'static pyo3::ffi::PyTypeObject {
```
kngwyu, line 390: Why `raw`?
Since it returns an FFI type?

src/type_object.rs
```diff
@@ -132,78 +136,65 @@ where
...
-// to allow sharing on the Rust side too.
-unsafe impl Sync for LazyHeapType {}
-
 /// Lazy type object for PyClass
 #[doc(hidden)]
 pub struct LazyStaticType {
-    value: UnsafeCell<ffi::PyTypeObject>,
-    initialized: AtomicBool,
+    // Boxed because Python expects the type object to have a stable address.
+    value: OnceCell<Box<ffi::PyTypeObject>>,
```
kngwyu, line 147: Why can't we use `ffi::PyTypeObject` here?
I think `&value` is a consistent address since `LazyStaticType` is only used as a static variable.

src/type_object.rs
```diff
@@ -132,78 +136,65 @@ where
...
+    // Threads which have begun initialization. Used for reentrant initialization detection.
+    initializing_threads: Mutex<Option<HashSet<ThreadId>>>,
```
kngwyu, line 149: Just a nit but I think `Mutex<Vec<ThreadId>>` is sufficient.

kngwyu:
Thank you @davidhewitt and @scalexm for the discussion and pointing out the unsoundness! :+1: 

> It’s ok to disable recursive class attributes for now, as it seems clear that a solution to this problem is out of scope for this PR. 

Agreed. I want to adopt @scalexm's 'two-stage' initialization strategy if it's possible, but in this PR I want to concentrate on how `OnceCell` should be.

src/type_object.rs
kngwyu: Ah but I think `AtomicBool` is sufficient as @scalexm pointed out in the above thread because we don't release the GIL in `initialize_type_object`. Or I missed something?

src/type_object.rs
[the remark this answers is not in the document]
kngwyu: I'm feeling it difficult to understand why `f()` matters here. Now `f()` is this closure and we don't call any user-defined function in this closure, right?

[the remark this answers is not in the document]
scalexm: We initialize class attributes inside `initialize_type_object`, and the initialization calls into user-defined code.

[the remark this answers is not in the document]
kngwyu: > We initialize class attributes inside initialize_type_object, and the initialization calls into user-defined code.

Ah, I understand, thanks.
But it looks the only candidate is `#[classattr]` :thinking: ...

src/type_object.rs
kngwyu: I noticed that `#[classattr]` functions can be called in `intialize_type_object`. Please ignore the above comment.

davidhewitt: Yes I considered `Mutex<Vec<ThreadId>>` but was a little put off by the O(N) lookup. Usually will be just one thread at most (or very) few so `Vec` should be fine 👍

src/type_object.rs
davidhewitt: `OnceCell::get_or_init` moves the result. I'm not sure that it's safe to do this after calling `PyType_Ready`, which is why I boxed it...

pyo3-derive-backend/src/pyclass.rs
davidhewitt: Yeah, having `TypeObject::type_object` and `PyTypeInfo::type_object` have the same name is a little annoying, which is why I wanted to rename this one. I'm happy to revert this or change the name to another new one as you prefer?

davidhewitt:
> In addition, can the name OnceCell be confusing?

I chose the name `OnceCell` because it's a write-once cell exactly like the ones found in the `once_cell` crate.

But I would be very happy to use a different name if anyone can find one. A nice thing about calling it `OnceCell` is we can document this API as being a drop-in replacement for `once_cell::sync::OnceCell` which won't deadlock due to the GIL :)

src/type_object.rs
kngwyu: > I'm not sure that it's safe to do this after calling PyType_Ready, which is why I boxed it...

Yeah, it's not safe. But then why we need to use `OnceCell` here? Just for simplification?

pyo3-derive-backend/src/pyclass.rs
kngwyu: :+1:

kngwyu:
What I meant is about using the same name as the famous `OnceCell`.
Our `OnceCell` and their one can be used at the same time. If one initializes their OnceCell before getting GIL, it works without deadlock.

davidhewitt:
> If one initializes their OnceCell before getting GIL, it works without deadlock.

Not sure exactly what you mean by this? I can create a deadlock with `once_cell::sync` and the GIL pretty easily:

<details><summary>Click to expand</summary>

```
use pyo3::prelude::*;
use once_cell::sync::OnceCell;
use std::thread::sleep;

static OC: OnceCell<()> = OnceCell::new();
static DURATION: std::time::Duration = std::time::Duration::from_secs(1);

#[test]
fn test1() {
    let py = gil.python();
    OC.get_or_init(|| py.allow_threads(|| sleep(DURATION)));
}

#[test]
fn test2() {
    let py = gil.python();
    OC.get_or_init(|| py.allow_threads(|| sleep(DURATION)));
}
```
</details>

> What I meant is about using the same name as the famous OnceCell.

That's fair. What about `GILOnceCell` ? I am happy to take any suggestions too!

src/type_object.rs
davidhewitt: I wanted to use `OnceCell` because then all our internal statics (datetime API, imported type objects, created type objects) are initialised the same way, using `OnceCell`. Changing to use `OnceCell` also helped us discover the unsafety in previous implementation's interaction with class attributes, so I think using `OnceCell` here is proving a good idea.

kngwyu:
I like `GILOnceCell` or `GILCell`.

davidhewitt:
I think let's have `GILOnceCell` and maybe later we can make `GILLazy`. I think those names have nice symmetry with the well-known `OnceCell` and `Lazy` (just stick `GIL` on the front :smile:)

davidhewitt:
(I was thinking maybe in the future `GILCell` could be something which is a bit more like `RefCell` but can be shared between threads - kind of like `PyCell` we already have but for all types, not just for `PyClass`)

davidhewitt:
Renamed to `GILOnceCell`, added documentation and CHANGELOG, and rebased on master. This is now ready for final review / consideration for merging.

Let's create a follow-up issue to solve self-typed class attributes?

src/type_object.rs
kngwyu: Hmm I still don't like this redundant `Box, though I'm not strongly against it.

davidhewitt: How about I change it to `Pin<Box<ffi::PyTypeObject>>`? That way it clearly shows that it should have a fixed address.

kngwyu: My concern is about the allocation cost.
`ffi::PyTypeObject` is (roughly) 320bytes.

kngwyu: `Pin` looks nice, though.

scalexm: `Pin<P>` is not really about having a stable address, it’s about not being able to move out of the place pointed by `P`. `Box` already guarantees to have a stable address, I think `Pin` only adds confusion.

davidhewitt: Is thta not kinda the idea though - once `PyType_Ready` is called, the type object should not be moved in memory? (Hence why I thought `Pin` is applicable here.)

scalexm: You can still move in or out of the logical `ffi::PyTypeObject` without problem: it’s `Copy` and `Unpin` anyway (it basically amounts to overwriting all the fields with a `memcpy`).

It’s the place in memory where the `ffi::PyTypeObject` is allocated that should not move.

davidhewitt: That's a fair point. Ok I'm convinced adding `Pin` doesn't help here.

@kngwyu I don't think it's possible to remove the `Box` without adding some `unsafe` inplace initialization trick to `GILOnceCell`- and I'm not sure even then I can make it sound.

With regards to the allocation concern, I think this Python code will cause the same allocation cost anyway:

```
class Foo:
    pass
```

kngwyu: > I'm not sure even then I can make it sound.

What we need is just having `GILOnceCell<()>` and `UnsafeCell<ffi::PyTypeObject>`, and then initializing the type object in `get_or_init`.

> With regards to the allocation concern, I think this Python code will cause the same allocation cost anyway:

Yeah, I think I can be *too nervous* about allocation. This is why I said ' I'm not strongly against it'.

davidhewitt: > What we need is just having `GILOnceCell<()>` and `UnsafeCell<ffi::PyTypeObject>`, and then initializing the type object in `get_or_init`.

Hmm I guess that can work. I think we would need to be careful to ensure thread safety and no deadlocks.

As it's an implementation detail, maybe we should focus on this in a new PR (perhaps the same one that also solves recursive class attributes problem)?

scalexm: @kngwyu’s idea is doable once we’re sure we don’t release the GIL during the initialization of the type object. So I agree with @davidhewitt that is linked to the class attributes problem (the only part that calls into user code).

kngwyu: @davidhewitt @scalexm 
Thank you for your comments.

> As it's an implementation detail, maybe we should focus on this in a new PR (perhaps the same one that also solves recursive class attributes problem)?

Agreed :+1:

src/once_cell.rs
```diff
@@ -0,0 +1,92 @@
+use crate::Python;
+use std::cell::UnsafeCell;
+
+/// A write-once cell similar to `once_cell::OnceCell`.
```
kngwyu, line 4: We might need to link https://github.com/matklad/once_cell or crates.io

src/once_cell.rs
```diff
@@ -0,0 +1,92 @@
...
+///
+/// Unlike `once_cell::sync` which uses locks to achieve thread safety, this implementation uses
```
kngwyu, line 6: Just a nit, but it looks like that the current `OnceCell` implementation does not use `Mutex`. It adopts a lock-free approach the same as the `std::sync::Once`.

src/once_cell.rs
```diff
@@ -0,0 +1,92 @@
...
+/// the Python GIL to mediate concurrent access.
```
kngwyu, line 7: Let's make it clear why we need GIL(i.e., the interaction between GIL and another mutex can cause a deadlock).
Or it is also good to link the FAQ section in the guide.

src/once_cell.rs
```diff
@@ -0,0 +1,92 @@
...
+///
+/// # Example
+///
+/// The following example shows how to use `GILOnceCell` to share a reference to a Python list
+/// between threads:
+///
+/// ```
+/// use pyo3::prelude::*;
```
kngwyu, line 15: This example is nice :+1:

guide/src/faq.md
```diff
@@ -0,0 +1,16 @@
+# Frequently Asked Questions / Troubleshooting
```
kngwyu, line 1: Nice :+1:

src/once_cell.rs
davidhewitt: Ah good catch - I'd looked at the `parking_lot` implementation of `OnceCell` - which does use a Mutex, so it depends on the `once_cell` feature being used :).

I'll reword this

davidhewitt:
Rebased on master and reworded the docs as suggested - thanks for the feedback!

kngwyu:
LGTM, thanks!

kngwyu:
Opened #982.

PR #986, opened 2020-06-20
Do not use PyObject_RichCompareBool to detect invalid comparison

### LINKED ISSUE #985: nan == nan is True when other is the same object as self
## 🐛 Bug Reports

### 🌍 Environment

 - Your operating system and version: MacOS Mojave 10.14.6
 - Your python version: 3.7.4
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: virtualenv
 - Your Rust version (`rustc --version`): 1.46.0-nightly (feb3536eb 2020-06-09)
 - Your PyO3 version: 0.10
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: Yes

### 💥 Reproducing

```rust
use pyo3::prelude::*;

fn main() {
    let py = gil.python();
    let a = py.eval("float('nan')", None, None).unwrap();
    // nan == nan must not be Equal, but this test fails.
    assert_ne!(a.compare(a).unwrap(), std::cmp::Ordering::Equal);
}
```

This will fail with the below command

```
$ cargo run
```
```
thread 'main' panicked at 'assertion failed: `(left != right)`
  left: `Equal`,
 right: `Equal`', src/main.rs:8:5
```

Here is the repo to reproduce: https://github.com/tamuhey/pyo3_nan_bug

kngwyu:
Fixes #985

### DIFF
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -5,7 +5,7 @@ use crate::conversion::{
 use crate::err::{PyDowncastError, PyErr, PyResult};
 use crate::exceptions::TypeError;
 use crate::types::{PyDict, PyIterator, PyList, PyString, PyTuple, PyType};
-use crate::{err, ffi, Py, PyNativeType, Python};
+use crate::{err, ffi, Py, PyNativeType, PyObject, Python};
 use libc::c_int;
 use std::cell::UnsafeCell;
 use std::cmp::Ordering;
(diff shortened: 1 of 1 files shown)


src/types/any.rs
```diff
@@ -146,36 +146,36 @@ impl PyAny {
...
-        ) -> PyResult<Ordering> {
-            let result = ffi::PyObject_RichCompareBool(a, b, ffi::Py_EQ);
-            if result == 1 {
-                return Ok(Ordering::Equal);
-            } else if result < 0 {
+            op: c_int,
+        ) -> PyResult<bool> {
+            let result = ffi::PyObject_RichCompare(a, b, op);
+            if result.is_null() {
                 return Err(PyErr::fetch(py));
```
tamuhey, line 158: is it okay if you don't call `ffi::PyDECREF(result)` here?

kngwyu: Oh, good catch :+1: , thank you.
I actually forgot to call it.

src/types/any.rs
```diff
@@ -146,36 +146,25 @@ impl PyAny {
...
+        };
+        other.with_borrowed_ptr(py, |other| {
+            if do_compare(other, ffi::Py_EQ)? {
                 return Ok(Ordering::Equal);
-            } else if result < 0 {
-                return Err(PyErr::fetch(py));
             }
-            let result = ffi::PyObject_RichCompareBool(a, b, ffi::Py_LT);
-            if result == 1 {
+            if do_compare(other, ffi::Py_LT)? {
```
davidhewitt, line 159: Maybe `else if do_compare(...` ?

src/types/any.rs
```diff
@@ -146,36 +146,25 @@ impl PyAny {
...
                 return Ok(Ordering::Less);
-            } else if result < 0 {
-                return Err(PyErr::fetch(py));
             }
-            let result = ffi::PyObject_RichCompareBool(a, b, ffi::Py_GT);
-            if result == 1 {
+            if do_compare(other, ffi::Py_GT)? {
```
davidhewitt, line 162: Ditto `else if` ?

src/types/any.rs
kngwyu: :+1:

src/types/any.rs
```diff
@@ -146,36 +146,24 @@ impl PyAny {
...
-            let result = ffi::PyObject_RichCompareBool(a, b, ffi::Py_GT);
-            if result == 1 {
+            } else if do_compare(other, ffi::Py_GT)? {
                 return Ok(Ordering::Greater);
-            } else if result < 0 {
-                return Err(PyErr::fetch(py));
             }
             Err(TypeError::py_err(
                 "PyAny::compare(): All comparisons returned false",
             ))
```
davidhewitt, line 166: I just realised we could change this to:

```suggestion
            } else {
                Err(TypeError::py_err(
                    "PyAny::compare(): All comparisons returned false",
                ))
            }
```

Then we could also remove the `return` keyword from all the other branches.

kngwyu: :+1:

pganssle:
I think this is the right thing to do in the general case, but I suspect that there's a reasonable speed-up associated with using `PyObject_RichCompareBool`, since _most_ objects actually _are_ equal when the object identity is the same.

I have not looked at this issue in any detail, and I haven't tried to do any benchmarks to see if this change even really had a major speed impact, but I'm wondering how much work it would be to either special case certain types where it is known that pointer comparison is valid, like `int`, `str`, `list`, `tuple`, `datetime.datetime`, etc, or even something more elaborate using traits (or an extra field in `PyAny` indicating whether pointer comparison is "safe" for a given ).

I would be curious to see how this change affected the speed of comparison of, say, two identical tuples or lists, since for container types this optimization is already made in `__eq__`:

```python
>>> a = [float('nan')]
>>> b = [float('nan')]
>>> a == b
False
>>> a == a
True
>>> b == b
True
```

Another possible performance regression would be two lists that are not the same object but consistent of a large number of elements all of which *are* the same object, like so:

```python
>>> OBJECT = "bbbbb"
>>> a = [OBJECT] * 10000
>>> b = [OBJECT] * 10000
>>> a == b
```

PR #989, opened 2020-06-21
Bump version to 0.11.0

kngwyu:
The first version that can be used with stable Rust.
- [x] Check the compatibility with `rust-numpy`

### DIFF
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -4,7 +4,72 @@ This guide can help you upgrade code through breaking changes from one PyO3 vers
 For a detailed list of all changes, see [CHANGELOG.md](https://github.com/PyO3/pyo3/blob/master/CHANGELOG.md)
 
 ## from 0.10.* to 0.11
-Now PyO3 supports stable Rust toolchain. The minimum required version is 1.39.0.
+
+### Stable Rust
+PyO3 now supports the stable Rust toolchain. The minimum required version is 1.39.0.
+
+### `#[pyclass]` structs must now be `Send`
+Because `#[pyclass]` structs can be sent between threads by the Python interpreter, they must implement
+`Send` to guarantee thread safety. This bound was added in PyO3 `0.11.0`.
+
+This may "break" some code which previously was accepted, even though it was unsound. To resolve this,
+consider using types like `Arc` instead of `Rc`, `Mutex` instead of `RefCell`, and add `Send` to any
+boxed closures stored inside the `#[pyclass]`.
+
+Before:
+```rust,compile_fail
+use pyo3::prelude::*;
+use std::rc::Rc;
+use std::cell::RefCell;
+
+#[pyclass]
+struct NotThreadSafe {
+    shared_bools: Rc<RefCell<Vec<bool>>>,
+    closure: Box<Fn()>
+}
+```
+
+After:
+```rust
+use pyo3::prelude::*;
+use std::sync::{Arc, Mutex};
+
+#[pyclass]
+struct ThreadSafe {
+    shared_bools: Arc<Mutex<Vec<bool>>>,
+    closure: Box<Fn() + Send>
+}
+```
+
+Or in situations where you cannot change your `#[pyclass]` to automatically implement `Send`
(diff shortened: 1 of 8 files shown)


davidhewitt:
Shall we add a migration guide? I think we can mention stable Rust as well as `PyClass: Send`, maybe some other things too?

kngwyu:
> I think we can mention stable Rust as well as PyClass: Send

Sounds good :+1: 
You can push changes directly into this branch.

davidhewitt:
Cool I'll try to write this early evening today. Tomorrow morning at the latest.

davidhewitt:
Pushed a migration guide. I think it would be great to also merge #990 and @scalexm's recursive class attribute fix as part of `0.11`.

guide/src/migration.md
```diff
@@ -3,6 +3,69 @@
...
+## from 0.10.* to 0.11
+
+### Stable Rust
+PyO3 now supports the stable Rust toolchain. The minimum required version is 1.39.0.
+
+### `#[pyclass]` structs must now be `Send`
+Because `#[pyclass]` structs can be sent between threads by the Python interpreter, they must implement
+`Send` to guarantee thread safety. This bound was added in PyO3 `0.11.0`.
+
+This may "break" some code which previously was accepted, even though it was unsound. To resolve this,
```
kngwyu, line 15: We should note that 'if you are sure that your class is not sent to another thread, you can use `unsafe impl Send for YourClass {}`' first. Actually it's rare to send pyclass to another thread, and I saw some `#[pyclass]`es that have a raw pointer internally and need `unsafe impl Send` when investigating the effect of this change.

> This may "break" some code which previously was accepted, even though it was unsound.

Also, we have to note that it's *theoretically* unsound, but not problematic when the class is not actually sent to another thread by:
- Python's threading module
- `py.allow_threads` and `Py`

birkenfeld: > We should note that 'if you are sure that your class is not sent to another thread, you can use `unsafe impl Send for YourClass {}`' first.

I don't think that's good advice.  It undermines protections on the Rust side, too.

> Also, we have to note that it's theoretically unsound, but not problematic when the class is not actually sent to another thread by:

I don't know about your experience, but I use the threading module quite a lot in Python, and objects are accessed freely from any thread there.

kngwyu: Thank you for the feedback. 

> I don't know about your experience, but I use the threading module quite a lot in Python, and objects are accessed freely from any thread there.

Maybe it's because I use Python mainly for machine learning research, where CPU/GPU bound is dominant.
Then I would rather take back my words... but I still think we should note that `unsafe impl Sync` can be the easiest fix. Maybe it's also helpful to link [`std::marker::Send`](https://doc.rust-lang.org/std/marker/trait.Send.html) or [nomicon](https://doc.rust-lang.org/nomicon/send-and-sync.html).

davidhewitt: I'm not keen on suggesting `unsafe impl Send` or `unsafe impl Sync` - to me that feels like the kind of hack where "I know that I'm not using threads now, so I can skip writing safe code". Later on this hack is forgotten, the code gets edited to include threading and suddenly very strange bugs / crashes / undefined behaviour.

kngwyu:
I added a sentence about `unsafe impl Send`. Please check it.

guide/src/migration.md
```diff
@@ -3,6 +3,74 @@
...
+
+#[pyclass]
+struct ThreadSafe {
+    shared_bools: Arc<Mutex<Vec<bool>>>,
+    closure: Box<Fn() + Send>
+}
+```
+
+Or if you feel it hard to make your `#[pyclass]` struct automatically sendable
+(e.g., in case it has a raw pointer), you can use `unsafe impl Send`.
```
davidhewitt, line 45:
```suggestion
Or in situations where you cannot change your `#[pyclass]` to automatically implement `Send`
(e.g., when it contains a raw pointer), you can use `unsafe impl Send`.
```

kngwyu:
Thanks!

PR #990, opened 2020-06-21
Use subclass correctly in tp_new

### LINKED ISSUE #947: Subclassing pyclass from python - new returns wrong type
_Originally posted by @thesketh in https://github.com/PyO3/pyo3/issues/945#issuecomment-635465330_

> I am using `#[pyclass(subclass)]`, but `ParentClass.__new__` seems to ignore the first argument and always returns an instance of the parent class instead. The only thing it seems I can do with subclassing is create new constructors

This needs investigating, sounds like a possible bug.

kngwyu:
Fixes #947.
As [documented](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_new), we have to call `subtype->tp_alloc(subtype, nitems)` in tp_new.
In this PR, I also renamed `PyClassAlloc::alloc` to `new`, since it does not actually override `tp_alloc`.


kngwyu:
image

This segfault means that our `tp_dealloc` does something wrong when it is called for subtype :thinking:

CHANGELOG.md
```diff
@@ -25,13 +25,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - `#[pyproto]` is re-implemented without specialization. [#961](https://github.com/PyO3/pyo3/pull/961)
+- `PyClassAlloc::alloc` is renamed to `PyClassAlloc::new`. [#990](https://github.com/PyO3/pyo3/pull/990)
 
 ### Removed
 - Remove `ManagedPyRef` (unused, and needs specialization) [#930](https://github.com/PyO3/pyo3/pull/930)
 - Disable `#[classattr]` where the class attribute is the same type as the class. (This may be re-enabled in the future; the previous implemenation was unsound.) [#975](https://github.com/PyO3/pyo3/pull/975)
 
 ### Fixed
 - Fix passing explicit `None` to `Option<T>` argument `#[pyfunction]` with a default value. [#936](https://github.com/PyO3/pyo3/pull/936)
+- Fix `PyClass.__new__`'s behavior when inherited by a Python class. [#990](https://github.com/PyO3/pyo3/pull/990)
```
davidhewitt, line 36:
```suggestion
- Fix `PyClass.__new__` not respecting subclasses when inherited by a Python class. [#990](https://github.com/PyO3/pyo3/pull/990)
```

examples/rustapi_module/tests/test_subclassing.py
```diff
@@ -6,10 +6,11 @@
...
-    pass
+    def __str__(self):
+        return "Subclass"
 
 
 def test_subclassing():
     if not PYPY:
         a = SomeSubClass()
-        _b = str(a) + repr(a)
+        assert str(a) == "Subclass"
```
davidhewitt, line 16: Maybe also check `type(a)` ?

src/pyclass_init.rs
```diff
@@ -114,14 +114,27 @@ impl<T: PyClass> PyClassInitializer<T> {
...
-        let cell = PyCell::internal_new(py)?;
+        self.create_cell_from_subtype(py, T::type_object_raw(py) as *const _ as _)
+    }
+
+    /// Create a new PyCell and initialize it given a typeobject `subtype`.
+    /// Called by our `tp_new` generated by the `#[new]` attribute.
+    pub unsafe fn create_cell_from_subtype(
+        self,
+        py: Python,
+        subtype: *mut crate::ffi::PyTypeObject,
```
davidhewitt, line 131:
```suggestion
    /// Create a new PyCell and initialize it given a typeobject `subtype`.
    /// Called by our `tp_new` generated by the `#[new]` attribute.
    ///
    /// # Safety
    /// `cls` must be a subclass of T.
    pub unsafe fn create_cell_with_type(
        self,
        py: Python,
        cls: *mut crate::ffi::PyTypeObject,
```

src/pyclass_init.rs
```diff
@@ -114,14 +114,27 @@ impl<T: PyClass> PyClassInitializer<T> {
         PyClassInitializer::new(subclass_value, self)
     }
 
-    // Create a new PyCell + initialize it
-    #[doc(hidden)]
-    pub unsafe fn create_cell(self, py: Python) -> PyResult<*mut PyCell<T>>
+    // Create a new PyCell and initialize it.
+    pub(crate) unsafe fn create_cell(self, py: Python) -> PyResult<*mut PyCell<T>>
```
davidhewitt, line 118: This function doesn't actually look unsafe to me? From the signature, worst case seems to be memory leak which is sound.

```suggestion
    // Create a new PyCell and initialize it.
    pub(crate) fn create_cell(self, py: Python) -> PyResult<*mut PyCell<T>>
```

pyo3-derive-backend/src/pymethod.rs
```diff
@@ -193,7 +193,7 @@ pub fn impl_wrap_new(cls: &syn::Type, spec: &FnSpec<'_>) -> TokenStream {
     quote! {
         #[allow(unused_mut)]
         unsafe extern "C" fn __wrap(
-            _cls: *mut pyo3::ffi::PyTypeObject,
+            subcls: *mut pyo3::ffi::PyTypeObject,
```
davidhewitt, line 196: I think it's better to keep the `cls` name convention maybe.

```suggestion
            cls: *mut pyo3::ffi::PyTypeObject,
```

kngwyu: image
Maybe `subtype` is better?

src/pyclass_init.rs
kngwyu: :+1:

src/pyclass_init.rs
kngwyu: :+1:

kngwyu:
Thank you for the reviews :+1:

pyo3-derive-backend/src/pymethod.rs
davidhewitt: Ah yes good find 👍 I think subtype is probably better than `cls` here and also for `create_cell_with_type` in that case!

PR #992, opened 2020-06-22
Expose some APIs for rust-numpy

kngwyu:
`EnsureGIL` is not necessary but I want to use it.


src/lib.rs
```diff
@@ -195,6 +195,12 @@ mod python;
 pub mod type_object;
 pub mod types;
 
+/// Internal APIs exposed for rust-numpy
+#[doc(hidden)]
+pub mod internal_apis {
```
davidhewitt, line 200: This name is fine, though not one I think I've seen before. Some alternatives which I've seen in other include `imp` (which `once_cell` uses) and `r#impl` (which `inventory` uses) if you like either of those. Maybe also `_impl` or `impl_` could work if you don't like `r#`.

I also wonder if we should expose `derive_utils` under this mod?

kngwyu: I don't like `imp` or `impl` since they're not actually `implementation` itself. They are just utilties for internal implementations. So `internal_utils` is better?

> I also wonder if we should expose derive_utils under this mod?

Hmm :thinking:, I think it's not bad but have no strong motivation to do so.

davidhewitt: That's fair. Personally I think of anything that's not the "stable" public API as implementation. Let's keep as is 👍 

> Hmm 🤔, I think it's not bad but have no strong motivation to do so.

Maybe I'll open a PR sometime if I have a quiet moment and want to do some tidying up 😄

PR #994, opened 2020-06-22
`LazyStaticType::get_or_init` returns an `*mut` instead of a `&` ref

scalexm:
Previously, we were returning a `&ffi::PyTypeObject`. However, we were often casting this immutable reference to a `*mut ffi::PyTypeObject` and then passing that pointer to FFI functions.

But we don't know whether the FFI functions mutate the `PyTypeObject` behind our back or not, and if they do, I believe that this is Undefined Behavior because the pointer comes from a shared reference.

Unfortunately, these things are very badly documented: I'm only aware of this because I followed many Unsafe Code Guideline discussions. Best source that I could get is the following [comment](https://github.com/rust-lang/rust/issues/66136#issuecomment-550009220) from @RalfJung, who is very knowledgeable on the subject. The details are probably written in the Stacked Borrows paper, to be fair.

Even if it's not UB *right now* because of ill-specification or because of the compiler not taking advantage of it, we'd better be safe.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -927,7 +927,7 @@ unsafe impl pyo3::PyTypeInfo for MyClass {
     const FLAGS: usize = 0;
 
     #[inline]
-    fn type_object_raw(py: pyo3::Python) -> &'static pyo3::ffi::PyTypeObject {
+    fn type_object_raw(py: pyo3::Python) -> *mut pyo3::ffi::PyTypeObject {
         use pyo3::type_object::LazyStaticType;
         static TYPE_OBJECT: LazyStaticType = LazyStaticType::new();
         TYPE_OBJECT.get_or_init::<Self>(py)
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -401,7 +401,7 @@ fn impl_class(
             const FLAGS: usize = #(#flags)|* | #extended;
 
             #[inline]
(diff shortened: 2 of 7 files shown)


davidhewitt:
I wonder about a possible further refactoring: `PyTypeInfo` could be `PyTypeInfo: PyTypeObject`, and we could add an `.as_raw()` method to `pyo3::types::PyType` which returns `*mut ffi::PyTypeObject`.

kngwyu:
> I wonder about a possible further refactoring: PyTypeInfo could be PyTypeInfo: PyTypeObject, and we could add an .as_raw() method to pyo3::types::PyType which returns *mut ffi::PyTypeObject.

Since `impl<T: PyTypeInfo> PyTypeObject for T` uses `type_object_raw`, I don't think it's possible.

kngwyu:
Thanks!

RalfJung:
> Unfortunately, these things are very badly documented: I'm only aware of this because I followed many Unsafe Code Guideline discussions. Best source that I could get is the following comment from @RalfJung, who is very knowledgeable on the subject. The details are probably written in the Stacked Borrows paper, to be fair.

Yeah, the details of our reference rules are pretty fuzzy, and Stacked Borrows is a promising next step but it is not an official model yet.

However, [this page](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) is clear at least about some thing. It is UB to do the following:

> Mutating immutable data. All data inside a const item is immutable. Moreover, all data reached through a shared reference or data owned by an immutable binding is immutable, unless that data is contained within an UnsafeCell<U>.

PR #995, opened 2020-06-22
Re-enable recursive class attributes

scalexm:
Builds on #994 (it's actually the call to `PyType_Modify` through a shared reference which made me realize the problem!).

Last thing to do will be to replace `value: GILOnceCell<*mut ffi::PyTypeObject>` with `UnsafeCell<ffi::PyTypeObject>` + `GILOnceCell<()>` as suggested by @kngwyu.

### DIFF
--- a/src/once_cell.rs
+++ b/src/once_cell.rs
@@ -58,6 +58,10 @@ impl<T> GILOnceCell<T> {
     ///     calling `f()`. Even when this happens `GILOnceCell` guarantees that only **one** write
     ///     to the cell ever occurs - other threads will simply discard the value they compute and
     ///     return the result of the first complete computation.
+    ///  3) if f() does not release the GIL and does not panic, it is guaranteed to be called
+    ///     exactly once, even if multiple threads attempt to call `get_or_init`
+    ///  4) if f() can panic but still does not release the GIL, it may be called multiple times,
+    ///     but it is guaranteed that f() will never be called concurrently
     pub fn get_or_init<F>(&self, py: Python, f: F) -> &T
     where
         F: FnOnce() -> T,
--- a/src/pyclass.rs
+++ b/src/pyclass.rs
@@ -1,10 +1,10 @@
 //! `PyClass` trait
 use crate::class::methods::{PyClassAttributeDef, PyMethodDefType, PyMethods};
 use crate::class::proto_methods::PyProtoMethods;
-use crate::conversion::{AsPyPointer, FromPyPointer, IntoPyPointer, ToPyObject};
+use crate::conversion::{AsPyPointer, FromPyPointer};
 use crate::pyclass_slots::{PyClassDict, PyClassWeakRef};
 use crate::type_object::{type_flags, PyLayout};
-use crate::types::{PyAny, PyDict};
+use crate::types::PyAny;
 use crate::{class, ffi, PyCell, PyErr, PyNativeType, PyResult, PyTypeInfo, Python};
 use std::ffi::CString;
 use std::os::raw::c_void;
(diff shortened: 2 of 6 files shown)


src/once_cell.rs
```diff
@@ -58,6 +58,10 @@ impl<T> GILOnceCell<T> {
     ///     calling `f()`. Even when this happens `GILOnceCell` guarantees that only **one** write
     ///     to the cell ever occurs - other threads will simply discard the value they compute and
     ///     return the result of the first complete computation.
+    ///  3) if f() does not release the GIL and does not panic, it is guaranteed to be called
+    ///     exactly once, even if multiple threads attempt to call `get_or_init`
+    ///  4) if f() can panic but still does not release the GIL, it may be called multiple times,
+    ///     but it is guaranteed that f() will never be called concurrently
```
davidhewitt, line 64: 👍 nice

src/type_object.rs
```diff
@@ -135,64 +131,117 @@ where
...
+            unsafe { ffi::PyType_Modified(type_object) };
+
+            // Initialization successfully complete, can clear the thread list.
+            // (No further calls to get_or_init() will try to init, on any thread.)
+            *self.initializing_threads.lock() = Vec::new();
+            result
+        });
+
+        if let Err(err) = result {
+            err.clone_ref(py).print(py);
```
kngwyu, line 219: Why we need `clone` here?

src/pyclass.rs
```diff
@@ -248,14 +239,22 @@ fn py_class_flags<T: PyTypeInfo>(type_object: &mut ffi::PyTypeObject) {
     }
 }
 
+pub(crate) fn py_class_attributes<T: PyMethods>() -> Vec<PyClassAttributeDef> {
```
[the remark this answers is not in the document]
davidhewitt: I _think_ we need to finish evaluating all class attributes before we can then safely start modifying `tp_dict`, which is why we have to collect them in a `Vec`. I could be wrong though.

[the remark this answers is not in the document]
scalexm: We’ll need to collect them in a `Vec` for the reason evoked by @davidhewitt, but we already do that inside `get_or_init`.

So I agree this intermediate `Vec` is unnecessary and could be an `impl Iterator`.

src/type_object.rs
scalexm: We only have a `&Result<(), PyErr>`, and it seems that `PyErr::print` takes `self` as an argument, not `&self`. So I used `clone_ref` (we are in an unlikely error path anyway), but if there is a better way I can change that.

kngwyu: Ah, I see, thanks. Then I think we need `clone_ref`.

src/type_object.rs
```diff
@@ -135,64 +131,117 @@ where
...
+        // return from the function.
+        let result = self.tp_dict_filled.get_or_init(py, move || {
+            let tp_dict = unsafe { (*type_object).tp_dict };
+            let result = initialize_tp_dict(py, tp_dict, items);
+            // See discussion on #982 for why we need this.
+            unsafe { ffi::PyType_Modified(type_object) };
+
+            // Initialization successfully complete, can clear the thread list.
+            // (No further calls to get_or_init() will try to init, on any thread.)
+            *self.initializing_threads.lock() = Vec::new();
```
kngwyu, line 214: How about 
```suggestion
            self.initializing_threads.lock().clear();
```

davidhewitt: `.clear()` doesn't release the allocated memory.

davidhewitt:
@scalexm could you also please remove the CHANGELOG entry which I'd previously written about disabling class attributes (as that's now incorrect):

> - Disable `#[classattr]` where the class attribute is the same type as the class. (This may be re-enabled in the future; the previous implemenation was unsound.) [#975](https://github.com/PyO3/pyo3/pull/975)

src/type_object.rs
kngwyu: :+1: 
That makes sense.

kngwyu:
Thank you for the update.

> Last thing to do will be to replace value: GILOnceCell<*mut ffi::PyTypeObject> with UnsafeCell<ffi::PyTypeObject> + GILOnceCell<()>

Are you going to push the change to this branch?
Or can I merge this PR as is?

scalexm:
You can merge it, we can do that other change in another PR as it’s not a bug fix but rather a technical improvement.

PR #996, opened 2020-06-23
Refactor `#[pyproto]` Result types

davidhewitt:
Fixes #798 

I reworked `#[pyproto]` result types to use the `IntoPyCallbackOutput` trait. This led to a couple of nice improvements:
- Return values can now be `PyResult<T>` or just `T` (the `PyResult` wrapping is now optional, previously was required). 
- It was easy to solve #798 without some hack.

Following up will be two new PRs which come easily from this:
- Enable return value from `__next__` / `__anext__`
- Allow skipping return type completely when it is `-> ()`

Also the overall code is simpler, which is nice.

PR #997, opened 2020-06-23
Add __next__ return

davidhewitt:
This builds on #996 by enabling ability to `return` a value in the `StopIteration` exception.

If you like this approach I'll write some docs in the guide before we merge this PR.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -808,10 +808,9 @@ It includes two methods `__iter__` and `__next__`:
   * `fn __iter__(slf: PyRefMut<Self>) -> PyResult<impl IntoPy<PyObject>>`
   * `fn __next__(slf: PyRefMut<Self>) -> PyResult<Option<impl IntoPy<PyObject>>>`
 
-Returning `Ok(None)` from `__next__` indicates that that there are no further items.
+Returning `None` from `__next__` indicates that that there are no further items.
 These two methods can be take either `PyRef<Self>` or `PyRefMut<Self>` as their
 first argument, so that mutable borrow can be avoided if needed.
-
 
 Example:
 
@@ -890,6 +889,14 @@ impl PyIterProtocol for Container {
 
 For more details on Python's iteration protocols, check out [the "Iterator Types" section of the library
 documentation](https://docs.python.org/3/library/stdtypes.html#iterator-types).
+
+#### Returning a value from iteration
+
+This guide has so far shown how to use `Option<T>` to implement yielding values during iteration.
+In Python a generator can also return a value. To express this in Rust, PyO3 provides the
+[`IterNextOutput`](https://docs.rs/pyo3/latest/pyo3/class/iter/enum.IterNextOutput.html) enum to
+both `Yield` values and `Return` a final value - see its docs for further details and an example.
+
 
 ## How methods are implemented
 
--- a/src/class/iter.rs
+++ b/src/class/iter.rs
@@ -11,6 +11,39 @@ use crate::{ffi, IntoPy, IntoPyPointer, PyClass, PyObject, Python};
 ///
 /// Check [CPython doc](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_iter)
 /// for more.
+///
+/// # Example
+/// The following example shows how to implement a simple Python iterator in Rust which yields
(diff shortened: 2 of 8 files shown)


src/class/iter.rs
```diff
@@ -64,22 +64,37 @@ impl PyIterMethods {
     }
 }
 
-pub struct IterNextOutput(Option<PyObject>);
+pub enum IterNextOutput {
+    Yield(PyObject),
+    Return(Option<PyObject>),
+}
```
kngwyu, line 70: I suggest
```rust
pub enum IterNextOutput<T, U> {
    Yield(T),
    Return(U),
}

pub type PyIterNextOutput = IterNextOutput<PyObject, PyObject>;

impl IntoPyCallbackOutput<*mut ffi::PyObject> for PyIterNextOutput {
    fn convert(self, _py: Python) -> PyResult<*mut ffi::PyObject> {
        match self {
            IterNextOutput::Yield(o) => Ok(o.into_ptr()),
            IterNextOutput::Return(opt) => Err(crate::exceptions::StopIteration::py_err((opt,))),
        }
    }
}

impl<T, U> IntoPyCallbackOutput<PyIterNextOutput> for IterNextOutput<T, U>
where
    T: IntoPy<PyObject>,
    U: IntoPy<PyObject>,
{
    fn convert(self, py: Python) -> PyResult<PyIterNextOutput> {
        match self {
            IterNextOutput::Yield(o) => Ok(IterNextOutput::Yield(o.into_py(py))),
            IterNextOutput::Return(o) => Ok(IterNextOutput::Return(o.into_py(py))),
        }
    }
}

impl<T> IntoPyCallbackOutput<PyIterNextOutput> for Option<T>
where
    T: IntoPy<PyObject>,
{
    fn convert(self, py: Python) -> PyResult<PyIterNextOutput> {
        match self {
            Some(o) => Ok(PyIterNextOutput::Yield(o.into_py(py))),
            None => Ok(PyIterNextOutput::Return(py.None())),
        }
    }
}
```

kngwyu:
Thank you! Looks good but I want to make general `IterNextOutput<T, U>` work.

davidhewitt:
Thanks, I like that design. I've pushed that to the branch and added a small sample to the guide.

guide/src/class.md
```diff
@@ -891,6 +890,52 @@ impl PyIterProtocol for Container {
 For more details on Python's iteration protocols, check out [the "Iterator Types" section of the library
 documentation](https://docs.python.org/3/library/stdtypes.html#iterator-types).
 
+#### Returning a value from iteration
```
kngwyu, line 893: I think `PyIterProtocols`'s doc would be a better place for this document. `class.md` is too long now.

guide/src/class.md
```diff
@@ -891,6 +890,52 @@ impl PyIterProtocol for Container {
...
+for example with the following code:
+
+```python
+def iter_():
+    for i in range(5):
+        yield i
+    return "Ended"
+```
+
+When the generator `iter_()` returns with `"Ended"`, this is translated to a `StopIteration` exception `e` with `e.value = "Ended"`.
```
kngwyu, line 905: This is an implementation detail and I don't think it's practically important.

davidhewitt:
Good feedback, thanks - I've shortened the doc and moved most of it to `PyIterProtocol`.

kngwyu:
Looks rustfmt complains...

guide/src/class.md
```diff
@@ -891,6 +890,14 @@ impl PyIterProtocol for Container {
 For more details on Python's iteration protocols, check out [the "Iterator Types" section of the library
 documentation](https://docs.python.org/3/library/stdtypes.html#iterator-types).
 
+#### Returning a value from iteration
+
+This guide has so far shown how to use `Option<T>` to implement yielding values during iteration.
+In Python a generator can also return a value. To express this in Rust, PyO3 provides the
+[`IterNextOutput`](https://docs.rs/pyo3/latest/pyo3/class/iter/enum.IterNextOutput.html) enum to
```
kngwyu, line 897: Nice :+1:

davidhewitt:
🤦 pushed fixed formatting.

kngwyu:
Thanks!

PR #998, opened 2020-06-23
Allow omitting return type for `#[pyproto]`

davidhewitt:
This builds on #996 by removing the need to write `-> ()` in the return type for `#[pyproto]` methods.


kngwyu:
LGTM, thank you!

PR #999, opened 2020-06-23
Allow #[getter] and #[setter] functions to take PyRef

### LINKED ISSUE #837: Property methods cannot take self as a PyRef
I would like to be able to write code like the following:

```rust
#[pyclass]
struct Proxy {
    parent: Py<Foo>,
}

#[pyclass]
struct Foo {}

#[pymethods]
impl Foo {
    #[getter]
    fn proxy(slf: PyRef<Self>) -> Proxy {
        Proxy {
            parent: slf.into(),
        }
    }
}
```

However, this currently fails with the following error:
```
error: Getter function can only have one argument of type pyo3::Python
  --> src/lib.rs:12:20
   |
12 |     fn proxy(slf: PyRef<Self>) -> Proxy {
   |                   ^^^^^^^^^^^
```

(For completeness, this should also apply to setters, as well as support all of the other wrapper types for Python objects)

davidhewitt:
Closes #837 

I rewrote the proc_macro code to allow `#[getter]` and `#[setter]` functions to take `&self` as well as `PyRef<Self>` etc.

It might also be possible to make a similar change so that all `#[pyproto]` methods can also choose beteween `&self`, `&mut self`, `PyRef<Self>`, etc. but this change was necessary first so I began with this PR.

### DIFF
--- a/pyo3-derive-backend/src/method.rs
+++ b/pyo3-derive-backend/src/method.rs
@@ -132,6 +132,23 @@ impl<'a> FnSpec<'a> {
         let mut arguments = Vec::new();
         let mut inputs_iter = sig.inputs.iter();
 
+        let mut parse_receiver = |msg: &'static str| {
+            inputs_iter
+                .next()
+                .ok_or_else(|| syn::Error::new_spanned(sig, msg))
+                .and_then(parse_method_receiver)
+        };
+
+        // strip get_ or set_
+        let strip_fn_name = |prefix: &'static str| {
+            let ident = sig.ident.unraw().to_string();
+            if ident.starts_with(prefix) {
+                Some(syn::Ident::new(&ident[prefix.len()..], ident.span()))
+            } else {
+                None
+            }
+        };
+
         // Parse receiver & function type for various method types
         let fn_type = match fn_type_attr {
             Some(MethodTypeAttribute::StaticMethod) => FnType::FnStatic,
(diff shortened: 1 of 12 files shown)


kngwyu:
Thank you! The approach looks good, but it seems like there are some rooms for refactoring. I'll give a detailed review later.
In addition, 
> It might also be possible to make a similar change so that all #[pyproto] methods can also choose beteween &self, &mut self, PyRef<Self>, etc. but this change was necessary first so I began with this PR.

Is it possible? 
Each protocol method has trait defintion(e.g., `fn __len__(&'p self) -> Self::Result`) and it is really restrictive.

davidhewitt:
> I'll give a detailed review later.

:+1: no rush. I'm sure there are things we can improve in here further. Also I don't mind if you want to wait for this until after `0.11` is released as it's quite a hefty rewrite of the proc macro code!

> Is it possible?

That's why I say _might_ :smile: . The pyproto code uses the same function parser so this PR would at least enable us to parse such function definitions. I need to experiment another time to understand if the rest of the implementation is possible!

davidhewitt:
I spotted a refactoring around `SelfType` to simplify the code there.

Also I was able to use `quote_spanned!` instead of `quote!` so that invalid self types get highlighted appropriately in error messages (see the `compile_error` test). With just `quote!` the error message just highlights the `#[pymethods]` block. This will be worth remembering in the future when we edit proc macro code!

pyo3-derive-backend/src/method.rs
```diff
@@ -75,27 +124,107 @@ impl<'a> FnSpec<'a> {
...
+                                syn::Error::new_spanned(
+                                    sig,
+                                    "Static method needs #[staticmethod] attribute",
+                                )
+                            },
+                        )
+                        .and_then(parse_method_receiver)?,
+                )
+            }
+        };
```
kngwyu, line 214:
```suggestion
        let mut parse_receiver = |msg: &'static str| {
            inputs_iter
                .next()
                .ok_or_else(|| syn::Error::new_spanned(sig, msg))
                .and_then(parse_method_receiver)
        };

        // strip get_ or set_
        let strip_fn_name = |prefix: &'static str| {
            let ident = sig.ident.unraw().to_string();
            if ident.starts_with(prefix) {
                Some(syn::Ident::new(&ident[prefix.len()..], ident.span()))
            } else {
                None
            }
        };

        // Parse receiver & function type for various method types
        let fn_type = match fn_type_attr {
            Some(MethodTypeAttribute::StaticMethod) => FnType::FnStatic,
            Some(MethodTypeAttribute::ClassAttribute) => {
                if !sig.inputs.is_empty() {
                    return Err(syn::Error::new_spanned(
                        name,
                        "Class attribute methods cannot take arguments",
                    ));
                }
                FnType::ClassAttribute
            }
            Some(MethodTypeAttribute::New) => FnType::FnNew,
            Some(MethodTypeAttribute::ClassMethod) => {
                // Skip first argument for classmethod - always &PyType
                let _ = inputs_iter.next();
                FnType::FnClass
            }
            Some(MethodTypeAttribute::Call) => {
                FnType::FnCall(parse_receiver("expected receiver for #[call]")?)
            }
            Some(MethodTypeAttribute::Getter) => {
                // Strip off "get_" prefix if needed
                if python_name.is_none() {
                    python_name = strip_fn_name("get_");
                }

                FnType::Getter(parse_receiver("expected receiver for #[getter]")?)
            }
            Some(MethodTypeAttribute::Setter) => {
                // Strip off "set_" prefix if needed
                if python_name.is_none() {
                    python_name = strip_fn_name("set_");
                }

                FnType::Setter(parse_receiver("expected receiver for #[setter]")?)
            }
            None => FnType::Fn(parse_receiver(
                "Static method needs #[staticmethod] attribute",
            )?),
        };
```

pyo3-derive-backend/src/method.rs
```diff
@@ -322,44 +408,51 @@ fn parse_method_attributes(
...
+    macro_rules! set_ty {
+        ($new_ty:expr, $ident:expr) => {
+            if ty.is_some() {
+                return Err(syn::Error::new_spanned(
+                    $ident,
+                    "Cannot specify a second method type",
+                ));
+            } else {
+                ty = Some($new_ty);
+            }
```
kngwyu, line 423:
```suggestion
            if ty.replace($new_ty).is_some() {
                return Err(syn::Error::new_spanned(
                    $ident,
                    "Cannot specify a second method type",
                ));
            }
```

pyo3-derive-backend/src/method.rs
```diff
@@ -454,77 +540,33 @@ fn parse_method_attributes(
...
-                ))
+        if let Some(ty) = ty {
+            match ty {
+                New | Call | Getter | Setter => {
+                    return Err(syn::Error::new_spanned(
+                        name,
+                        "name not allowed with this method type",
+                    ))
+                }
+                _ => {}
```
kngwyu, line 560:
```suggestion
    if let (Some(name), Some(ty)) = (&name, ty) {
        match ty {
            New | Call | Getter | Setter => {
                return Err(syn::Error::new_spanned(
                    name,
                    "name not allowed with this method type",
                ))
            }
            _ => {}
        }
    }

```

davidhewitt:
Thanks for the review - I merged all of the suggestions!

tests/test_compile_error.rs
```diff
@@ -5,6 +5,8 @@ fn test_compile_errors() {
     t.compile_fail("tests/ui/invalid_macro_args.rs");
     t.compile_fail("tests/ui/invalid_property_args.rs");
     t.compile_fail("tests/ui/invalid_pyclass_args.rs");
+    t.compile_fail("tests/ui/invalid_pymethod_names.rs");
```
davidhewitt, line 8: This test already existed but seemed to accidentally be lost from the compile error test list.

kngwyu: Ooops, I accidently removed it when tackling the CI problem. Thanks!

kngwyu:
Thanks!

PR #1001, opened 2020-06-26
Use Nightly for coverage

kngwyu:
Somehow we get
```
error: can't instrument with gcov profiling when compiling incrementally
```
with stable Rust, even if [`RUSTC_BOOTSTRAP=1`](https://github.com/mozilla/grcov/issues/433) is set.


kngwyu:
It does not work due to the test failure on nightly. I fixed it in #1002.

kngwyu:
Wow :roll_eyes: 
image

Issue #365, opened 2019-02-18
Run clippy and test for all crates in the workspace


konstin:
pyo3-derive-backend has tests which do not run on ci and clippy does currently only chech the pyo3 crate, leaving out all other. It shouldn't be difficult to run those checks on all crates. EDIT: This should also include the tests.

kngwyu:
Now we run `make lint` on travis

PR #1006, opened 2020-06-28
Fix clippy warning

kngwyu:
Fix this warning:
```
error: large size difference between variants
  --> pyo3-derive-backend/src/method.rs:56:5
   |
56 |     TryFromPyCell(syn::Type),
   |     ^^^^^^^^^^^^^^^^^^^^^^^^ this variant is 304 bytes
   |
   = note: `-D clippy::large-enum-variant` implied by `-D warnings`
note: and the second-largest variant is 1 bytes:
  --> pyo3-derive-backend/src/method.rs:55:5
   |
55 |     Receiver { mutable: bool },
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
help: consider boxing the large fields to reduce the total size of the enum
   |
56 |     TryFromPyCell(Box<syn::Type>),
   |                   ^^^^^^^^^^^^^^
```
, though I don't know why this is not detected by CI...

### DIFF
--- a/pyo3-derive-backend/src/method.rs
+++ b/pyo3-derive-backend/src/method.rs
@@ -38,7 +38,7 @@ pub enum MethodTypeAttribute {
     Setter,
 }
 
-#[derive(Clone, PartialEq, Debug)]
+#[derive(Clone, Debug)]
 pub enum FnType {
     Getter(SelfType),
     Setter(SelfType),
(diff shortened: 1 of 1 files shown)


kngwyu:
> I think there's an issue #365 to add CI and testing for all crates?

No, we run `make lint` on travis Python3.8 :thinking:

PR #1007, opened 2020-06-28
Mention `send_wrapper` in the migration guide

davidhewitt:
Based on @konstin's comment in Gitter.

### DIFF
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -41,10 +41,13 @@ struct ThreadSafe {
 }
 ```
 
-Or in situations where you cannot change your `#[pyclass]` to automatically implement `Send`
-(e.g., when it contains a raw pointer), you can use `unsafe impl Send`.
-In such cases, care should be taken to ensure the struct is actually thread safe.
-See [the Rustnomicon](ttps://doc.rust-lang.org/nomicon/send-and-sync.html) for more.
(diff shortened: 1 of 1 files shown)


davidhewitt:
(Sorry for the force-pushes; I tweaked the wording and commit message shortly after opening the PR.)

kngwyu:
Hmm... :thinking: 
`SendWrapper` is internally a pair of `Box<T>` and `ThreadId`.
I don't want to recommend it as a thread-safe alternative of `T` since it requires redundant `Box`.

davidhewitt:
Ok :+1:

konstin:
While it gives a small performance hit, it is imho still preferable to `unsafe impl Send` which can cause nasty bugs if used incorrectly.

birkenfeld:
This change showed both possibilities, so it's definitely helpful when the contents are *not* threadsafe after all.

davidhewitt:
I would also be in favour of encouraging users to use `send_wrapper` instead of unsafe impl if they don't know what they're doing.

As a few of us are in favour of this, reopening.

kngwyu:
I agree that we provide no good interface for a class that we don't want to send.
So, as a *builtin* alternative of send-wrapper, I opened #1009.

kngwyu:
Closed in favor of #1009

PR #1015, opened 2020-07-01
Add introduction to class.md

davidhewitt:
This could probably be better, but at least it's a starting point and will help users discover all the possible attributes.


programmerjake:
the list is missing `#[text_signature]`

davidhewitt:
There's also `#[name]`, but both of those also apply to `#[pyfunction]` as well so I was thinking a new part of the guide might be better for the common attributes...

kngwyu:
Thanks!

PR #1020, opened 2020-07-04
Remove PyString::as_bytes since it cannot return raw bytes

### LINKED ISSUE #1018: Why does PyString::as_bytes insist on valid unicode?
Hello

I'm reading through the documentation of 0.11.1 and I'm a bit confused about the purpose behind https://docs.rs/pyo3/0.11.1/pyo3/types/struct.PyString.html#method.as_bytes. Why it returns bytes but still returns `&[u8]` slice?

On one hand, if it returns the slice, there doesn't seem to be any particular need for refusing invalid unicode. On the other hand, if it already is checked for valid unicode, why not have `as_str` and return it as (more convenient) `&str`? Or, would it make sense to have both, one that returns the raw (potentially broken unicode) data as `[u8]` and the other that does the checking and returns `&str`?

Or is the documentation simply stale about the returned error? It seems the low-level function doesn't explicitly say what happens if there's *invalid* utf8 inside.

kngwyu:
Fixes #1018

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -6,7 +6,6 @@ use crate::{
     PyTryFrom, Python, ToPyObject,
 };
 use std::borrow::Cow;
-use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::str;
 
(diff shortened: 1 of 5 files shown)


src/types/string.rs
```diff
@@ -136,24 +127,9 @@ impl<'a> IntoPy<PyObject> for &'a String {
...
-
-/// Allows extracting strings from Python objects.
-/// Accepts Python `str` and `unicode` objects.
-impl<'a> crate::FromPyObject<'a> for &'a str {
-    fn extract(ob: &'a PyAny) -> PyResult<Self> {
-        let s: Cow<'a, str> = crate::FromPyObject::extract(ob)?;
-        match s {
-            Cow::Borrowed(r) => Ok(r),
-            Cow::Owned(r) => {
-                let r = ob.py().register_any(r);
```
davidhewitt, line 153: This was actually the only use of the api `register_any`. IMO it adds complexity and now doesn't have a clear use case, should we remove this API (or mark deprecated)?

We could discuss this in a separate issue.

kngwyu: Good catch :+1:

davidhewitt: See #1023

Issue #295, opened 2018-12-01
Add a proc macro for exceptions


konstin:
I would be great to have a custom derive for exceptions, similar but more powerful than [create_exception](https://pyo3.rs/master/doc/pyo3/macro.create_exception.html). This would e.g. allow reimplementing [`JSONDecodeError`](https://docs.python.org/3/library/json.html#json.JSONDecodeError):

```rust
#[pyexception(module = json)]
struct JSONDecodeError {
    msg: String,
    doc: String,
    pos: usize,
    lineno: Option<usize>,
    colno: Option<usize>,
}
```

The steps required are roughly:

 1. Look at [examples in syn](https://github.com/dtolnay/syn/tree/master/examples) and pyo3's existing proc macros to learn how to implement a proc macro. [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen) could be another useful resource.
 2. Implement the macro using the existing [impl_exception_boilerplate](https://pyo3.rs/master/doc/pyo3/macro.impl_exception_boilerplate.html) and [create_exception_type_object](https://pyo3.rs/master/doc/pyo3/macro.create_exception_type_object.html) macros. The `#[proc_macro_attribute]` function should go in `pyo3cls/src/lib.rs`, while the actual implementation should be in `pyo3-derive-backend`.
 3. Write tests. Most important is one for a struct without members and one with at least 2 members.
 4. Add a usage example to the guide. You can just copy one of the tests for that.

If you want to work on implementing this or have any questions, feel free ask here or on gitter!

pradyunsg:
I'd like to try my hand at this. Would that be fine?

If so, I'm not super familiar with the details here, so would this issue be a good place to ask questions?

kngwyu:
Thanks, please don't hesitate to ask us!

konstin:
Yes, just ask here!

davidhewitt:
I'm nominating this for `0.12` - I actually already have a PR almost ready for #682 which would lay the groundwork for this.

davidhewitt:
Nearly everything else in `0.12` is close to being ready to go, and this is a potentially complex feature but not yet started. As much as I'm looking forward to having this, I'm going to delay this to the `0.13` release so we have time to design and implement properly.

davidhewitt:
For those watching this, in #1591 we made it possible to write `#[pyclass(extends=PyException)]` (or other exception types) to create exception types.

[...]

PR #1024, opened 2020-07-04
Rework exceptions to be native types

davidhewitt:
This is a rework of pyo3's exception types to be more in line with the rest of the Python native types.

I introduce new exception types `PyException`, `PyRuntimeError` etc. These must be accessed through either `Py<PyBaseException>`, `Py<PyRuntimeError>` or gil-scoped references `&PyBaseException` , `&PyRuntimeError`, just like other native types.

To minimise breakage the old types `Exception`, `RuntimeError` continue to exist and behave as they did before, but I mark them deprecated with a suggestion to use the new types.

I think this is an important first step of #295 - we should then be able to make something very similar to `#[pyclass]` for exceptions. It should also enable #682 to be done relatively cleanly by building off these implementations.

I based this PR on the work of #686, which I think can now be closed. As well as `Py<PyBaseException>`, I was able to implement `std::error::Error` for `Py<T>` and `&T` for all the error types `T`.

This PR will introduce a couple of minor breaking changes, but I think it's worth it:
  - `PyBorrowError` and `PyBorrowMutError` no longer have dedicated Python types and just raise `RuntimeError` in Python.
     I could undo this change, however the `try_borrow()` would have to return `&PyBorrowError` and we'll take a little performance hit. 
  - The imported exceptions e.g. `socket::herror` are now only accessible as native types; existing code using these will break slightly.

If you like this, I'll press ahead with documentation.

TODO:
- [ ] ~~Documentation~~ _EDIT: I'm going to write docs after some follow-up PRs, when it's easier to see what the final result looks like._


kngwyu:
Thank you, looks good except the approach to the breaking change(i.e., Py prefix) :thinking: 
Is there any simpler solution? (e.g., `type Exception = Py<PyException>`)

davidhewitt:
> Is there any simpler solution? (e.g., type Exception = Py<PyException>)

I think any alternative, including the above, would break more user code than this PR already does.

The simplest alternative would be to embrace the breaking change and just remove the old types.

davidhewitt:
> The simplest alternative would be to embrace the breaking change and just remove the old types.

If you think it's okay, I would be very happy to do this. It keeps the pyo3 codebase smaller and I can document these breaking changes immediately in the migration guide.

kngwyu:
How about
```rust
type BaseException = PyBaseException;
```
?
This is enough for me.

programmerjake:
> How about
> 
> ```rust
> type BaseException = PyBaseException;
> ```
> 
> ?
> This is enough for me.

sounds good to me if it's marked deprecated

davidhewitt:
Agreed marking the old functionality as deprecated. I'll try to finish off this PR at the weekend.

davidhewitt:
I've updated this PR and rebased.

I was originally intending to write more documentation as part of this PR, but now as I've been thinking about related designs such as `impl std::error::Error for PyErr` I want to get those into shape before writing new docs.

PR #1036, opened 2020-07-13
Various fixes to edge cases with GILGuard

### LINKED ISSUE #1019: Is this „nasty“ code around GIL sound?
Hello

When reading through the docs, I sometimes have the habit of clicking on the `[src]` button and looking under the hood. And I've come up with some code I'm not sure is not UB (besides me not using unsafe). It might be fine, but I think it makes sense to share it and ask *why* it is fine.

```rust
use pyo3::prelude::*;
use pyo3::wrap_pyfunction;

#[pyfunction]
fn messed_gil_cnt(py: Python<'_>) {
    eprintln!("Start");
    let guard = py.allow_threads(|| {
        eprintln!("Released");
        let guard = GILGuard::acquire();
        eprintln!("Re-acquired");
        guard
    });
    eprintln!("Returned");
    drop(guard);
}

#[pyfunction]
fn cross_release() {
    let g1 = GILGuard::acquire();
    let g2 = GILGuard::acquire();

    drop(g1);
    drop(g2);
}

#[pymodule]
fn broken(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
    m.add_wrapped(wrap_pyfunction!(messed_gil_cnt))?;
    m.add_wrapped(wrap_pyfunction!(cross_release))?;

    Ok(())
}
```

The thing I worry about the `cross_release` is, the documentation for `PyGILState_Ensure` seems to hint (not very explicitly) that the corresponding `PyGILState_Release` should be called in reverse order:

> After this call, Python’s state will be the same as it was prior to the corresponding PyGILState_Ensure() call (but generally this state will be unknown to the caller, hence the use of the GILState API).

If taken literally, the interpreter should be left in the state after acquiring `g1`, because it's the state it should return to after releasing `g2`. I don't know how to „prove“ anything wrong is happening, though, I haven't seen anything weird manifest.

The `messed_gil_cnt` is similar, but I've noticed one further problem. When I'm looking at the source code here: https://docs.rs/pyo3/0.11.1/src/pyo3/python.rs.html#129-150, I *think* this is happening:

* Starts with `GIL_COUNT` set to some unknown `n`.
* `GIL_COUNT` is set to 0.
* A new GILGuard is created, increasing `GIL_COUNT` to `1`. Furthermore, as it is 0, it creates the `GILPool` (because the `gil_is_acquired` works based on the `GIL_COUNT`, https://docs.rs/pyo3/0.11.1/src/pyo3/gil.rs.html#167), but at least the comment above says there should not be two pools on the same thread.
* When leaving the scope, the `GIL_COUNT` is reset back to `n`, losing the one from the created `GILGuard`
* But the `GILGuard is „exported“ out through the return value, so that number is wrong.

I don't know if it is related, but the code deadlocks, it never prints `Returned`.

davidhewitt:
Closes #1019 

I introduce panics for the two "nasty" cases discussed in that issue.

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -12,13 +12,13 @@ static START: sync::Once = sync::Once::new();
 thread_local! {
     /// This is a internal counter in pyo3 monitoring whether this thread has the GIL.
     ///
-    /// It will be incremented whenever a GILPool is created, and decremented whenever they are
-    /// dropped.
+    /// It will be incremented whenever a GILGuard or GILPool is created, and decremented whenever
+    /// they are dropped.
     ///
     /// As a result, if this thread has the GIL, GIL_COUNT is greater than zero.
     ///
     /// pub(crate) because it is manipulated temporarily by Python::allow_threads
-    pub(crate) static GIL_COUNT: Cell<u32> = Cell::new(0);
+    pub(crate) static GIL_COUNT: Cell<usize> = Cell::new(0);
 
     /// Temporally hold objects that will be released when the GILPool drops.
     static OWNED_OBJECTS: RefCell<Vec<NonNull<ffi::PyObject>>> = RefCell::new(Vec::with_capacity(256));
@@ -110,7 +110,8 @@ pub fn prepare_freethreaded_python() {
     });
 }
 
-/// RAII type that represents the Global Interpreter Lock acquisition.
+/// RAII type that represents the Global Interpreter Lock acquisition. To get hold of a value
+/// of this type, see [`Python::acquire_gil`](struct.Python.html#method.acquire_gil).
 ///
 /// # Example
 /// ```
@@ -128,15 +129,17 @@ pub struct GILGuard {
 }
 
 impl GILGuard {
-    /// Acquires the global interpreter lock, which allows access to the Python runtime. This is
-    /// safe to call multiple times without causing a deadlock.
-    ///
-    /// If the Python runtime is not already initialized, this function will initialize it.
-    /// See [prepare_freethreaded_python()](fn.prepare_freethreaded_python.html) for details.
+    /// Retrieves the marker type that proves that the GIL was acquired.
+    #[inline]
+    pub fn python(&self) -> Python {
+        unsafe { Python::assume_gil_acquired() }
+    }
+
+    /// PyO3 internal API for acquiring the GIL. The public API is Python::acquire_gil.
     ///
     /// If PyO3 does not yet have a `GILPool` for tracking owned PyObject references, then this
     /// new `GILGuard` will also contain a `GILPool`.
-    pub fn acquire() -> GILGuard {
+    pub(crate) fn acquire() -> GILGuard {
         prepare_freethreaded_python();
 
         let gstate = unsafe { ffi::PyGILState_Ensure() }; // acquire GIL
@@ -146,6 +149,8 @@ impl GILGuard {
         let pool = if !gil_is_acquired() {
             Some(unsafe { GILPool::new() })
         } else {
+            // As no GILPool was created, need to update the gil count manually.
+            increment_gil_count();
             None
         };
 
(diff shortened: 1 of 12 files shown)


kngwyu:
- For `allow_threads`
I think it's OK to add `T: Send`.
It's not so meaningful to return `T: !Send` in this context.

- For nested case
I'll leave some comments.

src/gil.rs
```diff
@@ -20,6 +20,9 @@ thread_local! {
     /// pub(crate) because it is manipulated temporarily by Python::allow_threads
     pub(crate) static GIL_COUNT: Cell<u32> = Cell::new(0);
 
+    /// Counter purely for GILGuards, to ensure they drop in the same order they are created.
+    static GILGUARD_COUNT: Cell<usize> = Cell::new(0);
```
kngwyu, line 24: `bool` is sufficient. I'll explain why in the below comment.

src/gil.rs
```diff
@@ -165,6 +175,15 @@ impl GILGuard {
 /// The Drop implementation for `GILGuard` will release the GIL.
 impl Drop for GILGuard {
     fn drop(&mut self) {
+        // Safety check & update counter
+        GILGUARD_COUNT.with(|c| {
+            if c.get() - 1 != self.previous_count {
+                panic!("GILGuards must be dropped in the reverse of the order they are created.")
```
kngwyu, line 181: **We should panic only when the gstate is `PyGILState_UNLOCKED` and GILGuard is nested.**
In this nested case we care, `gstate` is always `PyGILState_LOCKED` and it's not problematic to set it in an incorrect order.

src/gil.rs
davidhewitt: I don't think it is. Consider the case where three GILGuards get created:

```
let gil1 = Python::acquire_gil();
let gil3 = Python::acquire_gil();

drop(gil2);
```

What should dropping `gil2` do? It can't modify the `bool`, because we're still in a nested case with `gil3`.

kngwyu: Ah yes you're right, please don't mind the word

src/gil.rs
kngwyu: What I meant is:
```rust
GIL_COUNT.try_with(|c| {
    if c.get() > 0 and self.gstate == PyGILState_UNLOCKED { panic!("...") }
});
```

davidhewitt: Good idea, I'll go for that. Avoids needing to add `GILGUARD_COUNT`

davidhewitt:
Having looked further into these and subsequent tests, I've concluded that `panic!` is not the right solution for either of these two issues. In both cases, a panic will make it hard or impossible for user code to correct the GIL state, so Python will likely abort the process or deadlock shortly after the panic.

For `allow_threads`, I now agree we should add `T: Send` bound.

For the `GILGuard` drop out of order, I think the solution is less clear. I'm tempted to just add an `abort()` to this instead of panic, to be on the safe side.

kngwyu:
> so Python will likely abort the process or deadlock shortly after the panic.

1. In extension case (we get `GILGuard` in `#[pyfunction]` or so), we don't need to worry about that so seriously. `gstate` is mostly `LOCKED`.
2. In embedding case (we get `GILGuard` from main function or so), we don't need to recover the interpreter state, right?

davidhewitt:
It's not quite so clear-cut as that. Even for extensions, use of `Python::allow_threads` could lead to acquisition of new `UNLOCKED` GILGuards. Also, for the embedding case (~ tests) I was seeing the test process deadlock after a panic.

I have to say I really don't like the `abort` UX, so in the end I think I will keep `panic!` as the implementation. I'll add a whack of documentation to `Python::acquire_gil` to warn against possible issues here, and encourage users to use `Python::with_gil`.

Also, I think I'll change `GILGuard::acquire` to be just `pub(crate)`, which will push users of it onto `Python::acquire_gil`. This means that all the safety notes can just go on `Python::acquire_gil`, as well as discouragement from using it!

davidhewitt:
I've pushed those discussed changes.

One interesting discovery from this was that `PyErr` is not `Send`. For now, I've pushed a couple of tweaks to this branch to make `PyErr: Send`. If you think this is a good idea, I'll first open a separate PR to make `PyErr: Send` properly. I think this would be a good idea anyway.

kngwyu:
Thank you, though I don't understand the intention of some changes in the test files.

> For now, I've pushed a couple of tweaks to this branch to make PyErr: Send. If you think this is a good idea, I'll first open a separate PR to make PyErr: Send properly. I think this would be a good idea anyway.

I agree that `PyErr` should be `Send`, but I'm sorry I can't understand what changes are related to `PyErr`.

davidhewitt:
> I agree that PyErr should be Send, but I'm sorry I can't understand what changes are related to PyErr.

The example for `Python::allow_threads` returns `PyResult<()>`, and to make `PyResult: Send`, I had to make `PyErr: Send`. The alternative was to change the example, but I think the example is a good one.

Anyway, please hold on merging this PR; I'll make a new PR for `PyErr: Send` shortly.

> I don't understand the intention of some changes in the test files.

Maybe the changes you are referring to are where I had to remove `GILGuard::acquire` usage from tests, because I've taken that function out of the PyO3 public API.

davidhewitt:
Having made the `PyErr` changes separately in #1067, I've rebased this PR and proceeding now to merge.

PR #1037, opened 2020-07-13
Add `Python::with_gil`

davidhewitt:
As suggested in #1019, this is a new `Python::with_gil` API for executing a closure with the GIL.

It avoids some of the potential panics introduced in #1036, and also might be "nicer" because the usual `acquire_gil(); gil.python()` dance can be replaced by a single `with_gil` call.

I've updated some key examples in the documentation to use `with_gil`, as I think it's probably better to encourage users to use this new API rather than `acquire_gil()`.

### DIFF
--- a/guide/src/trait_bounds.md
+++ b/guide/src/trait_bounds.md
@@ -79,37 +79,36 @@ struct UserModel {
 impl Model for UserModel {
     fn set_variables(&mut self, var: &Vec<f64>) {
         println!("Rust calling Python to set the variables");
-        let gil = Python::acquire_gil();
-        let py = gil.python();
-        let values: Vec<f64> = var.clone();
-        let list: PyObject = values.into_py(py);
-        let py_model = self.model.as_ref(py);
-        py_model
-            .call_method("set_variables", (list,), None)
-            .unwrap();
+        Python::with_gil(|py| {
+            let values: Vec<f64> = var.clone();
+            let list: PyObject = values.into_py(py);
+            let py_model = self.model.as_ref(py);
+            py_model
+                .call_method("set_variables", (list,), None)
+                .unwrap();
+        })
     }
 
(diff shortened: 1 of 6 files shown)


guide/src/trait_bounds.md
```diff
@@ -180,61 +179,55 @@ This wrapper will also perform the type conversions between Python and Rust.
...
+#              .unwrap();
+#      })
+#
 #  }
 # }
 
 #[pymethods]
 impl UserModel {
-    pub fn set_variables(&mut self, var: Vec<f64>) -> PyResult<()> {
+    pub fn set_variables(&mut self, var: Vec<f64>) {
```
kngwyu, line 218: This method should take `py: Python`.

src/python.rs
```diff
@@ -46,6 +46,31 @@ pub use gil::prepare_freethreaded_python;
...
+    /// ```
+    /// use pyo3::prelude::*;
+    /// Python::with_gil(|py| -> PyResult<()> {
+    ///     let x: i32 = py.eval("5", None, None)?.extract()?;
+    ///     assert_eq!(x, 5);
+    ///     Ok(())
+    /// });
+    /// ```
+    #[inline]
+    pub fn with_gil<F, R>(f: F) -> R
```
kngwyu, line 66: Nice :+1:

guide/src/trait_bounds.md
davidhewitt: I agree, but it's actually quite hard to solve.

The trait signature `impl Model for UserModel` has no way to pass a `Python` into the trait methods. The idea for this example is to show how to wrap a Rust trait which doesn't know anything about `Python`, so I don't think changing the trait definition is a solution either.

For now I would like to leave it as it is, but if we come up with a good idea how to clean this up later I would be happy to resolve it. (I'll create an issue so we remember this.)

Issue #539, opened 2019-07-15
Rename `PYTHON_SYS_EXECUTABLE`


konstin:
We should rename `PYTHON_SYS_EXECUTABLE` to something more more understandable like `PYO3_INTERPRETER`. The new name is up for bikeshedding.

pyo3 should continue to support both the `PYTHON_SYS_EXECUTABLE` and the new variable for some time, and we need to migrate setuptools-rust and pyo3-pack.

davidhewitt:
Closed via #1039

PR #1039, opened 2020-07-13
Rename PYTHON_SYS_EXECUTABLE to PYO3_PYTHON

davidhewitt:
As suggested in #539 .

Once merged, I'll update `setuptools-rust` and `maturin` to output both environment variables for now, as a compatible way to manage the migration.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -386,7 +386,7 @@ fn get_rustc_link_lib(config: &InterpreterConfig) -> Result<String> {
 ///
 /// The following locations are checked in the order listed:
 ///
-/// 1. If `PYTHON_SYS_EXECUTABLE` is set, this intepreter is used and an error is raised if the
+/// 1. If `PYO3_PYTHON` is set, this intepreter is used and an error is raised if the
 /// version doesn't match.
 /// 2. `python`
 /// 3. `python{major version}`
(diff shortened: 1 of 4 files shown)


davidhewitt:
cc @konstin, who suggested originally. I'm also toying with the name `PYO3_PYTHON`.

build.rs
```diff
@@ -637,7 +640,7 @@ fn main() -> Result<()> {
         // TODO: Find out how we can set -undefined dynamic_lookup here (if this is possible)
     }
 
-    let env_vars = ["LD_LIBRARY_PATH", "PATH", "PYTHON_SYS_EXECUTABLE", "LIB"];
+    let env_vars = ["LD_LIBRARY_PATH", "PATH", "PYO3_INTERPRETER", "LIB"];
```
kngwyu, line 643: Is this compatible with old tools?

davidhewitt: Good catch, we'll need both variables in here for now!

davidhewitt:
Pushed the correction, and also renamed the variable to `PYO3_PYTHON`. I think the meaning is still clear, and it's shorter than `PYO3_INTERPRETER`.

If nobody has any objections to the new name I'll probably merge this on Sunday and write patches for `setuptools-rust` and `maturin`.

davidhewitt:
:+1: I plan to add patches to `maturin` and `setuptools-rust` before merging this PR!

davidhewitt:
I've now pushed patches both tools, so I've rebased this PR and will merge shortly.

Issue #991, opened 2020-06-22
Extend Python Class from Rust


davidhewitt:
A user asked a question on Gitter just now about extending a Python class from Rust.

While I think it's possible to do this by hand with a lot of unsafe code by implementing `PyTypeInfo` for the base type manually, it's pretty complicated. 

This is an issue to think one day about how to make this easier. Possibly it's good enough to document the current process, or maybe there are design changes we can make internally.

nekitdev:
I had an initial idea of calling ``PyType_Type``, just like from python, in order to construct new classes with multiple inheritance:
```python
class A:
    a = 13

class B:
    b = 42

C = type("C", (A, B), {})
```

davidhewitt:
Interesting. You might find #1152 interesting which I think might use that or something similar (haven't had a chance to review it yet)

chinedufn:
> Possibly it's good enough to document the current process

Are there any existing examples that you can point me to of how this is done now?

Thanks!

LegNeato:
I recently hit this and asked in [gitter](https://gitter.im/PyO3/Lobby?at=60321eef8621811d588eae1c) too

> "How do I use abstract base classes / make my Rust code with the same api as a built in class pass an isinstance check? I 
 can't seem to find it in the docs or any examples. I specifically want my Rust class to pass isinstance(i, uuid.UUID) where i is a  instance created by my Rust code.
> (if it isn't supported, happy to put up a PR if someone will point me in the right direction and it is reasonable for a pyo3 beginner)"

> David Hewitt @davidhewitt Feb 20 22:50
> @LegNeato you're looking at a special case of PyO3/pyo3#991. Unfortunately this is not supported at the moment. It's quite a hard issue to solve, but if you're interested in implementing it I can discuss the design and implementation steps with you on that issue.

@davidhewitt I likely won't have time soon due to work so don't sweat it, but if you find some cycles and can throw some information here I can possibly take a crack at it in the future when work settles down.

LegNeato:
Potentially relevant: https://www.python.org/dev/peps/pep-0253/

calbaker:
I tried using `subclass` in a `pyclass` in https://github.com/NREL/fastsim (can grant access to this private repo if needed), and when I tried:  

```
class ChildVehicle(fsr.RustVehicle):
    def get_veh_kg(self) -> float:
        return self.veh_kg

[...]

PR #1045, opened 2020-07-17
enum WIP

stillinbeta:

### DIFF
--- a/pyo3-derive-backend/src/pyenum.rs
+++ b/pyo3-derive-backend/src/pyenum.rs
@@ -85,6 +85,11 @@ fn impl_enum(
 
         impl pyo3::pyclass::PyClassAlloc for #enum_ {}
 
+        // TODO: handle not in send
+        impl pyo3::pyclass::PyClassSend for #enum_ {
+            type ThreadChecker = pyo3::pyclass::ThreadCheckerStub<#enum_>;
+        }
+
         #inventory
     })
 }
--- a/pyo3-derive-backend/src/pyenum.rs
+++ b/pyo3-derive-backend/src/pyenum.rs
@@ -65,6 +65,7 @@ fn impl_enum(
         #[pymethods]
         impl #enum_ {
             #(
+                #[allow(non_upper_case_globals)]
                 #variant_consts
             )*
         }
--- a/pyo3-derive-backend/src/pyenum.rs
+++ b/pyo3-derive-backend/src/pyenum.rs
@@ -1,0 +1,90 @@
+// Copyright (c) 2017-present PyO3 Project and Contributors
+
+use crate::pyclass::impl_methods_inventory;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+pub fn build_py_enum(enum_: &syn::ItemEnum) -> syn::Result<TokenStream> {
+    let mut variants = Vec::new();
+
+    for variant in enum_.variants.iter() {
+        if !variant.fields.is_empty() {
+            return Err(syn::Error::new_spanned(
+                variant,
+                "#[pyenum] only supports unit enums",
+            ));
+        }
+        if let Some((_, syn::Expr::Lit(lit))) = &variant.discriminant {
+            variants.push((variant.ident.clone(), lit.clone()))
+        } else {
+            return Err(syn::Error::new_spanned(
+                variant,
+                "#[pyenum] requires explicit discriminant (MyVal = 4)",
+            ));
+        }
+    }
+
+    impl_enum(&enum_.ident, variants)
+}
+
+fn impl_enum(
+    enum_: &syn::Ident,
+    _variants: Vec<(syn::Ident, syn::ExprLit)>,
+) -> syn::Result<TokenStream> {
+    let inventory = impl_methods_inventory(enum_);
+
+    let enum_name = enum_.to_string();
+
+    Ok(quote! {
+        unsafe impl pyo3::type_object::PyTypeInfo for #enum_ {
+            type Type = #enum_;
+            type BaseType = pyo3::PyAny;
+            type Layout = pyo3::PyCell<Self>;
+            type BaseLayout = pyo3::pycell::PyCellBase<pyo3::PyAny>;
+
+            type Initializer = pyo3::pyclass_init::PyClassInitializer<Self>;
+            type AsRefTarget = pyo3::PyCell<Self>;
+
+            const NAME: &'static str = #enum_name;
+            const MODULE: Option<&'static str> = None;
+            const DESCRIPTION: &'static str = "y'know, an enum\0"; // TODO
+            const FLAGS: usize = 0;
+
+            #[inline]
+            fn type_object_raw(py: pyo3::Python) -> *mut pyo3::ffi::PyTypeObject {
+                use pyo3::type_object::LazyStaticType;
+                static TYPE_OBJECT: LazyStaticType = LazyStaticType::new();
+                TYPE_OBJECT.get_or_init::<Self>(py)
+            }
+
+        }
+
+        impl pyo3::PyClass for #enum_ {
+            type Dict =  pyo3::pyclass_slots::PyClassDummySlot ;
+            type WeakRef = pyo3::pyclass_slots::PyClassDummySlot;
+            type BaseNativeType = pyo3::PyAny;
+        }
+
+        impl<'a> pyo3::derive_utils::ExtractExt<'a> for &'a #enum_
+        {
+            type Target = pyo3::PyRef<'a, #enum_>;
+        }
+
+        impl<'a> pyo3::derive_utils::ExtractExt<'a> for &'a mut #enum_
+        {
+            type Target = pyo3::PyRefMut<'a, #enum_>;
+        }
+
+        impl pyo3::class::proto_methods::HasProtoRegistry for #enum_ {
+            fn registry() -> &'static pyo3::class::proto_methods::PyProtoRegistry {
+                static REGISTRY: pyo3::class::proto_methods::PyProtoRegistry
+                    = pyo3::class::proto_methods::PyProtoRegistry::new();
+                &REGISTRY
+            }
+        }
+
+        impl pyo3::pyclass::PyClassAlloc for #enum_ {}
+
+        #inventory
+    })
+}
--- a/pyo3-derive-backend/src/pyenum.rs
+++ b/pyo3-derive-backend/src/pyenum.rs
@@ -1,6 +1,6 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
-use crate::pyclass::impl_methods_inventory;
+use crate::common::{impl_extractext, impl_methods_inventory, impl_proto_registry};
 use proc_macro2::TokenStream;
 use quote::quote;
 
@@ -29,15 +29,30 @@ pub fn build_py_enum(enum_: &syn::ItemEnum) -> syn::Result<TokenStream> {
 
 fn impl_enum(
     enum_: &syn::Ident,
-    _variants: Vec<(syn::Ident, syn::ExprLit)>,
+    variants: Vec<(syn::Ident, syn::ExprLit)>,
 ) -> syn::Result<TokenStream> {
-    let inventory = impl_methods_inventory(enum_);
-
-    let enum_name = enum_.to_string();
+    let enum_cls = impl_class(enum_)?;
+    let variant_cls = variants
+        .iter()
+        .map(|(ident, _)| impl_class(ident))
+        .collect::<syn::Result<Vec<_>>>()?;
 
     Ok(quote! {
-        unsafe impl pyo3::type_object::PyTypeInfo for #enum_ {
-            type Type = #enum_;
+        #enum_cls
+        #(#variant_cls)*
+    })
+}
+
+fn impl_class(cls: &syn::Ident) -> syn::Result<TokenStream> {
+    let inventory = impl_methods_inventory(cls);
+    let extractext = impl_extractext(cls);
+    let protoregistry = impl_proto_registry(cls);
+
+    let clsname = cls.to_string();
+
+    Ok(quote! {
+        unsafe impl pyo3::type_object::PyTypeInfo for #cls {
+            type Type = #cls;
             type BaseType = pyo3::PyAny;
             type Layout = pyo3::PyCell<Self>;
             type BaseLayout = pyo3::pycell::PyCellBase<pyo3::PyAny>;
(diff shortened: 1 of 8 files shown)


davidhewitt:
@stillinbeta many thanks for kicking this off, I'm thrilled to see this! With respect to your comment on the other thread:

> I made an attempt (#1045), but I got frustrated by the amount of boilerplate and traits required.

If we were to work together to design this functionality, have you got interest in continuing onwards with this PR?

If you've lost interest that's also fine, I'll pick this up eventually when I get round to thinking about this feature harder.

Reading through what you've started here, I have a few initial thoughts:

- The variant parsing code you've written looks clean and straightforward for these "C-style" enums, which is nice to see. 👍 
- I'm wondering whether we need to implement `PyClass` and all its gubbins for `#[pyenum]`, or whether we should go down a different direction.
  - I don't _think_ that we'll need either inheritance or being able to have `&mut self` receivers in `#[pymethods]` for these "C-style" enums. These are two of the main reasons `PyClass` is so messy, so maybe we can design something leaner here.
  - On the other hand, at the moment the `#[pyproto]` implementations are _heavily_ coupled to `PyClass`. That's not necessarily an issue, but it does mean that if we want to use `#[pyproto]` with `#[pyenum]` then we're either forced to use `PyClass`, or we'll have to do some refactoring...
- It would likely be nice to add each enum variant as a `#[classattr]` automatically.

kngwyu:
Thanks!

> I don't think that we'll need either inheritance or being able to have &mut self receivers in #[pymethods] for these "C-style" enums.

Yeah, but I don't think both inheritance and `&mut self` don't contribute the code complexity here.
I think what we need is just to use the common code as pyclass.

stillinbeta:
My initial intent was to inherit from enum, but that blocked on #991. I
agree that modularizing the pyclass macro is probably the way forward,
since we're definitely going to need some of it (`EnumName.EnumVariant` was
the UX I was going for, a la the enum modules, and I think that basically
necessitates a class)

davidhewitt:
Agreed. Perhaps if I drafted up a solution to #991, we could get that merged which would pave the way for inheriting from `enum.Enum` in this PR?

I guess we can continue building out the other elements of this PR in the meanwhile!

> `EnumName.EnumVariant` was the UX I was going for, a la the enum modules, and I think that basically necessitates a class

If you haven't seen it yet, you should check out the `#[classattr]` annotation we support in `#[pymethods]` blocks. We already went through a lot of trouble to allow that to safely support attributes that are instances of the containing class, so you should definitely be able to re-use that to expose the variants.

stillinbeta:
I took a stab at getting rid of all the PyClass constraint, but it's going to require a slot of boilerplate on its own:

We need a Layout, and the only blanket implementation is for PyClass: https://docs.rs/pyo3/0.11.1/pyo3/type_object/trait.PyLayout.html#impl-PyLayout%3CT%3E
So does Initializer, which is only implemented in for PyClass and native types (which this definitely isn't). https://docs.rs/pyo3/0.11.1/pyo3/pyclass_init/struct.PyClassInitializer.html

I managed to get enough of the traits in there to compile (was just missing one, it turns out), but it feels... janky. We don't actually need the Inventory or the ProtoRegistry, but the constraints on PyClass mandate them. Do we think we could loosen some of those constraints, or would that break other things?

davidhewitt:
Thanks for continuing to iterate on this!

Regarding the layout, I think that we could try adding a new struct like this:

```
struct PyEnumLayout<T> {
    base: PyAny,
    value: T
}
```

This new type could get a new blanket layout?

As for the initializer, I think that we can use `PyClassInitializer` by adding a new function similar to `create_cell_from_subtype`. The new function would instead return `PyResult<*mut PyEnumLayout<T>>`.

> We don't actually need the Inventory or the ProtoRegistry

I actually think we probably do want to support `#[pymethods]` and `#[pyproto]` for enum classes? So I think we do actually need the inventories.

kngwyu:
> I took a stab at getting rid of all the PyClass constraint, but it's going to require a slot of boilerplate on its own:

I **really** don't think we should do so.
So now we're going to introduce new lots of boilerplates for reducing boilerplates for pyclass?
I don't think it's so meaningful and believe that reusing all `pyclass` components is the simplest way and also easy to maintain.

kngwyu:
I think what we really need here is to provide a nice API for *our* Rustic enum. How to implement it does not matter so much. So I recommend to implement it just as pyclass.

gilescope:
I keep wondering if another way to think about enums is if we supported static getters on pyclasses? That would look from the python side very similar esp if we could subclass Enum...

stillinbeta:
> I actually think we probably do want to support #[pymethods] and #[pyproto] for enum classes? So I think we do actually need the inventories.

I'm not sure I agree with this. Do we have examples of Enums in python with methods on them? It seems like usually they're just sigils in Python.

programmerjake:
> I'm not sure I agree with this. Do we have examples of Enums in python with methods on them? It seems like usually they're just sigils in Python.

Example:
https://git.libre-soc.org/?p=ieee754fpu.git;a=blob;f=src/ieee754/div_rem_sqrt_rsqrt/core.py;h=9f925b871493a6885d92366d648ea6c57bb39c20;hb=HEAD#l28
```python3
class DivPipeCoreOperation(enum.Enum):
    """ Operation for ``DivPipeCore``.

    :attribute UDivRem: unsigned divide/remainder.
    :attribute SqrtRem: square-root/remainder.
    :attribute RSqrtRem: reciprocal-square-root/remainder.
    """

    SqrtRem = 0
    UDivRem = 1
    RSqrtRem = 2

    def __int__(self):
        """ Convert to int. """
        return self.value

    @classmethod
    def create_signal(cls, *, src_loc_at=0, **kwargs):
        """ Create a signal that can contain a ``DivPipeCoreOperation``. """
        return Signal(range(min(map(int, cls)), max(map(int, cls)) + 2),
                      src_loc_at=(src_loc_at + 1),
                      decoder=lambda v: str(cls(v)),
                      **kwargs)
```

stillinbeta:
oh yep, fair enough. In that case, I think PyClass is the way to go. Though I think perhaps initially I'll try to land the change without pymethod support, then we can add it.

gilescope:
Any luck @stillinbeta ?

stillinbeta:
Sorry, I don't have a lot of time during the week to work on this (day job
drains my programming energy). Will give it a shot tomorrow!

kngwyu:
Don't worry, that’s the order of OSS

stillinbeta:
Made some progress this weekend! You can see that you can now return enums from pyfunctions, and they'll be Python-ified properly. However, I ran into issues when trying to pass one back:

```
error[E0119]: conflicting implementations of trait `pyo3::FromPyObject<'_>` for type `MyEnum`:
 --> tests/test_enum.rs:6:1
  |
6 | #[pyenum]
  | ^^^^^^^^^
  |
  = note: conflicting implementation in crate `pyo3`:
          - impl<'a, T> pyo3::FromPyObject<'a> for T
            where T: pyo3::PyClass, T: std::clone::Clone;
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

Any ideas how to get around this? Should be implementing a different trait, like `PyTryInto`?

pyo3-derive-backend/src/pyenum.rs
```diff
@@ -0,0 +1,214 @@
...
+    variants: &Vec<(syn::Ident, syn::ExprLit)>,
+) -> syn::Result<TokenStream> {
+    let matches = variants
+        .iter()
+        .map(|(ident, _)| {
+            variant_enumname(enum_, ident).map(|cls| {
+                let name = cls.to_string();
+                // TODO: this should be
+                // #cls as pyo3::type_object::PyTypeObject>::type_object(py).compare()
+                // but I can't figure out how to get py inside extract()
```
kngwyu, line 89: You can use `pyo3::PyNativeType::py(ob)`

kngwyu:
How does this work without `impl FromPyObject`?

stillinbeta:
Do you mean “does it compile without that part?”

If so, yes, and the appropriate test cases pass

gilescope:
Nice work @stillinbeta ! As an extension I wonder if it would be possible to use enums where the numeric value is not specified? I know on the rust side you could access the https://doc.rust-lang.org/std/mem/fn.discriminant.html but whether that helps at all I'm less certain.

kngwyu:
> If so, yes, and the appropriate test cases pass

Then why should we need that impl?

pyo3-derive-backend/src/pyenum.rs
```diff
@@ -0,0 +1,214 @@
...
+            type BaseNativeType = pyo3::PyAny;
+        }
+
+        #protoregistry
+
+        #extractext
+
+        impl pyo3::pyclass::PyClassAlloc for #typ {}
+
+        // TODO: handle not in send
```
davidhewitt, line 207: I think it's reasonable to expect `#[pyenum]` to always be `Send`? It wouldn't ever carry nontrivial data afaik.

davidhewitt:
> Then why should we need that impl?

+1. Like for `#[pyclass]`, I think this `impl` gets derived automatically if you have `#[derive(Clone)]`.

stillinbeta:
The impl is because I want passing the class in to rust to turn into the
enum variant, not the a type object. I don't think that'll work
automatically

tests/test_enum.rs
```diff
@@ -0,0 +1,50 @@
...
+    let f = wrap_pyfunction!(return_enum)(py);
+    let mynum = py.get_type::<MyEnum>();
+
+    py_run!(py, f mynum, "assert f() == mynum.Variant")
+}
+
+#[pyfunction]
+fn enum_arg(e: MyEnum) {
+    assert_eq!(MyEnum::OtherVariant, e)
+}
```
davidhewitt, line 40: I suggest making this 

```suggestion
fn enum_arg(e: MyEnum) -> String {
    format!("{}", e)
}
```

tests/test_enum.rs
```diff
@@ -0,0 +1,50 @@
...
+
+#[test]
+fn test_enum_arg() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+    let f = wrap_pyfunction!(enum_arg)(py);
+    let mynum = py.get_type::<MyEnum>();
+
+    py_run!(py, f mynum, "f(mynum.Variant)")
```
davidhewitt, line 49: If you take my suggestion from above, we can test both cases:

```suggestion
    py_run!(py, f mynum, "assert f(mynum.Variant) == 'Variant'")
    py_run!(py, f mynum, "assert f(mynum.OtherVariant) == 'OtherVariant'")
```

stillinbeta:
the `test_enum_arg` behaviour doesn't work right now, that's the problem: 
```
TypeError: Can't convert <class 'Variant'> to MyEnum
test test_enum_arg ... FAILED
```

There's no way to automatically determine that I want the class Variant to map to MyEnum::Variant, I need to produce that code myself.

pyo3-derive-backend/src/pyenum.rs
```diff
@@ -0,0 +1,214 @@
...
+
+        #enum_cls
+
+        #(
+            struct #variant_names;
+        )*
+
+        #(
+            #variant_cls
+        )*
```
davidhewitt, line 62: I don't think this bit is needed - instead you should just be making the variants be instances of the the enum class. (Like how it is with the `enum` module.)

davidhewitt:
Oh, I see what you mean now! I don't think that `MyClass.Variant` should be a separate type? I've marked that code in the comment above.

stillinbeta:
Ah yeah, that makes sense. I misinterpreted how Enum was architected.

stillinbeta:
The one advantage of every variant having its own class is it made mapping the variants to their classes fit better into the model of struct -> class. I think to create instances of Variant, I need to make some use of IntoPython, any ideas what that would look like? I'm a little lost in the type system right now.

pyo3-derive-backend/src/pyenum.rs
```diff
@@ -0,0 +1,214 @@
...
+            variant_enumname(enum_, ident).map(|cls| {
+                quote! {
+                    #enum_::#ident => <#cls as pyo3::type_object::PyTypeObject>::type_object(py).to_object(py),
+                }
+            })
+        })
+        .collect::<syn::Result<Vec<_>>>()?;
+
+    Ok(quote! {
+        impl pyo3::FromPy<#enum_> for pyo3::PyObject {
```
davidhewitt, line 132: With the rebase on master, you'll need to change this to `impl IntoPy<PyObject> for #enum`

davidhewitt:
I think `MyClass::Variant.into_py(py)` should already do the into-python conversion you're asking for?

FWIW if you rebase on master I just merged some simplifications to the type system, which hopefully will help!

stillinbeta:
```rust
impl pyo3::IntoPy<pyo3::PyObject> for #enum_ {
            fn into_py(self, py: pyo3::Python) -> pyo3::PyObject {
                self.into_py(py)
            }
        }
```

is definitely not going to do it, that's just recursion.

davidhewitt:
Oh heh I see what you're asking 😄

You probably want to use the same implementation as `#[pyclass]`, which uses `Py::new`

https://github.com/PyO3/pyo3/blob/bcb90775b43659a58f44755d3fbaf92ac493d8c2/pyo3-derive-backend/src/pyclass.rs#L379

pyo3-derive-backend/src/pyenum.rs
gilescope: Maybe go further and have pyenum always be 'Copy' as we are doing small data here rather than ADTs?

davidhewitt: Perhaps, though I'm not sure it's our place to add `#[derive(Copy, Clone)]` automatically to the struct.

gilescope:
Given https://github.com/PyO3/pyo3/pull/1065 is merged in, do we want to also try and get this landed so they can be in 0.12 together or does that PR also make it possible to convert simple rust enums to python enums so this PR is now not needed?

davidhewitt:
#1065 enables the Python -> Rust direction, but it doesn't really solve the case of data enums like we see here (afaik).

I'm now thinking that this new macro `#[pyenum]` makes sense to land in 0.13, with `#[pyexception]`. That should also give time for this to be finished off.

gilescope:
Still keen on using this and having proper python enums! What's left to do?

hwchen:
Same here, I'm definitely looking forward to this and I'm happy to help.

davidhewitt:
I think at this point it's reasonable to assume the original PR author has unfortunately not been able to find the time to finish this off, so I guess they wouldn't mind if anyone is interested in copying this branch and reopening it as a new PR.

I think that rebasing it on master, applying the review changes I've made here, and then fixing any subsequent test failures would be a great starting point. I don't think there's particularly far to go for us to merge this!

stillinbeta:
Yeah, I've mostly moved onto other projects, sorry for not un-licking this
particular cookie sooner. Go forth with my blessing, and I'll try to answer
questions about my code if they come up.

aviramha:
Hey, I'll take this.

davidhewitt:
This is now superseded by #2002. Thanks again for kicking this off.

PR #1047, opened 2020-07-17
Functions which cast reference to pointer are not unsafe.

davidhewitt:
... it's the later usage of that pointer that's unsafe! 😄


davidhewitt:
cc @kngwyu - I just want to check you agree with me on this one? Best to have a reviewer other than me as it's changing "`unsafe`" code...

kngwyu:
Thanks!

Issue #652, opened 2019-10-26
More verbose conversion/downcasting errors.


sebpuetz:
pyo3 tries to convert arguments to the specified type in the signature, if that fails it returns `TypeError` without any message through:

~~~Rust
/// Converts `PyDowncastError` to Python `TypeError`.
impl std::convert::From<PyDowncastError> for PyErr {
    fn from(_err: PyDowncastError) -> PyErr {
        exceptions::TypeError.into()
    }
}
~~~

Since conversion happens before entering the user-defined method body, there's no way to be verbose about what happened unless some new-type is defined for which `extract()` is implemented manually or the method is changed to take `&PyAny` or `PyObject` to begin with - which also entails manual extraction.

Are there obvious downsides to changing `PyDowncastError` to wrap a String holding an error message that gets propagated into Python? 

In that case, the failed conversions could return something along these lines:

~~~Rust
PyDowncastError(format!(Can't extract some_type from {}", ob.get_type().name())
~~~

kngwyu:
Looks nice :+1: 
Thanks.

sebpuetz:
I was looking at the code generation yesterday and it would be possible to be verbose there, but not in a clean way: Either the (possibly in `extract()` user-defined) error message is lost or the generic error message would need to be extended by the one coming out of `extract()`.

Another option could be changing the conversion impls in `types` to something like:

~~~Rust
/// Allows extracting strings from Python objects.
/// Accepts Python `str` and `unicode` objects.
impl<'a> crate::FromPyObject<'a> for &'a str {
    fn extract(ob: &'a PyAny) -> PyResult<Self> {
        let s: Cow<'a, str> = crate::FromPyObject::extract(ob).map_err(|_| {
            TypeError::py_err(format!(
                "Failed to extract string from {}",
                ob.get_type().name()
            ))
        })?;
        match s {
            Cow::Borrowed(r) => Ok(r),
            Cow::Owned(r) => {
                let r = ob.py().register_any(r);
                Ok(r.as_str())
            }
        }
    }
}
~~~

As a workaround within my own project, this little helper method comes in handy:

~~~Rust
pub(crate) fn t_from_any<'a, T>(any: &'a PyAny, to: Option<&str>) -> PyResult<T>
where
    T: FromPyObject<'a>,
{
    let any_type = any.get_type().name();
    let msg = to
        .map(|to| format!("Expected '{}' not '{}'", to, any_type))
[...]

PR #1050, opened 2020-07-18
Add type info to conversion errors.

sebpuetz:
This is a draft to address #652. Initially, I tried to come up with a solution that keeps the downcast error 0-sized by parameterizing `PyDowncastError<T1, T2> where T1: PyTypeInfo, T2: PyTypeInfo` but I think there are cases where the underlying type is not easily accessible. 

I haven't worked much with pyo3 recently, so any pointers if I missed some obvious solution are much appreciated!

----------------------

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -321,10 +321,10 @@ where
 /// This trait is similar to `std::convert::TryFrom`
 pub trait PyTryFrom<'v>: Sized + PyNativeType {
     /// Cast from a concrete Python object type to PyObject.
-    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError>;
+    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
 
     /// Cast from a concrete Python object type to PyObject. With exact type check.
-    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError>;
+    fn try_from_exact<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
 
     /// Cast a PyAny to a specific type of PyObject. The caller must
     /// have already verified the reference is for this type.
@@ -358,24 +358,24 @@ impl<'v, T> PyTryFrom<'v> for T
 where
     T: PyTypeInfo + PyNativeType,
 {
-    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError> {
+    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
         let value = value.into();
         unsafe {
             if T::is_instance(value) {
                 Ok(Self::try_from_unchecked(value))
             } else {
-                Err(PyDowncastError)
+                Err(PyDowncastError::new(value, T::NAME))
             }
         }
     }
 
-    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError> {
+    fn try_from_exact<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
         let value = value.into();
         unsafe {
             if T::is_exact_instance(value) {
                 Ok(Self::try_from_unchecked(value))
             } else {
-                Err(PyDowncastError)
+                Err(PyDowncastError::new(value, T::NAME))
             }
         }
     }
@@ -390,23 +390,23 @@ impl<'v, T> PyTryFrom<'v> for PyCell<T>
 where
     T: 'v + PyClass,
 {
-    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError> {
+    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
         let value = value.into();
         unsafe {
             if T::is_instance(value) {
                 Ok(Self::try_from_unchecked(value))
             } else {
-                Err(PyDowncastError)
+                Err(PyDowncastError::new(value, T::NAME))
             }
         }
     }
-    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError> {
+    fn try_from_exact<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
         let value = value.into();
         unsafe {
             if T::is_exact_instance(value) {
                 Ok(Self::try_from_unchecked(value))
             } else {
-                Err(PyDowncastError)
+                Err(PyDowncastError::new(value, T::NAME))
             }
         }
     }
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -321,10 +321,10 @@ where
 /// This trait is similar to `std::convert::TryFrom`
 pub trait PyTryFrom<'v>: Sized + PyNativeType {
     /// Cast from a concrete Python object type to PyObject.
-    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
+    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
 
     /// Cast from a concrete Python object type to PyObject. With exact type check.
-    fn try_from_exact<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
+    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
 
     /// Cast a PyAny to a specific type of PyObject. The caller must
     /// have already verified the reference is for this type.
@@ -358,7 +358,7 @@ impl<'v, T> PyTryFrom<'v> for T
 where
     T: PyTypeInfo + PyNativeType,
 {
-    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
+    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
         let value = value.into();
         unsafe {
             if T::is_instance(value) {
@@ -369,7 +369,7 @@ where
         }
     }
 
-    fn try_from_exact<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
+    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
         let value = value.into();
         unsafe {
             if T::is_exact_instance(value) {
@@ -390,7 +390,7 @@ impl<'v, T> PyTryFrom<'v> for PyCell<T>
 where
     T: 'v + PyClass,
 {
-    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
+    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
         let value = value.into();
         unsafe {
             if T::is_instance(value) {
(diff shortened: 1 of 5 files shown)


src/err.rs
```diff
@@ -56,7 +56,29 @@ pub struct PyErr {
 pub type PyResult<T> = Result<T, PyErr>;
 
 /// Marker type that indicates an error while downcasting
-pub struct PyDowncastError;
+pub struct PyDowncastError {
+    from: Option<String>,
+    to: Option<String>,
+}
```
davidhewitt, line 62: I have a concern that creating two strings for most conversion errors may lead to a noticeable performance overhead.

How about this alternative:

```
pub struct PyDowncastError<'a> {
    from: &'a PyAny,
    to: Cow<'a, str>,
}
```

I suggest `from` becomes `&'a PyAny` because I think I'd rather see `from.repr()` rather than just the source type. That'd make it easier for me to debug out exactly what value is failing to convert.

And `from` as `Cow<'a, str>` just so we can avoid copying the string when we don't have to :)

src/err.rs
```diff
@@ -450,14 +472,25 @@ impl<'a> IntoPy<PyObject> for &'a PyErr {
...
+            }
+            (None, None) => exceptions::TypeError::py_err("Failed to convert"),
+        }
     }
 }
 
 impl<'p> std::fmt::Debug for PyDowncastError {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
-        f.write_str("PyDowncastError")
+        write!(f, "PyDowncastError from {:?} to {:?}", self.from, self.to)
```
davidhewitt, line 493: This now looks more like a `Display` impl - can you please change this impl to be for `Display` and just add `#[derive(Debug)]` to the `PyDowncastError` struct?

src/err.rs
```diff
@@ -450,14 +472,25 @@ impl<'a> IntoPy<PyObject> for &'a PyErr {
...
+                exceptions::TypeError::py_err(format!("Failed to convert from {} to {}", from, to))
+            }
+            (Some(from), None) => {
+                exceptions::TypeError::py_err(format!("Failed to convert from {}", from))
+            }
+            (None, Some(to)) => {
+                exceptions::TypeError::py_err(format!("Failed to convert to {}", to))
+            }
+            (None, None) => exceptions::TypeError::py_err("Failed to convert"),
+        }
```
davidhewitt, line 487: If you change the `PyDowncastError` struct to my suggested form (esp. without `Option`), and change `Debug` impl to `Display` as suggested, then I think that this can just re-use the `Display` implementation:

```suggestion
        exceptions::TypeError::py_err(e.to_string())
```

src/types/iterator.rs
```diff
@@ -38,7 +38,7 @@ impl<'p> PyIterator<'p> {
             // Returns NULL if an object cannot be iterated.
             if ptr.is_null() {
                 PyErr::fetch(py);
-                return Err(PyDowncastError);
+                return Err(PyDowncastError::new(None, Some("Iterator")));
```
davidhewitt, line 41: Hmm it looks like this is the only situation where you're not able to provide a source input. Reading through this function I actually think we can simplify it a lot, so I'm going to open a PR for it right now so that you don't need `PyDowncastError` at all here.

sebpuetz:
Thanks for the review! I'll amend the PR, probably some time tomorrow!

davidhewitt:
:+1: I'll try to remember to merge #1051 in the morning so that you can use that to simplify the `PyIterator` case. Feel free to ping me if I forget!

sebpuetz:
>  *  there's a second trait `FromPyObject` which has implementations for types like `Option<T>`.  It would be really cool if extraction for that `Option<T>` mentioned `Optional[int]` etc as the target type.

I spent some time trying to implement this, but I don't think it's easily doable:

```Rust
impl<'a, T> FromPyObject<'a> for Option<T>
where
    T: FromPyObject<'a>,
{
        if obj.as_ptr() == unsafe { ffi::Py_None() } {
            Ok(None)
        } else {
            match T::extract(obj) {
                Ok(v) => Ok(Some(v)),
                Err(e) => Err(e),
            }
        }
    }
}
```

assumes that `T` is any Rust type that implements `FromPyObject`, so we don't get any `PyTypeInfo` about `T`. `T::extract()` implementations know the target types of the casts, but these details aren't available on `T`. 

Unless I'm missing something, `PyErr` only makes the message available through its `pvalue`, which is already some Python type that in itself requires `extract`ing a `String`. We'd have to ensure that the error is in fact a `TypeError` and that it's a message that we generated in order to manipulate the `to` part. This seems quite messy and error prone.

In other words, I don't think we can get the nice `Option` messages without changing `extract` errors to be generic over `E: Into<PyErr>`...

davidhewitt:
Ah right, thanks for exploring the `Option` case further. I agree let's not pursue that right now - maybe I'll rework another time though as you say will be more breaking. The mini PR would be very welcome though!

I've merged #1051 so you should be unblocked here now!

src/err.rs
```diff
@@ -56,7 +57,21 @@ pub struct PyErr {
 pub type PyResult<T> = Result<T, PyErr>;
 
 /// Marker type that indicates an error while downcasting
-pub struct PyDowncastError;
+#[derive(Debug)]
+pub struct PyDowncastError<'a, 'b> {
+    from: &'a PyAny,
+    to: Cow<'b, str>,
```
kngwyu, line 63: Isn't this too complex?
I prefer to `Cow<'static str>` or `String`.

src/err.rs
```diff
@@ -56,7 +57,21 @@ pub struct PyErr {
...
+}
+
+impl<'a, 'b> PyDowncastError<'a, 'b> {
+    pub fn new(from: &'a PyAny, to: Cow<'b, str>) -> Self {
+        PyDowncastError { from, to }
+    }
+
+    pub fn set_to(&mut self, to: Cow<'b, str>) {
```
kngwyu, line 71: What is this method for?

src/err.rs
```diff
@@ -56,7 +57,21 @@ pub struct PyErr {
...
 /// Marker type that indicates an error while downcasting
-pub struct PyDowncastError;
+#[derive(Debug)]
+pub struct PyDowncastError<'a, 'b> {
+    from: &'a PyAny,
+    to: Cow<'b, str>,
+}
+
+impl<'a, 'b> PyDowncastError<'a, 'b> {
+    pub fn new(from: &'a PyAny, to: Cow<'b, str>) -> Self {
```
kngwyu, line 67: `to` should be `Into<Cow<...>>` to allow `&str` as an argument.

src/err.rs
davidhewitt: I think we should just be able to do `Cow<'a, str>` rather than needing a separate lifetime? Agreed having two lifetimes is too much. I'd rather avoid the copy into a `String` if possible!

Maybe you're right and non-copy case is always `'static`, so perhaps `Cow<'static, str>' would work. Or yeah just `&'static str` if there's now no uses where we create a new `String`.

src/err.rs
sebpuetz: That's an accidental leftover from trying to get the `Optional` info to work, I'll remove it!

src/err.rs
sebpuetz: A `String` is allocated anyways when the conversion to a `PyErr` happens. The lifetime of `to` is generally not tied to the object that we try to convert from.

If you don't like the `'b` parameter, I think @kngwyu's suggestion of setting it to `'static` is the easiest. Otherwise, I don't think the additional `String` allocation would hurt, if it was set to be a `String`.

davidhewitt: My motivation for avoiding the `String` allocation is that then users who are handling this error on the Rust side (e.g. never converting to a `TypeError`) can have as little overhead as possible. A small thing but they'll add up over time!

I'm happy with `Cow<'static, str>` here 👍

sebpuetz: Thanks for pointing out! I didn't see that `PyTryFrom` returns `PyDowncastError` instead of `PyErr`.

sebpuetz:
> If I can make one further ask, could you please add a CHANGELOG entry? I suggest:

I had already amended the previous commit with:
```
- Implement type information for conversion failures  [#1050](https://github.com/PyO3/pyo3/pull/1050)
```

I pushed out the version with the fallback to the `from` type!

davidhewitt:
Ah brilliant, thanks very much, I missed that. Once CI passes let's get this merged! :rocket:

src/conversion.rs
```diff
@@ -321,10 +321,10 @@ where
 /// This trait is similar to `std::convert::TryFrom`
 pub trait PyTryFrom<'v>: Sized + PyNativeType {
     /// Cast from a concrete Python object type to PyObject.
-    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError>;
+    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
```
davidhewitt, line 324:
```suggestion
    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
```

src/conversion.rs
```diff
@@ -321,10 +321,10 @@ where
...
 
     /// Cast from a concrete Python object type to PyObject. With exact type check.
-    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError>;
+    fn try_from_exact<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
```
davidhewitt, line 327:
```suggestion
    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>>;
```

src/conversion.rs
```diff
@@ -358,24 +358,24 @@ impl<'v, T> PyTryFrom<'v> for T
 where
     T: PyTypeInfo + PyNativeType,
 {
-    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError> {
+    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
```
davidhewitt, line 361:
```suggestion
    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
```

src/conversion.rs
```diff
@@ -358,24 +358,24 @@ impl<'v, T> PyTryFrom<'v> for T
...
                 Ok(Self::try_from_unchecked(value))
             } else {
-                Err(PyDowncastError)
+                Err(PyDowncastError::new(value, T::NAME))
             }
         }
     }
 
-    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError> {
+    fn try_from_exact<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
```
davidhewitt, line 372:
```suggestion
    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
```

src/conversion.rs
```diff
@@ -390,23 +390,23 @@ impl<'v, T> PyTryFrom<'v> for PyCell<T>
 where
     T: 'v + PyClass,
 {
-    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError> {
+    fn try_from<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
```
davidhewitt, line 393:
```suggestion
    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
```

src/conversion.rs
```diff
@@ -390,23 +390,23 @@ impl<'v, T> PyTryFrom<'v> for PyCell<T>
...
             if T::is_instance(value) {
                 Ok(Self::try_from_unchecked(value))
             } else {
-                Err(PyDowncastError)
+                Err(PyDowncastError::new(value, T::NAME))
             }
         }
     }
-    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError> {
+    fn try_from_exact<'b, V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
```
davidhewitt, line 403:
```suggestion
    fn try_from_exact<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, PyDowncastError<'v>> {
```

src/conversion.rs
sebpuetz: That should probably be a `clippy` lint, good catch!

davidhewitt: I think it is, but I decided not to wait for CI to tell us ;)

davidhewitt:
CI is happy so let's merge it! Many thanks again

src/err.rs
```diff
@@ -56,7 +57,20 @@ pub struct PyErr {
 pub type PyResult<T> = Result<T, PyErr>;
 
 /// Marker type that indicates an error while downcasting
```
gilescope, line 59: technically we should stop calling it a marker type now.

PR #558, opened 2019-07-27
Fix NULL pointer deref when creating iterator for non-iterable

danieldk:
Fixes #494.

PR #1051, opened 2020-07-18
Change PyIterator::from_object` to return underlying TypeError

davidhewitt:
Reviewing #1050 I realised that `PyIterator::from_object()` makes an unneccessary `PyIter_Check()` call as a successful result is already guaranteed to be an iterator.

I simplified it so that the function now returns the error from `PyObject_GetIter()` instead of a `PyDowncastError`. This helps the implementation of #1050 and I think also improves the experience here.

Now this function will return the well-known `TypeError: 'Foo' object is not iterable` which comes from `PyObject_GetIter()`: (https://github.com/python/cpython/blob/8182cc2e68a3c6ea5d5342fed3f1c76b0521fbc1/Objects/abstract.c#L2624)


sebpuetz:
Just commenting on this since this was mentioned in my PR: https://github.com/PyO3/pyo3/pull/558 introduced this null check because trying to iterate some non-sense would cause segfaults. Is that behaviour fixed in some other way or would this lead to a regression here?

davidhewitt:
Thanks for double-checking! `Python::from_borrowed_ptr_or_err` makes the null check inside it, so that case is still handled.

It actually happens exactly that the test I added to this PR tries to iterate an int and confirms that it results in a `TypeError` :smile:

Issue #1016, opened 2020-07-02
Bridging respective logging facilities


vorner:
Hello

I'm not sure this is in scope of the project, but I'd still want to present the idea here.

I'm wondering about a use case where I write an extension module in Rust and use it from python application. And I'd certainly like to be able to log from the rust part, or let my rust dependencies to log too. One way to do it is to log from the python and rust parts independently, but that doesn't sound exactly right.

So the other way would be to implement a logger on the Rust side that would delegate logging to the python side (I do see some possible issues with that approach ‒ python doesn't seem to have trace level, the performance might be a problem).

And my question is, if I give it a try and something usable comes of it:
* Would other people be interested in having this functionality?
* Would someone with more experience of using `pyo3` give it a look/review?
* How much closely related to `pyo3` should this be? Should it be a completely separate crate, with separate git repo and possibly mentioned as useful in some guide of `pyo3` (is there a place for such „handy“ crates for various tasks like this?)? Or should it live under the same repo so it can easily follow the same release cycle? Or even be a feature-flag directly in the crate?

I guess one could also want the bridge in other direction (eg. python logging piping stuff to the rust side), but I don't currently have such use case myself and haven't thought much about how to implement it; I have some about this direction.

(I assume there isn't such thing yet, I haven't found it but maybe I haven't looked well enough)

davidhewitt:
I'd be happy to help review it. For now I wonder if it makes sense to be a separate crate along the lines of `android_logger` and `console_log`?

Possible names to me include `python_logger`, or `pyo3_logger` if you want to take the pyo3 dependency explicit.

As to marketing this crate, I think it would be worth mentioning in the pyo3 README.

vorner:
Thank you for the offer. I've put something together. It feels more or less ready for release (the test coverage could be better and I'll have to figure out how to test such a thing anyway). But if you could have a look if there's something obviously broken or could be done better, I'd be glad. https://github.com/vorner/pyo3-log

> As to marketing this crate, I think it would be worth mentioning in the pyo3 README.

That would be nice :-). I was also thinking of adding an example somewhere into the „Advanced topics“ or adding another appendix to the guide, if that would be fine.

davidhewitt:
Thanks - the crate looks nice! A collection of random thoughts:

The caching stuff looks very advanced. I guess because log configuration is typically global and done once at program startup, this is why you are caching the Python log hierarchy?

[...]

PR #1052, opened 2020-07-19
docs: Link to pyo3-log in README and the guide

vorner:
Show how one can log from Rust through the Python's logging facilities.

Related to #1016.

I've tried running `mdbook test`, as it is more relevant to the changes I did than the suggested commands in contributing guidelines. However, the examples fail to compile throughout the whole book, so it seems they are not being checked :-( and I didn't find a place to mention the code examples need to depend on additional package.

### DIFF
--- a/guide/src/logging.md
+++ b/guide/src/logging.md
@@ -1,0 +1,68 @@
+# Logging
+
+It is desirable if both the Python and Rust parts of the application end up
+logging using the same configuration into the same place.
+
+This section of the guide briefly discusses how to connect the two languages'
+logging ecosystems together. The recommended way for Python extension modules is
+to configure Rust's logger to send log messages to Python using the `pyo3-log`
+crate. For users who want to do the opposite and send Python log messages to
+Rust, see the note at the end of this guide.
+
+## Using `pyo3-log` to send Rust log messages to Python
+
+The [pyo3-log] crate allows sending the messages from the Rust side to Python's
+[logging] system. This is mostly suitable for writing native extensions for
+Python programs.
+
+Use [`pyo3_log::init`][init] to install the logger in its default configuration.
+It's also possible to tweak its configuration (mostly to tune its performance).
+
+```rust
+use log::info;
+use pyo3::prelude::*;
+use pyo3::wrap_pyfunction;
+
+#[pyfunction]
+fn log_something() {
+    // This will use the logger installed in `my_module` to send the `info`
+    // message to the Python logging facilities.
+    info!("Something!");
+}
+
+#[pymodule]
+fn my_module(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
+    // A good place to install the Rust -> Python logger.
+    pyo3_log::init();
+
+    m.add_wrapped(wrap_pyfunction!(log_something))?;
+    Ok(())
+}
+```
+
+Then it is up to the Python side to actually output the messages somewhere.
+
+```python
+import logging
+import my_module
+
+FORMAT = '%(levelname)s %(name)s %(asctime)-15s %(filename)s:%(lineno)d %(message)s'
+logging.basicConfig(format=FORMAT)
+logging.getLogger().setLevel(logging.INFO)
+my_module.log_something()
+```
+
+It is important to initialize the Python loggers first, before calling any Rust
+functions that may log. This limitation can be worked around if it is not
+possible to satisfy, read the documentation about [caching].
+
+## The Python to Rust direction
+
+To best of our knowledge nobody implemented the reverse direction yet, though it
+should be possible. If interested, the `pyo3` community would be happy to
+provide guidance.
+
+[logging]: https://docs.python.org/3/library/logging.html
+[pyo3-log]: https://crates.io/crates/pyo3-log
+[init]: https://docs.rs/pyo3-log/*/pyo3_log/fn.init.html
+[caching]: https://docs.rs/pyo3-log/*/pyo3_log/#performance-filtering-and-caching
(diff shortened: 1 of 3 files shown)


guide/src/logging.md
```diff
@@ -0,0 +1,60 @@
...
+
+FORMAT = '%(levelname)s %(name)s %(asctime)-15s %(filename)s:%(lineno)d %(message)s'
+logging.basicConfig(format=FORMAT)
+logging.getLogger().setLevel(logging.INFO)
+my_module.log_something()
+```
+
+It is important to initialize the Python loggers first, before calling any Rust
+functions that may log. This limitation can be worked around if it is not
+possible to satisfy, read the documentation about [caching]
```
davidhewitt, line 55:
```suggestion
possible to satisfy, read the documentation about [caching].
```

guide/src/logging.md
```diff
@@ -0,0 +1,60 @@
...
+It is desirable if both the Python and Rust parts of the application end up
+logging using the same configuration into the same place.
+
+This can trivially be accomplished by not logging at all in one of them. If
+that's not a good enough solution, it is possible to install a logger that sends
+the log messages to the counterpart.
+
+# The Rust -> Python direction
+
+The [pyo3-log] allows sending the messages from the Rust side to Python's
```
davidhewitt, line 12:
```suggestion
The [pyo3-log] crate allows sending the messages from the Rust side to Python's
```

guide/src/logging.md
```diff
@@ -0,0 +1,60 @@
...
+[logging] system. This is mostly suitable for writing native extensions for
+Python programs.
+
+The [`pyo3_log::init`][init] installs the logger in its default configuration.
```
davidhewitt, line 16:
```suggestion
Use [`pyo3_log::init`][init] to install the logger in its default configuration.
```

guide/src/logging.md
```diff
@@ -0,0 +1,60 @@
...
+It's also possible to tweak it's configuration (mostly to tune its performance).
```
davidhewitt, line 17:
```suggestion
It's also possible to tweak its configuration (mostly to tune its performance).
```

guide/src/logging.md
```diff
@@ -0,0 +1,60 @@
+# Logging
+
+It is desirable if both the Python and Rust parts of the application end up
+logging using the same configuration into the same place.
+
+This can trivially be accomplished by not logging at all in one of them. If
+that's not a good enough solution, it is possible to install a logger that sends
+the log messages to the counterpart.
+
+# The Rust -> Python direction
```
davidhewitt, line 10: I wonder if this title would be even stronger if it said what the reader will learn in this section:

```suggestion
## Using `pyo3-log` to send Rust log messages to Python
```

Also I would prefer only have one top level `#` heading so I added an extra `#` in the suggestion.

guide/src/logging.md
```diff
@@ -0,0 +1,60 @@
...
+logging.basicConfig(format=FORMAT)
+logging.getLogger().setLevel(logging.INFO)
+my_module.log_something()
+```
+
+It is important to initialize the Python loggers first, before calling any Rust
+functions that may log. This limitation can be worked around if it is not
+possible to satisfy, read the documentation about [caching]
+
+[logging]: https://docs.python.org/3/library/logging.html
```
davidhewitt, line 57: How about maybe adding a very short section about the `Python -> Rust` direction - that nobody has built a solution yet and if would be awesome if anyone wants to contribute a PR / crate for that (with design advice available from the PyO3 community)?

guide/src/logging.md
```diff
@@ -0,0 +1,60 @@
+# Logging
+
+It is desirable if both the Python and Rust parts of the application end up
+logging using the same configuration into the same place.
+
+This can trivially be accomplished by not logging at all in one of them. If
```
davidhewitt, line 6: I think this sentence doesnt add so much - users presumably come here because they want to do something rather than not log at all :smile: 

I'm tempted to suggest this paragraph gets replaced with a new one that describes what the rest of this chapter will be, e.g.

```suggestion
This section of the guide briefly discusses how to connect the two languages' logging ecosystems together. The recommended way for Python extension modules is to configure Rust's logger to send log messages to Python using the `pyo3-log` crate. For users who want to do the opposite and send Python log messages to Rust, see the note at the end of this guide.
```

This will help readers find what section of this chapter could be relevant to them by just scanning the introduction.
 
Of course please feel free to change what's written here it you don't like this suggestion!

guide/src/logging.md
```diff
@@ -0,0 +1,60 @@
...
+fn log_something() {
+    // This will use the logger installed in `my_module` to send the `info`
+    // message to the Python logging facilities.
+    info!("Something!");
+}
+
+#[pymodule]
+fn my_module(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
+    // A good place to install the Rust -> Python logger.
+    pyo3_log::init();
```
kngwyu, line 34: Nice interface :+1:

vorner:
I've adopted the suggestions, as a fixup commit. Once satisfied, please tell me to rebuild the history in the proper way before merging.

> We actually test these examples through cargo test - you can add the new file to the doctests in lib.rs: 

I tried that. The downside is, if I add the `pyo3-log` as a dependency, it also brings the version on crates.io in as a dependency, so pyo3-log is twice in the dep graph (the git version and crates.io version). If this is deemed too heavy for the check, it is in a separate commit so it can be easily removed from the branch.

davidhewitt:
This looks great, thanks!

> pyo3-log is twice in the dep graph 

I think you mean `pyo3` is twice in the dep graph? Yeah this is a little unfortunate. How about making the `pyo3-log` dev-dependency optional and adding `#[cfg(pyo3-log)]` to the doctest? And then finally, adding `pyo3-log` to the features list in `ci/actions/test.sh` and `ci/actions/travis.sh`?

I don't think there's so much code in this guide that it's likely to be a frequent frustration if we typically only test on ci...

> please tell me to rebuild the history in the proper way before merging

We're not super strict in this crate, though if you want to squash to a single commit I won't mind ;)

vorner:
> How about making the pyo3-log dev-dependency optional and adding #[cfg(pyo3-log)] to the doctest?

Dev-dependencies are not allowed to be optional :-(.

Furthermore, apparently pyo3-log needs newer rust than pyo3, which breaks the CI. So I'd say I'll exclude this test again, at least for now.

> We're not super strict in this crate, though if you want to squash to a single commit I won't mind ;)

Actually, I wanted to sort & do two commits, just squashing the fixup to the right one, but as I'm going to kill the doctests, it'll end up being just one commit. And it's me being pedantic about git history 😇.

davidhewitt:
Yep that makes sense.

In which case I think this is now ready to merge! Thanks very much for working on this.

Issue #1022, opened 2020-07-04
Segmentation fault before method call with dict and unsendable


konstin:
A `#[pyclass(dict, unsendable)]` crashes with a segmentation fault when trying to call a method. 

### 🌍 Environment

 - Your operating system and version: Ubuntu 18.04
 - Your python version: 3.7
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: pyenv and virtualenv
 - Your Rust version (`rustc --version`): rustc 1.44.1 (c7087fe00 2020-06-17)
 - Your PyO3 version: 0.11.1
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: Yes

### 💥 Reproducing

lib.rs

```rust
use pyo3::prelude::*;

#[pyclass(dict, unsendable)] // <- NOTE THESE TWO
struct BeautifulSoup {}

#[pymethods]
impl BeautifulSoup {
    #[new]
    fn new() -> PyResult<BeautifulSoup> {
        println!("new called");
        Ok(BeautifulSoup {})
    }

    fn select_one(&self) -> Option<()> {
        println!("select_one called");
        Some(())
    }
}

#[pymodule]
fn html_py_ever_repo(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<BeautifulSoup>()?;

    Ok(())
}
```

Cargo.toml

```toml
[package]
name = "html-py-ever-repo"
version = "0.1.0"
authors = ["konstin <konstin@mailbox.org>"]
edition = "2018"

[dependencies]
#pyo3 = { version = "0.11.1", features = ["extension-module"] }
pyo3 = { git = "https://github.com/PyO3/pyo3", features = ["extension-module"] }

[lib]
name = "html_py_ever_repo"
crate-type = ["cdylib"]
```

Crashing python code

```python
from html_py_ever_repo import BeautifulSoup
BeautifulSoup().select_one()
```

backtrace

```
#0  _PyObject_GetMethod (obj=obj@entry=0x7ffff7e516c0, name=0x7ffff7e65070, method=method@entry=0x7fffffffd770) at Objects/object.c:1165
#1  0x00005555555b1075 in _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at Python/ceval.c:3040
[...]

PR #1058, opened 2020-07-20
Move the ThreadChecker field in front of dict and weakref.

sebpuetz:
Addresses #1022

Changing the order doesn't seem to cause any other issues. If changing the order seems risky, we can also calculate the initial offsets based on `offset -= std::mem::size_of::<T::ThreadChecker>() as isize;`

----------------
Offsets for dict and weakref are calculated from the end of the PyCell struct. When using the non-dummy ThreadChecker, the offsets were invalid since the `ThreadCheckerImpl` is not zero-sized.

----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Conversion from types with an `__index__` method to Rust BigInts. [#1027](https://github.com/PyO3/pyo3/pull/1027)
+- Fix segfault with #[pyclass(dict, unsendable)] [#1058](https://github.com/PyO3/pyo3/pull/1058)
 
 ## [0.11.1] - 2020-06-30
 ### Added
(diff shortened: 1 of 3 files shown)


sebpuetz:
I keep forgetting about the changelog! Thanks for the reminder.

The amended commit has a (hopefully correct) test case and the changelog entry.

kngwyu:
@sebpuetz Thanks!
@davidhewitt Thank you for your review, but could you please wait for my review if a PR is related to what I solely wrote ?
In such cases, I want to understand _what I misunderstood_.

davidhewitt:
> but could you please wait for my review if a PR is related to what I solely wrote ?

:+1: apologies for going too fast here!

sebpuetz:
Since there's still some life in here: 

It probably would be nice to add an in-line comment on `PyCell` mentioning that the dict and weakref fields must be at the end of the struct. I'm also wondering if there is a cleaner solution for this that doesn't depend on the order of struct fields in one place and subtracting the correct offsets in a **different** place.

kngwyu:
>  I'm also wondering if there is a cleaner solution for this that doesn't depend on the order of struct fields 

I think it's difficult since we don't have the concrete `PyCell` at that time, though we can do some cleanups around that.

PR #1060, opened 2020-07-21
Calculate offsets for weakreflist and dict in PyCell.

sebpuetz:
As demonstrated by #1022, depending on the order of structmembers to calculate offsets in bytes is fragile. ~~This PR adds a dependency to `memoffset` which provides an interface to find offsets into structs in a more robust manner.~~

~~Caveat: This macro is technically unsound, see https://github.com/Gilnaa/memoffset/issues/24, so I'm marking this as a draft for now.~~

This PR moves the calcualtion of class dict and weakreflist offsets to `PyCell` rather than `initialize_type_object`.

-----------------

### DIFF
--- a/src/pyclass.rs
+++ b/src/pyclass.rs
@@ -137,18 +137,14 @@ where
     // type size
     type_object.tp_basicsize = std::mem::size_of::<T::Layout>() as ffi::Py_ssize_t;
 
-    let mut offset = type_object.tp_basicsize;
-
     // __dict__ support
-    if let Some(dict_offset) = T::Dict::OFFSET {
-        offset += dict_offset as ffi::Py_ssize_t;
-        type_object.tp_dictoffset = offset;
+    if let Some(dict_offset) = PyCell::<T>::dict_offset() {
+        type_object.tp_dictoffset = dict_offset as ffi::Py_ssize_t;
     }
 
     // weakref support
-    if let Some(weakref_offset) = T::WeakRef::OFFSET {
-        offset += weakref_offset as ffi::Py_ssize_t;
-        type_object.tp_weaklistoffset = offset;
+    if let Some(weakref_offset) = PyCell::<T>::weakref_offset() {
+        type_object.tp_weaklistoffset = weakref_offset as ffi::Py_ssize_t;
     }
 
     // GC support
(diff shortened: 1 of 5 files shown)


sebpuetz:
An alternative implementation that makes maintenance easier without pulling in unsoundness:

~~~Rust
#[repr(C)]
pub struct PyCell<T: PyClass> {
    inner: PyCellInner<T>,
    thread_checker: T::ThreadChecker,
    // DO NOT CHANGE THE ORDER OF THESE FIELDS WITHOUT CHANGING PyCell::dict_offset()
    // AND PyCell::weakref_offset()
    dict: T::Dict,
    weakref: T::WeakRef,
}

impl<T: PyClass> PyCell<T> {
    /// Get the offset of the dictionary from the start of the struct in bytes.
    pub(crate) fn dict_offset() -> usize {
        mem::size_of::<Self>() - mem::size_of::<T::Dict>() - mem::size_of::<T::WeakRef>()
    }

    /// Get the offset of the weakref list from the start of the struct in bytes.
    pub(crate) fn weakref_offset() -> usize {
        mem::size_of::<Self>() - mem::size_of::<T::WeakRef>()
    }
}
~~~

sebpuetz:
@kngwyu

src/pyclass.rs
```diff
@@ -137,18 +137,14 @@ where
     // type size
     type_object.tp_basicsize = std::mem::size_of::<T::Layout>() as ffi::Py_ssize_t;
 
-    let mut offset = type_object.tp_basicsize;
-
     // __dict__ support
-    if let Some(dict_offset) = T::Dict::OFFSET {
-        offset += dict_offset as ffi::Py_ssize_t;
-        type_object.tp_dictoffset = offset;
+    if !T::Dict::IS_DUMMY {
```
kngwyu, line 141: How about moving this if statement to the `PyCell`?
I don't like that `dict_offset` can return a meaningless value even if it doesn't have a dict.

kngwyu:
Thanks!

src/pyclass.rs
sebpuetz: Good catch, I `PyCell::*_offset` now returns `Option<usize>`

sebpuetz:
Do you want to add a commit with the test case you wrote for #1061?

kngwyu:
> Do you want to add a commit with the test case you wrote for #1061?

Yeah, could you please copy-paste it to this branch?

sebpuetz:
Should be good to go if CI passes

kngwyu:
Thanks!

PR #1061, opened 2020-07-21
MAINT: Move offset calculation code to PyCell

kngwyu:
Follow up of #1058. Make it cleaner and correct (current code has a bug in that it assumes `weakref, dict` order).

### DIFF
--- a/src/pycell.rs
+++ b/src/pycell.rs
@@ -361,6 +361,20 @@ impl<T: PyClass> PyCell<T> {
         (*self_).thread_checker = T::ThreadChecker::new();
         Ok(self_)
     }
+
+    /// Calculate the offset of the `dict` from the end of the struct.
+    /// E.g., in 64bit devices, offset is -4 (without weakref) or -8 (without weakref).
+    /// Please take care that the offset depends on the order of the members.
+    pub(crate) fn dict_offset() -> Option<isize> {
+        T::Dict::OFFSET.map(|o| o + T::WeakRef::OFFSET.unwrap_or(0))
+    }
+
+    /// Calculate the offset of the `weakref` from the end of the struct.
+    /// E.g., in 64bit devices, offset is -4.
+    /// Please take care that the offset depends on the order of the members.
+    pub(crate) fn weakref_offset() -> Option<isize> {
+        T::WeakRef::OFFSET
+    }
 }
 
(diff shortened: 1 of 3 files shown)


sebpuetz:
I had just proposed something similar over at #1060, returning the offset from the start of the struct instead of the end eliminates some additional calculations, too.

~~~Rust
#[repr(C)]
pub struct PyCell<T: PyClass> {
    inner: PyCellInner<T>,
    thread_checker: T::ThreadChecker,
    // DO NOT CHANGE THE ORDER OF THESE FIELDS WITHOUT CHANGING PyCell::dict_offset()
    // AND PyCell::weakref_offset()
    dict: T::Dict,
    weakref: T::WeakRef,
}

impl<T: PyClass> PyCell<T> {
    /// Get the offset of the dictionary from the start of the struct in bytes.
    pub(crate) fn dict_offset() -> usize {
        mem::size_of::<Self>() - mem::size_of::<T::Dict>() - mem::size_of::<T::WeakRef>()
    }

    /// Get the offset of the weakref list from the start of the struct in bytes.
    pub(crate) fn weakref_offset() -> usize {
        mem::size_of::<Self>() - mem::size_of::<T::WeakRef>()
    }
}
~~~

and

~~~Rust
    // __dict__ support
    if !T::Dict::IS_DUMMY {
        type_object.tp_dictoffset = PyCell::<T>::dict_offset() as ffi::Py_ssize_t;
    }

    // weakref support
    if !T::WeakRef::IS_DUMMY {
        type_object.tp_weaklistoffset = PyCell::<T>::weakref_offset() as ffi::Py_ssize_t;
    }
~~~

kngwyu:
Oops, I'm sorry I overlooked that.

>  returning the offset from the start of the struct instead of the end eliminates some additional calculations, too.

The offset should be zero if the type doesn't have dict or weakref slots, so I think rerunning `Option` is more convenient in that it doesn't require `if offset != type_object.tp_basicsize` or so.

sebpuetz:
As far as I can tell, my proposed solution doesn't need that if condition either? `offset` is eliminated from `initialize_type_object` entirely in that. I'll amend the commit over at #1060, should make it easier to discuss then!

kngwyu:
Closed in favor of #1060

Issue #1055, opened 2020-07-20
Function argument error handling improvements


davidhewitt:
This is a sketch of ideas for a follow-up of #1050 to enable `#[pyfunctions]` to report their error type.

Imagine that the following `#[pyfunction]`:

```rust
#[pyfunction]
fn foo(tup: (i32, &str)) {
    println!("{:?}", tup)
}
```

After #1050, the error message when calling this function with invalid arguments is already much improved from PyO3 `0.11.1`:

```python
>>> foo(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't convert None to PyTuple
```

I think that this error message can yet be improved further. There's two ways in which I'd like to see this improve:

1. **Mention the argument which failed to convert.**
   Naming the function argument which failed to convert will help users locate exactly which function input is incorrect.

2. **Provide mypy-style type annotations for the target type, where possible.**
   It's great to see that `None` can't convert to `PyTuple`, but as a Python user really what I need to see is that the expected type is `Tuple[int, str]`.

Put together, this might mean that our `foo` function above would output error messages like the below:

```python
>>> foo(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't convert None to `Tuple[int, str]` (when trying to extract argument `tup`)
```

I'm not super attached to the exact wording of the above if someone else has a better idea / something easier to implement.

Later today I'm going to add some thoughts on how these two pieces could be implented if anyone in the community is interested in championing these improvements.

sebpuetz:
I was looking into this and I think the biggest issue is that the extraction methods in the proc macros return `PyResult` which is already the opaque Python error that we can't easily manipulate.  

```Rust
if spec.is_args(&name) {
    return quote! {
        let #arg_name = <#ty as pyo3::FromPyObject>::extract(_args.as_ref())?;
    };
}
```

generates code using `FromPyObject::extract` which fails with `PyErr` that's propagated into Python. 

I'm not sure how to proceed from this point without touching error-handling in `FromPyObject` which has more failure conditions than just downcasting errors.

birkenfeld:
One way would be to use a custom `TypeError` subclass that can be recognized in error-handling code and therefore have its message easily amended.

sebpuetz:
[...]

Issue #301, opened 2018-12-05
Accepting Union types for function arguments 


mehcode:
I'd like to define a `#[pyfunction]` that accepts either a String or some custom `#[pyclass]`.

davidhewitt:
Renaming to mention Python's `Union` type - which is what this issue is really talking about integration for imo.

birkenfeld:
I think this is a hopelessly complex topic and you're better off taking a `&PyAny` and doing the extraction yourself.

(Stdlib functions implemented C do the same.)

davidhewitt:
I agree that this is a complete moonshot and unlikely to be supported for a long long time! If rust had variadic generics a la C++, this would be easy ;)

sebpuetz:
You can probably emulate this behaviour to the point where it's fairly similar to Python where you'd often need something like `isinstance. Say you want a method to take both `Union[str, int]`, then you could implement some wrapper enum that you implement extraction logic for:

~~~Rust
use pyo3::prelude::*;
use pyo3::types::{PyString, PyLong};
use pyo3::{PyTypeInfo, exceptions, wrap_pyfunction};

#[pyfunction]
pub fn bar(foo: Union) {
    match foo {
        Union::Str(s) => println!("{}", s),
        Union::Int(i) => println!("{}", i),
    }
}

pub enum Union {
    Str(String),
    Int(isize),
}

impl<'source> FromPyObject<'source> for Union {
    fn extract(ob: &'source PyAny) -> PyResult<Self> {
        if let Ok(s) = PyString::try_from(ob) {
            return Ok(Union::Str(s.to_string()))
        } else if let Ok(i) = PyLong::try_from(ob) {
            return Ok(Union::Int(i.extract()?))
        } else {
            Err(exceptions::PyTypeError::py_err(format!("Can't convert from {} to Union[str, int]", ob.get_type().name())))
        }

    }
}

#[pymodule]
fn foo(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
    m.add_wrapped(wrap_pyfunction!(bar))?;
    Ok(())
}
~~~

For Python there's essentially no difference to using a Python `def`'d function dealing with both int and str. 

~~~Python
>>> import foo
>>> foo.bar("Test")
Test
>>> foo.bar(1)
1
>>> foo.bar(None)
TypeError: Can't convert from NoneType to Union[str, int]
~~~

E.g. in pure Python something like this would probably be used:

~~~Python
[...]

Issue #884, opened 2020-05-01
How to implement PyDateTime From<chrono::DateTime>


milesgranger:
## :question: Question

What is the best way to do the following?
```rust
use chrono::DateTime;
use pyo3::types::PyDateTime;

let chrono_datetime = DateTime::Utc::now();
let pydatetime: PyDateTime = chrono_datetime.into();
```
This obviously doesn't work now, and currently doing silly things to get the same effect. Most the the problem is getting [`chrono::offset::Utc`](https://docs.rs/chrono/0.4.11/chrono/offset/struct.Utc.html) into a [`PyTzInfo`](https://docs.rs/pyo3/0.9.2/pyo3/types/struct.PyTzInfo.html) for [`PyDateTime::from_timestamp`](https://docs.rs/pyo3/0.9.2/pyo3/types/struct.PyDateTime.html#method.from_timestamp)

There must be a better way. If not, I'd be happy to provide a PR (with a `chrono` feature flag(?)) with some pointers on where to start.

davidhewitt:
Hi @milesgranger, good question!

This would absolutely be welcome as a PR - you would have to implement `FromPyObject`, `IntoPy<PyObject>` and `ToPyObject` for the chrono types. It would be great to have this behind a feature flag.

There's a few recent PRs which added similar conversions for `HashSet` - you could look at those (or just the implementation in the code) if you want some templates to get started.

Any questions just pop them on here or open a WIP review and I can comment directly on the code.

kngwyu:
@milesgranger 
Thank you for your interest.
Since `chrono` is a major library, I think it's worth supporting.
We have some conditional implementation for conversion.
For example, please take a glance at [complex coversion](https://github.com/PyO3/pyo3/blob/master/src/types/complex.rs#L133).

pganssle:
I guess I can begrudgingly accept why you'd want to do this, but is there any way this can be a third-party library providing this trait instead, or does it have to be provided in one of `chrono` or `pyo3` (I can never remember the rule on this).

Perhaps irrationally, I oppose this because in my experience `chrono` is a terrible datetime library, and I'm shocked that people seem to have mostly standardized on it. Perhaps it's improved since I last used it a year or two ago.

[...]

Issue #566, opened 2019-08-06
Crazy idea: serde support


birkenfeld:
A few times now, I had to take in complex-ish Python data and convert it into an equivalent in Rust.  `FromPyObject` already does the basic types, but I'd also like to convert e.g. nested dicts with certain keys into structs.

Do you think it would make sense to let `serde` handle this? It already has all the options you'd like, e.g. flattening of sub-structure and renaming of things.  For structs, it should support both dicts (taking keys) and other objects (taking attributes).

The "ser" direction could also be supported, although I don't need it, and it would probably only construct dicts for structs.

ohadravid:
> The "ser" direction could also be supported, although I don't need it, and it would probably only construct dicts for structs.

We actually wanted something like this, mostly when you want to return something which is 'data', but don't want (or can't) create a dedicated python wrapper for each struct.

pganssle:
@birkenfeld I think that this sort of thing is a good idea, but is it really something that belongs in PyO3?

I think it makes sense to leave the core C API wrappers in PyO3 and break out useful additional functionality into their own crates.

jmrgibson:
I took a stab at something like this, by implementing the custom serde traits, but then ran into lifetime issues with all the python data since everything relies on the interpreter lifetime. Not that it probably isn't possible, but It'll take someone with more experience than me.

I ended up just dumping everything to a raw json string and passing it over, which is a trivial amount of code on both python and rust sides, but not an elegant solution.

gperinazzo:
I've made [a crate](https://github.com/gperinazzo/dict-derive) to help deal with this. It let's you derive the necessary traits for PyO3 to convert a dict into your struct, as long as it knows how to convert every field.

I plan on having serde-like field attributes in the future, but if you're not using them the base functionality should beat passing things around as json.

m3047:
I would encourage you all not to overthink this. So allow me to step back and do exactly that at a more abstract level I guess... :-/

**Who is your audience?** Is it people who are primarily writing in _Rust_ or _Python_?

**What are they struggling with?** Are they struggling with learning one language or the other, or learning how to call something from _Python_ using e.g. `ctypes` or have they stumbled on `pyo3` and are they wondering how that works?

[...]

PR #1065, opened 2020-07-23
FromPyObject derivation for structs and enums

sebpuetz:
This is a draft addressing a bit of both topics in  #301 and #1055.

I have not written proc-macro code before, so this might be quite ugly. This is mostly proof-of-concept, but I think it'd be a great quality-of-life improvement if something like this was available.

This produces the following behaviour:

~~~Python
>>> union.foo(["Test"])
["Test"]
>>> union.foo("Test")
Test
>>> union.foo(1)
1
>>> union.foo(None)
TypeError: Can't convert NoneType to Union[Str, Int, StringList]
~~~

This implementation covers both the cheap downcast variant and extraction of proper Rust types. Although I'm not sure whether relying on the variant holding a reference is the proper heuristic to apply here.

--------

Todos are marked in the code, a proper implementation should in my opinion:

* make the reported names of the variants in the error configurable through an attribute on the variant, i.e.:
    ~~~Rust
    #[union]
    pub enum MyUnion {
        #[rename(str)]
        SomeVariant(String)
    } 
    ~~~
* support named fields & variants with multiple fields


examples/union/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+use pyo3::prelude::*;
```
kngwyu, line 1: This should be moved under `tests`.

kngwyu:
Thank you, but...
Should this go this way?

Some complaints about the `#[union]`
- The conversion result depends on the order of the enum elements.
  E.g., we get  different result from `#[union] enum<'a> { A(&'a PyAny), B(&'a PyDict) }` and `#[union] enum<'a> { B(&'a PyDict), A(&'a PyAny) }`.
- It's not clear what the `union` means for who don't use `typing` usually.
- It's not full-featured attribute. Just `derive`.

So, how about just making it `#[derive(FromPyObject)]`?

sebpuetz:
Thanks for the feedback. As mentioned, this is just a rough draft to figure out whether this is the right path or not.

> Thank you, but...
> Should this go this way?
> 
> Some complaints about the `#[union]`
> 
> * The conversion result depends on the order of the enum elements.
>   E.g., we get  different result from `#[union] enum<'a> { A(&'a PyAny), B(&'a PyDict) }` and
> `#[union] enum<'a> { B(&'a PyDict), A(&'a PyAny) }`.

That's a valid point although I'm not sure how to address this or whether this needs to be addressed. 

> * It's not clear what the `union` means for who don't use `typing` usually.

I chose that name since it's the most specific term within Python, also it was the one discussed over at #301. Different naming would be fine, too. 

> * It's not full-featured attribute. Just `derive`.
> So, how about just making it `#[derive(FromPyObject)]`?

Sure, at this point that's all that the macro is doing.

-------

Before proceeding, we should probably discuss whether deriving `FromPyObject` is something that should be offered for enums and whether dependence on order is an issue.

davidhewitt:
Thanks for this.

I think @kngwyu is right that because `#[union]` is just deriving `FromPyObject`, it should really be `#[derive(FromPyObject)]`.

_However_, I think that if we proceed to add this `#[derive]` macro to `PyO3`, we should probably consider also offering it for structs in the same way that the `dict-derive` crate does. And in addition, we should aim to provide `#[derive]` for `IntoPyObject` in the same way.

Order dependence - I think this is a desirable feature and should be documented. e.g. consider this:

```
#[pyclass] struct Base { }
#[pyclass(extends = "Base")] struct Child { };

#[derive(FromPyObject)]
enum ChildOrBase<'a> {
    Child(PyRef<'a, Child>),
    Base(PyRef<'a, Base>),
}
```

In the above, `ChildOrBase::Base` can always be extracted from any input value that would also successfully extract the `Child` variant. So I think the only sensible strategy for this example to work is to state that variants are tested in order.

kngwyu:
And another question is how should we distinguish this from the enum type introduced in #1045

davidhewitt:
IMO that's a question for documentation and easy enough for us to solve:

- if you want to just convert Rust data back and forth with Python, use `#[derive(FromPyObject, IntoPy)]`

- if you want to share data between Rust and Python, use `#[pyclass]` or `#[pyenum]`

sebpuetz:
I've been working a bit on this on and off:

Should the `FromPyObject` impl handle structs / variants with multiple fields? E.g., `struct AllTheStrings<'a>(String, &'a PyString, &'a PyAny)` could all be extracted from the same object. My current (local) implementation handles all that but I'm unsure whether we want this. What do you think?

Wrt. unit structs / variants I have a check that rules those out since it wouldn't be possible to decide what to extract, or rather, any unit struct can always be extracted.

~~~Rust
pub struct A<'a>
{
    s: String,
    t: &'a PyString,
    p: &'a PyAny,
}

//generated
impl<'a> ::pyo3::FromPyObject<'a> for A<'a> {
    fn extract(ob: &'a ::pyo3::PyAny) -> ::pyo3::PyResult<Self> {
        Ok(Self {
            s: ::pyo3::FromPyObject::extract(ob)?,
            t: ::pyo3::FromPyObject::extract(ob)?,
            p: ::pyo3::FromPyObject::extract(ob)?,
        })
    }
}
~~~

for structs and for enums:

~~~Rust
pub enum Union<'a> {
    #[err_rename = "str"]
    Named {
        a: B,
        b: String,
        c: &'a PyString,
    },
    Int(&'a PyLong),
    #[err_rename = "List[str]"]
    StringList(Vec<String>),
}

// generated
impl<'a> ::pyo3::FromPyObject<'a> for Union<'a> {
    fn extract(ob: &'a ::pyo3::PyAny) -> ::pyo3::PyResult<Self> {
        if let (Ok(a), Ok(b), Ok(c)) = (
            ::pyo3::FromPyObject::extract(ob),
            ::pyo3::FromPyObject::extract(ob),
            ::pyo3::FromPyObject::extract(ob),
        ) {
            return Ok(Union::Named { a, b, c });
        };
        if let (Ok(_field_0)) = (::pyo3::FromPyObject::extract(ob)) {
            return Ok(Union::Int(_field_0));
        };
        if let (Ok(_field_0)) = (::pyo3::FromPyObject::extract(ob)) {
            return Ok(Union::StringList(_field_0));
        };
        let type_name = ob.get_type().name();
        let from = ob
            .repr()
            .map(|s| {
                let res = ::alloc::fmt::format(::core::fmt::Arguments::new_v1(
                    &["", " (", ")"],
                    &match (&s.to_string_lossy(), &type_name) {
                        (arg0, arg1) => [
                            ::core::fmt::ArgumentV1::new(arg0, ::core::fmt::Display::fmt),
                            ::core::fmt::ArgumentV1::new(arg1, ::core::fmt::Display::fmt),
                        ],
                    },
                ));
                res
            })
            .unwrap_or_else(|_| type_name.to_string());
        let err_msg = {
            let res = ::alloc::fmt::format(::core::fmt::Arguments::new_v1(
                &["Can\'t convert ", " to "],
                &match (&from, &"str, Int, List[str]") {
                    (arg0, arg1) => [
                        ::core::fmt::ArgumentV1::new(arg0, ::core::fmt::Display::fmt),
                        ::core::fmt::ArgumentV1::new(arg1, ::core::fmt::Display::fmt),
                    ],
                },
            ));
            res
        };
        Err(::pyo3::exceptions::PyTypeError::py_err(err_msg))
    }
}
~~~

davidhewitt:
> Should the FromPyObject impl handle structs / variants with multiple fields? E.g., struct AllTheStrings<'a>(String, &'a PyString, &'a PyAny) could all be extracted from the same object. My current (local) implementation handles all that but I'm unsure whether we want this. What do you think?

It seems reasonable to me that the Python "equivalent" of `AllTheThings<'a>` is `Tuple[str, str, Any]`, so I'm happy for tuple structs & variants to be handled in that fashion in a first draft. We can see how this feels in practice.

> `#[err_rename = "List[str]"]`

Can I propose that we instead name this attribute `#[pyo3(annotation = "List[str]")]`? In the future I'd hope we can generate type annotations from these impls so let's make a forward-compatible name.

> ```
> //generated
> impl<'a> ::pyo3::FromPyObject<'a> for A<'a> {
>     fn extract(ob: &'a ::pyo3::PyAny) -> ::pyo3::PyResult<Self> {
>         Ok(Self {
>             s: ::pyo3::FromPyObject::extract(ob)?,
>             t: ::pyo3::FromPyObject::extract(ob)?,
>             p: ::pyo3::FromPyObject::extract(ob)?,
>         })
>     }
> }
> ```

I'm not sure I agree with this implementation - it looks to me that `A<'a>` can basically only be extracted from a Python `str`, and that `s`, `t`, `p` will all end up with the same data?

I think we need to consider what `A` -> (python) -> `A` conversion should look like. I think that the conversion from `A` to Python will probably be a dict with keys `"s"`, `"t"`, and `"p"`, so I think the from-python conversion should probably be similar?

If I were to write it as a Python type annotation, I think the Python "type" that `A<'a>` above should map to / from would be:

```
class A(TypedDict):
    s: str
    t: str
    p: Any
```

sebpuetz:
> > Should the FromPyObject impl handle structs / variants with multiple fields? E.g., struct AllTheStrings<'a>(String, &'a PyString, &'a PyAny) could all be extracted from the same object. My current (local) implementation handles all that but I'm unsure whether we want this. What do you think?
> 
> It seems reasonable to me that the Python "equivalent" of `AllTheThings<'a>` is `Tuple[str, str, Any]`, so I'm happy for tuple structs & variants to be handled in that fashion in a first draft. We can see how this feels in practice.

The Python equivalent would actually be `str` or something for that on the Rust side extractions to `String` ,`&'a PyString` and `&'a PyAny` has been implemented. `FromPyObject` just gets a single argument, if that argument is a `Tuple[str, str, Any]`, then the corresponding type would be `(String, String, &'a PyAny)` (although I don't think extraction is implemented for tuples).

~~~Rust
#[derive(FromPyObject)]
struct Tuple<'a>((String, String, &'a PyAny));
~~~

> > `#[err_rename = "List[str]"]`
> 
> Can I propose that we instead name this attribute `#[pyo3(annotation = "List[str]")]`? In the future I'd hope we can generate type annotations from these impls so let's make a forward-compatible name.

Sure that was just whatever I came up with at the time.

> I'm not sure I agree with this implementation - it looks to me that `A<'a>` can basically only be extracted from a Python `str`, and that `s`, `t`, `p` will all end up with the same data?
> 
> I think we need to consider what `A` -> (python) -> `A` conversion should look like. I think that the conversion from `A` to Python will probably be a dict with keys `"s"`, `"t"`, and `"p"`, so I think the from-python conversion should probably be similar?
> 
> If I were to write it as a Python type annotation, I think the Python "type" that `A<'a>` above should map to / from would be:
> 
> ```
> class A(TypedDict):
>     s: str
>     t: str
>     p: Any
> ```

This is more or less the same as above: If a dict should be extracted, that needs to be guided from the type of the field that it gets extracted to. I'm not sure how this would work otherwise...

davidhewitt:
For example I think the implementation for `A<'a>` could be:

```
impl<'a> ::pyo3::FromPyObject<'a> for A<'a> {
    fn extract(ob: &'a ::pyo3::PyAny) -> ::pyo3::PyResult<Self> {
        Ok(Self {
            s: ob.get_item("s")?.extract()?,
            t: ob.get_item("t")?.extract()?,
            p: ob.get_item("p")?.extract()?,
        })
    }
}
```

and for `AllTheThings<'a>`:

```
impl<'a> ::pyo3::FromPyObject<'a> for AllTheThings<'a> {
    fn extract(ob: &'a ::pyo3::PyAny) -> ::pyo3::PyResult<Self> {
        let tup: (String, &'a PyString, &'a PyAny) = ob.extract()?;
        Ok(Self(tup.0, tup.1, tup.2))
    }
}
```

gilescope:
Any news? Can't wait to try out pyo3 with ADTs. Is there anything we can do to help?

sebpuetz:
I got caught up in some other stuff, I might find some time later this week / towards the weekend to get back to this.

I'm not completely sold on the assumed dict-struct relation in the derivation. IMO, a more intuitive parallel for struct fields would be properties / attributes in Python classes, so what's your opinion on `getattr` instead of `getitem`?

The PyTuple-to-tuple-struct extraction makes sense to me as proposed by @davidhewitt

davidhewitt:
> a more intuitive parallel for struct fields would be properties / attributes in Python classes

This is a fair view and I don't disagree that `getattr` seems more pure than `get_item`, but at the same time what type would that Python class be?

I think that whatever design we build here should have an equivalent to-python implementation we can also derive in pyo3. (Even if we don't implement the to-python conversion as part of this PR.) 

I can't think of an obvious class type for the to-python conversion, so I think struct -> dict seems best for that direction. And equivalently this makes me think the from-python conversion should be dict->struct.

Another viewpoint I've been thinking on: how to implement a `serde` serializer & deserializer that maps Rust <-> Python. I seems to be me that for a serde implementation, the following must be true:

1.`Python -> Rust -> Python` roundrips always produce the original result.
2. `Rust -> JSON` should be equivalent to `Rust -> Python -> JSON`, i.e. the output of such a serializer should be Python objects that would json-serialize the same way as `serde_json` would serialize the original Rust type.

I _think_ that both of these two rules should probably apply to the design here?

sebpuetz:
> > a more intuitive parallel for struct fields would be properties / attributes in Python classes
> 
> This is a fair view and I don't disagree that `getattr` seems more pure than `get_item`, but at the same time what type would that Python class be?
> 
> I think that whatever design we build here should have an equivalent to-python implementation we can also derive in pyo3. (Even if we don't implement the to-python conversion as part of this PR.)
> 
> I can't think of an obvious class type for the to-python conversion, so I think struct -> dict seems best for that direction. And equivalently this makes me think the from-python conversion should be dict->struct.
> 
> Another viewpoint I've been thinking on: how to implement a `serde` serializer & deserializer that maps Rust <-> Python. It seems to me that for a serde implementation, the following should be true:
> 
> 1.`Python -> Rust -> Python` roundrips always produce the original result.
> 2. `Rust -> JSON` is equivalent to `Rust -> Python -> JSON`, i.e. the output of such a serializer should be Python objects that would json-serialize the same way as `serde_json` would serialize the original Rust type.
> 
> I _think_ that both of these two rules should probably apply to the design here?

I don't think serde's `derive(Serialize, Deserialize)` guarantees that all data survives roundtrips. And I don't think that deriving `FromPyObject` / `ToPyObject` should make such guarantees either. Assuming that we are always extracting from a dict will quite possibly cause some odd code on the Python side where you'd naturally have some `class Foo` with field `_bar` exposed through a property `bar`. Being forced to store this in a dictionary as `{"bar": value}` in Python seems like a strange constraint to me.

I wouldn't expect `derive(ToPyObject)` to implement conversion to Python dictionaries, I'd expect that from `derive(ToPyDict)`. A much more involved solution could generate a `pyclass` and expose the fields as properties / attributes. Or maybe we can find a middleground and use `namedtuples` in this case? Although I have no clue whether we can easily generate that in PyO3.

davidhewitt:
>I don't think serde's derive(Serialize, Deserialize) guarantees that all data survives roundtrips.

True, but I think in practice if data does not roundtrip, it's extremely suprising to unexpecting users. If it's possible for us to make roundtripping work, it should. See e.g. https://github.com/RReverser/serde-xml-rs/issues/130

> A much more involved solution could generate a pyclass and expose the fields as properties / attributes.

This is true, but I'm not convinced that adding a `#[derive]` macro which expands to `#[pyclass]` under the hood is a good design. That kinda defeats the rule from [this comment](https://github.com/PyO3/pyo3/pull/1065#issuecomment-667482693).

> Or maybe we can find a middleground and use namedtuples in this case?

I'd be much more happy with this solution. There's still questions of how to construct the `namedtuple` type, though perhaps we can find a way to overcome that.

----

A possible interesting case of discussion: I've now started a WIP implementation on such a serde/python implementation. See https://github.com/davidhewitt/pythonize

I've not quite settled on the final design, so I'd welcome any feedback on that.

I kinda see these `#[derive(IntoPyObject, FromPyObject)]` macros as being roughly equivalent to the serde implementation but without requiring all the overhead of the serde machinery. And so I would hope they behave the same as that crate. But I could be convinced that shouldn't be the case.

davidhewitt:
(See also #566 and #884 which are kind of related to this problem area.)

sebpuetz:
I tried implementing some of the suggestions but I think there's some issues to smooth out:

~~~Rust
#[derive(FromPyObject)]
struct Wrapper(String);
~~~

With such a struct it doesn't seem intuitive to extract the `String` from a `PySequence` / `PyTuple`. So we'd probably want some option that switches to simply `String::from_pyobject(obj)?`, i.e. something like:

~~~Rust
#[derive(FromPyObject)]
#[pyo3(extract_wrapper)]
struct Wrapper(String);
~~~

The same applies to a wrapper struct with a named field:
~~~Rust
#[derive(FromPyObject)]
struct Wrapper { wrapped: String }
~~~

should offer the same switch.

This probably extends to enums, too.

------------

Regarding the extraction of named fields, I think it'd be nice to decide this based on a field attribute, i.e.:

~~~Rust
#[derive(FromPyObject)]
struct WithNamedFields {
    #[pyo3(getattr)]
    from_attr: String,
    #[pyo3(get_item)]
    from_item: String,
}

// expands to

impl<'a> FromPyObject<'a> for WithNamedFields {
        Ok(WithNamedFields {
            from_attr: obj.getattr("from_attr")?.extract()?,
            from_item: obj.getitem("from_item")?.extract()?
        }
    }
}
~~~

It'd probably be nice to untie the field name from the Rust implementation. Otherwise trying to derive `FromPyObject` for some type from a foreign Python package would force the Rust struct to use the name the Python package chose.

davidhewitt:
Thanks for proceeding forwards with this implementation! I think we can can draw inspiration to make configurable attributes from [serde's set of attributes](https://serde.rs/field-attrs.html).

> With such a struct it doesn't seem intuitive to extract the String from a PySequence / PyTuple. So we'd probably want some option that switches to simply String::from_pyobject(obj)?

Let's make `#[pyo3(transparent)]`, the same as serde has [`#[serde(transparent)]`](https://serde.rs/container-attrs.html#transparent)?

> It'd probably be nice to untie the field name from the Rust implementation. Otherwise trying to derive FromPyObject for some type from a foreign Python package would force the Rust struct to use the name the Python package chose.

Serde does this with `#[serde(rename = "...")]`. In other places in pyo3 we've used `#[name = "..."]`, though I think that was a mistake and it should have been _at least_ `#[pyo3(name = "...")]`.

So perhaps `#[pyo3(name = "...")]` here? I could also accept `rename` if you prefer it, though think `name` is just as clear when documented. (And then I might make a PR to change the attribute in other contexts.)

> Regarding the extraction of named fields, I think it'd be nice to decide this based on a field attribute

Agreed, though I think probably should go on the struct rather than the fields - probably you want either all `getattr`, or all `get_item`? 

As it's not clear right now whether attributes or items is a better default: how about making it required for one of `#[pyo3(from_attributes)]` or `#[pyo3(from_items)]` to be specified for now, if we're in one of these cases where the strategy needs to be specified?

If you have ideas for better names for this pair of attributes I'm happy for you to use something else.

Later we could make it optional when we figure out the better default strategy.

davidhewitt:
> How about making it only for enum so far?

I think unfortunately because struct variants and tuple variants encode all the possibilites, solving for enum pretty much requires solving for structs too.

If the scope was to reduce on this PR I think we'd have to dial it back pretty much to just the original design (i.e. single-item tuple variants).

I'm happy with whatever @sebpuetz is planning to build! 😄

sebpuetz:
I have an ugly WIP solution that handles most of the desired cases for both enums and structs (I haven't handled the wrapper-struct-variant yet).

It's pretty hard to abstract over tuple structs, tuple variants, proper structs and struct variants, at least I'm getting lost with all the different `syn` types. Before this is reviewable, I'll definitely have to factor out some common logic.

I'm pushing the current state to this branch so this doesn't look all that stale. Perhaps you can voice your opinion on some of the decisions, too. 

~~~Rust
#[pyfunction]
pub fn foo(mut inp: A) {
    println!("{:?}", inp);
}
#[pyfunction]
pub fn bar(mut inp: Foo) {
    println!("{:?}", inp);
}

#[pyfunction]
pub fn baz(mut inp: B) {
    println!("{:?}", inp);
}

pub struct A<'a> {
    #[extract(getattr)]
    s: String,
    #[extract(get_item)]
    t: &'a PyString,
    #[extract(getattr)]
    #[name = "foo"]
    p: &'a PyAny,
}

#[wrapper]
pub struct B {
    test: String
}

pub struct C {
    #[extract(getattr)]
    #[name = "bla"]
    test: String
}

pub enum Foo<'a> {
    #[err_rename = "bla"]
    FF(&'a PyAny, String),
    #[wrapper]
    F(&'a PyAny)
}
~~~

~~~Python
>>>import from_py_object
>>> from_py_object.bar(("test", "test"))                                                                                                                                                                       
F('test')
>>> from_py_object.bar(("test", "test"))                                                                                                                                                                       
FF('test', "test")
>>> class Foo:
...     def __init__(self):
...         self.s = "test"
...         self.foo = None
...     def __getitem__(self, key):
...         return "Foo"
>>> from_py_object.foo(Foo())
A { s: "test", t: 'Foo', p: None }
>>> from_py_object.baz("test")                                                                                                                                                                                
B { test: "test" }
~~~

davidhewitt:
Awesome! 👍

I'll try to find time to review whatever state this PR is in at some point in the next few days. Perhaps that can help with some of the refactoring.

sebpuetz:
I pushed a cleaned up version, some stuff that's left to do (apart from what reviewers will mention):

* [ ] Naming (and scoping?) of attributes
* [ ] Write docs (depends on names)
* [ ] Compilation-fail tests?

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    /// Construct a container based on fields, identifier and attributes.
+    ///
+    /// Fails if the variant has no fields or incompatible attributes.
+    fn new(fields: &'a Fields, ident: &'a Ident, attrs: &'a [Attribute]) -> Result<Self> {
+        let transparent = attrs.iter().any(|a| a.path.is_ident("transparent"));
+        if transparent {
+            Self::check_transparent_len(fields)?;
+        }
+        let style = match fields {
```
davidhewitt, line 200: It might be easier to match `(fields, transparent)` together.

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+        let mut fields: Punctuated<Expr, syn::Token![,]> = Punctuated::new();
+        let mut ext: Punctuated<Expr, syn::Token![,]> = Punctuated::new();
+        for (ident, ext_fn) in tups {
+            field_pats.elems.push(parse_quote!(Ok(#ident)));
+            fields.push(parse_quote!(#ident));
+            ext.push(parse_quote!(obj.#ext_fn.and_then(|o| o.extract())));
+        }
+        quote!(if let #field_pats = #ext {
+            return Ok(#enum_ident::#var_ident{#fields});
+        })
```
davidhewitt, line 147: I think if you generate this as a closure, you might be able to re-use the non-variant case. This would cut down on the amount of macro, and also might lead to performance benefits, as with this current form it'll try to extract all the fields. Whereas the struct form below gives up as soon as any one fails.

Something vaguely like the below. 

```
let build_struct_variant = build_struct(...);
quote!( || -> pyo3::PyResult<_> { #build_struct_variant })
```

If you change all of the variant cases to be like this, the overall enum logic could probably become something like:

```
#(if let Ok(variant) = (#var_extracts)() { return variant; })*
```

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+                        fields.push((ident, ext_fn(&field.attrs, ident)?))
+                    }
+                    Style::Struct(fields)
+                }
+            }
+            Fields::Unit => {
+                return Err(syn::Error::new_spanned(
+                    &fields,
+                    "Cannot derive FromPyObject for Unit structs and variants",
+                ))
```
davidhewitt, line 236: Again would be nice to see what serde does here with json. It might be acceptable for `None` to be provided?

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+        }
+        Ok(())
+    }
+}
+
+/// Get the extraction function that's called on the input object.
+///
+/// Valid arguments are `get_item`, `get_attr` which are called with the
+/// stringified field identifier or a function call on `PyAny`, e.g. `get_attr("attr")`
+fn ext_fn(attrs: &[Attribute], field_ident: &Ident) -> Result<syn::ExprCall> {
```
davidhewitt, line 315: For attributes, it may be nice to have an `enum` which encodes all the attributes, along with a `parse_derive_attributes` function which returns a `Vec` of the attributes found.

Awkward cases that might need to be handled: 
- what happens if the same attribute is parsed twice?
- what happens if the attribute is provided in an invalid context? E.g. `extract` attribute in a tuple struct - would probably be desirable to raise a helpful error.

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+    }
+    Ok(None)
+}
+
+fn verify_and_get_lifetime(generics: &syn::Generics) -> Result<Option<&syn::LifetimeDef>> {
+    let lifetimes = generics.lifetimes().collect::<Vec<_>>();
+    if lifetimes.len() > 1 {
+        return Err(syn::Error::new_spanned(
+            &generics,
+            "Only a single lifetime parameter can be specified.",
```
davidhewitt, line 372:
```suggestion
            "FromPyObject can only be derived with at most one lifetime parameter.",
```

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+
+/// Derive FromPyObject for enums and structs.
+///
+///   * Max 1 lifetime specifier, will be tied to `FromPyObject`'s specifier
+///   * At least one field, in case of `#[transparent]`, exactly one field
+///   * At least one variant for enums.
+///   * Fields of input structs and enums must implement `FromPyObject`
+///   * Derivation for structs with generic fields like `struct<T> Foo(T)`
+///     adds `T: FromPyObject` on the derived implementation.
+pub fn build_derive_from_pyobject(tokens: &mut DeriveInput) -> Result<TokenStream> {
```
davidhewitt, line 386: I'm not sure I've spotted a case where tokens is modified.

```suggestion
pub fn build_derive_from_pyobject(tokens: &DeriveInput) -> Result<TokenStream> {
```

tests/test_frompyobject.rs
```diff
@@ -0,0 +1,308 @@
...
+use pyo3::prelude::*;
+use pyo3::types::{PyDict, PyString, PyTuple};
+use pyo3::{PyErrValue, PyMappingProtocol};
+
+#[macro_use]
+mod common;
+
+#[derive(Debug, FromPyObject)]
+pub struct A<'a> {
+    #[extract(getattr)]
```
davidhewitt, line 11: Considering the possibility of a to-python conversion later, I'd prefer use a term that could be symmetric. e.g. `#[pyo3(field_type = "attribute")]`, or maybe just `#[pyo3(attribute)]`

tests/test_frompyobject.rs
```diff
@@ -0,0 +1,308 @@
...
+    s: String,
+    #[extract(get_item)]
```
davidhewitt, line 13: Similarly `#[pyo3(field_type = "item")]` or `#[pyo3(item)]`

tests/test_frompyobject.rs
```diff
@@ -0,0 +1,308 @@
...
+    t: &'a PyString,
+    #[extract(getattr("foo"))]
```
davidhewitt, line 15: The nice thing about the pyo3 form is that it could compose

```suggestion
    #[pyo3(attribute, name = "foo")]
```

tests/test_frompyobject.rs
```diff
@@ -0,0 +1,308 @@
...
+            let d = <&PyDict>::extract(any).expect("Expected pydict");
+            assert!(d.is_empty());
+        }
+        _ => panic!("Expected extracting Foo::CatchAll, got {:?}", f),
+    }
+}
+
+#[derive(FromPyObject)]
+pub enum Bar {
+    #[rename_err = "str"]
```
davidhewitt, line 282: Similarly I think annotation can be used more broadly later (if we started being able to generate type hints from these, for example).

```suggestion
    #[pyo3(annotation = "str")]
```

davidhewitt:
(Also, much kudos for writing a comprehensive set of tests 👍)

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+    /// Construct a new enum representation.
+    ///
+    /// `data_enum` is the `syn` representation of the input enum, `ident` is the
+    /// `Identifier` of the enum.
+    fn new(data_enum: &'a DataEnum, ident: &'a Ident) -> Result<Self> {
+        if data_enum.variants.is_empty() {
+            return Err(syn::Error::new_spanned(
+                &data_enum.variants,
+                "Cannot derive FromPyObject for empty enum.",
+            ));
```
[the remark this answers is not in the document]
sebpuetz: Iirc, it depends on the attributes for serde, without adding `tag(type)` or whatever the attribute is, it will serialize empty structs & variants the same as `None`, e.g. as `null` in `serde_json`. I think deserialization works almost the same, it's possible to deserialize `SomeEnum::Foo` from the serialization of `SomeEnum::Bar` if the enum is untagged.

In any way, I think having a `#[transparent]` variant around `&'a PyAny` is a better solution for a catch-all extraction than discarding the input and generating an empty variant. At least, using such an empty variant as catch-all is the only reasonable use-case that I could come up with, unless we want to tie such variants to something like type names.

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+            match <::pyo3::types::PyTuple as ::pyo3::conversion::PyTryFrom>::try_from(obj) {
+                Ok(s) => {
+                    if s.len() == #len {
+                        let slice = s.as_slice();
+                        if let (#field_pats) = (#ext) {
+                            return Ok(#enum_ident::#var_ident(#fields))
+                        }
+                    }
+                },
+                Err(_) => {}
```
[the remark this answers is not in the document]
sebpuetz: I think the failed extractions are communicated through the error message, even with some room for customizations by renaming the failed extraction types. Keeping all the error messages from e.g. simple fields or tuple extractions seems too complex.

pyo3-derive-backend/src/frompy.rs
```diff
@@ -0,0 +1,428 @@
...
+            return Err(syn::Error::new_spanned(
+                tokens,
+                "FromPyObject can only be derived for structs and enums.",
+            ))
+        }
+    };
+
+    let ident = &tokens.ident;
+    Ok(quote!(
+        #[automatically_derived]
```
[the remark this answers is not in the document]
sebpuetz: Saw this on the output of `cargo expand` on `derive(Debug)` generated code, this [SO post](https://stackoverflow.com/questions/51481551/what-does-automatically-derived-mean) mentions that it doesn't produce unused warnings among other stuff.

kngwyu:
Thank you, so in my understanding:

1. `#[transparent]`.
Only struct-like (i.e. `name: value` style) variants can have `transparent`? Maybe it's better to have a test for it.
2. `#[extract]`
What does happen if `extract` is not specified? `getattr` is used?
In addition, does it accepts only `getattr` and `getitem`? Then why does it need both key (`extract`) and value (`getattr`)?

> Naming (and scoping?) of attributes

Could you please explain more precisely?

sebpuetz:
> Thank you, so in my understanding:
> 
>     1. `#[transparent]`.
>        Only struct-like (i.e. `name: value` style) variants can have `transparent`? Maybe it's better to have a test for it.

Looks like I forgot to properly implement the check for the `#[transparent]` tuple struct. I'll amend this.

`#[transparent]` is applicable to all single-field struct and variants, i.e. 
~~~Rust
#[transparent]
struct TransparentStruct { field: String }
struct TransparentTuple(String);

enum SomeEnum {
    #[transparent]
    TransparentStructVariant { field: String },
    TransparentTupleVariant(String),
}
~~~
are all valid, since it's possible to extract all these variants' fields through a simple `obj.extract()` call. Multi-field structs and variants would be restricted to extract all fields from the same input `obj`, this was one of the earlier solutions in this PR, but I don't think it makes much sense to have this.

>     2. `#[extract]`
>        What does happen if `extract` is not specified? `getattr` is used?
>        In addition, does it accepts only `getattr` and `getitem`? Then why does it need both key (`extract`) and value (`getattr`)?

Yes, it defaults to `getattr`. The `extract` key is arbitrary. Is the nesting an issue for you? We could also extend the legal inputs to any method defined on `& PyAny`, but I feel like book-keeping might become too hard. Third option would be to give the user free reign over the inputs. 

> > Naming (and scoping?) of attributes
> 
> Could you please explain more precisely?

How should each of the attributes be called, i.e. is `extract`, `transparent`, `err_rename` clear enough?

"Scoping" was referring to whether we should check that the attributes are applied to valid contexts, so that there are no attributes without any effect.

sebpuetz:
Thanks for the thorough review @davidhewitt, especially the suggestion to use a closure was awesome!

I'm still not sure about the attributes, or rather, how to handle duplicates / invalid combinations / invalid applications. E.g. `#[derive(FromPyObject)]` on a `#[pyclass]` and adding `pyo3(attribute)` on one of the fields collides with the other `pyo3` attributes defined in `pyclass`. Probably not an issue since it'd be unusual to derive `FromPyObject` on a `pyclass`, but I think a principled approach would be nice.

kngwyu:
> I'm still not sure about the attributes, or rather, how to handle duplicates / invalid combinations / invalid applications

Since `#[pyclass]` must be placed before `derive` (i.e., `#[pyclass] #[derive(...)] struct`), we cannot inihibit that from `#[derive(FromPyObject)]` code and it should be done in `#[pyclass]`.
So I think it's better to do so in another PR.

And, looking at the new `#[pyo3(attribute)]` and `item`, I feel a bit confused. One can easily remember of `getattr` and `getitem` from them?

sebpuetz:
> And, looking at the new `#[pyo3(attribute)]` and `item`, I feel a bit confused. One can easily come up with `getattr` and `getitem` from them?

That's based on suggestions by @davidhewitt, using `pyo3(attribute)` and `pyo(item)` makes these attributes usable for `#[derive(ToPyObject)]` implementations. If we used `pyo3(getattr)`, it'd be awkward as that's not a fitting name for setting an attribute in `ToPyObject`.

kngwyu:
> That's based on suggestions by @davidhewitt, using pyo3(attribute) and pyo(item) makes these attributes usable for #[derive(ToPyObject)] implementations. If we used pyo3(getattr), it'd be awkward as that's not a fitting name for setting an attribute in ToPyObject.

Ah, thanks, that makes sense.
Then my points are:
- I'm not sure we should have a *serializer* like this. But I have not so strong opinions about that.
- `item` sounds still confusing to me and I'm continuing to look for another one.

sebpuetz:
The `item` variant is motivated by being the suffix for both `set_item` and `get_item`. So that's the common terminology for values of Python dicts / sequences.

But we could stick to the verbose variants `getattr` and `get_item` and have a potential `ToPyObject` derivation use `setattr` and `set_item` attributes. That would actually give room for more customization, but it would also allow breaking roundtrips. I'm not really sold on having a serializer either, but I think that's something @davidhewitt would like to have.

davidhewitt:
I have a feeling that as soon as we have from-python enum support, users will soon be asking for something like this:

```
#[derive(IntoPy)]
enum MyEnum {
    String(String),
    Int(i64),
}

#[pyfunction]
fn string_or_int(text: bool) -> MyEnum {
    if text {
        MyEnum::String("1".into()),
    } else {
        MyEnum::Int(1)
    }
}
```

sebpuetz:
I feel like it'd be better to point out `into_py()` in that case, rather than adding more proc-macro-magic... At least I think the into impl would just boil down to: unwrap the enum variant and call `into_py()` on it

davidhewitt:
That's true. Either way I'm happy to wait for a serializer until it actually seems helpful. I was just suggesting that the names we pick here try to be forward compatible with the possibility of adding one.

If you folks prefer names that are specific to the from-python case, because they are potentially clearer for the user, I'm very happy to be outvoted :smile:

sebpuetz:
No worries, I'm fine with either version. I'd leave it up to @kngwyu to decide!

kngwyu:
What makes me confused is `#[pyo3(item)]` modifies the name of struct member which would be a key for the container(, or a key for `getitem`).
This is inconsistent with `#[pyo3(attribute)]` that modifies what would be an attribute.
On this line, I came up with some names such `#[pyo3(itemkey)]` or `#[pyo3(key)]`, but am still not happy with these names ... :thinking:

sebpuetz:
I see where you're coming from, I was thinking of

~~~Python
getattr(obj, attr)
~~~

while you're thinking about 

~~~Python
obj.attr
~~~

Maybe we can make the valid arguments the following:

`pyo3(attribute)`, `pyo3(item)`, `pyo3(key = "key")`, where `key` is combinable with both `attribute` and `item`?

kngwyu:
Thanks, then I think the pair of `#[pyo3(attribute, name = "foo")]` and `#[pyo3(item, key = "foo")]` would be better, though the implementation would be a bit complicated.

davidhewitt:
Sounds reasonable - it's a bit of a shame that `name` and `key` will be two separate ideas, but it's definitely the most clear. (Also hopefully `name` / `key` will not be needed in most cases!)

sebpuetz:
I went with `pyo3(item("key"))` and `pyo3(attribute("name"))` for now. That's motivated by being simpler to implement while - in my opinion - not leaving room for any ambiguous / unexpected naming. I think it should be clear in the context what the literal arguments for the `attribute` and `item` attributes mean.

davidhewitt:
Thanks, that seems reasonable to me!

sebpuetz:
CI failure seems due to 1.46?

davidhewitt:
Yep once #1121 is merged you should be able to rebase on that.

davidhewitt:
(I just merged #1121.)

sebpuetz:
All green, do we want UI tests & docs done within this PR?

davidhewitt:
:tada:

UI tests - I think it's desirable to get as much test coverage as possible in the original PR, so if you have ideas what you would make for UI tests it would be great to have here.

Docs - I'm okay with merging PRs without docs, but especially for a new feature like this nobody will use it until we write good documentation. So if we choose to wait on docs we should open an issue when closing this PR so that we remind ourselves docs are needed.

sebpuetz:
I'm working on the UI tests, docs are up next. Should be done in the afternoon!

sebpuetz:
If everything turns out green, this should be reviewable!

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -0,0 +1,545 @@
...
+                    }
+                }
+                FieldAttribute::GetItem(key) => {
+                    if let Some(key) = key.as_ref() {
+                        quote!(get_item(#key))
+                    } else {
+                        quote!(get_item(stringify!(#ident)))
+                    }
+                }
+            };
```
kngwyu, line 281:
```suggestion
            let ext_fn = match attr {
                FieldAttribute::GetAttr(Some(name)) => quote!(getattr(#name)),
                FieldAttribute::GetAttr(None) => quote!(getattr(stringify!(#ident))),
                FieldAttribute::GetItem(Some(key)) => quote!(getitem(#key)),
                FieldAttribute::GetItem(None) => quote!(getitem(stringify!(#ident))),
            };
```

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -0,0 +1,545 @@
...
+            return Ok(None);
+        }
+        if list.nested.len() > 1 {
+            return Err(spanned_err(
+                list.nested,
+                "Only one of `item`, `attribute` can be provided, possibly with an \
+                additional argument: `item(\"key\")` or `attribute(\"name\").",
+            ));
+        }
+        let metaitem = list.nested.into_iter().next().unwrap();
```
kngwyu, line 385:
```suggestion
        let metaitem = match list.nested.len() {
            0 => return Ok(None),
            1 => list.nested.into_iter().next().unwrap(),
            _ => {
                return Err(spanned_err(
                    list.nested,
                    "Only one of `item`, `attribute` can be provided, possibly with an \
                     additional argument: `item(\"key\")` or `attribute(\"name\").",
                ))
            }
        };
```

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -0,0 +1,545 @@
...
+        } else {
+            Err(spanned_err(meta, "Expected `attribute` or `item`."))
+        }
+    }
+
+    fn attribute_arg(meta: syn::Meta) -> syn::Result<Option<syn::LitStr>> {
+        let arg_list = match meta {
+            syn::Meta::List(list) => list,
+            syn::Meta::Path(_) => return Ok(None),
+            Meta::NameValue(nv) => {
```
kngwyu, line 409: Maybe it's better to use only one of `syn::Meta` or `Meta`?

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -0,0 +1,545 @@
...
+                return Err(spanned_err(nv, err_msg));
+            }
+        };
+        let arg_msg = "Expected a single string literal argument.";
+        if arg_list.nested.is_empty() {
+            return Err(spanned_err(arg_list, arg_msg));
+        } else if arg_list.nested.len() > 1 {
+            return Err(spanned_err(arg_list.nested, arg_msg));
+        }
+        let first = arg_list.nested.first().unwrap();
```
kngwyu, line 420: Don't `arg_list` and `arg_list.nested` have almost the same span?
```suggestion
        let first = match arg_list.nested.len() {
            1 => arg_list.nested.first().unwrap(),
            _ => return Err(spanned_err(arg_list, arg_msg)),
        };
```

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -0,0 +1,545 @@
...
+        if let syn::NestedMeta::Lit(lit) = first {
+            if let syn::Lit::Str(litstr) = lit {
+                if litstr.value().is_empty() {
+                    return Err(spanned_err(litstr, "Attribute name cannot be empty."));
+                }
+                return Ok(Some(parse_quote!(#litstr)));
+            }
+        }
```
kngwyu, line 428:
```suggestion
        if let syn::NestedMeta::Lit(syn::Lit::Str(litstr)) = first {
            if litstr.value().is_empty() {
                return Err(spanned_err(litstr, "Attribute name cannot be empty."));
            }
            return Ok(Some(parse_quote!(#litstr)));
        }
```

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -0,0 +1,545 @@
...
+        Err(spanned_err(first, arg_msg))
+    }
+
+    fn item_arg(meta: syn::Meta) -> syn::Result<Option<syn::Lit>> {
+        let arg_list = match meta {
+            syn::Meta::List(list) => list,
+            syn::Meta::Path(_) => return Ok(None),
+            Meta::NameValue(nv) => {
```
kngwyu, line 436: Same as the above.

kngwyu (approved):
LGTM, thank you for the hard work!
My suggestions are mostly for reducing LOC and not critical. Please address them if you like.

sebpuetz:
Thanks for the suggestions, I pushed the sugested changes. Hoping for all-green CI!

guide/src/conversions.md
```diff
@@ -119,6 +119,147 @@ mutable references, you have to extract the PyO3 reference wrappers [`PyRef`]
...
+    Coordinates3d { // needs to be in front of 2d
+        x: usize,
+        y: usize,
+        z: usize,
+    },
+    Coordinates2d { // only gets checked if the input did not have `z`
+        #[pyo3(attribute("x"))]
+        a: usize,
+        #[pyo3(attribute("y"))]
+        b: usize,
```
davidhewitt, line 235: As `attribute` is redundant here, I'd just do

```suggestion
        a: usize,
        b: usize,
```

to be consistent with the 3D case above

guide/src/conversions.md
```diff
@@ -119,6 +119,147 @@ mutable references, you have to extract the PyO3 reference wrappers [`PyRef`]
...
+enum RustyEnum {
+    #[pyo3(transparent, annotation = "str")]
+    String(String),
+    #[pyo3(transparent, annotation = "int")]
+    Int(isize),
+}
+```
+
+If the input is neither a string nor an integer, the error message will be:
+`"Can't convert <INPUT> to str, int"`, where `<INPUT>` is replaced by the type name and
```
davidhewitt, line 260: I think it'd be really nice to use the `Union` type here:

```suggestion
`"Can't convert <INPUT> to Union[str, int]"`, where `<INPUT>` is replaced by the type name and
```

guide/src/conversions.md
```diff
@@ -119,6 +119,147 @@ mutable references, you have to extract the PyO3 reference wrappers [`PyRef`]
...
+
+```
+use pyo3::prelude::*;
+
+#[derive(FromPyObject)]
+enum RustyEnum<'a> {
+    #[pyo3(transparent)]
+    Int(usize), // input is a positive int
+    #[pyo3(transparent)]
+    String(String), // input is a string
```
davidhewitt, line 223: I wonder if `#[pyo3(transparent)]` should just be assumed for single-element tuple variants like this? It slightly breaks symmetry, but I would think most people would commonly want that.

The 1-tuple case could still be used explicitly with e.g. `StringOneTuple((String,))`.

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -0,0 +1,531 @@
...
+                        ))
+                    }
+                }
+            } else {
+                return Err(spanned_err(
+                    meta,
+                    "Unknown container attribute, expected `transparent` or \
+                    `annotation(\"err_name\")`",
+                ));
+            }
```
davidhewitt, line 346: How about dropping the else outside and doing the `Err` return just once? 

Also, I'd be tempted to not list the possible options here as it can get easily outdated.

```suggestion
                    _ => {} // return Err below
                }
            }
            
            return Err(spanned_err(
                meta,
                "Unrecongised `pyo3` container attribute",
            ));
```

guide/src/conversions.md
```diff
@@ -119,6 +119,147 @@ mutable references, you have to extract the PyO3 reference wrappers [`PyRef`]
...
+    String(String),
+    #[pyo3(transparent, annotation = "int")]
+    Int(isize),
+}
+```
+
+If the input is neither a string nor an integer, the error message will be:
+`"Can't convert <INPUT> to str, int"`, where `<INPUT>` is replaced by the type name and
+`repr()` of the input object. 
+
```
davidhewitt, line 262: At the end of this section it may be very nice just to have bulleted list of all the valid container and field attributes, the same way [serde does it](https://serde.rs/container-attrs.html):

```suggestion
#### `#[derive(FromPyObject)]` Container Attributes
- list goes here

#### `#[derive(FromPyObject)]` Field Attributes
- list goes here
```

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -0,0 +1,531 @@
...
+            let attrs = ContainerAttribute::parse_attrs(&tokens.attrs)?;
+            Container::verify_struct_container_attrs(&attrs, &tokens.attrs)?;
+            let ident = &tokens.ident;
+            let st = Container::new(&st.fields, parse_quote!(#ident), attrs, false)?;
+            st.build()
+        }
+        syn::Data::Union(_) => {
+            return Err(spanned_err(
+                tokens,
+                "FromPyObject can not be derived for unions.",
```
davidhewitt, line 517:
```suggestion
                "#[derive(FromPyObject)] is not supported for unions.",
```

davidhewitt (approved):
This is looking absolutely excellent; great tests and docs! I just have a few final suggestions/questions and then let's merge it!

guide/src/conversions.md
sebpuetz: It's not actually reduntant, the arguments are the relevant parts! So, it's possible to extract from `coordinate.x` and `coordinate.y`, rather than `coordinate.a` and `coordinate.b`. That's also relevant to show the importance of order, `Coordinates2D` is only attempted to extract if `Coordinates3d` failed!

sebpuetz:
Once again, thanks for the suggestions. CI is slowly turning green, too!

guide/src/conversions.md
davidhewitt: Good point; I misread that the names were different!

davidhewitt (approved):
Looks absolutely excellent to me. Thanks so much! 🚀

sebpuetz:
Is there still something left to be addressed or is this good to go?

davidhewitt:
I'm happy, just leaving some time for @kngwyu to have a final chance to review and make the merge.

kngwyu:
Thanks, I'm happy to see this became a reality in such a nice way.

PR #1067, opened 2020-07-24
Make `PyErr: Send + Sync`

davidhewitt:
Changes `PyErr` to implement `Send + Sync`. This allows easier integration with libraries such as [anyhow](https://docs.rs/anyhow/1.0.31/anyhow/trait.Context.html#foreign-impls), which often require `Send + Sync` errors.

While technically breaking, I think it is unlikely to cause much frustration, because most types convertible to Python are `Send + Sync` already since we added the `#[pyclass]` requirements.

Also, `PyErr::from_value` can be used where the new bounds are restrictive.


davidhewitt:
@kngwyu unless you have any objections, I plan to merge this later today

kngwyu:
Sorry I overlooked this PR, thanks!

PR #1072, opened 2020-07-28
#1064: Comparisons with __eq__ should not raise TypeError

### LINKED ISSUE #1064: Comparisons with __eq__ should not raise TypeError
Consider this implementation of `__richcmp__`:

```
#[pyproto]
impl<'p> PyObjectProtocol<'p> for MyClass {
    fn __richcmp__(
        &'p self,
        other: PyRef<'p, MyClass>,
        op: CompareOp,
    ) -> PyResult<PyObject> {
        match op {
            CompareOp::Eq => Ok(self.eq(&*other).into_py(other.py())),
            _ => Ok(other.py().NotImplemented()),
        }
    }
}
```

With this comparison, comparing `MyClass() == 1` (for example) fails with `TypeError: Can't convert 1 to MyClass`.

This is fairly un-Pythonic; the usual situation for Python 3 is that `==` should return `False` if the types can't be compared.

mvaled:
RichComparisonToSelf expects to compare itself with objects of the same type, but then we shouldn't raise TypeError but return NotImplemented.

Fixes #1064

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -758,7 +758,7 @@ Each method corresponds to Python's `self.attr`, `self.attr = value` and `del se
 
 ### Emulating numeric types
 
-The [`PyNumberProtocol`] trait allows [emulate numeric types](https://docs.python.org/3/reference/datamodel.html?highlight=__ipow__#emulating-numeric-types).
+The [`PyNumberProtocol`] trait allows [emulate numeric types](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).
 
   * `fn __add__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
   * `fn __sub__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
@@ -775,9 +775,8 @@ The [`PyNumberProtocol`] trait allows [emulate numeric types](https://docs.pytho
   * `fn __or__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
   * `fn __xor__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
 
-These methods are called to implement the binary arithmetic operations (`+`,
-`-`, `*`, `@`, `/`, `//`, `%`, `divmod()`, `pow()` and `**`, `<<`, `>>`, `&`,
-`^`, and `|`).
+These methods are called to implement the binary arithmetic operations
+(`+`, `-`, `*`, `@`, `/`, `//`, `%`, `divmod()`, `pow()` and `**`, `<<`, `>>`, `&`, `^`, and `|`).
 
 If `rhs` is not of the type specified in the signature, the generated code
 will automatically `return NotImplemented`.  This is not the case for `lhs`
@@ -822,8 +821,7 @@ This trait also has support the augmented arithmetic assignments (`+=`, `-=`,
   * `fn __ior__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
   * `fn __ixor__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
 
-The following implement the unary arithmetic operations (`-`, `+`, `abs()` and
-`~`):
+The following methods implement the unary arithmetic operations (`-`, `+`, `abs()` and `~`):
 
   * `fn __neg__(&'p self) -> PyResult<impl ToPyObject>`
   * `fn __pos__(&'p self) -> PyResult<impl ToPyObject>`
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -803,6 +803,11 @@ The reflected operations are also available:
 The code generated for these methods expect that all arguments match the
 signature, or raise a TypeError.
 
+*Note*: Currently implementing the method for a binary arithmetic operations
+(e.g, `__add__`) shadows the reflected operation (e.g, `__radd__`).  This is
+being addressed in [#844](https://github.com/PyO3/pyo3/issues/844).  to make
+these methods
+
 
 This trait also has support the augmented arithmetic assignments (`+=`, `-=`,
 `*=`, `@=`, `/=`, `//=`, `%=`, `**=`, `<<=`, `>>=`, `&=`, `^=`, `|=`):
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -756,6 +756,91 @@ Each method corresponds to Python's `self.attr`, `self.attr = value` and `del se
 
     Determines the "truthyness" of the object.
 
+### Emulating numeric types
+
+The [`PyNumberProtocol`] trait allows [emulate numeric types](https://docs.python.org/3/reference/datamodel.html?highlight=__ipow__#emulating-numeric-types).
+
+  * `fn __add__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __sub__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __mul__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __matmul__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __truediv__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __floordiv__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __mod__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __divmod__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __pow__(lhs: impl FromPyObject, rhs: impl FromPyObject, modulo: Option<impl FromPyObject>) -> PyResult<impl ToPyObject>`
+  * `fn __lshift__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rshift__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __and__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __or__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __xor__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+
+These methods are called to implement the binary arithmetic operations (`+`,
+`-`, `*`, `@`, `/`, `//`, `%`, `divmod()`, `pow()` and `**`, `<<`, `>>`, `&`,
+`^`, and `|`).
+
+If `rhs` is not of the type specified in the signature, the generated code
+will automatically `return NotImplemented`.  This is not the case for `lhs`
+which must match signature or else raise a TypeError.
+
+
+The reflected operations are also available:
+
+  * `fn __radd__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rsub__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rmul__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rmatmul__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rtruediv__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rfloordiv__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rmod__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rdivmod__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rpow__(lhs: impl FromPyObject, rhs: impl FromPyObject, modulo: Option<impl FromPyObject>) -> PyResult<impl ToPyObject>`
+  * `fn __rlshift__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rrshift__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rand__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __ror__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rxor__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+
+The code generated for these methods expect that all arguments match the
+signature, or raise a TypeError.
+
+
+This trait also has support the augmented arithmetic assignments (`+=`, `-=`,
+`*=`, `@=`, `/=`, `//=`, `%=`, `**=`, `<<=`, `>>=`, `&=`, `^=`, `|=`):
+
+  * `fn __iadd__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __isub__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __imul__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __imatmul__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __itruediv__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ifloordiv__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __imod__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ipow__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ilshift__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __irshift__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __iand__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ior__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ixor__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+
+The following implement the unary arithmetic operations (`-`, `+`, `abs()` and
+`~`):
+
+  * `fn __neg__(&'p self) -> PyResult<impl ToPyObject>`
+  * `fn __pos__(&'p self) -> PyResult<impl ToPyObject>`
+  * `fn __abs__(&'p self) -> PyResult<impl ToPyObject>`
+  * `fn __invert__(&'p self) -> PyResult<impl ToPyObject>`
+
+Support for coercions:
+
+  * `fn __complex__(&'p self) -> PyResult<impl ToPyObject>`
+  * `fn __int__(&'p self) -> PyResult<impl ToPyObject>`
+  * `fn __float__(&'p self) -> PyResult<impl ToPyObject>`
+
+Other:
+
+  * `fn __index__(&'p self) -> PyResult<impl ToPyObject>`
+  * `fn __round__(&'p self, ndigits: Option<impl FromPyObject>) -> PyResult<impl ToPyObject>`
+
 ### Garbage Collector Integration
 
 If your type owns references to other Python objects, you will need to
--- a/src/class/macros.rs
+++ b/src/class/macros.rs
@@ -92,8 +92,11 @@ macro_rules! py_binary_num_func {
             $crate::callback_body!(py, {
                 let lhs = py.from_borrowed_ptr::<$crate::PyAny>(lhs);
                 let rhs = py.from_borrowed_ptr::<$crate::PyAny>(rhs);
-
-                $class::$f(lhs.extract()?, rhs.extract()?).convert(py)
+                let param = match rhs.extract::<<$class as $trait>::Right>() {
+                    Ok(param) => param,
+                    _ => return py.NotImplemented().convert(py),
+                };
+                $class::$f(lhs.extract()?, param).convert(py)
             })
         }
         Some(wrap::<$class>)
--- a/src/class/macros.rs
+++ b/src/class/macros.rs
@@ -91,12 +91,8 @@ macro_rules! py_binary_num_func {
         {
             $crate::callback_body!(py, {
                 let lhs = py.from_borrowed_ptr::<$crate::PyAny>(lhs);
-                let rhs = py.from_borrowed_ptr::<$crate::PyAny>(rhs);
-                let param = match rhs.extract::<<$class as $trait>::Right>() {
-                    Ok(param) => param,
-                    _ => return py.NotImplemented().convert(py),
-                };
-                $class::$f(lhs.extract()?, param).convert(py)
+                let rhs = extract_or_return_not_implemented!(py, rhs);
+                $class::$f(lhs.extract()?, rhs).convert(py)
             })
         }
         Some(wrap::<$class>)
@@ -225,14 +221,8 @@ macro_rules! py_ternary_num_func {
                 let arg1 = py
                     .from_borrowed_ptr::<$crate::types::PyAny>(arg1)
                     .extract()?;
-                let arg2 = match py.from_borrowed_ptr::<$crate::types::PyAny>(arg2).extract() {
-                    Ok(arg) => arg,
-                    _ => return py.NotImplemented().convert(py),
-                };
-                let arg3 = match py.from_borrowed_ptr::<$crate::types::PyAny>(arg3).extract() {
-                    Ok(arg) => arg,
-                    _ => return py.NotImplemented().convert(py),
-                };
+                let arg2 = extract_or_return_not_implemented!(py, arg2);
+                let arg3 = extract_or_return_not_implemented!(py, arg3);
                 $class::$f(arg1, arg2, arg3).convert(py)
             })
         }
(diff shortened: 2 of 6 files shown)


tests/test_arithmetics.rs
```diff
@@ -423,3 +437,57 @@ fn rich_comparisons_python_3_type_error() {
     py_expect_exception!(py, c2, "c2 >= 1", PyTypeError);
     py_expect_exception!(py, c2, "1 >= c2", PyTypeError);
 }
+
+#[test]
+fn rich_comparisons_python3_not_implemented() {
```
kngwyu, line 442: Thank you, but isn't this test too long and complicated?
I recommend separating this into three test functions.

tests/test_arithmetics.rs
```diff
@@ -423,3 +437,57 @@ fn rich_comparisons_python_3_type_error() {
...
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+
+    let c2 = PyCell::new(py, RichComparisonToSelf {}).unwrap();
+    for operation in vec!["==", "<", ">", "<=", ">="] {
```
kngwyu, line 447: We don't need vec here:
```suggestion
    for operation in ["==", "<", ">", "<=", ">="].iter() {
```

tests/test_arithmetics.rs
```diff
@@ -423,3 +437,57 @@ fn rich_comparisons_python_3_type_error() {
...
+            )
+            .as_ref()
+        );
+    }
+
+    for operation in vec!["!=", "==", "<", ">", "<=", ">="] {
+        py_expect_exception!(
+            py,
+            c2,
+            format!("class Nope:  pass; assert c2 {} Nope()", operation).as_ref(),
```
kngwyu, line 489:
```suggestion
            &format!("class Nope:  pass; assert c2 {} Nope()", operation)
```

kngwyu:
I think the implementation would be straightforward as @davidhewitt [commented](https://github.com/PyO3/pyo3/issues/1064#issuecomment-665509930), but please give us any question you have :smiley_cat:

tests/test_arithmetics.rs
mvaled: Thanks @kngwyu 

I'm still a newly born rustacean ;)

tests/test_arithmetics.rs
mvaled: I'll do it.

mvaled:
Is it ok to rewrite the commit for the changes proposed?

davidhewitt:
Yes, force-pushing to this branch is fine if you want to rewrite it as a single commit! Really up to you how you prefer to make PR :)

mvaled:
@kngwyu 

> I think the implementation would be straightforward as @davidhewitt commented, but please give us any question you have 

I'm reading the comment very carefully.  I want to understand every bit of it.  As soon as I have questions you'll know.

mvaled:
@kngwyu @davidhewitt

I've now made the first bit of progress.

1. The tests

The tests are indirect because they don't actually test the pyclass returns NotImplemented.  Instead they rely heavily on Python [reversing the operations](https://github.com/python/cpython/blob/master/Objects/object.c#L664) (BTW, I didn't know Python would try first the reversed operator of `other` when it is a subclass of `type(self)`. Cool!)

So, the setup is always the same: a Python implementation of `Other` with all the reversed dunder methods.  The idea is that when the Rust pyclass returns NotImplemented, CPython will find the reversed method on `Other` and succeed.  This would work even for in-place assignment operators, which fallback to assign the result of the reversed operation.

There's a second part of each test when `Other` doesn't implement any of the reversed operations, and they should fail with a TypeError.

For the time being, I'm focusing on richcmp.  The rest of the protocols are laid out in the tests but not being run yet.


2. Further tests?

The previous tests are simply minded.  They only test the cases when the pyclass don't have an implementation of the protocol or when extracting the arguments fails.

I'm not really sure if we need to test what happens when the pyclass does provide an implementation.  Whether this code returns NotImplemented or fails, is up to the author.  Right?


3. Implementing rich comparison first.

Following the 3rd point in [David's comment](https://github.com/PyO3/pyo3/issues/1064#issuecomment-665509930) I came up with commit 52210e0d81988611a29c1cbebf0a3fcba74489a4.

I had to comment the case test expecting a TypeError when `Other` doesn't implement the reversed method.  I think this is one of the things I need to find in the macros.

I plan to make richcmp pass all the tests this weekend.  Do you any more pointers?

mvaled:
> I had to comment the case test expecting a TypeError when Other doesn't implement the reversed method. I think this is one of the things I need to find in the macros.

It wasn't the macro.  It was an error in the Python code of the error.  Fixed.

mvaled:
I managed to also change the implementation of `py_binary_num_func`.  

The next step would be `py_binary_self_func`, but the extract of args down the rabbit hole ;)  I'm following the source code path...

Cargo.toml
```diff
@@ -41,6 +41,8 @@ macros = ["ctor", "indoc", "inventory", "paste", "pyo3cls", "unindent"]
 # Optimizes PyObject to Vec conversion and so on.
 nightly = []
 
+no-return-notimplemented = []
```
davidhewitt, line 44: I think unfortunately this cannot be a feature, because it can change behaviour of other user's code.

Imagine one crate A which uses this feature depends on another crate B which does not. Then suddenly all crate B's APIs change behaviour when compiled as a dependency of crate A.

mvaled: Actually, it doesn't do much right now but opting-out the tests, the actual code would still return NotImplemented.  I found that trying to have the switch led to very ugly code in my hands ;)

I'll remove it.

mvaled:
> If py_binary_self_func is too complicated to change in it's current form, we can discuss thinking about refactors that would make this easier?

My first attempt at it was disastrous :zany_face:.  A change in `_call_impl` so that we could return NotImplemented:

```patch
diff --git a/src/class/macros.rs b/src/class/macros.rs
index e3a0a3401..094466a1e 100644
--- a/src/class/macros.rs
+++ b/src/class/macros.rs
@@ -141,7 +141,7 @@ macro_rules! py_binary_self_func {
             $crate::callback_body!(py, {
                 let slf_ = py.from_borrowed_ptr::<$crate::PyCell<T>>(slf);
                 let arg = py.from_borrowed_ptr::<$crate::PyAny>(arg);
-                call_mut!(slf_, $f, arg).convert(py)?;
+                call_proto_mut!(py, slf_, $f, arg).convert(py)?;
                 ffi::Py_INCREF(slf);
                 Ok(slf)
             })
@@ -385,6 +385,16 @@ macro_rules! _call_impl {
     ($slf: expr, $fn: ident, $raw_arg: expr $(,$raw_args: expr)* $(; $args: expr)*) => {
         _call_impl!($slf, $fn $(,$raw_args)* $(;$args)* ;$raw_arg.extract()?)
     };
+    (proto $py:ident, $slf: expr, $fn: ident, $raw_arg: expr $(,$raw_args: expr)* $(; $args: expr)*) => {
+        _call_impl!(
+            $slf, $fn ;
+            (match $raw_arg.extract() {
+                Ok(res) => res,
+                _=> return Ok($py.NotImplemented().convert($py))
+            })
+            $(;$args)*
+        )
+    }
 }
 
 /// Call `slf.try_borrow()?.$fn(...)`
@@ -400,3 +410,9 @@ macro_rules! call_mut {
         _call_impl!($slf.try_borrow_mut()?, $fn $(,$raw_args)* $(;$args)*)
     };
 }
+
+macro_rules! call_proto_mut {
+    ($py:ident, $slf: expr, $fn: ident $(,$raw_args: expr)* $(; $args: expr)*) => {
+        _call_impl!(proto $py, $slf.try_borrow_mut()?, $fn $(,$raw_args)* $(;$args)*)
+    };
+}

```

You can imagine how that went.  Pages of errors...

tests/test_arithmetics.rs
```diff
@@ -423,3 +423,193 @@ fn rich_comparisons_python_3_type_error() {
...
+            "RC_Self"
+        }
+
+        fn __richcmp__(&self, other: PyRef<'p, Self>, _op: CompareOp) -> PyObject {
+            other.py().None()
+        }
+    }
+
+    #[pyproto]
+    impl<'p> PyNumberProtocol<'p> for RichComparisonToSelf {
```
kngwyu, line 447: I think there are too many tests.
Two or three operators are sufficient.

tests/test_arithmetics.rs
```diff
@@ -423,3 +423,193 @@ fn rich_comparisons_python_3_type_error() {
...
+            py_expect_exception!(
+                py,
+                c2,
+                &format!("class Other: pass\nc2 {} Other()", operator),
+                PyTypeError
+            );
+        }
+    }
+
+    macro_rules! not_implemented_test {
```
kngwyu, line 526: I want to avoid using this macro to generate test *functions*. It makes it more difficult to maintain our tests.
What is important is to place each test with clear purpose.

kngwyu:
> The next step would be py_binary_self_func, but there the extract of args is down the rabbit hole ;) I'm following the source code path...

We don't have to make the situation *perfect* in a PR. Let's do that in the next one.

> no-return-notimplemented = []

So... when do we want not to return `NotImplemented` error?
I don't think we have to implement it.

tests/test_arithmetics.rs
davidhewitt: I disagree here; I tend to think that if it's possible to write a simple macro which generates tests to cover the entire set expected of expected behaviour, then it's easier to maintain that rather than many individual tests to cover all the interface.

But I accept this is a style choice, so willing to cede to your preference @kngwyu

kngwyu: Hmm :thinking: 
It can be true sometimes but in this case, what we do is simply calling `test_dunder`.
So I think simply repeating `test_dunder` is not so bad.
```rust
#[test] 
fn test_operators_return_not_implemented() {
    test_impl("<");
    test_impl(">");
    ....
}
```

tests/test_arithmetics.rs
mvaled: I've just push commit 5a0cdbe1f78f23c8189be1832b6506a44156737b.

I kept all the operators for two reasons:

- It helps to have a list comprehensive list of which operators implement.

- Although right now all the operators are implemented nicely in a macro, having them here can help to detect errors in future refactors.

mvaled: However, I group them in just four tests.

tests/test_arithmetics.rs
mvaled: See https://github.com/PyO3/pyo3/pull/1072#discussion_r464069161.

mvaled:
> We don't have to make the situation perfect in a PR. Let's do that in the next one.

Ok.  Cool.  It buys a little bit of time.

Nevertheless, let me solve one remaining puzzle in this PR:  I think the tests are not catching one error in this PR.  I didn't changed `py_ternary_num_func` which is backing the `__pow__`; and still the tests are passing.  I fail to see why.

mvaled:
Definitively.  There's something wrong in the tests as they stand now.  I added a trivial one focusing on `**` and it fails -- rightfully so.

```rust
    #[test]
    fn mistery_power() {
        let py = gil.python();
        let c2 = PyCell::new(py, RichComparisonToSelf {}).unwrap();
        py_run!(
            py,
            c2,
            "\
class Other:
    def __rpow__(self, other):
        return other

assert c2 ** Other() == c2"
        );
    }
```

mvaled:
Ah! The `mistery_power` was failing for another reason, I'm always returning `None` in `__richcmp__`; chaging to `assert c2 ** Other() is c2` makes it pass. 

And I noticed I didn't implemented `__pow__` for RichComparisonToSelf.

mvaled:
I rewrote the tests to strengthen them a little bit.  Now they properly fail at `**`.

tests/test_arithmetics.rs
mvaled: Reworked the tests in commit 366892fb2bdd96e7b5ad770464907d276e974222.

mvaled:
Good news! 
I've managed to fix the case for inplace operators with the expection of  `**=`.

tests/test_arithmetics.rs
```diff
@@ -423,3 +423,226 @@ fn rich_comparisons_python_3_type_error() {
...
+            py,
+            c2,
+            &format!(
+                "\
+class Other:
+    def __radd__(self, other):
+        return other
+    __rand__ = __ror__ = __rxor__ = __radd__
+    __rsub__ = __rmul__ = __rtruediv__ = __rfloordiv__ = __rpow__ = __radd__
+    __rmatmul__ = __rlshift__ = __rrshift__ = __rmod__ = __rdivmod__ = __radd__
```
davidhewitt, line 570: I don't know whether it's worth it, but if you wanted to make only one definition of `Other` you might be able to do a helper function like:

```
def get_other_type(py: Python) -> &PyAny {
    let locals = PyDict::new(py);
    py.run("""
class Other:
    # put all operators here
""", None, Some(locals)).unwrap();
    locals.get_item("Other").unwrap()
}

**Comment by kngwyu** [comment]
Thank you, but I feel every `match ~.extract()` a bit annoying.
How about adding `PyErr::not_implemted` and using it?
You can do so by just copy-pasting the following snippet to the impl block in `err.rs`.
```rust
    /// Shorthand methods for returning `NotImplementedError`.
    pub fn not_implemented(py: Python) -> Self {
        let value = py.NotImplemented();
        Self {
            ptype: exceptions::PyNotImplementedError::type_object(py).into(),
            pvalue: PyErrValue::Value(value),
            ptraceback: None,
        }
    }
```

Then you can write
```rust
                let arg3 = py
                    .from_borrowed_ptr::<$crate::types::PyAny>(arg3)
                    .extract()
                    .map_err(|_| $crate::PyErr::not_implemented(py));
```
or so.

davidhewitt:
> How about adding PyErr::not_implemted and using it?

`NotImplementedError` and `NotImplemented` are actually two different things, so I'm not sure if that works?

I confess I did wonder myself if we could remove some of the boilerplate. Maybe an `extract_or_return_not_implemented!()` macro in derive_utils?

kngwyu:
Ah, sorry, I misunderstood them.

mvaled:
@kngwyu 

> Thank you, but I feel every match ~.extract() a bit annoying.
> How about adding PyErr::not_implemted and using it? 
> ...

We must not return a `PyErr` in such cases because `NotImplemented` is not (at this point) an exception, but a signal to CPython to look for the reversed method of the second operand.  That's why `_call_impl` returns an `Ok($py.NotImplemented()...)`.

I agree having `matches` everywhere is kinda ugly,  but I don't see how we could avoid it because at this point we need to `return`. Maybe a macro could help for `py_binary_num_func` and in `tp_richcompare`.    But the `match` in `_call_impl` is a bit different because it must return a Result, and most importantly, the '?' operator after `convert(py)`.

mvaled:
> I confess I did wonder myself if we could remove some of the boilerplate. Maybe an extract_or_return_not_implemented!() macro in derive_utils?

Exactly my thoughts.   However, it would only be usable in two of the three cases, because of what I noted in https://github.com/PyO3/pyo3/pull/1072#issuecomment-667686309

kngwyu:
```rust
macro_rules! extract_or_notimplemented {
    ($py: ident, $arg: ident) => {
        match $py
            .from_borrowed_ptr::<$crate::types::PyAny>($arg)
            .extract()
        {
            Ok(value) => value,
            Err(_) => return $py.NotImplemented().convert($py),
        }
    };
}
```

This macro would work for simplification.

kngwyu:
@mvaled 
> But the match in _call_impl is a bit different because it must return a Result, and most importantly, the '?' operator after convert(py).

I don't think we have to take care of the code complexity in macros. `match` appears once in the macro and it's not so bad.

src/class/macros.rs
```diff
@@ -397,3 +411,9 @@ macro_rules! call_mut {
         _call_impl!($slf.try_borrow_mut()?, $fn $(,$raw_args)* $(;$args)*)
     };
 }
+
+macro_rules! call_proto_mut {
```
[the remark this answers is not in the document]
mvaled: Done.  AFAIK, only operators which can be reversed should return NotImplemented.  I would have to check if that actually makes up for the whole set of operators.

tests/test_arithmetics.rs
mvaled: Nice.  Done.

src/class/macros.rs
```diff
@@ -222,13 +221,14 @@ macro_rules! py_ternary_num_func {
...
                     .from_borrowed_ptr::<$crate::types::PyAny>(arg1)
                     .extract()?;
-                let arg2 = py
-                    .from_borrowed_ptr::<$crate::types::PyAny>(arg2)
-                    .extract()?;
-                let arg3 = py
-                    .from_borrowed_ptr::<$crate::types::PyAny>(arg3)
-                    .extract()?;
-
+                let arg2 = match py.from_borrowed_ptr::<$crate::types::PyAny>(arg2).extract() {
```
kngwyu, line 224: Why not use the macro here?

mvaled:
@davidhewitt @kngwyu 

I created the macro `extract_or_return_not_implemented` as in https://github.com/PyO3/pyo3/pull/1072#issuecomment-667686968.  However, I put in `macros.rs`.

tests/test_arithmetics.rs
```diff
@@ -423,3 +424,227 @@ fn rich_comparisons_python_3_type_error() {
     py_expect_exception!(py, c2, "c2 >= 1", PyTypeError);
     py_expect_exception!(py, c2, "1 >= c2", PyTypeError);
 }
+
```
kngwyu, line 427: Maybe it's better to have a comment on this mod?
E.g., `// Checks that undefined binary operations return NotImplemented.`

src/class/macros.rs
mvaled: Oops.  I forgot.

mvaled: Done.

tests/test_arithmetics.rs
mvaled: Done.

src/class/macros.rs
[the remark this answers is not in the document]
mvaled: [`object.__length_hint__`](https://docs.python.org/3/reference/datamodel.html#object.__length_hint__) also allows for NotImplemented but we don't have it.

mvaled:
@davidhewitt 

> Finally, for __ipow__ (i.e. **=) , it uses a separate macro py_dummy_ternary_self_func, which you could just alter in-place (I guess the arg1 extract)

Right now those lines are like this:

```rust
                let arg1 = py.from_borrowed_ptr::<$crate::PyAny>(arg1);
                call_operator_mut!(py, slf_cell, $f, arg1).convert(py)?;
```
`arg1` should be being extracted by `call_operator_mut!`.  

I tried to use `extract_or_return_not_implemented!` here like this:

```rust
                let arg1 = extract_or_return_not_implemented!(py, arg1);
                call_mut!(slf_cell, $f; arg1).convert(py)?;
```

But it didn't work.  I'm gonna give another try.

src/class/macros.rs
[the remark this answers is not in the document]
mvaled: @davidhewitt 

Currently, there are other uses of `call_mut!` in `sequence.rs`: 

- `set_inplace_concat`
- `set_inplace_repeat`

Those methods set the values for `sq_inplace_concat` and `sq_inplace_repeat`.    I detected Python only uses those in:

- PyNumber_InPlaceAdd: https://github.com/python/cpython/blob/6a613f90bf13038255bca028feeab6fad816edfd/Objects/abstract.c#L1096
- PyNumber_InPlaceMultiply: https://github.com/python/cpython/blob/6a613f90bf13038255bca028feeab6fad816edfd/Objects/abstract.c#L1117
- PySequence_InPlaceConcat: https://github.com/python/cpython/blob/6a613f90bf13038255bca028feeab6fad816edfd/Objects/abstract.c#L1643
- PySequence_InPlaceRepeat: https://github.com/python/cpython/blob/6a613f90bf13038255bca028feeab6fad816edfd/Objects/abstract.c#L1668

It seems that these are not used if we fill `nb_inplace_add` or `nb_inplace_multiply`; and we do have those.  So I think there's no *protocol* to fill those slots.

So I prefer to let `call_mut` as it is.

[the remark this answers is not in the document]
davidhewitt: Ok, that sounds good to me, thanks for investigating 👍

mvaled:
I think that the issue behind `__ipow__` is that our tests requires it to call `__rpow__`, but I've just noticed the following in the [docs](https://docs.python.org/3/reference/datamodel.html?highlight=__ipow__#object.__rpow__):

> Note that ternary pow() will not try calling __rpow__() (the coercion rules would become too complicated).

I changed this test to explicitly test that we're returning NotImplemented and it passes.

mvaled:
@davidhewitt @kngwyu 

I think we can considering merging this PR.  I added the entry in the CHANGELOG.

mvaled:
I've just noticed that the [guide](https://pyo3.rs/v0.11.1/class.html#comparison-operators) promised to return NotImplemented for `__richcmp__`:

>     If `other` is not of the type specified in the signature, the generated code will automatically `return NotImplemented`.

Hum... was that implemented previously and got lost?

kngwyu:
I'm sorry but please wait one or two days until I find the time to review the `ipow` problem in detail.

mvaled:
> I'm sorry but please wait for one or two days until I find the time to review the ipow problem in detail.

Ok.  I'm completing the guide, which didn't have a section on PyNumberProtocol.

mvaled:
@kngwyu 

I noticed that `PyNumber_Add` and other operators in CPython expect that *reflected* operators also return NotImplemented and we didn't do that in this PR.  

Should we include such change in this PR?

mvaled:
I'm a little bit puzzled about how Python tells apart `__add__` and `__radd__`, I see that both are *assigned* to the `nb_add` slot.  Yet they are both there:

```python
>>> class X: 
...     def __add__(self, other): 
...         return self 
...     def __radd__(self, other): 
...         return other 
...     

>>> X() + 1                                                                                                                                                                                                        
<__main__.X at 0x7fd308820df0>

>>> 1 + X()                                                                                                                                                                                                        
1
```
:thinking:

mvaled:
@kngwyu 

I confirmed that `__ipow__` doesn't fallback to the reversed operator in Python itself:

```python
>>> class Foo: 
...     def __ipow__(self, other): 
...         return NotImplemented 
...                                                                                                                                                                                                                

>>> class X: 
...     def __rpow__(self, other): 
...         return other 
...                                                                                                                                                                                                                

>>> f = Foo()                                                                                                                                                                                                      
>>> f **= X()                                                                                                                                                                                                      
---------------------------------------------------------------------------
<ipython-input-4-16ca508b06d5> in <module>
----> 1 f **= X()

TypeError: unsupported operand type(s) for ** or pow(): 'Foo' and 'X'
```

mvaled:
> Is there anything else you're planning to do with it?

I was thinking we should return NotImplemented in reversed operators as well.  While studying the code of `py_binary_reversed_num_func` I noticed the `nb_` thing, and took a look a Python's code of `PyNumber_Add`, which basically relies on `binary_op1(v, w, NB_SLOT(nb_add))` (https://github.com/python/cpython/blob/582aaf19e8b94a70c1f96792197770d604ba0fdf/Objects/abstract.c#L945)

But, then, I don't see how CPython itself chooses the `__radd__` of my programs :exploding_head:

mvaled:
> This is actually a bug: #844. There's a little bit of discussion there already. If it interests you, I'd be very happy to mentor you through a solution to that ticket also? (I was planning to look at it myself soon, but so many pieces I'm busy with I'm always happy to share the work!)

I will look at it.  Maybe there's already an explanation there.  If think I'm up to the challenge I'll let you know.

davidhewitt:
> I rather let this PR as it is now and work the __r*__ in another one.

Great! I'm satisfied with this PR, let's just wait for @kngwyu to confirm.

> But, then, I don't see how CPython itself chooses the __radd__ of my programs 🤯

That's pretty much what I'm trying to get to in #844 - CPython doesn't just set `__add__` or `__radd__` directly to the `nb_add` slot, but instead uses a macro to generate a special wrapper function which does different things depending on which of these are present:

https://github.com/python/cpython/blob/05e4a296ecc127641160a04f39cc02c0f66a8c27/Objects/typeobject.c#L6358

guide/src/class.md
```diff
@@ -756,6 +756,91 @@ Each method corresponds to Python's `self.attr`, `self.attr = value` and `del se
 
     Determines the "truthyness" of the object.
 
+### Emulating numeric types
+
+The [`PyNumberProtocol`] trait allows [emulate numeric types](https://docs.python.org/3/reference/datamodel.html?highlight=__ipow__#emulating-numeric-types).
```
kngwyu, line 761: Please don't highlight ipow ...
```suggestion
The [`PyNumberProtocol`] trait allows [emulate numeric types](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).
```

guide/src/class.md
```diff
@@ -756,6 +756,91 @@ Each method corresponds to Python's `self.attr`, `self.attr = value` and `del se
...
+  * `fn __pow__(lhs: impl FromPyObject, rhs: impl FromPyObject, modulo: Option<impl FromPyObject>) -> PyResult<impl ToPyObject>`
+  * `fn __lshift__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __rshift__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __and__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __or__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+  * `fn __xor__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
+
+These methods are called to implement the binary arithmetic operations (`+`,
+`-`, `*`, `@`, `/`, `//`, `%`, `divmod()`, `pow()` and `**`, `<<`, `>>`, `&`,
+`^`, and `|`).
```
kngwyu, line 780: Just a nit, 
```suggestion
These methods are called to implement the binary arithmetic operations
(`+`, `-`, `*`, `@`, `/`, `//`, `%`, `divmod()`, `pow()` and `**`, `<<`, `>>`, `&`, `^`, and `|`).
```

guide/src/class.md
```diff
@@ -756,6 +756,91 @@ Each method corresponds to Python's `self.attr`, `self.attr = value` and `del se
...
+  * `fn __imod__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ipow__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ilshift__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __irshift__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __iand__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ior__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ixor__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+
+The following implement the unary arithmetic operations (`-`, `+`, `abs()` and
+`~`):
```
kngwyu, line 826:
```suggestion
The following methods implement the unary arithmetic operations (`-`, `+`, `abs()` and `~`):
```

kngwyu:
Thank you!
I totally forgot about #844 and, yeah, I agree that we should address it first (and maybe I can do so this weekend).

mvaled:
@kngwyu @davidhewitt 

Thank you for all your guidance and support.

I've committed the suggestions and also added a note to #844.  I was looking at it to see if I could tackle it, but it would take some time before I could go back to it and study the CPython code (some deadlines here to meet). I'll let you know.

kngwyu:
Thank you for all your work!

> I've committed the suggestions and also added a note to #844. I was looking at it to see if I could tackle it, but it would take some time before I could go back to it and study the CPython code (some deadlines here to meet). I'll let you know.

Since our proc-macro implementation for `__r*__` methods is a bit tricky, I'm not sure CPython code helps you.
If you want to tackle it, please let me know that so I can explain the detail. Or I'll implement it next Saturday.

mvaled:
> Since our proc-macro implementation for __r*__ methods is a bit tricky, I'm not sure CPython code helps you. If you want to tackle it, please let me know that so I can explain the detail. Or I'll implement it next Saturday.

Ok.  I think I'll pass on this, but I'll be looking at your work to learn a bit more.  

I'm gonna review the list of issues remaining for 0.12 to see if there's anything I could do.

davidhewitt:
❤️ as well as `0.12` milestone you're also free to investigate anything else which strikes your interest. You might be interested in the new contributing notes I've written which I've meant to do for a while: https://github.com/PyO3/pyo3/pull/1086

Issue #617, opened 2019-10-10
memoryview, python-side allocation


elbaro:
- Feature request: PyMemoryView
- Feature request: PyByteArray::new_with_size(num_bytes)
- Question: Can we allocate some buffer in python-side?

Consider this example:
```
let vec = vec![0;length];
// write on vec
// ...
return PyBytes::new(&vec[..]).to_object(py);
```

PyBytes::new is very slow because it copies. A memoryview on Rust slice works. but it is hard to manage the lifetime. If we can allocate on the python side from the beginning,  the copy is not necessary. Is this possible?

```
let buf = some_allocation(length);
let bytes = buf.as_mut_bytes();
// write directly on bytes
// ...
return bytes;
```

kngwyu:
Thank you for letting us know that.
[rust-numpy](https://github.com/rust-numpy/rust-numpy) has this kind of feature, so maybe we can implement this for `PyBuffer`.
But, since the current `PyBuffer` design is not good and complicated, it needs some more design decisions.

davidhewitt:
I think we might be able to do something here with a closure like `PyBytes::new_with(py, 1024, |b: &mut [u8]| { ... }`.

My thinking is that perhaps during a controlled initialization in this fashion, nothing else can yet access the bytes object so it can be mutated safely.

If anyone strongly wants this feature, please open a PR and we can flesh it out.

davidhewitt:
Closed in #1074

PR #1074, opened 2020-08-01
Added Rust initialisation of Python-allocated bytes

MoritzLangenstein:
Adds `new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyBytes` as suggested by @davidhewitt in #617. This allows initialising new PyBytes in Rust like such:
```rust
let py_bytes = PyBytes::new_with(py, 10, |b: &mut [u8]| {
    b.copy_from_slice(b"Hello Rust");
});
```
Currently, it follows the semantics of `PyBytes::new` in that it panics if a memory error occurs in Python. Maybe my implementation of that could be improved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add `Python::with_gil` for executing a closure with the Python GIL. [#1037](https://github.com/PyO3/pyo3/pull/1037)
 - Implement `Debug` for `PyIterator`. [#1051](https://github.com/PyO3/pyo3/pull/1051)
 - Implement type information for conversion failures. [#1050](https://github.com/PyO3/pyo3/pull/1050)
+- Add `PyBytes::new_with` and `PyBytes::new_with_uninit` for initialising Python allocated bytes inside Rust. [#1074](https://github.com/PyO3/pyo3/pull/1074)
 
 ### Changed
 - Exception types have been renamed from e.g. `RuntimeError` to `PyRuntimeError`, and are now only accessible by `&T` or `Py<T>` similar to other Python-native types. The old names continue to exist but are deprecated. [#1024](https://github.com/PyO3/pyo3/pull/1024)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add `Python::with_gil` for executing a closure with the Python GIL. [#1037](https://github.com/PyO3/pyo3/pull/1037)
 - Implement `Debug` for `PyIterator`. [#1051](https://github.com/PyO3/pyo3/pull/1051)
 - Implement type information for conversion failures. [#1050](https://github.com/PyO3/pyo3/pull/1050)
-- Add `PyBytes::new_with` and `PyBytes::new_with_uninit` for initialising Python allocated bytes inside Rust. [#1074](https://github.com/PyO3/pyo3/pull/1074)
+- Add `PyBytes::new_with` and `PyByteArray::new_with` for initialising Python allocated bytes and bytearrays inside Rust. [#1074](https://github.com/PyO3/pyo3/pull/1074)
 
 ### Changed
 - Exception types have been renamed from e.g. `RuntimeError` to `PyRuntimeError`, and are now only accessible by `&T` or `Py<T>` similar to other Python-native types. The old names continue to exist but are deprecated. [#1024](https://github.com/PyO3/pyo3/pull/1024)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add `Python::with_gil` for executing a closure with the Python GIL. [#1037](https://github.com/PyO3/pyo3/pull/1037)
 - Implement `Debug` for `PyIterator`. [#1051](https://github.com/PyO3/pyo3/pull/1051)
 - Implement type information for conversion failures. [#1050](https://github.com/PyO3/pyo3/pull/1050)
-- Add `PyBytes::new_with` and `PyByteArray::new_with` for initialising Python allocated bytes and bytearrays inside Rust. [#1074](https://github.com/PyO3/pyo3/pull/1074)
+- Add `PyBytes::new_with` and `PyByteArray::new_with` for initialising Python-allocated bytes and bytearrays using a closure. [#1074](https://github.com/PyO3/pyo3/pull/1074)
 
 ### Changed
 - Exception types have been renamed from e.g. `RuntimeError` to `PyRuntimeError`, and are now only accessible by `&T` or `Py<T>` similar to other Python-native types. The old names continue to exist but are deprecated. [#1024](https://github.com/PyO3/pyo3/pull/1024)
--- a/src/types/bytearray.rs
+++ b/src/types/bytearray.rs
@@ -39,7 +39,7 @@ impl PyByteArray {
     /// ```
     pub fn new_with<F>(py: Python, len: usize, init: F) -> &PyByteArray
     where
-        F: FnOnce(&mut [u8])
+        F: FnOnce(&mut [u8]),
     {
         unsafe {
             let length = len as ffi::Py_ssize_t;
--- a/src/types/bytearray.rs
+++ b/src/types/bytearray.rs
@@ -21,9 +21,8 @@ impl PyByteArray {
         unsafe { py.from_owned_ptr::<PyByteArray>(ffi::PyByteArray_FromStringAndSize(ptr, len)) }
     }
 
-    /// Creates a new Python bytearray object.
-    /// The bytearray is zero-initialised and can be read inside `init`.
-    /// The `init` closure can further initialise the bytearray.
+    /// Creates a new Python `bytearray` object with an `init` closure to write its contents.
+    /// Before calling `init` the bytearray is zero-initialised.
     ///
     /// Panics if out of memory.
     ///
@@ -38,7 +37,10 @@ impl PyByteArray {
     ///     assert_eq!(bytearray, b"Hello Rust");
     /// });
     /// ```
-    pub fn new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyByteArray {
+    pub fn new_with<F>(py: Python, len: usize, init: F) -> &PyByteArray
+    where
+        F: FnOnce(&mut [u8])
+    {
         unsafe {
             let length = len as ffi::Py_ssize_t;
             let pyptr = ffi::PyByteArray_FromStringAndSize(std::ptr::null(), length);
@@ -48,7 +50,7 @@ impl PyByteArray {
             debug_assert!(!buffer.is_null());
             // Zero-initialise the uninitialised bytearray
             std::ptr::write_bytes(buffer, 0u8, len);
-            // (Furher) Initialise the bytearray in init
+            // (Further) Initialise the bytearray in init
             init(std::slice::from_raw_parts_mut(buffer, len));
             pybytearray
         }
--- a/src/types/bytearray.rs
+++ b/src/types/bytearray.rs
@@ -21,6 +21,39 @@ impl PyByteArray {
         unsafe { py.from_owned_ptr::<PyByteArray>(ffi::PyByteArray_FromStringAndSize(ptr, len)) }
     }
 
+    /// Creates a new Python bytearray object.
+    /// The bytearray is zero-initialised and can be read inside `init`.
+    /// The `init` closure can further initialise the bytearray.
+    ///
+    /// Panics if out of memory.
+    ///
+    /// # Example
+    /// ```
+    /// use pyo3::{prelude::*, types::PyByteArray};
+    /// Python::with_gil(|py| {
+    ///     let py_bytearray = PyByteArray::new_with(py, 10, |bytes: &mut [u8]| {
+    ///         bytes.copy_from_slice(b"Hello Rust");
+    ///     });
+    ///     let bytearray: &[u8] = unsafe { py_bytearray.as_bytes() };
+    ///     assert_eq!(bytearray, b"Hello Rust");
+    /// });
+    /// ```
+    pub fn new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyByteArray {
+        unsafe {
+            let length = len as ffi::Py_ssize_t;
+            let pyptr = ffi::PyByteArray_FromStringAndSize(std::ptr::null(), length);
+            // Iff pyptr is null, py.from_owned_ptr(pyptr) will panic
+            let pybytearray = py.from_owned_ptr(pyptr);
+            let buffer = ffi::PyByteArray_AsString(pyptr) as *mut u8;
+            debug_assert!(!buffer.is_null());
+            // Zero-initialise the uninitialised bytearray
+            std::ptr::write_bytes(buffer, 0u8, len);
+            // (Furher) Initialise the bytearray in init
+            init(std::slice::from_raw_parts_mut(buffer, len));
+            pybytearray
+        }
+    }
+
     /// Creates a new Python bytearray object from another PyObject that
     /// implements the buffer protocol.
     pub fn from<'p, I>(py: Python<'p>, src: &'p I) -> PyResult<&'p PyByteArray>
@@ -226,5 +259,25 @@ mod test {
 
         bytearray.resize(20).unwrap();
         assert_eq!(20, bytearray.len());
+    }
+
+    #[test]
+    fn test_byte_array_new_with() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let py_bytearray = PyByteArray::new_with(py, 10, |b: &mut [u8]| {
+            b.copy_from_slice(b"Hello Rust");
+        });
+        let bytearray: &[u8] = unsafe { py_bytearray.as_bytes() };
+        assert_eq!(bytearray, b"Hello Rust");
+    }
+
+    #[test]
+    fn test_byte_array_new_with_zero_initialised() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let py_bytearray = PyByteArray::new_with(py, 10, |_b: &mut [u8]| ());
+        let bytearray: &[u8] = unsafe { py_bytearray.as_bytes() };
+        assert_eq!(bytearray, &[0; 10]);
     }
 }
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -2,7 +2,6 @@ use crate::{
     ffi, AsPyPointer, FromPy, FromPyObject, PyAny, PyObject, PyResult, PyTryFrom, Python,
     ToPyObject,
 };
-use std::mem::MaybeUninit;
 use std::ops::Index;
 use std::os::raw::c_char;
 use std::slice::SliceIndex;
@@ -27,25 +26,6 @@ impl PyBytes {
         unsafe { py.from_owned_ptr(ffi::PyBytes_FromStringAndSize(ptr, len)) }
     }
 
-    /// Creates a new Python bytestring object and a mutable slice to its underlying buffer.
-    /// The bytestring is uninitialised and must not be read.
-    ///
-    /// Panics if out of memory.
-    #[inline]
-    unsafe fn new_with_uninit_impl(
-        py: Python<'_>,
-        len: usize,
-    ) -> (&mut [MaybeUninit<u8>], &PyBytes) {
-        let length = len as ffi::Py_ssize_t;
-        let pyptr = ffi::PyBytes_FromStringAndSize(std::ptr::null(), length);
-        // Iff pyptr is null, py.from_owned_ptr(pyptr) will panic
-        let pybytes = py.from_owned_ptr(pyptr);
-        let buffer = ffi::PyBytes_AsString(pyptr) as *mut MaybeUninit<u8>;
-        debug_assert!(!buffer.is_null());
-        let slice = std::slice::from_raw_parts_mut(buffer, len);
-        (slice, pybytes)
-    }
-
     /// Creates a new Python bytestring object.
     /// The bytestring is zero-initialised and can be read inside `init`.
     /// The `init` closure can further initialise the bytestring.
@@ -65,47 +45,19 @@ impl PyBytes {
     /// });
     /// ```
     pub fn new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyBytes {
-        let (slice, pybytes) = unsafe { Self::new_with_uninit_impl(py, len) };
-        slice.iter_mut().for_each(|x| *x = MaybeUninit::zeroed());
-        let slice: &mut [u8] = unsafe { &mut *(slice as *mut [MaybeUninit<u8>] as *mut [u8]) };
-        init(slice);
-        pybytes
-    }
-
-    /// Creates a new Python bytestring object.
-    ///
-    /// # Safety
-    /// * The bytestring is uninitialised and must not be read inside `init`.
-    /// * The entire bytestring must be written to inside `init` such that any code that
-    /// follows does not read uninitialised memory.
-    ///
-    /// Panics if out of memory.
-    ///
-    /// # Example
-    /// ```
-    /// use pyo3::{prelude::*, types::PyBytes};
-    /// use std::mem::MaybeUninit;
-    /// Python::with_gil(|py| -> PyResult<()> {
-    ///     let py_bytes =
-    ///         unsafe { PyBytes::new_with_uninit(py, 10, |uninit_bytes: &mut [MaybeUninit<u8>]| {
-    ///             uninit_bytes
-    ///                 .iter_mut()
-    ///                 .zip(b"Hello Rust".iter())
-    ///                 .for_each(|(ub, b)| *ub = MaybeUninit::new(*b));
-    ///         }) };
-    ///     let bytes: &[u8] = FromPyObject::extract(py_bytes)?;
-    ///     assert_eq!(bytes, b"Hello Rust");
-    ///     Ok(())
-    /// });
-    /// ```
-    pub unsafe fn new_with_uninit<F: Fn(&mut [MaybeUninit<u8>])>(
-        py: Python<'_>,
-        len: usize,
-        init: F,
-    ) -> &PyBytes {
-        let (slice, pybytes) = Self::new_with_uninit_impl(py, len);
-        init(slice);
-        pybytes
+        unsafe {
+            let length = len as ffi::Py_ssize_t;
+            let pyptr = ffi::PyBytes_FromStringAndSize(std::ptr::null(), length);
+            // Iff pyptr is null, py.from_owned_ptr(pyptr) will panic
+            let pybytes = py.from_owned_ptr(pyptr);
+            let buffer = ffi::PyBytes_AsString(pyptr) as *mut u8;
+            debug_assert!(!buffer.is_null());
+            // Zero-initialise the uninitialised bytestring
+            std::ptr::write_bytes(buffer, 0u8, len);
+            // (Furher) Initialise the bytestring in init
+            init(std::slice::from_raw_parts_mut(buffer, len));
+            pybytes
+        }
     }
 
     /// Creates a new Python byte string object from a raw pointer and length.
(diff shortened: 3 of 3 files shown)


MoritzLangenstein:
I have also implemented a more specialised variant
```rust
new_with_and_truncate<F: Fn(&mut [u8]) -> usize>(
    py: Python<'_>,
    len: usize,
    init_and_truncate: F,
) -> &PyBytes
```
where the closure returns the truncated length of the final PyBytes. I think this might be too specific of a use case for this PR, but I can add it as well if you would be interested. This change can be viewed at https://github.com/PyO3/pyo3/compare/master...MoritzLangenstein:py_bytes_new_with_and_truncate.

kngwyu:
With this function, we can get uninitialized `PyBytes` by
```rust
fn do_nothing(_bytes: &mut [u8]) {}
let py_bytes = PyBytes::new_with(py, 10, do_nothing);
```

So I think simply we should expose `unsafe fn uninitialized` or so

src/types/bytes.rs
```diff
@@ -26,6 +26,21 @@ impl PyBytes {
         unsafe { py.from_owned_ptr(ffi::PyBytes_FromStringAndSize(ptr, len)) }
     }
 
+    /// Creates a new Python bytestring object.
+    /// The uninitialised bytestring must be initialised in the `init` closure.
```
davidhewitt, line 30: Maybe specifically say the _entire_ uninitialised bytestring?

Also, if you're up for it, adding a little example would be very much appreciated.

src/types/bytes.rs
```diff
@@ -26,6 +26,21 @@ impl PyBytes {
...
+    ///
+    /// Panics if out of memory.
+    pub fn new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyBytes {
+        let length = len as ffi::Py_ssize_t;
+        let pyptr = unsafe { ffi::PyBytes_FromStringAndSize(std::ptr::null(), length) };
```
davidhewitt, line 35: As this and the next 4 lines all contain unsafe, could perhaps just refactor to be a single unsafe block?

davidhewitt:
> With this function, we can get uninitialized PyBytes

Ah, this is a very good point. Is it actually UB if we allow that? Might need to consult cpython docs further. What if we make this function `unsafe`, or alternatively always zero-initialise the bytes object before calling `init`?

MoritzLangenstein:
> With this function, we can get uninitialized `PyBytes` by
> 
> ```rust
> fn do_nothing(_bytes: &mut [u8]) {}
> let py_bytes = PyBytes::new_with(py, 10, do_nothing);
> ```
> 
> So I think simply we should expose `unsafe fn uninitialized` or so

Would it be possible to use MaybeUninit here? I’ve never tried wrapping an existing value in it (and it will unfortunately take until tonight before I get back to my computer), but maybe we could pass the bytes explicitly uninitialised into the closure and return an initialised slice with the same lifetime from the closure? One problem I could see with that, though, would be that you could then return subslices ... On second thought, default-initialising the bytes or making the function unsafe might be easier.

kngwyu:
> Would it be possible to use MaybeUninit here? 

It's basically `union {(), T}` so not applicable here directly.

> On second thought, default-initialising the bytes or making the function unsafe might be easier.
> What if we make this function unsafe, or alternatively always zero-initialise the bytes object before calling init?

Both of them are OK, but maybe zero-initialization is better?

MoritzLangenstein:
Ok, let’s go with zero-initialisation then? In that case, should we explicitly add support for a zero- and / or an unsafe uninitialised constructor (neither of which would require a closure)? Personally, I can’t think of a use-case for that in Rust code right now, but maybe you might have some ideas.

davidhewitt:
How about `new_with` zero-initializes first, and `unsafe new_with_uninit` which has a `# Safety` note pointing out two things:
- the slice should not be read, because it's uninitialised
- the entire slice must be written, so that code which follows does not read uninitialised memory

davidhewitt:
I'm also happy only exposing the safe API if it's simpler. An extra write to the whole array can't be _that_ bad, can it? :)

MoritzLangenstein:
> * Regarding the truncate form, if you have examples of uses which are common enough then I'd be happy to consider it.

I initially came across #617 after having implemented pickling for my wrapper class as suggested in https://gist.github.com/ethanhs/fd4123487974c91c7e5960acc9aa2a77. The data structure I am serialising to bytes is large enough that the copying from Vec<u8> to PyBytes can lead to OutOfMemory errors. Therefore, I have a small compacting algorithm which reduces the size of the serde + bincode output. For this algorithm I have an allocation upper bound, which allows me to do the compaction in place. Afterwards, I can then usually truncate the buffer. For me, pushing the allocation of the buffer to Python and being able to truncate it after its initialisation would make the entire serialisation zero-copy. This is probably a very specific use-case.

MoritzLangenstein:
>     * Can you check if any other objects should support the same API? For example maybe `PyString::new_with` or `PyTuple::new_with`. I haven't thought about this too hard, so it might not make sense for them. If it does, would be great to support a uniform API across the whole library!

@davidhewitt I have looked through the native types and think that only `str`, `bytes` and `bytearray` should share this API, as only they have the `Py..._FromStringAndSize` constructor methods (tuples unfortunately do not, see https://docs.python.org/3/c-api/tuple.html).

For the `PyString`, there still exists `PyUnicode_FromStringAndSize(const char *u, Py_ssize_t size)` but setting u to `NULL`
has been deprecated in favour of `PyUnicode_New(Py_ssize_t size, Py_UCS4 maxchar)`. However, to use the latter we would have to know the largest char value, i.e. does it fit into 1-4 bytes, up front. How should we go about this - should we just expose `maxchar` as a required parameter as well?

src/types/bytes.rs
MoritzLangenstein: Fixed in 20d27ff050b57a5ceb446c881efaf301b97b4cda

src/types/bytes.rs
MoritzLangenstein: Where and in what format should I add an example?

davidhewitt:
Many thanks, will give this all a thorough read through in the morning!

For the example, you can just add an `# Example` section to the docstring of the method in question. e.g. like this: https://github.com/PyO3/pyo3/blob/51171f7475cdc671693c5c6f48350569e6dfac62/src/python.rs#L56

kngwyu:
> For the PyString, there still exists PyUnicode_FromStringAndSize(const char *u, Py_ssize_t size) but setting u to NULL
has been deprecated in favour of PyUnicode_New(Py_ssize_t size, Py_UCS4 maxchar). However, to use the latter we would have to know the largest char value, i.e. does it fit into 1-4 bytes, up front. How should we go about this - should we just expose maxchar as a required parameter as well?

I don't think we need `PyString::new_with`

src/types/bytes.rs
```diff
@@ -26,6 +26,60 @@ impl PyBytes {
...
+    /// Panics if out of memory.
+    pub fn new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyBytes {
+        let (slice, pybytes) = unsafe { Self::new_with_uninit_impl(py, len) };
+        #[cfg(feature = "slice_fill")]
+        {
+            slice.fill(0);
+        }
+        #[cfg(not(feature = "slice_fill"))]
+        {
+            slice.iter_mut().for_each(|x| *x = 0);
```
kngwyu, line 59: This is optimized by the compiler and not so slow.
I don't see any clear reason we should use (unstable) `fill` here. It's actually for loop.

programmerjake:
Technically, even making a `&mut [u8]` that points to uninitialized bytes is potentially undefined behavior even if you don't read from it (assuming Rust decides that uninitialized integers are undefined behavior, they still haven't decided but it's better to be safe than sorry).

I think you may want to use either `*mut u8` or `&mut [MaybeUninit<u8>]`, since they are specifically designed to handle pointing to uninitialized data.

src/types/bytes.rs
```diff
@@ -26,6 +26,60 @@ impl PyBytes {
...
+
+    /// Creates a new Python bytestring object.
+    ///
+    /// # Safety
+    /// * The bytestring is uninitialised and must not be read inside `init`.
+    /// * The entire bytestring must be written to inside `init` such that any code that
+    /// follows does not read uninitialised memory.
+    ///
+    /// Panics if out of memory.
+    pub unsafe fn new_with_uninit<F: Fn(&mut [u8])>(
```
kngwyu, line 73: Do we need this?

src/types/bytes.rs
```diff
@@ -26,6 +26,60 @@ impl PyBytes {
...
+        let buffer = ffi::PyBytes_AsString(pyptr) as *mut u8;
+        debug_assert!(!buffer.is_null());
+        let slice = std::slice::from_raw_parts_mut(buffer, len);
+        (slice, pybytes)
+    }
+
+    /// Creates a new Python bytestring object.
+    /// The `init` closure can initialise the bytestring.
+    ///
+    /// # Safety
```
kngwyu, line 47: We don't need `#Safety` document for safe functions

src/types/bytes.rs
kngwyu: If so, the closure should take `&mut [MaybeUninit<u8>]` as @programmerjake says.

kngwyu:
Thank you for the change. Let's make it simpler and add `#Example` as David says.

src/types/bytes.rs
MoritzLangenstein: I've added examples in 533c105b73d214afdb14033f45b1659cf0723233

src/types/bytes.rs
MoritzLangenstein: Fixed in 533c105b73d214afdb14033f45b1659cf0723233

src/types/bytes.rs
MoritzLangenstein: Changed in 533c105b73d214afdb14033f45b1659cf0723233

src/types/bytes.rs
MoritzLangenstein: Fixed in 533c105b73d214afdb14033f45b1659cf0723233

MoritzLangenstein:
I have attempted to implement all of your suggestions. One everyone is satisfied with the current API, I would then implement it for `PyByteArray` as well.

> I don't think we need PyString::new_with

`PyString` would be the only other remaining candidate I could think of, but it probably does not usually transport as much data around as `PyBytes`. I'll follow your recommendations here.

kngwyu:
Why do we need `new_with_uninit`?

MoritzLangenstein:
On an unrelated note, could you mention `cargo clippy --tests` and `cargo test --doc` (and any other checks made by your CI pipeline) in `Contributing.md`? Personally, I was, unfortunately, not aware that I had to invoke these explicitly in order to catch all the errors your pipeline does.

MoritzLangenstein:
> Why do we need new_with_uninit?

We could also remove it entirely - I just included it so we could see what an interface using `MaybeUninit` would look like.

kngwyu:
> On an unrelated note, could you mention cargo clippy --tests and cargo test --doc (and any other checks made by your CI pipeline) in Contributing.md? 

Agreed that `clippy --tests` should be mentioned, but I think `cargo test` runs all doc tests by default

> We could also remove it entirely - I just included it so we could see what an interface using MaybeUninit would look like.

If you don't have any use case, I think it's better to remove it.

davidhewitt:
> could you mention cargo clippy --tests and cargo test --doc (and any other checks made by your CI pipeline) in Contributing.md?

Good suggestion, see #1075 - I also plan to extend CONTRIBUTING.md later today.

> We could also remove it entirely - I just included it so we could see what an interface using MaybeUninit would look like.

Yeah I was thinking when it was essentially `new_with` but without the zero-intialisation it made more sense. Now, I think most users will choose to use `new_with` anyway. OTOH, it shows how complicated it is to get this API well-defined, so maybe it still serves some use? I don't have a strong opinion either way.

src/types/bytes.rs
```diff
@@ -26,6 +27,87 @@ impl PyBytes {
...
+    ///         bytes.copy_from_slice(b"Hello Rust");
+    ///     });
+    ///     let bytes: &[u8] = FromPyObject::extract(py_bytes)?;
+    ///     assert_eq!(bytes, b"Hello Rust");
+    ///     Ok(())
+    /// });
+    /// ```
+    pub fn new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyBytes {
+        let (slice, pybytes) = unsafe { Self::new_with_uninit_impl(py, len) };
+        slice.iter_mut().for_each(|x| *x = MaybeUninit::zeroed());
```
davidhewitt, line 69: If you choose to remove `new_with_uninit`, you may find `std::ptr::write_bytes` a nice alternative to making a `MaybeUninit` slice.

davidhewitt:
Also, on assessment I agree `PyString::new_with` is probably not so useful (and also I'm not sure it's easy to implement safely; the Python representation of unicode data has lots of representations for different platforms / optimisations).

For `PyByteArray::new_with` - I agree this can work. It's not documented, but it looks like `PyByteArray_FromStringAndSize` also returns an uninitialized object if the first argument is null: https://github.com/python/cpython/blob/8963a7f1f84a05412178b56629508b660d38861b/Objects/bytearrayobject.c#L145

Many thanks for your continued revisions to this PR - it's looking really really good 👍

MoritzLangenstein:
> Agreed that clippy --tests should be mentioned, but I think cargo test runs all doc tests by default

Mhm, maybe I was doing something wrong then. But I checked the docs and you are completely right that `cargo test` should run documentation tests by default.

src/types/bytes.rs
MoritzLangenstein: Thanks for the suggestion - I’ll keep that in mind.

programmerjake:
Read through the code, LGTM!

MoritzLangenstein:
I've thought about `PyBytes::new_with_uninit` and using it seemed to much of a hastle that I think most people would stick with the safe version anyway (and zeroing memory should be extremely well optimised anyway) - so as per @kngwyu's suggestion I've removed the method. I've integrated @davidhewitt's suggestion to use `std::ptr::write_bytes` for the zeroing. Finally, I've added `PyByteArray::new_with` which is a functional clone just with differently named API calls.

src/types/bytearray.rs
```diff
@@ -21,6 +21,39 @@ impl PyByteArray {
...
+        unsafe {
+            let length = len as ffi::Py_ssize_t;
+            let pyptr = ffi::PyByteArray_FromStringAndSize(std::ptr::null(), length);
+            // Iff pyptr is null, py.from_owned_ptr(pyptr) will panic
+            let pybytearray = py.from_owned_ptr(pyptr);
+            let buffer = ffi::PyByteArray_AsString(pyptr) as *mut u8;
+            debug_assert!(!buffer.is_null());
+            // Zero-initialise the uninitialised bytearray
+            std::ptr::write_bytes(buffer, 0u8, len);
+            // (Furher) Initialise the bytearray in init
```
davidhewitt, line 51:
```suggestion
            // (Further) Initialise the bytearray in init
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add `Python::with_gil` for executing a closure with the Python GIL. [#1037](https://github.com/PyO3/pyo3/pull/1037)
 - Implement `Debug` for `PyIterator`. [#1051](https://github.com/PyO3/pyo3/pull/1051)
 - Implement type information for conversion failures. [#1050](https://github.com/PyO3/pyo3/pull/1050)
+- Add `PyBytes::new_with` and `PyByteArray::new_with` for initialising Python allocated bytes and bytearrays inside Rust. [#1074](https://github.com/PyO3/pyo3/pull/1074)
```
davidhewitt, line 13:
```suggestion
- Add `PyBytes::new_with` and `PyByteArray::new_with` for initialising Python-allocated bytes and bytearrays using a closure. [#1074](https://github.com/PyO3/pyo3/pull/1074)
```

src/types/bytearray.rs
```diff
@@ -21,6 +21,39 @@ impl PyByteArray {
...
+    /// use pyo3::{prelude::*, types::PyByteArray};
+    /// Python::with_gil(|py| {
+    ///     let py_bytearray = PyByteArray::new_with(py, 10, |bytes: &mut [u8]| {
+    ///         bytes.copy_from_slice(b"Hello Rust");
+    ///     });
+    ///     let bytearray: &[u8] = unsafe { py_bytearray.as_bytes() };
+    ///     assert_eq!(bytearray, b"Hello Rust");
+    /// });
+    /// ```
+    pub fn new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyByteArray {
```
davidhewitt, line 41: I think we can make a stronger guarantee here (that the closure will only be called once).

```suggestion
    pub fn new_with<F>(py: Python, len: usize, init: F) -> &PyByteArray
    where
        F: FnOnce(&mut [u8])
    {
```

src/types/bytes.rs
```diff
@@ -26,6 +26,40 @@ impl PyBytes {
         unsafe { py.from_owned_ptr(ffi::PyBytes_FromStringAndSize(ptr, len)) }
     }
 
+    /// Creates a new Python bytestring object.
+    /// The bytestring is zero-initialised and can be read inside `init`.
+    /// The `init` closure can further initialise the bytestring.
```
davidhewitt, line 31: Same as above

```suggestion
    /// Creates a new Python `bytes` object with an `init` closure to write its contents.
    /// Before calling `init` the contents are zero-initialised.
```

src/types/bytearray.rs
```diff
@@ -21,6 +21,39 @@ impl PyByteArray {
         unsafe { py.from_owned_ptr::<PyByteArray>(ffi::PyByteArray_FromStringAndSize(ptr, len)) }
     }
 
+    /// Creates a new Python bytearray object.
+    /// The bytearray is zero-initialised and can be read inside `init`.
+    /// The `init` closure can further initialise the bytearray.
```
davidhewitt, line 26:
```suggestion
    /// Creates a new Python `bytearray` object with an `init` closure to write its contents.
    /// Before calling `init` the bytearray is zero-initialised.
```

src/types/bytes.rs
```diff
@@ -26,6 +26,40 @@ impl PyBytes {
...
+    /// Python::with_gil(|py| -> PyResult<()> {
+    ///     let py_bytes = PyBytes::new_with(py, 10, |bytes: &mut [u8]| {
+    ///         bytes.copy_from_slice(b"Hello Rust");
+    ///     });
+    ///     let bytes: &[u8] = FromPyObject::extract(py_bytes)?;
+    ///     assert_eq!(bytes, b"Hello Rust");
+    ///     Ok(())
+    /// });
+    /// ```
+    pub fn new_with<F: Fn(&mut [u8])>(py: Python<'_>, len: usize, init: F) -> &PyBytes {
```
davidhewitt, line 47: Same as above

```suggestion
    pub fn new_with<F>(py: Python, len: usize, init: F) -> &PyBytes
    where
        F: FnOnce(&mut [u8])
    {
```

src/types/bytes.rs
```diff
@@ -26,6 +26,40 @@ impl PyBytes {
...
+        unsafe {
+            let length = len as ffi::Py_ssize_t;
+            let pyptr = ffi::PyBytes_FromStringAndSize(std::ptr::null(), length);
+            // Iff pyptr is null, py.from_owned_ptr(pyptr) will panic
+            let pybytes = py.from_owned_ptr(pyptr);
+            let buffer = ffi::PyBytes_AsString(pyptr) as *mut u8;
+            debug_assert!(!buffer.is_null());
+            // Zero-initialise the uninitialised bytestring
+            std::ptr::write_bytes(buffer, 0u8, len);
+            // (Furher) Initialise the bytestring in init
```
davidhewitt, line 57: Same as above

```suggestion
            // (Further) Initialise the bytestring in init
```

MoritzLangenstein:
Does anyone know what went wrong with the tests there? I don't think this PR should have affected the iterator type - maybe we just came across a probabilistic bug?

davidhewitt:
I think looks like a probabilistic bug to me. As it looks quite hard to isolate and unrelated to this PR, I'm going to re-run CI and see what happens. (Any test which accesses the `None` singleton could be the cause of that.)

davidhewitt:
As CI is green, I'm going to proceed to merge. Thanks very much for your work on this ❤️

MoritzLangenstein:
Thank you very much to @davidhewitt, @kngwyu and @programmerjake for all your feedback and assistance in getting the idea into a great (and merged) PR! It has been an absolute joy to collaborate with you ❤️

davidhewitt:
👍 glad to hear! If you feel like joining in for any more - please feel free to claim any issue you want to work on, or contribute to any design discussion, or write improved docs / examples, or even just hang out on Gitter and help others to use pyo3. (I'm going to write a more extensive CONTRIBUTING.md soon 😄)

MoritzLangenstein:
I am sorry that I have to follow up on this discussion, but I came accross an interesting problem. (How) Can we support failure inside the `init` closure? Right now, we are implicitly assuming that nothing inside it can fail, which might not be the case. One way to resolve this might be to accept a `Result<R, E>` as a return type of `init` and to return `Result<(&PyBytes, R), E>` from `new_with`. What do you think?

davidhewitt:
Ah, great question! We _could_ perhaps modify the return type of `init` to be `PyResult<()>`?

Note that because `F: FnOnce` it's possible to mutate references to other data outside the closure, so I don't think that we need to support an arbitrary return type `R`.

Im also happy to leave the API it as-is, and let users who need error handling use a similar trick as suggested above for other return values. It'd be less ergonomic, so it depends if we expect users of this API to typically be doing fallible operations inside `init`.

I guess we the answer is yes, because if they already had the data materialised then `PyBytes::new` would already be sufficient.

davidhewitt:
In fact, reflecting on the original truncate case further, perhaps the return type for `init` should be `PyResult<Option<usize>>`, where:
- if the return value is `Ok(Some(x))` then the bytes are truncated to length x
- if the return value is `Ok(None)`, this is equivalent to the current behaviour
- if the return value is `Err(e)` then the bytes are deallocates immediately and the error is returned.

Also, if we change this function to be fallible, I think we should also remove the panic for out-of-memory from this API and related functions.

Does the above seem reasonable?

MoritzLangenstein:
I also tried out the approach of mutating an outside variable - it just felt very un-Rusty to mutate a result variable inside a closure so it can be read outside. But for just returning some value, I think mutating an outside variable could be fine.

MoritzLangenstein:
> In fact, reflecting on the original truncate case further, perhaps the return type for `init` should be `PyResult<Option<usize>>`, where:
> 
>     * if the return value is `Ok(Some(x))` then the bytes are truncated to length x
> 
>     * if the return value is `Ok(None)`, this is equivalent to the current behaviour
> 
>     * if the return value is `Err(e)` then the bytes are deallocated immediately and the error is returned.
> 
> 
> Also, if we change this function to be fallible, I think we should also remove the panic for out-of-memory from this API and related functions.
> 
> Does the above seem reasonable?

That sounds like a very reasonable approach. Should we change the general `new` such that it returns a result as well?

davidhewitt:
> Should we change the general new such that it returns a result as well?

Yes please :)

I think when the allocation fails a `MemoryError` should actually already be already be set by the underlying c apis?

birkenfeld:
I don't see why PyO3 should go out of its way not panicking for out-of-memory while the standard library does.

MoritzLangenstein:
After fiddling around with the required changes to the API, I agree that returning `PyResult` from all constructors for `PyBytes` and `PyByteArray` makes the code more cumbersome and will just lead to a lot of `unwrap()`s. We could either remain at not catching any OutOfMemory errors or just catching them when we already return a result (though that semantic inconsistency might then become confusing as well). At the same time, I think forcing the user to think about those error cases also has advantages -- especially if they can forward them.

MoritzLangenstein:
If the `init` closure returns an error, do we need to free the allocated bytes in some way?

davidhewitt:
> If the init closure returns an error, do we need to free the allocated bytes in some way?

I think you would just need to call Py_DECREF on the pointer instead of `py.from_owned_ptr`.

davidhewitt:
> I don't see why PyO3 should go out of its way not panicking for out-of-memory while the standard library does.

That's a reasonable opinion, and I'm happy to leave this as-is if you feel strongly about so. My motivation was that the equivalent Python just raises `MemoryError` exception, and usually for APIs where we have a Python exception possibly being raised, we return `PyResult`.

```
>>> bytearray(10**11)
MemoryError
```

From a quick check, at the moment it looks like our only constructors which return `PyResult` are for `PySet` / `PyFrozenSet` and the datetime bindings.

birkenfeld:
And `Py::new`, unfortunately.

davidhewitt:
> And Py::new, unfortunately.

Yeah, that's a tricky one. While `MemoryError` is likely to be the error there, it's also possible for arbitrary user code to run e.g. from the `tp_alloc` slot.

Have you got examples where `Py::new` returning `PyResult` is not ergonomic? My general perception is that because `#[pyfunction]` and `#[pymethods]` can return `PyResult`, it's easy to handle APIs that return `PyResult` with `?` unwrapping.

MoritzLangenstein:
I think if we wanted to change the API such that out-of-memory panics are caught, it would have to be updated everywhere and would be a significant and breaking change. Allowing the initialisation of `PyBytes` and `PyByteArray` to fail is a separate issue. Should I add the error handling (and truncation) to `new_with`? If so, should `new_with` still panic on out-of-memory errors or already report them as `Err`? Since `new_with` would presumably be most useful in memory-constrained scenaries, I think reporting memory errors would be helpful here.

davidhewitt:
Agreed, it's a big change, so let's leave existing constructors as-is for now. Perhaps at some point we can review them across the library including the `Py::new` case.

Regarding `new_with` - I'm happy if you open a PR to add the error handling and truncation to it, I think it's useful in memory-constrained situations. And I agree that `new_with` should return `PyErr` on out-of-memory - this matches the behaviour of our other fallible constructors as well as being potentially helpful, as you say.

birkenfeld:
> Yeah, that's a tricky one. While MemoryError is likely to be the error there, it's also possible for arbitrary user code to run e.g. from the tp_alloc slot.

Right, although I would have expected that for the types that are allowed there no fancy `tp_alloc` is possible.

> Have you got examples where Py::new returning PyResult is not ergonomic?

I was using some closure heavy code where it can be a pain to thread the Results through to the toplevel. But I don't think it's a big issue.

davidhewitt:
> Right, although I would have expected that for the types that are allowed there no fancy `tp_alloc` is possible.

This is true and I agree, though we might need to refactor some internals. I'll open an issue to discuss.

MoritzLangenstein:
One issue I have come across is that testing for a memory error is quite difficult cross-platform. On my machine, the OOM kicks in instead (because the virtual address space allows the allocation) and just kills the program instead. Should we just do a best effort error handling (i.e. if we are lucky enough to encounter the error in Rust we will return an `Err`, but it might just kill your program as well) or just go back to the "it will panic/abort in some way" semantic? Is there a way to test this behaviour?

MoritzLangenstein:
In the test for a failing initialisation (i.e. `init` returns `Err`), I would like to assert that the execution has not leaked any objects. Is there a way to do so in the Rust testsuite?

davidhewitt:
> On my machine, the OOM kicks in instead (because the virtual address space allows the allocation) and just kills the program instead. Should we just do a best effort error handling (i.e. if we are lucky enough to encounter the error in Rust we will return an Err, but it might just kill your program as well) or just go back to the "it will panic/abort in some way" semantic? Is there a way to test this behaviour?

Ah, that's a fair point. I think when we're hitting limits of the machine like this it's hard to guarantee particular execution, so I'm happy to just leave it as "we'll try to give you an error if possible". I'm willing to accept that there probably can't be a test written for this case.

> In the test for a failing initialisation (i.e. init returns Err), I would like to assert that the execution has not leaked any objects. Is there a way to do so in the Rust testsuite?

Good question. I'm actually not aware of any C API for this. I wonder if you can call into Python's [`tracemalloc`](https://docs.python.org/3/library/tracemalloc.html) module to get what you need?

If writing such a test is massively complicated, again I'm willing to forgo a test as long as the implementation is appropriately commented (e.g. `// XXX: Make sure to deallocate here or bytes will be leaked!`)

MoritzLangenstein:
I have found two different ways of calling into `tracemalloc`:

1) Using https://docs.python.org/3/c-api/memory.html#tracemalloc-c-api (I have not tried that yet as the relevant `ffi` is incomplete in `pyo3`)
2) Using `Python::run`: this results in the following import error for `tracemalloc`: `.pyenv/versions/3.8.3/lib/python3.8/lib-dynload/_struct.cpython-38-x86_64-linux-gnu.so: undefined symbol: PyFloat_Type`

One slightly unrelated note: when I call `Py_REFCNT` inside `init` (which is now called before `py.from_owned_ptr(pyptr)`), the refcount is 0. Is that the expected behaviour?

davidhewitt:
> Using Python::run: this results in the following import error for tracemalloc: .pyenv/versions/3.8.3/lib/python3.8/lib-dynload/_struct.cpython-38-x86_64-linux-gnu.so: undefined symbol: PyFloat_Type

Looks like you're using pyenv, so this discussion might be the cause of that: #763

> One slightly unrelated note: when I call Py_REFCNT inside init (which is now called before py.from_owned_ptr(pyptr)), the refcount is 0. Is that the expected behaviour?

That's surprising to me, I thought it would be 1. Will need to check the cpython source to figure that out.

MoritzLangenstein:
> Looks like you're using pyenv, so this discussion might be the cause of that: #763

Thank you for the suggestion - in the end setting `config.shared` to `true` seemed to fix it. But from the discussion it was not clear to me where the config was so I had to grep for it. Should that information be included in the README as well or is that a developer-only fix?

MoritzLangenstein:
> > One slightly unrelated note: when I call Py_REFCNT inside init (which is now called before py.from_owned_ptr(pyptr)), the refcount is 0. Is that the expected behaviour?
>
> That's surprising to me, I thought it would be 1. Will need to check the cpython source to figure that out.

Maybe I was just measuring it wrong - anyways, I am now able to produce the behaviour where an allocation of 1MB bytestring without `ffi::Py_DECREF` has a memory difference >1MB whereas an allocation with `ffi::Py_DECREF` has a memory difference of >10kB. That should suffice to create a leak test.

PR #1092, opened 2020-08-07
Experimental support for PyPy 3.6

davidhewitt:
I added Windows and MacOS actions builds for PyPy. This caused quite a bit of pain; eventually I did manage to get things building.

Issues became apparent on pypy initially when I tried to use it for https://github.com/PyO3/setuptools-rust/pull/78


src/types/num.rs
```diff
@@ -216,17 +161,141 @@ int_convert_u64_or_i64!(
     ffi::PyLong_AsUnsignedLongLong
 );
 
-#[cfg(not(Py_LIMITED_API))]
-int_convert_128!(i128, 16, 1);
-#[cfg(not(Py_LIMITED_API))]
-int_convert_128!(u128, 16, 0);
+// manual implementation for 128bit integers
+#[cfg(not(any(Py_LIMITED_API, PyPy)))]
+mod int128_conversion {
```
kngwyu, line 166: What is your intention of this change?

.github/workflows/test.yml
```diff
@@ -9,35 +9,63 @@ jobs:
...
+        run: cargo build --features "num-bigint num-complex" --verbose
+
+      # Run tests (except on PyPy, because no embedding API).
+      - if: matrix.python-version != 'pypy3'
+        name: Test
+        run: cargo test --features "num-bigint num-complex"
+
+      - name: Test proc-macro code
+        shell: bash
+        run: cd pyo3-derive-backend && cargo test
```
kngwyu, line 54:
```suggestion
        run: cargo test --manifest-path=pyo3-derive-backend/Cargo.tom
```

src/types/num.rs
davidhewitt: Without grouping in a `mod`, I had to add many `#[cfg(not(PyPy)]` annotations (e.g. also for all the tests).

As this mod is private and only contains trait implementations, I thought it would keep our code more tidy to group in a module with a single top-level `#[cfg]`.

.github/workflows/test.yml
davidhewitt: Thanks, nice one! 🚀

davidhewitt:
>  It looks like you forgot to remove ci/actions/test.sh?

Good catch, removed now. I mistakenly thought it was used in the `Makefile` for some reason.

src/ffi/pyerrors.rs
```diff
@@ -281,9 +282,11 @@ extern "C" {
     #[cfg_attr(PyPy, link_name = "PyPyExc_TimeoutError")]
     pub static mut PyExc_TimeoutError: *mut PyObject;
 
+    #[cfg(not(all(windows, PyPy)))]
     pub static mut PyExc_EnvironmentError: *mut PyObject;
+    #[cfg(not(all(windows, PyPy)))]
     pub static mut PyExc_IOError: *mut PyObject;
```
messense, line 288: Since `IOError` is just a alias of `OSError` on Python 3, I think we can expose `PyExc_IOError` as `PyPyExc_OSError`?

Use case: https://github.com/milesgranger/pyrus-cramjam/runs/2227542030?check_suite_focus=true

```
Installing collected packages: cramjam
Successfully installed cramjam-2.2.0
ImportError: /opt/hostedtoolcache/PyPy/3.6.12/x64/site-packages/cramjam.pypy36-pp73-x86_64-linux-gnu.so: undefined symbol: PyExc_IOError
```

https://github.com/milesgranger/pyrus-cramjam/pull/48/commits/f60849c4f2f9992e2b7525b4479887be8808c22b#diff-76866598ce8fd16261a27ac58a84b2825e6e77fc37c163a6afa60f0f4477e569R343

kngwyu: Hmm, but does it seem that PyPy takes a different strategy for that error?
https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/exceptions/moduledef.py#L21

messense: I'm not sure, but it seems to be the same on the Python side on my local PyPy installation.

```
Python 3.7.10 (7035471788c8, Mar 19 2021, 02:02:14)
[PyPy 7.3.5-alpha0 with GCC Apple LLVM 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
And now for something completely different: ``finally, mercurial migration is
happening!''
>>>> IOError
<class 'OSError'>
>>>> OSError
<class 'OSError'>
>>>> IOError is OSError
True
```

messense: Actually it is the same on pypy 3.6 and 3.7 branch https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.7/pypy/module/exceptions/moduledef.py#L31

kngwyu: Ah, I got it, thanks.

PR #1093, opened 2020-08-08
Improve lifetime insertions for #[pyproto]

kngwyu:
This PR allows:
```rust
#[pyproto]
impl PyIterProtocol for Iter {
    fn __iter__(slf: PyRef<Self>) -> PyRef<Self> {
        slf
    }
}
```


pyo3-derive-backend/src/func.rs
```diff
@@ -108,19 +113,8 @@ pub(crate) fn impl_method_proto(
...
-                {
-                    if let syn::GenericArgument::Lifetime(_) = args.args[0] {
-                    } else {
-                        let lt = syn::parse_quote! {'p};
-                        args.args.insert(0, lt);
-                    }
-                }
-            }
+            // Insert lifetime to PyRef (i.e., PyRef<Self> -> PyRef<'p, Self>)
+            insert_lifetime_to_pyref(&mut slf_ty);
```
davidhewitt, line 117: I think you should add this inside `get_arg_ty` instead of calling here.

pyo3-derive-backend/src/func.rs
```diff
@@ -77,7 +77,12 @@ pub(crate) fn impl_method_proto(
 ) -> TokenStream {
     let ret_ty = match &sig.output {
         syn::ReturnType::Default => quote! { () },
-        syn::ReturnType::Type(_, ty) => ty.to_token_stream(),
+        syn::ReturnType::Type(_, ty) => {
+            let mut ty = ty.clone();
+            // Insert lifetime to PyRef (i.e., PyRef<Self> -> PyRef<'p, Self>)
+            insert_lifetime_to_pyref(&mut ty);
```
davidhewitt, line 83: It would be nice if we could support adding lifetimes to references too. I notice that `get_arg_ty` does this.

kngwyu:
Thanks, I made it more general. Now your `descr` example works too.

> I was actually thinking about a refactoring the other day which refactors the MethodProto enum to a struct instead. I can't find my branch right now, but this was the essence of the change:

It sounds nice since the current implementation is crafty. However, I want to leave it for another chance, since I have to complete another PR first (for #844).

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Don't rely on the order of structmembers to compute offsets in PyCell. Related to
   [#1058](https://github.com/PyO3/pyo3/pull/1058). [#1059](https://github.com/PyO3/pyo3/pull/1059)
 - Allows `&Self` as a `#[pymethods]` argument again. [#1071](https://github.com/PyO3/pyo3/pull/1071)
+- Improve lifetime insertion in `#[pyproto]`. [#1093](https://github.com/PyO3/pyo3/pull/1093)
```
davidhewitt, line 38: I think elision is the technical term for the user-facing behavior.

```suggestion
- Improve lifetime elision in `#[pyproto]`. [#1093](https://github.com/PyO3/pyo3/pull/1093)
```

PR #1094, opened 2020-08-08
Move fmt and clippy to actions

davidhewitt:
I think that having these checks in actions makes it easier to see when it fails, rather than having to look at the bottom of the min-stable travis job.

I made it so that tests don't bother running if fmt fails, but I can remove that "needs" if you think it's not so nice. I think it's a good habit to not waste energy running tests for a commit we won't merge! 😄

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,31 @@ name: Test
 on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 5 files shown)


davidhewitt:
I slightly preferred having two separate as they're different tools, but didn't mind too strongly so have merged them into one.

Issue #1082, opened 2020-08-05
extension-module should link libpython for Android and Cygwin


davidhewitt:
When browsing documentation earlier this morning I came across a note suggesting that `libpython` should be linked on both `android` and `cygwin` targets (see last paragraph [here](https://docs.python.org/3/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build)).

At the moment this is not the case.

We should change `build.rs` so that even with `extension-module` set (or whatever mechanism might replace `extension-module`) we always link to `libpython` on these targets.

Progdrasil:
Not really sure what target `cygwin` would correspond to in the [target platforms], `x86_64-pc-windows-gnu`?

[target platforms]: https://forge.rust-lang.org/release/platform-support.html

davidhewitt:
> Not really sure what target cygwin would correspond to in the target platforms, x86_64-pc-windows-gnu?

Good question. I can't see evidence that Rust provides a target for cygwin. I think if it existed it might be `x64_64-pc-cygwin`, which is based on briefly trawling the internet and finding [this stackoverflow post](https://stackoverflow.com/questions/32624082/how-to-use-clang-in-cygwin).

IMO cygwin is increasingly redundant with WSL anyway, so I think let's ignore support for it for now until somebody comes along asking.

Progdrasil:
fair enough, makes dealing with one less exception :smile:

davidhewitt:
#1095 solved this for android, and at the moment there is no Rust cygwin target (afaik), so closing this issue for now.

Issue #1077, opened 2020-08-02
Cross compile linking issue: cannot locate symbol "_Py_NoneStruct"


Progdrasil:
## 🐛 Bug Description

I need to cross compile a pyo3 library for android using [kivy/buildozer](https://github.com/kivy/buildozer). I've set the `LD_LIBRARY_PATH`, `PYO3_CROSS_INCLUDE_DIR` and `PYO3_CROSS_LIB_DIR` environment variables and am using [`cargo ndk`](https://github.com/bbqsrc/cargo-ndk) for the build process and copying the `.so` to the site_packages directory.

Everything compiles correctly, unfortunately at runtime I get the message `cannot locate symbol "_Py_NoneStruct"`. 

### 🌍 Environment

 - Your operating system and version: Ubuntu 20.04 (docker)
 - Your python version: 3.8
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: compiled from source
 - Your Rust version (`rustc --version`): 1.45.0
 - Your PyO3 version: 0.11.1
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3"`)?: yes

### 💥 Reproducing

I have a repo for the [minimal working example](https://github.com/Progdrasil/pyo3-buildozer). The necessary commands are shown in the README.md

davidhewitt:
Thanks for posting. This looks like a packaging issue, possibly in `kivy/python-for-android`. Here's an example of a very similar looking upstream bug from 2016 which was closed. https://github.com/kivy/python-for-android/issues/887

As they will have much better knowledge than I for diagnosing the buildozer config, I suggest reporting upstream there before I attempt any debugging here.

Progdrasil:
I'll check that out, however I did have a similar issue when compiling the same project in [termux](https://termux.com/) I'm not sure if this constitutes a different issue, but it is the same target.

```log
ImportError: dlopen failed: cannot locate symbol "PyExc_BaseException" referenced by "/data/data/com.termux/files/home/pyo3-buildozer/target/debug/my_lib.so"
```

Progdrasil:
After reading through the comments it seems what has fixed it was adding the `LDFLAG+="-lpythonX.Ym"` flag. I did get a linking issue when removing the `extension-module` where the linking flags had `-lpython` without the version. So it could not find the appropriate python library.

davidhewitt:
[...]

PR #923, opened 2020-05-12
Fix cross-compilation from unix to windows

davidhewitt:
I think `build.rs` functions slightly incorrectly for cross-compilation.

If I understand correctly, the build script is always compiled for the *host* platform, so we can't actually use `#[cfg(target_os = "macos")]` to set link args, for example. Instead we need to check the `CARGO_CFG_TARGET_OS` env var which is set during build script execution.

This might fix #666. I want to work with feedback from users before we merge this PR.

PR #1095, opened 2020-08-08
Load compilation options from _sysconfigdata_*.py file

Progdrasil:
Following the discussion in #1077 this change allows the compilation
script to load the configurations from a `_sysconfigdata_*.py` file
in the library directory.

This file is also provided on target systems in the same directory.
At least on Manjaro Linux.
Which could remove the need to run a python script at compile time
for compiling the the host.

I've also addressed the linking need for android in #1082.

The paths provided are now checked if they exist.

## Future improvements
The build script subsequently grew quite a bit.
I will probably contribute to https://github.com/davidhewitt/python-config in order to refactor most of the python configuration logic there. It should also improve the debugability of the build script.

~~Another issue I encountered was when I miss-spelled the path to one of the cross-compilation paths, checking if the paths supplied are valid before even trying to get the configs would be a good simple improvement. This would also help consumers of the library when they inevitably make mistakes.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `PyObject` is now just a type alias for `Py<PyAny>`. [#1063](https://github.com/PyO3/pyo3/pull/1063)
 - Implement `Send + Sync` for `PyErr`. `PyErr::new`, `PyErr::from_type`, `PyException::py_err` and `PyException::into` have had these bounds added to their arguments. [#1067](https://github.com/PyO3/pyo3/pull/1067)
 - Change `#[pyproto]` to return NotImplemented for operators for which Python can try a reversed operation. [1072](https://github.com/PyO3/pyo3/pull/1072)
+- Change method for getting cross-compilation configurations using the file which provides the `sysconfig` module with its data, `_sysconfig_*.py` located in the lib directory. [1077](https://github.com/PyO3/pyo3/issues/1077)
 
 ### Removed
 - Remove `PyString::as_bytes`. [#1023](https://github.com/PyO3/pyo3/pull/1023)
@@ -42,6 +43,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Allows `&Self` as a `#[pymethods]` argument again. [#1071](https://github.com/PyO3/pyo3/pull/1071)
 - Fix best-effort build against PyPy 3.6. #[1092](https://github.com/PyO3/pyo3/pull/1092)
 - Improve lifetime elision in `#[pyproto]`. [#1093](https://github.com/PyO3/pyo3/pull/1093)
+- Linking against libpython when compiling for android even with `extension-module` set[#1082](https://github.com/PyO3/pyo3/issues/1082)
 
 ## [0.11.1] - 2020-06-30
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `PyObject` is now just a type alias for `Py<PyAny>`. [#1063](https://github.com/PyO3/pyo3/pull/1063)
 - Implement `Send + Sync` for `PyErr`. `PyErr::new`, `PyErr::from_type`, `PyException::py_err` and `PyException::into` have had these bounds added to their arguments. [#1067](https://github.com/PyO3/pyo3/pull/1067)
 - Change `#[pyproto]` to return NotImplemented for operators for which Python can try a reversed operation. [1072](https://github.com/PyO3/pyo3/pull/1072)
-- Change method for getting cross-compilation configurations using the file which provides the `sysconfig` module with its data, `_sysconfig_*.py` located in the lib directory. [1077](https://github.com/PyO3/pyo3/issues/1077)
 
 ### Removed
 - Remove `PyString::as_bytes`. [#1023](https://github.com/PyO3/pyo3/pull/1023)
@@ -43,7 +42,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Allows `&Self` as a `#[pymethods]` argument again. [#1071](https://github.com/PyO3/pyo3/pull/1071)
 - Fix best-effort build against PyPy 3.6. #[1092](https://github.com/PyO3/pyo3/pull/1092)
 - Improve lifetime elision in `#[pyproto]`. [#1093](https://github.com/PyO3/pyo3/pull/1093)
-- Linking against libpython when compiling for android even with `extension-module` set[#1082](https://github.com/PyO3/pyo3/issues/1082)
+- Fix python configuration detection when cross-compiling. [1077](https://github.com/PyO3/pyo3/issues/1077)
+- Link against libpython on android with `extension-module` set. [#1082](https://github.com/PyO3/pyo3/issues/1082)
 
 ## [0.11.1] - 2020-06-30
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Allows `&Self` as a `#[pymethods]` argument again. [#1071](https://github.com/PyO3/pyo3/pull/1071)
 - Fix best-effort build against PyPy 3.6. #[1092](https://github.com/PyO3/pyo3/pull/1092)
 - Improve lifetime elision in `#[pyproto]`. [#1093](https://github.com/PyO3/pyo3/pull/1093)
-- Fix python configuration detection when cross-compiling. [1077](https://github.com/PyO3/pyo3/issues/1077)
-- Link against libpython on android with `extension-module` set. [#1082](https://github.com/PyO3/pyo3/issues/1082)
+- Fix python configuration detection when cross-compiling. [#1095](https://github.com/PyO3/pyo3/pull/1095)
+- Link against libpython on android with `extension-module` set. [#1095](https://github.com/PyO3/pyo3/pull/1095)
 
 ## [0.11.1] - 2020-06-30
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,8 +32,8 @@ trybuild = "1.0.23"
 rustversion = "1.0"
 
 [build-dependencies]
-walkdir = "~2.3"
-regex = "~1.3"
+walkdir = {version = "~2.3", optional = true }
+regex = { version = "~1.3", optional = true }
 
 [features]
 default = ["macros"]
@@ -53,6 +53,9 @@ extension-module = []
 # many compilation errors. Pull Requests working towards fixing that
 # are welcome.
 # abi3 = []
+
+# Use this feature when cross-compiling the library
+cross-compile = ["walkdir", "regex"]
 
 [workspace]
 members = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,10 @@ assert_approx_eq = "1.1.0"
 trybuild = "1.0.23"
 rustversion = "1.0"
 
+[build-dependencies]
+walkdir = "~2.3"
+regex = "~1.3"
+
 [features]
 default = ["macros"]
 macros = ["ctor", "indoc", "inventory", "paste", "pyo3cls", "unindent"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,10 +31,6 @@ assert_approx_eq = "1.1.0"
 trybuild = "1.0.23"
 rustversion = "1.0"
 
-[build-dependencies]
-walkdir = {version = "~2.3", optional = true }
-regex = { version = "~1.3", optional = true }
-
 [features]
 default = ["macros"]
 macros = ["ctor", "indoc", "inventory", "paste", "pyo3cls", "unindent"]
@@ -53,9 +49,6 @@ extension-module = []
 # many compilation errors. Pull Requests working towards fixing that
 # are welcome.
 # abi3 = []
-
-# Use this feature when cross-compiling the library
-cross-compile = ["walkdir", "regex"]
 
 [workspace]
 members = [
--- a/build.rs
+++ b/build.rs
@@ -270,12 +270,10 @@ fn find_sysconfigdata(path: impl AsRef<Path>, cross: &CrossPython) -> Option<Pat
                 } else {
                     &cross.os
                 }) {
-                    println!("{:?}", f);
                     continue;
                 }
                 // Check if right arch
                 if !name.to_string_lossy().contains(&cross.arch) {
-                    println!("{:?}", f);
                     continue;
                 }
                 find_sysconfigdata(f.path(), cross)
@@ -316,7 +314,7 @@ fn load_cross_compile_from_sysconfigdata(
 
     let interpreter_config = InterpreterConfig {
         version: python_version,
-        libdir: Some(python_paths.lib_dir.to_owned()),
+        libdir: Some(python_paths.lib_dir),
         shared,
         ld_version,
         base_prefix: "".to_string(),
@@ -363,7 +361,7 @@ fn load_cross_compile_from_headers(
 
     let interpreter_config = InterpreterConfig {
         version: python_version,
-        libdir: Some(python_paths.lib_dir.to_owned()),
+        libdir: Some(python_paths.lib_dir),
         shared,
         ld_version: format!("{}.{}", major, minor),
         base_prefix: "".to_string(),
@@ -382,10 +380,10 @@ fn load_cross_compile_info(
     // Because compiling for windows on linux still includes the unix target family
     if target_family == "unix" {
         // Configure for unix platforms using the sysconfigdata file
-        return load_cross_compile_from_sysconfigdata(python_paths);
+        load_cross_compile_from_sysconfigdata(python_paths)
     } else {
         // Must configure by headers on windows platform
-        return load_cross_compile_from_headers(python_paths);
+        load_cross_compile_from_headers(python_paths)
     }
 }
 
--- a/build.rs
+++ b/build.rs
@@ -175,7 +175,6 @@ fn parse_sysconfigdata(config_path: impl AsRef<Path>) -> Result<HashMap<String, 
 }
 
 fn load_cross_compile_from_sysconfigdata(
-    python_include_dir: &Path,
     python_lib_dir: &str,
 ) -> Result<(InterpreterConfig, HashMap<String, String>)> {
     // find info from sysconfig
@@ -207,7 +206,7 @@ fn load_cross_compile_from_sysconfigdata(
 
     let (major, minor) = match config_map.get("VERSION") {
         Some(s) => {
-            let split = s.split(".").collect::<Vec<&str>>();
+            let split = s.split('.').collect::<Vec<&str>>();
             (split[0].parse::<u8>()?, split[1].parse::<u8>()?)
         }
         None => bail!("Could not find python version"),
@@ -237,9 +236,10 @@ fn load_cross_compile_from_sysconfigdata(
 }
 
 fn load_cross_compile_from_headers(
-    python_include_dir: &Path,
+    python_include_dir: &str,
     python_lib_dir: &str,
 ) -> Result<(InterpreterConfig, HashMap<String, String>)> {
+    let python_include_dir = Path::new(&python_include_dir);
     let patchlevel_defines = parse_header_defines(python_include_dir.join("patchlevel.h"))?;
 
     let major = match patchlevel_defines
@@ -294,13 +294,12 @@ fn load_cross_compile_info(
     python_include_dir: String,
     python_lib_dir: String,
 ) -> Result<(InterpreterConfig, HashMap<String, String>)> {
-    let python_include_dir = Path::new(&python_include_dir);
     // Try to configure from the sysconfigdata file which is more accurate for the information
     // provided at python's compile time
-    match load_cross_compile_from_sysconfigdata(python_include_dir, &python_lib_dir) {
+    match load_cross_compile_from_sysconfigdata(&python_lib_dir) {
         Ok(ret) => Ok(ret),
         // If the config could not be loaded by sysconfigdata, failover to configuring from headers
-        Err(_) => load_cross_compile_from_headers(python_include_dir, &python_lib_dir),
+        Err(_) => load_cross_compile_from_headers(&python_include_dir, &python_lib_dir),
     }
 }
 
--- a/build.rs
+++ b/build.rs
@@ -134,10 +134,112 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
     config_map
 }
 
-fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, String>)> {
-    let python_include_dir = env::var("PYO3_CROSS_INCLUDE_DIR")?;
-    let python_include_dir = Path::new(&python_include_dir);
+fn parse_sysconfigdata(config_path: impl AsRef<Path>) -> Result<HashMap<String, String>> {
+    let config_reader = BufReader::new(File::open(config_path)?);
+    let mut entries = HashMap::new();
+    let entry_re = regex::Regex::new(r#"'([a-zA-Z_0-9]*)': ((?:"|')([\S ]*)(?:"|')|\d+)($|,$)"#)?;
+    let subsequent_re = regex::Regex::new(r#"\s+(?:"|')([\S ]*)(?:"|')($|,)"#)?;
+    let mut previous_finished = None;
+    for maybe_line in config_reader.lines() {
+        let line = maybe_line?;
+        if previous_finished.is_none() {
+            let captures = match entry_re.captures(&line) {
+                Some(c) => c,
+                None => continue,
+            };
+            let key = captures[1].to_owned();
+            let val = if let Some(val) = captures.get(3) {
+                val.as_str().to_owned()
+            } else {
+                captures[2].to_owned()
+            };
+            if &captures[4] != "," && &captures[4] != "}" {
+                previous_finished = Some(key.clone());
+            }
+            entries.insert(key, val);
+        } else if let Some(ref key) = previous_finished {
+            let captures = match subsequent_re.captures(&line) {
+                Some(c) => c,
+                None => continue,
+            };
+            let prev = entries.remove(key).unwrap();
+            entries.insert(key.clone(), prev + &captures[1]);
 
+            if &captures[2] == "," || &captures[2] == "}" {
+                previous_finished = None;
+            }
+        }
+    }
+
+    Ok(entries)
+}
+
+fn load_cross_compile_from_sysconfigdata(
+    python_include_dir: &Path,
+    python_lib_dir: &str,
+) -> Result<(InterpreterConfig, HashMap<String, String>)> {
+    // find info from sysconfig
+    // first find sysconfigdata file
+    let sysconfig_re = regex::Regex::new(r"_sysconfigdata_m?_linux_([a-z_\-0-9]*)?\.py$")?;
+    let mut walker = walkdir::WalkDir::new(&python_lib_dir).into_iter();
+    let sysconfig_path = loop {
+        let entry = match walker.next() {
+            Some(Ok(entry)) => entry,
+            None => bail!("Could not find sysconfigdata file"),
+            _ => continue,
+        };
+        let entry = entry.into_path();
+        if sysconfig_re.is_match(entry.to_str().unwrap()) {
+            break entry;
+        }
+    };
+    let config_map = parse_sysconfigdata(sysconfig_path)?;
+
+    let shared = match config_map
+        .get("Py_ENABLE_SHARED")
+        .map(|x| x.as_str())
+        .ok_or("Py_ENABLE_SHARED is not defined")?
+    {
+        "1" | "true" | "True" => true,
+        "0" | "false" | "False" => false,
+        _ => panic!("Py_ENABLE_SHARED must be a bool (1/true/True or 0/false/False"),
+    };
+
+    let (major, minor) = match config_map.get("VERSION") {
+        Some(s) => {
+            let split = s.split(".").collect::<Vec<&str>>();
+            (split[0].parse::<u8>()?, split[1].parse::<u8>()?)
+        }
+        None => bail!("Could not find python version"),
+    };
+
+    let ld_version = match config_map.get("LDVERSION") {
+        Some(s) => s.clone(),
+        None => format!("{}.{}", major, minor),
+    };
+    let python_version = PythonVersion {
+        major,
+        minor: Some(minor),
+        implementation: PythonInterpreterKind::CPython,
+    };
+
+    let interpreter_config = InterpreterConfig {
+        version: python_version,
+        libdir: Some(python_lib_dir.to_owned()),
+        shared,
+        ld_version,
+        base_prefix: "".to_string(),
+        executable: PathBuf::new(),
+        calcsize_pointer: None,
+    };
+
+    Ok((interpreter_config, fix_config_map(config_map)))
+}
+
+fn load_cross_compile_from_headers(
+    python_include_dir: &Path,
+    python_lib_dir: &str,
+) -> Result<(InterpreterConfig, HashMap<String, String>)> {
     let patchlevel_defines = parse_header_defines(python_include_dir.join("patchlevel.h"))?;
 
     let major = match patchlevel_defines
@@ -177,15 +279,29 @@ fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, Strin
 
     let interpreter_config = InterpreterConfig {
         version: python_version,
-        libdir: Some(env::var("PYO3_CROSS_LIB_DIR")?),
+        libdir: Some(python_lib_dir.to_owned()),
         shared,
-        ld_version: "".to_string(),
+        ld_version: format!("{}.{}", major, minor),
         base_prefix: "".to_string(),
         executable: PathBuf::new(),
         calcsize_pointer: None,
     };
 
     Ok((interpreter_config, fix_config_map(config_map)))
+}
+
+fn load_cross_compile_info(
+    python_include_dir: String,
+    python_lib_dir: String,
+) -> Result<(InterpreterConfig, HashMap<String, String>)> {
+    let python_include_dir = Path::new(&python_include_dir);
+    // Try to configure from the sysconfigdata file which is more accurate for the information
+    // provided at python's compile time
+    match load_cross_compile_from_sysconfigdata(python_include_dir, &python_lib_dir) {
+        Ok(ret) => Ok(ret),
+        // If the config could not be loaded by sysconfigdata, failover to configuring from headers
+        Err(_) => load_cross_compile_from_headers(python_include_dir, &python_lib_dir),
+    }
 }
 
 /// Examine python's compile flags to pass to cfg by launching
@@ -567,10 +683,27 @@ fn main() -> Result<()> {
     // If you have troubles with your shell accepting '.' in a var name,
     // try using 'env' (sorry but this isn't our fault - it just has to
     // match the pkg-config package name, which is going to have a . in it).
-    let cross_compiling =
-        env::var("PYO3_CROSS_INCLUDE_DIR").is_ok() && env::var("PYO3_CROSS_LIB_DIR").is_ok();
+    //
+    // Detecting if cross-compiling by checking if the target triple is different from the host
+    // rustc's triple.
+    let cross_compiling = env::var("TARGET") != env::var("HOST");
     let (interpreter_config, mut config_map) = if cross_compiling {
-        load_cross_compile_info()?
+        // If cross compiling we need the path to the cross-compiled include dir and lib dir, else
+        // fail quickly and loudly
+        let python_include_dir = match env::var("PYO3_CROSS_INCLUDE_DIR") {
+            Ok(v) => v,
+            Err(_) => bail!(
+                "Must provide PYO3_CROSS_INCLUDE_DIR environment variable when cross-compiling"
+            ),
+        };
+        let python_lib_dir = match env::var("PYO3_CROSS_LIB_DIR") {
+            Ok(v) => v,
+            Err(_) => {
+                bail!("Must provide PYO3_CROSS_LIB_DIR environment variable when cross-compiling")
+            }
+        };
+
+        load_cross_compile_info(python_include_dir, python_lib_dir)?
     } else {
         find_interpreter_and_get_config()?
     };
--- a/build.rs
+++ b/build.rs
@@ -227,12 +227,9 @@ fn parse_script_output(output: &str) -> HashMap<String, String> {
 
 /// Parse sysconfigdata file
 ///
-/// The sysconfigdata is basically a dictionary, and since we can't really use this library to read
-/// it for us, egg and chicken type thing, we parse it with some regex. Here there are two regex's.
-/// The first one is to match an entry in the dictionary (`entry_re`). Then when the entry is on
-/// multiple lines we use the second regex to capture the additional string entry and add it to the
-/// previously captured key. We detect if this is a multi line entry by checking if the last capture
-/// group contains either `,` or `}`.
+/// The sysconfigdata is simply a dictionary containing all the build time variables used for the
+/// python executable and library. Here it is read and added to a script to extract only what is
+/// necessary. This necessitates a python interpreter for the host machine to work.
 fn parse_sysconfigdata(config_path: impl AsRef<Path>) -> Result<HashMap<String, String>> {
     let mut script = fs::read_to_string(config_path)?;
     script += r#"
@@ -265,6 +262,37 @@ fn ends_with(entry: &DirEntry, pat: &str) -> bool {
     name.to_string_lossy().ends_with(pat)
 }
 
+/// Finds the `_sysconfigdata*.py` file in the library path
+///
+/// From the python source this file is always going to be located at `build/lib.{PLATFORM}-{PY_MINOR_VERSION}`
+/// when built from source. The [exact line][1] is defined as:
+///
+/// ```py
+/// pybuilddir = 'build/lib.%s-%s' % (get_platform(), sys.version_info[:2])
+/// ```
+///
+/// Where get_platform returns a kebab-case formated string containing the os, the architecture and
+/// possibly the os' kernel version (not the case on linux). However, when installed using a package
+/// manager, the `_sysconfigdata*.py` file is installed in the `${PREFIX}/lib/python3.Y/` directory.
+/// The `_sysconfigdata*.py` is generally in a sub-directory of the location of `libpython3.Y.so`.
+/// So we must find the file in the following possible locations:
+///
+/// ```sh
+/// # distribution from package manager, lib_dir should include lib/
+/// ${INSTALL_PREFIX}/lib/python3.Y/_sysconfigdata*.py
+/// ${INSTALL_PREFIX}/lib/libpython3.Y.so
+/// ${INSTALL_PREFIX}/lib/python3.Y/config-3.Y-${HOST_TRIPLE}/libpython3.Y.so
+///
+/// # Built from source from host
+/// ${CROSS_COMPILED_LOCATION}/build/lib.linux-x86_64-Y/_sysconfigdata*.py
+/// ${CROSS_COMPILED_LOCATION}/libpython3.Y.so
+///
+/// # if cross compiled, kernel release is only present on certain OS targets.
+/// ${CROSS_COMPILED_LOCATION}/build/lib.{OS}(-{OS-KERNEL-RELEASE})?-{ARCH}-Y/_sysconfigdata*.py
+/// ${CROSS_COMPILED_LOCATION}/libpython3.Y.so
+/// ```
+///
+/// [1]: https://github.com/python/cpython/blob/3.5/Lib/sysconfig.py#L389
 fn find_sysconfigdata(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Option<PathBuf> {
     for f in fs::read_dir(path).expect("Path does not exist") {
         return match f {
@@ -665,9 +693,10 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<String> {
     }
 
     check_target_architecture(interpreter_config)?;
+    let target_os = env::var_os("CARGO_CFG_TARGET_OS").unwrap();
 
     let is_extension_module = env::var_os("CARGO_FEATURE_EXTENSION_MODULE").is_some();
-    if !is_extension_module || cfg!(target_os = "windows") {
+    if !is_extension_module || target_os == "windows" || target_os == "android" {
         println!("{}", get_rustc_link_lib(&interpreter_config)?);
         if let Some(libdir) = &interpreter_config.libdir {
             println!("cargo:rustc-link-search=native={}", libdir);
--- a/build.rs
+++ b/build.rs
@@ -297,7 +297,7 @@ fn ends_with(entry: &DirEntry, pat: &str) -> bool {
 /// [1]: https://github.com/python/cpython/blob/3.5/Lib/sysconfig.py#L389
 fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
     let mut sysconfig_paths = search_lib_dir(&cross.lib_dir, &cross);
-    if sysconfig_paths.len() == 0 {
+    if sysconfig_paths.is_empty() {
         bail!(
             "Could not find either libpython.so or _sysconfigdata*.py in {}",
             cross.lib_dir.display()
--- a/build.rs
+++ b/build.rs
@@ -76,56 +76,84 @@ impl FromStr for PythonInterpreterKind {
     }
 }
 
-struct CrossPython {
-    lib_dir: String,
-    include_dir: Option<String>,
+trait GetPrimitive {
+    fn get_bool(&self, key: &str) -> Result<bool>;
+    fn get_numeric<T: FromStr>(&self, key: &str) -> Result<T>;
+}
+
+impl GetPrimitive for HashMap<String, String> {
+    fn get_bool(&self, key: &str) -> Result<bool> {
+        match self
+            .get(key)
+            .map(|x| x.as_str())
+            .ok_or(format!("{} is not defined", key))?
+        {
+            "1" | "true" | "True" => Ok(true),
+            "0" | "false" | "False" => Ok(false),
+            _ => Err(format!("{} must be a bool (1/true/True or 0/false/False", key).into()),
+        }
+    }
+
+    fn get_numeric<T: FromStr>(&self, key: &str) -> Result<T> {
+        self.get(key)
+            .ok_or(format!("{} is not defined", key))?
+            .parse::<T>()
+            .map_err(|_| format!("Could not parse value of {}", key).into())
+    }
+}
+
+struct CrossCompileConfig {
+    lib_dir: PathBuf,
+    include_dir: Option<PathBuf>,
     os: String,
     arch: String,
 }
 
-impl CrossPython {
+impl CrossCompileConfig {
     fn both() -> Result<Self> {
-        Ok(CrossPython {
-            include_dir: Some(CrossPython::validate_variable("PYO3_CROSS_INCLUDE_DIR")?),
-            ..CrossPython::lib_only()?
+        Ok(CrossCompileConfig {
+            include_dir: Some(CrossCompileConfig::validate_variable(
+                "PYO3_CROSS_INCLUDE_DIR",
+            )?),
+            ..CrossCompileConfig::lib_only()?
         })
     }
 
     fn lib_only() -> Result<Self> {
-        Ok(CrossPython {
-            lib_dir: CrossPython::validate_variable("PYO3_CROSS_LIB_DIR")?,
+        Ok(CrossCompileConfig {
+            lib_dir: CrossCompileConfig::validate_variable("PYO3_CROSS_LIB_DIR")?,
             include_dir: None,
             os: env::var("CARGO_CFG_TARGET_OS").unwrap(),
             arch: env::var("CARGO_CFG_TARGET_ARCH").unwrap(),
         })
     }
 
-    fn validate_variable(var: &str) -> Result<String> {
-        let path = match env::var(var) {
-            Ok(v) => v,
-            Err(_) => bail!(
+    fn validate_variable(var: &str) -> Result<PathBuf> {
+        let path = match env::var_os(var) {
+            Some(v) => v,
+            None => bail!(
                 "Must provide {} environment variable when cross-compiling",
                 var
             ),
         };
 
         if fs::metadata(&path).is_err() {
-            bail!("{} value of {} does not exist", var, path)
+            bail!("{} value of {:?} does not exist", var, path)
         }
 
-        Ok(path)
+        Ok(path.into())
     }
 }
 
-fn cross_compiling() -> Result<Option<CrossPython>> {
+fn cross_compiling() -> Result<Option<CrossCompileConfig>> {
     if env::var("TARGET")? == env::var("HOST")? {
         return Ok(None);
     }
 
     if env::var("CARGO_CFG_TARGET_FAMILY")? == "windows" {
-        Ok(Some(CrossPython::both()?))
+        Ok(Some(CrossCompileConfig::both()?))
     } else {
-        Ok(Some(CrossPython::lib_only()?))
+        Ok(Some(CrossCompileConfig::lib_only()?))
     }
 }
 
@@ -197,26 +225,6 @@ fn parse_script_output(output: &str) -> HashMap<String, String> {
         .collect()
 }
 
-fn as_bool(config: &HashMap<String, String>, key: &str) -> Result<bool> {
-    match config
-        .get(key)
-        .map(|x| x.as_str())
-        .ok_or(format!("{} is not defined", key))?
-    {
-        "1" | "true" | "True" => Ok(true),
-        "0" | "false" | "False" => Ok(false),
-        _ => Err(format!("{} must be a bool (1/true/True or 0/false/False", key).into()),
-    }
-}
-
-fn as_numeric<T: FromStr>(config: &HashMap<String, String>, key: &str) -> Result<T> {
-    config
-        .get(key)
-        .ok_or(format!("{} is not defined", key))?
-        .parse::<T>()
-        .map_err(|_| format!("Could not parse value of {}", key).into())
-}
-
 /// Parse sysconfigdata file
 ///
 /// The sysconfigdata is basically a dictionary, and since we can't really use this library to read
@@ -257,7 +265,7 @@ fn ends_with(entry: &DirEntry, pat: &str) -> bool {
     name.to_string_lossy().ends_with(pat)
 }
 
-fn find_sysconfigdata(path: impl AsRef<Path>, cross: &CrossPython) -> Option<PathBuf> {
+fn find_sysconfigdata(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Option<PathBuf> {
     for f in fs::read_dir(path).expect("Path does not exist") {
         return match f {
             Ok(ref f) if starts_with(f, "_sysconfigdata") && ends_with(f, "py") => Some(f.path()),
@@ -291,20 +299,20 @@ fn find_sysconfigdata(path: impl AsRef<Path>, cross: &CrossPython) -> Option<Pat
 ///
 /// [1]: https://github.com/python/cpython/blob/3.8/Lib/sysconfig.py#L348
 fn load_cross_compile_from_sysconfigdata(
-    python_paths: CrossPython,
+    python_paths: CrossCompileConfig,
 ) -> Result<(InterpreterConfig, HashMap<String, String>)> {
     let sysconfig_path = find_sysconfigdata(&python_paths.lib_dir, &python_paths)
         .expect("_sysconfigdata*.py not found");
     let config_map = parse_sysconfigdata(sysconfig_path)?;
 
-    let shared = as_bool(&config_map, "Py_ENABLE_SHARED")?;
-    let major = as_numeric(&config_map, "version_major")?;
-    let minor = as_numeric(&config_map, "version_minor")?;
+    let shared = config_map.get_bool("Py_ENABLE_SHARED")?;
+    let major = config_map.get_numeric("version_major")?;
+    let minor = config_map.get_numeric("version_minor")?;
     let ld_version = match config_map.get("LDVERSION") {
         Some(s) => s.clone(),
         None => format!("{}.{}", major, minor),
     };
-    let calcsize_pointer = as_numeric(&config_map, "SIZEOF_VOID_P").ok();
+    let calcsize_pointer = config_map.get_numeric("SIZEOF_VOID_P").ok();
 
     let python_version = PythonVersion {
         major,
@@ -314,7 +322,7 @@ fn load_cross_compile_from_sysconfigdata(
 
     let interpreter_config = InterpreterConfig {
         version: python_version,
-        libdir: Some(python_paths.lib_dir),
+        libdir: python_paths.lib_dir.to_str().map(String::from),
         shared,
         ld_version,
         base_prefix: "".to_string(),
@@ -326,29 +334,14 @@ fn load_cross_compile_from_sysconfigdata(
 }
 
 fn load_cross_compile_from_headers(
-    python_paths: CrossPython,
+    python_paths: CrossCompileConfig,
 ) -> Result<(InterpreterConfig, HashMap<String, String>)> {
     let python_include_dir = python_paths.include_dir.unwrap();
     let python_include_dir = Path::new(&python_include_dir);
     let patchlevel_defines = parse_header_defines(python_include_dir.join("patchlevel.h"))?;
 
-    let major = match patchlevel_defines
-        .get("PY_MAJOR_VERSION")
-        .map(|major| major.parse::<u8>())
-    {
-        Some(Ok(major)) => major,
-        Some(Err(e)) => bail!("Failed to parse PY_MAJOR_VERSION: {}", e),
-        None => bail!("PY_MAJOR_VERSION undefined"),
-    };
-
-    let minor = match patchlevel_defines
-        .get("PY_MINOR_VERSION")
-        .map(|minor| minor.parse::<u8>())
-    {
-        Some(Ok(minor)) => minor,
-        Some(Err(e)) => bail!("Failed to parse PY_MINOR_VERSION: {}", e),
-        None => bail!("PY_MINOR_VERSION undefined"),
-    };
+    let major = patchlevel_defines.get_numeric("PY_MAJOR_VERSION")?;
+    let minor = patchlevel_defines.get_numeric("PY_MINOR_VERSION")?;
 
     let python_version = PythonVersion {
         major,
@@ -357,11 +350,11 @@ fn load_cross_compile_from_headers(
     };
 
     let config_map = parse_header_defines(python_include_dir.join("pyconfig.h"))?;
-    let shared = as_bool(&config_map, "Py_ENABLE_SHARED")?;
+    let shared = config_map.get_bool("Py_ENABLE_SHARED")?;
 
     let interpreter_config = InterpreterConfig {
         version: python_version,
-        libdir: Some(python_paths.lib_dir),
+        libdir: python_paths.lib_dir.to_str().map(String::from),
         shared,
         ld_version: format!("{}.{}", major, minor),
         base_prefix: "".to_string(),
@@ -372,9 +365,8 @@ fn load_cross_compile_from_headers(
     Ok((interpreter_config, fix_config_map(config_map)))
 }
 
-#[allow(unused_variables)]
 fn load_cross_compile_info(
-    python_paths: CrossPython,
+    python_paths: CrossCompileConfig,
 ) -> Result<(InterpreterConfig, HashMap<String, String>)> {
     let target_family = env::var("CARGO_CFG_TARGET_FAMILY")?;
     // Because compiling for windows on linux still includes the unix target family
--- a/build.rs
+++ b/build.rs
@@ -126,7 +126,7 @@ impl CrossCompileConfig {
             include_dir: None,
             os: env::var("CARGO_CFG_TARGET_OS").unwrap(),
             arch: env::var("CARGO_CFG_TARGET_ARCH").unwrap(),
-            version: env::var_os("PYO3_PYTHON_VERSION").map(|s| s.into_string().unwrap()),
+            version: env::var_os("PYO3_CROSS_PYTHON_VERSION").map(|s| s.into_string().unwrap()),
         })
     }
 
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -35,6 +35,10 @@ assert_approx_eq = "1.1.0"
 trybuild = "1.0.23"
 rustversion = "1.0"
 
+[build-dependencies]
+walkdir = "~2.3"
+regex = "~1.3"
```
davidhewitt, line 40: I think let's make these dependencies optional with a `cross-compile` feature please 👍

CHANGELOG.md
```diff
@@ -35,6 +36,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Don't rely on the order of structmembers to compute offsets in PyCell. Related to
   [#1058](https://github.com/PyO3/pyo3/pull/1058). [#1059](https://github.com/PyO3/pyo3/pull/1059)
 - Allows `&Self` as a `#[pymethods]` argument again. [#1071](https://github.com/PyO3/pyo3/pull/1071)
+- Linking against libpython when compiling for android even with `extension-module` set[#1082](https://github.com/PyO3/pyo3/issues/1082)
```
davidhewitt, line 39:
```suggestion
- Link against libpython on android with `extension-module` set. [#1082](https://github.com/PyO3/pyo3/issues/1082)
```

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Change `PyIterator::from_object` to return `PyResult<PyIterator>` instead of `Result<PyIterator, PyDowncastError>`. [#1051](https://github.com/PyO3/pyo3/pull/1051)
 - Implement `Send + Sync` for `PyErr`. `PyErr::new`, `PyErr::from_type`, `PyException::py_err` and `PyException::into` have had these bounds added to their arguments. [#1067](https://github.com/PyO3/pyo3/pull/1067)
 - Change `#[pyproto]` to return NotImplemented for operators for which Python can try a reversed operation. [1072](https://github.com/PyO3/pyo3/pull/1072)
+- Change method for getting cross-compilation configurations using the file which provides the `sysconfig` module with its data, `_sysconfig_*.py` located in the lib directory. [1077](https://github.com/PyO3/pyo3/issues/1077)
```
davidhewitt, line 26: I'd prefer skip the implementation detail

```suggestion
- Fix python configuration detection when cross-compiling. [1077](https://github.com/PyO3/pyo3/issues/1077)
```

build.rs
```diff
@@ -134,10 +134,112 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
 }
 
-fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, String>)> {
-    let python_include_dir = env::var("PYO3_CROSS_INCLUDE_DIR")?;
-    let python_include_dir = Path::new(&python_include_dir);
+fn parse_sysconfigdata(config_path: impl AsRef<Path>) -> Result<HashMap<String, String>> {
+    let config_reader = BufReader::new(File::open(config_path)?);
+    let mut entries = HashMap::new();
+    let entry_re = regex::Regex::new(r#"'([a-zA-Z_0-9]*)': ((?:"|')([\S ]*)(?:"|')|\d+)($|,$)"#)?;
+    let subsequent_re = regex::Regex::new(r#"\s+(?:"|')([\S ]*)(?:"|')($|,)"#)?;
```
davidhewitt, line 141: Might be worth adding a comment just above with examples of what these are matching? They're quite complicated expressions!

build.rs
```diff
@@ -188,6 +290,19 @@ fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, Strin
...
+fn load_cross_compile_info(
+    python_include_dir: String,
+    python_lib_dir: String,
+) -> Result<(InterpreterConfig, HashMap<String, String>)> {
+    // Try to configure from the sysconfigdata file which is more accurate for the information
+    // provided at python's compile time
+    match load_cross_compile_from_sysconfigdata(&python_lib_dir) {
+        Ok(ret) => Ok(ret),
+        // If the config could not be loaded by sysconfigdata, failover to configuring from headers
+        Err(_) => load_cross_compile_from_headers(&python_include_dir, &python_lib_dir),
```
davidhewitt, line 302: Seems reasonable. Though I wonder if we thought the header method was broken anyway, maybe we should just remove it to simplify the build script.

build.rs
```diff
@@ -574,10 +689,27 @@ fn main() -> Result<()> {
     // If you have troubles with your shell accepting '.' in a var name,
     // try using 'env' (sorry but this isn't our fault - it just has to
     // match the pkg-config package name, which is going to have a . in it).
-    let cross_compiling =
-        env::var("PYO3_CROSS_INCLUDE_DIR").is_ok() && env::var("PYO3_CROSS_LIB_DIR").is_ok();
+    //
+    // Detecting if cross-compiling by checking if the target triple is different from the host
+    // rustc's triple.
+    let cross_compiling = env::var("TARGET") != env::var("HOST");
```
davidhewitt, line 695: It's nice to have this check, but the condition might need to be different. 

E.g. this looks like it'll break 32-bit builds on windows (where the host rustc is usually 64-bit).

build.rs
```diff
@@ -574,10 +689,27 @@ fn main() -> Result<()> {
...
     let (interpreter_config, mut config_map) = if cross_compiling {
-        load_cross_compile_info()?
+        // If cross compiling we need the path to the cross-compiled include dir and lib dir, else
+        // fail quickly and loudly
+        let python_include_dir = match env::var("PYO3_CROSS_INCLUDE_DIR") {
+            Ok(v) => v,
+            Err(_) => bail!(
+                "Must provide PYO3_CROSS_INCLUDE_DIR environment variable when cross-compiling"
+            ),
+        };
```
davidhewitt, line 704: Maybe if we drop the header parsing then we can also drop this env var? Which might be nice.

Progdrasil:
> It looks like it was simpler than I feared, which is pleasing to see :)

Its not often you hear these words when it comes to compilation and dynamic linking

As for the porting strategy, thats pretty much what I was planing on doing. I'll probably only have time to put into it next weekend.

build.rs
Progdrasil: I did not know that detail! i'll look into what needs to be done here in such a case. Maybe looking at how other popular crates detect cross compilation.

build.rs
Progdrasil: While I havn't delved into the actual library code, don't you need the include directory for the FFI functions? If not then removing this variable would in fact improve cross compilation ergonomics

build.rs
Progdrasil: I'm not entirely sure. I think we can be fairly confident that the `sysconfigdata`  file needs to be present for the module to work properly. However, I'd like to confirm through documentation and if this is true for all versions that pyo3 support before making this removal.

build.rs
Progdrasil: Are you refering to when building a 32bit windows binary with a 64bit windows machine? if so the target should be different as according to [rust platform support] windows 32 bit target start with `i686` where as a 64bit windows should have `x86_64`.

Unfortunately I dont have access to windows with a development environment to test this out... However, by inspecting the [openssl-sys] crates [build] seems to basically boil down to this same comparison

[openssl-sys]: https://docs.rs/openssl-sys/0.9.58/openssl_sys/
[build]: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/find_normal.rs#L24
[rust platform support]: https://forge.rust-lang.org/release/platform-support.html

build.rs
Progdrasil: As I feared, after looking at the python source for sysconfig, [non-posix] os' don't load a `_sysconfigdata_` file, so I don't think it is possible to fully drop the header file parsing if windows is to be supported as a crosscompilation target. However, what I could do is do some conditional compilation here so all unix platforms use the `_sysconfigdata_` file and windows targets stay with the old header parsing method.

[non-posix]: https://github.com/python/cpython/blob/3.8/Lib/sysconfig.py#L426

Cargo.toml
Progdrasil: I've applied the feature. However I believe this will be useful when targeting one's own system (on unix) instead of running the python script. Although not all the information is present.

build.rs
Progdrasil: I've dropped it for unix platforms, unfortunately, windows won't have this information

Progdrasil:
@davidhewitt  What is the need for the `cfg_line_for_var` from the config map?

<details>
<summary>because the new config map generated by the `_sysconfigdata_` is huge</summary>

```sh
rustc --crate-name pyo3 --edition=2018 src/lib.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type lib --emit=dep-info,metadata,link -Cembed-bitcode=no -C debuginfo=2 --cfg 'feature="cross-compile"' --cfg 'feature="ctor"' --cfg 'feature="default"' --cfg 'feature="indoc"' --cfg 'feature="inventory"' --cfg 'feature="macros"' --cfg 'feature="paste"' --cfg 'feature="pyo3cls"' --cfg 'feature="regex"' --cfg 'feature="unindent"' --cfg 'feature="walkdir"' -C metadata=4f5db729a86d41e0 -C extra-filename=-4f5db729a86d41e0 --out-dir /home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/deps --target arm-linux-androideabi -C incremental=/home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/incremental -L dependency=/home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/deps -L dependency=/home/rene/git/perso/pyo3/target/debug/deps --extern ctor=/home/rene/git/perso/pyo3/target/debug/deps/libctor-d23edd316d8286d6.so --extern indoc=/home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/deps/libindoc-4782a3271805adbd.rmeta --extern inventory=/home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/deps/libinventory-e88ee21875d02d87.rmeta --extern libc=/home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/deps/liblibc-e4daaed138e3377f.rmeta --extern parking_lot=/home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/deps/libparking_lot-f8127f7c92732ae6.rmeta --extern paste=/home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/deps/libpaste-fde5f3099432a925.rmeta --extern pyo3cls=/home/rene/git/perso/pyo3/target/debug/deps/libpyo3cls-fc433adb7e94fbec.so --extern unindent=/home/rene/git/perso/pyo3/target/arm-linux-androideabi/debug/deps/libunindent-3007c61f26869942.rmeta -L native=/home/rene/git/perso/python3.7_armv7/android-build --cfg Py_3_5 --cfg Py_3_6 --cfg Py_3_7 --cfg Py_3 --cfg 'py_sys_config="CC"' --cfg 'py_sys_config="INSTSONAME"' --cfg 'py_sys_config="RETSIGTYPE"' --cfg 'py_sys_config="LIBSUBDIRS"' --cfg 'py_sys_config="HAVE_DIRFD"' --cfg 'py_sys_config="HAVE_GETPEERNAME"' --cfg 'py_sys_config="HAVE_SIGNAL_H"' --cfg 'py_sys_config="POBJS"' --cfg 'py_sys_config="HAVE_SPAWN_H"' --cfg 'py_sys_config="HAVE_CLOCK_GETTIME"' --cfg 'py_sys_config="HAVE_SETRESGID"' --cfg 'py_sys_config="SCRIPTDIR"' --cfg 'py_sys_config="DLINCLDIR"' --cfg 'py_sys_config="HAVE_ROUND"' --cfg 'py_sys_config="HAVE_TRUNCATE"' --cfg 'py_sys_config="HAVE_SYS_RESOURCE_H"' --cfg 'py_sys_config="HAVE_DECL_RTLD_NOLOAD"' --cfg 'py_sys_config="TESTPYTHONOPTS"' --cfg 'py_sys_config="VERSION"' --cfg 'py_sys_config="HAVE_FCHOWN"' --cfg 'py_sys_config="BASEMODLIBS"' --cfg 'py_sys_config="HAVE_DLOPEN"' --cfg 'py_sys_config="HAVE_UNLINKAT"' --cfg 'py_sys_config="HAVE_FSTATAT"' --cfg 'py_sys_config="CONFIG_ARGS"' --cfg 'py_sys_config="HAVE_MKFIFO"' --cfg 'py_sys_config="INSTALL_DATA"' --cfg 'py_sys_config="HAVE_ACCEPT4"' --cfg 'py_sys_config="HAVE_GETPAGESIZE"' --cfg 'py_sys_config="LIBS"' --cfg 'py_sys_config="UPDATE_FILE"' --cfg 'py_sys_config="INCLDIRSTOMAKE"' --cfg 'py_sys_config="abs_srcdir"' --cfg 'py_sys_config="SIZEOF__BOOL"' --cfg 'py_sys_config="ABIFLAGS"' --cfg 'py_sys_config="HAVE_EPOLL"' --cfg 'py_sys_config="DTRACE_OBJS"' --cfg 'py_sys_config="TESTTIMEOUT"' --cfg 'py_sys_config="HAVE_SETEUID"' --cfg 'py_sys_config="HAVE_UNSETENV"' --cfg 'py_sys_config="HAVE_LGAMMA"' --cfg 'py_sys_config="TIME_WITH_SYS_TIME"' --cfg 'py_sys_config="GITBRANCH"' --cfg 'py_sys_config="BLDSHARED"' --cfg 'py_sys_config="HAVE_FORKPTY"' --cfg 'py_sys_config="READELF"' --cfg 'py_sys_config="HAVE_FSTATVFS"' --cfg 'py_sys_config="HAVE_FCHMODAT"' --cfg 'py_sys_config="CONFIGURE_CPPFLAGS"' --cfg 'py_sys_config="HAVE_FDATASYNC"' --cfg 'py_sys_config="SIZEOF_UINTPTR_T"' --cfg 'py_sys_config="TESTRUNNER"' --cfg 'py_sys_config="HAVE_WMEMCMP"' --cfg 'py_sys_config="HAVE_SYS_SELECT_H"' --cfg 'py_sys_config="HAVE_FUTIMES"' --cfg 'py_sys_config="COREPYTHONPATH"' --cfg 'py_sys_config="HAVE_LIBDL"' --cfg 'py_sys_config="HAVE_MEMRCHR"' --cfg 'py_sys_config="CFLAGS_NODIST"' --cfg 'py_sys_config="HAVE_DECL_ISINF"' --cfg 'py_sys_config="HAVE_STRLCPY"' --cfg 'py_sys_config="HAVE_WAITPID"' --cfg 'py_sys_config="SHELL"' --cfg 'py_sys_config="HAVE_READLINK"' --cfg 'py_sys_config="STRIPFLAG"' --cfg 'py_sys_config="HAVE_LINUX_CAN_H"' --cfg 'py_sys_config="HAVE_SETRESUID"' --cfg 'py_sys_config="HAVE_PIPE2"' --cfg 'py_sys_config="HAVE_ALIGNED_REQUIRED"' --cfg 'py_sys_config="HAVE_SIGINFO_T_SI_BAND"' --cfg 'py_sys_config="HAVE_GETADDRINFO"' --cfg 'py_sys_config="HAVE_SYS_RANDOM_H"' --cfg 'py_sys_config="HAVE_SYS_UIO_H"' --cfg 'py_sys_config="HAVE_SYS_POLL_H"' --cfg 'py_sys_config="COVERAGE_REPORT"' --cfg 'py_sys_config="HAVE_MKFIFOAT"' --cfg 'py_sys_config="SIZEOF_INT"' --cfg 'py_sys_config="VPATH"' --cfg 'py_sys_config="HAVE_WAIT4"' --cfg 'py_sys_config="HAVE_ACOSH"' --cfg 'py_sys_config="HAVE_DECL_RTLD_NOW"' --cfg 'py_sys_config="BUILDEXE"' --cfg 'py_sys_config="HAVE_ERRNO_H"' --cfg 'py_sys_config="HAVE_DECL_RTLD_GLOBAL"' --cfg 'py_sys_config="AR"' --cfg 'py_sys_config="MULTIARCH_CPPFLAGS"' --cfg 'py_sys_config="BUILD_GNU_TYPE"' --cfg 'py_sys_config="TCLTK_INCLUDES"' --cfg 'py_sys_config="HAVE_TERMIOS_H"' --cfg 'py_sys_config="HAVE_CLOCK_SETTIME"' --cfg 'py_sys_config="HAVE_FDOPENDIR"' --cfg 'py_sys_config="HAVE_STRINGS_H"' --cfg 'py_sys_config="CONFIGURE_LDFLAGS"' --cfg 'py_sys_config="HAVE_LANGINFO_H"' --cfg 'py_sys_config="HAVE_MKNODAT"' --cfg 'py_sys_config="HAVE_DECL_RTLD_NODELETE"' --cfg 'py_sys_config="HAVE_X509_VERIFY_PARAM_SET1_HOST"' --cfg 'py_sys_config="HAVE_INTTYPES_H"' --cfg 'py_sys_config="HAVE_PREAD"' --cfg 'py_sys_config="MACHDESTLIB"' --cfg 'py_sys_config="HAVE_ERF"' --cfg 'py_sys_config="HAVE_SYMLINK"' --cfg 'py_sys_config="HAVE_SIGACTION"' --cfg 'py_sys_config="PY3LIBRARY"' --cfg 'py_sys_config="HAVE_PROTOTYPES"' --cfg 'py_sys_config="HAVE_SETPGRP"' --cfg 'py_sys_config="HAVE_FACCESSAT"' --cfg 'py_sys_config="HAVE_GETPGID"' --cfg 'py_sys_config="LLVM_PROF_ERR"' --cfg 'py_sys_config="HAVE_DYNAMIC_LOADING"' --cfg 'py_sys_config="CCSHARED"' --cfg 'py_sys_config="PY_CFLAGS_NODIST"' --cfg 'py_sys_config="HAVE_PTHREAD_GETCPUCLOCKID"' --cfg 'py_sys_config="HAVE_GETRANDOM_SYSCALL"' --cfg 'py_sys_config="HAVE_DEV_PTMX"' --cfg 'py_sys_config="HAVE_LONG_DOUBLE"' --cfg 'py_sys_config="SIZEOF_VOID_P"' --cfg 'py_sys_config="HAVE_TM_ZONE"' --cfg 'py_sys_config="DTRACE_HEADERS"' --cfg 'py_sys_config="HAVE_SYS_FILE_H"' --cfg 'py_sys_config="MACHDEP_OBJS"' --cfg 'py_sys_config="CFLAGS"' --cfg 'py_sys_config="WITH_PYMALLOC"' --cfg 'py_sys_config="HAVE_SIGALTSTACK"' --cfg 'py_sys_config="HAVE_LINUX_CAN_BCM_H"' --cfg 'py_sys_config="HAVE_OPENPTY"' --cfg 'py_sys_config="HAVE_EPOLL_CREATE1"' --cfg 'py_sys_config="OPENSSL_LIBS"' --cfg 'py_sys_config="HAVE_CLOCK_GETRES"' --cfg 'py_sys_config="HAVE_GETRESUID"' --cfg 'py_sys_config="PURIFY"' --cfg 'py_sys_config="HAVE_DLFCN_H"' --cfg 'py_sys_config="MODULE_OBJS"' --cfg 'py_sys_config="ENSUREPIP"' --cfg 'py_sys_config="HAVE_FCHOWNAT"' --cfg 'py_sys_config="HAVE_FCNTL_H"' --cfg 'py_sys_config="HAVE_STRUCT_STAT_ST_BLKSIZE"' --cfg 'py_sys_config="PYTHONFRAMEWORK"' --cfg 'py_sys_config="SIZEOF_OFF_T"' --cfg 'py_sys_config="SIZEOF_TIME_T"' --cfg 'py_sys_config="HAVE_SEM_TIMEDWAIT"' --cfg 'py_sys_config="SHLIBS"' --cfg 'py_sys_config="HAVE_SEM_GETVALUE"' --cfg 'py_sys_config="HAVE_SNPRINTF"' --cfg 'py_sys_config="PY_FORMAT_SIZE_T"' --cfg 'py_sys_config="HAVE_TEMPNAM"' --cfg 'py_sys_config="HAVE_WAITID"' --cfg 'py_sys_config="SIZEOF_PID_T"' --cfg 'py_sys_config="SIZEOF_SHORT"' --cfg 'py_sys_config="RUNSHARED"' --cfg 'py_sys_config="HAVE_BUILTIN_ATOMIC"' --cfg 'py_sys_config="BYTESTR_DEPS"' --cfg 'py_sys_config="CONFINCLUDEDIR"' --cfg 'py_sys_config="HAVE_BROKEN_SEM_GETVALUE"' --cfg 'py_sys_config="HAVE_SYS_STAT_H"' --cfg 'py_sys_config="HAVE_LINKAT"' --cfg 'py_sys_config="DESTLIB"' --cfg 'py_sys_config="HAVE_LUTIMES"' --cfg 'py_sys_config="HAVE_PTHREAD_SIGMASK"' --cfg 'py_sys_config="HAVE_READLINKAT"' --cfg 'py_sys_config="IO_H"' --cfg 'py_sys_config="DESTSHARED"' --cfg 'py_sys_config="HAVE_GETHOSTBYNAME_R_6_ARG"' --cfg 'py_sys_config="HAVE_POSIX_FADVISE"' --cfg 'py_sys_config="HAVE_SIGRELSE"' --cfg 'py_sys_config="HAVE_TIMES"' --cfg 'py_sys_config="BUILDPYTHON"' --cfg 'py_sys_config="SUBDIRS"' --cfg 'py_sys_config="HAVE_POSIX_FALLOCATE"' --cfg 'py_sys_config="HAVE_SYS_XATTR_H"' --cfg 'py_sys_config="HAVE_DEVICE_MACROS"' --cfg 'py_sys_config="MANDIR"' --cfg 'py_sys_config="SYS_SELECT_WITH_SYS_TIME"' --cfg 'py_sys_config="HAVE_STRING_H"' --cfg 'py_sys_config="BINDIR"' --cfg 'py_sys_config="HAVE_LOG2"' --cfg 'py_sys_config="OPENSSL_LDFLAGS"' --cfg 'py_sys_config="HAVE_CHOWN"' --cfg 'py_sys_config="HAVE_GETHOSTBYNAME_R"' --cfg 'py_sys_config="HAVE_SYS_SENDFILE_H"' --cfg 'py_sys_config="WITH_THREAD"' --cfg 'py_sys_config="LLVM_PROF_MERGER"' --cfg 'py_sys_config="HAVE_MBRTOWC"' --cfg 'py_sys_config="HAVE_MREMAP"' --cfg 'py_sys_config="MODBUILT_NAMES"' --cfg 'py_sys_config="HAVE_ASINH"' --cfg 'py_sys_config="PY_LDFLAGS_NODIST"' --cfg 'py_sys_config="SRCDIRS"' --cfg 'py_sys_config="HAVE_STDLIB_H"' --cfg 'py_sys_config="HAVE_INET_ATON"' --cfg 'py_sys_config="HAVE_READV"' --cfg 'py_sys_config="STDC_HEADERS"' --cfg 'py_sys_config="HAVE_SETUID"' --cfg 'py_sys_config="HAVE_MKNOD"' --cfg 'py_sys_config="HAVE_LINUX_NETLINK_H"' --cfg 'py_sys_config="EXT_SUFFIX"' --cfg 'py_sys_config="HAVE_RENAMEAT"' --cfg 'py_sys_config="HAVE_TIMEGM"' --cfg 'py_sys_config="HAVE_UTIMES"' --cfg 'py_sys_config="PY_COERCE_C_LOCALE"' --cfg 'py_sys_config="PY_SSL_DEFAULT_CIPHERS"' --cfg 'py_sys_config="HAVE_GETGROUPLIST"' --cfg 'py_sys_config="HAVE_POLL"' --cfg 'py_sys_config="PTHREAD_KEY_T_IS_COMPATIBLE_WITH_INT"' --cfg 'py_sys_config="LDLIBRARY"' --cfg 'py_sys_config="HAVE_TGAMMA"' --cfg 'py_sys_config="PY_STDMODULE_CFLAGS"' --cfg 'py_sys_config="HAVE_WCSCOLL"' --cfg 'py_sys_config="SIZEOF_DOUBLE"' --cfg 'py_sys_config="DFLAGS"' --cfg 'py_sys_config="HAVE_LINUX_CAN_RAW_H"' --cfg 'py_sys_config="HAVE_SEM_UNLINK"' --cfg 'py_sys_config="SYSLIBS"' --cfg 'py_sys_config="HAVE_STATVFS"' --cfg 'py_sys_config="HAVE_GETC_UNLOCKED"' --cfg 'py_sys_config="OPENSSL_INCLUDES"' --cfg 'py_sys_config="HAVE_SYS_SYSCALL_H"' --cfg 'py_sys_config="PYTHON_FOR_BUILD"' --cfg 'py_sys_config="HAVE_SETLOCALE"' --cfg 'py_sys_config="LIBRARY_OBJS"' --cfg 'py_sys_config="HAVE_FORK"' --cfg 'py_sys_config="EXTRA_CFLAGS"' --cfg 'py_sys_config="prefix"' --cfg 'py_sys_config="HAVE_GAI_STRERROR"' --cfg 'py_sys_config="PY_LDFLAGS"' --cfg 'py_sys_config="HAVE_PUTENV"' --cfg 'py_sys_config="HAVE_SETVBUF"' --cfg 'py_sys_config="HAVE_FSEEKO"' --cfg 'py_sys_config="HAVE_NET_IF_H"' --cfg 'py_sys_config="HAVE_TMPFILE"' --cfg 'py_sys_config="LDFLAGS_NODIST"' --cfg 'py_sys_config="DISTFILES"' --cfg 'py_sys_config="HAVE_GETPWENT"' --cfg 'py_sys_config="LDLIBRARYDIR"' --cfg 'py_sys_config="HAVE_SYS_UTSNAME_H"' --cfg 'py_sys_config="datarootdir"' --cfg 'py_sys_config="HAVE_HSTRERROR"' --cfg 'py_sys_config="HAVE_ATANH"' --cfg 'py_sys_config="LIBM"' --cfg 'py_sys_config="SIZEOF_FPOS_T"' --cfg 'py_sys_config="HAVE_SETITIMER"' --cfg 'py_sys_config="HAVE_STD_ATOMIC"' --cfg 'py_sys_config="HAVE_HTOLE64"' --cfg 'py_sys_config="CONFINCLUDEPY"' --cfg 'py_sys_config="HAVE_SETHOSTNAME"' --cfg 'py_sys_config="HAVE_LARGEFILE_SUPPORT"' --cfg 'py_sys_config="IO_OBJS"' --cfg 'py_sys_config="RESSRCDIR"' --cfg 'py_sys_config="LDSHARED"' --cfg 'py_sys_config="PY_CORE_LDFLAGS"' --cfg 'py_sys_config="UNIVERSALSDK"' --cfg 'py_sys_config="MODDISABLED_NAMES"' --cfg 'py_sys_config="CONFIGURE_CFLAGS"' --cfg 'py_sys_config="HAVE_HYPOT"' --cfg 'py_sys_config="OTHER_LIBTOOL_OPT"' --cfg 'py_sys_config="LIBRARY_OBJS_OMIT_FROZEN"' --cfg 'py_sys_config="WITH_DOC_STRINGS"' --cfg 'py_sys_config="LIBRARY"' --cfg 'py_sys_config="HAVE_ZLIB_COPY"' --cfg 'py_sys_config="HAVE_PTHREAD_H"' --cfg 'py_sys_config="HAVE_MMAP"' --cfg 'py_sys_config="INCLUDEDIR"' --cfg 'py_sys_config="HAVE_PATHCONF"' --cfg 'py_sys_config="PY_BUILTIN_MODULE_CFLAGS"' --cfg 'py_sys_config="LIBDIR"' --cfg 'py_sys_config="CONFIGFILES"' --cfg 'py_sys_config="DESTPATH"' --cfg 'py_sys_config="DTRACE_DEPS"' --cfg 'py_sys_config="HAVE_DECL_RTLD_LOCAL"' --cfg 'py_sys_config="SIZEOF_FLOAT"' --cfg 'py_sys_config="PYTHON_OBJS"' --cfg 'py_sys_config="INSTALL"' --cfg 'py_sys_config="GITVERSION"' --cfg 'py_sys_config="INSTALL_SHARED"' --cfg 'py_sys_config="HAVE_GETPGRP"' --cfg 'py_sys_config="COVERAGE_INFO"' --cfg 'py_sys_config="LINKFORSHARED"' --cfg 'py_sys_config="HAVE_SYS_EPOLL_H"' --cfg 'py_sys_config="MODLIBS"' --cfg 'py_sys_config="PGENOBJS"' --cfg 'py_sys_config="HAVE_SETREGID"' --cfg 'py_sys_config="HAVE_FUTIMESAT"' --cfg 'py_sys_config="HAVE_KILLPG"' --cfg 'py_sys_config="CONFIGURE_LDFLAGS_NODIST"' --cfg 'py_sys_config="HAVE_CLOCK"' --cfg 'py_sys_config="LOCALMODLIBS"' --cfg 'py_sys_config="HAVE_WRITEV"' --cfg 'py_sys_config="HAVE_IF_NAMEINDEX"' --cfg 'py_sys_config="HAVE_SYS_TIME_H"' --cfg 'py_sys_config="SOABI"' --cfg 'py_sys_config="LDVERSION"' --cfg 'py_sys_config="PGOBJS"' --cfg 'py_sys_config="HAVE_COPYSIGN"' --cfg 'py_sys_config="HAVE_MEMORY_H"' --cfg 'py_sys_config="SHLIB_SUFFIX"' --cfg 'py_sys_config="HAVE_SCHED_SETPARAM"' --cfg 'py_sys_config="LIBOBJS"' --cfg 'py_sys_config="SIZEOF_SIZE_T"' --cfg 'py_sys_config="HAVE_LSTAT"' --cfg 'py_sys_config="SGI_ABI"' --cfg 'py_sys_config="HAVE_SYS_SOCKET_H"' --cfg 'py_sys_config="HAVE_REALPATH"' --cfg 'py_sys_config="HAVE_LINUX_VM_SOCKETS_H"' --cfg 'py_sys_config="PY_CORE_CFLAGS"' --cfg 'py_sys_config="HAVE_CHROOT"' --cfg 'py_sys_config="HAVE_STDARG_PROTOTYPES"' --cfg 'py_sys_config="HAVE_KILL"' --cfg 'py_sys_config="HAVE_SYS_UN_H"' --cfg 'py_sys_config="LIBPC"' --cfg 'py_sys_config="HOST_GNU_TYPE"' --cfg 'py_sys_config="HAVE_SSIZE_T"' --cfg 'py_sys_config="PYTHONFRAMEWORKINSTALLDIR"' --cfg 'py_sys_config="HAVE_FTELLO"' --cfg 'py_sys_config="HAVE_SYS_PARAM_H"' --cfg 'py_sys_config="SITEPATH"' --cfg 'py_sys_config="HAVE_PTHREAD_KILL"' --cfg 'py_sys_config="ANDROID_API_LEVEL"' --cfg 'py_sys_config="HAVE_ALLOCA_H"' --cfg 'py_sys_config="HAVE_SCHED_SETSCHEDULER"' --cfg 'py_sys_config="HAVE_SYS_TIMES_H"' --cfg 'py_sys_config="DYNLOADFILE"' --cfg 'py_sys_config="HAVE_SYSCONF"' --cfg 'py_sys_config="HAVE_FLOCK"' --cfg 'py_sys_config="QUICKTESTOPTS"' --cfg 'py_sys_config="HAVE_SETPRIORITY"' --cfg 'py_sys_config="LDLAST"' --cfg 'py_sys_config="MODOBJS"' --cfg 'py_sys_config="MACHDEP"' --cfg 'py_sys_config="PROFILE_TASK"' --cfg 'py_sys_config="HAVE_DECL_ISNAN"' --cfg 'py_sys_config="MAKESETUP"' --cfg 'py_sys_config="HAVE_DUP2"' --cfg 'py_sys_config="HAVE_TMPNAM"' --cfg 'py_sys_config="STRICT_SYSV_CURSES"' --cfg 'py_sys_config="LIBFFI_INCLUDEDIR"' --cfg 'py_sys_config="LIBOBJDIR"' --cfg 'py_sys_config="srcdir"' --cfg 'py_sys_config="HAVE_SCHED_GET_PRIORITY_MAX"' --cfg 'py_sys_config="HAVE_CTERMID"' --cfg 'py_sys_config="HAVE_DIRENT_H"' --cfg 'py_sys_config="HAVE_GETITIMER"' --cfg 'py_sys_config="PYTHONPATH"' --cfg 'py_sys_config="SIZEOF_PTHREAD_T"' --cfg 'py_sys_config="TCLTK_LIBS"' --cfg 'py_sys_config="HAVE_DUP3"' --cfg 'py_sys_config="HAVE_PWRITEV"' --cfg 'py_sys_config="HAVE_SIGTIMEDWAIT"' --cfg 'py_sys_config="HAVE_LOCKF"' --cfg 'py_sys_config="HAVE_WCHAR_H"' --cfg 'py_sys_config="PY_CFLAGS"' --cfg 'py_sys_config="FILEMODE"' --cfg 'py_sys_config="SIZEOF_PTHREAD_KEY_T"' --cfg 'py_sys_config="DTRACE"' --cfg 'py_sys_config="HAVE_GETPID"' --cfg 'py_sys_config="HAVE_FCHDIR"' --cfg 'py_sys_config="XMLLIBSUBDIRS"' --cfg 'py_sys_config="LIBPL"' --cfg 'py_sys_config="GITTAG"' --cfg 'py_sys_config="HAVE_STAT_TV_NSEC"' --cfg 'py_sys_config="LIBDEST"' --cfg 'py_sys_config="HAVE_SOCKADDR_ALG"' --cfg 'py_sys_config="HAVE_GETTIMEOFDAY"' --cfg 'py_sys_config="HAVE_DIRENT_D_TYPE"' --cfg 'py_sys_config="SUBDIRSTOO"' --cfg 'py_sys_config="LLVM_PROF_FILE"' --cfg 'py_sys_config="HAVE_SYSEXITS_H"' --cfg 'py_sys_config="PARSER_OBJS"' --cfg 'py_sys_config="SIZEOF_LONG_DOUBLE"' --cfg 'py_sys_config="PYTHON_HEADERS"' --cfg 'py_sys_config="CONFIGURE_CFLAGS_NODIST"' --cfg 'py_sys_config="HAVE_ADDRINFO"' --cfg 'py_sys_config="HAVE_MKDIRAT"' --cfg 'py_sys_config="SIZEOF_LONG"' --cfg 'py_sys_config="HAVE_SCHED_RR_GET_INTERVAL"' --cfg 'py_sys_config="HAVE_STRUCT_STAT_ST_RDEV"' --cfg 'py_sys_config="HAVE_MAKEDEV"' --cfg 'py_sys_config="HAVE_GETPRIORITY"' --cfg 'py_sys_config="HAVE_GETGROUPS"' --cfg 'py_sys_config="HAVE_OPENAT"' --cfg 'py_sys_config="HAVE_FTRUNCATE"' --cfg 'py_sys_config="LIPO_32BIT_FLAGS"' --cfg 'py_sys_config="LIBC"' --cfg 'py_sys_config="HAVE_SETGID"' --cfg 'py_sys_config="HAVE_SETPGID"' --cfg 'py_sys_config="INSTALL_PROGRAM"' --cfg 'py_sys_config="HAVE_INITGROUPS"' --cfg 'py_sys_config="SRC_GDB_HOOKS"' --cfg 'py_sys_config="HAVE_GETRESGID"' --cfg 'py_sys_config="DISTDIRS"' --cfg 'py_sys_config="HAVE_SCHED_H"' --cfg 'py_sys_config="HAVE_ERFC"' --cfg 'py_sys_config="Py_ENABLE_SHARED"' --cfg 'py_sys_config="BASECPPFLAGS"' --cfg 'py_sys_config="BINLIBDEST"' --cfg 'py_sys_config="DESTDIRS"' --cfg 'py_sys_config="HAVE_FCHMOD"' --cfg 'py_sys_config="DIRMODE"' --cfg 'py_sys_config="EXEMODE"' --cfg 'py_sys_config="EXE"' --cfg 'py_sys_config="HAVE_SIGINTERRUPT"' --cfg 'py_sys_config="MAJOR_IN_SYSMACROS"' --cfg 'py_sys_config="HAVE_SEM_OPEN"' --cfg 'py_sys_config="HAVE_FPATHCONF"' --cfg 'py_sys_config="LDFLAGS"' --cfg 'py_sys_config="BASECFLAGS"' --cfg 'py_sys_config="HAVE_LINUX_TIPC_H"' --cfg 'py_sys_config="HAVE_SENDFILE"' --cfg 'py_sys_config="HAVE_WCSFTIME"' --cfg 'py_sys_config="LDCXXSHARED"' --cfg 'py_sys_config="LN"' --cfg 'py_sys_config="HAVE_FSYNC"' --cfg 'py_sys_config="PYTHON"' --cfg 'py_sys_config="PYTHON_FOR_REGEN"' --cfg 'py_sys_config="HAVE_ENDIAN_H"' --cfg 'py_sys_config="INSTALL_SCRIPT"' --cfg 'py_sys_config="HAVE_POLL_H"' --cfg 'py_sys_config="HAVE_STDINT_H"' --cfg 'py_sys_config="HAVE_FINITE"' --cfg 'py_sys_config="PYTHONFRAMEWORKPREFIX"' --cfg 'py_sys_config="exec_prefix"' --cfg 'py_sys_config="HAVE_ALARM"' --cfg 'py_sys_config="HAVE_SYS_SYSMACROS_H"' --cfg 'py_sys_config="DLLLIBRARY"' --cfg 'py_sys_config="HAVE_STRUCT_TM_TM_ZONE"' --cfg 'py_sys_config="ENABLE_IPV6"' --cfg 'py_sys_config="HAVE_SETREUID"' --cfg 'py_sys_config="HAVE_LOG1P"' --cfg 'py_sys_config="CXX"' --cfg 'py_sys_config="HAVE_LCHOWN"' --cfg 'py_sys_config="HAVE_SYS_STATVFS_H"' --cfg 'py_sys_config="OPT"' --cfg 'py_sys_config="SIZEOF_LONG_LONG"' --cfg 'py_sys_config="HAVE_UTIMENSAT"' --cfg 'py_sys_config="HAVE_SYNC"' --cfg 'py_sys_config="MACOSX_DEPLOYMENT_TARGET"' --cfg 'py_sys_config="PY_CPPFLAGS"' --cfg 'py_sys_config="HAVE_GAMMA"' --cfg 'py_sys_config="COVERAGE_REPORT_OPTIONS"' --cfg 'py_sys_config="HAVE_UTIME_H"' --cfg 'py_sys_config="HAVE_GRP_H"' --cfg 'py_sys_config="LINKCC"' --cfg 'py_sys_config="HAVE_MKTIME"' --cfg 'py_sys_config="DOUBLE_IS_LITTLE_ENDIAN_IEEE754"' --cfg 'py_sys_config="HAVE_LINUX_RANDOM_H"' --cfg 'py_sys_config="HAVE_SETSID"' --cfg 'py_sys_config="HAVE_SYS_ENDIAN_H"' --cfg 'py_sys_config="HAVE_PWRITE"' --cfg 'py_sys_config="HAVE_EXPM1"' --cfg 'py_sys_config="HAVE_NICE"' --cfg 'py_sys_config="HAVE_GETLOGIN"' --cfg 'py_sys_config="HAVE_UNAME"' --cfg 'py_sys_config="HAVE_SIGWAITINFO"' --cfg 'py_sys_config="HAVE_SOCKETPAIR"' --cfg 'py_sys_config="HAVE_ASM_TYPES_H"' --cfg 'py_sys_config="EXTRATESTOPTS"' --cfg 'py_sys_config="HAVE_PTY_H"' --cfg 'py_sys_config="HAVE_STRDUP"' --cfg 'py_sys_config="HAVE_DECL_ISFINITE"' --cfg 'py_sys_config="CFLAGSFORSHARED"' --cfg 'py_sys_config="HAVE_SIGPENDING"' --cfg 'py_sys_config="INCLUDEPY"' --cfg 'py_sys_config="HAVE_SETGROUPS"' --cfg 'py_sys_config="HAVE_PREADV"' --cfg 'py_sys_config="HAVE_WCSXFRM"' --cfg 'py_sys_config="HAVE_SYS_TYPES_H"' --cfg 'py_sys_config="MAINCC"' --cfg 'py_sys_config="HAVE_SETEGID"' --cfg 'py_sys_config="GNULD"' --cfg 'py_sys_config="HAVE_SYS_WAIT_H"' --cfg 'py_sys_config="HAVE_INET_PTON"' --cfg 'py_sys_config="HAVE_STRUCT_PASSWD_PW_GECOS"' --cfg 'py_sys_config="PARSER_HEADERS"' --cfg 'py_sys_config="HAVE_SCHED_SETAFFINITY"' --cfg 'py_sys_config="PGO_PROF_GEN_FLAG"' --cfg 'py_sys_config="HAVE_SOCKADDR_STORAGE"' --cfg 'py_sys_config="HAVE_UNISTD_H"' --cfg 'py_sys_config="HAVE_STRFTIME"' --cfg 'py_sys_config="HAVE_PRLIMIT"' --cfg 'py_sys_config="abs_builddir"' --cfg 'py_sys_config="TESTOPTS"' --cfg 'py_sys_config="HAVE_FUTIMENS"' --cfg 'py_sys_config="HAVE_GETNAMEINFO"' --cfg 'py_sys_config="ARFLAGS"' --cfg 'py_sys_config="HAVE_SIGWAIT"' --cfg 'py_sys_config="PGO_PROF_USE_FLAG"' --cfg 'py_sys_config="CFLAGS_ALIASING"' --cfg 'py_sys_config="HAVE_EXECV"' --cfg 'py_sys_config="NO_AS_NEEDED"' --cfg 'py_sys_config="HAVE_SYMLINKAT"' --cfg 'py_sys_config="PYTHONFRAMEWORKDIR"' --cfg 'py_sys_config="HAVE_GETSID"' --cfg 'py_sys_config="HAVE_STRUCT_STAT_ST_BLOCKS"' --cfg 'py_sys_config="OBJECT_OBJS"' --cfg 'py_sys_config="PGEN"' --cfg 'py_sys_config="TESTPATH"' --cfg 'py_sys_config="HAVE_STRUCT_PASSWD_PW_PASSWD"' --cfg 'py_sys_config="HAVE_LINK"' --cfg 'py_sys_config="HAVE_PAUSE"' --cfg 'py_sys_config="DIST"' --cfg 'py_sys_config="SIZEOF_WCHAR_T"' --cfg 'py_sys_config="HAVE_TCGETPGRP"' --cfg 'py_sys_config="HAVE_SYS_IOCTL_H"' --cfg 'py_sys_config="MULTIARCH"' --cfg 'py_sys_config="HAVE_TCSETPGRP"' --cfg 'py_sys_config="CPPFLAGS"' --cfg 'py_sys_config="HAVE_NETPACKET_PACKET_H"' --cfg 'py_sys_config="HAVE_DECL_RTLD_LAZY"' --cfg 'py_sys_config="TESTPYTHON"' --cfg 'py_sys_config="HAVE_LINUX_CAN_RAW_FD_FRAMES"' --cfg 'py_sys_config="WITH_DECIMAL_CONTEXTVAR"' --cfg 'py_sys_config="BLDLIBRARY"' --cfg 'py_sys_config="UNICODE_DEPS"' --cfg 'py_sys_config="MKDIR_P"' -l python3.7m`
```

</details>

davidhewitt:
> What is the need for the cfg_line_for_var from the config map?
It's used to ensure that our ffi definitions match the structs generated by similar configuration on the C side, e.g. here:
https://github.com/PyO3/pyo3/blob/b71018689dead517ae7389f0618b4d9aaf4e57a4/src/ffi/object.rs#L12

However, from searching the codebase it looks like the only ones we need are `WITH_THREAD`, `Py_TRACE_REFS`, `COUNT_ALLOCS`, and `Py_REF_DEBUG`. https://github.com/PyO3/pyo3/search?q=py_sys_config&unscoped_q=py_sys_config

So I think we should probably change `build.rs` to check the config map just for those variables?

build.rs
```diff
@@ -134,10 +185,180 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
+/// 'ANDROID_API_LEVEL': 26,
+/// 'AR': 'arm-linux-androideabi-ar',
+/// 'ARFLAGS': 'rcs',
+/// 'BASECFLAGS': '-mfloat-abi=softfp -mfpu=vfpv3-d16 -Wno-unused-result '
+///               '-Wsign-compare -Wunreachable-code',
+/// 'BASECPPFLAGS': '-IObjects -IInclude -IPython'}
+/// ```
+#[cfg(feature = "cross-compile")]
+fn parse_sysconfigdata(config_path: impl AsRef<Path>) -> Result<HashMap<String, String>> {
+    let config_reader = BufReader::new(File::open(config_path)?);
```
kngwyu, line 234: How about making Python output only what we need?
E.g,
```rust
    let mut script = fs::read_to_string(config_path)?;
    s += r#"
print("version_major", build_time_vars["VERSION"][0])  # 3
print("version_minor", build_time_vars["VERSION"][2])  # E.g., 8
...
"#;
    let config = run_python_script(find_interpreter()?, script)?;
```
(Note that  I introduced `find_interpreter` in #1097)

Then we can use the same code as `get_config_from_interpreter`.

davidhewitt: It's possible when cross-compiling we may not have a Python interpreter for the host machine which can run the script? Though this approach is also quite nice.

build.rs
davidhewitt: Yep, what I mean though is that a 32-bit windows Python interpreter is still available for us to execute in the normal fashion, so we can get the config from the Python interpreter. Especially as we now know we can't use the `sysconfigdata` file on Windows, we definitely need to go down the "not cross compiling" configuration route.

I think if you rebase on master (where I've made a couple of CI fixes), the 32-bit windows builds will fail, which'll show what I mean 👍

build.rs
```diff
@@ -134,10 +185,180 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
+        let entry = match walker.next() {
+            Some(Ok(entry)) => entry,
+            None => bail!("Could not find sysconfigdata file"),
+            _ => continue,
+        };
+        let entry = entry.into_path();
+        if sysconfig_re.is_match(entry.to_str().unwrap()) {
+            break entry;
+        }
+    };
```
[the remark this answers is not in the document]
Progdrasil: The point is to go recursively within the directory. Let's say the lib directory is at `/lib/`, while the `libpython3.X.so` is present at the root, the `_sysconfigdata_` file might be at `/lib/python3.x/` as it is on my manjaro system. However my cross compiled library, the file is located at `/lib/build/lib.linux-arm-3.x/` Thats why I went for a micro crate which has one dependency (or 3 if on windows) because i didn't want to do this recursive work simply for the sake of not having a dependency.

build.rs
Progdrasil: i'll check it out. it'll help since i dont have access to a windows dev machine

Progdrasil:
> However, from searching the codebase it looks like the only ones we need are `WITH_THREAD`, `Py_TRACE_REFS`, `COUNT_ALLOCS`, and `Py_REF_DEBUG`. https://github.com/PyO3/pyo3/search?q=py_sys_config&unscoped_q=py_sys_config
> 
> So I think we should probably change `build.rs` to check the config map just for those variables?

I can fix that!

build.rs
Progdrasil: This would definitively be cleaner, with less complicated regex's. I would be curious to see which environments dont have a host python interpreter when cross compiling python. For example, the build system to compile python for android (buildozer) first compiles the same version of python for the host before cross-compiling python for the target.

Progdrasil:
Bad news is, none of these variables are exposed in the `_sysconfigdata_` file... I think that file will help to get the ABI flags and LDVERSION. but for the rest we will need to parse the header files. Good news is, this seems to be the same thing python itself does within the `sysconfig` module

davidhewitt:
> Bad news is, none of these variables are exposed in the _sysconfigdata_ file... I think that file will help to get the ABI flags and LDVERSION. but for the rest we will need to parse the header files. Good news is, this seems to be the same thing python itself does within the sysconfig module

Ah, blast, sorry I suggested a bad approach. 😢 Thanks for continuing to figure this out!

Progdrasil:
>  Ah, blast, sorry I suggested a bad approach. cry Thanks for continuing to figure this out!

No problem, at this point it's basically research and experimentation. Happens all the time.

I wont have much time to attribute to this before the end of this coming week though. I might read through python's documentation during lost time to maybe find a better solution than header parsing. `cbindgen` might be a good option to get the options out of the headers easily, not sure how that will impact build times though.

build.rs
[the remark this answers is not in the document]
Progdrasil: Although if we go with the python way to get the information out of the file i'll take the `starts_with` approach so we can remove regex as a build dependency.

kngwyu:
@Progdrasil 
How about calling `sysconfig.parse_config_h` in Python script?
It might be easier to maintain than having our regex.

davidhewitt:
> How about calling sysconfig.parse_config_h in Python script?

That seems like a good idea to me!

Wrt presence of working Python interpreter on host - I think you're right most host systems will have one. I also think we're ok to require one (as long as we document it)!

kngwyu:
> How about calling sysconfig.parse_config_h in Python script?

I tried
```python
In [5]: with open(sysconfig.get_config_h_filename()) as f: 
   ...:     d = sysconfig.parse_config_h(f) 
```
but `d` was empty.

So I think `config.h` does not work for this (maybe Makefile matters?) and think `_sysconfigdata_*` more reliable.

> Bad news is, none of these variables are exposed in the _sysconfigdata_ file... I think that file will help to get the ABI flags and LDVERSION. 

For Py_TRACE_REFS, COUNT_ALLOCS, and Py_REF_DEBUG: we should just set these variables if `Py_DEBUG` is 1.
And for `WITH_THREAD`, I have no way to detect this automatically.
However, since `WITH_THREAD=1` is common and always true in Python >= 3.7, it would be enough to set `WITH_THREAD=1` by default and make it customizable by env vars.

build.rs
```diff
@@ -134,10 +185,180 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
+
+    let interpreter_config = InterpreterConfig {
+        version: python_version,
+        libdir: Some(python_paths.lib_dir.to_owned()),
+        shared,
+        ld_version,
+        base_prefix: "".to_string(),
+        executable: PathBuf::new(),
+        calcsize_pointer: None,
```
kngwyu, line 351: It looks like we can use `SIZEOF_VOID_P` for this.

Progdrasil:
> For Py_TRACE_REFS, COUNT_ALLOCS, and Py_REF_DEBUG: we should just set these variables if Py_DEBUG is 1.
> And for WITH_THREAD, I have no way to detect this automatically.
> However, since WITH_THREAD=1 is common and always true in Python >= 3.7, it would be enough to set WITH_THREAD=1 by default and make it customizable by env vars.

So after more investigation, `_sysconfigdata` is [built from the `makefiles` and `pyconfig.h`][1] so this file should be complete.
If there are variables missing such as `Py_TRACE_REFS`, `COUNT_ALLOCS`, and `Py_REF_DEBUG` it is because they are not set.
`Py_DEBUG` is se to 0 in all my builds but I can set a fall back.
`WITH_THREAD` is present in this file for versions 3.5 and 3.6, since it is missing in more recent versions I will take for granted that it is set by default.

as for the placement of this file, it is [hard coded][2] when generated, so i can search for this placement explicitly and if not found search for the file at the top level. This might however be less accurate than simply searching recursively within the lib directory, let me know which you prefer @davidhewitt  and @kngwyu 
 
 [1]: https://github.com/python/cpython/blob/3.5/Lib/sysconfig.py#L342

kngwyu:
> as for the placement of this file, it is hard coded when generated, so i can search for this placement explicitly and if not found search for the file at the top level. This might however be less accurate than simply searching recursively within the lib directory, let me know which you prefer @davidhewitt and @kngwyu

I'm sorry I don't understand the context completely, but please choose a simpler way for finding the file.
If it cannot cover all cases, let's make it customizable by some environmental variable.

Progdrasil:
Python saves the file during build time at the path defined by:
```py
pybuilddir = 'build/lib.%s-%s' % (get_platform(), sys.version_info[:2])
```

This would be the rust equivalent of
```rs
let pybuilddir = format!("build/lib.{}-{}-{}", env!("CARGO_CFG_TARGET_OS"), env!("CARGO_CFG_TARGET_ARCH"), minor)
```
at least with my cross compiled version, I get the added `-arm` between the os and the version.

However since were searching for the file in order to find the version, we have to search for all the versions that PyO3 supports.

Progdrasil:
And depending on the os used, sometimes the kernel version is included. So I would need to re-implement [`get_platform`][1] in order for this to be accurate on other unix systems other than linux.

[1]: https://github.com/python/cpython/blob/master/Lib/sysconfig.py#L605

kngwyu:
> Python saves the file during build time at the path defined by:

Sorry what is 'the file'... is it different from `/usr/lib/python3.8/_sysconfigdata__linux_x86_64-linux-gnu.py` on my Linux PC?
Is it placed on a different situation in cross-compile case?

build.rs
Progdrasil: [seems to pass](https://github.com/PyO3/pyo3/pull/1095/checks?check_run_id=994290675)

Progdrasil:
The `sysconfigdata` file, sorry. When building from source, whether it be for cross compilation or not, it will always be saved at a path that is equivalent to:
```py
pybuilddir = 'build/lib.%s-%s' % (get_platform(), sys.version_info[:2])
```
where `get_platform()` is defined as

```
def get_platform():
    """Return a string that identifies the current platform.
    This is used mainly to distinguish platform-specific build directories and
    platform-specific built distributions.  Typically includes the OS name and
    version and the architecture (as supplied by 'os.uname()'), although the
    exact information included depends on the OS; on Linux, the kernel version
    isn't particularly important.
    Examples of returned values:
       linux-i586
       linux-alpha (?)
       solaris-2.6-sun4u
    Windows will return one of:
       win-amd64 (64bit Windows on AMD64 (aka x86_64, Intel64, EM64T, etc)
       win32 (all others - specifically, sys.platform is returned)
    For other non-POSIX platforms, currently just returns 'sys.platform'.
    """
```
And it seems when it is packaged for distribution, the `sysconfigdata` file is placed top levle along side the `libpythonX.Y.so`  file.
So these are the possibilities

```sh
{PREFIX}/_sysconfigdata*.py # distribution from package manager
{PREFIX}/build/lib.linux-x86_64-{MINOR}/_sysconfigdata*.py # if built from source for host
{PREFIX}/build/lib.{OS}(-{OS-KERNEL-RELEASE})?-{ARCH}-{MINOR}/_sysconfigdata*.py # if cross compiled, kernel release is only present on certain OS targets.
```

Cargo.toml
Progdrasil: no longer necessary

build.rs
Progdrasil: applied

build.rs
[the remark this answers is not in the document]
Progdrasil: please see
https://github.com/PyO3/pyo3/pull/1095/files#diff-a7b0a2dee0126cddf994326e705a91eaR260

build.rs
Progdrasil: applied

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Allows `&Self` as a `#[pymethods]` argument again. [#1071](https://github.com/PyO3/pyo3/pull/1071)
 - Fix best-effort build against PyPy 3.6. #[1092](https://github.com/PyO3/pyo3/pull/1092)
 - Improve lifetime elision in `#[pyproto]`. [#1093](https://github.com/PyO3/pyo3/pull/1093)
+- Fix python configuration detection when cross-compiling. [1077](https://github.com/PyO3/pyo3/issues/1077)
```
davidhewitt, line 45: So far we've always linked against the PR that changes things, not the issue. Though maybe this is ok, @kngwyu?

build.rs
```diff
@@ -76,6 +76,59 @@ impl FromStr for PythonInterpreterKind {
     }
 }
 
+struct CrossPython {
```
davidhewitt, line 79:
```suggestion
struct CrossCompileConfig {
```

build.rs
davidhewitt: Nice, happy to see I'm wrong!

build.rs
```diff
@@ -134,10 +187,149 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
+    output
+        .lines()
+        .filter_map(|line| {
+            let mut i = line.splitn(2, ' ');
+            Some((i.next()?.into(), i.next()?.into()))
+        })
+        .collect()
+}
+
+fn as_bool(config: &HashMap<String, String>, key: &str) -> Result<bool> {
```
davidhewitt, line 200: I think `.as` in the stdlib is typically used for zero-cost coercions.

```suggestion
fn get_bool(config: &HashMap<String, String>, key: &str) -> Result<bool> {
```

build.rs
```diff
@@ -134,10 +187,149 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
+        .map(|x| x.as_str())
+        .ok_or(format!("{} is not defined", key))?
+    {
+        "1" | "true" | "True" => Ok(true),
+        "0" | "false" | "False" => Ok(false),
+        _ => Err(format!("{} must be a bool (1/true/True or 0/false/False", key).into()),
+    }
+}
+
+fn as_numeric<T: FromStr>(config: &HashMap<String, String>, key: &str) -> Result<T> {
```
davidhewitt, line 212: Same as above

```suggestion
fn get_numeric<T: FromStr>(config: &HashMap<String, String>, key: &str) -> Result<T> {
```

build.rs
```diff
@@ -188,6 +372,21 @@ fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, Strin
     Ok((interpreter_config, fix_config_map(config_map)))
 }
 
+#[allow(unused_variables)]
```
davidhewitt, line 375: Looks like temporary hack that's no longer needed?

build.rs
```diff
@@ -76,6 +76,59 @@ impl FromStr for PythonInterpreterKind {
     }
 }
 
+struct CrossPython {
+    lib_dir: String,
+    include_dir: Option<String>,
```
davidhewitt, line 81: Maybe `PathBuf` and `Option<PathBuf>`?

build.rs
```diff
@@ -76,6 +76,59 @@ impl FromStr for PythonInterpreterKind {
...
+        Ok(CrossPython {
+            lib_dir: CrossPython::validate_variable("PYO3_CROSS_LIB_DIR")?,
+            include_dir: None,
+            os: env::var("CARGO_CFG_TARGET_OS").unwrap(),
+            arch: env::var("CARGO_CFG_TARGET_ARCH").unwrap(),
+        })
+    }
+
+    fn validate_variable(var: &str) -> Result<String> {
+        let path = match env::var(var) {
```
davidhewitt, line 104: I think this should use `var_os`, which doesn't attempt to sanitize the variable.

build.rs
```diff
@@ -188,6 +372,21 @@ fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, Strin
...
+    python_paths: CrossPython,
+) -> Result<(InterpreterConfig, HashMap<String, String>)> {
+    let target_family = env::var("CARGO_CFG_TARGET_FAMILY")?;
+    // Because compiling for windows on linux still includes the unix target family
+    if target_family == "unix" {
+        // Configure for unix platforms using the sysconfigdata file
+        load_cross_compile_from_sysconfigdata(python_paths)
+    } else {
+        // Must configure by headers on windows platform
+        load_cross_compile_from_headers(python_paths)
```
davidhewitt, line 386: I'd also be happy for us to just hard-code an implementation for windows which is inspired by `sysconfig.py`'s non-posix function: https://github.com/python/cpython/blob/2bcd0fe7a5d1a3c3dd99e7e067239a514a780402/Lib/sysconfig.py#L426-L435

I can do that in a separate PR, if you like. If so let's remember to create an issue once this PR is merged.

guide/src/building_and_distribution.md
```diff
@@ -45,20 +45,28 @@ Cross compiling PyO3 modules is relatively straightforward and requires a few pi
 * A toolchain for your target.
 * The appropriate options in your Cargo `.config` for the platform you're targeting and the toolchain you are using.
 * A Python interpreter that's already been compiled for your target.
+* A Python interpreter that is built for your host and available through the `PATH`.
```
davidhewitt, line 48: I think technically it can also be specified using `PYO3_PYTHON` variable?

build.rs
Progdrasil: You're right, folllowing the [api guidelines][1] it should probably be `to_` or `into_` but would make more sense to implement as a method on `Option<String>` in our own trait.

[1]: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

guide/src/building_and_distribution.md
Progdrasil: You are absolutely right, but its not documented anywhere. I'll add it in the crosscompiling and in an appropriate section in building

build.rs
Progdrasil: But if its kept as an accessor, get effectively more accurate.

build.rs
Progdrasil: I think it would be more appropriate in a separate PR. It might seem out of scope since this one is more geared towards using the sysconfigdata. While windows requires a whole other tactic.

guide/src/building_and_distribution.md
Progdrasil: Oh it is documented, I'm just blind :sweat_smile:

build.rs
davidhewitt: I'll do it in #923; I'll rebase and improve after this merges.

build.rs
```diff
@@ -134,52 +215,148 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
 
-    let major = match patchlevel_defines
-        .get("PY_MAJOR_VERSION")
-        .map(|major| major.parse::<u8>())
-    {
-        Some(Ok(major)) => major,
-        Some(Err(e)) => bail!("Failed to parse PY_MAJOR_VERSION: {}", e),
-        None => bail!("PY_MAJOR_VERSION undefined"),
+fn find_sysconfigdata(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Option<PathBuf> {
+    for f in fs::read_dir(path).expect("Path does not exist") {
```
davidhewitt, line 269: This might need adjusting just very slightly so that it works from just the "lib" dir on unix platforms? See https://github.com/PyO3/pyo3/issues/1110#issuecomment-675179064

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Link against libpython on android with `extension-module` set. [#1082](https://github.com/PyO3/pyo3/issues/1082)
```
davidhewitt, line 46: Did the fix for this get lost along the way? I don't see it in the current version of `build.rs`.

CHANGELOG.md
kngwyu: I think PR is better since it also contains the issue link, though I don't have any strong opinion.

build.rs
```diff
@@ -134,52 +215,148 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
+            Some((i.next()?.into(), i.next()?.into()))
+        })
+        .collect()
+}
 
-    let patchlevel_defines = parse_header_defines(python_include_dir.join("patchlevel.h"))?;
+/// Parse sysconfigdata file
+///
+/// The sysconfigdata is basically a dictionary, and since we can't really use this library to read
+/// it for us, egg and chicken type thing, we parse it with some regex. Here there are two regex's.
```
kngwyu, line 231: Outdated comments?

build.rs
```diff
@@ -76,6 +76,87 @@ impl FromStr for PythonInterpreterKind {
...
+
+    fn get_numeric<T: FromStr>(&self, key: &str) -> Result<T> {
+        self.get(key)
+            .ok_or(format!("{} is not defined", key))?
+            .parse::<T>()
+            .map_err(|_| format!("Could not parse value of {}", key).into())
+    }
+}
+
+struct CrossCompileConfig {
```
kngwyu, line 105: Nicely abstracted :+1:

build.rs
```diff
@@ -134,52 +215,148 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
     config_map
 }
 
-fn load_cross_compile_info() -> Result<(InterpreterConfig, HashMap<String, String>)> {
-    let python_include_dir = env::var("PYO3_CROSS_INCLUDE_DIR")?;
-    let python_include_dir = Path::new(&python_include_dir);
+fn parse_script_output(output: &str) -> HashMap<String, String> {
```
kngwyu, line 218: :+1:

build.rs
```diff
@@ -134,52 +215,148 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
-    let major = match patchlevel_defines
-        .get("PY_MAJOR_VERSION")
-        .map(|major| major.parse::<u8>())
-    {
-        Some(Ok(major)) => major,
-        Some(Err(e)) => bail!("Failed to parse PY_MAJOR_VERSION: {}", e),
-        None => bail!("PY_MAJOR_VERSION undefined"),
+fn find_sysconfigdata(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Option<PathBuf> {
+    for f in fs::read_dir(path).expect("Path does not exist") {
+        return match f {
```
kngwyu, line 270: I think it's better we have some comments that describe why this complicated match is necessary.
[Your issue comment](https://github.com/PyO3/pyo3/pull/1095#issuecomment-674976471) is nice and copy-pasting a part of it is sufficient, though.

build.rs
Progdrasil: Yeah so from what I understand is the fact that if we link to a python installation we get a structure like so

```
/usr/lib/libpython3.Y.so
/usr/lib/python3.Y/_sysconfigdata*.py
```

The issue now is matching against the right python version, so for example if python 3.6, 3.7 and 3.8 are installed on the system, this will simply match the first one found . We might need a new environment variable just to define the version of python if the first one found is not the wanted version.

CHANGELOG.md
Progdrasil: changed it to the pr for both changes, for the linking against android and for the cross-compilation issue

CHANGELOG.md
Progdrasil: it did in fact get lost, I just re-added it

build.rs
Progdrasil: yes just changed it

build.rs
davidhewitt: As it seems like this script will now already work correctly if pointed at `/usr/lib/python3.Y`, maybe we can just document that instead of adding a new variable?

Along that line: to avoid suprises I suggest if multiple python versions are matched, pyo3 should bail and ask for a more specific `PYO3_CROSS_LIB_DIR` to be set. (This seems better to me than silently picking the first one.)

Progdrasil: The problem would arise when needing to link to `libpython` as it would not always be in the lib dir. For example, the raspberry pi seems to have it within the `python3.Y` directory, how ever on Manjaro, only the static library is within that directory and the shared library is top level under `/usr/lib`. So if the user needs to link to `libpython.so` such as when building for android or making an executable, we wouldn't be able to find it.

What I'm thinking of doing is keeping the recursive search, adding the search for the `python3.Y` directory and instead of returning immediately when a match is found, keeping a record of the locations found. Like that we can bail with a useful error message, or we get all the necessary paths. I would still add an optional environment variable to set the version when needed. Since if I have a system with the following file placements, I will not be the only one.

```
/usr/lib/libpython3.7.so
/usr/lib/python3.7/_sysconfigdata*.py
/usr/lib/libpython3.8.so
/usr/lib/python3.8/_sysconfigdata*.py
/usr/lib/python3.8/config-3.8-x86_64-linux-gnu/libpython3.8.a
```

Progdrasil: see commit 7beb272 for the application of this, I noticed that linking still works even if the `libpython` is deep in the directory. Such as when the DSO is only at `/usr/lib/python3.8/config-3.8-x86_64-linux-gnu/libpython3.8.so` and you set `PYO3_CROSS_LIB_DIR=/usr/lib`

build.rs
```diff
@@ -76,6 +76,89 @@ impl FromStr for PythonInterpreterKind {
...
+        })
+    }
+
+    fn lib_only() -> Result<Self> {
+        Ok(CrossCompileConfig {
+            lib_dir: CrossCompileConfig::validate_variable("PYO3_CROSS_LIB_DIR")?,
+            include_dir: None,
+            os: env::var("CARGO_CFG_TARGET_OS").unwrap(),
+            arch: env::var("CARGO_CFG_TARGET_ARCH").unwrap(),
+            version: env::var_os("PYO3_PYTHON_VERSION").map(|s| s.into_string().unwrap()),
```
davidhewitt, line 129: As it's only used when cross compiling, let's keep that explicit:

```suggestion
            version: env::var_os("PYO3_CROSS_PYTHON_VERSION").map(|s| s.into_string().unwrap()),
```

build.rs
```diff
@@ -134,52 +217,202 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
+/// [1]: https://github.com/python/cpython/blob/3.5/Lib/sysconfig.py#L389
+fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
+    let mut sysconfig_paths = search_lib_dir(&cross.lib_dir, &cross);
+    if sysconfig_paths.is_empty() {
+        bail!(
+            "Could not find either libpython.so or _sysconfigdata*.py in {}",
+            cross.lib_dir.display()
+        );
+    } else if sysconfig_paths.len() > 1 {
+        bail!(
```
davidhewitt, line 306: Maybe here before bailing, as a last hope try calling `fs::canonicalize` on all the paths and see if they all resolve to the same thing? That'd fix my machine's edge case.

build.rs
```diff
@@ -134,52 +217,202 @@ fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, St
...
-        .get("Py_ENABLE_SHARED")
-        .map(|x| x.as_str())
-        .ok_or("Py_ENABLE_SHARED is not defined")?
-    {
-        "1" | "true" | "True" => true,
-        "0" | "false" | "False" => false,
-        _ => panic!("Py_ENABLE_SHARED must be a bool (1/true/True or 0/false/False"),
+    let interpreter_config = InterpreterConfig {
+        version: python_version,
+        libdir: python_paths.lib_dir.to_str().map(String::from), //libpython_path.to_str().map(String::from),
```
davidhewitt, line 381: Bonus comment 🚀 

```suggestion
        libdir: python_paths.lib_dir.to_str().map(String::from),
```

davidhewitt (commented):
`7beb272` looks great!

I'm afraid there's one final edge case on my machine:

```
$ ls -l /usr/lib/python*/_sysconfigdata*
-rw-r--r-- 1 root root 24866 Apr 18 03:59 /usr/lib/python3.7/_sysconfigdata_m_linux_x86_64-linux-gnu.py
lrwxrwxrwx 1 root root    35 Apr 27 16:53 /usr/lib/python3.8/_sysconfigdata__linux_x86_64-linux-gnu.py -> _sysconfigdata__x86_64-linux-gnu.py
-rw-r--r-- 1 root root 25708 Apr 27 16:53 /usr/lib/python3.8/_sysconfigdata__x86_64-linux-gnu.py
```

With `PYO3_CROSS_LIB_DIR=/usr/lib` and `PYO3_PYTHON_VERSION=3.8`, the current commit will match both the `3.8` configdatas and bail. See my other comment for a suggestion.

(Not sure why the default python 3.8 install on ubuntu 20.04 is laid out like that. Alas!)

build.rs
Progdrasil: oops

build.rs
Progdrasil: fairly easy fix, see 441d7f5

davidhewitt (approved):
This looks brilliant to me now, thanks so much for your dedication over many revisions! I'll wait a bit to merge just incase others have anything to say.

Progdrasil:
Thanks, it was a nice experience. That being said, having this in a crate of its own will help greatly in testing and debugging all this logic.

kngwyu (approved):

kngwyu:
Thank you!

PR #1098, opened 2020-08-09
Py::as_ref and Py::into_ref (remove AsPyRef)

### LINKED ISSUE #1041: PyObject -> &'py PyAny
I'm implementing a function within an API that's expected to return `&'py PyAny`. The function will serialize a Rust object to a Python object and return a reference to it, bound to the GIL lifetime of `Python<'py>`.

It would be easy to implement a function that returns `PyObject`, and there are various types in PyO3 that will return `&'py PySomething`, eg `PyBool::new`, `PyDict::new`. However I don't see a route to expose Rust objects generically as `&'py PyAny`.

Conceptually I think I want to: allocate an object on the Python heap; move my instance `T` into it; add the new Python object to the auto-release pool; return a reference to the Python object.

The closest I could see was: `PyObject::as_ref` (but that returns a reference bound to the lifetime of `&self`, not to `&'py`); or something involving `*mut ffi::PyObject` and unsafe `Python::from_owned_ptr`.

In case it's relevant, I'm working against PyO3 v0.8.5.

davidhewitt:
I introduce `Py::as_ref` and `Py::into_ref`. We spoke about `Py::into_ref` in #1041.

I moved `AsPyRef::as_ref` to `Py::as_ref` because I think it's easier to document and for users to discover. Also means one more trait can be removed!

Closes #1041

### DIFF
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -60,9 +60,12 @@ This should change very little from a usage perspective. If you implemented trai
 `PyObject` and `Py<T>`, you may find you can just remove the `PyObject` implementation.
 
 ### `AsPyRef` has been removed
-The only implementor of `AsPyRef` was `Py<T>`, so the `AsPyRef::as_ref` method has been moved to
-`Py::as_ref`. This should require no code changes except removing the old `use` for code which
-did not use `pyo3::prelude`.
+As `PyObject` has been changed to be just a type alias, the only remaining implementor of `AsPyRef`
+was `Py<T>`. This removed the need for a trait, so the `AsPyRef::as_ref` method has been moved to
+`Py::as_ref`.
+
+This should require no code changes except removing `use pyo3::AsPyRef` for code which did not use
+`pyo3::prelude::*`.
 
 Before:
 ```rust,ignore
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -14,7 +14,6 @@ Now, the canonical implementation is always `IntoPy`, so downstream crates may n
 accordingly.
 
 Before:
-
 ```rust,ignore
 # use pyo3::prelude::*;
 struct MyPyObjectWrapper(PyObject);
@@ -27,7 +26,6 @@ impl FromPy<MyPyObjectWrapper> for PyObject {
 ```
 
 After
-
 ```rust
 # use pyo3::prelude::*;
 struct MyPyObjectWrapper(PyObject);
@@ -42,7 +40,6 @@ impl IntoPy<PyObject> for MyPyObjectWrapper {
 Similarly, code which was using the `FromPy` trait can be trivially rewritten to use `IntoPy`.
 
 Before:
-
 ```rust,ignore
 # use pyo3::prelude::*;
 # Python::with_gil(|py| {
@@ -61,6 +58,30 @@ let obj: PyObject = 1.234.into_py(py);
 ### `PyObject` is now a type alias of `Py<PyAny>`
 This should change very little from a usage perspective. If you implemented traits for both
 `PyObject` and `Py<T>`, you may find you can just remove the `PyObject` implementation.
+
+### `AsPyRef` has been removed
+The only implementor of `AsPyRef` was `Py<T>`, so the `AsPyRef::as_ref` method has been moved to
+`Py::as_ref`. This should require no code changes except removing the old `use` for code which
+did not use `pyo3::prelude`.
+
+Before:
+```rust,ignore
+use pyo3::{AsPyRef, Py, types::PyList};
+# pyo3::Python::with_gil(|py| {
+let list_py: Py<PyList> = PyList::empty(py).into();
+let list_ref: &PyList = list_py.as_ref(py);
+# })
+```
+
+After:
+```rust
+use pyo3::{Py, types::PyList};
+# pyo3::Python::with_gil(|py| {
+let list_py: Py<PyList> = PyList::empty(py).into();
+let list_ref: &PyList = list_py.as_ref(py);
+# })
+```
+
 
 ## from 0.10.* to 0.11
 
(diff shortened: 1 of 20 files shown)


guide/src/migration.md
```diff
@@ -62,6 +59,30 @@ let obj: PyObject = 1.234.into_py(py);
 This should change very little from a usage perspective. If you implemented traits for both
 `PyObject` and `Py<T>`, you may find you can just remove the `PyObject` implementation.
 
+### `AsPyRef` has been removed
+The only implementor of `AsPyRef` was `Py<T>`, so the `AsPyRef::as_ref` method has been moved to
```
kngwyu, line 63: and `PyObject`? (though it's removed)

birkenfeld:
I'm confused by `into_ref`. In what cases do you need to / should you use this instead of using `as_ref` and letting the `Py` be dropped naturally?

guide/src/types.md
```diff
@@ -140,9 +140,10 @@ Can be cloned using Python reference counts with `.clone()`.
 # let py = gil.python();
 let list: Py<PyList> = PyList::empty(py).into();
 
-// Access the native type using AsPyRef::as_ref(py)
-// (For #[pyclass] types, as_ref() will return &PyCell<T>)
+// Access the native type using Py::as_ref(py) or Py::into_ref(py)
+// (For #[pyclass] types, these will return &PyCell<T>)
 let _: &PyList = list.as_ref(py);
+let _: &PyList = list.clone().into_ref(py);
```
birkenfeld, line 146: Describe difference(s) here?

src/instance.rs
```diff
@@ -33,15 +33,23 @@ pub unsafe trait PyNativeType: Sized {
...
 }
 
-/// A Python object of known type.
+/// A Python object of known type T.
 ///
-/// Accessing this object is thread-safe, since any access to its API requires a
-/// `Python<'py>` GIL token.
+/// Accessing this object is thread-safe, since any access to its API requires a `Python<'py>` GIL
+/// token. There are a few different ways to use the Python object contained:
+///  - [`Py::as_ref`](#method.as_ref) to borrow a GIL-bound reference to the contained object,
```
birkenfeld, line 40:
```suggestion
///  - [`Py::as_ref`](#method.as_ref) to borrow a GIL-bound reference to the contained object.
```

src/instance.rs
```diff
@@ -33,15 +33,23 @@ pub unsafe trait PyNativeType: Sized {
...
+///    [the `PyCell` guide entry](https://pyo3.rs/master/class.html#pycell-and-interior-mutability)
+///    ).
+///  - Use methods directly on `Py`, such as [`Py::call`](#method.call) and
+///    [`Py::call_method`](#method.call_method).
 ///
 /// See [the guide](https://pyo3.rs/master/types.html) for an explanation
 /// of the different Python object types.
 ///
-/// Technically, it is a safe wrapper around `NonNull<ffi::PyObject>` with
+/// `Py<T>` is implemented as a safe wrapper around `NonNull<ffi::PyObject>` with
```
birkenfeld, line 52: Do we need this sentence? It always struck me as more confusing than helpful to the user, and the hardcore developer will click the [src] link anyway to check it out.

src/instance.rs
```diff
@@ -63,7 +71,83 @@ where
...
+    pub fn as_ref<'py>(&'py self, _py: Python<'py>) -> &'py T::AsRefTarget {
+        let any = self.as_ptr() as *const PyAny;
+        unsafe { PyNativeType::unchecked_downcast(&*any) }
+    }
 
+    /// Similar to [`as_ref`](#method.as_ref), but instead consumes this `Py` and registers the
+    /// Python object reference in PyO3's object storage. The reference count for the Python
+    /// object will not be decreased until the GIL lifetime ends.
+    ///
+    /// # Examples
```
birkenfeld, line 119: These are the same examples as above, with `as_ref` changed to `into_ref`.  Instead, it would be nice to (sorry, again) mention the differences and show something that *can't* be done with `as_ref`.

davidhewitt:
> I'm confused by into_ref. In what cases do you need to / should you use this instead of using as_ref and letting the Py be dropped naturally?

Great feedback. We've seen in rare cases APIs want to return a reference but have a `Py` value they have created (e.g. from `into_py()`). This has come up in #1041 as well as #1083.

I'll rework documentation later based on the comments here.

birkenfeld:
Ah, good point. I did not realize that `as_ref` had the self lifetime and the Python lifetime tied to the return's lifetime. (I'm sure there is a reason this is necessary?)

davidhewitt:
Yep - the returned reference for `as_ref` needs to be the shorter of the two lifetimes, which is why they're both set the same. 

(It cannot outlive `&self` because that's the data it's borrowing, and it cannot outlive `Python` because it's GIL-bound.)

src/instance.rs
davidhewitt: I'm happy to take it out 👍

davidhewitt:
Thanks for the feedback; I've pushed the suggested improvements to the docs.

src/instance.rs
```diff
@@ -63,7 +68,73 @@ where
...
+    /// Similar to [`as_ref`](#method.as_ref), and also consumes this `Py` and registers the
+    /// Python object reference in PyO3's object storage. The reference count for the Python
+    /// object will not be decreased until the GIL lifetime ends.
+    ///
+    /// # Example
+    ///
+    /// Useful when returning GIL-bound references from functions:
+    /// ```
+    /// # use pyo3::prelude::*;
+    /// fn new_py_any(py: Python, value: impl IntoPy<PyObject>) -> &PyAny {
```
birkenfeld, line 121: One last suggestion: add an explicit lifetime to `Python` and `&PyAny`?

davidhewitt: Agreed 👍

PR #805, opened 2020-03-13
Can we export exceptions?

kngwyu:
In this PR I added a new macro named `export_exceptions`, which inserts
```rust
#[pymethods]
fn _pyo3_exceptions(_py: pyo3::Python, m: &pyo3::types::PyModule) -> pyo3::PyResult<()> {
    use pyo3::type_object::PyTypeObject;
    m.add("PyBorrowError", pyo3::pycell::PyBorrowError::type_object())?;
    m.add(
        "PyBorrowMutError",
        pyo3::pycell::PyBorrowMutError::type_object(),
    )?;
    Ok(())
}
```
to exension, so that our custom exceptions can be imported to Python code.

But somehow `except PyBorrowError` fails oddly like
```python
____________________________________________________________ test_catch_borrrow ____________________________________________________________

    def test_catch_borrrow():
        m = Mutable()
        try:
            m.invalid_borrow(m)
            assert False
        except PyBorrowError:
            pass
        except Exception as e:
>           raise e

tests/test_pycell_exceptions.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def test_catch_borrrow():
        m = Mutable()
        try:
>           m.invalid_borrow(m)
E           _pyo3_exceptions.PyBorrowError
```
and also `isintance(e, PyBorrowError)` returns False.

PR #1099, opened 2020-08-10
guide: show exporting create_exception! exceptions in a module

birkenfeld:
This was asked about in Gitter; #805 seemed to suggest to them that it is not possible.

### DIFF
--- a/guide/src/exception.md
+++ b/guide/src/exception.md
@@ -33,6 +33,23 @@ fn main() {
 }
 ```
 
+When using PyO3 to create an extension module, you can add the new exception to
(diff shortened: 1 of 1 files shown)


davidhewitt:
Thank you for adding this!

kngwyu:
Can we use these exceptions for catching error?
In #805, somehow an exception raised in `#[pyfunction]` and imported one are treated as different errors.

birkenfeld:
This worked when I tried it:
```
try:
    rustmod.func()
except rustmod.CustomError:
    ...
```
where `rustmod.func()` raises the exception.

PR #1107, opened 2020-08-15
Left-hand operands are fellback to right-hand ones for type mismatching

### LINKED ISSUE #844: __radd__ etc don't get called if __add__ is also defined

kngwyu:
Fixes #844. 
Include some other refactorings:
- ` SlotSetter` is refactored not to have `skipped_setters` field.
- `py_ternary_num_func` macros are removed since they are just called once.
- `#[macro_export]` for macros in `class/macros.rs` are removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Improve lifetime elision in `#[pyproto]`. [#1093](https://github.com/PyO3/pyo3/pull/1093)
 - Fix python configuration detection when cross-compiling. [#1095](https://github.com/PyO3/pyo3/pull/1095)
 - Link against libpython on android with `extension-module` set. [#1095](https://github.com/PyO3/pyo3/pull/1095)
+- Fix support for both `__add__` and `__radd__`  in the `+` operator when both are defined in `PyNumberProtocol`
+  (and similar for all other reversible operators). [#1107](https://github.com/PyO3/pyo3/pull/1107)
 
 ## [0.11.1] - 2020-06-30
 ### Added
--- a/pyo3-derive-backend/src/defs.rs
+++ b/pyo3-derive-backend/src/defs.rs
@@ -65,18 +65,13 @@ pub struct SlotSetter {
     pub proto_names: &'static [&'static str],
     /// The name of the setter called to the method table.
     pub set_function: &'static str,
-    /// Represents a set of setters disabled by this setter.
-    /// E.g., `set_setdelitem` have to disable `set_setitem` and `set_delitem`.
-    pub skipped_setters: &'static [&'static str],
 }
 
 impl SlotSetter {
-    const EMPTY_SETTERS: &'static [&'static str] = &[];
     const fn new(names: &'static [&'static str], set_function: &'static str) -> Self {
         SlotSetter {
             proto_names: names,
             set_function,
-            skipped_setters: Self::EMPTY_SETTERS,
         }
     }
 }
@@ -144,11 +139,7 @@ pub const OBJECT: Proto = Proto {
         SlotSetter::new(&["__hash__"], "set_hash"),
         SlotSetter::new(&["__getattr__"], "set_getattr"),
         SlotSetter::new(&["__richcmp__"], "set_richcompare"),
-        SlotSetter {
-            proto_names: &["__setattr__", "__delattr__"],
-            set_function: "set_setdelattr",
-            skipped_setters: &["set_setattr", "set_delattr"],
-        },
+        SlotSetter::new(&["__setattr__", "__delattr__"], "set_setdelattr"),
         SlotSetter::new(&["__setattr__"], "set_setattr"),
         SlotSetter::new(&["__delattr__"], "set_delattr"),
         SlotSetter::new(&["__bool__"], "set_bool"),
@@ -379,11 +370,7 @@ pub const MAPPING: Proto = Proto {
     slot_setters: &[
         SlotSetter::new(&["__len__"], "set_length"),
         SlotSetter::new(&["__getitem__"], "set_getitem"),
-        SlotSetter {
-            proto_names: &["__setitem__", "__delitem__"],
-            set_function: "set_setdelitem",
-            skipped_setters: &["set_setitem", "set_delitem"],
-        },
+        SlotSetter::new(&["__setitem__", "__delitem__"], "set_setdelitem"),
         SlotSetter::new(&["__setitem__"], "set_setitem"),
         SlotSetter::new(&["__delitem__"], "set_delitem"),
     ],
@@ -446,11 +433,7 @@ pub const SEQ: Proto = Proto {
         SlotSetter::new(&["__concat__"], "set_concat"),
         SlotSetter::new(&["__repeat__"], "set_repeat"),
         SlotSetter::new(&["__getitem__"], "set_getitem"),
-        SlotSetter {
-            proto_names: &["__setitem__", "__delitem__"],
-            set_function: "set_setdelitem",
-            skipped_setters: &["set_setitem", "set_delitem"],
-        },
+        SlotSetter::new(&["__setitem__", "__delitem__"], "set_setdelitem"),
         SlotSetter::new(&["__setitem__"], "set_setitem"),
         SlotSetter::new(&["__delitem__"], "set_delitem"),
         SlotSetter::new(&["__contains__"], "set_contains"),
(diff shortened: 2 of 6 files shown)


kngwyu:
And I changed right hand methods to use `extract_or_return_not_implemented`, so now we can write tests that confirm that `NotImplemented` is correctly returned when both of `__add__` and `__radd__` is defined. cc: @mvaled

kngwyu:
Sigh, coverage decreased after I added some tests... :sweat:

tests/test_arithmetics.rs
```diff
@@ -292,46 +328,182 @@ impl PyNumberProtocol for LhsAndRhsArithmetic {
...
+    py_run!(py, c, "assert 1 >> c == '1 >> BA'");
+    py_run!(py, c, "assert 1 & c == '1 & BA'");
+    py_run!(py, c, "assert 1 ^ c == '1 ^ BA'");
+    py_run!(py, c, "assert 1 | c == '1 | BA'");
     py_run!(py, c, "assert 1 ** c == '1 ** BA'");
 }
 
+#[pyclass]
+#[derive(Debug)]
+struct LhsFellbackToRhs {}
```
davidhewitt, line 397: Rather than making a new `#[pyclass]` here, was it possible to just use the renamed `LhsAndRhsArithmetic` class? There's four sets of tests which we can write:

```
#[test]
fn can_call_lhs_explicitly() {
    py_run!(py, c, "assert c.__add__(1) == '1 + BA'");
}

#[test]
fn can_call_rhs_explicitly() {
    py_run!(py, c, "assert c.__radd__(1) == '1 + RA'");
    // etc for all operators
}

#[test]
fn can_call_lhs_operation() {
     py_run!(py, c, "assert c + 1 == '1 + BA'");
    // etc for all operators
}

#[test]
fn can_call_rhs_operation() {
     py_run!(py, c, "assert 1 + c == '1 + RA'");
    // etc for all operators
}
```

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -134,23 +134,22 @@ fn slot_initialization(
...
     // Collect initializers
     let mut initializers: Vec<TokenStream> = vec![];
-    'outer_loop: for m in proto.slot_setters {
-        if skipped_setters.contains(&m.set_function) {
+    for m in proto.slot_setters {
+        // Skip if any required protocol are not implemented or already used.
+        if m.proto_names
+            .iter()
+            .any(|name| !method_names.contains(*name) || used_protocols.contains(name))
+        {
```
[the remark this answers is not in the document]
kngwyu: Thanks, I encapsuled it to an iterator.

tests/test_arithmetics.rs
kngwyu: Thanks, I simplified the test cases, though I still stuck with `lhs_overrides_rhs` and `lhs_fellback_to_rhs`.

tests/test_arithmetics.rs
```diff
@@ -292,44 +338,94 @@ impl PyNumberProtocol for LhsAndRhsArithmetic {
...
+    fn __rmatmul__(&self, other: &PyAny) -> String {
+        format!("{:?} @ RA", other)
     }
 }
 
 #[pyproto]
-impl PyObjectProtocol for LhsAndRhsArithmetic {
+impl PyObjectProtocol for LhsAndRhs {
     fn __repr__(&self) -> &'static str {
         "BA"
```
davidhewitt, line 377: I wonder if `LhsAndRhs` should have a field `name: String`, and then this could be:

```suggestion
        &self.name
```

That way the tests could be written as e.g. `"assert a.__radd__(b) == 'B + A'"`

(I find `RA` and `BA` slightly confusing. It's easy enough to look back up at the pyclass to figure it out though, so this isn't a strong feeling.)

pyo3-derive-backend/src/defs.rs
```diff
@@ -30,6 +31,40 @@ impl Proto {
     {
         self.py_methods.iter().find(|m| query == m.name)
     }
+    // Since the order matters, we expose only the iterator instead of the slice.
+    pub(crate) fn setters(&self, implemented: HashSet<String>) -> SlotSetterIter {
+        SlotSetterIter {
+            iter: self.slot_setters.iter(),
+            implemented_protocols: implemented,
```
davidhewitt, line 38: I quite like `std::iter::from_fn`, and return `impl Iterator<Item = &'static str>` instead of making a new struct, but I am happy with this way too 👍

davidhewitt (approved):
This looks really great! Feel free to take on or ignore any of these minor ideas as you like before merging 😄

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Allows `&Self` as a `#[pymethods]` argument again. [#1071](https://github.com/PyO3/pyo3/pull/1071)
 - Fix best-effort build against PyPy 3.6. #[1092](https://github.com/PyO3/pyo3/pull/1092)
 - Improve lifetime elision in `#[pyproto]`. [#1093](https://github.com/PyO3/pyo3/pull/1093)
+- `__add__` is correctly fellback to `__radd__` when both of them are defined. [#1107](https://github.com/PyO3/pyo3/pull/1107)
```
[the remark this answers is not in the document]
davidhewitt: Actually I think the fallback is an implementation detail the user doesn't need to know about, maybe this version says what they need to care about:

```suggestion
- Fix support for both `__add__` and `__radd__`  in the `+` operator when both are defined in `PyNumberProtocol` (and similar for all other reversible operators). [#1107](https://github.com/PyO3/pyo3/pull/1107)
```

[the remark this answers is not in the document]
kngwyu: :+1:

tests/test_arithmetics.rs
kngwyu: I tried so but it wasn't so meaningful when we want to test which operator is called.

pyo3-derive-backend/src/defs.rs
kngwyu: Thanks, I didn't know that function.

tests/test_arithmetics.rs
```diff
@@ -280,10 +280,56 @@ fn rhs_arithmetic() {
 }
 
 #[pyclass]
-struct LhsAndRhsArithmetic {}
+struct LhsAndRhs {}
```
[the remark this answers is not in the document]
kngwyu: Since the behavior of `__add__` is almost the same as `+`, I don't see strong necessity for it.

pyo3-derive-backend/src/defs.rs
```diff
@@ -30,6 +31,28 @@ impl Proto {
     {
         self.py_methods.iter().find(|m| query == m.name)
     }
+    // Since the order matters, we expose only the iterator instead of the slice.
+    pub(crate) fn setters(
+        &self,
+        mut implemented_protocols: HashSet<String>,
+    ) -> impl Iterator<Item = &'static str> {
+        self.slot_setters.iter().filter_map(move |setter| {
```
davidhewitt, line 39: Nice one, this implementation is really elegant!

davidhewitt (approved):
💯

PR #1108, opened 2020-08-16
Test MSRV on github actions

kngwyu:
And some (boring) refactorings.


kngwyu:
> I notice the CI jobs are running twice - "pull_request" and "push". Maybe we should change the "push" trigger to only run on the master branch?

Agreed, changed to do so.

PR #1114, opened 2020-08-25
Add optional support for conversion from Hashbrown types

mtreinish:
This commit adds optional support for conversion from hashbrown's [1]
HashMap [2] and HashSet [3] types. The HashMap and HashSet implementation
in std::collections is a copy from HashBrown, but Hashbrown still
provides some features over the std::collections version. Primarily this
is rayon support and also using a default hasher which is faster
(although not DOS resistant). The hashbrown versions provide a drop in
replacement over std::collections to get these features. To take
advantage of native type conversion in PyO3 this commit adds hashbrown
as an optional dependency and when the feature is enabled the traits for
going between python and hashbrown::HashMap and hashbrown::HashSet are
available. This is handy for users of hashbrown which have to inline
these conversions manually in functions that take dicts as args.

[1] https://github.com/rust-lang/hashbrown
[2] https://docs.rs/hashbrown/0.8.2/hashbrown/struct.HashMap.html
[3] https://docs.rs/hashbrown/0.8.2/hashbrown/struct.HashSet.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Implement type information for conversion failures. [#1050](https://github.com/PyO3/pyo3/pull/1050)
 - Add `PyBytes::new_with` and `PyByteArray::new_with` for initialising Python-allocated bytes and bytearrays using a closure. [#1074](https://github.com/PyO3/pyo3/pull/1074)
 - Add `Py::as_ref` and `Py::into_ref`. [#1098](https://github.com/PyO3/pyo3/pull/1098)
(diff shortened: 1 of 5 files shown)


davidhewitt:
Thanks, could you please add a `CHANGELOG` entry and also update the `dict` and `set` entries of `guide/conversions.md` (with a note to hint at the optional dependency similar to `num_complex::Complex` 👍

kngwyu (approved):
Thanks!

mtreinish:
> Thanks, could you please add a `CHANGELOG` entry and also update the `dict` and `set` entries of `guide/conversions.md` (with a note to hint at the optional dependency similar to `num_complex::Complex` +1

Sure thing, added in: https://github.com/PyO3/pyo3/pull/1114/commits/659d24e1a9e4e208054e94a7353aed5bbd908bd9

davidhewitt (approved):
Awesome, thank you!

PR #1115, opened 2020-08-25
Implement std::error::Error for PyErr

### LINKED ISSUE #682: PyErr does not implement std::error::Error

### LINKED ISSUE #1034: Question: How to convert Python Exception (PyErr) into string for logging
I have a python script which raises exception. How do I print this in the rust log?

davidhewitt:
This is a rework of `PyErr` to implement `std::error::Error` (and `std::fmt:Display`).

The rework was necessary because functions like `PyErr::normalize()` taking `&mut self` and `PyErr::instance()` taking `self` were incompatible with `std::error::Error`.

A couple of notes about the implementation:
- `Display` for `PyError` automatically uses `Python::with_gil` internally. I think as this is the error path, and also it's rare that the GL will not already be acquired, this overhead is fine.
- `PyErr` now contains an `UnsafeCell`. The safety is assured because, once normalized, the `UnsafeCell` value never changes. Almost every nontrivial access to the `PyErr` immediately normalizes it.

Closes #682
Closes #1034 

TODO: 
- [x] ~~Lots of documentation / guide updates!~~
- [x] ~~Add more test coverage.~~
- [x] ~~Rename `PyConcreteException::py_err` (... to `PyConcreteException::new_err` ?)~~
- [ ] Add `std::error::Error::cause()` ?


kngwyu:
Thank you for this hard work and refactoring!

It's well designed overall, but I want to list up some design questions before reviewing in detail:
- `PyErrValue`. Now it's only used in `Lazy` variant, so maybe another name is suitable. In addition, maybe it's better to have `Lazy { ptype: Py<PyType>, value: Option<PyErrValue> }` instead of `PyErrValue::None`.
- `Normalizing` variant of `PyErrState`. I think some other ways (`Option` or boolean flag) also work and should be considered.
- Visibility. Though I haven't checked it precisely, are `PyErrState` and `PyErrValue` need to be `pub`?

davidhewitt:
Thanks, good points:

- I had kept `None` / `Normalizing` in their respective enums because I thought it was more layout-efficient. But from a quick test I see that rustc is clever enough to optimize the size of `Option<SomeEnum>`:

   ```
   std::mem::size_of::<PyErrState>() = 40
   std::mem::size_of::<Option<PyErrState>>() = 40
   std::mem::size_of::<PyErrValue>() = 24
   std::mem::size_of::<Option<PyErrValue>>() = 24
   ```

   So I think in both cases let's use `Option`.

- Visibility: You're correct, I think both `PyErrValue` and `PyErrState` can be `pub(crate)`.

I'll push a commit tweaking as suggested tonight.

davidhewitt:
(still planning to push fix later, sorry didn't get round to it yesterday.)

davidhewitt:
This is ready for a first round of review. Once we've made any design changes necessary, I also need to write more tests and the migration guide entry before we can merge this.

src/exceptions.rs
```diff
@@ -609,9 +610,8 @@ mod test {
         let py = gil.python();
         let err = py
             .run("raise Exception('banana')", None, None)
-            .expect_err("raising should have given us an error")
-            .instance(py);
-        write!(&mut out, "{}", err).expect("successful format");
+            .expect_err("raising should have given us an error");
+        write!(&mut out, "{}", err.instance(py)).expect("successful format");
```
birkenfeld, line 614: if `PyErr` now implements Display, shouldn't that work without the `instance` call?

birkenfeld (commented):
While overhauling `PyErr`, can we fix the pretty confusing and duplicate APIs around creating new errors?

I mean mostly `PyConcreteError::py_err` and `PyConcreteError::into`:

* for me, the first doesn't really make the association with `PyErr`. I'd much rather this be called `new`: AFAICT, that method is not defined currently.

* the second reads really strange, since I'm hardwired to think of `into()` as a self-consuming method.

The two are also named completely differently, although the do very similar things (AFAICS, `into()` is just `Err(py_err())`).  IMO `into()` can be removed, I can't think of any other library that provides this shortcut.

I also think the `PyErr: Into<PyResult>` impl is unidiomatic and should be removed.

davidhewitt:
👍 I'm completely happy to rework `PyException::py_err` and `PyException::into`. I was wondering along similar lines.

I'm definitely in favour of removing `.into` and agree on the impls being unidiomatic, so I'll remove those immediately.

Regarding `py_err` vs `new`, the only thing that makes me unsure about this is that `PyConcreteException::new()` could also return `&PyConcreteException` similar to all the rest of the `PyType::new()`. Perhaps `PyConcreteException::new_err() -> PyErr` is a less ambiguous choice?

birkenfeld:
Yes, I was thinking about that too. `new_err` is certainly nicer than `py_err`, but since the concrete exception structs are just void structs, I don't think you can have a useful other definition of `new` there?

davidhewitt:
Since #1024 it's possible to have a reference to all the exception objects in just the same way as for the rest of the native types (i.e. they're no longer void structs).

davidhewitt:
The last piece of the puzzle, which I haven't finished yet but plan to do soon, is have a `#[pyexception]` macro which is pretty much just `#[pyclass]` with extra goodness to make exception types.

birkenfeld:
Oops, I missed that. Yes, then it makes sense not to call that `new`.

kngwyu:
`new_err` sounds good to me, though I mostly use `PyErr::new` and don't have any strong opinion about how the exception constructor would be.

birkenfeld:
BTW, could `PyErr::new` be made nicer to use with turbofish by making `V` an `impl ...` parameter?

davidhewitt:
Sadly it's not possible to combine turbofish with `impl Trait` parameters: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=81e2e2f20891dbb7456b3480b2fede79

src/err/err_value.rs
```diff
@@ -0,0 +1,12 @@
+use crate::{PyObject, Python, ToPyObject};
+
+pub(crate) enum PyErrValue {
```
kngwyu, line 3: How about moving this to `err_state.rs`?

src/err/mod.rs
```diff
@@ -164,27 +140,80 @@ impl PyErr {
...
+        };
+
+        PyErr::from_state(state)
+    }
+
+    /// Get the type of this exception object.
+    ///
+    /// The object will be normalized first if needed.
+    ///
+    /// # Example
```
kngwyu, line 172: Nice examples :+1:

davidhewitt:
One last point of discussion: At the moment there's no implementation for `std::error::Error::cause()`, because it needs to return `Option<&dyn std::error::Error>`.

At the moment we only offer this implementation for `&PyBaseException`, because we can attach the cause object to the GIL lifetime.

I think it should be possible to implement for `PyErr`. We'd need to have a kind of "mini" object storage which the `.cause()` could be stored in, which could be cleaned up in `impl Drop for PyErr`. It feels complicated to get right, which is why I wanted to separate it from the rest of this PR (if we choose to implement it at all).

src/err/mod.rs
```diff
@@ -412,62 +387,95 @@ impl PyErr {
...
+            (*self.state.get())
+                .take()
+                .expect("Cannot normalize a PyErr while already normalizing it.")
+        };
+        let (mut ptype, mut pvalue, mut ptraceback) = state.into_ffi_tuple(py);
+
+        unsafe {
+            ffi::PyErr_NormalizeException(&mut ptype, &mut pvalue, &mut ptraceback);
+            let self_state = &mut *self.state.get();
+            *self_state = Some(PyErrState::Normalized(PyErrStateNormalized {
```
kngwyu, line 424: Nicer than the previous one with `transmute` :+1:

kngwyu (approved):
:100:

kngwyu:
> I think it should be possible to implement for PyErr. We'd need to have a kind of "mini" object storage which the .cause() could be stored in, which could be cleaned up in impl Drop for PyErr.

But what is the `source`? `PyErrValue`? Or Python's backtrace?

src/err/mod.rs
```diff
@@ -100,8 +79,7 @@ impl PyErr {
     /// In most cases, you can use a concrete exception's constructors instead:
```
birkenfeld, line 79: Nit: constructors -> constructor

davidhewitt:
> But what is the source? PyErrValue? Or Python's backtrace?

It would be the normalized python exceptions's cause, as seen at https://docs.python.org/3/c-api/exceptions.html#c.PyException_GetCause

davidhewitt:
Thanks for the comments, updated as suggested.

I'll start writing docs and tests later 👍

davidhewitt:
Rebased, added some further test coverage, and also simplified `PyErrState` further.

I was able to take out the `PyErrValue` enum completely, and changed `PyErrArguments` to use `IntoPy` rather than `ToPyObject` (which means less copying!).

src/gil.rs
```diff
@@ -365,10 +365,8 @@ fn decrement_gil_count() {
     });
 }
 
-/// Ensure the GIL is held, useful in implementation of APIs like PyErr::new where it's
-/// inconvenient to force the user to acquire the GIL.
-#[doc(hidden)]
-pub fn ensure_gil() -> EnsureGIL {
+/// Ensure the GIL is held, used in the implementation of Python::with_gil
+pub(crate) fn ensure_gil() -> EnsureGIL {
```
davidhewitt, line 369: `rust-numpy` will need to migrate to `Python::with_gil`

davidhewitt:
Right, I've just pushed a load of documentation for this. I now think this is ready to merge, so would appreciate some final reviews from anyone who's got a few minutes!

guide/src/migration.md
```diff
@@ -5,13 +5,91 @@ For a detailed list of all changes, see [CHANGELOG.md](https://github.com/PyO3/p
...
+without needing to worry about breaking downstream code.
+
+If you needed to access the contents of the `PyErr`, you can use the new methods `PyErr::ptype()`,
+`PyErr::pvalue()` and `PyErr::ptraceback()` to safely do so.
+
+#### `PyErrValue` and `PyErr::from_value` have been removed
+
+As these were part the internals of `PyErr` which have been reworked, these APIs no longer exist.
+
+If you used this API, it is recommended to use `PyException::new_err` (see the next section on
```
kngwyu, line 38: Inserting markdown link would be better than just 'the next section'

guide/src/migration.md
```diff
@@ -5,13 +5,91 @@ For a detailed list of all changes, see [CHANGELOG.md](https://github.com/PyO3/p
...
+Exception types).
+
+#### `Into<PyResult<T>>` for `PyErr` has been removed
+
+This implementation was not idiomatic Rust. Just construct the `Result::Err` variant directly.
```
kngwyu, line 43: Is 'not idiomatic' a too strong argument? Just 'redundant' would make sense.

guide/src/migration.md
```diff
@@ -5,13 +5,91 @@ For a detailed list of all changes, see [CHANGELOG.md](https://github.com/PyO3/p
...
+```rust
+# use pyo3::{PyErr, PyResult, exceptions::PyTypeError};
+let result: PyResult<()> = Err(PyTypeError::new_err("error message"));
+```
+
+### Exception types have been reworked
+
+Previously exception types were zero-sized marker types purely used to construct `PyErr`. In PyO3
+0.12, these types have been replaced with real definitions similar to `PyAny`, `PyDict` etc. This
+makes it is possible to interact with Python exception objects.
```
kngwyu, line 60:
```suggestion
makes it possible to interact with Python exception objects.
```

guide/src/migration.md
```diff
@@ -5,13 +5,91 @@ For a detailed list of all changes, see [CHANGELOG.md](https://github.com/PyO3/p
...
+After (also using the new reworked exception types; see the following section):
+```rust
+# use pyo3::{PyErr, PyResult, exceptions::PyTypeError};
+let result: PyResult<()> = Err(PyTypeError::new_err("error message"));
+```
+
+### Exception types have been reworked
+
+Previously exception types were zero-sized marker types purely used to construct `PyErr`. In PyO3
+0.12, these types have been replaced with real definitions similar to `PyAny`, `PyDict` etc. This
```
kngwyu, line 59: I feel ` real definitions similar to ...` a bit difficult.
What is important is we can use them like `&PyAny`, so I think `these type now has real definitions and is usable like PyAny, PyDict etc.` would be better.

guide/src/migration.md
```diff
@@ -5,13 +5,91 @@ For a detailed list of all changes, see [CHANGELOG.md](https://github.com/PyO3/p
...
+#### `PyErr::new` and `PyErr::from_type` now require `Send + Sync` for their argument
+
+For most uses no change will be needed. If you are trying to construct `PyErr` from a value that is
+not `Send + Sync`, you will need to first create the Python object and then use
+`PyErr::from_instance`.
+
+Similarly, any types which implemented `PyErrArguments` will now need to be `Send + Sync`.
+
+#### `PyErr`'s contents are now private
+
```
kngwyu, line 27: I think it's beneficial to concretely note what has been made private. I.e. we cannot access `err.ptype/pvalue/ptraceback` and should use `err.ptype()/pvalue()/ptraceback()`.

guide/src/migration.md
```diff
@@ -5,13 +5,91 @@ For a detailed list of all changes, see [CHANGELOG.md](https://github.com/PyO3/p
...
+Making the `PyErr` internals private makes it possible for PyO3 to implement optimizations safely
```
kngwyu, line 28: I think this is not *perfectly* true: isn't it more of a safety reason considering normalization?
And it would be maybe too detailed for just a migration guide.

kngwyu (approved):
LGTM, but it would be better if we have another review for the document.
@birkenfeld 
Could you take a glance on it?

davidhewitt:
Thanks for the great feedback on the docs; I've pushed some quick changes based on your review.

davidhewitt:
I'd like to merge this tonight; maybe @sebpuetz is interested in reviewing?

sebpuetz:
Sure, I should find some time within the next hour

src/err/err_state.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+pub(crate) fn boxed_args(
+    args: impl PyErrArguments + 'static,
+) -> Box<dyn FnOnce(Python) -> PyObject + Send + Sync> {
+    Box::new(|py| args.arguments(py))
+}
+
+impl PyErrState {
+    pub fn into_ffi_tuple(
```
sebpuetz, line 47: `PyErrState` is `pub (crate)` but this fn is `pub`, given that this function returns raw pointers, I'd expect that this is not used externally. Could this also be marked `pub (crate)`? Otherwise, adding some documentation would be nice!

src/err/mod.rs
```diff
@@ -0,0 +1,568 @@
...
+    /// Constructs a new error, with the usual lazy initialization of Python exceptions.
+    ///
+    /// `exc` is the exception type; usually one of the standard exceptions
+    /// like `exceptions::PyRuntimeError`.
+    /// `args` is the a tuple of arguments to pass to the exception constructor.
+    pub fn from_type<A>(ty: &PyType, args: A) -> PyErr
+    where
+        A: PyErrArguments + Send + Sync + 'static,
+    {
+        assert_ne!(unsafe { ffi::PyExceptionClass_Check(ty.as_ptr()) }, 0);
```
sebpuetz, line 96: I'm not sure whether I like this kind of assertion here, on the other hand making the error-creation fallible seems weird...

At least the requirements for `ty` should be documented.

sebpuetz: I just saw that the constructor below handles this case by returning a `PyTypeError`, perhaps this is also the right solution here?

src/err/mod.rs
```diff
@@ -0,0 +1,568 @@
...
+            None => std::ptr::null_mut(),
+            Some(obj) => obj.as_ptr(),
+        };
+
+        unsafe {
+            let null_terminated_name =
+                CString::new(name).expect("Failed to initialize nul terminated exception name");
+
+            NonNull::new_unchecked(ffi::PyErr_NewException(
+                null_terminated_name.as_ptr() as *mut c_char,
```
sebpuetz, line 271: Does this take ownership of the `CString` or does `PyErr_NewException` copy the string? Else, I'm not sure how the pointer stays valid after `null_terminated_name` goes out of scope.

sebpuetz (approved):
I didn't have time to do a very thorough review but this looks pretty good, just a handful of questions from my side.

src/err/mod.rs
davidhewitt: It copies the string. (I believe this is the norm for all C APIs - it's pretty scary to steal ownership of things in C unless clearly documented, lest double-free!)

sebpuetz: Knowing this could also have some implications for CStrings in other interfaces like docstrings

src/err/mod.rs
davidhewitt: 👍 refactored to use the same `PyTypeError`.

davidhewitt:
Thanks for the review! In the end I didn't have time to merge yesterday evening. I've just pushed tweaks to address your comments and will merge shortly...

PR #1121, opened 2020-08-28
Update UI tests for Rust 1.46

davidhewitt:
With the new Rust 1.46 it looks like the output on the UI tests has improved slightly. Unfortunately this has broken CI 😄


davidhewitt:
Hmm unsuprisingly MSRV test fails after the update!

I'll find a way to disable the UI tests on MSRV.

kngwyu:
@davidhewitt 
Thank you for the update!

> Hmm unsuprisingly MSRV test fails after the update!

Please check how `rustversion` works for this in `test_compile_errors.rs`.

davidhewitt:
Thanks, I've updated the code in `test_compile_errors.rs`.

PR #1126, opened 2020-08-30
Reduce the direct access to tp_name which is not part of the limited API

alex:
Refs #1125

This reduces the compilation errors with `--features abi3` from over 200 to 87.


alex:
Coverage results are because `impl Debug for $name` was never covered. Let me know if I need to write a test for that to be considered for merging.

davidhewitt:
Thanks, looks good to me. I'm actively improving this section of the code at the moment, so I'll add the tests 👍

alex:
Thanks!

PR #1127, opened 2020-08-30
Do not include PyBuffer APIs when building for abi3

alex:
Refs #1125

Reduces the number of compilation failures with `--features abi3` from 87 to 56.


davidhewitt:
Thanks!

Issue #1076, opened 2020-08-02
Minimum supported Rust version - policy


davidhewitt:
Following on from #1069, where we decided not to update MSRV right now, I just wanted to post a discussion over what we do want to support for MSRV.

Discussion in #1069 suggested we support _at least_ the current version of RHEL, and maybe also Debian.  

**TLDR: I think we should continue to aim to always support at least RHEL, but don't think it's worth looking all the way back to Debian's version of rustc.**

#### RHEL

RHEL updates `rustc` every three months; this is equivalent to every odd-numbered compiler version. https://access.redhat.com/documentation/en-us/red_hat_developer_tools/1/

Currently supports `1.43`.

#### Debian

I also took a look at what Debian requires. It looks like their `rustc` does not receive version updates so frequently; their current version is `1.34`. https://packages.debian.org/buster/rust/rustc

The next Debian version looks like it will support `1.43`. https://packages.debian.org/bullseye/rust/rustc

So we already don't support current Debian, but if we keep MSRV back for a while, eventually we will.

#### Proposal

First I should clarify that _we shouldn't update `rustc` without reason_. Eventually, new dependencies or language features will create desire to update.

When we bump MSRV in a pyo3 update, here's three rough policies we could follow:

1. **Don't care about the rest of the ecosytem.** We could just bump MSRV as high as we want whenever it suits us, but I think that this risks annoying quite a number of users if we're always demanding the latest `rustc`.
2. **Try to keep some level of backwards compatibility.** If we take latest RHEL as an example, supporting this would mean we're ~3 months or so behind latest `rustc`. I think this allows us to get new features when we need them, while also giving users time to upgrade.
3. **Support a wide ecosystem.** Debian's `rustc` is already over a year old, and looks like it will probably be about 2.5 years old by the time the next Debian version is released. Supporting Debian would probably support almost all Rust users, but we would be very constrained with regards to updating PyO3 to leverage new language features and packages.

Overall, it seems to me like (2) is a good compromise.

programmerjake:
Agreed. Nearly anyone doing serious Rust dev will have a recent version of rustc.

birkenfeld:
Well, as a developer, of course. When I deploy stuff, I try using the toolchains available on those systems.

But for now I agree with (2), as long as Rust's iteration cycle and pace of development is still pretty high it makes sense not to drag too much.

davidhewitt:
[...]

PR #1128, opened 2020-08-30
Add support for arbitrary arrays

c410-f3r:
Extends my previous work on #778 and an additional PR will soon be provided to remove the `Default` bound restriction.

If https://github.com/rust-lang/rust/pull/75644 is going to be accepted, then the auxiliary functions won't be necessary in the near future.

### DIFF
--- a/src/conversions/array.rs
+++ b/src/conversions/array.rs
@@ -1,0 +1,273 @@
+use crate::{FromPyObject, IntoPy, PyAny, PyObject, PyResult, PyTryFrom, Python, ToPyObject};
+
+#[cfg(not(min_const_generics))]
+macro_rules! array_impls {
+    ($($N:expr),+) => {
+        $(
+            impl<'a, T> FromPyObject<'a> for [T; $N]
+            where
+                T: Copy + Default + FromPyObject<'a>,
+            {
+                #[cfg(not(feature = "nightly"))]
+                fn extract(obj: &'a PyAny) -> PyResult<Self> {
+                    let mut array = [T::default(); $N];
+                    _extract_sequence_into_slice(obj, &mut array)?;
+                    Ok(array)
+                }
+
+                #[cfg(feature = "nightly")]
+                default fn extract(obj: &'a PyAny) -> PyResult<Self> {
+                    let mut array = [T::default(); $N];
+                    _extract_sequence_into_slice(obj, &mut array)?;
+                    Ok(array)
+                }
+            }
+
+            #[cfg(feature = "nightly")]
+            impl<'source, T> FromPyObject<'source> for [T; $N]
+            where
+                for<'a> T: Default + FromPyObject<'a> + crate::buffer::Element,
+            {
+                fn extract(obj: &'source PyAny) -> PyResult<Self> {
+                    let mut array = [T::default(); $N];
+                    // first try buffer protocol
+                    if unsafe { crate::ffi::PyObject_CheckBuffer(obj.as_ptr()) } == 1 {
+                        if let Ok(buf) = crate::buffer::PyBuffer::get(obj) {
+                            if buf.dimensions() == 1 && buf.copy_to_slice(obj.py(), &mut array).is_ok() {
+                                buf.release(obj.py());
+                                return Ok(array);
+                            }
+                            buf.release(obj.py());
+                        }
+                    }
+                    // fall back to sequence protocol
+                    _extract_sequence_into_slice(obj, &mut array)?;
+                    Ok(array)
+                }
+            }
+        )+
+    }
+}
+
+#[cfg(not(min_const_generics))]
+array_impls!(
+    0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25,
+    26, 27, 28, 29, 30, 31, 32
+);
+
+#[cfg(min_const_generics)]
+impl<'a, T, const N: usize> FromPyObject<'a> for [T; N]
+where
+    T: FromPyObject<'a>,
+{
+    #[cfg(not(feature = "nightly"))]
+    fn extract(obj: &'a PyAny) -> PyResult<Self> {
+        create_array_from_obj(obj)
+    }
+
+    #[cfg(feature = "nightly")]
+    default fn extract(obj: &'a PyAny) -> PyResult<Self> {
+        create_array_from_obj(obj)
+    }
+}
+
+#[cfg(not(min_const_generics))]
+macro_rules! array_impls {
+    ($($N:expr),+) => {
+        $(
+            impl<T> IntoPy<PyObject> for [T; $N]
+            where
+                T: ToPyObject
+            {
+                fn into_py(self, py: Python) -> PyObject {
+                    self.as_ref().to_object(py)
+                }
+            }
+        )+
+    }
+}
+
+#[cfg(not(min_const_generics))]
+array_impls!(
+    0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25,
+    26, 27, 28, 29, 30, 31, 32
+);
+
+#[cfg(min_const_generics)]
+impl<T, const N: usize> IntoPy<PyObject> for [T; N]
+where
+    T: ToPyObject,
+{
+    fn into_py(self, py: Python) -> PyObject {
+        self.as_ref().to_object(py)
+    }
+}
+
+#[cfg(all(min_const_generics, feature = "nightly"))]
+impl<'source, T, const N: usize> FromPyObject<'source> for [T; N]
+where
+    for<'a> T: FromPyObject<'a> + crate::buffer::Element,
+{
+    fn extract(obj: &'source PyAny) -> PyResult<Self> {
+        let mut array: core::mem::MaybeUninit<[T; N]> = core::mem::MaybeUninit::uninit();
+        // first try buffer protocol
+        if unsafe { crate::ffi::PyObject_CheckBuffer(obj.as_ptr()) } == 1 {
+            if let Ok(buf) = crate::buffer::PyBuffer::get(obj) {
+                if buf.dimensions() == 1 && buf.copy_to_slice(obj.py(), &mut array).is_ok() {
+                    buf.release(obj.py());
+                    // SAFETY: The array should be fully filled by `copy_to_slice`
+                    return Ok(unsafe { array.assume_init() });
+                }
+                buf.release(obj.py());
+            }
+        }
+        // fall back to sequence protocol
+        _extract_sequence_into_slice(obj, &mut array)?;
+        // SAFETY: The array should be fully filled by `_extract_sequence_into_slice`
+        Ok(unsafe { array.assume_init() })
+    }
+}
+
+#[cfg(min_const_generics)]
+fn create_array_from_obj<'s, T, const N: usize>(obj: &'s PyAny) -> PyResult<[T; N]>
+where
+    T: FromPyObject<'s>,
+{
+    let seq = <crate::types::PySequence as PyTryFrom>::try_from(obj)?;
+    let expected_len = seq.len()? as usize;
+    let mut counter = 0;
+    try_create_array(&mut counter, |idx| {
+        seq.get_item(idx as isize)
+            .map_err(|_| crate::utils::invalid_sequence_length(expected_len, idx + 1))?
+            .extract::<T>()
+    })
+}
+
+fn _extract_sequence_into_slice<'s, T>(obj: &'s PyAny, slice: &mut [T]) -> PyResult<()>
+where
+    T: FromPyObject<'s>,
+{
+    let seq = <crate::types::PySequence as PyTryFrom>::try_from(obj)?;
+    let expected_len = seq.len()? as usize;
+    if expected_len != slice.len() {
+        return Err(crate::utils::invalid_sequence_length(
+            expected_len,
+            slice.len(),
+        ));
+    }
+    for (value, item) in slice.iter_mut().zip(seq.iter()?) {
+        *value = item?.extract::<T>()?;
+    }
+    Ok(())
+}
+
+#[cfg(min_const_generics)]
+fn try_create_array<E, F, T, const N: usize>(counter: &mut usize, mut cb: F) -> Result<[T; N], E>
+where
+    F: FnMut(usize) -> Result<T, E>,
+{
+    // Helper to safely create arrays since the standard library doesn't
+    // provide one yet. Shouldn't be necessary in the future.
+    struct ArrayGuard<'a, T, const N: usize> {
+        dst: *mut T,
+        initialized: &'a mut usize,
+    }
+
+    impl<T, const N: usize> Drop for ArrayGuard<'_, T, N> {
+        fn drop(&mut self) {
+            debug_assert!(*self.initialized <= N);
+            let initialized_part = core::ptr::slice_from_raw_parts_mut(self.dst, *self.initialized);
+            unsafe {
+                core::ptr::drop_in_place(initialized_part);
+            }
+        }
+    }
+
+    let mut array: core::mem::MaybeUninit<[T; N]> = core::mem::MaybeUninit::uninit();
+    let guard: ArrayGuard<T, N> = ArrayGuard {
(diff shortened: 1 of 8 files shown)


kngwyu:
Thank you, but is there any strong reason we should have this `nightly` feature?
I think we can wait for stabilization.

davidhewitt:
Agreed in that I was hoping the goal was to remove the nightly feature. Maybe instead of providing this in the pyo3 core, we can extend the new `FromPyObject` derive to be able to plug in custom converters similar to `#[serde(deserialize_with = ...)]`, and then the array support could be provided in a separate crate?

birkenfeld:
TBH I'd rather this to be kept behind a feature flag for longer, than get merged without the second that const-generics are stabilized (#1076 calling).

davidhewitt:
Agreed whatever we do we should give time for MSRV support to propagate.

c410-f3r:
I did include it within the `nightly` feature because... you know... it is currently only available in the nightly version of Rustc.

Constant generics is now behind its own opt-in flag and all `Default` and `Copy` bounds were removed

src/types/sequence.rs
```diff
@@ -257,6 +257,41 @@ impl PySequence {
...
+    fn extract(obj: &'source PyAny) -> PyResult<Self> {
+        let mut array = create_array_from_obj(obj)?;
+        if let Ok(buf) = crate::buffer::PyBuffer::get(obj) {
+            if buf.dimensions() == 1 && buf.copy_to_slice(obj.py(), &mut array).is_ok() {
+                buf.release(obj.py());
+                return Ok(array);
+            }
+            buf.release(obj.py());
+        }
+        Ok(array)
```
c410-f3r, line 290: This is kind of tricky. `copy_to_slice` expects a slice of initialized memory and that is why I inverted the order of the operations. I.e.: From `copy_to_slice` -> `create_array_from_obj` to `create_array_from_obj` -> `copy_to_slice`

Maybe this can be improved using the `get_item` method of Buffer.

kngwyu:
> I did include it within the nightly feature because... you know... it is currently only available in the nightly version of Rustc.

What I meant as `feature` is what this PR does and not the feature flag itself. My intention was keeping the code base reasonably small.

Anyway, let's consider merging after the upstream PR is merged.

kngwyu:
Ah, and I realized that now we don't have any tests for `nightly` feature.

davidhewitt:
> Ah, and I realized that now we don't have any tests for nightly feature.

As the `nightly` feature is just adding optimizations, there's not much testing to be added other than an extra CI job which runs with the flag enabled.

I'd prefer to move away from the pyo3 core crate having any CI which requires nightly, as it consumes contributor time (e.g. PRs spuriously failing because something changed in nightly). Hence my question above whether we can make this easy to integrate via a separate crate for now.

davidhewitt:
I have been thinking further on this topic, and I wanted to propose that we do support this PR, behind a new feature. I think it's beneficial to the Rust ecosystem that we support this because it enables more users to experiment with const generics. This can help the feature to eventually stabilize and also benefits immediately our users on nightly Rust.

(Even serde has an `unstable` feature for things like this: https://github.com/serde-rs/serde/blob/e3d871ff7bf10dadf10bdc234a55692228358d0e/serde/Cargo.toml#L45)

I'd prefer to keep away from having a single `nightly` feature, because I want users to know what they're opting in to and because I want to name it clearly that it's not guaranteed to be supported forever.

So I propose to move forward on this PR with a new feature named `unstable-const-generics`.

c410-f3r:
@davidhewitt Maybe wait for the 1.50 release?

https://github.com/rust-lang/rust/pull/79135

c410-f3r:
The time has come.
Are you guys still using the latest stable compiler on master? Maybe I should put back the `cons-generics` feature flag.

davidhewitt:
We support all the way back to Rust 1.41, so this will need to go behind a `const-generics` feature flag, yes.

davidhewitt:
An alternative to a user-facing feature flag is we could detect the compiler version in the build script, and then conditionally enable const generics based on that.

c410-f3r:
> We support all the way back to Rust 1.41, so this will need to go behind a const-generics feature flag, yes.

I miss the good ol' times when nightly was the only thing available for this project

> An alternative to a user-facing feature flag is we could detect the compiler version in the build script, and then conditionally enable const generics based on that.

Sure, this is a good approach. I will update everything accordingly. Thanks!

davidhewitt:
👍 you might be interested in `anyhow`'s `build.rs` which enables backtrace support in this fashion. We might be able to follow a similar pattern here.

https://github.com/dtolnay/anyhow/blob/master/build.rs

c410-f3r:
Clippy failure is not related to this PR

davidhewitt (changes requested):
Thanks very much for updating this PR. I verified locally that the `build.rs` feature detection is working - nice solution! There's still a couple of changes I would like to see before this is merged:
- As there's now a fair amount of array code sprinkled around the place, could this all please be moved into a new file `conversions/array.rs`. 
- The buffer specialization needs reworking, as at the moment the new implementation is a de-optimisation compared to the base case. See my other comment.

c410-f3r:
Hey @davidhewitt 
Thanks for the review! I will address everything in the next couple of days

src/conversions/array.rs
```diff
@@ -0,0 +1,72 @@
...
+// provide one yet. Shouldn't be necessary in the future.
+#[cfg(min_const_generics)]
+struct ArrayGuard<T, const N: usize> {
+    dst: *mut T,
+    initialized: usize,
+}
+
+#[cfg(min_const_generics)]
+impl<T, const N: usize> Drop for ArrayGuard<T, N> {
+    fn drop(&mut self) {
```
davidhewitt, line 13: Based on the coverage, it looks like this code never gets exercised.

As it's pretty fiddly `unsafe` code, we should add a test for it. I guess we could call `try_create_array` with a function that generates, and confirm that the drop code modified some counter as expected.

src/conversions/array.rs
```diff
@@ -0,0 +1,72 @@
...
+#[cfg(any(all(min_const_generics, feature = "nightly"), not(min_const_generics)))]
+pub fn extract_sequence_into_slice<'s, T>(obj: &'s PyAny, slice: &mut [T]) -> PyResult<()>
+where
+    T: FromPyObject<'s>,
+{
+    let seq = <PySequence as PyTryFrom>::try_from(obj)?;
+    if seq.len()? as usize != slice.len() {
+        return Err(exceptions::PyBufferError::new_err(
+            "Slice length does not match buffer length.",
+        ));
```
davidhewitt, line 46: As this particular error is now duplicated in a lot of places, I would like to see this refactored into a standalone function - e.g. `invalid_sequence_length()`.

For bonus points, it would be great to improve the error at the same time:
- it's not involving the Python buffer protocol when casting sequences to arrays in this way, so `ValueError` & `sequence` seems more appropriate terminology
- add size information which would help the user

e.g. maybe this:

```
fn invalid_sequence_length(expected: usize, actual: usize) -> PyErr {
    exceptions::PyValueError::new_err(
        format!("expected a sequence of length {} (got {})", expected, actual)
    )
}
```

Happy for you to go with something different if you think alternative wording is clearer.

src/conversions/mod.rs
```diff
@@ -1,5 +1,6 @@
 //! This module contains conversions between non-String Rust object and their string representation
 //! in Python
 
+pub mod array;
```
davidhewitt, line 4: I think there is no need for this to be `pub` if we move all code which is concerned with arrays into it.

src/types/sequence.rs
```diff
@@ -297,19 +297,61 @@ macro_rules! array_impls {
...
+        crate::conversions::array::create_array_from_obj(obj)
+    }
+}
+
+#[cfg(all(min_const_generics, feature = "nightly"))]
+impl<'source, T, const N: usize> FromPyObject<'source> for [T; N]
+where
+    for<'a> T: FromPyObject<'a> + crate::buffer::Element,
+{
+    fn extract(obj: &'source PyAny) -> PyResult<Self> {
```
davidhewitt, line 335: 👍 this is great; thanks for fixing!

davidhewitt (commented):
Thanks very much for the continued refinements! This is looking great.

Afraid I have another round of comments; I think we're really close now, very excited to see this land! 😅  

As well as the comments below, I have two additional points:
  - When I suggested moving code into `conversions/array.rs`, I meant all of the existing `array_impls!` macros and tests too. If it's all together in one place it's much easier for reviewers to audit what's tested and also maintain in the future once we don't need to support older rustc < 1.51.
  - This still needs a CHANGELOG entry. I propose the following:
    ```
    ### Added
    - Add conversion for `[T; N]` for all `N` on Rust 1.51 and up. [#1128](https://github.com/PyO3/pyo3/pull/1128)
    ```
    
    Happy for you to choose alternative wording if you think there's a better phrase out there.

src/conversions/array.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+#[cfg(min_const_generics)]
+impl<T, const N: usize> Drop for ArrayGuard<'_, T, N> {
+    fn drop(&mut self) {
+        debug_assert!(*self.initialized <= N);
+        let initialized_part = core::ptr::slice_from_raw_parts_mut(self.dst, *self.initialized);
+        unsafe {
+            core::ptr::drop_in_place(initialized_part);
+        }
```
kngwyu, line 146: Why do we need to `drop` the initialized slice?

c410-f3r: Guard drops the initialized part when the closure panics to prevent leaks

c410-f3r: This data structure was used by previous releases of the nightly compiler and is battle tested by many people around the globe. Please consider avoid changing its original formulation.

src/conversions/array.rs
```diff
@@ -0,0 +1,273 @@
...
+        thread::sleep,
+        time,
+    };
+
+    #[cfg(min_const_generics)]
+    #[test]
+    fn try_create_array() {
+        #[allow(clippy::mutex_atomic)]
+        let counter = Arc::new(Mutex::new(0));
+        let counter_unwind = Arc::clone(&counter);
```
davidhewitt, line 217: Because this test panics you'll need to disable it on the coverage CI job (which has panic=abort) by inserting this at the top of the test:

```suggestion
        // TODO replace with #[cfg(panic = "unwind")] once stable
        if !crate::cfg_panic_unwind() {
            return;
        }
        
        #[allow(clippy::mutex_atomic)]
        let counter = Arc::new(Mutex::new(0));
        let counter_unwind = Arc::clone(&counter);
```

src/conversions/array.rs
```diff
@@ -0,0 +1,273 @@
...
+        let _ = catch_unwind_silent(move || {
+            let mut locked = counter_unwind.lock().unwrap();
+            let _: Result<[i32; 4], _> = super::try_create_array(&mut *locked, |idx| {
+                if idx == 2 {
+                    panic!("peek a boo");
+                }
+                Ok::<_, ()>(1)
+            });
```
davidhewitt, line 225: Rather than counting the number of invocations of the callback, I wanted the counter to show how many times Drop has been called.


```suggestion
            static drop_counter: AtomicUsize = AtomicUsize::new(0);
            struct CountDrop { }
            impl Drop for CountDrop {
                fn drop(&mut self) {
                    drop_counter.fetch_add(1, Ordering::SeqCst);
                }
            }
            let ctr: usize = 0;
            let _: Result<[CountDrop; 4], _> = super::try_create_array(&mut ctr, |idx| {
                if idx == 2 {
                    panic!("peek a boo");
                }
                Ok::<_, ()>(CountDrop { })
            });
```

src/conversions/array.rs
```diff
@@ -0,0 +1,273 @@
...
+        }
+    }
+
+    let mut array: core::mem::MaybeUninit<[T; N]> = core::mem::MaybeUninit::uninit();
+    let guard: ArrayGuard<T, N> = ArrayGuard {
+        dst: array.as_mut_ptr() as _,
+        initialized: counter,
+    };
+    unsafe {
+        for (idx, value_ptr) in (&mut *array.as_mut_ptr()).iter_mut().enumerate() {
```
davidhewitt, line 192: I could be wrong; isn't this UB?

We're creating `value_ptr: &mut T` which refers to uninitialized memory; I'm reasonably certain that this breaks Rust's semantics of references and so the compiler is allowed to break this code at will.

I think the easiest way to avoid this could be to use `[MaybeUninit<T>; N]`.

src/buffer.rs
```diff
@@ -441,9 +442,7 @@ impl<T: Element> PyBuffer<T> {
 
     fn copy_to_slice_impl(&self, py: Python, target: &mut [T], fort: u8) -> PyResult<()> {
         if mem::size_of_val(target) != self.len_bytes() {
-            return Err(exceptions::PyBufferError::new_err(
-                "Slice length does not match buffer length.",
-            ));
+            return Err(invalid_sequence_length(self.item_count(), target.len()));
```
davidhewitt, line 445: I'm sorry for communicating poorly earlier; the two uses of `invalid_sequence_length` were actually correct to stay as `BufferError`.

I'll push a separate PR to improve these error messages; in this PR can you please just reset them to what they were (ultimately leaving `buffer.rs` unchanged by this PR, to avoid a merge conflict).

src/utils.rs
```diff
@@ -0,0 +1,8 @@
+use crate::{exceptions, PyErr};
```
davidhewitt, line 1: I'm personally not a fan of "utils" files if we can have code in more precisely named files.

In this case, I think once the `buffer.rs` changes are reverted then this function is only used inside `conversions/array.rs`, so can be moved there.

src/conversions/array.rs
```diff
@@ -0,0 +1,273 @@
...
+        create_array_from_obj(obj)
+    }
+
+    #[cfg(feature = "nightly")]
+    default fn extract(obj: &'a PyAny) -> PyResult<Self> {
+        create_array_from_obj(obj)
+    }
+}
+
+#[cfg(not(min_const_generics))]
```
davidhewitt, line 74: Now that everything's in one file, I think it would be nice if this macro was merged with the other `array_impls!` macro just above.

davidhewitt (commented):
Whew, this PR is turning out to be quite fiddly! Thank you for all the effort you've put into this.

I'm apologetic to say I've got another round of comments. In particular I have a concern that we might have some UB inside the `try_create_array` function.

src/conversions/array.rs
davidhewitt: @kngwyu note that eventually an API for this might be added to `std`: https://github.com/rust-lang/rust/pull/75644

Personally I'm happy as long as we're sure the implementation we use is sound.

davidhewitt:
@c410-f3r thanks very much for all the work you've done getting this PR through to this point.

I'm doing some tidying up of pending items ahead of the 0.14 release so I'm shortly going to rebase this PR and push a commit to finish it off so that we can include it in the upcoming release. Hope that's ok.

c410-f3r:
Hey @davidhewitt Thanks for finishing this PR!

davidhewitt (approved):

src/conversions/array.rs
```diff
@@ -0,0 +1,282 @@
...
+                Ok(CountDrop)
+            });
+        });
+        assert_eq!(DROP_COUNTER.load(Ordering::SeqCst), 2);
+    }
+
+    #[test]
+    fn test_extract_small_bytearray_to_array() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
```
kngwyu, line 230: Lets's use `with_gil` in new tests

src/conversions/array.rs
```diff
@@ -0,0 +1,282 @@
...
+    use crate::{PyResult, Python};
+    #[cfg(min_const_generics)]
+    use std::{
+        panic,
+        sync::atomic::{AtomicUsize, Ordering},
+    };
+
+    #[cfg(min_const_generics)]
+    #[test]
+    fn array_try_from_fn() {
```
kngwyu, line 208: How about moving these to `#[cfg(min_const_generics)]` block?

kngwyu (changes requested):
Thanks! I have some minor requests for tests

src/conversions/array.rs
davidhewitt: Agreed; actually I'll rearrange the whole file so there's only one `#[cfg(min_const_generics)]` on an inner `mod`.

PR #1129, opened 2020-08-31
Use CPython stable APIs for implementing tuples.

alex:
Refs #1125

Reduces the number of compilation failures with `--features abi3` from 56 to 53.

I chose to implement this by simple replacing all functions with their stable equivilant. This potentially regresses performance somewhat. If it's preferable, I can switch which call is used based on the `#[cfg]` at the price of some verbosity.

### DIFF
--- a/src/types/tuple.rs
+++ b/src/types/tuple.rs
@@ -42,7 +42,7 @@ impl PyTuple {
     pub fn len(&self) -> usize {
         unsafe {
             // non-negative Py_ssize_t should always fit into Rust uint
-            ffi::PyTuple_GET_SIZE(self.as_ptr()) as usize
+            ffi::PyTuple_Size(self.as_ptr()) as usize
         }
     }
 
@@ -62,8 +62,7 @@ impl PyTuple {
     /// Takes a slice of the tuple from `low` to the end and returns it as a new tuple.
     pub fn split_from(&self, low: isize) -> &PyTuple {
         unsafe {
-            let ptr =
-                ffi::PyTuple_GetSlice(self.as_ptr(), low, ffi::PyTuple_GET_SIZE(self.as_ptr()));
+            let ptr = ffi::PyTuple_GetSlice(self.as_ptr(), low, self.len() as Py_ssize_t);
             self.py().from_owned_ptr(ptr)
         }
     }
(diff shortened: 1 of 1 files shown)


kngwyu:
How about making a single specific branch for this instead of opening series of PRs?
If you are OK, I will create a separate branch (e.g., `pyo3/abi3`) and you can send a PR to that branch.

alex:
I personally find it easier to work in small PRs that can each be reviewed by themselves. This ensures that I'm constantly working off the latest version and didn't break anything.

The next PR is probably going to be type definition support, which will be the biggest part of this anyways :-)

alex:
Coverage change is because `as_slice()` lost coverage since it's no longer used in `iter()`. Let me write a test for it.

kngwyu:
> The next PR is probably going to be type definition support, which will be the biggest part of this anyways :-)

That patch would require careful design decisions and it's easier to work for me if we have an upstream branch for it.

alex:
Yes, I agree that PR will be very subtle and require careful thought! Is it ok to review/merge this tuple PR as is, and then we can do a branch for the type definition work?

src/types/tuple.rs
```diff
@@ -110,8 +112,8 @@ impl<'a> Iterator for PyTupleIterator<'a> {
 
     #[inline]
     fn next(&mut self) -> Option<&'a PyAny> {
-        if self.index < self.slice.len() {
-            let item = self.slice[self.index];
+        if self.index < self.tuple.len() {
```
kngwyu, line 115: Just a nit, but I think it would be better to have `length` in `PyTupleIterator`.
`slice.len()` is zero-cost, but `tuple.len()` is not because of FFI boundary.

kngwyu (approved):
Thanks, mostly LGTM.

src/types/tuple.rs
alex: Will do!

kngwyu:
> Is it ok to review/merge this tuple PR as is, and then we can do a branch for the type definition work?

Of course. I'll create `pyo3/abi3` branch and please open a PR to that branch first. Then let's work on the branch collaboratory.

alex:
Ok, should be good now!

kngwyu:
Now, https://github.com/PyO3/pyo3/tree/abi3 is available.

PR #1131, opened 2020-08-31
structure code to avoid a use-after-free

alex:
I don't understand why the existing code works -- I'd expect the `CString` value to be `Drop`'d before `PyDict_SetItemString` was called, but it doesn't appear to work that way. Structuring the code like this makes it clear the lifetimes are safe.

### DIFF
--- a/src/type_object.rs
+++ b/src/type_object.rs
@@ -235,7 +235,8 @@ fn initialize_tp_dict(
     // the POV of other threads.
     for (key, val) in items {
         let ret = unsafe {
-            ffi::PyDict_SetItemString(tp_dict, CString::new(key)?.as_ptr(), val.into_ptr())
+            let key_cstr = CString::new(key)?;
+            ffi::PyDict_SetItemString(tp_dict, key_cstr.as_ptr(), val.into_ptr())
         };
         if ret < 0 {
             return Err(PyErr::fetch(py));


alex:
CI failures are unrelated.

The code currently works due to https://doc.rust-lang.org/nightly/reference/expressions.html#temporaries but I think writing it like this makes it much more obvious why it's safe.

kngwyu:
This is a frequent pattern and I'm not sure we should avoid this.

alex:
I think with `as_ptr()` not being called on a temporary, it's much clearer that the code is safe, but if you disagree I can close this.

birkenfeld:
Actually, there is a much better solution: don't call `PyDict_SetItemString` at all. It's a convenience API for when you have a C string already (but we don't).

The first thing it does is to make a string object with `PyUnicode_FromString`, so the only thing this gets us is an unnecessary allocation for the CString adding the NUL and an unnecessary traversal on Python's side to find the NUL again.

We should rather just create the string object here using `PyUnicode_FromStringAndSize`.

birkenfeld:
I checked all `CString::new` instances, and there are two others where we can use a different API instead of creating a `CString`: `PyImport_ImportModule => PyImport_Import` and `PyModule_New => PyModule_NewObject`.  All others either create leaked strings for structures, or are used to call APIs that have no equivalent taking `PyObject*`.

Also, in all other places where the CString is used with `as_ptr()`, it is `let`-bound before the C function is called.

alex:
I'm going to close this PR then. Thanks Georg!

PR #1133, opened 2020-09-01
Avoid using CString where unnecessary

birkenfeld:
Use APIs that take a Python string instead of those taking a C string,
which are a convenience in C but require an unncessary allocation
for us.

### DIFF
--- a/src/ffi/import.rs
+++ b/src/ffi/import.rs
@@ -63,6 +63,7 @@ pub unsafe fn PyImport_ImportModuleEx(
 
 extern "C" {
     pub fn PyImport_GetImporter(path: *mut PyObject) -> *mut PyObject;
+    #[cfg_attr(PyPy, link_name = "PyPyImport_Import")]
     pub fn PyImport_Import(name: *mut PyObject) -> *mut PyObject;
     #[cfg_attr(PyPy, link_name = "PyPyImport_ReloadModule")]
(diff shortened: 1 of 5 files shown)


kngwyu:
Thanks!
The CI failure in PyPy looks weird... :thinking: 
Maybe we need `#[cfg_attr(PyPy, link_name = "PyPyImport_Import")]`?

birkenfeld:
Looks like the rustup CDN is currently flaky...

davidhewitt:
Latest failure looks specific to pypy on windows, symbol not included only on that build? Perhaps worth posting an issue upstream?

kngwyu:
I confirmed that ` error LNK2019: unresolved external symbol _PyPyModule_NewObject referenced ` is the problem and will some investigation tomorrow.

kngwyu:
It looks like that pypy doesn't have `PyModule_NewObject`: https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/cpyext/modsupport.py

birkenfeld:
Is that a PyPy bug, or should I remove that one from here?

kngwyu:
> Is that a PyPy bug, or should I remove that one from here?

I'm not sure it's a bug or intended, but at least we should use `#[cfg(PyPy)]` if we want to use `PyModule_NewObject`.

birkenfeld:
well, since it's only an optimization I removed that one.

kngwyu (approved):
Thanks!

PR #1135, opened 2020-09-01
Remove duplicate pool and py in py_init.

sebpuetz:
`callback_body!` already creates a pool and acquires `Python`:

~~~Rust
let pool = ::pyo3::GILPool::new();
let unwind_safe_py = std::panic::AssertUnwindSafe(pool.python());
~~~

I think these two lines can be removed safely, at least all tests still pass locally!


davidhewitt (approved):
Agreed, thanks for cleaning this up!

PR #1136, opened 2020-09-01
Remove proc-macro panics.

sebpuetz:
In `pyo3-derive-backend`:

Previously:
~~~sh
$ grep "panic" ./**/*.rs -n
./src/method.rs:335:                                panic!("argument type is not supported by python method: {:?} ({:?}) {:?}",
./src/method.rs:346:                            panic!(
./src/module.rs:100:                        _ => panic!("The first parameter of pyfn must be a MetaItem"),
./src/module.rs:107:                        _ => panic!("The second parameter of pyfn must be a Literal"),
./src/module.rs:116:                    panic!("can not parse 'pyfn' params {:?}", attr);
./src/proto_method.rs:347:        ty => panic!("Unsupported argument type: {:?}", ty),
./src/proto_method.rs:388:        _ => panic!(),
./src/proto_method.rs:414:        _ => panic!("not supported"),
./src/proto_method.rs:438:        panic!("func.rs::296")
~~~

Now:
~~~sh
$ grep "panic" ./**/*.rs -n
./src/proto_method.rs:394:        _ => panic!(),
~~~

This should improve errors with incorrect macro inputs. Not sure whether the remaining panic could / should also be changed to returning `Result`:

~~~Rust
fn extract_decl(spec: syn::Item) -> syn::Signature {
    match spec {
        syn::Item::Fn(f) => f.sig,
        _ => panic!(),
    }
}
~~~

### DIFF
--- a/pyo3-derive-backend/src/method.rs
+++ b/pyo3-derive-backend/src/method.rs
@@ -224,7 +224,7 @@ impl<'a> FnSpec<'a> {
                         ty,
                         optional: opt,
                         py,
-                        reference: is_ref(name, ty),
+                        reference: is_ref(name, ty)?,
                     });
                 }
             }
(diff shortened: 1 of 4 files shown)


sebpuetz:
CI failures are due to coverage, does this need to be fixed?

davidhewitt (approved):
I think for new features it's well worth adding tests, but for this I'm very happy as-is.

PR #1137, opened 2020-09-01
Refactor proc_macro optional / python detection

davidhewitt:
Some tidy ups to reduce some proc macro code I noticed could be simplified.


kngwyu (approved):
Good catch, thank you!
Looks reasonable :+1:

PR #1138, opened 2020-09-01
Disable fail-fast in CI test

davidhewitt:
Switching this off allows all platforms to continue testing, rather than stop when the first one fails.

When there's spurious failures, this allows the process to continue to (hopefully) mean fewer rounds of iteration / re-running jobs needed overall.


kngwyu (approved):
Thanks!

PR #1141, opened 2020-09-01
Start splitting long guide pages into subchapters

davidhewitt:
I've been meaning to start breaking the longest pages into subchapters for a little while.

I've avoided making any changes to the wording. `git`'s pretty bad at rebasing changes when files are split like this, so it would be nice to try to merge this quickly if you agree this is a good idea. We can then improve the contents of each page / maybe shuffle things around more.


kngwyu:
Looks nice and please continue working.

davidhewitt:
👍 I'll merge this one now so that it doesn't create conflicts with work other contributors are doing. Somewhere I have a few notes of other small tweaks I want to push to the docs...

Issue #759, opened 2020-02-08
ModuleNotFoundError: No module named 'supermodule.submodule'; 'supermodule' is not a package


tcztzy:
## 🐛 Bug Reports
I am working for writing Python extension, but facing import submodule error.
```
ModuleNotFoundError: No module named 'supermodule.submodule'; 'supermodule' is not a package
```

### What do I expected
Import submodule like usual Python package.

### 🌍 Environment

 - Your operating system and version:
[ArchWSL](https://github.com/yuk7/ArchWSL) version 19.11.16.0 on [Windows Subsystem Linux 2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-index) version Windows Insiders Slow ring 10.0.19041
```
$ uname -a
Linux xxxxxx 4.19.84-microsoft-standard #1 SMP Wed Nov 13 11:44:37 UTC 2019 x86_64 GNU/Linux
```
 - Your python version:
```shell
$ python -V
Python 3.8.1
```
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?:
```
$ pyenv virtualenv system xxxxxx
```
 - Your rust version (`rustc --version`):
```
rustc 1.43.0-nightly (c9290dcee 2020-02-04)
```
 - Are you using the latest pyo3 version? Have you tried using latest master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?` 
Yes.

### 💥 Reproducing

Please provide a [minimal working example](https://stackoverflow.com/help/mcve). This means both the rust code and the python.

Please also write what exact flags are required to reproduce your results.
*src/lib.rs*
```rust
use pyo3::prelude::*;
use pyo3::{wrap_pyfunction, wrap_pymodule};
use pyo3::types::IntoPyDict;

#[pyfunction]
fn subfunction() -> String {
    "Subfunction".to_string()
}

#[pymodule]
fn submodule(_py: Python, module: &PyModule) -> PyResult<()> {
    module.add_wrapped(wrap_pyfunction!(subfunction))?;
    Ok(())
}

#[pymodule]
fn supermodule(_py: Python, module: &PyModule) -> PyResult<()> {
    module.add_wrapped(wrap_pymodule!(submodule))?;
    Ok(())
}
```
*Cargo.toml*
```toml
[package]
name = "supermodule"
version = "0.1.0"
authors = ["Tang Ziya <tcztzy@gmail.com>"]
edition = "2018"

[...]

Issue #694, opened 2019-12-18
Unify `#[pyfunction]` and `#[pyfn]`


davidhewitt:
It seems confusing to me to have two different attributes, `#[pyfunction]` and `#[pyfn]`, which both wrap Rust functions to create python functions. Pyo3 users need to know which applies in two different contexts.

I'd argue we should merge these. Then there's only one attribute which pyo3 users need to know about. It would also help us to simplify / refactor the macro code if we didn't need two different parsing systems.

I think this would need to go in the following steps:
1. Allow `#[pyfunction]` to be used in place of `#[pyfn]`
2. Change the docs to describe `#[pyfunction]` everywhere
3. Change `#[pyfn]` to expand to `#[pyfunction]`
3. (Later on) change `#[pyfn]` to emit an error, stating the `#[pyfunction]` invocation to use instead
4. (Much later on) remove `#[pyfn]`

I've already opened a PR #693 to achieve step 1.

kngwyu:
But... are they so confusing?
For me, the different looks clear from the [guide](https://pyo3.rs/v0.8.4/function.html).

Or am I too accustomed?

davidhewitt:
The guide is clear enough on the differences. It's only a little confusing - but mostly I think for new users of pyo3.

Another downside to having two different attributes is that they might drift apart over time if we're not careful. For example, if #692 lands, then the way that each of these specify the python name is different:

    #[pyfn(m, "foo")]
    fn bar() { }

vs

    #[pyfunction]
    #[name = "foo"]
    fn bar() { }

It seems like a good idea to me to simplify the codebase by unifying these attributes.

pganssle:
@kngwyu I would not say they are confusing if you read the guide, but I agree with @davidhewitt - these two procedural macros apparently do the *same thing* but in two different contexts.

Consider that even if the documentation clarifies this, it's definitely going to be hard to remember which one to use in which context unless you do it basically every day, so you'll have to look it up every time. Readers of the code who aren't necessarily reading the documentation will wonder why one or the other is used, etc.

If it's possible to unify them into a single macro, IMO, it should definitely be done.

kngwyu:
I'd rather introduce
```rust
#[pymodule] 
mod mymod {
    #[pyfunction] 
    fn myfunction(py: Python, ...) {}
}
``` 
and make `pyfn` soft-duplicated.

davidhewitt:
[...]

PR #1143, opened 2020-09-02
PyModule in #[pyfunction]

### LINKED ISSUE #828: Accessing the module object in `pyfunction`s
In the C API, module-level functions get passed their registered module object as the `self` parameter.  This is very nice to interact with other module level API.

Can we (optionally?) make the module reference available in `pyfunction`s too?

sebpuetz:
Closes #828 

- [x] Error handling for module names
- [x] how to signal when `PyModule` should be passed
- [x] Documentation


pyo3-derive-backend/src/module.rs
```diff
@@ -191,9 +192,25 @@ pub fn add_fn_to_module(
...
+        name = quote!(
+            let name = unsafe { pyo3::ffi::PyModule_GetNameObject(mod_ptr) };
+            if name.is_null() {
+                //TODO Error handling
+            }
+        );
+    }
     Ok(quote! {
-        fn #function_wrapper_ident(py: pyo3::Python) -> pyo3::PyObject {
+        fn #function_wrapper_ident(py: pyo3::Python #maybe_module_arg) -> pyo3::PyObject {
```
davidhewitt, line 213: How about here, instead of requiring the module arg optionally depending on the `free` attribute, we add a little trait to `derive_utils`, e.g.

```rust
trait WrapPyFunctionArguments<'a> {
    fn arguments(self) -> (Python<'a>, Option<&'a PyModule>);
}

impl<'a> WrapPyFunctionArguments<'a> for Python<'a> {
    fn arguments(self) -> (Python<'a>, Option<&'a PyModule>>) { (self, None) }
}

impl<'a> WrapPyFunctionArguments<'a> for &'a PyModule {
    fn arguments(self) -> (Python<'a>, Option<&'a PyModule>>) { (self.py(), Some(self)) }
}
```

Then it should be possible to call either `wrap_pyfunction!(foo)(py)` or `wrap_pyfunction!(foo)(module)` if we change this line to:

```suggestion
        fn #function_wrapper_ident(args: impl pyo3::derive_utils::WrapPyFunctionArguments) -> pyo3::PyObject {
```

And in this generated wrapper code here we could do different things depending on whether a `PyModule` is provided.

davidhewitt (commented):
Thanks, I have a few quick thoughts on this:

- 👍 on adding `add_function`, and `add_module`, though I would like to also keep support for `add_wrapped` for now, because I don't see a strong reason that users would need to migrate existing code. I'm thinking in 0.12 we can add a doc to `add_wrapped` saying prefer the new methods, and in 0.13 we can mark as deprecated or consider a broader cleanup. (I still would love to have the new module syntax discussed in #694)

- Similarly, I think asking users to migrate to `#[pyfunction(free)]` and also changing the arguments to `wrap_pyfunction!` is quite a lot of churn to existing code. I agree you're probably right that most pyfunctions want to be part of modules, but I think we don't necessarily need to force users to change existing code immediately.

I've added an inline comment with a suggestion of how we might be able to achieve a similar result to what you've got, but without forcing ecosystem churn.

pyo3-derive-backend/src/module.rs
davidhewitt: EDIT: This idea _might not_ work because `wrap_pyfunction!` takes a reference to the wrapper, and by adding the `impl Trait` argument I don't think it can have a reference to it.

If that's the case, I'd still like us to look for a similar trick so that the new functionality comes with new syntax, without changing the existing code.

sebpuetz:
Thanks for the comments!
 
>  * Similarly, I think asking users to migrate to `#[pyfunction(free)]` and also changing the arguments to `wrap_pyfunction!` is quite a lot of churn to existing code. I agree you're probably right that most pyfunctions want to be part of modules, but I think we don't necessarily need to force users to change existing code immediately.

I don't think there are many uses of `pyfunction(free)` at all! It's fairly unnatural for a `pyfunction` to not be associated with a `PyModule`. Almost every case within the pyo3 codebase that needed to be fixed after this change was about tests where no module was set up and code was called from within Rust. So I don't think that many users would even notice this change!

I'll look into implementing your suggestions tomorrow!

davidhewitt:
> I don't think there are many uses of pyfunction(free) at all!

👍 I guess you're probably right about this! It's sometimes hard to judge how much certain patterns are used downstream.

My point still stands that if we can have our cake and eat it (i.e. add new feature without changing any existing code), that'd be the best thing =)

davidhewitt (commented):
:tada: looks like my suggestion worked!

sebpuetz:
This works pretty well, although figuring out the lifetimes was a bit of work. Nice suggestions!

Now we have to figure out a way to signal *which* `&PyModule` argument in the signature is the one that we should inject.

src/derive_utils.rs
```diff
@@ -207,3 +207,24 @@ where
         <R as std::convert::TryFrom<&'a PyCell<T>>>::try_from(cell)
     }
 }
+
+/// Trait to abstract over the arguments of Python function wrappers.
+#[doc(hidden)]
+pub trait WrapPyFunctionArguments<'a, 'b> {
```
davidhewitt, line 213: Is there a reason to have two lifetime arguments in this trait? If we set them to the same, then Rust can just shorten the longer lifetime to match, which seems good enough to me. 

---

Also, looking again at this trait, I noticed a different pattern to achieve the same thing:

```
pub enum WrapPyFunctionArguments<'a> {
    Python(Python<'a>),
    Module(&'a PyModule)
}
```

And can then just implement `Into<WrapPyFunctionArguments>` for each of `Python` and `&PyModule`.

I think that _maybe_ the enum is less confusing to understand in the long run, I'm also happy to leave it as-is (as long as we take the second lifetime argument out). Just an idea in cause you find you prefer one design over the other.

sebpuetz: > Is there a reason to have two lifetime arguments in this trait? If we set them to the same, then Rust can just shorten the 
> longer lifetime to match, which seems good enough to me.

That's weird, I was trying that for almost an hour, now I went back to verify that it in fact doesn't work...and...it compiles!

I'll check how nice the enum variant turns out

sebpuetz:
PyPy doesn't seem to have a `PyModule_GetNameObject` / `PyPyModule_GetNameObject` function, currently digging through PyPy source...

kngwyu:
@sebpuetz 
https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/cpyext/modsupport.py#L136
Looks it has only `PyModule_GetName`

pyo3-derive-backend/src/module.rs
```diff
@@ -193,7 +193,17 @@ pub fn add_fn_to_module(
     let wrapper = function_c_wrapper(&func.sig.ident, &spec);
 
     Ok(quote! {
-        fn #function_wrapper_ident(py: pyo3::Python) -> pyo3::PyObject {
+        fn #function_wrapper_ident<'a>(
+            args: impl Into<pyo3::derive_utils::WrapPyFunctionArguments<'a>>
+        ) -> pyo3::PyObject {
+            let arg = args.into();
+            let (py, maybe_module) = match arg {
```
kngwyu, line 200: How about making this a method of `WrapPyFunctionArguments`?

sebpuetz:
Yeah, I'm now going through `PyModule::name()` instead, I'll also open a PR that feature-gates the `GetNameObject`

kngwyu (approved):
Thanks, nicely designed :+1: 

BTW, how about changing the argument of `add_function` to `wrapper: &impl Fn(&'a Self) -> PyResult<PyObject>`?

sebpuetz:
> BTW, how about changing the argument of `add_function` to `wrapper: &impl Fn(&'a Self) -> PyResult<PyObject>`?

Looking into this now, there are also a few `expect()` calls in these `add` functions that should be replaced with `?`.

sebpuetz:
I don't see how that's possible while not breaking existing code: 

If `add_function` takes `wrapper: &impl Fn(&'a Self) -> PyResult<PyObject>`, we need to change the code-gen accordingly to return  `-> PyResult<PyObject>` wrappers.

If the proc-macro generates `-> PyResult<PyObject>` wrappers, the old `add_wrapped` way breaks, because the argument in `add_wrapped` is `wrapper: &impl Fn(&'a Self) -> PyObject`. 

Fwiw, I'm fine with making that breaking change and have *all* `add_X` fns on `PyModule` take wrappers that return results. With my proposed solution we'd see some weird behaviour, if `PyModule::name()` fails while creating the method, `PyErr` is converted to `PyObject` and returned. The error only gets recognized when `function.getattr(self.py(), "__name__")?;` fails - assuming that `PyErr` doesn't have a `__name__` attribute. 

So, should we change all add functions to take `-> PyResult<PyObject>` wrappers?

kngwyu:
> I don't see how that's possible while not breaking existing code:

If `add_wrapped(wrap_*!())` still works, I think it's not *very* breaking and OK.

sebpuetz:
Well, that doesn't work because the argument to `add_wrapped()` is required to be a function returning `PyResult<PyObject>`. I'll push how it'd look as an additional commit in a couple of minutes. We can then decide which version we want.

sebpuetz:
I think I found a solution by making the return value of the `wrapper` fn  `T: IntoPyCallback<PyObject>` in `add_wrapped`. I'm not getting compilation errors when using `add_wrapped(wrap_pyfunction!())` now, so I hope that this is a workable solution.

src/types/module.rs
```diff
@@ -194,11 +195,50 @@ impl PyModule {
...
+    pub fn add_wrapped<'a, T>(&'a self, wrapper: &impl Fn(Python<'a>) -> T) -> PyResult<()>
+    where
+        T: IntoPyCallbackOutput<PyObject>,
+    {
+        let function = wrapper(self.py()).convert(self.py())?;
+        let name = function.getattr(self.py(), "__name__")?;
+        self.add(name.extract(self.py())?, function)
+    }
+
+    /// Adds a (sub)module to a module.
```
davidhewitt, line 210: I think it's fine to just call it a submodule. (I see you got this wording from add_wrapped)

```suggestion
    /// Add a submodule to this module.
```

src/types/module.rs
```diff
@@ -194,11 +195,50 @@ impl PyModule {
...
+    /// ```rust,ignore
+    /// m.add_module(wrap_pymodule!(utils));
+    /// ```
+    pub fn add_module<'a>(&'a self, wrapper: &impl Fn(Python<'a>) -> PyObject) -> PyResult<()> {
+        let module = wrapper(self.py());
+        let name = module.getattr(self.py(), "__name__")?;
+        self.add(name.extract(self.py())?, module)
+    }
+
+    /// Adds a function to a module, using the functions __name__ as name.
```
davidhewitt, line 223: I think the wording can be identical as for `add_module`.

```suggestion
    /// Add a function to this module.
```

src/types/module.rs
```diff
@@ -194,11 +195,50 @@ impl PyModule {
...
+    }
+
+    /// Adds a (sub)module to a module.
+    ///
+    /// Use this together with `#[pymodule]` and [wrap_pymodule!].
+    ///
+    /// ```rust,ignore
+    /// m.add_module(wrap_pymodule!(utils));
+    /// ```
+    pub fn add_module<'a>(&'a self, wrapper: &impl Fn(Python<'a>) -> PyObject) -> PyResult<()> {
```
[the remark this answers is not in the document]
sebpuetz: Yep, that sounds better!

guide/src/function.md
```diff
@@ -189,3 +189,53 @@ If you have a static function, you can expose it with `#[pyfunction]` and use [`
...
+
+Functions are usually associated with modules, in the C-API, the self parameter in a function call corresponds
+to the module of the function. It is possible to access the module of a `#[pyfunction]` and `#[pyfn]` in the
+function body by passing the `need_module` argument to the attribute:
+
+```rust
+use pyo3::wrap_pyfunction;
+use pyo3::prelude::*;
+
+#[pyfunction(need_module)]
```
davidhewitt, line 203: Bikeshedding: I'd like to propose to call the attribute `pass_module`.

Motivation is that this was already the verb used in #828

Also I've seen this phrasing before e.g. in Python's `click` library: https://click.palletsprojects.com/en/7.x/api/#click.pass_context

guide/src/function.md
```diff
@@ -189,3 +189,53 @@ If you have a static function, you can expose it with `#[pyfunction]` and use [`
 [`PyAny::call1`]: https://docs.rs/pyo3/latest/pyo3/struct.PyAny.html#tymethod.call1
 [`PyObject`]: https://docs.rs/pyo3/latest/pyo3/type.PyObject.html
 [`wrap_pyfunction!`]: https://docs.rs/pyo3/latest/pyo3/macro.wrap_pyfunction.html
+
+### Accessing the module of a function
+
+Functions are usually associated with modules, in the C-API, the self parameter in a function call corresponds
```
davidhewitt, line 195: I think this first sentence is probably an implementation detail that the user doesn't need to know? Might want to take it out for simplicity.

guide/src/function.md
```diff
@@ -189,3 +189,53 @@ If you have a static function, you can expose it with `#[pyfunction]` and use [`
...
+
+#[pymodule]
+fn module_with_fn(py: Python, m: &PyModule) -> PyResult<()> {
+    m.add_function(wrap_pyfunction!(pyfunction_with_module))
+}
+
+# fn main() {}
+```
+
+If `need_module` is set, the first argument **must** be the `&PyModule`. It is then possible to interact with
```
davidhewitt, line 218: Maybe "it is then possible to use the module in the function body"?

guide/src/function.md
```diff
@@ -189,3 +189,53 @@ If you have a static function, you can expose it with `#[pyfunction]` and use [`
...
+    fn module_name(module: &PyModule) -> PyResult<&str> {
+        module.name()
+    }
+    Ok(())
+}
+
+# fn main() {}
+```
+
+Within Python, the name of the module that a function belongs to can be accessed through the `__module__`
```
davidhewitt, line 240: I think this is probably true of all Python functions (although you've just fixed this for pyO3) so not sure this sentence needs to be in this section?

davidhewitt (approved):
Thanks, this looks great to me🚀 

Great docs and tests much appreciated 😊

Couple of final suggestions from me.

davidhewitt:
Oh also, CHANGELOG entry please! :smile:

sebpuetz:
Looks pretty green to me ;)

guide/src/function.md
```diff
@@ -189,3 +189,49 @@ If you have a static function, you can expose it with `#[pyfunction]` and use [`
...
+
+It is possible to access the module of a `#[pyfunction]` and `#[pyfn]` in the
+function body by passing the `pass_module` argument to the attribute:
+
+```rust
+use pyo3::wrap_pyfunction;
+use pyo3::prelude::*;
+
+#[pyfunction(pass_module)]
+fn pyfunction_with_module(
```
birkenfeld, line 203: Does the signature need to be formatted so strangely here?

src/types/module.rs
```diff
@@ -194,11 +195,50 @@ impl PyModule {
...
+    /// Use this together with the`#[pyfunction]` and [wrap_pyfunction!].
+    ///
+    /// ```rust,ignore
+    /// m.add_function(wrap_pyfunction!(double));
+    /// ```
+    ///
+    /// You can also add a function with a custom name using [add](PyModule::add):
+    ///
+    /// ```rust,ignore
+    /// m.add("also_double", wrap_pyfunction!(double)(py, m));
```
birkenfeld, line 234: Does this need to handle the `Result` from wrap_pyfunction?

src/types/module.rs
```diff
@@ -194,11 +195,50 @@ impl PyModule {
...
-            .expect("A function or module must have a __name__");
-        self.add(name.extract(self.py()).unwrap(), function)
+    ///
+    /// **This function will be deprecated in the next release. Please use the specific
+    /// [add_function] and [add_submodule] functions instead.**
+    pub fn add_wrapped<'a, T>(&'a self, wrapper: &impl Fn(Python<'a>) -> T) -> PyResult<()>
+    where
+        T: IntoPyCallbackOutput<PyObject>,
+    {
+        let function = wrapper(self.py()).convert(self.py())?;
```
birkenfeld, line 205: might be nicer to do a `let py = self.py()` at the beginning?

tests/test_module.rs
```diff
@@ -60,8 +65,8 @@ fn module_with_functions(py: Python, m: &PyModule) -> PyResult<()> {
 
     m.add("foo", "bar").unwrap();
 
-    m.add_wrapped(wrap_pyfunction!(double)).unwrap();
-    m.add("also_double", wrap_pyfunction!(double)(py)).unwrap();
+    m.add_function(wrap_pyfunction!(double)).unwrap();
+    m.add("also_double", wrap_pyfunction!(double)(m)?).unwrap();
```
birkenfeld, line 69: now it takes only the `m` argument. What am I missing here?

src/types/module.rs
```diff
@@ -194,11 +195,50 @@ impl PyModule {
     /// ```rust,ignore
     /// m.add("also_double", wrap_pyfunction!(double)(py));
```
birkenfeld, line 196: arguments are `(py)` here, but `(py, m)` below?

src/types/module.rs
davidhewitt: Ah, good spot. This I think is an outdated doc and probably shows why using `ignore` in docstrings is usually a bad idea! Maybe this should be updated to not use ignore?

src/types/module.rs
sebpuetz: I missed these because the tests are ignored, I'll go through the guide again to make the `add_X` calls consistent. The correct way to add a function is to pass the `PyModule` to the wrapper. `Python` can be acquired from  `PyModule`.

davidhewitt: `(py, m)` was an outdated design we removed. `ignore` docstring example probably was a mistake.

`wrap_pyfunction!` output can take either `py` (for backwards compatibility, will deprecate eventually) or `m` (preferred, sets the `__module__` correctly).

src/types/module.rs
sebpuetz: Yup, simplifies all the add methods

src/types/module.rs
birkenfeld: OK, thanks!

src/types/module.rs
kngwyu: Yeah, nice catch. Looks we need to update this.

sebpuetz: I made all the examples in the docs  of `fn add_X` runnable.

sebpuetz:
@davidhewitt I pushed the changes discussed in #1149 as a new commit

davidhewitt:
Thanks, I'll try to review this later today (I have some related experiments from importing submodules for #759 from a few weeks' ago).

kngwyu:
>  I pushed the changes discussed in #1149 as a new commit

LGTM, let's merge this after David reviews.

davidhewitt:
👍 I did some thinking this morning on this and I believe it will cause no issues with what I was trying out; I'll confirm for definite later.

davidhewitt:
Yep looks great, let's merge this! 🎉

davidhewitt:
Thank you for your hard work on this!

PR #1146, opened 2020-09-03
Add #[allow(non_snake_case)] to function with generated name

### LINKED ISSUE #1145: getting warning: function `__init_Object_8818130445816726333` should have a snake case name
Adding a `#[allow(non_snake_case)]` annotation in the code below should fix it, since `#init` is not always a snake case identifier:

https://github.com/PyO3/pyo3/blob/9477ee85e7ea1a3d624b9294934cd7aeb5c74815/pyo3-derive-backend/src/pyproto.rs#L154-L161

programmerjake:
Fixes #1145


davidhewitt (approved):
:+1: thanks!

PR #1147, opened 2020-09-03
Gate PyModuleGetNameObject on PyPy.

sebpuetz:
`PyModuleGetNameObject` doesn't seem to exist on PyPy, therefore don't include it in the ffi.


davidhewitt:
👍 thanks

PR #1148, opened 2020-09-04
Add dependabot config

mtreinish:
This commit adds a dependabot config to automatically open pull requests
with version bumps for dependencies. This will configure dependabot to
check daily if any pyo3 dependency versions are out of date and if they
are it will open a PR for each dependency. You can find more details on
the dependabot configuration and usage in in the github
documentation. [1]

[1] https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "cargo" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
   - package-ecosystem: "cargo" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
-      interval: "daily"
+      interval: "weekly"


mtreinish:
This was prompted by me preparing a PR to bump the hashbrown version in the `Cargo.toml` since there was a new release today, but I figured instead of doing it manually we could just have the bot do it (that's how I was notified about the new release: https://github.com/Qiskit/retworkx/pull/133 )

sebpuetz:
I think keeping the dependencies up-to-date through a bot is nice, but I'm wondering if the MSRV can be configured such that it only proposes to update dependencies that are compatible with currently set minimal supported version.

mtreinish:
> I think keeping the dependencies up-to-date through a bot is nice, but I'm wondering if the MSRV can be configured such that it only proposes to update dependencies that are compatible with currently set minimal supported version.

I don't think dependabot has enough logic for that yet. It probably wouldn't be a hard feature to implement in the bot, but there is no metadata in the `Cargo.toml` for the MSRV yet (there is an approved rfc for it, but it hasn't been implemented). So it's not like the bot could easily know which version is the MSRV for pyo3 and for any dependency updates. I think at least for now we'd just have to rely on the MSRV ci job to detect this after the bot opens a PR.

sebpuetz:
I guess it's worth a try, if it ends up being too spammy it can still be disabled

konstin:
FYI dependabot is currently already configured through the GUI, though afaik github wants to move everyone to the file-based configuration

grafik

davidhewitt:
My view on this:

- yes, let's move to dependabot file, but weekly not daily please!
- if dependabot opens a PR which bumps MSRV before we want to do so, we can just close the PR and remember for later. I actually need to make an issue for the package updates waiting for 1.45. :)

kngwyu:
> but weekly not daily please!

Agreed

mtreinish:
Ok, I updated the update interval to be weekly instead of daily in https://github.com/PyO3/pyo3/pull/1148/commits/e0e82e37ec94a61b3d762e75dcc8ce35aab22a49

kngwyu (approved):

kngwyu:
Thanks!

PR #1157, opened 2020-09-06
docs: add detail to wrapper type conversions

### LINKED ISSUE #376: Add Table with conversions between wrapper types
We should add a table to the documentation that explains with trait and method to use to convert one wrapper type to another. This is complementary to #374.

For example you use `ToPyObject::to_object` to convert `&PySomeNativeType` to `PyObject` and there's `AsPyRef` to convert `PyObject` and `Py<T>` to `PyRef<T>`s.

davidhewitt:
I took a look at writing a table for #376. This table ended up being big and confusing, and also not possible to doctest. So in the end I decided it was better to extend the existing examples on `types.md`.

At the same time I also made the language more consistent and simpler.

Now every example is of the same format, e.g.

```rust 
// To &PyList with PyAny::downcast
let _: &PyList = any.downcast()?;
```

Closes #376

### DIFF
--- a/guide/src/types.md
+++ b/guide/src/types.md
@@ -61,23 +61,54 @@ such as `getattr`, `setattr`, and `.call`.
 
 **Conversions:**
 
+For a `&PyAny` object reference `any` where the underlying object is a Python-native type such as
+a list:
+
 ```rust
 # use pyo3::prelude::*;
-# use pyo3::types::PyList;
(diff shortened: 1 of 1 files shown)


kngwyu (approved):
LGTM, please merge this at your own convenience.

birkenfeld (commented):
:+1:

PR #1160, opened 2020-09-06
Remove dead fields from fptr structs

alex:


alex:
@kngwyu The compilation errors inside `Default`. At first I thought I could just use `derive(Default)` but that doesn't work because of the pointer fields :-)

kngwyu:
That makes sense, thanks.
Since current code is written assuming something like `type_object.tp_as_mapping = table;`, I think there's a room for simplification. (i.e. now we don't need table).

alex:
Yes, I think there's probably room for simplification. For now my focus is on getting abi3 all the way working though :-)

PR #1165, opened 2020-09-08
Fill tp_dict on types in an abi3-friendly way

alex:
I think this might technically be backwards incompatible if you had a custom metaclass with fancy behavior, but pyo3 doesn't seem to have any native support for those.


kngwyu (approved):
Thank you, this is the same approach as I was thinking as of :slightly_smiling_face:

PR #1166, opened 2020-09-08
Make PyType::name abi3 compatible

alex:


kngwyu:
Some points:
- How about `__qualname__`?
- If we have to extract String, we don't have to `Cow<str>`. This is a minor API and complete compatibility is not required. 
- As the same reason above, we can return `PyResult` if it is needed.

alex:
Ah yes, `__qualname__` looks much better!

src/err.rs
```diff
@@ -481,7 +481,7 @@ impl<'a> std::fmt::Display for PyDowncastError<'a> {
             self.from
                 .repr()
                 .map(|s| s.to_string_lossy())
-                .unwrap_or_else(|_| self.from.get_type().name()),
+                .unwrap_or_else(|_| Cow::Borrowed(self.from.get_type().name())),
```
kngwyu, line 484:
```suggestion
                .unwrap_or_else(|_| self.from.get_type().name().into()),
```

src/types/typeobject.rs
```diff
@@ -39,8 +37,11 @@ impl PyType {
     }
 
     /// Gets the name of the `PyType`.
-    pub fn name(&self) -> Cow<str> {
-        unsafe { CStr::from_ptr((*self.as_type_ptr()).tp_name).to_string_lossy() }
+    pub fn name(&self) -> &str {
+        self.getattr("__qualname__")
+            .expect("Error accessing __qualname__")
+            .extract()
+            .expect("__qualname__ not a string")
```
kngwyu, line 44: How about returning `PyResult<&str>`?
It's not fun to see the program crashed in such a trivial operation.
```suggestion
    pub fn name(&self) -> PyResult<&str> {
        self.getattr("__qualname__")?.extract()
```

kngwyu (changes requested):
Thanks, suggested some minor changes

tests/test_methods.rs
```diff
@@ -76,7 +76,7 @@ impl ClassMethod {
     #[classmethod]
     /// Test class method.
     fn method(cls: &PyType) -> PyResult<String> {
-        Ok(format!("{}.method()!", cls.name()))
+        Ok(format!("{}.method()!", cls.name().unwrap()))
```
kngwyu, line 79: Just `?` would work

alex:
sorry this has taken so many revisions

kngwyu (approved):

kngwyu:
Thanks!

PR #1169, opened 2020-09-09
Add null-check for function's documents

kngwyu:
We have a feature that extracts Python's doc string from doc comments in proc-macro code.

In the related parts, I found all of the code assumes that the corresponding document is null-terminated and does not check that.
This can make it difficult to debug our proc-macro code when it produces not null-terminated doc by accident, so I added a check using `CStr`.

### DIFF
--- a/pyo3-derive-backend/src/from_pyobject.rs
+++ b/pyo3-derive-backend/src/from_pyobject.rs
@@ -133,7 +133,9 @@ impl<'a> Container<'a> {
                 "Cannot derive FromPyObject for empty structs and variants.",
             ));
         }
-        let transparent = attrs.iter().any(ContainerAttribute::transparent);
+        let transparent = attrs
+            .iter()
+            .any(|attr| *attr == ContainerAttribute::Transparent);
         if transparent {
             Self::check_transparent_len(fields)?;
         }
@@ -182,7 +184,6 @@ impl<'a> Container<'a> {
         let err_name = attrs
             .iter()
             .find_map(|a| a.annotation())
-            .cloned()
             .unwrap_or_else(|| path.segments.last().unwrap().ident.to_string());
 
         let v = Container {
@@ -306,18 +307,10 @@ enum ContainerAttribute {
 }
 
 impl ContainerAttribute {
-    /// Return whether this attribute is `Transparent`
-    fn transparent(&self) -> bool {
-        match self {
-            ContainerAttribute::Transparent => true,
-            _ => false,
-        }
-    }
-
     /// Convenience method to access `ErrorAnnotation`.
-    fn annotation(&self) -> Option<&String> {
+    fn annotation(&self) -> Option<String> {
         match self {
-            ContainerAttribute::ErrorAnnotation(s) => Some(s),
+            ContainerAttribute::ErrorAnnotation(s) => Some(s.to_string()),
             _ => None,
         }
     }
--- a/src/types/function.rs
+++ b/src/types/function.rs
@@ -1,9 +1,6 @@
-use std::ffi::{CStr, CString};
-
 use crate::derive_utils::PyFunctionArguments;
-use crate::exceptions::PyValueError;
 use crate::prelude::*;
-use crate::{class, ffi, AsPyPointer, PyMethodType};
+use crate::{ffi, AsPyPointer, PyMethodDef};
 
 /// Represents a builtin Python function object.
 #[repr(transparent)]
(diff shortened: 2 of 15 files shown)


sebpuetz:
I think this is a nice improvement, maybe we can also change the docs to take `&str` instead of `&'static str` and rely on `CString::into_raw` for this, too?

It might also make sense to change the `name` and `doc` fields of `PyMethodDef` to `CString` and ensure well-formedness in constructors and in setter methods. If we also provide getter methods, we can encapsulate these fields, too.

E.g. something like

~~~Rust
impl PyMethodDef {
    fn set_name(&mut self, name: &str) -> Result<()> {
        let name = CString::new(name)?;
        self.name = name;
        Ok(())
    }
}

kngwyu:
> It might also make sense to change the name and doc fields of PyMethodDef to CString and ensure well-formedness in constructors and in setter methods.

It must be static and we cannot do so.

sebpuetz:
We can allocate a `String` and leak it, thus creating our own `&'static str` or we can mirror what is done now for `name` and use  `CString::into_raw` since this transfers ownership to the C-caller.

kngwyu:
Probably we can, but it would require lots of refactoring.

kngwyu:
Now I want to investigate some other solutions. E.g.,
```rust
pub struct PyMethodDef {
    pub ml_name: CString,
}
```

So I leave this PR unmerged for two or three days (it looks that I did too much OSS work this week and I have to do some research works as a student...)

kngwyu:
**UPDATED**
Now the internal representations of `ml_name` and `ml_doc` are `&'static CStr`. I.e., they are static strings with terminating null bytes, for efficeincy and null safety.
However, this change makes the `PyMethodDef` constructor not-const, and now `Pyo3MethodsInventory~` has `Vec<PyMethodDef>` instead of `&'static [PyMethodDef]`,

kngwyu:
image
:thinking:

kngwyu:
@davidhewitt 
If you're OK to use `Vec` instead of `&'static []`, I will merge this PR.

kngwyu:
And I'm sorry that this PR contains many changes to make fields of `Py~Def` types private, but I believe it's not good to keep them public.

davidhewitt (approved):
LGTM; I think `Vec` is fine here. Maybe _slightly_ less optimal but it's one-time-setup anyway and the safety is a huge plus. Maybe one day it'll be possible to do this with `const fn` 🤷 

I guess with the removal of public fields this is _technically_ a breaking change, so maybe let's merge this and abandon the idea of making a `0.12.2` patch release? We can proceed with iterating on improvements for `0.13`.

davidhewitt:
(Also this could probably benefit from a CHANGELOG entry static that it fixes #1211 )

kngwyu:
> I guess with the removal of public fields this is technically a breaking change, so maybe let's merge this and abandon the idea of making a 0.12.2 patch release? We can proceed with iterating on improvements for 0.13.

Hmm... I'm not sure we should treat this as *public* API :thinking:

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -133,7 +133,9 @@ impl<'a> Container<'a> {
                 "Cannot derive FromPyObject for empty structs and variants.",
             ));
         }
-        let transparent = attrs.iter().any(ContainerAttribute::transparent);
+        let transparent = attrs
```
sebpuetz, line 136: Changes here are unrelated?

src/types/function.rs
```diff
@@ -17,56 +14,32 @@ impl PyCFunction {
...
         fun: ffi::PyCFunctionWithKeywords,
-        name: &str,
+        name: &'static str,
         doc: &'static str,
         py_or_module: PyFunctionArguments<'a>,
     ) -> PyResult<&'a PyCFunction> {
-        let fun = PyMethodType::PyCFunctionWithKeywords(fun);
-        Self::new_(fun, name, doc, py_or_module)
+        Self::new_(
+            PyMethodDef::cfunction_with_keywords(name, fun, 0, doc),
```
sebpuetz, line 22: This removes error-handling of non-null terminated names and docs and instead causes panics in this API without documenting it. If panics are the way you want to handle non-terminated names and docs, it should at least be documented.

src/types/function.rs
```diff
@@ -17,56 +14,32 @@ impl PyCFunction {
     /// See [raw_pycfunction] for documentation on how to get the `fun` argument.
     pub fn new_with_keywords<'a>(
         fun: ffi::PyCFunctionWithKeywords,
-        name: &str,
+        name: &'static str,
```
sebpuetz, line 17: It's pretty awkward to expect null-terminated `&str` in a public API like this, taking `CStr` would signal this expectation on the type level.

pyo3-derive-backend/src/from_pyobject.rs
kngwyu: I did for address clippy warnings.

sebpuetz: Although clippy only asked to replace the implementation with `matches!()`, no?

src/types/function.rs
kngwyu: I think it's the same for `doc`... but am willing to avoid this.

pyo3-derive-backend/src/from_pyobject.rs
kngwyu: Yeah, but since we support MSRV, we cannot use `macthes`.

kngwyu:
@sebpuetz 
I updated `PyCFunction` to make it more backward-compatible, except now `doc` doesn't need to be null-terminated.

davidhewitt:
> Hmm... I'm not sure we should treat this as public API 🤔

Should we maybe be marking `pyo3::class::PyMethodDef` etc. with `#[doc(hidden)]` then? I'm more ok with treating this as a "patch fix" if we can claim that these should never have been public API 😄

kngwyu:
I made them `#[doc(hidden)]`, though I leave the release decision to you.

davidhewitt:
👍 anything else you want to change in this PR before it merges? I will release `0.12.2` soon after this is merged.

kngwyu:
No, I thin it's ready

PR #1174, opened 2020-09-11
Add documentation for raw_pycfuntion

sebpuetz:
This adds instructions on how to use the `raw_pycfunction!()` macro to the guide. I forgot to include this in #1163.

It doesn't touch any `.rs` file, so this could probably be merged post release, too.

### DIFF
--- a/guide/src/function.md
+++ b/guide/src/function.md
@@ -173,15 +173,31 @@ Type:      builtin_function_or_method
 
 ## Closures
 
-Currently, there are no conversions between `Fn`s in Rust and callables in Python. This would definitely be possible and very useful, so contributions are welcome. In the meantime, you can do the following:
+Currently, there are no conversions between `Fn`s in Rust and callables in Python. This would
+definitely be possible and very useful, so contributions are welcome. In the meantime, you can do
+the following:
 
 ### Calling Python functions in Rust
 
-You can use [`PyAny::is_callable`] to check if you have a callable object. `is_callable` will return `true` for functions (including lambdas), methods and objects with a `__call__` method. You can call the object with [`PyAny::call`] with the args as first parameter and the kwargs (or `None`) as second parameter. There are also [`PyAny::call0`] with no args and [`PyAny::call1`] with only positional args.
+You can pass Python `def`'d functions and built-in functions to Rust functions `[PyFunction]`
+corresponds to regular Python functions while `[PyCFunction]` describes built-ins such as
+`repr()`.
+
+You can also use [`PyAny::is_callable`] to check if you have a callable object. `is_callable` will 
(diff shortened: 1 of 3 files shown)


davidhewitt:
(Just a note to say thanks very much for writing these docs; I've been very busy lately, hoping to find time to read through these and review hopefully Monday.)

kngwyu:
I think the doc comment of `raw_pycfunction` could be a better home for this doc, and then we can link it from the guide.

guide/src/function.md
```diff
@@ -233,3 +249,41 @@ fn module_with_fn(py: Python, m: &PyModule) -> PyResult<()> {
...
+    module.add_function(py_cfunction)
+}
+
+# fn main() {}
+```
+
+The `wrap_pyfunction` macro can be used to directly get a `PyCFunction` given a
+`#[pyfunction]` and a `PyModule`: `wrap_pyfunction!(rust_fun, module)`. In the single-argument
+variant, the macro only takes the identifier of the Rust function and returns a function that
+can be called with either `Python` or `PyModule` and returns a `PyCFunction`.
```
davidhewitt, line 289: I think this paragraph may go into more detail than needed. (I'd rather not talk about the single-argument variant of `wrap_pyfunction` any more so that the ecosystem naturally migrates to the two-argument form.)

guide/src/function.md
```diff
@@ -233,3 +249,41 @@ fn module_with_fn(py: Python, m: &PyModule) -> PyResult<()> {
...
+fn module(_py: Python, module: &PyModule) -> PyResult<()> {
+    let ffi_wrapper_fun = raw_pycfunction!(some_fun);
+    let docs = "Some documentation string with null-termination\0";
+    let py_cfunction =
+        PyCFunction::new_with_keywords(ffi_wrapper_fun, "function_name", docs, module.into())?;
+    module.add_function(py_cfunction)
+}
+
+# fn main() {}
+```
```
davidhewitt, line 284: I think @kngwyu is correct; this document is great to go as a docstring on `raw_pycfunction` or maybe even `PyCFunction::new_with_keywords.

And then in the paragraph above can just have a link e.g. `(See the docs for [raw_pycfunction](https://docs.rs/pyo3/0.12.0/pyo3/macro.raw_pycfunction.html)).

I can't decide whether it's better for `raw_pycfunction` or `PyCFunction::new_with_keywords` though - happy for you to decide!

davidhewitt (commented):
Sorry for the delay, just got round to reading it. 

Thanks for taking the time to write this. It's some great documentation. I agree with @kngwyu that the example on the bottom is probably better to go as API documentation rather than letting the guide get too long.

davidhewitt:
Thanks!

PR #1175, opened 2020-09-11
Fix cargo test with Anaconda python, and divergent gcc versions.

thatcr:
This change forces pyo3 to use python from a shared library when using Anaconda python distributions.  Currently cargo test fails with:
```
= note: lto1: fatal error: bytecode stream in file '.../target/debug/deps/libpyo3-90ae49a6516fe776.rlib' generated with LTO version 6.0 instead of the expected 8.1
          compilation terminated.
```

unless you match the gcc version with Anaconda's.

Anaconda includes both static and shared versions of python however the python executable is static, so build.rs pyo3 embeds the  libpython*.a file, built with Anaconda's gcc 7.3.

This causes the tests and downstream crates to see it also and breaks builds where a different version of gcc is used as the LTO versions in the archive become unsupported. 

With this change we always use the .so and avoid any embedded LTO issues altogether.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Fix `build.rs` to work with Anaconda python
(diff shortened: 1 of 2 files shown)


gilescope (commented):
LGTM

davidhewitt:
🚀 thanks very much for this!

PR #1176, opened 2020-09-11
Remove special case of PyIterator

birkenfeld:
Instead of `PyIterator<'p>` implement the iter protocol on `&'p PyIterator`.

This lets us implement `PyNativeType`, pass `PyIterator` as a function argument etc.

Is there a reason I don't get why this was handled specially?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Change `PyIterator` to be consistent with other native types: it is now used as `&PyIterator` instead of `PyIterator<'a>`. [#1176](https://github.com/PyO3/pyo3/pull/1176)
+
 ### Removed
 - Remove unused `python3` feature. [#1209](https://github.com/PyO3/pyo3/pull/1209)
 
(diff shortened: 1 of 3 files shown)


davidhewitt:
Thanks, it's nice to make `PyIterator` consistent with the rest of the api!

> Is there a reason I don't get why this was handled specially?

🤔 yeah I'm not entirely sure of the reason for this either. I have a couple guesses:
- possibly before the "owned references" design was added to PyO3, PyIterator was already a special case, and nobody else ever noticed it could be updated.
- I also wonder whether the early drop behaviour of `PyIterator<'a>` versus `&PyIterator` is important for some runtime behaviour. But from a quick test just now I couldn't prove that. (And the test suite appears to pass.)

However, I'm of split opinion whether to merge this before releasing 0.12.

Pros:
- it would simplify the api (which is always awesome imo 🚀)

Cons:
- we're not entirely sure there wasn't a reason for this (which isn't a reason not to do it, but it could be a reason to wait and think about this a little for 0.13)
- I think we _might_ be revisiting the owned object design for 0.13 based off discussions in #1056, and it might be nice to change the `PyIterator` API just once in 0.13 rather than once now and once in the next release too.

davidhewitt:
I think on balance I'd vote to wait a moment with this PR until 0.12 releases, and then let's merge it on the road to 0.13. That should give us plenty of time to learn if this change was ok - and if we make any bigger changes for 0.13 we can keep everything consistent.

davidhewitt:
I rebased on master, added CHANGELOG entry and test. Will merge shortly...

PR #1179, opened 2020-09-12
Don't consider it cross-compilation when building for 32-bit Windows on 64-bit windows

alex:
This is a bit of an untested hack :-)

In https://github.com/PyO3/setuptools-rust/pull/77 we introduced the ability to automatically build for 32-bit Windows when running a 32-bit Python on a 64-bit Windows (and Rust toolchain). The 0.12 release of pyo3 unfortunately broke this. I believe this PR may resolve it.


davidhewitt:
Thanks, nice catch. I thought our CI would catch this when I commented it seemed a possible breakage in #1095, but at the time CI suggested it was ok (and I was suprised but let it slide). Can you add a CHANGELOG entry please? (fixed section) 

I will rebase the setuptools-rust github actions CI PR I had, which I guess will find the same issue.

Might give this a couple days once merged before releasing just so that 0.12.1 can include fixes for any other painful oversights!

alex:
Added a changelog entry. Thanks!

kngwyu (approved):
Good catch, thank you!

PR #1184, opened 2020-09-13
Replace poking at type internals with usage of a public API

alex:


kngwyu:
So here we assume that `tp_alloc==NULL` for all subtypes... and I'm not sure it's OK.

alex:
I wasn't sure either :-) But there's no test where it fails, perhaps we need a new test for some other case?

alex:
This is rolled ino #1183

Issue #1185, opened 2020-09-14
Not raising a TypeError when it should after reversing the logical operator.


mvaled:
When our implementation of a pyclass returns NotImplemented and the other side also returns NotImplemented, a TypeError must be raised.

This is maybe a consequence of the implementation for #1064.  

The following code reproduces the issue:

```rust
    #[pyclass]
    #[derive(PartialEq, PartialOrd, Eq, Ord, Clone, Debug)]
    pub(crate) struct ExternalObject {
        name: String,
        id: u64,
    }

    #[pymethods]
    impl ExternalObject {
        #[new]
        fn new(name: String, id: u64) -> Self {
            Self { name, id }
        }
    }

    #[pyproto]
    impl<'p> PyObjectProtocol<'p> for ExternalObject {
        fn __richcmp__(&'p self, other: Self, op: CompareOp) -> PyResult<bool> {
            unimplemented!()
        }
    }
```

When calling from Python:

```python
>>> ExternalObject("name", 1).__eq__(("name", 1))                                                                                                                                                                  
NotImplemented
>>> ("name", 1).__eq__(ExternalObject("name", 1))                                                                                                                                                                  
NotImplemented

>>> ExternalObject("name", 1) == ("name", 1)                                                                                                                                                                       
False

>>> ("name", 1) == ExternalObject("name", 1)                                                                                                                                                                       
False
```

Even though both calls to `__eq__` return NotImplemented the operator `==` doesn't raise a TypeError and it should.

kngwyu:
I'm sorry for the delay, but is it really true for `==`?
I can't see `TypeError` in Python. Here's my attempt:
```python
In [10]: class C:
    ...:     def __eq__(self, a):
    ...:         return NotImplemented

In [11]: 1 == C()
Out[11]: False

In [12]: C() == 1
Out[12]: False
```

But I'm sure this is true for some number methods and our current implementation is incorrect.
```python
In [10]: class C:
    ...:     def __eq__(self, a):
    ...:         return NotImplemented

In [11]: 1 == C()
Out[11]: False

In [12]: C() == 1
Out[12]: False

In [13]: class D:
    ...:     def __add__(self, a):
    ...:         return NotImplemented
    ...: 

In [14]: D() + D()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-14-715d6e17ac3a> in <module>
----> 1 D() + D()

[...]

PR #1186, opened 2020-09-14
Add regression test for TypeError after returning NotImplemented.

mvaled:
This PR adds a regression test for issue #1185.

### DIFF
--- a/src/class/macros.rs
+++ b/src/class/macros.rs
@@ -125,10 +125,22 @@ macro_rules! py_binary_fallback_num_func {
                 match (lhs.extract(), rhs.extract()) {
                     (Ok(l), Ok(r)) => $class::$lop(l, r).convert(py),
                     _ => {
-                        // Next, try the right hand method (e.g., __radd__)
-                        let slf: &$crate::PyCell<T> = extract_or_return_not_implemented!(rhs);
-                        let arg = extract_or_return_not_implemented!(lhs);
-                        $class::$rop(&*slf.try_borrow()?, arg).convert(py)
+                        // Next, try the right hand method (e.g., __radd__),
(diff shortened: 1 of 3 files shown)


kngwyu:
Thank you for finding this!
Looks the bug is around https://github.com/PyO3/pyo3/blob/a0960f891801c0534856cb90fa90451828579470/src/class/macros.rs#L128.
Here we use `extract_or_not_implemented!` even after the fallback call, but we should raise type error instead.

Are you willing to fix this yourself on this branch?
Or I can send a PR on this.

mvaled:
@kngwyu Now that you have pinpointed the place where it happens I could try to fix it.  If I fail, I'll pass it to you ;)

kngwyu:
@mvaled 
Thank you, let us know any question if you get stucked.

mvaled:
@kngwyu 

I tried several things, without luck.  The reversed operator may itself return NotImplemented (which I think would be the case of our test), so I need to compare the result of the reversed call to NotImplemented.

The last commit (7d15e51ef7989a34402a06f8f77f1d3c56bc97c3) is the last attempt I've made so far.

mvaled:
@kngwyu 

Well, the code of our `py_binary_fallback_num_func` is not being run for `obj == 1`, because `==` is implemented with `py_binary_reversed_num_func`.

I'm going to change that as well, while extracting the *upgrade* of NotImplemented to a TypeError in a macro.

mvaled:
Forget about my previous comment.  Still doesn't work.  I'd need to know how `tp_richcompare` is handled.  

But I think @kngwyu could possibly fix this faster than I could.  I'm gonna close this PR and let you take it over, ok?

PR #1188, opened 2020-09-15
Make check warning clean in limited API mode

alex:


src/pyclass.rs
```diff
@@ -51,6 +51,8 @@ pub(crate) unsafe fn default_new<T: PyTypeInfo>(
         }
         #[cfg(Py_LIMITED_API)]
         {
+            // Silence unused parameter warning.
+            let _ = py;
```
sebpuetz, line 55: Could this be solved in a cleaner manner by splitting the function into #[cfg(PY_LIMITED_API)] and #[cfg(not(PY_LIMITED_API))] fns? This method is currently rather hard to read!

src/pyclass.rs
```diff
@@ -51,6 +51,8 @@ pub(crate) unsafe fn default_new<T: PyTypeInfo>(
...
             unreachable!("Subclassing native types isn't support in limited API mode");
```
sebpuetz, line 56: Is this line actually unreachable? Or is this here to panic if hitting this codepath on limited API? At least the control flow in the function seems to allow hitting this block...

src/pycell.rs
```diff
@@ -8,9 +8,9 @@ use crate::type_object::{PyBorrowFlagLayout, PyLayout, PySizedLayout, PyTypeInfo
 use crate::types::PyAny;
 use crate::{ffi, IntoPy, PyErr, PyNativeType, PyObject, PyResult, Python};
 use std::cell::{Cell, UnsafeCell};
+use std::fmt;
```
sebpuetz, line 11: There's a bit of inconsistency, in other files, e.g. types::complex, the imports are feature-gated. Although, I think your approach here is more legible since there's just less conditionals involved.

sebpuetz (commented):
I moved the comments from the other PR to this one.

src/pycell.rs
alex: I agree it's not super consistent. I'm not particularly worried though, choices like this area very easy to change later :-)

FWIW, I did `std::mem::size_of` in this file specifically because there were several other existing callers that used that style in this file.

src/pyclass.rs
alex: It could be solved that way, but I'm not sure it'd be any cleaner -- splitting it can also make it harder to reuse logic between the different cases.

src/pyclass.rs
alex: You're right, it is reachable. @kngwyu asked for it to be `unreachable!()` instead of `panic!()` with the idea that maybe in the future we'd make changes in `pyo3-derive-backend` to make it impossible.

kngwyu: I'll make it unreachable by hacking the proc-macro code.

src/pycell.rs
kngwyu: I actually don't care about how we limit imports.
Please chooes the way you like.

kngwyu (approved):
LGTM, thank you!

kngwyu:
So...
For now, all things don't have to be perfect. We can fix some coding styles later when in #1152

PR #1189, opened 2020-09-15
Get all the tests building and passing!

alex:
This incorporates #1188, so I'll rebase it once that merges.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -271,7 +271,7 @@ You can also inherit native types such as `PyDict`, if they implement
 However, because of some technical problems, we don't currently provide safe upcasting methods for types
 that inherit native types. Even in such cases, you can unsafely get a base class by raw pointer conversion.
 
-```rust
+```rust,no_run
 # use pyo3::prelude::*;
 use pyo3::types::PyDict;
 use pyo3::{AsPyPointer, PyNativeType};
--- a/pyo3-derive-backend/src/from_pyobject.rs
+++ b/pyo3-derive-backend/src/from_pyobject.rs
@@ -249,7 +249,7 @@ impl<'a> Container<'a> {
         let self_ty = &self.path;
         let mut fields: Punctuated<TokenStream, syn::Token![,]> = Punctuated::new();
         for i in 0..len {
-            fields.push(quote!(slice[#i].extract()?));
+            fields.push(quote!(s.get_item(#i).extract()?));
         }
         let msg = if self.is_enum_variant {
             quote!(format!(
@@ -265,7 +265,6 @@ impl<'a> Container<'a> {
             if s.len() != #len {
                 return Err(::pyo3::exceptions::PyValueError::new_err(#msg))
             }
-            let slice = s.as_slice();
             Ok(#self_ty(#fields))
         )
     }
--- a/src/types/floatob.rs
+++ b/src/types/floatob.rs
@@ -83,7 +83,9 @@ impl<'source> FromPyObject<'source> for f32 {
 mod test {
     #[cfg(not(Py_LIMITED_API))]
     use crate::ffi::PyFloat_AS_DOUBLE;
-    use crate::{AsPyPointer, Python, ToPyObject};
+    #[cfg(not(Py_LIMITED_API))]
+    use crate::AsPyPointer;
+    use crate::{Python, ToPyObject};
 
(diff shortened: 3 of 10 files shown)


alex:
I think I've resolved the doctest issue.

pyo3-derive-backend/src/from_pyobject.rs
```diff
@@ -249,7 +249,7 @@ impl<'a> Container<'a> {
         let self_ty = &self.path;
         let mut fields: Punctuated<TokenStream, syn::Token![,]> = Punctuated::new();
         for i in 0..len {
-            fields.push(quote!(slice[#i].extract()?));
+            fields.push(quote!(s.get_item(#i).extract()?));
```
sebpuetz, line 252: Is the memory layout different for `PyTuple`s with limited ABI? I know this change here is necessary because `as_slice()` is no longer available on `PyTuple`:
https://github.com/PyO3/pyo3/blob/2ec1c3b0b9bb0c2e5ecd07d9014ac1a9ae890634/src/types/tuple.rs#L84-L93

It probably won't have a measurable impact anyways, just curious..

src/pyclass.rs
```diff
@@ -51,6 +51,8 @@ pub(crate) unsafe fn default_new<T: PyTypeInfo>(
         }
         #[cfg(Py_LIMITED_API)]
         {
+            // Silence unused parameter warning.
+            let _ = py;
```
sebpuetz, line 55: Could this be solved in a cleaner manner by splitting the function into `#[cfg(PY_LIMITED_API)]` and `#[cfg(not(PY_LIMITED_API))]` fns? This method is currently rather hard to read!

src/pycell.rs
```diff
@@ -8,9 +8,9 @@ use crate::type_object::{PyBorrowFlagLayout, PyLayout, PySizedLayout, PyTypeInfo
 use crate::types::PyAny;
 use crate::{ffi, IntoPy, PyErr, PyNativeType, PyObject, PyResult, Python};
 use std::cell::{Cell, UnsafeCell};
+use std::fmt;
```
sebpuetz, line 11: There's a bit of inconsistency, in other files, e.g. `types::complex`, the imports are feature-gated. Although, I think your approach here is more legible since there's just less conditionals involved.

src/pyclass.rs
```diff
@@ -51,6 +51,8 @@ pub(crate) unsafe fn default_new<T: PyTypeInfo>(
...
             unreachable!("Subclassing native types isn't support in limited API mode");
```
sebpuetz, line 56: Is this line actually unreachable? Or is this here to panic if hitting this codepath on limited API? At least the control flow in the function seems to allow hitting this block...

sebpuetz (commented):
I took a look at this, I'm not deep into the topic, so most of the comments are superficial or aimed at legibility etc.

alex:
Just a note that this PR includes https://github.com/PyO3/pyo3/pull/1188 so it's probably better to review that one first.

sebpuetz:
I moved the applicable comments to #1188 and deleted them in this PR. There's still a question based on changes in this PR. The actual changes that make the change in this PR necessary happened earlier and are already merged in the abi3 branch, so I think this is the correct place to discuss it.

pyo3-derive-backend/src/from_pyobject.rs
alex: They _can_ be, which is the whole point of the limited ABI :-)

alex:
Ok, rebased and ready to go!

guide/src/class.md
```diff
@@ -271,7 +271,7 @@ You can also inherit native types such as `PyDict`, if they implement
 However, because of some technical problems, we don't currently provide safe upcasting methods for types
 that inherit native types. Even in such cases, you can unsafely get a base class by raw pointer conversion.
 
-```rust
+```rust,no_run
```
kngwyu, line 274: What is the problem with this?

alex: subclasses a builtin type which isn't support on abi3 (and I don't know how to make a doc test `no_run` only in some cfgs)

kngwyu: OK, though I'll have to modify this when it is changed to raise a compile error.

kngwyu (approved):
Thanks!

PR #1191, opened 2020-09-15
Release 0.12.1

davidhewitt:
There's been a couple of bugfixes to 0.12 which I think warrant a patch release before we start merging any new / breaking changes on the way to 0.13.

I'd like to publish this tomorrow evening (16th Sep). We can always make further patch releases if necessary (hopefully not).

I think it would be nice if a bugfix for #1185 was also included in this release. @kngwyu, I'm not sure if you're already working on one, I'll have a go at writing a fix tomorrow evening if I don't hear a reason not to.


davidhewitt:
As #1185 has been closed with conclusion that no fix needed, I'm going to proceed to release in a moment...

kngwyu:
Thanks!

PR #1199, opened 2020-09-16
Tidy up guide release

davidhewitt:
The changes from #1198 worked nicely, but the process is a bit messy. It currently runs three times and also doesn't have the version number in the commit that gets pushed to github pages.

This PR just tidies up those loose ends for when the next release happens.

### DIFF
--- a/.github/workflows/guide.yml
+++ b/.github/workflows/guide.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
   release:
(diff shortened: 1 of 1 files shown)


kngwyu:
You mean `release` has three types so it runs three times?
Sounds confusing... and thank you for this good tip :+1:

davidhewitt:
Yep exactly 😊

PR #1201, opened 2020-09-19
Hack __text_signature__ back to working with abi3

alex:


alex:
This gross hack seems to break PyPy -- not a big surprise. I guess I'll just disable it there.

tests/test_text_signature.rs
```diff
@@ -72,7 +70,7 @@ fn class_with_docs_and_signature() {
 // Ignored because heap types don't have working __text_signature__:
 // https://github.com/python/cpython/blob/master/Objects/typeobject.c#L864-L870
```
kngwyu, line 71: It's also outdated?

kngwyu (approved):
Thanks, looks nice as a remedy :+1:

tests/test_text_signature.rs
alex: Ooops. Fixed.

Issue #1195, opened 2020-09-16
Specifying the minimum supported python version


konstin:
Currently, there is no way to specify minimal supported python version, and compiling against a too low version or fail with an error about missing symbols (or worse, at runtime due to some failed import). With maturin, I'd like to build abi3 wheels for the lowest supported python version, so there's only one wheel and that supports all versions. I'm currently thinking about how to best specify this. Ideally, pyo3's build script or at least setuptools-rust should also check for this version and exit with an understandable error message about the wrong python version (I can write the pull request).

I currently have three different ideas:

 * In the cargo metadata section. The big plus is that user already configure a Cargo.toml. 
 * In pyproject.toml. We could start in the tools section and switch over the official `python-requires` when PEP 621 is accepted. 
 * Feature flags. Add `py36-abi3`, `py37-abi3`, etc. features

@kngwyu @davidhewitt What do you think?

davidhewitt:
Is it possible to use metadata from the package itself in pyo3's build.rs? Even if it is I'm not enthusiastic about doing that. OTOH Maturin could definitely check the cargo metadata, and setuptools-rust the `pyproject.toml`.

I wonder if we can add macros like [rustversion](https://github.com/dtolnay/rustversion) ?

kngwyu:
I think the combination of feature flags (for build scripts) + toml config (for maturin or setuptools) is good, though we have to invert `unstable-api` flags to do so.

> I wonder if we can add macros like rustversion ?

We already have `#[cfg(Py_3_7)]` or so for that purpose.

davidhewitt:
> We already have #[cfg(Py_3_7)] or so for that purpose.

`#[cfg(Py_3_7)]` etc. only works inside of the `pyo3` code though; downstream crates don't see these flags unless we give them a way to add them in their own build scripts.

konstin:
I checked that only setting the cfg's in build.rs works nicely with abi3. I've just checked with this hacky snippet:

```rust
fn main() -> Result<()> {
    if env::var_os("CARGO_FEATURE_ABI3").is_some() {
        let msg = r#"cargo:rustc-cfg=Py_3_5
cargo:rustc-cfg=Py_3_6
cargo:rustc-cfg=Py_3_7
cargo:rustc-cfg=Py_3
cargo:rustc-cfg=py_sys_config="WITH_THREAD"
[...]

PR #1203, opened 2020-09-20
Start documenting abi3 support

alex:


davidhewitt (changes requested):
This PR documents the `abi3` feature, but I understood we decided to invert the pattern and instead have an `unstable-abi` feature which would be enabled by default?

There was a commit on the `abi3` branch to this effect: https://github.com/PyO3/pyo3/pull/1152/commits/1941f4daefa940d5f5426061ba8598e781186031

However that change doesn't appear to be present in the latest version of #1152 - perhaps came unstuck with a merge conflict?

kngwyu:
> I understood we decided to invert the pattern and instead have an unstable-abi feature which would be enabled by default?

I inverted the change to introduce `py3*-abi3` features.

davidhewitt:
> I inverted the change to introduce py3*-abi3 features.

Can you elaborate? I don't think I can see these py3*-abi3 features on that branch at the moment either.

kngwyu:
Sorry, I meant that 'I changed it to `abi3` so that it would be integrated well with `py3*-abi3` features that we're going to introduce based on the discussion in #1195'

alex:
Yes, this documents the branch as it exists today, and I followed this documentation when I wrote the tests for setuptools-rust, so I'm sure it works :-)

davidhewitt (approved):
:+1: in that case I propose let's merge it and review everything on the abi3 branch!

PR #1207, opened 2020-09-23
Link python3.lib instead of python3x.lib on Windows in abi3 mode

alex:
Without this, on windows we'd link `python35.lib` (for example) which doesn't exist on other pythons of course :-)

### DIFF
--- a/tests/test_pyfunction.rs
+++ b/tests/test_pyfunction.rs
@@ -1,7 +1,9 @@
 #[cfg(not(Py_LIMITED_API))]
 use pyo3::buffer::PyBuffer;
 use pyo3::prelude::*;
-use pyo3::types::{PyCFunction, PyFunction};
+use pyo3::types::PyCFunction;
+#[cfg(not(Py_LIMITED_API))]
+use pyo3::types::PyFunction;
 use pyo3::{raw_pycfunction, wrap_pyfunction};
 
 mod common;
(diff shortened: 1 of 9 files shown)


alex:
It looks like we still rely on a few things that aren't present in `python3.lib`. I'll add more more commits to this to fix it and we can split them out into a separate PR if it's too much.

tests/test_pyfunction.rs
```diff
@@ -81,21 +84,32 @@ fn function_with_pycfunction_arg(fun: &PyCFunction) -> PyResult<&PyAny> {
...
     let py_cfunc_arg = wrap_pyfunction!(function_with_pycfunction_arg)(py).unwrap();
     let bool_to_string = wrap_pyfunction!(optional_bool)(py).unwrap();
 
     pyo3::py_run!(
         py,
-        py_func_arg
-        py_cfunc_arg
-        bool_to_string,
+            py_cfunc_arg
+            bool_to_string,
```
kngwyu, line 93: Looks badly formatted?

kngwyu (approved):
LGTM, except for a formatting thing, thanks!

tests/test_pyfunction.rs
alex: fixed

alex:
Ok, green now!

PR #1214, opened 2020-10-07
Add wrapper for PyErr_CheckSignals() to Python.

birkenfeld:
I don't know how to test this...

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -498,6 +498,26 @@ impl<'p> Python<'p> {
     pub fn xdecref<T: IntoPyPointer>(self, ptr: T) {
         unsafe { ffi::Py_XDECREF(ptr.into_ptr()) };
     }
+
+    /// Lets the Python interpreter check for pending signals and invoke the
+    /// corresponding signal handlers. This can run arbitrary Python code.
+    ///
+    /// If an exception is raised by the signal handler, or the default signal
+    /// handler raises an exception (such as `KeyboardInterrupt` for `SIGINT`),
+    /// an `Err` is returned.
+    ///
+    /// This is a wrapper of the C function `PyErr_CheckSignals()`. It is good
+    /// practice to call this regularly in a long-running calculation since
+    /// SIGINT and other signals handled by Python code are left pending for its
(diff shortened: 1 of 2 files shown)


kngwyu:
Looks good but we have to fix PyPy CI first (it's unrelated to this patch, though).

davidhewitt:
CI actually only prevents merging if any of the python 3.8 jobs fail atm, so no need to rebase on #1215.

I was going to ask though - should this api be `PyErr::check_signals` instead of `Python::check_signals`? We already also have `PyErr_WarnEx` wrapped on `PyErr` in this way.

birkenfeld:
Well IMO this only has a marginal connection to error handling, and the C version is named unfortunately. It's rather an interpreter related thing.

In newer Pythons, they added APIs with PySignal_ prefix but this didn't exist back then.

davidhewitt (approved):

davidhewitt:
👍

I think it's ok to let a test slide on this one for now as it's a complicated test to write, and this is basically just wrapping the C-API?

Could use a CHANGELOG entry stating the addition of the new method.

src/python.rs
```diff
@@ -498,6 +498,21 @@ impl<'p> Python<'p> {
     pub fn xdecref<T: IntoPyPointer>(self, ptr: T) {
         unsafe { ffi::Py_XDECREF(ptr.into_ptr()) };
     }
+
+    /// Lets the Python interpreter check for pending signals and invoke the
+    /// corresponding signal handlers. This can run arbitrary Python code.
+    ///
+    /// If an exception is raised by the signal handler, or the default signal
+    /// handler raises an exception (such as `KeyboardInterrupt` for `SIGINT`),
+    /// an `Err` is returned.
```
davidhewitt, line 507: Can this doc maybe also mention that it's a safe wrapper about `PyErr_CheckSignals`? I think it can be helpful for some people to know this without having to look at the source.

birkenfeld: Sure!

kngwyu (approved):
Thanks!

PR #1215, opened 2020-10-08
Exclude PyPy CI for windows

kngwyu:
I don't know what to do but for now, push some changes *experimentally*.


kngwyu:
It looks like `tox.interpreters.via_path.get_python_info("pypy3")` causes the problem, but I have no idea why this started to fail.

kngwyu:
I finally gave up...

davidhewitt:
Let's create an issue to fix? And then I'll try to have a look some time.

kngwyu:
:+1:

PR #1226, opened 2020-10-10
Use stable Python 3.9 on CI

davidhewitt:
Use released Python 3.9, add Python 3.10-dev, and remove Python 3.5.


davidhewitt:
Looks like 3.10-dev is not available yet, so I've removed it again for now and put back 3.5. (I wanted to keep the number of jobs down.)

kngwyu (approved):
Thanks!

PR #1231, opened 2020-10-12
Don't watch PYTHON_SYS_EXECUTABLE and PATH when unnecessary

### LINKED ISSUE #1229: PyO3 always retriggers compilation when used with IDE/Shell combination.
## 🐛 Bug Reports

### 🌍 Environment

 - Your operating system and version: Linux, MacOS
 - Your python version: `3.6`, `3.7`,`3.8`
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: Pyenv and system native (Ubuntu)
 - Your Rust version (`rustc --version`): `nightly-2020-07-11`
 - Your PyO3 version: `0.11.1`, `0.9.2`
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: No

### 💥 Reproducing

Any PyO3 code should face the same behaviour.

Hello,

Thank you very much for the PyO3 project. This has been a massive game changer in accelerating python based projects. I noticed that PyO3 was triggering recompilation (or rechecking) after every change I made, even if the change was in an unrelated crate.

The source was found out to be rust-analyzer running in the IDE, where, due to many factors, the `PATH` can be different than in the `shell`. Under certain other environments, the `PATH` can be different between two shells as well. (Something like `fnm` uses `PATH` to set the node versions, changing the variable). I'm not certain, but IDEs such as VSCode also modify PATH before executing.

All of this makes for a very excruciating development environment, especially in very large projects where PyO3 based code is a relatively static and small part of the codebase. We've resorted to using a forked version with the `PATH` rechecking disabled, but I thought I should still file the bug since other people may also have the same issue.

I understand the need for checking `PATH` to trigger recompilation, but I was wondering if there was an alternative for situations where the `PATH` is expected to change unrelated to the location of the `python` executable. My build.rs knowlege is a bit lacking, but I was wondering if this would be solved by adding a cargo feature flag or something.

Once again, thank you so much for the time you've put into PyO3.

kngwyu:
Fixes #1229. 
I confirmed that this works manually, though it's better if @nikhilmitrax can check this.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -893,16 +893,19 @@ fn main() -> Result<()> {
         // TODO: Find out how we can set -undefined dynamic_lookup here (if this is possible)
     }
 
-    let env_vars = [
-        "LD_LIBRARY_PATH",
-        "PATH",
-        "PYTHON_SYS_EXECUTABLE",
-        "PYO3_PYTHON",
-        "LIB",
-    ];
-
-    for var in env_vars.iter() {
+    for var in ["LIB", "LD_LIBRARY_PATH", "PYO3_PYTHON"].iter() {
         println!("cargo:rerun-if-env-changed={}", var);
+    }
+
+    if env::var_os("PYO3_PYTHON").is_none() {
+        // When PYO3_PYTHON is not used, PYTHON_SYS_EXECUTABLE has the highest priority.
+        // Let's watch it.
+        println!("cargo:rerun-if-env-changed=PYTHON_SYS_EXECUTABLE");
+        if env::var_os("PYTHON_SYS_EXECUTABLE").is_none() {
+            // When PYTHON_SYS_EXECUTABLE is also not used, then we use PATH.
+            // Let's watch this, too.
+            println!("cargo:rerun-if-env-changed=PATH");
+        }
     }
 
     Ok(())


kngwyu:
Oops, my first attempt was completely broken... 
Please try the force-pushed one.

nikhilmitrax:
yeah figured. I will test now and respond back. Unfortunately my test code depends on multiple versions of PyO3, so will backport the patch and test.

nikhilmitrax:
Seems to be working.

Btw, just my 2c,

It might be more ergonomic to check if `PYO3_PYTHON` is set at the top of the build script, and setting it to the path of `python` or `python3` to the env variable, and then just simply check `PYO3_PYTHON` for changes.

Either way, this unblocks me so I'm happy with this.

kngwyu:
> It might be more ergonomic to check if PYO3_PYTHON is set at the top of the build script, and setting it to the path of python or python3 to the env variable, and then just simply check PYO3_PYTHON for changes.

But then, if `PYO3_PYTHON` originates from `PATH` or `PYTHON_SYS_EXECUTABLE`, don't we need to check them altogether?

nikhilmitrax:
I meant something like this:

https://github.com/PyO3/pyo3/compare/master...nikhilmitrax:master

(non working code, and the structure definitely needs to be different, but should explain the idea.)

This _should_ achieve the same effect as the original MR, with reduced cargo instructions.

However, this sort of introduces a magical coupling between `find_interpreter()` and `main()`, which I'm not so sure I love. I'll defer to you for what you think is better.

davidhewitt:
I'm not entirely convinced that'll work for users that are relying on `PATH` instead of `PYO3_PYTHON`, as I don't think cargo will know in that case that it needs to rerun when `PATH` changes.

As we are happy the existing one implementation works, I'm going to merge this as-is and proceed with releasing it in `0.12.2`.

PR #1233, opened 2020-10-12
Update environment setting for Github Actions API updates

alex:


alex:
The failed build looks spurious?

davidhewitt:
Agreed. Thanks very much for the tidy up, the new syntax is much nicer 👍

alex:
Also helps github to be able to remove it to address a security issue https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

PR #1243, opened 2020-10-18
guide: improve "calling Python from Rust" section

davidhewitt:
I was looking at the "conversion traits" section of the guide when I noticed the "args and kwargs" section was probably in the wrong place.

I moved it to the "Calling Python from Rust" chapter and reworded it slightly.


kngwyu (approved):
Thanks!

PR #1246, opened 2020-10-18
Make get_free_list API sound

alex:


davidhewitt:
👍 I suspect that so far we've been lucky because this code is in practice only called with the GIL held anyway. Nice catch, thanks!

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -267,7 +267,7 @@ fn impl_class(
             quote! {
                 impl pyo3::freelist::PyClassWithFreeList for #cls {
                     #[inline]
-                    fn get_free_list() -> &'static mut pyo3::freelist::FreeList<*mut pyo3::ffi::PyObject> {
```
davidhewitt, line 270: I already merged this PR, but now I think about it should this actually return just `&mut Freelist`, so that it's not possible to get a reference that outlives the GIL lifetime?

alex: Hmm, yes, probably.

davidhewitt: 👍 I'm going to write a quick follow-up PR to clean this up.

kngwyu:
Since `get_free_list` doesn't call CPython API, I'm not sure what the motivation for this.
Or you need to override the default implementation using `py`?

davidhewitt:
It's fixing a thread-safety issue: this function is handing out `& mut` references to a global variable. Introducing the `Python` token ensures that no two threads can access this concurrently.

Issue #1241, opened 2020-10-17
sys.argv support: Consider call PySys_SetArgv on python initialization.


XUWeijiang:
Currently if I use sys.argv in python code, it would complain:  'module' object has no attribute 'argv'
This is because python's argv is not initialized. 

I hope PyO3 can provide an option to initialize argv in its initialization.

davidhewitt:
I wonder if we should be using the [python initialization configuration](https://docs.python.org/3/c-api/init_config.html#init-config) from Python 3.8 onwards? I'll have a play with this this morning.

davidhewitt:
@XUWeijiang I took a look this morning and I can't reproduce this issue. On my system the snippet you provide above works fine, and `sys.argv` exists.

Can you please provide detail on your OS and Python version?

XUWeijiang:
@davidhewitt  I am using ubuntu 18.04 on wsl, and python3.6.

davidhewitt:
So I can confirm that on Python 3.6 this issue reproduces on my system.

On Python 3.7 and up `sys.argv` exists but is empty, so the original reported issue no longer applies.

However, I'm unsure whether pyo3 should be setting `sys.argv` unconditionally. Perhaps we can add an API to do this.

davidhewitt:
Note for the future - the folks in `pybind11` look like they might merge something similar to this: https://github.com/pybind/pybind11/pull/2341

The design of their implementation would take arguments passed from c++ `main`, which isn't quite how Rust does it. There's also a question of whether we would ever want something like this for the `auto-initialize` feature.

Additionally, it's clearly less of a problem once we move away from supporting Python 3.6, because after that point accessing `sys.argv` stops being an `AttributeError`.

PR #1247, opened 2020-10-18
ffi: add bindings for PEP 587

davidhewitt:
While investigating #1241 I noticed we didn't have bindings for these APIs, so I added them!


alex:
What do you think about waiting to merge this until after abi3, so as to hopefully avoid more merge conflicts on that branch?

davidhewitt:
Sure thing, I'm happy to wait. Next big PR I'm working on is the drop of Python 3.5 support 😉

Issue #219, opened 2018-09-11
Incorrect #[cfg] before PyEval_InitThreads breaks Python::acquire_gil() in standalone scripts on <= 3.6


joar:
## 🐛 Bug Reports

When reporting a bug, please provide the following information. If this is not a bug report you can just discard this template.

### 🌍 Environment

 - Your operating system and version: Ubuntu 18.04
 - Your python version: Python 3.6 in this case
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?:
   - Python is installed via apt.
   - I use python's `venv`.
 - Your rust version (`rustc --version`): `rustc 1.30.0-nightly (2d4e34ca8 2018-09-09)`

### 💥 Reproducing

Check out the code in the fork at https://github.com/joar/pyo3/tree/bug/pyeval-threadsnotinitialized/examples/verify_gil.

`cd examples/verify_gil`

if you haven't already

```console
$ python -VV
Python 3.6.5 (default, Apr  1 2018, 05:46:30) 
[GCC 7.3.0]

$ python setup.py develop
[...]

$ cat ../../target/debug/build/pyo3-4833970e0e0d17d0/output
cargo:rustc-cfg=Py_3_5
cargo:rustc-cfg=Py_3_6
cargo:rustc-cfg=Py_3
cargo:rustc-cfg=py_sys_config="WITH_THREAD"
cargo:python_flags=FLAG_WITH_THREAD=1,CFG_Py_3_5,CFG_Py_3_6
cargo:rerun-if-env-changed=LD_LIBRARY_PATH
cargo:rerun-if-env-changed=PATH
cargo:rerun-if-env-changed=PYTHON_SYS_EXECUTABLE
cargo:rerun-if-env-changed=LIB

$ python naked-script.py
thread '<unnamed>' panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`', src/pythonrun.rs:42:13
note: Run with `RUST_BACKTRACE=1` for a backtrace.
fatal runtime error: failed to initiate panic, error 5
fish: “python naked-script.py” terminated by signal SIGABRT (Abort)
```

*Note:* Importing the script via ipython still works. I supect that is because some other extension that ipython uses calls `PyEval_InitThreads`.

### :tada: Solution?

[...]

PR #1249, opened 2020-10-18
Remove unneeded test

davidhewitt:
I couldn't figure a reason for this test existing, so I've removed.


davidhewitt:
EDIT: actually there seems to be some merit based on the discussion in #219 for a test like this which simply calls acquire_gil, but this test could benefit from a comment and also being housed somewhere other than the `test_datetime` test suite. I'll tidy up another evening.

kngwyu:
I agree to move this test.

davidhewitt:
Done so in #1268.

PR #1250, opened 2020-10-18
Drop support for Python 3.5

### LINKED ISSUE #208: Bounds checking in Python <=3.5
There are a slew of `xfail` tests in both the [`rustapi_module`](https://github.com/PyO3/pyo3/blob/master/examples/rustapi_module/tests/test_datetime.py#L73) and [`test_datetime.rs`](https://github.com/PyO3/pyo3/blob/master/tests/test_datetime.rs#L142) because the Python C API only started doing bounds checking on inputs in Python 3.6.

It *may* be that an exception is set in Python <=3.5 and it's just never actually *raised*, but I think more likely you get a malformed date/time object.

If no exception is set, I believe we should fix this by doing bounds checking ourselves on the Rust side (`objects/datetime.rs`), probably with a macro under a conditional compilation flag.

### LINKED ISSUE #685: Drop support for python3.5
python3.5 has been "security fixes only" for a while: https://www.python.org/downloads/release/python-356/

Support for it should at some point be dropped.

There are 45 `cfg` gating on python3.6+ , 14 on 3.7+, and 9 on 3.8+, with differences in datetime, asyncio, modules, and the FFI. It would be nice to reduce complexity.

The last few PyPy releases have upgraded to 3.6.

davidhewitt:
This drops support for Python 3.5, as notionally suggested in #685 . (Python 3.5 is now end-of-life.)

This also helps unblock the abi3 work, which currently has a segmentation fault under Python 3.5 on Windows which looks like it'll be very hard to debug.

Closes #208
Closes #685


build.rs
```diff
@@ -770,7 +770,7 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<String> {
     }
 
     if let Some(minor) = interpreter_config.version.minor {
-        for i in 5..(minor + 1) {
+        for i in 6..(minor + 1) {
```
kngwyu, line 773: :slightly_smiling_face:

kngwyu (approved):
Thanks. I'm happy to see that dropping this reduces LOC.

build.rs
programmerjake: you could have used `6..=minor` instead of `6..(minor + 1)`

davidhewitt: :+1: good idea!

PR #1254, opened 2020-10-20
Renew #[pyproto] implementation for abi3 branch

kngwyu:
So what's the problem?
Currently, on the master, `#[pyproto]` is implemented via 'table initialization'.
I.e., `#[pyproto]` generates basically such a function:
```rust
#[ctor] 
fn __table_initialization() {
    let mut methods = PyNumberMethods::default();
    methods.set_add::<MyClass>();
    <MyClass as HasProtoRegistory>::add_number_methods(methods);  // I don't remember the function name correctly...
}
```
I designed this since many protocols required such kinds of tables before. 
However, when we are to use ABI3, most protocols are initialized via `PyType_Slot` and tables are unnecessary.
So I decided to renew the whole design, using `Vec<PyType_Slot>` instead of tables.
I believe the new implementation is simpler but has one disadvantage: weak typing.
`PyType_Slot` casts function pointers to `*mut c_void` so the function type is eliminated, which can lead to bugs. We should carefully check the `slot` and `pfunc` of each `PyType_Slot` are correct.

Now `#[pyproto]` generates (roughly):
```rust
pyo3::inventory::submit!( {
    let mut proto_methods = vec![];
    proto_methods.push(<MyClass as PyNumberSlots>::get_add());
    proto_methods.into()
});
```

### DIFF
--- a/pyo3-derive-backend/src/pyproto.rs
+++ b/pyo3-derive-backend/src/pyproto.rs
@@ -106,16 +106,16 @@ fn impl_proto_impl(
             }
         }
     }
-    let inventory_submission = inventory_submission(py_methods, ty);
-    let slot_initialization = slot_initialization(method_names, ty, proto)?;
+    let normal_methods = submit_normal_methods(py_methods, ty);
+    let protocol_methods = submit_protocol_methods(method_names, ty, proto)?;
     Ok(quote! {
         #trait_impls
-        #inventory_submission
-        #slot_initialization
+        #normal_methods
+        #protocol_methods
     })
 }
 
-fn inventory_submission(py_methods: Vec<TokenStream>, ty: &syn::Type) -> TokenStream {
+fn submit_normal_methods(py_methods: Vec<TokenStream>, ty: &syn::Type) -> TokenStream {
     if py_methods.is_empty() {
         return quote! {};
     }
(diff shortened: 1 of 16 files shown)


alex:
I only skimmed it, but the design makes sense to me.

davidhewitt:
👍 looks like a good idea! I've started reviewing but need more time to finish reading. I have a very busy day today so tomorrow is more likely I'm afraid.

alex (approved):

davidhewitt (commented):
Sorry for the delay in reviewing. Looks good to me. I have a couple of suggestions.

> I believe the new implementation is simpler but has one disadvantage: weak typing.
PyType_Slot casts function pointers to *mut c_void so the function type is eliminated, which can lead to bugs. We should carefully check the slot and pfunc of each PyType_Slot are correct.

How about adding an enum so that when we put the proto methods in the inventory we still have strong types? See my other comment.

Also, I think we should be able to remove the special case for the buffer methods and just use slots for them too.

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -129,42 +129,46 @@ fn inventory_submission(py_methods: Vec<TokenStream>, ty: &syn::Type) -> TokenSt
...
+            let field = syn::Ident::new(&format!("bf_{}", &getter[4..]), Span::call_site());
+            tokens.push(quote! { proto_methods.#field = Some(<#ty as #ext_trait>::#get()); });
+        }
+    } else {
+        // For other protocols, we construct `PyProtoMethods` from Vec<ffi::PyType_Slot>
+        tokens.push(quote! { let mut proto_methods = vec![]; });
+        for getter in proto.slot_getters(method_names) {
+            let get = syn::Ident::new(getter, Span::call_site());
+            tokens.push(quote! { proto_methods.push(<#ty as #ext_trait>::#get()); });
+        }
```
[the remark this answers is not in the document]
davidhewitt: (and similar for the buffer method tokens just above, but I think that we can probably remove those completely.

kngwyu:
Thanks for the review!

> How about adding an enum so that when we put the proto methods in the inventory we still have strong types? 

This sounds the same as the current approach, and if so, it does not address the code complexity issue.
I think what we need is adding some assertion before we cast a function pointer to `*mut c_void`. I'd rather try to implement this using a macro and ask your review again.

src/class/proto_methods.rs
```diff
@@ -1,160 +1,74 @@
...
 
-/// Defines all method tables we need for object protocols.
 // Note(kngwyu): default implementations are for rust-numpy. Please don't remove them.
 pub trait PyProtoMethods {
-    fn async_methods() -> Option<NonNull<PyAsyncMethods>> {
-        None
-    }
-    fn basic_methods() -> Option<NonNull<PyObjectMethods>> {
-        None
+    fn get_type_slots() -> Vec<ffi::PyType_Slot> {
```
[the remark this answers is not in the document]
kngwyu: > Also this would work very nicely with the buffer, where slot definitions do exist as long as it is not limited API. https://github.com/python/cpython/blob/cb115e36e1aba04b90b0ecac6f043e60064ac65b/Include/typeslots.h#L7-L8

Nice catch, thanks!

kngwyu:
Oh, I just recalled that type slots does not work for buffer... (https://github.com/PyO3/pyo3/pull/1132#issuecomment-684218818)

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -129,42 +129,46 @@ fn inventory_submission(py_methods: Vec<TokenStream>, ty: &syn::Type) -> TokenSt
...
+            let get = syn::Ident::new(getter, Span::call_site());
+            let field = syn::Ident::new(&format!("bf_{}", &getter[4..]), Span::call_site());
+            tokens.push(quote! { proto_methods.#field = Some(<#ty as #ext_trait>::#get()); });
+        }
+    } else {
+        // For other protocols, we construct `PyProtoMethods` from Vec<ffi::PyType_Slot>
+        tokens.push(quote! { let mut proto_methods = vec![]; });
+        for getter in proto.slot_getters(method_names) {
+            let get = syn::Ident::new(getter, Span::call_site());
+            tokens.push(quote! { proto_methods.push(<#ty as #ext_trait>::#get()); });
```
davidhewitt, line 157: Maybe we can do the creation of `PyType_Slot` objects and conversion to `*mut void` here?

Then the slot getters can return well-typed function pointers.

```suggestion
            tokens.push(quote! { proto_methods.push(pyo3::ffi::PyType_Slot {
                slot: #slot,
                pfunc: <#ty as #ext_trait>::#get()
            }); });
```

We would need to move the `#slot` definitions into the proc macro. I think that could be a good idea anyway to help keep the definitions in one place.

src/class/basic.rs
```diff
@@ -133,94 +133,109 @@ pub trait PyObjectRichcmpProtocol<'p>: PyObjectProtocol<'p> {
...
-    pub nb_bool: Option<ffi::inquiry>,
-}
-
+/// Extension trait for proc-macro backend.
 #[doc(hidden)]
-impl PyObjectMethods {
-    // Set functions used by `#[pyproto]`.
-    pub fn set_str<T>(&mut self)
+pub trait PyBasicSlots {
+    fn get_str() -> ffi::PyType_Slot
```
davidhewitt, line 139: If you take my suggestion above then this should just return the typed function pointer:

```suggestion
    fn get_str() -> ffi::reprfunc {
        py_unary_func!(PyObjectStrProtocol, Self::__str__)
    }
```

And same for all other methods. I think this'll reduce the LOC quite a lot!

davidhewitt (commented):
> Oh, I just recalled that type slots does not work for buffer... (#1132 (comment))

👍 let's leave it as-is for this PR. I think it _might_ work for 3.9 - I'll experiment in a follow-up PR sometime.

> This sounds the same as the current approach, and if so, it does not address the code complexity issue.
I think what we need is adding some assertion before we cast a function pointer to *mut c_void. I'd rather try to implement this using a macro and ask your review again.

Fair point. See my new suggestion in comments below, which I think can reduce LOC quite a bit.

pyo3-derive-backend/src/pyproto.rs
kngwyu: Sounds nice :+1: !

kngwyu:
Chaned to use `struct TypedSlot<T: Sized>(pub std::os::raw::c_int, pub T)` as internal representation of `ffi::PyType_Slot` to resolve the weak typing problem.

I also tried to `slot: #slot` that @davidhewitt introduced, but found myself hard to write `"ffi::Py_*"`s without code completion.

davidhewitt (approved):
Nice, LGTM!

PR #1255, opened 2020-10-20
Add PyEval_SetProfile, PyEval_SetTrace

aviramha:
Not sure if I missed anything, I relied on existing infra to understand what needs to be done.
Feedback will be appreciated.


kngwyu:
Thanks!
Could you please consider placing those on `ceval.rs` since they are originally in `ceval.h`?
Also, it would be better if we have
```rust
pub type Py_tracefunc = extern "C" fn(
    obj: *mut PyObject,
    frame: *mut PyFrameObject,
    what: c_int,
    arg: *mut PyObject,
) -> c_int;
```
in `pystate.rs`.

aviramha:
@kngwyu Thanks for the review! I tried to do what you have suggested but have failed due to lack of Rust/FFI experience 👍 
I'll change it ASAP.

aviramha:
Done!

kngwyu (approved):
Thanks!

PR #1256, opened 2020-10-21
Docs: Fixed example word_count

Daggy1234:
No change in actual py03 code. this just fixes the word_count example.

### DIFF
--- a/examples/word-count/src/lib.rs
+++ b/examples/word-count/src/lib.rs
@@ -55,7 +55,7 @@ fn count_line(line: &str, needle: &str) -> usize {
 
 #[pymodule]
 fn word_count(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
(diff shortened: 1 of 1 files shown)


examples/word-count/src/lib.rs
```diff
@@ -55,7 +55,7 @@ fn count_line(line: &str, needle: &str) -> usize {
 
 #[pymodule]
 fn word_count(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
-    m.add_wrapped(wrap_pyfunction!(search))?;
+    m.add_wrapped(wrap_pyfunction!(search, m)?)?;
```
kngwyu, line 58: You mean ` m.add_function(wrap_pyfunction!(search, m)?)?;` instead?

Daggy1234: Whoops yeah will fix

kngwyu (approved):
Thanks!

PR #1259, opened 2020-10-22
Add context.h functions (PyContext_New, PyContext*)

aviramha:
Added context.h functions for FFI. I've limited it to Py3.8+ because API changed between 3.7 and 3.7.1 and I couldn't find any way to limit by patch and didn't want to add patch treatment just for this.


kngwyu (approved):
Thanks!

PR #1260, opened 2020-10-24
Fix PyCodeObject struct

aviramha:
Added missing field to PyCodeObject.


aviramha:
Not sure why 3.9 Ubuntu fails..

davidhewitt:
Thanks!

PR #1267, opened 2020-11-10
build: not cross-compiling when musl from gnu

### LINKED ISSUE #1264: Building with maturin in manylinux fails because of missing PYO3_CROSS_LIB_DIR
The command which is failing is at https://github.com/mozilla/pyo3-parsepatch/blob/e531232ee6760a9fe7f2bca24905fae626c51495/Dockerfile#L41 (`maturin build --target x86_64-unknown-linux-musl --manylinux off`).

```
error: failed to run custom build command for `pyo3 v0.12.3`

Caused by:
  process didn't exit successfully: `/rs_pp/target/debug/build/pyo3-a884c8252d1eb1d1/build-script-build` (exit code: 1)
  --- stderr
  Error: "Must provide PYO3_CROSS_LIB_DIR environment variable when cross-compiling"
warning: build failed, waiting for other jobs to finish...
error: build failed
💥 maturin failed
  Caused by: Failed to build a native library through cargo
  Caused by: Cargo build finished with "exit code: 101": `cargo rustc --message-format json --manifest-path Cargo.toml --lib --target x86_64-unknown-linux-musl --`
```

It used to work with pyo3 0.9.2.

davidhewitt:
Fixes #1264 

This patch allows compiling for musl on linux (and possibly a few other cases, I tried to write a general patch which was reasonable). 

I think as long as the extension module is statically linked against libc (e.g. such as in the musl case) then there is no issue loading such an extension module. I was able to test locally that building a musl extension module worked fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-### Fixed
-- Fix missing field in `PyCodeObject` struct (`co_posonlyargcount`) - caused invalid access to other fields in Python >3.7. [#1260](https://github.com/PyO3/pyo3/pull/1260)
 ### Packaging
 - Drop support for Python 3.5 (as it is now end-of-life). [#1250](https://github.com/PyO3/pyo3/pull/1250)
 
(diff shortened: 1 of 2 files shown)


kngwyu:
So we cannot use `PYO3_CROSS_INCLUDE_DIR` in this setting?
I'm not familiar with cross-compiling, but isn't this hurt some generality?
E.g., we use `PY_MAJOR_VERSION` and `Py_ENABLE_SHARED` from `PYO3_CROSS_LIB_DIR`.

davidhewitt:
We don't need `PYO3_CROSS_LIB_DIR` in this setting at all - we can use the normal `PYO3_PYTHON` (or `PATH`) in this case and extract the build configuration using `get_config_from_interpreter`.

The extension module statically linked with musl libc can be run with the normal system python - so it's not really a cross-compile. Just linking against a different libc.

kngwyu:
Ah, that makes sense, thanks.

Issue #1117, opened 2020-08-25
Refactor pyproto internals


davidhewitt:
Just a reminder of an idea to refactor pyproto internals.

See the idea in the comment from https://github.com/PyO3/pyo3/pull/1093#pullrequestreview-463806345

davidhewitt:
@dvermd following up from your Gitter question: I had some experiments but no branch which is worth pushing. 

The main idea I was playing with is that we have an enum `MethodProto` which causes a lot of the `pyproto` code to be duplicated for each enum branch.

I was wondering if something like the following would be more flexible.

```
pub struct MethodProto {
    name: &'static str,
    proto: &'static str,
    receiver: SelfType,
    args: &[&'static str],         // names of all the arguments
    result: Option<&'static str>   // name of the return type, if any (I think only `Free` does not have)
}
```

Happy to answer more questions if you're interested in working on this!

dvermd:
All the firsts fields of this new struct (`name`, `proto`, `receiver`, `args`) can be set in [defs.rs](https://github.com/PyO3/pyo3/blob/master/pyo3-derive-backend/src/defs.rs) while filling the `Proto` structs. 

`result` however seems to be taken from the [method signature](https://github.com/PyO3/pyo3/blob/master/pyo3-derive-backend/src/proto_method.rs#L77). 

Should this still be the case and the `result` field removed from the new `MethodProto` struct ?
Did you thought about "forcing" return type `Proto` structs definitions in `defs.rs` ?

kngwyu:
BTW, isn't it nice to first resolve #1206, where `Unary` is replaced by `Unary*`? I have a working branch for that, though I only have changed `basic` protocol.

davidhewitt:
> BTW, isn't it nice to first resolve #1206, where Unary is replaced by Unary*?

I think it can be done either way around - this refactoring makes it quite easy because `receiver` in the defs will control what the self type should be. After #1206 we wouldn't need `receiver` in the defs at all.

> result however seems to be taken from the method signature.
>
> Should this still be the case and the result field removed from the new MethodProto struct ?
> Did you thought about "forcing" return type Proto structs definitions in defs.rs ?

I think the link you give is for the return type - what the def needs is whether this trait has a return type at all (because there is `MethodProto::Free` which has no `Result`. So maybe a `bool` is best. Or alternatively maybe the traits with `MethodProto::Free` should be changed to allow returning either `()` or `PyResult<()>`.

[...]

PR #1270, opened 2020-11-10
refactor pyproto internals

dvermd:
This PR aims to refactor MethodProto.

`make test` passed OK

Still to do:
- [x] `Free` methods 

Closes: #1117

### DIFF
--- a/pyo3-derive-backend/src/proto_method.rs
+++ b/pyo3-derive-backend/src/proto_method.rs
@@ -1,5 +1,4 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
-use crate::utils::print_err;
 use proc_macro2::{Span, TokenStream};
 use quote::{quote, ToTokens};
 use syn::Token;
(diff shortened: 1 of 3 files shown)


pyo3-derive-backend/src/utils.rs
```diff
@@ -1,12 +1,7 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 use proc_macro2::Span;
-use proc_macro2::TokenStream;
 use std::fmt::Display;
 
-pub fn print_err(msg: String, t: TokenStream) {
```
kngwyu, line 6: Good catch :+1:

kngwyu:
Quite a nice refactoring, thanks.
I think your approach to `Free` methods is correct. They doesn't have `type Result = ...`.

How about adding some constructors to `MethodProto`? For example, `with_result: false` is only used for `GC` methods, so we can use a 'short-cut' constructor for the common case.

kngwyu:
BTW, I came up with a shorter `modify_arg_ty`:
```rust

fn modify_arg_ty(
    sig: &mut syn::Signature,
    idx: usize,
    decl1: &syn::FnArg,
    decl2: &syn::FnArg,
) -> syn::Result<()> {
    let arg = sig.inputs[idx].clone();
    match arg {
        syn::FnArg::Typed(ref cap) if crate::utils::option_type_argument(&*cap.ty).is_some() => {
            sig.inputs[idx] = fix_name(&cap.pat, &decl2.inputs[idx])?;
        }
        syn::FnArg::Typed(ref cap) => {
            sig.inputs[idx] = fix_name(&cap.pat, &decl1.inputs[idx])?;
        }
        _ => return Err(syn::Error::new_spanned(arg, "not supported")),
    }

    Ok(())
}
```

dvermd:
> How about adding some constructors to `MethodProto`? For example, `with_result: false` is only used for `GC` methods, so we can use a 'short-cut' constructor for the common case.

I thought and tried some constructors but then the declarations in `defs.rs` won't be that explicit without constructors' arguments names. If you're OK with that I can create them.

Else, there's the builder pattern that can be a solution `builder(name, proto).args(&[...]).with_self().with_result()`.

Which one do you like better ?

pyo3-derive-backend/src/proto_method.rs
```diff
@@ -1,70 +1,59 @@
...
+            args: &[],
+            with_self: false,
+            with_result: true,
+        }
+    }
+    pub const fn args(self, args: &'static [&'static str]) -> MethodProtoBuilder {
+        let mut with_args = self;
+        with_args.args = args;
+        with_args
+    }
```
davidhewitt, line 39: I think you can probably take `mut self` in each of these builder methods, e.g.

```suggestion
    pub const fn args(mut self, args: &'static [&'static str]) -> MethodProtoBuilder {
        self.args = args;
        self
    }
```

pyo3-derive-backend/src/pyproto.rs
```diff
@@ -66,6 +66,7 @@ fn impl_proto_impl(
     for iimpl in impls.iter_mut() {
         if let syn::ImplItem::Method(ref mut met) = iimpl {
             // impl Py~Protocol<'p> { type = ... }
+            // TODO: MethodProto.result = meth.sig.output
```
davidhewitt, line 69: Not sure what this TODO is asking to happen?

davidhewitt (approved):
🚀 Thanks, this looks awesome to me! Lovely to see a great reduction in LOC and overall it's much easier to understand the affected files.

The `.has_self()` builder method is really nice as it'll make it super easy to refactor the self-types later as discussed on the issue.

Regarding CHANGELOG entry, I don't think this affects users so imo it shouldn't have an entry.

davidhewitt:
Regarding MSRV failure - I think there's now maybe enough reasons for us to bump to rust 1.45, so I'm probably going to open a PR for that later.

pyo3-derive-backend/src/pyproto.rs
dvermd: This TODO was forgotten there. It's removed.

pyo3-derive-backend/src/proto_method.rs
dvermd: You're right, I didn't think `mut self` was possible but the code is much nicer this way.

davidhewitt:
Looks great, thanks! If you squash and rebase on master I think you should find that MSRV CI starts working.

I notice this PR is still in draft, but I think other than a rebase it's now hopefully ready to be merged?

dvermd:
Squashed, rebased, removed a `use` comment in `defs.rs` and un-drafted.

Still some problems with MSRV but 1.45 should contain the right commit

davidhewitt:
Ugh. 

Based on [this comment](https://github.com/rust-lang/rust/pull/73862/#issue-441477879)

>  This PR stabilizes these casts (which are already stable in const outside const fn)

I wonder if this works as a workaround:

```rust

// TODO: workaround for no unsized casts in const fn on Rust 1.45 (stable in 1.46)
const EMPTY_ARGS: &[&str] = &[];

impl MethodProtoBuilder {
        MethodProtoBuilder {
            name,
            proto,
            args: EMPTY_ARGS,
            with_self: false,
            with_result: true,
        }
    }
}
```

If you don't have a chance before I do, I might push a commit to this branch tonight to try that.

davidhewitt:
... or I guess if that doesn't work, maybe could just always take `args` as an input to the `const fn new`.

kngwyu:
Just a nit, but your `EMPTY_ARGS` can be an associated constant, like:
```rust
impl MethodProtoBuilder {
    const EMPTY_ARGS: &[&str] = &[];
...
}
```

kngwyu:
Why use `build`? 
It looks like that `MethodProtoBuilder` and `MethodProto` is the same and I don't see any reason for defining the builder as another struct.

dvermd:
> ```rust
> impl MethodProtoBuilder {
>     const EMPTY_ARGS: &[&str] = &[];
> ...
> }
> ```
This workaround works !

> Why use `build`?
> It looks like that `MethodProtoBuilder` and `MethodProto` are the same and I don't see any reason for defining the builder as another struct.

I merged the `MethodProtoBuilder` into the `MethodProto` and removed the `build()` method.

davidhewitt:
> It looks like that MethodProtoBuilder and MethodProto are the same and I don't see any reason for defining the builder as another struct.

👍 good spot

> I merged the MethodProtoBuilder into the MethodProto and removed the build() method.

🎉 this refactoring is really great; it's reduced LOC by just over 50% 2 for this part of the codebase. Thanks so much for your work on this!

PR #1271, opened 2020-11-12
coverage: simplify mechanism for panic=unwind tests

davidhewitt:
I recently added `#[cfg(panic = ...)]` to nightly rust, and wanted to use it here but it turned out to be a little awkward. So I simplified the way to check for `panic=abort` in tests which care about it while we wait for `#[cfg(panic = ...)]` to become stable.

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -526,7 +526,6 @@ fn exceptions_must_derive_from_base_exception(py: Python) -> PyErr {
 mod tests {
     use super::PyErrState;
     use crate::exceptions;
-    use crate::panic::PanicException;
     use crate::{PyErr, Python};
 
     #[test]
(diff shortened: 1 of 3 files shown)


kngwyu:
> I recently added #[cfg(panic = ...)] to nightly rust,

Where's the commit?

And should this be done in the build.rs?

davidhewitt:
https://github.com/rust-lang/rust/commit/46bce9f8efbd4c57a23e822588650250d1e74b26

> And should this be done in the build.rs?

Once `cfg_panic` is stable we won't need anything in `build.rs`. For the moment I moved the logic of checking `RUSTFLAGS` there so that we don't repeat it in all relevant tests.

kngwyu:
Then I'd rather pack the logic to the test code by a function or macro.

davidhewitt:
Changed to a `cfg_panic_unwind` helper function. It's now slightly increased LOC compared to the original master, but I think it's ok because it notes the cleanup path once we get the stable attribute.

PR #1272, opened 2020-11-12
msrv: bump to 1.45

davidhewitt:
Bumps minimum supported Rust version to 1.45.

I think there's a few places now where the older Rust version is beginning to hold us up, and also RHEL now supports Rust 1.45: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/8.3_release_notes/overview

From #1269 I discovered that maturin needs MSRV 1.40
In #1270 the usage of `const fn` there needs a slightly newer compiler than 1.39 (not sure exactly which one).

Closes #1252 
Closes #1262

### DIFF
--- a/pyo3-derive-backend/src/pymethod.rs
+++ b/pyo3-derive-backend/src/pymethod.rs
@@ -505,11 +505,10 @@ fn impl_arg_param(
             // Get Option<&T> from Option<PyRef<T>>
             (
                 quote! { Option<<#tref as pyo3::derive_utils::ExtractExt>::Target> },
-                // To support Rustc 1.39.0, we don't use as_deref here...
                 if mut_.is_some() {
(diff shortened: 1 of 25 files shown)


pyo3-derive-backend/src/pymethod.rs
```diff
@@ -712,13 +711,8 @@ pub(crate) fn impl_py_getter_def(
...
-    if args
-        .get(0)
-        .map(|py| utils::is_python(&py.ty))
-        .unwrap_or(false)
-    {
-        (Some(&args[0]), &args[1..])
-    } else {
-        (None, args)
+    match args {
+        [py, args @ ..] if utils::is_python(&py.ty) => (Some(py), args),
```
kngwyu, line 715: :+1:

kngwyu (approved):

kngwyu (approved):

PR #1275, opened 2020-11-14
pyerr: improve debug & display impls

davidhewitt:
This PR aims to improve the ergonomics of `PyErr`'s debug output. 

Now, instead of `PyErr { type: 0x... }` the repr of type and value are printed (using `Python::with_gil` to acquire the GIL if needed). We already acquire the GIL in the `Display` impl so I think we have acceptable precedent to do this. It also makes the output much more useful for PyO3 users.

As a simplification, I also removed the special `impl Debug` and `impl Display` for `PyException`. They now just use the underlying Python object's `.repr()` and `.str()`, just like `PyAny`. I think there's a strong consistency argument for doing this.

Before:

```
PyErr debug         = PyErr { type: 0x7ff6f7b50420 }
PyErr display       = ValueError: bar
PyException debug   = ValueError
PyException display = ValueError: bar
PyAny debug         = ValueError('bar')
PyAny display       = bar
```

After:

```
PyErr debug         = PyErr { type: <class 'ValueError'>, value: ValueError('bar'), traceback: None }
PyErr display       = ValueError: bar
PyException debug   = ValueError('bar')
PyException display = bar
PyAny debug         = ValueError('bar')
PyAny display       = bar
```

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -176,7 +176,8 @@ impl PyErr {
     /// use pyo3::{Python, PyErr, exceptions::PyTypeError, types::PyType};
     /// Python::with_gil(|py| {
     ///     let err = PyTypeError::new_err(("some type error",));
-    ///     assert_eq!(err.pvalue(py).to_string(), "TypeError: some type error");
+    ///     assert!(err.is_instance::<PyTypeError>(py));
+    ///     assert_eq!(err.pvalue(py).to_string(), "some type error");
     /// });
     /// ```
     pub fn pvalue<'py>(&'py self, py: Python<'py>) -> &'py PyBaseException {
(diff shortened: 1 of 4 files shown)


kngwyu (approved):
Awesome :100:

src/err/mod.rs
```diff
@@ -559,6 +574,31 @@ mod tests {
...
     }
 
+    #[test]
+    fn err_debug() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let err = py
+            .run("raise Exception('banana')", None, None)
+            .expect_err("raising should have given us an error");
+        assert!(
```
kngwyu, line 584: Looks this assertion fails at Python 3.6. Maybe you can just weaken the assertion.

davidhewitt: 👍 modified slightly to check the individual fields and use `#[cfg(Py_3_7)]` on the bit that's different

PR #1276, opened 2020-11-15
pyany: add is_instance

davidhewitt:
Adds `PyAny::is_instance::<T>()`.

This is an alternative to `Python::is_instance()` which offers slightly nicer syntax. I think there's a good consistency argument to adding this:
- We expose other builtins like `dir()`, `hash()`, `len()` as `PyAny::dir` instead of `Python::dir` - `is_instance` has been a special case.
- It's also consistent with `PyErr::is_instance::<T>(py)`.

To encourage only one way to do things, I'm very tempted to mark `#[deprecated]` on `Python::is_instance` and `Python::is_subclass` (which is already redundant to `PyType::is_subclass`). What do you think of this?

Comparing from the documentation:

`Python::is_instance`:

```
fn main() {
    let py = gil.python();
    assert!(py.is_instance::<PyBool, _>(PyBool::new(py, true)).unwrap());
    let list = PyList::new(py, &[1, 2, 3, 4]);
    assert!(!py.is_instance::<PyBool, _>(list).unwrap());
    assert!(py.is_instance::<PyList, _>(list).unwrap());
}
```

`PyAny::is_instance`:

```
fn main() {
    let py = gil.python();
    assert!(PyBool::new(py, true).is_instance::<PyBool>().unwrap());
    let list = PyList::new(py, &[1, 2, 3, 4]);
    assert!(!list.is_instance::<PyBool>().unwrap());
    assert!(list.is_instance::<PyList>().unwrap());
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add FFI definitions for PEP 587 "Python Initialization Configuration". [#1247](https://github.com/PyO3/pyo3/pull/1247)
 - Add `PyEval_SetProfile` and `PyEval_SetTrace` to FFI. [#1255](https://github.com/PyO3/pyo3/pull/1255)
 - Add context.h functions (`PyContext_New`, etc) to FFI. [#1259](https://github.com/PyO3/pyo3/pull/1259)
+- Add `PyAny::is_instance()` method. [#1276](https://github.com/PyO3/pyo3/pull/1276)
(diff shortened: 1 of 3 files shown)


davidhewitt:
cc @kngwyu - as this is a new public API that would be added to `PyAny`, I'd like to check you're ok with this before I merge.

kngwyu (approved):

kngwyu:
Nice catch, thanks.

> I'm very tempted to mark #[deprecated] on Python::is_instance and Python::is_subclass

:+1:

PR #1278, opened 2020-11-15
pytypeinfo: rename is_instance to is_type_of

davidhewitt:
While working on #1276 I realised that `PyTypeInfo::is_instance` is actually a bit misleading.

The default implementation of it uses `PyObject_TypeCheck`, which is _similar_ to `isinstance()` but doesn't call the `__isinstance__` hook.

So I think it's better to make this name clearer. I went for `check` and `check_exact`, inspired by `PyObject_TypeCheck`.

This is a little breaking, but I think in practice this should only affect people who are manually implementing `PyTypeInfo` (hopefully very few users).


alex:
What do you think about `type_check` and `type_check_exact`?

davidhewitt:
Yeah that's probably even clearer!

kngwyu:
I understand the motivation, but I still prefer the `is_` prefix.

davidhewitt:
> I understand the motivation, but I still prefer the is_ prefix.

I do agree `is_` is nicer. I'd really like to move away from `is_instance` now that I'm aware this is subtly different to the other `is_instance` functions. Otherwise I think we're asking for people to write bugs by accident!

How about `T::is_type_of` and `T::is_exact_type_of` ?

I think that reads quite naturally - e.g. `PyList::is_type_of(list)` compared to the original `PyList::is_instance(list)`.

davidhewitt:
Changed to `is_type_of` and `is_exact_type_of`. @kngwyu would you be ok with these names?

kngwyu (approved):
Looks nice :+1:

PR #1280, opened 2020-11-15
wrap_pyfunction: allow working on all crate functions

davidhewitt:
This tweak allows for importing pyfunctions across modules in crates using glob imports.

e.g.

`functions.rs`

```
#[pyfunction]
fn foo() { }
```

`module.rs`

```
use functions::*;  // this will now pull in the pub(crate) generated function wrapper

// ...
wrap_pyfunction!(foo, m)?;
```

It's still not very user-friendly, so I don't really want to document this, but some users might want to do it, and I don't see why not. (I am aware of at least one person who's asked me offline about this.)

I think it might also be helpful to have support for this internally for the proposed pymodule syntax in #694 .


kngwyu (approved):
Thanks!

PR #1282, opened 2020-11-19
Char support

### LINKED ISSUE #1281: Add support for Rust char type
Currently it doesn't seem to be possible to wrap a function accepting a char as an argument using PyO3. A char is a (particular kind of) unicode code point, so it could be converted to a single-character Python string. I think this would be a useful feature for users wanting to wrap such functions.

kngwyu:
Resolve #1281. Support char and `PyString` conversion.
I'm sorry this PR includes some unrelated cleanups. Since recently I don't have much time for PyO3, I think it would be better to do some cleanups when I find the necessity.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -174,10 +174,7 @@ pub trait FromPyObject<'source>: Sized {
 
 /// Identity conversion: allows using existing `PyObject` instances where
 /// `T: ToPyObject` is expected.
-impl<'a, T: ?Sized> ToPyObject for &'a T
-where
-    T: ToPyObject,
-{
+impl<'a, T: ?Sized + ToPyObject> ToPyObject for &'a T {
     #[inline]
     fn to_object(&self, py: Python) -> PyObject {
         <T as ToPyObject>::to_object(*self, py)
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -174,7 +174,7 @@ pub trait FromPyObject<'source>: Sized {
 
 /// Identity conversion: allows using existing `PyObject` instances where
 /// `T: ToPyObject` is expected.
-impl<'a, T: ?Sized + ToPyObject> ToPyObject for &'a T {
+impl<T: ?Sized + ToPyObject> ToPyObject for &'_ T {
     #[inline]
     fn to_object(&self, py: Python) -> PyObject {
         <T as ToPyObject>::to_object(*self, py)
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -177,16 +177,16 @@ impl<'source> FromPyObject<'source> for &'source str {
 
 /// Allows extracting strings from Python objects.
 /// Accepts Python `str` and `unicode` objects.
-impl<'source> FromPyObject<'source> for String {
-    fn extract(obj: &'source PyAny) -> PyResult<Self> {
+impl FromPyObject<'_> for String {
+    fn extract(obj: &PyAny) -> PyResult<Self> {
         <PyString as PyTryFrom>::try_from(obj)?
             .to_str()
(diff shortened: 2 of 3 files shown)


src/conversion.rs
```diff
@@ -174,10 +174,7 @@ pub trait FromPyObject<'source>: Sized {
 
 /// Identity conversion: allows using existing `PyObject` instances where
 /// `T: ToPyObject` is expected.
-impl<'a, T: ?Sized> ToPyObject for &'a T
-where
-    T: ToPyObject,
-{
+impl<'a, T: ?Sized + ToPyObject> ToPyObject for &'a T {
```
davidhewitt, line 177: 👍 nice! I think the lifetime can maybe also be anonymous

```suggestion
impl<T: ?Sized + ToPyObject> ToPyObject for &'_ T {
```

src/types/string.rs
```diff
@@ -172,6 +185,20 @@ impl<'source> FromPyObject<'source> for String {
     }
 }
 
+impl<'source> FromPyObject<'source> for char {
+    fn extract(obj: &'source PyAny) -> PyResult<Self> {
```
davidhewitt, line 189: Similarly can skip the lifetime here:

```suggestion
impl FromPyObject<'_> for char {
    fn extract(obj: &PyAny) -> PyResult<Self> {
```

src/types/string.rs
```diff
@@ -172,6 +185,20 @@ impl<'source> FromPyObject<'source> for String {
...
+        let s = PyString::try_from(obj)?.to_str()?;
+        let mut iter = s.chars();
+        if let (Some(ch), None) = (iter.next(), iter.next()) {
+            Ok(ch)
+        } else {
+            Err(crate::exceptions::PyValueError::new_err(format!(
+                "Expected a sting of length 1",
```
davidhewitt, line 196: typo `sting` -> `string`, and I think the recommendation in https://github.com/PyO3/pyo3/pull/1212#issuecomment-707557245 was that it's better for us to make errors start with lowercase
```suggestion
                "expected a string of length 1",
```

src/types/string.rs
```diff
@@ -180,80 +207,90 @@ mod test {
...
+            assert_eq!(s, s2);
+        })
+    }
 
-        let s2: &str = FromPyObject::extract(py_string.as_ref(py)).unwrap();
-        assert_eq!(s, s2);
+    #[test]
+    fn test_extract_char() {
+        Python::with_gil(|py| {
+            let ch = '😃';
```
davidhewitt, line 231: 😃

davidhewitt (approved):
👍 looks great, thanks! I made some minor nitpicks, only one I'd definitely like to see is the error message formatting.

Also CHANGELOG entry 😉

davidhewitt:
I'm working on a PR to fix the clippy errors with new Rust 1.48.

src/types/string.rs
```diff
@@ -180,80 +207,90 @@ mod test {
...
+            let py_string = ch.to_object(py);
+            let ch2: char = FromPyObject::extract(py_string.as_ref(py)).unwrap();
+            assert_eq!(ch, ch2);
```
birkenfeld, line 234: Is there a test for the error case (string with len > 1)?

davidhewitt: 👍 good idea

davidhewitt:
Rebase on master to fix CI (see #1284).

src/conversion.rs
kngwyu: :+1:

src/types/string.rs
kngwyu: Thanks!

davidhewitt (approved):
💯

PR #1286, opened 2020-11-19
Fix #1285, text_signature and raw ident interaction

### LINKED ISSUE #1285: text_signature and raw identifiers don't interact well
## 🐛 Bug Reports

The `text_signature` attributes automatically sets the first line of the python documentation to the function signature, which is supposed to be `fnname(arg1, arg2)`. However, this seems to interact badly with rust's raw identifiers. For instance, a methods defined like this:

```rust
#[pymethods]
impl PythonStuff {
    #[text_signature = "()"]
    fn r#match(&self) {}
}
```

Will have a signature of `r#match()` instead of the correct `match()`. This breaks several tools, like sphinx documentation generation.

wrongly documents match's signature as `r#match()` instead of simply `match()`. Python script showcasing the issue:

```python
>>> import pyo3_bug
>>> help(pyo3_bug.PythonStuff.match)
Help on method_descriptor:

match(...)
    r#match()
    --
```

### 🌍 Environment

 - Your operating system and version: Ubuntu 20
 - Your python version: Python 3.8.5
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: apt, yes
 - Your Rust version (`rustc --version`): 1.47.0
 - Your PyO3 version: 0.12
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: yes, same issue.

### 💥 Reproducing

Here's a quick script that will quickly reproduce the bug. It creates a simple pyo3 module with the `r#match` method and tries to access its help text.

```bash
cargo new --lib pyo3_bug
cd pyo3_bug
echo 'pyo3 = { version = "0.12", features = ["extension-module"] }' >> Cargo.toml
echo '[lib]' >> Cargo.toml
echo 'crate-type = ["rlib", "cdylib"]' >> Cargo.toml
cat > src/lib.rs << HERE
use pyo3::prelude::*;
#[pymodule]
fn pyo3_bug(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<PythonStuff>()?;
    Ok(())
}
#[pyclass]
struct PythonStuff {}
#[pymethods]
impl PythonStuff {
    #[text_signature = "()"]
    fn r#match(&self) {}
}
HERE
maturin develop
cat << HERE | python3
import pyo3_bug
help(pyo3_bug.PythonStuff.match)
HERE
```

This script should eventually show:

```
Help on method_descriptor:

match(...)
    r#match()
    --
```

roblabla:
Fixes #1285

Fix ended up being simple: `parse_text_signature_attrs` took the raw name instead of the pythonized name. I added some simple tests to make sure we avoid regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fix missing field in `PyCodeObject` struct (`co_posonlyargcount`) - caused invalid access to other fields in Python >3.7. [#1260](https://github.com/PyO3/pyo3/pull/1260)
 - Fix building for `x86_64-unknown-linux-musl` target from `x86_65-unknown-linux-gnu` host. [#1267](https://github.com/PyO3/pyo3/pull/1267)
+- Fix `#[text_signature]` interacting badly with rust `r#raw_identifiers`. [#1286](https://github.com/PyO3/pyo3/pull/1286)
 
 ## [0.12.3] - 2020-10-12
 ### Fixed
--- a/pyo3-derive-backend/src/method.rs
+++ b/pyo3-derive-backend/src/method.rs
@@ -237,7 +237,7 @@ impl<'a> FnSpec<'a> {
 
         let text_signature = match &fn_type {
             FnType::Fn(_) | FnType::FnClass | FnType::FnStatic => {
-                utils::parse_text_signature_attrs(&mut *meth_attrs, name)?
+                utils::parse_text_signature_attrs(&mut *meth_attrs, &python_name)?
             }
             FnType::FnNew => parse_erroneous_text_signature(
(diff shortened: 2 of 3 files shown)


davidhewitt (approved):
👍 Thanks very much for fixing this, will merge shortly assuming CI passes!

davidhewitt:
Hmm looks like the test isn't passing? 🤔

roblabla:
I got no clue what's going on, the test passes when locally running `cargo test test_raw_identifiers`.

davidhewitt:
Ahh I think `__text_signature__` doesn't work with abi3 feature (until Python 3.10, where it might start working).

I've pushed a tweak to your commit which just ignores the test in that situation.

kngwyu (approved):
Wow, what a simple fix :heart:

kngwyu:
Thanks @roblabla  and @davidhewitt

Issue #684, opened 2019-12-08
Vectorcall support


kngwyu:
Python 3.8 has a faster `__call__` protocol for C-FFI namd [vectorcall](https://www.python.org/dev/peps/pep-0590/).
#641 added FFI functions for it, but we need more high level functions implemented.

davidhewitt:
@kngwyu I'd like to take this on

kngwyu:
Thank you, but I don't think we should change all things to use vectorcall now.
How about starting from adding a `call_*` variant to `ObjectProtocol` or so?

davidhewitt:
Agreed. I plan to run a few experiments before I make any final PRs. Just wanted to give you a heads-up before I go too far so we don't end up duplicating the investigation!

davidhewitt:
Based on what I've been reading, I think that there's a few places we can potentially use vectorcall:

- `call_vectorcall` to `ObjectProtocol` is a first point to start.
- `PyObject_Vectorcall` internally uses the old calling convention if vectorcall isn't available, so in the long run I think we want to make an API where the users of pyo3 don't have to care about vectorcall, and we just dispatch the optimisation for them if they're on Python >= 3.8.
- Also, it looks like our PyCFunctions defined with `METH_VARARGS` will not support vectorcall - so we should be changing these to be `METH_FASTCALL`. We will only want to do this for Python >= 3.8, so it will take some refactoring, but should lead to good speedups for users on those Python versions.

Before this goes anywhere though, I've hit a stumbling block that `_PyObject_Vectorcall` doesn't actually seem to be exported in `libpython`. It's defined as a `static inline` function in CPython's `abstract.h`.

I've asked about that in [bpo-39773](https://bugs.python.org/issue39773).

davidhewitt:
>I've asked about that in bpo-39773.

A follow up to that. `PyObject_Vectorcall` is not actually exposed in `libpython`. This is by design as they want is as a performance optimisation for C.

This means that at the very least the wrappers added in #641 are not quite correct, because they're defining symbols which can't be linked to. I'll come up with a PR sometime to clean that up.

alex:
I think this can be closed now, yeah?

davidhewitt:
Yes, we use vectorcall in both directions. There are cases where we can probably get more efficient (e.g. we always pass keywords as a dict) bit that conversation can continue in #4503 and related threads.

PR #1287, opened 2020-11-19
vectorcall: fix bindings & use for call0 and call_method0

### LINKED ISSUE #1283: Fix Vectorcall Bindings
Looks like these are not implemented correctly and are now breaking CI with Rust 1.48.  https://github.com/PyO3/pyo3/pull/1282/checks?check_run_id=1425446591

I'm going to disable them in an initial PR and then fix them in a follow-up. We should fix before 0.13 release.

davidhewitt:
This PR fixes the broken vectorcall ffi bindings, and proves that the new ones work by using them in `.call0` and `.call_method0` implementations (where possible).

This yields quite a tidy ~20% performance speedup on those functions, because it no longer creates an empty `PyTuple` and `PyDict` each time:

before: 

```
test bench_call_0        ... bench:      50,712 ns/iter (+/- 5,555)
test bench_call_method_0 ... bench:     164,581 ns/iter (+/- 10,141)
```

after:

```
test bench_call_0        ... bench:      40,996 ns/iter (+/- 4,490)
test bench_call_method_0 ... bench:     138,826 ns/iter (+/- 18,465)
```

I'd love to follow this PR up with some ways to use vectorcall for functions with arguments too. Will leave that for #684 instead of pushing this further right now.

Closes #1283

### DIFF
--- a/src/ffi/cpython/abstract_.rs
+++ b/src/ffi/cpython/abstract_.rs
@@ -1,0 +1,306 @@
+use crate::ffi::{PyObject, Py_TYPE, Py_buffer, Py_ssize_t};
+use libc::{c_char, c_int, c_void};
+
+#[cfg(all(Py_3_8, not(PyPy)))]
+use crate::ffi::{
+    vectorcallfunc, PyCallable_Check, PyThreadState, PyThreadState_GET, PyTuple_Check,
+    PyType_HasFeature, Py_TPFLAGS_HAVE_VECTORCALL,
+};
+#[cfg(all(Py_3_8, not(PyPy)))]
+use libc::size_t;
+
+extern "C" {
+    #[cfg(all(Py_3_8, not(PyPy)))]
+    pub fn _PyStack_AsDict(values: *const *mut PyObject, kwnames: *mut PyObject) -> *mut PyObject;
+}
+
+#[cfg(all(Py_3_8, not(PyPy)))]
+const _PY_FASTCALL_SMALL_STACK: size_t = 5;
+
+extern "C" {
+    #[cfg(all(Py_3_8, not(PyPy)))]
+    pub fn _Py_CheckFunctionResult(
+        tstate: *mut PyThreadState,
+        callable: *mut PyObject,
+        result: *mut PyObject,
+        where_: *const c_char,
+    ) -> *mut PyObject;
+
+    #[cfg(all(Py_3_8, not(PyPy)))]
+    pub fn _PyObject_MakeTpCall(
+        tstate: *mut PyThreadState,
+        callable: *mut PyObject,
+        args: *const *mut PyObject,
+        nargs: Py_ssize_t,
+        keywords: *mut PyObject,
+    ) -> *mut PyObject;
+}
+
+#[cfg(all(Py_3_8, not(PyPy)))]
+const PY_VECTORCALL_ARGUMENTS_OFFSET: Py_ssize_t =
+    1 << (8 * std::mem::size_of::<Py_ssize_t>() as Py_ssize_t - 1);
+
+#[cfg(all(Py_3_8, not(PyPy)))]
+#[inline(always)]
+pub unsafe fn PyVectorcall_NARGS(n: size_t) -> Py_ssize_t {
+    assert!(n <= (Py_ssize_t::MAX as size_t));
+    (n as Py_ssize_t) & !PY_VECTORCALL_ARGUMENTS_OFFSET
+}
+
+#[cfg(all(Py_3_8, not(PyPy)))]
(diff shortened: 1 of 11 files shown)


davidhewitt:
Windows failing to link on abi3 - presumably it's us doing something wrong; I'm asking about it at https://bugs.python.org/issue42415

davidhewitt:
Rather than wait for that bug to be resolved (which I guess might not happen until Python 3.9.1) I've put a TODO in the code and just enabling the optimization for the unlimited API for now.

src/ffi/cpython/abstract_.rs
```diff
@@ -0,0 +1,304 @@
...
+    _PyObject_VectorcallTstate(PyThreadState_GET(), callable, args, nargsf, kwnames)
+}
+
+extern "C" {
+    #[cfg(all(Py_3_8, not(PyPy)))]
+    #[cfg_attr(not(Py_3_9), link_name = "_PyObject_VectorcallDict")]
+    pub fn PyObject_VectorcallDict(
+        callable: *mut PyObject,
+        args: *const *mut PyObject,
+        nargsf: usize,
```
kngwyu, line 104: `libc::size_t` would be better?

src/ffi/cpython/abstract_.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub fn PyVectorcall_Call(
+        callable: *mut PyObject,
+        tuple: *mut PyObject,
+        dict: *mut PyObject,
+    ) -> *mut PyObject;
+}
+
+#[cfg(all(Py_3_8, not(PyPy)))]
+#[inline(always)]
+pub unsafe fn _PyObject_FastCallTstate(
```
kngwyu, line 119: Where is this function? (I'm looking at https://github.com/python/cpython/blob/v3.8.6/Include/cpython/abstract.h ).

src/ffi/cpython/abstract_.rs
```diff
@@ -0,0 +1,304 @@
...
+        2 | PY_VECTORCALL_ARGUMENTS_OFFSET as usize,
+        std::ptr::null_mut(),
+    )
+}
+
+// skipped _PyObject_VectorcallMethodId
+// skipped _PyObject_CallMethodIdNoArgs
+// skipped _PyObject_CallMethodIdOneArg
+
+// skipped _PyObject_HasLen
```
kngwyu, line 206: Just curious, why are they skipped?

src/ffi/cpython/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod abstract_;
```
kngwyu, line 1: :+1:  for creating this new module. Maybe we can reduce `cfg` using this.

src/instance.rs
```diff
@@ -272,7 +272,18 @@ impl<T> Py<T> {
     ///
     /// This is equivalent to the Python expression `self()`.
     pub fn call0(&self, py: Python) -> PyResult<PyObject> {
-        self.call(py, (), None)
+        cfg_if::cfg_if! {
+            // TODO: Use PyObject_CallNoArgs instead after https://bugs.python.org/issue42415,
+            // at which point can enable optimization for limited API.
```
kngwyu, line 277: > at which point can enable

I'm sorry but I don't understand the grammar (caveat: I'm not a native English speaker, so please just ignore if I'm wrong)

kngwyu (changes requested):
Thanks. This is a pretty good step towards vectorizing function calls. Comments are mainly around FFI definitions.

src/ffi/cpython/abstract_.rs
davidhewitt: I'm looking at the same file but the version in `master` (so `3.10-dev`). I figured it was best for these inline functions to be as up-to-date as possible.

src/ffi/cpython/abstract_.rs
davidhewitt: I didn't consider them relevant to my PR; I could add them but I'd also quite like to keep as few unlimited API definitions as possible (as they're more likely for us to be work to maintain).

See also #1289 which talks a bit about what I'm thinking.

src/ffi/cpython/mod.rs
davidhewitt: Yes, I think for sure (again see #1289).

src/instance.rs
davidhewitt: Yes, this is me being lazy; I've skipped the subject. I'll tidy this up 😄

kngwyu (approved):

kngwyu:
Thanks!

PR #1291, opened 2020-11-21
Reintroduce PyPy for Windows in CI

### LINKED ISSUE #1216: PyPy CI on Windows
Now it's disabled by #1215 but should be enabled again.

AleksaC:
Fixes #1216 

The  reason tox is failing with pypy on Windows is because `os.path.exists('')` returns `True` in that version of pypy which causes things to break during interpreter discovery. I found a wokaround using `--discover` which basically points tox to the interpreter. Once the next release of pypy lands in azure pipelines the issue will be fixed and `--discover` can be removed.


kngwyu (approved):
Nice, thank you very much!

PR #1292, opened 2020-11-22
python: deprecate some redundant methods

davidhewitt:
This PR deprecates `Python::is_instance`, `Python::is_subclass`, `Python::release`, and `Python::xdecref`. (Follow up to #1276 plus a little extra.)

All these methods have simpler alternatives, so I see no reason for us to keep them around.

I moved all the deprecated methods to the bottom and also moved some `unsafe` functions lower in the file, which will put the methods we actually want users to read at the top of the docs.


kngwyu (approved):
Thanks!

PR #1294, opened 2020-11-24
readme: add pythonize

davidhewitt:
I think enough people are asking about serde on gitter recently that it'd be nice to make `pythonize` a little bit easier for people to find and experiment with.


kngwyu (approved):

kngwyu:
Thanks!

PR #1295, opened 2020-11-25
Allow `#[pyclass(name = "string literal")]`

scalexm:
This is also more in line with the `#[name = "..."]` attribute on methods.


kngwyu (approved):
Thanks, maybe worth adding a CHANGELOG entry?

alex:
merge conflict :-(

kngwyu:
Thanks!

PR #1301, opened 2020-11-30
Add section about Python::check_signals to the FAQ

Palladinium:
I think `Python::check_signals` should be mentioned somewhere in the user guide, as it can be difficult to figure out why Ctrl-C doesn't work while running Rust code and the function is somewhat buried in the documentation.


davidhewitt:
👍 thank you!

PR #1303, opened 2020-11-30
Do not include the double quotes in `#[pyclass(name = "literal")]`

scalexm:
Embarrassing... My previous PR was quite wrong actually, as it would set the class name to `'"literal"'`, double quotes included :)

### DIFF
--- a/pyo3-derive-backend/src/pyclass.rs
+++ b/pyo3-derive-backend/src/pyclass.rs
@@ -15,7 +15,7 @@ use syn::{parse_quote, Expr, Token};
 /// The parsed arguments of the pyclass macro
 pub struct PyClassArgs {
     pub freelist: Option<syn::Expr>,
-    pub name: Option<syn::Expr>,
+    pub name: Option<syn::ExprPath>,
     pub flags: Vec<syn::Expr>,
     pub base: syn::TypePath,
     pub has_extends: bool,
@@ -92,18 +92,17 @@ impl PyClassArgs {
                 self.freelist = Some(syn::Expr::clone(right));
             }
             "name" => match &**right {
-                syn::Expr::Lit(
-                    lit
-                    @
-                    syn::ExprLit {
-                        lit: syn::Lit::Str(..),
-                        ..
-                    },
-                ) => {
-                    self.name = Some(lit.clone().into());
+                syn::Expr::Lit(syn::ExprLit {
+                    lit: syn::Lit::Str(lit),
+                    ..
+                }) => {
+                    self.name = match lit.parse() {
+                        Ok(name) => Some(name),
+                        Err(..) => expected!("type name (e.g., Name or \"Name\")"),
+                    }
                 }
                 syn::Expr::Path(exp) if exp.path.segments.len() == 1 => {
-                    self.name = Some(exp.clone().into());
+                    self.name = Some(exp.clone());
                 }
                 _ => expected!("type name (e.g., Name or \"Name\")"),
             },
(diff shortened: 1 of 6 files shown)


pyo3-derive-backend/src/pyclass.rs
```diff
@@ -15,7 +15,7 @@ use syn::{parse_quote, Expr, Token};
 /// The parsed arguments of the pyclass macro
 pub struct PyClassArgs {
     pub freelist: Option<syn::Expr>,
-    pub name: Option<syn::Expr>,
+    pub name: Option<syn::ExprPath>,
```
davidhewitt, line 18: I actually wonder if this should be `syn::LitStr`?

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -92,18 +92,17 @@ impl PyClassArgs {
...
+                }) => {
+                    self.name = match lit.parse() {
+                        Ok(name) => Some(name),
+                        Err(..) => expected!("type name (e.g., Name or \"Name\")"),
+                    }
                 }
                 syn::Expr::Path(exp) if exp.path.segments.len() == 1 => {
-                    self.name = Some(exp.clone().into());
+                    self.name = Some(exp.clone());
                 }
```
davidhewitt, line 106: I think this is the branch for `name` without quotes? I would actually be quite happy to deprecate this as I think `name="CustomName"` is clearer than `name=CustomName`.

If you agree, we could make this return an error message suggesting the user wrap the name in quotes. 

cc @kngwyu - do you agree with a breaking change here to tidy this up?

davidhewitt (commented):
Thanks, sorry for the slow review by us! I just have a couple of small final thoughts on this one...

pyo3-derive-backend/src/pyclass.rs
scalexm: Could easily be `LitStr` if we deprecate the other branch as you said, although it may be better not to allow arbitrary strings, so maybe `Ident` is even better.

pyo3-derive-backend/src/pyclass.rs
scalexm: I’d support that as well, also `ExprPath` is actually too restrictive as it doesn’t play well with substituted macro variables which are basically `None` delimited `ExprGroup`. That was the very reason for my initial PR.

kngwyu: Yeah, I agree to remove the branch.

kngwyu (approved):
Thank you and sorry for the lazy review, too. I just overlooked this PR...

davidhewitt:
@scalexm I'm going to finish and force push this branch this morning, so we can include it in the release. Hope that's okay!

davidhewitt:
@kngwyu this branch should now be ready to merge for the release!

pyo3-derive-backend/src/pyclass.rs
```diff
@@ -92,20 +92,25 @@ impl PyClassArgs {
...
-                        lit: syn::Lit::Str(..),
-                        ..
-                    },
-                ) => {
-                    self.name = Some(lit.clone().into());
+                syn::Expr::Lit(syn::ExprLit {
+                    lit: syn::Lit::Str(lit),
+                    ..
+                }) => {
+                    self.name = Some(lit.parse()?);
```
scalexm, line 99: Maybe we should catch the error from `parse` here, in order to display a friendlier error.

davidhewitt: Agreed, pushed a fix.

kngwyu:
Thanks!

PR #1304, opened 2020-12-01
experimental: owned objects API (alternative)

davidhewitt:
This is a fork of #1300 to use `PyAny<'py>` as an *owned* pointer, as per discussion in that thread. The `PyOwned` type is removed.

I haven't implemented any design around `PyAnyRaw` / `PySend` / `Py` yet.

Overall this design feels simpler to work with, so I plan to continue refining this branch over the original one.

The tests aren't all passing yet, but I did run a benchmark:

| benchmark    | current api                           | experimental api (PyOwned)        | experimental api (PyAny)          | 
| -------------- | ----------------------------        | ------------------------------    | -------------------------------   |
| bench_call_0 | `52,895 ns/iter (+/- 6,956)`          | `54,954 ns/iter (+/- 2,175)`      | `51,247 ns/iter (+/- 1,852)`      |
| bench_call_method_0 | `153,977 ns/iter (+/- 9,671)`  | `153,001 ns/iter (+/- 6,288)`     | `141,100 ns/iter (+/- 10,980)`    |
| dict_get_item | `1,933,300 ns/iter (+/- 328,382)`    | `1,780,027 ns/iter (+/- 79,625)`  | `1,700,250 ns/iter (+/- 52,607)`  |
| extract_btreemap | `9,538,110 ns/iter (+/- 634,895)` | `8,688,905 ns/iter (+/- 590,359)` | `8,201,100 ns/iter (+/- 153,532)` |
| extract_btreeset | `8,099,900 ns/iter (+/- 570,799)` | `7,842,395 ns/iter (+/- 535,694)` | `7,569,645 ns/iter (+/- 144,288)` |
| extract_hashmap | `5,013,300 ns/iter (+/- 483,341)`  | `4,292,595 ns/iter (+/- 358,152)` | `4,133,425 ns/iter (+/- 231,738)` |
| extract_hashset | `5,418,065 ns/iter (+/- 440,259)`  | `4,900,090 ns/iter (+/- 355,389)` | `4,730,320 ns/iter (+/- 86,596)`  |
| iter_dict | `1,759,862 ns/iter (+/- 209,509)`        | `1,304,683 ns/iter (+/- 133,726)` | `1,328,632 ns/iter (+/- 203,382)` |
| iter_list | `1,262,143 ns/iter (+/- 76,108)`         | `1,109,227 ns/iter (+/- 52,926)`  | `1,163,203 ns/iter (+/- 86,611)`  |
[...]

PR #1308, opened 2020-12-08
experimental: owned objects API

davidhewitt:
This is active branch to replace #1300 and #1304 

See the discussions in those PRs to the motivation for the design (and some preliminary benchmarks).

The idea is that we can use this new `experimental` API to iterate on types and traits. We should be able to make the `macros` and `prelude` be compatible with both APIs until we are ready to call the new API stable and then deprecate the original one.

TODO (list probably not complete):
 - [ ] Make `experimental` forms of macros
 - [ ] Lots of documentation for all new functions
 - [ ] Implement `type PyAnyRaw = Py<Any>` ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ hashbrown = { version = "0.9", optional = true }
 assert_approx_eq = "1.1.0"
 trybuild = "1.0.23"
 rustversion = "1.0"
+maplit = "1.0"
 
 [features]
 default = ["macros"]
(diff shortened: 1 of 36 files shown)


davidhewitt:
(Similar to the `abi3` branch, I expect it might take a while for this one to be ready to merge. I'll try to work on it when I can; I'm quite busy at the moment too!)

kngwyu:
Thanks!

aviramha:
How can we make progress with this? Anything that can be split into smaller changes and be contributed by less experienced coders? ;)

davidhewitt:
I'd like to start looking at this as part of the 0.17 release work, in perhaps a month or so. I'll try to find a good way to split this at that point.

davidhewitt:
Closing in favour of #3361

PR #1310, opened 2020-12-12
Implement 128bit integer conversion for limited API

### LINKED ISSUE #1309: abi3 causes compile failures where u128s are involved
A `#[staticmethod]` that takes a `u128` as an argument will fail to compile with abi3, but compiles fine without abi3

### 🌍 Environment

 - Your operating system and version: Arch Linux running in WSL2
 - Your python version: Python 3.9.0
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: `pacman`
 - Your Rust version (`rustc --version`): `rustc 1.48.0 (7eac88abb 2020-11-16)`
 - Your PyO3 version: Latest master (commits `92b7a9736c9585883df5ec5bc01004caa7f65106` and `9aa70f7c89afdbafad47bce02e1cdbdb893406bb`)
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: I already was.

### 💥 Reproducing

Please provide a [minimal working example](https://stackoverflow.com/help/mcve). This means both the Rust code and the Python.

```rust
use pyo3::prelude::*;

#[pyclass]
struct Example {
    a: u128,
}

#[pymethods]
impl Example {
    #[staticmethod]
    pub fn new_seed(seed: u128) -> Self {
        Example { a: seed }
    }
}
```

```toml
[dependencies.pyo3]
git = "https://github.com/PyO3/pyo3"
rev = "9aa70f7c89afdbafad47bce02e1cdbdb893406bb"
features = ["extension-module", "abi3"]
```

Result:
```
error[E0277]: the trait bound `u128: PyClass` is not satisfied
 --> src/lib.rs:8:1
  |
8 | #[pymethods]
  | ^^^^^^^^^^^^ the trait `PyClass` is not implemented for `u128`
  |
  = note: required because of the requirements on the impl of `pyo3::FromPyObject<'_>` for `u128`
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info
```

kngwyu:
Fix #1309 
They would be slow since they use `eval`... but could there be any alternative?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add FFI definitions for `PyBuffer_SizeFromFormat`, `PyObject_LengthHint`, `PyObject_CallNoArgs`, `PyObject_CallOneArg`, `PyObject_CallMethodNoArgs`, `PyObject_CallMethodOneArg`, `PyObject_VectorcallDict`, and `PyObject_VectorcallMethod`. [#1287](https://github.com/PyO3/pyo3/pull/1287)
 - Allow the use of a string literal in `#[pyclass(name = "string literal")]`. [#1295](https://github.com/PyO3/pyo3/pull/1295)
 - Add section about Python::check_signals to the FAQ page of the user guide. [#1301](https://github.com/PyO3/pyo3/pull/1301)
-- Add conversions between u128/i128 and PyLong for abi3 and PyPy. [#1310](https://github.com/PyO3/pyo3/pull/1310)
+- Add conversions between u128/i128 and PyLong for PyPy. [#1310](https://github.com/PyO3/pyo3/pull/1310)
 
 ### Changed
 - Change return type `PyType::name()` from `Cow<str>` to `PyResult<&str>`. [#1152](https://github.com/PyO3/pyo3/pull/1152)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add FFI definitions for `PyBuffer_SizeFromFormat`, `PyObject_LengthHint`, `PyObject_CallNoArgs`, `PyObject_CallOneArg`, `PyObject_CallMethodNoArgs`, `PyObject_CallMethodOneArg`, `PyObject_VectorcallDict`, and `PyObject_VectorcallMethod`. [#1287](https://github.com/PyO3/pyo3/pull/1287)
 - Allow the use of a string literal in `#[pyclass(name = "string literal")]`. [#1295](https://github.com/PyO3/pyo3/pull/1295)
 - Add section about Python::check_signals to the FAQ page of the user guide. [#1301](https://github.com/PyO3/pyo3/pull/1301)
+- Add conversions between u128/i128 and PyLong for abi3 and PyPy. [#1310](https://github.com/PyO3/pyo3/pull/1310)
 
 ### Changed
 - Change return type `PyType::name()` from `Cow<str>` to `PyResult<&str>`. [#1152](https://github.com/PyO3/pyo3/pull/1152)
--- a/src/types/num.rs
+++ b/src/types/num.rs
@@ -157,23 +157,13 @@ int_convert_u64_or_i64!(
     ffi::PyLong_AsUnsignedLongLong
 );
 
-// manual implementation for 128bit integers
 #[cfg(not(any(Py_LIMITED_API, PyPy)))]
-mod int128_conversion {
-    use crate::{
-        ffi, AsPyPointer, FromPyObject, IntoPy, PyAny, PyErr, PyNativeType, PyObject, PyResult,
-        Python, ToPyObject,
-    };
-    use std::os::raw::{c_int, c_uchar};
-
-    #[cfg(target_endian = "little")]
-    const IS_LITTLE_ENDIAN: c_int = 1;
-    #[cfg(not(target_endian = "little"))]
-    const IS_LITTLE_ENDIAN: c_int = 0;
+mod fast_128bit_int_conversion {
+    use super::*;
 
     // for 128bit Integers
     macro_rules! int_convert_128 {
-        ($rust_type: ty, $byte_size: expr, $is_signed: expr) => {
+        ($rust_type: ty, $is_signed: expr) => {
             impl ToPyObject for $rust_type {
                 #[inline]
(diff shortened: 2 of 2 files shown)


birkenfeld:
Why use `eval` when you can call the method directly with `call_method`?

birkenfeld:
This is the snippet I tested:
```
        let bytes = obj.call_method("to_bytes", (16, "little"), None)?;
        let res = u128::from_le_bytes(bytes.extract()?);
```

Of course, this is missing the `PyNumber_Index` call that the other conversion is doing.

birkenfeld:
Alternative, as I said in #1309, is to use `PyNumber_` APIs to shift and mask the number to get two values that fit into u64's.

kngwyu:
> Why use eval when you can call the method directly with call_method?

Thanks, I was just bad not to come up with it. We can call `PyNumber_Index` manually.

>  to use PyNumber_ APIs to shift and mask the number to get two values that fit into u64's.

I tried that first but I found it too complex, especially for `i128`, though it is faster...

birkenfeld:
So in my tests (FromPyObject only) it was slightly faster to call with `call_method`, but then use your method of `downcast` instead of `extract`.

src/types/num.rs
```diff
@@ -183,11 +173,12 @@ mod int128_conversion {
             impl IntoPy<PyObject> for $rust_type {
                 fn into_py(self, py: Python) -> PyObject {
                     unsafe {
-                        let bytes = self.to_ne_bytes();
+                        // Always use little endian
+                        let bytes = self.to_le_bytes();
```
kngwyu, line 177: Though it's not related to this PR, I noticed that `to_le_bytes` is added in Rust 1.32. 
Using always little-endian simplified the code a lot.

src/types/num.rs
```diff
@@ -183,11 +173,12 @@ mod int128_conversion {
...
                         let obj = ffi::_PyLong_FromByteArray(
-                            bytes.as_ptr() as *const c_uchar,
-                            $byte_size,
```
kngwyu, line 189: I don't know why `16` is not hard-coded...

davidhewitt: maybe `bytes.len()` is most correct? I would think that the compiler will still optimize this at compile time.

src/types/num.rs
```diff
@@ -202,83 +193,131 @@ mod int128_conversion {
                         if num.is_null() {
                             return Err(PyErr::fetch(ob.py()));
                         }
-                        let mut buffer = [0; $byte_size];
+                        let mut buffer = [0; 16];
```
davidhewitt, line 196:
```suggestion
                        let mut buffer = [0; std::mem::size_of::<$rust_type>()];
```

src/types/num.rs
```diff
@@ -202,83 +193,131 @@ mod int128_conversion {
...
                         let ok = ffi::_PyLong_AsByteArray(
                             num as *mut ffi::PyLongObject,
                             buffer.as_mut_ptr(),
-                            $byte_size,
-                            IS_LITTLE_ENDIAN,
+                            16,
```
davidhewitt, line 200:
```suggestion
                            buffer.len(),
```

src/types/num.rs
```diff
@@ -202,83 +193,131 @@ mod int128_conversion {
...
+                }
+            }
+
+            impl<'source> FromPyObject<'source> for $rust_type {
+                fn extract(ob: &'source PyAny) -> PyResult<$rust_type> {
+                    let py = ob.py();
+                    let num = unsafe { ffi::PyNumber_Index(ob.as_ptr()) };
+                    let num: &PyLong =
+                        unsafe { crate::FromPyPointer::from_owned_ptr_or_err(py, num)? };
+                    let args = (16, "little".to_object(py));
```
davidhewitt, line 254:
```suggestion
                    let args = (std::mem::size_of::<$rust_type>(), "little".to_object(py));
```

or could define a `const SIZE: usize = std::mem::size_of::<$rust_type>()` and use here and on line 258.

davidhewitt (commented):
Nice!

I think it's better to use `std::mem::size_of{_val}` instead of hard-coding 16, as it's easier to understand what that means for a future reader rather than just seeing a magic number 16.

src/types/num.rs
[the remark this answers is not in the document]
kngwyu: This new function is added in Rust 1.32, which always returns little-endian bytes. To simplify the code, I switched to use this.

src/types/num.rs
```diff
@@ -202,83 +193,134 @@ mod int128_conversion {
...
+            impl<'source> FromPyObject<'source> for $rust_type {
+                fn extract(ob: &'source PyAny) -> PyResult<$rust_type> {
+                    let py = ob.py();
+                    let num: &PyLong = unsafe {
+                        crate::FromPyPointer::from_owned_ptr_or_err(
+                            py,
+                            ffi::PyNumber_Index(ob.as_ptr()),
+                        )
+                    }?;
+                    let kwargs = [("signed", $is_signed)].into_py_dict(py);
```
birkenfeld, line 257: ah damn, it requires a dict here too then. Can you at least make this
```
let kwargs = if $is_signed { Some(...) } else { None };
```
?

kngwyu: Sure,

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add FFI definitions for `PyBuffer_SizeFromFormat`, `PyObject_LengthHint`, `PyObject_CallNoArgs`, `PyObject_CallOneArg`, `PyObject_CallMethodNoArgs`, `PyObject_CallMethodOneArg`, `PyObject_VectorcallDict`, and `PyObject_VectorcallMethod`. [#1287](https://github.com/PyO3/pyo3/pull/1287)
 - Allow the use of a string literal in `#[pyclass(name = "string literal")]`. [#1295](https://github.com/PyO3/pyo3/pull/1295)
 - Add section about Python::check_signals to the FAQ page of the user guide. [#1301](https://github.com/PyO3/pyo3/pull/1301)
+- Add conversions between u128/i128 and PyLong for abi3 and PyPy. [#1310](https://github.com/PyO3/pyo3/pull/1310)
```
davidhewitt, line 27: I think because `abi3` is new this release, we don't need it mentioned specifically in this line:

```suggestion
- Add conversions between u128/i128 and PyLong for PyPy. [#1310](https://github.com/PyO3/pyo3/pull/1310)
```

kngwyu:
Thank you for your reviews!

davidhewitt (approved):
💯

PR #1315, opened 2020-12-13
Faster conversion from u128/i128 to PyLong with abi3

### LINKED ISSUE #1314: Use shifting and masking to convert between 128-bit Rust integers and Python integers
@kngwyu in https://github.com/PyO3/pyo3/issues/1310#issuecomment-743722886_
> >  to use PyNumber_ APIs to shift and mask the number to get two values that fit into u64's.
> 
> I tried that first but I found it too complex, especially for `i128`, though it is faster...

We should use the faster method anyway, I don't think it'll be that much more complex.

kngwyu:
Fixes #1314 
I hope it's correct...

Also, adding 30 lines is acceptable for such a minor feature?

### DIFF
--- a/src/types/num.rs
+++ b/src/types/num.rs
@@ -281,6 +281,32 @@ mod slow_128bit_int_conversion {
 mod test_128bit_intergers {
     use super::*;
 
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn test_i128_roundtrip(x: i128) {
+            Python::with_gil(|py| {
+                let x_py = x.into_py(py);
+                crate::py_run!(py, x_py, &format!("assert x_py == {}", x));
+                let roundtripped: i128 = x_py.extract(py).unwrap();
+                assert_eq!(x, roundtripped);
+            })
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn test_u128_roundtrip(x: u128) {
+            Python::with_gil(|py| {
+                let x_py = x.into_py(py);
+                crate::py_run!(py, x_py, &format!("assert x_py == {}", x));
+                let roundtripped: u128 = x_py.extract(py).unwrap();
+                assert_eq!(x, roundtripped);
+            })
+        }
+    }
+
     #[test]
     fn test_i128_max() {
         Python::with_gil(|py| {
--- a/src/types/num.rs
+++ b/src/types/num.rs
@@ -216,24 +216,24 @@ mod fast_128bit_int_conversion {
     int_convert_128!(u128, 0);
 }
 
-// For ABI3 and PyPy, we implement the conversion using `call_method`.
+// For ABI3 and PyPy, we implement the conversion (almost) manually.
 #[cfg(any(Py_LIMITED_API, PyPy))]
 mod slow_128bit_int_conversion {
     use super::*;
-    use crate::types::{IntoPyDict, PyBytes, PyDict, PyType};
+    use crate::types::{IntoPyDict, PyBytes};
+    const BYTE_SIZE: usize = std::mem::size_of::<i128>();
 
-    // Returns `{'signed': True}` if is_signed else None
-    fn kwargs(py: Python, is_signed: bool) -> Option<&PyDict> {
-        if is_signed {
-            Some([("signed", is_signed)].into_py_dict(py))
-        } else {
-            None
-        }
+    fn split_128int(bytes: [u8; BYTE_SIZE]) -> ([u8; BYTE_SIZE / 2], [u8; BYTE_SIZE / 2]) {
+        let mut first = [0u8; BYTE_SIZE / 2];
+        let mut last = [0u8; BYTE_SIZE / 2];
+        first.copy_from_slice(&bytes[..BYTE_SIZE / 2]);
+        last.copy_from_slice(&bytes[BYTE_SIZE / 2..]);
+        (first, last)
     }
 
     // for 128bit Integers
     macro_rules! int_convert_128 {
-        ($rust_type: ty, $is_signed: literal) => {
+        ($rust_type: ty, $half_type: ty, $is_signed: literal) => {
             impl ToPyObject for $rust_type {
                 #[inline]
                 fn to_object(&self, py: Python) -> PyObject {
@@ -243,12 +243,21 @@ mod slow_128bit_int_conversion {
 
             impl IntoPy<PyObject> for $rust_type {
                 fn into_py(self, py: Python) -> PyObject {
-                    let bytes = PyBytes::new(py, &self.to_le_bytes());
-                    let longtype = unsafe { PyType::from_type_ptr(py, &mut ffi::PyLong_Type) };
-                    longtype
-                        .call_method("from_bytes", (bytes, "little"), kwargs(py, $is_signed))
-                        .expect("Integer conversion (u128/i128 to PyLong) failed")
-                        .into_py(py)
+                    let (first, last) = split_128int(self.to_le_bytes());
+                    let (first, last) =
+                        (u64::from_le_bytes(first), <$half_type>::from_le_bytes(last));
+                    let (first, last) = (first.into_py(py), last.into_py(py));
+                    let shift = 64u64.into_py(py);
+                    unsafe {
+                        let shifted = PyObject::from_owned_ptr(
+                            py,
+                            ffi::PyNumber_Lshift(last.as_ptr(), shift.as_ptr()),
+                        );
+                        PyObject::from_owned_ptr(
+                            py,
+                            ffi::PyNumber_Or(shifted.as_ptr(), first.as_ptr()),
+                        )
+                    }
                 }
             }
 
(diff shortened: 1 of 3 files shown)


kngwyu:
Now I understand why I though this approach too complex: I was thinking of using `lshift` and `add`, but `lshift` and `or` is easier.

src/types/num.rs
```diff
@@ -261,9 +270,14 @@ mod slow_128bit_int_conversion {
...
-                    let mut bytes = [0u8; std::mem::size_of::<$rust_type>()];
+                    let mut bytes = [0u8; BYTE_SIZE];
+                    let kwargs = if $is_signed {
+                        Some([("signed", true)].into_py_dict(py))
+                    } else {
+                        None
+                    };
                     let pybytes: &PyBytes = num
-                        .call_method("to_bytes", (bytes.len(), "little"), kwargs(py, $is_signed))?
+                        .call_method("to_bytes", (bytes.len(), "little"), kwargs)?
```
birkenfeld, line 280: this seems to revert a refactoring, or did you want to change the FromPy part as well?

kngwyu: Since now the `kwargs` function is called once, I moved it in the function.

src/types/num.rs
```diff
@@ -243,12 +243,21 @@ mod slow_128bit_int_conversion {
 
             impl IntoPy<PyObject> for $rust_type {
                 fn into_py(self, py: Python) -> PyObject {
-                    let bytes = PyBytes::new(py, &self.to_le_bytes());
-                    let longtype = unsafe { PyType::from_type_ptr(py, &mut ffi::PyLong_Type) };
-                    longtype
-                        .call_method("from_bytes", (bytes, "little"), kwargs(py, $is_signed))
-                        .expect("Integer conversion (u128/i128 to PyLong) failed")
-                        .into_py(py)
+                    let (first, last) = split_128int(self.to_le_bytes());
```
davidhewitt, line 246: Can I suggest we use the terminology `lower` / `upper` instead of `first` / `last`? For me it's easier to understand which bits are most significant / least significant with that terminology.

```suggestion
                    let (lower, upper) = split_128int(self.to_le_bytes());
```

davidhewitt (commented):
Thanks. It's 14 more lines of code total for a more efficient solution so I think this is a desirable change. I expect abi3 builds to become quite common once we have them working.

I think it'd be quite good to have a test here to be sure that we've got the right solution without any edge cases.

`proptest` can be pretty nice for tests like this where data is passed in a roundtrip. I think we can have a test like this, for example:

```rust
use proptest::prelude::*;

proptest! {
    #[test]
    fn test_i128_roundtrip(x: i128) {
        Python::with_gil(|py| {
            let x_py = x.into_py(py);
            py_run!(py, x_py, format!("assert x_py == {}", x));
            let roundtripped: i128 = x_py.extract().unwrap();
            assert_eq!(x, roundtripped);
        })
    }
}
```

and similar for u128.

src/types/num.rs
kngwyu: Thanks, I wanted better names too 😉

kngwyu:
Just a status update: renamed some variables and re-implement PyLong to i128/u128 conversion using `PyLong_AsUnsignedLongLongMask` and shift.
Proptest looks good but I'm still reading the document.

src/types/num.rs
```diff
@@ -243,37 +241,49 @@ mod slow_128bit_int_conversion {
...
 
             impl<'source> FromPyObject<'source> for $rust_type {
                 fn extract(ob: &'source PyAny) -> PyResult<$rust_type> {
                     let py = ob.py();
-                    let num: &PyLong = unsafe {
-                        crate::FromPyPointer::from_owned_ptr_or_err(
+                    unsafe {
+                        let lower = err_if_invalid_value(
                             py,
-                            ffi::PyNumber_Index(ob.as_ptr()),
```
davidhewitt, line 261: I think the call to `PyNumber_Index` is still needed before calling `PyLong_AsUnsignedLongLongMask`, so that types which implement `__index__` are converted to `PyLong`?

davidhewitt (commented):
Nice! Couple more suggestions...

src/types/num.rs
kngwyu: No, `PyLong_AsUnsigned...` calls it internally.

src/types/num.rs
```diff
@@ -243,37 +241,49 @@ mod slow_128bit_int_conversion {
...
-                    let bytes = PyBytes::new(py, &self.to_le_bytes());
-                    let longtype = unsafe { PyType::from_type_ptr(py, &mut ffi::PyLong_Type) };
-                    longtype
-                        .call_method("from_bytes", (bytes, "little"), kwargs(py, $is_signed))
-                        .expect("Integer conversion (u128/i128 to PyLong) failed")
-                        .into_py(py)
+                    let le_bytes = self.to_le_bytes();
+                    let lower = u64::from_le_bytes(slice_to_bytearr(&le_bytes[..BYTE_SIZE / 2]));
+                    let upper =
+                        <$half_type>::from_le_bytes(slice_to_bytearr(&le_bytes[BYTE_SIZE / 2..]));
```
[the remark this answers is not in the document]
kngwyu: Looks nice 👍🏽

kngwyu:
Added an overflow test for i128.
And for `proptest`, I think in this case `max/min` is sufficiently strong and we don't *strongly* need randomized tests.

birkenfeld:
There should also be a test with values in the `i/u64` range, to ensure that the two halves are not mixed up.

src/types/num.rs
davidhewitt: Ah perfect, thanks for checking!

davidhewitt:
Agreed; if we're not proptesting we should make sure the tests we do have cover for mistakes like halves being mixed up. We know it's correct now but who knows what will happen next time I try to refactor this code 👀

programmerjake:
Why are you using all the le bytes methods when you can just use:
```rust
let v: u128 = ...;
let low_half = v as u64;
let high_half = (v >> 64) as u64;
```
and
```rust
let v: i128 = ...;
let low_half = v as u64;
let high_half = (v >> 64) as i64;
```
and then on the python side:
```Python
(high_half << 64) | low_half
```

programmerjake:
Good test values are:
```rust
let u128_test_value = 0xFF0102030405060788090A0B0C0D0E0Fu128;
let i128_test_value = u128_test_value as i128;
```
since it tests the halves are in the right order with the right signedness.

davidhewitt:
> Why are you using all the le bytes methods when you can just use:

😅 great suggestion, thanks! I think when this implementation started with `to_bytes` we got stuck on one idea too much!

kngwyu:
> Why are you using all the le bytes methods when you can just use:

Wow, thanks. It seems like I wrote too many TeX documents to forget some bit hacks 🙄

kngwyu:
Added a basic proptest but is this sufficient?

src/types/num.rs
```diff
@@ -243,43 +234,79 @@ mod slow_128bit_int_conversion {
...
-                        )
-                    }?;
-                    let mut bytes = [0u8; std::mem::size_of::<$rust_type>()];
-                    let pybytes: &PyBytes = num
-                        .call_method("to_bytes", (bytes.len(), "little"), kwargs(py, $is_signed))?
-                        .extract()?;
-                    bytes.copy_from_slice(pybytes.as_bytes());
-                    Ok(<$rust_type>::from_le_bytes(bytes))
+                            -1 as _,
+                            ffi::PyLong_AsUnsignedLongLongMask(ob.as_ptr()),
```
programmerjake, line 262: I think you'd probably want an assertion of some sort that `PyLong_UnsignedLongLongMask` returns the expected type, since C/C++ don't guarantee that it's `u64` -- it could be `u128` or some other type on an unusual OS.

programmerjake:
You accidentally changed `src/lib.rs` to an executable file.

programmerjake:
Btw, sorry to sound all negative, thanks for all your work!

programmerjake:
other than all that, looks good to me!

src/types/num.rs
kngwyu: Hmm 🤔 , but even the Rust stdlib does not support it https://doc.rust-lang.org/nightly/std/os/raw/type.c_ulonglong.html.

kngwyu:
> Btw, sorry to sound all negative, thanks for all your work!

No worry, thank you for your comments 👍🏼

src/types/num.rs
programmerjake: The docs are just showing the type used on the system they used to build the docs (`x86_64-unknown-linux-gnu` iirc), it doesn't mean it will always be `u64`.

programmerjake: For an example, see https://doc.rust-lang.org/nightly/std/os/raw/type.c_long.html

kngwyu: I meant the description: 
> The C standard technically only requires that this type be an unsigned integer with the size of a long long, although in practice, no system would have a long long that is not a u64, as most systems do not have a standardised u128 type.

Also, it has no `#[cfg` block:
image
So I think we don't need to consider the system where ull is u128 *for now*

programmerjake: ok, yeah. I did check both the AS/400 and RISC-V 128-bit compilers (or at least their ABI proposals/docs) and they both have a 64-bit ull -- AS/400 just doesn't have a 128-bit int type (no intptr_t at all), and RV128 proposes `long long long` for 128-bit types.

davidhewitt (approved):
Looks good! Just need to `chmod -x src/lib.rs` before merging. Not sure how that happened 🤔

kngwyu:
Thanks!
> Not sure how that happened

Maybe my emacs setting does something wrong, but I'm also not sure 😓

PR #1316, opened 2020-12-13
Add symbolic links to LICENSE in subcrates

### LINKED ISSUE #1313: please include LICENSE files in all published crates
The `pyo3-derive-backend` and `py3cls` crates published to crates.io do not contain the LICENSE text file of the project. For Apache-2.0, it's necessary to distribute the license text alongside the source code it covers. This crates issues for (linux) distributions which redistribute the crate source code.

Adding a symlink from the `pyo3-derive-backend` and `pyo3cls` directories to the LICENSE file in the root of the repository should be enough to fix this, assuming you're running `cargo publish` on linux or mac OS (AFAIK, full copies of the files are necessary on windows, since symlinks do not work there).

kngwyu:
Fixes #1313 
But copies can be better if we sometimes publish the crates on Windows?
cc: @davidhewitt


davidhewitt (approved):
I also publish from linux (WSL2) so I think that it's fine to have symlinks.

Perhaps we could move the `cargo publish` commands to Github actions?

birkenfeld:
What's the problem with making copies? It's not like the license changes.

davidhewitt:
I just took a look at `serde` and `wasm-bindgen` - both of those projects use symlinks so I think it's fine for us to use also.

PR #1317, opened 2020-12-13
macros: rename crates for consistency

davidhewitt:
This PR is a simple rename of `pyo3cls` to `pyo3-macros` and `pyo3-derive-backend` to `pyo3-macros-backend`.

I did this for a couple reasons:
 - I think the original names are not very clear.
 - The name `pyo3-macros` matches nicely with the `macros` feature.


birkenfeld:
Oh yes please, this is much better.

konstin:
For historical context, I initially named it pyo3-derive-backend because I wanted to rename pyo3cls to pyo3-derive, but then didn't follow through because I realized you can't rename crates on crates.io

davidhewitt:
Mmm I think I can manually release `0.12.5` versions of the old names, with READMEs saying to use the new ones.

I chose `macros` / `macros-backend` over `derive` / `derive-backend` because it's not just `#[derive()]` macros we provide.

kngwyu:
I agree that the names are confusing (especially `pyo3cls`) and renaming is acceptable when we note that in README or so.

davidhewitt:
👍 I propose I merge this now and immediately release these crates with versions `0.13.0-alpha.1`, so that they are available. I will also push `0.12.5` versions of `pyo3cls` and `pyo3-derive-backend` which note the change of name.

Then when `0.13` release comes the main pyo3 crate will be published to use these new crates.

davidhewitt:
@kngwyu unless you say otherwise I plan to merge this tomorrow.

kngwyu (approved):
Sure, please.

davidhewitt:
That's done, and all crates updated on `crates.io`:

https://crates.io/crates/pyo3cls
https://crates.io/crates/pyo3-derive-backend
https://crates.io/crates/pyo3-macros
https://crates.io/crates/pyo3-macros-backend

PR #1318, opened 2020-12-13
Clearly document the limitations of abi3

### LINKED ISSUE #1288: #[text_signature] doesn't work with abi3
The solution in #1201 for `text_signature` with abi3 got disabled until Python 3.10, so at the moment I think `__text_signature__` will just silently be `None` if abi3 is enabled. 

We might want to consider a more user-friendly solution like failing at compile-time. 

Not sure if this blocks 0.13 release - I think it's potentially quite annoying but also I expect no users are using abi3 yet, so it'll only break for them when they switch to abi3.

At the very least we should document the problem alongside abi3 docs before releasing.

alex:
closes #1288


davidhewitt:
👍 that's great, thanks so much!

PR #1319, opened 2020-12-14
pyclass #[new]: allow using custom error type

davidhewitt:
Similar to #1118 - in that PR we missed out allowing a different result type for `#[new]`. This PR fixes that.

(Came up in a real-world use case for someone I was speaking with earlier today.)


kngwyu (approved):
LGTM, thank you!

PR #1320, opened 2020-12-15
abi3-py36 now passes tests

kngwyu:
Now `cargo test --features=abi3-py36` fails since `ffi::PyEval_InitThreads()` is called after `Py_InitializeEx`.
So now `py36` build is not actually forward-compatible.
We have a small fix for this problem (this PR), but this also suggests that `py3*` build can be broken in the future version.
Though I'm not sure the breakage is likely to happen in the user code, maybe we want to add a note about it.

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -65,12 +65,8 @@ pub fn prepare_freethreaded_python() {
             // If Python is already initialized, we expect Python threading to also be initialized,
             // as we can't make the existing Python main thread acquire the GIL.
             #[cfg(not(Py_3_7))]
-            assert_ne!(ffi::PyEval_ThreadsInitialized(), 0);
+            debug_assert_ne!(ffi::PyEval_ThreadsInitialized(), 0);
         } else {
-            // If Python isn't initialized yet, we expect that Python threading
-            // isn't initialized either.
-            #[cfg(not(Py_3_7))]
-            assert_eq!(ffi::PyEval_ThreadsInitialized(), 0);
             // Initialize Python.
             // We use Py_InitializeEx() with initsigs=0 to disable Python signal handling.
             // Signal handling depends on the notion of a 'main thread', which doesn't exist in this case.
@@ -97,7 +93,9 @@ pub fn prepare_freethreaded_python() {
             // > Changed in version 3.7: This function is now called by Py_Initialize(), so you don’t have
             // > to call it yourself anymore.
             #[cfg(not(Py_3_7))]
-            ffi::PyEval_InitThreads();
+            if ffi::PyEval_ThreadsInitialized() == 0 {
+                ffi::PyEval_InitThreads();
+            }
             // PyEval_InitThreads() will acquire the GIL,
             // but we don't want to hold it at this point
             // (it's not acquired in the other code paths)
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -64,7 +64,6 @@ pub fn prepare_freethreaded_python() {
         if ffi::Py_IsInitialized() != 0 {
             // If Python is already initialized, we expect Python threading to also be initialized,
             // as we can't make the existing Python main thread acquire the GIL.
-            #[cfg(not(Py_3_7))]
             debug_assert_ne!(ffi::PyEval_ThreadsInitialized(), 0);
         } else {
             // Initialize Python.
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -64,7 +64,7 @@ pub fn prepare_freethreaded_python() {
         if ffi::Py_IsInitialized() != 0 {
             // If Python is already initialized, we expect Python threading to also be initialized,
             // as we can't make the existing Python main thread acquire the GIL.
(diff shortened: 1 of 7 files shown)


kngwyu:
Oops, it looks like that datetime is not declared with `Py_LIMITED_API`....
I'm glad to find this major bug before releasing 0.13!

alex (approved):

src/err/mod.rs
```diff
@@ -602,10 +602,10 @@ mod tests {
...
 
         assert_eq!(fields.next().unwrap(), "type: <class 'Exception'>");
-        #[cfg(not(Py_3_7))] // Python 3.6 and below formats the repr differently
-        assert_eq!(fields.next().unwrap(), ("value: Exception('banana',)"));
-        #[cfg(Py_3_7)]
-        assert_eq!(fields.next().unwrap(), "value: Exception('banana')");
+        assert!(fields
+            .next()
+            .unwrap()
+            .starts_with("value: Exception('banana'"));
```
davidhewitt, line 608: Can we please not make this change?

I left the `#[cfg(...)]` variants in, even though it is more LOC, because it is clearer what the exact behaviour is. `starts_with` is a much weaker check than `assert_eq`.

When we drop Python 3.6 support (presumably next year) it is easy to update this test then to remove the `#[cfg]`s.

src/exceptions.rs
```diff
@@ -598,18 +598,10 @@ mod test {
...
-        #[cfg(not(Py_3_7))]
-        assert_eq!(format!("{:?}", exc), "Exception('banana',)");
+        assert!(format!("{:?}", exc).starts_with("Exception('banana'"));
 
         let source = exc.source().expect("cause should exist");
-
-        #[cfg(Py_3_7)]
-        assert_eq!(format!("{:?}", source), "TypeError('peach')");
-        #[cfg(not(Py_3_7))]
-        assert_eq!(format!("{:?}", source), "TypeError('peach',)");
```
davidhewitt, line 612: Same here.

src/gil.rs
```diff
@@ -64,13 +64,8 @@ pub fn prepare_freethreaded_python() {
         if ffi::Py_IsInitialized() != 0 {
             // If Python is already initialized, we expect Python threading to also be initialized,
             // as we can't make the existing Python main thread acquire the GIL.
-            #[cfg(not(Py_3_7))]
-            assert_ne!(ffi::PyEval_ThreadsInitialized(), 0);
+            debug_assert_ne!(ffi::PyEval_ThreadsInitialized(), 0);
```
davidhewitt, line 67: I think `assert_ne` is better than `debug_assert_ne` here:
- it's not a hot code path (only called once), so the performance overhead doesn't really matter in `--release`
- code will crash in weird ways if threads are not initialised.

davidhewitt (changes requested):
👍 thanks, that's a nice catch!

I think a couple of these changes are done with the goal of minimising LOC & optimizations, but I don't agree with them. I'd prefer we kept what we had before for those cases. See my other comments.

src/err/mod.rs
kngwyu: I did this not for reducing LOC, but for passing tests with `abi3-py36` with Python >= 3.7.
I'm happy to go in another way, e.g., using actual Python version instead of cfg here, though I'm not sure wa can.

davidhewitt: Ah, sorry, I see now. With abi3 flags the `#[cfg()]` applies only `Py_3_6`, even on Python 3.7. 

I guess this is actually a runtime dependency on the Python version - something like `if / else` instead of `#[cfg]`...

davidhewitt: I have an idea for checking this at runtime that I'll push slightly later today.

davidhewitt: See #1322. I fixed these tests to use a new `Python::version()` API.

src/gil.rs
kngwyu: 👍🏼 , I changed to use `assert_ne`

davidhewitt:
I merged #1329 which should fix the pypy issues - if you rebase on master CI should hopefully be green 🤞

davidhewitt (approved):

davidhewitt:
Ok I merged #1331 and now hopefully this will work. I'm going to rebase this on master and then merge if it's ok.

davidhewitt:
Ugh, still failing pypy tests!

I think the issue is that on pypy we can _build_ the tests, but can't ever run them.

I'm going to work on yet another PR to improve the pypy CI...

kngwyu:
Thanks.
Could I help to debug?
What is the precise commit where this starts failing? 
#1325?

davidhewitt:
Ah, yes, that's it - for now I have disabled the pypy abi3-py36 tests being built; that should fix this PR and then I'll open a new PR later which re-enables building the tests.

alex (approved):

davidhewitt:
🎉

PR #1322, opened 2020-12-17
python: add version() to get running version

davidhewitt:
This PR adds a new API `Python::version()` which is a safe convenience wrapper around `Py_GetVersion()`.

Now that we offer abi3 wheels which may need to support multiple different Python versions, such an API will come in helpful. We saw this already in #1320.

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -7,11 +7,95 @@ use crate::gil::{self, GILGuard, GILPool};
 use crate::type_object::{PyTypeInfo, PyTypeObject};
 use crate::types::{PyAny, PyDict, PyModule, PyType};
 use crate::{ffi, AsPyPointer, FromPyPointer, IntoPyPointer, PyNativeType, PyObject, PyTryFrom};
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::marker::PhantomData;
-use std::os::raw::c_int;
+use std::os::raw::{c_char, c_int};
 
 pub use gil::prepare_freethreaded_python;
+
+/// Represents the major, minor, and patch (if any) versions of this interpreter.
+///
+/// See [Python::version].
+#[derive(Debug)]
+pub struct PythonVersionInfo<'p> {
+    pub major: u8,
+    pub minor: u8,
+    pub patch: u8,
+    pub suffix: Option<&'p str>,
+}
+
+impl<'p> PythonVersionInfo<'p> {
+    /// Parses a hard-coded Python interpreter version string (e.g. 3.9.0a4+).
+    ///
+    /// Panics if the string is ill-formatted.
+    fn from_str(version_number_str: &'p str) -> Self {
+        fn split_and_parse_number(version_part: &str) -> (u8, Option<&str>) {
+            match version_part.find(|c: char| !c.is_ascii_digit()) {
+                None => (version_part.parse().unwrap(), None),
+                Some(version_part_suffix_start) => {
+                    let (version_part, version_part_suffix) =
+                        version_part.split_at(version_part_suffix_start);
+                    (version_part.parse().unwrap(), Some(version_part_suffix))
+                }
+            }
+        }
+
+        let mut parts = version_number_str.split('.');
+        let major_str = parts.next().expect("Python major version missing");
+        let minor_str = parts.next().expect("Python minor version missing");
+        let patch_str = parts.next();
+        assert!(
+            parts.next().is_none(),
+            "Python version string has too many parts"
+        );
+
+        let major = major_str
+            .parse()
+            .expect("Python major version not an integer");
+        let (minor, suffix) = split_and_parse_number(minor_str);
+        if suffix.is_some() {
+            assert!(patch_str.is_none());
+            return PythonVersionInfo {
+                major,
+                minor,
+                patch: 0,
+                suffix,
+            };
+        }
+
+        let (patch, suffix) = patch_str.map(split_and_parse_number).unwrap_or_default();
+        PythonVersionInfo {
+            major,
+            minor,
+            patch,
+            suffix,
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


davidhewitt:
PyPy failure is upstream bug in `setup-python`, but we might not want to merge this until this is resolved as I'd like to know that this API works correctly with PyPy.

kngwyu (changes requested):
Thanks!

src/python.rs
```diff
@@ -7,12 +7,129 @@ use crate::gil::{self, GILGuard, GILPool};
...
 pub use gil::prepare_freethreaded_python;
 
+/// Represents the major, minor, and patch (if any) versions of this interpreter.
+///
+/// See [Python::version].
+#[derive(Debug)]
+pub struct PythonVersion {
+    major: u8,
+    minor: (u8, Option<&'static str>),
+    patch: Option<(u8, Option<&'static str>)>,
```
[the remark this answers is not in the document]
kngwyu: Also, I think we can make all members transparent (=`pub`) with this style.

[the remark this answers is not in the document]
davidhewitt: I like this idea, let's try it. And also I agree it's probably correct to assume `patch == 0`.

It's possible all version strings we will see in practice include a patch, though the example for `Py_GetVersion` doesn't!

src/python.rs
```diff
@@ -618,4 +760,39 @@ mod test {
...
+        assert!(version >= (3, 9, 0));
+    }
+
+    #[test]
+    fn test_python_version_parse() {
+        assert!(PythonVersion::from_str("3.5.0a1") >= (3, 5, 0));
+        assert!(PythonVersion::from_str("3.5+") >= (3, 5, 0));
+        assert!(PythonVersion::from_str("3.5+") == (3, 5, 0));
+        assert!(PythonVersion::from_str("3.5+") != (3, 5, 1));
+        assert!(PythonVersion::from_str("3.5+") == (3, 5, 0));
```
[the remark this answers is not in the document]
davidhewitt: Yes, good spot.

src/python.rs
```diff
@@ -7,12 +7,129 @@ use crate::gil::{self, GILGuard, GILPool};
...
+    pub fn minor(&self) -> u8 {
+        self.minor.0
+    }
+
+    /// Gets the patch version, if any.
+    pub fn patch(&self) -> Option<u8> {
+        self.patch.as_ref().map(|patch| patch.0)
+    }
+
+    /// Parses a hard-coded Python interpreter version string.
```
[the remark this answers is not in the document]
davidhewitt: This is a private API so I don't think it needs a full doctest example snippet, I will add a note of the form as you suggest though.

davidhewitt:
Thanks for the review, I pushed some changes to address.

Also I renamed this API to `PythonVersionInfo` and `Python::version_info()`, and also added `Python::version() -> &'static str`. This matches better with `sys.version_info` and `sys.version`.

src/python.rs
```diff
@@ -7,12 +7,112 @@ use crate::gil::{self, GILGuard, GILPool};
...
+
+impl PartialEq<(u8, u8, u8)> for PythonVersionInfo {
+    fn eq(&self, other: &(u8, u8, u8)) -> bool {
+        self.major == other.0 && self.minor == other.1 && self.patch == other.2
+    }
+}
+
+impl PartialOrd<(u8, u8)> for PythonVersionInfo {
+    fn partial_cmp(&self, other: &(u8, u8)) -> Option<std::cmp::Ordering> {
+        match self.major.cmp(&other.0) {
```
kngwyu, line 91: I think you can compare the tuples direcly here, like
```
(self.major, self.minor).partial_cmp(other)
```

davidhewitt:
Thanks, simplified `partial_cmp` implementations and force-pushed.

src/python.rs
```diff
@@ -302,6 +386,49 @@ impl<'p> Python<'p> {
...
+    /// # use pyo3::Python;
+    /// Python::with_gil(|py| {
+    ///     // The full string could be, for example:
+    ///     // "3.0a5+ (py3k:63103M, May 12 2008, 00:53:55) \n[GCC 4.2.3]"
+    ///     assert!(py.version().starts_with("3."));
+    /// });
+    /// ```
+    pub fn version(self) -> &'static str {
+        unsafe {
+            CStr::from_ptr(ffi::Py_GetVersion() as *const c_char)
```
alex, line 404: I'm not sure treating the result of `Py_GetVersion` as ` `&'static str` is safe. Because of the way it's implemented, if you have a thread calling `version()` at the same time as another thread is attempting to read the contents of the str, you have a data race: https://github.com/python/cpython/blob/master/Python/getversion.c

Is there a reason it wouldn't work to give it the `'py` lifetime?

davidhewitt: Ah, yes the claim in the docs it was "static storage" misled me. Nice catch - I'll change the lifetime 👍

src/python.rs
```diff
@@ -7,12 +7,99 @@ use crate::gil::{self, GILGuard, GILPool};
...
 
+/// Represents the major, minor, and patch (if any) versions of this interpreter.
+///
+/// See [Python::version].
+#[derive(Debug)]
+pub struct PythonVersionInfo {
+    pub major: u8,
+    pub minor: u8,
+    pub patch: u8,
+    pub suffix: Option<String>,
```
alex, line 24: This could be `Option<&'a str>` if you want to avoid the malloc.

davidhewitt: Yeah; I was trying to keep lifetimes simple on the type but perhaps it's worth doing.

alex (approved):

kngwyu:
Thanks!

kngwyu (approved):

PR #1324, opened 2020-12-17
ci: pin to working setup-python version

davidhewitt:
Workaround for https://github.com/actions/setup-python/issues/171 by using an old working version for now.


kngwyu:
Thanks, it looks like that we need this also on L71.

davidhewitt:
Looks like `setup-python` has been fixed, so this PR is no longer needed.

PR #1326, opened 2020-12-18
CHANGELOG: remove unneeded entry

davidhewitt:
Don't think we need to ever include guide changes in the changelog .


kngwyu (approved):
Yeah, it makes sense, thanks!

Issue #1206, opened 2020-09-23
Support all receiver types for all protocol methods


davidhewitt:
At the moment the protocol methods are in an inconsistent state: some of them can take `PyRef` or `PyRefMut`, and some of them take `&self` or `&mut self`.

This is confusing to users and also gets in the way in certain cases like needing to access `Python` inside a protocol method (which can be obtained from `PyRef::py()` for example) or wanting to return `PyRef<Self>`.

The simple solution is to just change all protocol methods to use `TryFromPyCell` trait. This is however a breaking change.

The _better_ solution is to change all `#[pyproto]` methods to support any of the five of `&PyCell`, `PyRef<Self>`, `PyRefMut<Self>`, `&self` and `&mut self`, just like we support for `#[pymethods]`.

I've had some ideas how to approach this second point so would like to take a shot at it soon.

kngwyu:
> The better solution is to change all #[pyproto] methods to support any of the five of &PyCell, PyRef<Self>, PyRefMut<Self>, &self and &mut self, just like we support for #[pymethods].

So what would the trait definition would be?
Or we remove `__dunder__` methods?

davidhewitt:
I think I see a way to make a `PyMethodReceiver` trait which is implemented by all five types.

davidhewitt:
Quick update from me: over the weekend I had a little time to play with a trait along the lines of:

```
pub trait PyMethodReceiver<'a>: Sized {
    fn receive<T, U>(slf: &'a PyAny, f: impl FnOnce(Self) -> T) -> PyResult<U>
    where
        T: IntoPyCallbackOutput<U>,
        U: 'static;
}
```

It _almost_ works, but there's some lifetime challenges about making sure the receiver lifetime is scoped correctly and safely. For the `&self` and `&mut self` receivers, I currently have some issues:

```
// THIS IMPL ISN'T SAFE BECAUSE &'a C IN THE CLOSURE CAN OUTLIVE THE `PYREF` GUARD
impl<'a, C: PyClass + 'a> PyMethodReceiver<'a> for &'a C {
    fn receive<T, U>(slf: &'a PyAny, f: impl FnOnce(Self) -> T) -> PyResult<U>
    where
        T: IntoPyCallbackOutput<U>,
        U: 'static
    {
        let cell: &PyCell<C> = slf.extract()?;

        // Introduce a PyRef to hold the guard. The lifetime of this is not long enough.
        let _ref = cell.try_borrow()?;

[...]

PR #1328, opened 2020-12-19
pyproto: remove inventory from implementation

davidhewitt:
This is a refactoring for `#[pyproto]` which uses dtolnay specialization instead of inventory as the underlying implementation.

The dtolnay specialization is done for a type `PyClassProtocols<T: PyClass>`. There are traits `PyObjectProtocolSlots<T>` etc. which are what get specialized. The default implementation is `PyObjectProtocolSlots<T> for &'_ PyClassProtocols<T>`, and the specialized implementation is `PyObjectProtocolSlots<T> for PyClassProtocols<T>`.

The traits with slot getter methods have been replaced by free functions to implement the slots, which the pyproto macro code directly uses.

It works out as slightly fewer LOC and also because all the protocols are known at compile time, the compiler should be able to optimise pyclass initialization better.

At the moment we lost the type checking we had with `TypedSlot`. If that's a concern I think I can reintroduce a type check in the slot defs in the macro code.

Follow-ups:
- I think I can see a way to refactor the code further to implement #1206, but I'm not entirely sure it'll work out. So I stopped here for now which works nicely and I think is a nice simplification.
- I think that we can make `inventory` an optional feature where it's only needed when users want multiple `#[pymethods]` blocks for a single PyClass. That'd be a nice simplification to remove overheads.

### DIFF
--- a/pyo3-macros-backend/src/pyproto.rs
+++ b/pyo3-macros-backend/src/pyproto.rs
@@ -107,7 +107,7 @@ fn impl_proto_impl(
         }
     }
     let normal_methods = submit_normal_methods(py_methods, ty);
-    let protocol_methods = submit_protocol_methods(method_names, ty, proto)?;
+    let protocol_methods = impl_proto_methods(method_names, ty, proto)?;
     Ok(quote! {
         #trait_impls
         #normal_methods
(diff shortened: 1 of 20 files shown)


kngwyu:
Awesome, thank you so much!
Though I don't think the inventory overhead is large, it seems good for me to have a simplified implementation.

> The traits with slot getter methods have been replaced by free functions to implement the slots, which the pyproto macro code directly uses.

Aren't these changes actually unrelated to removing inventory? I like the way it flattens the ffi functions, though.

At a glance,  this PR looks almost OK for me. 
I *feel* that we can make things simpler in `pyo3-derive-backend/src/defs.rs`, but am not sure and need to check it again.

davidhewitt:
> Aren't these changes actually unrelated to removing inventory?

Sort of - I wanted the new specialised traits to provide `&'static [PyType_Slot]` lists. I ran into various issues with `const fn` unless I flattened the getter methods away.

> I feel that we can make things simpler in pyo3-derive-backend/src/defs.rs, but am not sure and need to check it again.

I think I can hard-code some simplifications about the paths in `defs.rs` - I'll push a commit to do so.

davidhewitt:
I pushed a commit which puts the module path on the top-level `Proto` def rather than each slot def. This reduces repetition and LOC a bit.

As I think these defs might have to change again to implement #1206 in the way I imagine, I think it's not worth doing further refactoring right now.

davidhewitt:
@kngwyu ok if I merge this today before the release? It technically "breaks" any code using the `#[doc(hidden)]` public implementation details of `pyproto` (though hopefully that's nobody), so it could be nicer to release as part of a major version.

pyo3-macros-backend/src/pyproto.rs
```diff
@@ -129,49 +130,61 @@ fn submit_normal_methods(py_methods: Vec<TokenStream>, ty: &syn::Type) -> TokenS
...
-                let slot = <#ty as #ext_trait>::#get();
-                proto_methods.push(pyo3::ffi::PyType_Slot { slot: slot.0, pfunc: slot.1 as _ });
-            });
-        }
-    };
-    if tokens.len() <= 1 {
-        return Ok(quote! {});
+        return quote! {
+            impl pyo3::class::proto_methods::#slots_trait<#ty>
+            for pyo3::class::proto_methods::PyClassProtocols<#ty> {
```
kngwyu, line 149: Just a nit, but how about more `rustfmt`ic?
```suggestion
            impl pyo3::class::proto_methods::#slots_trait<#ty>
                for pyo3::class::proto_methods::PyClassProtocols<#ty> 
            {
```

pyo3-macros-backend/src/pyproto.rs
```diff
@@ -129,49 +130,61 @@ fn submit_normal_methods(py_methods: Vec<TokenStream>, ty: &syn::Type) -> TokenS
...
+        })
+        .peekable();
+
+    if tokens.peek().is_none() {
+        return TokenStream::default();
+    }
+
+    quote! {
+        impl pyo3::class::proto_methods::#slots_trait<#ty>
+        for pyo3::class::proto_methods::PyClassProtocols<#ty> {
```
kngwyu, line 184: Same as above,

pyo3-macros-backend/src/pyproto.rs
```diff
@@ -129,49 +130,61 @@ fn submit_normal_methods(py_methods: Vec<TokenStream>, ty: &syn::Type) -> TokenS
...
-                    <#ty as pyo3::class::proto_methods::HasProtoInventory>::ProtoMethods;
-                <Inventory as pyo3::class::proto_methods::PyProtoInventory>::new(
-                    { #(#tokens)* proto_methods.into() }
-                )
+
+    let mut tokens = proto
+        .slot_defs(method_names)
+        .map(|def| {
+            let slot = syn::Ident::new(def.slot, Span::call_site());
+            let slot_impl: syn::Path = syn::parse_str(def.slot_impl).unwrap();
```
kngwyu, line 168: I can understand it merely panics, but returning a `Result` would be more kind for future developers.

src/class/proto_methods.rs
```diff
@@ -1,78 +1,135 @@
...
+
+impl<T> Copy for PyClassProtocols<T> {}
 
-#[doc(hidden)]
-pub enum PyProtoMethodDef {
-    Slots(Vec<ffi::PyType_Slot>),
-    Buffer(PyBufferProcs),
+// All traits describing slots, as well as the fallback implementations for unimplemented protos
+//
+// Protos which are implented use dtolnay specialization to implement for PyClassProtocols<T>
```
kngwyu, line 36: How about adding a link to https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md?
I hope our codebase would be kind for new contributors.

src/class/proto_methods.rs
```diff
@@ -1,78 +1,135 @@
...
+}
+
+impl<T> PySequenceProtocolSlots<T> for &'_ PyClassProtocols<T> {
+    fn sequence_protocol_slots(self) -> &'static [ffi::PyType_Slot] {
+        &[]
+    }
+}
+
+// Buffer protocol has a slightly different implementation because the slots can't actually be used
+// to set the buffer protocol...
```
kngwyu, line 119: Let's be more precise, like:
```
Initializing buffer protocol via `ffi::PyType_Slot` is supported in Python 3.9.1, so we have to manually set `PyBufferProcs`.
```

kngwyu (approved):
Sure, I noted some additional suggestions.

src/class/proto_methods.rs
davidhewitt: 👍 actually I went further and make `PyBufferProtocolSlots` work using slots for `#[cfg(Py_3_9)]`, so that maybe in a few years it's easy for us to clean up this hack 😂

src/class/proto_methods.rs
davidhewitt: 👍 good idea!

pyo3-macros-backend/src/pyproto.rs
davidhewitt: Ah actually since the refactoring to `defs.rs` this only needs `syn::Ident` here, so I can remove the panic!

davidhewitt:
👍 will merge this shortly, then rebase the release and push it out!

Issue #1330, opened 2020-12-20
Compilation failed with M1 macOS, PyO3 0.12, Python 3


LER0ever:
### 🌍 Environment

 - Your operating system and version: macOS Big Sur 11.1
 - Your python version: Python 3.8 (system), Python 3.9 (homebrew), Python 3.9 (homebrew, Rosetta)
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: brew install, no.
 - Your Rust version (`rustc --version`): rustc 1.49.0-beta.4 (877c7cbe1 2020-12-10)
 - Your PyO3 version: 0.12
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: no

### 💥 Reproducing

This happens on my own project, but I can reproduce it with maturin's test-crate

- install rust beta with rustup (stable does not yet have darwin aarch64 support)
- install python3.9 aarch64 with native homebrew

#### 1. Compile with System python 3.8

```bash
git clone https://github.com/PyO3/maturin.git
cd maturin/test-crates/pyo3-pure
cargo build
```

would produce the following error:

```
  = note: Undefined symbols for architecture arm64:
            "_PyExc_ValueError", referenced from:
                _$LT$pyo3..exceptions..PyValueError$u20$as$u20$pyo3..type_object..PyTypeInfo$GT$::type_object_raw::h059c73bcf950bd88 in libpyo3-a087e80261990d6e.rlib(pyo3-a087e80261990d6e.pyo3.1al8rrrd-cgu.0.rcgu.o)
            "_PyExc_BaseException", referenced from:
                _$LT$pyo3..exceptions..PyBaseException$u20$as$u20$pyo3..type_object..PyTypeInfo$GT$::type_object_raw::h82828fb355186050 in libpyo3-a087e80261990d6e.rlib(pyo3-a087e80261990d6e.pyo3.1al8rrrd-cgu.0.rcgu.o)
            "_PyList_Append", referenced from:
                pyo3::types::list::PyList::append::_$u7b$$u7b$closure$u7d$$u7d$::h5630ce7dc966791a in libpyo3-a087e80261990d6e.rlib(pyo3-a087e80261990d6e.pyo3.1al8rrrd-cgu.4.rcgu.o)
            "_PyExc_SystemError", referenced from:
[...]

PR #1332, opened 2020-12-20
Document that macOS linker flags apply on aarch64 as well

alex:
Based on #1330


davidhewitt:
👍 thanks!

jonathan-s:
@alex I noticed that this has been removed from the docs. At least I can't really find it, I think it'd be useful to keep this, as I was searching for it and eventually found my way here.

davidhewitt:
It can still be found in the docs here: https://pyo3.rs/v0.19.1/building_and_distribution#macos

PR #1334, opened 2020-12-21
Fix the pydebug with trace refs and count_allocs for python > 3.7

cecini:
There is some change for the trace refs . So fix it .
Since COUNT_ALLOCS used rarely, leave it unimpl now.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -230,13 +230,17 @@ fn parse_header_defines(header_path: impl AsRef<Path>) -> Result<HashMap<String,
     Ok(definitions)
 }
 
-fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, String> {
+fn fix_config_map(
+    mut config_map: HashMap<String, String>,
+    version_minor: Option<u8>,
+) -> HashMap<String, String> {
     if let Some("1") = config_map.get("Py_DEBUG").as_ref().map(|s| s.as_str()) {
         config_map.insert("Py_REF_DEBUG".to_owned(), "1".to_owned());
-        config_map.insert("Py_TRACE_REFS".to_owned(), "1".to_owned());
-        config_map.insert("COUNT_ALLOCS".to_owned(), "1".to_owned());
+        if version_minor.map_or(false, |minor| minor <= 7) {
+            // Py_DEBUG only implies Py_TRACE_REFS until Python 3.7
+            config_map.insert("Py_TRACE_REFS".to_owned(), "1".to_owned());
+        }
     }
-
     config_map
 }
 
@@ -414,7 +418,7 @@ fn load_cross_compile_from_sysconfigdata(
         calcsize_pointer,
     };
 
-    Ok((interpreter_config, fix_config_map(config_map)))
+    Ok((interpreter_config, config_map))
 }
 
 fn load_cross_compile_from_headers(
@@ -446,7 +450,7 @@ fn load_cross_compile_from_headers(
         calcsize_pointer: None,
     };
 
-    Ok((interpreter_config, fix_config_map(config_map)))
+    Ok((interpreter_config, config_map))
 }
 
 fn load_cross_compile_info(
@@ -501,7 +505,7 @@ fn get_config_vars(python_path: &Path) -> Result<HashMap<String, String>> {
             memo
         });
 
-    Ok(fix_config_map(all_vars))
+    Ok(all_vars)
 }
 
 fn get_config_vars_windows(_: &Path) -> Result<HashMap<String, String>> {
@@ -529,7 +533,7 @@ fn get_config_vars_windows(_: &Path) -> Result<HashMap<String, String>> {
     // map.insert("Py_REF_DEBUG", "1");
     // map.insert("Py_TRACE_REFS", "1");
     // map.insert("COUNT_ALLOCS", 1");
-    Ok(fix_config_map(map))
+    Ok(map)
 }
 
 fn is_value(key: &str) -> bool {
@@ -889,6 +893,7 @@ fn main() -> Result<()> {
         find_interpreter_and_get_config()?
     };
 
+    config_map = fix_config_map(config_map, interpreter_config.version.minor);
     let flags = configure(&interpreter_config)?;
 
     // These flags need to be enabled manually for PyPy, because it does not expose
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ PyO3 versions, please see the [migration guide](https://pyo3.rs/master/migration
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Fixed
+- Stop including `Py_TRACE_REFS` config setting automatically if `Py_DEBUG` is set on Python 3.8 and up. [#1334](https://github.com/PyO3/pyo3/pull/1334)
+
 ## [0.13.0] - 2020-12-22
 ### Packaging
 - Drop support for Python 3.5 (as it is now end-of-life). [#1250](https://github.com/PyO3/pyo3/pull/1250)


build.rs
```diff
@@ -230,13 +230,29 @@ fn parse_header_defines(header_path: impl AsRef<Path>) -> Result<HashMap<String,
...
+    if interpreter_config.version.major == 3 && interpreter_config.version.minor.unwrap_or(0) > 7 {
+        if let Some("1") = config_map.get("Py_DEBUG").as_ref().map(|s| s.as_str()) {
+            config_map.insert("Py_REF_DEBUG".to_owned(), "1".to_owned());
+        }
+        if let Some("1") = config_map.get("Py_TRACE_REFS").as_ref().map(|s| s.as_str()) {
+            config_map.insert("Py_TRACE_REFS".to_owned(), "1".to_owned());
+        }
+        // find CFLAGS to whether contains COUNT_ALLOCS then set it as 1
+        // if interpreter_config.version.minor.unwrap_or(0) < 9 {
+        // }
```
davidhewitt, line 246: Dead code?

```suggestion
```

build.rs
```diff
@@ -230,13 +230,29 @@ fn parse_header_defines(header_path: impl AsRef<Path>) -> Result<HashMap<String,
...
+    mut config_map: HashMap<String, String>,
+    interpreter_config: &InterpreterConfig,
+) -> HashMap<String, String> {
+    if interpreter_config.version.major == 3 && interpreter_config.version.minor.unwrap_or(0) > 7 {
+        if let Some("1") = config_map.get("Py_DEBUG").as_ref().map(|s| s.as_str()) {
+            config_map.insert("Py_REF_DEBUG".to_owned(), "1".to_owned());
+        }
+        if let Some("1") = config_map.get("Py_TRACE_REFS").as_ref().map(|s| s.as_str()) {
+            config_map.insert("Py_TRACE_REFS".to_owned(), "1".to_owned());
+        }
```
davidhewitt, line 243: Can you provide some details on why this change is needed?

davidhewitt (commented):
Thanks for your contribution! 

I have a couple of questions... see my other comments...

build.rs
cecini: py >= 3.9, remove the COUNT_ALLOCS; for  COUNT_ALLOCS, there is a Cflag option -DCOUNT_ALLOCS to enable it .
py >= 3.8, enabled Py_DEBUG will   no longer enable the trace refs, meaning enabled Py_DEBUG with disabled trace refs.   If we need to trace the refs , we need configure with-trace-refs.

build.rs
cecini: Since COUNT_ALLOCS used rarely, leave it unimpl now.

build.rs
```diff
@@ -230,13 +230,29 @@ fn parse_header_defines(header_path: impl AsRef<Path>) -> Result<HashMap<String,
...
+        // }
+        return config_map;
+    }
+
     if let Some("1") = config_map.get("Py_DEBUG").as_ref().map(|s| s.as_str()) {
         config_map.insert("Py_REF_DEBUG".to_owned(), "1".to_owned());
         config_map.insert("Py_TRACE_REFS".to_owned(), "1".to_owned());
-        config_map.insert("COUNT_ALLOCS".to_owned(), "1".to_owned());
+        // find CFLAGS to whether contains the COUNT_ALLOCS then set
+        // config_map.insert("COUNT_ALLOCS".to_owned(), "1".to_owned());
```
davidhewitt, line 254: I think this code might originally have been wrong (I can't see any suggestion that `Py_DEBUG` would set `COUNT_ALLOCS` in documentation), so please just delete:

```suggestion
```

build.rs
davidhewitt: Let's delete this code then. `COUNT_ALLOCS` will still work anyway; this `fix_config_map` function was only trying to add `COUNT_ALLOCS` if `Py_Debug` was set.

build.rs
```diff
@@ -230,13 +230,29 @@ fn parse_header_defines(header_path: impl AsRef<Path>) -> Result<HashMap<String,
     Ok(definitions)
 }
 
-fn fix_config_map(mut config_map: HashMap<String, String>) -> HashMap<String, String> {
+fn fix_config_map(
+    mut config_map: HashMap<String, String>,
+    interpreter_config: &InterpreterConfig,
+) -> HashMap<String, String> {
+    if interpreter_config.version.major == 3 && interpreter_config.version.minor.unwrap_or(0) > 7 {
```
davidhewitt, line 237: I suggest the following:

```suggestion
    assert_eq!(interpreter_config.version.major, 3);
    if interpreter_config.version.minor.unwrap_or(0) >= 8 {
```

Because we only support Python 3, I think it's good to just make this code fail if we ever try to run it on Python 4.

build.rs
davidhewitt, line 243: This code looks like a no-op, please delete? It just sets `Py_TRACE_REFS` to `1` if `Py_TRACE_REFS` == `1`.

```suggestion
```

davidhewitt (changes requested):
Thanks for the explanation!

I would like to see some simplifications before we merge this. Please see my other comments.

build.rs
kngwyu: Yeah, but is this a proper place to do this assertion....?
I'd rather pass only the mimor version to this function.

build.rs
kngwyu: Where are these changes documented?

davidhewitt: I've been looking at the special builds doc for each version, e.g. https://github.com/python/cpython/blob/3.8/Misc/SpecialBuilds.txt

Also in the "what's new" documents - e.g. `COUNT_ALLOCS` removal is here: https://docs.python.org/3/whatsnew/3.9.html#build-changes

build.rs
cecini: Yes,  I also said according the CFlags to decide whether to set COUNT_ALLOCS as 1.
But always we build python with pydebug, then  in this case , we then  can decide to enable COUNT_ALLOCS.  So i place the comment in the if Py_DEBUG block. 
Of course, if decide never support COUNT_ALLOCS( we know > 3.9 have remove this code), we just delete this code.

build.rs
davidhewitt: Passing just the minor version instead of the full `interpreter_config` to this function also seems like a good idea!

build.rs
cecini: This code is for 3.8 only. according the CFlags to decide whether to set COUNT_ALLOCS as 1., no by the Py_DEBUG, 
Again, Of course, if  we decide never support COUNT_ALLOCS( we know >= 3.9 have remove COUNT_ALLOCS  this code), we just delete this comment.

build.rs
cecini: Oh, you are right. get config can get this info. I will fix.

build.rs
cecini: will we support python 4? if support, we should keep the interpreter_config as the function argument. and leave room for the possible change in the python 4.

cecini:
refactor the pr code： pass the minor version; fix  some code by suggestion.
Again, if we plan not support count_allocs,  must delete the last comment in the pr. 
Please give your suggestions.

build.rs
```diff
@@ -230,13 +230,25 @@ fn parse_header_defines(header_path: impl AsRef<Path>) -> Result<HashMap<String,
...
         config_map.insert("Py_REF_DEBUG".to_owned(), "1".to_owned());
         config_map.insert("Py_TRACE_REFS".to_owned(), "1".to_owned());
-        config_map.insert("COUNT_ALLOCS".to_owned(), "1".to_owned());
     }
 
+    // py3.9 remove the COUNT_ALLOCS related code
+    // if interpreter_config.version.minor.unwrap_or(0) <= 8 {
+    //     // find CFLAGS to whether contains the COUNT_ALLOCS then set
+    //     // config_map.insert("COUNT_ALLOCS".to_owned(), "1".to_owned());
+    // }
```
davidhewitt, line 251: Yes let's delete this 👍 

```suggestion
```

build.rs
```diff
@@ -230,13 +230,25 @@ fn parse_header_defines(header_path: impl AsRef<Path>) -> Result<HashMap<String,
...
+    version_minor: Option<u8>,
+) -> HashMap<String, String> {
+    // defining the Py_DEBUG macro no longer implies the Py_TRACE_REFS macro by py38
+    if version_minor.unwrap_or(0) >= 8 {
+        if let Some("1") = config_map.get("Py_DEBUG").as_ref().map(|s| s.as_str()) {
+            config_map.insert("Py_REF_DEBUG".to_owned(), "1".to_owned());
+        }
+    } else if let Some("1") = config_map.get("Py_DEBUG").as_ref().map(|s| s.as_str()) {
         config_map.insert("Py_REF_DEBUG".to_owned(), "1".to_owned());
         config_map.insert("Py_TRACE_REFS".to_owned(), "1".to_owned());
```
davidhewitt, line 244: I suggest this version with less repetition:

```suggestion
    if let Some("1") = config_map.get("Py_DEBUG").as_ref().map(|s| s.as_str()) {
        config_map.insert("Py_REF_DEBUG".to_owned(), "1".to_owned());
        if version_minor.map_or(false, |minor| minor <= 7) {
            // Py_DEBUG only implies Py_TRACE_REFS until Python 3.7
            config_map.insert("Py_TRACE_REFS".to_owned(), "1".to_owned());
        }
```

davidhewitt (commented):
Perfect, thanks! See my final suggestions, then let's merge 👍

build.rs
cecini: Nice.

davidhewitt (approved):
Thank you very much for this, looks good!

I'm going to push a CHANGELOG entry to this branch and then will proceed to merge.

Issue #1321, opened 2020-12-16
Integration tests can't find crate


OrangeTux:
This is not a bug in PyO3, but it can hit Pyo3 users nonetheless. 

I'm using PyO3 for a project and wrote some integration tests. I put those tests inside a `tests/` directory at the crate root. However my tests fail to compile because my crate couldn't be found:
   
``` bash
 error[E0463]: can't find crate for `yolo`
   --> tests/tests.rs:4:1
    |
  4 | use yolo;
    | ^^^^^^^^^ can't find crate
```

It turns out that `cdylib` crate types can't use tests inside a `tests/` directory. See https://github.com/rust-lang/cargo/issues/6659 

I think this should be documented, as this is not very obvious. What would be the best place to do that? Inside the FAQ?

kngwyu:
Thanks.

> Inside the FAQ?

Yeah, I think FAQ & Troubleshooting is good and am really appriciate your PR.

OrangeTux:
Sure. I just created a PR. 

Happy holidays!

PR #1336, opened 2020-12-22
Add note to FAQ about failing integration tests

OrangeTux:
Integration tests of a PyO3 project can't be put inside a `tests/`
directory. Due to a known problem Cargo fails to import the crate.

Fixes: #1321

### DIFF
--- a/guide/src/faq.md
+++ b/guide/src/faq.md
@@ -28,6 +28,22 @@ extension-module = ["pyo3/extension-module"]
 default = ["extension-module"]
 ```
 
+## I can't run `cargo test`: Compiler fails to import crate for tests in `tests/` directory!
+
+The Rust book suggests to [put integration tests inside a `tests/` directory](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests). However, you can't do that for a PyO3 project. The compiler wouldn't be able to find your crate and displays an error like this:
+
+```
+error[E0463]: can't find crate for `my-crate`
+   --> tests/tests.rs:4:1
+    |
+  4 | use crate;
+    | ^^^^^^^^^ can't find crate
+```
+
+This is not a bug in PyO3. But it's a [known problem of Cargo](https://github.com/rust-lang/cargo/issues/6659).
+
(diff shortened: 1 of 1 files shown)


guide/src/faq.md
```diff
@@ -28,6 +28,22 @@ extension-module = ["pyo3/extension-module"]
...
+error[E0463]: can't find crate for `my-crate`
+   --> tests/tests.rs:4:1
+    |
+  4 | use crate;
+    | ^^^^^^^^^ can't find crate
+```
+
+This is not a bug in PyO3. But it's a [known problem of Cargo](https://github.com/rust-lang/cargo/issues/6659).
+
+For now you can work around it by putting the integration tests inside the `src/` directory.
```
davidhewitt, line 45: I think an alternative workaround is also to add `rlib` to the `crate-types` in your `Cargo.toml`?

davidhewitt (commented):
Thanks for the PR! I think there might be one other alternative...

guide/src/faq.md
kngwyu: Agreed.

OrangeTux: That is also what a colleague suggested. However that yielded a new error:

``` 
  = note: /usr/bin/ld: /home/developer/projects/pyo3-ftp/target/debug/deps/libpyo3-9f95204c15f7bad5.rlib(ceval.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
          /usr/bin/ld: final link failed: Bad value
          collect2: error: ld returned 1 exit status
``` 

Any idea if this could that also happen to others? Or is this problem isolated for my project?

kngwyu: Ah, I understand, thanks.
It looks like that we cannot link Python with rlib.
How about lib?

davidhewitt: GLIBC 2.2.5 is apparently from 2002... what OS are you using / where did you get the python interpreter from?

https://sourceware.org/glibc/wiki/Glibc%20Timeline

Issue #1289, opened 2020-11-20
FFI module cleanup


davidhewitt:
The modules in `src/ffi` have become a bit out-of-date, and also quite disorganised so it's kinda hard to figure out what's missing.

Since Python 3.9 the cpython include files have been split so that (roughly) the limited api is in `cpython/Include/` and the unlimited api is in `cpython/Include/cpython/`. I propose for simplicity in maintaining the `pyo3::ffi` module we should reorganise code to match that structure.

I suggest that we do the following things:

- [ ] reorganise `ffi/mod.rs` to be alphabetical and have `// skipped foo.h` comments for each header present in cpython master which is not present in our bindings at the moment.
- [ ] add `ffi/cpython/mod.rs` to contain the unlimited api headers, and do the same as with `ffi/mod.rs`.
- [ ] for each file in `ffi/`, reorganise the definitions in it appropriately into `ffi` and `ffi/cpython` parts which match the cpython master definitions. This should include:
  - Put the existing definitions in the same order as upstream cpython.
  - Either:
    - Add `// skipped foo` for each definition `foo` we don't currently have in our definitions.
    - Add new definition for each we don't currently have in our definitions.

The idea with each `// skipped foo` being in alphabetical order is that it's then hopefully easier for us to compare against upstream in the future.

Perhaps in general the above guidelines of how to organise our ffi modules should go in CONTRIBUTING.md ?

nw0:
I've started implementing this, and ran into a question:

Do we want to mark the `Py_DEPRECATED` functions in any way? I'm not sure whether this is useful or desired in PyO3, but this would be a convenient time to do it.

davidhewitt:
@nw0 thanks so much for being willing to help us out! In response to your questions:

1. With `Py_DEPRECATED` functions I think:
   - any we currently have ffi bindings for, we should add `#[deprecated()]` also.
   - any that we are missing, let's skip adding.
 
2. With `#ifndef Py_LIMITED_API` inside the `cpython/Include/*.h` headers - that's quite suprising as I thought that those headers are only `#include`-d if it's not `Py_LIMITED_API`. I guess we'll have to match anything that `cpython` does with those definitions. If you want to file upstream PRs against cpython, the whole ecosystem will be extremely grateful for you helping us to keep things tidy!

[...]

PR #1338, opened 2020-12-24
ffi module cleanup

nw0:
Partial fix for #1289, as suggested in https://github.com/PyO3/pyo3/issues/1289#issuecomment-750367548


src/ffi/ceval.rs
```diff
@@ -1,7 +1,5 @@
-#[cfg(not(Py_LIMITED_API))]
-use crate::ffi::code::FreeFunc;
 use crate::ffi::object::PyObject;
-use crate::ffi::pystate::{PyThreadState, Py_tracefunc};
+use crate::ffi::pystate::PyThreadState;
 use std::os::raw::{c_char, c_int, c_void};
 
 extern "C" {
```
nw0, line 5: Can't select the exact line, but PyEval_CallObjectWithKeywords is marked deprecated in 3.9. However, if I mark it with `#[cfg_attr(Py_3_9, deprecated(note = "Python 3.9"))]`, I can't compile due to `deny(warnings)` on this crate (it's used by `PyEval_CallObject` below).

Maybe we could consider being a bit more specific with the crate's `deny` attribute? https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md

src/ffi/code.rs
```diff
@@ -1,156 +1,2 @@
...
-    (Py_TYPE(op) == &mut PyCode_Type) as c_int
-}
-
-#[inline]
-#[cfg(not(PyPy))]
-pub unsafe fn PyCode_GetNumFree(op: *mut PyCodeObject) -> Py_ssize_t {
-    crate::ffi::tupleobject::PyTuple_GET_SIZE((*op).co_freevars)
-}
+// TODO: newtype
+pub use crate::ffi::PyCodeObject;
```
nw0, line 2: The definition in `Include/code.h` is `typedef struct PyCodeObject PyCodeObject`, where the actual def is in `Include/cpython/code.h`.

I plan to change this into a newtype, but this requires some other code changes I'm not prepared to make yet. Just making a note here.

nw0:
Hmm, I'm not sure what to make of 
```
 Falsifying example: test_datetime_from_timestamp(
    dt=datetime.datetime(3001, 1, 19, 0, 0, fold=1),
)
```
(from failing python3.6-x86 windows-latest). Is this something you know about or do I need to try smaller changes?

src/ffi/code.rs
kngwyu: I think what we need is 
```rust
#[cfg(Py_LIMITED_API)]
opaque_struct!(PyCodeObject);
```
Also, we have to remove `#[cfg(not(Py_LIMITED_API)]` from `mod code;`.

src/ffi/mod.rs
```diff
@@ -75,6 +75,56 @@ pub use self::weakrefobject::*;
...
+mod bytesobject;
+// skipped cellobject.h
+mod ceval; // TODO supports PEP-384 only; needs adjustment for Python 3.3 and 3.5
+
+// skipped classobject.h
+mod codecs; // TODO supports PEP-384 only; needs adjustment for Python 3.3 and 3.5
+#[cfg(Py_LIMITED_API)]
+mod code {}
+#[cfg(not(Py_LIMITED_API))]
+mod code;
```
kngwyu, line 93: See the above comment.
```suggestion
mod code {}
```

kngwyu (changes requested):
Thanks!
I don't think the datetime test failure is related to this change and we need to file it as a separate issue.

src/ffi/code.rs
nw0: I've made this change

kngwyu: Also, let's remove `pub use crate::ffi::PyCodeObject;`.
Since we import all items in `mod.rs`, this is not so mearningful.

src/ffi/ceval.rs
davidhewitt: Thanks, I recently added this and I agree it's now an antipattern.

I've opened #1340 to remove the attribute.

As specifically here, I think we will need to put `#[allow(deprecated)]` on the call inside `PyEval_CallObject` so that this passes CI. Mildly annoying, but hey!

davidhewitt (approved):
Thanks very much for kicking this efforts off! ❤️ 🚀

I think once we've got that last deprecated attribute in place, this batch of cleanups is ready to merge? 🙇

davidhewitt:
The `deprecated` stuff should work better if you rebase on master.

Also, if you are willing to add a CHANGELOG entry documented the new and deprecated APIs, that'd be excellent 🙏

src/ffi/ceval.rs
nw0: This is fine, I think. I think the only gotcha is that we need to add the deprecation to `PyEval_CallObject` as well.

nw0:
All done, I think. Added changelog entry for deprecations (not adding any functions yet). Thanks!

kngwyu:
Thanks!

kngwyu (approved):

Issue #1298, opened 2020-11-28
DeprecationWarning: builtin type X has no __module__ attribute


davidhewitt:
On Python 3.7 and up I see warnings of the following form when running the pyo3 test suite:

```
DeprecationWarning: builtin type X has no __module__ attribute
```

I _think_ this is probably related to using a `#[pyclass]` when not adding it to a module. I haven't investigated yet.

Example test which demonstrates this issue:

```
$ cargo test --test test_various return_custom_class
    Finished test [unoptimized + debuginfo] target(s) in 0.03s
     Running target/debug/deps/test_various-a73ba5b7affaefde

running 1 test
<string>:1: DeprecationWarning: builtin type PyUsize has no __module__ attribute
test return_custom_class ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 7 filtered out
```

I'm not sure what changes we need to make to resolve.

ChillFish8:
> I think this is probably related to using a #[pyclass] when not adding it to a module. I haven't investigated yet.
This does seem to be the trend with the issue when re-producing, when exposed to the module this error goes away. Only when you dont expose it directly does this warning get raised from my findings

bderrett:
This appears to cause an error when running doctests:
```
$ python -m doctest -v myfile.py
Traceback (most recent call last):
  File "/home/ben/miniconda3/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/ben/miniconda3/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/ben/miniconda3/lib/python3.8/doctest.py", line 2793, in <module>
    sys.exit(_test())
  File "/home/ben/miniconda3/lib/python3.8/doctest.py", line 2783, in _test
    failures, _ = testmod(m, verbose=verbose, optionflags=options)
  File "/home/ben/miniconda3/lib/python3.8/doctest.py", line 1955, in testmod
    for test in finder.find(m, name, globs=globs, extraglobs=extraglobs):
  File "/home/ben/miniconda3/lib/python3.8/doctest.py", line 939, in find
    self._find(tests, obj, name, module, source_lines, globs, {})
  File "/home/ben/miniconda3/lib/python3.8/doctest.py", line 1000, in _find
    self._from_module(module, val)):
  File "/home/ben/miniconda3/lib/python3.8/doctest.py", line 967, in _from_module
[...]

PR #1341, opened 2020-12-27
ffi module cleanup: context.h to frameobject.h

nw0:
Partial fix for #1289. This set seems relatively more straightforward, but touches a lot of files.

PR #1342, opened 2020-12-27
abi3: add support for dict and weakref from Python 3.9

davidhewitt:
This PR adds support for `#[pyclass(dict)]` and `#[pyclass(weakref)]` from Python 3.9 and up.

I was looking into how to do #1298 when I noticed this was possible! :rocket:

### DIFF
--- a/src/pyclass.rs
+++ b/src/pyclass.rs
@@ -178,6 +178,14 @@ where
     slots.maybe_push(ffi::Py_tp_new, new.map(|v| v as _));
     slots.maybe_push(ffi::Py_tp_call, call.map(|v| v as _));
 
+    #[cfg(Py_3_9)]
+    {
+        let members = py_class_members::<T>();
+        if !members.is_empty() {
+            slots.push(ffi::Py_tp_members, into_raw(members))
+        }
+    }
+
     // normal methods
     if !methods.is_empty() {
         slots.push(ffi::Py_tp_methods, into_raw(methods));
@@ -203,7 +211,7 @@ where
         basicsize: std::mem::size_of::<T::Layout>() as c_int,
         itemsize: 0,
         flags: py_class_flags::<T>(has_gc_methods),
-        slots: slots.0.as_mut_slice().as_mut_ptr(),
+        slots: slots.0.as_mut_ptr(),
     };
 
     let type_object = unsafe { ffi::PyType_FromSpec(&mut spec) };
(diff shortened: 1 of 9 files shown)


src/pyclass.rs
```diff
@@ -357,11 +412,21 @@ fn py_class_properties<T: PyClass>() -> Vec<ffi::PyGetSetDef> {
...
+            name: "__dict__\0".as_ptr() as *mut c_char,
+            get: Some(ffi::PyObject_GenericGetDict),
+            set: Some(ffi::PyObject_GenericSetDict),
+            doc: ptr::null_mut(),
+            closure: ptr::null_mut(),
+        });
     }
     if !props.is_empty() {
-        props.push(ffi::PyGetSetDef_INIT);
+        props.push(unsafe { std::mem::zeroed() });
```
kngwyu, line 429: Why do we need these changes?

src/pyclass.rs
```diff
@@ -333,6 +348,46 @@ fn py_class_method_defs<T: PyMethods>() -> (
...
+    // weakref support
+    if let Some(weakref_offset) = PyCell::<T>::weakref_offset() {
+        members.push(ffi::structmember::PyMemberDef {
+            name: WEAKLISTOFFSET.as_ptr() as _,
+            type_code: ffi::structmember::T_PYSSIZET,
+            offset: weakref_offset as _,
+            flags: ffi::structmember::READONLY,
+            doc: std::ptr::null_mut(),
+        });
+    }
```
kngwyu, line 382: These two `PyMemberDef` have many common fields and you might want to shorten the code by using a closure or function.

kngwyu (approved):
Thanks!

src/pyclass.rs
davidhewitt: I'll use a macro (so that the optimizer doesn't have to try to eliminate the function call overhead).

src/pyclass.rs
davidhewitt: See comment at https://github.com/PyO3/pyo3/pull/1341#discussion_r549083130

Basically `PyGetSetDef_INIT` and `PyGetSetDef_DICT` are ffi definitions we invented ourselves; we should deprecate and remove them so that our `ffi` module matches Python's headers.

src/pyclass.rs
kngwyu: How about `const fn`?

src/pyclass.rs
kngwyu: Thanks, it makes sense.

src/pyclass.rs
davidhewitt: `const fn` won't work here, because the `T: PyClass` generic type bound is unstable with `const fn`.

kngwyu: I just meant:
```rust
pub const fn member_def(name: *mut c_char, offset: Py_ssize_t) -> PyMemberDef {
    PyMemberDef {
        name,
        type_code: crate::ffi::structmember::T_PYSSIZET,
        offset,
        flags: crate::ffi::structmember::READONLY,
        doc: std::ptr::null_mut(),
    }
}
```

PR #1343, opened 2020-12-27
pyclass: fix deprecation warning for no __module__ attribute

### LINKED ISSUE #1298: DeprecationWarning: builtin type X has no __module__ attribute

davidhewitt:
Closes #1298 .

This ensures that `__module__` is always set (usually to "builtins", which is consistent with what we already do).


kngwyu (approved):
Nice catch, thank you!

PR #1345, opened 2020-12-28
Use const fn instead of macro_rules to construct PyMemberDef

kngwyu:
A tweak for #1342


davidhewitt:
I mean, sure, but the compiler can't actually run this code at compile time, because `PyCell::<T>::dict_offset` (e.g.) isn't `const fn`. So this might as well be a closure - it'll be called at runtime either way.

kngwyu:
Ah, sorry, now I understand what you said.

kngwyu:
Reopened with `#[inline(always)]`, but I don't have a strong opinion about inline vs macro so rejecting is OK.

PR #1346, opened 2020-12-28
Use inline fn instead of macro_rules to construct PyMemberDef

kngwyu:
A tweak for #1342.
As I commented in #1345, I don't have any strict rule about macro vs inline fn so rejecting is OK if you don't like this.


davidhewitt:
It's one line shorter, let's use it :joy:

davidhewitt (approved):

PR #1347, opened 2020-12-28
auto-initialize: new feature to control initializing Python

### LINKED ISSUE #742: test suite failures and missing symbols on macOS

### LINKED ISSUE #762: Importing Python random lead to an error

### LINKED ISSUE #1213: Failing tests on master
## 🐛 Bug Reports

After I cloned the repo for the first time I ran the tests to check for errors and I got two. I'm reporting them here by suggestion of @kngwyu .

```
failures:

---- buffer::test::test_array_buffer stdout ----
thread 'buffer::test::test_array_buffer' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: 0x55fb0678ffe0 }', src/buffer.rs:675:14

---- exceptions::test::test_check_exception stdout ----
thread 'exceptions::test::test_check_exception' panicked at 'Can not import module: socket: PyErr { type: 0x55fb0678ffe0 }', src/exceptions.rs:499:5


failures:
    buffer::test::test_array_buffer
    exceptions::test::test_check_exception
```

### 🌍 Environment

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.5 LTS
Release:        18.04
Codename:       bionic
$ uname -rsm       
Linux 5.4.0-39-generic x86_64
```
 - Your python version: 
```
$ python --version
Python 2.7.17
$ python3 --version
Python 3.8.3
```
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: Python 2 - apt, Python 3 - Compiled from source, installed to system.
 - Your Rust version (`rustc --version`):
```
$ rustc --version
rustc 1.48.0-nightly (8876ffc92 2020-10-02)
```
 - Your PyO3 version: Master @ 37ce406ba18eb4f42bebf14c4bf33596f1f0dc3b
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: See above

### 💥 Reproducing

I ran:
```
$ cargo test
```

davidhewitt:
This PR add a new feature `auto-initialize` to control whether `Python::with_gil` and `Python::acquire_gil` call `prepare_freethreaded_python` automatically.

I wanted to add this for a couple of reasons:
- When building an `extension-module` then this functionality is not needed.
- When linking against a static python lib (or PyPy) then this functionality doesn't even work correctly.

The PR checks for the static python and PyPy cases and removes the `prepare_freethreaded_python` API, and instead emits a `compile_error!` if the `auto-initialize` feature is enabled.

Closes #742
Closes #762
Closes #1213 

---

**TODO:**

It would be nice to make this feature opt-in instead of opt-out, but that's a breaking change so we should delay that until 0.14 release.

I'd like to write some more documentation for this in the guide before merging. 

Also needs CHANGELOG entry.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,27 +85,28 @@ jobs:
         run: echo LD_LIBRARY_PATH=${pythonLocation}/lib >> $GITHUB_ENV
 
       - name: Build docs
-        run: cargo doc --features "num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
+        run: cargo doc --no-default-features --features "macros num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
 
-      - name: Build without default features
+      - name: Build (no features)
         run: cargo build --no-default-features --verbose --target ${{ matrix.platform.rust-target }}
 
-      - name: Build with default features
-        run: cargo build --features "num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
+      - name: Build (all additive features)
+        run: cargo build --no-default-features --features "macros num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
 
       # Run tests (except on PyPy, because no embedding API).
       - if: matrix.python-version != 'pypy-3.6'
         name: Test
-        run: cargo test --features "num-bigint num-complex hashbrown" --target ${{ matrix.platform.rust-target }}
+        run: cargo test --no-default-features --features "macros num-bigint num-complex hashbrown" --target ${{ matrix.platform.rust-target }}
+
       # Run tests again, but in abi3 mode
       - if: matrix.python-version != 'pypy-3.6'
         name: Test (abi3)
-        run: cargo test --no-default-features --features "abi3,macros" --target ${{ matrix.platform.rust-target }}
+        run: cargo test --no-default-features --features "abi3 macros num-bigint num-complex hashbrown" --target ${{ matrix.platform.rust-target }}
 
       # Run tests again, for abi3-py36 (the minimal Python version)
       - if: (matrix.python-version != 'pypy-3.6') && (matrix.python-version != '3.6')
         name: Test (abi3-py36)
-        run: cargo test --no-default-features --features "abi3-py36,macros" --target ${{ matrix.platform.rust-target }}
+        run: cargo test --no-default-features --features "abi3-py36 macros num-bigint num-complex hashbrown" --target ${{ matrix.platform.rust-target }}
 
       - name: Test proc-macro code
         run: cargo test --manifest-path=pyo3-macros-backend/Cargo.toml --target ${{ matrix.platform.rust-target }}
@@ -125,6 +126,9 @@ jobs:
     env:
       RUST_BACKTRACE: 1
       RUSTFLAGS: "-D warnings"
+      # TODO: this is a hack to workaround compile_error! warnings about auto-initialize on PyPy
+      # Once cargo's `resolver = "2"` is stable (~ MSRV Rust 1.52), remove this.
+      PYO3_CI: 1
 
   coverage:
     needs: [fmt]
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -3,11 +3,11 @@
 //! Interaction with python's global interpreter lock
 
 use crate::{ffi, internal_tricks::Unsendable, Python};
-use parking_lot::{const_mutex, Mutex};
+use parking_lot::{const_mutex, Mutex, Once};
 use std::cell::{Cell, RefCell};
-use std::{mem::ManuallyDrop, ptr::NonNull, sync};
+use std::{mem::ManuallyDrop, ptr::NonNull};
 
-static START: sync::Once = sync::Once::new();
+static START: Once = Once::new();
 
 thread_local! {
     /// This is a internal counter in pyo3 monitoring whether this thread has the GIL.
@@ -45,16 +45,20 @@ pub(crate) fn gil_is_acquired() -> bool {
 /// If both the Python interpreter and Python threading are already initialized,
 /// this function has no effect.
 ///
+/// # Availability
+///
+/// This function is only available when linking against Python distributions that contain a
+/// shared library.
+///
+/// This function is not available on PyPy.
+///
 /// # Panic
 /// If the Python interpreter is initialized but Python threading is not,
 /// a panic occurs.
 /// It is not possible to safely access the Python runtime unless the main
 /// thread (the thread which originally initialized Python) also initializes
 /// threading.
-///
-/// When writing an extension module, the `#[pymodule]` macro
-/// will ensure that Python threading is initialized.
-///
+#[cfg(all(Py_SHARED, not(PyPy)))]
 pub fn prepare_freethreaded_python() {
     // Protect against race conditions when Python is not yet initialized
     // and multiple threads concurrently call 'prepare_freethreaded_python()'.
@@ -72,22 +76,18 @@ pub fn prepare_freethreaded_python() {
             // Note that the 'main thread' notion in Python isn't documented properly;
             // and running Python without one is not officially supported.
 
-            // PyPy does not support the embedding API
-            #[cfg(not(PyPy))]
-            {
-                ffi::Py_InitializeEx(0);
+            ffi::Py_InitializeEx(0);
 
-                // Make sure Py_Finalize will be called before exiting.
-                extern "C" fn finalize() {
-                    unsafe {
-                        if ffi::Py_IsInitialized() != 0 {
-                            ffi::PyGILState_Ensure();
-                            ffi::Py_Finalize();
-                        }
+            // Make sure Py_Finalize will be called before exiting.
+            extern "C" fn finalize() {
+                unsafe {
+                    if ffi::Py_IsInitialized() != 0 {
+                        ffi::PyGILState_Ensure();
+                        ffi::Py_Finalize();
                     }
                 }
-                libc::atexit(finalize);
             }
+            libc::atexit(finalize);
 
             // > Changed in version 3.7: This function is now called by Py_Initialize(), so you don’t have
             // > to call it yourself anymore.
@@ -95,11 +95,10 @@ pub fn prepare_freethreaded_python() {
             if ffi::PyEval_ThreadsInitialized() == 0 {
                 ffi::PyEval_InitThreads();
             }
-            // PyEval_InitThreads() will acquire the GIL,
-            // but we don't want to hold it at this point
+
+            // Py_InitializeEx() will acquire the GIL, but we don't want to hold it at this point
             // (it's not acquired in the other code paths)
             // So immediately release the GIL:
-            #[cfg(not(PyPy))]
             let _thread_state = ffi::PyEval_SaveThread();
             // Note that the PyThreadState returned by PyEval_SaveThread is also held in TLS by the Python runtime,
             // and will be restored by PyGILState_Ensure.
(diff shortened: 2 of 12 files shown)


kngwyu:
Thanks!
I really like the idea to raise a compile error before failing to link, but I need some clarifications:
- About `prepare_freethreaded_python`:
  - I agree that this should be disabled for PyPy or static linking. Then, how about introducing some new configuration(e.g., `Py_SHARED`) and mark `#[cfg(not(any(PyPy, Py_SHARED)))] fn prepare_freethreaded_python`? I'm not sure this should be a feature flag.
- About `auto-initialize`
  - Isn't this a `not(extension-module)`? Why do we need a new feature?

Again, this is a difficult problem, and thank you for tackling it.

davidhewitt:
Thanks for the questions. I'm very happy to change the design if we think alternatives are better. Here's my thinking about your proposed alternatives:

- using `Py_SHARED` flag, I don't think we can fail compilation in `build.rs`. This means the error message when building will be that `prepare_freethtreaded_python` is missing. A nicer error than link failures maybe, especially if we add documentation to that function do that users understand what might have happened.

  The failure will still be quite hard to diagnose for the user compared to a friendly error message from the build script.

- With `auto-initialize` being `not(extension-module)` - I suppose in practice yes, this is true. But I think this would then be another case of #771 because then `extension-module` feature is taking away functionality again.

  Also, this would mean that users who want to run `cargo test` which relies on auto-initialize would have to _always_ use the `--no-default-features` workaround to disable `extension-module`, even after #1123 is stable. It seems more idiomatic Rust to me to be able to control this properly with a separate feature flag.

  The feature flag is also more general. For example, in some special cases users might want to embed python and manually initialize it before calling PyO3 code. In those cases they don't need `auto-initialize` functionality either.

davidhewitt:
... I'm thinking that maybe we can use `Py_SHARED` plus the `auto-initialize` feature. We can use `compile_error!` to produce a friendly error message, so my argument above for the `embedding` feature now seems weak.

davidhewitt:
Reworked this PR as discussed but I expect CI to continue to fail until #1350 is merged, because one of the simplifications in that PR fixes how Py_SHARED would be set.

davidhewitt:
Rebased and rebuilt this PR as a single commit; hopefully CI is green this time 🤞

src/gil.rs
```diff
@@ -137,7 +136,56 @@ impl GILGuard {
...
+        //    auto-initialize so this avoids breaking existing builds.
+        //  - Otherwise, just check the GIL is initialized.
+        cfg_if::cfg_if! {
+            if #[cfg(all(feature = "auto-initialize", Py_SHARED, not(PyPy)))] {
+                prepare_freethreaded_python();
+            } else if #[cfg(all(feature = "auto-initialize", not(feature = "extension-module"), not(Py_SHARED), not(rustdoc)))] {
+                compile_error!(concat!(
+                    "The `auto-initialize` feature is not supported when linking Python ",
+                    "statically instead of with a shared library.\n",
+                    "\n",
```
kngwyu, line 152: How about making this a bit shorter?
```suggestion
                    "statically instead of with a shared library.\n\n",
```

src/gil.rs
```diff
@@ -137,7 +136,56 @@ impl GILGuard {
...
+        // Maybe auto-initialize the GIL:
+        //  - If auto-initialize feature set and supported, try to initalize the interpreter.
+        //  - If the auto-initialize feature is set but unsupported, emit hard errors only when
+        //    the extension-module feature is not activated - extension modules don't care about
+        //    auto-initialize so this avoids breaking existing builds.
+        //  - Otherwise, just check the GIL is initialized.
+        cfg_if::cfg_if! {
+            if #[cfg(all(feature = "auto-initialize", Py_SHARED, not(PyPy)))] {
+                prepare_freethreaded_python();
+            } else if #[cfg(all(feature = "auto-initialize", not(feature = "extension-module"), not(Py_SHARED), not(rustdoc)))] {
```
kngwyu, line 148: I think we can make this simpler by `else if all(auto-initialize, PyPy) { ... } else if all(auto_initialize, ...)`

.github/workflows/ci.yml
```diff
@@ -84,28 +84,31 @@ jobs:
...
-      - name: Build docs
-        run: cargo doc --features "num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
+      # Validate docs have no warnings (except on PyPy; auto-initialize feature fails the compile until `cfg(rustdoc)` is stable).
+      - if: matrix.python-version != 'pypy-3.6'
+        name: Build docs
+        run: cargo doc --no-default-features --features "macros num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
 
-      - name: Build without default features
+      - name: Build (no features)
         run: cargo build --no-default-features --verbose --target ${{ matrix.platform.rust-target }}
```
kngwyu, line 93: It looks like that `auto-initialize` and `macros` are enabled in this build, though I don't know why 😕

.github/workflows/ci.yml
```diff
@@ -84,28 +84,31 @@ jobs:
...
-      - name: Build with default features
-        run: cargo build --features "num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
+      - name: Build (all additive features)
+        run: cargo build --no-default-features --features "macros num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
 
       # Run tests (except on PyPy, because no embedding API).
       - if: matrix.python-version != 'pypy-3.6'
         name: Test
-        run: cargo test --features "num-bigint num-complex hashbrown" --target ${{ matrix.platform.rust-target }}
+        run: cargo test --no-default-features --features "macros num-bigint num-complex hashbrown" --target ${{ matrix.platform.rust-target }}
```
kngwyu, line 101: Why not using `auto-initialize`?

kngwyu (approved):
LGTM, thanks!
CI failures are because `auto-initialise` is unintentionally invoked. I'm sorry but I have no idea why this happens.

src/gil.rs
davidhewitt: `not(extension-module)` is unfortunately necessary to avoid a breaking change - otherwise all extension modules would suddenly start failing. We could make it as you suggest in `0.14`.

.github/workflows/ci.yml
davidhewitt: It's automatically included in tests because I put this in `[dev-dependencies]`:

```
# features needed to run the PyO3 test suite
pyo3 = { path = ".", default-features = false, features = ["macros", "auto-initialize"] }
```

However, that's also the reason why `auto-initialize` and `macros` are included in the `cargo build` command above. Interestingly `cargo` is about to stabilize a fix for this as a new `resolver = "2"` option in Cargo.toml. https://github.com/rust-lang/cargo/pull/8997 (I tested on nightly and the features are no longer included in `cargo build`.)

Because in the future what I've done here _will_ work correctly in the future, I'm going to keep the code mostly as-is for the moment and introduce a hack to avoid the immediate issue with a TODO to tidy it up once we bump MSRV in the future (probably to Rust 1.52).

.github/workflows/ci.yml
davidhewitt: It's because of the `Cargo.toml` changes - see other comment.

src/gil.rs
kngwyu: Ah, thanks. That makes sense.
BTW, do you think that we should keep auto-initialize in default-features?

davidhewitt: I think in the long-run it's better to remove `auto-initialize` from default-features, but that's also a breaking change, so let's remove it in `0.14`.

I'll create an issue to track these cleanups and add it to the `0.14` milestone.

davidhewitt:
@kngwyu I pushed a CHANGELOG entry and a new guide page `guide/src/features.md` to document this functionality.

I also added a tiny "hack" with `cfg(__pyo3_ci)` to suppress the error message in our CI until we are able to bump MSRV to a Rust version which includes the cargo `resolver = "2"`. Might be nearly a year though - I put `TODO` on the location in `ci.yml` and `build.rs` where I added the hack.

If you are happy with this, I think this is ready to merge! 🚀

kngwyu:
> I also added a tiny "hack" with cfg(__pyo3_ci) to suppress the error message in our CI until we are able to bump MSRV to a Rust version which includes the cargo resolver = "2". Might be nearly a year though - I put TODO on the location in ci.yml and build.rs where I added the hack.

Is it much simpler to remove the dev-dependency for now?

src/gil.rs
kngwyu: 👍🏼

davidhewitt:
If we remove the `dev-dependency` and also remove `auto-initialize` from default features then to run the PyO3 test suite we'll have to write `cargo test --features auto-initialize`.

I'm sure I'll forget to add the feature a lot and it's also annoying to type something so long.

I think it's better to leave the very small hack in so that we can make exactly the API we want for users. Especially as we know the fix which removes the need for the hack is going to be stabilised very soon.

(Because we only need the hack to fix the pypy CI jobs, we only need to wait until it's in the most recent stable rust version.)

kngwyu:
> If we remove the dev-dependency and also remove auto-initialize from default features

OK, if you are to release 0.14 before the next stable release, let's go in this way 👍🏼 #

davidhewitt:
(I force-pushed to fix the commit message. Will merge shortly.)

PR #1348, opened 2020-12-29
ffi module cleanup: genobject.h to iterobject.h

nw0:
Partial fix for #1289. This one is slightly more complex.


davidhewitt (approved):
These are looking great, thanks very much again!

Issue #1337, opened 2020-12-23
build.rs fails to parse hand-written pyconfig.h headers installed by python for windows (MSVC-compiled)


ravenexp:
## 🐛 Bug Reports

When trying to cross-compile a python extension for windows on linux using the mingw target I get the following error:

    $ cargo build --target x86_64-pc-windows-gnu

    error: failed to run custom build command for `pyo3 v0.13.0`

    Caused by:
      process didn't exit successfully: `./target/debug/build/pyo3-7cf3ac8bb8af293b/build-script-build` (exit code: 1)
      --- stderr
      Error: "Py_ENABLE_SHARED is not defined"

pyconfig.h from python 3.9.1 for windows (built with the MSVC compiler), which I have installed, contains the following lines:

```
/* For Windows the Python core is in a DLL by default.  Test
Py_NO_ENABLE_SHARED to find out.  Also support MS_NO_COREDLL for b/w compat */
#if !defined(MS_NO_COREDLL) && !defined(Py_NO_ENABLE_SHARED)
#       define Py_ENABLE_SHARED 1 /* standard symbol for shared library */
#       define MS_COREDLL       /* deprecated old symbol */
#endif /* !MS_NO_COREDLL && ... */
```

When I manually add the line

    #define Py_ENABLE_SHARED 1

to pyconfig.h the error disappears.

I believe the following code fails to parse C headers with white space between # and define,
as well as with the comments:

https://github.com/PyO3/pyo3/blob/master/build.rs#L224

The C header parsing code should somehow be restricted to generated pyconfig.h headers only.
Manually written pyconfig.h from python for MSVC does not work.

### 🌍 Environment

Arch Linux
python 3.9.1
mingw-w64-python39-bin 3.9.1 <- this is the MSVC-compiled python, not the MinGW-compiled python
rustc 1.48.0
pyo3 0.13.0

davidhewitt:
Hi @ravenexp thanks for the report. Is there a `_sysconfigdata.py` file (might have a slightly different name) anywhere in your mingw lib directory? Perhaps we can parse this; it's usually more reliable than header parsing.

(See `find_sysconfigdata` function in `build.rs` for more detail.)

ravenexp:
No, there isn't. It could be that the ArchLinux package author has chosen not to install it, though.

BTW, what kind of configuration data pyo3 needs to build an extension module for Windows?
I'm asking because by applying the above workaround, as well as some DLL import library symlinking hacks,
[...]

PR #1350, opened 2020-12-31
Refactor build rs

davidhewitt:
This PR makes some simplifications to build.rs, based on #1337:

- Removes unicode flags
- Support "default" windows cross-compile without `PYO3_CROSS_INCLUDE_DIR` header parsing
- Some other cleanups/simplifications which seemed reasonable at the same time.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -97,9 +97,7 @@ struct CrossCompileConfig {
 impl CrossCompileConfig {
     fn both() -> Result<Self> {
         Ok(CrossCompileConfig {
-            include_dir: Some(CrossCompileConfig::validate_variable(
-                "PYO3_CROSS_INCLUDE_DIR",
-            )?),
+            include_dir: env::var_os("PYO3_CROSS_INCLUDE_DIR").map(Into::into),
             ..CrossCompileConfig::lib_only()?
         })
     }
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -195,24 +174,16 @@ fn cross_compiling() -> Result<Option<CrossCompileConfig>> {
 ///
 /// (hrm, this is sort of re-implementing what distutils does, except
 /// by passing command line args instead of referring to a python.h)
-static SYSCONFIG_FLAGS: [&str; 7] = [
-    "Py_USING_UNICODE",
-    "Py_UNICODE_WIDE",
+type BuildFlag = &'static str;
```
kngwyu, line 177: Is this *really* so meaningful?
I think maybe `struct BuildFlags(HashSet<&'static str>)` would be a better abstraction.
Then we can move `BUILD_FLAGS` into `impl BuildFlags` block.

build.rs
```diff
@@ -195,24 +174,16 @@ fn cross_compiling() -> Result<Option<CrossCompileConfig>> {
...
+
+static BUILD_FLAGS: [BuildFlag; 5] = [
     "WITH_THREAD",
     "Py_DEBUG",
     "Py_REF_DEBUG",
     "Py_TRACE_REFS",
     "COUNT_ALLOCS",
 ];
 
-static SYSCONFIG_VALUES: [&str; 1] = [
```
kngwyu, line 208: 👍🏼

kngwyu (approved):
Nice, thank you!
Also, thanks @ravenexp for researching.

davidhewitt:
Pushed a commit to use `struct BuildFlags`; will merge shortly so I can rebase #1347 on this.

PR #1352, opened 2020-12-31
ci: deny clippy warnings for examples

davidhewitt:
I noticed that `make clippy` was not denying warnings for the examples.


kngwyu (approved):
Thanks!

PR #1353, opened 2021-01-02
Fix #[pyclass] arguments description

daniil-konovalenko:
This is a minor change to the `#[pyclass]` parameters documentation.
I added quotes to reflect the changes in 0.13

I initially thought that the `name` value should be just an identifier without quotes, unlike `module`, but the compiler was quick to correct me: `error: since PyO3 0.13 a pyclass name should be in double-quotes, e.g. "Profile"`


davidhewitt (approved):
👍 thanks very much for tidying this up!

PR #1355, opened 2021-01-03
gil: tidy ups to finalization

### LINKED ISSUE #1073: Importing tensorflow causes a crash on exit
### 🌍 Environment

 - Your operating system and version: Ubuntu 18.04.4 LTS (Bionic Beaver)
 - Your python version: Python 3.7.8
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: Installed via apt. No virtualenv.
 - Your Rust version (`rustc --version`): 1.45.0
 - Your PyO3 version: 0.11.1
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: Yes

### 💥 Reproducing

Tensorflow version: 2.1.0, installed via pip: `pip3 install tensorflow==2.1.0`

Minimal reproduction:

```rust
use pyo3::prelude::*;

fn main() {
    let py = gil.python();

    let _mod = PyModule::import(py, "tensorflow").unwrap();
}
```
On exit:
```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
          Running `target/debug/pyo3`
corrupted double-linked list
[1]    4387 abort      cargo run
```
or

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
          Running `target/debug/pyo3`
corrupted size vs. prev_size
[1]    5299 abort      cargo run
```

Running import tensorflow directly in python (and exiting the interpreter) exits normally.

Looks like a similar issue to [Importing mxnet causes a segfault on exit #1044](https://github.com/PyO3/pyo3/issues/1044#issue-657897568).

If I comment out `libc::atexit(finalize)` as @kngwyu suggested in [comment](https://github.com/PyO3/pyo3/issues/1044#issuecomment-660464673), the error does not happen and the process exits successfully.

### LINKED ISSUE #1261: Importing SciPy causes segfault in PyFinalize()
### 🌍 Environment

 - Your operating system and version: macOS 10.15.5
 - Your python version: 3.8.3
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: homebrew and pyenv, no venv
 - Your Rust version (`rustc --version`): rustc 1.47.0 (18bf6b4f0 2020-10-07)
 - Your PyO3 version: v0.12.3
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: yes, same error

### 💥 Reproducing

The below code dropped in a fresh Cargo project with a pyo3 dependency is sufficient. 
```
fn main() {
    let py = gil.python();

    let _mod = PyModule::import(py, "scipy").unwrap();
}
```

```cargo run```

### Discussion
```
tentactl(1184,0x10f08cdc0) malloc: *** error for object 0x7ff13f5a5b70: pointer being freed was not allocated
tentactl(1184,0x10f08cdc0) malloc: *** set a breakpoint in malloc_error_break to debug
```
This is the error you observe when running, I am writing a chunk of code that is a Rust, C++ and Python hybrid and thought that was the issue, after doing some debugging I found it is crashing during PyFinalize and is free-ing memory that has likely already been freed. 

My hunch is this is a deinitilization ordering issue, if extensions are not careful they can leave memory around that is no longer owned by Python causing faults like this (have encountered same problem while developing Apache TVM).

Similar to https://github.com/PyO3/pyo3/issues/1073. I traced down what is happening GDB.

Removing all SciPy imports from the code temporarily fixed this for me, I do not have the expertise to dig into SciPy and figure out what is going on.

### PR DESCRIPTION
**Pull request by davidhewitt**
EDIT: I changed strategy. See https://github.com/PyO3/pyo3/pull/1355#issuecomment-756446786 for the latest design.

This is a follow-up to #1347 which adds some tidy-ups to our finalization behavior:

- Calling `Py_Finalize` in a different thread to `Py_Initialize` will cause an `AssertionError`. (See https://bugs.python.org/issue26693)

  ~~So I changed `prepare_freethreaded_python` to initialize and finalize in a dedicated thread.~~

- We have some C-extensions which have crashes with `Py_Finalize`. ~~So after disabling the `auto-initialize` feature we can ask users with problematic extensions to call the new `prepare_freethreaded_python_without_finalizer` to workaround these. I added this to the FAQ.~~

Closes #1073 
Closes #1261

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -42,13 +42,6 @@ pub(crate) fn gil_is_acquired() -> bool {
 /// Python signal handling depends on the notion of a 'main thread', which must be
 /// the thread that initializes the Python interpreter.
 ///
-/// Additionally, this function will register an `atexit` callback to finalize the Python
-/// interpreter. Usually this is desirable - it flushes Python buffers and ensures that all
-/// Python objects are cleaned up appropriately. Some C extensions can have memory issues during
-/// finalization, so you may get crashes on program exit if your embedded Python program uses
-/// these extensions. If this is the case, you should use [`prepare_freethreaded_python_without_finalizer`]
-/// which does not register the `atexit` callback.
-///
 /// If both the Python interpreter and Python threading are already initialized,
 /// this function has no effect.
 ///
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by davidhewitt** [comment]
Hmm, the coverage job seems to have some kind of deadlock...

**Comment by davidhewitt** [comment]
@kngwyu this PR is now ready for review - would be interested to know what you think of it.

**Comment by kngwyu** [comment]
> So I changed prepare_freethreaded_python to initialize and finalize in a dedicated thread.

If we really need this trick I don't want to use `libc::atexit` to call finalize. Is there any way to call finalize from the main thread?

**Comment by kngwyu** [comment]
I think it's also good to provide `#[pyo3::main] fn main() { .. }`, but am not sure about which approach is better.

**Comment by davidhewitt** [comment]
> If we really need this trick I don't want to use libc::atexit to call finalize. Is there any way to call finalize from the main thread?

The problem is in tests there is no main thread, so we will always need some kind of trick like this for tests.

`#[pyo3::main(finalize = false)]` is a really nice idea; I guess we will build it on top of this API. I'll push a commit to experiment with that later.

**Comment by kngwyu** [comment]
Then how about memorizing the thread id and calling finalize if the current thread == initialized thread?

Also, I think we need to rethink the use cases of `Py_Finalize`.

The [document](https://docs.python.org/3/c-api/init.html#c.Py_FinalizeEx) notes some examples:

> This function is provided for a number of reasons. An embedding application might want to restart Python without having to restart the application itself. An application that has loaded the Python interpreter from a dynamically loadable library (or DLL) might want to free all memory allocated by Python before unloading the DLL. During a hunt for memory leaks in an application a developer might want to free all memory allocated by Python before exiting from the application.

Our original motivation (clearing the buffer) does not match any of these, and I'm kind of doubtful about the idea that it must be called.

**Comment by davidhewitt** [comment]
I think I don't want to do anything like memorize the thread id, as then it's unpredictable in tests whether finalize is called, which doesn't seem good for reproducability. Also I'm now thinking that because we know lots of crashes can occur at finalization it's better not to have it as a default.

I pushed a new commit which removes `atexit` completely from `prepare_freethreaded_python` and adds a new api `with_embedded_python_interpreter` which does both initialization and finalization guaranteed to be on the current thread.

We could implement `#[pyo3::main]` and `#[pyo3::main(finalize = true)]` (opt-in finalization) on top of these APIs trivially as a future PR.

File: src/gil.rs [hunk 1]
```diff
@@ -46,66 +46,125 @@ pub(crate) fn gil_is_acquired() -> bool {
...
+#[cfg(all(Py_SHARED, not(PyPy)))]
+pub unsafe fn with_embedded_python_interpreter<F, R>(f: F) -> R
+where
+    F: for<'p> FnOnce(Python<'p>) -> R,
+{
+    assert_eq!(
+        ffi::Py_IsInitialized(),
+        0,
+        "called `with_embedded_python_interpreter` but a Python interpreter is already running."
+    );
```
kngwyu, line 144: If we force some error type (e.g., `PyErr`), we can return an `Err` here. It's an optional choice, though.

src/gil.rs
```diff
@@ -46,66 +46,125 @@ pub(crate) fn gil_is_acquired() -> bool {
...
+
+    ffi::Py_InitializeEx(0);
+
+    #[cfg(not(Py_3_7))] // Called by Py_InitializeEx in Python 3.7 and up.
+    ffi::PyEval_InitThreads();
+
+    // Import the threading module - this ensures that it will associate this thread as the "main"
+    // thread, which is important to avoid an `AssertionError` at finalization.
+    let pool = GILPool::new();
+    pool.python().import("threading").unwrap();
```
kngwyu, line 154: Why not acquring a GIL?

kngwyu (commented):
Thanks!
This is a breaking change, but I believe it would be applicable.

src/gil.rs
davidhewitt: I think `PyErr` doesn't work here because then the user will have a `PyErr` outside of the scope of a Python interpreter!

The hazards of `unsafe` 😅

src/gil.rs
davidhewitt: It's already aquired by `Py_Initialize`. I'll add a comment.

src/gil.rs
```diff
@@ -46,66 +46,125 @@ pub(crate) fn gil_is_acquired() -> bool {
...
-            libc::atexit(finalize);
-
-            // > Changed in version 3.7: This function is now called by Py_Initialize(), so you don’t have
-            // > to call it yourself anymore.
-            #[cfg(not(Py_3_7))]
-            if ffi::PyEval_ThreadsInitialized() == 0 {
-                ffi::PyEval_InitThreads();
-            }
+            #[cfg(not(Py_3_7))] // Called by Py_InitializeEx in Python 3.7 and up.
+            ffi::PyEval_InitThreads();
```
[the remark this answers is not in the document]
davidhewitt: 👍 I think you're right, I'll put it back.

davidhewitt:
@kngwyu this PR is ready for another look. I'm quite happy with it now, would be great to hear if you think there should be any further changes.

kngwyu (approved):
Thanks, LGTM 💯

PR #1356, opened 2021-01-03
readme/faq: document `rlib` crate type better

### LINKED ISSUE #1321: Integration tests can't find crate

davidhewitt:
This simplifies the documentation around `cdylib` and `rlib` crate types in the readme and includes a similar note in the FAQ as #1336 .

I think that `rlib` crate type is the best solution - I tested manually and this works great for me locally.

I think the issue described in https://github.com/PyO3/pyo3/pull/1336#discussion_r547271216 is likely a result of some static linking, because the error is talking about relocating a symbol incorrectly. This is probably prevented by #1347.

Closes #1321
Closes #1336

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,11 @@ edition = "2018"
 
 [lib]
 name = "string_sum"
+# "cdylib" is necessary to produce a shared library for Python to import from.
+#
+# Downstream Rust code (including code in `bin/`, `examples/`, and `tests/`) will not be able
+# to `use string_sum;` unless the "rlib" or "lib" crate type is also included, e.g.:
+# crate-type = ["cdylib", "rlib"]
 crate-type = ["cdylib"]
 
 [dependencies.pyo3]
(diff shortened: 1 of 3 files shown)


kngwyu:
Thanks.
Could you please rebuild this PR on #1336 to give credit to the original author?
E.g., [gh pr pull](https://cli.github.com/) command would work nicely.

davidhewitt:
👍 will do so when I get a chance

Issue #1221, opened 2020-10-10
Cross-compiling wasm32-unknown-unknown on 64bit debian


jruiss:
## 🐛 Bug Reports

When cross-compiling pyo3s build.rs fails on two places without any pointers on what lines, but rather just "Error: NotPresent", 

The guilty lines in build.rs:
https://github.com/PyO3/pyo3/blob/master/build.rs#L157, env::var("CARGO_CFG_TARGET_FAMILY") is not set
https://github.com/PyO3/pyo3/blob/master/build.rs#L433, env::var("CARGO_CFG_TARGET_FAMILY") is not set

After that, a custom python 32bit compilation requires PYO3_CROSS_INCLUDE_DIR which is only set for windows in fn cross_compilings https://github.com/PyO3/pyo3/blob/master/build.rs#L150

And finally, Include/pyconfig.h doesn't exists in https://www.python.org/downloads/release/python-379/
Which is required at https://github.com/PyO3/pyo3/blob/master/build.rs#L414

pyconfig.h gets generated when using ./configure and ends up in the root dir.

Also, when python is configured without --enable-shared and such, options retrieved from config_map crashes instead of defaulting to false, dunno if there's a reason for that. https://github.com/PyO3/pyo3/blob/master/build.rs#L415 but unwrap_or(false) might be better than current solution?

Accidently posted this issue before successfully compiling, will update on progress

**update:**
Well, seem to have gotten rid of all errors in build.rs, it seems to find all neccessary files and and so on but now I get about 100 errors regarding libc.

Using wasm32-wasi instead pulls the errors down to only 3 regarding wchar_t in libc. Seems to be a problem with libc and wasm32 target from here on out. 
```
cargo build --release --target wasm32-wasi
pyo3 git:(master) ✗ cargo build --release --target wasm32-wasi           
   ...
   Compiling libc v0.2.79
   Compiling pyo3 v0.12.1 (/home/cfr/workspace/clones/pyo3)
   ...
error[E0432]: unresolved import `libc::wchar_t`
 --> src/ffi/unicodeobject.rs:3:5
  |
3 | use libc::wchar_t;
  |     ^^^^^^^^^^^^^ no `wchar_t` in the root

error[E0432]: unresolved import `libc::wchar_t`
 --> src/ffi/pylifecycle.rs:2:5
  |
2 | use libc::wchar_t;
  |     ^^^^^^^^^^^^^ no `wchar_t` in the root

[...]

PR #1359, opened 2021-01-04
Let pyconfig.h not define Py_ENABLE_SHARED

dalcde:
If there is no line that defines Py_ENABLE_SHARED in `pyconfig.h`, then it is false

This fixes an issue described in #1221

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -488,7 +488,7 @@ fn load_cross_compile_from_headers(
     let interpreter_config = InterpreterConfig {
         version: python_version,
         libdir: cross_compile_config.lib_dir.to_str().map(String::from),
-        shared: config_data.get_bool("Py_ENABLE_SHARED")?,
+        shared: config_data.get_bool("Py_ENABLE_SHARED").unwrap_or(false),
         ld_version: format!("{}.{}", major, minor),
         base_prefix: "".to_string(),
         executable: PathBuf::new(),


davidhewitt:
Hi @dalcde thanks very much for the PR.

I'm slightly reluctant to merge this, because the header parsing is slightly broken for a few various reasons and I'm thinking about removing it anyway in PyO3 0.14.

Can you tell me a bit more about your use case? Perhaps we can come up with a better way to configure PyO3 in your situation.

dalcde:
On Mon, Jan 04, 2021 at 12:24:04PM -0800, David Hewitt wrote:
>Can you tell me a bit more about your use case? Perhaps we can come up with a better way to configure PyO3 in your situation.

I'm looking into building pyo3 packages to wasm32-unknown-unknown for use with https://github.com/iodide-project/pyodide .

If I set CARGO_CFG_TARGET_FAMILY=unix I get

> Your Rust target architecture (32-bit) does not match your python interpreter (64-bit)

and if I set it to windows it doesn't complain about this but it uses the header file instead.

I'm most likely not going about this the right way, but on the other hand what this PR does is more correct behaviour.

kngwyu (approved):
Thanks!

kngwyu:
> I'm slightly reluctant to merge this, because the header parsing is slightly broken for a few various reasons and I'm thinking about removing it anyway in PyO3 0.14. 

I don't think it's not very straightforward to remove the header parsing and it'll take time. So I don't think merging this now is harmful.

davidhewitt:
Ok 👍

PR #1360, opened 2021-01-05
Update mdbook version

kngwyu:
< 0.4.5 has a security issue.
See https://blog.rust-lang.org/2021/01/04/mdbook-security-advisory.html.


davidhewitt:
Good catch, thanks!

PR #1361, opened 2021-01-05
Don't check target architecture if cross-compiling

dalcde:
Part of the point of cross-compiling is to be able to compile to other architectures.

This is needed in order to compile to wasm32-unknown-unknown: see #1221

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -733,7 +733,9 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<()> {
         );
     }
 
-    check_target_architecture(interpreter_config)?;
+    if cross_compiling()?.is_none() {
+        check_target_architecture(interpreter_config)?;
+    }
     let target_os = env::var_os("CARGO_CFG_TARGET_OS").unwrap();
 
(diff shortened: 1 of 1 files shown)


davidhewitt:
I'm wary of this... The purpose of `check_target_architecture` is to give a friendly error message at build time if the Rust target doesn't match what PyO3 has detected from the config.

If it's wrong for you, this could mean that the library you're getting build configuration from thinks the pointer size is 64-bit?

If your build configuration really is correct, I think rather than disable this check for all cross compiles I'd rather have an environment variable `PYO3_ALLOW_INCORRECT_ARCHITECTURE` (or a better name if you can think of one). So that in cases like this, users who know what they're doing can proceed anyway.

davidhewitt:
... maybe `PYO3_SKIP_TARGET_CHECK` is a better name for the environment variable.

dalcde:
Ah I think I misunderstood the purpose of this code. It correctly identified that I previously had a broken _sysconfigdata_*.py and threw an error. The problem is now fixed in upstream pyodide.

PR #1362, opened 2021-01-05
Enable wasm32 compilation by wrapping libc

dalcde:
My plan is to use this with [pyodide](https://github.com/iodide-project/pyodide).

It is unclear how to test this at the moment. I think a reasonable solution is to make sure it compiles with `--target wasm32-unknown-unknown` after supplying suitable header files, possibly copying from pyodide, and run no tests. When/if things get ready with pyodide, we can consider running integration tests with it, but that would require a very different setup. Due to the nature of the builds, I also think it is unlikely that it can be usable without nightly in the near future.

I'm also happy to leave this as a PR until then. See also #1221

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -357,7 +357,7 @@ pub(crate) fn impl_wrap_setter(
         #[allow(unused_mut)]
         unsafe extern "C" fn __wrap(
             _slf: *mut pyo3::ffi::PyObject,
-            _value: *mut pyo3::ffi::PyObject, _: *mut ::std::os::raw::c_void) -> ::std::os::raw:c_int
+            _value: *mut pyo3::ffi::PyObject, _: *mut ::std::os::raw::c_void) -> ::std::os::raw::c_int
         {
             const _LOCATION: &'static str = concat!(stringify!(#cls),".",stringify!(#python_name),"()");
             pyo3::callback_body_without_convert!(_py, {
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -357,7 +357,7 @@ pub(crate) fn impl_wrap_setter(
         #[allow(unused_mut)]
         unsafe extern "C" fn __wrap(
             _slf: *mut pyo3::ffi::PyObject,
-            _value: *mut pyo3::ffi::PyObject, _: *mut ::std::os::raw::c_void) -> pyo3::libc::c_int
+            _value: *mut pyo3::ffi::PyObject, _: *mut ::std::os::raw::c_void) -> ::std::os::raw:c_int
         {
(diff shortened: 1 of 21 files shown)


davidhewitt:
... so as I understand it, `libc` doesn't provide any exports on `wasm32`? Should this be fixed upstream?

dalcde:
libc made a conscious decision to not support wasm32-unknown-unknown,
which makes sense since wasm doesn't come with any libc-like runtime.
For the purposes of building python extensions, this is less of a
problem, since we are building a dynamic library, to be loaded by an
emscripten runtime, which does provide libc-like functionalities.
An in any case, apart from the type definitions, all we need is
libc::atexit, which emscripten does happen to provide (should we provide
failsafes if it isn't present?)

There *is* a different target, wasm32-unknown-wasi, and wasi is a
standard for provide posix-like system calls in webassembly. libc is
working on an implementation for wasm32-unknown-wasi.

This wrapper is only enabled withen architecture = wasm32 and os =
unknown, so wasm32-unknown-wasi should use the "real" wasi-implemented
libc.

davidhewitt:
I see, thanks. In which case to have wasm32 support we're definitely going to need something like this.

To keep the "wrapper" as small as possible, for all types we use which are exported in `std::os::raw` can you please change this PR to use `std::os::raw` in the usage site and remove them from the wrapper?

Let's see what's left after they're gone - I think except for `size_t`, they probably account for most of the usages.

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -357,7 +357,7 @@ pub(crate) fn impl_wrap_setter(
         #[allow(unused_mut)]
         unsafe extern "C" fn __wrap(
             _slf: *mut pyo3::ffi::PyObject,
-            _value: *mut pyo3::ffi::PyObject, _: *mut ::std::os::raw::c_void) -> pyo3::libc::c_int
+            _value: *mut pyo3::ffi::PyObject, _: *mut ::std::os::raw::c_void) -> ::std::os::raw::c_int
```
kngwyu, line 360: BTW, #1367 may conflict with this line (I'm sorry!) and you will need to merge that manually.

kngwyu:
I think we can remove `atexit` but are still discussing at #1355

dalcde:
I think before resolving any rustc/wasm32 shenanigans, we can merge the
changes that convert `libc` to `std::os::raw` when applicable, which I
think is an positive change in any case (currently both are used,
seemingly arbitrarily)

davidhewitt:
> we can merge the
changes that convert `libc` to `std::os::raw` when applicable,

Agreed - are you willing to factor those changes out into a separate PR?

dalcde:
>Agreed - are you willing to factor those changes out into a separate PR?

Sure

dalcde:
Closing in favor of #1368. May be reopened if this turns out to be useful/needed

PR #1365, opened 2021-01-05
pyclass: fix reference count issue in subclass new

### LINKED ISSUE #1363: Python subclassing doesn't inherit from pyclass on py3.6 or 3.7
## 🐛 Bug Reports

In retworkx I'm trying to update to use the latest pyo3 version in https://github.com/Qiskit/retworkx/pull/224 ci jobs on python 3.6 and 3.7: https://github.com/Qiskit/retworkx/pull/224/checks?check_run_id=1650749331#step:6:16

Several of the tests use a python subclass of a rust class (it's just a backwards compat alias so there are no changes besides the name it's defined here: https://github.com/Qiskit/retworkx/blob/master/retworkx/__init__.py#L15-L83 ) and those tests are all failing on python 3.6 and 3.7 (both locally and in CI) with a `TypeError` on method calls (the python subclass type isn't valid for the rust method) or on the constructor (saying no constructor is defined for the subclas). The issue looks like the changes in subclassing for 0.13.0 don't work with python 3.6 or 3.7 as the 3.8 and 3.9 jobs work fine. 

### 🌍 Environment

 - Your operating system and version: linux x86_64
 - Your python version: 3.6.12 and 3.7.9
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: distro package with virtualenv
 - Your Rust version (`rustc --version`): 1.49.0
 - Your PyO3 version: 0.13.0
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: no

### 💥 Reproducing

I tried to make a minimum reproduce case with:

```rust
#[pyclass(module = "mymodule", subclass)]
pub struct MyClass {
    pub value: bool,
}

#[pymethods]
impl MyClass {
    #[new]
    fn new() -> Self {
        MyClass {
            value: false,
        }
    }
    
    fn get_value(&self) -> bool {
        self.value
    }
}
```

```python3
import mymodule
class MySubClass(mymodule.MyClass):
    pass
    
obj = MySubClass()
print(obj.get_value())
```

but this worked fine locally (both in lib.rs and importing the class definition from another rust module). So I'm not sure what I'm doing differently in retworkx that could be causing this. But the same exact code worked fine with 0.12.x and the upgrade notes only called out using the `subclass` argument on the pyclass macro which I already was using.

davidhewitt:
This fixes a missing reference count increase (to work around a cpython bug) on Python versions 3.7 and older - see https://github.com/PyO3/pyo3/issues/1363#issuecomment-754926607

It's a fairly nasty bug which leads to pretty immediate weird behaviour / crashes when tripped, so I think it is worth releasing 0.13.1 once this is merged.

Closes #1363

### DIFF
--- a/src/freelist.rs
+++ b/src/freelist.rs
@@ -76,6 +76,8 @@ where
         if subtype == Self::type_object_raw(py) {
             if let Some(obj) = <Self as PyClassWithFreeList>::get_free_list(py).pop() {
                 ffi::PyObject_Init(obj, subtype);
+                #[cfg(not(Py_3_8))]
+                crate::pyclass::bpo_35810_workaround(py, subtype);
                 return obj as _;
             }
         }
@@ -87,13 +89,13 @@ where
         let obj = PyAny::from_borrowed_ptr_or_panic(py, self_ as _);
 
         if let Some(obj) = <Self as PyClassWithFreeList>::get_free_list(py).insert(obj.as_ptr()) {
-            match get_type_free(ffi::Py_TYPE(obj)) {
-                Some(free) => {
-                    let ty = ffi::Py_TYPE(obj);
-                    free(obj as *mut c_void);
-                    ffi::Py_DECREF(ty as *mut ffi::PyObject);
-                }
-                None => tp_free_fallback(obj),
+            let ty = ffi::Py_TYPE(obj);
+            let free = get_type_free(ty).unwrap_or_else(|| tp_free_fallback(ty));
+            free(obj as *mut c_void);
+
+            #[cfg(Py_3_8)]
+            if ffi::PyType_HasFeature(ty, ffi::Py_TPFLAGS_HEAPTYPE) != 0 {
+                ffi::Py_DECREF(ty as *mut ffi::PyObject);
             }
         }
     }
--- a/src/pyclass.rs
+++ b/src/pyclass.rs
@@ -1,11 +1,9 @@
 //! `PyClass` and related traits.
 use crate::class::methods::{PyClassAttributeDef, PyMethodDefType, PyMethods};
 use crate::class::proto_methods::PyProtoMethods;
-use crate::conversion::{AsPyPointer, FromPyPointer};
 use crate::derive_utils::PyBaseTypeUtils;
 use crate::pyclass_slots::{PyClassDict, PyClassWeakRef};
 use crate::type_object::{type_flags, PyLayout};
-use crate::types::PyAny;
 use crate::{ffi, PyCell, PyErr, PyNativeType, PyResult, PyTypeInfo, Python};
 use std::convert::TryInto;
 use std::ffi::CString;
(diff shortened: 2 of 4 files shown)


davidhewitt:
Hmm the tests ran fine on my local Python 3.9, I'll take another look tomorrow...

src/freelist.rs
```diff
@@ -76,6 +76,11 @@ where
         if subtype == Self::type_object_raw(py) {
             if let Some(obj) = <Self as PyClassWithFreeList>::get_free_list(py).pop() {
                 ffi::PyObject_Init(obj, subtype);
+                // Workaround for Python issue 35810; no longer necessary in Python 3.8
+                #[cfg(not(Py_3_8))]
+                if ffi::PyType_HasFeature(subtype, ffi::Py_TPFLAGS_HEAPTYPE) != 0 {
```
kngwyu, line 81: Is `subtype` always heap type?

src/pyclass.rs
```diff
@@ -64,15 +71,14 @@ pub trait PyClassAlloc: PyTypeInfo + Sized {
...
-                let ty = ffi::Py_TYPE(obj.as_ptr());
-                free(obj.as_ptr() as *mut c_void);
-                ffi::Py_DECREF(ty as *mut ffi::PyObject);
-            }
-            None => tp_free_fallback(obj.as_ptr()),
+        let ty = ffi::Py_TYPE(obj);
+        let free = get_type_free(ty).unwrap_or_else(|| tp_free_fallback(ty));
+        free(obj as *mut c_void);
+
+        if ffi::PyType_HasFeature(ty, ffi::Py_TPFLAGS_HEAPTYPE) != 0 {
```
kngwyu, line 80: `#[cfg(Py_3_8)]` is missing.

kngwyu (changes requested):
Nice catch, thanks!

src/freelist.rs
davidhewitt: 👍 I think it must be, because all our types are now heap types. Similarly for dealloc.

src/pyclass.rs
davidhewitt: Good catch!

davidhewitt:
@kngwyu I think this is ready for another look!

src/pyclass.rs
```diff
@@ -44,8 +42,24 @@ pub(crate) unsafe fn default_new<T: PyTypeInfo>(
...
+    let ptr = alloc(subtype, 0);
+
+    // Workaround for Python issue 35810; no longer necessary in Python 3.8
+    #[cfg(not(Py_3_8))]
+    {
+        use crate::once_cell::GILOnceCell;
+
+        // Must check version at runtime because of abi3 wheels which could run against a higher
+        // version than the config.
+        static IS_NOT_PYTHON_3_8: GILOnceCell<bool> = GILOnceCell::new();
```
kngwyu, line 56: How about checking the current refcount instead of the version?

davidhewitt: I couldn't find a way to detect this case using the type reference count - it increases by one on all versions I tested.

I guess I need to make this check only with `Py_Limited_API`, so I could avoid checking the version in more cases.

davidhewitt:
I pushed a new version of this PR which only does the runtime version check if `Py_LIMITED_API` is set.

Let me know if there's any other changes you would like to this. Once you're happy, I'll merge this and release 0.13.1 after.

kngwyu (approved):

kngwyu:
LGTM, thanks!

Issue #1358, opened 2021-01-04
pyo3 ClassB referencing ClassA instead of Py<ClassA> disables ClassA pyo3(set) pymethod


aagit:
## 🐛 Bug Reports

Hello,

During the holidays I spent some spare time experimenting with pyo3. I found something didn't work as I would have expected, fom a non-rust programmer point of view, this looks really strange. Since I run out of holiday spare time, I setup  a reproducer and I send this bugreport to notify about this. Overall I'm curious to hear why it is happening, if it's a fixable bug, if it's a unfixable bug or if it's just a bug in my test code.

What feels like is that python side or the rust side are silently losing writes. At the very least I'd expect to see some runtime error from either side and not just silent data loss.

Taking a look at similar cases like in tests/test_class_conversion.rs I see Py<BaseClass> is used, that's how I got the classB2 setup and then it works, but it's not clear if it's just working by luck in such case or if Py is used precisely for that purpose of activating all BaseClass methods like pyo3(set).

### 🌍 Environment

 - Your python version: 3.8.6
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: system python on Gentoo
 - Your Rust version (`rustc --version`): 1.47.0
 - Your PyO3 version: 0.13.0
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: yes

### 💥 Reproducing

$ git clone https://github.com/aagit/pyo3-reproducer.git
$ cd pyo3-reproducer
$ firejail --private=`pwd` --seccomp # optional
$ cargo build
$ ./test.py
case1 B 0
0 False
case2 A 0
1 True
case3 B 0
0 False
case4 B 0
1 True
1 False
case1 B2 0
1 True
case3 B2 0
1 True
case4 B2 0
1 True
2 True
case4 B workaround 0
1 True

All lines with False show the unexpected silent write lost.

In all False cases the pyo3(set) for member "x" of ClassA is silently not invoked when python sets b.a.x .

ClassB2 shows how Py<ClassA> stabilizes the python object lifetime and doing so also happen to re-activate pyo3(set) method of ClassA and then everything works as expected.

It's just not clear if Py<ClassaA> should be necessary to make it work and why "set" doesn't trigger an error if cannot hit the memory of the real ClassA rust side.

[...]

PR #1366, opened 2021-01-06
implement Serialize, Deserialize for Py<T>

daniil-konovalenko:
David Hewitt wrote in #1358
> As for Serialize / Deserialize for `Py<T>` - I'm beginning to think that PyO3 should have an optional serde feature to support this...

I thought this too and hacked together a `"serde"` feature that includes implementations of Serialize and Deserialize for `Py<T>`

TODO:
 - [x] tidy up tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ PyO3 versions, please see the [migration guide](https://pyo3.rs/master/migration
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- Add `serde` feature to support `Serialize/Deserialize` for `Py<T>`. [#1366](https://github.com/PyO3/pyo3/pull/1366)
+
 ## [0.13.1] - 2021-01-10
 ### Added
 - Add support for `#[pyclass(dict)]` and `#[pyclass(weakref)]` with the `abi3` feature on Python 3.9 and up. [#1342](https://github.com/PyO3/pyo3/pull/1342)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ paste = { version = "1.0.3", optional = true }
 pyo3-macros = { path = "pyo3-macros", version = "=0.13.1", optional = true }
 unindent = { version = "0.1.4", optional = true }
 hashbrown = { version = "0.9", optional = true }
+serde = {version = "1.0", optional = true}
 
 [dev-dependencies]
 assert_approx_eq = "1.1.0"
@@ -35,6 +36,7 @@ rustversion = "1.0"
 proptest = { version = "0.10.1", default-features = false, features = ["std"] }
 # features needed to run the PyO3 test suite
 pyo3 = { path = ".", default-features = false, features = ["macros", "auto-initialize"] }
+serde_json = "1.0.61"
 
 [features]
 default = ["macros", "auto-initialize"]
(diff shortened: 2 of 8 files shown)


daniil-konovalenko:
@davidhewitt Hi, could you please take a look at this? The PR is by no means ready for the real review, but I really would like your feedback on the general implementation idea.

The thing that bothers me the most is error handling — should we just convert pyo3 errors to string and create custom serde errors from it, or maybe we should attach more context to them?

kngwyu:
Thanks!
> The PR is by no means ready for the real review

I don't see any clear problem... what do you have in your mind as a blocker?

> The thing that bothers me the most is error handling

I think both `se` and `de` would rarely fail and we don't have to take care of it so seriously.

daniil-konovalenko:
Thank you for quick response!


> I don't see any clear problem... what do you have in your mind as a blocker?
I was anxious about error handling (I'm new to Rust), and the tests could be better, but the implementation itself looks ready to review.

Also I haven't written the docs yet, but this doesn't block the review of course

guide/src/features.md
```diff
@@ -28,6 +28,27 @@ These features are an extension of the `abi3` feature to specify the exact minim
 
 See the [building and distribution](building_and_distribution.md#minimum-python-version-for-abi3) section for further detail.
 
+### `serde`
```
davidhewitt, line 31: This needs to go in the "Advanced Features" section (which should maybe be renamed to "Other Features") as this feature could be useful in the embedding case too.

Cargo.toml
```diff
@@ -27,14 +27,16 @@ paste = { version = "1.0.3", optional = true }
...
+serde = {version = "1.0", optional = true}
 
 [dev-dependencies]
 assert_approx_eq = "1.1.0"
 trybuild = "1.0.23"
 rustversion = "1.0"
 proptest = { version = "0.10.1", default-features = false, features = ["std"] }
 # features needed to run the PyO3 test suite
-pyo3 = { path = ".", default-features = false, features = ["macros", "auto-initialize"] }
+pyo3 = { path = ".", default-features = false, features = ["macros", "auto-initialize", "serde"] }
```
davidhewitt, line 38: Strictly speaking, we don't _need_ this feature to run the tests - they straight up don't work without the other features enabled.

I'd rather just see the `serde` feature added to the list of features tested in `ci.yml`.

davidhewitt (commented):
Thanks very much for this! 

TBH I agree, the code, tests and docs all look very good, so I'm happy! I just have some very minor nits.

Cargo.toml
daniil-konovalenko: Thanks for the tip! Added serde to ci.yml

guide/src/features.md
daniil-konovalenko: fixed, thanks!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add FFI definitions `PyOS_BeforeFork`, `PyOS_AfterFork_Parent`, `PyOS_AfterFork_Child` for Python 3.7 and up. [#1348](https://github.com/PyO3/pyo3/pull/1348)
 - Add `auto-initialize` feature to control whether PyO3 should automatically initialize an embedded Python interpreter. For compatibility this feature is enabled by default in PyO3 0.13.1, but is planned to become opt-in from PyO3 0.14.0. [#1347](https://github.com/PyO3/pyo3/pull/1347)
 - Add support for cross-compiling to Windows without needing `PYO3_CROSS_INCLUDE_DIR`. [#1350](https://github.com/PyO3/pyo3/pull/1350)
+- Add `serde` feature to support `Serialize/Deserialize` for `Py<T>`
```
davidhewitt, line 14: Just noticed this needs the PR number:

```suggestion
- Add `serde` feature to support `Serialize/Deserialize` for `Py<T>`. [#1366](https://github.com/PyO3/pyo3/pull/1366)
```

davidhewitt (approved):
Thanks, LGTM!

@kngwyu unless you think otherwise, I'm going to merge this after we release 0.13.1, to allow some time to get feedback on this on `master` before this makes a release.

CHANGELOG.md
daniil-konovalenko: Thanks!

davidhewitt:
I force-pushed a squash and rebase; will merge assuming CI is happy. Thanks again for the contribution!

PR #1368, opened 2021-01-07
Prefer to use std::os::raw type definitions

dalcde:
Certain types e.g. c_char are provided by both std::os::raw and libc.  Previously, both are used interchangably. This commit changes all instances to use std::os::raw, which is slightly more portable, since libc is not implemented for all targets (e.g. wasm32-unknown-unknown).


kngwyu (approved):
Thanks!

PR #1370, opened 2021-01-08
Deprecate functions from PEP 623

konstin:
I think those are all the remaining functions scheduled for removal by PEP 623

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Deprecate FFI definition `PyOS_AfterFork` for Python 3.7 and up. [#1348](https://github.com/PyO3/pyo3/pull/1348)
 - Deprecate FFI definition `PyCoro_Check` and `PyAsyncGen_Check` in favor of `PyCoro_CheckExact` and `PyAsyncGen_CheckExact` respectively to mirror Python API. [#1348](https://github.com/PyO3/pyo3/pull/1348)
 - Deprecate FFI definition `PyCoroWrapper_Check` which has never been in the Python API. [#1348](https://github.com/PyO3/pyo3/pull/1348)
+- Deprecate FFI definitions for `PyUnicode_FromUnicode`, `PyUnicode_AsUnicode` and `PyUnicode_AsUnicodeAndSize`, which will be removed from 3.12 and up due to [PEP 613](https://www.python.org/dev/peps/pep-0623/). [#1370](https://github.com/PyO3/pyo3/pull/1370)
 
 ### Removed
 - Remove FFI definition `PyFrame_ClearFreeList` when building for Python 3.9. [#1341](https://github.com/PyO3/pyo3/pull/1341)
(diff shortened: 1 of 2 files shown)


davidhewitt:
I think we should maybe use `#[deprecated]` to match Python? Just in case some user wants to use these apis.

kngwyu:
I think `#[cfg_attr(Py_3_9, deprecated)]` or just `#[cfg(Py_3_9)]` is sufficient.

konstin:
I've deprecated them unconditionally and cfg'ed them out with a (not yet existing) `Py_3_12`, as they'd only cause linker errors on those platforms anyway.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Remove FFI definition `PyFrame_ClearFreeList` when building for Python 3.9. [#1341](https://github.com/PyO3/pyo3/pull/1341)
 - Remove FFI definition `_PyDict_Contains` when building for Python 3.10. [#1341](https://github.com/PyO3/pyo3/pull/1341)
 - Remove FFI definitions `PyGen_NeedsFinalizing`, `PyImport_Cleanup` (for 3.9 and up), and `PyOS_InitInterrupts` (3.10). [#1348](https://github.com/PyO3/pyo3/pull/1348)
+- Deprecate FFI definitions for `PyUnicode_FromUnicode`, `PyUnicode_AsUnicode` and `PyUnicode_AsUnicodeAndSize`, which will be removed from 3.12 and up due to [PEP 613](https://www.python.org/dev/peps/pep-0623/).
```
davidhewitt, line 27: Needs the PR number:

```suggestion
- Deprecate FFI definitions for `PyUnicode_FromUnicode`, `PyUnicode_AsUnicode` and `PyUnicode_AsUnicodeAndSize`, which will be removed from 3.12 and up due to [PEP 613](https://www.python.org/dev/peps/pep-0623/). [#1370](https://github.com/PyO3/pyo3/pull/1370)
```

davidhewitt (approved):
In the other deprecations in this patch release we're using `#[cfg_attr(Py_3_9, deprecated)]`.

I _think_ I'm ok with just straight-up `#[deprecated]` on these APIs, because they're intended for removal with no replacement. Many of the deprecated methods only have their replacement added on the same Python version.

CHANGELOG.md
davidhewitt: (Also if #1372 is merged, this should go in the new `Deprecated` section.)

konstin:
> I _think_ I'm ok with just straight-up `#[deprecated]` on these APIs, because they're intended for removal with no replacement. Many of the deprecated methods only have their replacement added on the same Python version.

Yep, as I understood [https://docs.python.org/3/c-api/unicode.html](https://docs.python.org/3/c-api/unicode.html) and [PEP 623](https://www.python.org/dev/peps/pep-0623/#id9) those methods were legacy for some time already:

> PEP 393 was implemented in Python 3.3 which is released in 2012. Old APIs were deprecated since then, and the removal was scheduled in Python 4.0.
>
> Python 4.0 was expected as next version of Python 3.9 when PEP 393 was accepted. But the next version of Python 3.9 is Python 3.10, not 4.0. This is why this PEP schedule the removal plan again.

CHANGELOG.md
konstin: I'll change that once #1372 this is merged

davidhewitt:
To get the release moving I'm going to force-push a fix for the commit and merge this. Thanks very much!

PR #1371, opened 2021-01-08
Make abi3 linking work on windows

konstin:
On mac and linux, you can build abi3 wheels without the python interpreter being present. On windows however, we still need to tell the linker how to find a .lib matching the `extern` blocks in pyo3. First, pyo3 tells the linker that we're looking for definitions in a collection we call `pythonXY` with `#[cfg_attr(windows, link(name = "pythonXY"))]`. Then we use `rustc-link-lib` (or `-l`) to tell the linker that the `pythonXY` is actually a file called `python3.lib`. This part needs to be done by pyo3 itself, as [cargo rejects doing this as a rustc option](https://github.com/PyO3/maturin/runs/1670164610?check_suite_focus=true#step:9:409). This pull request adds this part. Finally, maturin needs to tell the linker the directory containing `python3.lib` (with `-L`).

The advantage is that the pyo3 build script doesn't need to call python anymore and becomes independent of most environment variables.

I've also tried to use `/FORCE:UNDEFINED` to avoid the library completely but then python crashes with exit code 0xc0000005.

This pull request is the pyo3 part of https://github.com/PyO3/maturin/pull/399.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -831,11 +831,21 @@ fn abi3_without_interpreter() -> Result<()> {
     }
     println!("cargo:rustc-cfg=py_sys_config=\"WITH_THREAD\"");
     println!("cargo:python_flags={}", flags);
+
+    // Unfortunately, on windows we can't build without at least providing
+    // python.lib to the linker. While maturin tells the linker the location
+    // of python.lib, we need to do the renaming here, otherwise cargo
+    // complains that the crate using pyo3 does not contains a `#[link(...)]`
+    // attribute with pythonXY.
+    if env::var("CARGO_CFG_TARGET_FAMILY")? == "windows" {
+        println!("cargo:rustc-link-lib=pythonXY:python3");
+    }
+
     Ok(())
 }
 
 fn main() -> Result<()> {
-    // If PYO3_NO_PYTHON is set with abi3, we can build PyO3 without calling Python (UNIX only).
+    // If PYO3_NO_PYTHON is set with abi3, we can build PyO3 without calling Python.
     // We only check for the abi3-py3{ABI3_MAX_MINOR} because lower versions depend on it.
     if env::var_os("PYO3_NO_PYTHON").is_some()
         && env::var_os(format!("CARGO_FEATURE_ABI3_PY3{}", ABI3_MAX_MINOR)).is_some()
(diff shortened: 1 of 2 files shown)


kngwyu (approved):
Thanks!
- You may want to fix the comment on line 848
- Although I don't fully understand what maturin does for this, can I understand that we need some trick for Windows? And should this feature be still optional, invoked by `PYO3_NO_PYTHON`?

davidhewitt (approved):
👍 

I think this needs to continue to be optional, because if you want to build tests with the `abi3-py36` feature present then you'll need to be able to link still.

Only thing I wonder is whether we should also add `PYO3_LIB_DIR` so that maturin can pass that to pyo3 rather than have to add a link flag itself.

konstin:
> You may want to fix the comment on line 848

Thanks, done.

> Although I don't fully understand what maturin does for this, can I understand that we need some trick for Windows?

On [maturin's side](https://github.com/PyO3/maturin/pull/399/files#diff-38031c4b383891f1cdddffbaad442f2ade8185879641edaf9cbaca059f1d29a6R69-R71), I set `-L native={base_prefix}\libs` to rustc's arguments. 

> And should this feature be still optional, invoked by `PYO3_NO_PYTHON`?

> I think this needs to continue to be optional, because if you want to build tests with the `abi3-py36` feature present then you'll need to be able to link still.

yep, it's guarded by `PYO3_NO_PYTHON`.

> Only thing I wonder is whether we should also add `PYO3_LIB_DIR` so that maturin can pass that to pyo3 rather than have to add a link flag itself.

I've also thought about this, but I think it's simpler the current way.

davidhewitt:
> I've also thought about this, but I think it's simpler the current way.

Ok. It's easy to add in the future if we decide we want this.

kngwyu:
> Thanks, done.

Maybe you forgot to push the commit ... ?

davidhewitt:
To get the release moving I'm going to force-push a fix for the commit and merge this. Thanks very much!

konstin:
> Maybe you forgot to push the commit ... ?

Oops, sorry 

> To get the release moving I'm going to force-push a fix for the commit and merge this. 

Thanks!

PR #1372, opened 2021-01-09
changelog: add Deprecated sections

davidhewitt:
[Keep a Changelog](https://keepachangelog.com/en/1.0.0/) suggests having deprecations as a separate section... I think we should also do that as it's helpful for users to know about methods that will be removed.


kngwyu (approved):
Thanks!

Issue #1377, opened 2021-01-10
Implement PyClass for PathBuf and Path


kangalioo:
There exists an infallible conversion from String to PathBuf (and hence Path), so if I'm not missing something, there is no real reason not to implement PyClass for PathBuf/Path.

That would allow Rust functions exposed to Python to take paths as parameters with the actual appropriate Rust type, instead of having to use String as a workaround.

kangalioo:
I'm trying to make a PR for this

davidhewitt:
I think rather than `PyClass`, you should implement `IntoPy<PyObject>`, `ToPyObject`, and `FromPyObject` for `PathBuf` (and maybe `Path`, if you can make it work).

kangalioo:
I think it will be better to implement those three traits on OsStr and OsString, because that's what the this work is really about: converting between Python's magic strings and Rust's opaque OS strings.

Path and PathBuf can just redirect to the implementations on OsStr and OsString.

davidhewitt:
That sounds sensible to me! If you're looking at string types, it could be worth assessing if `CStr` and `CString` should also be supported. I think Python allows null bytes in the middle of its strings so definitely from-Python would be fallible for those.

kangalioo:
Should the conversion functions from/to OsStr/Cstr go onto the PyString struct? It would get four new methods:
- `from_osstr`
- `to_osstr`
- `from_cstr`
- `to_cstr`

The alternative would be to create new modules `types/cstr.rs` and `types/osstr.rs`, and implement the conversions there

davidhewitt:
I would suggest avoiding new methods on `PyString` if we can - as much as possible the methods on the python-native types are just wrappers around the underlying Python C API.

Is there a reason the traits above don't provide enough functionality for what you need? The traits are all we need for `#[pyfunction]` / `#[pymethods]` etc.

kangalioo:
I was just asking what you preferred, I'm fine with both.


In any case, I wrote the trait implementations for OsStr/OsString: https://github.com/kangalioo/pyo3/blob/bbc578fda767134ff678c1ebc62eae5e0e756cb2/src/types/osstr.rs. I'd greatly appreciate if you could check it out to make sure I didn't do anything totally stupid. Commenst with `// HELP` mark places where I was unsure if I was doing the right thing.

davidhewitt:
👍 totally makes sense - nice work and thanks for the PR; I'll do my best to make time to review it as soon as I can.

[...]

PR #1379, opened 2021-01-10
OsStr and Path conversions

kangalioo:
Implements conversions between OsStr/OsString/Path/PathBuf and Python strings: #1377

This is my first PR on this project and my first interaction with the codebase

This PR is a draft because in several places of the code I'm not sure if I did the right thing. Those places are marked with a `// HELP` comment. I'd greatly appreciate if an experienced PyO3 maintainer could check those places.

### DIFF
--- a/src/conversions/osstr.rs
+++ b/src/conversions/osstr.rs
@@ -1,0 +1,202 @@
+use crate::types::PyString;
+#[cfg(windows)]
+use crate::PyErr;
+use crate::PyNativeType;
+use crate::{
+    ffi, AsPyPointer, FromPyObject, IntoPy, PyAny, PyObject, PyResult, PyTryFrom, Python,
+    ToPyObject,
+};
+use std::borrow::Cow;
+use std::ffi::{OsStr, OsString};
+#[cfg(not(windows))]
+use std::os::raw::c_char;
+
+impl ToPyObject for OsStr {
+    fn to_object(&self, py: Python) -> PyObject {
+        // If the string is UTF-8, take the quick and easy shortcut
+        if let Some(valid_utf8_path) = self.to_str() {
+            return valid_utf8_path.to_object(py);
+        }
+
+        // All targets besides windows support the std::os::unix::ffi::OsStrExt API:
+        // https://doc.rust-lang.org/src/std/sys_common/mod.rs.html#59
+        #[cfg(not(windows))]
+        {
+            let bytes = std::os::unix::ffi::OsStrExt::as_bytes(self);
+
+            let ptr = bytes.as_ptr() as *const c_char;
+            let len = bytes.len() as ffi::Py_ssize_t;
+            unsafe {
+                // DecodeFSDefault automatically chooses an appropriate decoding mechanism to
+                // parse os strings losslessly (i.e. surrogateescape most of the time)
+                let pystring = ffi::PyUnicode_DecodeFSDefaultAndSize(ptr, len);
+                PyObject::from_owned_ptr(py, pystring)
+            }
+        }
+
+        #[cfg(windows)]
+        {
+            let wstr: Vec<u16> = std::os::windows::ffi::OsStrExt::encode_wide(self).collect();
+
+            unsafe {
+                // This will not panic because the data from encode_wide is well-formed Windows
+                // string data
+                PyObject::from_owned_ptr(
+                    py,
+                    ffi::PyUnicode_FromWideChar(wstr.as_ptr(), wstr.len() as ffi::Py_ssize_t),
+                )
+            }
+        }
+    }
+}
+
+// There's no FromPyObject implementation for &OsStr because albeit possible on Unix, this would
+// be impossible to implement on Windows. Hence it's omitted entirely
+
+impl FromPyObject<'_> for OsString {
+    fn extract(ob: &PyAny) -> PyResult<Self> {
+        let pystring = <PyString as PyTryFrom>::try_from(ob)?; // Cast PyAny to PyString
+
+        #[cfg(not(windows))]
+        {
+            // Decode from Python's lossless bytes string representation back into raw bytes
+            let fs_encoded_bytes = unsafe {
+                crate::Py::<crate::types::PyBytes>::from_owned_ptr(
+                    ob.py(),
+                    ffi::PyUnicode_EncodeFSDefault(pystring.as_ptr()),
+                )
+            };
+
+            // Create an OsStr view into the raw bytes from Python
+            let os_str: &OsStr = std::os::unix::ffi::OsStrExt::from_bytes(
+                fs_encoded_bytes.as_ref(ob.py()).as_bytes(),
+            );
+
+            Ok(os_str.to_os_string())
+        }
+
+        #[cfg(windows)]
+        {
+            // Take the quick and easy shortcut if UTF-8
+            if let Ok(utf8_string) = pystring.to_str() {
+                return Ok(utf8_string.to_owned().into());
+            }
+
+            // Get an owned allocated wide char buffer from PyString, which we have to deallocate
+            // ourselves
+            let size =
+                unsafe { ffi::PyUnicode_AsWideChar(pystring.as_ptr(), std::ptr::null_mut(), 0) };
+            if size == -1 {
+                return Err(PyErr::fetch(ob.py()));
+            }
+
+            let mut buffer = vec![0; size as usize];
+            let bytes_read =
+                unsafe { ffi::PyUnicode_AsWideChar(pystring.as_ptr(), buffer.as_mut_ptr(), size) };
+            assert_eq!(bytes_read, size);
+
+            // Copy wide char buffer into OsString
+            let os_string = std::os::windows::ffi::OsStringExt::from_wide(&buffer);
+
+            Ok(os_string)
+        }
+    }
+}
+
+impl IntoPy<PyObject> for &'_ OsStr {
+    #[inline]
+    fn into_py(self, py: Python) -> PyObject {
+        self.to_object(py)
+    }
+}
+
+impl ToPyObject for Cow<'_, OsStr> {
+    #[inline]
+    fn to_object(&self, py: Python) -> PyObject {
+        (&self as &OsStr).to_object(py)
+    }
+}
+
+impl IntoPy<PyObject> for Cow<'_, OsStr> {
+    #[inline]
+    fn into_py(self, py: Python) -> PyObject {
+        self.to_object(py)
+    }
+}
+
+impl ToPyObject for OsString {
+    #[inline]
+    fn to_object(&self, py: Python) -> PyObject {
+        (&self as &OsStr).to_object(py)
+    }
+}
+
+impl IntoPy<PyObject> for OsString {
+    fn into_py(self, py: Python) -> PyObject {
+        self.to_object(py)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::{types::PyString, IntoPy, PyObject, Python, ToPyObject};
+    use std::fmt::Debug;
+    use std::{
+        borrow::Cow,
+        ffi::{OsStr, OsString},
+    };
+
+    #[test]
+    #[cfg(not(windows))]
+    fn test_non_utf8_conversion() {
+        Python::with_gil(|py| {
+            use std::os::unix::ffi::OsStrExt;
+
+            // this is not valid UTF-8
+            let payload = &[250, 251, 252, 253, 254, 255, 0, 255];
+            let os_str = OsStr::from_bytes(payload);
+
+            // do a roundtrip into Pythonland and back and compare
+            let py_str: PyObject = os_str.into_py(py);
+            let os_str_2: OsString = py_str.extract(py).unwrap();
+            assert_eq!(os_str, os_str_2);
+        });
+    }
+
+    #[test]
+    fn test_topyobject_roundtrip() {
+        Python::with_gil(|py| {
+            fn test_roundtrip<T: ToPyObject + AsRef<OsStr> + Debug>(py: Python, obj: T) {
+                let pyobject = obj.to_object(py);
+                let pystring: &PyString = pyobject.extract(py).unwrap();
+                assert_eq!(pystring.to_string_lossy(), obj.as_ref().to_string_lossy());
+                let roundtripped_obj: OsString = pystring.extract().unwrap();
+                assert_eq!(obj.as_ref(), roundtripped_obj.as_os_str());
+            }
+            let os_str = OsStr::new("Hello\0\n🐍");
+            test_roundtrip::<&OsStr>(py, os_str);
+            test_roundtrip::<Cow<'_, OsStr>>(py, Cow::Borrowed(os_str));
+            test_roundtrip::<Cow<'_, OsStr>>(py, Cow::Owned(os_str.to_os_string()));
+            test_roundtrip::<OsString>(py, os_str.to_os_string());
+        });
+    }
+
+    #[test]
+    fn test_intopy_roundtrip() {
+        Python::with_gil(|py| {
+            fn test_roundtrip<T: IntoPy<PyObject> + AsRef<OsStr> + Debug + Clone>(
+                py: Python,
+                obj: T,
+            ) {
+                let pyobject = obj.clone().into_py(py);
+                let pystring: &PyString = pyobject.extract(py).unwrap();
+                assert_eq!(pystring.to_string_lossy(), obj.as_ref().to_string_lossy());
+                let roundtripped_obj: OsString = pystring.extract().unwrap();
+                assert!(obj.as_ref() == roundtripped_obj.as_os_str());
+            }
+            let os_str = OsStr::new("Hello\0\n🐍");
+            test_roundtrip::<&OsStr>(py, os_str);
+            test_roundtrip::<OsString>(py, os_str.to_os_string());
+        })
+    }
+}
--- a/src/conversions/path.rs
+++ b/src/conversions/path.rs
@@ -1,0 +1,115 @@
+use crate::{FromPyObject, IntoPy, PyAny, PyObject, PyResult, Python, ToPyObject};
+use std::borrow::Cow;
+use std::ffi::OsString;
+use std::path::{Path, PathBuf};
+
+impl ToPyObject for Path {
+    fn to_object(&self, py: Python) -> PyObject {
+        self.as_os_str().to_object(py)
+    }
+}
+
+// See osstr.rs for why there's no FromPyObject impl for &Path
+
+impl FromPyObject<'_> for PathBuf {
+    fn extract(ob: &PyAny) -> PyResult<Self> {
+        Ok(PathBuf::from(OsString::extract(ob)?))
+    }
+}
+
+impl<'a> IntoPy<PyObject> for &'a Path {
+    #[inline]
+    fn into_py(self, py: Python) -> PyObject {
+        self.as_os_str().to_object(py)
+    }
+}
+
+impl<'a> ToPyObject for Cow<'a, Path> {
+    #[inline]
+    fn to_object(&self, py: Python) -> PyObject {
+        self.as_os_str().to_object(py)
+    }
+}
+
+impl<'a> IntoPy<PyObject> for Cow<'a, Path> {
+    #[inline]
+    fn into_py(self, py: Python) -> PyObject {
+        self.to_object(py)
+    }
+}
+
+impl ToPyObject for PathBuf {
+    #[inline]
+    fn to_object(&self, py: Python) -> PyObject {
+        self.as_os_str().to_object(py)
+    }
+}
+
+impl IntoPy<PyObject> for PathBuf {
+    fn into_py(self, py: Python) -> PyObject {
+        self.into_os_string().to_object(py)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::{types::PyString, IntoPy, PyObject, Python, ToPyObject};
+    use std::borrow::Cow;
+    use std::fmt::Debug;
+    use std::path::{Path, PathBuf};
+
+    #[test]
+    #[cfg(not(windows))]
+    fn test_non_utf8_conversion() {
+        Python::with_gil(|py| {
+            use std::ffi::OsStr;
+            use std::os::unix::ffi::OsStrExt;
+
+            // this is not valid UTF-8
+            let payload = &[250, 251, 252, 253, 254, 255, 0, 255];
+            let path = Path::new(OsStr::from_bytes(payload));
+
+            // do a roundtrip into Pythonland and back and compare
+            let py_str: PyObject = path.into_py(py);
+            let path_2: PathBuf = py_str.extract(py).unwrap();
+            assert_eq!(path, path_2);
+        });
+    }
+
+    #[test]
+    fn test_topyobject_roundtrip() {
+        Python::with_gil(|py| {
+            fn test_roundtrip<T: ToPyObject + AsRef<Path> + Debug>(py: Python, obj: T) {
+                let pyobject = obj.to_object(py);
+                let pystring: &PyString = pyobject.extract(py).unwrap();
+                assert_eq!(pystring.to_string_lossy(), obj.as_ref().to_string_lossy());
+                let roundtripped_obj: PathBuf = pystring.extract().unwrap();
+                assert_eq!(obj.as_ref(), roundtripped_obj.as_path());
+            }
+            let path = Path::new("Hello\0\n🐍");
+            test_roundtrip::<&Path>(py, path);
+            test_roundtrip::<Cow<'_, Path>>(py, Cow::Borrowed(path));
+            test_roundtrip::<Cow<'_, Path>>(py, Cow::Owned(path.to_path_buf()));
+            test_roundtrip::<PathBuf>(py, path.to_path_buf());
+        });
+    }
+
+    #[test]
+    fn test_intopy_roundtrip() {
+        Python::with_gil(|py| {
+            fn test_roundtrip<T: IntoPy<PyObject> + AsRef<Path> + Debug + Clone>(
+                py: Python,
+                obj: T,
+            ) {
+                let pyobject = obj.clone().into_py(py);
+                let pystring: &PyString = pyobject.extract(py).unwrap();
+                assert_eq!(pystring.to_string_lossy(), obj.as_ref().to_string_lossy());
+                let roundtripped_obj: PathBuf = pystring.extract().unwrap();
+                assert!(obj.as_ref() == roundtripped_obj.as_path());
+            }
+            let path = Path::new("Hello\0\n🐍");
+            test_roundtrip::<&Path>(py, path);
+            test_roundtrip::<PathBuf>(py, path.to_path_buf());
+        })
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,9 @@ PyO3 versions, please see the [migration guide](https://pyo3.rs/master/migration
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [Unreleased]
 ### Added
+- Add conversions between `OsStr`/`OsString`/`Path`/`PathBuf` and Python strings. [#1379](https://github.com/PyO3/pyo3/pull/1379)
 - Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
 
 ### Changed
--- a/guide/src/conversions/tables.md
+++ b/guide/src/conversions/tables.md
@@ -13,7 +13,7 @@ The table below contains the Python type and the corresponding function argument
 | Python        | Rust                            | Rust (Python-native) |
 | ------------- |:-------------------------------:|:--------------------:|
 | `object`      | -                               | `&PyAny`             |
-| `str`         | `String`, `Cow<str>`, `&str`    | `&PyUnicode`         |
+| `str`         | `String`, `Cow<str>`, `&str`, `OsString`, `PathBuf` | `&PyUnicode` |
 | `bytes`       | `Vec<u8>`, `&[u8]`              | `&PyBytes`           |
 | `bool`        | `bool`                          | `&PyBool`            |
 | `int`         | Any integer type (`i32`, `u32`, `usize`, etc) | `&PyLong` |
--- a/src/conversions/mod.rs
+++ b/src/conversions/mod.rs
@@ -1,0 +1,5 @@
+//! This module contains conversions between non-String Rust object and their string representation
+//! in Python
+
+mod osstr;
+mod path;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -180,6 +180,7 @@ pub mod buffer;
 pub mod callback;
 pub mod class;
 pub mod conversion;
+mod conversions;
 #[macro_use]
 #[doc(hidden)]
 pub mod derive_utils;
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -131,6 +131,13 @@ impl<'a> ToPyObject for Cow<'a, str> {
     }
 }
 
+impl IntoPy<PyObject> for Cow<'_, str> {
+    #[inline]
+    fn into_py(self, py: Python) -> PyObject {
+        self.to_object(py)
+    }
+}
+
 /// Converts a Rust `String` to a Python object.
 /// See `PyString::new` for details on the conversion.
 impl ToPyObject for String {


src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+        let fs_encoded_bytes: &crate::types::PyBytes = unsafe {
+            ob.py()
+                .from_borrowed_ptr(ffi::PyUnicode_EncodeFSDefault(pystring.as_ptr()))
+        };
+
+        Ok(std::os::unix::ffi::OsStrExt::from_bytes(
+            fs_encoded_bytes.as_bytes(),
+        ))
+    }
+}
```
davidhewitt, line 74: Hmm yeah I agree it's undesirable to have this trait only work on some platforms. A lot of users turn to Rust for the easy multiple-platform support.

Also, we're in the process of designing a rewrite of PyO3's object system which would result in better performance and more predictable memory management (#1308). However I think with that change, this trait would be incompatible because it depends on the temporary `PyBytes` extraction.

As we have `FromPyObject for OsString`, perhaps best to skip out on this trait in this PR? Can leave an issue and/or FIXME and I'm happy to revisit once #1308 is more complete.

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+            let wstr: Vec<u16> = std::os::windows::ffi::OsStrExt::encode_wide(self).collect();
+
+            let ptr = wstr.as_ptr();
+            let len = wstr.len() as ffi::Py_ssize_t;
+            unsafe {
+                // HELP: is there some better thing to do here than panic? Some last resort thing to
+                // try?
+                py.from_owned_ptr_or_err::<PyString>(ffi::PyUnicode_FromWideChar(ptr, len))
+                    .expect("OsStr conversion to Python string via FromWideChar failed")
+                    .into()
```
davidhewitt, line 50: I'd just use `PyObject::from_owned_ptr` here.

```suggestion
                PyObject::from_owned_ptr(py, ffi::PyUnicode_FromWideChar(ptr, len))
```

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+        {
+            let bytes = std::os::unix::ffi::OsStrExt::as_bytes(self);
+
+            let ptr = bytes.as_ptr() as *const c_char;
+            let len = bytes.len() as ffi::Py_ssize_t;
+            unsafe {
+                // HELP: is this the appropriate function? Does it use the surrogateescape error
+                // handler? Should it be PyUnicode_DecodeLocaleAndSize instead maybe?
+                let pystring = ffi::PyUnicode_DecodeFSDefaultAndSize(ptr, len);
+                py.from_owned_ptr::<PyString>(pystring).into()
```
davidhewitt, line 35: Marginally faster:

```suggestion
                PyObject::from_owned_ptr(py, pystring)
```

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+use std::borrow::Cow;
+use std::ffi::{OsStr, OsString};
+#[cfg(not(windows))]
+use std::os::raw::c_char;
+
+impl ToPyObject for OsStr {
+    fn to_object(&self, py: Python) -> PyObject {
+        // If the string is UTF-8, take the quick and easy shortcut
+        if let Some(valid_utf8_path) = self.to_str() {
+            return PyString::new(py, valid_utf8_path).into();
```
davidhewitt, line 17: Could even just use `ToPyObject` to `str`.

```suggestion
            return valid_utf8_path.to_object();
```

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+            // Get an owned allocated wide char buffer from PyString, which we have to deallocate
+            // ourselves
+            let mut size: ffi::Py_ssize_t = 0;
+            let buffer_ptr =
+                unsafe { ffi::PyUnicode_AsWideCharString(pystring.as_ptr(), &mut size) };
+
+            // Copy wide char buffer into OsString
+            let os_string = std::os::windows::ffi::OsStringExt::from_wide(unsafe {
+                std::slice::from_raw_parts(buffer_ptr, size as usize)
+            });
```
davidhewitt, line 110: Needs a check for null pointer on `buffer_ptr` before trying to copy from it! I think this should do:

```
if buffer_ptr.is_null() {
    return PyErr::fetch(py);
}
```

NB maybe better to use `Vec` - see comment on line 105.

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+
+            // do a roundtrip into Pythonland and back and compare
+            let py_str = os_str.into_py(py);
+            let os_str_2: OsString = py_str.extract(py).unwrap();
+            assert_eq!(os_str, os_str_2);
+        })
+    }
+
+    // HELP: Are these tests enough? I was unsure how to best test it, because of the platform-
+    // dependant nature of OsStr/OsString
```
davidhewitt, line 190: We recently used `proptest` to test roundtripping of u128 / i128 - would be a great fit here too as it'll help us sniff out any edge cases where we have issues.

src/types/path.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn to_object(&self, py: Python) -> PyObject {
+        self.as_os_str().to_object(py)
+    }
+}
+
+// See osstr.rs for why this is Unix-only
+// HELP: should this implementation be here at all? It may be confusing for PyO3 users when a trait
+// is implemented only on some platforms
+#[cfg(not(windows))]
+impl<'a> FromPyObject<'a> for &'a Path {
```
davidhewitt, line 18: Mmm same again I'll try to come back to this one in the future.

src/types/path.rs
```diff
@@ -0,0 +1,66 @@
...
+impl<'a> IntoPy<PyObject> for &'a PathBuf {
+    #[inline]
+    fn into_py(self, py: Python) -> PyObject {
+        self.as_os_str().to_object(py)
+    }
+}
+
+// These trait implementations simply forward to the OsStr/OsString trait implementations, so I
+// haven't put the effort to test all of this
+// HELP: should I have?
```
davidhewitt, line 66: I appreciate it's a little tedious but would prefer to have some tests here too! You never know when I might come along and accidentally break things 🙃

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+            let buffer_ptr =
+                unsafe { ffi::PyUnicode_AsWideCharString(pystring.as_ptr(), &mut size) };
+
+            // Copy wide char buffer into OsString
+            let os_string = std::os::windows::ffi::OsStringExt::from_wide(unsafe {
+                std::slice::from_raw_parts(buffer_ptr, size as usize)
+            });
+
+            unsafe {
+                ffi::PyMem_Free(buffer_ptr as *mut libc::c_void);
```
davidhewitt, line 113: NB we will leak memory if `OsStringExt::from_wide` panics - easiest fix is probably to wrap `buffer_ptr` in a struct which calls `PyMem_Free` in a `Drop` impl. NB maybe better to use `Vec` - see comment on line 105.

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+        #[cfg(not(windows))]
+        {
+            let bytes = std::os::unix::ffi::OsStrExt::as_bytes(self);
+
+            let ptr = bytes.as_ptr() as *const c_char;
+            let len = bytes.len() as ffi::Py_ssize_t;
+            unsafe {
+                // HELP: is this the appropriate function? Does it use the surrogateescape error
+                // handler? Should it be PyUnicode_DecodeLocaleAndSize instead maybe?
+                let pystring = ffi::PyUnicode_DecodeFSDefaultAndSize(ptr, len);
```
davidhewitt, line 34: Good question - I think probably `DecodeLocaleAndSize` is the right choice 👍

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+            // copying.
+            // Second, we have to convert the wide char buffer into a OsString. Rust provides no
+            // way to cheaply convert an owned buffer into an OsString; it only exposes a method
+            // involving a copy and an allocation.
+
+            // Get an owned allocated wide char buffer from PyString, which we have to deallocate
+            // ourselves
+            let mut size: ffi::Py_ssize_t = 0;
+            let buffer_ptr =
+                unsafe { ffi::PyUnicode_AsWideCharString(pystring.as_ptr(), &mut size) };
```
davidhewitt, line 105: It could be easier to allocate a `Vec<u16>` of the correct size and use `PyUnicode_AsWideChar` instead.

That would avoid potential issues with `buffer_ptr` being null and also risk of memory leak on panic. (See other comments.)

src/types/osstr.rs
```diff
@@ -0,0 +1,191 @@
...
+        #[cfg(windows)]
+        {
+            let pystring = <PyString as PyTryFrom>::try_from(ob)?; // Cast PyAny to PyString
+
+            // Take the quick and easy shortcut if UTF-8
+            if let Ok(utf8_string) = pystring.to_str() {
+                return Ok(utf8_string.to_owned().into());
+            }
+
+            // The Windows way involves two allocations+copies, unfortunately.
```
davidhewitt, line 93: Technically the unix one does too (just that one of them is `PyBytes` which makes it less obvious). 😄

davidhewitt (commented):
Thanks very much for this PR! This is a really great start. I've given it a thorough read and got various feedback and suggestions from my time spent in battle with the PyO3 codebase 💂.

kangalioo:
Wow, thank you!! That is one detailed code review. I will go over it tomorrow and implement all the suggestions

src/types/osstr.rs
kngwyu: > However I think with that change, this trait would be incompatible because it depends on the temporary PyBytes extraction.

`&'a PyAny` would become `PyAny<'a>`, and how the lifetime works would not change?
I agree that this trait is not very necessary (since paths are often not so long and copying is not a large burden), though.

src/types/osstr.rs
kngwyu: It's somewhat complex, but would look like:
```rust
let size = ffi::PyUnicode_AsWideChar(pystring.as_ptr(), ptr::null_mut(), 0);
if size == -1 {
    return Err(...)
}
let mut buffer = vec![0; size];
assert!(ffi::PyUnicode_AsWideChar(pystring.as_ptr(), buffer.as_mut_ptr(), size), size - 1);

File: src/types/osstr.rs [hunk 13]
```diff
@@ -0,0 +1,191 @@
...
+                py.from_owned_ptr::<PyString>(pystring).into()
+            }
+        }
+
+        #[cfg(windows)]
+        {
+            let wstr: Vec<u16> = std::os::windows::ffi::OsStrExt::encode_wide(self).collect();
+
+            let ptr = wstr.as_ptr();
+            let len = wstr.len() as ffi::Py_ssize_t;
```
kngwyu, line 44: BTW, you can just do `ffi::PyUnicode_FromWideChar(ptr, wstr.len() as _)`

davidhewitt:
I was thinking this morning that we should check to understand carefully the differences in behavior on unix vs windows. If String -> OsString -> PyString gives different results on the two platforms this might trip up users.

Also `pathlib.Path` doesn't have a C-API, but if it gets one we probably want to make Rust paths convert back and forth with those instead of strings. Potentially worth keeping in mind.

src/types/osstr.rs
kangalioo: I'm curious why? I'd tentatively guess that PyUnicode_DecodeFSDefaultAndSize is a special case of PyUnicode_DecodeLocaleAndSize, but it chooses the error handler appropriate for file paths; the same one that Python itself uses. The Python docs don't say much about specifics, unfortunately

src/types/path.rs
kangalioo: Yes, perhaps an integration test with an accompanied folder of weird file names (think emojis, long filenames, maybe invalid filenames etc.), and the integration test will have both Python and Rust read those filenames, transmit the filenames to the other language respectively, and assert_eq.

src/types/osstr.rs
kngwyu: I checked that `interp->fs_codec.error_handler` is used in `PyUnicode_DecodeFSDefaultAndSize`, but how is it different from other handlers?

kangalioo: The two error handlers I know of, and are mentioned in the Python docs, are "strict" and "surrogateescape". "strict" will error out if invalid Unicode data is encountered. "surrogateescape" will handle any invalid bytes by encoding them in the Low Surrogates Unicode block.

Therefore, "surrogateescape" can decode any arbitrary bytes losslessly by parsing UTF-8 where it can, and saving the rest as surrogates instead. Hence "surrogateescape" is well-suited for parsing filenames, since Unix filenames _are_ actually arbitrary bytes.

I did an experiment:
```sh
➜  ~ touch $(echo -e "\xFA\xFB\xFC\xFD\xFE\xFF") # invalid UTF-8
➜  ~ python3
>>> import os; os.listdir()
['\udcfa\udcfb\udcfc\udcfd\udcfe\udcff']
```
As you can see, the invalid bytes I put into the filename are being encoded in the Unicode `0xdc__` block, which strongly suggests that Python uses "surrogateescape" as its file system string parser.

All this complexity is encapsulated by using PyUnicode_DecodeFSDefaultAndSize, which is specifically made for filesystem operations and does _the right thing™_. Hence I tend to prefer PyUnicode_DecodeFSDefaultAndSize over PyUnicode_DecodeLocaleAndSize

kangalioo:
> I was thinking this morning that we should check to understand carefully the differences in behavior on unix vs windows. If String -> OsString -> PyString gives different results on the two platforms this might trip up users.
> 
> Also `pathlib.Path` doesn't have a C-API, but if it gets one we probably want to make Rust paths convert back and forth with those instead of strings. Potentially worth keeping in mind.

I think this can be taken care of using unit or integration tests. We could feed many different strings through the String -> OsString -> PyString conversion and back. This test will be run on all OSes by CI, so we know if something breaks somewhere.

Also, is there a reason you reacted confused on my previous comment? Now I am confused :smile:

src/types/osstr.rs
kangalioo: I put the `.expect()` there as a sort of temporary measure, because I didn't know under what circumstances PyUnicode_FromWideChar would fail. There may be circumstances which we can handle appropriately.

Rust's encode_wide "re-encodes an OsStr as a wide character sequence, i.e., potentially ill-formed UTF-16." and PyUnicode_FromWideChar "creates a Unicode object from the wchar_t buffer w of the given size." The "potentially ill-formed" part trips me up. We don't want PyO3 to panic on regular usage, right?

src/types/osstr.rs
kngwyu: Thank you for your explanation, but isn't it look like the handler is not always `surrogateescape`?
https://github.com/python/cpython/blob/v3.9.1/Python/initconfig.c#L1695

Anyway, I agree that `DecodeFSDefault` is better. There's no merit in choosing the error handler by ourselves.

kangalioo: Heh, seems like in reality it's even more complex than what I described above. It applies a third error handler for "legacy Windows". I guess this makes it even clearer that it's probably better to leave the error handler decision to Python

src/types/osstr.rs
kangalioo: I decided to stick with PyUnicode_AsWideCharString and write a `defer_pymem_free` function to handle the deallocation robustly. In my opinion, that way of doing it is simpler and less error-prone than the code snippet by kngwyu

src/types/osstr.rs
kangalioo: On second thought, the "potentially ill-formed" probably refers to the fact that Windows uses some unholy modified variant of UTF-16 internally. I'm pretty sure that the data emitted from encode_wide exactly compatible with PyUnicode_FromWideChar. So, no errors should happen and the hidden panic is fine (because it signifies some unforeseen error like a programming bug)

src/types/osstr.rs
kangalioo: I did it for consistency with the block above. Also, with rustfmt, that change would actually increase the number of lines. Hence I'm gonna keep it like before

kangalioo:
By the way, why do all Rust->Python conversions have to be duplicated? There's two traits to implement: `IntoPy<PyObject>` and `ToPyObject` and I don't really see the meaningful difference.

This is a bit awkward in the test because I essentially have to test everything twice to account for all the trait implementations

kngwyu:
>  why do all Rust->Python conversions have to be duplicated?

It's a common consume/ not consume pattern in Rust, though it does not work nicely sometimes.
I think there are two main problems here (1. the generic parameter of `IntoPy` is not effectively used 2. we rarely need `into` operation), which should be discussed in the meantime.

> I essentially have to test everything twice

Since they share the implementation, I don't think we need to test all things twice.

src/conversions/path.rs
```diff
@@ -0,0 +1,116 @@
...
+                let pystring: &PyString = pyobject.extract(py).unwrap();
+                assert_eq!(pystring.to_string_lossy(), obj.as_ref().to_string_lossy());
+                let roundtripped_obj: PathBuf = pystring.extract().unwrap();
+                assert_eq!(obj.as_ref(), roundtripped_obj.as_path());
+            }
+            let path = Path::new("Hello\0\n🐍");
+            test_roundtrip::<&Path>(py, path);
+            test_roundtrip::<Cow<'_, Path>>(py, Cow::Borrowed(path));
+            test_roundtrip::<Cow<'_, Path>>(py, Cow::Owned(path.to_path_buf()));
+            test_roundtrip::<PathBuf>(py, path.to_path_buf());
```
kngwyu, line 93: If `.as_ref()` is called at last, do we need to test all these?

kangalioo: .as_ref() is only called to verify the end result. The conversion is done on the types as is.

kangalioo:
It seems like the codecov check fail is a false positive. The places it marked as "not covered by tests" _are_ actually covered by the tests. Perhaps codecov is getting confused by the generic trickery in the tests.

src/conversions/osstr.rs
```diff
@@ -0,0 +1,230 @@
...
+    ToPyObject,
+};
+use std::borrow::Cow;
+use std::ffi::{OsStr, OsString};
+#[cfg(not(windows))]
+use std::os::raw::c_char;
+
+// TODO: move to other module to prevent accidentally circumventing the new function?
+#[cfg(windows)]
+struct DropGuard<T>(*mut T);
```
kngwyu, line 16: I don't think it's good to abstract such a specific operation.
In this struct, `*mut T` *implicitly* must be a pointer in the Python heap without any reference to Python objects, so it's unsafe to construct this object.
So I recommend making this a more limited one without `new` in the function scope if you like this solution much more than `Vec.`

src/types/osstr.rs
davidhewitt: Yes I agree let's go with `DecodeFSDefault` - I'd mistakenly understood previously that we want `surrogateescape` in all circumstances.

src/conversions/osstr.rs
davidhewitt: Agreed, this type is super `unsafe` so I'd rather it was inside the function too. That way only the function can misuse it.

davidhewitt:
> Also, is there a reason you reacted confused on my previous comment? Now I am confused 😄

Hehe nope that's just a fat-finger error when checking on the conversation on my phone!

> There's two traits to implement: `IntoPy<PyObject>` and `ToPyObject` and I don't really see the meaningful difference.

Yes; I would love to rework this pair to also potentially model fallible conversions, but to avoid ecosystem churn until we're really sure what a better design is I think best to leave as-is!

> It seems like the codecov check fail is a false positive. The places it marked as "not covered by tests" are actually covered by the tests. Perhaps codecov is getting confused by the generic trickery in the tests.

I think `#[inline]` functions can have issues with coverage. Probably the generics in the tests got inlined too. Don't need to worry too hard about it; it's annoying and I'd love to improve that one day, but it's fine for now.

----

Also, I was thinking we might want to consider whether `PyString::to_string_lossy` gives the same output as converting first to `OsString` and then calling `OsStr::to_string_lossy`. Users might assume they're equivalent, though I'm not sure if we should expect them to be.

konstin:
> > There's two traits to implement: `IntoPy<PyObject>` and `ToPyObject` and I don't really see the meaningful difference.
> 
> Yes; I would love to rework this pair to also potentially model fallible conversions, but to avoid ecosystem churn until we're really sure what a better design is I think best to leave as-is!

My original plan was to have `IntoPy<T>` and `FromPy<T>` mirror std's excellent [`Into<T>`](https://doc.rust-lang.org/std/convert/trait.Into.html) and [`From<T>`](https://doc.rust-lang.org/std/convert/trait.From.html), with everything equivalent except for an additional gil token parameter. I also removed some of the conversion traits (such as `IntoPyObject` and `IntoPyTuple`), but I never finished so the current design is kinda half-baked.

src/types/osstr.rs
kangalioo: Actually, I changed my mind :sweat_smile: The version with a custom allocated Vec is not so bad after all.

However I'm wondering why we can assume that PyUnicode_AsWideChar won't return any error in the second invocation?

Also, if I understand the Python docs correctly, the return value of PyUnicode_AsWideChar is the number of bytes read, not the total number of bytes in the string. If that is true, won't the `size` variable be zero and the whole code blow up?

davidhewitt:
> However I'm wondering why we can assume that PyUnicode_AsWideChar won't return any error in the second invocation?
> 
> Also, if I understand the Python docs correctly, the return value of PyUnicode_AsWideChar is the number of bytes read, not the total number of bytes in the string. If that is true, won't the size variable be zero and the whole code blow up?

Docs seem to be slightly misleading in this case - the implementation of `PyUnicode_AsWideChar` has slightly more helpful docs: https://github.com/python/cpython/blob/fb35fa49d192368e94ffec09c092260ed0fea2e1/Objects/unicodeobject.c#L3270

From that implementation it looks like that function will never raise an exception (as long as the first argument is a valid unicode object), and also that doc describes that passing `NULL` for the buffer causes the full string size to be returned.

src/types/osstr.rs
kngwyu: See https://github.com/PyO3/pyo3/pull/1379#issuecomment-758467258 (thanks David) and I found a small mistake in my snippet:
```rust
assert!(ffi::PyUnicode_AsWideChar(pystring.as_ptr(), buffer.as_mut_ptr(), size), size);
```
would be correct.

kangalioo:
I'm currently working on something else but I hope to continue work on this in the next few days

kangalioo:
> Also, I was thinking we might want to consider whether `PyString::to_string_lossy` gives the same output as converting first to `OsString` and then calling `OsStr::to_string_lossy`. Users might assume they're equivalent, though I'm not sure if we should expect them to be. Perhaps we can consider changing `PyString::to_string_lossy` implementation in this PR.

I looked at the source code of Python (`unicodeobject.c`) in an attempt to figure out what the "surrogatepass" error handler does exactlyy, which is used in PyO3's `PyString::to_string_lossy`. The Python documentation isn't helping either.

In other words, I'm not able to verify if `PyString::to_string_lossy` does the same thing as `PyString -> OsString -> OsString::to_string_lossy`, so at least I can't do it in this PR.

kangalioo:
Is there anything left to work on in this PR?

kngwyu (approved):
LGTM, thanks!

src/conversions/osstr.rs
```diff
@@ -0,0 +1,203 @@
...
+    fn extract(ob: &PyAny) -> PyResult<Self> {
+        #[cfg(not(windows))]
+        {
+            let pystring = <PyString as PyTryFrom>::try_from(ob)?; // Cast PyAny to PyString
+
+            // Decode from Python's lossless bytes string representation back into raw bytes
+            let fs_encoded_bytes: &crate::types::PyBytes = unsafe {
+                ob.py()
+                    .from_borrowed_ptr(ffi::PyUnicode_EncodeFSDefault(pystring.as_ptr()))
+            };
```
davidhewitt, line 67: `PyUnicode_EncodeFSDefault` returns a new reference, so we need to own this pointer or we leak memory. I suggest using `Py<PyBytes>`.

```suggestion
            let fs_encoded_bytes: Py<crate::types::PyBytes> = unsafe {
                Py::from_owned_ptr(ob.py(), ffi::PyUnicode_EncodeFSDefault(pystring.as_ptr()))
            };
```

src/conversions/path.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+impl<'a> IntoPy<PyObject> for &'a Path {
+    #[inline]
+    fn into_py(self, py: Python) -> PyObject {
+        self.as_os_str().to_object(py)
+    }
+}
+
+impl<'a> ToPyObject for Cow<'a, Path> {
```
davidhewitt, line 27: Also needs `IntoPy<PyObject>` for `Cow<'_, Path>`.

davidhewitt (changes requested):
I think we need to adjust the set of traits provided just slightly, see other comments.

---

R.E. `PyString::to_string_lossy` - I'm going to open a separate PR for that case. I ran a quick test script:

```
use pyo3::prelude::*;
use pyo3::types::{PyDict, PyString};
use std::borrow::Cow;
use std::error::Error;
use std::ffi::OsString;

fn main() -> PyResult<()> {
    Python::with_gil(|py| -> PyResult<()> {
        let locals = PyDict::new(py);
        py.run(
            r#"x = '\udcfa\udcfb\udcfc\udcfd\udcfe\udcff'"#,
            None,
            Some(locals),
        )?;
        let py_str = locals.get_item("x").unwrap().downcast::<PyString>()?;
        dbg!(py_str, py_str.len()?);
        let os_string: OsString = py_str.extract()?;
        dbg!(&os_string, &os_string.len());
        let string: Cow<str> = py_str.to_string_lossy();
        dbg!(&string, &string.chars().count());
        Ok(())
    })
}
```

The output was this:

```
[src/main.rs:16] py_str = '\udcfa\udcfb\udcfc\udcfd\udcfe\udcff'
[src/main.rs:16] py_str.len()? = 6
[src/main.rs:18] &os_string = "\xFA\xFB\xFC\xFD\xFE\xFF"
[src/main.rs:18] &os_string.len() = 6
[/home/david/dev/pyo3/src/types/string.rs:86] bytes = b'\xed\xb3\xba\xed\xb3\xbb\xed\xb3\xbc\xed\xb3\x
bd\xed\xb3\xbe\xed\xb3\xbf'
[src/main.rs:20] &string = "������������������"
[src/main.rs:20] &string.chars().count() = 18
```

It looks like the "surrogatepass" escape handler literally hands the three-byte surrogate sequences to Rust, which then results in three replacement characters being produced by `std::String::from_utf8_lossy`. I think this is wrong - Python thinks these surrogates are a single codepoint, but they're currently treated as three by `PyString::to_string_lossy`.

I think that using `PyUnicode_EncodeFSDefault` is probably better and will then match `PyString -> OsString -> String`.

davidhewitt:
> I think that using `PyUnicode_EncodeFSDefault` is probably better and will then match `PyString -> OsString -> String`.

I changed my mind on this - it actually would change `PyString::to_string_lossy` to potentially throw exceptions, which would not be great.

Also the inconsistency does not seem to matter; I found someone else who stumbled across similar conversions at https://github.com/rust-lang/rust/issues/56786 - in the end they decided it was better left as-is.

At least the current implementation of `PyString::to_string_lossy` is infallible and exactly consistent with what `std::String::from_utf8_lossy` does.

src/conversions/osstr.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl IntoPy<PyObject> for OsString {
+    fn into_py(self, py: Python) -> PyObject {
+        self.to_object(py)
+    }
+}
+
+impl<'a> IntoPy<PyObject> for &'a OsString {
```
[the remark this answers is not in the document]
kangalioo: types/string.rs has an equivalent implementation. I oriented myself at those trait implementations

src/conversions/path.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+impl IntoPy<PyObject> for PathBuf {
+    fn into_py(self, py: Python) -> PyObject {
+        self.into_os_string().to_object(py)
+    }
+}
+
+impl<'a> IntoPy<PyObject> for &'a PathBuf {
```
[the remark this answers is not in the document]
kangalioo: Same thing; I oriented myself at the trait implementations for String

src/conversions/osstr.rs
```diff
@@ -0,0 +1,203 @@
...
+}
+
+impl IntoPy<PyObject> for &'_ OsStr {
+    #[inline]
+    fn into_py(self, py: Python) -> PyObject {
+        self.to_object(py)
+    }
+}
+
+impl ToPyObject for Cow<'_, OsStr> {
```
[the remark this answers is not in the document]
kangalioo: Haha same thing, seems like `types/string.rs` is missing this as well

src/conversions/osstr.rs
kangalioo: I will apply the suggestion. If you have time, I would be interested to hear why you couldn't just replace `from_borrowed_ptr` with `from_owned_ptr`

src/conversions/osstr.rs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        #[cfg(windows)]
+        {
+            let wstr: Vec<u16> = std::os::windows::ffi::OsStrExt::encode_wide(self).collect();
+
+            unsafe {
+                // This will not panic because the data from encode_wide is well-formed Windows
+                // string data
+                py.from_owned_ptr::<PyString>(ffi::PyUnicode_FromWideChar(
```
davidhewitt, line 46: Just noticed that this can use `PyObject::from_owned_ptr`, which will remove the need for `.into()` also.

```suggestion
                PyObject::from_owned_ptr(py, ffi::PyUnicode_FromWideChar(
```

src/conversions/osstr.rs
```diff
@@ -0,0 +1,212 @@
...
+
+impl FromPyObject<'_> for OsString {
+    fn extract(ob: &PyAny) -> PyResult<Self> {
+        #[cfg(not(windows))]
+        {
+            let pystring = <PyString as PyTryFrom>::try_from(ob)?; // Cast PyAny to PyString
+
+            // Decode from Python's lossless bytes string representation back into raw bytes
+            let fs_encoded_bytes: Py<crate::types::PyBytes> = unsafe {
+                Py::from_owned_ptr(ob.py(), ffi::PyUnicode_EncodeFSDefault(pystring.as_ptr()))
```
davidhewitt, line 67: Using `crate::Py` here may allow you to avoid the OS-specific import.

src/conversions/osstr.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// There's no FromPyObject implementation for &OsStr because albeit possible on Unix, this would
+// be impossible to implement on Windows. Hence it's omitted entirely
+
+impl FromPyObject<'_> for OsString {
+    fn extract(ob: &PyAny) -> PyResult<Self> {
+        #[cfg(not(windows))]
+        {
+            let pystring = <PyString as PyTryFrom>::try_from(ob)?; // Cast PyAny to PyString
```
davidhewitt, line 63: This line is shared between the two implementations so could be pulled out the top before them both.

src/conversions/osstr.rs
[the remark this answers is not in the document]
davidhewitt: Ah - if you're willing, can you add it there also? 🙏 

I think that adding `IntoPy` for `Cow` makes sense (as it's the trait needed to be able to return this type from `#[pyfunction]`).

src/conversions/osstr.rs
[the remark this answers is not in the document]
davidhewitt: Ahh ok. 

I'm not entirely sure why that impl exists... it might be because `#[pyo3(get)]` on `String` fields needs it? Not sure.

If we can't figure out why we need this impl, I'd rather skip it for now. We can always add it later!

src/conversions/osstr.rs
davidhewitt: Sure thing. `from_owned_ptr` returns `&PyAny` (or another native type) where the owned pointer has to be stored by PyO3 in a thread-local vector. In comparison, `Py` (and `PyObject`, aka `Py<PyAny>`) directly hold the owned pointer inside them. This is marginally more efficient, and also means that the temporary `bytes` will be cleaned up immediately, instead of when PyO3 has a chance to cleanup its internal vector safely.

This has been a long thorn in PyO3's API imo - there's discussion at #1056 and #1308 where I hope to eventually remove this difference and make everything as efficient as possible!

davidhewitt (approved):
👍 I think there's some final decisions to be made on the traits to provide, and then this is good to merge. Thanks again!

davidhewitt (approved):
Thank you for the many rounds of iteration - this PR is looking great to me now 🚀

davidhewitt:
I'm going to rebase and merge this. Thanks again!

PR #1380, opened 2021-01-10
Exclude more files from published crates

decathorpe:
Exclude files that are only useful for CI or development, but not in the crates published on crates.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/pyo3/pyo3"
 documentation = "https://docs.rs/crate/pyo3/"
 categories = ["api-bindings", "development-tools::ffi"]
 license = "Apache-2.0"
-exclude = ["/.gitignore", ".cargo/config"]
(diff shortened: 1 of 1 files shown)


davidhewitt:
Thanks. Was the presence of these files causing you some kind of issue?

I wonder if we should go further e.g. the `guide/` subdirectory might not be interesting on crates.io either?

decathorpe:
Those files get automatically installed by the packaging automation we have for RPM packages in Fedora. The primary "problem" is only that they unnecessarily clutter the filesystem - but some files can cause unintended side effects (e.g. pulling in additional script interpreters, or causing hard errors if they have an executable bit set; this can sometimes happen when developers publish crates from a windows machine). I hope that explains our reasoning.

I tried excluding the "guide" directory as well, but it caused the build of "pyo3-macros" to fail when I ran `cargo package`, so I didn't exclude it after all.

davidhewitt (approved):
I see. Sounds good to me - thanks!

decathorpe:
Thanks for merging! :)

PR #1383, opened 2021-01-12
macros-backend: fix raw idents in pymethods

### LINKED ISSUE #1382: Using rust keywords as parameter names
Rust allows you to use keywords as identifiers by prepending `r#`, e.g. a parameter may be named `r#type`. 

When used as parameter in e.g. a constructor, passing e.g. `type="foo"` on Python's side is not accepted by Rust. Example

```python
arch = report.macho.CPUArchitecture(type=type_, subtype=subtype)
TypeError: CPUArchitecture.__new__() missing required positional argument: r#type
```

In Rust:
```rust
#[pymethods]
impl CPUArchitecture {
    #[new]
    pub fn new(_py: Python, r#type: Py<CPUType>, subtype: Py<CPUSubType>) -> PyResult<Self> {
        Ok(Self { r#type, subtype })
    }
}
```

I believe `r#type` should be translated to `type` for Python.

davidhewitt:
Closes #1382 

I added a comprehensive test and then added some fixes in the code for the cases uncovered by the test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - `prepare_freethreaded_python` will no longer register an `atexit` handler to call `Py_Finalize`. [#1355](https://github.com/PyO3/pyo3/pull/1355)
 
+### Fixed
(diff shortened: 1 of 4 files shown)


Lehona:
I ran this against my code (given that I opened the original issue) and it works perfectly.

davidhewitt:
Thanks for the feedback!

PR #1384, opened 2021-01-13
Define _PyCFunctionFastWithKeywords() on CPython 3.7+

ijl:
https://docs.python.org/3/c-api/structures.html#c._PyCFunctionFastWithKeywords

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Add unsafe API `with_embedded_python_interpreter` to initalize a Python interpreter, execute a closure, and finalize the interpreter. [#1355](https://github.com/PyO3/pyo3/pull/1355)
 - Add `serde` feature to support `Serialize/Deserialize` for `Py<T>`. [#1366](https://github.com/PyO3/pyo3/pull/1366)
+- Define `_PyCFunctionFastWithKeywords()` on CPython 3.7+. [#1384](https://github.com/PyO3/pyo3/pull/1384)
 
 ### Changed
 - `prepare_freethreaded_python` will no longer register an `atexit` handler to call `Py_Finalize`. [#1355](https://github.com/PyO3/pyo3/pull/1355)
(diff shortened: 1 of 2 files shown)


ijl:
Regarding the 3.7 cfg, this was introduced in cpython commit 6969eaf4682beb01bc95eeb14f5ce6c01312e297 and first tagged in v3.7.0a1.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Add unsafe API `with_embedded_python_interpreter` to initalize a Python interpreter, execute a closure, and finalize the interpreter. [#1355](https://github.com/PyO3/pyo3/pull/1355)
 - Add `serde` feature to support `Serialize/Deserialize` for `Py<T>`. [#1366](https://github.com/PyO3/pyo3/pull/1366)
+- Define `_PyCFunctionFastWithKeywords()` on CPython 3.7+. [#1384](https://github.com/PyO3/pyo3/pull/1384)
```
davidhewitt, line 12: Just a small nit to make this wording match other ones in the changelog:

```suggestion
- Add FFI definition `_PyCFunctionFastWithKeywords` on Python 3.7 and up. [#1384](https://github.com/PyO3/pyo3/pull/1384)
```

davidhewitt (approved):
Thanks for this!

I've got a tiny nit on the changelog and this is good to merge. Looking at the docs you linked I noticed there's a couple of other mismatches we have:
- we incorrectly use `args: *mut *mut PyObject` in `_PyCFunctionFast` when it should also be `*const *mut PyObject` like you have used here.
- we're also missing `PyCMethod` for Python 3.9 and up.

This PR is good to merge with the suggested changelog entry. However, if you've got a moment and willing to fix the above points at the same time I'd be very grateful! 👍

nw0:
>  we incorrectly use `args: *mut *mut PyObject` in `_PyCFunctionFast` when it should also be `*const *mut PyObject` like you have used here.

This is actually correct for 3.6:

```c
typedef PyObject *(*_PyCFunctionFast) (PyObject *self, PyObject **args,
                                       Py_ssize_t nargs, PyObject *kwnames);
```

src/ffi/methodobject.rs
```diff
@@ -30,6 +30,14 @@ pub type PyCFunctionWithKeywords = unsafe extern "C" fn(
     kwds: *mut PyObject,
 ) -> *mut PyObject;
 
+#[cfg(all(Py_3_7, not(Py_LIMITED_API)))]
```
nw0, line 33: In `methodobject.h`, this is unconditionally defined (i.e. part of the limited API). (It should probably be considered CPython-specific, however)

```suggestion
#[cfg(all(Py_3_7))]
```

davidhewitt:
👍 thanks for the review @nw0 - in which case I'll merge this now and we can leave further corrections for #1289

ijl:
I see you committed the rewording, thanks.

PR #1388, opened 2021-01-15
add some safety notes

nw0:
For #698.

I'm not convinced that all the APIs defined should be unsafe. For example:

https://github.com/PyO3/pyo3/blob/b6f595ba018b3be8e14cd398fb347caf067bd1bf/src/type_object.rs#L23

Implemented in:
https://github.com/PyO3/pyo3/blob/b6f595ba018b3be8e14cd398fb347caf067bd1bf/src/pycell.rs#L69

Since `py_init` takes ownership of `value`, it seems the worst thing that could happen is a memory leak, which is generally considered safe. But do correct me if there are other things that could go wrong here.

Likewise, `PyClassDict::clear_dict` seems safe, as all it does is call `PyDict_Clear` on a dict which is (afaik) managed by Python refcounting and doesn't include anything that we own. So it seems to me that there is some scope to reduce the `unsafe` labeling.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -401,7 +401,7 @@ pub unsafe trait FromPyPointer<'p>: Sized {
     ///
     /// # Safety
     ///
-    /// Relies on unsafe fn `from_owned_ptr_or_opt`.
+    /// Relies on [`from_owned_ptr_or_opt`](#method.from_owned_ptr_or_opt).
     unsafe fn from_owned_ptr_or_panic(py: Python<'p>, ptr: *mut ffi::PyObject) -> &'p Self {
         Self::from_owned_ptr_or_opt(py, ptr).unwrap_or_else(|| err::panic_after_error(py))
     }
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -411,7 +411,7 @@ pub unsafe trait FromPyPointer<'p>: Sized {
     ///
     /// # Safety
     ///
-    /// Relies on unsafe fn `from_owned_ptr_or_opt`.
+    /// Relies on [`from_owned_ptr_or_opt`](#method.from_owned_ptr_or_opt).
     unsafe fn from_owned_ptr(py: Python<'p>, ptr: *mut ffi::PyObject) -> &'p Self {
         Self::from_owned_ptr_or_panic(py, ptr)
     }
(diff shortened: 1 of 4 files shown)


src/types/bytes.rs
```diff
@@ -67,6 +67,10 @@ impl PyBytes {
     /// Creates a new Python byte string object from a raw pointer and length.
     ///
     /// Panics if out of memory.
+    ///
+    /// # Safety
+    ///
+    /// Unsafe as it deferences the raw pointer `ptr`.
```
kngwyu, line 73: Let's make this more explicit. 
https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety can be a good reference.
Or just noting a link to is sufficient.

src/instance.rs
```diff
@@ -356,6 +356,9 @@ impl<T> Py<T> {
     }
 
     /// Deprecated alias for [`from_owned_ptr`](#method.from_owned_ptr).
+    ///
+    /// # Safety
+    /// `ptr` must be a pointer to a Python object of type T.
```
kngwyu, line 361: Also, it should be noted that `ptr` must be owned. We can remove this function in the near future, though.

kngwyu (changes requested):
Thanks!
I think there is some room for improvement.

kngwyu:
> I'm not convinced that all the APIs defined should be unsafe. 

If you are OK, could you please file a separate issue for these 'actually not unsafe' functions?

src/conversion.rs
```diff
@@ -387,24 +391,64 @@ impl IntoPy<Py<PyTuple>> for () {
 
 /// Raw level conversion between `*mut ffi::PyObject` and PyO3 types.
 pub unsafe trait FromPyPointer<'p>: Sized {
+    /// Convert from an arbitrary `PyObject`.
+    ///
+    /// # Safety
+    ///
+    /// Implementations must ensure the object does not get freed during `'p` and avoid type confusion.
```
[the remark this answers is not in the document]
nw0: Updated.

> type confusion

[Basically when the program has a `&T` reference to what is actually a `U` object, and uses that `U` as if it were a `T`.](https://www.microsoft.com/security/blog/2015/06/17/understanding-type-confusion-vulnerabilities-cve-2015-0336/) Anyway I've changed this to "...ensure that `ptr` is of the correct type", which is hopefully clearer.

src/instance.rs
nw0: Updated below (and above for `from_owned_ptr`)!

src/types/bytes.rs
```diff
@@ -67,6 +67,11 @@ impl PyBytes {
     /// Creates a new Python byte string object from a raw pointer and length.
     ///
     /// Panics if out of memory.
+    ///
+    /// # Safety
+    ///
+    /// This function dereferences the raw pointer `ptr`,
```
kngwyu, line 73: I'm sorry for my unclear words, but I meant that 'let's note that `ptr` is a leading pointer of an array (or contiguous memory?) of which length is `len` by referencing https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety

davidhewitt (approved):
👍 this looks great to me, thanks for this!

kngwyu (approved):

kngwyu:
Thanks!

Issue #1274, opened 2020-11-12
Questions about multithreading and asyncio


awestlake87:
Hey, I've got a rather complicated program that involves mixing Rust async/await with Python asyncio. I know this problem isn't sorted out yet in pyo3 (although I think it's being worked on), but I figured I could get around those limitations by running two event loops, one for Rust and one for Python on its own dedicated thread.

So here's what I'm doing:
- Spin up an async Rust application with async-std runtime
- Spawn a thread to act as the main thread for the asyncio event loop and call `asyncio.set_event_loop(loop)` and `loop.run_forever()` on that thread
- Host a websocket server with `async-tungstenite` that occasionally interacts with the Python event loop via pyo3.
- Interact with the Python event loop via `call_soon_threadsafe` and `asyncio.run_coroutine_threadsafe` to run my application
- When the application exits, I call `loop.call_soon_threadsafe(loop.stop)` and join my thread

My tests run exactly as expected with pyo3. The only problem is that when the test exits, I get one of three errors:

If I'm lucky, my test exits successfully this warning:
```
Exception ignored in: <module 'threading' from '/usr/lib/python3.8/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 1373, in _shutdown
    assert tlock.locked()
AssertionError: 
```

If I'm unlucky I either get this error:
```
Fatal Python error: This thread state must be current when releasing

Thread 0x00007f7dd1dd8b40 (most recent call first):

Current thread 0x00007f7dcece9700 (most recent call first):
  File "/usr/lib/python3.6/asyncio/selector_events.py", line 148 in _write_to_self
  File "/usr/lib/python3.6/asyncio/base_events.py", line 643 in call_soon_threadsafe

Thread 0x00007f7dcc4a0700 (most recent call first):
  File "/usr/lib/python3.6/selectors.py", line 445 in select
  File "/usr/lib/python3.6/asyncio/base_events.py", line 1415 in _run_once
  File "/usr/lib/python3.6/asyncio/base_events.py", line 438 in run_forever

Thread 0x00007f7dce8e7700 (most recent call first):
error: test failed
```
or this one:
```
FATAL: exception not rethrown
error: test failed
```

Things I've tried:
- Spawning native rust threads with `std::thread` to run the event loop
- Spawning python threads with the `threading` module to run my event loop
- Calling `pyo3::prepare_freethreaded_python()` before the first call to `Python::with_gil`

[...]

PR #1393, opened 2021-01-19
Added async/await example to the guide

awestlake87:
Here's the async/await example for the guide as discussed in #1274. The example code is actually based on the `other-runtimes` branch instead of the `master` branch since I expect to merge that branch soon.

Let me know if you have any feedback on the guide / examples / API (since it changed a bit with the `other-runtimes` branch to support other runtimes)

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ about this topic.
  * [pyre](https://github.com/Project-Dream-Weaver/Pyre) _Fast Python HTTP server written in Rust_
  * [jsonschema-rs](https://github.com/Stranger6667/jsonschema-rs/tree/master/bindings/python) _Fast JSON Schema validation library_
  * [css-inline](https://github.com/Stranger6667/css-inline/tree/master/bindings/python) _CSS inlining for Python implemented in Rust_
+ * [cryptography](https://github.com/pyca/cryptography/tree/master/src/rust) _Python cryptography lirbary with some functionality in Rust_
 
 ## License
 


kngwyu:
Thank you for this hard work!

guide/src/async-await.md
```diff
@@ -0,0 +1,242 @@
...
+## Awaiting an Async Python Function in Rust
+
+Let's take a look at a dead simple async Python function:
+
+```python
+async def py_sleep():
+    await asyncio.sleep(1)
+```
+
+This function simply sleeps for 1 second and returns. So what does this look like to PyO3?
```
kngwyu, line 20: > So what does this look like to PyO3? 

Isn't it better to first describe async function in Python, and then throw this question?

guide/src/async-await.md
```diff
@@ -0,0 +1,242 @@
...
+for our purposes, we really don't need to know much about them. The key factor here is that calling
+an `async` function is _just like calling a regular function_, the only difference is that we have
+to do something special with the object that it returns.
+
+Normally in Python, that something special is the `await` keyword, but in Rust, we don't have the 
+luxury of using Python's syntax. Luckily, Rust also has an `await` keyword that does something
+similar, we just need to find a way of converting a `&PyAny` into a Rust future so we can use the
+`await` keyword on it.
+
+That's where pyo3-asyncio comes in. `pyo3_asyncio::into_future` performs this conversion for us:
```
kngwyu, line 34: How about describing *a bit* deeper about what the function does for conversion?

guide/src/async-await.md
```diff
@@ -0,0 +1,242 @@
...
+## PyO3 Asyncio in Cargo Tests
+
+The default Cargo Test harness does not currently allow test crates to provide their own main 
+function, so there doesn't seem to be a good way to allow Python to gain control over the main
+thread.
+
+We can, however, override the default test harness and provide our own. `pyo3-asyncio` provides some
+utilities to help us do just that!
+
+### Creating A PyO3 Asyncio Integration Test
```
kngwyu, line 155: How about flattening this subsection?

kngwyu (approved):

guide/src/async-await.md
awestlake87: I was thinking the details might be a bit convoluted in the guide. What do you think about linking to the API docs for this function to provide the implementation details instead of putting them here inline?

Currently the API docs don't go over the nitty gritty details for this function, so I'll have to expand that a little bit. It'd probably read something like this:

> This function converts the `coroutine` into a Python `Task` using `run_coroutine_threadsafe`. A completion handler sends the result of this `Task` through a `futures::channel::oneshot::Sender<PyResult<PyObject>>` and the future returned by this function simply awaits the result through the `futures::channel::oneshot::Receiver<PyResult<PyObject>>`.

guide/src/async-await.md
awestlake87: Yeah, I agree. The subsections here get a bit too deeply nested and `Creating a PyO3 Asyncio Integration Test` is kinda redundant.

guide/src/async-await.md
awestlake87: My intention here was to somehow lead into the explanation of the async function, so maybe I should word it a bit differently.

The point that I was trying to make in this section is that even though async functions and normal functions are different, calling them via PyO3 is the same, you just have to treat the return value differently. I didn't want to dive too far into the specifics for awaitables and coroutines, just enough to explain the how/why for the `pyo3_asyncio::into_future` function.

Any tips on how I should reword this section? Should I just skip this line and go straight into the explanation in the next few paragraphs?

guide/src/async-await.md
kngwyu: > What do you think about linking to the API docs for this function to provide the implementation details instead of putting them here inline?

Sounds nice :+1:

guide/src/async-await.md
kngwyu: > The point that I was trying to make in this section is that even though async functions and normal functions are different, calling them via PyO3 is the same, you just have to treat the return value differently. 

That makes sense, thanks.

> Should I just skip this line and go straight into the explanation in the next few paragraphs?

I don't think we should remove the whole paragraph, but packing the line into the next paragraph would make the document easier to follow.
E.g.,
```
This function simply sleeps for 1 second and returns a `coroutine` object. ...
```

guide/src/async-await.md
```diff
@@ -0,0 +1,240 @@
...
+Async functions in python are simply functions that return a `coroutine` object. You can read more about 
+`coroutine` objects in the [Python 3 docs](https://docs.python.org/3/library/asyncio-task.html), but
+for our purposes, we really don't need to know much about them. The key factor here is that calling
+an `async` function is _just like calling a regular function_, the only difference is that we have
+to do something special with the object that it returns.
+
+Normally in Python, that something special is the `await` keyword, but in Rust, we don't have the 
+luxury of using Python's syntax. Luckily, Rust also has an `await` keyword that does something
+similar, we just need to find a way of converting a `&PyAny` into a Rust future so we can use the
+`await` keyword on it.
```
kngwyu, line 32: I feel the whole paragraph a little bit redundant. The key point is we need `pyo3_asyncio::into_future` to make the Python coroutine `await`able in Rust, so we might want to note that earlier than pointing out that both languages have the same keyword `await`.

kngwyu:
@davidhewitt 
Could you please take a glance at the document?

davidhewitt:
👍 sorry have been quite busy this week; intending to review later today.

guide/src/async-await.md
awestlake87: Yeah, I see what you mean. How about condensing it down to this?

> Normally in Python, that something special is the `await` keyword, but in order to await this coroutine in Rust, we first need to convert it into a `Future`. 

It could probably be joined up with line 34 too.

guide/src/async-await.md
awestlake87: K, I'll probably have it point to the docs.rs version, so getting a proper link might have to wait until it's published to crates.io. That should happen before this PR is merged anyway I think.

guide/src/async-await.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+## PyO3 Asyncio in Cargo Tests
+
+The default Cargo Test harness does not currently allow test crates to provide their own main 
+function, so there doesn't seem to be a good way to allow Python to gain control over the main
+thread.
+
+We can, however, override the default test harness and provide our own. `pyo3-asyncio` provides some
+utilities to help us do just that!
```
davidhewitt, line 153: I would add a quick note here before going into the subsections, saying what we'll need to do (e.g. `Cargo.toml`, custom `pytests` directory, different way to write tests.

davidhewitt (commented):
🚀 

The more I read this, the more my mind is blown by how awesome this is. I've got some personal projects which use a mix of Rust + Python; at the moment I've got a jungle of threads and blocking code because I haven't had time to think about async for PyO3.

Many thanks to @ChillFish8 and @awestlake87 - what you've been figuring out here is really awesome.

On topic of the review - I think this guide entry is a great start and has all the right content. I made a couple of suggestions to improve the flow of the document. I see there's also some outstanding suggestions from @kngwyu.

Once they're all addressed we'll probably be much closer to a final copy; I'd like to give it another read through then.

guide/src/SUMMARY.md
```diff
@@ -15,6 +15,7 @@
 - [Calling Python from Rust](python_from_rust.md)
 - [GIL, mutability and object types](types.md)
 - [Parallelism](parallelism.md)
+- [Async / Await](async-await.md)
```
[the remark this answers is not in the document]
awestlake87: Ok, should I go ahead and do that here or do you want to do that as part of #1401?

guide/src/async-await.md
```diff
@@ -0,0 +1,240 @@
...
+let future = rust_sleep();
+```
+
+We can convert this future object into Python to make it "Awaitable". This tells Python that you can
+use the `await` keyword with it. In order to do this, we'll call 
+`pyo3_asyncio::async_std::into_coroutine`:
+
+```rust
+#[pyfunction]
+fn call_rust_sleep(py: Python) -> PyResult<PyObject> {
```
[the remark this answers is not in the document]
awestlake87: The return type here would actually be `asyncio.Future` here since the event loop waits on us to run the future on the Rust event loop and call `asyncio.Future.set_result`.

`into_coroutine` is a bit of a misnomer, but I wasn't really sure what to call it honestly. `into_future` is already taken, maybe `into_awaitable` would be a better name?

guide/src/async-await.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+Similar to Python, Rust's async functions also return a special object called a
+`Future`:
+
+```rust
+let future = rust_sleep();
+```
+
+We can convert this future object into Python to make it "Awaitable". This tells Python that you can
```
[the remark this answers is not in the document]
awestlake87: I was thinking about using `code quotes` here since the [Python 3 docs](https://docs.python.org/3/library/asyncio-task.html) seem to highlight the word `awaitable` the same as `coroutine`.

guide/src/async-await.md
awestlake87: I added the future docs.rs links to the guide for many of the pyo3-asyncio functions. The only difference is that I noticed `-` and `_` are different between the crate name and the folder name in the docs so the actual links should be something like this: `https://docs.rs/pyo3-asyncio/latest/pyo3_asyncio/fn.into_future.html`. I discovered this by poking around with the async-std docs.rs links.

guide/src/async-await.md
```diff
@@ -0,0 +1,240 @@
...
+pyo3-asyncio, we decided the best way to handle Rust/Python interop was to just surrender the main
+thread to Python and run Rust's event loops in the background. Unfortunately, since most event loop 
+implementations _prefer_ control over the main thread, this can still make some things awkward.
+
+### PyO3 Asyncio Initialization
+
+Because Python needs to control the main thread, we can't use the convenient proc macros from Rust
+runtimes to handle the main function or test functions. 
+
+Instead, the initialization for PyO3 has to be done manually from the main function and the main
```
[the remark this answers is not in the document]
awestlake87: I haven't been following the conversation on the `auto-initialize`, so I'm not sure what I should add here. AFAIK there are no problems with the current implementation using the `auto-initialize` feature. Are you suggesting that we add an example of what someone should do if they want to disable `auto-initialize`?

Signal handling _appears_ to work properly in my project that uses pyo3-asyncio, so that difference between `Py_InitializeEx(0)` and `Py_InitializeEx(1)` kinda surprises me. Maybe that's something we should look at a bit further.

awestlake87:
Alright, I think I've addressed (or at least responded to) all of your comments and added some of my own minor changes. Let me know what you guys think!

awestlake87:
Also, I created a PR to create proc macros like `#[pyo3_asyncio::async_std::main]` to simplify the initialization. If we were to merge [PR #10](https://github.com/awestlake87/pyo3-asyncio/pull/10) in `pyo3-asyncio` we could condense the initialization example down to this:

```rust
#[pyo3_asyncio::async_std::main]
async fn main() -> PyResult<()> {
    let fut = Python::with_gil(|py| {
        let asyncio = py.import("asyncio")?;
        // convert asyncio.sleep into a Rust Future
        pyo3_asyncio::into_future(asyncio.call_method1("sleep", (1.into_py(py),))?)
    })?;
    fut.await?;
    Ok(())
}
```

awestlake87:
I also took a look at the `#[test]` proc macros, and it turns out they were a bit easier than I thought they would be. Tests can be simplified to this now:

```rust
use std::{time::Duration, thread};

use pyo3::prelude::*;

#[pyo3_asyncio::tokio::test]
async fn test_async_sleep() -> PyResult<()> {
    tokio::time::sleep(Duration::from_secs(1)).await;
    Ok(())
}

#[pyo3_asyncio::tokio::test]
fn test_blocking_sleep() -> PyResult<()> {
    thread::sleep(Duration::from_secs(1));
    Ok(())
}

pyo3_asyncio::testing::test_main!(#[pyo3_asyncio::tokio::main], "Example Test Suite");
```

Although many of the caveats still apply as I discussed [here](https://github.com/awestlake87/pyo3-asyncio/issues/9). I think it's a cleaner solution, but it is a breaking change, so idk if it should be included in the `0.13` release.

guide/src/async-await.md
```diff
@@ -0,0 +1,268 @@
+# Async / Await
+
+If you are working with a Python library that makes use of async functions or wish to provide
```
kngwyu, line 3: Nice introduction 👍🏼

kngwyu:
I think now the contents are fine. Could you please rebase on #1401?

> I think it's a cleaner solution, but it is a breaking change. On one hand it'd be nice to merge before the release, but on the other it'll involve more changes to the guide and the design hasn't been reviewed by anyone else.

Though I do not fully understand the macros, they look better especially for newcomers to Rust 😃 
Also, I'm happy to review the PR (though it'll take some time...

davidhewitt:
I think something's gone weird in the rebase so I'm going to force-push...

davidhewitt:
😅 well I force-pushed incorrectly and now github's locked me out of this PR, so I've had to re-open it at #1415. Sorry!

PR #1397, opened 2021-01-20
Make PyTimeAccess::get_fold() return bool instead of u8

kangalioo:
The fold property of a Python datetime stores whether the datetime lies on a leap second or not. The only allowed values are 0 and 1 ([source](https://docs.python.org/3/library/datetime.html#datetime.datetime.fold)). Therefore, the [`PyTimeAccess::get_fold()`](https://docs.rs/pyo3/0.13.1/pyo3/types/trait.PyTimeAccess.html#tymethod.get_fold) method should expose this property as a `bool`.

Coincidentally, the [`PyTime::new_with_fold()`](https://docs.rs/pyo3/0.13.1/pyo3/types/struct.PyTime.html) already uses bool as the type of fold.

In a later PR, I will add the currently missing `PyDateTime::new_with_fold()` method, which is also going to accept bool for the fold parameter.

This is a breaking change

### DIFF
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -58,7 +58,7 @@ pub trait PyTimeAccess {
     fn get_second(&self) -> u8;
     fn get_microsecond(&self) -> u32;
     #[cfg(not(PyPy))]
-    fn get_fold(&self) -> u8;
+    fn get_fold(&self) -> bool;
 }
 
 /// Bindings around `datetime.date`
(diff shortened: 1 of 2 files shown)


src/types/datetime.rs
```diff
@@ -224,8 +224,10 @@ impl PyTimeAccess for PyDateTime {
     }
 
     #[cfg(not(PyPy))]
-    fn get_fold(&self) -> u8 {
-        unsafe { PyDateTime_DATE_GET_FOLD(self.as_ptr()) as u8 }
+    fn get_fold(&self) -> bool {
+        // TODO: Python's fold values must be either 0 or 1 as per Python documentation. Should we
+        // panic if they do, for whatever reason, fall outside that range?
```
kngwyu, line 229: I don't think we should cause panic here.

kngwyu (approved):
Thanks, sounds reasonable to me.
> This is a breaking change

Then let's add a CHANGELOG entry

davidhewitt:
As this is another breaking change, I'm thinking it's time to release 0.13.2 before we then start merging all changes for 0.14 . Will put a PR for this out tonight.

davidhewitt:
I'm rebasing and merging this shortly. Thanks again!

PR #1398, opened 2021-01-21
Add PyDateTime::new_with_fold() method

kangalioo:
`PyTime` already has a new_with_fold() method; `PyDateTime` needs one for consistency and, more importantly, to be able to handle leap seconds correctly.


davidhewitt:
👍 thanks very much for this! While adding this function, would you be willing to add an accompanying test please?

(I know the other methods in this file aren't tested; that's an accident of history. 😄 )

davidhewitt (approved):
Thanks very much for this!

kangalioo:
Thank you for all the kind feedback you keep giving! :)

PR #1401, opened 2021-01-24
guide: organise more chapters into sublevels

davidhewitt:
With #1393 I thought that a new section of the guide would be appropriate to house that chapter.


davidhewitt:
I pushed a tweak to the `doctest!` macro which makes it easier (imo) to ensure all the guide files are tested.

src/lib.rs
```diff
@@ -359,42 +359,46 @@ macro_rules! py_run_impl {
...
     macro_rules! doctest {
-        ($x:expr, $y:ident) => {
-            doc_comment!(include_str!($x), mod $y {});
+        ($($path:ident)/+.md) => {
+            $crate::paste::paste!(
+                doc_comment!(
+                    include_str!(concat!("..", $("/", stringify!($path)),+, ".md")),
+                    mod [<$($path)_+ _md>] {}
+                );
+            );
```
kngwyu, line 379: Why do we need this?
I prefer to keep this simple for future contributors.

kngwyu (approved):
Thanks!

src/lib.rs
kngwyu: Also, why is `paste!` necessary here?

davidhewitt: I was using some trickery with `paste!` to generate the module names from the path - paste is used to join the idents into one.

I thought it would be nice to simplify usage of the macros. TBH I was 50/50 if this was a nice result in the end, so I'll revert this change.

kngwyu:
Thanks!

PR #1402, opened 2021-01-25
Use fully-qualified Result in pyobject_native_type_base

maffoo:
This allows using macros such as `create_exception!` in modules where `Result` has been aliased, for example to a single-parameter type with a module-specific error type. And in any event, seems like good macro hygiene to use fully-qualified names wherever possible.


davidhewitt (approved):
Thanks for the fix, much appreciated!

kngwyu:
Nice catch 👍🏼

PR #1404, opened 2021-01-26
Remove `unsafe` for some safe functions

### LINKED ISSUE #1389: Some safe functions are marked `unsafe`?
When I was looking at the remaining unsafe functions without safety notes for #698, I noticed that a couple of functions didn't seem to be `unsafe` at all.

This issue is about whether these functions should be safe to call (i.e. no need to be marked `unsafe`) because they wrap up the unsafety, or aren't actually unsafe.

1. `PyLayout::py_init`. This function takes ownership of a `value: T`, which could be leaked in the worst case, but leaking memory is considered safe Rust. For reference, both implementations don't use any `unsafe` code.
https://github.com/PyO3/pyo3/blob/b6f595ba018b3be8e14cd398fb347caf067bd1bf/src/type_object.rs#L23

1. `PyClassDict::clear_dict`. In this case the implementation calls the FFI function `PyDict_Clear`, which seems to be safe as it operates solely on the refcounted contents of the class dict.
https://github.com/PyO3/pyo3/blob/3de51d59df906e4db44bd54c27323b21cc17cb06/src/pyclass_slots.rs#L9

1. `PyClassAlloc::new`. This feels like implementations should encapsulate any unsafety, but I'm not sure.
https://github.com/PyO3/pyo3/blob/3de51d59df906e4db44bd54c27323b21cc17cb06/src/pyclass.rs#L75

Opening this issue as suggested in https://github.com/PyO3/pyo3/pull/1388#issuecomment-761519711, even if it's not obvious that we want to change the `unsafe` marking on any of the functions. I realise that most of these APIs are probably only used internally.

nw0:
Closes #1389, which noted that some safe functions are marked `unsafe`.

Noted that this is a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Mark FFI definitions `PyMarshal_WriteObjectToString`, `PyMarshal_ReadObjectFromString` as available in limited API.
 - Mark FFI definitions `PyListObject` and those from `funcobject.h` as requiring non-limited API.  [#1387](https://github.com/PyO3/pyo3/pull/1387)
 - Fix typo in FFI definition `PyFunction_Code` to `PyFunction_GetCode`. [#1387](https://github.com/PyO3/pyo3/pull/1387)
+- Mark `PyLayout::py_init`, `PyClassDict::clear_dict`, and `opt_to_pyobj` safe, as they do not perform any unsafe operations. [#1404](https://github.com/PyO3/pyo3/pull/1404)
 
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


programmerjake:
iirc, removing `unsafe` shouldn't actually be a breaking change. all functions can be called in an `unsafe` block and also can be coerced to `unsafe` function pointers, so the worst that can happen is code gets a warning that it has a unnecessary `unsafe` block.

davidhewitt (approved):
I did say in #1389 that this would be breaking, but actually I think @programmerjake is correct. It's the same level of breaking as if we added `#[deprecated]` - it won't break builds automatically, only for those who `#[deny]` certain warnings.

Thanks for following up the discussion with this PR!

Issue #1239, opened 2020-10-15
Add #[pyo3(from_python_with = ...)] attribute


davidhewitt:
I've been thinking for a while it would be interesting to have an attribute which can be used to customize the from-python conversion in our proc macros. This need has come up in Gitter as well as #884.

The attribute will go on `#[pyfunction]` arguments or `#[derive(FromPyObject)]` fields. It takes a value which is a path to a function `fn (&PyAny) -> Result<T, E>`, where `T` is the annotated type and `E: Into<PyErr>`. For examples see the detailed design below.

This issue is invite users to comment on the design and for anyone interested to volunteer to implement it (before I eventually get around to it).

In detail, the two applications I see for this attribute are:

#### In `#[pyfunction]` arguments

Consider the following pyfunction:

```
#[pyfunction]
fn foo(datetime: chrono::DateTime) { ... }
```

Currently `chrono::DateTime` does not implement `FromPyObject`, so this function won't compile. An existing workaround is to create a newtype `MyDateTime(chrono::DateTime)` and implement `FromPyObject` for that, but it's a lot of boilerplate.

The attribute would allow the user to leverage a single function without adding extra types:

```
fn extract_datetime(any: &PyAny) -> PyResult<chrono::DateTime> { ... }

#[pyfunction]
fn foo(
    #[pyo3(from_python_with = extract_datetime)]
    datetime: chrono::DateTime
) { ... }
```

#### On `#[derive(FromPyObject)]` fields

Similarly, this applies in a `#[derive(FromPyObject)]` struct:

```
#[derive(FromPyObject)]
#[pyo3(transparent)]
struct Foo {
    #[pyo3(from_python_with = extract_datetime)]
    datetime: chrono::DateTime
}
```

The cool thing here is that this lets us add `#[derive(FromPyObject)]` to a struct which has a field which itself doesn't implement `FromPyObject`.

#### Bikeshedding

- should it be `from_py_with` instead of `from_python_with`?
- should the function be allowed to take more specific argument types than `&PyAny`? E.g. in the examples above it could be helpful to take `&PyDateTime`. But it's also trivial to convert `PyAny` with `.downcast()`.

davidhewitt:
cc @gilescope

daniil-konovalenko:
Hi, I would like to try to implement this, if it's okay.
I haven't touched procedural macros yet though so I think it may take a while

gilescope:
[...]

PR #1411, opened 2021-02-01
Add from_py_with attribute

daniil-konovalenko:
This PR addresses #1239. 

It adds a  `#[pyo3(from_py_with = "function")]` attribute that allows to customize conversion between Python-native and Rust-native types.  
At the moment only argument attributes are supported. I plan on adding support for derive in the near future.

I've made a couple of design decisions along the way, that can be changed if needed:
- I chose `from_py_with` instead of `from_python_with` from the original proposal because it was shorter to type. Also it goes well with the trait name `FromPyObject`.
- The path to the conversion function is quoted instead of being a raw identifier, mainly because [serde does it](https://serde.rs/field-attrs.html#serialize_with). If there are any reasons it shouldn't be quoted, please let me know.

There's still a bunch of stuff left to do:
- [x] support `from_py_with` in `derive(FromPyObject)`
- [x] refactor `parse_arg_attributes` into smaller functions
- [x] add tests with usage in #[pymethods]
- [ ] support `from_py_with` for enums/tuple structs
- [ ] add tests for enums/tuple structs
- [x] add tests for errors
- [x] add a CHANGELOG entry
- [x] add docs (tracking issue #1432)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Added
 - Add conversions between `OsStr`/`OsString`/`Path`/`PathBuf` and Python strings. [#1379](https://github.com/PyO3/pyo3/pull/1379)
+- Add #[pyo3(from_py_with = "...")]` attribute for function arguments and struct fields to override the default from-Python conversion. [#1411](https://github.com/PyO3/pyo3/pull/1411)
 - Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
 
 ### Changed
--- a/pyo3-macros-backend/src/attrs.rs
+++ b/pyo3-macros-backend/src/attrs.rs
@@ -1,0 +1,22 @@
+use syn::spanned::Spanned;
+use syn::{ExprPath, Lit, Meta, MetaNameValue, Result};
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct FromPyWithAttribute(pub ExprPath);
+
+impl FromPyWithAttribute {
+    pub fn from_meta(meta: Meta) -> Result<Self> {
+        let string_literal = match meta {
+            Meta::NameValue(MetaNameValue {
+                lit: Lit::Str(string_literal),
+                ..
+            }) => string_literal,
+            meta => {
+                bail_spanned!(meta.span() => "expected a name-value: `pyo3(from_py_with = \"func\")`")
+            }
+        };
+
+        let expr_path = string_literal.parse::<ExprPath>()?;
+        Ok(FromPyWithAttribute(expr_path))
+    }
+}
--- a/pyo3-macros-backend/src/from_pyobject.rs
+++ b/pyo3-macros-backend/src/from_pyobject.rs
@@ -1,3 +1,4 @@
+use crate::attrs::FromPyWithAttribute;
 use proc_macro2::TokenStream;
 use quote::quote;
 use syn::punctuated::Punctuated;
@@ -85,7 +86,7 @@ enum ContainerType<'a> {
     /// Struct Container, e.g. `struct Foo { a: String }`
     ///
     /// Variant contains the list of field identifiers and the corresponding extraction call.
-    Struct(Vec<(&'a Ident, FieldAttribute)>),
+    Struct(Vec<(&'a Ident, FieldAttributes)>),
     /// Newtype struct container, e.g. `#[transparent] struct Foo { a: String }`
     ///
     /// The field specified by the identifier is extracted directly from the object.
@@ -156,9 +157,8 @@ impl<'a> Container<'a> {
                         .ident
                         .as_ref()
                         .expect("Named fields should have identifiers");
-                    let attr = FieldAttribute::parse_attrs(&field.attrs)?
-                        .unwrap_or(FieldAttribute::GetAttr(None));
-                    fields.push((ident, attr))
+                    let attrs = FieldAttributes::parse_attrs(&field.attrs)?;
+                    fields.push((ident, attrs))
                 }
                 ContainerType::Struct(fields)
             }
@@ -235,17 +235,24 @@ impl<'a> Container<'a> {
         )
     }
 
-    fn build_struct(&self, tups: &[(&Ident, FieldAttribute)]) -> TokenStream {
+    fn build_struct(&self, tups: &[(&Ident, FieldAttributes)]) -> TokenStream {
         let self_ty = &self.path;
         let mut fields: Punctuated<TokenStream, syn::Token![,]> = Punctuated::new();
-        for (ident, attr) in tups {
-            let ext_fn = match attr {
-                FieldAttribute::GetAttr(Some(name)) => quote!(getattr(#name)),
-                FieldAttribute::GetAttr(None) => quote!(getattr(stringify!(#ident))),
-                FieldAttribute::GetItem(Some(key)) => quote!(get_item(#key)),
-                FieldAttribute::GetItem(None) => quote!(get_item(stringify!(#ident))),
+        for (ident, attrs) in tups {
+            let getter = match &attrs.getter {
+                FieldGetter::GetAttr(Some(name)) => quote!(getattr(#name)),
+                FieldGetter::GetAttr(None) => quote!(getattr(stringify!(#ident))),
+                FieldGetter::GetItem(Some(key)) => quote!(get_item(#key)),
+                FieldGetter::GetItem(None) => quote!(get_item(stringify!(#ident))),
             };
-            fields.push(quote!(#ident: obj.#ext_fn?.extract()?));
+
+            let get_field = quote!(obj.#getter?);
+            let extractor = match &attrs.from_py_with {
+                None => quote!(#get_field.extract()?),
+                Some(FromPyWithAttribute(expr_path)) => quote! (#expr_path(#get_field)?),
+            };
+
+            fields.push(quote!(#ident: #extractor));
         }
         quote!(Ok(#self_ty{#fields}))
     }
@@ -309,40 +316,59 @@ impl ContainerAttribute {
 
 /// Attributes for deriving FromPyObject scoped on fields.
 #[derive(Clone, Debug)]
-enum FieldAttribute {
+struct FieldAttributes {
+    getter: FieldGetter,
+    from_py_with: Option<FromPyWithAttribute>,
+}
+
+#[derive(Clone, Debug)]
+enum FieldGetter {
     GetItem(Option<syn::Lit>),
     GetAttr(Option<syn::LitStr>),
 }
 
-impl FieldAttribute {
-    /// Extract the field attribute.
+impl FieldAttributes {
+    /// Extract the field attributes.
     ///
-    /// Currently fails if more than 1 attribute is passed in `pyo3`
-    fn parse_attrs(attrs: &[Attribute]) -> Result<Option<Self>> {
+    fn parse_attrs(attrs: &[Attribute]) -> Result<Self> {
+        let mut getter = None;
+        let mut from_py_with = None;
+
         let list = get_pyo3_meta_list(attrs)?;
-        let metaitem = match list.nested.len() {
-            0 => return Ok(None),
-            1 => list.nested.into_iter().next().unwrap(),
-            _ => bail_spanned!(
-                list.nested.span() =>
-                "only one of `attribute` or `item` can be provided"
-            ),
-        };
-        let meta = match metaitem {
-            syn::NestedMeta::Meta(meta) => meta,
-            syn::NestedMeta::Lit(lit) => bail_spanned!(
-                lit.span() =>
-                "expected `attribute` or `item`, got a literal"
-            ),
-        };
-        let path = meta.path();
-        if path.is_ident("attribute") {
-            Ok(Some(FieldAttribute::GetAttr(Self::attribute_arg(meta)?)))
-        } else if path.is_ident("item") {
-            Ok(Some(FieldAttribute::GetItem(Self::item_arg(meta)?)))
-        } else {
-            bail_spanned!(meta.span() => "expected `attribute` or `item`");
+
+        for meta_item in list.nested {
+            let meta = match meta_item {
+                syn::NestedMeta::Meta(meta) => meta,
+                syn::NestedMeta::Lit(lit) => bail_spanned!(
+                    lit.span() =>
+                    "expected `attribute`, `item` or `from_py_with`, got a literal"
+                ),
+            };
+            let path = meta.path();
+
+            if path.is_ident("attribute") {
+                ensure_spanned!(
+                    getter.is_none(),
+                    meta.span() => "only one of `attribute` or `item` can be provided"
+                );
+                getter = Some(FieldGetter::GetAttr(Self::attribute_arg(meta)?))
+            } else if path.is_ident("item") {
+                ensure_spanned!(
+                    getter.is_none(),
+                    meta.span() => "only one of `attribute` or `item` can be provided"
+                );
+                getter = Some(FieldGetter::GetItem(Self::item_arg(meta)?))
+            } else if path.is_ident("from_py_with") {
+                from_py_with = Some(Self::from_py_with_arg(meta)?)
+            } else {
+                bail_spanned!(meta.span() => "expected `attribute`, `item` or `from_py_with`")
+            };
         }
+
+        Ok(FieldAttributes {
+            getter: getter.unwrap_or(FieldGetter::GetAttr(None)),
+            from_py_with,
+        })
     }
 
     fn attribute_arg(meta: Meta) -> syn::Result<Option<syn::LitStr>> {
@@ -389,6 +415,10 @@ impl FieldAttribute {
 
         bail_spanned!(arg_list.span() => "expected a single literal argument");
     }
+
+    fn from_py_with_arg(meta: Meta) -> syn::Result<FromPyWithAttribute> {
+        FromPyWithAttribute::from_meta(meta)
+    }
 }
 
 /// Extract pyo3 metalist, flattens multiple lists into a single one.
@@ -426,7 +456,7 @@ fn verify_and_get_lifetime(generics: &syn::Generics) -> Result<Option<&syn::Life
 ///   * Max 1 lifetime specifier, will be tied to `FromPyObject`'s specifier
 ///   * At least one field, in case of `#[transparent]`, exactly one field
 ///   * At least one variant for enums.
-///   * Fields of input structs and enums must implement `FromPyObject`
+///   * Fields of input structs and enums must implement `FromPyObject` or be annotated with `from_py_with`
 ///   * Derivation for structs with generic fields like `struct<T> Foo(T)`
 ///     adds `T: FromPyObject` on the derived implementation.
 pub fn build_derive_from_pyobject(tokens: &DeriveInput) -> Result<TokenStream> {
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -1,11 +1,12 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
 use crate::pyfunction::Argument;
-use crate::pyfunction::{parse_name_attribute, PyFunctionAttr};
+use crate::pyfunction::{parse_name_attribute, PyFunctionArgAttrs, PyFunctionAttr};
 use crate::utils;
 use proc_macro2::TokenStream;
 use quote::ToTokens;
 use quote::{quote, quote_spanned};
+use std::ops::Deref;
 use syn::ext::IdentExt;
 use syn::spanned::Spanned;
 
@@ -17,6 +18,7 @@ pub struct FnArg<'a> {
     pub ty: &'a syn::Type,
     pub optional: Option<&'a syn::Type>,
     pub py: bool,
+    pub attrs: PyFunctionArgAttrs,
 }
 
 #[derive(Clone, PartialEq, Debug, Copy, Eq)]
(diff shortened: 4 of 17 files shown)


nw0:
Just bikeshedding: would you consider 'from_py'? It's even more succinct and I think "with" is pretty clear once you add the equals sign.

davidhewitt:
Thanks very much for this PR. I'm very excited to see it!

I'm very busy at the moment though hope to get a moment to read it by the weekend and share my full thoughts and feedback then.

pyo3-macros-backend/src/pyfunction.rs
```diff
@@ -183,9 +188,50 @@ pub fn parse_name_attribute(attrs: &mut Vec<syn::Attribute>) -> syn::Result<Opti
...
+                }
+            } else {
+                new_attrs.push(attr.clone())
+            }
+        } else {
+            new_attrs.push(attr.clone())
+        }
+    }
+    cap.attrs.clear();
+    cap.attrs.extend(new_attrs);
```
kngwyu, line 231: Why do we need to remove `from_py_with`?
If you do so because another function raises an error for that, I think there would be a better way.

pyo3-macros-backend/src/pyfunction.rs
```diff
@@ -27,6 +27,11 @@ pub struct PyFunctionAttr {
     pub pass_module: bool,
 }
 
+#[derive(Clone, PartialEq, Debug)]
+pub struct PyFunctionArgAttr {
+    pub from_py_with: Option<syn::ExprPath>,
+}
```
kngwyu, line 33: I think just `Option<ExprPath>` is not so bad in readability, though either is fine.

kngwyu (commented):
Thanks!
I think the API is fine, but there's some room for improvements in the implementation (can we reduce `mut`)?

pyo3-macros-backend/src/pyfunction.rs
daniil-konovalenko: We have to remove the attributes, because we can't yet specify helper attributes in `#[proc_macro_attribute]`
https://github.com/rust-lang/rust/issues/65823

If we don't remove the attributes, the compiler doesn't recognize it.
```
error: cannot find attribute `pyo3` in this scope
   --> tests/test_pyfunction.rs:124:7
    |
124 |     #[pyo3(from_py_with = "datetime_to_timestamp")] timestamp: i64,
    |       ^^^^
```

The same approach is used for `#[pyo3(get, set)]` attributes inside `#[pyclass]`: https://github.com/PyO3/pyo3/blob/88872eba4fbe895a0728e86505caef8c85a98b07/pyo3-macros-backend/src/pyclass.rs#L184-L211

This solution forces us to make function signature mutable, and it's not very nice, so please let me know if there's another way

daniil-konovalenko: I guess we could do it in two passes:
1. Parse all attributes from an immutable reference to a TokenStream
2. Obtain a mutable reference and remove all pyo3 attributes

pyo3-macros-backend/src/pyfunction.rs
```diff
@@ -183,9 +188,50 @@ pub fn parse_name_attribute(attrs: &mut Vec<syn::Attribute>) -> syn::Result<Opti
 pub fn build_py_function(ast: &mut syn::ItemFn, args: PyFunctionAttr) -> syn::Result<TokenStream> {
     let python_name =
         parse_name_attribute(&mut ast.attrs)?.unwrap_or_else(|| ast.sig.ident.unraw());
+
     add_fn_to_module(ast, python_name, args)
 }
 
+/// Parses #[pyo3(from_python_with = "func")]
+pub fn parse_arg_attributes(cap: &mut syn::PatType) -> syn::Result<PyFunctionArgAttr> {
```
kngwyu, line 196: How about taking `&mut Vec<syn::Attribute>` instead of PatType?
Then it's obvious what is mutated.

pyo3-macros-backend/src/pyfunction.rs
kngwyu: I understand, thank you for the explanation!
Then I think it's fine to leave this mutable, but let's use `&mut Vec<_>` instead.

pyo3-macros-backend/src/pyfunction.rs
daniil-konovalenko: Thanks for the tip! Fixed

daniil-konovalenko:
> Just bikeshedding: would you consider 'from_py'?

I think it's worth considering, but I personally prefer `from_py_with` more because it feels more familiar to me.

In any case, I feel that the final naming decision should be made by the maintainers, as they have much better understanding of pyo3's API as a whole

tests/test_class_basics.rs
```diff
@@ -244,3 +244,45 @@ fn panic_unsendable_base() {
 fn panic_unsendable_child() {
     test_unsendable::<UnsendableChild>().unwrap();
 }
+
+fn to_string(obj: &PyAny) -> PyResult<String> {
+    Ok(obj.str()?.to_str()?.to_string())
```
davidhewitt, line 249:
```suggestion
    obj.extract()
```

pyo3-macros-backend/src/method.rs
```diff
@@ -104,7 +106,7 @@ pub fn get_return_info(output: &syn::ReturnType) -> syn::Type {
     }
 }
 
-pub fn parse_method_receiver(arg: &syn::FnArg) -> syn::Result<SelfType> {
+pub fn parse_method_receiver(arg: &mut syn::FnArg) -> syn::Result<SelfType> {
```
davidhewitt, line 109: It doesn't look like `&mut` is needed here.

```suggestion
pub fn parse_method_receiver(arg: &syn::FnArg) -> syn::Result<SelfType> {
```

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -426,7 +466,7 @@ fn verify_and_get_lifetime(generics: &syn::Generics) -> Result<Option<&syn::Life
 ///   * Max 1 lifetime specifier, will be tied to `FromPyObject`'s specifier
 ///   * At least one field, in case of `#[transparent]`, exactly one field
 ///   * At least one variant for enums.
-///   * Fields of input structs and enums must implement `FromPyObject`
+///   * Fields of input structs and enums must implement `FromPyObject` or be annotated with from_py_with
```
davidhewitt, line 469:
```suggestion
///   * Fields of input structs and enums must implement `FromPyObject` or be annotated with `from_py_with`
```

pyo3-macros-backend/src/method.rs
```diff
@@ -128,12 +130,17 @@ impl<'a> FnSpec<'a> {
         } = parse_method_attributes(meth_attrs, allow_custom_name)?;
 
         let mut arguments = Vec::new();
-        let mut inputs_iter = sig.inputs.iter();
 
+        // TODO: maybe there's a cleaner solution
+        let inputs_empty = sig.inputs.is_empty();
+        let sig_span = sig.span();
+        let inputs_span = sig.inputs.span();
```
davidhewitt, line 137: Looks like this is down to lifetime of the `iter_mut()` borrow? Perhaps we can remove the need for these by rearranging the code.

pyo3-macros-backend/src/module.rs
```diff
@@ -142,12 +145,12 @@ pub fn add_fn_to_module(
...
     let mut arguments = Vec::new();
 
-    for (i, input) in func.sig.inputs.iter().enumerate() {
+    for (i, input) in func.sig.inputs.iter_mut().enumerate() {
         match input {
             syn::FnArg::Receiver(_) => {
                 bail_spanned!(input.span() => "unexpected receiver for #[pyfn]");
             }
-            syn::FnArg::Typed(cap) => {
+            syn::FnArg::Typed(ref mut cap) => {
```
davidhewitt, line 153: Is "ref mut" actually needed here? I had a feeling that with default match bindings `cap` will automatically be the mutable reference we want...

pyo3-macros-backend/src/pyfunction.rs
```diff
@@ -27,6 +27,11 @@ pub struct PyFunctionAttr {
     pub pass_module: bool,
 }
 
+#[derive(Clone, PartialEq, Debug)]
+pub struct PyFunctionArgAttrs {
+    pub from_py_with: Option<syn::ExprPath>,
```
davidhewitt, line 32: It looks like there's duplication here with `from_pyobject.rs`, both for this type declaration and the attribute parsing code itself.

Perhaps we can create a newtype in a new file `src/attrs.rs`, and put a newtype `struct FromPyWithAttribute(syn::ExprPath)` in there along with the parsing code which can be shared between the two?

Then here we can have:

```suggestion
    pub from_py_with: Option<FromPyWithAttribute>,
```

pyo3-macros-backend/src/pyfunction.rs
```diff
@@ -183,9 +188,63 @@ pub fn parse_name_attribute(attrs: &mut Vec<syn::Attribute>) -> syn::Result<Opti
 pub fn build_py_function(ast: &mut syn::ItemFn, args: PyFunctionAttr) -> syn::Result<TokenStream> {
     let python_name =
         parse_name_attribute(&mut ast.attrs)?.unwrap_or_else(|| ast.sig.ident.unraw());
+
```
davidhewitt, line 191: Bonus newline!

```suggestion
```

davidhewitt (commented):
Thanks again for this, and sorry for the delay from me getting round to reading it!

The implementation looks good and has some decent test coverage; I added a few comments on places in the code where I have some questions and ideas to improve it a bit.

Regarding name - I think you're correct that `from_py_with` is better than `from_python_with`.

I also think that `from_py_with` is better than `from_py`, as we're keeping in line with the precedent made by serde.

Would be nice to hear if @kngwyu has any strong preference on name also.

pyo3-macros-backend/src/module.rs
daniil-konovalenko: You're right, thanks! I guess it's one of the remnants of my fight with the borrow checker ;)

pyo3-macros-backend/src/method.rs
daniil-konovalenko: Yes, I couldn't leave them inline, because then it would be an immutable borrow during a mutable reference lifetime. I'll try to rearrange it somehow

tests/test_class_basics.rs
daniil-konovalenko: This one is tricky: I wanted to perform conversion explicitly via python's `str()` function, so that we could accept any object, like `dict`, for example. With `extract` we'll only be able to accept `FromPy<String>` objects and using `from_py_with` wouldn't make sense.

It's quite confusing, so I'll try to come up with a better example for this test that shows the need for a custom conversion function more clearly.

pyo3-macros-backend/src/pyfunction.rs
daniil-konovalenko: Thanks for the suggestion! It's much cleaner this way

pyo3-macros-backend/src/attrs.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct FromPyWithAttribute(syn::ExprPath);
+
+impl FromPyWithAttribute {
+    pub fn from_meta(meta: Meta) -> Result<Self> {
+        if let Meta::NameValue(MetaNameValue {
+            lit: syn::Lit::Str(string_literal),
+            ..
+        }) = meta
```
davidhewitt, line 14: I personally prefer do something like

```rust
let string_literal = match meta {
    Meta::NameValue(MetaNameValue { lit: syn::Lit::Str(string_literal), .. } => string_literal,
    meta => bail_spanned!(meta.span() => "expected a name-value: `pyo3(from_py_with = \"func\")` "),
};
```

... so that the rest of the function isn't indented one level.

But I'm happy as-is if you prefer the current form.

pyo3-macros-backend/src/attrs.rs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            let expr_path = string_literal.parse::<syn::ExprPath>()?;
+            return Ok(FromPyWithAttribute(expr_path));
+        }
+        bail_spanned!(meta.span() => "expected a name-value: `pyo3(from_py_with = \"func\")` ")
+    }
+}
+
+impl ToTokens for FromPyWithAttribute {
```
davidhewitt, line 23: Curious what this impl is needed for?

pyo3-macros-backend/src/pyfunction.rs
```diff
@@ -186,6 +192,50 @@ pub fn build_py_function(ast: &mut syn::ItemFn, args: PyFunctionAttr) -> syn::Re
...
+            if meta_list.path.is_ident("pyo3") {
+                for meta in meta_list.nested {
+                    metas.push(meta);
+                }
+            } else {
+                new_attrs.push(attr)
+            }
+        }
+    }
+    attrs.extend(new_attrs);
```
davidhewitt, line 210: After the drain, `attrs` will be empty, so it's more efficient to just overwrite.

```suggestion
    *attrs = new_attrs;
```

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -491,6 +491,12 @@ fn impl_arg_param(
         (None, false) => quote! { panic!("Failed to extract required method argument") },
     };
 
+    let extract = if let Some(expr) = &arg.attrs.from_py_with {
+        quote! {#expr(_obj.extract()?).map_err(#transform_error)?}
```
davidhewitt, line 495: This looks like it'll allow the `from_py_with` function to accept any type which is convertible from Python, instead of just `PyAny`.

This is probably more flexible than we need it, so I wonder if we should dial it back a bit and force the functions to take `PyAny`?

```suggestion
        quote! {#expr(_obj).map_err(#transform_error)?}
```

Also as written if the `.extract()` call failed it would not be passed through `#transform_error`, so if we want to keep support for typed inputs to the extraction function we should have:

```suggestion
        quote! {_obj.extract().and_then(#expr).map_err(#transform_error)?}
```

pyo3-macros-backend/src/pyproto.rs
```diff
@@ -53,7 +53,7 @@ fn impl_proto_impl(
     let module = proto.module();
 
     for iimpl in impls.iter_mut() {
-        if let syn::ImplItem::Method(met) = iimpl {
+        if let syn::ImplItem::Method(ref mut met) = iimpl {
```
davidhewitt, line 56:
```suggestion
        if let syn::ImplItem::Method(met) = iimpl {
```

tests/test_pyfunction.rs
```diff
@@ -111,6 +111,41 @@ fn test_functions_with_function_args() {
...
+        custom_conv_func,
+        r#"
+        import datetime
+
+        dt = datetime.datetime.fromtimestamp(1612040400)
+        assert custom_conv_func(dt) == 1612040400
+        "#
+    )
+}
+
```
davidhewitt, line 148: It would be nice to add a test where the custom extraction fails.

davidhewitt (commented):
Thanks for the continued work on this, this is looking increasingly solid.

I took another read just now and had a few more ideas...

pyo3-macros-backend/src/attrs.rs
daniil-konovalenko: Thanks! I agree, less indentation is better

pyo3-macros-backend/src/attrs.rs
daniil-konovalenko: Since I'd wrapped ExprPath in a struct, I couldn't use it in `quote!` without getting the nested `ExprPath` by index:
```rust
match attrs.from_py_with {
    Some(expr) => quote!(#expr.0),
    ...
}
```

I didn't like how `#expr.0` looked, so I proxied the`ToTokens` of the nested `ExprPath`. It became possible to use `FromPyWithAttribute` directly:
```rust
match attrs.from_py_with {
    Some(expr) => quote!(#expr),
    ...
}
```

Only now I see that it's not actually necessary, we can match for the nested ExprPath instead:

```rust
match attrs.from_py_with {
    Some(FromPyWithAttribute(expr)) => quote!(#expr),
    ...
}
```

pyo3-macros-backend/src/pyfunction.rs
daniil-konovalenko: Fixed, thanks!

pyo3-macros-backend/src/pymethod.rs
daniil-konovalenko: I think you're right, we should restrict inputs to `&PyAny`.
 
One can downcast or extract to a specific type explicitly in the conversion function, if neccessary. 
Also, seeing a non-python type in a conversion function signature might be confusing.

daniil-konovalenko:
CI just got updated to Rust 1.50 and it broke formatting, plus there's a couple of new Clippy rules

davidhewitt:
👍 I'll fix in a bit and you can rebase this PR on it after

daniil-konovalenko:
Awesome, thank you!

tests/test_class_basics.rs
daniil-konovalenko: Changed the test a bit, hope it's less confusing now

pyo3-macros-backend/src/method.rs
daniil-konovalenko: I can't get rid of it yet because I still use `iter_mut()` to iterate over arguments:
https://github.com/PyO3/pyo3/blob/68137635d2eeecd8f8da610ca44a0c99806d91d8/pyo3-macros-backend/src/method.rs#L139-L145

I'll remove it once I figure out how to deal with this mutable/immutable arguments iterator

pyo3-macros-backend/src/pymethod.rs
davidhewitt: Sounds good to me! If someone has a really compelling reason why something other than `&PyAny` should be supported later, we can revisit then.

davidhewitt:
See #1422 - just waiting for a second pair of eyes on it...

davidhewitt:
#1422 is merged, so if you rebase things should work now!

daniil-konovalenko:
Rebased, I hope it wouldn't mess up the comments too much

tests/test_class_basics.rs
```diff
@@ -244,3 +244,47 @@ fn panic_unsendable_base() {
...
+
+    Ok(length)
+}
+
+#[pyclass]
+struct ClassWithFromPyWithMethods {}
+
+#[pymethods]
+impl ClassWithFromPyWithMethods {
+    fn instance_method(&self, #[pyo3(from_py_with = "get_length")] argument: usize) -> usize {
```
davidhewitt, line 259: Just curious; what happens if you try `"PyAny::len" instead of `"get_length"` in the macro? I guess it'll work? 🤔

tests/test_frompyobject.rs
```diff
@@ -299,3 +299,34 @@ fn test_err_rename() {
...
+#[derive(Debug, FromPyObject)]
+pub struct Zap {
+    #[pyo3(item)]
+    name: String,
+
+    #[pyo3(from_py_with = "to_string", item("my_object"))]
+    some_object: String,
+}
+
+#[test]
```
davidhewitt, line 316: Final thought; could we please add tests for tuple structs and enum variants? Just to make sure that all cases are covered and remain covered.

davidhewitt (approved):
Thanks, the implementation looks great to me!

I'm happy with the design here, I think this is good to merge. @kngwyu are you also happy with this?

I see that the TODO list still includes CHANGELOG entry and docs.

I think docs should go in the guide. TBH the `#[name = ...]` attribute is also undocumented; I was going to move it to `#[pyo3(name = ...)]` and add a guide section to document attributes better at the same time. If you like you can just open an issue to document `#[pyo3(from_py_with)]` and I'll do it at the same time.

CHANGELOG entry - I suggest something like this:

```
- Add #[pyo3(from_py_with = "...")]` attribute for function arguments and struct fields to override the default from-Python conversion. [#1411](https://github.com/PyO3/pyo3/pull/1411)
```

tests/test_class_basics.rs
daniil-konovalenko: Hmm, I think it should work? I'm going to check it to be sure

tests/test_frompyobject.rs
daniil-konovalenko: Sure, thank you for the reminder!

daniil-konovalenko:
@davidhewitt 
Thank you for the thorough review! 
I like the changelog line that you proposed, I think it summarizes the changes really well.
I'll try to fix the remaining issues on the following week, probably closer to the weekend.
If you'll be able to update the guide it would be awesome! I've created an issue for this: #1432. 


I'm still a little bit bummed about this:
https://github.com/PyO3/pyo3/blob/d43c19b6a2888f0b947bd84f57394d88bf4859b9/pyo3-macros-backend/src/method.rs#L136-L138

I have tried a few things but I can't seem to fix it completely. 
My approach was to split up `FnSpec::parse` into different functions, it helped to find out which parts of the parsing code use which references. 
Turns out the main problem is this:
1. an immutable reference to `sig` must be valid during the whole `FnSpec::parse`
2. we need `sig.inputs.iter()` to parse `fn_spec` etc.
2. we need `sig.inputs.iter_mut()` to parse attributes.

If we just use the mutable iterator everywhere, then we would need the additional variables listed above. Also it introduces unnecessary mutability to `parse_method_receiver`. 

The other approach I tried is this:
1. Use an immutable iterator to parse `fn_type`. Count how many elements (N) from `sig.inputs` we used.
2. Obtain a mutable iterator and skip the first N elements, then parse the rest of the arguments.
This approach is somewhat awkward. I've got it to compile, but it does not pass all the tests yet.

There's also the third approach:  
1. Parse everything immutably
2. [remove attributes from the arguments](https://github.com/extrawurst/alpaca-rust/commit/46dbba66cf73f6d05f7fdda2d6e1b9d6e30384c2#diff-8f097454b072a95a96c470f9c3f84c5fR136-R151). 
see https://github.com/PyO3/pyo3/pull/1411#discussion_r570174799 for the details.  
I'm not a fan of this one because it would be easy to forget to remove attributes. Also it adds an implicit dependency between the parsing code and the removal code.

At this point I'm stuck. I would love to hear any ideas that you have on this.

kngwyu:
> Would be nice to hear if @kngwyu has any strong preference on name also.

No, I'm fine with the current name.

kngwyu:
I am now still playing in the code but will leave some suggestions in a day.

daniil-konovalenko:
I've pushed the second approach implementation. It's awkward around skipping arguments, but otherwise I like it more that what I had before.

tests/test_class_basics.rs
daniil-konovalenko: Yep, it works! Do you think we should use it instead?

davidhewitt: Might be nice to use one of each? :)

daniil-konovalenko: done!

kngwyu (commented):
Thanks, your last commit nicely simplifies the code. 👍🏼 
I left some comments to make the intention clearer.

pyo3-macros-backend/src/method.rs
```diff
@@ -148,7 +249,6 @@ impl<'a> FnSpec<'a> {
             }
         };
 
-        // Parse receiver & function type for various method types
         let fn_type = match fn_type_attr {
             Some(MethodTypeAttribute::StaticMethod) => FnType::FnStatic,
```
kngwyu, line 253:
```suggestion
        let (fn_type, skip_first_arg) = match fn_type_attr {
            Some(MethodTypeAttribute::StaticMethod) => (FnType::FnStatic, false),
```

daniil-konovalenko:
@kngwyu Thanks for the suggestions! It looks a lot better now

kngwyu (approved):
Thanks, I think now it's ready 👍🏼

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Add conversions between `OsStr`/`OsString`/`Path`/`PathBuf` and Python strings. [#1379](https://github.com/PyO3/pyo3/pull/1379)
 - Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
+- Add #[pyo3(from_py_with = "...")]` attribute for function arguments and struct fields to override the default from-Python conversion. [#1411](https://github.com/PyO3/pyo3/pull/1411)
```
davidhewitt, line 12: Nit: these tend to be sorted by PR number:

```suggestion
- Add #[pyo3(from_py_with = "...")]` attribute for function arguments and struct fields to override the default from-Python conversion. [#1411](https://github.com/PyO3/pyo3/pull/1411)
- Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
```

pyo3-macros-backend/src/method.rs
```diff
@@ -115,146 +117,188 @@ pub fn parse_method_receiver(arg: &syn::FnArg) -> syn::Result<SelfType> {
...
-                utils::parse_text_signature_attrs(meth_attrs, &python_name)?
-            {
-                bail_spanned!(text_signature.span() => error_msg)
+    fn parse_fn_type(
+        sig: &syn::Signature,
+        fn_type_attr: Option<MethodTypeAttribute>,
+        python_name: &mut Option<syn::Ident>,
+    ) -> syn::Result<(FnType, bool)> {
+        let name = &sig.ident;
+        let mut inputs_iter = sig.inputs.iter().enumerate();
```
davidhewitt, line 233: It looks to me that this is only used once in a `let _ = inputs_iter.next()` for `#[classmethod]`, so my conclusion is that we don't need this at all any more and can throw it away?

```suggestion
```

davidhewitt (approved):
This all looks great to me, thanks for your hard work on this. The implementation of `FnSpec::parse` is a lot better now, really nice job.

Few final nits and then let's merge this. It could also be worth squashing this down to one or two commits from 22.

pyo3-macros-backend/src/method.rs
daniil-konovalenko: fixed, thanks!

tests/test_frompyobject.rs
daniil-konovalenko: I've finally got to this and I have a question. At the moment we provide `from_py_with` for struct fields only.
Do you have in mind that `from_py_with` should be supported for tuple structs and enums as well?

davidhewitt: Hmm, that's a good point, I didn't really specify those on #1239 when I proposed the idea.

I think for tuple structs with `#[derive(FromPyObject)]` it's quite natural to want to be able to annotate fields in the same way. Maybe let's skip on deciding what to do with enum variants for now. That'll be easy to add later anyway.

If tuple structs are going to be a lot of extra work, perhaps let's merge this PR for now and add an issue to remember to support them later?

daniil-konovalenko: Thanks for clearing this out! I'll try supporting tuple structs, maybe it won't be that hard.
In any case, I think it's worth adding a compiler error (warning?) when `from_py_with` is used on types that are not supported

kngwyu:
> add tests for enums/tuple structs

Is this the only blocker?

daniil-konovalenko:
> Is this the only blocker?

I forgot to update the TODO: I still need to implement `from_py_with` for tuple structs and enums, but it can be deferred to a future PR.
See our discussion with David:
https://github.com/PyO3/pyo3/pull/1411#discussion_r578385714

kngwyu:
I see, thank you!

PR #1413, opened 2021-02-08
List pyca/cryptography in the README

alex:
We shipped our first release with a Rust extension module today! You can see the results https://pepy.tech/project/setuptools-rust :-)


kngwyu (approved):
Thank you and congrats :tada:

davidhewitt:
This is really awesome, congratulations, pyo3 download numbers on crates.io picking up too! (Afaik they show 7-day averages only now, so it'll take a few days to see where we level off at...)

PR #1415, opened 2021-02-08
Added async/await example to the guide

davidhewitt:
This is a copy of #1393, which I accidentally closed and can't reopen after a bad force-push... 🤦


davidhewitt (commented):
Thanks very much @awestlake87, this guide looks great to me!

I note that in https://github.com/awestlake87/pyo3-asyncio/pull/10 there's a few changes which might bring need to edit this document.

I'm thinking I'd like to merge this PR soon so that we can get on with the release. Perhaps shall I add a note at the top of the guide that `pyo3-asyncio` is still a new and fast-evolving library, so the exact API may change faster than this document can keep up?

awestlake87:
I have the API docs updated for the `#[main]` and `#[test]` attributes, so I could add those changes to the guide pretty quick if we merged awestlake87/pyo3-asyncio#10. It seems like the only thing that we'd need to decide is whether or not to include the suite name argument in the test harness main function.

davidhewitt:
👍 Given that I broke the permissions on the original PR, I'm going to merge this guide entry now. Let's update it once awestlake87/pyo3-asyncio#10 is merged, and then release.

There's no specific urgency on the release, so I think that ordering is fine.

awestlake87:
Works for me, I'll open another PR to update it once it's ready!

PR #1417, opened 2021-02-10
Updated guide after merging awestlake87/pyo3-asyncio#10 

awestlake87:
Updated the guide to describe the initialization and testing for `pyo3-asyncio` using the new proc macro attributes.


davidhewitt (approved):
👍 thanks very much for turning this around so fast!

PR #1419, opened 2021-02-10
Fix the PyPy link in README.md

cfbolz:
I noticed that the pypy link in the readme is broken and saw that the file was moved. Hope that fix is acceptable! Thanks for caring about PyPy :-)


davidhewitt:
👍 thanks for the fix!

We try to maintain decent PyPy support for PyO3. I haven't tested with PyPy 3.7 yet though 😅

PR #1421, opened 2021-02-10
Restore compatibility with Rust 1.41(.1).

### LINKED ISSUE #1420: Support for older Rust versions (1.41)?
## Feature request

pyca/cryptography users have been asking for support of Debian Buster (current stable), Alpine 3.12, and other platforms with older Rust compiler versions. The common baseline seems to be 1.41. pyo3 does not compile with Rust 1.41 and seem to require at least Rust 1.45. Would it be feasible to support 1.41?

https://buildd.debian.org/status/package.php?p=rustc&suite=buster

### 🌍 Environment

 - Your operating system and version: Debian Buster, Alpine <= 3.12
 - Your python version: /
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: system python
 - Your Rust version (`rustc --version`): 1.41
 - Your PyO3 version: git master
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`:

### Reproducer

```
$ podman run -ti debian:stable
# apt update
# apt install -y git cargo python3-dev build-essential
# git clone https://github.com/PyO3/pyo3.git
# cd pyo3
# rustc -V
rustc 1.41.1
```

```
# cargo build
   Compiling pyo3-macros-backend v0.13.1 (/root/pyo3/pyo3-macros-backend)
error[E0658]: subslice patterns are unstable
   --> pyo3-macros-backend/src/pyfunction.rs:177:36
    |
177 |         [_first_attr, second_attr, ..] => bail_spanned!(
    |                                    ^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/62254

error[E0658]: subslice patterns are unstable
   --> pyo3-macros-backend/src/pymethod.rs:707:14
    |
707 |         [py, args @ ..] if utils::is_python(&py.ty) => (Some(py), args),
    |              ^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/62254

error[E0658]: use of unstable library feature 'str_strip': newly added
   --> pyo3-macros-backend/src/method.rs:144:18
    |
144 |                 .strip_prefix(prefix)
    |                  ^^^^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/67302

error: aborting due to 3 previous errors

For more information about this error, try `rustc --explain E0658`.
error: could not compile `pyo3-macros-backend`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

birkenfeld:
This version is currently supported by Debian stable and Alpine Linux.

Fixes #1420

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![Actions Status](https://github.com/PyO3/pyo3/workflows/Test/badge.svg)](https://github.com/PyO3/pyo3/actions)
 [![codecov](https://codecov.io/gh/PyO3/pyo3/branch/master/graph/badge.svg)](https://codecov.io/gh/PyO3/pyo3)
 [![crates.io](http://meritbadge.herokuapp.com/pyo3)](https://crates.io/crates/pyo3)
-[![minimum rustc 1.45](https://img.shields.io/badge/rustc-1.45+-blue.svg)](https://rust-lang.github.io/rfcs/2495-min-rust-version.html)
+[![minimum rustc 1.41](https://img.shields.io/badge/rustc-1.41+-blue.svg)](https://rust-lang.github.io/rfcs/2495-min-rust-version.html)
 [![Join the dev chat](https://img.shields.io/gitter/room/nwjs/nw.js.svg)](https://gitter.im/PyO3/Lobby)
 
 [Rust](http://www.rust-lang.org/) bindings for [Python](https://www.python.org/). This includes running and interacting with Python code from a Rust binary, as well as writing native Python modules.
@@ -18,7 +18,7 @@ A comparison with rust-cpython can be found [in the guide](https://pyo3.rs/maste
 
 ## Usage
 
-PyO3 supports Python 3.6 and up. The minimum required Rust version is 1.45.0.
+PyO3 supports Python 3.6 and up. The minimum required Rust version is 1.41.
 
 Building with PyPy is also possible (via cpyext) for Python 3.6, targeted PyPy version is 7.3+.
 Please refer to the [pypy section in the guide](https://pyo3.rs/master/pypy.html).
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -115,6 +115,7 @@ pub fn parse_method_receiver(arg: &syn::FnArg) -> syn::Result<SelfType> {
 
 impl<'a> FnSpec<'a> {
     /// Parser function signature and function attributes
+    #[allow(clippy::manual_strip)] // for strip_prefix replacement supporting rust < 1.45
     pub fn parse(
         sig: &'a syn::Signature,
         meth_attrs: &mut Vec<syn::Attribute>,
@@ -139,10 +140,12 @@ impl<'a> FnSpec<'a> {
 
         // strip get_ or set_
         let strip_fn_name = |prefix: &'static str| {
-            name.unraw()
-                .to_string()
-                .strip_prefix(prefix)
-                .map(|rest| syn::Ident::new(rest, name.span()))
+            let ident = name.unraw().to_string();
+            if ident.starts_with(prefix) {
+                Some(syn::Ident::new(&ident[prefix.len()..], ident.span()))
+            } else {
+                None
+            }
         };
 
         // Parse receiver & function type for various method types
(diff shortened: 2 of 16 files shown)


pyo3-macros-backend/src/method.rs
```diff
@@ -115,6 +115,7 @@ pub fn parse_method_receiver(arg: &syn::FnArg) -> syn::Result<SelfType> {
 
 impl<'a> FnSpec<'a> {
     /// Parser function signature and function attributes
+    #[allow(clippy::manual_strip)] // for strip_prefix replacement below
```
davidhewitt, line 118:
```suggestion
    #[allow(clippy::manual_strip)] // for strip_prefix replacement supporting rust < 1.45
```

README.md
```diff
@@ -18,7 +18,7 @@ A comparison with rust-cpython can be found [in the guide](https://pyo3.rs/maste
 
 ## Usage
 
-PyO3 supports Python 3.6 and up. The minimum required Rust version is 1.45.0.
+PyO3 supports Python 3.6 and up. The minimum required Rust version is 1.41.1.
```
davidhewitt, line 21: Imo patch releases probably aren't that interesting unless we do actually depend on the patch fix. Shall we just use `major.minor` here?

```suggestion
PyO3 supports Python 3.6 and up. The minimum required Rust version is 1.41.
```

davidhewitt (approved):
Thanks! It's a shame to give up the ergonomics of the newer compiler. However I think making the adoption of cryptography into as many packaging solutions as possible is worth the (admittedly still small) costs to us.

Could you please add this into the `CHANGELOG` in a `## Packaging` section at the top? (Similar to the 0.13 release ordering.)

davidhewitt:
Regarding UI tests, we use `#[rustversion]` to move problematic test cases into gated functions. See `test_compile_error.rs`.

I suggest moving anything that doesn't pass on Rust 1.41 into a `tests_rust_1_45` function.

README.md
birkenfeld: Agreed.

pyo3-macros-backend/src/method.rs
birkenfeld: Done.

src/freelist.rs
```diff
@@ -93,9 +94,10 @@ where
             let free = get_type_free(ty).unwrap_or_else(|| tp_free_fallback(ty));
             free(obj as *mut c_void);
 
-            #[cfg(Py_3_8)]
-            if ffi::PyType_HasFeature(ty, ffi::Py_TPFLAGS_HEAPTYPE) != 0 {
-                ffi::Py_DECREF(ty as *mut ffi::PyObject);
+            if cfg!(Py_3_8) {
```
kngwyu, line 97: Why do you prefer this instead of `if cfg!(Py_3_8) && .. `?
Or, if you want to avoid `#[allow(clippy)]` you can use
```rust
            cfg_if::cfg_if! {
                if #[cfg(Py_3_8)] {
                    if ffi::PyType_HasFeature(ty, ffi::Py_TPFLAGS_HEAPTYPE) != 0 {
                        ffi::Py_DECREF(ty as *mut ffi::PyObject);
                    }
                }
            }
```
Though I'm not sure which is bettter.

kngwyu (commented):
Thanks 😉

src/freelist.rs
birkenfeld: I was thinking of later, when the `#[cfg]` form can be used again, it is easier to replace this back.

In general I hate the "collapsible_if" lint, since it is in at least 50% of cases *not* an oversight and makes the code clearer.

davidhewitt (approved):
👍 This looks great to me, thanks for fixing up!

Will merge this and rebase the patch release.

src/buffer.rs
```diff
@@ -48,7 +48,12 @@ impl ElementType {
     pub fn from_format(format: &CStr) -> ElementType {
         match format.to_bytes() {
             [char] | [b'@', char] => native_element_type_from_type_char(*char),
-            [modifier, char] if matches!(modifier, b'=' | b'<' | b'>' | b'!') => {
+            [modifier, char]
+                if (*modifier == b'='
+                    || *modifier == b'<'
+                    || *modifier == b'>'
+                    || *modifier == b'!') =>
+            {
```
chris-morgan, line 56:
```suggestion
            [b'=', char] | [b'<', char] | [b'>', char] | [b'!', char] => {
```

PR #1422, opened 2021-02-11
rust 1.50: clippy and lint fixes

davidhewitt:


src/class/methods.rs
```diff
@@ -10,15 +10,15 @@ use std::os::raw::c_int;
 #[derive(Debug)]
 pub enum PyMethodDefType {
     /// Represents class `__new__` method
-    New(PyMethodDef),
+    New(PyMethodDef<ffi::newfunc>),
```
davidhewitt, line 13: The changes in this module are _technically_ breaking, but it's `#[doc(hidden)]` internals so I think this refactor is fine.

I did it because it removes some of the type assertions from runtime to compile time, which can only be a good thing for performance 🚀

davidhewitt: I decided this was fine based on comment by dtolnay here: https://www.reddit.com/r/rust/comments/ey0ul0/handling_breaking_api_changes/fgew8d2

kngwyu:
Thanks!

PR #1423, opened 2021-02-12
Support building with Mingw Python

lazka:
The DLL of the mingw Python in MSYS2 is named libpython3.8.dll:

$ python3 -m sysconfig | grep LIBPYTHON
        LIBPYTHON = "-lpython3.8"

Add another special case to in get_rustc_link_lib() to handle that case.
Afaik the mingw build doesn't support the limited ABI, so skip that as well.

This makes all tests pass in an MSYS2 environment and lets us build
python-cryptography.


lazka:
I know mingw Python isn't officially supported, but I thought this change is small enough to be considered anyway.

davidhewitt:
Thanks for the PR! I'm willing to merge this. Before we do, I wanted to ask whether we also need to consider adding a similar special case for PyPy?

lazka:
Thanks. PyPy doesn't support mingw, so for that case nothing is needed. As for PyPy on Windows in general, I'd need to check.

davidhewitt:
Ah great. We already support PyPy on Windows so it sounds like there's no further changes needed here. 

Would you mind rebasing on master? That should fix the CI issues.

(If you're willing to add a note to the `## Fixed` section of the CHANGELOG documenting this at the same time, that'd be great.)

davidhewitt:
Actually nvm, I'll do this now as I'm preparing the next patch release atm. Sorry if you already started!

lazka:
Thanks!

PR #1425, opened 2021-02-13
ffi cleanup: methodobject to moduleobject

nw0:
Main changes are to deprecate some functions to reflect the C API. This one isn't too challenging either, but next is `object.h`, which will require some major refactoring to `cpython/`.

This PR reflects `PyModule_GetFilename` (returning a `char*`) being deprecated in Python 3.2 in favour of `PyModule_GetFilenameObject` (returning a `PyObject`). We do use the former, but for now I've just put an `#[allow(deprecated)]` on that line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 ### Added
-- Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later.
+- Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
 
 ### Changed
-- Deprecate FFI definition `PyCFunction_Call` for Python 3.9 and later.
-- Deprecate FFI definitions `PyModule_GetFilename`, `PyMethodDef_INIT`.
+- Deprecate FFI definition `PyCFunction_Call` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
+- Deprecate FFI definitions `PyModule_GetFilename`, `PyMethodDef_INIT`. [#1425](https://github.com/PyO3/pyo3/pull/1425)
 
 ### Removed
-- Remove FFI definition `PyCFunction_ClearFreeList` for Python 3.9 and later.
+- Remove FFI definition `PyCFunction_ClearFreeList` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
 
 ## [0.13.2] - 2021-02-12
 ### Packaging
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ PyO3 versions, please see the [migration guide](https://pyo3.rs/master/migration
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Added
+- Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later.
+
+### Changed
+- Deprecate FFI definition `PyCFunction_Call` for Python 3.9 and later.
+- Deprecate FFI definitions `PyModule_GetFilename`, `PyMethodDef_INIT`.
+
+### Removed
+- Remove FFI definition `PyCFunction_ClearFreeList` for Python 3.9 and later.
+
 ## [0.13.2] - 2021-02-12
 ### Packaging
 - Lower minimum supported Rust version to 1.41. [#1421](https://github.com/PyO3/pyo3/pull/1421)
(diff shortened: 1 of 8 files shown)


src/types/module.rs
```diff
@@ -121,7 +121,12 @@ impl PyModule {
     /// May fail if the module does not have a `__file__` attribute.
     #[cfg(not(all(windows, PyPy)))]
     pub fn filename(&self) -> PyResult<&str> {
-        unsafe { self.str_from_ptr(ffi::PyModule_GetFilename(self.as_ptr())) }
+        unsafe {
+            self.str_from_ptr(
+                #[allow(deprecated)]
+                ffi::PyModule_GetFilename(self.as_ptr()),
```
nw0, line 127: I think we should change this to use `PyModule_GetFilenameObject`.

So far this series of PRs has only been the minimum changes for #1289, but maybe an update here would be appropriate?

davidhewitt: I agree we can change this. It looks like `PyModule_GetFileameObject` is guaranteed to return `PyString` on success, or `NULL` on failure, so you should be able to do something like `self.py().from_owned_ptr_or_err::<PyString>(PyModule_GetFilenameObject(self.as_ptr()))?.to_str()`

src/ffi/methodobject.rs
```diff
@@ -8,6 +8,12 @@ extern "C" {
...
 
+#[cfg(Py_3_9)]
+#[inline]
+pub unsafe fn PyCFunction_CheckExact(op: *mut PyObject) -> c_int {
+    (Py_TYPE(op) == &mut PyCFunction_Type) as c_int
+}
+
 #[inline]
 pub unsafe fn PyCFunction_Check(op: *mut PyObject) -> c_int {
     (Py_TYPE(op) == &mut PyCFunction_Type) as c_int
```
davidhewitt, line 19: This definition `PyCFunction_Check` is the same as the new `PyCFunction_CheckExact` - does it need updating?

CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ PyO3 versions, please see the [migration guide](https://pyo3.rs/master/migration
...
+## Unreleased
+### Added
+- Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
+
+### Changed
+- Deprecate FFI definition `PyCFunction_Call` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
+- Deprecate FFI definitions `PyModule_GetFilename`, `PyMethodDef_INIT`. [#1425](https://github.com/PyO3/pyo3/pull/1425)
+
+### Removed
+- Remove FFI definition `PyCFunction_ClearFreeList` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
```
davidhewitt, line 17: I think this can count as a `Fixed` ffi definition, as this wouldn't actually work, so it's not like we're taking functionality away!

davidhewitt (approved):
As always these changes look great. Thank you so much for your careful efforts on this cleanup!

src/ffi/methodobject.rs
nw0: Oops, thanks. Looks like this one was renamed in Python 3.9, with the new definition of `PyCFunction_Check` using `PyObject_TypeCheck`.

nw0: Looks like 3.9 introduced the `Py_IS_TYPE` macro in `object.h`, which I'll add -- should help catch these changes when eyeballing

src/types/module.rs
```diff
@@ -121,7 +121,11 @@ impl PyModule {
     /// May fail if the module does not have a `__file__` attribute.
     #[cfg(not(all(windows, PyPy)))]
     pub fn filename(&self) -> PyResult<&str> {
-        unsafe { self.str_from_ptr(ffi::PyModule_GetFilename(self.as_ptr())) }
+        unsafe {
+            self.py()
+                .from_owned_ptr_or_err::<PyString>(ffi::PyModule_GetFilenameObject(self.as_ptr()))?
```
kngwyu, line 126: 👍🏼

src/ffi/moduleobject.rs
```diff
@@ -85,6 +91,10 @@ pub struct PyModuleDef {
     pub m_free: Option<freefunc>,
 }
 
+/// Helper initial value of [`PyModuleDef`] for a Python class.
+///
+/// Not present in the Python C API.
+#[deprecated(note = "not present in Python headers; to be removed")]
 pub const PyModuleDef_INIT: PyModuleDef = PyModuleDef {
```
kngwyu, line 98: This would be breaking, but I guess the only user is `pyclass.rs` and we can remove this.

kngwyu (approved):
Thanks!

src/ffi/moduleobject.rs
davidhewitt: Agreed, as I think we're going to do a breaking release next anyway it's fine to just scrap this.

nw0: @davidhewitt You'll do this in #1426 or should I add to this PR?

davidhewitt: I think probably simpler for you to do it in this PR so there's no risk of merge conflicts.

davidhewitt: Actually nvm, I see that other than that, this PR is ready to go. I'll merge this PR now and do it in #1426.

davidhewitt:
👍 Thanks again!

PR #1426, opened 2021-02-13
0.14: deprecations

davidhewitt:
This PR just cleans up a few deprecated things now that I think we're probably next going to release 0.14.

- Old exception names like `BaseException` from before we renamed them to `PyBaseException` etc.
- Some ffi definitions identified by @nw0 which aren't actually upstream APIs.
- Some deprecated methods on `Python` and `Py`.


kngwyu (approved):
Thanks 💯

Issue #1385, opened 2021-01-14
Most efficient way of returning large byte array from Rust to Python


cswinter:
I'm looking for some guidance on the best way to return a large byte array to Python code and/or share access to raw memory between Rust and Python.

Basically, my application transfers large amounts of data over the network in Rust and then hands the data over to Python (but in some cases I still want concurrent access to the data on the Rust side as well).
One of my ideas is to allocate a numpy array and have the networking code write into it directly, though in that case I don't think the Rust code can still safely access the array because it might be modified concurrently (related question, if I store a reference to a numpy array object on the Rust side, will GC work properly or could the numpy array get deallocated once all references on the Python side disappear?).

I think [`PyBytes::from_ptr`](https://docs.rs/pyo3/0.13.1/pyo3/types/struct.PyBytes.html#method.from_ptr) might be what I want but I haven't found any documentation on what invariants it expects to be maintained without triggering unsafety.
If I have a `Vec<u8>` on the Rust side and return direct access to it with `PyBytes::from_ptr`, I assume this is fine as long as I don't deallocate the `Vec` and the Python code doesn't do anything funky? Is there way to ever safely/automatically deallocate the `Vec`, or would that require something like exposing a `free` method that has to be called explicitly by the Python code after it stops accessing the bytes object?

cswinter:
I see the `PyBytes` has a [`get_refcnt`](https://docs.rs/pyo3/0.13.1/pyo3/types/struct.PyBytes.html#method.get_refcnt) function, if I keep one copy of the `PyBytes` object on the Rust side, is it then safe to automatically deallocate it and the underlying storage once the reference count reaches 1?

althonos:
Sounds like a job for `memoryview`, but there is no stable interface in `pyo3` for that!

davidhewitt:
Ahhh yes that looks like it! https://docs.python.org/3/c-api/memoryview.html

We have some ffi definitions at `src/ffi/memoryobject.rs` which look slightly out of date. They could probably use an update (see #1289). It would also be completely appropriate to add this as a new `&PyMemoryView` native type, I guess, in `src/types/memoryview.rs`.

cswinter:
Thanks for the pointers, looks like `memoryview` might be the way to go. Will do some more reading.

cswinter:
[...]

PR #1427, opened 2021-02-13
[WIP] Efficient movement of Vec<u8> to PyBytes

milesgranger:
Hi!

This is related to #1385 

I have a  de/compression lib which gets some bytes in from Python, then passes them back.
Really like how `PyBytes::new_with` works; unfortunately, I don't know the exact size of the slice I want until _after_ the de/compression operation. Using this approach, I end up with trailing null bytes. 
```python
>>> compress(b'bytes')
b'\xff\x06\x00\x00sNaPpY\x01\t\x00\x00\xb5\x8b\xa8\xdbbytes\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

Using the standard `PyBytes::new` from a slice, ends up making the new allocation and thus _just_ puts the lib slower than the existing Python/C options. However, the approach above can be reliably 1.5x - 2x faster; so really pulling that something like this is possible. :crossed_fingers: 

As you see in the PR, I naively attempted to use `ffi::_PyBytes_Resize` but it seems to have no effect?

Thanks for the help!

(related to benchmarks: https://github.com/milesgranger/pyrus-cramjam/pull/12#issuecomment-777811957)

### DIFF
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -1,7 +1,4 @@
-use crate::{
-    ffi, AsPyPointer, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, PyTryFrom, Python,
-    ToPyObject,
-};
+use crate::{ffi, AsPyPointer, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, PyTryFrom, Python, ToPyObject, PyErr};
 use std::ops::Index;
 use std::os::raw::c_char;
 use std::slice::SliceIndex;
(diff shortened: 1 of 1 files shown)


birkenfeld:
The fact that `_PyBytes_Resize` gets a double pointer is a hint to how it works: it does a realloc, so the object's address can change, and the old address is invalid.

This is definitely not safe to use with the `Py` reference, I think you'd have to do everything in `PyObject*` form.

However, this (overallocating and then truncating) sounds more like a task for `bytearray`, which natively supports truncation?

milesgranger:
Thanks for the pointers! :+1:

indygreg:
I was going to file an issue requesting a higher-level API utilizing `_PyBytes_Resize` because I want it for https://github.com/indygreg/python-zstandard so (de)compression uses 0-copy and stumbled upon this issue. (The existing C extension code uses `_PyBytes_Resize` heavily.)

While I agree `bytearray` is a better type for mutable binary buffers, the reason I don't want to use it is because it is a separate type from `bytes`. python-zstandard today uses `bytes` as its output type and switching this to `bytearray` would likely cause issues with existing code. Yes, you can coerce from `bytearray` to `bytes` easily. However, this isn't 0-copy and is therefore prone to performance overhead.

I could certainly call the Python C APIs directly. But I was hoping there would be room in PyO3 to support some (possibly `unsafe`) methods on its `PyBytes` wrapper to support resizing buffers. But from the looks of this conversation it appears the mutable object address (via a potential realloc) is a deal-breaker due to how PyO3 tracks objects?

davidhewitt:
@indygreg we actually once considered an extension to the `PyBytes::new_with` API to allow safely resizing at the end of the initialization step - see https://github.com/PyO3/pyo3/pull/1074#issuecomment-668740716.

You could potentially roll something yourself which does that - or add it as a PR to pyo3. There's a long discussion in that thread why we decided not to do it at the time.

In addition, we could support `unsafe fn resize(&mut self, new_size: usize)` on `PyBytes<'py>` after the owning object model in #1308.

PR #1429, opened 2021-02-14
ffi cleanup: object.h

nw0:
This one's a fairly big change to account for the new `cpython/object.h` file.

Deprecate some FFI definitions which are not in the API
- PyObject_Check
- PySuper_Check
- FreeFunc (note: there's also a `freefunc` which is subtly different -- fixed `_PyEval_RequestCodeExtraIndex`, which takes a `freefunc` as an argument)

I thought the INCREF, DECREF, ... methods were a bit off. ~~See review (soon).~~ EDIT: ~~will open new issue~~ this is fine

### DIFF
--- a/src/ffi/cpython/object.rs
+++ b/src/ffi/cpython/object.rs
@@ -1,0 +1,483 @@
+use crate::ffi::PyObject;
+use std::os::raw::c_int;
+
+// skipped _Py_NewReference
+// skipped _Py_ForgetReference
+// skipped _Py_GetRefTotal
+
+// skipped _Py_Identifier
+
+// skipped _Py_static_string_init
+// skipped _Py_static_string
+// skipped _Py_IDENTIFIER
+
+mod bufferinfo {
+    use crate::ffi::Py_ssize_t;
+    use std::os::raw::{c_char, c_int, c_void};
+    use std::ptr;
+
+    #[repr(C)]
+    #[derive(Copy, Clone)]
+    pub struct Py_buffer {
+        pub buf: *mut c_void,
+        /// Owned reference
+        pub obj: *mut crate::ffi::PyObject,
+        pub len: Py_ssize_t,
+        pub itemsize: Py_ssize_t,
+        pub readonly: c_int,
+        pub ndim: c_int,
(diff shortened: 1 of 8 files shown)


src/ffi/object.rs
```diff
@@ -1,30 +1,18 @@
+// FFI note: this file changed a lot between 3.6 and 3.10.
+// Some missing definitions may not be marked "skipped".
```
davidhewitt, line 2: 👍 yep good idea; over time I'm sure that we'll eventually fill in any gaps users need!

src/ffi/cpython/object.rs
```diff
@@ -0,0 +1,483 @@
...
+        tp_allocs: 0,
+        tp_frees: 0,
+        tp_maxalloc: 0,
+        tp_prev: ptr::null_mut(),
+        tp_next: ptr::null_mut(),
+    };
+
+    #[cfg(not(py_sys_config = "COUNT_ALLOCS"))]
+    #[deprecated(note = "not present in Python headers; to be removed")]
+    pub const PyTypeObject_INIT: PyTypeObject = type_object_init!();
```
davidhewitt, line 383: TBH I think this falls under the same bracket of the other `_INIT` constants which aren't actually defined upstream; we can just throw it away for the 0.14 release. Nobody should be using it.

Looks like that would allow quite a lot of macro code above to be deleted too!

davidhewitt (approved):
Looks great, thanks very much again. This is definitely one of the more awkward files; as always thank you for your diligent reconciliation.

src/ffi/cpython/object.rs
nw0: I thought so, but didn't want to presume. Updated.

Shall we also remove `PyObject_Check`, `PySuper_Check`, and `FreeFunc`? They're not in the Python API either, and I struggle to imagine why, for example, a downstream user would need need a function to verify that a PyObject is a Python object (always returns 1).

davidhewitt: Agreed that we can remove those three too!

kngwyu (approved):
Awesome 💯 , thank you.

PR #1436, opened 2021-02-19
ffi cleanup: abstract.h

nw0:
* This PR makes PyIter_Check available to Python earlier than 3.8 under the limited API.
This was due to some interaction between the `cfg(not(Py_LIMITED_API)` on `ffi::cpython`, and the cfg on PyIter_Check within the `ffi::cpython` module.
However, it's always been in the limited API: as a macro before 3.8, and as a function since 3.8. Under the non-limited API it's always a macro.

* Consequently we can implement PyTryFrom for PyIterator for all versions. I've (speculatively) commented out the cfg preventing that.

* Similar changes for PyIndex_Check.

### DIFF
--- a/src/ffi/abstract_.rs
+++ b/src/ffi/abstract_.rs
@@ -89,7 +89,10 @@ extern "C" {
     pub fn PyObject_GetIter(arg1: *mut PyObject) -> *mut PyObject;
 }
 
-#[cfg(not(any(all(Py_3_8, Py_LIMITED_API), PyPy)))]
+// Defined as this macro in Python 3.6, 3.7 limited API, but relies on
+// non-limited PyTypeObject. Don't expose this since it cannot be used.
+#[cfg(not(any(Py_LIMITED_API, PyPy)))]
+#[inline]
 pub unsafe fn PyIter_Check(o: *mut PyObject) -> c_int {
     (match (*crate::ffi::Py_TYPE(o)).tp_iternext {
         Some(tp_iternext) => {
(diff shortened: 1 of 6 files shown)


kngwyu (approved):
Good catch 👍🏼 , thank you!

davidhewitt (approved):
👍 this is a great catch, thanks!

davidhewitt:
I'm going to rebase this, force-push and merge shortly. Thanks again.

davidhewitt:
Hmm, after rebasing we still have issues: the "macro" forms of `PyIter_Check` and `PyIndex_Check` use the contents of the type object, which isn't available with the limited api...

nw0:
> Hmm, after rebasing we still have issues: the "macro" forms of `PyIter_Check` and `PyIndex_Check` use the contents of the type object, which isn't available with the limited api...

Ouch. I've double-checked and PyIter_Check in 3.7 tries to use the contents of the opaque PyTypeObject, even under the limited API. I wonder why nobody noticed?

I think the best thing to do here is to remove the functions again :( will do this by the weekend.

src/ffi/abstract_.rs
```diff
@@ -84,16 +87,30 @@ extern "C" {
...
 
-    // PyIter_Check for unlimited API is in cpython/abstract_.rs
-    #[cfg(any(all(Py_LIMITED_API, Py_3_8), PyPy))]
-    #[cfg_attr(PyPy, link_name = "PyPyIter_Check")]
-    pub fn PyIter_Check(obj: *mut PyObject) -> c_int;
+// Defined as this macro in Python 3.6, 3.7 limited API, but relies on
+// non-limited PyTypeObject. Don't expose this since it cannot be used.
+#[cfg(not(any(Py_LIMITED_API, PyPy)))]
+#[inline]
+pub unsafe fn PyIter_Check(o: *mut PyObject) -> c_int {
```
davidhewitt, line 96: 👍

davidhewitt (approved):
Thanks for finishing this off.

PR #1437, opened 2021-02-20
Refactor Python scripts in build.rs

kngwyu:
I also considered placing these scripts in separate files and applying black, but it seemed overkill.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -318,18 +318,18 @@ fn parse_sysconfigdata(config_path: impl AsRef<Path>) -> Result<HashMap<String, 
     script += r#"
 print("version_major", build_time_vars["VERSION"][0])  # 3
 print("version_minor", build_time_vars["VERSION"][2])  # E.g., 8
-if "WITH_THREAD" in build_time_vars:
-    print("WITH_THREAD", build_time_vars["WITH_THREAD"])
-if "Py_TRACE_REFS" in build_time_vars:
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -690,30 +690,34 @@ fn find_interpreter_and_get_config() -> Result<(InterpreterConfig, BuildFlags)>
...
 print("version_major", sys.version_info[0])
 print("version_minor", sys.version_info[1])
 print("implementation", platform.python_implementation())
 if libdir is not None:
     print("libdir", libdir)
-print("ld_version", sysconfig.get_config_var('LDVERSION') or sysconfig.get_config_var('py_version_short'))
+print(
+    "ld_version",
+    ld_version or sysconfig.get_config_var("py_version_short"),
+)
```
kngwyu, line 717: I formatted this by black with its default line width (=88).

davidhewitt: Might be able to reduce line width for this and the lines below by caching `sysconfig.get_config_var`, e.g.:
```suggestion
var = sysconfig.get_config_var
print("ld_version", ld_version or var("py_version_short"))
```

davidhewitt (approved):
👍 thanks for cleaning these up! Just one optional suggestion.

build.rs
kngwyu: Thanks, I choose `from sysconfig import get_config_var`.

davidhewitt:
Even better! Thanks 🚀

Issue #1439, opened 2021-02-20
Support positional-only arguments


davidhewitt:
I noticed this morning that we don't yet have support for Python 3.8's positional-only arguments (PEP 570). This should be a nice standalone feature which anyone interested is welcome to implement. I can help mentor with any questions.

Some examples of how they interact:

```
>>> def foo(a, /, b):
...     print(a, b)
...
>>> foo(1, 3)
1 3

>>> foo(a=1, b=2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: foo() got some positional-only arguments passed as keyword arguments: 'a'

# Also combines with keyword-only separator
>>> def foo(a, /, *, b):
...     print(a, b)
...
>>> foo(1, 3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: foo() takes 1 positional argument but 2 were given
```

This will need changes in a couple of places:
 - The proc macro code will need to be extended to support positional-only arguments. A starting-point would be to teach [add_literal](https://github.com/PyO3/pyo3/blob/f1fc17cbf0624e74ae01168b8d94d3e551ff35a5/pyo3-macros-backend/src/pyfunction.rs#L66) which already parses `"*"` to also parse `"/"`.
 - The runtime code to [parse_fn_args](https://github.com/PyO3/pyo3/blob/f1fc17cbf0624e74ae01168b8d94d3e551ff35a5/src/derive_utils.rs#L35) will need adjusting to handle these positional-only arguments.

Both parts of the code are welcome to be refactored at the same time to deal with the additional complexity.

davidhewitt:
The argument extraction was updated in #1440 so now we just need to add this to our macros!

Issue #661, opened 2019-11-06
Class creation is slow


ethanhs:
My environment is rustc 1.40, python 3.8, Windows 10.

### 💥 Reproducing

Make a `pyclass` and benchmark it!

I realize this could be a quirk of my current setup, but I was doing some benchmarking, and realized that it can take up to 440ns to create a Python class object. This is fine in isolation, but when performance matters, and you are making a lot of these objects (or want to make them as fast as possible) it matters. I created a simple Cython test file to compare, and it was more than 2x faster in creating the class (~200ns).

Here is the sample Cython code:
```cython
# cython: language_level=3
cdef class Example:
    def __cinit__(self):
        self.a = 4
        self.b = "Test"
    cdef public long a
    cdef public str b
```
You can compile an extension from it with `cythonize -b test.pyx`.

The Rust code is the obvious:
```Rust
use pyo3::prelude::*;

#[pyclass]
pub struct Example {
    #[pyo3(get, set)]
    pub a: i64,
    #[pyo3(get, set)]
    pub b: String,
}

#[pymethods]
impl Example {
    #[new]
    fn new(obj: &PyRawObject) {
        obj.init({
            Example {
                a: 4,
                b: String::from("Hello"),
            }
        });
    }
}

#[pymodule]
fn example(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<Example>()?;

    Ok(())
}
```

I realize this is probably not the biggest priority, but it seems odd that class creation would be so much slower.

ehiggs:
Could you post your rust build commands and commands for how you benchmarked this?

Thanks

ethanhs:
Sure! I just used `cargo build --release` (I'd have used target-cpu=native but I'm getting a heap corruption ICE if I do that). Then I write a simple benchmark using pytest-benchmark like the following:

```python
import example
def test_class_creation_rust(benchmark):
    benchmark(example.Example)
```

This calls `example.Example()` which is the best benchmark I could think of atm.

kngwyu:
Thank you for investingating that, but...
PyO3 has some *fundamental* overhead in its core, so I'm not sure we can fix it :thinking:

ethanhs:
Ah I see, that's too bad :/

I was under the assumption pyo3 was zero cost. Thank you for the information!

ijl:
Why does pyo3 track allocations and use PyRef rather than just increment and decrement the PyObject's reference count?

kngwyu:
[...]

PR #1440, opened 2021-02-20
pyfunction: refactor argument extraction

davidhewitt:
This PR fixes a bug in `#[pyfunction]` where a keyword-only argument is incorrectly rejected when used in combination with `*args`:

This is an example of the problem code:

```
#[pyfunction(args = "*", param)]
fn my_func(
    args: &PyTuple,
    param: bool,
) -> PyResult<(&PyTuple, bool)> {
    Ok((args, param))
}
```

And this is the call which didn't work:
```
>>> my_func(1, param=True)
TypeError: my_func() got multiple values for argument: param
```

At the same time I also changed error messages from `argument: foo` to `argument 'foo'`, so that they match what CPython errors look like.

### DIFF
--- a/examples/pyo3_benchmarks/src/lib.rs
+++ b/examples/pyo3_benchmarks/src/lib.rs
@@ -1,0 +1,33 @@
+use pyo3::prelude::*;
+use pyo3::types::{PyDict, PyTuple};
+use pyo3::wrap_pyfunction;
+
+#[pyfunction(args = "*", kwargs = "**")]
+fn args_and_kwargs<'a>(
+    args: &'a PyTuple,
+    kwargs: Option<&'a PyDict>,
+) -> (&'a PyTuple, Option<&'a PyDict>) {
+    (args, kwargs)
+}
+
+#[pyfunction(a, b = 2, args = "*", c = 4, kwargs = "**")]
+fn mixed_args<'a>(
+    a: i32,
+    b: i32,
+    args: &'a PyTuple,
+    c: i32,
+    kwargs: Option<&'a PyDict>,
+) -> (i32, i32, &'a PyTuple, i32, Option<&'a PyDict>) {
+    (a, b, args, c, kwargs)
+}
+
+#[pyfunction]
+fn no_args() {}
+
+#[pymodule]
+fn _pyo3_benchmarks(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
+    m.add_function(wrap_pyfunction!(args_and_kwargs, m)?)?;
+    m.add_function(wrap_pyfunction!(mixed_args, m)?)?;
+    m.add_function(wrap_pyfunction!(no_args, m)?)?;
+    Ok(())
+}
--- a/examples/rustapi_module/tests/test_datetime.py
+++ b/examples/rustapi_module/tests/test_datetime.py
@@ -143,12 +143,11 @@ def test_time_fold(fold):
     assert t.fold == fold
 
 
-@pytest.mark.xfail
 @pytest.mark.parametrize(
     "args", [(-1, 0, 0, 0), (0, -1, 0, 0), (0, 0, -1, 0), (0, 0, 0, -1)]
 )
-def test_invalid_time_fails_xfail(args):
-    with pytest.raises(ValueError):
+def test_invalid_time_fails_overflow(args):
+    with pytest.raises(OverflowError):
         rdt.make_time(*args)
 
 
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -4,8 +4,8 @@ use crate::konst::ConstSpec;
 use crate::method::{FnArg, FnSpec, FnType, SelfType};
 use crate::utils;
 use proc_macro2::{Span, TokenStream};
-use quote::quote;
-use syn::{ext::IdentExt, spanned::Spanned};
+use quote::{quote, quote_spanned};
+use syn::{ext::IdentExt, spanned::Spanned, Result};
 
 pub enum PropertyType<'a> {
     Descriptor(&'a syn::Field),
@@ -22,29 +22,29 @@ pub fn gen_py_method(
     cls: &syn::Type,
     sig: &mut syn::Signature,
     meth_attrs: &mut Vec<syn::Attribute>,
-) -> syn::Result<GeneratedPyMethod> {
+) -> Result<GeneratedPyMethod> {
     check_generic(sig)?;
     let spec = FnSpec::parse(sig, &mut *meth_attrs, true)?;
 
     Ok(match &spec.tp {
         FnType::Fn(self_ty) => GeneratedPyMethod::Method(impl_py_method_def(
             &spec,
-            &impl_wrap(cls, &spec, self_ty, true),
+            &impl_wrap(cls, &spec, self_ty, true)?,
         )),
         FnType::FnNew => {
-            GeneratedPyMethod::New(impl_py_method_def_new(cls, &impl_wrap_new(cls, &spec)))
+            GeneratedPyMethod::New(impl_py_method_def_new(cls, &impl_wrap_new(cls, &spec)?))
         }
         FnType::FnCall(self_ty) => GeneratedPyMethod::Call(impl_py_method_def_call(
             cls,
-            &impl_wrap(cls, &spec, self_ty, false),
+            &impl_wrap(cls, &spec, self_ty, false)?,
         )),
         FnType::FnClass => GeneratedPyMethod::Method(impl_py_method_def_class(
             &spec,
-            &impl_wrap_class(cls, &spec),
+            &impl_wrap_class(cls, &spec)?,
         )),
         FnType::FnStatic => GeneratedPyMethod::Method(impl_py_method_def_static(
             &spec,
-            &impl_wrap_static(cls, &spec),
+            &impl_wrap_static(cls, &spec)?,
         )),
         FnType::ClassAttribute => GeneratedPyMethod::Method(impl_py_method_class_attribute(
             &spec,
@@ -98,7 +98,7 @@ pub fn impl_wrap(
     spec: &FnSpec<'_>,
     self_ty: &SelfType,
     noargs: bool,
-) -> TokenStream {
+) -> Result<TokenStream> {
     let body = impl_call(cls, &spec);
     let slf = self_ty.receiver(cls);
     impl_wrap_common(cls, spec, noargs, slf, body)
@@ -110,10 +110,10 @@ fn impl_wrap_common(
     noargs: bool,
     slf: TokenStream,
     body: TokenStream,
-) -> TokenStream {
+) -> Result<TokenStream> {
     let python_name = &spec.python_name;
     if spec.args.is_empty() && noargs {
-        quote! {
+        Ok(quote! {
             unsafe extern "C" fn __wrap(
                 _slf: *mut pyo3::ffi::PyObject,
                 _args: *mut pyo3::ffi::PyObject,
@@ -126,11 +126,10 @@ fn impl_wrap_common(
                     pyo3::callback::convert(_py, #body)
                 })
             }
-        }
+        })
     } else {
-        let body = impl_arg_params(&spec, Some(cls), body);
-
-        quote! {
+        let body = impl_arg_params(&spec, Some(cls), body)?;
+        Ok(quote! {
             unsafe extern "C" fn __wrap(
                 _slf: *mut pyo3::ffi::PyObject,
                 _args: *mut pyo3::ffi::PyObject,
@@ -146,18 +145,22 @@ fn impl_wrap_common(
                     pyo3::callback::convert(_py, #body)
                 })
             }
-        }
+        })
     }
 }
 
 /// Generate function wrapper for protocol method (PyCFunction, PyCFunctionWithKeywords)
-pub fn impl_proto_wrap(cls: &syn::Type, spec: &FnSpec<'_>, self_ty: &SelfType) -> TokenStream {
+pub fn impl_proto_wrap(
+    cls: &syn::Type,
+    spec: &FnSpec<'_>,
+    self_ty: &SelfType,
+) -> Result<TokenStream> {
     let python_name = &spec.python_name;
     let cb = impl_call(cls, &spec);
-    let body = impl_arg_params(&spec, Some(cls), cb);
+    let body = impl_arg_params(&spec, Some(cls), cb)?;
     let slf = self_ty.receiver(cls);
 
-    quote! {
+    Ok(quote! {
         #[allow(unused_mut)]
         unsafe extern "C" fn __wrap(
             _slf: *mut pyo3::ffi::PyObject,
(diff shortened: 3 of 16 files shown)


davidhewitt:
cc @1frag thanks for reporting on Gitter

davidhewitt:
Given the clippy failure I might rework this PR to also provide a solution for #1439 at the same time.

davidhewitt:
I've refactored the argument extraction in `derive_utils` to be future-compatible with keyword-only arguments. I'm happy with the overall implementation I've written, however I think there's a couple of kinks that need to be worked out before this is review-worthy.

davidhewitt:
Ok, tests all passing, and I added a benchmark for curiosity. Performance hasn't really changed versus the original implementation. However the new implementation seems easier to maintain to me - it should be much easier to add support for vectorcall (#684) and positional-only arguments (#1439).

examples/rustapi_module/tests/test_misc.py
```diff
@@ -4,3 +4,19 @@
...
+def test_args_and_kwargs(benchmark):
+    benchmark(rustapi_module.misc.args_and_kwargs, 1, 2, 3, a=4, foo=10)
+
+
+def test_mixed_args(benchmark):
+    benchmark(rustapi_module.misc.mixed_args, 1, 2, 3, bar=4, foo=10)
+
+
+def mixed_args_py(a, b=2, *args, c=4, **kwargs):
+    (a, b, args, c, kwargs)
```
1frag, line 18: Should there be a `return` here?

davidhewitt: It'll probably not change the benchmark much, but yes, to exactly match the Rust benchmark we want this here. Thanks for reading 👍

examples/rustapi_module/tests/test_datetime.py
```diff
@@ -143,12 +143,11 @@ def test_time_fold(fold):
     assert t.fold == fold
 
 
-@pytest.mark.xfail
 @pytest.mark.parametrize(
     "args", [(-1, 0, 0, 0), (0, -1, 0, 0), (0, 0, -1, 0), (0, 0, 0, -1)]
 )
-def test_invalid_time_fails_xfail(args):
-    with pytest.raises(ValueError):
+def test_invalid_time_fails_overfloaw(args):
```
1frag, line 149: Typo in `overflow`.

src/derive_utils.rs
```diff
@@ -252,3 +395,21 @@ impl<'a> From<&'a PyModule> for PyFunctionArguments<'a> {
         PyFunctionArguments::PyModule(module)
     }
 }
+
+fn write_parameter_list(msg: &mut String, parameter_names: &[&str]) {
```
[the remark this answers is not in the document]
davidhewitt: Do you mean call the function `push_parameter_list`?

Agreed will definitely add a test 👍

src/derive_utils.rs
```diff
@@ -252,3 +395,21 @@ impl<'a> From<&'a PyModule> for PyFunctionArguments<'a> {
...
+    let mut num_written = 0;
```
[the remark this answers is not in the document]
davidhewitt: Good spot, I needed it previously before I did some other refactoring!

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -415,15 +419,23 @@ pub fn impl_arg_params(
...
 
-        params.push(quote! {
-            pyo3::derive_utils::ParamDescription {
-                name: #name,
-                is_optional: #opt,
-                kw_only: #kwonly
+        if kwonly {
+            if !opt {
+                required_keyword_only_parameters.push(params.len());
             }
```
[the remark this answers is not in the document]
davidhewitt: Actually keyword-only arguments can come in any order after the varargs separator. They're not always default, for example consider this function:

```
>>> def foo(*, a=1, b, c, d=2, e):
...     ...
...
>>> foo()
TypeError: foo() missing 3 required keyword-only arguments: 'b', 'c', and 'e'
```

However I think I might do some refactoring here to make keyword-only arguments a separate slice...

examples/rustapi_module/tests/test_datetime.py
davidhewitt: Thanks!

davidhewitt:
Thanks for the reviews. I pushed a couple of commits addressing the comments and making the keyword arguments portion clearer (hopefully).

kngwyu:
I'm sorry, but I'm struggling to understand all changes. Comments may delay.

davidhewitt:
No problem. Feel free to ask as many questions on the code as necessary. I might need to add better inline comments & documentation.

kngwyu:
Just curious, why did you change the current
```python
for param in positional_params + keyword_params:
    if param in kwargs:
        ...
```
loop to
```python
for name, value in kwargs:
    if name in positional_params:
        ....
    elif name in keyword_params:
        ....
```
?

davidhewitt:
Great question. There's a few reasons I did this, however I am happy to change this design:
- This is [exactly the same as what cpython does](https://github.com/python/cpython/blob/cc3df6368d4f3f6c9c9b716876c7e7b79c7abf3f/Python/ceval.c#L4490)
- When we implement `Vectorcall` support then we get keyword names as an array, and we don't have a dict/hashmap to look up any more in the way the code on master currently does.
- Looking up the parameter name in the kwargs dict requires making a new `PyString` for each lookup. With the new way, we don't make any allocations.

I know the algorithmic complexity of the new method looks worse. For most functions we're never going to have more than 10 parameters, so the fact that the new algorithm is O(n) whereas the old one was O(1) matters less. The cost of allocating thePyString (or even for Vectorcall, a whole new `HashMap`) is relevant.

I'm primarily motivated by benchmarks - in the end we should pick whichever algorithm is fastest. But I think we should first implement vectorcall and the owned objects API, because they will also have big performance effects. So for now I wanted to keep the algorithm the same as cpython for simplicity of maintenance.

kngwyu:
Thank you for your explanation. I don't care about the computational complexity here, let's choose a simpler one :)

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -491,15 +497,31 @@ fn impl_arg_param(
     };
 
     if spec.is_args(&name) {
-        return quote! {
-            let #arg_name = <#ty as pyo3::FromPyObject>::extract(_args.as_ref())
-                .map_err(#transform_error)?;
+        return if arg.optional.is_some() {
```
kngwyu, line 500: Does this mean we'll start supporting `args: Option<&PyTuple>`?
Then let's add a test and document for that.

src/derive_utils.rs
```diff
@@ -10,106 +10,245 @@ use crate::exceptions::PyTypeError;
...
+
+        // Check that there's sufficient positional arguments once keyword arguments are specified
+        if args_provided < self.required_positional_parameters {
+            let missing_positional_arguments: Vec<_> = self.positional_parameter_names
+                [..self.required_positional_parameters]
+                .iter()
+                .copied()
+                .zip(output.iter())
+                .filter_map(|(param, out)| if out.is_none() { Some(param) } else { None })
+                .collect();
```
kngwyu, line 98: Just a nit:
```suggestion
            let missing_positional_arguments: Vec<_> = self
                .positional_parameter_names
                .iter()
                .take(self.required_positional_parameters)
                .zip(output.iter())
                .filter_map(|(param, out)| if out.is_none() { Some(*param) } else { None })
                .collect();
```

src/derive_utils.rs
```diff
@@ -10,106 +10,245 @@ use crate::exceptions::PyTypeError;
...
-            Some(kwarg) => {
-                if i < nargs {
-                    raise_error!("got multiple values for argument: {}", p.name)
+
+    #[inline]
+    fn extract_keyword_arguments<'p>(
+        &self,
+        kwargs: &'p PyDict,
+        output: &mut [Option<&'p PyAny>],
+        mut unexpected_keyword_handler: impl FnMut(&'p PyAny, &'p PyAny) -> PyResult<()>,
```
kngwyu, line 132: How about returning `Vec<(&'p PyAny, &'p PyAny)>` instead of taking this closure?
Then we can display all unexpected keyword args.

src/derive_utils.rs
```diff
@@ -10,106 +10,245 @@ use crate::exceptions::PyTypeError;
...
+            };
+
+            // Compare the keyword name against each parameter in turn. This is exactly the same method
+            // which CPython uses to map keyword names. Although it's O(num_parameters), the number of
+            // parameters is expected to be small so it's not worth constructing a mapping.
+            for (param, out) in self.keyword_only_parameters.iter().zip(&mut *kwargs_output) {
+                if utf8_string == param.name {
+                    *out = Some(value);
+                    continue 'kwarg_loop;
                 }
```
kngwyu, line 155: Just a nit:
```suggestion
            if let Some(i) = self
                .keyword_only_parameters
                .iter()
                .position(|p| p.name == utf8_string)
            {
                kwargs_output[i] = Some(value);
                continue;
```

src/derive_utils.rs
```diff
@@ -10,106 +10,245 @@ use crate::exceptions::PyTypeError;
...
+                    if i < self.positional_only_parameters {
+                        positional_only_keyword_arguments.push(param);
+                    } else {
+                        match out {
+                            Some(_) => return Err(self.multiple_values_for_argument(param)),
+                            None => {
+                                *out = Some(value);
+                            }
+                        }
                     }
```
kngwyu, line 175: Same as above, 
```suggestion
            if let Some((i, &param)) = self
                .positional_parameter_names
                .iter()
                .enumerate()
                .find(|(_, &p)| p == utf8_string)
            {
                if i < self.positional_only_parameters {
                    positional_only_keyword_arguments.push(param);
                } else {
                    if args_output[i].is_some() {
                        return Err(self.multiple_values_for_argument(param));
                    }
                    args_output[i] = Some(value);
                }
```

kngwyu (commented):
Thanks, I left some comments.

davidhewitt:
Thanks for the review! I will add the extra tests and try out all the suggested changes over the weekend. As (imo) this is performance-sensitive code I'll try each change in turn and see whether it has any noticeable effect on performance. I suspect that your suggested `.find()` / `.position()` is probably better than using `.zip()`, we'll see ;)

src/derive_utils.rs
davidhewitt: I want to avoid `Vec` in the case where we accept `**kwargs`, because it's an unnecessary allocation. However I agree we can display all unexpected keyword arguments - I'll change the code to do so.

davidhewitt: Actually I've changed my mind on changing what I've currently got - the existing code for printing all unexpected positional arguments relies on having `&str` forms of the parameter names, whereas here we only have `&PyAny` which are not guaranteed to be UTF8.

Also the current form I've implemented is consistent with CPython which only lists the first unexpected keyword argument.

If someone badly wants the full list printed, it can be a follow-up PR 😄

pyo3-macros-backend/src/pymethod.rs
davidhewitt: Actually it's still not possible, because I can't figure out the right macro invocation to get `Option<&PyTuple>` and `Option<Vec>` to work correctly (we currently support e.g. both `&PyTuple` and `Vec`.

I'll add a check that the type is _not_ `Option<_>` for now.

src/derive_utils.rs
kngwyu: > Also the current form I've implemented is consistent with CPython which only lists the first unexpected keyword argument.

OK, that makes sense 👍🏼

src/derive_utils.rs
```diff
@@ -10,106 +10,242 @@ use crate::exceptions::PyTypeError;
...
+    fn extract_keyword_arguments<'p>(
+        &self,
+        kwargs: &'p PyDict,
+        output: &mut [Option<&'p PyAny>],
+        mut unexpected_keyword_handler: impl FnMut(&'p PyAny, &'p PyAny) -> PyResult<()>,
+    ) -> PyResult<()> {
+        let (args_output, kwargs_output) =
+            output.split_at_mut(self.positional_parameter_names.len());
+        let mut positional_only_keyword_arguments = Vec::new();
+        'kwarg_loop: for (kwarg_name, value) in kwargs {
```
kngwyu, line 137: I think this label is unnecessary.

src/derive_utils.rs
```diff
@@ -10,106 +10,242 @@ use crate::exceptions::PyTypeError;
...
 }
 
-/// Parse argument list
-///
-/// * fname:  Name of the current function
-/// * params: Declared parameters of the function
-/// * args:   Positional arguments
-/// * kwargs: Keyword arguments
-/// * output: Output array that receives the arguments.
-///           Must have same length as `params` and must be initialized to `None`.
```
kngwyu, line 34: How about retrieving some of these comments?

examples/pyo3_benchmarks/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+use pyo3::prelude::*;
```
kngwyu, line 1: I'm not sure how this benchmark is useful.
It may be difficult to observe some speedup when trying to extract only a few args.

src/derive_utils.rs
davidhewitt: Good catch; it was needed before the other changes to `position` / `find`!

examples/pyo3_benchmarks/src/lib.rs
davidhewitt: Actually although the variance is quite high, there's still a reasonably clear signal as to how good a change is using these benchmarks. I wouldn't take these numbers super seriously for any kind of scientific conclusion. However I can see (for example) that there's still some overhead calling pyo3 functions vs CPython functions, and also that returning `Vec` from `extract_keyword_arguments` leads to noticeable slowdown.

Some example numbers I see at the moment:

```
-------------------------------------------------------------------------------------------- benchmark: 6 tests -------------------------------------------------------------------------------------------
Name (time in ns)                Min                    Max                Mean              StdDev              Median                IQR             Outliers  OPS (Mops/s)            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_no_args_py              60.0000 (1.0)       4,658.0000 (2.86)      67.6030 (1.0)       34.3033 (1.08)      66.0000 (1.0)       2.0000 (inf)      1857;3556       14.7923 (1.0)      149254         100
test_no_args                 70.0000 (1.17)      1,627.0000 (1.0)       78.7283 (1.16)      31.8182 (1.0)       74.0000 (1.12)      2.0000 (inf)      3261;8374       12.7019 (0.86)     128206         100
test_args_and_kwargs_py     219.0476 (3.65)      5,566.6667 (3.42)     245.2637 (3.63)      57.6773 (1.81)     242.8571 (3.68)      4.7619 (inf)     1156;10254        4.0772 (0.28)     196079          21
test_mixed_args_py          288.2353 (4.80)      6,329.4118 (3.89)     320.9264 (4.75)      83.6458 (2.63)     317.6471 (4.81)      5.8824 (inf)     1406;18940        3.1160 (0.21)     192308          17
test_args_and_kwargs        299.9999 (5.00)     86,200.0001 (52.98)    412.2963 (6.10)     466.4672 (14.66)    400.0000 (6.06)      0.0000 (1.0)      359;47106        2.4254 (0.16)     163935           1
test_mixed_args             340.0000 (5.67)     21,400.0000 (13.15)    385.6509 (5.70)     207.9585 (6.54)     380.0000 (5.76)     15.0000 (inf)      1187;1616        2.5930 (0.18)     128206          20
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

At the moment it's clear we're a bit slower than CPython for function calls, but I hope vectorcall support and #1308 will improve things.

Later I would like to add more numbers to these benchmarks, for example #661.

src/derive_utils.rs
davidhewitt: Good idea.

kngwyu (approved):

kngwyu:
LGTM 💯 , thank you!

PR #1445, opened 2021-02-24
Fix wordcount example docs

Winxton:
Fix unused references in the wordcount example


davidhewitt (approved):
Thank you for cleaning this up!

PR #1446, opened 2021-02-25
pyproto: don't use inventory for methods

davidhewitt:
This is the next step in making inventory optional, by changing `#[pyproto]` methods which use inventory to also use dtolnay specialization.

It works very similarly to #1328.

From a very basic benchmark, this doesn't appear to change performance of type object initialization at all. (Perf doesn't matter very much here anyway, as this is only something that runs once on library startup.)

The next (and final) PR in this series will make it so that `inventory` is optional. Without inventory, you will only be allowed one `#[pymethods]` for each `#[pyclass]`. However you will get faster compile times and support on platforms where inventory isn't available :)

### DIFF
--- a/pyo3-macros-backend/src/module.rs
+++ b/pyo3-macros-backend/src/module.rs
@@ -210,7 +210,7 @@ pub fn add_fn_to_module(
             pyo3::types::PyCFunction::internal_new(
                 name,
                 doc,
-                pyo3::class::PyMethodType::PyCFunctionWithKeywords(#wrapper_ident),
+                unsafe { std::mem::transmute(#wrapper_ident as *const std::os::raw::c_void) },
                 pyo3::ffi::METH_VARARGS | pyo3::ffi::METH_KEYWORDS,
                 args.into(),
             )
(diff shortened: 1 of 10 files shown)


pyo3-macros-backend/src/module.rs
```diff
@@ -210,7 +210,7 @@ pub fn add_fn_to_module(
             pyo3::types::PyCFunction::internal_new(
                 name,
                 doc,
-                pyo3::class::PyMethodType::PyCFunctionWithKeywords(#wrapper_ident),
+                unsafe { std::mem::transmute(#wrapper_ident as *const std::os::raw::c_void) },
```
kngwyu, line 213: Transmuting to `enum` is safe?

kngwyu (approved):
Thanks, I don't see any problem, but maybe it's better we have a comment on `transmute`.

pyo3-macros-backend/src/module.rs
davidhewitt: This is actually a function-pointer-to-function-pointer cast (I changed the arguments to `PyCFunction::internal_new`).

Actually I think I can get rid of this by refactoring `PyCFunction::internal_new` a bit more. It will have a slight user-facing change (but a good one imo) so I'll do it in a follow-up PR.

PR #1450, opened 2021-02-27
pycfunction: take &'static str arguments to new

davidhewitt:
This is a follow-up to #1446 to refactor `PyCFunction` a little bit to re-use code and get rid of one `transmute` call at the same time.

I had to change the `name` and `doc` arguments to `PyCFunction::new()` to take `&'static str` instead of just `&str`. This is a good change in my opinion because the underlying Python APIs need `'static` data anyway. Previously we always had to copy the string and leak the copy; now we can forward the pointer to the underlying string (as long as it's a valid c-string).

### DIFF
--- a/src/class/methods.rs
+++ b/src/class/methods.rs
@@ -1,7 +1,7 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
 use crate::{ffi, PyObject, Python};
-use std::ffi::CStr;
+use std::ffi::{CStr, CString};
 use std::fmt;
 use std::os::raw::c_int;
 
@@ -73,15 +73,6 @@ unsafe impl Sync for PyGetterDef {}
 
 unsafe impl Sync for PySetterDef {}
 
-fn get_name(name: &str) -> &CStr {
-    CStr::from_bytes_with_nul(name.as_bytes())
-        .expect("Method name must be terminated with NULL byte")
-}
-
-fn get_doc(doc: &str) -> &CStr {
-    CStr::from_bytes_with_nul(doc.as_bytes()).expect("Document must be terminated with NULL byte")
-}
-
 impl PyMethodDef {
     /// Define a function with no `*args` and `**kwargs`.
     pub const fn cfunction(name: &'static str, cfunction: PyCFunction, doc: &'static str) -> Self {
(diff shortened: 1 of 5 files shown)


src/class/methods.rs
```diff
@@ -209,3 +200,25 @@ pub trait PyMethodsInventory: inventory::Collect {
 pub trait HasMethodsInventory {
     type Methods: PyMethodsInventory;
 }
+
+#[derive(Debug)]
+pub struct NulByteInString(pub(crate) &'static str);
+
+fn get_name(name: &'static str) -> Result<&'static CStr, NulByteInString> {
+    extract_cstr_or_leak_cstring(name, "Function name cannot contain contain NULL byte.")
```
messense, line 208:
```suggestion
    extract_cstr_or_leak_cstring(name, "Function name cannot contain NUL byte.")
```

src/class/methods.rs
```diff
@@ -209,3 +200,25 @@ pub trait PyMethodsInventory: inventory::Collect {
...
+}
+
+fn get_doc(doc: &'static str) -> Result<&'static CStr, NulByteInString> {
+    extract_cstr_or_leak_cstring(doc, "Document cannot contain contain NULL byte.")
```
messense, line 212:
```suggestion
    extract_cstr_or_leak_cstring(doc, "Document cannot contain NUL byte.")
```

src/class/methods.rs
```diff
@@ -209,3 +200,25 @@ pub trait PyMethodsInventory: inventory::Collect {
 pub trait HasMethodsInventory {
     type Methods: PyMethodsInventory;
 }
+
+#[derive(Debug)]
+pub struct NulByteInString(pub(crate) &'static str);
```
kngwyu, line 205: Do we really want to expose this as a public error type?
I think `std::ffi::NulError` is better at least for `as_method_def`, though then we need a bit more glue codes to display specific error messages.

kngwyu (commented):
Thanks, I think it's almost OK.

src/class/methods.rs
davidhewitt: Actually I think that `as_method_def` can now be `pub (crate)` - so I'll apply that and also to this error type.

kngwyu: Sounds good

davidhewitt:
Changes pushed - thanks both for the reviews!

messense (approved):

PR #1454, opened 2021-03-01
tuples: support conversions up to length 12

davidhewitt:
The stdlib supports tuples up to length 12, so I think we should too. https://doc.rust-lang.org/std/primitive.tuple.html#trait-implementations-1


kngwyu (approved):
> The stdlib supports tuples up to length 12, 

That makes sense, thanks!

PR #1455, opened 2021-03-01
Generate less code

konstin:
When building a test crate without the macros feature for maturin, I saw that pyo3 generates a lot of identical boilerplate code. This removes a bunch of it, mainly by turning `callback_body_without_convert` into a function. 

I've used this code to test:

```rust
use pyo3::prelude::*;
use pyo3::wrap_pyfunction;

#[pyfunction]
fn add(x: usize, y: usize) -> usize {
    x + y
}

#[pyclass]
struct Data {
    pub x: usize,
}

#[pymodule]
fn minimal(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(add, m)?)?;
    m.add_class::<Data>()?;

    Ok(())
}
```

And this is the resulting diff: https://gist.github.com/konstin/3c9e6b9047e2aa06a4cc47d6a235cff4 (with `cargo +nightly rustc -- -Zunstable-options --pretty=expanded` and after rustfmt)

### DIFF
--- a/src/callback.rs
+++ b/src/callback.rs
@@ -216,7 +216,7 @@ macro_rules! callback_body {
 /// ```ignore
 /// pyo3::callback::handle_panic(|_py| {
 ///     let _slf = #slf;
-///     pyo3::callback::convert(py, #foo)
+///     pyo3::callback::convert(_py, #foo)
 /// })
 /// ```
 ///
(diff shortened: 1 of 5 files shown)


src/callback.rs
```diff
@@ -231,33 +232,33 @@ macro_rules! callback_body {
...
-        let pool = $crate::GILPool::new();
-        let unwind_safe_py = std::panic::AssertUnwindSafe(pool.python());
-        let result = match std::panic::catch_unwind(move || -> $crate::PyResult<_> {
-            let $py = *unwind_safe_py;
-            $body
-        }) {
+pub unsafe fn handle_panic<
+    R: PyCallbackOutput,
+    F: FnOnce(Python) -> crate::PyResult<R> + UnwindSafe,
+>(
```
kngwyu, line 238: Maybe `where` clause would be better for such long type parameter definitions.

kngwyu (approved):
Solid PR, thanks!

davidhewitt (approved):
Seems reasonable to me. IIRC I made a macro originally to avoid function call overhead, but tbh we probably should not worry about that and the optimizer might remove it anyway.

There's benchmarks beginning to be written in #1440 but I doubt that they'll show anything from this change.

src/callback.rs
konstin: good idea

konstin:
Yeah, and if it should actually cause overhead we can still add `#[inline]`

PR #1456, opened 2021-03-02
pyclass: move flags to PyClassImpl

davidhewitt:
This is a (mostly internal) refactoring to move some pieces of `PyTypeInfo` which aren't needed for anything other than `#[pyclass]` onto the `PyClassImpl` trait.

At the same time I cleaned up the old `FLAGS` constant which still had unneeded entries for e.g. weakref.

### DIFF
--- a/src/class/impl_.rs
+++ b/src/class/impl_.rs
@@ -1,6 +1,6 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
-use crate::{derive_utils::PyBaseTypeUtils, ffi, PyMethodDefType, PyNativeType, PyTypeInfo};
+use crate::{derive_utils::PyBaseTypeUtils, ffi, PyMethodDefType, PyNativeType};
 use std::{marker::PhantomData, thread};
 
 /// This type is used as a "dummy" type on which dtolnay specializations are
@@ -31,9 +31,9 @@ impl<T> Copy for PyClassImplCollector<T> {}
 ///
 /// Users are discouraged from implementing this trait manually; it is a PyO3 implementation detail
 /// and may be changed at any time.
-pub trait PyClassImpl: Sized + PyTypeInfo {
+pub trait PyClassImpl: Sized {
     /// Class doc string
-    const DESCRIPTION: &'static str = "\0";
+    const DOC: &'static str = "\0";
 
     /// #[pyclass(gc)]
     const IS_GC: bool = false;
--- a/src/class/impl_.rs
+++ b/src/class/impl_.rs
@@ -1,6 +1,6 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
-use crate::{derive_utils::PyBaseTypeUtils, ffi, PyMethodDefType, PyNativeType};
+use crate::{derive_utils::PyBaseTypeUtils, ffi, PyMethodDefType, PyNativeType, PyTypeInfo};
 use std::{marker::PhantomData, thread};
 
 /// This type is used as a "dummy" type on which dtolnay specializations are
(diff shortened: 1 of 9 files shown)


src/pyclass.rs
```diff
@@ -153,6 +153,7 @@ pub trait PyClass:
 struct TypeSlots(Vec<ffi::PyType_Slot>);
 
 impl TypeSlots {
+    #[inline]
```
kngwyu, line 156: Is this effective?
This is not a public function so the compiler would automatically make this inline if needed.

src/pyclass.rs
```diff
@@ -174,6 +175,7 @@ fn get_type_name<T: PyTypeInfo>(module_name: Option<&str>) -> PyResult<*mut c_ch
     })
 }
 
+#[inline]
```
kngwyu, line 178: Same as above.

src/class/impl_.rs
```diff
@@ -31,7 +31,19 @@ impl<T> Copy for PyClassImplCollector<T> {}
 ///
 /// Users are discouraged from implementing this trait manually; it is a PyO3 implementation detail
 /// and may be changed at any time.
-pub trait PyClassImpl: Sized {
+pub trait PyClassImpl: Sized + PyTypeInfo {
```
kngwyu, line 34: I'm not sure this reduces the complexity of our (somewhat complex) trait hierarchy...

kngwyu (commented):
Thanks!
Looks like reasonable changes, but why `PyClassImpl` instead of `PyClass`?
I understand that `PyClassImpl` is for collecting methods by pseudo-specialization trick, so I don't see the necessity to place these flags under `PyClassImpl`.

src/pyclass.rs
davidhewitt: I felt it couldn't hurt to add this attribute here. I can try and measure the benchmarks to see if I can get any meaningful difference.

src/class/impl_.rs
davidhewitt: I agree it's sad, this is because `PyClassAlloc` wanted `PyTypeInfo`. Maybe I can do some other adjusting.

davidhewitt:
> Looks like reasonable changes, but why `PyClassImpl` instead of `PyClass`?

Overall I see `PyClassImpl` as the pieces of `PyClass` which we don't consider stable API (it's `#[doc(hidden)]`). So unless we are sure this is useful for users and it's not going to need to change, it should go on `PyClassImpl`.

To reduce the complex trait hierarchy I was thinking to merge `PyClassAlloc` into `PyClassImpl` also.

kngwyu:
How about moving some flags under `PyClassAlloc`?
They are actually related only to allocation.

davidhewitt:
> How about moving some flags under PyClassAlloc? They are actually related only to allocation.

Only the `IS_SUBCLASS` flag is used during the `default_new` implementation - the others are used in `py_class_flags` which isn't anything to do with allocation.

I'll have a go a follow-up commit to try refactoring `PyClassAlloc` a bit. 👍

src/pyclass.rs
davidhewitt: Okay, removed!

davidhewitt:
I've pushed a commit which removes the `#[inline]` and puts `PyTypeInfo` bound onto `PyClassAlloc` instead of `PyClassImpl`. Also renamed `T::DESCRIPTION` to `T::DOC` for simplification.

I looked at doing further work to `PyClassAlloc` in this PR but realised there are a lot of complications with that trait as-is, especially with #991.

So I will try and push something experimental in a separate PR later.

kngwyu (approved):
Nice, thank you!

PR #1457, opened 2021-03-02
pymethods: make inventory optional

davidhewitt:
With this PR, `inventory` is opt-in (via the `multiple-pymethods` feature) and only needed if you want to have multiple `#[pymethods]` for the same `#[pyclass]`.

This is breaking, so we could make it opt-out. However I think most users won't use multiple `#[pymethods]` per class so will enjoy the reduced dependencies & faster compile times without noticing anything changed 😄 

I plan to push a number of updates to the guide before this PR is ready to be merged. (e.g. information about the feature, migration guide etc.)

### DIFF
--- a/src/class/methods.rs
+++ b/src/class/methods.rs
@@ -1,7 +1,8 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
+use crate::internal_tricks::{extract_cstr_or_leak_cstring, NulByteInString};
 use crate::{ffi, PyObject, Python};
-use std::ffi::{CStr, CString};
+use std::ffi::CStr;
 use std::fmt;
 use std::os::raw::c_int;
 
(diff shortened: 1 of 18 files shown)


davidhewitt:
After #1458 is merged I will rebase on it, which should fix the issues I put into the CI workflow 😄

src/class/methods.rs
```diff
@@ -29,16 +30,22 @@ pub enum PyMethodType {
     PyCFunctionWithKeywords(PyCFunctionWithKeywords),
 }
 
-// These two newtype structs serve no purpose other than wrapping the raw ffi types (which are
-// function pointers) - because function pointers aren't allowed in const fn, but types wrapping
-// them are!
+// These newtype structs serve no purpose other than wrapping which are function pointers - because
```
kngwyu, line 33: ?
```suggestion
// These newtype structs serve no purpose other than wrapping (raw ffi) function pointers - because
```

kngwyu (approved):
LGTM, thanks!

src/class/methods.rs
davidhewitt: Actually the `ClassAttributeFactory` is a Rust function pointer that we use internally and not for ffi :)

davidhewitt:
I've pushed a bunch of documentation tweaks. Unless any changes are needed to those, this PR should be good to merge!

PR #1458, opened 2021-03-02
ci: simplify workflow

davidhewitt:
This extracts just the ci workflow file changes from #1457.

Instead of having separate build & test stages, this now just runs tests for various feature combinations. On pypy this command is overridden to just `build --lib --tests`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,32 +84,43 @@ jobs:
         name: Prepare LD_LIBRARY_PATH (Ubuntu only)
         run: echo LD_LIBRARY_PATH=${pythonLocation}/lib >> $GITHUB_ENV
 
+      - name: Prepare workflow settings
+        id: settings
+        shell: bash
+        run: |
+          echo "::set-output name=all_additive_features::macros num-bigint num-complex hashbrown serde"
+
       - name: Build docs
-        run: cargo doc --no-default-features --features "macros num-bigint num-complex hashbrown" --verbose --target ${{ matrix.platform.rust-target }}
+        run: cargo doc --no-default-features --features "${{ steps.settings.outputs.all_additive_features }}"
 
       - name: Build (no features)
-        run: cargo build --no-default-features --verbose --target ${{ matrix.platform.rust-target }}
+        run: cargo build --lib --tests --no-default-features
+
+      # Run tests (except on PyPy, because no embedding API).
+      - if: matrix.python-version != 'pypy-3.6'
+        name: Test (no features)
+        run: cargo test --no-default-features
 
       - name: Build (all additive features)
-        run: cargo build --no-default-features --features "macros num-bigint num-complex hashbrown serde" --verbose --target ${{ matrix.platform.rust-target }}
+        run: cargo build --lib --tests --no-default-features --features "${{ steps.settings.outputs.all_additive_features }}"
 
       # Run tests (except on PyPy, because no embedding API).
       - if: matrix.python-version != 'pypy-3.6'
         name: Test
-        run: cargo test --no-default-features --features "macros num-bigint num-complex hashbrown serde" --target ${{ matrix.platform.rust-target }}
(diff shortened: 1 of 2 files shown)


kngwyu:
Thanks, but ... 🤔,  is it good to use environmental variables to shorten commands?
I'm not familiar with CI good practices, though.

davidhewitt:
I've pushed a commit to use "outputs" of a github actions step, maybe you prefer that?

I wanted to use variables of some form because it simplifies maintenance of the CI scripts - every time we add a new feature, there were a lot of places to edit. Also the various cargo commands were getting inconsistent with PyPy. Now the structure is more consistent on all CI runs, and there's only one place to edit if we change a new feature.

kngwyu:
> I've pushed a commit to use "outputs" of a github actions step, maybe you prefer that?

I'm sorry but I'm thinking that Github actions might provide something better than `env`... 
But it looks like that `env` is what we want for `all-additive-features`: https://docs.github.com/en/actions/reference/environment-variables

davidhewitt:
Ah ok, I'll revert that last commit? Always good to try things 😄

.github/workflows/ci.yml
```diff
@@ -84,32 +84,38 @@ jobs:
...
-        run: cargo build --no-default-features --features "macros num-bigint num-complex hashbrown serde" --verbose --target ${{ matrix.platform.rust-target }}
+      - name: Test (no features)
+        run: ${{ steps.settings.outputs.cargo_command }}
 
-      # Run tests (except on PyPy, because no embedding API).
-      - if: matrix.python-version != 'pypy-3.6'
-        name: Test
-        run: cargo test --no-default-features --features "macros num-bigint num-complex hashbrown serde" --target ${{ matrix.platform.rust-target }}
+      - name: Test (all additive features)
+        run: ${{ steps.settings.outputs.cargo_command }} --features "${{ steps.settings.outputs.all_additive_features }}"
```
kngwyu, line 109: How about something like
```yaml
       run: |
          if [[ $PYPY == "true" ]]; then
              cargo build --no-default-features --features "${{ steps.settings.outputs.all_additive_features }}" --target ${{ matrix.platform.rust-target }}
             cargo test --no-default-features --features "abi3 ${{ steps.settings.outputs.all_additive_features }}" --target ${{ matrix.platform.rust-target }}
          else
              cargo test --no-default-features --features "abi3 ${{ steps.settings.outputs.all_additive_features }}" --target ${{ matrix.platform.rust-target }}
              cargo test --no-default-features --features "abi3 ${{ steps.settings.outputs.all_additive_features }}" --target ${{ matrix.platform.rust-target }}
...
          fi
```

davidhewitt: I like having the different feature combinations as different steps, because it makes it much easier to see which combinations of features went wrong.

kngwyu:
I noticed that we can use `with` instead of env: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepswith, but... I really don't know which is better. They look the same 😕

build.rs
```diff
@@ -618,7 +618,11 @@ fn get_rustc_link_lib(config: &InterpreterConfig) -> String {
             // See https://www.python.org/dev/peps/pep-0384/#linkage
             //
             // This contains only the limited ABI symbols.
-            if env::var_os("CARGO_FEATURE_ABI3").is_some() {
+            if env::var_os("CARGO_FEATURE_ABI3").is_some()
+                // FIXME: PyPy doesn't currently ship a stable ABI dll.
```
kngwyu, line 622: Does it better to abort the build with some warnings?

kngwyu:
I think we can only refactor `all-additive-features` and leave the PyPy conditioning as is. Since we cannot build the crate with PyPy and abi3, I don't see any strong motivation to commonize them.

davidhewitt:
Ok will push a commit to rework this PR later today. Thanks for looking at this with me 👍

davidhewitt:
I'm gonna merge this so that I can rebase #1457 on it. Hope that's alright.

PR #1464, opened 2021-03-03
Update setuptools-rust/maturin comparison

konstin:
I tried to make differences clearer and make it a bit less biased towards maturin


kngwyu (approved):
Looks better, thank you!

PR #1467, opened 2021-03-04
PyModule: rename call* to call_function*

birkenfeld:
For consistency with PyObject/PyAny, where call() means to call the object itself.


davidhewitt (approved):
Seems like a good idea to me. I can't think of a better name.

Needs a CHANGELOG entry please!

kngwyu (approved):
Thanks 👍🏼

PR #1469, opened 2021-03-05
Show warning when abi3 is used with PyPy

kngwyu:
Thanks @davidhewitt  for contacting to PyPy maintainers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Packaging
+- The `abi3` feature shows a warning when building for `PyPy`. [#1469](https://github.com/PyO3/pyo3/pull/1469)
+
 ### Added
 - Add conversions between `OsStr`/`OsString`/`Path`/`PathBuf` and Python strings. [#1379](https://github.com/PyO3/pyo3/pull/1379)
 - Add #[pyo3(from_py_with = "...")]` attribute for function arguments and struct fields to override the default from-Python conversion. [#1411](https://github.com/PyO3/pyo3/pull/1411)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-### Packaging
-- The `abi3` feature shows a warning when building for `PyPy`. [#1469](https://github.com/PyO3/pyo3/pull/1469)
-
 ### Added
 - Add conversions between `OsStr`/`OsString`/`Path`/`PathBuf` and Python strings. [#1379](https://github.com/PyO3/pyo3/pull/1379)
 - Add #[pyo3(from_py_with = "...")]` attribute for function arguments and struct fields to override the default from-Python conversion. [#1411](https://github.com/PyO3/pyo3/pull/1411)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -145,6 +145,12 @@
 //! }
 //! ```
 
+#[cfg(all(PyPy, Py_LIMITED_API))]
+compile_error!(concat!(
+    "PyPy still does not support abi3. ",
+    "See https://foss.heptapod.net/pypy/pypy/-/issues/3397 for the discussion."
+));
```
davidhewitt, line 152: Just a couple of wording changes which I think makes it read easier:

```suggestion
compile_error!(
    "PyPy does not yet support abi3. \
    See https://foss.heptapod.net/pypy/pypy/-/issues/3397 for more information."
);
```

davidhewitt (commented):
I think we should definitely include this in the `## Packaging` feature at the top of the changelog, because it could break user's builds:

```
### Packaging
- The `abi3` feature is now a compile error when building for `PyPy`. [#1469](https://github.com/PyO3/pyo3/pull/1469)
```

On `setuptools-rust` they'll have to write like this example: https://github.com/PyO3/setuptools-rust/blob/14703b86325e2c53cebec22937e9ca48c7cdeb70/examples/rust_with_cffi/setup.py#L24-L25

konstin:
I think it would be good if there was a way to build for pypy when the abi3 feature is enabled; I'd say the abi3 feature should require that only abi3 apis are used, while it should be possible to build non-abi3 wheels. Otherwise someone who wants to ship wheels for both cpython abi3 and pypy would need to add an optional feature for enabling abi3 and make sure cpython wheels are build with the feature and the pypy wheels without it.

messense:
> I think it would be good if there was a way to build for pypy when the abi3 feature is enabled; I'd say the abi3 feature should require that only abi3 apis are used, while it should be possible to build non-abi3 wheels. Otherwise someone who wants to ship wheels for both cpython abi3 and pypy would need to add an optional feature for enabling abi3 and make sure cpython wheels are build with the feature and the pypy wheels without it.

Example: https://github.com/milesgranger/pyrus-cramjam/blob/37306e46ddb288a164af7d65db16eae3e3ce53ff/Makefile#L37

kngwyu:
Thanks, I changed it to show a warning.
It would look like:
image

In addition, now this PR includes some refactorings of the build script.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Packaging
+- The `abi3` feature shows a warning when building for `PyPy`. [#1469](https://github.com/PyO3/pyo3/pull/1469)
+
```
davidhewitt, line 11: I think now that it's just a warning, there's no need for a CHANGELOG entry. Sorry for my u-turn!

```suggestion
```

build.rs
```diff
@@ -781,14 +792,21 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<()> {
...
     }
 
-    if interpreter_config.version.implementation == PythonInterpreterKind::PyPy {
+    let is_abi3 = is_abi3();
+
+    if interpreter_config.is_pypy() {
         println!("cargo:rustc-cfg=PyPy");
+        if is_abi3 {
+            warn!(
+                "PyPy does not yet support abi3 and the resulting wheel would be version-specific. \
```
davidhewitt, line 801:
```suggestion
                "PyPy does not yet support abi3 so the resulting wheel will be version-specific. \
```

davidhewitt (approved):
LGTM - just a couple of final suggestions. 

The build script refactoring is very nice 👍

davidhewitt:
👍 thanks!

konstin:
Thanks, the warning is a good solution :+1:

PR #1472, opened 2021-03-06
ci: fix nightly warnings

davidhewitt:
This PR fixes the warnings seen in the coverage job, but doesn't fix the one test failure.

It looks like an upstream Rust bug; I was able to isolate the issue to an example which doesn't involve PyO3. https://github.com/rust-lang/rust/issues/82833

I think we might have to just accept the coverage job will be failing for a little while until upstream resolves.


kngwyu:
Thanks!

PR #1473, opened 2021-03-06
ffi: move initconfig.rs to cpython/initconfig.rs

davidhewitt:
No material user-facing changes; just moving this file to the right place 😄

### DIFF
--- a/src/ffi/cpython/pylifecycle.rs
+++ b/src/ffi/cpython/pylifecycle.rs
@@ -1,0 +1,48 @@
+use crate::ffi::{PyConfig, PyPreConfig, PyStatus, Py_ssize_t};
+use libc::wchar_t;
+use std::os::raw::{c_char, c_int};
+
+// "private" functions in cpython/pylifecycle.h accepted in PEP 587
+extern "C" {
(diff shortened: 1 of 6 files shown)


davidhewitt:
... also pushed a commit to update the config structures for what it looks like they'll be in Python 3.10.

cc @nw0 in case you're looking at these files in any of your PRs.

src/ffi/cpython/pylifecycle.rs
```diff
@@ -0,0 +1,48 @@
...
+    ) -> PyStatus;
+    pub fn Py_PreInitializeFromArgs(
+        src_config: *const PyPreConfig,
+        argc: Py_ssize_t,
+        argv: *mut *mut wchar_t,
+    ) -> PyStatus;
+    pub fn _Py_IsCoreInitialized() -> c_int;
+
+    pub fn Py_InitializeFromConfig(config: *const PyConfig) -> PyStatus;
+    // skipped _Py_InitializeMain
```
indygreg, line 22: I'll need this API for PyOxidizer.

davidhewitt: Thanks; added!

PR #1475, opened 2021-03-06
Define missing import APIs

indygreg:
I need some of these APIs for the pyembed crate (used as part of
PyOxidizer). I figured I might as well define them all.

All these APIs are limited.

I don't have a PyPy install with a shared library to test. We may need to disable these when targeting PyPy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
 - Add FFI definition `Py_IS_TYPE`. [#1429](https://github.com/PyO3/pyo3/pull/1429)
 - Add FFI definition `_Py_InitializeMain`. [#1473](https://github.com/PyO3/pyo3/pull/1473)
+- Add FFI definitions from `cpython/import.h`.[#1475](https://github.com/PyO3/pyo3/pull/1475)
 - Add tuple and unit struct support for `#[pyclass]` macro. [#1504](https://github.com/PyO3/pyo3/pull/1504)
 
 ### Changed
--- a/src/ffi/cpython/import.rs
+++ b/src/ffi/cpython/import.rs
@@ -1,0 +1,71 @@
+use crate::ffi::{PyInterpreterState, PyObject};
+use std::os::raw::{c_char, c_int, c_uchar};
+
+// skipped PyInit__imp
+
+extern "C" {
+    pub fn _PyImport_IsInitialized(state: *mut PyInterpreterState) -> c_int;
+    // skipped _PyImport_GetModuleId
+    #[cfg(Py_3_7)]
+    pub fn _PyImport_SetModule(name: *mut PyObject, module: *mut PyObject) -> c_int;
+    #[cfg(Py_3_7)]
+    pub fn _PyImport_SetModuleString(name: *const c_char, module: *mut PyObject) -> c_int;
+    pub fn _PyImport_AcquireLock();
+    pub fn _PyImport_ReleaseLock() -> c_int;
+    #[cfg(not(Py_3_7))]
+    pub fn _PyImport_FindBuiltin(name: *const c_char) -> *mut PyObject;
+    #[cfg(all(Py_3_7, not(Py_3_9)))]
+    pub fn _PyImport_FindBuiltin(name: *const c_char, modules: *mut PyObject) -> *mut PyObject;
+    #[cfg(not(Py_3_10))]
+    pub fn _PyImport_FindExtensionObject(a: *mut PyObject, b: *mut PyObject) -> *mut PyObject;
+    #[cfg(not(Py_3_7))]
+    pub fn _PyImport_FixupBuiltin(module: *mut PyObject, name: *const c_char) -> c_int;
+    #[cfg(Py_3_7)]
+    pub fn _PyImport_FixupBuiltin(
+        module: *mut PyObject,
+        name: *const c_char,
+        modules: *mut PyObject,
+    ) -> c_int;
+    #[cfg(not(Py_3_7))]
+    pub fn _PyImport_FixupExtensionObject(
+        a: *mut PyObject,
+        b: *mut PyObject,
+        c: *mut PyObject,
+    ) -> c_int;
+    #[cfg(Py_3_7)]
+    pub fn _PyImport_FixupExtensionObject(
+        a: *mut PyObject,
+        b: *mut PyObject,
+        c: *mut PyObject,
+        d: *mut PyObject,
+    ) -> c_int;
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct _inittab {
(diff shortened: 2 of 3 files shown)


src/ffi/import.rs
```diff
@@ -82,3 +84,68 @@ extern "C" {
...
+#[derive(Copy, Clone)]
+#[cfg(not(Py_LIMITED_API))]
+pub struct _frozen {
+    pub name: *const c_char,
+    pub code: *const c_uchar,
+    pub size: c_int,
+}
+
+#[cfg(not(Py_LIMITED_API))]
+extern "C" {
```
davidhewitt, line 106: This extern block contains `pub static mut` so it's going to need `#[cfg_attr(windows, link(name = "pythonXY"))]`

davidhewitt (changes requested):
Thanks for adding these. Can they please be located in `src/ffi/cpython/import.rs`?

We try to match the cpython header structure as much as possible to make it easier to diff against upstream changes.

If you're willing to put the definitions as the same order as they currently are [in the cpython header](https://github.com/python/cpython/blob/master/Include/import.h), that'd also be much appreciated.

davidhewitt:
Also can confirm that none of these APIs appear to exist for PyPy.

kngwyu (commented):
Thanks. As David says, please consider placing some APIs in `cpython` mod. We're refactoring the FFI module in that way (https://github.com/PyO3/pyo3/issues/1289).

src/ffi/cpython/import.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub name: *const c_char,
+    pub initfun: Option<unsafe extern "C" fn()>,
+}
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    #[allow(unused)]
+    pub static mut PyImport_Inittab: *mut _inittab;
+    #[allow(unused)]
+    pub fn PyImport_ExtendInittab(newtab: *mut _inittab) -> c_int;
```
indygreg, line 55: For some reason I'm getting a compiler warning about `PyImport_ExtendInittab()` being unused, so I suppressed this. Unsure why. Maybe the presence of the `pub static mut` above triggers it?

indygreg: Yeah, that was it. I moved this to its own block 🤷‍♂️

src/ffi/cpython/import.rs
```diff
@@ -0,0 +1,72 @@
...
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct _inittab {
+    pub name: *const c_char,
+    pub initfun: Option<unsafe extern "C" fn()>,
+}
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    #[allow(unused)]
```
davidhewitt, line 52: I don't think that this is needed?

```suggestion
```

src/ffi/cpython/import.rs
```diff
@@ -0,0 +1,72 @@
...
+#[derive(Copy, Clone)]
+pub struct _frozen {
+    name: *const c_char,
+    code: *const c_uchar,
+    size: c_int,
+}
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    #[allow(unused)]
```
davidhewitt, line 70: Likewise here.

```suggestion
```

src/ffi/cpython/import.rs
```diff
@@ -0,0 +1,72 @@
...
+        c: *mut PyObject,
+        d: *mut PyObject,
+    ) -> c_int;
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct _inittab {
+    pub name: *const c_char,
+    pub initfun: Option<unsafe extern "C" fn()>,
```
davidhewitt, line 47: @kngwyu's original comment still stands here - looks like this is missing the return type:

```suggestion
    pub initfun: Option<unsafe extern "C" fn() -> *mut PyObject>,
```

davidhewitt (commented):
Thanks! Just a final couple of tweaks needed before merging.

Strange, I can't reproduce the unused code warning if I remove those two attributes from the `static mut` variables...

src/ffi/cpython/import.rs
```diff
@@ -0,0 +1,72 @@
+use crate::ffi::{PyInterpreterState, PyObject};
+use std::os::raw::{c_char, c_int, c_uchar};
+
```
nw0, line 3:
```suggestion

// skipped PyInit__imp

```

src/ffi/cpython/import.rs
```diff
@@ -0,0 +1,72 @@
...
+    ) -> c_int;
+    #[cfg(not(Py_3_7))]
+    pub fn _PyImport_FixupExtensionObject(
+        a: *mut PyObject,
+        b: *mut PyObject,
+        c: *mut PyObject,
+        d: *mut PyObject,
+    ) -> c_int;
+    #[cfg(Py_3_7)]
+    pub fn _PyImport_FixupExtensionObject(
```
nw0, line 35: Am I missing something or is this def the same for Py_3_7 and not(Py_3_7) ?

nw0 (commented):
I think the current comments cover everything in import.h from CPython master.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add FFI definition `PyCFunction_CheckExact` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
 - Add FFI definition `Py_IS_TYPE`. [#1429](https://github.com/PyO3/pyo3/pull/1429)
 - Add FFI definition `_Py_InitializeMain`. [#1473](https://github.com/PyO3/pyo3/pull/1473)
+- Add FFI definitions from `cpython/import.h`.
```
nw0, line 19:
```suggestion
- Add FFI definitions from `cpython/import.h`. [#1475](https://github.com/PyO3/pyo3/pull/1475)
```

src/ffi/cpython/import.rs
```diff
@@ -0,0 +1,72 @@
...
+        name: *const c_char,
+        modules: *mut PyObject,
+    ) -> c_int;
+    #[cfg(not(Py_3_7))]
+    pub fn _PyImport_FixupExtensionObject(
+        a: *mut PyObject,
+        b: *mut PyObject,
+        c: *mut PyObject,
+        d: *mut PyObject,
+    ) -> c_int;
```
davidhewitt, line 33: Good catch @nw0 - looks like for Python 3.6 this had one fewer arguments:

https://github.com/python/cpython/blob/a64de63248267acec87a3ffb6de66ee7008977b2/Include/import.h#L107
```suggestion
    pub fn _PyImport_FixupExtensionObject(
        a: *mut PyObject,
        b: *mut PyObject,
        c: *mut PyObject,
    ) -> c_int;
```

davidhewitt:
@indygreg thanks for pushing this PR along to this point; I'm going to take it over and push a commit shortly to finish it off as part of some tidy-ups I'm doing for the 0.14 release.

davidhewitt (approved):

PR #1476, opened 2021-03-07
Add context manager example to user guide

isosphere:
A simple illustrative example on how to use context managers. I required this to use pymc3, which relies heavily on the context stack for modelling.

My example has a gap in that it calls `__exit__` with blank arguments. I don't honestly know how to use it without them, or if I even need to call it at all. If anyone wants to add some detail there, I'd appreciate it!

### DIFF
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -261,7 +261,8 @@ class House(object):
 
         let result = py.eval("undefined_variable + 1", None, None);
         
-        // If the eval threw an exception we'll pass it through to the context manager. Otherwise, __exit__  is called with empty arguments.
+        // If the eval threw an exception we'll pass it through to the context manager. 
+        // Otherwise, __exit__  is called with empty arguments (Python "None").
         match result {
             Ok(_) => {
                 let none = py.None();
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -253,7 +253,7 @@ class House(object):
         else:
             print(f"Thank you for visiting {self.address}, come again soon!")
 
-        "#, "objects.py", "objects").unwrap();
+        "#, "house.py", "house").unwrap();
         
         let house = custom_manager.call_method1("House", ("123 Main Street",)).unwrap();
 
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -255,7 +255,7 @@ class House(object):
 
         "#, "house.py", "house").unwrap();
         
-        let house = custom_manager.call_method1("House", ("123 Main Street",)).unwrap();
+        let house = custom_manager.call_function1("House", ("123 Main Street",)).unwrap();
 
         house.call_method0("__enter__").unwrap();
 
@@ -266,10 +266,10 @@ class House(object):
         match result {
             Ok(_) => {
                 let none = py.None();
-                house.call_method1("__exit__", (&none, &none, &none)).unwrap();
+                house.call_function1("__exit__", (&none, &none, &none)).unwrap();
             },
             Err(e) => {
-                house.call_method1(
+                house.call_function1(
                     "__exit__",
                     (e.ptype(py), e.pvalue(py), e.ptraceback(py))
                 ).unwrap();
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -255,11 +255,13 @@ class House(object):
 
         "#, "objects.py", "objects").unwrap();
         
-        let house = custom_manager.call1("House", ("123 Main Street",)).unwrap();
+        let house = custom_manager.call_method1("House", ("123 Main Street",)).unwrap();
 
         house.call_method0("__enter__").unwrap();
 
         let result = py.eval("undefined_variable + 1", None, None);
+        
+        // If the eval threw an exception we'll pass it through to the context manager. Otherwise, __exit__  is called with empty arguments.
         match result {
             Ok(_) => {
                 let none = py.None();
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -266,10 +266,10 @@ class House(object):
         match result {
             Ok(_) => {
                 let none = py.None();
-                house.call_function1("__exit__", (&none, &none, &none)).unwrap();
+                house.call_method1("__exit__", (&none, &none, &none)).unwrap();
             },
             Err(e) => {
-                house.call_function1(
+                house.call_method1(
                     "__exit__",
                     (e.ptype(py), e.pvalue(py), e.ptraceback(py))
                 ).unwrap();
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -257,7 +257,7 @@ class House(object):
         
         let house = custom_manager.call_function1("House", ("123 Main Street",)).unwrap();
 
-        house.call_method0("__enter__").unwrap();
+        house.call_function0("__enter__").unwrap();
 
         let result = py.eval("undefined_variable + 1", None, None);
         
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -237,11 +237,11 @@ Use context managers by directly invoking `__enter__` and `__exit__`.
 
 ```rust
 use pyo3::prelude::*;
-use pyo3::types::{IntoPyDict, PyModule};
+use pyo3::types::PyModule;
 
 fn main() {   
     Python::with_gil(|py| {
-        let CustomManager = PyModule::from_code(py, r#"
+        let custom_manager = PyModule::from_code(py, r#"
 class House(object):
     def __init__(self, address):
         self.address = address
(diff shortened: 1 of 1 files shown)


isosphere:
CI failure looks unrelated:

```

failures:

---- types::string::test::test_extract_char stdout ----
---- types::string::test::test_extract_char stderr ----
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'ValueError'>, value: ValueError('expected a string of length 1'), traceback: None }', src/types/string.rs:239:73


failures:
    types::string::test::test_extract_char

test result: FAILED. 236 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.32s
```

davidhewitt:
Thanks, will read later! Yes, the coverage failure is a regression in upstream Rust nightly that'll be fixed in good time ;)

guide/src/python_from_rust.md
```diff
@@ -230,3 +230,30 @@ def leaky_relu(x, slope=0.01):
...
+        self.address = address
+    def __enter__(self):
+        print(f"Welcome to {self.address}!")
+    def __exit__(self, type, value, traceback):
+        print(f"Thank you for visiting {self.address}, come again soon!")
+        "#, "objects.py", "objects").unwrap();
+        
+        let house = custom_manager.call1("House", ("123 Main Street",)).unwrap();
+        house.call_method0("__enter__").unwrap();
+        house.call_method1("__exit__", ("", "", "")).unwrap();
```
davidhewitt, line 256: So in Python the arguments to `__exit__` are the exception information in the case of an error occurring inside the `with` block.

When working from Rust we'll need to set this ourselves - we should either use `py.None()` if no error occurred, or the exception tuple in the case where there was an error.

... we in theory could get even fancier than this - according to the [docs for `object.__exit__()`](https://docs.python.org/3/reference/datamodel.html#object.__exit__) if `__exit__` returns a truthy value then we should suppress the original exception, otherwise we should raise it.

E.g.

```suggestion
        let result = do_some_python_operation_that_might_fail();
        match result {
            Ok(_) => {
                let none = py.None();
                house.call_method1("__exit__", (none, none, none)).unwrap();
            },
            Err(e) => house.call_method1(
                "__exit__",
                (e.ptype(py), e.pvalue(py), e.ptraceback(py))
            ).unwrap();
        }
```

the "re-raising" in the case of a falsy value would be better demonstrated in a `#[pyfunction]` where the `Err` can be returned back to Python.

davidhewitt (commented):
Thanks very much for writing this. You're right that the `__exit__` part probably needs some expansion - I've added a comment below.

isosphere:
Thanks for your suggestion! I ran with it, and did a little eval with an undefined variable which will always through an exception. Here's the output I get:

```
Welcome to 123 Main Street!
Sorry you had <class 'NameError'> trouble at 123 Main Street
```

Your use of py.None() there results in a moved value, so I passed it by reference (even though that branch of the match will never execute, but it should be helpful as an example of what to do). However, I'm not certain that it's okay to pass py.None() by reference here.

guide/src/python_from_rust.md
```diff
@@ -230,3 +230,48 @@ def leaky_relu(x, slope=0.01):
...
+        print(f"Welcome to {self.address}!")
+    def __exit__(self, type, value, traceback):
+        if type:
+            print(f"Sorry you had {type} trouble at {self.address}")
+        else:
+            print(f"Thank you for visiting {self.address}, come again soon!")
+
+        "#, "objects.py", "objects").unwrap();
+        
+        let house = custom_manager.call1("House", ("123 Main Street",)).unwrap();
```
davidhewitt, line 258: Should be `call_function1`, because we actually just deprecated `call1` on master.

@birkenfeld strictly speaking this isn't a function, so maybe `call_attribute` is better? Although I wonder if actually we should plan to remove `module::call{_function}` etc. completely and instead just ask users to do `.getattr("...")?.call1("...")?` etc? It's not much longer and doesn't suffer with naming ambiguity.

guide/src/python_from_rust.md
```diff
@@ -230,3 +230,48 @@ def leaky_relu(x, slope=0.01):
...
+
+        house.call_method0("__enter__").unwrap();
+
+        let result = py.eval("undefined_variable + 1", None, None);
+        match result {
```
davidhewitt, line 263: It might be nice to explain a comment above here to explain why there's two branches calling `__exit__` differently.

davidhewitt (approved):
Thanks, just two final nits!

guide/src/python_from_rust.md
isosphere: I think you must mean `call_method1`, so I've changed it to that!

davidhewitt: See https://github.com/PyO3/pyo3/commit/e065f9b5170f76a529aaf5bad25256ed5b7d75f7 - `call_function1` is very hot off the press :)

guide/src/python_from_rust.md
```diff
@@ -261,7 +261,8 @@ class House(object):
 
         let result = py.eval("undefined_variable + 1", None, None);
         
-        // If the eval threw an exception we'll pass it through to the context manager. Otherwise, __exit__  is called with empty arguments.
+        // If the eval threw an exception we'll pass it through to the context manager. 
+        // Otherwise, __exit__  is called with empty arguments (Python "None").
```
davidhewitt, line 265: 👍

guide/src/python_from_rust.md
```diff
@@ -230,3 +230,51 @@ def leaky_relu(x, slope=0.01):
...
+        self.address = address
+    def __enter__(self):
+        print(f"Welcome to {self.address}!")
+    def __exit__(self, type, value, traceback):
+        if type:
+            print(f"Sorry you had {type} trouble at {self.address}")
+        else:
+            print(f"Thank you for visiting {self.address}, come again soon!")
+
+        "#, "objects.py", "objects").unwrap();
```
kngwyu, line 256: Just a nit, but is `house.py` better?

kngwyu (approved):
Simple and effective example 👍🏼 , thank you!

guide/src/python_from_rust.md
isosphere: > 
> 
> See [e065f9b](https://github.com/PyO3/pyo3/commit/e065f9b5170f76a529aaf5bad25256ed5b7d75f7) - `call_function1` is very hot off the press :)

My understanding is that the `call_function1` change will require a larger-scale rewrite of the documentation, which is out of scope for this specific pull request.  For example: call is still used earlier in this same document.

I would propose leaving this as-is, and doing the broader change to `call_function1` in a separate pull request that can cover a wide scope and capture all the appropriate uses of `call` and otherwise that should be changed.

guide/src/python_from_rust.md
isosphere: I agree, changed!

guide/src/python_from_rust.md
isosphere: Oops, you updated the documentation in that commit. Ignore earlier message if you saw it! I'll make the change!

guide/src/python_from_rust.md
```diff
@@ -230,3 +230,51 @@ def leaky_relu(x, slope=0.01):
...
+        if type:
+            print(f"Sorry you had {type} trouble at {self.address}")
+        else:
+            print(f"Thank you for visiting {self.address}, come again soon!")
+
+        "#, "house.py", "house").unwrap();
+        
+        let house = custom_manager.call_function1("House", ("123 Main Street",)).unwrap();
+
+        house.call_function0("__enter__").unwrap();
```
davidhewitt, line 260: Confusingly, this one _should_ be call `call_method0` - the new name `call_function` is just for `&PyModule`. But now the confusion makes me think this was a mistake.

Sorry 😬 

```suggestion
        house.call_method0("__enter__").unwrap();
```

guide/src/python_from_rust.md
```diff
@@ -230,3 +230,51 @@ def leaky_relu(x, slope=0.01):
...
+        
+        // If the eval threw an exception we'll pass it through to the context manager. 
+        // Otherwise, __exit__  is called with empty arguments (Python "None").
+        match result {
+            Ok(_) => {
+                let none = py.None();
+                house.call_function1("__exit__", (&none, &none, &none)).unwrap();
+            },
+            Err(e) => {
+                house.call_function1(
```
davidhewitt, line 272: Same for these.

```suggestion
                house.call_method1("__exit__", (&none, &none, &none)).unwrap();
            },
            Err(e) => {
                house.call_method1(
```

birkenfeld:
As for `call_function` vs `call_method` vs `call`: you're right, while `call_function` is better than `call`, it's not really such a good name and `call_attr` would be better.  Or even no special API.

In the end I wish we could implement some `Fn` trait for PyAny which would make this much easier :)

davidhewitt:
> In the end I wish we could implement some Fn trait for PyAny which would make this much easier :)

I have been abstractly dreaming recently about a macro to solve all this, spread args and kwargs and also use the Vectorcall optimization. E.g.:

```rust
py_call!(module.attr(a, b, *args, **kwargs));
```

but that's probably a pipe dream 😄

davidhewitt:
👍 Thanks very much for this example; sorry for the complications caused by our API changes.

Issue #1465, opened 2021-03-03
__bytes__() method does not get a 'text_signature' by default 


ravenexp:
## 🐛 Bug Reports

I'm using pydoc to check the doc strings and function signatures in the compiled extension module.
When implementing `PyObjectProtocol` trait for my class, all trait methods except for `__bytes__()` seem
to get their #[text_signature]'s set implicitly (no Rust code attributes required).
I expect the same to happen for `__bytes__()` method as well.

Example pydoc output:

```
    class Instruction(builtins.object)
     |  Python class wrapper for Rust `Instruction` enum.
     |  
     |  Methods defined here:
     |  
     |  __bytes__(...)
     |  
     |  __repr__(self, /)
     |      Return repr(self).
     |  
     |  __str__(self, /)
     |      Return str(self).
     |  
     |  size(self, /)
     |      Returns the encoded instruction length in bytes.
     |  
     |  ----------------------------------------------------------------------
```

### 🌍 Environment

 - Your operating system and version: Arch Linux
 - Your python version: 3.9.2
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: pacman
 - Your Rust version (`rustc --version`): 1.50
 - Your PyO3 version: 13.2
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: no

kngwyu:
Thanks, I'll investigate this bug.

davidhewitt:
I wonder if this is because `__bytes__` is a method rather than a slot? Afaik we don't generate signature for methods automatically.

ravenexp:
I've tried adding

```
/// Return bytes(self).                     
#[text_signature = "($self, /)"]
```

to the trait method implementation, but it had no effect.

ravenexp:
I confirm that https://github.com/PyO3/pyo3/pull/1483 fixed the problem and both `#[text_signature]` and doc comments now work with `__bytes__()`.

I think we can close this issue now.

PR #1483, opened 2021-03-09
pymethods: macros tidy-ups

davidhewitt:
This is just a little bit of refactoring that I spotted could be done in the proc-macro code.

It allows re-use of a little bit more code between `#[pymethods]` and `#[pyproto]`, with the effect that "normal methods" in `#[pyproto]` such as `__bytes__` and `__enter__` can now be generated as `METH_NOARGS`, which I think should be a little bit better for performance.

### DIFF
--- a/pyo3-macros-backend/src/pyproto.rs
+++ b/pyo3-macros-backend/src/pyproto.rs
@@ -64,33 +64,22 @@ fn impl_proto_impl(
             if let Some(m) = proto.get_method(&met.sig.ident) {
                 let fn_spec = FnSpec::parse(&mut met.sig, &mut met.attrs, false)?;
 
+                let flags = if m.can_coexist {
+                    // We need METH_COEXIST here to prevent __add__  from overriding __radd__
+                    Some(quote!(pyo3::ffi::METH_COEXIST))
+                } else {
+                    None
+                };
+
                 let method = if let FnType::Fn(self_ty) = &fn_spec.tp {
-                    pymethod::impl_proto_wrap(ty, &fn_spec, &self_ty)?
+                    pymethod::impl_py_method_def(ty, &fn_spec, &self_ty, flags)?
                 } else {
                     bail_spanned!(
                         met.sig.span() => "expected method with receiver for #[pyproto] method"
                     );
                 };
 
-                let coexist = if m.can_coexist {
-                    // We need METH_COEXIST here to prevent __add__  from overriding __radd__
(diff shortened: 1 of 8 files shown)


pyo3-macros-backend/src/pyproto.rs
```diff
@@ -64,33 +64,22 @@ fn impl_proto_impl(
...
+                let flags = if m.can_coexist {
+                    // We need METH_COEXIST here to prevent __add__  from overriding __radd__
+                    Some(quote!(pyo3::ffi::METH_COEXIST))
+                } else {
+                    None
+                };
+
                 let method = if let FnType::Fn(self_ty) = &fn_spec.tp {
-                    pymethod::impl_proto_wrap(ty, &fn_spec, &self_ty)?
+                    pymethod::impl_py_method_def(ty, &fn_spec, &self_ty, flags)?
```
davidhewitt, line 75: @kngwyu making this change appears to have partially fixed #1465

The `__bytes__` text signature is still not generated automatically, but the `#[text_signature]` attribute does seem to work now.

kngwyu: Great, thanks!

src/class/methods.rs
```diff
@@ -95,17 +95,21 @@ impl PyMethodDef {
     pub const fn cfunction_with_keywords(
         name: &'static str,
         cfunction: PyCFunctionWithKeywords,
-        flags: c_int,
```
kngwyu, line 98: How about `Option<c_int>`?
Passing `None` would be (slightly) better than `0`.
I'm writing this because I don't like using `.flags()` pattern for a non-builder struct, but this is my personal taste, and please choose what you like.

kngwyu (approved):
Nice refactoring, thank you!

src/class/methods.rs
davidhewitt: Thanks, I changed to do so.

davidhewitt:
Hmmm the clippy failure is annoying - `.unwrap_or` is not a `const fn` so the suggestion is incorrect.

src/class/methods.rs
davidhewitt: Ah, also with MSRV we cannot use `match` in `const fn`, so we'd need to store `Option<c_int>` in the struct and do slightly more work at runtime.

So I'm going to keep `.flags()` for now. Maybe in the future after we increase MSRV we can change to use `Option<c_int>`.

Issue #1480, opened 2021-03-08
Passing generic types to functions


jafioti:
I have a function written in Rust accepting a Vec<T>, and I am trying to build a wrapper around it in PyO3. Currently this is what my function header looks like:

`pub fn pad_mask<T: std::cmp::PartialEq>(batch: &Vec<Vec<T>>, pad_value: T) -> Vec<Vec<bool>> `

I have been having trouble implementing a wrapper for this. I get an error on T: cannot find type `T` in this scope
Is this possible to wrap, or will I need to use some hacks to get around it?

nw0:
By "implementing a wrapper", do you mean with the `#[pyfunction]` macro, or something else?

I could be entirely wrong, but looking at the macro code, I think generic parameters aren't handled currently.

jafioti:
Yes I mean #[pyfunction]. Is this something that is possible at all, is there any workarounds, or will this be a feature coming soon?

davidhewitt:
Unfortunately this isn't possible at a technical level - a generic Rust function has a separate function created for each different type you use it with. And this set of types is known at compile time.

However we can only wrap _one_ function to Python.

Instead of a generic type parameter `T` you should probably use `&PyAny`.

jafioti:
Ok, I've changed my function to:
`#[pyfunction]
pub fn pad_mask(batch: &Vec<Vec<&PyAny>>, pad_value: &PyAny) -> Vec<Vec<bool>> `
But I am now getting the error the trait bound &Vec<Vec<&pyo3::PyAny>>: PyClass is not satisfied

davidhewitt:
Does `Vec<Vec<_>>` work without the reference?

FWIW I would recommend using `PyArray` from the numpy crate instead of `Vec<Vec<_>>` as it'll be much much faster than converting all the data into Rust. 

However you're going to need https://github.com/PyO3/rust-numpy/issues/175 to have a `PyObject` array. I could PR that for you sometime so it's on numpy `master` - or if you want it faster you could try opening a PR yourself?

PR #1486, opened 2021-03-09
pyfunction: reject generic functions

davidhewitt:
This improves the error message noted in #1480 when using generics with `#[pyfunction]`.

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -6,7 +6,6 @@ use crate::utils;
 use proc_macro2::TokenStream;
 use quote::ToTokens;
 use quote::{quote, quote_spanned};
-use std::ops::Deref;
 use syn::ext::IdentExt;
 use syn::spanned::Spanned;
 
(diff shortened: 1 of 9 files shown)


pyo3-macros-backend/src/method.rs
```diff
@@ -21,6 +20,44 @@ pub struct FnArg<'a> {
     pub attrs: PyFunctionArgAttrs,
 }
 
+impl<'a> FnArg<'a> {
+    /// Transforms a rust fn arg parsed with syn into a method::FnArg
+    pub fn parse(arg: &'a mut syn::FnArg) -> syn::Result<Self> {
```
kngwyu, line 25: A bit nitpicking, but syn has a [Parse::parse](https://docs.rs/syn/latest/syn/parse/trait.Parse.html) which takes `TokenStream` and returns `Option<Self>`.
So, isn't this naming confusing?
I think the boring `from_*` style naming is better (e.g.,  `from_syn_fnarg`).

kngwyu (approved):
Looks great, thanks!

PR #1487, opened 2021-03-10
Use _PYTHON_SYSCONFIGDATA_NAME in find_sysconfigdata when cross-compiling

### LINKED ISSUE #1484: Failure if there are multiple sysconfigdata files during cross-compilation
PyO3 searches for `_sysconfigdata*.py` and crashes if it finds any other number than 1 of them:

https://github.com/PyO3/pyo3/blob/4c80ce93c39110f1d65944c2ce4ea2149e241613/build.rs#L411

If you are cross compiling, you can set the target lib dir with `PYO3_CROSS_LIB_DIR` and specify the python version with `PYO3_CROSS_PYTHON_VERSION` if there are multiple versions.

Within these contraints, any matching `_sysconfigdata*.py` file will be found:

https://github.com/PyO3/pyo3/blob/4c80ce93c39110f1d65944c2ce4ea2149e241613/build.rs#L431

However, there is a problem with Anaconda Python versions, which is that all alternative `_sysconfigdata*.py` files are always bundled with Python, not only the target platform variant. So we have among others, the following files, where only one of them is relevant:

- `/lib/python3.8/_sysconfigdata_arm64_apple_darwin20_0_0.py`
- `/lib/python3.8/_sysconfigdata_powerpc64le_conda_linux_gnu.py`
- `/lib/python3.8/_sysconfigdata_x86_64_conda_cos6_linux_gnu.py`

If I understand correctly, CPython uses a `_PYTHON_SYSCONFIGDATA_NAME` environment variable as a workaround for this specific cross-compilation issue.

My proposed solution is to take the sysconfigdata that corresponds with `_PYTHON_SYSCONFIGDATA_NAME` if there are multiple sysconfigdata files, the environment variable is set, and if that particular file exists.

### 🌍 Environment

 - Your operating system and version: macOS 11.2
 - Your python version: conda-forge/python 3.8.8
 - Your PyO3 version: 0.12.4+

This problem affects everyone who tries to do PyO3 cross compilation with Anaconda Python.

messense:
Fixes #1484 

CPython reference code: 
1. https://hg.python.org/cpython/rev/04b679dd11eb
2. https://github.com/python/cpython/blob/b4f9089d4aa787c5b74134c98e5f0f11d9e63095/Lib/sysconfig.py#L366-L372

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -397,9 +397,14 @@ fn ends_with(entry: &DirEntry, pat: &str) -> bool {
 /// [1]: https://github.com/python/cpython/blob/3.5/Lib/sysconfig.py#L389
 fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
     let sysconfig_paths = search_lib_dir(&cross.lib_dir, &cross);
+    let sysconfig_name = env::var_os("_PYTHON_SYSCONFIGDATA_NAME");
     let mut sysconfig_paths = sysconfig_paths
         .iter()
-        .filter_map(|p| fs::canonicalize(p).ok())
+        .filter_map(|p| {
+            fs::canonicalize(p)
+                .ok()
+                .filter(|p| p.file_stem() == sysconfig_name.as_deref())
+        })
         .collect::<Vec<PathBuf>>();
     sysconfig_paths.dedup();
     if sysconfig_paths.is_empty() {
@@ -407,18 +412,13 @@ fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
             "Could not find either libpython.so or _sysconfigdata*.py in {}",
             cross.lib_dir.display()
         );
-    } else {
-        if let Some(sysconfig_name) = env::var_os("_PYTHON_SYSCONFIGDATA_NAME") {
-            sysconfig_paths.retain(|p| *p == Path::new(&sysconfig_name).with_extension("py"));
-        }
-        if sysconfig_paths.len() > 1 {
-            bail!(
-                "Detected multiple possible python versions, please set the PYO3_PYTHON_VERSION \
-                variable to the wanted version on your system or set the _PYTHON_SYSCONFIGDATA_NAME \
-                variable to the wanted sysconfigdata file name\nsysconfigdata paths = {:?}",
-                sysconfig_paths
-            )
-        }
+    } else if sysconfig_paths.len() > 1 {
+        bail!(
+            "Detected multiple possible python versions, please set the PYO3_PYTHON_VERSION \
+            variable to the wanted version on your system or set the _PYTHON_SYSCONFIGDATA_NAME \
+            variable to the wanted sysconfigdata file name\nsysconfigdata paths = {:?}",
+            sysconfig_paths
+        )
     }
 
     Ok(sysconfig_paths.remove(0))
--- a/build.rs
+++ b/build.rs
@@ -407,12 +407,18 @@ fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
             "Could not find either libpython.so or _sysconfigdata*.py in {}",
             cross.lib_dir.display()
         );
-    } else if sysconfig_paths.len() > 1 {
-        bail!(
-            "Detected multiple possible python versions, please set the PYO3_PYTHON_VERSION \
-            variable to the wanted version on your system\nsysconfigdata paths = {:?}",
-            sysconfig_paths
-        )
+    } else {
+        if let Some(sysconfig_name) = env::var_os("_PYTHON_SYSCONFIGDATA_NAME") {
+            sysconfig_paths.retain(|p| *p == Path::new(&sysconfig_name).with_extension("py"));
+        }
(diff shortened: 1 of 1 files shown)


kngwyu (commented):
Thanks!
How about filtering paths which are not equal to `sysconfig_name` before constructing a Vec?
```rust
    let sysconfig_paths = search_lib_dir(&cross.lib_dir, &cross);
    let sysconfig_name = env::var_os("_PYTHON_SYSCONFIGDATA_NAME");
    let mut sysconfig_paths = sysconfig_paths
        .iter()
        .filter_map(|p| {
            fs::canonicalize(p)
                .ok()
                .filter(|p| p.file_name() == sysconfig_name.as_deref())
        })
        .collect::<Vec<PathBuf>>();
```

build.rs
```diff
@@ -397,9 +397,14 @@ fn ends_with(entry: &DirEntry, pat: &str) -> bool {
...
 fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
     let sysconfig_paths = search_lib_dir(&cross.lib_dir, &cross);
+    let sysconfig_name = env::var_os("_PYTHON_SYSCONFIGDATA_NAME");
     let mut sysconfig_paths = sysconfig_paths
         .iter()
-        .filter_map(|p| fs::canonicalize(p).ok())
+        .filter_map(|p| {
+            fs::canonicalize(p)
+                .ok()
+                .filter(|p| p.file_stem() == sysconfig_name.as_deref())
```
messense, line 406: I've changed `file_name()` to `file_stem()` since `_PYTHON_SYSCONFIGDATA_NAME` does not include the `.py` extension.

messense:
```
error[E0658]: use of unstable library feature 'matches_macro'
   --> pyo3-macros-backend/src/method.rs:153:18
    |
153 |                 !matches!(&**ty, syn::Type::ImplTrait(_)),
    |                  ^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/65721
```

Hmm, interesting. Seems introduced in https://github.com/PyO3/pyo3/pull/1486

davidhewitt:
Yes that's my fault, sorry. I'll push a fix in a moment.

build.rs
davidhewitt, line 406: What happens if `sysconfig_name` is unset? It looks to me like we need to do something more like:

```suggestion
            let canonical = fs::canonicalize(p).ok();
            match &sysconfig_name {
                Some(name) => canonical.filter(|p| p.file_stem() == name),
                None => canonical,
            }
```

kngwyu (approved):
LGTM, thank you! 💯

mattip:
Anything starting with an `_` is considered an implementation detail. Deriving values from such interfaces is going to be fragile and not supported on all python implementations, like PyPy. If you wish python to formally support this, you should start a discussion with the python developers on the python ideas discourse channel https://discuss.python.org/c/ideas

davidhewitt:
@mattip are you referring to the usage of this env var, or generally the whole practice of parsing sysconfigdata?

mattip:
Both. 

There is no standard to `_sysconfigdata` (with the underline), no consensus of how to ship multiple files, and no documentation of what values are expected to be in it. 

There is also no standard environment variable `_PYTHON_SYSCONFIGDATA_NAME`, the [discussion in the issue](https://bugs.python.org/issue28046#msg275433) with the commit pointed to in the cpython code apparently happened at a sprint so there is no documentation of the thought that went into to the change and what is expected to be in a shipped python installation. So it is very difficult for PyPy to be compliant with this idea. 

See issues PyO3/maturin#559 and the corresponding PyPy issue https://foss.heptapod.net/pypy/pypy/-/issues/3483. I am trying to reconstruct how this is supposed to be used in cross-compiling for PyPy (which does not use `configure`, and how, for instance, the use of these internal interfaces works on Ubuntu or a standard manylinux docker image where I fail to see any additional `_sysconfigdata*` files other than the ones for the host interpreter. Is there a design document for cross-compilation?

messense:
I think it's fine to use `_PYTHON_SYSCONFIGDATA_NAME` here because it's only used for telling pyo3 to use that sysconfigdata file, not passed to Python interpreter. We can call it with different name but since CPython already use this one, it's easier for users to relate and understand.

> Is there a design document for cross-compilation?

I wrote a little about that here: https://foss.heptapod.net/pypy/pypy/-/issues/3483#note_160232

davidhewitt:
FWIW @mattip I agree with your view that parsing `_sysconfigdata` is potentially brittle. With lack of an official technique, it's a system that we've come up with empirically to resolve needs of our users ad-hoc. For now I'm prepared to live with the consequences that it may break with any future Python release, and we'll find workarounds as needed then.

I've created #1672 to record an intention to work with upstream to find a permanent stable solution, though I'm unsure when I would be able to find time to champion doing so myself.

Given we've gone off-piste, I think it's perfectly reasonable that PyPy should be free to ignore our desire for support for a private interface. I do see that on the PyPy issue tracker (3484) you've been doing a lot of work to support our needs, so I'm very grateful for that ❤️

PR #1489, opened 2021-03-12
msrv: remove use of matches! macro

davidhewitt:
Fixes the MSRV issue introduced in #1486.

I've also changed the actions jobs settings so that MSRV failure prevents merging.


messense (approved):
LGTM

PR #1492, opened 2021-03-13
pymodule: remove call_function etc.

davidhewitt:
So on `master` we renamed `PyModule::call` to `PyModule::call_function` etc. in #1467 

But this is causing confusion to a user unfamiilar with the project in #1476 .

So I suggest we just remove `PyModule::call_function` and instead point users to use `.getattr(...)?.call()`. There's no optimization in `PyModule::call_function` etc., so I think there's no reason we need to offer these APIs.

I also deprecated `PyModule::get` because it's just a dumb wrapper around `PyAny::getattr`, by the same logic.

### DIFF
--- a/src/types/module.rs
+++ b/src/types/module.rs
@@ -129,65 +129,6 @@ impl PyModule {
         }
     }
 
-    /// Calls a function in the module.
-    ///
-    /// This is equivalent to the Python expression `module.name(*args, **kwargs)`.
-    pub fn call_function(
-        &self,
-        name: &str,
-        args: impl IntoPy<Py<PyTuple>>,
-        kwargs: Option<&PyDict>,
-    ) -> PyResult<&PyAny> {
-        self.getattr(name)?.call(args, kwargs)
-    }
-
-    /// Calls a function in the module with only positional arguments.
-    ///
-    /// This is equivalent to the Python expression `module.name(*args)`.
-    pub fn call_function1(&self, name: &str, args: impl IntoPy<Py<PyTuple>>) -> PyResult<&PyAny> {
-        self.getattr(name)?.call1(args)
-    }
-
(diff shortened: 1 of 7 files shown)


davidhewitt:
cc @birkenfeld

src/types/module.rs
```diff
@@ -310,4 +251,44 @@ impl PyModule {
...
+    /// This is equivalent to the Python expression `module.name()`.
+    #[deprecated(since = "0.14.0", note = "use getattr(name)?.call0() instead")]
+    pub fn call0(&self, name: &str) -> PyResult<&PyAny> {
+        self.getattr(name)?.call0()
+    }
+
+    /// Gets a member from the module.
+    ///
+    /// This is equivalent to the Python expression `module.name`.
+    #[deprecated(since = "0.14.0", note = "use getattr(name)?.call0() instead")]
```
birkenfeld, line 290:
```suggestion
    #[deprecated(since = "0.14.0", note = "use getattr(name) instead")]
```

tests/test_datetime.rs
```diff
@@ -108,7 +108,7 @@ fn test_datetime_utc() {
     let gil = Python::acquire_gil();
     let py = gil.python();
     let datetime = py.import("datetime").map_err(|e| e.print(py)).unwrap();
-    let timezone = datetime.get("timezone").unwrap();
+    let timezone = datetime.getattr("tattrimezone").unwrap();
```
birkenfeld, line 111:
```suggestion
    let timezone = datetime.getattr("timezone").unwrap();
```

birkenfeld (approved):

birkenfeld:
Whoops, should have been now "approve" yet.

kngwyu (approved):
Thanks!

messense (approved):

Issue #1463, opened 2021-03-03
Create macro for creating kwargs dict?


birkenfeld:
While calling methods with only positional arguments of different types is pretty easy (with tuples of `IntoPy`), for kwargs it gets complicated (see #1462).

Would it be useful to have a macro that can be used like so:
```
obj.call((pos_arg1, pos_arg2), kwargs! { key1=value1, key2=value2 })
```

davidhewitt:
Absolutely! I've thought about this before and I think we can do one better and support python collections a-la `vec!` and `maplit::hashmap!` etc. I think I'd make them look like this:

```
py_list!(py, [a, b, c]);
py_tuple!(py, (a, b, c));
py_dict!(py, {a => 1, b => 2, c => 3});
py_set!(py, {a, b, c});
py_frozenset!(py, {a, b, c});
```

I think these have more general use than a `kwargs!` macro.

Also when we add functionality to use `vectorcall` to call Python functions, then I think we might need some kind of `py_call!` macro which does the magic.

daniil-konovalenko:
I wonder if we can take the `Python` instance from the scope implicitly:
```rust
list![a, b, c];
tuple!(a, b, c);
dict!{a => 1, b => 2, c => 3};
set!{a, b, c};
frozenset!{a, b, c};
```
so then
```rust
Python::with_gil(|py| {
    let my_list = list![1, 2, 3]
});
```

would expand to something like
```rust
Python::with_gil(|py| {
    let my_list = PyList::empty().unwrap();
    my_list.append(py, 1).unwrap();
    my_list.append(py, 2).unwrap();
    my_list.append(py, 3).unwrap();
});
```

I think it would look more familiar to Python users. I'm not sure though that being implicit here is a good idea. Also it relies heavily on the convention to name the `Python` instance `py`, which is brittle.

birkenfeld:
And it doesn't even work (with macro_rules macros) since you can't use outer identifiers in the expansion without passing them in. That shows that Rust doesn't want us to do this in the first place :)

daniil-konovalenko:
Ah, right, thanks. Never mind then :)

davidhewitt:
Reminder for myself: I was thinking it might be nice if we can also use spread syntax in these macros.

e.g.

```rust
py_list!(py, [*left, *right]);
```

PR #1494, opened 2021-03-14
Extend py_run! to take locals dict and refactor tests using it

kngwyu:
Currently, our tests have many lines like:
```rust
    let run = |code| py.run(code, None, Some(d)).unwrap();
    let err = |code| py.run(code, None, Some(d)).unwrap_err();
    run("...");
```
I felt this redundant.

So I prepared this PR.
`py_run!` now can take `locals` dict by `py_run!(py, *locals, "some_python_code()")` syntax.
This odd `*` is to distinguish dict with normal `py_run` usage with variables (e.g., `py_run!(py, obj1 obj2, "code")` and inspired by `*` syntax in Python. I believe this is not *very* confusing, but please propose alternatives if you have.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -331,10 +331,11 @@ macro_rules! wrap_pymodule {
 /// });
 /// ```
 ///
-/// If you need to prepare the `locals` dict by yourself, you can pass it by `*locals`.
+/// If you need to prepare the `locals` dict by yourself, you can pass it as `*locals`.
 ///
 /// ```
-/// # use pyo3::prelude::*;
+/// use pyo3::prelude::*;
+/// use pyo3::types::IntoPyDict;
 /// #[pyclass]
 /// struct MyClass {}
 /// #[pymethods]
@@ -343,7 +344,7 @@ macro_rules! wrap_pymodule {
 ///     fn new() -> Self { MyClass {} }
 /// }
 /// Python::with_gil(|py| {
-///    let locals = [("C", py.get_type::<MyClass>())];
+///    let locals = [("C", py.get_type::<MyClass>())].into_py_dict(py);
 ///    pyo3::py_run!(py, *locals, "c = C()");
 /// });
 /// ```
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -289,10 +289,10 @@ macro_rules! wrap_pymodule {
 /// # Example
 /// ```
 /// use pyo3::{prelude::*, py_run, types::PyList};
-/// let gil = Python::acquire_gil();
-/// let py = gil.python();
-/// let list = PyList::new(py, &[1, 2, 3]);
-/// py_run!(py, list, "assert list == [1, 2, 3]");
+/// Python::with_gil(|py| {
+///     let list = PyList::new(py, &[1, 2, 3]);
+///     py_run!(py, list, "assert list == [1, 2, 3]");
+/// });
 /// ```
 ///
 /// You can use this macro to test pyfunctions or pyclasses quickly.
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -320,15 +320,32 @@ macro_rules! wrap_pymodule {
...
+///     let time_as_tuple = (8, 43, 16);
+///     py_run!(py, time time_as_tuple, r#"
+///         assert time.hour == 8
+///         assert time.repl_japanese() == "8時43分16秒"
+///         assert time.as_tuple() == time_as_tuple
+///     "#);
+/// });
+/// ```
+///
+/// If you need to prepare the `locals` dict by yourself, you can pass it by `*locals`.
```
davidhewitt, line 334:
```suggestion
/// If you need to prepare the `locals` dict by yourself, you can pass it as `*locals`.
```

tests/test_buffer_protocol.rs
```diff
@@ -93,8 +95,7 @@ fn test_buffer() {
         )
         .unwrap();
         let env = [("ob", instance)].into_py_dict(py);
-        py.run("assert bytes(ob) == b' 23'", None, Some(env))
-            .unwrap();
+        py_assert!(py, *env, "bytes(ob) == b' 23'");
```
davidhewitt, line 98: 🚀 This is nice!

src/lib.rs
```diff
@@ -355,8 +378,10 @@ macro_rules! py_run_impl {
         use $crate::types::IntoPyDict;
         use $crate::ToPyObject;
         let d = [$((stringify!($val), $val.to_object($py)),)+].into_py_dict($py);
-
-        if let Err(e) = $py.run($code, None, Some(d)) {
+        $crate::py_run_impl!($py, *d, $code)
```
davidhewitt, line 381: What do you think of making a `#[doc(hidden)]` module `pyo3::impl_` and moving `py_run_impl` into there?

I think we could also move `pyo3::impl_::derive_utils` and `pyo3::impl_::internal_tricks` into such a module.

This way it'd be easier for users to see what bits of pyo3 are considered internals that they should not depend upon!

tests/test_mapping.rs
```diff
@@ -66,61 +69,58 @@ impl PyMappingProtocol for Mapping {
     }
 }
 
+/// Return a dict with `m = Mapping(['1', '2', '3'])`.
+fn map_dict(py: Python) -> &pyo3::types::PyDict {
+    let d = [("Mapping", py.get_type::<Mapping>())].into_py_dict(py);
+    py_run!(py, *d, "m = Mapping(['1', '2', '3'])");
+    d
+}
```
davidhewitt, line 77: If it were possible to return a value from `py_run!`, I would suggest the following:

```suggestion
fn make_mapping(py: Python) -> &PyAny {
    let cls = py.get_type::<Mapping>();
    py_run!(py, cls, "cls(['1', '2', '3'])")
}
```

Use like

```
let mapping = make_mapping(py);
py_assert!(py, mapping, "mapping['1'] == 0");

davidhewitt (commented):
Thanks, I think it's a nice improvement!

Two ideas to make this feature even nicer:

- Support returning a value from `py_run!` (if it's possible? - maybe would need to be `py_eval!` macro for returning values)
- #1463

davidhewitt (approved):
(This is also good to merge as-is once the small nits addressed. The ideas could be follow-up PRs.)

src/lib.rs
kngwyu: Hmm 🤔, I think there are advantages and disadvantages to doing so.
Placing here would be clearer for someone who wants to read the implementation.

src/lib.rs
kngwyu: Thanks 👍🏼

kngwyu:
> Support returning a value from py_run! (if it's possible? - maybe would need to be py_eval! macro for returning values)

We need to use `py.eval` for returning a value.

davidhewitt:
> We need to use py.eval for returning a value.

👍. It would be awesome to have a `py_eval!` macro in that case, however I guess that's a separate PR.

PR #1495, opened 2021-03-14
macros-backend: make __wrap fns expressions

davidhewitt:
This is a tiny internal refactoring for the macros to make the "wrapper" functions generated as expressions. Reduces LOC just slightly and means that the `__wrap` name is less coupled between various parts of the macros codebase.


messense (approved):
Nice!

PR #1498, opened 2021-03-15
Fix wrong class name in function call error message

### LINKED ISSUE #1497: Wrong name in __new__ error message
### 🌍 Environment

 - Your operating system and version: macOS 11.2.2
 - Your python version: 3.7.5
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: pyenv virtualenv
 - Your Rust version (`rustc --version`): rustc 1.50.0 (cb75ad5db 2021-02-10)
 - Your PyO3 version: 0.13.2
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: no

### 💥 Reproducing

```bash
pip install crfs
```

https://github.com/messense/crfs-rs/blob/359b176f04d7353cc99a3c4dc5c49e594a568e14/python/src/lib.rs#L7-L16

```python
In [1]: from crfs import Attribute

In [2]: Attribute()
---------------------------------------------------------------------------
<ipython-input-2-f90d654a11d0> in <module>
----> 1 Attribute()

TypeError: PyAttribute.__new__() missing required positional argument: name
```

I'd expect it to be `TypeError: Attribute.__new__() missing required positional argument: name`, it should be `Attribute` instead of `PyAttribute` since I've specified `name` for it.

messense:
Fixes #1497

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -101,7 +101,7 @@ pub fn impl_wrap_cfunction_with_keywords(
             _args: *mut pyo3::ffi::PyObject,
             _kwargs: *mut pyo3::ffi::PyObject) -> *mut pyo3::ffi::PyObject
         {
-            const _LOCATION: &'static str = concat!(stringify!(#python_name), "()");
+            const _LOCATION: &'static str = stringify!(#python_name);
             pyo3::callback::handle_panic(|_py| {
                 #slf
                 let _args = _py.from_borrowed_ptr::<pyo3::types::PyTuple>(_args);
@@ -125,7 +125,7 @@ pub fn impl_wrap_noargs(cls: &syn::Type, spec: &FnSpec<'_>, self_ty: &SelfType) 
             _args: *mut pyo3::ffi::PyObject,
         ) -> *mut pyo3::ffi::PyObject
         {
-            const _LOCATION: &'static str = concat!(stringify!(#python_name), "()");
+            const _LOCATION: &'static str = stringify!(#python_name);
             pyo3::callback::handle_panic(|_py| {
                 #slf
                 #body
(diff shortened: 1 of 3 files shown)


pyo3-macros-backend/src/pymethod.rs
```diff
@@ -408,11 +406,18 @@ pub fn impl_arg_params(
...
+        quote! { None }
+    };
+
     // create array of arguments, and then parse
     Ok(quote! {
         {
             const DESCRIPTION: pyo3::derive_utils::FunctionDescription = pyo3::derive_utils::FunctionDescription {
-                fname: _LOCATION,
+                cls_name: #cls_name,
+                func_name: _LOCATION,
```
davidhewitt, line 420: How about writing `#python_name` here, and then adding the `()` inside either `full_name` or each of the error messages?

This is the only use of `_LOCATION` and it's always seemed like awkward coupling to me - if we remove this usage then we can also delete `_LOCATION` from all of the `__wrap` functions.

davidhewitt (changes requested):
Thanks, nice bugfix and refactoring!

I have a suggestion for a little bit more refactoring which I'd like to see...

pyo3-macros-backend/src/pymethod.rs
messense: > How about writing #python_name here, and then adding the () inside either full_name() or each of the error messages?

I've also thought about it. But it gets a little complicated with getters & setters: https://github.com/PyO3/pyo3/blob/6ab61a15609e81de9cab1d8cbe4033c24a4638c4/pyo3-macros-backend/src/pymethod.rs#L261-L279

davidhewitt: I'm pretty sure `_LOCATION` is actually already unused in getters / setters - what happens if you just delete it there?

messense: Done in https://github.com/PyO3/pyo3/pull/1498/commits/e6374240b1b78c614aa6a2c72219336c67d016b1

kngwyu (approved):
Nice 💯 , thank you!

PR #1502, opened 2021-03-16
Check buffer protocol support before getting buffer in sequence protocol specialization

messense:
This avoids calling a expensive [`PyErr_Format` inside of `PyObject_GetBuffer`](https://github.com/python/cpython/blob/cc02b4f2e810ab524d845daa18bc94df5b092dd8/Objects/abstract.c#L304) when buffer protocol is unsupported

### DIFF
--- a/src/types/sequence.rs
+++ b/src/types/sequence.rs
@@ -287,12 +287,14 @@ macro_rules! array_impls {
                 fn extract(obj: &'source PyAny) -> PyResult<Self> {
                     let mut array = [T::default(); $N];
                     // first try buffer protocol
-                    if let Ok(buf) = crate::buffer::PyBuffer::get(obj) {
-                        if buf.dimensions() == 1 && buf.copy_to_slice(obj.py(), &mut array).is_ok() {
+                    if unsafe { ffi::PyObject_CheckBuffer(obj.as_ptr()) } == 1 {
+                        if let Ok(buf) = crate::buffer::PyBuffer::get(obj) {
+                            if buf.dimensions() == 1 && buf.copy_to_slice(obj.py(), &mut array).is_ok() {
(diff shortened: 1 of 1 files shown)


src/types/sequence.rs
```diff
@@ -287,12 +287,14 @@ macro_rules! array_impls {
                 fn extract(obj: &'source PyAny) -> PyResult<Self> {
                     let mut array = [T::default(); $N];
                     // first try buffer protocol
-                    if let Ok(buf) = crate::buffer::PyBuffer::get(obj) {
-                        if buf.dimensions() == 1 && buf.copy_to_slice(obj.py(), &mut array).is_ok() {
+                    if unsafe { ffi::PyObject_CheckBuffer(obj.as_ptr()) } == 1 {
+                        if let Ok(buf) = crate::buffer::PyBuffer::get(obj) {
```
davidhewitt, line 291: I guess that this ultimately results in using `PyObject_CheckBuffer` twice? I wonder if we can have a `pub(crate) fn get_unchecked` method for `PyBuffer` to avoid this.

davidhewitt (commented):
Minor further suggestion.

Also tbh I'm not a fan of the `nightly` feature and probably wouldn't recommend it for users trying to get performance. Specialization is known to have soundness holes; I don't know if it's possible for users to hit them with our unstable implementations but I don't want to have to debug that 😅

src/types/sequence.rs
messense: I don't think so? `PyBuffer::get` doesn't call `PyObject_CheckBuffer` internally.

davidhewitt: 👍 ah ok

davidhewitt (approved):

Issue #287, opened 2018-11-26
Wrapping external rust crate


rth:
I have an external rust crate, and I would like to build some subset of it as a Python extension, all the while keeping the original crate independent from pyo3 (among other things due to the Rust nightly requirement). Is this currently possible and could you please outline the general approach?

For instance, taking the example from https://pyo3.rs/v0.5.2/class.html , but assuming that we have a struct with an implementation in an external crate that we would like to expose in Python. Is it possible to apply the `pyclass` macro to an external struct? If one needs to write some wrapper, how would it look roughly?  Thank you!

rth:
I tried the above suggestion, however,
```rust
#[pyclass]
pub struct PyWrapper(ExternalStruct);
```
produces 
```
error: custom attribute panicked
  --> src/lib.rs:31:1
   |
31 | #[pyclass]
   | ^^^^^^^^^^
   |
   = help: message: #[pyclass] can only be used with C-style structs

error: aborting due to previous error
```
where `ExternalStruct` is defined in a separate crate as,
```rust
pub struct ExternalStruct {
    lowercase: bool,
    token_pattern: String,
    n_features: u32,
}
```

This happens due to [this condition not passing in PyO3](https://github.com/PyO3/pyo3/blob/531f9db99e243678d7d4f48a1ff0a83e9695ea8b/pyo3-derive-backend/src/py_class.rs#L15) but I'm not sure what could be a way around it?

kngwyu:
Sorry, it looks tuple struct is currently unsupported :sweat: 
How about 
```rust
#[pyclass]
pub struct PyWrapper {
    inner: ExternalStruct,
}
```
?

rth:
Yes, that works, thank you @kngwyu !

messense:
```rust
#[pyclass]
pub struct PyWrapper(ExternalStruct);
```

@kngwyu  Is there any reason why tuple struct cannot be supported? Is it possible to add support for it?

davidhewitt:
From a technical level, I think it should be possible to support. Design-wise I think it makes sense to support it; the prevailing direction of pyO3 is (imo) write natural Rust code when writing Python bindings.

I think the first step would be to add support in the parsing code here: https://github.com/PyO3/pyo3/blob/acf72718792b52c5b3048d485dfec4fdbf930b48/pyo3-macros-backend/src/pyclass.rs#L182

[...]

PR #1504, opened 2021-03-18
Add tuple and unit struct support for pyclass macro

messense:
Fixes https://github.com/PyO3/pyo3/issues/287#issuecomment-800776622

### DIFF
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -506,18 +506,21 @@ fn impl_descriptors(
         .flat_map(|(field, fns)| {
             fns.iter()
                 .map(|desc| {
-                    let name = field.ident.as_ref().unwrap().unraw();
-                    let doc = utils::get_doc(&field.attrs, None, true)
-                        .unwrap_or_else(|_| syn::LitStr::new(&name.to_string(), name.span()));
-                    let property_type = PropertyType::Descriptor(&field);
-                    match desc {
-                        FnType::Getter(self_ty) => {
-                            impl_py_getter_def(cls, property_type, self_ty, &name, &doc)
+                    if let Some(name) = field.ident.as_ref().map(|ident| ident.unraw()) {
+                        let doc = utils::get_doc(&field.attrs, None, true)
+                            .unwrap_or_else(|_| syn::LitStr::new(&name.to_string(), name.span()));
+                        let property_type = PropertyType::Descriptor(&field);
+                        match desc {
+                            FnType::Getter(self_ty) => {
+                                impl_py_getter_def(cls, property_type, self_ty, &name, &doc)
+                            }
+                            FnType::Setter(self_ty) => {
+                                impl_py_setter_def(cls, property_type, self_ty, &name, &doc)
+                            }
+                            _ => unreachable!(),
                         }
-                        FnType::Setter(self_ty) => {
-                            impl_py_setter_def(cls, property_type, self_ty, &name, &doc)
-                        }
-                        _ => unreachable!(),
+                    } else {
+                        bail_spanned!(field.span() => "get/set are not supported on tuple struct field");
                     }
                 })
                 .collect::<Vec<syn::Result<TokenStream>>>()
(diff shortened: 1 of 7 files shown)


tests/test_class_basics.rs
```diff
@@ -18,6 +18,20 @@ fn empty_class() {
     py_assert!(py, typeobj, "typeobj.__name__ == 'EmptyClass'");
 }
 
+#[pyclass]
+struct UnitClass;
+
+#[test]
+fn unit_class() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
```
kngwyu, line 27: For new tests, please use `Python::with_gil` instead.

tests/test_getter_setter.rs
```diff
@@ -130,3 +130,31 @@ fn ref_getter_setter() {
...
+}
+
+#[test]
+fn tuple_struct_getter_setter() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+
+    let inst = Py::new(py, TupleClassGetterSetter(10)).unwrap();
+
+    py_run!(py, inst, "assert inst.num == 10");
```
kngwyu, line 157:
```suggestion
    py_assert!(py, inst, "inst.num == 10");
```

tests/test_getter_setter.rs
```diff
@@ -130,3 +130,31 @@ fn ref_getter_setter() {
...
+    py_run!(py, inst, "inst.num = 20");
+    py_run!(py, inst, "assert inst.num == 20");
```
kngwyu, line 159: Same as above.

kngwyu (commented):
Nice, thank you!

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -497,18 +506,21 @@ fn impl_descriptors(
...
+                                impl_py_getter_def(cls, property_type, self_ty, &name, &doc)
+                            }
+                            FnType::Setter(self_ty) => {
+                                impl_py_setter_def(cls, property_type, self_ty, &name, &doc)
+                            }
+                            _ => unreachable!(),
                         }
-                        _ => unreachable!(),
+                    } else {
+                        bail_spanned!(field.span() => "get/set are not supported on tuple struct field");
```
davidhewitt, line 523: I'm curious what the reason for disallowing `get/set` is? I was thinking that it should be possible, but a `name` would be required.

e.g. this seems a plausible combination to me:

```rust
#[pyclass]
struct PyFoo(
    #[pyo3(get, set, name = "inner", from_py_with = "pytimestamp_to_int")] i32,
);
```

I'm happy for us to create an issue to leave this for a follow-up PR.

tests/test_class_basics.rs
```diff
@@ -18,6 +18,20 @@ fn empty_class() {
...
 
+#[pyclass]
+struct UnitClass;
+
+#[test]
+fn unit_class() {
+    Python::with_gil(|py| {
+        let typeobj = py.get_type::<UnitClass>();
+        // By default, don't allow creating instances from python.
+        assert!(typeobj.call((), None).is_err());
```
davidhewitt, line 29: 👍

davidhewitt (commented):
👍 Thanks for implementing this! Changes look good to me, though I have one follow-up thought around `#[pyo3(get, set)]`.

Needs a CHANGELOG entry.

pyo3-macros-backend/src/pyclass.rs
messense: It was disallowed just to avoid proc-maco panic when user try to use `get/set` on it. We can certainly add support for it, I think we can do the follow-up after #1507 merged.

kngwyu:
Thanks!

kngwyu (approved):

pyo3-macros-backend/src/pyclass.rs
austin362667: Hi @davidhewitt  and @messense  This's nice!  And as a local wrapper of external structure, I think we still can't get them work with nested getter and setter, which is `get_all` and `set_all`?

PR #1505, opened 2021-03-18
Ignore `syn::Type::Group` in `is_python`

scalexm:
Macro expansion can add invisible delimiters around types, modelled by `syn::Type::Group` in `syn`. This causes the following code to fail to compile:
```rust
#[pyclass]
struct Foo {}

macro_rules! foo {
    (
        #[pymethods]
        impl $ty: ty {
            fn $fn:ident (&self, $arg:ident : $arg_ty:ty) {}
        }
    ) => {
        #[pymethods]
        impl $ty {
            fn $fn(&self, $arg: $arg_ty) {}
        }
    };
}

foo!(
    #[pymethods]
    impl Foo {
        fn foo(&self, py: Python<'_>) {}
    }
);
```

> error[E0277]: the trait bound `pyo3::Python<'_>: PyClass` is not satisfied

because `pyo3` does not "see" the special `py` argument.


davidhewitt (approved):
Thanks! I would say this needs a CHANGELOG entry, though I find it hard to express what the end-user benefit it.

Maybe "Fix use of `Python` argument for `#[pymethods]` inside macro expansions" ?

davidhewitt:
I'm going to push a CHANGELOG entry and test this morning quickly.

messense (approved):

PR #1507, opened 2021-03-18
Allow `#[name]` with `#[getter]` and `#[setter]`

scalexm:
A custom name can now be specified either via `#[getter(custom_name)]` or `#[name = "custom_name"]` like normal methods.

It's easier to have this kind of uniformity when dealing with code generation.

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -468,7 +468,16 @@ fn parse_method_attributes(
     *attrs = new_attrs;
 
     let python_name = if allow_custom_name {
-        parse_method_name_attribute(ty.as_ref(), attrs, property_name)?
+        match parse_method_name_attribute(ty.as_ref(), attrs)? {
+            Some(python_name) if property_name.is_some() => {
+                return Err(syn::Error::new_spanned(
+                    python_name,
+                    "name cannot be specified twice",
+                ));
+            }
+            Some(python_name) => Some(python_name),
+            None => property_name,
+        }
     } else {
         property_name
     };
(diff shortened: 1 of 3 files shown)


davidhewitt:
Thanks, this seems reasonable to me. I would actually like to ask your opinion on a further change relating to this. 

I was thinking of moving the `#[name]` attribute under the pyo3 namespace, e.g. `#[pyo3(name = "foo")]`. I wanted to make this change _everywhere_ that the `#[name]` attribute can currently be used (keeping the old form for compatibility for a little while).

This could compose with `#[getter]` in a slightly different form: `#[pyo3(getter, name = "foo")]`. See other examples where this is relevant e.g. https://github.com/PyO3/pyo3/pull/1504#discussion_r597162765

So the options would be

```rust
#[pymethods]
impl Foo {
    #[getter(custom_name)]
    fn get_foo(&self) { }

    // or

    #[pyo3(getter, name = "custom_name")]
    fn get_bar(&self) { }
}
```

The later one is longer, but I think much clearer what is going on.

scalexm:
Would you want to also allow multiple `#[pyo3(...)]` attributes like the following:
```rust
#[pyo3(getter)]
#[pyo3(name="custom_name")]
fn foo(&self) -> i32 { ... {
```
If yes, then that’s definitely a change I would support. This avoids clashing with other proc macros and adds uniformity with the existing `#[pyo3(get, set)]` on struct fields.

The ability to have the two attributes separately makes my life easier when generating code: in my case, adding a `#[name]` on every item through a macro.

davidhewitt:
> If yes, then that’s definitely a change I would support.

Awesome! Are you willing to support the new syntax now in this PR, or you want to leave this PR as-is and I'll make a change to use the new syntax soon?

scalexm:
Well, to be honest I’m in lack of a personal computer right now so my ability to write code in my free time is severely decreased :)

davidhewitt:
👍 let's merge this PR and I'll write the follow-up!

davidhewitt (approved):

kngwyu (approved):
Thanks!

kngwyu:
> I was thinking of moving the #[name] attribute under the pyo3 namespace, e.g. #[pyo3(name = "foo")].

I think it can be better, but first, shouldn't we summarize what is under the `#[pyo3(...)]` and what is not in an issue?

PR #1510, opened 2021-03-20
Correct `# Examples` in documents

kngwyu:
https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

It looks like that `# Examples` is recommended even for an example.


davidhewitt (approved):
Agreed, I think this is a good change for consistency!

PR #1511, opened 2021-03-20
Use with_gil instead of acquire_gil in examples

kngwyu:
We should encourage this.


messense (approved):
LGTM 💯

src/marshal.rs
```diff
@@ -21,15 +21,14 @@ pub const VERSION: i32 = 4;
 /// # Examples
```
messense, line 21: This file seems accidentally changed mode from `644` to `755`

kngwyu: Oops, thank you for pointing out that!

PR #1514, opened 2021-03-22
Ensure we are cross compiling when any cross env variables are set.

jameshilliard:
Potential partial fix for #1513.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix inability to use a named lifetime for `&PyTuple` of `*args` in `#[pyfunction]`. [#1440](https://github.com/PyO3/pyo3/pull/1440)
 - Fix inability to add `#[text_signature]` to some `#[pyproto]` methods. [#1483](https://github.com/PyO3/pyo3/pull/1483)
 - Fix use of Python argument for #[pymethods] inside macro expansions. [#1505](https://github.com/PyO3/pyo3/pull/1505)
+- Always use cross-compiling configuration if any of the environment variables are set. [#1514](https://github.com/PyO3/pyo3/pull/1514)
 
 ## [0.13.2] - 2021-02-12
 ### Packaging
--- a/build.rs
+++ b/build.rs
@@ -151,36 +151,44 @@ impl CrossCompileConfig {
 }
 
 fn cross_compiling() -> Result<Option<CrossCompileConfig>> {
-    let target = env::var("TARGET")?;
-    let host = env::var("HOST")?;
-    if target == host {
-        // Not cross-compiling
-        return Ok(None);
-    }
+    if env::var_os("PYO3_CROSS").is_none()
+        && env::var_os("PYO3_CROSS_LIB_DIR").is_none()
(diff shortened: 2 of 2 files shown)


kngwyu:
Thanks, 
- Did you confirm that this works for your case?
- What do you mean by 'partial'? What problem do you think remains?

davidhewitt (commented):
@jameshilliard thanks for this - can you check if this  patch in conjunction with the suggested `LD_LIBRARY_PATH` setting might provide enough of a working solution to unblock you?

This could also use a CHANGELOG entry - e.g. "Fixed - Always use cross-compiling configuration if any of the environment variables are set".

jameshilliard:
> Did you confirm that this works for your case?

At the moment I haven't been able to fully validate this properly due to the cargo issue, the rust toolchain is rather exotic and buildroot's rust support is different from most other language toolchains we use(mostly due to having to use prebuilt rust toolchains since rust can't easily be bootstrapped due to the rust requires rust to build issue) so I'm having difficulty testing stuff that requires modifications to stuff deep in the dependency tree(I'm quite familiar with python package cross compilation in buildroot since I help maintain a good number of python packages there but have not worked much with rust). In any case having this merged should make it easier for me to isolate the cargo issue since env variables are generally quite easy to control with buildroot.

> What do you mean by 'partial'? What problem do you think remains?

Well it looks like the main other issue is the cargo issue, but this best I can tell fixes the most obvious cross compilation detection logic issue I saw in pyo3. This at least shouldn't cause regressions since these env variables seem to be exclusively used in cross compilation enabled codepaths, so if any of them are set we should be able to safely assume that cross compilation should be unconditionally enabled.

> can you check if this patch in conjunction with the suggested `LD_LIBRARY_PATH` setting might provide enough of a working solution to unblock you?

Well the `LD_LIBRARY_PATH` workaround doesn't really look like it's viable to upstream for buildroot to me as we do in fact support compiling for x86_64 targets with newer instructions sets than the build host(buildroot uses a custom autobuilder CI system with older build hosts to ensure this sort of thing isn't broken). I'll play around with this a little though.

> This could also use a CHANGELOG entry - e.g. "Fixed - Always use cross-compiling configuration if any of the environment variables are set".

Added.

davidhewitt (approved):
👍 I'm satisfied that this patch is correct; and while there's other upstream issues at least we can get one step closer with this in place.

Thanks for the contribution!

jameshilliard:
> 👍 I'm satisfied that this patch is correct; and while there's other upstream issues at least we can get one step closer with this in place.

Great, by the way I made a few attempts to go through the cargo codebase to try and isolate how it detects cross compilation but so far I haven't been able to understand the internal logic all that well.

Would you happen to know if there's a good place to find someone familiar with those internals enough to give me some pointers to help me better understand how the auto-detection there currently works?

Best I can tell none of the longstanding github issues relating to this issue that were opened against the cargo or rust repos have been given any serious attention by anyone who has a good understanding of what's going on internally so far.

davidhewitt:
You could try the `t-cargo` stream on the [Rust zulip server](https://rust-lang.zulipchat.com)?

jameshilliard:
For reference I have an open pull request to fix the related cargo bug in [#9322](https://github.com/rust-lang/cargo/pull/9322).

PR #1516, opened 2021-03-23
build.rs: adjust pypy warning wording

davidhewitt:
Strictly speaking, the pyo3 build step just produces a `.so` file - it's up to the packaging tool (`setuptools-rust` or `maturin`, most likely) to name the `.so` file and the wheel, which really determines whether pypi considers it a "version-specific" wheel.

So I think the wording in the pypy abi3 warning could be tweaked slightly.


messense (approved):
LGTM

PR #1518, opened 2021-03-25
Always use correct abi3 Python DLL name for MinGW

ravenexp:
Compiling an abi3 extension module for `x86_64-pc-windows-gnu` target
links to `python3.dll` import library when `PYO3_NO_PYTHON` is set
and to `python3.Y.dll` import library when `PYO3_NO_PYTHON` is not set.

All abi3 extensions should link to `python3.dll` on Windows,
as required by

https://www.python.org/dev/peps/pep-0384/#linkage

Update the code path for the case when `PYO3_NO_PYTHON` is not set
to yield the same Python DLL import library name.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -625,26 +625,23 @@ fn run_python_script(interpreter: &Path, script: &str) -> Result<String> {
 
 fn get_rustc_link_lib(config: &InterpreterConfig) -> String {
     let link_name = if env::var("CARGO_CFG_TARGET_OS").unwrap().as_str() == "windows" {
-        if env::var("CARGO_CFG_TARGET_ENV").unwrap().as_str() == "gnu" {
+        if is_abi3() {
+            // Link against python3.lib for the stable ABI on Windows.
+            // See https://www.python.org/dev/peps/pep-0384/#linkage
+            //
+            // This contains only the limited ABI symbols.
+            "pythonXY:python3".to_owned()
+        } else if env::var("CARGO_CFG_TARGET_ENV").unwrap().as_str() == "gnu" {
             // https://packages.msys2.org/base/mingw-w64-python
-            // TODO: ABI3?
             format!(
                 "pythonXY:python{}.{}",
                 config.version.major, config.version.minor
             )
         } else {
-            // Link against python3.lib for the stable ABI on Windows.
-            // See https://www.python.org/dev/peps/pep-0384/#linkage
-            //
-            // This contains only the limited ABI symbols.
-            if is_abi3() {
-                "pythonXY:python3".to_owned()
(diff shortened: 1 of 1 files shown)


messense:
Close and reopen to trigger CI.

kngwyu (approved):
Sounds reasonable, thank you!

MehdiChinoune:
On MSYS2 we have only libpythonX.Y.dll/dll.a libs.
Building python-cryptography fails because of this change.
`lld: error: unable to find library -lpython3`
`D:/M/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lpython3: No such file or directory`

Is there a workaround to ask for libpythonX.Y?

davidhewitt:
I think you could write a custom `PYO3_CONFIG_FILE` (see https://pyo3.rs/v0.16.4/building_and_distribution.html#advanced-config-files).

You might also consider filling an upstream msys2 bug?

ravenexp:
It links to `python3.dll.a` because you are trying to compile an `abi3` extension, which do not seem to be supported by the MSYS2 Python distribution.

If you disable the `abi3` feature it should link to the version-specific library the PyO3 build script discovers via `_sysconfigdata*.py` parsing. This should work on MSYS2, I think. If it doesn't, please open an issue.

davidhewitt:
@ravenexp what you say is true, but I think in practice it should also be safe to link the version-specific library using only the abi3 apis? The build just wouldn't be portable between versions due to the link name.

ravenexp:
> @ravenexp what you say is true, but I think in practice it should also be safe to link the version-specific library using only the abi3 apis? The build just wouldn't be portable between versions due to the link name.

I also think it should work, but it's kinda.. pointless. You still need to build one extension DLL per supported Python version, so why give up performance by restricting PyO3 to the stable API?

MehdiChinoune:
> If you disable the `abi3` feature it should link to the version-specific library the PyO3 build script discovers via `_sysconfigdata*.py` parsing. This should work on MSYS2, I think.

I looks like It links after disabling `abi3`.
>If it doesn't, please open an issue.

It couldn't find `cryptography_rust.so` (unix extension).

`error: Rust build failed; unable to find any *cryptography_rust.so in C:\_\mingw-w64-python-cryptography\src\cryptography-37.0.4\src\rust\target/release`

Should I open an issue?

davidhewitt:
Was probably fixed by https://github.com/PyO3/setuptools-rust/pull/267 - just needs a new `setuptools-rust` release.

PR #1519, opened 2021-03-25
Always set `Py_SHARED` when building for Windows

ravenexp:
Setting `PYO3_NO_PYTHON` results in an extension module built with
a different config feature set (on Windows).
Specifically, `Py_SHARED` config option gets omitted in this code path.

Maturin always sets `PYO3_NO_PYTHON` when cross-compiling abi3 extensions,
which creates a subtle configuration mismatch between `cargo build`
and `maturin build` artifacts.

Always set `Py_SHARED` when compiling abi3 extensions for Windows.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -886,13 +886,17 @@ fn abi3_without_interpreter() -> Result<()> {
     println!("cargo:rustc-cfg=py_sys_config=\"WITH_THREAD\"");
     println!("cargo:python_flags={}", flags);
 
-    // Unfortunately, on windows we can't build without at least providing
-    // python.lib to the linker. While maturin tells the linker the location
-    // of python.lib, we need to do the renaming here, otherwise cargo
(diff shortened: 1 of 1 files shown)


ravenexp:
> Hmm but `PYO3_NO_PYTHON` should not be used for Windows, I think...

Well, it _may_ be used, provided that the Python DLL import library path is somehow supplied separately via `RUSTFLAGS`.
It's even mentioned in the user guide:
https://pyo3.rs/v0.13.2/building_and_distribution.html

Maturin does just that when cross-compiling abi3 extensions for Windows.
See the comments I moved into the condition block:
they led me to believe that this use case was already supported.
I just changed the config defaults so that `cargo build` and `maturin build` yield consistent results
whether `PYO3_NO_PYTHON`  is set or not.

kngwyu (approved):
Ah, OK, I just forgot what #1371 did.
cc: @konstin

kngwyu:
Thanks!

PR #1520, opened 2021-03-26
Fix clippy warnings in Rust 1.51.0

messense:


messense:
```
error: `#[link_name]` is ignored on struct fields, match arms and macro defs
   --> src/ffi/datetime.rs:368:22
    |
368 |     #[cfg_attr(PyPy, link_name = "_PyPyDate_FromDate")]
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D unused-attributes` implied by `-D warnings`
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: see issue #80564 <https://github.com/rust-lang/rust/issues/80564> for more information
```

I'm not sure about this so I've left it unchanged.

src/class/gc.rs
```diff
@@ -10,12 +10,16 @@ use std::os::raw::{c_int, c_void};
 pub struct PyTraverseError(c_int);
 
 /// GC support
+#[allow(clippy::upper_case_acronyms)]
```
birkenfeld, line 13: Should this just become a global allow? It is pretty common here (and nothing we can do about it).

davidhewitt (approved):
Thanks!

> I'm not sure about this so I've left it unchanged.

I'm 99% certain that those can just be deleted - as the error message says, they're ignored.

PR #1523, opened 2021-03-26
Refactor doctest

kngwyu:
- Use `#[cfg(doctest)]` instead of `#[doc(hidden)]`
- Some naming tweaks in the `doctest!` macro


davidhewitt (approved):
Thanks, a nice tidy up which reduces the amount of clutter for downstream users!

PR #1525, opened 2021-03-28
1506: fixes to macro hygiene

### LINKED ISSUE #1506: Hygiene issues with `quote_spanned!`
Commit fe74160a from #1440 introduced the use of `quote_spanned!` in a few places. Hygiene rules specific to `quote_spanned!` make the following macro expansion to fail:
```rust
#[pyclass]
struct Foo {}

macro_rules! foo {
    (#[pymethods] $($body:tt)*) => {
        #[pymethods]
        $($body)*
    };
}

foo!(
    #[pymethods]
    impl Foo {
        fn foo(&self, py: Python<'_>) {}
    }
);
```

> error[E0425]: cannot find value `_py` in this scope

Was there a legitimate reason for these uses of `quote_spanned!`? Otherwise removing them altogether solves the problem.

davidhewitt:
Fixes #1506 

Introducing `quote_spanned` improved the error messages for a number of common cases, but it came at the downside of changing the span of parts of our generated code.

If not careful, when expanding macros we can get issues where some of the code comes from one context and the rest from another.

This PR makes two changes to resolve some cases identified in #1506:
 - The `_py` ident is created once and shared around the macro code, so that it's not possible to make `_py` expand in two different contexts by accident. Similar for `output`.
 - A few more uses of `quote_spanned!` inside `impl_arg_param`.

### DIFF
--- a/pyo3-macros-backend/src/module.rs
+++ b/pyo3-macros-backend/src/module.rs
@@ -225,17 +225,18 @@ fn function_c_wrapper(
         };
         slf_module = None;
     };
-    let body = impl_arg_params(spec, None, cb)?;
+    let py = syn::Ident::new("_py", Span::call_site());
+    let body = impl_arg_params(spec, None, cb, &py)?;
     Ok(quote! {
         unsafe extern "C" fn #wrapper_ident(
             _slf: *mut pyo3::ffi::PyObject,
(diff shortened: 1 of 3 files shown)


davidhewitt:
No CHANGELOG necessary as this bug only exists on master!

Issue #1522, opened 2021-03-26
cross-compiling to wasm32-wasi


cymqqqq:
## 🐛 Bug Reports

When reporting a bug, please provide the following information. If this is not a bug report you can just discard this template.

### 🌍 Environment

 - Your operating system and version:
 - Your python version:
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?:
 - Your Rust version (`rustc --version`):
 - Your PyO3 version:
 - Have you tried using latest PyO3 main (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`:

### 💥 Reproducing

Please provide a [minimal working example](https://stackoverflow.com/help/mcve). This means both the Rust code and the Python.

Please also write what exact flags are required to reproduce your results.
ubuntu 20.04
py3.7
install python by apt. Did you use a virtualenv? No.
rust version  1.42
pyo3 version 0.13.2
yes, tried  `git = "https://github.com/PyO3/pyo3"
when I build my demo code, an error comes:
error: failed to run custom build command for `pyo3 v0.13.2`

Caused by:
  process didn't exit successfully: `/home/azureuser/rustpy/target/release/build/pyo3-c4d90697e291dc3a/build-script-build` (exit code: 1)
  --- stderr
  Error: NotPresent
warning: build failed, waiting for other jobs to finish...
error: build failed
Error: Compiling your crate to WebAssembly failed
Caused by: failed to execute `cargo build`: exited with exit code: 101
  full command: "cargo" "build" "--all-targets" "--release" "--target" "wasm32-wasi"

davidhewitt:
Hi @cymqqqq can you please share your demo code?

cymqqqq:
ok, here is my demo code:
use pyo3::prelude::*;
use pyo3::{prelude::*, types::{IntoPyDict, PyModule}};
use wasm_bindgen::prelude::*;
#[wasm_bindgen]
Python::with_gil(|py| {
        let activators = PyModule::from_code(py, r#"
        def relu(x):
            """see https://en.wikipedia.org/wiki/Rectifier_(neural_networks)"""
                return max(0.0, x)

                def leaky_relu(x, slope=0.01):
                    return x if x >= 0 else x * slope
                        "#, "activators.py", "activators")?;

                        let relu_result: f64 = activators.getattr("relu")?.call1((-1.0,))?.extract()?;
                            assert_eq!(relu_result, 0.0);

[...]

PR #1529, opened 2021-03-29
Clean up CARGO_CFG_TARGET_FAMILY variable parsing 

ravenexp:
Improves error reporting in e.g. https://github.com/PyO3/pyo3/issues/1522

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -144,8 +144,6 @@ impl CrossCompileConfig {
 fn cross_compiling() -> Result<Option<CrossCompileConfig>> {
     if env::var_os("PYO3_CROSS").is_none()
         && env::var_os("PYO3_CROSS_LIB_DIR").is_none()
-        && env::var_os("PYO3_CROSS_INCLUDE_DIR").is_none()
-        && env::var_os("PYO3_CROSS_VERSION").is_none()
         && env::var_os("PYO3_CROSS_PYTHON_VERSION").is_none()
     {
         let target = env::var("TARGET")?;
@@ -489,7 +487,7 @@ fn windows_hardcoded_cross_compile(
         ) {
             (Some(major), Some(minor), None) => (major, minor),
             _ => bail!(
-                "Expected major.minor version (e.g. 3.9) for PYO3_CROSS_VERSION, got `{}`",
+                "Expected major.minor version (e.g. 3.9) for PYO3_CROSS_PYTHON_VERSION, got `{}`",
                 version
             ),
         }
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -518,13 +516,15 @@ fn windows_hardcoded_cross_compile(
 fn load_cross_compile_info(
     cross_compile_config: CrossCompileConfig,
 ) -> Result<(InterpreterConfig, BuildFlags)> {
-    let target_family = env::var("CARGO_CFG_TARGET_FAMILY")?;
-    // Because compiling for windows on linux still includes the unix target family
-    if target_family == "unix" {
+    match env::var("CARGO_CFG_TARGET_FAMILY") {
```
davidhewitt, line 519: Hmm the `Err` can include utf8 decoding errors too. We can avoid those using `var_os`:

```suggestion
    match env::var_os("CARGO_CFG_TARGET_FAMILY") {
```

build.rs
```diff
@@ -518,13 +516,15 @@ fn windows_hardcoded_cross_compile(
...
-        load_cross_compile_from_sysconfigdata(cross_compile_config)
-    } else {
-        windows_hardcoded_cross_compile(cross_compile_config)
+        Ok(os) if os == "unix" => load_cross_compile_from_sysconfigdata(cross_compile_config),
+        // Use hardcoded interpreter config when targeting Windows
+        Ok(os) if os == "windows" => windows_hardcoded_cross_compile(cross_compile_config),
+        // ???
+        Ok(os) => Err(format!("Unsupported target OS family: {}", os).into()),
+        // wasm32-wasi / bare metal ?
+        Err(_) => Err("CARGO_CFG_TARGET_FAMILY is not set when cross-compiling".into()),
```
davidhewitt, line 527: With `var_os`:

```suggestion
        Some(os) if os == "unix" | None => load_cross_compile_from_sysconfigdata(cross_compile_config),
        // Use hardcoded interpreter config when targeting Windows
        Some(os) if os == "windows" => windows_hardcoded_cross_compile(cross_compile_config),
        // ???
        Some(os) => bail!("Unsupported target OS family for cross-compilation: {:?}", os),
```

I also moved the "missing" case to just load sysconfigdata - I don't see any harm in trying to load that, and it's probably better than just giving up.

davidhewitt (commented):
Thanks for the tidy up!

Couple of suggestions on the `match` statement...

ravenexp:
Oh, I expected everything coming from Cargo to be UTF-8-safe, but sure, it looks cleaner this way.

davidhewitt (approved):
This looks great now. Thanks very much for the PR!

davidhewitt:
> I expected everything coming from Cargo to be UTF-8-safe

I would think Cargo is, but I have no idea if a user can ever override these variables with non-utf8 content 🤷

Issue #416, opened 2019-03-23
Static Build of pyo3 (running without python in users system)


uncotion:
I would like to run python from rust, but in my situation, end users who will run the program, haven't preinstalled python and it's not possible to install (their system is extra restricted).
I possible build a rust program that uses pyo3 will <U>embed</U> python interpreter? (So that users don't need to install a python interpreter separately.)?
I is it possible please give me a hit, In documentation there wan't any point to this.

konstin:
That's currently not supported, though it would be possible. I've written the necessary instructions for linux in the fourth of #276

konstin:
To implement this `cargo:rustc-link-search=native` must be set to the value of LIBPL in the python sysconfig and `cargo:rustc-link-lib=static=` must be used instead of `cargo:rustc-link-lib`. It should be possible to enable this with a feature flag.

hfingler:
Did anyone have any success related to this? I'm about to try this and any information would be helpful. Will report back with results.

tr4cks:
Hello everyone,

I was also interested in this feature. I looked into it and I thought of directly integrating the embedded library provided by python rather than compiling it statically. I think it's much easier to deploy since you just have to download the python library in portable format and integrate it into your own project.
I first tried to indicate the python executable via the `PYTHON_SYS_EXECUTABLE` environment variable but this solution doesn't work. It doesn't work because the method used to know the path of the python library is to execute a python script and get the variable returned by the function named `sysconfig.get_config_var('LIBDIR')`. However, in the case of the portable python library, this function returns nothing.

To do this, I thought of providing another environment variable allowing to indicate the archive containing the portable python executable. This would then be unzipped into the target/{...}/out folder and automatically linked to the project.

Tell me what you think about this?

jmwright:
Has there been any progress on this? I see `cargo:rustc-link-lib=static=` in a couple of places in build.rs.

https://github.com/PyO3/pyo3/blob/master/build.rs#L333

davidhewitt:
I investigated this briefly, but I think even with the static linking "supported" we have plenty of missing symbol errors which are seen in other tickets #763 #742 . 

I believe that the problem is that unused symbols get removed at link-time, which is problematic when loading other Python extensions from a statically embedded Python interpreter.

I asked about this on URLO, but got no response: https://users.rust-lang.org/t/embed-whole-python-interpreter-using-static-linking/42509

Not sure if to get this working properly we need to add changes to cargo / rustc.

[...]

PR #1532, opened 2021-03-29
Add `PYO3_CROSS_NO_ENABLE_SHARED` variable

ravenexp:
It allows to override the default value of `Py_ENABLE_SHARED`
definition in `pyconfig.h` when cross-compiling for Windows.

Equivalent to defining `Py_NO_ENABLE_SHARED` when building
a C program statically linking the Python interpreter.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -145,6 +145,7 @@ fn cross_compiling() -> Result<Option<CrossCompileConfig>> {
     if env::var_os("PYO3_CROSS").is_none()
         && env::var_os("PYO3_CROSS_LIB_DIR").is_none()
         && env::var_os("PYO3_CROSS_PYTHON_VERSION").is_none()
+        && env::var_os("PYO3_CROSS_NO_ENABLE_SHARED").is_none()
     {
         let target = env::var("TARGET")?;
         let host = env::var("HOST")?;
@@ -503,10 +504,19 @@ fn windows_hardcoded_cross_compile(
 
     let python_version = PythonVersion { major, minor };
 
+    let is_extension_module = env::var_os("CARGO_FEATURE_EXTENSION_MODULE").is_some();
+    let no_enable_shared = env::var_os("PYO3_CROSS_NO_ENABLE_SHARED").is_some();
+
+    if no_enable_shared && is_extension_module {
+        warn!(
+            "PYO3_CROSS_NO_ENABLE_SHARED cannot be used for extension modules and will be ignored"
+        );
+    }
+
     let interpreter_config = InterpreterConfig {
         version: python_version,
         libdir: cross_compile_config.lib_dir.to_str().map(String::from),
-        shared: true,
+        shared: !no_enable_shared || is_extension_module,
         ld_version: format!("{}.{}", major, minor),
         base_prefix: "".to_string(),
         executable: PathBuf::new(),
@@ -904,6 +914,7 @@ fn main_impl() -> Result<()> {
         "PYO3_PYTHON",
         "PYO3_CROSS_LIB_DIR",
         "PYO3_CROSS_PYTHON_VERSION",
+        "PYO3_CROSS_NO_ENABLE_SHARED",
     ] {
         println!("cargo:rerun-if-env-changed={}", var);
     }
--- a/guide/src/building_and_distribution.md
+++ b/guide/src/building_and_distribution.md
@@ -119,6 +119,11 @@ cargo build --target x86_64-pc-windows-gnu
 
 Any of the `abi3-py3*` features can be enabled instead of setting `PYO3_CROSS_PYTHON_VERSION` in the above examples.
 
+As an advanced feature, `PYO3_CROSS_NO_ENABLE_SHARED` can be set to enable support for static linking
+of the Python interpreter library when targeting Windows.
+It can only be used for building Rust applications embedding a Python interpreter,
+and is ignored for the Python extension modules (when `extension-module` feature is enabled).
+
 ## Bazel
 
 For an example of how to build python extensions using Bazel, see https://github.com/TheButlah/rules_pyo3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - `PyObject_Check`, `PySuper_Check`, and `FreeFunc` [#1438](https://github.com/PyO3/pyo3/pull/1438)
 - Remove pyclass implementation details `Type`, `DESCRIPTION`, and `FLAGS` from `PyTypeInfo`. [#1456](https://github.com/PyO3/pyo3/pull/1456)
 - Remove `__doc__` from module's `__all__`. [#1509](https://github.com/PyO3/pyo3/pull/1509)
-- Remove `PYO3_CROSS_INCLUDE_DIR` environment variable and the associated C header parsing functionality.
+- Remove `PYO3_CROSS_INCLUDE_DIR` environment variable and the associated C header parsing functionality. [#1521](https://github.com/PyO3/pyo3/pull/1521)
 
 ### Fixed
 - Remove FFI definition `PyCFunction_ClearFreeList` for Python 3.9 and later. [#1425](https://github.com/PyO3/pyo3/pull/1425)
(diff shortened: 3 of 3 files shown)


build.rs
```diff
@@ -501,10 +501,19 @@ fn windows_hardcoded_cross_compile(
 
     let python_version = PythonVersion { major, minor };
 
+    let is_extension_module = env::var_os("CARGO_FEATURE_EXTENSION_MODULE").is_some();
+    let no_enable_shared = env::var_os("PYO3_CROSS_NO_ENABLE_SHARED").is_some();
+
+    if no_enable_shared && is_extension_module {
+        warn!(
+            "PYO3_CROSS_NO_ENABLE_SHARED can not be used for extension modules and will be ignored"
```
davidhewitt, line 509: Nit:

```suggestion
            "PYO3_CROSS_NO_ENABLE_SHARED cannot be used for extension modules and will be ignored"
```

guide/src/building_and_distribution.md
```diff
@@ -118,6 +118,11 @@ cargo build --target x86_64-pc-windows-gnu
 
 Any of the `abi3-py3*` features can be enabled instead of setting `PYO3_CROSS_PYTHON_VERSION` in the above examples.
 
+As an advanced feature, `PYO3_CROSS_NO_ENABLE_SHARED` can be set to enable support for static linking
+of the Python interpreter library when targeting Windows.
+It can only be used for building Rust applications embedding a Python interpreter,
+and is ignored for the Python extension modules (when `extension-module` feature is enabled).
```
davidhewitt, line 124:
```suggestion
It is only relevant when building Rust applications embedding a Python interpreter,
and so is ignored for Python extension modules (when the `extension-module` feature is enabled).
```

We should also comment on the downsides here - only symbols used in the program will be included, so importing modules in the embedded program will often fail.

davidhewitt (commented):
Thanks for implementing this. I have a couple of comments regarding the copy.

Also, for this to be useful for users we would need to remove the cfg-gates I put in `gil.rs`:

https://github.com/PyO3/pyo3/blob/f97aac64ec273d25aea2d7e196618fb2f22eee10/src/gil.rs#L73

https://github.com/PyO3/pyo3/blob/f97aac64ec273d25aea2d7e196618fb2f22eee10/src/gil.rs#L140

However I'm a bit worried about doing so. I added those because we got a lot of users having problems with statically-linked tests. 😄

guide/src/building_and_distribution.md
ravenexp: On Windows extensions can only import symbols from a Python DLL (either python3y.dll or python3.dll). When the interpreter is built as a static library no interface DLL is created, and dynamically loading extension modules will not work at all. This is different from *nix, where the extensions can grab symbols from the executable.

ravenexp:
> Also, for this to be useful for users we would need to remove the cfg-gates I put in `gil.rs`:
> ...
> However I'm a bit worried about doing so. I added those because we got a lot of users having problems with statically-linked tests.

Sorry, I don't quite understand why it is gated on `Py_SHARED` and not for example on `py_sys_config = "WITH_THREAD"`.

I grepped the current CPython sources for `Py_ENABLE_SHARED` and `Py_NO_ENABLE_SHARED` uses, and it seems to be mostly dealing with `__declspec(...)` shenanigans in the Windows build. It also affect the default module search path on Windows: apparently it is relative to the Python DLL location. I couldn't find any uses that would affect the Unix-like builds.

davidhewitt:
Yes, those `#[cfg]` are not motivated by upstream cpython limitations. I put them there because embedding the Python interpreter statically accidentally has caused lots of problems for users in the past.

See original PR in #1347

davidhewitt:
.... I am wondering if perhaps the right thing to do is remove those `#[cfg]` attributes and fix this instead with better documentation. I think it's still correct to disable `auto-initialize` with static linking as that'll prevent most of the accidents, however we could alter the error message for this case to be better - especially now that the `auto-initialize` feature is opt-in.

I'll try to put together a PR for this case tonight.

ravenexp:
I still can't figure out why is it illegal to call `Py_InitializeEx(0)` when statically linking to the interpreter library.

But assuming it is, then `Py_SHARED` can only be disabled for the extension modules. But then it is completely useless on Windows because loading extensions requires the dynamically-linked Python. Then this PR does not make sense to begin with.

davidhewitt:
> I still can't figure out why is it illegal to call Py_InitializeEx(0) when statically linking to the interpreter library.

Apologies, I'm communicating poorly - it is fine to call `Py_InitializeEx(0)` with a statically embedded Python. But users often then run code which imports other extensions, and creates issues.

Give me a few days to try out some experiments and write some documentation in a new PR. Then I will come back to re-review this PR.

kngwyu (approved):
Thanks, I think this is OK, but let's wait a bit for David to come up with a good document.

build.rs
```diff
@@ -501,10 +501,19 @@ fn windows_hardcoded_cross_compile(
 
     let python_version = PythonVersion { major, minor };
 
+    let is_extension_module = env::var_os("CARGO_FEATURE_EXTENSION_MODULE").is_some();
+    let no_enable_shared = env::var_os("PYO3_CROSS_NO_ENABLE_SHARED").is_some();
```
jameshilliard, line 505: This variable probably needs to be added here to ensure cross compilation gets enabled when it is set:
https://github.com/PyO3/pyo3/blob/7eb3aa3fd276d6f94684f4a3aff4c080b9816954/build.rs#L145-L147

ravenexp: I can add it, but this variable is only used when cross-compiling for Windows, which requires `PYO3_CROSS_LIB_DIR` to be set as well.
The cross compilation mode is also automatically entered when the `--target` argument is different from the host triple, and `PYO3_CROSS` variable can be set to force it. As a user, I'd rather expect `PYO3_CROSS_LIB_DIR` and `PYO3_CROSS_PYTHON_VERSION` not do to anything when not cross-compiling.

davidhewitt:
Ok, so as per https://github.com/PyO3/pyo3/issues/416#issuecomment-814650710 I was able to prove that linking a static python _can_ work. It is however super awkard so not something we can support first-class in PyO3 at the moment.

Given that this PR affects windows, I'm going to try to repeat a similar experiment on Windows and see what makes sense to change / put into documentation.

ravenexp:
> Given that this PR affects windows, I'm going to try to repeat a similar experiment on Windows and see what makes sense to change / put into documentation.

AFAIK on Windows the extensions can't import symbols from the embedding executable because the Python DLL name gets hardcoded into the extension modules. One could theoretically build a fake `python3y.dll` that forwards all its symbols to the user executable  (that's how abi3 extensions are implemented on Windows: `python3.dll` forwards to `python3y.dll`), but I can't imagine anyone actually doing it. When you are compiling the interpreter statically, you have to compile in the Python standard library extensions statically as well. And loading third-party extensions at the run time is out.

build.rs
jameshilliard: I think for now probably a good idea to add it, mostly just because cargo has issues with cross compiling and forcing it might make debugging easier in some cases.

ravenexp: Added and rebased on `main`.

davidhewitt:
👍 that sounds about right based on the reading I've been doing too.

Thanks for this; I'm satisfied let's merge it and I'll add a follow-up PR shortly to unlock more pyo3 functionality when linking statically.

davidhewitt (approved):

build.rs
```diff
@@ -145,6 +145,7 @@ fn cross_compiling() -> Result<Option<CrossCompileConfig>> {
     if env::var_os("PYO3_CROSS").is_none()
         && env::var_os("PYO3_CROSS_LIB_DIR").is_none()
         && env::var_os("PYO3_CROSS_PYTHON_VERSION").is_none()
+        && env::var_os("PYO3_CROSS_NO_ENABLE_SHARED").is_none()
```
jameshilliard, line 148: I think this also needs to be added here:
https://github.com/PyO3/pyo3/blob/c0988f62d14a702abc64bd1a44f0e4dd0d03e9a8/build.rs#L911-L919

ravenexp: Applied

ravenexp:
Obsoleted by https://github.com/PyO3/pyo3/pull/1553

PR #1534, opened 2021-03-31
buffer: tidy up exceptions

davidhewitt:
Motivated by https://github.com/PyO3/pyo3/pull/1128#discussion_r605227220

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -17,13 +17,15 @@
 // DEALINGS IN THE SOFTWARE.
 
 //! `PyBuffer` implementation
-use crate::err::{self, PyResult};
-use crate::{exceptions, ffi, AsPyPointer, FromPyObject, PyAny, PyNativeType, Python};
-use std::ffi::CStr;
+use crate::{
+    err, exceptions::PyBufferError, ffi, AsPyPointer, FromPyObject, PyAny, PyNativeType, PyResult,
+    Python,
+};
 use std::marker::PhantomData;
 use std::os::raw;
 use std::pin::Pin;
 use std::{cell, mem, ptr, slice};
+use std::{ffi::CStr, fmt::Debug};
 
 /// Allows access to the underlying buffer used by a python object such as `bytes`, `bytearray` or `array.array`.
 // use Pin<Box> because Python expects that the Py_buffer struct has a stable memory address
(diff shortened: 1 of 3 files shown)


src/buffer.rs
```diff
@@ -169,25 +158,32 @@ impl<'source, T: Element> FromPyObject<'source> for PyBuffer<T> {
...
+
+        if buf.0.shape.is_null() {
+            Err(PyBufferError::new_err("shape is null but PyBUF_FULL_RO requested"))
+        } else if buf.0.strides.is_null() {
+            Err(PyBufferError::new_err("strides is nullbut PyBUF_FULL_RO requested"))
+        } else if mem::size_of::<T>() != buf.item_size()
+            || (buf.0.buf as usize) % mem::align_of::<T>() != 0
+            || !T::is_compatible_format(buf.format())
+        {
+            Err(PyBufferError::new_err("incompatible type as buffer"))
```
kngwyu, line 184: How about displaying the [typename](https://doc.rust-lang.org/std/any/fn.type_name.html)?

src/buffer.rs
```diff
@@ -169,25 +158,32 @@ impl<'source, T: Element> FromPyObject<'source> for PyBuffer<T> {
...
+            mem::transmute(buf)
+        };
+        // Create PyBuffer immediately so that if validation checks fail, the PyBuffer::drop code
+        // will call PyBuffer_Release (thus avoiding any leaks).
+        let buf = PyBuffer(Pin::from(buf), PhantomData);
+
+        if buf.0.shape.is_null() {
+            Err(PyBufferError::new_err("shape is null but PyBUF_FULL_RO requested"))
+        } else if buf.0.strides.is_null() {
+            Err(PyBufferError::new_err("strides is nullbut PyBUF_FULL_RO requested"))
```
kngwyu, line 179: Typo:
```suggestion
            Err(PyBufferError::new_err("strides is null but PyBUF_FULL_RO requested"))
```

Also, is mentioning `PyBuF_FULL_R0` ergonomic?

src/buffer.rs
```diff
@@ -169,25 +158,32 @@ impl<'source, T: Element> FromPyObject<'source> for PyBuffer<T> {
 impl<T: Element> PyBuffer<T> {
     /// Get the underlying buffer from the specified python object.
     pub fn get(obj: &PyAny) -> PyResult<PyBuffer<T>> {
-        unsafe {
-            let mut buf = Box::pin(ffi::Py_buffer::new());
+        // TODO: use nightly API Box::new_uninit() once stable
+        let mut buf = Box::new(mem::MaybeUninit::uninit());
```
kngwyu, line 162: If you take this approach, then you can remove `Py_buffer::new()`.
Though I'm a bit doubtful about how effective `MaybeUninit` is (I mean that this is effective only when we raise an error, but then `buf` is immediately dropped...).

kngwyu (approved):
Thanks, looks reasonable for me.

src/buffer.rs
davidhewitt: > Also, is mentioning PyBuF_FULL_R0 ergonomic?

🤷 I'll take it out

src/buffer.rs
davidhewitt: In the future I would like to remove `Py_buffer::new()`, but I think to do so right now would be inconvenient for users because [`PyBufferProtocol`](https://docs.rs/pyo3/0.13.2/pyo3/class/buffer/trait.PyBufferProtocol.html) basically forces you to handle the raw ffi pointer (which probably points to uninitialized memory which C created).

I chose `MaybeUninit` here because that's pretty much identical to what C would do: we create a slab of memory and pass its location to `PyObject_GetBuffer` to initialize the contents.

kngwyu:
Could you please some tests where construction fails?

PR #1538, opened 2021-04-02
pypy: support PyPy 3.7

davidhewitt:
The next PyPy version [will not include Python 3.6](https://doc.pypy.org/en/latest/release-v7.3.4.html), so we should get ready to upgrade.

Fingers crossed things *just work* 🤞

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
       fail-fast: false  # If one platform fails, allow the rest to keep testing.
       matrix:
         rust: [stable]
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10-dev, pypy-3.6]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -129,7 +132,7 @@ jobs:
       - name: Install python test dependencies
         run: |
           python -m pip install -U pip setuptools
-          pip install setuptools-rust pytest pytest-benchmark tox tox-venv
+          pip install setuptools-rust pytest pytest-benchmark tox
```
davidhewitt, line 135: I removed `tox-venv` because it caused issues with PyPy 3.7 on windows, and according to its repository [it's deprecated](https://github.com/tox-dev/tox-venv#tox-venv) anyway.

messense (approved):
🎉

PR #1539, opened 2021-04-02
macos: automatically provide required linker arguments

davidhewitt:
This PR removes the need for specifying macOS linker arguments in `.cargo/config` - we can feed them in directly in the build script.

I tested this on my wife's macbook air, and it seems to work. According to the cargo changelog, the `rustc-cdylib-link-arg` key has been supported since Rust 1.35: https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-135-2019-05-23

cc @messense I think you might have a macOS computer if you're interested in verifying whether this also works for you?

Because `setuptools_rust` and `maturin` set this config automatically you'll have to run plain `cargo build` inside a project which contains a pyo3 extension module to confirm (once removing any relevant config).


messense:
🎉 It works perfectly on my machine with `cargo build` for the word-count example.

messense (approved):

PR #1544, opened 2021-04-03
actions: cache rust builds

davidhewitt:
Hopefully this will add some much-needed performance improvements to our CI times!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,17 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: actions/cache@v2
+        # macOS seems to have a problem when the proc_macro_hack crates are cached.
+        # We can cache macOS in CI once we drop support for Rust older than 1.45
+        if: matrix.platform.os != 'macOS-latest'
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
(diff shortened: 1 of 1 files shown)


davidhewitt:
Hmm strange. It cuts the CI times by about 50%. But there seem to be random bugs on macOS, and also I had an issue the first time where some jobs randomly failed in the post-run cache job (I think because too much got uploaded to new cache all at the same time).

So I don't really know whether the speedup, while nice, is worth the additional complexity...

messense (approved):

kngwyu:
I don't think we need to merge this ASAP. Optimistically, one of us will find the workaround eventually.

davidhewitt:
> I don't think we need to merge this ASAP. Optimistically, one of us will find the workaround eventually.

Even though we can't yet enable for macOS, I think there's value in merging this now. Github actions seems to have a limit to the number of jobs running across all PRs in parallel, so when several PRs are pushed close together it can take some time for all the CI to run.

If we cache the windows and linux jobs, we'll get through all of these jobs faster, so we get feedback on CI faster.

kngwyu (approved):
OK, I didn't check that you added `if: matrix.platform.os != 'macOS-latest'`

messense:
Looks like you can workaround the macOS cache issue by using `gnu-tar`, see https://github.com/PyO3/maturin/pull/493

PR #1545, opened 2021-04-04
coverage: use source-based coverage

davidhewitt:
The [source-based coverage](https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/source-based-code-coverage.html) in rust nightly is meant to be better than the "current" `-Zprofile` option, so let's try it here!

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -557,12 +557,6 @@ mod tests {
     #[test]
     #[should_panic(expected = "new panic")]
     fn fetching_panic_exception_resumes_unwind() {
-        // TODO replace with #[cfg(panic = "unwind")] once stable
-        if !crate::cfg_panic_unwind() {
-            // panic to meet the expected abort in panic=abort :-/
-            panic!("new panic");
-        }
-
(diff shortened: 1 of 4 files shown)


davidhewitt:
Well, looks like it'll make a tiny positive improvement to our coverage numbers! 😄

davidhewitt:
In addition I guess it means we don't need to support running our test suite with `panic = "abort"` any more, so I pushed a separate commit to remove that...

src/err/mod.rs
```diff
@@ -557,12 +557,6 @@ mod tests {
     #[test]
     #[should_panic(expected = "new panic")]
     fn fetching_panic_exception_resumes_unwind() {
-        // TODO replace with #[cfg(panic = "unwind")] once stable
-        if !crate::cfg_panic_unwind() {
-            // panic to meet the expected abort in panic=abort :-/
-            panic!("new panic");
-        }
-
```
kngwyu, line 565: Well I'm quite happy to see this is removed

kngwyu (approved):
Thanks!

PR #1546, opened 2021-04-05
Add SequenceProtocol and MappingProtocol descriptions to the guide

ravenexp:
References:

1. https://docs.python.org/3/reference/datamodel.html#emulating-container-types
1. https://docs.python.org/3/c-api/sequence.html
1. https://docs.python.org/3/c-api/typeobj.html#c.PySequenceMethods
1. https://docs.python.org/3/c-api/mapping.html
1. https://docs.python.org/3/c-api/typeobj.html#c.PyMappingMethods

### DIFF
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -61,7 +61,7 @@ Each method corresponds to Python's `self.attr`, `self.attr = value` and `del se
 
 ### Emulating numeric types
 
-The [`PyNumberProtocol`] trait allows [emulate numeric types](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).
+The [`PyNumberProtocol`] trait allows to emulate [numeric types](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).
 
   * `fn __add__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
   * `fn __sub__(lhs: impl FromPyObject, rhs: impl FromPyObject) -> PyResult<impl ToPyObject>`
@@ -105,12 +105,6 @@ The reflected operations are also available:
 
 The code generated for these methods expect that all arguments match the
 signature, or raise a TypeError.
-
-*Note*: Currently implementing the method for a binary arithmetic operations
-(e.g, `__add__`) shadows the reflected operation (e.g, `__radd__`).  This is
-being addressed in [#844](https://github.com/PyO3/pyo3/issues/844).  to make
-these methods
-
 
 This trait also has support the augmented arithmetic assignments (`+=`, `-=`,
 `*=`, `@=`, `/=`, `//=`, `%=`, `**=`, `<<=`, `>>=`, `&=`, `^=`, `|=`):
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -16,9 +16,9 @@ The [`PyObjectProtocol`] trait provides several basic customizations.
 
 To customize object attribute access, define the following methods:
 
-  * `fn __getattr__(&self, name: FromPyObject) -> PyResult<impl IntoPy<PyObject>>`
-  * `fn __setattr__(&mut self, name: FromPyObject, value: FromPyObject) -> PyResult<()>`
-  * `fn __delattr__(&mut self, name: FromPyObject) -> PyResult<()>`
+  * `fn __getattr__(&self, name: impl FromPyObject) -> PyResult<impl IntoPy<PyObject>>`
+  * `fn __setattr__(&mut self, name: impl FromPyObject, value: impl FromPyObject) -> PyResult<()>`
+  * `fn __delattr__(&mut self, name: impl FromPyObject) -> PyResult<()>`
 
 Each method corresponds to Python's `self.attr`, `self.attr = value` and `del self.attr` code.
 
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -204,6 +204,45 @@ where _N_ is the length of the sequence.
     Used by the `*=` operator, after trying the numeric in place multiplication via
     the `PyNumberProtocol` trait method.
 
+### Emulating mapping containers (such as dictionaries)
+
+The [`PyMappingProtocol`] trait allows to emulate
+[mapping container types](https://docs.python.org/3/reference/datamodel.html#emulating-container-types).
+
+For a mapping, the keys may be Python objects of arbitrary type.
+
+  * `fn __len__(&self) -> PyResult<usize>`
+
+    Implements the built-in function `len()` for the mapping.
+
+  * `fn __getitem__(&self, key: impl FromPyObject) -> PyResult<impl ToPyObject>`
+
+    Implements evaluation of the `self[key]` element.
+    If `key` is of an inappropriate type, `TypeError` may be raised;
+    if `key` is missing (not in the container), `KeyError` should be raised.
+
+  * `fn __setitem__(&mut self, key: impl FromPyObject, value: impl FromPyObject) -> PyResult<()>`
+
+    Implements assignment to the `self[key]` element or insertion of a new `key`
+    mapping to `value`.
+    Should only be implemented if the mapping support changes to the values for keys,
+    or if new keys can be added.
+    The same exceptions should be raised for improper key values as
+    for the `__getitem__()` method.
+
+  * `fn __delitem__(&mut self, key: impl FromPyObject) -> PyResult<()>`
+
+    Implements deletion of the `self[key]` element.
+    Should only be implemented if the mapping supports removal of keys.
+    The same exceptions should be raised for improper key values as
+    for the `__getitem__()` method.
+
+  * `fn __reversed__(&self) -> PyResult<impl ToPyObject>`
+
+    Called (if present) by the `reversed()` built-in to implement reverse iteration.
+    It should return a new iterator object that iterates over all the objects in
+    the container in reverse order.
+
 ### Garbage Collector Integration
 
 If your type owns references to other Python objects, you will need to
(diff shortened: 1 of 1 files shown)


guide/src/class/protocols.md
```diff
@@ -61,7 +61,7 @@ Each method corresponds to Python's `self.attr`, `self.attr = value` and `del se
 
 ### Emulating numeric types
 
-The [`PyNumberProtocol`] trait allows [emulate numeric types](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).
+The [`PyNumberProtocol`] trait allows to emulate [numeric types](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).
```
davidhewitt, line 64:
```suggestion
The [`PyNumberProtocol`] trait can be implemented to emulate [numeric types](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types).
```

guide/src/class/protocols.md
```diff
@@ -147,6 +141,108 @@ Other:
   * `fn __index__(&'p self) -> PyResult<impl ToPyObject>`
   * `fn __round__(&'p self, ndigits: Option<impl FromPyObject>) -> PyResult<impl ToPyObject>`
 
+### Emulating sequential containers (such as lists or tuples)
+
+The [`PySequenceProtocol`] trait allows to emulate
```
davidhewitt, line 146:
```suggestion
The [`PySequenceProtocol`] trait can be implemented to emulate
```

guide/src/class/protocols.md
```diff
@@ -147,6 +141,108 @@ Other:
...
+[sequential container types](https://docs.python.org/3/reference/datamodel.html#emulating-container-types).
+
+For a sequence, the allowable keys should be the integers _k_ for which _0 <= k < N_,
```
davidhewitt, line 149:
```suggestion
For a sequence, the keys are the integers _k_ for which _0 <= k < N_,
```

guide/src/class/protocols.md
```diff
@@ -147,6 +141,108 @@ Other:
...
+
+  * `fn __getitem__(&self, idx: isize) -> PyResult<impl ToPyObject>`
+
+    Implements evaluation of the `self[idx]` element.
+    If the `idx` value is outside the set of indexes for the sequence, `IndexError` should be raised.
+
+    *Note:* Negative integer indexes are handled as follows: if `__len__()` is defined,
+    it is called and the sequence length is used to compute a positive index,
+    which is passed to `__getitem__()`.
+    If `__len__()` is not defined, the index is passed as is to the function.
```
davidhewitt, line 164: This differs from https://docs.python.org/3/reference/datamodel.html#object.__getitem__, which says that negative integers are are passed directly to `__getitem__` which then can interpret the negative numbers as it likes?

guide/src/class/protocols.md
```diff
@@ -147,6 +141,108 @@ Other:
...
+    Implements deletion of the `self[key]` element.
+    Should only be implemented if the mapping supports removal of keys.
+    The same exceptions should be raised for improper key values as
+    for the `__getitem__()` method.
+
+  * `fn __reversed__(&self) -> PyResult<impl ToPyObject>`
+
+    Called (if present) by the `reversed()` built-in to implement reverse iteration.
+    It should return a new iterator object that iterates over all the objects in
+    the container in reverse order.
```
davidhewitt, line 244: Hmm we should probably move `__reversed__` to the `PyIterProtocol` trait? (I can do it in a separate PR.)

davidhewitt (commented):
Thank you for giving some much needed love to our documentation!

I have a couple of suggestions and questions; comments below.

Also, a more general question I have wondered about: the `PyResult` bit of all the return values is optional - e.g. for a proto which returns `PyResult<T>`, just `T` is also an acceptable return value. Do you think it's better if:
- we include `PyResult<T>` on all these return values, with a note in an `# Error Handling` section saying *PyResult is optional*
- we just have `T` on all the return value, with a note in an `# Error Handling` section saying *all return types can be wrapped in PyResult*

I think that just having `T` on the return value is easier to read, but it's maybe less obvious for new users to then figure out how to return errors unless they read the whole document and find the `Error Handling` section!

guide/src/class/protocols.md
ravenexp: I presume PyO3 plugs the `__getitem__` method implementation into the [`PySequenceMethods.sq_item`](https://docs.python.org/3/c-api/typeobj.html#c.PySequenceMethods) slot. The C function in that slot has a slightly different interface from the Python-native `__getitem__()` method, as described in the C-API docs linked above. `PyObject_GetItem()` and `PySequence_GetItem()` handle the negative indexes themselves when `__len__()` is also implemented.

My implementations of `__getitem__()` always start with `assert!(idx >= 0);` because I also implement `__len__()`, and they indeed work with the negative indexes.

guide/src/class/protocols.md
ravenexp: Well, `__reversed__()` is a standalone method, not a [`PyMappingMethods`](https://docs.python.org/3/c-api/typeobj.html#c.PyMappingMethods) slot, so there is no particular reason it should be a part of `PyMappingProtocol`.
However, implementing `__reversed__()` for sequences is not very useful, because [the default implementation](https://docs.python.org/3/reference/datamodel.html#object.__reversed__) seems sufficient for most cases.

I personally do not like the idea of adding anything to `PyIterProtocol` though. It's already semantically overloaded as it encompasses both `IntoIterator` and `Iterator` Rust traits. But then the user is expected to implement a different subset of methods for each of the intended uses (as an iterator or as an iterable).

Currently, we also have to write mandatory boilerplate code like
```rust
fn __iter__(slf: PyRef<Self>) -> PyRef<Self> {
    slf
}
```
for every custom iterator class.

Most Rust programmers would expect a blanket implementation like
```rust
impl<I> IntoIterator for I where
    I: Iterator, 
```
from the stdlib to exist instead.

I'm sorry about the long rant, but `PyIterProtocol` was the biggest road bump for me when working with PyO3.
I really wish it was two distinct traits, like for example `PyIterProtocol` and `PyIntoIterProtocol`, so that anyone who knows how Rust iterators work could easily implement Python iterators without scratching their head about "why do I need to return `PyRef<Self>` instead `Self` here?"

ravenexp:
> Also, a more general question I have wondered about: the `PyResult` bit of all the return values is optional - e.g. for a proto which returns `PyResult<T>`, just `T` is also an acceptable return value. Do you think it's better if:
> 
>     * we include `PyResult<T>` on all these return values, with a note in an `# Error Handling` section saying _PyResult is optional_
> 
>     * we just have `T` on all the return value, with a note in an `# Error Handling` section saying _all return types can be wrapped in PyResult_
> 
> 
> I think that just having `T` on the return value is easier to read, but it's maybe less obvious for new users to then figure out how to return errors unless they read the whole document and find the `Error Handling` section!

Hmm, either way is fine with me. I mostly followed the convention `PyObjectProtocol` and `PyNumberProtocol` docs already used.
It is also not very clear whether `PyResult<()>` return type can be removed entirely.
New Rust programmers would probably expect the trait method signatures to match exactly, and the fact that one can implement either `fn foo(&self)` or `fn foo(&self) -> PyResult<()>` is not obvious at all.

guide/src/class/protocols.md
davidhewitt: Thanks, I completely agree that the current situation is confusing in many ways. I wonder if alternatively here I should remove `__reversed__` from `#[pyproto]` completely. In my opinion `#[pymethods]` are very easy to learn, and implementing `__reversed__` in `#[pymethods]` will already work.

guide/src/class/protocols.md
davidhewitt: Ah yes, this is a really good point. Perhaps in the documentation here we should link all these methods to their relevant slot information?

All the relationships between these methods and their slots is defined in https://github.com/PyO3/pyo3/blob/main/pyo3-macros-backend/src/defs.rs

davidhewitt:
> New Rust programmers would probably expect the trait method signatures to match exactly

This is very true. I think in that case it's better to leave the documentation with `PyResult<_>` on all the methods.

> the fact that one can implement either fn foo(&self) or fn foo(&self) -> PyResult<()> is not obvious at all.

I agree. There's a lot of magic going on in these trait implementations, which I really wish we could simplify.

davidhewitt (approved):
Based on the discussion here, I'm happy that this is good to be merged to the guide as-is. Thanks very much for this; the `#[pyproto]` documentation is desperately in need of love!

guide/src/class/protocols.md
ravenexp: > Perhaps in the documentation here we should link all these methods to their relevant slot information?

I like this idea, but the mapping between methods and slots looks not so straightforward. There are methods that do not have a C function slot like `__bytes__()`, `__format__()` and `__reversed__()`. And then there are methods that share the same slot like `__setitem__()` and `__delitem__()`. I think we could add a symbolic tag like `PySequenceMethods.sq_item` to each trait method.

Another interesting issue here is the implicit slot method probing order: for example the evaluation of `obj[i]` Python expression probes `PyMappingMethods.mp_subscript` before `PySequenceMethods.sq_item`. This is relevant when someone implements both of these traits for some reason.

guide/src/class/protocols.md
ravenexp: I like this idea very much. There are other non-slotted methods that are included in the protocol traits, like `__bytes__()`. And then there are C function slots that have no corresponding `PyObjectProtocol` method like [`tp_call`](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_call) aka [`__call__()`](https://docs.python.org/3/reference/datamodel.html#object.__call__).

I'm all for implementing true non-slotted methods as `#[pymethods]`. This way we can avoid surprises like https://github.com/PyO3/pyo3/issues/1465 in the future.

PR #1552, opened 2021-04-09
Rerun if PYO3_CROSS env variable changes.

jameshilliard:
Looks like I missed this in #1514.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -902,6 +902,7 @@ fn main_impl() -> Result<()> {
         "LIB",
         "LD_LIBRARY_PATH",
         "PYO3_PYTHON",
+        "PYO3_CROSS",
         "PYO3_CROSS_LIB_DIR",
         "PYO3_CROSS_PYTHON_VERSION",
     ] {


messense (approved):
Thanks!

davidhewitt:
I must have missed this from #1514 - what's the purpose of the `PYO3_CROSS` env var? It's not documented; it looks like #1514 introduced this, but it doesn't do anything special except force cross compilation. However at least `PYO3_CROSS_LIB_DIR` is also required at the same time, so it doesn't seem that there's a point to having a separate `PYO3_CROSS` variable...

jameshilliard:
> I must have missed this from #1514 - what's the purpose of the `PYO3_CROSS` env var?

Oh, I think I was using that so that I could ensure pyo3 was running in cross compilation mode for cases where `PYO3_CROSS_LIB_DIR` is getting set in a different part of the build system. There's basically a chain of build systems when I cross compile, buildroot being the meta build system which invokes the setuptools build system which then invokes pyo3 using the cargo build system. So I had added that so at any point cross compilation can be turned on from any build system essentially to ensure one gets a useful error message in the event something in the build chain doesn't set PYO3_CROSS_LIB_DIR. Also added it so that we have a way to force cross compilation without affecting anything else in the event `PYO3_CROSS_LIB_DIR` gets removed/changed.

PR #1553, opened 2021-04-11
auto-initialize: better error messages and embedding docs

davidhewitt:
This commit removes the restriction that `prepare_freethreaded_python` cannot be called when linking statically.

I keep the error when `auto-initialize` is enabled with static linking, because it's usually an accidental bug which leads to confusing linker errors at runtime (e.g. #762). However I move it to `build.rs` and improve the error message contents a lot.

To improve the error message context further, I've added a section discussing dynamic and static embedding to the guide.

cc @kngwyu @ravenexp - follow-up from https://github.com/PyO3/pyo3/pull/1532#issuecomment-810607813

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -17,10 +17,15 @@ const CFG_KEY: &str = "py_sys_config";
 
 type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
-// A simple macro for returning an error. Resembles failure::bail and anyhow::bail.
+// A simple macro for returning an error. Resembles anyhow::bail.
 macro_rules! bail {
     ($msg: expr) => { return Err($msg.into()); };
-    ($fmt: literal $(, $args: expr)+) => { return Err(format!($fmt $(,$args)+).into()); };
+    ($fmt: literal $($args: tt)+) => { return Err(format!($fmt $($args)+).into()); };
+}
+
+// A simple macro for checking a condition. Resembles anyhow::ensure.
+macro_rules! ensure {
+    ($condition:expr, $($args: tt)+) => { if !($condition) { bail!($($args)+) } };
 }
 
 // Show warning. If needed, please extend this macro to support arguments.
@@ -757,8 +762,30 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<()> {
         _ => {}
     }
 
-    if interpreter_config.shared {
-        println!("cargo:rustc-cfg=Py_SHARED");
+    if env::var_os("CARGO_FEATURE_AUTO_INITIALIZE").is_some() {
+        ensure!(
+            interpreter_config.shared,
+            "The `auto-initialize` feature is enabled, but your python installation only supports \
+            embedding the Python interpreter statically. If you are attempting to run tests, or a \
+            binary which is okay to link dynamically, install a Python distribution which ships \
+            with the Python shared library.\n\
+            \n\
+            Embedding the Python interpreter statically does not yet have first-class support in \
+            PyO3. If you are sure you intend to do this, disable the `auto-initialize` feature.\n\
+            \n\
+            For more information, see \
+            https://pyo3.rs/v{pyo3_version}/\
+                building_and_distribution.html#embedding-python-in-rust",
+            pyo3_version = env::var("CARGO_PKG_VERSION").unwrap()
+        );
+
+        // TODO: PYO3_CI env is a hack to workaround CI with PyPy, where the `dev-dependencies`
+        // currently cause `auto-initialize` to be enabled in CI.
+        // Once cargo's `resolver = "2"` is stable (~ MSRV Rust 1.52), remove this.
+        ensure!(
+            !interpreter_config.is_pypy() || env::var_os("PYO3_CI").is_some(),
+            "The `auto-initialize` feature is not supported with PyPy."
+        );
     }
 
     let is_abi3 = is_abi3();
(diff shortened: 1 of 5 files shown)


ravenexp:
This solution looks good to me. I do have a question however: do we really need `cfg(Py_SHARED)` anymore?

https://github.com/PyO3/pyo3/pull/1532 can be dropped if `Py_SHARED` config is not required to restrict access to certain libpython APIs.

davidhewitt:
#1532 still has a use if anyone wants to force static-linking in the windows cross-compile, so I'm fine with it to stay?

ravenexp:
But defining `Py_ENABLE_SHARED` does not force static linking all by itself. On the PyO3 side, it simply enables the `Py_SHARED` config gate, and on the libpython side, it mostly deals with  `__declspec(dllexport)` and `__declspec(dllimport)` attributes in the C headers.

I'm just wondering why would anyone want to disable `Py_SHARED` for PyO3 manually, when it accomplishes almost nothing.

davidhewitt:
Ok. Let's close #1532 in that case?

ravenexp:
I closed #1532 , but I'll also look into whether `cfg(Py_SHARED)` can be safely removed as well.

davidhewitt:
> whether cfg(Py_SHARED) can be safely removed as well.

No need to look, this PR removes it!

build.rs
```diff
@@ -17,10 +17,15 @@ const CFG_KEY: &str = "py_sys_config";
 
 type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
-// A simple macro for returning an error. Resembles failure::bail and anyhow::bail.
```
kngwyu, line 20: Well, time flies...

build.rs
```diff
@@ -757,8 +762,36 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<()> {
         _ => {}
     }
 
-    if interpreter_config.shared {
-        println!("cargo:rustc-cfg=Py_SHARED");
+    if env::var_os("CARGO_FEATURE_AUTO_INITIALIZE").is_some() {
+        ensure!(
+            interpreter_config.shared,
+            "The `auto-initialize` feature is enabled, however your python installation only \
```
kngwyu, line 768: This document is nice, but I'm a bit doubtful that users want to read such a long message (this means that I'm embarrassingly lazy to read errors:sweat:).

build.rs
```diff
@@ -757,8 +762,36 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<()> {
...
+            option. For example, this is the default for `pyenv` on macOS.\n\
+            \n\
+            Embedding the Python interpreter statically has many complications and does not yet \
+            have first-class support in PyO3. If you are sure you intend to do this, disable the \
+            `auto-initialize` feature and call `pyo3::prepare_freethreaded_python` directly in \
+            your program.\n\
+            \n\
+            If you are attempting to run tests or a binary which is okay to link dynamically, \
+            install a Python distribution which ships with the Python shared library and direct \
+            PyO3 to use it using a virtual environment or the PYO3_PYTHON environment variable.\n\
```
kngwyu, line 780: Let's place this `if` sentence first. This is a major case, I guess. Also, I think just noting to use a `--enabled-shared` Python is sufficient. I don't see any reason to refer to virtualenv (can't we also use pyenv?).

build.rs
```diff
@@ -757,8 +762,36 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<()> {
...
+            interpreter_config.shared,
+            "The `auto-initialize` feature is enabled, however your python installation only \
+            supports embedding the Python interpreter statically. This is most often seen in \
+            Python distributions compiled from source without the `--enable-shared` configuration \
+            option. For example, this is the default for `pyenv` on macOS.\n\
+            \n\
+            Embedding the Python interpreter statically has many complications and does not yet \
+            have first-class support in PyO3. If you are sure you intend to do this, disable the \
+            `auto-initialize` feature and call `pyo3::prepare_freethreaded_python` directly in \
+            your program.\n\
```
kngwyu, line 776: I feel this a bit redundant. Users just need to know it's a rare usage and not supported very well, like: `Or, if you are sure to embed the Python interpreter statically (it's a rare usage and not supported very well, though), please ...`

guide/src/building_and_distribution.md
```diff
@@ -119,10 +119,53 @@ cargo build --target x86_64-pc-windows-gnu
 
 Any of the `abi3-py3*` features can be enabled instead of setting `PYO3_CROSS_PYTHON_VERSION` in the above examples.
 
+## Embedding Python in Rust
+
+If you want to embed the Python interpreter inside a Rust program, there are two modes in which this can be done: dynamically and statically. We'll cover each of these modes in the following sections. Each of them affect how you must distribute your program. Instead of learning how to do this yourself, you might want to consider using a project like [PyOxidizer] to ship your application and all of its dependencies in a single file.
```
kngwyu, line 124: > you might want to consider using a project like [PyOxidizer] to ship your application and all of its dependencies in a single file.

Nice to know that PyOxidzer has such a feature 👍🏼

guide/src/building_and_distribution.md
```diff
@@ -119,10 +119,53 @@ cargo build --target x86_64-pc-windows-gnu
...
+
+PyO3 automatically switches between the two linking modes depending on whether the Python distribution you have configured PyO3 to use ([see above](#python-version)) contains a shared library or a static library.
+
+### Dynamically embedding the Python interpreter
```
kngwyu, line 128: Nicely organized 👍🏼

kngwyu (changes requested):
Thank you!
The document is nice, but I prefer a shorter error message for displaying in a terminal emulator.

build.rs
davidhewitt: Thanks, this and the comments below were very helpful. I've pushed a new commit which makes the build error much shorter.

ravenexp:
> > whether cfg(Py_SHARED) can be safely removed as well.
> 
> No need to look, this PR removes it!

Oh, indeed. There's one place left though:

https://github.com/PyO3/pyo3/blob/83c6ec4f8ae5d597e09bd71b434a95e9d81f94df/build.rs#L887

davidhewitt:
> Oh, indeed. There's one place left though:

👍 nice catch, removed!

kngwyu (approved):

kngwyu:
Thank you! The new error message is quite easy to read.

PR #1557, opened 2021-04-11
build: avoid rebuilds when using clippy in a virtualenv

### LINKED ISSUE #1551: Switching between `cargo check` and `cargo clippy` forces pyo3 to rebuild
Hi, I observe that switching between `cargo clippy` and `cargo check` forces pyo3 and all downstream crates to rebuild.

### 💥 Reproducing

1. `cargo new --lib example && cd example`
2. Edit `Cargo.toml` to include `pyo3 = { version = "0.13", features = ["extension-module"] }`.
3. Run `cargo check`.
4. Run `cargo check` again, observe no rebuild.
5. Run `cargo clippy`.
6. Run `cargo check` again, observe `pyo3` and all downstream crates rebuild.

### 🌍 Environment

 - Your operating system and version: Arch Linux
 - Your python version: 3.9
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: pacman, no virtualenv
 - Your Rust version (`rustc --version`): rustc 1.51.0 (2fd73fabe 2021-03-23)
 - Your PyO3 version: 0.13.2

davidhewitt:
Closes #1551 

The fix comes in two pieces:
 - I add detection for `VIRTUAL_ENV` and `CONDA_PREFIX` variables as possible ways to detect the Python interpreter; this logic is borrowed from `maturin`.
 - I refactored the environment variable `rerun-if-env-changed` system to ensure we emit _exactly_ the environment variables checked by the current run of the build script. This should avoid issues like #1552 .

~~cc @jameshilliard I took out the (new, undocumented) `PYO3_CROSS` variable - was there a specific need for its creation?~~

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,6 +2,7 @@ use std::{
     collections::{HashMap, HashSet},
     convert::AsRef,
     env,
+    ffi::OsString,
     fs::{self, DirEntry},
     io,
     path::{Path, PathBuf},
@@ -33,6 +34,20 @@ macro_rules! warn {
     ($msg: literal) => {
         println!(concat!("cargo:warning=", $msg));
     };
+}
+
+/// Gets an environment variable owned by cargo.
+///
+/// Environment variables set by cargo are expected to be valid UTF8.
+fn cargo_env_var(var: &str) -> Option<String> {
+    env::var_os(var).map(|os_string| os_string.to_str().unwrap().into())
+}
+
+/// Gets an external environment variable, and registers the build script to rerun if
+/// the variable changes.
+fn env_var(var: &str) -> Option<OsString> {
+    println!("cargo:rerun-if-env-changed={}", var);
+    env::var_os(var)
 }
 
 /// Information returned from python interpreter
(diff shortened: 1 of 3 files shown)


davidhewitt:
(Ah sorry just seen your other comment, I'll restore `PYO3_CROSS` and add documentation.)

build.rs
```diff
@@ -145,7 +145,10 @@ fn cross_compiling() -> Result<Option<CrossCompileConfig>> {
     let target_vendor = cargo_env_var("CARGO_CFG_TARGET_VENDOR");
     let target_os = cargo_env_var("CARGO_CFG_TARGET_OS");
 
-    if cross_lib_dir.is_none() && cross_python_version.is_none() {
+    if env_var("PYO3_CROSS").is_none()
+        && cross_python_version.is_none()
```
jameshilliard, line 149: I think this should be `cross_lib_dir.is_none()`.

davidhewitt: 👍 good catch, thanks!

davidhewitt:
If nobody has any objections, I'll probably merge this later today. I think it's a useful tidy up that can avoid some user frustration.

PR #1558, opened 2021-04-11
ci: fix macOS cache (try 2)

davidhewitt:
🤦


.github/workflows/ci.yml
```diff
@@ -69,10 +69,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # macOS: install gnu-tar because BSD tar is buggy for github actions
+      # macos: install gnu-tar because BSD tar is buggy for github actions
```
kngwyu, line 72: I guess you didn't fix this? Anyway, thank you for lots of tries.

davidhewitt: I think it should be fixed now. (I just got the `if` wrong in #1555.)

Issue #1205, opened 2020-09-22
Trouble using PyContextProtocol


jothan:
I'm trying to make a file writing interface be its own context manager with PyO3 0.12. The trouble is that I don't know how to return self from the ``__enter__`` function. I don't have this issue for implementing iterators that are their own iterators.

I can work around this by not using the context protocol and simply using methods inside the pymethods block.

Code that I'm trying to make work: https://github.com/jothan/cordoba/blob/cf8c6deb809d73948c8c47677c6b92b799588131/src/pymod.rs#L200

Iterator that can return self: https://github.com/jothan/cordoba/blob/cf8c6deb809d73948c8c47677c6b92b799588131/src/pymod.rs#L118

davidhewitt:
Thanks, yes this is sadly a flaw with the `#[pyproto]` design at the moment which I've been aware of for a while but not found the right moment to fix. Perhaps this is now a good time. I've opened #1206 and will try to take a spin at this soon.

kngwyu:
For now, I'm happy to write a temporary solution (i.e. allow `PyRefMut`) for this.
@davidhewitt What do you think?

davidhewitt:
I think what I have in mind is non-breaking and not _too hard_ so I can do over the weekend and we can potentially even release in a 0.12.2.

indygreg:
Fun fact: `PyContextProtocol` isn't required to implement `__enter__` and `__exit__`. Since these methods don't occupy any special slots on `PyTypeObject` (from Python's C API), you can implement `__enter__` and `__exit__` next to your normal methods on a `#[pymethods] impl` block.

You can return `self` by doing something like:

```rust
#[pymethods]
impl MyClass {
    fn __enter__<'p>(slf: PyRef<'p, Self>, _py: Python<'p>) -> PyResult<PyRef<'p, Self>> {
        Ok(slf)
    }
}
```

You can also use `PyRefMut` if you need a mutable instance.

I am curious why `PyContextProtocol` exists in the first place. I naively assumed the various traits existed to facilitate conversion to the appropriate `PyTypeObject` slots. But `__enter__` and `__exit__` are defined on `tp_methods` and don't require special slot handling. So maybe there's another reason for their existence?

kngwyu:
@indygreg 
> So maybe there's another reason for their existence?

No, there's not some complex reason. It's just a convention for users: i.e., we treat `__dunder__` methods as protocol, even when they are not actually so in CPython API.

[...]

PR #1560, opened 2021-04-12
pyproto: deprecate py_methods

davidhewitt:
I've been thinking about `#[pyproto]` ergonomics for a while, prompted by questions like #1205, #1206, https://github.com/PyO3/pyo3/pull/1546#discussion_r608365767, and confusion I often see by users on Gitter.

There are a few methods like `__bytes__` which have no technical reason to be in `#[pyproto]`, and in fact can already be implemented in `#[pymethods]`. (See https://github.com/PyO3/pyo3/issues/1205#issuecomment-778529199.) I think that `#[pymethods]` is much easier to understand and use than `#[pyproto]`, so to make "one way" for users I propose in this PR to deprecate these `#[pyproto]` methods.

### DIFF
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -1,12 +1,14 @@
 ## Class customizations
 
-Python's object model defines several protocols for different object behavior, like sequence,
-mapping or number protocols. PyO3 defines separate traits for each of them. To provide specific
-Python object behavior, you need to implement the specific trait for your struct. Important note,
-each protocol implementation block has to be annotated with the `#[pyproto]` attribute.
+PyO3 uses the `#[pyproto]` attribute in combination with special traits to implement certain protocol (aka `__dunder__`) methods of Python classes. The special traits are listed in this chapter of the guide. See also the [documentation for the `pyo3::class` module]({{#PYO3_DOCS_URL}}/pyo3/class/index.html).
 
-All `#[pyproto]` methods which can be defined below can return `T` instead of `PyResult<T>` if the
-method implementation is infallible. In addition, if the return type is `()`, it can be omitted altogether.
+Python's object model defines several protocols for different object behavior, such as the sequence, mapping, and number protocols. You may be familiar with implementing these protocols in Python classes by "dunder" methods, such as `__str__` or `__repr__`.
+
+In the Python C-API which PyO3 is dependent upon, many of these protocol methods have to be provided into special "slots" on the class type object. To fill these slots PyO3 uses the `#[pyproto]` attribute in combination with special traits.
+
+All `#[pyproto]` methods can return `T` instead of `PyResult<T>` if the method implementation is infallible. In addition, if the return type is `()`, it can be omitted altogether.
+
+There are many "dunder" methods which are not included in any of PyO3's protocol traits, such as `__dir__`. These methods can be implemented in `#[pymethods]` as already covered in the previous section.
 
 ### Basic object customization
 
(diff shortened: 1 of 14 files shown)


guide/src/migration.md
```diff
@@ -15,6 +15,47 @@ For projects embedding Python in Rust, PyO3 no longer automatically initalizes a
 
 The limitation of the new default implementation is that it cannot support multiple `#[pymethods]` blocks for the same `#[pyclass]`. If you need this functionality, you must enable the `multiple-pymethods` feature which will switch `#[pymethods]` to the inventory-based implementation.
 
+### Deprecated `#[pyproto]` methods
+
+Some methods such as `__bytes__` and `__format__` have been possible to implement two ways in PyO3 for some time: via a `#[pyproto]` (e.g. `PyBasicProtocol` for the methods listed here), or by writing them directly in `#[pymethods]`. This is only true for a handful of the `#[pyproto]` methods (for technical reasons to do with the Python C-API).
```
kngwyu, line 20: Some protocol/dunder methods? Also, if we mention ` (for technical reasons to do with the Python C-API).`, we should also note that the situation can change in the future. `Technical reasons` can look like something unfixable for me.

guide/src/migration.md
```diff
@@ -15,6 +15,47 @@ For projects embedding Python in Rust, PyO3 no longer automatically initalizes a
...
+
+So as to have only one way to do things, and because the `#[pymethods]` macro has better documentation, error messages, and ergonomics, the `#[pyproto]` forms of these methods have been deprecated.
```
kngwyu, line 22: The second reason seems redundant for me. If we have two similar codes, often the one grows better than the other.
```suggestion
So as to have only one way to do things, the `#[pyproto]` forms of these methods have been deprecated.
```

guide/src/class/protocols.md
```diff
@@ -1,12 +1,14 @@
...
-mapping or number protocols. PyO3 defines separate traits for each of them. To provide specific
-Python object behavior, you need to implement the specific trait for your struct. Important note,
-each protocol implementation block has to be annotated with the `#[pyproto]` attribute.
+Python's object model defines several protocols for different object behavior, like sequence, mapping or number protocols. You may be familiar with implementing these protocol in Python classes by "dunder" methods, such as `__str__` or `__repr__`.
 
-All `#[pyproto]` methods which can be defined below can return `T` instead of `PyResult<T>` if the
-method implementation is infallible. In addition, if the return type is `()`, it can be omitted altogether.
+In the Python C-API which PyO3 is dependent upon, many of these protocol methods have to be provided into special "slots" on the class type object. To fill these slots PyO3 uses the `#[pyproto]` attribute in combination with special traits.
+
+The special traits are listed in this chapter of the guide. See also the [documentation for the `pyo3::class` module]({{#PYO3_DOCS_URL}}/pyo3/class/index.html).
```
kngwyu, line 7: Here, we explain Python's type slots first and then PyO3's pyproto. I propose to reverse the order and note that we have traits and pyproto for some special dunder methods first. We can add the reasons later.

kngwyu (approved):
Thanks, good intermediate changes if we are going to refactor protocols.

guide/src/migration.md
davidhewitt: Good point. I changed to `for technical reasons to do with the way PyO3 currently interacts with the Python C-API`, which I think captures the fact that this might change.

davidhewitt:
Thanks @kngwyu for the feedback. I've made some changes to the documentation based on your suggestions, it looks much better now!

ravenexp:
I'd like to bring to your attention that the `__bool__()` method is currently included in the `PyObjectProtocol` trait, but the corresponding C API slot [`nb_bool` ](https://docs.python.org/3/c-api/typeobj.html#c.PyNumberMethods.nb_bool) is a part of the [`PyNumberMethods`](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_as_number) struct, which roughly corresponds to the `PyNumberProtocol` trait. I don't know why it is done this way, but you may also consider deprecating `__bool__()` from `PyObjectProtocol` and reintroducing it into `PyNumberProtocol` for consistency.

davidhewitt:
Given that the current design decision in #1206 is that we might try completely replacing `#[pyproto]` with `#[pymethods]` for 0.15, I'm not going to bother moving `__bool__` for now.

If we can't make that change work I agree there's a couple of methods like this which should be shuffled around.

kngwyu:
> I don't know why it is done this way,

https://docs.python.org/3/reference/datamodel.html#object.__bool__

Though I don't like this inconsistency, but maybe this is the reason.

ravenexp:
> Given that the current design decision in #1206 is that we might try completely replacing #[pyproto] with #[pymethods] for 0.15 ...

That would be sweet. I'm really looking forward to it.

PR #1561, opened 2021-04-13
pyproto: no self receivers

### LINKED ISSUE #1205: Trouble using PyContextProtocol

davidhewitt:
Fixes #1205 

This PR changes all `#[pyproto]` methods to use `TryFromPyCell` trait instead of having `&self` or `&mut self` receivers, as proposed as option 1 in https://github.com/PyO3/pyo3/issues/1206#issuecomment-760549473

Needs a lot of documenting before this can merge - it's a pretty tedious breaking change for users to have to migrate from. It's so annoying that I'm not sure the advantages are worth the migration costs for most users. For users that have problems noted in #1206, this PR enables access to `self.py()` and base classes - `PyRef` can do these things, `&self` can't.

I'd welcome feedback on whether this is worth merging, or we should consider instead the more drastic option 2 in https://github.com/PyO3/pyo3/issues/1206#issuecomment-760549473 and remove `#[pyproto]` completely.

Moving all `#[pyproto]` methods to `#[pymethods]` might be a much easier migration for users than having to change all `&self` to `PyRef<Self>`.

### DIFF
--- a/pyo3-macros-backend/src/defs.rs
+++ b/pyo3-macros-backend/src/defs.rs
@@ -145,26 +145,18 @@ pub const OBJECT: Proto = Proto {
     name: "Object",
     module: "pyo3::class::basic",
     methods: &[
-        MethodProto::new("__getattr__", "PyObjectGetAttrProtocol")
-            .args(&["Name"])
-            .has_self(),
-        MethodProto::new("__setattr__", "PyObjectSetAttrProtocol")
-            .args(&["Name", "Value"])
-            .has_self(),
-        MethodProto::new("__delattr__", "PyObjectDelAttrProtocol")
-            .args(&["Name"])
-            .has_self(),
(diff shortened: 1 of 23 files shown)


davidhewitt:
I'm going to close this PR, as it looks like we now plan to attempt to merge `#[pyproto]` into `#[pymethods]`.

PR #1563, opened 2021-04-17
ffi: prevent segfault with datetime bindings

davidhewitt:
This fixes a segfault achievable entirely in safe Rust:

```
fn main() {
    assert!(pyo3::ffi::PyDateTimeAPI.DateType == std::ptr::null_mut())
}
```

The `Deref` implementation for `PyDateTimeAPI` can call Python APIs, but before this PR it didn't check whether the GIL is actually held.

### DIFF
--- a/src/ffi/datetime.rs
+++ b/src/ffi/datetime.rs
@@ -435,26 +435,18 @@ pub struct PyDateTime_CAPI {
 
 // Python already shares this object between threads, so it's no more evil for us to do it too!
 unsafe impl Sync for PyDateTime_CAPI {}
-static PY_DATETIME_API: GILOnceCell<&'static PyDateTime_CAPI> = GILOnceCell::new();
 
-#[derive(Debug)]
-pub struct PyDateTimeAPI {
-    __private_field: (),
-}
-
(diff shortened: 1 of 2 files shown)


src/ffi/datetime.rs
```diff
@@ -435,26 +435,18 @@ pub struct PyDateTime_CAPI {
...
-
-pub static PyDateTimeAPI: PyDateTimeAPI = PyDateTimeAPI {
-    __private_field: (),
+/// Safe wrapper around the Python datetime C-API global. Note that this object differs slightly
+/// from the equivalent C object: in C, this is implemented as a `static PyDateTime_CAPI *`. Here
+/// this is implemented as a wrapper which implements [`Deref`] to access a reference to a
+/// [`PyDateTime_CAPI`] object.
+///
+/// In the [`Deref`] implementation, if the underlying object has not yet been initialized, the GIL
+/// will automatically be acquired and [`PyDateTime_IMPORT()`] called.
```
kngwyu, line 445: Are these correctly linked?

kngwyu (approved):
Thanks!

src/ffi/datetime.rs
davidhewitt: Yes, I checked in the docs locally🚀

PR #1564, opened 2021-04-18
ci: allow failure on cache errors

davidhewitt:
... I'm hoping this will stop the cache upload at the end from randomly failing the jobs.


davidhewitt:
Well, looks like this worked. See post-run cache job here, where the cache upload failed but the job still succeeds: https://github.com/PyO3/pyo3/pull/1564/checks?check_run_id=2372900247

CI errors now are a new unrelated PyPy bug: https://foss.heptapod.net/pypy/pypy/-/issues/3442

So I'm going to merge this now and then get on with fixing the PyPy issue.

PR #1565, opened 2021-04-18
pypy: support PyPy 7.3.4

davidhewitt:
This works around a change in the `LIBDIR` variable in the PyPy 7.3.4 release.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -758,6 +758,14 @@ fn configure(interpreter_config: &InterpreterConfig) -> Result<()> {
             if let Some(libdir) = &interpreter_config.libdir {
                 println!("cargo:rustc-link-search=native={}", libdir);
             }
+            if interpreter_config.implementation == PythonInterpreterKind::PyPy {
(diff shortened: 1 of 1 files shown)


messense (approved):
LGTM

mattip:
How does this work on windows?

mattip:
Ahh, nevermind, I see the windows clause is just above this change. So on CPython, linux hits the `interpreter_config.libdir` path? What is that value for CPython and where does it come from?

davidhewitt:
`interpreter_config.libdir` is fished out of the Python interpreter using `sysconfig.get_config_var("LIBDIR")`.

We invoke `python` as part of PyO3's build script to extract the configuration we need:

https://github.com/PyO3/pyo3/blob/88d86a65aa78bf2d001753f994fe3f6db1d8d75e/build.rs#L704

(When cross-compiling we can't do this and resort to other tricks such as scraping the `sysconfig` data script for the target platform.)

mattip:
Thanks. I think I am coming around to the idea this is a regression in PyPy in version 7.3.4

Issue #207, opened 2018-08-22
Support Python 3.7 timezone in datetime bindings


pganssle:
As part of #200 (which added the Python datetime bindings), the update for the full Python 3.7 interface, which includes:

- [`PyDateTime_TimeZone_UTC`](https://docs.python.org/3/c-api/datetime.html#c.PyDateTime_TimeZone_UTC)
- [`PyTimeZone_FromOffset`](https://docs.python.org/3/c-api/datetime.html#c.PyTimeZone_FromOffset)
- [`PyTimeZone_FromOffsetAndName`](https://docs.python.org/3/c-api/datetime.html#c.PyTimeZone_FromOffsetAndName)

The timezone API was added to CPython in [PR 5032](https://github.com/python/cpython/pull/5032).

I believe this can be accompanied on the Rust side by a `PyFixedTimeZone` object or something. It's worth noting that no `_Check` methods accompany this, because you are really not intended to subclass these things.

pickfire:
Can I get some pointers on how to provide a wrapper for `PyDateTime_TimeZone_UTC`?

I see that we can do `PyObject::from_borrowed_ptr(py, PyDateTimeAPI.TimeZone_UTC)` but should we do it the same way as `python::NotImplemented()`? I wonder how should we handle the singleton in this case?

Also, how can one do `<PyXXXUTC as PyTryFrom>::try_from(ob)` when implementing `FromPyObject` for `chrono::Utc`? I don't get how do we `PyTryFrom` a thing that does not have a thin wrapper, even if we have, I am still confused if we can use `pyobject_native_type!` for this?

davidhewitt:
@pickfire I'm thinking about the correct design for this; please allow me a few days.

pickfire:
Just discussing, I will only get to use `FromOffset` (which correlates to chrono [`FixedOffset`](https://docs.rs/chrono/0.4.19/chrono/offset/struct.FixedOffset.html)) and `UTC` (which correlates to chrono [`UTC`](https://docs.rs/chrono/0.4.19/chrono/offset/struct.Utc.html)), so minimally I require those two, but while doing that I can probably add the `FixedOffsetWithName` as well.

I believe I can just start with `PyTimeZone_FromOffset` since that is straightforward. Maybe I will keep `UTC` for later.

davidhewitt:
[...]

PR #1573, opened 2021-04-21
ffi: support PyDateTime_TimeZone_UTC

davidhewitt:
This adds a symbol for `PyDateTime_TimeZone_UTC`, as discussed in #207.

I made it work like `PyDateTimeAPI`, where it has a `Deref` impl which in this case yields `&'static PyObject`.  I'm slightly split on whether I like this approach; in particular because it can run a Python import in the background on first use. On the other hand, it's safe and convenient for users.

~~TODO: add a test.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add FFI definition `Py_IS_TYPE`. [#1429](https://github.com/PyO3/pyo3/pull/1429)
 - Add FFI definition `_Py_InitializeMain`. [#1473](https://github.com/PyO3/pyo3/pull/1473)
 - Add tuple and unit struct support for `#[pyclass]` macro. [#1504](https://github.com/PyO3/pyo3/pull/1504)
+- Add FFI definition `PyDateTime_TimeZone_UTC`. [#1572](https://github.com/PyO3/pyo3/pull/1572)
 
 ### Changed
 - Change `PyTimeAcces::get_fold()` to return a `bool` instead of a `u8`. [#1397](https://github.com/PyO3/pyo3/pull/1397)
--- a/src/ffi/datetime.rs
+++ b/src/ffi/datetime.rs
@@ -399,13 +399,11 @@ pub struct PyDateTime_CAPI {
     pub TimeZone_FromTimeZone:
         unsafe extern "C" fn(offset: *mut PyObject, name: *mut PyObject) -> *mut PyObject,
 
-    // Defined for PyPy as `PyDateTime_FromTimestamp`
     pub DateTime_FromTimestamp: unsafe extern "C" fn(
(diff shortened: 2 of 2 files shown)


pickfire:
Nice, thanks a lot for showing me how to do this, I probably will get stuck forever on the UTC thing if I try it myself especially with the 3 cast.

Just wondering, what are the other approaches?

davidhewitt:
> Just wondering, what are the other approaches?

Well, for example we could make some kind of `static mut` global which would be more similar to the underlying C API, but that'd be quite fiddly for users to handle safely!

kngwyu (approved):
Nice, thank you!

pickfire:
https://github.com/python/cpython/blob/3.9/Lib/datetime.py#L2217

```python
class timezone(tzinfo):
    __slots__ = '_offset', '_name'
```

I am not sure how to implement slots within pyo3 src/types/datetime. From what I see we don't have any places that have a subclass of an existing python type which is a wrapper in rust for me to reference. I saw architecture.md and `pyclass_slots` but I only see `__dict__` and `__weakref__` there but not `__slots__`, am I missing something here?

Also, am I supposed to do
```rust
// src/types/datetime.rs
#[pyclass(extends=PyTzInfo)]
pub struct PyTimeZone {
    _offset: *mut PyObject,
    _name: *mut PyObject,
}
```
Or maybe am I doing it wrongly?

davidhewitt:
I think you should be aiming for a type similar to `PyList` defined in `src/types/list.rs`, where the `pyobject_native_var_type!` macro is used to set up all the right trait implementations.

That should be enough for getting a type which can be used to implement methods and conversions with.

pickfire:
But `PyTimeZone` does not have a C type, from what I see the type is being defined in python library. Which is why I am confused, like what do I put for `ffi::XXXType` for `pyobject_native_var_type!`?

davidhewitt:
Ah, I'd completely missed that was the case. In which case I think that we can just refer to the existing `PyTzInfo` type (which is the base class of `datetime.timezone`) - that should be good enough for now?

I opened #1588 - do you think that'll resolve what you need?

Also, regarding `chrono` implementation... I kind of wonder if it would be easier for the community if the conversions lived here in `src/conversions/chrono.rs`, as an optional feature? Depends if you think the progress of implementing them in `chrono` is going anywhere.

pickfire:
Chrono implementation is just missing the datetime, putting it in chrono pros is that we can use the internal representation directly and that would probably be faster if the compiler is not able to optimize it. But the cons is that it is troublesome without bumping the compiler version, maybe just for the feature it can have a higher compiler version. So far I am stuck because chrono is missing PyTzInfo (including FixedOffset) and UTC, I wanted to try adding it here but seemed harder than I thought.

Yup, the patch that you send seemed a lot better. I thought I need to `impl PyTzInfo` but I guess it is not necessary. I guess that's  all I need, thanks for the quick response.

PR #1574, opened 2021-04-24
build: refactor for simplicity

davidhewitt:
This is a collection of a few small refactorings to `build.rs`: 
 - Made the error message slightly nicer when accidentally compiling on a Python version which is too old.
 - Moved `BuildFlags` struct inside the `InterpreterConfig`, which I think is logically correct.
 - The `abi3_without_interpreter` configuration now uses the same code path to emit cargo configuration as the rest of the script. This is motivated by changes in #1128 where I noticed the `configure_const_generics()` call was nearly missed out of the `abi3_without_interpreter` function, which would have led to confusing bugs.
 - A few other minor updates to comments which had gotten very stale.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,18 +3,17 @@ use std::{
     convert::AsRef,
     env,
     ffi::OsString,
+    fmt::Display,
     fs::{self, DirEntry},
-    io,
     path::{Path, PathBuf},
     process::{Command, Stdio},
     str::FromStr,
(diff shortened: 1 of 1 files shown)


davidhewitt:
CI failure looks like https://github.com/rust-lang/rust/issues/84502 - I think we may have to live with coverage job being down for the next few days...

PR #1577, opened 2021-04-27
coverage: also run compile_error tests

davidhewitt:
At the moment the compile errors only get tested on stable Rust, so they aren't included in the coverage job.

This is a little experiment to see if they get enabled - it should improve coverage information for the macros if it worked.

### DIFF
--- a/tests/test_compile_error.rs
+++ b/tests/test_compile_error.rs
@@ -1,6 +1,19 @@
 #[rustversion::stable]
 #[test]
 fn test_compile_errors() {
+    // stable - require all tests to pass
+    _test_compile_errors()
+}
+
+#[rustversion::nightly]
+#[test]
+fn test_compile_errors() {
+    // nightly - don't care if test output is potentially wrong, to avoid churn in PyO3's CI thanks
+    // to diagnostics changing on nightly.
+    let _ = std::panic::catch_unwind(_test_compile_errors);
+}
+
+fn _test_compile_errors() {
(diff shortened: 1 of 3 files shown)


davidhewitt:
Hmm doesn't seem to have worked 😅 I'll try again soon...

tests/test_compile_error.rs
```diff
@@ -1,6 +1,19 @@
...
+    // stable - require all tests to pass
+    _test_compile_errors()
+}
+
+#[rustversion::nightly]
+#[test]
+fn test_compile_errors() {
+    // nightly - don't care if test output is potentially wrong, to avoid churn in PyO3's CI thanks
+    // to diagnostics changing on nightly.
+    let _ = std::panic::catch_unwind(_test_compile_errors);
```
kngwyu, line 13: How about displaying some warnings?

davidhewitt: `trybuild` actually already prints a lot of warnings if the output is wrong, this just catches the `panic!` when trybuild attempts to fail the test. So I don't think we need to print anything else.

davidhewitt:
I got this working using `cargo-llvm-cov` (many thanks @taiki-e!)

`multiple-pymethods` feature no longer has coverage measured (see https://github.com/taiki-e/cargo-llvm-cov/issues/50), however overall this has still increased coverage numbers a lot which is very cool!

messense (approved):

PR #1582, opened 2021-04-29
examples: use pip 21.1

davidhewitt:
pip 21.1 is now released so we can use that instead of master.


kngwyu (approved):
Nice 🎉 , thank you.

PR #1584, opened 2021-04-29
pypy: use x64 windows in CI for pypy 3.7

davidhewitt:
Since PyPy 7.3.4, the downloads for windows are only 64-bit, which is why CI started failing.


messense (approved):
LGTM

PR #1585, opened 2021-04-29
Proc macro documentation

mejrs:
Adds documentation for the proc macros. 

It's mostly meant as a quick reference, as well as a future place to add doc aliases (they are not very discoverable right now).

<p>

image

image

image

image

image

image

</p>
</details>

### DIFF
--- a/pyo3-macros/src/lib.rs
+++ b/pyo3-macros/src/lib.rs
@@ -74,11 +74,11 @@ pub fn pyproto(_: TokenStream, input: TokenStream) -> TokenStream {
 /// |  Parameter  |  Description |
 /// | :-  | :- |
 /// | `name = "python_name"` | Sets the name that Python sees this class as. Defaults to the name of the Rust struct. |
-/// | `freelist = N`  |  Implements a [free list][10] of size N. This can improve performance for types that are often created and deleted in a row. Profile your code to see whether `freelist` is right for you.  |
+/// | `freelist = N`  |  Implements a [free list][10] of size N. This can improve performance for types that are often created and deleted in quick succession. Profile your code to see whether `freelist` is right for you.  |
 /// | `gc`  | Participate in Python's [garbage collection][5]. Required if your type contains references to other Python objects. If you don't (or incorrectly) implement this, contained Python objects may be hidden from Python's garbage collector and you may leak memory. Note that leaking memory, while undesirable, [is safe behavior][7].|
 /// | `weakref` | Allows this class to be [weakly referenceable][6]. |
 /// | `extends = BaseType`  | Use a custom baseclass. Defaults to [`PyAny`][4] |
-/// | `subclass` | Allows Python classes to inherit from this class.  |
+/// | `subclass` | Allows other Python classes and `#[pyclass]` to inherit from this class.  |
 /// | `unsendable`<!-- hack to stop the column from wrapping -->&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160; | Required if your struct is not [`Send`][3]. Rather than using `unsendable`, consider implementing your struct in a threadsafe way by e.g. substituting [`Rc`][8] with [`Arc`][9]. By using `unsendable`, your class will panic when accessed by another thread.|
 /// | `module = "module_name"` |  Python code will see the class as being defined in this module. Defaults to `builtins`. |
 ///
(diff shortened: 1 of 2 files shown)


kngwyu (approved):
Wow, this is amazing. I hope this would improve experiences for new contributors.

mejrs:
I figured out that docs for #pyclass and #pymethods don't get rendered if the `multiple_pymethods` flag is on, this should fix that

pyo3-macros/src/lib.rs
```diff
@@ -55,26 +67,152 @@ pub fn pyproto(_: TokenStream, input: TokenStream) -> TokenStream {
...
+/// | `gc`  | Participate in Python's [garbage collection][5]. Required if your type contains references to other Python objects. If you don't (or incorrectly) implement this, contained Python objects may be hidden from Python's garbage collector and you may leak memory. Note that leaking memory, while undesirable, [is safe behavior][7].|
+/// | `weakref` | Allows this class to be [weakly referenceable][6]. |
+/// | `extends = BaseType`  | Use a custom baseclass. Defaults to [`PyAny`][4] |
+/// | `subclass` | Allows Python classes to inherit from this class.  |
+/// | `unsendable`<!-- hack to stop the column from wrapping -->&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160;&#160; | Required if your struct is not [`Send`][3]. Rather than using `unsendable`, consider implementing your struct in a threadsafe way by e.g. substituting [`Rc`][8] with [`Arc`][9]. By using `unsendable`, your class will panic when accessed by another thread.|
```
davidhewitt, line 82: Can you explain how this hack works? 🧠

pyo3-macros/src/lib.rs
```diff
@@ -55,26 +67,152 @@ pub fn pyproto(_: TokenStream, input: TokenStream) -> TokenStream {
...
 }
 
+/// A proc macro used to expose Rust structs as Python objects.
+///
+/// `#[pyclass]` accepts the following [parameters][2]:
+///
+/// |  Parameter  |  Description |
+/// | :-  | :- |
+/// | `name = "python_name"` | Sets the name that Python sees this class as. Defaults to the name of the Rust struct. |
+/// | `freelist = N`  |  Implements a [free list][10] of size N. This can improve performance for types that are often created and deleted in a row. Profile your code to see whether `freelist` is right for you.  |
```
davidhewitt, line 77: Slight wording adjustment.

```suggestion
/// | `freelist = N`  |  Implements a [free list][10] of size N. This can improve performance for types that are often created and deleted in quick succession. Profile your code to see whether `freelist` is right for you.  |
```

pyo3-macros/src/lib.rs
```diff
@@ -55,26 +67,152 @@ pub fn pyproto(_: TokenStream, input: TokenStream) -> TokenStream {
...
+/// | `gc`  | Participate in Python's [garbage collection][5]. Required if your type contains references to other Python objects. If you don't (or incorrectly) implement this, contained Python objects may be hidden from Python's garbage collector and you may leak memory. Note that leaking memory, while undesirable, [is safe behavior][7].|
+/// | `weakref` | Allows this class to be [weakly referenceable][6]. |
+/// | `extends = BaseType`  | Use a custom baseclass. Defaults to [`PyAny`][4] |
+/// | `subclass` | Allows Python classes to inherit from this class.  |
```
davidhewitt, line 81: Note that this includes other `#[pyclass]`es!

```suggestion
/// | `subclass` | Allows other classes to inherit from this class.  |
```

pyo3-macros/src/lib.rs
```diff
@@ -55,26 +67,152 @@ pub fn pyproto(_: TokenStream, input: TokenStream) -> TokenStream {
...
+/// [7]: https://doc.rust-lang.org/nomicon/leaking.html
+/// [8]: std::rc::Rc
+/// [9]: std::sync::Arc
+/// [10]: https://en.wikipedia.org/wiki/Free_list
 #[proc_macro_attribute]
 pub fn pyclass(attr: TokenStream, input: TokenStream) -> TokenStream {
     pyclass_impl(attr, input, PyClassMethodsType::Specialization)
 }
 
+/// A proc macro used to expose Rust structs as Python objects.
```
davidhewitt, line 103: Ungh, duplicating this doc kinda hurts, but it's my fault.

For now I think it's okay to copy-paste, I'll try to see if there's a way to make it work nicer in the future.

pyo3-macros/src/lib.rs
```diff
@@ -55,26 +67,152 @@ pub fn pyproto(_: TokenStream, input: TokenStream) -> TokenStream {
...
+///
+/// |  Annotation  |  Description |
+/// | :-  | :- |
+/// | [`#[new]`][4]  | Defines the class constructor, like Python's `__new__` method. |
+/// | [`#[getter]`][5] and [`#[setter]`][5] | These define getters and setters, similar to Python's `@property` decorator. This is useful for getters/setters that require computation or side effects; if that is not the case consider using [`#[pyo3(get, set)]`][11] on the struct's field(s).|
+/// | [`#[staticmethod]`][6]| Defines the method as a staticmethod, like Python's `@staticmethod` decorator.|
+/// | [`#[classmethod]`][7]  | Defines the method as a classmethod, like Python's `@classmethod` decorator.|
+/// | [`#[call]`][8]  | Allows Python code to call a class instance as a function, like Python's `__call__` method. |
+/// | [`#[classattr]`][9]  | Defines a class variable. |
+/// | [`#[args]`][10]  | Define a method's default arguments and allows the function to receive `*args` and `**kwargs`.  |
```
davidhewitt, line 148: There's actually also `#[name = "..."]`, but that's not documented. I am planning to move it to `#[pyo3(name = "...")]`, and same for `#[args]`. Then we can document all the `#[pyo3]` options. See #1567 - I'd be really interested in your opinion on this.

Individual arguments can also have `#[pyo3(from_py_with)]`.

davidhewitt (approved):
👍 it's _so nice_ to have this documentation! Thank you for taking the time to write it 💯 

I have just a few questions, suggestions and tweaks...

pyo3-macros/src/lib.rs
mejrs: They're whitespaces that do not permit wrapping.

If something tries minimizing the table column, "ham spam eggs" can wrap around their spaces, but "ham&#160;spam&#160;eggs" (`ham&#160;spam&#160;eggs`) will not.

pyo3-macros/src/lib.rs
mejrs: I think this is even better to note explicitly: "Allows Python classes and other `#[pyclass]` to inherit from this class."

pyo3-macros/src/lib.rs
mejrs: That part was copied from the guide 😶

pyo3-macros/src/lib.rs
mejrs: I think it's best to wait with making guides about `#pyo3(..)]` until that has happened.

pyo3-macros/src/lib.rs
davidhewitt: Ah, understood.

I think it's slightly more intuitive (although admittedly more ugly) to do something like 

```
`<span style="white-space: pre">`name = "python_name"`</span>`
```

on the rows that we don't want to wrap, but I'll do that in a separate PR.

PR #1588, opened 2021-04-30
datetime: support timezone bindings

### LINKED ISSUE #207: Support Python 3.7 timezone in datetime bindings

davidhewitt:
This PR implements safe wrappers for the Python 3.7 timezone bindings described in #207.

Rather than try to do anything complicated with a new type representing _exactly_ `datetime.timezone`, I return `&PyTzInfo` from the three new functions `timezone_utc()`, `timezone_from_offset()` and `timezone_from_offset_and_name()`. I think this is good enough to be useful; `datetime.timezone` is a subclass of `datetime.tzinfo` so doing this is correct.

I also adjusted a few constructors taking `tzinfo` arguments to take `Option<&PyTzInfo>` instead of `Option<&PyObject>`.

Finally, I tweaked the ffi symbol `PyDateTime_TimeZone_UTC` I added recently in #1573 to resolve to `&'static *mut ffi::PyObject`, which I think is a more correct representation of the "unsafe ffi" nature of that global.

Closes #207

cc @pganssle in case you might be interested / have an opinion on this.

### DIFF
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -21,7 +21,7 @@ use crate::ffi::{
 };
 use crate::instance::PyNativeType;
 use crate::types::PyTuple;
-use crate::{AsPyPointer, PyAny, PyObject, Python, ToPyObject};
+use crate::{AsPyPointer, IntoPy, Py, PyAny, Python};
 use std::os::raw::c_int;
 
 fn ensure_datetime_api(_py: Python<'_>) -> &'static PyDateTime_CAPI {
@@ -244,7 +244,7 @@ impl PyDateTime {
         minute: u8,
         second: u8,
         microsecond: u32,
-        tzinfo: Option<&PyObject>,
+        tzinfo: Option<&PyTzInfo>,
     ) -> PyResult<&'p PyDateTime> {
         let api = ensure_datetime_api(py);
         unsafe {
@@ -280,7 +280,7 @@ impl PyDateTime {
         minute: u8,
         second: u8,
         microsecond: u32,
-        tzinfo: Option<&PyObject>,
+        tzinfo: Option<&PyTzInfo>,
         fold: bool,
     ) -> PyResult<&'p PyDateTime> {
         let api = ensure_datetime_api(py);
@@ -303,20 +303,13 @@ impl PyDateTime {
 
     /// Construct a `datetime` object from a POSIX timestamp
     ///
-    /// This is equivalent to `datetime.datetime.from_timestamp`
+    /// This is equivalent to `datetime.datetime.fromtimestamp`
     pub fn from_timestamp<'p>(
         py: Python<'p>,
         timestamp: f64,
-        time_zone_info: Option<&PyTzInfo>,
+        tzinfo: Option<&PyTzInfo>,
     ) -> PyResult<&'p PyDateTime> {
-        let timestamp: PyObject = timestamp.to_object(py);
-
-        let time_zone_info: PyObject = match time_zone_info {
-            Some(time_zone_info) => time_zone_info.to_object(py),
-            None => py.None(),
-        };
-
-        let args = PyTuple::new(py, &[timestamp, time_zone_info]);
+        let args: Py<PyTuple> = (timestamp, tzinfo).into_py(py);
 
         // safety ensure API is loaded
         let _api = ensure_datetime_api(py);
@@ -396,7 +389,7 @@ impl PyTime {
         minute: u8,
         second: u8,
         microsecond: u32,
-        tzinfo: Option<&PyObject>,
+        tzinfo: Option<&PyTzInfo>,
     ) -> PyResult<&'p PyTime> {
         let api = ensure_datetime_api(py);
         unsafe {
@@ -419,7 +412,7 @@ impl PyTime {
         minute: u8,
         second: u8,
         microsecond: u32,
-        tzinfo: Option<&PyObject>,
+        tzinfo: Option<&PyTzInfo>,
         fold: bool,
     ) -> PyResult<&'p PyTime> {
         let api = ensure_datetime_api(py);
@@ -473,9 +466,11 @@ impl PyTzInfoAccess for PyTime {
     }
 }
 
-/// Bindings for `datetime.tzinfo`
+/// Bindings for `datetime.tzinfo`.
 ///
-/// This is an abstract base class and should not be constructed directly.
+/// This is an abstract base class and cannot be constructed directly.
+/// For concrete time zone implementations, see [`timezone_utc`] and
+/// the [`zoneinfo` module](https://docs.python.org/3/library/zoneinfo.html).
 #[repr(transparent)]
 pub struct PyTzInfo(PyAny);
 pyobject_native_type!(
@@ -485,6 +480,11 @@ pyobject_native_type!(
     #module=Some("datetime"),
     #checkfunction=PyTZInfo_Check
 );
+
+/// Equivalent to `datetime.timezone.utc`
+pub fn timezone_utc(py: Python<'_>) -> &PyTzInfo {
+    unsafe { &*(ensure_datetime_api(py).TimeZone_UTC as *const PyTzInfo) }
+}
 
 /// Bindings for `datetime.timedelta`
 #[repr(transparent)]
@@ -535,7 +535,7 @@ impl PyDeltaAccess for PyDelta {
 }
 
 // Utility function
-fn opt_to_pyobj(py: Python<'_>, opt: Option<&PyObject>) -> *mut ffi::PyObject {
+fn opt_to_pyobj(py: Python<'_>, opt: Option<&PyTzInfo>) -> *mut ffi::PyObject {
     // Convenience function for unpacking Options to either an Object or None
     match opt {
         Some(tzi) => tzi.as_ptr(),
@@ -545,12 +545,51 @@ fn opt_to_pyobj(py: Python<'_>, opt: Option<&PyObject>) -> *mut ffi::PyObject {
 
 #[cfg(test)]
 mod tests {
+    use super::*;
+    #[cfg(feature = "macros")]
+    use crate::py_run;
+
+    #[test]
+    #[cfg(feature = "macros")]
+    #[cfg_attr(target_arch = "wasm32", ignore)] // DateTime import fails on wasm for mysterious reasons
+    fn test_datetime_fromtimestamp() {
+        Python::with_gil(|py| {
+            let dt = PyDateTime::from_timestamp(py, 100.0, None).unwrap();
+            py_run!(
+                py,
+                dt,
+                "import datetime; assert dt == datetime.datetime.fromtimestamp(100)"
+            );
+
+            {
+                let dt = PyDateTime::from_timestamp(py, 100.0, Some(timezone_utc(py))).unwrap();
+                py_run!(
+                py,
+                dt,
+                "import datetime; assert dt == datetime.datetime.fromtimestamp(100, datetime.timezone.utc)"
+            );
+            }
+        })
+    }
+
+    #[test]
+    #[cfg(feature = "macros")]
+    #[cfg_attr(target_arch = "wasm32", ignore)] // DateTime import fails on wasm for mysterious reasons
+    fn test_date_fromtimestamp() {
+        Python::with_gil(|py| {
+            let dt = PyDate::from_timestamp(py, 100).unwrap();
+            py_run!(
+                py,
+                dt,
+                "import datetime; assert dt == datetime.date.fromtimestamp(100)"
+            );
+        })
+    }
+
     #[test]
     #[cfg_attr(target_arch = "wasm32", ignore)] // DateTime import fails on wasm for mysterious reasons
     fn test_new_with_fold() {
-        crate::Python::with_gil(|py| {
-            use crate::types::{PyDateTime, PyTimeAccess};
-
+        Python::with_gil(|py| {
             let a = PyDateTime::new_with_fold(py, 2021, 1, 23, 20, 32, 40, 341516, None, false);
             let b = PyDateTime::new_with_fold(py, 2021, 1, 23, 20, 32, 40, 341516, None, true);
 
@@ -559,29 +598,23 @@ mod tests {
         });
     }
 
-    #[cfg(not(PyPy))]
     #[test]
     #[cfg_attr(target_arch = "wasm32", ignore)] // DateTime import fails on wasm for mysterious reasons
     fn test_get_tzinfo() {
         crate::Python::with_gil(|py| {
-            use crate::conversion::ToPyObject;
-            use crate::types::{PyDateTime, PyTime, PyTzInfoAccess};
+            let utc = timezone_utc(py);
 
-            let datetime = py.import("datetime").map_err(|e| e.print(py)).unwrap();
-            let timezone = datetime.getattr("timezone").unwrap();
-            let utc = timezone.getattr("utc").unwrap().to_object(py);
+            let dt = PyDateTime::new(py, 2018, 1, 1, 0, 0, 0, 0, Some(utc)).unwrap();
 
-            let dt = PyDateTime::new(py, 2018, 1, 1, 0, 0, 0, 0, Some(&utc)).unwrap();
-
-            assert!(dt.get_tzinfo().unwrap().eq(&utc).unwrap());
+            assert!(dt.get_tzinfo().unwrap().eq(utc).unwrap());
 
             let dt = PyDateTime::new(py, 2018, 1, 1, 0, 0, 0, 0, None).unwrap();
 
             assert!(dt.get_tzinfo().is_none());
 
-            let t = PyTime::new(py, 0, 0, 0, 0, Some(&utc)).unwrap();
+            let t = PyTime::new(py, 0, 0, 0, 0, Some(utc)).unwrap();
 
-            assert!(t.get_tzinfo().unwrap().eq(&utc).unwrap());
+            assert!(t.get_tzinfo().unwrap().eq(utc).unwrap());
 
             let t = PyTime::new(py, 0, 0, 0, 0, None).unwrap();
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `timezone_utc()`. [#1588](https://github.com/PyO3/pyo3/pull/1588)
 - Implement `ToPyObject` for `[T; N]`. [#2313](https://github.com/PyO3/pyo3/pull/2313)
 - Added the internal `IntoPyResult` trait to give better error messages when function return types do not implement `IntoPy`. [#2326](https://github.com/PyO3/pyo3/pull/2326)
 - Add `PyDictKeys`, `PyDictValues` and `PyDictItems` Rust types to represent `dict_keys`, `dict_values` and `dict_items` types. [#2358](https://github.com/PyO3/pyo3/pull/2358)
@@ -26,10 +27,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `CompareOp::matches` to easily implement `__richcmp__` as the result of a
   Rust `std::cmp::Ordering` comparison. [#2460](https://github.com/PyO3/pyo3/pull/2460)
 - Supprt `#[pyo3(name)]` on enum variants [#2457](https://github.com/PyO3/pyo3/pull/2457)
-- Add `PySuper` object [#2049](https://github.com/PyO3/pyo3/issues/2049) 
+- Add `PySuper` object [#2049](https://github.com/PyO3/pyo3/issues/2049)
 
 ### Changed
 
+- Change datetime constructors taking a `tzinfo` to take `Option<&PyTzInfo>` instead of `Option<&PyObject>`: `PyDateTime::new()`, `PyDateTime::new_with_fold()`, `PyTime::new()`, and `PyTime::new_with_fold()`. [#1588](https://github.com/PyO3/pyo3/pull/1588)
 - Several methods of `Py` and `PyAny` now accept `impl IntoPy<Py<PyString>>` rather than just `&str` to allow use of the `intern!` macro. [#2312](https://github.com/PyO3/pyo3/pull/2312)
 - Move `PyTypeObject::type_object` method to `PyTypeInfo` trait, and deprecate `PyTypeObject` trait. [#2287](https://github.com/PyO3/pyo3/pull/2287)
 - The deprecated `pyproto` feature is now disabled by default. [#2322](https://github.com/PyO3/pyo3/pull/2322)
--- a/pyo3-ffi/src/datetime.rs
+++ b/pyo3-ffi/src/datetime.rs
@@ -574,8 +574,17 @@ pub unsafe fn PyTZInfo_CheckExact(op: *mut PyObject) -> c_int {
 // skipped non-limited PyTime_FromTime
 // skipped non-limited PyTime_FromTimeAndFold
 // skipped non-limited PyDelta_FromDSU
-// skipped non-limited PyTimeZone_FromOffset
-// skipped non-limited PyTimeZone_FromOffsetAndName
+
+pub unsafe fn PyTimeZone_FromOffset(offset: *mut PyObject) -> *mut PyObject {
+    ((*PyDateTimeAPI()).TimeZone_FromTimeZone)(offset, std::ptr::null_mut())
+}
+
+pub unsafe fn PyTimeZone_FromOffsetAndName(
+    offset: *mut PyObject,
+    name: *mut PyObject,
+) -> *mut PyObject {
+    ((*PyDateTimeAPI()).TimeZone_FromTimeZone)(offset, name)
+}
 
 #[cfg(not(PyPy))]
 pub unsafe fn PyDateTime_FromTimestamp(args: *mut PyObject) -> *mut PyObject {
--- a/pytests/src/datetime.rs
+++ b/pytests/src/datetime.rs
@@ -33,14 +33,7 @@ fn make_time<'p>(
     microsecond: u32,
     tzinfo: Option<&PyTzInfo>,
 ) -> PyResult<&'p PyTime> {
-    PyTime::new(
-        py,
-        hour,
-        minute,
-        second,
-        microsecond,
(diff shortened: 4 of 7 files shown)


src/types/datetime.rs
```diff
@@ -279,7 +252,7 @@ impl PyTime {
         minute: u8,
         second: u8,
         microsecond: u32,
-        tzinfo: Option<&PyObject>,
+        tzinfo: Option<&PyTzInfo>,
```
pickfire, line 255: Wouldn't this be a breaking change? It may probably affect pyo3-chrono crate. Maybe need to ping the author if this is indeed the case.

src/ffi/datetime.rs
```diff
@@ -445,12 +445,11 @@ pub static PyDateTimeAPI: _PyDateTimeAPI_impl = _PyDateTimeAPI_impl {
     inner: GILOnceCell::new(),
 };
 
-/// Safe wrapper around the Python C-API global `PyDateTime_TimeZone_UTC`. This follows a similar
+/// Wrapper around the Python C-API global `PyDateTime_TimeZone_UTC`. This follows a similar
```
pickfire, line 448: Is this not a safe wrapper now?

src/types/datetime.rs
```diff
@@ -356,6 +332,38 @@ pyobject_native_type!(
...
+    name: &str,
+) -> PyResult<&'py PyTzInfo> {
+    let name = name.into_py(py);
+    unsafe {
+        py.from_owned_ptr_or_err(ffi::PyTimeZone_FromOffsetAndName(
+            offset.as_ptr(),
+            name.as_ptr(),
+        ))
+    }
+}
```
pickfire, line 365: Wouldn't these be easier to discover if we put it within `impl PyTzInfo`?

src/ffi/datetime.rs
davidhewitt: Well, I made it less safe because it gives you `&'static *mut ffi::PyObject`, which is a pointer and thus less safe to use. The safe wrapper is now the `pyo3::types::timezone_utc()` function.

src/types/datetime.rs
davidhewitt: I went back and forth on that. Yes, it probably would. However I wasn't sure if that was totally "correct" because they're actually creating a subtype. If you also think it'd be nicer to have them on `impl PyTzInfo` then I'm persuaded that's the right thing to do 👍

src/types/datetime.rs
davidhewitt: Yes, and we'll release this as part of the 0.14 release. I've notified this in the CHANGELOG, and we are still pre-1.0, after all. If it turns out to be a big problem we can revert it later.

src/types/datetime.rs
pickfire: No, I am just thinking that `datetime` module could have tons of functions already and separating these would increase those later on, since these will only create `PyTzInfo` I don't see why not put it there.

pganssle:
I'm not sure I understand what is complicated about having a new type representing `datetime.timezone`. Is it the inheritance parts (`datetime.timezone` is not supposed to be a superclass, so maybe that's why it's weird compared to the other classes?).

All of the other type wrappers we have are Rust types with constructors corresponding to the equivalent C API constructors.

src/types/datetime.rs
pganssle: I can't quite remember why I went with `&PyObject` instead of `&PyTzInfo` here originally. It's entirely possible that I misunderstood how inheritance in Rust and Python can play together well, The `datetime.tzinfo` class is an abstract base class (but I think one that was created before Python supported either abstract base classes or Protocols).

In theory it should be a trait, but in practice `datetime` enforces that it is an actual subclass of `datetime.tzinfo` and I don't know if I'll be able to change that any time soon. Presumably it is safe to change this from `PyTzInfo` back to `PyObject` after the 1.0 release, since all `PyTzInfo` objects are `PyObject`s but not vice versa?

src/types/datetime.rs
pganssle: I definitely don't think this should go in `impl PyTzInfo`. `PyTzInfo` exists only to be a base class, so you shouldn't really be able to get instances of it. Despite Rust's lack of nominal type hierarchies, it needs Rust wrappers so you can do stuff like this: https://github.com/PyO3/pyo3/blob/3ee84a3c26a67686d618cfec8c85463be7506bb0/examples/pyo3-pytests/src/datetime.rs#L181-L202

pickfire:
But this one is a python wrapper. Wouldn't `PyAny` have some performance cost when doing the conversion unlike the C API? Since it uses `__slots__` within python do we have any way to convert it directly?

src/types/datetime.rs
pganssle: Ah, I remember why it has to be `&PyObject` now — `tzinfo` can also be `None` (and in fact probably _is_ `None` in the common case, since `datetime.timezone` is not a very good time zone representation but datetimes are useful anyway).

I'm actually fairly surprised that this doesn't break any tests.

src/types/datetime.rs
pickfire: I don't mean with `&self`, I just mean something like
```rust
impl PyTzinfo {
    fn utcoffset<'py>(py: Python<'py>, offset: &PyDelta) -> PyResult<&'py PyTzInfo> {
        ...
    }
}
```

src/types/datetime.rs
pganssle: Ah, sorry to keep going back and forth here, but I forgot that this is `Option<&PyTzInfo>`, and that the `None` is handled by the `None` case of `Option`, so I'm back to not remembering why this was `PyObject` instead of `PyTzInfo`.

src/types/datetime.rs
pganssle: @pickfire Regardless, I don't think it's the right abstraction, because it implies that this is a *constructor* for `PyTzInfo`, which is an abstract base class.

I do think it would make sense for these methods to turn into a wrapper class for `datetime.timezone` like the others (minus the `_Check` methods and such), in which case these could be alternate constructors for `PyFixedTimezone` or whatever it's called.

davidhewitt:
> I'm not sure I understand what is complicated about having a new type representing `datetime.timezone`.

We haven't yet had to model the inheritance hierarchy of types like this. Would `PyTimeZone` implement `Deref<Target=PyTzInfo>` instead of `Deref<Target=PyAny>`? It probably should (otherwise you wouldn't be able to use the return type of `PyTimeZone::utc()` in a call to `PyTime::new`, for example), but that'd mean reworking PyO3's internal macros.

... I'm going to work on a couple of PRs to see if implementing `Deref` like that is practical, please hold with this...

pganssle:
> We haven't yet had to model the inheritance hierarchy of types like this. Would `PyTimeZone` implement `Deref<Target=PyTzInfo>` instead of `Deref<Target=PyAny>`? It probably should (otherwise you wouldn't be able to use the return type of `PyTimeZone::utc()` in a call to `PyTime::new`, for example), but that'd mean reworking PyO3's internal macros.

It has unfortunately been a while since I had a halfway decent understanding of the way Python inheritance is modeled in PyO3 (if I ever did). This sounds appealing to me, but is there a more general problem that sometimes your `PyTzInfo` needs to be a `PyAny`, so we'd need to do the `Deref` trick twice?

Is it not sufficient to implement `From<&PyFixedTimezone>` for both `&PyAny` and `&PyTzInfo` and satisfy both requirements?

In this case, it would be _really_ nice if we could use a trait instead, since the actual memory layout of the `tzinfo` subclass doesn't matter at all for `datetime`'s purposes, but `datetime` does an `isinstance` check on the `tzinfo` object meaning it must be an actual subclass. We could theoretically change that in Python 3.11, but in some ways that feels unnecessary since: 1. no one has actually expressed any desire to write a time zone class that doesn't inherit from `datetime.tzinfo` and 2. all the PyO3 types will be wrappers around things that actually _are_ `tzinfo` subclasses, and our only struggle is in convincing Rust that that's the case. So while that change would make things more aligned with reality (we only care that you implement the `tzinfo` protocol, not that you pass an actual `tzinfo` object), it's also a totally "on paper" change that would have no practical benefits.

pganssle:
One other comment on this: Other than exposing the `UTC` singleton (which is a great convenience), I don't think the `datetime.timezone` class is terribly important. It's trivially simple to re-implement in Rust if you need something like that (see [this](https://github.com/PyO3/pyo3/blob/3ee84a3c26a67686d618cfec8c85463be7506bb0/examples/pyo3-pytests/src/datetime.rs#L181-L202), which could easily be tweaked to be equivalent to `datetime.timezone`), and I can't think of any particular benefits of using `datetime.timezone` if your `tzinfo` implementation is already in a compiled language.

Plus, it's generally not a good idea to use `datetime.timezone` in the first place. `datetime.timezone` exists so that stuff like the `datetime.fromisoformat` and `datetime.strptime` can construct fixed-point datetimes from some serialized datetime format with a fixed offset using just the standard library. If possible, you should strongly prefer to construct `datetime` objects using [`zoneinfo`](https://docs.python.org/3/library/zoneinfo.html). I realize that `chrono`'s various design flaws makes fixed offsets easier to use and it would be nice to have the `timezone` API exposed, but I'm ambivalent enough about making `datetime.timezone` easy to use that it's probably worth taking the time to come up with a decent interface rather than rushing to get _something_ implemented.

davidhewitt:
> This sounds appealing to me, but is there a more general problem that sometimes your PyTzInfo needs to be a PyAny, so we'd need to do the Deref trick twice?

Yes, and if I implemented this we'd have `impl Deref<Target = PyTzInfo> for PyTimeZone`, and `impl Deref<Target = PyAny> for PyTzInfo`.

> Is it not sufficient to implement From<&PyFixedTimezone> for both &PyAny and &PyTzInfo and satisfy both requirements?

Something like this could be done, yes. However I'm a bit uneasy about adding new trait mappings like `From` without making this an across-the-board change for PyO3.

I think that changes like #1056 need to be resolved before we commit to more ways to interconvert Python types. I'm similarly uneasy about the `Deref` design does at least have precedent in `web-sys`, where Javascript types `Deref` to their super types.

> Plus, it's generally not a good idea to use datetime.timezone in the first place.  [...]

Understood, however I think there's also a practical notion that these APIs _would_ be helpful for chrono as current, and these are exposed in the Python C-API so it's reasonable for PyO3 to expose these too.

----

Overall, I'm leaning towards merging this PR as-is, because it's a simple and practical approach even if the new functions aren't super discoverable. Perhaps their lack of discoverability may be a feature 😉 

We can make future improvements to this API as PyO3 and the Python C-API evolve...

pganssle:
> Overall, I'm leaning towards merging this PR as-is, because it's a simple and practical approach even if the new functions aren't super discoverable. Perhaps their lack of discoverability may be a feature

I'm still kinda -1 for the same reason that I didn't like the idea of adding `chrono` bindings directly to PyO3. It feels out of scope to be adding convenience functions that don't necessarily map to things in the C API. If you just want a nice convenient way to get any fixed offset timezone, it seems like it would be pretty simple to implement a Rust "fixed offset" type in its own crate that doesn't bother with `datetime.timezone`.  If `chrono` is growing PyO3 support ([in this PR](https://github.com/chronotope/chrono/pull/542) from @pickfire), it could also include a simple `FixedOffset` implementation like the one I linked to from the examples.

In general, it's not going to be useful to implement specifically `datetime.timezone` → `chrono` support for a number of reasons *anyway*, so we only need the `chrono` → `datetime` side, so `chrono` also has the option of simply calling the `datetime.timezone` constructor, no?

```rust
impl ToPyObject for FixedOffset {
    fn to_object(&self, py: pyo3::Python) -> pyo3::Result<pyo3::PyObject> {
        let dt_module = py.import("datetime")?;

        let seconds_offset = self.utc_minus_local();  // Throw exception if not ±24h?

        let td = pyo3::PyDelta::new(py, seconds_offset, 0, true)?;

        dt_module.call1((td,))?
    }
}
```

This would be slightly slower than calling the C API directly, but if you are concerned about that, a cache would likely make it _much_ faster than even calling the C API directly; since there are a maximum of ~172,800 possible fixed offsets, and in practice there's maybe 30-40 that you'll _actually_ encounter, any process doing a large number of conversions of fixed offsets to `datetime.timezone` objects would very quickly saturate the cache and you'd go from making python function calls to looking up offsets in a Rust hashmap. On my computer it takes ~500-600ns to call `datetime.timezone(datetime.timedelta(seconds=3600))`, so construction 172,800 of these would take 95ms, Depending on how much overhead Rust adds, I think it would be 12-14MB of memory to store all those `timedelta`s and `timezone`s, and that's the worst case scenario where you cache every `timezone` object you create and you create every possible `timezone` that represents a `chrono::FixedOffset`.

So at the end of the day, I'd say that we should concentrate on creating nice, consistent bindings for CPython in `pyo3` and leave the "convenient way to create an aware datetime from a `chrono` datetime" for either `chrono` or for its own crate.

pickfire:
> Plus, it's generally not a good idea to use datetime.timezone in the first place. datetime.timezone exists so that stuff like the datetime.fromisoformat and datetime.strptime can construct fixed-point datetimes from some serialized datetime format with a fixed offset using just the standard library. If possible, you should strongly prefer to construct datetime objects using zoneinfo. I realize that chrono's various design flaws makes fixed offsets easier to use and it would be nice to have the timezone API exposed, but I'm ambivalent enough about making datetime.timezone easy to use that it's probably worth taking the time to come up with a decent interface rather than rushing to get something implemented.

I think you missed the context here. What I am trying to do is provide chrono and pyo3 integration in which the python timezone is almost the same as chrono provided functions. chrono also does not provide timezone, so rather than having users do the conversion, we can just support it on library level so the experience is seamless.

> it seems like it would be pretty simple to implement a Rust "fixed offset" type in its own crate that doesn't bother with datetime.timezone.

I think that's probably not a good idea in the long run, keeping it align with the C API might be a better idea.

> In general, it's not going to be useful to implement specifically datetime.timezone → chrono support for a number of reasons anyway, so we only need the chrono → datetime side, so chrono also has the option of simply calling the datetime.timezone constructor, no?

Other than chrono there are also other time libraries or other stuff that needs the timezone. I think implementing it here is better than having everyone have their own implementation which they can easily get wrong (like me, I still can't figure out how to do the FFI thing by myself).

The code you showed doesn't seemed like a good idea. pyo3 is exposing safe wrappers over python API. Of course not all parts of C API is necessary but doing it this way seemed weird, the `utc_minus_local` thing feels like a hack to me.

pganssle:
> Other than chrono there are also other time libraries or other stuff that needs the timezone. I think implementing it here is better than having everyone have their own implementation which they can easily get wrong (like me, I still can't figure out how to do the FFI thing by myself).

The problem is that `chrono` does everything with fixed offsets, which has a number of problems. It's not a very well-designed library and I am very sad that it is the _de facto_ standard datetime library in Rust. I'm hoping that other libraries handling dates and times would use something other than fixed offsets, in which case `datetime.timezone` would not be especially useful anyway. For the most part, well-designed uses of `datetime` will either use UTC, local time (naïve, no offset) or an IANA time zone, so if you have something in Rust using fixed offsets that you want to send over to the Python side, you'll want to convert to one of those three before using it _anyway_. In any case, I'm mostly just saying that there's no reason to make a quick release to support this  since I'm fairly convinced that it will be an attractive nuisance in the long run anyway.

> The code you showed doesn't seemed like a good idea. pyo3 is exposing safe wrappers over python API. Of course not all parts of C API is necessary but doing it this way seemed weird, the `utc_minus_local` thing feels like a hack to me.

The `utc_minus_local` part of that is not the important part. If you have a preferred way to get the offset in minutes and seconds from a `FixedOffset`, then do that instead. To put my suggestion another way, any person could easily write their own version of @davidhewitt's `timezone_from_offset` like so:

```rust
fn timezone_from_offset<'py>(py: Python<'py>, offset: &PyDelta) -> PyResult<&'py PyTzInfo> {
    let dt_mod = py.import("datetime")?;
    dt_mod.call1("timezone", (offset))?
}
```

This is a trivial operation and it should be simple to do it from within your `chrono` ↔ `pyo3` bindings. The main benefit of using the C API and C struct is probably going to be speed since (I think) there will be fewer interactions between you and the Python interpreter, but as I mentioned earlier a cache-based solution to any speed problems will probably be much more fruitful if constructing `datetime.timezone` objects ends up as a performance bottleneck.

Additionally, it returns the exact same object that you'd get if you were using the `ffi` wrappers, meaning that if / when `pyo3` works out the right way to represent a `datetime.timezone` object, it would be a simple matter to replace the contents of the function with something using the C API wrappers.

pickfire:
> The problem is that chrono does everything with fixed offsets, which has a number of problems. It's not a very well-designed library and I am very sad that it is the de facto standard datetime library in Rust. I'm hoping that other libraries handling dates and times would use something other than fixed offsets, in which case datetime.timezone would not be especially useful anyway. For the most part, well-designed uses of datetime will either use UTC, local time (naïve, no offset) or an IANA time zone, so if you have something in Rust using fixed offsets that you want to send over to the Python side, you'll want to convert to one of those three before using it anyway. In any case, I'm mostly just saying that there's no reason to make a quick release to support this since I'm fairly convinced that it will be an attractive nuisance in the long run anyway.

I still don't see what is the issue with fixed offsets? Can you please explain it or open an issue with chrono? `FixedOffset` can be set to the second precision so I am not sure what is the issue with that, which is the same as python timezone.

Ah, I didn't know the code will work. Will try it out later and if this patch didn't get merged I will do that. But if other people come do they need to go search the pull request and find that comment? I think having a function there is simpler such that others does not have to even worry about the low level stuff and just use the same API for python.

pganssle:
> I still don't see what is the issue with fixed offsets? Can you please explain it or open an issue with chrono? `FixedOffset` can be set to the second precision so I am not sure what is the issue with that, which is the same as python timezone.

Python offsets can have sub-second precision as well, as of Python 3.7 or 3.8 (I think 3.7). It is also the case that in general, Python datetimes _should_ have a `zoneinfo.Zoneinfo` or equivalent time zone attached, not a fixed offset.

The biggest problem I have with fixed offsets is that they are rarely actually all that useful. Generally speaking, you either want an absolute time, which represents a fixed point in time (e.g. UTC, an epoch timestamp, etc) or you want a civil time, which represents what the clock will say on the wall at a specific time in a specific place. Absolute times are usually useful for recording when something happened in the past (e.g. timestamps on logs) and civil times are usually useful for human-centric events that may take place in the future. When working with absolute times, it's almost always sufficient to record it and work with it in UTC and convert it to local civil time when you want to display it to a human. When working with future times it is _not_ appropriate to use UTC or a fixed offset thereof  — imagine we plan an event at 14:00 a month from now, and then the country we're meeting in decides to eliminate DST. If we had recorded the meeting as 14:00-04:00 — in local time that is now 13:00, even though from our perspective _we always want the meeting to be at 14:00_.

The way Python works is that you attach a `tzinfo` which is an object mapping "civil" times to UTC (the UTC tzinfo is a no-op). This way what is stored is not "local time + offset" but "local time + rule set", which is much more versatile. Fixed offsets are basically just UTC in disguise, and may not be accurate for a given civil time. They are useful for when you want to be able to easily print a given datetime with a given fixed offset, but it's _very rare_ that you'd want to do this. It's also useful for representing all the information you have from, say, an ISO 8601 datetime — in case you want to losslessly re-serialize it. So, for example, if you are given `2021-05-22T03:21:19+02:00`, you don't know what rule set generated that, but you know both the local time (2021-05-22 at 03:21:19) and the offset from UTC (-02:00). `datetime.timezone` exists to satisfy basically this use case, but usually you'd want to convert this to UTC or an IANA zone as soon as you can before doing much with it.


> Ah, I didn't know the code will work. Will try it out later and if this patch didn't get merged I will do that. But if other people come do they need to go search the pull request and find that comment? I think having a function there is simpler such that others does not have to even worry about the low level stuff to have to go search the issues and pull requests and read the source code like I did to figure out how to do the same thing but using a lower-level API, just using a high-level API like the one in python would be easier to find and figure out how it can be used.

First of all, I don't think my suggestion here was terribly complicated. If you do any appreciable amount of coding with Rust ↔ Python or C ↔ Python interop, you'll quickly find that there aren't C API hooks for even _most_ of the standard library, so if you can't find a way to accomplish something using the C API, you just make a Python function call and get a `PyObject *` in your compiled code. Before Python 3.7, anyone would have to do that to construct a `datetime.timezone` from C anyway, and you still must do this to construct the much more useful `zoneinfo.ZoneInfo` object, since `zoneinfo` doesn't (and probably won't) have C API bindings.

I am in a weird position here because I _do_ think that we should expose `datetime.timezone` bindings, but I also think that there's a decent chance that they'll do more harm than good 😅. My point is that _in general_ it should be pretty rare that you want to use one of these things in the first place, so as long as the UTC singleton is already exposed, we should probably take our time carefully designing the timezone bindings, rather than introducing convenience functions that give you convenient access to a minor footgun.

pickfire:
> I am in a weird position here because I do think that we should expose datetime.timezone bindings, but I also think that there's a decent chance that they'll do more harm than good sweat_smile. My point is that in general it should be pretty rare that you want to use one of these things in the first place, so as long as the UTC singleton is already exposed, we should probably take our time carefully designing the timezone bindings, rather than introducing convenience functions that give you convenient access to a minor footgun.

If it's a bug that people rely on. It's not a bug. It's a feature. -- Linus

At least in this case, based on what you said taking DST into account it is a bad design. But considering that RFC 3339 and ISO 8601 supports fixed offsets by default in datetime but not IANA or something similar, I think it is still good to have an API for it, at least python already does it even though it is not recommended, it is not really a foot gun for this to be exact, even chrono does the same thing. But yeah, in the docs we can just put a note that it is more recommended to use IANA rather than the fixed offset (but we don't have a pointer to say what to use yet).

pickfire:
Some time has passed, is there any decisions on this or what's pending?

davidhewitt:
Overall I'm still leaning towards merging this eventually. I understand the concerns that this is not IANA timezone, however we also don't have a C API for the `zoneinfo` stdlib. The position I have on this one is that PyO3 remains unopinionated and offers the APIs that it can; chrono integration will find these useful even if we really need a `zoneinfo` API too.

The main reason I have not been rushing to merge this is that it's currently a bit clunky to use custom `PyTzInfo` . See #1662. I would like to solve that issue at the same time, which will probably be a 0.17 thing at current rate of progress. Having #1662 would make it easier for us (or third-party crate) to define custom `zoneinfo` for IANA support if they needed it.

pickfire:
Okay, then I guess I will just use the other method to call python code directly for chrono, when I have the time to work on that.

davidhewitt:
> One reason it seems unlikely that we would bother with zoneinfo bindings for the C API is that constructing them will basically never be the slow part of any operations you are doing, 

That's fine. I hope that one day we'll have easy ways to generate type-safe wrappers for Python types to reduce the amount of dynamic code users have to write. We could eventually use that to wrap `zoneinfo`.

I started such a project at https://github.com/davidhewitt/pyo3-pytypes, however I think it'll be a long time before I next get a chance to sit down and try again to make it happen.

davidhewitt:
I'd suggest opening as a separate PR, it's probably going to be quite a while before I have a chance to finish this off unfortunately.

davidhewitt:
After a long delay, I'm leaning towards merging this for 0.17.

I appreciate the concern that this does not expose IANA timezones. While that is perhaps something that we can consider in future, for now the Python C API lacks any bindings for `zoneinfo` (as far as I am aware). A future PyO3 could release a statically-typed wrapper around `zoneinfo` which just makes dynamic Python calls underneath (or this could be added as a community crate). For now, I see that https://github.com/chronotope/chrono/pull/542 has incorrect `unsafe` code which could have been avoided by this `timezone_utc` function being available, so I see immediate value to the community in having this safe API added.

pganssle:
> I appreciate the concern that this does not expose IANA timezones.

As I said before, this not exposing IANA time zones is *not* the problem. The problem is that these fixed offsets are both not useful and they **seem like they are useful**, which is a dangerous combination.

davidhewitt:
@pganssle Would it resolve your concern if I took the two `timezone_from_offset` functions out of this PR? (Leaving just `timezone_utc`, as the most commonly used timezone imo.)

src/types/datetime.rs
```diff
@@ -473,9 +466,11 @@ impl PyTzInfoAccess for PyTime {
     }
 }
 
-/// Bindings for `datetime.tzinfo`
+/// Bindings for `datetime.tzinfo`.
 ///
-/// This is an abstract base class and should not be constructed directly.
+/// While `tzinfo` is an abstract base class, the `datetime` module provides one concrete
```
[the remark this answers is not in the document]
davidhewitt: Sure thing, I'll just change to use that 👍

davidhewitt:
> The only comment I'd have about `timezone_utc` is that it seems like a bit of an odd interface to use a function to access a singleton like that, but I don't think it hurts anything to have it this way.

The complication for safety on the Rust side is that if the singleton hasn't yet been imported then we need to hold the GIL to safely import. This could be done behind the scenes in a `Deref` impl, however it seems better for users' control if they explicitly pass the GIL in.

Thanks for all the reviews, proceeding to merge this.

src/types/datetime.rs
```diff
@@ -545,12 +545,51 @@ fn opt_to_pyobj(py: Python<'_>, opt: Option<&PyObject>) -> *mut ffi::PyObject {
...
+            );
 
+            {
+                let dt = PyDateTime::from_timestamp(py, 100.0, Some(timezone_utc(py))).unwrap();
+                py_run!(
+                py,
+                dt,
+                "import datetime; assert dt == datetime.datetime.fromtimestamp(100, datetime.timezone.utc)"
+            );
+            }
```
pickfire, line 571: I think your indentation here is off.

davidhewitt: Thanks, #2508 to fix

Issue #1587, opened 2021-04-30
Proposal: use a `docsrs`  rustdocs flag


mejrs:
This is inspired by an user that had difficulty finding out about one of the optional features (num-bigint) the other day. Tokio, for example, [communicates these things very well](https://docs.rs/tokio/1.5.0/tokio/#modules), and I think we should too. This is how they do it:

Doc features like [doc_cfg](https://doc.rust-lang.org/unstable-book/language-features/doc-cfg.html) and [doc_alias](https://blog.rust-lang.org/2020/11/19/Rust-1.48.html#adding-search-aliases) can be gated behind this flag, so these only affect anything when rendering documentation. 

These features allow for some really useful things:

- Add redirects for search terms that don't appear in the documentation on their own:

![afbeelding](https://user-images.githubusercontent.com/59372212/116694466-9e9eff80-a9bf-11eb-9d04-1fe00ee7c5d8.png)

- Clearly show when things are not available on certain platforms:

![afbeelding](https://user-images.githubusercontent.com/59372212/116694799-03f2f080-a9c0-11eb-924d-db11735d7591.png)
![afbeelding](https://user-images.githubusercontent.com/59372212/116696381-1ff79180-a9c2-11eb-8f59-bf92a841eaeb.png)

- Clearly show when something is gated behind an optional feature:

![afbeelding](https://user-images.githubusercontent.com/59372212/116696111-c1caae80-a9c1-11eb-8887-5bb5dfa25267.png)

There are also some things that need to be done that I'm not 100% sure about how to do:

- Figure out the optimal set of features to compile the documentation with (`--all-features` won't work) .

- Add metadata for rustdoc to `cargo.toml`, probably something like this:
```
[package.metadata.docs.rs]
all-features = false
features = ["macros","num-bigint","num-complex","hashbrown","serde","nightly"]
rustdoc-args = ["--cfg", "docsrs"]
```

- Change the CI to do something like this:
```
$env:RUSTDOCFLAGS="--cfg docsrs"

cargo +nightly doc --no-deps --features=macros,num-bigint,num-complex,hashbrown,serde,nightly
```

Otherwise, the scope of the changes looks like this: https://github.com/mejrs/pyo3/commit/22958bdc5ae9cf22386fbdac2a5d73d11d18cc30 . It would also require putting these attributes everywhere relevant.

davidhewitt:
[...]

PR #1590, opened 2021-05-02
docsrs feature

mejrs:
As per #1587.

Adds: 
 - new workflow in ci that builds docs with the flag on
 - flags in the lib.rs of pyo3, macros and macros-backend
 - tags on a couple of items

This is more of a test to see whether the workflow plays ball, I plan to add more tags where applicable.


davidhewitt:
👍 do you want me to review this yet? unsure when things are marked WIP / draft whether my input is needed.

mejrs:
Right now I'm in the middle of putting these annotations where needed. I figured it was best to do these things in different commits.

Can you review the change to ci.yml? That is really the only substantial change so far.

.github/workflows/ci.yml
```diff
@@ -173,6 +173,23 @@ jobs:
...
+    needs: [fmt]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+         toolchain: nightly
+         override: true
+      - name: "Build docs with docsrs flag"
+        run: cargo doc --no-deps --features=default,multiple-pymethods,extension-module,num-bigint,num-complex,hashbrown,serde,nightly
```
davidhewitt, line 188: I think let's make this string match the others in the file, so that it's easy to correct later with find-and-replace when features change.

```suggestion
        run: cargo doc --no-deps --no-default-features --features "macros num-bigint num-complex hashbrown serde multiple-pymethods"
```

The features I took out (`extension-module`, `nightly`) don't affect visible APIs so no need to pass them here.

davidhewitt (commented):
Sure thing, seems reasonable with just one suggestion.

.github/workflows/ci.yml
mejrs: `nightly` exposes some conversions that play with the specialization feature which *are* part of the public api (if the nightly flag is on). Or is it the case that you just don't want to expose them as such?

mejrs:
I'm not sure how resolving actually works 😞

mejrs:
I think this is good :)

mejrs:
@davidhewitt can you merge this? 💖

messense (approved):
LGTM.

@davidhewitt I think we should also build docs with `docsrs` feature for the docs of the main branch? https://pyo3.rs/main/doc/pyo3/index.html

https://github.com/PyO3/pyo3/blob/350e7b2be743358edc07a0ca320c50461b97f3c8/.github/workflows/guide.yml#L38-L43

messense:
I took the liberty to change `guide` workflow to also build with `--cfg docsrs` and make it run on pull request. Now I think we can remove the `docs` job from `ci` workflow.

.github/workflows/guide.yml
```diff
@@ -38,11 +44,12 @@ jobs:
         # This adds the docs to gh-pages-build/doc
       - name: Build the doc
         run: |
-          cargo doc --features="default num-bigint num-complex" --no-deps
+          cargo +nightly rustdoc --lib --features="macros num-bigint num-complex hashbrown serde multiple-pymethods" -- --cfg docsrs
           cp -r target/doc gh-pages-build/doc
           echo "<meta http-equiv=refresh content=0;url=pyo3/index.html>" > gh-pages-build/doc/index.html
 
       - name: Deploy
+        if: github.ref == 'refs/heads/main'
```
davidhewitt, line 52: We also need to deploy if it's a tag e.g. `v0.14.0`

.github/workflows/guide.yml
```diff
@@ -38,11 +44,12 @@ jobs:
         # This adds the docs to gh-pages-build/doc
       - name: Build the doc
         run: |
-          cargo doc --features="default num-bigint num-complex" --no-deps
+          cargo +nightly rustdoc --lib --features="macros num-bigint num-complex hashbrown serde multiple-pymethods" -- --cfg docsrs
```
davidhewitt, line 47: `--no-default-features` also, maybe?

davidhewitt (approved):
Thanks both. Docs changes look great. 💯 


@messense just a couple of suggestions on your `gh-pages` ci changes. I agree we don't need the new docs job if we reuse gh-pages; that's a good idea.

PR #1591, opened 2021-05-03
pyclass: support extending Exception types

davidhewitt:
Implements support for `#[pyclass(extends=Exception)]`.

I expected there to be bigger issues around doing this; it turned out to be very easy.

cc @Stranger6667 who was asking about this on Gitter just the other day.

### DIFF
--- a/tests/test_inheritance.rs
+++ b/tests/test_inheritance.rs
@@ -153,7 +153,8 @@ except Exception as e:
 #[cfg(not(Py_LIMITED_API))]
 mod inheriting_native_type {
     use super::*;
-    use pyo3::types::{PyDict, PySet};
+    use pyo3::exceptions::PyException;
+    use pyo3::types::{IntoPyDict, PyDict, PySet};
 
     #[pyclass(extends=PySet)]
     #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


tests/test_inheritance.rs
```diff
@@ -208,6 +209,38 @@ mod inheriting_native_type {
...
+        }
+    }
+
+    #[test]
+    fn custom_exception() {
+        Python::with_gil(|py| {
+            let cls = py.get_type::<CustomException>();
+            let dict = [("cls", cls)].into_py_dict(py);
+            let res = py.run(
+            "e = cls('hello'); assert str(e) == 'hello'; assert e.context == 'Hello :)'; raise e",
```
birkenfeld, line 236: I guess it's a bit redundant, but do you also want to check that the exception can be caught in Python code?

davidhewitt: I don't see the harm in doing so; it's more comprehensive adding it 👍

ravenexp:
Does this feature make `create_exception!()` obsolete?

davidhewitt:
In the long run, yes. But that'll come when we have a full `#[pyexception]` macro. I'll try to write the implementation still needed on #295 soon.

PR #1592, opened 2021-05-03
types: clean up types macros

davidhewitt:
Related to #1591, #1588.

I've been looking at the types macros just now, which are getting a bit more complicated again (especially if we customize `Deref` implementations to model base types as per #1588).

This PR makes two simplifications:
 - optional arguments to the macros (`checkfunction`, `module`, soon-to-be `deref`) have an explicit syntax `#checkfunction=...` so that the macros can parse unambiguously.
 - I've removed `pyobject_native_var_type!` - just use `pyobject_native_type_core!` if the layout is not suitable for inheriting, and `pyobject_native_type!` if the layout is suitable for inheriting.

### DIFF
--- a/src/exceptions.rs
+++ b/src/exceptions.rs
@@ -2,7 +2,6 @@
 
 //! Exception types defined by Python.
 
-use crate::type_object::PySizedLayout;
 use crate::{ffi, PyResult, Python};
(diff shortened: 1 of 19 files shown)


messense (approved):
Looks great!

PR #1594, opened 2021-05-03
docs: use white-space style instead of &nbsp; blob

davidhewitt:
cc @mejrs


mejrs:
Thanks 👍

I have no idea why I didn't think of doing that...

Issue #1344, opened 2020-12-27
Support subclassing from native types in abi3


davidhewitt:
Currently unsupported, because it depends on the base type layout. There should be ways to achieve this.

davidhewitt:
Note to self: I think that this might be possible on Python 3.10, where `PyType_GetSlot` can be safely used on native types to get `tp_new` etc (which is necessary during class instantiation).

ppmpreetham:
is this still being worked on?

davidhewitt:
It's still relevant but nobody has been able to prioritise it. #4798 is probably going to help.

PR #1596, opened 2021-05-04
type_object: remove layout and base type from PyTypeInfo

davidhewitt:
This PR removes some associated types from `PyTypeInfo` which are really just implementation details of `#[pyclass]`:

- `Layout`: moved to `PyClassImpl`. This doesn't make sense in the long term anyway, because if we inherit from native types where we don't know the layout information (#991, #1344) we need `#[pyclass]` to work without layout information.
- `BaseType` - moved to `PyClassImpl`. Not all Python types have a base type (e.g. `PyAny`), or we might not statically know them (e.g. #991). All `#[pyclass]` do have a base type and we'll always know what it is.
- `BaseLayout` - removed entirely. 
- `Intitializer` - moved to `PyClassBaseType` (which is renamed from `PyBaseTypeUtils`).

Making this change frees us up to make internal implementation changes to eventually support #991 and #1344.

In addition, it also simplifies a lot of trait bounds dotted around the public interface and simplifies the macros in `types/mod.rs`.


messense (approved):
LGTM.

davidhewitt:
(I'm going to add some more tests before merging)

PR #1605, opened 2021-05-14
Add support for cross compilation for wasm/wasi

simonrainerson:
When using rust after this commit
https://github.com/rust-lang/rust/commit/1a491e2304387c2a50c5a0ae219e8d5a0ce1bc89
the target family for wasm targets is now set to "wasm". Also fix the
assumption that OsStrExt lives inside std::os::unix for all non-windows
targets.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -536,6 +536,8 @@ fn load_cross_compile_info(cross_compile_config: CrossCompileConfig) -> Result<I
         Some(os) if os == "unix" => load_cross_compile_from_sysconfigdata(cross_compile_config),
         // Use hardcoded interpreter config when targeting Windows
         Some(os) if os == "windows" => windows_hardcoded_cross_compile(cross_compile_config),
+        // sysconfigdata works fine on wasm/wasi
+        Some(os) if os == "wasm" => load_cross_compile_from_sysconfigdata(cross_compile_config),
         // Waiting for users to tell us what they expect on their target platform
         Some(os) => bail!(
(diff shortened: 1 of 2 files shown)


messense (approved):
LGTM

davidhewitt:
👍 This looks reasonable to me too.

Note for the folks using Wasm: I'm happy to have patches like this to fixup support where it's needed, but I don't think we can declare support for Wasm until someone with knowledge about how the current state of the Wasm ecosystem contributes a comprehensive CI job.

(Without CI, we could unknowingly break Wasm at any moment.)

gideongrinberg:
> 👍 This looks reasonable to me too.
> 
> Note for the folks using Wasm: I'm happy to have patches like this to fixup support where it's needed, but I don't think we can declare support for Wasm until someone with knowledge about how the current state of the Wasm ecosystem contributes a comprehensive CI job.
> 
> (Without CI, we could unknowingly break Wasm at any moment.)

Isn't adding CI for WASM just a matter of running the tests in a headless browser environment, or would it require new tests?

davidhewitt:
> Isn't adding CI for WASM just a matter of running the tests in a headless browser environment

It's pretty much just this, but there's two key questions I don't have the answer to:
- Which wasm runtime to use? (Node? Wasmer? Headless chrome?)
- How to obtain a Python distribution compiled to Wasm?

Anyone who knows a good answer to these is welcome to contribute CI so that we can add official support.

PR #1608, opened 2021-05-15
Reduce function calling overhead by ~25% in ReferencePool::update_counts:

alex:
1) Place both increfs and decrefs behind a single mutex, rather than two. Even uncontended mutexes aren't free to acquire.
2) Keep a dirty tracking bool to avoid acquiring any mutexes at all in the common case of no modifications (because everything happened with the GIL held)

# Benchmarks

## Before

```
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 63.6 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 63.7 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 64.1 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 63.6 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 63.5 nsec per loop
```

## With single lock

```
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 51.9 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 52 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 52.9 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 51.8 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 54.7 nsec per loop
```

## With both single lock and dirty flag

```
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 47.2 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 46.7 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 46.6 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 46.6 nsec per loop
(.venv) alex@media-01:~/p/pyo3$ python3 -mtimeit -s "from pyo3_benchmarks import no_args" "no_args()"
5000000 loops, best of 5: 47.1 nsec per loop
```

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -5,7 +5,11 @@
 use crate::{ffi, internal_tricks::Unsendable, Python};
 use parking_lot::{const_mutex, Mutex, Once};
 use std::cell::{Cell, RefCell};
-use std::{mem::ManuallyDrop, ptr::NonNull};
+use std::{
+    mem::{self, ManuallyDrop},
+    ptr::NonNull,
+    sync::atomic,
+};
 
 static START: Once = Once::new();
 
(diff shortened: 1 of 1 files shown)


src/gil.rs
```diff
@@ -287,50 +287,60 @@ impl Drop for GILGuard {
...
         // Always increase reference counts first - as otherwise objects which have a
         // nonzero total reference count might be incorrectly dropped by Python during
         // this update.
-        for ptr in swap_vec_with_lock!(self.pointers_to_incref) {
+        for ptr in swap_vec!(ops.0) {
             unsafe { ffi::Py_INCREF(ptr.as_ptr()) };
         }
 
-        for ptr in swap_vec_with_lock!(self.pointers_to_decref) {
+        for ptr in swap_vec!(ops.1) {
```
davidhewitt, line 343: We need to be careful here. Decrementing references can run arbitrary Python drop code, which can lead to GIL release which may eventually deadlock because we hold the lock here. (I think this mutex is also non reentrant so may deadlock on a single thread.)

After swapping `ops.1` we should release the lock before decreasing any references.

src/gil.rs
```diff
@@ -287,50 +287,60 @@ impl Drop for GILGuard {
...
 
     fn register_decref(&self, obj: NonNull<ffi::PyObject>) {
-        self.pointers_to_decref.lock().push(obj)
+        self.pointer_ops.lock().1.push(obj);
+        self.dirty.store(true, atomic::Ordering::Release);
     }
 
     fn update_counts(&self, _py: Python) {
-        macro_rules! swap_vec_with_lock {
+        macro_rules! swap_vec {
```
davidhewitt, line 319: Now that this macro doesn't do any locking, I suggest throwing it away and just using `mem::take`.

davidhewitt (changes requested):
Awesome! These are some really great optimizations which many people will be happy with. I think there's a potential deadlock though (which we can avoid), see comment below.

src/gil.rs
alex: Good call, should be good now.

davidhewitt (approved):
Looks great! Thanks very much 😌

kngwyu:
Looks like I was late for the party... Very interesting to see this optimization works, thanks.

PR #1614, opened 2021-05-17
Fix clippy ignore directives

birkenfeld:
My nightly compiler complained about these (not stable though), but both stable and nightly seem to be happy with the new version...


davidhewitt (approved):
LGTM!

PR #1617, opened 2021-05-19
pyo3_benchmarks: add a benchmark for the "only simple args" case

birkenfeld:
This is the case that will be helped by fastcall support,
so add it first as a baseline.

Currently, this benchmark takes about 2x time compared to the Python version.

### DIFF
--- a/examples/pyo3-benchmarks/src/lib.rs
+++ b/examples/pyo3-benchmarks/src/lib.rs
@@ -24,10 +24,16 @@ fn mixed_args<'a>(
 #[pyfunction]
 fn no_args() {}
 
+#[pyfunction(b = 2, "*", c = 4)]
+fn simple_args(a: i32, b: i32, c: i32) -> (i32, i32, i32) {
+    (a, b, c)
+}
+
 #[pymodule]
 fn _pyo3_benchmarks(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
     m.add_function(wrap_pyfunction!(args_and_kwargs, m)?)?;
     m.add_function(wrap_pyfunction!(mixed_args, m)?)?;
     m.add_function(wrap_pyfunction!(no_args, m)?)?;
+    m.add_function(wrap_pyfunction!(simple_args, m)?)?;
     Ok(())
 }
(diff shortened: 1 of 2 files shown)


birkenfeld:
Note: I plan to implement fastcall only for cases where there are no `*args` or `**kwds` catch-all arguments.  In these cases, we would have to create a tuple/dictionary anyway so we might as well use the one created by CPython.

In addition, it greatly simplifies the implementation of the fastcall extraction.

davidhewitt (approved):
LGTM!

> Note: I plan to implement fastcall only for cases where there are no *args or **kwds catch-all arguments. In these cases, we would have to create a tuple/dictionary anyway so we might as well use the one created by CPython.

That seems reasonable to me. We can always experiment with optimizing the catch-all cases later and see if it is worth the additional complexity!

(In particular we could potentially allow catch-all args to be extracted into `Vec` and `HashMap` in the future, which could be faster than making Python structures for certain use cases.)

davidhewitt:
Btw, I was sort of hoping that the argument extraction shouldn't be too bad: my last refactor to `FunctionDescription::extract_arguments` was done with the hope that it could be retrofitted to use fastcall args. I think the main changes that would be needed would be how the args and kwargs are split up into appropriate iterators. It might be possible to make it generic on `args: impl IntoIterator<Item = &PyAny>` and `kwargs: impl IntoIterator<Item = (&PyAny, &PyAny)>` if you're able to preprocess the fastcall input a little.

birkenfeld:
> Btw, I was sort of hoping that the argument extraction shouldn't be too bad: my last refactor to FunctionDescription::extract_arguments was done with the hope that it could be retrofitted to use fastcall args. I think the main changes that would be needed would be how the args and kwargs are split up into appropriate iterators. It might be possible to make it generic on args: impl IntoIterator<Item = &PyAny> and kwargs: impl IntoIterator<Item = (&PyAny, &PyAny)> if you're able to preprocess the fastcall input a little.

Ok, I haven't got to the adaptation of `extract_arguments` yet.  That sounds very attractive though, so I'll keep generality in mind.

PR #1620, opened 2021-05-19
pyany documentation

mejrs:
Expands on what can/can't be done with PyAny and adds examples to the compare methods

### DIFF
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -23,15 +23,13 @@ use std::os::raw::c_int;
 /// therefore its API does not require a [`Python<'py>`](crate::Python) token.
 /// - It can't be used in situations where the GIL is temporarily released,
 /// such as [`Python::allow_threads`](crate::Python::allow_threads)'s closure.
-/// - It can be mutated through any reference.
+/// - The underlying Python object, if mutable, can be mutated through any reference.
 /// - It can be converted to the GIL-independent [`Py`]`<`[`PyAny`]`>`,
 /// allowing it to outlive the GIL scope. However, using [`Py`]`<`[`PyAny`]`>`'s API
 /// *does* require a [`Python<'py>`](crate::Python) token.
 ///
-/// While basic functionality for interacting with Python objects is available on `PyAny`,
-/// doing useful things with it often requires downcasting `PyAny` to a concrete type.
-/// This can be done with [`PyAny::downcast`] (for native Python types only) and [`FromPyObject::extract`]. See their
-/// documentation for more information.
+/// It can be cast to a concrete type with PyAny::downcast (for native Python types only)
+/// and FromPyObject::extract. See their documentation for more information.
 ///
 /// See [the guide](https://pyo3.rs/main/types.html) for an explanation
 /// of the different Python object types.
@@ -179,12 +177,12 @@ impl PyAny {
     ///
     /// ```rust
     /// use pyo3::prelude::*;
-    /// use pyo3::types::PyFloat;
+    /// use pyo3::types::{PyFloat, PyString};
     ///
     /// # fn main() -> PyResult<()> {
     /// Python::with_gil(|py| -> PyResult<()> {
-    ///     let a = PyFloat::new(py, std::f64::NAN);
-    ///     let b = PyFloat::new(py, 100_f64);
+    ///     let a = PyFloat::new(py, 0_f64);
+    ///     let b = PyString::new(py, "zero");
     ///     assert!(a.compare(b).is_err());
     ///     Ok(())
     /// })?;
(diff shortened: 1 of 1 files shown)


src/types/any.rs
```diff
@@ -11,31 +11,30 @@ use std::cell::UnsafeCell;
...
-/// `PyAny` is used as a reference with a lifetime that represents that the GIL
-/// is held, therefore its API does not require a `Python<'py>` token.
+/// `PyAny` has some interesting properties, which it shares
+/// with the other [native Python types](crate::types):
 ///
-/// See [the guide](https://pyo3.rs/main/types.html) for an explanation
-/// of the different Python object types.
+/// - It can only be obtained and used while the GIL is held,
+/// therefore its API does not require a [`Python<'py>`](crate::Python) token.
+/// - It can't be used in situations where the GIL is temporarily released,
```
birkenfeld, line 24: What do you mean here?

src/types/any.rs
```diff
@@ -11,31 +11,30 @@ use std::cell::UnsafeCell;
...
+/// such as [`Python::allow_threads`](crate::Python::allow_threads)'s closure.
+/// - It can be mutated through any reference.
```
birkenfeld, line 26: This is not unique to `PyAny`, and "it can be mutated" only holds for Python types that actually *can* be mutated.

I would either remove or expand this.

src/types/any.rs
```diff
@@ -11,31 +11,30 @@ use std::cell::UnsafeCell;
...
+/// - It can be converted to the GIL-independent [`Py`]`<`[`PyAny`]`>`,
+/// allowing it to outlive the GIL scope. However, using [`Py`]`<`[`PyAny`]`>`'s API
+/// *does* require a [`Python<'py>`](crate::Python) token.
 ///
-/// # Examples
+/// While basic functionality for interacting with Python objects is available on `PyAny`,
+/// doing useful things with it often requires downcasting `PyAny` to a concrete type.
```
birkenfeld, line 32: That is a bit misleading - you can do a whole lot with the abstract object API.

src/types/any.rs
```diff
@@ -143,6 +157,39 @@ impl PyAny {
...
+    /// Python::with_gil(|py| -> PyResult<()> {
+    ///     let a = PyFloat::new(py, 0_f64);
+    ///     let b = PyFloat::new(py, 42_f64);
+    ///     assert_eq!(a.compare(b)?, Ordering::Less);
+    ///     Ok(())
+    /// })?;
+    /// # Ok(())}
+    /// ```
+    ///
+    /// It will return `PyErr` for values that cannot be compared:
```
birkenfeld, line 178: With floats, this is due to neither of the three comparisons returning True, which is already shown in the equivalent code above.

Can you make an example that involves values that actually cannot be compare, like `int < str`?

src/types/any.rs
```diff
@@ -169,16 +216,40 @@ impl PyAny {
...
-    ///   * CompareOp::Eq: `self == other`
-    ///   * CompareOp::Ne: `self != other`
-    ///   * CompareOp::Lt: `self < other`
-    ///   * CompareOp::Le: `self <= other`
-    ///   * CompareOp::Gt: `self > other`
-    ///   * CompareOp::Ge: `self >= other`
+    ///
+    /// <div style="width:1px">
+    ///
+    /// | `compare_op` | <span style="white-space: pre">Python expression</span> |
```
birkenfeld, line 226: Why is that `span` needed?

birkenfeld (changes requested):

src/types/any.rs
mejrs: Combined with the div that's 1px, it shrinks the table as much as possible. Otherwise it expands the table as big as it can get, which I think looks odd.

src/types/any.rs
mejrs: 👍

src/types/any.rs
mejrs: I'm not sure what is unclear here. You can't get or use it if you don't have the GIL.

src/types/any.rs
mejrs: Hmm, that's true, that line isn't really that useful. What about gutting it entirely:

```It can be cast to a concrete type with PyAny::downcast (for native Python types only) and FromPyObject::extract. See their documentation for more information.```

src/types/any.rs
birkenfeld: You're right, scratch that.

src/types/any.rs
birkenfeld: Or just `The underlying Python object, if mutable, can be mutated through any reference.`?

src/types/any.rs
birkenfeld: :+1:

src/types/any.rs
mejrs: Perfect

birkenfeld (approved):

davidhewitt (approved):
LGTM, thanks again for the authoring and reviewing folks!

PR #1621, opened 2021-05-20
Remove overly specific advice in `callable` docstring

birkenfeld:
The "don't use unwrap()" applies to any operation on Python API,
and seems out of place here specifically.

### DIFF
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -120,7 +120,7 @@ impl PyAny {
 
     /// Deletes an attribute.
     ///
-    /// This is equivalent to the Python expression `del self.attr_name`.
+    /// This is equivalent to the Python statement `del self.attr_name`.
     pub fn delattr<N>(&self, attr_name: N) -> PyResult<()>
     where
         N: ToPyObject,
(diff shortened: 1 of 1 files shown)


mejrs:
This example was never about "don't use unwrap" in my eyes; it's to illustrate that it's an antipattern (that people like to reach for for the wrong reasons).

Let me draw a parallel to something that people tend to get wrong:

```python
if os.path.isfile("foo.txt"):
	file = open("foo.txt")
```

This is bad for some reasons: 

- a file can exist but fail to open (for example, a permission error) and thus still raise exceptions
- you still can't rely on it because this is racy (i.e., something else could delete the file between checking and opening). Any code like this is, strictly speaking, incorrect.

Using `callable()` is bad for similar reasons:

- checking whether something is callable before you call it is mostly useless
- it doesn't save you from having to deal with exceptions anyway

I admit it's not a great example, but I feel *something* like "if you are looking for this function it's probably not what you should be using; use x instead" is appropriate.

birkenfeld:
OK, I get that, but the example is a "strawman" since even after checking with `callable()` you wouldn't call it and `unwrap()` the result, since the called function itself could still raise anything.

How about just saying "unless your API needs to distinguish between callable and non-callable objects, there is no point in using this instead of trying to call directly"?

mejrs:
Sounds good 👍

davidhewitt (approved):

PR #1626, opened 2021-05-24
build.rs: if found more than one candidate, filter on arch

alonblade:
Hi,

 This is my first RP on pyo3. Thanks for the great work, I'm using this right now to convert a library from python to rust for a raspberry pi running in 32 bit mode, that is armhf architecture, using ubuntu 20.04 as the target distribution. I have both the host x86_64 python3.8 sysconfigdata and the foreign architecture armhf in /usr/lib/python3.8, which results in more than one candidate during search_lib_dir, and an error of "Detected multiple possible python versions".

 I've conservatively added a filter in that case to keep only the paths with cross.arch in their name (to_lossy_string.contains(&config.arch))

 There is already an existing filter on config.arch but only in another branch, so I didn't want to pull it in to the first branch (Ok(f) if starts_with(f, "_sysconfigdata") && ends_with(f, "py")).

 I added a line in CHANGELOG, I think this is a bug so it should not change the documentation.

 I've run clippy (no errors), run fmt (no changes), "cargo test" fails but on unrelated code (I can attach, fails in examples).

Alon

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -484,6 +484,24 @@ fn search_lib_dir(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Vec<Pat
         };
         sysconfig_paths.extend(sysc);
     }
+    // If we got more than one file, only take those that contain the arch name.
+    // For ubuntu 20.04 with host architecture x86_64 and a foreign architecture of armhf
+    // this reduces the number of candidates to 1:
+    //
+    // $ find /usr/lib/python3.8/ -name '_sysconfigdata*.py' -not -lname '*'
+    //  /usr/lib/python3.8/_sysconfigdata__x86_64-linux-gnu.py
+    //  /usr/lib/python3.8/_sysconfigdata__arm-linux-gnueabihf.py
+    if sysconfig_paths.len() > 1 {
+        let temp = sysconfig_paths
+            .iter()
+            .filter(|p| p.to_string_lossy().contains(&cross.arch))
+            .cloned()
+            .collect::<Vec<PathBuf>>();
+        if !temp.is_empty() {
+            sysconfig_paths = temp;
+        }
+    }
+
     sysconfig_paths
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Update `num-complex` optional dependency to 0.4. [#1482](https://github.com/PyO3/pyo3/pull/1482)
 - Extend `hashbrown` optional dependency supported versions to include 0.11. [#1496](https://github.com/PyO3/pyo3/pull/1496)
 - Support PyPy 3.7. [#1538](https://github.com/PyO3/pyo3/pull/1538)
+- Try harder to filter sysconfigdata candidates on arch.config
 
 ### Added
 - Add conversions for `[T; N]` for all `N` on Rust 1.51 and up. [#1128](https://github.com/PyO3/pyo3/pull/1128)


pyo3-build-config/src/impl_.rs
```diff
@@ -484,6 +484,20 @@ fn search_lib_dir(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Vec<Pat
...
+    // $ find /usr/lib/python3.8/ -name '_sysconfigdata*.py' -not -lname '*'
+    //  /usr/lib/python3.8/_sysconfigdata__x86_64-linux-gnu.py
+    //  /usr/lib/python3.8/_sysconfigdata__arm-linux-gnueabihf.py
+    if sysconfig_paths.len() > 1 {
+        sysconfig_paths = sysconfig_paths
+            .iter()
+            .filter(|p| p.to_string_lossy().contains(&cross.arch))
+            .map(|p| p.clone())
+            .collect::<Vec<PathBuf>>();
+    }
```
davidhewitt, line 500: The only potential issue I see is that if the `arch` doesn't match _any_ sysconfig file (e.g. maybe the arch is spelt differently for some reason?) then this could reduce the number of candidates to zero.

In that case, it might be worth keeping the original full set of paths instead of listing zero paths and failing?

davidhewitt (commented):
Hi @alonblade, thanks for the PR!

We literally just merged a change to the build infrastructure, so I had to rebase and force-push your PR. Hope that's ok.

Let's see how CI does with this branch, also I have one thought...

alonblade:
ok, I made a mess by accidentally pushing with force my updated change, so
now I'm redoing the rebase you did for me.

alonblade:
@davidhewitt hmm, now that I rebased on latest I get a different error from my use case (pi armhf target, x86_64 host) - Trying to debug it.

```
   Compiling pyo3 v0.13.2 (/home/alon/catkin_ws/mpu6050/sub/pyo3)
error: failed to run custom build command for `pyo3 v0.13.2 (/home/alon/catkin_ws/mpu6050/sub/pyo3)`

Caused by:
  process didn't exit successfully: `/home/alon/catkin_ws/mpu6050/target/release/build/pyo3-6c31d3e862b113d6/build-script-build` (exit status: 1)
  --- stderr
  error: Your Rust target architecture (32-bit) does not match your python interpreter (64-bit)
```

davidhewitt:
@alonblade yikes, I'm pretty sure this is my fault; I have a guess at what's going on.

I'll try to open a PR to fix tonight.

alonblade:
@davidhewitt ok, I think I found the problem: the new refactoring of pyo3-build-config out of build.rs means that when
pyo3-build-config/src/impl_.rs is run the value of the cargo target environment variables is actually the host, not the real target. So cross.arch is actually x86_64, and I get a false error. To confirm, I commented out ensure_target_architecture and tested the resulting binary and it is the correct architecture.

I am not sure how to proceed - one option would be to pass the real target variables to pyo3-build-config in a different set of environment variables (i.e. CARGO_CFG_TARGET_ARCH becomes PYO3_CARGO_CFG_TARGET_ARCH)

What do you think?

alonblade:
> @alonblade yikes, I'm pretty sure this is my fault; I have a guess at what's going on.
> 
> I'll try to open a PR to fix tonight.

Thanks!

davidhewitt:
(Apologies, I've run out of time to try to fix this tonight. I'm going to attempt to find time again soon, hopefully on Thursday evening.)

alonblade:
Sweet. No worries.

davidhewitt:
I'm going to merge this and rebase my cross-compilation fix on it. Thanks for implementing this!

alonblade:
Super, thanks!

PR #1630, opened 2021-05-25
pymodule: tidy up module init

davidhewitt:
Three small tweaks: 
  - Remove `PyModuleDef_INIT` symbol.
  - Set module `__doc__` via `m_doc` def member rather than with a `setattr()` call.
  - Don't bother calling `ffi::PyEval_InitThreads()` - as far as I can tell this isn't necessary (even on Python 3.6).

### DIFF
--- a/src/derive_utils.rs
+++ b/src/derive_utils.rs
@@ -9,7 +9,7 @@ use crate::exceptions::PyTypeError;
 use crate::instance::PyNativeType;
 use crate::pyclass::PyClass;
 use crate::types::{PyAny, PyDict, PyModule, PyString, PyTuple};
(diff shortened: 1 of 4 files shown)


src/derive_utils.rs
```diff
@@ -312,14 +321,8 @@ impl ModuleDef {
         #[cfg(not(Py_3_7))]
         ffi::PyEval_InitThreads();
```
messense, line 322: Missing a `unsafe` block here.

davidhewitt: Actually I don't think this call to `PyEval_InitThreads` is needed at all. Pushed a commit to remove it, and updated the OP.

messense (approved):
LGTM

Issue #1632, opened 2021-05-25
`async fn` tracking issue


davidhewitt:
This issue is a placeholder issue which I'd like to use to keep track of what we need to do to support `async fn` for `#[pyfunction]` (and `#[pymethods]`).

While this would be a great feature to have, I think anything which we merge into PyO3 core should be simple to use, performant, and (ideally) async-runtime agnostic.

Users have started researching the design space of how possible cross-language implementations could be written. The most advanced candidate is [`pyo3-async-runtimes`](https://github.com/PyO3/pyo3-async-runtimes), (formerly `pyo3-asyncio`). See the guide as a great example of how to use it: https://pyo3.rs/latest/ecosystem/async-await.html

At the moment I suggest we allow some time for that and other crates to mature, and once we understand performance trade-offs / difficulties etc. we can consider upstreaming something here.

As this is supported by third-party crates I think we can afford patience for now.

awestlake87:
`pyo3-asyncio` still needs some time and development to mature. There are currently a few things that I think need to be addressed:

- Documentation and testing for `pyo3-asyncio` within native modules - seems like many users need this to get started
- Cleaner initialization - better asyncio configuration, potentially some `AlreadyInitialized` errors, etc.
- Stream <-> async generator conversions - my first draft had some performance concerns
- Cleaner error handling - no complaints yet, but I have a few nitpicks here and there

I've been taking a bit of a passive approach to this library lately to see what users need it for, and it seems like these are the main concerns right now.

JohnScience:
For the time being, using `async` conditionally can be enabled with [`qualifier_attr`](https://github.com/JohnScience/qualifier_attr).

wyfo:
I've just released [pyo3-async](https://github.com/wyfo/pyo3-async) (see [reddit](https://www.reddit.com/r/rust/comments/16y0yi9/pyo3async_new_bindings_to_various_python/)). 
Its approach is completely different from pyo3-asyncio as it is runtime-agnostic on Rust side. I've also added the support for [trio](https://github.com/python-trio/trio), as it's quite popular (and supporting it was simple enough, contrary to curio for example). Last but not least, it supports async generators (simply reusing its coroutine implementation for `__anext__`), and cancellation.

Adding a macro for `async fn` is on my TODO list, but I think I have to discuss about it here first, whether or not the implementation could be good enough to be quoted in the documentation.

wyfo:
[...]

PR #1633, opened 2021-05-25
pyfunction: better error message with `async fn`

davidhewitt:
This PR adds a better error message when users try to use `async fn`, which points them to the new tracking issue #1632. We can update that issue over time to keep useful up-to-date information on thirdparty options like `pyo3-asyncio`.


Eric-Arellano (approved):
I love this attention to detail! This is really useful and user-friendly.

davidhewitt:
😅 proc macros are awesome, but the error messages users get when proc macros expand to non-functional code are usually just gibberish! At least in this case we can catch this one early and give them something better to read.

messense (approved):
Nice!

Issue #1634, opened 2021-05-26
Automated continuous benchmarking


messense:
https://github.com/marketplace/actions/continuous-benchmark

Tokio has an example workflow: https://github.com/tokio-rs/tokio/blob/master/.github/workflows/bench.yml and tracking benchmark result with GH Pages: https://tokio-rs.github.io/tokio/dev/bench/

I can send a PR if this is desirable.

davidhewitt:
Personally I would love to see this! I was actually talking a look at that exact action the other day!

PR #1636, opened 2021-05-26
Migrate all benchmarks to criterion

messense:


davidhewitt:
I guess this is part of #1634 ? If you think it's necessary, I'm happy with this. Overall, I don't love the clutter which criterion adds versus the built-in benchmark runner, but at least it enables benchmarking on stable...

PR #1638, opened 2021-05-26
Fix memory leak when converting to u128 and i128

erikjohnston:
I had a quick look at the equivalent `slow_128bit_int_conversion` conversion and I don't think there is a leak there, but I haven't checked.


birkenfeld (approved):
LGTM, nice find!

PR #1643, opened 2021-05-29
 improve lib.rs documentation

mejrs:
- rewrites the ownership and lifetimes section into more of "this is a very basic overview of the important types"
- adds a list of supported features
- adds a description and link to the pyo3-build-config crate
- also adds a paragraph about the supported versions
- deletes the bit about renaming files (this was pretty confusing for me back in the day), which @davidhewitt said he'd like to move to the guide anyway

On another note - which of the modules and macros shown listed on the index page are implementation details? I'd like to follow up with a commit to `#[doc(hidden)]` the ones that are.

<details><summary>rendered</summary>
<p>

image

image

image

</p>
</details>

### DIFF
--- a/guide/src/building_and_distribution.md
+++ b/guide/src/building_and_distribution.md
@@ -35,9 +35,16 @@ pyo3 = { {{#PYO3_CRATE_VERSION}} }
 
 On Linux/macOS you might have to change `LD_LIBRARY_PATH` to include libpython, while on windows you might need to set `LIB` to include `pythonxy.lib` (where x and y are major and minor version), which is normally either in the `libs` or `Lib` folder of a Python installation.
 
-## Distribution
+## Testing, building and distribution
 
-There are two ways to distribute your module as a Python package: [setuptools-rust] and [maturin]. setuptools-rust needs several configuration files (`setup.py`, `MANIFEST.in`, `build-wheels.sh`, etc.). It allows (and sometimes requires) writing custom workflows in python. maturin has only few options and works without any additional configuration, instead it requires a rigid project structure and does not support some functionality of setuptools such as package data ([pyo3/maturin#258](https://github.com/PyO3/maturin/issues/258)), multiple extensions or running python scripts at build time.
+There are two main ways to test, build and distribute your module as a Python package: [setuptools-rust] and [maturin]. setuptools-rust needs several configuration files (`setup.py`, `MANIFEST.in`, `build-wheels.sh`, etc.). It allows (and sometimes requires) writing custom workflows in python. maturin has only few options and works without any additional configuration, instead it requires a rigid project structure and does not support some functionality of setuptools such as package data ([pyo3/maturin#258](https://github.com/PyO3/maturin/issues/258)), multiple extensions or running python scripts at build time.
+
+While developing, you can also symlink (or copy) and rename the shared library from the `target` folder: 
+- on macOS, rename `libyour_module.dylib` to `your_module.so`.
+- on Windows, rename  `libyour_module.dll` to `your_module.pyd`.
+- on Linux, rename `libyour_module.so` to `your_module.so`. 
+
+You can then open a Python shell in the same folder and you'll be able to use `import your_module`.
 
 ## `Py_LIMITED_API`/`abi3`
 
--- a/guide/src/building_and_distribution.md
+++ b/guide/src/building_and_distribution.md
@@ -39,7 +39,9 @@ On Linux/macOS you might have to change `LD_LIBRARY_PATH` to include libpython, 
 
 There are two main ways to test, build and distribute your module as a Python package: [setuptools-rust] and [maturin]. setuptools-rust needs several configuration files (`setup.py`, `MANIFEST.in`, `build-wheels.sh`, etc.). It allows (and sometimes requires) writing custom workflows in python. maturin has only few options and works without any additional configuration, instead it requires a rigid project structure and does not support some functionality of setuptools such as package data ([pyo3/maturin#258](https://github.com/PyO3/maturin/issues/258)), multiple extensions or running python scripts at build time.
 
-While developing, you can also symlink (or copy) and rename the shared library from the `target` folder: 
+### Manual builds
+
+You can also symlink (or copy) and rename the shared library from the `target` folder: 
 - on macOS, rename `libyour_module.dylib` to `your_module.so`.
 - on Windows, rename  `libyour_module.dll` to `your_module.pyd`.
 - on Linux, rename `libyour_module.so` to `your_module.so`. 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,7 @@
 
 //! Rust bindings to the Python interpreter.
 //!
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -4,41 +4,132 @@
...
+//!Use this feature when building an extension module.
+//
+//! - `hashbrown`: Enables conversions between Python objects and
+//! [hashbrown](https://docs.rs/hashbrown)'s
+//! [`HashMap`](https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html) and
+//! [`HashSet`](https://docs.rs/hashbrown/latest/hashbrown/struct.HashSet.html) types.
+//
+//! - `multiple-pymethods`: Enables the use of multiple
+//! [`#[pymethods]`](crate::proc_macro::pymethods) blocks per
+//! [`#[pyclass]`](crate::proc_macro::pyclass).
```
birkenfeld, line 80:
```suggestion
//! [`#[pyclass]`](crate::proc_macro::pyclass), at the cost of some portability.
```

src/lib.rs
```diff
@@ -4,41 +4,132 @@
...
+//! is also implicitly acquired by PyO3 as it wraps Rust functions and structs
+//! into Python functions and objects.
 //!
-//! The vast majority of operations in this library will return `PyResult<...>`.
+//! ## The GIL-dependent types
+//!
+//! For example `&`[`PyAny`](crate::types::PyAny).
+//! These are only ever seen as references, with a lifetime that is only valid for as long
+//! as the GIL is held, which is why using them doesn't require a  [`Python<'py>`](crate::Python) token.
+//!  The underlying Python object, if mutable, can be mutated through any reference.
```
birkenfeld, line 29:
```suggestion
//!  The underlying Python object, if mutable, can be mutated through any reference.
```

src/lib.rs
```diff
@@ -4,41 +4,132 @@
...
+//!
+//! See the [guide](https://pyo3.rs/main/types.html) for an explanation of the different Python object types.
+//!
+//! ## The GIL-independent types
+//!
+//! When wrapped in [`Py`]`<...>`, like with [`Py`]`<`[`PyAny`](crate::types::PyAny)`>` or [`Py`]`<SomePyClass>`, Python objects
+//! no longer have a limited lifetime which makes them easier to store in structs and pass between functions.
+//! However, you cannot do much with them without a
+//! [`Python<'py>`](crate::Python) token, for which you’d need to reacquire the GIL.
```
birkenfeld, line 38: Not mentioning `PyObject` here is intended, I assume? (It's probably for the best since it's supposed to be phased out anyway.)

birkenfeld (approved):
Looks good, just a few nits.

src/lib.rs
mejrs: What does cost of portability mean in this context? Does it not work on some platforms? If so, which ones?

src/lib.rs
mejrs: Yeah; imo it's pretty confusing to have both uses of `Py<PyAny>` and `PyObject` mixed through documentation, which is why I thought it best not to mention it.

src/lib.rs
birkenfeld: From the guide: "Most users should only need a single `#[pymethods]` per `#[pyclass]`. In addition, not all platforms (e.g. Wasm) are supported by `inventory`. For this reason this feature is not enabled by default, meaning fewer dependencies and faster compilation for the majority of users."

mejrs: I'll add this:

> This adds a dependency on the [`inventory`](https://docs.rs/inventory) crate, which is not supported on all platforms.

src/lib.rs
```diff
@@ -4,41 +4,132 @@
...
+//! Building with PyPy is also possible (via cpyext) for Python 3.6,
+//! targeted PyPy version is 7.3+. Please refer to the
+//! [pypy section](https://pyo3.rs/main/building_and_distribution/pypy.html)
+//! in the guide for more information.
+//!
+//! # Example: Building a native Python module
+//!
+//! To build, test and publish your crate as a Python module, it is recommended that you use
+//! [maturin](https://github.com/PyO3/maturin) or
+//! [setuptools-rust](https://github.com/PyO3/setuptools-rust). You can also do this manually. See the
```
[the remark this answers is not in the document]
mejrs: I've now added it to the guide

mejrs:
@birkenfeld I think that's everything. Thanks for the review 😇

birkenfeld (approved):
+1

src/lib.rs
```diff
@@ -4,41 +4,133 @@
...
+//! [`Deserialize`](https://docs.rs/serde/latest/serde/trait.Deserialize.html) traits for
+//! [`Py`]`<T>` for all `T` that implement
+//! [`Serialize`](https://docs.rs/serde/latest/serde/trait.Serialize.html) and
+//! [`Deserialize`](https://docs.rs/serde/latest/serde/trait.Deserialize.html).
+//!
+//! ## Unstable features
+//!
+//! - `nightly`: Gates some optimizations that rely on
+//!  [`#![feature(specialization)]`](https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md),
+//! for which you'd also need nightly Rust. You should not use this feature.
```
davidhewitt, line 103: Fully agree 😂. I keep wondering if we should remove this feature!

guide/src/building_and_distribution.md
```diff
@@ -35,9 +35,16 @@ pyo3 = { {{#PYO3_CRATE_VERSION}} }
...
+There are two main ways to test, build and distribute your module as a Python package: [setuptools-rust] and [maturin]. setuptools-rust needs several configuration files (`setup.py`, `MANIFEST.in`, `build-wheels.sh`, etc.). It allows (and sometimes requires) writing custom workflows in python. maturin has only few options and works without any additional configuration, instead it requires a rigid project structure and does not support some functionality of setuptools such as package data ([pyo3/maturin#258](https://github.com/PyO3/maturin/issues/258)), multiple extensions or running python scripts at build time.
+
+While developing, you can also symlink (or copy) and rename the shared library from the `target` folder:
```
davidhewitt, line 42: I'd be tempted to give this a section header `### Manual builds`.

Also, I don't think we need the prefix "while developing" - it's totally possible users might have their own build scripts that they use to release their code!

davidhewitt (approved):
Thanks both, looks fantastic! Just a couple of tiny suggestions from me.

I'm traveling this week with a significantly worse internet connection than I was hoping for, so I may struggle to do much reviewing or other progress before next weekend. Apologies in advance.

src/lib.rs
mejrs: We might be able to substitute with `min_specialization`? It is less unsound than full `specialization`, at least...

I figured that since people can see the feature exists it should at least be mentioned with a note that you shouldn't use it.

guide/src/building_and_distribution.md
mejrs: Done!

PR #1645, opened 2021-05-31
faq: add section about pyo3(get) cloning fields

mejrs:
Explains that pyo3(get) clones the field and to use Py<...> if that is undesirable. This has come up various times on SO and gitter.

### DIFF
--- a/guide/src/faq.md
+++ b/guide/src/faq.md
@@ -35,7 +35,7 @@ The Rust book suggests to [put integration tests inside a `tests/` directory](ht
 For a PyO3 `extension-module` project where the `crate-type` is set to `"cdylib"` in your `Cargo.toml`,
 the compiler won't be able to find your crate and will display errors such as `E0432` or `E0463`:
 
-```
+```text
 error[E0432]: unresolved import `my_crate`
  --> tests/test_my_crate.rs:1:5
   |
@@ -45,7 +45,7 @@ error[E0432]: unresolved import `my_crate`
 
 The best solution is to make your crate types include both `rlib` and `cdylib`:
 
-```
+```text
 # Cargo.toml
 [lib]
 crate-type = ["cdylib", "rlib"]
--- a/guide/src/faq.md
+++ b/guide/src/faq.md
@@ -45,7 +45,7 @@ error[E0432]: unresolved import `my_crate`
 
 The best solution is to make your crate types include both `rlib` and `cdylib`:
 
-```text
+```toml
 # Cargo.toml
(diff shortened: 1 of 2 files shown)


birkenfeld (approved):
LGTM!

mejrs:
Figured I'd add a doctest for it, too.

davidhewitt (approved):
👍 Looks great, thanks!

FWIW I was wondering about adding a new guide page about this and the other common noob traps (e.g. `&PyAny` instead of `PyAny` as argument) etc.

I was thinking a new page instead of adding to the FAQ because the FAQ is kind of about troubleshooting. I was thinking that a new page would be called "common patterns in PyO3" and would have all sorts of tips like this which a noob might find helpful to scan.

mejrs:
> FWIW I was wondering about adding a new guide page about this and the other common noob traps (e.g. `&PyAny` instead of `PyAny` as argument) etc.
>
> I was thinking a new page instead of adding to the FAQ because the FAQ is kind of about troubleshooting. I was thinking that a new page would be called "common patterns in PyO3" and would have all sorts of tips like this which a noob might find helpful to scan.

I actually really like how the nomicon does this through examples, e.g.:

> Naively, this means we just want this design:
>
> ```rs 
> pub struct Vec<T> {
>     ptr: *mut T,   
>     cap: usize,
>     len: usize,
> }
> ```
>
> And indeed this would compile. Unfortunately, it would be incorrect. First, the compiler will give us too strict variance. So a &Vec<&'static str> couldn't be used where an &Vec<&'a str> was expected. More importantly, it will give incorrect ownership information to the drop checker, as it will conservatively assume we don't own any values of type T. See the chapter on ownership and lifetimes for all the details on variance and drop check.

Generally I think the way to push good practices is to just do it through examples, documentation and good compile errors (can this be done for PyAny?), as well by just making good patterns easier to use than bad patterns. Python::with_gil vs GilGuard is a great example of this


As for where it's located I don't really mind as long as it's easy to link to.

davidhewitt:
👍 Fully agree with all of that. I was imagining this new section of the guide would be full of good examples pretty much like you suggest.

As for making bad patterns harder to use than good patterns... I feel optimistic each PyO3 release gets slightly more ergonomic 😅

mejrs:
Can you merge this? I'm not planning to make a new guide section at this point.

davidhewitt:
Sorry yes I'll merge now! I might start the guide section in Saturday, if I find time after catching up with everything else...

Issue #1640, opened 2021-05-27
Better Type error messages when FromPyObject fails on complex types


batconjurer:
I am currently working on a project where complex types are passed to Rust from Python and must be extracted to a Rust type. Most of the `FromPyObject` traits are implemented via the corresponding derive macro. 

However, the type errors that we get don't provide a lot of help in pinpointing what went wrong when incorrect input is given. Consider the following example 
```rust
// Rust
#[derive(FromPyObject)]
pub struct Inner {
   hi: i32,
   low: i32,
}

#[derive(FromPyObject)]
pub struct Outer {
    inner: Inner,
    extra: i32,
}

#[pyfunction]
fn foo(bar: Outer, baz: i32) {
    // --- snip -- //
}
```
```python
# Python 

class Inner(object):

    def __init__(self, hi, low):
        self.hi = hi
        self.low = low


class Outer(object):
    
    def __init__(self, inner, extra):
        self.inner = inner
        self.extra = extra
```
And then I call `foo(Outer(Inner(1.0, 0), 2), 3)` by accident. The error I get in Python will be something along the lines of
``` TypeError: 'float' object cannot be interpreted as an integer ```

What I would like to see is something along the lines of  
``` 
TypeError: Failed to extract Outer.inner because :: 
-- TypeErrror: Failed to extract Inner.hi because ::
-- TypeError: 'float' object cannot be interpreted as an integer 
```
although I am not at all attached to the formatting.

I am willing to work on this and playing around, I got the above to work by doing the following in 
https://github.com/PyO3/pyo3/tree/5446fe2062cb3bf11bf61bd4a2c58a7ed8b408d2/pyo3-macros-backend/src/from_pyobject.rs#L243
```rust
 let extractor = match &attrs.from_py_with {
     None => quote!(#get_field.extract().map_err(|inner| {
         let err_msg = format!("Failed to extract field {}.{} because:: \n-- {} ",
             stringify!(#self_ty),
             stringify!(#ident),
             inner);
             pyo3::exceptions::PyTypeError::new_err(err_msg)
     })?),
    Some(FromPyWithAttribute(expr_path)) => quote! (#expr_path(#get_field).
        map_err(|inner| {
            let err_msg = format!("Failed to extract field {}.{} because:: \n-- {} ",
                stringify!(#self_ty),
                stringify!(#ident),
                inner);
            pyo3::exceptions::PyTypeError::new_err(err_msg)
      })?),
};
```
[...]

PR #1649, opened 2021-06-01
WIP: Enhanced extract type errors (Issue #1640)

batconjurer:
This PR includes allows the `FromPyObject` derive macro give more detailed error messages when conversion fails. It constructs a "traceback" of the nested types that failed the conversion to help more accurately pinpoint from where the type error originated. 

The following gives an examples for different complex types (REMEMBER TO CHANGE THIS WHEN FORMATTING IS FINALIZED):

### Enum
```rust
pub enum Bar {
    #[pyo3(annotation = "str")]
    A(String),
    #[pyo3(annotation = "uint")]
    B(usize),
    #[pyo3(annotation = "int", transparent)]
    C(isize),
}

fn main() {
    let py = gil.python();
    let dict = PyDict::new(py);
    let f = Bar::extract(dict.as_ref()).unwrap();
}
```

This produces the following error message:
```
TypeError: Failed to extract type Bar ('Union[str, uint, int]')
- variant A (str): 'dict' object cannot be converted to 'PyString'
- variant B (uint): 'dict' object cannot be interpreted as an integer
- variant C (int): 'dict' object cannot be interpreted as an integer
```

Note that for Enums, a reason for the failure of each variant is provided.  

### Struct

```rust
#[pyclass]
struct PyBaz {
    #[pyo3(get)]
    tup: (String, String),
    #[pyo3(get)]
    e: PyE,
}

#[pyclass]
#[derive(Clone)]
pub struct PyE {
    #[pyo3(get)]
    test: String,
    #[pyo3(get)]
    test2: usize,
}


pub struct E<T, T2> {
    test: T,
    test2: T2,
}

struct Baz<U, T> {
    e: E<U, T>,
    tup: Tuple,
}

fn main() {
  let py = gil.python();
  let pybaz = PyBaz {
      tup: ("test".into(), "test".into()),
      e: PyE {
          test: "foo".into(),
          test2: 0,
      },
  }
  .into_py(py);
  let test: Baz<usize, String> = FromPyObject::extract(pybaz.as_ref(py)).unwrap();
}
```
This produces the following error message (in the Python interpreter):
```python
TypeError                       Traceback (most recent call last)
TypeError: 'str' object cannot be interpreted as an integer

TypeError                       Traceback (most recent call last)
TypeError: failed to extract field E.test

TypeError                       Traceback (most recent call last)
<python-process> in <module>
----> 1 module_name.module_func(x)

TypeError: failed to extract field Baz.e
```

### Transparent Struct

```rust
#[pyo3(transparent)]
pub struct B {
    test: String,
}

fn main() {
    let py = gil.python();
    let tup: PyObject = 1.into_py(py);
    let tup = B::extract(tup.as_ref(py)).unwrap();
}
```
This produces the error message (in Python)
```python
TypeError                       Traceback (most recent call last)
TypeError: 'int' object cannot be converted to 'PyString',

TypeError                       Traceback (most recent call last)
<python-process> in <module>
----> 1 module_name.module_func(x)

TypeError: failed to extract field B.test
```
### Tuple Struct

```rust
pub struct Tuple(String, usize);

fn main() {
    let py = gil.python();
    let tup: PyObject = (1, "test").into_py(py);
    let tup = Tuple::extract(tup.as_ref(py)).unwrap();
}
```
This produces the following error message: (in Python)
```python
TypeError                       Traceback (most recent call last)
TypeError: 'int' object cannot be converted to 'PyString'

TypeError                       Traceback (most recent call last)
<python-process> in <module>
----> 1 module_name.module_func(x)

TypeError: failed to extract field Tuple.0
```

### Transparent Tuple Struct

```rust
#[pyo3(transparent)]
pub struct TransparentTuple(String);

fn main() {
    let py = gil.python();
    let tup: PyObject = 1.into_py(py);
    let tup = TransparentTuple::extract(tup.as_ref(py)).unwrap();
}
```
This produces the following error message (in Python):
```python
TypeError                       Traceback (most recent call last)
TypeError: 'int' object cannot be converted to 'PyString'

TypeError                       Traceback (most recent call last)
<python-process> in <module>
----> 1 module_name.module_func(x)

TypeError: failed to extract inner field of TransparentTuple
```


davidhewitt:
@batconjurer - thank you for working on this and sorry for my delay in reviewing.

Reading the generated error message and thinking about this some more, I'd like to suggest a slightly denser format which I think is easier for the user to see what happened:

Enum case:

```
TypeError: failed to extract enum Bar (Union[str,uint,int])
- variant A (str): 'dict' object cannot be converted to 'PyString'
- variant B (uint): 'dict' object cannot be interpreted as an integer
- variant C (int): 'dict' object cannot be interpreted as an integer
```

Struct field:

```
TypeError: failed to extract field Bar.foo: 'dict' object cannot be converted to 'PyString'
```

You can get the error message without the error type using `py_err.instance(py).str()`.

I'd prefer to avoid adding the `error_message` option for now. I think that customizing errors on that level is rare enough that I'm not in favour of extending the PyO3 macro API yet. (Perhaps my view on that will change in the future.)

As for why `PyString` appears - this is because the `TypeError` has come from `PyDowncastError`. In the future I'd like to improve that error to say `'dict' object cannot be converted to 'str'`. I don't think `PyString` is at all useful for a Python user to see.

batconjurer:
@davidhewitt I have made the suggested changes. A couple of comments
 * In order to get the message body of errors without the error type, I have to acquire the gil many times. I'm wondering if that is a problem
 * Again, for very complex structs, these error messages are going to be quite long to be in a single line.

davidhewitt:
Thanks, I will re-review at the weekend.

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -61,6 +61,11 @@ impl<'a> Enum<'a> {
                 if maybe_ret.is_ok() {
                     return maybe_ret
                 }
+                if let Err(inner) = maybe_ret {
+                    let gil = Python::acquire_gil();
+                    let py = gil.python();
```
davidhewitt, line 66: I wonder if you are able to use `obj.py()` here to avoid needing to acquire the GIL?

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -61,6 +61,11 @@ impl<'a> Enum<'a> {
                 if maybe_ret.is_ok() {
                     return maybe_ret
                 }
+                if let Err(inner) = maybe_ret {
```
davidhewitt, line 64: This pair of `if ok() / if let Err` looks like a `match` could be more appropriate.

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -74,10 +79,15 @@ impl<'a> Enum<'a> {
...
+        let ty_name = self.enum_ident.to_string();
         quote!(
+            let mut err_reasons = String::new();
             #(#var_extracts)*
             let type_name = obj.get_type().name()?;
-            let err_msg = format!("'{}' object cannot be converted to '{}'", type_name, #error_names);
+            let mut err_msg = format!("failed to extract enum {} ('{}')\n",
+                #ty_name,
+                #error_names);
+            err_msg.push_str(&err_reasons);
```
davidhewitt, line 90:
```suggestion
            let err_msg = format!("failed to extract enum {} ('{}')\n{}",
                #ty_name,
                #error_names,
                &err_reasons);
```

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -74,10 +79,15 @@ impl<'a> Enum<'a> {
         } else {
             error_names
         };
+        let ty_name = self.enum_ident.to_string();
         quote!(
+            let mut err_reasons = String::new();
             #(#var_extracts)*
             let type_name = obj.get_type().name()?;
```
davidhewitt, line 86: This looks now unused.

```suggestion
```

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -196,19 +206,55 @@ impl<'a> Container<'a> {
...
+            let err_msg = format!(
+                "failed to extract field {}.{}",
+                quote!(#self_ty),
+                quote!(#ident)
+            );
             quote!(
-                Ok(#self_ty{#ident: obj.extract()?})
+                Ok(#self_ty{#ident: obj.extract().map_err(|inner| {
+                    let gil = Python::acquire_gil();
+                    let py = gil.python();
```
davidhewitt, line 217: You can use `obj.py()` here to avoid acquiring the GIL.

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -196,19 +206,55 @@ impl<'a> Container<'a> {
...
+            let err_msg = if self.is_enum_variant {
+                let variant_name = &self.path.segments.last().unwrap();
+                format!("- variant {} ({})", quote!(#variant_name), &self.err_name)
+            } else {
+                format!("failed to extract inner field of {}", quote!(#self_ty))
+            };
+            quote!(
+                Ok(#self_ty(obj.extract().map_err(|inner| {
+                    let gil = Python::acquire_gil();
+                    let py = gil.python();
```
davidhewitt, line 234: Similarly use `obj.py()`.

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -196,19 +206,55 @@ impl<'a> Container<'a> {
...
     fn build_tuple_struct(&self, len: usize) -> TokenStream {
         let self_ty = &self.path;
         let mut fields: Punctuated<TokenStream, syn::Token![,]> = Punctuated::new();
         for i in 0..len {
-            fields.push(quote!(s.get_item(#i).extract()?));
+            let error_msg = format!("failed to extract field {}.{}", quote!(#self_ty), i);
+            fields.push(quote!(
+                s.get_item(#i).extract().map_err(|inner| {
+                let gil = Python::acquire_gil();
+                let py = gil.python();
```
davidhewitt, line 252: I think `s.py()` can be used here.

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -238,11 +284,29 @@ impl<'a> Container<'a> {
...
+            let conversion_error_msg =
+                format!("failed to extract field {}.{}", quote!(#self_ty), ident);
             let get_field = quote!(obj.#getter?);
             let extractor = match &attrs.from_py_with {
-                None => quote!(#get_field.extract()?),
-                Some(FromPyWithAttribute(expr_path)) => quote! (#expr_path(#get_field)?),
+                None => quote!(
+                    #get_field.extract().map_err(|inner| {
+                    let gil = Python::acquire_gil();
+                    let py = gil.python();
```
davidhewitt, line 294: Looks like `obj.py()` can be used here.

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -238,11 +284,29 @@ impl<'a> Container<'a> {
...
+                    let err_msg = format!("{}: {}",
+                        #conversion_error_msg,
+                        inner.instance(py).str().unwrap());
+                    pyo3::exceptions::PyTypeError::new_err(err_msg)
+                })?),
+                Some(FromPyWithAttribute(expr_path)) => quote! (
+                    #expr_path(#get_field).map_err(|inner| {
+                        let gil = Python::acquire_gil();
+                        let py = gil.python();
```
davidhewitt, line 303: Looks like `obj.py()` can be used here.

davidhewitt (commented):
@batconjurer I'm sorry this took me so long to re-review. If you've lost interest because of the delay, that's understandable, and I'm happy to take up finishing this off.

The `acquire_gil` uses can definitely be avoided. I've added some comments on what to do.

I'm really happy with the enum error message.

I agree that the struct message is still too long.

New idea - we added [`PyErr::set_cause`](https://docs.rs/pyo3/0.14.1/pyo3/prelude/struct.PyErr.html#method.set_cause) in PyO3 0.14. Rather than formatting the old error into the new error, how about setting the original extraction error as the cause of the "failed to extract B.a" error?

That way it should break up across lines naturally thanks to Python.

batconjurer:
@davidhewitt I am still interested in working on it. I will try to incorporate your comments as soon as I can.

batconjurer:
@davidhewitt So I changed the structs to use the `set_cause` method and I'm pretty happy with it. However, I'm pretty sure that absolutely need the gil to do. On top of that, I could not understand your suggestion for getting rid of the gil above. I can't find any mention of a `.py()` method

birkenfeld:
@batconjurer what David means is that you don't need to call `acquire_gil` since must already be held any time you have a `&PyAny`.  The `py()` method is on the `PyNativeType` trait, which is not in scope 
(the error message when you try `obj.py()` should have told you that), so you have to `use` the trait or call it explicitly as `pyo3::PyNativeType::py(obj)` etc.

batconjurer:
Ok, I have removed the `acquire_gil` calls. However, instead of importing the `PyNativeType` to get access to the `.py()` method, I found it simpler to just copy the body of that function. So now we have 
```rust
let py = unsafe{ Python::assume_gil_acquired() };
```
in all places

davidhewitt:
Hmmm. I really would like us to avoid `unsafe` in the code generated by this macro if possible. I've opened #1751 which would make it possible to use `.py()` without the import. If merged, we should update to use that. Otherwise, we should definitely use the `PyNativeType` trait here, even if it produces a bit of noise.

tests/test_frompyobject.rs
```diff
@@ -6,6 +6,21 @@ use pyo3::PyMappingProtocol;
 #[macro_use]
 mod common;
 
+/// Helper function that concatenates the error message from
+/// each error in the traceback into a single string that can
+/// be tested.
+fn extract_traceback(mut error: PyErr) -> String {
```
davidhewitt, line 12: As this function seems to be called from tests which all run with the GIL, it'd be faster to just take `py: Python` as an argument rather than calling `acquire_gil`.

tests/test_frompyobject.rs
```diff
@@ -170,6 +186,108 @@ fn test_transparent_tuple_struct() {
...
+#[derive(Debug, FromPyObject)]
+struct Baz<U, T> {
+    e: E<U, T>,
+    tup: Tuple,
+}
+
+#[test]
+fn test_struct_nested_type_errors() {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
```
davidhewitt, line 206: Nit: can we please use `Python::with_gil` rather than `acquire_gil` in this and all the other new tests please? We're intending to move away from `acquire_gil` because of underlying challenges with that API.

tests/test_frompyobject.rs
```diff
@@ -6,6 +6,21 @@ use pyo3::PyMappingProtocol;
...
+/// each error in the traceback into a single string that can
+/// be tested.
+fn extract_traceback(mut error: PyErr) -> String {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+    let mut error_msg = error.to_string();
+    while let Some(err) = error.cause(py) {
+        error_msg.push_str(": ");
+        error_msg.push_str(&err.to_string());
+        error = err;
```
davidhewitt, line 19: Nit: would be nice to use `cause` for the binding:

```suggestion
    while let Some(cause) = error.cause(py) {
        error_msg.push_str(": ");
        error_msg.push_str(&cause.to_string());
        error = cause;
```

davidhewitt (commented):
👍 looking really great! I like the way these errors look now, thank you for the repeated iterations of this PR.

Just a couple of tiny nits and then the question of which safe way to get `.py()` (maybe #1751 will be merged before long) and then I'd be keen to see this merged!

birkenfeld:
I don't understand:
```
let py = pyo3::PyNativeType::py(obj);
```
is shorter than your solution and does not need to introduce unsafe code.

batconjurer:
> @batconjurer I don't understand:
> 
> ```
> let py = pyo3::PyNativeType::py(obj);
> ```
> 
> is shorter than your solution and does not need to introduce unsafe code.

Incorportated this. Did not realize at first that `obj` is in scope the whole time.

tests/test_frompyobject.rs
batconjurer: :+1:

tests/test_frompyobject.rs
batconjurer: :+1:

tests/test_frompyobject.rs
batconjurer: :+1:

davidhewitt (approved):
This is looking perfect. Finger crossed CI is green - if so, I think it's good to merge! Thanks for all your efforts!

birkenfeld:
> Incorportated this. Did not realize at first that `obj` is in scope the whole time.

Great, thanks!

birkenfeld (approved):

Issue #1601, opened 2021-05-11
Macro attribute reworks


davidhewitt:
With #1567 merged we created the `#[pyo3(name = "...")]` attribute for `#[pyfunction]`.

For consistency we should also support this pattern in the other macros too:
 - [x] name of `#[pymodule]` (implemented in #1650)
 - [x] name of `#[pyfn]` should become optional and support the `#[pyo3(name = "...")]` syntax (implemented in #1610)
 - [x] `#[pyo3(get, set, name = "...")]` on `#[pyclass]` struct fields (including tuple structs, where setting the name will be required). (implemented in #1651)
 - [ ] `#[pyo3(getter, name = "...")]` and `#[pyo3(setter, name = "...")]` in `#[pymethods]`.

There's also some other attributes which I would like to move to the `#[pyo3(...)]` namespace for the 0.14 release:
 - [x] `#[text_signature]` on `#[pyfunction]` and `#[pymethods]` (implemented in #1658)
 - [ ] `#[args]` on `#[pymethods]` (perhaps call it `#[pyo3(signature = (...))]` instead? Or rename the undocumented signature merged in #1567 to `#[pyo3(args = (...))]`)
 - [x] `#[pyfunction(pass_module)]` -> `#[pyfunction] #[pyo3(pass_module)]` (implemented in #1567; documented in #1659)

Once we have all of these in place, we'll be much closer to a standard syntax for these attributes. This can lead to possible future improvements like a possible `#[pyo3]` macro (https://github.com/PyO3/pyo3/pull/1567#issuecomment-830697538, #211).

davidhewitt:
I'll aim to implement each of these as individual PRs over the coming weeks heading into 0.14.

If anyone else is interested in implenting any of these, please feel welcome to claim them on this issue! I can give further information if the notes above aren't clear enough in any spots.

davidhewitt:
I think once #1650 is merged then I'm happy with the changes done in advance of the 0.14 release; I think the rest of the changes here can wait for a future release:

- Changing `#[getter]` to `#[pyo3(getter)]` potentially affects a lot of users of `#[pymethods]` so I think we can pause for a moment and take stock before doing it.
- `#[pyo3(signature = (...))]` would be nice to support on `#[pyfunction]` and `#[pymethods]`, however I think there's a few known issues with the current `#[args]` annotation (#794 #1439), so I think it's a bigger piece of work than I'd like to delay the 0.14 release for.

[...]

PR #1650, opened 2021-06-04
pymodule: accept `#[pyo3(name = "...")]` option

davidhewitt:
Part of #1601 

This PR makes `#[pymodule]` accept the `#[pyo3(name = "...")]` attribute, and deprecates the previous `#[pymodule(name)]` way to set the name.


messense (approved):
LGTM!

PR #1639, opened 2021-05-26
Add automated continuous benchmarking

messense:
Closes #1634

PR #1652, opened 2021-06-04
Run pytest pyo3-benchmarks on CI

messense:
Follow-up of #1639


davidhewitt (approved):
LGTM, thanks for doing this!

Let's merge it and see what the result looks like on https://pyO3.rs/dev/bench !

PR #1653, opened 2021-06-05
PyTuple: use fast macros on !abi3

birkenfeld:
They were already defined, but not used so far.  For abi3 they now forward
to the C functions.

**Question:** should these be defined at all for abi3, or should I add the conditional code in the `PyTuple` impl?

### DIFF
--- a/src/types/tuple.rs
+++ b/src/types/tuple.rs
@@ -26,6 +26,9 @@ impl PyTuple {
         unsafe {
             let ptr = ffi::PyTuple_New(len as Py_ssize_t);
             for (i, e) in elements_iter.enumerate() {
+                #[cfg(not(Py_LIMITED_API))]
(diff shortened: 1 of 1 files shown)


davidhewitt:
It looks like these macros are only defined in https://github.com/python/cpython/blob/main/Include/cpython/tupleobject.h, i.e. for the unlimited abi.

So I'd vote for doing the same for our `ffi` module - only define these macros for the unlimited abi and write the conditional code inside our tuple code.

If there's lots of places in the code where this happens, can always add tiny `#[inline] pub(crate) fn pytuple_set_item` helpers which are defined differently for each `#[cfg]` and call to the relevant ffi function.

birkenfeld:
So far, these were each only used in one place, so I went and added the conditional there.

davidhewitt (approved):
LGTM, nice one!

FWIW, I get ~2% speedup with `tuple_get_item` benchmark on this branch, and ~8% speedup on a `tuple_new` benchmark I'll push in a minute.

PR #1654, opened 2021-06-05
Add support for extracting PathBuf from pathlib.Path

messense:


messense:
datafusion probably wants this: https://github.com/apache/arrow-datafusion/issues/491

src/conversions/osstr.rs
```diff
@@ -58,7 +58,22 @@ impl ToPyObject for OsStr {
...
+        let pystring = match <PyString as PyTryFrom>::try_from(ob) {
+            // Cast PyAny to PyString
+            Ok(s) => s,
+            Err(err) => {
+                let py = ob.py();
+                let pathlib = py.import("pathlib")?;
+                let pathlib_path = pathlib.getattr("Path")?;
+                // FIXME: right way to get the type object of pathlib.Path?
+                let path_instance = pathlib_path.call0()?;
+                if path_instance.get_type().is_instance(ob)? {
```
davidhewitt, line 70: You already got the type object just above ;)

```suggestion
                let pathlib_path: &PyType = pathlib.getattr("Path")?.downcast()?;
                if pathlib_path.is_instance(ob)? {
```

src/conversions/osstr.rs
```diff
@@ -58,7 +58,22 @@ impl ToPyObject for OsStr {
 
 impl FromPyObject<'_> for OsString {
```
davidhewitt, line 59: I feel like this new code should probably go in `src/converstions/path.rs` rather than also adding a conversion for `OsString` ?

src/conversions/osstr.rs
```diff
@@ -58,7 +58,22 @@ impl ToPyObject for OsStr {
...
     fn extract(ob: &PyAny) -> PyResult<Self> {
-        let pystring = <PyString as PyTryFrom>::try_from(ob)?; // Cast PyAny to PyString
+        let pystring = match <PyString as PyTryFrom>::try_from(ob) {
+            // Cast PyAny to PyString
+            Ok(s) => s,
+            Err(err) => {
+                let py = ob.py();
+                let pathlib = py.import("pathlib")?;
+                let pathlib_path = pathlib.getattr("Path")?;
+                // FIXME: right way to get the type object of pathlib.Path?
```
birkenfeld, line 68: Can't you "extract" a `PyType` object from `pathlib_path`?

messense:
```
error[E0277]: the trait bound `PyType: PyClass` is not satisfied
  --> src/conversions/osstr.rs:68:46
   |
68 |                 let path_type = pathlib_path.extract::<PyType>()?;
   |                                              ^^^^^^^ the trait `PyClass` is not implemented for `PyType`
   |
   = note: required because of the requirements on the impl of `conversion::FromPyObject<'_>` for `PyType`


error[E0277]: the trait bound `PyType: Clone` is not satisfied
  --> src/conversions/osstr.rs:68:46
   |
68 |                 let path_type = pathlib_path.extract::<PyType>()?;
   |                                              ^^^^^^^ the trait `Clone` is not implemented for `PyType`
   |
   = note: required because of the requirements on the impl of `conversion::FromPyObject<'_>` for `PyType`
```

davidhewitt (commented):
Looks like a good idea, I was wondering about doing similar in the past. Couple of thoughts...

Also definitely worthy of a CHANGELOG entry! 🚀

src/conversions/osstr.rs
davidhewitt: Need to use `.downcast()` or `.extract::<&PyType>()` - see my comment below!

src/conversions/osstr.rs
messense: 😅 Ahh, it's not very clear to me that you can get `&PyType` by downcasting.

src/conversions/osstr.rs
messense: Done.

src/conversions/osstr.rs
birkenfeld: Sorry, should have specified `&PyType`.

davidhewitt (approved):
LGTM!

Issue #1647, opened 2021-06-01
feature suggestion: extending PyTuple


alonblade:
Right now it is impossible to

```rust
#[pyclass(extends=PyTuple)]
struct Bla {
// ...
}
```

I'd like to do that to get a rust namedtuple like class.

What would that entail?

davidhewitt:
At the moment this is disabled because we weren't confident that subclassing `PyTuple` would work correctly.

The complexity is during `__new__` implementation. A subclass of `PyTuple` needs to pass the number of items in the tuple up to the `PyTuple` base class, which we don't do correctly right now. In addition, the items are laid out in memory after the main class data. This shouldn't be a problem, but we weren't sure.

I actually have a patch which might tentatively fix the above (among with some other problems). If you give me a few days, I'll try to push that patch and cc you. Once it works, we can try subclassing `PyTuple` and see whether we can make it work.

alonblade:
Sounds good, thanks.

Sh3mm:
I'm a bit late to the party, but for those who found this post after me: 

It's not perfect, but here's a macro to creates a `struct` that can be converted to and from an equivalent `namedtuple`. It does this by calling the python function, so conversion is costly. 
``` rust
#[macro_export]
macro_rules! namedtuple {
    ( $(#[$meta:meta])* $name:ident, $($field:ident: $f_type:ty),+ ) => {
        $(#[$meta])*
        pub struct $name{
            $(pub $field: $f_type,)+
        }

        impl $name {
            pub fn new($($field: $f_type,)+) -> Self {
                return Self{$($field,)+};
            }
        }

        impl<'py> FromPyObject<'py> for $name {
            fn extract_bound(obj: &Bound<'py, PyAny>) -> PyResult<Self> {
                $(let $field = obj.getattr(stringify!($field))?.extract::<$f_type>()?;)+
                return Ok($name{$($field,)+});
            }
        }

        impl<'py> IntoPyObject<'py> for $name {
            type Target = PyAny;
            type Output = Bound<'py, PyAny>;
            type Error = PyErr;

            fn into_pyobject(self, py: Python<'py>) -> Result<Self::Output, Self::Error> {
                let nt_type = py.import("collections")?
                    .getattr("namedtuple")?
                    .call((stringify!($name), ($(stringify!($field),)+)), None)?;
                Ok(nt_type.call(($(self.$field,)+), None)?)
            }
        }
    };
}
```

It can then be used this way:
``` rust
[...]

Issue #1637, opened 2021-05-26
[Question] Return different sub-classes in function of an enum?


azerupi:
Is it possible to return different sub-classes from a function?

In Rust I have an enum containing data
```rust
pub enum Frame {
    Type,
    TypeResponse(u8, Version, u8),
    Ping(u8),
    Pong(u8),
}
```

In Python I though I would expose this by having a `Frame` base class and a sub-class for each type of frame:
```rust
#[pyclass(name="Frame", subclass)]
pub struct PyFrame { /* some fields */ }

#[pyclass(extends=PyFrame, subclass)]
pub struct PingFrame {
    #[pyo3(get)]
    sequence: u8,
}

#[pymethods]
impl PingFrame {
    #[new]
    pub fn new(sequence: u8) -> (Self, PyFrame) {
        let frame = Frame::new_ping(sequence);
        ((&frame).try_into().unwrap(), frame.into())
    }
}

impl TryFrom<&Frame<'_>> for PingFrame {
    type Error = ();
    
    fn try_from(frame: &Frame) -> Result<Self, Self::Error> {
        match frame.payload() {
            Frame::Ping(sequence) => Ok(PingFrame { sequence: *sequence }),
            _ => Err(())
        }
    }
}
```

Now I'm trying to have a function that can return any of the sub-classes depending on the frame enum variant. Is that possible?

```rust
fn parse(&mut self, py: Python, byte: u8) -> PyResult<Option<PyObject>> {
        match self.parser.parse(byte) {
            Ok(f) => Ok(Some(PingFrame::new(5).into_py(py))),
            Err(e) => return Err(ParseError { err: e }.into())
        }
    }
```

I tried to use `PyFrame` and `PyObject` in the return type, but that doesn't seem to work.

birkenfeld:
I feel like I've figured out the solution to this once already, but I don't remember where :(

The tuple (Subclass, Baseclass) *should* be possible to convert to a PyObject but it clearly doesn't right now.

Maybe we should rethink the subclassing strategy and require a field of the subclass to be of type Base?

mejrs:
The guide has a [paragraph on inheritance](https://pyo3.rs/v0.13.2/class.html#inheritance) with an example that looks like what you are trying to do.

[...]

PR #1657, opened 2021-06-05
pyclass: refactor `tp_new` / `tp_dealloc` / remove `PyCellInner`

davidhewitt:
This PR is a bunch of cleanups to `PyCell` and `PyClassInitializer` code for the slot implementations of `tp_new`, `tp_alloc`, `tp_dealloc` and `tp_free`:

 - `tp_alloc` and `tp_free` are only implemented for `freelist` classes, where they provide custom memory allocation / release
 - `tp_new` now does the same thing for all `#[pyclass]` (call super `tp_new` and initialize the `PyCell` contents)
 - `tp_dealloc` does the reverse of `tp_new` (drop the `PyCell` contents and then call super `tp_dealloc`)

This work seems necessary to me as a precursor to #1647, #1637, #1344 and #991.

While doing this I also found a nasty bug that we don't call `tp_dealloc` of inherited native types, which means that dict subclasses never decrease reference counts of all their contents on drop (for example).

I added a benchmark for class instantiation. Sadly these cleanups don't seem to have meaningfully changed the benchmark. I did notice that we're a lot faster than the numbers reported in #661, so there must have been progress over time.


davidhewitt:
Oooh nice, a segfault purely on windows 32-bit 👀

davidhewitt:
Turns out I forgot `#[repr(C)]` from `PartiallyInitializedPyCell` 😬

davidhewitt:
If nobody reviews of raises any concerns by the weekend I plan to merge this.

PR #1660, opened 2021-06-05
Refactor wrapper generation, and enable fastcall for static/class methods.

birkenfeld:
So I managed it after all...

This has, unfortunately, pretty large changes moving stuff around which are not very review friendly.  I suggest to look mostly at the new code.

`impl_arg_param(s)` are just moved, no changes to review there.


davidhewitt (approved):
This is absolutely brilliant; I'd been wondering about a similar refactoring but never got around to it.

Really happy with the new code arrangement and adds a performance optimization on top! 💯

PR #1664, opened 2021-06-06
PyList: remove get_parked_item, use macros for speed on !abi3

birkenfeld:
I assume that the `get_parked_item` API is a remnant from long ago?  I don't see any other reference to "parked" in the whole codebase, and for users this is just confusing.

BTW, why are the `PyListObject` struct and the fast "macros" defined in `ffi::cpython::listobject`, but in `ffi::tupleobject` for `PyTuple`?


davidhewitt:
Fully agree `get_parked_item` should go; do you think we should give it one release version deprecation cycle just in case anyone is using it?

> BTW, why are the PyListObject struct and the fast "macros" defined in ffi::cpython::listobject, but in ffi::tupleobject for PyTuple?

I think that's just because the "ffi cleanup" #1289 hasn't made it to `tupleobject` yet (it's mostly been done alphabetically).

davidhewitt (approved):
Regarding the macro changes: these make about 30% speedup to list `get_item` and iteration for me, which is awesome 🚀

birkenfeld:
At least I'm not seeing any uses searching on GitHub.  I think this is fine without deprecation cycle, considering the various other changes that might have to be made when updating.

davidhewitt:
👍 I'm persuaded! Thanks for making all these nice optimizations 🚀

PR #1665, opened 2021-06-06
Argument parsing: use VARARGS for *any* function taking **kwds

birkenfeld:
This is a test balloon - the benchmark seems a little faster but not too much, it's also just a single signature there.  Would be nice to test it with more different situations...


davidhewitt:
As it looks like this is marginally better then I think it makes sense to do this. We can always tweak again as future changes might affect the balance.

If you've got a moment to add some more benchmarks testing only-*args, only-*kwargs and some other combinations I'm sure that'll be useful in the future too.

davidhewitt (approved):
The extra functions look great. Thanks again for the hard work you've put into the recent optimizations!

Issue #1667, opened 2021-06-06
PyList indexing: negative indices?


birkenfeld:
`PyList::get_item` and `PySequence::get_item` accept `isize` indices.

`PyList::get_item`, however, does not support them and segfaults :(  (Also before the change to use the macro, since there was no check for error returns of `PyList_GetItem`.)

The `assert` in there is written as if negative indices were supported.

`PyTuple::get_item` accepts `usize`.

Also, `PySequence::get_item` returns a Result while `PyList` and `PyTuple` panic on invalid indices.

I see several things to think about here:

* Should we use `Index` for the panicing impls? `get_item` should then always return a Result.
* Should `PyList` support negative indices (we'd have to implement the correction ourselves)?  If yes, the same should apply to `PyTuple`.
* What about other APIs that take indices? E.g. `set_item`.

I also noticed some inconsistencies of the tuple and list APIs.  E.g., tuple as `slice` (and I don't mean `as_slice`, it's clear that list can't have that) but list doesn't.

birkenfeld:
OK, I checked all current APIs of list, tuple and sequence that take indices:

**Doesn't work**:
- PyTuple::slice - always `()`
- PyTuple::split_from - always full slice
- PyList::get_item - crashes
- PyList::set_item - raises IndexError

**Works**:
- PyList::insert
- all PySequence APIs

**Takes usize, not isize**:
- PyTuple::get_item

davidhewitt:
Mmm, thanks for taking the time to look at this; I'd noticed the inconsistencies but not had time to reason about them.

> Should we use `Index` for the panicing impls? `get_item` should then always return a `Result`.

I would 100% support implementing `Index`. Rather than `get_item` returning `Result`, it's perhaps possible also to mirror `Vec::get()` which returns `Option` and also does slicing? We might want to play around with that to figure out what feels nice.

> Should `PyList` support negative indices (we'd have to implement the correction ourselves)? If yes, the same should apply to `PyTuple`.

I think Python users would appreciate this. Note that if we implemented `Index<usize>` above then we could implement `Index<isize>` later...

> If we take only "real" indices, should the type switch to `usize`?

I think the function signature should mirror possible usage, so yes, `usize` if positive indices only`, `isize` if we allow negative-from-the-end lookup.

> What about other APIs that take indices? E.g. `set_item`.

I wonder if we can draw similar inspiration from `std::Vec` and friends.

birkenfeld:
[...]

PR #1668, opened 2021-06-06
PyList: fix segfault on get_item with negative indices

birkenfeld:
See #1667.


davidhewitt (approved):
Thanks for tidying up the tests at the same time.

I would think this is a user-facing crash worthy of a "Fixed" CHANGELOG entry?

birkenfeld:
Sure. Looks like I also unintentionally relied on `IntoIter` for arrays :)

davidhewitt:
Benchmark looks completely flaky so don't worry about that.

PR #1669, opened 2021-06-06
Reorganize method wrapper generation

birkenfeld:
Remove unneded splitting of functions in two, and reorder the
functions to order of the main match {} in gen_py_method().


davidhewitt (approved):
Thanks, nice reduction in line count and improved clarity! 💯

PR #1670, opened 2021-06-07
Add docs + slight refactor for num-complex and num-bigint dependencies

mejrs:
- adds docs and examples for the num-complex and num-bigint dependencies
- also adds a paragraph to the features section of the guide, which was missing them
- adds `nalgebra` as a dev-dependency - this is required to test the num-complex example. I really wanted an example that shows using `Complex` in a linear algebra crate, which is probably what most people using the feature want to do. I don't like adding dependencies, but given that it requires two other crates I would feel uncomfortable not testing it.
- shuffles around some code:
  - I put the relevant feature-gated code in the same module as the docs
  - I put the code in scr/types/num that depends on the Py_LIMITED/PyPy cfg's in a single block
  - both these things reduce a lot of the cfg littering in that module
  - this does make the diffs a pain to review 😳 sorry!

<p>

image
image

image
image

</p>
</details>


src/num_bigint.rs
```diff
@@ -0,0 +1,288 @@
...
+//!
+//! # Setup
+//!
+//! To use this feature, add this to your **`Cargo.toml`**:
+//!
+//! ```toml
+//! [dependencies]
+//! num-bigint = "0.4"
+//! pyo3 = { version = "0.14", features = ["num-bigint"] }
+//! ```
```
davidhewitt, line 22: Agreed this is useful to have but I worry we'll forget to update the versions...

src/types/complex.rs
```diff
@@ -17,205 +19,191 @@ pyobject_native_type!(
...
-    /// Creates a new Python `complex` object, from its real and imaginary values.
+    /// Creates a new `PyComplex` from the given real and imaginary values.
     pub fn from_doubles(py: Python, real: c_double, imag: c_double) -> &PyComplex {
         unsafe {
             let ptr = ffi::PyComplex_FromDoubles(real, imag);
             py.from_owned_ptr(ptr)
         }
     }
     /// Returns the real part of the complex number.
+    #[must_use = "method returns a new number and does not mutate the original value"]
```
davidhewitt, line 30: Is this `#[must_use]` really necessary? We could make a similar argument for all "getter" methods, but I don't see `#[must_use]` on most of them...

src/types/complex.rs
```diff
@@ -17,205 +19,191 @@ pyobject_native_type!(
...
     pub fn real(&self) -> c_double {
         unsafe { ffi::PyComplex_RealAsDouble(self.as_ptr()) }
     }
-    /// Returns the imaginary part the complex number.
+    /// Returns the imaginary part of the complex number.
+    #[must_use = "method returns a new number and does not mutate the original value"]
```
davidhewitt, line 35: Same query about `#[must_use]` as above.

davidhewitt (commented):
Overall this is a really nice tidy-up, thank you!

My main query is regarding the `#[must_use]` annotations. I highlighted two but there's a few more. Are they really necessary?

src/num_bigint.rs
mejrs: I don't think there is a way to auto update it :/ This is definitely going to be an annoying maintenance burden going forward.

I've seen other crates document this like:

```toml
[dependencies]
# replace * with the current pyo3 version
pyo3 = { version = "*", features = ["num-bigint"] }
```

src/types/complex.rs
mejrs: Right, real/imag don't actually return a PyComplex. I'll remove them. How do you feel about must_use on `.pow()`?

davidhewitt: I've tended to treat `#[must_use]` as "it's a bad bug if the result of this is not used". The existing two uses in PyO3, in `GilGuard` and `IntoPyPointer::into_ptr` are in the spirit of this.

In https://github.com/rust-lang/rust/issues/48926 the stdlib itself is trying to decide on the appropriate use of this attribute. I think I'd rather we stayed away from overuse of it until there's been consensus in that issue.

(I note that `f64::pow` is `#[must_use]`, but `num_complex::Complex::powf` is not.)

Did you come across a real-world bug arising from discarding the return value of these functions?

src/num_bigint.rs
davidhewitt: 👍 That seems like a really good idea. Could also say "check pyo3's `Cargo.toml`" for supported `num-bigint` version?

... or we could do something fun with an `mdbook` preprocessor to parse `Cargo.toml` and put the supported `num-bigint` version in the guide in a table 🙈

src/types/complex.rs
mejrs: Imho, must_use is appropriate if not using the return value makes the entire operation useless.

For example, I would argue that something like this is incorrect - it contains dead code and/or won't do what the author intends. Therefore a warning is appropriate.

```rust
fn main() {
    let mut i = 42_usize;
    i += 1;
    i.pow(2);
    println!("{:?}", i)
}
```
I'm happy to remove them from the methods that don't return `Self`. I'll remove them from `pow` if you feel strongly about it (?)

> Did you come across a real-world bug arising from discarding the return value of these functions?

Not these in particular; but I would have been burned by not handling `Result<()>` a few times, if it wasn't for that having a lint.

src/num_bigint.rs
mejrs: > or we could do something fun with an mdbook preprocessor to parse Cargo.toml and put the supported num-bigint version in the guide in a table 🙈

Is that even possible?

mejrs:
I'm not sure how to fix the failing `python3.9-x64 ubuntu-latest MSRV` test. The problem is that `nalgebra` won't build on msrv (officially, they only support latest stable).

I see the following solutions:
- don't test the example 😢
- put `nalgebra` as an optional dependency (dev-dependencies can't be optional) and only turn it on when testing on latest stable. I'm not a fan of that showing up as a feature when it doesn't really do anything useful for users.

src/types/complex.rs
birkenfeld: I think for getters it's overkill to apply this, especially when it is obvious that the method is a getter.

Personally, I tend to apply must_use only in cases where not using the return value can cause a hard-to-find bug. I.e. with `pow()`, if you test your code at least once, you'll find the bug.  For something returning `Result`, you won't if you only test the happy path.

davidhewitt:
Hmm, that's an awkward one.

I think I'd rather not do anything too complicated but agree it'd be nice to test the example. How about ditching `nalgebra` and writing example to find the complex roots of a quadratic polynomial instead? That's a reasonably simple formula which could be written without needing a third-party dependency.

src/num_bigint.rs
davidhewitt: I'm not aware of similar tricks to do dynamic docs, however if you want to go for something dynamic in the guide you might find the existing preprocessor I wrote useful: https://github.com/PyO3/pyo3/blob/main/guide/pyo3_version.py

src/types/complex.rs
davidhewitt: @birkenfeld's description of `#[must_use]` is pretty much exactly the same as my own view of it.

Overall I'm slightly leaning towards not adding it. I'd be happy to reverse that position if appropriate once `std` has settled on a convention how to use the attribute.

If you feel strongly that it should be there, I can be persuaded. My reservation if that if it doesn't fall into the convention `std` eventually sets then we'll want to remove it later.

src/num_bigint.rs
mejrs: This works and will update to the current pyo3 version automagically 🙌
```
//! ```toml
//! [dependencies]
//! num-complex = "*"
#![doc = concat!("pyo3 = { version = \"", env!("CARGO_PKG_VERSION"),  "\", features = [\"num-complex\"] }")]
//! ```
```

mejrs: ..but not on stable 😭

It's a syntax related thing so `cfg_attr(docsrs, ..)` won't work..a proc macro maybe? 🤔

birkenfeld: Since we have other locations where the current version must be substituted, I think the best way is a small (Python) script that replaces everything at once. This can then be run simply as part of the release process.

mejrs: I tried a couple of things but didn't find anything that didn't feel like a gross hack. So I've just taken the `*` approach as above.

birkenfeld: What's the "gross hack" with my suggestion?

mejrs: It's not, really. I figured out a proc macro that does it by just search+replacing text in a module, but it doesn't work in lib.rs and messes up the [src] links (which is pretty gross).

I don't like the idea of adding things to the release process for something that's at best "nice to have". It's not that big of a deal on its own but these things do add up.

mejrs:
I've `ignore`'d the example to avoid the problem but I do want to keep it because it shows a good use case for using the feature. 

I think that is all 👌

src/num_bigint.rs
davidhewitt: TBH I would be happy to have a python script to automate releasing (at the moment it's done manually by me), but i think that's best in a separate PR.

davidhewitt (approved):

kngwyu:
If you can afford it, could you please revise the tests to use `with_gil`?
BTW, thank you for all your efforts. Unfortunately, recently I don't have enough time for OSS, so I'm happy to see the project goes on this nice way.

mejrs:
@kngwyu sure!

davidhewitt:
Thanks!

PR #1677, opened 2021-06-13
Add a new `PyList::try_new` method

messense:
`PyList::new` requires a infallible iterator which is sometimes awkward to use.

Is this new API desirable?

### DIFF
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -31,6 +31,31 @@ unsafe fn new_from_iter<T>(
     ptr
 }
 
+#[inline]
+unsafe fn try_new_from_iter<T>(
+    elements: impl ExactSizeIterator<Item = PyResult<T>>,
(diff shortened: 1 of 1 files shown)


davidhewitt:
While I agree that our collection interfaces can definitely be made more user-friendly, I think we should take a bit of time to design something new before adding too many new functions.

I think this loosely combines well with the changes proposed in #1667 as something we can experiment with for 0.15, if you're ok with waiting for it for a bit?

messense:
> While I agree that our collection interfaces can definitely be made more user-friendly, I think we should take a bit of time to design something new before adding too many new functions.
> 
> I think this loosely combines well with the changes proposed in #1667 as something we can experiment with for 0.15, if you're ok with waiting for it for a bit?

No problem, it’s not blocking anything.

PR #1679, opened 2021-06-14
Add support for setting and retrieving exception cause

messense:


src/err/mod.rs
```diff
@@ -390,6 +390,29 @@ impl PyErr {
...
+        obj.map(|x| Self::from_instance(x))
+    }
+
+    /// Set the cause associated with the exception
+    pub fn set_cause(&self, py: Python, cause: Self) {
+        let cause = cause.into_instance(py);
+        unsafe {
+            ffi::PyException_SetCause(self.pvalue(py).as_ptr(), cause.as_ptr());
+        }
+    }
```
davidhewitt, line 407: What do you think about `set_cause(&self, py: Python, cause: Option<Self>)` instead? Then there's no need for `clear_cause`

It has a nice symmetry with `get_cause() -> Option<Self>`, but I guess it will force users to type `Some()` a lot.

davidhewitt (commented):
👍 I'm in favour of adding this. I have a few design choices I want to ask about:

- possibly merge `set_cause` / `clear_cause` - see below
- also, these APIs could potentially work with `&PyBaseException` instead of `Self`. Would be slightly closer to the underlying C APIs, though might in practice just force users to call `.into_instance(py)`.

I think the answers mostly depend on how you expect users to want to use these APIs? I assume you might have a use case which prompted you to add them.

messense:
I do have a use case for this, consider that a Rust extension module that accepts and calls a user supplied Python callback, the Python callback can raise many kind of exceptions, but the Rust extension module will wrap them as an unified exception for example `RuntimeError`, it'd be nice to set the exception cause to whatever exception the Python callback throws so it's easier to debug.

src/err/mod.rs
messense: 👍

davidhewitt (approved):
👍 thanks for tweaking, I think we can add this!

PR #1682, opened 2021-06-17
Allow calling Py::as_ref() and Py::into_ref() against PySequence, PyIterator and PyMapping. 

moriyoshi:
This patch allows one to call `Py::as_ref()` against `PySequence` by supplying `PyTypeInfo` for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add implementations for `Py::as_ref()` and `Py::into_ref()` for `Py<PySequence>`, `Py<PyIterator>` and `Py<PyMapping>`. [#1682](https://github.com/PyO3/pyo3/pull/1682)
+
 ### Changed
 
 - `#[classattr]` constants with a known magic method name (which is lowercase) no longer trigger lint warnings expecting constants to be uppercase. [#1969](https://github.com/PyO3/pyo3/pull/1969)
--- a/src/types/iterator.rs
+++ b/src/types/iterator.rs
@@ -2,7 +2,7 @@
 //
 // based on Daniel Grunwald's https://github.com/dgrunwald/rust-cpython
 
-use crate::{ffi, AsPyPointer, PyAny, PyErr, PyResult, Python};
+use crate::{ffi, AsPyPointer, IntoPyPointer, Py, PyAny, PyErr, PyNativeType, PyResult, Python};
 #[cfg(any(not(Py_LIMITED_API), Py_3_8))]
 use crate::{PyDowncastError, PyTryFrom};
 
@@ -94,6 +94,22 @@ impl<'v> PyTryFrom<'v> for PyIterator {
     }
 }
 
+impl Py<PyIterator> {
+    /// Borrows a GIL-bound reference to the PyIterator. By binding to the GIL lifetime, this
+    /// allows the GIL-bound reference to not require `Python` for any of its methods.
+    pub fn as_ref<'py>(&'py self, _py: Python<'py>) -> &'py PyIterator {
+        let any = self.as_ptr() as *const PyAny;
+        unsafe { PyNativeType::unchecked_downcast(&*any) }
+    }
+
+    /// Similar to [`as_ref`](#method.as_ref), and also consumes this `Py` and registers the
+    /// Python object reference in PyO3's object storage. The reference count for the Python
+    /// object will not be decreased until the GIL lifetime ends.
+    pub fn into_ref(self, py: Python) -> &PyIterator {
+        unsafe { py.from_owned_ptr(self.into_ptr()) }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::PyIterator;
@@ -101,8 +117,8 @@ mod tests {
     use crate::gil::GILPool;
     use crate::types::{PyDict, PyList};
     #[cfg(any(not(Py_LIMITED_API), Py_3_8))]
-    use crate::{Py, PyAny, PyTryFrom};
-    use crate::{Python, ToPyObject};
+    use crate::PyTryFrom;
+    use crate::{Py, PyAny, Python, ToPyObject};
     use indoc::indoc;
 
     #[test]
@@ -208,5 +224,27 @@ mod tests {
             let iter: &PyIterator = PyIterator::try_from(obj.as_ref(py)).unwrap();
             assert_eq!(obj, iter.into());
         });
+    }
+
+    #[test]
+    fn test_as_ref() {
+        Python::with_gil(|py| {
+            let iter: Py<PyIterator> = PyAny::iter(PyList::empty(py)).unwrap().into();
+            let mut iter_ref: &PyIterator = iter.as_ref(py);
+            assert!(iter_ref.next().is_none());
+        })
+    }
+
+    #[test]
+    fn test_into_ref() {
+        Python::with_gil(|py| {
+            let bare_iter = PyAny::iter(PyList::empty(py)).unwrap();
+            assert_eq!(bare_iter.get_refcnt(), 1);
+            let iter: Py<PyIterator> = bare_iter.into();
+            assert_eq!(bare_iter.get_refcnt(), 2);
+            let mut iter_ref = iter.into_ref(py);
+            assert!(iter_ref.next().is_none());
+            assert_eq!(iter_ref.get_refcnt(), 2);
+        })
     }
 }
--- a/src/types/sequence.rs
+++ b/src/types/sequence.rs
@@ -1,10 +1,10 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
 use crate::err::{self, PyDowncastError, PyErr, PyResult};
-use crate::ffi;
 use crate::internal_tricks::get_ssize_index;
 use crate::types::{PyAny, PyList, PyTuple};
-use crate::AsPyPointer;
+use crate::{ffi, PyNativeType};
+use crate::{AsPyPointer, IntoPyPointer, Py, Python};
 use crate::{FromPyObject, PyTryFrom, ToBorrowedObject};
 
 /// Represents a reference to a Python object supporting the sequence protocol.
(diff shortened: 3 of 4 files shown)


src/ffi/abstract_.rs
```diff
@@ -3,6 +3,12 @@ use crate::ffi::pyport::Py_ssize_t;
 use std::os::raw::{c_char, c_int};
 use std::ptr;
 
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    #[cfg_attr(PyPy, link_name = "PyPySequence_Type")]
+    pub static mut PySequence_Type: PyTypeObject;
```
birkenfeld, line 9: There is no `PySequence_Type` in the CPython API.  This is because "sequence" is not a concrete type, it's an abstract API that works with any type that has the corresponding `tp_as_sequence` members implemented.

birkenfeld (changes requested):

src/ffi/abstract_.rs
moriyoshi: Definitely that was my understanding and what I did here is due to tiredness... I'm going to come up with an alternative idea.  Thanks for pointing it out.

moriyoshi: I think I managed to fix this.

CHANGELOG.md
```diff
@@ -86,6 +86,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix segfault when dereferencing `ffi::PyDateTimeAPI` without the GIL. [#1563](https://github.com/PyO3/pyo3/pull/1563)
 - Fix memory leak when converting to u128 and i128. [#1638](https://github.com/PyO3/pyo3/pull/1638)
 - Fix segfault when calling `PyList::get_item` with negative indices. [#1668](https://github.com/PyO3/pyo3/pull/1668)
+- Fix inability to call `Py::as_ref()` against `PySequence`. [#1682](https://github.com/PyO3/pyo3/pull/1682)
```
messense, line 89: I think is should go to the `### Added` part since it's a new API not a bugfix.

moriyoshi: That's what I was wondering.  `PySequence` had already implemented `PyTryFrom` at the time I had to take a look into this, so I thought it simply didn't get modified to reflect the past changes to `Py`.

davidhewitt:
Sorry for being AWOL the past week, I've been swamped and not able to keep up with OSS. This is top of my list to review either tonight or tomorrow evening.

CHANGELOG.md
davidhewitt: Agreed this is an `Added` API.

src/instance.rs
```diff
@@ -128,6 +128,32 @@ where
...
+        unsafe { PyNativeType::unchecked_downcast(&*any) }
+    }
+
+    /// Similar to [`as_ref`](#method.as_ref), and also consumes this `Py` and registers the
+    /// Python object reference in PyO3's object storage. The reference count for the Python
+    /// object will not be decreased until the GIL lifetime ends.
+    pub fn into_ref(self, py: Python) -> &PySequence {
+        unsafe { py.from_owned_ptr(self.into_ptr()) }
+    }
+}
```
davidhewitt, line 155: I think it makes sense for this to live in `types/sequence.rs`.

src/types/sequence.rs
```diff
@@ -328,12 +316,25 @@ impl<'v> PyTryFrom<'v> for PySequence {
...
+    T: FromPyObject<'s>,
+{
+    let seq = <PySequence as PyTryFrom>::try_from(obj)?;
+    let mut v = Vec::with_capacity(seq.len().unwrap_or(0) as usize);
+    for item in seq.iter()? {
+        v.push(item?.extract::<T>()?);
+    }
+    Ok(v)
+}
+
```
davidhewitt, line 330: This function having moved doesn't seem relevant, can you please put it back in the original spot?

src/types/sequence.rs
```diff
@@ -674,4 +675,13 @@ mod test {
...
     }
+
+    #[test]
+    fn test_as_ref() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let seq: Py<PySequence> = PyList::empty(py).cast_as::<PySequence>().unwrap().into();
+        let seq_ref: &PySequence = seq.as_ref(py);
+        assert_eq!(seq_ref.len().unwrap(), 0);
+    }
```
davidhewitt, line 686: 👍 thanks for adding a test. Could you please also add one for `into_ref`?

davidhewitt (commented):
Thanks, overall I'm in favour of having these two APIs for `Py<PySequence>`. I think it's perfectly fine to add these in manually. Maybe in the future we'll have a trait which automatically produces this API again.

If it's not too much to ask, would you be willing to add the same for `Py<PyIterator>`, which is in almost exactly the same position?

src/types/sequence.rs
moriyoshi: This was absolutely unintentional.

moriyoshi:
> If it's not too much to ask, would you be willing to add the same for Py<PyIterator>, which is in almost exactly the same position?
That's also what I've been wondering.  Sounds nice.

moriyoshi:
I think I've done with them. Can you take a look?

src/types/any.rs
```diff
@@ -43,6 +44,15 @@ impl crate::AsPyPointer for PyAny {
...
 }
 
+impl crate::IntoPyPointer for PyAny {
+    #[inline]
+    fn into_ptr(self) -> *mut ffi::PyObject {
+        let pointer = self.as_ptr();
+        mem::forget(self);
+        pointer
+    }
+}
```
davidhewitt, line 54: I don't think this impl is necessary and it's also not sound - moving a PyAny like this onto the stack and then converting it to a pointer is just creating a dangling pointer. Please remove.

```suggestion
```

src/types/iterator.rs
```diff
@@ -213,4 +229,30 @@ mod tests {
         let iter: &PyIterator = PyIterator::try_from(obj.as_ref(py)).unwrap();
         assert_eq!(obj, iter.into());
     }
+
+    #[test]
+    fn test_as_ref() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
```
davidhewitt, line 236: We prefer `Python::with_gil` in new tests (`acquire_gil` is probably going to be deprecated or marked `unsafe` soon).

```suggestion
        Python::with_gil(|py| {
```

src/types/iterator.rs
```diff
@@ -213,4 +229,30 @@ mod tests {
...
+        let iter: Py<PyIterator> = PyList::empty(py).to_object(py).as_ref(py).iter().unwrap().into();
+        let mut iter_ref: &PyIterator = iter.as_ref(py);
+        assert!(matches!(iter_ref.next(), Option::<PyResult::<&PyAny>>::None));
+    }
+
+    #[test]
+    fn test_into_ref() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
```
davidhewitt, line 245: `Python::with_gil`

src/types/sequence.rs
```diff
@@ -674,4 +699,29 @@ mod test {
         let empty_seq = empty_list.cast_as::<PySequence>(py).unwrap();
         assert!(empty_seq.is_empty().unwrap());
     }
+
+    #[test]
+    fn test_as_ref() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
```
davidhewitt, line 706: `Python::with_gil`

src/types/sequence.rs
```diff
@@ -674,4 +699,29 @@ mod test {
...
+        let seq: Py<PySequence> = PyList::empty(py).cast_as::<PySequence>().unwrap().into();
+        let seq_ref: &PySequence = seq.as_ref(py);
+        assert_eq!(seq_ref.len().unwrap(), 0);
+    }
+
+    #[test]
+    fn test_into_ref() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
```
davidhewitt, line 715: `Python::with_gil`

src/types/iterator.rs
```diff
@@ -213,4 +229,30 @@ mod tests {
         let iter: &PyIterator = PyIterator::try_from(obj.as_ref(py)).unwrap();
         assert_eq!(obj, iter.into());
     }
+
+    #[test]
+    fn test_as_ref() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let iter: Py<PyIterator> = PyList::empty(py).to_object(py).as_ref(py).iter().unwrap().into();
```
davidhewitt, line 237: I think this can skip the `to_object(py)` part like this:

```suggestion
        let iter: Py<PyIterator> = PyAny::iter(PyList::empty(py)).unwrap().into();
```

src/types/iterator.rs
```diff
@@ -213,4 +229,30 @@ mod tests {
...
+        let l = PyList::empty(py).to_object(py);
+        let bare_ref = l.as_ref(py).iter().unwrap();
+        assert_eq!(bare_ref.get_refcnt(), 1);
+        let iter: Py<PyIterator> = bare_ref.into();
+        assert_eq!(bare_ref.get_refcnt(), 2);
+        Python::with_gil(|py| {
+            let mut iter_ref = iter.into_ref(py);
+            assert!(matches!(iter_ref.next(), Option::<PyResult::<&PyAny>>::None));
+            assert_eq!(iter_ref.get_refcnt(), 2);
+        });
```
davidhewitt, line 255: You already have the GIL, so no need to re-acquire

```suggestion
        let mut iter_ref = iter.into_ref(py);
        assert!(matches!(iter_ref.next(), Option::<PyResult::<&PyAny>>::None));
        assert_eq!(iter_ref.get_refcnt(), 2);
```

src/types/sequence.rs
```diff
@@ -674,4 +699,29 @@ mod test {
...
+        let bare_ref = PyList::empty(py).cast_as::<PySequence>().unwrap();
+        assert_eq!(bare_ref.get_refcnt(), 1);
+        let seq: Py<PySequence> = bare_ref.into();
+        assert_eq!(bare_ref.get_refcnt(), 2);
+        Python::with_gil(|py| {
+            let seq_ref = seq.into_ref(py);
+            assert_eq!(seq_ref.len().unwrap(), 0);
+            assert_eq!(seq_ref.get_refcnt(), 2);
+        });
```
davidhewitt, line 724: Similar

```suggestion
        let seq_ref = seq.into_ref(py);
        assert_eq!(seq_ref.len().unwrap(), 0);
        assert_eq!(seq_ref.get_refcnt(), 2);
```

davidhewitt (commented):
Thanks, it's looking almost there!

There's a new implementation for `IntoPyPointer` for `PyAny` which isn't sound and needs to be removed.

Also I made a few comments to tidy up the tests a little bit.

Finally, looks like you may need to run `cargo fmt` before you repush.

davidhewitt:
I'm going to finish this up and also add support for `PyMapping`; will push onto this branch. Thanks for getting it to this point.

Issue #1085, opened 2020-08-06
Iterator examples without `Clone`


davidhewitt:
As originally posted on Gitter by @microgravitas:
> Hi y'all! Can anyone point me to an example of how to implement an Python iterator which does not copy the whole underlying datastructure? All examples I could find used .clone(), something I'd like to avoid---if possible.

I've actually noticed this too about our iterator examples - it would be nice if we could find a way to make the `.clone()` unnecessary!

birkenfeld:
There is obviously no way when using `vec::IntoIter`, but that is also not really necessary.

The iterator type should just need a reference to the Py object, and an index.

davidhewitt:
Pretty much. I was thinking similar to the Python standard library types we might also want a guard to check that the base object is not modified during iteration. I think that'd have to be built into the base object, maybe as an additional field.

rapodaca:
To be clear, is this issue about:

- using a Python Iterator from Rust, or
- using a Rust Iterator from Python?

If this issue isn't about using Rust Iterator from Python, what would be the best way to do this? In other words, can anyone answer [this question](https://stackoverflow.com/questions/61955227/)?

From what I can tell, the problem is that iterator types returning references require a lifetime. And that lifetime propagates to any type that holds it. In PyO3, it's not possible for a struct to use a lifetime, as noted in [this issue](https://github.com/PyO3/pyo3/issues/502).

davidhewitt:
This is exactly about using a "Rust iterator from Python". The example I'm thinking of is [this one in the guide](https://pyo3.rs/v0.11.1/class.html#iterator-types), which uses `.clone().into_iter()`, which is essentially just as bad as the stackoverflow link you post.

rapodaca:
Yes, I've seen that one from the guide - thanks for the confirmation.

Is the problem that such a Rust Iterator to be used from Python (without `clone`) isn't possible, or that it's just not documented currently?

If it is possible, can you point me in the general direction of how to do it?

davidhewitt:
It should be possible! From the discussion between myself and birkenfeld above, I think this might be the pair of types that you'd want:

```
#[pyclass]
struct Container {
    vec: Vec<usize>,

    // if you want to reproduce Python's "X modified during iteration" behaviour, you'll probably need
    // a flag like this to:
    //  - set it false whenever iteration starts
    //  - set it true every time the vec is modified
    //  - check it every time Iter.__next__ is called
    has_been_modified: bool,
}

[...]

PR #1684, opened 2021-06-19
Add error message for trying to store references in pyclasses

mejrs:
People keep getting this wrong, so I figured an error for it is appropriate. Also note that this error can never reject a valid pyclass.

```rust
#[pyclass]
pub struct Foo<'a> {
    field: &'a PyAny,
}
```
creates the following error:

image

```rust
#[pyclass]
pub struct Bar {
    field: &'a PyAny,
}
```
creates the following error:

image

### DIFF
--- a/pyo3-macros-backend/src/diagnostics.rs
+++ b/pyo3-macros-backend/src/diagnostics.rs
@@ -1,0 +1,72 @@
+use quote::ToTokens;
+use syn::spanned::Spanned;
+
+/// Returns a `syn::Error` if the struct has any generic parameters,
+/// or its fields have any non-static lifetime parameters
+pub fn check_pyclass_generics_error(class: &syn::ItemStruct) -> Result<(), syn::Error> {
+    if !class.generics.params.is_empty() {
+        let mut base_error = syn::Error::new(
+            class.generics.params.span(),
+            "#[pyclass] cannot have generic parameters.",
+        );
+        if let Err(another) = check_pyclass_field_lifetimes(class) {
+            base_error.combine(another);
+        }
+
+        Err(base_error)
+    } else {
+        check_pyclass_field_lifetimes(class)
+    }
+}
+
+/// Returns a `syn::Error` if any field has any non-static lifetime parameters.
+fn check_pyclass_field_lifetimes(class: &syn::ItemStruct) -> Result<(), syn::Error> {
+    for field in &class.fields {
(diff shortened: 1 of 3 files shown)


messense:
Sometimes it's not a reference to a pyclasses but a Rust self-referencing struct, see https://github.com/PyO3/pyo3/issues/1089#issuecomment-794026403

mejrs:
This should only trigger if the referenced object looks like a Python object, by checking whether the name looks like `Py....`.  Obviously that is inherently limited, as macros can't resolve the actual type.

davidhewitt:
It's nice to have this error, however at the same time it scares be a bit because matching on the name is kinda brittle. Maybe we have a self-referencing struct with a `&'a PyMyCustomSelfReferencingThing` field? Suddenly this error blocks the user doing it.

I've been meaning for a long time to experiment with self-referencing struct to write a doc for #1085 . Perhaps we can experiment with that as part of any work we do here?

Also, I wonder if the error should only trigger when the class has a lifetime parameter, as that's a known case which we already don't allow to compile? (So an incorrect heuristic wouldn't matter too much.)

mejrs:
> It's nice to have this error, however at the same time it scares me a bit because matching on the name is kinda brittle. Maybe we have a self-referencing struct with a `&'a PyMyCustomSelfReferencingThing` field? Suddenly this error blocks the user doing it.

Building a self referential struct requires unsafe code and that reference would probably be a pointer, in which case this error won't pop up.

> I've been meaning for a long time to experiment with self-referencing struct to write a doc for #1085 . Perhaps we can experiment with that as part of any work we do here?

Hmm, I got an idea for that actually. I don't think it is that complicated if you're willing to sacrifice a little bit of ergonomics.

> Also, I wonder if the error should only trigger when the class has a lifetime parameter, as that's a known case which we already don't allow to compile? (So an incorrect heuristic wouldn't matter too much.)

The other way (a struct without lifetime parameters but with a field that has one) shouldn't compile either, right? Unless I'm missing something this shouldn't error on a valid program.

davidhewitt:
There are proc macros to make self-referencing types safer, such as https://github.com/joshua-maros/ouroboros

Depending on the order `#[pyclass]` and `#[self_referencing]` is expanded we might have lifetimes on the fields but not on the type.

I know @messense is doing this in the wild, so erroring on this case would definitely break existing users!

mejrs:
I've thought some on this and I think it's a bad idea to do this without any real support for it (there's no type resolution in macros, for example).

Issue #1675, opened 2021-06-11
_PyFrameEvalFunction has different definitions in Python 3.8 and Python 3.9


itamarst:
### 🌍 Environment

This `pyo3` 0.13.2.

### 💥 Reproducing

Per [The docs](https://docs.rs/pyo3/0.13.2/pyo3/ffi/type._PyFrameEvalFunction.html), `_PyFrameEvalFunction` is `extern "C" fn(_: *mut PyFrameObject, _: c_int) -> *mut PyObject;` This is in fact the case in Python 3.8; however, it's worth noting that this is private in Python 3.8.

In Python 3.9 this was made public, alongside some related APIs for PEP 523 that aren't in PyO3's FFI (e.g. https://docs.python.org/3/c-api/init.html#c._PyInterpreterState_SetEvalFrameFunc).

However, the signature in Python 3.9 has changed, it's now `PyObject* (*_PyFrameEvalFunction)(PyThreadState *tstate, PyFrameObject *frame, int throwflag)`.

### Bigger picture

As someone currently munging the internals of Python 3 for profiling purposes, I want to use the private API on Python 3.8, and the public API on Python 3.9. The current way `pyo3::ffi` works in pyo3 makes this difficult insofar as it's exposing types that change over time across Python releases, so I'm going to have to make my own types... it would be handy if PyO3 had some way for me to detect at compile time which Python API version was available.

Alternatively, perhaps PyO3 should have different versions of the low-level `pyo3::ffi` module depending which version of Python you're building against? Which I imagine would be some work.

### Smaller picture

On Python 3.9 or later `pyo3::ffi::_PyFrameEvalFunction` definition is wrong, and given that it's private before that, probably the newer signature should be there if there's only going to be one `pyo3:ffi` module variant for all Python versions.

### Final thoughts

This is an obscure API for a rare use case. However, the bigger picture issue of `pyo3:ffi` not matching the Python API version that code is compiling against is perhaps more significant.

itamarst:
Ah, I found the `cfg()` options for Python API versions: https://github.com/PyO3/pyo3/blob/97d6f15b32c3090498911b8bff2edcef496bf228/pyo3-build-config/src/impl_.rs#L72

mejrs:
Thanks for the bug report :)

> Ah, I found the cfg() options for Python API versions: 

[...]

PR #1687, opened 2021-06-23
implement ffi/cpython/pystate

mejrs:
Per #1675 

Also:
-  adds a bit more documentation to /ffi
- makes the cpython modules `pub(crate)` to avoid double glob imports (is this OK?)


src/ffi/pystate.rs
```diff
@@ -7,10 +8,29 @@ use std::os::raw::{c_int, c_long};
...
 #[repr(C)]
-#[derive(Copy, Clone)]
+#[derive(Debug, Copy, Clone)]
 pub struct PyInterpreterState {
+    /// Not actually a public field
     pub ob_base: PyObject,
+
+    /// Not actually a public field
+    /// Also, it has an additional argument in 3.9
+    #[cfg(not(Py_3_9))]
```
mejrs, line 18: I'm not sure that this is a good way to handle this, any suggestions?

src/ffi/pystate.rs
```diff
@@ -7,10 +8,29 @@ use std::os::raw::{c_int, c_long};
 pub const MAX_CO_EXTRA_USERS: c_int = 255;
 
 #[repr(C)]
-#[derive(Copy, Clone)]
+#[derive(Debug, Copy, Clone)]
 pub struct PyInterpreterState {
```
davidhewitt, line 12: Hmm, this looks dangerously broken, because it doesn't match at all the full contents of the corresponding struct in the Python headers: https://github.com/python/cpython/blob/af5fb6706219d7949c1db5c9f2b7da53198123f3/Include/internal/pycore_pystate.h#L67

Appreciate that this existed before you wrote this PR, but I don't think editing this struct in this PR achieves much.

As it's an internal type (and probably very different on PyPy),I think we should just replace with an opaque type:

```
opaque_struct!(PyInterpreterState);
```

davidhewitt (commented):
I don't think we should be restricting the `ffi::cpython` modules to `pub(crate)`. They're already public in the CPython API, so I think it's correct for them to be `pub`, not `pub(crate)`.

Also, I think that we should probably remove the `PyInterpreterState` struct as that looks scary - see other comment.

src/ffi/pystate.rs
mejrs: Done.

mejrs:
> I don't think we should be restricting the `ffi::cpython` modules to `pub(crate)`. They're already public in the CPython API, so I think it's correct for them to be `pub`, not `pub(crate)`.

Their contents would still be public, just not the modules themselves.

Currently when you search for something that is defined in the cpython module this is what you see, because it's glob imported twice:

image

davidhewitt:
> Their contents would still be public, just not the modules themselves.
>
> Currently when you search for something that is defined in the cpython module this is what you see, because it's glob imported twice:

👍 I'm persuaded; also the modules in `pyo3::ffi` are not `pub`, so this is consistent.

davidhewitt:
Happy to merge this once conflict caused by #1692 is resolved (sorry!)

PR #1688, opened 2021-06-23
Update class.md - add example of `new` returning a `PyResult`

aviramha:
It took me some time to realize how to return error in the new class API. I think it's better to have explicit documentation to save time for others :)

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -169,7 +169,7 @@ impl MyClass {
 }
 ```
 
-Alternatively, if your `new` method might throw errors you can return `PyResult<Self>`.
+Alternatively, if your `new` method may fail you can return `PyResult<Self>`.
 ```rust
 # use pyo3::prelude::*;
 #[pyclass]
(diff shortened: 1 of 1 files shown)


messense:
See also https://pyo3.rs/v0.13.2/exception.html#handling-rust-errors

Eric-Arellano (commented):
Note that this is already covered in the section right below: https://github.com/PyO3/pyo3/blob/main/guide/src/class.md#return-type

It's a bit easy to miss, but it has 3 hashtags compared to 2, meaning it's a subsection of the constructor section.

davidhewitt (approved):
I still think it's fine to explicitly call this out.

guide/src/class.md
```diff
@@ -169,6 +169,23 @@ impl MyClass {
 }
 ```
 
+Alternatively, if your `new` method might throw errors you can return `PyResult<Self>`.
```
Eric-Arellano, line 172: To be consistent with the below table, maybe word this as "may fail" instead of "might throw errors"?

aviramha: Done

PR #1689, opened 2021-06-23
Add a badge and link to benchmark tracking page

messense:
[Rendered](https://github.com/PyO3/pyo3/blob/6504f2598b5f06915597634da34df8b32d04ff60/README.md)


davidhewitt (approved):
Imo that benchmark page is not _super_ helpful for users, as it's very noisy and also a bit unclear to people who aren't familiar with the project as to what we're actually benchmarking.

However overall it's still nice to show the badge to give easy access to them!

PR #1692, opened 2021-06-23
Make the `Py_tracefunc` parameters optional

GabrielMajeri:
The `PyEval_SetProfile`/`SetTrace` methods [support receiving `NULL`](https://docs.python.org/3/c-api/init.html#c.PyEval_SetProfile) for the `Py_tracefunc` parameter, in order to disable profiling/tracing. However, the only way to safely pass `NULL` as a function pointer in Rust is by wrapping it in an `Option<>`.

This PR updates the respective functions to make it easier for developers to disable profiling/tracing hooks, without having to resort to `transmute` or `zeroed`.


davidhewitt (approved):
👍 thanks!

PR #1693, opened 2021-06-23
Rewrite `module.md` for clarity and add tip on code organization

Eric-Arellano:
This adds a recommendation on how to split FFI code over multiple Rust modules, which we have found useful at [Pants](https://github.com/pantsbuild/pants/tree/main/src/rust/engine/engine_pyo3).

It also rewrites some of `module.md` and a bit of `function.md`:

- Consistently use a simpler example function `double`, rather than `sum_as_string`.
- Use `#[pyfunction]` instead of `#[pyfn]` in the module example for simplicity.
- Add a code example of renaming a `#[pymodule]` with `#[pyo3(name = "..")]`.
- Rewrite submodule hierarchy section, including adding the disclaimer about packages.


guide/src/function.md
```diff
@@ -14,52 +14,48 @@ fn double(x: usize) -> usize {
 }
 
 #[pymodule]
-fn module_with_functions(py: Python, m: &PyModule) -> PyResult<()> {
+fn my_extension(py: Python, m: &PyModule) -> PyResult<()> {
     m.add_function(wrap_pyfunction!(double, m)?)?;
     Ok(())
 }
-
-# fn main() {}
```
Eric-Arellano, line 22: This isn't necessary as of version 1.34 of Rust: https://doc.rust-lang.org/rustdoc/documentation-tests.html#using--in-doc-tests.

guide/src/module.md
```diff
@@ -1,44 +1,49 @@
 # Python Modules
 
-You can create a module as follows:
+You can create a module using `#[pymodule]`:
```
Eric-Arellano, line 3: This is something we picked up on with docs for Pants: for bigger code blocks, preview what the main thing to look for is.

guide/src/module.md
```diff
@@ -1,44 +1,49 @@
...
 
 ```rust
 use pyo3::prelude::*;
+use pyo3::wrap_pyfunction;
 
-// add bindings to the generated Python module
-// N.B: "rust2py" must be the name of the `.so` or `.pyd` file.
```
Eric-Arellano, line 9: This is said later in the paragraph. It's repetitive and distracting to say here too.

guide/src/module.md
```diff
@@ -1,44 +1,49 @@
...
+    x * 2
+}
 
 /// This module is implemented in Rust.
 #[pymodule]
-fn rust2py(py: Python, m: &PyModule) -> PyResult<()> {
-    // PyO3 aware function. All of our Python interfaces could be declared in a separate module.
-    // Note that the `#[pyfn()]` annotation automatically converts the arguments from
-    // Python objects to Rust values, and the Rust return value back into a Python object.
-    // The `_py` argument represents that we're holding the GIL.
```
Eric-Arellano, line 17: That's explained in function.md. This page is instead focused on modules, and there's value imo in keeping this example short and focused.

guide/src/module.md
```diff
@@ -1,44 +1,49 @@
...
-    #[pyfn(m)]
```
Eric-Arellano, line 18: I personally prefer the `#[pyfunction]` approach and think it makes things easier to understand. Better separation of concerns, such that the `#[pymodule]` function is solely for FFI registration and has no business logic.

I think the example is easier to follow by using `#[pyfunction]` - `function.md` still demos `#[pyfn]`.

guide/src/module.md
```diff
@@ -48,53 +53,110 @@ To import the module, either:
...
 ```
 
-Which means that the above Python code will print `This module is implemented in Rust.`.
+## Organizing your module registration code
+
+As your Python module grows, it can be helpful to split your Rust code into several Rust modules to
+keep your code readable.
+
+You can pass references to the `PyModule` so that each 
+module registers its own FFI code, which can give your project better
```
Eric-Arellano, line 73: Is FFI a safe word to use here? Generally acronyms should be avoided in docs.

guide/src/module.md
```diff
@@ -48,53 +53,110 @@ To import the module, either:
...
 
-Which means that the above Python code will print `This module is implemented in Rust.`.
+## Organizing your module registration code
+
+As your Python module grows, it can be helpful to split your Rust code into several Rust modules to
+keep your code readable.
+
+You can pass references to the `PyModule` so that each 
+module registers its own FFI code, which can give your project better 
+[encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)). For example:
```
Eric-Arellano, line 74: I don't want to assume every future PyO3 user knows what encapsulation is and its benefits, hence the link.

guide/src/module.md
```diff
@@ -48,53 +53,110 @@ To import the module, either:
...
+
+```rust,ignore
```
Eric-Arellano, line 76: I'd like to not ignore this, but don't know how to simulate multiple distinct files in a doc test.

guide/src/module.md
```diff
@@ -48,53 +53,110 @@ To import the module, either:
...
+    "linux".to_owned()
+}
+```
 
-## Modules as objects
+## Python submodules
 
-In Python, modules are first class objects. This means that you can store them as values or add them to
-dicts or other modules:
+You can create a module hierarchy within a single extension module by using `PyModule.add_submodule()`.
```
Eric-Arellano, line 122: Should this cross-reference with Rustdoc magic? I'm not sure how to do that.

Eric-Arellano (commented):
I hope some of these suggestions are useful :) I love that y'all have a guide, that was a big draw to PyO3 for me personally with the Pants project.

guide/src/module.md
mejrs: You can do https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example in rustdoc. I'm not sure if this also works in mdbook (?)

guide/src/module.md
```diff
@@ -48,53 +53,110 @@ To import the module, either:
...
 ```python
-import rust2py
+import my_extension
 
-print(rust2py.__doc__)
+print(my_extension.__doc__)
 ```
 
-Which means that the above Python code will print `This module is implemented in Rust.`.
+## Organizing your module registration code
```
mejrs, line 67: I would probably avoid this indirection and just add the classes and functions directly:

```rust
#[pymodule]
fn my_extension(py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<dirutil::SomeClass>()?;
    m.add_function(wrap_pyfunction!(osutil::determine_current_os, m)?)?;
    Ok(())
}
```

I guess this makes more sense if you have a big hierarchy of modules with register functions that each import other registers? But at that point you should probably just organize them in submodules.

mejrs (commented):
Hey, thanks for this :), this looks great.

guide/src/module.md
Eric-Arellano: That's great! I was going to open an issue proposing that very thing :) I think there's value in having a consolidate way to do things. 

Would you like me to open an issue for this? I'd be happy to implement.

guide/src/module.md
Eric-Arellano: As in, rustdoc has a way to refer to other parts of the code base and then it will hyperlink for you. I'm not sure how to do it, but I know it exists. 

I'm wondering here if you happen to know if it's possible to cross reference from the book to the API docs?

mejrs: You can just hyperlink to the latest versions with e.g.:
- the pyo3.rs documentation: `https://pyo3.rs/main/doc/pyo3/types/struct.PyModule.html#method.add_submodule`
- docs.rs: https://docs.rs/pyo3/latest/pyo3/types/struct.PyModule.html#method.add_submodule
- 
I'm actually not sure which one you should prefer...fwiw pyo3.rs reflects the current main branch, docs.rs the latest release.

guide/src/module.md
mejrs: Yeah, I don't like having two slightly different ways to do the same thing.

As for an issue, there is one: #694

davidhewitt: Completely agree that `#[pyfn]` should go. The syntax will at least be _almost_ the same in PyO3 0.14.

Regarding #694 - the vision was that rather than remove `#[pyfn]`, we would first create a new `#[pymodule]` syntax which is used on `mod my_module` instead of `fn my_module`. Then we could deprecate the `#[pymodule]` function syntax and `#[pyfn]` together.

guide/src/module.md
davidhewitt: Yes I think FFI is fairly prevalent in PyO3's guide. TBH we could add a glossary if we really were worried, to include other terms like GIL which we also use a lot.

guide/src/module.md
davidhewitt: It does. You could use hidden `# mod lib { }` etc to break the code into multiple "files" (modules).

guide/src/module.md
davidhewitt: The `mdbook` setup has a plugin to substitute in the appropriate link, the following should do it:

`{#PYO3_DOCS_URL}}/pyo3/prelude/struct.PyModule.html#method.add_submodule`

davidhewitt (commented):
Looks great, many thanks for doing this! A few quick comments on some of the conversations...

Eric-Arellano:
FYI I'm traveling next few days on a road trip, so I likely won't get to this till I'm back. 

Thanks for the review though!

guide/src/module.md
Eric-Arellano: I remember now why I originally came up with this pattern of passing `&PyModule` around:

```rust
error: no rules expected the token `::`
  --> src/lib.rs:591:43
   |
14 |     m.add_function(wrap_pyfunction!(osutil::determine_current_os, m)?)?;
   |                                           ^^ no rules expected this token in macro call
```

Is that expected? Any known workarounds other than what I'm suggesting?

Eric-Arellano:
> FYI I'm traveling next few days on a road trip, so I likely won't get to this till I'm back.

Took me longer to get back to this than I hoped, but updating based on feedback now :) Congrats on 0.14 release btw!

guide/src/module.md
Eric-Arellano: Hm my next idea didn't work either to leverage `use`:

```rust
// src/lib.rs
use pyo3::prelude::*;
use osutil::determine_current_os; 

#[pymodule]
fn my_extension(py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<dirutil::SomeClass>()?;
    m.add_function(wrap_pyfunction!(determine_current_os, m)?)?;
    Ok(())
}

// src/dirutil.rs
# mod dirutil { 
use pyo3::prelude::*;

#[pyclass]
pub(crate) struct SomeClass {
    x: usize,
}
# }

// src/osutil.rs
# mod osutil { 
use pyo3::prelude::*;

#[pyfunction]
pub(crate) fn determine_current_os() -> String {
    "linux".to_owned()
}  
# }
```

```
❯ cargo test --doc -- doc_test::guide_module_md
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/eric/code/pyo3/examples/pyo3-pytests/Cargo.toml
workspace: /Users/eric/code/pyo3/Cargo.toml
    Finished test [unoptimized + debuginfo] target(s) in 0.10s
   Doc-tests pyo3

running 5 tests
test src/lib.rs - doc_test::guide_module_md (line 620) ... ignored
test src/lib.rs - doc_test::guide_module_md (line 578) ... FAILED
test src/lib.rs - doc_test::guide_module_md (line 535) ... ok
test src/lib.rs - doc_test::guide_module_md (line 513) ... ok
test src/lib.rs - doc_test::guide_module_md (line 670) ... ok

failures:

---- src/lib.rs - doc_test::guide_module_md (line 578) stdout ----
error[E0425]: cannot find value `__pyo3_get_function_determine_current_os` in this scope
  --> src/lib.rs:586:20
   |
11 |     m.add_function(wrap_pyfunction!(determine_current_os, m)?)?;
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

```

davidhewitt: Yeah, you're running into https://github.com/PyO3/pyo3/issues/1709

For now you can work around this by writing `use osutil::*;`

(As stated in the issue above, I hope we can fix it in a nicer way at the end of the year when we do a dependency update release.)

davidhewitt:
Thanks and no worries, hope you had a great trip! Feel free to ping when it's ready for re-review.

Eric-Arellano:
> Thanks and no worries, hope you had a great trip!

Thanks! Washington was magical, my first time spending more than a day there and I loved the outdoors :) 

> Feel free to ping when it's ready for re-review.

All ready! I previewed it using the instructions you sent on Gitter, and it looks good from my POV. I'm not sure I got the `{#PYO3_DOCS_URL}}` thing right though - the link didn't work when I was previewing locally.

guide/src/module.md
```diff
@@ -48,53 +51,133 @@ To import the module, either:
...
 #[pyfunction]
-fn subfunction() -> String {
-    "Subfunction".to_string()
+fn determine_current_os() -> String {
+    "linux".to_owned()
 }
+# }
+```
+
+Another workaround for splitting FFI code across multiple modules is to add `use module::*`, like this:
```
davidhewitt, line 117: It might be worth adding a quick note about #1709, both so users can understand why the glob is suggested (which is usually seen as an antipattern) and as a reminder for us to improve it when we implement a better solution.

guide/src/module.md
```diff
@@ -48,53 +51,133 @@ To import the module, either:
...
+pub(crate) fn determine_current_os() -> String {
+    "linux".to_owned()
+}
+# }
+```
+
+## Python submodules
+
+You can create a module hierarchy within a single extension module by using 
+[`PyModule.add_submodule()`]({#PYO3_DOCS_URL}}/pyo3/prelude/struct.PyModule.html#method.add_submodule).
```
davidhewitt, line 144: I think missing a brace?

```suggestion
[`PyModule.add_submodule()`]({{#PYO3_DOCS_URL}}/pyo3/prelude/struct.PyModule.html#method.add_submodule).
```

davidhewitt (approved):
Looks really great! Just a few final suggestions.

guide/src/module.md
davidhewitt: EDIT: oh that's my fault from originally typing the URL wrong above. Sorry! 😬

davidhewitt:
Perfect! Thanks again 🚀

PR #1694, opened 2021-06-24
automagically update pyo3 version in toml examples

mejrs:
I figured out how to automagically update the pyo3 version used in doc examples: see https://github.com/rust-lang/rust/issues/82768#issuecomment-803935643


davidhewitt:
👍 nicely done, and one day we will be able to simplify this after it comes to stable!

PR #1695, opened 2021-06-24
Add `wrap_pyfunction` macro to prelude

### LINKED ISSUE #1691: Move `wrap_pyfunction` into prelude?
Hello! We're using `wrap_pyfunction` a lot in pantsbuild/pants, and found we prefer `#[pyfunction]` over `#[pyfn]` to keep our `#[pymodule]` block clean. (I'm submitting a docs PR on other things we've found to help.)

I'm wondering if it makes sense to move the macro into the prelude, given how common this type seems to be, for fewer imports? Maybe the `create_exception` macro too?

I'm happy to implement if you think so!

messense:
Closes #1691


Eric-Arellano (approved):
Awesome, thanks!

davidhewitt (approved):
Thanks!

PR #1696, opened 2021-06-24
Unwrap `syn::*::Group` in a few places

scalexm:
Feel free to push to this branch with similar fixes.

### DIFF
--- a/pyo3-macros-backend/src/params.rs
+++ b/pyo3-macros-backend/src/params.rs
@@ -4,6 +4,7 @@ use crate::{
     attributes::FromPyWithAttribute,
     method::{FnArg, FnSpec},
     pyfunction::Argument,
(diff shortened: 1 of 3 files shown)


davidhewitt:
👍 thanks, looks good.

I think to make coverage happy it would be great to add some test code which checks these functions (e.g. executes them). It can be pretty simple, like just call them once.

scalexm:
You mean a `#[test]` function which calls `issue_1696`? The code coverage check seems to be complaining about `issue_1696`, which only purpose is to check that code generated by the proc-macro compiles.

davidhewitt:
That's what I was thinking, but actually maybe it's better to just ignore coverage of test files. I've opened #1706.

And for now, I'll merge this. Thanks!

PR #1697, opened 2021-06-24
Use explicit lifetimes in proc-macro expansion

scalexm:
Otherwise this triggers `#[deny(rust_2018_idioms)]` in projects using it.


messense (approved):
Thanks!

PR #1699, opened 2021-06-27
Added `size_hint` impls for `{PyDict,PyList,PySet,PyTuple}Iterator`s

ohadravid:
Hi!

I've added `size_hint` impls for the iterators which can support it (while defaulting to 0 on Python-related errors).
Should make `some_py_dict.iter().map(...).collect()` faster by avoiding allocations!


birkenfeld:
I think this needs to take into account the current position of the iterator, since size_hint is supposed to return the remaining length, not the total length.

davidhewitt:
As well as the above, it would be awesome to add tests for these ☺️

ohadravid:
@birkenfeld 🤦‍♂️ right! fixed.
@davidhewitt added the tests, and indeed `PyTuple` moves the `self.pos` to `len() + 1` in the end, so I got an ugly overflow.

davidhewitt (approved):
Thanks, this is great!

PR #1702, opened 2021-06-27
Make `ToPyObject` impl for `HashSet` accept non-default hashers.

ohadravid:
I had some code like 
```
let s = HashSet<u32>::new();
s.to_object(py)
```
I switched to `HashSet<_, FxBuildHasher>`, and it stopped working.
So this PR changes the `ToPyObject` impl for `HashSet` to accept non-default hashers.

I didn't add a test since I saw that the other converter impls (for example `HashMap`, `into_py`) don't have tests for non-default `BuildHasher`s, but let me know if you want me to add them anyway :)


davidhewitt (approved):
Thanks for the fix! As this is just adding a generic parameter I'm fine to skip adding a test.

PR #1703, opened 2021-06-27
Shortcut macros for collections

daniil-konovalenko:
This is an implementation of #1463
It currently provides the following macros:
- `py_list!(py, [1, 2, 3]) → &PyList`
- `py_tuple!(py, (1, 2, 3)) → &PyTuple`
- `py_dict!(py, {"key": value}) → PyResult<&PyDict>`
- `py_set!(py, {"item1", "item2", 1, 2, 3}) → PyResult<&PySet>`
- `py_frozenset!(py, {"item1", "item2", 1, 2, 3}) → PyResult<&PyFrozenSet>`

`py_dict!` only supports literals as keys because we use colons `:` as key-value separators, and colons are not allowed after `expr` fragments. If we would like to use expressions as keys, we have to use `=>` as a separator.

TODO:
- [ ] `:` vs `=>` for key-value separator
- [x] proc macro for py_dict!
- [x] PyDict::from_sequence doesn't exist on PyPy
 - [ ] docs to all new macros and details in the guide
 - [ ] more tests probably

### DIFF
--- a/pyo3-macros-backend/src/dict.rs
+++ b/pyo3-macros-backend/src/dict.rs
@@ -15,7 +15,6 @@ pub struct PyDictLiteral {
 #[derive(Debug)]
 pub struct KeyValue {
     key: syn::Expr,
-    sep: Token![:],
     value: syn::Expr,
 }
 
@@ -40,14 +39,10 @@ impl Parse for Key {
 impl Parse for KeyValue {
     fn parse(input: ParseStream) -> syn::Result<Self> {
         let key: Key = input.parse()?;
-        let sep: Token![:] = input.parse()?;
+        let _sep: Token![:] = input.parse()?;
         let value: syn::Expr = input.parse()?;
 
-        Ok(Self {
-            key: key.0,
-            sep,
-            value,
-        })
+        Ok(Self { key: key.0, value })
     }
 }
 
--- a/pyo3-macros-backend/src/dict.rs
+++ b/pyo3-macros-backend/src/dict.rs
@@ -1,0 +1,78 @@
+use proc_macro2::{Ident, TokenStream, TokenTree};
+use quote::{quote, ToTokens};
+use std::iter::FromIterator;
+use syn::parse::{Parse, ParseBuffer, ParseStream};
+use syn::punctuated::Punctuated;
+use syn::Token;
+use syn::{braced, Expr};
+
+#[derive(Debug)]
+pub struct PyDictLiteral {
+    pub py: Ident,
+    pub items: Vec<KeyValue>,
+}
+
+#[derive(Debug)]
+pub struct KeyValue {
+    key: syn::Expr,
+    sep: Token![:],
+    value: syn::Expr,
+}
+
+#[derive(Debug)]
+struct Key(syn::Expr);
+
+impl Parse for Key {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let mut tokens = vec![];
+
+        while !input.peek(Token![:]) || input.peek(Token![::]) {
+            let tt = input.parse::<TokenTree>()?;
+            tokens.push(tt);
+        }
+        let stream = TokenStream::from_iter(tokens.into_iter());
+
+        let expr = syn::parse2::<Expr>(stream)?;
+        Ok(Self(expr))
+    }
+}
+
+impl Parse for KeyValue {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let key: Key = input.parse()?;
+        let sep: Token![:] = input.parse()?;
+        let value: syn::Expr = input.parse()?;
+
+        Ok(Self {
+            key: key.0,
+            sep,
+            value,
+        })
+    }
+}
+
+impl Parse for PyDictLiteral {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let py: Ident = input.parse()?;
+        let _arrow: Token![=>] = input.parse()?;
+
+        let body: ParseBuffer;
+        braced!(body in input);
+
+        let items: Punctuated<KeyValue, Token![,]> = Punctuated::parse_terminated(&body)?;
+
+        Ok(Self {
+            py,
+            items: items.into_iter().collect(),
+        })
+    }
+}
+
+impl ToTokens for KeyValue {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let key = &self.key;
+        let value = &self.value;
+        let ts = quote! {(#key, #value)};
+        tokens.extend(ts);
+    }
+}
--- a/pyo3-macros-backend/src/lib.rs
+++ b/pyo3-macros-backend/src/lib.rs
@@ -11,6 +11,7 @@ mod utils;
 mod attributes;
 mod defs;
 mod deprecations;
+mod dict;
 mod from_pyobject;
 mod konst;
 mod method;
@@ -23,6 +24,7 @@ mod pyimpl;
 mod pymethod;
 mod pyproto;
 
+pub use dict::PyDictLiteral;
 pub use from_pyobject::build_derive_from_pyobject;
 pub use module::{process_functions_in_module, py_init, PyModuleOptions};
 pub use pyclass::{build_py_class, build_py_enum, PyClassArgs};
--- a/pyo3-macros/src/lib.rs
+++ b/pyo3-macros/src/lib.rs
@@ -204,9 +204,11 @@ pub fn py_dict(input: proc_macro::TokenStream) -> TokenStream {
     let PyDictLiteral { items, py } = parse_macro_input!(input as PyDictLiteral);
     let stream = quote! {
             (|| {
-                let dict = ::pyo3::types::PyDict::new(#py);
+                use pyo3 as _pyo3;
+
+                let dict = _pyo3::types::PyDict::new(#py);
                 #(dict.set_item#items?;)*
-                ::pyo3::prelude::PyResult::Ok(dict)
+                _pyo3::PyResult::Ok(dict)
             })()
     };
     stream.into()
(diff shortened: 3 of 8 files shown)


davidhewitt:
Very cool, thanks for working on this! Let's take some time to get this right; we can merge it after the 0.14 release so that there's plenty of time to make the right compromises.

> py_dict! only supports literals as keys because we use colons : as key-value separators, and colons are not allowed after expr fragments. If we would like to use expressions as keys, we have to use => as a separator.

Ah, that's really frustrating. There is precedent for `=>` from the `maplit::hashmap!` macro, though it's a bit of a shame to deviate from the python syntax. Perhaps as well as literals, we could allow idents? Or I wonder if we could use a proc macro for `py_dict!` to allow us to have exactly the syntax we want. (proc macro in expression position is possible from Rust 1.42 if I recall correctly; I think it's fine if these macros don't work on MSRV though.)

I'd be interested in hearing some other maintainer's opinions on the design considerations here, as the way they're implemented at the moment is basically exactly the way I suggested in #1463.

One final syntax idea I have is that we could consider using `py =>` at the beginning, e.g.

```rust
py_list!(py => [1, 2, 3]);
```

I saw this in the `quote::quote_spanned!` macro and quite liked it. Wonder what you think of it?

daniil-konovalenko:
Thank you for such a quick reaction!

> Let's take some time to get this right

Absolutely! Let's try to make this as pleasant to use as possible.

> Perhaps as well as literals, we could allow idents?

Sure, idents shouldn't be a problem. A proc macro sounds even better, I did not know it was an option here, thanks for the hint! I will see what we can do with it.

> I'd be interested in hearing some other maintainer's opinions on the design considerations here, as the way they're implemented at the moment is basically exactly the way I suggested in #1463.

Me too!  
One thing that bugs me at the moment is that deeply nested collections look really convoluted because of all the `py`'s and result handling:

```rust
let my_complex_dict = py_dict!(py, {
    "username": "admin",
    "permissions": py_set!(py, {"read", "write"}).unwrap(),
    "default_args": py_tuple!(py, (py_list!(py, [1, "order_id", py_dict!(py, {"meta": py_list!(py, ["abcd", "efg"])}).unwrap()])))
}).unwrap();
```
compared to python
```python
my_complex_dict = {
    'default_args': (
        [1, 'order_id', {'meta': ['abcd', 'efg']}],
    ),
    'permissions': {'read', 'write'},
    'username': 'admin',
}
```
I wonder if we can do better here.


> One final syntax idea I have is that we could consider using py => at the beginning, e.g.
`py_list!(py => [1, 2, 3]);`

I like it! It makes for a nice distinction between `py` and the actual collection items.

davidhewitt:
> One thing that bugs me at the moment is that deeply nested collections look really convoluted because of all the py's and result handling:

Hmm, that's true. There could be some kind of `py_collection!` macro which had all of the syntaxes in one, though I'm not sure I love that. I'd be open to hearing suggestions on this.

mejrs:
I love this, thanks ❤️

> One thing that bugs me at the moment is that deeply nested collections look really convoluted because of all the py's and result handling:

I think this is also because nested objects *are* convoluted.

> Hmm, that's true. There could be some kind of py_collection! macro which had all of the syntaxes in one, though I'm not sure I love that. 

This sounds pretty complicated to use. I can't say I've ever had a good time making/manipulating complex structures in Python. So I don't think this is something we should encourage.

> I'd be open to hearing suggestions on this.

Something like this could work for that (although this would be out of the scope of this pr):

```rust
pub trait IntoNativePythonCollections {
    fn into_collection(&self, py: Python) -> PyResult<PyAny>;
}

#[derive(IntoNativePythonCollections)]
pub struct Meta {
    inner: Vec<String>,
}

#[derive(IntoNativePythonCollections)]
pub struct MyComplexType {
    default_args: (usize, String, Meta),
    permissions: HashSet<String>,
    username: String,
}

fn main() {
    Python::with_gil(|py| {
        let s = MyComplexType { /* ... */ };
        let nested: &PyAny = s.into_collection(py).unwrap();
    });
}
```

davidhewitt:
> People can just define nested Rust structs and derive ToPyObject for it? Is this something we could make a derive macro for?

Yes definitely. I think at the time that `#[derive(FromPyObject)]` was implemented there were some design questions about the to-python direction, but it's very much something I'd support having.

I think you're right that that would be a cleaner way to create a deeply-nested object.

> result handling

Just noticed this item. Do we expect many errors here? It might be nice to panic on invalid input to avoid the error handling, but it depends on how easy it is to write invalid code.

daniil-konovalenko:
> I think this is also because nested objects are convoluted.

I agree, creating deeply nested objects is probably not the intended use of these macros. I really like the `#[derive(IntoPyObject)]` idea though.

> Do we expect many errors here?

I was thinking about passing unhashable objects as dict keys or set elements. I'm not sure how common that might be, so I made `py_set!` and `py_dict!` return `Result` for now. Perhaps we could panic on invalid input by default but also provide an alternative that returns Result instead, like `try_py_dict!`.

davidhewitt:
> I was thinking about passing unhashable objects as dict keys or set elements.

That's a fair point. I still think that even if users make an error here, it's pretty much hard-coded into the program with the macro, so it'll be fixed pretty quick after they see a panic.

> I made `py_set!` and `py_dict!` return Result for now

To avoid suprising users I would suggest making either all of the macros of none of them return `Result`, so that they're not repeatedly having to remember the special cases.

I'm not particularly convinced we would need non-panicking `try_` variants - when the user is unsure about the input, they can just use normal methods.

davidhewitt:
Also - prior art on complex "collections" is the `serde_json::json!` macro.

Though there is also the `inline_python` crate which exposes the full Python syntax, which is probably a better way to do such a thing than us trying to make a `py_collection!` macro.

birkenfeld:
I'm in favor or the non-panicking variant - the case of unhashable objects is one of those instances where everything works perfectly in tests, and then somebody passes an unexpected object, and poof.  Let the user handle that case and panic if they want.  It's not like memory allocation where you can consistently expect a panic from Rust.

(I also think py_run! should not panic; it's pretty useful but hampered by that behavior.)

daniil-konovalenko:
Update:

I took a long break, but now I'm back!

I have recently figured out how to allow expressions in dict literal macros, thanks to @dtolnay's awesome [proc-macro workshop](https://github.com/dtolnay/proc-macro-workshop) and https://github.com/dtolnay/syn/issues/704.
We now allow using expressions as dict keys and values:
```rust
Python::with_gil(|py| {
    let my_dict = py_dict!(py => {"KeY".to_lowercase(): 7 + 10}).unwrap();
    py_run!(py, my_dict, "assert my_dict['key'] == 17");
})
```
I have also implemented the `py => {...}` syntax that @davidhewitt suggested and it looks really nice! 
I also discovered that, technically, we can ditch the requirement to pass `py` explicitly and rely on it being in scope. Although it makes for nicer syntax (`let dict = py_dict!{"key": "value"}`), I'm not sure if it's a good idea to be implicit here. Let me know your thoughts on this!

The code for `py_dict!` is pretty rough still, I am just excited to share that it works :) 
I'll take some time to clean it up, write tests, make errors nicer, etc., in the meantime, please feel free to review the current implementation or wait until it's more ready — I'll appreciate any comments or suggesstions!

davidhewitt:
Awesome! I've been quite busy recently so don't wait for my review, although I will definitely try to give one soon.

> `"KeY".to_lowercase(): 7 + 10}` 

Wow, I didn't even realise this was valid Python syntax! I confirmed it is indeed 😅 

```python
>>> {"KeY".lower(): 7 + 10}
{'key': 17}
```

> I also discovered that, technically, we can ditch the requirement to pass py explicitly and rely on it being in scope. Although it makes for nicer syntax (let dict = py_dict!{"key": "value"}), I'm not sure if it's a good idea to be implicit here. Let me know your thoughts on this!

Yeah, I was thinking on this. The awesome thing about macros is that actually we can support both. As long as it's clearly documented, this seems ok to me. Something like:

```
py_dict! { key: value }  // Assumes `py: Python` is already in scope
py_dict!( other_py => { key: value })  // To pass `other_py: Python` explicitly
```

I'd be interested to hear whether everyone else thinks that's an acceptable idea, or if we should aim to choose just one syntax.

I think if we decided to pick just one, I'd go for `py_dict!( py=> { key: value })` for clarity.

daniil-konovalenko:
> The awesome thing about macros is that actually we can support both
It didn't even come to me, thanks for pointing that out! Let's see what others think of it :)

mejrs:
I prefer the original syntax, e.g. `py_list!(py, [1, 2, 3])`. I don't get the meaning of this  `=>` syntax.

> we should aim to choose just one syntax.

I agree

> we can ditch the requirement to pass py explicitly and rely on it being in scope

What kind of errors do you get if you do this but `Python` is not in scope? Also this relies on it being named `py`, of which I'm not a fan.

daniil-konovalenko:
Hi everyone!

I'm sorry for the lack of progress and updates here. 2022 was an awful year for me, I had to move countries, and I didn't really have time time to finish this PR.

I don't think that I'll be able to finish this anytime soon, so if anyone wants to pick this up, I'll be happy to share the existing code and findings. If not, that's fine too, I think I'll get back to it eventually, but I can't really say when.

davidhewitt:
@daniil-konovalenko no problem, hope things settle down soon for you and family.

I have had this PR in the back of my mind, I think I'd like to create a `pyo3-typed-collections` library which has helpers like `PyList<T>` as an experiment. I'd pull this PR onto that library to see how it feels. Unfortunately I haven't gotten there yet 😄

mejrs:
Since this doesn't seem to be going anywhere I'm going ahead and close it. We also have more derive macros now; as a result of that this feature is less appealing to me now.

PR #1706, opened 2021-06-29
coverage: ignore tests

davidhewitt:
It doesn't seem that useful to measure coverage of test files, so let's perhaps not count them?


davidhewitt:
It looks like this is working (and benchmark failure is flaky).

The coverage number has gone down because tests ofc have very high coverage! A bit sad, but a reason for us to work to get it higher again 😄

messense (approved):

PR #1710, opened 2021-07-03
Improve Py<...> documentation

mejrs:
Adds some docs and examples to `Py<...>`.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -272,7 +272,7 @@ where
     ///     obj.as_ref(py)
     /// }
     /// ```
-    ///This can be solved by using [`Py::into_ref`] instead, which does not suffer from this issue.
+    /// This can be solved by using [`Py::into_ref`] instead, which does not suffer from this issue.
     /// Note that the lifetime of the [`Python<'py>`](crate::Python) token is transferred to
     /// the returned reference.
     ///
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -383,7 +383,7 @@ where
         self.as_ref(py).try_borrow()
     }
 
-    /// Attempts to immutably borrow the value `T`, returning an error if the value is currently borrowed.
+    /// Attempts to mutably borrow the value `T`, returning an error if the value is currently borrowed.
     ///
     /// The borrow lasts while the returned [`PyRefMut`] exists.
     ///
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -48,12 +48,13 @@ pub unsafe trait PyNativeType: Sized {
 /// # Example: Storing Python objects in structs
 ///
 /// As all the native Python objects only appear as references, storing them in structs doesn't work well.
-/// For example, there is no way to make something like this compile!
+/// For example, this won't compile:
 ///
 /// ```compile_fail
 /// # use pyo3::prelude::*;
 /// # use pyo3::types::PyDict;
 /// #
+/// #[pyclass]
 /// struct Foo<'py> {
 ///     inner: &'py PyDict,
 /// }
@@ -84,10 +85,12 @@ pub unsafe trait PyNativeType: Sized {
 /// # use pyo3::prelude::*;
 /// # use pyo3::types::PyDict;
 /// #
+/// #[pyclass]
 /// struct Foo {
 ///     inner: Py<PyDict>,
 /// }
 ///
+/// #[pyclass]
 /// impl Foo {
 ///     fn new() -> Foo {
 ///         Python::with_gil(|py| {
@@ -105,6 +108,7 @@ pub unsafe trait PyNativeType: Sized {
 /// #[pyclass]
 /// struct Bar {/* fields omitted */}
 ///
+/// #[pyclass]
 /// struct Foo {
 ///     inner: Py<Bar>,
 /// }
@@ -158,13 +162,13 @@ pub unsafe trait PyNativeType: Sized {
 ///
 /// It is easy to accidentally create reference cycles using [`Py`]`<T>`.
 /// The Python interpreter can break these reference cycles within pyclasses if they
-/// implement the [`PyGCProtocol`](crate::class::gc::PyGCProtocol). If you permit your pyclass
-/// to create reference cycles you should implement this protocol to avoid leaking memory.
+/// implement the [`PyGCProtocol`](crate::class::gc::PyGCProtocol). If your pyclass
+/// contains other Python objects you should implement this protocol to avoid leaking memory.
 ///
 /// # A note on `Send` and `Sync`
 ///
 /// Accessing this object is threadsafe, since any access to its API requires a [`Python<'py>`](crate::Python) token.
-/// As you can only get this by acquiring the GIL, `Py<...>` is safe against concurrent access.
+/// As you can only get this by acquiring the GIL, `Py<...>` "implements [`Send`] and [`Sync`].
 ///
 /// [`Rc`]: std::rc::Rc
 /// [`RefCell`]: std::cell::RefCell
@@ -208,9 +212,11 @@ impl<T> Py<T>
 where
     T: PyTypeInfo,
 {
-    /// Borrows a GIL-bound reference to the contained `T`. By binding to the GIL lifetime, this
-    /// allows the GIL-bound reference to not require [`Python<'py>`](crate::Python) for any of its methods,
-    /// which makes calling methods on it more ergonomic.
+    /// Borrows a GIL-bound reference to the contained `T`.
+    ///
+    /// By binding to the GIL lifetime, this allows the GIL-bound reference to not require
+    /// [`Python<'py>`](crate::Python) for any of its methods, which makes calling methods
+    /// on it more ergonomic.
     ///
     /// For native types, this reference is `&T`. For pyclasses, this is `&PyCell<T>`.
     ///
@@ -252,13 +258,18 @@ where
         unsafe { PyNativeType::unchecked_downcast(&*any) }
     }
 
-    /// Similar to [`as_ref`](#method.as_ref) but consumes `self` and registers the
+    /// Borrows a GIL-bound reference to the contained `T` independently of the lifetime of `T`.
+    ///
+    /// This method is similar to [`as_ref`](#method.as_ref) but consumes `self` and registers the
     /// Python object reference in PyO3's object storage. The reference count for the Python
     /// object will not be decreased until the GIL lifetime ends.
     ///
+    /// You should prefer using [`as_ref`](#method.as_ref) if you can as it'll have less overhead.
+    ///
     /// # Examples
     ///
-    /// [`Py::as_ref`]'s lifetime limitation forbids creating a function that references a .
+    /// [`Py::as_ref`]'s lifetime limitation forbids creating a function that references a
+    /// variable created inside the function.
     ///
     /// ```rust,compile_fail
     /// # use pyo3::prelude::*;
@@ -272,6 +283,7 @@ where
     ///     obj.as_ref(py)
     /// }
     /// ```
+    ///
     /// This can be solved by using [`Py::into_ref`] instead, which does not suffer from this issue.
     /// Note that the lifetime of the [`Python<'py>`](crate::Python) token is transferred to
     /// the returned reference.
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -32,21 +32,142 @@ pub unsafe trait PyNativeType: Sized {
     }
 }
 
-/// A Python object of known type T.
+/// A GIL-independent reference to an object allocated on the Python heap.
 ///
-/// Accessing this object is thread-safe, since any access to its API requires a `Python<'py>` GIL
-/// token. There are a few different ways to use the Python object contained:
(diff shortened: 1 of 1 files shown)


src/instance.rs
```diff
@@ -32,21 +32,142 @@ pub unsafe trait PyNativeType: Sized {
...
+/// });
+/// # }
+/// ```
+///
+/// # Preventing reference cycles
+///
+/// It is easy to accidentally create reference cycles using [`Py`]`<T>`.
+/// The Python interpreter can break these reference cycles within pyclasses if they
+/// implement the [`PyGCProtocol`](crate::class::gc::PyGCProtocol). If you permit your pyclass
+/// to create reference cycles you should implement this protocol to avoid leaking memory.
```
messense, line 162: I think usually users should always implement the `PyGCProtocol` and opt-out only when they are sure it's not needed. This would align with [what Cython does](https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#disabling-cycle-breaking-tp-clear).

src/instance.rs
```diff
@@ -32,21 +32,142 @@ pub unsafe trait PyNativeType: Sized {
...
+///  - You can call methods directly on `Py` with [`Py::call`], [`Py::call_method`] and friends.
+/// These require passing in the [`Python<'py>`](crate::Python) token but are otherwise similar to the corresponding
+/// methods on [`PyAny`].
 ///
-/// See [the guide](https://pyo3.rs/main/types.html) for an explanation
-/// of the different Python object types.
+/// # Example: Storing Python objects in structs
+///
+/// As all the native Python objects only appear as references, storing them in structs doesn't work well.
+/// For example, there is no way to make something like this compile!
```
davidhewitt, line 51: I feel like it'd be helpful to the reader to explain what "something like this" is. I tried to word it as "no way to store `&PyDict` directly in a `#[pyclass]`.

src/instance.rs
```diff
@@ -32,21 +32,142 @@ pub unsafe trait PyNativeType: Sized {
...
+///             // so won't be able to outlive this closure.           
+///             let dict: &PyDict = PyDict::new(py);
+///
+///             // because `Foo` contains `dict` its lifetime
+///             // is now also tied to `py`.
+///             Foo { inner: dict }
+///         });
+///         // Foo is no longer valid.
+///         // Returning it from this function is a 💥 compiler error 💥
+///         foo
```
davidhewitt, line 76: Might want to use a `#[pyclass]`, because otherwise this could be made to compile by adding `py: Python` argument to `Foo::new`.

(I've actually used this exact pattern to bind temporarily to Python-owned data just for a short moment, it's also essentially what `#[derive(FromPyObject)]` does if you put a `&PyDict` reference in it.)

davidhewitt (approved):
Thank you, as usual this is brilliant! Just a couple of tiny suggestions.

src/instance.rs
birkenfeld: I think it would be enough to show
```
/// #[pyclass]
/// struct Foo<'py> {
///     inner: &'py PyDict,
/// }
```
without the impl block.

Structs used purely in Rust can of course have a `'py` lifetime attached and contain `&'py` members, they just can't "escape" to Python.

src/instance.rs
```diff
@@ -32,21 +32,142 @@ pub unsafe trait PyNativeType: Sized {
...
+/// }
+/// ```
+///
+/// [`Py`]`<T>` can be used to get around this by converting `dict` into a GIL-independent reference:
+///
+/// ```rust
+/// # use pyo3::prelude::*;
+/// # use pyo3::types::PyDict;
+/// #
+/// struct Foo {
```
birkenfeld, line 87: derive pyclass?  I feel like that's the most important use case here.

src/instance.rs
birkenfeld: This advice should be placed prominently where `pyclass` is documented, then.

src/instance.rs
```diff
@@ -32,21 +32,142 @@ pub unsafe trait PyNativeType: Sized {
...
+///
+/// It is easy to accidentally create reference cycles using [`Py`]`<T>`.
+/// The Python interpreter can break these reference cycles within pyclasses if they
+/// implement the [`PyGCProtocol`](crate::class::gc::PyGCProtocol). If you permit your pyclass
+/// to create reference cycles you should implement this protocol to avoid leaking memory.
+///
+/// # A note on `Send` and `Sync`
+///
+/// Accessing this object is threadsafe, since any access to its API requires a [`Python<'py>`](crate::Python) token.
+/// As you can only get this by acquiring the GIL, `Py<...>` is safe against concurrent access.
```
birkenfeld, line 167: replace "safe against concurrent access" which is imprecise with "implements Send and Sync".

src/instance.rs
```diff
@@ -71,55 +209,82 @@ where
...
     /// ```
     /// # use pyo3::prelude::*;
     /// # use pyo3::types::PyList;
-    /// # Python::with_gil(|py| {
-    /// let list: Py<PyList> = PyList::empty(py).into();
-    /// let list: &PyList = list.as_ref(py);
-    /// assert_eq!(list.len(), 0);
-    /// # });
+    /// #
+    /// Python::with_gil(|py| {
```
birkenfeld, line 229: why make the `with_gil` visible?  Where the `py` token comes from is accidental, it could just as well be within a `#[pymethods]` method that gets it as an argument.

src/instance.rs
```diff
@@ -71,55 +209,82 @@ where
...
-    /// Similar to [`as_ref`](#method.as_ref), and also consumes this `Py` and registers the
+    /// Similar to [`as_ref`](#method.as_ref) but consumes `self` and registers the
     /// Python object reference in PyO3's object storage. The reference count for the Python
     /// object will not be decreased until the GIL lifetime ends.
     ///
     /// # Examples
     ///
-    /// Useful when returning GIL-bound references from functions. In the snippet below, note that
-    /// the `'py` lifetime of the input GIL lifetime is also given to the returned reference:
+    /// [`Py::as_ref`]'s lifetime limitation forbids creating a function that references a .
```
birkenfeld, line 261: Something missing here?

src/instance.rs
```diff
@@ -71,55 +209,82 @@ where
...
+    /// fn new_py_any<'py>(py: Python<'py>, value: impl IntoPy<Py<PyAny>>) -> &'py PyAny {
+    ///     let obj: Py<PyAny> = value.into_py(py);
+    ///
+    ///     // The lifetime of the return value of this function is the shortest
+    ///     // of `obj` and `py`. As `obj` is owned by the current function,
+    ///     // Rust won't let the return value escape this function!
+    ///     obj.as_ref(py)
+    /// }
     /// ```
+    ///This can be solved by using [`Py::into_ref`] instead, which does not suffer from this issue.
```
birkenfeld, line 275:
```suggestion
    /// This can be solved by using [`Py::into_ref`] instead, which does not suffer from this issue.
```

src/instance.rs
```diff
@@ -71,55 +209,82 @@ where
...
+    ///     assert!(my_class_cell.try_borrow().is_ok());
+    /// });
     /// ```
     pub fn as_ref<'py>(&'py self, _py: Python<'py>) -> &'py T::AsRefTarget {
         let any = self.as_ptr() as *const PyAny;
         unsafe { PyNativeType::unchecked_downcast(&*any) }
     }
 
-    /// Similar to [`as_ref`](#method.as_ref), and also consumes this `Py` and registers the
+    /// Similar to [`as_ref`](#method.as_ref) but consumes `self` and registers the
```
birkenfeld, line 255: Should mention that this has a cost; otherwise it sounds like `into_ref` is always preferable.

src/instance.rs
```diff
@@ -132,22 +297,72 @@ impl<T> Py<T>
...
 {
-    /// Immutably borrows the value `T`. This borrow lasts untill the returned `PyRef` exists.
+    /// Immutably borrows the value `T`.
+    ///
+    /// This borrow lasts while the returned [`PyRef`] exists.
+    /// Multiple immutable borrows can be taken out at the same time.
     ///
     /// Equivalent to `self.as_ref(py).borrow()` -
-    /// see [`PyCell::borrow`](../pycell/struct.PyCell.html#method.borrow)
+    /// see [`PyCell::borrow`](crate::pycell::PyCell::borrow)
```
birkenfeld, line 306:
```suggestion
    /// see [`PyCell::borrow`](crate::pycell::PyCell::borrow).
```

src/instance.rs
```diff
@@ -132,22 +297,72 @@ impl<T> Py<T>
...
     }
 
-    /// Mutably borrows the value `T`. This borrow lasts untill the returned `PyRefMut` exists.
+    /// Mutably borrows the value `T`.
+    ///
+    /// This borrow lasts while the returned [`PyRefMut`] exists.
     ///
     /// Equivalent to `self.as_ref(py).borrow_mut()` -
-    /// see [`PyCell::borrow_mut`](../pycell/struct.PyCell.html#method.borrow_mut)
+    /// see [`PyCell::borrow_mut`](crate::pycell::PyCell::borrow_mut)
```
birkenfeld, line 343:
```suggestion
    /// see [`PyCell::borrow_mut`](crate::pycell::PyCell::borrow_mut).
```

src/instance.rs
```diff
@@ -156,24 +371,26 @@ where
...
-    /// see [`PyCell::try_borrow`](../pycell/struct.PyCell.html#method.try_borrow)
+    /// Equivalent to `self.as_ref(py).borrow_mut()` -
+    /// see [`PyCell::try_borrow`](crate::pycell::PyCell::try_borrow).
     pub fn try_borrow<'py>(&'py self, py: Python<'py>) -> Result<PyRef<'py, T>, PyBorrowError> {
         self.as_ref(py).try_borrow()
     }
 
-    /// Mutably borrows the value `T`, returning an error if the value is currently borrowed.
-    /// This borrow lasts untill the returned `PyRefMut` exists.
+    /// Attempts to immutably borrow the value `T`, returning an error if the value is currently borrowed.
```
birkenfeld, line 386:
```suggestion
    /// Attempts to mutably borrow the value `T`, returning an error if the value is currently borrowed.
```

birkenfeld (changes requested):

src/instance.rs
mejrs: Oops, thanks.

src/instance.rs
mejrs: > I think usually users should always implement the PyGCProtocol and opt-out only when they are sure it's not needed. 

Good idea, I'll change that.

> This advice should be placed prominently where pyclass is documented, then.

That is what the documentation for the `pyclass` macro recommends: https://pyo3.rs/main/doc/pyo3/proc_macro/attr.pyclass.html, as does the guide.

src/instance.rs
mejrs: I think doing away with that and just "For example, this won't compile" would be better.

src/instance.rs
mejrs: > Might want to use a #[pyclass]

Done.

> without the impl block.

I think the impl block is important to show that the lifetime will always be bound to the Python token used to create it. I would rather *show* this than explain it.

Also, while it is "lifetimes 101" and should be obvious to most Rust programmers, I think it's important to explain this in detail because I think a lot of people that use pyo3 are not experienced Rust programmers.

src/instance.rs
mejrs: I think it's weird to see an example conjuring variables out of nowhere. It also makes the example harder to copy-paste.

src/instance.rs
birkenfeld: Ok, great.

src/instance.rs
birkenfeld: I never understood this argument - next you'll have to supply a function around it, and finally a whole project including `Cargo.toml`.  It is very well understood what `py` is in the context of PyO3.

People are not supposed to blindly copy-paste code examples, they're supposed to understand what they show and apply it to their situation.  (end of rant)

src/instance.rs
birkenfeld: Ok, fair enough.

src/instance.rs
mejrs: Is the additional cost just in situations where a `Py<T>` is needlessly created and destroyed? Or are there performance penalties beyond that?

davidhewitt: It requires placing the pointer into pyO3's internal pool, so will involve a thread-local access. Tbh I'm not sure how much that's relevant right now as most of PyO3's APIs do that!

messense (approved):

davidhewitt:
Are there any outstanding changes or questions from @birkenfeld's review?

mejrs:
> 
> 
> Are there any outstanding changes or questions from @birkenfeld's review?

That should be everything but making `with_gil` visible - I'm not sure what to do about that one.

davidhewitt:
👍 in general I'm also in favour of having `Python::with_gil` present around examples, at least for now. A bunch of examples already use it, and it's nice to keep consistent.

Maybe we'll change our minds on this in the future and end up hiding a pile of `with_gil` usages, but for now let's proceed to merge this. Thanks again!

PR #1711, opened 2021-07-03
Add full path for PyCell

mejrs:
This makes it so that users no longer have to bring  `PyCell` into scope to make the `#[pyclass]` macro work (either with the prelude or explicit import).


messense:
You need to update UI tests output with `TRYBUILD=overwrite cargo test`

davidhewitt:
Thanks! I'll do this myself and force-push now, because this looks like a very good catch that would be a potentially annoying bug in 0.14.

davidhewitt (approved):

PR #1712, opened 2021-07-04
Implement `IntoPy<PyObject>` for `&PathBuf` and `&OsString`

messense:
Without this impl users need to call `as_path()` to convert a `&PathBuf` to a `&Path`, for example https://github.com/messense/rjsonnet-py/commit/b29e161a37e0f9642dca8b5e3b4836c816615647

Otherwise the error message is not easy to understand:

```rust
error[E0277]: the trait bound `PathBuf: AsPyPointer` is not satisfied
  --> src/lib.rs:34:55
   |
34 |             Python::with_gil(|py| match self.callback.call(py, (from, path), None) {
   |                                                       ^^^^ the trait `AsPyPointer` is not implemented for `PathBuf`
   |
   = note: required because of the requirements on the impl of `pyo3::IntoPy<Py<PyAny>>` for `&PathBuf`
   = note: 1 redundant requirements hidden
   = note: required because of the requirements on the impl of `pyo3::IntoPy<Py<PyTuple>>` for `(&PathBuf, &PathBuf)`
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ PyO3 versions, please see the [migration guide](https://pyo3.rs/main/migration.h
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Implement `IntoPy<PyObject>` for `&PathBuf` and `&OsString`. [#1721](https://github.com/PyO3/pyo3/pull/1712)
+
 ## [0.14.0] - 2021-07-03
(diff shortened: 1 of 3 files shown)


davidhewitt:
👍 Can you add it for `&OsString` too please? That would then be consistent with `&String`, which already exists. Otherwise LGTM!

Btw I am preparing a 0.14.1 release today to fix a PyPy crash found in https://github.com/pyca/cryptography/pull/6154, happy to see this merged and included in the release too.

messense:
> 👍 Can you add it for `&OsString` too please? That would then be consistent with `&String`, which already exists. 

Added.

davidhewitt (approved):
That's great, thanks!

PR #1713, opened 2021-07-04
PyPy: the PyList/Tuple_GET/SET macros are defined as functions

birkenfeld:


davidhewitt (approved):
👍 thanks! We should definitely add a CHANGELOG note, maybe something like 

```
- Fix crashes on PyPy due to incorrect definitions of `PyList_SET_ITEM`
```

davidhewitt:
Awesome, thanks!

PR #1715, opened 2021-07-04
Add `PyErr::take` method

messense:
Just like [`Option::take`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.take).

If no error is set, the existing `PyError::fetch`method  returns a `SystemError` makes it inconvenient to use in Rust side for example when setting the cause for `PyErr`: https://github.com/messense/rjsonnet-py/blob/e144e9c08621135e03d06c8d0fc53af51ff51b3c/src/lib.rs#L224-L233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Implement `IntoPy<PyObject>` for `&PathBuf` and `&OsString`. [#1721](https://github.com/PyO3/pyo3/pull/1712)
+- Add `PyErr::take`. [#1715](https://github.com/PyO3/pyo3/pull/1715)
 
 ## [0.14.0] - 2021-07-03
 
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -243,6 +243,21 @@ impl PyErr {
         }
     }
 
+    /// Retrieves the current error from the Python interpreter's global state.
+    ///
+    /// The error is cleared from the Python interpreter.
+    /// If no error is set, returns a `None`.
+    ///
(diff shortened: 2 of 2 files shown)


davidhewitt:
Hmmm. I agree this is useful, but I feel that `PyErr::fetch() -> Option<PyErr>` is probably the more accurate representation of the underlying `PyErr_Fetch` API.

At the moment the `SystemError` comes from the "normalizing" process internally. We could remove that hidden "feature" and make some kind of function like this to use inside the PyO3 codebase:

```rust
pub(crate) fn api_call_failed(py: Python) -> PyErr {
    PyErr::fetch(py).unwrap_or_else(|| PySystemError::new_err("exception value missing"))
}
```

Though TBH maybe it's a serious enough issue if the error is missing that we should panic.

messense:
> Hmmm. I agree this is useful, but I feel that `PyErr::fetch() -> Option<PyErr>` is probably the more accurate representation of the underlying `PyErr_Fetch` API.

Yes, It feels weird that `PyErr::fetch()` doesn't return a `Option<PyErr>`, although changing it is a breaking change.

davidhewitt:
Agreed it would be breaking. I'm not sure how many downstream users would be affected that, I suspect most don't use `PyErr::fetch` directly?

Perhaps let's put this on the list to decide for 0.15 - either we can add this new `take()`, or we can consider breaking `fetch()`. If doing so can lead to simpler code think that's worth experimenting with.

(For now it doesn't look too hard to implement `pyerr_take()` as local helper function, even if it's mildly annoying to repeat in multiple crates...)

messense:
I don't mind waiting for 0.15, will experiment with `PyErr:fetch() -> Option<PyErr>`.

davidhewitt:
@messense should we close this now that it looks like #1717 will be merged for 0.15?

messense:
Sure.

PR #1716, opened 2021-07-04
release: 0.14.1

davidhewitt:
This is a patch release to resolve the issues with PyPy. I think they're nasty enough to warrant a swift release.

I'll put it live a few hours from now if we don't hear any other feedback on 0.14 in the meanwhile which also needs attention.


davidhewitt:
Release is live! Thanks to all involved today.

PR #1717, opened 2021-07-05
Change `PyErr::fetch` to return an `Option<PyErr>`

messense:
Closes #1715

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -215,16 +215,21 @@ impl PyErr {
     /// Retrieves the current error from the Python interpreter's global state.
     ///
     /// The error is cleared from the Python interpreter.
-    /// If no error is set, returns a `SystemError`.
+    /// If no error is set, returns a `None`.
     ///
     /// If the error fetched is a `PanicException` (which would have originated from a panic in a
     /// pyo3 callback) then this function will resume the panic.
-    pub fn fetch(py: Python) -> PyErr {
+    pub fn fetch(py: Python) -> Option<PyErr> {
         unsafe {
             let mut ptype: *mut ffi::PyObject = std::ptr::null_mut();
             let mut pvalue: *mut ffi::PyObject = std::ptr::null_mut();
             let mut ptraceback: *mut ffi::PyObject = std::ptr::null_mut();
             ffi::PyErr_Fetch(&mut ptype, &mut pvalue, &mut ptraceback);
+
+            // If the error indicator is not set, all three variables are set to NULL
+            if ptype.is_null() && pvalue.is_null() && ptraceback.is_null() {
+                return None;
+            }
 
             let err = PyErr::new_from_ffi_tuple(py, ptype, pvalue, ptraceback);
 
@@ -242,8 +247,16 @@ impl PyErr {
                 std::panic::resume_unwind(Box::new(msg))
             }
 
-            err
+            Some(err)
         }
+    }
+
+    /// Retrieves the current error from the Python interpreter's global state.
+    ///
+    /// The error is cleared from the Python interpreter.
+    /// Panics if no error is set
+    pub(crate) fn api_call_failed(py: Python) -> PyErr {
+        PyErr::fetch(py).expect("exception missing")
     }
 
     /// Creates a new exception type with the given name, which must be of the form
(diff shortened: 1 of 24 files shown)


src/err/mod.rs
```diff
@@ -239,10 +244,18 @@ impl PyErr {
...
 
-            err
+            Some(err)
         }
     }
 
+    /// Retrieves the current error from the Python interpreter's global state.
+    ///
+    /// The error is cleared from the Python interpreter.
+    /// Panics if no error is set
```
birkenfeld, line 254: This behavior is new compared to before, where it returned a `SystemError`.

Should we deviate from the Python convention here?

src/types/dict.rs
```diff
@@ -41,11 +41,11 @@ impl PyDict {
...
         unsafe {
             let dict = py.from_owned_ptr::<PyDict>(ffi::PyDict_New());
-            match ffi::PyDict_MergeFromSeq2(dict.into_ptr(), seq.into_ptr(), 1i32) {
-                0 => Ok(dict),
-                -1 => Err(PyErr::fetch(py)),
-                _ => unreachable!(),
-            }
+            err::error_on_minusone(
+                py,
+                ffi::PyDict_MergeFromSeq2(dict.into_ptr(), seq.into_ptr(), 1i32),
```
birkenfeld, line 46: nit: does this have to be `i32`? Seems better to actually leave the type unspecified since it's a `c_int`.

src/types/typeobject.rs
```diff
@@ -50,9 +50,8 @@ impl PyType {
     {
         let result =
             unsafe { ffi::PyObject_IsSubclass(self.as_ptr(), T::type_object(self.py()).as_ptr()) };
-        if result == -1 {
-            Err(PyErr::fetch(self.py()))
-        } else if result == 1 {
+        err::error_on_minusone(self.py(), result)?;
+        if result == 1 {
```
birkenfeld, line 54: `Ok(result == 1)`?

src/types/typeobject.rs
```diff
@@ -64,9 +63,8 @@ impl PyType {
     /// Equivalent to Python's `isinstance` function.
     pub fn is_instance<T: AsPyPointer>(&self, obj: &T) -> PyResult<bool> {
         let result = unsafe { ffi::PyObject_IsInstance(obj.as_ptr(), self.as_ptr()) };
-        if result == -1 {
-            Err(PyErr::fetch(self.py()))
-        } else if result == 1 {
+        err::error_on_minusone(self.py(), result)?;
+        if result == 1 {
```
birkenfeld, line 67: same

src/err/mod.rs
messense: I'm not sure, see https://github.com/PyO3/pyo3/pull/1715#issuecomment-873612388

birkenfeld: I mean, I fully agree with `PyErr::fetch` returning an option. But the thing that we used for our own calls, now called `api_call_failed` used to return a SystemError and now panics.

messense: We're not sure if it really needs to return a `SystemError`, so I'm trying to just panic now to see if anything breaks.

davidhewitt: @birkenfeld what do you mean by "the Python convention" above?

I'm personally going back and forth on whether this should panic. On one hand, it feels like if this ever happens it's a genuine bug - we think a Python error has occurred but we found no exception when we went to retrieve it.

On the other hand, we translate panics into exceptions which are pretty hard to catch for Python users, so it might be nicer to continue to use `SystemError` anyway.

birkenfeld: When a CPython function returns NULL/-1/... but no exception is set (i.e. someone wrote a buggy C function), the eval loop raises a `SystemError: error return with no exception set`.  This is equivalent to the case where `api_call_failed` is used.

davidhewitt: Aha yes, found it: https://github.com/python/cpython/blob/17f94e28882e1e2b331ace93f42e8615383dee59/Python/ceval.c#L4330-L4333

In which case I would definitely vote for us returning a `SystemError` with the same error message as CPython!

messense: Looks like CPython only raises a `SystemError` in release build, do we want to follow that too and only raises a `SystemError` in release mode and panic in debug mode?

davidhewitt: I don't mind. Having the panic message printed would probably help debugging, because it'll include the Rust traceback. The Python `SystemError` will only have a traceback which starts at the point where Python called Rust.

@birkenfeld what do you think on that option?

birkenfeld: Yeah it's fine.  This "should not happen" anyway, and if it does it's as likely to be a bug in the Rust code due to misunderstanding the error condition.

davidhewitt (approved):
This looks good, but let's wait to merge it until we're closer to the 0.15 release because I think we might have a patch release or two in the 0.14 line yet.

I'll happily rebase and add a CHANGELOG at that point.

Thanks!

davidhewitt:
I think it's ok to start merging stuff for 0.15, so I'm going to rebase and merge this now.

Issue #1683, opened 2021-06-18
Avoiding panics on recursive GIL acquisitions


LegionMammal978:
For a personal project, I'm trying to provide a Rust callback function to a Python module as a raw function pointer. However, if I attempt to acquire the GIL within the callback function, it causes the program to panic due to out-of-order GIL dropping. As a minimal example:

    use pyo3::prelude::*;

    extern "C" fn callback() {
        println!("begin callback");
        let _ = Python::acquire_gil();
        println!("end callback");
    }

    fn main() -> PyResult<()> {
        Python::with_gil(|py| {
            PyModule::import(py, "ctypes")?
                .call_method1("CFUNCTYPE", ((),))?
                .call1((callback as *const () as usize,))?
                .call0()?;
            Ok(())
        })
    }

The program panics before `end callback` can be printed:

    $ target/debug/gil_test 
    begin callback
    thread 'main' panicked at 'The first GILGuard acquired must be the last one dropped.', /home/lm978/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.13.2/src/gil.rs:290:17
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
    thread 'main' panicked at 'The first GILGuard acquired must be the last one dropped.', /home/lm978/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.13.2/src/gil.rs:290:17
    stack backtrace:
       0:     0x5599ade9a3b0 - std::backtrace_rs::backtrace::libunwind::trace::h04d12fdcddff82aa
                                   at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/../../backtrace/src/backtrace/libunwind.rs:100:5
       1:     0x5599ade9a3b0 - std::backtrace_rs::backtrace::trace_unsynchronized::h1459b974b6fbe5e1
                                   at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
       2:     0x5599ade9a3b0 - std::sys_common::backtrace::_print_fmt::h9b8396a669123d95
                                   at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/sys_common/backtrace.rs:67:5
[...]

PR #1724, opened 2021-07-13
err: don't use with_gil internally in PyErr::new()

davidhewitt:
This is a partial follow-on from #1683 

If we're going to remove the optimization of using PyO3's thread-local to check for the GIL being held in `with_gil`, then it seems desirable that `PyErr::new` really doesn't need the GIL at all.

I added a benchmark to accompany the change; it only slightly impacts performance (<2% slower and this is on the error path anyway). It also has the nice side effect that users already creating `PyErr` outside of a GIL will get a significant speedup, because the GIL won't be temporarily acquired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Add `indexmap` feature to add `ToPyObject`, `IntoPy` and `FromPyObject` implementations for `indexmap::IndexMap`. [#1728](https://github.com/PyO3/pyo3/pull/1728)
 
+### Changed
+
+- `PyErr::new` no longer acquires the Python GIL internally. [#1724](https://github.com/PyO3/pyo3/pull/1724)
+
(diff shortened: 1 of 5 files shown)


messense:
> Warning: Performance alert! Previous value was 5255746.530213239 and current value is 2283663.643596061. It is 2.301453869947796x worse than previous exceeding a ratio threshold 2
> Warning: Performance alert! Previous value was 6427231.592283239 and current value is 2608965.9845569944. It is 2.4635168224987765x worse than previous exceeding a ratio threshold 2

This action is dumb. 😅

davidhewitt:
Heh indeed, it could use some work on output formatting! I would write a PR myself but it looks like the maintainer isn't looking after it, and I don't want to devote time to maintaining a fork... https://github.com/rhysd/github-action-benchmark/pulls

davidhewitt:
Anyone got any comments or concerns here, or should I look to merge this?

davidhewitt:
Proceeding to merge this!

messense (approved):

PR #1725, opened 2021-07-14
pyo3-build-config: improve error messaging

davidhewitt:
This PR tweaks the implementation of `Result` in the build scripts to use a simple `Error` type which supports chaining the "source".

I was upgrading a private codebase to 0.14 and hit an unhelpful error message `error: No such file or directory (os error 2)`.

I suspect it was probably caused by this line: https://github.com/PyO3/pyo3/blob/584fc6fe42a8ec6fe7fa55ff98aec8bf8471e134/build.rs#L197

... which is a different problem (I'm worried that it might be caused by a bad interaction with `sccache`).

At least after this PR, this error (and a few others) have a more helpful context message added to the original error.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,22 +1,13 @@
 use std::{env, process::Command};
 
-use pyo3_build_config::{InterpreterConfig, PythonImplementation, PythonVersion};
-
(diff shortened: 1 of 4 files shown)


messense (approved):
Nice!

PR #1726, opened 2021-07-15
workaround an issue that causes function arguments to show up as if they are uncovered

alex:
https://github.com/rust-lang/rust/issues/86972 describes this issue in some detail and links to a minimal reproducer

making this change, and this change alone, resolves the coverage issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Fix regression in 0.14.0 rejecting usage of `#[doc(hidden)]` on structs and functions annotated with PyO3 macros. [#1722](https://github.com/PyO3/pyo3/pull/1722)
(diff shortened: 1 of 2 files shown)


davidhewitt (approved):
Hmmm, interesting.

So if I understand correctly, we've generated a branch which spans the argument, however in your tests it's not covered. You (or any other user doing similar things) would have to invoke every argument with incorrect types to achieve full coverage.

I think it's reasonable to adjust this in this way. Can we add a CHANGELOG message please, in case others are also noticing this behaviour?

I am curious though: if branches like this are a problem, I wonder why you're not having issues with the branching around default arguments slightly further down the file?

https://github.com/PyO3/pyo3/blob/8107c86682093e4e52ecf07a39f369ba61030b41/pyo3-macros-backend/src/params.rs#L244-L260

alex:
I have really no explanation for why it's not impacting all of these spans or anything like that.

I'll add the changelog entry.

davidhewitt:
👍 I guess we can always tweak further if other users find other cases. Thanks very much for fixing.

alex:
Thanks for the quick merge!

PR #1728, opened 2021-07-16
Add optional support for conversion from `indexmap::IndexMap`

IvanIsCoding:
Add support for converting [`indexmap::IndexMap`](https://docs.rs/indexmap/1.7.0/indexmap/map/struct.IndexMap.html) into `&PyDict` under an optional feature

IndexMap is a hash table that is closely compatible with the standard HashMap, with the difference that it preserves the insertion order when iterating over keys. [It was inspired by Python's 3.6+ dict implementation.](https://github.com/bluss/indexmap#background)

The use case for this new dict conversion is for having a more deterministic iteration behaviour when passing hash tables between Rust and Python. Because of the guarantees Python makes about dictionaries, some users are puzzled that looping through a dict or calling `.items()` is not identical on multiple runs of a function that originally returned a HashMap


IvanIsCoding:
Also, this PR is almost identical to @mtreinish's #1114 but with `indexmap::IndexMap` instead of `hashbrown::HashMap`. I cut `indexmap::IndexSet` from this though because converting into `&PySet` does not preserve the insertion order, it didn't make as much sense to include it.

src/types/dict.rs
```diff
@@ -455,6 +455,99 @@ mod hashbrown_hashmap_conversion {
     }
 }
 
+#[cfg(feature = "indexmap")]
+#[cfg_attr(docsrs, doc(cfg(feature = "indexmap")))]
+mod indexmap_indexmap_conversion {
```
davidhewitt, line 460: Could this code actually please go in `src/conversions/indexmap.rs`? I realise you added alongside the `hashbrown` code but I think it's easier to find trait implementations / manage all optional features if we isolate them to new files.

src/types/dict.rs
```diff
@@ -455,6 +455,99 @@ mod hashbrown_hashmap_conversion {
...
+            let mut ret = indexmap::IndexMap::with_capacity_and_hasher(dict.len(), S::default());
+            for (k, v) in dict.iter() {
+                ret.insert(K::extract(k)?, V::extract(v)?);
+            }
+            Ok(ret)
+        }
+    }
+
+    #[test]
+    fn test_indexmap_indexmap_to_python() {
```
davidhewitt, line 506: Tests inside a `mod tests` please. Also, given the motivation is to use `indexmap` to preserve order, it would be great to write a test which round-trips from Rust -> Python -> Rust and then confirms the original order is preserved.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Add `indexmap` feature to convert `indexmap::IndexMap` into `&PyDict`, an alternative that preserves the insertion order of the elements.
```
davidhewitt, line 13: We usually add a PR number to the entry. Also, I think the extra detail is probably more than necessary for a changelog.

```suggestion
- Add `indexmap` feature to add `ToPyObject`, `IntoPy` and `FromPyObject` implementations for `indexmap::IndexMap`. [#1728](https://github.com/PyO3/pyo3/pull/1728)
```

davidhewitt (commented):
Thanks for this, I agree it's a useful feature and worth supporting. It could also be worth adding a note in the guide somewhere in the `conversions` section, as I suspect users might want this from time to time.

Few nits to follow...

mejrs:
Thanks!

Would you be willing to add something like [this](https://docs.rs/pyo3/0.14.1/pyo3/num_bigint/index.html) with a) an example of the cargo.toml that people can copy paste b) a doc example illustrating why people would want to use it (probably a quick example of the conversion maintaining insertion order)?

Also, python dictionaries maintaining insertion order is only guaranteed on 3.7 and up - from https://docs.python.org/3/library/stdtypes.html :

> Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.

It doesn't appear to be a problem to me, but is this something we should be concerned about?

src/types/dict.rs
IvanIsCoding: I added one, see if you agree with the test

IvanIsCoding:
> Thanks!
> 
> Would you be willing to add something like [this](https://docs.rs/pyo3/0.14.1/pyo3/num_bigint/index.html) with a) an example of the cargo.toml that people can copy paste b) a doc example illustrating why people would want to use it (probably a quick example of the conversion maintaining insertion order)?
> 
> Also, python dictionaries maintaining insertion order is only guaranteed on 3.7 and up - from https://docs.python.org/3/library/stdtypes.html :
> 
> > Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.
> 
> It doesn't look like a problem to me, but is this something we should be concerned about?

I will try to add an example like you listed, I will come up with some use cases.

For the 3.6 vs 3.7 detail. CPython has insertion order since 3.6, and [PyPy had it even before](https://twitter.com/pypyproject/status/942070907715182592). That covers all Python 3.6 implementations PyO3 supports no? 

So I think we can add in the `indexmap` section that:
> Since Python 3.7, Dictionary order is guaranteed to be insertion order and the conversions with IndexMap will reflect that guarantee. In practice, the guarantee will also be mainted in Python 3.6 because of the implementation details of CPython and PyPy.

src/conversions/indexmap.rs
```diff
@@ -0,0 +1,132 @@
+#[cfg(feature = "indexmap")]
+#[cfg_attr(docsrs, doc(cfg(feature = "indexmap")))]
+mod indexmap_indexmap_conversion {
```
davidhewitt, line 3: Let's move these to `conversions/mod.rs` which allows you to drop one level of nesting.

```suggestion
```

src/conversions/mod.rs
```diff
@@ -1,5 +1,6 @@
 //! This module contains conversions between various Rust object and their representation in Python.
 
 mod array;
+mod indexmap;
```
davidhewitt, line 4: As per above:

```suggestion
#[cfg(feature = "indexmap")]
#[cfg_attr(docsrs, doc(cfg(feature = "indexmap")))]
mod indexmap;
```

src/conversions/indexmap.rs
```diff
@@ -0,0 +1,132 @@
...
+                ret.insert(K::extract(k)?, V::extract(v)?);
+            }
+            Ok(ret)
+        }
+    }
+}
+
+#[cfg(feature = "indexmap")]
+#[cfg(test)]
+mod test_indexmap {
```
davidhewitt, line 53: With the cfg above moved to `conversions/mod.rs`, don't need to repeat it here any more.

Also let's call the test module `tests`, as that's a pretty universal convention.

```suggestion
#[cfg(test)]
mod tests {
```

src/conversions/indexmap.rs
```diff
@@ -0,0 +1,132 @@
...
+
+            let py_map = map.into_py_dict(py);
+
+            assert_eq!(py_map.len(), 1);
+            assert_eq!(py_map.get_item(1).unwrap().extract::<i32>().unwrap(), 1);
+        });
+    }
+
+    #[test]
+    fn test_indexmap_indexmap_insertion_order_round_trip() {
```
davidhewitt, line 104: 👍 This is great!

davidhewitt (approved):
This looks great, just some tiny nits and it might be nice to squash the commits if you're willing to do that too.

EDIT: I see there is a suggestion for more documentation too, which would also be much appreciated. It can go at the top of `indexmap.rs` as an inline comment.

Cargo.toml
```diff
@@ -27,6 +27,7 @@ paste = { version = "0.1.18", optional = true }
 pyo3-macros = { path = "pyo3-macros", version = "=0.14.1", optional = true }
 unindent = { version = "0.1.4", optional = true }
 hashbrown = { version = ">= 0.9, < 0.12", optional = true }
+indexmap = { version = "1.6.2", optional = true }
```
davidhewitt, line 30: Why is it fixed to the exact patch version? Could allow maybe `>= 1.0, < 2` here, perhaps.

IvanIsCoding: I fixed it to the latest version that was compatible with PyO3's MSRV. I will try something similar to hashbrown to have more flexibility

davidhewitt:
👍 looking great. Just waiting on a doc at the top of `indexmap.rs` as per https://github.com/PyO3/pyo3/pull/1728#issuecomment-881944913, and then this is ready to merge!

Thanks very much for your work and refinements on this!

IvanIsCoding:
>  looking great. Just waiting on a doc at the top of `indexmap.rs` as per [#1728 (comment)](https://github.com/PyO3/pyo3/pull/1728#issuecomment-881944913), and then this is ready to merge!
> 
> Thanks very much for your work and refinements on this!

I have added the doc at the top of the file, hopefully it gets the point accross

src/conversions/indexmap.rs
```diff
@@ -1,3 +1,97 @@
+//!  Conversions to and from [indexmap](https://docs.rs/indexmap/)’s
+//! `IndexMap`.
+//!
+//! [`indexmap::IndexMap`] is a hash table that is closely compatible with the standard [`std::collections::HashMap`],
+//! with the difference that it preserves the insertion order when iterating over keys. It was inspired
+//! by Python's 3.6+ dict implementation.
+//!
+//! Dictionary order is guaranteed to be insertion order since Python 3.7, and the conversions with IndexMap will reflect
```
mejrs, line 8: Given that people don't need to worry about this, I probably wouldn't mention the order guarantees at all.

src/conversions/indexmap.rs
```diff
@@ -1,3 +1,97 @@
...
+//! ```python
+//! from my_module import calculate_statistics
+//!
+//! data = [1, 1, 1, 3, 4, 5]
+//! print(calculate_statistics(data))
+//! # always prints {"median": 2.0, "mean": 2.5, "mode": 1.0} in the same order
+//! # if another hash table was used, the order could be random
+//! ```
+
 mod indexmap_indexmap_conversion {
```
mejrs, line 95: You can omit this module wrapper 😉 It's already in its own module.

mejrs (commented):
Thanks 💖

Is this documentation reachable in the docs? It's in a private module, so it's not (?) visible in the documentation, and there's no link to it like the other features for conversions.

IvanIsCoding:
> Thanks 
> 
> Is this documentation reachable in the docs? It's in a private module, so it's not (?) visible in the documentation, and there's no link to it like the other features for conversions.

I edited `lib.rs` to export the module, so the document should be visible now

mejrs (approved):

davidhewitt:
Looks brilliant. Thank you very much for your work on this!

Issue #1616, opened 2021-05-18
Python 3.x interpreter not found


Johnz86:
### 🌍 Environment

 - Your operating system and version: Microsoft Windows 10 Enterprise, 10.0.18363 Build 18363
 - Your python version: Python 3.8.5
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: no
 - Your Rust version (`rustc --version`): rustc 1.52.1 (9bc8c42bb 2021-05-09)
 - Your PyO3 version: 0.13.2
 - Have you tried using latest PyO3 main (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: yes

### 💥 Reproducing
Check the python version available with commands
```
PS C:\GIT\Rust\call-python> py --version
Python 3.8.5
```
I set the python3 command alias in powershell with:
```
Set-Alias python3 py
Set-Alias python py
```
And check the python3 interpreter availability
```
PS C:\GIT\Rust\call-python> python --version
Python 3.8.5
PS C:\GIT\Rust\call-python> python3 --version
Python 3.8.5
```
cargo.toml
```toml
[package]
name = "call-python"
version = "0.3.0"
authors = ["Johnz86"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies.pyo3]
version = "0.13.2"
features = ["auto-initialize"]
```
main.rs
```Rust
use pyo3::prelude::*;
use pyo3::types::IntoPyDict;

fn main() -> Result<(), ()> {
    Python::with_gil(|py| {
        main_(py).map_err(|e| {
          // We can't display Python exceptions via std::fmt::Display,
          // so print the error here manually.
          e.print_and_set_sys_last_vars(py);
        })
    })
}

fn main_(py: Python) -> PyResult<()> {
    let sys = py.import("sys")?;
    let version: String = sys.get("version")?.extract()?;
    let locals = [("os", py.import("os")?)].into_py_dict(py);
    let code = "os.getenv('USER') or os.getenv('USERNAME') or 'Unknown'";
    let user: String = py.eval(code, None, Some(&locals))?.extract()?;
    println!("Hello {}, I'm Python {}", user, version);
    Ok(())
}
```
Try to run the project and fail
```
PS C:\GIT\Rust\call-python> cargo run
   Compiling proc-macro2 v1.0.26
   Compiling unicode-xid v0.2.2
[...]

PR #1729, opened 2021-07-17
pyo3-build-config: add print-config helper

davidhewitt:
This adds a `print-config` bin to the `pyo3-build-config` crate.

The idea is that we can use this to help debug installation issues, e.g. #1616 

Some example outputs on my machine (Ubuntu 20.04, WSL2):

```
cargo:rerun-if-env-changed=PYO3_PYTHON
cargo:rerun-if-env-changed=VIRTUAL_ENV
cargo:rerun-if-env-changed=CONDA_PREFIX
cargo:rerun-if-env-changed=PATH
implementation: CPython
interpreter version: 3.8
interpreter path: Some("/usr/bin/python")
libdir: Some("/usr/lib")
shared: true
base prefix: Some("/usr")
ld_version: Some("3.8")
pointer width: Some(8)
```

and with a 3.9 virtualenv up:

```
cargo:rerun-if-env-changed=PYO3_PYTHON
cargo:rerun-if-env-changed=VIRTUAL_ENV
cargo:rerun-if-env-changed=CONDA_PREFIX
implementation: CPython
interpreter version: 3.9
interpreter path: Some("/home/david/.virtualenvs/pyo3/bin/python")
libdir: Some("/usr/lib/x86_64-linux-gnu")
shared: true
base prefix: Some("/usr")
ld_version: Some("3.9")
pointer width: Some(8)
```

### DIFF
--- a/pyo3-build-config/src/bin/print-config.rs
+++ b/pyo3-build-config/src/bin/print-config.rs
@@ -1,0 +1,16 @@
+use pyo3_build_config::{find_interpreter, get_config_from_interpreter};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let config = get_config_from_interpreter(&find_interpreter()?)?;
+
+    println!("implementation: {}", config.implementation);
+    println!("interpreter version: {}", config.version);
+    println!("interpreter path: {:?}", config.executable);
+    println!("libdir: {:?}", config.libdir);
+    println!("shared: {}", config.shared);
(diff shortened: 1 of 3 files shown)


messense (approved):
This is good, maybe we should also add some docs about this utility somewhere.

davidhewitt:
I agree some docs would be useful. Maybe in the FAQ, if we can find the right words?

For now I was just planning to ask users to run it if they report an issue which we can't reproduce / sounds like an environment error.

PR #1733, opened 2021-07-19
Tuple/List get_item to fallible

aviramha:
Hey, trying to help with:
https://github.com/PyO3/pyo3/issues/1667
specifically:
make PyTuple::get_item and PyList::get_item fallible
I'm not quite sure about the extract thingy, any guidance or feedback will be welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- Add `PyList::get_item_unchecked()` and `PyTuple::get_item_unchecked()` to get items without bounds checks. [#1733](https://github.com/PyO3/pyo3/pull/1733)
 - Add `PyAny::py()` as a convenience for `PyNativeType::py()`. [#1751](https://github.com/PyO3/pyo3/pull/1751)
 
 ### Changed
 
 - Change `PyErr::fetch()` to return `Option<PyErr>`. [#1717](https://github.com/PyO3/pyo3/pull/1717)
+- `PyList`, `PyTuple` and `PySequence`'s `get_item` now accepts only `usize` indices instead of `isize`. [#1733](https://github.com/PyO3/pyo3/pull/1733)
+- `PyList` and `PyTuple`'s `get_item` now return `PyResult<&PyAny> instead of panicking. [#1733](https://github.com/PyO3/pyo3/pull/1733)
 - Deprecate `PyTuple::split_from`. [#1804](https://github.com/PyO3/pyo3/pull/1804)
 
 ### Fixed
--- a/benches/bench_list.rs
+++ b/benches/bench_list.rs
@@ -32,7 +32,22 @@ fn list_get_item(b: &mut Bencher) {
     let mut sum = 0;
     b.iter(|| {
         for i in 0..LEN {
-            sum += list.get_item(i as isize).extract::<usize>().unwrap();
+            sum += list.get_item(i).unwrap().extract::<usize>().unwrap();
+        }
+    });
+}
+
+fn list_get_item_unchecked(b: &mut Bencher) {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+    const LEN: usize = 50_000;
+    let list = PyList::new(py, 0..LEN);
+    let mut sum = 0;
+    b.iter(|| {
+        for i in 0..LEN {
+            unsafe {
+                sum += list.get_item_unchecked(i).extract::<usize>().unwrap();
+            }
         }
     });
 }
@@ -41,6 +56,7 @@ fn criterion_benchmark(c: &mut Criterion) {
     c.bench_function("iter_list", iter_list);
     c.bench_function("list_new", list_new);
     c.bench_function("list_get_item", list_get_item);
+    c.bench_function("list_get_item_unchecked", list_get_item_unchecked);
 }
 
 criterion_group!(benches, criterion_benchmark);
--- a/benches/bench_tuple.rs
+++ b/benches/bench_tuple.rs
@@ -32,7 +32,22 @@ fn tuple_get_item(b: &mut Bencher) {
     let mut sum = 0;
     b.iter(|| {
         for i in 0..LEN {
-            sum += tuple.get_item(i).extract::<usize>().unwrap();
+            sum += tuple.get_item(i).unwrap().extract::<usize>().unwrap();
+        }
+    });
+}
+
+fn tuple_get_item_unchecked(b: &mut Bencher) {
+    let gil = Python::acquire_gil();
+    let py = gil.python();
+    const LEN: usize = 50_000;
+    let tuple = PyTuple::new(py, 0..LEN);
+    let mut sum = 0;
+    b.iter(|| {
+        for i in 0..LEN {
+            unsafe {
+                sum += tuple.get_item_unchecked(i).extract::<usize>().unwrap();
+            }
         }
     });
 }
@@ -41,6 +56,7 @@ fn criterion_benchmark(c: &mut Criterion) {
     c.bench_function("iter_tuple", iter_tuple);
     c.bench_function("tuple_new", tuple_new);
     c.bench_function("tuple_get_item", tuple_get_item);
+    c.bench_function("tuple_get_item_unchecked", tuple_get_item_unchecked);
 }
 
 criterion_group!(benches, criterion_benchmark);
--- a/pyo3-macros-backend/src/from_pyobject.rs
+++ b/pyo3-macros-backend/src/from_pyobject.rs
@@ -243,7 +243,7 @@ impl<'a> Container<'a> {
         for i in 0..len {
             let error_msg = format!("failed to extract field {}.{}", quote!(#self_ty), i);
             fields.push(quote!(
-                s.get_item(#i).extract().map_err(|inner| {
+                s.get_item(#i).and_then(PyAny::extract).map_err(|inner| {
                 let py = pyo3::PyNativeType::py(obj);
                 let new_err = pyo3::exceptions::PyTypeError::new_err(#error_msg);
                 new_err.set_cause(py, Some(inner));
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -66,21 +66,37 @@ impl PyList {
         self.len() == 0
     }
 
-    /// Gets the item at the specified index.
-    ///
-    /// Panics if the index is out of range.
-    pub fn get_item(&self, index: isize) -> &PyAny {
-        assert!(index >= 0 && index < self.len() as isize);
+    /// Gets the list item at the specified index.
+    /// # Example
+    /// ```
+    /// use pyo3::{prelude::*, types::PyList};
+    /// Python::with_gil(|py| {
+    ///     let list = PyList::new(py, &[2, 3, 5, 7]);
+    ///     let obj = list.get_item(0);
+    ///     assert_eq!(obj.unwrap().extract::<i32>().unwrap(), 2);
+    /// });
+    /// ```
+    pub fn get_item(&self, index: usize) -> PyResult<&PyAny> {
         unsafe {
-            #[cfg(not(Py_LIMITED_API))]
-            let ptr = ffi::PyList_GET_ITEM(self.as_ptr(), index as Py_ssize_t);
-            #[cfg(Py_LIMITED_API)]
-            let ptr = ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t);
-
+            let item = ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t);
             // PyList_GetItem return borrowed ptr; must make owned for safety (see #890).
-            ffi::Py_INCREF(ptr);
-            self.py().from_owned_ptr(ptr)
+            ffi::Py_XINCREF(item);
+            self.py().from_owned_ptr_or_err(item)
         }
+    }
+
+    /// Gets the list item at the specified index. Undefined behavior on bad index. Use with caution.
+    ///
+    /// # Safety
+    ///
+    /// Caller must verify that the index is within the bounds of the list.
+    #[cfg(not(any(Py_LIMITED_API, PyPy)))]
+    #[cfg_attr(docsrs, doc(cfg(not(any(Py_LIMITED_API, PyPy)))))]
+    pub unsafe fn get_item_unchecked(&self, index: usize) -> &PyAny {
+        let item = ffi::PyList_GET_ITEM(self.as_ptr(), index as Py_ssize_t);
+        // PyList_GET_ITEM return borrowed ptr; must make owned for safety (see #890).
+        ffi::Py_XINCREF(item);
+        self.py().from_owned_ptr(item)
     }
 
     /// Takes the slice `self[low:high]` and returns it as a new list.
@@ -163,7 +179,7 @@ impl PyList {
 /// Used by `PyList::iter()`.
 pub struct PyListIterator<'a> {
     list: &'a PyList,
-    index: isize,
+    index: usize,
 }
 
 impl<'a> Iterator for PyListIterator<'a> {
@@ -171,8 +187,11 @@ impl<'a> Iterator for PyListIterator<'a> {
 
     #[inline]
     fn next(&mut self) -> Option<&'a PyAny> {
-        if self.index < self.list.len() as isize {
-            let item = self.list.get_item(self.index);
+        if self.index < self.list.len() {
+            #[cfg(any(Py_LIMITED_API, PyPy))]
+            let item = self.list.get_item(self.index).expect("tuple.get failed");
+            #[cfg(not(any(Py_LIMITED_API, PyPy)))]
+            let item = unsafe { self.list.get_item_unchecked(self.index) };
             self.index += 1;
             Some(item)
         } else {
@@ -185,8 +204,8 @@ impl<'a> Iterator for PyListIterator<'a> {
         let len = self.list.len();
 
         (
-            len.saturating_sub(self.index as usize),
-            Some(len.saturating_sub(self.index as usize)),
+            len.saturating_sub(self.index),
+            Some(len.saturating_sub(self.index)),
         )
     }
 }
@@ -237,10 +256,10 @@ mod tests {
     fn test_new() {
         Python::with_gil(|py| {
             let list = PyList::new(py, &[2, 3, 5, 7]);
-            assert_eq!(2, list.get_item(0).extract::<i32>().unwrap());
-            assert_eq!(3, list.get_item(1).extract::<i32>().unwrap());
-            assert_eq!(5, list.get_item(2).extract::<i32>().unwrap());
-            assert_eq!(7, list.get_item(3).extract::<i32>().unwrap());
+            assert_eq!(2, list.get_item(0).unwrap().extract::<i32>().unwrap());
+            assert_eq!(3, list.get_item(1).unwrap().extract::<i32>().unwrap());
+            assert_eq!(5, list.get_item(2).unwrap().extract::<i32>().unwrap());
+            assert_eq!(7, list.get_item(3).unwrap().extract::<i32>().unwrap());
         });
     }
 
@@ -256,19 +275,10 @@ mod tests {
     fn test_get_item() {
         Python::with_gil(|py| {
             let list = PyList::new(py, &[2, 3, 5, 7]);
-            assert_eq!(2, list.get_item(0).extract::<i32>().unwrap());
-            assert_eq!(3, list.get_item(1).extract::<i32>().unwrap());
-            assert_eq!(5, list.get_item(2).extract::<i32>().unwrap());
-            assert_eq!(7, list.get_item(3).extract::<i32>().unwrap());
-        });
-    }
-
-    #[test]
-    #[should_panic]
-    fn test_get_item_invalid() {
-        Python::with_gil(|py| {
-            let list = PyList::new(py, &[2, 3, 5, 7]);
-            list.get_item(-1);
+            assert_eq!(2, list.get_item(0).unwrap().extract::<i32>().unwrap());
+            assert_eq!(3, list.get_item(1).unwrap().extract::<i32>().unwrap());
+            assert_eq!(5, list.get_item(2).unwrap().extract::<i32>().unwrap());
+            assert_eq!(7, list.get_item(3).unwrap().extract::<i32>().unwrap());
         });
     }
 
@@ -289,9 +299,9 @@ mod tests {
             let list = PyList::new(py, &[2, 3, 5, 7]);
             let val = 42i32.to_object(py);
             let val2 = 42i32.to_object(py);
-            assert_eq!(2, list.get_item(0).extract::<i32>().unwrap());
+            assert_eq!(2, list.get_item(0).unwrap().extract::<i32>().unwrap());
             list.set_item(0, val).unwrap();
-            assert_eq!(42, list.get_item(0).extract::<i32>().unwrap());
+            assert_eq!(42, list.get_item(0).unwrap().extract::<i32>().unwrap());
             assert!(list.set_item(10, val2).is_err());
         });
     }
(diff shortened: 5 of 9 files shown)


src/types/tuple.rs
```diff
@@ -458,4 +470,12 @@ mod test {
             );
         })
     }
+
+    #[test]
+    fn test_invalid_get_item() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
+        let ob = PyTuple::new(py, &[1, 2, 3]);
+        assert!(ob.get_item(100).is_err())
```
davidhewitt, line 479: 👍 good idea to add a test, it could be nice to also check the error type (`IndexError`?) or message?

src/types/tuple.rs
```diff
@@ -124,9 +132,13 @@ impl<'a> Iterator for PyTupleIterator<'a> {
...
-            let item = self.tuple.get_item(self.index);
-            self.index += 1;
-            Some(item)
+            match self.tuple.get_item(self.index) {
+                Ok(item) => {
+                    self.index += 1;
+                    Some(item)
+                }
+                _ => None,
+            }
```
davidhewitt, line 141: I think it's fine to use `.expect` here.


```suggestion
            let item = self.tuple.get_item(self.index).expect("index is less than length");;
            self.index += 1;
            Some(item)
```

If you don't feel comfortable about the possble panic, it would also be possible to implement a `#[cfg(not(Py_LIMITED_API))]` version of `PyTupleIterator` which uses `as_slice`.

davidhewitt (commented):
Thanks for kicking this off!

Another way to design this which is potentially for backwards compatibility is to deprecate `.get_item()` and to introduce two new methods:
 - `.get() -> PyResult<&PyAny>` 
 - `.get_unchecked() -> &PyAny`, doesn't do checking and only available on `#[cfg(not(Py_LIMITED_API))]`

Or `.get() -> Option<&PyAny>`, if handling the error internally feels appropriate. I'd be open to opinions on this.

Regarding the `.extract()` error, you should check for uses of `.get_item()` inside `pyo3-macros-backend`. TBH this is an example why deprecating `get_item` might be the better idea - the compilation will still go through and you'll just get warnings of where it's used, which can be fixed instead.

src/types/tuple.rs
aviramha: Overall I'm against having safe code that can panic (even if it's induced by very odd human error)
I'll check the second suggestion.

aviramha:
> Or .get() -> Option<&PyAny>, if handling the error internally feels appropriate. I'd be open to opinions on this.

I like this better. I don't like that the user might panic when using the API (while the API is safe).

aviramha:
@davidhewitt I've changed - I wasn't sure if to handle the deprecations now or maybe later?

src/types/list.rs
```diff
@@ -83,6 +84,23 @@ impl PyList {
...
+    pub fn get(&self, index: usize) -> PyResult<&PyAny> {
+        unsafe {
+            let item = ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t);
+            self.py().from_borrowed_ptr_or_err(item)
+        }
+    }
+
+    /// Gets the list item at the specified index. Undefined behavior on bad index. Use with caution.
+    #[cfg(not(any(Py_LIMITED_API, PyPy)))]
+    pub fn get_unchecked(&self, index: usize) -> &PyAny {
```
davidhewitt, line 97: This one is definitely `unsafe`:

```suggestion
    /// Gets the list item at the specified index. Undefined behavior on bad index. Use with caution.
    #[cfg(not(any(Py_LIMITED_API, PyPy)))]
    pub unsafe fn get_unchecked(&self, index: usize) -> &PyAny {
```

src/types/list.rs
```diff
@@ -431,7 +440,48 @@ mod test {
...
-        assert_eq!(1, list.get_item(0).extract::<i32>().unwrap());
-        assert_eq!(2, list.get_item(1).extract::<i32>().unwrap());
+        assert_eq!(1, list.get(0).unwrap().extract::<i32>().unwrap());
+        assert_eq!(2, list.get(1).unwrap().extract::<i32>().unwrap());
+    }
+
+    #[test]
+    fn test_list_get_invalid_index() {
+        let gil = Python::acquire_gil();
+        let py = gil.python();
```
davidhewitt, line 450: Can we use `Python::with_gil` in new tests please :)

src/types/tuple.rs
```diff
@@ -76,6 +76,7 @@ impl PyTuple {
     /// Gets the tuple item at the specified index.
     ///
     /// Panics if the index is out of range.
+    #[deprecated = "this is a deprecated alias for get and get_unchecked"]
```
davidhewitt, line 79:
```suggestion
    #[deprecated = "use PyList::get, which does not panic"]
```

src/types/tuple.rs
```diff
@@ -124,7 +142,7 @@ impl<'a> Iterator for PyTupleIterator<'a> {
     #[inline]
     fn next(&mut self) -> Option<&'a PyAny> {
         if self.index < self.length {
-            let item = self.tuple.get_item(self.index);
+            let item = self.tuple.get(self.index).expect("tuple.get failed");
```
davidhewitt, line 145: This can use `get_unchecked` when available.

src/types/tuple.rs
```diff
@@ -201,7 +219,7 @@ macro_rules! tuple_conversion ({$length:expr,$(($refN:ident, $n:tt, $T:ident)),+
             let t = <PyTuple as PyTryFrom>::try_from(obj)?;
             if t.len() == $length {
                 Ok((
-                    $(t.get_item($n).extract::<$T>()?,)+
+                    $(t.get($n)?.extract::<$T>()?,)+
```
davidhewitt, line 222: Same here r.e. `get_unchecked`.

davidhewitt (commented):
Thanks, this is taking shape!

@birkenfeld as you came up with the proposals in #1667 I'd be really interested for your opinion on this.

mejrs (commented):
Thanks :)

I do have some suggestions..

src/types/list.rs
```diff
@@ -83,6 +84,24 @@ impl PyList {
         }
     }
 
+    /// Gets the list item at the specified index.
+    pub fn get(&self, index: usize) -> PyResult<&PyAny> {
```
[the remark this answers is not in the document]
aviramha: We want negative indexing, but when using index (slicing). There's no real sense in implementing it at this function level.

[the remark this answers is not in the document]
mejrs: Personally I'm leaning towards using usize everywhere, anyway 👍

mejrs (approved):

src/types/list.rs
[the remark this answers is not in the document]
birkenfeld: I don't see why this should only be relevant for slicing. Python allows negative indices for single item access as well as for slices.  IMO either we allow negative indices for both, or none.

src/types/list.rs
```diff
@@ -142,16 +171,19 @@ impl PyList {
...
 impl<'a> Iterator for PyListIterator<'a> {
     type Item = &'a PyAny;
 
     #[inline]
     fn next(&mut self) -> Option<&'a PyAny> {
-        if self.index < self.list.len() as isize {
-            let item = self.list.get_item(self.index);
+        if self.index < self.list.len() {
+            #[cfg(any(Py_LIMITED_API, PyPy))]
+            let item = self.list.get(self.index).expect("tuple.get failed");
```
birkenfeld, line 184: It's a list here

birkenfeld (changes requested):
I'm happy with returning `PyResult`.  The new name makes the API asymmetric (`get` vs `set_item`, and btw why is there is no `del_item` or `remove`...)  Are we still allowed to break APIs if the result is a clear compile error?

`usize` vs `isize` needs to be discussed still.

A follow-up to this is the implementation of `Index` traits.

davidhewitt:
> The new name makes the API asymmetric (get vs set_item, and btw why is there is no del_item or remove...) Are we still allowed to break APIs if the result is a clear compile error?

Yes, we could change `get_item` in 0.15. I do agree we should keep consistent names; maybe it is better to have `get_item` and `get_item_unchecked` here. I think I prefer having `isize` because it's more similar to the Python semantics, but I could be persuaded either way. The underlying C-API only accepts unsigned integers, so we'd need to do the conversion (but that's not _too_ bad)...

Sorry @aviramha you picked a ticket which has a lot of open design questions 🙃 

I definitely think if we call it `.get()` then we should accept `usize` only.

aviramha:
> >  Sorry @aviramha you picked a ticket which has a lot of open design questions 🙃

That's OK. the get_item of Python is actually usize, so it might be more confusing that it works with negative indices. I guess it's a question whether you want the API to feel like a Rust wrapper to the CPython or a rust wrapper to the Python API.
Let me know which you prefer and I'll fix the PR.

davidhewitt:
That's a fair question. I think I'm now leaning towards this matching the C API, so would vote for `get_item(usize) -> PyResult<&PyAny>` and `unsafe get_item_unchecked(usize) -> &PyAny`.

aviramha:
> That's a fair question. I think I'm now leaning towards this matching the C API, so would vote for `get_item(usize) -> PyResult<&PyAny>` and `unsafe get_item_unchecked(usize) -> &PyAny`.

Goes without saying that I am too. How do we make this call? 🙃

davidhewitt:
If it looks like a majority of us agree, I'm happy for us to commit to this plan (cc @birkenfeld @mejrs @messense).

birkenfeld:
Where would `isize` be allowed then? This should really be designed keeping all places with indices in mind (see #1667), not decided in a case-by-case basis just looking at the C API, which is inconsistent. For example, `PyList_Insert` accepts negative indices.

birkenfeld:
To be clear, I'm not opposed to doing away with `isize` everywhere and only accepting "real" indices, Rust-std-style.  I just want it to be consistent.

aviramha:
@birkenfeld I understand your concern but I believe that's CPython inconsistency issue. It depends on what we value more - being consistent with it or improving it.

aviramha:
To make myself more clear (previous comment was from mobile) - when coding in PyO3 I often use the CPython docs to understand how stuff works - having different API (i.e get_item of PyO3 accepts negative indices vs CPython not) can easily cause confusion for anyone.

birkenfeld:
The expectation of each PyO3 method mapping exactly to a Python C-API call is misguided, though.  If you are missing info from the PyO3 docs which is integral to understanding the API in question, it is a bug and should be fixed; please report that here.

And of course, in the case we're discussing here, the documentation needs to state that negative indices are supported and how they are interpreted. (That this is not documented currently is one of the points of #1667).

davidhewitt:
I definitely agree that we should go for all-`isize`, or all-`usize`. It's not fun as a user to have to look up argument types for every function. (And the right document source should definitely be PyO3 docs, not CPython docs.)

Here's the arguments I can see either way:

---
#### usize
Good:
  - Matches `std`
  - Means we can use the `SliceIndex` trait to allow doing `list[0..5]` etc

Bad:
  - `PySequence` and `PyList_Insert` APIs work correctly with negative indices, but we'd be preventing users from doing so.

---
#### isize
Good:
  - Matches python usage, e.g. `list[-5]`
  - We'd have to work around APIs that don't support negative indexing. It's probably as simple as using `PySequence_GetItem` instead of `PyList_GetItem`, and similar, but not sure. Some edge cases might be annoying.

Bad:
   - Supporting negative indexing presumably has a (tiny) performance overhead, because negative indices must be converted to actual positive ones.
   - To have slices with negative indices such as `list[-4..-2]` we'd probably need to create a `PySliceIndex` trait which is implemented for various `Range<isize>`, `RangeFrom<isize>`, `RangeFull<isize>`. Could be a lot of work.
   - `PyList::get_item_unchecked` would almost certainly want to take `usize`. Either that or it could have a simple check `if idx < self.len() { idx += self.len() }`, But that would again imply performance overhead, which kinda sucks in an `unsafe` low-level API.

---

An alternative... what if we made `PyList` take `usize` and `PySequence` take `isize`? That way we'd get the simplicity of `usize` for most usage but also give users the "full API" if they chose to treat the list as a sequence. We'd have to document this really clearly, but at least there's then a clear pattern to it.

I think this "mixed-mode" would match the C-API in every way except `PyList::insert`, which we would make only take `usize`.

mejrs:
@davidhewitt thanks for the recap

Indexing and slicing with negative numbers doesn't seem like a good idea to me:
- For someone not coming from Python, having negative indices work when indexing/slicing is something that I feel will mostly just be surprising and unexpected.
- For someone who *is* coming from Python (to Rust), I think it'll be confusing if negative indexing works somewhere (pyo3) but not in other places (the rest of the Rust ecosystem).
- It will only be a convenience for people making Python libraries rather than the people using said libraries. After all, authors can implement negative slicing and indexing for their own types.
- I'm not a fan of doing a bunch of math tricks in what should be a low level api.

davidhewitt:
Agreed, I'm now feeling very sure that `usize` is definitely the correct choice for `PyList`.

The question for me then becomes whether `PySequence`should be based around `isize` to match the underlying Python APIs. On balance, I think I slightly err towards `PuSequence` using `usize` too for simplicity.

aviramha:
@davidhewitt can I finish the PR then? :) (finalizing the changes)

davidhewitt:
Sure thing, let's go with the `usize` plan!

birkenfeld:
Agreed!

aviramha:
@davidhewitt @birkenfeld should I include anything else then in this PR?

davidhewitt:
I understood we decided above that we should have `get_item(&self, usize) -> PyResult<&PyAny>` and `unsafe.get_item_unchecked(&self, usize) -> &PyAny`?

The `.get()` suggestion by me will need to be rolled back - sorry about that.

This also will need a CHANGELOG entry.

In addition, if we're changing `PyList` and `PyTuple` APIs to take `usize` in this PR it would be ideal if other relevant APIs were updated at the same time. (Although this is not entirely necessary.)

I'm about to release 0.14.2 tomorrow, and happy to see this merged after that.

aviramha:
@davidhewitt should I change get_slice to also use usize?

aviramha:
I'm not sure how I broke the extract thingy

davidhewitt:
Check for uses of .get_item inside `pyo3-macros-backend`. I expect code is being generated which used to just use the value, but now there's results being returned instead.

aviramha:
Cool. @davidhewitt Waiting for your review :

davidhewitt:
👍 will try to give a full read within the next few days!

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix regression in 0.13.2 leading to linking to incorrect Python library on Windows "gnu" targets. [#1759](https://github.com/PyO3/pyo3/pull/1759)
 - Fix compiler warning: deny trailing semicolons in expression macro. [#1762](https://github.com/PyO3/pyo3/pull/1762)
 
+### Changed
+- `PyList`, `PyTuple` and `PySequence`'s `get_item` now accepts only `usize` indices instead of `isize`.
+- `PyList` and `PyTuple`'s `get_item` now always uses the safe API. See `get_item_unchecked` for retrieving index without checks.
```
davidhewitt, line 32: Afraid since we released 0.14.2 you'll need to rebase on `main` and move this CHANGELOG entry up to the new "Unreleased" section building for 0.15.

I think for users the return type is more relevant, needs PR number, and I'd mention `get_item_unchecked` separately:

```suggestion
### Added
- Add `PyList::get_item_unchecked()` and `PyTuple::get_item_unchecked()` to get items without bounds checks. [#1733](https://github.com/PyO3/pyo3/pull/1733)

### Changed 
- `PyList` and `PyTuple`'s `get_item` now return `PyResult<&PyAny> instead of panicking. [#1733](https://github.com/PyO3/pyo3/pull/1733)
```

benches/bench_list.rs
```diff
@@ -32,7 +32,7 @@ fn list_get_item(b: &mut Bencher) {
     let mut sum = 0;
     b.iter(|| {
         for i in 0..LEN {
-            sum += list.get_item(i as isize).extract::<usize>().unwrap();
+            sum += list.get_item(i).unwrap().extract::<usize>().unwrap();
```
davidhewitt, line 35: It might be nice to add a `list_get_item_unchecked` benchmark.

benches/bench_tuple.rs
```diff
@@ -32,7 +32,7 @@ fn tuple_get_item(b: &mut Bencher) {
     let mut sum = 0;
     b.iter(|| {
         for i in 0..LEN {
-            sum += tuple.get_item(i).extract::<usize>().unwrap();
+            sum += tuple.get_item(i).unwrap().extract::<usize>().unwrap();
```
davidhewitt, line 35: Similarly it might be nice to add a `tuple_get_item_unchecked` benchmark.

pyo3-macros-backend/src/from_pyobject.rs
```diff
@@ -243,7 +243,7 @@ impl<'a> Container<'a> {
         for i in 0..len {
             let error_msg = format!("failed to extract field {}.{}", quote!(#self_ty), i);
             fields.push(quote!(
-                s.get_item(#i).extract().map_err(|inner| {
+                s.get_item(#i)?.extract().map_err(|inner| {
```
davidhewitt, line 246: I think this probably needs a `.and_then` or something rather than returning early, because we still need the error message:

```suggestion
                s.get_item(#i).and_then(PyAny::extract).map_err(|inner| {
```

src/types/list.rs
```diff
@@ -66,23 +66,34 @@ impl PyList {
...
-            #[cfg(not(Py_LIMITED_API))]
-            let ptr = ffi::PyList_GET_ITEM(self.as_ptr(), index as Py_ssize_t);
-            #[cfg(Py_LIMITED_API)]
-            let ptr = ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t);
-
-            // PyList_GetItem return borrowed ptr; must make owned for safety (see #890).
-            ffi::Py_INCREF(ptr);
-            self.py().from_owned_ptr(ptr)
+            let item = ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t);
+            self.py().from_borrowed_ptr_or_err(item)
```
davidhewitt, line 82: The original code incremented the reference and used `from_owned_ptr()`.

That's still necessary here - the problem with `from_borrowed_ptr` is that if a user then edits the `PyList` the "borrowed" reference could be deleted out from underneath us.

See #890 for more detail.

```suggestion
            // PyList_GetItem return borrowed ptr; must make owned for safety (see #890).
            ffi::Py_INCREF(item);
            self.py().from_owned_ptr(item)
```

src/types/list.rs
```diff
@@ -66,23 +66,34 @@ impl PyList {
...
 
+    /// Gets the list item at the specified index. Undefined behavior on bad index. Use with caution.
+    ///
+    /// # Safety
+    ///
+    /// Caller must verify that the index is within the bounds of the list.
+    #[cfg(not(any(Py_LIMITED_API, PyPy)))]
+    pub unsafe fn get_item_unchecked(&self, index: usize) -> &PyAny {
+        let item = ffi::PyList_GET_ITEM(self.as_ptr(), index as Py_ssize_t);
+        self.py().from_borrowed_ptr(item)
```
davidhewitt, line 94: Same again here:

```suggestion
        // PyList_GET_ITEM return borrowed ptr; must make owned for safety (see #890).
        ffi::Py_INCREF(item);
        self.py().from_owned_ptr(item)
```

src/types/list.rs
```diff
@@ -142,16 +153,19 @@ impl PyList {
 /// Used by `PyList::iter()`.
 pub struct PyListIterator<'a> {
     list: &'a PyList,
-    index: isize,
+    index: usize,
```
davidhewitt, line 156: I think with this change there's some conversions in `size_hint` (`self.index as usize`) which aren't needed any more.

src/types/tuple.rs
```diff
@@ -200,9 +218,12 @@ macro_rules! tuple_conversion ({$length:expr,$(($refN:ident, $n:tt, $T:ident)),+
...
             if t.len() == $length {
-                Ok((
-                    $(t.get_item($n).extract::<$T>()?,)+
-                ))
+                #[cfg(any(Py_LIMITED_API, PyPy))]
+                return Ok(($(t.get_item($n)?.extract::<$T>()?,)+));
+
+                #[cfg(not(any(Py_LIMITED_API, PyPy)))]
+                unsafe {return Ok(($(t.get_item_unchecked($n).extract::<$T>()?,)+));}
+
```
davidhewitt, line 226: Nit: bonus newline

```suggestion
```

davidhewitt (changes requested):
Thanks for pushing forward on this! I'm afraid there's a few bits and pieces which still need tidying up. Comments scattered below.

> @davidhewitt should I change get_slice to also use usize?

Before we release 0.15 we'll need to change all these tuple / list / sequence APIs which currently take `isize` to take `usize`. This PR is already big enough, so maybe it's better to make a list of what still needs to change in #1667 and then make follow-up PRs.

aviramha:
Thanks for the review. I think I fixed all of the comments :)

src/types/tuple.rs
```diff
@@ -74,20 +74,35 @@ impl PyTuple {
...
+            self.py().from_borrowed_ptr_or_err(item)
         }
     }
 
+    /// Gets the tuple item at the specified index. Undefined behavior on bad index. Use with caution.
+    ///
+    /// # Safety
+    ///
+    /// Caller must verify that the index is within the bounds of the list.
+    #[cfg(not(any(Py_LIMITED_API, PyPy)))]
```
davidhewitt, line 100: Similarly here:

```suggestion
    #[cfg(not(any(Py_LIMITED_API, PyPy)))]
    #[cfg_attr(docsrs, doc(cfg(not(any(Py_LIMITED_API, PyPy))))]
```

davidhewitt (commented):
The changes are looking reasonable to me... but CI has a segfault?

src/types/list.rs
```diff
@@ -427,8 +436,49 @@ mod tests {
...
+    }
+
+    #[cfg(not(any(Py_LIMITED_API, PyPy)))]
+    #[test]
+    fn test_list_get_item_unchecked_invalid_index() {
+        Python::with_gil(|py| {
+            let list = PyList::new(py, &[2, 3, 5, 7]);
+            unsafe { list.get_item_unchecked(5) };
         });
     }
```
davidhewitt, line 483: Ah I guess this test is just not worth having - you're reading an invalid memory location. No wonder it crashes!

aviramha:
Oopsie, I thought the CI failure was main's fault

src/types/list.rs
```diff
@@ -66,23 +66,39 @@ impl PyList {
...
-            #[cfg(not(Py_LIMITED_API))]
-            let ptr = ffi::PyList_GET_ITEM(self.as_ptr(), index as Py_ssize_t);
-            #[cfg(Py_LIMITED_API)]
-            let ptr = ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t);
-
+            let item = ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t);
             // PyList_GetItem return borrowed ptr; must make owned for safety (see #890).
-            ffi::Py_INCREF(ptr);
-            self.py().from_owned_ptr(ptr)
+            ffi::Py_INCREF(item);
```
davidhewitt, line 83: Ahh, the segfault on CI is my fault. When there's an error item is null. Best way to handle that is use `Py_XINCREF`:

```suggestion
            ffi::Py_XINCREF(item);
```

davidhewitt (commented):
Ahh I think I've found another cause of segfault... Fingers crossed CI green after 🤞

davidhewitt (approved):
👍 looks brilliant! Thanks again for your patience and many iterations as we worked out the design for this. 

I'll merge this shortly (just going to cherry-pick some fixes for a 0.14.3 bugfix release first).

davidhewitt:
(See #1796 - as soon as that release is live I'll merge this. It just makes my life easier if I try to cherry-pick other stuff and there's no more breaking changes 😉)

birkenfeld (approved):
LGTM.

aviramha:
> 👍 looks brilliant! Thanks again for your patience and many iterations as we worked out the design for this. 
> 
> 
> 
> I'll merge this shortly (just going to cherry-pick some fixes for a 0.14.3 bugfix release first).

You're welcome! Glad to contribute to an awesome framework.
Where do we wish to track progress on this so I might pick up the next task?

davidhewitt:
> You're welcome! Glad to contribute to an awesome framework.
Where do we wish to track progress on this so I might pick up the next task?

🚀

If you want to own this, that would be awesome! The easiest way might be to create a new issue where you can own a list of checkboxes and close them as we make progress?

birkenfeld:
Since 0.14.x is going to be in a branch, can we merge this?

davidhewitt:
Ok 👍 I'm getting more used to doing the cherry-picking onto the 0.14.3 branch now, so let's proceed with this.

I see I created a merge conflict by merging #1802, so I'll fixup this quickly now...

PR #1734, opened 2021-07-20
guide: add latest symlink

davidhewitt:
This add a symlink to the `gh-pages` branch so that the latest released guide will always be available underneath `https://pyo3.rs/latest/`.

E.g. `https://pyo3.rs/latest/changelog.html` would be the link to the latest released changelog.

If approved, I'll also add this manually now to make `https://pyo3.rs/latest/` symlink to `https://pyo3.rs/v0.14.1/`.


Eric-Arellano (approved):
This sounds really useful!

messense (approved):

PR #1737, opened 2021-07-22
pypy: fix FFI definition of Py_Buffer

### LINKED ISSUE #1736: PyBuffer::get leads to memory corruption under PyPy
I've put up a repository containing a minimal repro of this bug: https://github.com/oconnor663/pyo3_pybuffer_repro. You can run the `repro.sh` script there to execute all the repro steps below.

I'm on x86-64 Arch Linux, with CPython 3.9.6, PyPy 7.3.5, and Rust 1.53.0. Here is a minimal Rust extension demonstrating this bug:

Cargo.toml:
```toml
[package]
name = "repro"
version = "0.1.0"
edition = "2018"

[lib]
name = "repro"
crate-type = ["cdylib"]

[dependencies.pyo3]
version = "0.14.1"
```

src/lib.rs:
```rust
use pyo3::buffer::PyBuffer;
use pyo3::prelude::*;

#[pyfunction]
fn buffer_len(obj: &PyAny) -> PyResult<usize> {
    let buffer = PyBuffer::<u8>::get(obj)?;              // <-- THIS IS THE INTERESTING LINE
    Ok(buffer.item_size() * buffer.item_count())
}

#[pymodule]
fn repro(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(buffer_len, m)?)?;
    Ok(())
}
```

test.py:
```python
from repro import buffer_len

assert buffer_len(b"foo") == 3
```

I create a `virtualenv -p pypy3`, and inside that virtualenv I build and install with extension module with `maturin develop`. (Note that this relies on the fix at https://github.com/PyO3/maturin/pull/596, which only just shipped in v0.11.2.) Finally I execute `test.py` over and over in a loop something like this in Bash (where `python` refers to the virtualenv Python interpreter):

```bash
i=1
while python ./test.py ; do
    echo "run number $i"
    i=$(($i + 1))
done
```

Here's an example run of `repro.sh` (which executes all the steps above) on my box:

```
❯ ./repro.sh
created virtual environment PyPy3.7.10.final.0-64 in 144ms
  creator PyPy3Posix(dest=/tmp/tmp.n8Evydf1LW, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/jacko/.local/share/virtualenv)
    added seed packages: pip==21.1.2, setuptools==57.0.0, wheel==0.36.2
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator
🔗 Found pyo3 bindings
🐍 Found PyPy 3.7 at python
   Compiling pyo3 v0.14.1
   Compiling repro v0.1.0 (/home/jacko/tmp/repro)
    Finished dev [unoptimized + debuginfo] target(s) in 3.63s
run number 1
run number 2
run number 3
malloc_consolidate(): invalid chunk size
./repro.sh: line 14: 125301 Aborted                 (core dumped) python ./test.py
```

Note that the assert in `test.py` did pass a few times before the crash in this case. On some runs it crashes immediately for me, and on others it gets through a handful of executions before crashing like it did here.

If I replace `virtualenv -p pypy3` with `virtualenv -p /usr/bin/python` (CPython 3.9.6), `test.py` always succeeds and that while loop runs forever.

davidhewitt:
Fixes #1736 

Looks like PyPy has extra fields for the `Py_Buffer` struct - https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/cpyext/parse/cpyext_object.h#L78

Needs a `CHANGELOG` entry and also tests (in `pyo3-pytests`, because this is PyPy) before this can be merged. Will write them later, got to go to work now! 💼

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Fix regression in 0.14.0 rejecting usage of `#[doc(hidden)]` on structs and functions annotated with PyO3 macros. [#1722](https://github.com/PyO3/pyo3/pull/1722)
 - Fix regression in 0.14.0 leading to incorrect code coverage being computed for `#[pyfunction]`s. [#1726](https://github.com/PyO3/pyo3/pull/1726)
+- Fix incorrect FFI definition of `Py_Buffer` on PyPy. [#1737](https://github.com/PyO3/pyo3/pull/1737)
 
 ## [0.14.1] - 2021-07-04
 
--- a/examples/pyo3-pytests/src/buf_and_str.rs
+++ b/examples/pyo3-pytests/src/buf_and_str.rs
@@ -1,4 +1,5 @@
 //! Objects related to PyBuffer and PyStr
+use pyo3::buffer::PyBuffer;
 use pyo3::prelude::*;
 use pyo3::types::{PyBytes, PyString};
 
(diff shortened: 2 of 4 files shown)


oconnor663:
Confirmed, this fixes the crash on my box!

messense (approved):
LGTM

PR #1738, opened 2021-07-24
docs: use pyo3.rs/latest instead of pyo3.rs/main

davidhewitt:
After #1734 I think it's desirable to link to `pyo3.rs/latest` (i.e. most recent released guide) instead of `pyo3.rs/main` within crate documentation. I guess we could also use the `#[doc]` attributes, but that's quite painful to maintain for a lot of these links.

cc @mejrs - I'd be interested to hear what you think about doing this?


messense (approved):

mejrs (approved):
Sounds like a good idea. I'm not hot for doing a lot of magic with `#[doc]` tags.

Issue #1719, opened 2021-07-06
error: invalid instruction `cargo:rustc-cdylib-link-arg` from build script of `pyo3 v0.14.1`


Hywan:
## 🐛 Bug Reports

I try to compile `wasmer-python` with pyo3 0.14.1. It uses pyo3 to create a Python extension with Rust. Whe compiling, I'm facing this error:

```
error: invalid instruction `cargo:rustc-cdylib-link-arg` from build script of `pyo3 v0.14.1`
The package pyo3 v0.14.1 does not have a cdylib target.
```

### 🌍 Environment

- macOS 11.1 **aarch64**
- Python 3.9.4 from Homebrew
- Rust 1.55 nightly
- PyO3 0.14.1

### 💥 Reproducing

When compiling `examples/word-count/` with `cargo build --release`, we get a warning:

```
warning: cargo:rustc-cdylib-link-arg was specified in the build script of pyo3 v0.14.1 (/Users/hwhost/Development/pyo3/pyo3), but that package does not contain a cdylib target

Allowing this was an unintended change in the 1.50 release, and may become an error in the future. For more information, see <https://github.com/rust-lang/cargo/issues/9562>.
warning: cargo:rustc-cdylib-link-arg was specified in the build script of pyo3 v0.14.1 (/Users/hwhost/Development/pyo3/pyo3), but that package does not contain a cdylib target

Allowing this was an unintended change in the 1.50 release, and may become an error in the future. For more information, see <https://github.com/rust-lang/cargo/issues/9562>.
```

In my project (`wasmer-python`), it's an error.

Hywan:
To conclude on a more positive aspect: PyO3 is awesome, and you're all doing an amazing job!

davidhewitt:
Ah, that's a shame. This is #1539, but I guess that https://github.com/rust-lang/cargo/issues/9562 is saying that was an unintended change.

This probably has implications for #1123, which is using similar cargo flags.

I wonder if we can provide some hooks in the `pyo3-build-config` crate which work around this.

@Hywan do I understand correctly that this is a blocking error for your attempting to upgrade? If so, we should probably figure out an appropriate patch fix...

Hywan:
Yeah it's an error that forbids me to update.

[...]

PR #1743, opened 2021-07-30
readme: slightly improve getting started notes

davidhewitt:
This PR tweaks the README a bit to prefer `maturin` to a manual build for users getting started with PyO3 for the first time.

Especially if we have to require users to add macOS config for manual `extension-module` builds (as per #1719), I think it's nice to give users something simple like `maturin` for their first experiments with PyO3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,17 +20,13 @@ PyO3 supports Python 3.6 and up. The minimum required Rust version is 1.41.
 
 PyPy is also supported. Some minor features are unavailable on PyPy - please refer to the [pypy section in the guide](https://pyo3.rs/latest/building_and_distribution/pypy.html) for more information.
 
-You can either write a native Python module in Rust, or use Python from a Rust binary.
+You can use PyO3 to write a native Python module in Rust, or to embed Python in a Rust binary. The following sections explain each of these in turn.
 
-However, on some OSs, you need some additional packages. E.g. if you are on _Ubuntu 18.04_, please run
+### Using Rust from Python
 
-```bash
-sudo apt install python3-dev python-dev
-```
+PyO3 can be used to generate a native Python module. The easiest way to try this out for the first time is to use [`maturin`](https://github.com/PyO3/maturin). `maturin` is a tool for building and publishing Rust-based Python packages with minimal configuration. The following steps set up some files for an example Python module, install `maturin`, and then show how build and import the Python module.
 
-## Using Rust from Python
-
-PyO3 can be used to generate a native Python module.
+First, create a new folder (let's call it `string_sum`) containing the following two files:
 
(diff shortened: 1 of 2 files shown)


Eric-Arellano (approved):
Nice improvement!

README.md
```diff
@@ -20,23 +20,20 @@ PyO3 supports Python 3.6 and up. The minimum required Rust version is 1.41.
...
-
-PyO3 can be used to generate a native Python module.
+First, create a new folder (let's call it `string_sum`) containing the following two files:
 
 **`Cargo.toml`**
 
 ```toml
 [package]
 name = "string-sum"
+authors = "PyO3 Contributors"
```
birkenfeld, line 36: Is `authors` not usually a list? And now it's optional anyway...

README.md
```diff
@@ -65,23 +62,50 @@ fn sum_as_string(a: usize, b: usize) -> PyResult<String> {
     Ok((a + b).to_string())
 }
 
-/// A Python module implemented in Rust.
+/// A Python module implemented in Rust. The name of this function should match
```
birkenfeld, line 65:
```suggestion
/// A Python module implemented in Rust. The name of this function must match
```

?

README.md
```diff
@@ -65,23 +62,50 @@ fn sum_as_string(a: usize, b: usize) -> PyResult<String> {
...
+'25'
+```
+
+As well as with `maturin`, it is possible to build using [`setuptools-rust`](https://github.com/PyO3/setuptools-rust) or [manually](https://pyo3.rs/latest/building_and_distribution.html#manual-builds). Both offer more flexibility than `maturin` but require further configuration.
+
+### Using Python from Rust
+
+To embed Python into a Rust binary, you need to ensure that your Python installation contains a shared library. The following steps demonstrate how to ensure this (for Ubuntu), and then give some example code which runs an embedded Python interpreter.
+
+To install the Python shared library, if you are on _Ubuntu 18.04_, you can run:
```
birkenfeld, line 102: This should work with any (supported) Ubuntu version.

README.md
messense: I think we should just remove it. https://github.com/rust-lang/rfcs/blob/master/text/3052-optional-authors-field.md

davidhewitt: Yep, this is a typo from me on top of the fact I was erroneously testing with an old `maturin` version which still required `authors`. Will remove 👍

README.md
davidhewitt: Yeah, technically they can write a custom import hook (or use `#[pyo3(name = "foobar")]` to workaround name mismatch, but for a README "must" is probably simpler.

birkenfeld (approved):

PR #1745, opened 2021-07-30
rust: updates for rust & clippy 1.54

davidhewitt:


Eric-Arellano (approved):

davidhewitt:
Windows 32-bit segfault is because `dictoffset` in the `PyCell` seems to be calculated incorrectly. I think the types in [pyclass_slots.rs](https://github.com/PyO3/pyo3/blob/main/src/pyclass_slots.rs) need to be `#[repr(C)]`, will confirm when I get home later

davidhewitt:
Looks like we calculated `dictoffset` incorrectly on 32-bit windows. The dictoffset was resolving to the padding bytes at the end of the `PyCell` struct. I guess that Rust 1.54 might have stopped zero-initialising the padding (or vice versa?) which caused the crash to become noticeable.

I've replaced the offset calculations with more robust ones.

birkenfeld:
Sounds like this should be released as 0.14.2?

davidhewitt:
Yes, I'm going to open a PR to fix #1719 and then prepare the full lot as 0.14.2 release. (Hopefully today, depending on some family stuff I've got to do first)

PR #1746, opened 2021-07-30
pyo3-macros-backend: support macros inside doc attributes

davidhewitt:
~~WIP. (Needs tests, CHANGELOG entry.)~~

In Rust 1.54 it's now valid to write 

```
#![doc = include_str!("README.md")]
struct Foo {}
```

This PR attempts to support these macros-inside-docs for `pyclass` and the other PyO3 macros. It works by collecting all the `#[doc]` attributes into one big `concat!` macro invocation, forwarding all the inner macros through so that `concat!` can evaluate them all and create the final docstring.


davidhewitt:
I think this is now ready for review if anyone's willing to take a look at it!

PR #1749, opened 2021-08-01
guide: rewrite "Building and Distribution" chapter

davidhewitt:
In preparation for documenting changes for #1719, I ended up rewriting the "building and distribution" chapter of the guide.

It's mostly a reorganisation and extra text to make it flow a bit better.

I ended up taking out the "PyPy" section as I don't think it's particularly informative. Our docs already have information on all APIs which aren't available for PyPy e.g. https://docs.rs/pyo3/0.14.1/pyo3/types/struct.PyDict.html#method.from_sequence, which seems good enough and much more likely to stay up-to-date than a list in the guide.


guide/src/building_and_distribution.md
```diff
@@ -178,11 +170,57 @@ The known complications are:
...
+
+cargo build --target x86_64-pc-windows-gnu
+```
+
+Any of the `abi3-py3*` features can be enabled instead of setting `PYO3_CROSS_PYTHON_VERSION` in the above examples.
 
+The following resources may also be useful for cross-compiling:
+ - [github.com/japaric/rust-cross](https://github.com/japaric/rust-cross) is a primer on cross compiling Rust.
+ - [github.com/rust-embedded/cross](https://github.com/rust-embedded/cross) uses Docker to make Rust cross-compilation easier.
+ - [github.com/messense/manylinux-cross](https://github.com/messense/manylinux-cross) contains custom Docker images for `cross` that are `manylinux` compliant. (Useful if cross-compiling Python extension modules.)
```
messense, line 219: Actually `manylinux-cross` docker images are not built for `cross`, it was built for cross compiling with maturin (utilized in [maturin-action](https://github.com/messense/maturin-action)). Maybe it could work with `cross` but I haven't tried yet.

davidhewitt: Ah ok, thanks 😅 . I'll take that link out for now, if they look like they work we can add it back later!

messense (approved):
Nice improvement!

guide/src/building_and_distribution.md
```diff
@@ -1,58 +1,91 @@
 # Building and Distribution
 
-## Python version
+This chapter of the guide goes into detail on how to build and distribute projects using PyO3. The way to achieve this is very different depending on whether the project is a Python module implemented in Rust, or a Rust binary embedding Python. For both types of project there are also common problems such as the Python version to build for and the [linker](https://en.wikipedia.org/wiki/Linker_(computing)) arguments to use.
+
+The material in this chapter is intended for users who have already read the PyO3 [README](#index.md). It covers in turn the choices that can be made for Python modules and for Rust binaries. There is also a section on the end about cross-compiling projects using PyO3.
+
+There is an additional sub-chapters dedicated to [supporting multiple Python versions](./building_and_distribution/multiple_python_versions.html).
```
davidhewitt, line 7:
```suggestion
There is an additional sub-chapter dedicated to [supporting multiple Python versions](./building_and_distribution/multiple_python_versions.html).
```

guide/src/building_and_distribution.md
```diff
@@ -1,58 +1,91 @@
 # Building and Distribution
 
-## Python version
+This chapter of the guide goes into detail on how to build and distribute projects using PyO3. The way to achieve this is very different depending on whether the project is a Python module implemented in Rust, or a Rust binary embedding Python. For both types of project there are also common problems such as the Python version to build for and the [linker](https://en.wikipedia.org/wiki/Linker_(computing)) arguments to use.
+
+The material in this chapter is intended for users who have already read the PyO3 [README](#index.md). It covers in turn the choices that can be made for Python modules and for Rust binaries. There is also a section on the end about cross-compiling projects using PyO3.
```
birkenfeld, line 5: Non-native-speaker question: is there a semantic difference between "at the end" and "on the end"?

guide/src/building_and_distribution.md
```diff
@@ -1,58 +1,91 @@
...
+PyO3 uses a build script (backed by the [`pyo3-build-config`] crate) to determine the Python version and set the correct linker arguments. By default it will attempt to use the following in order:
+ - Any active Python virtualenv.
+ - The `python` executable (if it's a Python 3 interpreter).
+ - The `python3` executable.
+
+You can override the Python interpreter by setting the `PYO3_PYTHON` environment variable, e.g. `PYO3_PYTHON=python3.6`, `PYO3_PYTHON=/usr/bin/python3.9`, or even a PyPy interpreter `PYO3_PYTHON=pypy3`.
+
+Once the Python interpreter is located, `pyo3-build-config` executes it to query the information in the `sysconfig` module which is needed to configure the rest of the compilation.
+
+To validate the configuration which PyO3 will use, you can run the `print-config` binary from the `pyo3-build-config` crate. An example usage of this binary is shown below:
```
birkenfeld, line 20: Could `print-config` be useful as a feature of `pyo3`, so that you can get this output simply by activating it within Cargo.toml, avoiding the need to clone pyo3?

guide/src/building_and_distribution.md
```diff
@@ -1,58 +1,91 @@
...
+base prefix: Some("/usr")
+ld_version: Some("3.8")
+pointer width: Some(8)
+```
 
-PyO3 uses a build script to determine the Python version and set the correct linker arguments. By default it uses the `python3` executable. You can override the Python interpreter by setting `PYO3_PYTHON`, e.g., `PYO3_PYTHON=python3.6`.
+## Building Python extension modules
 
-## Linking
+Python extension modules need to be compiled differently depending on the OS (and architecture) that they are being compiled for. As well as multiple OSes (and architectures). There are also many different Python versions which are actively supported  Packages uploaded to [PyPI](https://pypi.org/) usually want to upload prebuilt "wheels" covering many OS/arch/version combinations so that users on all these different platforms don't have to compile the package themselves. Package vendors can opt-in to the "abi3" limited Python API which allows their wheels to be used on multiple Python versions but restricts the functionality they can use, which reduces the number of wheels they need to compile.
```
birkenfeld, line 43:
```suggestion
Python extension modules need to be compiled differently depending on the OS (and architecture) that they are being compiled for. As well as multiple OSes (and architectures), there are also many different Python versions which are actively supported. Packages uploaded to [PyPI](https://pypi.org/) usually want to upload prebuilt "wheels" covering many OS/arch/version combinations so that users on all these different platforms don't have to compile the package themselves. Package vendors can opt-in to the "abi3" limited Python API which allows their wheels to be used on multiple Python versions, reducing the number of wheels they need to compile, but restricts the functionality they can use.
```

guide/src/building_and_distribution.md
```diff
@@ -1,58 +1,91 @@
...
+Once built, symlink (or copy) and rename the shared library from Cargo's `target/` directory to your desired output directory:
+- on macOS, rename `libyour_module.dylib` to `your_module.so`.
+- on Windows, rename  `libyour_module.dll` to `your_module.pyd`.
+- on Linux, rename `libyour_module.so` to `your_module.so`.
 
-## Testing, building and distribution
+You can then open a Python shell in the output directory and you'll be able to run `import your_module`.
 
-There are two main ways to test, build and distribute your module as a Python package: [setuptools-rust] and [maturin]. setuptools-rust needs several configuration files (`setup.py`, `MANIFEST.in`, `build-wheels.sh`, etc.). It allows (and sometimes requires) writing custom workflows in python. maturin has only few options and works without any additional configuration, instead it requires a rigid project structure and does not support some functionality of setuptools such as package data ([pyo3/maturin#258](https://github.com/PyO3/maturin/issues/258)), multiple extensions or running python scripts at build time.
+See as an example, Bazel rules to build PyO3 on linux at https://github.com/TheButlah/rules_pyo3
```
birkenfeld, line 74:
```suggestion
See, as an example, Bazel rules to build PyO3 on Linux at https://github.com/TheButlah/rules_pyo3.
```

guide/src/building_and_distribution.md
```diff
@@ -1,58 +1,91 @@
...
+### The `extension-module` feature
 
-You can also symlink (or copy) and rename the shared library from the `target` folder:
-- on macOS, rename `libyour_module.dylib` to `your_module.so`.
-- on Windows, rename  `libyour_module.dll` to `your_module.pyd`.
-- on Linux, rename `libyour_module.so` to `your_module.so`.
+PyO3's `extension-module` feature is used to disable [linking](https://en.wikipedia.org/wiki/Linker_(computing)) to `libpython` on unix targets.
 
-You can then open a Python shell in the same folder and you'll be able to run `import your_module`.
+This is necessary because by default PyO3 links to `libpython`. This makes binaries, tests, and examples "just work". However, Python extensions on unix must not link to libpython for [manylinux](https://www.python.org/dev/peps/pep-0513/) compliance).
```
birkenfeld, line 80:
```suggestion
This is necessary because by default PyO3 links to `libpython`. This makes binaries, tests, and examples "just work". However, Python extensions on unix must not link to libpython for [manylinux](https://www.python.org/dev/peps/pep-0513/) compliance.
```

guide/src/building_and_distribution.md
davidhewitt: Hmm interesting idea. I guess there could be a `PYO3_PRINT_CONFIG=1` environment variable which prints the config and then stops the build? I'm not aware of a way to make the output be printed and successfully compile...

guide/src/building_and_distribution.md
davidhewitt: Heh yes, I'd use "at the end" to describe a location of something, e.g. "at the end of the document" or "at the end of the path", versus "on the end" to describe 

This is a good spot and I think a typo by me, because I would actually use "at the end" in this sentence. I think I started writing something else and then didn't correct the start when I changed tack.

```suggestion
The material in this chapter is intended for users who have already read the PyO3 [README](#index.md). It covers in turn the choices that can be made for Python modules and for Rust binaries. There is also a section at the end about cross-compiling projects using PyO3.
```

guide/src/building_and_distribution.md
davidhewitt: 👍 thanks for catching so many typos in this paragraph!

guide/src/building_and_distribution.md
birkenfeld: Right, an env var is probably even easier. And yes, I also think it needs to error out in order to see stdout/stderr.

davidhewitt: Added a commit to remove the `print-config` binary and put it behind an environment variable instead. Feels really nice!

birkenfeld (approved):

PR #1750, opened 2021-08-01
docs: reduce duplication of examples

davidhewitt:
This PR removes the inline examples from `lib.rs` to avoid duplicating them with the README (I noticed they fell out of sync). Instead I put a bunch of links into `lib.rs` to guide users to relevant materials.

I also added a quick `README.md` to the `examples/` directory to make it a nicer landing page as a link from `lib.rs`.

CC @mejrs - you added the really good docs recently to lib.rs, so I wonder what you think of this change?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,7 +179,7 @@ jobs:
       - name: Test example extension modules
         shell: bash
         run: |
-          for example_dir in examples/*; do
+          for example_dir in examples/*/; do
               tox -c $example_dir -e py
           done
         env:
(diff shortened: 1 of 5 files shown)


mejrs:
I don't like this, it's important to have some code snippets in lib.rs. The ones we have now can certainly be improved though.

I think some amount of duplication across the docs and guide is a good thing and not a problem that needs solving.

> Instead I put a bunch of links into lib.rs to guide users to relevant materials.

Not many people a) click through links b) even notice the link, especially if they are visually impaired or are using some kind of reader. I'd rather have a (very) simple example and then "For more, see -link-" below that.

davidhewitt:
Thanks, that's fair points. I'll replace this PR with a different one to resync lib.rs and the README tomorrow 👍

PR #1759, opened 2021-08-05
pyo3-build-config: fix build for windows gnu targets

davidhewitt:
This PR tweaks the logic from #1423 to use mingw lib names only if `sysconfig.get_platform()` returns `"mingw"`.

It's built on top of #1758, because I did a lot of changes in that PR and I didn't want to have to rebase 😅. Only the last commit is relevant for this PR.

Fixes https://github.com/PyO3/setuptools-rust/issues/147

cc @Quogu - I verified this appears to work locally. If you have the chance to check this patch with your use case in the next couple days before I put PyO3 0.14.2 live, it'd be great to have confirmation this resolves your issue.


messense (approved):

davidhewitt:
Cherry-picked this onto #1758 to write more test coverage.

PR #1762, opened 2021-08-07
fix compiler warning

deantvv:
Fix issue [link](https://github.com/rust-lang/rust/issues/79813) which will eventually deny trailing semicolons in expression macro in rust compiler.


davidhewitt:
Thanks 👍

PR #1763, opened 2021-08-07
clean up ffi from pyframe to pyhash

deantvv:
Try to cleanup ffi module. Related to #1289.

### DIFF
--- a/src/ffi/pyframe.rs
+++ b/src/ffi/pyframe.rs
@@ -8,5 +8,4 @@ opaque_struct!(PyFrameObject);
 extern "C" {
     pub fn PyFrame_GetLineNumber(f: *mut PyFrameObject) -> c_int;
 }
-// skipped PyFrame_GetLineNumber
 // skipped PyFrame_GetCode
--- a/src/ffi/pyhash.rs
+++ b/src/ffi/pyhash.rs
@@ -1,6 +1,26 @@
+#[cfg(not(Py_LIMITED_API))]
 use crate::ffi::pyport::{Py_hash_t, Py_ssize_t};
-use std::os::raw::{c_char, c_int, c_void};
+#[cfg(not(Py_LIMITED_API))]
+use std::os::raw::{c_char, c_void};
 
+use std::os::raw::{c_int, c_ulong};
+
+extern "C" {
+    // skipped non-limited _Py_HashDouble
+    // skipped non-limited _Py_HashPointer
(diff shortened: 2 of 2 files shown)


src/ffi/pyframe.rs
```diff
@@ -8,5 +8,4 @@ opaque_struct!(PyFrameObject);
 extern "C" {
     pub fn PyFrame_GetLineNumber(f: *mut PyFrameObject) -> c_int;
 }
-// skipped PyFrame_GetLineNumber
```
davidhewitt, line 11: This symbol is still present in the C header? 

https://github.com/python/cpython/blob/a40675c659cd8c0699f85ee9ac31660f93f8c2f5/Include/pyframe.h#L15

src/ffi/pyhash.rs
```diff
@@ -1,6 +1,18 @@
 use crate::ffi::pyport::{Py_hash_t, Py_ssize_t};
-use std::os::raw::{c_char, c_int, c_void};
+use std::os::raw::{c_char, c_int, c_ulong, c_void};
 
+// skipped non-limited _Py_HashDouble
+// skipped non-limited _Py_HashPointer
+// skipped non-limited _Py_HashPointerRaw
+// skipped non-limited _Py_HashBytes
```
davidhewitt, line 7: Looks like we do define this down below, perhaps move it up here?

davidhewitt (approved):
👍 thanks for doing these. Just two comments, once those are resolved let's merge!

src/ffi/pyframe.rs
deantvv: Maybe I get it wrong, but isn't there a `PyFrame_GetLineNumber` at line 9?

src/ffi/pyhash.rs
deantvv: Yeah, I missed this!

src/ffi/pyframe.rs
davidhewitt: Haha yes sorry that's my bad for reviewing at midnight 😮‍💨

davidhewitt (approved):
👍 as well as relocating `_Py_HashBytes` it looks like `PyHash_GetFuncDef` should also be `#[cfg(not(any(Py_LIMITED_API, PyPy)))]`

With those tweaked and CI green we can merge :)

PR #1771, opened 2021-08-09
Problems with the new Async/Await guide

awestlake87:
Hey, looks like we've got some issues with the async/await guide after that update. I took a look at https://pyo3.rs/v0.14.2/ecosystem/async-await.html this morning to check the links. The link to the testing module appears to be busted and this PR fixes that.

Aside from that, it looks like someone's been trying to integrate it into the rustdoc tests since I see some main functions that shouldn't be visible, and the syntax highlighting is missing on some code sections. Also the README.md still points to 0.14.1 (but that might be by design). Idk if that's a WIP or not, but those changes don't seem to be on the `main` branch, so I wasn't sure where to find them. My README contains most of this guide, and its integrated into rustdoc as well if you want to take a look at how I did it.

My advice for integrating them into the rustdoc tests is to mark the problematic code sections with `rust compile_fail` even though it may seem like a bit of a cop-out. 
- `rust no_run` is also a good idea for tests that require external deps like `uvloop`, but it will still fail to compile some sections without a hidden main (or not-so-hidden in markdown code sections).
- `rust ignore` will still try to build a test for it, it just won't run it by default.

### DIFF
--- a/guide/src/ecosystem/async-await.md
+++ b/guide/src/ecosystem/async-await.md
@@ -498,4 +498,4 @@ fn main() -> PyResult<()> {
 
 ## Additional Information
 - Managing event loop references can be tricky with pyo3-asyncio. See [Event Loop References](https://docs.rs/pyo3-asyncio/#event-loop-references) in the API docs to get a better intuition for how event loop references are managed in this library.
-- Testing pyo3-asyncio libraries and applications requires a custom test harness since Python requires control over the main thread. You can find a testing guide in the [API docs for the `testing` module](https://docs.rs/pyo3-asyncio/testing)
+- Testing pyo3-asyncio libraries and applications requires a custom test harness since Python requires control over the main thread. You can find a testing guide in the [API docs for the `testing` module](https://docs.rs/pyo3-asyncio/latest/pyo3_asyncio/testing)


davidhewitt:
Thanks for checking this! I'm a little confused, are you intending to push more to this PR or should I merge it now?

>  My advice for integrating them into the rustdoc tests is to mark the problematic code sections with rust compile_fail even though it may seem like a bit of a cop-out.

Agreed.

> Also the README.md still points to 0.14.1 (but that might be by design)

Actually the 0.14.2 redirect CI job failed, I'm hoping to look at that later tonight...

awestlake87:
Well what's weird is I see some changes to the guide that just aren't present on the main branch. I didn't see any activity on other pyo3 branches this morning related to the guide either so I'm not sure how those other problems I mentioned happened. I'm willing to fix them, but idk where to look.

awestlake87:
For example on the 0.14.2 guide I see this section:
```rust
#![allow(unused)]
fn main() {
let future = rust_sleep();
}
```

On the main branch I see the same section like it should be:

```rust
let future = rust_sleep();
```

davidhewitt:
It looks like that's to do with mdbook transformations.

If I change it locally to this and build locally 

```
```rust,compile_fail
let future = rust_sleep();
```

Then the transformation doesn't the unwanted extra stuff and the syntax highlighting works again.

TBH, all of these tests would be `compile_fail` anyway because we don't have the `pyo3_asyncio` deps available when the guide builds. So I'm just going to remove the doctest markers completely.

Issue #1776, opened 2021-08-10
Feature request: raw data API for PyString


indygreg:
PyUnicode internally stores its data in various variations. See https://docs.python.org/3/c-api/unicode.html.

PyO3's `PyString` currently only allows you to get at UTF-8 / Rust `str` compatible variations of the data.

rust-cpython - by contrast - exposes a `PyString.data()` returning a `PyStringData` enum:

```rust
pub enum PyStringData<'a> {
    Latin1(&'a [u8]),
    Utf8(&'a [u8]),
    Utf16(&'a [u16]),
    Utf32(&'a [u32]),
}
```

This API enables Rust to have access to the raw bytes backing a Python string, not the UTF-8 normalization of it (if different).

PyOxidizer was relying on this API for testing. (There are some low-level tests around encoding handling that need to verify exact byte sequences and Python string representations are being handled properly.)

While I'm certainly capable of using `unsafe` Python C APIs to get at the raw string data to close this feature gap, I was curious if PyO3 would be interested in a PR to expose a `PyStringData` enumeration for `PyString` instances. Here is my proposal:

1. `PyString` gains a `pub fn data(&self) -> PyStringData<'_>`
2. `PyStringData` is an enum with a variant for each internal Python string variation.
3. `PyString.data()` calls out to `PyUnicode_READY()` + `PyUnicode_{KIND, DATA, GET_LENGTH}` and constructs a `PyStringData` with a slice.

I'd be willing to contribute a PR for this feature if there is interest.

davidhewitt:
👍 I'm ok with adding this, however I think it's going to be quite hard. See my comments on #1777

FWIW given that PEPs 393 & 623 are removing the "legacy" kinds (and if an object is in the legacy state, we can convert it with `PyUnicode_READY`), I'd vote we aim to expose just a three-member enum which matches the remaining kinds exposed:

```
pub enum PyStringData<'a> {
    Ucs1(&'a [u8]),
    Ucs2(&'a [u16]),
    Ucs4(&'a [u32]),
}
```

There's perhaps room for an `Ascii` variant too. (A special-case of `Ucs1` when the `ascii` flag is set.)

PR #1777, opened 2021-08-10
ffi: define some cpython/unicodeobject bindings

indygreg:
pyo3 doesn't currently define various Unicode bindings that allow the
retrieval of raw data from Python strings. Said bindings are a
prerequisite to possibly exposing this data in the Rust API (#1776).
Even if those high-level APIs never materialize, the FFI bindings are
necessary to enable consumers of the raw C API to utilize them.

This commit partially defines the FFI bindings as defined in
CPython's Include/cpython/unicodeobject.h file.

I used the latest CPython 3.9 Git commit for defining the order
of the symbols and the implementation of various inline preprocessor
macros. I tried to be as faithful as possible to the original
implementation, preserving intermediate `#define`s as inline functions.

The structs are a bit wonky and probably warrant the most review
scrutiny. I haven't tested this code thoroughly.

Missing symbols have been annotated with `skipped` and symbols currently
defined in `src/ffi/unicodeobject.rs` have been annotated with `move`.

### DIFF
--- a/src/ffi/cpython/unicodeobject.rs
+++ b/src/ffi/cpython/unicodeobject.rs
@@ -1,4 +1,6 @@
-use crate::ffi::{PyObject, PyUnicode_Check, Py_UCS1, Py_UCS2, Py_UCS4, Py_hash_t, Py_ssize_t};
+use crate::ffi::{
+    PyObject, PyUnicode_Check, Py_UCS1, Py_UCS2, Py_UCS4, Py_UNICODE, Py_hash_t, Py_ssize_t,
+};
 use libc::wchar_t;
 use std::os::raw::{c_char, c_int, c_uint, c_void};
 
@@ -226,22 +228,59 @@ pub unsafe fn PyUnicode_READY(op: *mut PyObject) -> c_int {
 // skipped PyUnicode_WSTR_LENGTH
 
 extern "C" {
-    // move PyUnicode_New
-
+    pub fn PyUnicode_New(size: Py_ssize_t, maxchar: Py_UCS4) -> *mut PyObject;
     pub fn _PyUnicode_Ready(unicode: *mut PyObject) -> c_int;
 
     // skipped _PyUnicode_Copy
-    // move PyUnicode_CopyCharacters
+
+    pub fn PyUnicode_CopyCharacters(
+        to: *mut PyObject,
+        to_start: Py_ssize_t,
+        from: *mut PyObject,
+        from_start: Py_ssize_t,
+        how_many: Py_ssize_t,
+    ) -> Py_ssize_t;
+
     // skipped _PyUnicode_FastCopyCharacters
-    // move PyUnicode_Fill
+
+    pub fn PyUnicode_Fill(
+        unicode: *mut PyObject,
+        start: Py_ssize_t,
+        length: Py_ssize_t,
+        fill_char: Py_UCS4,
+    ) -> Py_ssize_t;
+
     // skipped _PyUnicode_FastFill
-    // move PyUnicode_FromUnicode
-    // move PyUnicode_FromKindAndData
+
+    #[cfg(not(Py_3_12))]
+    #[deprecated]
+    #[cfg_attr(PyPy, link_name = "PyPyUnicode_FromUnicode")]
+    pub fn PyUnicode_FromUnicode(u: *const Py_UNICODE, size: Py_ssize_t) -> *mut PyObject;
+
+    pub fn PyUnicode_FromKindAndData(
+        kind: c_int,
+        buffer: *const c_void,
+        size: Py_ssize_t,
+    ) -> *mut PyObject;
+
     // skipped _PyUnicode_FromASCII
     // skipped _PyUnicode_FindMaxChar
-    // move PyUnicode_AsUnicode
+
+    #[cfg(not(Py_3_12))]
+    #[deprecated]
+    #[cfg_attr(PyPy, link_name = "PyPyUnicode_AsUnicode")]
+    pub fn PyUnicode_AsUnicode(unicode: *mut PyObject) -> *mut Py_UNICODE;
+
     // skipped _PyUnicode_AsUnicode
-    // move PyUnicode_AsUnicodeAndSize
+
+    #[cfg(not(Py_3_12))]
+    #[deprecated]
+    #[cfg_attr(PyPy, link_name = "PyPyUnicode_AsUnicodeAndSize")]
+    pub fn PyUnicode_AsUnicodeAndSize(
+        unicode: *mut PyObject,
+        size: *mut Py_ssize_t,
+    ) -> *mut Py_UNICODE;
+
     // skipped PyUnicode_GetMax
 }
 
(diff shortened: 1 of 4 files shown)


src/ffi/cpython/unicodeobject.rs
```diff
@@ -0,0 +1,314 @@
...
+pub struct PyASCIIObjectState {
+    // SSTATE_* constants.
+    pub interned: c_uint,
+    // PyUnicode_*_KIND constants.
+    pub kind: c_uint,
+    pub compact: c_uint,
+    pub ascii: c_uint,
+    pub ready: c_uint,
+    pub _a: c_uint,
+}
```
davidhewitt, line 42: This struct looks wrong; the upstream CPython source implements this state as a 32-bit [bit field](https://www.tutorialspoint.com/cprogramming/c_bit_fields.htm).

It's not possible to have an equivalent bitfield definition in Rust (e.g. see https://github.com/rust-lang/rfcs/issues/314)

Looks like the `cpython` crate defines the `state` member as a single `u32` field and uses bitwise operators to read values at various offsets. That seems to make sense, but as far as I'm aware the bitfield order isn't actually well-defined (https://stackoverflow.com/a/19376742) so that seems like it's actually a bug. I guess it'll work on some platforms and be UB on others?

I think some more research is needed on the correct way to handle this.

davidhewitt (changes requested):
Thanks for working on this. It's a particularly gnarly one; the representation of unicode objects is one of the more aggressively optimized parts of the Python interpreter. On top of that, any work here needs to be aware of PEP 623 (affecting Python 3.12).

As-is, it looks to me that the `PyASCIIObjectState` struct is incorrect. I'm not entirely sure what the correct solution is. (See the review comment below.)

Because there's some very complex inline macros going on here, I would like to either see tests within this ffi module, or a thoroughly-tested safe API for #1776. I'm particuarly worried there's OS, arch and Python-version edge cases for this low-level code. For example this'll presumably crash on Python 3.12, when that comes along, so we'll need to catch that with tests.

> symbols currently defined in `src/ffi/unicodeobject.rs` have been annotated with `move`.

I understand not wanting to edit the existing code, however I'd much rather that these were moved as part of this PR, perhaps as a standalone commit.

indygreg:
I was just typing up a comment talking about the limitations of this PR and your review beat me to it :)

Yes, the struct definition is wrong. And the macros could use some tests. I'll address your review feedback.

Also, I noticed that PyO3 is using soon-to-be-deprecated Python Unicode APIs. I'm unsure how much you care about that. If I proceed with #1776, I may have a go at flushing things out and improving the string handling story. Time will tell.

davidhewitt:
> Also, I noticed that PyO3 is using soon-to-be-deprecated Python Unicode APIs.

Oh interesting, I'd mistakenly thought #1370 flushed them out! Yes, I'd really appreciate any help moving off deprecated APIs.

indygreg (commented):
I fixed the bitfield definition with this version.

I've added accessor functions for resolving each logical component of the bitfield to a `c_uint`. The macros/functions now use these accessor functions. I figured it was clearer this way rather than doing bitwise manipulation inline in each macro function.

Tests over bitfield APIs and macros have been added. The behavior asserted in the ASCII and UCS-4 tests seem reasonable to me. Although I wouldn't be surprised if the assertions are non-portable across Python implementations.

A 2nd commit has been introduced to move limited API symbols to the appropriate file.

davidhewitt:
👍 thanks, I'll give this a proper review when I get a chance, got a busy few days ahead. Would be great to have it rebased once #1780 is merged so we can see whether the bitfield order is indeed stable on all platforms tested on CI.

src/ffi/cpython/unicodeobject.rs
```diff
@@ -0,0 +1,596 @@
...
+    ) -> Py_ssize_t;
+
+    // skipped _PyUnicode_FastFill
+
+    #[cfg(not(Py_3_12))]
+    #[deprecated]
+    #[cfg_attr(PyPy, link_name = "PyPyUnicode_FromUnicode")]
+    pub fn PyUnicode_FromUnicode(u: *const Py_UNICODE, size: Py_ssize_t) -> *mut PyObject;
+
+    pub fn PyUnicode_FromKindAndData(
```
davidhewitt, line 260:
```suggestion
    #[cfg_attr(PyPy, link_name = "PyPyUnicode_FromKindAndData")]
    pub fn PyUnicode_FromKindAndData(
```

src/ffi/cpython/unicodeobject.rs
```diff
@@ -0,0 +1,596 @@
...
+        _PyUnicode_Ready(op)
+    }
+}
+
+// skipped PyUnicode_MAX_CHAR_VALUE
+// skipped _PyUnicode_get_wstr_length
+// skipped PyUnicode_WSTR_LENGTH
+
+extern "C" {
+    pub fn PyUnicode_New(size: Py_ssize_t, maxchar: Py_UCS4) -> *mut PyObject;
```
davidhewitt, line 231:
```suggestion
    #[cfg_attr(PyPy, link_name = "PyPyUnicode_New)]
    pub fn PyUnicode_New(size: Py_ssize_t, maxchar: Py_UCS4) -> *mut PyObject;
```

src/ffi/cpython/unicodeobject.rs
```diff
@@ -0,0 +1,596 @@
...
+    pub fn _PyUnicode_Ready(unicode: *mut PyObject) -> c_int;
```
davidhewitt, line 232:
```suggestion
    #[cfg_attr(PyPy, link_name = "_PyPyUnicode_Ready")]
    pub fn _PyUnicode_Ready(unicode: *mut PyObject) -> c_int;
```

src/ffi/cpython/unicodeobject.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    // skipped _PyUnicode_Copy
+
+    pub fn PyUnicode_CopyCharacters(
```
davidhewitt, line 236:
```suggestion
    #[cfg(not(PyPy))]
    #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
    pub fn PyUnicode_CopyCharacters(
```

src/ffi/cpython/unicodeobject.rs
```diff
@@ -0,0 +1,596 @@
...
+        to: *mut PyObject,
+        to_start: Py_ssize_t,
+        from: *mut PyObject,
+        from_start: Py_ssize_t,
+        how_many: Py_ssize_t,
+    ) -> Py_ssize_t;
+
+    // skipped _PyUnicode_FastCopyCharacters
+
+    pub fn PyUnicode_Fill(
```
davidhewitt, line 246:
```suggestion
    #[cfg(not(PyPy))]
    #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
    pub fn PyUnicode_Fill(
```

src/ffi/cpython/unicodeobject.rs
```diff
@@ -0,0 +1,596 @@
...
+                assert!(!_PyUnicode_COMPACT_DATA(ptr).is_null());
+                // _PyUnicode_NONCOMPACT_DATA isn't valid for compact strings.
+                assert!(!PyUnicode_DATA(ptr).is_null());
+
+                assert_eq!(PyUnicode_GET_LENGTH(ptr), s.len().unwrap() as _);
+                assert_eq!(PyUnicode_IS_READY(ptr), 1);
+
+                // This has potential to mutate object. But it should be a no-op since
+                // we're already ready.
+                assert_eq!(PyUnicode_READY(ptr), 0);
```
davidhewitt, line 554:
```suggestion
                #[cfg_attr(Py_3_10, allow(deprecated))]
                assert_eq!(PyUnicode_READY(ptr), 0);
```

src/ffi/cpython/unicodeobject.rs
```diff
@@ -0,0 +1,596 @@
...
+                assert!(!_PyUnicode_COMPACT_DATA(ptr).is_null());
+                // _PyUnicode_NONCOMPACT_DATA isn't valid for compact strings.
+                assert!(!PyUnicode_DATA(ptr).is_null());
+
+                assert_eq!(PyUnicode_GET_LENGTH(ptr), py_string.len().unwrap() as _);
+                assert_eq!(PyUnicode_IS_READY(ptr), 1);
+
+                // This has potential to mutate object. But it should be a no-op since
+                // we're already ready.
+                assert_eq!(PyUnicode_READY(ptr), 0);
```
davidhewitt, line 592:
```suggestion
                #[cfg_attr(Py_3_10, allow(deprecated))]
                assert_eq!(PyUnicode_READY(ptr), 0);
```

davidhewitt (approved):
Looks like the CI is passing on all platforms, so I'm reasonably convinced this is a safe way to handle the bitfield :thumbsup:

I've added some tweaks for PyPy and Python 3.10 below, I'll add them as a commit.

davidhewitt:
Looks like I broke this - I'll fixup and merge later, sorry!

indygreg:
> Looks like I broke this - I'll fixup and merge later, sorry!

Twas a missing `"`. I amended your commit and force pushed.

indygreg:
Rust didn't seem to honor `[allow(deprecated)]` on an `assert_eq!(...)` macro. Unsure why. But moving the annotation to the `#[test]` function seems to make Python 3.10 happy. I dunno.

indygreg:
Ugh, something doesn't like `PyUnicode_AsUTF8AndSize`. Something about the annotations on its availability is off.

davidhewitt:
Thanks!

PR #1785, opened 2021-08-11
ci: pin bitflags for MSRV job

davidhewitt:
Looks like bitflags 1.3.0 and greater breaks MSRV in CI.

(bitflags 1.3.0 was just yanked, but at least this change makes it ready for whenever 1.3.1 comes.)


messense (approved):
LGTM

PR #1786, opened 2021-08-12
tests: switch to python_with_gil

mejrs:
This replaces most uses of `acquire_gil` with `with_gil`.

This is only in /src, not in /tests (yet)

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -805,6 +805,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(not(Py_LIMITED_API))]
     fn test_acquire_gil() {
         const GIL_NOT_HELD: c_int = 0;
         const GIL_HELD: c_int = 1;
--- a/src/python.rs
+++ b/src/python.rs
@@ -709,64 +709,62 @@ mod tests {
 
     #[test]
     fn test_eval() {
-        let gil = Python::acquire_gil();
-        let py = gil.python();
+        Python::with_gil(|py| {
+            // Make sure builtin names are accessible
+            let v: i32 = py
+                .eval("min(1, 2)", None, None)
+                .map_err(|e| e.print(py))
+                .unwrap()
+                .extract()
+                .unwrap();
+            assert_eq!(v, 1);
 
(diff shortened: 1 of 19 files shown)


src/python.rs
```diff
@@ -805,4 +803,24 @@ mod tests {
         assert!(PythonVersionInfo::from_str("3.5.2a1+") < (3, 6));
         assert!(PythonVersionInfo::from_str("3.5.2a1+") > (3, 4));
     }
+
+    #[test]
+    fn test_acquire_gil() {
```
davidhewitt, line 808: Looks like `PyGILState_Check` requires the unlimited api. (See #1787)

```suggestion
    #[test]
    #[cfg(not(Py_LIMITED_API))]
    fn test_acquire_gil() {
```

src/types/iterator.rs
```diff
@@ -107,65 +107,62 @@ mod tests {
...
     fn iter_refcnt() {
-        let obj;
-        let count;
-        {
-            let gil_guard = Python::acquire_gil();
-            let py = gil_guard.python();
-            obj = vec![10, 20].to_object(py);
-            count = obj.get_refcnt(py);
-        }
+        let (obj, count) = Python::with_gil(|py| -> (Py<PyAny>, isize) {
```
davidhewitt, line 122: Looks like you'll either need to tweak imports for abi3 or remove the closure return type.

```suggestion
        let (obj, count) = Python::with_gil(|py| {
```

davidhewitt (approved):
This is heroic! Thank you so much for doing this, I'm inspired to open a PR to deprecate `acquire_gil` now 😄 

There's a few CI issues, which I think I've managed to point out below...

src/types/iterator.rs
mejrs: Right that probably should have been a c_int anyway

davidhewitt (approved):
I think that the PyPy failures were independently fixed by #1787 (which I just merged), so I'm going to merge this now and worry about CI if stuff still breaks 😅 

Thanks again for this tidy-up!

PR #1787, opened 2021-08-12
ffi: cleanup pystate

davidhewitt:
- moves `PyGILState_Check` and `Py_tracefunc` to unlimited api
- makes `PyThreadState` an opaque struct (the current definition was totally broken, so haven't bothered mentioning this in the CHANGELOG as nobody can have been using it successfully).

### DIFF
--- a/src/ffi/cpython/pystate.rs
+++ b/src/ffi/cpython/pystate.rs
@@ -1,7 +1,19 @@
-use crate::ffi::pystate::{PyInterpreterState, PyThreadState};
+#[cfg(not(PyPy))]
+use crate::ffi::PyThreadState;
+use crate::ffi::{PyFrameObject, PyInterpreterState, PyObject};
 use std::os::raw::c_int;
 
-// Py_tracefunc is defined in ffi::pystate
+// skipped _PyInterpreterState_RequiresIDRef
+// skipped _PyInterpreterState_RequireIDRef
+
+// skipped _PyInterpreterState_GetMainModule
+
+pub type Py_tracefunc = extern "C" fn(
+    obj: *mut PyObject,
+    frame: *mut PyFrameObject,
+    what: c_int,
+    arg: *mut PyObject,
+) -> c_int;
 
 pub const PyTrace_CALL: c_int = 0;
 pub const PyTrace_EXCEPTION: c_int = 1;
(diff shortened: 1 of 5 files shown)


src/ffi/pystate.rs
```diff
@@ -1,26 +1,46 @@
...
-#[repr(C)]
-#[derive(Copy, Clone)]
-pub struct PyThreadState {
-    pub ob_base: PyObject,
-    pub interp: *mut PyInterpreterState,
-}
-
 extern "C" {
+    #[cfg(not(PyPy))]
+    #[cfg_attr(docsrs, doc(not(PyPy)))]
```
mejrs, line 12:
```suggestion
    #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
```
Same for all the other ones. 

Maybe this is worth putting on top of the `extern "C"` blocks (splitting them if necessary) to reduce littering attributes everywhere?

Note that doc_cfg doesn't pwork with glob imports, so documenting everything in ffi with these is going to be a huge pain.

src/ffi/cpython/pystate.rs
```diff
@@ -12,13 +22,38 @@ pub const PyTrace_C_EXCEPTION: c_int = 5;
 pub const PyTrace_C_RETURN: c_int = 6;
 pub const PyTrace_OPCODE: c_int = 7;
 
+// skipped PyTraceInfo
+// skipped CFrame
+// skipped _PyErr_StackItem
+// skipped _PyStackChunk
+// skipped _ts (aka PyThreadState)
+
 extern "C" {
```
mejrs, line 31: The cfgs in this entire block don't do anything because the entire module import is already gated with `#[cfg(all(Py_3_8, not(PyPy)))]` (https://github.com/PyO3/pyo3/blob/main/src/ffi/cpython/mod.rs#L43).

src/ffi/cpython/pystate.rs
```diff
@@ -1,7 +1,17 @@
...
+use crate::ffi::{PyFrameObject, PyInterpreterState, PyObject, PyThreadState};
 use std::os::raw::c_int;
 
-// Py_tracefunc is defined in ffi::pystate
+// skipped _PyInterpreterState_RequiresIDRef
+// skipped _PyInterpreterState_RequireIDRef
+
+// skipped _PyInterpreterState_GetMainModule
+
+pub type Py_tracefunc = extern "C" fn(
```
mejrs, line 9: These functions were in `Include/pystate` before 3.8. The problem with moving this here now is that the `cpython/pystate` module is gated behind `Py_3_8`, so this removes it from 3.6 and 3.7, which is why CI is now failing. (the same thing may be true for other things)

davidhewitt: Ah. I'd rather have the ffi modules look as much like the Python headers as possible, to minimise confusion. So I'll remove that version gating and tweak things.

src/ffi/cpython/pystate.rs
davidhewitt: As above will remove that gating and make the cfgs here meaningful.

src/ffi/pystate.rs
davidhewitt: Hmm. For now I'm going to keep it like this, but I agree it would be nice to find a long-term solution.

PR #1792, opened 2021-08-14
update README

deantvv:
Key `authors` is necessary for maturin to build


davidhewitt:
See https://github.com/PyO3/pyo3/pull/1743#discussion_r680322789

I think you're using an out of date `maturin`.

I'll close this, please comment if I'm wrong and we can reconsider.

PR #1793, opened 2021-08-14
Build configuration settings to facilitate complete link control

indygreg:
This PR makes a handful of changes to the crate build configuration to support advanced control over linking, which is required by PyOxidizer.

I wrote extensive commit messages detailing each change. Please read the commit messages for more context.

I'm creating this as a draft until I have more confidence that these changes are sufficient to support PyOxidizer on all platforms. I figured I'd publish this as a PR in hopes of getting feedback about the general direction sooner.

The first commit is generally useful and could be cherry picked to `main` if you want.

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -101,6 +101,18 @@ pub struct InterpreterConfig {
     /// Serialized to `build_flags`.
     pub build_flags: BuildFlags,
 
+    /// Whether to suppress emitting of `cargo:rustc-link-*` lines from the build script.
+    ///
+    /// Typically, `pyo3`'s build script will emit `cargo:rustc-link-lib=` and
+    /// `cargo:rustc-link-search=` lines derived from other fields in this struct. In
+    /// advanced building configurations, the default logic to derive these lines may not
+    /// be sufficient. This field can be set to `Some(true)` to suppress the emission
+    /// of these lines.
+    ///
+    /// If suppression is enabled, `extra_build_script_lines` should contain equivalent
+    /// functionality or else a build failure is likely.
+    pub suppress_build_script_link_lines: bool,
+
     /// Additional lines to `println!()` from Cargo build scripts.
     ///
     /// This field can be populated to enable the `pyo3` crate to emit additional lines from its
@@ -244,6 +256,7 @@ print("mingw", get_platform() == "mingw")
             executable: map.get("executable").cloned(),
             pointer_width: Some(calcsize_pointer * 8),
             build_flags: BuildFlags::from_interpreter(interpreter)?.fixup(version, implementation),
+            suppress_build_script_link_lines: false,
             extra_build_script_lines: vec![],
         })
     }
@@ -284,6 +297,7 @@ print("mingw", get_platform() == "mingw")
         let mut executable = None;
         let mut pointer_width = None;
         let mut build_flags = None;
+        let mut suppress_build_script_link_lines = None;
         let mut extra_build_script_lines = vec![];
 
         for (i, line) in lines.enumerate() {
@@ -307,6 +321,9 @@ print("mingw", get_platform() == "mingw")
                 "executable" => parse_value!(executable, value),
                 "pointer_width" => parse_value!(pointer_width, value),
                 "build_flags" => parse_value!(build_flags, value),
+                "suppress_build_script_link_lines" => {
+                    parse_value!(suppress_build_script_link_lines, value)
+                }
                 "extra_build_script_line" => {
                     extra_build_script_lines.push(value.to_string());
                 }
@@ -335,6 +352,7 @@ print("mingw", get_platform() == "mingw")
                 }
                 .fixup(version, implementation)
             }),
+            suppress_build_script_link_lines: suppress_build_script_link_lines.unwrap_or(false),
             extra_build_script_lines,
         })
     }
@@ -374,6 +392,7 @@ print("mingw", get_platform() == "mingw")
         write_option_line!(executable)?;
         write_option_line!(pointer_width)?;
         write_line!(build_flags)?;
+        write_line!(suppress_build_script_link_lines)?;
         for line in &self.extra_build_script_lines {
             writeln!(writer, "extra_build_script_line={}", line)
                 .context("failed to write extra_build_script_line")?;
@@ -950,6 +969,7 @@ fn load_cross_compile_from_sysconfigdata(
         executable: None,
         pointer_width,
         build_flags: BuildFlags::from_config_map(&sysconfig_data).fixup(version, implementation),
+        suppress_build_script_link_lines: false,
         extra_build_script_lines: vec![],
     })
 }
@@ -970,6 +990,7 @@ fn windows_hardcoded_cross_compile(
         executable: None,
         pointer_width: None,
         build_flags: BuildFlags::windows_hardcoded(),
+        suppress_build_script_link_lines: false,
         extra_build_script_lines: vec![],
     })
 }
@@ -1174,6 +1195,7 @@ mod tests {
             lib_dir: Some("lib_dir".into()),
             shared: true,
             version: MINIMUM_SUPPORTED_VERSION,
+            suppress_build_script_link_lines: true,
             extra_build_script_lines: vec!["cargo:test1".to_string(), "cargo:test2".to_string()],
         };
         let mut buf: Vec<u8> = Vec::new();
@@ -1204,6 +1226,7 @@ mod tests {
                 major: 3,
                 minor: 10,
             },
+            suppress_build_script_link_lines: false,
             extra_build_script_lines: vec![],
         };
         let mut buf: Vec<u8> = Vec::new();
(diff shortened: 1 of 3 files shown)


pyo3-build-config/src/impl_.rs
```diff
@@ -101,6 +101,18 @@ pub struct InterpreterConfig {
...
+    ///
+    /// Typically, `pyo3`'s build script will emit `cargo:rustc-link-lib=` and
+    /// `cargo:rustc-link-search=` lines derived from other fields in this struct. In
+    /// advanced building configurations, the default logic to derive these lines may not
+    /// be sufficient. This field can be set to `Some(true)` to suppress the emission
+    /// of these lines.
+    ///
+    /// If suppression is enabled, `extra_build_script_lines` should contain equivalent
+    /// functionality or else a build failure is likely.
+    pub suppress_build_script_link_lines: Option<bool>,
```
davidhewitt, line 114: Does it need to be `Option<bool>`? Seems like it could just default to `false` if not present.

Cargo.toml
```diff
@@ -12,6 +12,7 @@ categories = ["api-bindings", "development-tools::ffi"]
 license = "Apache-2.0"
 exclude = ["/.gitignore", ".cargo/config", "/codecov.yml", "/Makefile", "/pyproject.toml", "/tox.ini"]
 edition = "2018"
+links = "python"
```
davidhewitt, line 15: Yeah I was wondering about this in the past too.

It's also good because it prevents users from accidentally trying to depend on different PyO3 versions, so I think this is 👍.

I guess it is technically breaking, so probably needs to go in 0.15

davidhewitt (commented):
👍 Overall I haven't really read the code super hard yet, I wanted to comment and ask questions about the high-level first. Covering each commit in turn:
1. Yep this is much appreciated, thank you for writing this. It could be worth mentioning the defaults on deserialization (if the corresponding config key is missing)? In theory only `version` is required, depending on use case.
2. I'm suprised that cargo/rustc doesn't have what you need already. Can you give an example of the lines you need to emit? It seems like something like using RUSTFLAGS env var to pass link args might do what you need?
3. The `links = "python"` bit is a good idea, I think it probably does need to happen. A couple of questions about the use case though:
  - There's already `pyo3_build_config::get()` which can be used from build scripts to get the resolved `InterpreterConfig`. If that were stabilised and documented would it meet your needs?
  - If the above is no good, what about passing config file contents through the `DEP_` env var, rather than writing out to a path? Seems like an extra step which may not be needed, however I'm happy to cede on this point.
4. Yep I think having this option makes sense as an advanced feature (it's basically the `extension-module` cargo feature on steroids).

indygreg:
>     2. I'm suprised that cargo/rustc doesn't have what you need already. Can you give an example of the lines you need to emit? It seems like something like using RUSTFLAGS env var to pass link args might do what you need?

`RUSTFLAGS` isn't sufficient because you can't easily target single crates with environment variables.

One use case is we want to tell `pyo3` to link additional static libraries. This is because PyOxidizer can derive a custom `libpython`. That `libpython` is defined as a static library of CPython's object files plus static libraries from other libraries, like `libssl`. It is easier to point Cargo at a bunch of static libraries than to assemble all the requisite object files into a single `libpython` static library/archive. The caveat is it currently fragments the build across multiple crates, as we need to emit `cargo:rustc-link-*` lines outside `pyo3`. But this feature will enable `pyo3` to consume/link a working `libpython` and enable PyOxidizer to simplify build logic.

We also want to make use of `cargo:rustc-link-lib=static-nobundle=pythonXY`. Although come to think of it, if we tell pyo3 about all the static libraries, we may no longer need to use that `static-nobundle` hack!

>     * There's already `pyo3_build_config::get()` which can be used from build scripts to get the resolved `InterpreterConfig`. If that were stabilised and documented would it meet your needs?

Maybe? I'm not sure if that will work correctly since it relies on `env!("OUT_DIR")`. Won't the `pyo3-build-config` crate get built multiple times and have multiple configurations in play? By passing state through `cargo:variable` + environment variables, you ensure state is correct.

>     * If the above is no good, what about passing config file contents through the `DEP_` env var, rather than writing out to a path? Seems like an extra step which may not be needed, however I'm happy to cede on this point.

The commit message explains my reasoning on this topic.

indygreg:
I've got PyOxidizer working with this PR!

As hypothesized by my last comment, I was able to get rid of the `cargo:rustc-link-lib=static-nobundle=pythonXY` Windows hack because the ability to emit extra `cargo:rustc-link-lib` lines via the config file means that the crate linking `libpython` now has total awareness of the Python library and we no longer need to supplement that linking awareness with another crate. I have yet to clean up PyOxidizer's code, but this should simplify things considerably!

davidhewitt:
> I've got PyOxidizer working with this PR!

Awesome! I think that if we need to add the `links = "python"` mechanism then it might take a bit longer for this to be released, because I wasn't planning to do another breaking release (0.15) too soon.

So in that spirit, it might be worth trying out `pyo3_build_config::get()` ? As long as all uses of `pyo3_build_config` are as `[build-dependency]` and/or dependencies of proc macro crates then I understood that it'd only be built once, for the host configuration.

(And I was expecting that you'd have the `PYO3_BUILD_CONFIG` env var present for the whole compilation.)

If `pyo3_build_config::get()` works for your use-case, then we can also avoid a bit of the complexity of adding a file-writing mechanism.

It's worth noting I had a bit of pain with sccache interacting with the config-file-writing - it wasn't reliably making it available to downwind crates. It was probably related to my (ab)use of getting PyO3's build script to write to the `pyo3_build_config` `OUT_DIR`. That was one of the additional motivations of #1758 - I made `pyo3_build_config` inline a lot more of the configuration to avoid bad interactions with sccache.

> The commit message explains my reasoning on this topic.

Looks like you considered having multiple `DEP_` keys? I was thinking of having a single `cargo:PYO3_BUILD_CONFIG=<serialized config file>` key. It might depend whether we can escape newlines and have a single key like this.

indygreg:
I'm starting to split functionality from this into separate PRs to reduce the scope of this one.

I understand why we can't land some of this functionality in 0.14. But I want to make it clear that I'm blocked on adopting PyO3 in PyOxidizer until some aspects of this PR land. I'm on the fence about merging my PyO3 migration branch until there is a published `pyo3` crate with the functionality I need. I've temporarily used Git branches/commits before and gotten burned because they block my ability to release a new version to crates.io. I don't like being blocked by factors outside my control.

With my discovery that `cargo:rustc-link-lib=static-nobundle=pythonXY` is no longer needed, it _might_ be possible for me to reduce scope of this PR somewhat. But I'm pretty sure I need the `links = ` Cargo.toml entry (#1819) in order to enable downstream crates to access exported variables. IMO that PR is backwards incompatible and requires a minor version bump. So PyOxidizer's dependence on 0.15 seems inevitable.

davidhewitt:
👍 I've merged the piece on interpreter config docs and also the new `PyStringData` api just earlier.

> I'm on the fence about merging my PyO3 migration branch until there is a published pyo3 crate with the functionality I need. I've temporarily used Git branches/commits before and gotten burned because they block my ability to release a new version to crates.io. I don't like being blocked by factors outside my control.

I fully agree that having a git dependency blocking release for you is undesirable. Most of what you need is going into 0.14.3, which I'm about to release shortly.

If we can find ways to get the rest of what you need for PyOxidizer in as non-breaking changes I'd be happy to backport / cherry-pick to make a 0.14.4 to get you everything that you need.

I guess we _could_ also release 0.15 before long to make breaking changes if necessary. However I do like to try to include new features (or other improvements) relevant to most users alongside breaking changes so that people have a reason to upgrade / don't get burned out by churn. 😄

davidhewitt:
Based on https://github.com/indygreg/PyOxidizer/pull/433, it sounds like if we change this PR to just contain just the two commits adding extra options to `InterpreterConfig` then we can proceed to merge this and include it in 0.14.5. (We don't need the mechanism to write the file out to the `DEP_`-advertised location any more, nor the links key?)

indygreg:
> Based on [indygreg/PyOxidizer#433](https://github.com/indygreg/PyOxidizer/pull/433), it sounds like if we change this PR to just contain just the two commits adding extra options to `InterpreterConfig` then we can proceed to merge this and include it in 0.14.5. (We don't need the mechanism to write the file out to the `DEP_`-advertised location any more, nor the links key?)

Yeah, assuming `pyo3_build_config::get()`  from my crate's build script won't diverge from what `pyo3`` is actually using using, I just need the new fields in `InterpreterConfig` to provide explicit control over `cargo:` directives. I'll rebase this PR accordingly.

pyo3-build-config/src/impl_.rs
indygreg: Fixed in latest push.

indygreg:
I rebased and reduced scope. I think this is ready for review.

(Although I haven't tested this exact branch against PyOxidizer yet. I may find time to do that in ~8 hours. But no guarantees.)

davidhewitt (approved):
Thanks! This looks good to merge now!

PR #1794, opened 2021-08-15
string: implement API to access raw string data

### LINKED ISSUE #1776: Feature request: raw data API for PyString

indygreg:
With the recent implementation of non-limited unicode APIs, we're
able to query Python's low-level state to access the raw bytes that
Python is using to store string objects.

This commit implements a safe Rust API for obtaining a view into
Python's internals and representing the raw bytes Python is using
to store strings.

Not only do we allow accessing what Python has stored internally,
but we also support coercing this data to a `Cow<str>`.

Closes #1776.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Add `PyList::get_item_unchecked()` and `PyTuple::get_item_unchecked()` to get items without bounds checks. [#1733](https://github.com/PyO3/pyo3/pull/1733)
 - Add `PyAny::py()` as a convenience for `PyNativeType::py()`. [#1751](https://github.com/PyO3/pyo3/pull/1751)
+- Implement `PyString.data()` to access the raw bytes storing a Python string. [#1794](https://github.com/PyO3/pyo3/issues/1794)
 
 ### Changed
 
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -1,13 +1,132 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
+#[cfg(not(Py_LIMITED_API))]
+use crate::exceptions::PyUnicodeDecodeError;
 use crate::types::PyBytes;
 use crate::{
     ffi, AsPyPointer, FromPyObject, IntoPy, PyAny, PyObject, PyResult, PyTryFrom, Python,
     ToPyObject,
 };
 use std::borrow::Cow;
+#[cfg(not(Py_LIMITED_API))]
+use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::str;
+
+/// Represents raw data backing a Python `str`.
+///
+/// Python internally stores strings in various representations. This enumeration
+/// represents those variations.
+#[cfg(not(Py_LIMITED_API))]
+#[cfg_attr(docsrs, doc(cfg(not(Py_LIMITED_API))))]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum PyStringData<'a> {
+    /// UCS1 representation.
+    Ucs1(&'a [u8]),
+
+    /// UCS2 representation.
+    Ucs2(&'a [u16]),
+
+    /// UCS4 representation.
+    Ucs4(&'a [u32]),
+}
+
+#[cfg(not(Py_LIMITED_API))]
+impl<'a> PyStringData<'a> {
+    /// Obtain the raw bytes backing this instance as a [u8] slice.
+    pub fn as_bytes(&self) -> &[u8] {
+        match self {
+            Self::Ucs1(s) => s,
+            Self::Ucs2(s) => unsafe {
+                std::slice::from_raw_parts(
+                    s.as_ptr() as *const u8,
+                    s.len() * self.value_width_bytes(),
+                )
+            },
+            Self::Ucs4(s) => unsafe {
+                std::slice::from_raw_parts(
+                    s.as_ptr() as *const u8,
+                    s.len() * self.value_width_bytes(),
+                )
+            },
+        }
+    }
+
+    /// Size in bytes of each value/item in the underlying slice.
+    #[inline]
+    pub fn value_width_bytes(&self) -> usize {
+        match self {
+            Self::Ucs1(_) => 1,
+            Self::Ucs2(_) => 2,
+            Self::Ucs4(_) => 4,
+        }
+    }
+
+    /// Convert the raw data to a Rust string.
+    ///
+    /// For UCS-1 / UTF-8, returns a borrow into the original slice. For UCS-2 and UCS-4,
+    /// returns an owned string.
+    ///
+    /// Returns [PyUnicodeDecodeError] if the string data isn't valid in its purported
+    /// storage format. This should only occur for strings that were created via Python
+    /// C APIs that skip input validation (like `PyUnicode_FromKindAndData`) and should
+    /// never occur for strings that were created from Python code.
+    pub fn to_string(self, py: Python) -> PyResult<Cow<'a, str>> {
+        match self {
+            Self::Ucs1(data) => match str::from_utf8(data) {
+                Ok(s) => Ok(Cow::Borrowed(s)),
+                Err(e) => Err(crate::PyErr::from_instance(PyUnicodeDecodeError::new_utf8(
+                    py, data, e,
+                )?)),
+            },
+            Self::Ucs2(data) => match String::from_utf16(data) {
+                Ok(s) => Ok(Cow::Owned(s)),
+                Err(e) => {
+                    let mut message = e.to_string().as_bytes().to_vec();
+                    message.push(0);
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Implement `PyString.data()` to access the raw bytes storing a Python string. [#1776](https://github.com/PyO3/pyo3/issues/1776)
```
davidhewitt, line 13: The convention in PyO3 has been to attach the PR number to the CHANGELOG, with the thinking that users can start there and read backwards through the conversation.

```suggestion
- Implement `PyString.data()` to access the raw bytes storing a Python string. [#1794](https://github.com/PyO3/pyo3/issues/1794)
```

davidhewitt (commented):
Thanks, this looks useful to me! I wonder whether eventually we'd use this internally in PyO3 too to potentially avoid causing strings to generate (and cache) their UTF-8 representation.

I have a bunch of small nitty comments below.

I'd also very much like to see some more tests of `PyStringData::to_string()`  - it would be good to cover both the success case (which looks like it's covered) and the error cases (which don't appear to be) for each of the three character widths.

davidhewitt:
Also, `types::string` is currently not a public module, which is why CI is unhappy. I don't personally see the harm in making it public? I think I'd rather users wrote `use pyo3::types::string::PyStringData` over `use pyo3::types:::PyStringData` - the latter seems misleading.

src/types/string.rs
```diff
@@ -89,6 +199,48 @@ impl PyString {
...
+            ffi::PyUnicode_2BYTE_KIND => Ok(PyStringData::Ucs2(unsafe {
+                std::slice::from_raw_parts(raw_data as *const u16, length)
+            })),
+            ffi::PyUnicode_4BYTE_KIND => Ok(PyStringData::Ucs4(unsafe {
+                std::slice::from_raw_parts(raw_data as *const u32, length)
+            })),
+            kind => Err(crate::exceptions::PyValueError::new_err(format!(
+                "unknown string kind: {}",
+                kind
+            ))),
```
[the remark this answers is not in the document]
indygreg: I'm OK with it if you are. I usually try to avoid panics if possible. But as long as we stay on top of API contracts, this should be fine.

indygreg:
OK. I think the most recent push addresses most (all?) of your feedback.

I did disagree with you on exposing `PyStringData` under `pyo3::types::string`. If you insist on that, suggest a commit and I'll apply it.

I added tests for verifying `.to_string()` errors. This required using `PyUnicode_FromKindAndData()` to construct instances with invalid byte sequences, which Python appears to accept without validation.

As part of writing the tests, I noticed there may be a `c_int` / `c_uint` mismatch in the ffi definitions. When I implemented the FFI bindings, I attempted to copy signedness from the C headers. However, some of these things were macros and lacked type annotations. I may have used the wrong types for symbols like `PyUnicode_1BYTE_KIND` :/

src/types/string.rs
```diff
@@ -1,14 +1,129 @@
...
+                    Err(crate::PyErr::from_instance(PyUnicodeDecodeError::new(
+                        py,
+                        CStr::from_bytes_with_nul(b"utf-16\0").unwrap(),
+                        self.as_bytes(),
+                        0..self.as_bytes().len(),
+                        CStr::from_bytes_with_nul(&message).unwrap(),
+                    )?))
+                }
+            },
+            Self::Ucs4(data) => match data.iter().map(|&c| char::from_u32(c)).collect() {
```
indygreg, line 95: `std::char::from_u32` and the corresponding `TryFrom<u32>` conversion appear to be added in Rust 1.34. The implementation literally just checks for `if (i > MAX as u32) || (i >= 0xD800 && i <= 0xDFFF)` so it would be trivial to inline. I suppose we'll need to do that to support Rust 1.32.

Unless you have any better suggestions...

birkenfeld: PyO3 MSRV is 1.41 though? What am I missing?

indygreg: Err. You are right. CI was complaining when I used `char::from_u32`.

It appears `char::from_u32` and `std::char::from_u32` are different. The former works on stable but not on 1.41. TIL. I'll push a fix.

src/types/string.rs
```diff
@@ -89,6 +199,48 @@ impl PyString {
             }
         }
     }
+
+    /// Obtains the raw data backing the Python string.
+    ///
+    /// If the Python string object was created through legacy APIs, its internal
+    /// storage format will be canonicalized before data is returned.
+    #[cfg(not(Py_LIMITED_API))]
```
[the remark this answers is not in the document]
birkenfeld: is this available on PyPy?

src/types/string.rs
```diff
@@ -1,14 +1,129 @@
...
+        }
+    }
+
+    /// Convert the raw data to a Rust string.
+    ///
+    /// For UCS-1 / UTF-8, returns a borrow into the original slice. For UCS-2 and UCS-4,
+    /// returns an owned string.
+    ///
+    /// Returns [PyUnicodeDecodeError] if the string data isn't valid in its purported.
+    #[allow(clippy::wrong_self_convention)]
```
birkenfeld, line 71: Since this struct is `Copy`, why can't you take `self` instead of silencing the lint?

src/types/string.rs
```diff
@@ -1,14 +1,129 @@
...
+            Self::Ucs4(s) => unsafe {
+                std::slice::from_raw_parts(
+                    s.as_ptr() as *const u8,
+                    s.len() * self.character_width_bytes(),
+                )
+            },
+        }
+    }
+
+    /// Size in bytes of each *character* in the underlying buffer.
```
birkenfeld, line 55: Is "character" the correct terminology here? IMO it should be "codepoint".

src/types/string.rs
```diff
@@ -1,14 +1,129 @@
...
+            Self::Ucs4(_) => 4,
+        }
+    }
+
+    /// Convert the raw data to a Rust string.
+    ///
+    /// For UCS-1 / UTF-8, returns a borrow into the original slice. For UCS-2 and UCS-4,
+    /// returns an owned string.
+    ///
+    /// Returns [PyUnicodeDecodeError] if the string data isn't valid in its purported.
```
birkenfeld, line 70: IIUC exceptions can only happen if strings with invalid encoding are created using the C API, right? Would be nice to clarify that (also in to_string_lossy below).

src/types/string.rs
indygreg: Good suggestion!

src/types/string.rs
indygreg: It is debatable.

It is unclear to me if Python's internal storage format actually guarantees that a slice N byte value is the same as a _code point_. UTF-8 and UTF-16 are variable width encodings after all. I _think_ Python will do things like store surrogate pairs without expanding the item width of the stored value.

The documentation at https://docs.python.org/3/c-api/unicode.html uses _character_ to refer to the values in these buffers, which is the main reason I used _character_ in the API.

But, I don't likely blindly copying things, especially if they are wrong.

What I know for certain is that the values stored in the memory slice are stored with a fixed width. So maybe we should go with a name that says what they actually are: `item width in bytes`.

🤷‍♂️

src/types/string.rs
indygreg: I'll update the docs in the next push.

src/types/string.rs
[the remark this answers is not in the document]
davidhewitt: Looks like the PyPy CI is working fine!

davidhewitt (approved):
Nice work, this looks good to me!

PR #1795, opened 2021-08-15
ci: update cargo-llvm-cov

davidhewitt:
Updates to `cargo-llvm-cov` 0.1.0 (no longer a prerelease).

This version restores compatibility with trybuild (using new trybuild 1.0.45), and should also enable testing the `multiple-pymethods` feature again.

If everything works, I guess coverage numbers will go up 🤞

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -213,6 +213,8 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/cache@v2
         with:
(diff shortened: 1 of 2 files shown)


davidhewitt:
Working brilliantly! 🎉 

Thank you @taiki-e! With `cargo-llvm-cov` we've managed to detect coverage for about 5% of the PyO3 codebase which our previous CI setup was not measuring. 🔎

messense (approved):

PR #1796, opened 2021-08-15
release: 0.14.3

davidhewitt:
Since there have been a few FFI fixes, in particular #1791 which causes a nasty crash, I felt it would be worth cherry-picking these on top of 0.14.2 and creating a 0.14.3 release.

As these commits all exist in `main` I don't think it's worth actually merging this branch, I'll create a separate PR to update the CHANGELOG to separate out the contents released in 0.14.3 when this goes live. Perhaps there's a lesson for the future to consider having an 0.14.x branch, 0.15.x branch etc., although that in itself also creates extra work.

I propose I'll put this live next weekend, I was thinking it could be nice to also cherry-pick #1794 (and maybe non-breaking parts of #1793) into this release, so that PyOxidizer is able to move ahead with its PyO3 port without waiting for us to release 0.15 (which might be a couple months off).

cc @indygreg

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,7 @@ jobs:
         run: |
           cargo update -p indexmap --precise 1.6.2
           cargo update -p hashbrown:0.11.2 --precise 0.9.1
+          cargo update -p bitflags --precise 1.2.1
 
       - name: Build docs
         run: cargo doc --no-deps --no-default-features --features "${{ steps.settings.outputs.all_additive_features }}"
--- a/.github/workflows/guide.yml
+++ b/.github/workflows/guide.yml
@@ -50,7 +50,7 @@ jobs:
 
       - name: Deploy
         if: ${{ github.ref == 'refs/heads/main' || github.event_name == 'release' }}
-        uses: peaceiris/actions-gh-pages@v3.7.0-8
+        uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./gh-pages-build/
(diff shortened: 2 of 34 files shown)


Eric-Arellano:
It's worked well for us with Pants to create a release branch as soon as rc0 goes out for a particular release. It allows us to immediately land breaking changes to `main` without a cooling off period, and keeps us disciplined with cherry picking bug fixes / non-breaking improvements.

davidhewitt:
Thanks, that's definitely interesting food for thought. 

I also realised this morning that in our CHANGELOG we use git ranges to produce comparison links between each PyO3 version. It might get more fiddly to create those if we have a slightly nonlinear history. But maybe that's fine.

davidhewitt:
CI seems to be borked so I'm going to close and open the PR... ?

davidhewitt:
Hmm still no CI. Very confusing.

messense:
> Hmm still no CI. Very confusing.

Because of the conflicting file? See https://github.community/t/run-actions-on-pull-requests-with-merge-conflicts/17104

davidhewitt:
Ah interesting, somehow I'd never been aware of that.

I think I'll have to make a 0.14 release branch in that case, and change this PR to merge into that.

davidhewitt:
Aha looks like that's working now.

tiran:
Could you please cherry-pick https://github.com/PyO3/pyo3/pull/1812 into 0.14.3, too? Thanks!

mejrs:
And also both of the decref ones? https://github.com/PyO3/pyo3/pull/1810 and https://github.com/PyO3/pyo3/pull/1806

davidhewitt:
Thanks, cherry-picked the new fixes. 

I'll put this live tonight unless I hear a reason to wait!

davidhewitt:
Release is live!

PR #1799, opened 2021-08-16
Expand supported num-complex versions

### LINKED ISSUE #1798: pyo3, rust-numpy, ndarray, and num-complex MSRV issue
## 🐛 Bug Reports

There is an MSRV conflict between ndarray, rust-numpy, and pyo3. Specifically the msrv of ndarray was increased to 1.42 in ndarray 0.14.0 and to rust 1.49 in 0.15 (see https://github.com/rust-ndarray/ndarray/blob/master/RELEASES.md ) so to use ndarray with numpy we need ndarray 0.13.x which still supports building with rust 1.41. However, if we need to use num-complex with all of the above for a complex number we have to use the version pinned in the ndarray 0.13.1 package which is 0.2:

https://github.com/rust-ndarray/ndarray/blob/0.13.1/Cargo.toml#L33

but this has an interop issue with pyo3's num-complex optional dependency which requires 0.4:

https://github.com/PyO3/pyo3/blob/main/Cargo.toml#L24

which causes a version mismatch if there is an object taken from a numpy array converted to a readonly ndarray that gets converted to a pyobject (which I hit).

### 🌍 Environment

 - Your operating system and version: linux
 - Your python version: 3.9
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: arch linux package
 - Your Rust version (`rustc --version`): 1.41.1
 - Your PyO3 version: 0.14.2
 - Have you tried using latest PyO3 main (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: no (but it has the same issue)

### 💥 Reproducing

```toml
[dependencies]
numpy = "0.14.1"
num-complex = "0.2"
ndarray = "0.13.1"

[dependencies.pyo3]
version = "0.14.2"
features = ["extension-module", "num-complex"]
```

```rust
use ndarray::prelude::*;
use pyo3::prelude::*;
use pyo3::Python;
use numpy::PyReadonlyArray2;
use num_complex::Complex64;

#[pyfunction]
fn reproduce<'p>(py: Python<'p>, matrix: PyReadonlyArray2<'p, Complex64>) -> Complex64 {
    let array = matrix.as_array();
    array[[0, 0]]
}
```

The only way I can get this to work with all 4 libs is by using the latest rust version and the latest ndarray and num-complex. Then everything builds as expected and is using num-complex 0.4. I really can't wait for cargo to support msrv package metadata so this kind of situation can be automatically resolved.

mtreinish:
When building an extension with rust-numpy and ndarray on the MSRV of
1.41 with complex numbers. The num-complex crate version needs to be
0.2 which was the pinned version as of ndarray 0.13.1 which was the last
release of ndarray that supported building with rust 1.41. However, the
pyo3 pinned version of 0.4 is incompatible with this and will cause an
error when building because of the version mismatch. To fix this This
commit expands the supported versions for num-complex to match what
rust-numpy uses [1] so that we can build pyo3, numpy, ndarray, and
num-complex in an extension with rust 1.41.

Fixes #1798

[1] https://github.com/PyO3/rust-numpy/blob/v0.14.1/Cargo.toml#L19

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ inventory = { version = "0.1.4", optional = true }
 libc = "0.2.62"
 parking_lot = "0.11.0"
 num-bigint = { version = "0.4", optional = true }
-num-complex = { version = ">= 0.2, <= 0.4", optional = true }
+num-complex = { version = ">= 0.2, < 0.5", optional = true }
 # must stay at 0.1.x for Rust 1.41 compatibility
 paste = { version = "0.1.18", optional = true }
 pyo3-macros = { path = "pyo3-macros", version = "=0.14.2", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ inventory = { version = "0.1.4", optional = true }
 libc = "0.2.62"
 parking_lot = "0.11.0"
 num-bigint = { version = "0.4", optional = true }
-num-complex = { version = "0.4", optional = true }
+num-complex = { version = ">= 0.2, <= 0.4", optional = true }
 # must stay at 0.1.x for Rust 1.41 compatibility
 paste = { version = "0.1.18", optional = true }
 pyo3-macros = { path = "pyo3-macros", version = "=0.14.2", optional = true }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ inventory = { version = "0.1.4", optional = true }
 libc = "0.2.62"
 parking_lot = "0.11.0"
 num-bigint = { version = "0.4", optional = true }
-num-complex = { version = "0.4", optional = true }
+num-complex = { version = ">= 0.2, <= 0.4", optional = true }
```
davidhewitt, line 24: as `"0.4"` alone really means `">= 0.4.0, < 0.5.0", I think this needs to be:

```suggestion
num-complex = { version = ">= 0.2, < 0.5", optional = true }
```

davidhewitt (approved):
👍 I agree this is a good idea, if this gets merged this week I can also cherry-pick it onto 0.14.3 this weekend.

davidhewitt (approved):

PR #1802, opened 2021-08-17
Some PyList index fixes

birkenfeld:
NB: the behavior on out-of-range indices hasn't changed;
it was merely wrongly documented before.

If we want to panic instead, to match e.g. Vec::insert, this needs to be implemented specially.

### DIFF
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -4,6 +4,7 @@
 
 use crate::err::{self, PyResult};
 use crate::ffi::{self, Py_ssize_t};
+use crate::internal_tricks::get_ssize_index;
 use crate::{
     AsPyPointer, IntoPy, IntoPyPointer, PyAny, PyNativeType, PyObject, Python, ToBorrowedObject,
     ToPyObject,
@@ -83,17 +84,35 @@ impl PyList {
         }
     }
 
+    /// Takes the slice `self[low:high]` and returns it as a new list.
+    ///
+    /// Indices must be nonnegative, and out-of-range indices are clipped to
+    /// `self.len()`.
+    pub fn slice(&self, low: usize, high: usize) -> &PyList {
+        unsafe {
+            self.py().from_owned_ptr(ffi::PyList_GetSlice(
+                self.as_ptr(),
+                get_ssize_index(low),
+                get_ssize_index(high),
+            ))
+        }
+    }
+
     /// Sets the item at the specified index.
     ///
-    /// Panics if the index is out of range.
-    pub fn set_item<I>(&self, index: isize, item: I) -> PyResult<()>
+    /// Raises `IndexError` if the index is out of range.
+    pub fn set_item<I>(&self, index: usize, item: I) -> PyResult<()>
     where
         I: ToPyObject,
     {
         unsafe {
             err::error_on_minusone(
                 self.py(),
-                ffi::PyList_SetItem(self.as_ptr(), index, item.to_object(self.py()).into_ptr()),
+                ffi::PyList_SetItem(
+                    self.as_ptr(),
+                    get_ssize_index(index),
+                    item.to_object(self.py()).into_ptr(),
+                ),
             )
         }
     }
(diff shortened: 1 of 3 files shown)


src/types/list.rs
```diff
@@ -83,17 +83,34 @@ impl PyList {
         }
     }
 
+    /// Takes the slice `self[low:high]` and returns it as a new list.
```
davidhewitt, line 86: Given that you're using the Python slice syntax here it may be worth calling out in docs that this doesn't take negative (from-the-end) indices.

(Maybe that's also worth commenting on at the top of this file on the `PyList` type?)

src/types/list.rs
```diff
@@ -83,17 +83,34 @@ impl PyList {
...
+    ///
+    /// Out-of-range indices are clipped to `self.len()`.
+    pub fn slice(&self, low: usize, high: usize) -> &PyList {
+        unsafe {
+            self.py().from_owned_ptr(ffi::PyList_GetSlice(
+                self.as_ptr(),
+                low as Py_ssize_t,
+                high as Py_ssize_t,
```
davidhewitt, line 94: Because this is an "as" cast, there's actually something weird here where users can (ab)use twos complement to smuggle negative indices into this call: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ec2001b00be41c47404ab07d3826633c

```
[src/main.rs:2] std::usize::MAX as isize = -1
```

We might need to consider overflow behaviour like this for all of the functions in this file?

For `get_slice`, it given that out-of-range indices are already clipped at the top end, we may wish to just clamp values to std::isize::MAX before converting? E.g.

```suggestion
                std::cmp::min(low, std::isize::MAX as usize) as Py_ssize_t,
                std::cmp::min(high, std::isize::MAX as usize) as Py_ssize_t,
```

davidhewitt (changes requested):
Hmm, staring at this, I now think that we need to be careful when casting the `usize` function arguments to `Py_ssize_t`...

birkenfeld:
Yeah that's a good point. I think the simplest way would be clamping to `Py_ssize_t::MAX` in all cases; that will let Python raise IndexError or clamp as desired.  A possible false positive case, i.e. a Python collection with `Py_ssize_t::MAX` items, is not realistic in any case.

src/types/list.rs
birkenfeld: Can do, although negative indices are out of the question due to the argument type already :)

src/types/tuple.rs
```diff
@@ -57,18 +58,28 @@ impl PyTuple {
...
+                self.as_ptr(),
+                get_ssize_index(low),
+                get_ssize_index(high),
+            ))
         }
     }
 
     /// Takes a slice of the tuple from `low` to the end and returns it as a new tuple.
-    pub fn split_from(&self, low: isize) -> &PyTuple {
+    pub fn split_from(&self, low: usize) -> &PyTuple {
```
davidhewitt, line 76: Looks like this function lacks a test, so it would be great to add one if you're willing!

davidhewitt (approved):
👍 needs a CHANGELOG entry, I think, otherwise LGTM!

It might be worth adding tests with out-of-range usize values, now that we've realised that edge case...

src/types/tuple.rs
birkenfeld: Actually I was wondering if it's needed. The operation exists neither in Python nor in Rust, is not used internally, and it's easy enough to use `slice` instead.

birkenfeld:
I was thinking that this should have a common changelog entry with #1733 and the other followups.

BTW, is `internal_tricks` the right module for the helper function? Happy to move it somewhere else.

src/types/tuple.rs
davidhewitt: Fair point. Yes let's deprecate it and plan to remove.

As you say, it's easy to reproduce with either `.slice()` or `.as_slice().split_at()`

davidhewitt:
> BTW, is internal_tricks the right module for the helper function? Happy to move it somewhere else.

Not really sure where it fits either. I think it's fine there for now. As it's `pub(crate)` only, quite easy for us to relocate later.

davidhewitt:
It looks like this is unlikely to cause much pain when cherry-picking other things from `main` for 0.14.3, so I'll proceed to merge this now.

PR #1803, opened 2021-08-17
PySequence: use usize everywhere, fix in-place methods

birkenfeld:


davidhewitt:
(I guess this needs to wait for #1802)

davidhewitt:
👍 this is looking good!

It looks like there's a few `PySequence` apis which are lacking tests; would you be willing to add them as part of this PR?

birkenfeld:
Sure.

davidhewitt (approved):
Looks good to me, thanks for the new tests!

I guess still to be added for #1667 is to add the `impl std::ops::Index` implementations, and maybe also an entry in the migration guide.

PR #1805, opened 2021-08-17
Test proc macro hygiene for `#[pyclass]` macro.

mejrs:
This forces unambiguous types in proc macro output, preventing things like https://github.com/PyO3/pyo3/pull/1711

For example, using ` quote! { pyo3::PyAny }` will now fail, and require an explicit use: ` quote! { ::pyo3::PyAny }`.

For example, the error for the above looks like this:

```
error[E0223]: ambiguous associated type
  --> tests\test_proc_macro_hygiene.rs:25:1
   |
25 | #[::pyo3::proc_macro::pyclass]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use fully-qualified syntax: `<Shadows_pyo3<()> as Trait>::PyAny`
   |
   = note: this error originates in the attribute macro `::pyo3::proc_macro::pyclass` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### DIFF
--- a/tests/test_proc_macro_hygiene.rs
+++ b/tests/test_proc_macro_hygiene.rs
@@ -1,22 +1,26 @@
 #![no_implicit_prelude]
-#![allow(non_camel_case_types)]
-#![allow(dead_code)]
 
-trait Use_unambiguous_imports<T> {
-    type Error;
+macro_rules! shadow {
+    ($name: ident) => {
+        ::paste::item! {
+            #[allow(non_camel_case_types, dead_code)]
+            unsafe trait [<NobodyImplsThis_ $name>]  {}
+
+            #[allow(non_camel_case_types, dead_code)]
+            struct [<Shadows_ $name>]<T: [<NobodyImplsThis_ $name>]> {
+                _ty: ::core::marker::PhantomData<T>,
+              }
+
+            #[allow(non_camel_case_types, dead_code)]
+            type $name = [<Shadows_ $name>]<()>;
+        }
+    };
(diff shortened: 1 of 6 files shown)


tests/test_proc_macro_hygiene.rs
```diff
@@ -0,0 +1,66 @@
...
+            #[allow(non_camel_case_types, dead_code)]
+            type $name = [<Shadows_ $name>]<()>;
+        }
+    };
+}
+
+shadow!(std);
+shadow!(alloc);
+shadow!(core);
+shadow!(pyo3);
```
birkenfeld, line 23: All this shadowing is unneeded - with the `no_implicit_prelude` all the crates aren't available anyway and you'll get `use of undeclared crate or module`.

birkenfeld (changes requested):

tests/test_proc_macro_hygiene.rs
davidhewitt: (if that is indeed the case, that would mean we also wouldn't be dependent on the `paste!` macro in this test, which would be nice)

mejrs: Thanks, I didn't realize that. I've removed it. 👍

davidhewitt:
👍 this is looking great, though it looks like it might take a few more iterations to satisfy CI! 🙃

davidhewitt (approved):
This is great, thank you!

PR #1806, opened 2021-08-17
Fix memory leak when calling Python from Rust

### LINKED ISSUE #1801: Potential memory leak when calling Python from Rust
## 🐛 Bug Reports

Repeatedly calling Python from Rust appears to cause a memory leak of ~700 bytes per invocation. This poses a problem for long-lived Rust programs that repeatedly call Python. In my case, I ran into this issue while adding support for Python UDFs to a Rust-based streaming system.

As a side note, thanks for all the hard work on this fantastic library! Hoping this can be resolved, and that I haven't missed something on my end.

### 🌍 Environment

 - Your operating system and version: Ubuntu 20.04 LTS
 - Your python version: 3.8.10
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: Installed via apt, using a virtualenv.
 - Your Rust version (`rustc --version`): 1.54.0
 - Your PyO3 version: 0.14.2

### 💥 Reproducing

```rust
use pyo3::prelude::*;

fn main() {
    loop {
        Python::with_gil(move |py| {
            py.run("1 + 1", None, None).unwrap();
        }); 
    }
}
```

The following also causes the issue:
```rust
use pyo3::prelude::*;

fn main() {
    Python::with_gil(move |py| {
        loop {
            py.run("1 + 1", None, None).unwrap();
        }
    }); 
}
```

Run with `cargo run --release` and observe memory use grow.

pschafhalter:
`Python::run_code` transforms the Rust code string to an owned pointer. The reference count to the owned pointer isn't decremented, which results in a memory leak when calling Python from Rust.

This PR adds a missing `Py_DECREF` call on the owned pointer which fixes the memory leak.

Fixes #1801.

Thanks to @mejrs for the tip on the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Loosened the lower bound on the `num-complex` optional dependency to support
   interop with `rust-numpy` and `ndarray` when building with the MSRV of 1.41
   [#1799](https://github.com/PyO3/pyo3/pull/1799)
+- Add missing `Py_DECREF` to fix memory leak when calling Python from Rust. [#1806](https://github.com/PyO3/pyo3/pull/1806)
 
 ## [0.14.2] - 2021-08-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Loosened the lower bound on the `num-complex` optional dependency to support
   interop with `rust-numpy` and `ndarray` when building with the MSRV of 1.41
   [#1799](https://github.com/PyO3/pyo3/pull/1799)
-- Add missing `Py_DECREF` to fix memory leak when calling Python from Rust. [#1806](https://github.com/PyO3/pyo3/pull/1806)
+- Add missing `Py_DECREF` to `Python::run_code` which fixes a memory leak when
+  calling Python from Rust. [#1806](https://github.com/PyO3/pyo3/pull/1806)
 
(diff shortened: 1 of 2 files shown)


mejrs (approved):
Thanks!

> The owned pointer isn't dereferenced

That's not what `Py_DECREF` does (see https://docs.python.org/3/c-api/refcounting.html for more on refcounting).

What happened here was that the raw pointer never had its reference count decremented, so the inner value was leaked and never deallocated. It's similar to `mem::forget`'ing a `Rc` in Rust.

pschafhalter:
Thanks for the correction! Updated the comment on the PR

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Loosened the lower bound on the `num-complex` optional dependency to support
   interop with `rust-numpy` and `ndarray` when building with the MSRV of 1.41
   [#1799](https://github.com/PyO3/pyo3/pull/1799)
+- Add missing `Py_DECREF` to fix memory leak when calling Python from Rust. [#1806](https://github.com/PyO3/pyo3/pull/1806)
```
birkenfeld, line 31: This should explicitly state which function contained the bug.

birkenfeld (changes requested):

birkenfeld (approved):
LGTM, thanks! @davidhewitt needs cherry-pick to 0.14...

messense (approved):
Thanks!

messense:
[auto-merge](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request) works! 🎉

PR #1816, opened 2021-08-19
deduplicate pyo3-macros documentation

mejrs:
This adds a feature flag to the `pyo3-macros` crate, so that it can be invoked from `pyo3` itself and we can use that feature flag there. This allows unifying the macros there rather than in pyo3/lib.rs itself.

I'm not sure if there are any gotchas I didn't see. Is anyone using `pyo3-macros` directly? Does this count as a breaking change for anyone that does?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ default = ["macros"]
 macros = ["pyo3-macros", "indoc", "paste", "unindent"]
 
 # Enables multiple #[pymethods] per #[pyclass]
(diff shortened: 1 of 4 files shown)


davidhewitt (approved):
This is definitely a good idea as far as I can tell (I should have done it when originally implementing `multiple-pymethods`).

> Is anyone using `pyo3_macros` directly?

I wouldn't have thought so, given we re-export it in the main crate. And I think this is fine to be breaking anyway - presumably it'll become part of 0.15.

Issue #1817, opened 2021-08-21
Remove self-dev-dependency from Cargo.toml


davidhewitt:
From the discussion in #1811, it sounds like the use of `path = .` dev-dependency is ill-advised.

https://github.com/PyO3/pyo3/blob/79c7e2879979cc8f1d44c542f8bb95a789df13db/Cargo.toml#L41

@decathorpe - the use of this arose because `cargo test` without the features enabled by the self-dev-dependency is more-or-less pointless - we need the `auto-initialize` feature enabled else every test will fail when trying to get the Python interpreter. Perhaps I can come up with a better solution for that.

FWIW this was recommended in https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481

davidhewitt:
I think a `#[pyo3::test]` proc-macro which automatically called `prepare_freethreaded_python()` and `Python::with_gil` could remove a lot of boilerplate as well as the need for `auto-initialize` from most tests.

doctests / example would still need an alternative solution. Could just put `pyo3::prepare_freethreaded_python()` as a hidden statement in the top of all of them for now.

davidhewitt:
Hmm, this actually makes the problem worse - it means that the tests even stop building without the `macros` feature enabled. Arguably this is even worse. 

I'm not sure we can resolve easily without upstream support in Cargo?

decathorpe:
I think the "easiest" solution would be to either move those tests to `tests/foo.rs` and specify their `required-features` in `Cargo.toml` (if that's possible, e.g. the tests are only using public API), or to wrap unit tests that use non-default features in something like `#[cfg(all(feature = "macros", feature = "auto-initialize"))]` attributes. This way, they won't be compiled if those features are not enabled, but you can still compile and run them with `cargo test --all-features` etc.

davidhewitt:
I agree that's the most theoretically correct path, however it also seems impractical for PyO3's general maintenance.

The `auto-initialize` feature in particular seems unavoidable to test without.

Pretty much all of the unit tests and doctests in PyO3 need to interact with the Python interpreter, so unless every test added boilerplate to initialize the GIL (i.e. a call to `pyo3::prepare_freethreaded_python`) then pretty much all of these tests and doctests would be decorcated in `cfg` attributes.

[...]

PR #1822, opened 2021-08-23
Add `anyhow-integration` feature which implements From<anyhow::Error> for PyErr

chris-laplante:
This makes it possible to use anyhow with pyo3. Includes documentation for feature in guide.


Cargo.toml
```diff
@@ -74,6 +75,10 @@ auto-initialize = []
 # Optimizes PyObject to Vec conversion and so on.
 nightly = []
 
+# Provides a (lossy) From<anyhow::Error> impl for PyErr.
+# There is currently no way to convert the PyErr back into the original anyhow::Error.
+anyhow-integration = ["anyhow"]
```
davidhewitt, line 80: There's no need for this to have a separate feature - users can just enable `anyhow` feature directly.

guide/src/features.md
```diff
@@ -36,6 +36,10 @@ This feature changes [`Python::with_gil`]({{#PYO3_DOCS_URL}}/pyo3/struct.Python.
 
 If you do not enable this feature, you should call `pyo3::prepare_freethreaded_python()` before attempting to call any other Python APIs.
 
+### `anyhow-integration`
+
+This feature makes it possible to return [`anyhow::Result<T>`](https://docs.rs/anyhow/1.0.43/anyhow/type.Result.html) from functions and methods exposed to Python. It does so by adding `impl From<anyhow::Error> for PyErr`. Currently, the conversion simply stringifies the `anyhow::Error` and shoves it into a `PyRuntimeError`. As a consequence, there is no way to convert a `PyErr` back to the original `anyhow::Error`. It is mostly intended as a developer convenience.
```
davidhewitt, line 41: We should probably make the section with `num_bigint` etc into an `Optional Dependencies` section and add `anyhow` there.

davidhewitt (commented):
Hmmm.

I agree `anyhow` is a very useful library, and I understand why this is desirable.

I don't much like how this makes `anyhow::Error` always map to `RuntimeError`, which misses out on the richness of Python's range of exception types.

I'd be interested in hearing other maintainer's opinions on whether they think this is a useful feature to add to the library. I could go either way, however my stance at the moment is that I'd like to be persuaded this is necessary.

messense:
IMO it's fine to use `RuntimeError` since using `anyhow` usually means the developer don't care about the exact error type.

birkenfeld:
I tend to agree with @messense, this is a nice convenience feature for application code.

For library code, many Python libraries make their own exception class or hierarchy and convert upstream exceptions to those. This is obviously not possible to offer from PyO3 and needs some converter functions in the respective library.

I'm *not* sure about the formatting, though: is `{:?}` the correct one?  (`anyhow::Error` also can have context information; this should not be discarded if possible.)

birkenfeld:
Of course, it's also possible to provide, under the same feature, a method on `PyErr` for converting `anyhow::Error` into arbitrary exception classes if wanted.

mejrs:
Thanks!

Would you be willing to add an entry and doc example [here](https://pyo3.rs/main/doc/pyo3/index.html#optional-feature-flags)?

mejrs:
> my stance at the moment is that I'd like to be persuaded this is necessary.

I think this is also how I feel. 

I also think it is inconsistent that eyre has this conversion there and that we'd have anyhows here. But that isn't necessarily a problem to me.

Cargo.toml
chris-laplante: Didn't know that, will fix!

chris-laplante:
> I'm _not_ sure about the formatting, though: is `{:?}` the correct one? (`anyhow::Error` also can have context information; this should not be discarded if possible.)

From my testing, the context is correctly preserved :)

chris-laplante:
> Thanks!
> 
> Would you be willing to add an entry and doc example [here](https://pyo3.rs/main/doc/pyo3/index.html#optional-feature-flags)?

I assume this was directed at me :) in which case will do!

chris-laplante:
> > my stance at the moment is that I'd like to be persuaded this is necessary.
> 
> I think this is also how I feel.
> 
> I also think it is inconsistent that eyre has this conversion there and that we'd have anyhows here. But that isn't necessarily a problem to me.

I did think about submitting this change to anyhow instead. But, there are only a handful of popular error handling crates. Contrast that to the bazillion possible crates you may want to use in an application. I feel like it would be asking too much to add support for all those crates into the error handler ones.

davidhewitt (commented):
👍 it sounds like we're generally in favour of adding this, so let's proceed with doing so!

src/err/mod.rs
```diff
@@ -559,6 +559,13 @@ impl<'a> std::fmt::Display for PyDowncastError<'a> {
     }
 }
 
+#[cfg(feature = "anyhow")]
+impl From<anyhow::Error> for PyErr {
+    fn from(err: anyhow::Error) -> Self {
+        crate::exceptions::PyRuntimeError::new_err(format!("{:?}", err))
+    }
+}
```
davidhewitt, line 567: From a code organisation perspective I'd quite like this to go into a new file `src/conversions/anyhow.rs`, and it would be nice to also add tests for it!

guide/src/features.md
chris-laplante: Are you saying I should rename the 'Advanced features' section in features.md to 'Optional features'?

src/conversions/anyhow.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    #[pyfunction]
+    fn produce_ok_result() -> anyhow::Result<String> {
+        Ok(String::from("OK buddy"))
+    }
+
+    #[pyfunction]
+    fn produce_err_result() -> anyhow::Result<String> {
+        anyhow::bail!("error time")
+    }
```
davidhewitt, line 25: It's actually not ideal to use `#[pyfunction]` within unit tests of pyo3's lib, because it relies on `pyo3` being available as in the global namespace. We happen to have a self-dev-dependency on pyo3 but it causes unexpected breakages - see #1817 for more detail.

The point here is that it's probably better to just test that `let x: PyErr = produce_err_result().unwrap_err().into()` works ok, and could check the string contents of the error are as expected.

(If you really want to check the `#[pyfunction]` case, that's good too, but it'll need to go in a separate file in `tests/`.)

mejrs:
I think that is everything.

chris-laplante:
Thank you @mejrs and @davidhewitt so much for finishing this up :). I appreciate it!

PR #1823, opened 2021-08-23
ffi cleanup: pylifecycle to pystate

deantvv:
For #1289

### DIFF
--- a/src/ffi/cpython/pystate.rs
+++ b/src/ffi/cpython/pystate.rs
@@ -55,7 +55,10 @@ extern "C" {
     #[cfg(not(PyPy))]
     #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
     pub fn PyThreadState_Next(tstate: *mut PyThreadState) -> *mut PyThreadState;
-    // skipped PyThreadState_DeleteCurrent
+
+    #[cfg(py_sys_config = "WITH_THREAD")]
+    #[cfg_attr(PyPy, link_name = "PyPyThreadState_DeleteCurrent")]
+    pub fn PyThreadState_DeleteCurrent();
 }
 
 #[cfg(Py_3_9)]
--- a/src/ffi/pylifecycle.rs
+++ b/src/ffi/pylifecycle.rs
@@ -37,6 +37,8 @@ extern "C" {
     pub fn Py_GetPath() -> *mut wchar_t;
     pub fn Py_SetPath(arg1: *const wchar_t);
 
+    // skipped _Py_CheckPython3
+
     #[cfg_attr(PyPy, link_name = "PyPy_GetVersion")]
     pub fn Py_GetVersion() -> *const c_char;
     pub fn Py_GetPlatform() -> *const c_char;
--- a/src/ffi/pystate.rs
+++ b/src/ffi/pystate.rs
@@ -2,6 +2,7 @@
 use crate::ffi::moduleobject::PyModuleDef;
 use crate::ffi::object::PyObject;
 use std::os::raw::c_int;
+
 #[cfg(not(PyPy))]
 use std::os::raw::c_long;
 
@@ -27,15 +28,16 @@ extern "C" {
 
     #[cfg(all(Py_3_8, not(PyPy)))]
     #[cfg_attr(docsrs, doc(all(Py_3_8, not(PyPy))))]
-    pub fn PyInterpreterState_GetDict() -> *mut PyObject;
+    pub fn PyInterpreterState_GetDict(arg1: *mut PyInterpreterState) -> *mut PyObject;
 
     #[cfg(all(Py_3_7, not(PyPy)))]
     #[cfg_attr(docsrs, doc(all(Py_3_7, not(PyPy))))]
-    pub fn PyInterpreterState_GetID() -> i64;
+    pub fn PyInterpreterState_GetID(arg1: *mut PyInterpreterState) -> i64;
 
     #[cfg(not(PyPy))]
     #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
     pub fn PyState_AddModule(arg1: *mut PyObject, arg2: *mut PyModuleDef) -> c_int;
+
     #[cfg(not(PyPy))]
     #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
     pub fn PyState_RemoveModule(arg1: *mut PyModuleDef) -> c_int;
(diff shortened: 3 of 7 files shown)


src/ffi/pystate.rs
```diff
@@ -1,13 +1,17 @@
...
 pub const MAX_CO_EXTRA_USERS: c_int = 255;
 
-opaque_struct!(PyThreadState);
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct PyThreadState {
+    pub ob_base: PyObject,
+    pub interp: *mut PyInterpreterState,
+}
+
```
davidhewitt, line 14: Can you show me where you got this from?

I removed this exact definition in #1787, as I understood it to be incorrect. It looked to me like the corresponding struct in the C header is super complex and I didn't think it was worth trying to wrap.

https://github.com/python/cpython/blob/e41912c6348362489d2514565a70782591f23902/Include/cpython/pystate.h#L73

src/ffi/pystate.rs
```diff
@@ -21,43 +25,45 @@ extern "C" {
     #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
     pub fn PyInterpreterState_Delete(arg1: *mut PyInterpreterState);
 
-    #[cfg(all(Py_3_9, not(PyPy)))]
+    #[cfg(all(any(not(Py_LIMITED_API), Py_3_9), not(PyPy)))]
     #[cfg_attr(docsrs, doc(all(Py_3_9, not(PyPy))))]
-    pub fn PyInterpreterState_Get() -> *mut PyInterpreterState;
+    pub fn PyInterpreterState_Get(arg1: *mut PyInterpreterState) -> *mut PyInterpreterState;
```
davidhewitt, line 30: This change looks incorrect, the corresponding C definition takes no arguments.

src/ffi/pystate.rs
```diff
@@ -21,43 +25,45 @@ extern "C" {
     #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
     pub fn PyInterpreterState_Delete(arg1: *mut PyInterpreterState);
 
-    #[cfg(all(Py_3_9, not(PyPy)))]
+    #[cfg(all(any(not(Py_LIMITED_API), Py_3_9), not(PyPy)))]
```
davidhewitt, line 28: This isn't exactly right: on Python 3.8 and below this API would have been called `_PyInterpreterState_Get` (with a leading underscore).

I didn't think it was worth the added complexity of covering this, hence why only enabled on 3.9 and up (where it was immediately added to the stable API).

src/ffi/pystate.rs
```diff
@@ -21,43 +25,45 @@ extern "C" {
...
+    #[cfg(all(any(not(Py_LIMITED_API), Py_3_8), not(PyPy)))]
     #[cfg_attr(docsrs, doc(all(Py_3_8, not(PyPy))))]
-    pub fn PyInterpreterState_GetDict() -> *mut PyObject;
+    pub fn PyInterpreterState_GetDict(arg1: *mut PyInterpreterState) -> *mut PyObject;
 
-    #[cfg(all(Py_3_7, not(PyPy)))]
+    #[cfg(all(any(not(Py_LIMITED_API), Py_3_7), not(PyPy)))]
     #[cfg_attr(docsrs, doc(all(Py_3_7, not(PyPy))))]
-    pub fn PyInterpreterState_GetID() -> i64;
+    pub fn PyInterpreterState_GetID(arg1: *mut PyInterpreterState) -> i64;
```
davidhewitt, line 38: Similar point about these having different names before they were added to the limited API.

src/ffi/pystate.rs
```diff
@@ -21,43 +25,45 @@ extern "C" {
...
-    #[cfg(not(PyPy))]
-    #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
+    #[cfg(not(Py_LIMITED_API))]
     pub fn PyState_AddModule(arg1: *mut PyObject, arg2: *mut PyModuleDef) -> c_int;
-    #[cfg(not(PyPy))]
-    #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
-    pub fn PyState_RemoveModule(arg1: *mut PyModuleDef) -> c_int;
+
+    #[cfg(not(Py_LIMITED_API))]
+    pub fn PyState_RemoveModule(arg2: *mut PyModuleDef) -> c_int;
```
davidhewitt, line 44: These APIs are definitely available on the limited API as far aw we care, the header suggests that they're available on limited API since Python 3.3.

Also, if these are available for PyPy we should have the PyPy link names set. However from my own checks I couldn't find them, hence the `not(PyPy)` I'd previously written.

src/ffi/cpython/pystate.rs
```diff
@@ -55,7 +55,10 @@ extern "C" {
     #[cfg(not(PyPy))]
     #[cfg_attr(docsrs, doc(cfg(not(PyPy))))]
     pub fn PyThreadState_Next(tstate: *mut PyThreadState) -> *mut PyThreadState;
-    // skipped PyThreadState_DeleteCurrent
+
+    #[cfg(py_sys_config = "WITH_THREAD")]
+    #[cfg_attr(PyPy, link_name = "PyPyThreadState_DeleteCurrent")]
+    pub fn PyThreadState_DeleteCurrent();
```
davidhewitt, line 61: 👍

src/ffi/pylifecycle.rs
```diff
@@ -7,6 +7,7 @@ extern "C" {
     pub fn Py_Initialize();
     pub fn Py_InitializeEx(arg1: c_int);
     pub fn Py_Finalize();
+    #[cfg(any(not(Py_LIMITED_API), Py_3_6))]
```
davidhewitt, line 10: This isn't necessary - `Py_3_6` is always set, because we don't support older Python versions than 3.6.

```suggestion
```

davidhewitt (changes requested):
Thank you for working on this!

I have a few concerns about the `pystate` bindings, commented below. Everything else looks great and correct!

src/ffi/pystate.rs
deantvv: Thanks for catching this.

src/ffi/pystate.rs
deantvv: Change as suggested. Thanks!

src/ffi/pystate.rs
deantvv: Changed as suggested.

src/ffi/pystate.rs
deantvv: I checked that `PyState_RemoveModule` is not in PyPy. Will revert this change.

src/ffi/pylifecycle.rs
deantvv: Doesn't know the Python version 3.6 support before. Now got it and fixed.

deantvv:
@davidhewitt I got it almost all wrong because I wrongly use the old branch which caused a lot of unnecessary trouble. Thanks for carefully reviewing the code! Wish I won't get it wrong the next time.

davidhewitt (approved):
👍 not a problem at all, mistakes happen! I just try to give PRs thorough reviews so that the PyO3 codebase is as correct as possible 😄 

This new patch looks great to me, thanks very much for helping clean all this up!

PR #1825, opened 2021-08-24
list/tuple/sequence: implement Index

birkenfeld:
See #1667


davidhewitt (commented):
👍 looks like a great start!

What do you think of supporting full slice indexing? I think this means either implementing [`SliceIndex`](https://doc.rust-lang.org/std/slice/trait.SliceIndex.html) or just implementing `Index` directly for `Range<usize>`, `RangeFull`, `RangeInclusive<usize>` etc.

birkenfeld:
Yeah, I though about it, but that runs into API consistency again: do we want to panic on invalid indices or not?
In the end, I didn't (and don't) care about it enough to make the change.

davidhewitt (approved):
That's fair enough. I think for those out of bounds slice indexing we should probably panic (to match Rust), given that we already deviated from Python by not supporting negative indexing. That also has the nice dual that again `get_slice` is the non-panicking companion, even if it doesn't return an error.

I'm happy to implement `SliceIndex` myself in a follow-up.

PR #1826, opened 2021-08-24
sequence: tidy up implementation

davidhewitt:
Just noticed this while reviewing #1803


birkenfeld (approved):
BTW, when I looked at the code earlier I got a bit worried about returning `&PySequence` here: there is no obligation for the functions on the (C)Python side to return something usable as a sequence.

But since `PySequence` is just an abstract API, not a concrete object type, there is no unsoundness, and it makes sense to keep the convenience on the Rust side.

davidhewitt:
👍 agreed, as I've understood it, calling the pysequence APIs with an incompatible object will just raise an error.

PR #1828, opened 2021-08-24
list/tuple: rename slice -> get_slice

birkenfeld:
Only tuple's method is renamed; list's method is a new API in 0.15.


davidhewitt (changes requested):
👍 I 100% agree with doing this!

Can you please re-add `.slice()` as a deprecated API for one release cycle? It makes it easier for users to upgrade their code.

birkenfeld:
So, should the deprecated method continue to use `isize` indices as in 0.14? Otherwise, people will get a compile error instead of a deprecation warning anyway...

davidhewitt:
Oooh good question! I think the original `isize` implementation is probably best? (Could potentially even copy it from 0.14 rather than try to implement it using the new APIs, for maximum backcompat.)

davidhewitt (approved):
Thanks, LGTM!

PR #1829, opened 2021-08-24
list,tuple,sequence: add slice indexing

### LINKED ISSUE #1667: PyList indexing: negative indices?

davidhewitt:
Follow up to #1825 to add range indexing.

Given that the decision in #1733 was that we are trying to make these APIs feel as much like idiomatic Rust as possible, I opted to make them panic on out of bounds.

This probably closes #1667 ?


src/types/list.rs
```diff
@@ -177,17 +177,83 @@ impl PyList {
...
+            .unwrap_or_else(|_| panic_out_of_range!(index, self.len()))
+    }
+}
+
+impl Index<Range<usize>> for PyList {
+    type Output = PyList;
+
+    fn index(&self, Range { start, end }: Range<usize>) -> &Self::Output {
+        let len = self.len();
+        if start >= len {
```
birkenfeld, line 200: Actually, Rust allows `start` to be `== len` too.

You need an additional check for `start <= end`.

src/types/list.rs
```diff
@@ -177,17 +177,83 @@ impl PyList {
...
+        }
+    }
+}
+
+impl Index<RangeFrom<usize>> for PyList {
+    type Output = PyList;
+
+    fn index(&self, RangeFrom { start }: RangeFrom<usize>) -> &Self::Output {
+        let len = self.len();
+        if start >= len {
```
birkenfeld, line 215: Same here.

src/types/list.rs
```diff
@@ -177,17 +177,83 @@ impl PyList {
...
+
+impl Index<RangeInclusive<usize>> for PyList {
+    type Output = PyList;
+
+    fn index(&self, range: RangeInclusive<usize>) -> &Self::Output {
+        &self[*range.start()
+            ..range
+                .end()
+                .checked_add(1)
+                .expect("range end exceeds Python limit")]
```
birkenfeld, line 240: ugh, that rustfmt output...

src/types/sequence.rs
```diff
@@ -254,13 +254,86 @@ impl PySequence {
...
+    }
+}
+
+impl Index<Range<usize>> for PySequence {
+    type Output = PyAny;
+
+    fn index(&self, Range { start, end }: Range<usize>) -> &Self::Output {
+        let len = self.len().expect("failed to get sequence length");
+        if start >= len {
+            panic_out_of_range!(start);
```
birkenfeld, line 278: I guess the length is now known here too, for the panic message.

src/types/sequence.rs
```diff
@@ -254,13 +254,86 @@ impl PySequence {
...
     fn index(&self, index: usize) -> &Self::Output {
-        self.get_item(index).unwrap_or_else(|_| {
-            panic!("index {} out of range for sequence", index);
-        })
+        self.get_item(index)
+            .unwrap_or_else(|_| panic_out_of_range!(index))
+    }
+}
+
+impl Index<Range<usize>> for PySequence {
```
birkenfeld, line 272: These impls mostly look identical - use a macro?

birkenfeld (changes requested):

PR #1844, opened 2021-08-28
types: add PyMapping

davidhewitt:
Adds a `PyMapping` type - similar to `PySequence`, but specially to interact with the mapping protocol.

It doesn't have `get_item`, `set_item`, and `del_item` methods because they're already present because of `Deref` to `PyAny`. However it contains a few other helpful pieces such as `.contains_key()`, `.keys()`, `.values()`, and `.items()`.

Seemed like it might be useful for https://github.com/davidhewitt/pythonize/pull/16 (cc @i1i1)

WIP - needs tests.


src/types/mapping.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Determines if self contains the key `key`.
+    ///
+    /// This is equivalent to the Python expression `key in self`.
+    #[inline]
+    pub fn contains_key<K>(&self, key: K) -> PyResult<bool>
+    where
+        K: ToBorrowedObject,
+    {
+        let r = key.with_borrowed_ptr(self.py(), |ptr| unsafe {
+            ffi::PyMapping_HasKey(self.as_ptr(), ptr)
```
birkenfeld, line 44: If we use `HasKey` which suppresses exceptions, the method can return just `bool`?

Otherwise, use `GetItem` internally to propagate the exceptions...

birkenfeld (commented):
As I proposed in #1845 for sequences, the API here should mirror PyDict as closely as possible.

davidhewitt:
👍 will update this to follow what you've done in #1849 but for `PyDict` / `PyMapping`

davidhewitt:
Updated to contain the other relevant `PyDict` APIs and added tests.

In the end I couldn't come up with a good decision for `.contains_key`. I felt it was potentially confusing to implement it using `PyMapping_HasKey`, because the suppressed exception might not be a `KeyError` but some other cause (e.g. failed comparison when comparing two keys).

So for now I left `.contains_key` out.

davidhewitt:
Any objections if I merge this?

birkenfeld (approved):
Could add an entry in `guide/src/conversions/tables.md` I guess.

davidhewitt:
👍 good idea, done!

Gobot1234:
Is there a reason that `contains_key` wasn't implemented, it would be much better than `obj.getattr("__contains__")?.call1((key,))?.is_true()?`

davidhewitt:
See https://github.com/PyO3/pyo3/pull/1844#issuecomment-913787734

I just couldn't make an obvious decision on the semantics of `contains_key` in this initial PR. If you're willing to come up with an appropriate implementation and happy to open a PR for it, I'd be happy to add it to this type.

FYI you might also want to know about #2072

birkenfeld:
Don't we have a method for the `in` operator on PyAny?

Gobot1234:
Not that I could find?

PR #1848, opened 2021-08-29
pyo3-build-config: make InterpreterConfig public

indygreg:
PyOxidizer will want to create interpreter config files. Rather
than reinvent the logic for reading/writing these files, I think
it makes sense to consume the `pyo3-build-config` crate so we can
use the `InterpreterConfig` type directly. But the symbol needs
to be public to allow us to do that.

(Strictly speaking I could probably hack my way into getting an
instance of `InterpreterConfig` via `get()` and use that. But that
is exceptionally hacky.)

### DIFF
--- a/pyo3-build-config/src/lib.rs
+++ b/pyo3-build-config/src/lib.rs
@@ -13,11 +13,7 @@ use std::io::Cursor;
 
 use once_cell::sync::OnceCell;
 
-use impl_::InterpreterConfig;
-
-// Used in `pyo3-macros-backend`; may expose this in a future release.
-#[doc(hidden)]
-pub use impl_::PythonVersion;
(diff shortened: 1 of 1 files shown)


davidhewitt:
👍 I think this needs to happen.

Shall we go a little further and remove the `#[doc(hidden)]`?

I think to make this useful in practice you may also want `PythonVersion`, `PythonImplementation` , `BuildFlags` and `BuildFlag` made public too?

indygreg:
Yeah, we need to make all referenced fields public as well. I would have realized that if I actually tried to use this branch :) Good catch.

I rebased and force pushed a change to expose the additional types.

davidhewitt (approved):
👍 thanks!

Issue #1845, opened 2021-08-29
Complete sequence API for list/tuple?


birkenfeld:
I propose to add all methods present in `PySequence` (such as `del_item`) to `PyList` and `PyTuple`.  It is quite convenient not to have to cast to `PySequence` for some operations but not for others.

davidhewitt:
Specifically on `del_item` - it should already be available thanks to deref to `PyAny`?

Though in general I'm in favour of making it easier for users, so happy to find improvements here.

We could also have `.as_sequence()` method on `PyList` and `PyTuple` to make the cast less awkward?

birkenfeld:
Well, there's a few inconsistencies: for example, since the `del_item` is the generic version it will take negative indices :)

davidhewitt:
Ah, that's a good point! In which case a specialized `del_item` on `PyList` would be a good idea!

PR #1849, opened 2021-08-30
PyList: add more sequence APIs

birkenfeld:
See #1845

Since these are just literal copies of the sequence methods, this makes me think - could `PyList` deref to `PySequence`?

### DIFF
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -5,8 +5,10 @@
 use crate::err::{self, PyResult};
 use crate::ffi::{self, Py_ssize_t};
 use crate::internal_tricks::get_ssize_index;
+use crate::types::PySequence;
 use crate::{
-    AsPyPointer, IntoPy, IntoPyPointer, PyAny, PyObject, Python, ToBorrowedObject, ToPyObject,
+    AsPyPointer, IntoPy, IntoPyPointer, PyAny, PyObject, PyTryFrom, Python, ToBorrowedObject,
+    ToPyObject,
 };
 
 /// Represents a Python `list`.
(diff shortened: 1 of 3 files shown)


src/types/list.rs
```diff
@@ -132,6 +132,50 @@ impl PyList {
...
+
+    /// Assigns the sequence `seq` to the slice of `self` from `low` to `high`.
+    ///
+    /// This is equivalent to the Python statement `self[low:high] = v`.
+    #[inline]
+    pub fn set_slice(&self, low: usize, high: usize, seq: &PyAny) -> PyResult<()> {
+        unsafe {
+            err::error_on_minusone(
+                self.py(),
+                ffi::PySequence_SetSlice(
```
MatthieuDartiailh, line 156: The CPython API provides `PyList_SetSlice` that may be optimized for list and may be worth using here.  `PyList_GetSlice` also exists but there is no  `PyList_DelSlice`.

birkenfeld: Good catch, from other methods I had assumed that all existing `PyList_` functions were already covered. I'll have to recheck!

birkenfeld:
Now also forwarding the methods to `PySequence` instead of copy-pasting the impl.

davidhewitt:
Thanks, I've definitely come around to this being the right thing to do.

> Since these are just literal copies of the sequence methods, this makes me think - could `PyList` deref to `PySequence`?

It's tempting, however I've been wondering also whether we should have Python types `Deref` to their Python base, just like [what web-sys does](https://rustwasm.github.io/docs/wasm-bindgen/web-sys/inheritance.html). `PyList` deref to `PySequence` doesn't fit that pattern.

How about adding `.as_sequence()` method, and maybe even `impl AsRef<PySequence> for PyList`? That would also be useful within the implementation of these methods?

Also, needs CHANGELOG entry 😄

davidhewitt (approved):
Thanks, this looks good to me! I still think `.as_sequence()`  could be useful, however I'm happy to see that as a separate PR...

PR #1853, opened 2021-09-01
pyo3-build-config: derive Clone on BuildFlags

indygreg:
This isn't strictly necessary since you can `.0.clone()`. But it is
more ergonomic.

Found this when porting PyOxidizer to the 0.14.5 branch.


davidhewitt (approved):
👍 there's one or two places I would have used this too!

PR #1861, opened 2021-09-03
ffi: cleanup setobject

deantvv:
For #1289

- Change the order of declaration as cpython does
- Document skipped function/macro

### DIFF
--- a/src/ffi/setobject.rs
+++ b/src/ffi/setobject.rs
@@ -1,30 +1,12 @@
 use crate::ffi::object::*;
-use crate::ffi::pyport::{Py_hash_t, Py_ssize_t};
+#[cfg(not(Py_LIMITED_API))]
+use crate::ffi::pyport::Py_hash_t;
+use crate::ffi::pyport::Py_ssize_t;
 use std::os::raw::c_int;
-
-#[cfg_attr(windows, link(name = "pythonXY"))]
-extern "C" {
-    #[cfg_attr(PyPy, link_name = "PyPySet_Type")]
-    pub static mut PySet_Type: PyTypeObject;
-    #[cfg_attr(PyPy, link_name = "PyPyFrozenSet_Type")]
-    pub static mut PyFrozenSet_Type: PyTypeObject;
-    pub static mut PySetIter_Type: PyTypeObject;
-
-    #[cfg(PyPy)]
-    #[link_name = "PyPyFrozenSet_CheckExact"]
-    pub fn PyFrozenSet_CheckExact(ob: *mut PyObject) -> c_int;
-
-    #[cfg(PyPy)]
-    #[link_name = "PyPyAnySet_CheckExact"]
-    pub fn PyAnySet_CheckExact(ob: *mut PyObject) -> c_int;
-
-    #[cfg(PyPy)]
-    #[link_name = "PyPySet_Check"]
-    pub fn PySet_Check(ob: *mut PyObject) -> c_int;
-}
 
 pub const PySet_MINSIZE: usize = 8;
 
+#[cfg(not(Py_LIMITED_API))]
 #[repr(C)]
 #[derive(Debug)]
 pub struct setentry {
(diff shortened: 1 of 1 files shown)


src/ffi/setobject.rs
```diff
@@ -46,12 +29,78 @@ pub struct PySetObject {
...
+    #[cfg_attr(PyPy, link_name = "PyPySet_Discard")]
+    pub fn PySet_Discard(set: *mut PyObject, key: *mut PyObject) -> c_int;
+    #[cfg_attr(PyPy, link_name = "PyPySet_Pop")]
+    pub fn PySet_Pop(set: *mut PyObject) -> *mut PyObject;
+    #[cfg_attr(PyPy, link_name = "PyPySet_Size")]
+    pub fn PySet_Size(anyset: *mut PyObject) -> Py_ssize_t;
+}
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
```
davidhewitt, line 79: Usually the `windows` link name attribute is only needed for `static mut` symbols, you might be able to merge this `extern` block with the one above:

```suggestion
```

src/ffi/setobject.rs
```diff
@@ -46,12 +29,78 @@ pub struct PySetObject {
...
 #[cfg(not(PyPy))]
 pub unsafe fn PyFrozenSet_CheckExact(ob: *mut PyObject) -> c_int {
     (Py_TYPE(ob) == &mut PyFrozenSet_Type) as c_int
 }
 
+#[inline]
+pub unsafe fn PyFrozenSet_Check(ob: *mut PyObject) -> c_int {
+    (Py_TYPE(ob) == &mut PyFrozenSet_Type
+        || PyType_IsSubtype(Py_TYPE(ob), &mut PyFrozenSet_Type) != 0) as c_int
+}
```
davidhewitt, line 95: This one should also be `not(PyPy)` like `PyFrozenSet_CheckExact`  - I see a `PyPyFrozenSet_Check` in my list of symbols from last time I extracted the PyPy lib.

src/ffi/setobject.rs
```diff
@@ -65,42 +114,17 @@ pub unsafe fn PyAnySet_Check(ob: *mut PyObject) -> c_int {
         || PyType_IsSubtype(Py_TYPE(ob), &mut PyFrozenSet_Type) != 0) as c_int
 }
 
+// skipped PySet_CheckExact
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
```
davidhewitt, line 119: I think no need for this attribute (the imported symbol is a function, not `static mut`)

```suggestion
```

davidhewitt (commented):
Thanks again - looks great! Just a few final pieces I spotted when cross-checking...

src/ffi/setobject.rs
deantvv: > Usually the `windows` link name attribute is only needed for `static mut` symbols
Could you please elaborate why?
In my understanding, `link` statement will be emitted in build script. Why this is necessary to add the attribute only to `static mut` symbols but not function?
Thanks in advance. I couldn't understand this after researching for several hours.

>, you might be able to merge this `extern` block with the one above:
Will do.

src/ffi/setobject.rs
deantvv: Same as above.

src/ffi/setobject.rs
deantvv: Thanks for catching this.

davidhewitt (approved):
Perfect, thanks again!

src/ffi/setobject.rs
davidhewitt: Yeah, unfortunately I haven't found this documented anywhere. I suspect it's not intended, however in practice I've found that we experience windows linker errors for `static mut` symbols if they aren't covered explicitly by a `link_name` attribute. I should really report this upstream...

PR #1863, opened 2021-09-06
conversions: move more features inside the module

davidhewitt:
Moves the `hashbrown` conversions into `src/conversions/hashbrown.rs`, and also relocates the `num_bigint` and `num_complex` files inside that subdirectory.


mejrs (commented):
Thanks, this looks good. Do the links work in lib.rs? I can't render the docs myself at the moment.

davidhewitt:
I'm pretty sure they do, tested locally! I'll go ahead and merge this and then double-check https://pyo3.rs/main/doc

PR #1865, opened 2021-09-10
chore: update rust 1.55

mejrs:


davidhewitt (changes requested):
In `test_compile_error` we use `rustversion` to skip running certain tests on older compilers - you'll need to make a new `test_rust_1_55` function and move each test which has changed output into it.

mejrs:
I have done that now, but I don't understand how that influences whether any of the ci tests fail.

for reference, here is the error:

```text
test tests/ui/invalid_pymethod_receiver.rs ... mismatch

EXPECTED:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error[E0277]: the trait bound `i32: From<&PyCell<MyClass>>` is not satisfied
   --> $DIR/invalid_pymethod_receiver.rs:8:43
    |
8   |     fn method_with_invalid_self_type(slf: i32, py: Python, index: u32) {}
    |                                           ^^^ the trait `From<&PyCell<MyClass>>` is not implemented for `i32`
    |
    = help: the following implementations were found:
              <i32 as From<NonZeroI32>>
              <i32 as From<bool>>
              <i32 as From<i16>>
              <i32 as From<i8>>
            and 2 others
    = note: required because of the requirements on the impl of `Into<i32>` for `&PyCell<MyClass>`
    = note: required because of the requirements on the impl of `TryFrom<&PyCell<MyClass>>` for `i32`
note: required by `std::convert::TryFrom::try_from`
   --> $DIR/mod.rs:477:5
    |
477 |     fn try_from(value: T) -> Result<Self, Self::Error>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈

ACTUAL OUTPUT:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error[E0277]: the trait bound `i32: From<&PyCell<MyClass>>` is not satisfied
   --> $DIR/invalid_pymethod_receiver.rs:8:43
    |
8   |     fn method_with_invalid_self_type(slf: i32, py: Python, index: u32) {}
    |                                           ^^^ the trait `From<&PyCell<MyClass>>` is not implemented for `i32`
    |
    = help: the following implementations were found:
              <i32 as From<NonZeroI32>>
              <i32 as From<bool>>
              <i32 as From<i16>>
              <i32 as From<i8>>
            and 2 others
    = note: required because of the requirements on the impl of `Into<i32>` for `&PyCell<MyClass>`
    = note: required because of the requirements on the impl of `TryFrom<&PyCell<MyClass>>` for `i32`
note: required by `std::convert::TryFrom::try_from`
   --> $DIR/mod.rs:477:5
    |
477 |     fn try_from(value: T) -> Result<Self, Self::Error>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
note: If the actual output is the correct output you can bless it by rerunning
      your test with the environment variable TRYBUILD=overwrite
```

davidhewitt:
👍 sorry I should have explained better - doing the `rustversion` stuff is useful because then `cargo test` still works for people on Rust 1.54. (and backwards all the way to MSRV)

So mostly not for CI, though it is useful when we will update MSRV because then it should just be a case of removing the old version gates.

The continued failure on CI is a bit confusing - did you update it using `TRYBUILD=overwrite` environment variable?

mejrs:
I guess the issue is that locally I get different errors than in CI? 🤔

birkenfeld:
The "actual output" in CI is quite strange.

It has the added padding in the line numbers, i.e.
```
    |
8   |     fn method_with_invalid_self_type(slf: i32, py: Python, index: u32) {}
^^^^^ this padding
```
which is present in (our local) 1.55 because of the added line with line number 477 below. But there is no such line in CI, so the padding should be the same as in 1.54.

davidhewitt:
Hmmm, I'm puzzled why CI would be consistently different on all OSes! Could it be an environment variable? I'll see if I can repro later...

mejrs:
So it looks like this makes CI happy. However this will error if I run `cargo test` locally. 🤔

davidhewitt:
So it looks like this is probably a rust 1.55 issue related to the new diagnostic?

In #1867 we get 

```
error[E0277]: the trait bound `i32: From<&PyCell<MyClass>>` is not satisfied
   --> tests/test_class_basics.rs:350:43
    |
350 |     fn method_with_invalid_self_type(slf: i32, py: Python, index: u32) {}
    |                                           ^^^ the trait `From<&PyCell<MyClass>>` is not implemented for `i32`
    |
    = help: the following implementations were found:
              <i32 as From<NonZeroI32>>
              <i32 as From<bool>>
              <i32 as From<i16>>
              <i32 as From<i8>>
            and 2 others
    = note: required because of the requirements on the impl of `Into<i32>` for `&PyCell<MyClass>`
    = note: required because of the requirements on the impl of `TryFrom<&PyCell<MyClass>>` for `i32`
note: required by `std::convert::TryFrom::try_from`
```

... which is exactly the same form as what you've found here, with the definition of `std::convert::TryFrom::try_from` truncated. Still haven't managed to repro this behaviour locally.

davidhewitt:
Aha, I've figured it out! If I uninstall the `rust-src` component locally, I get the same error message as in CI. Presumably because `rustc` then doesn't have a source location to display the definition of `std::convert::TryFrom::try_from`.

So assuming that most local installations will contain the `rust-src` component (rustup installs it by default afaik), we should add this to CI? I think it should be as simple as adding `components: rust-src` after line 154 of `ci.yml`

https://github.com/PyO3/pyo3/blob/0c5ac220de0227a209b2acb2111731ba99db0548/.github/workflows/ci.yml#L154

Also, given that this is a lot of WIP commits it would be nice to squash them into one commit before merging please 🙂

tests/ui/abi3_nativetype_inheritance.stderr
```diff
@@ -1,11 +1,16 @@
 error[E0277]: the trait bound `PyDict: PyClass` is not satisfied
```
davidhewitt, line 1: Please move this test into the `tests_rust_1_55` function too!

mejrs:
I'll do those things tonight :)

birkenfeld:
> Aha, I've figured it out! If I uninstall the `rust-src` component locally, I get the same error message as in CI. Presumably because `rustc` then doesn't have a source location to display the definition of `std::convert::TryFrom::try_from`.

Ugh, that's nasty. I don't know if trybuild can/wants to do anything about that, but it might be at least a good idea to bring it to their attention.

davidhewitt:
Good idea, done.

mejrs:
@davidhewitt Looks like this works 🎉 

Can you "Squash and merge" this? Thanks!

davidhewitt (approved):
Perfect, thanks!

PR #1866, opened 2021-09-11
ffi: cleanup sliceobject

deantvv:
For #1289

### DIFF
--- a/src/ffi/sliceobject.rs
+++ b/src/ffi/sliceobject.rs
@@ -12,6 +12,15 @@ pub unsafe fn Py_Ellipsis() -> *mut PyObject {
     &mut _Py_EllipsisObject
 }
 
+#[cfg(not(Py_LIMITED_API))]
+#[repr(C)]
+pub struct PySliceObject {
+    pub ob_base: PyObject,
+    pub start: *mut PyObject,
+    pub stop: *mut PyObject,
+    pub step: *mut PyObject,
+}
+
 #[cfg_attr(windows, link(name = "pythonXY"))]
 extern "C" {
     #[cfg_attr(PyPy, link_name = "PyPySlice_Type")]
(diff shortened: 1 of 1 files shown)


src/types/slice.rs
```diff
@@ -61,6 +61,7 @@ impl PySlice {
             let start: isize = 0;
             let stop: isize = 0;
             let step: isize = 0;
+            #[allow(deprecated)]
             let r = ffi::PySlice_GetIndicesEx(
```
davidhewitt, line 65: What's the recommended replacement to using `PySlice_GetIndicesEx`? It would be great to remove this usage of a deprecated API...

davidhewitt (commented):
Thanks! Looks good, just one suggestion related to the deprecation...

src/types/slice.rs
deantvv: In cpython, they define a macro `PySlice_GetIndicesEx` (shown below)

```c
#define PySlice_GetIndicesEx(slice, length, start, stop, step, slicelen) (  \
    PySlice_Unpack((slice), (start), (stop), (step)) < 0 ?                  \
    ((*(slicelen) = 0), -1) :                                               \
    ((*(slicelen) = PySlice_AdjustIndices((length), (start), (stop), *(step))), \
     0))
```

Which I didn't implement since currently we didn't have `PySlice_Unpack` and `PySlice_AdjustIndices`.
These method seems not so hard. If you like, I can also implement these.

davidhewitt: That would be great, yes please!

Given the function is deprecated in Python 3.7, and there's [quite a complex condition on the macro, `PySlice_Unpack` and `PySlice_AdjustIndices` which suggest they were added inPython 3.6.1](https://github.com/python/cpython/blob/5277ffe12d492939544ff9c54a3aaf448b913fb3/Include/sliceobject.h#L49), I think we can only use the new macro form in Python 3.7 and up.

davidhewitt (approved):
Looks great, thank you! If you rebase on `main` it should fix the CI issues with Rust 1.55 and then we can merge 👍

davidhewitt:
Perfect, thanks again!

PR #1867, opened 2021-09-12
ci: try to debug trybuild issue

davidhewitt:
For #1865 

I get the same output as everyone else locally, for some reason the error message is cut off in CI. Just adding the code which fails to compile to a different file so we can see what CI gives us...


davidhewitt:
We learned what we wanted; closing!

PR #1868, opened 2021-09-13
clean up lib.rs

mejrs:
- moves serde.rs to /conversions, like the other optional features
 - moves the proc macro imports to the top level so they are more discoverable on the front page
 - moves the link declarations in lib.rs down, to remove the noise of defining links inside the text itself
 - moves the macro_rules! macros to macros.rs to declutter lib.rs

### DIFF
--- a/src/conversions/mod.rs
+++ b/src/conversions/mod.rs
@@ -7,3 +7,4 @@ pub mod num_bigint;
 pub mod num_complex;
 mod osstr;
 mod path;
+pub mod serde;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -353,7 +353,7 @@ pub mod proc_macro {
 
 #[cfg_attr(docsrs, doc(cfg(feature = "macros")))]
 #[cfg(feature = "macros")]
-pub use pyo3_macros::{pyclass, pyfunction, pymethods, pymodule, pyproto};
+pub use pyo3_macros::{pyclass, pyfunction, pymethods, pymodule, pyproto, FromPyObject};
 
 #[macro_use]
 mod macros;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,11 +1,15 @@
 #![cfg_attr(feature = "nightly", feature(specialization))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
-
+#![deny(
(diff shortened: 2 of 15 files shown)


src/lib.rs
```diff
@@ -327,166 +336,27 @@ pub mod pyclass;
...
-        $crate::py_run_impl!($py, $($val)+, &$crate::unindent::unindent($code))
-    }};
-    ($py:expr, *$dict:expr, $code:literal) => {{
-        $crate::py_run_impl!($py, *$dict, $crate::indoc::indoc!($code))
-    }};
-    ($py:expr, *$dict:expr, $code:expr) => {{
-        $crate::py_run_impl!($py, *$dict, &$crate::unindent::unindent($code))
-    }};
-}
+pub use pyo3_macros::{pyclass, pyfunction, pymethods, pymodule, pyproto};
```
davidhewitt, line 356: I wonder if we should add a `FromPyObject` import here so that users can write `#[derive(pyo3::FromPyObject)]`? (I think we can also switch to using this in `test_proc_macro_hygiene`.)

And if we add that, I think it's equivalent to

```suggestion
pub use pyo3_macros::*;
```

src/conversions/mod.rs
```diff
@@ -7,3 +7,4 @@ pub mod num_bigint;
 pub mod num_complex;
 mod osstr;
 mod path;
+pub mod serde;
```
davidhewitt, line 10: 👍 it doesn't feel _exactly_ like a conversion to me but I do like having it in the same place as all the other optional dependencies!

davidhewitt (approved):
Thanks, this is a much-appreciated tidy-up! One thought on the proc macro imports...

src/lib.rs
mejrs: I've imported it explicitly. If I do that with a glob import, this happens, which makes it really confusing where a macro is coming from.

```rust
pub use pyo3_macros::*;

#[macro_use]
mod macros;
```

mejrs:
Looks like CI is breaking on windows runners, see also https://github.com/rust-lang/rust/issues/88924

davidhewitt:
Mmm, following that thread it looks like Github did something which has now been reverted! 😬

Issue #1782, opened 2021-08-11
Python 3.6 end-of-life, MSRV to 1.48?


davidhewitt:
Python 3.6 end-of-life is scheduled for 23rd December this year. (https://endoflife.date/python)

This means our first breaking release in 2022 can drop support for Python 3.6 and also bump MSRV. It looks like Debian bullseye will release with Rust 1.48, so I expect that to be the next MSRV as per https://github.com/PyO3/pyo3/blob/main/Contributing.md#python-and-rust-version-support-policy

This ticket is just a placeholder which we can use to mark TODOs in code related to Python 3.6 or MSRV which we can clean up once reaching Python 3.7 and Rust 1.48 as minimum versions.

birkenfeld:
+1 for making these bumps together.

PR #1874, opened 2021-09-16
gil: try to initialize threads on Python 3.6 if possible

### LINKED ISSUE #1870: PyEval_InitThreads is not called in `#[pymodule]` for Python3.6
## 🐛 Bug Reports

I was writing an extension module (python call rust) using pyo3 and recently upgraded the pyo3 dependency from 0.11 to 0.14. 

Then once the`Python::with_gil()` was called, the pyo3 would panic:

```
thread 'tokio-runtime-worker' panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`: Python threading is not initalized and the `auto-initialize` feature is not enabled.

Consider calling `pyo3::prepare_freethreaded_python()` before attempting to use Python APIs.', /home/andylokandy/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.14.4/src/gil.rs:232:21
```

Then I added the `auto-initialize` feature flag and also tried to add `pyo3::prepare_freethreaded_python()` to the first line in `#[pymodule]` function without `auto-initialize` feature flag, it still panicked:

```
thread '<unnamed>' panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`', /home/andylokandy/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.14.4/src/gil.rs:84:13
```

After investigating a little bit, I believe that the latter panic in `gil.rs:84:13` is caused by the uninitialized threading while the interpreter is already initialized by `#[pymodule]`.

The `PyEval_InitThreads` is called in `#[pymodule]` before https://github.com/PyO3/pyo3/pull/1630, in which @davidhewitt commented that `PyEval_InitThreads` is not necessary which I think is not true because `PyEval_InitThreads` is required to be called before Python3.7 (https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads).

After manually calling `PyEval_InitThreads` in `#[pymodule]`, pyo3 no longer panics.
https://github.com/tikv/client-py/blob/6225c2ffcafbbed70dde3cf9efe3460a5e34d84e/src/lib.rs#L14-L16

### 🌍 Environment

 - Your operating system and version: Ubuntu 2004
 - Your python version: 3.6.9
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: apt, yes
 - Your Rust version (`rustc --version`): rustc 1.53.0-nightly (16bf626a3 2021-04-14)
 - Your PyO3 version: 0.14.4
 - Have you tried using latest PyO3 main (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: not yet 

### 💥 Reproducing

Not minimized the project yet but I've found the root cause.

davidhewitt:
Fixes #1870

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -52,10 +52,12 @@ pub(crate) fn gil_is_acquired() -> bool {
 /// software). Support for this is tracked on the
 /// [PyPy issue tracker](https://foss.heptapod.net/pypy/pypy/-/issues/3286).
 ///
+/// Python 3.6 only: If the Python interpreter is initialized but Python threading is not, this
+/// function will initialize Python threading.
+///
 /// # Panics
-/// - If the Python interpreter is initialized but Python threading is not, a panic occurs.
-///   It is not possible to safely access the Python runtime unless the main thread (the thread
-///   which originally initialized Python) also initializes threading.
+/// - Python 3.6 only: If this function needs to initialize Python threading but the calling thread
+///   is not the thread which initialized Python, this function will panic.
 ///
 /// # Examples
 /// ```rust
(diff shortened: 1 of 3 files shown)


src/gil.rs
```diff
@@ -52,10 +52,12 @@ pub(crate) fn gil_is_acquired() -> bool {
 /// software). Support for this is tracked on the
 /// [PyPy issue tracker](https://foss.heptapod.net/pypy/pypy/-/issues/3286).
 ///
+/// Python 3.6 only: If the Python interpreter is initialized but Python threading is not, this
```
andylokandy, line 55: Is it only for python3.6? I can't see that python36 is different from the previous versions from the FFI doc. But I haven't tested on python35 or earlier though.

davidhewitt: We don't support Python 3.5 or earlier, as their end of life date has already been reached.

Note also that we will stop supporting Python 3.6 next year: https://github.com/PyO3/pyo3/issues/1782

andylokandy: Got it. Thank you!

PR #1880, opened 2021-09-21
build-config: Always pass `abi3` flag to `default_lib_name_windows()`

ravenexp:
Hardcoding `false` here leads to an incorrect abi3 DLL name
when cross-compiling to Windows.


davidhewitt (approved):
Thanks, good catch! I'll push a CHANGELOG entry onto your branch and then merge.

Issue #1781, opened 2021-08-11
Preparation for Python 3.10 release


davidhewitt:
Python 3.10 RC1 has just been released, so I think we should be looking at testing full compatibility with it. We already run it in CI, however I think there's potentially missing FFI definitions and other interesting changes.

This is a great opportunity for us to ensure everything works before upstream commits to a final release. 

There's two sections in the Python 3.10 changelog which are of note to us:

https://docs.python.org/3.10/whatsnew/3.10.html#changes-in-the-c-api
https://docs.python.org/3.10/whatsnew/3.10.html#c-api-changes

I would also propose we adopt PEP 604 union syntax `X | Y` in the guide and any error messages where we currently mention `Union`.

Finally, I'd love to find some time to experiment with PEP 634 match syntax (#1714), however this is probably a bigger feature that we might want to take some time to explore.

Anyone interested in helping out with this is welcome. Just ping here what you're working on so that we don't end up duplicating work.

davidhewitt:
We also need to add an `abi3_py310` feature.

mejrs:
I skimmed it and I think these are the ones we care about?

> Add PyUnicode_AsUTF8AndSize() to the limited C API. (Contributed by Alex Gaynor in bpo-41784.)

We can use this at https://pyo3.rs/main/doc/pyo3/types/struct.PyString.html#method.to_str

> Add a new orig_argv member to the PyConfig structure: the list of the original command line arguments passed to the Python executable. (Contributed by Victor Stinner in bpo-23427.)

It needs that field added.

> Add the Py_Is(x, y) function to test if the x object is the y object, the same as x is y in Python. Add also the Py_IsNone(), Py_IsTrue(), Py_IsFalse() functions to test if an object is, respectively, the None singleton, the True singleton or the False singleton. (Contributed by Victor Stinner in bpo-43753.)

Should we care about this in e.g. `PyAny::is_none`? Looks like all those functions do (for now) is forward to a pointer comparison.

Other than that a few functions need implementing in the ffi module, I think.

davidhewitt:
👍 thanks, I'll put together a PR for the ffi changes shortly.

> It needs that field added.

I think it's already there: https://github.com/PyO3/pyo3/blob/4a34cc63a67d14397b30d5cc064742a0848a556d/src/ffi/cpython/initconfig.rs#L104-L105

> Should we care about this in e.g. `PyAny::is_none`? Looks like all those functions do (for now) is forward to a pointer comparison.

[...]

PR #1885, opened 2021-09-24
pep 604: use `T | U` instead of `Union[T, U]` in messaging

davidhewitt:
See #1781 

I think `T | U` is nicer syntax than `Union[T, U]`. Also, we only use it in error messages, so I think we can use this syntax on all Python versions (not just 3.10+).


birkenfeld (approved):
I like it.

PR #1888, opened 2021-09-25
Remove use of unwrap

momirza:
Propagates error back instead of panicking. We had an issue in production when a worker in the process of shutting down returned `SystemExit` when calling `inner.instance(py).str()` on the Python object. This should be propagated back instead of resulting in a `PanicException`. 

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'SystemExit'>, value: SystemExit(1), traceback: Some(<traceback object at 0x7fe1bbcacd70>) }'
```


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Move Py_DecodeLocale from sysmodule to fileutils. [#1887](https://github.com/PyO3/pyo3/pull/1887)
 - Deprecate `PySys_AddWarnOption`, `PySys_AddWarnOptionUnicode` and `PySys_HasWarnOptions`. [#1887](https://github.com/PyO3/pyo3/pull/1887)
 - Remove function PyTuple_ClearFreeList from python 3.9 above. [#1887](https://github.com/PyO3/pyo3/pull/1887)
+- Replace `unwrap` with `?` in derivation body for enums. [#1888](https://github.com/PyO3/pyo3/pull/1888)
```
davidhewitt, line 38: The user-facing experience is probably "Fixed: panic inside generated `#[derive(FromPyObject)]` for enums"

- I'll push a commit to tweak this shortly and then merge!

davidhewitt (approved):
Thanks for fixing, sorry for letting this panic slip through review 😬

I'm going to adjust the CHANGELOG message - I prefer them to be more informative to how the typical user experience will change rather than the actual code change we made.

I'll push a commit to tweak and then merge.

Issue #1714, opened 2021-07-04
Support for PEP 622 structural pattern matching


fschutt:
Just wanted to ask: New versions of Python have added support for pattern matching ([PEP 622](https://www.python.org/dev/peps/pep-0622/#reference-implementation)):

```python
def make_point_3d(pt):
    match pt:
        case (x, y):
            return Point3d(x, y, 0)
        case (x, y, z):
            return Point3d(x, y, z)
        case Point2d(x, y):
            return Point3d(x, y, 0)
        case Point3d(_, _, _):
            return pt
        case _:
            raise TypeError("not a point we support")
```

Would it be possible to extend Pyo3 to support `#[derive(FromPyObject)]` for Rust enums, so that Rust can work natively with these Python enums? According to the specification, all that needs to be done is that the Python interpreter tries to look for new functions

Right now I am emulating this behaviour by having a struct with `#[repr(transparent)]` and a function to create the case:

```rust
#[repr(C, u8)]
enum Point {
    CaseA(PointCaseA),
    CaseB(PointCaseB),
}

#[pyclass]
#[repr(C)]
struct PointCaseA {
    x: f32,
    y: f32,
}

#[pyclass]
#[repr(C)]
struct PointCaseB {
    x: f32,
    y: f32,
    z: f32,
}

#[pyclass]
#[pyo3(name = "Point")]
#[repr(transparent)]
struct PointWrapper {
    _0: Point,
}

#[pymethods]
impl PointWrapper {
    // not the exact code, just pseudocode
    pub fn destructure(&self) -> (&'static str, PyObject) {
        let casted: &Point = unsafe { mem::transmute(&self) };
        match casted {
            CaseA(x, y) => ("CaseA", PyObject::Tuple(&[x, y]),
            CaseB(x, y, z) => ("CaseB", PyObject::Tuple(&[x, y, z]),
        }
    }
}
```

```python
def make_point_3d(pt): # pt = object created by Rust
    tag, payload = pt.destructure()
    if tag == "CaseA":
        return Point3d(payload.x, payload.y, 0)
    elif tag == "CaseB":
        return Point3d(payload.x, payload.y, payload.z)
    else:
        raise TypeError("not a point we support")
```

It's not a very high priority, but it would be nice to have. Just wanted to ask if the PyO3 team is aware of this PEP.

birkenfeld:
This is definitely interesting, but:

* I'd let Python's `match` conventions shake out themselves for at least a release
[...]

PR #1889, opened 2021-09-26
packaging: formal support for Python 3.10

### LINKED ISSUE #1781: Preparation for Python 3.10 release

davidhewitt:
This PR goes through the FFI changes listed in #1781, and also adds an `abi3-py310` feature.

Given that we have also tested Python 3.10 on CI for a long time, and this PR adds the documented C API changes, I think we can call Python 3.10 officially supported now.

I also made a tidy-up pass to use `_Py_NewRef` in expressions in a bunch of places, because imo it's nicer than calling `Py_INCREF` in a separate satement. And we can define `_Py_NewRef` on all supported Python versions (not just 3.10).

Closes #1781 

(The PEP 622 support is tracked separately in #1714 - I don't think we'll have this ready by the time 3.10 is released.)


CHANGELOG.md
```diff
@@ -17,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add range indexing implementations of `std::ops::Index` for `PyList`, `PyTuple` and `PySequence`. [#1829](https://github.com/PyO3/pyo3/pull/1829)
 - Add commonly-used sequence methods to `PyList` and `PyTuple`. [#1849](https://github.com/PyO3/pyo3/pull/1849)
 - Add `as_sequence` methods to `PyList` and `PyTuple`. [#1860](https://github.com/PyO3/pyo3/pull/1860)
+- Add `abi3-py310` feature. [#1885](https://github.com/PyO3/pyo3/pull/1885)
```
davidhewitt, line 24:
```suggestion
- Add `abi3-py310` feature. [#1889](https://github.com/PyO3/pyo3/pull/1889)
```

src/ffi/modsupport.rs
```diff
@@ -53,7 +53,13 @@ extern "C" {
     // skipped non-limited _PyArg_UnpackKeywords
     // skipped non-limited _PyArg_Fini
 
-    // skipped PyModule_AddObjectRef
+    #[cfg(Py_3_10)]
+    #[cfg_attr(docsrs, doc(cfg(Py_3_10)))]
+    pub fn PyModule_AddObjectRef(
+        arg1: *mut PyObject,
+        arg2: *const c_char,
+        arg3: *mut PyObject,
```
mejrs, line 61:
```suggestion
        module: *mut PyObject,
        name: *const c_char,
        value: *mut PyObject,
```

That's the names these have in the documentation

src/ffi/object.rs
```diff
@@ -494,7 +530,13 @@ pub const Py_NE: c_int = 3;
 pub const Py_GT: c_int = 4;
 pub const Py_GE: c_int = 5;
 
-// skipped non-limited / 3.10 PySendResult
+#[cfg(Py_3_10)]
+#[repr(C)]
```
mejrs, line 534:
```suggestion
#[repr(C)]
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```

mejrs (changes requested):
Thanks!..just two nits

mejrs (approved):

PR #1893, opened 2021-09-29
Implement `eyre` feature

mejrs:
Per https://github.com/yaahc/eyre/issues/57

image
image
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ links = "python"
 [dependencies]
 cfg-if = { version = "1.0" }
 # must stay at 0.3.x for Rust 1.41 compatibility
+eyre = {version = ">= 0.4" , optional = true}
 indoc = { version = "0.3.6", optional = true }
 inventory = { version = "0.1.4", optional = true }
 libc = "0.2.62"
--- a/src/conversions/eyre.rs
+++ b/src/conversions/eyre.rs
@@ -1,0 +1,158 @@
+#![cfg(feature = "eyre")]
+#![cfg_attr(docsrs, doc(cfg(feature = "eyre")))]
+//! A conversion from [eyre]’s [`Report`] type to [`PyErr`].
+//!
+//! Use of an error handling library like [eyre] is common in application code and when you just
+//! want error handling to be easy. If you are writing a library or you need more control over your
+//! errors you might want to design your own error type instead.
+//!
+//! This implementation always creates a Python [`RuntimeError`]. You might find that you need to
+//! map the error from your Rust code into another Python exception. See [`PyErr::new`] for more
+//! information about that.
+//!
+//! For information about error handling in general, see the [Error handling] chapter of the Rust
+//! book.
+//!
+//! # Setup
+//!
+//! To use this feature, add this to your **`Cargo.toml`**:
+//!
+//! ```toml
+//! [dependencies]
+//! ## change * to the version you want to use, ideally the latest.
+//! eyre = "*"
+// workaround for `extended_key_value_attributes`: https://github.com/rust-lang/rust/issues/82768#issuecomment-803935643
+#![cfg_attr(docsrs, cfg_attr(docsrs, doc = concat!("pyo3 = { version = \"", env!("CARGO_PKG_VERSION"),  "\", features = [\"eyre\"] }")))]
+#![cfg_attr(not(docsrs), doc = "pyo3 = { version = \"*\", features = [\"eyre\"] }")]
+//! ```
+//!
+//! Note that you must use compatible versions of eyre and PyO3.
+//! The required eyre version may vary based on the version of PyO3.
+//!
+//! # Example: Propagating a `PyErr` into [`eyre::Report`]
+//!
+//! ```rust
+//! use pyo3::prelude::*;
+//! use pyo3::wrap_pyfunction;
+//! use std::path::PathBuf;
+//!
+//! // An example function that returns `eyre::Result<...>`.
+//! fn rust_open(filename: PathBuf) -> eyre::Result<Vec<u8>> {
+//!     let data = std::fs::read(filename)?;
+//!     Ok(data)
+//! }
+//!
+//! // A wrapper around a Rust function.
+//! #[pyfunction]
+//! fn py_open(filename: PathBuf) -> PyResult<Vec<u8>> {
+//!     // The `?` ("try") operator performs the conversion
+//!     // into a `PyErr` - if it is an error.
+//!     let data = rust_open(filename)?;
+//!     Ok(data)
+//! }
+//!
+//! fn main() {
+//!     let error = Python::with_gil(|py| -> PyResult<Vec<u8>> {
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -17,6 +17,7 @@ links = "python"
 [dependencies]
 cfg-if = { version = "1.0" }
 # must stay at 0.3.x for Rust 1.41 compatibility
+eyre = {version = ">= 0.4" , optional = true}
```
davidhewitt, line 20: Think these need to be swapped (new dep has split comment from its target). Also, should we place an upper bound?

```suggestion
eyre = {version = ">= 0.4, <0.5" , optional = true}
# must stay at 0.3.x for Rust 1.41 compatibility
```

src/conversions/eyre.rs
```diff
@@ -0,0 +1,158 @@
...
+//! }
+//!
+//! // A wrapper around a Rust function.
+//! #[pyfunction]
+//! fn py_open(filename: PathBuf) -> PyResult<Vec<u8>> {
+//!     // The `?` ("try") operator performs the conversion
+//!     // into a `PyErr` - if it is an error.
+//!     let data = rust_open(filename)?;
+//!     Ok(data)
+//! }
```
davidhewitt, line 52: I think it should be possible to also just have a `#[pyfunction]` which returns `eyre::Result<Vec<u8>>` - which might be useful to have as an example (as a kind of test)?

davidhewitt (approved):
Thanks, looks great! Should we merge this at the same time as #1822 ?

davidhewitt:
Oh, also this is definitely worth a CHANGELOG line :)

Cargo.toml
mejrs: (oops)

Eyre is at 0.6.5 now, so that upper bound is too strict. 0.4 is the first version that has the eyre::Report type (?), so I think any version above that should be fine.

I'll admit that I don't know anything about how cargo resolves dependency versions, but this looks like how you should say "works with any eyre version of 0.4 and up". Is that too liberal and should it be pinned to `">= 0.4, <0.7"` instead?

src/conversions/eyre.rs
mejrs: I thought about doing that but I want to point at the where the conversion happens. A `#[pyfunction]` returning `eyre::Result<Vec<u8>>` has that conversion hidden somewhere in the `pyfunction` macro.

CHANGELOG.md
```diff
@@ -22,8 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `PyMapping` type to represent the Python mapping protocol. [#1844](https://github.com/PyO3/pyo3/pull/1844)
 - Add commonly-used sequence methods to `PyList` and `PyTuple`. [#1849](https://github.com/PyO3/pyo3/pull/1849)
 - Add `as_sequence` methods to `PyList` and `PyTuple`. [#1860](https://github.com/PyO3/pyo3/pull/1860)
+- Add `eyre` feature to convert `eyre::report` into `PyErr`. [#1893](https://github.com/PyO3/pyo3/pull/1893)
 - Add `abi3-py310` feature. [#1889](https://github.com/PyO3/pyo3/pull/1889)
 
+
```
davidhewitt, line 28: Bonus newline!

```suggestion
```

CHANGELOG.md
```diff
@@ -22,8 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `PyMapping` type to represent the Python mapping protocol. [#1844](https://github.com/PyO3/pyo3/pull/1844)
 - Add commonly-used sequence methods to `PyList` and `PyTuple`. [#1849](https://github.com/PyO3/pyo3/pull/1849)
 - Add `as_sequence` methods to `PyList` and `PyTuple`. [#1860](https://github.com/PyO3/pyo3/pull/1860)
+- Add `eyre` feature to convert `eyre::report` into `PyErr`. [#1893](https://github.com/PyO3/pyo3/pull/1893)
```
davidhewitt, line 25:
```suggestion
- Add `eyre` feature to convert `eyre::Report` into `PyErr`. [#1893](https://github.com/PyO3/pyo3/pull/1893)
```

Also for dependencies I quite like to have something in the  `Packaging` section along the lines of "added optional `eyre` dependency" - what do you think of having that as well (or instead)?

Cargo.toml
davidhewitt: I think that by default `cargo` will pick the highest `eyre` possible for the pyo3 dependency, not the one which the user already has in their own Cargo.toml. (This is a bit awkward.) The user can adjust it manually using some incantation of `cargo update --precise` (like we do in our MSRV CI jobs).

> Is that too liberal and should it be pinned to `">= 0.4, <0.7"` instead?

Generally we've only offered a range for dependencies when the MSRV support requires an older version. I'd be tempted to ask the other question - is just supporting `eyre = "0.6"` acceptable? 

As for the upper bound: if we were to take semver literally, eyre `0.7` could hypothetically be very breaking and completely remove the `eyre::Report` type. (Although I would be suprised if this happened in practice.) So I'd definitely prefer have an upper bound of `< 0.7` if we go for a range. Dependabot can remind us to update when 0.7 releases 😄

src/conversions/eyre.rs
```diff
@@ -0,0 +1,158 @@
...
+//!
+//! fn main() -> eyre::Result<()> {
+//!     let bytes: &[u8] = b"x\x9c\x8b\xcc/U(\xce\xc8/\xcdIQ((\xcaOJL\xca\xa9T\
+//!                         (-NU(\xc9HU\xc8\xc9LJ\xcbI,IUH.\x02\x91\x99y\xc5%\
+//!                         \xa9\x89)z\x00\xf2\x15\x12\xfe";
+//!     let text = decompress(bytes)?;
+//!
+//!     println!("The text is \"{}\"", text);
+//! # assert_eq!(text, "You should probably use the libflate crate instead.");
```
davidhewitt, line 103: 😂

src/conversions/eyre.rs
davidhewitt: That's a fair point. I wonder if your second example "using `eyre` in general" could be tweaked slightly so that the function `decompress` is a `pyfunction` returning `eyre::Result`? The conversion from `PyErr` -> `eyre::Result` can still be there.

Alternatively maybe it's better to keep away from that here, and instead in the guide's `function.md` there is probably scope for an improved "error handling" section which comments on how to use `Result` with error types that convert to `PyErr`. (And can use `Result<_, eyre::Report>` as an example?)

davidhewitt (commented):
Thanks, all looks really good. Sorry for the delay - I was ill over the weekend. Feeling a bit more capable this morning finally!

Just commented on the remaining points with some thoughts. All is great as-is though so also happy if you want to call this done for now and merge it!

mejrs:
> at the same time as finishing up #1822?

I don't see a reason to let this sit around here. I think we'd still want to have them both in 0.15 though.

Issue #804, opened 2020-03-12
Wasm-bindgen-like Closure wrapper around Rust functions?


davidhewitt:
As noted in this [part of the book](https://github.com/PyO3/pyo3/blob/master/guide/src/function.md#calling-rust-fns-in-python). I didn't see any issue documenting this intention.

davidhewitt:
As far as I'm concerned this was implemented in #1901, can close this one.

PR #1901, opened 2021-10-03
Support for wrapping rust closures as python functions

LaurentMazare:
Hello and thanks for all the work on this very useful crate!
My understanding is that there is currently no easy way to wrap Rust closures and call them from Python. This PR adds a new function `PyCFunction::new_closure` to help doing so. Note that I'm not very familiar with the pyo3 internals, so it's unclear to me whether this is the right api for exposing this, any feedback is very welcome, and maybe this should be discussed in a 'need-design' issue first (though I haven't found any related such issue).

Technically, this works by boxing the Rust closure twice (to get around the fat pointer size) and the resulting pointer is stored in a Python capsule. Then `PyCFunction_NewEx` is used to create a Python callable function, this uses a 'dispatch' function `run_closure` that unpacks the capsule and calls the closure. The capsule also registers a destructor to collect the closure when it's not used anymore.
There are two small tests, one of them where the closure mutates some Rust based state.

Current limitations include:
- The wrapped closure gets both the positional and keyword arguments, it would be possible to also have a function that only handles positional arguments or even no argument.
- There is a `RefUnwindSafe` constraint that is used for `handle_panic`. I'm not sure if it's possible to remove it somehow, if this constraint is necessary it should probably be added to the `new_closure` function too.
- The function name and documentation are hard-coded, maybe having a version of `PyMethodDef` using owned string instead of `&'static str` could help get around this and allow the user to specify the function name and documentation.
Again happy to get feedback on any of these or other limitations I have missed (or feel free to edit the PR too).


mejrs (commented):
Thanks 👍 I really want an api such as this. Thank you for doing this 🙏🏻

However there are a couple of issues with this implementation.

1) For starters, you can do use-after-frees by transferring references into the closure:

```rust
use pyo3::prelude::*;
use pyo3::types::{PyCFunction, PyDict, PyTuple};

fn main() {
    let fun = Python::with_gil(|py| {
        let stuff = vec![0, 1, 2, 3, 4];
        let ref_: &[u8] = &stuff;

        let counter_fn = |_args: &PyTuple, _kwargs: Option<&PyDict>| {
            println!("called");
            println!("This is five: {:?}", ref_.len());

            Ok("".into_py(py))
        };

        let counter_py: Py<PyCFunction> = PyCFunction::new_closure(counter_fn, py).unwrap().into();

        counter_py
    });

    Python::with_gil(|py| {
        fun.call0(py).unwrap();
    });
}
```
```text
This is five: 738180135880
```

Similarly you can probably (I haven't checked) use `Py<PyCFunction>` to smuggle non-`Send` types to other threads. So you'll want (at least) a `Send + 'static` bound.

2) The `Box<dyn Fn(&types::PyTuple, Option<&types::PyDict>) -> PyResult<PyObject>>` bound is quite rigid. It would be nice to have closures that don't need args/kwargs arguments or need to return a PyResult. Having to return `Ok(py.None())` is going to get annoying. Maybe we can do something clever with traits here?

In the past I've used a pyclass with a call method for this, like so:
```rust
use pyo3::prelude::*;
use pyo3::types::{PyDict, PyTuple};


type C = Box<dyn Fn(&PyTuple, Option<&PyDict>) -> PyResult<()> + Send + 'static>;

#[pyclass]
struct Closure {
    wraps: C,
}

#[pymethods]
impl Closure {
    #[call]
    #[args(args = "*", kwargs = "**")]
    fn __call__(
        &self,
        args: &PyTuple,
        kwargs: Option<&PyDict>,
    ) -> PyResult<()> {
        (self.wraps)(args, kwargs)
    }
}

fn main() {
    Python::with_gil(|py| {
        let f = |_: &PyTuple, _: Option<&PyDict>| {
            println!("called");
            Ok(())
        };

        let closure = Closure{wraps: Box::new(f)};
		closure.__call__(PyTuple::empty(py), None).unwrap();
		closure.__call__(PyTuple::empty(py), None).unwrap();
		closure.__call__(PyTuple::empty(py), None).unwrap();
    });
}
```

It'll be nice to have a more idiomatic interface for that.

src/types/function.rs
```diff
@@ -39,6 +76,38 @@ impl PyCFunction {
         )
     }
 
+    /// Create a new function from a closure.
+    pub fn new_closure<F>(f: F, py: Python) -> PyResult<&PyCFunction>
+    where
+        F: Fn(&types::PyTuple, Option<&types::PyDict>) -> PyResult<PyObject>,
```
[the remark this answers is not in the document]
LaurentMazare: Added the bound, thanks!

src/types/function.rs
```diff
@@ -3,15 +3,52 @@ use crate::exceptions::PyValueError;
...
+        boxed_fn(args, kwargs)
+    })
+}
+
+unsafe extern "C" fn drop_closure(capsule_ptr: *mut ffi::PyObject) {
+    let boxed_fn: Box<BoxedFn> = Box::from_raw(ffi::PyCapsule_GetPointer(
+        capsule_ptr,
+        closure_capsule_name(),
+    ) as *mut Box<_>);
+    drop(boxed_fn)
```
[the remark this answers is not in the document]
LaurentMazare: Good point, I'm not sure what the best way is to handle this though: the [PyCapsule documentation](https://docs.python.org/3/c-api/capsule.html) does not mention any error/exception mechanism for the destructor. We could just silently swallow the panic but I imagine that there is something better to do than this?

src/types/function.rs
```diff
@@ -3,15 +3,52 @@ use crate::exceptions::PyValueError;
...
 
 /// Represents a builtin Python function object.
 #[repr(transparent)]
 pub struct PyCFunction(PyAny);
 
 pyobject_native_type_core!(PyCFunction, ffi::PyCFunction_Type, #checkfunction=ffi::PyCFunction_Check);
 
+const CLOSURE_CAPSULE_NAME: &[u8] = b"pyo3-closure\0";
+
+fn closure_capsule_name() -> *const i8 {
```
birkenfeld, line 18: Why is this a `fn`? Also, `char` is either `i8` or `u8` depending on platform, so it's easiest to use

`CLOSURE_CAPSULE_NAME.as_ptr() as *const _`

inline, see e.g. https://github.com/PyO3/pyo3/blob/main/src/derive_utils.rs#L311

LaurentMazare:
Thanks for the feedback and the detailed example showing why `Send + 'static` are required, this is very helpful.

Re making it more idiomatic having some helper functions for the no-args or no-returns case would be an easy thing to do. Alternatively a dedicated macro similar to `#[pyfunction]/wrap_pyfunction!` could be helpful there though it would add quite some complexity.
For the return type of the closure, maybe just requiring it to be `IntoPy` would do the trick, though not sure that there could be an equivalent for the closure arguments.

```rust
      pub fn new_closure<F, R>(f: F, py: Python) -> PyResult<&PyCFunction>
      where
          F: Fn(&types::PyTuple, Option<&types::PyDict>) -> PyResult<R> + Send + 'static,
          R: IntoPy<PyObject>,
      {
```

src/types/function.rs
```diff
@@ -39,6 +76,38 @@ impl PyCFunction {
...
     }
 
+    /// Create a new function from a closure.
+    pub fn new_closure<F>(f: F, py: Python) -> PyResult<&PyCFunction>
+    where
+        F: Fn(&types::PyTuple, Option<&types::PyDict>) -> PyResult<PyObject>,
+    {
+        let boxed_fn: Box<dyn Fn(&types::PyTuple, Option<&types::PyDict>) -> PyResult<PyObject>> =
+            Box::new(f);
+        let function_ptr = Box::into_raw(Box::new(boxed_fn));
```
birkenfeld, line 86: Please add a comment here why the function is boxed twice.

src/types/function.rs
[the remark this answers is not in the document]
birkenfeld: Not much we can do - swallow or print. Maybe emit as a Python warning?

src/types/function.rs
```diff
@@ -39,6 +76,38 @@ impl PyCFunction {
...
+        let capsule_ptr = unsafe {
+            ffi::PyCapsule_New(
+                function_ptr as *mut c_void,
+                closure_capsule_name(),
+                Some(drop_closure),
+            )
+        };
+        let method_def = methods::PyMethodDef::cfunction_with_keywords(
+            "pyo3_closure",
```
birkenfeld, line 95: Would it be useful to allow this to be passed as an argument (still `&'static str` of course)?

src/types/function.rs
```diff
@@ -39,6 +76,38 @@ impl PyCFunction {
...
+            methods::PyCFunctionWithKeywords(run_closure),
+            "",
+        );
+        let def = method_def
```
birkenfeld, line 99: I would prefer a new variant for `PyFunctionArguments` and using `new_internal` instead of copying its code here.

src/types/function.rs
davidhewitt: If we used `Box<F>` and made `run_closure` generic on `F` too we might be able to avoid double-boxing at the cost of additional code size. Potentially worth it for the performance improvement?

davidhewitt:
Thanks, I've wanted this for a long time too! There's a very old issue #804. I don't know if we want to take any inspiration from `wasm_bindgen` as to how we want to implement this?

> Alternatively a dedicated macro similar to `#[pyfunction]/wrap_pyfunction!` could be helpful there though it would add quite some complexity.

I agree that if we wanted to support varied argument types, it would be best to have a proc macro. `py_closure!` perhaps? We would be able to piggy-back off a lot of the `pyfunction` codegen, although might force some refactoring.

> For the return type of the closure, maybe just requiring it to be `IntoPy` would do the trick

I think `IntoPyCallbackOutput<*mut ffi::PyObject>` is probably the best trait to use for the return value. It'll allow anything which implements `IntoPy<PyObject>`, whether wrapped in a result or not, as well as `()`. 

(This is essentially what `#[pyfunction]` uses.)

src/types/function.rs
LaurentMazare: Thanks, I've moved to your suggestion, the goal of the function was to avoid repeating the 'as_ptr()' bit but it certainly not a big gain (and might not be very idiomatic).

src/types/function.rs
LaurentMazare: Nice suggestion, indeed using a generic on `F` for `run_closure` removes the need for double boxing or using `dyn` at all. I've just pushed some changes doing this, I imagine that it's at the cost of some slightly larger generated code but avoiding the double indirection seems worth it.

src/types/function.rs
LaurentMazare: I've exposed it, as it may indeed be useful. Longer term, being able to pass an owned string would be even better, but no need to cram everything in here.

src/types/function.rs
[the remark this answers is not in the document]
LaurentMazare: Got it, I've tweaked the code to print to stderr. Not sure how safe using Python function is from this callback so I've kept it to be only Rust printing.

LaurentMazare:
Thanks all for the quick feedback and the reviews. One thing I forgot to mention, mostly for context, is that this wrapping of a `PyCapsule` in `PyCFunction_NewEx` is what we use in [pyml](https://github.com/thierry-martinez/pyml) to expose OCaml closures to the Python runtime.

> Thanks, I've wanted this for a long time too! There's a very old issue #804. I don't know if we want to take any inspiration from `wasm_bindgen` as to how we want to implement this?

Ah sorry I haven't noticed this issue. I didn't know about `wasm_bindgen` `Closure` so I only glimpsed at the [documentation](https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen/closure/struct.Closure.html). My (very basic) understanding is that this brings mostly a way to control the lifetime of the closure on the Rust side and invalidate the wrapping function once it has been dropped. For the use cases that I have at the moment I would have thought that tying the lifetime to the Python function still being reachable is somewhat more natural, but maybe there are other use cases where being able to drop the function from the Rust side is helpful?

> I agree that if we wanted to support varied argument types, it would be best to have a proc macro. `py_closure!` perhaps? We would be able to piggy-back off a lot of the `pyfunction` codegen, although might force some refactoring.

Makes sense, this seems like a bit of a more involved change so maybe would it be ok to defer it to another PR?
 
> I think `IntoPyCallbackOutput<*mut ffi::PyObject>` is probably the best trait to use for the return value. It'll allow anything which implements `IntoPy<PyObject>`, whether wrapped in a result or not, as well as `()`.

Got it, the PR has been tweaked to use this `IntoPyCallbackOutput<*mut ffi::PyObject>`, and one of the example has been adapted to benefit from this.

src/types/function.rs
LaurentMazare: Right, duplicating the code seems indeed sad, I've made some small changes to move the duplicated bits in a new function. This seemed slightly simpler than adding a variant to `PyFunctionArguments`, especially now that we have generics on `F` at a couple places (my understanding is that we should avoid having a variant case involving pointers in `PyFunctionArguments` but I'm very unsure about this).

src/types/function.rs
mejrs: Hm, do we even want to name anonymous functions? I don't have a strong opinion either way. However given that you can't (ergonomically) create `&'static str` at runtime, you can't easily create names dynamically at runtime.

mejrs:
I think it would be better if users can omit the type annotations in the closure arguments - like below. This doesn't compile now:

```rust
fn main() {
    Python::with_gil(|py| {
        let counter_fn = |_, _| {
            println!("called");
            Ok(())
        };
	
        let counter_py: Py<PyCFunction> = PyCFunction::new_closure("", counter_fn, py).unwrap().into();
    });
}
```

```text
error: implementation of `FnOnce` is not general enough
  --> src\main.rs:11:43
   |
11 |         let counter_py: Py<PyCFunction> = PyCFunction::new_closure("", counter_fn, py).unwrap().into();
   |                                           ^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `FnOnce` is not general enough
   |
   = note: closure with signature `fn(&'2 PyTuple, Option<&PyDict>) -> Result<(), PyErr>` must implement `FnOnce<(&'1 PyTuple, Option<&PyDict>)>`, for any lifetime `'1`...
   = note: ...but it actually implements `FnOnce<(&'2 PyTuple, Option<&PyDict>)>`, for some specific lifetime `'2`

error: implementation of `FnOnce` is not general enough
  --> src\main.rs:11:43
   |
11 |         let counter_py: Py<PyCFunction> = PyCFunction::new_closure("", counter_fn, py).unwrap().into();
   |                                           ^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `FnOnce` is not general enough
   |
   = note: closure with signature `fn(&PyTuple, Option<&'2 PyDict>) -> Result<(), PyErr>` must implement `FnOnce<(&PyTuple, Option<&'1 PyDict>)>`, for any lifetime `'1`...
   = note: ...but it actually implements `FnOnce<(&PyTuple, Option<&'2 PyDict>)>`, for some specific lifetime `'2`
```

Perhaps that can actually work. However, this requires higher rank trait bounds (?) and I couldn't get it to work.

src/types/function.rs
LaurentMazare: I don't have a strong opinion here, the names appear when printing the function object itself, e.g.:
```
<built-in method my-function-name of PyCapsule object at 0x7fb400b55660>
```
But not sure whether that makes it worth it to expose to the end user. @birkenfeld any thoughts there?

birkenfeld: How does it look in tracebacks? That's where I was initially thinking it might be useful.

LaurentMazare:
> I think it would be better if users can omit the type annotations in the closure arguments - like below.

Agreed that it would be better, though maybe if we have a `pyclosure!` macro this is less important? The type annotations can be reduced so as just to mention the references but it's not as nice as removing them entirely.
```rust
      let counter_fn = move |_: &_, _: Option<&_>| {
          Ok(())
      };
```

src/types/function.rs
LaurentMazare: I actually don't see the function name (or the closure at all) appearing in the traceback though I might be doing something wrong here.
```rust
      let counter_fn = move |_args: &types::PyTuple, _kwargs: Option<&types::PyDict>| {
          let py = gil.python();
          py.run("raise Exception('banana')", None, None)?;
          Ok(())
      };
```
This prints the following traceback:
```python
import traceback
try:
  counter_py()
except Exception:
  print(traceback.format_exc())

>> Traceback (most recent call last):
>>  File "<string>", line 3, in <module>
>>  File "<string>", line 1, in <module>
>> Exception: banana
```

birkenfeld: ok, then I guess the name is not useful.

LaurentMazare: Ok, I've switched it back to being hard-coded then.

davidhewitt:
I've just rebased and added a CHANGELOG entry, will proceed to merge this! Thanks again!

Issue #954, opened 2020-06-05
Guide entry for PyBuffer


davidhewitt:
I can't see `PyBuffer` mentioned anywhere in the guide...

At the very least, we might want to add it to the [argument types](https://pyo3.rs/master/conversions.html#argument-types) section of `conversions.md`? Or maybe on [`types.md`](https://pyo3.rs/master/types.html).

PR #1905, opened 2021-10-09
Issue/#954 Add PyBuffer to docs

Askaholic:
Some things I'm not sure about:
- Why are there two types, `PyBuffer` and `PyBufferProtocol`?
- Is it intentional that `PyBuffer` can be used as a function argument for pyfunctions but `&PyBuffer` cannot? I guess `PyBuffer` is just a pointer type, but it confused me for a while since all the other types on that list can be taken as references.
- Why would anyone want to call `PyBuffer::release` explicitly and not just use scopes or `std::mem::drop`?

#954

### DIFF
--- a/guide/src/conversions/tables.md
+++ b/guide/src/conversions/tables.md
@@ -14,7 +14,7 @@ The table below contains the Python type and the corresponding function argument
 | ------------- |:-------------------------------:|:--------------------:|
 | `object`      | -                               | `&PyAny`             |
 | `str`         | `String`, `Cow<str>`, `&str`, `OsString`, `PathBuf` | `&PyUnicode` |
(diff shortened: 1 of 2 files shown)


mejrs:
If you happen to be already holding the gil then it'll check a thread_local cell and see that the gil is already held.

I would rather not make statements like "fast", "slow", "cheap", etc, as these are all relative and you might not notice anyway. - but the implementation of `thread_local!` leaves some things to be desired. For example, read https://matklad.github.io/2020/10/03/fast-thread-locals-in-rust.html

Still, "check nothing" is always going to be faster than "fast".


(minor nit - you'd actually write this:)
```rust
Python::with_gil(|py| {
    buf1.release(py);
    buf2.release(py);
    buf3.release(py);
    buf4.release(py);
});
```

davidhewitt:
Closing in favour of #2593

PR #1906, opened 2021-10-09
chore: pin trybuild version

mejrs:
Pins trybuild to a (latest) specific version. In particular, versions 1.0.45 and 1.0.49 produce different outputs.


mejrs:
PyPy on MacOs is failing. I have no idea how to fix that.

PR #1907, opened 2021-10-12
Fix additional tests after trybuild update

aganders3:
I saw #1906 but when I cloned the repo a number of tests were still failing for me. I believe this covers the recent changes to `trybuild` behavior.

I am not sure about the changes in `tests/ui/invalid_pymethod_receiver.stderr` - these seem to depend on platform, etc.

I am using Rust 1.55.0 in the official Docker image (buster, Python 3.7).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,8 @@ jobs:
             platform: { os: "windows-latest", python-architecture: "x86" }
         include:
           # PyPy3.6 still runs on macos-10.15
-          - python-version: pypy-3.6
+          - rust: stable
+            python-version: pypy-3.6
             platform:
               {
                 os: "macos-10.15",


aganders3:
> When I run the tests for this PR locally the output is just cut off. Is that what you are seeing as well?
> I installed rust-src (with `rustup component add rust-src`) and that problem seemed to go away. I swear I already had that installed though...

Yes, mine was cut off but then I installed `rust-src` and it added the output pointing to `src/convert/mod.rs`, but weirdly mine is missing the line number in the file path as well as in the gutter next to the source, so it looks like this:

```
   --> $RUST/core/src/convert/mod.rs
     |
     |     fn try_from(value: T) -> Result<Self, Self::Error>;
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

aganders3:
> 👍 thanks, this seems correct to me. The tests pass either way for me, so maybe trybuild did something in backcompat? A bit confusing.

Thanks for reviewing! The trybuild stuff is indeed confusing. The changes I made in this PR may actually only have been necessary because I was developing in a container and had the repo mounted at `/pyo3`. With it mounted instead at `/src/pyo3` things pass as-is.

Would you like me to push the CI changes for macos/pypy3.6 to this PR?

I will check out the `setuptools-rust` on my macOS 11 machine - it seems like a promising lead.

aganders3:
Okay I was able to reproduce this hang/issue on my machine with PyPy3.7-v7.3.5. It actually installs successfully but then hangs at the `Processing dependencies...` part. If I cancel and re-run it works fine. Also if I run `python setup.py build` separately from `python setup.py install` it seems to work.

 Reverting to PyPy3.7-v7.3.4 fixes it for me, so this may be a PyPy bug? There were some changes in `distutils.command.install` between these versions:
https://foss.heptapod.net/pypy/pypy/-/compare/release-pypy3.7-v7.3.4...release-pypy3.7-v7.3.5?from_project_id=76

PR #1909, opened 2021-10-12
examples: use pip 21.3

davidhewitt:
Now that pip 21.3 is released, we can update to it and clean up the usage of `--use-feature=in-tree-build`. 

I kinda wonder if this might end up fixing the PyPy 3.7 macOS 11 failure we haven't got to the bottom of yet.


davidhewitt:
> I kinda wonder if this might end up fixing the PyPy 3.7 macOS 11 failure we haven't got to the bottom of yet.

Looks like it did indeed fix it! 🎉

PR #1922, opened 2021-10-15
Pin transitive dev-dependencies to msrv

mejrs:
These two are the last msrv compatible versions, so that `cargo test` works locally without having to use `cargo update precise`


mejrs:
> 
> 
> Good idea! Perhaps remove the now-redundant `cargo update` lines from `ci.yml` at the same time?

Sure

davidhewitt:
Perfect, thanks!

PR #1928, opened 2021-10-17
Remove doc_cfg attributes

mejrs:
with "Make cfg imply doc(cfg)" https://github.com/rust-lang/rust/pull/89596, these attributes are no longer necessary.

One pain point is that now doc_cfg generates info that's sort of...wonky for some things in the ffi module. 

For example...
```rust
#[repr(C)]
#[derive(Copy, Clone, Debug)]
#[cfg(not(PyPy))]
pub struct PyObject {
    #[cfg(py_sys_config = "Py_TRACE_REFS")]
    _ob_next: *mut PyObject,
    #[cfg(py_sys_config = "Py_TRACE_REFS")]
    _ob_prev: *mut PyObject,
    pub ob_refcnt: Py_ssize_t,
    pub ob_type: *mut PyTypeObject,
}

#[repr(C)]
#[derive(Debug, Copy, Clone)]
#[cfg(PyPy)]
pub struct PyObject {
    pub ob_refcnt: Py_ssize_t,
    pub ob_pypy_link: Py_ssize_t,
    pub ob_type: *mut PyTypeObject,
}
```
...looks like this...

image

..and this..

image

I think the best way to fix that is to combine them into a single struct (?). That way it's easy to document the fields as well.

Alternatively there's `doc(cfg(all))` to override that.

I'm personally leaning towards the first option. What do you think?

### DIFF
--- a/src/conversions/num_bigint.rs
+++ b/src/conversions/num_bigint.rs
@@ -80,6 +80,7 @@ unsafe fn extract(ob: &PyLong, buffer: &mut [c_uchar], is_signed: c_int) -> PyRe
 
 macro_rules! bigint_conversion {
     ($rust_ty: ty, $is_signed: expr, $to_bytes: path, $from_bytes: path) => {
+        #[cfg_attr(docsrs, doc(cfg(feature = "num-bigint")))]
         impl ToPyObject for $rust_ty {
             fn to_object(&self, py: Python) -> PyObject {
                 unsafe {
@@ -94,11 +95,15 @@ macro_rules! bigint_conversion {
                 }
             }
         }
+
+        #[cfg_attr(docsrs, doc(cfg(feature = "num-bigint")))]
         impl IntoPy<PyObject> for $rust_ty {
             fn into_py(self, py: Python) -> PyObject {
                 self.to_object(py)
             }
         }
+
+        #[cfg_attr(docsrs, doc(cfg(feature = "num-bigint")))]
         impl<'source> FromPyObject<'source> for $rust_ty {
             fn extract(ob: &'source PyAny) -> PyResult<$rust_ty> {
                 let py = ob.py();
--- a/src/conversions/num_complex.rs
+++ b/src/conversions/num_complex.rs
@@ -117,12 +117,15 @@ impl PyComplex {
 
 macro_rules! complex_conversion {
     ($float: ty) => {
+        #[cfg_attr(docsrs, doc(cfg(feature = "num-complex")))]
         impl ToPyObject for Complex<$float> {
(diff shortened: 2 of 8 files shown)


mejrs:
> I agree that sounds better, because presumably that way the individual fields get their cfg information rendered correctly?

It doesn't look like it. I think this feature is designed for a system where all features are additive. PyO3 is not one of those.

With it condensed in one struct, this is how it renders:

image

I've experimented with this and it seems the conditional compilation happens before the rustdoc pass. So only if those fields are active at the time are the field and the "is supported on..." banner rendered. For example, it *does* render the `ob_pypy_link` field if `--cfg PyPy` is passed to rustdoc. However that also *gets rid* of all the `non(PyPy)` banners elsewhere.

There is a **cursed** workaround for it, of course:

```rust
#[repr(C)]
#[derive(Copy, Clone, Debug)]
pub struct PyObject {
    #[cfg(py_sys_config = "Py_TRACE_REFS")]
    pub _ob_next: *mut PyObject,
    #[cfg(py_sys_config = "Py_TRACE_REFS")]
    pub _ob_prev: *mut PyObject,
    #[cfg(docsrs)]
    #[doc(cfg(py_sys_config = "Py_TRACE_REFS"))]
    pub _ob_next: *mut PyObject,
    #[cfg(docsrs)]
    #[doc(cfg(py_sys_config = "Py_TRACE_REFS"))]
    pub _ob_prev: *mut PyObject,
    pub ob_refcnt: Py_ssize_t,
    #[cfg(PyPy)]
    pub ob_pypy_link: Py_ssize_t,
    #[cfg(docsrs)]
    #[doc(cfg(PyPy))]
    pub ob_pypy_link: Py_ssize_t,
    pub ob_type: *mut PyTypeObject,
}
```

image

However, given that...
  - the feature is nightly
  - it adds some tedious and error prone cfg magic
  - it would be dangerous/confusing if somehow someone uses pyo3 with a `docsrs` cfg active
  
...I'm really not that enthusiastic about doing that. I'm not sure how many things would require covering it like this. I suppose we could do it for a couple of really important things...?

davidhewitt:
Unngh, ok. I think let's not worry about the cfg at field-level for now then? As you say, the feature is nightly so with luck it'll improve (especially if we report these papercuts upstream).

> Oh yeah, it also doesn't work correctly if the cfg is on the module and glob imports are used. [...] So a lot of our ffi module has missing not(Py_LIMITED_API) banners.

Hmm, that's frustrating. I think at least it's no worse that what we currently have? I guess that that's another papercut which hopefully will get fixed in time if we report it.

Issue #1884, opened 2021-09-24
Migrating `#[pyproto]` to `#[pymethods]`


davidhewitt:
This is a tracking issue for the work remaining to complete the support for `__dunder__` (aka "protocol") methods in `#[pymethods]`.

As per the proposal in #1864, the current intention is to have a working but experimental implementation of this in 0.15. If any feedback from this implementation is ok, I'd suggest to deprecate `#[pyproto]` in 0.16 or 0.17. If we're happy with the whole implementation earlier / later we can always adjust these aims.

The items outstanding after #1864 
- [x] Documentation
- [x] Some protocols which need design discussion:
   - [x] The GC protocol - doesn't have a well-specified `__dunder__`  at the Python level, so we'll need to design one
   - [x] The buffer protocol - same thing for this
   - [x] The sequence protocol - it overlaps messily with the mapping protocol; we might want to do the same as what happens at the Python level, or we might want to make `__sequence_getitem__` methods etc to be explicit. See also #1855 
- [x] Suppport for `multiple-pymethods` feature
- [x] tests for
  - [x] `__delete__`
  - [x] `__anext__`
  - [x] `__aiter__`
  - [x] `__index__`
  - [x] `__int__`
  - [x] `__float__`
  - [x] `__invert__`
- [x] ~(maybe) better argument extraction errors (if there's an error, the generated code won't say which argument failed, athough that's probably obvious as most these methods take at most one argument).~
- [x] `__call__`
- [ ] `__new__` / `__init__` ?
- [x] #1927 ?

birkenfeld:
How do we make progress here? Brainstorming for missing protocols?

davidhewitt:
Yes. I have a reasonably complete plan for the sequence protocol which I just need to find time to write up.

GC and Buffer I currently have no better idea than copying the pyproto design exactly. Maybe that's fine, however both those traits are quite clunky in my opinion so I was hoping we can come up with new proposals for them while we're breaking stuff. Ideas very welcome!

At the same time it's probably fine to just copy those two traits in, and then we just might end up wanting to make breaking improvements to them eventually.

davidhewitt:
Just as a note, I'm hoping to get an implementation for the GC protocol done soon (maybe within the next two weeks). That's the last one we need to cover before marking pyproto deprecated and releasing 0.16 🎉

davidhewitt:
[...]

PR #1945, opened 2021-10-24
guide: magic methods in pymethods

davidhewitt:
A first pass at documentation for #1884.

It's a bit sloppy; a fair chunk of it is just listing out the methods that are supported. We can definitely refine this further, however this is at least enough of a structure that users should be able to figure out things if we release the current state as 0.15, and also gives me some space to add documentation for the less-obvious protocols such as gc and buffer integration.


birkenfeld:
How do signatures work for the `pymethod` protocol methods? The docs don't say anything about it.

`__call__` is obviously special, and it already existed in `pymethods` before.  I think there should be another example for a method with a rigid signature.

mejrs:
> I think there should be another example for a method with a rigid signature.

`__len__` has to return an usize (as opposed to any other integer type, or an option/result).

PR #1954, opened 2021-10-30
add two missing feature restrictions for nightly

birkenfeld:
Found with `cargo hack --feature-powerset test`.


davidhewitt (approved):
👍 at some point we might want to consider adding CI for the nightly job?

PR #1955, opened 2021-10-30
Cargo.toml: sort dependencies by type

birkenfeld:
In particular, distinguish optional crates between those
that are necessary for some named feature (e.g. macros),
and those that can be activated as their own feature
(e.g. anyhow).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,23 +15,29 @@ edition = "2018"
 links = "python"
 
 [dependencies]
-cfg-if = { version = "1.0" }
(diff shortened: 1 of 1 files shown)


birkenfeld:
Not sure what's up with the pypy/3.7 builds?

davidhewitt (approved):
Thanks, this is a nice idea.

I think PyPy 3.7 builds are expected to be broken until GHA runners update from the broken PyPy 7.3.6 release 7.3.7.

PR #1958, opened 2021-10-31
macros: clean up protocol argument extraction a bit

davidhewitt:
This PR just tidies up the code which does argument extraction for protocol `#[pymethods]` a little.

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -4,7 +4,9 @@ use std::borrow::Cow;
 
 use crate::attributes::NameAttribute;
 use crate::method::{CallingConvention, ExtractErrorMode};
-use crate::utils::{ensure_not_async_fn, unwrap_ty_group, PythonDoc};
+use crate::utils::{
+    ensure_not_async_fn, remove_lifetime, replace_self, unwrap_ty_group, PythonDoc,
+};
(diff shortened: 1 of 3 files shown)


pyo3-macros-backend/src/pymethod.rs
```diff
@@ -452,55 +454,55 @@ const __FLOAT__: SlotDef = SlotDef::new("Py_nb_float", "unaryfunc");
 const __BOOL__: SlotDef = SlotDef::new("Py_nb_bool", "inquiry").ret_ty(Ty::Int);
 
 const __IADD__: SlotDef = SlotDef::new("Py_nb_inplace_add", "binaryfunc")
-    .arguments(&[Ty::ObjectOrNotImplemented])
+    .arguments(&[Ty::Object])
     .extract_error_mode(ExtractErrorMode::NotImplemented)
     .return_self();
```
davidhewitt, line 459: @birkenfeld: in answer to https://github.com/PyO3/pyo3/pull/1945#issuecomment-955548364

These definitions state what the generated function signature will be and place constraints on what is allowed in the pymethod. For example this `__iadd__` definiion will require there to be only one argument, of any type, and the return type to be `()` or `PyResult<()>`.

If the user deviates from these constraints there should be a compile error, however the error messages can probably be improved.

PR #1962, opened 2021-10-31
err: add `PyErr::fetch_last_error`

davidhewitt:
An alternative to #1957.


davidhewitt:
We decided not to proceed with this alternative.

Issue #538, opened 2019-07-15
Send trait requested for Python::allow_threads


jimy-byerley:
I'm making a library that uses non-thread-safe objects in python-wrapped objects, one of them is a `Rc<glium::Context>` (the opengl context wrapper)
But since the Python::allow_threads requests the `Send` trait, it's not possible to use the `Rc<glium::Context>` while releasing the gil. I read in the pyo3 source code that the `Send` trait was used only to prevent from moving the Py object within the closure. Is possible to use an other marker trait than Send for Python::allow_threads ?

working code:
```rust
#[pyclass]
struct MyData {
    internal:  Arc<Mutex<Rc<Context>>>,    // Rc<Context> doesn't implement `Send`
}

#[pymethods]
impl MyData {
    fn do_something(&mut self, ... ) -> PyResult<()> {
          // do some preparations
          // ...
          self.internal.lock().unwrap().some_methods());
          Ok(())
    }
```
failing code:
```rust
#[pyclass]
struct MyData {
    internal:  Arc<Mutex<Rc<Context>>>,   // Rc<Context> doesn't implement `Send`
}

#[pymethods]
impl MyData {
    fn do_something(&mut self, ... ) -> PyResult<()> {
          // do some preparations
          // ...
          let gil = Python::acquire_gil();
          let py = gil.python();
          py.allow_threads(|| {        
                   // compilation fails because self.internal doesn't implement `Send` as requested by `allow_threads`
                   self.internal.lock().unwrap().some_methods());  
                   });
          Ok(())
    }
```

nagisa:
If making pyo3 sound is an eventual goal, this API must continue to use `Send`.

jimy-byerley:
I'm trying to understand why `Send` is necessary here, as it's a trait for thread-sharable objects and `allow_threads` doesn't execute the closure un a separated thread. I mean: If it's to prevent python to use these data in a bad thread (for example, an openGL data used from out of the graphic thread), it's not safe to allow it in the method's body (in a `#[pyfunction]`). As Python can execute this method from everywhat thread it want.
For situations where using the structure's data in multiple threads is safe, then request `Send` is not necessary because the closure execution takes place in the same thread as the whole method.

If I'm missing something in my interpretation: would it be a question to rust's developpers to make Arc<Mutex<T>> always Send ?

nagisa:
[...]

PR #1963, opened 2021-11-01
Clean up `Python` documentation

mejrs:
Adds/changes some documentation in the python.rs module.

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -15,7 +15,7 @@ use std::os::raw::{c_char, c_int};
 ///
 /// See [Python::version].
 #[derive(Debug)]
-pub struct PythonVersionInfo<'p> {
+pub struct PythonVersionInfo<'py> {
     /// Python major version (e.g. `3`).
     pub major: u8,
     /// Python minor version (e.g. `11`).
@@ -23,14 +23,14 @@ pub struct PythonVersionInfo<'p> {
     /// Python patch version (e.g. `0`).
     pub patch: u8,
     /// Python version suffix, if applicable (e.g. `a0`).
-    pub suffix: Option<&'p str>,
+    pub suffix: Option<&'py str>,
 }
 
-impl<'p> PythonVersionInfo<'p> {
+impl<'py> PythonVersionInfo<'py> {
     /// Parses a hard-coded Python interpreter version string (e.g. 3.9.0a4+).
     ///
     /// Panics if the string is ill-formatted.
-    fn from_str(version_number_str: &'p str) -> Self {
+    fn from_str(version_number_str: &'py str) -> Self {
         fn split_and_parse_number(version_part: &str) -> (u8, Option<&str>) {
             match version_part.find(|c: char| !c.is_ascii_digit()) {
                 None => (version_part.parse().unwrap(), None),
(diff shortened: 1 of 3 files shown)


src/python.rs
```diff
@@ -99,26 +99,42 @@ impl PartialOrd<(u8, u8, u8)> for PythonVersionInfo<'_> {
     }
 }
 
-/// Marker type that indicates that the GIL is currently held.
+/// A marker token that represents holding the GIL.
```
davidhewitt, line 102: Is it worth expanding on GIL for this introduction? Possibly also with a link to some Python docs.

```suggestion
/// A marker token that represents holding the Python interpreter's Global Interpreter Lock (GIL).
```

src/python.rs
```diff
@@ -127,57 +143,63 @@ impl PartialOrd<(u8, u8, u8)> for PythonVersionInfo<'_> {
...
 ///
-/// The [`Python`] type can be used to generate references to variables in
-/// Python's memory e.g. using [`Python::eval()`] and indirectly e.g.
-/// using [`PyModule::import()`], which takes a [`Python`] token
-/// as one if its arguments to prove the GIL is held.  The lifetime of these
-/// references is bound to the GIL (more precisely the [`GILPool`], see
-/// [`Python::new_pool()`]), which can cause surprising results with respect to
-/// when a variable's reference count is decreased so that it can be released to
-/// the Python garbage collector.  For example:
+/// The [`Python`] type can be used to create references to variables owned by the Python
```
davidhewitt, line 151: "owner by the Python interpreter" isn't quite true - the references are owned by PyO3 (but point to Python objects).

src/python.rs
```diff
@@ -189,133 +211,135 @@ impl Python<'_> {
...
+    /// - If multiple [`GILGuard`]s are not dropped in in the reverse order of acquisition, PyO3
+    /// may panic. It is recommended to use [`Python::with_gil`] instead to avoid this.
+    ///
+    /// # Notes
+    ///
+    /// The return type from this function, [`GILGuard`], is implemented as a RAII guard
+    /// around [`PyGILState_Ensure`]. This means that multiple `acquire_gil()` calls are
+    /// allowed, and will not deadlock. However, [`GILGuard`]s must be dropped in the reverse order
+    /// to acquisition. If PyO3 detects this order is not maintained, it may be forced to begin
+    /// a panic.
```
davidhewitt, line 277:
```suggestion
    /// to acquisition. If PyO3 detects this order is not maintained, it will panic when the out-of-order drop occurs.
```

src/python.rs
```diff
@@ -189,133 +211,135 @@ impl Python<'_> {
...
+    /// If you don't need to touch the Python
+    /// interpreter for some time and have other Python threads around, this will let you run
+    /// Rust-only code while letting those other Python threads make progress.
+    ///
+    /// The closure is impermeable to types that are tied to holding the GIL, such as `&`[`PyAny`]
+    /// and its concrete-typed siblings like `&`[`PyString`]. This is achieved via the [`Send`]
+    /// bound on the closure and the return type. This is slightly
+    /// more restrictive than necessary, but it's the most fitting solution available in stable
+    /// Rust. In the future this bound may be relaxed by using an "auto-trait" instead, if
+    /// [auto-traits] ever become a stable feature of the Rust language.
```
davidhewitt, line 299: Maybe link to https://github.com/PyO3/pyo3/issues/538 here?

davidhewitt (commented):
Great clean-ups, thank you! Some quick thoughts from reading below...

src/python.rs
mejrs: I can see how the wording is confusing.

I'm (in general) not happy with how "references" are explained so far. Maybe it would be better to just use "owned (or strong) reference" and "borrowed (or weak) reference" everywhere, rather than "GIL-independent" and "GIL-dependent" or *just* "reference".

src/python.rs
mejrs: I don't think a link to the issue really adds anything. It's not some sort of "needs-implementer" thing where someone can just contribute a fix if they need to get around it. And we're not getting auto-traits any time soon (if ever) so it's not something that people can look forward to getting fixed.

src/python.rs
davidhewitt: Yes. Perhaps over the holiday season I will finally get another chance to look at #1056 and we can come up with better terminology then?

davidhewitt (approved):
Looks great to me, thanks again!

PR #1964, opened 2021-11-01
guide: add hints for the signature of pymethods protos

birkenfeld:
A start to give a hint about pymethod-proto args.

By the way, I discovered that it is possible to omit arguments to the methods, e.g. to define a `__richcmp__` without arguments.  It seems to be handled all right, just not passing the slot arguments to Rust, but I wonder if that's intentional.

### DIFF
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -21,22 +21,40 @@ The magic methods handled by PyO3 are very similar to the standard Python ones o
  - Magic methods for the sequence protocol
 
 When PyO3 handles a magic method, a couple of changes apply compared to other `#[pymethods]`:
-  - The `#[pyo3(text_signature = "...")]` attribute is not allowed
-  - The types of the arguments are fixed according to the magic method
+ - The `#[pyo3(text_signature = "...")]` attribute is not allowed
+ - The signature is restricted to match the magic method
 
-The following sections list of all magic methods PyO3 currently handles:
+The following sections list of all magic methods PyO3 currently handles.  The
+given signatures should be interpreted as follows:
+ - All methods take a receiver as first argument, shown as `<self>`. It can be
+   `&self`, `&mut self` or a `PyCell` reference like `self_: PyRef<Self>` and
+   `self_: PyRefMut<Self>`, as described [here](../class.md#inheritance).
+ - An optional `Python<'py>` argument is always allowed as the first argument.
+ - Return values can be optionally wrapped in `PyResult`.
+ - `object` means that any type is allowed that can be extracted from a Python
+   object (if argument) or converted to a Python object (if return value).
+ - Other types must match what's given, e.g. `pyo3::basic::CompareOp` for
+   `__richcmp__`'s second argument.
+ - For the comparison and arithmetic methods, extraction errors are not
+   propagated as exceptions, but lead to a return of `NotImplemented`.
+ - For some magic methods, the return values are not restricted by PyO3, but
+   checked by the Python interpreter. For example, `__str__` needs to return a
+   string object.  This is indicated by `object (Python type)`.
+
 
 #### Basic object customization
 
-  - `__str__`
-  - `__repr__`
-  - `__hash__`
-  - `__richcmp__`
-  - `__getattr__`
-  - `__setattr__`
-  - `__delattr__`
-  - `__bool__`
-  - `__call__`
+  - `__str__(<self>) -> object (str)`
+  - `__repr__(<self>) -> object (str)`
+  - `__hash__(<self>) -> isize`
+  - `__richcmp__(<self>, object, pyo3::basic::CompareOp) -> object`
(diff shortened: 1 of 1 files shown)


guide/src/class/protocols.md
```diff
@@ -21,22 +21,40 @@ The magic methods handled by PyO3 are very similar to the standard Python ones o
...
-  - The types of the arguments are fixed according to the magic method
+ - The `#[pyo3(text_signature = "...")]` attribute is not allowed
+ - The signature is restricted to match the magic method
+
+The following sections list of all magic methods PyO3 currently handles.  The
+given signatures should be interpreted as follows:
+ - All methods take a receiver as first argument, which can be `&self`, `&mut
+   self` or a `PyCell` reference like `self_: PyRef<Self>` and `self_:
+   PyRefMut<Self>`, as described [here](../class.md#inheritance).
+ - An optional `Python<'py>` argument is always allowed as the first argument.
```
davidhewitt, line 32: I think in the way currently implemented `Python` arguments can be in any position. I'd be happy to restrict this, however.

guide/src/class/protocols.md
```diff
@@ -21,22 +21,40 @@ The magic methods handled by PyO3 are very similar to the standard Python ones o
...
+ - Return values can be optionally wrapped in `PyResult`.
+ - `object` means that any type is allowed that can be extracted from a Python
+   object (if argument) or converted to a Python object (if return value).
+ - Other types must match what's given, e.g. `pyo3::basic::CompareOp` for
+   `__richcmp__`'s second argument.
+ - For the comparison and arithmetic methods, extraction errors are not
+   propagated as exceptions, but lead to a return of `NotImplemented`.
+ - For some slots, the return values are not restricted by PyO3, but checked by
```
davidhewitt, line 40: I've been trying to stick to "magic method".

```suggestion
 - For some slots, the return values are not restricted by PyO3, but checked by
```
```suggestion
 - For some magic methods, the return values are not restricted by PyO3, but checked by
```

guide/src/class/protocols.md
```diff
@@ -21,22 +21,40 @@ The magic methods handled by PyO3 are very similar to the standard Python ones o
...
-  - `__str__`
-  - `__repr__`
-  - `__hash__`
-  - `__richcmp__`
-  - `__getattr__`
-  - `__setattr__`
-  - `__delattr__`
-  - `__bool__`
-  - `__call__`
+  - `__str__() -> object (str)`
```
davidhewitt, line 47: Although arguably redundant, should receivers be included?

```suggestion
  - `__str__(&self) -> object (str)`
```

I guess we don't want to pin users to a single receiver type, so seeing this here will either remind them to include it or confuse them!

davidhewitt (commented):
Thanks for kicking this off!

I keep wondering... Should we add variable names too? `name: ty` is common to both rust and Python, and it'll help users know what to do with the incoming arguments.

> By the way, I discovered that it is possible to omit arguments to the methods, e.g. to define a `__richcmp__` without arguments. It seems to be handled all right, just not passing the slot arguments to Rust, but I wonder if that's intentional.

Yeah I kinda knew about this and was thinking it was a minor bug. It probably should be checked and a nice error message emitted. Users missing arguments off is probably an incorrect implementation imo so we should at least force them to be explicit about unused arguments.

The code responsible is https://github.com/PyO3/pyo3/blob/0f92f28ffd0f57bed1d8e8af2543e8f0b3a02a2f/pyo3-macros-backend/src/pymethod.rs#L955

`method_args` are the arguments the user wrote for the method; at the moment the code just iterates it without checking its length. Slightly complicated by the `Python` arguments but not too bad.

guide/src/class/protocols.md
birkenfeld: Or use something like `<self>`.

guide/src/class/protocols.md
birkenfeld: ah ok, I didn't even test this. I suggest keeping it like this in the docs anyway.

birkenfeld:
> I keep wondering... Should we add variable names too? name: ty is common to both rust and Python, and it'll help users know what to do with the incoming arguments.

As long as we don't also add a blurb about what the function does, this might not make it much clearer.

When the old system is deprecated, or when the new system is finalized, I'd like to move the descriptions from the old chapters below to the new chapter; maybe at this point it would make sense to add names.

On the other hand, the short form makes it clear that these are not copy-pastable signatures.

guide/src/class/protocols.md
davidhewitt: `<self>` sounds like a great idea to me; seems like a good placeholder that's clearly not valid syntax.

birkenfeld:
Updated!

davidhewitt (approved):
Thanks, I'm glad we have this in time for the 0.15 release!

Issue #1959, opened 2021-10-31
Put information about stub files in the documentation


KRunchPL:
### What is the request about

Describing in a documentation what to do, so your module has hints shown by IDE and types checking tools.

### Reasoning

I have just started using pyo3 to create a Python package that I can use in my Python application. It took me some time googling on how to make the hints and typing which is a very common need in my opinion. The simple "How to" would be a nice improvement to the documentation and help newbies, especially that even although my code is working I am not sure if I have found the best solution.

### Details of proposed solution

There are two issues I have found that are related to this topic:
* https://github.com/PyO3/pyo3/issues/1112
* https://github.com/PyO3/pyo3/issues/510

As far as I understood, the best way right now is to write the *.pyi files manually and just attach it to the package (with maturin it is easy).

I would be glad to help with this part of documentation, but I am not sure if above statement is completely true or we can somehow use comments/docstrings/signatures in rust code to make it happen.

davidhewitt:
Thanks, some additions to the guide would be very welcome. You're correct that the best solution at the moment is to maintain `.pyi` files manually. It would be awesome to have #510 implemented some day, however it's a future feature!

PR #1968, opened 2021-11-06
#1959 - Add Python Typing Hints to the Guide

KRunchPL:
PR adds another Appendix to the Guide describing how to use `pyi` files to provide types hinting in Python packages built using PyO3+Maturin as suggested in https://github.com/PyO3/pyo3/issues/1959


messense:
See also https://maturin-guide.netlify.app/project_layout.html#adding-python-type-information

KRunchPL:
Thanks for pointing it out, I somehow missed that information before.

I have added the section about simpler solution to the document and also fixes the formatting pointed out by CI/CD.

davidhewitt (approved):
Thank you, this is a useful resource for future readers!

Issue #1927, opened 2021-10-17
__getitem__ iteration.


fsh:
```
>>> a=PyZZ(17)
>>> len(a)
5
>>> [a[k] for k in range(len(a))]
[True, False, False, False, True]
>>> list(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'builtins.PyZZ' object is not iterable
```

I have both `__len__()` `__getitem__()` which I think is normally enough for Python to do iteration without me needing to create a separate iter object, at least that's how it's worked for me in other bindings.

From Python docs:

> Setting a special method to None indicates that the corresponding operation is not available. For example, if a class sets __iter__() to None, the class is not iterable, so calling iter() on its instances will raise a TypeError (without falling back to __getitem__()).

Does PyO3 do something like that to block the fallback behavior? Is it possible to stop it from doing it?

I tried returning `PyRef<Self>` from `__iter__` but then I get the error that a non-iterable was returned. If I define also `__next__` to some dummy, of course it just doesn't iterate. I'd like to avoid having to use the `__iter__` `__next__` if I can, because I have a lot of small types and merely want them to be iterable out of convenience, and the _desired_ behavior in my case is for Python to fall back to using `__getitem__[]` iteration.

For reference my actual `__getitem__` is as:

```rust
  fn __getitem__(&self, idx: Either2<&PySlice,isize>) -> PyResult<bool> {
    let bz = self.el.significant_bits() as isize;
    match idx {
      Either2::A(sl) => todo!(),
      Either2::B(i) => {
        let i = if i < 0 { bz + i } else { i };
        if i < 0 || i >= bz {
          Err(PyIndexError::new_err("out of bounds"))
        } else {
          Ok(self.el.get_bit(i as u32))
        }
      }
    }
  }
```

Possibly related to #1855 or the ongoing migration work?

fsh:
Oops, I definitely skimmed the other issue too quickly. This is indeed a duplicate of #1855. Feel free to mark as such.

davidhewitt:
Hmm, thanks for the head-up. I don't think this is a duplicate.

Looks like there are four methods with special fallbacks implemented in CPython: https://docs.python.org/3/reference/datamodel.html#id9: `__hash__`, `__iter__`, `__reversed__`, and `__contains__`

We don't implement equivalent fallbacks in PyO3, but should consider doing so.

fsh:
[...]

PR #1971, opened 2021-11-07
pymethods: test and document opt-out of protos

davidhewitt:
This is inspired by the discussion in #1927.

Some protocol methods such as `__hash__`, `__iter__`, `__contains__`, and `__reversed__` have automatically-generated default implementations which CPython's operators use if other protocols are available:
- `__hash__` always gets a default
- `__iter__` if *sequence* `__getitem__` is implemented (`Py_sq_item` slot)
- `__contains__` if `__iter__` is implemented
- `__reversed__` if `__iter__` and *sequence* length is implemented (`Py_sq_length` slot)

This PR tests and documents the defaults and opt-outs for `__hash__` and `__contains__`. `__iter__` and `__reversed__` require the sequence protocol support in #1884 first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ PyO3 versions, please see the [migration guide](https://pyo3.rs/latest/migration
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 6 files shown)


davidhewitt:
I'd really appreciate some feedback on this one - in particular, I chose to use `<details>` in the magic methods docs to make the examples collapsible. I think that's a good idea to reduce length of the page, but does have the drawback that users may miss information. How does everyone else feel about use of `<details>`?

birkenfeld:
I think the `<details>` is fine, but it shouldn't be labeled as "Example".

davidhewitt:
👍 that's fair. I'll remove "Example" from the wording and then proceed to merge this.

PR #1975, opened 2021-11-10
ci: try running coverage for all OSes

davidhewitt:
According to https://docs.codecov.com/docs/merging-reports if we upload multiple coverage reports they all get automatically merged. So I'm opening this PR to see what happens if I try that!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -266,7 +266,11 @@ jobs:
 
   coverage:
     needs: [fmt]
-    runs-on: ubuntu-latest
+    name: coverage-${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ["windows", "macos", "ubuntu"]
+    runs-on: ${{ matrix.os }}-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


davidhewitt:
Hmmm @taiki-e have you seen errors like these from running `cargo-llvm-cov` on windows before?

```
libpyo3-8bd03e2eb0b9a94d.rlib(pyo3-8bd03e2eb0b9a94d.pyo3.e32d66c5-cgu.2.rcgu.o) : fatal error LNK1227: conflicting weak extern definition for '_RNvNtCs9ZwDYBEfEGf_4pyo311type_object12get_tp_allocB3_'.  New default '.weak._RNvNtCs9ZwDYBEfEGf_4pyo311type_object12get_tp_allocB3_.default._RINvMNtCsk1Jcy3OMEV4_5alloc5sliceSh9to_vec_inNtNtB5_5alloc6GlobalECs9ZwDYBEfEGf_4pyo3' conflicts with old default '.weak._RNvNtCs9ZwDYBEfEGf_4pyo311type_object12get_tp_allocB3_.default._RINvMNtCsjJUIXhzXbX_4core3stre13get_uncheckedINtNtNtB5_3ops5range5RangejEECs9ZwDYBEfEGf_4pyo3' in libpyo3-8bd03e2eb0b9a94d.rlib(pyo3-8bd03e2eb0b9a94d.pyo3.e32d66c5-cgu.9.rcgu.o)
```

taiki-e:
@davidhewitt  That is a rustc bug: https://github.com/rust-lang/rust/issues/85461

I'll implement a workaround on the cargo-llvm-cov side.

davidhewitt:
Aha brilliant, thank you!

taiki-e:
@davidhewitt I've released cargo-llvm-cov 0.1.11, which implements a workaround for that issue. Could you try that version?

.github/workflows/ci.yml
```diff
@@ -280,12 +297,24 @@ jobs:
...
+        with:
+          toolchain: nightly
+          override: true
+          profile: minimal
+          components: llvm-tools-preview
+      - if: matrix.os.name != 'windows'
+        name: install cargo-llvm-cov (unix)
+        run: |
+          wget https://github.com/taiki-e/cargo-llvm-cov/releases/download/v${CARGO_LLVM_COV_VERSION}/cargo-llvm-cov-${CARGO_LLVM_COV_ARCH}.tar.gz -qO- | tar -xzvf -
           mv cargo-llvm-cov ~/.cargo/bin
```
taiki-e, line 317: Btw, cargo-llvm-cov has both .tar.gz and .zip distribution on windows, so you can unify the installation script on both windows and unix (with `shell: bash`). For example:

```
host=$(rustc -Vv | grep host | sed 's/host: //')
curl -fsSL --retry 3 https://github.com/taiki-e/cargo-llvm-cov/releases/download/v${CARGO_LLVM_COV_VERSION}/cargo-llvm-cov-"$host".tar.gz | tar xzf - -C ~/.cargo/bin
```

davidhewitt:
Thanks @taiki-e, the windows job working great now!

I _think_ this looks like codecov is merging reports? Not super clear, I'm going to go ahead and merge this, might play around more another day.

taiki-e:
> I _think_ this looks like codecov is merging reports? Not super clear, I'm going to go ahead and merge this, might play around more another day.

This may be due to the different path between windows and unix. (I feel it's a bit odd that codecov can't handle it, but we may need to do [what cargo-llvm-cov's test is doing](https://github.com/taiki-e/cargo-llvm-cov/blob/01a7fe32585113f707d7c5ccda236b081b9ea54f/tests/auxiliary/mod.rs#L94-L99))

PR #1977, opened 2021-11-10
types: add PyTraceback

### LINKED ISSUE #592: pretty python exception traceback in rust
The `print` method on `PyErr` (https://pyo3.rs/master/doc/pyo3/struct.PyErr.html#method.print)   is super handy but relies on stderr being available on the python side. When embedding rust + python into another application this is not always available and so it fails silently. It would be better if we had a way of getting the string of the error or the traceback as a string in *rust*, so that we can use it with whatever logging/error infrastructure is already set up on the rust side (eg `log!`, the `failure` crate, etc)

At the moment I've been doing some variation of `py.eval("str(err)")` or `py.eval(import traceback; # etc)` which works, but is cumbersome.

davidhewitt:
Adds `PyTraceback` type, with a `.format()` method to stringify the traceback object.

Closes #592 

This changes `PyErr::ptraceback()` to return `Option<&PyTraceback>` instead of `Option<&PyAny>`, but I don't think this is breaking as `&PyTraceback` will coerce to `&PyAny`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add implementations for `Py::as_ref()` and `Py::into_ref()` for `Py<PySequence>`, `Py<PyIterator>` and `Py<PyMapping>`. [#1682](https://github.com/PyO3/pyo3/pull/1682)
+- Add `PyTraceback` type to represent and format Python tracebacks. [#1977](https://github.com/PyO3/pyo3/pull/1977)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


birkenfeld:
This is totally a bikeshed, but the `TraceBack` capitalization looks funny to me. Unfortunately CPython uses both `TraceBack` and `Traceback`, so there's no good example to follow.

davidhewitt:
👍 agreed, I was on the fence about this as I also thought `PyTraceBack` looked a bit clunky. Changed to `PyTraceback`.

davidhewitt:
I'll probably merge this early next week if there's no further concerns from anyone.

PR #1980, opened 2021-11-10
Add PyCapsule API

### LINKED ISSUE #1193: Expose PyCapsule API
The capsule API (https://docs.python.org/3.8/c-api/capsule.html) can be useful for safely storing Rust data in a way that's not usable by Python without leaking it. This is possible because the capsule creation takes a destructor function.

Today it came up on Gitter that for `rust-numpy` it may be useful to do zero-copy transfer of arrays from Rust to Python, by stashing the backing storage (e.g. vec) in a capsule which is set as the base object to the numpy array.

milesgranger:
Will close #1193 

Ref: https://docs.python.org/3/c-api/capsule.html#capsules

TODO: 
- [x]  an entry in CHANGELOG.md
- [x]  docs to all new functions and / or detail in the guide
- [x]  tests for all new or changed functions


davidhewitt:
Just wanted to say thank-you for working on this - it's a really nice feature to add to PyO3. I've been a bit busy recently, however trying to find a moment to review when I can...

milesgranger:
No worries; thank you! I'm still chipping away as I get some time here and there. I'm sure there are some glaring errors I've missed, so put it here anyway if someone wanted to take a peak.

mejrs:
Thanks! Sorry for not getting to this sooner.

Unfortunately casting away lifetimes and types like this is really, really unsafe.

For example, this segfaults:

```rust
fn main() {
    let cap: Py<PyCapsule> = Python::with_gil(|py| {
        let name = CString::new("foo").unwrap();
        let cap = PyCapsule::new(py, (), &name, None).unwrap();

        let stuff: Vec<u8> = vec![1, 2, 3, 4];

        cap.set_context(py, &stuff).unwrap();

        cap.into()
    });

    Python::with_gil(|py| {
        let ctx: Option<&&Vec<u8>> = cap.as_ref(py).get_context(py).unwrap();
        dbg!(ctx);
    })
}
```

Or, you can implement transmute:

```rust
fn transmute<'p, T, U>(py: Python<'p>, t: T) -> &'p U {
    let name = CString::new("foo").unwrap();
    let cap = PyCapsule::new(py, (), &name, None).unwrap();
    cap.set_context(py, &t).unwrap();
    let ctx: Option<&U> = cap.get_context(py).unwrap();
    ctx.unwrap()
}

fn main() {
    Python::with_gil(|py| {
        let foo = 32u8;
        let boom: &Vec<u32> = transmute(py, &foo);
        dbg!(boom);
    });
}
```

And you can also use `Py<PyCapsule>` to smuggle things like `Rc` to other threads.

A good way to fix some of this is to introduce some variance - i.e., make the type `PyCapsule<T>`, so that type information is not lost. However this may make a safe api of PyCapsule almost useless.

Maybe you can use `dyn Any` for this? You can do some runtime reflection with it.

milesgranger:
I suspected I'd fouled that up a bit. Thanks for the thorough explanation, the failing example is nice to have. I'll work on it; suspect this PR will be "slow to update". Working on it as time presents itself. :-)

milesgranger:
My current understanding was capsules were used to store by value and not by reference (unless static reference). If I modify your example to pass by value, it works fine. Thus, if I update the signatures of `new` and `set_context` to have `T: 'static` then at least your example would no longer compile. Then current tests and the updated example here all pass:

```rust
pub fn new<'py, T: 'static>(
        py: Python<'py>,
        value: T,
        name: &CStr,
        destructor: Option<ffi::PyCapsule_Destructor>,
    ) -> PyResult<&'py Self>;

pub fn set_context<'py, T: 'static>(&self, py: Python<'py>, context: T) -> PyResult<()>;
```

I'm still pretty confident the value of `Box::into_raw(T) as *mut c_void` leaked. Which I suppose is another issue.

src/pycapsule.rs
```diff
@@ -0,0 +1,276 @@
...
+
+impl PyCapsule {
+    /// Constructs a new capsule of whose contents are `T` associated with `name`.
+    /// Can optionally provide a deconstructor for when `PyCapsule` is destroyed
+    /// it will be passed the capsule.
+    pub fn new<'py, T>(
+        py: Python<'py>,
+        value: T,
+        name: &CStr,
+        destructor: Option<ffi::PyCapsule_Destructor>,
```
davidhewitt, line 55: I think that asking the user to provide the destructor which receives `*mut void` is hard for users to use correctly; for example the user should use `Box::from_raw` to avoid leaking the value, but that's not really obvious at all. Also possible panics and maybe GIL interactions.

One possible alternative would be to ask users to pass a destructor which takes T:

```rust
destructor: Option<impl FnOnce(T)>
```

And then we can box this up with the value (instead of just boxing the value), and internally create the "real" destructor which first unboxes the value and calls the closure:

```rust
struct ValueAndDestructor<T, D: FnOnce(T)> {
    value: T,
    destructor: D
}

unsafe extern "C" fn capsule_destructor<T, F: FnOnce<T>>(capsule: *mut PyObject) {
    // FIXME might need to use `handle_panic` here and create a gilpool?
    let ptr = unsafe { PyCapsule_GetPointer(capsule, PyCapsule_GetName(capsule)) };
    let ValueAndDestructor { value, destructor } = *unsafe { Box::from_raw(ptr as *mut ValueAndDestructor<T, F>) };
    destructor(value)
}
```

And then I think if the user does not provide a destructor, would default to `std::mem::drop` or some other no-op method.

... This design doesn't take into account the `context` pointer which can also be stored in the capsule (or it's name). Pehaps the capsule could be a second argument to the `FnOnce`?

src/pycapsule.rs
```diff
@@ -0,0 +1,276 @@
...
+#[repr(transparent)]
+pub struct PyCapsule(PyAny);
+
+pyobject_native_type_core!(PyCapsule, ffi::PyCapsule_Type, #checkfunction=ffi::PyCapsule_CheckExact);
+
+impl PyCapsule {
+    /// Constructs a new capsule of whose contents are `T` associated with `name`.
+    /// Can optionally provide a deconstructor for when `PyCapsule` is destroyed
+    /// it will be passed the capsule.
+    pub fn new<'py, T>(
```
davidhewitt, line 51: It might be nice to call this `new_with_destructor` (making the argument non-optional) and have `new` which doesn't have a destructor?

src/pycapsule.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// Import an existing capsule.
+    ///
+    /// The `name` should match the path to module attribute exactly in the form
+    /// of `module.attribute`, which should be the same as the name within the
+    /// capsule. `no_block` indicates to use
+    /// [PyImport_ImportModuleNoBlock()](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModuleNoBlock)
+    /// or [PyImport_ImportModule()](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModule)
+    /// when accessing the capsule.
+    pub fn import<'py, T>(py: Python<'py>, name: &CStr, no_block: bool) -> PyResult<&'py T> {
```
davidhewitt, line 77: I think this should either be `unsafe` or alternatively should return `PyResult<*mut void>` and let the user cast the pointer at their own peril; this function looks like a hidden safe transmute.

src/pycapsule.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /// Check if this is a valid capsule.
+    pub fn is_valid(&self) -> bool {
+        let r = unsafe { ffi::PyCapsule_IsValid(self.as_ptr(), self.name().as_ptr()) } as u8;
+        r != 0
+    }
+
+    /// Get the capsule deconstructor, if any.
+    pub fn get_deconstructor(&self, py: Python) -> PyResult<Option<ffi::PyCapsule_Destructor>> {
```
davidhewitt, line 132: Typo: `deconstructor` -> `destructor`.

```suggestion
    /// Get the capsule destructor, if any.
    pub fn get_destructor(&self, py: Python) -> PyResult<Option<ffi::PyCapsule_Destructor>> {
```

src/pycapsule.rs
```diff
@@ -0,0 +1,276 @@
...
+            unsafe { ffi::PyCapsule_SetContext(self.as_ptr(), Box::into_raw(ctx) as _) as u8 };
+        if result != 0 {
+            Err(PyErr::fetch(py))
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Get a reference to the context `T` in the capsule, if any.
+    pub fn get_context<T>(&self, py: Python) -> PyResult<Option<&T>> {
```
davidhewitt, line 99: Similar concern as `import` r.e. transmute risk.

davidhewitt (commented):
Thanks, in addition to discussion above I finally had some time to give this a read today. I think there's a few things which need to be solved yet.

src/pycapsule.rs
milesgranger: I've updated it with a similar implementation, but I end up being confused. 

The pointer value of the capsule is pointing to a struct for `T` and `F`, and _somehow_ `get_reference<T>(&self) -> &T` still works, even though it's really just getting the pointer to this `CapsuleContents` struct that holds the value _and_ destructor. Maybe it's some Rust/memory internals I don't know about yet that allows that to work.

I was thinking it ought to be something more like:
```rust
pub unsafe fn reference<T, F>(&self) -> &T {
   let contents = Box::from_raw(self.get_pointer() as *mut &CapsuleContents<T, F> };
   &contents.value
}
```

but again, it works as-is. And even if it didn't, would this be good? Seems having to supply type parameters for everything in `CapsuleContents` would be annoying, especially for methods like `reference` or `get_destructor` where the user is only interested in getting a single value, and yet required to supply types for all values in a capsule. Or maybe it isn't that bad...

Do you think having a macro of some sort for a destructor would be helpful for the user?
ie:

```rust
#[capsule_destructor]
fn destructor(capsule: PyCapsule) {
    ...smart stuff...
}

// expands to something like
unsafe extern "C" fn capsule_destructor(capsule: *mut PyObject) {
    let ptr = unsafe { PyCapsule_GetPointer(capsule, PyCapsule_GetName(capsule)) };
    let capsule = *Box::from_raw(ptr as *mut PyCapsule);
    ...smart stuff...
}
```

src/pycapsule.rs
```diff
@@ -0,0 +1,330 @@
...
+    ///
+    /// The `name` should match the path to module attribute exactly in the form
+    /// of `module.attribute`, which should be the same as the name within the
+    /// capsule. `no_block` indicates to use
+    /// [PyImport_ImportModuleNoBlock()](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModuleNoBlock)
+    /// or [PyImport_ImportModule()](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModule)
+    /// when accessing the capsule.
+    ///
+    /// ## Safety
+    /// This is unsafe, as there is no guarantee when casting `*mut void` into `T`.
```
davidhewitt, line 103: I think it's more helpful for users to indicate what invariant they must uphold. Also, elsewhere we've done these as only one-level headings:

```suggestion
    /// # Safety
    /// 
    /// It must be known that this capsule's pointer is to an item of type `T`.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,330 @@
...
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Get a reference to the context `T` in the capsule, if any.
+    ///
+    /// ## Safety
+    ///
+    /// This is unsafe, as there is no guarantee when casting `*mut void` into `T`.
```
davidhewitt, line 129: Similar comment on the safety note here.

src/pycapsule.rs
```diff
@@ -0,0 +1,330 @@
...
+            }
+        } else {
+            Ok(Some(&*(ctx as *const T)))
+        }
+    }
+
+    /// Obtain a reference to the value `T` of this capsule.
+    ///
+    /// # Safety
+    /// This is unsafe because there is no guarantee the pointer is `T`
```
davidhewitt, line 146: And again.

src/pycapsule.rs
```diff
@@ -0,0 +1,330 @@
...
+    pub fn is_valid(&self) -> bool {
+        let r = unsafe { ffi::PyCapsule_IsValid(self.as_ptr(), self.name().as_ptr()) } as u8;
+        r != 0
+    }
+
+    /// Get the capsule destructor, if any.
+    pub fn get_destructor<T: 'static, D: FnOnce(T)>(&self) -> PyResult<Option<&D>> {
+        let val = unsafe { Box::from_raw(self.get_pointer() as *mut &CapsuleContents<T, D>) };
+        Ok(Some(&val.destructor))
+    }
```
davidhewitt, line 166: TBH this method looks nearly impossible to use correctly due to the type annotations. It's also `unsafe` because there's no guarantee the capsule holds these types. Perhaps better to just not have this, and let users go for raw FFI if they need this?

src/pycapsule.rs
```diff
@@ -0,0 +1,330 @@
...
+    fn new(value: T, destructor: D) -> Self {
+        Self { value, destructor }
+    }
+}
+
+unsafe extern "C" fn capsule_destructor<T: 'static, F: FnOnce(T)>(capsule: *mut ffi::PyObject) {
+    let ptr = ffi::PyCapsule_GetPointer(capsule, ffi::PyCapsule_GetName(capsule));
+    let CapsuleContents { value, destructor } = *Box::from_raw(ptr as *mut CapsuleContents<T, F>);
+    destructor(value)
+}
```
davidhewitt, line 61: As these are private helpers unique to this file, perhaps put them after the `impl PyCapsule` block?

src/pycapsule.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// Get a reference to the context `T` in the capsule, if any.
+    ///
+    /// ## Safety
+    ///
+    /// This is unsafe, as there is no guarantee when casting `*mut void` into `T`.
+    pub unsafe fn get_context<T>(&self, py: Python) -> PyResult<Option<&T>> {
+        let ctx = ffi::PyCapsule_GetContext(self.as_ptr());
+        if ctx.is_null() {
+            if self.is_valid() & PyErr::occurred(py) {
```
davidhewitt, line 133: Bitwise operator looks potentially unintentional

```suggestion
            if self.is_valid() && PyErr::occurred(py) {
```

davidhewitt:
> The pointer value of the capsule is pointing to a struct for T and F, and somehow get_reference<T>(&self) -> &T still works, even though it's really just getting the pointer to this CapsuleContents struct that holds the value and destructor.

Because the first field of the `CapsuleContents` struct is the `T`, through blind luck the compiler's put that at the front of the struct and so the cast works. This is UB because the default struct layout is not guaranteed. If you add `#[repr(C)]` to `CapsuleContents` (maybe with a comment why that repr is used), it's sound to rely on first field `T` always being at zero offset inside the struct.

https://doc.rust-lang.org/reference/type-layout.html#the-c-representation

> Do you think having a macro of some sort for a destructor would be helpful for the user?

It's not clear to me what the use case for this would be. Also, macros are hard for users to figure out how to use without lots of good documentation, so without a strong motivation I'm unconvinced it's worth the maintenance cost.

davidhewitt:
Thanks for the continued progress on this, I think it's getting there and will be a nice feature to add to PyO3!

src/pycapsule.rs
milesgranger: Nice! I agree, it was struggling to understand how to get back the original function the user wrote (and would expect) from this method after wrapping it into `ffi::PyCapsule_Destructor` type.

src/pycapsule.rs
milesgranger: Agreed, sloppy on my part to put it here.

milesgranger:
Thank you for the thorough reviews here, a lot of new stuff here, so appreciate it. I know you're busy! :pray:

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+                Box::into_raw(val) as *mut c_void,
+                name.as_ptr(),
+                Some(capsule_destructor::<T, F>),
+            )
+        };
+        if cap_ptr.is_null() {
+            Err(PyErr::fetch(py))
+        } else {
+            Ok(unsafe { py.from_owned_ptr::<PyCapsule>(cap_ptr) })
+        }
```
davidhewitt, line 74: This one can be simplified:

```suggestion
        unsafe { py.from_owned_ptr_or_err(cap_ptr) }
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// or [PyImport_ImportModule()](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModule)
+    /// when accessing the capsule.
+    ///
+    /// # Safety
+    ///
+    /// It must be known that this capsule's pointer is to an item of type `T`.
+    pub unsafe fn import<'py, T>(py: Python<'py>, name: &CStr, no_block: bool) -> PyResult<&'py T> {
+        let ptr = ffi::PyCapsule_Import(name.as_ptr(), no_block as c_int);
+        if ptr.is_null() {
+            Err(PyErr::fetch(py))
```
davidhewitt, line 92: Might be nice to add a test which tries to call `PyCapsule::import` with some made up name so that we can exercise this line checking for the error.

davidhewitt (approved):
👍 This is looking great to me! Just needs an `Added` CHANGELOG entry and then a couple of final suggestions.

Thanks for implementing this.

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+///  assert!(r.is_ok());
+/// ```
+#[repr(transparent)]
+pub struct PyCapsule(PyAny);
+
+pyobject_native_type_core!(PyCapsule, ffi::PyCapsule_Type, #checkfunction=ffi::PyCapsule_CheckExact);
+
+impl PyCapsule {
+    /// Constructs a new capsule of whose contents are `T` associated with `name`.
+    pub fn new<'py, T: 'static>(py: Python<'py>, value: T, name: &CStr) -> PyResult<&'py Self> {
```
birkenfeld, line 49: Doesn't this need at least a `T: Send` bound?  Once given to Python the object can be transferred to any thread, so both `import()` as well as the destructor can be called from any thread.

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// of `module.attribute`, which should be the same as the name within the
+    /// capsule. `no_block` indicates to use
+    /// [PyImport_ImportModuleNoBlock()](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModuleNoBlock)
+    /// or [PyImport_ImportModule()](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModule)
+    /// when accessing the capsule.
+    ///
+    /// # Safety
+    ///
+    /// It must be known that this capsule's pointer is to an item of type `T`.
+    pub unsafe fn import<'py, T>(py: Python<'py>, name: &CStr, no_block: bool) -> PyResult<&'py T> {
```
birkenfeld, line 89: `no_block` is not needed anymore since since Python 3.3, the special "noblock" functionality has been removed.

See https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModuleNoBlock

I'll make a PR to CPython later to fix this in the C-API docs too.

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+
+pyobject_native_type_core!(PyCapsule, ffi::PyCapsule_Type, #checkfunction=ffi::PyCapsule_CheckExact);
+
+impl PyCapsule {
+    /// Constructs a new capsule of whose contents are `T` associated with `name`.
+    pub fn new<'py, T: 'static>(py: Python<'py>, value: T, name: &CStr) -> PyResult<&'py Self> {
+        Self::new_with_destructor(py, value, name, std::mem::drop)
+    }
+
+    /// Constructs a new capsule of whose contents are `T` associated with `name`
```
birkenfeld, line 53: nit: leave an empty line between first and subsequent lines.
```suggestion
    /// Constructs a new capsule whose contents are `value`, associated with `name`.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Provide a destructor for when `PyCapsule` is destroyed it will be passed the capsule.
```
birkenfeld, line 54:
```suggestion
    /// Also provides a destructor: when the `PyCapsule` is destroyed, it will be passed the original object.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+///  });
+///  assert!(r.is_ok());
+/// ```
+#[repr(transparent)]
+pub struct PyCapsule(PyAny);
+
+pyobject_native_type_core!(PyCapsule, ffi::PyCapsule_Type, #checkfunction=ffi::PyCapsule_CheckExact);
+
+impl PyCapsule {
+    /// Constructs a new capsule of whose contents are `T` associated with `name`.
```
birkenfeld, line 48:
```suggestion
    /// Constructs a new capsule whose contents are `value`, associated with `name`.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
+// Copyright (c) 2017-present PyO3 Project and Contributors
+use crate::Python;
+use crate::{ffi, AsPyPointer, PyAny};
+use crate::{pyobject_native_type_core, PyErr, PyResult};
+use std::ffi::{c_void, CStr};
+use std::os::raw::c_int;
+
+/// Represents a Python Capsule
+/// As described in [Capsules](https://docs.python.org/3/c-api/capsule.html#capsules)
```
birkenfeld, line 9:
```suggestion
/// as described in [Capsules](https://docs.python.org/3/c-api/capsule.html#capsules):
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+            )
+        };
+        if cap_ptr.is_null() {
+            Err(PyErr::fetch(py))
+        } else {
+            Ok(unsafe { py.from_owned_ptr::<PyCapsule>(cap_ptr) })
+        }
+    }
+
+    /// Import an existing capsule.
```
birkenfeld, line 77:
```suggestion
    /// Imports an existing capsule.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    ///
+    /// The `name` should match the path to module attribute exactly in the form
```
birkenfeld, line 79:
```suggestion
    /// The `name` should match the path to the module attribute exactly in the form
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    pub unsafe fn import<'py, T>(py: Python<'py>, name: &CStr, no_block: bool) -> PyResult<&'py T> {
+        let ptr = ffi::PyCapsule_Import(name.as_ptr(), no_block as c_int);
+        if ptr.is_null() {
+            Err(PyErr::fetch(py))
+        } else {
+            Ok(&*(ptr as *const T))
+        }
+    }
+
+    /// Set a context pointer in the capsule to `T`
```
birkenfeld, line 98:
```suggestion
    /// Sets the context pointer in the capsule to `T`.
```

This Box is leaked; there is no place where the context is destructed. I guess this is unavoidable, but it must be at least be made explicit in the docstring.

Also, either needs a `T: Send` or `get_context` must make sure not to be called from a different thread (but that would limit its usability severely).

Also needs a little explanation of what the context is, and that it's different from `value`.

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+        let result =
+            unsafe { ffi::PyCapsule_SetContext(self.as_ptr(), Box::into_raw(ctx) as _) as u8 };
+        if result != 0 {
+            Err(PyErr::fetch(py))
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Get a reference to the context `T` in the capsule, if any.
```
birkenfeld, line 110:
```suggestion
    /// Gets a reference to the context of the capsule, if any.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    ///
+    /// # Safety
+    ///
+    /// It must be known that this capsule's contains a context item of type `T`.
```
birkenfeld, line 114:
```suggestion
    /// It must be known that this capsule's contains a context of type `T`.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+                Err(PyErr::fetch(py))
+            } else {
+                Ok(None)
+            }
+        } else {
+            Ok(Some(&*(ctx as *const T)))
+        }
+    }
+
+    /// Obtain a reference to the value `T` of this capsule.
```
birkenfeld, line 128:
```suggestion
    /// Obtains a reference to the value of this capsule.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    ///
+    /// # Safety
+    ///
+    /// It must be known that this capsule's pointer is to an item of type `T`.
+    pub unsafe fn reference<T>(&self) -> &T {
+        &*(self.get_pointer() as *const T)
+    }
+
+    /// Get the raw `c_void` pointer to the value in this capsule.
```
birkenfeld, line 137:
```suggestion
    /// Gets the raw `c_void` pointer to the value in this capsule.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    pub fn get_pointer(&self) -> *mut c_void {
+        unsafe { ffi::PyCapsule_GetPointer(self.0.as_ptr(), self.name().as_ptr()) }
+    }
+
+    /// Check if this is a valid capsule.
```
birkenfeld, line 142:
```suggestion
    /// Checks if this is a valid capsule.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,296 @@
...
+    pub fn is_valid(&self) -> bool {
+        let r = unsafe { ffi::PyCapsule_IsValid(self.as_ptr(), self.name().as_ptr()) } as u8;
+        r != 0
+    }
+
+    /// Retrieve the name of this capsule.
```
birkenfeld, line 148:
```suggestion
    /// Retrieves the name of this capsule.
```

birkenfeld (changes requested):

src/pycapsule.rs
milesgranger: Thanks for the review! 

I think this will handicap the functionality quite a bit. I personally, was hoping to use capsules to pass some functions, but [it doesn't appear fn is Send](https://doc.rust-lang.org/std/primitive.fn.html#)

Do you know how one would deal with this situation? (ie the current `test_pycapsule_func` then fails.

src/pycapsule.rs
milesgranger: nice!

src/pycapsule.rs
birkenfeld: I don't see why a function pointer wouldn't be `Send` (and the [playground seems to agree](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dbabdd6b7cebdd11ceedd4b9f2a2d3ae))...

I think the docs don't show the auto traits if they are derived. Only if you explicitly do an `impl !Send` or `impl Send where ...`.

src/pycapsule.rs
milesgranger: Okay, so I've updated it to pass false then, to `PyCapsule_Import` and removed the parameter, is that the right way to go?

birkenfeld: Yes, sounds good.

src/pycapsule.rs
davidhewitt: `fn` should be `Send` - and yes good catch, this should definitely require `Send`.

birkenfeld: Ah, I now see the original test was casting the pointer to `*const c_void`, which is not Send. Looking at e.g. https://internals.rust-lang.org/t/shouldnt-pointers-be-send-sync-or/8818 it seems that this is not really a question of safety.

The easiest workaround seems to be to make a struct `SendPtr(*const c_void)` with `#[repr(transparent)]` and have an `unsafe impl Send` for it. Maybe we can supply this in this module? Let's hear @davidhewitt 's opinion...

src/pycapsule.rs
milesgranger: Took a swing at it, hard to find any information on exactly what context is used for itself, so I at least explains what I think it is used for. :sweat_smile: 

Maybe we can alter `capsule_destructor` to pass the whole of `PyCapsule`  (un borrowed) to the destructor, then add `take_value` / `take_context`  methods which would, in my mind at least, allow the user supplied `destructor` to be `Fn(PyCapsule)` and thus full ownership to cleanup anything?

src/pycapsule.rs
davidhewitt: `0bd03de` looks like it's solved this one by removing the cast?

src/pycapsule.rs
davidhewitt: Ah, good point. The nice thing about the `FnOnce(T)` design we've currently got is that the user doesn't have to do any `unsafe` type casting in their destructor implementation.

> full ownership to cleanup anything

We can't actually have unborred `PyCapsule` as we don't know its true memory layout (that's private to CPython).

We could potentially have `FnOnce(T, *mut void)` and let the user handle non-null context? That's a bit icky, however I don't see much better. `new_with_context` constructor? Then we'll have combinatorial explosion of constructors :(

src/pycapsule.rs
milesgranger: That one (just passing the function itself) gives a segfault now, which I haven't figured out yet.

birkenfeld: IMO you'll have to cast the function to `fn(u32) -> u32` when making the box. Function items (`foo` alone) are zero sized types, not function pointers, so the box will be "empty".

Since it never makes sense to put ZSTs into a capsule, we should at least document this and put an `assert!(size_of::<T>() > 0)` to panic at capsule creation if the invariant is violated.

milesgranger: Great! I had no idea, casting it fixed it; thank you. Learning a lot here. :-) 
I've updated; for my own understanding, why is it preferred to panic instead of returning an error in this case?

src/pycapsule.rs
milesgranger: I see, for now, I've went with your last suggestion (`FnOnce(T, *mut void)`)

CHANGELOG.md
```diff
@@ -47,6 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix build failure on PyPy when abi3 features are activated. [#1991](https://github.com/PyO3/pyo3/pull/1991)
 - Fix mingw platform detection. [#1993](https://github.com/PyO3/pyo3/pull/1993)
 - Fix panic in `__get__` implementation when accessing descriptor on type object. [#1997](https://github.com/PyO3/pyo3/pull/1997)
+>>>>>>> main
```
messense, line 52: Looks like a bad merge here.

milesgranger: Oh my, it's amateur hour over here on my end. :man_facepalming:  Thank you.

src/pycapsule.rs
birkenfeld: Just an instinct; this is a violation of the contract (at least when it's documented as such) and can never work, so failing fast would be my preference here.

It would of course be preferred to have a static assert that fires at compile time. There is no such facility in `std` yet, AFAIK.

src/pycapsule.rs
```diff
@@ -0,0 +1,325 @@
...
+///
+///
+/// # Example
+/// ```
+///  use std::ffi::CString;
+///  use pyo3::{prelude::*, pycapsule::PyCapsule};
+///
+///  #[repr(C)]
+///  struct Foo {
+///     pub val: u32,
```
birkenfeld, line 25: One more nit, the examples here are indented with 3 spaces instead of 4.

src/pycapsule.rs
```diff
@@ -0,0 +1,325 @@
...
+/// ```
+#[repr(transparent)]
+pub struct PyCapsule(PyAny);
+
+pyobject_native_type_core!(PyCapsule, ffi::PyCapsule_Type, #checkfunction=ffi::PyCapsule_CheckExact);
+
+impl PyCapsule {
+    /// Constructs a new capsule whose contents are `value`, associated with `name`.
+    ///
+    /// # Notes
```
birkenfeld, line 51: Normally the heading is `# Panics` to make it obvious.

src/pycapsule.rs
```diff
@@ -0,0 +1,325 @@
...
+///  });
+///  assert!(r.is_ok());
+/// ```
+#[repr(transparent)]
+pub struct PyCapsule(PyAny);
+
+pyobject_native_type_core!(PyCapsule, ffi::PyCapsule_Type, #checkfunction=ffi::PyCapsule_CheckExact);
+
+impl PyCapsule {
+    /// Constructs a new capsule whose contents are `value`, associated with `name`.
```
birkenfeld, line 49: I guess the meaning of `name` should be explained here as well. Sorry, should have pointed that out before.

src/pycapsule.rs
```diff
@@ -0,0 +1,325 @@
...
+        }
+    }
+
+    /// Sets the context pointer in the capsule to `T`.
+    ///
+    /// # Notes
+    ///
+    /// Context is not destructed like the value of the capsule is by `destructor`. Therefore,
+    /// it's likely this value will be leaked. If set, it's up to the user to either ignore this
+    /// side effect or figure another (unsafe) method of clean up.
```
birkenfeld, line 115: This seems a bit outdated now that the value is passed into the destructor. Suggestion:

```
The default destructor does not drop the context value. If you need to ensure it is not leaked, use `new_with_destructor`, let it call `Box::from_raw()` with the correct type, and drop it.

Finally, if `set_context` is called twice in a row, the previous value is always leaked.
```

It seems that it would be better to treat the context like the value, i.e. another type parameter for `new` and the destructor. That would also allow dropping the previous context properly on `set_context`.

However, this depends on how context is supposed to be used (which I don't know either): if the context is supposed to be set by the other side (which might well be C), this would not help, and rather be detrimental since the destructor would try to drop something non-Rust.

It's all a bit iffy. I wonder if we should instead simply have the API operate on raw pointers. People can still call `Box::into_raw` for Rust objects, and it would be more symmetric between here and the destructor. @davidhewitt ?

src/pycapsule.rs
milesgranger: Added a line, referencing much of the same description for name given in the [C-API docs](https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_New).

src/pycapsule.rs
birkenfeld: Well, it *is* possible: https://users.rust-lang.org/t/is-it-possible-to-assert-at-compile-time-that-foo-t-is-not-called-with-a-zst

src/pycapsule.rs
davidhewitt: Yes, I was beginning to wonder whether the `set_context` should just take `*mut void` too. If we have an example which demonstrates `into_raw` (and `from_raw` in the destructor) then there's something for them to copy.

Because anything can mutate the capsule using the raw FFI calls, I'm leaning away from us trying to do anything too smart - at some point it's better if the `unsafe` aspects of this are acknowledged by the user rather than just papered over by a wrapper from PyO3 which has invariants we can't guarantee are maintained.

davidhewitt: (`get_context` may need an equivalent change to `*mut void` too, I guess?)

davidhewitt:
👍 this is shaping up really well, just wanted to say thanks again for working on this. It's a tricky one with a few design gotchas which are taking a bit of effort to hash out, so the patience with repeated iterations is appreciated. I think with luck we're close now 🤞

src/pycapsule.rs
birkenfeld: Agreed, let's do it that way.

src/pycapsule.rs
milesgranger: Cool! Added with https://github.com/PyO3/pyo3/pull/1980/commits/7a87a15760bedeeffb0fdaf47bc117a9f3548289, see what you think.

src/pycapsule.rs
```diff
@@ -0,0 +1,417 @@
...
+        } else {
+            Ok(&*(ptr as *const T))
+        }
+    }
+
+    /// Sets the context pointer in the capsule.
+    ///
+    /// # Notes
+    ///
+    /// The default destructor does not drop the context value. If you need to ensure it is not
```
birkenfeld, line 141: These two paragraphs (this one and "Finally...") are now out of date again :( Suggestion:

```
If you want to store a Rust value as the context, and drop it from the destructor, use `Box::into_raw` to convert it into a pointer, see the example.
```

src/pycapsule.rs
```diff
@@ -0,0 +1,417 @@
...
+    ///     let name = CString::new("foo").unwrap();
+    ///     let capsule =
+    ///         PyCapsule::new_with_destructor(py, 123, &name, destructor as fn(u32, *mut c_void))
+    ///             .unwrap();
+    ///     let context = Box::new(tx);  // `Sender<String>` is our context, box it up and ship it!
+    ///     capsule.set_context(py, Box::into_raw(context) as *mut c_void).unwrap();
+    ///     // This scope will end, causing our destructor to be called...
+    /// });
+    ///
+    /// assert_eq!(rx.recv(), Ok("Destructor called!".to_string()));
```
birkenfeld, line 179: That's a cool example!

src/pycapsule.rs
```diff
@@ -0,0 +1,417 @@
...
+) {
+    let ptr = ffi::PyCapsule_GetPointer(capsule, ffi::PyCapsule_GetName(capsule));
+    let ctx = ffi::PyCapsule_GetContext(capsule);
+    let CapsuleContents { value, destructor } = *Box::from_raw(ptr as *mut CapsuleContents<T, F>);
+    destructor(value, ctx)
+}
+
+/// Guarantee `T` is not zero sized at compile time.
+/// credit: https://users.rust-lang.org/t/is-it-possible-to-assert-at-compile-time-that-foo-t-is-not-called-with-a-zst/67685
+pub trait PanicWhenZeroSized: Sized {
```
birkenfeld, line 249: I think this name is a bit nicer:
```suggestion
#[doc(hidden)]
pub trait AssertNotZeroSized: Sized {
```

src/pycapsule.rs
```diff
@@ -0,0 +1,417 @@
...
+    const _CONDITION: usize = (std::mem::size_of::<Self>() == 0) as usize;
+    const _CHECK: &'static str = ["type must not be zero-sized!"][Self::_CONDITION];
```
birkenfeld, line 251:
```suggestion
    const _CHECK: &'static str = ["PyCapsule value type T must not be zero-sized!"][Self::_CONDITION];
```

src/pycapsule.rs
```diff
@@ -0,0 +1,417 @@
...
+pyobject_native_type_core!(PyCapsule, ffi::PyCapsule_Type, #checkfunction=ffi::PyCapsule_CheckExact);
+
+impl PyCapsule {
+    /// Constructs a new capsule whose contents are `value`, associated with `name`.
+    /// `name` is the identifier for the capsule; if it is stored as an attribute of a module,
+    /// the name should be in the format `modulename.attribute`.
+    ///
+    /// # Panics
+    ///
+    /// An attempt to add a zero sized value will panic.
```
birkenfeld, line 54: This is now outdated.

src/pycapsule.rs
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    /// Constructs a new capsule whose contents are `value`, associated with `name`.
+    ///
+    /// Also provides a destructor: when the `PyCapsule` is destroyed, it will be passed the original object,
+    /// as well as `*mut c_void` which will point to the capsule's context, if any.
+    ///
+    /// # Panics
+    ///
+    /// An attempt to add a zero sized value will panic.
```
birkenfeld, line 96: This is now outdated.

src/pycapsule.rs
```diff
@@ -0,0 +1,417 @@
...
+#[repr(transparent)]
+pub struct PyCapsule(PyAny);
+
+pyobject_native_type_core!(PyCapsule, ffi::PyCapsule_Type, #checkfunction=ffi::PyCapsule_CheckExact);
+
+impl PyCapsule {
+    /// Constructs a new capsule whose contents are `value`, associated with `name`.
+    /// `name` is the identifier for the capsule; if it is stored as an attribute of a module,
+    /// the name should be in the format `modulename.attribute`.
+    ///
```
birkenfeld, line 51: add something like:

It is checked at compile time that the type `T` is not zero-sized.
Rust function items need to cast to a function pointer (`fn(args) -> result`)
to be put into a capsule.

birkenfeld (commented):
Only a few nits left :)

src/pycapsule.rs
```diff
@@ -147,6 +147,10 @@ impl PyCapsule {
     /// Context itself, is treated much like the value of the capsule, but should likely act as
     /// a place to store any state managment when using the capsule.
     ///
+    /// # Safety
+    ///
+    /// This might dereference the raw pointer `context`, which is unsafe.
```
birkenfeld, line 152: Does it? How?

milesgranger: Good question. :sweat_smile: 

I tried to place the unsafe block within the method, but clippy doesn't agree:
```
error: this public function might dereference a raw pointer but is not marked `unsafe`
   --> src/pycapsule.rs:182:72
    |
182 |         let result = unsafe { ffi::PyCapsule_SetContext(self.as_ptr(), context) as u8 };
    |                                                                        ^^^^^^^
    |
    = note: `#[deny(clippy::not_unsafe_ptr_arg_deref)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref

error: could not compile `pyo3` due to previous error
```

birkenfeld: I think clippy is pulling that out of its steely behind, so yes, please ignore it. No `unsafe` needed on this side.

src/pycapsule.rs
```diff
@@ -0,0 +1,414 @@
...
+            Ok(ctx)
+        }
+    }
+
+    /// Obtains a reference to the value of this capsule.
+    ///
+    /// # Safety
+    ///
+    /// It must be known that this capsule's pointer is to an item of type `T`.
+    pub unsafe fn reference<T>(&self) -> &T {
```
messense, line 197: Should this be called `as_ref` and add a `as_mut` companion if applicable? ([`NonNull`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.as_ref) does something like this).

mejrs: I'm inclined to agree with https://github.com/PyO3/pyo3/pull/1980#discussion_r753810681 and just not have this function - leave it up to the user to handle the pointers.

milesgranger: I'm not heavily bound to it, however, it seems if basically every use of fetching the value is `&*(self.get_pointer() as *const T)`, why not make it slightly easier on the user/reader? If they'd like to treat it differently, `PyCapsule::get_pointer` is available.

milesgranger: > Should this be called as_ref and add a as_mut companion if applicable? (NonNull does something like this).

No objection to changing it, if that's the consensus.

birkenfeld: I vote for keeping the `reference` name and returning a `&T` - `as_ref` usually is a cheap pointer conversion and not unsafe.
`as_mut` is not safe in any case once the capsule is created, since the capsule can be shared like other PyObjects.

In order to be able to also use capsules with arbitrary pointers, `get_pointer` is also needed. Maybe it should be named just `pointer` for consistency.

milesgranger: Sounds good to me. Have updated `get_pointer` -> `pointer`

davidhewitt:
@birkenfeld as I think you're very much on top of this PR, do you want to approve, squash-and-merge it once you're happy? Thanks for comprehensive reviewing, and thanks @milesgranger for all the iterations put in! 35 commits 😮

milesgranger:
It's been a pleasure, learned a lot, and really appreciate the repeated and helpful reviews! :pray:

birkenfeld:
@davidhewitt will do - should the module be renamed to `capsule` though (to be consistent with `buffer` which is also a CPython type, as opposed to `pycell`)?

@milesgranger I'll do this change if @davidhewitt agrees and then merge it, everything else LGTM!

davidhewitt:
Ah yes good spot, it should probably even live in `src/types/capsule.rs` and be exposed as `types::PyCapsule` ?

birkenfeld:
That's probably even better. Not sure why `buffer` is so lonely at toplevel :D

davidhewitt:
`PyBuffer` is this weird thing that isn't really a Python object, so it's kind of lost wandering the wasteland of murky design 😅

birkenfeld (approved):

birkenfeld:
Thanks again @milesgranger!

PR #1981, opened 2021-11-10
add doc_auto_cfg feature

mejrs:
This makes the feature flag thingies render again.

See https://github.com/rust-lang/rust/pull/90502


davidhewitt (approved):
Thanks for keeping on top of these developments!

PR #1985, opened 2021-11-11
add PyType::is_subclass_of and PyAny::is_instance_of

### LINKED ISSUE #1984: `PyType.is_subclass` does not work with two dynamic `PyType`s
Hello! During Pants's [migration to PyO3](https://github.com/pantsbuild/pants/pull/13526), I'm stuck on how to get `&PyType.is_subclass()` working.

This is what I would like to do:

```rust
#[pyfunction]
fn check_subclass(typ1: &PyType, typ2: &PyType) -> PyResult<bool> {
  typ1.is_subclass(typ2)
}
```

If I understand correctly, this is not possible because the API is `is_subclass<T>(&self)`, and it doesn't seem possible to say this due to how Generics work in Rust:

```rust
typ1.is_subclass::<typ2>()
```

For now, I'm thinking I can work around this by writing our own `unsafe` implementation, but I'd love to be able to contribute a more flexible API to PyO3. I'm thinking `is_subclass(&self, T: ToPyObject) -> bool`.

What do you think? How would we handle deprecations? (Note that rust-cpython calls it `is_subtype_of`, maybe that's an option)

Also I'm curious what you think about it returning `bool` instead of `PyResult<bool>`?

birkenfeld:
which get the type to check against as an arguments, as opposed to a compile-time generic type.

Also added some tests for the instance/subclass methods.

**Question:** is it better to name the new methods like this with `_of`, or to move the current methods to `_of` and name the new methods with the conventional arguments more like in Python?

This would also fix the current (slight) inconsistency between `PyType::is_instance` and `PyAny::is_instance_of`.

Fixes #1984

### DIFF
--- a/guide/src/exception.md
+++ b/guide/src/exception.md
@@ -100,21 +100,19 @@ PyErr::from_instance(py, err).restore(py);
 ## Checking exception types
 
 Python has an [`isinstance`](https://docs.python.org/3/library/functions.html#isinstance) method to check an object's type.
-In PyO3 every native type has access to the [`PyAny::is_instance`] method which does the same thing.
+In PyO3 every object has the [`PyAny::is_instance`] and [`PyAny::is_instance_of`] methods which do the same thing.
 
 ```rust
 use pyo3::Python;
 use pyo3::types::{PyBool, PyList};
 
 Python::with_gil(|py| {
-    assert!(PyBool::new(py, true).is_instance::<PyBool>().unwrap());
+    assert!(PyBool::new(py, true).is_instance_of::<PyBool>().unwrap());
     let list = PyList::new(py, &[1, 2, 3, 4]);
-    assert!(!list.is_instance::<PyBool>().unwrap());
-    assert!(list.is_instance::<PyList>().unwrap());
+    assert!(!list.is_instance_of::<PyBool>().unwrap());
+    assert!(list.is_instance_of::<PyList>().unwrap());
 });
 ```
-[`PyAny::is_instance`] calls the underlying [`PyType::is_instance`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyType.html#method.is_instance)
-method to do the actual work.
 
 To check the type of an exception, you can similarly do:
 
@@ -123,7 +121,7 @@ To check the type of an exception, you can similarly do:
 # use pyo3::prelude::*;
 # Python::with_gil(|py| {
 # let err = PyTypeError::new_err(());
-err.is_instance::<PyTypeError>(py);
+err.is_instance_of::<PyTypeError>(py);
 # });
 ```
 
@@ -184,7 +182,7 @@ fn main() {
     Python::with_gil(|py| {
         let fun = pyo3::wrap_pyfunction!(connect, py).unwrap();
         let err = fun.call1(("0.0.0.0",)).unwrap_err();
-        assert!(err.is_instance::<PyOSError>(py));
+        assert!(err.is_instance_of::<PyOSError>(py));
     });
 }
 ```
@@ -201,21 +199,21 @@ fn parse_int(s: String) -> PyResult<usize> {
 }
 #
 # use pyo3::exceptions::PyValueError;
-# 
+#
 # fn main() {
 #     Python::with_gil(|py| {
 #         assert_eq!(parse_int(String::from("1")).unwrap(), 1);
 #         assert_eq!(parse_int(String::from("1337")).unwrap(), 1337);
-# 
+#
 #         assert!(parse_int(String::from("-1"))
 #             .unwrap_err()
-#             .is_instance::<PyValueError>(py));
+#             .is_instance_of::<PyValueError>(py));
 #         assert!(parse_int(String::from("foo"))
 #             .unwrap_err()
-#             .is_instance::<PyValueError>(py));
+#             .is_instance_of::<PyValueError>(py));
 #         assert!(parse_int(String::from("13.37"))
 #             .unwrap_err()
-#             .is_instance::<PyValueError>(py));
+#             .is_instance_of::<PyValueError>(py));
 #     })
 # }
 ```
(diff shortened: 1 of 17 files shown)


src/types/typeobject.rs
```diff
@@ -53,12 +54,55 @@ impl PyType {
         Ok(result == 1)
     }
 
+    /// Checks whether `self` is a subclass of `other`.
+    ///
+    /// Equivalent to the Python expression `issubclass(self, other)`.
+    pub fn is_subclass_of(&self, other: &PyType) -> PyResult<bool> {
+        let result = unsafe { ffi::PyObject_IsSubclass(self.as_ptr(), other.as_ptr()) };
+        err::error_on_minusone(self.py(), result)?;
+        Ok(result == 1)
```
Eric-Arellano, line 63: Maybe DRY `is_subclass` to call this?

src/types/typeobject.rs
```diff
@@ -53,12 +54,55 @@ impl PyType {
...
+    }
+
     /// Check whether `obj` is an instance of `self`.
     ///
-    /// Equivalent to Python's `isinstance` function.
+    /// Equivalent to the Python expression `isinstance(obj, self)`.
     pub fn is_instance<T: AsPyPointer>(&self, obj: &T) -> PyResult<bool> {
```
Eric-Arellano, line 69: Should this also have `is_instance_of`? Ack that `PyAny` already has that

Eric-Arellano (approved):
Yay! Thank you!

If we temporarily disregard backwards compatibility, I wonder if there's still value in keeping `is_instance`? It's only for boilerplate reduction, right? If so, I wonder how common it is to know the types at compile time.

src/types/typeobject.rs
birkenfeld: If we kept the names consistent, this would be `is_instance_of`. `is_instance` is also possible, it would be a static method.

birkenfeld:
> If we temporarily disregard backwards compatibility, I wonder if there's still value in keeping `is_instance`? It's only for boilerplate reduction, right? If so, I wonder how common it is to know the types at compile time.

It's quite convenient if you're checking against your own `pyclass` types. Otherwise you need the `PyClass::type_object(py)` which is a bit clumsy.

But since we can still break compat, it would be also fine (IMO) to exchange the names.

src/types/typeobject.rs
birkenfeld: Yeah I'll put that in the next rev.

Eric-Arellano:
> But since we can still break compat, it would be also fine (IMO) to exchange the names.

Imo, that makes sense. I was surprised that `is_subclass()` did not take any args because it conflicted with my mapping to Python.

This reads naturally to me:

```rust
typ1.is_subclass(typ2)
type.is_subclass_of::<MyType>()
```

While I don't love breaking backwards compatibility, I agree it's less offensive to do so now than to have an unclear API for all future developers, given how much PyO3 seems to be picking up steam.

src/types/typeobject.rs
davidhewitt: I wonder if we should just throw this method away entirely? `obj.is_instance(other)` and `type.is_subclass(base)` work in the same order `isinstance(obj, other)`, whereas this `type.is_instance(obj)` method is the reversed `isinstance(obj, type)`.

Seems like a great way to introduce confusion and bugs, so I'd be up for deprecating (can point users to `PyAny::is_instance`).

davidhewitt (commented):
👍 this is definitely a necessary addtion, thanks! In particular actually for PEP 489 (#1982) I suspect we may need to remove the static type objects, so having runtime arguments may become necessary eventually.

Regarding the names - I don't feel strongly either way with which way round `is_instance_of` and `is_instance` should be for generic vs dynamic form. I'm inclined to call the new ones `is_instance_of` and `as_subclass_of` as already proposed in this PR and avoid making a breaking change given there's not a clear advantage of making the breaking change IMO.

Needs a CHANGELOG entry 😉

src/types/typeobject.rs
Eric-Arellano: Oh yeah, that's a good point...In Python, you're meant to use `issubclass()` for types, `isinstance()` for objects. So I agree with deprecating this method on `PyType` - it's too confusing of sugar for something you already can achieve.

Eric-Arellano:
> In particular actually for PEP 489 (#1982) I suspect we may need to remove the static type objects, so having runtime arguments may become necessary eventually.

Oh cool!!

> I'm inclined to call the new ones is_instance_of and as_subclass_of as already proposed in this PR and avoid making a breaking change given there's not a clear advantage of making the breaking change IMO.

Given that we may need to remove the static implementations for PEP 489, my two cents is to name these methods what you want them to be after that removal happens in the future. I imagine you'd prefer `is_instance` over `is_instance_of` - the `_of` would be unnecessary because there is no ambiguity what it refers to anymore. 

That way, you're not going to be adding a ticking deprecation time bomb for users who use the more correct implementation (this PR) from the start.

--

One thing I don't know though is PyO3s approach to deprecations, if you are willing to make completely breaking changes in a release or if you have to have a deprecation cycle first. The latter would complicate this all.

birkenfeld:
> +1 this is definitely a necessary addtion, thanks! In particular actually for PEP 489 (#1982) I suspect we may need to remove the static type objects, so having runtime arguments may become necessary eventually.

Hm, I suspect you'll still need access to it, so either `type_object()` will need to do some thread-local magic to figure out its current interpreter, or the interpreter info needs to be in the `py` token?

birkenfeld:
Ok, review comments addressed, changelog entry added and `PyType::is_instance` removed. I'm still slightly in favor of switching the names.

Eric-Arellano (approved):
>  I'm still slightly in favor of switching the names.

Looks great! I agree with you here though.

davidhewitt:
👍 Ok go for it, let's switch the names. (This can go as a part of 0.16 I guess.)

> Hm, I suspect you'll still need access to it, so either type_object() will need to do some thread-local magic to figure out its current interpreter, or the interpreter info needs to be in the py token?

Yes that's definitely possible. It's a bit complex so I've buried my head in the sand about it so far! 😂 

> One thing I don't know though is PyO3s approach to deprecations, if you are willing to make completely breaking changes in a release or if you have to have a deprecation cycle first. The latter would complicate this all.

We don't have any hard restrictions; I like it if we can give users a helpful migration pathway. Deprecations and compile errors are both ways to achieve that. Here it isn't too hard to figure out what to do to update especially if we write a helpful CHANGELOG message.

birkenfeld:
So this is updated; I also missed `PyErr::is_instance` at first which is now handled as well.

guide/src/exception.md
```diff
@@ -123,7 +121,7 @@ To check the type of an exception, you can similarly do:
 # use pyo3::prelude::*;
 # Python::with_gil(|py| {
 # let err = PyTypeError::new_err(());
-err.is_instance::<PyTypeError>(py);
```
davidhewitt, line 126: Ah, just been reminded from this that we probably need to make equivalent changes to `PyErr`?

Either that or we could deprecate `Err::is_instance()` too (and advise just using `err.pvalue(py).is_instance_of::<PyTypeError>()`).

davidhewitt (changes requested):

davidhewitt (approved):

Eric-Arellano (approved):
👏

davidhewitt:
I'll rebase this and then merge it. Thanks!

PR #1988, opened 2021-11-12
err: tidy up some uses of PyErr::occurred

davidhewitt:
With new `PyErr::take` API, I just noticed a few bits that could be made slightly more efficient.


birkenfeld:
LGTM.

PR #1990, opened 2021-11-14
allow_threads: switch from `catch_unwind` to guard pattern

davidhewitt:
See https://github.com/pantsbuild/pants/pull/13526#issuecomment-968248522

I think that use of `catch_unwind` in `allow_threads` may currently catch non-Rust exceptions too (it's not particularly clear), leading to nasty crashes.

Switching to a guard pattern might avoid this.

cc @Eric-Arellano can you try your pants PR against this branch and see if it fixes?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix creating `#[classattr]` by functions with the name of a known magic method. [#1969](https://github.com/PyO3/pyo3/pull/1969)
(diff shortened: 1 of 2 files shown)


Eric-Arellano:
Yay this fixes the two problems I was seeing! 1) The crash on Linux, and 2) the async thread was not actually finishing on macOS. Thank you so much 🙌 Everything now works without needing to make any more changes to Pants, woot!

Thanks to Cargo's git support, once this merges, I think Pants is fully unblocked now from hitting merge on https://github.com/pantsbuild/pants/pull/13526 🚀

davidhewitt:
Great! I was thinking of putting out 0.15.1 within a week anyway to include a few non-breaking additions before starting breaking stuff for 0.16, so I'll make sure to include this in that release.

PR #1991, opened 2021-11-15
Don't emit `Py_LIMITED_API` cfg for PyPy

messense:
See https://github.com/PyO3/maturin/pull/678#issuecomment-968312293

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -219,7 +219,12 @@ print("mingw", get_platform().startswith("mingw"))
         let implementation = map["implementation"].parse()?;
 
         let lib_name = if cfg!(windows) {
-            default_lib_name_windows(version, abi3, map["mingw"].as_str() == "True")
+            default_lib_name_windows(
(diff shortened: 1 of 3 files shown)


pyo3-build-config/src/impl_.rs
```diff
@@ -219,7 +219,12 @@ print("mingw", get_platform() == "mingw")
         let implementation = map["implementation"].parse()?;
 
         let lib_name = if cfg!(windows) {
-            default_lib_name_windows(version, abi3, map["mingw"].as_str() == "True")
+            default_lib_name_windows(
+                version,
+                implementation,
+                abi3,
+                map["mingw"].as_str() == "True",
```
messense, line 226: According to https://github.com/PyO3/maturin/issues/676#issuecomment-967083567 , `sysconfig.get_platform()` returns `mingw_x86_64` instead of just `mingw` in L203.

davidhewitt (approved):
I think this is also worth a "Fixed" CHANGELOG entry 😄

PR #1993, opened 2021-11-15
Fix mingw platform detection

messense:
According to https://github.com/PyO3/maturin/issues/676#issuecomment-967083567 , `sysconfig.get_platform()` returns `mingw_x86_64` instead of just `mingw`.

See also https://github.com/msys2/MINGW-packages/blob/c95de6c2045def876948c20fe6185faccff3f576/mingw-w64-python/0083-Change-the-get_platform-method-in-sysconfig.patch


davidhewitt (approved):
👍 I think this is worth a "Fixed" CHANGELOG entry please!

PR #1995, opened 2021-11-15
add missing annotation to PyCounter.__call__

gertjanvanzwieten:
This patch annotates the `__call__` method of `PyCounter` in example: callable objects with `#[args(args="*", kwargs="**")]`. Without it (at least in PyO3 0.15.0) the example fails with `TypeError: counter.__call__() missing 1 required positional argument: 'args'`.


davidhewitt (approved):
Thanks!

Issue #1628, opened 2021-05-25
Better `pyo3-build-config` example


davidhewitt:
See https://github.com/PyO3/pyo3/pull/1622#issuecomment-847568063

Issue #1627, opened 2021-05-25
Expose `pyo3-build-config` APIs for maturin


davidhewitt:
See https://github.com/PyO3/pyo3/pull/1622#discussion_r636678392

aganders3:
I'm interested in taking this on, but I'm not sure I completely understand. I get that the current method of getting the config using `cargo_env_var` is not suitable for maturin, but I also don't see maturin depending on pyo3 or pyo3-build-config to call this function directly.

It seems there are a few possible ways to handle this in pyo3:
* refactor `cross_compiling` to take the "target triple" as args (and I suspect move the `cargo_env_var` parts to another function)
* refactor `cross_compiling` to optionally take these args (else check the Cargo env vars as it does now)
* refactor `cross_compliling` to look for non-Cargo env vars (e.g. `PYO3_TARGET_ARCH`), falling back to the Cargo env vars

It would be nice to pair this with a PR in maturin for whatever changes would be needed there. I am happy to be the implementer there as well but will definitely need help figuring out where that might go.

davidhewitt:
Some help with this would be awesome!

> refactor `cross_compiling` to take the "target triple" as args (and I suspect move the cargo_env_var parts to another function)

That is the design that I was thinking. There's also a couple of other functions that would be nice to expose (see below).

> I also don't see maturin depending on pyo3 or pyo3-build-config to call this function directly.

> It would be nice to pair this with a PR in maturin for whatever changes would be needed there.

Yep, so at the moment `maturin` doesn't have a dependency, however the idea is that it should be able to depend on `pyo3-build-config` (with `default-features = false`).

If you look at the following file, you'll recognise pretty much all of it is logic that also lives in `pyo3-build-config/src/impl_.rs`.

https://github.com/PyO3/maturin/blob/main/src/cross_compile.rs

That file was basically a copypasta of bits of PyO3's build script some time ago before we made `pyo3-build-config` into a separate crate. So the goal of this issue would be to add sufficient pub functions to `pyo3-build-config` so that `maturin` could re-use as much of that logic as possible.

I'm happy to help feed back on the other functions we'd need if you'd like pointers on those too.

aganders3:
Thanks - this is super helpful and seems like enough for me to get started. I'll come back here when I have more questions.

aganders3:
Okay - I'm just documenting my process/progress here, but please let me know if I'm off course. Otherwise I will work on a PR next.

[...]

PR #1996, opened 2021-11-15
Expose cross compiling configuration from pyo3-build-config

aganders3:
These modifications were proposed in #1627.

So far this exposes three functions:
* `cross_compiling` (altered to take arguments, with current behavior moved to `cross_compiling_from_cargo_env`)
* `find_sysconfigdata`
* `parse_sysconfigdata`

and modifies `InterpreterConfig` with three new fields:
* `abi_flags`
* `abi_tag`
* `ext_suffix`

The refactored `cross_compiling` takes `host` (triple), `arch`, `vendor`, and `os`, but `vendor` is not currently used separately from what I see in maturin. It might be better to change it to take only the complete "target triple" to simplify things, but splitting `os` and `arch` from there for `CrossCompileConfig` may not be straightforward as the overall format seems a bit complicated (i.e. not always only three parts). Maturin uses another crate (`target_lexicon`) for this, which could be pulled in here so it could perhaps take a `target_lexicon::Triple`. This might be the best option for supporting the most platforms.

I'm also not sure I have the correct behavior for the fields added to `InterpreterConfig` - specifically I'm not sure if the defaults are handled correctly. Also, do they need to change depending on the value of `abi3`?

It seems there's more in pyo3-build-config that could be exposed so I'm happy to bundle that in this PR or work on the different features separately.

TODO: 
 - [x] docs to all new functions and / or detail in the guide (possibly related to #1628, #1976)


pyo3-build-config/src/impl_.rs
```diff
@@ -864,7 +952,7 @@ fn ends_with(entry: &DirEntry, pat: &str) -> bool {
 /// ```
 ///
 /// [1]: https://github.com/python/cpython/blob/3.5/Lib/sysconfig.py#L389
-fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
+pub fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
```
davidhewitt, line 955: I would probably refactor this function a bit further, so that the `find_sysconfigdata` function which we share with `maturin` just returns sorted / deduplicated / canonicalized `Vec<PathBuf>`. 

The error messages for empty or too many results are pretty specific to the pyo3 build script.

pyo3-build-config/src/lib.rs
```diff
@@ -15,7 +15,12 @@ use std::io::Cursor;
 #[cfg(feature = "resolve-config")]
 use once_cell::sync::OnceCell;
 
-pub use impl_::{BuildFlag, BuildFlags, InterpreterConfig, PythonImplementation, PythonVersion};
+pub use impl_::{
+    BuildFlag, BuildFlags, CrossCompileConfig, InterpreterConfig, PythonImplementation,
+    PythonVersion,
+};
+
+pub use impl_::{cross_compiling, find_sysconfigdata, parse_sysconfigdata};
```
davidhewitt, line 23: I think these can just be a single `pub use` rather than splitting the functions tbh.

davidhewitt (commented):
Looking good! Few comments below.

I try to avoid adding dependencies to pyo3 and sub-crates as much as possible. It looks like `target_lexicon` is itself a simple dependency with no other dependencies, so it's relatively lightweight. Which platforms do you feel we're lacking support for which `target_lexicon` could help with?

pyo3-build-config/src/impl_.rs
```diff
@@ -486,47 +549,58 @@ pub fn any_cross_compiling_env_vars_set() -> bool {
...
 
-        let target = cargo_env_var("TARGET").unwrap();
-        let host = cargo_env_var("HOST").unwrap();
-        if target == host {
-            // Not cross-compiling
-            return Ok(None);
-        }
-
-        if target == "i686-pc-windows-msvc" && host == "x86_64-pc-windows-msvc" {
+        if target_triple == "i686-pc-windows-msvc" && host == "x86_64-pc-windows-msvc" {
```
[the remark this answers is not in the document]
aganders3: For now I think this check could instead use `starts_with` and a decision on using `target_lexicon` can be put off.

I understand the hesitation on adding a dependency, especially without a concrete example of new functionality it would add.

aganders3:
Thanks for the quick feedback! I'll work on these suggestions.

pyo3-build-config/src/impl_.rs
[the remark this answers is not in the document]
aganders3: Hm, that still misses the `msvc` part though which is probably important for this check. That specific check could also move back to the `cross_compiling_from_cargo_env` function.  I'll give this a bit more thought.

pyo3-build-config/src/impl_.rs
```diff
@@ -486,47 +549,58 @@ pub fn any_cross_compiling_env_vars_set() -> bool {
...
+    if target == "i686-pc-windows-msvc" && host == "x86_64-pc-windows-msvc" {
+        // Not cross-compiling to compile for 32-bit Python from windows 64-bit
+        return Ok(None);
+    }
+
+    let target_arch = cargo_env_var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let target_vendor = cargo_env_var("CARGO_CFG_TARGET_VENDOR").unwrap();
+    let target_os = cargo_env_var("CARGO_CFG_TARGET_OS").unwrap();
+
+    cross_compiling(&host, &target_vendor, &target_arch, &target_os)
```
aganders3, line 570: Do we know that `TARGET` and `CARGO_CFG_TARGET_*` vars will always match?

messense:
Thank you for helping with this. Just a reminder, this branch has conflicts that must be resolved.

pyo3-build-config/src/impl_.rs
```diff
@@ -1560,6 +1675,9 @@ mod tests {
             parsed_config,
             InterpreterConfig {
                 abi3: false,
+                abi_flags: Some("m".to_string()),
+                abi_tag: Some("37m".to_string()),
+                ext_suffix: Some(".cpython-37m-x86_64-linux-gnu.so".to_string()),
```
aganders3, line 1680: I think this `assert_eq` needs to change to be successful even when these fields don't match exactly, as per the note above:
>  // A best effort attempt to get test coverage for the sysconfigdata parsing.
// Might not complete successfully depending on host installation; that's ok as long as
// CI demonstrates this path is covered!

aganders3: Oh I see what's happening - this test is conditional for x86_64 linux but I don't know how to make it conditional on Python version in CI.

messense:
FYI, we recently merged https://github.com/PyO3/maturin/pull/677 , it introduced new dependency on `sysconfig.get_platform()` of Python.

aganders3:
> FYI, we recently merged [PyO3/maturin#677](https://github.com/PyO3/maturin/pull/677) , it introduced new dependency on `sysconfig.get_platform()` of Python.

Hm, thanks for the heads up. This is already called by `InterpreterConfig::from_interpreter` but `parse_sysconfigdata` does not necessarily use the target interpreter.

pyo3-build-config/src/impl_.rs
```diff
@@ -739,7 +810,7 @@ fn parse_script_output(output: &str) -> HashMap<String, String> {
 /// The sysconfigdata is simply a dictionary containing all the build time variables used for the
 /// python executable and library. Here it is read and added to a script to extract only what is
 /// necessary. This necessitates a python interpreter for the host machine to work.
-fn parse_sysconfigdata(sysconfigdata_path: impl AsRef<Path>) -> Result<InterpreterConfig> {
+pub fn parse_sysconfigdata(sysconfigdata_path: impl AsRef<Path>) -> Result<InterpreterConfig> {
```
davidhewitt, line 813: Hmm, I hate to be "that guy" who suggests a whole load of changes after you've done all this work. This just struck me and it seems really nice.

What if the signature of `parse_sysconfigdata` is 

```
pub fn parse_sysconfigdata(sysconfigdata_path: impl AsRef<Path>) -> Result<Sysconfigdata>

// where Sysconfigdata is just a wrapper around a hashmap:
struct Sysconfigdata(Hashmap<String, String>);
```

We could then have `Sysconfigdata::get_value(key: &str) -> Option<&str>` and `Interpreterconfig::from_sysconfigdata(sysconfigdata: Sysconfigdata) -> Result<InterpreterConfig>`.

Couple of reasons I suggest this:
 - PyO3 doesn't need to care about `abi_flags`, `abi_tag` and `ext_suffix` in the `InterpreterConfig` struct or its config files. Especially it seems slightly weird to have these in the config files even though they don't affect PyO3's build at all.
 - `maturin` can get those vars (and any others it wants in the future) directly from `Sysconfigdata.get_value`. Especially good is that then `maturin` doesn't need PyO3 to change if it wants extra sysconfigdata values in the future.
 - `IntepreterConfig::from_sysconfigdata` has a nice match with `InterpreterConfig::from_interpreter`

I'm sorry to suggest this so late 😢

pyo3-build-config/src/impl_.rs
davidhewitt: I think they will, yes.

pyo3-build-config/src/impl_.rs
aganders3: Hey no worries - you obviously know much more about the project so I appreciate the input and guidance. I'm not just trying to rack up merged PRs so I'm happy to take the time to get the interface right. Doubly so on a public API subject to SemVer, etc. I am learning along the way and that (along with helping) is my goal. Hopefully I am not holding things up by taking more time.

Anyway - this makes sense! I think the hashmap is basically what maturin uses now but without the `Sysconfigdata` wrapper. The simple struct basically says (to me) "this is a dict where you can make a pretty good guess about the keys, but we make no specific guarantees" which seems like a good fit for this file which can be quite long and varied.

davidhewitt: Great! Your contributions are warmly welcomed, you're not holding things up at all. Contributors are the lifeblood of OSS and if PyO3 will carry on should I ever be unable to keep working on it then a healthy group of knowledgeable contributors is even more important. (My intention is to continue with PyO3 until Rust and Python are dead languages, though I'm going to be a dad soon and you never know what life throws at you.) There's also way more things we want to improve in PyO3 than I as a single person have time for. People take time to familiarize with the project; no question is too silly!

I simply respect your time too and if I can suggest design ideas which influence a PR before too much work is spent on it, then I much prefer to do that! 🙂

aganders3: Well thanks - I think we're on the same page! I'll work on these changes next.

aganders3:
I know I definitely still have a couple tests to write, but I'm a bit confused by the codecov results. Is there a good way for me to run this locally?

Otherwise I think this refactor definitely provides a better API now than before, so thanks for the suggestion! I'm not sure if the `parse_sysconfigdata` is what you had in mind, as this will parse the whole file where before there was a list of keys to keep.

davidhewitt:
If you go into the `pyo3-build-config` subdirectory you should be able to run `cargo test` to exercise the tests in these files? Or is there something more specific you want to run.

This is definitely looking great, let me know when you want a re-review. (I think there will only be minor nits left tbh, nice work.)

> this will parse the whole file where before there was a list of keys to keep

Yes indeed I didn't state it but I was assuming that we would parse (and keep) the whole file contents 👍

davidhewitt:
(#2011 may be relevant if you're getting weird coverage results.)

aganders3:
I wrote a few more tests to improve coverage (not pushed), but the cases where it _is_ cross compiling all rely on environment variables, which I am having a hard time controlling for tests. I can set/unset in each test, but the default for `cargo test` is to run in parallel which causes intermittent failures. Do you have any suggestions?

Aside from that I am mostly happy with the interface and would appreciate feedback when you have time to review. I don't have a great feeling for whether `cross_compiling` should take `AsRef<str>` or `Into<String>` instead of `&str` for some of its args, for example. It seems maybe it should be something like this, because the host can always be borrowed but the target values need to become owned by the CrossCompileConfig if cross-compiling:
```rust
pub fn cross_compiling(
    host: AsRef<str>,
    target_arch: Into<String>,
    target_vendor: Into<String>,
    target_os: Into<String>,
) -> Result<Option<CrossCompileConfig>>
```

davidhewitt:
👍 I'll give this a full review in the morning, thanks!

aganders3:
> 👍 I'll give this a full review in the morning, thanks!

Much appreciated - no rush!

davidhewitt (approved):
Thank you, this is looking excellent!

I have a few minor tidy ups commented below, really just nits though.

I agree also that it's probably worth having a whirl with using this as a git dependency with `maturin` (and maybe even opening that branch as a PR on the maturin repo) to confirm that we're happy with how this works downstream before we merge.

pyo3-build-config/src/impl_.rs
```diff
@@ -485,47 +573,61 @@ pub fn any_cross_compiling_env_vars_set() -> bool {
...
+/// Detect whether we are cross compiling and return an assembled CrossCompileConfig if so.
+pub fn cross_compiling(
+    host: &str,
+    target_arch: &str,
+    target_vendor: &str,
+    target_os: &str,
+) -> Result<Option<CrossCompileConfig>> {
     let cross = env_var("PYO3_CROSS");
     let cross_lib_dir = env_var("PYO3_CROSS_LIB_DIR");
     let cross_python_version = env_var("PYO3_CROSS_PYTHON_VERSION");
```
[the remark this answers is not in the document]
messense: It's fine to leave it here.

pyo3-build-config/src/impl_.rs
```diff
@@ -32,7 +32,9 @@ pub fn cargo_env_var(var: &str) -> Option<String> {
 /// Gets an external environment variable, and registers the build script to rerun if
 /// the variable changes.
 pub fn env_var(var: &str) -> Option<OsString> {
-    println!("cargo:rerun-if-env-changed={}", var);
+    if cfg!(feature = "resolve-config") {
+        println!("cargo:rerun-if-env-changed={}", var);
+    }
```
aganders3, line 37: @davidhewitt let me know if this is problematic or if there's a better way to suppress this output in maturin.

aganders3:
Okay https://github.com/PyO3/maturin/pull/718 seems good to go and I think this is ready for another look. I will make a few inline comments for areas I have questions but of course appreciate any additional feedback

Merging probably needs to wait for https://github.com/PyO3/pyo3/pull/2040 but otherwise hopefully we can close this out soon!

pyo3-build-config/src/impl_.rs
```diff
@@ -826,7 +920,7 @@ fn ends_with(entry: &DirEntry, pat: &str) -> bool {
 /// So we must find the file in the following possible locations:
 ///
 /// ```sh
-/// # distribution from package manager, lib_dir should include lib/
+/// # distribution from package manager, (lib_dir may or may not include lib/)
```
aganders3, line 923: Hopefully this is okay - see implementation below. This was requested in Maturin to support cross-compiling with PyPy, where the `_sysconfigdata*.py` is not under the `lib` dir.

messense:
I think we have a use case for the new `Sysconfigdata` in [cryptography](https://github.com/pyca/cryptography/pull/6695/files#r767289574), how about exposing the current detected Python interpreter and it's sysconfigdata in `pyo3-build-config` for downstream consumption?

aganders3:
> I think we have a use case for the new `Sysconfigdata` in [cryptography](https://github.com/pyca/cryptography/pull/6695/files#r767289574), how about exposing the current detected Python interpreter and it's sysconfigdata in `pyo3-build-config` for downstream consumption?

Just as I merge `main` and get back to green 😆.

I can look into adding this here if you all think it's best to have in the same PR. It doesn't seem like too much work. In any case I'd be happy to take this on as another project.

messense:
> I can look into adding this here if you all think it's best to have in the same PR. It doesn't seem like too much work. In any case I'd be happy to take this on as another project.

Thanks! Do it in another PR is fine.

messense (approved):

messense:
Looks good to me. Let's give @davidhewitt a chance to take another look before merging.

pyo3-build-config/src/impl_.rs
davidhewitt: This seems like a sensible way to do it to me!

davidhewitt (approved):
Looks brilliant! Thank you for all the work put into this, and I'm sorry it took me so long to find time to review.

PR #1999, opened 2021-11-15
Fix broken relative markdown link in guide

dansvo:
While browsing the guide on https://pyo3.rs/v0.13.2/conversions/traits.html, I discovered that the link to the class documentation in that doc is broken. The path to `class.md` in `traits.md` is incorrect and clicking it yields a 404.

I propose this fix to the link.


davidhewitt (approved):
Thanks!

Issue #834, opened 2020-03-25
Support exporting Rust enums to Python (old school / without data)


davidhewitt:
Following suggestion this is a subset of #417 to focus just on exporting Rust enums without data to Python. These should be fairly straightforward to support - we could borrow inspiration from C++'s `boost.Python` and `pybind11`. We might also want compatibility with the Python `enum` library.

programmerjake:
I have some messy code that you can use as a reference (originally mentioned in #417): https://salsa.debian.org/Kazan-team/simple-soft-float/-/blob/e19291a0b8eb17e9b52c09b8e670bf0fe3244989/src/python_macros.rs

davidhewitt:
I'm nominating this for `0.12`.

gilescope:
Jake's code looks like a good starting place. Has anyone had a go at it? Happy to test out PRs. I could have a crack at it myself but I think the PR would need a fair bit of hand holding...

davidhewitt:
I haven't had a go yet, but I plan to have a go soon unless someone else opens a PR sooner!

stillinbeta:
I made an attempt (#1045), but I got frustrated by the amount of boilerplate and traits required.

davidhewitt:
As we've got a fair bit to release in 0.12 I'm going to move this to the 0.13 milestone

Legend-of-iPhoenix:
Are there any recommended workarounds for the time being? I was considering just adding a python implementation, but I really don't want to have to maintain two copies of the same code in my project :)

davidhewitt:
You could construct a `#[pyclass]` with a bunch of `#[classattr]` to represent all the variants.

It's worth pointing out that #1045 is nearly there, just needs a willing person to rebase and finish it off.

b05902132:
So when I was rebasing #1045, I found there are two ways to implement #834:

- The origin PR implements each variant as a subclass of the enum, and the reviewers suggests [not to do that](https://github.com/PyO3/pyo3/pull/1045/commits/3cc9d7fa80ccc26d371efda99e957a92924f1a52#r467628540). But I think if we plan to implement #417 and #1714, then we should implement  `pyenum` this way. According to [PEP634](https://www.python.org/dev/peps/pep-0634/#class-patterns),  class patterns in Python are actually just a series of `isinstance` check. If each variant is a subclass of the enum, then we already support a small subset of class patterns, and this may lead us toward supporting Python pattern matching. Going down this way needs more design works. I can think of two issues:
[...]

Issue #417, opened 2019-03-24
Support exporting Rust enums to Python  


thedrow:
This issue replaces the closed #131.

We need to provide a way to export an enum to Python.

programmerjake:
I haven't checked, but doing [the same thing as boost python](https://www.boost.org/doc/libs/1_71_0/libs/python/doc/html/reference/high_level_components/boost_python_enum_hpp.html) might work

birkenfeld:
well, C++ enums are a fair bit different from Rust enums.

gilescope:
We should split this into two issues. One is supporting old school Enums and the other is supporting ADTs. I'd suggest that for now having a pleasant way to define a python Enum in rust would be what this issue should focus on and if there's any state in any of the rust enum states we say not currently supported.

davidhewitt:
Agreed with the above

programmerjake:
I have some messy code that you can use as a reference: https://salsa.debian.org/Kazan-team/simple-soft-float/-/blob/e19291a0b8eb17e9b52c09b8e670bf0fe3244989/src/python_macros.rs

davidhewitt:
I've made this the full ADT issue and created #834 for tracking the simple case.

gilescope:
Cool - that works too :-)

gilescope:
For full ADTs I would initially limit it to the one pyclass per variant

gilescope:
This is now released in 0.12 if people want to try it out.

thedrow:
In that case, we should close this issue no?

davidhewitt:
To clarify - in 0.12 we added `#[derive(FromPyObject)]` which lets you convert a Python type into a Rust enum. Having a full two-way binding analogue equivalent to `#[pyclass]` would be even more awesome, but it's not clear what the design would be.

gilescope:
Well the intial feedback seems to be that it works well enough (one less big switch statement - yay), 
but if we're going to do that shouldn't we also support `#[derive(IntoPy)]`  for enums?

davidhewitt:
I would be very open to having `#[derive(IntoPy)]` for enums.

[I made such a comment on the FromPyObject PR](https://github.com/PyO3/pyo3/pull/1065#issuecomment-681810136), but we did agree that it feels easier to use `.into_py()` to handle the to-python case than it does handling the FromPyObject with an if-let chain.

[...]

PR #2002, opened 2021-11-17
Derive enum

b05902132:
This PR is about #834. Actually it's just #1045 rebased on the current main branch.

However, due to PyO3 internal API changes, I have to rewrite how `pyenum` implements PyO3 traits, and I am not sure whether I got it right. I am sure I omitted docstrings and descriptors, but maybe there are some other traits that I need to implement, or maybe I implemented something wrong.

The origin implement tries to create a subclass for each variant. [This isn't compatible with current PyO3 API](https://github.com/PyO3/pyo3/issues/417#issuecomment-971767941), because we don't have a way to convert the subclass back to the enum.  To make PyO3 support this is beyond my capability, so this PR will use a simpler approach. But to support #417 and #1714 we will eventually have to do that.

This PR will only support enums with unit variants, by implement the enum as a `PyClass` with an `__eq__` method, and add a `classattr` for each variant.

Unsolved issues:
  - ~~How should we treat custom discriminant values? (Add an `__int__` implementation?)~~ Will not touch it at this PR.
  - How should we implement `__repr__`?
  - How does docstring on each variant work?

### DIFF
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -2,7 +2,8 @@
 
 use crate::attributes::{self, take_pyo3_options, NameAttribute, TextSignatureAttribute};
 use crate::deprecations::Deprecations;
-use crate::pyimpl::PyClassMethodsType;
+use crate::konst::{ConstAttributes, ConstSpec};
+use crate::pyimpl::{gen_py_const, PyClassMethodsType};
 use crate::pymethod::{impl_py_getter_def, impl_py_setter_def, PropertyType};
 use crate::utils::{self, unwrap_group, PythonDoc};
 use proc_macro2::{Span, TokenStream};
@@ -10,7 +11,14 @@ use quote::quote;
 use syn::ext::IdentExt;
 use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
-use syn::{parse_quote, spanned::Spanned, Expr, Result, Token};
+use syn::{parse_quote, spanned::Spanned, Expr, Result, Token}; //unraw
+
+/// If the class is derived from a Rust `struct` or `enum`.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum PyClassKind {
+    Struct,
+    Enum,
+}
 
 /// The parsed arguments of the pyclass macro
 pub struct PyClassArgs {
@@ -24,22 +32,28 @@ pub struct PyClassArgs {
     pub has_extends: bool,
     pub has_unsendable: bool,
     pub module: Option<syn::LitStr>,
+    pub class_kind: PyClassKind,
 }
 
-impl Parse for PyClassArgs {
-    fn parse(input: ParseStream) -> Result<Self> {
-        let mut slf = PyClassArgs::default();
-
+impl PyClassArgs {
+    fn parse(input: ParseStream, kind: PyClassKind) -> Result<Self> {
+        let mut slf = PyClassArgs::new(kind);
         let vars = Punctuated::<Expr, Token![,]>::parse_terminated(input)?;
         for expr in vars {
             slf.add_expr(&expr)?;
         }
         Ok(slf)
     }
-}
 
-impl Default for PyClassArgs {
-    fn default() -> Self {
+    pub fn parse_stuct_args(input: ParseStream) -> syn::Result<Self> {
+        Self::parse(input, PyClassKind::Struct)
+    }
+
+    pub fn parse_enum_args(input: ParseStream) -> syn::Result<Self> {
+        Self::parse(input, PyClassKind::Enum)
+    }
+
+    fn new(class_kind: PyClassKind) -> Self {
         PyClassArgs {
             freelist: None,
             name: None,
@@ -51,11 +65,10 @@ impl Default for PyClassArgs {
             is_basetype: false,
             has_extends: false,
             has_unsendable: false,
+            class_kind,
         }
     }
-}
 
-impl PyClassArgs {
     /// Adda single expression from the comma separated list in the attribute, which is
     /// either a single word or an assignment expression
     fn add_expr(&mut self, expr: &Expr) -> Result<()> {
@@ -113,6 +126,9 @@ impl PyClassArgs {
             },
             "extends" => match unwrap_group(&**right) {
                 syn::Expr::Path(exp) => {
+                    if self.class_kind == PyClassKind::Enum {
+                        bail_spanned!( assign.span() =>  "enums cannot extend from other classes" );
+                    }
                     self.base = syn::TypePath {
                         path: exp.path.clone(),
                         qself: None,
@@ -147,6 +163,9 @@ impl PyClassArgs {
                 self.has_weaklist = true;
             }
             "subclass" => {
+                if self.class_kind == PyClassKind::Enum {
+                    bail_spanned!(exp.span() => "enums can't be inherited by other classes");
+                }
                 self.is_basetype = true;
             }
             "dict" => {
@@ -328,41 +347,6 @@ impl FieldPyO3Options {
     }
 }
 
-/// To allow multiple #[pymethods] block, we define inventory types.
-fn impl_methods_inventory(cls: &syn::Ident) -> TokenStream {
-    // Try to build a unique type for better error messages
-    let name = format!("Pyo3MethodsInventoryFor{}", cls.unraw());
-    let inventory_cls = syn::Ident::new(&name, Span::call_site());
-
-    quote! {
-        #[doc(hidden)]
-        pub struct #inventory_cls {
-            methods: ::std::vec::Vec<::pyo3::class::PyMethodDefType>,
-            slots: ::std::vec::Vec<::pyo3::ffi::PyType_Slot>,
-        }
-        impl ::pyo3::class::impl_::PyMethodsInventory for #inventory_cls {
-            fn new(
-                methods: ::std::vec::Vec<::pyo3::class::PyMethodDefType>,
-                slots: ::std::vec::Vec<::pyo3::ffi::PyType_Slot>,
-            ) -> Self {
-                Self { methods, slots }
-            }
-            fn methods(&'static self) -> &'static [::pyo3::class::PyMethodDefType] {
-                &self.methods
-            }
-            fn slots(&'static self) -> &'static [::pyo3::ffi::PyType_Slot] {
-                &self.slots
-            }
-        }
-
-        impl ::pyo3::class::impl_::HasMethodsInventory for #cls {
-            type Methods = #inventory_cls;
-        }
-
-        ::pyo3::inventory::collect!(#inventory_cls);
-    }
-}
-
 fn get_class_python_name<'a>(cls: &'a syn::Ident, attr: &'a PyClassArgs) -> &'a syn::Ident {
     attr.name.as_ref().unwrap_or(cls)
 }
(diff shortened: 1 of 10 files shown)


src/types/enum.rs
```diff
@@ -0,0 +1,3 @@
+pub struct Enum;
```
davidhewitt, line 1: I think this file is dead code which can be thrown away

pyo3-macros-backend/src/pyenum.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        #to_py
+
+        #[pymethods]
+        impl #enum_ {
+            #(
+                #[allow(non_upper_case_globals)]
+                #variant_consts
+            )*
+        }
```
davidhewitt, line 76: This is clever, but I think the downside of this is that without `multiple-pymethods` feature it won't be possible for users to add their own `pymethods` block.

You should instead use [`gen_py_const`](https://github.com/PyO3/pyo3/blob/2f6ea2f96cf8d9f213ea6c8b047e95d5e4aecaca/pyo3-macros-backend/src/pyimpl.rs#L121) to create "method definitions" which then need to be fed into the collection machinery via a "methods trait" - probably `PyClassDescriptors` can be "abused" here (that's the one which collects all the `#[pyo3(get, set)]` annotations from regular `#[pyclass]`).

Let me know if you want me to give some more pointers - this code is rather messy and not well documented as it's in a permanent state of evolution / hackery :)

pyo3-macros-backend/src/pyenum.rs
```diff
@@ -0,0 +1,163 @@
...
+    let enum_cls = impl_class(enum_, None, methods_type)?;
+    let variant_names: Vec<syn::Ident> = variants
+        .iter()
+        .map(|(ident, _)| variant_enumname(enum_, ident))
+        .collect::<syn::Result<Vec<_>>>()?;
+
+    let variant_cls = variants
+        .iter()
+        .map(|(ident, _)| impl_class(ident, Some(enum_), methods_type))
+        .collect::<syn::Result<Vec<_>>>()?;
```
davidhewitt, line 48: Does there need to be separate classes for the variants? I understood that was not the plan any more, maybe this is code to be killed off?

pyo3-macros-backend/src/pyimpl.rs
```diff
@@ -6,18 +6,188 @@ use crate::{
...
+    doc: Option<PythonDoc>,
+    is_gc: bool,
+    is_basetype: bool,
+    has_extends: bool,
+    is_unsendable: bool,
+    methods_type: PyClassMethodsType,
+    base: syn::TypePath,
+}
+
+impl<'a> PyClassImplBuilder<'a> {
```
davidhewitt, line 37: 👍 this builder class is a nice refactoring!

tests/test_enum.rs
```diff
@@ -0,0 +1,50 @@
+use pyo3::prelude::*;
+use pyo3::{py_run, wrap_pyfunction};
+
+mod common;
+
+#[pyenum]
+#[derive(Debug, PartialEq, Clone)]
+pub enum MyEnum {
```
davidhewitt, line 8: Crazy idea, does this need to be a separate `#[pyenum]` macro?

This still creates a Python class, so maybe this can just be a new feature for the `#[pyclass]` macro?

davidhewitt (commented):
Thank you, I'm really excited to see progress being made for this again!

I've posted a few comments to get started. I think that may lead to a fair few changes so I've held off on more detailed review for now.

tests/test_enum.rs
b05902132: I think that's a great idea. I'm not sure how do that interact with [https://pyo3.rs/v0.15.0/class.html#customizing-the-class] though. Maybe we just disable `extends`, `subclass` and `dict`? I'll implement this PR as a `#[pyclass]` if you are okay with it.

pyo3-macros-backend/src/pyenum.rs
b05902132: Yes, I already killed that in my local repository.

pyo3-macros-backend/src/pyenum.rs
b05902132: Let me try to figure that out. Meanwhile there's a new problem: I need to implement __eq__ (via __richcmp__) for the enum to be usable, but I'll need a way to do that.

src/types/enum.rs
b05902132: I didn't see that. Will do.

tests/test_enum.rs
davidhewitt: Implementing as `#[pyclass]` sounds great to me!

> Maybe we just disable extends, subclass and dict?

I think `extends` makes sense to disable, however `subclass` and `dict` could reasonably be things that users want to add to customize their enum? I don't think they'll add much complexity to this PR, we can see as we go and decide that later.

pyo3-macros-backend/src/pyenum.rs
davidhewitt: Mmm ultimately you'll need to generate the code to create a `PyType_Slot` struct which points to the implementation of `__richcmp__` you want, which then gets fed into the collection machinery as part of the `for_each_proto_slot` function.

There's some code in `pyo3-macros-backend/src/pymethod.rs` to which handles the magic methods as part of `pymethods` which might be useful if you need to generate a `__richcmp__` method. Or alternatively maybe you're able to hard-code an "enum protocols trait" in `src/class/impl_.rs` which you can then add to enum classes only using the `PyClassImplBuilder`.

Hopefully again those are some notes to get started reading the right code - please do ask more questions as you need!

davidhewitt:
> - How should we treat custom discriminant values? (Add an `__int__` implementation?)

Can you elaborate a little more what you mean by this?

> - How should we implement `__repr__`?
> - How does docstring on each variant work?

I guess that for these two it would make sense to look at how Python does it (maybe Python 3.10?) and aim for a similar end result for users.

tests/test_enum.rs
b05902132: I just want to be future-proof. In the future when we add ADT support, each variant will be represented by a subclass (I explained why in [another comment](https://github.com/PyO3/pyo3/issues/417#issuecomment-972623184) of #417), and we want them to be the only subclasses of this enum, otherwise pattern matching will not work. (Python interpreter cannot recognize those subclasses as a variant of the enum.)

I am now okay with enabling `dict`, as I cannot think of a potential reason why we would wan't to disable `dict` in the future.

davidhewitt: > and we want them to be the only subclasses of this enum

Ahh good point, ok yes let's disable `subclass` too.

davidhewitt:
> they will expect some relationship between `Foo.Bar` and `1` at the Python side. I'm not sure what should that be.

Perhaps like Python `IntEnum`, `Foo.Bar.value` should be `1` and also `Foo.Bar == 1`?

tests/test_enum.rs
b05902132: Then we just need to figure out how to prevent subclassing from the Python side 😕.

davidhewitt: Worth writing a test but I'm fairly sure that neither Python nor Rust can subclass without `subclass`.

b05902132:
> > they will expect some relationship between `Foo.Bar` and `1` at the Python side. I'm not sure what should that be.
> 
> Perhaps like Python `IntEnum`, `Foo.Bar.value` should be `1` and also `Foo.Bar == 1`?

https://docs.python.org/3/library/enum.html#enum.IntEnum

That's implemented by subclassing `int`... Not sure if we should & how to do that.

I think I'll forbid that for now, and leave that for another PR.

tests/test_enum.rs
b05902132: Well, I just want to stop user from writing `#[pyclass(subclass)]`. We still need to secretly enable that in order to write our variant subclasses.

davidhewitt: I see. I'm not sure it can work like that; as far as Python is concerned either it's legal to subclass this or it isn't.

Might be able to control this a bit via a metaclass? Could be overkill.

b05902132:
Hi, I did a major refactoring to implementations of `#[pyclass]`. You might want to review that.

tests/test_enum.rs
b05902132: Python's `Enum` also makes use of metaclass to make sure its enums are immutable. I'll try to implement that after I finish `__eq__`.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -413,54 +414,66 @@ fn impl_enum(
...
+
     })
 }
 
+fn unit_variants_as_descriptors<'a>(
+    cls: &'a syn::Ident,
+    variant_names: impl IntoIterator<Item = &'a syn::Ident>,
+) -> TokenStream {
+    let cls_type = syn::parse_quote!(#cls);
+    let variant_to_attribute = |ident: &syn::Ident| ConstSpec {
```
b05902132, line 453: This is dirty. Have no idea how to change that though.

pyo3-macros-backend/src/pyenum.rs
b05902132: I have an idea, how about we derive from the enum and provide `__richcmp__` in the subclass? That's less complex, and we don't have to handle the situation where the user want to provide their own `__richcmp__`.
Equivalent code in Python:
```Python
class Enum:
    @classmethod
    @property
    def Var1(cls):
        return EnumVar1()
...

class EnumVar1(Enum):
    def __eq__(self, other):
        # Return true only if the other is a `Enum` and represents Var1
```

tests/test_enum.rs
b05902132: I'm sorry, for some reasons I thought `PyClass::BaseNativeType` must be a class created by the metaclass ... I found I was wrong after reading `create_type_object` and a bunch of CPython code... 😅😅
But I still have no idea whether it's possible create metaclass in PyO3, and if so what are the limitations. Did any one else try this before? Maybe I can do some experiments...

davidhewitt: Indeed there isn't yet a way to set metaclasses - #906.

Agreed, let's complete this PR without solving the metaclass issue and we can always explore that as a follow-up.

davidhewitt:
👍 will try to re-review in the next couple days when I get a chance!

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -383,242 +371,127 @@ fn impl_class(
...
+fn extract_variant_data(variant: &syn::Variant) -> syn::Result<VariantPyO3> {
+    use syn::Fields;
+    let ident = match variant.fields {
+        Fields::Unit => &variant.ident,
+        _ => {
+            return Err(syn::Error::new(
+                variant.span(),
+                "Currently only support unit variants.",
+            ))
+        }
```
davidhewitt, line 486: There's a `bail_spanned!` macro which can help with boilerplate for this kind of stuff

```suggestion
        _ => bail_spanned!(variant.span() => "Currently only support unit variants.")
```

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -383,242 +371,127 @@ fn impl_class(
...
-        }
-        PyClassMethodsType::Inventory => {
-            quote! {
-                for inventory in ::pyo3::inventory::iter::<<Self as ::pyo3::class::impl_::HasMethodsInventory>::Methods>() {
-                    visitor(::pyo3::class::impl_::PyMethodsInventory::slots(inventory));
-                }
-            }
-        }
-    };
+struct VariantPyO3<'a> {
```
davidhewitt, line 391: Perhaps this name can be more descriptive:

```suggestion
struct PyClassEnumVariant<'a> {
```

pyo3-macros-backend/src/pyclass.rs
davidhewitt: What's your dislike about it? It seems ok to me go through the existing machinery.

davidhewitt (commented):
👍  this is looking really solid to me, and nice refactoring work on the `#[pyclass]` implementation.

I'm leaning towards we tidy this PR up just a tiny bit and then merge it, because it's got a lot of changes which might conflict with other work going on.

We can create an issue to record all the pieces of follow-up work to do (such as `__eq__` implementation and guide contents). Hopefully you'd be interested in carrying on with the future work on this? It's an exciting piece of new functionality, thank you for picking this up!

pyo3-macros-backend/src/pyclass.rs
b05902132: I... can't remember 😅 I guess I wanted some constructor function.

b05902132:
> We can create an issue to record all the pieces of follow-up work to do (such as `__eq__` implementation and guide contents).
I'm okay with that, but don't we already have #417 and #834?

> Hopefully you'd be interested in carrying on with the future work on this? It's an exciting piece of new functionality, thank you for picking this up!
I'd love to implement rest of this feature.

pyo3-macros-backend/src/pyenum.rs
davidhewitt: Perhaps, I think if the user provides their own `__richcmp__` and it overrides our default one that's fine. Let's see how this goes.

davidhewitt:
> I'm okay with that, but don't we already have #417 and #834?

Yes, but I think once this merges we should aim to complete this initial functionality (e.g. `__eq__` and guide entry) before 0.16 releases, so it would be useful to have an issue to remind us of those bits.

pyo3-macros-backend/src/pyenum.rs
davidhewitt: Another thought could be to enforce singletons for each variant, this removing need for `__richcmp__`

davidhewitt (approved):
Created #2013 for follow-up work. Post on there if you think anything is missing.

Would you mind rebasing and force-pushing this PR (and potentially squashing it into one or two commits)? The CI seems to have glitched out and so I can't merge it 😅

pyo3-macros-backend/src/pyenum.rs
b05902132: I wanted to use subclass approach because I didn't understand how PyO3 collects methods and thought if we provided `__richcmp__`, the user couldn't create their own. Now I've read the implementation, I know it's not the case and I think I like introducing a new trait. 
I don't know how to enforce singletons.

b05902132:
Rebase done. Sorry there are some hiccups...

davidhewitt:
Brilliant, thanks again!

PR #2003, opened 2021-11-18
release: 0.15.1

davidhewitt:
There's a couple of PRs which include breaking changes (and I have a few in mind too which I'll author soon), so let's put out the current batch of fixes as 0.15.1 before we start chopping and changing!

I'll put this live tomorrow unless I hear a reason to wait.


davidhewitt:
Release is live!

PR #2009, opened 2021-11-19
Add Py<T>.setattr method

vxgmichel:
`PyAny` exposes a `setattr` method, but `Py<T>` currently doesn't.

TODO:
 - [x] add `Py::setattr` method
 - [x] add an entry in `CHANGELOG.md`
 - [ ] add the corresponding test


davidhewitt (approved):
LGTM, thanks!

birkenfeld:
It looks like there are other methods missing from `Py`. Should they get the same API in general?

davidhewitt:
> Should they get the same API in general?

Probably, yes, though I hadn't been rushing to add to `Py` because I'm not yet sure how much needs to change with #1056.

PR #2011, opened 2021-11-20
coverage: try again to merge all OS coverage

davidhewitt:
After #1975 the three OS jobs run but don't seem to be giving reliable numbers at the project level. I suspect that first job to finish and upload is what gets reported on the PR. I'm going to try tweaking settings like this to see if it improves things...


davidhewitt:
Hmm browsing the codecov site I'm not convinced this will improve things at all, so I'll just close it for now.

Issue #2010, opened 2021-11-20
Magic methods in `#[pymethods]` don't check they have the correct number of arguments


davidhewitt:
From #1964:

> By the way, I discovered that it is possible to omit arguments to the methods, e.g. to define a `__richcmp__` without arguments. It seems to be handled all right, just not passing the slot arguments to Rust, but I wonder if that's intentional.

Yeah I kinda knew about this and was thinking it was a minor bug. It probably should be checked and a nice error message emitted. Users missing arguments off is probably an incorrect implementation imo so we should at least force them to be explicit about unused arguments.

The code responsible is:

 https://github.com/PyO3/pyo3/blob/0f92f28ffd0f57bed1d8e8af2543e8f0b3a02a2f/pyo3-macros-backend/src/pymethod.rs#L955

method_args are the arguments the user wrote for the method; at the moment the code just iterates it without checking its length. Slightly complicated by the Python arguments but not too bad.

---- 

Anyone who's interested in helping out with this, it should be a reasonably self-contained piece.

PR #2012, opened 2021-11-21
Pymethod args

buggsaff:
Fixed issue #2010. Added a condition to check the `methods_arg` if passed then continue iterating. <br>  logs are added in the `CHANGELOG.md` file.

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -1031,19 +1031,23 @@ fn extract_proto_arguments(
     let mut non_python_args = 0;
 
     let mut args_conversions = Vec::with_capacity(proto_args.len());
-
-    for arg in method_args {
-        if arg.py {
-            arg_idents.push(py.clone());
-        } else {
-            let ident = syn::Ident::new(&format!("arg{}", non_python_args), Span::call_site());
-            let conversions = proto_args.get(non_python_args)
-                .ok_or_else(|| err_spanned!(arg.ty.span() => format!("Expected at most {} non-python arguments", proto_args.len())))?
-                .extract(cls, py, &ident, arg, extract_error_mode);
-            non_python_args += 1;
-            args_conversions.push(conversions);
-            arg_idents.push(ident);
+    if(method_args.len()!=0){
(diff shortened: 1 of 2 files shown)


pyo3-macros-backend/src/pymethod.rs
```diff
@@ -1031,20 +1031,24 @@ fn extract_proto_arguments(
...
-            arg_idents.push(py.clone());
-        } else {
-            let ident = syn::Ident::new(&format!("arg{}", non_python_args), Span::call_site());
-            let conversions = proto_args.get(non_python_args)
-                .ok_or_else(|| err_spanned!(arg.ty.span() => format!("Expected at most {} non-python arguments", proto_args.len())))?
-                .extract(cls, py, &ident, arg, extract_error_mode);
-            non_python_args += 1;
-            args_conversions.push(conversions);
-            arg_idents.push(ident);
+    if(method_args.len()!=0){
```
birkenfeld, line 1034: style issue: in Rust, no parens are needed around `if` (or `while`) conditions.

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -1031,20 +1031,24 @@ fn extract_proto_arguments(
...
+                    .ok_or_else(|| err_spanned!(arg.ty.span() => format!("Expected at most {} non-python arguments", proto_args.len())))?
+                    .extract(cls, py, &ident, arg, extract_error_mode);
+                non_python_args += 1;
+                args_conversions.push(conversions);
+                arg_idents.push(ident);
+            }
         }
     }
+    else{
+        println("Arguments missing");
```
birkenfeld, line 1050: An `Error` should be generated here and propagated upwards using `bail_spanned`.

birkenfeld (changes requested):
Hello, and thanks for your interest in working on PyO3!

There's a few points that need to be done here:

* you're only checking for empty args, but there's also the case that the user only added one of two required args. Better to check if the amount of (non-`Python`) args, i.e. `args_conversions.len()` after the `for` loop, matches the expected.
* a new test case to check this new error, probably as a UI test (see `tests/ui`).

Issue #2013, opened 2021-11-21
`#[pyclass]` on enums MVP


davidhewitt:
After #2002 it's legal to use `#[pyclass]` on enums that have only unit variants.

However, there's a few pieces which need finishing (at a minimum) before releasing this change in 0.16:
 - [ ] The generated class most likely wants a `__richcmp__` implementation so that comparisons work as expected
 - [ ] `__repr__` implementation
 - [ ] The generated class may want a metaclass to prevent subclassing
 - [ ] Docs in the guide
 - [ ] It would be nice to support custom discriminants a-la #834
 - [ ] CHANGELOG entry

davidhewitt:
(Also, if anyone has strong opinion that this to be a separate `#[pyenum]` macro instead of part of `#[pyclass]`, it would be better to split it out before 0.16. Personally I quite like having just the one `#[pyclass]` macro.)

b05902132:
> After #2002 it's legal to use `#[pyclass]` on enums that have only unit variants.
> 
> However, there's a few pieces which need finishing (at a minimum) before releasing this change in 0.16:
> 
> * [ ]  The generated class most likely wants a `__richcmp__` implementation so that comparisons work as expected
> * [ ]  `__repr__` implementation

How about we introduce a new trait in `src/class/impl_.rs` that allows us to add default implementations for magic methods? This may be useful for other situations.


> * [ ]  The generated class may want a metaclass to prevent subclassing

Related to #906.

> * [ ]  Docs in the guide
> * [ ]  It would be nice to support custom discriminants a-la [Support exporting Rust enums to Python (old school / without data) #834](https://github.com/PyO3/pyo3/issues/834)

It's not clear what's the expected behavior. Python's `IntEnum` subclasses from `int`, but that's not currently supported in PyO3, and it makes the enum immutable.


> * [ ]  CHANGELOG entry

davidhewitt:
> How about we introduce a new trait in src/class/impl_.rs that allows us to add default implementations for magic methods? This may be useful for other situations.

Agreed for sure. See also #1375 / #1376 which spoke about making "dataclass"-style `#[pyclass]` with hash, eq etc.

> Related to #906.

Yes I think we can live without this for now as it's an ergonomics improvement in my eyes rather than core functionality.

> It's not clear what's the expected behavior. 

[...]

PR #2014, opened 2021-11-21
Support default method implementation

b05902132:
Precursor to #2013. This allows us to inject default implementation of slot methods.

Add a trait (`PyClassDefaultSlots`) to support default methods.
Add `pyimpl::gen_default_slot_impls` to help generate definition of default methods in the macro backend.
Add default implementation of `__repr__` for `#[pyclass]` enums.

### DIFF
--- a/pyo3-macros-backend/src/pyimpl.rs
+++ b/pyo3-macros-backend/src/pyimpl.rs
@@ -139,6 +139,47 @@ pub fn gen_py_const(cls: &syn::Type, spec: &ConstSpec) -> TokenStream {
     }
 }
 
+pub fn gen_default_slot_impls(cls: &syn::Ident, method_defs: Vec<TokenStream>) -> TokenStream {
+    // This function uses a lot of `unwrap()`; since method_defs are provided by us, they should
+    // all succeed.
+    let ty: syn::Type = syn::parse_quote!(#cls);
+
+    let mut method_defs: Vec<_> = method_defs
+        .into_iter()
+        .map(|token| syn::parse2::<syn::ImplItemMethod>(token).unwrap())
+        .collect();
+
+    let mut proto_impls = Vec::new();
+
+    for meth in &mut method_defs {
+        let options = PyFunctionOptions::from_attrs(&mut meth.attrs).unwrap();
+        match pymethod::gen_py_method(&ty, &mut meth.sig, &mut meth.attrs, options).unwrap() {
+            GeneratedPyMethod::Proto(token_stream) => {
+                let attrs = get_cfg_attributes(&meth.attrs);
+                proto_impls.push(quote!(#(#attrs)* #token_stream))
+            }
+            GeneratedPyMethod::SlotTraitImpl(..) => {
+                panic!("SlotFragment methods cannot have default implementation!")
+            }
+            GeneratedPyMethod::Method(_) | GeneratedPyMethod::TraitImpl(_) => {
+                panic!("Only protocol methods can have default implementation!")
+            }
+        }
+    }
+
+    quote! {
+        impl #cls {
+            #(#method_defs)*
+        }
+        impl ::pyo3::class::impl_::PyClassDefaultSlots<#cls>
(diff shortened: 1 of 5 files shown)


tests/test_default_impls.rs
```diff
@@ -0,0 +1,156 @@
+#![allow(non_snake_case)]
```
davidhewitt, line 1: Heh, I see what you mean about these tests being a bit painful!

They're very coupled to the implementation, which is going to hurt whenever things get refactored.

How about instead we go a bit further in this PR and implement a default `__repr__` for the `#[pyclass]` enums. Then the tests for the overridable slots can instead be that the default `__repr__` works and a second test to confirm it's also overrideable.

... and similar for the overrideable methods, however there isn't a method we want to default right now which springs to mind.

b05902132: > however there isn't a method we want to default right now which springs to mind.

Maybe we don't support ordinary methods in this PR? We can always add a new trait later if that's needed.

davidhewitt: Sounds reasonable to me!

b05902132:
I think it's ready to be reviewed again. I need to touch a lot of implementation details in order to rename the generated method... Hope it won't be an issue in the future.

b05902132:
So I forgot there is `#[pyo3(name= ...)]` and it supports magic methods....
The new commit respects privacy better, though there's still code duplication between `pyimpl::impl_methods` and `pyimpl::gen_default_slot_impls`.

pyo3-macros-backend/src/pyimpl.rs
```diff
@@ -139,6 +139,47 @@ pub fn gen_py_const(cls: &syn::Type, spec: &ConstSpec) -> TokenStream {
...
+
+    for meth in &mut method_defs {
+        let options = PyFunctionOptions::from_attrs(&mut meth.attrs).unwrap();
+        match pymethod::gen_py_method(&ty, &mut meth.sig, &mut meth.attrs, options).unwrap() {
+            GeneratedPyMethod::Proto(token_stream) => {
+                let attrs = get_cfg_attributes(&meth.attrs);
+                proto_impls.push(quote!(#(#attrs)* #token_stream))
+            }
+            GeneratedPyMethod::SlotTraitImpl(..) => {
+                todo!()
```
b05902132, line 162: Not sure if ::pyo3::class::impl_::PyClassDefaultSlots<#cls> supports that. I think it does, but will have to check that when I have time.

b05902132:
I was very sick from my vaccine shot yesterday, so couldn't do anything. Sorry.

You mentioned there would be trait conflicts from SlotFragment implementations, so I reverted my last commit.

davidhewitt:
No worries, I have also been very busy. I'll do my best to review this over the weekend, or end of Monday at the latest.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `Py::setattr` method. [#2009](https://github.com/PyO3/pyo3/pull/2009)
 - Add `PyCapsule`, exposing the [Capsule API](https://docs.python.org/3/c-api/capsule.html#capsules). [#1980](https://github.com/PyO3/pyo3/pull/1980)
+- Add a mechanism for PyO3 to generate default definition of magic methods. [#2014](https://github.com/PyO3/pyo3/pull/2014)
```
davidhewitt, line 24: TBH this is an internal change rather than something that's meaningful to users. The output is part of `#[pyclass]` for enum, which is what we should have a CHANGELOG entry for. So I would take this one out.

```suggestion
```

davidhewitt (approved):
Thanks, this looks fine to me! Sorry for the delay.

Just needs the CHANGELOG entry removed please (and maybe squash this branch at the same time) and then I think it's good to merge.

EDIT: oh and also the `#[doc(hidden)]` addition to the generated methods.

davidhewitt (approved):
Perfect, thanks again!

PR #2018, opened 2021-11-22
Let #[pyclass] reject empty enums.

b05902132:
You can't instantiate them, it makes no sense in Python.
Rejecting them makes it easier to write default methods.


davidhewitt (approved):
Looks like a good idea to me!

PR #2019, opened 2021-11-22
inventory: update to 0.2

davidhewitt:
Reworked things a little bit internally to make inventory 0.2 work. Removed the `HasMethodsInventory` trait and just made it part of `PyClassImpl`.


birkenfeld:
Yeah looks like we should deprecate the `multiple-pymethods` feature...

davidhewitt:
See my thoughts in #2210.

PR #2021, opened 2021-11-23
hygiene: normal macros should generate `$crate`

birkenfeld:


davidhewitt (approved):
LGTM, thanks! In case there are more of these I'll let you merge this when ready.

Issue #2016, opened 2021-11-21
Allow adding docstrings to the classes created with `create_exception!`


fjarri:
At the moment the recommended way of creating custom exceptions is the `create_exception!` macro. Unfortunately, it does not allow a docstring to be specified (since that docstring will be attached to the macro invocation and not to the actual class it creates), which leads to several problems:
- Rust compiler gives a missing docstring warning, which cannot be silenced by `#[allow(missing_docs)]` (for the same reason an actual docstring cannot be specified)
- One cannot define a docstring for the created exception and have it picked up by introspecting tools on the Python side (like `help()` in REPL, or `sphinx.autodoc`) 

I tried doing an analogue of `class MyError(Exception): pass` in Rust (that's really all I need), but couldn't figure out how to write a constructor.

davidhewitt:
Personally I'd prefer not solve this and instead finish refining `#[pyclass]` for creating exceptions (#295)

davidhewitt:
(help with finishing the final bits of design and implementation written there is very much wanted!)

fjarri:
Yes, that seems reasonable. Is there some workaround one could use at the moment? I would like to be able to call the custom exception as a regular `Exception`, with a string message as a parameter, but for that I need to create a new `PyException` object in the constructor, and it doesn't seem anything but `new_err()` available.

mejrs:
The impl_native_exception macro takes a docstring as argument, perhaps we could update the other macros as well? See https://github.com/PyO3/pyo3/blob/main/src/exceptions.rs#L210

davidhewitt:
Good suggestion - adding an optional final argument would be backwards-compatible!

PR #2027, opened 2021-11-26
Allow user defined exceptions to have docstrings

mejrs:
For #2016


src/err/mod.rs
```diff
@@ -327,15 +333,70 @@ impl PyErr {
...
+    /// - `base` can be an existing exception type to subclass, or a tuple of classes.
+    /// - `dict` specifies an optional dictionary of class variables and methods.
+    /// - `doc` will be the docstring seen by python users.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if:
+    /// - `name` is not of the form `<module>.<ExceptionName>`.
+    /// - `name` or `doc` cannot be converted to [`CString`]s.
+    pub fn new_type_with_doc<'p>(
```
mejrs, line 360: We could also just merge this and new_doc, if a breaking change is acceptable here.

src/exceptions.rs
```diff
@@ -165,7 +199,22 @@ macro_rules! create_exception {
 
         $crate::impl_exception_boilerplate!($name);
 
-        $crate::create_exception_type_object!($module, $name, $base);
+        $crate::create_exception_type_object!($module, $name, $base, ::std::option::Option::None);
```
mejrs, line 202: We could also have a default docstring. I chose not to because I'm not sure whether we should write things that may be visible to end users.

davidhewitt: Yeah I agree that we should not generate a default docstring.

src/err/mod.rs
```diff
@@ -327,15 +333,70 @@ impl PyErr {
...
+        let null_terminated_name =
+            CString::new(name).expect("Failed to initialize nul terminated exception name");
+
+        let ptr = unsafe {
+            ffi::PyErr_NewException(null_terminated_name.as_ptr(), base, dict)
+                as *mut ffi::PyTypeObject
+        };
+        match NonNull::new(ptr) {
+            Some(not_null) => not_null,
+            None => panic!("{}", PyErr::fetch(py)),
```
davidhewitt, line 345: Mmm good idea checking for errors. If we're breaking `new_type` anyway, maybe we should change it to return `PyResult<Py<PyType>>` - this pushes the error handling downstream and also makes the pointer ownership clearer?

src/exceptions.rs
```diff
@@ -129,33 +129,67 @@ macro_rules! import_exception {
...
+/// #         globals.set_item("raise_myerror", fun)?;
+/// #
+/// #         py.run(
+/// # "try:
+/// #     raise_myerror()
+/// # except MyError as e:
+/// #     assert e.__doc__ == 'Some description.'
+/// #     assert str(e) == 'Some error happened.'",
+/// #             Some(globals),
+/// #             None,
```
davidhewitt, line 173: You might find it works if set locals instead of globals?

src/err/mod.rs
davidhewitt: Hmmm. I'm really on the fence about this. I do think that a `_with_doc` method should probably take `&str` rather than `Option<&str>`. These two methods can probably share a common implementation which takes option.

Otoh, I feel more strongly that the return value of `new_type` should be changed (see other comment). So that does open up the chance for us to make more breakages.

Which design do you prefer? Single method taking option or two methods?

davidhewitt:
Sorry it took me a while to get to this, thanks for implementing! Got a few ideas and comments r.e. the `new_type` stuff

src/err/mod.rs
mejrs: I'd definitely prefer a single method that takes an option.

src/exceptions.rs
mejrs: I think so too, I just find it weird it works intermittently like this.

mejrs:
I have combined the functions and made it return Result<Py<PyAny>>. I think this is good to go after the 1.57 changes merge.

davidhewitt:
Sorry for the catastrophically slow reviewing. Looks great to me! Just needs rebasing and a CHANGELOG entry :)

mejrs:
> Sorry for the catastrophically slow reviewing. Looks great to me! Just needs rebasing and a CHANGELOG entry :)

Should be good to go now :)

davidhewitt (approved):
Perfect, thanks again!

9999years:
Any chance we could cut a release with this feature? I forked pyo3 to add this but it's already here, just unreleased.

davidhewitt:
@9999years please follow #2079 - we're getting close to last piece we want to complete before next release lands. You might be able to use a git dependency in the meanwhile?

PR #2039, opened 2021-12-03
dev: remove self dev dependency

### LINKED ISSUE #1817: Remove self-dev-dependency from Cargo.toml

davidhewitt:
Closes #1817 

I thought that it wasn't going to be possible to remove `pyo3` from the dev dependencies and still be able to run plain `cargo test` meaningfully.

As #2022 would be much easier to test without the self dev dep, I took another look.

With `CARGO_PRIMARY_PACKAGE` I was actually able to put some conditional code into `acquire_gil` with the effect of automatically initializing when running the pyo3 test suite.

It does mean that on 1.48, `cargo test --features auto-initialize` is needed, as `CARGO_PRIMARY_PACKAGE` was released in 1.49. However I think that's an acceptable trade-off, and a problem which will eventually go away whenever the next msrv bump is.

This might actually help CI times a fair bit, because the `--no-default-features` steps now don't compile and run any code related to macros.

### DIFF
--- a/tests/test_buffer.rs
+++ b/tests/test_buffer.rs
@@ -1,3 +1,4 @@
+#![cfg(feature = "macros")]
 #![cfg(not(Py_LIMITED_API))]
 
 use pyo3::{
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,7 +185,7 @@ jobs:
       # Run tests (except on PyPy, because no embedding API).
       - if: ${{ !startsWith(matrix.python-version, 'pypy') }}
         name: Test (no features)
-        run: cargo test --no-default-features
+        run: cargo test --no-default-features --lib --tests
 
       # --no-default-features when used with `cargo build/test -p` doesn't seem to work!
       - name: Test pyo3-build-config (no features)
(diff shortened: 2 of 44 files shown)


tests/test_buffer.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg(feature = "macros")]
```
adamreichold, line 1: If several tests are deactivated and the others become harder to read, is this really better than requiring users to run
```
cargo test --features "auto-initialize macros"
```
or have the tests fail loudly otherwise?

davidhewitt: The end result of this PR is that `cargo test` will still run pretty much the whole pyO3 test suite, because the `macros` feature is enabled by default, and the change to `acquire_gil` means that pyo3's own tests effectively always have auto initialize enabled.

The reason a lot of test files had to get this cfg added is because `cargo test --no-default-features` now actually does disable the `macros` feature.

adamreichold: > because the macros feature is enabled by default

Ah yes, this was the thing that I was missing.

davidhewitt:
Sorry, this got merged too soon 🤦. I'll push some fixes later.

PR #2040, opened 2021-12-03
rust: support 1.57

davidhewitt:
Clippy fixes for Rust 1.57.

There's a new lint which catches that our macro code generates useless `Option::as_deref` in certain situations. I'm not entirely sure what to do about that; I might need to poke around at things a bit to find an alternative implementation.


davidhewitt:
This `Option::as_deref` lint is proving awkward - I haven't found a place in the codegen where I can add an attribute to silence it. I've been experimenting with alternative implementations, but haven't found anything which works cleanly yet.

Suggestions welcome! Otherwise I'll just keep musing over this when I have spare moments.

mejrs:
I suggest sprinkling some `#[allow(clippy::all)]` over the generated code, like

```rust
fn impl_py_method_def_new(cls: &syn::Type, spec: &FnSpec) -> Result<TokenStream> {
    let wrapper_ident = syn::Ident::new("__wrap", Span::call_site());
    let wrapper = spec.get_wrapper_function(&wrapper_ident, Some(cls))?;
    Ok(quote! {
        #[allow(clippy::all)]
        impl ::pyo3::class::impl_::PyClassNewImpl<#cls> for ::pyo3::class::impl_::PyClassImplCollector<#cls> {
            fn new_impl(self) -> ::std::option::Option<::pyo3::ffi::newfunc> {
                ::std::option::Option::Some({
                    #wrapper
                    #wrapper_ident
                })
            }
        }
    })
}```

davidhewitt:
>or in a more granular way by putting it directly on the use site:

Ah, that's solved it nicely, thanks 👍. I was trying to place it on the generated expression for `#borrow_tmp`, which the compiler didn't like (attributes on expressions aren't fully supported).

PR #2043, opened 2021-12-08
Organize examples + add __call__ example

mejrs:
This does the following:
- Moves the pyo3-benchmark and pyo3-pytests  examples to their own folder. My reasoning is that they are more intended as tests than examples.
- Adds an `[[example]]` array in examples/cargo.toml
- To not duplicate them, examples are included in the guide with the `{{#include <path>}}` macro.

My primary motivation for these changes is to play nice with [rustdoc-scrape-examples](https://github.com/rust-lang/rfcs/blob/master/text/3123-rustdoc-scrape-examples.md). (This pull request does not include an implementation of that).

This will, in the future when this feature is a bit more polished, allow us to automagically link code examples:

image
(To render that, run `cargo +nightly doc -Z unstable-options -Z rustdoc-scrape-examples=examples --open` with a sufficiently up to date toolchain).


examples/decorator/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies.pyo3]
+# If you copy this example, you should uncomment this...
+# version = "0.15.1"
+
+# ...and delete this path
+path = "../.."
+features = ["extension-module"]
+
+[workspace]
```
davidhewitt, line 19: Presumably these `[workspace]` keys are to trick cargo into treating these as new workspaces, now that we've removed these from the `Cargo.toml`?

Maybe users don't need to copy that either

```suggestion
features = ["extension-module"]
# If you copy this example, you should uncomment this...
# version = "0.15.1"

# ...and delete everything below here until the end of the file
path = "../.."

[workspace]
```

Cargo.toml
```diff
@@ -119,11 +119,7 @@ harness = false
 members = [
     "pyo3-macros",
     "pyo3-macros-backend",
-    "examples/pyo3-benchmarks",
-    "examples/pyo3-pytests",
-    "examples/maturin-starter",
-    "examples/setuptools-rust-starter",
-    "examples/word-count"
+    "examples"
```
davidhewitt, line 122: Is this change for simplicity?

One downside of this is that CI will be slower because it'll build a new fresh set of dependencies in a unique target dir for each example. We might be able to hack around that by forcing the target directory via `CARGO_TARGET_DIR`.

Another downside is that whole-workspace commands like `cargo fmt` and `cargo clippy --all-targets --workspace` won't catch the example and pytest crates any more. Opened #2047 which should resolve that (if this PR then updates the `Makefile` to ensure everything is covered).

davidhewitt (commented):
Thanks, changing things around like this seems very logical. I really love the advances this makes for us in terms of showcasing examples & reusing code in the guide 👍 

Some thoughts about the CI jobs, see other comments...

examples/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.0.0"
+publish = false
+edition = "2018"
+
+[dev-dependencies]
+pyo3 = { version = "0.15.1", path = "..", features = ["auto-initialize", "extension-module"] }
+
+[[example]]
+name = "decorator"
+path = "decorator/src/lib.rs"
```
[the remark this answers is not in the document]
mejrs: That depends on what you mean with "use in CI". If you run `cargo fmt --all -- --check` for example, it will check the contents of `decorator/src/lib.rs` as well.

Cargo.toml
mejrs: That's a good question, and I'm not sure the current approach is best.

My motivation for this is that I'd rather split all this example related stuff off from pyo3/cargo.toml. We could inline it in the pyo3/cargo.toml, but this means we have to put a lot of stuff there:

About 50-100 lines of this, depending on how many examples we end up with:
```
[[examples]]
name = ...
path = ...
```

To make example discovery work, we need to add dev dependencies to support any dependencies we only need for examples, so we end up with something like this:

```
[dev-dependencies]
assert_approx_eq = "1.1.0"
criterion = "0.3.5"
trybuild = "1.0.49"
rustversion = "1.0"
# 1.0.0 requires Rust 1.50
proptest = { version = "0.10.1", default-features = false, features = ["std"] }
serde_json = "1.0.61"

# Only needed for examples
rayon = "1"
bytes = "1"
..etc
```

I would rather be explicit about this and put this in an example crate.

For comparison, it could end up looking like [tokio](https://github.com/tokio-rs/tokio/blob/master/examples/Cargo.toml)'s example folder.

examples/decorator/Cargo.toml
mejrs: Yes, it is to make cargo play nice.

If an user clones the entire pyo3 repository and starts working in the example crate, do they still need the `[workspace]` key?

I'm not entirely sure. Is there some git command to create a repository containing only the example crate? Other than us using git submodules, which I guess we could do...

I have to say I'm not a fan of "to use this example, change this and that and this". I would really like something that Just Works straight out of the box.

examples/decorator/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "decorator"
+crate-type = ["cdylib"]
+
+[dependencies.pyo3]
+# If you copy this example, you should uncomment this...
+# version = "0.15.1"
+
+# ...and delete this path
+path = "../.."
+features = ["auto-initialize", "extension-module"]
```
davidhewitt, line 17: Hmmm baffling; the CI failure makes it look like this line is not being picked up. The build script shouldn't be emitting a link instruction...

davidhewitt:
I've pushed a commit which should fix CI. The new decorator example was missing a `tox.ini`.

examples/decorator/Cargo.toml
davidhewitt: That's fair. How about #2051 to solve the user issue?

examples/Cargo.toml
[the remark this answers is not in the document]
davidhewitt: You're right; however the `pytests/` content was not getting checked. I've tweaked the makefile, which should cover it.

Cargo.toml
davidhewitt: Ah ok, understood, and I think this is very nice 👍 

I've added a commit which tweaks CI to re-use the same target dir.

davidhewitt (approved):
I took a another look at this today and I now understand the motivations for doing this.

I like the way the `examples/decorator` file is split out and I'm keen to proceed down this path. I was looking at the sequence magic methods again and would benefit from having a similar example.

I pushed an extra pair of commits which make CI happy - if it's green then let's merge.

mejrs:
Thanks for fixing these CI errors 🙏

davidhewitt:
Took a few tries between us, we got there 😅

PR #2045, opened 2021-12-08
Fix the `#[macro_export]` absolute path within the same crate issue

danielhenrymantilla:
This takes care of https://github.com/PyO3/pyo3/runs/4455431379?check_suite_focus=true, reported in https://github.com/PyO3/pyo3/pull/2022#issuecomment-988660238 (cc @mejrs, @birkenfeld), by applying one of the workarounds suggested in https://discord.com/channels/273534239310479360/512792629516173323/918123959680122881 (mainly the one where, because we are the only callers of the meta-macro, and because we are located in a non root-of-the-crate module, to append a `pub use macro_exported_macro;` right after its definition, so that the _fully_ qualified path (`$crate::path::to::non_root_module::macro_exported_macro;`) works even when within the crate where the export occurs 🙂


davidhewitt:
Brilliant find, thanks very much! I'm going to close this one as the commit has been added into #2022 and is working nicely. 👍

PR #2047, opened 2021-12-09
ci: clippy whole workspace

davidhewitt:
I noticed there were a few things we weren't applying clippy to -- now we do!

Also changed `fmt` CI jobs to directly use the makefile, which might be useful in #2043 if we need to check clippy and formatting for examples / pytest crates outside the workspace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test test_py publish clippy lint fmt
+.PHONY: test test_py publish clippy lint fmt fmt_py fmt_rust
 
 ALL_ADDITIVE_FEATURES = macros multiple-pymethods num-bigint num-complex hashbrown serde indexmap eyre anyhow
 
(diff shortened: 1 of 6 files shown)


birkenfeld (approved):

mejrs (approved):

Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: test test_py publish clippy lint fmt
+.PHONY: test test_py publish clippy lint fmt fmt_py fmt_rust
 
 ALL_ADDITIVE_FEATURES = macros multiple-pymethods num-bigint num-complex hashbrown serde indexmap eyre anyhow
```
mejrs, line 3: I know this wasn't changed with this PR, but I'm not a fan of how we have this defined here. Can we just have a single (private) feature for all of that in our cargo.toml?

davidhewitt: I'm not a huge fan of having "private" stuff in Cargo.toml, which is why I put this here.

We could make a public feature though! What do you think of a "full" feature, which follows the convention of `syn` and `tokio`?

mejrs: I don't think "full" would be a good default choice. It makes sense for tokio, as huge swathes of their api are locked behind features.

I personally don't use make (I have some cargo aliases), so a feature would be convenient for me.

davidhewitt: That's fair. I'm going to merge this PR now, although please open a PR with a private feature if desired. (I'd also be happy for us to explore solutions such as `cargo-make` https://github.com/sagiegurari/cargo-make)

PR #2057, opened 2021-12-18
ci: try rust-cache action

davidhewitt:
https://github.com/Swatinem/rust-cache

Potentially will improve perf of CI runs; only dependencies are cached which should mean we face less cache pressure. We can also re-use the same cache across multiple Python versions.

Also has the workaround for macOS tar built-in.


davidhewitt:
Looks like it's working nicely so will proceed to merge!

PR #2061, opened 2021-12-20
Avoid UB in PySlice::indices

Amanieu:
Mutating a variables that is not declared with `mut` is undefined behavior.


birkenfeld:
Just to check that my understanding is correct: the problem is not the missing `let mut` but the mutation through a `&shared` ref, right? And `let mut` is necessary to get a `&mut` ref.

Amanieu:
From the [list of undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) in the reference:
> Mutating immutable data. All data inside a `const` item is immutable. Moreover, all data reached through a shared reference or data owned by an immutable binding is immutable, unless that data is contained within an `UnsafeCell<U>`.

davidhewitt:
Thank you for submitting this, appreciate the fix! Out of curiosity, did you find this via automated tooling or because you happened to be browsing the PyO3 source?

I might take a moment to check our other pointer casts in case this pattern was also employed elsewhere.

davidhewitt (approved):

Amanieu:
I was working on some code which used `indices` and had a look into the source code to see how it dealt with the `isize`/`usize` differences. Then I noticed the bug which immediately stood out.

PR #2064, opened 2021-12-20
ci: ignore coverage for test_hygiene, try 2

davidhewitt:
Follow up to #2056 

Looks like I can't copy and paste correctly. 😅 

In addition, looks like according to https://github.com/rust-lang/rust/issues/84605 that `#![no_coverage]` doesn't work as a module inner attribute yet, so I added the files to be ignored in `codecov.yml` too.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,3 +10,4 @@ coverage:
 
 ignore:
   - tests/*.rs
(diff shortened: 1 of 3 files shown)


davidhewitt:
Looks to have worked 🎉

PR #2065, opened 2021-12-20
pymethods: seq methods from mapping methods

davidhewitt:
Cut out from #2060

This is the part of that PR which changes `__getitem__`, `__setitem__`, and `__delitem__` to implement the sequence slots as well as the mapping slots.

I think it's reasonably clear that this design makes sense, because it matches pure-Python behavior. It defers how to solve "pure-sequence" and "pure-mapping" classes to #2060 - which might either be the `__seqlen__` (etc.) methods or attributes like `#[pyo3(sequence)]` and `#[pyo3(mapping)]`.

Note that I ended up making the decision to _not_ implement `sq_length` from `__len__`, because this way Python won't automatically try to convert negative indices to positive ones by adding the sequence length. This behaviour is inconsistent, because it only applies to calls to `__getitem__` etc from the sequence slot (not the mapping slot). We can't reverse it on the PyO3 side either, because if we get a positive index to the sequence slot we don't know whether the original index was actually a negative one but had the length added. From searching the CPython source I don't see any downsides to not implementing `sq_length` except that `PySequence::len()` will fail for pyclasses.

I'm tempted to submit a PR to upstream CPython to see if there can be a different way to opt out of this length-correcting behaviour so that we can implement `sq_length` too.

### DIFF
--- a/src/class/impl_.rs
+++ b/src/class/impl_.rs
@@ -9,7 +9,12 @@ use crate::{
     type_object::{PyLayout, PyTypeObject},
     PyClass, PyMethodDefType, PyNativeType, PyResult, PyTypeInfo, Python,
 };
-use std::{marker::PhantomData, os::raw::c_void, ptr::NonNull, thread};
+use std::{
+    marker::PhantomData,
+    os::raw::{c_int, c_void},
+    ptr::NonNull,
+    thread,
+};
 
 /// This type is used as a "dummy" type on which dtolnay specializations are
 /// applied to apply implementations from `#[pymethods]` & `#[pyproto]`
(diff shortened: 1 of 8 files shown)


src/class/impl_.rs
```diff
@@ -795,3 +800,34 @@ pub(crate) unsafe extern "C" fn fallback_new(
 pub(crate) unsafe extern "C" fn tp_dealloc<T: PyClass>(obj: *mut ffi::PyObject) {
     crate::callback_body!(py, T::Layout::tp_dealloc(obj, py))
 }
+
+pub(crate) unsafe extern "C" fn sq_item_from_mapping(
```
aviramha, line 804: I know this is similar name to the original, but maybe we should use a more clear naming ? `seq` would make more sense than `sq`. overall I'd go for `sequence` but it might be long for some peeps. same for below

src/class/impl_.rs
```diff
@@ -795,3 +800,34 @@ pub(crate) unsafe extern "C" fn fallback_new(
...
+    let index = ffi::PyLong_FromSsize_t(index);
+    if index.is_null() {
+        return std::ptr::null_mut();
+    }
+    let result = ffi::PyObject_GetItem(obj, index);
+    ffi::Py_DECREF(index);
+    result
+}
+
+pub(crate) unsafe extern "C" fn sq_ass_item_from_mapping(
```
aviramha, line 817: ass -> assign for clearer naming?

aviramha:
I think https://github.com/PyO3/pyo3/pull/2083 should be in 0.16 too!

aganders3:
For what it's worth this makes sense to me given that (as you mention) it mimics the pure-Python behavior.

I also like the proposed `#[pyo3(sequence)]` and `#[pyo3(mapping)]` attributes for the "pure" sequence and mapping types.

davidhewitt:
👍I have since also come to the conclusion that `#[pyo3(sequence)]` and `#[pyo3(mapping)]` would set the underlying type flags, which would be necesary for Python `match` support and _may_ also interact with #2072.

davidhewitt:
Took me three weeks to find a chance to finish this off! Merging now 😄

PR #2067, opened 2021-12-21
xtask: add coverage command

davidhewitt:
This is a bit of an experiment to add a `cargo xtask coverage` alias.

Ultimately this will re-implement the coverage recipe in ci. I also was playing around with adding coverage from out examples and python tests. It doesn't seem to be working quite yet; needs upstream support in `cargo-llvm-cov` which I'm still trying to figure out.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,8 @@ harness = false
 members = [
     "pyo3-macros",
     "pyo3-macros-backend",
-    "examples"
+    "examples",
+    "xtask"
 ]
 
 [package.metadata.docs.rs]
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0.51"
+clap = { version = "3.0.0-rc.7", features = ["derive"] }
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -1,8 +1,11 @@
 [package]
 name = "xtask"
 version = "0.1.0"
-edition = "2021"
+edition = "2018"
 
 [dependencies]
 anyhow = "1.0.51"
-clap = { version = "3.0.0-rc.7", features = ["derive"] }
+
+# Clap 3 requires MSRV 1.54
+rustversion = "1.0"
+structopt = { version = "0.3", default-features = false }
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -119,7 +119,8 @@ harness = false
 members = [
     "pyo3-macros",
     "pyo3-macros-backend",
-    "examples"
+    "examples",
+    "xtask"
```
mejrs, line 123: We don't necessarily need to name this "xtask". I don't have a motivation for changing it. But we could.

xtask/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0.51"
+clap = { version = "3.0.0-rc.7", features = ["derive"] }
```
mejrs, line 8: Does this add `clap` to pyo3's dependency tree? If so can we avoid that?

mejrs (approved):
Thanks :)

I've played around a bit with this "xtask' approach on a project and I like it. I'd be happy to see this go forward. And to help implement it ❤️

xtask/Cargo.toml
davidhewitt: I don't believe it does, no. This is an excerpt of `Cargo.lock` from a test project built against this branch:

```
[[package]]
name = "pyo3"
version = "0.15.1"
dependencies = [
 "cfg-if",
 "indoc",
 "libc",
 "num-bigint",
 "parking_lot",
 "paste",
 "pyo3-build-config",
 "pyo3-macros",
 "unindent",
]
```

So `clap` hasn't come through as a dependency here.

Cargo.toml
davidhewitt: Yeah, I don't personally _love_ the name xtask but I do like that it's a documented "approach" (https://github.com/matklad/cargo-xtask), so without good motivation I don't feel a need to change it.

For example I'd personally enjoy the alias being called `cargo run-script`. There's nothing stopping us adding a couple extra aliases that all do the same later, if we wanted ;)

davidhewitt:
Great! I've tidied this up a bit and added `cargo xtask test-py`, which now gets used in CI instead of `make test_py`. 

If CI passes I'll merge this now; I'll follow up on the coverage command improvements later once the `cargo-llvm-cov` support is finished, and @mejrs please do feel welcome to add to and expand to the set of commands implemented here to replace the Makefile :)

Issue #2053, opened 2021-12-16
No API docs for pyo3-build-config and others on main


messense:
https://pyo3.rs/main/doc/pyo3/index.html

![image](https://user-images.githubusercontent.com/1556054/146294316-15909db9-55f1-4b2a-8c2b-8ce7c3084bf7.png)

mejrs:
Do you think they should be?

We can have `pyo3-build-config` on the sidebar. I don't think we should have pyo3-macros and pyo3-macros-backend there.

messense:
> We can have pyo3-build-config on the sidebar. I don't think we should have pyo3-macros and pyo3-macros-backend there, because they're really not public api.

I agree.

mejrs:
It might be nice to have a `pyo3.rs/internal`, that shows all the hidden and internal api, similar to https://stdrs.dev/ Probably much more convenient than searching through a repo.

PR #2070, opened 2021-12-22
Implement rustdoc-scrape-examples

mejrs:
This implements [`rustdoc-scrape-examples`](https://rust-lang.github.io/rfcs/3123-rustdoc-scrape-examples.html) for the docs that we host ourselves (not the docs.rs ones - I figure it's best to see how this goes first). [Example](https://mejrs.github.io/pyo3/main/doc/pyo3/struct.Py.html#scraped-examples-1).

This also:
  - Implements a `full` feature that incorporates all additional features
  - Resolves [#2053](https://github.com/PyO3/pyo3/issues/2053)
  - Fixes a link I messed up on the `__call__` example in the guide
  - Fixes some links to anyhow and eyre's documentation
  - Adds cargo aliases for rendering docs
  - Renders and hosts internal documentation: [Example](https://mejrs.github.io/pyo3/internal/doc/pyo3/index.html)

Note that this creates a dead link to "decorator" in the sidebar. This should be fixed by https://github.com/rust-lang/rust/pull/92146

I also haven't touched the makefile yet - should I?

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,7 @@
+[alias]
+pyo3_doc = "doc --lib --no-default-features --features=full --no-deps --workspace --open --exclude pyo3-macros --exclude pyo3-macros-backend"
+pyo3_doc_scrape = "doc --lib --no-default-features --features=full --no-deps --workspace --open --exclude pyo3-macros --exclude pyo3-macros-backend -Z unstable-options -Z rustdoc-scrape-examples=examples"
+pyo3_doc_internal = "doc --lib --no-default-features --features=full --no-deps --workspace --open --document-private-items -Z unstable-options -Z rustdoc-scrape-examples=examples"
+
+[build]
+rustdocflags = ["--cfg", "docsrs"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,7 +1,0 @@
-[alias]
-pyo3_doc = "doc --lib --no-default-features --features=full --no-deps --workspace --open --exclude pyo3-macros --exclude pyo3-macros-backend"
-pyo3_doc_scrape = "doc --lib --no-default-features --features=full --no-deps --workspace --open --exclude pyo3-macros --exclude pyo3-macros-backend -Z unstable-options -Z rustdoc-scrape-examples=examples"
-pyo3_doc_internal = "doc --lib --no-default-features --features=full --no-deps --workspace --open --document-private-items -Z unstable-options -Z rustdoc-scrape-examples=examples"
-
-[build]
-rustdocflags = ["--cfg", "docsrs"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,8 +64,8 @@ jobs:
             echo "suppress_build_script_link_lines=true" >> config.txt
             PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features
             PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features "abi3"
-            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features "$(make list_all_additive_features)"
-            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features "abi3 $(make list_all_additive_features)"
+            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features full
+            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features "abi3 full"
           done
 
   build:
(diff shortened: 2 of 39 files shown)


Cargo.toml
```diff
@@ -79,6 +79,10 @@ auto-initialize = []
 # Optimizes PyObject to Vec conversion and so on.
 nightly = []
 
+# Activates all additional features
+# This is mostly intended for testing purposes - activating *all* of these isn't particularly useful.
+full = ["macros", "macros", "multiple-pymethods", "num-bigint", "num-complex", "hashbrown", "serde", "indexmap", "eyre", "anyhow"]
```
davidhewitt, line 84: double macros

```suggestion
full = ["macros", "multiple-pymethods", "num-bigint", "num-complex", "hashbrown", "serde", "indexmap", "eyre", "anyhow"]
```

.cargo/config.toml
```diff
@@ -0,0 +1,7 @@
+[alias]
+pyo3_doc = "doc --lib --no-default-features --features=full --no-deps --workspace --open --exclude pyo3-macros --exclude pyo3-macros-backend"
+pyo3_doc_scrape = "doc --lib --no-default-features --features=full --no-deps --workspace --open --exclude pyo3-macros --exclude pyo3-macros-backend -Z unstable-options -Z rustdoc-scrape-examples=examples"
+pyo3_doc_internal = "doc --lib --no-default-features --features=full --no-deps --workspace --open --document-private-items -Z unstable-options -Z rustdoc-scrape-examples=examples"
+
+[build]
+rustdocflags = ["--cfg", "docsrs"]
```
davidhewitt, line 7: I suspect adding this flag here means that `cargo doc` will no longer work on stable rust? That seems undesirable.

.github/workflows/ci.yml
```diff
@@ -64,8 +64,8 @@ jobs:
             echo "suppress_build_script_link_lines=true" >> config.txt
             PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features
             PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features "abi3"
-            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features "$(make list_all_additive_features)"
-            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features "abi3 $(make list_all_additive_features)"
+            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features full
+            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --no-default-features --features "abi3 full"
```
davidhewitt, line 68: Given `"macros"` is the only default feature _and_ it's in full maybe we should just kill the `--no-default-features` flag for these invocations?

```suggestion
            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --features full
            PYO3_BUILD_CONFIG=$(pwd)/config.txt cargo check --all-targets --features "abi3 full"
```

There's also similar cases further down this file (and also in the cargo aliases in `.cargo/config`).

.github/workflows/ci.yml
```diff
@@ -146,20 +146,14 @@ jobs:
...
-
       - if: matrix.msrv == 'MSRV'
         name: Prepare minimal package versions (MSRV only)
         run: |
           cargo update -p indexmap --precise 1.6.2
           cargo update -p hashbrown:0.11.2 --precise 0.9.1
 
       - name: Build docs
-        run: cargo doc --no-deps --no-default-features --features "${{ steps.settings.outputs.all_additive_features }}"
+        run: cargo doc --no-deps --no-default-features --features full
```
davidhewitt, line 156: Should this become `cargo pyo3_doc` ?

davidhewitt (commented):
Nice, this looks great to me!

Few various tidy-up suggestions etc.

> I also haven't touched the makefile yet - should I?

If you agree we should move towards cargo aliases (with "xtask" like in #2067 for more complex scripts), we can work to kill off the Makefile? If so, then don't worry about updating it here.

.cargo/config.toml
mejrs: Unfortunately there seems to be no way to pass these over the command line when using `cargo doc`. It has to be an environment variable. I used `cargo doc` because you can't pass the `--workspace` to rustdoc. I'd be happy to be wrong here :)

.github/workflows/ci.yml
mejrs: Actually, would you be OK with removing this step? Building docs is also tested in `guide.yml`.

.cargo/config.toml
davidhewitt: Ah ok, let's leave it like this for now and then maybe I'll have a play with it another time if it bothers me 😅

.github/workflows/ci.yml
davidhewitt: Hmmm, I do think there's value in checking the docs build on all each of windows, Linux and macos. Probably doesn't need to be run for each Python variant. Could make it 3.10-only?

davidhewitt (approved):
Looks good to me! Feel free to merge this as-is. I'll let you decide whether to change the docs build CI step here or in a follow up PR.

mejrs:
> Looks good to me! Feel free to merge this as-is. I'll let you decide whether to change the docs build CI step here or in a follow up PR.

I'm not sure I even want to change it now, seeing as it seems to work fine as it is (I'm guessing the rustdocflags don't get passed for some reason?).

Moving it to guide.yml and rendering docs for each possible combination of cfgs would require rebuilding pyo3 for each of those (?).

davidhewitt:
Mmm ok, let's keep as-is for now!

PR #2083, opened 2021-12-31
verify py method args count

### LINKED ISSUE #2010: Magic methods in `#[pymethods]` don't check they have the correct number of arguments

aviramha:
This fixes #2010. During the change I found out that `__ipow__` didn't receive the correct number of arguments.
While fixing it, I encountered a CPython 3.7 bug that passes invalid 3rd parameter in 3.7, so I made the parameters version dependant (3.7>= use 2 params, 3.8>= 3 params)


aviramha:
trying to fix https://github.com/PyO3/pyo3/issues/2010
I'm currently stuck having segfault.. help?

tests/test_arithmetics_protos.rs
```diff
@@ -108,7 +108,7 @@ impl PyNumberProtocol for InPlaceOperations {
         self.value |= other;
     }
 
-    fn __ipow__(&mut self, other: u32) {
+    fn __ipow__(&mut self, other: u32, _modulo: Option<u32>) {
```
mejrs, line 111: I managed to reproduce this segfault.

By changing this, it no longer segfaults.

```suggestion
    fn __ipow__(&mut self, other: u32, _modulo: u32) {
```

However, it now raises a TypeError:

```
TypeError: unsupported operand type(s) for ** or pow(): 'builtins.InPlaceOperations' and 'int'
```

mejrs: I'm using the following incantation to test, by the way.

```cargo test --test test_arithmetics --no-default-features --features full```

On Python 3.7, Windows.

Comment (aviramha) [reply tests/test_arithmetics_protos.rs:111]: Thanks. I don't think the solution is to make it non optional..

Comment (mejrs) [reply tests/test_arithmetics_protos.rs:111]: I agree but I hope this helps tracking down the issue :)

**Comment by aviramha** [comment]
Ah! Found it https://bugs.python.org/issue36379

**Comment by aviramha** [comment]
I'm not sure why the `Py_3` cfg doesn't work for the tests?

**Comment by aviramha** [comment]
Ready for review @davidhewitt

File: pyo3-macros-backend/build.rs [hunk 2]
```diff
@@ -0,0 +1,14 @@
+use pyo3_build_config::pyo3_build_script_impl::{errors::Result, resolve_interpreter_config};
+
+fn configure_pyo3() -> Result<()> {
+    let interpreter_config = resolve_interpreter_config()?;
```
davidhewitt, line 4: If I remember correctly, using `resolve_interpreter_config` from a proc-macro crate won't work correctly when cross compiling, because the environment variables provided by cargo are for the host system, not the cross-compile target.

So I don't think we can add this build script and use `#[cfg]` in `pyo3_build_config`. Instead in `pyo3_build_config` we should use the runtime APIs, like in https://github.com/PyO3/pyo3/blob/2503a2dd5e9d8b703d3b9c6ce3ce12ce0126affe/pyo3-macros-backend/src/method.rs#L210-L215

Sorry that I did not document this 😭

tests/test_arithmetics.rs
```diff
@@ -505,6 +550,77 @@ mod return_not_implemented {
     #[pyclass]
     struct RichComparisonToSelf {}
 
+    #[cfg(Py_3_8)]
+    #[pymethods]
+    impl RichComparisonToSelf {
```
davidhewitt, line 555: To avoid passing this `#[cfg]` pain onto users, do you think we can make it so that users always write three-argument `__ipow__`, and then on Python 3.7 we just always fill in `None` as the third argument?

guide/src/class/protocols.md
```diff
@@ -288,7 +288,7 @@ This trait also has support the augmented arithmetic assignments (`+=`, `-=`,
   * `fn __itruediv__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
   * `fn __ifloordiv__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
   * `fn __imod__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
-  * `fn __ipow__(&'p mut self, other: impl FromPyObject) -> PyResult<()>`
+  * `fn __ipow__(&'p mut self, other: impl FromPyObject, _modulo: impl FromPyObject) -> PyResult<()>`
```
davidhewitt, line 291:
```suggestion
  * `fn __ipow__(&'p mut self, other: impl FromPyObject, modulo: impl FromPyObject) -> PyResult<()>`
```

We should also document the special case for 3.7 (whether it's to have two arguments, or always pass `None`).

tests/ui/invalid_pymethod_proto_args.stderr
```diff
@@ -0,0 +1,5 @@
+error: Expected 1 arguments, got 0
+ --> tests/ui/invalid_pymethod_proto_args.rs:8:8
+  |
+8 |     fn __truediv__(&self) -> PyResult<()> {
+  |        ^^^^^^^^^^^
```
davidhewitt, line 5: Do you think we should include the receiver in these counts? E.g. the above becomes `Expected 2 arguments, got 1` ?

davidhewitt (changes requested):
Wow, thanks for fixing this up, and finally getting to the bottom of the issue with `__ipow__` is awesome 👍 

As an FYI my laptop keyboard is currently broken so I'm not sure how well I can check in on PRs for the next couple of weeks. Writing longer comments is hard on mobile so I may be a bit slow replying sometimes.

I think the `build-dependency` on `pyo3-build-config` for `pyo3-macros-backend` is unfortunately not correct, however there are alternative ways to resolve it. See other comments.

As a bit of bikeshedding, I also wonder if we should always allow users to write three-argument `__ipow__` and then just pass `None` always on 3.7. If we document this behaviour it seems like the easiest API for users to implement, rather than forcing them to use `#[cfg]`.

tests/ui/invalid_pymethod_proto_args.stderr
aviramha: Hmm, I was thinking about it but it might be a bit of a burden because we can also have "py" argument and take it into account which might be confusing?

tests/test_arithmetics.rs
aviramha: I'm not sure how to implement this approach and not sure if it's correct. How will the user know he wont ever receive values in Py 3.7 (besides doc?) I think compile error and strict adjustment is better...?
The issue happens where the extract layer of PyO3 tries to convert Option<u32> when it's actually an invalid value... The interface would be then OptionInvalid<u32> where it will check for python version and do it based on that..?

pyo3-macros-backend/build.rs
aviramha: I thought of using it but I was pretty sure I can't do ifs and logic when defining globals? I am still pretty confident?

guide/src/class/protocols.md
aviramha: Changed the doc.. A suggestion for formatting would be welcome (couldn't think of anything pretty)

tests/ui/invalid_pymethod_proto_args.stderr
davidhewitt: Ah I see. Can we add a test with `py` to confirm the behavior is correct in that situation too?

Other than that, I'm happy for us to leave the error message like this for now. We can always make adjustments later after we get user feedback 👍

pyo3-macros-backend/build.rs
davidhewitt: Agreed, I think you'll need to define both globals and then check which one to use based on the Python version later in the macro code.

E.g. `__IPOW__` and `__IPOW_PY_3_7__`, and then check which one to use at runtime.

This approach might not be so easy for `#[pyproto]`, in which case I'd be fine with leaving `#[pyproto]` broken. We're deprecating it anyway 😁

tests/test_arithmetics.rs
davidhewitt: Agreed that compiler errors are more explicit.

I was thinking that given this is such an edge case anyway, and Python code using `=**` will always pass `None`, it's only if users explicitly invoke `__ipow__` that there's ever a non-none value. So it doesn't seem so bad if we say "on CPython 3.7, due to an interpreter bug, you will always get None".

(Actually this makes me wonder, does it work with PyPy?)

If you're willing to give me a few days I'll write a PR which attempts to implement this always-pass-none case, to see if we like it?

aviramha: Okay. Would we need the globals in that case though? I think not? so I'll wait with further changes until you do it.
If you want you can give me some pointers on how to get it done. I thought maybe setting the `Ty::Object` to be `Ty::ObjectOrInvalid`?

davidhewitt:
I've pushed a commit which implements the strategy to always pass `None` on Python 3.7. Let me know what you think of it?

aviramha:
The changes look great. I'm still a bit concerned about the user getting confused why `__ipow__` doesn't work as expected.
Can we maybe shoot a compiler warning if `__ipow__` is implemented and compiled for Py3.7?

davidhewitt:
Hmm, I take your point. It's not possible for us to emit a compiler warning (unless on nightly).

Maybe instead of always returning `None`, we should `panic!` in `IPowModulo::extract` on Python 3.7? That way there is a safe crash instead of a segfault, and we can accompany the crash with a clear error message. Anyone implementing `__ipow__` with decent testing will then understand they can't use that argument on 3.7.

At least that way users will only need to have the `#[cfg(Py_3_8)]` or whatever for a few statements inside the function body, rather than on the whole function or `#[pymethods]` block.

If you agree with that idea, I'll pass this PR back into your hands to finish off (at your leisure) if that's ok? I'm a bit tight on time for the next few days, so I'll end up being a blocker most likely 😬

aviramha:
I'm not sure how you mean to achieve that without blocking users the usage of `__ipow__` in Py3.7?

aviramha:
If tests pass now I'm okay with merging as is.

davidhewitt (approved):
👍 thanks very much again for fixing this!

Issue #2084, opened 2022-01-01
Cargo test fails on M1 Pro


aviramha:
### Bug Description

Running cargo test fails with linker errors

### Steps to Reproduce

1. cargo test

### Backtrace

```shell
= note: Undefined symbols for architecture arm64:
            "_libintl_bind_textdomain_codeset", referenced from:
                _PyIntl_bind_textdomain_codeset in libpython3.9.a(_localemodule.o)
            "_libintl_bindtextdomain", referenced from:
                _PyIntl_bindtextdomain in libpython3.9.a(_localemodule.o)
            "_libintl_textdomain", referenced from:
                _PyIntl_textdomain in libpython3.9.a(_localemodule.o)
            "_libintl_dcgettext", referenced from:
                _PyIntl_dcgettext in libpython3.9.a(_localemodule.o)
            "_libintl_setlocale", referenced from:
                _PyLocale_setlocale in libpython3.9.a(_localemodule.o)
                _PyLocale_localeconv in libpython3.9.a(_localemodule.o)
            "_libintl_dgettext", referenced from:
                _PyIntl_dgettext in libpython3.9.a(_localemodule.o)
            "_libintl_gettext", referenced from:
                _PyIntl_gettext in libpython3.9.a(_localemodule.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
```


### Your operating system and version

macOS Monterey 12.1 (MBP 14, 2021, M1 Pro)

### Your Python version (`python --version`)

Python 3.9.9

### Your Rust version (`rustc --version`)

rustc 1.57.0 (f1edd0429 2021-11-29)

### Your PyO3 version

main branch

### How did you install python? Did you use a virtualenv?

pyenv + venv

### Additional Info

_No response_

aviramha:
`env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install VERSION`
this seems to solve it

aviramha:
@davidhewitt I think this should be documented. It seems there's no docs for setting up a dev environment for PyO3 (usually it's pretty simple) so where can we put it?

davidhewitt:
Yeah, this is one of the things which I'd like to cover in #2017. Just needs someone to find the time to write something :)

davidhewitt:
Maybe also `Contributing.md` ?

PR #2087, opened 2022-01-02
Add setting up a dev environment notes

aviramha:
Following discussions on https://github.com/PyO3/pyo3/issues/2084

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -15,6 +15,17 @@ You can browse the API of the non-public parts of PyO3 [here](https://pyo3.rs/in
 
 The following sections also contain specific ideas on where to start contributing to PyO3.
 
+## Setting up a development environment
+
+To work and develop PyO3, you need Python & Rust installed on your system.
+* We encourage the use of [rustup](https://rustup.rs/) to be able to select and choose specific toolchains based on the project.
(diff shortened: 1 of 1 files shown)


Contributing.md
```diff
@@ -15,6 +15,16 @@ You can browse the API of the non-public parts of PyO3 [here](https://pyo3.rs/in
 
 The following sections also contain specific ideas on where to start contributing to PyO3.
 
+## Setting up a development environment
+
+To work and develop PyO3, you need Python & Rust installed on your system.
+* We encourage the use of [rustup](https://rustup.rs/) to be able to select and choose specific toolchains based on the project.
+* [Pyenv](https://github.com/pyenv/pyenv) is also highly recommended for being able to choose a specific Python version.
```
davidhewitt, line 22: Actually on Ubuntu I've just used the `deadsnakes` repository to install Python versions 3.7 through 3.10 and then have made virtualenvs for each of them. Perhaps that can be added as an alternative option?

(Having multiple virtualenvs also works on Windows, where `pyenv` does not work at all.)

davidhewitt (commented):
Thanks for kicking these notes off! Please can a little extra detail be added (see below)...

Contributing.md
aviramha: lmk what you think now

davidhewitt (approved):
Great, thanks!

PR #2091, opened 2022-01-06
pyfunction: allow wrap_pyfunction to work on imports (even cross-crate)

davidhewitt:
I'm really pumped about this! I had an idea earlier today how to make `wrap_pyfunction!` work on a `#[pyfunction]` imported from another module.

How it works: `#[pyfunction] fn foo` now generates a hidden `mod foo` alongside the function. They're allowed to coexist, and `use foo` will bring both into the new scope. The hidden module contains just enough exports for `wrap_pyfunction` to be able to get what it needs. See `pyo3-macros-backend/src/pyfunction.rs` for the details.

This works _really_ nicely - it even works across different crates! 🚀 

(This trick could potentially also be used for `#[pymodule]` / `wrap_pymodule!`, though I'll save that for another day, because things may also change when we implement `#[pymodule]` on `mod` as proposed in https://github.com/PyO3/pyo3/issues/694#issuecomment-568590494.)


adamreichold:
This will really remove a lot of boiler plate `init` functions from some code bases I work with and remove a cost factor from having more fine-grained Rust-side modules. Thanks!

frameadvisors:
I am still getting this error. Is there a example/template that someone could share?

Issue #2033, opened 2021-11-29
Expose a way to access the `PYO3_CONFIG_FILE` values


davidhewitt:
See original comment from #2000 below.

If anyone has need for this, I'm happy to help review designs and implementation.

----

It would be nice to have access to the library path from `PYO3_CONFIG` (or all values) but that is not a problem for me.

> E.g. it could be based off the `PYO3_CONFIG` build mechanism - maybe we could have `pyo3_config_var!("lib_path")`. Or maybe just `pyo3_lib_path!()`. I'm open to hearing ideas on what's better.

An alternative could be a module filled with many const values, similar to what [`built` does](https://docs.rs/built/0.5.1/built/).

_Originally posted by @jonasbb in https://github.com/PyO3/pyo3/issues/2000#issuecomment-979985994_

aganders3:
This seems related to the request here https://github.com/pyca/cryptography/pull/6695#discussion_r767289574 as well? 

@messense can correct me but it seems they want to compile another static lib with the same version of Python PyO3 is using as part of a build script.

The other piece missing there is that the include path is not part of the `pyo3_build_config::InterpreterConfig`. It seems weird to add that if pyo3 itself doesn't use it, but I can come up with some alternative ideas:
1. expose `run_python_script` (or implement it on `InterpreterConfig`), which is right now duplicated in that `build.rs`
2. create something like `Sysconfigdata::from_interpreter_config`

messense:
What we want in pyca/cryptography#6695 are:

1. A way to get the currently detected Python interpreter for running a Python script file and determining the interpreter kind for example disable ABI3 if it's PyPy.
2. Get Python `include` file path from sysconfig

An `InterpreterConfig` with a `run_python_script` method sounds reasonable to me.

PR #2092, opened 2022-01-08
Add export-config feature to pyo3-build-config

aganders3:
This is intended to address the needs outlined in #2033.

Basically how this works is that a dependent crate would add `pyo3-build-config` as a build dep and enable the `export-config` feature. I made this a feature (as opposed to e.g. an environment variable) such that it could be configured via `Cargo.toml` in the dependent crate (though I think newer versions of cargo may allow you to set environment variables there as well...).

Enabling this feature causes the resolved pyo3 config to be written out during compilation. It is written in the standard pyo3 config file format because the functions for reading/writing this config already existed. This file is written to `OUT_DIR` (I think it has to be).

The only means I could find to pass information from one build script to another was [the `links` manifest key](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key): 
> The purpose of this manifest key is to give Cargo an understanding about the set of native dependencies that a package has, as well as providing a principled system of passing metadata between package build scripts.

Fortunately this was already enabled in the pyo3 crate for "python" which at least somewhat makes sense for this purpose. So in addition to writing out the config I added a key/value such that the config can be read back in by a dependent crate build script using the path in `DEP_PYTHON_PYO3_EXPORT_CONFIG`. Fortunately however this is all transparent to a dependent user, as the crate that actually reads the file back in is `pyo3-build-config` itself!

In addition I exposed two functions implemented on `InterpreterConfig` that allow running a python script using the resolved interpreter with and without additional environment variables (e.g. `PYTHONPATH`).

Feedback on this design is very welcome! It took me a while to figure out how to pass info from one build script to another.

### TODO
- [x] add an entry in CHANGELOG.md
- [x] add docs for all new functions and / or detail in the guide
- [x] add tests for all new or changed functions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,7 +227,7 @@ jobs:
         run: cargo test --manifest-path=pyo3-macros-backend/Cargo.toml
 
       - name: Test build config
-        run: cargo test --manifest-path=pyo3-build-config/Cargo.toml
+        run: cargo test --manifest-path=pyo3-build-config/Cargo.toml --all-features
 
       - name: Test python examples and tests
         shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,8 +175,13 @@ jobs:
       - if: matrix.msrv == 'MSRV'
         name: Prepare minimal package versions (MSRV only)
         run: |
+          set -x
           cargo update -p indexmap --precise 1.6.2
           cargo update -p hashbrown:0.12.0 --precise 0.9.1
+          PROJECTS=("." "examples/decorator" "examples/maturin-starter" "examples/setuptools-rust-starter" "examples/word-count")
+          for PROJ in ${PROJECTS[@]}; do
+            cargo update --manifest-path "$PROJ/Cargo.toml" -p parking_lot --precise 0.11.0
+          done
 
       - name: Build docs
         run: cargo doc --no-deps --no-default-features --features "full ${{ matrix.extra_features }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,7 +227,7 @@ jobs:
         run: cargo test --manifest-path=pyo3-macros-backend/Cargo.toml
 
       - name: Test build config
-        run: cargo test --manifest-path=pyo3-build-config/Cargo.toml --all-features
+        run: cargo test --manifest-path=pyo3-build-config/Cargo.toml
 
       - name: Test python examples and tests
         shell: bash
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -4,7 +4,7 @@ use std::{
     env,
     ffi::{OsStr, OsString},
     fmt::Display,
-    fs::{self, DirEntry},
+    fs::{self, DirEntry, File},
     io::{BufRead, BufReader, Read, Write},
     path::{Path, PathBuf},
     process::{Command, Stdio},
@@ -21,6 +21,8 @@ use crate::{
 const MINIMUM_SUPPORTED_VERSION: PythonVersion = PythonVersion { major: 3, minor: 7 };
 /// Maximum Python version that can be used as minimum required Python version with abi3.
 const ABI3_MAX_MINOR: u8 = 9;
+/// Name of config file exported by "export-config" feature.
+const EXPORT_CONFIG_FILENAME: &str = ".pyo3-export-config";
 
 /// Gets an environment variable owned by cargo.
 ///
@@ -154,6 +156,24 @@ impl InterpreterConfig {
 
         for flag in &self.build_flags.0 {
             println!("cargo:rustc-cfg=py_sys_config=\"{}\"", flag)
+        }
+
+        if cfg!(feature = "export-config") {
+            let output_path =
+                Path::new(&cargo_env_var("OUT_DIR").unwrap()).join(EXPORT_CONFIG_FILENAME);
+            if let Ok(config_file) = File::create(&output_path) {
+                if self.to_writer(config_file).is_err() {
+                    warn!(
+                        "Failed to export build config - this may be a problem for external crates that \
+                        depend on the pyo3 config."
+                    );
+                } else {
+                    println!(
+                        "cargo:PYO3_EXPORT_CONFIG={}",
+                        output_path.canonicalize().unwrap().to_str().unwrap()
+                    );
+                }
+            }
         }
     }
 
@@ -339,6 +359,14 @@ print("mingw", get_platform().startswith("mingw"))
         InterpreterConfig::from_reader(reader)
     }
 
+    /// Create an InterpreterConfig generated from pyo3-build-config using the "export-config"
+    /// feature
+    pub fn from_pyo3_export_config() -> Result<Self> {
+        InterpreterConfig::from_path(Path::new(
+            &cargo_env_var("DEP_PYTHON_PYO3_EXPORT_CONFIG").unwrap(),
+        ))
+    }
+
     #[doc(hidden)]
     pub fn from_reader(reader: impl Read) -> Result<Self> {
         let reader = BufReader::new(reader);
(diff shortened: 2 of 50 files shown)


davidhewitt:
Thank you for this! I will do my best to find a chance to review within the next few days.

.github/workflows/ci.yml
```diff
@@ -201,7 +201,7 @@ jobs:
         run: cargo test --manifest-path=pyo3-macros-backend/Cargo.toml
 
       - name: Test build config
-        run: cargo test --manifest-path=pyo3-build-config/Cargo.toml
+        run: cargo test --manifest-path=pyo3-build-config/Cargo.toml --all-features
```
aganders3, line 204: I'm not sure this is working how I thought it would since codecov is still complaining about the uncovered block in `emit_pyo3_configs()`.

aganders3: Drat - I thought I figured it out but still it's still not running the export-config test in the coverage run.

aganders3: Ha - I think it's "fixed" now but failing windows because the final coverage report command (called by `cargo-llvm-cov`) is very long.

aganders3:
This is failing patch coverage slightly, but I'm not sure it makes sense to test the un-covered lines highlighted as they involve environment variables and unlikely error cases. If it's a blocker I can refactor or come up with something.

davidhewitt:
Sorry that I'm still failing to get to this. I currently have a clear agenda on Sunday (OMG) so I have every intention to sit down and write a review then.

> This is failing patch coverage slightly, but I'm not sure it makes sense to test the un-covered lines highlighted as they involve environment variables and unlikely error cases. If it's a blocker I can refactor or come up with something.

I think that's fine; this is an awkward bit of the code base to cover because of the reasons you mention.

aganders3:
No worries! I appreciate the update but also understand this is not the most pressing feature/issue in the project or your queue. Whenever you have time is fine with me. It's typically only a minor chore to keep this rebased on main and ready to merge, since this crate is not changing too quickly.

pyo3-build-config/src/impl_.rs
```diff
@@ -155,6 +157,24 @@ impl InterpreterConfig {
         for flag in &self.build_flags.0 {
             println!("cargo:rustc-cfg=py_sys_config=\"{}\"", flag)
         }
+
+        if cfg!(feature = "export-config") {
```
davidhewitt, line 161: I think this might get called by all crates which call `pyo3_build_config::use_pyo3_cfgs`. Might be better to just put this in the PyO3 build script?

pyo3-build-config/src/impl_.rs
```diff
@@ -155,6 +157,24 @@ impl InterpreterConfig {
...
+                Path::new(&cargo_env_var("OUT_DIR").unwrap()).join(EXPORT_CONFIG_FILENAME);
+            if let Ok(config_file) = File::create(&output_path) {
+                if self.to_writer(config_file).is_err() {
+                    warn!(
+                        "Failed to export build config - this may be a problem for external crates that \
+                        depend on the pyo3 config."
+                    );
+                } else {
+                    println!(
+                        "cargo:PYO3_EXPORT_CONFIG={}",
```
davidhewitt, line 172: Rather than relying on the filesystem, should we just dump the config directly to cargo here? Only complexity I can think of is that we might need to escape newlines.

A nice thing about not touching the filesystem is that it's then simple enough that we probably can always have this functionality, rather than needing to enable it with a feature?

pyo3-build-config/src/impl_.rs
```diff
@@ -334,6 +354,15 @@ print("mingw", get_platform().startswith("mingw"))
         InterpreterConfig::from_reader(reader)
     }
 
+    /// Create an InterpreterConfig generated from pyo3-build-config using the "export-config"
+    /// feature
+    #[cfg(feature = "export-config")]
+    pub fn from_pyo3_export_config() -> Result<Self> {
```
davidhewitt, line 360: See also the hidden API `pyo3_build_config::get` (which probably should not be hidden), which looks like it achieves very similar result. Maybe we can combine the two?

pyo3-build-config/src/impl_.rs
```diff
@@ -456,6 +485,37 @@ print("mingw", get_platform().startswith("mingw"))
...
+                    .expect("No interpreter executable!"),
+            ),
+            script,
+            envs,
+        )
+    }
+
+    /// Run a python script using the executable of this InterpreterConfig.
+    pub fn run_python_script(&self, script: &str) -> Result<String> {
+        run_python_script(
```
davidhewitt, line 510: Might be worth making this call `self.run_python_script_with_envs`, for simplicity?

davidhewitt (commented):
Thank you for working on this, and sorry it's taken me so long to get to review this.

Finally starting to get my head around it; I've started by putting a few suggestions / asking some questions...

pyo3-build-config/src/impl_.rs
aganders3: Sure - that makes sense to me. The only case I can think of would be another crate basically substituting for PyO3  but also using `pyo3-build-config`. Does that make sense?

pyo3-build-config/src/impl_.rs
aganders3: Yes I think that could work. It seems a bit nice to have the artifact in the filesystem for easy inspection, but there are surely tools/methods to get that data from cargo as well.

As an alternative to dealing with newlines, the data could be put into separate `DEP_PYTHON_<VAR>` env variables so they wouldn't need to be parsed out separately.

pyo3-build-config/src/impl_.rs
aganders3: Yes this makes sense if it's okay make `pyo3_build_config::get` un-hidden. That function seems a bit magic so maybe the different cases should also be available to be called explicitly.

aganders3:
Thanks for the input! I'll do another iteration to make it a bit simpler based on your suggestions.

aganders3:
I (finally 🙁) found time to pick this back up but it looks like there have been a few changes even quite recently.

If this is still wanted my plan would be to emit to a dependency/link env var (as before) but serialize the whole `InterpreterConfig` as suggested above.

davidhewitt:
Sounds good to me 👍

pyo3-build-config/src/lib.rs
```diff
@@ -72,6 +71,8 @@ pub fn get() -> &'static InterpreterConfig {
             Ok(abi3_config())
         } else if impl_::cross_compile_env_vars().any() {
             InterpreterConfig::from_path(DEFAULT_CROSS_COMPILE_CONFIG_PATH)
+        } else if impl_::link_env_var_set() {
```
davidhewitt, line 74: To be honest I'd be tempted to try this first; it's exactly the configuration which was found by `pyo3-ffi` so it seems to make sense to use it if it exists.

Also, to avoid needing to read the environment twice, it would be cool to do something like

```rust
if let Some(parse_result) = InterpreterConfig::from_cargo_dep_env() {
    parse_result
} else if ...
```

Where `from_cargo_dep_env` returning `Option<Result<InterpreterConfig>>` should mean it doesn't need to panic.

pyo3-build-config/src/impl_.rs
```diff
@@ -461,6 +484,30 @@ print("mingw", get_platform().startswith("mingw"))
         }
         Ok(())
     }
+
+    /// Run a python script using the executable of this InterpreterConfig with additional
+    /// environment variables (e.g. PYTHONPATH) set.
+    pub fn run_python_script_with_envs<I, K, V>(&self, script: &str, envs: I) -> Result<String>
```
davidhewitt, line 490: Probably worth adding a CHANGELOG entry for this and the below new function.

Would also be great to document they panic if this config doesn't know the interpreter in a `# Panics` section.

davidhewitt (approved):
Looking great, thanks! Just a couple of final tweaks I think.

pyo3-build-config/src/lib.rs
aganders3: Thanks - this is smart!

I also appreciate the input on the ordering.

davidhewitt (approved):
Looks great to me, thanks for finishing this off! As before, do you want to test this against cryptography using a git dependency before we merge, or should I just go ahead and merge it?

aganders3:
After all this time I might as well run the test either with cryptography or at least an example package. It would be annoying to find out there's some issue after merging. I'll try to test that in the next day or so!

aganders3:
Okay I ran this on that `cryptography` PR and it seems to work but is basically stuck in the same spot it was before.

I also created a small test crate and it works as expected there: https://github.com/aganders3/hello-pyo3-build-config
You can't see the output in the CI run but running locally I can see it in the `output` file:
```
❯ cat ./target/debug/build/hello-pyo3-build-config-b6062149f81c7f0e/output
cargo:rustc-cdylib-link-arg=-undefined
cargo:rustc-cdylib-link-arg=dynamic_lookup
Hello from a Python build script run using `/Users/aanderson/src/oss/test-crates/hello-pyo3-build-config/.venv/bin/python`!
```

One quirk I noticed when building that is that it panics with `abi3-py37`, but I guess this makes sense as it does not require an interpreter for that.

Let me know if there's anything else you'd like tested or demonstrated!

davidhewitt:
Great!

> Okay I ran this on that `cryptography` PR and it seems to work but is basically stuck in the same spot it was before.

To clarify, you mean that this is working but that PR needs further implementation on the `cryptography` side once this lands?

aganders3:
> To clarify, you mean that this is working but that PR needs further implementation on the cryptography side once this lands?

Oh sorry my wording was indeed confusing. Yes you are correct - the remaining work would be on the cryptography side.

In my opinion this PR is good to go!

davidhewitt:
Excellent! Many thanks again 👍

PR #2093, opened 2022-01-10
pyfunction: allow required positional after option

davidhewitt:
I was testing out the current `main` on a large-ish project to see how easy upgrading would be, and also what the effect of the compile time improvements I've been making have been.

Overall, it seems like since 0.15.1 the reduction in LLVM lines produced has indeed improved compile times a little; in a release build the optimizer spent about 10% less time on the project.

One thing I _did_ notice was that upgrading to `main` was hard, purely because of the restriction on arguments introduced by #2041. The project in question has a lot of cases with arguments after an `Option`, and they're all used correctly without panics.

Therefore, this PR relaxes the restriction from #2041, and instead changes the behaviour such that if a positional required argument follows an `Option`, instead the `Option` is also treated as required (user must explicitly pass `None` or a value).

This option-has-implicit-None-default is not well documented anywhere, and I'd like to write an issue about this at some point to consider formalizing design of this before we hit 1.0. Maybe some time after 0.16 is out.


davidhewitt:
I'm going to proceed to merge this. Hopefully not controversial - it's less breaking than the previous solution in #2041. We can always amend again.

PR #2095, opened 2022-01-11
docs(guide): example how to include multiple Python files into the binary

mrl5:
As a Rust and pyo3 newbie I find useful to find in docs:
* information about how to include python file(s) into my binary
* an example how to call a python function that needs function imported from another file

### DIFF
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -209,6 +209,9 @@ assert userdata.as_tuple() == userdata_as_tuple
 can be used to generate a Python module which can then be used just as if it was imported with
 `PyModule::import`.
 
+**Warning**: This will compile and execute code. **Never** pass untrusted code
+to this function!
+
 ```rust
 use pyo3::{prelude::*, types::{IntoPyDict, PyModule}};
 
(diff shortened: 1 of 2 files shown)


mejrs:
Thanks ❤️ I think it's a good idea to have an example like this.

However I have a few suggestions:
- This fits better at the bottom of the page, where the guide talks about `from_code`.
- I'd prefer to (also) improve the example at https://pyo3.rs/main/doc/pyo3/prelude/struct.PyModule.html#method.from_code. This might need two examples, one including the file at compile time using `include_str!` macro, and another loading a file at runtime. I don't mind doing that myself but if you're up for it that would be great :)
- The reason this works is because `from_code` adds the module to `sys.modules`, not because they happen to be in the same directory (which is how this would work if you just ran the python file directly). That may be a bit confusing? I don't know enough about the import machinery to say whether that's a bad idea.

mrl5:
thanks for enthusiastic reception of my PR :D

> This fits better at the bottom of the page, where the guide talks about from_code.

ok no problem :) will do

> I'd prefer to (also) improve the example at https://pyo3.rs/main/doc/pyo3/prelude/struct.PyModule.html#method.from_code. 
This might need two examples, one including the file at compile time using include_str! macro, and another loading a file at runtime. I don't mind doing that myself but if you're up for it that would be great :)

I guess I should edit [src/types/module.rs#L75-L105](https://github.com/PyO3/pyo3/blob/edf03c12a7fb5f45b4dec71745e31498d43dfad8/src/types/module.rs#L75-L105) - if I'm right then no problem. I think it's a good idea to also have richer crate docs

> The reason this works is because from_code adds the module to sys.modules, not because they happen to be in the same directory (which is how this would work if you just ran the python file directly). That may be a bit confusing? I don't know enough about the import machinery to say whether that's a bad idea.

I'm not quite sure if I follow but the way I understand this section is that you would like an example that imports something nested? Like:
```
.
├── Cargo.lock
├── Cargo.toml
├── python
│   ├── utils
│   │   └── greeter.py
│   └── app.py
└── src
    └── main.rs
```
if yes then sorry I don't know how to do it. Maybe it should be done in next iteration in separate PR?

mejrs:
> you would like an example that imports something nested?

I mean to say that your approach works but not for the reason that people might expect, and that might be confusing. I wonder whether there's a better way to do it.

src/types/module.rs
```diff
@@ -93,16 +93,53 @@ impl PyModule {
...
+    ///       let from_python = Python::with_gil(|py| -> PyResult<Py<PyAny>> {
+    ///           PyModule::from_code(py, py_foo, "utils.foo", "utils.foo")?;
+    ///           let app: Py<PyAny> = PyModule::from_code(py, py_app, "", "")?
+    ///               .getattr("run")?
+    ///               .into();
+    ///           app.call0(py)
+    ///       });
+    ///
+    ///       println!("py: {}", from_python?);
+    ///       Ok(())
```
mejrs, line 113: Given that the page is pretty long already, it can be a bit shorter and simpler:

```suggestion
    ///       let code = include_str!("../example.py");
    ///
    ///       Python::with_gil(|py| {
    ///            let module = PyModule::from_code(py, code, "example", "example")?;
    ///            Ok(())
    ///       })?;
    ///       # Ok(())
```

mejrs (approved):
Thanks!

src/types/module.rs
mrl5: changed in https://github.com/PyO3/pyo3/commit/467b1be4e54e965c889cb70d15dd65c30d721726 and made sure it compiles

guide/src/python_from_rust.md
```diff
@@ -236,6 +239,100 @@ def leaky_relu(x, slope=0.01):
...
+[`std::include_str`](https://doc.rust-lang.org/std/macro.include_str.html) macro.
+
+Or you can load a file at runtime by using
+[`std::fs::read_to_string`](https://doc.rust-lang.org/std/fs/fn.read_to_string.html) function.
+
+Many Python files can be included and loaded as modules. If one file depends on
+another you must preserve correct order while declaring `PyModule`.
+
+Example directory structure:
+```
```
mejrs, line 254: These files are `include_str!`'d at the bottom of our lib.rs, so unless you tell it otherwise, rustdoc is going to try to run this as a Rust code example.

```suggestion
```text
```

src/types/module.rs
```diff
@@ -93,16 +93,41 @@ impl PyModule {
     ///
     /// # Examples
     ///
+    /// Include a file at compile time by using [`std::include_str` macro][1]:
+    ///
     /// ```no_run
```
mejrs, line 98: `no_run` will still compile the test (and try to find the file), but not run it. It has to not be compiled.

```suggestion
    /// ```ignore
```

guide/src/python_from_rust.md
```diff
@@ -236,6 +239,100 @@ def leaky_relu(x, slope=0.01):
...
+    return "baz"
+```
+
+The example below shows:
+* how to include content of `app.py` and `utils/foo.py` into your rust binary
+* how to call function `run()` (declared in `app.py`) that needs function
+  imported from `utils/foo.py`
+
+`src/main.rs`:
+```rust
```
mejrs, line 287:
```suggestion
```ignore
```

mejrs (commented):
The other CI failures are just due to the 1.58 release, don't worry about them.

mejrs:
I've fixed the Rust 1.58 failures, if you rebase on the main branch those should be fixed too 😃

mrl5:
fingers crossed it will pass now :crossed_fingers:

guide/src/python_from_rust.md
```diff
@@ -236,6 +239,100 @@ def leaky_relu(x, slope=0.01):
...
+The example below shows:
+* how to load content of `app.py` and `utils/foo.py` at runtime
+* how to call function `run()` (declared in `app.py`) that needs function
+  imported from `utils/foo.py`
+
+It is recommended to use an absolute path to your Python files because then
+your binary can be run from anywhere.
+
+`src/main.rs`:
+```rust
```
mrl5, line 315: leftover causing error on pipeline

mrl5:
final final (definitely final) pushed :D

mrl5:
I guess this is not related to my changes: https://github.com/PyO3/pyo3/runs/4828089085?check_suite_focus=true
```bash
running 4 tests
test pyfunction::tests::test_all ... ok
test pyfunction::tests::test_errs ... ok
test pyfunction::tests::test_simple_args ... ok
test pyfunction::tests::test_varargs ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: failed to generate report: process didn't exit successfully:
```

davidhewitt:
Yep coverage jobs seem to have been broken by rust nightly, don't worry about it!

@mejrs thanks for reviewing this - I'll leave in your hands to continue to review / approve / merge.

mejrs:
Thanks!

Issue #2072, opened 2021-12-22
PyMapping is inconsistent with typing.Mapping


Gobot1234:
### Bug Description

The [conversion table](https://pyo3.rs/v0.15.1/conversions/tables.html#argument-types) for PyMapping suggests anything that is a typing.Mapping should be able to be converted to PyMapping, this however is not the case seemly list and str are also downcast-able to a PyMapping.

This causes errors when you try to access things like items on the PyMapping as you get a AttributeError str has no attribute items.

### Steps to Reproduce

```rs
use pyo3::prelude::*;
use pyo3::types::PyMapping;

#[pyfunction]
fn downcast_(a: &PyAny) -> PyResult<&PyMapping> {
    Ok(a.downcast::<PyMapping>()?)
}

#[pymodule]
fn downcast(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(downcast_, m)?)?;

    Ok(())
}
```

```pycon
>>> import typing, downcast
>>> isinstance({}, typing.Mapping)
True
>>> downcast.downcast_({})
{}
>>> isinstance("", typing.Mapping)
False
>>> downcast.downcast_("")
''
```

### Your operating system and version

MacOS 11.1 (20C69)

### Your Python version (`python --version`)

3.9.3

### Your Rust version (`rustc --version`)

rustc 1.53.0-nightly (07e0e2ec2 2021-03-24)

### Your PyO3 version

0.15.1

### How did you install python? Did you use a virtualenv?

venv

davidhewitt:
Yikes, thanks for reporting.

So I'm happy to take the fall for this one. This is the [C-API definition of PyMapping_Check](https://docs.python.org/3/c-api/mapping.html?highlight=mapping%20protocol#c.PyMapping_Check):

```
.. c:function:: int PyMapping_Check(PyObject *o)

   Return ``1`` if the object provides mapping protocol or supports slicing,
   and ``0`` otherwise.  Note that it returns ``1`` for Python classes with
   a :meth:`__getitem__` method since in general case it is impossible to
   determine what type of keys it supports. This function always succeeds.
```

Reading this again, it's such a broad definition that it doesn't really seem practically useful.

[...]

PR #2098, opened 2022-01-13
Fix sequences that can be cast into PyMapping

aviramha:
Fixes https://github.com/PyO3/pyo3/issues/2072
Thanks @aganders3 for the finding the workaround.


aviramha:
I think tests are broken due to main being broken?

aganders3:
> I think tests are broken due to main being broken?

Yes seems like it coincides with the 1.58 release. Mostly UI tests failing from what I see.

mejrs:
Yes, this is because of 1.58 I'll try to fix it ⚒️

davidhewitt:
Hmmm, this is clever but I'm unsure it's the right fix.

As far as I know, classes implemented in Python with a `__getitem__` will pass both `PyMapping_Check` and `PySequence_Check`, and so never be convertible to `PyMapping` with this change.

I was also proposing to match the Python behavior in #2065, which would mean pyO3 mappings would also fail this check.

Maybe we need to ask for help from upstream?

aganders3:
Ah, bummer.

> As far as I know, classes implemented in Python with a __getitem__ will pass both PyMapping_Check and PySequence_Check, and so never be convertible to PyMapping with this change.

Yeah I think this is true unless they also subclass `dict`.

aviramha:
Closing this, let's return to the original issue to discuss.

PR #2100, opened 2022-01-13
chore: Fix 1.58 release

mejrs:


src/buffer.rs
```diff
@@ -34,6 +34,7 @@ pub struct PyBuffer<T>(Pin<Box<ffi::Py_buffer>>, PhantomData<T>);
 
 // PyBuffer is thread-safe: the shape of the buffer is immutable while a Py_buffer exists.
 // Accessing the buffer contents is protected using the GIL.
+#[allow(clippy::non_send_fields_in_send_ty)]
```
mejrs, line 37: See https://github.com/rust-lang/rust-clippy/issues/8045

We could also just make the fields `Send`, because right now the only types that implement `Element` are threadsafe already. However that would be a breaking change for anyone who has implemented `Element` on their own types.

src/ffi/unicodeobject.rs
```diff
@@ -77,10 +77,9 @@ extern "C" {
     ) -> *mut PyObject;
     #[cfg_attr(PyPy, link_name = "PyPyUnicode_FromObject")]
     pub fn PyUnicode_FromObject(obj: *mut PyObject) -> *mut PyObject;
+    // #[cfg_attr(PyPy, link_name = "PyPyUnicode_FromFormatV")]
+    // pub fn PyUnicode_FromFormatV(format: *const c_char, vargs: va_list) -> *mut PyObject;
```
mejrs, line 81: I'm not sure what's up with this..it seems that all functions using the `va_list` are commented out?

davidhewitt: I guess it's because support in std is not stable? We could enable it behind an optional flag.

https://doc.rust-lang.org/std/ffi/struct.VaList.html

src/buffer.rs
davidhewitt: Good question. Should we mark this with a FIXME or TODO to decide later?

davidhewitt (approved):
Thanks for picking this up!

src/ffi/unicodeobject.rs
mejrs: Ah, I had no idea that existed. Given that it's been on nightly for 3 years now and it doesn't look like it's close to stabilization, I think I prefer just not having these.

src/buffer.rs
mejrs: Actually that's not going to silence the lint, because it's triggering on `ffi::Py_buffer`, and we can't make that Send. 🤔

davidhewitt: Ah ok. This is just one of the many cases where Python ffi makes clippy uneasy, we'll just have to live with it.

PR #2101, opened 2022-01-13
cleanup: always enable #[track_caller]

davidhewitt:
... as `#[track_caller]` was stabilized in 1.46, and we now have MSRV 1.48.


davidhewitt:
Coverage failure looks like https://github.com/rust-lang/rust/issues/79645, I guess coverage might be down for a bit. I'll proceed to merge this.

PR #2105, opened 2022-01-16
ci: add coverage for pytests

davidhewitt:
This changes the ci job to use `cargo xtask coverage`, which now is able to generate coverage for the `pytests` modules.

Not ready to merge; depends on a branch of `maturin` (which I'll push shortly).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,7 +209,7 @@ jobs:
           python -m pip install -U pip nox
           cargo xtask test-py
         env:
-          CARGO_TARGET_DIR: ${{ github.workspace }}
+          CARGO_TARGET_DIR: ${{ github.workspace }}/target
 
       - name: Test cross compilation
         if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.9' }}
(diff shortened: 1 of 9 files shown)


davidhewitt:
Hmm, looks like there's a bit more work to do here before this is mergable. All the coverage jobs pass fine, but no coverage gets uploaded. This is the relevant output line:

> Failed to load coverage: Truncated coverage data

aganders3:
> Hmm, looks like there's a bit more work to do here before this is mergable. All the coverage jobs pass fine, but no coverage gets uploaded. This is the relevant output line:
> > Failed to load coverage: Truncated coverage data


This is failing across builds, but I think it's a rust bug in nightly. See https://github.com/taiki-e/cargo-llvm-cov/issues/128 where they recommend pinning rust nightly to `2022-01-14` or older.

davidhewitt:
> Edit: see this commit and the associated run, which now passes: c183765

👍 would you be willing to open a PR contianing that commit please? It looks like Rust nightly is battling this for a few days now, so probably best to pin so that we can have business-as-usual here.

aganders3:
Sure thing. PR is here just for reference: https://github.com/PyO3/pyo3/pull/2111

aganders3:
I *think* this now fails to generate and upload coverage because the pinned toolchain is `nightly-2022-01-14` but then this is overridden in the `xtask` by this line:
```
env.insert("RUSTUP_TOOLCHAIN".to_owned(), "nightly".to_owned());
```
...which seems to then install the `nightly` toolchain (which is broken, but also in this instance is missing the necessary `llvm-tools-preview` component). 😓

davidhewitt:
I think you're right; I'll try to tweak this again when I get a chance. It looks like I'm not checking process return codes on `xtask` either.

aganders3:
This is neat - I hadn't seen `cargo xtask` before.

I was reading through to understand how it works. It seems this branch may be missing coverage for the other crates in the workspace and doesn't account for the list in `ALL_PACKAGES`. Sorry if this is false or intentional!

messense:
I'm a bit worried that `cargo xtask` will increase compile time on CI when we add more features to it.

```bash
Compiling xtask v0.1.0 (/home/runner/work/pyo3/pyo3/xtask)
    Finished dev [unoptimized + debuginfo] target(s) in 51.77s
```

Another option is to use nox instead and just write Python code to do automations.

davidhewitt:
I had `CARGO_TARGET_DIR` for the python test step set incorrectly, which was why `xtask` build was quite so long. Last builds had xtask take more like ~7s, which is a one-off cost.

That said, I'm not super invested so we could also use nox. We could still have cargo aliases which redirect to `nox` commands. Anyway, that's something we can worry about in another PR.

I'm going to push this PR one last time to also test the other pyo3 sub-crates, and then I think we're ready to merge.

PR #2106, opened 2022-01-16
Switch from tox to nox

messense:
Get rid of `.ini` files. A lot of projects switched to nox already: https://nox.thea.codes/en/stable/#projects-that-use-nox


adamreichold:
> Get rid of .ini files. 

The wording appears a bit misleading to me as the .ini files are merely replaced by .py files.

I don't really have a stake in this and admittedly no good overview of the reasons projects would choose tox or nox, but the .py files being declarative only by convention seems to actually be a downside IMHO.

examples/decorator/cargo-generate.toml
```diff
@@ -1,5 +1,5 @@
 [template]
-ignore = [".tox"]
+ignore = [".tox", ".nox"]
```
davidhewitt, line 2:
```suggestion
ignore = [".nox"]
```

and similar for the other files

examples/decorator/noxfile.py
```diff
@@ -0,0 +1,8 @@
+import nox
+
+
+@nox.session
+def python(session):
+    session.install("-rrequirements-dev.txt")
+    session.install(".")
```
davidhewitt, line 7: To avoid having yet another environment generated for the pep 517 build

```suggestion
    session.install("-e", ".", "--no-build-isolation")
```

... and similar for the other examples.

noxfile.py
```diff
@@ -0,0 +1,6 @@
+import nox
```
davidhewitt, line 1: Should we maybe just remove top-level noxfile? I'm currently thinking `cargo xtask` can be all the top-level CI jobs, so plan to remove the makefile eventually too.

davidhewitt (commented):
This is nice, thanks! On first sight my instinct was to stick to `tox` because it's well known. I do quite like the flexibility `nox` provides, now that I've read the changes I want to proceed with this.

I think there are also some extra `tox.ini` files inside `.template` directories of the examples. Looks like they can simply be removed!

davidhewitt (approved):
Thanks, looks great to me!

The suggestion to use `--no-build-isolation` seems to work ok but requires `maturin` to be in the requirements-dev files, so I pushed a commit to add that. Will merge shortly!

Issue #148, opened 2018-05-02
How can I import Python code into Rust? 


jamesray1:
The documentation doesn't seem to be clear on that.

Context: what I'd like to do is to import [this code](https://github.com/ethereum/py-evm/pull/555/files) into my Rust [Drops of Diamond](https://github.com/Drops-of-Diamond/diamond_drops) project for blob serialization with building sharding with Ethereum. It would be nice if I could copy and paste a Python file into a Rust file, and make a few modifications to make it compatible with Rust. At the moment the documentation doesn't seem to be clear on how to do that. It seems to talk about using Rust to make a Python compatible library, but not using Python in Rust.

althonos:
Hi @jamesray1, 

I think this is feasible (although not documented at all) by doing the following:

**To embed the Python file within Rust**:

* use `include_str!` to embed your Python source into your compiled executable / library
* use `pyo3::Python.eval` to run *import* your source

**To use an external Python file**:

* make sure it is in the path (possibly, edit it with `py.import("sys")?.get("path")?.try_into<PyList>::()
 (you can also simply edit `sys.path` within Rust to import the required Python library, but from that will work only if you have installed the Python script to a particular location)
* use `eval` to define your function within the interpreter

Then you can have access to your function and call it with an argument that implements `Into<PyList>` for instance.

With an external file, the code looks like so:
```rust
extern crate pyo3;

use pyo3::prelude::*;

fn main() {
   let gil = Python::acquire_gil();
   let py = gil.python();

   let syspath: &PyList = py.import("sys")
     .unwrap()
     .get("path")
     .unwrap()
     .try_into()
     .unwrap();

   syspath.insert(0, "path/to/your/module").unwrap();

   let evm_utils = py.import("evm.utils").unwrap();
   evm_utils.call1("serialize_blobs", ...);

}
```

vertexclique:
I have tried to make importing work with different packages inside the Rust code with the example above. But I wasn't successful. Can you point me out how do you import external python code?

```python
import numpy as np

def test(args):
    return args
```

Code above throws `PyErr`:
```
Err(PyErr { type: Py(0x100a28cd0, PhantomData) })
```

abrcdf1023:
Hello guys I also has trouble in the example of using PyModule.
The piece of code is not work properly.
```rust
use pyo3::{prelude::*, types::{IntoPyDict, PyModule}};
let gil = Python::acquire_gil();
let py = gil.python();
[...]

PR #2109, opened 2022-01-17
docs(guide): example of loading python file from path at runtime in a way that it sees its dependencies

mrl5:
UC: As a pyo3 user I want example how to load at runtime python file in a way that it can import all of its modules

basically example how to modify python `sys.path` so that I dont need to load all dependencies of given file - this improves previous PR: https://github.com/PyO3/pyo3/pull/2095

this (outdated) response was helpful while doing this PR: https://github.com/PyO3/pyo3/issues/148#issuecomment-385923427

### DIFF
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -255,7 +255,7 @@ Example directory structure:
 .
 ├── Cargo.lock
 ├── Cargo.toml
-├── python
+├── python_app
 │   ├── app.py
 │   └── utils
 │       └── foo.py
@@ -263,7 +263,7 @@ Example directory structure:
     └── main.rs
 ```
 
-`python/app.py`:
+`python_app/app.py`:
 ```python
 from utils.foo import bar
 
@@ -272,7 +272,7 @@ def run():
     return bar()
 ```
 
-`python/utils/foo.py`:
+`python_app/utils/foo.py`:
 ```python
 def bar():
     return "baz"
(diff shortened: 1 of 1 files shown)


mrl5:
I guess failing coverage (https://github.com/PyO3/pyo3/runs/4846771114?check_suite_focus=true) is business as usual: https://github.com/PyO3/pyo3/pull/2095#issuecomment-1013972152

mejrs:
I don't think it's a good idea to encourage use of absolute paths, because that will make your program non-portable.

However (assuming you embed the script name in your program), you can use the `concat!` and `env!` macros to create absolute paths at compile time:

```rust
let path = concat!(env!("CARGO_MANIFEST_DIR"), "/python_script/foo.py");
```

mrl5:
@mejrs the main goal of this PR is to show how to set correct `PYTHONPATH` so that I dont need to load all dependencies of python script at runtime.

`let path = concat!(env!("CARGO_MANIFEST_DIR"), "/python_script/foo.py");` is a great improvement but for an example with `include_str!`.

I believe it doesn't make sense with reading python script from FS at runtime because:
- imagine usecase when you compile a binary that will run in some other system. in order to work this would need to have python script needed at runtime at the exact location when it was at build time
- my motivation of giving absolute path would be that if you put your python app in any machine somewhere in `/usr/share` or `/opt` or whatever then your compiled rust app will work if this condition is fulfilled

so I will update the `include_str!` example with your proposal (which is really great, much better than `../../python_dir` and then it's up to you if you want to merge this PR or not

guide/src/python_from_rust.md
```diff
@@ -304,23 +304,28 @@ fn main() -> PyResult<()> {
...
+* how to load content of `app.py` at runtime so that it sees its dependencies
+  automatically
 * how to call function `run()` (declared in `app.py`) that needs function
   imported from `utils/foo.py`
 
-It is recommended to use an absolute path to your Python files because then
-your binary can be run from anywhere.
+It is recommended to use an absolute paths because then your binary can be run
+from anywhere as long as your `app.py` is in expected directory (in example
+expected directory is `/usr/share/python_app`).
```
mejrs, line 314:
```suggestion
It is recommended to use absolute paths because then your binary can be run
from anywhere as long as your `app.py` is in the expected directory (in this example
that directory is `/usr/share/python_app`).
```

mejrs (approved):
Thanks, I have just some suggestions on a line that doesn't read quite right 🙂

mrl5:
> Thanks, I have just some suggestions on a line that doesn't read quite right slightly_smiling_face

thanks for reviewing :)

mejrs:
Thanks again :)

PR #2111, opened 2022-01-19
ci: pin rust nightly for coverage step

aganders3:
This just  pins the rust nightly version used in the coverage step to work around a current bug.  I also removed what looked like a duplicate step of installing the nightly toolchain. 

See https://github.com/taiki-e/cargo-llvm-cov/issues/128


davidhewitt (approved):
Perfect, many thanks!

PR #2115, opened 2022-01-20
Add 'contains' method to PyAny

### LINKED ISSUE #2107: Add `in` operator method to `PyAny`
We don't have a method for `x in foo` on `PyAny`.

We have `PySequence::contains` - probably this just wants to be copied as `PyAny::contains`, with some tests to check this does behave like Python's `in` operator.

As per https://github.com/PyO3/pyo3/pull/1844#issuecomment-1002551273

aganders3:
Fixes #2107 

This seems pretty straightforward as expected.


davidhewitt (approved):
Looks great, thanks!

PR #2119, opened 2022-01-21
Remove useless allow(lint)'s

mejrs:
Now that 1.58.1 is out, the send in not send lint can be removed.

Also removes some old allows that are no longer necessary.


davidhewitt (approved):
Thanks!

PR #2124, opened 2022-01-24
Fix soundness issues w/r `ExactSizeIterator`

### LINKED ISSUE #2118: Usage of `ExactSizeIterator` is incorrect
The other day, I noticed the following in the [docs for `ExactSizeIterator`](https://doc.rust-lang.org/stable/std/iter/trait.ExactSizeIterator.html):

> Note that this trait is a safe trait and as such does not and cannot guarantee that the returned length is correct. This means that unsafe code must not rely on the correctness of Iterator::size_hint. The unstable and unsafe `TrustedLen` trait gives this additional guarantee.

Unfortunately, we currently rely on the correctness of the length of `ExactSizeIterator` in `PyTuple::new` and `PyList::new` to assume that we are filling all the elements of the newly-created list (or tuple).

Instead, I guess we should track the number of elements we've inserted and the current container of the list / tuple, and then ensure that when we hit the capacity of the container we do a few things:
- For tuples, we can use `_PyTuple_Resize` to grow the allocation manually
- For lists, we probably just fall back to using `PyList_Append`

Similarly, when the iterator is exhausted we may need to shrink the allocation down so there's no uninitialized elements in the tail of the container. For tuples, this again works with `_PyTuple_Resize`. For lists, I'm not sure - we might need to fill the tail of the list with `None` for safety reasons and then slice the list to create a new list of the correct size. That seems a bit painful so maybe there's a better idea.

Unfortunately this will be a de-optimization of our tuple / list creation code, but I think it's necessary. On the upside, we can switch from `ExactSizeIterator` to just `Iterator` bound on `PyTuple::new` and `PyList::new`, allowing them to construct from more types!

If someone wants to take this on, might be a nice fun algorithmic one :)

mejrs:
Fixes #2118

### DIFF
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -2,12 +2,14 @@
 //
 // based on Daniel Grunwald's https://github.com/dgrunwald/rust-cpython
 
+use std::convert::TryInto;
+
 use crate::err::{self, PyResult};
 use crate::ffi::{self, Py_ssize_t};
 use crate::internal_tricks::get_ssize_index;
 use crate::types::PySequence;
 use crate::{
-    AsPyPointer, IntoPy, IntoPyPointer, PyAny, PyObject, PyTryFrom, Python, ToBorrowedObject,
+    AsPyPointer, IntoPy, IntoPyPointer, Py, PyAny, PyObject, PyTryFrom, Python, ToBorrowedObject,
     ToPyObject,
 };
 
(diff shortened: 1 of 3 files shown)


src/types/list.rs
```diff
@@ -18,30 +20,79 @@ pub struct PyList(PyAny);
...
 
 #[inline]
-unsafe fn new_from_iter<T>(
-    elements: impl ExactSizeIterator<Item = T>,
+#[track_caller]
+fn new_from_iter<T>(
+    py: Python,
+    mut elements: impl ExactSizeIterator<Item = T>,
     convert: impl Fn(T) -> PyObject,
 ) -> *mut ffi::PyObject {
```
mejrs, line 28: This function could return `Py<PyList>` to make clear that the pointer is owned. What do you think?

davidhewitt: Agreed, I'd support that!

src/types/list.rs
```diff
@@ -18,30 +20,79 @@ pub struct PyList(PyAny);
...
+
+        let list = ffi::PyList_New(len);
+
+        // - Panics if the ptr is null
+        // - Cleans up the list if `convert` or the asserts panic
+        let guard: Py<PyAny> = Py::from_owned_ptr(py, list);
+
+        let mut counter: Py_ssize_t = 0;
+
+        for (i, e) in elements.by_ref().enumerate().take(len as usize) {
```
davidhewitt, line 43: I think you can use `counter` as the loop variable instead of needing `enumerate()`.

Use of `.take(len)` is a really smart way to avoid the buffer overflow 👍

src/types/list.rs
```diff
@@ -18,30 +20,79 @@ pub struct PyList(PyAny);
...
+    /// function with (for example) [`Vec`]`<T>` or `&[T]` will always succeed.
+    #[track_caller]
     pub fn new<T, U>(py: Python<'_>, elements: impl IntoIterator<Item = T, IntoIter = U>) -> &PyList
     where
         T: ToPyObject,
         U: ExactSizeIterator<Item = T>,
     {
         unsafe {
-            py.from_owned_ptr::<PyList>(new_from_iter(elements.into_iter(), |e| e.to_object(py)))
+            py.from_owned_ptr::<PyList>(new_from_iter(py, elements.into_iter(), |e| {
```
davidhewitt, line 93: How about `.map()` instead of making a separate `convert` argument?

I also wonder if we should consider using `&mut dyn ExactSizeIterator<Item = PyObject>` as the argument to `new_from_iter`, to avoid generic code bloat?

src/types/list.rs
```diff
@@ -720,4 +773,124 @@ mod tests {
             assert!(list.index(42i32).is_err());
         });
     }
+
+    use std::ops::Range;
+
+    struct FaultyIter(Range<usize>, usize);
```
davidhewitt, line 779: Might be worth adding a doc and/or using named fields to make it clearer what's "faulty" about this.

src/types/list.rs
```diff
@@ -720,4 +773,124 @@ mod tests {
...
+        Python::with_gil(|py| {
+            let iter = FaultyIter(0..0, usize::MAX);
+
+            let _list = PyList::new(py, iter);
+        })
+    }
+
+    #[cfg(feature = "macros")]
+    #[test]
+    fn bad_clone_mem_leaks() {
```
davidhewitt, line 831: So do I gather correctly from this test that decrementing the list pointer even though it's not fully initialized is fine? (The NULL / uninitialized members don't cause segfault?)

davidhewitt (commented):
Lovely work, thank you! Excellent set of tests. I think worthy of a CHANGELOG?

I only commented on the list implementation, however all comments apply equally to the tuple case too.

src/types/list.rs
mejrs: > So do I gather correctly from this test that decrementing the list pointer even though it's not fully initialized is fine?

The [documentation](https://docs.python.org/3/c-api/list.html#c.PyList_New) explicitly mentions its items are `NULL` when initially initialized, rather than unitialized.

> The NULL / uninitialized members don't cause segfault?

`list_dealloc` uses `Py_XDECREF` to drop its items (see https://github.com/python/cpython/blob/main/Objects/listobject.c#L352), so a `PyList` having `NULL` items shouldn't be a problem. However I don't see this guaranteed anywhere.

davidhewitt: Yep makes sense, I think it's fine for us to assume it's sound then.

davidhewitt:
Looks great! I just want to take a moment to play with this locally to confirm if the `&dyn mut` has noticeable benchmark / code size impact. Hopefully tomorrow at some point...

davidhewitt (approved):
Yep works great!

mejrs:
> I just want to take a moment to play with this locally to confirm if the &dyn mut has noticeable benchmark / code size impact.

@davidhewitt did it make a difference? I'm curious now :)

davidhewitt:
No measurable perf difference. Code size - a little bit, figured as I couldn't see a perf change overall it seemed right 👍

PR #2126, opened 2022-01-25
move ffi module to separate crate

DSPOM2:
This PR moves the ffi bindings to a seperate crate so they can be used independently from the rest of the crate.
The motivation here is that some special use cases might want to use the ffi bindings directly without touching the rest of pyo3.
Elimination of the dependency on the rest of pyo3 makes for much easier auditing and can improve compile times:
For my project this this reduces compile times for debug build from ~6 seconds to ~3 seconds. Note that the macros feature was not used in this comparison. I am using a fairly beefy CPU (8 core ryzen). On a laptop the improvements might be even larger.

## Open Questions

There are some open questions on my side. I am not sure how to handle these as I am not as familiar with the pyo3 codebase

* There is quite a lot of duplication between the buildscripts of pyo3-ffi and pyo3. Is there a way to address this?
* Some tests and the datetime componenent of the ffi bindings had to remain in pyo3 itself due to integration with GILCell. Is this the desired solution?


pyo3-ffi/src/pyerrors.rs
```diff
@@ -158,9 +158,9 @@ pub unsafe fn PyUnicodeDecodeError_Create(
     end: Py_ssize_t,
     _reason: *const c_char,
 ) -> *mut PyObject {
-    crate::ffi::PyObject_CallFunction(
+    crate::PyObject_CallFunction(
         PyExc_UnicodeDecodeError,
-        std::ffi::CStr::from_bytes_with_nul(b"sy#nns\0")
+        std::crate::CStr::from_bytes_with_nul(b"sy#nns\0")
```
mejrs: This looks like search+replace gone wrong.

src/ffi/mod.rs
```diff
@@ -1,25 +1,3 @@
...
-//! generally the following apply:
-//! - Pointer arguments have to point to a valid Python object of the correct type,
-//! although null pointers are sometimes valid input.
-//! - The vast majority can only be used safely while the GIL is held.
-//! - Some functions have additional safety requirements, consult the
-//! [Python/C API Reference Manual][capi]
-//! for more information.
-//!
-//! [capi]: https://docs.python.org/3/c-api/index.html
 #![allow(
```
mejrs, line 1: These allows should probably go into pyo3-ffi's lib.rs.

pyo3-ffi/src/lib.rs
```diff
@@ -0,0 +1,206 @@
+//! Raw FFI declarations for Python's C API.
+//!
+//! This module provides low level bindings to the Python interpreter.
```
mejrs, line 3: It's not a module anymore 😊 (same for other uses of "module")
```suggestion
//! This crate provides low level bindings to the Python interpreter.
```

pyo3-ffi/src/lib.rs
```diff
@@ -0,0 +1,206 @@
+//! Raw FFI declarations for Python's C API.
```
mejrs, line 1: This documentation could be fleshed out a bit more. This crate should make sense on its own, so 
- Descriptions of its features
- Descriptions about the various cfg and the pyo3-build-config crate
- Something about the minimum supported Rust and Python versions
- One or two simple code examples using only the C api

You can copy paste most of these from pyo3's lib.rs

mejrs (commented):
Thanks! From glancing through this I have a few questions and comments.

> Some tests and the datetime componenent of the ffi bindings had to remain in pyo3 itself due to integration with GILCell. Is this the desired solution?

Perhaps we can have some sort of "raw api" in pyo3-ffi, with pyo3 offering a safe wrapper?

pyo3-ffi/src/pyerrors.rs
DSPOM2: I noticed this myself a couple minutes before your review (fixed in the last push) so this is already fixed

davidhewitt:
> For my project this this reduces compile times for debug build from ~6 seconds to ~3 seconds.

Is that for a clean or incremental build?

DSPOM2:
> > For my project this this reduces compile times for debug build from ~6 seconds to ~3 seconds.
> 
> Is that for a clean or incremental build?

This is for a clean build. I wouldn't expect huge changes for an incremental build (after all pyo3 is not rebuild in that case)

DSPOM2:
> I know you just moved it, but this is pretty outdated and the link is broken. Would you be able to fix it? :)

I will adress that along with your other comments :)

davidhewitt:
> This is for a clean build. I wouldn't expect huge changes for an incremental build (after all pyo3 is not rebuild in that case)

Understood! And is that 3 seconds difference just between pyo3 and pyo3-ffi, or is that the whole compile including dependencies like `parking_lot` which aren't needed any more?

DSPOM2:
The times I reported above were just how long I had to wait for cargo build to finish (after a cargo clean command).
Because I have very few dependencies and 16 threads so cargo is able to compile all of them in parallel so the 3 seconds are mostly due to pyo3 itself. To get a clearer picture I decided to do some tests with cargo bloat.

Both tests were performed after a clean build with the following command:

> cargo bloat -p verilogae_py --time -j 1

#### Compiled with pyo3
```
 Time Crate

4.98s pyo3
2.63s pyo3_build_config
0.98s libc
0.95s verilogae_py
0.78s pyo3_ffi
0.58s parking_lot_core
0.49s parking_lot
0.33s verilogae_ffi
0.20s once_cell
0.19s smallvec
0.19s lock_api
0.05s instant
0.04s scopeguard
0.02s cfg_if
```
#### Compiled with pyo3-ffi
```
 Time Crate

2.61s pyo3_build_config
1.04s verilogae_py
1.01s libc
0.78s pyo3_ffi
0.33s verilogae_ffi
0.20s once_cell
```

davidhewitt:
Thanks! I'm reasonably convinced that if we can make the split tidy this will be helpful for projects like yours 👍

pyo3-ffi/src/lib.rs
DSPOM2: I have added both documentation and a readme (that for the most part just copies said documentation). Writing these kinds of docs is not my strength so I would welcome some feedback :smile:

src/ffi/mod.rs
DSPOM2: I have addresses these attributes in both the lib.rs and mod.rs file, because the datetime.rs file (which is still inside pyo3 itself) still requires these allows. I could move these to the `datetime.rs` file however.

DSPOM2:
> Perhaps we can have some sort of "raw api" in pyo3-ffi, with pyo3 offering a safe wrapper?

I created such a raw API. Instead of a GILCell this API uses a mutable static (like the original c macros in cpython). For backwards compatibility I then shadow this static with a wrapper that ensure that the datetime api is imported before use. Furthermore I reexport all functions/c macros that depends on this static in the ffi module. These shadowed implementations ensure that the API is imported (like the previous implementations did automatically with deref) and then call the raw C API. This should retain backwards compatibility while still exposing all raw functionality in pyo3-ffi. 

The implementation should compile down to basically the same thing as previously but its still very unsafe and it would be good if someone more familiar with the codebase checked this thoroughly.

DSPOM2:
While writing the example for the documentation I noticed that the definition of `_PyCFunctionFast` was incorrect. The argument  `    kwnames: *mut PyObject` is only present in the `PyCFunctionWithKeywords` funciton according to the docs (and that makes sense if you look at the name) but in pyo3 it was also present in the `_PyCFunctionFast` definition. I included a small fix for this as a separate commit. I use `_PyCFunctionFast` in the docs so so the commit is somewhat related. If this is undesired I can drop the commit and create a separate PR

DSPOM2:
I have moved most of the buildscript to pyo3-ffi as that is where all the linking takes places (this is what was causing the build failures on windows I believe). This is now ready for review

adamreichold:
> Instead of a GILCell this API uses a mutable static (like the original c macros in cpython).

In light of https://github.com/rust-lang/rust/issues/53639, `UnsafeCell` might be preferable to `static mut`?

DSPOM2:
> In light of rust-lang/rust#53639, UnsafeCell might be preferable to static mut?

While I am pretty sure this implementation is safe (as there are essentially only ever reads and writes and no references are used) providing a cell would potentially be better API for downstream crates. I will take a look at this tomorrow (its quite late in my timezone)

src/ffi/mod.rs
```diff
@@ -29,182 +30,93 @@
...
-// skipped odictobject.h
-// skipped opcode.h
-// skipped osdefs.h
-mod osmodule;
-// skipped parser_interface.h
-// skipped patchlevel.h
-// skipped picklebufobject.h
-// skipped pyctype.h
-// skipped py_curses.h
+pub static PyDateTimeAPI: _PyDateTimeAPI_impl = _PyDateTimeAPI_impl {};
```
davidhewitt, line 54: I actually think we should remove these `PyDateTimeAPI` and `PyDateTime_TimeZone_UTC` statics, I don't think it's good for them to differ from the ffi crate.

src/ffi/mod.rs
```diff
@@ -29,182 +30,93 @@
...
-pub mod structmember;
-
-// "Limited API" definitions matching Python's `include/cpython` directory.
+pub static PyDateTime_TimeZone_UTC: _PyDateTime_TimeZone_UTC_impl = _PyDateTime_TimeZone_UTC_impl {
+    inner: &PyDateTimeAPI,
+};
+
 #[cfg(not(Py_LIMITED_API))]
-mod cpython;
+macro_rules! reexport_with_autoinit {
```
davidhewitt, line 62: These "autoinit" helpers are very useful for the safe wrapper, but we should explore whether they should instead be moved to be private features of `types/datetime.rs`.

davidhewitt (commented):
Thanks! Got a few thoughts...

src/ffi/mod.rs
DSPOM2: Now that the datetime API has been removed from the pyo3 ffi module I was able to remove these attributes aswell

build.rs
```diff
@@ -124,18 +23,6 @@ fn emit_link_config(interpreter_config: &InterpreterConfig) -> Result<()> {
 fn configure_pyo3() -> Result<()> {
     let interpreter_config = resolve_interpreter_config()?;
```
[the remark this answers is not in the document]
DSPOM2: done :)

pyo3-ffi/build.rs
```diff
@@ -0,0 +1,151 @@
...
+use pyo3_build_config::{
+    bail, ensure,
+    pyo3_build_script_impl::{
+        cargo_env_var, env_var, errors::Result, resolve_interpreter_config, InterpreterConfig,
+        PythonVersion,
+    },
+};
+
+/// Minimum Python version PyO3 supports.
+const MINIMUM_SUPPORTED_VERSION: PythonVersion = PythonVersion { major: 3, minor: 6 };
```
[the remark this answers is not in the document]
DSPOM2: done :)

pyo3-ffi/src/datetime.rs
```diff
@@ -0,0 +1,561 @@
...
+        tzinfo: *mut PyObject,
+        fold: c_int,
+        cls: *mut PyTypeObject,
+    ) -> *mut PyObject,
+}
+
+// Python already shares this object between threads, so it's no more evil for us to do it too!
+unsafe impl Sync for PyDateTime_CAPI {}
+
+pub static mut PyDateTimeAPI: *mut PyDateTime_CAPI = ptr::null_mut();
```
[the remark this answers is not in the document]
DSPOM2: I ended up using a wrapped around Cell (since pointers are copy). When using pyo3_ffi directly I was exactly missing this kind of type for a similar purpose (for example initializing numpy api). Therefore I exposed this cell type as `UnsafeGILCell`. While this is conceptually a very simple type I think it makes writing unsafe ffi code in rust must more ergonomic (this is much better than mutable statics)

src/ffi/mod.rs
DSPOM2: I have removed these statics and adjusted the save code. This actually simplified the code as all access to these statics (and also the appropriate functions) only happens when the GIL is already acquired which removed the need to call `Python::aquire_gil`. While this is trivial to see in most places (as a `Python` instance is already available) I had to use `Python::assume_gil` in the  `pyobject_native_type!` macros as these don't provide direct access to a `Python` instance. However I am sure this is correct from looking at the source code. Nonetheless I marked all instances of this with a `REVIEW` comment. It would be good to check that my understand is correct here

src/ffi/mod.rs
DSPOM2: I have addressed this. There is no more special datetime handeling in pyo3::ffi compared to pyo3::ffi anymore. See above for details

pyo3-ffi/build.rs
```diff
@@ -0,0 +1,151 @@
...
+            rust_target == pointer_width,
+            "your Rust target architecture ({}-bit) does not match your python interpreter ({}-bit)",
+            rust_target,
+            pointer_width
+        );
+    }
+    Ok(())
+}
+
+fn ensure_auto_initialize_ok(interpreter_config: &InterpreterConfig) -> Result<()> {
```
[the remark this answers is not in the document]
DSPOM2: done :)

pyo3-ffi/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+    /// Returns a copy of the contained value.
+    ///
+    /// # Safety
+    /// The caller must assure that the GIL is held
+    pub unsafe fn get(&self) -> T {
+        self.0.get()
+    }
+}
+
+unsafe impl<T> Sync for UnsafeGILCell<T> {}
```
adamreichold, line 456: Shouldn't these still include bounds on `T`? Otherwhise while the contract of `UnsafeGILCell` would ensure that updating the shared value is synchronized via the GIL, this would still allow to transfer a `T: !Send` from one thread to another.

Therefore, I think this needs at least
```rust
 unsafe impl<T: Send> Sync for UnsafeGILCell<T> {}
 unsafe impl<T: Send> Send for UnsafeGILCell<T> {}
```
similar to std's [`Mutex`](https://doc.rust-lang.org/stable/std/sync/struct.Mutex.html#impl-Sync).

DSPOM2: Good point. Trouble is we are storing a raw pointer in the cell... which is not Send (and this is exactly the usecase I have in mind for downstream crates aswell). This is technically still sound because set and get are unsafe anyway so maybe documenting the extra bound would be sufficent?

adamreichold: Is the current generality actually required for these use cases? So maybe
```rust
pub struct UnsafeGILCell<T>(Cell<*const T>);
```
would be a better definition and would allow the bounds to be formulated in terms of the pointee `T`? (Which would have to be `Sync` because `&T: Send` if `T: Sync` I guess?)

mejrs: I don't really know much about the datetime api - but this looks like we're reinventing `std::sync::atomic::AtomicPtr`. Can we not use that instead?

adamreichold: In principle we do. This is only an optimization insofar it avoids the costs of the atomic accesses if we already have the GIL. (I think the accesses have to be acquire/release pairs at least since we want to publish pointers to objects which can impede some compiler optimizations even though they do not require memory barriers on strongly ordered architectures.)

adamreichold: (And taking the GIL is almost surely more expensive if we would have to do that just to access the static.)

DSPOM2: Also the reason I implemented it this way was that my goal with the API was to mirror the python C API as close as possible (which uses a mutable static) in the pyo3_ffi crate. Other python APIs do the same thing (accessing some mutable static) which we also guard against with require ring the GIL. So it seemed more consistent to me to provide such a raw unsafe mechanism in the pyo3_ffi crate aswell. This also makes a lot of sense because all places where the datetime API is called within pyo3 itself a GIL is always present anyways (so no additional locking mechanism is required). In fact I can't really come up with a usecase where one would want to access this API without holding the GIL. 

With regard to your earlier comment: My other usecase for this generality (in downstream crates) was static `PyTypeObject` instances like they are typically found in python extensions written in C. However I need a pointer to those instances anyway (which would require an UnsafeCell and at that point very few advantages of just a plain static mut) therefore I will follow your suggestion to change this to a 
```
pub struct UnsafeGILCell<T>(Cell<*const T>);

File: pyo3-ffi/src/lib.rs [hunk 12]
```diff
@@ -0,0 +1,455 @@
...
+#[cfg(not(Py_LIMITED_API))]
+pub use self::cpython::*;
+
+/// This cell type allows shared mutability for python types.
+/// It is intended to hold raw pointers to the python heap and therefore only accepts copy types to
+/// reduce the changes of accidental undefined behaviour.
+/// Thread safety is ensured by the python GIL.
+/// Therefore accessing the contents of this cell in any form while the GIL is not held is UB
+#[repr(transparent)]
+pub struct UnsafeGILCell<T>(Cell<*mut T>);
```
davidhewitt, line 430: Two thoughts:
- `Cell` just is a wrapper around `UnsafeCell`, so I'd prefer we use `UnsafeCell` as the inner type.
- I don't think that we need a generic type here, given it's only used once in the . Also, we don't really need to give downstream users the possibility to write to the datetime pointer. Let me suggest something directly in the `datetime` bindings...

pyo3-ffi/README.md
```diff
@@ -0,0 +1,188 @@
...
+    let version = env!("CARGO_PKG_VERSION");
+    PyModule_AddObject(
+        mptr,
+        "__version__\0".as_ptr() as *const c_char,
+        PyUnicode_FromStringAndSize(version.as_ptr() as *const c_char, version.len() as isize),
+    );
+
+    let wrapped_sum_as_string = PyMethodDef {
+        ml_name: "sum_as_string\0".as_ptr() as *const c_char,
+        ml_meth: Some(transmute::<_PyCFunctionFast, PyCFunction>(sum_as_string)),
```
mejrs, line 79: This transmute looks a bit fishy, can you add a comment explaining why this is sound? Transmuting function pointers (and pyo3 users copypasting unsafe code they might not understand) always makes me nervous. This also fails to compile.

pyo3-ffi/README.md
```diff
@@ -0,0 +1,188 @@
...
+) -> *mut PyObject {
+    if nargs != 2 {
+        return raise_type_error("sum_as_string() expected 2 positional arguments");
+    }
+
+    let arg1 = *args;
+    if PyLong_Check(arg1) == 0 {
+        return raise_type_error("sum_as_string() expected an int for positional argument 1");
+    }
+    let arg1 = PyLong_AsLong(arg1);
```
mejrs, line 123: `PyLong_AsLong` will return an error on overflow, which must be handled (same for the other one)

pyo3-ffi/src/datetime.rs
```diff
@@ -434,62 +431,29 @@ pub struct PyDateTime_CAPI {
...
-/// from the equivalent C object: in C, this is implemented as a `static PyDateTime_CAPI *`. Here
-/// this is implemented as a wrapper which implements [`Deref`] to access a reference to a
-/// [`PyDateTime_CAPI`] object.
-///
-/// In the [`Deref`] implementation, if the underlying object has not yet been initialized, the GIL
-/// will automatically be acquired and [`PyDateTime_IMPORT()`] called.
-pub static PyDateTimeAPI: _PyDateTimeAPI_impl = _PyDateTimeAPI_impl {
-    inner: GILOnceCell::new(),
-};
+pub static PyDateTimeAPI: UnsafeGILCell<PyDateTime_CAPI> = UnsafeGILCell::new(ptr::null_mut());
```
davidhewitt, line 434: How about we make this static not `pub`, and then we have `pub fn PyDateTimeAPI() -> *mut PyDateTime_CAPI` to read the value of it?

The implementation itself can be as simple as

```rust
struct PyDateTimeAPISingleton(UnsafeCell<*mut PyDateTime_CAPI>);
unsafe impl Sync for PyDateTimeAPISingleton;

static PyDateTimeAPI_impl: PyDateTimeAPISingleton = PyDateTimeAPISingleton(UnsafeCell::new());

#[inline]
pub unsafe fn PyDateTimeAPI() -> *mut PyDateTime_CAPI {
    *PyDateTimeAPISingleton().get()
}
```

... and then `PyDateTime_IMPORT` in this file is the only place where this is ever written.

src/types/datetime.rs
```diff
@@ -26,8 +25,62 @@ use crate::ffi::{
...
+//
+// These are bindings around the C API typecheck macros, all of them return
+// `1` if True and `0` if False. In all type check macros, the argument (`op`)
+// must not be `NULL`. The reexported implementations here call ensure_datetime_api is initalized
+//
+// # Safety
+//
+// These functions must only be called when the GIL is held!
+
+macro_rules! reexport_with_autoinit {
```
davidhewitt, line 49: as these aren't re-exports any more, just wrappers, might be nice to give the macro a different name.

src/types/datetime.rs
```diff
@@ -111,7 +164,9 @@ pub struct PyDate(PyAny);
 pyobject_native_type!(
     PyDate,
     crate::ffi::PyDateTime_Date,
-    *PyDateTimeAPI.DateType,
+    // REVIEW: this is save because this is only ever called when a py object is aqured we just
+    // have no way to access it due to the macro
```
davidhewitt, line 168: Yep agreed on this (and in all cases below too).

davidhewitt (commented):
Thanks, this is shaping up nicely.

Some additional comments added. Also, some housekeeping:
 - The user-facing changes will need a CHANGELOG entry please.
 - It would be great to update `Architecture.md` and `Contributing.md` where relevant.
 - The `coverage` CI in `xtask/src/main.rs` will need to be updated to include this new package in coverage numbers.

pyo3-ffi/src/lib.rs
DSPOM2: My Intention with the pub cell was to let downstream crates (of pyo3_ffi) use the same cell. While this is trivial to implement it would be a nice QOL addition because having mutable statics for exactly the same purpose as here (loading some python type/api capsule) is a pretty common pattern if one uses raw ffi. However I your solution below is much nicer and in the end my UnsafeGILCell is probably not gonna cover all needs of downstream crates anyway and those crates should probably just use similar wrapper types instead

pyo3-ffi/src/lib.rs
DSPOM2: After the feedback below I decided not to expose this type after all (see comment below)

DSPOM2:
Thanks for the feedback. I will implement your suggestions + housekeeping tomorrow.

In regards to the Changelog: I noticted that the removal of the `PyDateTime_TimeZone_UTC` removes the only safe way to access `PyDateTime_TimeZone_UTC`. While this was hidden in the ffi module and probably rarely used this was the only safe way to access `PyDateTime_TimeZone_UTC` with pyo3 (except for calling back into python). Should some safe replacement be provdied somewhere or ist that okay?

davidhewitt:
Don't worry about `PyDateTime_TimeZone_UTC` - I'll make a note to myself to rebase part of #1588 after this PR lands to add that back in.

pyo3-ffi/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+As well as with `maturin`, it is possible to build using [setuptools-rust] or
+[manually][manual_builds]. Both offer more flexibility than `maturin` but require further
+configuration.
+
+
+While most projects use the save wrapper provided by pyo3,
+you can take a look at the [`orjson`] as an example on how to use ffi directly.
+For those well versed in C and Rust the [tutorials] from the cpython documentation
+can be easily converted to rust aswell.
```
mejrs, line 177: Thanks for including this, this is a great example :)

mejrs (changes requested):
Thanks, this looks a lot better :)

> Writing these kinds of docs is not my strength so I would welcome some feedback 😄

I think they're pretty good, thanks! Don't let perfect be the enemy of good.

pyo3-ffi/README.md
```diff
@@ -0,0 +1,188 @@
...
+        mptr,
+        "__version__\0".as_ptr() as *const c_char,
+        PyUnicode_FromStringAndSize(version.as_ptr() as *const c_char, version.len() as isize),
+    );
+
+    let wrapped_sum_as_string = PyMethodDef {
+        ml_name: "sum_as_string\0".as_ptr() as *const c_char,
+        ml_meth: Some(transmute::<_PyCFunctionFast, PyCFunction>(sum_as_string)),
+        ml_flags: METH_FASTCALL,
+        ml_doc: "returns the sum of two integers as a string".as_ptr() as *const c_char,
```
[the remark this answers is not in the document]
DSPOM2: good catch. Thanks :) I usually use a macro for static c strings but didn't want to make this example more complicated :D

pyo3-ffi/src/datetime.rs
mejrs: What about providing a macro to do this? There's quite a bit of code around that uses `static mut` to store pointers to Python objects. E.g. https://github.com/ijl/orjson/blob/master/src/typeref.rs#L24 

(this is probably out of scope for this PR, but perhaps a nice future addition?)

pyo3-ffi/README.md
DSPOM2: Sure I will add a comment. Essentially this is save because we add the appropriate [`METH_FASTCALL`](https://docs.python.org/3/c-api/structures.html#METH_FASTCALL) flag below. Essentially there are different types of functions signatures accepted by python (which does the exact same pointer cast just backwards when the flag is specified).
In regards to your compiler failure: I am guessing that you are either using master pyo3 or didn't pull my newest changes. I changed the defintion of `_PyCFunctionFast` in the FFI bindings because they were actually incorrect (see above and compare the changes I made to the defintion to the one in the [python docs](https://docs.python.org/3/c-api/structures.html#c._PyCFunctionFast)). If this is not the cause could you send me your exact error because I actually tested and ran this example locally (and it worked :D)

pyo3-ffi/README.md
DSPOM2: Its where I took a lot of inspiration from when I was working on the project which motivated this PR :) It's indeed very helpful

pyo3-ffi/README.md
DSPOM2: Good catch. Will fix this. Thanks :+1:

DSPOM2:
> I think they're pretty good, thanks! Don't let perfect be the enemy of good.

Thanks, I give it my best :) I haven't done this kind of contribution before so its just a bit of imposter syndrome.

pyo3-ffi/src/datetime.rs
DSPOM2: Something similar to orjson is exactly my kind of usecase :) A macro would be nice but I agree its probably better for a followup PR

pyo3-ffi/README.md
mejrs: > I changed the defintion of _PyCFunctionFast in the FFI bindings

Ah, I was just importing `pyo3::fii::*;` from main, that makes sense.

pyo3-ffi/README.md
mejrs: I just noticed - it should be "safe" instead of "save" 😉

birkenfeld: ... and "aswell" is not a word.

pyo3-ffi/src/datetime.rs
DSPOM2: >  f I squint a bit I also think we could have static PyDateTimeAPI: PyDateTimeApiSingleton and PyDateTimeApiSingleton has a Deref impl which produces &*mut PyDateTime_CAPI. That might feel more like the Python FFI, as PyDateTimeAPI is then a static rather than a function?!

I opted not to do this because reading from the UnsafeCell is always unsafe (because even for reading the caller has to guarantee that the GIL is held). Instead I opted for your former suggestion

DSPOM2:
Sorry for the delay some unexpected things came up yesterday.
I have applied the fixes as suggested.

> It would be great to update Architecture.md and Contributing.md where relevant.

I found nothing relevant to update in `Contributing.md `. Did you have anyhing specific in mind?

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `pyo3-build-config` no longer enables the `resolve-config` feature by default. [#2008](https://github.com/PyO3/pyo3/pull/2008)
 - Update `inventory` optional dependency to 0.2. [#2019](https://github.com/PyO3/pyo3/pull/2019)
 - Drop `paste` dependency. [#2081](https://github.com/PyO3/pyo3/pull/2081)
+- The ffi bindings found in the previously in the `pyo3::ffi` module are now also available as the
+  separate `pyo3-ffi` crate. [#2126](https://github.com/PyO3/pyo3/pull/2126)
```
davidhewitt, line 20: They're still available in `pyo3::ffi`, so maybe:

```suggestion
- The bindings found `pyo3::ffi` are now a re-export of the new `pyo3-ffi` crate. [#2126](https://github.com/PyO3/pyo3/pull/2126)
```

davidhewitt (approved):
Thanks, this is looking great to me! @mejrs any further changes you want before this is merged?

> I found nothing relevant to update in `Contributing.md`. Did you have anyhing specific in mind?

Nope that's fine, I was just listing off a few files which I thought would be worth checking in the light of a big structural change like this :)

I just have one tiny nit on the CHANGELOG...

mejrs (approved):
> any further changes you want before this is merged?

LGTM :)

Thanks for doing all this work @DSPOM2

DSPOM2:
Perfect. I will apply the change to the changelog and then squash all the commits tomorrow. Then we are good to go :) 
 I noticted that both setup-tools-rust and maturin require patches to allow building wheels that only depend on pyo3-ffi and not on pyo3. I will try to do a PRs for that aswell but I have quite a lot on my plate so not sure how soon I will manage that

CHANGELOG.md
```diff
@@ -50,6 +52,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `#[pyclass]` type object creation [#2076](https://github.com/PyO3/pyo3/pull/2076) [#2081](https://github.com/PyO3/pyo3/pull/2081)
 - `__ipow__` now supports modulo argument on Python 3.8+. [#2083](https://github.com/PyO3/pyo3/pull/2083)
 - `pyo3-macros-backend` is now compiled with PyO3 cfgs to enable different magic method definitions based on version. [#2083](https://github.com/PyO3/pyo3/pull/2083)
+- `_PyCFunctionFast` now correctly reflects the signature defined in the [python docs](https://docs.python.org/3/c-api/structures.html#c._PyCFunctionFast). [#2126](https://github.com/PyO3/pyo3/pull/2126)
```
birkenfeld, line 55:
```suggestion
- `_PyCFunctionFast` now correctly reflects the signature defined in the [Python docs](https://docs.python.org/3/c-api/structures.html#c._PyCFunctionFast). [#2126](https://github.com/PyO3/pyo3/pull/2126)
```

CHANGELOG.md
```diff
@@ -50,6 +52,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `PyDateTimeAPI` and `PyDateTime_TimeZone_UTC` are are now unsafe functions instead of statics. [#2007](https://github.com/PyO3/pyo3/pull/2007)
```
birkenfeld, line 56: PR number seems wrong, did you mean 2126?
```suggestion
- `PyDateTimeAPI` and `PyDateTime_TimeZone_UTC` are are now unsafe functions instead of statics. [#2126](https://github.com/PyO3/pyo3/pull/2126)
```

CHANGELOG.md
```diff
@@ -50,6 +52,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `PyDateTimeAPI` does not implicitly call `PyDateTime_IMPORT` anymore to reflect the original python API more closely. Before the first call to `PyDateTime_IMPORT` a null pointer is returned. Therefore before calling any of the following ffi functions `PyDateTime_IMPORT` must be called to avoid undefined behaviour: [#2007](https://github.com/PyO3/pyo3/pull/2007)
```
birkenfeld, line 57: Same here
```suggestion
- `PyDateTimeAPI` does not implicitly call `PyDateTime_IMPORT` anymore to reflect the original Python API more closely. Before the first call to `PyDateTime_IMPORT` a null pointer is returned. Therefore before calling any of the following FFI functions `PyDateTime_IMPORT` must be called to avoid undefined behaviour: [#2126](https://github.com/PyO3/pyo3/pull/2126)
```

Cargo.toml
```diff
@@ -60,16 +63,16 @@ multiple-pymethods = ["inventory", "pyo3-macros/multiple-pymethods"]
...
-abi3 = ["pyo3-build-config/abi3"]
+abi3 = ["pyo3-build-config/abi3", "pyo3-ffi/abi3"]
 
 # With abi3, we can manually set the minimum Python version.
-abi3-py37 = ["abi3-py38", "pyo3-build-config/abi3-py37"]
-abi3-py38 = ["abi3-py39", "pyo3-build-config/abi3-py38"]
-abi3-py39 = ["abi3-py310", "pyo3-build-config/abi3-py39"]
-abi3-py310 = ["abi3", "pyo3-build-config/abi3-py310"]
+abi3-py37 = ["abi3-py38", "pyo3-build-config/abi3-py37", "pyo3-ffi/abi3-py37"]
+abi3-py38 = ["abi3-py39", "pyo3-build-config/abi3-py38", "pyo3-ffi/abi3-py37"]
```
birkenfeld, line 73: Should this be `abi3-py38` (etc. on the the following lines)?

pyo3-ffi/README.md
```diff
@@ -0,0 +1,199 @@
+# pyo3-ffi
+
+This crate provides [Rust](https://www.rust-lang.org/) FFI declarations for Python 3.
+It supports both the stable and the unstable component of the ABI trough the use of cfg flags.
```
birkenfeld, line 4:
```suggestion
It supports both the stable and the unstable component of the ABI through the use of cfg flags.
```

pyo3-ffi/README.md
```diff
@@ -0,0 +1,199 @@
...
+        m_methods: std::ptr::null_mut(),
+        m_slots: std::ptr::null_mut(),
+        m_traverse: None,
+        m_clear: None,
+        m_free: None,
+    };
+
+    let mptr = PyModule_Create(Box::into_raw(Box::new(init)));
+    let version = env!("CARGO_PKG_VERSION");
+    PyModule_AddObject(
```
birkenfeld, line 69: Can `PyModule_AddObject` not fail?

pyo3-ffi/README.md
```diff
@@ -0,0 +1,199 @@
...
+>>> string_sum.sum_as_string(5, 20)
+'25'
+```
+
+As well as with `maturin`, it is possible to build using [setuptools-rust] or
+[manually][manual_builds]. Both offer more flexibility than `maturin` but require further
+configuration.
+
+
+While most projects use the save wrapper provided by pyo3,
```
birkenfeld, line 185:
```suggestion
While most projects use the safe wrapper provided by PyO3,
```

pyo3-ffi/README.md
```diff
@@ -0,0 +1,199 @@
...
+you can take a look at the [`orjson`] as an example on how to use ffi directly.
+For those well versed in C and Rust the [tutorials] from the cpython documentation
```
birkenfeld, line 187:
```suggestion
For those well versed in C and Rust the [tutorials] from the CPython documentation
```

pyo3-ffi/src/lib.rs
```diff
@@ -0,0 +1,430 @@
...
+//! although null pointers are sometimes valid input.
+//! - The vast majority can only be used safely while the GIL is held.
+//! - Some functions have additional safety requirements, consult the
+//! [Python/C API Reference Manual][capi]
+//! for more information.
+//!
+//!
+//! # Feature flags
+//!
+//! PyO3 uses [feature flags] to enable you to opt-in to additional functionality.For a detailed
```
birkenfeld, line 27:
```suggestion
//! PyO3 uses [feature flags] to enable you to opt-in to additional functionality. For a detailed
```

pyo3-ffi/src/lib.rs
```diff
@@ -0,0 +1,430 @@
...
+//! pub unsafe extern "C" fn sum_as_string(
+//!     _self: *mut PyObject,
+//!     args: *mut *mut PyObject,
+//!     nargs: Py_ssize_t,
+//! ) -> *mut PyObject {
+//!     if nargs != 2 {
+//!         return raise_type_error("sum_as_string() expected 2 positional arguments");
+//!     }
+//!
+//!    let arg1 = *args;
```
birkenfeld, line 160: wrong indentation (3 spaces) from here on

src/types/datetime.rs
```diff
@@ -26,8 +25,64 @@ use crate::ffi::{
...
+    unsafe {
+        if pyo3_ffi::PyDateTimeAPI().is_null() {
+            PyDateTime_IMPORT()
+        }
+
+        &*pyo3_ffi::PyDateTimeAPI()
+    }
+}
+
+/// Type Check macros
```
birkenfeld, line 39: This doesn't seem to be ta docstring for the macro?

pyo3-ffi/src/lib.rs
```diff
@@ -0,0 +1,430 @@
...
+//!
+//!     let mptr = PyModule_Create(Box::into_raw(Box::new(init)));
+//!     let version = env!("CARGO_PKG_VERSION");
+//!     PyModule_AddObject(
+//!         mptr,
+//!         "__version__\0".as_ptr() as *const c_char,
+//!         PyUnicode_FromStringAndSize(version.as_ptr() as *const c_char, version.len() as isize),
+//!     );
+//!
+//!     // Its necessary to transmute `sum_as_string` here because functions marked with `METH_FASTCALL`
```
birkenfeld, line 114:
```suggestion
//!     // It's necessary to transmute `sum_as_string` here because functions marked with `METH_FASTCALL`
```

pyo3-ffi/src/lib.rs
```diff
@@ -0,0 +1,430 @@
...
+//! To embed Python into a Rust binary, you need to ensure that your Python installation contains a
+//! shared library. The following steps demonstrate how to ensure this (for Ubuntu).
+//!
+//! To install the Python shared library on Ubuntu:
+//! ```bash
+//! sudo apt install python3-dev
+//! ```
+//!
+//! While most projects use the safe wrapper provided by pyo3,
+//! you can take a look at the [`orjson`] as an example on how to use ffi directly.
```
birkenfeld, line 232:
```suggestion
//! you can take a look at the [`orjson`] library as an example on how to use `pyo3-ffi` directly.
```

pyo3-ffi/src/lib.rs
```diff
@@ -0,0 +1,430 @@
...
+//! For those well versed in C and Rust the [tutorials] from the cpython documentation
```
birkenfeld, line 233:
```suggestion
//! For those well versed in C and Rust the [tutorials] from the CPython documentation
```

pyo3-ffi/README.md
```diff
@@ -0,0 +1,199 @@
...
+'25'
+```
+
+As well as with `maturin`, it is possible to build using [setuptools-rust] or
+[manually][manual_builds]. Both offer more flexibility than `maturin` but require further
+configuration.
+
+
+While most projects use the save wrapper provided by pyo3,
+you can take a look at the [`orjson`] as an example on how to use ffi directly.
```
birkenfeld, line 186:
```suggestion
you can take a look at the [`orjson`] library as an example on how to use `pyo3-ffi` directly.
```

birkenfeld (changes requested):
Looks very good! Mostly copy-edits, a few questions.

CHANGELOG.md
DSPOM2: you are right was copy past error :)

pyo3-ffi/README.md
DSPOM2: Yes it can technically fail. Practically given the inputs here it can never fail (I just checked the cpython sourcecode again, so long as name is a valid string, mod is valid module with a dict and value is not null this can never fail) so handling the error here just needlessly complicates the example so I choose to omit it.

src/types/datetime.rs
DSPOM2: I added this doc comment to the macro because it seemed the best place to put it (it describes the function the macro generates) but I can make it a comment instead if you think that would be clearer

Cargo.toml
DSPOM2: ah yes. Good catch!

DSPOM2:
Thanks for the thorough review :) 
I added 2 comments. Once we address those I will squash the commits

pyo3-ffi/README.md
birkenfeld: Looking at the CPython sources they also seem pretty inconsistent about checking the return value. Of course, `PyUnicode_FromStringAndSize` can *also* theoretically fail. Do you want to add a comment at least?

src/types/datetime.rs
birkenfeld: Well, in the rendered docs it will show up as doc for the macro called ffi_fun_with_autoinit which seems to have nothing to do with type checking.

pyo3-ffi/README.md
DSPOM2: Yeah adding a comment sounds like a good idea so users are aware that they might need to check for errors in other circumstances

src/types/datetime.rs
DSPOM2: Okay I will put it as a comment then so its lets confusing in the docs

birkenfeld (approved):

DSPOM2:
commits are squashed and I think this is ready to merge now now :smile:

DSPOM2:
The windows coverage job keeps failing but I doesn't sound related to this PR... Not too sure what going on there

davidhewitt:
Agreed. Thanks again everyone for working on this!

ijl:
@DSPOM2 thanks for adding this. I have a proposal to add it to maturin in pyo3/maturin#804.

PR #2127, opened 2022-01-26
ci: unpin nightly for coverage

davidhewitt:
As per https://github.com/rust-lang/rust/issues/79645, latest nightly should hopefully be working for coverage again...


davidhewitt:
Drop in coverage seems unfortunate, but from looking at the codecov this does seem to be more correct. Will proceed to merge.

PR #2133, opened 2022-01-29
Add Mapping::contains

Gobot1234:
Pretty self explainitory.

This function like the `collections.abc.Mapping` equivalent cannot fail (https://github.com/python/cpython/blob/8fb36494501aad5b0c1d34311c9743c60bb9926c/Lib/_collections_abc.py#L823-L829).


davidhewitt:
The analogy with `collections.abc.Mapping` is reasonable, however it makes me uncomfortable too. This isn't equivalent to the `in` operator; we just merged #2115 which added `PyAny::contains`. Given that, do we really need to add this too?

In particular, the thing I don't like about `PyMapping_HasKey` (and particular wrapping it in a method called `contains`) is that it looks like it should call `__contains__`, but it does not.

Gobot1234:
Ok fair enough, what about just making it call PyAny::contains?

davidhewitt:
I'd prefer that :)

davidhewitt (approved):
Thanks!

PR #2134, opened 2022-01-31
opt: improve generated code for argument extraction

davidhewitt:
I had a bit of fun yesterday playing around with the generated code for function argument extraction. There's three main things that I've changed:

- The "fastcall" and "normal" methods of argument extraction have less of a shared implementation. This allows them to make specific optimizations unique to each. e.g. "fastcall" can make use of `copy_from_slice`, "normal" can use `PyTuple::get_slice` to create varargs tuple.
- The argument extraction methods have been changed so that the decision whether to accept varargs and varkeywords is done at compile-time (by a pair of traits) rather than at runtime (by checking booleans in `FunctionDescription` struct).
- I've changed the generated code which extracts individual arguments to always use helper functions defined in `src/impl_/extract_argument.rs`. The current generated code uses `Option::map` etc directly to implement various combinations of (required / defaulted) extraction, however this came at a code size cost because the common logic was duplicated across all generated code.

Overall, has a modest effect on generated code size, and also local benchmarking suggests that there are some (very small) performance gains won here.

### DIFF
--- a/src/impl_/extract_argument.rs
+++ b/src/impl_/extract_argument.rs
@@ -6,6 +6,7 @@ use crate::{
     FromPyObject, PyAny, PyErr, PyResult, Python,
 };
 
+/// The standard implementation of how PyO3 extracts a `#[pyfunction]` or `#[pymethod]` function argument.
 #[doc(hidden)]
 #[inline]
 pub fn extract_argument<'py, T>(obj: &'py PyAny, arg_name: &str) -> PyResult<T>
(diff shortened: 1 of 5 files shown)


src/impl_/extract_argument.rs
```diff
@@ -6,18 +6,92 @@ use crate::{
...
+/// - `obj` must not be None (this helper is only used for required function arguments).
 #[doc(hidden)]
 #[inline]
-pub fn extract_argument<'py, T>(obj: &'py PyAny, arg_name: &str) -> PyResult<T>
+pub unsafe fn extract_argument<'py, T>(obj: Option<&'py PyAny>, arg_name: &str) -> PyResult<T>
 where
     T: FromPyObject<'py>,
 {
+    // Safety: obj is not None (see safety requirement above).
+    // The unwrap is done here rather than by the caller (generated proc macro code) to minimize the amount of
```
adamreichold, line 20: But should it also be possible to put the `unwrap_unchecked` into the generated code which should not actually generate any machine code (besides offsetting a pointer or something, but not calling into the panic machinery) so that this function would stay safe and generated code could just use an unsafe block?

davidhewitt: Yes, I'll have a go at that when I next get a chance 👍

It might be a bit more fiddly - the bit missed out in this PR at the moment is that `.unwrap_unchecked()` is only on 1.58 - so I'll need some `cfg` to support most of our users (probably with `.expect()`).

mejrs: > .unwrap_unchecked() is only on 1.58

You can use `.unwrap_or_else(|| unsafe { unreachable_unchecked() })`

mejrs: Or if you're a bit more nervous, like I am:
```rust
let obj = match obj{
   Some(obj) => obj,
   #[cfg(debug_assertions)]
   None => unreachable!(),
   #[cfg(not(debug_assertions))]
   None => unsafe { hint::unreachable_unchecked() },
}

Comment (mejrs) [reply src/impl_/extract_argument.rs]: @davidhewitt that is actually not quite what I meant. There is no need to use the cfg or unwrap_unchecked, you can just use either of the expressions above whether or not we are on 1.58. That is literally how unwrap_unchecked works:

```rust
    pub unsafe fn unwrap_unchecked(self) -> T {
        debug_assert!(self.is_some());
        match self {
            Some(val) => val,
            // SAFETY: the safety contract must be upheld by the caller.
            None => unsafe { hint::unreachable_unchecked() },
        }
    }
```

davidhewitt: That's fair. My thinking was that `unwrap_unchecked()` leads to the simplest generated code. IMHO the less code generated by the macro the better, so using the `#[cfg]` to get the desired "end state" for 1.58 and up seemed worth it?

mejrs: What about a helper function?

```rust
// somewhere in internal_tricks...

pub unsafe fn unwrap_unchecked<T>(t: Option<T>) -> T {
    match t {
        Some(obj) => obj,
        #[cfg(debug_assertions)]
        None => unreachable!(),
        #[cfg(not(debug_assertions))]
        None => std::hint::unreachable_unchecked(),
    }
}
```

My main concerns with the current implementation are:

1) I feel it's overkill to use cfgs and create a build script for a convenience method around a match statement.
2) I also feel uncomfortable with the scope of the unsafe here, it's actually quite hard to see if this really is ok. I'd sleep better at night knowing there's a debug assertion in there. (the debug_assertion in unwrap_unchecked only works if you `build-std` in debug mode...)

davidhewitt: Agreed the helper is much better than what I last pushed.

I almost feel like this has come full circle and we should consider again making `extract_argument` unsafe and take the `Option` - that way the macro expansion inserts as little code as possible, and the assertion message on failure in debug can be more informative than generic "unreachable".

adamreichold: I am not sure we are talking about the same kind of code. Are you concerned about the syntax level code that the compiler will have to process after the proc-macro was expanded or about the machine code generated due to calling into the panic machinery?

If it is the syntax-level code, then moving everything into `extract_argument` would probably be the way to go. If it is the machine code, the helper function seems as nicer as it keeps `extract_argument` safe. I also don't think it needs to be a generic `unwrap_unchecked` function, but could also be `unwrap_argument` with an informative error message.

But I also think this is a pretty small concern at this point in either direction. For example, keeping the checked assertions in debug mode seems more important in my opinion.

davidhewitt: I was thinking at syntax level; I've recently been staring at some huge PyO3-based crates with compile times in the minutes, so reducing the amount of work the macros put on the compiler has been front of my mind.

However agreed it's likely to be such a small difference in this case. I'm persuaded we should go with the helper function for clarity. Will probably have a chance to update this PR tomorrow sometime 👍

PR #2147, opened 2022-02-05
example: add cargo test which doesn't link

davidhewitt:
@aviramha - try running `cargo test` in `examples/maturin-starter`. You should get the linker error which #2135 can resolve.


davidhewitt:
@aviramha as this PR served its purpose (I think) I'm gonna kill it.

PR #2156, opened 2022-02-08
Add examples for object and numeric protocols

mejrs:
Todo:

- [x]  Add tests for all of this


guide/src/class.md
```diff
@@ -41,20 +39,59 @@ Because Python objects are freely shared between threads by the Python interpret
 
 The above example generates implementations for [`PyTypeInfo`], [`PyTypeObject`], and [`PyClass`] for `MyClass` and `MyEnum`. To see these generated implementations, refer to the [implementation details](#implementation-details) at the end of this chapter.
 
+## Constructor
+
+By default it is not possible to create an instance of a custom class from Python code.
+To declare a constructor, you need to define a method and annotate it with the `#[new]`
+attribute. Only Python's `__new__` method can be specified, `__init__` is not available.
```
davidhewitt, line 46: (I know this is moved from lower down...)

I would love for us to think about whether we could / should support `__init__` in the future. Think it's worth making an issue and linking it here for interested users to join discussion?

guide/src/class.md
```diff
@@ -23,12 +23,10 @@ This chapter will discuss the functionality and configuration these attributes o
...
 # #![allow(dead_code)]
-# use pyo3::prelude::*;
+use pyo3::prelude::*;
+
 #[pyclass]
-struct MyClass {
-    # #[pyo3(get)]
-    num: i32,
-}
+struct Number(i32);
```
davidhewitt, line 29: This is the first `#[pyclass]` example most users will see. I tend to think structs with named fields are more typical than tuple structs (although tuple structs definitely have their place). 

Maybe have examples of all three - named fields, tuple struct, and enum?

Then the rest of the examples can proceed to use the tuple struct.

guide/src/class.md
```diff
@@ -41,20 +39,59 @@ Because Python objects are freely shared between threads by the Python interpret
 
 The above example generates implementations for [`PyTypeInfo`], [`PyTypeObject`], and [`PyClass`] for `MyClass` and `MyEnum`. To see these generated implementations, refer to the [implementation details](#implementation-details) at the end of this chapter.
 
+## Constructor
```
davidhewitt, line 42: 👍 agree this should come earlier!

guide/src/class/numeric.md
```diff
@@ -0,0 +1,373 @@
+# Emulating numeric types
```
davidhewitt, line 1: I really like how this new page reads. It tells a nice story along the lines of "implementing your first PyO3 class". 

That said, the existing `class.md` is somewhere between a reference and a story itself. I'd love it if we could make the distinction clearer in splitting out the reference bits of `class.md` into a `#[pyclass]` attribute reference and then have a whole "getting started with PyO3 classes" story (of which this page is one part).

guide/src/class/numeric.md
```diff
@@ -0,0 +1,373 @@
...
+OverflowError: Python int too large to convert to C long
+```
+
+Instead of relying on the default [`FromPyObject`] extraction to parse arguments, we can specify our
+own extraction function, using the `#[pyo3(from_py_with = "...")]` attribute. Unfortunately PyO3
+ doesn't provide a way to wrap Python integers out of the box, but there's a Python C API function that does:
+
+```c
+unsigned long PyLong_AsUnsignedLongMask(PyObject *obj)
+```
```
davidhewitt, line 35: Should we add one? It might be nice to not encourage beginners to use `unsafe`.

guide/src/class/numeric.md
```diff
@@ -0,0 +1,373 @@
...
+                return Err(err);
+            }
+        }
+
+        Ok(ret as i32)
+    }
+}
+
+/// Did you ever hear the tragedy of Darth Signed The Overfloweth? I thought not.
+/// It's not a story C would tell you. It's a Rust legend.
```
davidhewitt, line 76: Haha 🤣

guide/src/class/numeric.md
```diff
@@ -0,0 +1,373 @@
...
+use pyo3::exceptions::{PyZeroDivisionError, PyValueError};
+
+# use pyo3::prelude::*;
+#
+# #[pyclass]
+# struct Number(i32);
+#
+#[pymethods]
+impl Number {
+    fn __add__(&self, other: &Self) -> Self {
```
davidhewitt, line 103: Aside: because this type wraps `i32` it's probably slightly more efficient to pass by `self`. But I don't think we support that. Might be nice to do so (maybe for `Copy` types only)?

guide/src/class/numeric.md
```diff
@@ -0,0 +1,373 @@
...
+    fn __neg__(&self) -> Self {
+        Self(-self.0)
+    }
+
+    fn __abs__(&self) -> Self {
+        Self(self.0.abs())
+    }
+
+    fn __invert__(&self) -> Self {
+        Self(!self.0)
```
davidhewitt, line 168: TIL (that `!` is bitwise not when used on integer types).

guide/src/class/object.md
```diff
@@ -0,0 +1,232 @@
...
+    fn __richcmp__(&self, other: &Self, op: CompareOp) -> PyResult<bool> {
+        match op {
+            CompareOp::Lt => Ok(self.0 < other.0),
+            CompareOp::Le => Ok(self.0 <= other.0),
+            CompareOp::Eq => Ok(self.0 == other.0),
+            CompareOp::Ne => Ok(self.0 != other.0),
+            CompareOp::Gt => Ok(self.0 > other.0),
+            CompareOp::Ge => Ok(self.0 >= other.0),
+        }
+    }
```
davidhewitt, line 151: I look forward to making this nicer soon :)

davidhewitt (commented):
Lovely to see additional docs and examples added, thank you as always! Sprinkled some comments :)

guide/src/class/numeric.md
mejrs: I think even if that was supported we shouldn't do it, because if users copy code and try that with non-copy types it'll throw some cryptic compiler errors.

guide/src/class.md
mejrs: Sure, I'll show all forms 👍

guide/src/class.md
mejrs: Is there an use case for it?

Given that Rust doesn't really do partial initialization (there's no clever constructor stuff, everything just bottoms out at creating a struct and initializing all its fields at once), it doesn't map well to that either.

I think it's better to not have an issue for now. We can discuss it on an issue if/when someone makes one.

guide/src/class/numeric.md
mejrs: I remember being a bit annoyed with this myself, when trying to figure out how to make something work I had to go back and forth between the book and the api documentation which was quite annoying.

> splitting out

I think a bit of duplication is also fine, it doesn't need to be a hard split.

guide/src/class/numeric.md
mejrs: They can go through the python apis to do `x & 0xFFFFFFFF` before extracting. So strictly speaking unsafe is not necessary.

Maybe I should do that and make an appendix with the unsafe way later.

guide/src/class/numeric.md
davidhewitt: Agreed its impossible to avoid duplication; I guess what I meant by splitting out is that the reference portion would always contain full detail, the story might introduce the key elements and then just link to the reference portion for readers who want to see the whole lot.

guide/src/class/numeric.md
davidhewitt: Ok 👍

guide/src/class.md
```diff
@@ -23,38 +23,84 @@ This chapter will discuss the functionality and configuration these attributes o
...
 }
 
+// A "tuple" struct
 #[pyclass]
-enum MyEnum {
-    Variant,
-    OtherVariant = 30, // PyO3 supports custom discriminants.
+struct Number(i32);
+
+// PyO3 supports custom discriminants.in enums
```
davidhewitt, line 37:
```suggestion
// PyO3 supports custom discriminants in enums
```

guide/src/class/object.md
```diff
@@ -58,7 +58,7 @@ impl Number {
         // format string, followed by any number of parameters which replace the
         // `{}`'s in the format string.
         //
-        // `.0` is how in Rust you get the first value of a tuple struct.
+        //                       👇 Tuple field access in Rust uses a dot
```
davidhewitt, line 61: Haha love this!

davidhewitt (approved):
LGTM; feel free to merge unless you plan to add more to this branch? One tiny typo spotted...

guide/src/class/object.md
mejrs: I saw this in a https://fasterthanli.me blog the other day and liked it. I figured I'd `~~steal~~ ~~copy~~ learn from him :)

mejrs:
I don't understand why this test is failing.

guide/src/class/numeric.md
```diff
@@ -373,8 +373,12 @@ fn my_module(_py: Python, m: &PyModule) -> PyResult<()> {
 # assert float(Number(13)) == 13
 # assert Number.__doc__ == "Did you ever hear the tragedy of Darth Signed The Overfloweth? I thought not.\nIt's not a story C would tell you. It's a Rust legend."
 # assert Number(12345234523452) == Number(1498514748)
-# import inspect
-# assert inspect.signature(Number).__str__() == '(int)'
+# try:
+#     import inspect
+#     assert inspect.signature(Number).__str__() == '(int)'
+# except ValueError:
+#     # Not supported on all Python versions
```
davidhewitt, line 380:
```suggestion
#     # Not supported with `abi3` before Python 3.10
```

https://pyo3.rs/v0.15.1/building_and_distribution.html#missing-features

davidhewitt (approved):
Thanks for adding the test! One final suggestion on a comment, then let's merge?

davidhewitt:
This'll need a rebase (or merge from master) since #2165 went in. Feel free to do that (maybe squash some of the "try fix CI" commits together) and then merge this at your leisure 👍

PR #2157, opened 2022-02-08
refactor: simpler pyclass internals

davidhewitt:
Just makes a few cleanups to the `#[pyclass]` internals which I noticed were possible the other day:
- Merges `pyclass_default_items` trait into the inlined items created in #2153
- Changes new / free / alloc to all be passed via slots, rather than using autoref specialization.

Overall seems like a nice simplification; makes the machinery of the pyclass internals a bit more uniform. Also reduces `llvm-lines` count of `pytests` crate by around 1%. So nothing major, still seems nice.

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -20,7 +20,6 @@ use syn::{ext::IdentExt, spanned::Spanned, Result};
 pub enum GeneratedPyMethod {
     Method(TokenStream),
     Proto(TokenStream),
-    TraitImpl(TokenStream),
     SlotTraitImpl(String, TokenStream),
 }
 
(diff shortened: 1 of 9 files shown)


pyo3-macros-backend/src/pymethod.rs
```diff
@@ -191,18 +190,18 @@ pub fn impl_py_method_def(
...
-                ::std::option::Option::Some({
-                    #wrapper
-                    #wrapper_ident
-                })
-            }
+    Ok(quote! {{
+        impl #cls {
+            #[doc(hidden)]
+            #wrapper
         }
```
davidhewitt, line 199: Realising that the `wrapper` could be an associated method of the struct is interesting for a couple of reasons:
- We have some rather ugly code to recursively `replace_self` (in method argument types and return values) - if all Pymethod wrappers were placed on the struct like this, we won't need it.
- In addition, because the struct is a shared namespace, we could use this safeguard against multiple Python definitions with the same name. (That's exactly why I did it here; so that `#[new]` can only be implemented once.)

I'll probably play around with this in a follow-up PR.

PR #2159, opened 2022-02-10
pymethods: support gc protocol

davidhewitt:
This is the last step for #1884. Copies the same design for `__traverse__` and `__clear__` from `PyGCProtocol` onto `#[pymethods]`.

It's not the cleanest implementation, but it works fine. Once merged, we can mark `#[pyproto]` as deprecated and get on with releasing 0.16 😄

### DIFF
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -3,7 +3,7 @@
 use crate::attributes::{
     self, take_pyo3_options, CrateAttribute, NameAttribute, TextSignatureAttribute,
 };
-use crate::deprecations::Deprecations;
+use crate::deprecations::{Deprecation, Deprecations};
 use crate::konst::{ConstAttributes, ConstSpec};
 use crate::pyimpl::{gen_default_items, gen_py_const, PyClassMethodsType};
 use crate::pymethod::{impl_py_getter_def, impl_py_setter_def, PropertyType};
@@ -29,12 +29,12 @@ pub struct PyClassArgs {
     pub base: syn::TypePath,
     pub has_dict: bool,
     pub has_weaklist: bool,
-    pub is_gc: bool,
     pub is_basetype: bool,
     pub has_extends: bool,
     pub has_unsendable: bool,
     pub module: Option<syn::LitStr>,
     pub class_kind: PyClassKind,
+    pub deprecations: Deprecations,
 }
 
 impl PyClassArgs {
@@ -63,11 +63,11 @@ impl PyClassArgs {
             base: parse_quote! { _pyo3::PyAny },
             has_dict: false,
             has_weaklist: false,
-            is_gc: false,
             is_basetype: false,
             has_extends: false,
             has_unsendable: false,
             class_kind,
+            deprecations: Deprecations::new(),
         }
     }
 
@@ -158,9 +158,9 @@ impl PyClassArgs {
     fn add_path(&mut self, exp: &syn::ExprPath) -> syn::Result<()> {
         let flag = exp.path.segments.first().unwrap().ident.to_string();
         match flag.as_str() {
-            "gc" => {
-                self.is_gc = true;
-            }
+            "gc" => self
+                .deprecations
+                .push(Deprecation::PyClassGcOption, exp.span()),
             "weakref" => {
                 self.has_weaklist = true;
             }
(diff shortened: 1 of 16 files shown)


pyo3-macros-backend/src/pyclass.rs
```diff
@@ -981,26 +980,6 @@ impl<'a> PyClassImplsBuilder<'a> {
...
-                    use _pyo3::class;
-
-                    fn _assert_implements_protocol<'p, T: _pyo3::class::PyGCProtocol<'p>>() {}
-                    _assert_implements_protocol::<#cls>();
-                }
-            }
-        } else {
-            quote! {}
-        }
-    }
```
davidhewitt, line 1003: This check isn't needed on Python 3.11, however it _is_ needed on older Python versions else `#[pyclass(gc)]` without `__traverse__` will cause segfaults.

I think a better solution might be to deprecate `#[pyclass(gc)]` and then in `src/pyclass.rs` we can check for the presence of `__traverse__` to decide whether to set the `Py_TPFLAGS_HAVE_GC` flag.

davidhewitt:
I'll leave this one up over the weekend in case anyone wants to review! Then some time next week I'll merge, deprecate `#[pyproto]`, and begin preparing the 0.16 release.

birkenfeld:
It seems like you're generating a normal slot for `__clear__`, while under pyproto, `pyo3::class::gc::clear` did not have the callback mechanism, and no way to transport exceptions. Are you sure that exceptions can be raised normally in `tp_clear`? If yes, why do we have the special case for `tp_traverse` (PyTraverseError could be a normal PyError)?

davidhewitt:
Good question. I'm not 100% on this; but what I do see is in the usage of `tp_clear` in the `gc`, there's immediately following a check for errors. 

https://github.com/python/cpython/blob/af32b3ef1fbad3c2242627a14398320960a0cb45/Modules/gcmodule.c#L1014

Whereas uses of `tp_traverse` never seem to check for errors. I get the impression that it's not expected for it to be possible to error during GC traversal.

mejrs:
This piece of documentation still mentions pyclass(gc) and needs deleting: https://github.com/PyO3/pyo3/blob/main/pyo3-macros/src/lib.rs#L90

davidhewitt:
Regarding error messages, it does indeed look like errors in `tp_clear` are allowed. With this test class

```
#[pyclass]
struct ErrorDuringClear {
    self_ref: PyObject,
}

#[pymethods]
impl ErrorDuringClear {
    fn __traverse__(&self, visit: PyVisit) -> Result<(), PyTraverseError> {
        visit.call(&self.self_ref)
    }

    fn __clear__(&mut self, py: Python) -> PyResult<()> {
        self.self_ref = py.None();
        Err(exceptions::PyValueError::new_err("clear failed"))
    }
}
```

I get the following output when the GC runs:

```
Exception ignored in tp_clear of: <class 'builtins.ErrorDuringClear'>
ValueError: clear failed
```

davidhewitt:
> This piece of documentation still mentions pyclass(gc) and needs deleting: https://github.com/PyO3/pyo3/blob/main/pyo3-macros/src/lib.rs#L90

Thanks, pushed!

birkenfeld:
Ok, nice. The other question is then if `tp_traverse` can still execute any Python code (and then ignore the exception), otherwise it probably should be marked unsafe.

davidhewitt:
> Ok, nice. The other question is then if tp_traverse can still execute any Python code (and then ignore the exception), otherwise it probably should be marked unsafe.

I would argue even if it can't execute Python code, it doesn't need to be `unsafe`. I understand `unsafe fn` means it is not safe to call; the implementation could always use `unsafe` internally. In this case it isn't `unsafe` for other Rust code to call a correct implementation of `__traverse__`.

We should definitely document that a correct implementation of `__traverse__` should do nothing more than use `visit.call` on all Python objects contained within the struct.

davidhewitt:
I'm going to merge this shortly and proceed forwards with a final PR to deprecate `#[pyproto]`.

birkenfeld:
All the docs still need to be moved from proto-centric to method-centric, with (maybe?) mentioning the old proto stuff.

davidhewitt:
Agreed, I'll do that in the next PR 👍

birkenfeld:
Are you already working on the PR? Otherwise I'd make a start today.

davidhewitt:
See https://github.com/PyO3/pyo3/pull/2173#issuecomment-1042634725 - please feel free to take over that branch!

PR #2161, opened 2022-02-12
Fix the signature of _PyLong_NumBits

cmpute:
This commit is a simple fix of the signature of _PyLong_NumBits. It's correct C signature in [`longobject.h`](https://github.com/python/cpython/blob/47290e7642dd41d94437dd0e2c0f6bfceb0281b5/Include/longobject.h#L117) is
```c
PyAPI_FUNC(size_t) _PyLong_NumBits(PyObject *v);
```

### DIFF
--- a/src/conversions/num_bigint.rs
+++ b/src/conversions/num_bigint.rs
@@ -111,7 +111,7 @@ macro_rules! bigint_conversion {
                     let num: Py<PyLong> =
                         Py::from_owned_ptr_or_err(py, ffi::PyNumber_Index(ob.as_ptr()))?;
                     let n_bits = ffi::_PyLong_NumBits(num.as_ptr());
-                    let n_bytes = if n_bits == usize::MAX {
+                    let n_bytes = if n_bits == (-1isize as usize) {
                         return Err(PyErr::fetch(py));
                     } else if n_bits == 0 {
                         0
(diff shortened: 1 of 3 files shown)


src/conversions/num_bigint.rs
```diff
@@ -111,12 +111,12 @@ macro_rules! bigint_conversion {
                     let num: Py<PyLong> =
                         Py::from_owned_ptr_or_err(py, ffi::PyNumber_Index(ob.as_ptr()))?;
                     let n_bits = ffi::_PyLong_NumBits(num.as_ptr());
-                    let n_bytes = if n_bits == -1 {
+                    let n_bytes = if n_bits == usize::MAX {
```
davidhewitt, line 114: Maybe the following is clearer, so that it aligns with the documentation for `_PyLong_NumBits` returning `-1` on overflow?

```suggestion
                    let n_bytes = if n_bits == (-1isize as usize) {
```

davidhewitt (commented):
Thanks! Just one quick thought...

davidhewitt:
(Also please add a CHANGELOG entry in the "Fixed" section)

davidhewitt:
Thanks! Afraid you might need to rebase; there's a merge conflict...

davidhewitt:
Perfect, thanks again!

PR #2163, opened 2022-02-12
ci: replace makefile with nox

davidhewitt:
Ports the `Makefile` over to `nox`. Seems like a win for portability!


davidhewitt:
For simplicity I will probably replace `xtask` with this `noxfile` too

noxfile.py
```diff
@@ -0,0 +1,88 @@
...
+def clippy(session: nox.Session) -> None:
+    for feature_set in ["full", "abi3 full"]:
+        session.run(
+            "cargo",
+            "clippy",
+            f"--features={feature_set}",
+            "--all-targets",
+            "--workspace",
+            "--",
+            "-Dwarnings",
```
adamreichold, line 57: To improve self-documentation of such scripts, I would prefer the long option forms, i.e. `--deny warnings`.

noxfile.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@nox.session(name="fmt-rust", venv_backend="none")
+def fmt_rust(session: nox.Session):
+    session.run("cargo", "fmt", "--all", "--check", external=True)
+
+
+@nox.session(name="fmt-py")
+def fmt_py(session: nox.Session):
+    session.install("black==21.12b0")
```
adamreichold, line 43: I think `black==22.1.0` is current but I am not sure whether bumping this is wanted or sensible to do in this PR.

davidhewitt: I'm happy to do so. Will push as a separate commit 👍

Issue #2140, opened 2022-02-02
Add a `Python::allow_thread_unsafe` variant without a `Send` bound


Tpt:
The python  `Python::allow_thread` method allows to release the GIL when evaluating Rust code inside of a function exposed to the Python interpreter.
However, it requires the `Send` bound on the closure given for argument to it in order to make sure no object holding the GIL is an argument of this closure.
This creates a strong limitation on some code that is not thread safe (e.g. use `std::rc::Rc`) but "GIL safe" because it does not hold the GIL.
What about creating an unsafe version of the `Python::allow_thread` that would not enforce the `Send` bound and would be annotated as `unsafe` because it would require the caller to ensure the GIL is not leaked into the closure?

I (@Tpt) am willing to write an implementation PR if you think this addition is relevant to pyo3.

davidhewitt:
How about having an opt-out mechanism for individual variables?

We could have `unsafe fn assert_gil_safe<T>(t: T) -> GilSafe<T>` which is just a newtype struct that implements `Send` and also `Deref<Target = T>`.

My reason for suggesting this alternative is that it's a very far-reaching `unsafe` to remove the bound completely - accidentally accessing a Python API with the GIL released can range from invisible raciness through to loud crashes.

Tpt:
@davidhewitt  Thank you for your feedback!

> We could have `unsafe fn assert_gil_safe<T>(t: T) -> GilSafe<T>` which is just a newtype struct that implements Send and also `Deref<Target = T>`.

This sounds a bit scary to me. What if someone calls `assert_gil_safe` and then send the resulting `GilSafe<T>` to an other thread without much care? It seems to me it saves us from GIL concurrency bugs at the cost of thread safety bugs. What do you think about it?

davidhewitt:
That's a fair point, that's not ideal either. Scrap my suggestion.

It's possible to implement `allow_thread_unsafe` using raw FFI. Perhaps we can start with an example for how users can do this themselves, rather than providing the API?

Tpt:
> It's possible to implement allow_thread_unsafe using raw FFI. Perhaps we can start with an example for how users can do this themselves, rather than providing the API?

Sure! If you want I can write something and do a pull request. Should I add an example inside of the [examples](https://github.com/PyO3/pyo3/tree/main/examples) directory, add something the [the FAQ](https://github.com/PyO3/pyo3/blob/main/guide/src/faq.md) or do something else?

davidhewitt:
[...]

PR #2165, opened 2022-02-12
Implement Auto trait

mejrs:
Per https://github.com/PyO3/pyo3/issues/2140#issuecomment-1029440783

- [x] Make a job for `nightly` feature in the CI - I'm not sure where this fits best.


src/marker.rs
```diff
@@ -2,103 +2,161 @@
...
-    }
-}
+/// # Safety
+///
+/// The type must not carry borrowed Python references or, if it does, not allow access to them if
+/// the GIL is not held.
+///
+/// See the [module-level documentation](self) for more information.
+#[cfg(not(feature = "nightly"))]
+pub unsafe trait Ungil {}
```
mejrs, line 139: Bikesheds on the name are welcome.

src/version.rs
```diff
@@ -0,0 +1,138 @@
...
+        assert!(PythonVersionInfo::from_str("3.5+") == (3, 5, 0));
+        assert!(PythonVersionInfo::from_str("3.5+") != (3, 5, 1));
+        assert!(PythonVersionInfo::from_str("3.5.2a1+") < (3, 5, 3));
+        assert!(PythonVersionInfo::from_str("3.5.2a1+") == (3, 5, 2));
+        assert!(PythonVersionInfo::from_str("3.5.2a1+") == (3, 5));
+        assert!(PythonVersionInfo::from_str("3.5+") == (3, 5));
+        assert!(PythonVersionInfo::from_str("3.5.2a1+") < (3, 6));
+        assert!(PythonVersionInfo::from_str("3.5.2a1+") > (3, 4));
+    }
+}
```
mejrs, line 138: I thought it made sense to move this into its separate file.

Also, does this need to be attached to `Python`? `Py_GetVersion` can be called before the interpreter is even initialized, and returns a pointer to a static.

src/marker.rs
```diff
@@ -2,103 +2,161 @@
...
+//! pub unsafe trait Ungil {}
+//!
+//! unsafe impl<T: Send> Ungil for T {}
+//! ```
+//!
+//! In practice this API works quite well, but it comes with some drawbacks:
+//!
+//! ## Drawbacks
+//!
+//! There is no reason to prevent `!Send` types like [`Rc`] from crossing the closure. After all,
```
davidhewitt, line 35: It might be nice to explain before here that the reason we have the `Send` bound is because auto traits aren't stable, and we want to implement `Ungil` for as many things as possible to make `allow_threads` nice to use. The Python Gil-bound types are all `!Send`, which is why we piggy-back off the auto trait.

src/version.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub major: u8,
+    /// Python minor version (e.g. `11`).
+    pub minor: u8,
+    /// Python patch version (e.g. `0`).
+    pub patch: u8,
+    /// Python version suffix, if applicable (e.g. `a0`).
+    pub suffix: Option<&'py str>,
+}
+
+impl<'py> TryFrom<&'py str> for PythonVersionInfo<'py> {
```
davidhewitt, line 32: Note that implementing this trait rather than private `from_str` function has made this public. I think we should consider implementing the `FromStr` trait rather than `TryFrom`?

tests/test_compile_error.rs
```diff
@@ -71,10 +73,21 @@ fn _test_compile_errors() {
...
     fn tests_rust_1_58(_t: &trybuild::TestCases) {}
 }
+
+#[cfg(feature = "nightly")]
+fn _test_compile_errors() {
+    let t = trybuild::TestCases::new();
+
+    t.compile_fail("tests/ui/not_send_auto_trait.rs");
+    t.compile_fail("tests/ui/not_send_auto_trait2.rs");
+    t.compile_fail("tests/ui/send_wrapper.rs");
```
davidhewitt, line 92: Should we maybe put them in a `ui/nightly` subdirectory to differentiate these?

tests/ui/not_send_auto_trait2.stderr
```diff
@@ -0,0 +1,22 @@
...
+    |  ____________^^^^^^^^^^^^^_-
+    | |            |
+    | |            within `[closure@$DIR/tests/ui/not_send_auto_trait2.rs:8:26: 10:10]`, the trait `Ungil` is not implemented for `PyObject`
+9   | |             println!("{:?}", string);
+10  | |         });
+    | |_________- within this `[closure@$DIR/tests/ui/not_send_auto_trait2.rs:8:26: 10:10]`
+    |
+    = note: required because it appears within the type `UnsafeCell<PyObject>`
+    = note: required because it appears within the type `PyAny`
+    = note: required because it appears within the type `PyString`
```
davidhewitt, line 14: It may be nice to add `!Ungil` implementation to one of the macros in `src/types/mod.rs` so that the error stops here at `PyString`.

src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg_attr(feature = "nightly", feature(specialization))]
```
davidhewitt, line 1: With the specialization feature removed, isn't there a whole load of stuff in `src/conversion.rs` which breaks?

davidhewitt (commented):
Thank you, this looks great!

> Make a job for nightly feature in the CI

It might be nice to add a set of these in the normal `build` matrix? I don't think needs to run for all Python versions, maybe just Python 3.10, once for each OS?

pyo3-ffi/src/lib.rs
```diff
@@ -261,7 +261,7 @@
 macro_rules! opaque_struct {
     ($name:ident) => {
         #[repr(C)]
-        pub struct $name([u8; 0]);
+        pub struct $name([u8; 0], ::std::marker::PhantomData<$crate::PyObject>);
```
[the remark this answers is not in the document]
davidhewitt: EDIT ah I see this is necessary for stable, to make the opaque structs `!Send`?

src/version.rs
mejrs: Unfortunately we can't do that...
 > FromStr does not have a lifetime parameter, and so you can only parse types that do not contain a lifetime parameter themselves.

src/lib.rs
mejrs: I thought I deleted them...I must've missed some things.

src/marker.rs
mejrs: I just went for the shortest one I could think of - all the auto traits in std have really short names, so I figured it would fit. 😆

> cleanest in terms of description

IMO, the danger of names like these is that users might think they understand what it does but are wrong. So perhaps a more cryptic name is better.

pyo3-ffi/src/lib.rs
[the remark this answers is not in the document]
mejrs: All it does is say "this opaque struct (can) contains (pointers to) Python object(s)". It's not really needed, but in the abstract it seems like the right thing to do.

> It may be better to add separately beneath each opaque_struct usage.

`pyo3-ffi` has no idea of what `Ungil` is, because it can't depend on `pyo3`. Manual `!Ungil` impls would have to happen in `pyo3` itself, which would be error prone.

> EDIT ah I see this is necessary for stable, to make the opaque structs !Send?

That was actually not my intent -  they only exist as pointees and raw pointers are already `!Send`.

Without this phantomdata and on nightly, `Ungil` is automatically derived for things like `PyLongObject`. It means that you could create a `Py` like wrapper around `PyLongObject` and automatically have it marked `Ungil`.

That's why `impl !Ungil for ffi::PyObject` exists - without that, `Py<PyAny>` is automatically marked `Ungil`, which just seems wrong to me.

tests/test_compile_error.rs
mejrs: There aren't that many of them so I didn't feel the need to do that.

src/version.rs
davidhewitt: > Also, does this need to be attached to Python? `Py_GetVersion` can be called before the interpreter is even initialized, and returns a pointer to a static.

I wasn't entirely confident on the guarantees, so opted to be safe. If you're confident we can change this to use statics, seems good to me. Maybe as a separate PR to make it easier for users to track patches?

pyo3-ffi/src/lib.rs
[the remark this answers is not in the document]
davidhewitt: Right, I see. I think that the ffi types should not be concerned with the `!Ungil` construct. I think that the implementations we need to implement are:
- `impl !Ungil for pyo3::PyAny`
- `impl !Ungil for pyo3::PyCell`
- `impl !Ungil for pyo3::PyRef`

`Py<PyAny>` should be `Ungil` though, as it's not tied to the GIL lifetime? `&PyAny` should be `!Ungil`.

src/version.rs
davidhewitt: Right, so if we change to `&'static` then this works. That's probably why I didn't implement `FromStr` originally.

tests/ui/not_send_auto_trait2.stderr
davidhewitt: Fair point. The implementation of `!Ungil for PyAny` as suggested by my other comment would be enough anyway.

pyo3-ffi/src/lib.rs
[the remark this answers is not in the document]
mejrs: > I think that the ffi types should not be concerned with the !Ungil construct

This feels like "should raw pointers be `!Send` and `!Sync`?".

I think marking (pointers to) ffi types as `!Ungil` is the correct and conservative thing to do here. It leads to all those types you mentioned being `!Ungil` and also requires manual impls for things like `PyErr` and `Py<T>`. I definitely think this is the correct behaviour.


To clarify, these are the current manual implementations:

afbeelding

src/version.rs
mejrs: > Maybe as a separate PR to make it easier for users to track patches?

Yeah I'd rather do this separately.

src/marker.rs
davidhewitt: That's a fair point. I'm happy to keep the name `Ungil` 👍

src/version.rs
davidhewitt: As we're not _entirely_ sure what the correct implementation for this type should be w.r.t. static etc, would you be ok if we rolled back this trait implementation for the moment to a private function? We can always commit again to the trait later.

pyo3-ffi/src/lib.rs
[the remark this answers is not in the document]
davidhewitt: Ok, I take your point that the ffi types should implement `!Ungil`.

I still don't much like achieving that by adding it to this opaque struct, for a few reasons:
 - Many FFI types don't use `opaque_struct!`, especially on the unlimited API.
 - Similarly, I'm not sure that all of the FFI types which use `opaque_struct!` need to be `!Ungil`.
 - If `extern type` ever stabilises (which `opaque_struct!` is currently a placeholder for), then we'll lose all the `!Ungil` implementations if we were to swap over.

I acknowledge it's a bit painful, however I do really think we'll get a better result if we have a file with manual implementations of `!Ungil` for all the relevant FFI types.

In addition, I still think it would be a nice UX improvement to add `impl !Ungil` for `pyo3::types::PyAny`.

davidhewitt (commented):
Thanks, this is getting really close now. Nice work on the CI. I can go into the repository settings and fix the "required jobs" list just before merging.

As well as the comments below, definitely needs a CHANGELOG!

davidhewitt (approved):
Looks great to me, thanks! Do you think we should consider squashing this down into fewer commits?

davidhewitt:
(I could just squash on merge however sometimes it's nice to not just have one mega commit either 🤷)

mejrs:
Yeah the commits are pretty messy 😭 

Let me see if I can git-fu this...

davidhewitt:
Thanks, nice work! Because this PR has updated the job names I had to tweak the branch protection rules. Anything else will have to rebase on this on in order to me mergeable.

Issue #2129, opened 2022-01-27
Should ml_meth be an union?


mejrs:
In https://github.com/PyO3/pyo3/pull/2126#discussion_r793126402 I mentioned that transmuting function pointers makes me nervous.

The particular code example is this:

```rust
let wrapped_sum_as_string = PyMethodDef {
    ml_name: "sum_as_string\0".as_ptr() as *const c_char,
    ml_meth: Some(transmute::<_PyCFunctionFast, PyCFunction>(sum_as_string)),
    ml_flags: METH_FASTCALL,
    ml_doc: "returns the sum of two integers as a string\0".as_ptr() as *const c_char,
};
```

Representing the `ml_meth` field with an union should make this less tricky and error-prone:
```rust
#[repr(C)]
#[derive(Copy, Clone)]
pub union Foo {
    PyCFunction: PyCFunction,
    PyCFunctionWithKeywords: PyCFunctionWithKeywords,
    #[cfg(not(Py_LIMITED_API))]
    _PyCFunctionFast: _PyCFunctionFast,
    #[cfg(not(Py_LIMITED_API))]
    _PyCFunctionFastWithKeywords: _PyCFunctionFastWithKeywords,
    None: *const libc::c_void,
}
```

It then becomes:

```rust
let wrapped_sum_as_string = PyMethodDef {
    ml_name: "sum_as_string\0".as_ptr() as *const c_char,
    ml_meth: Foo {
        _PyCFunctionFast: sum_as_string,
    },
    ml_flags: METH_FASTCALL,
    ml_doc: "returns the sum of two integers as a string\0".as_ptr() as *const c_char,
};
```

What do you think? Are there any caveats to this?

birkenfeld:
The enum shouldn't be a problem, as long as equal repr is guaranteed, but I the transmute just mirrors what is done on the CPython side, and I don't see how it could be problematic...

davidhewitt:
The union looks nicer for avoiding the transmute, sure, yet I'm not sure it really wins us much. CPython essentially does the C equivalent of a transmute with the function pointer in `ml_meth` anyway depending on the value of `ml_flags`, so it's not like it's any more safe in reality.

Personally I prefer that the current definition exactly matches the C. If the majority want the union I can be won round. Note that changing it would be a breaking change, so I'd also argue if there's not much clear benefit it's not worth the ecosystem churn. (Not sure how many users are using `ffi` directly, though there are clearly some!)

mejrs:
[...]

PR #2166, opened 2022-02-13
Implement ml_meth as an union.

mejrs:
Implements https://github.com/PyO3/pyo3/issues/2129

### DIFF
--- a/pyo3-ffi/src/methodobject.rs
+++ b/pyo3-ffi/src/methodobject.rs
@@ -31,7 +31,7 @@ pub unsafe fn PyCFunction_Check(op: *mut PyObject) -> c_int {
 pub type PyCFunction =
     unsafe extern "C" fn(slf: *mut PyObject, args: *mut PyObject) -> *mut PyObject;
 
-#[cfg(not(Py_LIMITED_API))]
+#[cfg(any(Py_3_10, not(Py_LIMITED_API)))]
 pub type _PyCFunctionFast = unsafe extern "C" fn(
     slf: *mut PyObject,
     args: *mut *mut PyObject,
@@ -52,7 +52,14 @@ pub type _PyCFunctionFastWithKeywords = unsafe extern "C" fn(
     kwnames: *mut PyObject,
 ) -> *mut PyObject;
 
-// skipped PyCMethod (since 3.9)
+#[cfg(all(Py_3_9, not(Py_LIMITED_API)))]
+pub type PyCMethod = unsafe extern "C" fn(
+    slf: *mut PyObject,
+    defining_class: *mut PyTypeObject,
+    args: *const *mut PyObject,
+    nargs: crate::pyport::Py_ssize_t,
+    kwnames: *mut PyObject,
+) -> *mut PyObject;
 
 extern "C" {
     #[cfg_attr(PyPy, link_name = "PyPyCFunction_GetFunction")]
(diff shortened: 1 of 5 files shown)


pyo3-ffi/src/methodobject.rs
```diff
@@ -71,11 +78,44 @@ extern "C" {
...
+/// otherwise the behavior is undefined.
+///
+/// See the [Python C API documentation][1] for more information.
+///
+/// [1]: https://docs.python.org/3/c-api/structures.html#implementing-functions-and-methods
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub union MlMeth {
+    /// This variant corresponds with [`METH_VARARGS`] *or* [`METH_NOARGS`] *or* [`METH_O`].
+    pub PyCFunction: Option<PyCFunction>,
```
mejrs, line 98: Can these things ever be null? Are the options necessary here?

birkenfeld: I don't think they can, a method slot always needs an implementation.

pyo3-ffi/src/methodobject.rs
```diff
@@ -71,11 +78,44 @@ extern "C" {
...
+///
+/// This union must be accompanied by the correct [ml_flags](PyMethodDef::ml_flags),
+/// otherwise the behavior is undefined.
+///
+/// See the [Python C API documentation][1] for more information.
+///
+/// [1]: https://docs.python.org/3/c-api/structures.html#implementing-functions-and-methods
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub union MlMeth {
```
birkenfeld, line 96: I don't particularly like the name. Can we call it `PyMethodDefPointer` (similar to a union we already have without equivalent in CPython, `PyUnicodeObjectData`)?

pyo3-ffi/src/methodobject.rs
davidhewitt: I think so too. Note that strictly speaking C function pointers are nullable whereas Rust ones aren't, and `pyo3-ffi` handles this slightly weirdly by making the typedefs non-nullable and then wrapping them in `Option` in every struct /  function where they're used.

Here I think it's probably ok to drop the `Option`.

davidhewitt:
Thanks, this does look a bit nicer than the transmutes. I think this needs a `CHANGELOG` entry!

pyo3-ffi/src/methodobject.rs
```diff
@@ -71,17 +78,45 @@ extern "C" {
...
-    pub ml_meth: Option<PyCFunction>,
+    pub ml_meth: PyMethodDefPointer,
     pub ml_flags: c_int,
     pub ml_doc: *const c_char,
 }
 
-impl Default for PyMethodDef {
-    fn default() -> PyMethodDef {
-        unsafe { mem::zeroed() }
-    }
```
mejrs, line 82: I don't think this default impl makes sense if PyMethodDefPointer is never null.

davidhewitt: Agreed; I also actually generally don't like having any implementations of traits or methods for `pyo3-ffi` types if possible (except for `Copy` & `Clone`). Makes it closer to the C that they come from.

davidhewitt (approved):
LGTM, thanks!

PR #2167, opened 2022-02-14
Fix example to prevent this issue: https://github.com/PyO3/maturin/issues/256

serg06:
The example currently triggers [this issue](https://github.com/PyO3/maturin/issues/256).

This change fixes it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ fn sum_as_string(a: usize, b: usize) -> PyResult<String> {
 /// the `lib.name` setting in the `Cargo.toml`, else Python will not be able to
 /// import the module.
 #[pymodule]
+#[pyo3(name = "string_sum")]
 fn string_sum(_py: Python, m: &PyModule) -> PyResult<()> {
     m.add_function(wrap_pyfunction!(sum_as_string, m)?)?;
     Ok(())


birkenfeld:
This is unexpected. Not specifying a name should be the same as the function name being used.

If it's not, I'd rather fix it to become equivalent, than add this workaround.

davidhewitt:
Agreed; @serg06 can you share with us some code which demonstrates the problem you had?

serg06:
I found the issue: It happens when my library name and function name are different. For example:

```toml
# Cargo.toml
[lib]
name = "rust_python_lib"
```

```rs
// lib.rs
#[pymodule]
fn string_sum(_py: Python, m: &PyModule) -> PyResult<()> {
```

Now I run:

```bash
# Recreate env just in case
conda activate base && conda remove --name test --all -y && conda create --name test python=3.7 -y && conda activate test && pip install maturin

maturin develop
python
>>> import rust_python_lib
  File "C:\Users\NAME\anaconda3\envs\test2\lib\site-packages\rust_python_lib\__init__.py", line 1, in <module>
    from .rust_python_lib import *
ImportError: dynamic module does not define module export function (PyInit_rust_python_lib)
>>> import string_sum
ModuleNotFoundError: No module named 'string_sum'

davidhewitt:
Right, I see, thanks for clarifying. So the example doesn't need this setting; instead users who rename the library should also rename the `#[pymodule]`. Instead I will add a comment to the `Cargo.toml` as in #2170. I'll close this one.

serg06:
@davidhewitt Perfect thanks.

PR #2172, opened 2022-02-16
Use addr_of_mut in pyo3-ffi

mejrs:
All these comparisons and getters were going through &mut references to mutable statics...

### DIFF
--- a/pyo3-ffi/build.rs
+++ b/pyo3-ffi/build.rs
@@ -100,7 +100,27 @@ fn configure_pyo3() -> Result<()> {
         println!("{}", line);
     }
 
+    let rustc_minor_version = rustc_minor_version().unwrap_or(0);
+
+    // Enable use of std::ptr::addr_of! on Rust 1.51 and greater
+    if rustc_minor_version >= 51 {
+        println!("cargo:rustc-cfg=addr_of");
+    }
+
     Ok(())
+}
+
+fn rustc_minor_version() -> Option<u32> {
+    use std::{env, process::Command};
+
+    let rustc = env::var_os("RUSTC")?;
+    let output = Command::new(rustc).arg("--version").output().ok()?;
+    let version = core::str::from_utf8(&output.stdout).ok()?;
+    let mut pieces = version.split('.');
+    if pieces.next() != Some("rustc 1") {
+        return None;
+    }
+    pieces.next()?.parse().ok()
 }
 
 fn print_config_and_exit(config: &InterpreterConfig) {
--- a/pyo3-ffi/build.rs
+++ b/pyo3-ffi/build.rs
@@ -1,5 +1,5 @@
 use pyo3_build_config::{
-    bail, ensure,
+    bail, ensure, print_feature_cfgs,
     pyo3_build_script_impl::{
         cargo_env_var, env_var, errors::Result, resolve_interpreter_config, InterpreterConfig,
         PythonVersion,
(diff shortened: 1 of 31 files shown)


pyo3-ffi/build.rs
```diff
@@ -100,9 +100,29 @@ fn configure_pyo3() -> Result<()> {
...
+
+    // Enable use of std::ptr::addr_of! on Rust 1.51 and greater
+    if rustc_minor_version >= 51 {
+        println!("cargo:rustc-cfg=addr_of");
+    }
+
     Ok(())
 }
 
+fn rustc_minor_version() -> Option<u32> {
```
adamreichold, line 113: PyO3 itself already uses the `rustversion`, wouldn't it make sense to apply it here as well?

pyo3-ffi/src/lib.rs
```diff
@@ -265,6 +265,19 @@ macro_rules! opaque_struct {
     };
 }
 
+macro_rules! addr_of_mut_shim {
```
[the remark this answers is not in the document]
mejrs: Short answer: It's a shim, so let's call it a shim 🤔

Long answer: It can work with the same name, but I'd rather not pretend it's the actual `addr_of_mut`. If it was named that the casual reader wouldn't be able to tell that it's not `split_at_mut`.

pyo3-ffi/build.rs
mejrs: The `rustversion` crate? That is only a dev dependency.

pyo3-ffi/src/lib.rs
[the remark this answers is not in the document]
adamreichold: > If it was named that the casual reader wouldn't be able to tell that it's not addr_of_mut.

That was my actual intent: Not having to think about the shim part while reading the code making use of it. (Also under the assumption that the shim will be removed once the MSRV passes 1.51.)

But fair enough calling it a shim if that is considered sufficiently important.

pyo3-ffi/build.rs
adamreichold: Indeed, wouldn't also just be build-dependency if used in the build script here?

I understand that it is meant to be used directly for conditional compilation, but that could also be applied to the `println!` invocation in the build script to avoid manually parsing the version.

Or at least could we somehow unify the parsing code between the build scripts of `pyo3` and `pyo3-ffi`?

pyo3-ffi/src/lib.rs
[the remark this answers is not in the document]
mejrs: > Also under the assumption that the shim will be removed once the MSRV passes 1.51.

Absolutely, there's quite of bit of addr_of/min_const_generics stuff that can be deleted with a 1.51 msrv 🎉

pyo3-ffi/build.rs
mejrs: Last time I discussed it @davidhewitt (I think?) didn't want to add dependencies, so I didn't do it here. I would prefer not to reinvent the wheel but I also don't want to add more macro crates as dependencies.

mejrs: I feel kind of split between deduplicating the code - I don't like duplication either but it's not a whole lot and it doesn't do anything clever, so i don't think the cost of duplication is great here. And we'll just delete it all next time we bump msrv.

davidhewitt: Yes, I keep going back and forth whether we should add `rustversion` and refactor our build scripts. On the one hand, it might be nice to use the dependency to keep our code a bit simpler. On the other, it keeps our dependency tree leaner to just do it ourselves. I slightly prefer the current implementation, though only slightly so if you folks wanted to change to `rustversion` that's fine by me.

pyo3-ffi/src/lib.rs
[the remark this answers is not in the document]
davidhewitt: I also wondered about just calling the macro `addr_of_mut`; we can always be explicit and use `crate::addr_of_mut` everywhere to make it obvious that it's different from the stdlib if we wanted. Calling it `addr_of_mut_shim` is also fine.

I'm already looking forward to the next MSRV / Python bump; if we wait until mid 2023 when 3.7 reaches end of life we might get quite a few upgrades :)

davidhewitt (approved):
Thanks, a nice tidy up!

Happy for you folks to decide on `rustversion`and `addr_of_mut` / `addr_of_mut_shim` and merge at your leisure.

pyo3-ffi/build.rs
davidhewitt: To avoid the duplication could always shove it into `pyo3-build-config` as a `#[doc(hidden)]` API :P

mejrs: The current implementation is pretty simple. I'll just keep that.

> To avoid the duplication could always shove it into pyo3-build-config as a #[doc(hidden)] API :P

Yeah, I'll do just that.

adamreichold (approved):

PR #2173, opened 2022-02-17
pyproto: deprecate protocol traits

davidhewitt:
The final step before I think we're ready to release 0.16!

### DIFF
--- a/Architecture.md
+++ b/Architecture.md
@@ -144,20 +144,12 @@ For example, you can see `type({})` shows `dict` and `type(type({}))` shows `typ
 
 ## 4. Protocol methods
 
-Python has some built-in special methods called dunder, such as `__iter__`.
-They are called [abstract objects layer](https://docs.python.org/3/c-api/abstract.html) in
+Python has some built-in special methods called dunder methods, such as `__iter__`.
+They are called "slots" in the [abstract objects layer](https://docs.python.org/3/c-api/abstract.html) in
 Python/C API.
-We provide a way to implement those protocols by using `#[pyproto]` and specific traits, such
-as `PyIterProtocol`.
-[`src/class`] defines these traits.
-Each protocol method has a corresponding FFI function.
-For example, `PyIterProtocol::__iter__` has
-`pub unsafe extern "C" fn iter<T>(slf: *mut PyObject) -> *mut PyObject`.
-When `#[pyproto]` finds that `T` implements `PyIterProtocol::__iter__`, it automatically
-sets `iter<T>` on the type object of `T`.
-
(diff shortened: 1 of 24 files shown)


davidhewitt:
@birkenfeld to answer your question in https://github.com/PyO3/pyo3/pull/2159#issuecomment-1042608176 - I started this branch last night. If you want to work on it please do take it over and push further commits! I'll get working on the 0.16 release branch instead.

I guess the remaining work here is to go through documentation, generally making `#[pyproto]` less obvious, and wherever it does exist point users to `#[pymethods]` instead?

birkenfeld:
Ok, I'll get on it. BTW, it is sad to see all the `#![allow(deprecated)]`; I'd have thought that deprecations in the same crate would be allowed, as this will be easy to overlook needing removal later.

davidhewitt:
Agreed, though I think most of these files can simply be deleted when we finally remove `#[pyproto]`

birkenfeld:
So, questions that came up while working on this:

1. Is `__getattr__` actually equivalent to Python `__getattr__` or rather `__getattribute__`?
2. In pyproto's `__getitem__` etc., negative indices were handled specially. I assume I was correct that this is not done for pymethods?
3. There doesn't seem to be support for `__inplace_concat__` and `__inplace_repeat__` sequence methods?

davidhewitt:
1. Yes, the two methods share the one slot, however we have [this code](https://github.com/PyO3/pyo3/blob/adc6400b83fe6071d6d60103806ec8c1aaf8a1d4/pyo3-macros-backend/src/pymethod.rs#L570-578) which makes the slot behave like `__getattr__`. I think in principle we could support both methods like CPython.

2. Yes, the summary of the issue is that if the `Py_sq_len` slot is implemented then CPython will add that length to a negative index before accessing the sequence slots. The mapping slots are always tried first, so in practice this means that only in weird edge cases (from C) would CPython do this adjustment to negative indices. I felt this was hard for users to understand and impossible to correct for, so I chose for our `__len__` to not implement the `Py_sq_len` slot and thus minimize potential confusion.

3. Ooh, good spot. I think I missed because these magic methods don't exist in pure Python; they're just lumped in with `__iadd__`. We could copy the existing design straight from `#[pyproto]`, seems clear enough?

davidhewitt:
Pushed a commit to add `__inplace_concat__` and `__inplace_repeat__` support.

davidhewitt:
I'll merge this tonight if nobody has any concerns!

birkenfeld:
Nice, __getattribute__ can be another PR.

PR #2175, opened 2022-02-19
Added `eq`, `ne`, `gt` etc. methods.

### LINKED ISSUE #2155: `is_eq`, `is_gt`, etc helper methods
Just a thought: why don't we add `is_eq()`, `is_gt()` etc methods (alongside the `compare` and `rich_compare`), for the sake of newcomer-friendliness? Sort of the same as `is_none()` etc.

Also, note explicitly in trait impl docstrings (and in `is_eq`/`is_ne` docstrings) that `PartialEq` compares *pointers*, which may otherwise lead to confusion.

(Here's a use case example. I've used pyo3 and rust-numpy for quite a lot and I've read large chunks of its sources but never needed to compare Python objects before for some reason. When I did, I started searching docs.rs and source code and gh repo for "eq" but was unable to find anything. It took me a good while till I randomly stumbled on `rich_compare`. Maybe I'm still missing something?)

Tom1380:
This is done for convenience and to be friendlier to beginners.

Closes #2155.

### DIFF
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -262,61 +262,61 @@ impl PyAny {
     /// Tests whether this object is less than another.
     ///
     /// This is equivalent to the Python expression `self < other`.
-    pub fn lt<O>(&self, other: O) -> PyResult<&PyAny>
+    pub fn lt<O>(&self, other: O) -> PyResult<bool>
     where
         O: ToPyObject,
     {
-        self.rich_compare(other, CompareOp::Lt)
+        self.rich_compare(other, CompareOp::Lt)?.is_true()
     }
 
     /// Tests whether this object is less than or equal to another.
     ///
     /// This is equivalent to the Python expression `self <= other`.
-    pub fn le<O>(&self, other: O) -> PyResult<&PyAny>
+    pub fn le<O>(&self, other: O) -> PyResult<bool>
     where
         O: ToPyObject,
     {
-        self.rich_compare(other, CompareOp::Le)
+        self.rich_compare(other, CompareOp::Le)?.is_true()
     }
 
     /// Tests whether this object is equal to another.
     ///
     /// This is equivalent to the Python expression `self == other`.
-    pub fn eq<O>(&self, other: O) -> PyResult<&PyAny>
+    pub fn eq<O>(&self, other: O) -> PyResult<bool>
     where
         O: ToPyObject,
     {
-        self.rich_compare(other, CompareOp::Eq)
+        self.rich_compare(other, CompareOp::Eq)?.is_true()
     }
 
     /// Tests whether this object is not equal to another.
     ///
     /// This is equivalent to the Python expression `self != other`.
-    pub fn ne<O>(&self, other: O) -> PyResult<&PyAny>
+    pub fn ne<O>(&self, other: O) -> PyResult<bool>
     where
         O: ToPyObject,
     {
-        self.rich_compare(other, CompareOp::Ne)
+        self.rich_compare(other, CompareOp::Ne)?.is_true()
     }
 
     /// Tests whether this object is greater than another.
     ///
     /// This is equivalent to the Python expression `self > other`.
-    pub fn gt<O>(&self, other: O) -> PyResult<&PyAny>
+    pub fn gt<O>(&self, other: O) -> PyResult<bool>
     where
         O: ToPyObject,
     {
-        self.rich_compare(other, CompareOp::Gt)
+        self.rich_compare(other, CompareOp::Gt)?.is_true()
     }
 
     /// Tests whether this object is greater than or equal to another.
     ///
     /// This is equivalent to the Python expression `self >= other`.
-    pub fn ge<O>(&self, other: O) -> PyResult<&PyAny>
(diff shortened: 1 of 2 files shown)


mejrs:
Thanks! 

Can you add tests for these? 😃 



Also, I'm not sure about the names for these. I'm tempted to go with `a.eq(b)` over  `a.is_eq(b)` and so on. @davidhewitt 
 since you were the one that suggested these - what do you think?

birkenfeld:
Also needs an entry in the changelog, please.

Tom1380:
@mejrs sure, I'd be happy to add tests.

I tried to find where `rich_compare` itself is tested, but I couldn't find it.
Can you point me to where I can find it? So I can add tests right below it.

Tom1380:
@mejrs Should I add the tests as documentation tests?

mejrs:
> @mejrs Should I add the tests as documentation tests?

I would recommend adding a new test at the bottom of the file where you create a vec with a bunch of different python objects and loop through every possible pair against every compare op, and check whether they give the expected results. That is probably a bit much for a doc example 😉

davidhewitt:
Thanks! Regular tests would be better than doctests please; we don't measure coverage for doctests yet.

> Also, I'm not sure about the names for these. I'm tempted to go with `a.eq(b)` over `a.is_eq(b)` and so on. @davidhewitt
since you were the one that suggested these - what do you think?

The only thing that makes me uneasy about that is that then `a.eq(b)` clashes with the `PartialEq` implementation, which uses pointer equality rather than Python equality. I think having the notably different names is a good thing? Or maybe we should consider changing the `PartialEq` implementation. (Though what to do if `__eq__` raises an Exception? Panic?)

mejrs:
Oh, that is unfortunate. But having `is_eq` and `eq` methods available is also uneasy for me.

In hindsight it may have been better to not have the PartialEq impls and have a `ptr_eq` method instead.

birkenfeld:
Yeah I agree that `obj1 == obj2` just comparing pointers is inviting confusion. I'd rather have that exposed as `a.is(b)` as in Python (`ptr_eq` is also ok but hard to find when coming from the Python side). Then the way would be clear for `eq` etc. methods using the richcmp functions, and `PartialEq` is removed.

The question is, can this have further repercussions than an easy-to-fix compile error? If we're not currently also providing `Eq`, at least it wasn't possible to put objects into HashMaps.

davidhewitt:
I would be happy to remove the current `PartialEq` implementation. I'm pretty convinced that pointer equality for that is surprising.

mejrs:
I've thought about it some more and I really don't like the PartialEq either.

So, I think the plan is?
- Remove PartialEq
- Add PyAny::is(other)
- Rename all the methods this PR added from `is_<op>` to just `<op>`

davidhewitt:
We may want to mention the removal in the migration guide, just in case it causes any surprises for users.

birkenfeld:
I did the PartialEq related stuff in #2183, so this PR just needs some tests and changelog entry to be ready.

Tom1380:
Please let me know if you think the tests are done right, I'm not sure that Rust's `cmp` logic is the same as Python's.

src/types/any.rs
```diff
@@ -834,4 +894,83 @@ class SimpleClass:
...
+        Python::with_gil(|py| {
+            for a in list {
+                for b in list {
+                    let a_py = a.to_object(py).into_ref(py);
+                    let b_py = b.to_object(py).into_ref(py);
+                    let unwrap_cmp = |cmp: PyResult<&PyAny>| cmp.unwrap().is_true().unwrap();
+                    assert_eq!(
+                        a.lt(b),
+                        unwrap_cmp(a_py.lt(b_py)),
+                        "{a_py} should be less than {b_py}"
```
birkenfeld, line 912: These inline captures will not work on MSRV (1.48) I think.

birkenfeld:
I wonder if we shouldn't go one step further here and do the `is_true` call in the functions.

I know that richcmp methods can return anything, but that is already quite esoteric, and usability here will be much better if only one `?` or `unwrap` is needed.

Those who need the general result can still use `rich_compare` after all.

davidhewitt:
> I wonder if we shouldn't go one step further here and do the is_true call in the functions.

I would support that.

Tom1380:
> I wonder if we shouldn't go one step further here and do the `is_true` call in the functions.
Done.

Tom1380:
I think it's ready. Please review it.

birkenfeld:
I think this is good to go. Don't know how the coverage could go down, all methods are tested.

birkenfeld (approved):

davidhewitt:
Codecov seems to do this weird thing where any PRs with the originating fork branch as `main` it takes that coverage as our current CI coverage. I noticed this the other day but haven't really had time to investigate or figure out why...

birkenfeld:
I also saw that, since rustfmt split the `assert` as much as possible, it complained about the arguments for the assert error format not being covered. While true, that is unavoidable for working tests :)

davidhewitt:
True, yes! Perhaps the upstream macro expansion should consider having the `#[no_coverage]` attribute added on the assertion failure branch?!

Issue #2177, opened 2022-02-21
Pass detailed error message to python when parsing an argument fails


ricohageman:
This issue might be seen as a continuation of https://github.com/PyO3/pyo3/issues/1640, which introduced more elaborated error messages when parsing objects using the derived FromPyObject trait. However, the extended error messages are not properly passed to python when parsing arguments provided by python to a rust function fails.

Consider for example the following case. When passing a wrong type for the `value` field, the error message contains the fact that `value` cannot be parsed. However, it fails to explain why it failed to parse value. This is valuable information and thus I would like to get this information out.

**Example rust code**
```
#[derive(FromPyObject)]
struct ValueClass {
   value: usize,
}

#[pyfunction]
fn function_to_call_with_value_class(value: ValueClass) {
    // Do stuff with the argument.
}

#[pymodule]
fn module(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(function_to_call_with_value_class, m)?)?;
    Ok(())
}
```

**Example python code**
```
Class ValueClass:
    def __init__(self, value):
        self.value = value

# Should be okay
module.function_to_call_with_value_class(ValueClass(0))  

# Should tell that a string cannot be converted into a integer.
# Currently tells: argument 'struct_arg': failed to extract field ValueClass.value
module.function_to_call_with_value_class(ValueClass("unparsable string")) 

# Should tell that a negative integer cannot be converted into an unsigned integer.
# Currently tells: argument 'struct_arg': failed to extract field ValueClass.value
module.function_to_call_with_value_class(ValueClass(-5)) 
```

RicoHageman:
This is closed by #2178

PR #2178, opened 2022-02-21
Pass detailed error message to python when parsing an argument fails

RicoHageman:
This pull requests includes some changes related to issue https://github.com/PyO3/pyo3/issues/2177

### DIFF
--- a/src/impl_/extract_argument.rs
+++ b/src/impl_/extract_argument.rs
@@ -101,21 +101,12 @@ pub fn from_py_with_with_default<'py, T>(
 /// single string.)
 #[doc(hidden)]
 #[cold]
-pub fn argument_extraction_error(py: Python, arg_name: &str, mut error: PyErr) -> PyErr {
+pub fn argument_extraction_error(py: Python, arg_name: &str, error: PyErr) -> PyErr {
     if error.get_type(py) == PyTypeError::type_object(py) {
-        let mut reason = error
-            .value(py)
-            .str()
-            .unwrap_or_else(|_| PyString::new(py, ""))
-            .to_string();
-
-        while let Some(cause) = error.cause(py) {
-            reason.push_str(":\n\t");
-            reason.push_str(&cause.to_string());
-            error = cause
-        }
-
-        PyTypeError::new_err(format!("argument '{}': {}", arg_name, reason))
+        let remapped_error =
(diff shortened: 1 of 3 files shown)


src/impl_/extract_argument.rs
```diff
@@ -101,9 +101,21 @@ pub fn from_py_with_with_default<'py, T>(
...
+            .unwrap_or_else(|_| PyString::new(py, ""))
+            .to_string();
+
+        while let Some(cause) = error.cause(py) {
+            reason.push_str(":\n\t");
+            reason.push_str(&cause.to_string());
+            error = cause
+        }
+
+        PyTypeError::new_err(format!("argument '{}': {}", arg_name, reason))
```
davidhewitt, line 118: I wonder, rather than flattening all the causes, should we be preserving the cause of the original error instead, like this:

```suggestion
        let remapped_error = PyTypeError::new_err(format!("argument '{}': {}", arg_name, error.value(py)));
        remapped_error.set_cause(py, error.cause(py));
        remapped_error
```

Then the error seen from Python might be more like this:

```
TypeError: 'str' object cannot be interpreted as an integer

  File "/Users/david/Dev/pyo3-scratch/test.py", line 19, in <module>
    pyo3_scratch.function_to_call_with_value_class(ValueClass("unparsable string"))
TypeError: argument 'value': failed to extract field ValueClass.value
```

I think preserving the cause tree like this is probably more correct. For example, the argument conversion code could run any old Python code (not just our PyO3 code) which generates some exception with its own separate cause. I think PyO3 flattening those causes into a single error message is a mistake; it might lose valuable debugging information.

davidhewitt (commented):
Thank you for looking into this; sorry I never found a chance to replying further to our conversation on Gitter.

You're right that this is definitely a bug. I wonder though, instead of flattening all the causes into a single exception, is it nicer to just preserve the existing cause, if any? See comment below...

src/impl_/extract_argument.rs
RicoHageman: You're totally right with this remark. Also, I didn't completely liked the flattening so this is a better approach while still preserving all the relevant information. 

In addition, I tried to comply to rustfmt and clippy where relevant to this pull request. However, they were also marking parts that were untouched (some strict comparisons between floating points). Let's see if the workflows succeed.

davidhewitt (approved):
Looks great, many thanks!

PR #2183, opened 2022-02-24
Py/PyAny: remove PartialEq impl and add is()

birkenfeld:
and fix the fallout, mostly in tests.

See https://github.com/PyO3/pyo3/pull/2175#issuecomment-1048162127 for discussion.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -463,6 +463,15 @@ where
 }
 
 impl<T> Py<T> {
+    /// Returns whether `self` and `other` point to the same object. To compare
+    /// the equality of two objects (the `==` operator), use [`eq`](PyAny::eq).
+    ///
+    /// This is equivalent to the Python expression `self is other`.
+    #[inline]
+    pub fn is<U: AsPyPointer>(&self, o: &U) -> bool {
+        self.as_ptr() == o.as_ptr()
+    }
+
     /// Gets the reference count of the `ffi::PyObject` pointer.
     #[inline]
     pub fn get_refcnt(&self, _py: Python) -> isize {
@@ -826,13 +835,6 @@ where
 {
     fn from(pyref: PyRefMut<'a, T>) -> Self {
         unsafe { Py::from_borrowed_ptr(pyref.py(), pyref.as_ptr()) }
-    }
-}
-
-impl<T> PartialEq for Py<T> {
-    #[inline]
-    fn eq(&self, o: &Py<T>) -> bool {
-        self.0 == o.0
     }
 }
 
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -87,6 +87,15 @@ impl PyAny {
         <T as PyTryFrom>::try_from(self)
     }
 
+    /// Returns whether `self` and `other` point to the same object. To compare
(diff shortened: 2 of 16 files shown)


birkenfeld:
The guide entry presumes #2175 as merged.

src/instance.rs
```diff
@@ -463,6 +463,13 @@ where
 }
 
 impl<T> Py<T> {
+    /// Returns whether the object is the same as the other.
+    ///
+    /// This is equivalent to the Python expression `self is other`.
+    pub fn is<U: AsPyPointer>(&self, o: &U) -> bool {
```
adamreichold, line 469: Should this inherit the `#[inline]` attribute from the `PartialEq` method?

davidhewitt: I think probably makes sense for this to be `#[inline]`.

davidhewitt (approved):
LGTM, thanks for updating migration notes too. It's a shame we can't deprecate trait implementations, however I would _hope_ that any breaking changes users have to undergo are also removing possible bugs from their code 😄

davidhewitt:
Clippy fix needs rebasing on #2184

src/instance.rs
birkenfeld: will do

birkenfeld:
I actually found a semi-bug; you can see it in the tests (test_sequence and _pyproto), where vector of objects was compared directly to `vec![1.into(),...]`.

It works since small integers are singletons in CPython, so I kept the logic the same for now, but that's not a guarantee and maybe should be changed to use `eq`.

birkenfeld:
BTW, a simple `cargo test` always results in
```
   Doc-tests pyo3
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> /home/gbr/devel/pyo3/src/lib.rs:2:21
  |
2 | #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
  |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
for me now.

I can run with nightly, but then the UI tests usually fail, which is not pretty.

mejrs:
> BTW, a simple `cargo test` always results in

Yeah I was getting annoyed by this too. It should be fixed with https://github.com/PyO3/pyo3/pull/2176


You can delete the rustdocflags in `.cargo/config`  to get around it for now.

davidhewitt:
Looks great, I'm just going to hold off on merging this for a moment until #2175 is ready.

src/types/any.rs
```diff
@@ -87,6 +87,14 @@ impl PyAny {
         <T as PyTryFrom>::try_from(self)
     }
 
+    /// Returns whether the object is the same as the other.
+    ///
+    /// This is equivalent to the Python expression `self is other`.
```
mejrs, line 92: I think this is worth explaining in a bit more detail, because this can be pretty surprising to people unfamiliar with Python. And even people that should know better get `==` and `is` wrong sometimes.

```suggestion
    /// Returns whether `self` and `other` point to the same object. To compare the equality of two objects
    /// (the `==` operator), use [`eq`](PyAny::eq).
    ///
    /// This is equivalent to the Python expression `self is other`.
```
(the link needs the other PR to work)

PR #2187, opened 2022-02-25
Protocols: implement __getattribute__

### LINKED ISSUE #2186: Support `__getattribute__` in `#[pymethods]`
Yes, [`__getattr__` and `__getattribute__`] share the one slot, however we have [this code](https://github.com/PyO3/pyo3/blob/adc6400b83fe6071d6d60103806ec8c1aaf8a1d4/pyo3-macros-backend/src/pymethod.rs#L570-578) which makes the slot behave like `__getattr__`. I think in principle we could support both methods like CPython.

_Originally posted by @davidhewitt in https://github.com/PyO3/pyo3/issues/2173#issuecomment-1044191311_

birkenfeld:
converting tp_getattro to a shared slot

Fixes #2186

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,11 +33,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `wrap_pyfunction!` can now wrap a `#[pyfunction]` which is implemented in a different Rust module or crate. [#2091](https://github.com/PyO3/pyo3/pull/2091)
 - Add `PyAny::contains` method (`in` operator for `PyAny`). [#2115](https://github.com/PyO3/pyo3/pull/2115)
 - Add `PyMapping::contains` method (`in` operator for `PyMapping`). [#2133](https://github.com/PyO3/pyo3/pull/2133)
-- Add support for the `__getattribute__` magic method. [#2187](https://github.com/PyO3/pyo3/pull/2187)
 - Add garbage collection magic methods `__traverse__` and `__clear__` to `#[pymethods]`. [#2159](https://github.com/PyO3/pyo3/pull/2159)
 - Add support for `from_py_with` on struct tuples and enums to override the default from-Python conversion. [#2181](https://github.com/PyO3/pyo3/pull/2181)
 - Add `eq`, `ne`, `lt`, `le`, `gt`, `ge` methods to `PyAny` that wrap `rich_compare`. [#2175](https://github.com/PyO3/pyo3/pull/2175)
 - Add `Py::is` and `PyAny::is` methods to check for object identity. [#2183](https://github.com/PyO3/pyo3/pull/2183)
+- Add support for the `__getattribute__` magic method. [#2187](https://github.com/PyO3/pyo3/pull/2187)
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


src/impl_/pyclass.rs
```diff
@@ -198,6 +198,80 @@ macro_rules! slot_fragment_trait {
...
+    /// # Safety: _slf and _attr must be valid non-null Python objects
+    #[inline]
+    unsafe fn __getattribute__(
+        self,
+        py: Python,
+        slf: *mut ffi::PyObject,
+        attr: *mut ffi::PyObject,
+    ) -> PyResult<*mut ffi::PyObject> {
+        let res = ffi::PyObject_GenericGetAttr(slf, attr);
+        if res.is_null() { Err(PyErr::fetch(py)) }
```
birkenfeld, line 213: I'm not sure about this implementation, does it look correct?

src/impl_/pyclass.rs
```diff
@@ -198,6 +198,80 @@ macro_rules! slot_fragment_trait {
...
+
+    /// # Safety: _slf and _attr must be valid non-null Python objects
+    #[inline]
+    unsafe fn __getattr__(
+        self,
+        py: Python,
+        _slf: *mut ffi::PyObject,
+        attr: *mut ffi::PyObject,
+        ) -> PyResult<*mut ffi::PyObject> {
+        Err(PyErr::new::<PyAttributeError, _>((Py::<PyAny>::from_owned_ptr(py, attr),)))
```
birkenfeld, line 229: Same here

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `wrap_pyfunction!` can now wrap a `#[pyfunction]` which is implemented in a different Rust module or crate. [#2091](https://github.com/PyO3/pyo3/pull/2091)
 - Add `PyAny::contains` method (`in` operator for `PyAny`). [#2115](https://github.com/PyO3/pyo3/pull/2115)
 - Add `PyMapping::contains` method (`in` operator for `PyMapping`). [#2133](https://github.com/PyO3/pyo3/pull/2133)
+- Add support for the `__getattribute__` magic method. [#2187](https://github.com/PyO3/pyo3/pull/2187)
```
davidhewitt, line 36: I quite like these to be in PR order.

src/impl_/pyclass.rs
davidhewitt: I think this pointer is not owned by the function call, so need to use from_borrowed:

```suggestion
        Err(PyErr::new::<PyAttributeError, _>((Py::<PyAny>::from_borrowed_ptr(py, attr),)))
```

src/impl_/pyclass.rs
davidhewitt: Looks good to me!

src/impl_/pyclass.rs
```diff
@@ -198,6 +198,80 @@ macro_rules! slot_fragment_trait {
...
+                    // Strategy:
+                    // - Try __getattribute__ first.  Its default is PyObject_GenericGetAttr.
+                    // - If it returns a result, use it.
+                    // - If it fails with AttributeError, try __getattr__.
+                    // - If it fails otherwise, reraise.
+                    match collector.__getattribute__(py, _slf, attr) {
+                        Ok(obj) => obj.convert(py),
+                        Err(e) if e.is_instance_of::<$crate::exceptions::PyAttributeError>(py) => {
+                            collector.__getattr__(py, _slf, attr).convert(py)
+                        }
```
davidhewitt, line 260: I don't think the `.convert()` calls needed here because these are `PyResult<*mut ffi::PyObject>` and so already an FFI type.

```suggestion
                        Ok(obj) => obj,
                        Err(e) if e.is_instance_of::<$crate::exceptions::PyAttributeError>(py) => {
                            collector.__getattr__(py, _slf, attr)
                        }
```

davidhewitt (commented):
Awesome, nice one! Just a couple tidy ups needed I think...

birkenfeld:
Thanks for the review, should be good now.

PR #2181, opened 2022-02-23
Add support for `from_py_with` on struct tuples and enums

RicoHageman:
Fixes https://github.com/PyO3/pyo3/issues/1441 by introducing `from_py_with` for tuple structs and enum.

PR #2190, opened 2022-02-26
Make gil tests not racy

mejrs:
Changes the refcount checking to happen from a thread that's actually holding the GIL.

Per https://github.com/PyO3/pyo3/pull/2181#issuecomment-1049741564

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -700,71 +700,120 @@ mod tests {
 
     #[test]
     fn test_clone_without_gil() {
-        let gil = Python::acquire_gil();
-        let py = gil.python();
-        let obj = get_object(py);
-        let count = obj.get_refcnt(py);
+        use crate::{Py, PyAny};
+        use std::sync::atomic::{AtomicBool, Ordering};
+        use std::{sync::Arc, thread};
 
-        // Cloning without GIL should not update reference count
-        drop(gil);
-        let c = obj.clone();
-        assert_eq!(
-            count,
-            obj.get_refcnt(unsafe { Python::assume_gil_acquired() })
-        );
+        // Some spinlocks for synchronizing
+        static GIL_ACQUIRED: AtomicBool = AtomicBool::new(false);
+        static OBJECT_CLONED: AtomicBool = AtomicBool::new(false);
+        static REFCNT_CHECKED: AtomicBool = AtomicBool::new(false);
 
-        // Acquring GIL will clear this pending change
-        let gil = Python::acquire_gil();
-        let py = gil.python();
+        Python::with_gil(|py| {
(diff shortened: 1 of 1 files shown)


mejrs:
> While the extend test case looks basically sound to me

These tests don't use unsafe anymore, what do you mean with "unsound"?

>  I wonder whether these test cases are special enough to be put into a separate integration test

I don't think they are any more or less special than the other tests so I'm not sure. Fwiw, `test_clone_in_other_thread` accesses `POOL` directly, so it can only be written in `gil.rs`.

> (This would also close some more theoretical objects, for example in the current implementation, another test case could acquire the GIL between this test case releasing it and the background thread created by it acquiring thereby masking if one or more acquisitions of the GIL happened between the synchronization points.)

Other threads shouldn't be able to interfere with these tests, I think I got the synchronization right so that the GIL is always held during the critical sections (I hope I didn't mess up).

adamreichold:
Hhhmmm, I think that the `test_clone_in_other_thread` test case is susceptible to interference from other tests, i.e. the assertion
```rust
// The pointer should appear once in the incref pool, and once in the
// decref pool (for the clone being created and also dropped)
assert_eq!(&*POOL.pointer_ops.lock().0, &[ptr]);
assert_eq!(&*POOL.pointer_ops.lock().1, &[ptr]);
```
and
```rust
// Acquiring the gil clears the pool
assert!(POOL.pointer_ops.lock().0.is_empty());
assert!(POOL.pointer_ops.lock().1.is_empty());
```
and probably too strict as other test threads could use `POOL` while the test is running (this probably even less likely than the race which led us here). I guess that could be fixed by checking of presence and absence of `ptr` instead of comparing with a single-element slice and checking for emptiness.

Which brings me back to thinking that isolating these tests within a separate process would really simplify this stuff. Of course, I also do not want to make what is essentially a bug fix unnecessarily complex and would not oppose merging this as-is.

One more idea to make process isolation simpler: Instead of integration tests, we could "just" execute the test binary using `std::process::Command` with a filter for the current test case and some environment variable set (to indicate to the test case that it is running within an isolated process and should do the actual test logic instead of spawning the subprocess). Of course, spawning the subprocess with the same flags as ourselves but the additional filter in a portable manner is probably a bit more than "just"...

davidhewitt:
What's the plan here? Sounds like we didn't want to overcomplicate... should I merge as-is?

adamreichold:
> What's the plan here? Sounds like we didn't want to overcomplicate... should I merge as-is?

Almost. I think we should still fix the assertions on the content of `POOL` to just use `contains(ptr)` to able be robust against interference from other test cases. And then merge as-is without doing any fundamental changes to how these tests work.

adamreichold (approved):

PR #2196, opened 2022-02-28
clippy: enable some more lints

davidhewitt:
Adds some extra clippy lints which I think are worth having. We can tweak this list further later.

### DIFF
--- a/src/impl_/pyclass.rs
+++ b/src/impl_/pyclass.rs
@@ -847,12 +847,12 @@ pub struct ThreadCheckerImpl<T>(thread::ThreadId, PhantomData<T>);
 
 impl<T> PyClassThreadChecker<T> for ThreadCheckerImpl<T> {
     fn ensure(&self) {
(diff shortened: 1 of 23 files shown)


src/impl_/pyclass.rs
```diff
@@ -847,12 +847,11 @@ pub struct ThreadCheckerImpl<T>(thread::ThreadId, PhantomData<T>);
 
 impl<T> PyClassThreadChecker<T> for ThreadCheckerImpl<T> {
     fn ensure(&self) {
-        if thread::current().id() != self.0 {
-            panic!(
-                "{} is unsendable, but sent to another thread!",
-                std::any::type_name::<T>()
-            );
-        }
+        assert!(
```
adamreichold, line 850: Wouldn't `assert_eq!` even be nicer as it would include the thread ID in the message?

davidhewitt: Good idea!

PR #2198, opened 2022-03-01
use platform.system to check if it is on windows

PanQL:
the code ```hasattr(platform, "win32_ver")``` will also be True on other os rather than Windows, this pr replace it by platform.system().


davidhewitt (approved):
Thanks, I verified you are correct about this, my macOS computer on Python 3.10 displays the wrong behaviour.

I've added a CHANGELOG & will now merge.

PanQL:
Hi, It seems that the CI has been failed due to network reason. And I didn't find where to rerun the CI. Would you mind to merge this pr manually? Thanks~

PR #2199, opened 2022-03-01
Enable the resolve-config feature of pyo3-build-config on docs.rs to make the dependent functionality visible.

adamreichold:
Noticed this while discussing how to handle https://github.com/PyO3/rust-numpy/pull/283 even though I am not sure if this is the correct way to resolve it, i.e. if the build environment on docs.rs does actually contain a Python interpreter?


davidhewitt:
👍 I'm pretty sure this is correct - PyO3 builds, so there must be some interpreter there!

PR #2201, opened 2022-03-02
Add aarch64-apple-darwin test to CI

### LINKED ISSUE #2200: universal build fails on new version?
### Bug Description

I upgraded ormsgpack to use newest version (0.16) and it caused the release job to fail.
https://github.com/aviramha/ormsgpack/actions/runs/1915644588
Seems it can't open the generated config file for some reason?

### Steps to Reproduce

[See the CI](https://github.com/aviramha/ormsgpack/actions/runs/1915644588)

### Backtrace

_No response_

macOS

all versions

1.58.1

### Your PyO3 version

0.16.0

actions/setup-python@v2

### Additional Info

_No response_

messense:
Fixes #2200

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -623,7 +623,16 @@ pub fn cross_compiling(
     let cross_lib_dir = env_var("PYO3_CROSS_LIB_DIR");
     let cross_python_version = env_var("PYO3_CROSS_PYTHON_VERSION");
 
-    let target_triple = format!("{}-{}-{}", target_arch, target_vendor, target_os);
+    let target_triple = format!(
+        "{}-{}-{}",
+        target_arch,
+        target_vendor,
+        if target_os == "macos" {
+            "darwin"
+        } else {
+            target_os
+        }
+    );
 
     if cross.is_none() && cross_lib_dir.is_none() && cross_python_version.is_none() {
         // No cross-compiling environment variables set; try to determine if this is a known case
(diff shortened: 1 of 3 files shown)


messense:
I think https://github.com/PyO3/pyo3/pull/1428 is broken in 0.16.

davidhewitt:
Yes, I think also hitting this on `setuptools-rust` - https://github.com/PyO3/setuptools-rust/pull/209

Think we should get this out as a patch release 0.16.1. Just need someone to figure out what's broken. I _might_ be able to look at this tomorrow, not sure.

pyo3-build-config/src/impl_.rs
```diff
@@ -623,7 +623,7 @@ pub fn cross_compiling(
     let cross_lib_dir = env_var("PYO3_CROSS_LIB_DIR");
     let cross_python_version = env_var("PYO3_CROSS_PYTHON_VERSION");
 
-    let target_triple = format!("{}-{}-{}", target_arch, target_vendor, target_os);
```
messense, line 626: The problem is that this results in `aarch64-apple-macos` instead of `aarch64-apple-darwin`.

pyo3-build-config/src/impl_.rs
```diff
@@ -623,7 +623,7 @@ pub fn cross_compiling(
...
+    let target_triple = cargo_env_var("TARGET").ok_or("expected TARGET env var")?;
```
messense, line 626: FIXME: we shouldn't rely on `TARGET` env var in this public API.

pyo3-build-config/src/impl_.rs
davidhewitt: Thank you! If tests pass and a `Fixed` CHANGELOG entry added, please feel free to merge this!

PR #2197, opened 2022-03-01
Extend  optional dependency supported versions to include 0.12.

adamreichold:
The tests seem to work when building against `hashbrown` version 0.12 to I opted to just extend the version specification.

PR #2204, opened 2022-03-04
Fix up changelog entry for `#2197`

messense:


adamreichold:
Thanks for catch and fixing this! And sorry for messing it up.

PR #2205, opened 2022-03-04
Remove typying

JerzySpendel:
Well, I guess it didn't make sense to add type annotation if the value was to be immediately returned?


birkenfeld (approved):
AFAICS that's invalid syntax anyway... maybe leftover from an automatic replacement?

PR #2208, opened 2022-03-04
release: 0.16.1

davidhewitt:
The patch fixes that we've had in the past few days seem worth putting out as a release.

If there's no objections, I'll put live tomorrow evening. Might rebase on #2207 first to pick up more guide tidy-ups.


davidhewitt:
Release went live last night!

PR #2213, opened 2022-03-06
guide: fix duplicated example for #[new]

birkenfeld:
The second one was probably intended to show a fallible `#[new]`.

Also show that the method does not need to be named `new()`, which
is nice because `new()` can be used for a Rust-level constructor
if they differ.


davidhewitt (approved):
Nice catch, thanks 👍

PR #2217, opened 2022-03-08
pypy: support 7.3.8

davidhewitt:
According to https://foss.heptapod.net/pypy/pypy/-/issues/3688#note_181074 the official PyPy line will be that PyPy 3.7 versions older than 7.3.8 should be treated as bugged.

Hence here I'm updating definitions to support 7.3.8 and add a warning if a user tries to import a PyO3 module with an older PyPy version.

```
  /Users/david/tools/pypy3.7-v7.3.8-osx64/lib-python/3/importlib/_bootstrap.py:228: UserWarning: PyPy 3.7
versions older than 7.3.8 are known to have binary compatibility issues which may cause segfaults. Please
upgrade.
```

### DIFF
--- a/src/impl_/pymodule.rs
+++ b/src/impl_/pymodule.rs
@@ -71,6 +71,21 @@ impl ModuleDef {
         panic_result_into_callback_output(
             py,
             std::panic::catch_unwind(move || -> PyResult<_> {
+                #[cfg(all(PyPy, not(Py_3_8)))]
+                {
+                    const PYPY_GOOD_VERSION: [u8; 3] = [7, 3, 8];
+                    let version = py
+                        .import("sys")?
+                        .getattr("implementation")?
+                        .getattr("version")?;
+                    if version.lt(crate::types::PyTuple::new(py, &PYPY_GOOD_VERSION))? {
+                        let warn = py.import("warnings")?.getattr("warn")?;
+                        warn.call1((
+                            "PyPy 3.7 versions older than 7.3.8 are known to have binary \
+                             compatibility issues which may cause segfaults. Please upgrade.",
+                        ))?;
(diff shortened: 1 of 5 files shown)


davidhewitt:
cc @mattip

src/impl_/pymodule.rs
```diff
@@ -71,6 +71,21 @@ impl ModuleDef {
         panic_result_into_callback_output(
             py,
             std::panic::catch_unwind(move || -> PyResult<_> {
+                #[cfg(all(PyPy, not(Py_3_8)))]
+                {
+                    const PYPY_GOOD_VERSION: [u8; 3] = [7, 3, 9];
```
mattip, line 76: Maybe I don't understand?

```suggestion
                    const PYPY_GOOD_VERSION: [u8; 3] = [7, 3, 8];
```

mattip:
> ` /Users/david/tools/pypy3.7-v7.3.8-osx64/lib-python/3/importlib/_bootstrap.py:228: UserWarning:`

Shouldn't v7.3.8 be clean, and the warning only emitted for v7.3.7 and under?

src/impl_/pymodule.rs
davidhewitt: Haha good catch, I set it higher when manually testing 😅

mattip: :+1:

davidhewitt:
> Shouldn't v7.3.8 be clean, and the warning only emitted for v7.3.7 and under?

Yes, I was running 7.3.8 locally so fiddled with versions to verify the warning. Force-pushed a corrected implementation which should now only warn on 7.3.7 and older ;)

PR #2226, opened 2022-03-15
Limit concurrency on GitHub Actions

messense:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

Hopefully this will reduce wasted CI run times.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -4,6 +4,10 @@ on:
       - main
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
 name: Benchmark
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
       - main
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
 env:
   CARGO_TERM_COLOR: always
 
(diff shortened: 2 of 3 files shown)


davidhewitt:
Nice, I had some other ideas too:
 - we could test 32-bit windows only with the latest Python version
 - we could split the pipeline into a "fast" set of checks, maybe run only for latest Python version, and then follow up with the full test suite

messense:
> we could split the pipeline into a "fast" set of checks, maybe run only for latest Python version, and then follow up with the full test suite

That might slow down the overall run time, I think we should experiment with it in a separate PR.

adamreichold:
> >  we could split the pipeline into a "fast" set of checks, maybe run only for latest Python version, and then follow up with the full test suite

> That might slow down the overall run time, I think we should experiment with it in a separate PR.

Having done this for `rust-numpy`, I would like to add that this is not just about the time to results, but also about not wasting computation on PR that have more fundamental issues, and thereby saving energy. I also think this is bearable if the "fast" is sufficiently fast, i.e. for `rust-numpy` the checks that run before the full test matrix is started usually complete in a minute or so.

So also from the point of view of environmental impact I would also argue for this, but of course this does not mean this can happen in a separate PR.

davidhewitt:
Agreed, this PR is already great, we can improve the pipeline in another PR. Just throwing out ideas 😊

messense:
https://github.com/PyO3/pyo3/blob/d0c124d25accbb678b4fbdf2baa824681fd69d3c/.github/workflows/ci.yml#L82

I think this `fail-fast: false` is a big source of waste. I'd suggest that if we want to debug CI workflow sometimes, we can add a label to PR, for example `CI-no-fail-fast` to signal it: `fail-fast: ${{ contains( github.event.pull_request.labels.*.name, 'CI-no-fail-fast') }}`.

PR #2227, opened 2022-03-15
CI: keep going on failure only if `CI-no-fail-fast` label is present

messense:
Implements https://github.com/PyO3/pyo3/pull/2226#issuecomment-1068131210

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,8 @@ jobs:
     needs: [fmt]
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


messense:
Looks like it's working✨.

* Without `CI-no-fail-fast`, the jobs quickly failed: https://github.com/PyO3/pyo3/actions/runs/1987785951 .
* WIth the label, the rest of the jobs keep going on failure: https://github.com/PyO3/pyo3/actions/runs/1987810013 .

adamreichold (approved):

Issue #2222, opened 2022-03-13
Cross-compilation to Windows randomly fails after upgrade to 0.16


ravenexp:
### Bug Description

After upgrading pyo3 from 0.15.1 to 0.16.1 the cross-compile builds have started failing with the following error:

```
error: failed to run custom build command for `pyo3 v0.16.1`

Caused by:
  process didn't exit successfully: `/home/raven/testlib/target/debug/build/pyo3-647dcf2b3938b785/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'failed to parse PyO3 config file: Error { value: "failed to open PyO3 config file at /home/raven/testlib/target/debug/build/pyo3-build-config-d6873a340c171ea3/out/pyo3-cross-compile-config.txt", source: Some(Os { code: 2, kind: NotFound, message: "No such file or directory" }) }', /home/raven/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-build-config-0.16.1/src/lib.rs:78:10
```

Restarting the failing build command 2 to 3 times results in a successful build.
I think there is a race condition somewhere in the PyO3 build scripts that causes these failures.

I've verified that `pyo3-cross-compile-config.txt` is not created in the build script output directory on the first (failing) invocation, but `pyo3-build-config.txt`, `pyo3-build-config-abi3.txt` and `pyo3-build-config-file.txt` are created.
After re-running the build command several times `pyo3-cross-compile-config.txt` is created and the build succeeds.

### Steps to Reproduce

1. Create a new PyO3 extension module with e.g. `maturin new -b pyo3 testlib`.
2. Run `PYO3_CROSS_LIB_DIR=/ cargo build --target x86_64-pc-windows-gnu`
3. Repeat the previous command several times until the build script errors disappear and the build fails at the linking error such as `x86_64-w64-mingw32/bin/ld: cannot find -lpython3`.

### Your operating system and version

Arch Linux

### Your Python version (`python --version`)

Python 3.10.2

### Your Rust version (`rustc --version`)

rustc 1.59.0 (9d1b2106e 2022-02-23)

### Your PyO3 version

0.16.1

### How did you install python? Did you use a virtualenv?

pacman

messense:
Same kind of issue in https://github.com/messense/maturin-action/pull/34

davidhewitt:
[...]

PR #2228, opened 2022-03-15
release: 0.16.2

adamreichold:
As discussed in #2222, it might make sense to put out another release due to another regression in cross compilation.


davidhewitt:
Thanks. I will put this live tonight when I have a moment of calm before bed. Currently stuck under a baby 🥰

davidhewitt:
Release is live! Thanks for the help setting this up.

Issue #2229, opened 2022-03-16
failed to open PyO3 config file at `pyo3-cross-compile-config.txt` in manylinux2010 docker container


messense:
### Bug Description

```
error: failed to run custom build command for `pyo3 v0.16.2`

Caused by:
  process didn't exit successfully: `/home/runner/work/maturin-action/maturin-action/maturin/test-crates/pyo3-mixed/target/debug/build/pyo3-241f0dea0e3e2ffc/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'failed to parse PyO3 config file: Error { value: "failed to open PyO3 config file at /home/runner/work/maturin-action/maturin-action/maturin/test-crates/pyo3-mixed/target/debug/build/pyo3-build-config-91712567431bbcb4/out/pyo3-cross-compile-config.txt", source: Some(Os { code: 2, kind: NotFound, message: "No such file or directory" }) }', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-build-config-0.16.2/src/lib.rs:78:10
```

After #2224 it still fails with the same kind of error message on Linux.

### Steps to Reproduce

https://github.com/messense/maturin-action/runs/5563946271?check_suite_focus=true

Run the following command in manylinux2010 docker image on [pyo3-mixed](https://github.com/PyO3/maturin/tree/main/test-crates/pyo3-mixed) test crate:

```
export PYO3_CROSS_LIB_DIR=/opt/python/cp38-cp38/lib
maturin build -i python3.8 --no-sdist --manylinux 2010
```

### Backtrace

```shell
0: rust_begin_unwind
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:498:5
     1: core::panicking::panic_fmt
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:116:14
     2: core::result::unwrap_failed
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/result.rs:1690:5
     3: core::result::Result<T,E>::expect
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/result.rs:975:23
     4: pyo3_build_config::get::{{closure}}
               at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-build-config-0.16.2/src/lib.rs:69:9
[...]

Issue #2191, opened 2022-02-26
Improving coverage


davidhewitt:
I believe Rust 1.60 (due in six weeks' time) will contain a stabilised `-C instrument-coverage` flag to support the coverage for which we currently use nightly.

It might be interesting when that lands to rewrite our CI a bit so that coverage is done as part of all jobs, and we can then remove the separate coverage jobs. That should hopefully help give a truer picture of sections which really are covered however not reflected by the current setup.

Once that change is done, this issue (or another one) can become a general call to action for interested volunteers who'd like to help us bring that coverage number up towards 100% (the ideal, not sure if realistic).

The general steps I'd recommend for anyone who'd be interested in helping achieve that would be to look at the table of files at the bottom of https://app.codecov.io/gh/PyO3/pyo3, identify uncovered sections, and then write PRs adding tests to cover them.

Some issues with coverage that I'm aware of:
 - Possible weirdness with codecov and PRs using `main` on the fork: https://github.com/PyO3/pyo3/pull/2175#issuecomment-1051104614 and 
 - Probably impossible to cover some `assert!` messages: https://github.com/PyO3/pyo3/pull/2175#issuecomment-1051111519
 - Doctests don't contribute to coverage https://github.com/taiki-e/cargo-llvm-cov/issues/2

davidhewitt:
Closing in favour of #2453

PR #2232, opened 2022-03-16
pyo3-build-config: fix windows "cross-compile" panic

davidhewitt:
Possible fix for #2229

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -540,6 +540,54 @@ fn is_abi3() -> bool {
     cargo_env_var("CARGO_FEATURE_ABI3").is_some()
 }
 
+#[derive(Debug, PartialEq)]
+struct TargetInfo {
+    /// The `arch` component of the compilation target triple.
+    ///
+    /// e.g. x86_64, i386, arm, thumb, mips, etc.
+    arch: String,
+
+    /// The `vendor` component of the compilation target triple.
+    ///
+    /// e.g. apple, pc, unknown, etc.
+    vendor: String,
+
+    /// The `os` component of the compilation target triple.
+    ///
+    /// e.g. darwin, freebsd, linux, windows, etc.
+    os: String,
+}
+
+impl TargetInfo {
+    fn from_cargo_env() -> Result<Self> {
+        Ok(Self {
+            arch: cargo_env_var("CARGO_CFG_TARGET_ARCH")
+                .ok_or("expected CARGO_CFG_TARGET_ARCH env var")?,
+            vendor: cargo_env_var("CARGO_CFG_TARGET_VENDOR")
+                .ok_or("expected CARGO_CFG_TARGET_VENDOR env var")?,
+            os: cargo_env_var("CARGO_CFG_TARGET_OS")
+                .ok_or("expected CARGO_CFG_TARGET_OS env var")?,
+        })
+    }
+
+    fn to_target_triple(&self) -> String {
+        format!(
+            "{}-{}-{}",
+            if self.arch == "x86" {
+                "i686"
+            } else {
+                &self.arch
+            },
+            self.vendor,
+            if self.os == "macos" {
+                "darwin"
+            } else {
+                &self.os
+            }
+        )
+    }
+}
+
 /// Configuration needed by PyO3 to cross-compile for a target platform.
 ///
 /// Usually this is collected from the environment (i.e. `PYO3_CROSS_*` and `CARGO_CFG_TARGET_*`)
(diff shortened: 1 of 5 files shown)


messense:
Actually it's linux not windows now.

pyo3-build-config/src/impl_.rs
```diff
@@ -638,6 +627,20 @@ pub fn cross_compiling(
         // No cross-compiling environment variables set; try to determine if this is a known case
         // which is not cross-compilation.
 
+        if host.starts_with(&target_triple) {
```
adamreichold, line 630: So this is basically subsuming the `host == target` condition with the idea that setting the environment variables should force cross compilation even if host and target are exactly equal?

pyo3-build-config/src/impl_.rs
```diff
@@ -638,6 +627,20 @@ pub fn cross_compiling(
...
+            // Not cross-compiling if arch-vendor-os is all the same
+            // e.g. x86_64-unknown-linux-musl on x86_64-unknown-linux-gnu host
+            //      x86_64-pc-windows-gnu on x86_64-pc-windows-msvc host
+            return Ok(None);
+        }
+
+        dbg!(&target_triple, &host);
+
+        if target_triple == "i686-pc-windows" && host.starts_with("x86_64-pc-windows") {
```
adamreichold, line 639: Minor nit: `target_tiple` seems to be `format!`ed unconditionally but is only used within the if clause

adamreichold:
> Actually it's linux not windows now.

I think this could still work, i.e. `host` and `target` are `x86_64-unknown-linux-gnu` in your test case which would have been rejected for cross compiling due to the `host == target` condition when writing but not when reading the configuration.

messense:
Here we correctly force cross compilation if `PYO3_CROSS_LIB_DIR` is set
https://github.com/PyO3/pyo3/blob/01788b722002d5c1494332fd04dbdbb23c268103/pyo3-build-config/src/lib.rs#L73

But in here the logic is a little different

https://github.com/PyO3/pyo3/blob/01788b722002d5c1494332fd04dbdbb23c268103/pyo3-build-config/src/lib.rs#L180

adamreichold:
Yeah, we should probably bite the bullet and properly unify those code paths. @davidhewitt If you are busy I can cook up something here or elsewhere factoring out the logic for checking for cross compilation and creating that configuration?

adamreichold:
> But in here the logic is a little different

I suspect that is actually necessary because of cases where we do detect cross-compilation, but do not have the environment variables set. I think this fix here only takes care of the other way around?

messense:
> I suspect that is actually necessary because of cases where we do detect cross-compilation, but do not have the environment variables set.

Yeah it's necessary, I think it only missed that it should also do cross compilation when the environment variables are set.

davidhewitt:
I think we should probably unify those code paths a bit, I have some ideas.

I'm enjoying having a go at this at the moment - might take me a few gaps through the day. Got a bit of a routine beginning to form. If things get busy later I might push, run, and ping you @adamreichold 😅

messense:
BTW cross compilation when host and target architectures are the same should be supported, we have an issue in the past: https://github.com/PyO3/pyo3/issues/1513

#2230 should be included in CI when we fix this.

pyo3-build-config/src/impl_.rs
```diff
@@ -1280,13 +1304,33 @@ fn fixup_config_for_abi3(
...
+    let target_os =
+        cargo_env_var("CARGO_CFG_TARGET_OS").ok_or("expected CARGO_CFG_TARGET_OS env var")?;
+
+    if env_vars.any() {
+        let cross_config =
+            cross_compiling_from_vars(&target_arch, &target_os, &target_vendor, env_vars)?;
+        let mut interpreter_config = load_cross_compile_config(cross_config)?;
+        fixup_config_for_abi3(&mut interpreter_config, get_abi3_version())?;
+        return Ok(Some(interpreter_config));
+    }
```
adamreichold, line 1325: I think if-Some-else-None would be easier to read than the early return here.

pyo3-build-config/src/impl_.rs
```diff
@@ -1280,13 +1304,33 @@ fn fixup_config_for_abi3(
...
+
+    ensure!(
+        host == target || is_not_cross_compiling(&host, &target_arch, &target_os, &target_vendor),
+        "PyO3 detected compile host {host} and build target {target}, but none of PYO3_CROSS, PYO3_CROSS_LIB_DIR \
```
adamreichold, line 1329: Using inline placeholders will probably not work on our MSRV? So this would be missing the `host=host` and `target=target` arguments to the `ensure!` macro?

pyo3-build-config/src/impl_.rs
```diff
@@ -677,6 +668,39 @@ pub fn cross_compiling(
...
+            target_os
+        }
+    );
+
+    // Not cross-compiling if arch-vendor-os is all the same
+    // e.g. x86_64-unknown-linux-musl on x86_64-unknown-linux-gnu host
+    //      x86_64-pc-windows-gnu on x86_64-pc-windows-msvc host
+    host.starts_with(&target_triple)
+        // Not cross-compiling to compile for 32-bit Python from windows 64-bit
+        || (target_triple == "i686-pc-windows" && host.starts_with("x86_64-pc-windows"))
```
adamreichold, line 697: The failure of the Windows job suggests that something is still off here, but debugging it still hindered by the `ensure!` formatting not working.

adamreichold:
> https://github.com/PyO3/pyo3/pull/2230 should be included in CI when we fix this.

I think it would make sense to cherry-pick that commit here so that we have only a single PR.

davidhewitt:
> I think it would make sense to cherry-pick that commit here so that we have only a single PR.

Agreed, I'll cherry-pick once windows job fixed.

pyo3-build-config/src/impl_.rs
adamreichold: So this should be
```rust
target_triple == "i686-windows-pc"
```
i.,e. the other of `windows` and `pc` being switched?

pyo3-build-config/src/impl_.rs
```diff
@@ -1280,13 +1306,36 @@ fn fixup_config_for_abi3(
...
+    let target_arch =
+        cargo_env_var("CARGO_CFG_TARGET_ARCH").ok_or("expected CARGO_CFG_TARGET_ARCH env var")?;
+    let target_vendor = cargo_env_var("CARGO_CFG_TARGET_VENDOR")
+        .ok_or("expected CARGO_CFG_TARGET_VENDOR env var")?;
+    let target_os =
+        cargo_env_var("CARGO_CFG_TARGET_OS").ok_or("expected CARGO_CFG_TARGET_OS env var")?;
+
+    let interpreter_config = if env_vars.any() {
+        let cross_config =
+            cross_compiling_from_vars(&target_arch, &target_os, &target_vendor, env_vars)?;
```
adamreichold, line 1323: `target_os` and `target_vendor` are switched here. Maybe this would benefit from a struct with named fields similar to `CrossCompileEnvVars`?

pyo3-build-config/src/impl_.rs
adamreichold: Actually no, the issue seems to be with the call site switching os and vendor.

pyo3-build-config/src/impl_.rs
davidhewitt: Added. Though I didn't change `cross_compiling` to take it as input because that's a public API. Think it's fine for now.

davidhewitt:
Looks great, let's merge? I'm hoping we can tidy the coverage later in #2191

adamreichold (approved):

PR #2233, opened 2022-03-17
Consider PYTHONFRAMEWORK when cross-compiling

kmp1:
When cross compiling on macos the `PYTHONFRAMEWORK` setting wasn't considered, though it is in the regular build process.  

This means that the build system believes it is a static Python when, actually, it's shared.  This adds an extra check to make `from_sysconfigdata` much more similar to `from_interpreter` in that regard.


messense (approved):
Looks correct to me, thanks!

kmp1:
Didn't know you had a code coverage check - I've added an additional test for this too.

PR #2235, opened 2022-03-19
Added an as_bytes method for Py<PyBytes>

alex:


src/types/bytes.rs
```diff
@@ -97,6 +97,24 @@ impl PyBytes {
     }
 }
 
+impl Py<PyBytes> {
+    /// Gets the Python bytes as a byte slice. Because Python bytes are
+    /// immutable, the result may be used for as long as the reference to
+    /// `self` is held.
+    pub fn as_bytes(&self, _py: Python<'_>) -> &[u8] {
```
adamreichold, line 104: I understand that it is implied, but since it seems to be the point of this, wouldn't it be nice to make it explicit which lifetime is attached to the return value, e.g.
```rust
pub fn as_bytes<'a>(&'a self, _py: Python<'_>) -> &'a [u8] {
```

alex: Good call, done.

alex:
Failing test is a network flake, can someone rerun those builds?

src/types/bytes.rs
mejrs: > I understand that it is implied, but since it seems to be the point of this, wouldn't it be nice to make it explicit which lifetime is attached to the return value, e.g.

It's not about being explicit - this function is unsound without it. Without `&'a self` this function can produce a slice with an unbounded lifetime.

src/types/bytes.rs
```diff
@@ -97,6 +97,24 @@ impl PyBytes {
     }
 }
 
+impl Py<PyBytes> {
+    /// Gets the Python bytes as a byte slice. Because Python bytes are
+    /// immutable, the result may be used for as long as the reference to
+    /// `self` is held.
+    pub fn as_bytes<'a>(&self, _py: Python<'_>) -> &'a [u8] {
```
mejrs, line 104: This produces a slice with an unbounded lifetime. It must be bound on the lifetime of self.

```suggestion
    /// Gets the Python bytes as a byte slice. Because Python bytes are
    /// immutable, the result may be used for as long as the reference to
    /// `self` is held.
    pub fn as_bytes<'a>(&'a self, _py: Python<'_>) -> &'a [u8] {
```

mejrs (changes requested):

src/types/bytes.rs
mejrs: As far as I understand it is OK to use this slice after dropping the GIL (?). This is probably a good idea to clarify in the docs.

alex: a) Whoops, that was an accident, I'd meant to put `&'a self`,
b) docs added

davidhewitt (approved):
Looks reasonable to me, thanks!

I think there's a few other immutable types like `str` and `tuple` which I guess can also have treatment like this where `py` is necessary for reading, but once read, the GIL can be released. Might be interesting to follow up this another time.

alex:
Yes, I'm sure there's other cases where this pattern would be helpful -- I send this particularly because I have a use case in pyca/cryptography for this, where we'll be able to optimize some code to avoid allocations/copying now.

PR #2236, opened 2022-03-20
Run tests on CI with cargo-nextest

messense:
https://github.com/nextest-rs/nextest

Try it out to see if it can help speeding up CI runs.


messense:
> test marker::tests::test_acquire_gil ... FAILED

This one seems problematic with cargo-nextest.

birkenfeld:
Last time I checked, nextest did not support and run doctests. So you'd have to add "cargo test --doc" as a separate step?

messense:
> Last time I checked, nextest did not support and run doctests. So you'd have to add "cargo test --doc" as a separate step?

In that case I think we should wait for https://github.com/nextest-rs/nextest/issues/16, otherwise the added `cargo test --doc` steps might just slow things down and make the CI setup more complex.

davidhewitt:
As a bit of housekeeping I'm going to close this for now, we can reopen when nextest supports doctests.

Issue #2225, opened 2022-03-15
#[pyclass] struct methods cannot move self


DecrepitHuman:
### Bug Description

Any methods implemented on #[pyclass] structs that attempt to take ownership and move `self` produce type mismatch errors

### Steps to Reproduce

1. Create #[pyclass] in module
2. Implement function taking ownership of `self`

### Backtrace

```shell
error[E0308]: mismatched types
 ---> src/tcp/server.rs:10:1
   |
10 | #[pymethods]
   | ^^^^^^^^^^^^ expected struct `Server`, found mutable reference
   |
   = note:         expected struct `Server`
           found mutable reference `&mut PyRefMut<'_, Server>`
```
```


### Your operating system and version

Ubuntu 20.04.3 LTS

### Your Python version (`python --version`)

3.8.10

### Your Rust version (`rustc --version`)

rustc 1.59-0-nightly (23f69235a 2021-12-20)

### Your PyO3 version

0.16.1

### How did you install python? Did you use a virtualenv?

apt

### Additional Info

_No response_

birkenfeld:
You can't take ownership of `self` since it's wrapped by a Python object, and that Python object can be referenced anywhere.

But the error could be better, so let's keep this open for an improvement.

birkenfeld:
Also, the guide only says "`&self` and `&mut self` can be used" about instance methods. This should be expanded to cover the other possible receivers, and the caveats with `&mut self`.

adamreichold:
> however I'm in a situation where I need to manually drop some properties on the struct, and for that I need ownership of self

I suspect that you will need to do "the option dance", i.e. replace a field of type `TcpStream` by `Option<TcpStream>` so you can drop the field using `Option::take` using only a `&mut self` to the struct.

birkenfeld:
Yes, or something like `enum Socket { Tcp(TcpStream), Udp(UdpSocket) [, None] }`

DecrepitHuman:
Great, thanks for the fast responses. I appreciate it. Will keep issue open as asked above

mejrs:
> But the error could be better, so let's keep this open for an improvement.

This should be fairy straightforward, right? Just emit a custom error if the receiver is `self`?

adamreichold:
From https://github.com/PyO3/pyo3/pull/2238#discussion_r830675173

[...]

PR #2238, opened 2022-03-20
Create a better error message for moving self

mejrs:


pyo3-macros-backend/src/method.rs
```diff
@@ -735,3 +742,5 @@ fn parse_method_attributes(
 }
 
 const IMPL_TRAIT_ERR: &str = "Python functions cannot have `impl Trait` arguments";
+const RECEIVER_BY_VALUE_ERR: &str = "PyO3 does not support `self` as a receiver, because it cannot move `self` out of the Python interpreter.
```
adamreichold, line 745: I do not really have a good proposal but since this is primarily about the error message, I wanted to say that "moving out of the Python interpreter" confused me a bit during the first reading.

AFAIU, the main issue is that all Python objects have shared ownership (similar to e.g. `Rc`), so we cannot leave a deinitialized object behind? (I think we could theoretically implement fallibly moving out of objects if their reference count is one, but I suspect that this basically never happens in Python code.)

So maybe we could word this in such a manner? Maybe something like "it cannot move from `self` as other Python references to it might still exist"?

mejrs: I've always thought of it as "the interpreter owns the object, you just own a reference to it". So to me it makes most sense to say that you can't move it out.

That said I think it doesn't really matter for the user to understand it perfectly if one of the suggestions will do.

adamreichold: > "the interpreter owns the object, you just own a reference to it"

That sounds a bit arbitrary as the interpreter will not hold on to objects just for the sake of it, only because other Python code has references to it (which should apply even to tracing JIT like PyPy).

If only the interpreter would hold on to those objects, it would probably also not be a problem since the only completely generic thing a compiler can do to any object is to destroy/deallocate it so we would just need to leave a `PyCell` that can be destroyed.

(Writing that, we could probably support moving out of `self` if `PyCell` were adjust to support only `2^64 - 3` instead of `2^64 - 2` shared borrows thereby creating another niche that can be used for moved from values turning it into something like a `RefCell<Option<T>>`, but the ergonomics of all subsequent method invocations not explicitly taking `slf: &PyCell<T>` panicking would probably be less than nice.)

mejrs: https://github.com/PyO3/pyo3/issues/2225 would be a better place to discuss whether we want to support something like that.

> "it cannot move from self as other Python references to it might still exist"?

I believe the reference count is incremented when the method is called (?) so there will always be other references to it?

adamreichold: > I believe the reference count is incremented when the method is called (?) so there will always be other references to it?

Yes, function arguments always seem to have an elevated reference count with CPython. This is what I meant with

> I think we could theoretically implement fallibly moving out of objects if their reference count is one, but I suspect that this basically never happens in Python code.

So maybe we could just say that "Python objects always have shared ownership"? (Mainly because "shared ownership" is term used in the official Rust documentation so people might have read it before.)

adamreichold: > https://github.com/PyO3/pyo3/issues/2225 would be a better place to discuss whether we want to support something like that.

Agreed. Moved the comment there.

mejrs: @adamreichold what do you think about "Python objects are shared, so 'self' cannot be moved out of the Python interpreter."

adamreichold: Agreed. I would find "Python object always have shared ownership, so `self` cannot be moved out of the Python interpeter" even nicer but I'd be happy as you propose.

birkenfeld: I have to agree that "Python interpreter" as the owner sounds strange here.

davidhewitt:
Thanks, it would be nice to improve this papercut!

For the sake of alternatives, can I throw out two ideas:
- allow `self` receiver only for `Copy` types. I think we could use a custom compile-time assertion to hand out a nice error message in non-copy cases.
- have a helper type which we could use as `slf: CopyFromPy<Self>`. (Or `CloneFromPy` even.) I think that all would be needed would be an implementation of `TryFromPyCell` for the helper type.

In particular, for small enums like `#[repr(u8)]` it seems a shame to take by reference & all accompanying borrow checking overhead.

birkenfeld:
Allowing self for copy types sounds nice, but may be a gotcha for Python oriented folks, where all methods take `self`. But a good guide explanation should counter that.

adamreichold:
> In particular, for small enums like `#[repr(u8)]` it seems a shame to take by reference & all accompanying borrow checking overhead.

This sounds rather like it would require something similar to #1979 ?

davidhewitt:
To avoid the borrow tracking, yes I think that would be needed too.

mejrs:
Yeah we could make it work, but I think it would be mistake. I think most of the time people do not actually want a by-value receiver and when they do they can, if it is `Copy`, just copy it out themselves. We would just be adding a footgun similar to "pyo3(get) clones my field". For example the `Server` struct in https://github.com/PyO3/pyo3/issues/2225 doesn't sound like it's `Copy`, so moving it out would never work.

That said, does this discussion block merging the PR? I'd prefer to move forward with this, and we can change/remove it if we decide `self` receivers is what we want.

davidhewitt (approved):
👍 fair counterpoint; I think let's start here. If later there are cases where we can demonstrate sufficient performance / convenience improvements by using `self` then we can worry about how to best support.

PR #2239, opened 2022-03-21
allow to compile with parking_lot 0.12

zh-jq:


messense:
Hmm, I'm not of ideas, even with `--lock --frozen --offline` it still updates the lockfile to use latest version of parking_lot in the setuptools-rust example.

adamreichold:
> Hmm, I'm not of ideas, even with `--lock --frozen --offline` it still updates the lockfile to use latest version of parking_lot in the setuptools-rust example.

I think the problem is that the `cargo update --precise` steps will crate a `Cargo.lock` in the repository root, i.e. for the `pyo3` crate. This is sufficient for its compile tests which where the only ones which needed it for now as the downgraded dependencies are optional and the examples do not use them. But `parking_lot` is not optional and hence I suspect that we would need to create a lock file for each of the  examples.

adamreichold (approved):
I think this is alright without a CHANGELOG entry as it is not a public dependency.

messense:
`parking-lot`'s MSRV is bumped to 1.49.0.

https://github.com/Amanieu/parking_lot/blob/master/CHANGELOG.md#parking_lot-0120-parking_lot_core-090-lock_api-046-2022-01-28

messense:
Thank you @zh-jq !

Issue #2231, opened 2022-03-16
Generate `python3.dll` import libraries for Windows targets in the build script


ravenexp:
I'm routinely cross-compiling Python extension modules for Windows from a Linux host using the well-supported `x86_64-pc-windows-gnu` rustc target. The main issue I'm having is finding a suitable `python3.dll` import library (for `abi3` extensions) to use with `PYO3_CROSS_LIB_DIR`.

MinGW and MSVC compilers use incompatible import library object formats, so downloading a Windows Python distribution does not work, and compiling Python with MinGW compiler is not supported by the upstream and requires about [a hundred](https://aur.archlinux.org/packages/mingw-w64-python) patches to work.

I've been working around this by using shell scripts to generate the MinGW-compatible `python3.dll` import library using `dlltool` from MinGW binutils. Luckily, the CPython project maintains [a list](https://github.com/python/cpython/blob/main/Misc/stable_abi.txt) of the stable ABI symbols in a machine-readable format, which can be used to generate a DLL module definition (.def) file and the DLL import library from it.

A few weeks ago I decided to ditch the shell script hacks and package everything into a self-contained Rust crate called [`python3-dll-a`](https://github.com/ravenexp/python3-dll-a). I'm currently using it in my own projects via `build.rs`, and the Windows cross-compile command is just this (no Windows Python files required):

```
PYO3_CROSS_LIB_DIR=target/python3-dll cargo build --target x86_64-pc-windows-gnu
```

Ideally, I'd like to integrate it into `pyo3-build-config`, so that cross-compiling `abi3` extension modules for Windows would just work out of the box without messing with any `PYO3_CROSS*` environment variables. This should be already the case for Unix-like platforms where linking extension modules does not require a Python shared library.

The current `python3-dll-a` version can generate `python3.dll` import libraries for 
`x86_64-pc-windows-gnu`, `i686-pc-windows-gnu`, `x86_64-pc-windows-msvc`, `i686-pc-windows-msvc` and `aarch64-pc-windows-msvc` compile targets and can support both cross and native compilation setting. As a limitation of this approach, targeting the `-msvc` environment ABI requires LLVM binutils to be installed on the compile host.

If there is an interest in integrating this solution, I can start working on a PR for PyO3.
This will be feature-gated at the `pyo3` crate level, of course.

messense:
See also https://github.com/PyO3/maturin/issues/830

messense:
[...]

PR #2241, opened 2022-03-22
pyo3-build-config: Make `PYO3_CROSS_LIB_DIR` optional

ravenexp:
Try to support most common extension module cross compilation targets
without requiring any of the `PYO3_CROSS_*` environment variables.

This is a work in progress tangentially related to https://github.com/PyO3/pyo3/issues/2231
I'm creating this mostly to get some initial feedback from the `pyo3-build-config` maintainers.

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -685,7 +685,7 @@ fn is_linking_libpython_for_target(target: &Triple) -> bool {
 #[derive(Debug, PartialEq)]
 pub struct CrossCompileConfig {
     /// The directory containing the Python library to link against.
-    pub lib_dir: PathBuf,
+    pub lib_dir: Option<PathBuf>,
 
     /// The version of the Python library to link against.
     version: Option<PythonVersion>,
@@ -745,8 +745,10 @@ impl CrossCompileConfig {
     ///
     /// The conversion can not fail because `PYO3_CROSS_LIB_DIR` variable
     /// is ensured contain a valid UTF-8 string.
-    fn lib_dir_string(&self) -> String {
-        self.lib_dir.to_str().unwrap().to_owned()
+    fn lib_dir_string(&self) -> Option<String> {
+        self.lib_dir
+            .as_ref()
+            .map(|s| s.to_str().unwrap().to_owned())
     }
 }
 
@@ -786,7 +788,7 @@ impl CrossCompileEnvVars {
     /// into a `PathBuf` instance.
     ///
     /// Ensures that the path is a valid UTF-8 string.
-    fn lib_dir_path(&self) -> Result<PathBuf> {
+    fn lib_dir_path(&self) -> Result<Option<PathBuf>> {
         let lib_dir = self.pyo3_cross_lib_dir.as_ref().map(PathBuf::from);
 
         if let Some(dir) = lib_dir.as_ref() {
@@ -794,11 +796,9 @@ impl CrossCompileEnvVars {
                 dir.to_str().is_some(),
                 "PYO3_CROSS_LIB_DIR variable value is not a valid UTF-8 string"
             );
-            Ok(dir.clone())
-        } else {
-            // FIXME: Relax this restriction in the future.
-            bail!("The PYO3_CROSS_LIB_DIR environment variable must be set when cross-compiling")
         }
+
+        Ok(lib_dir)
     }
 }
 
@@ -815,9 +815,9 @@ pub(crate) fn cross_compile_env_vars() -> CrossCompileEnvVars {
 /// This function relies on PyO3 cross-compiling environment variables:
 ///
 ///   * `PYO3_CROSS`: If present, forces PyO3 to configure as a cross-compilation.
-///   * `PYO3_CROSS_LIB_DIR`: Must be set to the directory containing the target's libpython DSO and
-///   the associated `_sysconfigdata*.py` file for Unix-like targets, or the Python DLL import
-///   libraries for the Windows target.
+///   * `PYO3_CROSS_LIB_DIR`: If present, must be set to the directory containing
+///   the target's libpython DSO and the associated `_sysconfigdata*.py` file for
+///   Unix-like targets, or the Python DLL import libraries for the Windows target.
 ///   * `PYO3_CROSS_PYTHON_VERSION`: Major and minor version (e.g. 3.9) of the target Python
 ///   installation. This variable is only needed if PyO3 cannnot determine the version to target
 ///   from `abi3-py3*` features, or if there are multiple versions of Python present in
@@ -1109,13 +1109,22 @@ fn ends_with(entry: &DirEntry, pat: &str) -> bool {
     name.to_string_lossy().ends_with(pat)
 }
 
-fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
+/// Finds the sysconfigdata file when the target Python library directory is set.
+///
+/// Returns `None` if the library directory is not available, and a runtime error
+/// when no or multiple sysconfigdata files are found.
+fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<Option<PathBuf>> {
     let mut sysconfig_paths = find_all_sysconfigdata(cross);
     if sysconfig_paths.is_empty() {
-        bail!(
-            "Could not find either libpython.so or _sysconfigdata*.py in {}",
-            cross.lib_dir.display()
-        );
+        if let Some(lib_dir) = cross.lib_dir.as_ref() {
+            bail!(
+                "Could not find either libpython.so or _sysconfigdata*.py in {}",
+                lib_dir.display()
+            );
+        } else {
+            // Continue with the default configuration when PYO3_CROSS_LIB_DIR is not set.
+            return Ok(None);
+        }
     } else if sysconfig_paths.len() > 1 {
         let mut error_msg = String::from(
             "Detected multiple possible Python versions. Please set either the \
@@ -1129,7 +1138,7 @@ fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
         bail!("{}\n", error_msg);
     }
 
-    Ok(sysconfig_paths.remove(0))
+    Ok(Some(sysconfig_paths.remove(0)))
 }
 
 /// Finds `_sysconfigdata*.py` files for detected Python interpreters.
@@ -1167,8 +1176,16 @@ fn find_sysconfigdata(cross: &CrossCompileConfig) -> Result<PathBuf> {
 /// ```
 ///
 /// [1]: https://github.com/python/cpython/blob/3.5/Lib/sysconfig.py#L389
+///
+/// Returns an empty vector when the target Python library directory
+/// is not set via `PYO3_CROSS_LIB_DIR`.
 pub fn find_all_sysconfigdata(cross: &CrossCompileConfig) -> Vec<PathBuf> {
-    let sysconfig_paths = search_lib_dir(&cross.lib_dir, cross);
+    let sysconfig_paths = if let Some(lib_dir) = cross.lib_dir.as_ref() {
+        search_lib_dir(lib_dir, cross)
+    } else {
+        return Vec::new();
+    };
+
     let sysconfig_name = env_var("_PYTHON_SYSCONFIGDATA_NAME");
     let mut sysconfig_paths = sysconfig_paths
         .iter()
@@ -1267,11 +1284,19 @@ fn search_lib_dir(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Vec<Pat
 /// first find sysconfigdata file which follows the pattern [`_sysconfigdata_{abi}_{platform}_{multiarch}`][1]
 ///
 /// [1]: https://github.com/python/cpython/blob/3.8/Lib/sysconfig.py#L348
+///
+/// Returns `None` when the target Python library directory is not set.
 fn cross_compile_from_sysconfigdata(
-    cross_compile_config: CrossCompileConfig,
-) -> Result<InterpreterConfig> {
-    let sysconfigdata_path = find_sysconfigdata(&cross_compile_config)?;
-    InterpreterConfig::from_sysconfigdata(&parse_sysconfigdata(sysconfigdata_path)?)
+    cross_compile_config: &CrossCompileConfig,
+) -> Result<Option<InterpreterConfig>> {
+    if let Some(path) = find_sysconfigdata(cross_compile_config)? {
+        let data = parse_sysconfigdata(path)?;
+        let config = InterpreterConfig::from_sysconfigdata(&data)?;
+
+        Ok(Some(config))
+    } else {
+        Ok(None)
+    }
 }
 
 fn windows_hardcoded_cross_compile(
@@ -1289,7 +1314,7 @@ fn windows_hardcoded_cross_compile(
     let implementation = PythonImplementation::CPython;
     let mingw = cross_compile_config.target.environment == Environment::Gnu;
 
-    let lib_dir = Some(cross_compile_config.lib_dir_string());
+    let lib_dir = cross_compile_config.lib_dir_string();
 
     Ok(InterpreterConfig {
         implementation,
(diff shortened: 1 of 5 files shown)


pyo3-build-config/src/impl_.rs
```diff
@@ -20,7 +20,7 @@ use crate::{
 /// Minimum Python version PyO3 supports.
 const MINIMUM_SUPPORTED_VERSION: PythonVersion = PythonVersion { major: 3, minor: 7 };
 /// Maximum Python version that can be used as minimum required Python version with abi3.
-const ABI3_MAX_MINOR: u8 = 9;
+const ABI3_MAX_MINOR: u8 = 10;
```
adamreichold, line 23: Isn't this a bug in the released implementation? Should this become a separate PR with a change log entry or is it inconsequential?

pyo3-build-config/src/impl_.rs
```diff
@@ -677,32 +787,22 @@ pub fn cross_compiling(
     target_os: &str,
 ) -> Result<Option<CrossCompileConfig>> {
     let env_vars = cross_compile_env_vars();
+    let target_info = TargetInfo::from_triple(target_arch, target_vendor, target_os, None);
 
-    let target_info = TargetInfo {
```
adamreichold, line 681: I think the intent of open-coding the construction of the struct was to explicitly avoid being able to mess up the positional arguments to a function like `TargetInfo::from_triple`.

pyo3-build-config/src/lib.rs
```diff
@@ -70,7 +71,7 @@ pub fn get() -> &'static InterpreterConfig {
             InterpreterConfig::from_reader(Cursor::new(CONFIG_FILE))
         } else if !ABI3_CONFIG.is_empty() {
             Ok(abi3_config())
-        } else if impl_::cross_compile_env_vars().any() {
+        } else if Path::new(DEFAULT_CROSS_COMPILE_CONFIG_PATH).exists() {
```
adamreichold, line 74: This does make sense to me, but I am wary whether this would pick up stale configuration files when the environment variables change and a project is rebuilt?

pyo3-build-config/src/impl_.rs
davidhewitt: Yes, I think we should split this out & aim to include in next patch release.

pyo3-build-config/src/impl_.rs
ravenexp: I've added an optional fourth field and it was getting really noisy syntax-wise.
`from_triple()` is a private method that is only ever called by unit test code.
It is also called from a public method that PyO3 itself does not seem to use...

pyo3-build-config/src/impl_.rs
```diff
@@ -1078,33 +1188,56 @@ fn search_lib_dir(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Vec<Pat
...
+    let version = cross_compile_config
+        .version
+        .or_else(get_abi3_version)
+        .ok_or(
+            "PYO3_CROSS_PYTHON_VERSION or an abi3-py3* feature must be specified \
+            when cross-compiling and PYO3_CROSS_LIB_DIR is not set.",
+        )?;
 
     let abi3 = is_abi3();
     let implementation = PythonImplementation::CPython;
```
davidhewitt, line 1216: While you're modifying this stuff, it might be nice to have a `PYO3_CROSS_PYTHON_IMPLEMENTATION` to choose between PyPy and CPython.

pyo3-build-config/src/lib.rs
ravenexp: This is a tricky question. I think this file is created by the `pyo3-ffi` build script, which should re-run when the rustc target or the PyO3 env vars change, but I'm not 100% sure.

`cargo build --target X` worked for me without running `cargo clean` before changing `X`.
Perhaps there is a better way to test this.

@davidhewitt What do you think about this approach to detecting whether we are really cross-compiling?

davidhewitt (commented):
Thanks; to check I understand the idea, it's that when `PYO3_CROSS_LIB_DIR` is not set then (for unix extension modules at least) we can infer all we need from the abi3 and version settings?

Also, a fair number of functions have changed to be `pub fn`. Was that necessary?

pyo3-build-config/src/impl_.rs
ravenexp: I think it is a good idea. I'm not sure that the current PyPy implementation supports `abi3` though, which will limit its usability in many cases.

I'll add `PYO3_CROSS_PYTHON_IMPLEMENTATION` in the next revision.

.gitignore
```diff
@@ -21,3 +21,4 @@ extensions/stamps/
 pip-wheel-metadata
 valgrind-python.supp
 *.pyd
+/tags
```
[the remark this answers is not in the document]
ravenexp: I'll remove it then. I didn't know about global git excludes, thanks!

pyo3-build-config/src/impl_.rs
davidhewitt: Right, but if they try to compile for Python 3.10 with the `abi3-py310` feature then I think we'll accidentally be using the 3.9 abi, thus missing out on newer features.

pyo3-build-config/src/impl_.rs
davidhewitt: I think there's compromise either way; I'm fine with `from_triple` like this.

ravenexp: I'll probably revert this because I'll be resubmitting with many updates anyways...

davidhewitt:
👍 makes sense, yes I think several commits (or even PRs) will be needed to get to the final goal for this one. These build scripts are quite fiddly, there's so many ways to distribute Python leading to lots of edge cases to handle!

ravenexp:
My current plan is to submit a new PR with `ABI3_MAX_MINOR` bug fix only, and convert the current PR into 2+ commits on top of it. I've added a lot of new code that doesn't have test coverage, so it needs to be addressed too...

It it OK to rebase and force push pull requests for this project, or do you prefer that I use merges and fixup commits, so that you can squash them yourself before the final merge?

davidhewitt:
Rebase & force push is fine! Thanks 👍

pyo3-build-config/src/lib.rs
adamreichold: > which should re-run when the rustc target or the PyO3 env vars change, but I'm not 100% sure.

I was thinking mainly about the case when switching from a cross build to a non-cross build as the build script probably does not get out of its way to delete a cross compilation config written by a prior run.

> cargo build --target X worked for me without running cargo clean before changing X. 

I would expect this to work as the file name is the same and the config would be overwritten.

I think https://github.com/PyO3/pyo3/pull/2092 would actually be a nice solution to this as the `DEP_*` environment variables are basically meant for this. But I guess we want to keep the text files around in any case?

davidhewitt: Yes the `DEP_` environment variables sadly don't work for the proc macro crate so we need this backup mechanism anyway.

ravenexp: > > which should re-run when the rustc target or the PyO3 env vars change, but I'm not 100% sure.
> 
> I was thinking mainly about the case when switching from a cross build to a non-cross build as the build script probably does not get out of its way to delete a cross compilation config written by a prior run.

But, isn't `pyo3-cross-compile-config.txt` created in the directory from the `OUT_DIR` Cargo variable, which should be pointing to somewhere in `target/<triple>/deps`? I think config file path should not collide at least as far as `--target` arguments are concerned.

ravenexp:
I'm resubmitting PR this as a set of smaller patches to get continuous feedback from the CI.

It's still a work in progress, but feel free to review the commits I've added so far. I'm highly confident in these, so if changes are requested the sooner I learn about them the better.

pyo3-build-config/src/impl_.rs
```diff
@@ -604,15 +627,43 @@ impl FromStr for PythonImplementation {
...
+///
+/// Must be called from a crate PyO3 build script.
+pub fn is_linking_libpython() -> bool {
+    let target = TargetInfo::from_cargo_env().expect("must be called from a build script");
+
+    target.os == "windows" || target.os == "android" || !is_extension_module()
+}
+
 #[derive(Debug, PartialEq)]
 struct TargetInfo {
```
messense, line 663: This is becoming much more like [target_lexicon::Triple](https://docs.rs/target-lexicon/latest/target_lexicon/struct.Triple.html), can we just wrap it instead?

ravenexp: This is doable, if everyone is OK with adding a new dependency to every build script.
`pyo3-build-config` already re-implements parts of `anyhow` crate to reduce the dependency footprint (I assume).

pyo3-build-config/src/impl_.rs
```diff
@@ -654,6 +723,32 @@ impl TargetInfo {
...
+
+        // Cross compiling if not compiler compatible
+        !host_target_compatible
+    }
+
+    fn is_windows(&self) -> bool {
+        self.os == "windows"
+    }
+
+    fn is_windows_mingw(&self) -> bool {
```
messense, line 749: Why not `is_windows_gnu`?

ravenexp: That's the convention the existing code uses:

https://github.com/PyO3/pyo3/blob/9f60b673b8a41dbd6b6ee5e77c34e5d9b750fd7b/pyo3-build-config/src/impl_.rs#L209-L211

https://github.com/PyO3/pyo3/blob/9f60b673b8a41dbd6b6ee5e77c34e5d9b750fd7b/pyo3-build-config/src/impl_.rs#L1370-L1380

pyo3-build-config/src/impl_.rs
```diff
@@ -1078,33 +1188,56 @@ fn search_lib_dir(path: impl AsRef<Path>, cross: &CrossCompileConfig) -> Vec<Pat
...
+        .ok_or(
+            "PYO3_CROSS_PYTHON_VERSION or an abi3-py3* feature must be specified \
+            when cross-compiling and PYO3_CROSS_LIB_DIR is not set.",
+        )?;
 
     let abi3 = is_abi3();
     let implementation = PythonImplementation::CPython;
 
+    let target = &cross_compile_config.target_info;
+    let lib_name = if target.is_windows() {
```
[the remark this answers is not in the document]
ravenexp: Fixed in the current revision.

pyo3-build-config/src/lib.rs
ravenexp: Actually, I was completely wrong about `pyo3-cross-compile-config.txt` location. It is created in the host build script directory because `pyo3-build-config` is always compiled for the host arch.

I'm using a different approach in https://github.com/PyO3/pyo3/pull/2241/commits/fd6a68e6798c31ee74563f6d29d6560ce38d94f3.

ravenexp:
I've just pushed version 2 of this PR, split into 5 smaller commits this time.
User documentation and the ChangeLog entry are still missing,
but I'll add them once all remaining code issues are solved.

Unresolved questions from the previous rounds of review (shortlist):
- Replacing ad-hoc `TargetInfo` with [`target_lexicon::Triple`](https://docs.rs/target-lexicon/latest/target_lexicon/struct.Triple.html). If the answer is yes, seems easier to address it in this PR.
- Adding `PYO3_CROSS_PYTHON_IMPLEMENTATION` environment variable. This can be addressed in a different PR because it involves writing new user documentation and adding a ChangeLog entry.
- Cross-compilation mode detection. See https://github.com/PyO3/pyo3/pull/2241#discussion_r832297622 and the later discussion. I *think* I've solved it in the current revision, but it's better reviewed once again.

pyo3-build-config/src/lib.rs
adamreichold: This seems like a nice clean-up to me! Thanks!

adamreichold:
> Replacing ad-hoc TargetInfo with [target_lexicon::Triple](https://docs.rs/target-lexicon/latest/target_lexicon/struct.Triple.html). If the answer is yes, seems easier to address it in this PR.

In my experience, build scripts tend to become choke points in parallel builds which is why I would be wary of adding more dependencies to build scripts. That said, `target-lexicon` seems lightweight enough and seems to have good provenance with the Bytecode Alliance.

adamreichold:
> Adding PYO3_CROSS_PYTHON_IMPLEMENTATION environment variable. This can be addressed in a different PR because it involves writing new user documentation and adding a ChangeLog entry.

More smaller PR are preferable IMHO.

ravenexp:
Thanks!

> I had been wondering whether we would need to use `target_lexicon`, it would definitely be worth having a go in a separate commit so we can see how much it helps +1

I'll add this commit on top to evaluate the extent of the required changes, but I'd like to fold it in the first commit for the final revision to reduce useless code churn within the same PR.

ravenexp:
> In my experience, build scripts tend to become choke points in parallel builds which is why I would be wary of adding more dependencies to build scripts. That said, `target-lexicon` seems lightweight enough and seems to have good provenance with the Bytecode Alliance.

The issue with `target-lexicon` is that it itself has a quite extensive build script, which among other things tries to run `rustc` to detect its version. Unfortunately, it can't be disabled via crate features.

davidhewitt:
Right, yes I think that would make it a lot less lightweight than we were hoping. I guess you decided not to try using it as a result?

Anything else to consider before merging this?

messense (approved):

ravenexp:
> Right, yes I think that would make it a lot less lightweight than we were hoping. I guess you decided not to try using it as a result?

I can still try to add it as an experiment. I haven't actually checked if it has any measurable impact on the build times.

> Anything else to consider before merging this?

Yes, updating user documentation and adding the changelog entry are still a TODO.

ravenexp:
> > Right, yes I think that would make it a lot less lightweight than we were hoping. I guess you decided not to try using it as a result?
> 
> I can still try to add it as an experiment. I haven't actually checked if it has any measurable impact on the build times.

I've added https://github.com/PyO3/pyo3/pull/2241/commits/65b382e55248be083ec15bb6ef489ccb1e007a24 on top of the current branch to evaluate the `target-lexicon::Triple` integration impact.

I think there's a demonstrable benefit from the code quality standpoint.

The downsides:

1. Build times: A simple PyO3 extension module build time increased (on average) from 15.05s to 15.25s (+0.2s) because of the new `target-lexicon` dependency and its build script.
2. I had to break the public crate API in `pyo3_build_config::cross_compile()` function because there's just not enough information to make sound decisions based on the existing function arguments.

In my opinion the build time cost is acceptable, and as for `cross_compile()` I'm not sure if it has any external users.
PyO3 and maturin themselves do not use it. Maybe it's better to deprecate it altogether.

messense:
> and as for cross_compile() I'm not sure if it has any external users.
PyO3 and maturin themselves do not use it. Maybe it's better to deprecate it altogether.

It was supposed to be used to replace the duplicated code in maturin, see https://github.com/PyO3/maturin/pull/718

ravenexp:
> > and as for cross_compile() I'm not sure if it has any external users.
> > PyO3 and maturin themselves do not use it. Maybe it's better to deprecate it altogether.
> 
> It was supposed to be used to replace the duplicated code in maturin, see [PyO3/maturin#718](https://github.com/PyO3/maturin/pull/718)

Can this code be adapted to the new `cross_compile()` signature? As I can see maturing already uses `Triple` internally.

messense:
> Can this code be adapted to the new cross_compile() signature?

I think it's ok.

> This would delay the feature merge until the next major PyO3 release, though...

I think it's fine to break that api since it was tailored for maturin anyway.

davidhewitt:
> I think it's fine to break that api since it was tailored for maturin anyway.

Even if someone else using it is unlikely, I would still prefer not change `cross_compiling` signature until the next breaking release. I think anyone who _is_ using it would have good reason to be upset if we broke them in a semver-noncompliant way.

I'm thinking to put out a 0.16.3 release in the next few days (just one last thing I'd like to add to it), and then I'd be happy to start merging breaking changes because I've got ideas anyway 😄

ravenexp:
> Even if someone else using it is unlikely, I would still prefer not change `cross_compiling` signature until the next breaking release. I think anyone who _is_ using it would have good reason to be upset if we broke them in a semver-noncompliant way.

I can try to keep this function as is and try to create a somewhat correct implementation for it until it's deprecated/removed in v0.17.
As a substitute, I'll add a new public API function that uses `Triple` arguments.

ravenexp:
I've pushed the `target-lexicon` patch version 2 in https://github.com/PyO3/pyo3/pull/2241/commits/fe9c1602345c7f1b7f7ff6bddb6b0474491b604b

Here I'm trying to keep the existing `cross_compile()` implementation afloat and introduce a new public function `cross_compile_from_to()` that uses `Triple` arguments directly.

If everyone is OK with moving on with `target_lexicon::Triple`, I'd rather split it off in a separate PR with its own ChangeLog entry and then rebase the current PR on top of it.

davidhewitt:
Sounds good to me 👍

ravenexp:
Rebased on main and reopened.

pyo3-build-config/src/impl_.rs
ravenexp: Postponed and will be addressed in a separate PR.

messense (approved):

davidhewitt (approved):
Thanks, looks good to me. Thanks all for reviewing on this; I didn't manage to find any time to look for a number of days.

PR #2242, opened 2022-03-23
Fix `abi3-py310` feature implementation: use Python 3.10 ABI when available.

ravenexp:
Split from #2241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix `abi3-py310` feature implementation: use Python 3.10 ABI when available. [#2242](https://github.com/PyO3/pyo3/pull/2242)
```
adamreichold, line 17: Maybe specific the symptoms of the error, e.g. "instead of silently falling back to the 3.9 ABI"?

ravenexp: Updated.

adamreichold: Thanks!

adamreichold (approved):

PR #2243, opened 2022-03-23
pyo3-macros-backend: Replace `pyo3-build-config` with `abi3` feature

ravenexp:
Python 3.6 and older are not supported by the current PyO3 version,
so the removed interpreter version check was a no-op.

`pyo3_build_config::get()` attempts to read a config file from disk
when PyO3 is cross-compiling, which is probably bad for rust-analyzer
and other IDEs that attempt to sandbox the proc macro code.


davidhewitt (approved):
Haha this is an excellent observation!

It used to not be possible when we also needed to support Python 3.6 (which didn't have fastcall). However, for now this method works well and I agree is more reliable.

PR #2244, opened 2022-03-23
Add more lints

mejrs:
This adds the `elided_lifetimes_in_paths`, `unused_lifetimes` and `rust_2021_prelude_collisions` lints.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -15,4 +15,7 @@ rustflags = [
     "-Dclippy::todo",
     "-Dclippy::unnecessary_wraps",
     "-Dclippy::useless_transmute",
+    "-Delided_lifetimes_in_paths",
+    "-Dunused_lifetimes",
+    "-Drust_2021_prelude_collisions"
 ]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,7 @@
         rustdoc::bare_urls
     )
 )]
+#![warn(elided_lifetimes_in_paths, unused_lifetimes)]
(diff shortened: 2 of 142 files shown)


.cargo/config
```diff
@@ -15,4 +15,7 @@ rustflags = [
     "-Dclippy::todo",
     "-Dclippy::unnecessary_wraps",
     "-Dclippy::useless_transmute",
+    "-Delided_lifetimes_in_paths",
+    "-Dunused_lifetimes",
+    "-Drust_2021_prelude_collisions"
```
davidhewitt, line 20: Because these are `rustc` lints, it might be nice to have them warn by default even when not using clips. Especially the `elided_lifetimes_in_paths` - I'm pretty sure I'll trip up on this a lot!

src/lib.rs
```diff
@@ -226,7 +234,7 @@
 //!         let code = "os.getenv('USER') or os.getenv('USERNAME') or 'Unknown'";
 //!         let user: String = py.eval(code, None, Some(&locals))?.extract()?;
 //!
-//!         println!("Hello {}, I'm Python {}", user, version);
+//!         println!("Hello {}, I'm Python<'_> {}", user, version);
```
davidhewitt, line 237: 😉 

```suggestion
//!         println!("Hello {}, I'm Python {}", user, version);
```

mejrs:
Thanks, done :)

davidhewitt (approved):
👍

PR #2245, opened 2022-03-23
Implement PEP 489 - Multi-phase extension module initialization

messense:
https://peps.python.org/pep-0489/

This is very much WIP, opening for discussion, previously: https://github.com/PyO3/pyo3/discussions/1982


src/impl_/pymodule.rs
```diff
@@ -52,10 +69,16 @@ impl ModuleDef {
...
+    /// Return module def
+    pub fn module_def(&'static self) -> *mut ffi::PyModuleDef {
+        self.ffi_def.get()
+    }
     /// Builds a module using user given initializer. Used for [`#[pymodule]`][crate::pymodule].
     pub fn make_module(&'static self, py: Python) -> PyResult<PyObject> {
         let module = unsafe {
-            Py::<PyModule>::from_owned_ptr_or_err(py, ffi::PyModule_Create(self.ffi_def.get()))?
+            let mod_def = self.ffi_def.get();
+            (*mod_def).m_slots = std::ptr::null_mut();
```
messense, line 80: `PyModule_Create` does not working with `m_slots` set:

```
SystemError('module foobar_module: PyModule_Create is incompatible with m_slots'), traceback: None }
```

davidhewitt: What would it take to change `wrap_pymodule` to use PEP 489?

messense: Using [PyModule_FromDefAndSpec](https://docs.python.org/3/c-api/module.html#c.PyModule_FromDefAndSpec) I guess, but I don't quite understand the `spec` argument yet.

messense: Unfortunately there isn't any example of `PyModule_FromDefAndSpec` usage in the docs: https://docs.python.org/3/c-api/module.html#low-level-module-creation-functions

cc @birkenfeld do you how these low level module creation functions are supposed to be used? I don't get how to obtain a `ModuleSpec` object for passing into `PyModule_FromDefAndSpec`.

birkenfeld: Not off the top of my head. Maybe they are supposed to be statically filled like PyTypeObjects used to be?

messense: It's not easy to be statically filled, [ModuleSpec](https://docs.python.org/3/library/importlib.html#importlib.machinery.ModuleSpec) isn't a C struct but a Python class from `importlib` and it requires a name and a [loader](https://docs.python.org/3/glossary.html#term-loader) to construct.

PEP 489 also mentioned [multiple modules in one library](https://peps.python.org/pep-0489/#multiple-modules-in-one-library), but it's doing that in Python instead of using C API.

davidhewitt: I think I saw somewhere that the builtin `sqlite3` module now supports PEP 489. Maybe this offers some clues?

https://github.com/python/cpython/blob/f66c857572a308822c70fd25e0197b6e0dec6e34/Modules/_sqlite/module.c#L762

messense: Well, that's the same function we use here for the main extension module:
https://github.com/PyO3/pyo3/pull/2245/files#diff-912566621a79aeb2b0463cd8d22a412c3166600d99e1cc7f164aef67492c2429R85, it returns a module definition object instead of a module: https://docs.python.org/3/c-api/module.html#c.PyModuleDef_Init

But `wrap_pymodule!` is more about making submodules, unable to use `PyModule_Create` makes it hard to implement.

davidhewitt: Ah, I see. Maybe we need to think about this in combination with the syntax suggested in #694, and get away from using `wrap_pymodule!`.

messense: BTW `wrap_pymodule!` is also used a lot in tests, without it it's also hard to write simple module related tests without setting up proper extension modules.

messense: Opened a discussion: https://discuss.python.org/t/how-to-use-pymodule-fromdefandspec-c-api-to-construct-a-module/14581

messense:
Looks like PyPy doesn't support PEP 489.

```
  = note: /usr/bin/ld: /home/runner/work/pyo3/pyo3/target/x86_64-unknown-linux-gnu/debug/deps/libpyo3-5c3820f48b335551.rlib(pyo3-5c3820f48b335551.pyo3.839e673f-cgu.2.rcgu.o): in function `pyo3_ffi::modsupport::PyModule_FromDefAndSpec':
          /home/runner/work/pyo3/pyo3/pyo3-ffi/src/modsupport.rs:146: undefined reference to `PyModule_FromDefAndSpec2'
          /usr/bin/ld: /home/runner/work/pyo3/pyo3/target/x86_64-unknown-linux-gnu/debug/deps/libpyo3-5c3820f48b335551.rlib(pyo3-5c3820f48b335551.pyo3.839e673f-cgu.5.rcgu.o): in function `pyo3::impl_::pymodule::ModuleDef::make_module':
          /home/runner/work/pyo3/pyo3/src/impl_/pymodule.rs:80: undefined reference to `PyModule_ExecDef'
```

src/impl_/pymodule.rs
messense: Sovled by https://discuss.python.org/t/how-to-use-pymodule-fromdefandspec-c-api-to-construct-a-module/14581/2?u=messense

davidhewitt:
:(

I guess we have to make a choice between making a new API that's not supported on PyPy, or just using different behaviour on CPython and PyPy. I think a separate API is probably better?

messense:
Maybe we should feature-gate it? See also https://github.com/cython/cython/issues/2343

davidhewitt:
It would be interesting to explore. There's also still a possibility to have a `#[pymodule(pep489)]` or similar.

adamreichold:
> I guess we have to make a choice between making a new API that's not supported on PyPy, or just using different behaviour on CPython and PyPy. I think a separate API is probably better?

But isn't "same interface but slightly incompatible behaviour" (w.r.t. CPython) already something that PyPy users expect? I fear that having CPython-only API will only end up with extensions having feature-gated PyPy-compatible code paths any way, so I would prefer it if the thankless task of  trying to paper this over was centralized in PyO3 (were we could also move the ecosystem over if/when PyPy adds support for this.)

davidhewitt:
I think it might depend how different the PyPy behaviour is. I do agree if we can solve the problem for users transparently, that would be nicest.

pyo3-macros-backend/src/module.rs
```diff
@@ -83,10 +86,60 @@ pub fn pymodule_impl(
             unsafe { #module_def_name.module_init() }
         }
 
+        #[cfg(any(not(feature = "pep489"), PyPy))]
         #[doc(hidden)]
         #visibility static #module_def_name: #krate::impl_::pymodule::ModuleDef = unsafe {
             #krate::impl_::pymodule::ModuleDef::new(concat!(stringify!(#name), "\0"), #doc, #krate::impl_::pymodule::ModuleInitializer(#fnname))
         };
+
+        #[cfg(all(feature = "pep489", not(PyPy)))]
```
messense, line 95: Hmm, looks like after macro expansion the code with this `cfg` are disappeared... 🧐

davidhewitt:
Superseded by #5142

PR #2249, opened 2022-03-25
ci: avoid race between bench and guide

davidhewitt:
The `main` guide build fails a lot, and I think it's because while the guide is building the benchmarks run and push their results to the `gh-pages` branch. I'm hoping by making the benchmark jobs run after the guide build, it'll fix this race.


.github/workflows/gh-pages.yml
```diff
@@ -103,3 +104,94 @@ jobs:
           publish_dir: ./public/
           full_commit_message: "Release ${{ needs.build.outputs.tag_name }}"
           keep_files: true
+
+  cargo-benchmark:
+    if: ${{ github.ref_name == 'main' }}
+    name: Cargo benchmark
+    needs: build
```
adamreichold, line 111: Isn't this called `guide-build` now?

davidhewitt: Yes, good spot.

davidhewitt:
Not sure why the actions wouldn't start until I force-toggled the PR.

adamreichold:
> Not sure why the actions wouldn't start until I force-toggled the PR.

GitHub still seems to have a bit of trouble with its mySQL1 cluster, c.f. https://www.githubstatus.com/#past-incidents and https://github.blog/2022-03-23-an-update-on-recent-service-disruptions/

messense (approved):

PR #2251, opened 2022-03-27
Add rustimport to tools section in readme

mityax:
I created a python module called "[rustimport](https://github.com/mityax/rustimport)" which simplifies development workflow when creating Python extensions with pyo3.

This PR adds it to the list in the readme, so others might profit from it too :)


davidhewitt:
👍 nice tool, thank you for building & sharing!

PR #2252, opened 2022-03-27
Show xtask stdout and stderr when running in GitHub Actions

messense:
This should make it easier to verify command execution result manually when they succeeded. The output should be grouped and collapsed by default so it's not too verbose.

### DIFF
--- a/xtask/src/cli.rs
+++ b/xtask/src/cli.rs
@@ -117,24 +117,35 @@ impl Subcommand {
 }
 
 pub fn run(command: &mut Command) -> Result<()> {
-    println!("Running: {}", format_command(command));
+    let command_str = format_command(command);
+    println!("Running: {}", command_str);
 
     let output = command
         .stdout(Stdio::piped())
         .stderr(Stdio::piped())
         .spawn()?
         .wait_with_output()?;
+    let out = String::from_utf8_lossy(&output.stdout);
+    let err = String::from_utf8_lossy(&output.stderr);
+
+    let github_actions = std::env::var_os("GITHUB_ACTIONS").is_some();
+    if github_actions {
+        println!("::group::{}", command_str);
+        println!("{}", out);
+        println!("{}", err);
+        println!("::endgroup::")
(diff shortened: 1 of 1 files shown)


xtask/src/cli.rs
```diff
@@ -117,24 +117,35 @@ impl Subcommand {
 }
 
 pub fn run(command: &mut Command) -> Result<()> {
-    println!("Running: {}", format_command(command));
+    let command_str = format_command(command);
+    println!("Running: {}", command_str);
 
     let output = command
         .stdout(Stdio::piped())
```
davidhewitt, line 124: Oh, I missed that we changed it to capture output. @mejrs, @messense how do you feel if we change this so that we just inherit output handles instead? I think it's nice to see what's happening in realtime; `nox` for example doesn't squash output by default.

messense: > how do you feel if we change this so that we just inherit output handles instead?

That's also fine.

davidhewitt (approved):
Looks great to me!

xtask/src/cli.rs
mejrs: >  how do you feel if we change this so that we just inherit output handles instead?

I wasn't really thinking that much when I wrote this, so a better approach would be welcome 👍

xtask/src/cli.rs
```diff
@@ -117,26 +117,29 @@ impl Subcommand {
 }
 
 pub fn run(command: &mut Command) -> Result<()> {
-    println!("Running: {}", format_command(command));
+    let command_str = format_command(command);
+    let github_actions = std::env::var_os("GITHUB_ACTIONS").is_some();
```
mejrs, line 121: It might be better to do this as part of parsing `Command`, especially if other parts of the program also need to know in which environment they're running.

mejrs: Oh I see you already merged 😅 It's not a big deal, but we might want to change it at some point.

PR #2253, opened 2022-03-28
pyo3-build-config: Replace `TargetInfo` with `target_lexicon::Triple`

ravenexp:
Add a new public crate function `cross_compile_from_to()` using
`target_lexicon::Triple` arguments instead of plain strings
used in `cross_compile()`.

Attempt to extract common code patterns into methods and standalone
helper functions. Add docstrings to private items.
Make some of the new helper functions public within the PyO3 crate
and reuse them in the build scripts.

No user visible functionality change.

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -12,6 +12,10 @@ use std::{
     str::FromStr,
 };
 
+pub use target_lexicon::Triple;
+
+use target_lexicon::{Architecture, BinaryFormat, Environment, OperatingSystem, Vendor};
+
 use crate::{
     bail, ensure,
     errors::{Context, Error, Result},
@@ -38,6 +42,16 @@ pub fn env_var(var: &str) -> Option<OsString> {
         println!("cargo:rerun-if-env-changed={}", var);
     }
     env::var_os(var)
+}
+
+/// Gets the compilation target triple from environment variables set by Cargo.
+///
+/// Must be called from a crate build script.
+pub fn target_triple_from_env() -> Triple {
+    env::var("TARGET")
+        .expect("target_triple_from_env() must be called from a build script")
+        .parse()
+        .expect("Unrecognized TARGET environment variable value")
 }
 
 /// Configuration needed by PyO3 to build for the correct Python implementation.
@@ -524,6 +538,29 @@ print("mingw", get_platform().startswith("mingw"))
             envs,
         )
     }
+
+    /// Lowers the configured version to the abi3 version, if set.
+    fn fixup_for_abi3_version(&mut self, abi3_version: Option<PythonVersion>) -> Result<()> {
+        // PyPy doesn't support abi3; don't adjust the version
+        if self.implementation.is_pypy() {
+            return Ok(());
+        }
+
+        if let Some(version) = abi3_version {
+            ensure!(
+                version <= self.version,
+                "cannot set a minimum Python version {} higher than the interpreter version {} \
+                (the minimum Python version is implied by the abi3-py3{} feature)",
+                version,
+                self.version,
+                version.minor,
+            );
+
+            self.version = version;
+        }
+
+        Ok(())
+    }
 }
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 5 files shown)


pyo3-build-config/src/impl_.rs
```diff
@@ -733,39 +828,74 @@ pub(crate) fn cross_compile_env_vars() -> CrossCompileEnvVars {
 ///   `PYO3_CROSS_LIB_DIR`.
 ///
 /// See the [PyO3 User Guide](https://pyo3.rs/) for more info on cross-compiling.
+///
+/// TODO: Deprecate in v0.17 favor of `cross_compiling_from_to()`.
```
adamreichold, line 832: Can't we deprecate this right now (as the alternative API is added) or would adding a deprecation be a breaking change?

(I also think this should get a change log entry as it modifies the public API of `pyo3-build-config` even most users will not interact with it directly.)

ravenexp: > Can't we deprecate this right now (as the alternative API is added) or would adding a deprecation be a breaking change?

It depends on whether this commit gets into the next point release or it is first released in v0.17.
I can replace `TODO:` with `#[deprecated(since="0.17.0", note="please use cross_compile_from_to() instead")]`.

I'm adding a ChangeLog entry in the next push.

ravenexp: I've added a deprecation attribute and a ChangeLog entry.

pyo3-build-config/src/impl_.rs
```diff
@@ -733,39 +828,76 @@ pub(crate) fn cross_compile_env_vars() -> CrossCompileEnvVars {
...
+    // for the current `CrossCompileConfig` implementation.
+    let environment = match operating_system {
+        OperatingSystem::Windows => Environment::Msvc,
+        _ => Environment::Gnu,
     };
 
-    if !env_vars.any() && is_not_cross_compiling(host, &target_info) {
-        return Ok(None);
-    }
+    // FIXME: This field is currently unused.
```
messense, line 854: I guess we don't need to address these `FIXME`s right now since we're going to remove `cross_compiling` in a new minor version.

messense (approved):

pyo3-build-config/src/impl_.rs
ravenexp: Oh, certainly. These FIXMEs will be removed with `cross_compiling()` in v0.17.
Perhaps I should've used `WONTFIX:` instead of `FIXME:` :grin:.

pyo3-ffi/build.rs
```diff
@@ -92,7 +89,7 @@ fn configure_pyo3() -> Result<()> {
     ensure_python_version(&interpreter_config)?;
     ensure_target_pointer_width(&interpreter_config)?;
 
-    if !interpreter_config.suppress_build_script_link_lines {
+    if is_linking_libpython() && !interpreter_config.suppress_build_script_link_lines {
```
adamreichold, line 92: This seems to have the effect that `is_linking_libpython()` now also controls whether ` interpreter_config.to_cargo_dep_env()?;` is called or not. Is that intentional? Shouldn't we serialize the configuration into those variables in any case?

pyo3-build-config/src/impl_.rs
```diff
@@ -1924,6 +2058,51 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_not_cross_compiling_from_to() {
```
adamreichold, line 2062: Maybe also add a case where it should cross-compile?

pyo3-build-config/src/impl_.rs
```diff
@@ -709,6 +766,44 @@ impl CrossCompileEnvVars {
...
+
+    /// Converts the stored `PYO3_CROSS_LIB_DIR` variable value (if any)
+    /// into a `PathBuf` instance.
+    ///
+    /// Ensures that the path is a valid UTF-8 string.
+    fn lib_dir_path(&self) -> Result<PathBuf> {
+        let lib_dir = self.pyo3_cross_lib_dir.as_ref().map(PathBuf::from);
+
+        if let Some(dir) = lib_dir.as_ref() {
+            if dir.to_str().is_none() {
```
adamreichold, line 797: This look like the `ensure!` macro could improve readability.

pyo3-build-config/src/impl_.rs
```diff
@@ -1303,7 +1440,7 @@ fn conda_env_interpreter(conda_prefix: &OsStr, windows: bool) -> PathBuf {
 
 fn get_env_interpreter() -> Option<PathBuf> {
     match (env_var("VIRTUAL_ENV"), env_var("CONDA_PREFIX")) {
-        // Use cfg rather can CARGO_TARGET_OS because this affects where files are located on the
+        // Use cfg rather can CARGO_CFG_TARGET_OS because this affects where files are located on the
```
adamreichold, line 1443: Typo: "rather ~~can~~than"

pyo3-build-config/src/impl_.rs
ravenexp: That requires setting `PYO3_CROSS_LIB_DIR` env var to something, and I don't know how to do that in a unit test.
https://github.com/PyO3/pyo3/pull/2241 lifts this restriction, so is becomes much easier to add there.

pyo3-ffi/build.rs
ravenexp: Good catch! It's the result of rebasing on top of https://github.com/PyO3/pyo3/pull/2092 and an automatic git merge.

I'll fix this in the new revision.

ravenexp: Fixed in https://github.com/PyO3/pyo3/commit/9b45a19161cd20fac1c83c25897aa9d213244a15

pyo3-build-config/src/impl_.rs
```diff
@@ -1780,6 +1894,26 @@ mod tests {
...
+            pyo3_cross_lib_dir: None,
+            pyo3_cross_python_version: Some("3.9".into()),
+        };
+
+        assert_eq!(
+            env_vars.parse_version().unwrap(),
+            Some(PythonVersion { major: 3, minor: 9 })
+        );
+
+        env_vars.pyo3_cross_python_version = None;
```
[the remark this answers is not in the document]
ravenexp: Applied in https://github.com/PyO3/pyo3/commit/9b45a19161cd20fac1c83c25897aa9d213244a15

pyo3-build-config/src/impl_.rs
```diff
@@ -668,32 +690,67 @@ pub struct CrossCompileConfig {
...
+            Some(CrossCompileConfig {
+                lib_dir,
+                target,
+                version,
+            })
+        } else {
+            None
+        };
+
+        Ok(maybe_config)
```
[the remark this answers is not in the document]
ravenexp: Personal preference.

Changed in https://github.com/PyO3/pyo3/commit/9b45a19161cd20fac1c83c25897aa9d213244a15

pyo3-build-config/src/impl_.rs
ravenexp: Applied in https://github.com/PyO3/pyo3/commit/9b45a19161cd20fac1c83c25897aa9d213244a15

pyo3-build-config/src/impl_.rs
ravenexp: Applied in https://github.com/PyO3/pyo3/commit/9b45a19161cd20fac1c83c25897aa9d213244a15

pyo3-build-config/src/impl_.rs
```diff
@@ -668,32 +690,67 @@ pub struct CrossCompileConfig {
...
+
+        !compatible
+    }
+
+    /// Converts `lib_dir` member field to an UTF-8 string.
+    ///
+    /// The conversion can not fail because `PYO3_CROSS_LIB_DIR` variable
+    /// is ensured contain a valid UTF-8 string.
+    #[allow(clippy::unnecessary_wraps)]
+    fn lib_dir_string(&self) -> Option<String> {
```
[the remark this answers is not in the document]
ravenexp: Removed in https://github.com/PyO3/pyo3/commit/9b45a19161cd20fac1c83c25897aa9d213244a15

adamreichold (approved):

PR #2256, opened 2022-03-28
Some more docs

mejrs:
- Use the new `is` method in some doc examples.
- Clarifies some safety docs


src/marker.rs
```diff
@@ -882,6 +866,29 @@ impl<'py> Python<'py> {
     }
 }
 
+impl<'unbound> Python<'unbound> {
```
mejrs, line 869: I've taken a liking to properly naming lifetimes as of recently. What would you think of renaming the `Python<'py>` lifetime to `Python<'gil>`? It's still short and I think it represents the meaning of the lifetime better.

src/marker.rs
```diff
@@ -882,6 +866,29 @@ impl<'py> Python<'py> {
...
+    /// Unsafely creates a Python token with an unbounded lifetime.
+    ///
+    /// Many of PyO3 APIs use `Python<'_>` as proof that the GIL is held, but this function can be
+    /// used to call them unsafely.
+    ///
+    /// # Safety
+    ///
+    /// - This token and any borrowed Python references derived from it can only be safely used
+    /// whilst the currently executing thread is actually holding the GIL.
```
mejrs, line 878: Re: threads, I was thinking about this:

```rust
fn foo(py: Python<'_>){
    let handle = std::thread::spawn(|| unsafe {
        // call some pyo3 apis
    });
    
    handle.join().unwrap();
}
```

Is this legal?

src/marker.rs
adamreichold: I agree with the sentiment, but I am not sure if the improvement is worth the churn?

src/marker.rs
adamreichold: As `spawn` and `join` are synchronization points, I think this is legal as long as there is absolutely nothing happening in between `spawn` and `join` which could prevent the `join` from happening, i.e. the thread must be scoped.

However, I would suggest not including this example in the documentation as it opens more questions than it answers IMHO and I think the current wording is the right thing to do in almost all of the cases. When someone has to handle a case where this is not so (as in your example), they should be sure enough to figure it out. Everybody is probably just confused by watering down the statement about the current thread.

mejrs: To clarify, the way I worded it would declare that it's not allowed.

src/marker.rs
davidhewitt: It should be a fairly mechanical change if we wanted to update. If we were consistent with always having `'gil` for the band of the GIL lifetime when present, that would also help make it clear when things did not depend on the GIL.

davidhewitt (approved):
Nice improvements, thanks!

Needs formatting fix, else I would have clicked merge ☺️

PR #2257, opened 2022-03-29
Enable `METH_FASTCALL` with abi3 for Python 3.10+

ravenexp:
Following https://github.com/PyO3/pyo3/commit/23e220985af786bda171f7d2b991e47e00d68b01

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -183,7 +183,7 @@ impl SelfType {
 pub enum CallingConvention {
     Noargs,   // METH_NOARGS
     Varargs,  // METH_VARARGS | METH_KEYWORDS
-    Fastcall, // METH_FASTCALL | METH_KEYWORDS (not compatible with `abi3` feature)
+    Fastcall, // METH_FASTCALL | METH_KEYWORDS
     TpNew,    // special convention for tp_new
 }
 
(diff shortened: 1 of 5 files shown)


pyo3-macros-backend/src/method.rs
```diff
@@ -199,8 +199,8 @@ impl CallingConvention {
         } else if accept_kwargs {
             // for functions that accept **kwargs, always prefer varargs
             Self::Varargs
-        } else if cfg!(not(feature = "abi3")) {
-            // Not available in the Stable ABI as of Python 3.10
+        } else if cfg!(not(feature = "abi3-py39")) {
```
birkenfeld, line 202: What about 3.7 and 3.8?

ravenexp: `abi3-py37` and `abi3-py38` imply `abi3-py39`, `abi3-py310` and `abi3`.

`abi3-*` features work backwards compared to `Py_3_*` configs.

ravenexp:
I think the problem here is that while the `METH_FASTCALL` bit flag was stabilized as a part of [PEP 573](https://peps.python.org/pep-0573/) implementation, the corresponding calling convention was not. [_PyCFunctionFast](https://docs.python.org/3/c-api/structures.html#c._PyCFunctionFast) and [_PyCFunctionFastWithKeywords](https://docs.python.org/3/c-api/structures.html#c._PyCFunctionFastWithKeywords) are still not a part of the Stable ABI.

This PR can be dropped for now.

PR #2262, opened 2022-03-31
Backport PyPy 3.9 support for pyo3 0.15

### LINKED ISSUE #2260: Version 0.15.2 with support for Pypy 3.9 and Python 3.6
Hi! First of all, thanks for building and maintaining this project!

The cryptography project, which depends on this PyO3, is currently unable to provide support for Pypy with Python 3.9, because a large portion of their users still use Python 3.6, which was dropped in version 0.16 of PyO3, the same that first includes support for Pypy 3.9.

This is the relevant issue: https://github.com/pyca/cryptography/issues/6924

It has been suggested that a new patch release could be used to solve this issue, for example a version 0.15.2 adding support for Pypy 3.9 while also keeping support for Python 3.6.

Is that feasible? Would a pull request be accepted? If yes, can you give me some pointers or tips so that I can start working on this?

messense:
Fixes #2260

### DIFF
--- a/src/callback.rs
+++ b/src/callback.rs
@@ -244,6 +244,21 @@ where
     let unwind_safe_py = AssertUnwindSafe(pool.python());
     let panic_result = panic::catch_unwind(move || -> PyResult<_> {
         let py = *unwind_safe_py;
+        #[cfg(all(PyPy, not(Py_3_8)))]
+        {
+            const PYPY_GOOD_VERSION: [u8; 3] = [7, 3, 8];
+            let version = py
+                .import("sys")?
+                .getattr("implementation")?
+                .getattr("version")?;
+            if version.lt(crate::types::PyTuple::new(py, &PYPY_GOOD_VERSION))? {
+                let warn = py.import("warnings")?.getattr("warn")?;
+                warn.call1((
+                    "PyPy 3.7 versions older than 7.3.8 are known to have binary \
+                             compatibility issues which may cause segfaults. Please upgrade.",
+                ))?;
+            }
+        }
         body(py)
     });
 
--- a/src/callback.rs
+++ b/src/callback.rs
@@ -251,7 +251,9 @@ where
                 .import("sys")?
                 .getattr("implementation")?
                 .getattr("version")?;
-            if version.lt(crate::types::PyTuple::new(py, &PYPY_GOOD_VERSION))? {
+            if version.compare(crate::types::PyTuple::new(py, &PYPY_GOOD_VERSION))?
+                == std::cmp::Ordering::Less
+            {
                 let warn = py.import("warnings")?.getattr("warn")?;
                 warn.call1((
                     "PyPy 3.7 versions older than 7.3.8 are known to have binary \
--- a/src/callback.rs
+++ b/src/callback.rs
@@ -244,23 +244,6 @@ where
     let unwind_safe_py = AssertUnwindSafe(pool.python());
     let panic_result = panic::catch_unwind(move || -> PyResult<_> {
         let py = *unwind_safe_py;
-        #[cfg(all(PyPy, not(Py_3_8)))]
-        {
-            const PYPY_GOOD_VERSION: [u8; 3] = [7, 3, 8];
-            let version = py
-                .import("sys")?
-                .getattr("implementation")?
-                .getattr("version")?;
-            if version.compare(crate::types::PyTuple::new(py, &PYPY_GOOD_VERSION))?
-                == std::cmp::Ordering::Less
-            {
-                let warn = py.import("warnings")?.getattr("warn")?;
-                warn.call1((
-                    "PyPy 3.7 versions older than 7.3.8 are known to have binary \
-                             compatibility issues which may cause segfaults. Please upgrade.",
-                ))?;
-            }
-        }
         body(py)
     });
 
--- a/src/ffi/datetime.rs
+++ b/src/ffi/datetime.rs
@@ -361,7 +361,7 @@ pub struct PyDateTime_CAPI {
     pub TimeType: *mut PyTypeObject,
     pub DeltaType: *mut PyTypeObject,
     pub TZInfoType: *mut PyTypeObject,
-    #[cfg(all(Py_3_7, any(not(PyPy), Py_3_8)))]
+    #[cfg(Py_3_7)]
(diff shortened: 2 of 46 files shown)


messense:
Looks like we need to raise MSRV to 1.48?

adamreichold:
> Looks like we need to raise MSRV to 1.48?

I think that would not be possible  in the context of #2260 as this is about supporting old platforms like RHEL 6 using rustc 1.41?

src/callback.rs
```diff
@@ -244,6 +244,21 @@ where
...
     let panic_result = panic::catch_unwind(move || -> PyResult<_> {
         let py = *unwind_safe_py;
+        #[cfg(all(PyPy, not(Py_3_8)))]
+        {
+            const PYPY_GOOD_VERSION: [u8; 3] = [7, 3, 8];
+            let version = py
+                .import("sys")?
+                .getattr("implementation")?
+                .getattr("version")?;
+            if version.lt(crate::types::PyTuple::new(py, &PYPY_GOOD_VERSION))? {
```
alex, line 254: This needs to use `.compare()` I think

messense:
I tried to bless the ui tests on macOS, but it still failed on CI.

src/ffi/datetime.rs
```diff
@@ -451,7 +451,7 @@ pub static PyDateTimeAPI: _PyDateTimeAPI_impl = _PyDateTimeAPI_impl {
 ///
 /// The type obtained by dereferencing this object is `&'static PyObject`. This may change in the
 /// future to be a more specific type representing that this is a `datetime.timezone` object.
-#[cfg(all(Py_3_7, any(not(PyPy), Py_3_8)))]
+#[cfg(Py_3_7)]
```
messense, line 454: This seems wrong.

messense:
Hmm, getting CI back to green is not easy.

davidhewitt:
Thanks for working on this! I think you probably need to include the switch from `tox` to `nox` (IIRC this fixes the windows link errors for some reason).

messense:
> Thanks for working on this! I think you probably need to include the switch from `tox` to `nox` (IIRC this fixes the windows link errors for some reason).

Tried [that](https://github.com/messense/pyo3/tree/pypy-0.15-backup) and got [more errors](https://github.com/PyO3/pyo3/actions/runs/2077203342), lots of things need cleanup and fixing.

messense:
I can only work on this very slowly due to the Covid lockdown situation in Shanghai China, so feel free to push to my branch or take over completely if anyone is interested in pushing this forward.

mejrs:
All green 🎉 

Thanks @messense for your work :) I think this is done now?

alex:
🎉 🎉 🎉 

Would it be helpful for us to do a CI run on pyca/cryptography with this branch to verify everything looks good?

mejrs:
I was just about to ask 😄 

Yes, that's a good idea 👍

alex:
Ok testing PR up at https://github.com/pyca/cryptography/pull/7062

alex:
Looks like there's an issue with pypy3 7.3.9, several tests are failing with `SystemError: Function returned a NULL result without setting an exception`

mejrs:
It looks like these tests go through the datetime api 😬 Your CI uses `PyPy 7.3.9`, try pinning to `pypy-3.7-v7.3.7`


This may be a case of https://github.com/PyO3/pyo3/pull/2217

alex:
I don't think we use the pyo3/C-API for handling datetimes directly, AFAIK everywhere we process datetimes we do it like this: https://github.com/pyca/cryptography/blob/main/src/rust/src/x509/common.rs#L663

mejrs:
Uh, strange. I don't think this PR introduces any difference between PyPy 3.7 and PyPy 3.8 ..... 🤔 

@davidhewitt, any ideas? 🤷

alex:
Confirmed that the tests pass on pypy3.7 7.3.7, but I'd like to understand what the regression on 7.3.9 is :-/

src/callback.rs
```diff
@@ -244,6 +244,23 @@ where
...
+            if version.compare(crate::types::PyTuple::new(py, &PYPY_GOOD_VERSION))?
+                == std::cmp::Ordering::Less
+            {
+                let warn = py.import("warnings")?.getattr("warn")?;
+                warn.call1((
+                    "PyPy 3.7 versions older than 7.3.8 are known to have binary \
+                             compatibility issues which may cause segfaults. Please upgrade.",
+                ))?;
+            }
+        }
```
alex, line 263: I think this is the one place where we're changing the behavior of the code on pypy3.7. I don't have an obvious reason this would cause a crash, but it does jump out to me that we're doing this import+operations on every single function invocation.

alex:
Alternatively... computers cursed. CI's now passing just fine on 7.3.7

h-vetinari:
> Confirmed that the tests pass on pypy3.7 7.3.7, but I'd like to understand what the regression on 7.3.9 is :-/

CC @mattip

messense:
Thanks! @mejrs

src/callback.rs
mejrs: ...huh?

It looks like this piece was committed to main with https://github.com/PyO3/pyo3/pull/2217 but then picked into this branch with https://github.com/PyO3/pyo3/pull/2262/commits/c18e60f92894173e7ff6fa1ce0d6ecd002ebc824 but put in a different location...

mejrs: > I don't have an obvious reason this would cause a crash,

I think it's most likely that this set an exception/warning when it shouldn't have. For example, if you raise[^1] a warning during `__traverse__`, you can segfault.


[^1]: `__traverse__` doesn't give you a Python token, so this requires unsafe, which makes this not a problem

davidhewitt:
I've pushed a commit to support PyPy 3.7 v7.3.8+ instead of pinning to v7.3.7, because this was agreed as the correct solution to the binary compatibility issues in https://foss.heptapod.net/pypy/pypy/-/issues/3688#note_181074

alex:
I'll continue to bump the pyca/cryptography side for any changes here -- it's still green (I'm not going to leave more status updates unless that changes :-))

davidhewitt:
I think this is now good to go! Unless I hear otherwise, I'll put this live some time later today.

PR #2263, opened 2022-03-31
Add PyTzInfoAccess

pickfire:

 - an entry in CHANGELOG.md
 - tests for all new or changed functions

PyO3's CI pipeline will check your pull request. To run its tests
locally, you can run ```cargo xtask ci```. See its documentation
 [here](https://github.com/PyO3/pyo3/tree/main/xtask#readme).

As discussed in https://github.com/PyO3/pyo3/pull/1588#issuecomment-1083632399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add an experimental `generate-abi3-import-lib` feature to auto-generate `python3.dll` import libraries for Windows. [#2282](https://github.com/PyO3/pyo3/pull/2282)
 - Add FFI definitions for `PyDateTime_BaseTime` and `PyDateTime_BaseDateTime`. [#2294](https://github.com/PyO3/pyo3/pull/2294)
+- Added `PyTzInfoAccess`. [#2263](https://github.com/PyO3/pyo3/pull/2263)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add an experimental `generate-abi3-import-lib` feature to auto-generate `python3.dll` import libraries for Windows. [#2282](https://github.com/PyO3/pyo3/pull/2282)
 - Add FFI definitions for `PyDateTime_BaseTime` and `PyDateTime_BaseDateTime`. [#2294](https://github.com/PyO3/pyo3/pull/2294)
-- Added `PyTzInfoAccess`. [#2263](https://github.com/PyO3/pyo3/pull/2263)
+- Added `PyTzInfoAccess` for safe access to time zone information. [#2263](https://github.com/PyO3/pyo3/pull/2263)
 
 ### Changed
 
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -4,11 +4,8 @@
 //! documentation](https://docs.python.org/3/library/datetime.html)
 
 use crate::err::PyResult;
-use crate::ffi;
-#[cfg(PyPy)]
-use crate::ffi::datetime::{PyDateTime_FromTimestamp, PyDate_FromTimestamp};
 use crate::ffi::{
-    PyDateTime_CAPI, PyDateTime_FromTimestamp, PyDateTime_IMPORT, PyDate_FromTimestamp,
+    self, PyDateTime_CAPI, PyDateTime_FromTimestamp, PyDateTime_IMPORT, PyDate_FromTimestamp,
 };
 #[cfg(not(PyPy))]
 use crate::ffi::{PyDateTime_DATE_GET_FOLD, PyDateTime_TIME_GET_FOLD};
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -5,6 +5,8 @@
 
 use crate::err::PyResult;
 use crate::ffi;
+#[cfg(PyPy)]
+use crate::ffi::datetime::{PyDateTime_FromTimestamp, PyDate_FromTimestamp};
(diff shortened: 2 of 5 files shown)


src/types/datetime.rs
```diff
@@ -524,4 +556,33 @@ mod tests {
...
+
+            assert!(dt.get_tzinfo().is_none());
+
+            let t = PyTime::new(py, 0, 0, 0, 0, Some(&utc)).unwrap();
+
+            assert!(t.get_tzinfo().unwrap().eq(&utc).unwrap());
+
+            let t = PyTime::new(py, 0, 0, 0, 0, None).unwrap();
+
+            assert!(t.get_tzinfo().is_none());
```
pickfire, line 585: I have no idea why this is not none, need to investigate.

pickfire: I think I might need some help here, still kind confuse why this did not work.

```rust
crate::py_run!(t.py(), t, "assert t.tzinfo is None");
```

This did not error but when it got to rust side it errors. I further check and saw that `PyDateTime_TIME_GET_TZINFO(self.as_ptr())` returns `0x0000000000000001` even though it should be `None`, I wonder why it isn't `0x0000000000000000` which was returned by `PyDateTime_DATE_GET_TZINFO(self.as_ptr())`.

davidhewitt: @pickfire I've opened #2289 to resolve, I don't think we need to worry about the different `BaseTime` struct layout as long as we check `has_tzinfo` first.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow dependent crates to access config values from `pyo3-build-config` via cargo link dep env vars. [#2092](https://github.com/PyO3/pyo3/pull/2092)
 - Added methods on `InterpreterConfig` to run Python scripts using the configured executable. [#2092](https://github.com/PyO3/pyo3/pull/2092)
 - Added FFI definitions for `PyType_FromModuleAndSpec`, `PyType_GetModule`, `PyType_GetModuleState` and `PyModule_AddType`. [#2250](https://github.com/PyO3/pyo3/pull/2250)
+- Added `PyTzInfoAccess`. [#2263](https://github.com/PyO3/pyo3/pull/2263)
```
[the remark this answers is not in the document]
pickfire: Thanks, but probably I will rebase after your PR is merged.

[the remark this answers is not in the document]
pickfire: Should be good now.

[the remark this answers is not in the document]
davidhewitt: CI still failing, note that the ffi function now returns `None` instead of null pointer, which matches the C.

src/types/datetime.rs
```diff
@@ -160,6 +164,16 @@ pub trait PyTimeAccess {
...
 }
 
+/// Trait for accessing the components of a struct containing a tzinfo.
+pub trait PyTzInfoAccess {
+    /// Returns the tzinfo (which may be None).
+    ///
+    /// Implementations should conform to the upstream documentation:
+    /// <https://docs.python.org/3/c-api/datetime.html#c.PyDateTime_DATE_GET_TZINFO>
+    /// <https://docs.python.org/3/c-api/datetime.html#c.PyDateTime_TIME_GET_TZINFO>
+    fn get_tzinfo(&self) -> Option<&PyTzInfo>;
```
davidhewitt, line 174: Because of the change to the ffi layer I guess this will now always be `Some` unless the Python `None` is handled specially.

pickfire: Thanks, fixed. But I wonder if we really need to do the none check twice, once on underlaying layer and once on rust layer, I think might be better if we can just do the check once.

src/types/datetime.rs
```diff
@@ -354,6 +368,20 @@ impl PyTimeAccess for PyDateTime {
...
+    fn get_tzinfo(&self) -> Option<&PyTzInfo> {
+        let tzinfo: &PyTzInfo = unsafe {
+            self.py()
+                .from_borrowed_ptr(PyDateTime_DATE_GET_TZINFO(self.as_ptr()))
+        };
+        if tzinfo.is_none() {
+            None
+        } else {
+            Some(tzinfo)
+        }
```
davidhewitt, line 381: Maybe to avoid the dance with Python `None` we should just read from the FFI structs directly?

```suggestion
        let ptr = self.as_ptr() as *mut ffi::PyDateTime_DateTime;
        unsafe { 
            if (*ptr).has_tzinfo {
                Some(self.py().from_borrowed_ptr((*ptr).tzinfo)
            } else {
                None
            }
        }
```

and similar for the time implementation

src/types/datetime.rs
```diff
@@ -5,6 +5,8 @@
 
 use crate::err::PyResult;
 use crate::ffi;
+#[cfg(PyPy)]
+use crate::ffi::datetime::{PyDateTime_FromTimestamp, PyDate_FromTimestamp};
```
davidhewitt, line 9: These lines look unnecessary.

```suggestion
```

src/types/datetime.rs
```diff
@@ -524,4 +566,33 @@ mod tests {
             assert!(b.unwrap().get_fold());
         });
     }
+
+    #[cfg(not(PyPy))]
```
davidhewitt, line 570: We should probably test this on PyPy too. Probably best to do that by adding test to `pytests/src/datetime.rs` (and corresponding python file).

davidhewitt (commented):
Hmm maybe to avoid the dance with Python `None` we should just inline an equivalent of the FFI symbols.

messense:
@pickfire You need to resolve conflicts otherwise CI won't run.

adamreichold (approved):

davidhewitt:
That would be great; thanks & I'll put the release live tonight 👍

PR #2265, opened 2022-04-02
pyclass: mapping flag

davidhewitt:
This adds a `#[pyclass(mapping)]` option, to disable the sequence-slot-flling behaviour introduced in PyO3 0.16 (#2065).

This is helpful in cases where you want to implement a mapping, because having the sequence slots filled in that case is simply incorrect.

I didn't add `#[pyclass(sequence)]` because the it wouldn't do anything - to implement sequence slicing the mapping slots have to be used anyway.

### DIFF
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -192,6 +192,16 @@ and `Return` a final value - see its docs for further details and an example.
 
 ### Mapping & Sequence types
 
+The magic methods in this section can be used to implement Python container types. They are two main categories of container in Python: "mappings" such as `dict`, with arbitrary keys, and "sequences" such as `list` and `tuple`, with integer keys.
+
+The Python C-API which PyO3 is built upon has separate "slots" for sequences and mappings. When writing a `class` in pure Python, there is no such distinction in the implementation - a `__getitem__` implementation will fill the slots for both the mapping and sequence forms, for example.
+
+By default PyO3 reproduces the Python behaviour of filling both mapping and sequence slots. This makes sense for the "simple" case which matches Python, and also for sequences, where the mapping slot is used anyway to implement slice indexing.
+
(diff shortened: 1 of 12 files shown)


messense (approved):
👍

guide/src/class/protocols.md
```diff
@@ -192,6 +192,16 @@ and `Return` a final value - see its docs for further details and an example.
 
 ### Mapping & Sequence types
 
+The magic methods in this section can be used to implement Python container types. They are two main categories of container in Python: "mappings" such as `dict`, with arbitrary keys, and "sequences" such as `list` and `tuple`, with integer keys.
+
+The Python C-API which PyO3 is built upon has separate "slots" for sequences and mappings. When writing a `class` in pure Python, there is no such distinction in the implementation - a `__getitem__` implementation will fill the slots for both the mapping and sequence forms, for example.
+
+By default PyO3 reproduces the Python behaviour of filling both mapping and sequence slots. This makes sense for the "simple" case which matches Python, and also for sequences, where the mapping slot is used anyway to implement slice indexing.
+
+For mapping types, it may be desirable to not have the sequence slots filled. Use the `#[pyclass(mapping)]` annotation to instruct PyO3 to only fill the mapping slots, leaving the sequence ones empty.
```
adamreichold, line 201: I am a bit confused by the wording. The cover letter says that this is "simply not correct" while the guide will say that "it may be desirable". Could this be formulated in a more specific way, maybe using a concrete example of what would happen without this change and why that is undesirable?

mejrs: I agree, "may be desirable" sounds a bit...weasily.

davidhewitt: Good point, I'll try to push a better wording later...

davidhewitt: New wording pushed which gives examples of what is likely to go wrong without this flag set. How's that?

PR #2267, opened 2022-04-02
release: 0.16.3

davidhewitt:
Since we've got quite a number of non-breaking improvements stacked up, I propose we put these out before we start iterating on stuff which would need to be breaking for 0.17.

I'd also like to include #2265 in this release, as I ran into that as potentially helpful for the ecosystem in https://github.com/davidhewitt/pythonize/pull/25#issuecomment-1059922845

Arbitrarily set the release date for this to Tuesday however no rush at all, we can push it once #2265 is reviewed & merged.


pyo3-build-config/src/impl_.rs
```diff
@@ -825,7 +825,7 @@ pub(crate) fn cross_compile_env_vars() -> CrossCompileEnvVars {
 ///
 /// See the [PyO3 User Guide](https://pyo3.rs/) for more info on cross-compiling.
 #[deprecated(
-    since = "0.17.0",
```
davidhewitt, line 828: I usually am cautious about deprecating in patch releases, however I think it's so unlikely that this is used widely that it's fine.

ravenexp:
https://github.com/PyO3/pyo3/pull/2272 was split out of https://github.com/PyO3/pyo3/pull/2241 to reduce its scope, but it still logically belongs there.
Perhaps it could also make it into 0.16.3?

davidhewitt:
Yes, assuming `main` doesn't pick up any breaking changes I'll rebase on it when #2265 and #2273 merge.

davidhewitt:
Rebased. Might rebase again tomorrow to pick up #2269 and #2273, and then put live.

davidhewitt:
Release is live! Thanks all

PR #2268, opened 2022-04-03
Add PyString::intern to enable access to Python's built-in string interning.

adamreichold:
I think the API is somewhat unfortunate but I am not sure how to get around the requirement for a null-terminated string without the cost of the allocation that this API is supposed to avoid in the first place. At least not without a `PyUnicode_InternFromStringAndSize`.

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -144,6 +144,26 @@ impl PyString {
         unsafe { py.from_owned_ptr(ffi::PyUnicode_FromStringAndSize(ptr, len)) }
     }
 
+    /// Intern the given string
+    ///
+    /// This will return a reference to the same Python string object if called repeatedly with the same string.
+    ///
+    /// Note that while this is more memory efficient than [`PyString::new`], it unconditionally allocates a
+    /// temporary Python string object and is thereby slower than [`PyString::new`].
+    ///
+    /// Panics if out of memory.
+    pub fn intern<'p>(py: Python<'p>, s: &str) -> &'p PyString {
+        let ptr = s.as_ptr() as *const c_char;
+        let len = s.len() as ffi::Py_ssize_t;
+        unsafe {
+            let mut ob = ffi::PyUnicode_FromStringAndSize(ptr, len);
+            if !ob.is_null() {
+                ffi::PyUnicode_InternInPlace(&mut ob);
+            }
+            py.from_owned_ptr(ob)
+        }
+    }
+
     /// Attempts to create a Python string from a Python [bytes-like object].
     ///
     /// [bytes-like object]: (https://docs.python.org/3/glossary.html#term-bytes-like-object).
@@ -590,6 +610,24 @@ mod tests {
                 .to_string()
                 .contains("'utf-32' codec can't decode bytes in position 0-7"));
             assert_eq!(data.to_string_lossy(), Cow::Owned::<str>("𠀀�".into()));
+        });
+    }
+
+    #[test]
+    fn test_intern_string() {
+        Python::with_gil(|py| {
+            let py_string1 = PyString::intern(py, "foo");
+            assert_eq!(py_string1.to_str().unwrap(), "foo");
+
+            let py_string2 = PyString::intern(py, "foo");
+            assert_eq!(py_string2.to_str().unwrap(), "foo");
+
+            assert_eq!(py_string1.as_ptr(), py_string2.as_ptr());
+
+            let py_string3 = PyString::intern(py, "bar");
+            assert_eq!(py_string3.to_str().unwrap(), "bar");
+
+            assert_ne!(py_string1.as_ptr(), py_string3.as_ptr());
         });
     }
 }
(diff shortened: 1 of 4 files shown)


birkenfeld:
I'd at least add a version that takes `&str`. This is so inconvenient that it's going to hurt uptake, and there is an allocation anyway in the Python function.

adamreichold:
> I'd at least add a version that takes `&str`. This is so inconvenient that it's going to hurt uptake

Do you mean a function taking `&str` and only checking that it is null-terminated?

> and there is an allocation anyway in the Python function.

While that is true for the [current implementation](https://github.com/python/cpython/blob/3faa9f78d4b9a8c0fd4657b434bdb08ae1f28800/Objects/unicodeobject.c#L15596), I think this is actually a performance bug as that implementation should probably use `PyDict_{Get,Set}ItemString` instead of immediately allocating a `PyUnicode` object.

Or put differently, if this is guaranteed to allocate a temporary `PyUnicode` object for every call, then this will actually be useless in the context of https://github.com/PyO3/pyo3/discussions/2266

birkenfeld:
>>  I'd at least add a version that takes &str. This is so inconvenient that it's going to hurt uptake
> Do you mean a function taking &str and only checking that it is null-terminated?

Either that, or one creating the intermediary CString by itself. It could even do both, by trying first to create a `CStr` and making a new null-terminated string if necessary. Then you can easily pass "abc\0" for speed, or any user given str for convenience.

>> and there is an allocation anyway in the Python function.
> While that is true for the current implementation, I think this is actually a performance bug as that implementation should probably use PyDict_{Get,Set}ItemString instead of immediately allocating a PyUnicode object.

No, I mean inside `PyUnicode_InternFromString`.

> Or put differently, if this is guaranteed to allocate a temporary PyUnicode object for every call, then this will actually be useless in the context of #2266 and we would probably be better off exposing PyDict_{Get,Set}ItemString ourselves.

Once you have interned the string objects, there is no need to allocate more for every call.

birkenfeld:
Sorry I wasn't thinking properly about the "benchmark dict access with strings" use case.

In general, having an API for interning is still useful for other cases, so I'd like this to go forward independently.

adamreichold:
> > > I'd at least add a version that takes &str. This is so inconvenient that it's going to hurt uptake
> > > Do you mean a function taking &str and only checking that it is null-terminated?
> 
> Either that, or one creating the intermediary CString by itself. It could even do both, by trying first to create a `CStr` and making a new null-terminated string if necessary. Then you can easily pass "abc\0" for speed, or any user given str for convenience.

Added a separate commit providing a more convenient API containing `CStr::from_bytes_with_nul` as a fast path.

birkenfeld:
Thanks! BTW, I can never figure out what people mean by the "eyes" reaction, what's your take on that?

src/types/string.rs
```diff
@@ -144,6 +144,33 @@ impl PyString {
...
+    /// # });
+    /// ```
+    pub fn intern<'p>(py: Python<'p>, s: &str) -> &'p PyString {
+        if let Ok(s) = CStr::from_bytes_with_nul(s.as_bytes()) {
+            unsafe { py.from_owned_ptr(ffi::PyUnicode_InternFromString(s.as_ptr())) }
+        } else {
+            let s = CString::new(s).unwrap();
+            unsafe { py.from_owned_ptr(ffi::PyUnicode_InternFromString(s.as_ptr())) }
+        }
+    }
```
mejrs, line 172: Note that the C function creates a PyString and interns that:

```c
PyObject *
PyUnicode_InternFromString(const char *cp)
{
    PyObject *s = PyUnicode_FromString(cp);
    if (s == NULL)
        return NULL;
    PyUnicode_InternInPlace(&s);
    return s;
}
```

Perhaps it would be better to do what that function does:

```rust
pub fn intern<'p>(py: Python<'p>, s: &str) -> &'p PyString {
        let pystring = PyUnicode_FromStringAndSize(s.as_ptr() as *const _, s.len());
        PyUnicode_InternInPlace(pystring);
        py.from_owned_ptr(pystring)
}
```

src/types/string.rs
```diff
@@ -144,6 +144,33 @@ impl PyString {
         unsafe { py.from_owned_ptr(ffi::PyUnicode_FromStringAndSize(ptr, len)) }
     }
 
+    /// Intern the given string
+    ///
+    /// This will return a reference to the same Python string object if called repeatedly with the same string.
+    ///
+    /// Note that a temporary allocation is required if the given string is not already null-terminated.
```
mejrs, line 151: This will always allocate though? Because a PyString is always created whether or not it is already interned.

I understand this isn't more performant than just using `PyString::new`, only more memory efficient? That should be mentioned IMO.

birkenfeld: The allocation of the CString is meant here.

src/types/string.rs
birkenfeld: Nice, that definitely makes more sense.

adamreichold:
> Thanks! BTW, I can never figure out what people mean by the "eyes" reaction, what's your take on that?

The meaning I wanted to convey was "I am looking into that", so when I am currently thinking about/working on something but do not have any positive or negative results to share yet.

src/types/string.rs
adamreichold: This is what I started with, but as written in https://github.com/PyO3/pyo3/pull/2268#issuecomment-1086797378, I consider the current implementation in CPython to be a performance bug as the `PyUnicode_InternFromString` does not need to literally call `PyUnicode_FromString` followed by `PyUnicode_InternInPlace`.

It could rather look up a possibly already interned string using only the null-terminated string and just increment its reference count without any allocation at all. The way it is written ATM would pick this up when `PyUnicode_InternFromString` is fixed.

src/types/string.rs
adamreichold: As written above, that the current implementation provided by CPython unconditionally allocates is a bug IMHO.

adamreichold:
> In general, having an API for interning is still useful for other cases, so I'd like this to go forward independently.

@mejrs @birkenfeld I rebooted this into a wrapper that is purely targeted at memory efficiency as `PyDict_{Set,Get}ItemString` actually have the same deficiency as `PyUnicode_InternFromString`.

birkenfeld:
LGTM.

mejrs (approved):
Thanks!

birkenfeld (approved):

PR #2272, opened 2022-04-04
pyo3-build-config: Add `PYO3_CROSS_PYTHON_IMPLEMENTATION` env var

ravenexp:
Adds a new cross-compile target interpreter configuration
environment variable.

This feature allows PyO3 to target PyPy on both Windows and Unix
cross compile targets.


davidhewitt (approved):
Thank you for implementing this! It's been a longstanding gap in the cross-compile configuration which I never got around to fixing...

PR #2273, opened 2022-04-04
Make use of intern! macro for attribute names used internally

adamreichold:
Since we do use several attribute names internally, I looked into whether some of them could be candidates for interning. Personally, I do not think this is obvious as while identifiers like `__name__` will most likely already be part of Python's string interner dictionary, they do not seem to be used in a hot path of PyO3.

I therefore split this into separate commits to solicit feedback on whether we want to include any of them:
1. `__qualname__` used in `PyType::name`. Here, I am not sure how often production code using PyO3 actually calls this.
2. `__all__` and `__name__` used in the implementation of `PyModule`. This seems to be limited to initialization, but the strings are almost surely interned already by other code.
3. The the attribute names used in `getattr` calls emitted by the `FromPyObject` derive macro. I think this has the largest possible benefit as this could end up being used in every Python-to-Rust call. But I am also not sure whether enabling this unconditionally is a good idea. (Since the names are given via code and not generated dynamically, this should be ok?)

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -534,6 +534,25 @@ impl<T> Py<T> {
     /// Retrieves an attribute value.
     ///
     /// This is equivalent to the Python expression `self.attr_name`.
+    ///
+    /// If calling this method becomes performance-critical, the [`intern!`] macro can be used
+    /// to intern `attr_name`, thereby avoiding repeated temporary allocations of Python strings.
+    ///
+    /// # Example: `intern!`ing the attribute name
+    ///
+    /// ```
+    /// # use pyo3::{intern, pyfunction, types::PyModule, IntoPy, Py, Python, PyObject, PyResult};
+    /// #
+    /// #[pyfunction]
+    /// fn version(sys: Py<PyModule>, py: Python<'_>) -> PyResult<PyObject> {
+    ///     sys.getattr(py, intern!(py, "version"))
+    /// }
+    /// #
+    /// # Python::with_gil(|py| {
+    /// #    let sys = py.import("sys").unwrap().into_py(py);
+    /// #    version(sys, py).unwrap();
+    /// # });
+    /// ```
     pub fn getattr<N>(&self, py: Python<'_>, attr_name: N) -> PyResult<PyObject>
     where
         N: ToPyObject,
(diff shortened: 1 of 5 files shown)


src/types/module.rs
```diff
@@ -168,12 +167,12 @@ impl PyModule {
...
     /// `__all__` declares the items that will be imported with `from my_module import *`.
     pub fn index(&self) -> PyResult<&PyList> {
-        match self.getattr("__all__") {
+        match self.getattr(__all__(self.py())) {
             Ok(idx) => idx.downcast().map_err(PyErr::from),
             Err(err) => {
                 if err.is_instance_of::<exceptions::PyAttributeError>(self.py()) {
                     let l = PyList::empty(self.py());
-                    self.setattr("__all__", l).map_err(PyErr::from)?;
+                    self.setattr(__all__(self.py()), l).map_err(PyErr::from)?;
```
davidhewitt, line 175: TBH here it looks like we should have a `let __all__ = __all__(self.py())` at the top to avoid looking up the interned value twice.

davidhewitt (approved):
Sneaky idea that should be nice for performance! It makes me a little uneasy if we know we might have to regress this for PEP 489 combatibility later, however it's still a nice win for the meanwhile.

birkenfeld:
Related idea: add a comment to `getattr`/`setattr` recommending `intern!`?

src/types/module.rs
adamreichold: Fixed.

adamreichold:
> Related idea: add a comment to `getattr`/`setattr` recommending `intern!`?

Add a comment to the methods of `Py` and `PyAny`. Would be glad for feedback on the wording. I think linking the `PyDict` example given at `intern!`'s docs is sufficient or should I add a `getattr`-specific one?

src/instance.rs
```diff
@@ -534,6 +534,9 @@ impl<T> Py<T> {
     /// Retrieves an attribute value.
     ///
     /// This is equivalent to the Python expression `self.attr_name`.
+    ///
+    /// If calling this method becomes performance-critical, using the [`intern!`] macro can be
+    /// used to intern `attr_name` thereby avoid repeated temporary allocations of Python strings.
```
birkenfeld, line 539:
```suggestion
    /// used to intern `attr_name`, thereby avoiding repeated temporary allocations of Python strings.
```

src/instance.rs
```diff
@@ -534,6 +534,9 @@ impl<T> Py<T> {
     /// Retrieves an attribute value.
     ///
     /// This is equivalent to the Python expression `self.attr_name`.
+    ///
+    /// If calling this method becomes performance-critical, using the [`intern!`] macro can be
```
birkenfeld, line 538:
```suggestion
    /// If calling this method becomes performance-critical, the [`intern!`] macro can be
```

birkenfeld (commented):
I think an example would not go amiss.

adamreichold:
> I think an example would not go amiss.

Added a single example to each of the four cases.

birkenfeld:
Thanks a lot!

Issue #2278, opened 2022-04-06
Add performance suggestions to docs


samuelcolvin:
Hi, thanks so much for pyo3 - it's wonderful, I'm using it a lot (see [rtoml](https://github.com/samuelcolvin/rtoml) and [watchfiles](https://github.com/samuelcolvin/watchfiles), I might be using it pydantic soon too).

I wonder if it would be worth adding a "Performance Suggestions" or similar section to the docs, perhaps the FAQs?

Examples:

## `cast_as` vs `extract`

I spent a lot of the morning very worried that pyo3 rust making heavy use of python types was actually slower than python, until by chance I tried `cast_as` instead of `extract`.

```rs
// fast
let list: &PyList = py_any.cast_as()?;
// slow
let list: &PyList = py_any.extract()?;
```

This change in my case made a ~4x improvement to performance.

## Optimised compilation

While investigating the above, I noticed [this SO question](https://stackoverflow.com/questions/66467640/why-cython-embeded-plugins-has-higher-performance-in-cpython-interpreter-than-ru) referenced on #1470 didn't mention that the performance is massively effected by debug/release compilation. I've added an answer to the question with suggestions.

----

These are both no doubt very obvious to the maintainers of this library and sessioned users, but to those new to rust and pyo3, they might be less obvious.

I'm sure there are other things which can have a dramatic effect on performance which I'm not ware of - I've love to hear any other general suggestions?

---

**Update:**

Other things to suggest when this section is written:
* interning strings where they're being used a lot, a good example is if you're creating a dict with common keys, performance can be significantly improved by reusing `PyString`s rather than rust strings

adamreichold:
> I spent a lot of the morning very worried that pyo3 rust making heavy use of python types was actually slower than python, until by chance I tried cast_as instead of extract.

I am surprised by this. Shouldn't both `cast_as` and `extract` for a native type end up calling `PyTryFrom::try_from`? `cast_as` [directly](https://docs.rs/pyo3/latest/src/pyo3/types/any.rs.html#699-704) and `extract` [via `pyobject_native_type_extract!`](https://docs.rs/pyo3/latest/src/pyo3/types/mod.rs.html#164) which should both end up [here](https://docs.rs/pyo3/latest/src/pyo3/conversion.rs.html#369).

Is it possible to show that performance difference using a self-contained reproducer?

adamreichold:
I tried the following benchmark

```diff
[...]

Issue #426, opened 2019-04-05
Revisit FromPyObject


kngwyu:
This is an experimental proposal. Any feedback is welcome, but please consider that it's experimental.

## TL; DR
We should have two traits `FromPyType` and `FromPyProto` instead of `FromPyObject`.

### The problem of `FromPyObject`
Currently, we have `FromPyObject` as a main interface to get a Rust type from a python object.
It is defined as 
```rust
pub trait FromPyObject<'source>: Sized {
    /// Extracts `Self` from the source `PyObject`.
    fn extract(ob: &'source PyAny) -> PyResult<Self>;
}
```
It's general since it requires `PyAny`, but it has a problem: we can't raise compile error using Rust's type system.
E.g.,
```rust
let l = PyComplex::from_doubles(py, 3.0, 1.2);
let v: Vec<i32> = FromPyObject::extract(l.as_ref()).unwrap();
```
this code raises a runtime error.
To prevent this, we should allow just a set of python types to become a rust type.
E.g., we should have
```rust
trait FromPyType<T: PyTypeInfo> {
    fn from_py(py: Python, obj: PyRef<T>) -> Self:
}
impl FromPyType<PyComplex> for Complex<f32> {
...
}
```

### The problem of `FromPyType`
Its problem is  we can not use it for conversions using object protocols.
We have some conversion codes using object protocol.
For example, to extract `Vec` from a Python object, we try either `Sequence` and `Buffer` protocol, or only `Sequence` protocol via specialization, like
```rust
impl<'a, T> FromPyObject<'a> for Vec<T>
where
    T: FromPyObject<'a>,
{
    default fn extract(obj: &'a PyAny) -> PyResult<Self> {
        extract_sequence(obj)
    }
}

impl<'source, T> FromPyObject<'source> for Vec<T>
where
    for<'a> T: FromPyObject<'a> + buffer::Element + Copy,
{
    fn extract(obj: &'source PyAny) -> PyResult<Self> {
        // first try buffer protocol
        if let Ok(buf) = buffer::PyBuffer::get(obj.py(), obj) {
            ...
        }
    }
}
```
(https://github.com/PyO3/pyo3/blob/72003ec37a16634e3fdff1cd7f03ab8814ecde42/src/types/sequence.rs#L237-#L264)
This looks difficult to give an abstraction via type systems.
One solution is to use types corresponding to each protocols.
```rust
impl<T> FromPyType<PyBuffer> for Vec<T> {...}
[...]

PR #2279, opened 2022-04-06
Add benchmark highlighting the costs of failed calls to FromPyObject::extract.

adamreichold:
This was brought up in #2278 and probably affects quite a bit of code that does a case-by-case analysis of argument types:

```
enum_from_pyobject      time:   [644.87 ns 648.64 ns 653.16 ns]                                
list_via_cast_as        time:   [3.4474 ns 3.4668 ns 3.4808 ns]                              
list_via_extract        time:   [4.3715 ns 4.3785 ns 4.3896 ns]                              
not_a_list_via_cast_as  time:   [3.4952 ns 3.5034 ns 3.5094 ns]                                    
not_a_list_via_extract  time:   [284.42 ns 284.79 ns 285.12 ns]                                   
not_a_list_via_extract_enum                                                                            
                        time:   [289.56 ns 289.78 ns 289.98 ns]
```

The overhead itself mostly comes from this

```rust
/// Convert `PyDowncastError` to Python `TypeError`.
impl<'a> std::convert::From<PyDowncastError<'a>> for PyErr {
    fn from(err: PyDowncastError<'_>) -> PyErr {
        exceptions::PyTypeError::new_err(err.to_string())
    }
}
```

conversion.

I find it especially troubling that the rather nice way of using an enum is also affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Default to "m" ABI tag when choosing `libpython` link name for CPython 3.7 on Unix. [#2288](https://github.com/PyO3/pyo3/pull/2288)
+- Improved performance of failing calls to `FromPyObject::extract` which is common when functions accept multiple distinct types. [#2279](https://github.com/PyO3/pyo3/pull/2279)
 
 ## [0.16.3] - 2022-04-05
 
--- a/benches/bench_frompyobject.rs
+++ b/benches/bench_frompyobject.rs
@@ -1,6 +1,9 @@
-use criterion::{criterion_group, criterion_main, Bencher, Criterion};
+use criterion::{black_box, criterion_group, criterion_main, Bencher, Criterion};
 
-use pyo3::{prelude::*, types::PyString};
+use pyo3::{
+    prelude::*,
+    types::{PyList, PyString},
+};
 
 #[derive(FromPyObject)]
 enum ManyTypes {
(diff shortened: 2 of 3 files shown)


adamreichold:
I pushed a commit adding `PyDowncastErrorArguments` to at least delay the actual call into the formatting machinery but I am not happy about duplicating the format string and it only improves thing to one order of magnitude slower instead of two:

```
enum_from_pyobject      time:   [669.27 ns 672.85 ns 676.48 ns]    
list_via_cast_as        time:   [3.4959 ns 3.4966 ns 3.4975 ns]                              
list_via_extract        time:   [3.0602 ns 3.0628 ns 3.0659 ns]                              
not_a_list_via_cast_as  time:   [3.2751 ns 3.4409 ns 3.5769 ns]                                    
not_a_list_via_extract  time:   [45.943 ns 46.069 ns 46.255 ns]                                    
not_a_list_via_extract_enum                                                                             
                        time:   [44.261 ns 44.301 ns 44.344 ns]
```

So IMHO, this still seems to be a performance foot gun, especially in derived code.

mejrs:
How about changing the `FromPyObject` trait itself?

```rust
pub trait FromPyObject<'source>: Sized {
    type Error;
    /// Extracts `Self` from the source `PyObject`.
    fn extract(ob: &'source PyAny) -> Result<Self, Self::Error>;
}
```
Where the Error type would be `PyDowncastError` for a Python type and PyErr otherwise.


(note: if we start changing conversion traits we should probably consider all of them)

adamreichold:
> How about changing the FromPyObject trait itself?

I think we should consider this, but this could be in a maintenance release helping existing deployments without obstructing such a change to `FromPyObject`, couldn't it?

Regarding adding an associated `Error` type to `FromPyObject`, I think we would need to consider the problem of composition, i.e.

```rust
#[derive(FromPyObject)]
struct Foo {
  bar: Bar,
  qux: Qux,
}
```

must derive an `Error` type that can be construct from either `<Bar as FromPyObject>::Error` and `<Qux as FromPyObject>::Error`. Do we create ad-hoc error types for this? Or do we require a trait bound and box them up?

davidhewitt:
> How about changing the `FromPyObject` trait itself?

There's a very old thread at #426 suggesting a rework. That thread is probably very out of date; it might be worth killing and starting a new discussion.

For what it's worth, the idea in #1308 would also change `FromPyObject`, by adding a separate lifetime (one becomes the lifetime of the data, the second the lifetime of the GIL).

davidhewitt (approved):
Yep makes sense. I think would be helpful to put out a 0.16.4 for the coverage fix as well as #2289, so let's merge this as-is and then consider further refactoring for 0.17 later.

Please add a `Changed` CHANGELOG entry documenting the performance tweak and then feel free to merge 👍

adamreichold:
> Please add a Changed CHANGELOG entry documenting the performance tweak and then feel free to merge +1

Done. Will give it some time if anyone fancies a look and merge in the afternoon.

PR #2282, opened 2022-04-07
pyo3-build-config: Add `python3-dll-a` crate support

### LINKED ISSUE #2231: Generate `python3.dll` import libraries for Windows targets in the build script

ravenexp:
Automatically generate `python3.dll` import libraries
for Windows compile targets in the build script.

Adds a new PyO3 crate feature `auto-abi3-import-lib` enabling
automatic import library generation.

Closes #2231

WIP: The documentation is still a TODO

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -272,17 +272,17 @@ jobs:
           args: --release -i python3.9 --no-sdist -m examples/maturin-starter/Cargo.toml
 
       - name: Test cross compile to Windows
-        if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.10' }}
+        if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.8' }}
         run: |
           sudo apt-get install -y mingw-w64
           rustup target add x86_64-pc-windows-gnu
           cargo build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-gnu
       - name: Test cross compile to Windows with maturin
-        if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.10' }}
+        if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.8' }}
         uses: messense/maturin-action@v1
         with:
           target: x86_64-pc-windows-gnu
-          args: --no-sdist -m examples/maturin-starter/Cargo.toml --cargo-extra-args="--features abi3"
+          args: -i python3.8 --no-sdist -m examples/maturin-starter/Cargo.toml --cargo-extra-args="--features abi3"
 
     env:
       CARGO_TERM_VERBOSE: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -274,12 +274,13 @@ jobs:
       - name: Test cross compile to Windows
         if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.8' }}
         env:
-          CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_LINKER: rust-lld
+          XWIN_ARCH: x86_64
         run: |
           sudo apt-get install -y mingw-w64 llvm
           rustup target add x86_64-pc-windows-gnu x86_64-pc-windows-msvc
+          which cargo-xwin > /dev/null || cargo install cargo-xwin
           cargo build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-gnu
-          cargo build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-msvc
+          cargo xwin build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-msvc
       - name: Test cross compile to Windows with maturin
         if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.8' }}
         uses: messense/maturin-action@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,10 +273,13 @@ jobs:
 
       - name: Test cross compile to Windows
         if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.8' }}
+        env:
+          CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_LINKER: rust-lld
         run: |
-          sudo apt-get install -y mingw-w64
-          rustup target add x86_64-pc-windows-gnu
+          sudo apt-get install -y mingw-w64 llvm
+          rustup target add x86_64-pc-windows-gnu x86_64-pc-windows-msvc
           cargo build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-gnu
+          cargo build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-msvc
       - name: Test cross compile to Windows with maturin
         if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.8' }}
         uses: messense/maturin-action@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -271,6 +271,19 @@ jobs:
           target: aarch64-apple-darwin
           args: --release -i python3.9 --no-sdist -m examples/maturin-starter/Cargo.toml
 
+      - name: Test cross compile to Windows
+        if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.10' }}
+        run: |
+          sudo apt-get install -y mingw-w64
+          rustup target add x86_64-pc-windows-gnu
+          cargo build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-gnu
+      - name: Test cross compile to Windows with maturin
+        if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.10' }}
+        uses: messense/maturin-action@v1
+        with:
+          target: x86_64-pc-windows-gnu
+          args: --no-sdist -m examples/maturin-starter/Cargo.toml --cargo-extra-args="--features abi3"
+
     env:
       CARGO_TERM_VERBOSE: true
       CARGO_BUILD_TARGET: ${{ matrix.platform.rust-target }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,9 @@ abi3-py38 = ["abi3-py39", "pyo3-build-config/abi3-py38", "pyo3-ffi/abi3-py38"]
 abi3-py39 = ["abi3-py310", "pyo3-build-config/abi3-py39", "pyo3-ffi/abi3-py39"]
 abi3-py310 = ["abi3", "pyo3-build-config/abi3-py310", "pyo3-ffi/abi3-py310"]
 
+# Automatically generates `python3.dll` import libraries for Windows targets.
+generate-abi3-import-lib = ["pyo3-build-config/python3-dll-a"]
+
 # Changes `Python::with_gil` and `Python::acquire_gil` to automatically initialize the
 # Python interpreter if needed.
 auto-initialize = []
--- a/guide/src/building_and_distribution.md
+++ b/guide/src/building_and_distribution.md
@@ -152,9 +152,19 @@ As your extension module may be run with multiple different Python versions you 
 
 PyO3 is only able to link your extension module to api3 version up to and including your host Python version. E.g., if you set `abi3-py38` and try to compile the crate with a host of Python 3.7, the build will fail.
 
-As an advanced feature, you can build PyO3 wheel without calling Python interpreter with the environment variable `PYO3_NO_PYTHON` set. On unix systems this works unconditionally; on Windows you must also set the `RUSTFLAGS` evironment variable to contain `-L native=/path/to/python/libs` so that the linker can find `python3.lib`.
-
 > Note: If you set more that one of these api version feature flags the lowest version always wins. For example, with both `abi3-py37` and `abi3-py38` set, PyO3 would build a wheel which supports Python 3.7 and up.
+
+#### Building `abi3` extensions without a Python interpreter
+
+As an advanced feature, you can build PyO3 wheel without calling Python interpreter with the environment variable `PYO3_NO_PYTHON` set.
+On Unix-like systems this works unconditionally; on Windows you must also set the `RUSTFLAGS` environment variable
+to contain `-L native=/path/to/python/libs` so that the linker can find `python3.lib`.
+
+If the `python3.dll` import library is not available, an experimental `generate-abi3-import-lib` crate
+feature may be enabled, and the required library will be created and used by PyO3 automatically.
+
+*Note*: MSVC targets require LLVM binutils (`llvm-dlltool`) to be available in `PATH` for
+the automatic import library generation feature to work.
 
 #### Missing features
 
@@ -218,8 +228,8 @@ Thanks to Rust's great cross-compilation support, cross-compiling using PyO3 is 
 
 * A toolchain for your target.
 * The appropriate options in your Cargo `.config` for the platform you're targeting and the toolchain you are using.
-* A Python interpreter that's already been compiled for your target.
-* A Python interpreter that is built for your host and available through the `PATH` or setting the [`PYO3_PYTHON`](#python-version) variable.
+* A Python interpreter that's already been compiled for your target (optional when building "abi3" extension modules).
+* A Python interpreter that is built for your host and available through the `PATH` or setting the [`PYO3_PYTHON`](#python-version) variable (optional when building "abi3" extension modules).
 
 After you've obtained the above, you can build a cross-compiled PyO3 module by using Cargo's `--target` flag. PyO3's build script will detect that you are attempting a cross-compile based on your host machine and the desired target.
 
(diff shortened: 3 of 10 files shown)


Cargo.toml
```diff
@@ -78,6 +78,9 @@ abi3-py38 = ["abi3-py39", "pyo3-build-config/abi3-py38", "pyo3-ffi/abi3-py38"]
 abi3-py39 = ["abi3-py310", "pyo3-build-config/abi3-py39", "pyo3-ffi/abi3-py39"]
 abi3-py310 = ["abi3", "pyo3-build-config/abi3-py310", "pyo3-ffi/abi3-py310"]
 
+# Generates `python3.dll` import libraries automatically.
+auto-abi3-import-lib = ["pyo3-build-config/python3-dll-a"]
```
adamreichold, line 82: Sorry for bike shedding, but I think `generate-abi3-import-lib` matching the function name would make the construction easier to follow.

pyo3-build-config/src/util.rs
```diff
@@ -0,0 +1,59 @@
+//! Additional functionality enabled by crate features
```
adamreichold, line 1: Should we have a "miscellaneous" module before having more than one optional function? I think a module that is specific to the python3-dll-a feature would nicer.

pyo3-build-config/src/util.rs
```diff
@@ -0,0 +1,59 @@
...
+    Ok(Some(out_lib_dir_string))
+}
+
+/// Generates `python3.dll` import library for Windows targets.
+///
+/// Places the import library into the build script output directory
+/// and returns the full library directory path.
+///
+/// Does nothing if the target OS is not Windows.
+#[cfg(not(feature = "python3-dll-a"))]
```
adamreichold, line 56: I think it would be nicer to make the invocation dependent on the feature in this case as right now, `_impls.rs` looks like something (maybe some fallback logic) might happen when actually this is a no-op. (If the module was feature-specific, the `mod` statement itself could carry the `cfg` directive (in addition to the invocation).)

Cargo.toml
ravenexp: This is fine with me.

pyo3-build-config/src/util.rs
ravenexp: This feature uses an external crate that is an optional dependency. I can add an `if cfg!() {}` statement to `_impl.rs`, but even then I'd still need to define two versions of `generate_abi3_import_lib()`.

adamreichold: I was thinking of something like

```rust
#[cfg(feature = "python3-dll-a")]
#[path = "util.rs"]
mod python3_dll_a;
...
pub fn make_interpreter_config() -> Result<InterpreterConfig> {
...
    // Auto generate python3.dll import libraries for Windows targets.
    #[cfg(feature = "python3-dll-a")]
    interpreter_config.lib_dir = python3_dll_a::generate_abi3_import_lib(&host)?;
...
}
```

in `_impl.rs` and then no `cfg`-directives at all in `python3_dll_a.rs`.

ravenexp: Do `#[cfg(...)]` directives work at the statement level? I didn't know that...

adamreichold: I think they are still experiment on the expression level, c.f. https://github.com/rust-lang/rust/issues/15701 but that can usually be worked around by adding a block, i.e. both

```rust
...
#[cfg(feature = "python3-dll-a")]
{
    // Auto generate python3.dll import libraries for Windows targets.
    interpreter_config.lib_dir = python3_dll_a::generate_abi3_import_lib(&host)?;
}
...
```

and

```rust
...
#[cfg(feature = "python3-dll-a")]
if abi3 && lib_dir.is_none() {
    // Auto generate python3.dll import libraries for Windows targets.
    lib_dir = python3_dll_a::generate_abi3_import_lib(&host)?;
}
...
```

should work.

(This can provide dead-code and unneeded-mut noise. So it isn't totally clear that it is the better. My primary aim here is to make it visible at the call site, that the operation is optional depending on whether the dependency is present.)

ravenexp: Thanks for the feedback! I'll try to apply these suggestions and submit the new patch tomorrow.

I can't name the submodule `python3_dll_a` because it's also the external crate name, so I'll go with `abi3_import_lib` or something.

guide/src/building_and_distribution.md
```diff
@@ -152,10 +152,14 @@ As your extension module may be run with multiple different Python versions you
...
 
-As an advanced feature, you can build PyO3 wheel without calling Python interpreter with the environment variable `PYO3_NO_PYTHON` set. On unix systems this works unconditionally; on Windows you must also set the `RUSTFLAGS` evironment variable to contain `-L native=/path/to/python/libs` so that the linker can find `python3.lib`.
-
 > Note: If you set more that one of these api version feature flags the lowest version always wins. For example, with both `abi3-py37` and `abi3-py38` set, PyO3 would build a wheel which supports Python 3.7 and up.
 
+As an advanced feature, you can build PyO3 wheel without calling Python interpreter with the environment variable `PYO3_NO_PYTHON` set.
+On Unix-like systems this works unconditionally; on Windows you must also set the `RUSTFLAGS` evironment variable to contain
+`-L native=/path/to/python/libs` so that the linker can find `python3.lib`. Alternatively, an experimental `generate-abi3-import-lib`
+crate feature may be enabled, and the DLL import library will be created automatically by PyO3.
```
adamreichold, line 160: Do I still need to modify `RUSTFLAGS` when the feature is enabled?

guide/src/building_and_distribution.md
```diff
@@ -230,6 +234,14 @@ When cross-compiling, PyO3's build script cannot execute the target Python inter
 * `PYO3_CROSS_PYTHON_VERSION`: Major and minor version (e.g. 3.9) of the target Python installation. This variable is only needed if PyO3 cannot determine the version to target from `abi3-py3*` features, or if `PYO3_CROSS_LIB_DIR` is not set, or if there are multiple versions of Python present in `PYO3_CROSS_LIB_DIR`.
 * `PYO3_CROSS_PYTHON_IMPLEMENTATION`: Python implementation name ("CPython" or "PyPy") of the target Python installation. CPython is assumed by default when this variable is not set, unless `PYO3_CROSS_LIB_DIR` is set for a Unix-like target and PyO3 can get the interpreter configuration from `_sysconfigdata*.py`.
 
+An experimental `pyo3` crate feature `generate-abi3-import-lib` enables the user to cross compile
+"abi3" extension modules for Windows targets without setting `PYO3_CROSS_LIB_DIR` environment
+variable or providing any Windows Python library files. It uses external `python3-dll-a` crate
```
adamreichold, line 239: A link to the crate homepage might be a nice courtesy?

guide/src/building_and_distribution.md
adamreichold: And if so, could we amend the build script to automatically set the search when we import library is automatically genereted?

adamreichold (approved):

guide/src/building_and_distribution.md
ravenexp: > Do I still need to modify `RUSTFLAGS` when the feature is enabled?

No, when the feature is enabled nothing else is required. I'll try to reword the sentence it to make that clearer.

guide/src/building_and_distribution.md
ravenexp: I'll add a link to `docs.rs` page. As for the GitHub project page, I actually hoped to transfer that project to the PyO3 organization after this feature is done. Unfortunately, there's a chance I might lose access to my GitHub or Gmail accounts because of the circumstances, and I don't want PyO3 to end up with an unmaintainable dependency.

guide/src/building_and_distribution.md
ravenexp: I've rewritten this part as a separate paragraph.

guide/src/building_and_distribution.md
ravenexp: I've added links to the `docs.rs` page to both the user guide and Architecture.md.

adamreichold: The changes look fine to me. Leaving this unresolved due to the question of long-term maintenance though.

ravenexp: I don't want to be misunderstood here. I _do_ intend to maintain it and to contribute to PyO3 in the future, but I also want to safeguard against the situation where I'd be unable to do so. GitHub has blocked accounts of users from sanctioned countries before, and if that happens I don't want it to become a problem for the projects that depend on mine. That's why I want to transfer the GitHub project and the `crates.io` package ownership to an organization as a backup measure.

davidhewitt (approved):
Sorry for the delay on my part, thanks @adamreichold for reviewing.

Last thing that would be nice to see IMO would be a check for this in CI. I think it could probably be done similarly to the other cross compilation check - https://github.com/PyO3/pyo3/blob/2a54641d9149831f10a94e8e5bc0790d48d56763/.github/workflows/ci.yml#L245

ravenexp:
> Last thing that would be nice to see IMO would be a check for this in CI. I think it could probably be done similarly to the other cross compilation check -

I'll try to figure out how to do it, but it'll probably take some time. I've only used GitLab CI before, and GitHub Actions are a complete mystery to me. Everybody just calls somebody else's actions, and those actions call somebody else's actions, and it's difficult for me to figure out where the resulting shell commands even come from...

davidhewitt:
I think it'll be something like this:

```
      - name: Test windows python3-dll-a cross compilation
        if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.9' }}
        uses: messense/maturin-action@v1
        with:
          target: x86_64-pc-windows-msvc
          args: --release -i python3.9 --no-sdist -m examples/maturin-starter/Cargo.toml --cargo-extra-args="--features=pyo3/generate-abi3-import-lib"
```

ravenexp:
> (might need `env: PYO3_NO_PYTHON` too? Not sure if `maturin` will set it.)

That's the issue with using maturin: it has its own cross-compilation logic and `abi3` bindings handling logic, which get special treatment. It used to set `PYO3_NO_PYTHON` and alter RUSTFLAGS accordingly, but apparently it does not do that anymore after https://github.com/PyO3/maturin/pull/851 was merged.

I'd rather start with plain `cargo build --target ...` and proceed from there, if that's possible.

adamreichold:
> I'd rather start with plain cargo build --target ... and proceed from there, if that's possible.

While I understand aiming for simplicity to be able to follow things and understand errors, I also think that a CI job should be as close as possible to what actual usage will look like. Which I expect to go through Maturin most of the time (definitely, in the case where I would like to use this: building Windows wheels using MinGW in GitLab CI which I currently do by checking in the import libraries into source control). Or putting it differently, if Maturin's abi3 handling logic interferes with using this features, this might become a problem for the intended use cases which a Maturin-based CI job would only surface early (which seems a good thing to me).

davidhewitt:
cc @messense - I'd be interested in your opinion on how we might be able to test this (if you get a chance to see this considering the situation in Shanghai, no worries if you are unable to comment atm)

ravenexp:
> While I understand aiming for simplicity to be able to follow things and understand errors, I also think that a CI job should be as close as possible to what actual usage will look like. Which I expect to go through Maturin most of the time (definitely, in the case where I would like to use this: building Windows wheels using MinGW in GitLab CI which I currently do by checking in the import libraries into source control).

Well, in my tests both `cargo build` and `maturin build` commands work (for the current version of maturin =0.12.12). I wouldn't submit this if they didn't. But, maturin's cross-compilation approach changes from version to version, and it is currently in the process of eliminating its internal `PYO3_NO_PYTHON` use, which will hopefully simplify things a lot.

> Definitely, in the case where I would like to use this: building Windows wheels using MinGW in GitLab CI which I currently do by checking in the import libraries into source control).

This is my main use case as well, and this works with just `maturin build --target x86_64-pc-windows-gnu`.

My original point was that, preferably, both commands should be tested, because I've encountered the situations where one of them works and the other one doesn't, like https://github.com/PyO3/maturin/issues/824#issuecomment-1066784006. Solving issues like that is hard when no additional information is available.

adamreichold:
> My original point was that, preferably, both commands should be tested

This certainly sounds like a good idea if even more work to implement.

messense:
> I'll try to figure out how to do it, but it'll probably take some time. I've only used GitLab CI before, and GitHub Actions are a complete mystery to me.

I can work on adding tests to CI by pushing to your branch once this PR is rebased.

ravenexp:
> > I'll try to figure out how to do it, but it'll probably take some time. I've only used GitLab CI before, and GitHub Actions are a complete mystery to me.
> 
> I can work on adding tests to CI by pushing to your branch once this PR is rebased.

I've rebased the PR branch on the current `main`. This would be a huge help, thanks!

Cross and native compilation to `*-pc-windows-gnu` targets should work out of the box, and cross-compilation to `*-pc-windows-msvc` I've only tested with `rust-lld` as the linker. I don't know how to run `link.exe` on Linux.

`*-msvc` targets also need LLVM binutils for `llvm-dlltool` and the MSVC runtime import libraries to link, as usual.
I don't use the native MSVC toolchain myself, so I can't help much here.

.github/workflows/ci.yml
```diff
@@ -271,6 +271,23 @@ jobs:
           target: aarch64-apple-darwin
           args: --release -i python3.9 --no-sdist -m examples/maturin-starter/Cargo.toml
 
+      - name: Test cross compile to Windows
+        if: ${{ matrix.platform.os == 'ubuntu-latest' && matrix.python-version == '3.8' }}
```
messense, line 275: Use 3.8 rather than 3.9 to split the work to lower the total run time.

.github/workflows/ci.yml
```diff
@@ -271,6 +271,23 @@ jobs:
...
+        env:
+          XWIN_ARCH: x86_64
+        run: |
+          sudo apt-get install -y mingw-w64 llvm
+          rustup target add x86_64-pc-windows-gnu x86_64-pc-windows-msvc
+          which cargo-xwin > /dev/null || cargo install cargo-xwin
+          cargo build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-gnu
+          cargo xwin build --manifest-path examples/maturin-starter/Cargo.toml --features abi3 --target x86_64-pc-windows-msvc
```
messense, line 283: Plain `cargo build` does not work because of missing Windows CRT and SDK files for example `kernel32.lib`.

https://github.com/PyO3/pyo3/runs/5962060548?check_suite_focus=true

davidhewitt: Thanks @messense! How come this doesn't need to refer to the new feature?

guide/src/building_and_distribution.md
```diff
@@ -152,10 +152,20 @@ As your extension module may be run with multiple different Python versions you
...
-
 > Note: If you set more that one of these api version feature flags the lowest version always wins. For example, with both `abi3-py37` and `abi3-py38` set, PyO3 would build a wheel which supports Python 3.7 and up.
 
+#### Building `abi3` extensions without a Python interpreter
+
+As an advanced feature, you can build PyO3 wheel without calling Python interpreter with the environment variable `PYO3_NO_PYTHON` set.
+On Unix-like systems this works unconditionally; on Windows you must also set the `RUSTFLAGS` environment variable
+to contain `-L native=/path/to/python/libs` so that the linker can find `python3.lib`.
+
+If the `python3.dll` import library is not available, an experimental `generate-abi3-import-lib` crate
```
davidhewitt, line 163: Just noticed it would also be great to mention this in features.md

.github/workflows/ci.yml
messense: The new feature is enabled by `maturin-starter`'s `abi3` feature, see https://github.com/PyO3/pyo3/pull/2282/files#diff-074902230e9e60917cbf6dcd41751d36ff5a6f3afb24bbd5a62317eeebc8a2b6R14

davidhewitt: Aha missed that, thanks!

guide/src/building_and_distribution.md
ravenexp: Added in https://github.com/PyO3/pyo3/pull/2282/commits/4347624bd36f31f79a9d48c8868601e33a0370c6

davidhewitt (approved):
LGTM! Thanks all!

PR #2285, opened 2022-04-08
Replace unhinted spin loops by sleeping events.

adamreichold:
Nightly Clippy complains about missing spin loop hints. Instead of just adding the hints, this utilizes that `parking_lot` can be used to initialize mutexes and condition variables in const context to create a sleeping `Event` struct.


adamreichold:
The `codecov/patch` failure seems to be questionable as the const fn is arguably covered if the dynamic part of the tests is executed.

mejrs (approved):
Nooo not my spinlocks 😢 

> The codecov/patch failure seems to be questionable

That's why I went with the spinlock 😆 (also because I didn't want to deal with non-const things like std's Barrier)

davidhewitt (approved):
👍

PR #2287, opened 2022-04-08
cleanup: deprecate PyTypeObject trait

davidhewitt:
The `PyTypeObject` trait has become a bit meaningless over time. The `PyTypeInfo` trait essentially covers the same role with additional functionality; there is even a blanket implementation of `PyTypeObject for T: PyTypeInfo` which is the only implementation I'm aware of.

To make things simpler, I suggest we fold the trait into `PyTypeInfo`.

### DIFF
--- a/src/type_object.rs
+++ b/src/type_object.rs
@@ -51,8 +51,13 @@ pub unsafe trait PyTypeInfo: Sized {
     /// Utility type to make Py::as_ref work.
     type AsRefTarget: PyNativeType;
 
-    /// PyTypeObject instance for this type.
+    /// Returns the PyTypeObject instance for this type.
     fn type_object_raw(py: Python<'_>) -> *mut ffi::PyTypeObject;
+
+    /// Returns the safe abstraction over the type object.
+    fn type_object(py: Python<'_>) -> &PyType {
+        unsafe { py.from_borrowed_ptr(Self::type_object_raw(py) as _) }
+    }
 
     /// Checks if `object` is an instance of this type or a subclass of this type.
     fn is_type_of(object: &PyAny) -> bool {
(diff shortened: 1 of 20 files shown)


src/type_object.rs
```diff
@@ -65,27 +70,19 @@ pub unsafe trait PyTypeInfo: Sized {
...
-    fn type_object(py: Python<'_>) -> &PyType {
-        unsafe { py.from_borrowed_ptr(Self::type_object_raw(py) as _) }
-    }
-}
+/// This trait used to have stringent safety requirements, but they are now irrelevant as it is deprecated.
+#[deprecated(
+    since = "0.17.0",
+    note = "PyTypeObject::type_object was moved to PyTypeInfo::type_object"
+)]
+pub unsafe trait PyTypeObject: PyTypeInfo {}
```
adamreichold, line 82: Since the trait is changed in an incompatible way, is there actually a use where keep it around but deprecated instead of removing it outright would help? As written above, having trait bounds to access `type_object` needs to be updated after this. If there is any explicit implementor, it also needs changing after this. Which kind of usage would survive unchanged with the deprecated trait?

adamreichold: I think now I possibly understand it: Existing trait bounds used to access `type_object` should continue to work due to having `PyTypeInfo` as a super trait? Albeit with a deprecation warning? Implementors would still be affected, but there should not be any useful manual ones due to conflicts with the existing blanket impl?

davidhewitt: That was the idea. It's not perfect, but at least it can serve as a waypoint for users to guide them towards how to migrate.

Pushed a test demonstrating `PyTypeObject` still working.

src/type_object.rs
```diff
@@ -232,3 +229,25 @@ pub(crate) unsafe fn get_tp_free(tp: *mut ffi::PyTypeObject) -> ffi::freefunc {
         std::mem::transmute(ptr)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    #[allow(deprecated)]
```
davidhewitt, line 236: If this line is commented out and `<PyList as crate::PyTypeInfo>` replaced with just `PyList`, this is the compiler output:

```
warning: use of deprecated trait `type_object::PyTypeObject`: PyTypeObject::type_object was moved to PyTypeInfo::type_object
   --> src/type_object.rs:240:20
    |
240 |         use super::PyTypeObject;
    |                    ^^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated trait `type_object::PyTypeObject`: PyTypeObject::type_object was moved to PyTypeInfo::type_object
   --> src/type_object.rs:243:31
    |
243 |         fn get_type_object<T: PyTypeObject>(py: Python<'_>) -> &PyType {
    |                               ^^^^^^^^^^^^

error[E0599]: no function or associated item named `type_object` found for struct `PyList` in the current scope
   --> src/type_object.rs:249:58
    |
249 |                 get_type_object::<PyList>(py).is(PyList::type_object(py)))
    |                                                          ^^^^^^^^^^^ function or associated item not found in `PyList`
    |
   ::: src/types/list.rs:18:1
    |
18  | pub struct PyList(PyAny);
    | ------------------------- function or associated item `type_object` not found for this
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
237 |     use crate::type_object::PyTypeInfo;
    |
```

... hopefully this is enough for users to figure out how to fix.

adamreichold (approved):

davidhewitt:
Will wait for merging this until 0.16.4 is done.

PR #2288, opened 2022-04-09
pyo3-build-config: Use "m" ABI tag for libpython 3.7 by default

ravenexp:
According to https://bugs.python.org/issue36707, this tag is useless
since version 3.4, but also the default until version 3.8.

For example, Debian 10 ships `libpython3.7m.so`.


davidhewitt (approved):
Yep agreed 👍

This probably needs a CHANGELOG entry - something along the lines of "default to `m` abi tag when choosing libpython link name for Python 3.7 on unix", perhaps?

davidhewitt (approved):

PR #2289, opened 2022-04-10
ffi: fix segfault in _GET_TZINFO methods

davidhewitt:
Looks like datetime types without `tzinfo` set will not have space allocated for the null pointer, so we should check `has_tzinfo` before trying to read it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Default to "m" ABI tag when choosing `libpython` link name for CPython 3.7 on Unix. [#2288](https://github.com/PyO3/pyo3/pull/2288)
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


adamreichold (approved):
The change looks good to me. I have to admit to feeling uneasy about ignoring the different struct layouts between `PyDateTime_DateTime` and `PyDateTime_BaseDateTime`. Is it ensured that we never write into a datetime that was allocated by someone else and might be missing that field? Maybe we could add a comment about this to the struct definition?

pickfire (approved):
Looks good to me, just slightly worried we are not doing the cast like C.

davidhewitt:
From the datetime docs (https://docs.python.org/3/library/datetime.html#available-types):

> Objects of these types are immutable.

I think it's a bug if we _ever_ write to these things.

That said, I still agree we should be careful about the memory layout. I'll push the extra definitions.

PR #2290, opened 2022-04-10
release: 0.16.4

adamreichold:
I added a changelog entry for the coverage change because https://github.com/PyO3/rust-numpy/pull/316 revealed that this also affects downstream crates depending on PyO3.

@davidhewitt Hopefully, this reduces the busy work for you a bit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,20 +8,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.16.4] - 2022-04-14
+
 ### Added
 
+- Add `PyTzInfoAccess` trait for safe access to time zone information. [#2263](https://github.com/PyO3/pyo3/pull/2263)
 - Add an experimental `generate-abi3-import-lib` feature to auto-generate `python3.dll` import libraries for Windows. [#2282](https://github.com/PyO3/pyo3/pull/2282)
 - Add FFI definitions for `PyDateTime_BaseTime` and `PyDateTime_BaseDateTime`. [#2294](https://github.com/PyO3/pyo3/pull/2294)
-- Added `PyTzInfoAccess` for safe access to time zone information. [#2263](https://github.com/PyO3/pyo3/pull/2263)
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


davidhewitt:
Thanks, it would be cool to add #2263 to this too

adamreichold:
> Thanks, it would be cool to add #2263 to this too

Yes, I think that would make sense. Will rebase when that PR is merged.

adamreichold:
@davidhewitt Keeping with the theme of reducing busy work, I tried drafting a GitHub release for you to review before publishing. Is there a command or script you use to generate the contributors list?

davidhewitt:
Thanks!

No script; I just use the GitHub UI https://github.com/pyo3/pyo3/compare/v0.16.3...HEAD

Every time I do it I think I should write a script and then never find the time 😂

adamreichold:
> Every time I do it I think I should write a script and then never find the time joy

https://gist.github.com/adamreichold/b62493c129a55c0929b742990acb4185 appears to work. If it turns out to be helpful, maybe it can get a more permanent place of residence (and some amenities like error checking).

davidhewitt:
Nice! Yes I think it's fine to commit it somewhere to the repo.

adamreichold:
Rebased and updated the draft release.

davidhewitt:
Thanks! I didn't push this yesterday in the end because I wanted to wait until after 0.15.2 published. I couldn't remember if I pushed this and _then_ 0.15.2 whether crates.io would list 0.16.4 or 0.15.2 as the latest version.

davidhewitt:
Force-pushed to update the date.

davidhewitt:
Rebased a final time to pick up #2293

davidhewitt:
Release is live. Thanks again @adamreichold for all the work writing fixes, reviewing, and preparing this release 👍

adamreichold:
I fear we might have regressed w.r.t. finding the Python interpreter: https://github.com/PyO3/rust-numpy/runs/6029114285?check_suite_focus=true

adamreichold:
It might also be Maturin, because the part where

```
  🔗 Found pyo3 bindings with abi3 support for Python ≥ 3.7
```

seems wrong as the crate does not use the stable ABI.

PR #2293, opened 2022-04-11
pyo3-build-config: Build "abi3" extensions without an interpreter

### LINKED ISSUE #2292: Skip the build host Python interpreter detection when building portable "abi3" extensions
Continuation of the discussion in https://github.com/PyO3/maturin/pull/875

I also think PyO3 should also be doing what maturin has been doing for a long time:
make `PYO3_NO_PYTHON=1` behavior the default when compiling "abi3" extensions.

This change will help to eliminate the last remaining use of `PYO3_NO_PYTHON` by maturin
and make `cargo build` and `maturin build` behave more or less the same.

I'll add a draft PR below.

ravenexp:
Ignore the build host Python interpreter configuration when
compiling portable "abi3" extension modules.
Such extensions should not depend on a concrete Python version
or build configuration by design.

Maturin already does this when building "abi3" extension wheels
by defining `PYO3_NO_PYTHON=1` for Cargo when an `abi3-py3*`
feature is detected.

Closes #2292

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -233,6 +233,13 @@ print("mingw", get_platform().startswith("mingw"))
 "#;
         let output = run_python_script(interpreter.as_ref(), SCRIPT)?;
         let map: HashMap<String, String> = parse_script_output(&output);
+
+        ensure!(
+            !map.is_empty(),
+            "broken Python interpreter: {}",
+            interpreter.as_ref().display()
+        );
+
         let shared = map["shared"].as_str() == "True";
 
         let version = PythonVersion {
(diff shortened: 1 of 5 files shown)


pyo3-build-config/src/impl_.rs
```diff
@@ -1651,14 +1651,19 @@ pub fn make_cross_compile_config() -> Result<Option<InterpreterConfig>> {
...
 #[allow(dead_code)]
 pub fn make_interpreter_config() -> Result<InterpreterConfig> {
+    let host = Triple::host();
     let abi3_version = get_abi3_version();
 
-    if have_python_interpreter() {
+    // See if we can skip the Python interpreter configuration detection.
+    // Unix "abi3" extension modules can be built without any interpreter,
+    // so do this by default just like maturin does.
+    let require_interpreter = abi3_version.is_none() || is_linking_libpython_for_target(&host);
```
messense, line 1660: I think we should also check `python-dll-a` feature on Windows once #2282 lands?

ravenexp: Yes, that should also work. It's draft, and if there's a consensus on the related issue, I'll implement it too.

This will also require updating the user-facing documentation in many places.

davidhewitt:
The only concern is cases like `cryptography`, where I think they build using `abi3-py36` yet also need to query the interpreter to get SSL information etc. - see https://github.com/pyca/cryptography/pull/6695#discussion_r767289574 and #2092

ravenexp:
> The only concern is cases like `cryptography`, where I think they build using `abi3-py36` yet also need to query the interpreter to get SSL information etc. - see [pyca/cryptography#6695 (comment)](https://github.com/pyca/cryptography/pull/6695#discussion_r767289574) and #2092

Oh, I can now see how this will break things like https://github.com/aganders3/hello-pyo3-build-config/blob/main/build.rs when an `abi3-py3*` feature is enabled.

I'll try a different approach then: making PyO3 not fail the build when the default Python interpreter is not found or too old or otherwise unusable, but the extension module can still be built without it. It will at least prevent unnecessary build failures like this: https://github.com/PyO3/maturin/pull/875#issuecomment-1090045846.

messense (approved):

davidhewitt (approved):
👍 this new approach looks good to me, thanks!

PR #2304, opened 2022-04-14
release: 0.15.2

### LINKED ISSUE #2260: Version 0.15.2 with support for Pypy 3.9 and Python 3.6

davidhewitt:
Closes #2260


davidhewitt:
Release is live! Thanks again all.

PR #2312, opened 2022-04-17
Allow more methods to take interned arguments

mejrs:
- [x] I have yet to add some docs to these methods
- [x] Changelog


src/types/string.rs
```diff
@@ -298,6 +298,13 @@ impl<'a> IntoPy<PyObject> for &'a str {
     }
 }
 
+impl<'a> IntoPy<Py<PyString>> for &'a str {
+    #[inline]
+    fn into_py(self, py: Python<'_>) -> Py<PyString> {
+        PyString::new(py, self).into()
+    }
+}
+
```
mejrs, line 307: This is a (mild) breaking change because of inference errors, but not too bad.

adamreichold: Just to make sure I understand things correctly: The comment is informational as the other changes are breaking anyway and this is therefore targeting 0.17.0?

birkenfeld: Is the previous instance into PyObject even useful anymore?

birkenfeld:
Very nice!

src/types/string.rs
adamreichold: > Is the previous instance into PyObject even useful anymore?

I think so as `Py<PyString>` does not automatically "decay" into `Py<PyAny>` and the existing impl might be used in generic context to produce `PyObject` for stashing away.

birkenfeld: Easy to get PyObject from a Py though.

adamreichold: > Easy to get PyObject from a Py though.

I think that in generic context, one needs to be able to express this in a single trait bound, e.g. calling `into_py` twice or `into_py` followed by `into` to go from `T` to `Py<S>` to `PyObject` is not necessarily the same as going from `T` to `PyObject`.

mejrs: > Just to make sure I understand things correctly: The comment is informational as the other changes are breaking anyway and this is therefore targeting 0.17.0?

Mainly it means that things like `let thing = "thing".into_py(py);` stop compiling

davidhewitt: To avoid making lots of annoying inference errors we could potentially combine this PR with #2316, which would enable us to select on types which have `IntoPyObject<Target = PyString>` instead of `IntoPy<Py<PyString>>`.

src/types/any.rs
```diff
@@ -99,6 +99,9 @@ impl PyAny {
     /// Determines whether this object has the given attribute.
     ///
     /// This is equivalent to the Python expression `hasattr(self, attr_name)`.
+    ///
+    /// To avoid repeated temporary allocations of Python strings, the [`intern!`] macro can be used
+    /// to intern `attr_name`.
     pub fn hasattr<N>(&self, attr_name: N) -> PyResult<bool>
     where
         N: ToPyObject,
```
adamreichold, line 107: As things are, this would leave us with two different kinds of bounds for names, either `N: ToPyObject` or `N: IntoPy<Py<PyString>>`. Would it make sense to unify them?

src/types/any.rs
```diff
@@ -488,91 +523,104 @@ impl PyAny {
...
-    ///     list.call_method("sort", (), Some(kwargs))?;
-    ///     assert_eq!(list.extract::<Vec<i32>>()?, vec![7, 6, 5, 4, 3]);
+    /// # fn main() -> PyResult<()> {
+    /// Python::with_gil(|py| {
+    ///     let module = PyModule::from_code(py, CODE, "", "")?;
+    ///     let instance = module.getattr("a")?;
+    ///     let args = ("hello",);
+    ///     let kwargs = PyDict::new(py);
+    ///     kwargs.set_item("cruel", "world")?;
+    ///     let result = instance.call_method("method", args, Some(kwargs))?;
```
adamreichold, line 551: Maybe this would be a good place to showcase using the `intern!` macro to make it easier for the reader of the documentation to understand how it is supposed to be used?

adamreichold (approved):
Looks good to me.

One remaining concern that I see is that we have made several methods "trivially generic" which were not generic before. Should we aim to reduce code bloat and factor out the non-generic parts into inner help functions? On the other hand, the functions in question do not appear to be huge, so I am unsure whether this would be wort it.

src/types/any.rs
mejrs: I looked into it and I think attribute names don't need to be strings - anything will do as long as it inherits from `str`.

I would be happy to limit it to `IntoPy<Py<PyString>>`, but - are people relying on this behaviour?

For example:
```rust
use pyo3::prelude::*;

#[pyfunction]
fn getattr(item: &PyAny, arg: &PyAny) -> PyResult<()> {
    item.getattr(arg)?;
    Ok(())
}

/// This module is implemented in Rust.
#[pymodule]
fn my_module(py: Python<'_>, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(getattr, m)?)?;
    Ok(())
}
```

```python
import my_module

class A:
    def __getattr__(self, attr):
        print(attr)
        print(attr.__class__)
        return "foo"

a = A()

class C(str):
    pass

my_module.getattr(a, "field")
my_module.getattr(a, C())
```

davidhewitt: I think `IntoPy<Py<PyString>>` makes sense for attribute names. Edge cases like this don't seem likely to me, and I think with a bit of type wrangling (e.g. change `arg: &PyAny` to `arg: &PyString`) things should still work.

mejrs:
> Should we aim to reduce code bloat and factor out the non-generic parts into inner help functions?

They are pretty small, I think it's fine.

> Would it make sense to unify them?

Done 👍

davidhewitt:
Looks good to me too. The only question I have outstanding is whether we want to merge this now or wait until we can also merge with #2316 so that we don't cause users to go through extra churn of type inference failures in the interim.

(FWIW I have a bad feeling that #2316 gets a little upset with some of the more gnarly `#[pyproto]` code, so that might be blocked until 0.18 when we can kill the feature off entirely.)

mejrs:
I'm feeling that what we're doing with traits is a bit ad-hoc.

Maybe we need to sit down and think about what we don't like, what we would like, and what a good path to getting there is.

src/types/any.rs
mejrs: Done

davidhewitt:
> I'm feeling that what we're doing with traits is a bit ad-hoc.
> 
> Maybe we need to sit down and think about what we don't like, what we would like, and what a good path to getting there is.

I agree very much with this sentiment. I'll write my current backlog of ideas up in a separate issue as a starting point.

As for this PR... looks good to me! The concern I have with merging before we make a decision on #2316 is that this may cause users to have to add `: PyObject` annotations to keep existing `.into_py()` uses compiling. If we then merge #2316 these annotations may then become type errors. (I guess #2316 will probably break a bit of existing code too, so maybe that's not actually a reason to hold back this PR?) What's your opinion on this?

src/instance.rs
```diff
@@ -657,20 +657,28 @@ impl<T> Py<T> {
...
+        let args: Py<PyTuple> = args.into_py(py);
+
+        let ptr = self.getattr(py, name)?.into_ptr();
+        let args = args.into_ptr();
+        let kwargs = kwargs.into_ptr();
+
         unsafe {
-            let args = args.into_py(py).into_ptr();
-            let kwargs = kwargs.into_ptr();
-            let ptr = ffi::PyObject_GetAttr(self.as_ptr(), name.to_object(py).as_ptr());
```
mejrs, line 670: Sorry this is a bit messy since resolving the merge conflict.

How do you feel about:

```rust
pyo3::ffi::Something(name.to_object(py).as_ptr());
```

versus 

```rust
let ptr =  name.to_object(py).into_ptr();
pyo3:;ffi::Something(ptr);
ffi::Py_DECREF(ptr);
```
(or something else?)

Personally I prefer the latter. IMO the former is error prone and gives a bad example.

mejrs:
>  What's your opinion on this?

I think that's fine, it's breaking either way.

src/instance.rs
```diff
@@ -682,24 +690,32 @@ impl<T> Py<T> {
...
+    where
+        N: IntoPy<Py<PyString>>,
+    {
         cfg_if::cfg_if! {
             if #[cfg(all(Py_3_9, not(any(Py_LIMITED_API, PyPy))))] {
                 // Optimized path on python 3.9+
                 unsafe {
-                    let name = name.into_py(py);
+                    let name: Py<PyString> = name.into_py(py);
                     PyObject::from_owned_ptr_or_err(py, ffi::PyObject_CallMethodNoArgs(self.as_ptr(), name.as_ptr()))
```
davidhewitt, line 719: Exactly, like this is how I think we should be making use of `Py` to avoid leaking memory 👍

src/types/any.rs
```diff
@@ -164,35 +174,40 @@ impl PyAny {
...
-                py,
-                ffi::PyObject_SetAttr(
-                    self.as_ptr(),
-                    attr_name.to_object(py).as_ptr(),
-                    value.to_object(py).as_ptr(),
-                ),
-            )
+            let ret = ffi::PyObject_SetAttr(self.as_ptr(), attr_name, value);
+            ffi::Py_DECREF(attr_name);
+            ffi::Py_XDECREF(value);
```
davidhewitt, line 188: e.g. here there's no need for `XDECREF` instead of `DECREF` because it's got to be a non-null pointer (it came via `to_object` call).

davidhewitt (commented):
Ok, let's head towards merging this in 0.17 and maybe thinking about `IntoPyObject` for 0.18. I'll try to write up my thoughts on traits tomorrow.

Got to confess I'm really not a fan of calling `Py_DECREF` or `Py_XDECREF` whenever we can just let `Py` do it in `Drop` implementation.

src/instance.rs
mejrs: > Why do you feel the former is error prone? I think the second is more error prone because missing the `Py_DECREF` call is leaking memory just like missing `free()` in C.

I'm mostly worried about someone reading it and and writing it slightly differently:
```rust
let ptr = name.to_object(py).as_ptr();
pyo3::ffi::Something(ptr);
```
which produces a dangling pointer, as the python object is dropped at the end of the statement (I should add documentation about this).

> Perhaps a middle ground is to make it obvious we're using RAII through `Py`?
> 
> ```rust
> let name: PyObject = name.to_object(py);
> pyo3::ffi::Something(name.as_ptr());
> ```

Sure 👍

davidhewitt: > 'm mostly worried about someone reading it and and writing it slightly differently:

This is an excellent point which I'd totally missed. IMO the middle ground above is probably the only sensible way to arrange this then. If you like, I'll submit a PR to fix the whole codebase to use this pattern once this PR merged? That way you can just worry about the modified functions here without another merge conflict.

davidhewitt (commented):
👍 Looks great, thanks very much for all the tidy ups!

A final thought: as we're committing to merging this and producing extra inference errors in PyO3 0.17, can we please add a section to `migration.md` to explain this to users?

mejrs:
> A final thought: as we're committing to merging this and producing extra inference errors in PyO3 0.17, can we please add a section to `migration.md` to explain this to users?

Sure.

I'm also concerned about some missing codecov here, so I want to add a couple of tests as well.

mejrs:
Should be done now :)

davidhewitt:
Thanks for merging, looks great, sorry I had a busy last few days and didn't get around to looking at this sooner. 👍

PR #2313, opened 2022-04-18
Implement ToPyObject for [T; N]

PigeonF:
It seems like this was forgotten to implement (unless there was a good reason not to implement this, which I did not find in my quick search).

Related: fusion-engineering/inline-python#51

### DIFF
--- a/src/conversions/array.rs
+++ b/src/conversions/array.rs
@@ -14,6 +14,15 @@ mod min_const_generics {
         }
     }
 
+    impl<T, const N: usize> ToPyObject for [T; N]
+    where
+        T: ToPyObject,
+    {
+        fn to_object(&self, py: Python<'_>) -> PyObject {
+            self.as_ref().to_object(py)
+        }
+    }
+
     impl<'a, T, const N: usize> FromPyObject<'a> for [T; N]
     where
         T: FromPyObject<'a>,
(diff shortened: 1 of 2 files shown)


adamreichold:
CI failure is unrelated and affects `main` as well ATM.

mejrs (approved):
Thanks!

I'm not sure why this was missing. I suppose in most cases, like when calling the method on an array, it can find the `&[T]` implementation - but `inline-python` uses it in a generic context where this does not work.

src/conversions/array.rs
```diff
@@ -213,6 +231,19 @@ mod tests {
             assert!(&v == b"abc");
         });
     }
+    #[test]
+    fn test_topyobject() {
```
mejrs, line 235: nit: This should have a more descriptive name.

PigeonF: Should I change this and force push, or just add a new commit?

mejrs: I don't mind either way, but I'd lean towards force pushing.

PigeonF: Naming things is hard, I hope you like the new one better :^)

konstin:
> It seems like this was forgotten to implement (unless there was a good reason not to implement this, which I did not find in my quick search).

`impl<T, const N: usize>` was only stabilized in rust 1.51, i'd have loved to have done this earlier! i'm surprised though we didn't have any impl for small slice lengths, since for tuples we have the `tuple_conversion` macro which implements tuples up to size 12.

messense:
Can we use `#[rustversion::since(1.51)] ` to guard it? Similar to https://github.com/PyO3/pyo3/blob/16ad15e04fd4dd283ef3ffb7e45d68c12637de96/xtask/src/utils.rs#L11

birkenfeld:
It's already guarded by a feature flag, which is set by build.rs depending on version. `rustversion` is not used as a non-dev dependency.

davidhewitt (approved):
Yep I think this was just an accidental omission 👍

davidhewitt:
Rebasing on `main` should fix the 3.11 error

mejrs:
Thanks!

PR #2316, opened 2022-04-19
Add new trait IntoPyObject

davidhewitt:
This is throwing out a concept that I've been pondering for a while to progress with the design of the to-python conversion traits.

It replaces `IntoPy<T>` generic trait with a trait `IntoPyObject` which has an associated type defined as follows:

```rust
pub trait IntoPyObject: Sized {
    type Target;

    /// Converts to the known target Python type.
    fn into_py(self, py: Python<'_>) -> Py<Self::Target>;

    /// Converts to a generic Python type. Equivalent to `self.into_py().into()`.
    fn into_object(self, py: Python<'_>) -> PyObject;
}
```

The names are intended to line up with the `ToPyObject` trait which has the `to_object` method. We could also have `TryIntoPyObject` trait, potentially, with a similar design.

The advantages I see of doing this:
- Users can call `into_py()` and get `Py<T>` rather than `Py<PyAny>`, helping avoid to need an immediate cast to work with the Python object.
- Avoids inference errors caused when multiple implementations of `IntoPy<T>` exist for a given type.
- Opens the way for us to have `#[derive(IntoPyObject)]`.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -184,6 +184,98 @@ pub trait IntoPy<T>: Sized {
     fn into_py(self, py: Python<'_>) -> T;
 }
 
+/// Defines a conversion from a Rust type to a Python object.
+///
+/// It functions similarly to std's [`Into`](std::convert::Into) trait,
+/// but requires a [GIL token](Python) as an argument.
+/// Many functions and traits internal to PyO3 require this trait as a bound,
+/// so a lack of this trait can manifest itself in different error messages.
+///
+/// # Examples
+/// ## With `#[pyclass]`
+/// The easiest way to implement `IntoPy` is by exposing a struct as a native Python object
+/// by annotating it with [`#[pyclass]`](crate::prelude::pyclass).
+///
+/// ```rust
+/// use pyo3::prelude::*;
+///
+/// #[pyclass]
+/// struct Number {
+///     #[pyo3(get, set)]
+///     value: i32,
+/// }
+/// ```
+/// Python code will see this as an instance of the `Number` class with a `value` attribute.
+///
+/// ## Conversion to a Python object
+///
+/// However, it may not be desirable to expose the existence of `Number` to Python code.
+/// `IntoPy` allows us to define a conversion to an appropriate Python object.
+/// ```rust
+/// use pyo3::prelude::*;
+///
+/// struct Number {
+///     value: i32,
+/// }
+///
+/// impl IntoPy<PyObject> for Number {
+///     fn into_py(self, py: Python<'_>) -> PyObject {
+///         // delegates to i32's IntoPy implementation.
+///         self.value.into_py(py)
+///     }
+/// }
+/// ```
+/// Python code will see this as an `int` object.
+///
+/// ## Dynamic conversion into Python objects.
+/// It is also possible to return a different Python object depending on some condition.
+/// This is useful for types like enums that can carry different types.
+///
+/// ```rust
+/// use pyo3::prelude::*;
+///
+/// enum Value {
+///     Integer(i32),
+///     String(String),
+///     None,
+/// }
+///
+/// impl IntoPy<PyObject> for Value {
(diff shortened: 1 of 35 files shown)


tests/test_buffer_protocol.rs
```diff
@@ -112,7 +112,7 @@ fn test_buffer_referenced() {
         let input = vec![b' ', b'2', b'3'];
         let gil = Python::acquire_gil();
         let py = gil.python();
-        let instance: PyObject = TestBufferClass {
+        let instance = TestBufferClass {
```
davidhewitt, line 115: Users migrating might have to remove a number of `: PyObject` annotations like this.

tests/test_class_conversion.rs
```diff
@@ -122,9 +122,9 @@ fn test_polymorphic_container_does_not_accept_other_types() {
 
     let setattr = |value: PyObject| p.as_ref(py).setattr("inner", value);
 
-    assert!(setattr(1i32.into_py(py)).is_err());
+    assert!(setattr(1i32.into_object(py)).is_err());
```
davidhewitt, line 125: Users migrating might have to replace `into_py` with `into_object` where they to want `PyObject` output. I'm not sure how annoying this might be.

Worst case I think we instruct users to replace all `.into_py()` with `.into_object()`, and then they can case-by-case change back to `into_py()` where the better typing is helpful.

src/conversion.rs
```diff
@@ -431,6 +534,13 @@ impl IntoPy<Py<PyTuple>> for () {
     }
 }
 
+impl IntoPyObject for () {
```
davidhewitt, line 537: Note that `IntoPy<PyObject> for ()` becomes `None`, and `IntoPy<Py<PyTuple>>` becomes `()`.

IMO this is a bit annoying, and `().into_py()` _should_ be an empty tuple 

However without specialization, we need to do some fudging in the macros to fix the `()` case as a `#[pyfunction]` return value. I'm pretty sure I know how to do this.

davidhewitt (commented):
Some compatibility comments before I push this too far, input from others very much welcomed!

birkenfeld:
I like it a lot. I agree that the `()` -> `None` transform only should be made for function returns.

src/conversion.rs
adamreichold: To me, it would seem preferable to fix the special case of the type of the return value of functions that do not return anything instead of ingraining this into the conversion traits.

src/conversion.rs
```diff
@@ -184,6 +184,98 @@ pub trait IntoPy<T>: Sized {
...
+/// Python code will see this as any of the `int`, `string` or `None` objects.
+#[cfg_attr(docsrs, doc(alias = "IntoPyCallbackOutput"))]
+pub trait IntoPyObject: Sized {
+    type Target;
+
+    /// Performs the conversion.
+    fn into_py(self, py: Python<'_>) -> Py<Self::Target>;
+
+    /// Performs the conversion.
+    fn into_object(self, py: Python<'_>) -> PyObject {
```
mejrs, line 274:
```suggestion
    fn into_object(self, py: Python<'_>) -> Py<PyAny> {
```

I think this communicates the difference between these methods better.

src/conversion.rs
davidhewitt: @adamreichold I think that we're saying the same thing? At the moment the conversion traits encode this special case, however really this should just be part of the `#[pyfunction]` / `#[pymethods]` macro. (With `min_specialization` it could probably be part of the `IntoPyCallbackOutput` trait in the future, however I won't hold my breath.)

adamreichold: Yes, certainly. I just wanted to echo

> IMO this is a bit annoying, and ().into_py() should be an empty tuple

which sounded like you did not fully convince yourself yet.

davidhewitt:
Closing as this is very outdated and #4041 is an attempt to kickstart future discussion in this space.

PR #2319, opened 2022-04-19
Add threading test

mejrs:
Continues https://github.com/PyO3/pyo3/pull/2314

### DIFF
--- a/src/marker.rs
+++ b/src/marker.rs
@@ -1015,4 +1015,30 @@ mod tests {
         let state = unsafe { crate::ffi::PyGILState_Check() };
         assert_eq!(state, GIL_NOT_HELD);
     }
+
+    #[test]
+    fn test_threading_id() {
+        const SOURCE: &str = r#"
+import threading
+
+current = threading.current_thread()
+main = threading.main_thread()
+
+assert main is not current, (main, current)
+"#;
+
+        // If somehow the interpreter has not yet been initialized,
+        // this will initialize it
+        Python::with_gil(|_| {});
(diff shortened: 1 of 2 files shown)


mejrs:
If the `threading` module is imported first, these tests succeed (at least on a6).

Interestingly, `with_embedded_python_interpreter` has this:

```rust
// Import the threading module - this ensures that it will associate this thread as the "main"
// thread, which is important to avoid an `AssertionError` at finalization.
pool.python().import("threading").unwrap();
```

Do we need this in `prepare_freethreaded_python` as well?

davidhewitt:
> Do we need this in `prepare_freethreaded_python` as well?

Ah, interesting. I previously decided "no", because:
 - We don't really have information whether this thread or any other is correct to think of as the "main thread". (I think PyO3's test suite is a good example where there isn't a main thread.)
 - We don't expect to ever finalize when initializing with `prepare_freethreaded_python()`, so the workaround there is not relevant.

I could be persuaded that the thread which calls `prepare_freethreaded_python()` is more likely to be the "main thread" than whatever thread happens to call `import threading` first. What's your opinion?

---

Actually I think in Python 3.11 it's also actually no longer necessary as a workaround ever - https://github.com/python/cpython/pull/28549. However I think it probably _is_ still correct to think of the thread calling `with_embedded_python_interpreter()` a "main thread".

mejrs:
> What's your opinion?

Hmm, I feel that could be surprising.

Let's just close this then 👍

Issue #2323, opened 2022-04-23
IntoPyCallbackOutput<_> is not implimented for arrays of pyclass


andyblarblar:
### Bug Description

I'm attempting to use a pyclass annotated struct array as the member of another pyclass annotated struct. When I attempt to annotate that array feild with `pyo3(get)`, it fails to compile. I'm building off of the current master (dea9eb7a) as it's latest commit seems to cover this use case.

### Steps to Reproduce

Minimal example:
```rust 
#[pyclass]
#[derive(Copy, Clone)]
struct Foo {
    x: i32
}

#[pyclass]
struct Bar {
    #[pyo3(get)]
    arr: [Foo; 12]
}
```

### Backtrace

```shell
error[E0277]: the trait bound `[Foo; 12]: IntoPyCallbackOutput<_>` is not satisfied
   --> src/ld06_driver.rs:21:1
    |
21  | #[pyclass]
    | ^^^^^^^^^^ the trait `IntoPyCallbackOutput<_>` is not implemented for `[Foo; 12]`
    |
    = help: the following other types implement trait `IntoPyCallbackOutput<Target>`:
              <() as IntoPyCallbackOutput<()>>
              <() as IntoPyCallbackOutput<i32>>
              <*mut PyObject as IntoPyCallbackOutput<*mut PyObject>>
              <HashCallbackOutput as IntoPyCallbackOutput<isize>>
              <IterANextOutput<Py<PyAny>, Py<PyAny>> as IntoPyCallbackOutput<*mut PyObject>>
              <IterANextOutput<T, U> as IntoPyCallbackOutput<IterANextOutput<Py<PyAny>, Py<PyAny>>>>
              <IterNextOutput<Py<PyAny>, Py<PyAny>> as IntoPyCallbackOutput<*mut PyObject>>
              <IterNextOutput<T, U> as IntoPyCallbackOutput<IterNextOutput<Py<PyAny>, Py<PyAny>>>>
            and 7 others
note: required by a bound in `pyo3::callback::convert`
   --> /home/andy/.cargo/git/checkouts/pyo3-a22e69bc62b9f0fd/dea9eb7/src/callback.rs:182:8
    |
182 |     T: IntoPyCallbackOutput<U>,
    |        ^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `pyo3::callback::convert`
```
```


### Your operating system and version

Ubuntu 20.04

### Your Python version (`python --version`)

Python 3.8

[...]

PR #2326, opened 2022-04-23
Fix IntoPyCallbackOutput paper cuts

mejrs:
- Changes `impl<T, const N: usize> IntoPy<PyObject> for [T; N]`  to use `IntoPy` for it elements instead. Fixes https://github.com/PyO3/pyo3/issues/2323
- Adds a new trait to result-wrap function outputs which runs before `IntoPyCallbackOutput `, which (generally) generates a better compile error.
For example, consider this:
```rust
struct Blah;

#[pyfunction]
fn blah() -> Blah{
    Blah
}
```

Current error:
```
error[E0277]: the trait bound `Blah: IntoPyCallbackOutput<_>` is not satisfied
   --> src\lib.rs:5:1
    |
5   | #[pyfunction]
    | ^^^^^^^^^^^^^ the trait `IntoPyCallbackOutput<_>` is not implemented for `Blah`
    |
note: required by a bound in `pyo3::callback::convert`
   --> C:\Users\bruno\rust\pyo3\src\callback.rs:182:8
    |
182 |     T: IntoPyCallbackOutput<U>,
    |        ^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `pyo3::callback::convert`
    = note: this error originates in the attribute macro `pyfunction` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
```

Error with this PR:
```
error[E0599]: the method `into_result` exists for struct `Blah`, but its trait bounds were not satisfied
   --> src\lib.rs:5:1
    |
3   | struct Blah;
    | ------------
    | |
    | method `into_result` not found for this
    | doesn't satisfy `Blah: IntoPy<Py<PyAny>>`
    | doesn't satisfy `Blah: IntoResult<Result<Blah, PyErr>>`
4   |
5   | #[pyfunction]
    | ^^^^^^^^^^^^^ method cannot be called on `Blah` due to unsatisfied trait bounds
    |
    = note: the following trait bounds were not satisfied:
            `Blah: IntoPy<Py<PyAny>>`
            which is required by `Blah: IntoResult<Result<Blah, PyErr>>`
            `&Blah: IntoPy<Py<PyAny>>`
            which is required by `&Blah: IntoResult<Result<&Blah, PyErr>>`
            `&mut Blah: IntoPy<Py<PyAny>>`
            which is required by `&mut Blah: IntoResult<Result<&mut Blah, PyErr>>`
note: the following trait must be implemented
   --> C:\Users\bruno\rust\pyo3\src\conversion.rs:182:1
    |
182 | / pub trait IntoPy<T>: Sized {
183 | |     /// Performs the conversion.
184 | |     fn into_py(self, py: Python<'_>) -> T;
185 | | }
    | |_^
    = note: this error originates in the attribute macro `pyfunction` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0599`.
```

Todo:

- [x]  Changelog
- [x] Tests
- [x] Some ui tests


src/callback.rs
```diff
@@ -35,6 +35,30 @@ impl PyCallbackOutput for () {
     const ERR_VALUE: Self = ();
 }
 
+#[doc(hidden)]
+pub trait IntoResult<R> {
+    fn into_result(self) -> R;
+}
```
davidhewitt, line 41: 👍 I think this is a good idea. It might take some refactoring to fully remove result-wrapping from the `IntoPyCallbackOutput` trait. IIRC it was very much needed for `#[pyproto]` macros, however maybe there's a way. Or the full separation could wait to 0.18 even if we merge this now.

src/callback.rs
```diff
@@ -35,6 +35,30 @@ impl PyCallbackOutput for () {
...
+
+impl<T> IntoResult<PyResult<T>> for T
```
davidhewitt, line 43: Do you think the trait should potentially just be `IntoPyResult<T>`, and the method `into_py_result()`?

That'll have the effect of making error messages clearly related to PyO3.

src/callback.rs
```diff
@@ -35,6 +35,30 @@ impl PyCallbackOutput for () {
...
+where
+    T: IntoPy<PyObject>,
+{
+    fn into_result(self) -> PyResult<Self> {
```
davidhewitt, line 47:
```suggestion
    #[inline]
    fn into_result(self) -> PyResult<Self> {
```

src/callback.rs
```diff
@@ -35,6 +35,30 @@ impl PyCallbackOutput for () {
...
+        Ok(self)
+    }
+}
+
+impl<T, E> IntoResult<PyResult<T>> for Result<T, E>
+where
+    T: IntoPy<PyObject>,
+    E: Into<PyErr>,
+{
+    fn into_result(self) -> PyResult<T> {
```
davidhewitt, line 57:
```suggestion
    #[inline]
    fn into_result(self) -> PyResult<T> {
```

src/conversions/array.rs
```diff
@@ -3,14 +3,39 @@ use crate::{exceptions, PyErr};
...
-    use crate::{FromPyObject, IntoPy, PyAny, PyObject, PyResult, PyTryFrom, Python, ToPyObject};
+    use crate::conversion::IntoPyPointer;
+    use crate::{
+        ffi, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, PyTryFrom, Python, ToPyObject,
+    };
 
     impl<T, const N: usize> IntoPy<PyObject> for [T; N]
     where
-        T: ToPyObject,
+        T: IntoPy<PyObject>,
```
davidhewitt, line 13: Yeah the original bound was definitely an error 👍

davidhewitt (commented):
Thanks! 

First commit message - "Make arrays of pyclasses easier to **sue**" - oops :)

src/callback.rs
mejrs: > It might take some refactoring to fully remove result-wrapping from the IntoPyCallbackOutput trait.

I did try doing this, but got a lot of errors from code related to pyproto 😰

src/callback.rs
mejrs: Does this even do anything on a generic function? Or on a trait implementation?

davidhewitt: Hmmm good question. Looking at https://matklad.github.io/2021/07/09/inline-in-rust.html

.... I think because this trait is generic this is necessarily a generic function, so implicitly inlineable anyway?

src/callback.rs
mejrs: Just to clarify - the only allowed return types for pyo3 functions and methods are:

```
Result<T, E>
where
    T: IntoPy<PyObject>,
    E: Into<PyErr>,
```
and
```
T
where
    T: IntoPy<PyObject>,
```
right?

I don't see it anywhere written in the guide, and CI doesn't seem to complain, but I'd like to make sure.

davidhewitt: Yes (except for `#[new]` and magic methods like `__hash__` which return specific types).

src/callback.rs
```diff
@@ -35,6 +35,31 @@ impl PyCallbackOutput for () {
     const ERR_VALUE: Self = ();
 }
 
+/// A helper trait to result-wrap the return values of functions and methods.
+#[doc(hidden)]
+pub trait IntoPyResult<R> {
+    fn into_py_result(self) -> R;
+}
```
davidhewitt, line 42: I was thinking that we'd restrict the return value to `PyResult`:

```suggestion
pub trait IntoPyResult<T> {
    fn into_py_result(self) -> PyResult<T>;
}
```

Was there a reason you preferred not to do so?

src/callback.rs
```diff
@@ -35,6 +35,31 @@ impl PyCallbackOutput for () {
     const ERR_VALUE: Self = ();
 }
 
+/// A helper trait to result-wrap the return values of functions and methods.
```
davidhewitt, line 38: For new code which is only public because it's proc-macro implementation detail, I've been shoving it somewhere underneath `src/impl_`. Think it makes sense to put this new trait in there?

src/callback.rs
davidhewitt, line 42: I was thinking that we'd restrict the return value to `PyResult`:

```suggestion
pub trait IntoPyResult<T> {
    fn into_py_result(self) -> PyResult<T>;
}
```

Was there a reason you preferred not to do so?

src/conversions/array.rs
```diff
@@ -149,17 +174,69 @@ mod min_const_generics {
 #[cfg(not(min_const_generics))]
```
davidhewitt, line 174: Unnngh just another year until we bump msrv and throw this one away 😅. Nicely done for figuring out the `ArrayGuard` technique!

src/conversions/array.rs
```diff
@@ -149,17 +174,69 @@ mod min_const_generics {
...
+                            // We create the  `Py` pointer here for two reasons:
+                            // - panics if the ptr is null
+                            // - its Drop cleans up the list if user code panics.
+                            let list: Py<PyAny> = Py::from_owned_ptr(py, ptr);
+
+                            let slf = ManuallyDrop::new(self);
+
+                            let mut guard = ArrayGuard{
+                                // the transmute size check is _very_ dumb around generics
+                                elements: transmute_copy(&slf),
```
davidhewitt, line 220: I'm curious why the `ManuallyDrop::new` followed by `transmute_copy`, instead of just `elements: transmute(self)`?

tests/ui/invalid_result_conversion.stderr
```diff
@@ -1,14 +1,12 @@
...
+  --> tests/ui/invalid_result_conversion.rs:21:1
+   |
+21 | #[pyfunction]
+   | ^^^^^^^^^^^^^ method cannot be called on `Result<(), MyError>` due to unsatisfied trait bounds
+   |
+   = note: the following trait bounds were not satisfied:
+           `&Result<(), MyError>: IntoPy<Py<PyAny>>`
+           which is required by `&Result<(), MyError>: IntoPyResult<Result<&Result<(), MyError>, PyErr>>`
+           `&mut Result<(), MyError>: IntoPy<Py<PyAny>>`
+           which is required by `&mut Result<(), MyError>: IntoPyResult<Result<&mut Result<(), MyError>, PyErr>>`
```
davidhewitt, line 11: Yeah I think that having `IntoPyResult<T>` as suggested above would make the errors here even nicer, because it'd remove one level of `Result<_, PyErr>` from these messages.

davidhewitt (commented):
Really nice! Just a few final thoughts, and I think it would be better to change from `IntoPyResult<R>` -> `IntoPyResult<T>`. After that, I think this is good to merge 👍

src/conversions/array.rs
mejrs: It doesn't work - you have to use `transmute_copy` because it doesn't have `transmute`'s size check. For example...

```rust
fn into<const N: usize>(input: [u8; N]) -> [i8; N] {
    unsafe { std::mem::transmute(input) }
}
```
...does not work.
```
 --> src\main.rs:8:14
  |
8 |     unsafe { std::mem::transmute(input) }
  |              ^^^^^^^^^^^^^^^^^^^
  |
  = note: source type: `[u8; N]` (this type does not have a fixed size)
  = note: target type: `[i8; N]` (this type does not have a fixed size)
```

davidhewitt: Hmmm, is that a compiler bug? An array has a fixed size 🤔

Either way, makes sense, let's stick with `transmute_copy`!

src/impl_/ghost.rs
```diff
@@ -0,0 +1,19 @@
...
+}
+
+impl<T> IntoPyResult<T> for T where T: IntoPy<PyObject> {}
+
+impl<T, E> IntoPyResult<T> for Result<T, E>
+where
+    T: IntoPy<PyObject>,
+    E: Into<PyErr>,
+{
+}
```
mejrs, line 19: I prefer this approach of code that doesn't actually do anything 😉

tests/ui/missing_intopy.stderr
```diff
@@ -0,0 +1,38 @@
...
+    |
+3   | #[pyo3::pyfunction]
+    | ^^^^^^^^^^^^^^^^^^^ the trait `IntoPyCallbackOutput<_>` is not implemented for `Blah`
+    |
+note: required by a bound in `pyo3::callback::convert`
+   --> src/callback.rs
+    |
+    |     T: IntoPyCallbackOutput<U>,
+    |        ^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `pyo3::callback::convert`
+    = note: this error originates in the attribute macro `pyo3::pyfunction` (in Nightly builds, run with -Z macro-backtrace for more info)
```
mejrs, line 38: It's unfortunate this `IntoPyCallbackOutput` part is back - but overall I think it's worth it. The first part of the error is very very good.

src/impl_/ghost.rs
```diff
@@ -0,0 +1,19 @@
+/// If it does nothing, was it ever really there? 👻
+///
+/// This is code that is just type checked to e.g. create better compile errors,
+/// but that never affects anything at runtime,
+use crate::{IntoPy, PyErr, PyObject};
+
+#[allow(clippy::wrong_self_convention)]
```
davidhewitt, line 7: Is this lint actually necessary? I wonder if it's a hangover from experimentation.

davidhewitt (approved):
Sneaky! Nicely done 👍

src/impl_/ghost.rs
davidhewitt: Ah I see! Do you think we should name the method differently then? E.g. `into_py_result_hint` or `assert_into_py_result`?

davidhewitt (approved):
Thanks, LGTM! To squash out the various interim commits I'm gonna squash-merge this one 👍

Issue #2324, opened 2022-04-23
Install Failures on Termux


xloem:
### Bug Description

I was unable to install a pyo3 pip package on termux without manually providing a config file. For some reason the library parameters were not set. From https://github.com/PyO3/pyo3/issues/1800#issuecomment-1106749656

### Steps to Reproduce

1. Install termux on android
2. In termux, install python and rust: `pkg install python rust`
3. In termux, install based58: `pip3 install based58` or manually from https://github.com/kevinheavey/based58 (`pkg install git`)

### Backtrace

```
$ PYO3_PRINT_CONFIG=1 maturin build
🔗 Found pyo3 bindings with abi3 support for Python ≥ 3.7
🐍 Not using a specific python interpreter (With abi3, an interpreter is only required on windows)
📦 Built source distribution to /data/data/com.termux/files/home/srctmp/based58/target/wheels/based58-0.1.0.tar.gz
   Compiling pyo3 v0.15.1
   Compiling pyo3-macros v0.15.1
error: failed to run custom build command for `pyo3 v0.15.1`

Caused by:
  process didn't exit successfully: `/data/data/com.termux/files/home/srctmp/based58/target/debug/build/pyo3-c92d6bef7e401530/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=PYO3_PRINT_CONFIG

  -- PYO3_PRINT_CONFIG=1 is set, printing configuration and halting compile --
  implementation=CPython
  version=3.7
  shared=true
  abi3=true
  build_flags=WITH_THREAD
  suppress_build_script_link_lines=false
warning: build failed, waiting for other jobs to finish...
error: build failed
💥 maturin failed
  Caused by: Failed to build a native library through cargo
  Caused by: Cargo build finished with "exit status: 101": `cargo rustc --manifest-path Cargo.toml --message-format json --lib --`
```

### Your operating system and version

[...]

PR #2328, opened 2022-04-24
pyo3-build-config: For `abi3`, link to `libpython3.so` on Unix

ravenexp:
Quoting PEP 384:

 If Python is compiled as a shared library, it is installed as both
 libpython3.so, and libpython3.y.so; applications conforming to this PEP
 should then link to the former (extension modules can continue to link
 with no libpython shared object, but rather rely on runtime linking).

Source: https://peps.python.org/pep-0384/#linkage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Move `PyTypeObject::type_object` method to `PyTypeInfo` trait, and deprecate `PyTypeObject` trait. [#2284](https://github.com/PyO3/pyo3/pull/2284)
 - The deprecated `pyproto` feature is now disabled by default. [#2321](https://github.com/PyO3/pyo3/pull/2321)
+- Link to `libpython3.so` on Unix-like targets when `abi3` feature is enabled. [#2328](https://github.com/PyO3/pyo3/pull/2328)
 
 ## [0.16.4] - 2022-04-14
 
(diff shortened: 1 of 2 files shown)


ravenexp:
Previous discussion: https://github.com/PyO3/pyo3/issues/2324#issuecomment-1107514760

ravenexp:
Ok, I'm not sure if `-lpython3` even works with modern gcc toolchains...

When compiling a C program that  only calls `Py_Initialize()`:

```
$ gcc test.c -lpython3
/usr/bin/ld: /tmp/ccQwlIiy.o: undefined reference to symbol 'Py_Initialize'
/usr/bin/ld: /usr/lib/libpython3.10.so.1.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

Adding [a workaround](https://stackoverflow.com/questions/8725572/gcc-4-5-vs-4-4-linking-with-dependencies) makes it kinda work:

```
$ gcc test.c -Wl,--copy-dt-needed-entries -lpython3
$ ./a.out
$ ldd a.out
        linux-vdso.so.1 (0x00007ffc38b86000)
        libpython3.so => /usr/lib/libpython3.so (0x00007f19c8d47000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f19c8b3d000)
        libpython3.10.so.1.0 => /usr/lib/libpython3.10.so.1.0 (0x00007f19c8782000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f19c8d6f000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f19c869a000)
```

but, it only results in the executable linking to both `libpython3.so` _and_ `libpython3.10.so`.

I'm going to conclude that this part of PEP 384 is dead on arrival, because I couldn't find anyone complaining about it not working as intended except for https://bugs.python.org/issue11347.

Issue #2349, opened 2022-05-02
Class attributes can be mutated from Python code


mejrs:
The guide [states](https://pyo3.rs/v0.16.4/class.html#class-attributes):
> Note that unlike class variables defined in Python code, class attributes defined in Rust cannot be mutated at all

But this is in fact possible:

```rust
use pyo3::prelude::*;

#[pyclass]
struct MyClass {}

#[pymethods]
impl MyClass {
    #[classattr]
    fn my_attribute() -> String {
        "hello".to_string()
    }
}
fn main() {
    Python::with_gil(|py| {
        let my_class = py.get_type::<MyClass>();

        // All good...
        pyo3::py_run!(py, my_class, "assert my_class.my_attribute == 'hello'");

        // OK
        pyo3::py_run!(py, my_class, "my_class.my_attribute = 'foo'");

        // Also OK
        pyo3::py_run!(py, my_class, "assert my_class.my_attribute == 'foo'");
    });
}

```

davidhewitt:
Hmm interesting. I suspect this is a behavioural change we missed when we switched to "heap types" to support abi3 - see https://pyo3.rs/v0.16.4/migration.html#runtime-changes-to-support-the-cpython-limited-api

> Type objects are now mutable - Python code can set attributes on them.

I'm unsure if this is something we should aim to fix, or just accept that it's been this way for a year and a half and update documentation?

mejrs:
> I'm unsure if this is something we should aim to fix

I think so yes. I would be really surprised if I wrote: 

```rust
#[pymethods]
impl MyClass {
    #[classattr]
    fn my_attribute() -> String {
        "hello".to_string()
    }
}
```
and python code was able to change it. IMO that kind of behaviour should be an opt-in. Maybe something like (to be bikeshedded) `#[pyo3(changeable_from_python]`.

birkenfeld:
Can you overwrite methods from `pymethods` by reassigning them on the class? If yes, I don't see why class attrs are different.

davidhewitt:
For what it's worth, a quick search suggests the type flag we'd want to use to implement this is only present in python 3.10: https://docs.python.org/3/c-api/typeobj.html#Py_TPFLAGS_IMMUTABLETYPE

I think for older versions than 3.10 the only way to protect would be a custom metaclass?

mejrs:
> Can you overwrite methods from `pymethods` by reassigning them on the class? If yes, I don't see why class attrs are different.

Yes

[...]

PR #2350, opened 2022-05-02
Expand on AsPyPointer docs and un-ignore doc examples

mejrs:
- Expands on the AsPyPointer/IntoPyPointer docs to warn about https://github.com/PyO3/pyo3/pull/2312#discussion_r859179032
- Un-ignores some doctests

### DIFF
--- a/guide/src/building_and_distribution.md
+++ b/guide/src/building_and_distribution.md
@@ -212,13 +212,13 @@ The known complications are:
 
     Significantly different compiler versions may see errors like this:
 
-    ```ignore
+    ```text
     lto1: fatal error: bytecode stream in file 'rust-numpy/target/release/deps/libpyo3-6a7fb2ed970dbf26.rlib' generated with LTO version 6.0 instead of the expected 6.2
     ```
 
     Mismatching flags may lead to errors like this:
 
-    ```ignore
+    ```text
     /usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libpython3.9.a(zlibmodule.o): relocation R_X86_64_32 against `.data' can not be used when making a PIE object; recompile with -fPIE
     ```
 
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -604,28 +604,6 @@ Python::with_gil(|py| {
 });
 ```
 
-Note that unlike class variables defined in Python code, class attributes defined in Rust cannot
-be mutated at all:
-```rust,should_panic
-# use pyo3::prelude::*;
-# #[pyclass]
-# struct MyClass {}
-# #[pymethods]
-# impl MyClass {
-#     #[classattr]
-#     fn my_attribute() -> String {
-#         "hello".to_string()
-#     }
-# }
-# 
-Python::with_gil(|py| {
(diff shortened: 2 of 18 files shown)


guide/src/building_and_distribution.md
```diff
@@ -212,13 +212,13 @@ The known complications are:
 
     Significantly different compiler versions may see errors like this:
 
-    ```ignore
+    ```text
```
davidhewitt, line 215: 👍

guide/src/class.md
```diff
@@ -606,9 +606,24 @@ Python::with_gil(|py| {
...
+#         "hello".to_string()
+#     }
+# }
+# 
+Python::with_gil(|py| {
+    let my_class = py.get_type::<MyClass>();
+
+    // Would raise a `TypeError: can't set attributes of built-in/extension type 'MyClass'`
+    pyo3::py_run!(py, my_class, "my_class.my_attribute = 'foo'")
+});
```
davidhewitt, line 626: As per #2349 shall we just change the documentation here? I'm not sure if it's worth us investing lots of effort to change this, or how hard it would be.

guide/src/migration.md
```diff
@@ -722,7 +722,7 @@ If `T` implements `Clone`, you can extract `T` itself.
 In addition, you can also extract `&PyCell<T>`, though you rarely need it.
 
 Before:
-```ignore
+```compile_fail
```
davidhewitt, line 725: I'm not sure it's worth changing bits like this which we know have old syntax, it's presumably just wasting CPU cycles to assert that the compile fails. Is there a particular intended benefit to this change?

davidhewitt (commented):
👍 a lot of great tidy-ups here, thank you! Few quick thoughts about the guide changes...

guide/src/python_from_rust.md
```diff
@@ -311,7 +311,7 @@ from anywhere as long as your `app.py` is in the expected directory (in this exa
 that directory is `/usr/share/python_app`).
 
 `src/main.rs`:
-```ignore
+```no_run
```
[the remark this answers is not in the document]
mejrs: I decided against running it because it loads things in a very platform dependent way.

guide/src/migration.md
mejrs: Well, I'd like to see to it that it doesn't start compiling again. I guess it doesn't matter because it's so old. IMO this one of these "don't waste time thinking about it, just do it" kind of things.

guide/src/class.md
mejrs: I've deleted it for now.

guide/src/migration.md
davidhewitt: Ok, though I might trim some of these back again another time if I ever try to optimise our CI pipeline. 👍

guide/src/python_from_rust.md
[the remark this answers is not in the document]
davidhewitt: I was just wondering whether the `rust` tag was also needed to give nice syntax highlighting.

davidhewitt (approved):
Looks great to me, thanks as always for the many documentation improvements which I'm so poor at! 🫣

PR #2356, opened 2022-05-04
pin cargo-llvm-cov

davidhewitt:
To fix the windows coverage job, let's try a slightly older `cargo-llvm-cov`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -328,7 +328,7 @@ jobs:
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@v1
         with:
-          tool: cargo-llvm-cov@0.3.0
+          tool: cargo-llvm-cov@0.2.4
       - run: python -m pip install -U pip nox
       - run: cargo xtask coverage --output-lcov coverage.lcov
       - uses: codecov/codecov-action@v2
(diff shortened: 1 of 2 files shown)


davidhewitt:
Hmmm 0.2.4 works if I remove the UI tests, I'll keep trying different combinations until I come to a useful conclusion.

davidhewitt:
Hmmm rewinding to 0.2.4 didn't solve it, and I don't think the windows coverage has been falling that long anyway.

Maybe it's a GitHub actions change?

@taiki-e have you by chance seen anything like the error we're getting in the Windows coverage jobs?

taiki-e:
https://github.com/PyO3/pyo3/runs/6301647762?check_suite_focus=true

> (never executed): The filename or extension is too long. (os error 206)

Looks like the same issue as https://github.com/taiki-e/cargo-llvm-cov/issues/124.

I don't know the underlying solution to this, but considering the way cargo-llvm-cov determines the object to pass to llvm-cov, disabling the cache (or cleaning target dir) on windows may alleviate the problem.

(Also, if more accurate build artifact tracking for https://github.com/taiki-e/cargo-llvm-cov/issues/102 is implemented in cargo-llvm-cov, it could be used to alleviate this problem somewhat.)

davidhewitt:
Ah, I see, it's the overall command that's too long! That makes sense, thanks.

Staring at the object files included in that command, it doesn't look like a caching issue, but more just that upstream build scripts are contributing a lot of the clutter. I opened https://github.com/taiki-e/cargo-llvm-cov/issues/154 to discuss whether you think they should be filtered.

PR #2358, opened 2022-05-05
types: add dict views

davidhewitt:
Inspired by #2347, having a play with adding the dictionary view types.

TODO

- [ ] What should their full API be? E.g. `impl IntoIterator`?
- [ ] Do we want to ask upstream CPython for a C-API rather than having to call by (interned) string method name?
- [ ] CHANGELOG entry

### DIFF
--- a/src/types/dict.rs
+++ b/src/types/dict.rs
@@ -22,6 +22,42 @@ pyobject_native_type!(
     #checkfunction=ffi::PyDict_Check
 );
 
+/// Represents a Python `dict_keys`.
+#[cfg(not(PyPy))]
+#[repr(transparent)]
+pub struct PyDictKeys(PyAny);
+
+#[cfg(not(PyPy))]
+pyobject_native_type_core!(
+    PyDictKeys,
+    ffi::PyDictKeys_Type,
+    #checkfunction=ffi::PyDictKeys_Check
+);
+
+/// Represents a Python `dict_values`.
+#[cfg(not(PyPy))]
+#[repr(transparent)]
+pub struct PyDictValues(PyAny);
+
+#[cfg(not(PyPy))]
+pyobject_native_type_core!(
+    PyDictValues,
+    ffi::PyDictValues_Type,
+    #checkfunction=ffi::PyDictValues_Check
+);
+
+/// Represents a Python `dict_items`.
+#[cfg(not(PyPy))]
+#[repr(transparent)]
+pub struct PyDictItems(PyAny);
+
+#[cfg(not(PyPy))]
+pyobject_native_type_core!(
+    PyDictItems,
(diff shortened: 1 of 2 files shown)


samuelcolvin:
Thank you so much.

Just a thought - not particularly necessary for me, would it be possible to add methods to `PyDict` to return keys as a `PyDictKeys` etc?

E.g. `.dict_keys()`?

src/types/dict.rs
```diff
@@ -175,6 +205,39 @@ impl PyDict {
         }
     }
 
+    /// Returns a view of dict keys.
+    ///
+    /// This is equivalent to the Python expression `dict.keys()`.
+    pub fn keys_view(&self) -> &PyDictKeys {
+        unsafe {
+            PyDictKeys::try_from_unchecked(
+                self.call_method0(intern!(self.py(), "keys")).unwrap(),
```
alex, line 214: Can `self` be a dict subclass -- which can therefore return a different type?

davidhewitt: Yes. I guess that would be the advantage of having a C API - then we could always hit the guaranteed pathway rather than lookup and allow subclasses to override things.

davidhewitt:
> Just a thought - not particularly necessary for me, would it be possible to add methods to PyDict to return keys as a PyDictKeys etc?
>
> E.g. `.dict_keys()`?

I think this is what I've called `PyDict::keys_view` etc. in this PR? Though I might punt on the methods for now and just add the type definitions as a first step.

mejrs:
It it safe to have a dictview while mutating the dict itself?

davidhewitt:
> Is it safe to have a dictview while mutating the dict itself?

Yes, looks like it: 

```python
>>> d = {}
>>> k = d.keys()
>>> k
dict_keys([])
>>> d['foo'] = 1
>>> k
dict_keys(['foo'])
```

davidhewitt:
I've pared this back to just be the `struct` definitions for now; the decision on what to do about methods might depend on what upstream API CPython will accept.

samuelcolvin:
For my use case, this looks great. Thanks so much.

davidhewitt:
> For my use case, this looks great. Thanks so much.

👍. Worth noting that it appears these types appear not to be supported on pypy, although they would probably be willing to add if requested on their issue tracker.

samuelcolvin:
Is there any chance that iterating over dict items will be faster than the list I get from `.items()`?

davidhewitt:
Off the top of my head I'm not sure I can call it either way. Iterating lists is probably faster than iterating the keys collection, but you've got to pay the upfront cost of allocating and filling the whole list. It probably depends on dictionary size - as always, best to benchmark for your use case 😊

PR #2365, opened 2022-05-10
Remove `#[doc(hidden)]` from trait impl items

messense:
See https://github.com/rust-lang/rust/pull/96008


mejrs:
If these do actually show up in the documentation maybe we should just use a sealed trait rather than this private marker trick?

messense:
And it also shows up on docs.rs: https://docs.rs/pyo3/latest/pyo3/pyclass_init/trait.PyObjectInit.html#tymethod.__private__

PR #2366, opened 2022-05-10
wrap_x: make implementation more consistent

davidhewitt:
Should make an implementation of #2362 a bit easier, as well as some ideas I have around `#[pymodule] mod foo { ... }`


davidhewitt:
Hmmm. So it seems like both this and #2363 break an edge case of importing a `#[pymodule]` named `foo` from a module `foo` via a glob import `use foo::*`.

This might actually be an unfortunate breakage but bring things more in line with normal resolution, I'll try to check that...

davidhewitt:
Yep, looks like "normal" code behaves the same way: The following won't compile:

```rust
mod foo {
    mod foo {
        const BAR: i32 = 1;
    }
}

use foo::*;

fn main() {
    dbg!(foo::BAR);  // despite use foo::*; the submodule foo::foo does not shadow the outer module.
}
```


In which case I'm going to reopen #2363 and scrap this one.

PR #2367, opened 2022-05-10
add `#[pymodule] mod some_module { ... }`

davidhewitt:
Implements using `#[pymodule]` directly on modules. 

Allows importing other wrapped items with a `#[pyo3] use foo;`. (Builds on #2366 to do this.)

Allows having a `#[pymodule_init]` function to run arbitrary code after adding all wrapped items. 

Still very unfinished, hopefully it's an interesting proof-of-concept.

See the declarative module in `test_module.rs` and also `pytests/src/lib.rs` for examples of how this could be used.


davidhewitt:
Superseded by #3294

PR #2371, opened 2022-05-14
ffi: add new 3.11 fields to PyConfig

indygreg:
This hopefully makes PyConfig compatible with 3.11. I haven't tested this though.

This is based on top of #2370 and should probably be rebased once that PR merges. Only the 2nd commit is unique to this PR. I wish GitHub let you chain stacks of commits together more easily...


davidhewitt (approved):
https://github.com/PyO3/pyo3/pull/2370#issuecomment-1126862857

PR #2372, opened 2022-05-14
Auto generate Windows import libraries when using a pyo3 config file

messense:
Currently the build script of `pyo3-ffi` calls `pyo3-build-config`'s `resolve_interpreter_config` function

https://github.com/PyO3/pyo3/blob/87bd10c9a33e0a189248f32e46fd1c602ab5606a/pyo3-ffi/build.rs#L79-L80

when `PYO3_CONFIG_FILE` env var is set, it simply read the file using `InterpreterConfig::from_reader` which missed the "generate import library" setup thus no `python3.dll`/`python3y.dll` is generated causing linking errors

https://github.com/PyO3/pyo3/blob/87bd10c9a33e0a189248f32e46fd1c602ab5606a/pyo3-build-config/src/lib.rs#L179-L180

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -456,6 +456,27 @@ print("mingw", get_platform().startswith("mingw"))
         })
     }
 
+    pub fn fixup_import_libs(&mut self) -> Result<()> {
+        let target = target_triple_from_env();
+        if self.lib_name.is_none() && target.operating_system == OperatingSystem::Windows {
+            self.lib_name = Some(default_lib_name_windows(
+                self.version,
+                self.implementation,
+                self.abi3,
+                false,
+            ));
+        }
+        // Auto generate python3.dll import libraries for Windows targets.
+        #[cfg(feature = "python3-dll-a")]
+        {
+            if self.lib_dir.is_none() {
+                let py_version = if self.abi3 { None } else { Some(self.version) };
+                self.lib_dir = self::import_lib::generate_import_lib(&target, py_version)?;
+            }
+        }
+        Ok(())
+    }
+
     #[doc(hidden)]
     /// Serialize the `InterpreterConfig` and print it to the environment for Cargo to pass along
     /// to dependent packages during build time.
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -456,6 +456,7 @@ print("mingw", get_platform().startswith("mingw"))
         })
     }
 
+    #[allow(clippy::unnecessary_wraps)]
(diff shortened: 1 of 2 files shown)


pyo3-build-config/src/impl_.rs
```diff
@@ -441,6 +441,24 @@ print("mingw", get_platform().startswith("mingw"))
         let implementation = implementation.unwrap_or(PythonImplementation::CPython);
         let abi3 = abi3.unwrap_or(false);
 
+        // Auto generate python3.dll import libraries for Windows targets.
+        #[cfg(feature = "python3-dll-a")]
```
messense, line 445: I'm not sure whether this is the right place to do this, feel free to suggest a better approach.

messense:
cc @ravenexp

pyo3-build-config/src/impl_.rs
ravenexp: Perhaps a new `InterpreterConfig` method like `fixup_import_libs(&mut self)` would be more readable and easier to feature-gate.

You can then invoke it explicitly after `from_reader()` here:

https://github.com/PyO3/pyo3/blob/87bd10c9a33e0a189248f32e46fd1c602ab5606a/pyo3-build-config/src/lib.rs#L179-L180

pyo3-build-config/src/impl_.rs
```diff
@@ -456,6 +456,28 @@ print("mingw", get_platform().startswith("mingw"))
...
+        {
+            let target = target_triple_from_env();
+            if self.lib_name.is_none() && target.operating_system == OperatingSystem::Windows {
+                self.lib_name = Some(default_lib_name_windows(
+                    self.version,
+                    self.implementation,
+                    self.abi3,
+                    false,
+                ));
+            }
```
ravenexp, line 472: I think this can be done unconditionally, because `lib_name` has to be set to _something_ on Windows, no matter whether `python3-dll-a` is enabled or not.

pyo3-build-config/src/impl_.rs
```diff
@@ -456,6 +456,28 @@ print("mingw", get_platform().startswith("mingw"))
...
+            if self.lib_dir.is_none() {
+                let py_version = if self.abi3 { None } else { Some(self.version) };
+                self.lib_dir = self::import_lib::generate_import_lib(&target, py_version)?;
+            }
```
ravenexp, line 476: and this we can do only when `python3-dll-a` is available.

ravenexp (commented):
LGTM in general, but I'd fix up `lib_name` either way.

ravenexp (approved):
Looks good, thanks!

davidhewitt (approved):
👍

PR #2374, opened 2022-05-15
add regression test for PEP 587

davidhewitt:
To follow up #2371 

Adds FFI definition `Py_ExitStatusException` so I can write the test.


davidhewitt:
Might need to unpin to 3.11 beta 1.

Issue #2359, opened 2022-05-05
[needs-design]Can we let macro generate function to append module to init table?


herquan:
Hi,
In order to debug module sometimes we need to run a python script inside the rust environment. One way to do this is through calling function [PyImport_AppendInittab](https://docs.python.org/3/c-api/import.html#c.PyImport_AppendInittab) before [pyo3::prepare_freethreaded_python()](https://docs.rs/pyo3/latest/pyo3/fn.prepare_freethreaded_python.html). We don't have this in pyo3 directly but one workaround would be adding such piece of code
```
pub extern "C" fn pyinit() -> *mut ::pyo3::ffi::PyObject {
	unsafe { PyInit_module() }
}
pub fn append_inittab(){
	unsafe {
		pyo3::ffi::PyImport_AppendInittab("module\0".as_ptr() as *const std::os::raw::c_char, Option::Some(pyinit));
	}
}
```
And I'm thinking of adding such support in macro [pymodule](https://github.com/PyO3/pyo3/blob/main/pyo3-macros-backend/src/module.rs#L82) with similar code like
```
pub fn append_inittab_wrapper(){
    unsafe {
        #krate::ffi::PyImport_AppendInittab(concat!(stringify!(#name), "\0").as_ptr() as *const std::os::raw::c_char, Option::Some(pyinit));
    }
}
```
Meanwhile make sure [PyImport_AppendInittab](https://github.com/PyO3/pyo3/blob/main/pyo3-ffi/src/import.rs#L79) contains unsafe so [this function](https://github.com/PyO3/pyo3/blob/16ee22c7ccd7dab66242ff95da20dd71fd7684b2/pyo3-macros-backend/src/module.rs#L82) can be reused. One thing I'm not sure is how to let the rust project that uses pyo3 recognize the `append_inittab_wrapper`  without guessing the function name.
Please let me know if there is any concern for this or if there is any better idea?

davidhewitt:
I would be ok with having a safe function `PyModule::append_to_inittab()`. It would probably make sense for it to panic if `Py_IsInitialized()` returns true.

davidhewitt:
Actually maybe this would need to a a macro `pyo3::append_to_inittab!(my_module)` which uses the same kind of tricks done in `wrap_pymodule`. PRs to experiment with this is welcome!

PR #2377, opened 2022-05-16
Add macro append_to_inittab (issue #2359)

herquan:
Sometimes we need to debug in a real environment with our module installed. `append_to_inittab` will be a wrapper for PyImport_AppendInittab (https://docs.python.org/3/c-api/import.html#c.PyImport_AppendInittab) and help us to do this


src/macros.rs
```diff
@@ -150,3 +150,24 @@ macro_rules! wrap_pymodule {
         }
     };
 }
+
+/// Add current module to the existing table of built-in modules.
+///
+/// Use it before [`prepare_freethreaded_python`](crate::prepare_freethreaded_python) and
+/// leave feature `auto-initialize` off
+#[macro_export]
```
davidhewitt, line 158: I don't think `PyImport_AppendInittab` is implemented by PyPy.

```suggestion
#[macro_export]
#[cfg(not(PyPy))]
```

tests/test_module.rs
```diff
@@ -443,3 +443,21 @@ fn test_module_doc_hidden() {
         py_assert!(py, m, "m.__doc__ == ''");
     })
 }
+
+#[test]
+fn test_module_append_to_inittab() {
+    use pyo3::append_to_inittab;
+    append_to_inittab!(module_with_functions);
```
davidhewitt, line 450: You will probably find that this needs to go in its own test file, because it's quite likely that another test will already have initialised the Python interpreter.

davidhewitt (commented):
Nice! Much simpler than the proc macro, I'm glad the refactoring seems to have helped :)

Couple of comments. Also please add a CHANGELOG entry.

src/macros.rs
herquan: Added above

davidhewitt: 👍, CI won't rerun until merge conflict is fixed.

src/macros.rs
```diff
@@ -150,3 +150,25 @@ macro_rules! wrap_pymodule {
         }
     };
 }
+
+#[cfg(not(PyPy))]
+/// Add current module to the existing table of built-in modules.
+///
+/// Use it before [`prepare_freethreaded_python`](crate::prepare_freethreaded_python) and
+/// leave feature `auto-initialize` off
```
mejrs, line 158: To be honest, I wasn't aware of this function, and I think the vast majority of people reading this documentation are not either.

In particular, a reader might have the following questions or thoughts:
- I want to add a module, this is what I need (is it?)
- What does this do and why should I use it?
- What is the point of adding builtin modules?
- You said you use this to debug modules, can you explain? How can a user use this for debugging?

mejrs (changes requested):
Thanks :)

Can you please add a test in https://github.com/PyO3/pyo3/blob/main/src/test_hygiene/misc.rs?

herquan:
> Thanks :)
> 
> Can you please add a test in https://github.com/PyO3/pyo3/blob/main/src/test_hygiene/misc.rs?

I'm not sure what do you want me to add in this test? Do you want `test_module_append_to_inittab` to be there?

src/macros.rs
herquan: Added here and `building_and_distribution.md`

tests/test_append_to_inittab.rs
```diff
@@ -0,0 +1,32 @@
+#![cfg(feature = "macros")]
```
davidhewitt, line 1: Instead of `#[cfg]` on the test, I think you want:

```suggestion
#![cfg(all(feature = "macros", not(PyPy))]
```

guide/src/building_and_distribution.md
```diff
@@ -225,6 +225,10 @@ The known complications are:
 
 If you encounter these or other complications when linking the interpreter statically, discuss them on [issue 416 on PyO3's Github](https://github.com/PyO3/pyo3/issues/416). It is hoped that eventually that discussion will contain enough information and solutions that PyO3 can offer first-class support for static embedding.
 
+### Import your module when embedding the CPython interpreter
```
davidhewitt, line 228: In general PyPy doesn't support embedding, so I don't think you need to make the distinction here.

```suggestion
### Import your module when embedding the Python interpreter
```

guide/src/building_and_distribution.md
```diff
@@ -225,6 +225,10 @@ The known complications are:
...
+
+When you run your Rust binary with CPython (not PyPy) interpreter, your newly created module won't be initialized unless the function defined with macro `#[pymodule]` is added to a table called `PyImport_Inittab`. This means Python statements like `import your_new_module` run by your Rust binary will fail. You can use macro `append_to_inittab` before function `prepare_freethreaded_python` being called to add the module function into that table. Also [`auto-initialize`](features.md#auto-initialize) needs to be turned off in such scenario.
```
davidhewitt, line 230: Some phrasing suggestions. 

```suggestion
When you run your Rust binary with an embedded interpreter, any `#[pymodule]` created modules won't be accessible to import unless added to a table called `PyImport_Inittab` before the embedded interpreter is initialized. This will cause Python statements in your embedded interpreter such as `import your_new_module` to fail. You can call the macro `append_to_inittab` with your module before initializing the Python interpreter to add the module function into that table. (The Python interpreter will be initialized by calling `prepare_freethreaded_python`, `with_embedded_interpreter`, or `Python::with_gil` with the [`auto-initialize`](features.md#auto-initialize) feature enabled.)
```

Also, it would be great to link `append_to_inittab` to the docs - see the `PYO3_DOCS_URL` substitutions we use.

davidhewitt (commented):
The purpose of `test_hygiene` is to ensure that the macros still work even when `pyo3` is in unusual import paths (i.e. `use ::pyo3` won't work). In `test_hygiene`, for example, `pyo3` is imported as `crate`.

I think adding the following to the bottom of `test_hygiene/misc.rs` will be sufficient - the generated code just needs to compile:

```
#[allow(dead_code)]
fn append_to_inittab(py: crate::Python<'_>) {
    #[crate::pymodule]
    fn my_module(_: crate::Python<'_>, _: &crate::PyModule) -> crate::PyResult<()> {
            ::std::result::Result::Ok(())
    }
    
    crate::append_to_inittab!(my_module);
}
```

herquan:
> The purpose of `test_hygiene` is to ensure that the macros still work even when `pyo3` is in unusual import paths (i.e. `use ::pyo3` won't work). In `test_hygiene`, for example, `pyo3` is imported as `crate`.
> 
> I think adding the following to the bottom of `test_hygiene/misc.rs` will be sufficient - the generated code just needs to compile:
> 
> ```
> #[allow(dead_code)]
> fn append_to_inittab(py: crate::Python<'_>) {
>     #[crate::pymodule]
>     fn my_module(_: crate::Python<'_>, _: &crate::PyModule) -> crate::PyResult<()> {
>             ::std::result::Result::Ok(())
>     }
>     
>     crate::append_to_inittab!(my_module);
> }
> ```

Macro pymodule alone is not working in this file and it seems there are still some code that needs to handle unusual import path. I think it is reasonable to fix it first before we add such test in this pull request.

davidhewitt:
> Macro pymodule alone is not working in this file 

My sample above was wrong, it needs to be:

```rust
#[allow(dead_code)]
fn append_to_inittab(py: crate::Python<'_>) {
    #[crate::pymodule]
    #[pyo3(path = "crate")]
    fn my_module(_: crate::Python<'_>, _: &crate::PyModule) -> crate::PyResult<()> {
            ::std::result::Result::Ok(())
    }
    
    crate::append_to_inittab!(my_module);
}

```

I forgot to tell the `#[pymodule]` macro what the PyO3 path was!

davidhewitt:
> I think it is reasonable to fix it first before we add such test in this pull request.

Please add the test to this PR. If there is an issue elsewhere it will be useful to see CI output anyway.

src/macros.rs
```diff
@@ -150,3 +150,26 @@ macro_rules! wrap_pymodule {
...
+macro_rules! append_to_inittab {
+    ($module:ident) => {
+        unsafe {
+            ::std::assert_eq!(
+                $crate::ffi::Py_IsInitialized(),
+                0,
+                "called `append_to_inittab_impl` but a Python interpreter is already running."
+            );
+            $crate::ffi::PyImport_AppendInittab(
+                concat!(stringify!($module), "\0").as_ptr() as *const ::std::os::raw::c_char,
```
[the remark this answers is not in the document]
herquan: I don't know about too much fancy grammars in rust and hopefully my new change looks good.

src/macros.rs
```diff
@@ -150,3 +150,26 @@ macro_rules! wrap_pymodule {
...
+/// Add the module to the initialization table in order to make embedded Python code to use it.
+/// Module name is the argument.
+///
+/// Use it before [`prepare_freethreaded_python`](crate::prepare_freethreaded_python) and
+/// leave feature `auto-initialize` off
+#[macro_export]
+macro_rules! append_to_inittab {
+    ($module:ident) => {
+        unsafe {
+            ::std::assert_eq!(
```
davidhewitt, line 164: Hmmm, looks like `assert_eq` is not hygienic until 1.50. @mejrs how do you feel if we just put `#[rust version::since(1.50)]` on this test and live with it? It's such a niche case that messying the implementation for the older compilers (to support only `no_implicit_prelude`) doesn't seem worth it.

mejrs: It can just be `if a != b { panic!()}` rather than an assert_eq (may need an allow to stop clippy from complaining about it though)

davidhewitt:
I'm going to force-push a formatting fix and then hopefully this can be merged. Thanks @herquan for proposing and implementing this!

Issue #2162, opened 2022-02-12
Iteration safety for containers


davidhewitt:
The other day I noticed this doc comment on `PyDict::iter`.

```
    /// Note that it's unsafe to use when the dictionary might be changed by
    /// other code.
```

This sounds scary; it's a safe function so shouldn't have arbitrary safety warnings like this.

I know that when iterating Python containers from Python they do have checks to prevent modification. We should probably check for our container types what happens if we try to modify during iteration, and also whether it really is unsound or just might introduce logical bugs.

I'd argue that if our current implementation is indeed unsound then I'd hope we can patch things up to prevent modification so that `PyDict::iter` and friends can remain safe.

(We should only need to check `PyList`, `PyDict`, and `PySet`; `PyFrozenSet` and `PyTuple` are immutable so shouldn't need this protection).

nw0:
The iterator code only uses `PyDict_Next` ([C-API doc](https://docs.python.org/3/c-api/dict.html#c.PyDict_Next)), which says

> The dictionary p should not be mutated during iteration. It is safe to modify the values of the keys as you iterate over the dictionary, but **only so long as the set of keys does not change**. (emphasis mine)

The phrasing is a bit ambiguous, so I read this statement as:

1. Don't add or remove keys from the dictionary
2. Don't modify the keys (which you shouldn't be doing anyway...)

I don't think we can add/remove keys on the (safe) Rust side as long as we hold the iterator. Is it possible to violate this by calling Python code?

Also, as I understand it (possibly incorrectly) the returned `&'py PyAny` could be used to mutate the underlying Python object -- in this case, the key. Arguably Python programmers already know they shouldn't do this: I'm not sure whether we can actually do something about that given #883.

birkenfeld:
You can do anything to the dict while the iterator is active since mutating operations are fine with shared borrows.

So it seems like to be safe here, we cannot use the C-layer iterators, but have to use the Python-layer iterators, which need to guard against all the same things that we do - and so e.g. raise RuntimeError when the dictionary changes during iteration.

Since `PyDict_Next` is probably a lot faster, we can still offer it as `unsafe` API, with the necessary warnings.

PR #2380, opened 2022-05-16
Protect iterators against concurrent modification

mejrs:
Fixes https://github.com/PyO3/pyo3/issues/2162

 - [x] Changelog
- [x] More tests?
- [x] Does PyList need to be protected? It seems fine as is.


src/types/dict.rs
```diff
@@ -213,13 +213,13 @@ impl PyDict {
 
     /// Returns an iterator of `(key, value)` pairs in this dictionary.
     ///
-    /// Note that it's unsafe to use when the dictionary might be changed by
-    /// other code.
+    /// # Panics
+    ///
+    /// If PyO3 detects that the dictionary is mutated during iteration, it will panic.
+    /// It is allowed to modify the values of the keys as you iterate over the dictionary, but only
```
birkenfeld, line 219: I know "modify the values of the keys" is copied verbatim from the C-API docs, but I think it's quite confusing.

src/types/dict.rs
```diff
@@ -228,51 +228,129 @@ impl PyDict {
...
+#[cfg(all(Py_3_8, not(PyPy), not(Py_LIMITED_API)))]
+mod impl_ {
+    use super::*;
+    use std::marker::PhantomData;
+
+    pub struct PyDictIterator<'py> {
+        di_dict: *mut ffi::PyDictObject,
+        di_pos: ffi::Py_ssize_t,
+        di_used: ffi::Py_ssize_t,
+        marker: PhantomData<&'py PyDict>,
```
davidhewitt, line 240: Is there a risk that the dict refcount can fall to 0 and can be accidentally released? Should this be `Py<PyDict>`?

src/types/dict.rs
```diff
@@ -228,51 +228,129 @@ impl PyDict {
...
-                ))
-            } else {
-                None
+    impl<'a> std::iter::IntoIterator for &'a PyDict {
+        type Item = (&'a PyAny, &'a PyAny);
+        type IntoIter = PyDictIterator<'a>;
+
+        fn into_iter(self) -> Self::IntoIter {
+            unsafe {
+                let di_dict: *mut ffi::PyDictObject = self.as_ptr().cast();
```
davidhewitt, line 294: I've always wondered if there should be an `as_raw() -> *mut ffi::PyDictObject` method on `PyDict` (and similar for other types)? Available only on not (limited_api / PyPy).

src/types/dict.rs
```diff
@@ -228,51 +228,129 @@ impl PyDict {
...
+        ///
+        /// # Panics
+        ///
+        /// If PyO3 detects that the dictionary is mutated during iteration, it will panic.
+        /// It is allowed to modify the values of the keys as you iterate over the dictionary, but only
+        /// so long as the set of keys does not change.
+        #[inline]
+        fn next(&mut self) -> Option<Self::Item> {
+            unsafe {
+                if self.di_used != (*(self.di_dict)).ma_used {
```
davidhewitt, line 256: What's the difference between `ma_used` and `PyDict_Size` used below?

src/types/dict.rs
```diff
@@ -228,51 +228,129 @@ impl PyDict {
...
+                    self.di_used = -1;
+                    panic!("dictionary changed size during iteration");
```
davidhewitt, line 258: Panicking seems much nicer than returning results 👍

src/types/frozenset.rs
```diff
@@ -0,0 +1,194 @@
...
+    PyFrozenSet,
+    ffi::PySetObject,
+    ffi::PyFrozenSet_Type,
+    #checkfunction=ffi::PyFrozenSet_Check
+);
+
+impl PyFrozenSet {
+    /// Creates a new frozenset.
+    ///
+    /// May panic when running out of memory.
```
davidhewitt, line 25: I know not from this PR, but is this actually true? It looks like returning `PyResult` means that we'll actually return `PyErr` containing a `MemoryError` if we run out of memory.

I suspect that actually the text may want to be more like `PySet` where we don't mention this and instead comment on the error return being caused by unhashable contents.

src/types/frozenset.rs
```diff
@@ -0,0 +1,194 @@
+// Copyright (c) 2017-present PyO3 Project and Contributors
+//
+
```
davidhewitt, line 3: 👍 good idea to make a new file.

davidhewitt (commented):
I love this! 🚀 

> Does PyList need to be protected? It seems fine as is.

Good question. I guess it's by luck that with the current implementation we re-fetch the `len()` on each call to `next()`? I guess checking for modifications would require that _and_ an extra check, so maybe it's ok as-is.

src/types/dict.rs
mejrs: The iterator borrows from the Pydict, so this should be fine.

davidhewitt (approved):
Looks great, smart new implementation!

davidhewitt:
I know I already approved... Looks like this needs a CHANGELOG and then good to merge?

mejrs:
Yes, I've been busy for a bit. I'll do it soon :)

mejrs:
That should be all 🎉

src/types/mod.rs
```diff
@@ -32,6 +33,42 @@ pub use self::traceback::PyTraceback;
...
+/// })
+/// # }
+///  ```
+///
+/// If PyO3 detects that the collection is mutated during iteration, it will panic.
+///
+/// These iterators use Python's C-API directly. However in certain cases, like when compiling for
+/// the Limited API and PyPy, the underlying structures are opaque and that may not be possible.
+/// In these cases the iterators are implemented by forwarding to [`PyIterator`].
+pub mod iter {
```
davidhewitt, line 66: 👍 it was long overdue for these to be public!

davidhewitt (approved):
Looks great!

PR #2385, opened 2022-05-17
Allow `#[classattr]` methods to be fallible

davidhewitt:
Noticed this morning while looking at #2382 that it's not possible to return `PyResult` from `#[classattr]` methods.

There's not much we can do with them but panic, but it is at least a nice consistency change to match other `#[pymethods]`.


davidhewitt:
Anyone got any concerns before I merge this? I don't think it's essential at all, just seems like a sensible refinement.

PR #2398, opened 2022-05-24
Add `GILOnceCell::get_or_try_init`

a1phyr:
This is similar to [`OnceCell::get_or_try_init`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html#method.get_or_try_init) and is very useful for fallible initialization.

### DIFF
--- a/src/once_cell.rs
+++ b/src/once_cell.rs
@@ -5,9 +5,19 @@ use std::cell::UnsafeCell;
 /// A write-once cell similar to [`once_cell::OnceCell`](https://docs.rs/once_cell/latest/once_cell/).
 ///
 /// Unlike `once_cell::sync` which blocks threads to achieve thread safety, this implementation
-/// uses the Python GIL to mediate concurrent access. This helps in cases where `once_sync` or
+/// uses the Python GIL to mediate concurrent access. This helps in cases where `once_cell` or
 /// `lazy_static`'s synchronization strategy can lead to deadlocks when interacting with the Python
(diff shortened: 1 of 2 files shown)


src/once_cell.rs
```diff
@@ -71,21 +71,39 @@ impl<T> GILOnceCell<T> {
...
 
+        match self.init(py, || Ok::<T, std::convert::Infallible>(f())) {
+            Ok(value) => value,
+            Err(void) => match void {},
+        }
+    }
+
+    /// Like `get_or_init`, but accepts a fallible initialization function.
+    ///
+    /// If it fails, the cell is left uninitialized.
```
mejrs, line 82: Can you expand on what reentrant initialization does here? (Does it deadlock like the `once_cell` impl?) And what about panics?

mejrs (changes requested):
Please add some tests 😉 Otherwise it looks good to me

src/once_cell.rs
adamreichold: We should probably centralize the documentation already present on `get_or_init` at the type level to avoid repeating ourselves here.

davidhewitt: Agreed, I've done this now.

davidhewitt (approved):
Many thanks for this, and sorry for the slow review. I agree with all comments above; I've just pushed a commit to move the docs around as suggested and will merge this now.

davidhewitt:
Hmmm, looks like bors probably wants you to give a green tick @mejrs 😄

mejrs (approved):
Thanks!

PR #2399, opened 2022-05-24
pymethods: prevent methods sharing the same name

davidhewitt:
This PR tweaks the `#[pymethods]` code generation so that the generated functions have meaningful names and are members of the associated class.

In general, the structure of generated code has changed from

```rust
unsafe extern "C" fn __wrap(...) { ... }
```

to 

```rust
impl SomeClass {
    unsafe extern "C" fn __pymethod_foo__(...) { ... }
}
```

There are a few nice benefits that this brings:
- As the generated function names are based on their names exposed to Python, and they share the struct's namespace, it's not possible to have multiple pymethods with the same `#[pyo3(name = "foo")]`.
- We have a hack to `replace_self` to remove `Self` from generated code; we don't need this now the generated code is associated with the class.
- Stack traces will be nicer in panic messages etc, because the frames will have meaningful names rather than `__wrap` (that was actually what motivated me to try this today as I was struggling to debug #2302; this idea's been in the back of my mind for a while)


mejrs:
Thanks, this should make reading `cargo expand` output a bit nicer too :)

davidhewitt:
Also true! While working on this I also realised that (with some extra work) it might be possible to rearrange the generated code so that we just add all these generated methods in a single `impl SomeClass` block, which might reduce the visual complexity of the generated output further.

PR #2404, opened 2022-05-28
Fixup `lib_name` when using `PYO3_CONFIG_FILE`

messense:
This matches the logic in `default_cross_compile`: https://github.com/PyO3/pyo3/blob/eafbbc54170e46c6b16067dabcf849a3864d6198/pyo3-build-config/src/impl_.rs#L1415-L1426

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -464,17 +464,21 @@ print("mingw", get_platform().startswith("mingw"))
         })
     }
 
+    #[cfg(feature = "python3-dll-a")]
     #[allow(clippy::unnecessary_wraps)]
-    pub fn fixup_import_libs(&mut self) -> Result<()> {
+    pub fn generate_import_libs(&mut self) -> Result<()> {
         // Auto generate python3.dll import libraries for Windows targets.
-        #[cfg(feature = "python3-dll-a")]
-        {
-            if self.lib_dir.is_none() {
-                let target = target_triple_from_env();
-                let py_version = if self.abi3 { None } else { Some(self.version) };
-                self.lib_dir = self::import_lib::generate_import_lib(&target, py_version)?;
-            }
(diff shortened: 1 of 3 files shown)


pyo3-build-config/src/impl_.rs
```diff
@@ -458,19 +477,11 @@ print("mingw", get_platform().startswith("mingw"))
 
     #[allow(clippy::unnecessary_wraps)]
     pub fn fixup_import_libs(&mut self) -> Result<()> {
```
adamreichold, line 479: Isn't the function name a misnomer now as it only generates the import libraries now?

pyo3-build-config/src/impl_.rs
```diff
@@ -440,6 +440,25 @@ print("mingw", get_platform().startswith("mingw"))
         let version = version.ok_or("missing value for version")?;
         let implementation = implementation.unwrap_or(PythonImplementation::CPython);
         let abi3 = abi3.unwrap_or(false);
+        // Fixup lib_name if it's not set
+        let lib_name = lib_name.or_else(|| {
+            if let Ok(Ok(target)) = env::var("TARGET").map(|target| target.parse::<Triple>()) {
+                if target.operating_system == OperatingSystem::Windows {
```
adamreichold, line 446: If this logic is the same as the one linked in the cover letter, would it make sense to factor it out into a helper function?

messense: Done.

pyo3-build-config/src/impl_.rs
messense: How about `generate_import_libs`? Or any other suggestions?

adamreichold: The suggested name sounds fine to me. I wonder if the cfg flags can be simplified, e.g. pub on the outside of the function instead using blocks?

adamreichold (approved):

PR #2406, opened 2022-05-28
Expose PyFrame_GetCode

alex:


mejrs (approved):
Thanks!

pyo3-ffi/src/pyframe.rs
```diff
@@ -7,5 +9,6 @@ opaque_struct!(PyFrameObject);
 
 extern "C" {
     pub fn PyFrame_GetLineNumber(f: *mut PyFrameObject) -> c_int;
+    #[cfg(Py_3_9)]
+    pub fn PyFrame_GetCode(f: *mut PyFrameObject) -> *mut PyCodeObject;
```
messense, line 13: Since it's part of the stable ABI since 3.10, I think we can annotate it with `#[cfg(any(Py_3_10, all(Py_3_9, not(Py_LIMITED_API))))]`?

alex: good catch, done

messense (approved):

PR #2408, opened 2022-05-29
Added new PyCode and PyFrame objects.

alex:
They currently do not expose any APIs, but are significantly easier to work with than raw pointers :-)


davidhewitt:
Thanks! Could you please add a CHANGELOG entry? Otherwise lgtm 👍

alex:
Done!

PR #2409, opened 2022-05-29
Expand on the C api definitions

mejrs:
I'm not entirely sure whether I got the cfg's right. When do I need to add (or not add) a PyPy gate?


davidhewitt:
Nice, these changes look along the right lines to me! It would be nice to add some tests for the inlined functions in [methodobject.rs](https://github.com/PyO3/pyo3/pull/2409/files#diff-8c91da94ff807b3e512891a48af8f3f70e41adbbab404de478af621bef678cc4)

> I'm not entirely sure whether I got the cfg's right. When do I need to add (or not add) a PyPy gate?

TBH I think struct definitions should be checked carefully whether they're PyPy compatible, because we could get UB/crashes if they're wrong. For functions, the worst that happens is that we get a linker error if there's a PyPy gate missing. So I tend to be a bit lax with PyPy gates on functions, though in principle they all could have either a `not(PyPy)` or a `link_name = "PyPy..."` attribute.

pyo3-ffi/src/cpython/object.rs
```diff
@@ -361,8 +361,15 @@ extern "C" {
...
 
-// skipped _PyNone_Type
-// skipped _PyNotImplemented_Type
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
+    pub static mut _PyNone_Type: PyTypeObject;
+}
+
+#[cfg_attr(windows, link(name = "pythonXY"))]
+extern "C" {
```
davidhewitt, line 370: I'd probably just merge these blocks.

```suggestion
```

mejrs:
I think that's all. I can't quite figure out how to squash them though. I suppose we can "squash and merge"?

davidhewitt:
👍 looks good to me, I'll click squash & merge. Thanks! (Hopefully will this PR, `pyo3-ffi-check` will now be happy for 3.11.)

PR #2413, opened 2022-06-01
macros: simpler expansion for `intern!`

davidhewitt:
Adds a helper `Interned` struct to contain the implementation of the `intern!` macro. Simplifies the expanded code a little, should allow the compiler to re-use the one monomorphization of `get_or_init`.

### DIFF
--- a/src/once_cell.rs
+++ b/src/once_cell.rs
@@ -1,5 +1,5 @@
 //! A write-once cell mediated by the Python GIL.
-use crate::Python;
+use crate::{types::PyString, Py, Python};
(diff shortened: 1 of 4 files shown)


adamreichold (approved):
Nice! Seems like a vastly cleaner implementation.

mejrs:
Does this implementation continue to forbid patterns like https://github.com/PyO3/pyo3/pull/2269#discussion_r842022306? I think that the const-ness of `new` takes care of that problem, but i'm not 100% sure.

PR #2422, opened 2022-06-02
frompyobject: improve error messages of derived impls

davidhewitt:
Just cleans up the quality of the error messages a little bit.


davidhewitt:
(There's definitely more that can be done here; I'm just going to merge this as-is for now as this implementation which doesn't drop error messages and types on the floor seems better than than what was here previously.)

mejrs:
Nice 👍

PR #2424, opened 2022-06-02
ffi: fix PyFrameObject definition

### LINKED ISSUE #2411: ffi:PyFrameObject doesn't match the structure in CPython
### Bug Description

https://github.com/PyO3/pyo3/blob/main/pyo3-ffi/src/cpython/frameobject.rs#L20 is the ffi's definition `PyFrameObject`, however this doesn't match CPython's (3.8): https://github.com/python/cpython/blob/3.8/Include/frameobject.h#L16

Notably the `f_exc_type`, `f_exc_value`, `f_exc_traceback` fields were removed in 3.7 (I believe), and the `f_trace_lines` and `f_trace_opcodes` fields were added at some point.

### Steps to Reproduce

1. Look at the links I shared :-)

Any attempt to actually use these fields would result in UB/a segfault.

### Backtrace

_No response_

all

3.8

any

### Your PyO3 version

HEAD

Not relevant

### Additional Info

I suspect there are other places where the struct definitions do not match upstream. It may make sense to add some sort of test to CI to verify they match on all the configurations tested in CI. The simplest version might be simply verifying that struct sizes match.

davidhewitt:
Closes #2411 

@alex - could be worth noting that in Python 3.11 this structure is moved to the "internal" API it looks like; which we haven't been exposing in PyO3 (and probably shouldn't be). Might need to add some of the accessor functions for you to use instead?


alex:
Yeah I've been adding the accessor functions as needed/available.

PR #2429, opened 2022-06-04
_PyOpcache does not exist on Python 3.11

alex:


davidhewitt (approved):
Thanks!

pyo3-ffi/src/cpython/code.rs
```diff
@@ -39,7 +39,7 @@ pub struct PyCodeObject {
     pub co_extra: *mut c_void,
     #[cfg(Py_3_8)]
     pub co_opcache_map: *mut c_uchar,
-    #[cfg(Py_3_8)]
+    #[cfg(all(Py_3_8, not(Py_3_11)))]
```
mejrs, line 42: With https://github.com/PyO3/pyo3/pull/2409 I split this struct's definition, so this cfg now does nothing.

davidhewitt: 👍 I'll tidy this up later.

davidhewitt: Done in #2432

PR #2430, opened 2022-06-04
ob_shash is available on CPython

alex:


mejrs:
Ah thanks - I think my brain took a wrong turn somewhere when I was looking at PyPy's definition.

davidhewitt:
Thanks both, looks like we might be in good shape for all released CPythons now: https://github.com/PyO3/pyo3-ffi-check/actions/runs/2441081510

 I might try to take another look at the PyPy and 3.11 diffs in the morning...

PR #2435, opened 2022-06-05
Add support for mappingproxy objects

### LINKED ISSUE #2108: Add support for mappingproxy? (standard library read-only dict)
Quoting Python docs for [`vars`](https://docs.python.org/3.8/library/functions.html#vars):

> Objects such as modules and instances have an updateable `__dict__` attribute; however, other objects may have write restrictions on their `__dict__` attributes (for example, classes use a [`types.MappingProxyType`](https://docs.python.org/3.8/library/types.html#types.MappingProxyType) to prevent direct dictionary updates).

There's `pyo3::ffi::PyDictProxy_Type` and ` pyo3::ffi::PyDictProxy_New` but they are currently unused.

It's almost a copy of `PyDict`, with only these methods present: `copy`, `get`, `items`, `keys`, `values`.

This has come up while implementing structured dtypes in `rust-numpy`, where `dtype.fields` is a read-only mappingproxy dict. It could be worked around via `PyAny`, but I figured maybe I should raise it here in case there's any interest.

dignissimus:
This pull request adds support for mappingproxy objects.

Resolves #2108


src/types/mappingproxy.rs
```diff
@@ -0,0 +1,563 @@
...
+/// Represents a Python `mappingproxy`.
+#[repr(transparent)]
+pub struct PyMappingProxy(PyAny);
+
+pyobject_native_type!(
+    PyMappingProxy,
+    ffi::PyDictProxyObject,
+    ffi::PyDictProxy_Type,
+    #checkfunction=PyDictProxy_Check
+);
```
davidhewitt, line 29: Because we don't have a definition for `PyDictProxyObject` (as above), we should instead do:

```suggestion
pyobject_native_type_core!(
    PyMappingProxy,
    ffi::PyDictProxy_Type,
    #checkfunction=PyDictProxy_Check
);
```

This is almost equivalent, essentially just means that we can't use `#[pyclass(extends=PyMappingProxy)]` because we don't know the layout of the struct in memory.

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,563 @@
...
+
+impl PyMappingProxy {
+    /// Creates a mappingproxy from an object.
+    pub fn new(py: Python<'_>, elements: impl IntoPyDict) -> PyResult<&'_ PyMappingProxy> {
```
davidhewitt, line 33: This is an interesting choice for `elements`. Why did you pick it? I'd like to make sure we get the right one before we publish this API.

Some alternatives

```rust
elements: &'py PyDict,
elements: &'py PyMapping,
elements: &'py PyAny,
```

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,563 @@
...
+    ///
+    /// Returns an error on invalid input. In the case of key collisions,
+    /// this keeps the last entry seen.
+    #[cfg(not(PyPy))]
+    pub fn from_sequence(py: Python<'_>, seq: PyObject) -> PyResult<&PyMappingProxy> {
+        unsafe {
+            let dict = py.from_owned_ptr::<PyDict>(PyDict::from_sequence(py, seq)?.as_ptr());
+            py.from_owned_ptr_or_err(ffi::PyDictProxy_New(dict.as_ptr()))
+        }
+    }
```
davidhewitt, line 53: I guess this is just a direct copy of the `PyDict` equivalent? Depending on what we decide for `PyMappingProxy::new`, this may or may not be necessary.

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,563 @@
...
+    /// This is equivalent to the Python expression `self.copy()`.
+    pub fn copy(&self) -> PyResult<&PyDict> {
+        self.call_method0("copy")
+            .and_then(|object| object.downcast().map_err(PyErr::from))
+    }
+
+    /// Checks if the mappingproxy is empty, i.e. `len(self) == 0`.
+    pub fn is_empty(&self) -> bool {
+        self.len().unwrap() == 0
+    }
```
davidhewitt, line 66: Looks like this one would benefit from having a test added (or just used in an existing test).

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,563 @@
...
+        unsafe {
+            let dict = py.from_owned_ptr::<PyDict>(PyDict::from_sequence(py, seq)?.as_ptr());
+            py.from_owned_ptr_or_err(ffi::PyDictProxy_New(dict.as_ptr()))
+        }
+    }
+
+    /// Returns a new mappingproxy that contains the same key-value pairs as self.
+    ///
+    /// This is equivalent to the Python expression `self.copy()`.
+    pub fn copy(&self) -> PyResult<&PyDict> {
```
davidhewitt, line 58: I don't think `&PyDict` is correct, looks like it calls `copy` on the inner mapping:

https://github.com/python/cpython/blob/66f5add1f0ac801cba5ece36e6cfd68985d82029/Objects/descrobject.c#L1135

So the return type should either be `&PyAny` or (maybe) `&PyMapping`.

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,563 @@
...
+    T: PyDictItem,
+    I: IntoIterator<Item = T>,
+{
+    fn into_py_mappingproxy(self, py: Python<'_>) -> PyResult<&PyMappingProxy> {
+        PyMappingProxy::new(py, self)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
davidhewitt, line 173: Very nice set of tests!

davidhewitt (changes requested):
Thanks very much for implementing this!

It's overall very good. To make sure we get this API right when we merge, I've placed quite a few comments around the place with some things that need changing and then also some places where we can discuss the design a little to ensure we make the right choices.

pyo3-ffi/src/lib.rs
```diff
@@ -264,6 +264,7 @@ macro_rules! opaque_struct {
     };
 }
 
+#[macro_export]
```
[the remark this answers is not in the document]
dignissimus: Resolved in 7f98dd9f26a15c07c560a5204c7fdbda3d0ba366 and 9bd2f440a5150176378bdcfbd6e68516aa020235

src/types/dict.rs
```diff
@@ -444,12 +444,19 @@ where
...
-            ret.insert(K::extract(k)?, V::extract(v)?);
+        match <PyDict as PyTryFrom>::try_from(ob) {
+            Ok(dict) => {
+                let mut ret = HashMap::with_capacity_and_hasher(dict.len(), S::default());
+                for (k, v) in dict.iter() {
+                    ret.insert(K::extract(k)?, V::extract(v)?);
+                }
+                Ok(ret)
+            }
+            Err(_) => {
```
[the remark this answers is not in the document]
dignissimus: Hopefully resolved in c5e434d68eb3548e83dad378e3312139b426c65f

src/types/mappingproxy.rs
dignissimus: Resolved in 651a007d2886323b12079816af192e4a26563a9b

davidhewitt:
Hi @dignissimus, just wondering what the status of this is and whether you're waiting for any input from me? No rush at all, just browsing our backlog of PRs.

dignissimus:
> Hi @dignissimus, just wondering what the status of this is and whether you're waiting for any input from me? No rush at all, just browsing our backlog of PRs.

Hey! For the elements parameter, I'm looking for input on the type? When writing it, I saw the benefit of `impl IntoPyDict` as the fact that it covered types of data, so they don't have to be converted. to other datatypes. But if something else works better, I'd be willing to change it.

dignissimus:
Also, some of the builds fail (emscripten) for reasons outside of testing, if I remember correctly it might be to do with imports? I'm not able to fix it on my own without help. I'd need guidance to amend it or have somebody else add a commit to the PR.

davidhewitt (commented):
Cool - added a bunch of suggestions which should help you continue moving!

dignissimus:
> Cool - added a bunch of suggestions which should help you continue moving!

Thanks! I'll make the changes this Saturday

DataTriny:
Hello @dignissimus 

Thank you!

davidhewitt:
No progress since what is seen here, it's also not a priority feature for me so unlikely to move any time soon without further external contributions.

davidhewitt:
Superseded by #3521

PR #2436, opened 2022-06-05
Set up CI for wasm32-emscripten target

### LINKED ISSUE #2412: wasm32-emscripten support
I am interested in wasm32-emscripten support for pyo3. We are able to build the cryptography package:
https://github.com/pyodide/pyodide/pull/2378
There are still some issues with the toolchain which require patching:
https://github.com/llvm/llvm-project/issues/55786

Another issue that I have is that for some reason pyo3 produces `blah.wasm` files rather than `blah.so` files. In our cargo/rustc wrapper, we look for `.wasm` files and move them to `.so` files. This would probably be an easy fix but I don't know where the problem comes from. We also use `-Zbuild-std` so we need to use nightly.
https://github.com/pyodide/pyodide/pull/2378/files#diff-dcdc2b45d6538a6db4fa7c76a3b29c56f5f7e7669ab7460385c4241d6d453ce2
If these are things that could be improved in PyO3 or if I am just missing some setting, it would be helpful to know.

I see on #1522 you said:

> If anyone with knowledge of how to build a Python interpreter for wasm wants to contribute a full CI job, that would be cool!

I would be interested in contributing a CI job for wasm32-emscripten. Options:
* build against 3.11b1 -- we wouldn't need to patch the Python interpreter, but this may require the rest of the PyO3 dependencies to work against 3.11b1 which it might not be ready for
* build against 3.10 -- Pyodide already does this. We would need a few patches, but probably some of our patches can be dropped.
 
Also, until wasm-ld is fixed to not be mad about `lib.rmeta` files we will need a patch for emscripten. (I'm actually a bit surprised you don't run into this problem with `wasm32-wasi`.)

Build time for the wasm32-emscripten Python interpreter is generally a bit shy of 15 minutes. I think we could run the tests themselves in node.

hoodmane:
This adds CI tests for wasm32-emscripten target. CI run builds `libpython3.11.a` for the wasm32-emscripten target and then runs `cargo test` against it.
Resolves #2412.

- [x] fix or xfail broken tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -339,3 +339,35 @@ jobs:
         with:
           file: coverage.lcov
           name: ${{ matrix.os }}
+
+  emscripten:
+    name: emscripten
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.11.0-beta.1
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: wasm32-unknown-emscripten
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
+      - run: pip install nox
+      - uses: actions/cache@v3
+        id: cache
+        with:
+          path: |
+            .nox/emscripten
+          key: ${{ hashFiles('emscripten/*') }} -  ${{ hashFiles('noxfile.py') }}
+      - uses: Swatinem/rust-cache@v1
+        with:
+          key: cargo-emscripten-wasm32
+      - name: Build
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: nox -s build_emscripten
+      - name: Test
+        run: nox -s test_emscripten
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,3 +168,8 @@ members = [
 no-default-features = true
 features = ["macros", "num-bigint", "num-complex", "hashbrown", "serde", "multiple-pymethods", "indexmap", "eyre"]
 rustdoc-args = ["--cfg", "docsrs"]
+
+[patch.crates-io]
+# Instant misspells emscripten_get_now by including a leading underscore.
+# https://github.com/sebcrozet/instant/pull/47
+instant = { git = 'https://github.com/hoodmane/instant/', branch= 'emscripten-no-leading-underscore' }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,8 +168,3 @@ members = [
 no-default-features = true
 features = ["macros", "num-bigint", "num-complex", "hashbrown", "serde", "multiple-pymethods", "indexmap", "eyre"]
 rustdoc-args = ["--cfg", "docsrs"]
-
-[patch.crates-io]
-# Instant misspells emscripten_get_now by including a leading underscore.
-# https://github.com/sebcrozet/instant/pull/47
-instant = { git = 'https://github.com/hoodmane/instant/', branch= 'emscripten-no-leading-underscore' }
--- a/emscripten/Makefile
+++ b/emscripten/Makefile
@@ -1,0 +1,81 @@
+CURDIR=$(abspath .)
+
+BUILDROOT ?= $(CURDIR)/builddir
+export EMSDKDIR = $(BUILDROOT)/emsdk
+
+PLATFORM=wasm32_emscripten
+SYSCONFIGDATA_NAME=_sysconfigdata__$(PLATFORM)
+
+# BASH_ENV tells bash to run pyodide_env.sh on startup, which sets various
+# environment variables. The next line instructs make to use bash to run each
+# command.
+export BASH_ENV := $(CURDIR)/env.sh
+SHELL := /bin/bash
+
+EMSCRIPTEN_VERSION=3.1.13
+
+PYMAJORMINORMICRO ?= 3.11.0
+PYPRERELEASE ?= b1
+
+version_tuple := $(subst ., ,$(PYMAJORMINORMICRO:v%=%))
+export PYMAJOR=$(word 1,$(version_tuple))
+export PYMINOR=$(word 2,$(version_tuple))
+export PYMICRO=$(word 3,$(version_tuple))
+PYVERSION=$(PYMAJORMINORMICRO)$(PYPRERELEASE)
+PYMAJORMINOR=$(PYMAJOR).$(PYMINOR)
+
+
+PYTHONURL=https://www.python.org/ftp/python/$(PYMAJORMINORMICRO)/Python-$(PYVERSION).tgz
+PYTHONTARBALL=$(BUILDROOT)/downloads/Python-$(PYVERSION).tgz
+PYTHONBUILD=$(BUILDROOT)/build/Python-$(PYVERSION)
+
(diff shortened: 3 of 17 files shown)


hoodmane:
So the segfault seems to be because `PyCapsule_Import("datetime.datetime_CAPI")` returns `NULL`.

hoodmane:
It would be good to return an error from `PyDateTime_IMPORT` if the import fails.

emscripten/Makefile
```diff
@@ -0,0 +1,81 @@
...
+	cd $(PYTHONBUILD) && \
+	CONFIG_SITE=Tools/wasm/config.site-wasm32-emscripten \
+  	emconfigure ./configure -C \
+		--host=wasm32-unknown-emscripten \
+		--build=$(shell $(PYTHONBUILD)/config.guess) \
+		--with-emscripten-target=browser \
+		--enable-wasm-dynamic-linking \
+		--with-build-python=python3.11
+
+$(PYTHONLIBDIR)/libpython$(PYMAJORMINOR).a : $(PYTHONBUILD)/Makefile
```
messense, line 69: I wonder if we can cache this artifact using [actions/cache](https://github.com/actions/cache) to reduce CI time.

.github/workflows/ci.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - emscripten
```
messense, line 7: This should be removed.

hoodmane: While working on the PR it's helpful if I can trigger CI on my fork.

emscripten/Makefile
hoodmane: Should be able to. I will try and see how it goes.

messense (approved):

davidhewitt:
Thanks, will try to give full review later!

> It would be good to return an error from `PyDateTime_IMPORT` if the import fails.

We shouldn't change the `ffi` symbol, however in `ensure_datetime_api` in `types/datetime.rs` we should _definitely_ assert the API is non-null after import.

hoodmane:
> Maybe worth correcting as a separate PR and documenting as a fix in the CHANGELOG.

Okay I will open a pr. I guess I should add a changelog entry here too.

tests/test_class_basics.rs
```diff
@@ -229,6 +229,7 @@ impl UnsendableChild {
     }
 }
 
+#[cfg_attr(target_arch = "wasm32", ignore)]
```
davidhewitt, line 232: This isn't a `#[test]` function so doesn't need the attribute.

```suggestion
```

tests/test_class_basics.rs
```diff
@@ -259,6 +260,7 @@ fn test_unsendable<T: PyClass + 'static>() -> PyResult<()> {
 
 /// If a class is marked as `unsendable`, it panics when accessed by another thread.
 #[test]
+#[cfg_attr(target_arch = "wasm32", ignore)]
```
davidhewitt, line 263: It might be nice to document this case as we know it's not expected to work.

```suggestion
#[cfg_attr(target_arch = "wasm32", ignore = "panic unwinding is not implemented on wasm")]
```

tests/test_class_basics.rs
```diff
@@ -267,6 +269,7 @@ fn panic_unsendable_base() {
 }
 
 #[test]
+#[cfg_attr(target_arch = "wasm32", ignore)]
```
davidhewitt, line 272: Similar here 

```suggestion
#[cfg_attr(target_arch = "wasm32", ignore = "panic unwinding is not implemented on wasm")]
```

tests/test_compile_error.rs
```diff
@@ -1,13 +1,15 @@
 #![cfg(feature = "macros")]
 
 #[rustversion::stable]
+#[cfg_attr(target_arch = "wasm32", ignore)]
```
davidhewitt, line 4: Similarly, for this and others in the file...

```suggestion
#[cfg_attr(target_arch = "wasm32", ignore = "cannot invoke rustc for UI tests with wasm")]
```

davidhewitt (commented):
This is really neat; I'm very excited to see this added to our CI. Thank you so much for putting the time in for this! (I really need to get my head around how shared libraries work in wasm sometime.)

> I guess I should add a changelog entry here too.

As strictly speaking there's not anything user-facing here I don't think a CHANGELOG is necessary for this one. Any subsequent MRs where we fix issues with wasm which users will encounter would be worth documenting. (e.g. the datetime fix.)

I've added a few comments - I think we can drop the `instant` patch, and also I'd like to have `ignore = "..."` notes on tests which we know will never work.

Let's keep #2412 open after this merges, and then the remaining tests which we believe _should_ work but currently don't can be tracked there. Hopefully we can close everything off before the 0.17 release. (I've still got a couple of axes of feature work I'd like to finish first, so there's some time!)

mejrs:
> I'd like to have ignore = "..." notes on tests which we know will never work.

I've always used `#[ignore]` to disable tests that are very expensive to run, rather than to not run tests that don't work at all. After all, what is the point of `cargo test -- --ignored` then? So I would personally prefer:
```rust
#[cfg(not(target_arch = "wasm32"))]
#[test]
fn test() {
    // ...
}
```

davidhewitt:
> I've always used `#[ignore]` to disable tests that are very expensive to run, rather than to not run tests that don't work at all. After all, what is the point of c`argo test -- --ignored` then? So I would personally prefer:

Good point, yes, tests which don't ever expect to work should use `#[cfg(not(...))]` - my bad 😄

Cargo.toml
```diff
@@ -168,3 +168,8 @@ members = [
 no-default-features = true
 features = ["macros", "num-bigint", "num-complex", "hashbrown", "serde", "multiple-pymethods", "indexmap", "eyre"]
 rustdoc-args = ["--cfg", "docsrs"]
+
+[patch.crates-io]
+# Instant misspells emscripten_get_now by including a leading underscore.
+# https://github.com/sebcrozet/instant/pull/47
+instant = { git = 'https://github.com/hoodmane/instant/', branch= 'emscripten-no-leading-underscore' }
```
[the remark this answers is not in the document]
hoodmane: Well it is a two character change and I need it anyways for cryptography.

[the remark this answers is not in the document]
hoodmane: Yeah it seems that cryptography has `pyo3 = { version = "0.15.2" }`  which requires `parking_lot = "0.11.0"` which requires `instant = "0.1.9"`. Indeed parking lot 0.12 has dropped the requirement. I wonder if it would work for me to build cryptography with a dependency-override giving it parking lot 0.12.1.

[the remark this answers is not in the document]
davidhewitt: Ah :(

Yes, there's https://github.com/pyca/cryptography/pull/6935, but they're blocked there waiting until 3.6 usage has decayed to a level where they're willing to drop support.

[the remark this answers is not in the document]
hoodmane: Maybe I can make a PR asking them to include that. Was there any reason that pyo3 v0.15 pinned parking lot 0.11?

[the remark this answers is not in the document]
davidhewitt: I think just was probably just released before 0.12 existed

davidhewitt (approved):
This is really great, thanks again! 💯

hoodmane:
Thanks for the help and the quick reviews @messense and @davidhewitt! Hopefully I'll get to some followup PRs soon.

PR #2446, opened 2022-06-09
LinkedList<T> to PyList conversion

### LINKED ISSUE #2405: LinkedList<T> to PyList conversion
For some functions, I'd like to return a `LinkedList`. However, `IntoPy` is only implemented for `Vec<T>`, so I currently first convert to a `Vec`, then allow pyo3 to convert the `Vec` to a `PyList`. This seems a bit inefficient, if I'm properly understanding how the conversions happen under the hood. It seems like going directly from `LinkedList` to `PyList` would be more efficient. I am willing to implement this, if such a PR would be welcome here.

This project's awesome!

JPDye:
Resolves #2405

Has been mentioned on Discord that converting a LinkedList<T> to a PyList could be surprising for users.

### DIFF
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -370,6 +370,28 @@ where
     }
 }
 
+impl<T> ToPyObject for std::collections::LinkedList<T>
+where
+    T: ToPyObject,
+{
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        let mut iter = self.iter().map(|e| e.to_object(py));
+        let list = new_from_iter(py, &mut iter);
+        list.into()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


mejrs:
CI failure is because `impl<T, const N: usize> From<[T; N]> for LinkedList<T>` is only available for Rust 1.56+. `vec![1,2.3].into_iter().collect()` is a decent way to do it for a test, I think.

> What would the alternative be? As a data point `pybind11` converts `std::list` to Python lists. https://github.com/pybind/pybind11/blob/0e956a2e4f7838fc26f42a6399bc358ae7248134/include/pybind11/stl.h#L200

I mean, they're very different data structures. I'm not convinced that linkedlist -> pylist is the obvious thing to do.

PR #2447, opened 2022-06-10
Prototype: Runtime inspection and Pyi generation

CLOVIS-AI:
The goal of this PR is to add a new `inspect` feature to PyO3.
When this feature is enabled, the PyO3 macros extract information about the Python data structures.
The generated information can be accessed at runtime.

This feature can be used to implement Rust functions that behave the same as `dict(...)`, etc.
In particular, this feature enables the automatic generation of Interface files (.pyi) which can be used for type checking (via MyPy, PyCharm…) or documentation (after small modifications, they can be fed to Sphinx or other tools).

Current status:
- [ ] collect data for modules:
  - [ ] name
  - [ ] classes
  - [ ] top-level fields
  - [ ] submodules?
- [ ] collect data for classes:
  - [x] name
  - [ ] base class
  - [x] direct fields
  - [x] fields in impl blocks
- [ ] collect data for fields:
  - [x] name
  - [x] kind (method, static method, …)
  - [ ] parameters
    - [x] name
    - [ ] kind (regular, positional-only, vararg, keyword arg)
    - [ ] has a default value
    - [x] type
  - [x] return type
- [x] convert types from Rust to Python
  - [x] for PyO3 builtins (`PyDict` -> `dict`, `PyAny` -> `Any`…) (no generics information available)
  - [x] for Rust stdlib builtins (`Vec<Foo>` -> `List[Foo]`, `u16` -> `int`…) (generic information is available)
- [x] merge information from the `#[pyclass]` and the `#[pymethods]` blocks at runtime into a single `inspect()` method
- [ ] entry in CHANGELOG.md
- [ ] docs to all new functions and / or detail in the guide
- [ ] tests for all new or changed functions
- [ ] create the `inspect` feature and gate all new behaviors behind it
- [ ] interface generation
  - [ ] generate a valid syntax for a .pyi file for each module
  - [ ] example in the documentation to write a build script / unit test that generates it
- [x] replace the `&'static` lifetimes with normal lifetimes, so objects can be constructed at runtime (useful for editing the .pyi?)

Future possible enhancements:
- Include the documentation in the extracted information, so the interface files can be used for HTML documentation via Sphinx or other
- Find a way to make this compatible with `multiple-pymethods` (the current implementation only works for a single `#[pymethods]` block)
- Is there a way to have Maturin transparently handle the PyI generation?

### DIFF
--- a/pyo3-macros-backend/src/inspect.rs
+++ b/pyo3-macros-backend/src/inspect.rs
@@ -150,12 +150,35 @@ pub(crate) fn generate_fields_inspection(
         .map(|it| it.to_token_stream())
         .unwrap_or_else(|| cls.to_token_stream());
 
+    let mut args: Vec<TokenStream> = vec![];
+    for arg in &field.spec.args {
+        let name = Literal::string(&*arg.name.to_string());
+        let typ = generate_type(arg.ty)
+            .map(|it| it.to_token_stream())
+            .unwrap_or_else(|| cls.to_token_stream());
+
+        let is_mutable = arg.mutability.is_some();
+
+        args.push(quote! {
+            _pyo3::inspect::fields::ArgumentInfo {
+                name: #name,
+                kind: _pyo3::inspect::fields::ArgumentKind::PositionOrKeyword, //TODO
+                py_type: ::std::option::Option::Some(|| <#typ as _pyo3::conversion::FromPyObject>::type_input()),
+                default_value: false,
+                is_modified: #is_mutable,
+            }
+        });
+    }
+    let args_size = Literal::usize_suffixed(args.len());
+
     let output = quote! {
         fn #field_type_fn_name() -> _pyo3::inspect::types::TypeInfo {
             <#field_type as _pyo3::conversion::IntoPy<_>>::type_output()
         }
 
-        const #field_args_name: [_pyo3::inspect::fields::ArgumentInfo<'static>; 0] = []; //TODO
+        const #field_args_name: [_pyo3::inspect::fields::ArgumentInfo<'static>; #args_size] = [
+             #(#args),*
+        ];
 
         const #field_info_name: _pyo3::inspect::fields::FieldInfo<'static> = _pyo3::inspect::fields::FieldInfo {
             name: #field_name,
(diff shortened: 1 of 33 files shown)


CLOVIS-AI:
Current state of this prototype:
- First and foremost, I've been figuring out what needs to be done as I go from one thing to the next, so the commit history / code is not greatly organized.
- I'm still learning how to write proc-macros, so there's a high likelihood that I'm breaking other things along the way. (and a lot of unit tests are broken)
- This prototype seems to show that the objective is possible.

Current generation:
```rust

#[pyclass]
struct Complicated {
    #[pyo3(get, set)]
    value: usize,
}

#[allow(unused_variables)]
#[pymethods]
impl Complicated {
    #[new]
    fn new(foo: PyObject, parent: PyObject) -> Self {
        unreachable!("This is just a stub")
    }

    #[pyo3(name = "staticmeth")]
    #[staticmethod]
    fn static_method(input: PyObject) -> Complicated {
        unreachable!("This is just a stub")
    }

    #[pyo3(name = "classmeth")]
    #[classmethod]
    fn class_method(cls: &PyType, input: ClassMethodInput) -> Complicated {
        unreachable!("This is just a stub")
    }
}

#[derive(FromPyObject)]
enum ClassMethodInput {
    // Complicated(PyCell<Complicated>),
    String(String),
    Int(usize),
}
```
generates:
```python
class Complicated:
    @property
    def value(self) -> int: ...

    @value.setter
    def value(self, value: int) -> None: ...

    def __new__(cls, /, foo: PyAny, parent: PyAny) -> None: ...

    @staticmethod
    def staticmeth(input: PyAny) -> Any: ...

    @classmethod
    def classmeth(cls, /, input: Any) -> Any: ...
```

CLOVIS-AI:
Hi @davidhewitt!

I opened #2454 to discuss the various ideas of this prototype. I think it's is good enough to show what the steps needed are, but it's far from perfect so I'd like your feedback on what the good/bad parts, so I can refactor out the good parts into independent PRs so we can finally stabilize this :)

beauxq:
> ```python
> def __new__(cls, /, foo: PyAny, parent: PyAny) -> None: ...
> ```

I think `__new__` normally returns `typing.Self`, rather than `None`

```python
class C:
    def __new__(cls, a: int) -> None:
        ...


c = C(3)
reveal_type(c)  # Type of "c" is "None"
```

It might be better to translate Rust's `new` to Python's `__init__`, which does normally return `None`.
But `__new__` could work if the return type is annotated to `Self`.

davidhewitt:
Sorry this got stuck for so long. We are now making progress on #5137 and this PR has lots of conflicts, so I will close this. Looking forward to getting pyi generation working... eventually!

PR #2449, opened 2022-06-11
xtask: expand on coverage

mejrs:


xtask/src/cli.rs
```diff
@@ -35,9 +35,9 @@ impl Default for Subcommand {
 
 #[derive(StructOpt, Default)]
 pub struct CoverageOpts {
-    /// Creates an lcov output instead of printing to the terminal.
```
mejrs, line 38: Was anyone actually printing it to terminal? It didn't seem that useful for me, but I'm happy to put it back if necessary.

davidhewitt: I hadn't used in practice, so fine to remove imo!

xtask/src/cli.rs
```diff
@@ -125,22 +127,23 @@ pub fn run(command: &mut Command) -> Result<()> {
         println!("Running: {}", command_str);
     }
 
-    let status = command.spawn()?.wait()?;
+    let output = command.output()?;
```
davidhewitt, line 130: Hmmm I think this change isn't right. This changes things to swallow all the output from lots of `xtask` commands, e.g. the Python examples https://github.com/PyO3/pyo3/runs/6846083356?check_suite_focus=true#step:20:97

Maybe there needs to be two variants of `run`? `run` and `run_output` ?

mejrs: Oops, thanks. I've created it.

xtask/src/cli.rs
```diff
@@ -116,6 +125,7 @@ impl Subcommand {
     }
 }
 
+// Run a command as a child process, inheriting stdin, stdout and stderr.
```
davidhewitt, line 128:
```suggestion
/// Run a command as a child process, inheriting stdin, stdout and stderr.
```

Contributing.md
```diff
@@ -128,6 +128,29 @@ First, there are Rust-based benchmarks located in the `benches` subdirectory. As
...
+## Code coverage
+
+You can view what code is and isn't covered by PyO3's tests. We aim to have 100% coverage - please check coverage and add tests if you notice a lack of coverage!
+
+- First, generate a `lcov.info` file with 
+```shell
+cargo xtask coverage
+```
+You can install an IDE plugin to view the coverage. For example, if you use VSCode:
+- Add the [coverage-gutters](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters) plugin.
```
davidhewitt, line 140: This sounds cool, I'm excited to test this when I next have a quiet moment with my laptop!

Contributing.md
```diff
@@ -128,6 +128,29 @@ First, there are Rust-based benchmarks located in the `benches` subdirectory. As
 
 Second, there is a Python-based benchmark contained in the `pytests` subdirectory. You can read more about it [here](pytests).
 
+## Code coverage
```
davidhewitt, line 131: 👍

davidhewitt (approved):
Thanks, great to make coverage more accessible for new contributors!

I've got just a couple formatting nits which I'll commit using the GitHub UI and then squash.

Contributing.md
mejrs: Please feel free to add information for other IDEs :)

PR #2450, opened 2022-06-12
Fix UB in *_offset functions

mejrs:
If the `addr_of` macro is not available, this created a zeroed `PyCell<T>`, but arbitrary `T`'s do not permit zero-initialization.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ edition = "2018"
 cfg-if = "1.0"
 libc = "0.2.62"
 parking_lot = ">= 0.11, < 0.13"
+memoffset = "0.6.5"
 
 # ffi bindings to the python interpreter, split into a separate crate so they can be used independently
 pyo3-ffi = { path = "pyo3-ffi", version = "=0.16.5" }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix FFI definitions `PyTypeObject`. `PyHeapTypeObject`, and `PyCFunctionObject` having incorrect members on PyPy 3.9. [#2428](https://github.com/PyO3/pyo3/pull/2428)
 - Fix FFI definition `PyGetSetDef` to have `*const c_char` for `doc` member (not `*mut c_char`). [#2439](https://github.com/PyO3/pyo3/pull/2439)
 - Fix `#[pyo3(from_py_with = "...")]` being ignored for 1-element tuple structs and transparent structs. [#2440](https://github.com/PyO3/pyo3/pull/2440)
+- Use `memoffset` for computing PyCell offsets [#2450](https://github.com/PyO3/pyo3/pull/2450)
 
 ## [0.16.5] - 2022-05-15
 
--- a/src/ffi/tests.rs
+++ b/src/ffi/tests.rs
@@ -135,7 +135,7 @@ fn ascii() {
             // _PyUnicode_NONCOMPACT_DATA isn't valid for compact strings.
             assert!(!PyUnicode_DATA(ptr).is_null());
 
-            assert_eq!(PyUnicode_GET_LENGTH(ptr), s.len().unwrap() as _);
+            assert_eq!(PyUnicode_GET_LENGTH(ptr), s.len().unwrap() as Py_ssize_t);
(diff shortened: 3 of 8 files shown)


adamreichold:
I suspect it had been considered when this code was written, but isn't [`memoffset`](https://docs.rs/memoffset/latest/memoffset/) the central place where these workarounds are collected?

mejrs:
> I suspect it had been considered when this code was written, but isn't [`memoffset`](https://docs.rs/memoffset/latest/memoffset/) the central place where these workarounds are collected?

`memoffset` just chooses the least UB way to do this, and it cannot avoid UB if `addr_of!` is unavailable. The fallback as written here is not UB.

FWIW I suspect the real reason that `memoffset` was not used is to avoid adding a dependency.

danielhenrymantilla:
`memoffset` shall never reach exploited UB using `cargo + rustc`:

  - before `addr_of!`, such fixed list of rustc implementations did not exploit its UB;
  - after `addr_of!`, it no longer has UB.

It is thus as fine as it could possibly be.

The current PR duplicates a structure's layout, which means these could theoretically end up out of sync as the codebase is updated, which _in a vacuum_ (I don't know the constraints of PyO3), seems orders of magnitude more error-prone than using an unofficial Rust implementation with no `addr_of!` which would exploit that UB

___

One approach to make the code robust to original struct updates, while keeping the current interesting stratefy, would be:

```rs
type Apply<Wrapper, T> = <Wrapper as CanWrap<T>>::T;

trait CanWrap<T> : 'static { type T; }

/// For any `T`, `Apply<Unwrapped, T> = T`.
enum Unwrapped {}
impl<T> CanWrap<T> for Unwrapped { type T = T; }

/// For any `T`, `Apply<InManuallyDrop, T> = ManuallyDrop<T>`
enum InManuallyDrop {}
impl<T> CanWrap<T> for InManuallyDrop<T> { type T = ManuallyDrop<T>; }

#[repr(C)]
pub(crate) struct PyCellContents<T: PyClassImpl, Wrapper = Unwrapped>
where
    // this would not be needed if we had (type) GATs
    Wrapper : CanWrap< ManuallyDrop<UnsafeCell<T>> >,
    Wrapper : CanWrap< <T::PyClassMutability as PyClassMutability>::Storage >,
    Wrapper : CanWrap< T::ThreadChecker >,
    Wrapper : CanWrap< T::Dict >,
    Wrapper : CanWrap< T::WeakRef >,
{
    pub(crate) value: Apply<Wrapper, ManuallyDrop<UnsafeCell<T> >,
    pub(crate) borrow_checker: Apply<Wrapper, <T::PyClassMutability as PyClassMutability>::Storage >,
    pub(crate) thread_checker: Apply<Wrapper, T::ThreadChecker >,
    pub(crate) dict: Apply<Wrapper, T::Dict >,
    pub(crate) weakref: Apply<Wrapper, T::WeakRef >,
}
```

that way, the `MaybeUninit`-projected struct could be queried by using `PyCellContents<T, InManuallyDrop>`, and would be guaranteed to be kept in sync.

  - (Sadly, without (type) GATs, this is a mouthful, so I still think that using `memoffset` (or inlining its logic) is the least error-prone approach)

mejrs:
Hmm, I like simple, and just using memoffset sounds simple.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ edition = "2018"
 cfg-if = "1.0"
 libc = "0.2.62"
 parking_lot = ">= 0.11, < 0.13"
+memoffset = "0.6.5"
```
davidhewitt, line 20: Should we add a note here that it's to support safe offset calculations on versions predating `addr_of`? I imagine in the future we might consider removing again.

davidhewitt (commented):
Thanks, using `memoffset` sounds reasonable to me. Please add a CHANGELOG entry. Not sure whether this classes as `Packaging` or `Fixed` category.

Thanks for cleaning up the tests with flaky type inference issues too! 😁

Cargo.toml
adamreichold: I think we are using it compute offsets in all cases since even when `addr_of` is available, using `memoffset` enables us to avoid writing out its usage ourselves.

Personally, I think as it takes care of what is available where and how to use it, `memoffset` does carry its weight as a dependency.

davidhewitt: Agree with this, I was thinking more that after a future MSRV bump that may change as `addr_of` would be trivial for us to implement safely. I'll worry about that in the future, perhaps 😄

PR #2451, opened 2022-06-12
add PyString::chars

robinvd:
Useful helper method to avoid allocating anything while iterating over a PyUnicode string. The AsUTF8 methods will allocate and cache the utf8 str on the python heap.

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -4,10 +4,11 @@
 use crate::exceptions::PyUnicodeDecodeError;
 use crate::types::PyBytes;
 use crate::{
-    ffi, AsPyPointer, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, PyTryFrom, Python,
-    ToPyObject,
+    exceptions, ffi, AsPyPointer, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, PyTryFrom,
+    Python, ToPyObject,
 };
 use std::borrow::Cow;
+use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::str;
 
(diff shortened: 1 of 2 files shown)


src/types/string.rs
```diff
@@ -227,6 +227,29 @@ impl PyString {
         }
     }
 
+    /// Returns an iterator over the PyString.
+    ///
+    /// Does not allocate anything (python or rust heap).
+    pub fn chars(&self) -> impl ExactSizeIterator<Item = PyResult<char>> + '_ {
+        unsafe {
+            let len = ffi::PyUnicode_GetLength(self.as_ptr());
+            (0..len).map(move |i| {
```
adamreichold, line 236: Since the implementation is based on indexing, would it make sense to expose this a the interface as well? Something like `char_at(&self, index: usize) -> Option<char>` so that the iterator can be produced on the outside?

mejrs:
Thanks for the PR :)

> The AsUTF8 methods will allocate and cache the utf8 str on the python heap.

- it won't if the string is ascii, which most strings are.
-  calling `PyUnicode_AsUTF8`/`PyUnicode_AsUTF8AndSize` caches the utf8 representation (as you mention), so any subsequent calls do not allocate (note that this is not true for `PyString::to_str` on its slower path - I'll fix that in another PR).

I've benchmarked this method and found that it generally is around twice as slow on ascii strings compared to `to_str().chars()`. It is only faster if the string is not ascii and long, which is not representative of most strings. For example, https://peps.python.org/pep-0393/ mentions: "out of 36,000 strings (with 1,310,000 chars), 35713 where ASCII strings". I suspect these figures are similar for other applications.

robinvd:
In my benchmarks this is slightly faster. But my usecase is quite specific. I have to modify some unicode chars. So any ascii data just gets returned as is. And all strings passed in are newly allocated and thus have no utf8 string cached.

As it is slower in the general case. Maybe its best to close this.

mejrs:
I see a use case where it forwards to `to_str().chars()` on ascii strings, but falls back to your method if it's not. The trouble with this approach is that there's no good way to check which representation it has - that information is stored in a  C bitfield 😭 . (See also the documentation of https://pyo3.rs/internal/doc/pyo3/types/struct.PyString.html#method.data)

PR #2457, opened 2022-06-16
Enum naming improvements

yodaldevoid:
This improves the interaction between `#[pyclass(name)]` and `enum`s.

First, the macro-defined name for the `enum` is used in the `enum`'s `__repr__` implementation so that everything is kept consistent from the Python point of view.

Secondly, `#[pyo3(name)]` is now supported on enum variants.


pyo3-macros-backend/src/pyclass.rs
```diff
@@ -407,6 +407,57 @@ pub fn build_py_enum(
     Ok(impl_enum(enum_, &args, doc, method_type))
 }
 
+fn get_variant_python_name<'a>(
```
messense, line 410:
```suggestion
fn get_enum_variant_python_name<'a>(
```

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -407,6 +407,57 @@ pub fn build_py_enum(
...
+) -> Cow<'a, syn::Ident> {
+    options
+        .name
+        .as_ref()
+        .map(|name_attr| Cow::Borrowed(&name_attr.value.0))
+        .unwrap_or_else(|| Cow::Owned(variant.unraw()))
+}
+
+/// `#[pyo3()]` options for pyclass enum variants
+struct VariantPyO3Options {
```
messense, line 422:
```suggestion
struct EnumVariantPyO3Options {
```

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -407,6 +407,57 @@ pub fn build_py_enum(
...
+    name: Option<NameAttribute>,
+}
+
+enum VariantPyO3Option {
```
messense, line 426:
```suggestion
enum EnumVariantPyO3Option {
```

messense (commented):
I think it's better to name some struct and fns `enum variant` instead of just `variant`.

yodaldevoid:
> I think it's better to name some struct and fns `enum variant` instead of just `variant`.

I've applied the naming suggestions.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -338,11 +338,11 @@ struct PyClassEnum<'a> {
     // The underlying #[repr] of the enum, used to implement __int__ and __richcmp__.
     // This matters when the underlying representation may not fit in `isize`.
     repr_type: syn::Ident,
-    variants: Vec<PyClassEnumVariant<'a>>,
+    variants: Vec<(PyClassEnumVariant<'a>, EnumVariantPyO3Options)>,
```
davidhewitt, line 341: Can `EnumVariantPyO3Options` be added to `PyClassEnumVariant`? Would be a bit nicer to have a `.options` field imo rather than a vec of tuples

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -407,6 +407,57 @@ pub fn build_py_enum(
     Ok(impl_enum(enum_, &args, doc, method_type))
 }
 
+fn get_enum_variant_python_name<'a>(
```
davidhewitt, line 410: Could probably be a method `PyClassEnumVariant::python_name` if options were stored in the struct as per above.

davidhewitt (commented):
Thanks very much for this, and sorry I have been slow to review.

Implementation is very complete and docs are great 👍

Just one tiny tweak I'd like to see to contain the options inside the `PyClassEnumVariant` struct (if it's straightforward to do so). Then let's merge!

pyo3-macros-backend/src/pyclass.rs
yodaldevoid: I implemented it this way to match what is done with struct fields. This would make it easier if in the future we want to replace `PyClassEnumVariant` with `syn::Field` as with struct fields. I think this would mainly be needed when adding support for non-C-like enums.

yodaldevoid:
I've incorporated the suggested changes and fixed the merge conflict in the CHANGELOG.

messense:
```
error: explicit lifetimes given in parameter types where they could be elided (or replaced with `'_` if needed by type declaration)
--> pyo3-macros-backend/src/pyclass.rs:337:5
|
337|     fn python_name<'b>(&'b self) -> Cow<'b, syn::Ident> {
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|
= note: `-D clippy::needless-lifetimes` implied by `-D warnings`
= help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
```

There is a clippy warning needs to be addressed.

davidhewitt (approved):
Looks great - many thanks again!

PR #2460, opened 2022-06-17
add `CompareOp::convert`

birkenfeld:
to easily implement `__richcmp__` as the result of a Rust `std::cmp::Ordering` comparison.


birkenfeld:
Questions:

* is this name obvious enough?
* should it take `Option<Ordering>` to allow for `partial_cmp` results?

CLOVIS-AI:
What about `op.from_ordering(…)` instead of `op.convert(…)`?

I don't think `convert` is clear enough on what it means.

birkenfeld:
`from_` methods normally construct an instance of their type, which this doesn't.

mejrs:
> * is this name obvious enough?

What about `matches`? Or could this be a PartialEq impl?

birkenfeld:
`matches` works for me! Not sure about `PartialEq`, it might be confusing that `Ordering::Less == CompareOp::Le` etc.

mejrs (approved):
LGTM

davidhewitt:
👍 good idea, thanks for this & sorry I've been slow on reviews recently!

PR #2461, opened 2022-06-18
macros: refactor codegen to remove PyMethodDefType

davidhewitt:
Just a bit of playing around to remove `PyMethodDefType` by making types in `pyo3-macros-backend` a bit more concrete and growing the `PyClassItems` struct to have explicit `class_attributes`, `getters` and `setters` fields. Think overall it leads to simpler generated code which is always nice.

As it is, the macro code changes on this branch are a bit messy - introduces a lot of types which are almost the same and the `Vec<(Vec<&'a syn::Attribute>, MethodAndMethodDef)>` having tuples feels a bit untidy.

Don't think this is very important to proceed with at all as it was primarily a refactor, so just pushing in case it's interesting for the future. Might close it again before long.


davidhewitt:
I think this is very stale. While it'd be nice to clean this up and move forward with it, it's just a refactoring and there are many higher priority things to be working on.

PR #2462, opened 2022-06-19
Increasing test coverage

aganders3:
Starting with `src/buffer.rs` - some simple tests in there to exercise the debug code and the type matching functions.

I also added some calls to the fortran-order-specific functions in `test_array_buffer`. These are pretty uninteresting because `array` produces a 1D buffer, but I don't know how to generate a higher-dim buffer with the standard library. Maybe I could just alloc and fill some memory with Rust and create a `PyBuffer` without using `PyBuffer::get()`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ categories = ["api-bindings", "development-tools::ffi"]
 license = "Apache-2.0"
 exclude = ["/.gitignore", ".cargo/config", "/codecov.yml", "/Makefile", "/pyproject.toml", "/noxfile.py"]
 edition = "2018"
-rust-version = "1.48"
 
 [dependencies]
 cfg-if = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ categories = ["api-bindings", "development-tools::ffi"]
 license = "Apache-2.0"
 exclude = ["/.gitignore", ".cargo/config", "/codecov.yml", "/Makefile", "/pyproject.toml", "/noxfile.py"]
 edition = "2018"
+rust-version = "1.48"
 
 [dependencies]
 cfg-if = "1.0"
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -687,37 +687,26 @@ mod tests {
 
     #[test]
     fn test_debug() {
-        // PyBuffer {
-        //    buf: 0x00000001012c65d0,
-        //    obj: 0x00000001012c65b0,
-        //    len: 5,
-        //    itemsize: 1,
-        //    readonly: 1,
-        //    ndim: 1,
-        //    format: 0x00000001016747d8,
-        //    shape: 0x0000600001b74010,
-        //    strides: 0x0000600001b74018,
-        //    suboffsets: 0x0000000000000000,
-        //    internal: 0x0000000000000000,
-        //}
         Python::with_gil(|py| {
             let bytes = py.eval("b'abcde'", None, None).unwrap();
             let buffer: PyBuffer<u8> = PyBuffer::get(bytes).unwrap();
-
-            let dbg = format!("{:?}", buffer);
-
-            assert!(dbg.starts_with("PyBuffer"));
-            assert!(dbg.contains("{ buf: "));
-            assert!(dbg.contains(", obj: "));
-            assert!(dbg.contains(", len: 5"));
-            assert!(dbg.contains(", itemsize: 1"));
-            assert!(dbg.contains(", readonly: 1"));
-            assert!(dbg.contains(", ndim: 1"));
-            assert!(dbg.contains(", format: "));
-            assert!(dbg.contains(", shape: "));
-            assert!(dbg.contains(", strides: "));
-            assert!(dbg.contains(", suboffsets: "));
-            assert!(dbg.contains(", internal: "));
+            let expected = format!(
+                concat!(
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -12,6 +12,7 @@ categories = ["api-bindings", "development-tools::ffi"]
 license = "Apache-2.0"
 exclude = ["/.gitignore", ".cargo/config", "/codecov.yml", "/Makefile", "/pyproject.toml", "/noxfile.py"]
 edition = "2018"
+rust-version = "1.48"
```
mejrs, line 15: Unfortunately this produces a warning on rust < 1.56: https://github.com/PyO3/pyo3/runs/6972028061?check_suite_focus=true#step:13:18 , so this won't be useful until we actually raise msrv above that :(

mejrs (commented):
>  I don't know how to generate a higher-dim buffer with the standard library.

Most array crates and numpy for example, will store everything in a continuous buffer, See https://en.wikipedia.org/wiki/Row-_and_column-major_order

aganders3:
> >  I don't know how to generate a higher-dim buffer with the standard library.
> 
> 
> 
> Most array crates and numpy for example, will store everything in a continuous buffer, See https://en.wikipedia.org/wiki/Row-_and_column-major_order 

Yeah this is what I meant. I am familiar with `numpy` as a producer in the Buffer Protocol and for n-dimensional arrays it allocates a block but may have more "interesting" values in dimensions, strides, shape, etc. for the PyBuffer. Unfortunately as far as I know the only (Python) standard library types that implement the Buffer Protocol are `bytes`, `bytearray`, and `array.array`; all 1D so don't provide much variability for writing tests. I don't think this is necessarily enough motivation to add `numpy` or something to the test environment though.

Cargo.toml
adamreichold: I don't think that warning is a reason to revert this as it is limited to building PyO3 itself and will not bother dependencies using the MSRV. And it is useful machine-readable and highly discoverable human-readable metadata.

davidhewitt:
Nice work! For 2D - you could potentially write a `#[pyclass]` which contains a `Vec` and a `__getbuffer__` implementation which describes it as a 2D buffer?

Cargo.toml
davidhewitt: Yeah I'm torn on this one. Can we think if there any cases where the warning might lead to build errors?

adamreichold: I am not aware of situations where warnings in dependencies would be upgraded to errors. At least when building using Cargo and going out of one's way by e.g. grepping build logs instead of using Cargo/rustc flags to deny warnings.

adamreichold: (Just as another data point to make this more palatable: `rust-numpy` has this since 0.16.0 with no reports so far.)

aganders3: Thanks - either way I agree I will revert from this PR and make a separate one. I wasn't trying to sneak it in 😛.

src/types/function.rs
```diff
@@ -74,7 +74,7 @@ impl PyCFunction {
         )
     }
 
-    /// Create a new built-in function without keywords.
+    /// Create a new built-in function without keywords (no arguments).
```
aganders3, line 77: This was confusing to me as it calls `&PyMethodDef::noargs()` below - this might be a quirk of the Python C-API which I'm not super familiar with.

src/buffer.rs
```diff
@@ -685,6 +685,171 @@ mod tests {
...
+            assert!(dbg.contains(", obj: "));
+            assert!(dbg.contains(", len: 5"));
+            assert!(dbg.contains(", itemsize: 1"));
+            assert!(dbg.contains(", readonly: 1"));
+            assert!(dbg.contains(", ndim: 1"));
+            assert!(dbg.contains(", format: "));
+            assert!(dbg.contains(", shape: "));
+            assert!(dbg.contains(", strides: "));
+            assert!(dbg.contains(", suboffsets: "));
+            assert!(dbg.contains(", internal: "));
```
davidhewitt, line 720: Checking the fragments to avoid the fact the pointers are different each time I guess? I wonder if an alternative could be to replace the pointers with `<some_ptr>` or similar so that you can do a `==` on the whole `dbg` string?

src/types/function.rs
davidhewitt: Good catch. How about making this even more explicit?

```suggestion
    /// Create a new built-in function which takes no arguments.
```

davidhewitt (commented):
Looks great - just suggested some quick tidy ups and then I'm happy to merge this. Nearly +1% coverage, lovely!

> I'm now not so convinced about the value

Fair enough, if it's painful and doesn't add much no need to toil over it.

src/buffer.rs
aganders3: Yeah I started with regex but it's only a secondary dependency. Replacing the pointers with known values is a good idea - I'll try that.

src/types/function.rs
aganders3: That's good I just wasn't sure if "keywords" had some meaning in the C-API that might be helpful for others with that experience.

src/buffer.rs
aganders3: I refactored to use a big format string to be able to `==` the whole thing. I think this is better but by nature this test may be a little tautological. At least it should catch changes in behavior.

davidhewitt (approved):
Looks brilliant, thanks again!

PR #2468, opened 2022-06-21
Add frozen documentation

mejrs:


pyo3-macros/src/lib.rs
```diff
@@ -79,21 +79,6 @@ pub fn pyproto(_: TokenStream, input: TokenStream) -> TokenStream {
     .into()
 }
 
-/// A proc macro used to expose Rust structs and fieldless enums as Python objects.
-///
-#[cfg_attr(docsrs, cfg_attr(docsrs, doc = include_str!("../docs/pyclass_parameters.md")))]
```
mejrs, line 84: This doesn't seem to actually work across crate boundaries 🤔

src/pyclass.rs
```diff
@@ -593,6 +595,7 @@ pub mod boolean_struct {
 }
 
 /// A trait which is used to describe whether a `#[pyclass]` is frozen.
+#[doc(hidden)]
```
mejrs, line 598: I've hidden this because it just seems...weird to be browsing the documentation and finding this. Imo this just just an implementation detail.

guide/src/class.md
```diff
@@ -195,7 +195,7 @@ Python::with_gil(|py|{
 
 ## Customizing the class
 
-{{#include ../../pyo3-macros/docs/pyclass_parameters.md}}
+{{#include ../pyclass_parameters.md}}
 
 [params-1]: {{#PYO3_DOCS_URL}}/pyo3/prelude/struct.PyAny.html
```
davidhewitt, line 200: You added these links to `pyclass_parameters.md` - does that create a problem with the duplicate definitions?

davidhewitt (commented):
Thanks, looks good to me! One brief concern that I wanted to double-check.

guide/src/class.md
mejrs: I don;t think so, but I've fixed it.

src/pyclass.rs
```diff
@@ -16,15 +16,16 @@ use std::{
...
-/// so you normally don't have to use this trait directly.
+/// The `#[pyclass]` attribute implements this trait for your Rust struct -
+/// you shouldn't implement this trait directly.
 pub trait PyClass:
     PyTypeInfo<AsRefTarget = PyCell<Self>> + PyClassImpl<Layout = PyCell<Self>>
 {
-    /// Frozen or not
+    /// Whether the pyclass is frozen.
+    ///
+    /// This can be enabled via `#[pyclass(frozen)]`.
```
davidhewitt, line 28: 👍 so much better than what I had written 😂

guide/src/class.md
davidhewitt: I think I slightly preferred the (slightly tedious) duplication of having the links written out in `class.md` and `lib.rs` because that way the links were mostly adjusted to link within the current version of the docs than always be stuck pointing at `latest`, though it's not a huge deal.

davidhewitt (approved):
Thanks!

Just a quick note on why I had originally written the duplication - feel free to decide whether or not you felt the duplication had benefit & then (squash?) merge!

PR #2471, opened 2022-06-23
readme: Adding an example to the README

jinlow:
I wanted to add this library to the examples, it's a machine learning library I have been working on to create gradient boosted decision trees, which uses pyo3 to wrap the rust code to create a python module.


messense (approved):
👍 Thanks!

PR #2477, opened 2022-06-24
Update PyTryFrom for PyMapping and PySequence to more accurately check types

### LINKED ISSUE #2072: PyMapping is inconsistent with typing.Mapping

aganders3:
This is intended to fix #2072 - see further discussion in that thread.

The implementation is such that for both `PyMapping` and `PyTryFrom`, they will call into python to check `isinstance(<obj>, collections.abc.Mapping)` before successfully downcasting. Please note this is a breaking change so it is important the documentation and migration guide are clear.

Also note the implementation has changed from the initial draft of this PR. See discussion below and in #2072 for more information.
 
### TODO
- [x]  an entry in CHANGELOG.md
- [x]  docs to all new functions and / or detail in the guide
- [x]  tests for all new or changed functions

### DIFF
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -5,6 +5,59 @@ For a detailed list of all changes, see the [CHANGELOG](changelog.md).
 
 ## from 0.16.* to 0.17
 
+### Downcasting (`PyTryFrom`) has been changed for `PyMapping` and `PySequence` types
+
+Previously the safety checks for downcasting to `PyTryFrom` and `PyMapping`
+used the Python C-API functions `PyMapping_Check` and `PySequence_Check`.
+Unfortunately these functions are not sufficient for distinguishing such types,
+leading to inconsistent behavior (see
+[pyo3/pyo3#2072](https://github.com/PyO3/pyo3/issues/2072)).
+
+PyO3 0.17 changes these downcast checks to explicityly test if the type is a
+subclass of the corresponding abstract base class `collections.abc.Mapping` or
+`collections.abc.Sequence`. Note this requires calling into Python, which may
+incur a performance penalty over the previous method. If this performance
+penatly is a problem, you may be able to perform your own checks and use
+`try_from_unchecked` (unsafe).
+
+Another side-effect is that a pyclass defined in Rust with PyO3 will need to
+be _registered_ with the corresponding Python abstract base class for
+downcasting to succeed. `PySequence::register_abc_subclass` and
+`PyMapping:register_abc_subclass` have been added to make it easy to do this
+from Rust code. These are equivalent to calling
+`collections.abc.Mapping.register(MappingPyClass)` or
+`collections.abc.Sequence.register(SequencePyClass)` from Python.
+
+For example, for a mapping class defined in Rust:
+```rust,compile_fail
+use pyo3::prelude::*;
+use std::collections::HashMap;
+
+#[pyclass(mapping)]
+struct Mapping {
+    index: HashMap<String, usize>,
+}
+
+#[pymethods]
+impl Mapping {
+    #[new]
+    fn new(elements: Option<&PyList>) -> PyResult<Self> {
+    // ...
+    // truncated implementation of this mapping pyclass - basically a wrapper around a HashMap
+}
+
+```
+
+You must register the class with `collections.abc.Mapping` before the downcast will work:
+```rust,compile_fail
+let m = Py::new(py, Mapping { index }).unwrap();
+assert!(m.as_ref(py).downcast::<PyMapping>().is_err());
+PyMapping::register_abc_subclass::<Mapping>(py).unwrap();
+assert!(m.as_ref(py).downcast::<PyMapping>().is_ok());
+```
+
+Note that this requirement may go away in the future when a pyclass is able to inherit from the abstract base class directly (see [pyo3/pyo3#991](https://github.com/PyO3/pyo3/issues/904)).
+
 ### The `multiple-pymethods` feature now requires Rust 1.62
 
 Due to limitations in the `inventory` crate which the `multiple-pymethods` feature depends on, this feature now
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -55,7 +55,7 @@ PyMapping::register::<Mapping>(py).unwrap();
 assert!(m.as_ref(py).downcast::<PyMapping>().is_ok());
 ```
 
-Note that this requirement may go away in the future when a pyclass is able to inherit from the abstract base class directly (see [pyo3/pyo3#991](https://github.com/PyO3/pyo3/issues/904)).
+Note that this requirement may go away in the future when a pyclass is able to inherit from the abstract base class directly (see [pyo3/pyo3#991](https://github.com/PyO3/pyo3/issues/991)).
 
 ### The `multiple-pymethods` feature now requires Rust 1.62
 
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -22,10 +22,9 @@ penatly is a problem, you may be able to perform your own checks and use
 
 Another side-effect is that a pyclass defined in Rust with PyO3 will need to
 be _registered_ with the corresponding Python abstract base class for
-downcasting to succeed. `PySequence::register_abc_subclass` and
-`PyMapping:register_abc_subclass` have been added to make it easy to do this
-from Rust code. These are equivalent to calling
-`collections.abc.Mapping.register(MappingPyClass)` or
+downcasting to succeed. `PySequence::register` and `PyMapping:register` have
+been added to make it easy to do this from Rust code. These are equivalent to
+calling `collections.abc.Mapping.register(MappingPyClass)` or
 `collections.abc.Sequence.register(SequencePyClass)` from Python.
 
 For example, for a mapping class defined in Rust:
@@ -52,7 +51,7 @@ You must register the class with `collections.abc.Mapping` before the downcast w
 ```rust,compile_fail
 let m = Py::new(py, Mapping { index }).unwrap();
 assert!(m.as_ref(py).downcast::<PyMapping>().is_err());
-PyMapping::register_abc_subclass::<Mapping>(py).unwrap();
+PyMapping::register::<Mapping>(py).unwrap();
 assert!(m.as_ref(py).downcast::<PyMapping>().is_ok());
 ```
 
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -5,9 +5,9 @@ For a detailed list of all changes, see the [CHANGELOG](changelog.md).
 
 ## from 0.16.* to 0.17
 
-### Downcasting (`PyTryFrom`) has been changed for `PyMapping` and `PySequence` types
+### Type checks have been changed for `PyMapping` and `PySequence` types
 
-Previously the safety checks for downcasting to `PyTryFrom` and `PyMapping`
+Previously the type checks for `PyMapping` and `PySequence` (implemented in `PyTryFrom`)
 used the Python C-API functions `PyMapping_Check` and `PySequence_Check`.
 Unfortunately these functions are not sufficient for distinguishing such types,
 leading to inconsistent behavior (see
--- a/src/pyclass.rs
+++ b/src/pyclass.rs
@@ -189,12 +189,7 @@ unsafe fn create_type_object_impl(
         name: py_class_qualified_name(module_name, name)?,
         basicsize: basicsize as c_int,
         itemsize: 0,
-        flags: py_class_flags(
-            has_traverse,
-            is_basetype,
-            #[cfg(all(Py_3_10, not(Py_LIMITED_API)))]
-            is_mapping,
-        ),
+        flags: py_class_flags(has_traverse, is_basetype),
         slots: slots.as_mut_ptr(),
     };
 
@@ -304,11 +299,7 @@ fn py_class_qualified_name(module_name: Option<&str>, class_name: &str) -> PyRes
     .into_raw())
 }
 
-fn py_class_flags(
-    is_gc: bool,
-    is_basetype: bool,
-    #[cfg(all(Py_3_10, not(Py_LIMITED_API)))] is_mapping: bool,
-) -> c_uint {
+fn py_class_flags(is_gc: bool, is_basetype: bool) -> c_uint {
     let mut flags = ffi::Py_TPFLAGS_DEFAULT;
 
     if is_gc {
@@ -317,14 +308,6 @@ fn py_class_flags(
 
     if is_basetype {
         flags |= ffi::Py_TPFLAGS_BASETYPE;
-    }
-
-    #[cfg(all(Py_3_10, not(Py_LIMITED_API)))]
-    {
-        if is_mapping {
-            flags |= ffi::Py_TPFLAGS_MAPPING;
-            flags &= !ffi::Py_TPFLAGS_SEQUENCE;
-        }
     }
 
     // `c_ulong` and `c_uint` have the same size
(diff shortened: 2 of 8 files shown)


src/pyclass.rs
```diff
@@ -309,6 +318,14 @@ fn py_class_flags(is_gc: bool, is_basetype: bool) -> c_uint {
         flags |= ffi::Py_TPFLAGS_BASETYPE;
     }
 
+    #[cfg(all(Py_3_10, not(Py_LIMITED_API)))]
+    {
+        if is_mapping {
+            flags |= ffi::Py_TPFLAGS_MAPPING;
+            flags &= !ffi::Py_TPFLAGS_SEQUENCE;
+        }
+    }
```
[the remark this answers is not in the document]
davidhewitt: Yeah, I thought about doing similar in the past and decided against it. I don't think it makes sense to only in select circumstances set this flag. I would prefer that we let users make this explicit, either by calling the abc `register` like we do elsewhere in this PR or hopefully use #991 to inherit from the abc.

src/types/mapping.rs
```diff
@@ -105,10 +105,34 @@ impl PyMapping {
...
+
+    // for Python >= 3.10 and not using the Py_LIMITED_API, check Py_TPFLAGS_MAPPING to determine
+    // downcastability
+    #[cfg(not(any(Py_LIMITED_API, Py_3_6, Py_3_7, Py_3_8, Py_3_9)))]
     fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v PyMapping, PyDowncastError<'v>> {
         let value = value.into();
         unsafe {
-            if ffi::PyMapping_Check(value.as_ptr()) != 0 {
+            let ty = ffi::Py_TYPE(value);
+            if ffi::PyType_HasFeature(ty, fft::Py_TPFLAGS_MAPPING) != 0 {
```
davidhewitt, line 135: As a sanity check, I wonder if there is any evidence we can collect (or tests we can write) that this check is functionally equivalent to the slow path? If there's any lingering concern that they may be different, we probably need to always use the slow path.

tests/test_mapping.rs
```diff
@@ -116,6 +117,16 @@ fn test_delitem() {
...
+        // downcast to PyMapping requires isinstance(<cls>, collections.abc.Mapping) to pass, so we
+        // have to register the class first
+        PyModule::import(py, "collections.abc")
+            .unwrap()
+            .getattr("Mapping")
+            .unwrap()
+            .getattr("register")
+            .unwrap()
+            .call1((Mapping::type_object(py),))
+            .unwrap();
```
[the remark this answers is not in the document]
davidhewitt: I suppose in absence of #991 users may find it useful to have this exposed as a documented API to improve discoverability?! Not completely sure, what do you think?

davidhewitt (commented):
Thanks for working on this! Got a few thoughts to get us started...

tests/test_mapping.rs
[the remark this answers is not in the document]
aganders3: Yeah I was thinking something like `py.register_abc_subclass<Mapping>(ABC::Mapping).unwrap()` for example. Like you say it might only be useful until #991 is available but easier and more discoverable than the above.

src/types/mapping.rs
aganders3: Good call - I'll give this some thought. I still think this is the right way going forward, but in the mean time (until 3.9 EOL I guess...) the behavior should be the same.

davidhewitt: I think it's not just 3.9 EOL - we also need to be sure that we won't have behaviour differences between abi3 and non-abi3 builds (just perf improvements).

aganders3:
Yeah it's a bit sad but it would be more consistent behavior, even if the discrepancy would be rare. Still I'll defer to your (and others') expertise and experience on this. I don't have a good feel for how big this performance hit will be and/or how common the operation is in real use.

For performance-critical situations could a user still do their own check and use `try_from_unchecked`? It might be sufficient to just document this in the `PyMapping` docs with an explanation that there is no fast perfect check for mapping types at this time.

davidhewitt:
Sorry for the slight delay - had a backlog of things to get through.

I think consistent behaviour is most important, so I think we should go for the `isinstance` check always. As you say, users can use `unsafe` APIs to check the flags / use `try_from_unchecked` if they really need performance and are willing to accept the differences.

src/types/mapping.rs
```diff
@@ -102,17 +108,52 @@ impl PyMapping {
...
+    /// This is required for a class to be downcastable from `PyAny` to `PyMapping`.
+    pub fn register_mapping_abc_subclass<T: PyTypeInfo>(py: Python<'_>) -> PyResult<()> {
+        let ty = T::type_object(py);
+        let mapping_abc = get_mapping_abc(py);
+        mapping_abc.getattr(py, "register")?.call1(py, (ty,))?;
+        Ok(())
+    }
+}
+
+fn get_mapping_abc(py: Python<'_>) -> &Py<PyType> {
```
davidhewitt, line 122: Slight preference to have this return `&'py PyType` - you can do that by adding `.is_ref(py)` after the `get_or_init` call.

src/types/mapping.rs
```diff
@@ -102,17 +108,52 @@ impl PyMapping {
...
+    /// `isinstance(<class>, collections.abc.Mapping) == True`.
     fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v PyMapping, PyDowncastError<'v>> {
         let value = value.into();
-        unsafe {
-            if ffi::PyMapping_Check(value.as_ptr()) != 0 {
-                Ok(<PyMapping as PyTryFrom>::try_from_unchecked(value))
-            } else {
-                Err(PyDowncastError::new(value, "Mapping"))
-            }
+        if is_mapping_abc_subclass(value) {
```
davidhewitt, line 153:
```suggestion
        if value.is_instance(get_mapping_abc(value.py()))? {
```

src/types/mapping.rs
```diff
@@ -102,17 +108,52 @@ impl PyMapping {
                 .from_owned_ptr_or_err(ffi::PyMapping_Items(self.as_ptr()))
         }
     }
+
+    /// Register a pyclass as a subclass of `collections.abc.Mapping` (from the Python standard library).
+    /// This is required for a class to be downcastable from `PyAny` to `PyMapping`.
```
davidhewitt, line 113: Might be nice to add "this is equivalent to `collections.abc.Mapping.register(T)`,  like we have for other docs with python equivalent functionality?

src/types/mapping.rs
```diff
@@ -102,17 +108,52 @@ impl PyMapping {
...
+    pub fn register_mapping_abc_subclass<T: PyTypeInfo>(py: Python<'_>) -> PyResult<()> {
+        let ty = T::type_object(py);
+        let mapping_abc = get_mapping_abc(py);
+        mapping_abc.getattr(py, "register")?.call1(py, (ty,))?;
```
davidhewitt, line 117:
```suggestion
        mapping_abc.call_method1(py, "register", (ty,))?;
```

davidhewitt (commented):
Thanks, this is shaping up well. Posted some implementation suggestions.

I'm beginning to wonder if we should do equivalent treatment for `PySequence`? Seems like it would be good to have both for consistency.

Also I keep going back and forth on whether we should have `#[pyclass(mapping)]` automatically register with the abc. I think it's still the right choice to make users opt-in, however if they don't when upgrading to 0.17 their casts to `PyMapping` will break so we should make a very clear entry about this in the migration guide.

src/types/mapping.rs
aganders3: Thanks - this is a distinction I'm still wrapping my head around a little. I'm also not-quite comfortable with lifetime specifications.

Would this make the fn signature look like this?
```
fn get_mapping_abc<'py>(py: Python<'py>) -> &'py PyType
```
which then can be replaced by placeholders:
```
fn get_mapping_abc(py: Python<'_>) -> &'_ PyType {
```

davidhewitt: Yes that's correct. I'm don't think the placeholder is required on the reference (as there's only one possible lifetime there), so I think you can even have 

```rust
fn get_mapping_abc(py: Python<'_>) -> &PyType {
```

davidhewitt:
Well @aganders3 nobody has raised concerns about also changing sequences, so I guess let's move ahead with that. Do you want to add to this PR, or shall I push separately?

aganders3:
> Well @aganders3 nobody has raised concerns about also changing sequences, so I guess let's move ahead with that. Do you want to add to this PR, or shall I push separately?

I think it makes sense to add it here as there will be overlap at least in the docs/changelog for it. I'm just getting back from vacation but should be able to do that some time this week.

davidhewitt:
That'd be great, thanks! There's still a few items to go in #2493 before we're ready to go, so there's some time to get this in. I'm crossing things off that list when I can! :)

guide/src/migration.md
```diff
@@ -5,6 +5,59 @@ For a detailed list of all changes, see the [CHANGELOG](changelog.md).
...
+subclass of the corresponding abstract base class `collections.abc.Mapping` or
+`collections.abc.Sequence`. Note this requires calling into Python, which may
+incur a performance penalty over the previous method. If this performance
+penatly is a problem, you may be able to perform your own checks and use
+`try_from_unchecked` (unsafe).
+
+Another side-effect is that a pyclass defined in Rust with PyO3 will need to
+be _registered_ with the corresponding Python abstract base class for
+downcasting to succeed. `PySequence::register_abc_subclass` and
+`PyMapping:register_abc_subclass` have been added to make it easy to do this
```
davidhewitt, line 26: I wonder if they should just be named `PySequence::register` and `PyMapping::register`?

src/types/sequence.rs
```diff
@@ -289,15 +302,33 @@ where
     Ok(v)
 }
 
+fn get_sequence_abc(py: Python<'_>) -> &PyType {
+    SEQUENCE_ABC
+        .get_or_init(py, || {
+            py.import("collections.abc")
+                .expect("coud not import 'collections.abc'")
```
davidhewitt, line 309: I wonder if it would be nice to avoid the panics by storing `PyResult<Py<PyType>>` in the static? Either that or we could finish off #2398 .

src/types/sequence.rs
```diff
@@ -289,15 +302,33 @@ where
...
         let value = value.into();
-        unsafe {
-            if ffi::PySequence_Check(value.as_ptr()) != 0 {
-                Ok(<PySequence as PyTryFrom>::try_from_unchecked(value))
-            } else {
-                Err(PyDowncastError::new(value, "Sequence"))
-            }
+        if value
+            .is_instance(get_sequence_abc(value.py()))
+            .unwrap_or(false)
```
davidhewitt, line 327: Is `unwrap_or(false)` correct here? I would have thought it would be better to just return the error.

davidhewitt (commented):
Thanks, this looks really solid. Just a couple of final things to decide on before we merge this in. I've commented on the sequence implementation, the equivalent ideas apply to the mapping implementation.

guide/src/migration.md
aganders3: Yeah I like that - it matches the Python function.

src/types/sequence.rs
aganders3: This would make sense but I wasn't sure how. Do I need to implement `From<err::PyErr>` for `PyDowncastError`?

src/types/sequence.rs
aganders3: This is a good idea. The public API (`register`) already returns a Result so this should be pretty straightforward I think.

src/types/sequence.rs
davidhewitt: Oh I see, we can't express that with the current return type. Maybe just leave a TODO in the code and we can figure that out another time?

aganders3: _That_ I know how to implement haha.

aganders3:
Thanks again for the review. It was trickier than I thought to change the static to hold a `PyResult`, but I'm getting a better feel for result and error types now and hopefully what I did makes sense. Does the `clone_ref` negate the benefit of the `GILOnceCell`? Without that the return was `Result<&PyType, &PyErr>` which was more awkward to use.

I rebased on `main` again so hopefully it's still green. Please take another look when you have a chance!

src/types/mapping.rs
```diff
@@ -102,18 +108,46 @@ impl PyMapping {
...
+        .get_or_init(py, || {
+            Ok(py
+                .import("collections.abc")?
+                .getattr("Mapping")?
+                .downcast::<PyType>()?
+                .into_py(py))
+        })
+        .as_ref()
+        .map(|t| t.as_ref(py))
+        .map_err(|e| e.clone_ref(py))
```
davidhewitt, line 133: The overall approach is definitely correct, you can actually do this bit in one line:

```suggestion
        .map_or_else(|e| e.clone_ref(py), |t| t.as_ref(py))
```

guide/src/migration.md
```diff
@@ -5,6 +5,58 @@ For a detailed list of all changes, see the [CHANGELOG](changelog.md).
 
 ## from 0.16.* to 0.17
 
+### Downcasting (`PyTryFrom`) has been changed for `PyMapping` and `PySequence` types
+
+Previously the safety checks for downcasting to `PyTryFrom` and `PyMapping`
```
davidhewitt, line 10: Looks the the wording in the first line might have meant to be `PyMapping` and `PySequence`? I suggest rephrasing like this:

```suggestion
### Type checks have been changed for `PyMapping` and `PySequence` types

Previously the type checks for `PyMapping` and `PySequence` (implemented in `PyTryFrom`)
```

davidhewitt (commented):
Looking brilliant! Just two final tiny suggestions and then let's merge this. Thanks again for a really solid PR here 👍

src/types/mapping.rs
aganders3: I had to add `Ok` and `Err` because `map_or_else` takes a `Result<T, E> -> U`, but otherwise this worked - thanks!

aganders3:
> Thanks, this looks really solid. Just a couple of final things to decide on before we merge this in. I've commented on the sequence implementation, the equivalent ideas apply to the mapping implementation.

Thanks for all the feedback and review!

davidhewitt (approved):
Looks great, thanks very much again!

Gobot1234:
@aganders3 thank you so much for fixing this!

PR #2478, opened 2022-06-25
opt: remove some generic code bloat

davidhewitt:
Comprised of three commits when just playing around this morning:
- Use `fn` instead of closure in `LazyStaticType::get_or_init` to avoid creating a new monomorphization of `GILOnceCell::get_or_init` for each PyClass
- Use `fn()` instead of `impl FnOnce` in `extract_argument` default / from_py_with arguments, reducing number of monomorphizations of those functions. I think this is more correct anyway (we don't need to support capturing closures).
- Change `#[pyclass]` machinery to create an iterator of `PyClassItems` instead of relying on a "visitor" closure. I think it's tidier, and again reduces the amount of code that needs to be generated per PyClass.

Overall this reduced llvm lines by about 2% in the `pytests` crate.

### DIFF
--- a/src/impl_/extract_argument.rs
+++ b/src/impl_/extract_argument.rs
@@ -38,7 +38,7 @@ where
 pub fn extract_argument_with_default<'py, T>(
     obj: Option<&'py PyAny>,
     arg_name: &str,
-    default: impl FnOnce() -> T,
+    default: fn() -> T,
 ) -> PyResult<T>
 where
     T: FromPyObject<'py>,
@@ -57,7 +57,7 @@ where
 pub fn from_py_with<'py, T>(
     obj: &'py PyAny,
     arg_name: &str,
-    extractor: impl FnOnce(&'py PyAny) -> PyResult<T>,
+    extractor: fn(&'py PyAny) -> PyResult<T>,
 ) -> PyResult<T> {
     // Safety: obj is not None (see safety
     match extractor(obj) {
@@ -71,8 +71,8 @@ pub fn from_py_with<'py, T>(
 pub fn from_py_with_with_default<'py, T>(
     obj: Option<&'py PyAny>,
     arg_name: &str,
-    extractor: impl FnOnce(&'py PyAny) -> PyResult<T>,
-    default: impl FnOnce() -> T,
+    extractor: fn(&'py PyAny) -> PyResult<T>,
+    default: fn() -> T,
 ) -> PyResult<T> {
     match obj {
         Some(obj) => from_py_with(obj, arg_name, extractor),
--- a/src/impl_/pyclass.rs
+++ b/src/impl_/pyclass.rs
@@ -177,7 +177,7 @@ pub trait PyClassImpl: Sized {
     #[cfg(feature = "multiple-pymethods")]
     type Inventory: PyClassInventory;
 
-    fn for_all_items(visitor: &mut dyn FnMut(&PyClassItems));
+    fn items_iter() -> PyClassItemsIter;
 
     #[inline]
     fn dict_offset() -> Option<ffi::Py_ssize_t> {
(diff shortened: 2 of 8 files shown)


src/pyclass.rs
```diff
@@ -1,11 +1,11 @@
 //! `PyClass` and related traits.
+use crate::impl_::pyclass::PyClassItemsIter;
```
adamreichold, line 2: Was this moved to a separate use clause intentionally?

src/impl_/pyclass.rs
```diff
@@ -189,6 +189,204 @@ pub trait PyClassImpl: Sized {
...
+            9 => {
+                self.idx += 1;
+                Some(self.sequence_protocol_items)
+            }
+            #[cfg(feature = "pyproto")]
+            10 => {
+                self.idx += 1;
+                Some(self.buffer_protocol_items)
+            }
+            _ => None,
```
adamreichold, line 329: I find it tricky that this case handles `2..=10` or not depending on the `pyproto` feature. I cannot suggest a nicer way, but maybe a comment would be useful?

src/impl_/pyclass.rs
```diff
@@ -189,6 +189,204 @@ pub trait PyClassImpl: Sized {
     }
 }
 
+/// Iterator used to process all class items during type instantiation.
+pub struct PyClassItemsIter {
```
adamreichold, line 193: Would it make sense to implement `size_hint` and/or `ExactSizeIterator` or do we rely on inlining and code expansion for performance anyway?

src/impl_/extract_argument.rs
```diff
@@ -38,7 +38,7 @@ where
 pub fn extract_argument_with_default<'py, T>(
     obj: Option<&'py PyAny>,
     arg_name: &str,
-    default: impl FnOnce() -> T,
+    default: fn() -> T,
```
adamreichold, line 41: This implies that argument extraction includes dynamic dispatch? Are we confident on the performance implications of this? Or does it actually improve performance by improving code density?

src/pyclass.rs
davidhewitt: No, IDE auto import, will move into the clause. 👍

src/impl_/pyclass.rs
davidhewitt: Yes, I agree that this is a bit horrible (esp. with the second lot of duplication for `multiple-pymethods`). In the end I decided not to bother refactoring further as (hopefully) 0.17 series will be the last set of releases with `pyproto`.

src/impl_/pyclass.rs
davidhewitt: I suspect that in this case it might be inlined anyway as the only code which is expected to be iterating this type is the `create_type_object` function inside of PyO3.

I don't think size_hint / `ExactSizeIterator` is likely to be relevant as that's usually (afaik) for deciding how much space to allocate in a collection as part of a call to `.collect()`? Also in `multiple-pymethods` case I'm not sure whether we can support it cleanly (doesn't look like `inventory::iter` struct implements these hints).

src/impl_/extract_argument.rs
davidhewitt: Great question, yes I was also concerned about this and thanks for encouraging me to write down my conclusions!

The primary motivation for this is that it reduces the quantity of llvm code for these methods twofold:
- fewer monomorphizations by removing the generic over `default` (note still generic over `T`)
- I saw the generated code itself is smaller, I would guess because calling `fn() -> T` is simpler than going through the `FnOnce` trait call shim

I believe that LLVM is still happy to inline through `fn()` calls where it knows the callsite - e.g. see https://godbolt.org/z/rYxToExc6. It looks like in that snippet that `example::main` has inlined the call to `do_thing ` and additionally the "dynamic" call to `get_x` so that ultimately it just pushes `5` straight onto the register without doing any function calls at all.

So the conclusion I've reached is that because these functions are all generic (thus code available for inlining) there's a good chance that LLVM is going to inline away the call to `default` completely (and maybe even the call to `extract_argument_with_default`).

Hopefully the outcome of this change is that we're passing a smaller blob of stuff to LLVM, saving on compile times / giving it less to optimize away, while still ending up with an equivalent result.

I looked at our existing benchmarks and I couldn't see any meaningful differences in performance. I could write a pathological case which exercises a lot of defaults, if you'd like to be extra sure.

src/impl_/pyclass.rs
davidhewitt: Perhaps worth pointing out I'm not overly concerned about maximising performance here as this will be part of one-time setup for pyclasses on module import and never executed again after that. As long as we're not doing anything terrible, perf isn't likely to be noticeable here.

src/impl_/pyclass.rs
davidhewitt: Pushed a comment as suggested.

src/impl_/extract_argument.rs
adamreichold: Thank you for the thorough write up. So we get equivalent performance at shorter compile times. Nice!

adamreichold (approved):

src/impl_/extract_argument.rs
davidhewitt: That's my understanding. If we do get users reporting performance concerns we can always revisit in the future. I think the thing to optimize next for performance is still the thread local used in `GILPool` - i.e. I really hope to look at #1308 again, probably in 0.18 when pyproto complexity is gone.

PR #2484, opened 2022-06-27
pymethods: add support for `__del__`

### LINKED ISSUE #2479: Support `__del__`
I'd like to be able to use a `__del__` method in `#[pymethods]`, however it's not being called.

There are two other related issue #402 and #774. #774 seems most relevant but the OP get's told they shouldn't be using `__del__`. AFAIK issuing a `ResourceWarning` is a legitimate use of `__del__`.

I could print a warning in `Drop` but that won't be a python warning and will confuse people, without support for `__del__` I don't see another way around this.

Related PR https://github.com/samuelcolvin/watchfiles/pull/164

messense:
Adds support for `__del__` in `#[pymethods]`, mapped to CPython's `tp_finalize` slot.

```rust
#[pymethods]
impl MyClass {
    fn __del__(&mut self) {
        println!("instance is being destroyed");
    }
}
```

Closes #2479

## Semantics

`__del__` must be an instance method taking no arguments besides an optional `Python<'_>` (enforced at compile time, with UI tests for the diagnostics). PyO3's `tp_dealloc` calls `PyObject_CallFinalizerFromDealloc` before any teardown, mirroring CPython's `subtype_dealloc`:

- the finalizer runs at most once per instance for GC types (tracked by the GC "finalized" bit) and may resurrect the object, in which case deallocation is aborted;
- for GC types it runs while the object is still GC-tracked, and before weakrefs and the instance dict are cleared;
- exceptions raised in `__del__` (including panics) are reported as unraisable, and any exception already set when the finalizer runs is saved and restored around the call, matching CPython's `slot_tp_finalize`.

## Platform limitations

- **abi3 targeting Python < 3.15:** `__del__` is a compile error. `PyObject_CallFinalizerFromDealloc` is only part of the limited API from 3.15, and the finalize-from-dealloc protocol (finalized bit, resurrection refcount handling) cannot be soundly reimplemented from older stable-ABI symbols. A const assertion emitted next to each `__del__` turns this into a clear diagnostic instead of a linker/name-resolution error.
- **PyPy / GraalPy:** the finalizer call in PyO3's dealloc is a no-op, so `__del__` timing diverges from CPython and may not fire on the dealloc path. This is documented in the guide.

Assisted with Claude Code Fable 5.


tests/test_proto_methods.rs
```diff
@@ -164,6 +168,16 @@ fn test_bool() {
     })
 }
 
+#[test]
+fn test_del() {
+    Python::with_gil(|py| {
+        let example_py = make_example(py);
+        example_py.setattr("special_custom_attr", 15).unwrap();
+        py_run!(py, example_py, "example_py.__del__()");
```
messense, line 176: Not sure what's the best way to test it.

> It looks like the tp_dealloc in that file also calls tp_finalize in some circumstances; we might need to check carefully that we reproduce the Python behaviour correctly.

@davidhewitt Could you elaborate on how do we check that?

davidhewitt (commented):
Thanks! Please also add a CHANGELOG and add to `protocols.md` in the guide.

Note that also on Python 3.7 we need to set [`Py_TPFLAGS_HAVE_FINALIZE`](https://docs.python.org/3/c-api/typeobj.html#Py_TPFLAGS_HAVE_FINALIZE), it's ignored on newer versions.

pytests/tests/test_pyclasses.py
```diff
@@ -25,3 +27,15 @@ def test_iter():
...
+def test_del():
+    d = pyclasses.PyClassDel()
+    del d
+
+
+@mock.patch("sys.unraisablehook")
+def test_del_error(unraisablehook):
+    d = pyclasses.PyClassDelError()
+    del d
+    assert unraisablehook.called
```
messense, line 41: This test case does not pass locally, adding a `panic!()` to `__del__` seems to indicate that `__del__` is not called as it did not trigger a panic.

davidhewitt: As per https://github.com/PyO3/pyo3/pull/2484#discussion_r907034479 this indicates to me that we may want to consider adding a block similar to the following to `PyCell::tp_dealloc`

https://github.com/python/cpython/blob/56310136170ef6653cb050f58dd2d32538997f59/Modules/_asynciomodule.c#L1534-L1542

But `PyObject_CallFinalizerFromDealloc` is only in the limited API, so it might be that we can only support `__del__` when not running `abi3`.

messense: That's unfortunate.

> ...or maybe we should be removing our custom tp_dealloc and moving all logic to tp_finalize ??

Could we do this instead so it does not require calling `PyObject_CallFinalizerFromDealloc` in extension module?

davidhewitt: Yes, that seems reasonable to me. I have no idea whether it will definitely work. I plan to play around with this once the 0.17 release is complete.

davidhewitt:
I found today that if we don't implement `tp_dealloc` then the default `tp_dealloc` implementation (at least for CPython) will call `tp_finalize`. This gives us a potentially route to support this and also fix #3064 at the same time.

However the documentation in this area is confusing, as it implies that `tp_dealloc` doesn't have a default - https://github.com/python/cpython/issues/105018 / https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_dealloc

I think it may be possible that we use that default implementation to support this case, but we'll need to test carefully whether this also works for all Python versions, PyPy etc.

davidhewitt:
I spoke to vstinner (thought no need to ping so no @) briefly about this, I understood that `PyObject_CallFinalizerFromDealloc` could be added to the limited API in 3.13. So maybe we can figure out a way to proceed with this where `__del__` is supported without `abi3` for now and then we can add `abi3` support on 3.13.

messense:
With the help from LLM, made some progress on this, tests are passing, review would be appreciated!

I have also looked into removing our `tp_dealloc` implementation but turns out to be pretty hard to get everything right.

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -510,6 +514,52 @@ fn impl_clear_slot(cls: &syn::Type, spec: &FnSpec<'_>, ctx: &Ctx) -> syn::Result
     })
 }
 
+fn impl_del_slot(cls: &syn::Type, spec: &FnSpec<'_>, ctx: &Ctx) -> syn::Result<MethodAndSlotDef> {
```
davidhewitt: Does this need a special codegen? I wonder if it can use a `SlotDef` to make the implementation less verbose.

src/impl_/trampoline.rs
```diff
@@ -224,6 +224,76 @@ pub mod releasebufferproc {
...
+
+    #[inline]
+    pub(crate) unsafe fn inner(slf: *mut ffi::PyObject, f: Func) {
+        // Dedicated trampoline for tp_finalize (__del__), matching CPython's
+        // slot_tp_finalize semantics: save the current exception, call __del__,
+        // write any error as unraisable, then restore the original exception.
+        //
+        // We cannot use trampoline_unraisable here because the exception
+        // save/restore must happen *around* the catch_unwind boundary — if __del__
+        // panics, we still need to restore the saved exception state.
```
davidhewitt: I think if we had a `struct ExceptionGuard` which called `PyErr::take` on construction and `PyErr::restore` on `Drop`, we could probably use `trampoline_unraisable`.

src/impl_/pyclass.rs
```diff
@@ -1120,6 +1120,14 @@ pub trait PyClassBaseType: Sized {
...
 /// Implementation of tp_dealloc for pyclasses without gc
 pub(crate) unsafe extern "C" fn tp_dealloc<T: PyClass>(obj: *mut ffi::PyObject) {
+    // Call tp_finalize if defined, matching CPython's subtype_dealloc behavior.
+    // PyO3 replaces the default subtype_dealloc with its own tp_dealloc, so we
+    // must explicitly invoke tp_finalize here.
+    #[cfg(all(not(Py_LIMITED_API), not(GraalPy)))]
+    if unsafe { ffi::PyObject_CallFinalizerFromDealloc(obj) } < 0 {
+        // Object was resurrected by the finalizer; abort deallocation.
+        return;
+    }
```
davidhewitt, line 1212: I think, for abi3 builds, for non-gc types it's impossible for the GC to have marked the object as finalised, so it should be always valid to call `tp_finalize` here (if needed).

src/impl_/pyclass.rs
```diff
@@ -1129,6 +1137,22 @@ pub(crate) unsafe extern "C" fn tp_dealloc_with_gc<T: PyClass>(obj: *mut ffi::Py
...
+    // then un-tracks it afterwards. We mirror that behavior here.
+    #[cfg(all(not(Py_LIMITED_API), not(PyPy), not(GraalPy)))]
+    unsafe {
+        ffi::PyObject_GC_Track(obj.cast());
+    }
+    #[cfg(all(not(Py_LIMITED_API), not(GraalPy)))]
+    if unsafe { ffi::PyObject_CallFinalizerFromDealloc(obj) } < 0 {
+        // Object was resurrected by the finalizer and is GC-tracked; abort deallocation.
+        return;
+    }
```
davidhewitt: I think on Python 3.9 and up we can reproduce this for abi3 by using `PyObject_GC_IsFinalized`.

So I think we might be able to support `__del__` fully on abi3 for Python 3.9+, and add a check similar to `dict` / `weakref` to forbid implementing `__del__` on abi3 3.7 / 3.8 (which we'll drop support for soon anyway).

tests/test_proto_methods.rs
```diff
@@ -893,3 +893,302 @@ fn test_contains_opt_out() {
...
     })
 }
+
+// __del__ tests
+
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
+#[pyclass(subclass)]
+struct ClassWithDel {
```
davidhewitt: Can probably merge this with `ClassWithDelPy`, it might also be interesting to add a test in `test_inheritance` which tests this behavior with subclassing. (Similarly, `#[pyclass(subclass)]` flag probably not needed here.)

tests/test_proto_methods.rs
```diff
@@ -893,3 +893,302 @@ fn test_contains_opt_out() {
...
+#[test]
+fn test_del_called_explicitly() {
+    Python::attach(|py| {
+        let flag = Arc::new(AtomicBool::new(false));
+        let obj = Bound::new(py, ClassWithDel { flag: flag.clone() }).unwrap();
+        assert!(!flag.load(Ordering::SeqCst));
+        obj.call_method0("__del__").unwrap();
+        assert!(flag.load(Ordering::SeqCst));
+    })
+}
```
davidhewitt: This looks duplicated with the test above.

davidhewitt (commented):
Thanks for moving this forward! There are a few typical LLM warts like duplicate tests, I think we might also be able to simplify a bit.

I'd like the behavior for abi3 to be well defined for users. The two options I see:
- Don't allow `__del__` on abi3 for now.
- Allow `__del__` on abi3 with Python 3.9 (see my comments with ideas how to support this below)

I also opened https://github.com/python/cpython/issues/146063 to ask upstream to add the needed methods to the stable ABI (probably I should have done this years ago).

davidhewitt:
> I have also looked into removing our `tp_dealloc` implementation but turns out to be pretty hard to get everything right.

I think it's probably necessary for us to continue to implement `tp_dealloc`, I don't think there's an alternative way for us to drop the Rust data within types at the point of destruction.

src/internal/get_slot.rs
```diff
@@ -133,6 +133,32 @@ impl_slots! {
     TP_TRAVERSE: (Py_tp_traverse, tp_traverse) -> Option<ffi::traverseproc>,
 }
 
+#[cfg(all(Py_LIMITED_API, not(GraalPy)))]
+pub(crate) const TP_FINALIZE: Slot<{ ffi::Py_tp_finalize }> = Slot;
```
davidhewitt: Can this go in `impl_slots!` above as `TP_FINALIZE: (Py_ty_finalize, tp_finalize) -> Option<ffi::destructor>`?

davidhewitt (commented):
Thanks, sorry for the slow review. I think this is pretty much ready, one small cleanup proposed.

src/impl_/pyclass.rs
```diff
@@ -1117,8 +1120,82 @@ pub trait PyClassBaseType: Sized {
     type Layout<T: PyClassImpl>;
 }
 
+/// Limited-API replacement for `PyObject_CallFinalizerFromDealloc`.
```
davidhewitt: As of https://github.com/python/cpython/issues/146063 it should be the case that 3.15.0b1 now contains `PyObject_CallFinalizerFromDealloc`.

On that issue I believe it was established that we can't fully mimic the behavior of that API (mostly deduplication), so I think it might be better if we instead only allow this method with abi3 on 3.15+

We can probably do this by inserting an assertion in the generated code, similar to `#[pyclass(dict)]` (which IIRC requires 3.9 on abi3).

davidhewitt (changes requested):
Actually, with apologies, I think we should probably roll back the limited-api implementation and instead only offer abi3 support on 3.15+ (which now contains `PyObject_CallFinalizerFromDealloc`).

pyo3-ffi/src/object.rs
```diff
@@ -538,6 +538,12 @@ extern_libpython! {
     pub fn PyCallable_Check(arg1: *mut PyObject) -> c_int;
     #[cfg_attr(PyPy, link_name = "PyPyObject_ClearWeakRefs")]
     pub fn PyObject_ClearWeakRefs(arg1: *mut PyObject);
+    // Added to the limited API in 3.15. PyO3 avoids calling this on PyPy
+    // because its cpyext implementation is not compatible with CPython
+    // deallocation semantics.
```
davidhewitt, line 543:
```suggestion
```

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -1513,10 +1586,31 @@ impl SlotDef {
...
+            #pyo3_path::impl_::trampoline::get_trampoline_function!(#trampoline_ty, #cls::#wrapper_ident)
+                as #pyo3_path::ffi::#ffi_cast_ty as _
+        };
+        // On abi3 before Python 3.15, `__del__` is unsupported because
+        // `PyObject_CallFinalizerFromDealloc` is not part of the limited API.
+        // Surface that as a clear diagnostic pointed at the user's `__del__`
+        // rather than an opaque "cannot find `finalizefunc`" error. The check is
+        // embedded in the slot's `pfunc` so it is const-evaluated when the
+        // type's slot table is built (and is a no-op where `__del__` is supported).
+        let pfunc = if *require_del_supported {
```
davidhewitt, line 1599: Rather than having a `require_del_supported` boolean state kicking around, can we just check if the slot is `Py_tp_finalize` here?

tests/test_inheritance.rs
```diff
@@ -539,3 +539,90 @@ fn test_subclass2_ref_counts() {
 fn test_simpleclass_ref_counts() {
     assert_type_refcount_stable!(SimpleClass);
 }
+
+// `__del__` requires `PyObject_CallFinalizerFromDealloc`, which is only part
+// of the limited API from Python 3.15.
+#[cfg(any(not(Py_LIMITED_API), Py_3_15))]
+mod inheritance_with_del {
```
davidhewitt, line 546: `tp_finalize` is automatically inherited by subtypes, perhaps we can just delete these tests?

tests/test_proto_methods.rs
```diff
@@ -1008,3 +1008,327 @@ fn test_contains_opt_out() {
...
+    impl ClassWithDelPy {
+        fn __del__(&self, _py: Python<'_>) {
+            self.flag.store(true, Ordering::SeqCst);
+        }
+    }
+
+    /// __del__ can be called explicitly via Python, and can optionally accept a
+    /// Python<'_> argument (with either receiver form).
+    #[test]
+    fn test_del_called_explicitly() {
```
davidhewitt, line 1056: If called explicitly, does it run again on dealloc?

tests/test_gc.rs
davidhewitt, line 1: The changes in this file seem unrelated to `__del__`?

guide/src/class/protocols.md
```diff
@@ -160,6 +159,23 @@ The given signatures should be interpreted as follows:
...
+
+  Note that `__del__` is called when the object is finalized, which is not the point at which the Rust struct is dropped (finalization happens earlier, and may not lead to destruction at all if the object is resurrected).
+  To run Rust code on destruction, implement [`Drop`](https://doc.rust-lang.org/std/ops/trait.Drop.html) instead; implement `__del__` only when finalization needs to run Python code or observe the Python object.
+
+  > `__del__` requires Python 3.15 or later when building with the `abi3` feature, because `PyObject_CallFinalizerFromDealloc` was only added to the limited API in Python 3.15.
+  >
+  > On PyPy, PyO3 does not call `PyObject_CallFinalizerFromDealloc` from `tp_dealloc` because PyPy's cpyext implementation is not CPython-compatible, but PyPy may still invoke `tp_finalize` from its own GC/finalization path.
+  > On GraalPy, PyO3 also skips dealloc-time finalization because the API has been observed to crash there.
+  > On these runtimes, finalization timing may differ from CPython, and `__del__` will still run when called explicitly.
```
davidhewitt, line 178: Do you know what the crashes / issues were with PyPy & GraalPy were? If we can reduce to clean minimal repros maybe we can report bugs upstream and get consistent behaviour on those too.

davidhewitt (approved):
Thanks, this looks pretty much ready to merge!

Just some minor suggestions / questions remaining. It would be nice if we can figure out what PyPy / GraalPy issues are before merging, in case upstream can fix them.

PR #2485, opened 2022-06-28
types: rework PyCapsule for soundness

davidhewitt:
Follow up to #2481.

This PR changes the `PyCapsule` API as discussed in that PR. In addition, I realised that error handling in the implementation was potentially incorrect - I had to rework a few things so that in case of an "invalid capsule" Python errors are always either propagated or cleared as I felt was appropriate.

### DIFF
--- a/src/types/capsule.rs
+++ b/src/types/capsule.rs
@@ -2,8 +2,8 @@
 use crate::Python;
 use crate::{ffi, AsPyPointer, PyAny};
 use crate::{pyobject_native_type_core, PyErr, PyResult};
-use std::ffi::{c_void, CStr, CString};
-use std::os::raw::c_int;
+use std::ffi::{CStr, CString};
+use std::os::raw::{c_char, c_int, c_void};
 
 /// Represents a Python Capsule
 /// as described in [Capsules](https://docs.python.org/3/c-api/capsule.html#capsules):
@@ -28,7 +28,7 @@ use std::os::raw::c_int;
 ///      let foo = Foo { val: 123 };
 ///      let name = CString::new("builtins.capsule").unwrap();
 ///
-///      let capsule = PyCapsule::new(py, foo, name.as_ref())?;
+///      let capsule = PyCapsule::new(py, foo, Some(name.clone()))?;
 ///
 ///      let module = PyModule::import(py, "builtins")?;
 ///      module.add("capsule", capsule)?;
@@ -60,7 +60,7 @@ impl PyCapsule {
     ///
     /// Python::with_gil(|py| {
     ///     let name = CString::new("foo").unwrap();
-    ///     let capsule = PyCapsule::new(py, 123_u32, &name).unwrap();
+    ///     let capsule = PyCapsule::new(py, 123_u32, Some(name)).unwrap();
     ///     let val = unsafe { capsule.reference::<u32>() };
     ///     assert_eq!(*val, 123);
     /// });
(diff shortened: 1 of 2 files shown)


src/types/capsule.rs
```diff
@@ -164,77 +167,113 @@ impl PyCapsule {
...
             Ok(())
         }
     }
 
     /// Gets the current context stored in the capsule. If there is no context, the pointer
     /// will be null.
-    pub fn get_context(&self, py: Python<'_>) -> PyResult<*mut c_void> {
+    ///
+    /// Returns an error if this capsule is not valid.
+    pub fn get_context(&self) -> PyResult<*mut c_void> {
```
adamreichold, line 194: Since we are making a breaking change in any case, shouldn't this be named `context` following general Rust conventions as well as `name` and `pointer`?

adamreichold (approved):
I think the changes are correct. I am not sure whether our API matches the intent of the CPython API it is wrapping though.

src/types/capsule.rs
davidhewitt: Pushed 👍

src/types/capsule.rs
```diff
@@ -164,77 +167,113 @@ impl PyCapsule {
...
     /// Checks if this is a valid capsule.
+    ///
+    /// Returns true if the stored `pointer()` is non-null.
     pub fn is_valid(&self) -> bool {
-        let r = unsafe { ffi::PyCapsule_IsValid(self.as_ptr(), self.name().as_ptr()) } as u8;
+        // As well as if the stored pointer is null, PyCapsule_IsValid also returns false if
+        // self.as_ptr() is null or not a ptr to a PyCapsule object. Both of these are guaranteed
+        // to not be the case thanks to invariants of this PyCapsule struct.
+        let r =
+            unsafe { ffi::PyCapsule_IsValid(self.as_ptr(), self.name_ptr_ignore_error()) } as u8;
```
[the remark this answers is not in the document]
davidhewitt: > (Why are the `i32` return values here and in a few other places cast to `u8` before comparing them to zero?)

Good question, seems pointless, removed.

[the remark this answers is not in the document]
davidhewitt: I've now pushed a change to the above effect.

src/types/capsule.rs
```diff
@@ -89,36 +89,39 @@ impl PyCapsule {
...
-        // Take ownership of a copy of `name` so that the string is guaranteed to live as long
-        // as the capsule. PyCapsule_new purely saves the pointer in the capsule so doesn't
-        // guarantee ownership itself.
-        let name = name.to_owned();
-        let name_ptr = name.as_ptr();
+
+        // Sanity check for capsule layout
+        debug_assert_eq!(memoffset::offset_of!(CapsuleContents::<T, F>, value), 0);
+
+        let name_ptr = name.as_ref().map_or(std::ptr::null(), |name| name.as_ptr());
```
adamreichold, line 110: I think this could also use `name_as_ptr`, probably via `name_as_ptr(name.as_deref())` (which should be in MSRV reach at 1.40)?

adamreichold (commented):
> Would you also change reference() to take name: Option<&CStr> argument, or given that method is unsafe would you leave it with the existing semantics it has here (where issues with validity / name are ignored)?

This is a difficult decision. I think going for the principle of least surprise, I would expect `reference` to call `pointer` and just take care of casting and dereferencing, so I would expect an equivalent signature like

```rust
pub unsafe fn reference<T>(&self, name: Option<&CStr>) -> PyResult<&T>;
```

I am also not sure if the currently implementation shouldn't at least clear any Python exceptions even if it ignores them?

But as written before, this under the assumption that we should match the CPython API where I do not understand _why_ the CPython API wants to independently verify the name. (Could this be a consistency check to prevent type confusion?)

davidhewitt:
After thinking about this for a while, I noticed that `pybind11` also deviates from the C API and automatically loads the name, as we were, so I think I will roll back a couple of these changes to be closer to our previous API. 

(https://github.com/pybind/pybind11/blob/75007dda72ad4508064c1f080394eae50d3a61ee/include/pybind11/pytypes.h#L1626)

adamreichold (approved):

Issue #2049, opened 2021-12-13
Bindings for `super()`


davidhewitt:
I don't think PyO3 currently exposes an API to get `super()` from Rust.

As far as I'm aware there aren't special C APIs to call `super()`, instead I think `ffi::PySuper_Type` can be cast to `&PyType` and we can then use `.call1((type, obj))` to fill in the arguments. I think we'd always need to use the two-argument form as I don't think the PEP 3135 syntax sugar can apply here.

Some ideas what could be done:
 - Add a `PySuper` type to `pyo3::types`, with constructors like `PySuper::new(ty: &PyType, obj: &PyAny) -> PyResult<PySuper>` 
 - Maybe add `PyAny::super()` which calls `PySuper::new` with the object and its type for convenience.

PR #2486, opened 2022-06-28
Add super object

ikrivosheev:
Hello! this PR close: https://github.com/PyO3/pyo3/issues/2049


tests/test_super.rs
```diff
@@ -0,0 +1,52 @@
...
+impl SubClass {
+    #[new]
+    fn new() -> (Self, BaseClass) {
+        (SubClass {}, BaseClass::new())
+    }
+
+    fn method2<'a>(self_: PyRef<'_, Self>, py: Python<'a>) -> PyResult<&'a PyAny> {
+        let any: Py<PyAny> = self_.into_py(py);
+        let super_ = any.into_ref(py).py_super()?;
+        super_.call_method("method1", (), None)
```
davidhewitt, line 35: I think it's more interesting if in the test both methods are just called `method`, right? Then it's proving that it's correctly calling the base object's `method` and not infinitely recursing.

tests/test_super.rs
```diff
@@ -0,0 +1,52 @@
...
+#[pymethods]
+impl SubClass {
+    #[new]
+    fn new() -> (Self, BaseClass) {
+        (SubClass {}, BaseClass::new())
+    }
+
+    fn method2<'a>(self_: PyRef<'_, Self>, py: Python<'a>) -> PyResult<&'a PyAny> {
+        let any: Py<PyAny> = self_.into_py(py);
+        let super_ = any.into_ref(py).py_super()?;
```
davidhewitt, line 34: Think this can just be written as
```suggestion
        let super_ = self_.py_super()?;
```

src/types/pysuper.rs
```diff
@@ -0,0 +1,26 @@
...
+#[repr(transparent)]
+pub struct PySuper(PyAny);
+
+pyobject_native_type_core!(PySuper, ffi::PySuper_Type, #checkfunction=ffi::PyType_Check);
+
+impl PySuper {
+    pub fn new<'py>(py: Python<'py>, ty: &'py PyType, obj: &'py PyAny) -> PyResult<&'py PySuper> {
+        let args = PyTuple::new(py, &[ty, obj]);
+        let type_ = PySuper::type_object_raw(py);
+        let super_ = unsafe { ffi::PyObject_CallObject(type_ as *mut _, args.as_ptr()) };
```
davidhewitt, line 18: Should we go through equivalent safe APIs? `PySuper::type_object(py).call1((ty, obj))` ?

src/types/pysuper.rs
```diff
@@ -0,0 +1,26 @@
...
+/// Represents a Python `super` object.
+///
+/// This type is immutable.
+#[repr(transparent)]
+pub struct PySuper(PyAny);
+
+pyobject_native_type_core!(PySuper, ffi::PySuper_Type, #checkfunction=ffi::PyType_Check);
+
+impl PySuper {
+    pub fn new<'py>(py: Python<'py>, ty: &'py PyType, obj: &'py PyAny) -> PyResult<&'py PySuper> {
```
davidhewitt, line 15: It would be nice to point users to the `super()` docs in Python so that it becomes clear what combination of arguments are valid.

src/types/pysuper.rs
```diff
@@ -0,0 +1,26 @@
...
+use crate::types::{PyTuple, PyType};
+use crate::{AsPyPointer, Py, PyAny, PyErr, PyResult, Python};
+
+/// Represents a Python `super` object.
+///
+/// This type is immutable.
+#[repr(transparent)]
+pub struct PySuper(PyAny);
+
+pyobject_native_type_core!(PySuper, ffi::PySuper_Type, #checkfunction=ffi::PyType_Check);
```
davidhewitt, line 12: I'm not convinced that `PyType_Check` is correct, that will check if objects are of type `type` ?

I think better to remove this and use the default implementation.

```suggestion
pyobject_native_type_core!(PySuper, ffi::PySuper_Type);
```

We could verify that this works correctly by writing a test which calls `PySuper::is_type_of(super_obj)` for some `super_obj: PySuper`.

davidhewitt (commented):
Many thanks for working on this! I think there's a few bits which need to be tidied up, see comments below.

tests/test_super.rs
ikrivosheev: Done

tests/test_super.rs
ikrivosheev: @davidhewitt I correctly understood that I need to add new method to `PyRef`?

src/types/pysuper.rs
ikrivosheev: Thank you! Yes, done!

src/types/pysuper.rs
ikrivosheev: Done! I add link to python doc and small example

src/types/pysuper.rs
ikrivosheev: Thank you! Done!

ikrivosheev:
> Many thanks for working on this! I think there's a few bits which need to be tidied up, see comments below.

@davidhewitt thank you for review! I fix some comments but I have one question: https://github.com/PyO3/pyo3/pull/2486#discussion_r908973299

tests/test_super.rs
davidhewitt: Ah right, yes `PyRef` doesn't deref to `PyAny`. `PyCell<T>` does, so I think let's instead go with:

```rust
    fn method<'py>(self_: &'py PyCell<Self>) -> PyResult<&'py PyAny> {
         let super_ = self_.py_super()?;
```

src/types/pysuper.rs
```diff
@@ -0,0 +1,61 @@
...
+    ///     }
+    ///
+    ///     fn method<'a>(self_: PyRef<'_, Self>, py: Python<'a>) -> PyResult<&'a PyAny> {
+    ///         let any: Py<PyAny> = self_.into_py(py);
+    ///         let super_ = any.into_ref(py).py_super()?;
+    ///         super_.call_method("method", (), None)
+    ///     }
+    /// }
+    /// ```
+    pub fn new<'py>(py: Python<'py>, ty: &'py PyType, obj: &'py PyAny) -> PyResult<&'py PySuper> {
```
davidhewitt, line 56: Given that this takes gil-bound types `&'py PyType` and `&'py PyAny`, I think we can skip taking `py: Python<'py>`:

```suggestion
    pub fn new<'py>(ty: &'py PyType, obj: &'py PyAny) -> PyResult<&'py PySuper> {
        let py = ty.py();
```

davidhewitt (commented):
Thanks for the amendments - I've given this another read and got a few final suggestions.

I think you will also need to add `#[cfg(not(PyPy))]` to this file - it looks like `PyPy` does not have the C APIs needed for this.

src/types/pysuper.rs
ikrivosheev: Thank you! Done

tests/test_super.rs
ikrivosheev: Done!

ikrivosheev:
@davidhewitt thank you for review! I have done with comment)

davidhewitt (approved):
Many thanks, this looks great to me now!

PR #2487, opened 2022-06-28
Make clippy happy with code emitted by impl_arg_param 

mejrs:
### Fixes some bad error message when Rust can't find a type:

```rust
// missing PathBuf import
use pyo3::prelude::*;

#[pyclass]
pub struct S;

#[pymethods]
impl S {
    #[new]
    #[args(path = "None")]
    fn new(path: Option<PathBuf>) -> PyResult<Self> {
        todo!()
    }
}
```

```
-- snip --

error[E0282]: type annotations needed for `Option<T>`
 --> src\lib.rs:7:1
  |
7 | #[pymethods]
  | ^^^^^^^^^^^^ consider giving `arg0` the explicit type `Option<T>`, where the type parameter `T` is specified
  |
  = note: this error originates in the attribute macro `pymethods` (in Nightly builds, run with -Z macro-backtrace for more info)

Some errors have detailed explanations: E0282, E0412.
For more information about an error, try `rustc --explain E0282`.
```

### Fixes https://github.com/rust-lang/rust-clippy/issues/8971

```rust
#[pyo3::pyfunction]
pub fn hash_djb2(_s: &str) -> i32 {
    0
}
```
```
warning: deref on an immutable reference
 --> src\lib.rs:3:22
  |
3 | pub fn hash_djb2(_s: &str) -> i32 {
  |                      ^^^^ help: if you would like to reborrow, try removing `&*`: `&str`
  |
  = note: `#[warn(clippy::borrow_deref_ref)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#borrow_deref_ref

warning: `my_module` (lib) generated 1 warning
```

### DIFF
--- a/pyo3-macros-backend/src/params.rs
+++ b/pyo3-macros-backend/src/params.rs
@@ -240,7 +240,6 @@ fn impl_arg_param(
                 }
             }
             (Some(default), _) => {
-                
                 quote_arg_span! {
                     _pyo3::impl_::extract_argument::from_py_with_with_default(#arg_value, #name_str, #expr_path, || #default)
                 }
@@ -263,7 +262,6 @@ fn impl_arg_param(
     } else {
         match (spec.default_value(name), arg.optional.is_some()) {
             (Some(default), true) if default.to_string() != "None" => {
-
                 quote_arg_span! {
                     _pyo3::impl_::extract_argument::extract_argument_with_default(#arg_value, #name_str, || Some(#default))
                 }
@@ -312,7 +310,7 @@ fn impl_arg_param(
 
         Ok(quote_arg_span! {
             let #mut_ _tmp: #target_ty = #arg_value_or_default?;
-            #[allow(clippy::needless_option_as_deref)]
+            #[allow(clippy::needless_option_as_deref, clippy::borrow_deref_ref)]
             let #arg_name = #borrow_tmp;
         })
     } else {
--- a/pyo3-macros-backend/src/params.rs
+++ b/pyo3-macros-backend/src/params.rs
@@ -240,6 +240,7 @@ fn impl_arg_param(
                 }
             }
             (Some(default), _) => {
+                
                 quote_arg_span! {
                     _pyo3::impl_::extract_argument::from_py_with_with_default(#arg_value, #name_str, #expr_path, || #default)
                 }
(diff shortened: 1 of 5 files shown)


pyo3-macros-backend/src/params.rs
```diff
@@ -310,12 +310,18 @@ fn impl_arg_param(
...
             let #arg_name = #borrow_tmp;
         })
-    } else {
+    } else if cfg!(feature = "pyproto") {
+        // arg.ty is kinda weird
         Ok(quote_arg_span! {
             let #arg_name = #arg_value_or_default?;
         })
+    } else {
+        let ty = arg.ty;
```
davidhewitt, line 322: This might need a call to `remove_lifetime`. Can you add a test for a pyfunction with something like `Vec<&'py PyAny>` as an argument? I've also been wondering if it's possible for us to copy lifetimes from the function definition onto the generated `pyfunction` - I think that might also solve the problem but not sure.

pyo3-macros-backend/src/params.rs
```diff
@@ -310,12 +310,18 @@ fn impl_arg_param(
 
         Ok(quote_arg_span! {
             let #mut_ _tmp: #target_ty = #arg_value_or_default?;
-            #[allow(clippy::needless_option_as_deref)]
+            #[allow(clippy::needless_option_as_deref, clippy::borrow_deref_ref)]
             let #arg_name = #borrow_tmp;
         })
-    } else {
+    } else if cfg!(feature = "pyproto") {
```
davidhewitt, line 316: Lack of `remove_lifetime` is probably why `pyproto` broke - it inserts a hidden lifetime to try and behave like lifetime elision :(. Super weird and so close to killing off now!

davidhewitt:
Thanks! I've paused on reviewing this one for a bit just because I've always viewed the deref juggling the macro does here as a bit janky. I finally had an idea how to [introduce a `PyFunctionArgument` trait](https://github.com/PyO3/pyo3/compare/main...davidhewitt:pyo3:extract_argument_holder) which should allow the macros to do less (relying on type inference instead).

I haven't quite finished off that branch, I'm hoping I'll get a chance to implement at the weekend and see what folks think.

davidhewitt:
Closing as superseded by #2503

PR #2489, opened 2022-07-01
Rust 1.62

mejrs:


tests/ui/missing_clone.stderr
```diff
@@ -4,9 +4,23 @@ error[E0277]: the trait bound `TestClass: Clone` is not satisfied
...
+    = help: the following other types implement trait `pyo3::FromPyObject<'source>`:
+              <&'a PyCell<T> as pyo3::FromPyObject<'a>>
+              <&'a [u8] as pyo3::FromPyObject<'a>>
+              <&'py CancelledError as pyo3::FromPyObject<'py>>
+              <&'py IncompleteReadError as pyo3::FromPyObject<'py>>
+              <&'py InvalidStateError as pyo3::FromPyObject<'py>>
+              <&'py LimitOverrunError as pyo3::FromPyObject<'py>>
+              <&'py PanicException as pyo3::FromPyObject<'py>>
+              <&'py PyAny as pyo3::FromPyObject<'py>>
+            and 142 others
```
birkenfeld, line 16: The number here seems to fluctuate. Can it be wildcarded out?

davidhewitt: Not aware of a wildcard feature.

Wild guess, could it be related to features? Maybe we need to run UI tests only with `feature = "full"`?

mejrs: Thanks, that's a good idea.

davidhewitt: Ah no-fail-fast has answered it, looks like different number of `Clone` implementations on Windows vs other OS.

I guess it's fine to run this particular UI test just on Linux?

mejrs: > Ah no-fail-fast has answered it, looks like different number of `Clone` implementations on Windows vs other OS.

That, and the abi3/full/etc features. I've just made a compile_fail doctest instead.

davidhewitt: Ah I think that's good enough here, thanks 👍

davidhewitt (approved):
Looks great, thanks again 👍

Issue #2454, opened 2022-06-15
Python Interface (.pyi) generation and runtime inspection


CLOVIS-AI:
Hi!

This issue is going to be a summary of my prototypes to generate Python Interface files ('stub files', '.pyi files') automatically. The prototypes are available as #2379 and #2447.
- #2379 aimed to generate Python stub files entirely at compile-time. This is not possible because proc-macros run before type inference and trait analysis, so proc-macros cannot know if a type implements a trait or not.
- #2447 aims to generate information at compile-time that represents the structure of `#[pyclass]` structs (which methods exist, what arguments do they accept) to be read at run-time by the stub generator.

I'm presenting the results here to get feedback on the current approach. I'm thinking of extracting parts of the prototypes as standalone features and PRs.

# Progress

Accessing type information at runtime:
- [x] #2490
- [ ] Generate precise type information as part of `#[pyclass]`

Accessing structural information at runtime:
- [ ] Declare the inspection API
- [ ] Generate the inspection API structures as part of `#[pyclass]` and `#[pymethods]`
- [ ] Collect the inspection data per module

Python interface generation:
- [ ] Generate the Python interface files
- [ ] Document how to generate PYI files in end user's projects

# Summary

The final goal is to provide a way for developers who use PyO3 to automatically generate Python Interface files (.pyi) with type information and documentation, to enable Rust extensions to 'feel' like regular Python code for end users via proper integration in various tools (MyPy, IDEs, Python documentation generators).

I have identified the following steps to achieve this goal. Ideally, each step will become its own PR as a standalone feature.
1. provide a way to extract the full Python type information from any object passed to/retrieved from Python (e.g. `List[Union[str]]`, not just `PyList`).
2. provide an API to describe Python objects at run-time (list of classes, list of methods for these classes, list of arguments of each method, etc).
3. improve the macros so they generate at compile-time the various inspection data structures (the API from 2.)
4. write a run-time pyi generator based on the inspection API

1 and 2 are independent, 3 and 4 are independent.

## Full type information

The goal of this task is to provide a simple way to access the string representation of the Python type of any object exposed to Python. This string representation should follow the exact format of normal Python type hints.

First, a structure representing the various types is created (simplified version below, [prototype here](https://github.com/PyO3/pyo3/blob/71ed3fe75af73bd1231df435330a9c9e13c42089/src/inspect/types.rs)):
```rust
struct TypeInfo {
    Any,
    None,
    Optional(Box<TypeInfo>),
[...]

Issue #2337, opened 2022-04-26
Smarter CHANGELOG management


davidhewitt:
One of the most common merge conflicts I encounter is our CHANGELOG.

Also, having to annotate the PR number at the end of the CHANGELOG message is a little fiddly. Since discussions were enabled on the repo I keep predicting the PR number incorrectly, leading to typos like #2336 (or just needing to push to correct).

I've seen systems like cpython's [`NEWS.d`](https://github.com/python/cpython/tree/main/Misc/NEWS.d/next) where the items are committed seperately and then merge as part of the release process.

Should we consider something similar here? Anyone know of a nice system for Rust?

messense:
https://github.com/changesets/changesets but in TypeScript, there is [a similar project in Rust](https://github.com/DmitryDodzin/mol) but it doesn't seem mature enough.

mtreinish:
I've been using [reno](https://docs.openstack.org/reno/latest/) for this but it's in python and only has built-in integration with sphinx/rst. The nice thing with it though is that it uses git history to determine which release a release note is part of.

jeertmans:
Hello, maybe you can have a look at [git-clif](https://github.com/orhun/git-cliff), a git tool written in Rust that can be used to automate CHANGELOG generation. This tool works best with [conventional commits](https://github.com/orhun/git-cliff#conventional_commits) (which I don't think you use) but can easily be tweaked to do (almost?) anything you want.

I'm not the author of this tool, but I really enjoy using it and I thought it was worth mentioning it ;-)

birkenfeld:
Should this be a discussion?

aganders3:
A number of prominent Python projects use [towncrier](https://towncrier.readthedocs.io/en/latest/). The drawback I see is that it wants to tie every change to an issue/ticket number. Maybe that could be changed to point to the PR number instead?

CLOVIS-AI:
If you want to go to low barrier of entry but with a little bit of manual work, here's what I do in other projects:
- Add the changelog sections to each pull request, each collaborator fills them in when creating the PR
- When you create a new release, simply run `git log --first-parent --format=%B main`, copy all the changelog sections and create a git tag

It can never create conflicts, and the changelog is available offline at any point by looking at the various tags.

PR #2490, opened 2022-07-01
Runtime type information for objects crossing the Rust–Python boundary

CLOVIS-AI:

 - ~~docs to all new functions and / or detail in the guide~~ Documentation will be added when the whole `inspect` module will be done

---

First step towards https://github.com/PyO3/pyo3/issues/2454.
The goal of this PR is to provide runtime information about Python types for any object that is transferred to/from Python.

Example usage:
```rust
println!("a: {}", <(bool, usize, Vec<HashMap<String, u32>>)>::type_input());
```
```python
a: Tuple[bool, int, Sequence[Mapping[str, int]]]
```

---

This PR introduces two new methods:
- `FromPyObject::type_input`: the Python type when the object appears as a parameter to a Rust function,
- `IntoPy::type_output`: the Python type when the object appears as a return value of a Rust function.

Both methods are implemented for the most important Rust types (primitives like `usize`, collections like `HashMap`).
The default implementation is to return `Any` (which is the recommended type when the exact type is missing, and avoids this PR being a breaking change).

In the future, this will allow the inspection module to generate code calling `type_input` and `type_output` to delay the Python type generation until runtime, which is required because macros are executed before the Rust compiler resolves the types.

Future work:
- Implementing these methods for all other objects (`PyDict`, hashbrown tables,...); this is fairly easy and may be done on a case-by-case basis when more precise information is needed
- Automatically generating the precise type information as part of `#[pyclass]`

### DIFF
--- a/src/inspect/types.rs
+++ b/src/inspect/types.rs
@@ -1,0 +1,396 @@
+use std::fmt::{Display, Formatter};
+
+/// Designation of a Python type.
+///
+/// This enum is used to handle advanced types, such as types with generics.
+/// Its [`Display`] implementation can be used to convert to the type hint notation (e.g. `List[int]`).
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum TypeInfo {
+    /// The type `typing.Any`, which represents any possible value (unknown type).
+    Any,
+    /// The type `typing.None`.
+    None,
+    /// The type `typing.NoReturn`, which represents functions that never return (they can still panic / throw, similar to `never` in Rust).
+    NoReturn,
+    /// The type `typing.Callable`.
+    ///
+    /// The first argument represents the parameters of the callable:
+    /// - `Some` of a vector of types to represent the signature,
+    /// - `None` if the signature is unknown (allows any number of arguments with type `Any`).
+    ///
+    /// The second argument represents the return type.
+    Callable(Option<Vec<TypeInfo>>, Box<TypeInfo>),
+    /// The type `typing.tuple`.
+    ///
+    /// The argument represents the contents of the tuple:
+    /// - `Some` of a vector of types to represent the accepted types,
+    /// - `Some` of an empty vector for the empty tuple,
+    /// - `None` if the number and type of accepted values is unknown.
+    ///
+    /// If the number of accepted values is unknown, but their type is, use [`Self::UnsizedTypedTuple`].
+    Tuple(Option<Vec<TypeInfo>>),
+    /// The type `typing.Tuple`.
+    ///
+    /// Use this variant to represent a tuple of unknown size but of known types.
+    ///
+    /// If the type is unknown, or if the number of elements is known, use [`Self::Tuple`].
+    UnsizedTypedTuple(Box<TypeInfo>),
+    /// A Python class.
+    Class {
+        /// The module this class comes from.
+        module: ModuleName,
+        /// The name of this class, as it appears in a type hint.
+        name: &'static str,
+        /// The generics accepted by this class (empty vector if this class is not generic).
+        type_vars: Vec<TypeInfo>,
+    },
+}
+
+/// Declares which module a type is a part of.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum ModuleName {
+    /// The type is built-in: it doesn't need to be imported.
+    Builtin,
+    /// The type is in the current module: it doesn't need to be imported in this module, but needs to be imported in others.
+    CurrentModule,
+    /// The type is in the specified module.
+    Module(&'static str),
+}
+
+impl TypeInfo {
+    /// Returns the module in which a type is declared.
+    ///
+    /// Returns `None` if the type is declared in the current module.
+    pub fn module_name(&self) -> Option<&'static str> {
+        match self {
+            TypeInfo::Any
+            | TypeInfo::None
+            | TypeInfo::NoReturn
+            | TypeInfo::Callable(_, _)
+            | TypeInfo::Tuple(_)
+            | TypeInfo::UnsizedTypedTuple(_) => Some("typing"),
+            TypeInfo::Class { module, .. } => match module {
+                ModuleName::Builtin => Some("builtins"),
+                ModuleName::CurrentModule => None,
+                ModuleName::Module(name) => Some(name),
+            },
+        }
+    }
+
+    /// Returns the name of a type.
+    ///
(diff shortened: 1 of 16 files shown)


mejrs:
You'll need to rebase on main to make CI work (because of https://github.com/PyO3/pyo3/pull/2489 ) :)

CLOVIS-AI:
@mejrs it's rebased :)

CLOVIS-AI:
Fixed `cargo fmt` and `cargo clippy`

davidhewitt:
👍 looks like you were thinking along the same lines for a first step as me with this PR! Nice work!

So I guess the only question here is whether `type_input` should be used for downcast errors as suggested in #2454? I think I'd like to confirm that use case works before we introduce new public APIs.

Similarly I wonder if users can already use these to include function signatures in docstrings? Might need some macro support. One to think about, can probably implement this in a separate PR if we want it.

Just for completeness, do other use cases for these beyond the introspection API come to mind? I'm sure users will always surprise us 😊

davidhewitt:
Also probably obvious from my other comments, though just wanted to add I'd prefer we merge this after 0.17 is released so that we allow ourselves time to iterate on this further once needed after merge.

CLOVIS-AI:
> Just for completeness, do other use cases for these beyond the introspection API come to mind? I'm sure users will always surprise us blush

Well, this PR itself is just the building block of "what is the actual Python type of this type as seen from Python". I expect that this can be used for hundreds of different things :sweat:. Indeed, debugging/error information seems like a good use case.

> So I guess the only question here is whether `type_input` should be used for downcast errors as suggested in #2454? I think I'd like to confirm that use case works before we introduce new public APIs.

'should' is your decision, but it definitely sounds like it 'could'. I don't think it should be a part of this PR though.

> I'd prefer we merge this after 0.17 is released so that we allow ourselves time to iterate on this further if needed after merge.

Do you have an estimate on when that will be? I'm a bit tight on schedule and there's a lot of things to do until .pyi generation is a thing :)

davidhewitt:
Tracking preparation for 0.17 at https://github.com/PyO3/pyo3/issues/2493 - TBH at the moment I'm working on getting those pieces complete whenever I can. I'd like it to be at most a week or two, though it depends how much time I can manage to free up to make progress.

CLOVIS-AI:
Fixed `rustdoc` & conflicts, rebased on current `main`.

CLOVIS-AI:
I'm not sure I understand why [test](https://github.com/PyO3/pyo3/runs/7608199286?check_suite_focus=true) fails. What should I do to fix it?

messense:
> I'm not sure I understand why [test](https://github.com/PyO3/pyo3/runs/7608199286?check_suite_focus=true) fails. What should I do to fix it?

Run `TRYBUILD=overwrite cargo test` to update ui tests output, see https://github.com/dtolnay/trybuild#workflow

davidhewitt:
@CLOVIS-AI sorry for the long delay with this caused by the 0.17 release. I'm now able to widen bandwidth again, I propose let's merge this and then we can continue iterating on the stub generation? I think this PR is good as-is, just needs a rebase :)

CLOVIS-AI:
Updated the expected test output & rebased.

src/inspect/types.rs
```diff
@@ -0,0 +1,477 @@
...
+    /// Use this variant to represent a tuple of unknown size but of known types.
+    ///
+    /// If the type is unknown, or if the number of elements is known, use [`Self::Tuple`].
+    UnsizedTypedTuple(Box<TypeInfo>),
+    /// A Python class.
+    Class {
+        /// The module this class comes from.
+        module: ModuleName,
+        /// The name of this class, as it appears in a type hint.
+        name: &'static str,
```
mejrs, line 43: Could someone want to create names at runtime? In that case `Cow<'static, str>` is more flexible 🤔

src/inspect/types.rs
```diff
@@ -0,0 +1,477 @@
...
+
+/// Declares which module a type is a part of.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum ModuleName {
+    /// The type is built-in: it doesn't need to be imported.
+    Builtin,
+    /// The type is in the current module: it doesn't need to be imported in this module, but needs to be imported in others.
+    CurrentModule,
+    /// The type is in the specified module.
+    Module(&'static str),
```
mejrs, line 57: Same as above.

src/inspect/types.rs
```diff
@@ -0,0 +1,477 @@
...
+///
+/// This enum is used to handle advanced types, such as types with generics.
+/// Its [`Display`] implementation can be used to convert to the type hint notation (e.g. `List[int]`).
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum TypeInfo {
+    /// The type `typing.Any`, which represents any possible value (unknown type).
+    Any,
+    /// The type `typing.None`.
+    None,
+    /// The type `typing.NoReturn`, which represents functions that never end (similar to `never` in Rust).
```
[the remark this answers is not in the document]
CLOVIS-AI: Done, added clarification in parenthesis

src/inspect/types.rs
CLOVIS-AI: Oh, I didn't know it was possible to create a `Cow` in a `const` context. That does sound more flexible.

src/inspect/types.rs
CLOVIS-AI: Done.

src/inspect/types.rs
CLOVIS-AI: I don't really see a use case for creating names at runtime, but I don't really see a reason to forbid it either.

davidhewitt:
Thanks! With apologies will need one final rebase as #2587 has conflicted on the CHANGELOG. (Really need to find some time to work on #2337 soon.)

Also the coverage job is reporting that a lot of new implementations are not covered by tests, it would be much appreciated if we could add a few more. I find it's much better to add coverage at the same time as new APIs wherever possible rather than hope it gets covered later 😄

CLOVIS-AI:
The rebase is done. I do not have the enough time on my hands at the moment to increase the code coverage. 

I'm not familiar with GitHub's UI for code coverage so I have some trouble reading it, but it seems to me like it is complaining about all the one-liner functions that simply return a `TypeInfo` instance? It's not possible for a bug to hide there (they are just Rust structure instanciations).

davidhewitt:
Understood. Given that I'm the source of the delay that caused you to lose your availability, seems reasonable to me to proceed to merge in that case.

PR #2491, opened 2022-07-03
Upgrade mdbook to 0.4.19

messense:
https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-0419


davidhewitt (approved):
Thanks!

PR #2496, opened 2022-07-06
Pin `plotters` to preserve MSRV

aganders3:
Hopefully this is the right fix/workaround for this. I noticed on my PR that the MSRV check was failing as `criterion` does not strictly pin `plotters` and it was updated with a MSRV-busting change.


aganders3:
Never mind - that didn't work :(

PR #2497, opened 2022-07-06
Pin `plotters` to preserve MSRV (Take 2)

aganders3:
I noticed on my PR that the MSRV check was failing as criterion does not strictly pin plotters and it was updated with a MSRV-busting change.

Hopefully this is the right fix/workaround for this, though it's a bit ugly.


davidhewitt:
Normally what we do is this: https://github.com/PyO3/pyo3/blob/eb2e080f3d735d45e16e225c6b91516ae10788e5/.github/workflows/ci.yml#L183

aganders3:
Thanks - I should have thought to check the workflow file. It's much nicer to not pollute the general Cargo.toml.

davidhewitt (approved):
👍 no worries, thanks for fixing CI!

Issue #2017, opened 2021-11-22
Guide page for "Installation" / "Getting Started"


davidhewitt:
As discussed in https://github.com/PyO3/pyo3/pull/763#issuecomment-972763362

I think it would be helpful to have a page in the guide for first-time users on documenting the ways we recommend to install Python, Rust, set up a virtualenv (maybe?) and install `maturin`.

This might need to be split out by OS. For example in #763 / #1800 we have quite a lot of experience of users who've had trouble on macOS because they've done things in ways that don't quite line up.

mejrs:
I like how https://rustwasm.github.io/docs/book/ approaches this. Very friendly to people new to Rust.

savente93:
I don't have access to a lot of different OSes so I can't necessarily test all the steps, but I'm willing to take a first crack at it so at least there is something people can add to over time, if you like

davidhewitt:
@savente93 if you're willing to spend some time to get this started, that would be brilliant. Anything is better than nothing! I still think this would be a very valuable resource for beginners that I just never quite manage to get around to myself...

PR #2499, opened 2022-07-07
add first version of getting started page to the guide

savente93:
As discussed in #2017 this is my first crack at a getting started page. I tried to dig through some of the issues that mentioned this and the problems I could find, but as said I don't have very in depth knowledge or a bunch of OSes to do research on so I consider this merely a starting point. I have tried to write it in such a way that it should be useful for both people coming from Python and coming from Rust. Hopefully it helps, let me know if any changes are necessary.

### DIFF
--- a/guide/src/getting_started.md
+++ b/guide/src/getting_started.md
@@ -1,0 +1,156 @@
+
+# Instalation
+
+To get started using PyO3 you will need three things: a rust toolchain, a python environment, and a way to build. We'll cover each of these below. 
+
+## Rust
+
+First, make sure you have rust installed on your system. If you haven't already done so you can do so by following the instructions [here](https://www.rust-lang.org/tools/install). PyO3 runs on both the `stable` and `nightly` versions so you can choose whichever one fits you best. The minimum required rust version is Rust 1.48.
+
+if you can run `rustc --version` and the version is high enough you're good to go! 
+
+## Python
+
+To use PyO3 you need at least Python 3.7. While you can simply use the default Python version on your system, it is recommended to use a virtual environment. 
+
+
+## Virtualenvs 
+
+While you can use any virtualenv manager you like, we recommend the use of `pyenv` especially if you want to develop or test for multiple different python versions, so that is what the examples in this book will use. The installation instructions for `pyenv` can be found [here](https://github.com/pyenv/pyenv#getting-pyenv). 
+
+Note that when using `pyenv` you should also set the following environment variable
+```bash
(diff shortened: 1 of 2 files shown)


guide/src/getting_started.md
```diff
@@ -0,0 +1,156 @@
...
+While you can use any virtualenv manager you like, we recommend the use of `pyenv` especially if you want to develop or test for multiple different python versions, so that is what the examples in this book will use. The installation instructions for `pyenv` can be found [here](https://github.com/pyenv/pyenv#getting-pyenv). 
+
+Note that when using `pyenv` you should also set the following environment variable
+```bash
+PYTHON_CONFIGURE_OPTS="--enable-shared"
+```
+### Building
+
+
+There are a number of build and python package management systems such as [`setuptools-rust`](https://github.com/PyO3/setuptools-rust) or [manually](https://pyo3.rs/latest/building_and_distribution.html#manual-builds)  we recommend the use of `maturin`. It is developed to work with PyO3 and is the most "batteries included" experience. `maturin` is just a python package so you can add it in any way that you install python packages.
```
messense, line 28: Add a link to https://maturin.rs/installation.html would be great!

guide/src/getting_started.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Starting a new project
+
+Firstly you should create the folder and virtual environment that are going to contain your new project. Here we will use the recommended `pyenv`: 
+
+```bash
+mkdir pyo3-example
+cd pyo3-example
+pyenv virtualenv pyo3
+pyenv local pyo3
```
messense, line 62: How about adding a paragraph about using `maturin new` if maturin was installed globally or with `pipx`?

```bash
maturin new -b pyo3 pyo3-example
cd pyo3-example
pyenv virtualenv pyo3
pyenv local pyo3
```

guide/src/getting_started.md
```diff
@@ -0,0 +1,156 @@
...
+[dependencies]
+pyo3 = { version = "0.16.5", features = ["extension-module"] }
+```
+
+## pyproject.toml
+You should also create a `pyproject.toml` with the following contents: 
+
+```toml
+[build-system]
+requires = ["maturin>=0.12,<0.13"]
```
messense, line 109:
```suggestion
requires = ["maturin>=0.13,<0.14"]
```

messense (approved):

messense:
Thanks!

Issue #2342, opened 2022-04-27
Don't accept `str` in `FromPyObject` for `Vec<&str>` and `Vec<String>`


davidhewitt:
As per the title, it would be nice if we didn't accept Python `str` in `FromPyObject` for `Vec<&str>` and `Vec<String>`. Some background below.

I'm not sure how hard this will be (without specialization).

### Discussed in https://github.com/PyO3/pyo3/discussions/2341

<div type='discussions-op-text'>

<sup>Originally posted by **atoav** April 27, 2022</sup>
In python one can iterate over a `str` as if it was a `List[str]`. And while this is occasionaly useful it can also be a major footgun if users pass in a `"string"` instead of a `["string"]` and suddenly have a function operate on single characters instead of the whole string, as they might (admittedly, errouniously) assumed one that one string.

In python I'd prevent this user error  by ensuring incoming str are never treated as a list of characters by wrapping a lone `str` in a list as follows:

```python
if isinstance(strings, str):
    strings = [strings]
```

In pyo3 this is an issue as well. Let's say i have a simple function that should take a `Vec<String>` and print out each string in the vec:

```Rust
/// Print out each string in a list of strings
#[pyfunction]
fn print_strings(strings: Vec<String>) {
    for s in strings {
        println!("{}", s);
    }
}
```

If I now run this in the python interpreter with a string that is not a list each character is printed out seperately:  

```python
>>> from my_module import print_strings
>>> print_strings("string")
s
t
r
i
n
g
```

Is there any simple way to ensure single `str` are treated as `[str]` using pyo3? Even an Error would be acceptable, as long as users don't pass a string and have it iterate over characters without them noticing.</div>

jeertmans:
Hi! Shouldn’t this be closed as #2500 was merged?

PR #2500, opened 2022-07-07
Automatically wrap `str` in a `vec![]` for `Vec<&str>` and `Vec<String>`

jeertmans:
# EDIT

After few discussions (see below), it was proposed to raise an error in the case when `str` can be split into a `Vec<T>`.
Thus, only a dynamic type checking is done with `PyAny.isinstance_of::<PyString>()`.

# Old proposition

As discussed in #2342, this PR proposes a way to wrap `str` arguments inside a vector to avoid iterating through chars when it is not desired.

Currently, only the `Vec<String>` is working, as I cannot manage to compile for `Vec<&str>` (code was commented out).

This "solution" leverages the use of the `specialization` features, which is currently unstable. As such, it must be compile with `nightly` feature activated as well as the nightly channel for the rust compiler.

Another solution, as discussed in #2342, would be to throw an error instead of wrapping in a vector.

# Demo

```toml
[package]
name = "fromstr"
# ...

[dependencies]
pyo3 = { path = "...", features = ["extension-module", "nightly"] }
```

```rust
use pyo3::prelude::*;   
                                                            
#[pyfunction]   
fn print_strings(strings: Vec<String>) {            
    for s in strings {                                                                          
        println!("{}", s);                       
    }           
}                              
                                        
                          
#[pyfunction]                                                                                   
fn print_str(strings: Vec<&str>) {
    for s in strings {                      
        println!("{}", s);     
    }                                      
}                
         
#[pymodule]              
fn fromstr(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(print_strings, m)?)?;
    m.add_function(wrap_pyfunction!(print_str, m)?)?;
    Ok(())                                                           
}   
```

```python
from fromstr import print_strings, print_str
                     
if __name__ == "__main__":
             
    print(1)                            
    print_strings("echo")
    print(2)              
    print_str("echo")
    print(3)
    print_strings(["echo"])
    print(4)
    print_str(["echo"])

# Outputs:
# 1
# echo
# 2
# e
# c
# h
# o
# 3
# echo
# 4
# echo
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Prevent multiple `#[pymethods]` with the same name for a single `#[pyclass]`. [#2399](https://github.com/PyO3/pyo3/pull/2399)
 - Fixup `lib_name` when using `PYO3_CONFIG_FILE`. [#2404](https://github.com/PyO3/pyo3/pull/2404)
 - Iterators over `PySet` and `PyDict` will now panic if the underlying collection is mutated during the iteration. [#2380](https://github.com/PyO3/pyo3/pull/2380)
+- `FromPyObject::extract` now raises an error if source type is `PyString` and target type is `Vec<T>`. [#2500](https://github.com/PyO3/pyo3/pull/2500)
 
 ### Fixed
 
--- a/src/types/sequence.rs
+++ b/src/types/sequence.rs
@@ -367,6 +367,19 @@ mod tests {
             assert!(<PySequence as PyTryFrom>::try_from(v.to_object(py).as_ref(py)).is_ok());
         });
     }
+
+    #[test]
+    fn test_strings_cannot_be_extracted_to_vec() {
+        Python::with_gil(|py| {
+            let v = "London Calling";
+            let ob = v.to_object(py);
+
+            assert!(ob.extract::<Vec<&str>>(py).is_err());
+            assert!(ob.extract::<Vec<String>>(py).is_err());
+            assert!(ob.extract::<Vec<char>>(py).is_err());
+        });
+    }
+
     #[test]
     fn test_seq_empty() {
         Python::with_gil(|py| {
--- a/src/types/sequence.rs
+++ b/src/types/sequence.rs
@@ -1,8 +1,8 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
-
 use crate::err::{self, PyDowncastError, PyErr, PyResult};
+use crate::exceptions::PyValueError;
 use crate::internal_tricks::get_ssize_index;
-use crate::types::{PyAny, PyList, PyTuple};
+use crate::types::{PyAny, PyList, PyString, PyTuple};
 use crate::{ffi, PyNativeType, ToPyObject};
 use crate::{AsPyPointer, IntoPyPointer, Py, Python};
 use crate::{FromPyObject, PyTryFrom};
@@ -270,6 +270,9 @@ where
     T: FromPyObject<'a>,
 {
     fn extract(obj: &'a PyAny) -> PyResult<Self> {
+        if let Ok(true) = obj.is_instance_of::<PyString>() {
+            return Err(PyValueError::new_err("Can't extract `str` to `Vec`"));
+        }
         extract_sequence(obj)
     }
 }


birkenfeld:
IMO it's not ok to silently change behavior when `nightly` is activated.

jeertmans:
@birkenfeld I agree that the user should be somehow notified of this, either through a warning or some error?

davidhewitt:
Gotta agree that I'm not a fan of having behavior change by a feature.

I know I copied a quote in #2342 which suggested this wrapping, but how about a simpler solution: just add a check when extracting `Vec<T>`  that the input is not a string (and return an error if it is).

Without specialization we'd have to pay that cost even for non-text types like `Vec<i32>`. Though if we measure it, I'd hope it would be almost unnoticeable.

That would at least hopefully work on stable...?

davidhewitt:
Do we need the `TypeId` checks? I was wondering about the simpler:

```rust
if let Ok(true) = obj.is_instance_of::<PyString>()
{
    return Err(PyValueError::new_err("Can't extract `str` to `Vec`"))
}
extract_sequence(obj)
```

Just depends how much that affects performance I think?

jeertmans:
The idea behind that kind of type checking (either with ˋTypeId` or specialization) was to still allow a `str` to be transformed into, e.g., a ˋVec<char>ˋ.
But maybe this is not desirable as well and using isinstance is sufficient in this case.

jeertmans:
@davidhewitt Here is a small benchmark I ran and its results:

*(Don't mind about the `print...` function names, it should be `count...`)*

`pyo3/src/types/sequence.rs`
```rust
impl<'a, T> FromPyObject<'a> for Vec<T>
where
    T: FromPyObject<'a>, 
{
        if let Ok(true) = obj.is_instance_of::<PyString>() {
            return Err(PyValueError::new_err("Can't extract `str` to `Vec`"));
        }
        extract_sequence(obj)
    }
}
```

`fromstr/src/lib.rs`
```rust
use pyo3::prelude::*;

#[pyfunction]
fn print_strings(strings: Vec<String>) -> usize {            
    strings.len()
}                              

  
#[pyfunction]  
fn print_str(strings: Vec<&str>) -> usize {
     strings.len()  
} 
                                         

#[pyfunction]  
fn print_int(strings: Vec<isize>) -> usize {
    strings.len() 
}                                                             
                                                                                
#[pyfunction]
fn print_char(strings: Vec<char>) -> usize {
    strings.len()
}  
  
#[pymodule]   
fn fromstr(_py: Python, m: &PyModule) -> PyResult<()> {                                                             
    m.add_function(wrap_pyfunction!(print_strings, m)?)?;                                                
    m.add_function(wrap_pyfunction!(print_str, m)?)?; 
    m.add_function(wrap_pyfunction!(print_int, m)?)?;        
    m.add_function(wrap_pyfunction!(print_char, m)?)?;         
    Ok(())                                             
}
```

`main.py`
```python
  from fromstr import *
  from timeit import timeit
             
  if __name__ == "__main__":                                 
                 
      number = 100000          
                                        
      i = list(range(10)) 
      c = list("abcdefg")                                                                       
      s = "some random sentence".split(" ")
                                                                     
      print(timeit(lambda: print_strings(s), number=number) / number)
      print(timeit(lambda: print_str(s), number=number) / number)
      print(timeit(lambda: print_char(c), number=number) / number)
      print(timeit(lambda: print_int(i), number=number) / number)
```

# Without the `isinstance` call

```console
(venv) ➜  fromstr python main.py
4.702104139996663e-06                               
3.5061912999935886e-06                              
5.810010350005541e-06                               
5.9074885100017125e-06                              
(venv) ➜  fromstr python main.py
4.845567360007408e-06                               
3.9516925000043556e-06                              
5.810035859994969e-06                               
5.726638340001955e-06
```

# With the `isinstance` call

```console
(venv) ➜  fromstr python main.py 
4.092408500000601e-06
2.954179230000591e-06
4.588532469997517e-06
4.370560449997356e-06
(venv) ➜  fromstr python main.py
3.9268650299982255e-06
2.893349019996094e-06
4.480811690000337e-06
4.1616317100033485e-06
```

davidhewitt:
Hmm, those `timeit` numbers are really hard to interpret for me.

I did some benchmarking myself, my rough conclusion is that on my machine extracting an empty `Vec` takes about 100ns, and this check adds an additional 10ns (so 10%).

Extracting 10 `&str` elements takes another 100ns, or another way at looking at this is that we pay a cost similar to if the `Vec` was just one element larger.

Seems ok to me for correctness?

davidhewitt:
Agreed. Correctness has to be the top priority, and as new compiler functionality emerges (and new language patterns) we can refine implementations to be more efficient.

A few things that this needs before ready for merge:
- An entry in the `Changed` section of the CHANGELOG
- A test for this new error case (extracting `str` as `Vec` should indeed fail).
- To fix CI you'll need to rebase on #2504 after merging (that PR includes a fix for the MSRV failure).

jeertmans:
@davidhewitt that should all be done now :-)
Please tell me if I am missing something

davidhewitt (approved):
Looks great to me, thanks for iterating on this!

jeertmans:
It was a pleasure, thank you for your time!

alex:
Hi all, I realize I'm leaving this comment many months after this was landed, but...

Is there a reason ValueError was chosen here? TypeError seems more appropriate to me, and I wanted to ask if there was a reason it was rejected.

davidhewitt:
@Alex that's a reasonable suggestion; I didn't have a strong opinion when reviewing, however `TypeError` probably is better. PR welcome!

If this behaviour is useful to you, would you be willing to also join the discussion in #2632? The proposal there would remove this behaviour, and I'm undecided as I see both sides of the debate, so I would value having more user input.

alex:
Sure, happy to do a PR to switch to TypeError :-)

PR #2503, opened 2022-07-12
pyfunction: use extract_argument with holder to avoid extractext

davidhewitt:
Motivated by #2487, this is hopefully a better long-term solution than patching to avoid upsetting clippy.

This is a reworking of the pyfunction macro implementations to have some smarter internal type inference for extracting reference values. At the moment there is an `ExtractExt` trait and a lot of generated macro code dancing with derefs which make clippy quite unhappy.

This PR replaces `ExtractExt` with a `PyFunctionArgument` trait which avoids the need for deref juggling by taking a `&mut holder` argument used as temporary storage. For `T: FromPyObject` types the holder is just `()` (and should hopefully be optimised out), and for `&T` / `&mut T` for `#[pyclass]` types the holder is `Option<PyRef<T>>` and `Option<PyRefMut<T>>` (`Option` because if the wrong argument is passed from Python then the holder never gets populated).

I was careful to check both benchmarks and generated lines of code, and I got no noticeable difference to either. The net result is more friendly on type inference, leads to simpler macro code, and avoids clippy trouble.

To have a practical example of what this change results as - for a function which looks like below:

```rust
#[pyclass]
struct Foo;

#[pyfunction]
fn foo(_value: &Foo) { }
```

The relevant generated code has changed from:

```rust
let _tmp: <&Foo as _pyo3::derive_utils::ExtractExt<'_>>::Target =
        _pyo3::impl_::extract_argument::unwrap_required_argument(output[0usize]),
        "_value",
    )?;
#[allow(clippy::needless_option_as_deref)]
let arg0 = &*_tmp;
foo(arg0);
```

to:

```rust
foo(
        _pyo3::impl_::extract_argument::unwrap_required_argument(output[0usize]),
        &mut { _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT },
        "_value",
    )?
);
```


davidhewitt:
For the "bad import" case described in #2487, this PR appears to also cleanly solve that:

```rust
// missing PathBuf import
use pyo3::prelude::*;

#[pyfunction]
fn foo(_path: Option<PathBuf>) { }
```

gives an error exactly as one would hope:

```
error[E0412]: cannot find type `PathBuf` in this scope
 --> src/lib.rs:6:22
  |
6 | fn foo(_path: Option<PathBuf>) { }
  |                      ^^^^^^^ not found in this scope
  |
help: consider importing this struct
  |
3 | use std::path::PathBuf;
  |
```

davidhewitt:
I'm sufficiently convinced that this is better than the `ExtractExt` implementation that I will proceed to merge now.

mejrs:
Thanks :)

PR #2507, opened 2022-07-14
Update Python to 3.11.0b4 in emscripten test

messense:
Try to fix emscripten test failure.


hoodmane:
Well the test failure on the main branch seems to be because `search_lib_dir` is failing to find `sysconfigdata`. I'm not sure why that started happening, the first commit where the problem occurs looks pretty unrelated. The memory error looks more confusing, is there any way you could make it print the traceback? I'm pretty busy today but I can spend some time looking into this over the weekend if you don't figure it out before then.

messense:
Traceback:

```python
---- exceptions::asyncio::tests::CancelledError stdout ----
thread 'main' panicked at 'Can not import module asyncio: Traceback (most recent call last):
  File "/lib/python3.11/asyncio/__init__.py", line 8, in <module>
    from .base_events import *
  File "/lib/python3.11/asyncio/base_events.py", line 45, in <module>
    from . import staggered
  File "/lib/python3.11/asyncio/staggered.py", line 6, in <module>
    import typing
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1149, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 936, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1083, in get_code
  File "<frozen importlib._bootstrap_external>", line 746, in _code_to_timestamp_pyc
', src/exceptions.rs:698:5
```

hoodmane:
Looks pretty weird...

davidhewitt:
@messense does this repro if you just `import typing`? I ask because that's the module in the traceback which it looks like the import is attempting when this fails.

Code raising the error is here: 

https://github.com/python/cpython/blob/5a7e1e0a92622c605ab28c95b3dd61a21d285ad5/Lib/importlib/_bootstrap_external.py#L746

Looks like it's trying to serialise the module source to a `bytearray`. Running out of memory or exceeding a page size or something? Either in `marshal.dumps` or `bytearray.extend`. @hoodmane maybe you have an idea how wasm memory constraints might affect this?

messense:
> does this repro if you just import typing?

Same `MemoryError`:

```
---- exceptions::tests::test_import_typing stdout ----
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'MemoryError'>, value: MemoryError(), traceback: None }', src/exceptions.rs:749:47
```

hoodmane:
There might be a linker flagthat will fix it. In Pyodide we pass a couple relevant flags:
https://github.com/pyodide/pyodide/blob/ebea82caffa6d05d6c56da764ac388ec9726d1dc/Makefile.envs#L106

noxfile.py
```diff
@@ -191,6 +191,7 @@ def test_emscripten(session: nox.Session):
     )
     session.env["CARGO_BUILD_TARGET"] = target
     session.env["PYO3_CROSS_LIB_DIR"] = pythonlibdir
+    session.env["PYTHONDONTWRITEBYTECODE"] = "1"
```
hoodmane, line 194: I'm pretty sure this won't work because it doesn't set the environment variable inside the VM.

messense:
> There might be a linker flagthat will fix it. In Pyodide we pass a couple relevant flags: https://github.com/pyodide/pyodide/blob/ebea82caffa6d05d6c56da764ac388ec9726d1dc/Makefile.envs#L106

`ALLOW_MEMORY_GROWTH=1` worked like a charm!

davidhewitt (approved):
Nice one, thanks for getting to the bottom of this!

Issue #2510, opened 2022-07-14
Support `Py_EnterRecursiveCall` and `Py_LeaveRecursiveCall`?


samuelcolvin:
I'm struggling with some recursive code in pydantic-core, particularly related to wasm and pyodide, see https://github.com/samuelcolvin/pydantic-core/pull/167.

It was suggested I use `Py_EnterRecursiveCall`, is there any case where this would help? Is it something pyo3 could help/support?

I had assumed recursive code in rust avoided the stack and python's recursion limit even when working with python objects, but messing with `setrecursionlimit` is effecting when tests pass and fail in wasm. 

To be honest, I'm out of my depth here, any guidance here would be appreciated. Thanks in advance.

davidhewitt:
> recursive code in rust avoided the stack 

Recursive Rust code definitely uses the stack - it's where all function locals and return addresses etc reside. e.g. here's the simplest stack overflow you could possibly hit: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f47a031011c7aff7d62a65519ca66700

Not sure if this is helpful, if you really need to have unlimited recursion the way I tend to do it is transform recursive function calls into loops.

e.g.

```rust
enum RecursiveData {
    Many(Vec<RecursiveData>),
    One(String),
}

fn recursive_flatten(data: RecursiveData) -> Vec<String> {
    match data {
        RecursiveData::Many(datas) => datas.into_iter().map(recursive_flatten).flatten().collect(),
        RecursiveData::One(string) => vec![string],
    }
}
```

can become this loop version which is bounded only by the memory of the machine:

```rust
enum RecursiveData {
    Many(Vec<RecursiveData>),
    One(String),
}

fn loop_flatten(data: RecursiveData) -> Vec<String> {
    match data {
        RecursiveData::Many(datas) => {
            let mut paused = Vec::new();
            let mut out = Vec::new();
            let mut current = datas.into_iter();
            loop {
                // Process next value from deepest iterator
                while let Some(next) = current.next() {
                    match next {
                        RecursiveData::Many(next_datas) => {
                            // Save current for later, descend into new level
                            paused.push(std::mem::replace(&mut current, next_datas.into_iter()));
                        },
                        RecursiveData::One(string) => out.push(string),
                    }
                }
                // Current is exhausted, try resuming
                if let Some(prev) = paused.pop() {
                    current = prev;
                } else {
                    // All done
                    return out;
                }
            }
        },
[...]

PR #2511, opened 2022-07-15
Add FFI definitions for `Py_EnterRecursiveCall` and `Py_LeaveRecursiveCall`

messense:
https://docs.python.org/3/c-api/exceptions.html#recursion-control

#2510 cc @samuelcolvin

### DIFF
--- a/pyo3-ffi/src/ceval.rs
+++ b/pyo3-ffi/src/ceval.rs
@@ -74,9 +74,11 @@ extern "C" {
 }
 
 extern "C" {
-    #[cfg(any(Py_3_9, not(Py_LIMITED_API)))]
+    #[cfg(Py_3_9)]
+    #[cfg_attr(PyPy, link_name = "PyPy_EnterRecursiveCall")]
     pub fn Py_EnterRecursiveCall(arg1: *const c_char);
-    #[cfg(any(Py_3_9, not(Py_LIMITED_API)))]
+    #[cfg(Py_3_9)]
+    #[cfg_attr(PyPy, link_name = "PyPy_LeaveRecursiveCall")]
     pub fn Py_LeaveRecursiveCall();
 }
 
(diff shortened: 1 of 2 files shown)


pyo3-ffi/src/ceval.rs
```diff
@@ -73,8 +73,12 @@ extern "C" {
     fn _Py_CheckRecursiveCall(_where: *mut c_char) -> c_int;
 }
 
-// skipped Py_EnterRecursiveCall
-// skipped Py_LeaveRecursiveCall
+extern "C" {
+    #[cfg(any(Py_3_9, not(Py_LIMITED_API)))]
+    pub fn Py_EnterRecursiveCall(arg1: *const c_char);
+    #[cfg(any(Py_3_9, not(Py_LIMITED_API)))]
+    pub fn Py_LeaveRecursiveCall();
```
davidhewitt, line 80: It looks like these were added to the stable API in 3.9, and also PyPy has them:

```suggestion
    #[cfg(Py_3_9)]
    #[cfg_attr(PyPy, link_name = "PyPy_EnterRecursiveCall")]
    pub fn Py_EnterRecursiveCall(arg1: *const c_char);
    #[cfg(Py_3_9)]
    #[cfg_attr(PyPy, link_name = "PyPy_LeaveRecursiveCall")]
    pub fn Py_LeaveRecursiveCall();
```

(In older versions they were implemented [as macros using unlimited API features](https://github.com/python/cpython/blob/f78733b332966cef7d0c61165389549f4ffe0727/Include/ceval.h#L89-L91), so we could maybe support in the unlimited API in older versions if @samuelcolvin needs that?)

davidhewitt (changes requested):
Think we can support more than just unlimited API - see other comment.

If people think it's worth it, we could potentially also add a safe API along the lines of `py.enter_recursive_call(some_cstr, || { /* stuff */ })`

davidhewitt (approved):
Thanks! Happy to leave the safe API to a separate PR if it's wanted.

davidhewitt:
Having some problems with squash on my phone, will try later...

PR #2518, opened 2022-07-19
Use Python::with_gil in tests

mejrs:
This just updates the tests - no actual changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,7 +181,7 @@ jobs:
         run: |
           set -x
           cargo update -p indexmap --precise 1.6.2
-          cargo update -p hashbrown:0.12.2 --precise 0.9.1
+          cargo update -p hashbrown:0.12.3 --precise 0.9.1
           PROJECTS=("." "examples/decorator" "examples/maturin-starter" "examples/setuptools-rust-starter" "examples/word-count")
           for PROJ in ${PROJECTS[@]}; do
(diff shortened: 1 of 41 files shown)


.github/workflows/ci.yml
```diff
@@ -181,7 +181,7 @@ jobs:
         run: |
           set -x
           cargo update -p indexmap --precise 1.6.2
-          cargo update -p hashbrown:0.12.2 --precise 0.9.1
+          cargo update -p hashbrown:0.12.3 --precise 0.9.1
```
mejrs, line 184: Uh, so does this need bumping every time hashbrown releases? 🤔

adamreichold: Yes. In `rust-numpy`, I avoid this using Python scripting but that is not pretty either: https://github.com/PyO3/rust-numpy/blob/main/.github/workflows/ci.yml#L133-L143

davidhewitt: Doing it programatically makes sense; could potentially put that into the `noxfile` / `xtask` so we can just do `nox -s set-mrsv-versions` here?

davidhewitt (approved):
Thanks very much for doing this! I'd better get on and rebase #1788 to avoid people adding new `acquire_gil` tests!

.github/workflows/ci.yml
davidhewitt: (We can leave this for a separate PR)

PR #2523, opened 2022-07-26
pymodule: only allow initializing once per process

davidhewitt:
Following the suggestion in https://github.com/PyO3/pyo3/discussions/2346#discussioncomment-2913833 this changes `#[pymodule]` to raise an `ImportError` if initialised more than once. This seems to be a necessary defence against soundness issues related to sub-interpreters.

### DIFF
--- a/src/impl_/pymodule.rs
+++ b/src/impl_/pymodule.rs
@@ -1,10 +1,13 @@
 //! Implementation details of `#[pymodule]` which need to be accessible from proc-macro generated code.
 
-use std::cell::UnsafeCell;
+use std::{
+    cell::UnsafeCell,
+    sync::atomic::{self, AtomicBool},
+};
 
 use crate::{
-    callback::panic_result_into_callback_output, ffi, types::PyModule, GILPool, IntoPyPointer, Py,
-    PyObject, PyResult, Python,
+    callback::panic_result_into_callback_output, exceptions::PyImportError, ffi, types::PyModule,
+    GILPool, IntoPyPointer, Py, PyObject, PyResult, Python,
 };
 
 /// `Sync` wrapper of `ffi::PyModuleDef`.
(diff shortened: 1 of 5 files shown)


src/impl_/pymodule.rs
```diff
@@ -50,13 +54,20 @@ impl ModuleDef {
...
             initializer,
+            initialized: AtomicBool::new(false),
         }
     }
     /// Builds a module using user given initializer. Used for [`#[pymodule]`][crate::pymodule].
     pub fn make_module(&'static self, py: Python<'_>) -> PyResult<PyObject> {
         let module = unsafe {
             Py::<PyModule>::from_owned_ptr_or_err(py, ffi::PyModule_Create(self.ffi_def.get()))?
         };
+        if self.initialized.load(atomic::Ordering::SeqCst) {
```
adamreichold, line 65: I think a single

```rust
if self.initialized.swap(true, atomic::Ordering::SeqCst) {
```

should suffice?

davidhewitt: 👍 thanks - I was tired yesterday!

adamreichold (approved):

linas:
This change seems to have broken multiple downstream users; see for example https://github.com/pyca/cryptography/issues/9016 and I'm trying to track the impacted projects here: https://github.com/pyca/cryptography/issues/12080 It's kind of overwhelming because of the large amounts of chatter on so many different discussion groups and distro bug trackers that have been hit by this bug. 

One possible fix is to apply this diff: https://github.com/zer0def/pyo3/commit/eba5243197995f6b37bef491f36d5a1ad0ee03e3 which allows unsafe operation. Unclear if there's an open pull req for this, or whether it has been merged.

alex:
FWIW, I think while it's probably not possible to have a fully safe sub-interpreter support (without huge perf overhead, at least), I do think if we can get to "pyo3 has no static state, and offers users a module state for their own stuff", then we can offer an unsafe "I promise I don't store any state statically" thing to allow sub-interpreters.

But the core work is:

1. Multi-phase init
2. Module state API

Issue #2521, opened 2022-07-25
Alternative recommendation for workspaces with an extension-module and linker issues


acshi:
Right now on the FAQ, there is an entry for `cargo test` problems when using an extension-module, recommending to use `cargo test --no-default-features`.

However, this is not the only way that these problems might surface. For me, I couldn't run `cargo check` from my workspace route, because of the additiveness of cargo features and using pyo3 both in an extension module and another use in different crates.

My solution was a combination and reversion of existing suggestions:
In my extension-module `Cargo.toml`
```
[lib]
name = "name"
crate-type = ["cdylib"]

[dependencies]
pyo3 = "0.15.2"

[features]
# NOT a default feature
extension-module = ["pyo3/extension-module"]
```

Then I build with `maturin develop --features extension-module --release`

This way, the feature is _only_ enabled when it is needed. I think this is a more composable and general solution than the current recommendation.

Ideally, this recommendation, or at least a link to it on the FAQ, could be linked from the main page (https://pyo3.rs/v0.14.2/index.html), since using PyO3 in a workspace seems a fairly common use-case to me and I see any drawbacks to this solution right now, besides a slightly longer maturin build command.

davidhewitt:
Thanks for sharing, that's an interesting approach, I agree it's probably better.

I think you could even go further and add the following to your `pyproject.toml`:

```
[tool.maturin]
cargo-extra-args = ["--features", "extension-module"]
```

That way I think `maturin` will pass the feature automatically. (Similar could also be done with `setuptools-rust`.)

Would you be willing to push a PR to add this recommendation to the FAQ?

acshi:
Great! Yeah, I'll see if I can send a PR for this later this week

PR #2524, opened 2022-07-28
Documentation of another workaround for enabling the extension-module feature

acshi:
Closes https://github.com/PyO3/pyo3/issues/2521

I don't think this needs an entry in the change log because it is just documentation.

### DIFF
--- a/guide/src/faq.md
+++ b/guide/src/faq.md
@@ -15,9 +15,21 @@ PyO3 provides a struct [`GILOnceCell`] which works equivalently to `OnceCell` bu
 
 [`GILOnceCell`]: {{#PYO3_DOCS_URL}}/pyo3/once_cell/struct.GILOnceCell.html
 
-## I can't run `cargo test`: I'm having linker issues like "Symbol not found" or "Undefined reference to _PyExc_SystemError"!
+## I can't run `cargo test`; or I can't build in a Cargo workspace: I'm having linker issues like "Symbol not found" or "Undefined reference to _PyExc_SystemError"!
 
-Currently, [#340](https://github.com/PyO3/pyo3/issues/340) causes `cargo test` to fail with linking errors when the `extension-module` feature is activated. For now you can work around this by making the `extension-module` feature optional and running the tests with `cargo test --no-default-features`:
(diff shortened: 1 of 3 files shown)


guide/src/faq.md
```diff
@@ -15,9 +15,21 @@ PyO3 provides a struct [`GILOnceCell`] which works equivalently to `OnceCell` bu
...
+
+```toml
+[dependencies.pyo3]
+{{#PYO3_CRATE_VERSION}}
+
+[features]
+extension-module = ["pyo3/extension-module"]
+```
+
+2. Make the `extension-module` feature optional and default. Run tests with `cargo test --no-default-features`:
```
messense, line 32: I think we can also mention `[tool.maturin]` section in `pyproject.toml`, see https://github.com/PyO3/pyo3/discussions/2271#discussioncomment-2497276

```toml
[tool.maturin]
```

acshi: I didn't include that because I couldn't get it to work for me. When I had errors in the format of `pyproject.toml`, it wouldn't compile, but once I fixed them, I still got the warning from maturin that it wasn't compiling with the `extension-module` feature. And actually I just grepped the pyo3 repo and I can't find documentation explaining `pyproject.toml` anywhere either.

If we do add this as a third workaround, I think we need to be able to link to some documentation that explains `pyproject.toml` so it is clear _why_ this works too?

messense: https://maturin.rs/metadata.html#add-maturin-build-options, you need maturin 0.13 for `features` or `cargo-extra-args` on maturin 0.12.

acshi: Okay, I was confused because I still get "Warning: You're building a library without activating pyo3's `extension-module` feature." but actually it does properly work. It is just that the warning is emitted before the pyproject.toml is parsed. I'll add this as a third workaround.

messense: Hmm, that warning might be a bug in maturin, can you create a issue in https://github.com/PyO3/maturin/issues? Thanks!

acshi: Done! https://github.com/PyO3/maturin/issues/1029

And how does the PR look now?

messense (approved):
Thanks!

PR #2526, opened 2022-07-29
Fix emscripten test and pin python version

messense:
I think it needs to be keeped in sync, otherwise every new 3.11 beta release breaks the test.


messense:
I don't have enough Emscripten knowledge to update it to Python 3.11b5, gonna just pin it to 3.11b4 for now.

davidhewitt:
👍 seems reasonable to me, thanks!

PR #2527, opened 2022-08-01
add `as_sequence()` method on dict views

PrettyWood:
Hi and thank you for this amazing library! I'm trying to help @samuelcolvin on [https://github.com/samuelcolvin/pydantic-core](pydantic-core). Following https://github.com/PyO3/pyo3/pull/2358, it would be great to have at list a `as_sequence()` method on dict views to be able to easily get length, convert them to list, iterate...
I didn't want to add any other methods following @davidhewitt 's comment. Let's keep it simple for now :)

### DIFF
--- a/src/types/dict.rs
+++ b/src/types/dict.rs
@@ -3,7 +3,7 @@
 use super::PyMapping;
 use crate::err::{self, PyErr, PyResult};
 use crate::ffi::Py_ssize_t;
-use crate::types::{PyAny, PyList};
+use crate::types::{PyAny, PyList, PySequence};
 #[cfg(not(PyPy))]
 use crate::IntoPyPointer;
 use crate::{ffi, AsPyPointer, FromPyObject, IntoPy, PyObject, PyTryFrom, Python, ToPyObject};
@@ -34,6 +34,13 @@ pyobject_native_type_core!(
     #checkfunction=ffi::PyDictKeys_Check
 );
 
+#[cfg(not(PyPy))]
+impl PyDictKeys {
+    pub fn as_sequence(&self) -> &PySequence {
+        unsafe { PySequence::try_from_unchecked(self) }
+    }
+}
+
 /// Represents a Python `dict_values`.
 #[cfg(not(PyPy))]
 #[repr(transparent)]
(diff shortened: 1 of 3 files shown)


samuelcolvin (approved):
LGTM :+1:

PrettyWood:
I don't know what's wrong on windows and if it's even related to this PR 🤔

src/types/dict.rs
```diff
@@ -46,6 +57,13 @@ pyobject_native_type_core!(
     #checkfunction=ffi::PyDictValues_Check
 );
 
+#[cfg(not(PyPy))]
+impl PyDictValues {
+    pub fn as_sequence(&self) -> &PySequence {
+        unsafe { PySequence::try_from_unchecked(self) }
+    }
+}
+
```
mejrs, line 66: This should have a comment similar to the existing `as_sequence` methods (same for PyDictItems)

mejrs (commented):
Thanks!

> I don't know what's wrong on windows and if it's even related to this PR 🤔

Looks like `The filename or extension is too long. (os error 206)`...ugh. I don't know what's up with that.

PrettyWood:
dankjewel @mejrs for the quick review! 🙏

davidhewitt:
Hey, thanks for this. 

I'm definitely willing to add some methods to these views to improve usability, although I'm not convinced `.as_sequence()` is strictly correct. This appears to break type-safety, as I don't think these collections are sequences:

```python
>>> import collections
>>> isinstance({}.keys(), collections.abc.Sequence)
False
```

You can already make use of the fact these views implement `Deref<Target=PyAny>` to use `.len()`, `.iter()` etc.

If we add methods to these types I would guess the main motivation would either be:
- performance if there's a C API we can make use of
- remove need for `PyResult<_>` on `.len()` etc.

Is it possible that someone could subtype `dict_keys` (e.g.) and create a `__len__` which raises an exception? I think so, in which case we have to make a judgement call whether a bespoke `fn len() -> usize` on these types would panic in that case. So far in PyO3 we've made the decision to return results in all such cases so as to minimise library-level panics.

PrettyWood:
Ah you're absolutely right @davidhewitt! Thank you!
Since order is preserved, I thought dict views would be sequences but since they are not we can close this PR.
----------------------
At first I also added a `len()` method indeed to avoid having a `Result` type but I understand the judgement call.

Thanks again for this library!

PR #2538, opened 2022-08-06
pyo3-build-config: add link args for wasm32-unknown-emscripten

davidhewitt:
Follow-up to https://github.com/PyO3/setuptools-rust/pull/244, I think it makes sense for PyO3 to set some link args directly for `wasm32-unknown-emscripten`.


davidhewitt:
NB these might eventually make it into a future stable rustc version: https://github.com/rust-lang/rust/pull/98358/files

messense (approved):

PR #2540, opened 2022-08-09
ci: downgrade libc to 0.2.127 for MSRV

aganders3:
Similar workaround as https://github.com/PyO3/pyo3/pull/2539, but for `libc` - hopefully this issue doesn't keep spreading!

See the error here:
https://github.com/PyO3/pyo3/runs/7753898046?check_suite_focus=true

and here, which is why I added this to the example directories as well:
https://github.com/PyO3/pyo3/runs/7754192650?check_suite_focus=true


aganders3:
I think this can probably be closed as that release has been yanked.

https://github.com/rust-lang/libc/issues/2866

PR #2542, opened 2022-08-10
pypy: disable PyFunction

davidhewitt:
Apparently this is an incomplete definition so it's not safe to use, see https://foss.heptapod.net/pypy/pypy/-/issues/3776


mejrs (approved):
This needs a rebase on the clippy fixes, I think.

mattip:
Is there a downside to making this an opaque type? Will this limit what can be done with PyO3 and PyPy?

davidhewitt:
Disabling this doesn't have a huge impact, it just means that users cannot check for this type using PyO3's safe APIs. They can still use the raw `ffi` call to do so.

The problem is that PyO3's type modelling assumes that a `type_object() -> &PyType` method can be implemented which hands out a valid `PyType` object to users. And because `PyFunction::type_object()` was handing out a `&PyType` with uninitialized fields, users hit crashes.

mattip:
PyPy tries to implement the c-api emulation on a "need to have" basis. Do you know what project was hitting crashes, so I can interact with them and make the changes necessary to make the code work?

davidhewitt:
Sure thing - it was [this code in `pydantic-core`](https://github.com/pydantic/pydantic-core/blob/03f7b4423c6400dc59a913297604f241971e1456/src/validators/typed_dict.rs#L458-L464).

mattip:
right, that is the `PyObject_IsInstance(<something>, &PyFunction_Type)` problem. Is that the only waythis type object is used?

davidhewitt:
I think if PyO3 exposed the `PyFunction` type object as a high-level API then I wouldn't place a limit on how it might be used. Methods on the `PyType` object include [getting the `__qualname__` attribute or using it in `PyObject_IsSubclass`](https://github.com/PyO3/pyo3/blob/676227d2a1b6fce8aa33973dca10684f6237b7bd/src/types/typeobject.rs#L37-L49), however users could also just use it anywhere a `PyAny` is expected so unless this object is fully initialised I wouldn't be confident that users wouldn't produce crashes.

PR #2544, opened 2022-08-11
safety: abort if dropping panic payloads panic

### LINKED ISSUE #2501: `run_closure` and `drop_closure` unsoundly drop payload on panic
### Bug Description

These two functions use `catch_unwind` to catch panics and transfer them as Python exceptions (for `run_closure`) or just report them (for `drop_closure`). However, both functions drop the payload returned by `catch_unwind`, which is an operation that can panic. Formally, unwinding from `extern "C"` functions [is undefined behavior](https://doc.rust-lang.org/1.62.0/reference/behavior-considered-undefined.html), but in practice, the program will effectively `longjmp` out of any C code as it unwinds. While this appears not to cause obvious memory unsafety in CPython, it can cause reference leaks and resource leaks. In particular, unwinding from `run_closure` prevents the recursion depth from being decremented, and unwinding from `drop_closure` prevents PyO3 from ever releasing the GIL and also prevents the trashcan mechanism from freeing any objects.

### Steps to Reproduce

An example of a recursion depth leak from `run_closure`:
```rust
/*
[dependencies]
pyo3 = { path = "pyo3", default-features = false }
*/

use pyo3::{types::PyCFunction, Python};
use std::panic::{self, AssertUnwindSafe};

struct Bomb;

impl Drop for Bomb {
    fn drop(&mut self) {
        panic!();
    }
}

fn main() {
    Python::with_gil(|py| {
        let f = PyCFunction::new_closure(|_, _| panic::panic_any(Bomb), py).unwrap();
        panic::set_hook(Box::new(|_| {}));
        for _ in 0..996 {
            panic::catch_unwind(AssertUnwindSafe(|| f.call0().unwrap())).unwrap_err();
        }
        let _ = panic::take_hook();
        py.eval("print('Hello, world!')", None, None).unwrap();
    });
}
```
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'RecursionError'>, value: RecursionError('maximum recursion depth exceeded while calling a Python object'), traceback: Some(<traceback object at 0x7f389f58d940>) }', src/main.rs:38:55
```

An example of a GIL leak from `drop_closure`:
```rust
/*
[dependencies]
pyo3 = { path = "pyo3", default-features = false }
pyo3-ffi = { path = "pyo3/pyo3-ffi" }
*/

use pyo3::{types::PyCFunction, Python};
use std::panic;

struct Bomb;

impl Drop for Bomb {
    fn drop(&mut self) {
        panic!();
    }
}

struct Fuse;

impl Drop for Fuse {
    fn drop(&mut self) {
        panic::panic_any(Bomb);
    }
}

fn main() {
    panic::set_hook(Box::new(|_| {}));
    let fuse = Fuse;
    panic::catch_unwind(|| {
        Python::with_gil(|py| {
            PyCFunction::new_closure(
                move |_, _| {
                    let _ = &fuse;
                },
                py,
            )
            .unwrap();
        })
    })
    .unwrap_err();
    let _ = panic::take_hook();
    let state = unsafe { pyo3_ffi::PyGILState_Check() };
    println!("GIL state: {state}");
}
```
```
--- PyO3 intercepted a panic when dropping a closure
Any { .. }
GIL state: 1
```

### Backtrace

_No response_

Ubuntu 22.04 LTS

Python 3.10.4

rustc 1.62.0 (a8314ef7d 2022-06-27)

### Your PyO3 version

commit a95485c

apt

### Additional Info

The error message in `drop_closure` can only ever print `Any { .. }`. To print the actual payload, it must be explicitly converted to a string with `.downcast_ref::<&str>()` and `.downcast_ref::<String>()`.

davidhewitt:
Fixes #2501

### DIFF
--- a/src/panic.rs
+++ b/src/panic.rs
@@ -16,7 +16,9 @@ Python interpreter to exit.
 );
 
 impl PanicException {
-    // Try to format the error in the same way panic does
+    /// Creates a new PanicException from a panic payload.
+    ///
+    /// Attempts to format the error in the same way panic does.
     #[cold]
     pub(crate) fn from_panic_payload(payload: Box<dyn Any + Send + 'static>) -> PyErr {
         if let Some(string) = payload.downcast_ref::<String>() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `pyo3_build_config::add_extension_module_link_args()` now also emits linker arguments for `wasm32-unknown-emscripten`. [#2538](https://github.com/PyO3/pyo3/pull/2538)
 - Downcasting (`PyTryFrom`) behavior has changed for `PySequence` and `PyMapping`: classes are now required to inherit from (or register with) the corresponding Python standard library abstract base class. See the [migration guide](https://pyo3.rs/latest/migration.html) for information on fixing broken downcasts. [#2477](https://github.com/PyO3/pyo3/pull/2477)
 - Disable `PyFunction` on `Py_LIMITED_API` and PyPy. [#2542](https://github.com/PyO3/pyo3/pull/2542)
+- Panics during drop of panic payload caught by PyO3 will now abort. [#2544](https://github.com/PyO3/pyo3/pull/2544)
 
 ### Removed
 
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -339,19 +339,20 @@ fn impl_traverse_slot(cls: &syn::Type, spec: FnSpec<'_>) -> MethodAndSlotDef {
             arg: *mut ::std::os::raw::c_void,
         ) -> ::std::os::raw::c_int
         {
+            let trap = _pyo3::impl_::panic::PanicTrap::new("uncaught panic inside __traverse__ handler");
             let pool = _pyo3::GILPool::new();
             let py = pool.python();
-            _pyo3::callback::abort_on_traverse_panic(::std::panic::catch_unwind(move || {
-                let slf = py.from_borrowed_ptr::<_pyo3::PyCell<#cls>>(slf);
+            let slf = py.from_borrowed_ptr::<_pyo3::PyCell<#cls>>(slf);
 
-                let visit = _pyo3::class::gc::PyVisit::from_raw(visit, arg, py);
-                let borrow = slf.try_borrow();
-                if let ::std::result::Result::Ok(borrow) = borrow {
-                    _pyo3::impl_::pymethods::unwrap_traverse_result(borrow.#ident(visit))
-                } else {
-                    0
-                }
-            }))
+            let visit = _pyo3::class::gc::PyVisit::from_raw(visit, arg, py);
+            let borrow = slf.try_borrow();
+            let retval = if let ::std::result::Result::Ok(borrow) = borrow {
+                _pyo3::impl_::pymethods::unwrap_traverse_result(borrow.#ident(visit))
+            } else {
+                0
+            };
+            trap.disarm();
+            retval
         }
     };
     let slot_def = quote! {
--- a/src/callback.rs
+++ b/src/callback.rs
@@ -5,6 +5,7 @@
 use crate::err::{PyErr, PyResult};
 use crate::exceptions::PyOverflowError;
 use crate::ffi::{self, Py_hash_t};
+use crate::impl_::panic::PanicTrap;
 use crate::panic::PanicException;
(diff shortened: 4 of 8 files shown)


src/panic.rs
```diff
@@ -16,15 +17,46 @@ Python interpreter to exit.
...
+    };
+    abort_on_unwind(move || drop(payload), "panic while dropping panic payload");
+    msg
+}
+
+/// Executes `f()`. If it panics, attempts to write `msg` to stderr and then aborts.
+pub(crate) fn abort_on_unwind<F: FnOnce()>(f: F, msg: &str) {
+    // Always fine to use AssertUnwindSafe here because a panic will abort.
+    std::panic::catch_unwind(std::panic::AssertUnwindSafe(f))
+        .unwrap_or_else(|_| abort_with_msg(msg));
```
adamreichold, line 55: Since `unwrap_or_else` is called for the closure's side effect, I think it would be clearer to match this like

```rust
if let Err(_err) = std::panic::catch_unwind(std::panic::AssertUnwindSafe(f)) {
  abort_with_msg(msg);
}
```

I would also feel better if the binding was `_err` instead of `_` to ensure the panic payload (the one caused by `drop(payload)`) is not dropped before calling `abort_with_msg`.

src/panic.rs
```diff
@@ -16,15 +17,46 @@ Python interpreter to exit.
...
+}
+
+/// Attempts to write `msg` to stderr, then aborts.
+pub(crate) fn abort_with_msg(msg: &str) -> ! {
+    let _ = writeln!(std::io::stderr(), "FATAL: {}, aborting", msg);
```
adamreichold, line 60: I think the `Stderr` wrapper returned by `stderr` used a `StderrLock` internally, the [`Write` impl of which calls `borrow_mut`](https://doc.rust-lang.org/stable/std/io/struct.StderrLock.html#impl-Write) so it seems this could panic itself. (The comment removed in `src/types/function.rs` seems to indicate this as well.)

Could this be a problem here since we are already unwinding? Is there a way to portably avoid calling into the whole `Stdio` locking machinery and the process go down my reliably but still make a best effort to print the message?

Is there

adamreichold: Since we already depend on `libc`, would calling `write` with `STDERR_FILENO` be portable? Or is this just being paranoid and overdoing it?

src/panic.rs
adamreichold: (Just as an idea, could we put a type on the stack which panics on `Drop` itself to trigger abort due to a panic during unwind before calling `f` instead of using `catch_unwind` and having to juggle its potential (second order) panic bomp? Maybe already in `panic_payload_to_string`?)

davidhewitt: 👍 this is a great idea! Will change to implement as this.

mejrs: I would prefer an aborting guard as well.

src/panic.rs
adamreichold: I looked into how std does this and they have a `rtabort` macro which calls into `rtprintpanic` which uses the normal `format!` machinery

```rust
macro_rules! rtprintpanic {
    ($($t:tt)*) => {
        if let Some(mut out) = crate::sys::stdio::panic_output() {
            let _ = crate::io::Write::write_fmt(&mut out, format_args!($($t)*));
        }
    }
}
```

but writes into a special `panic_output` which for example on Unix-like platforms is

```rust
impl Stderr {
    pub const fn new() -> Stderr {
        Stderr(())
    }
}

impl io::Write for Stderr {
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        unsafe { ManuallyDrop::new(FileDesc::from_raw_fd(libc::STDERR_FILENO)).write(buf) }
    }
}
```

(It is a bit more complicated on Windows since it handles UTF-8 to UTF-16 conversion, but since our messages are always plain ASCII, I think the Unix approach should generally work for us.)

adamreichold: Actually, I think we can let std handle all of this. My initial idea above was to use abort-due-to-panic-during-unwind, i.e. `PanicTrap::drop` panics instead of calling `abort_with_msg` which means std will take care of writing both panic messages and a backtrace without use having to muck around in the low-level details, c.f. [Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b75409ec1b4117f27e7da49c2451d411).

davidhewitt: This sounds like a very sensible suggestion, I'll push an amendment shortly!

adamreichold (approved):

adamreichold:
The jobs are queued for an unusual amount of time and I got some API errors while using the UI. Seems like GitHub has a hiccup...

davidhewitt:
Heh, I'd pushed a broken commit anyway. Hopefully this one passes CI...

PR #2549, opened 2022-08-13
Deprecate acquire_gil

mejrs:
Per https://github.com/PyO3/pyo3/pull/1788#issuecomment-1198244949

### DIFF
--- a/tests/test_compile_error.rs
+++ b/tests/test_compile_error.rs
@@ -52,6 +52,7 @@ fn _test_compile_errors() {
     tests_rust_1_58(&t);
     tests_rust_1_60(&t);
     tests_rust_1_62(&t);
+    tests_rust_1_63(&t);
 
     #[rustversion::since(1.49)]
     fn tests_rust_1_49(t: &trybuild::TestCases) {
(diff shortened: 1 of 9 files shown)


src/marker.rs
```diff
@@ -382,6 +382,7 @@ impl<'py> Python<'py> {
     /// [`PyGILState_Ensure`]: crate::ffi::PyGILState_Ensure
     /// [`auto-initialize`]: https://pyo3.rs/main/features.html#auto-initialize
     #[inline]
+    #[deprecated(since = "0.17.0", note = "prefer Python::with_gil")]
     pub fn acquire_gil() -> GILGuard {
         GILGuard::acquire()
```
davidhewitt, line 387: Is it worth adding a comment here to remind us as maintainers that when we remove this API for good we can also remove the drop-order tracking from `GILGuard`?

davidhewitt (commented):
Thanks for keeping this thread moving forward! Just a couple of tiny thoughts and then we can ship this as part of 0.17 :)

mejrs:
I don't know how I messed this up, but I'll just open a new PR

Issue #906, opened 2020-05-06
Add way to set metaclass for #[pyclass]


programmerjake:
apparently it works by setting the `ob_type` member to a custom instance (the metaclass) of a class that derives from `type`.

lausek:
Has there been progress on this issue? I'm looking for a way of implementing [\_\_instancecheck\_\_](https://docs.python.org/3/reference/datamodel.html#class.__instancecheck__) for a custom rust class.

lausek:
Has there been progress on this issue? I'm looking for a way of implementing a custom [instancecheck](https://docs.python.org/3/reference/datamodel.html#class.__instancecheck__) for a `pyclass`.

davidhewitt:
Any progress would have been on this issue! A proposal of how the API should look and / or a PR to implement this is always welcome 😄

b05902132:
Someone mentioned this issue on Gitter. https://bugs.python.org/issue15870

It seems that metaclass is not officially supported in C extensions.

mbway:
Since python 3.12 there is a function: `PyType_FromMetaclass` ([docs](https://docs.python.org/3/c-api/type.html#c.PyType_FromMetaclass)) which allows metaclasses to be set in extension modules. Below I have a c example (with error handling and memory cleanup removed for brevity) where I create `MyMetaclass` with a `__getitem__` equivalent and create a class `MyClass` with `MyMetaclass` as its metaclass. You can then do `MyClass[123]` and have it call `__getitem__` of the metaclass.

This could be used to implement more features that `EnumType` provides (as requested here: https://github.com/PyO3/pyo3/issues/2887).

In addition to the official way of doing things, I found that I was able to set the metaclass of a class by assigning to `ob_type` after calling `PyTypeReady` as suggested [here](https://stackoverflow.com/a/52957978) but that does seem to be a hack and might cause issues.

```c
#define PY_SSIZE_T_CLEAN
#include <Python.h>

typedef struct { PyTypeObject base_type; } MyMetaclass;

static PyMethodDef MyMetaclass_methods[] = { {NULL} };

static PyObject *myclass_getitem(PyObject *self, PyObject *key) { return PyLong_FromLong(123); }

static PyMappingMethods MyMetaClassMappingMethods = {
    .mp_length = NULL,
    .mp_subscript = myclass_getitem,
    .mp_ass_subscript = NULL,
};

static PyTypeObject MyMetaclassType = {
[...]

PR #2553, opened 2022-08-14
Created a builder for creating types, allowing for more flexibility

vultix:
This is the first in a series of pull requests as I implement issues #417 and #906.  This seemed like a relatively easy piece to review separately.

Creating metaclasses and custom types for different enum variants required more flexibility than our current `create_type_object_impl` allowed for.  I originally had lots of near-duplicates of the function, but found a builder like what I have here covers my use cases better.

### Question regarding unsafe
I'm not entirely sure which methods should / shouldn't be considered unsafe to use.  The approach I've taken here is to assume pushing a c pointer to a slot is always unsafe, but that functions doing so are considered safe if we've verified the slot type is correct.  For example, is `PyTypeBuilder::build` always safe to call if the earlier safety invariants are met? I believe so, but am not positive.

### DIFF
--- a/src/pyclass.rs
+++ b/src/pyclass.rs
@@ -298,7 +298,7 @@ impl PyTypeBuilder {
             }
 
             // Safety: Py_tp_members expects a raw vec of PyMemberDef
-            unsafe { self.push_raw_vec(ffi::Py_tp_members, members) };
+            unsafe { self.push_raw_vec_slot(ffi::Py_tp_members, members) };
         }
 
         // Setting buffer protocols, tp_dictoffset and tp_weaklistoffset via slots doesn't work until
--- a/src/pyclass.rs
+++ b/src/pyclass.rs
@@ -11,9 +11,10 @@ use crate::{
 };
 use std::{
     cmp::Ordering,
+    collections::HashMap,
     convert::TryInto,
     ffi::{CStr, CString},
-    os::raw::{c_char, c_int, c_uint, c_void},
+    os::raw::{c_char, c_int, c_ulong, c_void},
     ptr,
 };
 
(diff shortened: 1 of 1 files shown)


mejrs:
>  The approach I've taken here is to assume pushing a c pointer to a slot is always unsafe

In general, a function taking a raw pointer as an argument has to be unsafe, because it has no way to check that the pointer is correct.

src/pyclass.rs
```diff
@@ -154,122 +190,197 @@ unsafe fn create_type_object_impl(
...
     }
-}
 
-#[cold]
-fn type_object_creation_failed(py: Python<'_>, e: PyErr, name: &'static str) -> ! {
-    e.print(py);
-    panic!("An error occurred while initializing class {}", name)
-}
+    fn with_type_doc(mut self, type_doc: &'static str) -> Self {
+        if let Some(doc) = py_class_doc(type_doc) {
```
mejrs, line 241: Because you're taking a `&'static str`, this means `py_class_doc` shouldn't have to leak that CString.

src/pyclass.rs
```diff
@@ -30,122 +31,157 @@ pub trait PyClass:
...
     is_mapping: bool,
-) -> PyResult<*mut ffi::PyTypeObject> {
-    let mut slots = Vec::new();
+    has_new: bool,
+    has_dealloc: bool,
+    has_getitem: bool,
+    has_setitem: bool,
+    has_traverse: bool,
+    has_clear: bool,
+    has_dict: bool,
```
mejrs, line 70: When you have a lot of booleans it's often better to use enums. Something like:

```rust
enum Traverse{
    YES,
    NO,
}
```

src/pyclass.rs
```diff
@@ -154,122 +190,197 @@ unsafe fn create_type_object_impl(
...
-        if let Some(dict_offset) = dict_offset {
-            (*_type_object).tp_dictoffset = dict_offset;
+        if !self.has_new {
+            // Safety: This is the correct slot type for Py_tp_new
+            unsafe { self.push_slot(ffi::Py_tp_new, no_constructor_defined as *mut c_void) }
         }
 
-        if let Some(weaklist_offset) = weaklist_offset {
-            (*_type_object).tp_weaklistoffset = weaklist_offset;
+        if !self.has_dealloc {
```
mejrs, line 338: Just a thought: in principle the typestate pattern could move this (and most other checks in this file) from runtime to compile time. I don't think it'd be worth it though.

mejrs (commented):
I'm really happy to have a builder for this. It has gotten quite messy so thanks for the cleanup :)

src/pyclass.rs
davidhewitt: Given these are internal state of the builder, I think it's fine. Enums definitely improve readability in function calls, don't think that's applicable here though?

src/pyclass.rs
davidhewitt: Given that this code is expected to only execute once per pyclass I agree that runtime checks are preferred here over getting the compiler to do lots of codegen / compile-time checks.

src/pyclass.rs
davidhewitt: Is that true? If the `str` isn't null-terminated the copy will need to be made.

davidhewitt (commented):
Thanks very much for working on the enums and for kicking off with a much-needed refactor!

Using a builder is a great idea. For a bit of context why we had `create_type_object` and `create_type_object_impl`, the motivation was to keep the codgen cost of this code as low as possible by having minimal generic wrapper `create_type_object` around concrete `create_type_object_impl`. We had some severe compile-time regressions reported in the past by users in this code.

I think with a builder we could actually cut the codegen cost further. I wonder if we could have a trait like the following, shifting the builder creation to const eval and meaning we only need concrete runtime code to be generated:

```rust
trait HasPyClassBuilder {
    const BUILDER: PyClassBuilder;
}
``` 

I think it should be possible to give `HasPyClassBuilder` a blanket impl for `T: PyClassImpl`?

src/pyclass.rs
mejrs: Yeah I think it's fine either way.

src/pyclass.rs
mejrs: It usually will be null terminated. if it's not, it will still need to copy+leak that cstring yes.

adamreichold: > It usually will be null terminated.

So could we enforce this invariant by taking `&'static CStr`?

davidhewitt: I've wanted to do this in the past but I think const-constructed `CStr` isn't possible on MSRV. I guess could leak earlier for older compilers.

src/pyclass.rs
vultix: Yeah, if these were mutually exclusive in any way I'd make an enum variant for them, but as it is we'd just need to make 7 enum variants to represent true/false, and isn't worth the added effort.

src/pyclass.rs
vultix: If we enforce the `type_doc` is always a null-terminated `&'static str` I believe we could cast it directly into a pointer without leaking the memory. We could easily check that invariant at runtime to keep this function safe.

vultix: For now, I'd like this PR to keep the behavior exactly the same as before, so I won't make this change, but I will in a followup PR.

vultix (commented):
Alright, cleaned up that last clippy warning, and all other comments don't seem to be actionable.  Let me know the next steps from here!

vultix:
> I think it should be possible to give `HasPyClassBuilder` a blanket impl for `T: PyClassImpl`?

The builder currently allocates a Vec to store all of the slots, so can't be const.

src/pyclass.rs
```diff
@@ -30,122 +31,159 @@ pub trait PyClass:
...
-            T::IS_MAPPING,
-        )
+        PyTypeBuilder::default()
+            .with_type_doc(T::DOC)
+            .with_offsets(T::dict_offset(), T::weaklist_offset())
+            .with_slot(ffi::Py_tp_base, T::BaseType::type_object_raw(py))
+            .with_slot(ffi::Py_tp_dealloc, tp_dealloc::<T> as *mut c_void)
+            .set_is_basetype(T::IS_BASETYPE)
+            .set_is_mapping(T::IS_MAPPING)
+            .with_class_items(T::items_iter())
```
messense, line 46: I was under the impression that in Rust usually builder pattern naming conversions are:

* `with_xxx(val: T) -> Self` is an associate method for constructing the builder, for example `fn with_type_doc(doc: &'static str)`
* `xxx(mut self, val: T) -> Self` is a setter, no need to use `set_` prefix.

What do you think?

davidhewitt:
> The builder currently allocates a Vec to store all of the slots, so can't be const.

Ok fair, instead I pushed https://github.com/vultix/pyo3/pull/1 which captures all of the arguments we feed to the builder as a constant in the same fashion.

I checked `cargo llvm-lines` and `create_type_object` goes from 44 to 175 lines _per pyclass_ with this PR as currently implemented. If you accept my patch on top then the cost per pyclass is zero and my concern is resolved :)

src/pyclass.rs
davidhewitt: Following the example of `std::process::command`, I think dropping `with_` and `set_` prefixes from all methods would be conformant 👍

davidhewitt:
As discussed in https://github.com/vultix/pyo3/pull/1 we've shown that there _is_ a compile performance regression, but it doesn't appear to be so significant that it should be a blocker for merging.

So I think https://github.com/PyO3/pyo3/pull/2553#discussion_r946425058 is the last tweak proposed here, and then let's merge this.

src/pyclass.rs
vultix: Removing the `set_` from `set_is_basetype` leads to this clippy warning:
```
warning: methods called `is_*` usually take `self` by mutable reference or `self` by reference or no `self`
   --> src/pyclass.rs:223:19
    |
223 |     fn is_mapping(mut self, is_mapping: bool) -> Self {
    |                   ^^^^^^^^
    |
    = help: consider choosing a less ambiguous name
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
```

Unless there's an alternative, I'll leave those two as they are, but update the rest.

davidhewitt: Sounds reasonable to me 👍

vultix:
I'm not sure why CI is failing now - is this on my end?

mejrs:
No, it isnt anything you did - Bumpalo bumped its msrv, so we need to use an older version.

davidhewitt (approved):
I'll rebase this on main; that should then go through. Thanks!

davidhewitt:
EDIT: looks like you didn't check the box to give me permission to push to your branch - you'll need to do it :)

davidhewitt:
Actually I just pushed the rebase to my own fork, opened as #2565; will close this one. Thanks!

PR #2562, opened 2022-08-17
guide: deployment to netlify

davidhewitt:
See outputs at:
- https://netlify-test--pyo3.netlify.app/ - latest docs (also https://netlify-test--pyo3.netlify.app/latest/)
- https://netlify-test--pyo3.netlify.app/main - main branch (also https://netlify-test--pyo3.netlify.app/main/doc/pyo3)
- https://netlify-test--pyo3.netlify.app/v0.15.0 etc for old versioned docs

The main docs are built and hosted on netlify, the releases are stored on gh-pages and the docs on docs.rs. They are proxied through netlify redirects.

If this is setup we're happy with, I'll reach out to netlify to apply for the open-source plan and we can commit to this. (I've also reached out to @fafhrd91 to get access to the DNS settings so I could move pyo3.rs to point at this.)

When merged, the prefix `netlify-test` will drop and we'll just have https://pyo3.netlify.app working.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -39,23 +39,6 @@ jobs:
           TAG_NAME="${GITHUB_REF##*/}"
           echo "::set-output name=tag_name::${TAG_NAME}"
 
-      # Build some internal docs and inject a banner on top of it.
-      - name: Build the internal docs
-        run: |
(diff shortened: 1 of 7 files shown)


messense (approved):

davidhewitt:
Also note that this stops deploying the internal docs to `gh-pages`, which should reduce the repo size growth (and we can improve further once we move this to become the only host of the `main` docs).

PR #2563, opened 2022-08-18
misc package updates

davidhewitt:
- MSRV: pin bumpalo
- examples: update to maturin 0.13


adamreichold (approved):

adamreichold:
Just to make sure: `anyhow` was the Cargo bug which unintentionally bumped MSRV and that version has been yanked which is why it is removed here.

messense (approved):

messense:
> Just to make sure: `anyhow` was the Cargo bug which unintentionally bumped MSRV and that version has been yanked which is why it is removed here.

Yes, https://github.com/dtolnay/anyhow/issues/253

Issue #2392, opened 2022-05-20
PyClass Sequence of tuples or arrays not convertible to numpy array


mtreinish:
### Bug Description

When creating a custom pyclass that implements the sequence protocol with pyo3 0.16.x the output type is missing something (I haven't been able to figure out exactly what) that prevents numpy from interpreting it correctly to convert into an array. This only is an issue when using the creating a sequence type using `#[pymethods]`. It works as expected if you use the deprecated `#[pyproto]` with the `PySequenceProtocol` trait.

### Steps to Reproduce

1. create pyo3 pyclass that implement sequence protocol on array or tuple of numeric types, minimal example:

```rust
use pyo3::prelude::*;
use pyo3::exceptions::PyIndexError;

#[pyclass(module="reproduce")]
#[derive(Clone)]
pub struct MySequenceArray {
    pub data: Vec<[usize; 2]>
}

#[pymethods]
impl MySequenceArray {
    #[new]
    fn new(data: Vec<[usize; 2]>) -> Self {
        MySequenceArray {
            data
        }
    }

    fn __len__(&self) -> PyResult<usize> {
        Ok(self.data.len())
    }

    fn __getitem__(&self, idx: isize) -> PyResult<[usize; 2]> {
        if idx as usize >= self.data.len() {
            Err(PyIndexError::new_err("out of bounds"))
        } else {
            Ok(self.data[idx as usize])
        } 
    }
}

#[pyclass(module="reproduce")]
#[derive(Clone)]
pub struct MySequenceTuple {
    pub data: Vec<(usize, usize)>
}

#[pymethods]
impl MySequenceTuple {
    #[new]
    fn new(data: Vec<(usize, usize)>) -> Self {
        MySequenceTuple {
            data
        }
    }

    fn __len__(&self) -> PyResult<usize> {
        Ok(self.data.len())
    }

    fn __getitem__(&self, idx: isize) -> PyResult<(usize, usize)> {
        if idx as usize >= self.data.len() {
            Err(PyIndexError::new_err("out of bounds"))
        } else {
            Ok(self.data[idx as usize])
        }
    }
}

#[pymodule]
fn reproduce(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
    m.add_class::<MySequenceTuple>()?;
    m.add_class::<MySequenceArray>()?;
    Ok(())
}
```

2. In python create instances of `MySequenceArray` or `MySequenceTuple`:

```python
import reproduce
[...]

PR #2567, opened 2022-08-19
pyclass: add `sequence` option to implement `sq_length`

davidhewitt:
Implements `#[pyclass(sequence)]` as discussed at https://github.com/PyO3/pyo3/issues/2392#issuecomment-1146697738

This is the last thing I think needs to get merged before we are ready to push the releases!


messense (approved):
👍

PR #2578, opened 2022-08-21
Fix pyproto documentation.

mejrs:
Since https://github.com/PyO3/pyo3/pull/2322 this is not enabled by default.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -87,8 +87,7 @@
 //! - `multiple-pymethods`: Enables the use of multiple [`#[pymethods]`](macro@crate::pymethods)
 //! blocks per [`#[pyclass]`](macro@crate::pyclass). This adds a dependency on the [inventory]
 //! crate, which is not supported on all platforms.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -89,6 +87,8 @@
 //! - `multiple-pymethods`: Enables the use of multiple [`#[pymethods]`](macro@crate::pymethods)
 //! blocks per [`#[pyclass]`](macro@crate::pyclass). This adds a dependency on the [inventory]
 //! crate, which is not supported on all platforms.
+//! - `pyproto`: Enables the deprecated `#[pyproto]` attribute macro. This will be removed in a 
+//! future release.
```
mejrs, line 91: Do we want to mention a concrete version? 0.18 perhaps?

davidhewitt: I think it's fine to say 0.18 here 👍 (I'm already planning to have a bonfire PR as soon as 0.17 is out the door xD)

davidhewitt (commented):
Thanks!

davidhewitt (approved):

PR #2580, opened 2022-08-21
docs: pile of tweaks and additions

### LINKED ISSUE #804: Wasm-bindgen-like Closure wrapper around Rust functions?

### LINKED ISSUE #1602: impl WrappingCastTo<isize><HashCallbackOutput> for u128
## 🐛 Bug Reports

<warning - Rust newbie>
Hi pyo3 community! Love your library!

Basically, I would like to export the `__hash__` implementation from rust to python and to return an u128 hash out of it.

A snippet to reproduce the bug:

```
[pyproto]
impl PyObjectProtocol for MyStruct {
    fn __hash__(&self) -> u128 {
         0 as u128
    }

```
 
But it looks like the trait to do this is not implemented for u128. I could take a stab to implement it if it's something not too hardcore in the internals of the library. 

### 🌍 Environment

 - Your operating system and version: Ubuntu 20:04
 - Your python version: 3.7
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: conda env
 - Your Rust version (`rustc --version`): rustc 1.51.0 (2fd73fabe 2021-03-23)
 - Your PyO3 version: 0.13.2

### 💥 Reproducing

Try to return an u128 from a `__hash__` implementation.

### LINKED ISSUE #2122: Implementing `__richcmp__` makes object unhashable
### Bug Description

If you try to call Python's hash function on an object for which you've implemented `__richcmp__`, it fails with `unhashable type`. The only fix I've found is to also implement `__hash__`.

This also happens if you implement `__eq__` though I've heard `__eq__` isn't supported anyway.

### Steps to Reproduce

```rust
use pyo3::{basic::CompareOp, prelude::*};

#[pyclass]
#[derive(PartialEq, PartialOrd, Debug, Default)]
pub struct DoesNotHash(u32);

#[pymethods]
impl DoesNotHash {
    #[new]
    pub fn new() -> Self {
        Self(3)
    }
    
    // Uncomment the below to break hashing.
    // fn __richcmp__(&self, other: &Self, op: CompareOp) -> bool {
    //     match op {
    //         CompareOp::Eq => self == other,
    //         CompareOp::Ne => self != other,
    //         CompareOp::Lt => self < other,
    //         CompareOp::Gt => self > other,
    //         CompareOp::Le => self <= other,
    //         CompareOp::Ge => self >= other,
    //     }
    // }
}

/// A Python module implemented in Rust.
#[pymodule]
fn hash_example(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<DoesNotHash>()?;
    Ok(())
}
```

### Backtrace

_No response_

Ubuntu 20.04.3 LTS

Python 3.9

rustc 1.59.0-nightly (7abab1efb 2021-12-17)

### Your PyO3 version

0.15.1

Installed in fresh conda environment.

### Additional Info

_No response_

### LINKED ISSUE #2495: Thread state is lost between calls to `Python::with_gil`
### Bug Description

(Stemming from https://github.com/microsoft/debugpy/issues/968)

I've tracked down my debugability-of-python issue to the fact that the thread state is not persisted in a thread between multiple calls to `Python::with_gil`. The call to `PyGILState_Release` at the end of the `GILGuard` hits https://github.com/python/cpython/blob/45e62a2bc1c0000e2e9b613fff6bebf2c26fcb93/Python/pystate.c#L1715 which then clears and deletes the thread-state.

I'm still working on understanding Python's thread behavior, but ideally the thread state persists as long as the thread does. Not sure the right CPython calls to make this happen.

### Steps to Reproduce

1. Python calling a Python Rust-implemented C Extension module using PyO3
2. In the module, spawn a new thread
3. In the new thread calling `sys.settrace` and then `sys.gettrace` in two calls to `Python::with_gil` will show that thread state was cleared in between

### Backtrace

_No response_

Ubuntu 20

Python 3.9

rustc 1.61.0

### Your PyO3 version

0.16.5

apt

### Additional Info

_No response_

### LINKED ISSUE #2546: Please put in the documentation the common workarounds for the orphan rule
I'm new at rust, so this probably seems like a noob thing.

I believe rust has a 'easy', if a bit wordy solution for library errors in the form of the 'From' trait, which can be implemented if it's going to be implemented for your own errors that then makes ? transparently usable. Easy enough to turn a error from a library into your own error.

... however, pyo3 requires errors to be returned in PyResult<T> where the error is determined to be PyErr, which can't be changed afaik, so you hit the orphan rules by trying to implement something to a foreign type. So you have here some trouble of 'connecting' both parts.

tl;dr: for noobs like me, i feel you should mention the result method `.map_err` as a workaround to specifically the situation where you need to convert a series of library errors into a python exception (PyErr) to be returned into python.

And if you want to be really clear to the not so gifted, like me, mention that you _don't_ need a intermediate rust type to 'coalesce' the errors but can just return PyErr with the python exceptions you created with the macros in the function to pass to map_err which in turn the ? operator will return to python fine. And that often when you do this if the value was to be immediately returned, not wrapped or used to calculate something else, you can remove the Ok(...?) around it, although clippy will warn about it.

Just a small paragraph in the exceptions part of the documentation could save some aggregate man hours over the years.

### LINKED ISSUE #2547: Please put in the documentation for __richcmp__ the alternate implementation that allows using NotImplemented()
It's very common to only want to implement __richcmp__ partially, so although the current example is good, you could also mention the 'partial' much more complicated to find alternative that has access to NotImplemented() since it requires a python interperter reference:

```
    fn __richcmp__(&self, other: PyRef<Event>, op: CompareOp) -> Py<PyAny> {
        let py = other.py();
        match op {
          CompareOp::Eq => (self.0 == other.0).into_py(py),
          CompareOp::Ne => (self.0 != other.0).into_py(py),
          _ => py.NotImplemented(),
        }
    }
```

And you could maybe even mention in the original that the result return type is optional, just in case the type you're implementing can send exceptions from the operations (i guess) and just `-> bool` can work if you're sure your type does no such thing.

davidhewitt:
I was stuck on the train this afternoon with barely a working internet connection, so what better thing to do than close a whole pile of docs issues 😄 

I tried to land everything (mostly) as separate commits. Changes included:
- Change page titles and navigation links from _Title Case_ to _Sentence case_
- Remove `::pyo3` from guide samples (given that's not _quite_ what the macro expands to anyway!)
- Closes #2547 
- Closes #2546 
- Closes #2495 
- Closes #2122
- Closes #1602
- Closes #804

All reviews welcome, I'll probably merge this on Tuesday if I haven't heard anything so that it can be part of the published 0.17 guide. Figure even if it's not perfect it's probably better than not including in the release :)


messense:
[Rendered](https://deploy-preview-2580--pyo3.netlify.app/main/)

messense (approved):
Looks great!

guide/src/function.md
```diff
@@ -234,13 +211,12 @@ with only positional args.
...
-to get the corresponding [`PyCFunction`]. For dynamic functions, e.g. lambdas and functions that
-were passed as arguments, you must put them in some kind of owned container, e.g. a `Box`.
-(A long-term solution will be a special container similar to wasm-bindgen's `Closure`). You can
-then use a `#[pyclass]` struct with that container as a field as a way to pass the function over
-the FFI barrier. You can even make that class callable with `__call__` so it looks like a function
-in Python code.
+The ways to convert a Rust function into a Python object vary depending on the function:
+
+- Concrete functions, e.g. `fn foo()`: add `#[pyfunction]` and then use [`wrap_pyfunction!`] to get the corresponding [`PyCFunction`].
+- Dynamic functions, e.g. `foo: fn()` either:
```
mejrs, line 217: What about "named" and "anonymous/closures" versus "concrete" and "dynamic"? I don't think many people are familiar with those terms.

guide/src/function/error_handling.md
```diff
@@ -0,0 +1,222 @@
...
+#             .is_instance_of::<PyValueError>(py));
+#     })
+# }
+```
+
+## Foreign Rust error types
+
+The Rust compiler will not permit implementation of traits for types outside of the crate where the type is defined. (This is known as the "orphan rule".)
+
+Given a type `OtherError` which is defined in thirdparty code, there are two main strategies available to integrate it with PyO3:
```
mejrs, line 172: This should be `third-party` or `third party`? (same for the other usages). I'm not a native english speaker so I'm not sure 😺

mejrs (commented):
Thanks, this is great!

guide/src/function/error_handling.md
davidhewitt: Good catch! I think `third-party` is the adjective form and `third party` the noun from looking at https://dictionary.cambridge.org/dictionary/english/third-party, so I will use `third-party` here.

guide/src/function.md
davidhewitt: Much better

PR #2583, opened 2022-08-22
Improve ffi example and make const PyMethodDef::zeroed

mejrs:
Rendered:
image
image
image

### DIFF
--- a/pyo3-ffi/src/lib.rs
+++ b/pyo3-ffi/src/lib.rs
@@ -88,64 +88,40 @@
 //!
 //! use pyo3_ffi::*;
 //!
+//! static mut MODULE_DEF: PyModuleDef = PyModuleDef {
+//!     m_base: PyModuleDef_HEAD_INIT,
+//!     m_name: "string_sum\0".as_ptr().cast::<c_char>(),
+//!     m_doc: "A Python module written in Rust.\0"
+//!         .as_ptr()
+//!         .cast::<c_char>(),
+//!     m_size: 0,
+//!     m_methods: unsafe { METHODS.as_mut_ptr().cast() },
+//!     m_slots: std::ptr::null_mut(),
+//!     m_traverse: None,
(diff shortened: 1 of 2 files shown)


pyo3-ffi/src/methodobject.rs
```diff
@@ -94,6 +94,19 @@ pub struct PyMethodDef {
     pub ml_doc: *const c_char,
 }
 
+impl PyMethodDef {
+    pub const fn zeroed() -> PyMethodDef {
```
mejrs, line 98: I found it quite annoying to make a terminating PyMethodDef in a static, this would make it easier. I'm all for a better idea :)

davidhewitt: Yeah seems ok to me, it would be nice if `std::mem::zeroed` was `const`, but here we are 🙂

pyo3-ffi/src/lib.rs
```diff
@@ -88,64 +88,40 @@
...
-//!     let wrapped_sum_as_string = PyMethodDef {
-//!         ml_name: "sum_as_string\0".as_ptr() as *const c_char,
+//! static mut MODULE_DEF: PyModuleDef = PyModuleDef {
+//!     m_base: PyModuleDef_HEAD_INIT,
+//!     m_name: "string_sum\0".as_ptr().cast::<c_char>(),
+//!     m_doc: "A Python module written in Rust.\0"
+//!         .as_ptr()
+//!         .cast::<c_char>(),
+//!     m_size: 0,
+//!     m_methods: unsafe { METHODS.as_mut_ptr().cast() },
```
davidhewitt, line 98: 👍 I'm hoping that #2367 can enable us to also use this member in `#[pymodule]`.

davidhewitt (approved):
Thanks, definitely makes the `ffi` calls clearer and emphasises how much PyO3 can help 😄

PR #2589, opened 2022-08-24
netlify: fix redirects

davidhewitt:
I updated `create_redirects.py` previously but forgot to update the `_redirects` file.


messense (approved):
LGTM

PR #2590, opened 2022-08-24
fix: export new dict views types

PrettyWood:
Thank you for v0.17.0, which adds https://github.com/PyO3/pyo3/pull/2358!
Unfortunately we can't use them directly since those new types are not exposed and getting them directly from `dict` doesn't work as well since `dict` module is private

If this is approved, could we add this in a v0.17.1 please?

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -16,6 +16,8 @@ pub use self::datetime::{
     PyTzInfo, PyTzInfoAccess,
 };
 pub use self::dict::{IntoPyDict, PyDict};
+#[cfg(not(PyPy))]
+pub use self::dict::{PyDictItem, PyDictKeys, PyDictValues};
 pub use self::floatob::PyFloat;
 #[cfg(all(not(Py_LIMITED_API), not(PyPy)))]
 pub use self::frame::PyFrame;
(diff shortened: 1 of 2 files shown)


src/types/mod.rs
```diff
@@ -16,6 +16,8 @@ pub use self::datetime::{
     PyTzInfo, PyTzInfoAccess,
 };
 pub use self::dict::{IntoPyDict, PyDict};
+#[cfg(not(PyPy))]
+pub use self::dict::{PyDictItem, PyDictKeys, PyDictValues};
```
davidhewitt, line 20: I think you probably wanted the struct `PyDictItems` rather than the `PyDictItem` trait:

```suggestion
pub use self::dict::{PyDictItems, PyDictKeys, PyDictValues};
```

davidhewitt (commented):
Ah, whoops! Could you please also add a CHANGELOG entry? Maybe the following:

```
## [Unreleased]

### Fixed

- Fix visibility of `PyDictItems`, `PyDictKeys`, and `PyDictValues` types added in PyO3 0.17.0.
```

src/types/mod.rs
PrettyWood: Yup thank you so much!

PrettyWood:
Thank you @davidhewitt for the review! Changes applied

davidhewitt (approved):
Thanks for fixing 👍

Yes will push out a 0.17.1, maybe with other fixes if we get any other bug reports.

PR #2592, opened 2022-08-25
pyfunction: fix from_py_with on Option<T> argument

### LINKED ISSUE #2280: `from_py_with` doesn't seem to play nicely with optional arguments
### Bug Description

When trying to build optional arguments, using `from_py_with` I'm getting 

```
expected struct `InnerType`, found enum `Option`
```

### Steps to Reproduce

```rs

fn build_regex(obj: &PyAny) -> PyResult<Option<Regex>> {
    if obj.is_none() {
        Ok(None)
    } else {
        let regex_str: &str = obj.extract()?;
        let regex = Regex::new(regex_str).map_err(|e| PyValueError::new_err(e.to_string()))?;
        Ok(Some(regex))
    }
}

#[pyfunction]
pub fn regex_match(
    str: String,
    #[pyo3(from_py_with = "build_regex")] pattern: Option<Regex>,
) -> PyResult<bool> {
    if let Some(pattern) = pattern {
        if !pattern.is_match(&str) {
            return Ok(false);
        }
    }
    Ok(true)
}
```

(I know in this case I could build the regex inside `regex_match` but in practice I want to be able to call `regex_match` from other code with a pre-compiled regex.

### Backtrace

```shell
  --> src/validators.rs:36:52
   |
36 |     #[pyo3(from_py_with = "build_regex")] pattern: Option<Regex>,
   |                                                    ^^^^^^ expected struct `regex::Regex`, found enum `Option`
   |
   = note: expected struct `regex::Regex`
                found enum `Option<_>`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `pydantic-core` due to previous error
```
```


MacOS 12.2.1

3.10.4

rustc 1.59.0 (9d1b2106e 2022-02-23)

### Your PyO3 version

`{version = "0.16.3", features = ["extension-module"]}`

pyenv, yes

### Additional Info

_No response_

### PR DESCRIPTION
**Pull request by davidhewitt**
This fixes #2280 by making it do the "obvious" thing - an argument `#[pyo3(from_py_with = "foo")] x: Option<i32>` requires `foo` to be a function with returns `PyResult<Option<i32>>`. Turns out there was just a slight typo in the macro-generated code.

### DIFF
--- a/tests/test_pyfunction.rs
+++ b/tests/test_pyfunction.rs
@@ -165,6 +165,42 @@ fn test_function_with_custom_conversion_error() {
     });
 }
 
+#[test]
+fn test_from_py_with_defaults() {
+    fn optional_int(x: &PyAny) -> PyResult<Option<i32>> {
+        if x.is_none() {
+            Ok(None)
+        } else {
+            Some(x.extract()).transpose()
+        }
+    }
+
+    // issue 2280 combination of from_py_with and Option<T> did not compile
+    #[pyfunction]
+    fn from_py_with_option(#[pyo3(from_py_with = "optional_int")] int: Option<i32>) -> i32 {
+        int.unwrap_or(0)
+    }
+
+    #[pyfunction(len = "0")]
+    fn from_py_with_default(#[pyo3(from_py_with = "PyAny::len")] len: usize) -> usize {
+        len
+    }
+
+    Python::with_gil(|py| {
+        let f = wrap_pyfunction!(from_py_with_option)(py).unwrap();
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tests/test_pyfunction.rs [hunk 1]
```diff
@@ -114,23 +113,20 @@ fn test_functions_with_function_args() {
     });
 }
 
-#[cfg(not(Py_LIMITED_API))]
```
davidhewitt, line 117: Not directly related to this PR; just noticed that these attributes are probably not necessary so deleted them.

messense (approved):

PR #2593, opened 2022-08-25
docs: mention PyBuffer

### LINKED ISSUE #954: Guide entry for PyBuffer

davidhewitt:
Uses PEP 688 `types.Buffer` to describe `PyBuffer<T>` in the conversion tables. Will leave as draft until PEP 688 is finalised.

Closes #954

### DIFF
--- a/guide/src/conversions/tables.md
+++ b/guide/src/conversions/tables.md
@@ -33,6 +33,7 @@ The table below contains the Python type and the corresponding function argument
 | `datetime.time` | -                             | `&PyTime`            |
 | `datetime.tzinfo` | -                           | `&PyTzInfo`          |
 | `datetime.timedelta` | -                        | `&PyDelta`           |
+| `collections.abc.Buffer` | -                    | `PyBuffer<T>`        |
 | `typing.Optional[T]` | `Option<T>`              | -                    |
 | `typing.Sequence[T]` | `Vec<T>`                 | `&PySequence`        |
 | `typing.Mapping[K, V]` | `HashMap<K, V>`, `BTreeMap<K, V>`, `hashbrown::HashMap<K, V>`[^2], `indexmap::IndexMap<K, V>`[^3] | `&PyMapping` |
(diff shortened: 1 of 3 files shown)


adamreichold:
> Will leave as draft until PEP 688 is finalised.

The [PEP](https://peps.python.org/pep-0688/) seems to be accepted, so can this just be "undrafted"?

davidhewitt:
Good spot! I've rebased and updated this PR to mention `collections.abc.Buffer` as per the accepted form of the PEP.

Interestingly the PEP now also defines `__buffer__` and `__release_buffer__` methods. We may want to consider support for them eventually; I'll create a new ticket.

adamreichold:
It never stops... I think we need to bin `wasm-bindgen` to version 0.2.84 for the MSRV builds to avoid pulling in `syn` 2.0.

davidhewitt:
Thanks for fixing MSRV! How would you feel if I started a patch to bump MSRV to 1.56 as per #3113 ? Would mean we would be waiting until mid June to release 0.19 (i.e. around one month from now)

adamreichold:
> Thanks for fixing MSRV! How would you feel if I started a patch to bump MSRV to 1.56 as per https://github.com/PyO3/pyo3/issues/3113 ? Would mean we would be waiting until mid June to release 0.19 (i.e. around one month from now)

Personally, I would feel better to release the functionality created so far as 0.19 with our current MSRV (because we know it is compatible) and then immediately after that go about bumping both Rust and Python version requirements.

I do loathe the MSRV picket line, but I also think it is important work for us maintainers to make our crates available in as many environments as reasonably possible.

adamreichold:
(This of course purely psychological, but 0.20 would also look nicer than 0.19 as "break-stuff-that-used-to-work" release.)

adamreichold:
@davidhewitt Penny for your thoughts on whether to bump MSRV before or after releasing 0.19?

davidhewitt:
I think lets release 0.19 first as you suggest. I'll try to write up a list of what I was hoping to get into 0.19 later

Issue #1444, opened 2021-02-24
Sharing pyclasses between multiple Rust packages


davidhewitt:
I've seen several questions about re-using `#[pyclass]` types between multiple Rust packages, for example on [gitter](https://gitter.im/PyO3/Lobby?at=5fdc8b47aa6bb528c3592ec1), [stack overflow](https://stackoverflow.com/questions/62686840/unable-to-use-types-between-multiple-rust-libraries-via-python-bindings-created) and most recently in #1418.

This issue is indended to be a place for the discussion to centralize regarding understanding the difficulties in this and to brainstorm solutions / API designs which we can add to PyO3.

Below is my write-up of the issue. Note that I haven't tried to any of this myself; this is just where I think the difficulties are. I could be totally wrong - please use this thread to share evidence, ask questions, and test ideas.

---

Let's call package `A` the "original" package, which defines a `#[pyclass] MyClass`.  Package `B` is a child crate which makes use of `use a::MyClass` in Rust code, to re-use the pyclass.

The key issue is that `#[pyclass]` stores the pyclass type object in `static` storage. This means that (if Rust's usual `rlib` linkage is used) packages `A` and `B`  will have their own copies of the `MyClass` type object, and Python will think that they're actually different types coming from the two packages.

I think the core of the solution to this is to make `B` link against the package `A` C shared library - the same `.so` file which Python usess to load package `A`. Then `B` should be able to re-use the pyclass from `A`.

This comes with a few steps:
  - `A` will need to export additional symbols which `B` can use. This probably includes _at a minimum_ the `MyClass` static type object, and probably functions for converting `PyAny` <-> `MyClass`.
  - `B` will need to use `A` as a "C" dependency which it links to using an `extern "C"` block, where it imports symbols from `A`.
  - I'm unsure if `B` will be able to `use a::MyClass;` at all - it may need to have a duplicate definition which uses the exported symbols internally.
  
Looking at that, it might mean that we'd eventually have `pyclass_export!` and `pyclass_import!` macros in PyO3 to help with this.

The above is all a poorly-worded brain dump of what _might_ be a solution. I'm really not sure - anyone who is trying to do this, please comment, ask questions, provide error logs, and we can figure this out together.

milesgranger:
[...]

PR #2604, opened 2022-08-31
Add chrono support

pickfire:
Moved from https://github.com/chronotope/chrono/pull/542

What's left

- [x] port tests
- [ ] wait for chrono upstream to accept the changes https://github.com/chronotope/chrono/pull/812

cc @Psykopear I just pushed so that you can help review even though it's not complete yet, but all the types are now there

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ inventory = { version = "0.3.0", optional = true }
 
 # crate integrations that can be added using the eponymous features
 anyhow = { version = "1.0", optional = true }
-chrono = { version = "0.4", path = "../chrono", optional = true }
+chrono = { version = "0.5.0-alpha.1", path = "../chrono", optional = true }
 eyre = { version = ">= 0.4, < 0.7", optional = true }
 hashbrown = { version = ">= 0.9, < 0.13", optional = true }
 indexmap = { version = ">= 1.6, < 1.8", optional = true }
@@ -42,7 +42,7 @@ serde = { version = "1.0", optional = true }
 
 [dev-dependencies]
 assert_approx_eq = "1.1.0"
-chrono = { version = "0.4", path = "../chrono" }
+chrono = { version = "0.5.0-alpha.1", path = "../chrono" }
 criterion = "0.3.5"
 trybuild = "1.0.49"
 rustversion = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ inventory = { version = "0.3.0", optional = true }
 
 # crate integrations that can be added using the eponymous features
 anyhow = { version = "1.0", optional = true }
+chrono = { version = "0.4", path = "../chrono", optional = true }
 eyre = { version = ">= 0.4, < 0.7", optional = true }
 hashbrown = { version = ">= 0.9, < 0.13", optional = true }
 indexmap = { version = ">= 1.6, < 1.8", optional = true }
@@ -41,6 +42,7 @@ serde = { version = "1.0", optional = true }
 
 [dev-dependencies]
 assert_approx_eq = "1.1.0"
+chrono = { version = "0.4", path = "../chrono" }
 criterion = "0.3.5"
 trybuild = "1.0.49"
 rustversion = "1.0"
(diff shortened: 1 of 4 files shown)


src/types/datetime.rs
```diff
@@ -486,6 +486,39 @@ pub fn timezone_utc(py: Python<'_>) -> &PyTzInfo {
...
 }
 
+/// Equivalent to `datetime.timezone`
+// Not making this public yet as the implementation is slightly different from
+// original cpython.
+// #[crate::pyclass]
+// pub(crate) struct PyTimeZone {
+//     offset: PyDelta,
+//     // name
+// }
```
pickfire, line 496: I was thinking it is better to have our own timezone port, but I haven't figured out how to get `pyclass` working yet. Original source https://github.com/python/cpython/blob/29f1b0bb1ff73dcc28f0ca7e11794141b6de58c9/Lib/datetime.py#L2296-L2314

Psykopear: I had the same problem here, it's the first time I work in PyO3, but it seems those macros are never used inside the crate, I think this will have to be done without the macros

Psykopear:
That's great, thank you!
A couple of things:
- You mention the need for an update on the `time` crate, to get the `SECS_PER_DAY` const, but the const is defined inside `chrono` itself (in `src/time_delta.rs`), which doesn't depend on `time` anymore.
- What you do here for the conversion can be done with the current `chrono` api, it's less efficient (using `date.year()` ecc., rather than exposing `date.mdf()`), but it would work without any change to `chrono`. It might make sense to make this PR work with the current version of `chrono`, and maybe update it once `chrono` exposes the needed api? I say this because I feel like the changes on `chrono` would not be trivial, they'd need to expose some structs/functions that are currently private, and this might be a long process. Also, it's should be an implementation detail here, updating it later won't change anything in the exposed functions/structs.

src/types/datetime.rs
mejrs: The problem with having a pyclass such as this is that they behave poorly when shared between libraries. See also https://github.com/PyO3/pyo3/issues/1444

Most likely you (or your users) will run into weird errors like `TypeError: 'PyTimeZone' object cannot be converted to 'PyTimeZone'`.

Cargo.toml
```diff
@@ -56,7 +58,7 @@ widestring = "0.5.1"
 pyo3-build-config = { path = "pyo3-build-config", version = "0.17.1", features = ["resolve-config"] }
 
 [features]
-default = ["macros"]
+default = ["macros", "chrono"]
```
davidhewitt, line 61: When we get around to merging this we should remove it from the defaults here. Also mention in `features.md`.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,266 @@
+#![cfg(feature = "chrono")]
```
davidhewitt, line 1: If there are any `Error` types in `chrono`, it may be nice to also implement `From<Error> for PyErr` for them.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,266 @@
...
+//! `NaiveDate`, `NaiveTime`, `DateTime<Tz>`, `FixedOffset`, and `Utc`.
+//!
+//! # Setup
+//!
+//! To use this feature, add this to your **`Cargo.toml`**:
+//!
+//! ```toml
+//! [dependencies]
+//! # change * to the latest versions
+//! hashbrown = "*"
```
davidhewitt, line 13:
```suggestion
//! chrono = "*"
```

davidhewitt (commented):
Thanks for kicking this off! Please feel free to ping me whenever you need reviews.

src/types/datetime.rs
Psykopear: Just read that issue, it makes sense, so we won't have this here and we should ask users to implement the type in their own crate. We could add the example to the docs pointing at that issue, so that if it's fixed in the future we can move the PyClass here.

Psykopear:
@pickfire I opened a PR https://github.com/pickfire/pyo3/pull/1 over your fork, I updated this to work with the published version of `chrono`, without any change to the api. Let me know what you think, and if you like the idea feel free to merge it.

PR #2608, opened 2022-09-07
netlify: redirect to `main/` for deploy previews

messense:
So that the deploy preview url just works.

See https://docs.netlify.com/configure-builds/environment-variables/#build-metadata


messense:
It works:

```bash
$ curl https://deploy-preview-2608--pyo3.netlify.app/
<meta http-equiv=refresh content=0;url=main/>
```

davidhewitt (approved):
Thanks! I was thinking of adding this eventually too :)

PR #2612, opened 2022-09-09
Add chrono 0.4 integration

Psykopear:
This PR adds the integration for chrono 0.4, and it supersedes https://github.com/PyO3/pyo3/pull/2604
Most of the work here was done by @pickfire , I just adapted it to work with the published version of chrono.

Here the list of PR that brought us here:
https://github.com/chronotope/chrono/pull/542
https://github.com/PyO3/pyo3/pull/2604
https://github.com/pickfire/pyo3/pull/1
https://github.com/chronotope/chrono/pull/812

Thanks to @pickfire for doing the hard work, and to @djc and @davidhewitt for the reviews and help during the whole process.

Closes #2604

### DIFF
--- a/guide/src/features.md
+++ b/guide/src/features.md
@@ -96,12 +96,12 @@ Adds a dependency on [anyhow](https://docs.rs/anyhow). Enables a conversion from
 ### `chrono`
 
 Adds a dependency on [chrono](https://docs.rs/chrono). Enables a conversion from [chrono](https://docs.rs/chrono)'s types to python:
-- [Duration](https://docs.rs/chrono/0.4.22/chrono/struct.Duration.html) -> [`PyDelta`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDelta.html)
-- [FixedOffset](https://docs.rs/chrono/0.4.22/chrono/offset/struct.FixedOffset.html) -> [`PyDelta`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDelta.html)
-- [Utc](https://docs.rs/chrono/0.4.22/chrono/offset/struct.Utc.html) -> [`PyTzInfo`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyTzInfo.html)
-- [NaiveDate](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveDate.html) -> [`PyDate`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDate.html)
-- [NaiveTime](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveTime.html) -> [`PyTime`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyTime.html)
-- [DateTime](https://docs.rs/chrono/0.4.22/chrono/struct.DateTime.html) -> [`PyDateTime`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDateTime.html)
+- [Duration](https://docs.rs/chrono/0.4.22/chrono/struct.Duration.html) -> [`PyDelta`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDelta.html)
+- [FixedOffset](https://docs.rs/chrono/0.4.22/chrono/offset/struct.FixedOffset.html) -> [`PyDelta`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDelta.html)
+- [Utc](https://docs.rs/chrono/0.4.22/chrono/offset/struct.Utc.html) -> [`PyTzInfo`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyTzInfo.html)
+- [NaiveDate](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveDate.html) -> [`PyDate`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDate.html)
+- [NaiveTime](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveTime.html) -> [`PyTime`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyTime.html)
+- [DateTime](https://docs.rs/chrono/0.4.22/chrono/struct.DateTime.html) -> [`PyDateTime`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDateTime.html)
 
 ### `eyre`
 
--- a/guide/src/features.md
+++ b/guide/src/features.md
@@ -94,6 +94,7 @@ These features enable conversions between Python types and types from other Rust
 Adds a dependency on [anyhow](https://docs.rs/anyhow). Enables a conversion from [anyhow](https://docs.rs/anyhow)’s [`Error`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html) type to [`PyErr`](https://docs.rs/pyo3/latest/pyo3/struct.PyErr.html), for easy error handling.
 
 ### `chrono`
+
 Adds a dependency on [chrono](https://docs.rs/chrono). Enables a conversion from [chrono](https://docs.rs/chrono)'s types to python:
 - [Duration](https://docs.rs/chrono/0.4.22/chrono/struct.Duration.html) -> [`PyDelta`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDelta.html)
 - [FixedOffset](https://docs.rs/chrono/0.4.22/chrono/offset/struct.FixedOffset.html) -> [`PyDelta`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDelta.html)
--- a/guide/src/features.md
+++ b/guide/src/features.md
@@ -96,12 +96,12 @@ Adds a dependency on [anyhow](https://docs.rs/anyhow). Enables a conversion from
 ### `chrono`
 
 Adds a dependency on [chrono](https://docs.rs/chrono). Enables a conversion from [chrono](https://docs.rs/chrono)'s types to python:
-- [Duration](https://docs.rs/chrono/0.4.22/chrono/struct.Duration.html) -> [`PyDelta`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDelta.html)
-- [FixedOffset](https://docs.rs/chrono/0.4.22/chrono/offset/struct.FixedOffset.html) -> [`PyDelta`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDelta.html)
-- [Utc](https://docs.rs/chrono/0.4.22/chrono/offset/struct.Utc.html) -> [`PyTzInfo`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyTzInfo.html)
-- [NaiveDate](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveDate.html) -> [`PyDate`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDate.html)
-- [NaiveTime](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveTime.html) -> [`PyTime`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyTime.html)
-- [DateTime](https://docs.rs/chrono/0.4.22/chrono/struct.DateTime.html) -> [`PyDateTime`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDateTime.html)
+- [Duration](https://docs.rs/chrono/latest/chrono/struct.Duration.html) -> [`PyDelta`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDelta.html)
+- [FixedOffset](https://docs.rs/chrono/latest/chrono/offset/struct.FixedOffset.html) -> [`PyDelta`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDelta.html)
+- [Utc](https://docs.rs/chrono/latest/chrono/offset/struct.Utc.html) -> [`PyTzInfo`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyTzInfo.html)
+- [NaiveDate](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDate.html) -> [`PyDate`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDate.html)
+- [NaiveTime](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveTime.html) -> [`PyTime`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyTime.html)
+- [DateTime](https://docs.rs/chrono/latest/chrono/struct.DateTime.html) -> [`PyDateTime`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyDateTime.html)
 
 ### `eyre`
 
--- a/guide/src/features.md
+++ b/guide/src/features.md
@@ -93,6 +93,15 @@ These features enable conversions between Python types and types from other Rust
 
 Adds a dependency on [anyhow](https://docs.rs/anyhow). Enables a conversion from [anyhow](https://docs.rs/anyhow)’s [`Error`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html) type to [`PyErr`](https://docs.rs/pyo3/latest/pyo3/struct.PyErr.html), for easy error handling.
 
+### `chrono`
+Adds a dependency on [chrono](https://docs.rs/chrono). Enables a conversion from [chrono](https://docs.rs/chrono)'s types to python:
+- [Duration](https://docs.rs/chrono/0.4.22/chrono/struct.Duration.html) -> [`PyDelta`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDelta.html)
+- [FixedOffset](https://docs.rs/chrono/0.4.22/chrono/offset/struct.FixedOffset.html) -> [`PyDelta`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDelta.html)
+- [Utc](https://docs.rs/chrono/0.4.22/chrono/offset/struct.Utc.html) -> [`PyTzInfo`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyTzInfo.html)
+- [NaiveDate](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveDate.html) -> [`PyDate`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDate.html)
+- [NaiveTime](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveTime.html) -> [`PyTime`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyTime.html)
+- [DateTime](https://docs.rs/chrono/0.4.22/chrono/struct.DateTime.html) -> [`PyDateTime`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDateTime.html)
+
 ### `eyre`
 
 Adds a dependency on [eyre](https://docs.rs/eyre). Enables a conversion from [eyre](https://docs.rs/eyre)’s [`Report`](https://docs.rs/eyre/latest/eyre/struct.Report.html) type to [`PyErr`](https://docs.rs/pyo3/latest/pyo3/struct.PyErr.html), for easy error handling.
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -50,11 +50,14 @@ use crate::types::{
     timezone_utc, PyDate, PyDateAccess, PyDateTime, PyDelta, PyDeltaAccess, PyTime, PyTimeAccess,
     PyTzInfo, PyTzInfoAccess,
 };
-use crate::{FromPyObject, IntoPy, PyAny, PyObject, PyResult, PyTryFrom, Python, ToPyObject};
+use crate::{
+    AsPyPointer, FromPyObject, IntoPy, PyAny, PyObject, PyResult, PyTryFrom, Python, ToPyObject,
+};
 use chrono::offset::{FixedOffset, Utc};
 use chrono::{
     DateTime, Datelike, Duration, NaiveDate, NaiveDateTime, NaiveTime, Offset, TimeZone, Timelike,
 };
+use pyo3_ffi::{PyDateTime_IMPORT, PyTimeZone_FromOffset};
 use std::convert::TryInto;
 
 impl ToPyObject for Duration {
@@ -234,19 +237,22 @@ impl FromPyObject<'_> for DateTime<Utc> {
     }
 }
 
+// Utiliy function used to convert PyDelta to timezone
+fn pytimezone_fromoffset<'a>(py: &Python<'a>, td: &PyDelta) -> &'a PyAny {
+    // Safety: py.from_borrowed_ptr needs the cast to be valid.
+    // Since we are forcing a &PyDelta as input, the cast should always be valid.
+    unsafe {
+        PyDateTime_IMPORT();
+        py.from_borrowed_ptr(PyTimeZone_FromOffset(td.as_ptr()))
+    }
+}
+
 impl ToPyObject for FixedOffset {
     fn to_object(&self, py: Python<'_>) -> PyObject {
-        let dt_module = py.import("datetime").expect("Failed to import datetime");
-        let dt_timezone = dt_module
-            .getattr("timezone")
-            .expect("Failed to getattr timezone");
         let seconds_offset = self.local_minus_utc();
         let td =
             PyDelta::new(py, 0, seconds_offset, 0, true).expect("Failed to contruct timedelta");
-        let offset = dt_timezone
-            .call1((td,))
-            .expect("Failed to call timezone with timedelta");
-        offset.into()
+        pytimezone_fromoffset(&py, &td).into()
     }
 }
 
@@ -295,6 +301,7 @@ impl FromPyObject<'_> for Utc {
 
 #[cfg(test)]
 mod test_chrono {
+    use crate::chrono::pytimezone_fromoffset;
     use crate::types::*;
     use crate::{Python, ToPyObject};
     use chrono::offset::{FixedOffset, Utc};
@@ -575,15 +582,18 @@ mod test_chrono {
     #[test]
     fn test_pyo3_offset_fixed_topyobject() {
         Python::with_gil(|py| {
-            let py_module = py.import("datetime").unwrap();
-            let py_timezone = py_module.getattr("timezone").unwrap();
+            // Chrono offset
             let offset = FixedOffset::east(3600).to_object(py);
-            let py_timedelta = PyDelta::new(py, 0, 3600, 0, true).unwrap();
-            let py_timedelta = py_timezone.call1((py_timedelta,)).unwrap();
+            // Python timezone from timedelta
+            let td = PyDelta::new(py, 0, 3600, 0, true).unwrap();
+            let py_timedelta = pytimezone_fromoffset(&py, td);
+            // Should be equal
             assert!(offset.as_ref(py).eq(py_timedelta).unwrap());
+
+            // Same but with negative values
             let offset = FixedOffset::east(-3600).to_object(py);
-            let py_timedelta = PyDelta::new(py, 0, -3600, 0, true).unwrap();
-            let py_timedelta = py_timezone.call1((py_timedelta,)).unwrap();
+            let td = PyDelta::new(py, 0, -3600, 0, true).unwrap();
+            let py_timedelta = pytimezone_fromoffset(&py, td);
             assert!(offset.as_ref(py).eq(py_timedelta).unwrap());
         })
     }
@@ -591,10 +601,8 @@ mod test_chrono {
     #[test]
     fn test_pyo3_offset_fixed_frompyobject() {
         Python::with_gil(|py| {
-            let py_module = py.import("datetime").unwrap();
-            let py_timezone = py_module.getattr("timezone").unwrap();
             let py_timedelta = PyDelta::new(py, 0, 3600, 0, true).unwrap();
-            let py_tzinfo = py_timezone.call1((py_timedelta,)).unwrap();
+            let py_tzinfo = pytimezone_fromoffset(&py, &py_timedelta);
             let offset: FixedOffset = py_tzinfo.extract().unwrap();
             assert_eq!(FixedOffset::east(3600), offset);
         })
@@ -604,9 +612,7 @@ mod test_chrono {
     fn test_pyo3_offset_utc_topyobject() {
         Python::with_gil(|py| {
             let utc = Utc.to_object(py);
-            let py_module = py.import("datetime").unwrap();
-            let py_timezone = py_module.getattr("timezone").unwrap();
-            let py_utc = py_timezone.getattr("utc").unwrap();
+            let py_utc = timezone_utc(py);
             assert!(utc.as_ref(py).is(py_utc));
         })
     }
@@ -614,17 +620,17 @@ mod test_chrono {
     #[test]
     fn test_pyo3_offset_utc_frompyobject() {
         Python::with_gil(|py| {
-            let py_module = py.import("datetime").unwrap();
-            let py_timezone = py_module.getattr("timezone").unwrap();
-            let py_utc = py_timezone.getattr("utc").unwrap();
+            let py_utc = timezone_utc(py);
             let py_utc: Utc = py_utc.extract().unwrap();
             assert_eq!(Utc, py_utc);
+
             let py_timedelta = PyDelta::new(py, 0, 0, 0, false).unwrap();
-            let py_timezone_utc = py_timezone.call1((py_timedelta,)).unwrap();
+            let py_timezone_utc = pytimezone_fromoffset(&py, &py_timedelta);
             let py_timezone_utc: Utc = py_timezone_utc.extract().unwrap();
             assert_eq!(Utc, py_timezone_utc);
+
             let py_timedelta = PyDelta::new(py, 0, 3600, 0, false).unwrap();
-            let py_timezone = py_timezone.call1((py_timedelta,)).unwrap();
+            let py_timezone = pytimezone_fromoffset(&py, &py_timedelta);
             assert!(py_timezone.extract::<Utc>().is_err());
         })
     }
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -110,7 +110,6 @@ impl FromPyObject<'_> for Duration {
 
 impl ToPyObject for NaiveDate {
     fn to_object(&self, py: Python<'_>) -> PyObject {
-        // This cast should be safe, right?
         let month = self.month() as u8;
         let day = self.day() as u8;
         let date = PyDate::new(py, self.year(), month, day).expect("Failed to construct date");
@@ -207,7 +206,7 @@ impl FromPyObject<'_> for DateTime<FixedOffset> {
         let tz = if let Some(tzinfo) = dt.get_tzinfo() {
             tzinfo.extract()?
         } else {
-            return Err(PyTypeError::new_err("Not datetime.timezone.tzinfo"));
+            return Err(PyTypeError::new_err("Not datetime.tzinfo"));
         };
         let dt = NaiveDateTime::new(
             NaiveDate::from_ymd(dt.get_year(), dt.get_month().into(), dt.get_day().into()),
@@ -243,7 +242,7 @@ fn pytimezone_fromoffset<'a>(py: &Python<'a>, td: &PyDelta) -> &'a PyAny {
     // Since we are forcing a &PyDelta as input, the cast should always be valid.
     unsafe {
         PyDateTime_IMPORT();
-        py.from_borrowed_ptr(PyTimeZone_FromOffset(td.as_ptr()))
+        py.from_owned_ptr(PyTimeZone_FromOffset(td.as_ptr()))
     }
 }
 
@@ -423,7 +422,7 @@ mod proptests {
 }
 
 #[cfg(test)]
-mod test_chrono {
+mod tests {
     use crate::chrono::pytimezone_fromoffset;
     use crate::types::*;
     use crate::{Python, ToPyObject};
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -1,4 +1,4 @@
-#![cfg(all(feature = "chrono", not(Py_LIMITED_API))]
+#![cfg(all(feature = "chrono", not(Py_LIMITED_API)))]
 
 //! Conversions to and from [chrono](https://docs.rs/chrono/)’s `Duration`,
 //! `NaiveDate`, `NaiveTime`, `DateTime<Tz>`, `FixedOffset`, and `Utc`.
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -33,34 +33,30 @@ use chrono::{
 };
 use std::convert::TryInto;
 
-const SECS_PER_DAY: i64 = 86_400;
-const MICROS_PER_SEC: i64 = 1_000_000;
-
 impl ToPyObject for Duration {
     fn to_object(&self, py: Python<'_>) -> PyObject {
-        // This is the total number of microseconds in the duration,
-        // unlike std::time::Duration we can't have only the fractional part,
-        // so we need to calculate it ourselves.
-        let total_micros = self
-            .num_microseconds()
-            // num_microseconds returns None on overflow
-            .unwrap_or(i64::MAX);
-        let total_secs = total_micros / MICROS_PER_SEC;
-        let micros = total_micros - total_secs * MICROS_PER_SEC;
+        // Total number of days
         let days = self.num_days();
-        let secs = total_secs - days * SECS_PER_DAY;
+        // Remainder of seconds
+        let secs_dur = *self - Duration::days(days);
+        // .try_into() converts i64 to i32, but this should never overflow
+        // since it's at most the number of seconds per day
+        let secs = secs_dur.num_seconds().try_into().unwrap();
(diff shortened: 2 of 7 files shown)


src/conversions/chrono.rs
```diff
@@ -0,0 +1,573 @@
...
+                })
+                .is_err())
+            };
+
+            check(-86399999913600, 0, -999999999, 0, 0); // min
+            check(86399999999999, 999999000, 999999999, 86399, 999999); // max
+
+            check_panic(Duration::min_value());
+            check_panic(Duration::max_value());
+            // TODO: check timedelta underflow
```
pickfire, line 304: @Psykopear Thanks for taking over the PR, can you please help to take a look at this?

src/conversions/chrono.rs
```diff
@@ -0,0 +1,573 @@
...
+            .unwrap()
+            // Same for the conversion from i64 to i32
+            .try_into()
+            .unwrap();
+
+        // We do not need to check i64 to i32 cast from rust because
+        // python will panic with OverflowError.
+        // Not sure if we need normalize here.
+        let delta = PyDelta::new(py, days.try_into().unwrap_or(i32::MAX), secs, micros, false)
+            .expect("Failed to construct delta");
```
pickfire, line 60: I wasn't sure what to do here.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,573 @@
...
+        let dt_module = py.import("datetime").expect("Failed to import datetime");
+        let dt_timezone = dt_module
+            .getattr("timezone")
+            .expect("Failed to getattr timezone");
+        let seconds_offset = self.local_minus_utc();
+        let td =
+            PyDelta::new(py, 0, seconds_offset, 0, true).expect("Failed to contruct timedelta");
+        let offset = dt_timezone
+            .call1((td,))
+            .expect("Failed to call timezone with timedelta");
```
pickfire, line 224: I am not sure if there is a better way to do this or we should just create our own wrapper for `PyTimeZone` class like python `datetime.timezone`? That one I am not sure how to get the `#[pyclass]` working. Maybe there is a better way to refer to python `datetime.timezone` without doing this?

src/conversions/chrono.rs
Psykopear: Yes, I'll try to add tests for the underflow too

src/conversions/chrono.rs
Psykopear: I'll need to look at the implementation because this is something that is not present in the python api, I'll comment once I know more

src/conversions/chrono.rs
Psykopear: I know it's not ideal, but I couldn't find a better way to do this either, I'll have a deeper look at this this afternoon and let you know if I find something interesting

src/conversions/chrono.rs
Psykopear: Ok, I found the function that is called in cpython: https://github.com/python/cpython/blob/023c51d9d8e2fd91069eea2bf12e373f1c71e9d2/Modules/_datetimemodule.c#L1109
```
Normalize the members iff normalize is
true.  Passing false is a speed optimization, if you know for sure
that seconds and microseconds are already in their proper ranges.  In any
case, raises OverflowError and returns NULL if the normalized days is out
of range.
```

Since we don't check that the data is in proper range, we should probably pass `true` here

src/conversions/chrono.rs
Psykopear: So, when you say to check for timedelta (I assume it's `Duration` in this case) underflow, you mean something like this?
```rust
let delta = Duration::seconds(-(i64::MIN + 1)) - Duration::seconds(2);
```
This panics even before trying to convert it to python, but that's not something related to the pyo3-chrono integration, it's more of a `chrono` problem because we can use `std::ops::Sub` rather than `Duration::checked_sub`, which returns an `Option<Duration>`, so I don't think we need to test it here.

src/conversions/chrono.rs
Psykopear: So, the only way we could do this without executing python code, would be if we could construct a `PyTzInfo` instance ourselves, but the docstring above it says that we can't construct one directly. If I understood everything correctly, we would have to create a new class, which inherits form this abstract baseclass, and then implement the conversion between this class and the non abstract ones in `timezone.utc` and `zoneinfo`, but I think we go back to the problem we discussed in the previous PR? https://github.com/PyO3/pyo3/pull/2604#discussion_r960046250

src/conversions/chrono.rs
pickfire: Yes, but I am not sure if we need to normalize, if I am not wrong, it converts the microseconds to seconds if it is like too much.

src/conversions/chrono.rs
pickfire: Not sure if @davidhewitt have any suggestions for this.

src/conversions/chrono.rs
Psykopear: So, I did some more invetigation, and not normalizing the pydelta might lead to some problems, and we should do more checks/normalization ourselves if we don't want to normalize, so I'm going to set it to true.
For example this test won't pass if we don't normalize:
```rust
let delta = Duration::days(-1) + Duration::seconds(1) + Duration::microseconds(-10);
Python::with_gil(|py| {
    let delta = delta.to_object(py);
    let delta: &PyDelta = delta.extract(py).unwrap();
    let py_delta = PyDelta::new(py, -1, 1, -10, true).unwrap();
    assert!(delta.eq(py_delta).unwrap());
});
```

But if I compare the result of calling `.total_seconds()` they return the same result

Psykopear:
@davidhewitt I think the PR is ready to be reviewed. I still get CI failures on clippy, because it's trying to compile the integration with the limited python api, but I think in this case it doesn't make much sense since this integration needs it, right?
How would you like to proceed here?

src/conversions/chrono.rs
```diff
@@ -0,0 +1,663 @@
...
+            Python::with_gil(|py| {
+                let py_delta = PyDelta::new(py, py_days, py_seconds, py_ms, false).unwrap();
+                let py_delta: Duration = py_delta.extract().unwrap();
+                assert_eq!(py_delta, delta, "{}: {} != {}", name, py_delta, delta);
+            })
+        };
+
+        check(
+            "min value",
+            Duration::seconds(-86399999913600),
```
pickfire, line 332: Why not use `Duration::min_value()`?

src/conversions/chrono.rs
```diff
@@ -0,0 +1,663 @@
...
+            -999999999,
+            0,
+            0,
+        );
+        check(
+            "max value",
+            Duration::seconds(86399999999999) + Duration::microseconds(999999),
```
pickfire, line 339: Why not use `Duration::max_value()`?

guide/src/features.md
```diff
@@ -93,6 +93,15 @@ These features enable conversions between Python types and types from other Rust
 
 Adds a dependency on [anyhow](https://docs.rs/anyhow). Enables a conversion from [anyhow](https://docs.rs/anyhow)’s [`Error`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html) type to [`PyErr`](https://docs.rs/pyo3/latest/pyo3/struct.PyErr.html), for easy error handling.
 
+### `chrono`
```
pickfire, line 96:
```suggestion
### `chrono`

```

guide/src/features.md
```diff
@@ -93,6 +93,15 @@ These features enable conversions between Python types and types from other Rust
...
+Adds a dependency on [chrono](https://docs.rs/chrono). Enables a conversion from [chrono](https://docs.rs/chrono)'s types to python:
+- [Duration](https://docs.rs/chrono/0.4.22/chrono/struct.Duration.html) -> [`PyDelta`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDelta.html)
+- [FixedOffset](https://docs.rs/chrono/0.4.22/chrono/offset/struct.FixedOffset.html) -> [`PyDelta`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDelta.html)
+- [Utc](https://docs.rs/chrono/0.4.22/chrono/offset/struct.Utc.html) -> [`PyTzInfo`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyTzInfo.html)
+- [NaiveDate](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveDate.html) -> [`PyDate`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDate.html)
+- [NaiveTime](https://docs.rs/chrono/0.4.22/chrono/naive/struct.NaiveTime.html) -> [`PyTime`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyTime.html)
+- [DateTime](https://docs.rs/chrono/0.4.22/chrono/struct.DateTime.html) -> [`PyDateTime`](https://docs.rs/pyo3/0.17.1/pyo3/types/struct.PyDateTime.html)
```
pickfire, line 103: I don't think we need to manually link each type here, doing it this way can make the links pointing to the wrong version easily.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,663 @@
...
+//! pyo3 = { version = "*", features = ["chrono"] }
+//! chrono = "0.4"
+// workaround for `extended_key_value_attributes`: https://github.com/rust-lang/rust/issues/82768#issuecomment-803935643
+#![cfg_attr(docsrs, cfg_attr(docsrs, doc = concat!("pyo3 = { version = \"", env!("CARGO_PKG_VERSION"),  "\", features = [\"chrono\"] }")))]
+#![cfg_attr(
+    not(docsrs),
+    doc = "pyo3 = { version = \"*\", features = [\"chrono\"] }"
+)]
+//! ```
+//!
```
mejrs, line 22: I'd like (not require) an example showing why you might want to use the chrono feature. What does it let you do that you otherwise can't?

src/conversions/chrono.rs
Psykopear: Because that's the min value allowed by python, `Duration::min_value()` is actually a lot smaller:
```
Duration::seconds(-86399999913600) = Duration { secs: -86399999913600, nanos: 0 }
Duration::min_value() =              Duration { secs: -9223372036854776, nanos: 192000000 }
```

src/conversions/chrono.rs
Psykopear: Same thing as the min_value

mejrs (approved):
Can you add an entry to https://docs.rs/pyo3/latest/pyo3/#optional-feature-flags as well? :)

guide/src/features.md
Psykopear: yes I was not sure about this since pointing to the exact version could lead to outdated documentation if pyo3 upgrades the dependency, but pointing to latest will link to the wrong one if chrono changes anything relevant.

The reason I decided to point to the exact version is that at least pyo3 has control over this file, so it can be upgraded when the dependency is, but maybe just listing the structs without linking them could be enough.

@davidhewitt let me know what you prefer here and I'll fix this

src/conversions/chrono.rs
Psykopear: Sure, I'll add a brief example and an entry to the doc page you linked

Psykopear: wait, I think I broke the CI, pushing a fix

Psykopear: Ok, example added and optional-feature-flags section updated.
I still have an issue with the CI as I mentioned in a previous comment, Clippy fails because it's running with the limited python api, but this integration needs it, how do you want to proceed here?

src/conversions/chrono.rs
pickfire: Why not do both rust min value and python min value? And leave a comment saying that one is rust min and one is python min.

src/conversions/chrono.rs
```diff
@@ -14,14 +14,32 @@
...
+//!     Python::with_gil(|py| {
+//!         // Create an UTC datetime in python
+//!         let py_tz = Utc.to_object(py);
+//!         let py_tz = py_tz.cast_as(py).unwrap();
+//!         let pydatetime = PyDateTime::new(py, 2022, 1, 1, 12, 0, 0, 0, Some(py_tz)).unwrap();
+//!         println!("PyDateTime:\t{pydatetime}");
+//!         // Now convert it to chrono's DateTime<Utc>
+//!         let chrono_datetime: DateTime<Utc> = pydatetime.extract().unwrap();
+//!         println!("DateTime<Utc>:\t{chrono_datetime}");
+//!     });
```
pickfire, line 40: I wish there is an interesting example using `inline_python!`. https://docs.rs/inline-python/latest/inline_python/

I think it also helps demonstrate this nicely, although not sure if we want to run doctest since it takes in a new dependency.

src/conversions/chrono.rs
pickfire: Ah, I didn't know there is a C API for this https://github.com/python/cpython/blob/023c51d9d8e2fd91069eea2bf12e373f1c71e9d2/Modules/_datetimemodule.c#L1182-L1205, I only saw https://github.com/python/cpython/blob/023c51d9d8e2fd91069eea2bf12e373f1c71e9d2/Lib/datetime.py#L2301-L2314 so I thought the implementation was from python side.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,681 @@
...
+    }
+
+    #[test]
+    fn test_pyo3_offset_fixed_topyobject() {
+        Python::with_gil(|py| {
+            let py_module = py.import("datetime").unwrap();
+            let py_timezone = py_module.getattr("timezone").unwrap();
+            let offset = FixedOffset::east(3600).to_object(py);
+            let py_timedelta = PyDelta::new(py, 0, 3600, 0, true).unwrap();
+            let py_timedelta = py_timezone.call1((py_timedelta,)).unwrap();
```
pickfire, line 577: I guess this can also use the C API.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,681 @@
...
+        })
+    }
+
+    #[test]
+    fn test_pyo3_offset_utc_topyobject() {
+        Python::with_gil(|py| {
+            let utc = Utc.to_object(py);
+            let py_module = py.import("datetime").unwrap();
+            let py_timezone = py_module.getattr("timezone").unwrap();
+            let py_utc = py_timezone.getattr("utc").unwrap();
```
pickfire, line 604: `PyDateTime_TimeZone_UTC` can probably be used, IIRC pyo3 have a wrapper for this.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,681 @@
...
+        })
+    }
+
+    #[test]
+    fn test_pyo3_offset_fixed_frompyobject() {
+        Python::with_gil(|py| {
+            let py_module = py.import("datetime").unwrap();
+            let py_timezone = py_module.getattr("timezone").unwrap();
+            let py_timedelta = PyDelta::new(py, 0, 3600, 0, true).unwrap();
+            let py_tzinfo = py_timezone.call1((py_timedelta,)).unwrap();
```
pickfire, line 592: I guess this can also use the C API.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,681 @@
...
+    #[test]
+    fn test_pyo3_offset_utc_frompyobject() {
+        Python::with_gil(|py| {
+            let py_module = py.import("datetime").unwrap();
+            let py_timezone = py_module.getattr("timezone").unwrap();
+            let py_utc = py_timezone.getattr("utc").unwrap();
+            let py_utc: Utc = py_utc.extract().unwrap();
+            assert_eq!(Utc, py_utc);
+            let py_timedelta = PyDelta::new(py, 0, 0, 0, false).unwrap();
+            let py_timezone_utc = py_timezone.call1((py_timedelta,)).unwrap();
```
pickfire, line 618: I guess this can also use the C API.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,681 @@
...
+    fn into_py(self, py: Python<'_>) -> PyObject {
+        ToPyObject::to_object(&self, py)
+    }
+}
+
+impl FromPyObject<'_> for FixedOffset {
+    /// Convert python tzinfo to rust [`FixedOffset`].
+    ///
+    /// Note that the conversion will result in precision lost in microseconds as chrono offset
+    /// does not supports microseconds.
```
pickfire, line 258: @djc I guess even with chrono changes to `TimeDelta` this will stay the same?

src/conversions/chrono.rs
Psykopear: This is tested a few lines below that (currently 329-333), I just refactored the tests you wrote because it wasn't entirely clear what input data was making the test fail when they failed, but the assertions should not have changed.

Agree on the comment to make it clear why we are testing that specific value, I'll add it tomorrow once I work on it again

src/conversions/chrono.rs
Psykopear: thanks, I'm now using the `PyTimeZone_FromOffset` and `timezone_utc` instead of running python code

src/conversions/chrono.rs
Psykopear: yep, fixed

src/conversions/chrono.rs
Psykopear: yes, fixed

src/conversions/chrono.rs
Psykopear: yes, fixed

src/conversions/chrono.rs
Psykopear: fixed

src/conversions/chrono.rs
Psykopear: Didn't know about this project. This would make the examples much more readable, but I'm not sure we want to hide the details of how to do things with PyO3 in the documentation, new users of the library would be confused by the requirement of an external dependency to run PyO3 code. Anyway I think this is out of scope for this PR.

src/conversions/chrono.rs
djc: The `TimeDelta` changes we're thinking about for 0.5 don't have much to do with the `FixedOffset` you're using here, as far as I know. We could consider increasing precision in `FixedOffset::local_minus_utc` if you think that's useful? That might conceivably be done in a backwards compatible version for 0.4, too.

Psykopear:
Ok, I applied the suggested changes, moved to using the C API for offsets, added a couple of tests for bounds checking, added the newsfragment for the changelog and rebased everything on the `main` branch.

The CI is now finally passing!

@davidhewitt never used proptest, but the readme is quite clear, and I see it's already used in the `types/num` module, so I'll try using it for datetimes too and see if I catch any other edge case.

Psykopear:
Wow, I really like proptests. I hope I used it properly, please review the tests and let me know if you want me to change anything.
I added tests for the roundtrip of all the involved types with all the allowed values.
There are values that are valid in Rust but will panic when converting to python, so I only proptested for values that are allowed in Python.
Some considerations

#### `FixedOffset`
The roundtrip conversion of `FixedOffset` didn't work as it is.
During the trip, `FixedOffset::east` for negative values turned into a positive one: `-00:00:01` turns into `23:59:59`.
I made it work by not normalizing the `PyDelta` when converting the `FixedOffset` to Python, but I'm not sure this is the right approach. Can we have problems using the delta in python without the normalization?
I added a comment marked `XXX` where this happens.

#### `DateTime<Tz>`
This can't be tested this way with proptest, because we don't have a `FromPyObject` for `DateTime<Tz>`.
We convert `DateTime<Tz>` to fixed offset datetimes in python, so if we get anything back it's turned into a `DateTime<FixedOffset>`.

I think to be able to convert `DateTime<Tz>` (other than FixedOffset and Utc) back and forth from Python, we'd need to integrate `chrono_tz` and require the `zoneinfo` module in python (so python >=3.9), so this is probably out of scope for this PR.
Would it make sense to add a separate integration to properly handle IANA timezones?

#### Other tests
I think the proptests here make some of the test we wrote previously reduntant. Do you want me to remove them, or is it ok to keep both?

src/conversions/chrono.rs
pickfire: Yes, but it's for the users that are coming from inline_python, not pyo3. It was the motivation for me to contribute the initial pull request to chrono, because I wanted to use it in inline_python, and inline_python uses pyo3, I wouldn't have use pyo3 directly.

I think would be good to keep as a separate example given that users coming from inline_python probably have no idea what the pyo3 example is showing.

src/conversions/chrono.rs
pickfire: That would be good I guess if we increase precision in microseconds, I think it could have some impact to users as it it needs to store another byte in some cases but that would mean that there are no precision loss in conversion between chrono FixedOffset and python timedelta.

pickfire:
> I think the proptests here make some of the test we wrote previously reduntant. Do you want me to remove them, or is it ok to keep both?

The existing tests test out some hard-coded known bound values, I think would be nice to keep them, I think proptest does test it but it's more like a blackbox.

pickfire (approved):

Psykopear:
Looks like CI is failing due to throtlling from github, I think I triggered way too many runs today, sorry.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+        // -999999999 <= days <= 999999999
+        Ok(Duration::days(delta.get_days().into())
+            + Duration::seconds(delta.get_seconds().into())
+            + Duration::microseconds(delta.get_microseconds().into()))
+    }
+}
+
+impl ToPyObject for NaiveDate {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        // This cast should be safe, right?
```
davidhewitt, line 113: I think so, and `proptest` hasn't found anything, so we can probably conclude yes and drop this comment

```suggestion
```

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+        let secs_dur = *self - Duration::days(days);
+        // .try_into() converts i64 to i32, but this should never overflow
+        // since it's at most the number of seconds per day
+        let secs = secs_dur.num_seconds().try_into().unwrap();
+        // Fractional part of the microseconds
+        let micros = (secs_dur - Duration::seconds(secs_dur.num_seconds()))
+            .num_microseconds()
+            // This should never panic since we are just getting the fractional
+            // part of the total microseconds, which should never overflow.
+            .unwrap()
```
davidhewitt, line 77: I'm not entirely sure whether the plan is still to feedback upstream to `chrono` for possible API additions, this looks like one to me. It could be nice to be able to decompose the `Duration` into some kind of tuple like `(days, seconds, nanos)` where it's known that none of the components overflow, avoiding the need to do the subtraction here ourselves and then `unwrap`.

cc @djc

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+            // Same for the conversion from i64 to i32
+            .try_into()
+            .unwrap();
+
+        // We do not need to check i64 to i32 cast from rust because
+        // python will panic with OverflowError.
+        // We pass true as the `normalize` parameter since we'd need to do several checks here to
+        // avoid that, and it shouldn't have a big performance impact.
+        let delta = PyDelta::new(py, days.try_into().unwrap_or(i32::MAX), secs, micros, true)
+            .expect("Failed to construct delta");
```
davidhewitt, line 87: So if I read this right, this will panic if the `Duration` exceeds the possible range of Python timedeltas? We really need to come up with a fallible `IntoPy` equivalent...

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+    fn extract(ob: &PyAny) -> PyResult<DateTime<FixedOffset>> {
+        let dt = <PyDateTime as PyTryFrom>::try_from(ob)?;
+        let ms = dt.get_fold() as u32 * 1_000_000 + dt.get_microsecond();
+        let h = dt.get_hour().into();
+        let m = dt.get_minute().into();
+        let s = dt.get_second().into();
+        let tz = if let Some(tzinfo) = dt.get_tzinfo() {
+            tzinfo.extract()?
+        } else {
+            return Err(PyTypeError::new_err("Not datetime.timezone.tzinfo"));
```
davidhewitt, line 210: I think it should just be `datetime.tzinfo`? `datetime.timezone` being the fixed offset representation.

```suggestion
            return Err(PyTypeError::new_err("Not datetime.tzinfo"));
```

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+    fn extract(ob: &PyAny) -> PyResult<DateTime<Utc>> {
+        let dt = <PyDateTime as PyTryFrom>::try_from(ob)?;
+        let ms = dt.get_fold() as u32 * 1_000_000 + dt.get_microsecond();
+        let h = dt.get_hour().into();
+        let m = dt.get_minute().into();
+        let s = dt.get_second().into();
+        let tz = if let Some(tzinfo) = dt.get_tzinfo() {
+            tzinfo.extract()?
+        } else {
+            return Err(PyTypeError::new_err("Not datetime.timezone.utc"));
```
davidhewitt, line 230: Error message here needs to be better, presumably what happened here is the user passed a datetime which either didn't have a timezone or had a non-utc timezone.

In either case, should we attempt to convert to a utc timezone?

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+}
+
+impl FromPyObject<'_> for FixedOffset {
+    /// Convert python tzinfo to rust [`FixedOffset`].
+    ///
+    /// Note that the conversion will result in precision lost in microseconds as chrono offset
+    /// does not supports microseconds.
+    fn extract(ob: &PyAny) -> PyResult<FixedOffset> {
+        let py_tzinfo = <PyTzInfo as PyTryFrom>::try_from(ob)?;
+        let py_timedelta = py_tzinfo.call_method1("utcoffset", (ob.py().None(),))?;
```
davidhewitt, line 275: I'm not super familiar with the `utcoffset` method, in the Python docs it looks like it takes a datetime object as argument, but we're passing `None` here. Can you clarify how this works?

Is it also potentially brittle? E.g. the offset might change depending on the date if it's an IANA timezone?

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+}
+
+#[cfg(test)]
+mod proptests {
+    use super::*;
+
+    #[cfg(not(target_arch = "wasm32"))]
+    use proptest::prelude::*;
+
+    #[cfg(not(target_arch = "wasm32"))]
```
davidhewitt, line 312: Personally I'd just make this a submodule of `mod tests`, with `#[cfg(not(target_arch = "wasm32"))]`. 

i.e. full path would be `pyo3::conversions::tests::proptests`.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+                    let roundtripped: DateTime<FixedOffset> = pydt.extract(py).expect("Round trip");
+                    assert_eq!(dt, roundtripped);
+                }
+            })
+        }
+    }
+}
+
+#[cfg(test)]
+mod test_chrono {
```
davidhewitt, line 426: Nit: all test modules called `tests` please.

```suggestion
mod tests {
```

davidhewitt (commented):
Nice, use of proptests looks good!

> The existing tests test out some hard-coded known bound values, I think would be nice to keep them, I think proptest does test it but it's more like a blackbox.

Agreed, `proptest` is great for sniffing out weird edge cases and then it can be nice to make them actual tests so that anyone working on a new implementation sees all the known edge cases.

I'm not entirely sure about the `PyDelta` normalization either, if it's the same as the done documented for [`datetime.timedelta()`](https://docs.python.org/3/library/datetime.html#datetime.timedelta) I think we probably _do_ want this. Is there a `FixedOffset::west` for negative offsets?

I also am a little murky on whether we're doing the right thing when we convert datetimes into  `DateTime<FixedOffset>` and `DateTime<Utc>`. Do we accept the right varieties of timezones and how do these conversions interact with IANA timezones?

If we need to look at supporting `chrono-tz` here, I would be ok with us exploring it.

src/conversions/chrono.rs
djc: Unfortunately we don't really control the `Duration` API in chrono 0.4, since that's bound by the time 0.1 implementation of `Duration`.

pickfire:
> I also am a little murky on whether we're doing the right thing when we convert datetimes into DateTime<FixedOffset> and DateTime<Utc>. Do we accept the right varieties of timezones and how do these conversions interact with IANA timezones?

I didn't look at this previously because it does not come from datetime and instead could come from zoneinfo or pytz. Was also thinking of keeping the scope small and just include stuff from datetime so only utc and fixed offset is added.

src/conversions/chrono.rs
pickfire: IIRC the `TimeDelta` in 0.5 no longer have this issue.

src/conversions/chrono.rs
pickfire: Wouldn't a conversion be surprising and user may loss the original timezone if they converted to Utc? That is why I didn't add this at the first place, I don't think there is a way to retain this information if we converted to utc to convert back with original timezone.

davidhewitt: I agree, though I got the impression the fixedoffset case did some converting, let me reread....

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+    }
+}
+
+// Utiliy function used to convert PyDelta to timezone
+fn pytimezone_fromoffset<'a>(py: &Python<'a>, td: &PyDelta) -> &'a PyAny {
+    // Safety: py.from_borrowed_ptr needs the cast to be valid.
+    // Since we are forcing a &PyDelta as input, the cast should always be valid.
+    unsafe {
+        PyDateTime_IMPORT();
+        py.from_borrowed_ptr(PyTimeZone_FromOffset(td.as_ptr()))
```
davidhewitt, line 246: I think this needs to be `py.from_owned_ptr`? I assume that `PyTimeZone_FromOffset` produces a new object.
```suggestion
        py.from_owned_ptr(PyTimeZone_FromOffset(td.as_ptr()))
```

src/conversions/chrono.rs
```diff
@@ -0,0 +1,848 @@
...
+    }
+}
+
+impl FromPyObject<'_> for FixedOffset {
+    /// Convert python tzinfo to rust [`FixedOffset`].
+    ///
+    /// Note that the conversion will result in precision lost in microseconds as chrono offset
+    /// does not supports microseconds.
+    fn extract(ob: &PyAny) -> PyResult<FixedOffset> {
+        let py_tzinfo = <PyTzInfo as PyTryFrom>::try_from(ob)?;
```
davidhewitt, line 274: @pickfire this is my point - here it looks like we accept any tzinfo and convert it to a fixed offset. If we are going for a no-conversions approach we might need to be careful here? E.g. only accept `datetime.timezone` objects?

src/conversions/chrono.rs
pickfire: Ah, didn't know about that. Is this due the following text in https://docs.python.org/3.10/c-api/datetime.html#c.PyTimeZone_FromOffset

> Return value: New reference.

src/conversions/chrono.rs
pickfire: I think this part is fine because fixedoffset can be anything, but utc should always be utc I think, hence it has a special value for utc. At most, we lost utc when converting back for this part.

src/conversions/chrono.rs
Psykopear: Yes, here we unwrap to i32::MAX since that is already higher than the maximum allowed days in python, and it will fail trying to construct the `PyDelta`. I agree that something like `TryIntoPy` would be helpful, it's not always possible to transform a valid Rust type into a valid Python one

src/conversions/chrono.rs
djc: Yeah, we can definitely fix this for `TimeDelta` 0.5 (PR to main welcome).

src/conversions/chrono.rs
djc: Yeah, given that a Python `tzinfo` might not represent a fixed offset, it seems that we should only accept a `datetime.timezone` here.

src/conversions/chrono.rs
Psykopear: Passing `None` here only works for tzinfo that represents a fixed offset in Python, so it should be correct since that's what we are converting here. You are right, this would fail for any other kind of tzinfo, since the delta would change depending on the date.

See this example: https://docs.python.org/3/library/datetime.html?highlight=utcoffset#examples-of-usage-time
And this class: https://docs.python.org/3/library/datetime.html?highlight=utcoffset#datetime.timezone

The most notable example is UTC:
```python
from datetime import timezone, timedelta
timezone.utc.utcoffset(None) == timedelta(0)
```

But it works for any datetime with tzinfo of class `datetime.timezone` (which is just three UTCs in a trenchcoat):
```python
from datetime import timezone, timedelta, datetime
datetime.now(tz=timezone(timedelta(hours=2))).tzinfo.utcoffset(None) == timedelta(seconds=7200)
```

src/conversions/chrono.rs
Psykopear: sure, will fix it

src/conversions/chrono.rs
Psykopear: Yes, this will fail for any timezone that is not exactly `datetime.timezone.utc` since we don't do any conversion.

src/conversions/chrono.rs
djc: Fair enough.

Psykopear:
Talking about converting datetimes.

I think PyO3 should just move datetimes as they are represented in python:
- Python naive -> Chrono naive
- Python fixed offset -> Chrono fixed offset
- Python UTC -> Chrono UTC
- Python with proper timezone info (`pytz` or `zoneinfo`) -> chrono_tz

Then users of this library will have to decide what to do in case of naivedatetimes. In python naive datetimes are assumed to be in the same timezone of the interpreter that instanced the datetime. So if a user of PyO3 is sure that the datetime was generated on the same timezone, they can decide to convert it to either UTC/FixedOffset or to chrono_tz, but I feel like this is not something PyO3 should decide, since here we can't assume that datetimes were generated on the same timezone where the conversion takes place.

Psykopear:
Talking about converting datetimes, I'm still making my mind about it, but I'm starting to have a clearer vision now.

I think PyO3 should just represent datetimes as they are represented in python:
- Python naive -> Chrono naive
- Python fixed offset -> Chrono fixed offset
- Python UTC -> Chrono UTC
- Python with proper timezone info (`pytz` or `zoneinfo`) -> chrono_tz

Here I assume `chrono`/`chrono_tz` because that's what we are working on right now, but this is true for any other date time library. We could even think of a set of specific (simpler) traits that any datetime "backend" would have to implement to add an integration, and automatically implement `FromPy` and `IntoPy` for those who implement those traits, but one thing at a time.

Ideally naive datetimes should never be exposed to users or sent around the internet, but that's not the case in the real world, so we have to deal with them.

Users of this library will have to decide what to do in case of naive datetimes.
In python naive datetimes are [assumed to be in the same (local) timezone of the interpreter](https://docs.python.org/3/library/datetime.html#datetime.datetime.today) that instanced the datetime (see also the red warning [here](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow)).
If a user of PyO3 can reasonably assume that the datetime was generated in the same timezone their code is ran, they can decide to convert it to either UTC/FixedOffset or to chrono_tz, or even just decide to only use naive datetimes.

But I feel like this is not something PyO3 should decide, since here we can't assume that datetimes were generated on the same timezone where the conversion takes place (think about parsing a naive datetime coming from any source on the internet).

Given these considerations, I think we should:
- Expose a conversion for Python naive -> Chrono naive
- Add (in a different PR, using a different feature since requirements would be different) a chrono_tz integration for Python with IANA tzinfo

About the `ToPyObject` implementation for `DateTime<Tz>`.
Right now we convert it to a fixed offset representation.
This is not wrong, since we are doing it for a specific date, so there should be no ambiguity (apart from future changes in timezones definition, which is not exactly rare, but we can't do better without proper IANA timezones use), and it would work for any python that doesn't have IANA timezones info (so python<3.9 without `pytz` or `backports.zoneinfo` packages).
But if a PyO3 user wants to properly support IANA timezones, they could require python>=3.9 or the presence of a library, and then use a separate `chrono_tz` integration.
The `ToPyObject` for `DateTime<Tz>` would have to be removed from the `chrono` feature if the `chrono_tz` feature is enabled too, since a `ToPyObject` for `DateTime<chrono_tz::Tz>` would be a conflicting implementation, but that's doable.
With `chrono_tz` we should also be able to add the `FromPyObject` given the requirements mentioned before, I already tried that in another library and it seems to work.

src/conversions/chrono.rs
davidhewitt: Great, so it looks like `None` for `zoneinfo.ZoneInfo` returns `None`?

```python
>>> zoneinfo.ZoneInfo("Europe/London").utcoffset(None) == None
True
```

Three final suggestions here:
- Can we please add a comment here documenting that `None` will only work for fixed offsets (as it's not immediately clear to readers IMO).
- Please add a test confirming that `zoneinfo.ZoneInfo("Europe/London")` doesn't convert to `FixedOffset` (so that refactoring later doesn't break this design).
- Maybe add an improved error message here? e.g. `format!("`{:?}` is not a fixed offset timezone", py_tzinfo.repr().unwrap_or_else("repr failed"));` (I think that won't exactly compile but should give you the idea.)

src/conversions/chrono.rs
```diff
@@ -0,0 +1,843 @@
...
+        let py_timedelta = py_tzinfo.call_method1("utcoffset", (ob.py().None(),))?;
+        let py_timedelta = <PyDelta as PyTryFrom>::try_from(py_timedelta)?;
+        Ok(FixedOffset::east(py_timedelta.get_seconds()))
```
davidhewitt, line 276: I think the roundtrip issue related to the `XXX` is caused here - see e.g. this "negative" -18 hours offset:

```python
>>> datetime.timezone(datetime.timedelta(hours=-18)).utcoffset(None)
datetime.timedelta(days=-1, seconds=21600)
```

There's a days component, but we don't process that here. So I think this line is a bug.

src/conversions/chrono.rs
```diff
@@ -0,0 +1,843 @@
...
+}
+
+impl ToPyObject for FixedOffset {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        let seconds_offset = self.local_minus_utc();
+        // XXX: Here we don't normalize, otherwise the meaning
+        // of the offset changes. Normalizing a negative PyDelta and converting
+        // back to rust would transform a `-00:00:01` offset into `23:59:59`.
+        let td =
+            PyDelta::new(py, 0, seconds_offset, 0, false).expect("Failed to contruct timedelta");
```
davidhewitt, line 256: As per other comment I think we should be normalising here, and the bug is instead in the `FromPyObject` implementation.

davidhewitt (commented):
Yep, I agree with all of the above. So I think I'm comfortable that the design we've got here is now correct.

Just a few comments to finish off on the `FixedOffset` conversions, and then I'm tempted to include this in PyO3 0.17.2 (coming soon).

src/conversions/chrono.rs
Psykopear: uuuh, good catch, that's definitely wrong. I think I wanted to use the equivalent of python's `.total_seconds()` on the timedelta, but this only gives us the `seconds` part of it

src/conversions/chrono.rs
```diff
@@ -0,0 +1,926 @@
...
+                    );
+                })
+            };
+
+        check_fixed_offset("fold", 2014, 5, 6, 7, 8, 9, 1_999_999, 999_999, true);
+        check_fixed_offset("non fold", 2014, 5, 6, 7, 8, 9, 999_999, 999_999, false);
+
+        Python::with_gil(|py| {
+            // extract utc with fixedoffset should fail
+            // but fixedoffset from utc seemed to work, maybe because it is also considered fixedoffset?
```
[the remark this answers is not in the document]
Psykopear: ok, deleting the comment

src/conversions/chrono.rs
```diff
@@ -0,0 +1,926 @@
...
+}
+
+impl FromPyObject<'_> for NaiveDateTime {
+    fn extract(ob: &PyAny) -> PyResult<NaiveDateTime> {
+        let dt = <PyDateTime as PyTryFrom>::try_from(ob)?;
+        let h = dt.get_hour().into();
+        let m = dt.get_minute().into();
+        let s = dt.get_second().into();
+        let ms = dt.get_microsecond();
+        let dt = NaiveDateTime::new(
```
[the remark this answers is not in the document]
Psykopear: Right, given we are enforcing a 1:1 translation, returning a hard error here rather than removing tzinfo (or doing a conversion assuming local timezone) seems the reasonable thing to do :+1:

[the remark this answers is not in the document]
Psykopear: I'll add a test too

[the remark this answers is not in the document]
Psykopear: I think it's the right approach, it means that if a user really wants to convert a timezone aware datetime to a naive one, he has to remove the tzinfo himself in the python side.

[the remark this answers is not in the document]
Psykopear: I also added a test

Psykopear:
The CI fails on windows with this error:
```
ZoneInfoNotFoundError('No time zone found with key Europe/London'),
```

The reason is that windows does not offer a system IANA database:
https://docs.python.org/3/library/zoneinfo.html#data-sources

So we'd need to install the `pytz` package in CI, or skip this test on windows

davidhewitt:
I'd prefer the rust tests to not have an external Python dependency, so I think easiest for now to just skip that test on Windows.

davidhewitt (approved):
This looks complete to me! Test coverage looks excellent and I'm feeling confident the implementation behaves as users would expect.

Thank you very much @Psykopear, @pickfire, @mejrs and @djc for all the hard work here - it's been quite the epic journey for this feature. Will release it into 0.17.2 in the near future.

Psykopear:
Great! Thanks everybody, this was really a team effort

messense:
Some proptests failed in https://github.com/PyO3/pyo3/actions/runs/3104312627/jobs/5028714517

Psykopear:
I'm going to check what went wrong, where should I commit the fix? A separate branch and PR?

messense:
> where should I commit the fix? A separate branch and PR?

Yes.

PR #2613, opened 2022-09-10
docs(python_from_rust): codeblock labels to `rust`

ryanrussell:
Signed-off-by: Ryan Russell <git@ryanrussell.org>

### Issue
The demo code from [Include multiple Python files](https://pyo3.rs/latest/python_from_rust.html#include-multiple-python-files) was not labeled `rust`, making it harder to read and understand.

### Changes
- Updated the code block labels to `rust`

There are other instances of this as well, assuming this is merged, I'm happy to send another PR updating the rest. Feel free to advise if you prefer a single PR or multiple...

Best,
Ryan

### DIFF
--- a/.netlify/build.sh
+++ b/.netlify/build.sh
@@ -43,6 +43,8 @@ mv target/guide netlify_build/main/
 cargo xtask doc
 mv target/doc netlify_build/main/doc/
 
+echo "<meta http-equiv=refresh content=0;url=pyo3/>" > netlify_build/main/doc/index.html
+
 ## Build internal docs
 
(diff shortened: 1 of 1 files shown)


guide/src/python_from_rust.md
```diff
@@ -281,7 +281,7 @@ The example below shows:
   imported from `utils/foo.py`
 
 `src/main.rs`:
-```ignore
+```rust
```
adamreichold, line 284: I suspect these were intended to be ignored. I am not sure about the correct syntax, but maybe

~~~
```rust,ignore
...
```
~~~

which achieve the desired behaviour and the desired presentation?

adamreichold: (This does indeed appear to be the reason why the CI is failing.)

ryanrussell: Hey @adamreichold  -- thanks for the feedback.

This `rust,ignore` format appears [valid](https://github.com/rust-lang/crates.io/issues/1150), but I'm not familiar with it.

I'm not seeing this work on my end in rendering(could just be VS Code not rendering the tags?)
image

I updated the PR per your request, please let me know if this is better/working in the github formatter or I should make another change?

ryanrussell: This does indeed seem to render on [git](https://github.com/PyO3/pyo3/blob/f0ee2d037dc31bc7ccabdc76b931cded0344936e/guide/src/python_from_rust.md)

ryanrussell: Oops, I messed up the force push. Just going to start fresh with a new PR, my sincere apologies.

PR #2616, opened 2022-09-13
CHANGELOG: add link for unreleased changes

davidhewitt:
After moving unreleased changes to `newsfragments` (for towncrier) I thought it'd be helpful to add a link to the CHANGELOG to see the unreleased changes.


adamreichold (approved):
Change looks good. We still need to fix the coverage xtask though? Probably by just dropping `--workspace`?

davidhewitt:
> We still need to fix the coverage xtask though?

Yeah agreed, I see you took a stab in #2617, I'll see if I can get anything to work too.

PR #2617, opened 2022-09-13
Selecting the whole workspace conflicts with package selection in llvm_cov_command.

adamreichold:

PR #2619, opened 2022-09-13
ci: updates for cargo-llvm-cov 0.5

davidhewitt:
I think the Windows error in #2617 might be because `xtask` binary is running from the directory which `cargo llvm-cov clean` is trying to remove, so as an experiment am trying moving the `coverage` job to `nox` here:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -341,7 +341,7 @@ jobs:
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
       - run: python -m pip install -U pip nox
-      - run: cargo xtask coverage --output-lcov coverage.lcov
+      - run: nox -s coverage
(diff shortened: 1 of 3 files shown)


adamreichold:
Generally speaking, I think this does highlight a bit that keeping to xtask infrastructures working is additional work indeed. But that might also be me being grumpy due to running into the itertools regression while trying to understand the somewhat inscrutable cargo-llvm-cov error... (I have absolutely no preferences between the two xtask implementations though.)

davidhewitt:
I'm thinking similarly; I was excited to use Rust for xtask but the added CI burden coupled with the existence of Python / nox for samples combined with MSRV preventing use of clap 3 means it's seeming easier to just fold back into nox. For now merged #2617 and will tidy this up another day.

davidhewitt:
Superseded by #2634

PR #2621, opened 2022-09-14
Add FeOs to examples

prehner:
FeOs makes extensive use of PyO3 to provide python bindings to all Rust functionalities. It also provides an example on how to organize PyO3 features across multiple crates.


mejrs:
Thanks! I understand little of it but it looks like an interesting project. Would you be able to rebase and fix the merge conflict?

PR #2622, opened 2022-09-14
Some improvements to __richcmp__ on enums

### LINKED ISSUE #2618: __ne__ always returns True for enums
### Bug Description

pyo3 provides a default __richcmp__ implementation for enums which handles `CompareOp::Eq` but not `CompareOp::Ne`, instead returning `NotImplemented`. However, this means Python thinks it the two types cannot be compared for equality, and so always returns `True`.

### Steps to Reproduce

Here is a sample (currently failing) test case for [`tests/test_enum.rs`](https://github.com/PyO3/pyo3/blob/main/tests/test_enum.rs).

```rust
#[test]
fn test_enum_ne() {
    Python::with_gil(|py| {
        let var1 = Py::new(py, MyEnum::Variant).unwrap();
        let var2 = Py::new(py, MyEnum::Variant).unwrap();
        py_assert!(py, var1 var1, "(var1 != var1) == False"); // Same id(_), so doesn't call __ne__
        py_assert!(py, var1 var2, "(var1 != var2) == False"); // Calls __ne__, returns True instead.
    })
}
```

### Backtrace

_No response_

Linux 5.10.16.3-microsoft-standard-WSL2 (Ubuntu)

Python 3.10

rustc 1.64.0-nightly (f6f9d5e73 2022-08-04)

### Your PyO3 version

Built from trunk (7b3ad2b7189fcf003c8d6e9bcee8d1bddcab7d28)

apt. No virtual env.

### Additional Info

Not 100% relevant, but here's some pure-Python code just to demonstrate the current behaviour. It's definitely not what I was expecting!

```python
class Eq:
    def __eq__(self, other):
        return True

    def __ne__(self, other):
        return NotImplemented

print(Eq() == Eq()) # True
print(Eq() != Eq()) # True
```

SquidDev:
- Implement `__ne__` as well as `__eq__`. Fixes #2618.
 - Return NotImplemented when types cannot be converted, rather than throwing.
 - Compare the integer ids inside the __eq__/__ne__ implementation. Previously a match block was generated - this is now closer to what `#[derive(PartialEq)]` would generate.

### DIFF
--- a/newsfragments/2622.fixed.md
+++ b/newsfragments/2622.fixed.md
@@ -1,0 +1,1 @@
+Correctly implement `__richcmp__` for enums, fixing `__ne__` returning always returning `True`.
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -526,13 +526,6 @@ fn impl_enum_class(
     };
 
     let (default_richcmp, default_richcmp_slot) = {
-        let variants_eq = variants.iter().map(|variant| {
-            let variant_name = variant.ident;
-            quote! {
-                (#cls::#variant_name, #cls::#variant_name) =>
-                    Ok(true.to_object(py)),
-            }
-        });
         let mut richcmp_impl: syn::ImplItemMethod = syn::parse_quote! {
             fn __pyo3__richcmp__(
                 &self,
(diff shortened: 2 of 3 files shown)


SquidDev:
I don't _believe_ the build failure here is due to this change. From what I can tell the same failure has happened on a couple of other PRs.

adamreichold:
> I don't _believe_ the build failure here is due to this change. From what I can tell the same failure has happened on a couple of other PRs.

Indeed. But it should also be fixed on the current `main`. Could you rebase your PR?

davidhewitt (commented):
Thanks, this looks good to me, though it needs a news entry please - `newsfragments/2622.fixed.md`.

davidhewitt (approved):

SquidDev:
Thank you!

PR #2628, opened 2022-09-19
Enforce PR changelog entry

messense:
Trying out https://github.com/pllim/actions-towncrier-changelog


messense:
The https://github.com/pllim/actions-towncrier-changelog action does not good enough, will probably try other things later.

PR #2631, opened 2022-09-21
Relax extract_sequence to the status quo before 0.17 was released

### LINKED ISSUE #2615: `val` ndarray object cannot be converted to `Sequence`

adamreichold:
This is the part that should be relative uncontroversial and fit to release as part of 0.17.2.

Closes #2615


adamreichold:
Not sure the coverage drop by more than 18% is real...

davidhewitt (approved):
Thanks!

Issue #2482, opened 2022-06-26
implicit conversions


hiaselhans:
Thanks for the great work you do on PyO3.

Pybind11 has a feature called 'implicit conversions'
https://pybind11.readthedocs.io/en/stable/advanced/classes.html#implicit-conversions

This can be quite handy, let me give an example:
``` rust
#[pyclass]
#[derive(Clone, Copy)]
pub struct Vector3D {
    pub x: f64
    pub y: f64
    pub z: f64
}

#[pymethods]
impl Vector3D {
    #new
    pub fn __new__(v: [f64, 3]) -> Self {
        Self {x: v[0], y: v[1], z: v[2]}
    }
}


#[pyfunction]
fn do_something(vec: Vector3D) -> f64 {
    vec.x * 2
}
```

``` python
>>> from py_rust_module import *

>>> do_something([1,2,3])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument 'vec': 'list' object cannot be converted to 'Vector3D'

>>> do_something(Vector3D([1,2,3]))
2
```

In Pybind11 Vector3D could be declared as implicitly convertible from list and `do_something([1,2,3])` would be valid and implicitly casting the list to a Vector3D object.

Is this something that could be done in pyo3? It is quite handy and can save a lot of verbosity...

davidhewitt:
Hi, thanks for a great question. I'm aware of the `pybind11` implicit conversions and agree they can be very powerful. It's a good topic of discussion whether they should be implemented in PyO3.

One thing to note is that Rust as a language chose not to have any implicit conversions. So if you view packages created by PyO3 as exposing Rust APIs then it may not be relevant to consider implicit conversions. Similarly `pybind11` supports C++ function overloading, but I don't think overloading would be a good fit for PyO3 at all.

In my opinion Python as a language is also moving away from duck-typed functions to static typing. So the fit for implicit conversions on the Python side is perhaps less than it was when `pybind11` was designed.

When I last considered it myself I didn't think we need this right now and was undecided on whether I'd want it in the long term. If a majority of users would like to have this feature, we can of course consider it.

davidhewitt:
[...]

PR #2632, opened 2022-09-21
Further relax extract_sequence for the upcoming release 0.18

adamreichold:
These are the two ~~less~~ more controversial parts, i.e. turning anything that can be iterated into a `Vec` whether it calls itself a sequence or not and also accepting `str` where the changelog entry gives a concrete hint on how to handle the `str`-as-`Vec<String>` edge case ergonomically as is common in Python API.

### DIFF
--- a/newsfragments/2632.changed.md
+++ b/newsfragments/2632.changed.md
@@ -1,0 +1,11 @@
+`impl FromPyObject for Vec<T>` will accept any Python object that can be turned into an iterator via Python's built-in `iter` function. It will also not reject `str` anymore which can be iterated as a sequence of `str` objects, i.e. [#2500](https://github.com/PyO3/pyo3/pull/2500) was reverted. Please use a type like
+
+```rust
+#[derive(FromPyObject)]
+pub enum OneOrMany<'py> {
+    One(&'py PyString),
+    Many(Vec<&'py PyString>),
+}
+```
+
+if you would like to work around this edge case for callers of your API.
--- a/pytests/tests/test_sequence.py
+++ b/pytests/tests/test_sequence.py
@@ -17,8 +17,7 @@ def test_vec_from_bytes():
 
 
 def test_vec_from_str():
-    with pytest.raises(TypeError):
-        sequence.vec_to_vec_pystring("123")
+    assert sequence.vec_to_vec_pystring("123") == ["1", "2", "3"]
 
 
 @pytest.mark.skipif(
--- a/src/types/sequence.rs
+++ b/src/types/sequence.rs
@@ -1,11 +1,10 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 use crate::err::{self, PyDowncastError, PyErr, PyResult};
-use crate::exceptions::PyTypeError;
 use crate::inspect::types::TypeInfo;
 use crate::internal_tricks::get_ssize_index;
 use crate::once_cell::GILOnceCell;
 use crate::type_object::PyTypeInfo;
-use crate::types::{PyAny, PyList, PyString, PyTuple, PyType};
+use crate::types::{PyAny, PyList, PyTuple, PyType};
 use crate::{ffi, PyNativeType, ToPyObject};
 use crate::{AsPyPointer, IntoPy, IntoPyPointer, Py, Python};
 use crate::{FromPyObject, PyTryFrom};
@@ -284,9 +283,6 @@ where
     T: FromPyObject<'a>,
 {
     fn extract(obj: &'a PyAny) -> PyResult<Self> {
-        if let Ok(true) = obj.is_instance_of::<PyString>() {
-            return Err(PyTypeError::new_err("Can't extract `str` to `Vec`"));
-        }
         extract_sequence(obj)
     }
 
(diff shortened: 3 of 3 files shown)


davidhewitt:
So I'm still unsure about this. I think the removal of the `str` special-case handling could be included in 0.17.2 (it was added in 0.17.0). So that could be separated out too if you feel strongly that was incorrect.

The broader relax... is it too much? Passing e.g. a `dict` will then get a `Vec` of the keys. Is that helpful? I'm not sure. All opinions on this change would be welcome.

adamreichold:
> So I'm still unsure about this. I think the removal of the `str` special-case handling could be included in 0.17.2 (it was added in 0.17.0). So that could be separated out too if you feel strongly that was incorrect.

Personally, my primary motivation to change anything here is making NumPy users happy and that we already did. This follow-up is mainly about furthering the discussion on what our approach to type extraction should be.

I like the general approach of just requiring iterability and dislike the str special case. But I will certainly not throw a tantrum if we do not change either of these. 
 
> The broader relax... is it too much? Passing e.g. a `dict` will then get a `Vec` of the keys. Is that helpful? I'm not sure. All opinions on this change would be welcome.

I would also very much like a overview of the opinions of interested users and/or the other maintainers.

I am cc'ing the people already involved in the original issue trying to extract `Vec` from NumPy arrays: @mtreinish @ritchie46 @aganders3

adamreichold:
> Passing e.g. a dict will then get a Vec of the keys. Is that helpful? I'm not sure.

I have to admit that I consider that `iter(dict)` yields an iterator of the keys instead of the key-value pairs to be a historical Python edge case that is similar to the `str`-can-be-iterated-as-`str` thing. Getting a `Vec` of key-value pairs would definitely be useful IMHO but is ruled out the existing Python behaviour.

mejrs:
I haven't really been keeping up with this sequence stuff, but...

> Passing e.g. a `dict` will then get a `Vec` of the keys. Is that helpful?

...sounds like a big footgun. That's not what I'd expect when asking pyo3 to convert things into a Vec (rather, I'd expect a typeerror)

davidhewitt:
I'm still undecided what the right thing to do is. It should be noted that what we do is currently almost the same as [pybind11](https://github.com/pybind/pybind11/blob/0694ec6a15863bff2e0ea5efe07c78de39b9a33c/include/pybind11/stl.h#L169-L171) does, which is a plus for ecosystem consistency in my opinion.

Arguments for:
- Allowing any Python iterable has potential performance advantages by avoiding the intermediate Python collection.
- Rust `Vec` has to be created anyway as part of this conversion, so requiring a concrete Python type also seems wasteful.

Arguments against:
- It's a breaking change (many PyO3 APIs accept more types).
- I think it's harder to take this away again later if we decide it was an error to do so. (We saw how much breakage we got just from not accepting `numpy` arrays by accident; imagine if we stopped accepting all iterables how many callsites could break.)
- Semantically, Rust's `Vec` is roughly akin to Python `Sequence`, so the existing behaviour does seem correct from a philosophical angle.

I wonder if rather than changing the existing behaviour, it's better if we could add a type `PyIterable<T>` which can then have easy iteration access? Very similar to our existing `PyIterator` but with type safety.

(I've been wondering for a while about experimenting with a new crate which provides generic variants of existing PyO3 types, e.g. `PyList<T>`.)

adamreichold:
> Semantically, Rust's Vec is roughly akin to Python Sequence, so the existing behaviour does seem correct from a philosophical angle.

Personally, I think of `FromPyObject::extract` as constructing a Rust type based on a Python value rather than establishing type level equivalence between Rust and Python. I think that also fits with the result of that operation usually being independent of the Python value it originated from. Although we admittedly have it perform "double duty" for argument conversion purposes. From this point of view, `<Vec as FromPyObject>::extract` would be similar to Python's `list(iterable)` constructor.  

> I wonder if rather than changing the existing behaviour, it's better if we could add a type PyIterable<T> which can then have easy iteration access? Very similar to our existing PyIterator but with type safety.

I think this is almost orthogonal to the question posed here, i.e. if I want to work with `Vec<T>` in the end, taking `vals: &PyAny` and calling `let vals = vals.iter()?.map(|val| val.extract()).collect<PyResult<Vec<T>>>()?;` would almost have the same ergonomics IMHO.

> I'm still undecided what the right thing to do is.

All that being said, I think that we should drop the change if there is this much uncertainty attached to it. I fear that it takes up scarce review bandwidth that is needed elsewhere and the status quo certainly serves us insofar there is no stream of bug reports on this behaviour.

davidhewitt:
Ok. I am always sorry to see a contribution not be merged, as there is always an underlying motivation as well as time invested.

> Although we admittedly have it perform "double duty" for argument conversion purposes. From this point of view, `<Vec as FromPyObject>::extract` would be similar to Python's `list(iterable)` constructor.

I think you've hit the nail on the head here, and this perhaps suggests the avenue for future exploration another time. There are different levels of conversion which users can want on function inputs. The current PyO3 semantics are usually to accept the equivalent Python type - which I'll call "narrow" conversion range. The proposal here is more akin to "wide" conversion range. Implicit conversions like those asked about in #2482 also are flavours of "wide" conversions.

I would say `pybind11` defaults to "wide" conversions and [has a `.noconvert()` option](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#non-converting-arguments) to reduce to "narrow" mode.

It is easier for PyO3 users to use newtypes and `#[pyo3(from_py_with = ...)]` to modify conversion behaviour already, so I think there's less urgency to support something like `.noconvert()`. At the same time, these sorts of proposals do suggest there is a wish for more. We can definitely continue to be open to better designs to offer in the framework.

PR #2633, opened 2022-09-21
pyclass: better error and explanation why lifetimes are disallowed

davidhewitt:
After #2624 felt that there was a potentially easy win here to add some documentation to explain why lifetimes aren't allowed with `#[pyclass]`.


adamreichold (approved):

guide/src/class.md
```diff
@@ -50,10 +50,24 @@ enum MyEnum {
...
-
 The above example generates implementations for [`PyTypeInfo`] and [`PyClass`] for `MyClass` and `MyEnum`. To see these generated implementations, refer to the [implementation details](#implementation-details) at the end of this chapter.
 
+### Restrictions
+
+To integrate Rust types with Python, PyO3 needs to place some restrictions on the types which can be annotated with `#[pyclass]`. In particular, they must have no lifetime parameters, no generic parameters, and must implement `Send`. The reason for each of these is explained below.
+
+#### No lifetime parameters
+
+As soon as Rust data crosses into Python, there is no guarantee which the Rust compiler can make on how long the data will live. The only possible way to express this correctly is to require that no data in the struct is borrowed for any lifetime shorter than the `'static` lifetime, i.e. the struct cannot have any lifetime parameters.
```
mejrs, line 61: I'd like to add that Rust lifetimes are compile-time only; they cannot be tracked "at runtime". It's a misconception I see people have occasionally.

Also it might be nice to point people to alternative solutions here. Perhaps they can use `Arc` and maybe `Weak`?

mejrs (approved):
Thanks 👍

guide/src/class.md
birkenfeld: :+1:

davidhewitt: Good idea, added 👍

PR #2634, opened 2022-09-21
remove need to build xtask from CI

davidhewitt:
After discussions in #2619 I think we're leaning towards preferring `nox` over `xtask` for simplicity. This moves the coverage implementation from `xtask` to `nox`, and also removes `xtask` from the main CI runs. Hopefully shortens CI build times a bit.

Potentially a follow-up to move all of the `xtask ci` logic to a `nox -s ci` job.


noxfile.py
```diff
@@ -230,3 +271,39 @@ def _get_rust_target() -> str:
...
+    "--package=pyo3",
+    "--package=pyo3-build-config",
+    "--package=pyo3-macros-backend",
+    "--package=pyo3-macros",
+    "--package=pyo3-ffi",
+]
+
+
+def _cargo_test_package(package: str) -> List[str]:
+    return ["cargo", "test", "--manifest-path", f"{package}/Cargo.toml"]
```
adamreichold, line 287: Is there a particular reason to use `--manifest-path` instead of `--package` here?

adamreichold:
I think the grouping does not work as intended yet?

image

mejrs (approved):

noxfile.py
```diff
@@ -1,34 +1,57 @@
...
+    _run(session, "nox", "-f", "pytests/noxfile.py", env=env, external=True)
     for example in glob("examples/*/noxfile.py"):
-        session.run("nox", "-f", example, external=True)
+        _run(session, "nox", "-f", example, env=env, external=True)
+
+
+@nox.session(venv_backend="none")
+def coverage(session: nox.Session) -> None:
+    env = _get_coverage_env()
+    _run("cargo", "llvm-cov", "clean", "--workspace", env=env, external=True)
```
adamreichold, line 43:
```suggestion
    _run(session, "cargo", "llvm-cov", "clean", "--workspace", env=env, external=True)
```

noxfile.py
```diff
@@ -230,3 +249,47 @@ def _get_rust_target() -> str:
...
+
+
+def _run(session: nox.Session, *args: str, **kwargs: Any) -> None:
+    """Wrapper for _run(session, which creates nice groups on GitHub Actions."""
+    if "GITHUB_ACTIONS" in os.environ:
+        # Insert ::group:: at the start of nox's command line output
+        print("::group::", end="", flush=True)
+    session.run(*args, **kwargs)
+    if "GITHUB_ACTIONS" in os.environ:
+        print("::endgroup::")
```
messense, line 276: image

https://github.com/PyO3/pyo3/actions/runs/3103635220/jobs/5027187965#step:7:10

Looks odd, perhaps `::group::` should have a description like `::group::Running xxx`?

davidhewitt: It should be using nox's output from the next line but I think it's because nox's command output goes to stderr it's not lining up correctly in CI. I'm going to try printing to stderr.

davidhewitt: Yep that looks like it works much better now.

noxfile.py
davidhewitt: Just didn't realise `--package` would work! Have changed to use it 😄

messense (approved):

adamreichold (approved):

PR #2635, opened 2022-09-22
Add git repository and edit urls to user guide

messense:
https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options

image

### DIFF
--- a/guide/book.toml
+++ b/guide/book.toml
@@ -5,3 +5,7 @@ author = "PyO3 Project and Contributors"
 
 [preprocessor.pyo3_version]
 command = "python3 guide/pyo3_version.py"
+
+[output.html]
+git-repository-url = "https://github.com/PyO3/pyo3/tree/main/guide"
+edit-url-template = "https://github.com/PyO3/pyo3/edit/main/guide/{path}"


guide/book.toml
```diff
@@ -5,3 +5,7 @@ author = "PyO3 Project and Contributors"
 
 [preprocessor.pyo3_version]
 command = "python3 guide/pyo3_version.py"
+
+[output.html]
+git-repository-url = "https://github.com/PyO3/pyo3/tree/main/guide"
```
adamreichold, line 10: Is it customary to link into the tree here instead of just linking the PyO3 repository?

messense: mdbook also link into the tree: https://github.com/rust-lang/mdBook/blob/c8db0c8ec65e9ff8e82c580806a7c5dc03774352/guide/book.toml#L13

Psykopear:
Sorry, opened a PR with the fix: #2636

davidhewitt (approved):
Great idea!

CI failure looks related to `maturin-action`?

messense:
Rust 1.64.0 requires GLIBC 2.17+, so we need to change `manylinux: auto` in `ci.yml` to `manylinux: 2014`.

davidhewitt:
Thanks. I also pushed #2637 for the other pieces here

PR #2636, opened 2022-09-22
Fixed wrong proptest in chrono conversion module

Psykopear:
Two of the proptests in the `chrono` module were using an out of bound value to initialize a `FixedOffset` (see [here](https://docs.rs/chrono/latest/src/chrono/offset/fixed.rs.html#54-60)) , this slipped during review.


messense (approved):

davidhewitt (approved):
Thanks!

PR #2637, opened 2022-09-22
rust: updates for 1.64

davidhewitt:


.cargo/config
```diff
@@ -3,6 +3,7 @@ xtask = "run --package xtask --"
 
 [target.'cfg(feature = "cargo-clippy")']
 rustflags = [
+    "-Aclippy::needless_borrow", # TODO: remove once msrv increased from 1.48
```
davidhewitt, line 6: This lint started asking to remove `&` from `&[x, y, z]` slices in a lot of places which were incompatible with MSRV :(

pyo3-ffi/src/cpython/unicodeobject.rs
```diff
@@ -137,11 +137,7 @@ pub unsafe fn PyUnicode_IS_COMPACT(op: *mut PyObject) -> c_uint {
 #[inline]
 #[cfg(target_endian = "little")]
 pub unsafe fn PyUnicode_IS_COMPACT_ASCII(op: *mut PyObject) -> c_uint {
-    if (*(op as *mut PyASCIIObject)).ascii() != 0 && PyUnicode_IS_COMPACT(op) != 0 {
-        1
-    } else {
-        0
-    }
+    ((*(op as *mut PyASCIIObject)).ascii() != 0 && PyUnicode_IS_COMPACT(op) != 0) as c_uint
```
adamreichold, line 140: Wouldn't `.into()` work here as well? I think `Into` and `From` are generally preferable to casts where applicable?

davidhewitt: Sure, changed to into 👍

adamreichold (approved):

PR #2639, opened 2022-09-24
docs: fixed typo in error_handling.md

mrob95:
the sample code currently produces this error:
```
15 |     fn from(error: MyOtherError) -> Self {
   |        ---- this function doesn't have a `self` parameter
16 |         PyValueError::new_err(self.0.message())
   |                               ^^^^ `self` value is a keyword only available in methods with a `self` parameter
```


mejrs:
Is this part of the docs not tested? 🤔

PR #2640, opened 2022-09-24
docs: Fix pyo3-build-config dependency advice in guide

smheidrich:
Two things are wrong with the [guide's advice on how to get `#[cfg]` flags from `pyo3-build-config`](https://pyo3.rs/v0.17.1/building_and_distribution/multiple_python_versions.html#using-pyo3-build-config):

1. The dependency's `version = ...` specification to be put in `Cargo.toml` is enclosed in double quotes instead of curly braces.
2. Since https://github.com/PyO3/pyo3/pull/1856, [`use_pyo3_cfgs()` is only available if the `resolve-config` feature is enabled](https://github.com/PyO3/pyo3/blob/86ce4d1a137d6d62df84c7be2935078e08e46750/pyo3-build-config/src/lib.rs#L45), which isn't mentioned.

This PR fixes both of these.

### DIFF
--- a/guide/src/building_and_distribution/multiple_python_versions.md
+++ b/guide/src/building_and_distribution/multiple_python_versions.md
@@ -31,11 +31,11 @@ To see a full reference of all the `#[cfg]` flags provided, see the [`pyo3-build
 
 You can use the `#[cfg]` flags in just two steps:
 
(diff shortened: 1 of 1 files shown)


adamreichold (approved):

davidhewitt (approved):
Thanks!

PR #2646, opened 2022-09-29
add internal release notes

davidhewitt:
To accompany the 0.17.2 release, I thought I would document the release process so that we all know it and are aware of how it works, as it's also getting slightly more fiddly with the addition of `towncrier` :)

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -1,0 +1,56 @@
+# Releasing
+
+This is notes for the current process of releasing a new PyO3 version. Replace `<version>` in all instructions below with the new version.
+
+## 1. Prepare the release commit
+
+Follow the process below to update all required pieces to bump the version. All these changes are done in a single commit because it makes it clear to git readers what happened to bump the version. It also makes it easy to cherry-pick the version bump onto the `main` branch when tidying up branch history at the end of the release process.
+
+1. Replace all instances of the PyO3 current version with the new version to be released. Places to check:
+   - `Cargo.toml` for all PyO3 crates in the repository.
+   - PyO3 version embedded into documentation like the README.
+   - `pre-script.rhai` templates for the examples.
+   - `[towncrier]` section in `pyproject.toml`.
+
+   Places where you need to be a bit careful:
+   - Add _new_ lines to the netlify `_redirects` file rather than replacing the existing ones.
+   - Make sure not to modify the CHANGELOG during this step!
+
+2. Run `towncrier build` to generate the CHANGELOG. The version used by `towncrier` should automatically be correct because of the update to `pyproject.toml` in step 1.
+
+3. Manually edit the CHANGELOG for final notes. Steps to do:
+   - Adjust wording of any release lines to make them clearer for users / fix typos.
+   - Add a new link at the bottom for the new version, and update the `Unreleased` link.
+
+4. Create the commit containing all the above changes, with a message of `release: <version>`. Push to `release-<BRANCH_VER>` branch on the main PyO3 repository, where `<BRANCH_VER>` depends on whether this is a major or minor release:
+   - for O.X.0 minor releases, just use `0.X`, e.g. `release-0.17`. This will become the maintenance branch after release.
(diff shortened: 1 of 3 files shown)


Releasing.md
```diff
@@ -0,0 +1,48 @@
...
+## 3. Leave for a cooling off period
+
+Wait a couple of days in case anyone wants to hold up the release to add bugfixes etc.
+
+## 4. Put live
+
+To put live:
+- 1. run `nox -s publish` to put live on crates.io
+- 2. publish the release on Github
+- 3. merge the release PR
```
adamreichold, line 44: I noticed that we have quite a branches like `release-<major>.<minor>.<patch>` lying around in the repository. Aren't those redundant to the `v<major>.<minor>.<patch>` tags and could be delete in this step? (In contrast to the `release-<major>.<minor>` branches for diverging maintenance between the current and the next release on `main`.)

Releasing.md
```diff
@@ -0,0 +1,48 @@
...
+
+4. Create the commit containing all the above changes, with a message of `release: <version>`. Push to `release-<version>` branch on the main PyO3 repository.
+
+## 2. Create the release PR and draft release notes
+
+Open a PR for the branch, and confirm that it passes CI.
+
+On https://github.com/PyO3/pyo3/releases, click "Draft a new release". The tag will be a new tag of `v<version>` (note preceding `v`) and target should be the `release-<version>` branch you just pushed.
+
+Write release notes which match the style of previous releases. You can get the list of contributors by running `nox -s contributors -- v<prev-version> release-<version>` to get contributors from the previous version tag through to the branch tip you just pushed.
```
adamreichold, line 31: Maybe it would be good to slightly reword this to indicate that the Nox session will only _after_ pushing to GitHub as it uses the GitHub API instead of the local repository state?

Releasing.md
```diff
@@ -0,0 +1,48 @@
...
+   - Add _new_ lines to the netlify `_redirects` file rather than replacing the existing ones.
+   - Make sure not to modify the CHANGELOG during this step!
+
+2. Run `towncrier build` to generate the CHANGELOG. The version used by `towncrier` should automatically be correct because of the update to `pyproject.toml` in step 1.
+
+3. Manually edit the CHANGELOG for final notes. Steps to do:
+   - Adjust wording of any release lines to make them clearer for users / fix typos.
+   - Add a new link at the bottom for the new version, and update the `Unreleased` link.
+
+4. Create the commit containing all the above changes, with a message of `release: <version>`. Push to `release-<version>` branch on the main PyO3 repository.
```
messense, line 23: We can add a check to the existing changelog CI to ensure that `newsfragrments` directory contains no news entries when PR title starts with `release: `.

Releasing.md
davidhewitt: Yes, good idea to clean these up.

I'd been keeping the maintenance branches of each version as 0.14, 0.15, 0.16, etc.

I don't have a strong opinion whether it's better to call the maintenance branch `0.14` or `release-0.14`, any option you prefer? I can also add instructions here to create the maintenance branch if pushing a new breaking release.

Releasing.md
davidhewitt: Good idea, done 👍

Releasing.md
adamreichold: > I don't have a strong opinion whether it's better to call the maintenance branch 0.14 or release-0.14,

I have no strong opinion but I would be glad if we use consistent names.

> I can also add instructions here to create the maintenance branch if pushing a new breaking release.

That would be nice. Thanks!

davidhewitt: Let's go for `release-0.x`. That way the created branch from the PR can stick around as the maintenance branch when needed. I'll also tidy up the existing set of branches.

messense (approved):

Releasing.md
davidhewitt: Updated and cleaned up existing branches.

PR #2657, opened 2022-10-04
Update decorator to use atomics

mejrs:
Fixes https://github.com/PyO3/pyo3/discussions/2598

### DIFF
--- a/examples/decorator/src/lib.rs
+++ b/examples/decorator/src/lib.rs
@@ -1,14 +1,16 @@
 use pyo3::prelude::*;
 use pyo3::types::{PyDict, PyTuple};
+use std::sync::atomic::{AtomicU64, Ordering};
 
 /// A function decorator that keeps track how often it is called.
 ///
 /// It otherwise doesn't do anything special.
 #[pyclass(name = "Counter")]
 pub struct PyCounter {
-    // We use `#[pyo3(get)]` so that python can read the count but not mutate it.
-    #[pyo3(get)]
-    count: u64,
+    // Keeps track of how many calls have gone through.
+    //
+    // See the discussion at the end for why an atomic is used.
+    count: AtomicU64,
 
     // This is the actual function being wrapped.
     wraps: Py<PyAny>,
@@ -23,20 +25,28 @@ impl PyCounter {
     //    2. We still need to handle any exceptions that the function might raise
     #[new]
     fn __new__(wraps: Py<PyAny>) -> Self {
-        PyCounter { count: 0, wraps }
+        PyCounter {
+            count: AtomicU64::new(0),
+            wraps,
+        }
+    }
+
+    #[getter]
+    fn count(&self) -> u64 {
+        self.count.load(Ordering::Relaxed)
     }
 
     #[args(args = "*", kwargs = "**")]
     fn __call__(
-        &mut self,
+        &self,
         py: Python<'_>,
         args: &PyTuple,
         kwargs: Option<&PyDict>,
     ) -> PyResult<Py<PyAny>> {
-        self.count += 1;
+        let old_count = self.count.fetch_add(1, Ordering::Relaxed);
         let name = self.wraps.getattr(py, "__name__")?;
 
-        println!("{} has been called {} time(s).", name, self.count);
+        println!("{} has been called {} time(s).", name, old_count + 1);
 
         // After doing something, we finally forward the call to the wrapped function
         let ret = self.wraps.call(py, args, kwargs)?;
--- a/examples/decorator/src/lib.rs
+++ b/examples/decorator/src/lib.rs
@@ -1,6 +1,6 @@
 use pyo3::prelude::*;
 use pyo3::types::{PyDict, PyTuple};
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::cell::Cell;
 
 /// A function decorator that keeps track how often it is called.
 ///
@@ -9,8 +9,8 @@ use std::sync::atomic::{AtomicU64, Ordering};
 pub struct PyCounter {
     // Keeps track of how many calls have gone through.
     //
-    // See the discussion at the end for why an atomic is used.
-    count: AtomicU64,
+    // See the discussion at the end for why `Cell` is used.
+    count: Cell<u64>,
 
     // This is the actual function being wrapped.
     wraps: Py<PyAny>,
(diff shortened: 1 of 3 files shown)


guide/src/class/call.md
```diff
@@ -65,3 +65,51 @@ def Counter(wraps):
...
+```
+
+The implementation in this chapter fixes that by never borrowing exclusively; all the methods take `&self` as receivers, of which multiple may exist simultaneously. This requires a shared and threadsafe counter and the easiest way to do that is to use atomics, so that's what is used here.
+
+This shows the dangers of running arbitrary Python code - note that "running arbitrary Python code" can be far more subtle than the example above:
+- Python's asynchronous executor may park the current thread in the middle of Python code, even in Python code that *you* control, and let other Python code run.
+- Dropping arbitrary Python objects may invoke destructors defined in Python (`__del__` methods).
+- Calling Python's C-api (most PyO3 apis call C-api functions internally) may raise exceptions, which may allow Python code in signal handlers to run.
+
+This especially important if you are writing unsafe code; Python code must never be able to cause undefined behavior. You must ensure that your Rust code is in a consistent state before doing any of the above things.
```
adamreichold, line 113: > Python code must never be able to cause undefined behavior.

I am not sure that this a tenable position as Python code can already cause undefined behaviour (like data races) even without Rust code being involved, or to more succinct: Python code is unsafe/trusted, not safe/verified like safe Rust.

guide/src/class/call.md
```diff
@@ -65,3 +65,51 @@ def Counter(wraps):
...
+@Counter
+def say_hello():
+    if say_hello.count < 2:
+        print(f"hello from decorator")
+        
+say_hello()
+# RuntimeError: Already borrowed
+```
+
+The implementation in this chapter fixes that by never borrowing exclusively; all the methods take `&self` as receivers, of which multiple may exist simultaneously. This requires a shared and threadsafe counter and the easiest way to do that is to use atomics, so that's what is used here.
```
adamreichold, line 106: I think a `Cell<u64>` is sufficient to implement interior mutability here: The calls are protected by the GIL and it compiles and runs with a `Cell<u64>`, so either PyO3 is doing something wrong or thread safety is not actually required?

adamreichold (changes requested):
I think the function only needs to be reentrant which only needs `Cell` and friends, not atomics.

guide/src/class/call.md
mejrs: Oh that's true; I forgot that `Cell<u64>` is `Send` and we don't require pyclasses to be `Sync`.

adamreichold (approved):

guide/src/class/call.md
mejrs: I've always thought of the Python <-> C boundary as the same as safe Rust <-> unsafe Rust; C code can never trust python code, and Python code must never be able to cause undefined behavior.

Obviously there are some exceptions like `ctypes`, which is really just an escape hatch on the python side.

adamreichold: At least NumPy does not seem handle it that way, e.g.

```python
from threading import Thread

import numpy as np

def increment_array():
    global array

    for i in range(10_000):
        array += 1

array = np.zeros((100, 100))

threads = [Thread(target=increment_array) for _ in range(10)]
for thread in threads: thread.start()
for thread in threads: thread.join()

assert np.sum(array) == 10 * 10_000 * 100 * 100
```

sometimes passes, sometimes fails presumably because NumPy releases the GIL while doing `array += 1`.

I think there are similar issues in Python's standard library where the implementation of types releases the GIL before doing system calls which should also allow constructing such data races, but it is admittedly harder to do. (For example, a simple counter does not suffice as it only creates a logical race if the update is not done "GIL atomically" using e.g. `counter += 1`, but not a data race AFAIU.)

guide/src/class/call.md
```diff
@@ -37,7 +37,7 @@ say_hello has been called 4 time(s).
 hello
 ```
 
-#### Pure Python implementation
+### Pure Python implementation
```
davidhewitt, line 40: Was the `Example: ` heading above meant to just be `##`, perhaps ?

examples/decorator/src/lib.rs
```diff
@@ -1,14 +1,16 @@
...
 /// It otherwise doesn't do anything special.
 #[pyclass(name = "Counter")]
 pub struct PyCounter {
-    // We use `#[pyo3(get)]` so that python can read the count but not mutate it.
-    #[pyo3(get)]
-    count: u64,
+    // Keeps track of how many calls have gone through.
+    //
+    // See the discussion at the end for why `Cell` is used.
+    count: Cell<u64>,
```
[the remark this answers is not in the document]
mejrs: It seems nice for `Cell` - not sure about the others. Maybe Box, if it works out. I think `Mutex` is too tricky.

guide/src/class/call.md
adamreichold: > IIRC there's a cap on the number of bytecode operations before the GIL is released to give other threads a chance to make progress.

I don't think this applies here, i.e. the disassembly of `increment_array` is

```
  8           0 LOAD_GLOBAL              0 (range)
              2 LOAD_CONST               1 (10000)
              4 CALL_FUNCTION            1
              6 GET_ITER
        >>    8 FOR_ITER                 6 (to 22)
             10 STORE_FAST               0 (i)

  9          12 LOAD_GLOBAL              1 (array)
             14 LOAD_CONST               2 (1)
             16 INPLACE_ADD
             18 STORE_GLOBAL             1 (array)
             20 JUMP_ABSOLUTE            4 (to 8)

  8     >>   22 LOAD_CONST               0 (None)
             24 RETURN_VALUE
```

meaning that `INPLACE_ADD` which will call into NumPy eventually is a single bytecode operation.

davidhewitt: Ah I see what you mean; yes, that's scary, looks like `numpy` does indeed allow a data race while doing the increment!

adamreichold: NumPy does it when only primitive types are involved which can be access without involving the interpreter (which is the case in the above example), c.f. the [`NPY_METH_REQUIRES_PYAPI` flag](https://github.com/numpy/numpy/blob/26d690e5abe9dc7eb74deb807ea2c4842aa91c70/numpy/core/src/multiarray/array_method.h#L16) and its usage in the [low-level implementation details](https://github.com/numpy/numpy/blob/26d690e5abe9dc7eb74deb807ea2c4842aa91c70/numpy/core/src/multiarray/lowlevel_strided_loops.c.src#L1571).

mejrs: @adamreichold do you strongly disagree with the wording? I personally think it's fine to say this, especially in a Rust context, and I'd like to merge this PR.

adamreichold: I do _not_ strongly disagree would not oppose merging this. (I opted to approve the PR without resolving this conversation thereby trying to express this "I care about this but not enough to block the PR over it." position.)

PR #2663, opened 2022-10-05
[experimental] make better errors

mejrs:
This uses a clever hack to turn on `rustc_on_unimplemented` if available (this seems to be rather brittle, which is how I discovered it. So I'm not a fan of actually doing this).

Still, I'm sharing this to see and share how we can improve errors, and I'm also curious for any ideas you might have.

### DIFF
--- a/tests/ui/invalid_pymethod_receiver.stderr
+++ b/tests/ui/invalid_pymethod_receiver.stderr
@@ -1,18 +1,32 @@
-error[E0277]: the trait bound `i32: From<&PyCell<MyClass>>` is not satisfied
+error[E0277]: `i32` is not a valid receiver type for pyo3 methods
  --> tests/ui/invalid_pymethod_receiver.rs:8:43
   |
 8 |     fn method_with_invalid_self_type(slf: i32, py: Python<'_>, index: u32) {}
-  |                                           ^^^ the trait `From<&PyCell<MyClass>>` is not implemented for `i32`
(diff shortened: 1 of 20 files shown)


tests/ui/invalid_pymethod_receiver.stderr
```diff
@@ -1,18 +1,32 @@
...
-            <f64 as From<i32>>
-            <f64 as From<i8>>
-          and 67 others
-  = note: required because of the requirements on the impl of `Into<i32>` for `&PyCell<MyClass>`
-  = note: required because of the requirements on the impl of `TryFrom<&PyCell<MyClass>>` for `i32`
+  = note: Only `&self`, `&mut self`, `&PyCell<Self>, `&PyRef<'_, Self> and `&PyRefMut<'_, Self> are valid receiver types
+  = help: the following other types implement trait `pyo3::pycell::Receiver<T>`:
+            &'a PyCell<T>
+            PyRef<'a, T>
+            PyRefMut<'a, T>
```
mejrs, line 11: We don't need `rustc_on_unimplemented` for this, and this is already miles better. Is there a reason why we're using `TryFrom` for this rather than our own trait?

davidhewitt: Agreed, I think at the time it was simpler for good ol' `#[pyproto]`. I have considered making a similar change in the past, so a PR for this would be welcome!

PR #2664, opened 2022-10-06
Don't use intocallback in method macros

mejrs:
This makes the errors *much* nicer.


src/callback.rs
```diff
@@ -134,6 +134,31 @@ pub trait WrappingCastTo<T> {
     fn wrapping_cast(self) -> T;
 }
 
+pub trait OkWrap<T> {
```
mejrs, line 137: this should probably be hidden/sealed

src/pyclass.rs
```diff
@@ -378,7 +380,6 @@ impl PyTypeBuilder {
             itemsize: 0,
             // `c_ulong` and `c_uint` have the same size
             // on some platforms (like windows)
-            #[allow(clippy::useless_conversion)]
```
mejrs, line 381: this isn't working for some reason

tests/ui/missing_intopy.stderr
```diff
@@ -1,45 +1,18 @@
...
-              <IterANextOutput<T, U> as IntoPyCallbackOutput<IterANextOutput<Py<PyAny>, Py<PyAny>>>>
-              <IterNextOutput<Py<PyAny>, Py<PyAny>> as IntoPyCallbackOutput<*mut PyObject>>
-              <IterNextOutput<T, U> as IntoPyCallbackOutput<IterNextOutput<Py<PyAny>, Py<PyAny>>>>
-            and 7 others
-note: required by a bound in `pyo3::callback::convert`
-   --> src/callback.rs
-    |
-    |     T: IntoPyCallbackOutput<U>,
-    |        ^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `pyo3::callback::convert`
-    = note: this error originates in the attribute macro `pyo3::pyfunction` (in Nightly builds, run with -Z macro-backtrace for more info)
```
mejrs, line 45: I'm *so* happy to get rid of this

davidhewitt: +1 :)

davidhewitt (commented):
Thanks! I guess this goes in "changed" section of changelog?

There's similar for the proto methods, do you think we can use a similar approach?

src/pyclass.rs
mejrs: Nope - still getting this on stable clippy.

mejrs:
> There's similar for the proto methods, do you think we can use a similar approach?
 
Can users still call any callback trait since pyproto was removed?

Also..ugh 
```
  = help: the following other types implement trait `IntoPy<T>`:
            <&'a OsString as IntoPy<Py<PyAny>>>
            <&'a Path as IntoPy<Py<PyAny>>>
            <&'a PathBuf as IntoPy<Py<PyAny>>>
            <&'a PyErr as IntoPy<Py<PyAny>>>
            <&'a String as IntoPy<Py<PyAny>>>
            <&'a [u8] as IntoPy<Py<PyAny>>>
            <&'a str as IntoPy<Py<PyAny>>>
            <&'a str as IntoPy<Py<PyString>>>
          and 172 others
```
How do we deal with this in CI again? These counts change depending on the features 😭 

I have filed https://github.com/rust-lang/rust/pull/102752 which should do away both with the horrible error message and the "and ... others" thing, so these errors should get even better :)

mejrs:
Okay, I think this works.

mejrs:
Ah, I have a better idea,

src/pyclass.rs
davidhewitt: Hmm if moving it up resolves, I guess we'll live with it.

tests/ui/invalid_result_conversion.stderr
```diff
@@ -1,30 +1,18 @@
...
+   = help: the following other types implement trait `From<T>`:
+             <PyErr as From<&CancelledError>>
+             <PyErr as From<&IncompleteReadError>>
+             <PyErr as From<&InvalidStateError>>
+             <PyErr as From<&LimitOverrunError>>
+             <PyErr as From<&PanicException>>
+             <PyErr as From<&PyArithmeticError>>
+             <PyErr as From<&PyAssertionError>>
+             <PyErr as From<&PyAttributeError>>
+           and $N others
```
davidhewitt, line 16: Aha nice find!

davidhewitt (commented):
Thanks, this is looking really good! Sorry for my extended delay on reply, my son and I have been ill over the weekend previous and I'm finally better and catching up with things.

> Can users still call any callback trait since pyproto was removed?

So eventually I would like to remove `src/callback.rs` and put everything in it into `impl_` details. Users shouldn't be using that stuff directly as far as I'm concerned.

For the `__dunder__` methods, the generated code is built up in pieces (because it needed to be flexible to support the variety of forms those dunders take).

The last uses of `callback::convert` which might be doing ok-wrapping are [here](https://github.com/PyO3/pyo3/blob/43261c7b8c7d9c703539662fa5ee47a2eabca0f7/pyo3-macros-backend/src/pymethod.rs#L1089) and [here](https://github.com/PyO3/pyo3/blob/43261c7b8c7d9c703539662fa5ee47a2eabca0f7/pyo3-macros-backend/src/pymethod.rs#L950).

If you want, we could save those for a separate PR?

mejrs:
> Thanks, this is looking really good! Sorry for my extended delay on reply, my son and I have been ill over the weekend previous and I'm finally better and catching up with things.

Get well soon 👍 

> If you want, we could save those for a separate PR?

I think that's be a good idea

davidhewitt (approved):
Cool, in which case let's merge this 👍

> Get well soon 👍

Thanks, I'm fine now but my son still getting over it.

PR #2666, opened 2022-10-09
Add `github-actions` to dependabot config

messense:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


davidhewitt:
👍 thanks for adding this (and handling the follow-ups)

Issue #2611, opened 2022-09-09
Reproducable crash with pyo3 and tensorflow


dbr:
### Bug Description

We encountered a very similar sounding issue to #2032 #1623 - but "luckily" for us the crash happens quite reliably!

It doesn't seem specific to tensorflow as it seems to crash similarly with some pytorch code

### Steps to Reproduce

Create simple project, and virtual env with tensorflow (pandas also included as it was in the original Python snippet, but I don't think it is relevant)

```
$ cargo new --bin pyo3_tf_crash
$ cd pyo3_tf_crash/
$ python -m venv _venv
$ source _venv/bin/activate
$ pip install tensorflow pandas
```

```
$ cat Cargo.toml
[package]
name = "pyo3_tf_crash"
version = "0.1.0"
edition = "2021"

[dependencies]
pyo3 = {version = "0.17.1", features = ["auto-initialize"]}
```

And finally, `src/main.rs`

```rust
use pyo3::prelude::*;

static CODE: &'static str = r#"import pandas
d = [
    {'status': 'a', 'title': 'title 1'},
    {'status': 'b', 'title': 'title 2'},
]

data = pandas.DataFrame(d)
sentiment_label = data.status.factorize()

maxlen = 30

from tensorflow.keras.preprocessing.text import Tokenizer
from tensorflow.keras.preprocessing.sequence import pad_sequences
from tensorflow.keras.models import Sequential
from tensorflow.keras.layers import LSTM,Dense, Dropout, SpatialDropout1D
from tensorflow.keras.layers import Embedding



tokenizer = Tokenizer(num_words=5000)
tokenizer.fit_on_texts(data.title.values)

encoded_docs = tokenizer.texts_to_sequences(data.title.values)
padded_sequence = pad_sequences(encoded_docs, maxlen=maxlen)

vocab_size = len(tokenizer.word_index) + 1



embedding_vector_length = 32
model = Sequential()
model.add(Embedding(vocab_size, embedding_vector_length, input_length=maxlen))
model.add(SpatialDropout1D(0.25))
model.add(LSTM(50, dropout=0.5, recurrent_dropout=0.5))
model.add(Dropout(0.2))
model.add(Dense(1, activation='sigmoid'))

model.compile(loss='binary_crossentropy',optimizer='adam', metrics=['accuracy'])

history = model.fit(padded_sequence, sentiment_label[0], validation_split=0.2, epochs=2, batch_size=32)
"#;

fn start_py_thread() {
[...]

PR #2675, opened 2022-10-11
Also relax the PySequence check when extracting fixed-sized arrays.

### LINKED ISSUE #2674: Cannot extract Rust arrray from NumPy array
I am getting _TypeError: argument 'oned': 'ndarray' object cannot be converted to 'Sequence'_  running test.py with the following files:

src/lib.rs
```
use pyo3::prelude::*;

#[pyclass(module = "rust_module")]
#[derive(Default, Debug, serde::Serialize, serde::Deserialize, Clone)]
pub struct DummyClass{
    name: String,
    oned: [f64;3],
    twod: [[f64;3];3],
}

#[pymethods]
impl DummyClass {
    #[new]
    pub fn pynew(
        name: String,
        oned: [f64;3],
        twod: [[f64;3];3],
    ) -> PyResult<Self>{
        Ok(Self { name, oned, twod})
    }

    fn __call__(&self) -> PyResult<String>{
        Ok("Success".into())
    }
}

#[pymodule]
fn pyo3_pure(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<DummyClass>()?;
    Ok(())
}

```

Cargo.toml

```
[package]
name = "reproduce-extract"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
pyo3 = { version = "0.17.2", features = ["serde", "abi3-py37", "extension-module"] }
numpy = { version = "^0.17.2" }
serde = { version = "^1.0", features = ["derive"]}

[lib]
name = "pyo3_pure"
crate-type = ["cdylib"]

```

pyproject.toml
```
[build-system]
requires = ["maturin>=0.13,<0.14"]
build-backend = "maturin"

[tool.maturin]
bindings = "pyo3"

[project]
# The name pyo3_pure instead of pyo3-pure is intentional,
# it's used as a crate name resolution regression test
name = "pyo3_pure"
version = "0.1.0"

dependencies = [
    "numpy"
]
```

test.py
```
from pyo3_pure import DummyClass
import numpy as np

# test = DummyClass("hello",[0,0,0],[[0,0,0],[0,0,0],[0,0,0]])
test = DummyClass("hello",np.array([0,0,0]),np.array([[0,0,0],[0,0,0],[0,0,0]]))

print(test())
```

_Originally posted by @Billyangnz in https://github.com/PyO3/pyo3/issues/2615#issuecomment-1273884897_

adamreichold:
Closes #2674

### DIFF
--- a/newsfragments/2675.fixed.md
+++ b/newsfragments/2675.fixed.md
@@ -1,0 +1,1 @@
+Fix regression of `impl FromPyObject for [T; N]` no longer accepting types passing `PySequence_Check`, e.g. NumPy arrays, since version 0.17.0. This the same fix that was applied `impl FromPyObject for Vec<T>` in version 0.17.1 extended to fixed-size arrays.
--- a/src/conversions/array.rs
+++ b/src/conversions/array.rs
@@ -3,9 +3,11 @@ use crate::{exceptions, PyErr};
 #[cfg(min_const_generics)]
 mod min_const_generics {
     use super::invalid_sequence_length;
-    use crate::conversion::IntoPyPointer;
+    use crate::conversion::{AsPyPointer, IntoPyPointer};
+    use crate::types::PySequence;
     use crate::{
-        ffi, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, PyTryFrom, Python, ToPyObject,
+        ffi, FromPyObject, IntoPy, Py, PyAny, PyDowncastError, PyObject, PyResult, PyTryFrom,
+        Python, ToPyObject,
     };
 
     impl<T, const N: usize> IntoPy<PyObject> for [T; N]
@@ -61,8 +63,16 @@ mod min_const_generics {
     where
         T: FromPyObject<'s>,
     {
-        let seq = <crate::types::PySequence as PyTryFrom>::try_from(obj)?;
-        let seq_len = seq.len()? as usize;
+        // Types that pass `PySequence_Check` usually implement enough of the sequence protocol
+        // to support this function and if not, we will only fail extraction safely.
+        let seq = unsafe {
+            if ffi::PySequence_Check(obj.as_ptr()) != 0 {
+                <PySequence as PyTryFrom>::try_from_unchecked(obj)
+            } else {
+                return Err(PyDowncastError::new(obj, "Sequence").into());
+            }
+        };
+        let seq_len = seq.len()?;
         if seq_len != N {
             return Err(invalid_sequence_length(N, seq_len));
         }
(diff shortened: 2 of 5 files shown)


src/conversions/array.rs
```diff
@@ -61,8 +63,16 @@ mod min_const_generics {
...
     {
-        let seq = <crate::types::PySequence as PyTryFrom>::try_from(obj)?;
-        let seq_len = seq.len()? as usize;
+        // Types that pass `PySequence_Check` usually implement enough of the sequence protocol
+        // to support this function and if not, we will only fail extraction safely.
+        let seq = unsafe {
+            if ffi::PySequence_Check(obj.as_ptr()) != 0 {
+                <PySequence as PyTryFrom>::try_from_unchecked(obj)
+            } else {
+                return Err(PyDowncastError::new(obj, "Sequence").into());
```
davidhewitt, line 72: Willing to add a test for this error branch (ie. pass a non-iterable) to get the coverage to 100%? 😃

newsfragments/2675.fixed.md
```diff
@@ -0,0 +1 @@
+`impl FromPyObject for [T; N]` will accept anything passing `PySequence_Check`, e.g. NumPy arrays, in the same way that `impl FromPyObject for Vec<T>` was modified after [#2477](https://github.com/PyO3/pyo3/pull/2477).
```
davidhewitt, line 1: Given this is a revert, I think it'd be quite handy to state what version the behaviour is reverting to rather than the offending PR. Users who want to see the PR can start with the link from the CHANGELOG to this PR and follow the discussion thread.

E.g. "Fix regression of `impl FromPyObject for [T; N]` no longer accepting types passing `PySequence_Check`, e.g. NumPy arrays, since 0.17.0."

davidhewitt (commented):
Thanks for handling this, and sorry for my slow reply. My son and I were both sick with colds over the weekend previous and I'm just about getting back up to speed now.

Just a couple of tiny suggestions 👍

src/conversions/array.rs
adamreichold: Added a test case triggering this error path.

newsfragments/2675.fixed.md
adamreichold: Adjusted the wording, please re-review.

davidhewitt (approved):
Looks great, thanks again! I guess we should put out 0.17.3 - anything else we should include?

adamreichold:
> Looks great, thanks again! I guess we should put out 0.17.3 - anything else we should include?

I don't think we have any other bug fixes ready that are blocked only on review. Everything else seems to be enhancements.

We do have two undiagnosed crash bugs in https://github.com/PyO3/pyo3/issues/2611 and https://github.com/PyO3/pyo3/discussions/26778 though.

Personally, I am not sure if this fix here alone warrants a point release but then again it is a regression. (Turning NumPy arrays into fixed-size Rust arrays seems a bit fringe as NumPy arrays are relatively inefficient for small arrays which is where fixed-size Rust arrays are usually applied.)

davidhewitt:
Ok. Let's hold off for the moment; Python 3.11.0 is due at the end of the month so maybe we can target 0.17.3 for then once we can confirm PyO3 officially supports it. (Unless we can land some bigger features worthy of calling it 0.18!)

PR #2678, opened 2022-10-12
github pages example for sys.dict modules

flickpp:
Add an example into `python_from_rust.md` for creating a module and inserting it into the `sys.modules` dictionary.

As discussed here:
GH-2649 example for inserting moulde in sys.dict

### DIFF
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -245,20 +245,42 @@ def leaky_relu(x, slope=0.01):
 # }
 ```
 
-### Want to embed Python in Rust with exposed modules?
+### Want to embed Python in Rust with additional modules?
 
-Internally Python maintains the `sys.modules` dict.
-An import in Python is simply a lookup into this dict.
+Python maintains the `sys.modules` dict as a cache of all imported modules.
+An import in Python will first attempt to lookup the module from this dict,
+and if not present will use various strategies to attempt to locate and load
+the module.
 
-It is necessary to insert the new module into the dict
-before it is exposed to Python. There are two ways of
-doing this.
+The [`append_to_inittab`]({{*PYO3_DOCS_URL}}/pyo3/macro.append_to_inittab.html)
+macro can be used to add additional `#[pymodule]` modules to an embedded
+Python interpreter. The macro **must** be invoked _before_ initializing Python.
 
-   * manually inserting into `sys.modules`
-   * using [`append_to_inittab`]({{*PYO3_DOCS_URL}}/pyo3/macro.append_to_inittab.html) macro
+As an example, the below adds the module `foo` to the embedded interpreter:
 
+```rust
+use pyo3::prelude::*;
 
-Example using `sys.modules`
+#[pyfunction]
+fn add_one(x: i64) -> i64 {
+    x + 1
+}
+
+#[pymodule]
+fn foo(_py: Python<'_>, foo_module: &PyModule) -> PyResult<()> {
+    foo_module.add_function(wrap_pyfunction!(add_one, foo_module)?)?;
+    Ok(())
+}
+
+fn main() -> PyResult<()> {
+    pyo3::append_to_inittab!(foo);
+    Python::with_gil(|py| Python::run(py, "import foo; foo.add_one(6)", None, None))
+}
+```
+
+If `append_to_inittab` cannot be used due to constraints in the program,
+an alternative is to create a module using [`PyModule::new`]
+and insert it manually into `sys.modules`:
 
 ```rust
 use pyo3::prelude::*;
@@ -288,32 +310,6 @@ fn main() -> PyResult<()> {
 }
 ```
 
-Example using [`append_to_inittab`](https://docs.rs/pyo3/latest/pyo3/macro.append_to_inittab.html):
-
-**WARNING** You must call `append_to_inittab` _before_ initializing Python.
-
-
-```rust
-use pyo3::prelude::*;
-
-#[pyfunction]
-fn add_one(x: i64) -> i64 {
-    x + 1
-}
-
-#[pymodule]
-fn foo(_py: Python, foo_module: &PyModule) -> PyResult<()> {
-    foo_module.add_function(wrap_pyfunction!(add_one, foo_module)?)?;
-    Ok(())
-}
-
-fn main() -> PyResult<()> {
-    pyo3::append_to_inittab!(foo);
-    Python::with_gil(|py| Python::run(py, "import foo; foo.add_one(6)", None, None))
-}
-```
-
-
 ### Include multiple Python files
 
 You can include a file at compile time by using
@@ -331,9 +327,9 @@ Example directory structure:
 ├── Cargo.lock
 ├── Cargo.toml
 ├── python_app
-│   ├── app.py
-│   └── utils
-│       └── foo.py
+│   ├── app.py
+│   └── utils
+│       └── foo.py
 └── src
     └── main.rs
 ```
@@ -464,3 +460,6 @@ class House(object):
     })
 }
 ```
+
+
+[`PyModule::new`]: {{#PYO3_DOCS_URL}}/pyo3/types/struct.PyModule.html#method.new
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -245,6 +245,75 @@ def leaky_relu(x, slope=0.01):
 # }
 ```
 
+### Want to embed Python in Rust with exposed modules?
+
+Internally Python maintains the `sys.modules` dict.
+An import in Python is simply a lookup into this dict.
+
+It is necessary to insert the new module into the dict
+before it is exposed to Python. There are two ways of
+doing this.
+
+   * manually inserting into `sys.modules`
+   * using [`append_to_inittab`]({{*PYO3_DOCS_URL}}/pyo3/macro.append_to_inittab.html) macro
+
+
+Example using `sys.modules`
+
+```rust
+use pyo3::prelude::*;
+use pyo3::types::PyDict;
+
+#[pyfunction]
+pub fn add_one(x: i64) -> i64 {
+    x + 1
+}
+
+fn main() -> PyResult<()> {
+    Python::with_gil(|py| {
+        // Create new module
+        let foo_module = PyModule::new(py, "foo")?;
+        foo_module.add_function(wrap_pyfunction!(add_one, foo_module)?)?;
+
+        // Import and get sys.modules
+        let sys = PyModule::import(py, "sys")?;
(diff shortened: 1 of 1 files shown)


guide/src/python_from_rust.md
```diff
@@ -236,6 +236,42 @@ def leaky_relu(x, slope=0.01):
 # }
 ```
 
+### Want to embed Python in Rust with exposed modules?
+
+Internally Python maintains the `sys.modules` dict.
+An import in Python is simply a lookup into this dict.
+
+In pyo3 it is nessercary to insert the new module into the dict
+before it is exposed to Python.
```
davidhewitt, line 245:
```suggestion
In pyo3 it is necessary to insert the new module into the dict
before it is exposed to Python.
```

guide/src/python_from_rust.md
```diff
@@ -236,6 +236,42 @@ def leaky_relu(x, slope=0.01):
...
+        // Create new module
+        let foo_module = PyModule::new(py, "foo")?;
+        foo_module.add_function(wrap_pyfunction!(add_one, foo_module)?)?;
+
+        // Import and get sys.modules
+        let sys = PyModule::import(py, "sys")?;
+        let py_modules: &PyDict = sys.getattr("modules")?.downcast()?;
+
+        // Insert foo into sys.modules
+        py_modules.set_item("foo", foo_module)?;
```
davidhewitt, line 267: There is actually also another approach, which is to use a normal `#[pymodule] fn my_module` and `append_to_inittab!(my_module)`. Unfortunately the [docs for `append_to_inittab`](https://docs.rs/pyo3/latest/pyo3/macro.append_to_inittab.html) are woeful and lack an example.

It should be enough to call `append_to_inittab!(my_module)` _before_ calling `Python::with_gil`. Would you be willing to also document this option too?

davidhewitt (commented):
Thanks, sorry for me taking a long time to join this discussion; I'd been unwell.

guide/src/python_from_rust.md
flickpp: Job done.

One question - most of the links in `python_from_rust.md` use the following prefix 
`https://docs.rs/pyo3/0.12`. Is there a reason for using 0.12 instead of latest?

`append_to_inittab` isn't avaliable for 0.12 so I just used /latest/

davidhewitt: Looks great 👍 

Ah, for the docs URL, could you actually use `{{#PYO3_DOCS_URL}}` like [here in `class.md`](https://github.com/PyO3/pyo3/blob/61fd70c4764958898744dff90fadf2cc64851015/guide/src/class.md?plain=1#L1014) please? If you'd be willing to update the others to use that too that would be perfect, they shouldn't have hard-coded 0.12, I must have missed them when I originally built the substitution.

flickpp: Job done.

I haven't changed any of the other links in the file. Nor have I linked to the `append_to_inittab` example from the api reference (by api reference, I mean cargo doc). This seems to me to be out of scope for this PR?

guide/src/python_from_rust.md
```diff
@@ -236,6 +236,75 @@ def leaky_relu(x, slope=0.01):
 # }
 ```
 
+### Want to embed Python in Rust with exposed modules?
+
+Internally Python maintains the `sys.modules` dict.
+An import in Python is simply a lookup into this dict.
```
birkenfeld, line 242: That is a bit misleading as stated - if the lookup fails, Python's import machinery takes over and tries to locate the module on the filesystem - going through `sys.path`, trying import hooks etc. (All of which doesn't find modules simply created by `PyModule_New`, which is the problem here.)

guide/src/python_from_rust.md
```diff
@@ -236,6 +236,75 @@ def leaky_relu(x, slope=0.01):
...
+
+It is necessary to insert the new module into the dict
+before it is exposed to Python. There are two ways of
+doing this.
+
+   * manually inserting into `sys.modules`
+   * using [`append_to_inittab`]({{*PYO3_DOCS_URL}}/pyo3/macro.append_to_inittab.html) macro
```
birkenfeld, line 249:
```suggestion
   * declaring the module using the `pymodule` macro and using the [`append_to_inittab`]({{*PYO3_DOCS_URL}}/pyo3/macro.append_to_inittab.html) macro
```

guide/src/python_from_rust.md
```diff
@@ -236,6 +236,75 @@ def leaky_relu(x, slope=0.01):
...
+### Want to embed Python in Rust with exposed modules?
+
+Internally Python maintains the `sys.modules` dict.
+An import in Python is simply a lookup into this dict.
+
+It is necessary to insert the new module into the dict
+before it is exposed to Python. There are two ways of
+doing this.
+
+   * manually inserting into `sys.modules`
```
birkenfeld, line 248:
```suggestion
   * creating the module using `PyModule::new` and manually inserting it into `sys.modules`
```

guide/src/python_from_rust.md
```diff
@@ -236,6 +236,75 @@ def leaky_relu(x, slope=0.01):
...
+
+        // Now we can import + run our python code
+        Python::run(py, "import foo; foo.add_one(6)", None, None)
+    })
+}
+```
+
+Example using [`append_to_inittab`](https://docs.rs/pyo3/latest/pyo3/macro.append_to_inittab.html)
+
+**WARNING** You must call `append_to_inittab` _before_ initializing Python.
```
birkenfeld, line 284:
```suggestion
Note: You must call `append_to_inittab` _before_ initializing Python.
```

guide/src/python_from_rust.md
```diff
@@ -236,6 +236,75 @@ def leaky_relu(x, slope=0.01):
...
+        // Insert foo into sys.modules
+        py_modules.set_item("foo", foo_module)?;
+
+        // Now we can import + run our python code
+        Python::run(py, "import foo; foo.add_one(6)", None, None)
+    })
+}
+```
+
+Example using [`append_to_inittab`](https://docs.rs/pyo3/latest/pyo3/macro.append_to_inittab.html)
```
birkenfeld, line 282:
```suggestion
Example using [`append_to_inittab`](https://docs.rs/pyo3/latest/pyo3/macro.append_to_inittab.html):
```

guide/src/python_from_rust.md
```diff
@@ -236,6 +236,75 @@ def leaky_relu(x, slope=0.01):
...
+
+It is necessary to insert the new module into the dict
+before it is exposed to Python. There are two ways of
+doing this.
+
+   * manually inserting into `sys.modules`
+   * using [`append_to_inittab`]({{*PYO3_DOCS_URL}}/pyo3/macro.append_to_inittab.html) macro
+
+
+Example using `sys.modules`
```
birkenfeld, line 252:
```suggestion
Example using `PyModule::new` and `sys.modules`:
```

birkenfeld (changes requested):

guide/src/python_from_rust.md
flickpp: This is true, but probably not relevant here?

I don't think it's misleading but I'm happy to hear what others think.

guide/src/python_from_rust.md
flickpp: I personally find this verbose using [`append_to_inittab`] communicates the point...

guide/src/python_from_rust.md
flickpp: Again this seems not to be needed (for me).

guide/src/python_from_rust.md
flickpp: It's a WARNING, not a NOTE because the process will panic. It's also following the WARNING convention of the other examples in the file.

guide/src/python_from_rust.md
flickpp: Again, too verbose for me.

guide/src/python_from_rust.md
birkenfeld: The point is that they are two very different things, which have different use cases. You can't use `append_to_inittab` with `PyModule::new` created modules and vice versa.

guide/src/python_from_rust.md
birkenfeld: See above. `PyModule::new` is an integral part of this method.

guide/src/python_from_rust.md
birkenfeld: See above.

guide/src/python_from_rust.md
birkenfeld: OK, let me rephrase my comment in a less polite manner: calling all imports "simply a lookup" completely misrepresents what an import is. This is a wrong statement and should not go in like this.

guide/src/python_from_rust.md
```diff
@@ -236,6 +236,75 @@ def leaky_relu(x, slope=0.01):
...
 ```
 
+### Want to embed Python in Rust with exposed modules?
+
+Internally Python maintains the `sys.modules` dict.
+An import in Python is simply a lookup into this dict.
+
+It is necessary to insert the new module into the dict
+before it is exposed to Python. There are two ways of
+doing this.
```
birkenfeld, line 246: One more problem here: the `append_to_inittab` way does *not* add the module into `sys.modules`. It just makes it findable in one of the locations that are tried when it is *not found* there.

birkenfeld (changes requested):

guide/src/python_from_rust.md
flickpp: I take your point. That said, mentioning scanning the file system and running hooks seems to me to not be the purpose of this particular doc.

How about something like:

The python interpreter starts with a number of built-in modules being available in `sys.dict`. It is necessary to insert a newly created PyModule into this dictionary.

guide/src/python_from_rust.md
flickpp: Is this an implementation detail that might change between versions?

guide/src/python_from_rust.md
flickpp: I'm happy to repeat `PyModule::new` and `append_to_inittab` again and again. I don't see it as a huge issue.

My preference would still be to put it _once_, and then use less verbose more flowing English after that.

guide/src/python_from_rust.md
birkenfeld: I don't think so. `append_to_inittab` makes the module known without importing it (i.e. executing the module init function). This is a separate step and should not be mixed with the import itself.

davidhewitt:
I've force-pushed to update this & address feedback, will now merge. Thanks again!

PR #2685, opened 2022-10-14
Use a TypeError, rather than a ValueError, when refusing to treat a str as a Vec

alex:
This is far more consistent with how these exceptions are usually used

### DIFF
--- a/newsfragments/2685.changed.md
+++ b/newsfragments/2685.changed.md
@@ -1,0 +1,1 @@
+Use a TypeError, rather than a ValueError, when refusing to treat a str as a Vec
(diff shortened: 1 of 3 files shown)


adamreichold:
@alex Before we proceed with an API change here, any opinions on #2632 which might remove this code path entirely?

alex:
I'm pretty ambivalent.

alex:
Given that #2632 hasn't reached a consensus, would it be appropriate to merge this in the event this code path lives on?

davidhewitt:
I think it would make sense to merge this, as it's IMO a reasonable correction over what already exists while we decide whether to merge the bigger change.

davidhewitt:
@adamreichold please forgive me if you disagree!

adamreichold:
> @adamreichold please forgive me if you disagree!

Please don't worry, a disagreement is not an attack, so I don't think there is anything to forgive here.

And while I personally would like #2632 to land, I very much understand that there is no consensus yet and that change should therefore not block reasonable improvements.

PR #2689, opened 2022-10-16
Support documentation when creating a python function from a closure.

LaurentMazare:
This adds a new function similar to `new_closure` but that lets the user specify the Python function documentation, this is not done by just adding an argument to `new_closure` so as to avoid breaking the current api but maybe would it be simpler to always require this argument.

A dedicated test has been added checking that the proper documentation can then be seen from the Python side.

### DIFF
--- a/newsfragments/2689.added.md
+++ b/newsfragments/2689.added.md
@@ -1,0 +1,1 @@
+Add a `new_closure_with_doc` function so that one can specify some documentation when wrapping a Rust closure.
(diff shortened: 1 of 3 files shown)


birkenfeld:
Thanks for the PR, this seems to be a in the same vein as #2686?

LaurentMazare:
Oh that's a good point, funny that this was created two days ago. Anyway, I'll close this PR as the discussion seems more advanced on the other one. Thanks for pointing this out.

PR #2690, opened 2022-10-16
Stop leaking in `new_closure`

mejrs:
Continues https://github.com/PyO3/pyo3/pull/2686#issuecomment-1279842232

Previously this leaked memory on every call:
```rust
use pyo3::prelude::*;
use pyo3::types::*;

fn main() {
    Python::with_gil(|_py| loop {
        let pool = unsafe { _py.new_pool() };
        let py = pool.python();
        let add_one = |args: &PyTuple, _kwargs: Option<&PyDict>| -> PyResult<_> {
            let i = args.extract::<(i64,)>()?.0;
            Ok(i + 1)
        };
        let closure = PyCFunction::new_closure(add_one, py).unwrap();
    });
}
```

Now it does no more:
- The `PyMethodDef` is cleaned up when the closure gets dropped
- Any Cstrings are deallocated if necessary

I am not sure what the best next step is. I'm leaning towards an api that takes `&'str` (and always allocates a CString) and another api taking `&'static CStr` that does not allocate cstrings.

### DIFF
--- a/src/internal_tricks.rs
+++ b/src/internal_tricks.rs
@@ -1,6 +1,6 @@
 use crate::ffi::{Py_ssize_t, PY_SSIZE_T_MAX};
 use std::ffi::{CStr, CString};
-
+use std::os::raw::c_char;
 pub struct PrivateMarker;
 
 macro_rules! private_decl {
@@ -35,15 +35,45 @@ macro_rules! pyo3_exception {
 #[derive(Debug)]
 pub(crate) struct NulByteInString(pub(crate) &'static str);
 
+#[derive(Copy, Clone)]
+pub(crate) enum MaybeLeaked {
+    Static(*const c_char),
+    Leaked(*mut c_char),
+}
+
+impl MaybeLeaked {
+    pub(crate) fn as_ptr(&self) -> *const c_char {
+        match *self {
+            Self::Static(ptr) => ptr,
+            Self::Leaked(ptr) => ptr as *const c_char,
+        }
+    }
+}
+
 pub(crate) fn extract_cstr_or_leak_cstring(
     src: &'static str,
     err_msg: &'static str,
-) -> Result<&'static CStr, NulByteInString> {
-    CStr::from_bytes_with_nul(src.as_bytes())
-        .or_else(|_| {
-            CString::new(src.as_bytes()).map(|c_string| &*Box::leak(c_string.into_boxed_c_str()))
-        })
-        .map_err(|_| NulByteInString(err_msg))
+) -> Result<MaybeLeaked, NulByteInString> {
+    let bytes = src.as_bytes();
+    let nul_pos = memchr::memchr(0, bytes);
+    match nul_pos {
+        Some(nul_pos) if nul_pos + 1 == bytes.len() => {
+            // SAFETY: We know there is only one nul byte, at the end
+            // of the byte slice.
+            let ptr = unsafe { CStr::from_bytes_with_nul_unchecked(bytes).as_ptr() };
+            Ok(MaybeLeaked::Static(ptr))
+        }
+        Some(_) => Err(NulByteInString(err_msg)),
+        None => {
+            let mut v = Vec::with_capacity(bytes.len() + 1);
+            v.extend_from_slice(bytes);
+
+            // SAFETY: There is no null byte, and `from_vec_unchecked` will append one.
+            let cstring = unsafe { CString::from_vec_unchecked(v) };
+            let ptr = cstring.into_raw();
+            Ok(MaybeLeaked::Leaked(ptr))
+        }
+    }
 }
 
 /// Convert an usize index into a Py_ssize_t index, clamping overflow to
--- a/src/types/function.rs
+++ b/src/types/function.rs
@@ -1,6 +1,7 @@
 use crate::derive_utils::PyFunctionArguments;
 use crate::exceptions::PyValueError;
 use crate::impl_::panic::PanicTrap;
+use crate::methods::PyMethodDefDestructor;
 use crate::panic::PanicException;
 use crate::{
     ffi,
@@ -8,6 +9,7 @@ use crate::{
     types, AsPyPointer,
 };
 use crate::{prelude::*, GILPool};
+use std::mem::ManuallyDrop;
 use std::os::raw::c_void;
 
 /// Represents a builtin Python function object.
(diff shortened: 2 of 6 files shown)


src/types/function.rs
```diff
@@ -162,7 +173,19 @@ impl PyCFunction {
         } else {
             (std::ptr::null_mut(), std::ptr::null_mut())
         };
-        Self::internal_new_from_pointers(method_def, py, mod_ptr, module_name)
+        let (def, _destructor) = method_def
+            .as_method_def()
+            .map_err(|err| PyValueError::new_err(err.0))?;
+
+        let def = Box::into_raw(Box::new(def));
```
mejrs, line 180: This still leaks; I haven't looked at what actually calls this function so I'm not sure if that's important.

src/types/function.rs
```diff
@@ -106,45 +117,45 @@ impl PyCFunction {
...
     where
         F: Fn(&types::PyTuple, Option<&types::PyDict>) -> R + Send + 'static,
         R: crate::callback::IntoPyCallbackOutput<*mut ffi::PyObject>,
     {
-        let function_ptr = Box::into_raw(Box::new(f));
-        let capsule = unsafe {
+        let method_def = pymethods::PyMethodDef::cfunction_with_keywords(
+            "pyo3-closure\0",
+            pymethods::PyCFunctionWithKeywords(run_closure::<F, R>),
+            "\0",
```
mejrs, line 128: This should probably be a `None`

src/internal_tricks.rs
```diff
@@ -35,15 +35,45 @@ macro_rules! pyo3_exception {
 #[derive(Debug)]
 pub(crate) struct NulByteInString(pub(crate) &'static str);
 
+#[derive(Copy, Clone)]
+pub(crate) enum MaybeLeaked {
```
adamreichold, line 39: Isn't this actually `MaybeOwned::Owned(*mut c_char)` because the point of the type is to not leak things?

Also could this also be something like

```rust
pub(crate) enum MaybeOwned {
  Static(&'static CStr),
  Owned(CString),
}
```

or does this violate any aliasing guarantees?

src/types/function.rs
```diff
@@ -45,11 +55,12 @@ where
     let trap = PanicTrap::new("uncaught panic during drop_closure");
     let pool = GILPool::new();
     if let Err(payload) = std::panic::catch_unwind(|| {
-        let boxed_fn: Box<F> = Box::from_raw(ffi::PyCapsule_GetPointer(
+        let destructor: Box<ClosureDestructor<F>> = Box::from_raw(ffi::PyCapsule_GetPointer(
```
adamreichold, line 58: Would this code benefit from using our generic capsule API?

src/internal_tricks.rs
mejrs: I went with my approach over this for some reasons:
- the type is smaller (doesn't really matter, I suppose)
- I don't think aliasing `CString`  is a problem in practice but using it would make `ClosureDestructor` (more of) a self referential struct, so I'd really prefer to use raw pointers 
- it's a lot easier to create a `*const c_char` at compile time than a `&'static CStr` (msrv 😭 )

adamreichold: Another option that comes to mind is to not store the `Static` variant at all (we do not need it for destruction) and use a signature similar to `as_method_def` above for `extract_cstr_or_leak_cstring`, e.g. make it run

```rust
Result<(*const c_char, ManuallyDrop<Option<CString>>), NulByteInString>
```

and use `Option<CString>` inside of `PyMethodDefDestructor` (which is a sort of "double-option" ATM).

src/types/function.rs
davidhewitt: It's called by the `PyCFunction::new` and `new_with_keywords` methods.

I think the only way around this would be to have a way to construct `&static ffi::PyMethodDef` safely and change the `PyCFunction` apis to take it (even if it's wrapped up in another type so users can't see it).

Either that, or should they be changed to use a capsule in the same way as `PyCFunction::new_closure`?

adamreichold: > Either that, or should they be changed to use a capsule in the same way as PyCFunction::new_closure?

I think this would be the smallest change to fix this. Everything else could be a follow-up optimization.

(Generally speaking, I think there would be an optimization opportunity by creating a `#[pyclass]` wrapping a `ffi::PyMethodDef` instead of using a capsule to avoid one layer of dynamic allocation as `ffi::PyMethodDef` could be stored inline in the `#[pyclass]` whereas the capsule is itself limited to a pointer if I understand things correctly.)

davidhewitt: While true, we've steered clear of having `#[pyclass]` types internal to PyO3 because the wouldn't be publicly accessible or re-used between PyO3 modules. Solutions to that seem hard. (Maybe it's not actually a problem though.)

adamreichold: > because the wouldn't be publicly accessible

I think that would not be a problem here as this should really be an implementation detail. I suspect having to manually write the `#[pyclass]` to avoid the macros as it used to be done `rust-numpy` would be a bit of a hassle though (but easier from within PyO3 than from within `rust-numpy`).

> re-used between PyO3 modules.

I guess this makes it a trade-off between less indirection but more type objects. There are usually multiple method definitions per extension module, so I suspect this could work out. At least I know of no complaints w.r.t `rust-numpy`'s `PySliceContainer` which suffers from the same issue.

(This does mean that this could be applied only for "one-way" types though, i.e. when we hand data to Python for taking care of. If we might happen to consume such a type created by a different PyO3-written extension module, this would break.)

davidhewitt: Actually can use the `#[pyclass(crate = "crate")]` to just use the macros I think. However it'd mean this feature wouldn't work if the macros feature was disabled.

Yes, perhaps you are right that in this "internal" case it would be suitable.

src/internal_tricks.rs
```diff
@@ -35,15 +35,45 @@ macro_rules! pyo3_exception {
...
     err_msg: &'static str,
-) -> Result<&'static CStr, NulByteInString> {
-    CStr::from_bytes_with_nul(src.as_bytes())
-        .or_else(|_| {
-            CString::new(src.as_bytes()).map(|c_string| &*Box::leak(c_string.into_boxed_c_str()))
-        })
-        .map_err(|_| NulByteInString(err_msg))
+) -> Result<MaybeLeaked, NulByteInString> {
+    let bytes = src.as_bytes();
+    let nul_pos = memchr::memchr(0, bytes);
```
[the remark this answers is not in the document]
mejrs: Otherwise we need to traverse `bytes` twice; both of those functions check for internal nul bytes.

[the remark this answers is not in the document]
adamreichold: I understand that. I would say that the price of a new dependency (`memchr` was only a dev-dependency until now) and more unsafe blocks is relatively steep for that. I also suspect that the compiler might just be able to optimize out the redundant checks in the safe alternatives.

(As an aside, the change is somewhat orthogonal to the topic of the PR. It might be good to propose it separately to avoid derailing the fix for the memory leak. But then, this is probably not too convincing coming from the one doing the derailing in the first place.)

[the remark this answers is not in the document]
davidhewitt: How about this alternative (pseudocode):

```rust
let bytes = src.as_bytes();
if !bytes.is_empty() && bytes[-1] == '\0' {
    CStr::from_bytes_with_nul(bytes)
} else {
    CString::new(bytes)
}
```

Checking for the trailing nul doesn't need an external dependency. We can then branch on the leak-or-not operation and we know that any errors must be due to internal nuls.

davidhewitt:
Sorry didn't mean to force-push here; opened #2842 instead as a simpler alternative inspired by this one.

PR #2691, opened 2022-10-16
Suggestions to update "getting started" docs so things work out of the box.

ctb:
Thank you so much for pyo3! This is an awesome project and I'm just getting started with it!

This PR adds some (IMO) missing specifics to `getting_started.md`:
* mention location of rust code explicitly - this was a minor confusion for me so I thought I'd suggest the change - it was the first place I looked, but I did have to think about it :)
* define a default package and edition so that `maturin develop` will work out of the box.

(Here is my version of the result of running through getting-started: https://github.com/ctb/pyo3-example)


guide/src/getting_started.md
```diff
@@ -129,7 +136,7 @@ classifiers = [
 
 ## Running code
 
-After this you can setup rust code to be available in python as such:
+After this you can setup rust code to be available in python as below; for example, you can place this code in `src/lib.rs`
```
birkenfeld, line 139:
```suggestion
After this you can setup rust code to be available in python as below; for example, you can place this code in `src/lib.rs`:
```

guide/src/getting_started.md
```diff
@@ -95,6 +95,13 @@ Make sure that the rust you want to be able to access from Python is compiled in
 
 
 ```toml
+[package]
+# Name of the package. If you already have a package defined in `Cargo.toml`, you can remove
```
birkenfeld, line 99: I was a bit confused by "name of the package", which refers only to the first entry, while the rest of the comment refers to the whole section.

ctb: thanks! addressed in ca1c9c1d - changed to

```
# If you already have [package] information in `Cargo.toml`, you can ignore
# this section!
[package]
# `name` here is name of the package.
name = "pyo3_start"
# these are good defaults:
version = "0.1.0"
edition = "2021"
```

birkenfeld (approved):
LGTM now, thanks!

ctb:
(I'm happy to add a Changelog entry if you like, but it seems too small to bother doing that - lmk)

davidhewitt:
No changelog needed (generally don't for docs changes). Thanks!

Issue #1003, opened 2020-06-26
`pyo3(get, set)` attribute not found when using `cfg_attr`


Waelwindows:
## 🐛 Bug Reports

When reporting a bug, please provide the following information. If this is not a bug report you can just discard this template.

### 🌍 Environment

 - Your operating system and version: `Arch Linux x86_64 5.7.4-arch1-1`
 - Your python version: `Python 3.8.3`
 - How did you install python (e.g. apt or pyenv)? Did you use a virtualenv?: `pacman`
 - Your Rust version (`rustc --version`): `rustc 1.46.0-nightly (6bb3dbfc6 2020-06-22)`
 - Your PyO3 version: `0.10.1`
 - Have you tried using latest PyO3 master (replace `version = "0.x.y"` with `git = "https://github.com/PyO3/pyo3")?`: No

### 💥 Reproducing

Please provide a [minimal working example](https://stackoverflow.com/help/mcve). This means both the Rust code and the Python.

This compiles
```rust
#[cfg_attr(feature="pyo3", pyclass)]
#[derive(Debug, Default)]
pub struct Foo {
    #[pyo3(get, set)]
    bar: bool,
}
```
However, this doesn't 
```rust
#[cfg_attr(feature="pyo3", pyclass)]
#[derive(Debug, Default)]
pub struct Foo {
    #[cfg_attr(feature="pyo3", pyo3(get, set))]
    bar: bool,
}
```
Because it can't find the `pyo3` attribute
```
error: cannot find attribute `pyo3` in this scope
  --> src/material/texture/mod.rs:45:32
   |
45 |     #[cfg_attr(feature="pyo3", pyo3(get, set))]
   |                                ^^^^

```
Please also write what exact flags are required to reproduce your results.

birkenfeld:
This is basically a duplicate of #780.

davidhewitt:
Based on https://github.com/PyO3/pyo3/issues/780#issuecomment-636124309 I wonder if we can resolve this by changing `#[pyclass]` expansion to (among other things) add a `#[derive(PyClassImpl)]` or something like that. Where the `PyClassImpl` derive macro would apparently be called with all `cfg_attr` attributes resolved.

This is kind of a hack, though, so I wonder if really we should be asking the compiler team if it's reasonable to expand all `#[cfg_attr]` clauses before calling custom attributes (as it already appears to do that for derive).

davidhewitt:
[...]

Issue #780, opened 2020-02-29
`#[new]` doesn't play nice with `cfg_attr`


c410-f3r:
Compiles:

```rust
use pyo3::prelude::*;

#[pyclass]
#[derive(Debug)]
pub struct Foo {
  field: i32
}

#[pymethods]
impl Foo {
  #[new]
  pub fn new(field: i32) -> Self {
    Self { field }
  }
}

```

Doesn't compile with `Invalid type as custom self` and `cannot find attribute 'new' in this scope` errors:

```rust
#[cfg(feature = "with_pyo3")]
use pyo3::prelude::*;

#[cfg_attr(feature = "with_pyo3", pyclass)]
#[derive(Debug)]
pub struct Foo {
  field: i32
}

#[cfg_attr(feature = "with_pyo3", pymethods)]
impl Foo {
  // Change `#[cfg_attr(feature = "with_pyo3", new)]` for `#[new]` and `cargo build --feature with_py03` to see a successful compilation.
  #[cfg_attr(feature = "with_pyo3", new)]
  pub fn new(field: i32) -> Self {
    Self { field }
  }
}
```

c410-f3r:
Ah, NVM. The error was trigger due to the lack of a constructor

davidhewitt:
According to gitter this is also an issue for `#[pyo3(get)]` etc.

Probably needs a thorough review of `cfg_attr` compatibility.

c410-f3r:
Version `0.10` worsens conditional compilation. Now `#[cfg_attr(feature = "with_pyo3", pymethods)]` and `#[cfg_attr(feature = "with_pyo3", pyclass)]` doesn't compile either.

davidhewitt:
Thanks for the update. I'll try to give this some investigation soon.

birkenfeld:
I can't reproduce the latest report; the example given above compiles fine with 0.10 if I make the `#[new]` unconditional, but the other attributes remain conditional.

It does not compile without `#[new]`, but that has nothing to do with conditional attributes.

birkenfeld:
I'm not sure how to resolve this. We can of course look into `cfg_attr` and ignore it while processing attributes, but that is only correct when the condition is as reported here, switching PyO3 off.  Something like switching between two different `#[new]` implementations depending on flags would still fall down.

Is there a way at all to resolve `cfg_attr` at macro evaluation time?

davidhewitt:
[...]

PR #2692, opened 2022-10-17
Implement get/set all on pyclass

mejrs:
Fixes https://github.com/PyO3/pyo3/issues/1003 and most of https://github.com/PyO3/pyo3/issues/780

I'm happy to bikeshed the naming; I'm not sure whether I prefer `get_all` or just `get`.

### DIFF
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -1,6 +1,7 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 
 use std::borrow::Cow;
+use std::mem;
 
 use crate::attributes::{
     self, kw, take_pyo3_options, CrateAttribute, ExtendsAttribute, FreelistAttribute,
@@ -60,12 +61,14 @@ pub struct PyClassPyO3Options {
     pub krate: Option<CrateAttribute>,
     pub dict: Option<kw::dict>,
     pub extends: Option<ExtendsAttribute>,
+    pub get_all: Option<kw::get_all>,
     pub freelist: Option<FreelistAttribute>,
     pub frozen: Option<kw::frozen>,
     pub mapping: Option<kw::mapping>,
     pub module: Option<ModuleAttribute>,
     pub name: Option<NameAttribute>,
     pub sequence: Option<kw::sequence>,
+    pub set_all: Option<kw::set_all>,
     pub subclass: Option<kw::subclass>,
     pub text_signature: Option<TextSignatureAttribute>,
     pub unsendable: Option<kw::unsendable>,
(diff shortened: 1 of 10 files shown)


birkenfeld:
LGTM, I think `get_all` is better than `get`.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -230,11 +241,33 @@ pub fn build_py_class(
...
+            }
+        }
+    }
+
+    if let Some(attr) = args.options.set_all {
+        for (_, FieldPyO3Options { set, .. }) in &mut field_options {
+            if mem::replace(set, true) {
+                return Err(syn::Error::new(attr.span(), DUPE_SET));
+            }
+        }
```
davidhewitt, line 268: IMO the spans for the errors here aren't ideal because they use the span of `get_all` and `set_all`, but the error is describing the `get` and `set` as the source of the error.

Using the `syn::Field` span (bound as `_` in these snippets) may be better as it'll point to the field which has the invalid attribute?

The ideal span would probably be the `get` attribute itself - you could convert the `FieldPyO3Options` to have `get: Option<kw::get>`, and then use those spans. It would however mean that your elegant solution of using `mem::replace` here wouldn't be quite enough.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -1085,3 +1118,10 @@ fn define_inventory_class(inventory_class_name: &syn::Ident) -> TokenStream {
         _pyo3::inventory::collect!(#inventory_class_name);
     }
 }
+
+const DUPE_SET: &str = "duplicate `set` - the struct is already annotated with `set_all`";
+const DUPE_GET: &str = "duplicate `get` - the struct is already annotated with `get_all`";
```
davidhewitt, line 1123: Nit: I think I'd use "useless" or "redundant" rather than duplicate?

```suggestion
const DUPE_SET: &str = "useless `set` - the struct is already annotated with `set_all`";
const DUPE_GET: &str = "useless `get` - the struct is already annotated with `get_all`";
```

davidhewitt (commented):
Looks great to me! I also agree with `get_all`/`set_all` being better than `get`/`set`.

Will need `newsfragments/2692.added.md` please :)

pyo3-macros-backend/src/pyclass.rs
mejrs: That's a good idea, done.

davidhewitt (approved):
Looks great, thanks very much! 👍

joshparallel:
Thank you for implementing this! Any idea when it'll make it to crates.io?

mejrs:
See https://github.com/PyO3/pyo3/issues/2785 I think there's a few things we want to merge before the next release.


Also see https://github.com/PyO3/pyo3/pull/2786 - if this PR is useful for you then that one might also be.

PR #2695, opened 2022-10-18
Make `is_instance()` and `is_subclass()` take `&PyAny`

### LINKED ISSUE #2694: Allow any object as "type" in `is_instance`
### Discussed in https://github.com/PyO3/pyo3/discussions/2693

<div type='discussions-op-text'>

<sup>Originally posted by **samuelcolvin** October 18, 2022</sup>
I'd like to do an `is_instance` check with `typing.Sequence`, however

```rs
let sequence_type: &PyType = sequence_obj.extract();
```

Is failing with `TypeError: '_SpecialGenericAlias' object cannot be converted to 'PyType'`.

Is this a fundamental limitation of the python C API, or something that pyo3 could support?

Mostly for my record, here's the failing test from [pydantic-core](https://github.com/pydantic/pydantic-core):

```py
def test_is_instance_sequence():
    v = SchemaValidator({'type': 'is-instance', 'cls': typing.Sequence})
    assert v.isinstance_python(1) is False
    assert v.isinstance_python([1]) is True
```

Fails when creating the `SchemaValidator` with:

```
    def test_is_instance_sequence():
>       v = SchemaValidator({'type': 'is-instance', 'cls': typing.Sequence})
E       pydantic_core._pydantic_core.SchemaError: Error building "is-instance" validator:
E         TypeError: '_SpecialGenericAlias' object cannot be converted to 'PyType'

tests/validators/test_is_instance.py:204: SchemaError
```

birkenfeld:
Should be (mostly?) compatible since `&PyType` auto-derefs to `&PyAny`.

Fixes #2694

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -448,8 +448,8 @@ impl PyErr {
 
     /// Returns true if the current exception is instance of `T`.
     #[inline]
-    pub fn is_instance(&self, py: Python<'_>, typ: &PyType) -> bool {
-        unsafe { ffi::PyErr_GivenExceptionMatches(self.type_ptr(py), typ.as_ptr()) != 0 }
+    pub fn is_instance(&self, py: Python<'_>, ty: &PyAny) -> bool {
+        unsafe { ffi::PyErr_GivenExceptionMatches(self.type_ptr(py), ty.as_ptr()) != 0 }
     }
 
     /// Returns true if the current exception is instance of `T`.
--- a/guide/src/ecosystem/async-await.md
+++ b/guide/src/ecosystem/async-await.md
@@ -540,8 +540,6 @@ fn main() -> PyResult<()> {
                     uvloop
                         .as_ref(py)
                         .getattr("Loop")?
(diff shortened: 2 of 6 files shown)


adamreichold (approved):
It is a bit sad to make the types more loose here but I guess it can't be helped and I guess there is not too much room for confusion due to the method names.

src/err/mod.rs
```diff
@@ -448,7 +448,7 @@ impl PyErr {
 
     /// Returns true if the current exception is instance of `T`.
     #[inline]
-    pub fn is_instance(&self, py: Python<'_>, typ: &PyType) -> bool {
+    pub fn is_instance(&self, py: Python<'_>, typ: &PyAny) -> bool {
```
mejrs, line 451:
```suggestion
    pub fn is_instance(&self, py: Python<'_>, ty: &PyAny) -> bool {
```
For consistency with `PyAny::is_instance`

mejrs (approved):

Issue #2193, opened 2022-02-27
Improving function "signatures"


davidhewitt:
TLDR; I propose a `#[pyo3(signature)]` attribute for both `#[pyfunction]` and `#[pymethods]` which:
- automatically generates `text_signature`
- leaves no room for ambiguity
- is required to specify default arguments (including for `Option<T>` which implicitly has a `None` default in some cases at current).

Something like:

```rust
#[pyfunction]
#[pyo3(signature = (a, /, b, c = None, d = 5, **kwargs))]
fn foo(a: i32, b: i32, c: Option<i32>, d: i32, kwargs: Option<&PyDict>) {
    (a, b, c, d, kwargs)
}
```

----

I've been thinking for a while that there are a few ergonomics improvements we could make around `#[pyfunction]` signatures. By signatures I collectively mean:
 - The `#[pyo3(text_signature)]` feature
 - Default arguments, which we support implicitly through `Option<T>` and explicitly.
   - I'm unsure whether implicitly through `Option<T>` is really the right behaviour, 
   - Explicitly is inconsistent and poorly documented, this can be done directly as [arguments to the `#[pyfunction]` macro](https://pyo3.rs/v0.15.1/function.html#argument-parsing) or via the [`#[pyo3(args)]` attribute](https://pyo3.rs/v0.15.1/class.html#method-arguments) for `#[pymethods]`.
 - Other features such as positional-only arguments (`/`), or `*args` and `**kwargs`. These are again supported by the same syntaxes as the "explicit default arguments" above.

There's a few problems with all of this stuff:
- They can get out of sync: the `#[text_signature]` repeats a lot of information already present in the rest of the function definition.
- Implicit `Option<T>` having a `None` default makes the possibility of going out of sync worse.
- After #2093 we _don't_ make an `Option<T>` implicitly defaulted to `None` if a required parameter follows it. This was to avoid a large breaking change on existing code, however it does also make the rules very confusing.
- There are bugs with the `#[args]` annotation such as #794 .
- I think it's just way too much boilerplate.

Here's an example which exhibits some of these problems:

```rust
#[pyfunction(a, "/", d = 5, kwargs = "**")]
#[pyo3(text_signature = "(a1, b = None, /, d, **kwargs)")]
fn foo(a: i32, b: i32, c: Option<i32>, d: i32, kwargs: Option<&PyDict>) {
    (a, b, c, d, kwargs)
}
```

In particular:
[...]

PR #2702, opened 2022-10-21
add `#[pyo3(signature = (...))]` attribute

davidhewitt:
This is an implementation of the proposal in #2193 to add a new `#[pyo3(signature = (...))]` attribute to specify function arguments.

This PR also deprecates the old way of setting function argument signatures.

Compared to `#[pyfunction(...)]` / `#[args(...)]`:
  - It is more consistent: always `signature = (...)` compared to different forms for `#[pyfunction]` and `#[pymethods]`.
  - It is more strictly validated: _all_ arguments must appear in the signature in the correct order.
  - It has more Pythonic syntax, e.g.:
    - `*args` instead of `args = "*"`.
    - Default arguments don't need additional string quoting - `signature(name = "David")` instead of `args(name = "\"David\"")`

The possible argument I can see which could be made against this is the increased verbosity from requiring all arguments to appear. I argue this is a plus to ensure correctness. It would be easy to relax this in future if feedback frequently requested it.

Future extensions:
  - I'd like this to also generate a `text_signature` automatically
  - I'd like this to eventually support Python type annotation syntax e.g. `signature = (a: int = 5, b: str = "hello")`, although the mechanism how the type annotations would be exposed is still unclear.

To review this, I suggest looking at the guide and tests, and if you're satisfied with that give the implementation a scan.

### DIFF
--- a/guide/src/function/signature.md
+++ b/guide/src/function/signature.md
@@ -94,6 +94,18 @@ num=44
 num=-1
 ```
 
+> Note: for keywords like `struct`, to use it as a function argument, use "raw ident" syntax `r#struct` in both the signature and the function definition:
+>
+> ```rust
+> # !#[allow(unused_code)]
+> # use pyo3::prelude::*;
+> #[pyfunction(signature = (r#struct = "foo"))]
+> fn method_with_keyword<'a>(&self, r#struct: &'a str) {
+> #     let _ = r#struct;
+>     /* ... */
+> }
+> ```
+
 ## Deprecated form
 
 The `#[pyfunction]` macro can take the argument specification directly, but this method is deprecated in PyO3 0.18 because the `#[pyo3(signature)]` option offers a simpler syntax and better validation.
--- a/guide/src/function/signature.md
+++ b/guide/src/function/signature.md
@@ -2,9 +2,109 @@
 
 The `#[pyfunction]` attribute also accepts parameters to control how the generated Python function accepts arguments. Just like in Python, arguments can be positional-only, keyword-only, or accept either. `*args` lists and `**kwargs` dicts can also be accepted. These parameters also work for `#[pymethods]` which will be introduced in the [Python Classes](../class.md) section of the guide.
 
-Like Python, by default PyO3 accepts all arguments as either positional or keyword arguments. The extra arguments to `#[pyfunction]` modify this behaviour. For example, below is a function that accepts arbitrary keyword arguments (`**kwargs` in Python syntax) and returns the number that was passed:
+Like Python, by default PyO3 accepts all arguments as either positional or keyword arguments. There are two ways to modify this behaviour:
+  - The `#[pyo3(signature = (...))]` option which allows writing a signature in Python syntax.
+  - Extra arguments directly to `#[pyfunction]`. (See deprecated form)
+
+## Using `#[pyo3(signature = (...))]`
+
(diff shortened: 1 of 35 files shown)


pyo3-macros-backend/src/method.rs
```diff
@@ -302,19 +308,26 @@ impl<'a> FnSpec<'a> {
                 .collect::<Result<_>>()?
         };
 
+        let signature = if let Some(signature) = signature {
```
adamreichold, line 311: Is it currently possible to specify both `#[pyo3(signature)]` and `#[args()]` with signature overriding everything in args? If so, I think that should become a hard error.

adamreichold: Found the answer further down below. Sorry for the noise.

adamreichold:
The issue number of the news fragments (2302) does not match the PR issue number (2702) which is why that CI job fails.

pyo3-macros-backend/src/method.rs
davidhewitt: Actually good to call this out - I had made it a hard error for `#[pyfunction]` but forgotten to do the same for `#[pymethods]`. I've pushed a fix!

davidhewitt:
Thanks for the review - addressed both comments.

guide/src/function/signature.md
```diff
@@ -2,12 +2,112 @@
...
+# }
+#[pymethods]
+impl MyClass {
+    #[new]
+    #[pyo3(signature = (num=-1))]
+    fn new(num: i32) -> Self {
+        MyClass { num }
+    }
+
+    #[pyo3(signature = (num=10, *py_args, name="Hello", **py_kwargs))]
```
birkenfeld, line 58: Document how to use names that are keywords in Rust? `crate` here and `r#crate` in the Rust argument list?

pytests/src/datetime.rs
```diff
@@ -37,6 +37,7 @@ fn make_time<'p>(
 }
 
 #[pyfunction]
+#[pyo3(signature = (hour, minute, second, microsecond, tzinfo, fold))]
```
birkenfeld, line 40: Why is this annotation now necessary?

birkenfeld (approved):
Perfect, the new syntax is so much nicer!

pytests/src/datetime.rs
davidhewitt: That is because `tzinfo` is an `Option` and `fold` is not an option; the error message without the signature is this:

```
warning: use of deprecated constant `pyo3::impl_::deprecations::REQUIRED_ARGUMENT_AFTER_OPTION`: required arguments after an `Option<_>` argument are ambiguous and being phased out
         = help: add a `#[pyo3(signature)]` annotation on this function to unambiguously specify the default values for all optional parameters
  --> pytests/src/datetime.rs:48:5
   |
48 |     fold: bool,
   |     ^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

By adding the signature which doesn't give `tzinfo` a default, it makes it unambiguous that `tzinfo` is a required argument in this context.

To be honest, I should probably split this change out into a separate PR.

davidhewitt: Pushed #2703 to review that change later.

guide/src/function/signature.md
davidhewitt: Good idea, I also added a test which caught a slight bug!

PR #2703, opened 2022-10-22
deprecate required argument after `Option<T>` without signature

davidhewitt:
This PR is a follow-up to #2702 to make required arguments after `Option<T>` arguments require a `#[pyo3(signature)]` annotation to remove the possible ambiguity on whether the developer actually wanted to have a required option (which is what PyO3 is currently forced to assume).


davidhewitt:
This is now rebased; I would be glad to hear folks' opinions on whether this makes sense to include in 0.18.

Overall I like what this does for removing ambiguity, which fits Rust's conventions well, however it may generate a lot of churn for users (see all the adjustments I made to our tests).

davidhewitt:
If there's no strong opinions either way, I'll merge this later in the week ahead of preparing the 0.18 branch.

Issue #2704, opened 2022-10-22
LC_ID_DYLIB of api3 module is wrong on macOS


Aksem:
### Bug Description

[watchfiles](https://github.com/samuelcolvin/watchfiles) package includes _rust_notify module compiled with PyO3 in api3 mode. It has LC_ID_DYLIB equal to `@rpath/_rust_notify.cpython-310-darwin.so` however its name is `_rust_notify.abi3.so` and the library cannot be linked properly.

I faced this problem when I tried to compile an app with [nuitka](https://github.com/Nuitka/Nuitka) in which watchfiles package is used.

Initial issue in Nuitka: https://github.com/Nuitka/Nuitka/issues/1844

### Steps to Reproduce

1. Compile a module in api3 mode on macOS.
2. Check it LC_ID_DYLIB with `otool -l <lib.so>` command.

### Your operating system and version

macOS 11

### Your Python version (`python --version`)

3.10.7

### Your Rust version (`rustc --version`)

1.56.0

### Your PyO3 version

0.16.4

### How did you install python? Did you use a virtualenv?

actions/setup-python@v4 in github actions ([config](https://github.com/samuelcolvin/watchfiles/blob/main/.github/workflows/ci.yml#L39))
No virtualenv, just dependency installation with pip

PR #2705, opened 2022-10-23
introduce trampolines for methods

davidhewitt:
This is a refactoring to the generated `#[pyfunction]` and `#[pymethods]` with the intention of reducing compile time and binary size.

The motivation for the change is that `cargo llvm-lines` currently reports many monomorphizations of `std::panicking::try::do_call` and friends being instantiated - in fact, it's one per `#[pyfunction]` and every method in `#[pymethods]`, which is quite a lot of instantiations.

The change follows a similar idea to #2478 changing `extract_argument` to make use of "dynamic dispatch". Here, we create "trampoline" wrappers for all possible C-API function pointers we expose to Python. These wrappers take a function pointer to the real implementation of the function in question, and call it after starting the `std::panic::catch_unwind` machiner and creating a `GILPool`.

Much like in #2704 I suspect LLVM can do a pretty good job at optimizing out the dynamic dispatch. I saw no change at all in benchmarks for the `pytests` crate. So overall I believe this should have a compile-time win with little-to-no impact on runtime performance.

For our `pytests` crate I get a 10% reduction in `cargo llvm-lines` count, and for real-world use cases potentially with many small function wrappers (e.g. getters / setters) I suspect the reduction could be even greater. 

Detail-wise, for some method `MyClass::foo`, the generated code change from a single C-ABI symbol:

```rust
unsafe extern "C" fn __pymethod_foo__(
    slf: *mut ffi::PyObject,
    args: *const *mut ffi::PyObject,
    nargs: ffi::Py_ssize_t,
    kwnames: *mut ffi::PyObject,
) -> *mut ffi::PyObject {
    let gil = GILPool::new();
    let py = gil.python();
    pyo3::callback::panic_result_into_callback_output(
        py,
        panic::catch_unwind(move || -> PyResult<_> {
            /* implementation */
        }),
    )
}
```

to an C-ABI symbol which just calls the trampoline wrapper with a pointer to the actual implementation.

```rust
unsafe extern "C" fn trampoline(
    slf: *mut ffi::PyObject,
    args: *const *mut ffi::PyObject,
    nargs: ffi::Py_ssize_t,
    kwnames: *mut ffi::PyObject,
) -> *mut ffi::PyObject
{
    impl_::trampoline::fastcall_with_keywords(
        slf,
        args,
        nargs,
        kwnames,
        MyClass::__pymethod_foo__,
    )
}
unsafe fn __pymethod_foo__<'py>(
    py: Python<'py>,
    slf: *mut ffi::PyObject,
    args: *const *mut ffi::PyObject,
    nargs: ffi::Py_ssize_t,
    kwnames: *mut ffi::PyObject,
) -> PyResult<*mut ffi::PyObject> {
    /* implementation */
}
```

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -481,18 +481,13 @@ impl<'a> FnSpec<'a> {
             CallingConvention::Noargs => {
                 let call = rust_call(vec![]);
                 quote! {
-                    unsafe extern "C" fn #ident (
+                    unsafe fn #ident<'py>(
+                        #py: _pyo3::Python<'py>,
                         _slf: *mut _pyo3::ffi::PyObject,
-                        _args: *mut _pyo3::ffi::PyObject,
-                    ) -> *mut _pyo3::ffi::PyObject
-                    {
+                    ) -> _pyo3::PyResult<*mut _pyo3::ffi::PyObject> {
                         #deprecations
-                        let gil = _pyo3::GILPool::new();
-                        let #py = gil.python();
-                        _pyo3::callback::panic_result_into_callback_output(#py, ::std::panic::catch_unwind(move || -> _pyo3::PyResult<_> {
-                            #self_conversion
-                            #call
-                        }))
+                        #self_conversion
+                        #call
                     }
                 }
             }
@@ -500,20 +495,17 @@ impl<'a> FnSpec<'a> {
                 let (arg_convert, args) = impl_arg_params(self, cls, &py, true)?;
                 let call = rust_call(args);
                 quote! {
-                    unsafe extern "C" fn #ident (
+                    unsafe fn #ident<'py>(
+                        #py: _pyo3::Python<'py>,
                         _slf: *mut _pyo3::ffi::PyObject,
                         _args: *const *mut _pyo3::ffi::PyObject,
                         _nargs: _pyo3::ffi::Py_ssize_t,
-                        _kwnames: *mut _pyo3::ffi::PyObject) -> *mut _pyo3::ffi::PyObject
-                    {
+                        _kwnames: *mut _pyo3::ffi::PyObject
+                    ) -> _pyo3::PyResult<*mut _pyo3::ffi::PyObject> {
                         #deprecations
-                        let gil = _pyo3::GILPool::new();
-                        let #py = gil.python();
-                        _pyo3::callback::panic_result_into_callback_output(#py, ::std::panic::catch_unwind(move || -> _pyo3::PyResult<_> {
-                            #self_conversion
-                            #arg_convert
-                            #call
-                        }))
+                        #self_conversion
+                        #arg_convert
+                        #call
                     }
                 }
             }
@@ -521,19 +513,16 @@ impl<'a> FnSpec<'a> {
                 let (arg_convert, args) = impl_arg_params(self, cls, &py, false)?;
                 let call = rust_call(args);
                 quote! {
-                    unsafe extern "C" fn #ident (
+                    unsafe fn #ident<'py>(
+                        #py: _pyo3::Python<'py>,
                         _slf: *mut _pyo3::ffi::PyObject,
                         _args: *mut _pyo3::ffi::PyObject,
-                        _kwargs: *mut _pyo3::ffi::PyObject) -> *mut _pyo3::ffi::PyObject
-                    {
+                        _kwargs: *mut _pyo3::ffi::PyObject
+                    ) -> _pyo3::PyResult<*mut _pyo3::ffi::PyObject> {
                         #deprecations
-                        let gil = _pyo3::GILPool::new();
-                        let #py = gil.python();
-                        _pyo3::callback::panic_result_into_callback_output(#py, ::std::panic::catch_unwind(move || -> _pyo3::PyResult<_> {
-                            #self_conversion
-                            #arg_convert
-                            #call
-                        }))
+                        #self_conversion
+                        #arg_convert
+                        #call
                     }
                 }
             }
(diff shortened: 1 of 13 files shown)


src/impl_/pyclass.rs
```diff
@@ -943,7 +943,15 @@ impl<T: PyClass> PyClassBaseType for T {
 
 /// Implementation of tp_dealloc for all pyclasses
 pub(crate) unsafe extern "C" fn tp_dealloc<T: PyClass>(obj: *mut ffi::PyObject) {
-    crate::callback_body!(py, T::Layout::tp_dealloc(obj, py))
+    /// Trampoline so that std::panic::catch_unwind only gets monomorphized once for
+    /// all tp_dealloc implementations
+    unsafe fn trampoline_dealloc(
```
adamreichold, line 948: Why define that here instead of in `mod trampoline`?

adamreichold:
It is not really related, but I had a look while reviewing: Doesn't `panic_result_into_callback_output` need a `PanicTrap` as well since `PanicException::from_panic_payload` drops the payload which could make a panic in its Drop impl reach the FFI boundary?

adamreichold (approved):
Great code bloat reduction! A nit and an unrelated question, but this looks good to me.

src/impl_/pyclass.rs
davidhewitt: Because `T::Layout::tp_dealloc` takes `py` last it doesn't match the rest of the trampolines (which take `py` first). What I'll do is push a `dealloc` macro-generated trampoline, and leave a wrapper here along with a TODO here to clean up by rearranging arguments to `T::Layout::tp_dealloc`. (Which I think is follow-up-PR worthy as a breaking change.)

davidhewitt:
> It is not really related, but I had a look while reviewing: Doesn't `panic_result_into_callback_output` need a `PanicTrap` as well since `PanicException::from_panic_payload` drops the payload which could make a panic in its Drop impl reach the FFI boundary?

Ahh, yes, when working on PanicTrap was one of the moments when I was thinking a refactor like this could be in order... Then I forgot to install a PanicTrap here. Let me push 😆

pyo3-macros-backend/src/method.rs
```diff
@@ -571,21 +558,66 @@ impl<'a> FnSpec<'a> {
             CallingConvention::Noargs => quote! {
                 _pyo3::impl_::pymethods::PyMethodDef::noargs(
                     #python_name,
-                    _pyo3::impl_::pymethods::PyCFunction(#wrapper),
+                    _pyo3::impl_::pymethods::PyCFunction({
+                        unsafe extern "C" fn trampoline(
```
birkenfeld, line 562: Would be cool not to have to repeat the (almost) identical signature here and in the `trampoline` fn. But the required macro trickery isn't worth it, likely?

davidhewitt: I agree with this, however I'd like to defer it to a follow-up PR sometime (there's a fair bit of refactoring which could improve the macro code, and I'd like it not to distract readers of this PR from the behavioural change).

src/impl_/trampoline.rs
```diff
@@ -0,0 +1,173 @@
...
+    ) -> *mut ffi::PyObject;
+);
+
+#[inline]
+pub fn trampoline_inner<F, R>(body: F) -> R
+where
+    F: for<'py> FnOnce(Python<'py>) -> PyResult<R> + UnwindSafe,
+    R: PyCallbackOutput,
+{
+    let trap = PanicTrap::new("uncaught panic at ffi boundary");
```
davidhewitt, line 164: @adamreichold this is the solution I had in mind - now all the trampolines use a `PanicTrap` to guard against any panics leaking out of them.

adamreichold: This does put the `PanicTrap` closer to the FFI boundary itself, which is nice.

But I think the `getter` and `setter` trampolines are not covered by this? Couldn't they use `trampoline_inner` as well?

Also, wouldn't `pub(crate)` suffice for this function as the only placing which calls this outside of `mod trampoline` is within `mod pyclass`?

Personally, I think I would have felt safer if the `PanicTrap` was installed inside `panic_result_into_callback_output`, but admittedly that function itself does not imply an FFI boundary. Are there any places we use this function outside of FFI interactions at the moment?

davidhewitt: Ah, yes corrected `getter` and `setter` and made it `pub(crate)`.

Regarding `panic_result_into_callback_output`, I somewhat felt similar but on reflection I decided that the safety issue is not the panic payload drop panicking, but that secondary panic then making it past the FFI boundary unprotected. And on that reflection it seemed clearer to me that the responsibility for `PanicTrap` is with these trampolines.

adamreichold: > And on that reflection it seemed clearer to me that the responsibility for PanicTrap is with these trampolines.

Agreed. We might want to make another sweep over all users of `panic_result_into_callback_output` though whether those need similar treatment.

From a quick grep, it might make sense to investigate

* `ModuleDef::module_init` (Might be benign as no custom code is called, but why the `catch_unwind` then in the first place?)
* `macro_rules! $generate_macro`
* `macro_rules! generate_pyclass_getattro_slot`

Also, we might wanto adjust the module-level doc comment for `mod pycell` which discusses the generated code. (Or we might not want to do that since the trampolines are an optimization and would only complicate the exposition without adding anything to the discussion of aliasing and mutability aimed at over there.)

davidhewitt: Great observation. I've pushed a commit which makes `panic_result_into_callback_output` a private function in `trampoline.rs` and converted all callsites listed above to also make use of trampolines.

I did adjust the `mod pycell` documentation a bit, given I removed `panic_result_into_callback_output` from the (hidden) crate exports.

adamreichold: This was certainly sweeping! I feel much better with `panic_result_into_callback_output` being a module-private function since almost everything related to crossing the FFI boundary is now here.

PR #2708, opened 2022-10-25
packaging: support Python 3.11

davidhewitt:
This changes CI to run against released builds of Python 3.11.

No code changes are expected because we've already been testing for a while, however it also announces official support.

PR #2709, opened 2022-10-25
ci: add Python 3.12-dev jobs

davidhewitt:
Separately to #2708, time to start testing 3.12 alphas on CI. I imagine this might also take a while for the packages to come available.


davidhewitt:
Nearly there, however I came across a crash which reported upstream: https://github.com/python/cpython/issues/99708

davidhewitt:
blocked on #2817

davidhewitt:
I think it was fixed for 3.11 but the CPython developers intended to stick with the new changes for 3.12. It might be worth a rebase and rerun although I'm thinking we might want to consider updating to use 3.8's python interpreter initialization APIs and see if that unblocks us.

adamreichold:
> I'm thinking we might want to consider updating to use 3.8's python interpreter initialization APIs and see if that unblocks us.

I guess this would be something to do post a 0.19 release? Probably together with our MSRV bump when 3.7 reaches EoL and Bookworm hits the shelves.

adamreichold:
bors try

adamreichold:
bors try

adamreichold:
Looks like our string handling needs further work still, at least the necessary changes to make `ffi::tests::ascii` and `ffi::tests::ucs4` are not obvious to me.

davidhewitt:
@adamreichold I think this branch is now at a state where I'd like to propose we merge it, the falling 3.12 problems shouldn't block us in CI and then we can identify and fix them individually.

I think this same setting is what we can also apply for 3.7 so that contributors don't get blocked by the extended support we're trialling.

PR #2712, opened 2022-10-29
Also bump CI matrix include jobs to latest stable CPython version.

adamreichold:


adamreichold:
Not sure where those requirements are configured? Direclty in the GitHub UI?

grafik

.github/workflows/ci.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       - name: Run cargo checks
         run: |
           set -x
-          VERSIONS=("3.7" "3.8" "3.9" "3.10")
+          VERSIONS=("3.7" "3.8" "3.9" "3.10" "3.10" "3.11")
```
davidhewitt, line 69: Looks like 3.10 unintentionally duplicated

```suggestion
          VERSIONS=("3.7" "3.8" "3.9" "3.10" "3.11")
```

davidhewitt (commented):
Thanks, yes we'll need to update in the GitHub UI just before merging this (and everything else then has to be rebased).

davidhewitt (approved):

.github/workflows/ci.yml
```diff
@@ -409,7 +409,7 @@ jobs:
       - uses: actions/setup-python@v4
         id: setup-python
         with:
-          python-version: 3.11.0-beta.4
+          python-version: "3.11"
```
messense, line 412: https://github.com/PyO3/pyo3/blob/6766d9f93bd2f4130953627d7dc6e8758acc8e16/emscripten/Makefile#L6

I think this also needs to be updated.

adamreichold: Good catch! Removed the variable outright.

messense (approved):

messense:
> Not sure where those requirements are configured? Direclty in the GitHub UI for protected branch rules?

Updated.

davidhewitt:
Thanks both - my family and I were out of action sick for the past week, better now thankfully, just catching up on everything I missed!

PR #2715, opened 2022-10-31
Disable `PyModule::filename` on PyPy

messense:
See https://github.com/PyO3/maturin/pull/1214#issuecomment-1296705251


adamreichold (approved):
I think this is a good candidate for inclusion into 0.17.3.

messense:
Merging, I've confirmed that this patch fixes the test failure with PyPy on macOS: https://github.com/PyO3/maturin/actions/runs/3366045809/jobs/5582152498

PR #2717, opened 2022-11-01
Fix module name in getting started

datapythonista:
If I'm not missing anything, the docs of this same function seem to say that the function name needs to be the library name in cargo, and the python module that will be exported. But the function name is different from those.


messense:
Good catch, thanks!

PR #2721, opened 2022-11-04
Update ui tests for Rust 1.65.0

messense:


messense:
Looks like to maintain 1.48.0 MSRV, we can't use `cargo clippy --fix`, dropping https://github.com/PyO3/pyo3/pull/2721/commits/663172b221bc244e418e4cfff94e0aa4207ca184

PR #2726, opened 2022-11-08
Python 3.7 support for PyCodeObject.

### LINKED ISSUE #2725: PyCodeObject does not define fields on Python 3.7
Previously it did, but no longer.

itamarst:
See https://github.com/python/cpython/blob/3.7/Include/code.h#L21 for the original C code for validation.

Fixes #2725

### DIFF
--- a/newsfragments/2727.changed.md
+++ b/newsfragments/2727.changed.md
@@ -1,1 +1,0 @@
-`PyCodeObject` is now defined with fields on Python 3.7.
--- a/newsfragments/2727.changed.md
+++ b/newsfragments/2727.changed.md
@@ -1,0 +1,1 @@
+`PyCodeObject` is now defined with fields on Python 3.7.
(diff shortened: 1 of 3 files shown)


davidhewitt (approved):
Thanks for identifying & fixing this!

newsfragments/2727.changed.md
```diff
@@ -0,0 +1 @@
+`PyCodeObject` is now defined with fields on Python 3.7.
```
davidhewitt, line 1: This needs to be `2726` not `2727`. Maybe also better to class as "fixed" instead of "changed", if this was previously working for you?

itamarst: 2726 or the reported issue, 2725?

itamarst: Ah, the former.

itamarst:
Is there any ETA for next release, BTW? Or should I just rely on git version for now?

PR #2734, opened 2022-11-14
PyAny: deprecate cast_as() in favor of downcast()

birkenfeld:
They are (practically) identical, and `cast_as()` is already present on `Py<...>` as well, so I chose to keep that name for the future, although it reads a bit strange when no turbofished type follows.

If y'all prefer to keep `downcast`, sure, but then we should move to that name on `Py` as well.

There's also a `cast_as` on `Python` which is not quite the same, is unsafe and has a `checked_cast_as` companion...

### DIFF
--- a/newsfragments/2734.changed.md
+++ b/newsfragments/2734.changed.md
@@ -1,0 +1,2 @@
+`PyAny::cast_as()` and `Py::cast_as()` are now deprecated in favor of
+`PyAny::downcast()` and the new `Py::downcast()`.
--- a/newsfragments/2734.changed.md
+++ b/newsfragments/2734.changed.md
@@ -1,2 +1,3 @@
 `PyAny::cast_as()` and `Py::cast_as()` are now deprecated in favor of
-`PyAny::downcast()` and the new `Py::downcast()`.
+`PyAny::downcast()` and the new `Py::downcast()`. The `PyAny::downcast()`
+lifetime bounds where slightly relaxed.
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -113,7 +113,7 @@ pub unsafe trait PyNativeType: Sized {
 /// #         let m = pyo3::types::PyModule::new(py, "test")?;
 /// #         m.add_class::<Foo>()?;
 /// #
-/// #         let foo: &PyCell<Foo> = pyo3::PyTryFrom::try_from(m.getattr("Foo")?.call0()?)?;
+/// #         let foo: &PyCell<Foo> = m.getattr("Foo")?.call0()?.downcast()?;
 /// #         let dict = &foo.borrow().inner;
 /// #         let dict: &PyDict = dict.as_ref(py);
 /// #
@@ -150,7 +150,7 @@ pub unsafe trait PyNativeType: Sized {
 /// #         let m = pyo3::types::PyModule::new(py, "test")?;
 /// #         m.add_class::<Foo>()?;
 /// #
-/// #         let foo: &PyCell<Foo> = pyo3::PyTryFrom::try_from(m.getattr("Foo")?.call0()?)?;
+/// #         let foo: &PyCell<Foo> = m.getattr("Foo")?.call0()?.downcast()?;
 /// #         let bar = &foo.borrow().inner;
 /// #         let bar: &Bar = &*bar.borrow(py);
 /// #
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -995,11 +995,20 @@ impl PyObject {
     ///
     /// This can cast only to native Python types, not types implemented in Rust. For a more
     /// flexible alternative, see [`Py::extract`](struct.Py.html#method.extract).
+    pub fn downcast<'p, T>(&'p self, py: Python<'p>) -> Result<&T, PyDowncastError<'_>>
+    where
+        for<'py> T: PyTryFrom<'py>,
+    {
+        <T as PyTryFrom<'_>>::try_from(self.as_ref(py))
+    }
+
+    /// Casts the PyObject to a concrete Python object type.
+    #[deprecated(since = "0.18.0", note = "use downcast() instead")]
     pub fn cast_as<'p, D>(&'p self, py: Python<'p>) -> Result<&'p D, PyDowncastError<'_>>
     where
         D: PyTryFrom<'p>,
     {
-        <D as PyTryFrom<'_>>::try_from(unsafe { py.from_borrowed_ptr::<PyAny>(self.as_ptr()) })
+        <D as PyTryFrom<'_>>::try_from(self.as_ref(py))
     }
 }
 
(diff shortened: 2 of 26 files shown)


adamreichold:
I feel bad since you already did all the work, but I have to admit that I would prefer the `downcast` name as it more closely describes the use case (going from generic to specific) and is probably the first thing I would search for when to trying to accomplish this.

I agree about the need for consistency as to `Py<..>`. The existence of `Python::cast_as` with a different use case would be another argument for `downcast` from my point of view.

birkenfeld:
no matter, if others agree it's easily changed over!

mejrs:
I also prefer `downcast`. It's probably what people will be looking for, and it's also something that people expect to be fallible, which people don't expect with something named `cast_as`, I think.

birkenfeld:
Ok, that should do it. Unfortunately CI has some unrelated warnings in `chrono`...

newsfragments/2734.changed.md
```diff
@@ -0,0 +1,2 @@
+`PyAny::cast_as()` and `Py::cast_as()` are now deprecated in favor of
```
davidhewitt, line 1: Maybe also have a `2734.added.md` just to make it clear that `Py::downcast` is new?

davidhewitt:
Agreed that `downcast` is what I tend to use too, so deprecating `cast_as` makes sense to me.

I've still been struggling with lingering illness & sleepless baby, so I haven't got around to fixing CI, I think it's just some deprecations from the newest `chrono` release which should be straightforward to fix. I might be able to have a stab at it tomorrow evening 🤞

src/instance.rs
```diff
@@ -995,12 +995,21 @@ impl PyObject {
     ///
     /// This can cast only to native Python types, not types implemented in Rust. For a more
     /// flexible alternative, see [`Py::extract`](struct.Py.html#method.extract).
-    pub fn cast_as<'p, D>(&'p self, py: Python<'p>) -> Result<&'p D, PyDowncastError<'_>>
+    pub fn downcast<'p, D>(&'p self, py: Python<'p>) -> Result<&'p D, PyDowncastError<'_>>
     where
         D: PyTryFrom<'p>,
     {
         <D as PyTryFrom<'_>>::try_from(unsafe { py.from_borrowed_ptr::<PyAny>(self.as_ptr()) })
```
adamreichold, line 1002: Could the inner `unsafe` block be replaced by calling `self.as_ref(py)`?

src/types/any.rs
```diff
@@ -769,16 +746,40 @@ impl PyAny {
...
-    /// Casts `self` to a concrete Python object type.
-    ///
-    /// This can cast only to native Python types, not types implemented in Rust.
+    /// Converts this `PyAny` to a concrete Python type.
+    #[deprecated(since = "0.18.0", note = "use the equivalent .downcast()")]
     pub fn cast_as<'a, D>(&'a self) -> Result<&'a D, PyDowncastError<'_>>
     where
         D: PyTryFrom<'a>,
     {
         <D as PyTryFrom<'_>>::try_from(self)
```
adamreichold, line 755: I think it would be nice if this implementation would forward to `self.downcast` just to make it obvious that they do the same thing.

src/types/any.rs
```diff
@@ -769,16 +746,40 @@ impl PyAny {
...
+    ///     let dict = PyDict::new(py);
+    ///     assert!(dict.is_instance_of::<PyAny>().unwrap());
+    ///     let any: &PyAny = dict.as_ref();
+    ///     assert!(any.downcast::<PyDict>().is_ok());
+    ///     assert!(any.downcast::<PyList>().is_err());
+    /// });
+    /// ```
+    pub fn downcast<T>(&self) -> Result<&T, PyDowncastError<'_>>
+    where
+        for<'py> T: PyTryFrom<'py>,
```
adamreichold, line 778: Do we actually need the quantifier here? Wouldn't

```rust
pub fn downcast<'py, T>(&'py self) -> Result<&'py T, PyDowncastError<'_>>
where
    T: PyTryFrom<'py>,
```

suffice?

src/types/any.rs
birkenfeld: Not possible, since the bounds differ slightly...

src/instance.rs
birkenfeld: Done, thanks.

src/types/any.rs
birkenfeld: That's the signature of `cast_as`. I don't actually know if there's any meaningful difference here - so I didn't want to change `downcast`'s signature needlessly.

src/types/any.rs
adamreichold: See https://github.com/PyO3/pyo3/pull/2734#discussion_r1023580363, but then that would be a breaking change even though I don't think the would be real-world breaking due to relaxing the bound.

src/types/any.rs
birkenfeld: Agreed.

adamreichold (approved):

src/types/any.rs
davidhewitt: I'm actually not a fan of `PyTryFrom` at all - the `try_from` method it introduces also clashes with the 2021 prelude. I wonder if actually we want

```rust
pub fn downcast<'py, T>(&'py self) -> Result<&'py T, PyDowncastError<'_>>
where
    T: PyTypeInfo
```

or _maybe_

```rust
pub fn downcast<'py, T>(&'py self) -> Result<&'py T, PyDowncastError<'_>>
where
    T: PyNativeType,
    &'py T: FromPyObject<'py>
```

... which might open up the way for us to get rid of `PyTryFrom`? (We'd perhaps want to also have an `unsafe fn downcast_unchecked` too.)

---

On the original topic, yes I agree with correcting the lifetime bound as proposed.

src/instance.rs
```diff
@@ -995,11 +995,20 @@ impl PyObject {
     ///
     /// This can cast only to native Python types, not types implemented in Rust. For a more
     /// flexible alternative, see [`Py::extract`](struct.Py.html#method.extract).
+    pub fn downcast<'p, T>(&'p self, py: Python<'p>) -> Result<&T, PyDowncastError<'_>>
```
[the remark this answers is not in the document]
birkenfeld: Do we have another way to get from `Py<A>` to `Py<B>`?

birkenfeld:
If `Py::downcast` returns `Py`, probably it's a good idea to also add `downcast_ref()`; the operation is quite frequent.

birkenfeld:
I've added commits that replace usage of `try_from`, add `downcast_unchecked` and remove usage of `try_from_unchecked`.

src/instance.rs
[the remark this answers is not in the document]
davidhewitt: I think `.extract()` might do it though not immediately obvious. For `B = PyAny` there's `.into()`.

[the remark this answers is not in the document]
davidhewitt: Also with this downcast could do `.downcast::<B>().into()` to use `&B -> Py<B>`, maybe that's good enough?

davidhewitt (approved):
This is looking great to me, many thanks!

PR #2739, opened 2022-11-15
error when `#[pyo3(signature = ())]` used on invalid methods

davidhewitt:
A follow-up to #2702 to reject some invalid applications of `#[pyo3(signature = (...))]` attribute, specifically on magic methods and getters / setters / class attributes.


pyo3-macros-backend/src/method.rs
```diff
@@ -742,3 +722,29 @@ const IMPL_TRAIT_ERR: &str = "Python functions cannot have `impl Trait` argument
...
+    if let Some(signature) = signature {
+        if let FnType::Getter(_) | FnType::Setter(_) | FnType::ClassAttribute = fn_type {
+            bail_spanned!(signature.kw.span() => "signature not allowed with this method type")
+        }
+    }
+    if let Some(text_signature) = text_signature {
+        match &fn_type {
+            FnType::FnNew => bail_spanned!(
+                text_signature.kw.span() =>
+                "text_signature not allowed on __new__; if you want to add a signature on \
```
adamreichold, line 740: IIRC, the plan is to phase out `text_signature` entirely, isn't it? Does this imply we need to support `signature` on `#[pyclass]`? Or is this already supported?

pyo3-macros-backend/src/method.rs
```diff
@@ -742,3 +722,29 @@ const IMPL_TRAIT_ERR: &str = "Python functions cannot have `impl Trait` argument
...
 Try `&self`, `&mut self, `slf: PyRef<'_, Self>` or `slf: PyRefMut<'_, Self>`.";
+
+fn ensure_signatures_on_valid_method(
+    fn_type: &FnType,
+    signature: Option<&SignatureAttribute>,
+    text_signature: Option<&TextSignatureAttribute>,
+) -> syn::Result<()> {
+    if let Some(signature) = signature {
+        if let FnType::Getter(_) | FnType::Setter(_) | FnType::ClassAttribute = fn_type {
+            bail_spanned!(signature.kw.span() => "signature not allowed with this method type")
```
adamreichold, line 733: Since the span is on the signature and not the getter/setter/class attribute, I think it would be helpful to call out the method type in question.

pyo3-macros-backend/src/method.rs
davidhewitt: I think `text_signature` can continue to exist but I'd like to autogenerate them by default, we'd probably still support the existing form as an override and maybe `text_signature = false` to turn it off.

I was actually thinking the nicer solution here might be to move `text_signature` to work on `#[new]` and deprecate support for it on the class itself? It seems much more intuitive to me to put the signature documentation on the `#[new]`, after all. We could probably make that work with some macro mumbo-jumbo. Potentially this would be necessary if we wanted to autogenerate `text_signature` for `#[new]`.

adamreichold: > It seems much more intuitive to me to put the signature documentation on the #[new], after all.

I think so too.

> I was actually thinking the nicer solution here might be to move text_signature to work on #[new] and deprecate support for it on the class itself? 

And `signature` as well? So basically the same behaviour as for other `#[pymethods]` with `text_signature` derived from `signature`?

adamreichold (approved):

PR #2740, opened 2022-11-15
Make Clippy happy again

adamreichold:
grafik

I am not completely convinced those casts are unnecessary on all platforms but nightly Clippy is not happy. I guess the CI will decide...

### DIFF
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -127,11 +127,12 @@ impl IntoPy<PyObject> for NaiveDate {
 impl FromPyObject<'_> for NaiveDate {
     fn extract(ob: &PyAny) -> PyResult<NaiveDate> {
         let date = <PyDate as PyTryFrom>::try_from(ob)?;
-        Ok(NaiveDate::from_ymd(
+        Ok(NaiveDate::from_ymd_opt(
             date.get_year(),
             date.get_month() as u32,
             date.get_day() as u32,
-        ))
+        )
+        .expect("invalid or out-of-range date"))
     }
 }
 
@@ -164,7 +165,7 @@ impl FromPyObject<'_> for NaiveTime {
         let h = time.get_hour() as u32;
         let m = time.get_minute() as u32;
         let s = time.get_second() as u32;
-        Ok(NaiveTime::from_hms_micro(h, m, s, ms))
+        Ok(NaiveTime::from_hms_micro_opt(h, m, s, ms).expect("invalid or out-of-range date"))
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/conversions/chrono.rs
```diff
@@ -127,11 +127,12 @@ impl IntoPy<PyObject> for NaiveDate {
...
     fn extract(ob: &PyAny) -> PyResult<NaiveDate> {
         let date = <PyDate as PyTryFrom>::try_from(ob)?;
-        Ok(NaiveDate::from_ymd(
+        Ok(NaiveDate::from_ymd_opt(
             date.get_year(),
             date.get_month() as u32,
             date.get_day() as u32,
-        ))
+        )
+        .expect("invalid or out-of-range date"))
```
adamreichold, line 135: This is the absolute minimum change by basically copying over the code from `chrono`. (Personally, I am not convinced the deprecations in `chrono` are entirely helpful.)

We might want to use this occasion to turn these panics into `PyTypeError` but if I remember the discussion this should never happen as the ranges of `PyDate` and `NaiveData` more or less match up?

adamreichold:
> I am not completely convinced those casts are unnecessary on all platforms but nightly Clippy is not happy. I guess the CI will decide...

Seems like Clippy was right all along and I feel bad for questioning it...

messense (approved):

adamreichold:
@messense Thanks for the review! Note that I'll delay merging this until @davidhewitt had a change to consider https://github.com/PyO3/pyo3/pull/2740#discussion_r1023257127

src/conversions/chrono.rs
davidhewitt: I generally would argue that panics in a library are only ok if it's almost guaranteed that user code won't trigger them; on that basis I think the `chrono` APIs can be easily misused and so the deprecation seems reasonable (I wonder if they plan to replace them with forms that return errors instead in future).

As for here, I think it's fine to preserve behaviour in this PR however I think given we can't guarantee that these panics will never fire (especially if e.g. `chrono` changes the range it can handle), it's probably good to replace these with `PyValueError` as you suggest. Perhaps leave as an issue for a follow-up PR?

davidhewitt (approved):
Many thanks for sorting this!

adamreichold:
> I wonder if they plan to replace them with forms that return errors instead in future.

Exactly this. Just deprecating the short forms appears questionable to me if the `_opt` variants are the only alternative. Now every project starts open-coding the same `expect` calls. Just turning the default methods into ones returning `Result` would probably been more useful if technically breaking and we could have handled the errors in a systematic manner. 

> Perhaps open an issue to remind us to write a follow-up PR?

Will do.

PR #2742, opened 2022-11-16
Export warning classes and add PyErr::warn_explicit()

### LINKED ISSUE #2741: export `PyUserWarning`
Just as other exceptions are available from `pyo3::exceptions`.

Perhaps other warnings?

Perhaps from `pyo3::warnings`?

For know, I just need `UserWarning`.

In the meantime, I'll import it manually.

birkenfeld:
Fixes #2741

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -477,7 +477,28 @@ impl PyErr {
     }
 
     /// Issues a warning message.
-    /// May return a `PyErr` if warnings-as-errors is enabled.
+    ///
+    /// May return an `Err(PyErr)` if warnings-as-errors is enabled.
+    ///
+    /// Equivalent to `warnings.warn()` in Python.
+    ///
+    /// The `category` should be one of the `Warning` classes available in
+    /// [`pyo3::exceptions`](crate::exceptions), or a subclass.  The Python
+    /// object can be retrieved using [`PyTypeInfo::type_object()`].
+    ///
+    /// Example:
+    /// ```rust
+    /// use pyo3::prelude::*;
+    /// use pyo3::PyTypeInfo;
+    ///
+    /// # fn main() -> PyResult<()> {
+    /// Python::with_gil(|py| {
+    ///     let user_warning = pyo3::exceptions::PyUserWarning::type_object(py);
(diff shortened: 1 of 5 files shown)


samuelcolvin (commented):
LGTM, two comments:

* would it be clearer if we moved warnings to `pyo3::warning`?
* How hard would it be to add a method so we could call `PyErr::warn_from_type::<PyUserWarning>(....)` (like you can do `dict.is_instance_of::<PyAny>`)?

birkenfeld:
1. there wouldn't be much in it, and since Warnings are Exceptions anyway...
2. possible, but I feel the combinatoric explosion of methods (what about a counterpart for `warn_explicit`...) is not worth the few times when it is needed.

Of course, these are my opinions, other maintainers may disagree.

src/err/mod.rs
```diff
@@ -492,6 +511,48 @@ impl PyErr {
...
+            Some(s) => s.as_ptr(),
+        };
+        let registry: *mut ffi::PyObject = match registry {
+            None => std::ptr::null_mut(),
+            Some(obj) => obj.as_ptr(),
+        };
+        unsafe {
+            error_on_minusone(
+                py,
+                ffi::PyErr_WarnExplicit(
```
davidhewitt, line 544: It looks like `PyPyErr_WarnExplicit` does exist as a symbol in PyPy shared lib, so I think we can update the ffi definition and enable `warn_explicit` on pypy.

src/err/mod.rs
```diff
@@ -477,7 +477,26 @@ impl PyErr {
     }
 
     /// Issues a warning message.
```
davidhewitt, line 479: Perhaps worth expanding this doc to say it's equivalent to `warnings.warn` (and similar for `warn_explicit`)?

davidhewitt (commented):
Thanks for this! I'm ok with `pyo3::exceptions` as a reasonable home for these for now, though I'm also not sure either way. It should be possible to re-home them in the future if we want.

As for the typed methods, I'm unconvinced we need them for now - users can always define wrappers e.g. `user_warning(py, msg)` to do exactly what you want.

davidhewitt (approved):
Looks good to me, thanks!

PR #2749, opened 2022-11-19
Fix being able to call arg-less `#[new]` with any args from Python

### LINKED ISSUE #2748: Generated __new__ method does not check args to enforce signature
### Bug Description

Firstly, thank you for creating this amazing crate!

I noticed today that when using `#[new]` to generate a `__new__` method with an empty signature, the generated code doesn't enforce that _args and _kwargs must be empty. As a result, it's possible to call the constructor from Python and provide unused arguments without raising an exception. This could be confusing, for example the developer might expect the kwargs to set field values. I think this is a bug.

Here's part of the code generated by the macros, from `cargo expand`:

```
    impl Foo {
        unsafe extern "C" fn __pymethod___new____(
            subtype: *mut _pyo3::ffi::PyTypeObject,
            _args: *mut _pyo3::ffi::PyObject,
            _kwargs: *mut _pyo3::ffi::PyObject,
        ) -> *mut _pyo3::ffi::PyObject {
            use _pyo3::{callback::IntoPyCallbackOutput, pyclass_init::PyObjectInit};
            let gil = _pyo3::GILPool::new();
            let _py = gil.python();
            _pyo3::callback::panic_result_into_callback_output(
                _py,
                ::std::panic::catch_unwind(move || -> _pyo3::PyResult<_> {
                    let result = Foo::new();
                    let initializer: _pyo3::PyClassInitializer<Foo> = result.convert(_py)?;
                    initializer.into_new_object(_py, subtype)
                }),
            )
        }
    }
```

### Steps to Reproduce

```
#![allow(unused_variables)]
use pyo3::prelude::*;
use pyo3::types::IntoPyDict;

#[pyclass]
struct Foo {}

#[pymethods]
impl Foo {
    #[new]
    pub fn new() -> Self {
        Foo {}
    }
}

pub fn main() -> PyResult<()> {
    Python::with_gil(|py| {
    	let locals = [("Foo", py.get_type::<Foo>())].into_py_dict(py);
    	py.run("print(Foo()); print(Foo(123)); print(Foo(bar=123))", Some(locals), None)
    })
}
```

This code prints three objects, without raising any exceptions. The expected behavior would be to print the first and raise an exception on the second (and third of course).

### Backtrace

_No response_

WSL2

3.11.0

rustc 1.60.0-nightly (22e491ac7 2022-01-13)

### Your PyO3 version

0.17.3

venv

### Additional Info

_No response_

birkenfeld:
Fixes #2748

Note that the removed check only triggered for `#[new]` anyway, since other arg-less pymethods use the `NOARGS` argument convention which is not available for `tp_new`.

There's some confusion going on with the `py` argument, I'll open an issue for that.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -233,7 +233,7 @@ Consult the table below to determine which type your constructor should return:
 
 ## Inheritance
 
-By default, `PyAny` is used as the base class. To override this default,
+By default, `object`, i.e. `PyAny` is used as the base class. To override this default,
 use the `extends` parameter for `pyclass` with the full path to the base class.
 
 For convenience, `(T, U)` implements `Into<PyClassInitializer<T>>` where `U` is the
@@ -310,7 +310,8 @@ impl SubSubClass {
 ```
 
 You can also inherit native types such as `PyDict`, if they implement
-[`PySizedLayout`]({{#PYO3_DOCS_URL}}/pyo3/type_object/trait.PySizedLayout.html). However, this is not supported when building for the Python limited API (aka the `abi3` feature of PyO3).
+[`PySizedLayout`]({{#PYO3_DOCS_URL}}/pyo3/type_object/trait.PySizedLayout.html).
+This is not supported when building for the Python limited API (aka the `abi3` feature of PyO3).
 
 However, because of some technical problems, we don't currently provide safe upcasting methods for types
 that inherit native types. Even in such cases, you can unsafely get a base class by raw pointer conversion.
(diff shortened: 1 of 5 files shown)


birkenfeld:
Oh great, another MSRV bump (`rayon-core`) in a semver-compatible version bump. To the 2021 edition nonetheless.

davidhewitt (approved):
Ooof, thanks for fixing,!

birkenfeld:
Hmm, that test failure is interesting. The base class' `__new__` is called implicitly, it seems, with the same arguments?

birkenfeld:
Added a bit to the guide that explains the behavior.

davidhewitt (approved):
Thanks, this is definitely a correct fix and it's good to document existing behaviour.

I think there's potential for improvement around how to control invocation of `__new__` and `__init__` with inheritance, however I don't have a good answer for that right now and it's definitely for a future discussion anyway.

PR #2751, opened 2022-11-19
guide: better document negative index behavior w.r.t. PySequence

### LINKED ISSUE #2601: docs: `#[pyclass(sequence)]` does not make Python add the sequence length to negative indexes
### Bug Description

https://pyo3.rs/v0.17.1/class/protocols.html#mapping--sequence-types

> Use the #[pyclass(sequence)] annotation to instruct PyO3 to fill the sq_length slot instead of the mp_length slot for __len__. This will help libraries such as numpy recognise the class as a sequence, however will also cause CPython to automatically add the sequence length to any negative indices before passing them to __getitem__. (__getitem__, __setitem__ and __delitem__ mapping slots are still used for sequences, for slice operations.)

What actually happens is that negative indices are passed as is to `__getitem__`.

I guess this happens because PyO3 only fills the mapping slot for `__getitem__`, and the negative index correction thing only happens when calling the sequence slot for `__getitem__`.

### Steps to Reproduce

Minimal reproduction:

```rust
use pyo3::prelude::*;                                  
                                                       
#[pyclass(sequence)]                                   
struct MyList;                                         
                                                       
#[pymethods]                                           
impl MyList {                                          
    #[new]                                             
    fn new() -> MyList {                               
        MyList {}                                      
    }                                                  
                                                       
    fn __len__(&self) -> usize {                       
        42                                             
    }                                                  
                                                       
    fn __getitem__(&self, idx: isize) -> bool {        
        assert!(idx >= 0);                             
        true                                           
    }                                                  
}                                                      
                                                       
#[pymodule]                                            
fn testmod(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<MyList>()?;                          
    Ok(())                                             
}                                                      
```

### Backtrace

```shell
>>> from testmod import MyList
>>> a = MyList()
>>> a[-1]
thread '<unnamed>' panicked at 'assertion failed: idx >= 0', src/lib.rs:18:9
pyo3_runtime.PanicException: assertion failed: idx >= 0
>>> a[0]
True
```
```


Arch Linux

Python 3.10.6

rustc 1.63.0

### Your PyO3 version

0.17.1

pacman

Yes

### Additional Info

_No response_

birkenfeld:
Fixes #2601

### DIFF
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -243,17 +243,24 @@ Use the `#[pyclass(sequence)]` annotation to instruct PyO3 to fill the `sq_lengt
 
     Implements retrieval of the `self[a]` element.
 
-    *Note:* Negative integer indexes are not handled specially.
+    *Note:* Negative integer indexes are not handled specially by PyO3.
+    However, for classes with `#[pyclass(sequence)]`, when a negative index is
+    accessed via `PySequence::get_item`, the underlying C API already adjusts
+    the index to be positive.
 
   - `__setitem__(<self>, object, object) -> ()`
 
     Implements assignment to the `self[a]` element.
     Should only be implemented if elements can be replaced.
 
+    Same behavior regarding negative indices as for `__getitem__`.
+
   - `__delitem__(<self>, object) -> ()`
 
     Implements deletion of the `self[a]` element.
     Should only be implemented if elements can be deleted.
+
+    Same behavior regarding negative indices as for `__getitem__`.
 
   * `fn __concat__(&self, other: impl FromPyObject) -> PyResult<impl ToPyObject>`
 


davidhewitt:
Possibly worth mentioning that this negative index handling only applies with `#[pyclass(sequence)]`?

birkenfeld:
Updated.

birkenfeld:
I don't understand the test failure...

birkenfeld:
@messense I see that maturin 0.14 has been released recently. Can this be a problem with it?

messense:
> @messense I see that maturin 0.14 has been released recently. Can this be a problem with it?

Yes, it's resolved by https://github.com/PyO3/maturin/pull/1279.

For now you can pin maturin to 0.13.x in 
https://github.com/PyO3/pyo3/blob/3408cc420052b9a61cbe535c9f7229982f0d20a5/pytests/noxfile.py#L12

birkenfeld:
Do we need to do this or will the fix be released soon?

messense:
> Do we need to do this or will the fix be released soon?

I'll cut a new release now, but it's good idea to add a version constraint there anyway.

davidhewitt (approved):
Thanks 👍

PR #2752, opened 2022-11-19
ffi: link to _PyPyBytes_Resize

### LINKED ISSUE #2737: Adding _PyBytes_Resize for PyPy
Hi there,

Would it be possible to link against PyPy's version of [_PyBytes_Resize](https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.9/pypy/module/cpyext/bytesobject.py#L191) somewhere [here](https://github.com/PyO3/pyo3/blob/main/pyo3-ffi/src/bytesobject.rs#L22-L56)? 

I think that would be something like
```rust
#[cfg_attr(PyPy, link_name = "Py_PyBytes_Resize")]
pub fn _PyBytes_Resize(arg1: *mut PyObject, arg2: Py_ssize_t) -> c_int;
```

Big thanks!

birkenfeld:
Fixes #2737

### DIFF
--- a/pyo3-ffi/src/cpython/bytesobject.rs
+++ b/pyo3-ffi/src/cpython/bytesobject.rs
@@ -15,5 +15,6 @@ pub struct PyBytesObject {
 opaque_struct!(PyBytesObject);
 
 extern "C" {
+    #[cfg_attr(PyPy, link_name = "_PyPyBytes_Resize")]
     pub fn _PyBytes_Resize(bytes: *mut *mut PyObject, newsize: Py_ssize_t) -> c_int;
 }


messense (approved):

darleybarreto:
When compiling for PyPy, `_PyBytes_Resize` wont be available, I think we could move the definition of this function out of `cpython` folder, like adding it here (or somewhere else in src/bytesobject.rs):
```rust
#[cfg_attr(windows, link(name = "pythonXY"))]
extern "C" {
    #[cfg_attr(PyPy, link_name = "PyPyBytes_Type")]
    pub static mut PyBytes_Type: PyTypeObject;
    pub static mut PyBytesIter_Type: PyTypeObject;
    #[cfg_attr(PyPy, link_name = "_PyPyBytes_Resize")]
    pub fn _PyBytes_Resize(bytes: *mut *mut PyObject, newsize: Py_ssize_t) -> c_int;
}
```

birkenfeld:
Damn, I saw there were already `PyPy` conditionals in `bytesobject.rs`, so I didn't check. Removing the conditional in`mod.rs` seems to be the right course of action.

birkenfeld:
See also #2758

messense:
The CI failure on main is caused by https://github.com/oli-obk/cargo_metadata/issues/212

PR #2753, opened 2022-11-19
CI: fix rayon-core version to keep MSRV compat

birkenfeld:


.github/workflows/ci.yml
```diff
@@ -182,6 +182,8 @@ jobs:
           cargo update -p plotters-backend --precise 0.3.2
           cargo update -p bumpalo --precise 3.10.0
           cargo update -p once_cell --precise 1.14.0
+          cargo update -p rayon --precise 1.5.3
```
adamreichold, line 185: Since we use it in the `word-count`, it should use `--manifest-path "examples(word-count/Cargo.toml" I think? (Not sure if we need it for the `dev-dependency` of the top-level project at all?)

birkenfeld: I removed it there. However it still needs to be version-adjusted in toplevel since it's a dependency of `criterion`.

birkenfeld: Oops, it's still used in a docstring.

davidhewitt (approved):
Thanks!

PR #2756, opened 2022-11-19
Two smallish cleanups

birkenfeld:
RFC?

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -995,6 +995,7 @@ impl PyObject {
     ///
     /// This can cast only to native Python types, not types implemented in Rust. For a more
     /// flexible alternative, see [`Py::extract`](struct.Py.html#method.extract).
+    #[inline]
     pub fn downcast<'p, T>(&'p self, py: Python<'p>) -> Result<&T, PyDowncastError<'_>>
     where
         T: PyTryFrom<'p>,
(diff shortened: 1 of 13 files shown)


src/instance.rs
```diff
@@ -995,6 +995,7 @@ impl PyObject {
     ///
     /// This can cast only to native Python types, not types implemented in Rust. For a more
     /// flexible alternative, see [`Py::extract`](struct.Py.html#method.extract).
+    #[inline]
     pub fn downcast<'p, T>(&'p self, py: Python<'p>) -> Result<&T, PyDowncastError<'_>>
```
mejrs, line 999: `#[inline]` here does nothing, because generic, functions are already implicitly inlineable (same for every other inline in this PR)

davidhewitt: Actually I think there is still some benefit - it's true that llvm may always choose to inline generics but adding `#[inline]` does add `inline` hint to the llvm code so may encourage llvm to inline when it might have decided not to.

davidhewitt:
I've slightly preferred `T::type_object(py)` to `py.get_type::<T>()` in my own code, however agree we don't need two ways to do this so happy to agree on what the majority prefers.

birkenfeld:
If `PyTypeInfo` were in the `prelude` it might be different, but as it is now, it feels strange to have to import this extra trait for that simple operation...

davidhewitt:
👍 that's a good point, and I'm not entirely confident we want `PyTypeInfo` in the prelude, so that's a compelling reason to me.

PR #2759, opened 2022-11-20
ffi: use bytesobject/dictobject on PyPy

### LINKED ISSUE #2758: Conditional compilation confusion in `pyo3-ffi/src/cpython`
Some submodules have `cfg(not(PyPy))` attributes in `cpython/mod.rs`, however the modules themselves also contain definitions depending on `PyPy`, e.g. `dictobject.rs`.

Someone who is more familiar with these should check this. Ideally the attributes in `mod.rs` should go away if possible.

birkenfeld:
Fixes #2758


davidhewitt (approved):
👍 many thanks for all these fixes!

PR #2760, opened 2022-11-20
macros: fix the check for applying `METH_NOARGS`

### LINKED ISSUE #2750: Check for "method has no arguments" considers `py` an argument
When deciding whether to use `METH_NOARGS` for a pymethod, the [check](https://github.com/PyO3/pyo3/blob/main/pyo3-macros-backend/src/method.rs#L203) only considers `signature.arguments.is_empty()` which includes the `py` argument.

I.e., argument-less methods with `py` get a different calling convention, which should not be necessary.

Before I endeavour to fix something: was it intended that `py` is in the `arguments` list?

birkenfeld:
to only consider the Python argument list.

Fixes #2750

### DIFF
--- a/newsfragments/2760.fixed.md
+++ b/newsfragments/2760.fixed.md
@@ -1,0 +1,2 @@
+Also apply the `NOARGS` argument convention to methods that have a single
+`py: Python` argument.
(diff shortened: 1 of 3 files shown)


davidhewitt (approved):
Thanks 👍

PR #2714, opened 2022-10-30
ci: configure bors

davidhewitt:
Just for the moment seeing how to interact with bors, will experiment with CI setup in #2710 once I understand it.

PR #2761, opened 2022-11-20
ci: configure bors

### LINKED ISSUE #2710: add bors to CI
Sounds like we are generally in favour of doing this.

@messense - are you planning to do this? Otherwise I might try to have a play next in the next week or so.

messense:
Closes #2710 
Closes #2714

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ env:
 
 jobs:
   fmt:
+    if: github.ref != 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -34,6 +35,7 @@ jobs:
         run: nox -s fmt-rust
 
   clippy:
+    if: github.ref != 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -47,7 +49,7 @@ jobs:
   check-target:
     needs: [fmt]
     runs-on: ubuntu-latest
-    if: github.event_name != 'pull_request'
+    if: ${{ github.event_name != 'pull_request' && github.ref != 'refs/heads/main' }}
     strategy:
       # If one platform fails, allow the rest to keep testing if `CI-no-fail-fast` label is present
       fail-fast: ${{ !contains(github.event.pull_request.labels.*.name, 'CI-no-fail-fast') }}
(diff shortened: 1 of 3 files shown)


messense:
https://github.com/actions/runner/issues/1985 `matrix` doesn't work in `jobs.<job_id>.if`, so annoying.

messense:
bors try

.github/workflows/ci.yml
```diff
@@ -416,3 +485,24 @@ jobs:
...
+      - clippy
+      - check-target
+      - build
+      - valgrind
+      - careful
+      - coverage
+      - emscripten
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
```
adamreichold, line 501: Is this really necessary? Doesn't `needs` already require these job to finish successfully? Also what the reason to have `if: always()` here?

Generally speaking, I am not a fan of the manual JavaScript-based approach to CI workflows and would prefer if just split our workflows into one pull requests and for used by bors even if it means having some redundancies between the two workflow definitions if this allows us to keep the declarative approach.

I am also not sure if just adjusting the matrix of the `build` job is sufficient differentiation between the two use cases (pre-review check and pre-merge check). More fundamentally, I feel that the `build` could use some re-engineering as it has a collected a lot of different use cases in conditional steps which would work better as individual jobs IMHO.

messense: > Is this really necessary? Doesn't `needs` already require these job to finish successfully? Also what the reason to have `if: always()` here?

It gives a nicer overview: https://github.com/PyO3/pyo3/actions/runs/3508244920/jobs/5876898372#step:2:1 even when canceled due to previous failures.

Copied from https://github.com/cross-rs/cross/blob/ab99baf7f4c39403a0b033cef95d798bd0b9a940/.github/workflows/ci.yml#L361-L373

messense: > I feel that the `build` could use some re-engineering as it has a collected a lot of different use cases in conditional steps which would work better as individual jobs IMHO.

Expanding more jobs may result in more cache misses thus longer build time, we are [already over the 10GB cache limit](https://github.com/PyO3/pyo3/actions/caches).

messense:
bors try

.github/workflows/ci.yml
adamreichold: > Expanding more jobs may result in more cache misses thus longer build time, we are [already over the 10GB cache limit](https://github.com/PyO3/pyo3/actions/caches).

We could also improve on this front, considering the approach outlined in https://github.com/PyO3/pyo3/issues/2710#issuecomment-1296259994, the caches for the pull requests could become much more lightweight. (I think the issue of missing caches could also be solved by running the lightweight on-pull-request workflow as an on-push event on `main`? The heavyweight jobs would only run on `trying`/`staging` I guess?)

messense: > run check (not test) for each OS / Python version, and then one run of the test suite on each OS w. latest Python

It's not really lightweight IMO, I'd drop the `run check (not test) for each OS / Python version` part, and coverage jobs are even heavier but they should be kept to ensure good test coverage.

> I think the issue of missing caches could also be solved by running the lightweight on-pull-request workflow as an on-push event on `main`?

It will do, but I think with bors we shouldn't re-run workflow on `main`, [bors was designed to avoid broken checks on `main` I think?](https://graydon2.dreamwidth.org/1597.html). Maybe we should explore using [sccache](https://github.com/mozilla/sccache) instead.

adamreichold: > I'd drop the run check (not test) for each OS / Python version part

I agree. I don't think full platform coverage is something we need during review.

> and coverage jobs are even heavier but they should be kept to ensure good test coverage.

As with the above, I think one coverage job should suffice. (And it could subsume the test job because the tests run to produce the coverage.)

> It will do, but I think with bors we shouldn't re-run workflow on main, 

Exactly, bors should only run on `trying/staging` which is `main` after merging the PR, hence one does not need to re-run bors on `main` because updating `main` to `trying/staging` is a fast-forward merge to something that was just tested.

So in summary:
* `main` and pull requests: lightweight pre-review workflow to seed cache.
* `trying/staging`: heavyweight bors pre-merge workflow. (Since bors pushes to our repo it should be able to use our caches without seeding.)

messense:
bors try

davidhewitt:
Thanks! I'm curious to see what it looks like without the conclusion job, so I've pushed that to my own branch.

messense:
> I'm curious to see what it looks like without the conclusion job

I'm a bit worried about it due to this comment in bors forum: https://forum.bors.tech/t/check-matrix/673/2

davidhewitt:
Right, good find.

In which case, I think this is a great starting point, shall we merge this one and then we can write follow-up PRs for the improvements in https://github.com/PyO3/pyo3/pull/2761#discussion_r1027721707 as well as bringing `pyo3-ffi-check` into this workflow etc?

PR #2763, opened 2022-11-20
Typos, grammar, code style, etc.

birkenfeld:


guide/src/memory.md
```diff
@@ -122,11 +122,12 @@ for more information on safety.
 ## GIL-independent memory
 
 Sometimes we need a reference to memory on Python's heap that can outlive the
-GIL.  Python's `Py<PyAny>` is analogous to `Rc<T>`, but for variables whose
+GIL.  Python's `Py<PyAny>` is analogous to `Arc<T>`, but for variables whose
```
adamreichold, line 125: But wasn't the comparison to `Rc` more apt than `Arc` since Python does not use atomic operations to modify the reference counts and relies on the GIL to synchronize access to them instead?

adamreichold (approved):
Thank you for taking the time to apply a little spit and polish! One question and the change from `Rc` to `Arc` though.

guide/src/memory.md
birkenfeld: I was looking at it from the usage point of view: `Py` references are `Send` and `Sync` like `Arc`. (And you can't use `Rc`s across threads even with a lock to synchronize access.)

adamreichold: That makes sense.

I would suggest that the document might be improved if we replaced the "It isn't magic" part by a more factual explanation that the reference count itself is protected by the GIL which there must be held when it is inspect (to access the inner object) or modified (to clone the smart pointer). It might even make sense to mention that the reference counts are stored intrusively with each Python object, i.e. they are not an optional wrapper like `Arc/Rc` but a fundamental feature of CPython's memory management (which cannot easily be swapped out for something else).

birkenfeld: And of course this goes beyond the reference counts, the GIL is just required for *any* access to a PyObject.

birkenfeld:
bors r=adamreichold

Issue #2770, opened 2022-11-22
`PyUnicodeDecodeError::new_utf8` raises `TypeError` on PyPy


samuelcolvin:
### Bug Description

```rs
PyUnicodeDecodeError::new_utf8(py, py_bytes.as_bytes(), err)?;
```
Appears to be raising/returning a `TypeError`, on PyPy, specifically

```
TypeError: expected str, got NoneType object
```

Here's my full code in case it helps:

```rs
pub fn py_bytes_to_str(py_bytes: &PyBytes) -> PyResult<&str> {
    match from_utf8(py_bytes.as_bytes()) {
        Ok(s) => Ok(s),
        Err(err) => {
            let py = py_bytes.py();
            let decode_err = PyUnicodeDecodeError::new_utf8(py, py_bytes.as_bytes(), err)?;
            Err(PyErr::from_value(decode_err))
        }
    }
}
```

### Steps to Reproduce

Run the above code on PyPy, with bytes that raise an error like `b'\x81'`.

### Your operating system and version

Ubuntu

### Your Python version (`python --version`)

Python 3.9.12 (05fbe3aa5b0845e6c37239768aa455451aa5faba, Mar 29 2022, 08:15:34) [PyPy 7.3.9 with GCC 10.2.1 20210130 (Red Hat 10.2.1-11)]

### Your Rust version (`rustc --version`)

rustc 1.67.0-nightly (c1a859b25 2022-11-10)

### Your PyO3 version

pyo3 = "0.17.3"

### How did you install python? Did you use a virtualenv?

pyenv, virtualenv yes

samuelcolvin:
Maybe `UnicodeDecodeError` expects `reason` as a last argument? But that's not passed:

https://github.com/PyO3/pyo3/blob/a3edbf4fcd595f0e234c87d4705eb600a9779130/pyo3-ffi/src/pyerrors.rs#L82

mejrs:
I think that `Py_BuildValue` format string is wrong.

PR #2772, opened 2022-11-22
fix PyUnicodeDecodeError_Create

mejrs:
Fixes https://github.com/PyO3/pyo3/issues/2770

### DIFF
--- a/newsfragments/2772.fixed.md
+++ b/newsfragments/2772.fixed.md
@@ -1,0 +1,1 @@
+Fix `PyUnicodeDecodeError_Create` failing unconditionally with a TypeError on PyPy.
--- a/pyo3-ffi/src/pyerrors.rs
+++ b/pyo3-ffi/src/pyerrors.rs
@@ -85,18 +85,17 @@ pub unsafe fn PyUnicodeDecodeError_Create(
     length: Py_ssize_t,
     start: Py_ssize_t,
     end: Py_ssize_t,
-    _reason: *const c_char,
+    reason: *const c_char,
 ) -> *mut PyObject {
     crate::PyObject_CallFunction(
         PyExc_UnicodeDecodeError,
-        std::ffi::CStr::from_bytes_with_nul(b"sy#nns\0")
-            .unwrap()
-            .as_ptr(),
+        b"sy#nns\0".as_ptr().cast::<c_char>(),
         encoding,
         object,
         length,
         start,
         end,
+        reason,
     )
 }
 


adamreichold (commented):
Do we have a test that should have failed?

samuelcolvin:
It looks for a quick inspection like both pyo3 and pydantic-core (where I found the error), test with the same three pypy versions:

```yaml
          - 'pypy3.7'
          - 'pypy3.8'
          - 'pypy3.9'
```

However it seems tests are not actually being run, either on merge with main or on pull requests, see [this build](https://github.com/PyO3/pyo3/actions/runs/3523220765). But it seems that changed yesterday, so not sure about the problem going back further.

Maybe @messense has some idea on what's going on in CI?

messense:
We switched to use bors, full test matrix run in staging branch, like this one: https://github.com/PyO3/pyo3/actions/runs/3522833722

samuelcolvin:
On that build it seems tests are not being run with pypy:

image

messense:
I think tests like `unicode_decode_error` have never been ran on PyPy because PyPy lacks the `auto-initialize` feature because you can't embed a PyPy interpreter in Rust.

samuelcolvin:
that would explain it.

messense:
> On that build it seems tests are not being run with pypy:

Yes, they are also not being run before the bors change: https://github.com/PyO3/pyo3/actions/runs/3507802969/jobs/5875825394

messense:
I guess we need to add a test in the `pytests` extension module: https://github.com/PyO3/pyo3/tree/main/pytests

adamreichold:
> I guess we need to add a test in the `pytests` extension module: https://github.com/PyO3/pyo3/tree/main/pytests

The idea might be bonkers: But Rust unit tests are basically a automatically generated `fn main` and test case discovery based on `#[test]`.

Could we somehow package ourselves up into an extension that exposes a single function to execute all the unit tests? Can we get at the main function generated by `libtest` somehow, maybe using nightly features?

I think the above is problematic as we would want to do this for all unit tests, not just this particular one. But `pytests` are less convenient to work during development and the effort to convert all unit tests into such integration tests seems quite large.

davidhewitt:
Hmm, that's an interesting idea. The amount of dark magic involved scares me a little.

I think the best solution would be for PyPy to implement `Py_InitializeEx` so we could just run tests in the same way. I actually took a look at this one weekend a good many months ago, but being unfamiliar with the PyPy codebase I confess I decided I couldn't manage it in a weekend and don't have the capacity to look at that myself any more :p

I have mostly taken the approach that if it works on CPython it should work on PyPy. We at least link all tests on PyPy (even if we can't run them) which should demonstrate we're at least hitting valid PyPy APIs. There are of course edge cases like this, so I kinda feel like it's ok for us to just add `pytests` specifically for those until we've found someone willing to help PyPy implement `Py_InitializeEx`.

davidhewitt:
Hmm, that's an interesting idea. The amount of dark magic involved scares me a little.

I think the best solution would be for PyPy to implement `Py_InitializeEx` so we could just run tests in the same way. I actually took a look at this one weekend a good many months ago, but being unfamiliar with the PyPy codebase I confess I decided I couldn't manage it in a weekend and don't have the capacity to look at that myself any more :p

I have mostly taken the approach that if it works on CPython it should work on PyPy. We at least link all tests on PyPy (even if we can't run them) which should demonstrate we're at least hitting valid PyPy APIs. There are of course edge cases like this, so I kinda feel like it's ok for us to just add `pytests` specifically for those until we've found someone willing to help PyPy implement `Py_InitializeEx`.

messense:
> I think the best solution would be for PyPy to implement `Py_InitializeEx`

The good news is that there is an active `py_initialize` branch: https://foss.heptapod.net/pypy/pypy/-/commits/branch/py_initialize/pypy/module/cpyext

birkenfeld:
> I think the above is problematic as we would want to do this for all unit tests, not just this particular one. But pytests are less convenient to work during development and the effort to convert all unit tests into such integration tests seems quite large.

Actually, IMO having tests separately (in whatever form...) would speed things up, since edits to the inline `#[test]`s force `cargo test` to recompile the whole `pyo3` crate, followed by all the non-inline `test` executables, which takes quite a bit of time.

If, as a nice side-effect, the same tests can also be compiled in a way that works on PyPy, so much the better.

davidhewitt:
@adamreichold my opinion here is let's merge this for 0.18?

Regarding the wider issue of testing on PyPy - I've offered to help out [on their branch](https://foss.heptapod.net/pypy/pypy/-/issues/3837). So in the short term I'd rather we didn't change anything in PyO3, and hopefully in the longer term testing on PyPy should "just work" the same as it does in CPython 😄

adamreichold:
> @adamreichold my opinion here is let's merge this for 0.18?

Agreed.

davidhewitt:
Rebased, added changelog entry, and will merge.

newsfragments/2772.fixed.md
```diff
@@ -0,0 +1 @@
+Fix `PyUnicodeDecodeError_Create` failing unconditionally with a TypeError on PyPy.
```
mejrs, line 1: Wouldn't this be a result of undefined behavior?

davidhewitt: Ah, yes, it's not super clear from the Python docs. It looks like it might be, so I'll amend this in a follow-up PR.

PR #2777, opened 2022-11-23
ci: remove `github.sha` from concurrency group

messense:
Otherwise duplicate workflow runs on staging/trying branches won't be cancelled automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref_name }}
   cancel-in-progress: true
 
 env:


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
```
davidhewitt, line 15: Will `github.event.pull_request.number` be defined when bors is running on those branches?

messense: No, it will just be empty. It's fine because workflows runs on bors branches are supposed to be run one at a time.

davidhewitt: I see. So if I understand correctly this makes try and r+ conflict with each other, I'm not sure we want that?

davidhewitt: Maybe we need to include the branch name instead of the sha?

messense: Multiple r+ are queued and ran one by one unless batched together under the same priority, so it doesn't matter.

davidhewitt (approved):
Thanks! I think this looks fine to me now 👍

messense:
bors r=davidhewitt

PR #2784, opened 2022-11-27
Automatically generate `__text_signature__` for all functions

davidhewitt:
This PR makes it so that PyO3 generates `__text_signature__` by default for all functions. It also introduces `#[pyo3(text_signature = false)]` to disable the built-in generation.

There are a few limitations which we can improve later:
 - All default values are currently set to `...`. I think this is ok because `.pyi` files often do the same. Maybe for numbers, strings, `None` and `True`/`False` we could render these in a future PR.
 - No support for `#[new]` yet.

Alternative design ideas:
- Only autogenerate for methods with `#[pyo3(signature = (...))]` annotation. I started with this, and then decided it made sense to do it for everything.
- Opt-out with `#[pyo3(text_signature = None)]`. This is slightly harder to parse in the macro, but matches the final result in Python better, so if this looks preferable to others, I can change from `text_signature = false` to `text_signature = None`.

There's some small tidying up / refactoring to do before this merges (happy to take suggestions on this), however the general logic, design and docs are ready for review.

### DIFF
--- a/guide/src/function/signature.md
+++ b/guide/src/function/signature.md
@@ -188,3 +188,138 @@ impl MyClass {
     }
 }
 ```
+
+## Making the function signature available to Python
+
+The function signature is exposed to Python via the `__text_signature__` attribute. PyO3 automatically generates this for every `#[pyfunction]` and all `#[pymethods]` directly from the Rust function, taking into account any override done with the `#[pyo3(signature = (...))]` option.
+
+This automatic generation has some limitations, which may be improved in the future:
+- It will not include the value of default arguments, replacing them all with `...`. (`.pyi` type stub files commonly also use `...` for all default arguments in the same way.)
+- Nothing is generated for the `#[new]` method of a `#[pyclass]`.
+
+In cases where the automatically-generated signature needs adjusting, it can [be overridden](#overriding-the-generated-signature) using the `#[pyo3(text_signature)]` option.)
+
+The example below creates a function `add` which accepts two positional-only arguments `a` and `b`, where `b` has a default value of zero.
+
+```rust
+use pyo3::prelude::*;
+
+/// This function adds two unsigned 64-bit integers.
+#[pyfunction]
+#[pyo3(signature = (a, b=0, /))]
+fn add(a: u64, b: u64) -> u64 {
+    a + b
+}
+#
+# fn main() -> PyResult<()> {
+#     Python::with_gil(|py| {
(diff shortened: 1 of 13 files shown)


adamreichold:
> Opt-out with #[pyo3(text_signature = None)]. This is slightly harder to parse in the macro, but matches the final result in Python better, so if this looks preferable to others, I can change from text_signature = false to text_signature = None.

To be honest, `None` appears be much less surprising to me compared to `false`, meaning I would maybe guess `None` but certainly not `false` without reading the documentation.

pyo3-macros-backend/src/pyfunction/signature.rs
```diff
@@ -556,4 +564,67 @@ impl<'a> FunctionSignature<'a> {
...
+    pub fn text_signature(&self, self_argument: Option<&str>) -> String {
+        let mut output = String::new();
+        output.push('(');
+
+        if let Some(arg) = self_argument {
+            output.push('$');
+            output.push_str(arg);
+        }
+
+        let mut first = self_argument.is_none();
```
adamreichold, line 577: Hiding the `first` binding would feel more robust to me, e.g.

```rust
let mut maybe_push_comma = {
    let mut first = self_argument.is_none();

    |output: &mut String| {
        if !first {
            output.push_str(", ");
        } else {
            first = false;
        }
    }
};
```

guide/src/function/signature.md
```diff
@@ -188,3 +188,136 @@ impl MyClass {
     }
 }
 ```
+
+## Making the function signature available to Python
+
+The function signature is exposed to Python via the `__text_signature__` attribute. PyO3 automatically generates this for every `#[pyfunction]` and all `#[pymethods]`.
```
adamreichold, line 194: I think this should mention that `text_signature` is generated based on the `signature` attribute optionally attached to the `#[pyfunction]`s and `#[pymethod]`s. Without that, one could expect this to be derived from the Rust function signature.

pyo3-macros-backend/src/pyfunction/signature.rs
davidhewitt: Good idea 👍

davidhewitt:
Will try to change this to use `None` for the opt-out when I get a moment 👍

guide/src/function/signature.md
davidhewitt: Thanks, done

davidhewitt:
bors retry

PR #2786, opened 2022-11-27
Add support for wrapping attributes in `#[cfg_attr(feature = "pyo3", …)]`

konstin:
This adds support for wrapping attributes in `#[cfg_attr(feature = "pyo3", ...)]`, making it possible to build rust libraries with optional python bindings. Check `optional_bindings.md` for an extensive example.

The main trouble with the PR was that some of the attribute parsing is done with `syn::parse::Parse` impls, while other parts are done with `syn::Meta` matching, while syn itself just exposes attribute contents as token stream. Another difficulty is that while `#[foo]` is a single attribute, in `#[cfg_attr(feature = "something", foo, bar)]` you can have completely unrelated foo and bar attributes. 

For testing, i had to add a `pyo3` feature to pyo3 itself. I couldn't find a better solution to get a `pyo3` active in the test, but OTOH that feature shouldn't cause any problems on the user side


birkenfeld:
Does this fix #780 ? (It was closed by the submitter, not fixed.)

konstin:
yep!

guide/src/class/optional_bindings.md
```diff
@@ -0,0 +1,62 @@
+# Optional bindings
+
+You might want to write a library the is usable both in pure rust and as a python library. For that, pyo3 support wrapping attributes in `#[cfg_attr(feature = "pyo3", ...)]` (the feature unfortunately has to be hardcoded, so the feature must be named `pyo3`). This does not only apply to classes and their methods but also to e.g. `#[pyfunction]`.
```
birkenfeld, line 3:
```suggestion
You might want to write a library the is usable both in pure Rust and as a Python library. For that, PyO3 supports wrapping attributes in `#[cfg_attr(feature = "pyo3", ...)]` (the feature name unfortunately has to be hardcoded, so the feature must be named `pyo3`). This does not only apply to classes and their methods but also to e.g. `#[pyfunction]`.
```

guide/src/class/optional_bindings.md
```diff
@@ -0,0 +1,62 @@
...
+
+Make pyo3 optional in Cargo.toml:
+
+```toml
+[dependencies]
+pyo3 = { version = "0.14", features = ["extension-module", "abi3"], optional = true }
+```
```
birkenfeld, line 10: What about `[lib] crate-type`? Must then be both `cdylib` and `rlib`, I guess?

birkenfeld (commented):
Looking at the new test file, that is a lot of `cfg_attr`...

Wild idea: how about a macro that can be applied to a whole module/scope and *removes* the pyo3 related attributes? It can then be added with `cfg_attr(not(feature = "pyo3"), ...)`

guide/src/class/optional_bindings.md
```diff
@@ -0,0 +1,62 @@
...
+            } else {
+                b -= a
+            }
+        }
+
+        Self::new(a)
+    }
+}
+
+#[cfg(features = "pyo3")] // We don't want that function at all in a rust library
```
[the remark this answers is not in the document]
konstin: oops thanks

pyo3-macros-backend/src/attributes.rs
```diff
@@ -134,28 +131,156 @@ pub fn is_attribute_ident(attr: &syn::Attribute, name: &str) -> bool {
...
 }
 
-pub fn take_pyo3_options<T: Parse>(attrs: &mut Vec<syn::Attribute>) -> Result<Vec<T>> {
+pub fn take_pyo3_options<T: Parse>(attrs: &mut Vec<Attribute>) -> Result<Vec<T>> {
     let mut out = Vec::new();
-    take_attributes(attrs, |attr| {
-        if let Some(options) = get_pyo3_options(attr)? {
+    let mut new_attrs = Vec::new();
+
+    for mut attr in attrs.drain(..) {
```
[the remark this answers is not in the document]
konstin: I find it easier to read when we have a distinct old-attr-drain and new attr building; iterating and editing/deleting at the same time is unfortunately difficult to write straightforward

Cargo.toml
```diff
@@ -107,8 +107,12 @@ full = [
     "indexmap",
     "eyre",
     "anyhow",
+    "pyo3",
 ]
 
+# This makes `#[cfg_attr(feature = "pyo3", pyclass)]` in our own tests work, it has no other function
+pyo3 = []
```
davidhewitt, line 114: I wonder... should we move this to the `pytests` crate or something similar, so that we don't need to expose a no-op feature to the users?

To avoid other code changes, you _might_ also just be able to hack it by adding `RUSTFLAGS=--cfg "feature=pyo3"`, which is all `cargo` really does under the hood I think?

birkenfeld:
> > Wild idea: how about a macro that can be applied to a whole module/scope and _removes_ the pyo3 related attributes? It can then be added with `cfg_attr(not(feature = "pyo3"), ...)`
> 
> I've thought about that too: We could have a `stub` feature in pyo3 that does nothing but removing the attributes and depends if on anything at all on syn, which would be really cool. This would however mean that by default all dependencies are optional and that build script must not set any linker options while still supporting the existing feature combinations, which is complex. I think it would also make sense to migrate all attributes to the `#[pyo3(...)]` form before so we can have "self-removing attributes"

It would have to be a separate crate `pyo3-stub` otherwise, but then imports get more complex.

pyo3-macros-backend/src/attributes.rs
```diff
@@ -134,28 +131,156 @@ pub fn is_attribute_ident(attr: &syn::Attribute, name: &str) -> bool {
...
+                    return Ok(true);
+                }
+            }
+            Ok(false)
+        };
+        if let Ok(mut meta) = attr.parse_meta() {
+            if handle_cfg_feature_pyo3(&mut attr, &mut meta, parse_attr)? {
+                continue;
+            }
+        }
```
davidhewitt, line 167: I think it would be good to check `if attr.path.is_ident("cfg_attr")` before calling `parse_meta`, to avoid a lot of unnecessary parsing work.

davidhewitt (commented):
This is very cool, thanks for implementing!

If I understand correctly, this works by assuming that the top-level macro was also wrapped in `#[cfg_attr(feature = "pyo3", pyclass)]` (for example).

There is some interesting edge cases from this, e.g. I think the following code will work even in crates which _don't_ define a "pyo3" feature:

```rust
#[pyclass]
struct Foo;

#[pymethods]
impl Foo {
    #[cfg_attr(feature = "pyo3", new)]
    fn new() -> Self { Self }
}
```

... because the current assumption seems to be that if the macro is running, the top-level macro has already passed the `cfg_attr` gate.

davidhewitt:
> Wild idea: how about a macro that can be applied to a whole module/scope and _removes_ the pyo3 related attributes? It can then be added with `cfg_attr(not(feature = "pyo3"), ...)`

Also agreed this is a very interesting idea. Note that inner macro attributes are unstable so it would have to go on each pyclass (or a thing containing pyclasses, maybe a whole module after #2367).

Assuming that typically users would set things up to have no pyo3 by default, and then expose a feature which adds a pyo3 dependency, I'm trying to think through the practical usage.

> It would have to be a separate crate `pyo3-stub` otherwise, but then imports get more complex.

I don't think it's so bad; if a PyO3 dependency is disabled then imports will have a `#[cfg]` anyway for any imported PyO3 items.

With a separate crate, users can probably just have something like

```
#[cfg_attr(any(test, not(feature = "py")), pyo3_stub::pyo3_stub)]
#[pyclass]
struct Foo;
```

Or with a stub feature in PyO3, I guess we'd potentially need to make the base PyO3 without features contain absolutely nothing and then enable e.g. `c-api`, `macros-stub`, `macros` as users want? So this would be breaking.

So I'm thinking maybe the separate crate is the way to go? It could still depend on `pyo3-macros-backend` but without any `pyo3-build-config`. I'd be quite interested to see how that compares to the approach in this PR!

pyo3-macros-backend/src/attributes.rs
konstin: Is parsing `Meta` expensive? This is the only location where we do it in this order, in the other two locations we also need `meta` for the normal parsing

Cargo.toml
konstin: so like moving the ui test to `pytests/src`? The `RUSTFLAGS` would also do, but imho it'd be nice if `cargo test` would continue to work.

I've also been thinking about using a different feature name for the tests, but i couldn't come up with a good design for that either, especially for cases such as the guide doc test

konstin:
> There is some interesting edge cases from this, e.g. I think the following code will work even in crates which _don't_ define a "pyo3" feature:
> 
> ```rust
> #[pyclass]
> struct Foo;
> 
> #[pymethods]
> impl Foo {
>     #[cfg_attr(feature = "pyo3", new)]
>     fn new() -> Self { Self }
> }
> ```
> 
> ... because the current assumption seems to be that if the macro is running, the top-level macro has already passed the `cfg_attr` gate.

yeah this would work, but tbh i don't think anybody will run into problems with that, `feature = "pyo3"` is too oddly specific

pyo3-macros-backend/src/attributes.rs
davidhewitt: So I've now written a benchmark in #2794 

It's based against main, what I see on my machine is that the cost per attribute (for a very basic attribute) goes from about 250ns to 1us. Most of that overhead is removed again by just adding this check.

So sure, while it's not much, maybe in a bigger codebase with lots of attributes this could eventually add up to a second or so. I'd like us to try to keep macro performance as good as possible, especially when it's so easy to add the check here.

(Generally I've been heading in the direction of avoiding `meta` parsing and writing more specific parse implementations like the one here, for best efficiency.)

davidhewitt:
On reflection I'm leaning towards merging this (preferably with slight perf tweak as noted above).

Although we agree it introduces a weird edge case, you could argue `#[cfg_attr]` not working _at all_ is also a weird edge case and definitely more annoying for users!

pyo3-macros-backend/src/attributes.rs
konstin: sweet, i didn't expect this much rigor! i've now gated both parse with `is_ident`

konstin:
> Although we agree it introduces a weird edge case, you could argue `#[cfg_attr]` not working _at all_ is also a weird edge case and definitely more annoying for users!

yeah this is really more a hack to get this case to work at all. For a real solution we'd need either cooperation from rustc to tell us which ctr_attr epressions would evaluate to true (or just being able to ask it to preprocess the attributes) or some help from cargo to stub out proc macros, as in "don't even compile the proc macro crate".

pyo3-macros-backend/src/attributes.rs
konstin: fwiw, i think moving everything into `pyo3(...)` would be help because then we can skip the non-pyo3 macro parsing completely for attributes that are not ours

konstin:
BTW this is the reason i wanted to have this feature: https://github.com/konstin/pep440-rs

mejrs:
> Although we agree it introduces a weird edge case, you could argue `#[cfg_attr]` not working _at all_ is also a weird edge case and definitely more annoying for users!

I'm reluctant (not necessarily opposed) to merge this. It's a rust edge case; not a pyo3 edge case, so I'd much rather point people to cfg_eval instead.

davidhewitt:
@mejrs I know what you mean, initially I felt similar, I've warmed to it a bit when looking at it through the lens of which edge case is least bad for the user.

Last I saw `#[cfg_eval]` it was proposed for stabilization but then postponed because it's unclear what extra features the language wants to give to proc macros here.

I wonder, to give us some control (and way to deprecate this when the language can do better), if we should gate this behavior behind an option? E.g. `#[pyclass(feature = "pyo3")]` or `#[pyclass(cfg = (feature = "pyo3"))] `? @mejrs do you think this would be beneficial?

davidhewitt:
> I wonder, to give us some control (and way to deprecate this when the language can do better), if we should gate this behavior behind an option? E.g. `#[pyclass(feature = "pyo3")]` or `#[pyclass(cfg = (feature = "pyo3"))]` ? @mejrs do you think this would be beneficial?

Anyone got any further thoughts on this? I can help push this over the line for 0.18, though I'd like some input on what configuration form folks like. Personally, I think having `#[pyclass(feature = "pyo3")]` as the way to enable this is best, as it's reasonably simple and gives the users the choice of the feature name.

joshparallel:
> > I wonder, to give us some control (and way to deprecate this when the language can do better), if we should gate this behavior behind an option? E.g. `#[pyclass(feature = "pyo3")]` or `#[pyclass(cfg = (feature = "pyo3"))]` ? @mejrs do you think this would be beneficial?
> 
> Anyone got any further thoughts on this? I can help push this over the line for 0.18, though I'd like some input on what configuration form folks like. Personally, I think having `#[pyclass(feature = "pyo3")]` as the way to enable this is best, as it's reasonably simple and gives the users the choice of the feature name.

Does this mean I as a user would have to use the above syntax every time I invoke `pyclass`? Is there a way to provide control and a deprecation pathway without exposing the extra verbosity to the user at the call site? E.g. if I could just specify the feature in one place in my workspace `Cargo.toml` that would be really easy for me. Also, would a more self-explanatory feature name be desirable, e.g. "feature_guards"? Sorry if I'm misunderstanding anything here...

mejrs:
> I wonder, to give us some control (and way to deprecate this when the language can do better), if we should gate this behavior behind an option? E.g. `#[pyclass(feature = "pyo3")]` or `#[pyclass(cfg = (feature = "pyo3"))] `? @mejrs do you think this would be beneficial?

This assumes people will read the documentation, which I assume they won't. If we do this then few users will discover it. So if we do this, then I'd want it to "just work".

konstin:
> This assumes people will read the documentation, which I assume they won't. If we do this then few users will discover it. So if we introduce this feature I'd want it to "just work".

Sorry i'm not following, but which is the solution you'd want?

davidhewitt:
> If we do this then few users will discover it. So if we introduce this feature I'd want it to "just work".

> Sorry i'm not following, but which is the solution you'd want?

I read @mejrs' comment as "don't add either option, and keep this as-is".

> Is there a way to provide control and a deprecation pathway without exposing the extra verbosity to the user at the call site?

I suppose it may be the case that if the language eventually supports a better solution we might be able to transparently just use it without any deprecation.

... in which case, should we just merge this as-is? (After rebase and maybe a squash)

mejrs:
> Sorry i'm not following, but which is the solution you'd want?

that either:
- we merge this PR as is, no gating behind an option.
- we don't merge it

I am in the same boat as you: I also conditionally use pyo3. Which is why I wrote https://github.com/PyO3/pyo3/pull/2796 and https://github.com/PyO3/pyo3/pull/2692 but these are done in a way that is unsurprising and composes well. I'm not sure that's true for this PR; it feels rather hacky.

I personally would prefer to just tell users to write...

```rust
#[cfg(feature = "pyo3")]
#[pymethods]
impl Foo{
    // ...
}
```
...which may result in some extra boilerplate, but that's not a disaster or anything.

That said, if you are convinced that there are no gotchas or nasty corner cases with it, I'm not opposed to merging this 👍

davidhewitt:
I don't _think_ there are nasty gotchas, though I agree with you it is hacky and the sort of thing we may regret later as a result.

I guess worst case if we merge this, we get reports of issues which mean we need to gate it behind the option later on?

davidhewitt:
The suggestion to simply recommend a separate pymethods block with a top-level `#[cfg]` also seems very reasonable to me, I wonder if this would be suitable for you @konstin ?

ChristopherRabotin:
Hi there,

I was wondering whether there was any chance this would merge in the coming weeks. I would use it. My use-case is that I want to allow Python users to set some of the fields of a struct (which are Python-compatible) but not touch the other ones.

Thanks

davidhewitt:
Sorry, I somehow missed the last reply on this thread from @konstin .

So the understanding we have is that `#[pymethods]` has a workaround but `#[pyclass]` currently does not.

The other day I had an idea to add [a `getters` option to `#[pyclass]`](https://github.com/PyO3/pyo3/discussions/2957#discussioncomment-5017023) (name to be bikeshed) with the context being that it could support more complex lookups. I was a bit unsure of its power, though I see here it would also solve the problem.

E.g. the code we have today 

```rust
#[pyclass]
pub struct Foo {
    #[pyo3(get, set, name = "bar")]
    pub baz: usize,
}
```

could be rewritten using this hypothetical `#[getters]` as

```rust
#[pyclass(
    getters = (bar = self.baz),
    setters = (bar = self.baz)
)]
pub struct Foo {
    pub baz: usize,
}
```

which would scale naturally to `#[cfg_attr]` without hacks:

```rust
#[cfg_attr(
    feature = "pyo3",
    pyclass(
        getters = (bar = self.baz),
        setters = (bar = self.baz)
    )
)]
pub struct Foo {
    pub baz: usize,
}
```

Perhaps we should explore that idea further?

ChristopherRabotin:
I have no preference ;-)

konstin:
we could definitely move all field attributes to into the container attribute, but this will most likely be more cumbersome both in pyo3 itself and on the user side. I still think the current solution is the best compromise we can get as a workaround for the lack of `#[cfg_eval]`.

ChristopherRabotin:
@davidhewitt sorry to bump this up again, but I'm about to embark on a new PyO3 development that will also be behind a feature flag. What is the best approach to solve what this PR solves with version 0.20 (or higher)? Thanks

davidhewitt:
I think https://github.com/PyO3/pyo3/pull/2786#issuecomment-1369214439 is still most applicable.

xnerhu:
what's the status?

konstin:
There does not seem to be progress on [`cfg_eval`](https://github.com/rust-lang/rust/issues/82679), but this PR has too many conflicts to still being merged.

PR #2788, opened 2022-11-29
Define _PyErr_ChainExceptions() FFI for CPython

ijl:
This appears to be the only way to chain exceptions via the FFI. It was introduced prior to CPython 3.7. It has some description of usage in PEP 490. There was a discussion of whether it should be made stable in BPO 44938.

### DIFF
--- a/newsfragments/2788.added.md
+++ b/newsfragments/2788.added.md
@@ -1,0 +1,1 @@
+Define `_PyErr_ChainExceptions()` FFI for CPython.
--- a/pyo3-ffi/src/cpython/pyerrors.rs
+++ b/pyo3-ffi/src/cpython/pyerrors.rs
@@ -150,6 +150,11 @@ pub struct PyStopIterationObject {
     pub value: *mut PyObject,
 }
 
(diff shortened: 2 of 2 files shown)


messense (approved):
Defined in https://github.com/python/cpython/blob/e0f91deb5930ecb02e7f8ced9bd82609e6889fb0/Include/cpython/pyerrors.h#L101

bors r+

messense:
It's not presented in PyPy C-API, might need a `#[cfg(not(PyPy)]`

bors r-

davidhewitt:
As another comment, it looks like the implementation basically just either calls `PyErr_Restore` if no exception is set, or otherwise adds the exception to the currently-set one with `PyException_SetContext`.

Some possible follow-ups (maybe we can track in new issues):
- We could add `context()` and `set_context()` to match `cause` / `set_cause` which we already bind.
- We could add a high-level implementation of this `PyErr::chain()` which could be used in the abi3 / PyPy cases.

davidhewitt (approved):

Issue #2711, opened 2022-10-28
add memory view wrapper


davidhewitt:
I noticed that we don't have a `PyMemoryView` wrapper, however we have C-API symbols for it. Someone interested would be welcome to add one. I think the implementation for `PyBytes` and `PyByteArray` would be good starting points to draw from.

ikrivosheev:
@davidhewitt hello! I will work on this PR.

davidhewitt:
Thanks! Please let me know if you need any pointers. It looks like `pyo3-ffi/src/memoryobject.rs` contains enough symbols for us to have a `PyMemoryView` type, which might be all that's possible with abi3.

ikrivosheev:
@davidhewitt hello! I created draft PR (https://github.com/PyO3/pyo3/pull/2792)  but I have some questions:

1. Can I edit `pyo3-ffi/src/memoryobject.rs`? Or it's auto generated code.
2. Which methods do we want to add to PyMemoryView?

davidhewitt:
> Can I edit pyo3-ffi/src/memoryobject.rs

Yes - it's handwritten based off current CPython headers.

> Which methods do we want to add to PyMemoryView?

 I think what you've written in #2792 is great, we possibly don't need more? _Maybe_ `PyMemoryView::get_base` ?

davidhewitt:
Done in #3514

PR #2792, opened 2022-11-30
issues-2711 Add PyMemoryView

ikrivosheev:
Closed: https://github.com/PyO3/pyo3/issues/2711.

### DIFF
--- a/src/types/memoryview.rs
+++ b/src/types/memoryview.rs
@@ -1,5 +1,7 @@
+use crate::buffer::{Element, PyBuffer};
 use crate::err::PyResult;
-use crate::{ffi, AsPyPointer, PyAny};
+use crate::types::PySequence;
+use crate::{ffi, AsPyPointer, PyAny, Python};
 
 /// Represents a Python `memoryview`.
 #[repr(transparent)]
@@ -15,6 +17,58 @@ impl PyMemoryView {
             src.py()
                 .from_owned_ptr_or_err(ffi::PyMemoryView_FromObject(src.as_ptr()))
         }
+    }
+
+    /// Creates a new Python memoryview from a raw pointer and length.
+    ///
+    /// Panics if out of memory.
+    ///
+    /// # Safety
+    ///
+    /// This function dereferences the raw pointer `ptr` as the
+    /// leading pointer of a slice of length `len`. [As with
+    /// `std::slice::from_raw_parts`, this is
+    /// unsafe](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety).
+    pub unsafe fn from_ptr(
+        py: Python<'_>,
+        ptr: *const u8,
+        len: usize,
+        is_writable: bool,
+    ) -> &PyMemoryView {
+        let flags = if is_writable {
+            ffi::PyBUF_WRITE
+        } else {
+            ffi::PyBUF_READ
+        };
+
+        py.from_owned_ptr(ffi::PyMemoryView_FromMemory(
+            ptr as *mut _,
+            len as ffi::Py_ssize_t,
+            flags,
+        ))
+    }
+
+    /// Returns the length of the memoryview.
+    pub fn len(&self) -> usize {
+        let size = unsafe { ffi::Py_SIZE(self.as_ptr()) };
+        // non-negative Py_ssize_t should always fit into Rust usize
+        size as usize
+    }
+
+    /// Checks if the memoryview is empty.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns `self` cast as a `PySequence`.
+    pub fn as_sequence(&self) -> &PySequence {
+        unsafe { self.downcast_unchecked() }
+    }
+
+    /// Gets the Python string as a byte slice.
+    #[inline]
+    pub fn as_buffer<T: Element>(&self) -> PyResult<PyBuffer<T>> {
+        PyBuffer::get(self)
     }
 }
 


src/types/memoryview.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::types::PySequence;
+
+/// Represents a Python `memoryview`.
+#[repr(transparent)]
+pub struct PyMemoryView<T>(PyAny, PhantomData<T>);
+
+pyobject_native_type_core!(PyMemoryView<T>, ffi::PyMemoryView_Type, #checkfunction=ffi::PyMemoryView_Check; T);
+
+impl PyMemoryView<u8> {
+    pub fn new<'p>(py: Python<'p>, memory: &mut [u8], is_writable: bool) -> &'p PyMemoryView<u8> {
```
birkenfeld, line 16: Unfortunately this API isn't safe - the memory would have to be borrowed for as long as the returned memoryview lives, but that can't be encoded since Python objects can't have associated lifetimes (except for the `'py` pseudo-lifetime).

It might be possible to offer it as `unsafe fn` with appropriate documentation.

src/types/memoryview.rs
```diff
@@ -0,0 +1,123 @@
...
+use std::os::raw::c_char;
+
+use crate::{AsPyPointer, PyAny, PyResult, Python};
+use crate::buffer::{Element, PyBuffer};
+use crate::ffi::{self, Py_ssize_t};
+use crate::types::PySequence;
+
+/// Represents a Python `memoryview`.
+#[repr(transparent)]
+pub struct PyMemoryView<T>(PyAny, PhantomData<T>);
```
davidhewitt, line 11: The choice to make this generic is interesting, and I've been thinking about this further. At the moment none of PyO3's types (other than buffer) are generic. Also `PyMemoryView_Check` won't check the generic type is correct, which makes it easy for this type to hold memory views with the wrong T.

For these reasons I think it's better if we remove the generic type from this and just have `PyMemoryView`.

src/types/memoryview.rs
davidhewitt: Agreed this isn't safe as-is. We technically already have the `unsafe fn` in `from_ptr`. It would also be _very_ hard to use correctly, as if you pass this object to other Python code it's extremely likely this memoryview may live longer than you expect.

A safe alternative may be to accept `&'static [u8]`, although that impairs usability a bit.

Maybe the `PyMemoryView::from` already started below is the thing we should encourage users to use (they can then wrap their buffers in `#[pyclass]` types which expose buffer access).

src/types/memoryview.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Creates a new Python memoryview from a raw pointer and length.
+    ///
+    /// Panics if out of memory.
+    ///
+    /// # Safety
+    ///
+    /// This function dereferences the raw pointer `ptr` as the
+    /// leading pointer of a slice of length `len`. [As with
+    /// `std::slice::from_raw_parts`, this is
+    /// unsafe](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety).
```
davidhewitt, line 38: Also needs to have a safety note requiring that the created memory view does not outlive the data in `ptr`. This is _extremely_ hard to use safely!

src/types/memoryview.rs
```diff
@@ -0,0 +1,123 @@
...
+            flags
+        ))
+    }
+}
+
+impl<T: Element> PyMemoryView<T> {
+
+    /// Creates a new Python `memoryview` object from another Python object that
+    /// implements the buffer protocol.
+    pub fn from<'p, I>(py: Python<'p>, src: &'p I) -> PyResult<&'p PyMemoryView<T>>
```
davidhewitt, line 59: I might be tempted to just have `pub fn from(&PyAny) -> PyResult<&PyMemoryView>` to simplify the signature a bit.

src/types/memoryview.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// Returns the length of the memoryview.
+    pub fn len(&self) -> usize {
+        unsafe {
+            let size = ffi::Py_SIZE(self.as_ptr());
+            // non-negative Py_ssize_t should always fit into Rust usize
+            size as usize
+        }
+    }
```
davidhewitt, line 73: I think this always returns length in `u8`, not units of `T`? So again this looks ok, but only with untyped `PyMemoryView`.

src/types/memoryview.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Returns `self` cast as a `PySequence`.
+    pub fn as_sequence(&self) -> &PySequence {
+        unsafe { self.downcast_unchecked() }
+    }
+
+    /// Gets the Python string as a byte slice.
+    #[inline]
+    pub fn as_buffer(&self) -> PyResult<PyBuffer<T>> {
```
davidhewitt, line 87: ... and with untyped `PyMemoryView`, this can still be generic:

```suggestion
    pub fn as_buffer<T>(&self) -> PyResult<PyBuffer<T>> {
```

src/types/memoryview.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Creates a new Python `memoryview` object from another Python object that
+    /// implements the buffer protocol.
+    pub fn from<'p, I>(py: Python<'p>, src: &'p I) -> PyResult<&'p PyMemoryView<T>>
+    where
+        I: AsPyPointer,
+    {
+        unsafe { py.from_owned_ptr_or_err(ffi::PyMemoryView_FromObject(src.as_ptr())) }
+    }
+
```
davidhewitt, line 65: There is possibly also scope to have `PyMemoryView::from_buffer` which takes `PyBuffer` as input.

davidhewitt (changes requested):
Hello! Sorry for the _long_ delay in reviewing this, I was a bit overloaded before the Christmas holidays.

Thank you for working on this. This is a great start, however we need to adjust some of the APIs a bit before this can be ready for merge. See other comments 🙂

src/types/memoryview.rs
ikrivosheev: Done

src/types/memoryview.rs
ikrivosheev: Done

src/types/memoryview.rs
ikrivosheev: Done

src/types/memoryview.rs
ikrivosheev: Removed

src/types/memoryview.rs
ikrivosheev: @davidhewitt hmm, I have a question: how I can `as_ptr()` for PyBuffer?

ikrivosheev:
@birkenfeld @davidhewitt hello! Sorry for the delay... I extended PyMemoryView and added some new methods.

src/types/memoryview.rs
```diff
@@ -16,6 +18,58 @@ impl PyMemoryView {
...
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns `self` cast as a `PySequence`.
+    pub fn as_sequence(&self) -> &PySequence {
+        unsafe { self.downcast_unchecked() }
+    }
+
+    /// Gets the Python string as a byte slice.
```
davidhewitt, line 68: Comment here looks out of date.

src/types/memoryview.rs
```diff
@@ -16,6 +18,58 @@ impl PyMemoryView {
...
+    #[inline]
+    pub fn as_buffer<T: Element>(&self) -> PyResult<PyBuffer<T>> {
```
davidhewitt, line 70: I'm trying to think, is there a memory safety concern casting from bytes to arbitrary elements `T` here?   Potentially as long as the buffer format is checked, it's fine? We require `Element: Copy` which implies these types are all POD and don't own heap allocations to trigger use-after-free etc

adamreichold: If there is, it was already be present in `PyBuffer::get`, wasn't it? And that one does indeed check the buffer format using `Element::is_compatible_format`.

What I would like to see is a unit test to check whether this actually works for anything but `PyBuffer<u8>`.

PR #2794, opened 2022-12-01
pyo3-macros-backend: benchmark some parsing

davidhewitt:
To support #2786 

I've been meaning to try out benchmarking the macros for a while, looks like this approach of building up fake token trees and then interacting with them can work.


davidhewitt:
For now I'm going to close this, it was certainly an interesting exercise and I'd like to return to benchmark / optimise our macros further in the future.

PR #2795, opened 2022-12-04
accept any iterator for `PySet::new` and `PyFrozenSet::new`

### LINKED ISSUE #2789: Change `PySet::new` to accept a `ExactSizeIterator`
Somewhat related to #2673.

`PyList::new` has signature:

```rust
    pub fn new<T, U>(py: Python<'_>, elements: impl IntoIterator<Item = T, IntoIter = U>) -> &PyList
    where
        T: ToPyObject,
        U: ExactSizeIterator<Item = T>,
    {
```

But `PySet::new` is defined as

```rs
    pub fn new<'p, T: ToPyObject>(py: Python<'p>, elements: &[T]) -> PyResult<&'p PySet> {
        let list = elements.to_object(py);
        unsafe { py.from_owned_ptr_or_err(ffi::PySet_New(list.as_ptr())) }
    }
```

`PySet::new` `elements` parameter type should be relaxed to `ExactSizeIterator<Item = T>`, not a slice.

davidhewitt:
Closes #2789 

Looks like it's about 20% faster in the case I benchmarked to use `PySet_Add` directly without the intermediate list.

### DIFF
--- a/src/types/frozenset.rs
+++ b/src/types/frozenset.rs
@@ -1,9 +1,12 @@
 // Copyright (c) 2017-present PyO3 Project and Contributors
 //
 
-use crate::err::{PyErr, PyResult};
 #[cfg(Py_LIMITED_API)]
 use crate::types::PyIterator;
+use crate::{
+    err::{self, PyErr, PyResult},
+    IntoPyPointer, Py, PyObject,
+};
 use crate::{ffi, AsPyPointer, PyAny, Python, ToPyObject};
 
 use std::ptr;
@@ -31,9 +34,12 @@ impl PyFrozenSet {
     /// Creates a new frozenset.
     ///
     /// May panic when running out of memory.
-    pub fn new<'p, T: ToPyObject>(py: Python<'p>, elements: &[T]) -> PyResult<&'p PyFrozenSet> {
-        let list = elements.to_object(py);
-        unsafe { py.from_owned_ptr_or_err(ffi::PyFrozenSet_New(list.as_ptr())) }
+    #[inline]
+    pub fn new<'a, 'p, T: ToPyObject + 'a>(
+        py: Python<'p>,
+        elements: impl IntoIterator<Item = &'a T>,
+    ) -> PyResult<&'p PyFrozenSet> {
+        new_from_iter(py, elements).map(|set| set.into_ref(py))
     }
 
     /// Creates a new empty frozen set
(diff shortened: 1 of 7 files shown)


src/types/frozenset.rs
```diff
@@ -157,6 +162,35 @@ mod impl_ {
 
 pub use impl_::*;
 
+#[inline]
+pub(crate) fn new_from_iter<T: ToPyObject>(
+    py: Python<'_>,
+    elements: impl IntoIterator<Item = T>,
+) -> PyResult<Py<PyFrozenSet>> {
+    #[inline]
+    fn new_from_iter_inner(
```
birkenfeld, line 171: Why is the inner function necessary?

src/types/frozenset.rs
```diff
@@ -31,9 +34,11 @@ impl PyFrozenSet {
     /// Creates a new frozenset.
     ///
     /// May panic when running out of memory.
-    pub fn new<'p, T: ToPyObject>(py: Python<'p>, elements: &[T]) -> PyResult<&'p PyFrozenSet> {
-        let list = elements.to_object(py);
-        unsafe { py.from_owned_ptr_or_err(ffi::PyFrozenSet_New(list.as_ptr())) }
+    pub fn new<'p, T: ToPyObject>(
```
birkenfeld, line 37: Just for the record, this is breaking since `&[T]` is `IntoIterator<Item = &T>`.

birkenfeld (approved):

src/types/frozenset.rs
davidhewitt: To avoid generic code bloat, this inner part is not generic - actually will quickly just force-push to remove `#[inline]` from this as it's a bulkier loop and unlikely to benefit much from inlining.

src/types/frozenset.rs
davidhewitt: Hmm. I think in which case let's use `&T` here for now too. We can revisit that as a different breaking change in future.

src/types/frozenset.rs
birkenfeld: Got it, I always forget about that aspect.

src/types/frozenset.rs
mejrs: If we're going to be breaking code anyway, what about taking `IntoPy` instead? i.e. inserting pyclasses into the set by value?

davidhewitt: Yes, I think I'd like to do this (and I've thought similar in the past for `PyList::new`). I will hold off from doing that here, as this is again non-breaking after I've corrected to use `Item = &T` for now.

I'd be keen to revisit this along with #2316 and also a design about how we maybe also handle fallible into-py conversions.

davidhewitt:
bors r=birkenfeld

PR #2796, opened 2022-12-06
Forward cfgs on pyclass fields to the method defs

mejrs:
With this and the cfg_attr PR, I don't need cfg_eval at all anymore :)

- [x] needs some more tests


pyo3-macros-backend/src/pymethod.rs
```diff
@@ -653,7 +662,15 @@ pub fn impl_py_getter_def(
         }
     };
 
+    let mut cfg_attrs = TokenStream::new();
+    if let PropertyType::Descriptor { field, .. } = &property_type {
+        for attr in field.attrs.iter().filter(|attr| attr.path.is_ident("cfg")) {
+            attr.to_tokens(&mut cfg_attrs);
+        }
+    }
```
davidhewitt, line 670: I suspect it probably works to use an `Option` instead of having the intermediate `TokenStream` buffer:

```suggestion
    let mut cfg_attrs = None;
    if let PropertyType::Descriptor { field, .. } = &property_type {
        cfg_attrs = Some(field.attrs.iter().filter(|attr| attr.path.is_ident("cfg")));
    }
```

davidhewitt (commented):
Seems like an uncontroversial good idea to me! We already forward cfgs in other places.

Needs a CHANGELOG entry - `changed`, perhaps?

As for tests, I'm not sure there's much else which could be added? The hygiene test will already fail if the `cfg` are not forwarded correctly. I suppose a unit test which checks the generated proc-macro output would be most precise test, though we don't do that at all yet :)

newsfragments/2796.md
```diff
@@ -0,0 +1 @@
+Mixing cfgs and pyo3 attributes on struct fields will now work
```
davidhewitt, line 1: Needs to be 2796.changed.md (category is important so that towncrier knows where this goes on the changelog).

davidhewitt (approved):
LGTM 👍

pyo3-macros-backend/src/pymethod.rs
mejrs: I don't think I can `collect` into a tokenstream :(

PR #2800, opened 2022-12-08
allow `**kwargs` to take arguments which conflict with positional-only parameters

### LINKED ISSUE #2799: Positional-only parameters do not follow PEP 570
### Bug Description

Amongst several benefits to positional-only arguments, [PEP 570](https://peps.python.org/pep-0570) demonstrates two which are not currently possible in PyO3:

- the author may rename a positional-only parameter without it being a breaking change
- it allows functions to accept any keyword argument but also accept a positional one

### Steps to Reproduce

Let us first look at a pure Python function:

```pycon
>>> def makedict(other=(), /, **kwargs):
...     return dict(other, **kwargs)
...
>>> makedict(other=1)
{'other': 1}
```

This is how the `dict` constructor works, for example. There is no possible naming of the positional-only argument that is visible (as an effect) to the caller.

When we try to implement this in Rust, like so:

```rust
#[pyfunction(other = "None", "/", kwargs = "**")]
fn makedict(py: Python<'_>, other: Option<&PyAny>, kwargs: Option<&PyDict>) -> PyResult<PyObject> {
    let dict = py.import("builtins")?.getattr("dict")?;
    Ok(dict.call((other.unwrap_or_else(|| PyTuple::empty(py)),), kwargs)?.into())
}
```

then we get an error

```pycon
>>> makedict(other=1)
TypeError: makedict() got some positional-only arguments passed as keyword arguments: 'other'
```

### Backtrace

_No response_

macOS Monterey

3.10.7

rustc 1.64.0 (a55dd71d5 2022-09-19)

### Your PyO3 version

0.17.2

pyenv

### Additional Info

_No response_

davidhewitt:
Closes #2799 

I did a little bit of refactoring at the same time, sorry! The bit which changes behaviour is that in the now-factored-out `handle_kwargs` method, if the keyword argument name is found in the positional-only range then the varkeywords handler is now given a chance to accept it. (Lines 387-390 in new `extract_argument.rs`.)

### DIFF
--- a/src/impl_/extract_argument.rs
+++ b/src/impl_/extract_argument.rs
@@ -237,17 +237,9 @@ impl FunctionDescription {
         V: VarargsHandler<'py>,
         K: VarkeywordsHandler<'py>,
     {
-        // Safety: Option<&PyAny> has the same memory layout as `*mut ffi::PyObject`
-        let args = args as *const Option<&PyAny>;
-        let positional_args_provided = nargs as usize;
-        let args_slice = if args.is_null() {
-            &[]
-        } else {
-            std::slice::from_raw_parts(args, positional_args_provided)
-        };
-
         let num_positional_parameters = self.positional_parameter_names.len();
 
+        debug_assert!(nargs >= 0);
         debug_assert!(self.positional_only_parameters <= num_positional_parameters);
         debug_assert!(self.required_positional_parameters <= num_positional_parameters);
         debug_assert_eq!(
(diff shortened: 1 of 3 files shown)


src/impl_/extract_argument.rs
```diff
@@ -360,50 +334,77 @@ impl FunctionDescription {
...
+            self.positional_parameter_names.len()
+        );
+        debug_assert_eq!(
+            output.len(),
+            num_positional_parameters + self.keyword_only_parameters.len()
+        );
+        let mut positional_only_keyword_arguments = Vec::new();
+        for (kwarg_name_py, value) in kwargs {
+            // All keyword arguments should be UTF8 strings, but we'll check, just in case.
+            if let Ok(kwarg_name) = kwarg_name_py.downcast::<PyString>()?.to_str() {
```
adamreichold, line 372: This was not changed in the PR, but shouldn't we fail instead of silently ignoring non-UTF-8 keyword arguments?

src/impl_/extract_argument.rs
```diff
@@ -255,57 +257,29 @@ impl FunctionDescription {
...
-            let (positional_parameters, varargs) = args_slice.split_at(num_positional_parameters);
-            output[..num_positional_parameters].copy_from_slice(positional_parameters);
-            V::handle_varargs_fastcall(py, varargs, self)?
-        } else {
-            output[..positional_args_provided].copy_from_slice(args_slice);
-            V::handle_varargs_fastcall(py, &[], self)?
-        };
+        // Handle positional arguments
+        let (positional_parameters, remaining) =
+            if positional_args_provided > num_positional_parameters {
```
adamreichold, line 262: If we are at code-golfing this, wouldn't

```rust
args_slice.split_at(positional_args_provided.min(num_positional_parameters))
```

suffice?

src/impl_/extract_argument.rs
davidhewitt: Non-UTF-8 keywords can't possibly be keywords to a `#[pyfunction]`, so we can just forward them to the varkeywords handler, which will either raise or add to `**kwargs` accordingly.

davidhewitt: I've pushed a comment to clarify this.

src/impl_/extract_argument.rs
davidhewitt: Nice! I force-pushed to also move some other code down to make this even tighter :)

src/impl_/extract_argument.rs
adamreichold: I see, we do not actually skip them, just forward them instead.

PR #2805, opened 2022-12-14
check changelog using nox

davidhewitt:
I wanted to get the green tick when the `CI-skip-changelog` label was applied 😝 

... and I didn't think it hurt to use the same tech we use for the rest of the pipeline.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -8,28 +8,8 @@ jobs:
   check:
     name: Check changelog entry
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


messense (approved):
Looks great!

davidhewitt:
bors r=messense

davidhewitt:
bors retry

davidhewitt:
bors r=messense

davidhewitt:
(I think what happened here is that the unicode emoji was blowing up with Windows file encodings, let's see if this passes now I removed it...)

PR #2806, opened 2022-12-14
add .python-version file

davidhewitt:
This looks like it's recommended for `actions/setup-python` if we don't specify a version in the CI.yml 

It may also be helpful for contributors testing using pyenv?


messense:
> It may also be helpful for contributors testing using pyenv?

IMO it's a bit annoying to have `.python-version` when you want to test another Python version. It'd be fine if `pyenv` provides override methods so you don't have remove `.python-version` or change its content.

davidhewitt:
It looks like pyenv does provide an override: https://github.com/pyenv/pyenv#understanding-python-version-selection

If I understand correctly, this means that having this file will default users to 3.11, and they can use `pyenv shell` command to override for a given shell session.

messense (approved):
bors r+ p=1

PR #2822, opened 2022-12-19
Add conversion support for `std::time::Duration`

messense:
Mostly copied from https://github.com/PyO3/pyo3/blob/main/src/conversions/chrono.rs

### DIFF
--- a/src/conversions/std/time.rs
+++ b/src/conversions/std/time.rs
@@ -1,0 +1,153 @@
+#![cfg(not(Py_LIMITED_API))]
+
+use crate::types::{PyDelta, PyDeltaAccess};
+use crate::{FromPyObject, IntoPy, PyAny, PyObject, PyResult, Python, ToPyObject};
+use std::convert::TryInto;
+use std::time::Duration;
+
+const DAY_SECONDS: u64 = 60 * 60 * 24;
+
+impl ToPyObject for Duration {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        let total_secs = self.as_secs();
+        // Total number of days
+        let days = total_secs / DAY_SECONDS;
+        // Remainder of seconds
+        // .try_into() converts i64 to i32, but this should never overflow
+        // since it's at most the number of seconds per day
+        let secs = (total_secs - days * DAY_SECONDS).try_into().unwrap();
+        // Fractional part of the duration
+        // Same for the conversion from i64 to i32
+        let micros = self.subsec_micros().try_into().unwrap();
+
+        // We do not need to check i64 to i32 cast from rust because
+        // python will panic with OverflowError.
+        // We pass true as the `normalize` parameter since we'd need to do several checks here to
+        // avoid that, and it shouldn't have a big performance impact.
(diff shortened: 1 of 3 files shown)


src/conversions/std/time.rs
```diff
@@ -0,0 +1,153 @@
...
+            let high_days: i32 = 1000000000;
+            // This panics on PyDelta::new
+            assert!(panic::catch_unwind(|| {
+                let pydelta = PyDelta::new(py, high_days, 0, 0, true).unwrap();
+                if let Ok(_duration) = pydelta.extract::<Duration>() {
+                    // So we should never get here
+                }
+            })
+            .is_err());
+        });
```
davidhewitt, line 151: Maybe rather than having the `.unwrap()` on the `PyDelta` construction, it's enough to have something like:

```rust
assert!(PyDelta::new(py, low_days, 0, 0, true).is_err());
let _ = Duration::new(low_days * DAY_SECONDS, 0); // doesn't panic
```

However, it looks like Python durations can be negative, whereas I think `std::time::Duration` cannot hold negative values, so these tests may want to check that behaviour.

src/conversions/std/time.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl FromPyObject<'_> for Duration {
+    fn extract(ob: &PyAny) -> PyResult<Duration> {
+        let delta: &PyDelta = ob.downcast()?;
+        // Python size are much lower than rust size so we do not need bound checks.
+        // 0 <= microseconds < 1000000
+        // 0 <= seconds < 3600*24
+        // -999999999 <= days <= 999999999
+        let days: u64 = delta.get_days().try_into().unwrap();
```
davidhewitt, line 46: Given `std::time::Duration` can't be negative, should we give a nicer error e.g. 

```
OverflowError: negative `timedelta` cannot be converted to `std::time::Duration`
```

src/conversions/std/time.rs
```diff
@@ -0,0 +1,153 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::panic;
+
+    #[test]
+    fn test_std_duration_topyobject() {
+        // Utility function used to check different durations.
+        // The `name` parameter is used to identify the check in case of a failure.
+        let check = |name: &'static str, delta: Duration, py_days, py_seconds, py_ms| {
```
davidhewitt, line 65: Personally I prefer to have this as a separate helper function `_test_duration_to_py` and then each `check` call becomes its own function e.g. `test_delta_zero_to_py`. Not a huge thing, just makes it slightly easier to see each test case and what their intended purpose is.

davidhewitt (commented):
Thanks! We can definitely add this, just some suggestions / questions around the edge cases & tests.

davidhewitt:
On further reflection, maybe adding this is a footgun given that negative durations in Python are trivial to construct? It might be better to just support the `chrono` duration, given that one can be negative.

messense:
Yeah, since it's quite easy to [convert std `Duration` to `chrono::Duration`](https://docs.rs/chrono/latest/chrono/struct.Duration.html#method.from_std), only support `chrono` duration is fine for my use case.

PR #2827, opened 2022-12-25
pypy: enable `PyList::get_item_unchecked`

davidhewitt:
Split out from #2826. Approved previously as part of that review.

PR #2828, opened 2022-12-25
move set-minimal-package-versions to nox

davidhewitt:
Split from #2826 because #2827 couldn't merge due to `hermit-abi` MSRV break, which this PR also fixes (by pinning its dependent `num_cpus` back).


adamreichold (approved):

davidhewitt:
bors r=adamreichold

PR #2829, opened 2022-12-25
netlify: add forward slashes using redirect files

davidhewitt:
Following up on #2820, another try to get the guide to redirect nicely, by adding files which handle the redirects.


davidhewitt:
Hmm, doesn't look like this is working either (at least, not without introducing infinite redirect loops). I can't find a nice solution for this, URLs such as https://pyo3.rs/v0.15.0 will have to continue to have bad CSS for now.

PR #2831, opened 2022-12-26
add gh-pages content directly in netlify site

davidhewitt:
Last attempt to find a solution for #2819.

By adding the GitHub pages content directly to the netlify site during build, netlify should be able to make better URL redirects.


davidhewitt:
Well this seems to work quite well, so I'm going to proceed to merge it ☺️

PR #2833, opened 2022-12-26
ci: fix flaky test_pyobject_drop_without_gil_doesnt_decrease_refcnt

### LINKED ISSUE #2723: Flaky test `test_pyobject_drop_without_gil_doesnt_decrease_refcnt`
```
---- gil::tests::test_pyobject_drop_without_gil_doesnt_decrease_refcnt stdout ----
thread 'gil::tests::test_pyobject_drop_without_gil_doesnt_decrease_refcnt' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', src/gil.rs:619:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:483
   1: std::panicking::begin_panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:437
   2: pyo3::gil::tests::test_pyobject_drop_without_gil_doesnt_decrease_refcnt
             at ./src/gil.rs:619
   3: pyo3::gil::tests::test_pyobject_drop_without_gil_doesnt_decrease_refcnt::{{closure}}
             at ./src/gil.rs:601
   4: core::ops::function::FnOnce::call_once
             at /home/runner/.rustup/toolchains/1.48.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:227
   5: core::ops::function::FnOnce::call_once
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/ops/function.rs:227
```

https://github.com/PyO3/pyo3/actions/runs/3402526740/jobs/5658406562

davidhewitt:
Closes #2723 

I think @adamreichold is correct in what causes the flakiness, and the solution in this case seems straightforward enough - just pass `reference` off to another thread to drop it (as that thread won't have the GIL), no events needed.

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -499,7 +499,7 @@ mod tests {
     use super::{gil_is_acquired, GILPool, GIL_COUNT, OWNED_OBJECTS, POOL};
     use crate::{ffi, gil, AsPyPointer, IntoPyPointer, PyObject, Python, ToPyObject};
     use parking_lot::{const_mutex, Condvar, Mutex};
-    use std::ptr::NonNull;
+    use std::{ptr::NonNull, sync::atomic::Ordering};
 
     fn get_object(py: Python<'_>) -> PyObject {
         // Convenience function for getting a single unique object, using `new_pool` so as to leave
@@ -513,6 +513,17 @@ mod tests {
 
     fn owned_object_count() -> usize {
         OWNED_OBJECTS.with(|holder| holder.borrow().len())
+    }
+
+    fn pool_not_dirty() -> bool {
+        !POOL.dirty.load(Ordering::SeqCst)
+    }
+
+    fn pool_dirty_with(
+        inc_refs: Vec<NonNull<ffi::PyObject>>,
+        dec_refs: Vec<NonNull<ffi::PyObject>>,
+    ) -> bool {
+        *POOL.pointer_ops.lock() == (inc_refs, dec_refs)
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


adamreichold (approved):
The solution of just never releasing the GIL is much tidier than what I anticipated in #2723. Very nice!

bors r+

adamreichold:
Yet Clippy can never be appeased...

davidhewitt:
bors r=adamreichold

src/gil.rs
```diff
@@ -577,54 +588,51 @@ mod tests {
...
+        let obj = Python::with_gil(|py| {
+            let obj = get_object(py);
+            // Create a reference to drop without the GIL.
+            let reference = obj.clone_ref(py);
+
+            assert_eq!(obj.get_refcnt(py), 2);
+            assert!(pool_not_dirty());
+
+            // Drop reference in a separate thread which doesn't have the GIL.
+            std::thread::spawn(move || drop(reference)).join().unwrap();
```
adamreichold, line 619: I suspect this test function needs a

```rust
#[cfg(not(target_arch = "wasm32"))] // We are building wasm Python with pthreads disabled
```

attribute like the other tests which spawn threads.

adamreichold:
(I suspect bors does not listen to me since I could not push to the protected branch any way, but let's try nevertheless.)

bors r+

davidhewitt:
> I suspect bors does not listen to me

Hmm looks like just worked here, can check its config if you think there's an issue?

Thanks for fixing up this one 😁

PR #2836, opened 2022-12-27
ci: use lld on windows and macos

### LINKED ISSUE #2834: Ensure fast linker is used by CI jobs
Our CI jobs, especially the ones getting coverage, spent a lot of time building and running doc tests. Since each test is compiled separately, a lot of time is often spent linking those executables and we should make sure our CI jobs use a faster linker, e.g. LLD is often significantly faster and easily enabled by adding `-Clink-arg=-fuse-ld=lld` to `RUSTFLAGS`.

Potentially, [mold](https://github.com/rui314/mold) could provide further improvements but I have no personal experience using it.

davidhewitt:
Closes #2834 

Looks like `lld` is supported by Rust on Windows and Linux only, and `mold` is only available for free on Linux, so I didn't bother investigating `mold`.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,6 +1,12 @@
 [alias]
 xtask = "run --package xtask --"
 
+# Faster linking in CI for doctests where lld is available
+[target.'cfg(any(target_os = "windows", target_os = "linux"))']
+rustflags = [
+    "-Clink-arg=-fuse-ld=lld"
+]
+
 [target.'cfg(feature = "cargo-clippy")']
 rustflags = [
     # TODO: remove these allows once msrv increased from 1.48


.cargo/config
```diff
@@ -1,6 +1,12 @@
 [alias]
 xtask = "run --package xtask --"
 
+# Faster linking in CI for doctests where lld is available
+[target.'cfg(any(target_os = "windows", target_os = "unix"))']
```
messense, line 5: I don't think `target_os = "unix"` is valid?

.cargo/config
```diff
@@ -1,6 +1,12 @@
 [alias]
 xtask = "run --package xtask --"
 
+# Faster linking in CI for doctests where lld is available
```
adamreichold, line 4: By putting it here instead of injecting it only in the CI, don't we force anybody building PyO3 locally to have LLD available as well? Or is it shipped with rustup nowadays?

adamreichold:
Before merging this, we should double-check that it actually helps, e.g. the last few `coverage-ubuntu` job on the `staging` branch took between 26 and 32 minutes.

.cargo/config
adamreichold: Indeed, https://doc.rust-lang.org/reference/conditional-compilation.html#target_os, so either `target_os = "linux"` or `target_familiy = "unix"`.

I think using the former is safer and would suffice for our CI, but considering https://github.com/PyO3/pyo3/pull/2836#discussion_r1057623555, we might want to handle this via the workflow configuration in any case?

.cargo/config
davidhewitt: Hmm, great question. Unfortunately I'm running on macOS so can't test that.

Anyway, it looks like putting this here breaks netlify, so, I guess we should put it in the GitHub Actions. I'll take a look at  doing so...

davidhewitt:
Hmm, it doesn't seem to make much of a difference (the test steps still appear to be taking ~2.5 minutes each). Plus it looks like it might be quite fiddly!

I'm going to back out of this one for now; maybe some other tidying up of CI can help first, e.g.:
- merge test & coverage jobs.
- run doctests fewer times - e.g. maybe just on latest Python? I think the moment they don't contribute to coverage anyway, and mostly we just want to check these examples compile and rely on unit tests for real verification.

PR #2837, opened 2022-12-27
netlify: fix examples option on latest nightly

davidhewitt:
Looks like `-Zrustdoc-scrape-examples` no longer needs `=examples` value.


davidhewitt:
(The netlify build reports as failed, but after clearing cache and retrying this did work, so I'm going to merge it and we should see the problems go away.)

PR #2842, opened 2022-12-27
Stop leaking in `new_closure`

davidhewitt:
This is a rebase of #2690 which simplifies the `MaybeLeaked` abstraction from that PR with just `Cow<'static, CStr>`.

This enabled me to annotate with `FIXME` all the places where we still leak; I wonder if we could potentially use `GILOnceCell` in future and statics to avoid those. As those callsities are in `#[pyclass]` and `#[pyfunction]` these are effectively in statics anyway, but it would be nice to tidy up.

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -594,11 +594,6 @@ impl PyErr {
         }
     }
 
-    pub(crate) fn write_unraisable(self, py: Python<'_>, context: PyObject) {
-        self.restore(py);
-        unsafe { ffi::PyErr_WriteUnraisable(context.as_ptr()) };
-    }
-
     #[inline]
     fn from_state(state: PyErrState) -> PyErr {
         PyErr {
--- a/src/impl_/pymethods.rs
+++ b/src/impl_/pymethods.rs
@@ -1,6 +1,7 @@
-use crate::internal_tricks::{extract_cstr_or_leak_cstring, NulByteInString};
+use crate::exceptions::PyValueError;
 use crate::{ffi, IntoPy, Py, PyAny, PyErr, PyObject, PyResult, PyTraverseError, Python};
-use std::ffi::CStr;
+use std::borrow::Cow;
+use std::ffi::{CStr, CString};
 use std::fmt;
 use std::os::raw::c_int;
 
@@ -95,6 +96,12 @@ pub struct PyClassAttributeDef {
     pub(crate) meth: PyClassAttributeFactory,
 }
 
+impl PyClassAttributeDef {
+    pub(crate) fn attribute_c_string(&self) -> PyResult<Cow<'static, CStr>> {
+        extract_c_string(self.name, "class attribute name cannot contain nul bytes")
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct PyGetterDef {
     pub(crate) name: &'static str,
@@ -161,7 +168,7 @@ impl PyMethodDef {
     }
 
     /// Convert `PyMethodDef` to Python method definition struct `ffi::PyMethodDef`
-    pub(crate) fn as_method_def(&self) -> Result<ffi::PyMethodDef, NulByteInString> {
+    pub(crate) fn as_method_def(&self) -> PyResult<(ffi::PyMethodDef, PyMethodDefDestructor)> {
         let meth = match self.ml_meth {
             PyMethodType::PyCFunction(meth) => ffi::PyMethodDefPointer {
                 PyCFunction: meth.0,
@@ -175,12 +182,16 @@ impl PyMethodDef {
             },
         };
 
-        Ok(ffi::PyMethodDef {
-            ml_name: get_name(self.ml_name)?.as_ptr(),
+        let name = get_name(self.ml_name)?;
+        let doc = get_doc(self.ml_doc)?;
+        let def = ffi::PyMethodDef {
+            ml_name: name.as_ptr(),
             ml_meth: meth,
             ml_flags: self.ml_flags,
-            ml_doc: get_doc(self.ml_doc)?.as_ptr(),
-        })
+            ml_doc: doc.as_ptr(),
+        };
+        let destructor = PyMethodDefDestructor { name, doc };
+        Ok((def, destructor))
     }
 }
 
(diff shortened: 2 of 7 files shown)


src/impl_/pymethods.rs
```diff
@@ -214,10 +228,15 @@ impl PyGetterDef {
...
+            let name = get_name(self.name).unwrap();
+            dst.name = name.as_ptr() as _;
+            // FIXME: stop leaking name
+            std::mem::forget(name);
         }
         if dst.doc.is_null() {
-            dst.doc = get_doc(self.doc).unwrap().as_ptr() as _;
+            let doc = get_doc(self.doc).unwrap();
+            dst.doc = doc.as_ptr() as _;
+            // FIXME: stop leaking doc
```
adamreichold, line 239: Isn't there a call to `mem::forget` missing here?

src/impl_/pymethods.rs
```diff
@@ -236,21 +255,26 @@ impl PySetterDef {
...
```
adamreichold, line 266: Isn't there a call to `mem::forget` missing here?

src/impl_/pymethods.rs
```diff
@@ -288,3 +320,16 @@ where
...
+    if bytes.last() == Some(&0) {
+        // Last byte is a nul; try to create as a CStr
+        let c_str = CStr::from_bytes_with_nul(bytes).map_err(|_| PyValueError::new_err(err_msg))?;
+        Ok(Cow::Borrowed(c_str))
+    } else {
+        // Allocate a new CString for this
+        let c_string = CString::new(bytes).map_err(|_| PyValueError::new_err(err_msg))?;
+        Ok(Cow::Owned(c_string))
+    }
+}
```
mejrs, line 335: Can we handle a `""` by pointing to a static `"\0"` instead? 

e.g. 
```
match bytes {
        [] => point to static "\0"
        [.., 0] => try &'static CStr
        _ => try CString
    }
```

src/impl_/pymethods.rs
```diff
@@ -1,6 +1,7 @@
-use crate::internal_tricks::{extract_cstr_or_leak_cstring, NulByteInString};
```
adamreichold, line 1: Does `extract_cstr_or_leak_cstring` have other call sites? Would they benefit from the same treatment?

src/impl_/pymethods.rs
```diff
@@ -175,12 +182,19 @@ impl PyMethodDef {
...
+        let def = ffi::PyMethodDef {
+            ml_name: name.as_ptr(),
             ml_meth: meth,
             ml_flags: self.ml_flags,
-            ml_doc: get_doc(self.ml_doc)?.as_ptr(),
-        })
+            ml_doc: doc.as_ptr(),
+        };
+        let destructor = PyMethodDefDestructor {
+            name: Some(name),
```
adamreichold, line 194: Is `PyMethodDefDestructor` ever constructed with `None` values? Can the outer `Option` layer be dropped entirely?

src/impl_/pymethods.rs
davidhewitt: Actually `extract_cstr_or_leak_cstring` is removed completely in this PR, the replacement `extract_c_string` in `pymethods.rs` is private to that module.

adamreichold: Sorry for the noise. This gets dropped later on...

src/impl_/pymethods.rs
davidhewitt: Sorry, yes, I tried to force-push to correct this but pushed to the old PR by mistake 🤦

src/impl_/pymethods.rs
davidhewitt: As above.

src/impl_/pymethods.rs
davidhewitt: Oooh, I like this, will try now!

adamreichold (approved):
Except for the question of whether we need to outer `Option` layer, this looks good to me.

I still wonder whether we should dog-food our own capsule API, but maybe this is best left as a separate PR?

src/impl_/pymethods.rs
davidhewitt: Will drop the option layer.

src/err/mod.rs
```diff
@@ -594,11 +594,6 @@ impl PyErr {
         }
     }
 
-    pub(crate) fn write_unraisable(self, py: Python<'_>, context: PyObject) {
-        self.restore(py);
-        unsafe { ffi::PyErr_WriteUnraisable(context.as_ptr()) };
-    }
```
davidhewitt, line 600: Was only used in the `drop_closure` code now replaced by our `PyCapsule` API.

src/types/function.rs
```diff
@@ -173,10 +116,62 @@ impl PyCFunction {
...
+            let boxed_fn: &ClosureDestructor<F> = &*(ffi::PyCapsule_GetPointer(
+                capsule_ptr,
+                CLOSURE_CAPSULE_NAME.as_ptr() as *const _,
+            ) as *mut ClosureDestructor<F>);
+            let args = py.from_borrowed_ptr::<PyTuple>(args);
+            let kwargs = py.from_borrowed_ptr_or_opt::<PyDict>(kwargs);
+            crate::callback::convert(py, (boxed_fn.closure)(args, kwargs))
+        },
+    )
+}
```
davidhewitt: This was just moved down from above.

davidhewitt (commented):
Now using our `PyCapsule::new` 😄

src/types/function.rs
```diff
@@ -173,10 +117,64 @@ impl PyCFunction {
...
+                mod_ptr,
+                module_name,
+            ))
+        }
     }
 }
 
+fn closure_capsule_name(py: Python<'_>) -> &'static CStr {
+    static CLOSURE_CAPSULE_NAME: GILOnceCell<&CStr> = GILOnceCell::new();
+    CLOSURE_CAPSULE_NAME.get_or_init(py, || CStr::from_bytes_with_nul(b"pyo3-closure\0").unwrap())
```
adamreichold, line 138: Does it really help to memoize calling `CStr::from_bytes_with_nul` using a byte slice literal? I would almost expect the compiler to fully evaluate this at compile time.

davidhewitt: Good point, I'll reduce this to just call `CStr::from_bytes_with_nul(b"pyo3-closure\0").unwrap()` and then merge.

davidhewitt:
bors r=adamreichold

PR #2844, opened 2022-12-28
update URL in comment to point at correct place

alex:


alex:
I don't believe this needs a newsfragment, can that check be skipped?

adamreichold:
> I don't believe this needs a newsfragment, can that check be skipped?

Yes, by adding the `CI-skip-changelog` label to the pull request.

alex:
thanks

tests/test_class_basics.rs
```diff
@@ -148,7 +148,7 @@ struct EmptyClassInModule {}
 
 // Ignored because heap types do not show up as being in builtins, instead they
 // raise AttributeError:
-// https://github.com/python/cpython/blob/master/Objects/typeobject.c#L544-L573
+// https://github.com/python/cpython/blob/f9c5e3f5f61cd380f8a17c814766fc3730b7fbdf/Objects/typeobject.c#L544-L573
```
adamreichold, line 151: Could we link a tag instead of a commit hash? Those should be stable as well but would give more context as to when this was valid.

alex: sure

adamreichold:
I took the liberty of squashing those commits (as bors cannot do it on-demand).

bors r+

PR #2851, opened 2023-01-03
RFC: use bors squash merges

davidhewitt:
This turns on bors' "squash merge" option.

Previously if a PR had a number of small revisions which were only really meaningful during the review process, I would use Github's squash-and-merge button. Now, with bors, I find myself occasionally force-pushing to flatten PRs, which slows me down as I can't do it until I get to a laptop.

By making bors do this, we will always get a squashed PR, which helps keep a clean git history while also speeding me up. The downside is that occasionally it can be nice to keep several commits separate; with this setting doing so would require separate PRs. (Which might be a good thing anyway?)

What do you folks think of enabling this setting?

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -5,3 +5,4 @@ status = [
     "conclusion",
 ]
 timeout_sec = 21600
+use_squash_merge = true


adamreichold:
> What do you folks think of enabling this setting?

As someone who usually invests a considerable amount of time into rebasing branches to provide a well-curated set of commits, I am not happy about enabling this unconditionally. It would be best if bors had an inline command to request squash merging on a case by case basis. Until that happens, I would prefer manually squashing those PR that need it using the command line. (Of course, this requires enabling the "maintainers can push changes" check box. It also may be easier using the GitHub CLI than plain Git, but have not tried that myself.)

PR #2859, opened 2023-01-05
Remove `libpython.so` from cross compile error message

messense:
See https://github.com/PyO3/pyo3/discussions/2858.

Same as https://github.com/PyO3/maturin/pull/1389


davidhewitt (approved):
bors r+

Issue #2853, opened 2023-01-03
Memory leak on pyo3 object created within Rust


haixuanTao:
### Bug Description

I think there is a memory leak when some args are used by a `call_method1` in a Rust function that is being packaged and called from Python. Those args does not get removed until the very end of the function.

In my case, the args in question were `PyBytes` and `PyDict`. I do not know if it is specifically linked to this type.

To fix it, I have manually removed the reference with:
```rust
                unsafe {
                    ffi::Py_DECREF(bytes.as_ptr());
                    ffi::Py_DECREF(input_dict.as_ptr());
                }
```
See: https://github.com/dora-rs/dora/pull/168
This fix was inspired by: https://github.com/PyO3/pyo3/issues/1801

### Steps to Reproduce

You'll find a minimal reproducible example here: https://github.com/haixuanTao/pyo3_memory_leak

The expected behaviour is that the PyBytes get unallocated at each loop step.

### Your operating system and version

Linux Ubuntu 20.04

### Your Python version (`python --version`)

Python 3.7.9

### Your Rust version (`rustc --version`)

rustc 1.66.0 (69f9c33d7 2022-12-12)

### Your PyO3 version

0.16

### How did you install python? Did you use a virtualenv?

conda

mejrs:
It doesn't leak memory for me (windows 11, python 3.10, pyo3 main). Can you try with a more recent pyo3 version?

Also, merging https://github.com/dora-rs/dora/pull/168 is a bad idea, this will blow up really badly if it either doesn't always happen or if you switch to a pyo3 version that doesn't leak.

mejrs:
Also, can you reproduce this without using numpy? We can't do anything if that has the memory leak.

haixuanTao:
I'm pretty sure, Numpy does not leak on `numpy.array`. I just use it as a placeholder, but, you can try with another module that can use a large bytearray. 

For info It also leak on `python 3.10, pyo3 0.17.3` for me. 


https://user-images.githubusercontent.com/22787340/210444879-e79af64f-e508-4335-aaea-a55d9fd42289.mp4

I would love another solution that does not use `unsafe` and therefore does not blow up, but it has been the only solution i have found.

adamreichold:
[...]

PR #2864, opened 2023-01-07
Add a section on memory management for `extension`

haixuanTao:
Adding a special case of memory management when writing an extension.

This is a documentation of: https://github.com/PyO3/pyo3/issues/1056 and https://github.com/PyO3/pyo3/issues/2853

### DIFF
--- a/guide/src/memory.md
+++ b/guide/src/memory.md
@@ -68,6 +68,10 @@ bound to the `GILPool`, not the for loop.  The `GILPool` isn't dropped until
 the end of the `with_gil()` closure, at which point the 10 copies of `hello`
 are finally released to the Python garbage collector.
 
+This is often the case when writing extension as `#[pyfunction]` or `#[pymethod]` will lock 
+the `GILPool` at the beginning of the function call and only release it at 
+the returns. For more on this issue: [#1056](https://github.com/PyO3/pyo3/issues/1056)
+
 In general we don't want unbounded memory growth during loops!  One workaround
 is to acquire and release the GIL with each iteration of the loop.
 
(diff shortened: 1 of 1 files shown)


guide/src/memory.md
```diff
@@ -226,3 +226,53 @@ Python::with_gil(|py| {
 # Ok(())
 # }
 ```
+
+## On writing extension
+
+Using `#[pyfunction]` and `#[pymethods]` makes `Python::with_gil` a no-op and
```
adamreichold, line 232: This is not really correct. Using `Python::allow_threads` implies that `Python::with_gil` is not a no-op.

I think, the main observation is that within `#[pyfunction]` and `#[pymethod]` is already held and hence Python objects owned references tied to the GIL lifetime will usually be released only when the `#[pyfunction]` or `#[pymethod]` returns.

(The above also suggests another correction: It isn't any Python objects, but only those owned by GIL-bound Rust references.)

I would also suggest that the text stresses that this is usually not an issue except when the `#[pyfunction]` or `#[pymethod]` are long-running and/or perform a lot of transient allocations. This does happen, but it is _not_ the common case.

Finally, the section is highly redundant to the existing one on "GIL-bound memory" and I would prefer if the information was integrated into that section. Since that section already contains everything related to the behaviour within loops and how to apply a `GILPool`, I think the main additions this PR should make are a) how this interactions with `#[pyfunction]` and `#[pymethod]` and b) link to #1056 for further information on future design directions.

haixuanTao: Got it. 

I have made the section a small comment on the "GIL-bound memory" section, with a small explanation of the mechanism.

davidhewitt (approved):
Thanks! I've force-pushed to adjust wording and remove the unneeded newsfragment (no need to log doc changes).

bors r+

PR #2871, opened 2023-01-10
netlify: generate all redirects automatically

davidhewitt:
Uses the versions from the pulled `gh-pages` tarball to generate the `docs.rs` redirects. Simplifies the release process.


messense (approved):
Nice!

davidhewitt:
bors r=messense

op8867555:
Sorry if this isn't the right place to report, I think there is an extra `>` character not removed (.netlify/build.sh:32), which make pyo3.rs always redirect to a 404 now.

davidhewitt:
Ooof! Many thanks @op8867555 for reporting, much appreciated and definitely a suitable place to report 😁

davidhewitt:
Will fix in #2874

Issue #2817, opened 2022-12-18
3.11.1 and 3.12 alphas need additional path configuration on Windows


davidhewitt:
See https://github.com/PyO3/pyo3/actions/runs/3721944864/jobs/6312442577#step:11:1094 - we have CI failures on Windows along the lines of 

```
thread 'exceptions::asyncio::tests::TimeoutError' panicked at 'Can not import module asyncio: ModuleNotFoundError: No module named '_socket'
```

Appears to be the following upstream issue: https://github.com/python/cpython/issues/100171

We may need to do something smarter in `auto-initialize` depending on what the outcome of decisions upstream are. We have had several issues of interacting with virtualenvs in the past, e.g. #1741 #1896, maybe it's now time anyway to implement a new "default" in `prepare_freethreaded_python` which covers what users are most likely to want?

davidhewitt:
It seems this is no longer the case.

PR #2873, opened 2023-01-10
A new example that shows how to integrate Python plugins that use pyclasses into a Rust app

alexpyattaev:
Example showing integration of a Python plugin into a Rust app while having option to test pyclass based API without the main app.  This also illustrates some aspects related to import of Python modules into a Rust app while also having an API module available for the Python code to be able to produce Rust objects. 

CI seems to fail on my local machine for reasons unrelated to the example just added:
```
error: unused macro definition: `check_struct`
  --> pyo3-ffi-check/src/main.rs:13:18
   |
13 |     macro_rules! check_struct {
   |                  ^^^^^^^^^^^^
   |
```


examples/plugin/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+    let path = Path::new("./");
+    //do useful work
+    Python::with_gil(|py| {
+        //add the current directory to import path of Python (do not use this in production!)
+        let syspath: &PyList = py.import("sys")?.getattr("path")?.extract()?;
+        syspath.insert(0, &path)?;
+        println!("Import path is: {:?}", syspath);
+
+        // Now we can load our plugin.py file
+        let plugin = PyModule::import(py, "plugin")?;
```
mejrs, line 21: You can just call the initialization function, and avoid messing with sys.path entirely:
```rust
let module = PyModule::new(py, "plugin")?;
plugin_api::plugin_api(py, &module)?;
```

examples/plugin/src/main.rs
```diff
@@ -0,0 +1,38 @@
+use plugin_api::plugin_api as pylib_module;
+use pyo3::prelude::*;
+use pyo3::types::PyList;
+use std::path::Path;
+
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
```
mejrs, line 6: These bounds shouldn't be necessary
```suggestion
fn main() -> Result<(), Box<dyn std::error::Error>> {
```

examples/README.md
```diff
@@ -10,6 +10,7 @@ Below is a brief description of each of these:
 | `maturin-starter` | A template project which is configured to use [`maturin`](https://github.com/PyO3/maturin) for development. |
 | `setuptools-rust-starter` | A template project which is configured to use [`setuptools_rust`](https://github.com/PyO3/setuptools-rust/) for development. |
 | `word-count` | A quick performance comparison between word counter implementations written in each of Rust and Python. |
+| `plugin` | Illustrates how to make Python plugins work within a Rust application. Allows to test plugin API separately from the main application. |
```
mejrs, line 13: This description is too long. Also, can this be useful for more things than just testing? It would be nice to use that as an example instead.

```suggestion
| `plugin` | Shows how to call a python extension module from Rust. |
```

Alternatively, you could focus on testing instead:

```suggestion
| `plugin` | Testing your python extension module from Rust. |
```

examples/plugin/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+        // Now we can load our plugin.py file
+        let plugin = PyModule::import(py, "plugin")?;
+        // and call start function there, which will return a python reference to Gadget
+        let gadget = plugin.getattr("start")?.call0()?;
+        {
+            //now we extract the rust struct from python object
+            let mut gadget_rs: PyRefMut<'_,plugin_api::Gadget> = gadget.extract()?;
+            // we can now modify it as if it was a native rust struct
+            gadget_rs.prop = 42;
+            //which includes access to rust-only fields that are not visible to python
```
mejrs, line 29: This is an important advantage of this approach 👍 . You should put more emphasis on it. This entire example is about testing your python module *while* accessing its internal state, but that's not really mentioned anywhere else.

mejrs (commented):
Thanks for the PR! I have some thoughts:

This PR uses the word "plugin" a lot. What does it mean? If you mean "extension module", you should call it that, not "plugin". We don't call it that anywhere, so this is very likely to confuse readers, and it confused me too, until I realized you meant extension module (I think).


I'm also not sure this is the correct approach. Personally I'd update the existing examples - just put this at the bottom of their lib.rs:
```rust
#[cfg(test)]
mod tests{
    #[test] 
    fn test_module(){
         // make module
         // call module etc
    }
}
```
This could be added to `maturin new` too. 


There are two advantages to this:
- When users create their first python module, they won't have to figure out how to test it. It will just be there
- Putting it in lib.rs (or anywhere else in the crate) allows users to access *crate* private things. `main.rs` does not let you do this.

What do you think?

examples/plugin/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+The plugin API is in a separate crate `plugin_api`, so you can test it separately from the main app. 
+
+To build the API only package, first install `maturin`:
+
+```shell
+pip install maturin
+```
+
+To build use `./build_extmod.sh`, simply using `maturin develop` will fail to produce a viable extension module due to the features arrangement of PyO3.
```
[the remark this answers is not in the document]
alexpyattaev: Yes, that could be done too, but I always got it wrong myself while working on this. Remembering to run the correct option was super counter-intuitive to me, as muscle memory kept me from doing it. I suppose a compromise would be to specify exact command in the readme, but keep the script also.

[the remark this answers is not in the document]
adamreichold: > I suppose a compromise would be to specify exact command in the readme, but keep the script also.

Since this is documentation and should be generally useful, I would also suggest putting the full invocation in the readme and reduce the clutter by leaving out the script. Different people use different tools to automate such tasks (shell scripts, Makefiles, Justfiles, Python scripts, etc.) and providing a single one-liner script will not materially help them IMHO.

alexpyattaev:
> Thanks for the PR!
You are welcome, I figured if it took me 2 days to figure this out, it may be worth sharing. Also PyO3 is awesome =)

> This PR uses the word "plugin" a lot. What does it mean? If you mean "extension module", you should call it that, not "plugin". We don't call it that anywhere, so this is very likely to confuse readers, and it confused me too until I realized you meant extension module (I think).
By plugin I mean a literal plugin you'd put in a bigger application. Specifically, if you are building something like blender video editor, you would want to have the "main" application in rust, but then also have some scripting API as well. For that to work, you would normally want to define a bunch of your application-specific interface classes, such that Python plugins can construct complex objects (in example it is the Gadget pyclass) and then pass them into the Rust code. It is much more efficient than parsing complex data structures from Python classes, and also allows you to test the interface properly before integration into the main app.


> I'm also not sure this is the correct approach. Personally I'd update the existing examples - just put this at the bottom of their lib.rs:
> 
> ```rust
> #[cfg(test)]
> mod tests{
>     #[test] 
>     fn test_module(){
>          // make module
>          // call module etc
>     }
> }
> ```
> 
> This could be added to `maturin new` too.
> 
> There are two advantages to this:
> 
>     * When users create their first python module, they won't have to figure out how to test it. It will just be there
> 
>     * Putting it in lib.rs (or anywhere else in the crate) allows users to access _crate_ private things. `main.rs` does not let you do this.
> 
> 
> What do you think?

This will only work with custom build configuration (as in my build_extmod.sh) to make the actual plugin. With the way linking is set up with Python you can not have both tests and extension modules working at the same time. Though it is a "larger issue" in some sense.

examples/README.md
alexpyattaev: I think the point was missed. The example is about making plugins in python and PyO3 for Rust application. It is not just about how  to test them (but also about that as if you just pile everything into one crate it becomes a nightmare to test). Imagine if you had to start an entire video editing software just to test a one line change in your python API. That would be a total disaster.

examples/plugin/src/main.rs
alexpyattaev: Well not just testing, the reason I built this entire example is to figure out how to make Python scripting API in my application. Since my application is fairly complicated, I wanted to test certain pieces of the API separately from the main app. So access to internal state was a must. Also it gives ability to "hide" stuff from Python plugins that they do not need to access, while having access from Rust.

examples/plugin/src/main.rs
alexpyattaev: No, this will not work as expected, since plugin.py will not land on the import path. And then, if plugin.py has any dependencies of its own, they will fail to import which will confuse the user until they figure out they need to mess with import path for it to work as you'd expect Python to work. I will actually add some deps to plugin.py to illustrate this.

adamreichold:
Depending on what you guys decide on w.r.t. the plug-ins-versus-better-testing angle, I think it should be stressed that while putting the plug-in API in a separate crate and testing plug-ins this way is possible, it is _not_ possible to build the plug-in API/host separately from the main application, i.e. it currently needs to be linked statically into the main application to keep their ABI compatible.

alexpyattaev:
That is very much true, and is a fundamental limitation of Rust toolchain at the moment. Not sure if such remarks belong in the example though, but feel free to suggest where they would go.

examples/plugin/README.md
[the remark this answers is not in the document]
alexpyattaev: This has been done as suggested.

examples/plugin/README.md
```diff
@@ -0,0 +1,48 @@
...
+interface types that can be used to exchange data between Rust and Python. This also deals with how to separately test and load python modules.
+
+# Building and Testing
+## Host application
+To run the app itself, you only need to run 
+
+```shell
+cargo run
+```
+It will build the app, as well as the plugin API, then run the app, load th plugin and show it working.
```
mejrs, line 13:
```suggestion
It will build the app, as well as the plugin API, then run the app, load the plugin and show it working.
```

examples/README.md
```diff
@@ -10,6 +10,7 @@ Below is a brief description of each of these:
 | `maturin-starter` | A template project which is configured to use [`maturin`](https://github.com/PyO3/maturin) for development. |
 | `setuptools-rust-starter` | A template project which is configured to use [`setuptools_rust`](https://github.com/PyO3/setuptools-rust/) for development. |
 | `word-count` | A quick performance comparison between word counter implementations written in each of Rust and Python. |
+| `plugin` | Illustrates how to make Python plugins work within a Rust application in a way that allows one to separate the Python API from the main application. |
```
mejrs, line 13: > By plugin I mean a literal plugin you'd put in a bigger application. Specifically, if you are building something like blender video editor, you would want to have the "main" application in rust, but then also have some scripting API as well. For that to work, you would normally want to define a bunch of your application-specific interface classes, such that Python plugins can construct complex objects (in example it is the Gadget pyclass) and then pass them into the Rust code. It is much more efficient than parsing complex data structures from Python classes, and also allows you to test the interface properly before integration into the main app.

Ah, I get it 👍  I still think the description is too long though, how about

```suggestion
| `plugin` | Illustrates how to use Python as a scripting language within a Rust application |
```

mejrs (commented):
Sorry for not getting back to you sooner 😅. I have some comments remaining.


Anyway, I wonder what the others think E.g. @davidhewitt ?

alexpyattaev:
No rush here=) Thanks for the suggestions, I think your description is quite well fitting, and indeed way shorter.

examples/plugin/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Copying this example
+
+Use [`cargo-generate`](https://crates.io/crates/cargo-generate):
+
+```bash
+$ cargo install cargo-generate
+$ cargo generate --git https://github.com/PyO3/pyo3 examples/plugin
+```
+
+(`cargo generate` will take a little while to clone the PyO3 repo first; be patient when waiting for the command to run.)
```
davidhewitt, line 48: I think to make `cargo-generste` work we will need to add a few template files, I'll check this and push when I have a moment.

davidhewitt (commented):
Yep, this looks excellent to me, thanks very much to both of you  for putting this example together!

I'll add the template files and merge when I get a moment 👍

alexpyattaev:
Not sure what bors is unhappy with... Any ideas?

davidhewitt:
bors retry

davidhewitt:
There's a compatibility issue with our test suite on 3.11.1 (#2817) so we're using 3.11.0 for testing at the moment (3.11.2 should fix this). I think we just hit a rate limit with all the bors tests running!

PR #2875, opened 2023-01-11
Update `memoffset`, `send_wrapper` and `hashbrown`

messense:


noxfile.py
```diff
@@ -430,7 +430,7 @@ def set_minimal_package_versions(session: nox.Session):
             )
 
     _run_cargo_set_package_version(session, "indexmap", "1.6.2")
-    _run_cargo_set_package_version(session, "hashbrown:0.12.3", "0.9.1")
+    _run_cargo_set_package_version(session, "hashbrown:0.13.1", "0.9.1")
```
adamreichold, line 433: I understand that this might be out of scope for this PR, but since we are running custom Python code here anyway, maybe we could make this less brittle by parsing `Cargo.lock` and downgrading all instances of `hashbrown` instead of a specific given one?

Similar to even if more principles than https://github.com/PyO3/rust-numpy/blob/main/.github/workflows/ci.yml#L153

Otherwise, I personally like the simplicity of having a `Cargo.lock.msrv` in the repository even if that is a little less self-documenting.

messense:
```bash
nox -s set-minimal-package-versions
nox > Running session set-minimal-package-versions
nox > Creating virtual environment (virtualenv) using python in .nox/set-minimal-package-versions
nox > cargo update
    Updating crates.io index
nox > cargo update --manifest-path=examples/decorator/Cargo.toml
    Updating crates.io index
nox > cargo update --manifest-path=examples/maturin-starter/Cargo.toml
    Updating crates.io index
nox > cargo update --manifest-path=examples/setuptools-rust-starter/Cargo.toml
    Updating crates.io index
nox > cargo update --manifest-path=examples/word-count/Cargo.toml
    Updating crates.io index
nox > cargo update -p indexmap:1.9.2 --precise 1.6.2
    Updating crates.io index
    Updating hashbrown v0.12.3 -> v0.9.1
    Updating indexmap v1.9.2 -> v1.6.2
nox > cargo update -p hashbrown:0.12.3 --precise 0.9.1
error: package ID specification `hashbrown@0.12.3` did not match any packages
Did you mean one of these?

  hashbrown@0.9.1
  hashbrown@0.13.2
nox > Command cargo update -p hashbrown:0.12.3 --precise 0.9.1 failed with exit code 101
nox > Session set-minimal-package-versions failed.
```

It's much more complicated here in pyo3 because they depend on each other. 🤦‍♂️

messense: Here is the changes I made: https://gist.github.com/messense/2f4aeadc263af0430f57622bbc8584ae

messense:
bors try

noxfile.py
adamreichold: Re-reading `Cargo.lock` is a great idea!

adamreichold:
Still needs a changelog entry.

noxfile.py
```diff
@@ -429,44 +452,37 @@ def set_minimal_package_versions(session: nox.Session):
...
-
-    _run_cargo_set_package_version(
-        session, "rayon", "1.5.3", project="examples/word-count"
-    )
-    _run_cargo_set_package_version(
-        session, "rayon-core", "1.9.3", project="examples/word-count"
-    )
+        lock_file = "Cargo.lock"
+        if project:
+            lock_file = os.path.join(project, "Cargo.lock")
```
adamreichold, line 458: We already import `Path` above, so I think you could simplify this a bit like

```python
lock_file = Path(project) / "Cargo.lock"
```

and

```python
cargo_lock = toml.loads(lock_file.read_text())
```

noxfile.py
```diff
@@ -429,44 +452,37 @@ def set_minimal_package_versions(session: nox.Session):
...
+                name = pkg["name"]
+                if name not in min_pkg_versions:
+                    continue
+                pkg_versions[name].append(pkg["version"])
+            return pkg_versions
+
+        pkg_versions = load_pkg_versions()
+        for (pkg_name, min_version) in min_pkg_versions.items():
+            versions = pkg_versions.get(pkg_name)
+            if not versions:
```
adamreichold, line 475: You could use

```python
versions = pkg_versions.get(pkg_name, list())
```

to avoid the extra check for `None`?

noxfile.py
```diff
@@ -407,13 +407,36 @@ def check_changelog(session: nox.Session):
 
 @nox.session(name="set-minimal-package-versions")
```
adamreichold, line 408: Somewhat unrelated, but I noticed it from the log you pasted here, shouldn't this session use `venv_backend="none"` as well since we don't need the venv here?

adamreichold (approved):
Thank you for going the extra mile on the minimal-versions Nox session!

The nits on the Python code are merely that, so LGTM.

davidhewitt:
This looks great, many thanks both!

PR #2881, opened 2023-01-15
Rework `PyAny::is_instance_of` for performance

davidhewitt:
I was talking to @samuelcolvin about the fastest way to identify object types (relevant e.g. for `pythonize` and also `pydantic-core`) and noticed that `PyAny::is_instance_of` is quite unoptimised because it expands to an ffi call to `PyObject_IsInstance`.

This PR proposes `PyAny::is_instance_of::<T>(obj)` is changed to be equivalent to `T::is_type_of(obj)`, plus add a sprinkling of inlining. We often have implementations such as `PyDict_Check` which can pretty much be optimised away to just checking a bit on the type object.

The accompanying benchmark to run through a bunch of object types is approx 40% faster after making this change.

For completeness, I've also added `PyAny::is_exact_instance` and `PyAny::is_exact_instance_of`, to pair with `T::is_exact_type_of`. (This could be split into a separate PR if preferred.)

### DIFF
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -665,6 +665,7 @@ impl PyAny {
     /// Returns whether the object is considered to be None.
     ///
     /// This is equivalent to the Python expression `self is None`.
+    #[inline]
     pub fn is_none(&self) -> bool {
         unsafe { ffi::Py_None() == self.as_ptr() }
     }
@@ -778,7 +779,7 @@ impl PyAny {
     ///
     /// Python::with_gil(|py| {
     ///     let dict = PyDict::new(py);
-    ///     assert!(dict.is_instance_of::<PyAny>().unwrap());
+    ///     assert!(dict.is_instance_of::<PyAny>());
     ///     let any: &PyAny = dict.as_ref();
     ///
     ///     assert!(any.downcast::<PyDict>().is_ok());
@@ -904,6 +905,7 @@ impl PyAny {
     /// Checks whether this object is an instance of type `ty`.
     ///
     /// This is equivalent to the Python expression `isinstance(self, ty)`.
+    #[inline]
     pub fn is_instance(&self, ty: &PyAny) -> PyResult<bool> {
         let result = unsafe { ffi::PyObject_IsInstance(self.as_ptr(), ty.as_ptr()) };
         err::error_on_minusone(self.py(), result)?;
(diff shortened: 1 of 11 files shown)


src/types/any.rs
```diff
@@ -863,18 +864,39 @@ impl PyAny {
...
         err::error_on_minusone(self.py(), result)?;
         Ok(result == 1)
     }
 
+    /// Checks whether this object is an instance of exactly type `ty` (not a subclass).
+    ///
+    /// This is equivalent to the Python expression `type(self) is ty`.
+    #[inline]
+    pub fn is_exact_instance(&self, ty: &PyAny) -> PyResult<bool> {
+        let result = unsafe { ffi::PyObject_IsInstance(self.as_ptr(), ty.as_ptr()) };
```
birkenfeld, line 879: Should this not be different from above?

davidhewitt: Oooh yes, shame on me for not adding tests. I'll split this out into a real PR later 🫣

src/types/any.rs
```diff
@@ -863,18 +864,39 @@ impl PyAny {
...
+        err::error_on_minusone(self.py(), result)?;
+        Ok(result == 1)
+    }
+
     /// Checks whether this object is an instance of type `T`.
     ///
     /// This is equivalent to the Python expression `isinstance(self, T)`,
     /// if the type `T` is known at compile time.
+    #[inline]
     pub fn is_instance_of<T: PyTypeInfo>(&self) -> PyResult<bool> {
```
birkenfeld, line 889: Hmm, now it would be nice again to remove the `Result` wrapping :)

davidhewitt:
Given the proximity to releasing 0.18 and that there's a couple of design choices / refinements to add here, I'm going to park this until after 0.18 is out.

samuelcolvin:
Idea looks great, thanks so much. But I agree, no massive hurry.

davidhewitt:
Rebased, changed `PyAny::is_instance_of` to return `bool` instead of `PyResult<bool>` and split out `is_exact_instance[_of]` into #3161.

The benchmark added in `bench_any` is about 30% faster for these changes compared to `main`. It's nice, though I'm not entirely sure it's worth the churn.

adamreichold (approved):
> The benchmark added in bench_any is about 30% faster for these changes compared to main. It's nice, though I'm not entirely sure it's worth the churn.

Considering the source of the churn is an improvement IMHO (dropping the `Result` layer) and that idiomatic Python API will often check types before committing to an implementation due the high cost of error/exception handling, I would say this is worth it.

Hot paths would probably still benefit from look-up tables as discussed in #3104 but this could give a good boost across the ecosystem where the LUT approach would be inappropriately heavy-weight.

davidhewitt:
Ok, that's a good argument in favour. Let's ship it!

bors r=adamreichold

PR #2886, opened 2023-01-18
send errors in `__releasebuffer__` to `sys.unraisablehook`

davidhewitt:
I noticed that in the `__releasebuffer__` implementation we currently diverge from documentation by allowing `-> PyResult<()>` return values, but they cause crashes later down the line.

e.g. without the other changes in this PR, the new test case would crash with the following message:

```
ValueError: oh dear

SystemError: <class 'bytes'> returned a result with an exception set
```

After some deliberation I decided to allow `-> PyResult<()>` returns because:
 - it keeps the macro implementation and usage simpler
 - errors might be produced by the `__releasebuffer__` internals anyway (e.g. due to `PyCell` locking, or invalid self type passed)

As a result, this PR cleans up the case discussed to send errors to `sys.unraisablehook`, and updates the documentation to be clearer on the allowed behaviour.

### DIFF
--- a/newsfragments/2886.fixed.md
+++ b/newsfragments/2886.fixed.md
@@ -1,0 +1,1 @@
+Send errors returned by `__releasebuffer__` to `sys.unraisablehook` rather than causing `SystemError`.
--- a/src/impl_/pyclass.rs
+++ b/src/impl_/pyclass.rs
@@ -931,13 +931,8 @@ pub(crate) unsafe extern "C" fn tp_dealloc<T: PyClass>(obj: *mut ffi::PyObject) 
     /// A wrapper because PyCellLayout::tp_dealloc currently takes the py argument last
     /// (which is different to the rest of the trampolines which take py first)
     #[inline]
-    #[allow(clippy::unnecessary_wraps)]
-    unsafe fn trampoline_dealloc_wrapper<T: PyClass>(
-        py: Python<'_>,
-        slf: *mut ffi::PyObject,
-    ) -> PyResult<()> {
+    unsafe fn trampoline_dealloc_wrapper<T: PyClass>(py: Python<'_>, slf: *mut ffi::PyObject) {
         T::Layout::tp_dealloc(slf, py);
-        Ok(())
     }
     // TODO change argument order in PyCellLayout::tp_dealloc so this wrapper isn't needed.
     crate::impl_::trampoline::dealloc(obj, trampoline_dealloc_wrapper::<T>)
--- a/src/impl_/trampoline.rs
+++ b/src/impl_/trampoline.rs
@@ -140,15 +140,39 @@ trampolines!(
     ) -> *mut ffi::PyObject;
 
     pub fn unaryfunc(slf: *mut ffi::PyObject) -> *mut ffi::PyObject;
-
-    pub fn dealloc(slf: *mut ffi::PyObject) -> ();
 );
 
 #[cfg(any(not(Py_LIMITED_API), Py_3_11))]
-trampolines! {
+trampoline! {
     pub fn getbufferproc(slf: *mut ffi::PyObject, buf: *mut ffi::Py_buffer, flags: c_int) -> c_int;
+}
 
-    pub fn releasebufferproc(slf: *mut ffi::PyObject, buf: *mut ffi::Py_buffer) -> ();
+#[cfg(any(not(Py_LIMITED_API), Py_3_11))]
+#[inline]
+pub unsafe fn releasebufferproc(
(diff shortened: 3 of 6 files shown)


src/impl_/trampoline.rs
```diff
@@ -201,3 +219,30 @@ where
...
+///
+/// ctx must be either a valid ffi::PyObject or NULL
+unsafe fn trampoline_inner_void<F>(body: F, ctx: *mut ffi::PyObject)
+where
+    F: for<'py> FnOnce(Python<'py>) -> PyResult<()> + UnwindSafe,
+{
+    let trap = PanicTrap::new("uncaught panic at ffi boundary");
+    let pool = unsafe { GILPool::new() };
+    let py = pool.python();
+    if let Err(py_err) = match panic::catch_unwind(move || body(py)) {
```
adamreichold, line 240: Isn't this `.map_err(|payload| PanicException::from_panic_payload(payload))`?

src/impl_/trampoline.rs
```diff
@@ -201,3 +219,30 @@ where
...
+///
+/// Panics during execution are trapped so that they don't propagate through any
+/// outer FFI boundary.
+///
+/// Exceptions produced are sent to `sys.unraisablehook`.
+///
+/// # Safety
+///
+/// ctx must be either a valid ffi::PyObject or NULL
+unsafe fn trampoline_inner_void<F>(body: F, ctx: *mut ffi::PyObject)
```
adamreichold, line 233: I initially stumbled while reading this and I think I would prefer if the name would reference "unraisable", e.g. "trampoline_inner_unraisable" instead of "void".

src/impl_/pyclass.rs
```diff
@@ -931,13 +931,8 @@ pub(crate) unsafe extern "C" fn tp_dealloc<T: PyClass>(obj: *mut ffi::PyObject)
...
     #[inline]
-    #[allow(clippy::unnecessary_wraps)]
-    unsafe fn trampoline_dealloc_wrapper<T: PyClass>(
-        py: Python<'_>,
-        slf: *mut ffi::PyObject,
-    ) -> PyResult<()> {
-        T::Layout::tp_dealloc(slf, py);
-        Ok(())
+    unsafe fn trampoline_dealloc_wrapper<T: PyClass>(py: Python<'_>, slf: *mut ffi::PyObject) {
+        T::Layout::tp_dealloc(slf, py)
```
adamreichold, line 935: I think the semicolon after the call would help immediately seeing that this is not forwarding a return value.

newsfragments/2886.fixed.md
```diff
@@ -0,0 +1 @@
+Send errors returned by `__releasebuffer__` to `sys.unraisablehook` rather than causing `SystemError`.
```
adamreichold, line 1: This does also apply the same change for our implementations of dealloc, does it not?

tests/test_buffer_protocol.rs
```diff
@@ -132,3 +133,93 @@ fn test_buffer_referenced() {
...
+    }
+
+    unsafe fn __releasebuffer__(&self, _view: *mut ffi::Py_buffer) -> PyResult<()> {
+        Err(PyValueError::new_err("oh dear"))
+    }
+}
+
+#[test]
+fn test_releasebuffer_unraisable_error() {
+    static UNRAISABLE_CALLED: Mutex<Option<(PyErr, PyObject)>> = const_mutex(None);
```
adamreichold, line 197: This is really a nit as the effort might not be worth it at all, but wouldn't it be possible to use a `#[pyclass]` to capture the unraisable value without resorting to global state by setting the unraisablehook to use a method call? (I am not sure if we have the API to bind the self poarameter though?)

src/impl_/trampoline.rs
davidhewitt: Actually the type coming from catch unwind is something like `Result<Result<(), PyErr>, PanicPayload>`. I think `.unwrap_or_else()` is what I'm looking for.

newsfragments/2886.fixed.md
davidhewitt: Yes, though `dealloc` is internal and we don't currently return any errors from it, so I don't think this would change any user-facing behaviour.

davidhewitt:
Thanks as always for the thorough review! I've pushed a commit which should address all of those points and hopefully pass CI.

adamreichold (approved):

davidhewitt:
bors r=adamreichold

davidhewitt:
bors r=adamreichold

davidhewitt:
bors r=adamreichold

PR #2897, opened 2023-01-20
add link on how to obtain GIL to guide

### LINKED ISSUE #1562: Docs: clarify handling of Python references while Rust holds the GIL
The relationship between periods holding the GIL and object reference lifetime is not clearly described in the library documentation, leading to bugs like #1547 and #1556.  The guide makes a mention in "Advanced topics", but this reads like something that isn't important to know, calling it "invisible".

There should be at least mention in the library reference (maybe just on the topmost level in "Ownership and Lifetimes"?) and also in the guide in a more introductory section.

### LINKED ISSUE #2164: Doc improvements: `pyfunction` usages and acquiring `Python` object
I found some details making me a little confused when reading the docs, which I think can be improved.

The first is about the argument parsing for `pyfunction` macro. I searched for how to add keyword arguments to function and surprised to find that the related article is the `arg` macro under "Method argument" section. And later I found that there's [one small section in function documentation](https://pyo3.rs/v0.15.1/function.html#argument-parsing) referring to the method argument section. I think it's more intuitive to put the documentation of specifying argument behavior under the page for normal function, and then add a referring link on the method argument section.

The second is about how to acquire the `Python` generally in PyO3. There are documentations to the usage of `py: Python` object in the [GIL page](https://pyo3.rs/v0.15.1/types.html), but it seems that there's no guide of how to acquire a `Python` object (e.g. for creating a Python list). In my case, I want a function to return objects of different type from the same function, so I have to use a `PyObject` as a return type (no sure if this is the designed way), then the code will look like
```rust
// just for demonstration
#[pyfunction]
fn some_func(input: u8) -> PyResult<PyObject> {
    match input {
        0 => Ok(py.None()), // how to get a `py`?
        1 => Ok(PyList::new(py)),
        2 => Ok(PyDict::new(py)),
        _ => Err(...)
    }
}
```
Then after looking around the docs, I found some example directly put `py: Python` as the first argument of the function, so I tried
```rust
#[pyfunction]
fn some_func(py: Python, input: u8) -> PyResult<PyObject> {
    match input {
        0 => Ok(py.None()), // how to get a `py`?
        1 => Ok(PyList::new(py)),
        2 => Ok(PyDict::new(py)),
        _ => Err(...)
    }
}
```
and it works! I'm not sure if this is the idiomatic way to do this, so it's better to document how to get a `py` object. Thanks!

davidhewitt:
Closes #2164 

Most of that issue is already addressed with the new `#[pyo3(signature = (...))]` option and its guide page. The final suggestion to make it easier to find recommendations on how to acquire the GIL seems reasonable to me.

(EDIT: original issue link was wrong)

### DIFF
--- a/guide/src/types.md
+++ b/guide/src/types.md
@@ -6,7 +6,7 @@ an overview of their intended meaning, with examples when each type is best
 used.
 
 
-## Mutability and Rust types
+## The Python GIL, mutability, and Rust types
 
 Since Python has no concept of ownership, and works solely with boxed objects,
 any Python object can be referenced any number of times, and mutation is allowed
@@ -43,6 +43,9 @@ an object's ownership has been passed to the Python interpreter, ensuring
 references is done at runtime using `PyCell`, a scheme very similar to
 `std::cell::RefCell`.
 
+### Accessing the Python GIL
+
+To get hold of a `Python<'py>` token to prove the GIL is held, consult [PyO3's documentation][obtaining-py].
 
 ## Object types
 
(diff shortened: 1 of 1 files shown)


birkenfeld (approved):

adamreichold (approved):

adamreichold:
bors r=birkenfeld

Issue #2891, opened 2023-01-19
`py_run!` creates unexpected scope issues


davidhewitt:
### Bug Description

Try the following code (originally discussed in #2889).

```rust
#[pyfunction]
fn report_unraisable(py: Python) {
    use pyo3::exceptions::PyRuntimeError;
    let err = PyRuntimeError::new_err("foo");
    err.write_unraisable(py, None);

    let err = PyRuntimeError::new_err("bar");
    err.write_unraisable(py, Some(py.NotImplemented().as_ref(py)));
}

#[test]
fn test_write_unraisable() {
    Python::with_gil(|py| {
        let report_unraisable = wrap_pyfunction!(report_unraisable, py).unwrap();
        let locals = PyDict::new(py);
        locals
            .set_item("report_unraisable", report_unraisable)
            .unwrap();

        py_run!(
            py,
            report_unraisable,
            r#"
                import sys

                captured = []
                def report(data):
                    captured.append(list(data))

                original_hook = sys.unraisablehook
                try:
                    sys.unraisablehook = report
                    report_unraisable()

                    assert len(captured) == 2

                    assert captured[0][0] is RuntimeError
                    assert str(captured[0][1]) == 'foo'
                    assert captured[0][4] is None

                    assert captured[1][0] is RuntimeError
                    assert str(captured[1][1]) == 'bar'
                    assert captured[1][4] is NotImplemented
                finally:
                    sys.unraisablehook = original_hook
            "#
        );
    });
}
```

The `report` function defined within the `py_run!` source will print errors like the following, and the test will fail:

```
Exception ignored in sys.unraisablehook: <function report at 0x102e94040>
Traceback (most recent call last):
  File "<string>", line 5, in report
NameError: name 'captured' is not defined
Exception ignored in sys.unraisablehook: <function report at 0x102e94040>
Traceback (most recent call last):
  File "<string>", line 5, in report
NameError: name 'captured' is not defined
Traceback (most recent call last):
  File "<string>", line 12, in <module>
AssertionError
```

### Steps to Reproduce

As above

### Your operating system and version

macOS Ventura 13.1

### Your Python version (`python --version`)

Python 3.10.9

### Your Rust version (`rustc --version`)

rustc 1.67.0-beta.2

### Your PyO3 version

0.18.0

### How did you install python? Did you use a virtualenv?

brew + vent

PR #2899, opened 2023-01-20
RFC: Provide a special purpose FromPyObject impl for byte slices

### LINKED ISSUE #2888: Improve performance when converting python bytes/bytearray to `Vec<u8>`
Hi all, I noticed a performance issue when extracting a `PyBytes` or a `PyByteArray` object into a `Vec<u8>`.

This is an issue one can easily run into without realizing it. Here's a scenario, let's say we'd like to expose a simple checksum function:
 ```rust
#[pyfunction]
fn checksum(data: &[u8]) -> PyResult<u8> {
    let mut result = 0;
    for x in data {
        result ^= x;
    }
    Ok(result)
}
```
See how it performs against the equivalent python implementation, processing 1MB a hundred times:
```
2.65s call     test_checksum.py::test_perf[py-bytes]
0.00s call     test_checksum.py::test_perf[rs-bytes]
```
Looks really fast! However, it won't accept a bytearray as an argument:
```python
TypeError: argument 'data': 'bytearray' object cannot be converted to 'PyBytes'
```
So we update our implementation to take a `Vec<u8>` instead:
```rust
#[pyfunction]
fn checksum(data: Vec<u8>) -> PyResult<u8> {
    let mut result = 0;
    for x in data {
        result ^= x;
    }
    Ok(result)
}
```
And now the results:
```
2.61s call     test_checksum.py::test_perf[py-bytearray]
2.55s call     test_checksum.py::test_perf[py-bytes]
1.92s call     test_checksum.py::test_perf[rs-bytearray]
1.87s call     test_checksum.py::test_perf[rs-bytes]
```
It performs roughly the same as python, which makes sense if we look at the `FromPyObject` implementation for `Vec<T>`:
https://github.com/PyO3/pyo3/blob/bed4f9d6ee60793b8582e70b9b06cc25b318628d/src/types/sequence.rs#L314-L318

The `bytes`/`bytearray` object is iterated and each item (i.e a python integer) is separately extracted into a `u8`.

This could be fixed by specializing the extract logic in the case of a `Vec<u8>` and use specific methods such as `PyBytes::as_bytes().to_vec()` and `PyByteArray::to_vec()`. Here's a possible patch:
https://gist.github.com/vxgmichel/367e01e8504cb9c9e700a22525e8b68d

With this patch applied, the performance is now similar to what we had with the `&[u8]` slice:
```
2.70s call     test_checksum.py::test_perf[py-bytearray]
2.65s call     test_checksum.py::test_perf[py-bytes]
0.00s call     test_checksum.py::test_perf[rs-bytes]
0.00s call     test_checksum.py::test_perf[rs-bytearray]
```

adamreichold:
This enables efficiently and safely getting a byte slice from either bytes or byte arrays.

The main issue I see here is discoverability, i.e. should this be mention in the docs of `PyBytes` and `PyByteArray` or in the guide?

It is also not completely clear whether this really _fixes_ the issue.

Closes #2888

### DIFF
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -1,4 +1,4 @@
-use crate::{ffi, AsPyPointer, FromPyObject, Py, PyAny, PyResult, Python};
+use crate::{ffi, AsPyPointer, FromPyObject, IntoPy, Py, PyAny, PyResult, Python, ToPyObject};
 use std::borrow::Cow;
 use std::ops::Index;
 use std::os::raw::c_char;
@@ -140,9 +140,23 @@ impl<'source> FromPyObject<'source> for Cow<'source, [u8]> {
     }
 }
 
+impl ToPyObject for Cow<'_, [u8]> {
+    fn to_object(&self, py: Python<'_>) -> Py<PyAny> {
+        PyBytes::new(py, self.as_ref()).into()
+    }
+}
+
+impl IntoPy<Py<PyAny>> for Cow<'_, [u8]> {
+    fn into_py(self, py: Python<'_>) -> Py<PyAny> {
+        self.to_object(py)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    use crate::types::IntoPyDict;
 
     #[test]
     fn test_bytes_index() {
@@ -205,6 +219,16 @@ mod tests {
             something_else_entirely
                 .extract::<Cow<'_, [u8]>>()
                 .unwrap_err();
+
+            let cow = Cow::<[u8]>::Borrowed(b"foobar").into_py(py);
+            let locals = [("cow", cow)].into_py_dict(py);
+            py.run("assert isinstance(cow, bytes)", Some(locals), None)
+                .unwrap();
+
+            let cow = Cow::<[u8]>::Owned(b"foobar".to_vec()).into_py(py);
+            let locals = [("cow", cow)].into_py_dict(py);
+            py.run("assert isinstance(cow, bytes)", Some(locals), None)
+                .unwrap();
         });
     }
 }
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -156,8 +156,6 @@ impl IntoPy<Py<PyAny>> for Cow<'_, [u8]> {
 mod tests {
     use super::*;
 
-    use crate::types::IntoPyDict;
-
     #[test]
     fn test_bytes_index() {
         Python::with_gil(|py| {
@@ -220,15 +218,11 @@ mod tests {
                 .extract::<Cow<'_, [u8]>>()
                 .unwrap_err();
 
-            let cow = Cow::<[u8]>::Borrowed(b"foobar").into_py(py);
-            let locals = [("cow", cow)].into_py_dict(py);
-            py.run("assert isinstance(cow, bytes)", Some(locals), None)
-                .unwrap();
+            let cow = Cow::<[u8]>::Borrowed(b"foobar").to_object(py);
+            assert!(cow.as_ref(py).is_instance_of::<PyBytes>().unwrap());
 
-            let cow = Cow::<[u8]>::Owned(b"foobar".to_vec()).into_py(py);
-            let locals = [("cow", cow)].into_py_dict(py);
-            py.run("assert isinstance(cow, bytes)", Some(locals), None)
-                .unwrap();
+            let cow = Cow::<[u8]>::Owned(b"foobar".to_vec()).to_object(py);
+            assert!(cow.as_ref(py).is_instance_of::<PyBytes>().unwrap());
         });
     }
 }
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -1,8 +1,11 @@
-use crate::{ffi, AsPyPointer, Py, PyAny, PyResult, Python};
+use crate::{ffi, AsPyPointer, FromPyObject, Py, PyAny, PyResult, Python};
+use std::borrow::Cow;
 use std::ops::Index;
 use std::os::raw::c_char;
 use std::slice::SliceIndex;
 use std::str;
+
+use super::bytearray::PyByteArray;
 
 /// Represents a Python `bytes` object.
 ///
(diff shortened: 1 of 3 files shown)


src/types/bytes.rs
```diff
@@ -121,6 +124,18 @@ impl<I: SliceIndex<[u8]>> Index<I> for PyBytes {
     }
 }
 
+/// Special-purpose trait impl to efficiently handle both [`PyBytes`] and [`PyByteArray`]
```
mejrs, line 127: This should also describe the performance characteristics:

```suggestion
/// Special-purpose trait impl to efficiently handle both [`PyBytes`] and [`PyByteArray`]
///
/// If the source object is a `bytes` object, the `Cow` will be borrowed and pointing into the source object, and no copying or heap allocations will happen. If it is a `bytearray`, its contents will be copied to an owned `Cow`.
```

mejrs (approved):
LGTM, please add some tests though 🙂

adamreichold:
> please add some tests though

Certainly, and also a changelog entry. I just wanted to discuss whether we really want this hack and where we need to do document it.

The doc comment on the impl would only be visible in the implementors section of `FromPyObject` AFAIU which is not easily discoverable IMHO.

davidhewitt:
Sneaky! I think this seems reasonable, perhaps we should also add into-py conversions for `Cow<'_, [u8]>` for consistency?

I think I would also try to make this discoverable in the guide, maybe both from https://pyo3.rs/v0.18.0/conversions/tables and on a new page discussing performance techniques? (I've been thinking about adding a performance page for ages anyway.) 

I wonder, should we also ban `Vec<u8>` from `#[pyfunction]` arguments and return values? That would help avoid the performance footgun and reserve them for specialization later without breaking user code at that time. (I mean in a separate PR, not here.)

adamreichold:
> I wonder, should we also ban Vec<u8> from #[pyfunction] arguments and return values? That would help avoid the performance footgun and reserve them for specialization later without breaking user code at that time. (I mean in a separate PR, not here.)

You specifically banning `Vec<u8>`, not `Vec<T>`? How would we do this? Try to parse the argument types in the proc macros? Is that realistically possible on a syntactical level and considering arbitrarily complex types containing `Vec<u8>`?

(I think raising an error in `FromPyObject` is not possible due to the missing `T: 'static` bound in the impl for `Vec<T>` and making this a runtime error would also be poor ergonomics.)

mejrs:
> I wonder, should we also ban Vec<u8> from #[pyfunction] arguments and return values? T

This seems more worthy of a lint than a compile error.

>How would we do this? Try to parse the argument types in the proc macros? Is that realistically possible on a syntactical level and considering arbitrarily complex types containing Vec<u8>?

We'd check whether the type looks like `"Vec<u8>"`. Which is rather messy, I feel.

adamreichold:
> I think this seems reasonable, perhaps we should also add into-py conversions for Cow<'_, [u8]> for consistency?

So `IntoPy<Py<PyAny>> for Cow<'_, [u8]>`? That seems reasonable and could be done symmetrically, i.e. `Borrowed` to `bytes` and `Owned` to `bytearray`. The symmetry is limited as we have to copy the values in both cases, but at least this gives us a "type-driven" way to produce `bytes` and `bytearray`.

> I think I would also try to make this discoverable in the guide, maybe both from https://pyo3.rs/v0.18.0/conversions/tables and on a new page discussing performance techniques? (I've been thinking about adding a performance page for ages anyway.)

I will adjust the table. I would ask for the performance section to be a separate PR which I will try to produce in reasonable amount of time (if no one else does so first). The reason is that I expect the wording to be somewhat contentions as we probably all care very much about performance and the techniques in question can be subtle and their applicability depend on a lot of context.

davidhewitt:
Potentially also `ToPyObject`? Really need to work out those traits 😂

The symmetry is an interesting feature. It's quite pretty, but I think it has a risk of surprising users? They are likely not going to think too hard what variant their `Cow` is on and suddenly their Python type changes. Maybe from different branches of the same function. Seems like the sort of thing which will cause bug reports by people who don't read documentation.

As an alternative, how about changing the `Cow::Owned` branches so that it accepts any byte buffer (using `PyBuffer::get` instead of downcast to `PyByteArray`), and always return `PyBytes`? Users who want a `PyByteArray` can get better performance constructing directly from a `&[u8]` without an extra allocation to have a `Cow::Owned`, and I'm not convinced we need round tripping here.

I feel more strongly about narrowing the to-python value to be `PyBytes` always than I do about widening the input to accept all byte buffers.

davidhewitt:
And yes, @mejrs is right on what I was thinking to do for `Vec`. It's not pretty, but we do do it already for `Python` and `Option`.

adamreichold:
> Potentially also `ToPyObject`? Really need to work out those traits joy

Done.

> As an alternative, how about changing the `Cow::Owned` branches so that it accepts any byte buffer (using `PyBuffer::get` instead of downcast to `PyByteArray`), and always return `PyBytes`? Users who want a `PyByteArray` can get better performance constructing directly from a `&[u8]` without an extra allocation to have a `Cow::Owned`, and I'm not convinced we need round tripping here.

Implemented, but this is conditional not using the limited API or Python 3.11 or later. It also not clear how to include this in the guide.
 
> And yes, @mejrs is right on what I was thinking to do for Vec. It's not pretty, but we do do it already for Python and Option.

I would still prefer a lint then, especially if this is imprecise due to working with syntax only. (I would also suggest a separate PR as well since this one seems to sprawl somewhat.)

One related idea that is a bit out in the woods: Since for `#[pyfunction]` we are not in generic context, but in macro context, [`Deref`-based specialization](https://github.com/dtolnay/case-studies/tree/master/autoref-specialization) might be an option to efficiently handle `Vec<[u8]>`. However, this would mean stepping outside of our `FromPyObject` trait hierarchy and doing something in the generated code that cannot be done just by invoking `FromPyObject`, so I am not sure whether this would be worth it. But it would open up a general avenue towards specialization without compiler support.

davidhewitt:
>  I would still prefer a lint then, especially if this is imprecise due to working with syntax only.

Note that there's no stable way to do lints from macros, so any lints are a clumsy hack like what we do with deprecations.

> One related idea that is a bit out in the woods: Since for #[pyfunction] we are not in generic context, but in macro context, [Deref-based specialization](https://github.com/dtolnay/case-studies/tree/master/autoref-specialization) might be an option to efficiently handle Vec<[u8]>

It's an interesting thought and might work nicely, we could have `PyFunctionInput` and `PyFunctionOutput` traits, and document them with users even? Worth considering in another PR, I agree.

> Implemented, but this is conditional not using the limited API or Python 3.11 or later. It also not clear how to include this in the guide.

Ah, I had overlooked `PyBuffer` was unavailable in all contexts. Yes, need to consider what feels better for users...

birkenfeld:
> Note that there's no stable way to do lints from macros,

I was just thinking about that too. Maybe time for a RFC?

davidhewitt:
@adamreichold what's your feeling about this currently? I think I regret my suggestion of `PyBuffer`, I'm thinking we should narrow the input types back to just `bytes` and `bytearray` to remove the conditionally. Keep the output type as `bytes` only?

After that, probably good to merge this?

(We could leave an issue in the backlog to remind ourselves to widen this conversion in a few years when 3.10 is dropped?)

adamreichold:
Concerning the build error, I am not sure how the builtin `isinstance` can be missing on Python 3.8. Would it be sufficient if I just used our `PyAny::is_instance_of::<PyBytes>()` instead or would that reduce testing power too much?

adamreichold:
Added a commit to use native code instead of inline Python for the type assertions.

bors r=davidhewitt

davidhewitt:
> Concerning the build error, I am not sure how the builtin `isinstance` can be missing on Python 3.8. Would it be sufficient if I just used our `PyAny::is_instance_of::<PyBytes>()` instead or would that reduce testing power too much?

I think probably this is related to #2891 somehow. Needs some investigation. Using `PyAny::is_instance_of` seems fine for now.

Issue #2866, opened 2023-01-08
Support `text_signature` and autogenerated text signature for `#[new]`


davidhewitt:
After #2784 we now support autogenerated `text_signature` for everything except `#[new]`, which can't have any text signature set (it must be put on the `#[pyclass]` instead).

Using some macro and trait magic we should be able to resolve this; I won't do it for 0.18 but perhaps shortly after if nobody else wants to pick this up first (in which case I can point you in the right direction for the implementation).

PR #2904, opened 2023-01-24
refactor docstring generation code

davidhewitt:
As a first step towards #2866 this is a tweak to the macro code which generates Python docstrings. This PR refactors the behaviour so that instead of always creating a `concat!` expression to generate a nul-terminated string, this will only happen if a Rust 1.54+ macro doc is present (e.g. `#[doc = include_str!(...)]`).

The default case now just collects all the `#[doc]` attributes into a single string.

This should make it easier to factor out the `text_signature` formatting, and avoids wasting compile time invoking the `concat!` macro when not necessary.

### DIFF
--- a/pyo3-macros-backend/src/utils.rs
+++ b/pyo3-macros-backend/src/utils.rs
@@ -1,9 +1,9 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, fmt::Write};
 
 // Copyright (c) 2017-present PyO3 Project and Contributors
 use proc_macro2::{Span, TokenStream};
 use quote::ToTokens;
-use syn::{spanned::Spanned, Ident};
+use syn::{punctuated::Punctuated, spanned::Spanned, Ident, Token};
 
 use crate::attributes::CrateAttribute;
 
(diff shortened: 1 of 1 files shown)


pyo3-macros-backend/src/utils.rs
```diff
@@ -57,7 +57,7 @@ pub fn option_type_argument(ty: &syn::Type) -> Option<&syn::Type> {
     None
 }
 
-/// A syntax tree which evaluates to a null-terminated docstring for Python.
+/// A syntax tree which evaluates to a nul-terminated docstring for Python.
 ///
 /// It's built as a `concat!` evaluation, so it's hard to do anything with this
```
mejrs, line 62: So these comments are no longer true? 🙂

pyo3-macros-backend/src/utils.rs
```diff
@@ -70,57 +70,71 @@ pub fn get_doc(
...
+        syn::token::Bang(Span::call_site()).to_tokens(&mut tokens);
+        syn::token::Bracket(Span::call_site()).surround(&mut tokens, |tokens| {
+            parts.to_tokens(tokens);
+            syn::token::Comma(Span::call_site()).to_tokens(tokens);
+            syn::LitStr::new("\0", Span::call_site()).to_tokens(tokens);
+        });
+
+        PythonDoc(tokens)
+    } else {
+        // Just a string doc - return directly will nul terminator
```
mejrs, line 134:
```suggestion
        // Just a string doc - return directly with nul terminator
```

mejrs (approved):

pyo3-macros-backend/src/utils.rs
davidhewitt: Good catch, will update.

davidhewitt:
bors r=mejrs

PR #2911, opened 2023-01-25
add `Ellipsis()` and `is_ellipsis()` methods

### LINKED ISSUE #2906: `PyEllipsis` type
Would be great to have an Ellipsis type, aka `...` in python.

The ffi type seems to exist

https://github.com/PyO3/pyo3/blob/18081ba14e8dfe4cf989ae68c308ea585dbfc734/pyo3-ffi/src/sliceobject.rs#L11-L13

But I'm not sure how I would use it in safe rust code.

samuelcolvin:
fix #2906

 - [x] an entry for this PR in newsfragments - see [https://pyo3.rs/main/contributing.html#documenting-changes]

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -519,6 +519,13 @@ impl<T> Py<T> {
         unsafe { ffi::Py_None() == self.as_ptr() }
     }
 
+    /// Returns whether the object is Ellipsis, e.g. `...`.
+    ///
+    /// This is equivalent to the Python expression `self is ...`.
+    pub fn is_ellipsis(&self) -> bool {
+        unsafe { ffi::Py_Ellipsis() == self.as_ptr() }
+    }
+
     /// Returns whether the object is considered to be true.
     ///
     /// This is equivalent to the Python expression `bool(self)`.
@@ -1145,6 +1152,22 @@ a = A()
             instance.setattr(py, "foo", "bar").unwrap_err();
             Ok(())
         })
+    }
+
+    #[test]
+    fn test_is_ellipsis() {
+        Python::with_gil(|py| {
+            let v = py
+                .eval("...", None, None)
+                .map_err(|e| e.print(py))
+                .unwrap()
+                .to_object(py);
+
+            assert!(v.is_ellipsis());
+
+            let not_ellipsis = 5.to_object(py);
+            assert!(!not_ellipsis.is_ellipsis());
+        });
     }
 
     #[cfg(feature = "macros")]
--- a/src/marker.rs
+++ b/src/marker.rs
@@ -619,6 +619,13 @@ impl<'py> Python<'py> {
         unsafe { PyObject::from_borrowed_ptr(self, ffi::Py_None()) }
     }
 
+    /// Gets the Python builtin value `Ellipsis`, or `...`.
+    #[allow(non_snake_case)] // the Python keyword starts with uppercase
+    #[inline]
+    pub fn Ellipsis(self) -> PyObject {
+        unsafe { PyObject::from_borrowed_ptr(self, ffi::Py_Ellipsis()) }
+    }
+
     /// Gets the Python builtin value `NotImplemented`.
     #[allow(non_snake_case)] // the Python keyword starts with uppercase
     #[inline]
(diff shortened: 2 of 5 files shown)


samuelcolvin:
Seems to really be failing on windows, error is 

```
note: pyo3-8a1fda91fe8449f2.pyo3.3a3efd38-cgu.1.rcgu.o : error LNK2019: unresolved external symbol __Py_EllipsisObject referenced in function 
```

Any ideas?

mejrs:
> Seems to really be failing on windows, error is
> 
> ```
> note: pyo3-8a1fda91fe8449f2.pyo3.3a3efd38-cgu.1.rcgu.o : error LNK2019: unresolved external symbol __Py_EllipsisObject referenced in function 
> ```
> 
> Any ideas?

I'm guessing it needs a `#[cfg_attr(windows, link(name = "pythonXY"))]` to link correctly. I can check tonight if you can't figure it out.

src/instance.rs
```diff
@@ -519,6 +519,13 @@ impl<T> Py<T> {
         unsafe { ffi::Py_None() == self.as_ptr() }
     }
 
+    /// Returns whether the object is an Ellipsis, e.g. `...`.
```
mejrs, line 522: Just a nitpick on the wording - Ellipsis is a singleton so something can't be "an" Ellipsis, it *is* Ellipsis.

samuelcolvin:
> I'm guessing it needs a `#[cfg_attr(windows, link(name = "pythonXY"))]` to link correctly. I can check tonight if you can't figure it out.

Ha, I'm trying exactly that having seen it [here](https://github.com/StarryInternet/pyo3/blob/3408cd206c04739f814d4f57212cff616225ca6f/src/ffi3/sliceobject.rs#L5).

mejrs (approved):
Thanks!

src/instance.rs
samuelcolvin: changed.

samuelcolvin:
Would be great to get this merged and released as I'm currently having to get pyo3 from my branch due to the windows issue. :rocket: :runner: 

See https://github.com/pydantic/pydantic-core/pull/370.

src/marker.rs
```diff
@@ -619,6 +619,13 @@ impl<'py> Python<'py> {
         unsafe { PyObject::from_borrowed_ptr(self, ffi::Py_None()) }
     }
 
+    /// Gets the Python builtin value `Ellipsis`, or `...`.
+    #[allow(non_snake_case)] // the Python keyword starts with uppercase
+    #[inline]
+    pub fn Ellipsis(self) -> PyObject {
```
davidhewitt, line 625: Rather than returning `PyObject` here it would probably be nicer if we returned `&PyEllipsis`, and created a new struct `types::PyEllipsis`. (Would avoid the need to increase a reference count here.)

That would mean that `x.is_instance::<PyEllipsis>()` would just work instead of needing `is_ellipsis`, which does have the upside of avoiding a special method. Not necessarily saying that it's better, perhaps `is_ellipsis` is still warranted.

Also, adding `types::PyEllipsis` would pave the way for instead just having `PyEllipsis::get(py)` to retrieve the singleton, rather than adding this to `Python` which I don't think really needs this method.

davidhewitt (changes requested):
Sorry for the delay. My family's sick with a cold _again_....

I think there's some API details which I'd like to at least discuss whether we can do better before we merge.

src/marker.rs
mejrs: Then we should do this for None and NotImplemented too. The current pr mirrors their api.

davidhewitt: Agreed, and I've been wanting to do that for a while. I guess we can't chance those for 0.18.x, so perhaps we merge this as-is and I'll probably raise a follow-up PR to adjust all three on the way to 0.19?

samuelcolvin: 👍

davidhewitt (approved):
Approved as per comment that I'll try to find some time to add to this on the way to 0.19.

I've just force-pushed to squash-and-rebase, will merge. Thanks!

davidhewitt:
bors r=mejrs,davidhewitt

PR #2912, opened 2023-01-25
Add `PyDict.update()` and `PyDict.update_if_missing()`

### LINKED ISSUE #2910: Support `PyDict_Merge` on `PyDict`
Would be great to have support for [`PyDict_Merge`](https://docs.python.org/3/c-api/dict.html#c.PyDict_Merge) on `PyDict`, e.g. via a `fn merge(other: &PyDict, override: bool)` method.

samuelcolvin:
Fix #2910

Note, I'd also be happy to remove the `override_` argument from merge and perhaps rename it to `update_missing` or similar to give a cleaner API. LMK what you think.

 - [x] an entry for this PR in newsfragments

### DIFF
--- a/newsfragments/2912.added.md
+++ b/newsfragments/2912.added.md
@@ -1,0 +1,1 @@
+Add `PyDict.update()` and `PyDict.update_if_missing()` methods.
--- a/src/types/dict.rs
+++ b/src/types/dict.rs
@@ -257,6 +257,28 @@ impl PyDict {
     pub fn as_mapping(&self) -> &PyMapping {
         unsafe { self.downcast_unchecked() }
     }
+
+    /// Update this dictionary with the key/value pairs from another.
+    ///
+    /// This is equivalent to the Python expression `self.update(other)`. If `other` is a `PyDict`, you may want
+    /// to use `self.update(other.as_mapping())`, note: `PyDict::as_mapping` is a zero-cost conversion.
+    pub fn update(&self, other: &PyMapping) -> PyResult<()> {
+        let py = self.py();
+        unsafe { err::error_on_minusone(py, ffi::PyDict_Update(self.as_ptr(), other.as_ptr())) }
+    }
+
+    /// Add key/value pairs from another dictionary to this one only when they do not exist in this.
+    ///
+    /// This is equivalent to the Python expression `self.update({k: v for k, v in other.items() if k not in self})`.
+    /// If `other` is a `PyDict`, you may want to use `self.update_if_missing(other.as_mapping())`,
+    /// note: `PyDict::as_mapping` is a zero-cost conversion.
+    ///
+    /// This method uses [`PyDict_Merge`](https://docs.python.org/3/c-api/dict.html#c.PyDict_Merge) internally,
+    /// so should have the same performance as `update`.
+    pub fn update_if_missing(&self, other: &PyMapping) -> PyResult<()> {
+        let py = self.py();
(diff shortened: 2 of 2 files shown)


mejrs:
Maybe it would be better to use an enum for this, like
```rust 
pub enum Override{
    ReplaceSelf,
    ReplaceOther,
}```

This gets rid of that _ too 🤮

**Comment by davidhewitt** [comment]
An enum instead of bool can help with documentation, so I'm fine with doing that. Not sure how it removes `_` from the variable name?

**Comment by davidhewitt** [comment]
If you'd like the underscore gone and to use an enum, my suggestion would be to call the parameter `replace_existing` and then have

```rust
enum ReplaceExisting {
    No,
    Yes,
}
```

samuelcolvin:
Surely we just change the variable name to `replace_existing`, although I think it's less clear than the current solution which shadows the cpython c API.

I also think using an enum is significantly uglier than a bool - another import, an unexpected API, and it would make me think there should be more than 2 options.

We could use `over_ride`, but personally I think what's here fine.

mejrs:
> Not sure how it removes `_` from the variable name?

Uh yeah...not sure how I thought that

> I also think using an enum is significantly uglier than a bool

My gut feeling is that this is one of those methods where people are going to mess up. I think I'd have to look up which overrides what half the time 😅 . So maybe splitting this up in multiple methods is a better idea.

samuelcolvin:
> So maybe splitting this up in multiple methods is a better idea.

Well in that case we can drop the case of `merge(&self, other: &PyDict, true)` since according to the C API docs, it has the same behaviour as `update(other)` (and presumably is slower???).

So we could then have:
* `fn update(&self, other: &PyDict)` - unchanged
* `fn update_if_missing(&self, other: &PyDict) -> {merge(other, true)}` - where existing values are not changed, somewhat similar to `setdefault` but for multiple values

The question then is what we call `update_if_missing`???

We could use `update_if_missing`, `update_weak` or just `update_missing` as I mentioned at the beginning.

davidhewitt:
So I looked in the CPython source and it turns out that `PyDict_Merge(x, y, 1)` is exactly equivalent to `PyDict_Update(x, y)`: https://github.com/python/cpython/blob/4246fe977d850f8b78505c982f055d33d52ff339/Objects/dictobject.c#L2932

So it's correct that there are really only two cases here. `update` and `update_if_missing` seem reasonable to me.

Final observation is that these functions support arbitrary mappings, not just `Dict`. So I suggest we go for:

```rust
fn update(&self, other: &PyMapping) -> PyResult<()>
fn update_if_missing(&self, other: &PyMapping) -> PyResult<()> 
```

If you really want `&PyDict` arguments, for completeness we'd need:

```rust
fn update(&self, other: &PyDict) -> PyResult<()>
fn update_if_missing(&self, other: &PyDict) -> PyResult<()> 
fn update_from_mapping(&self, other: &PyMapping) -> PyResult<()>
fn update_if_missing_from_mapping(&self, other: &PyMapping) -> PyResult<()> 
```
 
But there would be no functional difference, and `PyDict::as_mapping` is a zero-cost conversion.

samuelcolvin:
* Updated to `update` and `update_if_missing`
* Switched to `PyMapping`
* updated docs

davidhewitt (approved):
LGTM, thanks! I'll force-push to squash and then merge.

samuelcolvin:
Awesome, thank you.

PR #2914, opened 2023-01-25
correct ffi definition of PyIter_Check

### LINKED ISSUE #2913: `py_any.downcast::<PyIterator>()` behaviour is different on windows vs macos and linux
### Bug Description

This is an extension of https://github.com/PyO3/pyo3/discussions/2902

On Linux and MacOS, `py_any.downcast::<PyIterator>()` is only `Ok` for generators.

On windows it appears to be `Ok` for strings and anything which is a sequence.

### Steps to Reproduce

Should be as simple as

```rs
let s = "foobar".to_object().into_ref(py);
let iter: PyIterator = s.downcast()?;
```

Should fail on linux and macos, but pass on windows. 

### Backtrace

_No response_

Ubuntu

3.10

cargo 1.68.0-nightly (2381cbdb4 2022-12-23)

### Your PyO3 version

v0.18

pyenv

### Additional Info

_No response_

davidhewitt:
Closes #2913 

It looks like what is happening is that PyO3 was relying on an outdated macro form of `PyIter_Check` which is now a CPython implementation detail, which would explain why it was behaving inconsistently on different platforms (likely due to differences in linkers / implementations).

The test I've pushed succeeds, but fails to compile due to a hygiene bug! I'm done for tonight so I'll take a look at that soon and then rebase this after.

### DIFF
--- a/src/types/iterator.rs
+++ b/src/types/iterator.rs
@@ -248,4 +248,78 @@ def fibonacci(target):
             assert_eq!(iter_ref.get_refcnt(), 2);
         })
     }
+
+    #[test]
+    #[cfg(any(not(Py_LIMITED_API), Py_3_8))]
+    #[cfg(feature = "macros")]
+    fn python_class_not_iterator() {
+        use crate::PyErr;
+
+        #[crate::pyclass(crate = "crate")]
+        struct Downcaster {
+            failed: Option<PyErr>,
+        }
+
+        #[crate::pymethods(crate = "crate")]
+        impl Downcaster {
+            fn downcast_iterator(&mut self, obj: &PyAny) {
+                self.failed = Some(obj.downcast::<PyIterator>().unwrap_err().into());
+            }
+        }
+
+        // Regression test for 2913
+        Python::with_gil(|py| {
+            let downcaster = Py::new(py, Downcaster { failed: None }).unwrap();
+            crate::py_run!(
+                py,
+                downcaster,
+                r#"
+                    from collections.abc import Sequence
+
+                    class MySequence(Sequence):
+                        def __init__(self):
+                            self._data = [1, 2, 3]
+
+                        def __getitem__(self, index):
+                            return self._data[index]
+
+                        def __len__(self):
+                            return len(self._data)
+
+                    downcaster.downcast_iterator(MySequence())
+                "#
+            );
+
+            assert_eq!(
+                downcaster.borrow_mut(py).failed.take().unwrap().to_string(),
+                "TypeError: 'MySequence' object cannot be converted to 'Iterator'"
+            );
+        });
+    }
+
+    #[test]
+    #[cfg(any(not(Py_LIMITED_API), Py_3_8))]
+    #[cfg(feature = "macros")]
+    fn python_class_iterator() {
+        #[crate::pyfunction(crate = "crate")]
+        fn assert_iterator(obj: &PyAny) {
+            assert!(obj.downcast::<PyIterator>().is_ok())
+        }
+
+        // Regression test for 2913
+        Python::with_gil(|py| {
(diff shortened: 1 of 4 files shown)


src/types/iterator.rs
```diff
@@ -248,4 +248,49 @@ def fibonacci(target):
             assert_eq!(iter_ref.get_refcnt(), 2);
         })
     }
+
+    #[test]
+    #[cfg(not(all(not(Py_3_8), Py_LIMITED_API)))]
+    #[cfg(feature = "macros")]
+    fn python_class_not_iterator() {
```
samuelcolvin, line 255: might also be woth testing some of the other failure cases turned up in https://github.com/pydantic/pydantic-core/pull/372 (search for `xfail`):
* subclass of `str`
* subclass of `bytes`
* a plain class like:

```py
    class Foobar:
        def __repr__(self):
            return '<Foobar repr>'
```

davidhewitt: I'm confident enough in the underlying mechanism (really it's just checking the presence of a `__next__` implementation) that I think the one test is enough. What I _will_ add, though, is the opposite check - that one which has a `__next__` will pass the check.

davidhewitt:
Rebased & tests adjusted.

adamreichold:
@davidhewitt This (like all other PR) will fail the MSRV build due to trybuild bump MSRV (via basic-toml) in a point release. I think we need to pin it to 1.0.76...

adamreichold:
> @davidhewitt This (like all other PR) will fail the MSRV build due to trybuild bump MSRV (via basic-toml) in a point release. I think we need to pin it to 1.0.76...

I took the liberty of rebasing this onto the current `main` branch which contains that version pin.

davidhewitt:
Thanks! It would have been ok to just leave that to bors too :)

bors r+

davidhewitt:
Looks like I fluffed the `#[cfg]` on the test. Think it's fixed now.

davidhewitt:
bors retry

davidhewitt:
Hmm. It looks like this night have always been broken on Windows.

We might just have to remove this functionality on Python 3.7 (which only has this broken macro form, the function form of `PyIter_Check` was added in 3.8).

davidhewitt:
I've pushed a commit which removes `PyIter_Check` from 3.7, thus removing ability to `.downcast()` to `PyIterator` on 3.7 or accept it as a function argument.

This is a bit unfortunate, but it's broken and I can't see a way to fix. I think this is better than the alternative of leaving this defined but broken on CPython 3.7 on Windows.

I'd welcome feedback here, otherwise I'll merge this sometime ahead of the 0.18.1 release early next week.

samuelcolvin:
Is much prefer `PyIterator` to work in 3.7, at least in Linux and macos.

Otherwise we'd have to drop 3.7 support from pydantic v2, and I've promised to support all python versions that are still alive.

"This thing is broken on windows on 3.7" is fine, but "we don't support 3.7"  will annoy a lot of people.

birkenfeld:
Why not do a `getattr(obj.__class__, "__next__")` under the conditions where `PyIter_Check` is broken? It's slow, but should work?

samuelcolvin:
According to https://pypistats.org/packages/pydantic 3.7 is still the second most common version with 25% of downloads.

Screenshot_20230203-072020

davidhewitt:
> Why not do a `getattr(obj.__class__, "__next__")` under the conditions where `PyIter_Check` is broken? It's slow, but should work?

Good idea - I'll change to that (and I'll use it in `ffi::PyIter_Check` too).

davidhewitt:
Ok, that's pushed - let's see if it works on CI.

davidhewitt:
bors try

davidhewitt:
That's a success! Thanks all for reviews & design assistance here.

bors r+

PR #2918, opened 2023-01-26
Remove some Debug/Eq/PartialEq derives

mejrs:
I'd used them to do `let is_terminator = *(obj.as_ptr()) == PyMemberDef::default();`, which blew up pretty badly. 😆

### DIFF
--- a/pyo3-ffi/src/descrobject.rs
+++ b/pyo3-ffi/src/descrobject.rs
@@ -8,8 +8,13 @@ pub type getter = unsafe extern "C" fn(slf: *mut PyObject, closure: *mut c_void)
 pub type setter =
     unsafe extern "C" fn(slf: *mut PyObject, value: *mut PyObject, closure: *mut c_void) -> c_int;
 
+/// Represents the [PyGetSetDef](https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef)
+/// structure.
+///
+/// Note that CPython may leave fields uninitialized. You must ensure that
+/// `name` != NULL before dereferencing or reading other fields.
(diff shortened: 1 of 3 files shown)


davidhewitt:
I'm ok with removing these, though can a little more context on how you got to that point and what blew up? I don't see how these derives are themselves deferencing pointers, and `obj` was presumably non-null? 😖 

Also unless we think there are soundness concerns, probably need to keep these around until 0.19.

mejrs:
> I'm ok with removing these, though can a little more context on how you got to that point and what blew up?

I was reading these things from random python objects (for fun).

> I don't see how these derives are themselves deferencing pointers, and obj was presumably non-null? 😖

It's not that these derives are broken - the issue is that them existing makes it really easy to make the same mistake I did. So I think it would be better to not have them.

adamreichold:
> It's not that these derives are broken - the issue is that them existing makes it really easy to make the same mistake I did. So I think it would be better to not have them.

It is still not clear to me how the derives would cause any crashes. Was it maybe `*obj.as_ptr()` itself that was crashing?

adamreichold:
Ok, that makes sense. Removing those impls is a breaking change though, so we would at least need the news fragment. I am not sure the foot gun is worth the breakage? Also w.r.t. the `PartialEq` impls etc. present elsewhere in these modules. The impls also take `&self` which requires a fully initialized object, so it was the creation of that reference which triggered the UB. (So the array should have had the type `[MaybeUninit<PyMemberDef>; N]`?)

The comments are definitely helpful IMHO. Would you consider this PR reduced to the comments completely useless?

PR #2921, opened 2023-01-26
Check to see if object is `None` before traversing

### LINKED ISSUE #2915: PyVisit::call segfaults when passed `None`
### Bug Description

When `PyVisit::call` is passed a `None` value, this is [converted into a null pointer][1] and passed to the underlying `visitproc` function. However, (by my understanding) the `visitproc` function isn't meant to handle null pointers as the intended interface is actually [`Py_VISIT`][2] which _does_ do a null check.

[1]: https://github.com/PyO3/pyo3/blob/58596ac631d6385ce0fec1d0bca4e17e6362086e/src/pyclass/gc.rs#L26
[2]: https://docs.python.org/3/c-api/gcsupport.html#c.Py_VISIT

### Steps to Reproduce

Minimal example:
```rust
use pyo3::{prelude::*, PyVisit, PyTraverseError, types::PyType};

#[pyclass]
struct Foo {
    #[pyo3(get, set)]
    other: Option<Py<Foo>>,

    obj: Option<Py<Bar>>,
}

#[pymethods]
impl Foo {
    #[new]
    fn new() -> Self {
        Self { other: None, obj: None }
    }

    fn __traverse__(&self, visit: PyVisit<'_>)  -> Result<(), PyTraverseError> {
        // Segfaults
        visit.call(&self.obj)?;
        Ok(())
    }

    fn __clear__(&mut self) {
        self.obj = None;
    }
}

#[pyclass]
struct Bar;

fn main()
{
    Python::with_gil(|py| {
        let foo = PyType::new::<Foo>(py);
        pyo3::py_run!(py, foo foo, r#"
            import gc
            import time

            item = foo()
            item.other = item
            del item

            gc.collect()
        "#);
    });
}
```

### Backtrace

_No response_

Ubuntu 22, Ubuntu 20

Python 3.10.6, Python 3.8.10

rustc 1.66.1

### Your PyO3 version

PyO3 0.17.0, PyO3 0.18.0

System installation

### Additional Info

_No response_

neachdainn:
Closes #2915

When using the C API directly, the intended way to call `visitproc` is via the `Py_VISIT` macro, which checks to see that the provided pointer is not null before passing it along to `visitproc`. Because PyO3 isn't using the macro, it needs to manually check that the pointer isn't null. Without this check, calling `visit.call(&obj)` where `let obj = None;` will segfault.

### DIFF
--- a/newsfragments/2921.fixed.md
+++ b/newsfragments/2921.fixed.md
@@ -1,0 +1,1 @@
+Traversal visit calls to `Option<T: AsPyPointer>` no longer segfaults when `None`.
--- a/src/pyclass/gc.rs
+++ b/src/pyclass/gc.rs
@@ -23,11 +23,16 @@ impl<'p> PyVisit<'p> {
     where
         T: AsPyPointer,
     {
-        let r = unsafe { (self.visit)(obj.as_ptr(), self.arg) };
-        if r == 0 {
-            Ok(())
+        let ptr = obj.as_ptr();
+        if !ptr.is_null() {
+            let r = unsafe { (self.visit)(ptr, self.arg) };
+            if r == 0 {
+                Ok(())
+            } else {
+                Err(PyTraverseError(r))
+            }
         } else {
-            Err(PyTraverseError(r))
+            Ok(())
         }
     }
 
--- a/tests/test_gc.rs
+++ b/tests/test_gc.rs
@@ -120,6 +120,44 @@ fn gc_integration() {
     });
 }
 
+#[pyclass]
+struct GcNullTraversal {
+    cycle: Option<Py<Self>>,
+    null: Option<Py<Self>>,
+}
+
+#[pymethods]
+impl GcNullTraversal {
+    fn __traverse__(&self, visit: PyVisit<'_>) -> Result<(), PyTraverseError> {
+        visit.call(&self.cycle)?;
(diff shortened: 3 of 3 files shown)


neachdainn:
I read through the contributing guide and I _think_ I did the newsfragment correctly. If it isn't right, feel free to either let me know what I did wrong or just do whatever needs to be done to fix it.

neachdainn:
The failed CI jobs do not seem related to my changes, so I don't plan on addressing them.

mejrs:
Can you add a test? Your example in https://github.com/PyO3/pyo3/issues/2915 should do nicely. You can put it in a test function similar to https://github.com/PyO3/pyo3/pull/2914/files#diff-e28c611e920921574b15af9f361787212c273d1b4327337593639e619176bd6f

> The failed CI jobs do not seem related to my changes, so I don't plan on addressing them.

That's just because some diagnostics have changed in today's new Rust version. If you rebase after https://github.com/PyO3/pyo3/pull/2922 they should pass.

davidhewitt:
Looks good to me, thanks - I think rebasing will fix CI (new rust release).

davidhewitt:
@mejrs is right, a test write be appreciated 😂

neachdainn:
Rebased and unit test added.

davidhewitt (approved):
Thanks!

PR #2929, opened 2023-01-29
docs: Precise the abscense of `py: Python` for the #[pyo3(signature)]

jjerphan:
Hi,

First, thank you for working on PyO3!

I think this adds a precision which was not obvious when migrating to 0.18.0 conventions.

What do you think? Also, should something be added to [the migration guide](https://pyo3.rs/main/migration) in this regard?

Thank you!

PR #2930, opened 2023-01-29
add better error message for Python in signature

davidhewitt:
Inspired by #2929, this just adds a better error message when `Python` arguments are accidentally included in the signature.

### DIFF
--- a/pyo3-macros-backend/src/pyfunction/signature.rs
+++ b/pyo3-macros-backend/src/pyfunction/signature.rs
@@ -352,10 +352,21 @@ impl<'a> FunctionSignature<'a> {
         let mut parse_state = ParseState::Positional;
         let mut python_signature = PythonSignature::default();
 
-        let mut args_iter = arguments.iter_mut().filter(|arg| !arg.py); // Python<'_> arguments don't show on the Python side.
+        let mut args_iter = arguments.iter_mut();
 
-        let mut next_argument_checked = |name: &syn::Ident| match args_iter.next() {
-            Some(fn_arg) => {
+        let mut next_non_py_argument_checked = |name: &syn::Ident| {
+            for fn_arg in args_iter.by_ref() {
+                if fn_arg.py {
+                    // If the user incorrectly tried to include py: Python in the
+                    // signature, give a useful error as a hint.
+                    ensure_spanned!(
+                        name != fn_arg.name,
+                        name.span() => "arguments of type `Python` must not be part of the signature"
+                    );
+                    // Otherwise try next argument.
+                    continue;
+                }
+
                 ensure_spanned!(
                     name == fn_arg.name,
                     name.span() => format!(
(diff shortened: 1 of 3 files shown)


pyo3-macros-backend/src/pyfunction/signature.rs
```diff
@@ -352,29 +352,48 @@ impl<'a> FunctionSignature<'a> {
...
-
-        let mut next_argument_checked = |name: &syn::Ident| match args_iter.next() {
-            Some(fn_arg) => {
-                ensure_spanned!(
-                    name == fn_arg.name,
-                    name.span() => format!(
-                        "expected argument from function definition `{}` but got argument `{}`",
-                        fn_arg.name.unraw(),
-                        name.unraw(),
+        struct ArgsValidator<'b, 'a: 'b>(std::slice::IterMut<'b, FnArg<'a>>);
```
adamreichold, line 355: Was the change from closure to named type required for this to work?

davidhewitt: The simplest way I could see to implement this was to use recursion to skip over `Python`arguments, which makes a horrible mess of closures.

Not sure what I've come up with here is the prettiest either. Maybe give me another shot at this tomorrow and I'll try factor it out differently?

adamreichold: I see, yes recursion and closure would indeed be a mess. I am not opposed to the change, I just asked to make sure that I did not miss anything. (And indeed I did not see the recursion at a first glance.)

For such cases, I think `loop` might help, i.e. turn

```rust
fn func(iter: I) -> R {
  if let Some(val) = iter.next() {
    if condition(val) {
      return func(iter);
    }

    ...
  } else {
    ...
  }
}
```

into

```rust
fn func(iter: I) -> R {
  loop {
    if let Some(val) = iter.next() {
      if condition(val) {
        continue;
      }

      break ...;
    } else {
      break ...;
    }
  }
}
```

mejrs: I could be missing something, but this looks like an awful way to reinvent `retain`.

```rust
let mut err = None;
arguments.retain(|a| /* delete some things and/or set errr */);
if let Some(err){
     return ...
}
```

davidhewitt: Not exactly - we don't want to remove elements from `arguments`, just enrich them based off the attribute information. This might be able to be refactored better once deprecated `#[args]` is removed.

I'll have another go at tidying this up now.

adamreichold: A `for` loop is certainly nicer than an open-coded `loop` loop. LGTM.

(The `.by_ref()` is required because adding `move |...` to the closure would move more than just the iterator into it?)

davidhewitt: The `.by_ref()` is needed because I don't want to move the iterator into the closure - I check at the end of the function that we've consumed everything from it.

davidhewitt:
bors r=adamreichold

PR #2936, opened 2023-02-04
bump maturin version in getting_started.md

jmhodges:
Latest maturin is 0.14.x, so reference that in the docs


messense (approved):
bors r+

PR #2944, opened 2023-02-11
optimize sequence conversion for list and tuple

### LINKED ISSUE #2943: `2.5/3x` Performance regression with `PySequence as PyTryFrom` since `0.16`.
This has been reported downstream in https://github.com/pola-rs/polars/issues/6791

It seems that dispatch to a very simple function has got a lot more expensive. Take this simple program:

```rust
use pyo3::prelude::*;
use pyo3::types::PySequence;

#[pyfunction]
fn get_len(obj: &PyAny) -> PyResult<usize> {
    let seq = <PySequence as PyTryFrom>::try_from(obj)?;
    seq.len()
}

#[pymodule]
fn pyo3_example(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(get_len, m)?)?;
    Ok(())
}
```

And running this snippet with a release build:

```
$ python -m timeit -n 1000000 -s "import pyo3_example as p; values = [1]" "p.get_len(values)"
```

This takes:

- 0.16: best of 5: 92.7 nsec per loop
- 0.17: best of 5: 255 nsec per loop
- 0.18; best of 5: 241 nsec per loop

## Cargo.toml

```toml
[package]
name = "pyo3-example"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
[lib]
name = "pyo3_example"
crate-type = ["cdylib"]

[dependencies]
pyo3 = { version = "0.18", features = ["extension-module"] }
```

davidhewitt:
closes #2943 

Avoid using `PyObject_IsInstance` for checking if lists or tuples are sequences, as we know they are always sequences.

### DIFF
--- a/src/types/sequence.rs
+++ b/src/types/sequence.rs
@@ -8,10 +8,8 @@ use crate::once_cell::GILOnceCell;
 use crate::type_object::PyTypeInfo;
 use crate::types::{PyAny, PyList, PyString, PyTuple, PyType};
 use crate::{ffi, PyNativeType, ToPyObject};
-use crate::{AsPyPointer, IntoPy, IntoPyPointer, Py, Python};
+use crate::{AsPyPointer, IntoPyPointer, Py, Python};
 use crate::{FromPyObject, PyTryFrom};
-
-static SEQUENCE_ABC: GILOnceCell<PyResult<Py<PyType>>> = GILOnceCell::new();
 
 /// Represents a reference to a Python object supporting the sequence protocol.
 #[repr(transparent)]
@@ -318,17 +316,14 @@ where
     Ok(v)
 }
 
-fn get_sequence_abc(py: Python<'_>) -> Result<&PyType, PyErr> {
+static SEQUENCE_ABC: GILOnceCell<Py<PyType>> = GILOnceCell::new();
+
+fn get_sequence_abc(py: Python<'_>) -> PyResult<&PyType> {
     SEQUENCE_ABC
-        .get_or_init(py, || {
-            Ok(py
-                .import("collections.abc")?
-                .getattr("Sequence")?
-                .downcast::<PyType>()?
-                .into_py(py))
+        .get_or_try_init(py, || {
+            py.import("collections.abc")?.getattr("Sequence")?.extract()
         })
-        .as_ref()
-        .map_or_else(|e| Err(e.clone_ref(py)), |t| Ok(t.as_ref(py)))
+        .map(|ty| ty.as_ref(py))
 }
 
 impl<'v> PyTryFrom<'v> for PySequence {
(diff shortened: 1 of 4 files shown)


davidhewitt:
@ritchie46 would you be willing to run your benchmark suite built against this branch to verify this resolves the issue?

If so, we can prepare 0.18.2 patch release with this in a week or two.

src/types/sequence.rs
```diff
@@ -338,11 +333,15 @@ impl<'v> PyTryFrom<'v> for PySequence {
     fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v PySequence, PyDowncastError<'v>> {
         let value = value.into();
```
adamreichold, line 334: I think this ties in somewhat with the discussion to generally overhaul the conversion traits, but if the `Into::into` call is the only reason this method is generic, wouldn't an inner function taking a plain `&PyAny` be a reasonable measure against code bloat?

ritchie46:
> @ritchie46 would you be willing to run your benchmark suite built against this branch to verify this resolves the issue?
> 
> If so, we can prepare 0.18.2 patch release with this in a week or two.

Yeap, seems much better; :+1: 

- 0.18: best of 5: 359 nsec per loop
- this_branch: best of 5: 193 nsec per loop

src/types/sequence.rs
davidhewitt: I agree and actually in my #1308 proposal I hope to push soon I want to just make the whole method take `&PyAny`.

davidhewitt:
Great - so I agree with you @adamreichold that it won't solve the general case (and all the discussion in #2943 about not needing `PySequence` at all for this particular case in `polars`).

I still think this optimization seems worth it for a common case?

src/types/sequence.rs
adamreichold: Fair enough to leave it like this as-is until we resolve the underlying issue.

adamreichold (approved):
> I still think this optimization seems worth it for a common case?

Agreed. (I have not yet messaged bors to give the others a change to weigh in.)

mejrs:
Do we have a benchmark for this? We should catch similar regressions in the future.

adamreichold:
> Do we have a benchmark for this? We should catch similar regressions in the future.

I agree that a benchmark would be helpful but I don't think this is an unexpected regression. I think everybody involved in the review expected this to decrease performance with the memoization of the `collection.abc.Sequence` type object being a partial mitigation due to that. From my perspective, the original report is rather about very specific use cases which are unable or unwilling to tolerate even the intentional change in performance.

davidhewitt:
👍 I'll push a benchmark within a day or two.

davidhewitt:
Benchmark pushed. On my machine these drop from the 100-200ns range to about 2ns 😄

ritchie46:
> Benchmark pushed. On my machine these drop from the 100-200ns range to about 2ns smile

Wow... That that's 1-3 cpu cycles :see_no_evil:

davidhewitt:
Yes, checking for lists and tuples (and subclasses) can be done by checking for a bit set on the type object.

I should probably push a similar PR for dicts & mapping.

davidhewitt:
Think we've given others a few days, I'd like to prep the next patch release soon.

bors r=adamreichold

jakob-keller:
Sounds fantastic, can't wait to give it a spin! Thanks to everyone involved!

Issue #2942, opened 2023-02-10
Extending native types does not always work


AntoineRR:
### Bug Description

I tried to extend various classes implementing [PySizedLayout](https://docs.rs/pyo3/latest/pyo3/type_object/trait.PySizedLayout.html) as stated in the [docs](https://pyo3.rs/v0.18.1/class#inheritance), but it seems only a few of those classes can be used as base class. I tried:
- `PyDict` and `PySet`, which work fine.
- `PyBool` and `PySlice`, which gives this error: `TypeError: type 'bool' (or 'slice') is not an acceptable base type`
- Most importantly, `PyFloat` and `PyFrozenSet` give the following: `Job 1, 'python3 test.py' terminated by signal SIGSEGV (Address boundary error)`

### Steps to Reproduce

1. Cargo.toml

```toml
[package]
name = "pyo3_test"
version = "0.1.0"
edition = "2021"

[dependencies]
pyo3 = { git = "https://github.com/PyO3/pyo3", features = ["extension-module"] }

[lib]
name = "pyo3_test"

crate-type = ["cdylib"]
```

2. Rust code:

```rust
use pyo3::types::PyFloat;
use pyo3::PyAny;
use pyo3::{pyclass, pymethods, pymodule, types::PyModule, PyResult, Python};

#[pyclass(extends=PyFloat)] // Or replace with other Py* struct
struct Number;

#[pymethods]
impl Number {
    #[new]
    #[pyo3(signature = (*_args, **_kwargs))]
    pub fn new(_args: &PyAny, _kwargs: Option<&PyAny>) -> Self {
        Self {}
    }
}

#[pymodule]
fn pyo3_test(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
    m.add_class::<Number>()?;
    Ok(())
}
```

3. Python code:

```python
from pyo3_test import Number

test = Number()

print(test)
```

4. Commands (inside venv)

```
$ maturin develop
$ python3 test.py
```

### Backtrace

```shell
For PyFloat for example:

fish: Job 1, 'python3 test.py' terminated by signal SIGSEGV (Address boundary error)

and for PySlice:

TypeError: type 'slice' is not an acceptable base type
thread '<unnamed>' panicked at 'An error occurred while initializing class Number', /home/<user>/.cargo/git/checkouts/pyo3-a22e69bc62b9f0fd/6757391/src/pyclass.rs:412:5
[...]

PR #2947, opened 2023-02-12
change PyModule::add_class to return an error if class creation fails

davidhewitt:
Related to #2942 

At the moment there are panics deep in the `#[pyclass]` machinery when initialising the type fails. This PR adjusts a number of these functions to return `PyResult` instead, so that we can handle the error more appropriately further down the pipeline.

For example, take the following snippet:

```rust
#[pyclass(extends = PyBool)]
struct ExtendsBool;

#[pymodule]
fn pyo3_scratch(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<ExtendsBool>()?;
    Ok(())
}
```

Currently, importing this module will fail with a panic:

```
TypeError: type 'bool' is not an acceptable base type
thread '<unnamed>' panicked at 'An error occurred while initializing class ExtendsBool', /Users/david/Dev/pyo3/src/pyclass.rs:412:5
  File "/Users/david/.virtualenvs/pyo3/lib/python3.10/site-packages/pyo3_scratch/__init__.py", line 1, in <module>
    from .pyo3_scratch import *
pyo3_runtime.PanicException: An error occurred while initializing class ExtendsBool
```

After this PR, this import still fails, but with a slightly cleaner, more Pythonic error:

```
TypeError: type 'bool' is not an acceptable base type

  File "/Users/david/.virtualenvs/pyo3/lib/python3.10/site-packages/pyo3_scratch/__init__.py", line 1, in <module>
    from .pyo3_scratch import *
RuntimeError: An error occurred while initializing class ExtendsBool
```

### DIFF
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -758,6 +758,7 @@ fn impl_pytypeinfo(
 
                 <#cls as _pyo3::impl_::pyclass::PyClassImpl>::lazy_type_object()
                     .get_or_init(py)
+                    .as_type_ptr()
             }
         }
     }
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -756,9 +756,8 @@ fn impl_pytypeinfo(
             fn type_object_raw(py: _pyo3::Python<'_>) -> *mut _pyo3::ffi::PyTypeObject {
                 #deprecations
 
-                use _pyo3::impl_::pyclass::LazyStaticType;
-                static TYPE_OBJECT: LazyStaticType = LazyStaticType::new();
-                TYPE_OBJECT.get_or_init::<Self>(py)
+                <#cls as _pyo3::impl_::pyclass::PyClassImpl>::lazy_type_object()
+                    .get_or_init(py)
             }
         }
     }
@@ -1038,6 +1037,12 @@ impl<'a> PyClassImplsBuilder<'a> {
                 #dict_offset
 
                 #weaklist_offset
+
+                fn lazy_type_object() -> &'static _pyo3::impl_::pyclass::LazyTypeObject<Self> {
+                    use _pyo3::impl_::pyclass::LazyTypeObject;
+                    static TYPE_OBJECT: LazyTypeObject<#cls> = LazyTypeObject::new();
+                    &TYPE_OBJECT
+                }
             }
 
             #[doc(hidden)]
--- a/src/impl_/pyclass/lazy_type_object.rs
+++ b/src/impl_/pyclass/lazy_type_object.rs
@@ -1,22 +1,25 @@
 use std::{
     borrow::Cow,
     ffi::CStr,
+    marker::PhantomData,
     thread::{self, ThreadId},
 };
 
 use parking_lot::{const_mutex, Mutex};
 
 use crate::{
-    ffi, once_cell::GILOnceCell, pyclass::create_type_object, IntoPyPointer, PyClass,
-    PyMethodDefType, PyObject, PyResult, Python,
+    exceptions::PyRuntimeError, ffi, once_cell::GILOnceCell, pyclass::create_type_object,
+    IntoPyPointer, PyClass, PyErr, PyMethodDefType, PyObject, PyResult, Python,
 };
 
 use super::PyClassItemsIter;
 
 /// Lazy type object for PyClass.
 #[doc(hidden)]
-pub struct LazyStaticType {
-    // Boxed because Python expects the type object to have a stable address.
+pub struct LazyTypeObject<T>(LazyTypeObjectInner, PhantomData<T>);
+
+// Non-generic inner of LazyTypeObject to keep code size down
+struct LazyTypeObjectInner {
     value: GILOnceCell<*mut ffi::PyTypeObject>,
     // Threads which have begun initialization of the `tp_dict`. Used for
     // reentrant initialization detection.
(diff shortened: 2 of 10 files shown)


src/impl_/pyclass/static_type_object.rs
```diff
@@ -24,31 +28,61 @@ pub struct LazyStaticType {
     tp_dict_filled: GILOnceCell<PyResult<()>>,
```
adamreichold, line 28: The detection of re-entrant initialization seem surprisingly heavy weight. What do you think about turning this around in the future, meaning to use e.g. a thread-local list of type object addresses which the current thread is currently initializing instead of a list of threads for each type object?

src/impl_/pyclass/static_type_object.rs
```diff
@@ -24,31 +28,61 @@ pub struct LazyStaticType {
...
-        fn inner<T: PyClass>() -> *mut ffi::PyTypeObject {
+    /// Gets the type object contained by this `StaticTypeObject`, initializing it if needed.
+    pub(crate) fn get_or_try_init(&self, py: Python<'_>) -> PyResult<*mut ffi::PyTypeObject> {
+        fn inner<T: PyClass>() -> PyResult<*mut ffi::PyTypeObject> {
             // Safety: `py` is held by the caller of `get_or_init`.
             let py = unsafe { Python::assume_gil_acquired() };
             create_type_object::<T>(py)
         }
 
-        // Uses explicit GILOnceCell::get_or_init::<fn() -> *mut ffi::PyTypeObject> monomorphization
```
adamreichold, line 45: I think this comment is still valid and helpful to understand why `inner` is used instead of a closure.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -1038,6 +1037,12 @@ impl<'a> PyClassImplsBuilder<'a> {
                 #dict_offset
 
                 #weaklist_offset
+
+                fn static_type_object() -> &'static _pyo3::impl_::pyclass::StaticTypeObject<Self> {
+                    use _pyo3::impl_::pyclass::*;
```
adamreichold, line 1042: Do we really need the glob import here? Why not just `StaticTypeObject` as in the example code from the guide?

tests/test_class_attributes.rs
```diff
@@ -113,6 +110,52 @@ fn test_fallible_class_attribute() {
     }
 
     Python::with_gil(|py| {
-        py.get_type::<BrokenClass>();
+        let stderr = CaptureStdErr::new(py).unwrap();
+        let _ = std::panic::catch_unwind(|| py.get_type::<BrokenClass>());
```
adamreichold, line 114: Couldn't at least assert directly that a panic did happen?

src/impl_/pyclass/static_type_object.rs
```diff
@@ -1,21 +1,25 @@
...
 use super::PyClassItemsIter;
 
 /// Lazy type object for PyClass.
 #[doc(hidden)]
-pub struct LazyStaticType {
+pub struct StaticTypeObject<T>(StaticTypeObjectInner, PhantomData<T>);
+
+// Generic inner of StaticTypeObject to keep code size down
+struct StaticTypeObjectInner {
     // Boxed because Python expects the type object to have a stable address.
```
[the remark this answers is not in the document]
davidhewitt: Actually this comment is waaaay out of date from back when we used static types. Now we use `PyType_FromSpec` (aka heap types) so the type object is already a pointer to a stable address. Will just delete the comment :)

pyo3-macros-backend/src/pyclass.rs
davidhewitt: Fair point, fixed. 👍

tests/test_class_attributes.rs
davidhewitt: Good idea 👍

src/impl_/pyclass/static_type_object.rs
davidhewitt: I think it's not something I'll rush to do because this code only affects program start-up time. It's a good idea and will be necessary if we ever try to replace these static things with type objects which live as data on the module. Created #2950

davidhewitt:
On reflection I've renamed `StaticTypeObject` to `LazyTypeObject` (was `LazyStaticType` before this PR). It's got nothing to do with "static" any more now that we use heap types, and the `&'static` reference is encoded elsewhere in the `PyClassImpl` implementation (but this would function perfectly fine without being static).

I've also pushed a follow-up commit which switches this to store `Py<PyType>` instead of `*mut ffi::PyTypeObject` to track ownership properly, and adjusts the internals to make use of this and `&PyType` so there's less `unsafe`.

src/impl_/pyclass/lazy_type_object.rs
```diff
@@ -43,39 +44,33 @@ impl<T> LazyTypeObject<T> {
...
 
 impl LazyTypeObjectInner {
-    fn get_or_try_init(
-        &self,
-        py: Python<'_>,
-        init: fn() -> PyResult<*mut ffi::PyTypeObject>,
+    fn get_or_try_init<'py>(
+        &'py self,
+        py: Python<'py>,
+        init: fn(Python<'py>) -> PyResult<Py<PyType>>,
```
adamreichold, line 65: Does this imply dynamic dispatch through a function pointer? Is that actually the right thing to do since this is only called during the initialization slow path?

adamreichold (approved):
I think this looks much better with the follow-up commit! Just one question where I am not sure about the performance characteristics.

src/impl_/pyclass/lazy_type_object.rs
davidhewitt: Yes, I think the cost of dynamic dispatch is unlikely to be relevant when it's only called once for each type? Or am I missing something here... ?

davidhewitt:
Thanks! I've adjusted the comment around the code you've questioned, replied, and will proceed to merge.

If we think there's follow-up work I'll open a new PR.

davidhewitt:
bors r=adamreichold

src/impl_/pyclass/lazy_type_object.rs
adamreichold: No I don't think so. To be honest, this is mainly me always getting confused about function pointers and items (the singleton type of each function) and traits in Rust.

davidhewitt:
bors retry

tests/test_class_attributes.rs
```diff
@@ -96,11 +96,10 @@ fn recursive_class_attributes() {
 }
 
 #[test]
-#[should_panic(
-    expected = "An error occurred while initializing `BrokenClass.fails_to_init`: \
-                ValueError: failed to create class attribute"
-)]
+#[cfg(panic = "unwind")]
```
adamreichold, line 99: I think this config flag was stabilized only in Rust 1.60.0: https://releases.rs/docs/1.60.0/#language

adamreichold: But maybe we can just move `CaptureStdError` into the test definition itself to avoid the unused warnings?

davidhewitt: Yes agreed, I'll do when I have a chance in the next day or two ...

adamreichold:
bors retry

tests/test_class_attributes.rs
adamreichold: Added a single fix-up and let bors do its thing.

davidhewitt: Many thanks!

Issue #2951, opened 2023-02-14
Segfault when cloning PyObjects in allow_threads


OliverBalfour:
### Bug Description

If you clone a `PyObject` inside an `allow_threads` block, it uses PyO3's internal refcount tracking (`gil::ReferencePool`) instead of immediately incrementing the reference count. However, the refcount isn't updated until the next `GILPool::new` call. This means if all previous references to the object are dropped before the next explicit call to `Python::with_gil` or similar, the object can be deleted despite the clone from `allow_threads` still existing. Using this reference can cause use-after-free segfaults or give you garbage results.

I've reproduced this on Mac and Linux.

### Steps to Reproduce

Here is a failing test case that reproduces the issue:
```rs
    // Put this code in the `tests` module in `gil.rs`

    fn get_lambda(py: Python<'_>) -> PyObject {
        let pool = unsafe { py.new_pool() };
        let py = pool.python();
        let obj = py.eval("lambda: None", None, None).unwrap();
        obj.to_object(py)
    }

    #[test]
    fn test_allow_threads_does_not_delete_too_soon() {
        Python::with_gil(|py| {
            // Make a simple function with 1 reference
            let obj = get_lambda(py);
            assert!(obj.get_refcnt(py) == 1);

            // Clone the object without the GIL to use internal tracking
            let escaped_ref = py.allow_threads(|| obj.clone());
            println!("refcnt: {} (should be 2)", obj.get_refcnt(py));
            // Refcount isn't incremented at the end of `allow_threads`, only in the next `GILPool::new` call.

            // This means the refcount is still 1, and will be 2 later on.
            // But when escaped_ref is dropped, the refcount temporarily dips to 0 and it's deleted
            drop(escaped_ref);

            // Now `obj` is a dangling reference and calling it segfaults
            println!("{}", obj.call0(py).unwrap());
            // I think we need to call  `ReferencePool::update_counts` at the end of `allow_threads`
        });
    }
```

### Backtrace

```shell
running 1 test
refcnt: 1 (should be 2)
error: test failed, to rerun pass `-p pyo3 --lib`

Caused by:
  process didn't exit successfully: `/path/to/pyo3/target/debug/deps/pyo3-cbbebd45802d0ee8 'gil::tests::test_allow_threads_does_not_delete_too_soon' --exact --nocapture` (signal: 11, SIGSEGV: invalid memory reference)
```

### Your operating system and version

MacOS 13.1

### Your Python version (`python --version`)

3.9.16

[...]

PR #2952, opened 2023-02-14
Fix allow_threads segfault

OliverBalfour:
Please see the corresponding issue **#2951** for details. This PR adds the failing test from the issue and then a fix for it. The fix simply calls `ReferencePool::update_counts` at the end of `allow_threads` to ensure objects aren't accidentally deleted too soon.

### DIFF
--- a/src/marker.rs
+++ b/src/marker.rs
@@ -478,6 +478,8 @@ impl<'py> Python<'py> {
                 gil::GIL_COUNT.with(|c| c.set(self.count));
                 unsafe {
                     ffi::PyEval_RestoreThread(self.tstate);
+                    // Update counts of PyObjects / Py that were cloned or dropped during `f`.
+                    gil::POOL.update_counts(Python::assume_gil_acquired());
                 }
             }
         }
--- a/newsfragments/2952.fixed.md
+++ b/newsfragments/2952.fixed.md
@@ -1,0 +1,1 @@
+Fix a reference counting race condition affecting `PyObject`s cloned in `allow_threads` blocks.
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -277,7 +277,7 @@ impl Drop for GILGuard {
 type PyObjVec = Vec<NonNull<ffi::PyObject>>;
 
 /// Thread-safe storage for objects which were inc_ref / dec_ref while the GIL was not held.
-struct ReferencePool {
+pub(crate) struct ReferencePool {
     dirty: atomic::AtomicBool,
     // .0 is INCREFs, .1 is DECREFs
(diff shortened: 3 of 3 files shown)


davidhewitt:
Thank you very much for debugging this, a clear explanation and I agree with the proposed fix! Let's get this out as a patch release.

Please can you add a newsfragment as `2952.fixed.md`.

The test failure looks like a race. I think what is likely happening is that the new test you have added is writing an object into the pool during the `allow_threads` call, which then races with `test_pyobject_drop_without_gil_doesnt_decrease_refcnt`.

I think the `pool_not_dirty()` assertion in that test on line 635 is flawed. I think you can change it instead just check that the object in question is not in the pool:

```rust
assert!(!POOL.pointer_ops.lock().0.contains(&obj.as_ptr()));
assert!(!POOL.pointer_ops.lock().1.contains(&obj.as_ptr()));
```

Finally, can you please update the test `test_clone_without_gil`. On line 803 it has this comment:

```rust
// Returning from allow_threads doesn't clear the pool
```

That comment is now clearly wrong, so I think the `allow_threads` / `with_gil` construct immediately after it can be removed.

src/marker.rs
```diff
@@ -478,6 +478,8 @@ impl<'py> Python<'py> {
                 gil::GIL_COUNT.with(|c| c.set(self.count));
                 unsafe {
                     ffi::PyEval_RestoreThread(self.tstate);
+                    // Update counts of PyObjects / Py that were cloned or dropped during `f`.
+                    gil::POOL.update_counts(Python::assume_gil_acquired());
```
adamreichold, line 482: I wonder whether we should package up `RestoreGuard` in `mod gil` to centralize the usage of `Save/RestoreThread`, accessing `GIL_COUNT` and calling `update_counts` in that module. (I suspect that having this in separate modules, i.e. making `GIL_COUNT` a `pub(crate)` item, might have been the root cause of this slipping in.)

OliverBalfour:
Thanks for the feedback. I've added a newsfragment and made the proposed test modifications. I see there are a couple other `pool_not_dirty` assertions - are these safe? I don't know the codebase well.

As for packaging up `RestoreGuard`, this sounds like a good solution! I personally won't have time to do this though, I'm very time poor.

davidhewitt:
Thanks! I'm going to force-push here to squash. The `pool_not_dirty()` assertions where the GIL is never released are not a problem (won't race with other threads).

@adamreichold  - I agree with the idea of adding an abstraction layer, however it's fine IMO to not do it in this fix. I've opened #2969 for one of us to remember to do this later.

PR #2953, opened 2023-02-14
set csv minimal package version for 1.48 CI

davidhewitt:
New `csv 1.2.0` requires Rust 1.60.


davidhewitt:
Thanks @messense and @adamreichold, this new process is so much easier!

bors r+

PR #2954, opened 2023-02-15
optimize mapping conversion for dict

davidhewitt:
Equivalent of #2944 for dicts -> mapping.

The benchmark diff is not as significant as in #2944. Still something like 80ns -> 2ns on my machine.

### DIFF
--- a/src/types/mapping.rs
+++ b/src/types/mapping.rs
@@ -3,12 +3,8 @@
 use crate::err::{PyDowncastError, PyErr, PyResult};
 use crate::once_cell::GILOnceCell;
 use crate::type_object::PyTypeInfo;
-use crate::types::{PyAny, PySequence, PyType};
-use crate::{
-    ffi, AsPyPointer, IntoPy, IntoPyPointer, Py, PyNativeType, PyTryFrom, Python, ToPyObject,
-};
-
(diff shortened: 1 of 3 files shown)


src/types/mapping.rs
```diff
@@ -139,11 +132,14 @@ impl<'v> PyTryFrom<'v> for PyMapping {
     fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v PyMapping, PyDowncastError<'v>> {
         let value = value.into();
 
-        // TODO: surface specific errors in this chain to the user
-        if let Ok(abc) = get_mapping_abc(value.py()) {
-            if value.is_instance(abc).unwrap_or(false) {
-                unsafe { return Ok(value.downcast_unchecked()) }
-            }
+        // Using `is_instance` for `collections.abc.Mapping` is slow, so provide
+        // optimized case dict as a well-known mapping
```
birkenfeld, line 136: What about the TODO comment that was here before?

davidhewitt: Sure, I'll restore it (and the same to #2944).

davidhewitt:
To reduce the review burden, I'm going to proceed to merge this and #2944. I think both are not controversial.

bors r+

PR #2956, opened 2023-02-15
Move PyTypeInfo::AsRefTarget onto new trait HasPyGilBoundRef

davidhewitt:
NB This is a breaking change, would be part of 0.19, not any 0.18.x patch release. Despite this, I would generally assume users don't ever refer to this directly, so I don't expect much impact downstream.

This PR moves `PyTypeInfo::AsRefTarget` onto its own trait. There's a couple of reasons for this:
- `PyIterator`, `PySequence` and `PyMapping` can implement the new trait (they can't implement `PyTypeInfo`). This removes the need for them to provide special-cased methods.
- While looking at rebooting #1308, I found that this is the only item in `PyTypeInfo` which the experimental types cannot implement. So if we break this up, I think it will simplify that implementation.

(Maybe we can wait until I push the rebooted #1308 proposal, to see whether we actually like it enough to try to adopt it.)

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -7,7 +7,7 @@ use crate::pycell::{PyBorrowError, PyBorrowMutError, PyCell};
 use crate::types::{PyDict, PyString, PyTuple};
 use crate::{
     ffi, AsPyPointer, FromPyObject, IntoPy, IntoPyPointer, PyAny, PyClass, PyClassInitializer,
-    PyRef, PyRefMut, PyTypeInfo, Python, ToPyObject,
+    PyRef, PyRefMut, Python, ToPyObject,
 };
 use std::marker::PhantomData;
 use std::mem;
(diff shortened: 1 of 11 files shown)


src/instance.rs
```diff
@@ -39,6 +39,16 @@ pub unsafe trait PyNativeType: Sized {
     }
 }
 
+/// Trait to denote types which have a GIL-bound reference, e.g. PyAny has &'a PyAny.
+///
+/// # Safety
+/// `Py<T>` will hand out references to `Self::AsRefTarget`. This _must_ have the
+/// same layout as `*mut ffi::PyObject`.
+pub unsafe trait HasPyGilBoundRef {
```
mejrs, line 47: IMO the name could use some work. I don't have a name that is obviously good, but what about something like `AsGilBorrow`?

src/instance.rs
```diff
@@ -39,6 +39,16 @@ pub unsafe trait PyNativeType: Sized {
     }
 }
 
+/// Trait to denote types which have a GIL-bound reference, e.g. PyAny has &'a PyAny.
+///
+/// # Safety
+/// `Py<T>` will hand out references to `Self::AsRefTarget`. This _must_ have the
+/// same layout as `*mut ffi::PyObject`.
```
mejrs, line 46: Isn't the layout part guaranteed by `PyNativeType`? Unless "this" refers to `Self` rather than the target.

davidhewitt: Yes, I think so. Probably still worth repeating this constraint here?

src/instance.rs
davidhewitt: Sure thing. I think we can sit on this and think for a bit. I'm a bit lost for a good name too. `StorableInPy`?

adamreichold: Just add completely different direction for the bikeshed: `PyHeapType` meaning that can be stored on the Python heap (i.e. in `Py<T>`) and thereby shared references must be bound to the GIL (as it protects access to the heap).

src/instance.rs
mejrs: Sure 👍

davidhewitt:
It looks like we don't need this / will remove it with #3382 so I'll close this.

PR #2964, opened 2023-02-19
Update `README.md` by adding datafusion, ballista and delta-rs

iajoiner:
Add Python bindings of datafusion, ballista and the official Rust implementation of Delta Lake (delta-rs). The former two are Apache projects under Arrow and the last is from Delta Lake itself so all three projects are notable and need to be added as examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,9 @@ about this topic.
 
 ## Examples
 
+- [DataFusion in Python](https://github.com/apache/arrow-datafusion-python) _A very fast, extensible query engine for building high-quality data-centric systems in Rust, using the Apache Arrow in-memory format with bindings into Python_.
+- [Ballista in Python](https://github.com/apache/arrow-ballista-python) _Python bindings to Apache Arrow Ballista Distributed Query Engine_.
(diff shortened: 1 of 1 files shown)


kngwyu:
Thanks, maybe it's better to use the same notation as https://github.com/PyO3/maturin#examples 
(Maybe we can have the same showcase page as maturin and PyO3? But it would be a future job...

adamreichold:
> Thanks, maybe it's better to use the same notation as https://github.com/PyO3/maturin#examples (Maybe we can have the same showcase page as maturin and PyO3? But it would be a future job...

I think that we should definitely switch to an alphabetically sorted list to avoid examples being added to the front. @iajoiner Could you adjust the PR in that manner?

iajoiner:
@adamreichold @kngwyu Sure!

iajoiner:
@kngwyu Description has been changed to match maturin.
@adamreichold All examples have been sorted alphabetically.

Please review.

davidhewitt:
This is great, thanks very much for the additions & tidy ups!

PR #2966, opened 2023-02-19
Correct documentation for `Py::borrow_mut()`.

### LINKED ISSUE #2958: Little docs mistake
### Bug Description

[Docs for `Py::borrow_mut()`](https://docs.rs/pyo3/latest/pyo3/prelude/struct.Py.html#method.borrow_mut) say:

> Panics if the value is **currently mutably borrowed.** For a non-panicking variant, use [`try_borrow_mut`](https://docs.rs/pyo3/latest/pyo3/prelude/struct.Py.html#method.try_borrow_mut).

But the docs for the similar linked method say:

> returning an error if the value is **currently borrowed.**

(Emphasis mine.)

Seems "mutably" should be removed from the first quote.

Enyium:
Closes #2958.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -429,7 +429,7 @@ where
     ///  ```
     ///
     /// # Panics
-    /// Panics if the value is currently mutably borrowed. For a non-panicking variant, use
+    /// Panics if the value is currently borrowed. For a non-panicking variant, use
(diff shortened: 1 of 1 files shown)


adamreichold:
Thanks again!

bors r+

PR #2975, opened 2023-02-21
RFC: Add GILProtected mutex replacement and use it for LazyTypeObjectInner.

adamreichold:
I would also like to use that type in rust-numpy and it seems we can avoid both a manual unsafe impl and a full blown mutex if we apply it to `LazyTypeObjectInner`.

One downside might be that it ties us closer to the GIL when we want to enable nogil experimentation, but on the other hand, it may also help by reifying the GIL usage. (This is currently limited to comments in unsafe code in rust-numpy for example.)


src/impl_/pyclass/lazy_type_object.rs
```diff
@@ -170,7 +171,7 @@ impl LazyTypeObjectInner {
             // Initialization successfully complete, can clear the thread list.
             // (No further calls to get_or_init() will try to init, on any thread.)
             std::mem::forget(guard);
-            *self.initializing_threads.lock() = Vec::new();
+            *self.initializing_threads.with_gil(py).borrow_mut() = Vec::new();
```
kngwyu, line 174: Maybe you can use `RefCell::replace`?

mejrs:
I think this PR title and the documentation are deceiving. This isn't a mutex - it's just a wrapper for making a !Sync type implement Sync.

adamreichold:
> .

But isn't this what a mutex does, i.e. https://doc.rust-lang.org/stable/std/sync/struct.Mutex.html#impl-Sync-for-Mutex%3CT%3E ?

The mechanism is different of course, but as with std's `Mutex`, you can only access the inner value while holding locking. Acquiring the lock and accessing the value is just more disconnected than with the std's `Mutex` but still reified in the type system via the `Python` token.

adamreichold:
One thing that is still unsound though, is that I cannot return a bare reference, i.e. `with_gil` must return a `!Send` reference wrapper so that I cannot e.g. pass that reference into an `allow_threads` closure.

adamreichold:
Meaning that I do not need a reference wrapper after all.

Maybe another way to think about this is that `GILProtected` is basically the same thing we do at each `#[pyfunction]` entry when we provide shared references to values stored on the Python heap by ensuring via the GIL that only one threads accesses them at any time, just as a free-standing type and without handling how to get hold of the `Python` token in the first place.

adamreichold:
> Also, mutex means "mutually exclusive". Also, mutex means "mutually exclusive". This does not give you something exclusive, because you can call it multiple times as long as you are holding the gil. 

Well, it is mutually exclusive insofar as only one thread can hold the GIL, so one might call it "recursive mutex" in that schema.

> Calling it a mutex is just going to confuse people when they try swapping out their mutex but can't get a &mut T out of it. Something like GilWrapper would be a better name.

That said, I am open to give this a different name and/or removing the reference to `Mutex` from the comments. `GilWrapper` does not seem helpful IMHO though as it does not indicate what the wrapping achieves.

What about changing the doc comment to

```rust
/// This is a synchronization primitive based on Python's global interpreter lock (GIL),
/// it ensure that only one thread at a time can access the inner value via shared references,
/// but it usually needs to be combined with interior mutability to obtain mutable references.
```

?

src/gil.rs
```diff
@@ -494,6 +494,31 @@ impl EnsureGIL {
...
+        Self { value }
+    }
+
+    /// Gain access to the inner value by giving proof of having acquired the GIL.
+    pub fn with_gil<'py>(&'py self, _py: Python<'py>) -> &'py T {
+        &self.value
+    }
+}
+
+unsafe impl<T> Send for GILProtected<T> where T: Send {}
```
davidhewitt, line 518: I don't think this implementation is needed, the only field is `T` so I would assume that auto impl will cover this.

src/gil.rs
```diff
@@ -494,6 +494,31 @@ impl EnsureGIL {
...
+
+unsafe impl<T> Sync for GILProtected<T> where T: Send {}
```
davidhewitt, line 520: I'm wondering if `T: Send` requirement can be dropped here? As I see it, `GILProtected<T>` offers an API which always is thread-safe, `T` is irrelevant.

davidhewitt (commented):
This seems like a very reasonable idea to me. I like the way that it gives us an API which we can encourage downstream code to unify around!

cc @colesbury - might be interesting to see this RE nogil context

davidhewitt:
Should this perhaps live in a `pyo3::sync` module along with `GILOnceCell`?

src/gil.rs
adamreichold: Certainly, this was mindless cargo-culting from `Mutex`.

src/gil.rs
adamreichold: If I clone an `Rc<()>` on thread A, put it inside a `GILProtected<T>` and send `&GILProtected<Rc<()>>` to thread B, then on thread B, I can do `.with_gil(py).clone()` to produce an `Rc<()>` thereby racily increment the reference count against thread A with never had the GIL in the first place?

I have to admit to have a hard time phrasing this more generally, but I have the same problem with the equivalent bound for `Mutex`. But just because of the counter example if think the bound is required.

adamreichold: Another example would GUI objects bound to the main thread where even calling methods through shared references on them on other threads would not work, independently of whether those calls are synchronized, i.e. exclusively interleave, using the GIL.

davidhewitt: Right, understood. Those counter-examples are clear enough to me, thanks.

adamreichold: I think @birkenfeld had some fitting general terms for this: https://users.rust-lang.org/t/solved-how-to-move-non-send-between-threads-or-an-alternative/19928/9

mejrs: Yes, this bound is required. If `GILProtected` is `Sync` then `&GILProtected` can be sent to another thread, and sending `&GILProtected<Rc<T>>` to another thread is unsound.

adamreichold:
> Should this perhaps live in a `pyo3::sync` module along with `GILOnceCell`?

Yes, will move it there. (I managed to edit the original comment instead of replying to it. Sorry for that.)

src/sync.rs
```diff
@@ -1,7 +1,30 @@
...
+}
+
+impl<T> GILProtected<T> {
+    /// Place the given value under the protection of the GIL.
+    pub const fn new(value: T) -> Self {
+        Self { value }
+    }
+
+    /// Gain access to the inner value by giving proof of having acquired the GIL.
+    pub fn with_gil<'py>(&'py self, _py: Python<'py>) -> &'py T {
```
mejrs, line 21: I'd rather just call this `get`

src/sync.rs
```diff
@@ -1,7 +1,30 @@
...
+//! Synchronization mechanisms based on the Python GIL.
 use crate::{types::PyString, Py, Python};
 use std::cell::UnsafeCell;
 
+/// Value with concurrent access protected by the GIL.
+///
+/// This is a synchronization primitive based on Python's global interpreter lock (GIL).
+/// It ensures that only one thread at a time can access the inner value via shared references.
+/// It can be combined with interior mutability to obtain mutable references.
+pub struct GILProtected<T> {
```
mejrs, line 10: This should have a doc example. Something with a `static `GILProtected<RefCell<T>>`, I'd imagine.

src/lib.rs
```diff
@@ -409,7 +409,7 @@ mod instance;
 pub mod marker;
 pub mod marshal;
 #[macro_use]
-pub mod once_cell;
+pub mod sync;
```
mejrs, line 412: Can we add a deprecation for this? This will be much nicer than a compile error.

mejrs (commented):
Overall I like this, I have some nits though.

src/lib.rs
adamreichold: Do you think a `pub use sync as once_cell;` would suffice?

adamreichold: (With the deprecation of course, I just meant instead of keeping both modules around and having the re-exports in there.)

mejrs: I'm not sure. Is there actually a way to deprecate this?

src/sync.rs
adamreichold: Added an example. Do you think this is sufficient or would should it be more involved, e.g. using multiple threads or asserting on the result of the modification?

src/lib.rs
adamreichold: Hhhmm, it seems to work neither way, i.e. both marking the `pub use sync as once_cell;` and the `pub use crate::sync::GILOnceCell` inside an otherwise empty `mod once_cell;` does not produce a deprecation warning for usages of `pyo3::once_cell:GILOnceCell`.

Does anybody have an idea of how to keep the old path alive but deprecated without copy & pasting the code?

davidhewitt: Does having `#[deprecated] mod once_cell { pub use crate::sync::GILOnceCell }` work?

If not, I suggest keep both around for backwards compatibility, there's a rustc issue _somewhere_ on deprecating exports which I've been tempted to pick up sometime so that we can clean up some of our old module hierarchy :smile:

adamreichold: > Does having #[deprecated] mod once_cell { pub use crate::sync::GILOnceCell } work?

No, it did not seem to work. I suspect because re-exports are not "items" and only those can be deprecated?

> If not, I suggest keep both around for backwards compatibility

This is a heavy burden though. What if make a change to one copy but forget about the other?

davidhewitt: I mean just have `pub mod once_cell { pub use crate::sync::GILOnceCell }` as an item in `lib.rs`, with a "TODO deprecate this" just above?

adamreichold: > I mean just have pub mod once_cell { pub use crate::sync::GILOnceCell } as an item in lib.rs, with a "TODO deprecate this" just above?

This I can certainly do.

adamreichold: I think the relevant issue is https://github.com/rust-lang/rust/issues/30827 which I linked in the FIXME comment.

src/lib.rs
```diff
@@ -420,6 +420,14 @@ pub mod type_object;
...
 mod version;
 
+#[deprecated(since = "0.19.0")]
+/// A write-once cell mediated by the Python GIL.
+// FIXME: We want to deprecate this, but that does not yet work for re-exports,
+// c.f. https://github.com/rust-lang/rust/issues/30827
+pub mod once_cell {
+    pub use crate::sync::{GILOnceCell, Interned};
+}
+
```
mejrs, line 430: Maybe this should be `doc(hidden)`? It might confusing having both show up in docs.

adamreichold: Yes, I agree. I also added the missing deprecation note if this suddenly becomes effective via a rustc change.

src/sync.rs
```diff
@@ -1,7 +1,46 @@
...
 
+/// Value with concurrent access protected by the GIL.
+///
+/// This is a synchronization primitive based on Python's global interpreter lock (GIL).
+/// It ensures that only one thread at a time can access the inner value via shared references.
+/// It can be combined with interior mutability to obtain mutable references.
+///
+/// # Example
+///
+/// Combinding `GILProtected` with `RefCell` enables mutable access to static data:
```
davidhewitt, line 13:
```suggestion
/// Combining `GILProtected` with `RefCell` enables mutable access to static data:
```

davidhewitt (commented):
I'm thinking this would be nice to include in 0.18.2?

src/sync.rs
mejrs: Lovely, thanks.

davidhewitt:
> > I'm thinking this would be nice to include in 0.18.2?
> 
> Isn't replacing an item with a re-export technically a breaking change? (Note to self, we should probably setup cargo-semver-checks...)

Is there an observable compile issue that can be introduced by moving the item and re-exporting from the old location? I would hope that this would be unnoticeable from a user protective 🙂

adamreichold:
> > > I'm thinking this would be nice to include in 0.18.2?
> > 
> > 
> > Isn't replacing an item with a re-export technically a breaking change? (Note to self, we should probably setup cargo-semver-checks...)
> 
> Is there an observable compile issue that can be introduced by moving the item and re-exporting from the old location? I would hope that this would be unnoticeable from a user protective slightly_smiling_face

I was thinking of https://predr.ag/blog/moving-and-reexporting-rust-type-can-be-major-breaking-change/ but that seems to apply only to using `pub type` instead of `pub use` for re-exporting. So it should not be a problem in this case here.

davidhewitt:
👍 let's ship it!

bors r+

PR #2980, opened 2023-02-22
support `text_signature` on `#[new]`

### LINKED ISSUE #2866: Support `text_signature` and autogenerated text signature for `#[new]`

davidhewitt:
Closes #2866 

This is a breaking change for 0.19.0, because it starts autogenerating `text_signature` for `#[new]`. This could affect runtime behaviour if the user is relying on the class docs at runtime for some reason.

Guide & tests all updated accordingly.

`#[pyclass(text_signature = "...")]` is deprecated by this PR, however if it's set, it will be used in preference to `#[new]`.

(The signature / `text_signature` from `#[new]` will simply be ignored in this case. I figure that when users fix their deprecation warnings by removing `#[pyclass(text_signature = "...")]` then the `#[new]` signatures will start flowing properly, and this is good enough.)

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -722,22 +722,20 @@ py_args=(), py_kwargs=None, name=World, num=-1, num_before=44
 
 ## Making class method signatures available to Python
 
-The [`text_signature = "..."`](./function.md#text_signature) option for `#[pyfunction]` also works for classes and methods:
+The [`text_signature = "..."`](./function.md#text_signature) option for `#[pyfunction]` also works for `#[pymethods]`:
 
 ```rust
 # #![allow(dead_code)]
 use pyo3::prelude::*;
 use pyo3::types::PyType;
 
-// it works even if the item is not documented:
-#[pyclass(text_signature = "(c, d, /)")]
+#[pyclass]
 struct MyClass {}
 
 #[pymethods]
 impl MyClass {
-    // the signature for the constructor is attached
-    // to the struct definition instead.
     #[new]
+    #[pyo3(text_signature = "(c, d)")]
     fn new(c: i32, d: &str) -> Self {
         Self {}
     }
@@ -746,8 +744,9 @@ impl MyClass {
     fn my_method(&self, e: i32, f: i32) -> i32 {
         e + f
     }
+    // similarly for classmethod arguments, use $cls
     #[classmethod]
-    #[pyo3(text_signature = "(cls, e, f)")]
+    #[pyo3(text_signature = "($cls, e, f)")]
     fn my_class_method(cls: &PyType, e: i32, f: i32) -> i32 {
         e + f
     }
@@ -773,7 +772,7 @@ impl MyClass {
 #                 .call1((class,))?
 #                 .call_method0("__str__")?
 #                 .extract()?;
-#             assert_eq!(sig, "(c, d, /)");
+#             assert_eq!(sig, "(c, d)");
 #         } else {
 #             let doc: String = class.getattr("__doc__")?.extract()?;
 #             assert_eq!(doc, "");
@@ -802,7 +801,7 @@ impl MyClass {
 #                 .call1((method,))?
 #                 .call_method0("__str__")?
 #                 .extract()?;
-#             assert_eq!(sig, "(cls, e, f)");
+#             assert_eq!(sig, "(e, f)");  // inspect.signature skips the $cls arg
 #         }
 #
 #         {
@@ -822,7 +821,7 @@ impl MyClass {
 # }
 ```
 
-Note that `text_signature` on classes is not compatible with compilation in
+Note that `text_signature` on `#[new]` is not compatible with compilation in
 `abi3` mode until Python 3.10 or greater.
 
 ## #[pyclass] enums
@@ -1018,7 +1017,6 @@ impl pyo3::IntoPy<PyObject> for MyClass {
 }
 
 impl pyo3::impl_::pyclass::PyClassImpl for MyClass {
-    const DOC: &'static str = "Class for demonstration\u{0}";
     const IS_BASETYPE: bool = false;
     const IS_SUBCLASS: bool = false;
     type Layout = PyCell<MyClass>;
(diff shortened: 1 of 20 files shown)


pyo3-macros-backend/src/pyclass.rs
```diff
@@ -191,11 +195,7 @@ pub fn build_py_class(
     methods_type: PyClassMethodsType,
 ) -> syn::Result<TokenStream> {
     args.options.take_pyo3_options(&mut class.attrs)?;
-    let text_signature_string = text_signature_or_none(args.options.text_signature.as_ref());
```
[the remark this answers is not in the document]
adamreichold: Sorry to answer myself, but this seems to have moved to `fn new`. Does this imply that the text processing happens at runtime now?

guide/src/class.md
```diff
@@ -1039,6 +1038,15 @@ impl pyo3::impl_::pyclass::PyClassImpl for MyClass {
         static TYPE_OBJECT: LazyTypeObject<MyClass> = LazyTypeObject::new();
         &TYPE_OBJECT
     }
+
+    fn doc(py: Python<'_>) -> pyo3::PyResult<&'static ::std::ffi::CStr> {
+        use pyo3::impl_::pyclass::*;
+        static DOC: pyo3::once_cell::GILOnceCell<::std::borrow::Cow<'static, ::std::ffi::CStr>> = pyo3::once_cell::GILOnceCell::new();
+        DOC.get_or_try_init(py, || {
+            let collector = PyClassImplCollector::<Self>::new();
+            build_pyclass_doc(<MyClass as pyo3::PyTypeInfo>::NAME, "", None.or_else(|| collector.new_text_signature()))
```
adamreichold, line 1047: Isn't `None.or_else(|| foo())` equivalent to `foo()`?

adamreichold: Ah, this is expanded generated code. I think I should take a break for today...

davidhewitt: Yes, the `None` represents the deprecated `#[pyclass(text_signature = "(...)")]` pathway.

pyo3-macros-backend/src/pyclass.rs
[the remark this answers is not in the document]
davidhewitt: There's a test to make sure that if `#[pyclass(text_signature = "...")]` is set, it's still used in preference to anything from `#[new]`, for backwards compatibility. Updated the PR description to explain this, sorry for lacking detail.

Yes, the text processing now necessarily happens at runtime for `#[new]. The `#[pyclass(name = "...")]` is only known by `#[pyclass]`, similarly for the doc. However the text_signature will now be coming from `#[pymethods]` (using dtolnay specialization).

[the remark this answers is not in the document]
adamreichold: One thing I do not yet understand is why the specialization is required. Can't we just always generate an appropriate trait impl for each `#[pyclass]`? Is this for painlessly handling classes with out a `__new__` method at all?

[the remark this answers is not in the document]
davidhewitt: Yes, I think it's a valid use case to make a pyclass which can't be constructed from Python, so that's why I use the specialization to permit the option.

davidhewitt:
@adamreichold I've just rebased this, and I think (assuming CI passes) it might be ready to merge.

src/pyclass/create_type_object.rs
```diff
@@ -22,7 +22,7 @@ where
 {
     unsafe {
         PyTypeBuilder::default()
-            .type_doc(T::DOC)
+            .type_doc(T::doc(py).unwrap()) // FIXME remove unwrap
```
adamreichold, line 25: Why not just `?` as `create_type_object` by now returns `PyResult<..>`? Is an early return problematic inside the unsafe block?

adamreichold (approved):
Single nit. Please feel free to merge after resolving this.

src/pyclass/create_type_object.rs
davidhewitt: Ah yes, thanks, missed this was now fixable post-rebase!

davidhewitt:
bors r=adamreichold

adamreichold:
bors retry

adamreichold:
The nightly warning on `non_upper_case_globals` seems unrelated but also trivial to fix by adding an `#[allow(..)]` to the generated code so I'd say we might as well do it here?

The test failure in `test_auto_test_signature_method` seems real though. During review, I assumed that this would be the automatically generated Python signature based on the actual Rust signature? Does that not work on Python 3.7? (It would be very helpful to have a `py_assert_eq` to see the actual value in the logs.)

adamreichold:
> The nightly warning on non_upper_case_globals seems unrelated but also trivial to fix by adding an #[allow(..)] to the generated code so I'd say we might as well do it here?

Actually, I cannot reproduce this on `main`. Maybe this does need a rebuild? Probably useful anyway after #3114 lands.

davidhewitt:
Rebased; let's hope things are working again now. Sorry for the gap!

bors r=adamreichold

adamreichold:
I think

> The test failure in test_auto_test_signature_method seems real though. During review, I assumed that this would be the automatically generated Python signature based on the actual Rust signature? Does that not work on Python 3.7? (It would be very helpful to have a py_assert_eq to see the actual value in the logs.)

still applies.

adamreichold:
> Sadly, I cannot reproduce the issue locally even using Python 3.7, neither via pyenv (which is 3.7.16 as in GitHub actions) nor using indygreg's standalone builds which stopped at 3.7.9.

Sorry, I am an idiot for trying this on the `main` branch...

adamreichold:
I think this is due to us not being able to set `__text_signature__` on class using the limited API before version 3.10. I hence took the liberty of pushing a commit ignoring that one assertion as we already do for the `class_with_signature_no_doc` and `class_with_docs_and_signature` test cases.

bors r+

adamreichold:
bors retry

adamreichold:
bors retry

PR #2981, opened 2023-02-22
Remove 0.17 deprecations

davidhewitt:
Since #2980 starts a breaking change for 0.19, let's also clean up all 0.17's deprecations.

I've removed `Python::acquire_gil` in its own commit, as that was a reasonably chunky removal.

### DIFF
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -40,6 +40,78 @@ After, the same code will print `ValueError: original error message`, which is m
 
 However, if the `anyhow::Error` or `eyre::Report` has a source, then the original exception will still be wrapped in a `PyRuntimeError`.
 
+### The deprecated `Python::acquire_gil` was removed and `Python::with_gil` must be used instead
+
+While the API provided by [`Python::acquire_gil`](https://docs.rs/pyo3/0.18.3/pyo3/marker/struct.Python.html#method.acquire_gil) seems convenient, it is somewhat brittle as the design of the GIL token [`Python`](https://docs.rs/pyo3/0.18.3/pyo3/marker/struct.Python.html) relies on proper nesting and panics if not used correctly, e.g.
+
+```rust,should_panic
+# #![allow(dead_code, deprecated)]
+# use pyo3::prelude::*;
+
+#[pyclass]
+struct SomeClass {}
+
+struct ObjectAndGuard {
+    object: Py<SomeClass>,
+    guard: GILGuard,
+}
+
+impl ObjectAndGuard {
+    fn new() -> Self {
+        let guard = Python::acquire_gil();
+        let object = Py::new(guard.python(), SomeClass {}).unwrap();
+
+        Self { object, guard }
+    }
+}
+
+let first = ObjectAndGuard::new();
+let second = ObjectAndGuard::new();
+// Panics because the guard within `second` is still alive.
+drop(first);
+drop(second);
+```
+
+The replacement is [`Python::with_gil`]() which is more cumbersome but enforces the proper nesting by design, e.g.
+
+```rust
+# #![allow(dead_code)]
+# use pyo3::prelude::*;
+
+#[pyclass]
+struct SomeClass {}
+
+struct Object {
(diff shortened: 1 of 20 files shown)


src/gil.rs
```diff
@@ -139,25 +135,9 @@ where
...
-/// use pyo3::Python;
-///
-/// {
-///     #[allow(deprecated)]
-///     let gil_guard = Python::acquire_gil();
-///     let py = gil_guard.python();
-/// } // GIL is released when gil_guard is dropped
-/// ```
-#[must_use]
 pub struct GILGuard {
```
adamreichold, line 138: Does `GILGuard` need to stay `pub` at all? Would `pub(crate)` or even private suffice if we use `EnsureGIL` elsewhere?

davidhewitt: Ah yes I nearly went for private and then forgot at the last second, will push... 😄

mejrs (approved):
I was thinking about keeping `acquire_gil` around for a while longer, because I think a lot of people still use it, and maybe use it in ways that aren't trivial to rewrite using `with_gil`. That said I also don't like having both around, so 👍 to this.

newsfragments/2981.removed.md
```diff
@@ -0,0 +1 @@
+Remove all functionality deprecated in PyO3 0.17.
```
mejrs, line 1: This should mention `acquire_gil` by name, its removal should hit a lot of people upgrading from older pyo3 versions and skipping versions.

adamreichold:
> I was thinking about keeping `acquire_gil` around for a while longer, because I think a lot of people still use it, and maybe use it in ways that aren't trivial to rewrite using `with_gil`. That said I also don't like having both around, so +1 to this.

I think this would be worth it considering how much tricky/unsafe code can be removed or get reduced visibility due to this, especially the drop order checking which could inject unexpected and intermittent panics into downstream code.

adamreichold:
@davidhewitt I would like to help here if I can as I would very much like to see this merged for the reduction in complexity it allows. Would you mind if I tried to rebase this and add another commit for the migration guide entry?

davidhewitt:
Please do, at the moment when I get a scrap of time I'm trying to finish off #3029

newsfragments/2981.removed.md
adamreichold: Done in the rebase.

adamreichold:
I added an entry to the migration guide but admittedly the example is quite artificial. Please have a look and let me know if this is sufficient for our purposes.

guide/src/migration.md
```diff
@@ -3,6 +3,80 @@
 This guide can help you upgrade code through breaking changes from one PyO3 version to the next.
 For a detailed list of all changes, see the [CHANGELOG](changelog.md).
 
+## from 0.18.* to 0.19
+
+### The deprecated `Python::acquire_gil` was removed and `Python::with_gil` must be used instead
+
+While API provided by [`Python::acquire_gil`](https://docs.rs/pyo3/0.18.3/pyo3/marker/struct.Python.html#method.acquire_gil) seems convenient, it is somewhat brittle as the design of the GIL token [`Python`](https://docs.rs/pyo3/0.18.3/pyo3/marker/struct.Python.html) relies on proper nesting and panics if not used correct, e.g.
```
davidhewitt, line 10:
```suggestion
While API provided by [`Python::acquire_gil`](https://docs.rs/pyo3/0.18.3/pyo3/marker/struct.Python.html#method.acquire_gil) seems convenient, it is somewhat brittle as the design of the GIL token [`Python`](https://docs.rs/pyo3/0.18.3/pyo3/marker/struct.Python.html) relies on proper nesting and panics if not used correctly, e.g.
```

davidhewitt (commented):
Looks great to me, I agree it's artificial though it clearly demonstrates what a problematic pattern might be and how `Python::with_gil` would be used instead.

adamreichold:
bors r=adamreichold,davidhewitt

Issue #2219, opened 2022-03-09
Cannot run tests with the abi3 feature.


mejrs:
### Bug Description

When I run pyo3's tests, I can't run tests with the abi3 feature

### Steps to Reproduce

cargo test --features=abi3

### Backtrace

```shell
Finished test [unoptimized + debuginfo] target(s) in 2m 38s
     Running unittests (target\debug\deps\pyo3-47fd354bc7aaab3c.exe)
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `C:\Users\bruno\rust\pyo3\target\debug\deps\pyo3-47fd354bc7aaab3c.exe` (exit code: 0xc0000139, STATUS_ENTRYPOINT_NOT_FOUND)
```

### Your operating system and version

Windows 10

### Your Python version (`python --version`)

Python 3.10.2

### Your Rust version (`rustc --version`)

rustc 1.58.1 (db9d1b20b 2022-01-20)

### Your PyO3 version

0.16

### How did you install python? Did you use a virtualenv?

downloaded from python.org

### Additional Info

I'm always getting incremental compilation failures, e.g. "warning: Error finalizing incremental compilation session directory `\\?\C:\Users\bruno\Rust\pyo3\target\debug\incremental\test_wrap_pyfunction_deduction-30s32wv9zkd0t\s-g7rdsrec4r-1f5oeza-working`: Access is denied. (os error 5)"

davidhewitt:
Weird.
- Does a `cargo clean` fix it?
- For the incremental errors, do you have antivirus software scanning the target dir?

mejrs:
> Does a cargo clean fix it?

No

> For the incremental errors, do you have antivirus software scanning the target dir?

I ignore my target dirs, so i don't think so.

davidhewitt:
Hmmm strange. I just retested on my Windows box and I can't reproduce this. Try updating to Rust 1.59? I'd be suprised if that fixed it though.

Otherwise, I'm not really sure what else to suggest...

mejrs:
I ran it in Dependency Walker and it turned out it was linking to a `python3.dll` and `python38.dll` in `C:\Program Files\Amazon\AWSCLIV2`. I deleted those and now it works 🥲

PR #2982, opened 2023-02-22
explain STATUS_DLL_NOT_FOUND

mejrs:
I had someone ask me this today, and googling for it I found various other places where this was asked, with no (good) answers.

### DIFF
--- a/guide/src/faq.md
+++ b/guide/src/faq.md
@@ -180,3 +180,15 @@ done with the `crate` attribute:
 #[pyo3(crate = "reexported::pyo3")]
 struct MyClass;
 ```
+
+## I'm trying to call Python from Rust but I get `STATUS_DLL_NOT_FOUND` or `STATUS_ENTRYPOINT_NOT_FOUND`!
+
+This happens on Windows when linking to the python DLL fails or the wrong one is linked. Your `PATH` variable should contain something like...
+
+```text
+C:\Users\<You>\AppData\Local\Programs\Python\Python310
+```
+
+This directory should have, for Python 3.10, the files `python3.dll` and `python310.dll`. If you have made any changes to your `PATH` or installed things recently, try restarting your shell - any changes will not affect already open shells.
+
+Failing that, you can use [Dependency Walker](https://www.dependencywalker.com/) to diagnose linking errors.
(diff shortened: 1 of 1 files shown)


guide/src/faq.md
```diff
@@ -180,3 +180,15 @@ done with the `crate` attribute:
...
+
+## I'm trying to call Python from Rust but I get `STATUS_DLL_NOT_FOUND` or `STATUS_ENTRYPOINT_NOT_FOUND`!
+
+This happens on Windows when linking to the python DLL fails or the wrong one is linked. Your `PATH` variable should contain something like...
+
+```text
+C:\Users\<You>\AppData\Local\Programs\Python\Python310
+```
+
+This directory should have, for Python 3.10, the files `python3.dll` and `python310.dll`. If you have made any changes to your `PATH` or installed things recently, try restarting your shell - any changes will not affect already open shells.
```
adamreichold, line 192: I also had the converse issue of other programs adding themselves and their internally used DLL to the `PATH` thereby messing things. For example, programs written in Java often include a JRE which can mess up a dedicated installation thoroughly. Or ArcGIS contains a Python installation used for scripts and plug-ins which could place unexpected Python DLL there.

Not sure if this is worth mentioning explicitly here, just thought I'd bring it up for consideration.

guide/src/faq.md
```diff
@@ -180,3 +180,15 @@ done with the `crate` attribute:
...
+
+Failing that, you can use [Dependency Walker](https://www.dependencywalker.com/) to diagnose linking errors.
```
davidhewitt, line 194: Potentially also worth mentioning `PyOxidizer` as a way to build a self-contained `.exe`?

davidhewitt (commented):
Thanks, good idea to document this!

guide/src/faq.md
mejrs: Yes, I hit this in https://github.com/PyO3/pyo3/issues/2219 😓

adamreichold:
I do think we could bikeshed the wording some more, but I also think this is a definitive improvement over not having this section. So I'll just go ahead and

bors r+

and take the blame if there is any.

Issue #2998, opened 2023-03-02
Better handling of `anyhow::Error` and `eyre::Report`


BlueGlassBlock:
Both `anyhow::Error` and `eyre::Report` are for dynamic error handling, I think it's feasible to make some improvements:

- Unwrap inner error, if it's containing a `PyErr` (However extra information should be combined to?)
- Extract the error chain into an exception chain (since both libraries have `Error::chain`/`Report::chain`) 

I'm willing to contribute :)

adamreichold:
I think it would be best to propose concrete separate PR for these changes to discuss them based on a concrete implementation instead of in the abstract.

mejrs:
As far as I understand anyhow already prints its error in a similar manner to how nested exceptions are printed. What difference would this make and what things would it let you do that you cannot do right now?

BlueGlassBlock:
> As far as I understand anyhow already prints its error in a similar manner to how nested exceptions are printed. What difference would this make and what things would it let you do that you cannot do right now?

When using `anyhow::Result` as function return type, `RuntimeError` is always raised when an error occurred. I have some Python callbacks to call within the Rust function, whose exceptions will also be converted to `RuntimeError` (`PyErr` -> `anyhow::Error` -> `PyErr`)

mejrs:
I'm not entirely sure what you want - do you want to be able to return a anyhow::Error to python, have it converted to a RuntimeError and then be able to get the original error back? If so that'll be hard because right now we just convert it to a string and create a RuntimeError with that.

BlueGlassBlock:
> I'm not entirely sure what you want - do you want to be able to return a anyhow::Error to python, have it converted to a RuntimeError and then be able to get the original error back? If so that'll be hard because right now we just convert it to a string and create a RuntimeError with that.

I prefer letting the original error bubble up, other errors get converted to RuntimeError. (`Error::downcast(PyErr)`)

mejrs:
I don't understand you at all. Can you maybe give an example?

BlueGlassBlock:
```rust
impl From<eyre::Report> for PyErr {
    fn from(error: Report) -> Self {
        match error.downcast::<Self>() {
            Ok(py_err) => py_err,
            Err(report) => PyRuntimeError::new_err(format!("{:?}", report))
        }
    }
}

adamreichold:
> However I think it's too naive:

Probably, but getting those into the main branch would be an improvement already. Why not open a PR?

PR #3004, opened 2023-03-04
Unwrap dynamic error types when inner is simple `PyErr`

BlueGlassBlock:
This is the first part of suggested improvements in #2998.

This change will make bubbled `PyErr` wrapped in `eyre::Report` / `anyhow::Error` bubble up unchanged, instead of being wrapped in a `PyRuntimeError`.

### DIFF
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -3,6 +3,43 @@
 This guide can help you upgrade code through breaking changes from one PyO3 version to the next.
 For a detailed list of all changes, see the [CHANGELOG](changelog.md).
 
+## from 0.18.* to 0.19
+
+### Smarter `anyhow::Error` / `eyre::Report` conversion when inner error is "simple" `PyErr`
+
+When converting from `anyhow::Error` or `eyre::Report` to `PyErr`, if the inner error is a "simple" `PyErr` (with no source error), then the inner error will be used directly as the `PyErr` instead of wrapping it in a new `PyRuntimeError` with the original information converted into a string.
+
+```rust
+# #[cfg(feature = "anyhow")]
+# #[allow(dead_code)]
+# mod anyhow_only {
+# use pyo3::prelude::*;
+# use pyo3::exceptions::PyValueError;
+#[pyfunction]
+fn raise_err() -> anyhow::Result<()> {
+    Err(PyValueError::new_err("original error message").into())
+}
+
+fn main() {
+    Python::with_gil(|py| {
+        let rs_func = wrap_pyfunction!(raise_err, py).unwrap();
+        pyo3::py_run!(py, rs_func, r"
+        try:
+            rs_func()
+        except Exception as e:
+            print(repr(e))
+        ");
+    })
+}
+# }
+```
+
+Before, the above code would have printed `RuntimeError('ValueError: original error message')`, which might be confusing.
+
+After, the same code will print `ValueError: original error message`, which is more straightforward.
+
+However, if the `anyhow::Error` or `eyre::Report` has a source, then the original exception will still be wrapped in a `PyRuntimeError`.
+
 ## from 0.17.* to 0.18
 
 ### Required arguments after `Option<_>` arguments will no longer be automatically inferred
--- a/src/conversions/anyhow.rs
+++ b/src/conversions/anyhow.rs
@@ -9,9 +9,10 @@
 //! want error handling to be easy. If you are writing a library or you need more control over your
 //! errors you might want to design your own error type instead.
 //!
-//! This implementation always creates a Python [`RuntimeError`]. You might find that you need to
-//! map the error from your Rust code into another Python exception. See [`PyErr::new`] for more
-//! information about that.
+//! When the inner error is a [`PyErr`] without source, it will be extracted out.
+//! Otherwise a Python [`RuntimeError`] will be created.
+//! You might find that you need to map the error from your Rust code into another Python exception.
+//! See [`PyErr::new`] for more information about that.
 //!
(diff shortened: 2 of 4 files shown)


adamreichold:
@mejrs Will wait for your feedback before merging as you sounded unsure whether this is a sensible thing to do in the issue.

src/conversions/anyhow.rs
```diff
@@ -111,13 +111,21 @@ use crate::exceptions::PyRuntimeError;
 use crate::PyErr;
 
 impl From<anyhow::Error> for PyErr {
-    fn from(err: anyhow::Error) -> Self {
-        PyRuntimeError::new_err(format!("{:?}", err))
+    fn from(mut error: anyhow::Error) -> Self {
+        // Error chains & error with contexts are skipped
+        if error.source().is_none() {
+            error = match error.downcast::<Self>() {
```
davidhewitt, line 117: Rather than the assignment here, might it be simpler to just do `if let Ok(py_err) = error.downcast::<Self>() { return py_err; }`

And same for the eyre implementation.

src/conversions/anyhow.rs
```diff
@@ -111,13 +111,21 @@ use crate::exceptions::PyRuntimeError;
 use crate::PyErr;
 
 impl From<anyhow::Error> for PyErr {
-    fn from(err: anyhow::Error) -> Self {
-        PyRuntimeError::new_err(format!("{:?}", err))
+    fn from(mut error: anyhow::Error) -> Self {
+        // Error chains & error with contexts are skipped
```
davidhewitt, line 115: This comment is slightly confusing to me as a reader; why are they skipped?

I think the following would be clearer:

```suggestion
        // Errors containing a PyErr without chain or context are returned as the underlying error 
```

davidhewitt (commented):
While I think this change is strictly an improvement, I'd like to quickly take a moment to think about how we document this and / or provide backwards compatibility.

Changing the error type observed here may catch some users out. Probably module docs need updating. Should this go in the migration guide?

src/conversions/anyhow.rs
adamreichold: This does not work as `downcast` consumes `error`, but if the downcast fails, we still want to format it into a string. So either we duplicate that code path or use the assignment. Personally, I find the assignment to be an elegant solution to this problem.

davidhewitt: Ah great point, of course. Ignore me!

DataTriny:
Hello @BlueGlassBlock 

Thank you!

mejrs:
> @mejrs Will wait for your feedback before merging as you sounded unsure whether this is a sensible thing to do in the issue.

I have nothing, feel free to merge as you wish :)

adamreichold:
@BlueGlassBlock You would be up to writing an entry for the migration guide showing the difference when forwarding errors using `anyhow` or `eyre`? The module docs definitely must be changed to match the proposed implementation.

BlueGlassBlock:
> @BlueGlassBlock You would be up to writing an entry for the migration guide showing the difference when forwarding errors using `anyhow` or `eyre`? The module docs definitely must be changed to match the proposed implementation.

Certainly!

guide/src/migration.md
```diff
@@ -3,6 +3,42 @@
...
+fn raise_err() -> anyhow::Result<_> {
+    // Here we simulate a simple `PyErr` is raised
+    Err(PyValueError::new_err("Rust Value Error"))?;
+}
+
+fn main() {
+    Python::with_gil(|py| {
+        let res = raise_err();
+
+        // Similar conversion is performed when the function is wrapped with #[pyfunction]
```
adamreichold, line 27: I think it would be nice to just define a `#[pyfunction]` in the example and call that using `wrap_pyfunction!` and `py_run!` instead of hinting at this in a comment.

guide/src/migration.md
```diff
@@ -3,6 +3,42 @@
...
+
+```rust
+# #[cfg(feature = "anyhow")]
+# #[allow(dead_code)]
+# mod anyhow_only {
+# use pyo3::prelude::*;
+use anyhow::Result;
+fn raise_err() -> anyhow::Result<_> {
+    // Here we simulate a simple `PyErr` is raised
+    Err(PyValueError::new_err("Rust Value Error"))?;
```
adamreichold, line 20: While this works, it is bit confusing, maybe better do

```rust
Err(PyValueError::new_err("original error message").into())
```

adamreichold:
Thanks for drafting the migration guide entry. Please remember to adjust the existing module-level docs in `anyhow.rs` and `eyre.rs` which currently still state that a `RuntimeError` is produced unconditionally.

guide/src/migration.md
```diff
@@ -3,6 +3,43 @@
...
+            print(repr(e))
+        ");
+    })
+}
+# }
+```
+
+Before, the above code would have printed `RuntimeError('ValueError: original error message')`, which might be confusing.
+
+After, the same code will print `ValueError: Rust Value Error`, which is more straightforward.
```
adamreichold, line 39:
```suggestion
After, the same code will print `ValueError: original error message`, which is more straightforward.
```

adamreichold:
> Changing the error type observed here may catch some users out. Probably module docs need updating. Should this go in the migration guide?

@davidhewitt I think this better positioned now w.r.t. the above concerns and therefore ready to be merged. Do you agree?

davidhewitt (approved):
Yep this looks great to me now, thanks! I'll squash and merge this now.

davidhewitt:
bors r=davidhewitt,adamreichold,mejrs

PR #3014, opened 2023-03-05
feat: add #[pyo3(get, set)] for Cell, Arc and Box

### LINKED ISSUE #2659: `#[pyo3(get, set)]` for wrapper types
As an offhand thought, I wonder if we should be trying to make `Cell<T>` automatically work with `#[pyo3(get)]`? (And similar for other wrapper types such as `Box<T>` or `Mutex<T>`?)

Out of scope of this PR of course.

_Originally posted by @davidhewitt in https://github.com/PyO3/pyo3/pull/2657#discussion_r987222272_

----

Presumably there could be a trait which becomes part of the public PyO3 API to control this for people who wanted to implement it.

@mejrs suggested in the original thread that it might be too far to support `Mutex`, and I'm inclined to agree, we can see what makes sense if someone comes forward with a design.

AntoineRR:
This fixes #2659.
The types for which `#[pyo3(get, set)]` should now work are `Cell`, `Arc` and `Box`.

There is one issue regarding `Box`, the implementation of `FromPyObject` conflicts with another one. I could not find what the issue was, especially since the other implementations for `Arc` and `Cell` work as expected. The related code and test has been commented out for now. Maybe someone could help me fix this issue if I don't figure it out myself? There is also the possibility to remove the implementation for `Box` of course.


src/conversion.rs
```diff
@@ -370,6 +372,60 @@ where
     }
 }
 
+impl<T: ToPyObject> ToPyObject for Box<T> {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        self.as_ref().to_object(py)
+    }
+}
+
+impl<T: Clone + IntoPy<PyObject>> IntoPy<PyObject> for Box<T> {
```
adamreichold, line 381: The `Clone` bound is not necessary IIUC, e.g.

```rust
impl<T: IntoPy<PyObject>> IntoPy<PyObject> for Box<T> {
        (*self).into_py(py)
    }
}
```

src/conversion.rs
```diff
@@ -370,6 +372,60 @@ where
...
+    fn into_py(self, py: Python<'_>) -> PyObject {
+        self.as_ref().clone().into_py(py)
+    }
+}
+
+// impl<'a, T: FromPyObject<'a>> FromPyObject<'a> for Box<T> {
```
adamreichold, line 387: I think the coherence error is correct insofar `Box<T>` is a "fundamental" type IIRC, i.e. downstream crates can `impl PyClass for Box<T>` with the same coherence rules as if the implemented `impl PyClass for T`.

I am not sure whether we can get around this without introducer said additional trait to decouple `FromPyObject`'s task to "produce a `T` given a `PyAny`" from the new (sealed) trait's "I consider this a transparent wrapper of its inner value get/set purposes", so maybe

```rust
trait PyPropertyWrapper<T>: Sealed {
  fn set(val: T) -> Self;
  fn get(&self) -> T;
}
```

?

mejrs:
I think the Arc/Box impls are a bad idea. In all cases I've seen where people stumbled upon their lack of impls this would not do what people expect, for example if you have a binary tree like in https://github.com/PyO3/pyo3/discussions/2839, then `#[pyo3(get)]` will just make a full clone of the field. IMO it would be better to not have this impl so people are forced to think about what they actually want.

kngwyu:
I also feel like `Box` are often used with complex recurrent design patterns, so I'm not sure we can recommend exposing it in this way. ` ToPyObject` implies that the conversion can be expensive so it should be OK, though.

davidhewitt:
I have to agree with the reservations about `Box` and `Arc`. Similar to @adamreichold's proposed `PyPropertyWrapper` above, I think there's also a third member of `get`/`set`, which I've been thinking in my head as `bind`.

The thought is that for e.g. `Arc<T: PyClass>`, maybe rather than cloning into a new Python object we could make a new Python object which just stores `Arc<T>` and so shares ownership between Python and Rust. So I'm a bit wary of trying to go too far down this road with `Box` and `Arc` without understanding how that option might work out.

For `Cell`, I'm reasonably comfortable that adding this as-is will be an improvement.

AntoineRR:
Thanks everyone for your feedback! I didn't foresee the potentials flaws of implementing `IntoPy` for `Box` and `Arc`, I will remove those from the PR.
I think I understand the general idea behind the `PyPropertyWrapper` trait although I may need more explanations (especially on `bind`) to try to implement this. Maybe this trait introduction should go in an other PR if we keep the `IntoPy` implementation for `Cell`?
I also thought about other types, but I think they do not fit for an `IntoPy` impl : `RefCell` (because `borrow` can panic), `Mutex` (because `lock` can block), and `Rc` (because it is not thread safe). If you have other types in mind, please let me know.

adamreichold:
> Maybe this trait introduction should go in an other PR if we keep the IntoPy implementation for Cell?

Agreed. I think the `Cell` additions is the completely uncontroversial and definitely useful part here. Let's finish this up before trying to start into more ambitious directions.

src/conversion.rs
AntoineRR: Thanks for letting me know about fundamental types, this explains the issue with `Box`!

PR #3016, opened 2023-03-05
implement Decimal to rust_decimal conversions

cardoe:
Implement conversion between rust_decimal::Decimal and decimal.Decimal from Python's stdlib. The C API does not appear to be exposed on the Python side so we need to call into it via Python.

Implements #2774

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ hashbrown = { version = ">= 0.9, < 0.14", optional = true }
 indexmap = { version = "1.6", optional = true }
 num-bigint = { version = "0.4", optional = true }
 num-complex = { version = ">= 0.2, < 0.5", optional = true }
+rust_decimal = { version = "1.0.0", default-features = false, optional = true }
 serde = { version = "1.0", optional = true }
 
 [dev-dependencies]
@@ -53,6 +54,7 @@ send_wrapper = "0.6"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.61"
 rayon = "1.0.2"
+rust_decimal = { version = "1.8.0", features = ["std"] }
 widestring = "0.5.1"
 
 [build-dependencies]
@@ -110,6 +112,7 @@ full = [
     "eyre",
     "anyhow",
     "experimental-inspect",
+    "rust_decimal",
 ]
 
 [[bench]]
@@ -119,6 +122,11 @@ harness = false
 [[bench]]
 name = "bench_err"
 harness = false
+
+[[bench]]
+name = "bench_decimal"
+harness = false
+required-features = ["rust_decimal"]
 
 [[bench]]
 name = "bench_dict"
@@ -173,5 +181,5 @@ members = [
 
 [package.metadata.docs.rs]
 no-default-features = true
-features = ["macros", "num-bigint", "num-complex", "hashbrown", "serde", "multiple-pymethods", "indexmap", "eyre", "chrono"]
+features = ["macros", "num-bigint", "num-complex", "hashbrown", "serde", "multiple-pymethods", "indexmap", "eyre", "chrono", "rust_decimal"]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/benches/bench_decimal.rs
+++ b/benches/bench_decimal.rs
@@ -1,0 +1,32 @@
+use criterion::{black_box, criterion_group, criterion_main, Bencher, Criterion};
+
+use pyo3::prelude::*;
+use pyo3::types::PyDict;
+use rust_decimal::Decimal;
+
+fn decimal_via_extract(b: &mut Bencher<'_>) {
+    Python::with_gil(|py| {
+        let locals = PyDict::new(py);
+        py.run(
+            r#"
+import decimal
+py_dec = decimal.Decimal("0.0")
+"#,
+            None,
+            Some(locals),
+        )
+        .unwrap();
+        let py_dec = locals.get_item("py_dec").unwrap();
+
+        b.iter(|| {
+            let _: Decimal = black_box(py_dec).extract().unwrap();
+        });
+    })
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    c.bench_function("decimal_via_extract", decimal_via_extract);
+}
+
+criterion_group!(benches, criterion_benchmark);
+criterion_main!(benches);
--- a/src/conversions/rust_decimal.rs
+++ b/src/conversions/rust_decimal.rs
@@ -1,0 +1,221 @@
+#![cfg(feature = "rust_decimal")]
+//! Conversions to and from [rust_decimal](https://docs.rs/rust_decimal)'s [`Decimal`] type.
+//!
+//! This is useful for converting Python's decimal.Decimal into and from a native Rust type.
+//!
+//! # Setup
+//!
+//! To use this feature, add to your **`Cargo.toml`**:
+//!
+//! ```toml
+//! [dependencies]
+//! rust_decimal = "1.0"
+// workaround for `extended_key_value_attributes`: https://github.com/rust-lang/rust/issues/82768#issuecomment-803935643
+#![cfg_attr(docsrs, cfg_attr(docsrs, doc = concat!("pyo3 = { version = \"", env!("CARGO_PKG_VERSION"),  "\", features = [\"rust_decimal\"] }")))]
+#![cfg_attr(
+    not(docsrs),
+    doc = "pyo3 = { version = ..., features = [\"rust_decimal\"] }"
+)]
+//! ```
+//!
+//! Note that you must use a compatible version of rust_decimal and PyO3.
+//! The required rust_decimal version may vary based on the version of PyO3.
+//!
+//! # Example
+//!
+//! Rust code to create a function that adds one to a Decimal
+//!
+//! ```rust
+//! use rust_decimal::Decimal;
+//! use pyo3::prelude::*;
+//!
+//! #[pyfunction]
+//! fn add_one(d: Decimal) -> Decimal {
+//!     d + Decimal::ONE
+//! }
+//!
+//! #[pymodule]
+//! fn my_module(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
+//!     m.add_function(wrap_pyfunction!(add_one, m)?)?;
+//!     Ok(())
+//! }
+//! ```
+//!
+//! Python code that validates the functionality
+//!
+//!
+//! ```python
+//! from my_module import add_one
+//! from decimal import Decimal
+//!
+//! d = Decimal("2")
+//! value = add_one(d)
+//!
+//! assert d + 1 == value
+//! ```
+
+use crate::exceptions::PyValueError;
+use crate::once_cell::GILOnceCell;
+use crate::types::PyType;
+use crate::{intern, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, Python, ToPyObject};
+use rust_decimal::Decimal;
+use std::str::FromStr;
+
+impl FromPyObject<'_> for Decimal {
+    fn extract(obj: &PyAny) -> PyResult<Self> {
+        // use the string representation to not be lossy
+        if let Ok(val) = obj.extract() {
+            Ok(Decimal::new(val, 0))
+        } else {
+            Decimal::from_str(obj.str()?.to_str()?)
+                .map_err(|e| PyValueError::new_err(e.to_string()))
+        }
+    }
+}
+
+static DECIMAL_CLS: GILOnceCell<Py<PyType>> = GILOnceCell::new();
+
+fn get_decimal_cls(py: Python<'_>) -> PyResult<&PyType> {
+    DECIMAL_CLS
+        .get_or_try_init(py, || {
+            py.import(intern!(py, "decimal"))?
+                .getattr(intern!(py, "Decimal"))?
+                .extract()
+        })
+        .map(|ty| ty.as_ref(py))
+}
+
+impl ToPyObject for Decimal {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        // TODO: handle error gracefully when ToPyObject can error
+        // look up the decimal.Decimal
+        let dec_cls = get_decimal_cls(py).expect("failed to load decimal.Decimal");
+        // now call the constructor with the Rust Decimal string-ified
+        // to not be lossy
+        let ret = dec_cls
+            .call1((self.to_string(),))
+            .expect("failed to call decimal.Decimal(value)");
+        ret.to_object(py)
+    }
+}
+
+impl IntoPy<PyObject> for Decimal {
+    fn into_py(self, py: Python<'_>) -> PyObject {
+        self.to_object(py)
+    }
+}
+
+#[cfg(test)]
+mod test_rust_decimal {
+    use super::*;
+    use crate::err::PyErr;
+    use crate::types::PyDict;
+    use rust_decimal::Decimal;
+
+    #[cfg(not(target_arch = "wasm32"))]
+    use proptest::prelude::*;
+
+    macro_rules! convert_constants {
+        ($name:ident, $rs:expr, $py:literal) => {
+            #[test]
+            fn $name() {
+                Python::with_gil(|py| {
+                    let rs_orig = $rs;
+                    let rs_dec = rs_orig.into_py(py);
+                    let locals = PyDict::new(py);
+                    locals.set_item("rs_dec", &rs_dec).unwrap();
+                    // Checks if Rust Decimal -> Python Decimal conversion is correct
+                    py.run(
+                        &format!(
+                            "import decimal\npy_dec = decimal.Decimal({})\nassert py_dec == rs_dec",
+                            $py
+                        ),
+                        None,
+                        Some(locals),
+                    )
+                    .unwrap();
+                    // Checks if Python Decimal -> Rust Decimal conversion is correct
+                    let py_dec = locals.get_item("py_dec").unwrap();
+                    let py_result: Decimal = FromPyObject::extract(py_dec).unwrap();
+                    assert_eq!(rs_orig, py_result);
+                })
+            }
+        };
+    }
+
+    convert_constants!(convert_zero, Decimal::ZERO, "0");
+    convert_constants!(convert_one, Decimal::ONE, "1");
+    convert_constants!(convert_neg_one, Decimal::NEGATIVE_ONE, "-1");
+    convert_constants!(convert_two, Decimal::TWO, "2");
+    convert_constants!(convert_ten, Decimal::TEN, "10");
+    convert_constants!(convert_one_hundred, Decimal::ONE_HUNDRED, "100");
+    convert_constants!(convert_one_thousand, Decimal::ONE_THOUSAND, "1000");
+
+    #[cfg(not(target_arch = "wasm32"))]
+    proptest! {
+        #[test]
+        fn test_roundtrip(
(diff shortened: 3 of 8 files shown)


cardoe:
I created the PR as suggested by @davidhewitt in #2774 to start the conversation. I added an initial benchmark but I'm not sure if it's sufficient to start. The results were:
```
Benchmarking decimal_via_extract: Collecting 100 samples in estimated 5.0001 s (
decimal_via_extract     time:   [46.788 ns 46.803 ns 46.819 ns]
Found 15 outliers among 100 measurements (15.00%)
  7 (7.00%) low severe
  5 (5.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe
```

not sure why I had that trailing open paren.

The tests fail because I haven't written any docs yet. I've still got to address @adamreichold's feedback as well but I figured I'd start with the benchmarks to see if it makes sense and if not talk to CPython upstream about exposing an API. If the benchmarks are okay then I'll work in parallel. Add the necessary bits to raise this to completion while getting an API exposed in CPython.

adamreichold:
Personally, I think that independently of the kind of (modernish) system this was run on, that the performance is reasonable and would have this than have it not, understanding that we can always improve things if better CPython API become available.

mejrs:
IMO we should only have these conversions if they're better than what a user could naively write themselves. Python's decimal has no C api, so this isn't the case here.

adamreichold:
> IMO we should only have these conversions if they're better than what a user could naively write themselves. Python's decimal has no C api, so this isn't the case here.

But why force the user into reinventing this if we can solve the problem once for everybody? As long as it is the best possible solution. And if a better way comes along, we can improve things in a single place and everyone benefits immediately.

cardoe:
@mejrs My thought process for this PR was that decimal is part of the stdlib as a type. I don't think it makes sense to add every part of stdlib but looking through most of the types (I'm counting the types under the numerical as well) are covered except for big integers which is what this aims to cover.

kngwyu:
I'm not actively catching up on the project, so I'm not sure, but is there any other module where we support conversion between some Rust library and Python object which doesn't have C-API? If so, having this would be OK.

But I'm not very convinced with this string-based conversion... If we can use tuple instead, it would be faster.

davidhewitt:
> IMO we should only have these conversions if they're better than what a user could naively write themselves. Python's decimal has no C api, so this isn't the case here.

I keep going back and forth on this line of thinking. I think ultimately the types with a C-API are not significantly different except that we wrap away the `unsafe` parts.

I do agree with @adamreichold's argument that getting this once right here is better for the ecosystem than forcing fragmentation. Also we essentially use Python APIs for `PySequence` and `PyMapping` now that we use the collections base classes, so perhaps we have to accept that to make "nice" bindings between the two languages we have to accept the Python C-API limitations at times?

Although I still agree it would be great to have a native conversion provided upstream, if there is willingness.

> But I'm not very convinced with this string-based conversion... If we can use tuple instead, it would be faster.

It's not clear to me which conversion options are available, however it would be good to sense-check what we're offering here is the most efficient.

davidhewitt:
Also of note is that we do have similar precedent with support for converting `pathlib.Path` into Rust types.

cardoe:
So the reason I went with the string over the tuple is that the Python tuple doesn't map to the Rust interface nicely.

The Python tuple documented at https://docs.python.org/3/library/decimal.html#decimal.Decimal says its
> If value is a [tuple](https://docs.python.org/3/library/stdtypes.html#tuple), it should have three components, a sign (0 for positive or 1 for negative), a [tuple](https://docs.python.org/3/library/stdtypes.html#tuple) of digits, and an integer exponent. For example, Decimal((0, (1, 4, 1, 4), -3)) returns Decimal('1.414')
which would be returned from https://docs.python.org/3/library/decimal.html#decimal.Decimal.as_tuple

We'd have a couple of weird corner cases with Infinity and NaN via as_tuple as well.

```
>>> nan = Decimal('NaN')
>>> nan.as_tuple()
DecimalTuple(sign=0, digits=(), exponent='n')
>>> inf = Decimal('Infinity')
>>> inf.as_tuple()
DecimalTuple(sign=0, digits=(0,), exponent='F')
```

Now from the Rust side the closest to the tuple interface is probably `new()` or `from_str_radix()`. https://docs.rs/rust_decimal/1.28.1/rust_decimal/struct.Decimal.html#method.new and https://docs.rs/rust_decimal/1.28.1/rust_decimal/struct.Decimal.html#method.from_str_radix

- `new()`  takes in the digits as a signed number and the exponent. So that will require some conversion of the digits to an `i64`
- `from_str_radix()` would take the digits as a `&str` and would probably be easier to convert than the above but will cause a temp allocation. But this would probably perform better in the Rust side than the Python side.

So ultimately if that's approach you'd like me to take I think I'd steer towards `from_str_radix()` if that sounds okay?

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#[cfg(test)]
+mod test_rust_decimal {
+    use super::*;
+    use crate::types::PyDict;
+    use indoc::formatdoc;
+    use rust_decimal::Decimal;
+
+    macro_rules! convert_constants {
```
davidhewitt, line 54: It may be interesting to use `proptest` to try conversion of some arbitrary values as a round-trip (you can see examples of this elsewhere in the code).

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,91 @@
...
+        };
+    }
+
+    convert_constants!(convert_zero, Decimal::ZERO, "0");
+    convert_constants!(convert_one, Decimal::ONE, "1");
+    convert_constants!(convert_neg_one, Decimal::NEGATIVE_ONE, "-1");
+    convert_constants!(convert_two, Decimal::TWO, "2");
+    convert_constants!(convert_ten, Decimal::TEN, "10");
+    convert_constants!(convert_one_hundred, Decimal::ONE_HUNDRED, "100");
+    convert_constants!(convert_one_thousand, Decimal::ONE_THOUSAND, "1000");
```
davidhewitt, line 90: It sounds like we need tests for infinity and NaN?

davidhewitt (commented):
I don't have any preference for one implementation over the other. If you're willing to experiment, it would be good to compare `from_str_radix` and `from_str_exact` (if those are the candidates we're looking at), to see if there's any meaningful implementation and performance differences.

src/conversions/rust_decimal.rs
cardoe: There's no representation of Infinity however. Not sure on the best approach there.

kngwyu:
@cardoe 
Thank you for your explanation. String-based conversion sounds reasonable for me now.

src/conversions/rust_decimal.rs
davidhewitt: I think `Option<Decimal>` isn't appropriate here, because that really means Python `Decimal` or Python `None`.

Probably best to just make those cases return errors for the moment.

If someone needs NaN & infinity they could build a wrapper around `rust_decimal` and check for those values before using the conversion defined here.

src/conversions/rust_decimal.rs
cardoe: `rust_decimal` didn't support `proptest` so I've now done that at https://github.com/paupino/rust-decimal/pull/582

cardoe: Lifted that into pyo3 so that it would work for now without a bump.

src/conversions/rust_decimal.rs
cardoe: Added tests for `NaN` and `Infinity`

cardoe:
I believe this addresses all the feedback thus far and could be the first commit considered for a full review / inclusion.

cardoe:
I think this fixed the MSRV issue.

cardoe:
I've just got an M1 based Mac to test on right now so the MSRV won't build on here so fingers crossed really.

davidhewitt:
Hello, I'll do my best to review this tomorrow. A bit behind on OSS this week, had a sick family. We're better now finally :)

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,216 @@
...
+//! To use this feature, add to your **`Cargo.toml`**:
+//!
+//! ```toml
+//! [dependencies]
+//! rust_decimal = "1.0"
+// workaround for `extended_key_value_attributes`: https://github.com/rust-lang/rust/issues/82768#issuecomment-803935643
+#![cfg_attr(docsrs, cfg_attr(docsrs, doc = concat!("pyo3 = { version = \"", env!("CARGO_PKG_VERSION"),  "\", features = [\"rust_decimal\"] }")))]
+#![cfg_attr(
+    not(docsrs),
+    doc = "pyo3 = { version = \"*\", features = [\"rust_decimal\"] }"
```
birkenfeld, line 17: Maybe not encourage `version = "*"` but use `version = ...` to indicate something to replace?

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl FromPyObject<'_> for Decimal {
+    fn extract(obj: &PyAny) -> PyResult<Self> {
+        // use the string representation to not be lossy
+        // TODO: is PyTypeError the right thing? The Decimal code isn't
+        //       exposed via the C API so its a bit funky to make those
+        //       exceptions directly
+        if obj.is_instance_of::<PyLong>()? {
+            let num_int: i64 = FromPyObject::extract(obj)?;
+            Decimal::try_new(num_int, 0).map_err(|e| PyTypeError::new_err(e.to_string()))
```
birkenfeld, line 71: Since `try_new` can not fail with a known `scale=0`, use `new` instead, and no need to make an error.

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,216 @@
...
+        } else {
+            Decimal::from_str(obj.str()?.to_str()?).map_err(|e| PyTypeError::new_err(e.to_string()))
```
birkenfeld, line 73: Here, `ValueError` would be better.

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,216 @@
...
+impl FromPyObject<'_> for Decimal {
+    fn extract(obj: &PyAny) -> PyResult<Self> {
+        // use the string representation to not be lossy
+        // TODO: is PyTypeError the right thing? The Decimal code isn't
+        //       exposed via the C API so its a bit funky to make those
+        //       exceptions directly
+        if obj.is_instance_of::<PyLong>()? {
+            let num_int: i64 = FromPyObject::extract(obj)?;
+            Decimal::try_new(num_int, 0).map_err(|e| PyTypeError::new_err(e.to_string()))
+        } else {
```
birkenfeld, line 72: I'm not sure *any* object should be accepted. If it's not `Decimal`, check for `numbers.Number`? Or at least it should have a `__float__` conversion?

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+
+impl ToPyObject for Decimal {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        // TODO: how do we handle the error gracefully?
+        // look up the decimal.Decimal
+        let dec_mod = py
+            .import(intern!(py, "decimal"))
+            .expect("failed to import decimal");
+        let dec_cls = dec_mod
```
birkenfeld, line 85: Hm, I thought we had a similar instance where we cache a class object to avoid the lookup on each conversion. But I can't find it now.

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,216 @@
...
+            Decimal::try_new(num_int, 0).map_err(|e| PyTypeError::new_err(e.to_string()))
+        } else {
+            Decimal::from_str(obj.str()?.to_str()?).map_err(|e| PyTypeError::new_err(e.to_string()))
+        }
+    }
+}
+
+impl ToPyObject for Decimal {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        // TODO: how do we handle the error gracefully?
```
birkenfeld, line 80: Unfortunately `ToPyObject` currently can't. So panicking is ok here.

src/conversions/rust_decimal.rs
cardoe: I'll add a comment above like:
https://github.com/PyO3/pyo3/blob/ebd417c9ccea884d92792c75fbb14ad833f1eb9a/src/conversions/anyhow.rs#L31
https://github.com/PyO3/pyo3/blob/ebd417c9ccea884d92792c75fbb14ad833f1eb9a/src/conversions/chrono.rs#L21

The others don't have one. I followed num-bigint as an example when writing this.

src/conversions/rust_decimal.rs
cardoe: Yeah I'm not sure what the best approach would be here. I originally was only doing `Decimal` but along the way someone asked me to handle more. I never had generic floats working with tests cause precision differences. When you say `numbers.Number` is that `PyNumber`?

src/conversions/rust_decimal.rs
cardoe: I'll leave a TODO to address this if ToPyObject gets changed to a Result if that's okay? I'll rewrite the comment to be more appropriate.

cardoe:
> Hello, I'll do my best to review this tomorrow. A bit behind on OSS this week, had a sick family. We're better now finally :)

Take you time. Family first. OSS is the shoulder monkey that never goes away for better or worse. I'm not in any rush on this. I was playing around with a Python code base and grumping at "type whispering" biting me on something and though I would try to lift part of the code base into Rust and looked for `Decimal` didn't see it. So I wrote a wrapper type for my use case but figured since there was at least one other person looking for it that I would try to contribute back to make it easier for the next person.

src/conversions/rust_decimal.rs
adamreichold: Maybe you remember our handling of the `collections.Sequence` ABC in

https://github.com/PyO3/pyo3/blob/ebd417c9ccea884d92792c75fbb14ad833f1eb9a/src/types/sequence.rs#L319-L327

src/conversions/rust_decimal.rs
davidhewitt: Yep fine to leave like this for now. When we update MSRV above 1.54 we can simplify this and just have the `CARGO_PKG_VERSION` embedded everywhere.

src/conversions/rust_decimal.rs
davidhewitt: Yep unfortunately that's the best we can do for the moment, though I hope we can improve this soon.

DataTriny:
Hello @cardoe 

Thank you!

src/conversions/rust_decimal.rs
cardoe: I followed this style in the most recent change.

src/conversions/rust_decimal.rs
cardoe: I believe I addressed this in the latest commit.

cardoe:
I believe I've addressed all the PR feedback. Please let me know what I can do going forward. The current GitHub actions failures are happening for me locally on my machine with the latest main commit as well.

adamreichold:
@cardoe Could you rebase on `main` now that #3114 has landed?

benches/bench_decimal.rs
```diff
@@ -0,0 +1,35 @@
+use criterion::{black_box, criterion_group, criterion_main, Bencher, Criterion};
+
+use pyo3::prelude::*;
+use pyo3::types::PyDict;
+#[cfg(feature = "rust_decimal")]
```
adamreichold, line 5: All `#[cfg(feature = "rust_decimal")]` should be unnecessary due to the `required-features = ["rust-decimal"]` clause?

Cargo.toml
```diff
@@ -53,6 +54,7 @@ send_wrapper = "0.6"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.61"
 rayon = "1.0.2"
+rust_decimal = { version = "1.18.0", features = ["std"] }
```
adamreichold, line 57: Is this the oldest version which supports this functionality of our PR? If not, I would be glad if we could require the oldest version that is semver-compatible with 1.18.0.

src/conversions/rust_decimal.rs
adamreichold: I am sorry for re-opening, but I do think @birkenfeld's suggestion of using `version = ...` instead is better. As it stands now, I get the comment about `*` on docs.rs even though docs.rs uses `CARGO_PKG_VERSION` and hence there is no `*` in sight.

src/conversions/rust_decimal.rs
```diff
@@ -0,0 +1,229 @@
...
+use crate::once_cell::GILOnceCell;
+use crate::types::PyType;
+use crate::{intern, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, Python, ToPyObject};
+use rust_decimal::Decimal;
+use std::str::FromStr;
+
+impl FromPyObject<'_> for Decimal {
+    fn extract(obj: &PyAny) -> PyResult<Self> {
+        // use the string representation to not be lossy
+        unsafe {
```
adamreichold, line 70: I am not satisfied with how the numeric case is handled:
* `PyLong_AsLongLong` does call `PyNumber_Index` internally so it is unnecessary do it manually here.
* `PyLong_AsLongLong` can fail if the value overflows and hence we need something like `err_if_invalid_value` from the `conversions::num` module.

Finally, I would like this module to not delve down into these details would suggest something simpler like

```rust
if let Ok(val) = obj.extract() {
  Ok(Decimal::new(val, 0))
} else {
  Decimal::from_str(...)
}
```

that just re-uses the code from the `num` module and most importantly avoids the unsafe code.

adamreichold (changes requested):
Sorry doing another round on this.

src/conversions/rust_decimal.rs
cardoe: Yeah that's certainly better.

src/conversions/rust_decimal.rs
cardoe: Sure. I was just trying to match the style of the rest of the code base.

benches/bench_decimal.rs
cardoe: I think you're right. I've removed it.

Cargo.toml
cardoe: Honestly I don't even remember why 1.18.0 at this point. I know originally we really wanted to use `from_str_exact()` but that didn't come in until 1.21.0 which had a MSRV of 1.54 which is newer than what pyo3 has currently but is what pyo3 wants to move to. 1.19.0 and newer wants a newer version of arrayvec which wants const generics from 1.51 as well.

At this point all the call sites I see are using the more basic calls that I believe are there for 1.0 so I'll just go back to that. I originally had that in #2774 but it wasn't really compatible.

src/conversions/rust_decimal.rs
adamreichold: The version specification looks good now but it looks like you did not remove the comment about `*`?

adamreichold (approved):
There still is the left-over comment concerning the `*` version specification, otherwise I think this is good to land. Thanks for keeping at it!

src/conversions/rust_decimal.rs
adamreichold: Took the liberty of amending that commit and will merge this now as I think anything has been addressed now. Any other improvements can be follow-ups.

adamreichold:
bors retry

PR #3023, opened 2023-03-07
Emit a better error for bad argument names

mejrs:
This will emit a better error for code like 
```rust
#[pyfunction]
fn output([a,b,c]: [u8;3]) {}
```

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -40,14 +40,14 @@ impl<'a> FnArg<'a> {
                 }
 
                 let arg_attrs = PyFunctionArgPyO3Attributes::from_attrs(&mut cap.attrs)?;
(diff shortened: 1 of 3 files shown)


adamreichold:
I would be glad if you could add some UI tests which hit the new code path to ensure that we do not accidentally break this in the future.

mejrs:
> I would be glad if you could add some UI tests which hit the new code path to ensure that we do not accidentally break this in the future.

This is not a new code path (it just improves an error message) but sure 👍

adamreichold:
> This is not a new code path

I was referring to the two uncovered match arms.

davidhewitt:
I force-pushed to rebase and add the UI tests. Thanks & hope that's ok, I wanted to include this in 0.18.2 :)

bors r+

PR #3029, opened 2023-03-09
use dynamic trampoline for all getters and setters

davidhewitt:
This is an extension to the "trampoline" changes made in #2705 to re-use a single trampoline for all `#[getter]`s (and similar for all `#[setters]`). It works by setting the currently-unused `closure` member of the `ffi::PyGetSetDef` structure to point at a new `struct GetSetDefClosure` which contains function pointers to the `getter` / `setter` implementations.

A universal trampoline for all `getter`, for example, then works by reading the actual getter implementation out of the `GetSetDefClosure`.

Advantages of doing this:
- Very minimal simplification to the macro code / generated code size. It made a 4.4% reduction to `test_getter_setter` generated size, which is an exaggerated result as most code will probably have lots of bulk that isn't just the macro code.

Disadvantages:
- Additional level of complexity in the `getter` and `setter` trampolines and accompanying code.
- To keep the `GetSetDefClosure` objects alive, I've added them to the static `LazyTypeObject` inner.
- Very slight performance overhead at runtime (shouldn't be more than an additional pointer read). It's so slight I couldn't measure it.

Overall I'm happy to either merge or close this based on what reviewers think!

### DIFF
--- a/src/impl_/trampoline.rs
+++ b/src/impl_/trampoline.rs
@@ -5,7 +5,7 @@
 
 use std::{
     any::Any,
-    os::raw::{c_int, c_void},
+    os::raw::c_int,
     panic::{self, UnwindSafe},
 };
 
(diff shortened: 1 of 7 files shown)


src/impl_/trampoline.rs
```diff
@@ -65,26 +70,34 @@ trampolines!(
...
 ) -> c_int {
-    // PyO3 doesn't use the closure argument at present.
-    debug_assert!(closure.is_null());
-    trampoline_inner(|py| f(py, slf, value))
+    // Safety: PyO3 sets the closure when constructing the ffi setter so this cast should always be valid
+    let closure: &GetSetDefClosure = &*(closure as *const GetSetDefClosure);
+    // Safety: PyO3 sets the setter field when constructing the ffi setter
+    let setter = match closure.setter {
+        Some(setter) => setter,
+        None => std::hint::unreachable_unchecked(),
```
adamreichold, line 98: I am not happy about the `unreachable_unchecked` here. This is because we use `.or_default()` when constructing setter and getter one after the other, right? Is there no way to avoid these effectively invalid instances of `GetSetDefClosure`? Separate `HashMap`s for setters and getters?

davidhewitt: Agreed.  We can only put a single `*closure` into the `ffi::PyGetSetDef`, but we could encode the three cases with different structs instead of optionality. (Would need a getter-only, setter-only, and both, for a total of 4 trampolines instead of 2.)

I'll push a refactor to achieve that in the coming days.

davidhewitt:
Ok, I've pushed a commit which removes the use of `unreachable_unchecked`. It does this by moving the logic constructing the getters & setters into a builder / enum in `create_type_object.rs`, and picking an appropriate trampoline / closure pair inside match arms.

I've added some TODO where we could add some validation that only one `#[getter]` per property etc, which can be a future PR.

davidhewitt:
> - To keep the `GetSetDefClosure` objects alive, I've added them to the static `LazyTypeObject` inner.

I think this is no longer a disadvantage, as I've added the "name" and "doc" members to this too, which addresses an older TODO to avoid naively leaking those things.

PR #3038, opened 2023-03-11
Add note about garbage collection on PyPy

### LINKED ISSUE #3005: Garbage collection does not work on PyPy
### Bug Description

Garbage collection as described in [in the docs](https://pyo3.rs/v0.18.1/class/protocols.html?highlight=garba#garbage-collector-integration) works on CPython but no PyPy.

I know that PyPy does GC differently than CPython, but I thought API / behavior wise it would work the same. So I'm not sure if this is a PyO3 issue or a PyPy issue, but I'm starting here and we can always transfer the issue to PyPy if we determine that is the problem.

Discovered from https://github.com/pydantic/pydantic-core/pull/416

### Steps to Reproduce

https://github.com/adriangb/misc/actions/runs/4332105542/jobs/7564368491
https://github.com/adriangb/misc/tree/pyo3-pypy

### Backtrace

_No response_

Linux / PyPy 3.9

3.9

1.67

### Your PyO3 version

0.18.0

github actions

adriangb:
Closes #3005

### DIFF
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -407,6 +407,8 @@ impl ClassWithGCSupport {
 }
 ```
 
+> Note: these methods are part of the C API, PyPy does not necessarily honor them. If you are building for PyPy you should measure memory consumption to make sure you do not have runaway memory growth. See [this issue on the PyPy bug tracker](https://foss.heptapod.net/pypy/pypy/-/issues/3899).
+
 [`IterNextOutput`]: {{#PYO3_DOCS_URL}}/pyo3/class/iter/enum.IterNextOutput.html
 [`PySequence`]: {{#PYO3_DOCS_URL}}/pyo3/types/struct.PySequence.html
 [`CompareOp::matches`]: {{#PYO3_DOCS_URL}}/pyo3/pyclass/enum.CompareOp.html#method.matches


guide/src/class/protocols.md
```diff
@@ -410,3 +410,5 @@ impl ClassWithGCSupport {
 [`IterNextOutput`]: {{#PYO3_DOCS_URL}}/pyo3/class/iter/enum.IterNextOutput.html
 [`PySequence`]: {{#PYO3_DOCS_URL}}/pyo3/types/struct.PySequence.html
 [`CompareOp::matches`]: {{#PYO3_DOCS_URL}}/pyo3/pyclass/enum.CompareOp.html#method.matches
```
davidhewitt, line 412: I _think_ it's required that these markdown links go at the bottom of the document, i.e. the note needs to go above them?

guide/src/class/protocols.md
```diff
@@ -410,3 +410,5 @@ impl ClassWithGCSupport {
...
+
+> Note: these methods are part of the C API, PyPy does not necessarily honor them. If you are building for PyPy you should measure memory consumption to make sure you do not have runaway memory growth.
```
davidhewitt, line 414: It would be useful to add a link to the PyPy issue tracker here so that readers can click through and see the current state of things and explanation.

```suggestion
> Note: these methods are part of the C API, PyPy does not necessarily honor them. If you are building for PyPy you should measure memory consumption to make sure you do not have runaway memory growth. See [this issue on the PyPy bug tracker](https://foss.heptapod.net/pypy/pypy/-/issues/3899).
```

davidhewitt (commented):
Thanks, definitely worth adding this note. Couple of small suggestions...

guide/src/class/protocols.md
adriangb: moved just in case

davidhewitt (approved):
Thanks (I force-pushed to squash)

bors r+

PR #3042, opened 2023-03-12
add `as_tuple()` method to `PyList`

samuelcolvin:
From [benchmarks](https://github.com/samuelcolvin/rust-bench/blob/25fcf98e125e94e46459b2adfbc297b37969e6ba/benches/main.rs#L165-L192), this is significantly faster than `PyTuple::new(py, the_list)`:

```
test list_as_tuple_direct        ... bench:         299 ns/iter (+/- 51)
test list_as_tuple_iterate       ... bench:         521 ns/iter (+/- 41)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ valgrind-python.supp
 *.pyd
 lcov.info
 netlify_build/
+# pycharm files
+/.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -23,5 +23,3 @@ valgrind-python.supp
 *.pyd
 lcov.info
 netlify_build/
-# pycharm files
-/.idea/
--- a/newsfragments/3042.added.md
+++ b/newsfragments/3042.added.md
@@ -1,0 +1,1 @@
+Add `as_tuple()` method to `PyList`, to more efficiently convert a lists to a tuples.
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -7,7 +7,7 @@ use std::convert::TryInto;
 use crate::err::{self, PyResult};
 use crate::ffi::{self, Py_ssize_t};
 use crate::internal_tricks::get_ssize_index;
-use crate::types::PySequence;
+use crate::types::{PySequence, PyTuple};
 use crate::{AsPyPointer, IntoPyPointer, Py, PyAny, PyObject, Python, ToPyObject};
 
 /// Represents a Python `list`.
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -23,3 +23,5 @@ valgrind-python.supp
 *.pyd
 lcov.info
 netlify_build/
+# pycharm files
+/.idea/
```
davidhewitt, line 27: Personally I think that this is irrelevant for this `.gitignore`, e.g. we could add ignores for `.vscode` and every editor under the sun, but we don't. I think it seems more appropriate to add this to your global `.gitignore`, would you be amenable to removing it here (and adding it to your global `.gitignore` if suits)?

davidhewitt (commented):
Looks good other than the change on the `.gitignore` which I'm not convinced is necessary!

.gitignore
birkenfeld: :+1:

samuelcolvin: done.

davidhewitt (approved):
Thanks!

bors r+

mejrs:
I would prefer a `From<&PyList> for &PyTuple` implementation.

samuelcolvin:
> I would prefer a `From<&PyList> for &PyTuple` implementation.

I see that but it would be less obvious what's going on, and why the reverse can't be done.

davidhewitt:
> I would prefer a `From<&PyList> for &PyTuple` implementation.

Actually this is a great point, I think the reverse could be implemented by `tuple.as_sequence().list()`. However I would like to ask we put this on the backlog for the moment just because with reboot of #1308 there would be ownership difference and then it's less clear if the implementation should be `From<&'_ PyList<'py>> for PyTuple<'py>` or `From<PyList<'py>> for PyTuple<'py>` (the difference being whether a reference to the `PyList` is borrowed or consumed).

However this also makes me notice another snag, which is naming convention. This is `as_tuple()` (and matches the C API name) but in Rust the `as_` methods are usually zero-cost conversions, with `to_` conversions being more costly e.g. `to_string` vs `as_str`. So I think there's a few considerations:

- Call this just `.tuple()` to match `PySequence::tuple()`.
- Call this `to_tuple()` and rename sequence methods to `PySequence::to_tuple()` and `PySequence::to_list()`.
  - I'm tempted to open a PR to rename the sequence methods anyway, I think the `to_` prefix may be strictly better.
- Revert this and recommend just using the sequence API. It looks like `PySequence_Tuple`[ has a special case to call `PyList_AsTuple`](https://github.com/python/cpython/blob/bb396eece44036a71427e7766fbb8e0247373102/Objects/abstract.c#L2044) so there would be negligible perf difference.

mejrs:
The `to_*` renaming seems like a good idea

samuelcolvin:
👍 for `to_*`.

davidhewitt:
Pushed #3043 and I'll add separate PRs for other renaming and also `PyTuple::to_list` when I get a moment.

PR #3043, opened 2023-03-12
`PyList::as_tuple()` -> `PyList::to_tuple()`

davidhewitt:
As discussed as a follow-up to #3042

PR #3044, opened 2023-03-12
Add `PyTuple::to_list`

davidhewitt:
Companion to #3043. I've included benchmarks which suggests that this method is indeed about 25% faster on my machine.

### DIFF
--- a/src/types/tuple.rs
+++ b/src/types/tuple.rs
@@ -6,6 +6,7 @@ use crate::ffi::{self, Py_ssize_t};
 #[cfg(feature = "experimental-inspect")]
 use crate::inspect::types::TypeInfo;
 use crate::internal_tricks::get_ssize_index;
+use crate::types::PyList;
 use crate::types::PySequence;
 use crate::{
     exceptions, AsPyPointer, FromPyObject, IntoPy, IntoPyPointer, Py, PyAny, PyErr, PyObject,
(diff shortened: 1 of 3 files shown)


src/types/tuple.rs
```diff
@@ -213,6 +214,15 @@ impl PyTuple {
...
     }
+
+    /// Return a new list containing the contents of this tuple; equivalent to the Python expression `list(tuple)`.
+    ///
+    /// This method is equivalent to `self.as_sequence().list()` and faster than `PyList::new(py, self)`.
+    pub fn to_list(&self) -> &PyList {
+        self.as_sequence()
+            .list()
+            .expect("failed to convert tuple to list")
+    }
```
mejrs, line 225: (I think) you can do  `PyList::new(py, self.as_slice())` on `not(Py_LIMITED_API)`

davidhewitt: Oh great idea, I'll try that when I get a free moment and see how it performs...

davidhewitt: Hmm unfortunately that performs very similarly to `PyList::new(py, self)` (i.e. slower than going through the sequence API), so the bottleneck must be in constructing the list on the Rust side. An investigation for another time, perhaps...

davidhewitt:
Based on #3043 having already gone in, I'm just going to proceed to merge this so I can include it in a rebased 0.18.2.

bors r+

Issue #2863, opened 2023-01-06
Improve default values for automatically-generated `text_signature`


davidhewitt:
This issue is a follow-up to #2784 to remember to improve the formatting of default values in the generated `text_signature`.

Currently all default values are just rendered as `...`, however Rust integer/boolean/string literals  and `None` should be easy enough to format a Python representation for.

Support for more general Rust literals in the Python signature might need a larger refactoring to do more of the signature generation at runtime using traits.

Both of these will be easier after the signature code is simplified, i.e. once the deprecated `#[args]` attribute is removed and maybe also #2703 is merged.

PR #3050, opened 2023-03-16
Improve default values for str, numbers and bool in `text_signature`

messense:
xref #2863

### DIFF
--- a/pyo3-macros-backend/src/pyfunction/signature.rs
+++ b/pyo3-macros-backend/src/pyfunction/signature.rs
@@ -585,6 +585,29 @@ impl<'a> FunctionSignature<'a> {
         }
     }
 
+    fn default_value_for_parameter(&self, parameter: &str) -> String {
+        let mut default = "...".to_string();
+        if let Some(fn_arg) = self.arguments.iter().find(|arg| arg.name == parameter) {
+            if let Some(syn::Expr::Lit(syn::ExprLit { lit, .. })) = fn_arg.default.as_ref() {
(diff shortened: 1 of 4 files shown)


pyo3-macros-backend/src/pyfunction/signature.rs
```diff
@@ -624,8 +647,8 @@ impl<'a> FunctionSignature<'a> {
             output.push_str(parameter);
 
             if i >= py_sig.required_positional_parameters {
-                // has a default, just use ... for now
-                output.push_str("=...");
+                output.push('=');
+                output.push_str(&self.default_value_for_parameter(parameter));
```
davidhewitt, line 651: So because `self.default_value_for_parameter` does a search of the arguments by name, this is O(n²), but only for small n, so it's probably fine.

I was planning to avoid that by storing the default expressions in `self.python_signature`, but that was harder with the deprecated args attribute still present.

davidhewitt (commented):
Nice! I think this is technically a breaking change, so will land in 0.19.

pyo3-macros-backend/src/pyfunction/signature.rs
messense: > So because `self.default_value_for_parameter` does a search of the arguments by name, this is O(n²), but only for small n, so it's probably fine.

Yeah, I think the same.

davidhewitt (approved):
Thanks! With 0.18.2 released, let's merge this!

bors r+

Issue #3039, opened 2023-03-11
Cannot create slot (__len__) methods with staticmethod annotation


willstott101:
I was just exploring the expanded macro source of the library I am working on and was curious if I could increase the performance of, say `__len__` by avoiding the reference to `&self`. For context this is a `Vector3(x, y, z)` whose `__len__` will always return `3`.

So to start with I have this:
```rust
#[pymethods]
impl Vector3 {
    fn __len__(&self) -> usize {
        3
    }
}
```
which is fine, but there's a load of code generated around this call.
If I switch to an associated function (still within `#[pymethods]`):
```rust
    fn __len__() -> usize {
        3
    }
```
pyo3 asks me to mark it as a `#[staticmethod]`:
```
error: static method needs #[staticmethod] attribute                                                                                                  
   --> src/vec3.rs:140:5
    |
140 |     fn __len__() -> usize {
    |     ^^
```
However if I add that annotation:
```
error[E0061]: this function takes 0 arguments but 1 argument was supplied                                                                                    
   --> src/vec3.rs:40:6
    |
39  |   #[pymethods]
    |   ------------ argument of type `*mut pyo3::ffi::PyObject` unexpected
40  |   impl Vector3 {
    |  ______^
41  | |     #[new]
42  | |     fn new(x: Option<f64>, y: Option<f64>, z: Option<f64>) -> Self {
43  | |         Vector3(na::Vector3::new(
...   |
140 | |     #[staticmethod]
141 | |     fn __len__() -> usize {
    | |______________^
    |
note: associated function defined here
   --> src/vec3.rs:141:8
    |
141 |     fn __len__() -> usize {
    |        ^^^^^^^
help: remove the extra argument
```
As you can see the generated code is wrong:
```
               unsafe fn __pymethod___len____(_py: _pyo3::Python<'_>,
                    _raw_slf: *mut _pyo3::ffi::PyObject)
                    -> _pyo3::PyResult<_pyo3::ffi::Py_ssize_t> {
                    let _slf = _raw_slf;
                    _pyo3::callback::convert(_py, Vector3::__len__(_slf))
                }
```
So long story short, I think there's a bug here... specifically that slot methods that are also staticmethods try to pass an unsafe object pointer into the static method (at least for `__len__`).

birkenfeld:
I just tested and surprisingly, defining `__len__` as a staticmethod works in pure Python...

davidhewitt:
[...]

PR #3055, opened 2023-03-18
Fix compile error when trying to use static slot methods, fixes: #3039

willstott101:
Whilst having static slot methods is of dubious value in the first place, there's little reason not to support it by bringing the different function call macros more in-line with each other.

I had no idea which test file to use so I just made a new one... `test_methods.rs` has no slots in and the test files that do use slots seem to be specific to their protocol. Let me know if you'd like the test somewhere else.

Thanks!

### DIFF
--- a/tests/test_static_slots.rs
+++ b/tests/test_static_slots.rs
@@ -1,0 +1,57 @@
+#![cfg(feature = "macros")]
+
+use pyo3::exceptions::PyIndexError;
+use pyo3::prelude::*;
+use pyo3::types::IntoPyDict;
+
+use pyo3::py_run;
+
+mod common;
+
+#[pyclass]
+struct Vector3 {
+    elements: [f64; 3],
+}
+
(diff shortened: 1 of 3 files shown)


tests/test_static_slots.rs
```diff
@@ -0,0 +1,57 @@
...
+            elements: [x, y, z],
+        }
+    }
+
+    #[staticmethod]
+    fn __len__() -> usize {
+        3
+    }
+
+    fn __getitem__(&self, idx: isize) -> PyResult<f64> {
```
birkenfeld, line 30: Why not test a "selfless" `__getitem__` as well?

willstott101: yep, good suggestion, pushed tests for a very useful `Count5` class

davidhewitt (approved):
Thanks, this looks correct to me, will include it in 0.18.2

bors r+

PR #3057, opened 2023-03-18
Add feature to make `PanicException` inherits from Python's `Exception`

### LINKED ISSUE #2783: `PanicException` doesn't inherit from base `Exception`
From https://github.com/pola-rs/polars/issues/5606

`PanicException` doesn't inherit from the base python `Exception`, only from `BaseException`.

As used by libraries such as `polars`, `PanicException` is typically something comparable to `ValueError`, rather than something comparable to `SystemExit` or `KeyboardInterrupt` (built-ins which inherit from `BaseException` but not from `Exception`).

Using a library that relies on `pyo3` in some application/service makes error handling problematic, as one most typically adds something like `except Exception` in order to handle errors gracefully while still responding to interrupt signals for example, but such code would not catch errors coming from pyo3-wrapped libraries.

touilleMan:
fix #2783

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,10 @@ pyo3-build-config = { path = "pyo3-build-config", version = "0.18.1", features =
 [features]
 default = ["macros"]
 
+# By default `PanicException` inherits `BaseException`, use this feature to
+# make it inherits `Exception` instead
+panicexception-inherits-exception = []
+
 # Enables pyo3::inspect module and additional type information on FromPyObject
 # and IntoPy traits
 experimental-inspect = []
@@ -110,6 +114,7 @@ full = [
     "eyre",
     "anyhow",
     "experimental-inspect",
+    "panicexception-inherits-exception",
 ]
 
 [[bench]]
--- a/newsfragments/2783.added.md
+++ b/newsfragments/2783.added.md
@@ -1,0 +1,1 @@
+Add `panicexception-inherits-exception` feature to make `PanicException` inherits from Python's `Exception` (instead of `BaseException` by default)
--- a/pytests/Cargo.toml
+++ b/pytests/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 publish = false
 
 [dependencies]
-pyo3 = { path = "../", features = ["extension-module"] }
+pyo3 = { path = "../", features = ["extension-module", "panicexception-inherits-exception"] }
 
 [build-dependencies]
 pyo3-build-config = { path = "../pyo3-build-config" }
--- a/pytests/src/pyfunctions.rs
+++ b/pytests/src/pyfunctions.rs
@@ -63,6 +63,11 @@ fn args_kwargs<'a>(
     (args, kwargs)
 }
 
+#[pyfunction(signature = ())]
+fn panic() {
+    panic!("Dodge this");
+}
+
 #[pymodule]
 pub fn pyfunctions(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
     m.add_function(wrap_pyfunction!(none, m)?)?;
@@ -71,5 +76,6 @@ pub fn pyfunctions(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
     m.add_function(wrap_pyfunction!(simple_kwargs, m)?)?;
     m.add_function(wrap_pyfunction!(simple_args_kwargs, m)?)?;
     m.add_function(wrap_pyfunction!(args_kwargs, m)?)?;
+    m.add_function(wrap_pyfunction!(panic, m)?)?;
     Ok(())
 }
--- a/pytests/tests/test_pyfunctions.py
+++ b/pytests/tests/test_pyfunctions.py
@@ -1,3 +1,4 @@
+import pytest
 from pyo3_pytests import pyfunctions
 
 
@@ -83,3 +84,10 @@ def test_args_kwargs_rs(benchmark):
     rust = benchmark(pyfunctions.args_kwargs, 1, "foo", {1: 2}, bar=4, foo=10)
     py = args_kwargs_py(1, "foo", {1: 2}, bar=4, foo=10)
     assert rust == py
+
+def test_panic():
+    with pytest.raises(BaseException) as exc_info:
+        pyfunctions.panic()
+    assert exc_info.type.__name__ == "PanicException"
+    assert exc_info.type.__bases__ == (Exception, )
+    assert str(exc_info.value) == "Dodge this"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -87,6 +87,9 @@
 //! - `multiple-pymethods`: Enables the use of multiple [`#[pymethods]`](macro@crate::pymethods)
 //! blocks per [`#[pyclass]`](macro@crate::pyclass). This adds a dependency on the [inventory]
 //! crate, which is not supported on all platforms.
+//! - `panicexception-inherits-exception`: This will make [`PanicException`] inherits
+//! [`Exception`](https://docs.python.org/3/library/exceptions.html#Exception) (instead
+//! of [`BaseException`](https://docs.python.org/3/library/exceptions.html#BaseException))
 //!
 //! The following features enable interactions with other crates in the Rust ecosystem:
 //! - [`anyhow`]: Enables a conversion from [anyhow]’s [`Error`][anyhow_error] type to [`PyErr`].
--- a/src/panic.rs
+++ b/src/panic.rs
@@ -1,19 +1,32 @@
 //! Helper to convert Rust panics to Python exceptions.
+#[cfg(not(feature = "panicexception-inherits-exception"))]
 use crate::exceptions::PyBaseException;
+#[cfg(feature = "panicexception-inherits-exception")]
+use crate::exceptions::PyException;
 use crate::PyErr;
 use std::any::Any;
 
-pyo3_exception!(
-    "
+/// Helper macro to factorize PanicException's documentation
+#[macro_export]
+macro_rules! panic_exception_doc {
+    () => {
+        "
 The exception raised when Rust code called from Python panics.
 
-Like SystemExit, this exception is derived from BaseException so that
+By default, like SystemExit, this exception is derived from BaseException so that
 it will typically propagate all the way through the stack and cause the
 Python interpreter to exit.
-",
-    PanicException,
-    PyBaseException
-);
+
+However this behavior can be changed by enabling the `panicexception-inherits-exception`
+feature: this exception will then derived from Exception.
+"
+    };
+}
+
+#[cfg(not(feature = "panicexception-inherits-exception"))]
+pyo3_exception!(panic_exception_doc!(), PanicException, PyBaseException);
+#[cfg(feature = "panicexception-inherits-exception")]
+pyo3_exception!(panic_exception_doc!(), PanicException, PyException);
 
 impl PanicException {
     /// Creates a new PanicException from a panic payload.


davidhewitt:
Sorry for being late to the discussion. I have had an ill family only recently getting better.

I'm not enthusiastic about doing this. I appreciate the usability concerns however it is also well established in Rust that a panic is not intended to be used for error handling. A panic is a last resort to crash the program when there are programmer errors which have broken invariants, implying bugs.

In #2783 you give some examples of use cases where panics are undesirable which could equally apply regardless of whether the code is written in Rust or Python. What are the source of the panics? Presumably this is in your control, either in your own code or in your dependencies.

I agree with you that a `PanicException` per-module is awkward. In the long term pyo3 should have a Python package with a central `PanicException` which can be imported and handled once. At some point we just need to bite the bullet and commit to stability for some APIs, I think.

Perhaps if we considered these two approaches first, we can solve the pain points for your use case without needing to water down the mapping of Rust error handling to Python at this boundary?

mejrs:
Per my comments in https://github.com/PyO3/pyo3/issues/2783, I think we should not do this.

touilleMan:
> In https://github.com/PyO3/pyo3/issues/2783 you give some examples of use cases where panics are undesirable which could equally apply regardless of whether the code is written in Rust or Python. What are the source of the panics? Presumably this is in your control, either in your own code or in your dependencies.

The point of panics is they are unexpected and can come from various part of the application (typically the "I unwrap this because it can never fail" assumption that end up being wrong 😄 ). So there it's most likely the panic comes from the application code, but it can also come from a dependency, or even from the OS.

But again this is irrelevant, having an unexpected error is something any program in any language can encounter, and being able to recover from it in a controlled way is standard feature for any memory safe language.

Rust has [a concept of UnwindSafe](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html) to address this problem, I don't see what is the need for Pyo3 to think it knows better than it users about their own application.
Recover from panic is tricky, yes. But recover from panic with a library that raises weird exceptions which type is not exposed and who inherits `BaseException`  is yet another layer of suffering 😢.

@mejrs see the end of my comment https://github.com/PyO3/pyo3/issues/2783#issuecomment-1449824967, could you share an example of the buggy unsafe code you've written ?
Considering how trivial this PR is (it's basically 10 lines when test and documentation is excluded) vs how crazy complicated handling `PanicException` like a regular exception currently is, I'd like to see real world examples demonstrating that panic recovery is NEVER the right thing to do.

birkenfeld:
Per my comments in https://github.com/PyO3/pyo3/issues/2783, this might be a good idea :)

touilleMan:
> Please remember that there are two "problems" here:
>    panicking from rust creates an exception that inherits from baseexception
>    this exception is hard to import/name

This PR is only related to the inheritance of `BaseException` problem.

> > I don't see what is the need for Pyo3 to think it knows better than it users about their own application.
>
> Can you clarify what you mean with this?

A good example is you previous sentence:

> The right fix for that is to submit PRs/issues to the library to fix their bugs, or to use a library that doesn't have them.

1. My application requires not blowing up if an unexpected error occurs [^1]. This is a business requirement so PyO3 has no say on it 😄 
2. Rust & Python are memory safe language that provide mechanism to do unexpected error recovery.

My application is capable of doing unexpected error recovery that works for most case and provide crash report upstream. Then of course I shall work on a fix and release a new version.

But you consider the "right fix" is only to "fix the bugs" (or "use a library that doesn't have [bugs]"...). But I can't even fix the bugs because the application closed before crash report, on top of that I have angry users that report me that "the app keeps crashing" 😭

[^1]: Obviously this is a best effort requirement. But basically 99% of bugs are more "I made a typo in the code" than "CPython has segfaulted", so not blowing up on the typo bugs is a huge win.

> It's a new feature, which isn't trivial.

Have you read the code ? This is literally 10 lines once the comments and tests are excluded. Most of the PR is adding a test because raising `PanicException` wasn't currently tested.

Again @mejrs I provided you with a real life use cases of an application currently in production. If you consider this PR is not the right approach, please tell me what could be done instead to answer my needs.

davidhewitt:
Fully agree with the position that adding a feature does not seem to me to be the right way to solve your concern. Adding a feature would then require you to go to _every_ package compiled with PyO3 which you use and request that they change their package to be shipped with this feature enabled (or you compile patched versions of all these packages in house). Is that really better than asking those packages to go fix their bugs?

I understand you have production concerns, and I'm keen we explore ways to fix them here.

As well as improving access to importing `PanicException` (which we all agree we should do), how about another alternative: create an environment variable `PYO3_PANIC_EXCEPTION_EXTENDS_EXCEPTION`, which effects the change in behaviour implied here?

I don't love the idea of changing the type hierarchy based on an environment variable (Python type checkers would probably hate it), however it would have the property of solving your use-case by allowing you to apply a fix at point-of-use rather than on every upstream dependency of yours. 

There may be better solutions than the environment variable too. So I'd prefer we explore that design space and make progress on making `PanicException` easier to reach before we resort to adding this feature.

vxgmichel:
Sorry for jumping in this late, but I followed both this PR (and issue #2783) and I got curious about something. It seems like you got caught up in the debate of whether it's a good idea to recover from a panic or not, but I'm not sure that this should be the decisive factor for choosing `Exception` or `BaseException` as a base class for panics.

Many here share the opinion that crashing the program on a panic is a good idea, but this would still be the default behavior even if panics inherit from `Exception`, as far as I understand. I listed 5 examples below, all crashing the program after a panic (regardless of the panic base class):

```python
from structlog import get_logger
from my_binding import maybe_panic, PanicException

logger = get_logger()

# [Example 1]
# Crashes in both cases
maybe_panic(do_panic=True)

# [Example 2]
# Crashes in both cases
try:
    maybe_panic(do_panic=True)
except ValueError:
    logger.exception("Ignoring value errors")
    pass

# [Example 3]
# Crashes in both cases, with the panic properly logged
try:
    maybe_panic(do_panic=True)
except PanicException:
    logger.exception("Logging the panic before exiting")
    exit(1)

# [Example 4]
# Crashes in both cases, with the exception properly logged
# However, `SystemExit` and other base exceptions are going to
# be logged as well which is usually not what we want
try:
    maybe_panic(do_panic=True)
except:
    logger.exception("Logging all exceptions before exiting")
    exit(1)

# [Example 5]
# Crashes in both cases, but panics are not going to be logged.
try:
    maybe_panic(do_panic=True)
except Exception:
    logger.exception("Logging exceptions before exiting")
    exit(1)
```


An argument for inheriting from `Exception` is that example 5 would behave nicer than example 4 in the context of a python code base. On the other hand, an argument against it is that a panic might be caught by mistake by an `except Exception` clause.

However, my intuition is that using `except Exception` in a python code base already implies "we're catching anything so invariants might have been broken, we probably want to log what happened and exit safely". I failed to find a hypothetical example where a panic is caught and ignored by mistake.

Anyway that was my 2 cents, sorry for the noise.

konstin:
to add a different perspective: i'd have preferred to make a panic not even throw an exception but terminate the program. a rust program does not need to panic and should never panic, and if it does the only reasonable way to deal with it is termination, maybe asking the user to file a bug report just before that. `catch_unwind` was never meant as a control flow construct and it's not even guaranteed to work at all, it's e.g. [popular](https://github.com/search?q=path%3ACargo.toml+panic+%3D+%22abort%22&type=code) to compile with `panic = "abort"`. to cite the [`catch_unwind`  docs](https://doc.rust-lang.org/nightly/std/panic/fn.catch_unwind.html):

> It is **not** recommended to use this function for a general try/catch mechanism. The [Result](https://doc.rust-lang.org/nightly/std/thread/type.Result.html) type is more appropriate to use for functions that can fail on a regular basis. Additionally, this function is not guaranteed to catch all panics, see the “Notes” section below.

> Note that this function **might not catch all panics** in Rust. A panic in Rust is not always implemented via unwinding, but can be implemented by aborting the process as well. This function only catches unwinding panics, not those that abort the process.

panic in drop is also [likely to abort](https://doc.rust-lang.org/std/ops/trait.Drop.html#panics).

similarly, python's [`KeyboardInterrupt` docs](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt) warn about catching `KeyboardInterrupt`  since it may leave the program in an inconsistent state:

> **Note:** Catching a [KeyboardInterrupt](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt) requires special consideration. Because it can be raised at unpredictable points, it may, in some circumstances, leave the running program in an inconsistent state. It is generally best to allow [KeyboardInterrupt](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt) to end the program as quickly as possible or avoid raising it entirely. (See [Note on Signal Handlers and Exceptions](https://docs.python.org/3/library/signal.html#handlers-and-exceptions).)

The [panic! docs](https://doc.rust-lang.org/std/macro.panic.html) further say:

> The `panic!` macro is used to construct errors that represent a bug that has been detected in your program. With `panic!` you provide a message that describes the bug and the language then constructs an error with that message, reports it, and propagates it for you.

> ## [Current implementation](https://doc.rust-lang.org/std/macro.panic.html#current-implementation)

> If the main thread panics it will terminate all your threads and end your program with code 101.

this means rust panics do fundamentally break the contract of an `Exception`, which is guaranteed to interact well with the control flow of the program. given that, i don't think it is good to treat a panic like a catchable error and i strongly believe it is not sound to treat it as a normal `Exception`.

touilleMan:
> The problem is not whether the implementation is complex but that a features are a complicated kind of interface

@adamreichold  thanks for pointing this out, I'm not familiar enough with maintaining a popular Rust library, so I trust you on the fact adding a feature is more complicated that it seems 😃

From this, @davidhewitt suggestion to instead introduce a `PYO3_PANIC_EXCEPTION_EXTENDS_EXCEPTION` environ variable seems to solve the issue well.
On top of that, this approach allows to test both behaviors in the python tests (my PR could only test a single behavior given otherwise it would require recompiling the project to enable/disable the feature)

@davidhewitt Do you want me to update this PR to implement this ? this should be fairly simple to do 😃 

> panic in drop is also [likely to abort](https://doc.rust-lang.org/std/ops/trait.Drop.html#panics).
> [...]
> this means rust panics do fundamentally break the contract of an Exception,

I don't see your point here, sure we cannot catch *all* panics, but we can catch the unwinding ones which is already a big win.

> a rust program does not need to panic and should never panic, and if it does the only reasonable way to deal with it is termination, maybe asking the user to file a bug report just before that. 

So after something unexpected occurs, the only thing the program should do is terminate... unless some kind of bug report should be done before that.
So we're back to my point: PyO3 doesn't know how the end program does bug reporting. And making error handling more exotic than it should be is the best way to break it bug reporting code !

So now the application developer has the choice between having to spend extra efforts checking PyO3's rogue exception is correctly handled[^1] or, more realistically, do nothing and live with the risk of having unreported crashes.

[^1]: for instance in my application I had to check the behavior against three different components: PyQt5, Sentry and Quart

Again the zen of Python:

> There should be one-- and preferably only one --obvious way to do it.

What is the obvious way of catching the panic exception to do bug reporting before terminating the program ?
Once we have acknowledge the need of handling panic exception to do bug reporting, what is the benefit of not inheriting `Exception` ?

mejrs:
> I don't see what is the need for Pyo3 to think it knows better than it users about their own application.

> PyO3 doesn't know how the end program does bug reporting.

You have brought this up a few times, but this is exactly why the current behavior is what it is. Pyo3 doesn't know the user or what they're doing, so the safest option is to treat a bug as fatal. It shouldn't "know better" and try to make the program continue at all cost. The correct solution is that *someone* (be it the end user or library author) should fix their bugs.

I'm also not in favor of an environment variable. They have issues too, just different ones than features. Furthermore I'm reluctant to add things like this that subtly change behavior. They open up a whole can of worms such as people compiling and distributing extensions that are compiled with settings that the original author never intended.

davidhewitt:
Agree that I'm not keen on the environment variable either, for the reasons highlighted by @mejrs. Libraries which panic may not expect the panic to be caught, so _especially_ making it conditional makes it easy to add bugs which the crate author didn't anticipate by enabling this. At least the footgun would be squarely in the control of users rather than distributors, which makes pointing the finger of blame easier, however I would prefer we didn't add a footgun at all.

My reason to suggest the environment variable is to show there is potentially an even better design which we haven't thought of yet.

I would prefer that we work on the improvements which we know about and all agree are needed, such as improving access to the `PanicException` type. In the meanwhile we can allow ourselves time to come up with alternative ideas which may be agreeable to everyone.

adamreichold:
> such as improving access to the PanicException type.

Should we open an issue for a capsule-based C-ABI-compatible interface for accessing the type objects at least for exceptions? (I wonder if this needs ABI at all, i.e. do we ever rely on a `PanicException` being our DSO's `PanicException` or would it be sufficient to publish some type object as a attribute of globally available module?)

birkenfeld:
Do you mean, the first PyO3-based extension would create the `PanicException` class and put it e.g. in `sys._pyo3_PanicException` (or even a custom module `pyo3`), and others would reuse that one? Sounds like an interesting idea.

A similar idea would be to ship an explicit `pyo3` module on PyPI (I just emailed them to request keeping it available for us), which modules can depend on, and which can provide this (and maybe other common APIs) in a well known location, even if it's still generated internally by extensions.

adamreichold:
> Do you mean, the first PyO3-based extension would create the `PanicException` class and put it e.g. in `sys._pyo3_PanicException` (or even a custom module `pyo3`), and others would reuse that one? Sounds like an interesting idea.

Yes, this is basically what I would be trying first.
 
> A similar idea would be to ship an explicit `pyo3` module on PyPI (I just emailed them to request keeping it available for us), which modules can depend on, and which can provide this (and maybe other common APIs) in a well known location, even if it's still generated internally by extensions.

This would yield cleaner code, but my personal experience with Python dependency management makes me prefer to "zero effort" solution outlined above where whatever a PyO3-based extension is initialized, it checks if a global `PanicException` needs to be registered.

(I also wonder whether these approaches are actually exclusive. Couldn't we built a global `pyo3` module on the fly or does that require a file-system-backed structure somewhere?)

adamreichold:
> the one challenge with it is to make sure the type has the same layout across all pyo3 versions and modules using it and/or detect if there are incompatibilities.

If we go for a capsule-based API, we could instead provide a constructor which always takes the same data, e.g. a stringly-typed panic message, and produces an exception backed by the layout of the PyO3-based extension which registered the capsule.

We do something [like that](https://github.com/PyO3/rust-numpy/blob/main/src/borrow/shared.rs) in rust-numpy where the implementation of the capsule can be changed/evolved arbitrarily as long as the capsule API stays compatible (which implied that it provides a very narrow interface).

birkenfeld:
Sure, technically it's easy to generate the module at runtime. I'm just considering the increasing reliance on static analysis tools which won't be too happy with that, so an additional "physical" manifestation might be useful.

konstin:
A `__init__.py` in a `pyo3` package should work for that:

```
class PanicException(BaseException):
    """Use this class to catch panic exceptions coming """
    ...
```

adamreichold:
I opened #3073 to start discussing an approach similar to what rust-numpy so we can discuss concrete technical issues and decide whether this is worth pursuing to improve the `PanicException` ergonomics.

touilleMan:
> My reason to suggest the environment variable is to show there is potentially an even better design which we haven't thought of yet.

@davidhewitt I have yet another solution to allow the end user of PyO3 to opt in to `PanicException` behaving in a Pythonic way:

```rust
// Function to call once asap when the pyo3 module is initializing
fn patch_panic_exception_to_inherit_exception(py: Python) {
    let mut panic_exception_cls = <pyo3::panic::PanicException as pyo3::PyTypeInfo>::type_object(py)
        .as_ptr() as *mut pyo3::ffi::PyTypeObject;
    let exception_cls = <pyo3::exceptions::PyException as pyo3::PyTypeInfo>::type_object(py);
    let new_bases = pyo3::types::PyTuple::new(py, [exception_cls]);
    // SAFETY: `tp_mro` is a pointer to a tuple once the exception structure has been
    // initialized (which is done lazily the first time pyo3 accesses `PanicException`)
    let mro = unsafe { pyo3::types::PyTuple::from_borrowed_ptr(py, (*panic_exception_cls).tp_mro) };
    let new_mro = pyo3::types::PyTuple::new(
        py,
        [
            // 1. Take `PanicException`
            mro.get_item(0).expect("PanicException has 3 items mro"),
            // 2. Add `Exception`
            exception_cls,
            // 3. Take `BaseException` (as `Exception` inherits from it)
            mro.get_item(1).expect("PanicException has 3 items mro"),
            // 4. Take `<class 'object'>`
            mro.get_item(2).expect("PanicException has 3 items mro"),
        ],
    );
    // SAFETY: `tp_base/tp_bases/tp_mro` are pointers that in theory should not be modified
    // once the exception has been initialized. But this is fine as long as `PanicException`
    // has not yet been used (which is the case here since we are initializing the pyo3 module)
    unsafe {
        (*panic_exception_cls).tp_base = exception_cls.into_ptr() as *mut pyo3::ffi::PyTypeObject;
        (*panic_exception_cls).tp_bases = new_bases.into_ptr();
        (*panic_exception_cls).tp_mro = new_mro.into_ptr();
    }
}
```
(full PR in my own project: https://github.com/Scille/parsec-cloud/pull/4337)

Obviously I'm not thrilled by this solution since it introduce unsafe and hot patching, but I'd rather have a single well-signaled complex part than having to pray every third party error handler in my codebase is able to deal with `BaseException`.

The good think about this approach is has zero impact on PyO3 codebase.

Do you think I would be worth it to provide this helper or at least to document it in PyO3 documentation (among with the rational and pros of having `PanicException` inheriting `BaseException`) ?

@mejrs since you closed #2783 saying `Discussion [...] is still welcome, but please make a new issue/PR for doing so.` I didn't post this solution there, however I believe it would be good to add it so that people with similar need can find a working solution, what do you think ?

adamreichold:
> I have yet another solution to allow the end user of PyO3 to opt in to `PanicException` behaving in a Pythonic way:

Doesn't still leave the issue that every extension using PyO3 would need to call `patch_panic_exception_to_inherit_exception` as each has its own type object?

I think the main benefit of the environment variable was to ensure during deployment that every PyO3 user will make the switch. (Also a significant downside as it enables behaviour that extensions authors do not expect as @mejrs pointed out.)

DataTriny:
Hello @touilleMan 

Thank you!

touilleMan:
@DataTriny I've expressed my consent on #3108 :smiley: 

> Doesn't still leave the issue that every extension using PyO3 would need to call patch_panic_exception_to_inherit_exception as each has its own type object?

@adamreichold 

Yes, however this is kind of a good thing here: this fix works because the panic exception is patched before behind used. If panic exception is shared across multiple pyo3 extensions, then the extension author cannot do this patch given 1) it may impact other modules and 2) it may modify an already used exception which may leave to undefined behavior.

So if the panic exception is shared, it's up to the final application developer (as opposed to pyo3 dev and 3rd party module native dev that rely on pyo3) to add the patch to there code base (base creating a pyo3 module that does just that and load it before any other pyo3 module)

@davidhewitt @mejrs I'm still interested in your opinion on my fix ;-)

davidhewitt:
I think it's fine if you want to be responsible for applying that patch in your own application. Given we have agreement we want to centralise the PanicException into a `pyo3` namespace, I think it's best not to encourage people more generally to apply that patch or they will end up interfering with other extensions' code.

Issue #3059, opened 2023-03-20
Comparing enums by identity


ricohageman:
In python the convention is to compare enums by identity (`Enum.A is Enum.A` and `Enum.A is not Enum.B`) instead of equality (`Enum.A == Enum.A` and `Enum.A != Enum.B`). However, a straight forward implementation of enums in rust exposed to python using pyo3 does not support this. [The documentation](https://pyo3.rs/v0.18.1/class#pyclass-enums) also only mentions comparison by equality. 

For the particular use-case I'm working with, it is used like the following:

```rust
#[pyclass]
pub enum PyEnum {
   A,
   B,
   C,
}

#[pyclass(get_all)]
pub struct PyEnumHoldingStruct {
   py_enum: PyEnum,
   ...
}

#[pymethods]
impl PyEnumHoldingStruct {
   #[new]
   pub fn new(py_enum: PyEnum, ...) -> Self {
      Self {
         py_enum,
         ...
      }
}

#[pymodule]
fn py_module(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<PyEnum>()?;
    m.add_class::<PyEnumHoldingStruct>()?;
}
```

```python
import py_module

py_a = py_module.PyEnumHoldingStruct(
   py_module.PyEnum.A,
   ...
)

assert py_a.py_enum is py_module.PyEnum.A
assert py_a.py_enum is not py_module.PyEnum.B
assert py_a.py_enum is not py_module.PyEnum.C
```

In https://github.com/PyO3/pyo3/issues/2384 there is a description on how to use `GILOnceCell` and `#[classattr]` to return pointers to the same instance of an enum such that one can compare the enums by identity. Ideally there would be a way to achieve this result without that amount of boilerplate code per enum variant. Would it be possible to automatically generate this means of some setting in `#[pyclass]`?

I would be able to implement this if there is some guidance and a clear idea on how to support it.

davidhewitt:
Hello, I agree with you that this is important; I've been aware of this shortcoming in PyO3's enum implementation and while I've wanted to fix it for a while I haven't had any time available to invest in this particular issue.

Your offer to help implement is appreciated, I'd be happy to feedback on the design choices with you and help steer an implementation.

To point you in the right direction, there's a couple of inter-related pieces which would be worth reading and thinking about:
[...]

Issue #2274, opened 2022-04-04
PEP 489 Compatibility


davidhewitt:
Inspired by https://github.com/PyO3/pyo3/pull/2269#issuecomment-1087901951, this is an issue to track all of the places which we know need to change for full PEP 489 compatibility (draft implementation in #2245).

- Rust statics - `#[pyclass]` type objects, most implementations of `PyTypeInfo`, `intern!`

davidhewitt:
The above is currently incomplete - please post / edit things as you think of them

messense:
Linking some related PEPs here:

* [PEP 489 – Multi-phase extension module initialization](https://peps.python.org/pep-0489/)
* [PEP 573 – Module State Access from C Extension Methods](https://peps.python.org/pep-0573/)
* [PEP 630 – Isolating Extension Modules](https://peps.python.org/pep-0630/)
* [PEP 684 - A Per-Interpreter GIL](https://peps.python.org/pep-0684/)
* [PEP 687 – Isolating modules in the standard library](https://peps.python.org/pep-0687/)

davidhewitt:
Beginning to think that this is something that we may not want to do for now - see https://github.com/PyO3/pyo3/discussions/2346#discussioncomment-2678241

davidhewitt:
@bazaah following up from https://github.com/PyO3/pyo3/pull/5525#issuecomment-3450241874

Let's use this issue to track follow-up work.

Defining a way to set per-module state as part of the `#[pymodule]` macro will allow users to get a long way towards isolating their own modules. There is potentially a lot of work we need to do inside various function macro machinery to make that state accessible.

One other alternative to removing statics is to have interpreter-safe datastructures (e.g. store multiple values in some weakly-referenced form, indexed by interpreter ID).

bazaah:
Yeah, my experimenting so far has lead me to the realization that exposing generics sanely while interacting with `PyModule_GetState` and `#[pymodule]` is pretty difficult to expose via an (e.g) `PyModuleState<T>`, as we have to erase and then "remember" the `T` later, inside the closure callbacks. I'm currently exploring a typemap approach instead, which is worse ergonomically, but shields pyo3 from having to care about the user's state `T` at all.

I haven't yet really started to look at what `static`s will need to be updated / moved. In general I think it will be pretty difficult to handle `m_traverse` stuff without storing stuff directly in the module state, but we probably just don't have to do that immediately either.

bazaah:
[...]

PR #3061, opened 2023-03-24
Enable comparison of C-like enums by identity

ricohageman:
Fix https://github.com/PyO3/pyo3/issues/3059 by changing the `IntoPy` implementation of enums.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,55 @@ To see unreleased changes, please see the [CHANGELOG on the main branch guide](h
 
 <!-- towncrier release notes start -->
 
+## [0.24.0] - 2025-03-09
+
+### Packaging
+
+- Add supported CPython/PyPy versions to cargo package metadata. [#4756](https://github.com/PyO3/pyo3/pull/4756)
+- Bump `target-lexicon` dependency to 0.13. [#4822](https://github.com/PyO3/pyo3/pull/4822)
+- Add optional `jiff` dependency to add conversions for `jiff` datetime types. [#4823](https://github.com/PyO3/pyo3/pull/4823)
+- Bump minimum supported `inventory` version to 0.3.5. [#4954](https://github.com/PyO3/pyo3/pull/4954)
+
+### Added
+
+- Add `PyIterator::send` method to allow sending values into a python generator. [#4746](https://github.com/PyO3/pyo3/pull/4746)
+- Add `PyCallArgs` trait for passing arguments into the Python calling protocol. This enabled using a faster calling convention for certain types, improving performance. [#4768](https://github.com/PyO3/pyo3/pull/4768)
+- Add `#[pyo3(default = ...']` option for `#[derive(FromPyObject)]` to set a default value for extracted fields of named structs. [#4829](https://github.com/PyO3/pyo3/pull/4829)
+- Add `#[pyo3(into_py_with = ...)]` option for `#[derive(IntoPyObject, IntoPyObjectRef)]`. [#4850](https://github.com/PyO3/pyo3/pull/4850)
+- Add uuid to/from python conversions. [#4864](https://github.com/PyO3/pyo3/pull/4864)
+- Add FFI definitions `PyThreadState_GetFrame` and `PyFrame_GetBack`. [#4866](https://github.com/PyO3/pyo3/pull/4866)
+- Optimize `last` for `BoundListIterator`, `BoundTupleIterator` and `BorrowedTupleIterator`. [#4878](https://github.com/PyO3/pyo3/pull/4878)
+- Optimize `Iterator::count()` for `PyDict`, `PyList`, `PyTuple` & `PySet`. [#4878](https://github.com/PyO3/pyo3/pull/4878)
+- Optimize `nth`, `nth_back`, `advance_by` and `advance_back_by` for `BoundTupleIterator` [#4897](https://github.com/PyO3/pyo3/pull/4897)
+- Add support for `types.GenericAlias` as `pyo3::types::PyGenericAlias`. [#4917](https://github.com/PyO3/pyo3/pull/4917)
+- Add `MutextExt` trait to help avoid deadlocks with the GIL while locking a `std::sync::Mutex`. [#4934](https://github.com/PyO3/pyo3/pull/4934)
+- Add `#[pyo3(rename_all = "...")]` option for `#[derive(FromPyObject)]`. [#4941](https://github.com/PyO3/pyo3/pull/4941)
+
+### Changed
+
+- Optimize `nth`, `nth_back`, `advance_by` and `advance_back_by` for `BoundListIterator`. [#4810](https://github.com/PyO3/pyo3/pull/4810)
+- Use `DerefToPyAny` in blanket implementations of `From<Py<T>>` and `From<Bound<'py, T>>` for `PyObject`. [#4593](https://github.com/PyO3/pyo3/pull/4593)
+- Map `io::ErrorKind::IsADirectory`/`NotADirectory` to the corresponding Python exception on Rust 1.83+. [#4747](https://github.com/PyO3/pyo3/pull/4747)
+- `PyAnyMethods::call` and friends now require `PyCallArgs` for their positional arguments. [#4768](https://github.com/PyO3/pyo3/pull/4768)
+- Expose FFI definitions for `PyObject_Vectorcall(Method)` on the stable abi on 3.12+. [#4853](https://github.com/PyO3/pyo3/pull/4853)
+- `#[pyo3(from_py_with = ...)]` now take a path rather than a string literal [#4860](https://github.com/PyO3/pyo3/pull/4860)
+- Format Python traceback in impl Debug for PyErr. [#4900](https://github.com/PyO3/pyo3/pull/4900)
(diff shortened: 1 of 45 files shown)


pyo3-macros-backend/src/pyclass.rs
```diff
@@ -512,6 +531,57 @@ fn impl_enum(
     impl_enum_class(enum_, args, doc, methods_type, krate)
 }
 
+fn impl_into_py_enum(cls: &syn::Ident, variants: &[PyClassEnumVariant<'_>]) -> TokenStream {
+    // Assuming all variants are unit variants because they are the only type we support.
+    match variants.len() {
+        0 => quote! {},
+        1 => {
```
RicoHagemanPlotwise, line 538: What to do with the "no variants" case? It should never even reach here as it errors earlier already on `#[pyclass] can't be used on enums without any variants`. 

Though maybe it's not needed to branch on the number of variants. I believe that the general case could/should also work when there is just a single variant. But this is my first time writing macros and I could not get it to work.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -615,6 +685,7 @@ fn impl_enum_class(
         methods_type,
         enum_default_methods(cls, variants.iter().map(|v| (v.ident, v.python_name()))),
         default_slots,
+        Some(impl_into_py_enum(cls, &variants)),
```
RicoHagemanPlotwise, line 688: We're changing the `IntoPy` implementation of enums. Should we maybe put this behind some flag on the `pyclass` level? Or is adhering to the python implementation of enums sufficient to change this without such control?

pyo3-macros-backend/src/pyclass.rs
ricohageman, line 538: What to do with the "no variants" case? It should never even reach here as it errors earlier already on #[pyclass] can't be used on enums without any variants.

Though maybe it's not needed to branch on the number of variants. I believe that the general case could/should also work when there is just a single variant. But this is my first time writing macros and I could not get it to work.

pyo3-macros-backend/src/pyclass.rs
ricohageman, line 688: We're changing the `IntoPy` implementation of enums. Should we maybe put this behind some flag on the `pyclass` level? Or is adhering to the python implementation of enums sufficient to change this without such control?

ricohageman:
I wonder what changed that all of a sudden the dependencies failed to be installed

pyo3-macros-backend/src/pyclass.rs
davidhewitt: It might sufficient to have the special case for `1` and then let `0` be handled by the `_` case?

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -512,6 +531,57 @@ fn impl_enum(
...
+fn impl_into_py_enum(cls: &syn::Ident, variants: &[PyClassEnumVariant<'_>]) -> TokenStream {
+    // Assuming all variants are unit variants because they are the only type we support.
+    match variants.len() {
+        0 => quote! {},
+        1 => {
+            let variant_name = variants.first().unwrap().ident;
+            quote! {
+                impl _pyo3::IntoPy<_pyo3::PyObject> for #cls {
+                    fn into_py(self, py: _pyo3::Python) -> _pyo3::PyObject {
+                        static SINGLETON: _pyo3::once_cell::GILOnceCell<_pyo3::Py<#cls>> = _pyo3::once_cell::GILOnceCell::new();
```
davidhewitt, line 543: Use of `GILOnceCell` seems reasonable, however at the same time if we ever want to have a stab at #2274 (which is admittedly a long way off) then it'd be better to not introduce new statics containing Python objects.

I think an alternative implementation could look up variants as attributes from the enum's Python class object, however I think there's potentially a chicken-or-egg problem of how we create those attributes (as I think that currently uses `IntoPy`).

tests/test_enum.rs
```diff
@@ -72,10 +72,59 @@ fn test_enum_eq_incomparable() {
     })
 }
 
+#[test]
+fn test_enum_compare_by_identity() {
```
davidhewitt, line 76: So I noticed a test above `test_enum_class_attr` with the following line:

```rust
Py::new(py, MyEnum::Variant).unwrap();
```

This is potentially a problem; that is a _new_ instance of `MyEnum` type and so it won't compare successfully with identity.

Now, the question is, is that a good thing? Pro could be that it gives users who need it an escape hatch. However, it is probably also a footgun.

Unfortunately, changing how `Py::new` works would require a refactoring of our initialization machinery. I'm sure there's plenty of scope for improvement in that area (e.g. #2384), however it would likely be a much bigger patch than this PR...

pyo3-macros-backend/src/pyclass.rs
davidhewitt: I would rather not add a new flag, I think supporting identity comparison is strictly an improvement. What we would need is the entry in the changelog (which you've already added) and potentially a note in the migration guide.

davidhewitt (commented):
Thanks for putting together this PR! I think there's a couple things we need to think about, however I'm very glad for improvements in this area and so long as we have an idea how we want to move forward in future we don't need to solve everything right away 😄

tests/test_enum.rs
ricohageman: I agree that this possibility is a footgun which we should prevent. Maybe we should put this a bit on hold until the scope for changes to `Py::new` is defined?

davidhewitt: By "this" do you mean solving this particular problem or the whole PR?

ricohageman: I meant the whole pull request but I don't know what the best approach would be. It seems that comparing enums by identity is a desired feature but there are also two concerns https://github.com/PyO3/pyo3/pull/3061#discussion_r1150124372 and https://github.com/PyO3/pyo3/pull/3061#discussion_r1150128264 that would both require quite a refactor. Also happy to help in that regard but I don't know a lot about the existing problems and where to start. And I can imagine you don't want to complicate that work further by including some changes now?

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -512,6 +531,56 @@ fn impl_enum(
...
+                .map(|variant| {
+                    let variant_name = variant.ident;
+                    quote!((#cls::#variant_name as usize, _pyo3::Py::new(py, #cls::#variant_name).unwrap()))
+                })
+                .collect();
+
+            quote! {
+                impl _pyo3::IntoPy<_pyo3::PyObject> for #cls {
+                    fn into_py(self, py: _pyo3::Python) -> _pyo3::PyObject {
+                        static SINGLETON_PER_VARIANT: _pyo3::once_cell::GILOnceCell<::std::collections::HashMap<usize, _pyo3::Py<#cls>>> = _pyo3::once_cell::GILOnceCell::new();
```
alex, line 565: Is there a reason to use a `HashMap` here instead of a fixed-size array?

adamreichold: I think `#cls::#variant_name as usize` could be sparse, e.g.

```
enum Foo {
  Bar = 1,
  Baz = 999999,
}
```

so while we could store those as sorted pairs in a `[(usize, Py<#cls>); 2]`, we would need to do a binary search for the variants instead of a hash table look-up as direct access into a `[(usize, Option<Py<#cls>>); 1000000]` seems wasteful.

Personally, I see the simplicity of using the hash table but would probably opt for the sorted array in a follow-up. (But let's whether this lands using the `static`s at all.)

ricohageman: The discriminant can indeed be sparse which is also tested by `test_custom_discriminant_comparison_by_identity `. But using binary search in a sorted array is definitely a possible replacement for the hashmap.

davidhewitt: I think we can know from parsing the enum whether it's dense or sparse?

For the dense case we can just index into an array, no binary search needed, and for the sparse case we can generate a `match` in front to convert into a dense set of indices first.

adamreichold: > and for the sparse case we can generate a match in front to convert into a dense set of indices first.

This would really be nice as the compiler is pretty clever in how to handle `match` AFAIK, e.g. producing linear or binary searches depending on cardinality estimates.

tests/test_enum.rs
davidhewitt: So we've just had #3287 merged which might be enough of a tweak to make this possible. Given that `Py::new` relies on `Into<PyClassInitializer<T>>` we might be able to adjust that trait implementation for enums to somehow use `PyClassInitializerImpl::Existing` (by fetching the value from a `GILOnceCell` which is itself initialized by a `PyClassInitializerImpl::New` invocation to `Py::new`.

pyo3-macros-backend/src/pyclass.rs
davidhewitt: I think given that we're still no closer to solving the statics problem, and it'll require coming up with new patterns which will probably apply to all statics equally, let's just stick with a static `GILOnceCell` for now and leave migration off that for the future 😆

PR #3062, opened 2023-03-24
Stop panic on `fmt::Display`

### LINKED ISSUE #3060: `py_any.to_string()` should not panic
I've seen seg faults in python due to rust code panicing, the root cause of the panic was `py_thing.to_string()` which panics if `py_thing.__repr__()` raises an error.

I suggest that instead of panicing, it should return `"<unprintable object>"`, similar to [this](https://github.com/pydantic/pydantic-core/blob/dcf1fdd5d6cd16af11fdd3de551c3ac52429b276/src/build_tools.rs#L242-L250) code:

```rs
pub fn safe_repr(v: &PyAny) -> Cow<str> {
    if let Ok(s) = v.repr() {
        s.to_string_lossy()
    } else if let Ok(name) = v.get_type().name() {
        format!("<unprintable {name} object>").into()
    } else {
        "<unprintable object>".into()
    }
}
```

samuelcolvin:
Closes #3060

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -99,8 +99,15 @@ macro_rules! pyobject_native_type_base(
             fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>)
                    -> ::std::result::Result<(), ::std::fmt::Error>
             {
-                let s = self.str().or(::std::result::Result::Err(::std::fmt::Error))?;
-                f.write_str(&s.to_string_lossy())
+                match self.str() {
+                    ::std::result::Result::Ok(s) => return f.write_str(&s.to_string_lossy()),
+                    ::std::result::Result::Err(err) => err.write_unraisable(self.py(), ::std::option::Option::Some(self)),
+                }
+
+                match self.get_type().name() {
+                    ::std::result::Result::Ok(name) => ::std::write!(f, "<unprintable {} object>", name),
+                    ::std::result::Result::Err(_err) => f.write_str("<unprintable object>"),
+                }
             }
         }
 
--- a/newsfragments/3062.fixed.md
+++ b/newsfragments/3062.fixed.md
@@ -1,0 +1,1 @@
+Stop panic on `fmt::Display`, instead return `"<unprintable object>"` string and report error via `sys.unraisablehook()`


src/types/mod.rs
```diff
@@ -99,8 +99,13 @@ macro_rules! pyobject_native_type_base(
             fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>)
                    -> ::std::result::Result<(), ::std::fmt::Error>
             {
-                let s = self.str().or(::std::result::Result::Err(::std::fmt::Error))?;
-                f.write_str(&s.to_string_lossy())
+                if let ::std::result::Result::Ok(s) = self.str() {
+                    f.write_str(&s.to_string_lossy())
+                } else if let ::std::result::Result::Ok(name) = self.get_type().name() {
+                    ::std::write!(f, "<unprintable {} object>", name)
```
davidhewitt, line 105: If `self.str()` fails, it may be nice to try to include the error text in this message so that it helps users understand why the object was unprintable

Not sure exactly what that would look like, something like this maybe?

```
<repr for object of type {type} failed with: {exc_type} {exc_msg}>
```

Would be quite a lengthy repr though...

samuelcolvin: is there any chance that could be a security risk - e.g. leaking information about the failure in a place where you wouldn't expect it?

adamreichold: I fear that if one relies on `repr` to produce externally visible log messages, that concern already applies? Otherwise, if I know the possible types and that their `repr` is safe, I should also know that these errors do not happen, don't I?

samuelcolvin: Well, this is using `str` not `repr`, which might generally be considered "more user facing".

adamreichold: I don't think the distinction matters too much when sensitive information is involved, e.g. an API key stored as a `str` property of a Python class might be part of both its `str` and its `repr`?

davidhewitt: I agree with @adamreichold here that if there is sensitive information then it's the type's responsibility to protect that adequately, not something which this framework can do. I wouldn't see this as a security risk.

Besides, I've had an even better idea: this should probably report the error using `PyErr::write_unraisable`.

adamreichold: @samuelcolvin Would you be up to bending this PR towards `PyErr::write_unraisable`?

samuelcolvin: I'm a bit flat out right now, won't be able to look for a week or two.

I'm also a bit unclear what you need?

adamreichold: Ok, I took the liberty of rebasing this and making use of `PyErr::write_unraisable` to report the error. I think this is ready to merge now and hence also removed the draft status.

adamreichold:
Sorry for the many force pushes: I took a detour adding a new API

```rust
impl Python<'_> {
    pub fn try_with_gil<F, R>(f: F) -> Option<R>
    where
        F: for<'py> FnOnce(Python<'py>) -> R,
    {
        gil::try_ensure_gil().map(|gil| f(unsafe { gil.python() }))
    }
}
```

because I did not realise that I already had implicit access to the GIL in the `Display` impl. Properly removing this and fixing up the news fragment took a bit of back and forth.

davidhewitt (approved):
This looks good to me, thanks!

adamreichold:
bors r=davidhewitt

PR #3066, opened 2023-03-26
Improve default value for `None` in `text_signature`

messense:
xref #2863

### DIFF
--- a/guide/src/function/signature.md
+++ b/guide/src/function/signature.md
@@ -148,7 +148,7 @@ fn increment(x: u64, amount: Option<u64>) -> u64 {
 #             .extract()?;
 #
 #         #[cfg(Py_3_8)]  // on 3.7 the signature doesn't render b, upstream bug?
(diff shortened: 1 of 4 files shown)


davidhewitt (commented):
Thanks!

I think there's another test case to consider, which is what happens if the argument is implicitly defaulted to `None`, as per https://pyo3.rs/v0.18.2/function/signature#trailing-optional-arguments

i.e. this test case:

```rust
#[pyfunction]
fn increment(x: u64, amount: Option<u64>) -> u64 {
    x + amount.unwrap_or(1)
}
```

should have `__text_signature__ == (x, amount=None)`

messense:
Thanks for the review, I think I've addressed it now.

davidhewitt (approved):
Thanks, looks great, sorry I missed your reply earlier!

davidhewitt:
(whoops, looks like bors is fine and I just hadn't noticed another batch is already merging, sorry for the spam 😅)

Issue #1059, opened 2020-07-20
Document passing PyClass values as arguments


davidhewitt:
Just thinking it might be worth adding a table to the guide (`class.md`?) which talks about the different choices for taking `#[pyclass]` values as arguments.

Currently there's `MyClass`, `&MyClass`, `&self`, `&mut self`, `PyRef<MyClass>`, `PyRefMut<MyClass>`, and `&PyCell<MyClass>`.

All have moments when you might want them, but it would help users if this was documented better!

PR #3067, opened 2023-03-26
docs: show pattern for self.__class__.__name__ in __repr__

wjones127:
It took me a little while to figure out this pattern in PyO3, so I thought it would be a good addition to the guide.

It's not the cleanest pattern, so would welcome suggestions on how to make it shorter or easier.

### DIFF
--- a/guide/src/class/numeric.md
+++ b/guide/src/class/numeric.md
@@ -233,8 +233,10 @@ impl Number {
         Self(value)
     }
 
-    fn __repr__(&self) -> String {
-        format!("Number({})", self.0)
+    fn __repr__(slf: &PyCell<Self>) -> PyResult<String> {
+       // Get the class name dynamically in case `Number` is subclassed
+       let class_name: &str = slf.get_type().name()?;
+        Ok(format!("{}({})", class_name, slf.borrow().0))
     }
 
     fn __str__(&self) -> String {
--- a/guide/src/class/object.md
+++ b/guide/src/class/object.md
@@ -61,11 +61,35 @@ impl Number {
         //                       👇 Tuple field access in Rust uses a dot
         format!("Number({})", self.0)
     }
-
     // `__str__` is generally used to create an "informal" representation, so we
     // just forward to `i32`'s `ToString` trait implementation to print a bare number.
     fn __str__(&self) -> String {
         self.0.to_string()
+    }
+}
+```
+
+#### Accessing the class name
+
+In the `__repr__`, we used a hard-coded class name. This is sometimes not ideal,
+because if the class is subclassed in Python, we would like the repr to reflect
+the subclass name. This is typically done in Python code by accessing 
+`self.__class__.__name__`. In order to be able to access the Python type information
+*and* the Rust struct, we need to use a `PyCell` as the `self` argument.
+
+```rust
+# use pyo3::prelude::*;
+#
+# #[pyclass]
+# struct Number(i32);
+#
+#[pymethods]
+impl Number {
+    fn __repr__(slf: &PyCell<Self>) -> PyResult<String> {
+        // This is the equivalent of `self.__class__.__name__` in Python.
+        let class_name: &str = slf.get_type().name()?;
+        // To access fields of the Rust struct, we need to borrow the `PyCell`.
+        Ok(format!("{}({})", class_name, slf.borrow().0))
     }
 }
 ```
@@ -235,8 +259,9 @@ impl Number {
         Self(value)
     }
 
-    fn __repr__(&self) -> String {
-        format!("Number({})", self.0)
+    fn __repr__(slf: &PyCell<Self>) -> PyResult<String> {
+        let class_name: &str = slf.get_type().name()?;
+        Ok(format!("{}({})", class_name, slf.borrow().0))
     }
 
     fn __str__(&self) -> String {


guide/src/class/object.md
```diff
@@ -52,14 +52,25 @@ contained inside `Number`.
...
-    // the `Number`, like `Number(5)` for example.
-    fn __repr__(&self) -> String {
+    // the `Number`, like `Number(5)` for example. If this is subclassed in Python we would
+    // like the class name to reflect the subclass name, so we'll use self.__class__.__name__
+    // to get the name of the class.
+    fn __repr__(self_: PyRef<Self>) -> PyResult<String> {
+        // We need both a reference to the Rust data and the Python data, so we use the 
+        // parameter type PyRef<Self>. This dereferences into Self but also can be converted
+        // into a PyObject.
+        let pyself = self_.clone().into_py(self_.py());
```
wjones127, line 62: Hmm does this only work if `Self` implements clone, meaning this is actually cloning the underlying struct? Is there a way to get a reference to the `PyObject` without moving `self_`?

adamreichold:
Thank you for sharing your findings! I agree that it is not obvious how to approach this. However, I would suggest not burdening this particular example with the complication but to start a new example on how to access the class name.

Concerning the implementation, instead of starting with `PyRef`, I think it would simpler to use `PyCell` which is the actual shape of the Python object and therefore directly derefs into `PyAny` creating another strong reference, e.g.

```diff
diff --git a/guide/src/class/object.md b/guide/src/class/object.md
index 357fdcd7..5b099ed3 100644
--- a/guide/src/class/object.md
+++ b/guide/src/class/object.md
@@ -53,13 +53,17 @@ contained inside `Number`.
 impl Number {
     // For `__repr__` we want to return a string that Python code could use to recreate
     // the `Number`, like `Number(5)` for example.
-    fn __repr__(&self) -> String {
+    fn __repr__(self_: &PyCell<Self>, py: Python<'_>) -> PyResult<String> {
+        let class_name: String = self_
+            .getattr("__class__")?
+            .getattr("__name__")?
+            .extract()?;
         // We use the `format!` macro to create a string. Its first argument is a
         // format string, followed by any number of parameters which replace the
         // `{}`'s in the format string.
         //
         //                       👇 Tuple field access in Rust uses a dot
-        format!("Number({})", self.0)
+        Ok(format!("{class_name}({})", self_.borrow().0))
     }
 
     // `__str__` is generally used to create an "informal" representation, so we
```

However, for this particular issues (getting the name of Rust-implemented Python class), there is simpler way IIUC via our own `PyTypeInfo` trait, e.g.

```diff
diff --git a/guide/src/class/object.md b/guide/src/class/object.md
index 357fdcd7..c2d0865f 100644
--- a/guide/src/class/object.md
+++ b/guide/src/class/object.md
@@ -44,7 +44,7 @@ It can't even print an user-readable representation of itself! We can fix that b
 contained inside `Number`.
 
 ```rust
-# use pyo3::prelude::*;
+# use pyo3::{prelude::*, PyTypeInfo};
 #
 # #[pyclass]
 # struct Number(i32);
@@ -59,7 +59,7 @@ impl Number {
         // `{}`'s in the format string.
         //
         //                       👇 Tuple field access in Rust uses a dot
-        format!("Number({})", self.0)
+        format!("{}({})", Self::NAME, self.0)
     }
 
     // `__str__` is generally used to create an "informal" representation, so we
```

which I would probably prefer to show case. But I am not sure whether we want to guide user's towards these "infrastructure traits" too strongly?

birkenfeld:
The second one is also incorrect in subclasses.

But couldn't the lookup of class and name be done faster with native APIs? `self.get_type().name()`?

adamreichold:
> But couldn't the lookup of class and name be done faster with native APIs? self.get_type().name()?

Indeed! But you'd still have to go via `PyCell<Self>` to access `PyAny::get_type`, e.g. `self_.get_type().name()`, right?

adamreichold:
> I would be in favour of not doing so, I have been pondering a move of PyTypeInfo::NAME and PyTypeInfo::MODULE to PyClassImpl, so I'd prefer not encourage users to access those members :)

Ok, so I think the consensus so far is:
* Use a separate example, specifically on the topic "access the Python-level class object of Rust-defined `#[pyclass]`.
* Use `&PyCell<Self>` as the receiver as it directly derefs to `PyAny`.
* Use the native methods like `PyAny::get_type` instead of access attributes by name.

wjones127:
Thanks for the advice. Btw #1059 would still be really useful; I had no idea you could take `&PyCell<Self>` as a parameter.

guide/src/class/numeric.md
```diff
@@ -233,8 +233,9 @@ impl Number {
         Self(value)
     }
 
-    fn __repr__(&self) -> String {
-        format!("Number({})", self.0)
+    fn __repr__(slf: &PyCell<Self>) -> PyResult<String> {
+        let class_name: &str = slf.get_type().name()?;
```
mejrs, line 237: I think a comment here would be helpful, in case the user hasn't read the preceding chapter
 ```suggestion
        // Get the class name dynamically in case `Number` is subclassed
        let class_name: &str = slf.get_type().name()?;
```

mejrs (approved):
LGTM, just a small nit

adamreichold:
@wjones127 Could you squash your commits before merging? Thanks.

adamreichold:
I force-pushed to remove the usage of inline format arguments.

bors r+

PR #3073, opened 2023-03-28
RFC: Use a capsule-based API with a stable ABI for global, cross-extension functionality.

adamreichold:
This does not work (meaning build) due to multiple issues but I wanted to propose something concrete here so we can discuss whether this roundabout way is something we would want to pursue.

As this it is proposed, this should mean that the first extension built using PyO3 that calls `ensure_global_api` creates the module containing its `PanicException` type object and the corresponding capsule API. All other PyO3-based extensions would then use this capsule to create panic exceptions based on the same internals as first one.

Downstream code should then be able to `import pyo3.PanicException` to catch all panic exceptions raised by an PyO3-based extension. (At least if it is new enough to use the capsule API.)

### DIFF
--- a/src/global_api.rs
+++ b/src/global_api.rs
@@ -1,0 +1,85 @@
+//! TODO
+
+use std::{ffi::CString, mem::forget};
+
+use crate::{
+    conversion::PyTryInto,
+    exceptions::PyTypeError,
+    ffi,
+    sync::GILOnceCell,
+    type_object::PyTypeInfo,
+    types::{PyCapsule, PyDict, PyModule},
+    Py, PyResult, Python,
+};
+
+#[repr(C)]
+pub(crate) struct GlobalApi {
+    version: u64,
+    pub(crate) create_panic_exception:
+        unsafe extern "C" fn(msg_ptr: *const u8, msg_len: usize) -> *mut ffi::PyObject,
+}
+
+pub(crate) fn ensure_global_api(py: Python<'_>) -> PyResult<&GlobalApi> {
+    let api = GLOBAL_API.0.get_or_try_init(py, || init_global_api(py))?;
+
+    // SAFETY: We inserted the capsule if it was missing
+    // and verified that it contains a compatible version.
+    Ok(unsafe { &**api })
+}
+
+struct GlobalApiPtr(GILOnceCell<*const GlobalApi>);
+
+unsafe impl Send for GlobalApiPtr {}
+
+unsafe impl Sync for GlobalApiPtr {}
+
+static GLOBAL_API: GlobalApiPtr = GlobalApiPtr(GILOnceCell::new());
+
+#[cold]
+fn init_global_api(py: Python<'_>) -> PyResult<*const GlobalApi> {
+    let module = match PyModule::import(py, "pyo3") {
+        Ok(module) => module,
+        Err(_err) => {
+            let module = PyModule::new(py, "pyo3")?;
+
+            module.add(
+                "PanicException",
+                crate::panic::PanicException::type_object(py),
+            )?;
+
+            let sys = PyModule::import(py, "sys")?;
+            let modules: &PyDict = sys.getattr("modules")?.downcast()?;
+            modules.set_item("pyo3", module)?;
+
+            module
+        }
+    };
+
+    let capsule: &PyCapsule = match module.getattr("_GLOBAL_API") {
+        Ok(capsule) => PyTryInto::try_into(capsule)?,
+        Err(_err) => {
+            let api = GlobalApi {
+                version: 1,
+                create_panic_exception: crate::panic::create_panic_exception,
+            };
+
+            let capsule = PyCapsule::new(py, api, Some(CString::new("_GLOBAL_API").unwrap()))?;
+            module.setattr("_GLOBAL_API", capsule)?;
+            capsule
+        }
+    };
+
+    // SAFETY: All versions of the global API start with a version field.
+    let version = unsafe { *(capsule.pointer() as *mut u64) };
+    if version < 1 {
+        return Err(PyTypeError::new_err(format!(
+            "Version {} of global API is not supported by this version of PyO3",
+            version
+        )));
+    }
+
+    // Intentionally leak a reference to the capsule so we can safely cache a pointer into its interior.
+    forget(Py::<PyCapsule>::from(capsule));
+
+    Ok(capsule.pointer() as *const GlobalApi)
+}
--- a/src/impl_/trampoline.rs
+++ b/src/impl_/trampoline.rs
@@ -220,7 +220,7 @@ where
     let py_err = match panic_result {
         Ok(Ok(value)) => return value,
         Ok(Err(py_err)) => py_err,
-        Err(payload) => PanicException::from_panic_payload(payload),
+        Err(payload) => PanicException::from_panic_payload(py, payload),
     };
     py_err.restore(py);
     R::ERR_VALUE
@@ -245,7 +245,7 @@ where
     let pool = GILPool::new();
     let py = pool.python();
     if let Err(py_err) = panic::catch_unwind(move || body(py))
-        .unwrap_or_else(|payload| Err(PanicException::from_panic_payload(payload)))
+        .unwrap_or_else(|payload| Err(PanicException::from_panic_payload(py, payload)))
     {
         py_err.write_unraisable(py, py.from_borrowed_ptr_or_opt(ctx));
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -410,6 +410,7 @@ pub mod marker;
 pub mod marshal;
 #[macro_use]
 pub mod sync;
+mod global_api;
 pub mod panic;
 pub mod prelude;
 pub mod pycell;
--- a/src/panic.rs
+++ b/src/panic.rs
@@ -1,7 +1,12 @@
 //! Helper to convert Rust panics to Python exceptions.
+use crate::conversion::{FromPyPointer, IntoPyPointer};
 use crate::exceptions::PyBaseException;
-use crate::PyErr;
+use crate::ffi;
+use crate::global_api::ensure_global_api;
+use crate::{PyAny, Python};
 use std::any::Any;
+use std::slice;
+use std::str;
 
 pyo3_exception!(
     "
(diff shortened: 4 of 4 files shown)


src/global_api.rs
```diff
@@ -0,0 +1,83 @@
...
+static GLOBAL_API: GlobalApiPtr = GlobalApiPtr(GILOnceCell::new());
+
+#[cold]
+fn init_global_api(py: Python<'_>) -> PyResult<*const GlobalApi> {
+    let module = match PyModule::import(py, "pyo3") {
+        Ok(module) => module,
+        Err(_err) => {
+            let module = PyModule::new(py, "pyo3")?;
+
+            // TODO: Register module so it become globally visible and hence importable...
```
adamreichold, line 45: Do I have to insert this into `sys.modules`? Is there a native API for this?

src/impl_/trampoline.rs
```diff
@@ -220,7 +220,7 @@ where
     let py_err = match panic_result {
         Ok(Ok(value)) => return value,
         Ok(Err(py_err)) => py_err,
-        Err(payload) => PanicException::from_panic_payload(payload),
+        Err(payload) => PanicException::from_panic_payload(py, payload),
```
adamreichold, line 223: The types do not work out at all for now, but I wanted to discuss the concrete approach first before trying plough through the trampolines.

davidhewitt:
(I had intended to include `GILProtected` in 0.18.2, unfortunately bungled the rebase. We could push out a 0.18.3 however I'm tempted to just get on with some new features for 0.19 and get that out the door faster instead :) )

adamreichold:
> (I had intended to include `GILProtected` in 0.18.2, unfortunately bungled the rebase. We could push out a 0.18.3 however I'm tempted to just get on with some new features for 0.19 and get that out the door faster instead :) )

Currently, the only reason I would see for a 0.18.3 is the `PyDateTime` type object SNAFU.

adamreichold:
Any comments w.r.t. the proposed direction? Do want this solution for cross-extension functionality? Are there alternatives? Are we okay with a `static` for caching the capsule pointer? Again, are there alternatives?

davidhewitt:
There are the following questions / alternatives in my mind:

- Functionality: I can imagine this API may grow to also contain various infrastructure types beyond the exception:
  - `pyo3.Future` type which could be a wrapper of `Box<dyn Future>`
  - `pyo3.Iterator` type which could be a wrapper of `Box<dyn Iterator<Item = PyObject>>`
  - This may give us power to build designs which we can't easily ship at the moment.
- Discoverability: we could publish a `pyo3` package on PyPI. The upside of doing this is that then downstream code can `import pyo3` immediately without needing to first import a module which is built in PyO3. **I quite strongly believe downstream users will frequently be confused if `import pyo3` is path-dependent and they can't install it, so I think publishing is probably necessary.**
  - We can also ship `.pyi` type annotations for `PanicException`.
  - This capsule-based solution would likely still be used to build that `pyo3` package?
  - Downside of shipping on `PyPI` is that pyo3-built extensions may need to update their Python dependencies to list our PyO3 version, creating more churn.
- Versioning: this current draft crashes if there are conflicting PyO3 versions. I wonder if we can namespace them to support multiple versions. We may also want to version per portion of the API individually rather than at the global level.
  - I think the main goal is to ensure downstream code can be as agnostic as possible about individual PyO3 versions.
  - It would also be quite helpful if PyO3-built packages using different PyO3 versions can coexist. This might be at odds with `pyo3` being installable from PyPI.
    - Perhaps the install from PyPI defines the top-level API which is importable, and we do as best we can to make it forward and backward-compatible with other PyO3 versions?
      - Maybe it's even possible to make that work in a way where downstream crates do not need to require or pin to a specific `pyo3` package version as a Python dependency?
  - As an idea, `pyo3.PanicException` can be itself subclassed by versioned types, e.g. `pyo3.v0_20.PanicException`, `pyo3.PanicExceptionV1`, `pyo3.PanicExceptionV2`. That way we could freely rebuild the actual concrete type per version while users can still just write `except PanicException` and not care about versioning.
- I think the `static` falls under the same category that at the same time it's a necessary evil, however when we have a way to store this in module data then we should do so and remove the `static`.

adamreichold:
One thing I noticed that if we provide a separate pyo3 package which installs the capsule API, we have to be very very careful to not do something that would call into that capsule API until it is installed when that pyo3 package is initialized. Or we would need to not use PyO3 to implement the pyo3 package?

adamreichold:
Making the pyo3 package pure Python and caching a reference to the constructor of a PanicException defined in that Python package and just calling that via the normal Python calling convention would have a certain simplicity going for it, wouldn't it? The performance should also not be such a problem as long as we talk about raising exceptions?

davidhewitt:
> Making the pyo3 package pure Python and caching a reference to the constructor of a PanicException defined in that Python package and just calling that via the normal Python calling convention would have a certain simplicity going for it, wouldn't it? The performance should also not be such a problem as long as we talk about raising exceptions?

Definitely would be simpler. Would that rule out replacing it with a capsule-based version later? Any older PyO3 versions which just do `import pyo3` could potentially fall-back to using a private type if the import fails.

davidhewitt:
> One thing I noticed that if we provide a separate pyo3 package which installs the capsule API, we have to be very very careful to not do something that would call into that capsule API until it is installed when that pyo3 package is initialized. Or we would need to not use PyO3 to implement the pyo3 package?

Can you provide an example of how things might go wrong? I haven't thought super hard about this, however it's not readily apparent to me what the kind of failure you're anticipating here looks like.

davidhewitt:
> One thing I noticed that if we provide a separate pyo3 package which installs the capsule API, we have to be very very careful to not do something that would call into that capsule API until it is installed when that pyo3 package is initialized. Or we would need to not use PyO3 to implement the pyo3 package?

Can you provide an example of how things might go wrong? I haven't thought super hard about this, however it's not readily apparent to me what the kind of failure you're anticipating here looks like.

adamreichold:
> Can you provide an example of how things might go wrong? I haven't thought super hard about this, however it's not readily apparent to me what the kind of failure you're anticipating here looks like.

Well let's say initialization of the `pyo3` extension panics and wants to create a `PanicException` but alas it cannot because it panicked before the capsule API used to instantiate a `PanicException` was installed. We could silently fall back to an internal type, but as a general approach, this would probably mask a lot of configuration/deployment errors which only come to light when `except PanicException:` does not work, i.e. at a really bad time on the production system. So we might need to limit that fallback to the `pyo3` extension itself, but that brings us back to the self-referential dance showcased here which would be nice to avoid by using a canonical source for the capsule API.

> Definitely would be simpler. Would that rule out replacing it with a capsule-based version later? Any older PyO3 versions which just do import pyo3 could potentially fall-back to using a private type if the import fails.

I think we could always upgrade to a capsule API later and hence think that starting with a pure Python `pyo3` package (and by extension pure Python `PanicException` class) would probably be the best next step. We can get this published, integrate it into version 0.20.0 and gain some general experience with the approach without getting bogged down in the technicalities of the capsule API evolution.

davidhewitt:
Ah right I see, yes. There are options like aborting or falling back silently but they're all quite messy, so something it would be nice to learn about first.

I'm in favour of starting with a pure-Python package and we can learn from there.

So I think just a couple questions to consider:

- Do we call the PyPI package `pyo3` or something like `pyo3_runtime`? (`pyo3_runtime.PanicException` is the current name I gave the type object.) I am happy to consider arguments for both or any other suggestion.
  - It might be a good idea to go for `pyo3` and `pyo3.PanicException` to make a clean break from the non-importable `pyo3_runtime.PanicException` which will exist for 0.19 and older.
- Versioning: do we version this separately from the Rust? If downstream packages put bounds on, say, `pyo3==0.20`, they can't be installed with packages depending on `pyo3==0.21`. Maybe we document to always use lower bounds `pyo3>=0.20` and promise to users we will never break compatibility? That's really hard, `pyo3>=0.20,<1`? Same compatbility problem I guess when we release 1.0 and end users can't have compatible PyO3 runtimes.
  - Maybe use calendar versioning for this package rather than semver? That way there's no "breaking" releases and it's just about what runtime features the end user needs. `pyo3_runtime>=2023.06` ? `pyo3>=2023.06` ?

davidhewitt:
... I think I quite like going for `pyo3` with calendar versioning - it's relatively obvious what it means. It has the great property that `pyo3>=2023.05.31` is sufficiently far enough from the Rust version of `0.19.0` that users are likely to pay attention to our documentation if we ask them to always put just a lower bound.

adamreichold:
I am not really familiar with calendar version. From what you describe, is there a difference between `>=0.20` and promise to never break and `>=2023.06`? Does that somehow allow more breakage?

Personally, I would probably prefer to version this separately from PyO3 (the library), e.g. we just start at `1.0` with a standard bound `>=1.0,<2`, i.e. we can break if we ever need to but I don't see us doing that without a very very good reason, e.g. not "just" because PyO3 (the library) hits 1.0. Version numbers being independent would also match that I would envision pyo3 (the package) evolving much slower than PyO3 (the library).

davidhewitt:
The difference between `>=0.20` looking like a semantic version and `>=2023.06` would be that calendar version bumps don't imply any restrictions or changes on breakage, it just labels the release date of the package.

If we go with this package being `1.0` and followed semantic versioning I think it would be less confusing for users if we called it something different like `pyo3_runtime` or `pyo3_core`. Imagine we release the crate PyO3 1.0 a couple years from now and we're on version 1.7 of the `pyo3` PyPI package. I suspect many users may trip up and try to install the older PyPI package.

Regardless of which combination we go for, I've observed that one nice thing about documenting lower bounds is that we only need to care about backwards-compatibility of the PyPI package with older PyO3 versions. We don't need to gracefully degrade PyO3 to use older versions of the PyPI package, because we can always document "extensions built with PyO3 0.20 require `pyo3_core>=1.0`" (or `>=2023.06`). As long as downstream users have set their bounds appropriately, we'll always have a suitably new package.

... Just trying to write that I realise how confusing it is to write `pyo3>=1.0` and PyO3 0.20 in the same sentences, so I'd be quite keen to call the PyPI package something different unless we either:
- version and release it alongside the PyO3 crate, so everything is in sync
- (maybe) use calendar versioning, if we think the distinction that semver used for Rust crate, calver for Python package can be clear enough for users to follow.

davidhewitt:
So I think we're converging to the decision that the next step will be to push a `pyo3_runtime` package with version `1.0`? Shall we keep it in this repository for ease of development?

> I definitely think that being able to break pyo3_runtime at some point in the future if required is worth the disconnect between versions of the PyPI package and the Rust crate.

My gut feeling is that even if we reserve the capability to do so, we'll never want to break the package in practice, because then ecosystem of packages using PyO3 is forcibly split by the runtime version they need. I can imagine we'd be really popular for putting the whole ecosystem in dependency hell 🙃 

> I think this less an effect of lower bounds, but rather of Python accepting only one version of a package per dependency tree. Because the downgrade is required here because different versions of PyO3 could be combined to built extensions loaded into a single Python interpreter and each would have to be able to cope with the capsule API installed by the others. With any Python package like `pyo3_runtime` this is not possible as it will be loaded only in one version. Of course, the price to pay is that users have to manage this themselves in every extension built using PyO3.

Yes, this is essentially what I meant. As long as the bounds are set correctly for `pyo3_runtime` on downstream extension manifests then the extension with highest runtime requirement will win, and that runtime should support all the extensions needing older versions.

adamreichold:
> So I think we're converging to the decision that the next step will be to push a pyo3_runtime package with version 1.0? Shall we keep it in this repository for ease of development?

Yes, I think the ease of changing this together with the usage sites would be worth it.

I'll close this issue then because this not how we will proceed. I still a pure-Python `class PanicException` would be a good first export of `pyo3_runtime` though.

davidhewitt:
> I still a pure-Python class `PanicException` would be a good first export of `pyo3_runtime` though.

Agreed. (I'm up for us including this in 0.20, shall we track it in an issue?)

davidhewitt:
I just came across https://docs.python.org/3/c-api/init.html#c.PyInterpreterState_GetDict, which looks like it _might_ be useful for storing panic exception which is shared across extensions. I don't think it would offer importable symbols for downstream code though.

adamreichold:
Indeed, I still think a `pyo3_runtime` published on PyPI opens up whole new avenues of integration between extensions and would not avoid doing that just because we have alternative ways of stashing per-interpreter or per-module state somewhere.

PR #3087, opened 2023-04-07
PEP 623: wstr, wstr_length, state changes

ijl:
https://peps.python.org/pep-0623/

	Following members are removed from the Unicode structures:

	    wstr
	    wstr_length
	    state.compact
	    state.ready

Note state.compact is still present as of v3.12.0a7.

state.interned has changed width from 2 to 1 bits.


adamreichold:
The one concern I see here is that this is auto-generated code that we need now extend with new functionality (supporting both CPython 3.11 and 3.12) manually. We might want to just generate the same code using the CPython 3.12 headers and `#[cfg(Py_3_12)]` only the complete generated output but I am honestly not sure whether the result would be better than this localized change. But it does commit us somewhat to maintaining this by hand from now on.

davidhewitt:
I think we're probably stuck maintaining this by hand for now; I think there's some willingness upstream to make `state` a private implementation detail and just use accessor functions (and we'd just offer an opaque type here), however that would take a few years to make it through.

Note that we don't have 3.12 in CI yet - #2709 and #2817

I'm ok with landing this change without CI but I'd rather we take away the CHANGELOG entry in that case, as I don't think we can claim to support any 3.12 functionality without CI.

davidhewitt:
Thanks - will ship this in 0.18.3 👍

PR #3097, opened 2023-04-12
ci: use codecov coverage format

davidhewitt:
Newly supported on `cargo-llvm-cov` 0.5.12, might give some more detailed information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -312,7 +312,7 @@ jobs:
       - uses: codecov/codecov-action@v3
         if: steps.should-skip.outputs.skip != 'true'
         with:
-          file: coverage.lcov
+          file: coverage.json
           name: ${{ matrix.os }}
 
   emscripten:
--- a/noxfile.py
+++ b/noxfile.py
@@ -62,9 +62,9 @@ def coverage(session: nox.Session) -> None:
         "--package=pyo3-macros",
         "--package=pyo3-ffi",
         "report",
-        "--lcov",
+        "--codecov",
         "--output-path",
-        "coverage.lcov",
+        "coverage.json",
         external=True,
     )
 


adamreichold:
The main difference behind the 15 percent drop in coverage seems to be that tracking of error paths is more detailed, e.g. `foo()?` seems to be covered only if both `Ok` and `Err` paths are taken.

This seems to be an improvement even though the numbers do not give me that warm fuzzy feeling.

davidhewitt:
I agree that showing the unhandled error cases is an improvement, however it doesn't seem like there's an easy way to see what the uncovered part of a line is in the codecov UI. This seems like it'll be pretty hard to actually then make meaningful improvements on?

I'm trying to see if there's a way to look locally.

davidhewitt:
So I've been looking at the llvm-cov html report this morning and it shows what regions are covered for each instantiation. Presumably to get 100% region coverage each region just needs to be covered at least once, not for every instantiation.

Interestingly, the codecov numbers seem much more pessimistic than the llvm-cov html report.

Take `impl_/pyclass/lazy_type_object.rs`, for example. On codecov, this PR, apparently the coverage is 48%

image

https://app.codecov.io/gh/PyO3/pyo3/pull/3097/blob/src/impl_/pyclass/lazy_type_object.rs

But on the local html report, the function / line / region coverage are all in the 90% range:

image

@taiki-e @andrewgazelka I wonder if you have an idea why the codecov numbers are so much more pessimistic than the llvm-cov html report?

andrewgazelka:
I'll look into it when I wake up. The way coverage can be determined is kinda subjective but yea I think it would be ideal to match --html

andrewgazelka:
> Thanks. I can push an update to the `noxfile` to make it generate the html and codecov reports (and any other formats) if that would be useful.

It would be great if it is hard to run the tests locally. Otherwise, I'll just run llvm-cov locally with `cargo test`. :)

davidhewitt:
It's more complicated than `cargo test` because we have UI tests and also a python module which is tested using `pytest`.

I've pushed a commit to this branch which will make the coverage job generate html, llvm-cov json and codecov json formats.

You can run with:

```bash
pip install nox
nox -s coverage
```

andrewgazelka:
@davidhewitt @adamreichold could this be related?

https://github.com/taiki-e/cargo-llvm-cov/blob/80cd0175844183f470fd705648907640266222a6/src/json.rs#L134-L147

andrewgazelka:
I think the way they calculate code coverage is different. For instance

image


shows there are only 52 regions, when in fact there are _many_ more (given all the generics). Just open the file.

davidhewitt:
I think with the generics it's the same region instantiated many times over? Potentially if that region is covered by at least one instantiation, it's considered covered?

If that's true, the trick may be to identify equivalent regions?

davidhewitt:
EDIT: rereading your comment I see you think exactly that too. It seems correct to me to merge instantiations in this way. Otherwise to get perfect coverage you have to test every branch of your function with every type you instantiate. Plus your users will instantiate your generic functions with types you didn't even think about, so "coverage of all genetic instantiations" doesn't seem like a helpful property.

andrewgazelka:
Ok that logic makes sense. That's definitely the issue in my program then. I'll have a go at this tomorrow, I have a sense of how I'll approach this.

andrewgazelka:
Hey @davidhewitt and @adamreichold, could you take a look at https://github.com/taiki-e/cargo-llvm-cov/pull/255 and see if it resolves this issue?

davidhewitt:
Thanks, I was busy over the weekend, will try to get to testing this out tomorrow evening.

davidhewitt:
I just reran this and the numbers look much more like what I'd expect now, thanks @andrewgazelka !

@adamreichold I'm going to proceed to merge this now, I think showing the uncovered errors is strictly an improvement!

PR #3100, opened 2023-04-12
build: use runtime OUT_DIR

coffinmatician:
This helps building on systems where they may be different (bazel, buck2, etc.)

### DIFF
--- a/pyo3-build-config/src/lib.rs
+++ b/pyo3-build-config/src/lib.rs
@@ -11,10 +11,7 @@ mod errors;
 mod impl_;
 
 #[cfg(feature = "resolve-config")]
-use std::{
-    io::Cursor,
-    path::{Path, PathBuf},
-};
+use std::{io::Cursor, path::PathBuf};
 
 use std::{env, process::Command, str::FromStr};
 
@@ -115,11 +112,13 @@ const HOST_CONFIG: &str = include_str!(concat!(env!("OUT_DIR"), "/pyo3-build-con
 #[doc(hidden)]
 #[cfg(feature = "resolve-config")]
 fn resolve_cross_compile_config_path() -> Option<PathBuf> {
-    env::var_os("TARGET").map(|target| {
-        let mut path = PathBuf::from(env!("OUT_DIR"));
-        path.push(Path::new(&target));
-        path.push("pyo3-build-config.txt");
-        path
+    env::var_os("TARGET").and_then(|target| {
+        std::env::var_os("OUT_DIR").map(|out_dir| {
+            let mut path = PathBuf::from(out_dir);
+            path.push(&target);
+            path.push("pyo3-build-config.txt");
+            path
+        })
     })
 }
 
--- a/newsfragments/3100.changed.md
+++ b/newsfragments/3100.changed.md
@@ -1,0 +1,1 @@
+Use runtime OUT_DIR for pyo3-build-config (helps with build systems other than Cargo in which the compile, and runtime OUT_DIR are different for build scripts)


pyo3-build-config/src/lib.rs
```diff
@@ -115,11 +115,13 @@ const HOST_CONFIG: &str = include_str!(concat!(env!("OUT_DIR"), "/pyo3-build-con
...
 fn resolve_cross_compile_config_path() -> Option<PathBuf> {
-    env::var_os("TARGET").map(|target| {
-        let mut path = PathBuf::from(env!("OUT_DIR"));
-        path.push(Path::new(&target));
-        path.push("pyo3-build-config.txt");
-        path
+    env::var_os("TARGET").and_then(|target| {
+        std::env::var_os("OUT_DIR").map(|out_dir| {
+            let mut path = PathBuf::from(out_dir);
+            path.push(Path::new(&target));
```
adamreichold, line 121: It isn't really related to this PR, but is the explicit call to `Path::new` really necessary here? Could you remove it?

adamreichold:
> This helps building on systems where they may be different (bazel, buck2, etc.)

Could you mention this rationale (build systems other than Cargo) in the news fragment? Thanks!

kngwyu:
I'm also a bit confused. If this PR is meant to handle the case when 'OUT_DIR' is not set, the PR title and changelog need to be revised. I don't have any experience using another build system than cargo to compile Rust crate, but do you mean that some of them don't set OUT_DIR?

adamreichold:
I'd say this is basically about replacing `env!("OUT_DIR")` (evaluated when `build.rs` is compiled) by `env_os("OUT_DIR")` (evaluated when it is executed).

kngwyu:
Ah, I see, thank you.

pyo3-build-config/src/lib.rs
```diff
@@ -115,11 +115,13 @@ const HOST_CONFIG: &str = include_str!(concat!(env!("OUT_DIR"), "/pyo3-build-con
 #[doc(hidden)]
 #[cfg(feature = "resolve-config")]
 fn resolve_cross_compile_config_path() -> Option<PathBuf> {
-    env::var_os("TARGET").map(|target| {
-        let mut path = PathBuf::from(env!("OUT_DIR"));
-        path.push(Path::new(&target));
-        path.push("pyo3-build-config.txt");
-        path
+    env::var_os("TARGET").and_then(|target| {
+        std::env::var_os("OUT_DIR").map(|out_dir| {
```
konstin, line 119: Could you add a link to the documentation of e.g. bazel or buck2 that says how they set `OUT_DIR`? otherwise it will be hard to determine in the future why it's written in this way and how you may or may not change it

adamreichold: I also think this strictly an improvement as we never needed compile-time access to `OUT_DIR` in the first place, i.e. we could have written it this way from the start. But let's wait what @konstin thinks w.r.t. your answer before merging.

adamreichold (approved):

adamreichold:
@coffinmatician I know I am to blame because I asked for the removal of `Path::new`, but it seems like you need to fix the unused import warning triggered by that as well.

adamreichold:
> but I assumed that the CI here would test some cross compilation.

I want to say that we do, e.g. https://github.com/PyO3/pyo3/actions/runs/4692932437/jobs/8319284869, but it seems we are missing the `CARGO_BUILD_TARGET` environment variable for `clippy` CI jobs that we set for the `build` ones? cc @messense

davidhewitt:
> Hmm, I think we may want to wait on this patch a bit longer. A coworker is suspicious that I have broken things subtly for other targets. I probably should have tested it more in traditional cargo to make sure it was working as I expected, but I assumed that the CI here would test some cross compilation.

Sorry I didn't get to commenting on this thread sooner.

If I recall correctly, I made the choice to bake-in the `OUT_DIR` at build time because this code can get evaluated from multiple different build scripts and I wasn't sure that it would be the same each time.

I think #1758 is the right PR with the original context. At the time we needed to access the config from our proc macros. In #2243 that access was removed.

It's very possible there's a solution I didn't think of, or something enabled by newer cargo functionality.

PR #3111, opened 2023-04-18
Rename sequence `.list()` and `.tuple()` to `.to_list()` and `.to_tuple()`

davidhewitt:
As agreed in https://github.com/PyO3/pyo3/pull/3042#issuecomment-1465229510.

The motivation is that it is an emerging Rust convention for `to_x` to mean "a non-consuming conversion to type x which has some (small) overhead". E.g. `to_string`, `to_owned`.

### DIFF
--- a/src/types/sequence.rs
+++ b/src/types/sequence.rs
@@ -239,20 +239,34 @@ impl PySequence {
 
     /// Returns a fresh list based on the Sequence.
     #[inline]
-    pub fn list(&self) -> PyResult<&PyList> {
+    pub fn to_list(&self) -> PyResult<&PyList> {
         unsafe {
             self.py()
                 .from_owned_ptr_or_err(ffi::PySequence_List(self.as_ptr()))
         }
     }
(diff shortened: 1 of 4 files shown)


src/types/sequence.rs
```diff
@@ -239,6 +239,16 @@ impl PySequence {
 
     /// Returns a fresh list based on the Sequence.
     #[inline]
+    pub fn to_list(&self) -> PyResult<&PyList> {
+        unsafe {
```
adamreichold, line 243: Personally, I would prefer if one implementation, e.g. the deprecated one, would forward to the other to avoid the code duplication. I do understand that this is probably not worth it for one-liners, on the other hand these are `unsafe` one-liners.

davidhewitt: Agreed, I meant to do this and then forgot while actually doing the typing! Have force-pushed to correct.

messense:
bors retry

Issue #3112, opened 2023-04-21
Feature/Documentation: simpler FromPyObject for Python dict


BlueGlassBlock:
The documentation of `FromPyObject` mentioned the usage of `#[pyo3(item)]`, however I think it's quite feasible to have a `#[pyo3(item_all)]` on structs, which collects every field with `get_item`.

Apart from that, the `pythonize` project also provided the ability to build structs from Python dictionary, which worths mentioniing in the documentation as an alternative solution. However using `#[pyo3(item_all)]` should be easier for nested structures.

I'll work on the implementation and come up with a PR if `#[pyo3(item_all)]` is appropriate for PyO3 to have.

PR #3120, opened 2023-04-25
feat: add #[pyo3(item_all)] for FromPyObject

BlueGlassBlock:
This PR addresses #3112 by implementing `#[pyo3(item_all)]`.

### DIFF
--- a/guide/src/conversions/traits.md
+++ b/guide/src/conversions/traits.md
@@ -137,6 +137,35 @@ from a mapping with the key `"key"`. The arguments for `attribute` are restricte
 non-empty string literals while `item` can take any valid literal that implements
 `ToBorrowedObject`.
 
+You can use `#[pyo3(item_all)]` on a struct to extract every field with `get_item` method.
+In this case, you can't use `#[pyo3(attribute)]` or barely use `#[pyo3(item)]` on any field.
+However, using `#[pyo3(item("key"))]` to specify the key for a field is still allowed.
+
+```rust
+use pyo3::prelude::*;
+
+#[derive(FromPyObject)]
+#[pyo3(item_all)]
+struct RustyStruct {
+    foo: String,
+    bar: String,
+    #[pyo3(item("foobar"))]
+    baz: String,
+}
+#
+# fn main() -> PyResult<()> {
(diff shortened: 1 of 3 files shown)


BlueGlassBlock:
> I think UI tests exercising the error paths would be good to have here.

I've noticed that, and I'll figure out how to fix the tests and improve coverage today (that's why I marked this as a draft⁠) (⁠^⁠^⁠)

davidhewitt:
It looks like you may have run the UI tests with nightly Rust? Quite a few spans have changed. Suggest rebuilding with stable.

Overall I'm happy for this functionality although I'm unsure about the name `item_all`. I understand why you chose it though it's not immediately intuitive to me as a reader what it means.

I wonder if something like `all_from_items` or `from_item_all` are any clearer? I don't love those either so happy for further suggestions!

BlueGlassBlock:
I personally perfer `from_item_all`, since that's similar to `get_all` and `set_all` which we already have.

I'll update the PR again once I can continue to work on it.

BlueGlassBlock:
It seemed that I forgot some paths in the UI tests. I'll address them soon.

I'm uncertain about the name too. `from_items` or something like that seems okay to me.

I'm thinking about the divergence in behavior from `get_all` and `set_all`, which don't allow specifying the name with `get, name = ...`, so that seems inconsistent. However, I think allowing specifying args passed to `get_item` is still useful. Maybe I could separate that to an extra proposal along with changes to `get_all` and `set_all`?

davidhewitt:
> I'm thinking about the divergence in behavior from get_all and set_all, which don't allow specifying the name with get, name = ..., so that seems inconsistent.

This is a good point and something we overlooked when originally implementing `get_all` and `set_all`, it would be nice to allow them to accept `#[pyo3(name = ...)]` on individual fields.

BlueGlassBlock:
Ugh. I didn't manage to make coverage rate pass codecov's check. Are declarative macro invocations blocking llvm-cov to pick them up?

davidhewitt (approved):
This looks fine to me, the implementation is comprehensive. For some reason the coverage isn't reporting the error cases from the UI tests, but I see them here, and that normally works, so I think we are good to merge this. Thanks!

bors r+

PR #3130, opened 2023-05-01
Make sure to upgrade pip before trying to install nox.

adamreichold:


adamreichold:
Sorry for merging this without adequate review and without having exactly done a root cause analysis here, but I would like to avoid more contributors running into a broken CI.

bors r+

davidhewitt:
Thanks for fixing, all looks good to me here!

PR #3131, opened 2023-05-01
Extend lifetime of GIL token associated with PyRef(Mut).

adamreichold:


src/pycell.rs
```diff
@@ -583,8 +583,8 @@ pub struct PyRef<'p, T: PyClass> {
 
 impl<'p, T: PyClass> PyRef<'p, T> {
     /// Returns a `Python` token that is bound to the lifetime of the `PyRef`.
```
mejrs, line 585: This documentation is no longer correct (same for PyRefMut's)

src/pycell.rs
```diff
@@ -706,8 +706,8 @@ pub struct PyRefMut<'p, T: PyClass<Frozen = False>> {
 
 impl<'p, T: PyClass<Frozen = False>> PyRefMut<'p, T> {
     /// Returns a `Python` token that is bound to the lifetime of the `PyRefMut`.
```
mejrs, line 708: And here

mejrs (approved):
Lgtm, the docs need fixing though.

src/pycell.rs
adamreichold: I did consider changing this but ended up thinking that the current documentation actually matches the new implementation: We do bind to the lifetime `'p` of the `PyRef`, not the lifetime `&'a self` of the borrow of the `PyRef`.

Of course, I am very much open to changing this. I was just unable to come up with a better fitting text than the current one. Could you suggest an alternative wording?

davidhewitt (approved):
Thanks! Agree with @adamreichold that this aligns the implementation with the original documentation.

src/pycell.rs
mejrs: It has a lifetime equal to the Python token originally used to create the PyRef(Mut), right?

adamreichold: Not necessarily. When you have `py: Python<'py>, arg: PyRef<'py, T>` in your signature, it will match the GIL token lifetime. But if you manually call [`PyCell::borrow`](https://docs.rs/pyo3/latest/pyo3/pycell/struct.PyCell.html#method.borrow) somewhere, it will really only match the lifetime of that `PyRef` instance (i.e. the lifetime of the borrow of the `PyCell`) which might be much shorter than the original GIL token used to create e.g. an `py: Python<'py>, arg: &'py PyCell<T>` set of function arguments.

mejrs: Okay, feel free to disregard this then :)

adamreichold:
bors r=mejrs

PR #3140, opened 2023-05-06
Extend lifetime of holder references

### LINKED ISSUE #3138: E0716 "temporary value dropped while borrowed" in `#[pyfunction]`
### Bug Description

:wave: Love PyO3! 

Upgrading from `0.16` to `0.18` I'm seeing the following compiler error:

```
error[E0716]: temporary value dropped while borrowed
    --> src/externs/interface.rs:1024:17
     |
1021 | #[pyfunction]
     | -------------
     | |           |
     | |           temporary value is freed at the end of this statement
     | borrow later used here
...
1024 |   py_scheduler: &'py PyScheduler,
     |                 ^ creates a temporary value which is freed while still in use
     |
     = note: consider using a `let` binding to create a longer lived value
```

The code in question looks like:

```rust
#[pyclass]
struct PyScheduler(Scheduler);

// ...

#[pyfunction]
fn scheduler_metrics<'py>(
  py: Python<'py>,
  py_scheduler: &'py PyScheduler,
  py_session: &'py PySession,
) -> HashMap<&'py str, i64> {
  py_scheduler
    .0
    .core
    .executor
    .enter(|| py.allow_threads(|| py_scheduler.0.metrics(&py_session.0)))
}
```

### Steps to Reproduce

(Try and compile the code)

### Backtrace

_No response_

Ubuntu 20

3.8

### Your PyO3 version

0.18.3

`apt` and no venv

### Additional Info

_No response_

adamreichold:
Closes #3138

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -941,11 +941,13 @@ impl Ty {
         ident: &syn::Ident,
         arg: &FnArg<'_>,
         extract_error_mode: ExtractErrorMode,
+        holders: &mut Vec<TokenStream>,
     ) -> TokenStream {
         let name_str = arg.name.unraw().to_string();
         match self {
             Ty::Object => extract_object(
                 extract_error_mode,
+                holders,
                 py,
                 &name_str,
                 quote! {
@@ -954,6 +956,7 @@ impl Ty {
             ),
             Ty::MaybeNullObject => extract_object(
                 extract_error_mode,
+                holders,
                 py,
                 &name_str,
                 quote! {
@@ -968,6 +971,7 @@ impl Ty {
             ),
             Ty::NonNullObject => extract_object(
                 extract_error_mode,
+                holders,
                 py,
                 &name_str,
                 quote! {
@@ -976,6 +980,7 @@ impl Ty {
             ),
             Ty::IPowModulo => extract_object(
                 extract_error_mode,
+                holders,
                 py,
                 &name_str,
                 quote! {
@@ -1024,17 +1029,23 @@ fn handle_error(
 
 fn extract_object(
     extract_error_mode: ExtractErrorMode,
+    holders: &mut Vec<TokenStream>,
     py: &syn::Ident,
     name: &str,
     source: TokenStream,
 ) -> TokenStream {
+    let holder = syn::Ident::new(&format!("holder_{}", holders.len()), Span::call_site());
+    holders.push(quote! {
+        #[allow(clippy::let_unit_value)]
+        let mut #holder = _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT;
+    });
     handle_error(
         extract_error_mode,
         py,
         quote! {
             _pyo3::impl_::extract_argument::extract_argument(
                 #source,
-                &mut { _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT },
+                &mut #holder,
                 #name
             )
         },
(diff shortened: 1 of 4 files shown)


adamreichold:
@davidhewitt I wanted to push this here to ask your opinion of whether using local variables to extend the holder lifetimes is a suitable approach or whether we should do something else here entirely?

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -1024,17 +1029,22 @@ fn handle_error(
...
+    holders.push(
+        quote! { let mut #holder = _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT; },
+    );
     handle_error(
         extract_error_mode,
         py,
         quote! {
             _pyo3::impl_::extract_argument::extract_argument(
                 #source,
-                &mut { _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT },
```
adamreichold, line 1037: I could not produce a test case actually triggering the issue but changed it for completeness. I did put it into a separate commit though if it turns out that we do not need this after all.

davidhewitt:
Ah, well found!

So this is definitely a viable solution which has the nice property that it allows us to bind local variables inside the generated macro code.

I think the other option is to avoid binding locals at all and just have one giant tail expression, because then the holder references will just be temporaries which will get extended until the end of the expression (by which point the return value has been converted into Python types). Something like this patch seems to fix the bug to me:

```
diff --git a/pyo3-macros-backend/src/method.rs b/pyo3-macros-backend/src/method.rs
index 01aedf695a..b7f967472d 100644
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -425,9 +425,8 @@ impl<'a> FnSpec<'a> {

         let rust_call = |args: Vec<TokenStream>| {
             quote! {
-                let mut ret = function(#self_arg #(#args),*);
-                let owned = _pyo3::impl_::pymethods::OkWrap::wrap(ret, #py);
-                owned.map(|obj| _pyo3::conversion::IntoPyPointer::into_ptr(obj))
+                _pyo3::impl_::pymethods::OkWrap::wrap(function(#self_arg #(#args),*), #py)
+                    .map(|obj| _pyo3::conversion::IntoPyPointer::into_ptr(obj))
                     .map_err(::core::convert::Into::into)
             }
         };
```

(This same trick might need applying in a few of the other code paths which you've corrected too.)

Overall I'm happy with either approach. Being able to bind local variables makes the generated macro code easier for us to read and debug. On the other hand, the patch I put above keeps the implementation in `pyo3-macros-backend` simpler. So it's just a case of whatever we like best. I slightly prefer the conciseness of the patch I put above, though I am also fine with the PR as proposed here.

adamreichold:
> So it's just a case of whatever we like best. I slightly prefer the conciseness of the patch I put above, though I am also fine with the PR as proposed here.

I am prepared #3142 with the tail-expression-based solution which is certainly a much smaller change and results less generated code which makes it the obvious choice for me.

The only concern I am wrangling with myself is that it does seem like a "trick" to me and I would have a hard time inferring this effect by looking at the code. At least in this regard, I find the solution using local variables more pedestrian and more importantly explicit in that it is easier to understand why this is done.

I am not completely convinced the effect is real though as I have never been on friendly terms with PyO3's proc macro code and this might just be me having a hard time doing anything with it.

davidhewitt:
Ok - let's go with #3142 for now.

I agree the macro implementations could use some refactoring, they've grown very... organically. 😅 

What I'm hoping to move towards with that code is more structs which represent the parsed types, and then potentially more structs / functions to compose the generated code. E.g. this "ok-wrap and convert to PyErr" pattern can probably be its own function (or structure which implements `quote::ToTokens`, not sure what is more performant).

PR #3141, opened 2023-05-06
Add BaseExceptionGroup for Python >= 3.11

adriangb:
Not sure if this is totally off base, but it looks like it may be this easy to add support for ExceptionGroup?

### DIFF
--- a/newsfragments/3141.added.md
+++ b/newsfragments/3141.added.md
@@ -1,0 +1,1 @@
+Add FFI definition for `BaseExceptionGroup`, which was added in Python 3.11 by [PEP 0654](https://peps.python.org/pep-0654/).
--- a/pyo3-ffi/src/pyerrors.rs
+++ b/pyo3-ffi/src/pyerrors.rs
@@ -103,6 +103,8 @@ pub unsafe fn PyUnicodeDecodeError_Create(
 extern "C" {
     #[cfg_attr(PyPy, link_name = "PyPyExc_BaseException")]
     pub static mut PyExc_BaseException: *mut PyObject;
+    #[cfg(Py_3_11)]
+    pub static mut PyExc_BaseExceptionGroup: *mut PyObject;
     #[cfg_attr(PyPy, link_name = "PyPyExc_Exception")]
     pub static mut PyExc_Exception: *mut PyObject;
     #[cfg_attr(PyPy, link_name = "PyPyExc_StopAsyncIteration")]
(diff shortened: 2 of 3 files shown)


adamreichold:
Please add a newsfragment as described in https://pyo3.rs/v0.18.3/contributing#documenting-changes including a link to an upstream changelog, PEP, etc.

The CI failure looks like this is not available in CPython 3.11.3 on Windows? Or is the name incorrect?

adriangb:
> Please add a newsfragment

Done!

> The CI failure looks like this is not available in CPython 3.11.3 on Windows? Or is the name incorrect?

Hmm not sure. I was kinda guessing at the name. It should be in 3.11 on all platforms. I think this is where it's defined? https://github.com/python/cpython/blob/f5088006ca8e9654fbc3de119462f0ab764e408b/PC/python3dll.c#L785-L786

adamreichold:
The news fragment still mentions `PyExc_ExceptionGroup`?

adamreichold:
I am not sure whether the test defined by the `impl_native_exception!` macro makes sense for `BaseExceptionGroup´? Can one raise it via `raise BaseExceptionGroup`?

adriangb:
It requires an argument `raise BaseExceptionGroup("", [ValueError()])`. I see there's some tests like that already.

adriangb:
Any suggestions on how to make sure the right Python version is being used? `cargo test` and such seems to be picking a non 3.11 version.

adriangb:
Do we need to do anything / can we make the methods like `subset` available?

adamreichold:
> Do we need to do anything / can we make the methods like subset available?

I think you would need to manually bind them if they are available via a C API. Are they?

If they are only available as Python methods, you can already access them via `PyAny::call1` etc. and adding Rust bindings do that `call1` in the background is of limited value.

adriangb:
> I think you would need to manually bind them if they are available via a C API. Are they?

I'm not sure. I don't see any other errors that have methods, so I'm guessing none of those are exposed via the C API?

davidhewitt:
I don't see any C-API functions for BaseExceptionGroup, so I think this is probably all we would typically offer here :)

adriangb:
Is there anything else needed here?

davidhewitt:
I think this is fine to merge, though would you be willing to squash the commits? We like to keep a tidy commit history so bors isn't configured to squash everything

adriangb:
Done!

PR #3142, opened 2023-05-06
Do not store return values in locals so that holders benefit from lifetime extension for temporaries.

### LINKED ISSUE #3138: E0716 "temporary value dropped while borrowed" in `#[pyfunction]`

adamreichold:
Closes #3138

### DIFF
--- a/newsfragments/3142.fixed.md
+++ b/newsfragments/3142.fixed.md
@@ -1,0 +1,1 @@
+Extend lifetime of holder variables to avoid "temporary value dropped while borrowed" errors when `#[pyfunction]`s take references into `#[pyclass]`es
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -425,9 +425,8 @@ impl<'a> FnSpec<'a> {
 
         let rust_call = |args: Vec<TokenStream>| {
             quote! {
-                let mut ret = function(#self_arg #(#args),*);
-                let owned = _pyo3::impl_::pymethods::OkWrap::wrap(ret, #py);
-                owned.map(|obj| _pyo3::conversion::IntoPyPointer::into_ptr(obj))
+                _pyo3::impl_::pymethods::OkWrap::wrap(function(#self_arg #(#args),*), #py)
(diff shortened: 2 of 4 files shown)


davidhewitt (approved):
Thanks, as discussed on #3140, let's move ahead with this solution. I won't merge just in case you want to play with refactoring the `OkWrap` pattern into a function or struct as suggested on that thread. (I am also fine with leaving that for the future, the macro code will be simpler once we ditch the deprecated `#[args]` and also I have a refactoring or two in my mind which maybe I'll get to one day.)

adamreichold:
> I am also fine with leaving that for the future

I'd like to limit ourselves to the fix for now as I am seeing an obvious simplification by trying to extract the ok wrapping ATM.

bors r=davidhewitt

adamreichold:
bors retry

adamreichold:
bors retry

PR #3146, opened 2023-05-09
Added a few lines to document the main difference between maturin and…

why-not-try-calmer:
… setuptools-rust as far as building manylinux-compliant wheels is concerned.


adamreichold:
Thanks! This looks rather helpful. Could you squash the two commits please?

PR #3150, opened 2023-05-10
Fix Clippy CI job cache to really include the platform.

adamreichold:
Doesn't really matter since we never build the same target on different platforms, but fixing it will prevent future confusion when reading it. Sorry for the extra lap. 😫


messense (approved):
bors r+

PR #3108, opened 2023-04-15
Dual-license PyO3 under Apache 2.0 or MIT

DataTriny:
Closes #2339

## Why?

PyO3 is currently licensed under Apache 2.0. The PyO3 community wants the project to be better aligned with the rest of the Rust crates ecosystem. Dual licensing under Apache 2.0 or MIT will give more freedom to the users. In particular, it will make PyO3 GPLv2 compatible.

## We need your help!

If GitHub sent you a notification about this, it means that you have contributed to PyO3 in the past. We need your consent to move forward.

We ask you to please clearly express your agreement by posting a comment here.

Once this is done, you can safely [unsubscribe from this thread](https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/managing-subscriptions-for-activity-on-github/managing-your-subscriptions#unsubscribing-from-notifications-on-the-subscriptions-page).

Thank you!

## List of contributors

- [x] 1tgr (Tim Robinson)
- [x] a1phyr (Benoît du Garreau)
- [x] abingham (Austin Bingham)
- [x] acshi (Acshi Haggenmiller)
- [x] adamreichold (Adam Reichold)
- [x] AdilZouitine (Adil Zouitine)
- [x] adriangb (Adrian Garcia Badaracco)
- [x] aganders3 (Ashley Anderson)
- [x] AleksaC (Aleksa Cukovic)
- [x] alex (Alex Gaynor)
- [x] Alexander-N (Alexander Niederbühl)
- [x] alexpyattaev (Alex Pyattaev)
- [x] Alisa-lisa (Alisa Dammer)
- [x] alonblade (Alon Levy)
- [x] althonos (Martin Larralde)
- [x] Amanieu (Amanieu d'Antras)
- [x] andersk (Anders Kaseorg)
- [x] andrewwhitehead (Andrew Whitehead)
- [x] AntoineRR (Antoine Romero-Romero)
- [x] ArniDagur (Árni Dagur)
- [x] arxanas (Waleed Khan)
- [x] Askaholic
- [x] athre0z (Joel Höner)
- [x] atouchet (Alex Touchet)
- [x] aviramha (Aviram Hassan)
- [x] awestlake87 (Andrew J Westlake)
- [x] b05902132
- [x] batconjurer (Jacob Turner)
- [x] benkay86 (Benjamin Kay)
- [x] Ben-KC (Ben Kooiman-Cox)
- [x] birkenfeld (Georg Brandl)
[...]

PR #3157, opened 2023-05-16
Add support for `#[new]` which is also a `#[classmethod]`

### LINKED ISSUE #3077: `#[new]` which is also a `#[classmethod]`
### Discussed in https://github.com/PyO3/pyo3/discussions/3075

<div type='discussions-op-text'>

<sup>Originally posted by **adriangb** March 29, 2023</sup>
As far as I can tell, `#[new]` has to be a static method. It's a classmethod in Python, could it be made the same in PyO3? Context: I'd like to get the class name as seen by Python for the failure case of calling new.</div>

stuhood:
Fixes #3077.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -607,6 +607,27 @@ Declares a class method callable from Python.
 * For details on `parameter-list`, see the documentation of `Method arguments` section.
 * The return type must be `PyResult<T>` or `T` for some `T` that implements `IntoPy<PyObject>`.
 
+### Constructors which accept a class argument
+
+To create a constructor which takes a positional class argument, you can combine the `#[classmethod]` and `#[new]` modifiers:
+```rust
+# use pyo3::prelude::*;
+# use pyo3::types::PyType;
+# #[pyclass]
+# struct BaseClass(PyObject);
+#
+#[pymethods]
+impl BaseClass {
+    #[new]
+    #[classmethod]
+    fn py_new<'p>(cls: &'p PyType, py: Python<'p>) -> PyResult<Self> {
+      // Get an abstract attribute (presumably) declared on a subclass of this class.
+      let subclass_attr = cls.getattr("a_class_attr")?;
+      Ok(Self(subclass_attr.to_object(py)))
+    }
+}
+```
+
 ## Static methods
 
 To create a static method for a custom class, the method needs to be annotated with the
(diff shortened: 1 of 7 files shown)


DataTriny:
Hello @stuhood 

Thank you!

tests/test_class_new.rs
```diff
@@ -204,3 +204,29 @@ fn new_with_custom_error() {
...
+
+#[pymethods]
+impl NewWithClassMethod {
+    #[new]
+    #[classmethod]
+    fn new(cls: &PyType) -> PyResult<Self> {
+        assert!(cls.is_subclass_of::<NewWithClassMethod>()?);
+        Ok(Self)
+    }
+}
```
davidhewitt, line 220: It would be nice to have a test which extends this struct to show the subtyping works as expected. Maybe the method could take an argument of the type to expect:

```rust
    #[new]
     #[classmethod]
     fn new(cls: &PyType, expected_type: &PyType) -> PyResult<Self> {
         assert!(cls.is(expected_type));
         Ok(Self)
     }
```

It might be easier to do the subtyping in Python by putting this test in the `pytests/` test suite.

davidhewitt, line 220: It would be nice to have a test which extends this struct to show the subtyping works as expected. Maybe the method could take an argument of the type to expect:

```rust
    #[new]
     #[classmethod]
     fn new(cls: &PyType, expected_type: &PyType) -> PyResult<Self> {
         assert!(cls.is(expected_type));
         Ok(Self)
     }
```

It might be easier to do the subtyping in Python by putting this test in the `pytests/` test suite.

guide/src/class.md
```diff
@@ -121,6 +121,27 @@ created from Rust, but not from Python.
 
 For arguments, see the [`Method arguments`](#method-arguments) section below.
 
+### Constructors which accept a class argument
```
davidhewitt, line 124: I think I'd prefer to add this section as part of the `#[classmethod]` section later; the way this document currently reads this is both the documentation of `#[new]` but also a "getting started with classes". I think showing users the option of `#[classmethod]` here may be jumping the gun?

davidhewitt (commented):
Thanks! The implementation is fine; I'm not a huge fan of a new enum variant in `FnType` though I can't see an obviously better way to do it. I think better to deliver the working feature and then hopefully over time refactoring will present an option.

Some suggestions for docs & tests.

stuhood:
Thanks! Have applied that feedback.

davidhewitt (approved):
Looks great to me, thanks! I'll leave for a moment in case anyone else wants to review. If you're willing to squash that would also be appreciated; we like to keep the git history neat when we can 😊

adamreichold (approved):

davidhewitt (approved):
Thanks!

bors r+

Issue #3154, opened 2023-05-15
0.19 Release


davidhewitt:
As discussed in https://github.com/PyO3/pyo3/pull/2593#issuecomment-1548327079 let's put out 0.19 release soon and then make MSRV bump breaking changes to ease the pain with CI.

As usual, list of a few things which I was hoping to round off before the release:
- [x] I'd like to finish #2881, as I've had a few people ask me about performance of instance checking (e.g. #3104) 
- [x] In conversation with @adriangb it was reported that `#[pyclass(extends = PyBaseException)]` doesn't compile without the `unsendable` option, though I haven't investigated this (sounds like a bug).
- [x] I think `PyCell<T>` and `Py<T>` for `#[pyclass(frozen)] struct T` should have a method to access `&T` without needing the GIL. I believe this is safe and will help us make progress to nogil / sub-interpreter support (by encouraging use of frozen pyclasses which can then use Rust synchronization mechanisms).
- [x] ~~I wanted to add `PyNotImplemented`, `PyEllipsis` and `PyNone` singleton types: https://github.com/PyO3/pyo3/pull/2911#discussion_r1087252466~~ Parked for now as #3162, see https://github.com/PyO3/pyo3/issues/3154#issuecomment-1553677796
- [x] I suggested I would implement #3077, although it may be better to postpone this to 0.20 now (sorry!)
- [x] Prevent dropping unsendable classes on other threads #3176
- [x] Avoid acquiring the GIL during `__traverse__` #3168 #3175

adamreichold:
> I think PyCell<T> and Py<T> for #[pyclass(frozen)] struct T should have a method to access &T without needing the GIL. I believe this is safe and will help us make progress to nogil / sub-interpreter support (by encouraging use of frozen pyclasses which can then use Rust synchronization mechanisms).

I think this would require an additional `T: Send + Sync` bound though, wouldn't it? `Py<T>` is unconditionally `Send + Sync` and even a `#[pyclass(frozen)]` could contain types like `Rc` or `RefCell`, couldn't it?

adriangb:
Sorry for giving you so many problems. Let me at least look into the PyBassException thing.

davidhewitt:
> I think this would require an additional T: Send + Sync bound though, wouldn't it? Py<T> is unconditionally Send + Sync and even a #[pyclass(frozen)] could contain types like Rc or RefCell, couldn't it?

[...]

PR #3158, opened 2023-05-16
Add Py::get for GIL-independent access to frozen classes.

adamreichold:
@davidhewitt Is this what you had in mind for #3154?

The name is an obvious candidate for bikeshedding.

Trying to write an example, I noticed that making `PyCell::get_frozen` public is most likely not useful as there is no way to safely get a `&PyCell` without acquiring the GIL first?

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -211,6 +211,32 @@ Python::with_gil(|py| {
 });
 ```
 
+### frozen classes: Opting out of interior mutability
+
+As detailed above, runtime borrow checking is currently enabled by default. But a class can opt of out it by declaring itself `frozen`. It can still use interior mutability via standard Rust types like `RefCell` or `Mutex`, but it is not bound to the implementation provided by PyO3 and can choose the most appropriate strategy on field-by-field basis.
+
+Classes which are `frozen` and also `Sync`, e.g. they do use `Mutex` but not `RefCell`, can be accessed without needing the Python GIL via the `PyCell::get` and `Py::get` methods:
+
+```rust
+use std::sync::atomic::{AtomicUsize, Ordering};
+# use pyo3::prelude::*;
+
+#[pyclass(frozen)]
+struct FrozenCounter {
+    value: AtomicUsize,
+}
+
+let py_counter: Py<FrozenCounter>  = Python::with_gil(|py| {
+    let counter = FrozenCounter { value: AtomicUsize::new(0) };
+
+    Py::new(py, counter).unwrap()
+});
+
+py_counter.get().value.fetch_add(1, Ordering::Relaxed);
+```
+
+Frozen classes are likely to become the default thereby guiding the PyO3 ecosystem towards a more deliberate application of interior mutability. Eventually, this should enable further optimizations of PyO3's internals and avoid downstream code paying the cost of interior mutability when it is not actually required.
+
 ## Customizing the class
 
 {{#include ../pyclass_parameters.md}}
--- a/guide/pyclass_parameters.md
+++ b/guide/pyclass_parameters.md
@@ -6,7 +6,7 @@
 | `dict` | Gives instances of this class an empty `__dict__` to store custom attributes. |
 | <span style="white-space: pre">`extends = BaseType`</span>  | Use a custom baseclass. Defaults to [`PyAny`][params-1] |
 | <span style="white-space: pre">`freelist = N`</span> |  Implements a [free list][params-2] of size N. This can improve performance for types that are often created and deleted in quick succession. Profile your code to see whether `freelist` is right for you.  |
-| <span style="white-space: pre">`frozen`</span> | Declares that your pyclass is immutable. It removes the borrowchecker overhead when retrieving a shared reference to the Rust struct, but disables the ability to get a mutable reference. |
+| <span style="white-space: pre">`frozen`</span> | Declares that your pyclass is immutable. It removes the borrow checker overhead when retrieving a shared reference to the Rust struct, but disables the ability to get a mutable reference. |
 | `get_all` | Generates getters for all fields of the pyclass. |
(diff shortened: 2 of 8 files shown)


davidhewitt (commented):
Thanks, yes this is what I was thinking!

> Trying to write an example, I noticed that making `PyCell::get_frozen` public is most likely not useful as there is no way to get safely get a `&PyCell` without acquiring the GIL first?

Good point. It may still be nice to have this method for symmetry with `Py`? Also, it might be more performant (it can avoid the `.borrow()` / `.try_borrow()` / `PyRef` dance entirely, though that is probably optimised out anyway for frozen classes).

As for name... I'd quite like to have this just be something concise like `.get()` or `.as_ref()`, though maybe those names make more sense when `frozen` is the default, and it's good to have `frozen` in the name for now?

adamreichold:
> Good point. It may still be nice to have this method for symmetry with Py? Also, it might be more performant (it can avoid the .borrow() / .try_borrow() / PyRef dance entirely, though that is probably optimised out anyway for frozen classes).

I was also thinking that the thread and borrow checking would be optimized out, but then again it does not hurt to give the compiler less work to do. Will add it.

> As for name... I'd quite like to have this just be something concise like .get() or .as_ref(), though maybe those names make more sense when frozen is the default, and it's good to have frozen in the name for now?

I think `get` is fine and will change the name.

> Or I wonder if we should add a new section to the guide for frozen talking about the tradeoff (i.e. you get access to this new method but lose the ability for &mut self in pymethods and borrow_mut).

Indeed, if we want to push the whole ecosystem towards this as a new default, we should probably talk about it in the docs. Its not my favourite part but I'll give it a try.

> (We could potentially also disable PyCell::borrow and Py::borrow for frozen classes, to make the distinction greater that frozen does not need dynamic borrow checking.)

I think we should keep `borrow` as this allows to handle non-frozen (molten?) and frozen classes the same in generic context. A human should not write code borrowing frozen classes directly, but a generic type using `PyRef<T>` also should not have to care whether the class is frozen. I will add a note linking `get` to the docs of `(try_)borrow` though.

adamreichold:
Ok, so I gay the guide part a try. Please have another look when convenient.

guide/src/class.md
```diff
@@ -211,6 +211,32 @@ Python::with_gil(|py| {
 });
 ```
 
+### frozen classes: Opting out of interior mutability
+
+As detailed abvoe, runtime borrow checking is currently enabled by default. But a class can opt of out it by declaring itself `frozen`. It can still use interior mutability via standard Rust types like `RefCell` or `Mutex`, but it is not bound to the implementation provided by PyO3 and can choose the most appropriate strategy on field-by-field basis.
```
davidhewitt, line 216:
```suggestion
As detailed above, runtime borrow checking is currently enabled by default. But a class can opt of out it by declaring itself `frozen`. It can still use interior mutability via standard Rust types like `RefCell` or `Mutex`, but it is not bound to the implementation provided by PyO3 and can choose the most appropriate strategy on field-by-field basis.
```

guide/src/class.md
```diff
@@ -211,6 +211,32 @@ Python::with_gil(|py| {
...
+
+Classes which are `frozen` and also `Sync`, e.g. they do use `Mutex` but not `RefCell`, can also be accessed independently of the acquiring the GIL via the `PyCell::get` and `Py::get` methods:
```
davidhewitt, line 218:
```suggestion
Classes which are `frozen` and also `Sync`, e.g. they do use `Mutex` but not `RefCell`, can be accessed without needing the Python GIL via the `PyCell::get` and `Py::get` methods:
```

guide/src/class.md
```diff
@@ -211,6 +211,32 @@ Python::with_gil(|py| {
...
+    value: AtomicUsize,
+}
+
+let cell  = Python::with_gil(|py| {
+    let counter = FrozenCounter { value: AtomicUsize::new(0) };
+
+    Py::new(py, counter).unwrap()
+});
+
+cell.get().value.fetch_add(1, Ordering::Relaxed);
```
davidhewitt, line 235: It might be helpful to add a type annotation here to show the user what's going on:

```suggestion
let py_frozen_counter: Py<FrozenCounter> = Python::with_gil(|py| {
    let counter = FrozenCounter { value: AtomicUsize::new(0) };

    Py::new(py, counter).unwrap()
});

py_frozen_counter.get().value.fetch_add(1, Ordering::Relaxed);
```

guide/src/class.md
```diff
@@ -211,6 +211,32 @@ Python::with_gil(|py| {
...
+```
+
+In the long run, frozen classes are poised to become the new default thereby guiding the PyO3 ecosystem towards a more deliberate application of interior mutability. Eventually, this should enable further optimizations of PyO3's internals and avoid downstream code paying the cost of interior mutability even though it is not actually required.
```
davidhewitt, line 238: I think it's fine to have this though I may suggest not committing to a timeline:

```suggestion
Frozen classes are likely to become the default thereby guiding the PyO3 ecosystem towards a more deliberate application of interior mutability. Eventually, this should enable further optimizations of PyO3's internals and avoid downstream code paying the cost of interior mutability when it is not actually required.
```

davidhewitt (commented):
👍 thanks - overall looking great, just some small suggestions to wording. Needs tests!

davidhewitt:
EDIT: I see the examples function as tests, it's unfortunate they don't contribute to coverage yet.

adamreichold:
> EDIT: I see the examples function as tests, it's unfortunate they don't contribute to coverage yet.

Yes, this is also a big issue for rust-numpy but I don't think there is anything we can do until https://github.com/taiki-e/cargo-llvm-cov/issues/2 is resolved.

I will add some UI tests nevertheless as the duplication should be harmless. Especially tests that this cannot be used on unfrozen classes should be useful.

adamreichold:
Coverage appears reasonable now.

davidhewitt (approved):
Perfect, thanks!

bors r+

PR #3161, opened 2023-05-18
add PyAny::is_exact_instance and PyAny::is_exact_instance_of

davidhewitt:
Split off from #2881. I'll rebase after that merges.

### DIFF
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -912,6 +912,14 @@ impl PyAny {
         Ok(result == 1)
     }
 
+    /// Checks whether this object is an instance of exactly type `ty` (not a subclass).
+    ///
+    /// This is equivalent to the Python expression `type(self) is ty`.
+    #[inline]
+    pub fn is_exact_instance(&self, ty: &PyAny) -> bool {
+        self.get_type().is(ty)
+    }
+
     /// Checks whether this object is an instance of type `T`.
     ///
     /// This is equivalent to the Python expression `isinstance(self, T)`,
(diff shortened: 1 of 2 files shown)


src/types/any.rs
```diff
@@ -912,6 +912,15 @@ impl PyAny {
         Ok(result == 1)
     }
 
+    /// Checks whether this object is an instance of exactly type `ty` (not a subclass).
+    ///
+    /// This is equivalent to the Python expression `type(self) is ty`.
+    #[inline]
+    pub fn is_exact_instance(&self, ty: &PyType) -> bool {
```
adamreichold, line 919: `is_instance` takes `PyAny` instead of `PyType`. Would that be possible here as well or would the call to `Py_IS_TYPE` be invalid then?

davidhewitt: `Py_IS_TYPE` requires `*mut PyTypeObject` as the second argument, so if we didn't take `&PyType` we'd need to do a type check first. (And maybe return `PyResult<bool`> as a consequence?)

adamreichold: `Py_IS_TYPE` seems to be implemented inline, i.e.

```rust
#[inline]
pub unsafe fn Py_IS_TYPE(ob: *mut PyObject, tp: *mut PyTypeObject) -> c_int {
    (Py_TYPE(ob) == tp) as c_int
}

#[inline]
pub unsafe fn Py_TYPE(ob: *mut PyObject) -> *mut PyTypeObject {
    (*ob).ob_type
}
```

so we do control the type signature to a degree. It also basically uses `tp` as the right-hand side of a pointer comparison meaning that its real type is never required.

I would not suggest we change the public signature as this would be breaking with little cause. But maybe we could implement this as

```rust
pub fn is_exact_instance(&self, ty: &PyAny) -> bool {
    unsafe { (*self.as_ptr()).ob_type == ty.as_ptr() }
}
```

instead?

(IIRC, the reason for taking `PyAny` instead of `PyType` was exactly so that the caller could avoid downcasting just to check a type.)

davidhewitt: I see what you mean. I think we don't need to expand `Py_TYPE`, so how about using safe code?

```rust
pub fn is_exact_instance(&self, ty: &PyAny) -> bool {
    self.get_type().is(ty)
}
```

This should optimize away

adamreichold: Yes, that looks much better and should optimize out. Nice!

davidhewitt:
Thanks for finishing this up @adamreichold 👍

PR #3166, opened 2023-05-20
RFC: Drop EnsureGIL to remove one layer of indirection from the implementation of Python::with_gil

adamreichold:
I am not sure if other people would consider this a simplification as well, but it helped me to remove one layer of indirection used by the implementation of `Python::with_gil`, especially flattening the `Option<_>` layers in `EnsureGIL` and `GILGuard::pool`. This makes it obvious that we always create `GILPool` when we actually acquire the GIL. (Of course, there still might be extra `GILPool` instances created manually.)

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -29,6 +29,7 @@ thread_local! {
 ///  1) for performance
 ///  2) PyGILState_Check always returns 1 if the sub-interpreter APIs have ever been called,
 ///     which could lead to incorrect conclusions that the GIL is held.
+#[inline(always)]
 fn gil_is_acquired() -> bool {
     GIL_COUNT.try_with(|c| c.get() > 0).unwrap_or(false)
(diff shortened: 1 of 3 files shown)


kngwyu (approved):
Cool. Looks good to me. At least, I don't see any benefit of having `Option<GILPool>` in the struct.

adamreichold:
@davidhewitt Since this is strictly refactoring only, I think this would be good to merge based on one review. But then again this code is somewhat involved/unsafe and you wrote most of the module in question. Would you like to give this a review yourself or do you think we can merge this without it?

davidhewitt (commented):
Yep, I think this is good to merge. Nice tidy up! 👍

adamreichold:
bors r=davidhewitt

adamreichold:
bors retry

Issue #3064, opened 2023-03-25
ResourceWarning: Object of type XXX is not untracked before destruction


messense:
While investigating https://github.com/pydantic/pydantic-core/issues/451 I noticed that when running on a debug build of Python, the [`SchemaSerializer`](https://github.com/pydantic/pydantic-core/blob/70dcb069ff20bbab3942f123b32fc35205af2f19/src/serializers/mod.rs#L25-L32) pyclass [with gc protocol support](https://github.com/pydantic/pydantic-core/blob/70dcb069ff20bbab3942f123b32fc35205af2f19/src/serializers/mod.rs#L141-L154) raised a warning

```
ResourceWarning: Object of type pydantic_core._pydantic_core.SchemaSerializer is not untracked before destruction
```

and caused a segfault on debug build of Python only:

```
collected 3134 items / 3133 deselected / 1 selected

tests/validators/test_datetime.py
Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=140737350287360) at ./nptl/pthread_kill.c:44
Download failed: Invalid argument.  Continuing without source file ./nptl/./nptl/pthread_kill.c.
44	./nptl/pthread_kill.c: No such file or directory.
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737350287360) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737350287360) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737350287360, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff7c97476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff7c7d7f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff7c7d71b in __assert_fail_base (fmt=0x7ffff7e32150 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x77bda9 "!_PyErr_Occurred(tstate)", file=0x78fc93 "../Objects/typeobject.c", line=1068, function=<optimized out>)
    at ./assert/assert.c:92
[...]

PR #3168, opened 2023-05-20
Do not apply deferred ref count updates and prevent the GIL from being acquired inside of __traverse__ implementations.

### LINKED ISSUE #2301: Raising warning inside `__traverse__` causes a segfault
Code:

```rust
use pyo3::prelude::*;
use pyo3::py_run;
use pyo3::{PyTraverseError, PyVisit};

#[pyclass]
struct GcIntegration {
    self_ref: PyObject,
}

#[pymethods]
impl GcIntegration {
    fn __traverse__(&self, visit: PyVisit<'_>) -> Result<(), PyTraverseError> {
        Python::with_gil(|py| {
            let warn = py.import("warnings").unwrap().getattr("warn").unwrap();
            warn.call1(("blah",)).unwrap();
        });

        visit.call(&self.self_ref)
    }

    fn __clear__(&mut self) {
        self.self_ref = gil.python().None();
    }
}

fn main() {
    Python::with_gil(|py| {
        let inst = PyCell::new(
            py,
            GcIntegration {
                self_ref: py.None(),
            },
        )
        .unwrap();

        let mut borrow = inst.borrow_mut();
        borrow.self_ref = inst.to_object(py);

        py_run!(py, inst, "import gc; assert inst in gc.get_objects()");
    });

    Python::with_gil(|py| {
        py.run("import gc; gc.collect()", None, None).unwrap();
    });
}
```

Results in:
```
error: process didn't exit successfully: `target\debug\my_module.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

Meta:
```
rustc 1.62.0-nightly (1f7fb6413 2022-04-10)
binary: rustc
commit-hash: 1f7fb6413d6d6c0c929b223e478e44c3db991b03
commit-date: 2022-04-10
host: x86_64-pc-windows-msvc
release: 1.62.0-nightly
LLVM version: 14.0.0

Python 3.10.2 (tags/v3.10.2:a58ebcc, Jan 17 2022, 14:12:15) [MSC v.1929 64 bit (AMD64)]
```

### LINKED ISSUE #3165: `__traverse__` should not alter any refcounts, at least in CPython
## Bug Description

Back when we've got a support for the [GC protocol](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse), while it is not usual for `__traverse__` to do anything more than calling `visit.call`, it was assumed that [running any Python code is not harmful](https://github.com/PyO3/pyo3/pull/2159#issuecomment-1037002492) and currently `__traverse__` trampoline is no more than an ordinary trampoline with a slightly different signature.

It turns out that not only this assumption is wrong but also *any* currently generated `tp_traverse` implementation is prone to crash. CPython has [reused `PyGC_Head` fields to store intermediate refcounts](https://devguide.python.org/internals/garbage-collector/#optimization-reusing-fields-to-save-memory) for a long time, so `tp_traverse` should be never able to read or write any `PyGC_Head` fields.

In particular the `__traverse__` trampoline would first allocate a new `GILPool` and apply deferred refcount updates at once, possibly deallocating some objects in the pool and in turn somehow touching temporarily corrupted `PyGC_Head` (e.g. `PyObject_GC_Untrack` or `PyObject_GC_Del`). For this reason I believe this is the root cause of #1623 and #3064; both suggests that `PyGC_Head` is corrupted and threading seems to be relevant here (it is much easier to defer refcount updates in non-interpreter threads).

## Reproduction

<details><summary>Too lengthy to open by default</summary>

I was able to reliably trigger SIGSEGV on Python 3.8 through 3.11 in x86-64 Linux. Given the following `src/lib.rs` and typical `Cargo.toml` (which has been omitted):

```rust
use pyo3::{prelude::*, PyTraverseError, PyVisit};
use std::{cell::RefCell, thread};

#[pyclass]
struct Ref {
    inner: RefCell<Option<Py<Self>>>,
    clear_on_traverse: bool,
}

impl Ref {
    fn new(clear_on_traverse: bool) -> Self {
        Self {
            inner: RefCell::new(None),
            clear_on_traverse,
        }
    }
}

#[pymethods]
impl Ref {
    fn __traverse__(&self, visit: PyVisit<'_>) -> Result<(), PyTraverseError> {
        if self.clear_on_traverse {
            self.inner.take();
        } else {
            if let Some(inner) = &*self.inner.borrow() {
                visit.call(inner)?;
            }
        }
        Ok(())
    }

    fn __clear__(&mut self) {
        self.inner.take();
    }
}

#[pyfunction]
fn crash1() {
    thread::spawn(move || {
        _ = Python::with_gil(|py| -> PyResult<_> {
            let a = Py::new(py, Ref::new(false))?;
            a.borrow_mut(py).inner.replace(Some(a.clone_ref(py)));

            let b = Py::new(py, Ref::new(false)))?;
            Ok((a, b))
        });
        eprintln!("thread finished");
    });

    Ok(())
}

#[pyfunction]
fn crash2() {
    thread::spawn(move || {
        _ = Python::with_gil(|py| -> PyResult<_> {
            let a1 = Py::new(py, Ref::new(true))?;
            let a2 = Py::new(py, Ref::new(false))?;
            a1.borrow_mut(py).inner.replace(Some(a2.clone_ref(py)));
            a2.borrow_mut(py).inner.replace(Some(a1.clone_ref(py)));

            let b = Py::new(py, Ref::new(false)))?;
            Ok((a1, b))
        });
        eprintln!("thread finished");
    });
}

#[pymodule]
#[pyo3(name = "pyo3_crash_test")]
fn module_entry(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(crash1, m)?)?;
    m.add_function(wrap_pyfunction!(crash2, m)?)?;
    Ok(())
}
```

Running the following Python code will crash inside `gc.collect()`:

```python
import sys
import gc
import time
import pyo3_crash_test

pyo3_crash_test.crash1() # or crash2
time.sleep(0.1) # not strictly required, but ensures that the thread finishes before GC
print('collecting...', file=sys.stderr)
gc.collect()
assert False, 'crash expected during GC'
```

Here `Ref` is holding an optional reference to other `Ref`, possibly itself. Setting `clear_on_traverse` will clear the reference during the traversal, simulating an otherwise totally safe operation to do.

In the `crash1` the refcounts for `a` and `b` will be 2 and 1 respectively at the time of GC. GC will overwrite `PyGC_Head._gc_prev` for both `a` and `b` (effectively converting the list to singly-linked), and then `__traverse__` will be called for `a`, creating a new `GILPool` and triggering `ReferencePool::update_counts`. As both values are registered for `Py_DECREF` the refcount of `b` will be now 0, and this will eventually call `PyObject_GC_Del` which assumes that the list is still doubly-linked and crashes.

The situation is similar for `crash2`: `a2` gets dropped with the GIL still acquired, so the refcounts for `a1`, `a2` and `b` are 2, 1 and 1 respectively. Now assume that `GILPool::new` somehow recognizes `__traverse__` and doesn't call `update_counts`. Even in this case `a1` will now remove a reference to `a2` during `__traverse__`, falsely beliving `Py_DECREF` is safe to call as the GIL is acquired, so the same thing will happen for `a2`.

</details>

## Possible Solution

It seems evident to me that `__traverse__` should be handled as if no GIL is (and can be) acquired. This implies that `Python::with_gil` should probably fail and `gil_is_acquired` should return false inside `__traverse__`.

I believe it is also possible to have multiple nested `__traverse__` calls. The GC process drains the linked list of objects while updating `PyGC_Head` and it sets `gcstate->collecting` to 1 during the process anyway, so `__traverse__` cannot be called twice for the same object and nested calls should be rare, but there are at least three possible paths (as of the main branch of CPython):

* Creating new objects may trigger `PyObject_GC_Track`, which implicitly calls `tp_traverse` in debug builds.
* `gc.get_referrers` and `gc.get_referrents` use `tp_traverse` to collect objects, and are not affected by `gcstate->collecting`. They can be independently nested if `__traverse__` somehow calls those functions.
* While directly calling `gc.collect()` checks for `gcstate->collecting` and returns immediately, the automatic GC scheduling may result in a double GC. Specifically there is an interval between the scheduling (`_Py_ScheduleGC`) and the actual GC (`_Py_RunGC`) and some opcode may end up calling `gc.collect()` inbetween. This is extremely rare though, as this interval can be at most a single opcode, and is otherwise safe.

Probably it is sufficient to add the second "traversal" counter for the nested `__traverse__` calls in addition to `GIL_COUNT`. If the traversal counter is non-zero the GIL cannot be acquired and `gil_is_acquired` should return false regardless of `GIL_COUNT`. The `GIL_COUNT` is still updated as usual, but it is effectively ignored until the traversal counter drops to zero, at which point `update_counts` should be considered (but only when `GIL_COUNT` was non-zero). But I'm not very much confident about this solution for multiple reasons, including that there is no actual guarantee that we can call any other Python API from `tp_traverse` and someday CPython may touch `PyGC_Head` from unexpected functions---I mean, even `gcstate->collecting` check is not complete so who knows? If possible, a fully static solution to prevent anything but `PyVisit` would be much more desirable.

### Suggestions

Given a full solution needs much testing at least, I looked for possible workarounds without a local PyO3 update. Unfortunately `GILPool::new` call is mandatory for the trampoline and the only indirect approach would be temporarily untracking objects via `PyObject_GC_UnTrack`, but that pretty much contradicts why we have `tp_traverse` in the first place.

So it seems that PyO3 needs a minimal update to ensure that you remain safe if you don't do weird things inside `__traverse__`. No promises here, but I hope to file a small PR to disable `update_counts` from the trampoline soon. In my local testing this fixes `crash1` but not `crash2`.

x86-64 Linux (at least)

3.8.10, 3.9.16, 3.10.11, 3.11.13

1.69.0

### Your PyO3 version

0.18.3

[`deadsnakes` PPA](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa), the bundled version of `venv`, `pip` and the most recent version of `maturin` (0.15.2).

adamreichold:
Closes #2301
Closes #3165

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -401,35 +401,13 @@ fn impl_call_slot(cls: &syn::Type, mut spec: FnSpec<'_>) -> Result<MethodAndSlot
 }
 
 fn impl_traverse_slot(cls: &syn::Type, rust_fn_ident: &syn::Ident) -> MethodAndSlotDef {
-    // It is important the implementation of `__traverse__` cannot safely access the GIL,
-    // c.f. https://github.com/PyO3/pyo3/issues/3165, and hence we do not expose our GIL
-    // token to the user code and lock safe methods for acquiring the GIL.
-    // Since we do not create a `GILPool` at all, it is important that our usage of the GIL
-    // token does not produce any owned objects thereby calling into `register_owned`.
     let associated_method = quote! {
         pub unsafe extern "C" fn __pymethod_traverse__(
             slf: *mut _pyo3::ffi::PyObject,
             visit: _pyo3::ffi::visitproc,
             arg: *mut ::std::os::raw::c_void,
-        ) -> ::std::os::raw::c_int
-        {
-            let trap = _pyo3::impl_::panic::PanicTrap::new("uncaught panic inside __traverse__ handler");
-            let py = _pyo3::Python::assume_gil_acquired();
-            let slf = py.from_borrowed_ptr::<_pyo3::PyCell<#cls>>(slf);
-            let borrow = slf.try_borrow();
-            let visit = _pyo3::class::gc::PyVisit::from_raw(visit, arg, py);
-
-            let retval = if let ::std::result::Result::Ok(borrow) = borrow {
-                let _lock = _pyo3::impl_::LockGIL::during_traverse();
-                match ::std::panic::catch_unwind(::std::panic::AssertUnwindSafe(move || borrow.#rust_fn_ident(visit))) {
-                    ::std::result::Result::Ok(res) => _pyo3::impl_::pymethods::unwrap_traverse_result(res),
-                    ::std::result::Result::Err(_err) => -1,
-                }
-            } else {
-                0
-            };
-            trap.disarm();
-            retval
+        ) -> ::std::os::raw::c_int {
+            _pyo3::impl_::pymethods::call_traverse_impl::<#cls>(slf, #cls::#rust_fn_ident, visit, arg)
         }
     };
     let slot_def = quote! {
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -401,6 +401,11 @@ fn impl_call_slot(cls: &syn::Type, mut spec: FnSpec<'_>) -> Result<MethodAndSlot
 }
 
 fn impl_traverse_slot(cls: &syn::Type, rust_fn_ident: &syn::Ident) -> MethodAndSlotDef {
+    // It is important the implementation of `__traverse__` cannot safely access the GIL,
+    // c.f. https://github.com/PyO3/pyo3/issues/3165, and hence we do not expose our GIL
+    // token to the user code and lock safe methods for acquiring the GIL.
+    // Since we do not create a `GILPool` at all, it is important that our usage of the GIL
+    // token does not produce any owned objects thereby calling into `register_owned`.
     let associated_method = quote! {
         pub unsafe extern "C" fn __pymethod_traverse__(
             slf: *mut _pyo3::ffi::PyObject,
@@ -409,14 +414,17 @@ fn impl_traverse_slot(cls: &syn::Type, rust_fn_ident: &syn::Ident) -> MethodAndS
         ) -> ::std::os::raw::c_int
         {
             let trap = _pyo3::impl_::panic::PanicTrap::new("uncaught panic inside __traverse__ handler");
-            let pool = _pyo3::GILPool::new();
-            let py = pool.python();
+            let py = _pyo3::Python::assume_gil_acquired();
             let slf = py.from_borrowed_ptr::<_pyo3::PyCell<#cls>>(slf);
-
-            let visit = _pyo3::class::gc::PyVisit::from_raw(visit, arg, py);
             let borrow = slf.try_borrow();
+            let visit = _pyo3::class::gc::PyVisit::from_raw(visit, arg, py);
+
             let retval = if let ::std::result::Result::Ok(borrow) = borrow {
-                _pyo3::impl_::pymethods::unwrap_traverse_result(borrow.#rust_fn_ident(visit))
+                let _lock = _pyo3::impl_::LockGIL::during_traverse();
+                match ::std::panic::catch_unwind(::std::panic::AssertUnwindSafe(move || borrow.#rust_fn_ident(visit))) {
+                    ::std::result::Result::Ok(res) => _pyo3::impl_::pymethods::unwrap_traverse_result(res),
+                    ::std::result::Result::Err(_err) => -1,
+                }
             } else {
                 0
             };
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -29,11 +29,16 @@ thread_local_const_init! {
     /// they are dropped.
     ///
     /// As a result, if this thread has the GIL, GIL_COUNT is greater than zero.
-    static GIL_COUNT: Cell<usize> = const { Cell::new(0) };
+    ///
+    /// Additionally, we sometimes need to prevent safe access to the GIL,
+    /// e.g. when implementing `__traverse__`, which is represented by a negative value.
+    static GIL_COUNT: Cell<isize> = const { Cell::new(0) };
 
     /// Temporarily hold objects that will be released when the GILPool drops.
     static OWNED_OBJECTS: RefCell<Vec<NonNull<ffi::PyObject>>> = const { RefCell::new(Vec::new()) };
 }
+
+const GIL_LOCKED_DURING_TRAVERSE: isize = -1;
 
 /// Checks whether the GIL is acquired.
 ///
@@ -286,7 +291,7 @@ static POOL: ReferencePool = ReferencePool::new();
 
 /// A guard which can be used to temporarily release the GIL and restore on `Drop`.
 pub(crate) struct SuspendGIL {
-    count: usize,
+    count: isize,
     tstate: *mut ffi::PyThreadState,
 }
 
@@ -308,6 +313,30 @@ impl Drop for SuspendGIL {
             // Update counts of PyObjects / Py that were cloned or dropped while the GIL was released.
             POOL.update_counts(Python::assume_gil_acquired());
         }
+    }
+}
+
+/// Used to lock safe access to the GIL
+pub struct LockGIL {
+    count: isize,
+}
+
+impl LockGIL {
+    /// Lock access to the GIL while an implementation of `__traverse__` is running
+    pub fn during_traverse() -> Self {
+        Self::new(GIL_LOCKED_DURING_TRAVERSE)
+    }
+
+    fn new(reason: isize) -> Self {
+        let count = GIL_COUNT.with(|c| c.replace(reason));
+
+        Self { count }
+    }
+}
+
+impl Drop for LockGIL {
+    fn drop(&mut self) {
+        GIL_COUNT.with(|c| c.set(self.count));
     }
 }
 
@@ -421,7 +450,17 @@ pub unsafe fn register_owned(_py: Python<'_>, obj: NonNull<ffi::PyObject>) {
 #[inline(always)]
 fn increment_gil_count() {
     // Ignores the error in case this function called from `atexit`.
-    let _ = GIL_COUNT.try_with(|c| c.set(c.get() + 1));
+    let _ = GIL_COUNT.try_with(|c| {
+        let current = c.get();
+        match current {
+            GIL_LOCKED_DURING_TRAVERSE => panic!(
+                "Access to the GIL is prohibited while a __traverse__ implmentation is running."
+            ),
+            current if current < 0 => panic!("Access to the GIL is currently prohibited."),
+            _ => (),
+        }
+        c.set(current + 1);
+    });
 }
 
 /// Decrements pyo3's internal GIL count - to be called whenever GILPool or GILGuard is dropped.
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -332,6 +332,16 @@ impl LockGIL {
 
         Self { count }
     }
+
+    #[cold]
+    fn bail(current: isize) {
+        match current {
+            GIL_LOCKED_DURING_TRAVERSE => panic!(
+                "Access to the GIL is prohibited while a __traverse__ implmentation is running."
+            ),
+            _ => panic!("Access to the GIL is currently prohibited."),
+        }
+    }
 }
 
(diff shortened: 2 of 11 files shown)


pyo3-macros-backend/src/pymethod.rs
```diff
@@ -409,14 +414,17 @@ fn impl_traverse_slot(cls: &syn::Type, rust_fn_ident: &syn::Ident) -> MethodAndS
...
             let slf = py.from_borrowed_ptr::<_pyo3::PyCell<#cls>>(slf);
-
-            let visit = _pyo3::class::gc::PyVisit::from_raw(visit, arg, py);
             let borrow = slf.try_borrow();
+            let visit = _pyo3::class::gc::PyVisit::from_raw(visit, arg, py);
+
             let retval = if let ::std::result::Result::Ok(borrow) = borrow {
-                _pyo3::impl_::pymethods::unwrap_traverse_result(borrow.#rust_fn_ident(visit))
+                let _lock = _pyo3::impl_::LockGIL::new();
+                match ::std::panic::catch_unwind(::std::panic::AssertUnwindSafe(move || borrow.#rust_fn_ident(visit))) {
```
adamreichold, line 424: I am not sure if we want to the `catch_unwind` dance here? It does seem necessary to make this testable as the `PanicTrap` aborts the process otherwise. But what is the desired behaviour in production code? Is `tp_traverse` returning -1 sufficient or do we actually want to abort the process when this happens?

(I also wonder whether this dance should be encapsulated better, e.g. replace `unwrap_traverse_result` by `call_traverse_impl` which does that and creates the `LockGIL` instance and handles any resulting panic.)

adamreichold: (Added a commit which moves as much of this as possible into a library function instead of expanding it in the macros.)

adamreichold: After thinking about this some more, we definitely do not want to abort the process because something accidentally (and possibly indirectly) call `with_gil`. Furthermore, the `catch_unwind` dance is the same we do for the other trampolines.

src/gil.rs
```diff
@@ -290,7 +293,7 @@ static POOL: ReferencePool = ReferencePool::new();
 
 /// A guard which can be used to temporarily release the GIL and restore on `Drop`.
 pub(crate) struct SuspendGIL {
-    count: usize,
+    count: isize,
```
davidhewitt, line 296: Now that `acquire_gil` is not a thing, this count is limited to recursive stack frames, so I wonder if `i16` or `i32` might be better?

src/gil.rs
```diff
@@ -425,7 +449,14 @@ pub unsafe fn register_owned(_py: Python<'_>, obj: NonNull<ffi::PyObject>) {
 #[inline(always)]
 fn increment_gil_count() {
     // Ignores the error in case this function called from `atexit`.
-    let _ = GIL_COUNT.try_with(|c| c.set(c.get() + 1));
+    let _ = GIL_COUNT.try_with(|c| {
+        let current = c.get();
+        assert!(
+            current >= 0,
+            "Access to the GIL is currently prohibited, for example because a `__traverse__` implementation is currently running."
+        );
```
davidhewitt, line 457: We could potentially even use different negative integers to emit different error messages, if we wanted to be more definite here that a `__traverse__` _is_ running.

src/impl_/pymethods.rs
```diff
@@ -239,14 +245,46 @@ impl PySetterDef {
...
+                Ok(()) => 0,
+                Err(PyTraverseError(value)) => value,
+            },
+            Err(_err) => -1,
+        }
+    } else {
+        0
+    };
+    trap.disarm();
+    retval
```
davidhewitt, line 287: I feel like I want to refactor this to have a non-generic `fn inner` portion somehow, but the only way I see right now would involve dynamic dispatch and I don't know if that's any better...

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -406,22 +406,8 @@ fn impl_traverse_slot(cls: &syn::Type, rust_fn_ident: &syn::Ident) -> MethodAndS
...
-            let borrow = slf.try_borrow();
-            let retval = if let ::std::result::Result::Ok(borrow) = borrow {
-                _pyo3::impl_::pymethods::unwrap_traverse_result(borrow.#rust_fn_ident(visit))
-            } else {
-                0
-            };
-            trap.disarm();
-            retval
+        ) -> ::std::os::raw::c_int {
+            _pyo3::impl_::pymethods::call_traverse_impl::<#cls>(slf, #cls::#rust_fn_ident, visit, arg)
```
davidhewitt, line 410: +1 for moving this implementation out of the generated code!

tests/ui/traverse_bare_self.stderr
```diff
@@ -0,0 +1,17 @@
...
+9 | |     fn __traverse__(slf: PyRef<Self>, visit: PyVisit) {}
+  | |___________________^ expected fn pointer, found fn item
+  |
+  = note: expected fn pointer `for<'a, 'b> fn(&'a TraverseTriesToTakePyRef, PyVisit<'b>) -> Result<(), PyTraverseError>`
+                found fn item `for<'a, 'b> fn(pyo3::PyRef<'a, TraverseTriesToTakePyRef>, PyVisit<'b>) {TraverseTriesToTakePyRef::__traverse__}`
+note: function defined here
+ --> src/impl_/pymethods.rs
+  |
+  | pub unsafe fn call_traverse_impl<T>(
+  |               ^^^^^^^^^^^^^^^^^^
```
davidhewitt, line 17: To make this error nicer we could make the proc-macro check that the `__traverse__` implementation has `&self` receiver? Happy to leave this as an issue for a follow-up in another PR.

davidhewitt (commented):
Nice, yes I'm happy with things going in this direction. I added #2301 to the "closes" section in your OP.

I had quite a few tangential thoughts as I read through this one...

src/gil.rs
adamreichold: Since this is a single value (and not safe an array of them), I think anything below register size (hence `?size`) should not make a meaningful difference? With `i16` for example, I would want to audit the code for overflows and use checked operations.

I could settle for `i32`, but this already is `i32` for 32 bit platforms and I don't think it will make difference on 64 bit hardware.

tests/ui/traverse_bare_self.stderr
adamreichold: Due to my innate love for proc-macros, I would be glad if we could move this into a follow-up issue so it can get the proper attention it so rightfully deserves. ;)

src/impl_/pymethods.rs
adamreichold: Yes, I pondered this as well but since we need to produce the `PyRef<T>` in generic context and this also requires already having the panic trap on the stack, there is little code left which will end up in the inner function, basically just:

```rust
    let visit = PyVisit::from_raw(visit, arg, py);

        let _lock = LockGIL::new();

        match catch_unwind(AssertUnwindSafe(move || bound_impl(visit))) {
            Ok(res) => match res {
                Ok(()) => 0,
                Err(PyTraverseError(value)) => value,
            },
            Err(_err) => -1,
        }
```

src/gil.rs
adamreichold: Will add a named constant and a definite assertion message.

src/gil.rs
davidhewitt: Fair point on register size, let's leave this as it is. 👍

src/impl_/pymethods.rs
davidhewitt: Indeed; let's not worry about this for now.

tests/ui/traverse_bare_self.stderr
davidhewitt: Hehe absolutely 🤣

adamreichold: That issue is #3180

src/gil.rs
adamreichold: Add the constant and assertion message. It does look a bit more complicated now and out of an abundance of caution, I moved choosing the correct panic message into a cold function.

davidhewitt (approved):
Looks great to me, thanks!

I see you have linked this to #3064, it's not quite clear to me why?

adamreichold:
> I see you have linked this to https://github.com/PyO3/pyo3/issues/3064, it's not quite clear to me why?

I misunderstood it as one of the symptoms described in #3165 but reading it less tired, that was nonsense. I removed the reference will give bors a try.

bors r=davidhewitt

adamreichold:
So this was hit by the same problem that #3171 is trying to fix. I just moved the test case to the 1.63 section for now.

bors retry

PR #3171, opened 2023-05-21
RFC: Simplify version handling of UI tests.

adamreichold:
Due to checking in error outputs these tests only work reliably on stable and not on intermediate version between our MSRV (currently 1.48) and the current stable version.

Hence this simplifies things to run only MSRV-compatible tests for the MSRV builds, anything else for stable builds except for those tests which require the nightly feature, i.e. the `Ungil` being distinct from the `Send` trait.

Finally, `not_send3` is disabled when using the nightly feature since while `Rc` is not send, it also not GIL-bound and hence can be passed into `allow_threads` as it does not actually spawn a new thread.


adamreichold:
bors try

adamreichold:
bors try

davidhewitt:
> Due to checking in error outputs these tests only work reliably on stable and not on intermediate version between our MSRV (currently 1.48) and the current stable version.

Can you clarify where the current tests broke? It was a best effort but I was hoping that by gating tests on each version in the way it was currently done it should be the case that when a test output changed it would be bumped to run only on that rust version or newer. The effect hopefully being that older intermediate versions would only run tests which that version produced the same output as current stable.

adamreichold:
> Can you clarify where the current tests broke? 

I ran the tests using 1.63 and immediately, meaning on `main` unmodified, got changes in the error output of 8 test cases.

> It was a best effort but I was hoping that by gating tests on each version in the way it was currently done it should be the case that when a test output changed it would be bumped to run only on that rust version or newer. 

I suspect we failed to do this at some point in the past when we updated the expected error outputs to match a new stable release. These test cases would have had to be moved to a later `rustversion::since`-gated section since their output retroactively changed from stable to stable, but we did not do that and hence they are broken when using an older compiler.

Personally, I don't think the effort to keep this aligned is worth it: We do not test these intermediate versions in the CI. (Which is also why we did not notice them breaking.) Even for MSRV and nightly, it seems enough to check that these test cases fail to build as expected. The detailed error message could then just be kept in sync with the current stable release and that would be it. Hence the change proposed here.

(With this change, we could also drop the dev dependency on `rustversion` after the MSRV bump to 1.56 because the only other use (besides xtask) is in the tests that do not even parse on 1.48.)

.github/workflows/build.yml
```diff
@@ -59,6 +59,10 @@ jobs:
         name: Prepare minimal package versions (MSRV only)
         run: nox -s set-minimal-package-versions
 
+      - if: inputs.rust == 'nightly' || inputs.msrv == 'MSRV'
+        name: Ignore changed error messages when using trybuild
+        run: echo "TRYBUILD=overwrite" >> "$GITHUB_ENV"
```
davidhewitt, line 64: This is very smart! That said, I think it would be nice if we just skipped running on msrv completely, to save some CPU cycles.

Maybe as a first line in the test we could have `if env!(PYO3_SKIP_UI_TEST) { return }` or similar? This might also solve the distro packagers problem discussed (they can probably live without UI tests).

davidhewitt (commented):
Ah right. That makes sense and I'm in favour of simplification.

So the one concern I have is for system package distributors, who I think may sometimes run `cargo test` as part of their package build / distribution? If that's the case I suspect they use the distro's rust, which probably isn't latest stable.

That said, as you note, it's already broken. I suppose that if we go this route and only support the UI tests on latest stable, we can perhaps just recommend any distro packagers who hit this just ignore the UI tests.

.github/workflows/build.yml
adamreichold: I am not sure if the CPU cycles are worth it, but the tests do still serve a purpose even on MSRV insofar as they check that the UI tests do fail to compile (even though with maybe a different error message compared to stable).

tests/test_compile_error.rs
```diff
@@ -45,90 +18,31 @@ fn _test_compile_errors() {
...
+    t.compile_fail("tests/ui/not_send2.rs");
+    t.compile_fail("tests/ui/get_set_all.rs");
+    #[cfg(not(feature = "nightly"))]
+    t.compile_fail("tests/ui/not_send3.rs");
+    #[cfg(feature = "nightly")]
     t.compile_fail("tests/ui/not_send_auto_trait.rs");
+    #[cfg(feature = "nightly")]
     t.compile_fail("tests/ui/not_send_auto_trait2.rs");
+    #[cfg(feature = "nightly")]
     t.compile_fail("tests/ui/send_wrapper.rs");
```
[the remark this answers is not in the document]
adamreichold: While I agree in principle, I am not sure this is realistic as nightly is such a fast moving target. For example, the error messages for all three nightly-only tests do not match the output of today's nightly.

Maybe checking for expected pass/fail is sufficient?

adamreichold:
> So the one concern I have is for system package distributors, who I think may sometimes run cargo test as part of their package build / distribution? If that's the case I suspect they use the distro's rust, which probably isn't latest stable.

One idea I see to fix this is to just not ship `tests/test_compile_error.rs` on crates.io, i.e. adding it to `package.exclude` via `Cargo.toml` so that package distributors using the sources from crates.io do not get this particular test case at all. (We could also exclude `tests/ui` then to save a few bytes.)

While this may appear a bit heavy-handed, alternatives like environment variables or Cargo features seem to lose out on discoverability and brittleness.

davidhewitt:
> One idea I see to fix this is to just not ship tests/test_compile_error.rs on crates.io, i.e. adding it to package.exclude via Cargo.toml so that package distributors using the sources from crates.io do not get this particular test case at all. (We could also exclude tests/ui then to save a few bytes.)

That sounds like a very reasonable idea, as long as we don't have other references to those files which might otherwise break the distributed crate.

I suppose some vendors might be using the GitHub archive, they can always strip or ignore the UI tests themselves.

tests/test_compile_error.rs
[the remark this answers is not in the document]
davidhewitt: Sure thing, we can probably cover that using doctests and remove these nightly UI tests completely?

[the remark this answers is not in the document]
adamreichold: Done.

adamreichold:
I excluded the UI tests from the package for crates.io and converted the nightly-related UI tests into doc tests which is actually nice since it gives a bit more context. Please have another look.

src/marker.rs
```diff
@@ -140,6 +140,42 @@ use std::os::raw::c_int;
...
+/// Python::with_gil(|py| {
+///     let rc = Rc::new(42);
+///
+///     py.allow_threads(|| {
+///         println!("{:?}", rc);
+///     });
+/// });
+/// ```
+///
+/// This also implies that one can circumvent this protection using e.g. the [`send_wrapper`](https://docs.rs/send_wrapper/) crate:
```
davidhewitt, line 160: Maybe it should be made clearer here that doing this is not sound and the fix is available on nightly Rust. Possibly also link to #2141?

Actually, looking at https://github.com/PyO3/pyo3/issues/2141#issuecomment-1028921883 again I am now wondering if the time is right to make some changes to `allow_threads` to make it sound even in the presence of `send_wrapper`?

We've done the removal of `GilGuard` now so we can potentially offer a safe-but-awkward API and then an `unsafe` but neater one? We can then document the lack of auto traits as hampering us (or maybe someone comes up with a better idea how to model this conveniently + safely).

davidhewitt (commented):
A really nice improvement! Just one comment motivated by our new docs, which we might want to spin off into a separate discussion...

src/marker.rs
adamreichold: > Maybe it should be made clearer here that doing this is not sound and the fix is available on nightly Rust. Possibly also link to https://github.com/PyO3/pyo3/issues/2141?

Will amend the comment.

W.r.t. `allow_threads`, if you are referring to adding the `'static` bound then I think your observation that 

> I don't think 'static is very practical, because then the closure can't capture any variables except owned values by move. I think that'll turn out quite awkward.

still applies. For example, this would make `allow_threads` not work for rust-numpy when a `PyArray` is borrowed as an `ArrayView<'_>` which we could not safely work on without keeping the GIL held (even though we actually do proper borrow checking now).

While I agree that our usage of `Send` is not really sound, I don't fixing this now and without auto traits is worth it. Especially so as using `send_wrapper` is usually a big warning sign that something strange is going. Meaning that I don't think this will happen accidentally.

On the other hand, pushing currently and eventually safe usage towards unsafe code will make this more difficult for every user which I suspect will mean more errors in aggregate compared to us being unable to close the `send_wrapper` loophole.

davidhewitt: Fair enough, let's leave `allow_threads` unchanged for now. I've added #2141 to the 1.0 project to remind us to see if we can work out a better solution over time.

adamreichold: I added more explicit language concerning the soundness loophole and its fix. I also included the converse of the `Rc`-`compile_fail` doctest in the nightly version.

davidhewitt (approved):
Looks great, sorry for the delayed approval!

bors r+

adamreichold:
bors retry

PR #3175, opened 2023-05-22
Additional tests for #3168

lifthrasiir:
These were a part of tests I was preparing for #3165, and I believe it's worthy to add them (any single of them fails in the current main branch).

### DIFF
--- a/guide/src/class/protocols.md
+++ b/guide/src/class/protocols.md
@@ -407,6 +407,10 @@ impl ClassWithGCSupport {
 }
 ```
 
+Usually, an implementation of `__traverse__` should do nothing but calls to `visit.call`.
+Most importantly, safe access to the GIL is prohibited inside implementations of `__traverse__`,
+i.e. `Python::with_gil` will panic.
+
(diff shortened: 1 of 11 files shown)


adamreichold:
> any single of them fails in the current main branch

I am confused now: They do seem to pass in the CI? Do they fail only intermittently? Could you also check whether #3168 does resolve the issues you found?

lifthrasiir:
Sorry, your PR definitely passes all of those tests. I meant to say that those tests are crafted so that PyO3 don't pass them without your PR.

adamreichold:
Ah sorry, I missed the target branch of this PR! It is targetting #3168 and not `main`.

Also sorry for not starting with that: Thank you for going the extra mile and packaging up your tests so we can include them upstream!

DataTriny:
Hello @lifthrasiir 

Thank you!

adamreichold (approved):
Thanks again for this comprehensive set of tests.

I'll hold off merging this into #3168 until we have consensus that that approach is what we will pursue to avoid that this PR needs to be recreated if we decide to pursue another approach.

PR #3176, opened 2023-05-22
Prevent dropping unsendable classes on other threads.

adamreichold:
Continuing the discussed from https://github.com/PyO3/pyo3/pull/3169#issuecomment-1556571504 and https://github.com/PyO3/pyo3/pull/3169#issuecomment-1556661723:

We already have checks in place to avoid borrowing these classes on other threads but it was still possible to send them to another thread and drop them there (while holding the GIL).
    
This change avoids running the `Drop` implementation in such a case even though Python will still free the underlying memory. This might leak resources owned by the object, but it avoids undefined behaviour due to access the unsendable type from another thread.
    
This does assume that the object was not unsafely integrated into an intrusive data structures which still point to the now freed memory. In that case, the only recourse would be to abort the process as freeing the memory is unavoidable when the tp_dealloc slot is called. (And moving it elsewhere into a new allocation would still break any existing pointers.)

### DIFF
--- a/tests/test_class_basics.rs
+++ b/tests/test_class_basics.rs
@@ -526,3 +526,41 @@ fn access_frozen_class_without_gil() {
 
     assert_eq!(py_counter.get().value.load(Ordering::Relaxed), 1);
 }
+
+#[test]
+fn drop_unsendable_elsewhere() {
+    use std::sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    };
+    use std::thread::spawn;
+
+    #[pyclass(unsendable)]
+    struct Unsendable {
+        dropped: Arc<AtomicBool>,
+    }
+
+    impl Drop for Unsendable {
+        fn drop(&mut self) {
+            self.dropped.store(true, Ordering::SeqCst);
+        }
+    }
+
+    let dropped = Arc::new(AtomicBool::new(false));
+
+    let unsendable = Python::with_gil(|py| {
+        let dropped = dropped.clone();
+
+        Py::new(py, Unsendable { dropped }).unwrap()
+    });
+
+    spawn(move || {
+        Python::with_gil(move |_py| {
+            drop(unsendable);
+        });
+    })
+    .join()
+    .unwrap();
+
+    assert!(!dropped.load(Ordering::SeqCst));
+}
--- a/tests/test_class_basics.rs
+++ b/tests/test_class_basics.rs
@@ -228,31 +228,40 @@ impl UnsendableChild {
 }
 
 fn test_unsendable<T: PyClass + 'static>() -> PyResult<()> {
-    let obj = std::thread::spawn(|| -> PyResult<_> {
+    let obj = Python::with_gil(|py| -> PyResult<_> {
+        let obj: Py<T> = PyType::new::<T>(py).call1((5,))?.extract()?;
+
+        // Accessing the value inside this thread should not panic
+        let caught_panic =
+            std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| -> PyResult<_> {
+                assert_eq!(obj.as_ref(py).getattr("value")?.extract::<usize>()?, 5);
+                Ok(())
+            }))
+            .is_err();
+
+        assert!(!caught_panic);
+        Ok(obj)
+    })?;
+
+    let keep_obj_here = obj.clone();
+
+    let caught_panic = std::thread::spawn(move || {
+        // This access must panic
         Python::with_gil(|py| {
-            let obj: Py<T> = PyType::new::<T>(py).call1((5,))?.extract()?;
-
-            // Accessing the value inside this thread should not panic
-            let caught_panic =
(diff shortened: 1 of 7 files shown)


adamreichold:
> I think there's a further question about what to do with #[pyclass(unsendable)]. We do have the option of leaking because we could still move the T into a box and forget it. I think at the very least we should also use PyErr::write_unraisable to report the issue.

I am wondering whether someone might (unsafely) depend on `Py<T>` having a stable address on the Python heap after construction? Is that actually true for tracing GC like PyPy?

Put differently, I think we could actually just not run `Drop` but free the memory as anything still requiring that memory to exist would need a pointer to it anyway, e.g. some kind of intrusive data structure. But this pointer would by invalidated by moving the value into a new box.

(We could enforcing boxing up front for `unsendable` but that seems like a cure that is worse than the illness as `!Send` is often used for performance reasons, e.g. `Arc` versus `Rc.)

> I'm now feeling somewhat tempted to deprecate unsendable to simplify the framework and just put T: Send as a general #[pyclass] requirement in a couple releases time, if users don't report a compelling reason to keep it.

With my maintainer hat on, I whole-heartedly agree. But I actually am a user of that feature: I had simulations which we not thread safe internally for performance reasons (but parallelized by running multiple instances in multiple processes). They would never be moved between threads (or actually be used in multi-threaded programs) and I would be totally fine with aborting the process if dropping these classes on the wrong thread was attempted.

adamreichold:
> Put differently, I think we could actually just not run Drop but free the memory as anything still requiring that memory to exist would need a pointer to it anyway, e.g. some kind of intrusive data structure. But this pointer would by invalidated by moving the value into a new box.

Meaning that if we do want to give the guarantee of `Py<T>` being pinned in place after construction, we would have to abort instead of just moving `T` into a newly allocated box.

davidhewitt:
> I am wondering whether someone might (unsafely) depend on Py<T> having a stable address on the Python heap after construction? Is that actually true for tracing GC like PyPy?

At the moment I think the nature of the Python C-API exposing objects as `*mut PyObject` more or less forces them to have a stable address, I believe PyPy is even forced to make objects stable to work with the C-API. This is one of the attractive features of HPy just exposing opaque handles; PyPy can implement more efficiently.

It's probably best if we don't rely on `Py<T>` having a stable address in case we ever want to support HPy bindings.

So I think this means we're in agreement that if the object is sent to another thread, on drop we can report it with `write_unraisable` and then just skip running [Drop for that object](https://github.com/PyO3/pyo3/blob/a3c4fd2fa07cda81ac31d97c8444c74d772179f5/src/pycell.rs#L940)?

Python will then free the memory anyway as it controls the allocation.

adamreichold:
> So I think this means we're in agreement that if the object is sent to another thread, on drop we can report it with write_unraisable and then just skip running [Drop for that object](https://github.com/PyO3/pyo3/blob/a3c4fd2fa07cda81ac31d97c8444c74d772179f5/src/pycell.rs#L940)?

Yes, I think this the best course of action. Please feel free to hijack this PR if you have time available. Otherwise, I will try to implement the above here eventually.

davidhewitt:
I'll see how things shape up during the week; recently I have been able to scrap together just enough time in the evenings to comment on some of the backlog of issues and PRs in my inbox :)

adamreichold:
> I'll see how things shape up during the week; recently I have been able to scrap together just enough time in the evenings to comment on some of the backlog of issues and PRs in my inbox :)

I added the minimal implementation here and extended the test case to actually check the intended effect, i.e. that `Drop` is not invoked. Please have a look when there this time.

tests/test_class_basics.rs
```diff
@@ -526,3 +526,41 @@ fn access_frozen_class_without_gil() {
...
+
+    spawn(move || {
+        Python::with_gil(move |_py| {
+            drop(unsendable);
+        });
+    })
+    .join()
+    .unwrap();
+
+    assert!(!dropped.load(Ordering::SeqCst));
```
davidhewitt, line 565: It might be nice to add an assertion that the unraisable exception was printed. See the `UnraisableCapture` type used in a couple of tests, could potentially be moved to `tests/common.rs`.

davidhewitt (commented):
Looks great to me, thanks! Just one thought on the test.

tests/test_class_basics.rs
adamreichold: This was a good call indeed as it made me notice that our existing two `test_unsendable` tests actually had this problem all along, creating their types on one thread, then checking that accessing it on another thread panics but also dropping them there instead of on the originating thread.

davidhewitt: Aha that's a great result 😂

davidhewitt (approved):
💯 

bors r+

davidhewitt:
Ah, the tests using `unraisablehook` only can run on 3.8 or higher. I can fixup this evening.

adamreichold:
Added the relevant `#[cfg(..)]` directives.

bors r=davidhewitt

adamreichold:
And let's try again, but not on WASM...

bors retry

PR #3177, opened 2023-05-23
RFC: Use a const initializer for `GIL_COUNT` if possible

lifthrasiir:
Normally [`LocalKey::try_with`](https://doc.rust-lang.org/stable/std/thread/struct.LocalKey.html#method.try_with) needs to check for the initialization flag to lazily initialize the TLS. This behaves badly with a compiler optimization and it seems that multiple calls to `gil_is_acquired()` cannot be correctly eliminated. Rust 1.59 added a `const { ... }` initializer (a special form only allowed here for now) in `thread_local!` which removes the initialization flag, allowing those kind of optimizations.

I should note that the performance impact is probably minimal, because the check branch is extremely well predicted and the optimization is only possible when every PyO3 code that leads to `gil_is_acquired()` is inlined, a pretty uncommon situation. I couldn't demonstrate any consistent improvement or regression from my machine, which performance seems to be flaky as well. But at least we would have one less branch there. I'll leave this as an RFC so that someone else can prove or disprove that this is indeed beneficial.

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -15,8 +15,8 @@ cfg_if::cfg_if! {
         use std::thread_local as thread_local_const_init;
     } else {
         macro_rules! thread_local_const_init {
-            ($(#[$attr:meta])* static $name:ident: $ty:ty = const { $init:expr };) => (
-                thread_local! { $(#[$attr])* static $name: $ty = $init; }
+            ($($(#[$attr:meta])* static $name:ident: $ty:ty = const { $init:expr };)*) => (
+                thread_local! { $($(#[$attr])* static $name: $ty = $init;)* }
             )
         }
     }
@@ -30,11 +30,9 @@ thread_local_const_init! {
     ///
     /// As a result, if this thread has the GIL, GIL_COUNT is greater than zero.
     static GIL_COUNT: Cell<usize> = const { Cell::new(0) };
-}
 
-thread_local! {
     /// Temporarily hold objects that will be released when the GILPool drops.
-    static OWNED_OBJECTS: RefCell<Vec<NonNull<ffi::PyObject>>> = RefCell::new(Vec::with_capacity(256));
+    static OWNED_OBJECTS: RefCell<Vec<NonNull<ffi::PyObject>>> = const { RefCell::new(Vec::new()) };
 }
 
 /// Checks whether the GIL is acquired.
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -10,15 +10,29 @@ use std::{mem, ptr::NonNull, sync::atomic};
 
 static START: Once = Once::new();
 
-thread_local! {
+cfg_if::cfg_if! {
+    if #[cfg(thread_local_const_init)] {
+        use std::thread_local as thread_local_const_init;
+    } else {
+        macro_rules! thread_local_const_init {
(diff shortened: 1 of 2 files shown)


src/gil.rs
```diff
@@ -10,15 +10,29 @@ use std::{mem, ptr::NonNull, sync::atomic};
...
     /// they are dropped.
     ///
     /// As a result, if this thread has the GIL, GIL_COUNT is greater than zero.
-    static GIL_COUNT: Cell<usize> = Cell::new(0);
+    static GIL_COUNT: Cell<usize> = const { Cell::new(0) };
+}
 
+thread_local! {
     /// Temporarily hold objects that will be released when the GILPool drops.
     static OWNED_OBJECTS: RefCell<Vec<NonNull<ffi::PyObject>>> = RefCell::new(Vec::with_capacity(256));
```
adamreichold, line 37: I wonder whether we should apply this here as well, of course using `RefCell::new(Vec::new())` as the initializer. The flag needs to be checked every time while the initial allocation should be amortized after the first few calls.

adamreichold:
> I should note that the performance impact is probably minimal, because the check branch is extremely well predicted and the optimization is only possible when every PyO3 code that leads to gil_is_acquired() is inlined, a pretty uncommon situation. I couldn't demonstrate any consistent improvement or regression from my machine, which performance seems to be flaky as well. But at least we would have one less branch there. I'll leave this as an RFC so that someone else can prove or disprove that this is indeed beneficial.

I think we should apply this change even if we cannot measure an improvement, as long as we do not measure a regression. The reason is that the effects will most likely be hard to measure as you point, mostly due to branch prediction hiding the costs of the branch. But it still one branch less and hence less pressure on the global resources backing branch prediction in the first place. Meaning that effects would most likely only be measurable when these global resources are oversubscribed.

I also think the change is nicely localized and carries a reasonable maintenance cost which we will loose eventually as our MSRV passes 1.59 (which are targetting 1.56 for PyO3 0.20 at the moment, but nothing is set in stone yet).

src/gil.rs
adamreichold: (Maybe to add: I am not a huge fan of unconditionally pre-allocating long-lived buffers in the first place. I think pre-allocation is best applied when I have something to base the size in and am sure that I will use the allocation, e.g. collecting an iterator and pre-allocating based on its size hint.)

davidhewitt: It could be worth a go to see if this does affect benchmarks in any way, I think we likely access this more frequently than `GIL_COUNT`.

I agree with you about the long-lived buffers; I am very much hoping to resurrect work on #1308 come mid-summer with a goal of us coming up with the adjusted object API which doesn't use this mechanism.

davidhewitt:
Agreed, I think this is a worthwhile improvement - even if immeasurable now and requires a conditional flag, with a future Rust version this will have no maintainance cost and be strictly better.

src/gil.rs
lifthrasiir: I've made `OWNED_OBJECTS` to be const-initialized as well in 291fc36d0c6a62acfceb58255b3d5da11e1dfb55.

adamreichold:
LGTM

bors r+

PR #3178, opened 2023-05-23
Fix a bad span of `_slf` for custom receivers in `#[pymethods]`.

lifthrasiir:
This turned out to be a remnant of #1506. It notably resulted in a very confusing error:

```
    --> tests\test_methods.rs:1456:9
     |
1456 |           #[pymethods]
     |           ^^^^^^^^^^^^
     |           |
     |           expected `Py<Issue1506>`, found `*mut PyObject`
     |           arguments to this function are incorrect
...
1461 | / issue_1506!(
1462 | |     #[pymethods]
1463 | |     impl Issue1506 {
1464 | |         fn issue_1506(
...    |
1536 | |     }
1537 | | );
     | |_- in this macro invocation
     |
     = note:   expected struct `pyo3::Py<Issue1506>`
             found raw pointer `*mut pyo3::ffi::PyObject`
```

The actual cause is that `SelfType::receiver` is entirely ignored in this case and `_slf` refers to the original argument, but it sounds like that `TryFrom::try_from` somehow resulted in `*mut PyObject`...

### DIFF
--- a/newsfragments/3178.fixed.md
+++ b/newsfragments/3178.fixed.md
@@ -1,0 +1,1 @@
+Fix a compile error when `#[pymethods]` items come from somewhere else (for example, as a macro argument) and a custom receiver like `Py<Self>` is used.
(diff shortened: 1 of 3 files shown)


adamreichold:
Great find! Thanks.

bors r+

PR #3181, opened 2023-05-24
Upgrade dependencies for pyo3-ffi-check

alex:


adamreichold:
@alex Thanks for doing the chores! Could you squash the commits though? (We have not configured bors to do this as we mostly want to keep the commit history.)

alex:
I originally did a bindgen bump, but it broke the MSRV build.

PR #3184, opened 2023-05-25
Additional tests for #3168

adamreichold:
These were a part of tests @lifthrasiir was preparing for https://github.com/PyO3/pyo3/issues/3165, and I believe it's worthy to add them (any single of them fails in the current main branch).


adamreichold:
I moved the two commits from #3175 here as I could not push a rebase there.

bors r+

Issue #2927, opened 2023-01-29
`py.run` has import issues on 3.8


davidhewitt:
The following code crashes on 3.8:

```rust
 test_write_unraisable() {
     #[pyfunction]
     fn report_unraisable(py: Python<'_>) {
         use pyo3::exceptions::PyRuntimeError;
         let err = PyRuntimeError::new_err("foo");
         err.write_unraisable(py, None);

         let err = PyRuntimeError::new_err("bar");
         err.write_unraisable(py, Some(py.NotImplemented().as_ref(py)));
     }

     Python::with_gil(|py| {
         let report_unraisable = wrap_pyfunction!(report_unraisable, py).unwrap();
         let locals = PyDict::new(py);
         locals
             .set_item("report_unraisable", report_unraisable)
             .unwrap();

         let source = r#"if True:
         import sys

         captured = []
         def report(data):
             captured.append(list(data))

         original_hook = sys.unraisablehook
         try:
             sys.unraisablehook = report
             report_unraisable()

             assert len(captured) == 2

             assert captured[0][0] is RuntimeError
             assert str(captured[0][1]) == 'foo'
             assert captured[0][4] is None

             assert captured[1][0] is RuntimeError
             assert str(captured[1][1]) == 'bar'
             assert captured[1][4] is NotImplemented
         finally:
             sys.unraisablehook = original_hook
         "#;

         py.run(source, Some(locals), None).unwrap();
     });
 }
```

(From #2889)

I'm not yet sure what's going on here. Probably worth investigating sometime.

davidhewitt:
No longer reproduces on `main`. Probably was fixed by #3378 (thanks @GoldsteinE!)

PR #3185, opened 2023-05-26
Fix `abi3` conversion of `__complex__` classes

### LINKED ISSUE #3164: Inconsistent custom `__complex__` conversions with `abi3`
### Bug Description

Classes that are not `complex` but define `__complex__` convert to `num-complex` types successfully when the full C API is in use, but fail with varying `TypeError`s when the limited API is in use.

### Steps to Reproduce

This is a MWE PyO3 module (strictly the return value of `noop` is irrelevant, but I left it here for clarity):
```rust
use pyo3::prelude::*;

use num_complex::Complex64;

#[pyfunction]
fn noop(a: Complex64) -> Complex64 {
   a 
}

#[pymodule]
fn core(_py: pyo3::Python<'_>, module: &PyModule) -> PyResult<()> {
    module.add_function(wrap_pyfunction!(noop, module)?)?;
    Ok(())
}
```

In Python space, if I compile against the full API, classes that define `__complex__` will successfully convert:
```python
from pyo3_test import noop

class A:
    def __complex__(self):
        return 1j

noop(A())  # returns 1j
```

If I rebuild the extension in limited-API mode, I instead get a type error:
```text
Cell In[1], line 7
      4     def __complex__(self):
      5         return 1j
----> 7 noop(A())

TypeError: argument 'a': must be real number, not A
```

### Backtrace

_No response_

macOS 13.3.1

Python 3.10.6

rustc 1.67.1 (d5a82bbd2 2023-02-07)

### Your PyO3 version

0.18.3

From source, in a venv.

### Additional Info

The relevant conversions are here: https://github.com/PyO3/pyo3/blob/3ec966d97fad85ee048da5f68891552f14c06c52/src/conversions/num_complex.rs#L139-L166.
`PyComplex_AsCComplex` uses the full `__complex__` conversion machinery in CPython, but `PyComplex_RealAsDouble` only tests if the object is already of type `complex`, and if not, assumes it's real and uses the `__float__` conversion machinery, hence the "must be a real number" errors above.

Could we modify the abi3 conversion to do the full `__complex__` conversion before the rest of its logic if `PyComplex_Check` is false?  I'd be happy to make a PR.

jakelishman:
Python classes that were not `complex` but implemented the `__complex__` magic would have that method called via `PyComplex_AsCComplex` when running against the full API, but the limited-API version `PyComplex_RealAsDouble` does not attempt this conversion.  If the input object is not already complex, we can call the magic before proceeding.

Happy to modify if I've made a mess of testing strategy - I'm not sure I've managed to do it in the same style as other tests.

Fix #3164.

### DIFF
--- a/newsfragments/3185.fixed.md
+++ b/newsfragments/3185.fixed.md
@@ -1,0 +1,1 @@
+Fix conversion of classes implementing `__complex__` to `Complex` when using `abi3` or PyPy.
--- a/src/conversions/num_complex.rs
+++ b/src/conversions/num_complex.rs
@@ -152,6 +152,18 @@ macro_rules! complex_conversion {
 
                 #[cfg(any(Py_LIMITED_API, PyPy))]
                 unsafe {
+                    let obj = if obj.is_instance_of::<PyComplex>() {
+                        obj
+                    } else if let Some(method) =
+                        obj.lookup_special(crate::intern!(obj.py(), "__complex__"))?
+                    {
+                        method.call0()?
+                    } else {
+                        // `obj` might still implement `__float__` or `__index__`, which will be
+                        // handled by `PyComplex_{Real,Imag}AsDouble`, including propagating any
+                        // errors if those methods don't exist / raise exceptions.
+                        obj
+                    };
                     let ptr = obj.as_ptr();
                     let real = ffi::PyComplex_RealAsDouble(ptr);
                     if real == -1.0 {
@@ -172,6 +184,7 @@ complex_conversion!(f64);
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::types::PyModule;
 
     #[test]
     fn from_complex() {
@@ -196,5 +209,132 @@ mod tests {
             let obj = vec![1].to_object(py);
             assert!(obj.extract::<Complex<f64>>(py).is_err());
         });
+    }
+    #[test]
+    fn from_python_magic() {
+        Python::with_gil(|py| {
+            let module = PyModule::from_code(
+                py,
+                r#"
+class A:
+    def __complex__(self): return 3.0+1.2j
+class B:
+    def __float__(self): return 3.0
+class C:
+    def __index__(self): return 3
+                "#,
+                "test.py",
+                "test",
+            )
+            .unwrap();
+            let from_complex = module.getattr("A").unwrap().call0().unwrap();
+            assert_eq!(
+                from_complex.extract::<Complex<f64>>().unwrap(),
+                Complex::new(3.0, 1.2)
+            );
+            let from_float = module.getattr("B").unwrap().call0().unwrap();
+            assert_eq!(
+                from_float.extract::<Complex<f64>>().unwrap(),
+                Complex::new(3.0, 0.0)
+            );
+            // Before Python 3.8, `__index__` wasn't tried by `float`/`complex`.
+            #[cfg(Py_3_8)]
+            {
+                let from_index = module.getattr("C").unwrap().call0().unwrap();
+                assert_eq!(
+                    from_index.extract::<Complex<f64>>().unwrap(),
+                    Complex::new(3.0, 0.0)
+                );
+            }
+        })
+    }
+    #[test]
+    fn from_python_inherited_magic() {
+        Python::with_gil(|py| {
+            let module = PyModule::from_code(
+                py,
+                r#"
+class First: pass
+class ComplexMixin:
+    def __complex__(self): return 3.0+1.2j
+class FloatMixin:
+    def __float__(self): return 3.0
+class IndexMixin:
+    def __index__(self): return 3
+class A(First, ComplexMixin): pass
+class B(First, FloatMixin): pass
+class C(First, IndexMixin): pass
+                "#,
+                "test.py",
+                "test",
+            )
+            .unwrap();
+            let from_complex = module.getattr("A").unwrap().call0().unwrap();
+            assert_eq!(
+                from_complex.extract::<Complex<f64>>().unwrap(),
+                Complex::new(3.0, 1.2)
+            );
+            let from_float = module.getattr("B").unwrap().call0().unwrap();
+            assert_eq!(
+                from_float.extract::<Complex<f64>>().unwrap(),
+                Complex::new(3.0, 0.0)
+            );
+            #[cfg(Py_3_8)]
+            {
+                let from_index = module.getattr("C").unwrap().call0().unwrap();
+                assert_eq!(
+                    from_index.extract::<Complex<f64>>().unwrap(),
+                    Complex::new(3.0, 0.0)
+                );
+            }
+        })
+    }
+    #[test]
+    fn from_python_noncallable_descriptor_magic() {
+        // Functions and lambdas implement the descriptor protocol in a way that makes
+        // `type(inst).attr(inst)` equivalent to `inst.attr()` for methods, but this isn't the only
+        // way the descriptor protocol might be implemented.
+        Python::with_gil(|py| {
+            let module = PyModule::from_code(
+                py,
+                r#"
+class A:
+    @property
+    def __complex__(self):
+        return lambda: 3.0+1.2j
+                "#,
+                "test.py",
+                "test",
+            )
+            .unwrap();
+            let obj = module.getattr("A").unwrap().call0().unwrap();
+            assert_eq!(
+                obj.extract::<Complex<f64>>().unwrap(),
+                Complex::new(3.0, 1.2)
+            );
+        })
+    }
+    #[test]
+    fn from_python_nondescriptor_magic() {
+        // Magic methods don't need to implement the descriptor protocol, if they're callable.
+        Python::with_gil(|py| {
+            let module = PyModule::from_code(
+                py,
+                r#"
+class MyComplex:
+    def __call__(self): return 3.0+1.2j
+class A:
+    __complex__ = MyComplex()
+                "#,
+                "test.py",
+                "test",
+            )
+            .unwrap();
+            let obj = module.getattr("A").unwrap().call0().unwrap();
+            assert_eq!(
+                obj.extract::<Complex<f64>>().unwrap(),
+                Complex::new(3.0, 1.2)
+            );
+        })
     }
 }
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -124,6 +124,51 @@ impl PyAny {
         }
     }
 
+    /// Retrieve an attribute value, skipping the instance dictionary during the lookup but still
+    /// binding the object to the instance.
+    ///
+    /// This is useful when trying to resolve Python's "magic" methods like `__getitem__`, which
+    /// are looked up starting from the type object.  This returns an `Option` as it is not
+    /// typically a direct error for the special lookup to fail, as magic methods are optional in
+    /// many situations in which they might be called.
+    ///
+    /// To avoid repeated temporary allocations of Python strings, the [`intern!`] macro can be used
+    /// to intern `attr_name`.
+    #[allow(dead_code)] // Currently only used with num-complex+abi3, so dead without that.
+    pub(crate) fn lookup_special<N>(&self, attr_name: N) -> PyResult<Option<&PyAny>>
+    where
+        N: IntoPy<Py<PyString>>,
+    {
+        let py = self.py();
+        let self_type = self.get_type();
+        let attr = if let Ok(attr) = self_type.getattr(attr_name) {
+            attr
+        } else {
+            return Ok(None);
+        };
+
+        // Manually resolve descriptor protocol.
+        if cfg!(Py_3_10)
+            || unsafe { ffi::PyType_HasFeature(attr.get_type_ptr(), ffi::Py_TPFLAGS_HEAPTYPE) } != 0
+        {
+            // This is the preferred faster path, but does not work on static types (generally,
+            // types defined in extension modules) before Python 3.10.
+            unsafe {
+                let descr_get_ptr = ffi::PyType_GetSlot(attr.get_type_ptr(), ffi::Py_tp_descr_get);
+                if descr_get_ptr.is_null() {
+                    return Ok(Some(attr));
+                }
+                let descr_get: ffi::descrgetfunc = std::mem::transmute(descr_get_ptr);
+                let ret = descr_get(attr.as_ptr(), self.as_ptr(), self_type.as_ptr());
+                py.from_owned_ptr_or_err(ret).map(Some)
+            }
+        } else if let Ok(descr_get) = attr.get_type().getattr(crate::intern!(py, "__get__")) {
+            descr_get.call1((attr, self, self_type)).map(Some)
+        } else {
+            Ok(Some(attr))
+        }
+    }
+
     /// Sets an attribute value.
     ///
     /// This is equivalent to the Python expression `self.attr_name = value`.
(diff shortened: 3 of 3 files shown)


src/conversions/num_complex.rs
```diff
@@ -152,6 +152,19 @@ macro_rules! complex_conversion {
 
                 #[cfg(any(Py_LIMITED_API, PyPy))]
                 unsafe {
+                    use crate::{intern, type_object::PyTypeInfo};
+
+                    let obj = if PyComplex::is_type_of(obj) {
+                        obj
+                    } else if let Ok(method) =
+                        obj.get_type().getattr(intern!(obj.py(), "__complex__"))
```
adamreichold, line 160: Why `.get_type().getattr(..)` instead of just `.getattr(..)`?

src/conversions/num_complex.rs
```diff
@@ -152,6 +152,19 @@ macro_rules! complex_conversion {
 
                 #[cfg(any(Py_LIMITED_API, PyPy))]
                 unsafe {
+                    use crate::{intern, type_object::PyTypeInfo};
+
+                    let obj = if PyComplex::is_type_of(obj) {
```
adamreichold, line 157: Purely a style question, I think `obj.is_instance_of::<PyComplex>()` would be easier to read.

src/conversions/num_complex.rs
```diff
@@ -152,6 +152,19 @@ macro_rules! complex_conversion {
...
+                        obj
+                    } else if let Ok(method) =
+                        obj.get_type().getattr(intern!(obj.py(), "__complex__"))
+                    {
+                        method.call1((obj,))?
+                    } else {
+                        // This will likely result in an error in `PyComplex_RealAsDouble`, but
```
adamreichold, line 164: I don't think this too helpful and would probably to just return the error of calling `__complex__` directly.

If I read https://docs.python.org/3/c-api/complex.html#c.PyComplex_AsCComplex correctly, we would also need to do more, e.g.

```rust
let obj = obj.is_instance_of::<PyComplex>() {
  obj
} else if let Ok(obj) = obj.method_call0(intern!(obj.py(), "__complex__")) {
  obj
} else {
  let num = obj.extract::<f64>()?;
  return Ok(Complex::from(num));
};
```

jakelishman: Oh sorry, my comment is poor here: in your case, `PyComplex_RealAsDouble` does this conversion in the same way that `PyComplex_AsCComplex` does, so the code would continue on to succeed if it's a valid float.

Do you prefer me to change the comment or to switch to your implementation?

src/conversions/num_complex.rs
jakelishman: [The spec says magic methods are accessed off the type](https://docs.python.org/3/reference/datamodel.html#special-method-names), so I was trying to be consistent with this behaviour:
```python
class A:
    def __complex__(self):
        return 1j
a = A()
a.__complex__ = lambda self: 2j
print(complex(a))  # 1j
```

In saying this, though, I realise that I'm not invoking the descriptor protocol on access to the `__complex__` attribute of the type, if required, which is why I guess the word "roughly" is in:
> `x[i]` is roughly equivalent to `type(x).__getitem__(x, i)`

Does PyO3 have a convenient interface for manually invoking the descriptor protocol on an object? CPython does it very manually in its internal `_PyObject_LookupSpecial` method, which is what it typically uses to access `__complex__`.

src/conversions/num_complex.rs
jakelishman: No problem - changed in b15573f2.

src/conversions/num_complex.rs
jakelishman: I'd thought about this, but I was concerned about the additional logic in that for parsing `str` and subclasses.  I can swap it to that if you think it's preferable, likely with an catch to error out if we're given a string?  The current implementation is pretty similar to what CPython does (except for me forgetting to invoke the descriptor protocol right now), but I get that over-tight coupling between PyO3/CPython isn't ideal.

DataTriny:
Hello @jakelishman 

Thank you!

src/conversions/num_complex.rs
adamreichold: Yeah, this too much complexity, especially if we try to second guess it by avoid the string logic.

I am also not sure whether `.get_type().getattr(..)` only lacks the descriptor protocol compared to `_PyObject_LookupSpecial` which also seems to do MRO look-up via `_PyType_Lookup`?

To be honest, I wonder if the the current implementation is good enough and a comment on this being "almost" the same protocol as the one CPython uses.

src/conversions/num_complex.rs
jakelishman: Sounds good: done in b30169ae and aa1dcb04.

src/conversions/num_complex.rs
jakelishman: I think the MRO lookup functions correctly with `obj.get_type().getattr()` - the result of `obj.get_type()` is just a regular `PyTypeObject` so the subsequent `getattr` call should resolve the MRO.  I put in an inheritance test in aa1dcb0461ac428d0dc245fe4d3ad1b5e1322028, I put in the implementation of the descriptor protocol (unfortunately the `PyTypeObject` struct is non-limited API, so no `tp_descr_get` slot for us here) in a599e82e, but if you think that's too complex, I can just remove that commit.

jakelishman:
I've pushed a couple more commits on the basis of the review.  I'm happy to revert the descriptor-resolution commit and replace it just with a comment about similarity not equality to CPython's implementation, if you'd rather.

src/conversions/num_complex.rs
adamreichold: I think the code is fine and the tests are convincing.

I would like to avoid sinking this into this `extract` implementation. We should factor out a crate-wide (but internal) helper method, something like `lookup_special` implementing the whole `.get_type().attr(..)`-and-then-again dance starting from a name like `__complex__`.

Maybe as a method on `PyAny` that we could eventually (but not in this PR) make public?

> unfortunately the PyTypeObject struct is non-limited API, so no tp_descr_get slot for us here

I think this is fine for this use case as we already are in a lower performance fallback. For the general helper method, it would be nice to add both implementations (limited and non-limited API). But I do not want to steal your time as this is tangential to the issue you are working on, so a FIXME comment would work for me as well.

jakelishman: I'm happy to do that too, no worries (but thanks for the consideration!).  Would you prefer a separate clean PR for that implementation, then I can update this PR to use it after that's merged?

I'm not in any hurry on this PR so I've got time to follow whatever procedures you prefer.  I found the `__complex__` discrepancy because of a bug in our own code (we didn't mean to use the behaviour, but were incorrectly returning a single-valued Numpy array at a point we expected a scalar).  We fixed our bug, so we could immediately move to `abi3` regardless of this PR.

adamreichold: > Would you prefer a separate clean PR for that implementation, then I can update this PR to use it after that's merged?

I think it would be fine to do here as the PR is currently very self-contained and it is always nice to merge infrastructure together with at least one usage site.

jakelishman:
A minor procedural question: this PR is getting a little messy in commit history, and with the CI failures, needs to be updated over `main`.  Do you prefer me to leave the history as-is and push an additional merge commit into this PR, or do you prefer me to rebase these commits back into a coherent story (1. add the new PyAny/ffi method, 2. fix the complex conversion)?

adamreichold:
> A minor procedural question: this PR is getting a little messy in commit history, and with the CI failures, needs to be updated over `main`. Do you prefer me to leave the history as-is and push an additional merge commit into this PR, or do you prefer me to rebase these commits back into a coherent story (1. add the new PyAny/ffi method, 2. fix the complex conversion)?

We certainly prefer a coherent story and have explicit configured to not squash merges, so turning this back into two commits would be really helpful!

jakelishman:
No trouble - I've just rebased down to two commits.

Just for visibility, there's a couple of my decisions that I think other maintainers might want to question:
- the naming/location/visibility of the new `PyAny::getattr_special`. I put it on `PyAny` because it made most sense to me, and the naming/visibility are from https://github.com/PyO3/pyo3/pull/3185#discussion_r1210574089.  Maybe since it's on `PyAny` rather than being a loose helper it might make more sense as a full `pub`?
- the return type of that method being `PyResult<Option<&PyAny>>` rather than `PyResult<&PyAny>` - it's not typically an immediate error if a magic method doesn't exist in my head, at least.

newsfragments/3185.fixed.md
```diff
@@ -0,0 +1 @@
+Fix conversion of classes implementing `__complex__` to `Complex` when using `abi3`.
```
davidhewitt, line 1:
```suggestion
Fix conversion of classes implementing `__complex__` to `Complex` when using `abi3` or PyPy.
```

davidhewitt (approved):
Thanks, the tests are great and I'm happy that the implementation is correct!

I have a couple of suggestions to round this off, I'm happy to see this merged when those are either addressed or dismissed.

davidhewitt:
As for visibility, I suggest we keep `lookup_special` private for now until users give us a reason to make this public. `_PyObject_LookupSpecial` feels like something which is for diving into the weeds and it's not immediately clear to me when we'd want to recommend it to users over `self.get_type().getattr(...)` (plus descriptor) dance.

jakelishman:
Well, I'm certainly glad that you mentioned `PyType_GetSlots` and how it might not work before Python 3.10.  It transpires that you're correct on the versioning and that it appears frequently because things like `method` are static types, so trying `PyType_GetSlots` on one of those fails to spot the descriptor.  Thanks for that advice - I'm fairly new to programming against the C API, and so I've not good great knowledge of all the available surface.

More importantly, though, it made me realise that `_PyObject_LookupSpecial` changed signature just very recently.  I thought I'd been running my tests using Python 3.9 locally, but I hadn't - I'd just been using 3.11, and CI only tests 3.11 so I hadn't noticed any issues.  Prior to 3.11, `_PyObject_LookupSpecial` is what's now called `_PyObject_LookupSpecialId`, and I don't think that's something we can bind against, because the IDs are shared static objects in the `PyRuntimeState` of the main interpreter.

What course of action would you like here?
- I could expose `_PyObject_LookupSpecial` in the FFI for Python 3.11+ only and have `PyAny::lookup_special` fallback to the getattr-based limited API form in lower CPythons, in addition to the limited API and PyPy?
- Alternatively, I could remove `_PyObject_LookupSpecial` from the FFI and use only a limited-API-safe version `PyAny::lookup_special`, since `_PyObject_LookupSpecial` might not be stable enough (and a special case for latest-Python-only in a probably non-performance-critical code path might be a bit messy).  I can switch this to a `PyType_GetSlots`-based implementation with appropriate checking of the heap-type flags for pre-3.10 CPythons to fallback to a slower getattr-based version. 

Fwiw, the complexity I'm finding here in correctly resolving non-slotted magic-method lookups across all Pythons strengthens the case for a central `PyAny::lookup_special` in my mind, for other PyO3-internal uses - there's so much subtlety I think it'd be very tricky to reproduce correctly elsewhere.

(All your other suggestions are easy and uncontentious to me.)

jakelishman:
Yeah, I think that's what I'd have suggested if someone was contributing code on a project I maintained.  I'll write this all up and push a couple of new commits to show what it looks like.  Testing against older Pythons makes it look to me like I _may_ have found a bug in `Python::run`, but I'll investigate that a little more separately - it's totally orthogonal to this PR, except that the particular way I initially wrote the tests needed tweaking.

(Sorry this has got super deep in the weeds - I wasn't expecting that!  I can say that I'm having great fun learning way more about Python/CPython internals, though.)

adamreichold:
> Testing against older Pythons makes it look to me like I may have found a bug in Python::run, but I'll investigate that a little more separately - it's totally orthogonal to this PR, except that the particular way I initially wrote the tests needed tweaking.

Please have a look at #2927 and #2891, maybe you hit one of those.

src/types/any.rs
```diff
@@ -124,6 +124,56 @@ impl PyAny {
...
+            let attr_name = attr_name.into_py(py);
+            let ret = ffi::_PyObject_LookupSpecial(self.as_ptr(), attr_name.as_ptr());
+            if ret.is_null() {
+                // _PyObject_LookupSpecial doesn't set an exception on lookup failure, but should
+                // if there's an exception during evaluation of the attribute (such as during the
+                // descriptor resolution).
+                match PyErr::take(py) {
+                    Some(err) => Err(err),
+                    None => Ok(None),
+                }
```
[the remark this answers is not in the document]
jakelishman: Ah thanks - this _felt_ like something that'd have a chained method, but I couldn't spot the correct one on `Option`.  Done in rebased befabb2e

adamreichold:
Let's give it a whirl in the CI then.

bors try

src/types/any.rs
```diff
@@ -124,6 +124,55 @@ impl PyAny {
...
+                if descr_get_ptr.is_null() {
+                    return Ok(Some(attr));
+                }
+                let descr_get: ffi::descrgetfunc = std::mem::transmute(descr_get_ptr);
+                let ret = descr_get(attr.as_ptr(), self.as_ptr(), self_type.as_ptr());
+                if ret.is_null() {
+                    Err(PyErr::fetch(py))
+                } else {
+                    Ok(Some(py.from_owned_ptr(ret)))
+                }
```
davidhewitt, line 167: I think this could be

```suggestion
        py.from_owned_ptr_or_err(ret).map(Some)
```

src/types/any.rs
```diff
@@ -124,6 +124,55 @@ impl PyAny {
...
+        let attr = if let Ok(attr) = self_type.getattr(attr_name) {
+            attr
+        } else {
+            return Ok(None);
+        };
+
+        // Manually resolve descriptor protocol.
+        unsafe {
+            if cfg!(Py_3_10)
+                || ffi::PyType_HasFeature(attr.get_type_ptr(), ffi::Py_TPFLAGS_HEAPTYPE) != 0
```
davidhewitt, line 153: Given the `unsafe` isn't used in the slow paths, it might be nice to move it inside the fast path.

```suggestion
        if cfg!(Py_3_10)
            || unsafe { ffi::PyType_HasFeature(attr.get_type_ptr(), ffi::Py_TPFLAGS_HEAPTYPE) } != 0
        {
            unsafe { 
               ...
            }
```

src/types/any.rs
```diff
@@ -124,6 +124,55 @@ impl PyAny {
...
+    where
+        N: IntoPy<Py<PyString>>,
+    {
+        let py = self.py();
+        let self_type = self.get_type();
+        let attr = if let Ok(attr) = self_type.getattr(attr_name) {
+            attr
+        } else {
+            return Ok(None);
+        };
```
davidhewitt, line 148: let-else is going to be so nice once PyO3 gets to MSRV of 1.65... but that's going to be a long time yet! (We're just about to do a bump to 1.56 I think.)

davidhewitt (approved):
Nice one! I have just a couple of tiny code style suggestions for the new `lookup_special` body, otherwise this looks perfect.

`cargo test` on PyPy is known to not work, it's a deficiency on PyPy's side which I hope to resolve one day. We have a `pytests` crate which builds an extension module and runs a small set of tests, which PyPy does support.

src/types/any.rs
adamreichold: How did this get through the full CI though? Are we missing a feature configuration for the MSRV job?

adamreichold: Yes, right. In such cases, personally, I actually prefer `match` as it gives "tigher" formatting even though it does "stutter" a bit:

```rust
let attr = match self_type.getattr(attr_name) {
  Ok(attr) => attr,
  Err(_err) => return Ok(None),
};
```

src/types/any.rs
jakelishman: Sounds good - fixed in 194d0c79.

jakelishman:
There we go - style changes made.

> [about pypy]

Ah yeah, that'd explain it - I'll have a look at doing that properly next time round, thanks!

adamreichold:
I think this is good to go now having collected two approvals with all threads resolved.

bors r+

jakelishman:
Thanks for all the help, chaps!

PR #3187, opened 2023-05-28
release: 0.19.0

davidhewitt:
As per #3154 I think we're ready to push 0.19.

I'll put this live on Tuesday evening unless I hear a reason to postpone.

PR #3188, opened 2023-05-29
Verify that macros do work without any imports

lifthrasiir:
Currently virtually all (positive) tests import `pyo3::prelude::*`, making it hard to detect a certain class of bugs. This PR adds an explicit test that never imports from `pyo3` to fix this.

Also this fixes a minor bug from #3157 which didn't work without `use pyo3::types::PyType;`. I think this should be a part of 0.19.0 (#3187), so no additional changelog would be required (as this feature is new in 0.19.0).


adamreichold:
Great find with a principled solution! I agree that this should be part of 0.19.0.

bors r+

davidhewitt:
Agreed, thanks!

Issue #2089, opened 2022-01-06
Support `__eq__` for pyclass


Lunarmagpie:
I don't think is possible to create a class like this in pyo3. I couldn't find a way to write an equality method in the docs.
```python
class SomeClass:
    def __eq__(self, other):
        return True
```
I need to be able to write an implementation for this so I can make a python wrapper around rust enums. I found that there wasn't a way to do an equality check. I'm considered writing an `is_equal_to` method then making another class in python that knows how to use `is_equal_to` when you call `__eq__`. I want to avoid this because I think its clunky and requires a lot of boilerplate. Forcing an end user to use a method like `is_equal_to` is jarring so I can't do that.

davidhewitt:
At the moment you should implement `__richcmp__`, which is the C api where all these operators are defined. See https://github.com/PyO3/pyo3/issues/1625#issuecomment-886074580 for a current example, unfortunately the docs aren't great for it right at this moment.

Let's leave this issue open, because I think with the new `#[pymethods]` magic methods there might be a way to enable users to write the usual Python operations like `__eq__`. I just need to find a chance to have a play with enabling this.

Lunarmagpie:
Thanks for explaining!

davidhewitt:
@ssokolow there's been suggestions previously to have `#[pyclass(dataclass)]` to do exactly that, see https://github.com/PyO3/pyo3/issues/1375#issuecomment-766110548

Just not got around to implementing!

ssokolow:
I don't necessarily want dataclass support... just the ability to proxy implementations through easily.

(My most recent example *would* have been fine with `#[pyclass(dataclass)]` since it's just porting an `IntEnum` to Rust as part of migrating as much as possible of a PyQt codebase into its Rust backend, but I've had cases where I would want to easily expose derives as the corresponding dunder methods without having to make all fields public or refactor the thing to put the private methods in a field containing an opaque object.)

davidhewitt:
Ah, interesting. I will have a think about whether there's a nice way to do this - it might solve some papercuts I've had too!

alex:
Being able to define `__eq__`, instead of `__richcmp__`, would be a big QoL improvement. The specific advantages are: Not needing to open-code a not-equal or order-comparison implementations, since those could just offer standard behaviors.

davidhewitt:
Finally pushed #3203 to implement this!

PR #3195, opened 2023-05-31
RFC: automatically implement equality and ordering using Eq and Ord

davidhewitt:
I had a sudden insight today that we can automatically generate implementations of equality and ordering based on the presence of `Eq` and `Ord` implementations for `#[pyclass]`. (Note that `Ord: Eq`.) The implementation selects for these traits using autoderef specialization.

I think this could be very convenient, but it has the potential to be breaking because existing types will suddenly have new behaviour in Python. If we went with this automatic implementation, we could land it in 0.20 with appropriate documentation.

I think at the very least we should have `#[pyclass(eq = false)]` and `#[pyclass(order = false)]` opt-outs (names matching the same functionality in Python `@dataclass`, although note the defaults here are not `True` and `False` respectively but instead based on whether `T` implements `Eq` or `Ord`).

Design discussions:
- Alternative would be to always default to `false`, and require `eq = true` (and `order = true`) to opt-in to adding these implementations. This would be non-breaking but I think is not as nice for future usage.
  - Maybe we can use this as an opt-in over a deprecation period?
- What to do with interactions around subclasses and extensions? I am certain the default implementations I have currently written do the wrong thing for `#[pyclass(extends = ...)]` and also if `T` is subclassed.
- Also there may be an interaction between this and the implementation for #2089. Maybe it's better to land that first?

### DIFF
--- a/src/impl_/pyclass.rs
+++ b/src/impl_/pyclass.rs
@@ -1,4 +1,5 @@
 use crate::{
+    basic::CompareOp,
     exceptions::{PyAttributeError, PyNotImplementedError, PyRuntimeError, PyValueError},
     ffi,
     impl_::freelist::FreeList,
@@ -7,12 +8,15 @@ use crate::{
     pycell::PyCellLayout,
     pyclass_init::PyObjectInit,
     type_object::PyLayout,
-    Py, PyAny, PyCell, PyClass, PyErr, PyMethodDefType, PyNativeType, PyResult, PyTypeInfo, Python,
+    types::PyBool,
+    IntoPyPointer, Py, PyAny, PyCell, PyClass, PyErr, PyMethodDefType, PyNativeType, PyRef,
+    PyResult, PyTypeInfo, Python,
 };
 use std::{
     borrow::Cow,
     ffi::{CStr, CString},
     marker::PhantomData,
+    ops::Deref,
     os::raw::{c_int, c_void},
     ptr::NonNull,
     thread,
@@ -109,6 +113,7 @@ impl PyClassWeakRef for PyClassWeakRefSlot {
 pub struct PyClassImplCollector<T>(PhantomData<T>);
 
 impl<T> PyClassImplCollector<T> {
+    #[inline]
     pub fn new() -> Self {
         Self(PhantomData)
     }
@@ -200,6 +205,8 @@ pub trait PyClassImpl: Sized + 'static {
     }
 
     fn lazy_type_object() -> &'static LazyTypeObject<Self>;
+
+    fn default_tp_richcompare() -> Option<ffi::richcmpfunc>;
 }
 
 /// Runtime helper to build a class docstring from the `doc` and `text_signature`.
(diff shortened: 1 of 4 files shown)


davidhewitt:
If we like this, I think we can do the same thing with `Hash` (only for `frozen` classes by default). Potentially also `Display` and `Debug` could be used for `str` and `repr` (though maybe that's less desirable).

adamreichold:
> Also there may be an interaction between this and the implementation for https://github.com/PyO3/pyo3/issues/2089. Maybe it's better to land that first?

I think we could land this first if it is easier to implement, but we should make sure that we do not block manually implementing `__eq__` before committing to an API.

Reading [a comment](https://github.com/PyO3/pyo3/issues/2089#issuecomment-1130095387] from there and

> Alternative would be to always default to false, and require eq = true (and order = true) to opt-in to adding these implementations. This would be non-breaking but I think is not as nice for future usage. 

I would actually prefer if this worked like `#[derive(SomeTrait)]`, i.e. it is purely opt-in using something like `#[pyclass(eq = true)]` to follow the principle of least surprise. (I also feel that this is functionality would be more Rusty than Pythonic being based on trait implementations instead of implementing magic methods and hence think it would be sensible to align this with the Rust-side precedence.)

> If we like this, I think we can do the same thing with Hash (only for frozen classes by default). Potentially also Display and Debug could be used for str and repr (though maybe that's less desirable).

As a `#[derive(..)]`-like opt-in mechanism, I would support making the support as broad as possible.

I am not sure though that the concerns w.r.t. `frozen` as specific to `Hash` though: Would ordering-based containers have exactly the same issue with `Ord` and pervasive resp. interior mutability? Or this is mainly because `dict` is so common on the Python side? Additionally, with opt-in, I don't additional interaction with `frozen` would be necessary?

adamreichold:
As my role as local resident autoref specialization sceptic also seems to be a given, doing this opt-in would probably make that unnecessary as well? I would see that as a benefit because while autoref specialization is an ingenious trick, it is also a somewhat indirect/implicit method of realizing specialization that does make the code harder to maintain IMHO.

davidhewitt:
> As my role as local resident autoref specialization sceptic also seems to be a given, I'd also like to add that doing this opt-in would probably make that unnecessary as well? I would see that as a benefit because while autoref specialization is an ingenious trick, it is also a somewhat indirect/implicit method of realizing specialization that does make the code harder to maintain IMHO.

I think I see both sides of the coin. I agree that opt-in feels more Rusty and avoids the compile overhead and maintenance complexity of the specialization.

On the other hand, I think there's quite a strong argument that the user has already opted in by deriving or implementing these traits for a `#[pyclass]`, and more likely than not it's a bug if they then forget to also add `#[pyclass(eq = true)]`.

I think for me it may hinge what the solution for inheritance looks like. If we can only safely implement these methods in absence of insurance then the rules get very complicated for automatic implementation. If the inheritance cases can be cleanly handled by our automatic implementation then I think the bet that doing this automatically helps the user avoid the bug of forgetting to opt in is a strong motivator to me.

> I am not sure though that the concerns w.r.t. frozen as specific to Hash though: Would ordering-based containers have exactly the same issue with Ord and pervasive resp. interior mutability? 

Great question. I suppose ordering-based containers in Rust have the same problem with interior mutability? So perhaps automatic `Ord` is also only appropriate for frozen classes.

src/impl_/pyclass.rs
```diff
@@ -1027,3 +1092,57 @@ pub(crate) unsafe extern "C" fn assign_sequence_item_from_mapping(
...
+    let slf: PyRef<'_, T> = slf.try_borrow()?;
+    // FIXME: what to do about subtypes?
+    let other: &PyAny = unsafe { py.from_borrowed_ptr(other) };
+    let other: PyRef<'_, T> = other.extract()?;
+
+    match CompareOp::from_raw(op) {
+        Some(CompareOp::Eq) => Ok(PyBool::new(py, slf.eq(&other)).into_ptr()),
+        Some(CompareOp::Ne) => Ok(PyBool::new(py, slf.ne(&other)).into_ptr()),
+        Some(_) => Ok(py.NotImplemented().into_ptr()),
+        None => unreachable!("invalid compareop"),
```
adamreichold, line 1120: Maybe just `CompareOp::from_raw(op).expect("invalid compareop")`?

davidhewitt:
This is an interesting question. Overall I have had the philosophy that `#[pyclass]` should match the Python `class` as much as possible, because this is what users expect. Where there are differences I think we should take care to document them.

I am mostly in favour of not making `__richcmp__` directly implementable; I think the only argument to that would be whether any "split-out" implementation from #2089 has any performance difference to a direct `__richcmp__`. We will have to see about that in the future.

On reflection about whether this automatic implementation should be merged - I think PyO3 may not be ready for this yet, though I would like us to still keep open about it in the future. I think merging the opt-in forms `eq = true`, `hash = true` etc would be a nice end-user feature which could be deprecated in the future if we wanted to make automatic implementations later.

I think the clear next step is to complete #2089, as that's a clearly desirable addition and we'll then understand what implementation constraints we'd be working with for all of the possibilities discussed here. I'll try to get to that sometime in the near-ish future.

PR #3197, opened 2023-06-01
Add `IpAddr` <-> `ipaddress.IPv(4/6)Address` conversion

mhils:
This PR adds support for conversion between Python's `ipaddress.IPv4Address`/`ipaddress.IPv6Address` and `std::net::IpAddr`. :smiley:

### DIFF
--- a/src/conversions/std/ipaddr.rs
+++ b/src/conversions/std/ipaddr.rs
@@ -1,0 +1,110 @@
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+use crate::exceptions::PyValueError;
+use crate::sync::GILOnceCell;
+use crate::types::PyType;
+use crate::{intern, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, Python, ToPyObject};
+
+impl FromPyObject<'_> for IpAddr {
+    fn extract(obj: &PyAny) -> PyResult<Self> {
+        match obj.getattr(intern!(obj.py(), "packed")) {
+            Ok(packed) => {
+                if let Ok(packed) = packed.extract::<[u8; 4]>() {
+                    Ok(IpAddr::V4(Ipv4Addr::from(packed)))
+                } else if let Ok(packed) = packed.extract::<[u8; 16]>() {
+                    Ok(IpAddr::V6(Ipv6Addr::from(packed)))
+                } else {
+                    Err(PyValueError::new_err("invalid packed length"))
+                }
+            }
+            Err(_) => {
+                // We don't have a .packed attribute, so we try to construct an IP from str().
+                obj.str()?.to_str()?.parse().map_err(PyValueError::new_err)
+            }
+        }
+    }
+}
+
+impl ToPyObject for Ipv4Addr {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        static IPV4_ADDRESS: GILOnceCell<Py<PyType>> = GILOnceCell::new();
+        IPV4_ADDRESS
+            .get_or_try_init_type_ref(py, "ipaddress", "IPv4Address")
+            .expect("failed to load ipaddress.IPv4Address")
+            .call1((u32::from_be_bytes(self.octets()),))
+            .expect("failed to construct ipaddress.IPv4Address")
+            .to_object(py)
+    }
+}
+
+impl ToPyObject for Ipv6Addr {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        static IPV6_ADDRESS: GILOnceCell<Py<PyType>> = GILOnceCell::new();
+        IPV6_ADDRESS
+            .get_or_try_init_type_ref(py, "ipaddress", "IPv6Address")
+            .expect("failed to load ipaddress.IPv6Address")
+            .call1((u128::from_be_bytes(self.octets()),))
+            .expect("failed to construct ipaddress.IPv6Address")
+            .to_object(py)
+    }
+}
+
+impl ToPyObject for IpAddr {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        match self {
+            IpAddr::V4(ip) => ip.to_object(py),
+            IpAddr::V6(ip) => ip.to_object(py),
+        }
+    }
+}
+
+impl IntoPy<PyObject> for IpAddr {
+    fn into_py(self, py: Python<'_>) -> PyObject {
+        self.to_object(py)
+    }
+}
+
+#[cfg(test)]
+mod test_ipaddr {
+    use std::str::FromStr;
+
+    use crate::types::PyString;
+
+    use super::*;
+
+    #[test]
+    fn test_roundtrip() {
+        Python::with_gil(|py| {
+            fn roundtrip(py: Python<'_>, ip: &str) {
+                let ip = IpAddr::from_str(ip).unwrap();
+                let py_cls = if ip.is_ipv4() {
+                    "IPv4Address"
+                } else {
+                    "IPv6Address"
+                };
+
+                let pyobj = ip.into_py(py);
+                let repr = pyobj.as_ref(py).repr().unwrap().to_string_lossy();
+                assert_eq!(repr, format!("{}('{}')", py_cls, ip));
+
+                let ip2: IpAddr = pyobj.extract(py).unwrap();
+                assert_eq!(ip, ip2);
+            }
+            roundtrip(py, "127.0.0.1");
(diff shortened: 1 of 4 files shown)


src/conversions/std/ipaddr.rs
```diff
@@ -0,0 +1,123 @@
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+use crate::exceptions::PyValueError;
+use crate::once_cell::GILOnceCell;
+use crate::types::PyType;
+use crate::{intern, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, Python, ToPyObject};
+
+/// A GILOnceCell that holds a reference to a Python class.
+struct PyClassRef(GILOnceCell<Py<PyType>>, &'static str, &'static str);
```
mhils, line 9: There are a few occasions where this particular GILOnceCell pattern is used (e.g. [`types/mapping.rs`](https://github.com/PyO3/pyo3/blob/451729aef0e787a85011dc608433dea9c663ecee/src/types/mapping.rs#L118), [`conversions/rust_decimal.rs`](https://github.com/PyO3/pyo3/blob/451729aef0e787a85011dc608433dea9c663ecee/src/conversions/rust_decimal.rs#L76)). Happy to move PyClassRef outside of this module and use it consistently around the codebase if you feel that'd be an improvement. :)

mhils: Oh, that's neat! Implemented in https://github.com/PyO3/pyo3/pull/3197/commits/721c13ce30782415606c5a06a934378eace24415. If y'all feel that's generally useful I'd be happy to apply it outside of the module. :)

src/conversions/std/ipaddr.rs
```diff
@@ -0,0 +1,123 @@
...
+        py: Python<'py>,
+        module_name: &'static str,
+        attr_name: &'static str,
+    ) -> PyResult<&'py PyType> {
+        self.get_or_try_init(py, || py.import(module_name)?.getattr(attr_name)?.extract())
+            .map(|ty| ty.as_ref(py))
+    }
+}
+
+static IPV4_ADDRESS: GILOnceCell<Py<PyType>> = GILOnceCell::new();
```
adamreichold, line 20: I think those statics could have their scope reduced by moving them into the `to_object` bodies.

src/conversions/std/ipaddr.rs
mhils: Probably easiest to centralize that in a follow-up PR, any objections to that? :)

mhils:
> (Please note that you'll either need to squash or rebase into a coherent history at the end before meging.)

To clarify, does this mean that your automation does not squash automatically an I am expected squash myself manually? Happy to do that, I didn't find anything specific in `Contributing.md` and expected you'd prefer to just press GitHub's squash merge button in the end. :)

DataTriny:
Hello @mhils 

Thank you!

adamreichold:
> > (Please note that you'll either need to squash or rebase into a coherent history at the end before meging.)
> 
> To clarify, does this mean that your automation does not squash automatically an I am expected squash myself manually? Happy to do that, I didn't find anything specific in `Contributing.md` and expected you'd prefer to just press GitHub's squash merge button in the end. :)

Our automation is currently bors and it can only always merge or always squash. Since we do like coherent multi-commit histories, we do not squash and hence you need to either squash manually or provide a coherent multi-commit history.

(This is not currently documented, but it is probably not useful ATM as bors-ng will not available for long and we will need to switch to GitHub's merge queues (which seem to have the same issue) soon.)

src/conversions/std/ipaddr.rs
adamreichold: Please still move the method definition to `src/sync.rs` even it is only used here (the visibility is not affected by this in any case).

mhils: Done, thanks! 😃

adamreichold (approved):
The `extract` implementation is somewhat broad, but I think this is alright since the conversion are guided by the given types on the Rust side of things.

adamreichold:
CI fails due to any unused import in the `sync` module.

mhils:
> CI fails due to any unused import in the `sync` module.

Fixed, thanks! 🙇

adamreichold:
@davidhewitt I would want to merge this but would be glad for a second opinion on the broadness of the `extract` implementation.

davidhewitt (approved):
Yep, this looks great to me - thanks both and sorry for the delay on my part!

src/conversions/std/ipaddr.rs
```diff
@@ -0,0 +1,110 @@
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+use crate::exceptions::PyValueError;
+use crate::sync::GILOnceCell;
+use crate::types::PyType;
+use crate::{intern, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, Python, ToPyObject};
+
+impl FromPyObject<'_> for IpAddr {
+    fn extract(obj: &PyAny) -> PyResult<Self> {
+        match obj.getattr(intern!(obj.py(), "packed")) {
```
davidhewitt, line 10: Actually, I suppose there is a good question about the width of the conversion here. I feel that normally we do some kind of type check on the way in, so this is a break from most of PyO3's convention?

davidhewitt: Is the motivation not to check due to performance?

src/conversions/std/ipaddr.rs
```diff
@@ -0,0 +1,110 @@
...
+                    Ok(IpAddr::V4(Ipv4Addr::from(packed)))
+                } else if let Ok(packed) = packed.extract::<[u8; 16]>() {
+                    Ok(IpAddr::V6(Ipv6Addr::from(packed)))
+                } else {
+                    Err(PyValueError::new_err("invalid packed length"))
+                }
+            }
+            Err(_) => {
+                // We don't have a .packed attribute, so we try to construct an IP from str().
+                obj.str()?.to_str()?.parse().map_err(PyValueError::new_err)
```
davidhewitt, line 22: Or is what's going on here that we're actually allowing strings as input? So this it's arguably convenient but this obviously doesn't round trip. Is that a problem?

src/conversions/std/ipaddr.rs
davidhewitt: So not type-checking for exactly an ip address on the to-rust direction is consistent with what we've done for the recent `rust_decimal` integration, so I think this is fine?

adamreichold: Yes, I think so too. But it was a point of contention in that review (of the `rust_decimal` integration) so I was uncomfortable just merging it without further discussion. As written before, I am personally completely fine with board `extract` impls (even parsing strings) as the user's Rust code controls the target types and the order in which they are applied.

mhils: FWIW the entire `.packed` check is a performance optimization. Python's IP address classes internally store the address as an int. The fallback path (`obj.str()?.to_str()?.parse()`) would also work for `IPv4Address` and `IPv6Address` instances. I have no strong opinions here and would be happy to go with whatever you feel is most consistent. :)

From an API user perspective, the ability to pass in strings would be very nice as that matches existing Python APIs (`socket.socket` etc take strings, not IPvXAddress instances).

davidhewitt: Yep, I think what we're converging on is that these generous-on-input conversions are quite practical.

At some point I wonder if we should go back over all the conversions we've implemented, categorise them into whether they are "generous" or strict, and make sure we're consistent in that approach?

davidhewitt:
I think we're happy to see this move forward, so I'll click merge now. Thanks again!

src/conversions/std/ipaddr.rs
adamreichold: > At some point I wonder if we should go back over all the conversions we've implemented, categorise them into whether they are "generous" or strict, and make sure we're consistent in that approach?

It does sound like a good idea to check them all for consistency. I just wonder when someone will the time to do that since there quite a few of them by now. Certainly worth making an issue to be resolve before 1.0.

davidhewitt: Created #3226

mhils:
Thank you two again! 🍰

It seems like the merge queue has hit an unrelated race condition, happy to fix things here if I'm misreading it. :)

davidhewitt:
Yep no worries I will get #3225 merged and then retry.

Issue #1813, opened 2021-08-18
Fallible alternative to IntoPy?


kszucs:
While `FromPyObject::extract()` returns with a `PyResult` the `IntoPy` conversion trait returns with a `PyObject` making it hard to implement conversions which may return with an error. 

Tried to manually raise an exception, but unable to propagate the correct error.
```rust
impl<'a> IntoPy<PyObject> for CustomType {
    fn into_py(self, py: Python) -> PyObject {
        match self.to_python(py) {
            Ok(obj) => obj,
            Err(err) => {
                err.restore(py);
                py.None() // SystemError: <built-in function to_python_raises> returned a result with an error set
                //panic!("error occured") => pyo3_runtime.PanicException: eeeeeeee
            }
        }
    }
}
```

Is there a workaround or plans to add support for a fallible IntoPy alternative?

A bit unrelated, but I'd like to mention that the current naming is a bit confusing, something like the following would better align with the rust conversion traits.

```
FromPy[Object]
IntoPy[Object]
TryFromPy[Object]
TryIntoPy[Object]
```

davidhewitt:
Yes, I absolutely agree that something should be done here.

I think the overall question is just about defining a better iteration of PyO3's conversion traits. We already have `ToPyObject` and `IntoPy` for the to-Python direction, and I think we should aim to start with a plan of what these conversions should look like in the future.

elliottgorrell:
Hey @davidhewitt is there any update on this? It feels like a pretty large chink in the armour. Because if IntoPy panics python ends up with a PanicException which is more than likely going to bubble up uncaught and crash the python process as it is based off BaseException.

I could potentially help but curious if there is a fundamental reason this isn't fallible or hasn't been prioritised?

davidhewitt:
No update, simple answer is that I haven't had time to invest into this particular issue.

The challenges I see:
- How many new traits do we need? Fallible `IntoPy` or `ToPyObject` could both be useful.
- What is the migration path? E.g. `IntoPy for HashMap` probably needs to be fallible, but it might break a lot of code to replace it.
- Do we make this easier by having blanket implementation of fallible conversion based on non-fallible conversion? Probably, but then it's not possible to retain the panicking fallible implementation for backwards compatibility.

[...]

PR #3198, opened 2023-06-02
Add abi3 + num_bigint conversion

### LINKED ISSUE #3196: features = ["abi3", "num-bigint"] silently ignore "num-bigint"
### Bug Description

It happens because num_bigint.rs has `#![cfg(all(feature = "num-bigint", not(any(Py_LIMITED_API))))]` on the top, right?

If the features are actually not compatible, could there be any error or warning?

### Steps to Reproduce

1. Add `num-bigint` features and use the conversion features.
2. Add `abi3`
3. Then conversion code build fails.

### Backtrace

_No response_

MacOS / 13.3.1

Python 3.11.1

### Your PyO3 version

0.19.0

pyenv install 3.11.1 && pyenv virtualenv 3.11.1 pyo3 && pyenv local pyo3

### Additional Info

Another suggestion is implementing num-bigint for Py_LIMITED_API. Since it is already using `ffi::_PyLong_FromByteArray`, using a python call through `int.from_bytes` and `int.to_bytes` will makes sense (with additional overhead).
I am sorry if this is a silly idea. Is this expected to be implemented on user side?

youknowone:
Fix #3196


src/conversions/num_bigint.rs
```diff
@@ -93,6 +113,23 @@ macro_rules! bigint_conversion {
...
+                let kwargs = if $is_signed > 0 {
+                    let kwargs = PyDict::new(py);
+                    kwargs.set_item(intern!(py, "signed"), true).unwrap();
+                    Some(kwargs)
+                } else {
+                    None
+                };
+                py.get_type::<PyLong>()
+                    .call_method("from_bytes", (bytes_obj, "little"), kwargs)
+                    .expect("TODO")
```
youknowone, line 130: How should I handle errors?

src/conversions/num_bigint.rs
```diff
@@ -128,6 +166,32 @@ macro_rules! bigint_conversion {
                     }
                 }
             }
+
+            #[cfg(Py_LIMITED_API)]
```
youknowone, line 170: where will be the proper places to put the limited api implementation? Like now or a new file or module?

youknowone (commented):
@davidhewitt Could you give advices? I left questions.

src/conversions/num_bigint.rs
```diff
@@ -2,7 +2,7 @@
 //
 // based on Daniel Grunwald's https://github.com/dgrunwald/rust-cpython
 
-#![cfg(all(feature = "num-bigint", not(any(Py_LIMITED_API))))]
+#![cfg(all(feature = "num-bigint"))]
```
davidhewitt, line 5:
```suggestion
#![cfg(feature = "num-bigint")]
```

src/conversions/num_bigint.rs
davidhewitt: For now `.expect()` is ok , we will eventually improve this in #1813, it's helpful if you could add a `// FIXME #1813 or similar` here.

src/conversions/num_bigint.rs
davidhewitt: It looks like the only thing which changes here is the implementation of how to get `n_bytes`, so I think we can do something like:

```rust
let n_bytes = cfg_if::cfg_if! {
    if #[cfg(not(Py_LIMITED_API))] {
        // fast path
    } else { 
        // slow path
    }
}
```

DataTriny:
Hello @youknowone 

Thank you!

youknowone:
@DataTriny Thank you. https://github.com/PyO3/pyo3/pull/3108#issuecomment-1574004992

davidhewitt (approved):
Looks good to me, thanks for implementing this!

PR #3199, opened 2023-06-02
update PR template to detail state of licensing

davidhewitt:
Adds detail about #3108 to the README and the PR template.

@DataTriny - I thought that adding this would help encourage new contributors to be aware of the licensing terms and also that they would need to explicitly accept the future license.


DataTriny (approved):
Looks good! Should have been my first step actually.

PR #3200, opened 2023-06-02
RFC: remove copyright headers from source files

davidhewitt:
As part of our relicensing effort - I've observed that over time we haven't applied COPYRIGHT headers consistently to source files.

I've seen different organisations take different approaches to this. I note that for [`rust-lang/rust`, headers only exist where code has been moved inside the Rust repo without relicensing](https://github.com/search?q=repo%3Arust-lang%2Frust+copyright+language%3ARust&type=code&l=Rust).

For PyO3, I suggest we follow this and remove COPYRIGHT headers from source. We may wish to consider attributing to `rust-cpython` as the original project from which PyO3 was forked, possibly in the License section of the readme?


davidhewitt:
Based on the thumbs-up, it sounds like folks are generally in agreement with removing these lines. I'll proceed with merge, we can always revert if we change our minds.

bors r+

Issue #3190, opened 2023-05-29
`__anext__` should be able to return `&PyAny`


lifthrasiir:
Currently `__anext__` expects `IterANextOutput<_, _>`, which can be only converted from `Option<T1>` or `IterANextOutput<T2, T3>` where all `T`s implement `IntoPy<PyObject>`. But `__anext__` frequently returns an awaitable, and whether the result is going to be `Yield` or `Return` is up to the awaitable and not `__anext__` itself.

Searching for the existing `__anext__` implementations from GitHub confirms this issue, and everyone seems to come up with their own workaround (I also had one):

* https://github.com/Gelbpunkt/zangy/blob/a00fb19a62d2f2446c73d0c66999983c22b58627/src/pool.rs#L1373
* https://github.com/apache/incubator-opendal/blob/2d4c14437328bdf2bb96de8395f833777e777a55/bindings/python/src/asyncio.rs#L351
* https://github.com/hydro-project/hydroflow/blob/28f7c69cd73f0556b5ac2688105417d4ce0e8d08/hydro_cli/src/lib.rs#L414-L422
* https://github.com/awslabs/smithy-rs/blob/79ead4884c206f1c41eacb5e61feab2a96c06173/rust-runtime/aws-smithy-http-server-python/src/types.rs#L489

Therefore ideally we should allow any `T` (that implements `IntoPy<PyObject>` as above), but that impl will conflict with both existing impls. While the impl for `IterANextOutput<_, _>` can be merged into `IntoPyCallbackOutput` and so is redundant, the impl for `Option<T>` cannot be removed because (Rust) `None` should be converted to `PyStopAsyncIteration`, not (Python) `None`. So the only possibility would be an [autoref specialization](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md).

On the other hand the value of forcing `IterANextOutput<_, _>` from `__anext__` seems unclear. Assuming the last `__anext__` did return some awaitable and it yielded some value, the awaitable might or might not know that the iteration has stopped---both possibilities are plausible. The current design is only suitable if the stop is already known. Probably `__anext__` should not restrict return types and we should just make `IterANextOutput` implement `IntoPy`, ditching the `Option<T>` case (with a proper documentation of course).

adamreichold:
[...]

PR #3202, opened 2023-06-02
RFC: add error handling to IterOutput types

joshua-auchincloss:
Hello, hoping for comments on the following changes: 
- Addition for IterANextOutput::Err
- Addition for IterNextOutput::Err

In short, this is rationalized by the current error handling by PyO3 for custom iterators leading to unexpected behaviours at runtime. For example: 

```rust
use pyo3::prelude::*;
use pyo3::iter::IterNextOutput;

#[pyclass]
struct PyClassIter {
    count: usize,
}

#[pymethods]
impl PyClassIter {
    #[new]
    pub fn new() -> Self {
        PyClassIter { count: 0 }
    }

    fn __next__(&mut self, py: Python<'_>) -> IterNextOutput<usize, PyObject> {
        if self.count < 5 {
            self.count += 1;
            // Given an instance `counter`, First five `next(counter)` calls yield 1, 2, 3, 4, 5.
            return IterNextOutput::Yield(self.count)
        } else if self.count == 6 {
            return IterNextOutput::Return("the stuff".into_py(py))
       };
        IterNextOutput::Return(pyo3::exceptions::PyRuntimeError::new_err("oh no this shouldn't have happened").into_py(py))
    }
}
```
In this example, we are simply returning the value as `StopIteration(value=value)` - if the user attempts to return a PyError, pyo3 will serialize the <i>actual `PyRuntimeError` exception object</i>.

This is fixed with the `::Err` enum field, where we can:

```rust
use pyo3::prelude::*;
use pyo3::iter::IterNextOutput;

#[pyclass]
struct PyClassIter {
    count: usize,
}

#[pymethods]
impl PyClassIter {
    #[new]
    pub fn new() -> Self {
        PyClassIter { count: 0 }
    }

    fn __next__(&mut self) -> IterNextOutput<usize, &'static str, &'static str> {
        if self.count < 5 {
            self.count += 1;
            // Given an instance `counter`, First five `next(counter)` calls yield 1, 2, 3, 4, 5.
            return IterNextOutput::Yield(self.count)
        } else if self.count > 6 {
            // Runtime error
            return IterNextOutput::Err("exceeded depth")
        };
        IterNextOutput::Return("value")
    }
}
```

This output allows for an actual `RuntimeError` to be raised should `count` > 6 (trivial example, but actual use-cases require handling those race conditions)

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -774,12 +774,12 @@ const __GET__: SlotDef = SlotDef::new("Py_tp_descr_get", "descrgetfunc")
     .arguments(&[Ty::MaybeNullObject, Ty::MaybeNullObject]);
 const __ITER__: SlotDef = SlotDef::new("Py_tp_iter", "getiterfunc");
 const __NEXT__: SlotDef = SlotDef::new("Py_tp_iternext", "iternextfunc").return_conversion(
-    TokenGenerator(|| quote! { _pyo3::class::iter::IterNextOutput::<_, _> }),
+    TokenGenerator(|| quote! { _pyo3::class::iter::IterNextOutput::<_, _, _> }),
 );
 const __AWAIT__: SlotDef = SlotDef::new("Py_am_await", "unaryfunc");
 const __AITER__: SlotDef = SlotDef::new("Py_am_aiter", "unaryfunc");
 const __ANEXT__: SlotDef = SlotDef::new("Py_am_anext", "unaryfunc").return_conversion(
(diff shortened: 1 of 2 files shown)


adamreichold:
Please note the discussion in #3190 where multiple options are on the table. Basically, the current design is still rooted in the trait-based object protocols which have now been removed so that other designs can be explored. This includes `__(a)next__` just returning `T: IntoPyCallbackOutput`, i.e. return values being handled in the same manner as for all other `#[pyfunction]`s so that exceptions would be thrown by returning `PyResult::Err`.

PR #3203, opened 2023-06-03
support ordering magic methods for `#[pyclass]`

### LINKED ISSUE #2089: Support `__eq__` for pyclass

davidhewitt:
Closes #2089 

This adds `__lt__`, `__le__`, `__eq__`, `__ne__`, `__gt__`, and `__ge__` as per the Python implementations of what we call `__richcmp__`.

There's a UI test confirming that the user cannot implement split forms and `__richcmp__` simultaneously.

There's also a benchmark comparing implementing these split methods against using `__richcmp__`. I couldn't see a meaningful performance difference, so I'm tempted to deprecate `__richcmp__`, given that's not a magic method which exists in Python. Potentially we can provide options such as the opt-in `#[pyclass(eq, ord)]` to avoid boilerplate for people who don't want to implement six different methods.

### DIFF
--- a/benches/bench_comparisons.rs
+++ b/benches/bench_comparisons.rs
@@ -1,0 +1,70 @@
+use criterion::{criterion_group, criterion_main, Bencher, Criterion};
+
+use pyo3::{prelude::*, pyclass::CompareOp, Python};
+
+#[pyclass]
+struct OrderedDunderMethods(i64);
+
+#[pymethods]
+impl OrderedDunderMethods {
+    fn __lt__(&self, other: &Self) -> bool {
+        self.0 < other.0
+    }
+
+    fn __le__(&self, other: &Self) -> bool {
+        self.0 <= other.0
+    }
+
+    fn __eq__(&self, other: &Self) -> bool {
+        self.0 == other.0
+    }
+
+    fn __ne__(&self, other: &Self) -> bool {
+        self.0 != other.0
+    }
+
+    fn __gt__(&self, other: &Self) -> bool {
+        self.0 > other.0
+    }
+
+    fn __ge__(&self, other: &Self) -> bool {
+        self.0 >= other.0
+    }
+}
+
+#[pyclass]
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+struct OrderedRichcmp(i64);
+
+#[pymethods]
+impl OrderedRichcmp {
+    fn __richcmp__(&self, other: &Self, op: CompareOp) -> bool {
+        op.matches(self.cmp(other))
+    }
+}
+
+fn bench_ordered_dunder_methods(b: &mut Bencher<'_>) {
+    Python::with_gil(|py| {
+        let obj1 = Py::new(py, OrderedDunderMethods(0)).unwrap().into_ref(py);
+        let obj2 = Py::new(py, OrderedDunderMethods(1)).unwrap().into_ref(py);
+
+        b.iter(|| obj2.gt(obj1).unwrap());
+    });
+}
+
+fn bench_ordered_richcmp(b: &mut Bencher<'_>) {
+    Python::with_gil(|py| {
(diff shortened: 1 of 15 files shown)


davidhewitt:
As the new behavior may break users who were attempting to use `__eq__` etc unsuccessfully (unknowingly) this is a 0.20 feature.

benches/bench_comparisons.rs
```diff
@@ -0,0 +1,70 @@
...
+        op.matches(self.cmp(other))
+    }
+}
+
+fn bench_ordered_dunder_methods(b: &mut Bencher<'_>) {
+    Python::with_gil(|py| {
+        let obj1 = Py::new(py, OrderedDunderMethods(0)).unwrap().into_ref(py);
+        let obj2 = Py::new(py, OrderedDunderMethods(1)).unwrap().into_ref(py);
+
+        b.iter(|| obj2.gt(obj1).unwrap());
```
adamreichold, line 51: Could these references be `black_box`ed to make sure the compiler isn't to clever about pre-computing this?

adamreichold:
> so I'm tempted to deprecate __richcmp__, given that's not a magic method which exists in Python. Potentially we can provide options such as the opt-in #[pyclass(eq, ord)] to avoid boilerplate for people who don't want to implement six different methods.

I think this should happen only when that boiler plate reduction is in place.

guide/src/class/protocols.md
```diff
@@ -58,11 +58,28 @@ given signatures should be interpreted as follows:
...
     </details>
 
+  - `__lt__(<self>, object) -> object`
+  - `__le__(<self>, object) -> object`
+  - `__eq__(<self>, object) -> object`
+  - `__ne__(<self>, object) -> object`
+  - `__gt__(<self>, object) -> object`
+  - `__ge__(<self>, object) -> object`
+
+    The implementations of Python's rich-comparison operators `<`, `<=`, `==`, `!=`, `>` and `>=` respectively.
```
adamreichold, line 68: Aren't these just "comparison operators"? To what property does "rich" refer to here?

benches/bench_comparisons.rs
davidhewitt: I think the FFI call to `PyObject_RichCompare` inside the `PyAny::gt` implementation probably prevents the compiler making too many assumptions. I tried:

```rust
b.iter(|| black_box(black_box(obj2).gt(black_box(obj1))).unwrap());
```

And got no noticeable difference?

davidhewitt:
> I think this should happen only when that boiler plate reduction is in place.

Completely agree. This PR is intended to be an addition only. I'd propose to then work on the opt-in implementation of `eq` and `ord`, and then deprecate `__richcmp__` as a final step.

guide/src/class/protocols.md
davidhewitt: This is quoting directly from what Python calls them at https://docs.python.org/3/reference/datamodel.html#object.__ge__

(Well, they write it "rich comparison" so I'll adjust the wording to be the same.)

davidhewitt:
(Pushed the wording adjustment as a new commit, when we're happy with what's going on here I'll squash.)

adamreichold (approved):

davidhewitt:
Thanks, squashed!

bors r=adamreichold

adamreichold:
bors retry

Issue #3113, opened 2023-04-21
syn 2.0 and MSRV


alex:
👋 

Right now, for pyca/cryptography, we're in an awkward space where half our dependencies use syn 2.0, and half use 1.0 (including pyo3). This is unfortunate from a compilation time perspective. The reason not to upgrade, and presumably why pyo3 hasn't, is that syn 2.0 raises their MSRV to 1.56.

I wanted to start a conversation on at what point it would make sense to raise pyo3's MSRV, and thus be able to use syn 2.0?

For pyca/cryptography, our upcoming release will raise our own MSRV from 1.48 to 1.56. You can see some of the data/reasoning behind this here: https://github.com/pyca/cryptography/issues/7172, https://github.com/pyca/cryptography/pull/8195, and https://github.com/pyca/cryptography/pull/8587

adamreichold:
I was expecting us to bump our MSRV to 1.63 when Debian Bookworm becomes stable in "mid 2023" as per https://github.com/PyO3/pyo3/blob/main/Contributing.md#python-and-rust-version-support-policy

adamreichold:
Alpine 3.17 has 1.64 but RHEL 9 seems to carry 1.62, so we would actually end up with 1.62 instead of 1.63 following that policy.

alex:
Based on our experience with /metrics for pyca/cryptography, I would advocate for a slightly smaller jump.

Based on our metrics, our plan is to do 1.56.0 as MSRV for our next release, and then 1.60 as our expected (subject to change) MSRV.

adamreichold:
Out of curiosity and to understand if our policy needs to be amended: Do your metrics include any indication of what users in the 1.56 to 1.61 bracket are running on or how they installed Rust?

adamreichold:
Interesting, so mostly outdated Alpine versions. So glad everyone keeps their container image builds fresh as daisies...

alex:
It likely reflects that musllinux wheel support requires a newer pip, so those users _also_ have an older pip.

davidhewitt:
Sorry for the slow reply here on my part.

I think the aging MSRV in PyO3 is probably slowing us down more than continuing to support Python 3.7 is. I also had the impression that Python 3.7 was still a significant proportion of downloads, e.g. about a third of recent `cryptography` downloads according to https://pypistats.org/packages/cryptography

[...]

PR #3204, opened 2023-06-04
Bump minimum Python version to 3.8

adamreichold:
This is not trying to reap any of the benefits of making that change yet, just the minimum amount of bureaucracy to make it happen.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,12 +191,10 @@ jobs:
         extra-features: ["multiple-pymethods"]  # Because MSRV doesn't support this
         rust: [stable]
         python-version: [
-          "3.7",
           "3.8",
           "3.9",
           "3.10",
           "3.11",
-          "pypy-3.7",
           "pypy-3.8",
           "pypy-3.9"
         ]
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 ## Usage
 
 PyO3 supports the following software versions:
-  - Python 3.7 and up (CPython and PyPy)
+  - Python 3.8 and up (CPython and PyPy)
   - Rust 1.56 and up
 
 You can use PyO3 to write a native Python module in Rust, or to embed Python in a Rust binary. The following sections explain each of these in turn.
--- a/examples/plugin/plugin_api/pyproject.toml
+++ b/examples/plugin/plugin_api/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "maturin"
 
 [project]
 name = "plugin_api"
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 classifiers = [
     "Programming Language :: Rust",
     "Programming Language :: Python :: Implementation :: CPython",
--- a/guide/src/getting_started.md
+++ b/guide/src/getting_started.md
@@ -10,7 +10,7 @@ If you can run `rustc --version` and the version is new enough you're good to go
 
 ## Python
 
-To use PyO3, you need at least Python 3.7. While you can simply use the default Python interpreter on your system, it is recommended to use a virtual environment.
+To use PyO3, you need at least Python 3.8. While you can simply use the default Python interpreter on your system, it is recommended to use a virtual environment.
 
 ## Virtualenvs
 
@@ -128,7 +128,7 @@ build-backend = "maturin"
 
 [project]
 name = "pyo3_example"
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 classifiers = [
     "Programming Language :: Rust",
     "Programming Language :: Python :: Implementation :: CPython",
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -7,7 +7,7 @@ For a detailed list of all changes, see the [CHANGELOG](changelog.md).
 
 ### Drop support for older technologies
 
-PyO3 0.20 has increased minimum Rust version to 1.56. This enables use of newer language features and simplifies maintenance of the project.
+PyO3 0.20 has increased minimum Rust version to 1.56 and minimum Python version to 3.8. This enables use of newer language features and simplifies maintenance of the project.
 
 ## from 0.18.* to 0.19
 
--- a/newsfragments/3204.changed.md
+++ b/newsfragments/3204.changed.md
@@ -1,0 +1,1 @@
+Update the minimum Python version to 3.8.
--- a/noxfile.py
+++ b/noxfile.py
@@ -16,8 +16,8 @@ nox.options.sessions = ["test", "clippy", "fmt"]
 
 
 PYO3_DIR = Path(__file__).parent
-PY_VERSIONS = ("3.7", "3.8", "3.9", "3.10", "3.11")
-PYPY_VERSIONS = ("3.7", "3.8", "3.9")
+PY_VERSIONS = ("3.8", "3.9", "3.10", "3.11")
+PYPY_VERSIONS = ("3.8", "3.9")
 
 
 @nox.session(venv_backend="none")
--- a/pyo3-ffi/README.md
+++ b/pyo3-ffi/README.md
@@ -2,7 +2,6 @@
 
 This crate provides [Rust](https://www.rust-lang.org/) FFI declarations for Python 3.
 It supports both the stable and the unstable component of the ABI through the use of cfg flags.
-Python Versions 3.7+ are supported.
 It is meant for advanced users only - regular PyO3 users shouldn't
 need to interact with this crate at all.
 
@@ -13,7 +12,7 @@ Manual][capi] for up-to-date documentation.
 # Minimum supported Rust and Python versions
 
 PyO3 supports the following software versions:
-  - Python 3.7 and up (CPython and PyPy)
+  - Python 3.8 and up (CPython and PyPy)
   - Rust 1.56 and up
 
 # Example: Building Python Native modules
--- a/pyo3-ffi/src/lib.rs
+++ b/pyo3-ffi/src/lib.rs
@@ -50,7 +50,7 @@
 //! # Minimum supported Rust and Python versions
 //!
 //! PyO3 supports the following software versions:
-//!   - Python 3.7 and up (CPython and PyPy)
+//!   - Python 3.8 and up (CPython and PyPy)
 //!   - Rust 1.56 and up
 //!
 //! # Example: Building Python Native modules
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,7 +121,7 @@
 //! # Minimum supported Rust and Python versions
 //!
 //! PyO3 supports the following software versions:
-//!   - Python 3.7 and up (CPython and PyPy)
+//!   - Python 3.8 and up (CPython and PyPy)
 //!   - Rust 1.56 and up
 //!
 //! # Example: Building a native Python module


davidhewitt:
I'm 100% in favour of bumping MSRV to 1.56.

I'm still open to a decision on whether we think the bump to minimum Python 3.8 needs to happen right away or we should wait for more of the ecosystem to drop 3.7. I realise our policy stated in `Contributing` says the Python and Rust version bumps happen together, so if we delayed the 3.8 bump we would need to adjust wording there as well as consider the impact vs upsides of requiring distributors to adapt to our changing requirements.

If we drop 3.7 then it creates more pressure for the rest of the ecosystem to drop 3.7 too, however for projects like `cryptography` which try very hard to support the vast majority of the ecosystem then maybe we just block them from getting new PyO3 features as per https://github.com/PyO3/pyo3/issues/3113#issuecomment-1540951525

@alex, do you have a projection of how long `cryptography` is likely to want to keep 3.7 support? Do you think it's feasible for you to pin to an older PyO3 purely for 3.7 builds?

alex:
We don't yet know how long we'll keep 3.7. Our approach to deprecations/removals is based on empirical usage data, not upstream support. Per https://pypistats.org/packages/cryptography, 3.7 is still at nearly 1/4 of our downloads. One of my TODOs is to investigate where those are coming from, likely Debian, in which case perhaps they'll start to go down with the upcoming release.

In terms of pinning for 3.7 builds, I looked into this previously, and I could not find a way to potentially depend on multiple different pyo3 versions. `cargo` only lets you have a single version of pyo3-ffi in your `Cargo.toml`, because only one crate may link the same shared library.

alex:
FWIW, if y'all are willing to hold on to 3.7, we (@reaperhulk and I) are willing to do anything required in terms of keeping it working maintenance wise.

davidhewitt:
Understood; it would be helpful to have a sense of what we're waiting for if we do choose to wait.

Just curious - what's the main motivation for not wanting to drop the 3.7 userbase (beyond the notable size)? Releasing a new cryptography which doesn't support 3.7 may help provide them impetus to upgrade. 

> In terms of pinning for 3.7 builds, I looked into this previously, and I could not find a way to potentially depend on multiple different pyo3 versions. cargo only lets you have a single version of pyo3-ffi in your Cargo.toml, because only one crate may link the same shared library.

Indeed, I also took a quick look and came to the same conclusion.

davidhewitt:
Just had another MSRV failure in https://github.com/PyO3/pyo3/pull/3203; @adamreichold would you be potentially willing to split the MSRV bump into a separate PR which we could land immediately and then we can continue to debate the option of Python 3.7 support in this thread?

alex:
Our view is that we have a pretty foundational place in the ecosystem, since we're a dependency of all sorts of things (certbot, twisted, paramiko, etc.). As a result, when we drop something it has pretty wide-ranging impacts -- while old versions do continue to work, they don't get security updates or updates for other ecosystem things.

In terms of encouraging people to upgrade, there's definitely a game of chicken and egg. Us dropping 3.7 (or warning about it in advance) does encourage some people to upgrade, however if we're too aggressive, then we impact too large a set of users, at which point maintainers of those various other projects (e.g. certbot) get put in an uncomfortable place, and apply back-pressure to us.

For Python 3.6 (which we dropped in our latest release) we started issuing warnings somewhere around 10-20%, and we ultimately dropped it around 7%.

So we definitely want to encourage users to upgrade, but at 1/4 of our downloads, it's probably too early to start applying sticks yet, we need things like the new Debian release to come out, and 3.7 to go EOL with upstream, first, then we can come in and apply the sticks to the laggards.

FWIW, I pulled the stats on which Linux distros are the source of our 3.7 downloads, and its not what I expected. I don't yet know what the dominance of 

image

davidhewitt:
Oh wow. It looks like Amazon Linux 2 does indeed include 3.7 by default if you just install `python3`, and that Amazon will provide LTS support for Amazon Linux 2 until 2025.

https://aws.amazon.com/amazon-linux-2/faqs/
https://docs.aws.amazon.com/linux/al2023/release-notes/version-compare.html

Amazon Linux 2023 was released on 15th March. Sounds like these are distros derived from Red Hat Linux family with releases planned every two years and support lasting five. With 3.9 shipped as the default in Amazon Linux 2023, we may have a similar position in two years time with 3.9 EOL in Oct 2025 but Amazon Linux 2023 continuing until 2028.

https://aws.amazon.com/linux/amazon-linux-2023/faqs/
https://aws.amazon.com/about-aws/whats-new/2023/03/amazon-linux-2023/
https://docs.aws.amazon.com/linux/al2023/ug/relationship-to-fedora.html

So this 50% of 3.7 downloads does have an upgrade path already, and Amazon is committed to pushing security back ports for them (they have a `python-cryptography` package, although for Amazon Linux 2 I think it's for the Python 2 system python unfortunately).

alex:
Oops, I just noticed I left a sentence half written: I was not expecting Amazon Linux to be such a dominant thing, and I don't yet have a sense for what it means for our Python 3.7 support. @reaperhulk and I will need to discuss.

adamreichold:
> Just had another MSRV failure in #3203; @adamreichold would you be potentially willing to split the MSRV bump into a separate PR which we could land immediately and then we can continue to debate the option of Python 3.7 support in this thread?

Sure, but the bump seems to be more complicated than expected insofar there is a lot Clippy-related work to be done, i.e. I am not sure when I will be able to do this.

Concerning Python 3.7, what do you think about dropping support from PyO3's mainline but committing to keep maintaining 0.19.x upstream? cryptography could then stick to 0.19.x which should not be an issue functionality-wise as we do not support anything crossing extensions yet.

alex:
I have a strong preference for keeping 3.7 support on main -- this keeps us engaged with upstream and contributing back. As I said, we're happy to take on any work needed to keep it functional.

adamreichold:
> As I said, we're happy to take on any work needed to keep it functional.

I think the main issue is that this work if often poorly defined and not separable from the rest of the work, i.e. random things fail because of some incompatibility with 3.7 and the code stay more complicated as the various workarounds have to stay in place. And OTOH, if you do make changes to 0.19.x, porting them forward to main and making the necessary adjustments is well-defined and has limited chance of interference. Obviously, I see the downside that it can end up being twice the work.

> Sure, but the bump seems to be more complicated than expected insofar there is a lot Clippy-related work to be done, i.e. I am not sure when I will be able to do this.

@davidhewitt One solution to this would be to do check builds or at least allow warnings for the MSRV clippy CI job. I actually think that check builds would be ideal because we want to be Clippy-clean on stable, but we just want to build on MSRV. Especially since being Clippy-clean on stable and MSRV can actually conflict.

davidhewitt:
> @davidhewitt One solution to this would be to do check builds or at least allow warnings for the MSRV clippy CI job. I actually think that check builds would be ideal because we want to be Clippy-clean on stable, but we just want to build on MSRV. Especially since being Clippy-clean on stable and MSRV can actually conflict.

Agreed with this, if you're short on time I'm happy to give this a go tonight.

adamreichold:
> Agreed with this, if you're short on time I'm happy to give this a go tonight.

Would be glad if you looked into it as I won't be able to do anything about it tonight.

davidhewitt:
Have pushed #3208 for the MSRV bump, which looks like it is going to pass.

As for the Python bump, I'm very torn. As a maintainer I would be glad to see 3.7 go; our CI is already extremely heavy and we'll soon be adding more to it with 3.12. I also think that it's likely that reworks we add for interpreter initialization (which looks necessary for part of work to support 3.12 as per https://github.com/PyO3/pyo3/pull/2709#issuecomment-1540779321). We could just not offer the reworked API on 3.7, though it's a bit unfortunate.

On the other side of the coin, I'm very keen to help PyO3 extension authors support 3.7 if they feel that there is good reason to. I want PyO3 to be seen as a useful part of the ecosystem that authors can rely on, so if axing 3.7 support too fast will isolate extension authors to be stuck on older versions of PyO3, that seems like a marketing downside to me. Especially when we haven't yet added 3.12 support and I'm hopeful we'll also land other big improvements over the second half of this year.

What would we do here if PyO3 were past 1.0? I think we would keep the original plan of doing MSRV bumps and Python minimum bumps together, as both of them would be cause for a major release (rather than just a minor release with feature additions).

@alex, I'm keen to hear the outcome of the evaluation of Amazon Linux 2 impact on 3.7 support between yourself and @reaperhulk (but no need to rush, 0.20 won't be released for perhaps a month or so).

adamreichold:
> so if axing 3.7 support too fast will isolate extension authors to be stuck on older versions of PyO3, that seems like a marketing downside to me.

I think that depends on the marketing. If we do commit to maintaining multiple PyO3 versions for increases compatibility? Yes, packages will not get both the newest PyO3 features and compatibility with 3.7, but we could keep existing environments alive as long as they are in use with repeatedly facing hard choice between keeping the code base healthy and keeping it compatible.

reaperhulk:
I haven't had a chance to talk with @alex much about this yet, but we've recently come off the 0.15.x branch (we made it to 0.18.x with our latest release). During our time on 0.15.x we needed the pypy3.9 compatibility work that was backported into 0.15.2 (thanks to everyone on that!). Additionally, we recently got an issue filed about pypy3.10, which would have been problematic again except we were able to move to 0.18.x just in time. 😄 Loss of forward compatibility is definitely something we definitely worry about.

An official policy on supporting a long-lived branch (with some metrics for when that support ends) might assuage that concern. That said, I wonder if the maintenance required to keep an older release up-to-date and forward compatible will exceed keeping that support in `main`.

Thinking more broadly, having a primary release (where you get new APIs, features, etc) that isn't compatible with the set of versions in wide use increases the adoption cost for projects looking to switch to PyO3. Do you choose the newer features or do you support all the versions your users are asking for? That's a decision each project will need to make as part of evaluating PyO3 if you prioritize dropping support.

Regardless of how you choose, I want to express my thanks for the volunteers on this project. We love that we've been able to ship safer code to our users for the past 2+ years (and well over 1B downloads) and a huge amount of credit goes to this project for making that possible.

adamreichold:
> That said, I wonder if the maintenance required to keep an older release up-to-date and forward compatible will exceed keeping that support in main.

Personally, I am rather convinced that yes, forward porting will be more work quantitively. I just think that it would keep "compatibility work" from interfering with "feature work." For example, I have people contributing rarely in mind which I think often hit these complications while doing something else entirely.

> Thinking more broadly, having a primary release (where you get new APIs, features, etc) that isn't compatible with the set of versions in wide use increases the adoption cost for projects looking to switch to PyO3. Do you choose the newer features or do you support all the versions your users are asking for? That's a decision each project will need to make as part of evaluating PyO3 if you prioritize dropping support.

Indeed we would be pushing that choice out into the ecosystem somewhat which admittedly is not a nice thing to do. The main argument for doing this that I can see is that a specific project with specific goals is in a better to position to make an informed decision. We as an upstream project would need to make everyone happy though which notably includes trying to grow our contributor and probably even more important reviewer base.

alex:
As an alternative idea how about: 3.7 jobs are allowed to fail on PRs (i.e., they can be merged), and then we can get ping'd to resolved the issues asynchronously?

davidhewitt:
That's an interesting proposal and may be an acceptable compromise. Our `bors` infrastructure was quite crude and would fail to merge if any check was failing. This helped keep `main` greener but regularly we had CI blockages anyway due to MSRV.

I'm just trying to enable GH merge queue in #3206 - so I'm in the process of learning if it is possible for 3.7 to be failing but the merge queue to pass. 

Downsides I see of allowing 3.7 to fail on merge:
- There are issues of the kind like #2927 where contributors have run into issues with specific Python versions when writing tests. @alex I suspect this is likely to be the sort of busywork that would be created by 3.7 maintenance, i.e. rewriting tests which don't pass on 3.7. I understand from the above that you'd be willing to help out with that (until a time when you no longer need 3.7 support)?
- It may slow down cutting of releases if we have `main` knowingly broken on 3.7.

reaperhulk:
If you use something like https://github.com/re-actors/alls-green to set a list of required jobs that must pass then the merge queue should work fine with potentially failing jobs. You can see how we use all-green in our [ci.yml](https://github.com/pyca/cryptography/blob/main/.github/workflows/ci.yml#L396), although we just use auto-merge, not the queue feature.

alex:
Yes, that's right, any sort of issues required to keep 3.7 green (post-hoc) we could ping'd on and we'd do our best to resolves in a timely manner.

adamreichold:
Closing this as the consensus seems to be that we keep support for Python 3.7 on the main branch.

davidhewitt:
At least for 0.20, I'm willing to see how this works out. If it turns out to be a lot of bother for everyone, let's reconsider dropping. And hopefully users of cryptography will hurry up and update soon 😅

alex:
Thanks!

PR #3208, opened 2023-06-04
bump msrv to 1.56

davidhewitt:
The MSRV changes from #3204, plus a commit which should hopefully make CI pass.

With luck this is mergeable and resolves CI pain while we decide what to do about the Python version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,32 @@ jobs:
       - name: Check rust formatting (rustfmt)
         run: nox -s fmt-rust
 
+  check-msrv:
+    needs: [fmt]
+    runs-on: ubuntu-latest
+    if: github.ref != 'refs/heads/main'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.56.0
+          targets: x86_64-unknown-linux-gnu
+          components: clippy,rust-src
+      - uses: actions/setup-python@v4
+        with:
+          architecture: "x64"
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: check-msrv-1.56.0
+        continue-on-error: true
+      - run: python -m pip install --upgrade pip && pip install nox
+      - name: Prepare minimal package versions
+        run: nox -s set-minimal-package-versions
+      - run: nox -s check-all
+
+    env:
+      CARGO_BUILD_TARGET: x86_64-unknown-linux-gnu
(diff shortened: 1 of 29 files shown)


alex:
Looks like there's a bunch of clippy warnings come alive that are unhappy about passing slices where an array will do.

davidhewitt:
Yes, I will correct that later on. It's expected, the first commit of this PR removed some clippy exclusions incompatible with 1.48.

davidhewitt:
It looks like this is going to pass, so I'm going to push it through as we're all in favour of having this.

@adamreichold I'm sure there's plenty of tidy-ups we can land now that we've done the MSRV bump. I guess at the very least I will try sometime to go through the Rust release notes and audit for changes we can migrate to. Feel free to push any cleanup PRs for stuff you've noticed or have on a list!

.github/workflows/ci.yml
```diff
@@ -34,6 +34,32 @@ jobs:
...
+    needs: [fmt]
+    runs-on: ubuntu-latest
+    if: github.ref != 'refs/heads/main'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.56.0
+          targets: x86_64-unknown-linux-gnu
+          components: clippy,rust-src
```
adamreichold, line 47: Note for possible clean-up: We do not need `clippy`.

tests/test_class_attributes.rs
```diff
@@ -96,7 +96,7 @@ fn recursive_class_attributes() {
 }
 
 #[test]
-#[cfg(panic = "unwind")]
+#[cfg_attr(cfg_panic, cfg(panic = "unwind"))]
```
adamreichold, line 99: I wonder if we actually need this? IIRC, libtest does not work without unwinding in any case?

adamreichold (approved):
I see we already silently turned Clippy into check specificially for 1.48 but of course this broke when switching to 1.56. I think the solution with a separate CI job is strictly better so that people (include ourselves) understand what is going without remembering or finding the single if statement in our Nox manifest.

tests/test_class_attributes.rs
adamreichold: Apparently not as the cfg is name `panic_unwind` in `print_feature_cfgs` but `cfg_panic` here and still everything passed. Will remove this in a follow up.

davidhewitt: Haha oops 😅

PR #3209, opened 2023-06-05
Remove the conditional compilation flags which are made redundant  by the MSRV bump

adamreichold:


src/conversions/std/array.rs
```diff
@@ -1,308 +1,118 @@
+use crate::conversion::{AsPyPointer, IntoPyPointer};
```
adamreichold, line 1: The diff in this file looks messy due to the indentation changes but I did not touch a single line of code. I just merged the inner modules in the outer one, i.e. I removed and moved code.

xtask/src/utils.rs
```diff
@@ -1,47 +1,15 @@
...
-        buf.push(' ');
-        buf.push_str(&arg.to_string_lossy());
-    }
-    buf.push('`');
-    buf
-}
-
-#[rustversion::before(1.57)]
-pub fn format_command(command: &Command) -> String {
-    // Debug impl isn't as nice as the above, but will do on < 1.57
```
adamreichold, line 31: I chose to drop the distinction to get rid of rustversion entirely. I think this is possible since we do not use xtask in the CI any more and any one using it for local development will most likely be running stable or nightly.

At the danger of opening old wounds, I would like to bring up again that we currently have two xtask-like mechanisms and I think the emphasis has definitely shifted towards Nox. Having another one that is rarely exercised seems confusing to contributors at least. I also wonder whether we want to invest the effort for modernising this, e.g. replacing struct and clap v2 by clap v4 wholesale. Personally, I would suggest dropping the Rust-based build automation.

adamreichold:
> Agreed that xtask has been superseded by nox for us now. I think it'd be a reasonable reduction in complexity to move all remaining functionally from there to nox (which isn't already duplicated).

I had a look and if I am not mistaking the only thing remaining would be building the docs?

davidhewitt:
Yes, from memory I think that's the case too.

PR #3210, opened 2023-06-05
Drop the xtask helper as it is superseded by Nox.

adamreichold:
While the xtask code base is better engineered than our slightly messy Nox manifest, all functionality is available via Nox including some that is not available via xtask. Finally, only the Nox sessions are used in the CI by now so that xtask does not see regular automated usage.


noxfile.py
```diff
@@ -165,13 +164,9 @@ def _check(env: Dict[str, str]) -> None:
 @nox.session(venv_backend="none")
 def publish(session: nox.Session) -> None:
     _run_cargo_publish(session, package="pyo3-build-config")
```
davidhewitt, line 166: Thanks, I had removing these sleeps in the back of my mind too.

davidhewitt (commented):
Looks like the netlify build still has a reference to cargo xtask doc 😊

adamreichold:
> Looks like the netlify build still has a reference to cargo xtask doc blush

Indeed. I think that setting `RUSTDOCFLAGS` there never worked as the xtask implementation would override that. I tried to actually implement amending `RUSTDOCFLAGS`.

I also noticed that `Contributing.md` still references them and suggested `nox` as a replacement for `cargo xtask ci` which is not exactly the same as it does not build the docs. Should I add the `docs` session to the list of default sessions?

adamreichold:
> Where are those internal docs published?

Ah, https://pyo3.rs/internal/doc/pyo3/, and indeed I do not see the warning there.

adamreichold:
Ok, so the deploy preview does show the banner:

grafik

I suspect the design is intentional to be hard to miss but not block the UI below?

davidhewitt:
> Should I add the docs session to the list of default sessions?

I think it's worth it; that way a new contributor can run nox straightforwardly to get a sense of whether CI might be happy.

davidhewitt:
> > I think it's worth it; that way a new contributor can run nox straightforwardly to get a sense of whether CI might be happy.
> 
> Will do.
> 
> > I'm sure in the past it was centered and the page flowed around it, I wonder if the CSS has bitrotted somewhat...
> 
> Are you ok with tackling this separately? Wrangling style attributes is not my forte...

Sure thing, I'll make a note to revisit 👍

davidhewitt (approved):
Looks like CI will pass so I'll trigger bors. Thanks for tidying this all up!

bors r+

PR #3211, opened 2023-06-05
Enable v2 resolver on the workspace to enable dropping PYO3_CI workaround

adamreichold:


davidhewitt:
How do you feel about instead going to 2021 edition, which I think implies resolver = 2?

(Could be a follow-up I guess?)

adamreichold:
> How do you feel about instead going to 2021 edition, which I think implies resolver = 2?
> 
> (Could be a follow-up I guess?)

Yes, I would like to bump the edition separately.

It is trivial to remove the `resolver = "2"` line, but I would like to see the effects of this in isolation as environment variables are a classic case for [spooky action at a distance in software](https://blog.sunfishcode.online/no-ghosts/).

adamreichold:
bors try

davidhewitt (approved):
Hooray!

bors r+

PR #3213, opened 2023-06-06
try failing 3.7 check in the merge queue

davidhewitt:
Ref https://github.com/PyO3/pyo3/pull/3204#issuecomment-1578026122

I'm going to try deliberately fail 3.7 in the merge queue, I think at the moment this will fail to merge, we can consider if we want to change this later.


davidhewitt:
Closing in favour of #3227 which I think should implement what we want.

PR #3215, opened 2023-06-07
Bump Rust edition to 2021 and make current Clippy happy.

adamreichold:


davidhewitt (approved):
Thanks!

It should be that you can launch a full CI run (and merge) by clicking the "merge when ready" button.

I guess `bors try` probably works still too. I've been wondering if we can use a label to trigger a full CI run as a substitute...

PR #3217, opened 2023-06-07
ci: enable full builds with a label

davidhewitt:
As a replacement for `bors try`, this PR adds a `CI-build-full` label which, when present, should trigger the full build instead of just the 3.11 runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,7 +129,7 @@ jobs:
       CARGO_BUILD_TARGET: ${{ matrix.platform.rust-target }}
 
   build-pr:
-    if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


davidhewitt:
Looks like that's working as hoped! Note that this works like `CI-no-fail-fast` where it doesn't automatically re-trigger CI on application, you need to push / reopen the PR to trigger a new actions run.

messense (approved):

adamreichold:
One thing we should probably keep in mind is that this runs the full CI on the branch, not the result of merging the branch as the merge queue does or as bors would have.

davidhewitt:
Very true! The merge queue will still run on the merge result before completing the merge, so we may see failures at that stage.

PR #3219, opened 2023-06-08
ci: fail ffi-check on pypy failures

davidhewitt:
To try and catch possible issues with upcoming PyPy 3.10 releases.


davidhewitt:
Looks like this caught a tiny issue on pypy 3.10, and now passes! Will merge now as seems uncontroversial.

PR #3221, opened 2023-06-08
docs: remove unneeded cfg, fixup internal banner

davidhewitt:
Removes some `cfg_attr` attributes from docs which are no longer needed.

Also fixes up the internal banner as agreed in #3210

### DIFF
--- a/.netlify/internal_banner.html
+++ b/.netlify/internal_banner.html
@@ -1,0 +1,47 @@
+<div id='pyo3-internal-banner'>
+    <div style="white-space: nowrap;">
+        ⚠️ Internal Docs ⚠️ Not Public API 👉
+        <a href='https://pyo3.rs/main/doc/pyo3/index.html' style='color:red;text-decoration:underline;'>
+            Official Docs Here
+        </a>
+    </div>
+    <style id="pyo3-noscript-style">
+        body {
+            padding-top: 2em;
+        }
+    </style>
+    <style>
+        #pyo3-internal-banner {
+            position: fixed;
+            display: flex;
+            align-items: center;
+            justify-content:
+                center;
+            z-index: 99999;
+            color: red;
+            left: 0;
+            right: 0;
+            top: 0;
+            height: 2em;
+            border: 3px solid red;
+            width: 100%;
+            overflow-x: hidden;
+            background-color: var(--target-background-color);
+        }
+
+        @media (max-width: 700px) {
+            #pyo3-internal-banner {
+                top: 50px;
+                margin-bottom: 10px;
+            }
+        }
+    </style>
+</div>
+<script>
+    // when javascript is active, splice the banner into a "sticky" location
+    // inside the doc body for best appearance
+    banner = document.getElementById("pyo3-internal-banner")
+    banner.style.position = "sticky"
+    document.getElementsByTagName("main")[0].prepend(banner)
+    document.getElementById("pyo3-noscript-style").remove()
+</script>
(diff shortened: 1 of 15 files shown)


noxfile.py
```diff
@@ -305,6 +305,7 @@ def docs(session: nox.Session) -> None:
     if "nightly" in session.posargs and "internal" in session.posargs:
         rustdoc_flags.append("--Z unstable-options")
```
adamreichold, line 306: If this removes all usage of the `docsrs` flag, maybe we want to remove it form the Noxfile here as well?

.netlify/internal_banner.html
```diff
@@ -0,0 +1,47 @@
...
+        @media (max-width: 700px) {
+            #pyo3-internal-banner {
+                top: 50px;
+                margin-bottom: 10px;
+            }
+        }
+    </style>
+</div>
+<script>
+    // when javascript is active, splice the banner into a "sticky" location
```
adamreichold, line 41: Uff, this is clever, but maybe too clever?

What do you think about providing our own index.html which contains rustdoc's index.html as a frame and our warning as another frame on top? I hope this could be less brittle as it would not depend on the structure Rustdoc generates.

davidhewitt: Ahh that's an interesting idea, you're thinking as part of the nox process we'd rename the generated index.html and write a new one? (I can't see a rustdoc option to do such index.html processing?)

noxfile.py
davidhewitt: Unfortunately there's a few uses remaining, related to the `doc_cfg` feature.

.netlify/internal_banner.html
adamreichold: Yes, exactly. Yes, we are running our own code anyway. This may be tainted by me being more comfortable with doing Python-based processing in the CI than JavaScript-based processing on each visiting client. But mainly to avoid having to integrate with Rustdoc's DOM structure.

davidhewitt: Agreed, it's better to have the complexity where we can control it. I'll fixup tomorrow or early next week.

davidhewitt: Uff, actually it's not just `index.html` but every doc page that would have to be redone like this but every html page, if we want the banner to be on all of them (like it is here).

While the solution I've pushed here is a little icky, it does at least render on all pages without too much effort. @adamreichold would you be willing to accept us leaving this as proposed in this PR, given the internal docs are really just for implementation detail and we expect very few users to ever look at it?

Alternatively I could just drop the second commit; the banner rendering is a bit broken but it's not a huge problem.

adamreichold: > Uff, actually it's not just index.html but every doc page that would have to be redone like this, if we want the banner to be on all of them (like it is here).

I am fine to do this like it is proposed if the other way requires a lot of effort. I am just not sure I understand the problem though.

Do you mean because the banner would not be visible when a page is linked to directly?

My idea was that we just replace the landing page with a frameset-based on containing the banner in one frame and the original but renamed index.html in another frame. Any navigation in the docs would then happen within the second frame and not affect the parent frameset page and therefore the banner?

davidhewitt: > Do you mean because the banner would not be visible when a page is linked to directly?

Exactly this. While we could fix it, I think we'd have to transform every html page, and if we'd transformed every page in this way we would need to rewrite all links too. Otherwise the original links within the inner frame would go to a page itself with a frame and a banner, etc. I think they would end up nesting horribly.

So while I think you're right that reprocessing the html as part of deployment would be the better solution, I think it's hard enough that I'd like to step away from that for now and spend any dev time I get on other things. This PR already is an improvement imo so seems worth merging as-is until a time we get a chance to build the html reprocessing solution.

PR #3222, opened 2023-06-08
simple getitem example

krpatter-intc:
Figured I'd pass this recipe along some, but if this is too simple for inclusion, that's understandable and I'll try to find another place to put something like this.

### DIFF
--- a/examples/getitem/pyproject.toml
+++ b/examples/getitem/pyproject.toml
@@ -1,0 +1,16 @@
+[build-system]
+requires = ["maturin>=1,<2"]
+build-backend = "maturin"
+
+[project]
+name = "getitem"
+version = "0.1.0"
+classifiers = [
+    "License :: OSI Approved :: MIT License",
+    "Development Status :: 3 - Alpha",
+    "Intended Audience :: Developers",
+    "Programming Language :: Python",
+    "Programming Language :: Rust",
+    "Operating System :: POSIX",
+    "Operating System :: MacOS :: MacOS X",
+]
--- a/examples/getitem/src/lib.rs
+++ b/examples/getitem/src/lib.rs
@@ -1,0 +1,82 @@
+// This is a very fake example of how to check __getitem__ parameter and handle appropriately
+use pyo3::exceptions::PyTypeError;
+use pyo3::prelude::*;
+use pyo3::types::PySlice;
+use std::os::raw::c_long;
+
+#[derive(FromPyObject)]
+enum IntOrSlice<'py> {
+    Int(i32),
+    Slice(&'py PySlice),
+}
+
+#[pyclass]
+struct ExampleContainer {
+    // represent the maximum length our container is pretending to be
+    max_length: i32,
+}
+
+#[pymethods]
+impl ExampleContainer {
+    #[new]
+    fn new() -> Self {
+        ExampleContainer { max_length: 100 }
+    }
+
+    fn __getitem__(&self, key: &PyAny) -> PyResult<i32> {
+        if let Ok(position) = key.extract::<i32>() {
+            return Ok(position);
+        } else if let Ok(slice) = key.downcast::<PySlice>() {
+            // METHOD 1 - the use PySliceIndices to help with bounds checking and for cases when only start or end are provided
+            // in this case the start/stop/step all filled in to give valid values based on the max_length given
+            let index = slice.indices(self.max_length as c_long).unwrap();
+            let _delta = index.stop - index.start;
+
+            // METHOD 2 - Do the getattr manually really only needed if you have some special cases for stop/_step not being present
+            // convert to indices and this will help you deal with stop being the max length
+            let start: i32 = slice.getattr("start")?.extract()?;
+            // This particular example assumes stop is present, but note that if not present, this will cause us to return due to the
+            // extract failing. Not needing custom code to deal with this is a good reason to use the Indices method.
+            let stop: i32 = slice.getattr("stop")?.extract()?;
+            // example of grabbing step since it is not always present
+            let _step: i32 = match slice.getattr("step")?.extract() {
+                // if no value found assume step is 1
+                Ok(v) => v,
+                Err(_) => 1 as i32,
+            };
+
+            // Use something like this if you don't support negative stepping and want to give users
+            // leeway on how they provide their ordering
+            let (start, stop) = if start > stop {
+                (stop, start)
+            } else {
+                (start, stop)
+            };
+            let delta = stop - start;
+
+            return Ok(delta);
+        } else {
+            return Err(PyTypeError::new_err("Unsupported type"));
+        }
+    }
+    fn __setitem__(&self, idx: IntOrSlice, value: u32) -> PyResult<()> {
+        match idx {
+            IntOrSlice::Slice(slice) => {
+                let index = slice.indices(self.max_length as c_long).unwrap();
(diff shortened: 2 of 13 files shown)


examples/getitem/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn new() -> Self {
+        ExampleContainer{}
+    }
+
+    fn __getitem__(&self, key: Py<PyAny>, _py: Python) -> PyResult<i32>{
+        // test
+        let key_ref: &PyAny = &key.as_ref(_py);
+        let key_type = key_ref.get_type();
+        let type_name = key_type.name().unwrap();
+        let result = match type_name {
```
adamreichold, line 23: I think this should not match on the type name, rather trying to extract what is expected and retry otherwise, e.g.

```rust
if let Ok(position) = key.extract::<i32>() {
  ..
} else if let Ok(slice) = key.extract::<PySlice>() {
  ..
} else {
  Err(PyTypeError::new_err("Unsupported type"))
}
```

examples/getitem/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+#[pymethods]
+impl ExampleContainer {
+    #[new]
+    fn new() -> Self {
+        ExampleContainer{}
+    }
+
+    fn __getitem__(&self, key: Py<PyAny>, _py: Python) -> PyResult<i32>{
+        // test
+        let key_ref: &PyAny = &key.as_ref(_py);
```
adamreichold, line 20: I think you can just take `key: &PyAny` as the function argument.

examples/getitem/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[build-system]
+requires = ["maturin>=0.13,<0.14"]
```
adamreichold, line 2: I think Maturin just reached 1.0 and we should use that current version here.

examples/getitem/src/lib.rs
krpatter-intc: I never could get extract to work for PySlice:

```
27  |         } else if let Ok(slice) = key.extract::<PySlice>(_py){
    |                                                 ^^^^^^^ the trait `PyClass` is not implemented for `PySlice`
```

examples/getitem/src/lib.rs
krpatter-intc: Ah, sorry I missed how to do that when building this, switched it to use `&PyAny` instead (which means I no longer have to have the GIL for this particular function (?) which is great).

examples/getitem/pyproject.toml
krpatter-intc: Updated to 1

examples/getitem/src/lib.rs
adamreichold: > which means I no longer have to have the GIL for this particular function (?) which is great).

No, the GIL is always held when you call into a `#[pyfunction]`, for whole duration of that call. This just means that PyO3 automatically does that `as_ref` call for you. (And that you do not need the GIL token, but this can always be recovered from a `&PyAny` via [`PyAny::py`](https://docs.rs/pyo3/latest/pyo3/types/struct.PyAny.html#method.py).)

examples/getitem/src/lib.rs
adamreichold: You probably need to replace `extract` by `downcast` then.

adamreichold: > I still needed getattr. It wasn't intuitive to me how to get from PySlice to PySliceIndices (https://github.com/PyO3/pyo3/blob/main/src/types/slice.rs)...or if that was really what was the right thing....hopefully gettattr is still the right thing here.

I think you want to call [`PySlice::indices`](https://docs.rs/pyo3/latest/pyo3/types/struct.PySlice.html#method.indices) by passing in the conceptual length of your container. (You should probably amend your example to actually store a container like `Vec<i32>` created via `(0..length).collect()`.) The methods needs to know the length to be able to handle negative indices "counting from the end."

> Is it right that:

Yes, that is basically correct.

examples/getitem/src/lib.rs
```diff
@@ -4,27 +4,50 @@ use pyo3::prelude::*;
...
+            println!("Delta: {}", delta);
+
+            // METHOD 2 - Do the getattr manully really only needed if you have some special cases for stop/_step not being present
+            // convert to indices and this will help you deal with stop being the max length
             let start: i32 = slice.getattr("start")?.extract()?;
+            // This particular example assumes stop is present, but note that if not present, this will cause us to return due to the
+            // extract failing. Not needing custom code to deal with this is a good reason to use the Indices method.
             let stop: i32 = slice.getattr("stop")?.extract()?;
+            // example of grabbing step since it is not always present
+            let _step: i32 = match slice.getattr("step")?.extract() {
```
adamreichold, line 36: I think this is forwarding the wrong error, i.e. `getattr` will fail if the attribute is not present, not `extract`.

examples/getitem/src/lib.rs
adamreichold: Personally, I am not a fan of showing both usage of `indices` and manually calls to `getattr` as it might confuse users even more. But let's see what other people think...

krpatter-intc: It seems a lot of the examples currently showed getattr in the docs so that's how I started, but I agree that indices is better for 95% of the use cases. I could just see someone needing the original knowledge of "no end given" or "no step given" just to make sure we have same functionality here that is available in python.

If it is confusing, I could do two separate containers with each as an example?

Add links to the docs we've referenced so far as well....

examples/getitem/src/lib.rs
krpatter-intc: I believe the attribute is always present but the value is None

examples/getitem/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+#[pymethods]
+impl ExampleContainer {
+    #[new]
+    fn new() -> Self {
+        ExampleContainer { max_length: 100 }
+    }
+
+    fn __getitem__(&self, key: &PyAny) -> PyResult<i32> {
```
davidhewitt, line 20: As a passing thought, it would be quite nice to eventually put in this example  `key: IntOrSlice<'py>` where that type is defined as

```rust
#[derive(FromPyObject)]
enum IntOrSlice<'py> {
    Int(i32),
    Slice(&'py PySlice),
}
```

I think the only reason that you wouldn't want to consider that today is performance of the error cases, if I recall correctly?

krpatter-intc: I'm not familiar with the performance delta, but I added this enum possibility on the __setitem__ side to show that methodology as well

examples/getitem/tests/example.py
davidhewitt, line 1: Is this file intentionally (almost) duplicated with the one below?

davidhewitt (commented):
Thanks, this is looking good to me!

Just one closing question (see comment below). Also, would you be willing to squash the commits into one please? We can't squash from the Github web UI because we use the merge queue.

examples/getitem/tests/example.py
krpatter-intc: Yeah, it seemed that's what the other examples were doing, but I will remove that here if that is preferred and just leave the test.

davidhewitt: Ah ok - I haven't revisited the other examples in a little while, so had forgotten this. Thanks for tidying up 👍

davidhewitt (approved):
Thanks again, looks great!

PR #3225, opened 2023-06-11
ci: pool can be dirty from other threads

davidhewitt:
I hit this failure in CI: https://github.com/PyO3/pyo3/actions/runs/5237094306/jobs/9455062113?pr=2709#step:19:732

I'm pretty sure what's going on there is that `pool_not_dirty()` is too strong an assertion - another thread which is testing putting a delayed clone or drop onto the pool can trigger a race. I was able to reproduce the same race before applying this patch.


adamreichold (approved):
Agreed. I have also hit this a few times, but just never found the motivation to try to actually fix it. Thanks!

PR #3227, opened 2023-06-12
ci: allow failure on 3.7 and pypy nightly

davidhewitt:
This follows up #2709 to also allow 3.7 jobs to fail as per https://github.com/PyO3/pyo3/pull/3204#issuecomment-1577832966

Also makes pypy nightly allow failures.

This way 3.7 issues and dev versions won't block contributors from merging.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ on:
 
 jobs:
   build:
-    continue-on-error: ${{ inputs.python-version == '3.12-dev' }}
+    continue-on-error: ${{ contains(fromJSON('["3.7", "3.12-dev", "pypy3.7", "pypy3.10-nightly"]'), inputs.python-version) }}
     runs-on: ${{ inputs.os }}
     steps:
       - uses: actions/checkout@v3


adamreichold (approved):
I have to admit that I am uncomfortable with our CI results being harder to understand, but we discussed this at length and the trade-off has reached consensus.

davidhewitt:
I think at least for the dev versions it's nice for them to not block typical PRs. I do agree that if 3.7 failures become confusing for contributors and users we should reconsider dropping.

Also, I'd love to find a way to simplify or speed up CI if at all possible. I'm not sure we can avoid the full build matrix, however maybe there are ways to clean up the test suite or other optimisations. I wonder if `cargo nextest` can help?

adamreichold:
>  I wonder if cargo nextest can help?

My gut feeling is that it won't help that much. I will most likely help with test runtime, but we still need to handle doctests the old way as they are supported yet by nextest IIRC. But mainly, I suspect that our CI is mostly spent rebuilding our crates for different feature combinations.

Try to turn that gut feeling into something actionable, maybe we could limit the feature combinations to fewer targets or maybe even the Clippy runs and only run a full feature set for most of the builds?

davidhewitt:
I agree with this, though I'm not exactly sure what the right feature combinations are. I think we probably want to test `abi3` and unlimited api builds on all platforms, unsure if that's sufficient.

We probably rerun doctests more times than we need to; I think some platforms are slow to build them and we can maybe manage with only testing them once per platform?

davidhewitt:
I also wonder if #2865 may help here by changing the feature sets needed.

adamreichold:
I would think that we should basically run `--features="full"` and `--features="abi3 full"` everywhere and I am not sure how #2865 would change that.

> We probably rerun doctests more times than we need to; I think some platforms are slow to build them and we can maybe manage with only testing them once per platform?

Since I very much like doctests doing double duty as "just tests" instead of being only "compiler-checked documentation", I would like to run them everywhere we run the unit tests.

davidhewitt:
I can agree with all of that. The only addition is that we should not `#[multiple_pymethods]` is a separate way of initialising the `#[pyclass]` machinery. That should at least only deviate by OS and not by python version, so we could run additional tests with that feature enabled only for latest Python version?

PR #3228, opened 2023-06-12
rename examples package to fix dependabot

davidhewitt:
At the moment dependabot fails to run on the repo with the following error:

```
Dependabot encountered the following error:

error: failed to load manifest for workspace member `/home/dependabot/dependabot-updater/dependabot_tmp_dir/examples`

Caused by:
  failed to parse manifest at `/home/dependabot/dependabot-updater/dependabot_tmp_dir/examples/Cargo.toml`

Caused by:
  the binary target name `examples` is forbidden, it conflicts with with cargo's build directory names
```

From testing locally using the dependabot CLI, changing the package name to `pyo3-examples` in `examples/Cargo.toml` appears to resolve.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "examples"
+name = "pyo3-examples"
 version = "0.0.0"
 publish = false
 edition = "2021"


davidhewitt:
Looks like the merge queue hit the timeout of 1 hour; I'll double it to two for now, though that's not a great long-term solution.

PR #3232, opened 2023-06-13
remove all functionality deprecated in PyO3 0.18

davidhewitt:
This turned out to be a very satisfying purge of complexity, mostly due to removing the backwards-compatibility for `#[pyfunction(a, b)]` and `#[args(a, b)]` which have been replaced by `#[pyo3(signature = (a, b))]`.


messense (approved):
Nice!

alex:
Looks like this also removes the usage of `NestedMeta`, which is reworked in syn 2.0, so it'll make that migration easier!

newsfragments/3232.removed.md
```diff
@@ -0,0 +1 @@
+Remove all functionality deprecated in PyO3 0.18.
```
adamreichold, line 1: Can we mention the `#[ærgs(..)]` syntax explicitly here as the headline item?

tests/ui/invalid_pyfunctions.rs
```diff
@@ -15,4 +15,7 @@ fn wildcard_argument(_: i32) {}
 #[pyfunction]
 fn destructured_argument((a, b): (i32, i32)) {}
 
+#[pyfunction]
+fn function_with_required_after_option(_opt: Option<i32>, _x: i32) {}
```
adamreichold, line 19: I think this would be good to mention in the changelog as well as it did cause quite a bit of confusion when we initially changed this IIRC. So I think at least some users will just have silenced the deprecation warning.

adamreichold (approved):
Just a suggestion to make the change log more explicit. LGTM otherwise.

davidhewitt:
I think what I'll do is split this out into separate PRs with separate changelog items, and merge them in turn given the approvals here.

davidhewitt:
Actually I've changed my mind; I'll just update the CHANGELOG entry and also add a small migration guide entry.

newsfragments/3232.removed.md
davidhewitt: Good idea, done.

tests/ui/invalid_pyfunctions.rs
davidhewitt: Mentioned in the migration guide.

PR #3239, opened 2023-06-13
Upgrade to syn 2

### LINKED ISSUE #3113: syn 2.0 and MSRV

alex:
fixes #3113

### DIFF
--- a/pyo3-macros-backend/src/attributes.rs
+++ b/pyo3-macros-backend/src/attributes.rs
@@ -146,18 +146,10 @@ pub type FromPyWithAttribute = KeywordAttribute<kw::from_py_with, LitStrValue<Ex
 pub type CrateAttribute = KeywordAttribute<Token![crate], LitStrValue<Path>>;
 
 pub fn get_pyo3_options<T: Parse>(attr: &syn::Attribute) -> Result<Option<Punctuated<T, Comma>>> {
-    if is_attribute_ident(attr, "pyo3") {
+    if attr.path().is_ident("pyo3") {
         attr.parse_args_with(Punctuated::parse_terminated).map(Some)
     } else {
         Ok(None)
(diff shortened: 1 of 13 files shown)


newsfragments/3239.changed.md
davidhewitt, line 1: To fit usual conventions for our changelog this would be `3239.packaging.md` please.

pyo3-macros-backend/src/attributes.rs
```diff
@@ -154,8 +154,8 @@ pub fn get_pyo3_options<T: Parse>(attr: &syn::Attribute) -> Result<Option<Punctu
 }
 
 pub fn is_attribute_ident(attr: &syn::Attribute, name: &str) -> bool {
```
davidhewitt, line 156: Looking at this, I think this function can just be removed at call sites and replaced with `attr.path().is_ident(name)`?

davidhewitt (commented):
Thanks, this is a much appreciated tidy-up! Just a couple of small suggestions.

alex:
Done!

PR #3243, opened 2023-06-15
Extend guide section on classes regarding free functions

### LINKED ISSUE #1059: Document passing PyClass values as arguments

adamreichold:
It might not be obvious from the reading the sections on free functions and on classes that they combine in a frictionless manner, i.e. class instances can be parameters to free functions in the same manner that the self parameters of instance methods are handled.

This also explicitly calls out the interaction between `Clone` and `FromPyObject` for classes.

Closes #1059

cc @facundo-villa

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -226,8 +226,10 @@ struct FrozenCounter {
     value: AtomicUsize,
 }
 
-let py_counter: Py<FrozenCounter>  = Python::with_gil(|py| {
-    let counter = FrozenCounter { value: AtomicUsize::new(0) };
+let py_counter: Py<FrozenCounter> = Python::with_gil(|py| {
+    let counter = FrozenCounter {
+        value: AtomicUsize::new(0),
+    };
 
     Py::new(py, counter).unwrap()
 });
@@ -647,9 +649,9 @@ impl BaseClass {
     #[new]
     #[classmethod]
     fn py_new<'p>(cls: &'p PyType, py: Python<'p>) -> PyResult<Self> {
-      // Get an abstract attribute (presumably) declared on a subclass of this class.
-      let subclass_attr = cls.getattr("a_class_attr")?;
-      Ok(Self(subclass_attr.to_object(py)))
+        // Get an abstract attribute (presumably) declared on a subclass of this class.
+        let subclass_attr = cls.getattr("a_class_attr")?;
+        Ok(Self(subclass_attr.to_object(py)))
     }
 }
 ```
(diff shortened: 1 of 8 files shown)


guide/src/class.md
```diff
@@ -716,6 +716,54 @@ impl MyClass {
...
+
+#[pyfunction]
+fn print_field(my_class: PyRef<'_, MyClass>) {
+    println!("{}", my_class.my_field);
+}
+
+#[pyfunction]
+fn print_refcnt(my_class: Py<MyClass>, py: Python<'_>) {
+    println!("{}", my_class.get_refcnt(py));
+}
```
davidhewitt, line 744: Potentially add a comment to each of these three forms to specificially point out the GIL-bound reference, reference wrapper etc.?

guide/src/class.md
```diff
@@ -716,6 +716,54 @@ impl MyClass {
...
+
+#[pyfunction]
+fn dissamble_clone(my_class: MyClass) {
+    let MyClass { mut my_field } = my_class;
+    *my_field += 1;
+}
+```
+
+Note that `#[derive(FromPyObject)]` on a class is usually not useful as it tries to construct a new Rust value by filling in the fields by looking up attributes of any given Python value.
```
davidhewitt, line 765: Crazy thought - we might be able to guard against this footgun by emitting some code which fails to compile if `T: !Clone + FromPyObject`, which users could then turn off with a `#[pyclass(external_impl_frompyobject)]` setting.

(Would probably require some great specialization evils in the implementation, my favourite kind of madness 😂)

davidhewitt (commented):
Thanks for adding this section! It's actually been on my mind for a long time, see #1059. I just never got around to writing anything 🙈. Do you think this PR closes that issue?

guide/src/class.md
adamreichold: I am not sure about the additional customization knob on `#[pyclass]`. In an ideal world, I would prefer to emit a hint when compilation fails due to a missing `FromPyObject` impl on a class, i.e. suggesting to derive `Clone` but not `FromPyObject`.

adamreichold:
> Do you think this PR closes that issue?

I added taking `&PyCell<T>` to the examples and think that with this addition, it does indeed.

guide/src/class.md
davidhewitt: Yes, I think further down the line this may be possible (e.g. maybe `#[diagnostic::on_unimplemented]`). For now let's not worry about this, unless we see lots of users suffer here.

guide/src/class.md
```diff
@@ -716,6 +718,67 @@ impl MyClass {
...
+}
+
+// Take a GIL-bound reference wrapper when borrowing should be automatic,
+// but interaction with the underlying `PyCell` is desired.
+#[pyfunction]
+fn print_field(my_class: PyRef<'_, MyClass>) {
+    println!("{}", my_class.my_field);
+}
+
+// Take a GIL-bound reference to the underyling cell
```
davidhewitt, line 746:
```suggestion
// Take a GIL-bound reference to the underlying cell
```

guide/src/class.md
```diff
@@ -716,6 +718,67 @@ impl MyClass {
 }
 ```
 
+## Free functions
```
davidhewitt, line 721: I wonder, should this be called something broader like "Accepting `#[pyclass]` values as function arguments"? (That's a bit of a meal to say, maybe there's a nicer way to phrase it.)

It might also be helpful to add this section to the "list of links" at the top of the page.

guide/src/python_from_rust.md
```diff
@@ -448,13 +465,14 @@ class House(object):
         match result {
             Ok(_) => {
                 let none = py.None();
-                house.call_method1("__exit__", (&none, &none, &none)).unwrap();
-            },
+                house
+                    .call_method1("__exit__", (&none, &none, &none))
+                    .unwrap();
```
davidhewitt, line 470: Did you run rustfmt on the guide code samples? I didn't know that was possible 😄

davidhewitt (approved):
Looks great! Just caught a typo and a quick thought on the section heading. Feel free to merge this once those are resolved 👍

guide/src/python_from_rust.md
adamreichold: We have a Nox session `format-guide` for that. It does require a nightly toolchain though. I think @birkenfeld added this.

PR #3245, opened 2023-06-15
examples: bump edition to 2021

davidhewitt:
Had noticed that a couple of examples still used 2018 edition (in their templates) after we did #3215, I suspect this was an innocent omission.


adamreichold:
Sorry for that, this is my grep skipping hidden files by default...

PR #3248, opened 2023-06-16
Add PyErr_GetRaisedException(), PyErr_SetRaisedException() to FFI

ijl:
These are new and part of the stable API as of v3.12.0a6.

PyErr_Fetch(), PyErr_Restore(), and PyErr_NormalizeException() were simultaneously deprecated in favor of these.

python/cpython@feec49c40736fc05626a183a8d14c4ebbea5ae28


davidhewitt (approved):
In which case I guess we merge this and we'll find time to adjust the 3.12 usages later. Thanks!

PR #3251, opened 2023-06-17
Fix two left-over usages of cfg(addr_of).

adamreichold:


alex (approved):

src/types/mod.rs
```diff
@@ -189,13 +189,6 @@ macro_rules! pyobject_native_type_info(
...
             #[inline]
             fn type_object_raw(_py: $crate::Python<'_>) -> *mut $crate::ffi::PyTypeObject {
-                // Create a very short lived mutable reference and directly
-                // cast it to a pointer: no mutable references can be aliasing
-                // because we hold the GIL.
-                #[cfg(not(addr_of))]
-                unsafe { &mut $typeobject }
-
-                #[cfg(addr_of)]
                 unsafe { ::std::ptr::addr_of_mut!($typeobject) }
```
davidhewitt, line 192: Do we even need `unsafe` around `addr_of_mut`? I wonder if that was purely there because we needed a block.

adamreichold: The `unsafe` is necessary as we still mutably access `static` variables and something also to call `assume_gil_acquired`.

adamreichold: (Personally, I find the call to `assume_gil_acquired` untidy, but we would have to rework the macros here to pass the GIL available here into the `$typeobject` expression which sometimes really is just a global and sometimes calls a function to lazily initialize a type object.)

davidhewitt: Agreed, refactoring here is welcome though maybe not my first priority. I also have a vague feeling that `#[pyclass]` types will need significant adjusting here if we ever try to remove the statics.

davidhewitt (approved):
Thanks!

PR #3252, opened 2023-06-18
Rework pyobject_native_type_info! to support callables

adamreichold:
Most native types have static type objects which just need to be turned into a pointer using addr_of_mut!, but sometimes like for exceptions we do call a function which lazily initializes a type object.

This change adds a modifier to the pyobject_native_type_*! macros to opt into this usage pattern and enable passing the GIL token to the callable in question avoiding the need to call Python::assume_gil_acquired.

The programmable declarative is a bit of a code smell itself IMHO, but this is the only way to make this a localized incremental change as far I can see. (Alternatively, the callable mode could be the only one with the `addr_of_mut!` invocations being mode to the macro call sites changing almost all of them.)


davidhewitt:
Thanks for cleaning this up :)

> Alternatively, the callable mode could be the only one with the addr_of_mut! invocations being mode to the macro call sites changing almost all of them.

I think going only for the callable mode may be a simpler end state? We can presumably have a helper macro to generate the callables.

E.g.

```rust
pyobject_native_type!(
    PyDict,
    ffi::PyDictObject,
    get_static_type!(ffi::PyDict_Type),
    #checkfunction=ffi::PyDict_Check
);
```

where `get_static_type!(ffi::PyDict_Type)` expands to 

```rust
|_py| unsafe { ::std::ptr::addr_of_mut!(ffi::PyDict_Type) }
```

which presumably inlines away trivially.

adamreichold:
> I think going only for the callable mode may be a simpler end state? We can presumably have a helper macro to generate the callables.

Reworked it along those lines but named the helper macro `pyobject_native_static_type_object!` but fit into the general scheme.

src/exceptions.rs
```diff
@@ -94,7 +94,7 @@ macro_rules! import_exception {
 
         $crate::pyobject_native_type_core!(
             $name,
-            *$name::type_object_raw($crate::Python::assume_gil_acquired()),
+            |py| $name::type_object_raw(py),
```
davidhewitt, line 97: Is the closure required here? It might just be possible to have

```rust
$name::type_object_raw,
```

src/exceptions.rs
```diff
@@ -233,7 +233,7 @@ macro_rules! create_exception_type_object {
     ($module: expr, $name: ident, $base: ty, $doc: expr) => {
         $crate::pyobject_native_type_core!(
             $name,
-            *$name::type_object_raw($crate::Python::assume_gil_acquired()),
+            |py| $name::type_object_raw(py),
```
davidhewitt, line 236: Same here maybe.

davidhewitt (approved):
Thanks, always nice to remove assumptions like the `assume_gil_acquired` calls there. Just one observation re possible unnecessary closures, feel free to merge this whether that observation is correct or not.

adamreichold:
> Just one observation re possible unnecessary closures, feel free to merge this whether that observation is correct or not.

Definitely correct in both cases, I think I messed this up in the reboot because this was actually the change I started with. 🤦🏽‍♀️

PR #3255, opened 2023-06-19
Replace unreachable! with empty enum match

mejrs:


adamreichold (approved):
Just one super-duper nit: I would be glad if we use `value` as the argument instead of `x` to be consistent with the other impls in the module.

mejrs:
Sure!

PR #3263, opened 2023-06-21
Add Python::with_pool as a safer alternative to Python::new_pool.

adamreichold:
I am still somewhat sceptical about adding this as an unsafe API on stable. Both because the unsafe of calling `with_pool` "infects" the closure body which is now necessarily unsafe as well and because I see the argument applying to `allow_threads` which would need to be unsafe stable for the same reasons.

Personally, I think this somewhat diminishes gains made from switching to the more limited API and would hence suspect users sticking to the existing API even though it is much more prone to misuse, i.e. at the end of day I expect more programmer mistakes than if we made the API safe and ate the soundness hole just like we do for `allow_threads`.


davidhewitt:
> Both because the unsafe of calling `with_pool` "infects" the closure body which is now necessarily unsafe as well and because I see the argument applying to `allow_threads` which would need to be unsafe stable for the same reasons.

I admit I'd missed the power of both of these points previously, particularly the insight on the tainting of the closure.

Given we're intending to remove this again in 0.21 which reduces my unsease of expanding API surfacre with a footgun, I think I should relax my position. Let's just make it safe on stable too. 👍

davidhewitt (approved):
👍

PR #3264, opened 2023-06-21
add http headers to help search engines index the docs

davidhewitt:
Quite often we have users linking to old versions of the guide.

I became aware today that we can set HTTP headers to encourage search engines to prefer our latest version of the docs. So here's an experiment in doing so.

### DIFF
--- a/.netlify/build.sh
+++ b/.netlify/build.sh
@@ -17,21 +17,45 @@ mv pyo3-gh-pages netlify_build
 ## Configure netlify _redirects file
 
 # Add redirect for each documented version
+set +x  # these loops get very spammy and fill the deploy log
+
 for d in netlify_build/v*; do
     version="${d/netlify_build\/v/}"
     echo "/v$version/doc/* https://docs.rs/pyo3/$version/:splat" >> netlify_build/_redirects
+    if [ $version != $PYO3_VERSION ]; then
+        # for old versions, mark the files in the latest version as the canonical URL
+        for file in $(find $d -type f); do
+            file_path="${file/$d\//}"
+            # remove index.html and/or .html suffix to match the page URL on the
+            # final netlfiy site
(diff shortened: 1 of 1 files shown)


davidhewitt:
On the deploy preview I see the Link header coming through as expected, so I think this is good to merge.

.netlify/build.sh
```diff
@@ -17,21 +17,45 @@ mv pyo3-gh-pages netlify_build
 ## Configure netlify _redirects file
 
 # Add redirect for each documented version
+set +x  # these loops get very spammy and fill the deploy log
+
 for d in netlify_build/v*; do
     version="${d/netlify_build\/v/}"
     echo "/v$version/doc/* https://docs.rs/pyo3/$version/:splat" >> netlify_build/_redirects
+    if [ $version != $PYO3_VERSION ]; then
```
adamreichold, line 25: What if this run for an unpublished version, shortly before releasing it, e.g `0.20` and `PYO3_VERSION == 0.19`. Will it create links to the older files? Do have a note for example in the releasing docs that one must publish to crates.io before running the Netlify build?

davidhewitt: This will be slightly racy for sure, however given we run the netlify build on every merge to main, the window where this might be wrong is relatively short.

The order of things written in `Releasing.md` implies pushing to crates.io before running the GitHub release.

We could make the crates.io push part of the release process in CI if we wanted to codify this?

adamreichold: > This will be slightly racy for sure, however given we run the netlify build on every merge to main, the window where this might be wrong is relatively short.

Yeah, I think this makes it moot. So this might publish a few confused links, but eventually things will stabilize. I think this is good enough, as search engines will only eventually re-index those page in any case.

adamreichold (approved):

Issue #3266, opened 2023-06-23
`IntoPy` does not properly respect `chrono_tz::Tz` timezones


grantslatton:
The [`IntoPy` implementation](https://github.com/PyO3/pyo3/blob/edf47b575ffe01d8357fa91bd471815a17f1a3a6/src/conversions/chrono.rs#L232C23-L232C23) for `chrono::DateTime` is implemented generically for any `T: TimeZone` by converting everything to fixed offsets. This means it loses the `Tz` information in a timezone like `America/Los Angeles` and just converts it to `UTC-8` which is a lossy conversion (due to daylight savings, etc).

PyO3 should probably provide two different implements, one for `DateTime<FixedOffset>` and another for `DateTime<chrono_tz::Tz>` that creates a python [`ZoneInfo`](https://docs.python.org/3/library/zoneinfo.html) object.

This is technically a bug since this should be a lossless conversion but I have filed it under feature request since it doesn't match the bug template very well.

adamreichold:
@Psykopear @pickfire Do you have the bandwidth to look into this?

grantslatton:
FWIW I wound up implementing this in my own code like:

```
pub fn get_canonical_tz_py_obj_kwargs(tz: Tz) -> Py<PyDict> {
    lazy_static! {
        static ref TIMEZONES: HashMap<Tz, Py<PyDict>> = {
            let code = r#"
import zoneinfo
all_zones = zoneinfo.available_timezones()
zone_dict = {}
for zone_name in all_zones:
    zone = zoneinfo.ZoneInfo(zone_name)
    zone_dict[zone_name] = zone
            "#.trim();
            Python::with_gil(|py| {
                let globals = PyDict::new(py);
                py.run(code, Some(globals), None).unwrap();
                let zone_dict = globals.get_item("zone_dict").unwrap().extract::<HashMap<String, Py<PyAny>>>().unwrap();
                zone_dict
                    .into_iter()
                    .map(|(mut name, zone)| {
                        // Weird idiosyncratic "factory reset" type zone that python has but rust does not
                        if name == "Factory" {
                            name = "UTC".to_string();
                        }
                        let kwargs = [("tzinfo", zone)].into_py_dict(py);
                        (Tz::from_str(&name).unwrap(), kwargs.into_py(py))
                    })
                    .collect()
            })
        };
    }

    TIMEZONES.get(&tz).unwrap().clone()
}
```

[...]

PR #3269, opened 2023-06-24
Fix fixed offset timezone conversion bug.

### LINKED ISSUE #3267: PyO3 incorrectly converts local to UTC times on FixedOffset conversion
### Bug Description

When converting to/from `datetime.datetime`/`DateTime<FixedOffset>`, PyO3 is using `DateTime::naive_utc` and `DateTime::from_utc` -- I believe these should be `naive_local` and `from_local`, respectively.

### Steps to Reproduce

```
    #[test]
    fn pyo3_tz_bug() {
        Python::with_gil(|py| {
            let code = r#"
import datetime
t = datetime.datetime.fromtimestamp(86400).replace(tzinfo=datetime.timezone(datetime.timedelta(hours=-8)))
            "#.trim();

            let globals = PyDict::new(py);
            py.run(code, Some(globals), None).unwrap();

            // Get ISO 8601 string from python
            let t = globals.get_item("t").unwrap();
            let py_iso_str = t.call_method0("isoformat").unwrap();

            // Get ISO 8601 string from rust
            let t = t.extract::<DateTime<FixedOffset>>().unwrap();
            let rust_iso_str = t.to_rfc3339();

            // They should be equal
            assert_eq!(py_iso_str.to_string(), rust_iso_str);
        })
    }
```

This test fails with

```
assertion failed: `(left == right)`
  left: `"1970-01-02T00:00:00-08:00"`,
 right: `"1970-01-01T16:00:00-08:00"`
 ```
 
 I think the bug probably went unnoticed because it exists on both the `IntoPy` and `FromPy` paths so cancels itself out when you go to and from, but you can induce it by creating dates in native python as shown above.

### Backtrace

_No response_

OS X Ventura 13.4

3.11

1.70

### Your PyO3 version

0.19

brew, no

### Additional Info

_No response_

### PR DESCRIPTION
**Pull request by grantslatton**
Closes #3267

### DIFF
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -216,8 +216,8 @@ impl FromPyObject<'_> for NaiveDateTime {
 
 impl<Tz: TimeZone> ToPyObject for DateTime<Tz> {
     fn to_object(&self, py: Python<'_>) -> PyObject {
-        let date = self.naive_utc().date();
-        let time = self.naive_utc().time();
+        let date = self.naive_local().date();
+        let time = self.naive_local().time();
         let yy = date.year();
         let mm = date.month() as u8;
         let dd = date.day() as u8;
@@ -246,7 +246,7 @@ impl<Tz: TimeZone> IntoPy<PyObject> for DateTime<Tz> {
 impl FromPyObject<'_> for DateTime<FixedOffset> {
     fn extract(ob: &PyAny) -> PyResult<DateTime<FixedOffset>> {
         let dt: &PyDateTime = ob.downcast()?;
-        let ms = dt.get_fold() as u32 * 1_000_000 + dt.get_microsecond();
+        let ms = dt.get_microsecond();
         let h = dt.get_hour().into();
         let m = dt.get_minute().into();
         let s = dt.get_second().into();
@@ -261,7 +261,8 @@ impl FromPyObject<'_> for DateTime<FixedOffset> {
             NaiveTime::from_hms_micro_opt(h, m, s, ms)
                 .ok_or_else(|| PyValueError::new_err("invalid or out-of-range time"))?,
         );
-        Ok(DateTime::from_utc(dt, tz))
+        // `FixedOffset` cannot have ambiguities so we don't have to worry about DST folds and such
+        Ok(DateTime::from_local(dt, tz))
     }
 }
 
@@ -607,7 +608,7 @@ mod tests {
                         .and_hms_micro_opt(hour, minute, ssecond, ms)
                         .unwrap();
                     let datetime =
-                        DateTime::<FixedOffset>::from_utc(datetime, offset).to_object(py);
+                        DateTime::<FixedOffset>::from_local(datetime, offset).to_object(py);
                     let datetime: &PyDateTime = datetime.extract(py).unwrap();
                     let py_tz = offset.to_object(py);
                     let py_tz = py_tz.downcast(py).unwrap();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/conversions/chrono.rs [hunk 1]
```diff
@@ -729,6 +730,30 @@ mod tests {
...
+    #[test]
+    fn test_pyo3_offset_fixed_frompyobject_created_in_python() {
+        Python::with_gil(|py| {
+            let code = r#"
+import datetime
+t = datetime.datetime.fromtimestamp(12345).replace(tzinfo=datetime.timezone(datetime.timedelta(hours=-8)))
+            "#.trim();
+
+            let globals = PyDict::new(py);
+            py.run(code, Some(globals), None).unwrap();
```
adamreichold, line 742: I think this might look simpler by using [`eval`](https://docs.rs/pyo3/latest/pyo3/marker/struct.Python.html#method.eval)?

src/conversions/chrono.rs
```diff
@@ -729,6 +730,30 @@ mod tests {
         })
     }
 
+    #[test]
+    fn test_pyo3_offset_fixed_frompyobject_created_in_python() {
+        Python::with_gil(|py| {
+            let code = r#"
+import datetime
+t = datetime.datetime.fromtimestamp(12345).replace(tzinfo=datetime.timezone(datetime.timedelta(hours=-8)))
```
adamreichold, line 738: Since the particular value of the timestamp should not matter, this looks like a good candidate for a proptest by interpolate the timestamp and the offset into the Python code?

src/conversions/chrono.rs
grantslatton: If you structure the code with `eval`, how do you get the import? Maybe I am misunderstanding

adamreichold: A useful trick is to use `__import__` as for example [here](https://github.com/PyO3/rust-numpy/blob/08510a3aa31a09c45df4602c484ac81056f0c4e8/src/array.rs#L742).

adamreichold:
Since we are going through a license change, please also have a loot at #3108

src/conversions/chrono.rs
adamreichold: Wouldn't it make sense to make the timezone offset variable for the proptest as well?

src/conversions/chrono.rs
grantslatton: I think you'll wind up with an expression that looks like

```
let code = format!("__import('datetime').datetime.fromtimestamp({timestamp}).replace(tzinfo=__import('datetime').timezone(__import('datetime').timedelta(hours=-8)))");
```

Note the 3 uses of `__import__` to make this work, IMO the `run` solution is more readable

adamreichold: How about

```rust
let globals = [("datetime", py.import("datetime").unwrap())].into_py_dict(py);
let t = py.eval(format!("datetime.datetime.fromtimestamp({timestamp}).replace(tzinfo=datetime.timezone(datetime.timedelta(hours={hours})))"), Some(globals), None).unwrap();
```

then?

adamreichold: (Also note that you cannot use inline format literals due to our MSRV.)

adamreichold:
Not sure what the Windows build failures are about. Is the Python installation missing the time zone database there?

grantslatton:
> Not sure what the Windows build failures are about. Is the Python installation missing the time zone database there?

It's because Windows only supports times in 1970-2038 I believe, but the proptest is using a much larger range, I am fixing this now

grantslatton:
Hunting down some more bugs the more exhaustive proptest found. Found one in chrono: https://github.com/chronotope/chrono/issues/1154

There is another one that results in the python vs rust datetimes being 1 second off when the time falls exactly on the daylight savings switchover. Unsure if pyo3 conversion bug (with the `fold` parameter) or just a discrepancy between python and chrono. Investigating further...

pitdicker:
@grantslatton I am interested in the issues you may run into.

grantslatton:
> @grantslatton I am interested in the issues you may run into.

Does chrono have facilities for supporting ambiguous instants created at the end of daylight savings? i.e. https://peps.python.org/pep-0495/

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
adamreichold, line 1: This is IDE-specific and does not belong into version control. Please also squash your commits as we generally aim for a helpful commit history and do not squash automatically.

proptest-regressions/conversions/chrono.txt
```diff
@@ -0,0 +1,9 @@
+# Seeds for failure cases proptest has generated in the past. It is
```
adamreichold, line 1: I would like to avoid checking this in for now as well. AFAIK this file is also not stable w.r.t. dependency and toolchain updates and we have no procedure for managing it.

grantslatton:
I am wondering if we should also fix https://github.com/PyO3/pyo3/issues/3266 in the same change, because this change will result in `fold` information being lost on `FromPyObject`, because PyO3's doesn't have a conversion out to `DateTime<Tz>` -- only `DateTime<Utc>` and `DateTime<FixedOffset`. 

But doing this would require changing the `IntoPy for DateTime<Tz: TimeZone>` path to different impls: one for `DateTime<Utc>`, `DateTime<FixedOffset>`, `DateTime<Local>`, and `DateTime<Tz>`. This would be a breaking change that would break any downstream code that has defined their own custom `TimeZone` impl. There is probably no such code, as everyone is probably using timezones defined by `chrono` and `chrono_tz`.

I think such a change increases the correctness of this crate, because right now the conversion is lossy and loses daylight savings information present in `chrono_tz::Tz` locale timezones.

adamreichold:
> This would be a breaking change that would break any downstream code that has defined their own custom TimeZone impl. There is probably no such code, as everyone is probably using timezones defined by chrono and chrono_tz.

This would mean the fix would also be limited to 0.20, the next release where breakage is possible. So I think the question is whether the side effects of losing the fold information is serious enough that one would say the backwards-compatible version of the fix cannot be used anyway?

Personally, I would be fine with try to fix this properly for 0.20 only because the chrono integration is still a relatively new feature of PyO3 and some churn seems expected.

adamreichold:
P.S.: You seem to have included two unrelated commits during your rebase.

Psykopear:
Ok, sorry for the delay. I've reviewed the issue and the PR, the bug is there, and this fixes it, nice catch.

> I am wondering if we should also fix https://github.com/PyO3/pyo3/issues/3266 in the same change, because this change will result in fold information being lost on FromPyObject, because PyO3's doesn't have a conversion out to DateTime\<Tz> -- only DateTime\<Utc> and DateTime\<FixedOffset>.

The problem with the `chrono_tz` integration IIRC, is that python<3.9 doesn't have the `zoneinfo` package in the standard library, and I'm not sure it's reasonable to implement the integration without that. So, unless there is an alternative solution I missed (that doesn't require adding a python dependency like [backports.zoneinfo](https://pypi.org/project/backports.zoneinfo/) and/or tzdata), we'd probably need to put the `chrono_tz` integration under a separate feature (dependent on python>=3.9), and keep only the `chrono` one for python<3.9. The two features would also have to be mutually exclusive I think, because we can't implement the python traits for both `chrono::Tz` (the trait) and `chrono_tz::Tz` (the enum).

All that to say that it's probably ok to lose the `fold` info with just `chrono` (since we only handle fixedoffsets), and fix this for python>=3.9 in the `chrono_tz` integration.
Gating the new integration under a separate feature will also avoid breaking code that uses the current one, because you'd need to enable the `chrono_tz` feature.

adamreichold (approved):
@grantslatton @Psykopear Ok, you to summarize my understanding: We take as it is and it fixes #3267 but #3266 will need a more involved approach which might even use an independent code path. Do you both agree with this?

grantslatton:
> @grantslatton @Psykopear Ok, you to summarize my understanding: We take as it is and it fixes #3267 but #3266 will need a more involved approach which might even use an independent code path. Do you both agree with this?

Yes this is right

PR #3270, opened 2023-06-25
add `PyDict::get_item_with_error` for PyPy

davidhewitt:
Looks like PyPy implemented `PyDict_GetItemWithError` [a couple of years ago in 7.2.0](https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/doc/release-v7.2.0.rst)

While correcting I couldn't resist adjusting our wrapper to avoid checking the global exception state on a successful return, and also to use `PyErr::take` to only access the global exception state once on error.

### DIFF
--- a/src/types/dict.rs
+++ b/src/types/dict.rs
@@ -2,10 +2,9 @@ use super::PyMapping;
 use crate::err::{self, PyErr, PyResult};
 use crate::ffi::Py_ssize_t;
 use crate::types::{PyAny, PyList};
-use crate::{ffi, AsPyPointer, Python, ToPyObject};
 #[cfg(not(PyPy))]
-use crate::{IntoPyPointer, PyObject};
-use std::ptr::NonNull;
+use crate::IntoPyPointer;
+use crate::{ffi, AsPyPointer, PyObject, Python, ToPyObject};
 
 /// Represents a Python `dict`.
 #[repr(transparent)]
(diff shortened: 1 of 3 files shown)


src/types/dict.rs
```diff
@@ -157,19 +157,17 @@ impl PyDict {
...
     ///
     /// To get a `KeyError` for non-existing keys, use `PyAny::get_item_with_error`.
-    #[cfg(not(PyPy))]
     pub fn get_item_with_error<K>(&self, key: K) -> PyResult<Option<&PyAny>>
     where
         K: ToPyObject,
     {
         unsafe {
             let ptr =
                 ffi::PyDict_GetItemWithError(self.as_ptr(), key.to_object(self.py()).as_ptr());
```
adamreichold, line 166: Admittedly unrelated to change in question, but I was wondering whether this line becomes a potential use-after-free when we drop the pool? If the owning pointer returned by `key.to_object(self.py())` is dropped while the result of `.as_ptr()` is passed to `PyDict_GetItemWithError`? Is there some sort of lifetime extension for the temporary in play? Probably not as the raw point does not carry a lifetime?

src/types/dict.rs
```diff
@@ -157,19 +157,17 @@ impl PyDict {
...
-            if !ffi::PyErr_Occurred().is_null() {
-                return Err(PyErr::fetch(self.py()));
+            match NonNull::new(ptr) {
+                Some(p) => Ok(Some(self.py().from_owned_ptr(ffi::_Py_NewRef(p.as_ptr())))),
+                None => PyErr::take(self.py()).map(Err).transpose(),
```
adamreichold, line 169: This code is very dense. Maybe it would help to at least take `let py = self.py();`? Do we have some kind of helper method for `from_owned_ptr(_Py_NewRef(ptr))`? Isn't this `from_borrowed

src/types/dict.rs
davidhewitt: Good question, I believe it is sound due to the temporaries all being held until the end of the statement.

Actually this is against the style which @mejrs and I agreed in https://github.com/PyO3/pyo3/pull/2312#discussion_r860233477, so let me adjust this to bind the temporary on a new line, just to be sure.

davidhewitt:
Thanks for the great feedback - I've split this out into two commits. The second commit rebuilds both `get_item` and `get_item_with_error` to use a non-generic inner, avoid the pointer-to-temporary, and use `PyObject::from_borrowed_ptr_or_opt`.

src/types/dict.rs
adamreichold: Debloating the generics is certainly a nice flourish.

src/types/dict.rs
davidhewitt: `py.from_borrowed_ptr` doesn't actually increase the reference count, it just casts to the `&PyAny`. Pretty easy to misuse tbh, so I'll be glad to see that removed with the pool.

On the other hand, `Py::from_borrowed_ptr` and equivalents _do_ convert the borrowed ptr into an owned one, because `Py` is an owning type. So I'll switch to use that.

adamreichold: Ok, so as we are going all-out code golf on this one, I think combinators and a single transpose would work as well, i.e.

```rust
 PyObject::from_borrowed_ptr_or_opt(py, ptr)
                .map(|pyobject| Ok(pyobject.into_ref(py)))
                .or_else(|| PyErr::take(py).map(Err))
                .transpose()
```

davidhewitt: Great spot - I had a feeling the final `.transpose()` was possible but missed what to do with the `Some` variant.

adamreichold (approved):

PR #3271, opened 2023-06-25
stop suppressing unrelated exceptions in `PyAny::hasattr`

davidhewitt:
I realised today that our `hasattr` implementation suppresses all exceptions, which is inconsistent with `hasattr`, which only suppresses `AttributeError`.

```python
>>> class Foo:
...     def __getattribute__(self, *args):
...         raise ValueError("oh no")
...
>>> hasattr(Foo(), "bar")
  File "<stdin>", line 3, in __getattribute__
ValueError: oh no
```

See also https://github.com/capi-workgroup/problems/issues/51

TODO: Before merging, I should push a test for this.

### DIFF
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -1,7 +1,7 @@
 use crate::class::basic::CompareOp;
 use crate::conversion::{AsPyPointer, FromPyObject, IntoPy, IntoPyPointer, PyTryFrom, ToPyObject};
 use crate::err::{PyDowncastError, PyErr, PyResult};
-use crate::exceptions::PyTypeError;
+use crate::exceptions::{PyAttributeError, PyTypeError};
 use crate::type_object::PyTypeInfo;
 #[cfg(not(PyPy))]
 use crate::types::PySuper;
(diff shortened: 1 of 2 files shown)


src/types/any.rs
```diff
@@ -83,10 +83,24 @@ impl PyAny {
...
+    fn hasattr_impl(&self, attr_name: Py<PyString>) -> PyResult<bool> {
+        let py = self.py();
+        let result = unsafe {
+            PyObject::from_owned_ptr_or_err(
+                py,
+                ffi::PyObject_GetItem(self.as_ptr(), attr_name.as_ptr()),
+            )
+        };
+        match result {
+            Ok(_) => Ok(true),
```
adamreichold, line 100: This does imply an unnecessary increment-then-decrement of the reference count, doesn't it? Also why `GetItem` instead of `GetAttr` as suggested by [the docs](https://docs.python.org/3/c-api/object.html#c.PyObject_HasAttr)?

If we do eat the reference count modifications, couldn't this just call `getattr(name)` and match on the error?

davidhewitt: > This does imply an unnecessary increment-then-decrement of the reference count, doesn't it?

Sadly yes, though this was always the case, just abstracted away in `PyObject_HasAttr`.

> Also why `GetItem` instead of `GetAttr` as suggested by [the docs](https://docs.python.org/3/c-api/object.html#c.PyObject_HasAttr)?

A bug, should be `GetAttr`, this is exactly why I need to push a test 😄

davidhewitt: Changed the implementation to go via `getattr` as suggested (and pushed tests).

adamreichold: Just to make sure I understand it correctly: Calling `_getattr` instead of `getattr` avoids the `into_ref(py)` which involves the pool? So when we drop the pool and `getattr` returns `PyAny<'py>` this distinction can be removed?

adamreichold (approved):

src/types/any.rs
davidhewitt: Yes that's exactly right. Avoiding the pool means that we don't end up holding a reference to whatever we checked for in `hasattr` in the pool - so it's just a micro-optimization for now that should reduce away soon.

PR #3273, opened 2023-06-25
add some style guide to Contributing.md

davidhewitt:
Born out of the discussion in https://github.com/PyO3/pyo3/pull/3270#discussion_r1241229424

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -111,6 +111,36 @@ To include your changes in the release notes, you should create one (or more) ne
 - `removed` - for features which have been removed
 - `fixed` - for "changed" features which were classed as a bugfix
 
+### Style guide
+
+#### Generic code
+
+PyO3 has a lot of generic APIs to increase usability. These can come at the cost of generic code bloat. Where reasonable, try to implement a concrete sub-portion of generic functions. There are two forms of this:
+
+- If the concrete sub-portion doesn't benefit from re-use by other functions, name it `inner` and keep it as a local to the function.
+- If the concrete sub-portion is re-used by other functions, preferably name it `_foo` and place it directly below `foo` in the source code (where `foo` is the original generic function).
+
+#### FFI calls
+
+PyO3 makes a lot of FFI calls to Python's C API using raw pointers. Where possible try to avoid using pointers-to-temporaries in expressions:
+
+```rust
+// dangerous
+pyo3::ffi::Something(name.to_object(py).as_ptr());
+
+// because the following refactoring is a use-after-free error:
+let name = name.to_object(py).as_ptr();
+pyo3::ffi::Something(name)
+```
+
+Instead, prefer to bind the safe owned `PyObject` wrapper before passing to ffi functions:
+
+```rust
+let name: PyObject = name.to_object(py);
+pyo3::ffi::Something(name.as_ptr())
+// name will automatically be freed when it falls out of scope
+```
+
 ## Python and Rust version support policy
 
 PyO3 aims to keep sufficient compatibility to make packaging Python extensions built with PyO3 feasible on most common package managers.
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -9,7 +9,6 @@ use crate::{AsPyPointer, IntoPy, IntoPyPointer, Py, PyAny, PyObject, Python, ToP
 use std::borrow::Cow;
 use std::cell::UnsafeCell;
 use std::ffi::CString;
-use std::os::raw::c_int;
 
 mod err_state;
 mod impls;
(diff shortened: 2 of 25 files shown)


Contributing.md
```diff
@@ -111,6 +111,33 @@ To include your changes in the release notes, you should create one (or more) ne
 - `removed` - for features which have been removed
 - `fixed` - for "changed" features which were classed as a bugfix
 
+### Style guide
+
+#### Generic code
+
+PyO3 has a lot of generic APIs to increase usability. These can come at the cost of generic code bloat. Where reasonable, try to implement a concrete sub-portion of generic functions. See `get_item` and `get_item_impl` in `dict.rs` as examples of this pattern.
```
davidhewitt, line 118: Across the code I think we've now used `_impl` and `_inner` suffixes, and also just `_` prefix.

I suggest we pick one of the three and I'll push a commit to update the codebase to match. I propose the `_` prefix, because it's short and implies some level of privacy.

adamreichold: Hhhmmm, personally, I actually prefer the

```
fn foobar<...>(...) {
  fn inner/impl(...) {
    ...
  }

  inner/impl(..)
}
```

where the privacy is realized by tucking away the non-generic inner function within the outer function.

I am not sure if `_foobar` as the name of the inner function is helpful?

Of course, this style also does not work if the non-generic implementation is used in multiple outer functions.

davidhewitt: I agree actually, though I'd proposed `_` prefix thinking that combined with outer functions can lead to smallest diffs if the implementation needs to be reused later.

Given that we both prefer `fn inner`, I will adjust this to state preference for that, and suggest the `_` prefix when implementation needs to be shared.

davidhewitt:
Updated and added a commit which applies these two rules to a number of our FFI calls.

src/err/mod.rs
```diff
@@ -439,9 +439,8 @@ impl PyErr {
     where
         T: ToPyObject,
     {
-        unsafe {
-            ffi::PyErr_GivenExceptionMatches(self.type_ptr(py), exc.to_object(py).as_ptr()) != 0
-        }
+        let exc = exc.to_object(py);
+        unsafe { ffi::PyErr_GivenExceptionMatches(self.type_ptr(py), exc.as_ptr()) != 0 }
```
adamreichold, line 443: Candidate for an non-generic inner impl or too short to matter?

src/types/sequence.rs
```diff
@@ -193,13 +189,16 @@ impl PySequence {
...
         V: ToPyObject,
     {
-        let r =
-            unsafe { ffi::PySequence_Count(self.as_ptr(), value.to_object(self.py()).as_ptr()) };
-        if r == -1 {
-            Err(PyErr::fetch(self.py()))
-        } else {
-            Ok(r as usize)
+        fn inner(seq: &PySequence, value: PyObject) -> PyResult<usize> {
+            let r = unsafe { ffi::PySequence_Count(seq.as_ptr(), value.as_ptr()) };
```
adamreichold, line 193: This isn't `error_on_minusone(..).map(|r| r as usize)`?

src/types/sequence.rs
```diff
@@ -227,13 +229,16 @@ impl PySequence {
...
     {
-        let r =
-            unsafe { ffi::PySequence_Index(self.as_ptr(), value.to_object(self.py()).as_ptr()) };
-        if r == -1 {
-            Err(PyErr::fetch(self.py()))
-        } else {
-            Ok(r as usize)
+        fn inner(seq: &PySequence, value: PyObject) -> PyResult<usize> {
+            let r = unsafe { ffi::PySequence_Index(seq.as_ptr(), value.as_ptr()) };
+            if r == -1 {
```
adamreichold, line 234: `error_on_minusone` as above?

adamreichold:
Wow, this is a lot of generic code to support `impl ToPyObject` arguments. Certainly not for this PR, but I do wonder whether `&dyn ToPyObject` might even be a good trade-off for these kinds of API?

src/err/mod.rs
davidhewitt: Done 👍

src/types/sequence.rs
davidhewitt: `error_on_minusone` doesn't return a value, we could change it, but that's a future PR I think.

src/types/sequence.rs
davidhewitt: Done 👍

davidhewitt:
Pushed a few commits to apply the suggestions, and also move `unsafe { }` blocks inside of `error_on_minusone` calls as reducing scope of the unsafe seemed like a good idea.

> Wow, this is a lot of generic code to support `impl ToPyObject` arguments. Certainly not for this PR, but I do wonder whether `&dyn ToPyObject` might even be a good trade-off for these kinds of API?

Personally I think this is probably the right level, if users want to have only a single monomorphization I think they can select the `&dyn ToPyObject` instantiation directly?

adamreichold:
> Personally I think this is probably the right level, if users want to have only a single monomorphization I think they can select the &dyn ToPyObject instantiation directly?

Agreed, disciplined users can opt into this by always calling with `&dyn ToPyObject`.

adamreichold (approved):

PR #3277, opened 2023-06-26
Extended range of supported versions of `indexmap` optional dependency to include version 2.

adamreichold:


davidhewitt (approved):
If it builds, I'm happy! 😄

PR #3278, opened 2023-06-26
decouple pyo3-ffi-check from MSRV

davidhewitt:
Inspired by #3276, I figured we should perhaps just exclude pyo3-ffi-check from the workspace during MSRV jobs. It's only an internal tool, we already don't run it on MSRV, and hopefully dependabot version bumps for its deps can then be trivially merged.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,10 +136,10 @@ jobs:
               - '.github/workflows/build.yml'
 
       - name: Run pyo3-ffi-check
-        run: nox -s ffi-check
         # pypy 3.7 and 3.8 are not PEP 3123 compliant so fail checks here, nor
         # is pypy 3.9 on windows
         if: ${{ steps.ffi-changes.outputs.changed == 'true' && inputs.rust == 'stable' && inputs.python-version != 'pypy3.7' && inputs.python-version != 'pypy3.8' && !(inputs.python-version == 'pypy3.9' && contains(inputs.os, 'windows')) }}
+        run: nox -s ffi-check
 
 
       - name: Test cross compilation
--- a/noxfile.py
+++ b/noxfile.py
@@ -34,7 +34,7 @@ def test_rust(session: nox.Session):
 
     _run_cargo_test(session)
     _run_cargo_test(session, features="abi3")
-    if not "skip-full" in session.posargs:
+    if "skip-full" not in session.posargs:
         _run_cargo_test(session, features="full")
         _run_cargo_test(session, features="abi3 full")
 
@@ -76,7 +76,8 @@ def fmt(session: nox.Session):
 
 @nox.session(name="fmt-rust", venv_backend="none")
 def fmt_rust(session: nox.Session):
-    _run(session, "cargo", "fmt", "--all", "--check", external=True)
+    _run_cargo(session, "fmt", "--all", "--check")
+    _run_cargo(session, "fmt", *_FFI_CHECK, "--all", "--check")
 
 
 @nox.session(name="fmt-py")
@@ -103,9 +104,6 @@ def _clippy(session: nox.Session, *, env: Dict[str, str] = None) -> bool:
                 *feature_set,
                 "--all-targets",
                 "--workspace",
-                # linting pyo3-ffi-check requires docs to have been built or
-                # the macros will error; doesn't seem worth it on CI
-                "--exclude=pyo3-ffi-check",
                 "--",
                 "--deny=warnings",
                 external=True,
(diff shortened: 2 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -57,8 +57,11 @@ jobs:
         run: echo LD_LIBRARY_PATH=${pythonLocation}/lib >> $GITHUB_ENV
 
       - if: inputs.msrv == 'MSRV'
-        name: Prepare minimal package versions (MSRV only)
-        run: nox -s set-minimal-package-versions
+        name: Prepare package versions for MSRV
+        run: |
+          sed -i '/pyo3-ffi-check/d' Cargo.toml  # we allow pyo3-ffi-check to use newer packages, as an internal tool
+          git diff Cargo.toml
+          nox -s set-minimal-package-versions
```
alex, line 64: Can we just run the various commands with `--workspace --exclude pyo3-ffi-check`?

.github/workflows/build.yml
```diff
@@ -57,8 +57,11 @@ jobs:
         run: echo LD_LIBRARY_PATH=${pythonLocation}/lib >> $GITHUB_ENV
 
       - if: inputs.msrv == 'MSRV'
-        name: Prepare minimal package versions (MSRV only)
-        run: nox -s set-minimal-package-versions
+        name: Prepare package versions for MSRV
+        run: |
+          sed -i '/pyo3-ffi-check/d' Cargo.toml  # we allow pyo3-ffi-check to use newer packages, as an internal tool
```
adamreichold, line 62: I think the Noxfile already mentions `pyo3-ffi-check` by name, excluding it from linting.

So I see two alternatives to this "open-coded" approach:
* We move this change into the `set-minimal-package-versions` session to concentrate this in one place.
* We fully remove pyo3-ffi-check from the workspace, e.g. by adding an empty `[workspace]` to its `Cargo.toml`, and build it only when needed.

Personally, I would prefer the second option as the cleaner one.

.github/workflows/ci.yml
```diff
@@ -51,6 +51,11 @@ jobs:
           key: check-msrv-1.56.0
         continue-on-error: true
       - run: python -m pip install --upgrade pip && pip install nox
+      - name: Prepare package versions for MSRV
+        run: |
+          sed -i '/pyo3-ffi-check/d' Cargo.toml  # we allow pyo3-ffi-check to use newer packages, as an internal tool
+          git diff Cargo.toml
+          nox -s set-minimal-package-versions
       - name: Prepare minimal package versions
         run: nox -s set-minimal-package-versions
```
adamreichold, line 60: This is now called twice by the CI job.

davidhewitt: Thanks, fixed.

.github/workflows/build.yml
davidhewitt: That's probably cleaner; I've force-pushed to reimplement as the second option.

noxfile.py
```diff
@@ -534,8 +526,12 @@ def load_pkg_versions():
 
 @nox.session(name="ffi-check")
 def ffi_check(session: nox.Session):
-    session.run("cargo", "doc", "-p", "pyo3-ffi", "--no-deps", external=True)
-    _run(session, "cargo", "run", "-p", "pyo3-ffi-check", external=True)
+    path = ["--manifest-path", "pyo3-ffi-check/Cargo.toml"]
+    session.run("cargo", "doc", *path, "-p", "pyo3-ffi", "--no-deps", external=True)
+    _run(
+        session, "cargo", "clippy", "--workspace", "--all-targets", *path, external=True
```
adamreichold, line 532: This one I do not understand. This is to make sure the ffi-check binary is Clippy clean? But why the `--workspace` flag then?

davidhewitt: Yes, to check the ffi-check binary, and then `--workspace` is to also check `pyo3-ffi-check/macro`. I've force-pushed to add "macro" subdirectory to the workspace section in `pyo3-ffi-check/Cargo.toml`.

Also, good point about rustfmt - I've added that to the `fmt-rust` job.

davidhewitt:
I've pushed a second commit with a tiny tidy-up to the `noxfile` with a `_run_cargo` helper function.

noxfile.py
```diff
@@ -139,18 +133,8 @@ def _check(env: Dict[str, str]) -> None:
...
-                    *feature_set,
-                    "--all-targets",
-                    "--workspace",
-                    # linting pyo3-ffi-check requires docs to have been built or
-                    # the macros will error; doesn't seem worth it on CI
-                    "--exclude=pyo3-ffi-check",
-                    external=True,
-                    env=env,
+                _run_cargo(
+                    session, *feature_set, "--all-targets", "--workspace", env=env
```
alex, line 137: Missing "check"

adamreichold (approved):

noxfile.py
davidhewitt: Thanks, the classic editing-code-last-thing-at-night mistake - at least we have CI to check me! 😄

PR #3280, opened 2023-06-28
clarify ownership of opt_to_pyobj helper function

davidhewitt:
While looking at tidy ups to apply in #3273 I noticed that this function has a strange `py.None().as_ptr()`. This is technically dangling but also always sound in practice due to the singleton nature of `None`.

I changed it to clarify that this function returns a borrowed reference in both cases, and used the ffi `PyNone()` to get a borrowed reference directly here.

If nothing else this is a micro-optimization.

### DIFF
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -257,7 +257,7 @@ impl PyDateTime {
                 c_int::from(minute),
                 c_int::from(second),
                 microsecond as c_int,
-                opt_to_pyobj(py, tzinfo),
+                opt_to_pyobj(tzinfo),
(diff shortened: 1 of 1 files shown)


adamreichold:
> If nothing else this is a micro-optimization.

The safe alternative here would be `py.None().into_ref(py)` which will stop to work without the pool? I.e. dropping the pool would imply that we have to use this approach anyway?

If so, this is r+ by me.

davidhewitt:
Yes, with the addition that the `Some` branch has no ownership / reference count changes so it seems correct to not change any reference counts in the `None` form.

PR #3283, opened 2023-06-28
RFC: deny `clippy::pedantic` by default

davidhewitt:
This PR reconfigures the clippy configuration to enable `clippy::pedantic` and then explicitly opt-out of the lints which we don't currently pass.

In principle I believe as a foundational library we should try to lint ourselves as strictly as possible to guarantee highest code quality. Of course where lint rules are extremely expensive to support we should keep ourselves realistic.

By enabling `clippy::pedantic` like this I hope for a few outcomes:
- In Rust version bumps we'll pick up more new lints and decide whether to fix or allow them at the time
- We are more explicit about what we're _not_ currently linting for
- Interested contributors can submit PRs to fix compliance with and enable these rules one at a time, which may be a gentle starting point for newcomers (I'd be tempted to open and pin a "Good First Issue" to ask to fix these, similar to FFI cleanup we had in the past)


adamreichold:
As a general comment, I have to admit to being torn on this. Personally, I am for draconic linting as a way to automatically keep quality above a certain level.

But I do understand the arguments that increasing linting levels in general complicates the story for new contributors who may struggle to get their intended functional change completed under the conditions. But even for more experienced contributors, I fear that we opt into additional busy work with relatively limited benefits due to this

I also have no problems with a certain level of busy work, sometimes I am actually happy to do such soothing mechanical changes. But in the case of PyO3 right now, I think we are struggling with feature work due to time constraints as it is and I am unsure whether we should voluntarily make choices which funnel further time away for the main work items.

mejrs:
It already happens that PRs fail because of new lints triggering on code that the PR didn't touch. I'd prefer the other way around; if a pedantic lint makes sense, add it.

davidhewitt:
I agree that this would create us a certain amount of additional busywork. My one main motivator for doing this would be that several times I've found that downstream code which enforces stricter linting will hit lint failures originating from inside PyO3's macros which have to then be silenced. #3319 is the most recent one I've hit.

In practice it's not too bad - when I see these I fix them here (typically with a CHANGELOG entry as it's user-facing). Just a bit annoying for users to have to suffer that.

Will close this for now and we can revist.

PR #3287, opened 2023-07-02
Allow `#[new]` to return existing instances

### LINKED ISSUE #2384: Allow `#[new]` to return existing objects
I don't think it's possible for `#[new]` in PyO3 to return an existing object (at the moment), though it really should be.

_Originally posted by @davidhewitt in https://github.com/PyO3/pyo3/discussions/2382#discussioncomment-2765288_

alex:
fixes #2384

### DIFF
--- a/newsfragments/3287.added.md
+++ b/newsfragments/3287.added.md
@@ -1,0 +1,1 @@
+`#[new]` methods may now return `Py<Self>` in order to return existing instances
--- a/src/pyclass_init.rs
+++ b/src/pyclass_init.rs
@@ -1,7 +1,7 @@
 //! Contains initialization utilities for `#[pyclass]`.
 use crate::callback::IntoPyCallbackOutput;
 use crate::impl_::pyclass::{PyClassBaseType, PyClassDict, PyClassThreadChecker, PyClassWeakRef};
-use crate::{ffi, PyCell, PyClass, PyErr, PyResult, Python};
+use crate::{ffi, IntoPyPointer, Py, PyCell, PyClass, PyErr, PyResult, Python};
 use crate::{
     ffi::PyTypeObject,
     pycell::{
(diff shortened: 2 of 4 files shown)


src/pyclass_init.rs
```diff
@@ -284,6 +294,16 @@ where
     }
 }
 
+impl<T: PyClass> From<Py<T>> for PyClassInitializer<T>
```
adamreichold, line 297: The `T: PyClass` appears to be repeated in both short and long form.

src/pyclass_init.rs
```diff
@@ -242,13 +247,18 @@ impl<T: PyClass> PyObjectInit<T> for PyClassInitializer<T> {
             contents: MaybeUninit<PyCellContents<T>>,
         }
 
-        let obj = self.super_init.into_new_object(py, subtype)?;
+        let (init, super_init) = match self.0 {
+            PyClassInitializerImpl::Existing(value) => return Ok(value.clone_ref(py).into_ptr()),
```
adamreichold, line 251: Do we need to clone if we consume `self` and it contains an owning reference?

newsfragments/3287.added.md
```diff
@@ -0,0 +1 @@
+`#[new]` methods may now return `Py<Self>` in order to return existing instances
```
adamreichold, line 1: I think this change requires adjustments to the guide as well.

alex: Good point, if the code (or at least, the public API) looks good I can update the guide, wanted to hold off on there until we knew how the API/implementation would impact what needed to be documented.

alex: Ok, added an explanation. Let me know if you think it needs more.

davidhewitt (approved):
LGTM, thanks! While it's a subset of the full functionality I was thinking in #3287, it will be forwards-compatible with extending `#[new]` to allow any `T: IntoPy<PyObject>` so am happy to merge this as it unblocks users.

PR #3294, opened 2023-07-04
add #[pymodule] mod some_module { ... }, v2

birkenfeld:
Implements using `#[pymodule]` directly on modules.

Allows importing other wrapped items with a `#[pyo3] use foo;`.

Allows having a `#[pymodule_init]` function to run arbitrary code after adding all wrapped items.

This is #2367 brought up to date with main.


birkenfeld:
Unfortunately now `test_declarative_module` fails because the non-declarative submodule is initialized twice. This assertion probably hadn't been there when you made the original PR?

src/impl_/pymodule.rs
```diff
@@ -82,6 +82,10 @@ impl ModuleDef {
         (self.initializer.0)(py, module.as_ref(py))?;
         Ok(module)
     }
+
+    pub fn add_to_module(&'static self, module: &PyModule) -> PyResult<()> {
+        module.add_object(self.make_module(module.py())?.into())
+    }
```
davidhewitt, line 88: I think to avoid the reinitialization we could use a hypothetical `self.find_or_make_module()` here which uses `PyState_FindModule` to retrieve the existing module object (if it exists).

I think this exists on all CPython versions, for PyPy it only exists on PyPy 3.9 (with the wrong name) and PyPy 3.10. Probably it'll be good enough to just reinitialize on PyPy 3.8 and older, they're out of support anyway.

I'll send a PR now to fix the PyPy definitions for `PyState_FindModule`.

pyo3-macros-backend/src/module.rs
```diff
@@ -56,9 +57,132 @@ impl PyModuleOptions {
...
+                    )?;
+                }
+            }
+            syn::Item::Fn(item_fn) => {
+                let mut is_module_init = false;
+                item_fn.attrs.retain(|attr| {
+                    let found = attr.path().is_ident("pymodule_init");
+                    is_module_init |= found;
+                    !found
+                });
```
davidhewitt, line 144: I think this is where we can scan for `#[pyclass]`, `#[pyfunction]` and `#[pymodule]`.

If the implementation of how it would be cleanest to add these items to the module isn't readily available I'm happy to help figure this out.

pyo3-macros-backend/src/module.rs
```diff
@@ -56,9 +57,132 @@ impl PyModuleOptions {
...
+            }
+            syn::UseTree::Path(path) => extract_use_items(&path.tree, cfg_attrs, names, attrs)?,
+            syn::UseTree::Group(group) => {
+                for tree in &group.items {
+                    extract_use_items(tree, cfg_attrs, names, attrs)?
+                }
+            }
+            syn::UseTree::Glob(glob) => {
+                bail_spanned!(glob.span() => "#[pyo3] cannot import glob statements")
+            }
```
davidhewitt, line 103: It would be nice to add a UI test for declarative modules which checks this error.

pyo3-macros-backend/src/module.rs
```diff
@@ -56,9 +57,132 @@ impl PyModuleOptions {
...
+            syn::Item::Fn(item_fn) => {
+                let mut is_module_init = false;
+                item_fn.attrs.retain(|attr| {
+                    let found = attr.path().is_ident("pymodule_init");
+                    is_module_init |= found;
+                    !found
+                });
+                if is_module_init {
+                    ensure_spanned!(pymodule_init.is_none(), item_fn.span() => "only one pymodule_init may be specified");
```
davidhewitt, line 146: Similarly a UI test for this error here would be desirable.

pyo3-macros-backend/src/module.rs
```diff
@@ -56,9 +57,132 @@ impl PyModuleOptions {
...
+        vis,
+        ident,
+        mod_token,
+        content,
+        unsafety: _,
+        semi: _,
+    } = &mut module;
+    let items = match content {
+        Some((_, items)) => items,
+        None => bail_spanned!(module.span() => "`#[pymodule]` can only be used on inline modules"),
```
davidhewitt, line 72: UI test also makes sense for this one, I think.

davidhewitt (commented):
Brilliant! Thank you for picking this up 🙏 

I've put a comment below with a suggestion how we can avoid the test failure.

What do we need to finish here to make this useful? I think as a starting point it's good enough to merge if we can add submodules, classes and free functions with `#[pymodule]`, `#[pyclass]` and `#[pyfunction]` items inside of the module. (Add them automatically instead of needing to add them in the `#[module_init]`.)

We can add those to the test of `test_declarative_module`.

As follow ups we can deprecate `#[pyfn]` shorthand of functional modules, and maybe even completely deprecate functional modules (as `#[module_init]` should allow all the same functionality to be replicated).

holzschu:
Following the mention in #3382, I hope this is a better place for suggesting that it would be great if PyO3 modules could have `m_free` and `m_clear` functions, that can be called when we want to manually clear up the module from memory. 
Ideally, these functions would leave the module in a state where it can be reinitialized again.

My motivation is that I'm working with non-standard architectures, where the process continues (so memory is not automatically released) but the interpreter is restarted, so I manually free all modules when the interpreter ends and reinitialize them (I'll be the first to admit that's a niche concern).

davidhewitt:
Superseded by #3815

PR #3295, opened 2023-07-04
add PyState_*Module definitions for PyPy

davidhewitt:
Might be useful for #3294 

https://github.com/PyO3/python3-dll-a/blob/main/src/libpypy3.9-c.def#L781
https://github.com/PyO3/python3-dll-a/blob/main/src/libpypy3.10-c.def#L585


adamreichold (approved):

adamreichold:
The 

```
use crate::moduleobject::PyModuleDef;
```

at the top is still gated as `#[cfg(not(PyPy))]`.

PR #3302, opened 2023-07-08
deprecate PySizedLayout

davidhewitt:
I noticed that with the long-term existence of `PyClassBaseType`, there was a section in the guide about `PySizedLayout` which wasn't helpful, as the compiler errors refer to `PyClassBaseType` (if you try to insert from a class which doesn't implement `PyClassBaseType`).

`PySizedLayout` is not actually necessary so I removed it from the constraints in the `PyCell` machinery and updated the guide. I've deprecated the trait so we can get rid of it as a cleanup later.

An interesting effect of this is that a few traits now wanted `T: ?Sized` bounds. I think there was no harm in adding this bounds, all of these traits are implementation details in my opinion.

(Just not placed inside `impl_` modules; should really consider doing that. Maybe I should add that as a separate commit?)

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -340,8 +340,8 @@ impl SubSubClass {
 # });
 ```
 
-You can also inherit native types such as `PyDict`, if they implement
-[`PySizedLayout`]({{#PYO3_DOCS_URL}}/pyo3/type_object/trait.PySizedLayout.html).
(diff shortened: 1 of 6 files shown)


davidhewitt:
Actually thinking on this a bit further, I'd like to do this in a more structured way where we clean up these namespaces ahead of 1.0 sometime in the future.

PR #3303, opened 2023-07-08
RFC: Add implementation of Iterator::size_hint for PyIterator

adamreichold:
When the Python iterator backing `PyIterator` has a `__length_hint__` special method, we can use this as a lower bound for Rust's `Iterator::size_hint` to e.g. support pre-allocation of collections.

There are a few complications though:
* The cost of calling `lookup_special` and extracting the length hint might be too high compared to what typical users of `size_hint` expect.
* Python's `operator.length_hint` will first try `__len__` before falling back to `__length_hint__` which would be more robust but worsen the first issue.

Additionally, Python's and Rust's semantics do not really match up as the
single estimate given by `__length_hint__` might be wrong in both directions. To
support the use case of pre-allocating collections, it is used as the lower
bound and correctness should never depend on it being correct, but it might lead
to unnecessary over-allocations.

### DIFF
--- a/benches/bench_any.rs
+++ b/benches/bench_any.rs
@@ -5,7 +5,7 @@ use pyo3::{
         PyBool, PyByteArray, PyBytes, PyDict, PyFloat, PyFrozenSet, PyInt, PyList, PyMapping,
         PySequence, PySet, PyString, PyTuple,
     },
-    PyAny, Python,
+    PyAny, PyResult, Python,
 };
 
 #[derive(PartialEq, Eq, Debug)]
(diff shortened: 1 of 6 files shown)


davidhewitt:
It seems nice to me to do something along these lines for sure. Have you benchmarked to see if adding this improves performance for e.g. `PyIterator` -> `Vec<&PyAny>`?

Some possible alternatives:
 - Using `PyObject_LengthHint` may be faster (but not supported on `abi3`). That does try `len()` first.
 - We could potentially just use `self.len()`, as that should be guaranteed correct if defined (for both lower and upper bounds) and will be fast because we can go through the C API.

adamreichold:
> It seems nice to me to do something along these lines for sure. Have you benchmarked to see if adding this improves performance for e.g. PyIterator -> Vec<&PyAny>?

Not yet, will add a benchmark.

> Using PyObject_LengthHint may be faster (but not supported on abi3). That does try len() first.

I did look for FFI bindings, but I did not remember to camel case `length_hint`. 🤦🏽 Since `Iterator::size_hint` is just that a hint, I think it would be best to implement this when the full API is used, and to leave it out otherwise.

> We could potentially just use self.len(), as that should be guaranteed correct if defined (for both lower and upper bounds) and should be fast because we can go through the C API.

The few iterators like `iter([1, 2, 3])` I looked at did not define `__len__`. In any case, if we do the above then `__len__` will by checked first by `PyObject_LengthHint`.

I also don't think the upper bound is too critical as e.g. `impl FromIterator<T> for Vec<T>` will use only the lower bound for pre-allocation, c.f. [here](https://doc.rust-lang.org/stable/src/alloc/vec/spec_from_iter_nested.rs.html#29)

adamreichold:
Rebooted this into a pure optimization if the stable ABI is not used and added a benchmark. Benchmarking memory allocation seems hard and I get a very noisy signal with at most single digit percentages improvement. But since this optional implementation seems straight forward, I'd say this definitely rather helps than hurts. Nothing worthing of a changelog entry though...

davidhewitt (approved):
Yep I'm in favour of this approach 👍

PR #3304, opened 2023-07-09
Start adding a performance section to the guide.

adamreichold:
We discussed this quite often and I though that even if it is only a single thing, just making a start would improve our chances of getting it finished eventually.

### DIFF
--- a/guide/src/performance.md
+++ b/guide/src/performance.md
@@ -52,3 +52,43 @@ fn frobnicate(value: &PyAny) -> PyResult<&PyAny> {
     }
 }
 ```
+
+## Access to GIL-bound reference implies access to GIL token
+
+Calling `Python::with_gil` is effectively a no-op when the GIL is already held, but checking that this is the case still has a cost. If an existing GIL token can not be accessed, for example when implementing a pre-existing trait, but a GIL-bound reference is available, this cost can be avoided by exploiting that access to GIL-bound reference gives zero-cost access to a GIL token via `PyAny::py`.
+
+For example, instead of writing
+
+```rust
+# #![allow(dead_code)]
+# use pyo3::prelude::*;
+# use pyo3::types::PyList;
+
+struct Foo(Py<PyList>);
+
+struct FooRef<'a>(&'a PyList);
+
+impl PartialEq<Foo> for FooRef<'_> {
+    fn eq(&self, other: &Foo) -> bool {
+        Python::with_gil(|py| self.0.len() == other.0.as_ref(py).len())
+    }
+}
+```
+
+use more efficient
+
+```rust
+# #![allow(dead_code)]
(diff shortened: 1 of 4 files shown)


guide/src/performance.md
```diff
@@ -0,0 +1,54 @@
+# Performance
+
+To achieve the best possible performance, it is useful to be aware of several tricks and sharp edges concerning PyO3's API.
+
+## `extract` versus `downcast`
+
+Pythonic API implemented using PyO3 are often explicitly polymorphic, i.e. they will accept `&PyAny` and try to turn this into multiple more concrete types to which the requested operation is applied. This often leads to chains of calls to `extract`, e.g.
```
davidhewitt, line 7:
```suggestion
Pythonic API implemented using PyO3 are often polymorphic, i.e. they will accept `&PyAny` and try to turn this into multiple more concrete types to which the requested operation is applied. This often leads to chains of calls to `extract`, e.g.
```

davidhewitt (approved):
Nice, thanks for kicking this off!

One day it would be nice to remove this performance pitfall. Something to investigate after removing the pool 😄

adamreichold:
@davidhewitt I added another subsection on implicit access to GIL token. Please have another look before this is merged.

Also do you remember any other topics when wanted to discuss here? I do think I remember that there about four different things we wanted to have here, but I am unable find a list anywhere in the older issues, pull requests or discussions.

adamreichold:
> Also do you remember any other topics when wanted to discuss here?

Is dictionary dispatch, i.e. look-up tables based on type object identity, something we want to discuss here or is this too specialized?

davidhewitt:
Personally I think that's quite application-specific, but I'm not against having that section if you want to write it.

Some other topics I recalled over the day:
- Overhead of conversions (i.e. `PyList` vs `Vec` etc.) 
- Threading / allow_threads
- String `intern!`
- `Vec<u8>` becoming `list[int]` (and thus being very slow)
- `#[pyo3(get)]` deep-cloning non-Py data

We can potentially move these into an issue rather than write them all in this PR?

adamreichold:
> Threading / allow_threads

Shouldn't this be covered by the "parallelism" section?

> We can potentially move these into an issue rather than write them all in this PR?

Yes, I'd like to limit this PR to the two small sections already written.

> Personally I think that's quite application-specific, but I'm not against having that section if you want to write it.

Ok, let's include a note in the issue. Certainly not in this PR.

davidhewitt:
Sorry for the slow reply - I got a bit sidetracked by the various PRs I have pushed in the last 24h...

> Shouldn't this be covered by the "parallelism" section?

Yes, though I think it can't hurt to call it out i.e. "how do I multithread? See parallelism section". More visibility :)

Issue #3305, opened 2023-07-10
test failure with CPython 3.12b3: err::tests::fetching_panic_exception_resumes_unwind


decathorpe:
I'm trying to update the pyo3 packages on Fedora Linux to the latest version (v0.19.1), but I'm getting a weird test failure with CPython 3.12b3:

```
---- err::tests::fetching_panic_exception_resumes_unwind stdout ----
--- PyO3 is resuming a panic after fetching a PanicException from Python. ---
Python stack trace below:
note: panic did not contain expected string
      panic message: `"Unwrapped panic from Python code"`,
 expected substring: `"new panic"`

failures:
    err::tests::fetching_panic_exception_resumes_unwind
```

c.f. https://github.com/PyO3/pyo3/blob/v0.19.1/src/err/mod.rs#L855-L868

I'm not sure how or why this is happening, but it looks like PanicException message is no longer extracted successfully here:
https://github.com/PyO3/pyo3/blob/v0.19.1/src/err/mod.rs#L317-L321

davidhewitt:
Yes, we haven't got 3.12 passing in CI yet, I think realistically this will come in PyO3 0.20 which might be a month off. Is it helpful for you if we try to expedite and backport onto a 0.19.2 release?

decathorpe:
That would be very helpful, yes :+1: Let me know if I can help with something.

Note that I also need to support PyO3 v0.15, v0.16, v0.17, and v0.18 on top of Python 3.12 (mostly due to Python packages like `cryptography` using ancient PyO3 version to be compatible with debian stable or something) ... I'll likely need to do some patching to make that work, would you be interested in getting fixes out for those older branches as well?

davidhewitt:
Oof. I'll take a quick spin at fixing the 3.12 test suite now. (Might take a few days depending on how much is broken.)

TBH I think it will be a lot of work to backport proper releases for all those versions, as they will have multiple ways that their CI has bitrotted by now. I don't think I'll have capacity to do that, but if there's any information I can offer to help you with patching I will do my best to share it.

`cryptography` have updated their PyO3 dependency to 0.19 not too long ago, so you may be able to update that first to avoid some patching pain. I'd be interested in hearing what other packages require old PyO3 versions, potentially it would be better for the ecosystem if we collaborate to update those packages' PyO3 dependencies?

decathorpe:
[...]

PR #3306, opened 2023-07-10
fix exception handling on Python 3.12

davidhewitt:
Reworked exception handling for Python 3.12. Ref #3305 

The simplifications to `PyErrState` result in a ~10-15% perf improvement in the `bench_err` benchmarks on all Python versions. On 3.12 the enum size has been reduced from 32 to 24 bytes, which might have small global perf benefits too.

While working on this I noticed a bug on <3.12 where the original panic message is not retrieved properly in `PyErr::fetch` if the `PanicException` is normalised. I've added that as a second commit.

### DIFF
--- a/src/err/err_state.rs
+++ b/src/err/err_state.rs
@@ -7,9 +7,35 @@ use crate::{
 
 #[derive(Clone)]
 pub(crate) struct PyErrStateNormalized {
+    #[cfg(not(Py_3_12))]
     pub ptype: Py<PyType>,
     pub pvalue: Py<PyBaseException>,
+    #[cfg(not(Py_3_12))]
     pub ptraceback: Option<Py<PyTraceback>>,
+}
+
+impl PyErrStateNormalized {
+    #[cfg(not(Py_3_12))]
+    pub(crate) fn ptype<'py>(&'py self, py: Python<'py>) -> &'py PyType {
+        self.ptype.as_ref(py)
+    }
+
+    #[cfg(Py_3_12)]
+    pub(crate) fn ptype<'py>(&'py self, py: Python<'py>) -> &'py PyType {
+        self.pvalue.as_ref(py).get_type()
+    }
+
+    #[cfg(not(Py_3_12))]
+    pub(crate) fn ptraceback<'py>(&'py self, py: Python<'py>) -> Option<&'py PyTraceback> {
+        self.ptraceback
+            .as_ref()
+            .map(|traceback| traceback.as_ref(py))
+    }
+
+    #[cfg(Py_3_12)]
+    pub(crate) fn ptraceback<'py>(&'py self, py: Python<'py>) -> Option<&'py PyTraceback> {
+        unsafe { py.from_owned_ptr_or_opt(ffi::PyException_GetTraceback(self.pvalue.as_ptr())) }
+    }
 }
 
 pub(crate) struct PyErrStateLazyFnOutput {
@@ -22,6 +48,7 @@ pub(crate) type PyErrStateLazyFn =
 
 pub(crate) enum PyErrState {
     Lazy(Box<PyErrStateLazyFn>),
+    #[cfg(not(Py_3_12))]
     FfiTuple {
         ptype: PyObject,
         pvalue: Option<PyObject>,
@@ -53,9 +80,23 @@ impl PyErrState {
             pvalue: args.arguments(py),
         }))
     }
-}
 
-impl PyErrState {
+    pub(crate) fn normalized(pvalue: &PyBaseException) -> Self {
+        Self::Normalized(PyErrStateNormalized {
+            #[cfg(not(Py_3_12))]
+            ptype: pvalue.get_type().into(),
+            pvalue: pvalue.into(),
+            #[cfg(not(Py_3_12))]
+            ptraceback: unsafe {
+                Py::from_owned_ptr_or_opt(
+                    pvalue.py(),
+                    ffi::PyException_GetTraceback(pvalue.as_ptr()),
+                )
+            },
+        })
+    }
+
+    #[cfg(not(Py_3_12))]
     pub(crate) fn into_ffi_tuple(
         self,
         py: Python<'_>,
@@ -64,7 +105,11 @@ impl PyErrState {
             PyErrState::Lazy(lazy) => {
                 let PyErrStateLazyFnOutput { ptype, pvalue } = lazy(py);
                 if unsafe { ffi::PyExceptionClass_Check(ptype.as_ptr()) } == 0 {
-                    Self::exceptions_must_derive_from_base_exception(py).into_ffi_tuple(py)
+                    PyErrState::lazy(
+                        PyTypeError::type_object(py),
+                        "exceptions must derive from BaseException",
+                    )
+                    .into_ffi_tuple(py)
                 } else {
                     (ptype.into_ptr(), pvalue.into_ptr(), std::ptr::null_mut())
                 }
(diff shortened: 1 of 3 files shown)


src/err/mod.rs
```diff
@@ -181,27 +181,32 @@ impl PyErr {
...
-            return exceptions_must_derive_from_base_exception(obj.py());
+            return exceptions_must_derive_from_base_exception(py);
         };
 
         PyErr::from_state(state)
     }
 
+    #[cfg(not(Py_3_12))]
+    fn _from_value(obj: &PyAny) -> PyErr {}
+
```
alex, line 209: Looks like this is leftovers?

davidhewitt:
Pushed a more complete implementation which I'm happy with, and updated OP.

src/err/mod.rs
```diff
@@ -844,12 +833,31 @@ mod tests {
...
+        Python::with_gil(|py| {
+            let err: PyErr = PyErr::new::<crate::types::PyString, _>(());
+            assert!(err.is_instance_of::<exceptions::PySystemError>(py));
+            err.restore(py);
+            let err = PyErr::fetch(py);
+
+            assert!(err.is_instance_of::<exceptions::PySystemError>(py));
+            assert_eq!(
+                err.to_string(),
+                "SystemError: _PyErr_SetObject: exception <class 'str'> is not a BaseException subclass"
```
davidhewitt, line 847: While this is a different error message to the non-3.12 case, I think it's actually a better error message. If the difference is undesirable I can either:
- Update the 3.12 case to check and return our exception, or
- Make normalization on non-3.12 cases _also_ use `PyErr_SetObject`, which would change the exception message on all versions. (Maybe I should benchmark this, and consider splitting out into a separate PR as it would be breaking.)

For reference, the assertion in the non-3.12 equivalent is:

```rust
            assert_eq!(
                 err.to_string(),
                "TypeError: exceptions must derive from BaseException"
            );
```

src/err/mod.rs
```diff
@@ -439,15 +455,16 @@ impl PyErr {
...
+                != 0
         }
         inner(self, py, exc.to_object(py))
     }
 
     /// Returns true if the current exception is instance of `T`.
     #[inline]
     pub fn is_instance(&self, py: Python<'_>, ty: &PyAny) -> bool {
-        (unsafe { ffi::PyErr_GivenExceptionMatches(self.type_ptr(py), ty.as_ptr()) }) != 0
+        (unsafe { ffi::PyErr_GivenExceptionMatches(self.get_type(py).as_ptr(), ty.as_ptr()) }) != 0
```
davidhewitt, line 467: Changing these lines is probably a bug fix, because otherwise it uses the unnormalized exception type pointer for instance checks, and `self.value(py)` or `self.get_type(py)` may then return completely unrelated stuff if normalization fails.

I'll split this out into a separate PR for communication clarity.

src/exceptions.rs
```diff
@@ -359,7 +359,9 @@ Python::with_gil(|py| {
 impl_native_exception!(
     PyBaseException,
     PyExc_BaseException,
-    native_doc!("BaseException")
+    native_doc!("BaseException"),
+    ffi::PyBaseExceptionObject,
+    #checkfunction=ffi::PyExceptionInstance_Check
```
davidhewitt, line 364: This is just an optimization to use a "fast" check for `isinstance` for `BaseException`. I'll split this out into a separate PR too.

davidhewitt:
I think best if I circle back to this later to split out into multiple PRs for the changes of behaviour / fixes etc.

src/err/err_state.rs
```diff
@@ -1,26 +1,49 @@
...
+        self.pvalue.as_ref(py).get_type()
+    }
+
+    pub(crate) fn ptraceback<'py>(&'py self, py: Python<'py>) -> Option<&'py PyTraceback> {
+        unsafe { py.from_owned_ptr_or_opt(ffi::PyException_GetTraceback(self.pvalue.as_ptr())) }
+    }
+}
+
+pub(crate) type PyErrStateLazyFn =
+    dyn for<'py> FnOnce(Python<'py>) -> (PyObject, PyObject) + Send + Sync;
```
mejrs, line 42: What is this `(PyObject, PyObject)`? Can this have a comment, or maybe could it return a struct instead?

davidhewitt: Completely agreed, I've changed it to a `struct` in a second commit.

src/err/mod.rs
davidhewitt: For now I've just kept it the same in 3.12 as before, we could reconsider changing this another time.

davidhewitt:
Assuming CI is green, I believe this is now good to review.

davidhewitt:
On further reflection, I should be able to split this out to first make the `PyErrState::LazyValue` & `PyErrState::LazyTypeAndValue` -> `PyErrState::Lazy` change a separate PR, if that would help with reviewing.

adamreichold:
> if that would help with reviewing.

I am pretty swamped at the moment, so anything that turns this into smaller chunks would be appreciated.

davidhewitt:
> > if that would help with reviewing.
> 
> I am pretty swamped at the moment, so anything that turns this into smaller chunks would be appreciated.

👍 #3323

davidhewitt:
Ok, I think this is as small a diff as it can possibly get. I split off the second commit into #3326.

src/err/mod.rs
```diff
@@ -322,6 +321,31 @@ impl PyErr {
         }))
     }
 
+    #[cfg(Py_3_12)]
+    fn _take(py: Python<'_>) -> Option<PyErr> {
+        let pvalue = unsafe { Py::from_owned_ptr_or_opt(py, ffi::PyErr_GetRaisedException()) }?;
+        let state = PyErrStateNormalized { pvalue };
+        if state.ptype(py).as_ptr() == PanicException::type_object_raw(py).cast() {
+            let msg: String = state.pvalue.as_ref(py).to_string();
```
adamreichold, line 329: I think the same nit w.r.t. `.str().ok()` instead of calling into `Display` applies here.

src/err/mod.rs
```diff
@@ -265,6 +264,11 @@ impl PyErr {
     /// expected to have been set, for example from [`PyErr::occurred`] or by an error return value
     /// from a C FFI function, use [`PyErr::fetch`].
     pub fn take(py: Python<'_>) -> Option<PyErr> {
+        Self::_take(py)
+    }
+
+    #[cfg(not(Py_3_12))]
+    fn _take(py: Python<'_>) -> Option<PyErr> {
```
adamreichold, line 271: Just to make sure I understand this correctly. The `_take` layer is only used to avoid repeating the doc comment of the public function for `Py_3_12` and `not(Py_3_12)`?

src/err/mod.rs
```diff
@@ -178,8 +178,10 @@ impl PyErr {
     pub fn from_value(obj: &PyAny) -> PyErr {
         let state = if let Ok(obj) = obj.downcast::<PyBaseException>() {
             PyErrState::Normalized(PyErrStateNormalized {
```
adamreichold, line 180: I think it would be nice to have a `PyErrState::normalized` constructor to better consolidate the `#[cfg(Py_3_12)]` attributes in the `err_state` module.

src/err/mod.rs
```diff
@@ -443,15 +467,12 @@ impl PyErr {
     /// This is the opposite of `PyErr::fetch()`.
     #[inline]
     pub fn restore(self, py: Python<'_>) {
-        let state = match self.state.into_inner() {
-            Some(state) => state,
+        match self.state.into_inner() {
+            Some(state) => state.restore(py),
             // Safety: restore takes `self` by value so nothing else is accessing this err
             // and the invariant is that state is always defined except during make_normalized
             None => unsafe { std::hint::unreachable_unchecked() },
```
adamreichold, line 474: I know it is unrelated but the safety comment seems out of date as there is no `make_normalized` function any more? Asked more broadly, is the `unreachable_unchecked` performance critical here or would the simpler `expect("PyErr state should never be invalid outside of normalization")` suffice as well?

src/err/mod.rs
davidhewitt: Yes, I figured it was more DRY this way.

src/err/mod.rs
davidhewitt: Benchmarks indicated no difference with the `unreachable_unchecked`, so let's go with expect. 👍

adamreichold (approved):

PR #3311, opened 2023-07-11
improve error span for mutable access to `#[pyclass(frozen)]`

davidhewitt:
For the following cases:

```rust
#[pyclass(frozen)]
pub struct Foo {
    #[pyo3(set)]  // Error `set` on frozen class
    field: u32,
}

#[pymethods]
impl Foo {
    fn mut_method(&mut self) {}  // Error `&mut` receiver on frozen class
}
```

... the error was fairly unhelpful trait solver output which pointed at `#[pyclass]` or `#[pymethods]` attribute.

With this PR, the error is much better in both cases - `#[pyo3(set)]` gets rejected as an option on frozen classes, and the `#[pymethods]` case gives a hint about `PyClass<Frozen = False>` directly on the `&mut self`.

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -8,7 +8,7 @@ use quote::ToTokens;
 use quote::{quote, quote_spanned};
 use syn::ext::IdentExt;
 use syn::spanned::Spanned;
-use syn::Result;
+use syn::{Result, Token};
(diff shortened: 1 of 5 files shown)


adamreichold (approved):

adamreichold:
Nice that the extraction logic is not open-coded any more.

PR #3312, opened 2023-07-11
optimize is_instance for PyBaseException

davidhewitt:
Split off from #3306 

`PyExceptionInstance_Check` is an efficient way to check for exceptions, just checks a bit on the type flags. Worth making use of it, it should be equivalent in behaviour.

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -181,12 +181,10 @@ impl PyErr {
     /// });
     /// ```
     pub fn from_value(obj: &PyAny) -> PyErr {
-        let ptr = obj.as_ptr();
-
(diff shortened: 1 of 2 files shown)


davidhewitt:
If there are no concerns here, I'll merge it tonight or tomorrow so that I can rebase #3306

adamreichold:
> If there are no concerns here, I'll merge it tonight or tomorrow so that I can rebase #3306

Sorry, did not get a notification for this one. (I now have a mail thread with the first starting with `Re:` in my inbox.) Will have a look at it tonight.

adamreichold (approved):

PR #3313, opened 2023-07-11
normalize exception in `PyErr::matches` and `PyErr::is_instance`

davidhewitt:
Split off from #3306 

This fixes a case where `PyErr::matches` and `PyErr::is_instance` could return results inconsistent with `PyErr::get_type` due to the first two methods not ensuring the exception was normalised.

### DIFF
--- a/newsfragments/3313.fixed.md
+++ b/newsfragments/3313.fixed.md
@@ -1,0 +1,1 @@
+Fix case where `PyErr::matches` and `PyErr::is_instance` returned results inconsistent with `PyErr::get_type`.
--- a/src/err/err_state.rs
+++ b/src/err/err_state.rs
@@ -68,11 +68,17 @@ impl PyErrState {
                     )
                 }
             }
-            PyErrState::LazyValue { ptype, pvalue } => (
-                ptype.into_ptr(),
-                pvalue(py).into_ptr(),
-                std::ptr::null_mut(),
-            ),
+            PyErrState::LazyValue { ptype, pvalue } => {
+                if unsafe { ffi::PyExceptionClass_Check(ptype.as_ptr()) } == 0 {
+                    Self::exceptions_must_derive_from_base_exception(py).into_ffi_tuple(py)
+                } else {
+                    (
+                        ptype.into_ptr(),
+                        pvalue(py).into_ptr(),
+                        std::ptr::null_mut(),
+                    )
+                }
+            }
             PyErrState::FfiTuple {
                 ptype,
                 pvalue,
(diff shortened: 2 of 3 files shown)


src/err/err_state.rs
```diff
@@ -68,11 +68,17 @@ impl PyErrState {
...
             }
-            PyErrState::LazyValue { ptype, pvalue } => (
-                ptype.into_ptr(),
-                pvalue(py).into_ptr(),
-                std::ptr::null_mut(),
-            ),
+            PyErrState::LazyValue { ptype, pvalue } => {
+                if unsafe { ffi::PyExceptionClass_Check(ptype.as_ptr()) } == 0 {
+                    Self::exceptions_must_derive_from_base_exception(py).into_ffi_tuple(py)
+                } else {
```
davidhewitt, line 74: I moved this check from `PyErr::from_type` to here. This allowed me to observe the case of normalization failure easier in the test (by delaying this check until normalisation). It's also now more consistent with the match arm above, which also does this check.

newsfragments/3313.fixed.md
```diff
@@ -0,0 +1 @@
+Fix case where `PyErr::matches` and `PyErr::is_instance` returned inconsistent results with `PyErr::get_type()`.
```
adamreichold, line 1: I am bit confused:

I think in your cover letter (PR description), you mean `PyErr::matches` and `PyErr::is_instance` like here?

But also think you want `returned results inconsistent with` like in your cover letter, not `returned inconsistent results with` which might be read as "the results of `PyErr::matches` are different when `PyErr::get_type` is also used"?

src/err/mod.rs
```diff
@@ -138,10 +138,6 @@ impl PyErr {
     where
         A: PyErrArguments + Send + Sync + 'static,
     {
-        if unsafe { ffi::PyExceptionClass_Check(ty.as_ptr()) } == 0 {
```
adamreichold, line 141: The doc comment still seems to refer to normalization even though this does not happen any more? (That sentence is also really hard to read as it is.)

davidhewitt: The normalisation is still done later (where I moved this check to).

However it's an outdated term which is deprecated in python 3.12, so I'll update the doc 👍

newsfragments/3313.fixed.md
davidhewitt: Yes it's `matches` and `is_instance` which I am claiming had dodgy behaviour before this PR. I'll fixup, good catch.

davidhewitt:
I've corrected the header and changelog entry, and pushed a separate commit to action the removal of "normalization" as a term from the PyErr docs.

adamreichold (approved):

PR #3316, opened 2023-07-14
add the support of `fixed` and `fpdec` to the conversions

alexcwyu:
this is to add the support of `fixed` and `fpdec` to the conversions

https://docs.rs/fixed/latest/fixed/
https://docs.rs/fpdec/latest/fpdec/

(similary to `rust_decimal`)

### DIFF
--- a/src/conversions/fixed.rs
+++ b/src/conversions/fixed.rs
@@ -12,11 +12,6 @@
 #![doc = concat!("pyo3 = { version = \"", env!("CARGO_PKG_VERSION"),  "\", features = [\"fixed\"] }")]
 //! fixed = "1.23.1"
 //! ```
-//!
-//! Note that you must use a compatible version of fixed and PyO3.
-//! The required fixed version may vary based on the version of PyO3.
-//!
-//! ```
 
 use crate::exceptions::PyValueError;
 use crate::once_cell::GILOnceCell;
--- a/src/conversions/fixed.rs
+++ b/src/conversions/fixed.rs
@@ -1087,3 +1087,92 @@ fixed_conversion!(U2F126);
 fixed_conversion!(U1F127);
 /// [`FixedU128`] with no integer bits and 128 fractional bits.
 fixed_conversion!(U0F128);
+
+#[cfg(test)]
+mod test_fixed {
+    use super::*;
+    use crate::err::PyErr;
+    use crate::types::PyDict;
+    use fixed::types::*;
+
+    #[cfg(not(target_arch = "wasm32"))]
+    use proptest::prelude::*;
+
+    macro_rules! convert_constants {
+        ($name:ident, $rs:expr, $py:literal) => {
+            #[test]
+            fn $name() {
+                Python::with_gil(|py| {
+                    let rs_orig = $rs;
+                    let rs_dec = rs_orig.into_py(py);
+                    let locals = PyDict::new(py);
+                    locals.set_item("rs_dec", &rs_dec).unwrap();
+                    // Checks if Rust I64F64 -> Python Decimal conversion is correct
+                    py.run(
+                        &format!(
+                            "import decimal\npy_dec = decimal.Decimal({})\nassert py_dec == rs_dec",
+                            $py
+                        ),
+                        None,
+                        Some(locals),
+                    )
+                    .unwrap();
+                    // Checks if Python Decimal -> Rust I64F64 conversion is correct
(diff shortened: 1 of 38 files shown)


src/conversions/fixed.rs
```diff
@@ -0,0 +1,1178 @@
...
+//! ```
+
+use crate::exceptions::PyValueError;
+use crate::once_cell::GILOnceCell;
+use crate::types::PyType;
+use crate::{intern, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, Python, ToPyObject};
+use fixed::types::*;
+use std::str::FromStr;
+
+static DECIMAL_CLS: GILOnceCell<Py<PyType>> = GILOnceCell::new();
```
adamreichold, line 23: We cache this three times now, don't we? I think it might be useful to centralize the `get_decimal_cls` helper function somewhere.

adamreichold:
From looking at the code, I suspect this should still be in draft status because there are still left overs from copy and pasting the `rust_decimal` integration?

More generally, we have to consider whether we really want another decimal type integration, i.e. `fpdec` and `rust_decimal`, even as an optional feature.

I don't see a technical issue in principle, but we did consider multiple crates providing this functionality and settled on `rust_decimal` as the most popular choice.

It does certainly further increase code size and maintenance effort, e.g. the work required for us to overhaul our conversion traits.

davidhewitt:
> More generally, we have to consider whether we really want another decimal type integration, i.e. `fpdec` and `rust_decimal`, even as an optional feature.

From a quick look at `fixed` and `fpdec` on crates-io, my thoughts are:
- `fixed` is ~10x smaller in download count than `rust_decimal`. It also covers a difference use case, it is not decimals but rather binary fractions. I would need to be convinced it's appropriate for us to offer this as a standard mapping from a Python decimal, due to possible loss of precision.
- `fpdec` is another ~10x smaller (so ~100x smaller than `rust_decimal`). It _is_ a decimal library and won't lose precision converting from Python decimals (unless out of the fixed range, where we could `OverflowError`). It seems like a reasonable library and the fixed size will lead to performance improvements, so probably this is a crate which a lot of users of `rust_decimal` would find appropriate. Given the tiny download count I worry the utility for users may not align with the maintenance burden for us. I appreciate that not having integrations is exactly the sort of negative network effect which can make it hard for new ideas to replace established ones. Us not having one here doesn't prevent a wrapper crate or `fdpec` to have a pyo3 feature. This has the benefit of distributing the work of PyO3 integrations rather than bogging down us maintainers with an ever-growing list of them.

alexcwyu:
@davidhewitt @adamreichold 

thanks for the review.

the main reason for using `fixed` or `fpdec` is performance.
I am leveraging rust + pyo3 to build core library for trading system (the strategy part is implemented python).

for my use case, I only need the calculation to be accurated up to certain precision (e.g. 8 dp) and would like to have good performance for low latency trading.

compare `rust_decimal` and `fixed`, `fixed is better in addition, substraction and multiplication...
```
add = 10x
Subtraction = 10x
mul = 20x better
div = similar performance
```
https://github.com/alexcwyu/rust-perf-benchmark


I understand the `fixed` is just 1/10 download of `rust_decimal` , I am happy to close this and manage the binding in my own branch.

thanks!

alexcwyu:
or i can take out `fpdec` and just add `fixed` conversion?
then we can keep the library clean & lightweight and support the high performance use case? thanks

davidhewitt:
I am actually more concerned about `fixed` than `fpdec`. As above I understand that `fixed` has lossy conversions to python `Decimal` objects, and I am therefore not convinced that it's correct to offer this as a standard conversion.

alexcwyu:
understood, let me close it for now.

I am not familar with `fixed` as well, do u know if any loss less conversion to python Decimal?

Besides, is there an easy way for me to register the converters? so that I can just publisher the converter as crate and easily register to some converter manager ?

thanks!

adamreichold:
> Besides, is there an easy way for me to register the converters? so that I can just publisher the converter as crate and easily register to some converter manager ?

The standard approach here would be to have a crate which creates a newtype wrapper around `fpdec::Decimal` and then implements e.g. `FromPyObject` for it. (It cannot implement it directly on `fpdec::Decimal` due to coherence rules.) Downstream crates must then use the wrapper type instead of `fpdec::Decimal` which can be made somewhat seamless by implementing `Deref<Target=fpdec::Decimal>` on the wrapper type.

PR #3318, opened 2023-07-14
macros: change self_arg to be an expression

davidhewitt:
This is an internal refactoring following #3311.

Before this PR the macros for `#[pymethods]` prepare the `self` argument by first having one or more statements to create a `_slf` binding, and then at the call of the Rust method they pass the binding by name.

This PR simplifies to just create the binding as a temporary expression directly at the point of calling the Rust method. I think this is generally the direction I'd like the macro codebase to head in, as experience suggests composing expressions is easier to work with than working out the order to put statements.


davidhewitt:
To avoid consuming reviewer capacity I'm going to merge this small internal refactoring. If suggestions or alternatives come up later I'm happy to merge a follow-up PR :)

PR #3321, opened 2023-07-16
collect arrays of objects prior to filling tuple in fixed-size conversions

davidhewitt:
Replacement to #3296.

As discussed there, we believe it's safe to use `PyTuple_New` when there is confidence of no Python code running while filling the uninitialized tuple.

Therefore in this PR I adjust the fixed-size conversions to collect into an array and then fill the tuple. I also happened to manage to optimize this by use of a const-generic helper which uses `PyTuple_SET_ITEM` (on the stable abi).

Benchmark shows this is ~20% faster than `main`, and should be safer too:

```
main
tuple_into_py           time:   [59.337 ns 59.947 ns 60.588 ns]

safer-tuple-into-py
tuple_into_py           time:   [46.143 ns 46.567 ns 47.005 ns]
```

I tried the same thing for `PyTuple::new` (the variable sized constructor) with a `Vec`, but that worked out around 20% slower, so I haven't included that here for now.

### DIFF
--- a/src/types/tuple.rs
+++ b/src/types/tuple.rs
@@ -284,22 +284,12 @@ fn wrong_tuple_length(t: &PyTuple, expected_length: usize) -> PyErr {
 macro_rules! tuple_conversion ({$length:expr,$(($refN:ident, $n:tt, $T:ident)),+} => {
     impl <$($T: ToPyObject),+> ToPyObject for ($($T,)+) {
         fn to_object(&self, py: Python<'_>) -> PyObject {
-            unsafe {
-                let ptr = ffi::PyTuple_New($length);
-                let ret = PyObject::from_owned_ptr(py, ptr);
-                $(ffi::PyTuple_SetItem(ptr, $n, self.$n.to_object(py).into_ptr());)+
-                ret
-            }
+            array_into_tuple(py, [$(self.$n.to_object(py)),+]).into()
         }
     }
     impl <$($T: IntoPy<PyObject>),+> IntoPy<PyObject> for ($($T,)+) {
         fn into_py(self, py: Python<'_>) -> PyObject {
-            unsafe {
-                let ptr = ffi::PyTuple_New($length);
-                let ret =  PyObject::from_owned_ptr(py, ptr);
-                $(ffi::PyTuple_SetItem(ptr, $n, self.$n.into_py(py).into_ptr());)+
-               ret
-            }
+            array_into_tuple(py, [$(self.$n.into_py(py)),+]).into()
         }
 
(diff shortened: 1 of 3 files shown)


src/types/tuple.rs
```diff
@@ -346,6 +331,22 @@ fn type_input() -> TypeInfo {
     }
 });
 
+fn array_into_tuple<const N: usize>(py: Python<'_>, array: [PyObject; N]) -> Py<PyTuple> {
+    unsafe {
+        let ptr = ffi::PyTuple_New(N.try_into().expect("0 < N <= 12"));
+        let tup = Py::from_owned_ptr(py, ptr);
+        let mut index = 0;
```
adamreichold, line 338: `for (idx, obj) in array.into_iter().enumerate()` instead of open-coding the `index` part.

adamreichold (approved):
> I tried the same thing for PyTuple::new (the variable sized constructor) with a Vec, but that worked out around 20% slower, so I haven't included that here for now.

While it is nice that this is an improvement, I think we should compare the cost of the above to the safe version using `PyTuple_Pack`. We can either use the `Vec` and `PyTuple_New` or `PyTuple_Pack` for the dynamically sized case.

davidhewitt:
> > I tried the same thing for PyTuple::new (the variable sized constructor) with a Vec, but that worked out around 20% slower, so I haven't included that here for now.
> 
> While it is nice that this is an improvement, I think we should compare the cost of the above to the safe version using `PyTuple_Pack`. We can either use the `Vec` and `PyTuple_New` or `PyTuple_Pack` for the dynamically sized case.

I haven't worked out how to use `PyTuple_Pack` for the dynamically sized case; if there's something I'm missing, would be happy for someone to show me the trick :)

src/types/tuple.rs
davidhewitt: I had open coded because the index needs to be `isize`, but clippy wanted this changed too, so I've changed and just using `index as ffi::Py_ssize_t` at the callsites. (Should always be within range.)

adamreichold:
> I haven't worked out how to use `PyTuple_Pack` for the dynamically sized case; if there's something I'm missing, would be happy for someone to show me the trick :)

Ah, they chose a variadic interface which is not nice for composeability, c.f. https://c-faq.com/varargs/invvarargs.html

I guess `libffi` could do it, e.g. https://docs.rs/libffi/latest/libffi/low/fn.prep_cif_var.html, but I am pretty sure we do not actually want that.

So I guess optimizing the common case using something like `SmallVec<[PyObject; 12]>` might be helpful? Otherwise, we just leave that one open until a better replace for `PyTuple_New` comes along?

PR #3322, opened 2023-07-16
Add tzfpy

ringsaturn:
Add [tzfpy](https://github.com/ringsaturn/tzfpy), a simple Python binding of [tzf-rs](https://github.com/ringsaturn/tzf-rs) to convert longitude/latitude to timezone name.

It's also an example project about how to build&publish Rust binding package with maturin and GitHub Actions, since it's core code is [less than 30 lines](https://github.com/ringsaturn/tzfpy/blob/main/src/lib.rs)


davidhewitt (approved):
Thanks! I force-pushed to squash.

PR #3323, opened 2023-07-16
merge PyErr internal states for simplicity

davidhewitt:
Split out from #3306 

This PR merges `PyErrState`'s `LazyTypeAndValue` and `LazyValue` variants into a single `Lazy` variant. As per #3306, this was clearly a premature optimization from me as the result is shorter code and slightly faster, if anything.

The performance difference is within 5% so I don't think it's anything worth calling out in the CHANGELOG.

### DIFF
--- a/src/err/err_state.rs
+++ b/src/err/err_state.rs
@@ -2,7 +2,7 @@ use crate::{
     exceptions::{PyBaseException, PyTypeError},
     ffi,
     types::{PyTraceback, PyType},
-    AsPyPointer, IntoPy, IntoPyPointer, Py, PyObject, Python,
+    AsPyPointer, IntoPy, IntoPyPointer, Py, PyAny, PyObject, PyTypeInfo, Python,
 };
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


src/err/err_state.rs
```diff
@@ -45,9 +46,14 @@ where
 }
 
 pub(crate) fn boxed_args(
```
adamreichold, line 48: Since this is never used to construct anything but a `PyErrState`, why not turn it into a constructor, e.g.

```rust
impl PyErrState {
  fn lazy(ptype: &PyAny, args: impl PyErrArguments + 'static) -> Self { .. }
}
```

davidhewitt: Good idea, will rework tomorrow

adamreichold (approved):
I see one semantic change: The lazy value construction is now called before the `PyExceptionClass_Check` instead of after it, but I guess when that would be problem, the cause would be in our code base as well and we could fix it.

davidhewitt:
> I see one semantic change: The lazy value construction is now called before the `PyExceptionClass_Check` instead of after it, but I guess when that would be problem, the cause would be in our code base as well and we could fix it.

Yes, indeed passing an invalid exception type will be handled during the creation of the exception object.

PR #3326, opened 2023-07-18
Preserve panic message after exception is normalized

davidhewitt:
Split off from #3306 

This was a bug I noticed on pre-3.12 Pythons where the panic message would not be extracted correctly if the `PanicException` had been normalised. After being fetched by PyO3 the `PanicException` would have its message reset to "Unwrapped panic from Python code".

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -297,9 +297,10 @@ impl PyErr {
         };
 
         if ptype.as_ptr() == PanicException::type_object_raw(py).cast() {
-            let msg: String = pvalue
+            let msg = pvalue
                 .as_ref()
-                .and_then(|obj| obj.extract(py).ok())
+                .and_then(|obj| obj.as_ref(py).str().ok())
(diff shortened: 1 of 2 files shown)


src/err/mod.rs
```diff
@@ -299,7 +299,7 @@ impl PyErr {
         if ptype.as_ptr() == PanicException::type_object_raw(py).cast() {
             let msg: String = pvalue
                 .as_ref()
-                .and_then(|obj| obj.extract(py).ok())
+                .map(|obj| obj.as_ref(py).to_string())
```
adamreichold, line 302: IIRC, we changed the `Display` impl to call into `write_unraisable` if the call to `PyAny::str` fails. Is this appropriate here? Should we call `and_then(|obj| obj.as_ref(py).str().ok())` instead since we have an alternative fallback here?

davidhewitt: Agreed, great idea.

adamreichold (approved):

PR #3330, opened 2023-07-19
move PyDict::get_item_with_error to PyDict::get_item

davidhewitt:
As per https://github.com/capi-workgroup/problems/issues/51 the recommendation of the Python core devs is to avoid `PyDict_GetItem` because it can hide arbitrary exceptions.

Accordingly I've moved the `PyDict::get_item_with_error` to just become `PyDict::get_item`, and deprecated the `get_item_with_error` name.

### DIFF
--- a/src/types/dict.rs
+++ b/src/types/dict.rs
@@ -447,8 +447,15 @@ mod tests {
     fn test_new() {
         Python::with_gil(|py| {
             let dict = [(7, 32)].into_py_dict(py);
-            assert_eq!(32, dict.get_item(7i32).unwrap().extract::<i32>().unwrap());
-            assert!(dict.get_item(8i32).is_none());
+            assert_eq!(
+                32,
+                dict.get_item(7i32)
+                    .unwrap()
+                    .unwrap()
+                    .extract::<i32>()
+                    .unwrap()
+            );
+            assert!(dict.get_item(8i32).unwrap().is_none());
             let map: HashMap<i32, i32> = [(7, 32)].iter().cloned().collect();
             assert_eq!(map, dict.extract().unwrap());
             let map: BTreeMap<i32, i32> = [(7, 32)].iter().cloned().collect();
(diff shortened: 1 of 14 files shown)


davidhewitt:
This is obviously quite an irritating change for users because it changes the return type from `Option<&PyAny>` to `PyResult<Option<&PyAny>>`. This means adding an extra `.unwrap()` in a lot of our test code. For users it's probably less bad because they can hopefully add `?` to propagate the exception.

I still defend the change as worthwhile given it nudges them onto using a "correct" API instead of unintentionally ignoring inner errors.

mejrs (approved):
+1 for this change.

It'd be nice to flesh out the docs. e.g. to answer questions like
- is it OK to unwrap or `unwrap_or(None)`?
- should I just bubble up the error?
- when can this return an error?

src/types/dict.rs
```diff
@@ -134,35 +134,10 @@ impl PyDict {
 
     /// Gets an item from the dictionary.
     ///
-    /// Returns `None` if the item is not present, or if an error occurs.
+    /// Returns `Ok(None)` if the item is not present.
     ///
     /// To get a `KeyError` for non-existing keys, use `PyAny::get_item`.
```
adamreichold, line 139: This reference to `get_item` does not make sense after the change, does it?

davidhewitt: It is still applicable because `PyAny::get_item` returns `PyResult<&PyAny>` and so nonexistent values need to raise.

I'll add an example to the doc to illustrate the difference.

davidhewitt:
Sorry for the long delay here. I've added some detail to the documentation to cover the points raised above, they were all good suggestions 👍

davidhewitt:
As this has had an approval and I think any further changes which might be requested from further reviews would be docs rather than API, I'm going to merge this now so that I can start building an 0.20 release branch.

PR #3331, opened 2023-07-20
Install wheel for the setuptools-rust-starter example

alex:
Seems to be needed on python 3.12


davidhewitt (approved):
Thanks! I think the last remaining issue is that some of the reference counting ffi needs an update for 3.12's immortal objects. I can try to take a look tonight / tomorrow.

alex:
One of the 3.12 builders is still failing, I'm not immediately sure why: https://github.com/PyO3/pyo3/actions/runs/5614086772/job/15211483629 Old version of virtualenv is my initial guess.

PR #3334, opened 2023-07-21
add PyErr::display

davidhewitt:
When looking at 3.12 FFI changes I saw the new `PyErr_DisplayException` and this inspired me to add `PyErr::Display` for all versions.

While looking at this I noticed that `PyErr_PrintEx` will call `sys.excepthook`, so I added this to the documentation.

I think we generally want to be using `.display()` instead of `.print()` in tests because there's no meaning to calling `sys.excepthook` there, so I updated the callsites.

### DIFF
--- a/newsfragments/3334.added.md
+++ b/newsfragments/3334.added.md
@@ -1,0 +1,1 @@
+Add `PyErr::Display` for all Python versions, and FFI symbol `PyErr_DisplayException` for Python 3.12.
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -431,13 +431,32 @@ impl PyErr {
     }
 
     /// Prints a standard traceback to `sys.stderr`.
+    pub fn display(&self, py: Python<'_>) {
+        #[cfg(Py_3_12)]
+        unsafe {
+            ffi::PyErr_DisplayException(self.value(py).as_ptr())
+        }
+
+        #[cfg(not(Py_3_12))]
+        unsafe {
+            ffi::PyErr_Display(
+                self.get_type(py).as_ptr(),
+                self.value(py).as_ptr(),
+                self.traceback(py)
+                    .map_or(std::ptr::null_mut(), PyTraceback::as_ptr),
+            )
+        }
+    }
+
+    /// Calls `sys.excepthook` and then prints a standard traceback to `sys.stderr`.
     pub fn print(&self, py: Python<'_>) {
         self.clone_ref(py).restore(py);
         unsafe { ffi::PyErr_PrintEx(0) }
     }
 
-    /// Prints a standard traceback to `sys.stderr`, and sets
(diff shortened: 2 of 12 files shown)


src/err/mod.rs
```diff
@@ -431,18 +431,37 @@ impl PyErr {
...
-    /// Prints a standard traceback to `sys.stderr`, and sets
-    /// `sys.last_{type,value,traceback}` attributes to this exception's data.
+    /// Calls `sys.excepthook` and then prints a standard traceback to `sys.stderr`.
+    ///
+    /// Additionally sets `sys.last_{type,value,traceback,exc}` attributes to this exception.
     pub fn print_and_set_sys_last_vars(&self, py: Python<'_>) {
         self.clone_ref(py).restore(py);
         unsafe { ffi::PyErr_PrintEx(1) }
     }
-
```
adamreichold, line 445: This one seems unintentional?

newsfragments/3334.added.md
```diff
@@ -0,0 +1 @@
+Add `PyErr_Display` for all Python versions, and FFI symbol `PyErr_DisplayException` for Python 3.12.
```
adamreichold, line 1: Shouldn't this reference `PyErr::display` which was added? (`PyErr_Display` does not seem to be new.)

adamreichold:
> so I added the ...

There seems to be something missing from the above?

davidhewitt:
> > so I added the ...
> 
> There seems to be something missing from the above?

Yes, I meant to say "added to the documentation", got distracted halfway writing 🤦

newsfragments/3334.added.md
davidhewitt: Yes, typo, will force-push fix.

src/err/mod.rs
```diff
@@ -431,13 +431,33 @@ impl PyErr {
...
+    pub fn display(&self, py: Python<'_>) {
+        unsafe {
+            ffi::PyErr_Display(
+                self.get_type(py).as_ptr(),
+                self.value(py).as_ptr(),
+                self.traceback(py)
+                    .map_or(std::ptr::null_mut(), PyTraceback::as_ptr),
+            )
+        }
+    }
```
mejrs, line 450:
```suggestion
    pub fn display(&self, py: Python<'_>) {
        unsafe { 
            if cfg!(Py_3_12) {
                ffi::PyErr_DisplayException(self.value(py).as_ptr())
            } else {
               ffi::PyErr_Display(
                  self.get_type(py).as_ptr(),
                  self.value(py).as_ptr(),
                  self.traceback(py)
                      .map_or(std::ptr::null_mut(), PyTraceback::as_ptr),
            )
        }
    }
```
Otherwise this function will show up as being unavailable on Python 3.12

davidhewitt: Good catch, thanks - will fixup when I get a moment

adamreichold (approved):

PR #3335, opened 2023-07-21
update object.h definitions for Python 3.12

davidhewitt:
This fixes a flaky reference counting issue observed in tests which seems to be related to the new "immortal objects" of Python 3.12.

So many changes had happened to `object.h` due to the immortal objects I felt it better to do a full reconciliation of the definitions in that file against the 3.12 header.

The reference counting changes are complicated, so I'd like to see a full CI run and also add some more tests before calling this ready to merge.

### DIFF
--- a/pyo3-ffi/src/object.rs
+++ b/pyo3-ffi/src/object.rs
@@ -1,5 +1,3 @@
-// FFI note: this file changed a lot between 3.6 and 3.10.
-// Some missing definitions may not be marked "skipped".
 use crate::pyport::{Py_hash_t, Py_ssize_t};
 use std::mem;
 use std::os::raw::{c_char, c_int, c_uint, c_ulong, c_void};
@@ -14,11 +12,24 @@ pub use crate::cpython::object::PyTypeObject;
 // _PyObject_HEAD_EXTRA: conditionally defined in PyObject_HEAD_INIT
 // _PyObject_EXTRA_INIT: conditionally defined in PyObject_HEAD_INIT
 
+#[cfg(Py_3_12)]
+pub const _Py_IMMORTAL_REFCNT: Py_ssize_t = {
+    if cfg!(target_pointer_width = "64") {
+        c_uint::MAX as Py_ssize_t
+    } else {
+        // for 32-bit systems, use the lower 30 bits (see comment in CPython's object.h)
+        (c_uint::MAX >> 2) as Py_ssize_t
+    }
+};
+
 pub const PyObject_HEAD_INIT: PyObject = PyObject {
     #[cfg(py_sys_config = "Py_TRACE_REFS")]
     _ob_next: std::ptr::null_mut(),
     #[cfg(py_sys_config = "Py_TRACE_REFS")]
     _ob_prev: std::ptr::null_mut(),
+    #[cfg(Py_3_12)]
+    ob_refcnt: PyObjectObRefcnt { ob_refcnt: 1 },
+    #[cfg(not(Py_3_12))]
     ob_refcnt: 1,
     #[cfg(PyPy)]
     ob_pypy_link: 0,
@@ -29,20 +40,40 @@ pub const PyObject_HEAD_INIT: PyObject = PyObject {
 // skipped Py_INVALID_SIZE
 
 #[repr(C)]
+#[derive(Copy, Clone)]
+#[cfg(Py_3_12)]
+/// This union is anonymous in CPython, so the name was given by PyO3 because
+/// Rust unions need a name.
+pub union PyObjectObRefcnt {
+    pub ob_refcnt: Py_ssize_t,
+    #[cfg(target_pointer_width = "64")]
+    pub ob_refcnt_split: [crate::PY_UINT32_T; 2],
+}
+
+#[cfg(Py_3_12)]
+impl std::fmt::Debug for PyObjectObRefcnt {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", unsafe { self.ob_refcnt })
+    }
+}
+
+#[cfg(not(Py_3_12))]
+pub type PyObjectObRefcnt = Py_ssize_t;
+
+#[repr(C)]
 #[derive(Copy, Clone, Debug)]
 pub struct PyObject {
     #[cfg(py_sys_config = "Py_TRACE_REFS")]
     pub _ob_next: *mut PyObject,
     #[cfg(py_sys_config = "Py_TRACE_REFS")]
     pub _ob_prev: *mut PyObject,
-    pub ob_refcnt: Py_ssize_t,
+    pub ob_refcnt: PyObjectObRefcnt,
     #[cfg(PyPy)]
     pub ob_pypy_link: Py_ssize_t,
     pub ob_type: *mut PyTypeObject,
 }
 
 // skipped _PyObject_CAST
-// skipped _PyObject_CAST_CONST
 
 #[repr(C)]
 #[derive(Debug, Copy, Clone)]
@@ -52,18 +83,21 @@ pub struct PyVarObject {
 }
 
 // skipped _PyVarObject_CAST
-// skipped _PyVarObject_CAST_CONST
 
 #[inline]
 pub unsafe fn Py_Is(x: *mut PyObject, y: *mut PyObject) -> c_int {
     (x == y).into()
 }
 
-// skipped _Py_REFCNT: defined in Py_REFCNT
+#[inline]
+#[cfg(Py_3_12)]
+pub unsafe fn Py_REFCNT(ob: *mut PyObject) -> Py_ssize_t {
+    (*ob).ob_refcnt.ob_refcnt
+}
 
 #[inline]
+#[cfg(not(Py_3_12))]
 pub unsafe fn Py_REFCNT(ob: *mut PyObject) -> Py_ssize_t {
-    assert!(!ob.is_null());
     (*ob).ob_refcnt
 }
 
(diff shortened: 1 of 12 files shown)


pyo3-ffi/src/object.rs
```diff
@@ -28,21 +39,39 @@ pub const PyObject_HEAD_INIT: PyObject = PyObject {
 // skipped PyObject_VAR_HEAD
 // skipped Py_INVALID_SIZE
 
+#[repr(C)]
+#[derive(Copy, Clone)]
+#[cfg(Py_3_12)]
+pub union __pyo3_object_ob_refcnt {
+    ob_refcnt: Py_ssize_t,
```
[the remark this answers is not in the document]
davidhewitt: Agreed, great catch 👍

davidhewitt:
Ok, looks like we're there 🎉 

I changed tests which interact with reference counting to use `object()` instead of `None`, because now that `None` is immortal on 3.12 it's not a suitable value to use in tests which try to modify reference counts!

davidhewitt:
I'd quite like to merge this soon so that I can proceed with the patch release. The following summarises the diff, to help with reviewing:
- On Python 3.11 `buffer.h` was renamed to `pybuffer.h`, so I made the corresponding rename here.
- On Python 3.12 the definitions of `getbufferproc` and `releasebufferproc` have been moved to `pybuffer.h`. (So I moved them for all versions here.)
- On Python 3.12 the `object.h` file has new symbols `_Py_IMMORTAL_REFCNT` and `_Py_IsImmortal`. The implementation of these differs for 32-bit architectures:
  - On 32-bit architectures, the lower 30 bits of the 32-bit reference count are set to 1 to denote an immortal object.
  - On 64-bit architectures, the lower 32 bits are set instead.
  - I expect this will all change again in future when PEP 703 / nogil comes into the picture with thread-biased reference counting.
- I took the argument names out of typedefs for `unaryfunc` etc, because the Python header doesn't give those arguments names and the argument names were just `arg1` `arg2` rather than anything meaningful. 
- I updated the implementations of a few inline macros in `object.rs`, adding helper definitions to `pyport.rs` where these were used. Most of these are trivial, except for `Py_INCREF` and `Py_DECREF`.
  - I added some tests to `src/ffi/tests.rs` to verify the behaviour of these functions. To exercise these with `abi3` I had to change that file to exclude individual tests with `#[cfg(not(Py_LIMITED_API))]` rather than excluding the whole module.

- Finally, I updated other refcounting tests to use `object()` instead of `None`, as per comment above.

I appreciate this is quite a meaty patch 😢

pyo3-ffi/src/object.rs
```diff
@@ -28,21 +39,39 @@ pub const PyObject_HEAD_INIT: PyObject = PyObject {
 // skipped PyObject_VAR_HEAD
 // skipped Py_INVALID_SIZE
 
+#[repr(C)]
+#[derive(Copy, Clone)]
+#[cfg(Py_3_12)]
+pub union __pyo3_object_ob_refcnt {
+    pub ob_refcnt: Py_ssize_t,
+    #[cfg(not(target_pointer_width = "32"))]
```
adamreichold, line 47: I think this only works for `target_pointer_width = "64"` and would hence would prefer that as the `cfg` attribute here and above. (I don't think we can reasonably CPython on CHERI here in any case.)

pyo3-ffi/src/object.rs
```diff
@@ -28,21 +39,39 @@ pub const PyObject_HEAD_INIT: PyObject = PyObject {
 // skipped PyObject_VAR_HEAD
 // skipped Py_INVALID_SIZE
 
+#[repr(C)]
+#[derive(Copy, Clone)]
+#[cfg(Py_3_12)]
+pub union __pyo3_object_ob_refcnt {
```
adamreichold, line 45: I suspect the union is anonymous in the C header, but can't we give it a nicer name here, e.g. `PyObjectRefcnt` or something like that?

pyo3-ffi/src/object.rs
```diff
@@ -72,99 +104,85 @@ pub unsafe fn Py_TYPE(ob: *mut PyObject) -> *mut PyTypeObject {
...
 
 #[inline]
 pub unsafe fn Py_IS_TYPE(ob: *mut PyObject, tp: *mut PyTypeObject) -> c_int {
     (Py_TYPE(ob) == tp) as c_int
 }
 
+#[inline(always)]
+#[cfg(all(Py_3_12, not(target_pointer_width = "32")))]
+pub unsafe fn _Py_IsImmortal(op: *mut PyObject) -> c_int {
+    (((*op).ob_refcnt.ob_refcnt as crate::PY_INT32_T) < 0) as c_int
```
adamreichold, line 125: Couldn't this use the split representation? Or is this copied verbatim from the C header?

pyo3-ffi/src/object.rs
```diff
@@ -72,99 +104,85 @@ pub unsafe fn Py_TYPE(ob: *mut PyObject) -> *mut PyTypeObject {
...
+}
+
+#[inline(always)]
+#[cfg(all(Py_3_12, target_pointer_width = "32"))]
+pub unsafe fn _Py_IsImmortal(op: *mut PyObject) -> c_int {
+    ((*op).ob_refcnt.ob_refcnt == _Py_IMMORTAL_REFCNT) as c_int
```
adamreichold, line 131: Are the top two bits never modified for immortal objects?

pyo3-ffi/src/object.rs
```diff
@@ -337,6 +376,15 @@ extern "C" {
 // Flag bits for printing:
 pub const Py_PRINT_RAW: c_int = 1; // No string quotes etc.
 
+#[cfg(all(Py_3_12, not(Py_LIMITED_API)))]
+pub const _Py_TPFLAGS_STATIC_BUILTIN: c_ulong = 1 << 1;
+
+#[cfg(all(Py_3_12, not(Py_LIMITED_API)))]
+pub const Py_TPFLAGS_MANAGED_WEAKREF: c_ulong = 1 << 3;
```
adamreichold, line 383: Is this intentionally equal to `Py_TPFLAGS_MANAGED_DICT`?

pyo3-ffi/src/object.rs
```diff
@@ -394,37 +441,161 @@ pub const Py_TPFLAGS_DICT_SUBCLASS: c_ulong = 1 << 29;
...
+            let new_refcnt = cur_refcnt.wrapping_add(1);
+            if new_refcnt == 0 {
+                return;
+            }
+            (*op).ob_refcnt.ob_refcnt_split[crate::PY_BIG_ENDIAN] = new_refcnt;
+        }
+
+        #[cfg(all(Py_3_12, target_pointer_width = "32"))]
+        {
+            if _Py_IsImmortal(op) != 0 {
```
adamreichold, line 518: Couldn't this part be shared as in the decrement implementation?

adamreichold (commented):
I have to say that CPython is certainly not becoming less baroque.

Besides the inline questions on trying to better abstract out the differences in how the reference counts are handled, I wonder whether `PyAny::get_refcnt` would have to change as this looks more like `Option<isize>` then `isize` conceptionally?

pyo3-ffi/src/object.rs
davidhewitt: Yes it's anonymous in the C header. I quite wanted to give it a name which implied that PyO3 had been forced to invent a name. I'll go for `PyObjectObRefcnt` (as the field is `ob_refcnt`), and add a docstring explaining this type is invented by PyO3.

pyo3-ffi/src/object.rs
davidhewitt: The C header used `# if SIZEOF_VOID_P > 4`, of which the closest analogy I could find was `not(target_pointer_width = "32")`, disregarding 16 bits which I don't think we would try to support.

pyo3-ffi/src/object.rs
davidhewitt: Yep copied verbatim: https://github.com/python/cpython/blob/e3b5ed7b1cb8ad12aa285cf78303c6dda0953181/Include/object.h#L243

pyo3-ffi/src/object.rs
davidhewitt: https://github.com/python/cpython/blob/e3b5ed7b1cb8ad12aa285cf78303c6dda0953181/Include/object.h#L182-L184

pyo3-ffi/src/object.rs
davidhewitt: This implementation matches the C header: https://github.com/python/cpython/blob/e3b5ed7b1cb8ad12aa285cf78303c6dda0953181/Include/object.h#L245

I think in practice the top two bits are reserved for abi3 builds built against older Python implementations, for which `Py_INCREF` would still directly modify the refcount field and so immortal objects need to allow some space for the count to spill over without wrapping.

pyo3-ffi/src/object.rs
davidhewitt: `MANAGED_DICT` should be `1 << 4`, great catch 👍

pyo3-ffi/src/object.rs
davidhewitt: In the header it's split like this where the "immortality" in the 64-bit implementation is achieved by the `wrapping_add` to the lower 32 bits will hit 0, and checking for it.

https://github.com/python/cpython/blob/e3b5ed7b1cb8ad12aa285cf78303c6dda0953181/Include/object.h#L640-L654

I agree it looks a bit funky, presumably there was a technical reason this was preferred. I'd like to match the C implementation as faithfully as possible.

(I'm also hoping that when nogil comes along the whole macro is removed so that we don't have to try to reproduce it here 😂)

davidhewitt:
Pushed a commit with review corrections, I'll squash before merge.

I think there is an open question for `PyAny::get_refcnt`. I'll open an issue.

pyo3-ffi/src/object.rs
adamreichold: I think copying verbatim is not useful here and would prefer the direct formulation as the structure does not make sense for anything but 64 bit architectures.

pyo3-ffi/src/object.rs
adamreichold: Ok, the salient point here is probably

> The object would lose its immortality, but the execution would still be correct.

from the C header. So indeed, in 32-bit builds, these object would really not be considered "immortal" any more after such an extension touched their reference.

Weird, but matching upstream behaviour. (I really wonder whether the benefit of having immortal objects is worth this technical complexity.)

adamreichold (approved):
LGTM modulo the usage of `target_point_width`.

davidhewitt:
🙏 thanks for the thorough reviewing!

Changed to use `cfg(target_pointer_width = "64")` instead of `cfg(not(target_pointer_width = "32"))`. We won't support 16 or 128-bit systems for now, can cross that bridge then it seems like there's a reason to.

PR #3337, opened 2023-07-22
Add haem to example projects

BooleanCat:
Add haem to example projects


BooleanCat:
I see the changelog check is failing, but I don't think it's necessary to mention this change in release notes. Poke me if you disagree.

davidhewitt:
Indeed not necessary. Thanks for sharing this!

PR #3339, opened 2023-07-23
Add PyType_GetDict for Python 3.12

cdce8p:
`PyType_GetDict` was added in Python 3.12 to safely access `tp_dict`.

```c
PyAPI_FUNC(PyObject *) PyType_GetDict(PyTypeObject *);
```

https://github.com/python/cpython/blob/v3.12.0b4/Include/cpython/object.h#L286
https://github.com/python/cpython/blob/v3.12.0b4/Objects/typeobject.c#L241-L246

https://docs.python.org/3.12/c-api/type.html#c.PyType_GetDict
https://docs.python.org/3.12/c-api/typeobj.html#c.PyTypeObject.tp_dict
https://docs.python.org/3.12/whatsnew/changelog.html?highlight=pytype_getdict#c-api

### DIFF
--- a/newsfragments/3339.added.md
+++ b/newsfragments/3339.added.md
@@ -1,0 +1,1 @@
+Define `PyType_GetDict()` FFI for CPython 3.12 or later.
--- a/pyo3-ffi/src/cpython/object.rs
+++ b/pyo3-ffi/src/cpython/object.rs
@@ -349,6 +349,9 @@ pub unsafe fn PyHeapType_GET_MEMBERS(
 // skipped _PyType_GetModuleByDef
 
 extern "C" {
+    #[cfg(Py_3_12)]
+    pub fn PyType_GetDict(o: *mut PyTypeObject) -> *mut PyObject;
+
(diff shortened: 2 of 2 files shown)


adamreichold (approved):

cdce8p:
Would it be possible to get this backported and released with `1.19.2`? It's necessary to build some packages for `3.12`. E.g. https://github.com/ijl/orjson/pull/408

adamreichold:
> Would it be possible to get this backported and released with `1.19.2`? It's necessary to build some packages for `3.12`. E.g. [ijl/orjson#408](https://github.com/ijl/orjson/pull/408)

We are still working on compatibility with Python 3.12 and plan a 0.19.x point release which contains mainly that. I suspect we will need to make use of this ourselves, so I am sure it will be cherry-picked into that release too. It might still take a while though.

adamreichold:
(Note that while it should of course end up in a released version here, you can fix this in downstream crates by adding the `extern "C"` declaration there if you need this urgently. Not something I would generally recommend, but it should work as long as one of the dependencies pulls in the defining DSO.)

PR #3342, opened 2023-07-25
fix ffi check failures for 3.12.0b4

davidhewitt:
While trying to get #3335 green I hit some `ffi-check` failures in CI. As I descended down the rabbit hole, there became enough of these that it felt appropriate to split them out into a separate PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ on:
 
 jobs:
   build:
-    continue-on-error: ${{ contains(fromJSON('["3.7", "3.12-dev", "pypy3.7", "pypy3.10"]'), inputs.python-version) }}
+    continue-on-error: ${{ endsWith(inputs.python-version, '-dev') || contains(fromJSON('["3.7", "pypy3.7", "pypy3.10"]'), inputs.python-version) }}
     runs-on: ${{ inputs.os }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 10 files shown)


davidhewitt:
For now I've pushed these commits also to #3335, and hopefully we will get a green CI run there even if we don't quite get one here. If that PR is green, I'd suggest merging this one first and then that one after.

.github/workflows/build.yml
```diff
@@ -25,7 +25,7 @@ on:
 
 jobs:
   build:
-    continue-on-error: ${{ contains(fromJSON('["3.7", "3.12-dev", "pypy3.7", "pypy3.10"]'), inputs.python-version) }}
+    continue-on-error: ${{ endsWith(inputs.python-version, '-dev') || contains(fromJSON('["3.7", "pypy3.7", "pypy3.10"]'), inputs.python-version) }}
```
adamreichold, line 28: Do we really want to generalize this? I would hope that not all version bumps carry that much baggage as the 3.12 one.

pyo3-ffi-check/macro/src/lib.rs
```diff
@@ -130,16 +130,23 @@ pub fn for_all_fields(input: proc_macro::TokenStream) -> proc_macro::TokenStream
...
             .value()
             .id()
             .unwrap()
             .strip_prefix("structfield.")
             .unwrap();
 
-        let field_ident = Ident::new(id, Span::call_site());
+        let field_ident = Ident::new(field_name, Span::call_site());
+        let bindgen_field_ident = if struct_name == "PyObject" && field_name == "ob_refcnt" {
+            // PyObject since 3.12 implements ob_refcnt as a union; bindgen creates
```
adamreichold, line 142: This is likely not the time to make that change, but why scrape this instead of using Rustdoc's JSON output? Is that still nightly-only?

.github/workflows/build.yml
davidhewitt: I think it would be desirable to switch on CI with future alphas ASAP, however I'd make a bet that ffi-check (at minimum) will break with every new alpha version, and maybe other test failures. I'd personally prefer this because it should provide a nudge to make small fixups through the year as each alpha / beta release drops rather than what's happened this year where PyO3 has fallen relatively far behind on 3.12 beta compatibility.

If future version bumps are trivial then generalizing like this is not a big deal; on the other hand if a future version bump is nasty, then generalizing like this will minimize blocking failures for other contributors :)

pyo3-ffi-check/macro/src/lib.rs
davidhewitt: Looks like it's still unstable, https://github.com/rust-lang/rust/issues/76578

However it would be very nice to use this when available!

.github/workflows/build.yml
adamreichold: Fair enough.

adamreichold (approved):

PR #3345, opened 2023-07-27
optimize `float` -> `f64` conversions on non-abi3

davidhewitt:
On non-abi3 builds we can use `PyFloat_AS_DOUBLE` to avoid an FFI function call and instead read the double value directly from the `PyFloatObject` structure.

This gives nearly a 2x speedup to `f64` extraction when the input type is `float` according to the benchmark I added:

```
main

f64_from_pyobject       time:   [1.9698 ns 1.9870 ns 2.0047 ns]

branch

f64_from_pyobject       time:   [1.0878 ns 1.0977 ns 1.1081 ns]
                        change: [-45.172% -44.325% -43.403%] (p = 0.00 < 0.05)
```

Given I would expect this is the majority case of `f64` conversions I think the slight bump in complexity is worth it given the drop in overhead.

While I was at it I moved the ffi definitions around to match their positioning in 3.12's headers.

### DIFF
--- a/src/types/floatob.rs
+++ b/src/types/floatob.rs
@@ -1,7 +1,8 @@
 #[cfg(feature = "experimental-inspect")]
 use crate::inspect::types::TypeInfo;
 use crate::{
-    ffi, AsPyPointer, FromPyObject, IntoPy, PyAny, PyErr, PyObject, PyResult, Python, ToPyObject,
+    ffi, AsPyPointer, FromPyObject, IntoPy, PyAny, PyErr, PyNativeType, PyObject, PyResult, Python,
+    ToPyObject,
 };
 use std::os::raw::c_double;
 
@@ -29,7 +30,16 @@ impl PyFloat {
 
     /// Gets the value of this float.
     pub fn value(&self) -> c_double {
-        unsafe { ffi::PyFloat_AsDouble(self.as_ptr()) }
+        #[cfg(not(Py_LIMITED_API))]
+        unsafe {
+            // Safety: self is PyFloat object
+            ffi::PyFloat_AS_DOUBLE(self.as_ptr())
+        }
+
+        #[cfg(Py_LIMITED_API)]
+        unsafe {
+            ffi::PyFloat_AsDouble(self.as_ptr())
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/types/floatob.rs
```diff
@@ -54,6 +71,13 @@ impl<'source> FromPyObject<'source> for f64 {
     // PyFloat_AsDouble returns -1.0 upon failure
     #![allow(clippy::float_cmp)]
     fn extract(obj: &'source PyAny) -> PyResult<Self> {
+        #[cfg(not(Py_LIMITED_API))]
+        unsafe {
+            if obj.is_exact_instance_of::<PyFloat>() {
```
adamreichold, line 76: Is the check-then-unchecked-downcast faster than just using the checked downcast?

src/types/floatob.rs
```diff
@@ -29,7 +30,23 @@ impl PyFloat {
...
+        #[cfg(Py_LIMITED_API)]
+        unsafe {
+            return ffi::PyFloat_AsDouble(self.as_ptr());
+        }
+    }
+
+    // private safe wrapper around PyFloat_AS_DOUBLE
+    // not exposed because .value() uses this on the unlimited API anyway
+    #[cfg(not(Py_LIMITED_API))]
+    fn as_double_fastpath(&self) -> c_double {
```
adamreichold, line 47: Couldn't `extract` just call `value` directly, avoid the need for this additional helper method?

src/types/floatob.rs
davidhewitt: I use `is_exact_instance_,of` here because I don't want to go through `isinstance` machinery for subclasses (would probably be a deoptimization).

I have thought for a while there is space for a `PyAny::downcast_exact` method. If you like, I could PR that first and then use it here?

src/types/floatob.rs
davidhewitt: I started out like that but noted it's slower to do the downcast and then call `.value()` for `abi3` (the check is redundant on abi3). So I decided to use a fastpath method so we can't accidentally remove the `#[cfg]` from the branch in extract and deoptimize `abi3`.

src/types/floatob.rs
adamreichold: Sounds reasonable.

src/types/floatob.rs
adamreichold: My thinking was to still start with `downcast_exact` and only for `not(Py_LIMITED_API)` and then call into `.value()` which should expand only to the `not(Py_LIMITPED_API)` code path. Hence the redundant call to `PyFloat_AsDouble` would never be emitted.

So basically, literally replacing `.as_double_fastpath()` by `.value()` should yield the exact same generated code, should it not?

src/types/floatob.rs
davidhewitt: Done 👍

src/types/floatob.rs
davidhewitt: Done 👍

adamreichold (approved):

PR #3347, opened 2023-07-28
fix compile failure for getter with return lifetime of self

davidhewitt:
While testing `main` downstream I noticed a compile failure caused by #3318, when the value returned from a `#[getter]` has the same lifetime as the `&self` receiver.

This PR adds a corresponding test and fixes it. The solution is the usual trend for the macro function bodies towards expressions; there is a sequence of statements along the lines of `let item = ...; return convert(item)`, the PR changes this to just be `convert(...)`.

This doesn't need a changelog entry as the compile failure never hit a released version.

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -2,6 +2,7 @@ use crate::attributes::{TextSignatureAttribute, TextSignatureAttributeValue};
 use crate::params::impl_arg_params;
 use crate::pyfunction::{FunctionSignature, PyFunctionArgPyO3Attributes};
 use crate::pyfunction::{PyFunctionOptions, SignatureAttribute};
+use crate::quotes;
 use crate::utils::{self, PythonDoc};
 use proc_macro2::{Span, TokenStream};
 use quote::ToTokens;
@@ -413,11 +414,7 @@ impl<'a> FnSpec<'a> {
         let func_name = &self.name;
 
         let rust_call = |args: Vec<TokenStream>| {
-            quote! {
-                _pyo3::impl_::pymethods::OkWrap::wrap(function(#self_arg #(#args),*), #py)
-                    .map(|obj| _pyo3::conversion::IntoPyPointer::into_ptr(obj))
-                    .map_err(::core::convert::Into::into)
-            }
+            quotes::map_result_into_ptr(quotes::ok_wrap(quote! { function(#self_arg #(#args),*) }))
         };
 
         let rust_name = if let Some(cls) = cls {
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -115,12 +115,12 @@ impl FnType {
             }
             FnType::FnClass | FnType::FnNewClass => {
                 quote! {
-                    _pyo3::types::PyType::from_type_ptr(_py, _slf as *mut _pyo3::ffi::PyTypeObject),
+                    _pyo3::types::PyType::from_type_ptr(py, _slf as *mut _pyo3::ffi::PyTypeObject),
                 }
             }
             FnType::FnModule => {
                 quote! {
-                    _py.from_borrowed_ptr::<_pyo3::types::PyModule>(_slf),
+                    py.from_borrowed_ptr::<_pyo3::types::PyModule>(_slf),
                 }
             }
         }
@@ -155,7 +155,7 @@ impl ExtractErrorMode {
 
 impl SelfType {
     pub fn receiver(&self, cls: &syn::Type, error_mode: ExtractErrorMode) -> TokenStream {
-        let py = syn::Ident::new("_py", Span::call_site());
+        let py = syn::Ident::new("py", Span::call_site());
         let slf = syn::Ident::new("_slf", Span::call_site());
         match self {
             SelfType::Receiver { span, mutable } => {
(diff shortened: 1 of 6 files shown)


pyo3-macros-backend/src/pymethod.rs
```diff
@@ -634,31 +634,28 @@ pub fn impl_py_getter_def(
...
-    let conversion = match property_type {
-        PropertyType::Descriptor { .. } => {
+                syn::Index::from(field_index).to_token_stream()
+            };
             quote! {
-                let item: _pyo3::Py<_pyo3::PyAny> = _pyo3::IntoPy::into_py(item, _py);
-                ::std::result::Result::Ok(_pyo3::conversion::IntoPyPointer::into_ptr(item))
+                ::std::result::Result::Ok(
+                    _pyo3::conversion::IntoPyPointer::into_ptr(
+                        _pyo3::IntoPy::<_pyo3::Py<_pyo3::PyAny>>::into_py(::std::clone::Clone::clone(&(#slf.#field_token)), _py)
```
adamreichold, line 647: To be honest, I think this somewhat showcases the downside of the expression style insofar this expression is becoming an unreadable wall of text. The previous version with the extra binding was not great, but still more readable IMHO.

Could we at least try to use interpolation and multiple `quote!` invocation to try to pull this apart in our macro implementation while still emitting a single expression in the end?

davidhewitt: Sure thing, I'll push a refactoring later (might create a bit of a bigger diff).

pyo3-macros-backend/src/quotes.rs
```diff
@@ -0,0 +1,15 @@
...
+        _pyo3::impl_::pymethods::OkWrap::wrap(#obj, py)
+            .map_err(::core::convert::Into::into)
+    }
+}
+
+pub(crate) fn map_result_into_ptr(result: TokenStream) -> TokenStream {
+    quote! {
+        #result.map(_pyo3::IntoPyPointer::into_ptr)
+    }
+}
```
davidhewitt, line 15: I think this function can be removed (and code moved out of the macros) by moving the `Py<PyAny> -> *mut ffi::PyObject` conversion into the trampolines. That'll also have a positive effect on code size, so I'll enact that as a follow-up within a few days.

davidhewitt (commented):
Ok, I've pushed a commit which improves the implementation of `impl_py_getter_def` a little bit.

I've identified a follow-up which will move code out of the macro-generated code and into the trampolines, so I'd like to ask to merge this as-is (even if the macro expression is still a bit verbose) because I'll simplify further in the follow up.

davidhewitt:
I also pushed a commit which moves all usage of the `_py` ident to `py`, which allowed a general reduction in complexity across the macro code (including inside this `impl_py_getter_def` once I moved to the `ok_wrap` form).

pyo3-macros-backend/src/method.rs
```diff
@@ -164,19 +167,15 @@ impl SelfType {
...
-                    quote_spanned! { *span =>
-                        _pyo3::impl_::extract_argument::#method::<#cls>(
-                            #py.from_borrowed_ptr::<_pyo3::PyAny>(#slf),
-                            &mut { _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT },
-                        )
-                    },
-                )
+                error_mode.handle_error(quote_spanned! { *span =>
+                    _pyo3::impl_::extract_argument::#method::<#cls>(
+                        #py.from_borrowed_ptr::<_pyo3::PyAny>(#slf),
```
adamreichold, line 172: Below, `py` is used, but here it stays `#py`. Oversight or deliberate?

pyo3-macros-backend/src/method.rs
```diff
@@ -155,7 +156,9 @@ impl ExtractErrorMode {
 
 impl SelfType {
     pub fn receiver(&self, cls: &syn::Type, error_mode: ExtractErrorMode) -> TokenStream {
-        let py = syn::Ident::new("_py", Span::call_site());
```
adamreichold, line 158: Do we know that there was no reason to do this and/or that we have a regression test that which catch any issues introduce by not quoting the GIL token?

adamreichold (commented):
While I welcome the reduction in Perl-style sigils by not quoting the GIL token, I somewhat fear regressions in macro hygiene due to this.

I am also not sure what the motivation for doing that change now is? Sure, this implies another reduction in sigil noise in the proc macro code, but to me, it appears small compared to possible regression in macro hygiene, especially which such a common identifier as `py` (at least in PyO3 code bases).

I am pretty sure I am missing something, because this appears somewhat risky to me.

pyo3-macros-backend/src/method.rs
davidhewitt: Most likely oversight 🙈

pyo3-macros-backend/src/method.rs
davidhewitt: Let's move this to #3356

davidhewitt:
That's a reasonable concern. The `py` ident should always have `call_site` hygiene, and the only case that it's necessary to explicitly construct the ident is when mixing with `quote_spanned` to emit other tokens with a different hygiene. (The interpolated `#py` in a `quote_spanned` will keep its original span.)

The `py` ident was parameterised further in some functions because we have used both `py` and `_py` as the identifier inside the macros codebase.

To remove the hygiene concerns from this PR I've moved that commit to #3356 and replaced it here with a simpler commit which unifies all idents as `py` in the macros codebase. That's sufficient to allow the `ok_wrap` helper to work without needing to take `py` ident as an argument.

adamreichold (approved):

PR #3350, opened 2023-07-28
Derive Macro for IntoPyDict

vidur2:
Create derive macro for IntoPyDict trait. Derives all primitives and std::collections except for BTreeMap and HashMap. Currently does not work on enums or tuple structs, but support will be added later.


tests/test_intopydict.rs
```diff
@@ -0,0 +1,32 @@
+use pyo3::{prelude::IntoPyDict, types::PyDict, Python};
```
mejrs: I think this will fix the CI failure.

```suggestion
#![cfg(feature = "macros")]
use pyo3::{prelude::IntoPyDict, types::PyDict, Python};
```

mejrs (changes requested):
Thanks for your contribution :)

Before doing an in depth review, I'd like to see the following addressed:
- Can you give a motivating example for this? Is this useful to you personally, and how would you use it? Think in terms of another user seeing this and asking themselves "should i implement this for my structs? Is this useful?"
- Don't use `&str` and/or `String`. Instead use `syn`'s types and compose code with the `quote!` macro, like our other macro code does.
- Don't panic in macro code, the user experience with this is really bad. Instead use `syn::Error` and `syn::Result`.

tests/test_intopydict.rs
vidur2: Thank you so much for your reply! The reason I made this contribution is because I had to manually implement IntoPyDict on some of my rust structs even though the implementation seemed relatively trivial (ie only primitive fields). I figured there should be a derive macro for the trivial case. I am passing the output into an tf model, and I find it is easier to parse the elements from a dictionary to a tensor on the python side.

vidur2: Also I'm having a little bit of trouble with the quote! macro. Could you help me out with it?

mejrs: I can't say I am an expert on it. I'd recommend reading `quote`'s documentation, online guides and our existing macro code to see what it does.

vidur2: Do you agree with the use case? Also why did the CI stop running checks. This is my first time committing to an open source repo like this.

mejrs: > Do you agree with the use case?

The other maintainers would have to weigh in as well, but I think I can see where you are coming from. An example would help.

You can just run the tests locally, these should catch most issues.

vidur2: In the local tests it keeps on failing on the test 'test_compile_errors'

davidhewitt (commented):
Thank you for working on this functionality.

Overall I'm ok with having this, I've seen lots of manual implementations of `to_dict` so having a way to make to make this easier is welcome. An alternative is #1703 which could add a `py_dict!` macro, that's still more boilerplate than a derive.

The biggest design question in my head is what happens when a `#[pyclass]` implements `IntoPyDict` and is included in a sub-structure. e.g.

```rust
#[pyclass]
#[derive(IntoPyDict)]
struct A { }

#[pyclass]
#[derive(IntoPyDict)]
struct B {
    pub a: A,
}
```

Should `IntoPyDict` do a deep conversion into Python dictionaries, or should it be shallow? In the example above, should `IntoPyDict` for `B` result in `{'a': {} }` or `{'a': <object of type A>}`? 

Is there a right answer, or should this be configurable?

How does this macro interact with other PyO3 options? I can imagine at least `#[pyo3(crate = ...)]` and `#[pyo3(name = "...")]` would affect this.

vidur2:
Python 3.7 added the dataclasses module which enabled users of the language to write classes in a similar syntax to rust structs or typescript types. In addition to being syntactic sugar over the __init__ dunder, it also came with an asdict function which could turn any dataclass into a dictionary. If a field of a dataclass was a dataclass, then it would turn that field into a dataclass. If it was just a normal class then it would do a shallow copy. I suggest we follow this format (though I do not have it implemented this way yet). The official documentation for dataclass's asdict can be found [here](https://docs.python.org/3/library/dataclasses.html).

vidur2:
I don't think it would be affected by pyo3 options because it directly takes the field names of the class as the dictionary keys. Is this a wrong designwise?

vidur2:
Im having a little trouble understanding what ```#[pyo3(crate=...]``` does. Do you mind explaining it to me?

pyo3-macros-backend/src/intopydict.rs
```diff
@@ -0,0 +1,366 @@
...
+
+fn split_struct(binding: String) -> HashMap<String, (String, Option<String>)> {
+    let mut stack: Vec<char> = Vec::new();
+    let mut start = 0;
+    let binding = binding.replace('\n', "");
+    let mut name_map: HashMap<String, (String, Option<String>)> = HashMap::new();
+
+    for (i, char_val) in binding.chars().enumerate() {
+        if char_val == ',' && stack.is_empty() {
+            if binding[start..i].starts_with('#') {
```
davidhewitt, line 128: I think you might want to take a look at `take_pyo3_options` to see how the rest of the crate parses this.

tests/test_intopydict.rs
```diff
@@ -0,0 +1,74 @@
...
+#[derive(IntoPyDict, PartialEq, Debug, Clone)]
+pub struct Test1 {
+    x: u8,
+}
+
+#[derive(IntoPyDict, Clone, Debug)]
+pub struct Test {
+    #[pyo3(get, set, name = "hello")]
+    v: Vec<Vec<Test1>>,
+    j: Test1,
```
davidhewitt, line 28: Potentially add a `#[cfg(FALSE)] z: i32` field to check interaction with `#[cfg]`.

davidhewitt (changes requested):
The purpose of `#[pyo3(crate = "foo")]` is to redirect the macro to use `foo` instead of `pyo3` as the top-level symbol. There are several reasons to do this; within `test_hygiene` we use this to run the macros within pyo3's lib test suite. Users might have pyo3 as a transient dependency e.g. `foo::pyo3`.

The way I suggest the macro should handle it is emit an outer `const _: () = { use pyo3 as _pyo3; <impl> };` structure, where `<impl>` is the derive macro and then can use `_pyo3` everywhere internally.

At the moment this implementation is significantly different from the rest of our macro machinery. That's not the end of the world, but I do think there are a few edge cases which will need to be addressed which would be helped by using the existing machinery.

Also, the implementation definitely needs to move from parsing type names as strings to emitting code which gets the trait solver to work out the final details. Users can freely rename types or add new types etc, so parsing type names is generally something we use as a last resort rather than something core to the implementation.

pyo3-macros-backend/src/intopydict.rs
```diff
@@ -0,0 +1,366 @@
...
+const COL_NAMES: [&str; 8] = [
+    "BTreeSet",
+    "BinaryHeap",
+    "Vec",
+    "HashSet",
+    "LinkedList",
+    "VecDeque",
+    "BTreeMap",
+    "HashMap",
+];
```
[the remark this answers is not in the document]
vidur2: For autoref is there a particular crate that the rest of pyo3 uses/you reccomend?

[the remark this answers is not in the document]
vidur2: Would this apply to collections as well (ie Vecs, Sets etc)? These would get stored as a PyObject instead of a PyList?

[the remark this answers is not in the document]
vidur2: And how do you recommend I use autoref specialization? I don't have much experience with it. Is there a crate which makes it easier?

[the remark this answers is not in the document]
adamreichold: A separate crate usually doesn't help because abstracting the pattern way is often more complex than just applying it.

In this case, the general technique is described at https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md and you can an example of its application in PyO3's existing macros by looking for usages of the type `PyClassImplCollector` defined at 

https://github.com/PyO3/pyo3/blob/ae982b8ad0fff71d4d2fdbc2f7055500a6fb450f/src/impl_/pyclass.rs#L109

and for example its relation to the trait `PyClassNewTextSignature` defined at

https://github.com/PyO3/pyo3/blob/ae982b8ad0fff71d4d2fdbc2f7055500a6fb450f/src/impl_/pyclass.rs#L995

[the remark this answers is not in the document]
vidur2: Should I add the specialization feature as a crate attribute?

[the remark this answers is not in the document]
davidhewitt: No; the specialization feature is incomplete and unsound. Autoref specialization is a technique which works on stable Rust to get a specialization-like result for some limited cases.

pyo3-macros-backend/src/intopydict.rs
```diff
@@ -0,0 +1,366 @@
...
+    CollectionSing(Box<crate::intopydict::Pyo3Type>),
+    // Map(
+    //     Box<crate::intopydict::Pyo3Type>,
+    //     Box<crate::intopydict::Pyo3Type>,
+    // ),
+}
+
+#[derive(Debug, Clone)]
+pub struct Pyo3DictField {
+    name: String,
```
mejrs, line 34: This should probably be an `Ident` or `&Ident`, not `String`.

pyo3-macros/src/lib.rs
```diff
@@ -160,6 +161,38 @@ pub fn derive_from_py_object(item: TokenStream) -> TokenStream {
...
+    let generics = ast.generics.into_token_stream();
+
+    if let Some(clause) = clause_wrapped {
+        where_clause = clause.into_token_stream();
+    }
+    let mut dict_fields: Pyo3Collection = Pyo3Collection(Vec::new());
+    for token in item {
+        let token_stream: syn::__private::TokenStream = token.into();
+        dict_fields += parse_macro_input!(token_stream as Pyo3Collection);
+    }
```
mejrs, line 182: Don't go back and forth between `syn` types, tokenstreams and strings here and elsewhere. Parse it into `DeriveInput`, *once*, then just write functions that take that or the types contained in it.

pyo3-macros-backend/src/intopydict.rs
```diff
@@ -0,0 +1,366 @@
...
+            generics_parsed += ",";
+        }
+
+        generics_parsed = generics_parsed[0..generics_parsed.len() - 1].to_string();
+        generics_parsed += ">";
+        generics_parsed
+    } else {
+        String::new()
+    }
+}
```
mejrs, line 343: [Syn has a function for doing this](https://docs.rs/syn/latest/syn/struct.Generics.html#method.split_for_impl).

mejrs (commented):
My comment regarding strings still stands. You shouldn't be parsing a single `String`, `&str` or `char` in this code. Use the appropriate `syn` types instead. In general: when writing macros if you feel you have to use strings you are either using quote/syn wrong or you just haven't found the right function/method yet.

pyo3-macros-backend/src/intopydict.rs
vidur2: Thank you! Ill keep that in mind

pyo3-macros-backend/src/intopydict.rs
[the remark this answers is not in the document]
vidur2: There seems to be some issues with reference moving when I try to do autoref specialization. Sorry it took me a while for this, I was finishing up an internship

vidur2:
Im finding that I cannot parse the pyo3 options when the struct is annotated with a ```#[pyclass]``` attribute macro. Is there an alterantive way to do this within pyo3's macro system instead of that provided by syn

davidhewitt:
Not totally sure what the problem you're encountering is, probably it's related to the fact that one of the macros will be running on the result of the other.

The solution might be to detect this case and run both expansions at the same time.

davidhewitt:
Thanks for the contributions here. It seems like this work has stalled, and I'm doing a bit of housekeeping this morning so I'm going to close this PR for now.

I think there are still open design questions and the implementation needed further work. If anyone is interested in continuing this please comment and we can discuss design.

PR #3348, opened 2023-07-28
release: 0.19.2

davidhewitt:
I think we're almost ready to release a 0.19.2 patch release to improve compatibility with 3.12.

Closes #3305 
cc @decathorpe 

I'd like to get just a few last PRs approved and merged before releasing; I will then cherry-pick them and go through the motions of the version bump / CHANGELOG prep in this PR:

- #3335 
- #3345 
- #3346 
- #3347

PR #3355, opened 2023-07-29
decouple benchmarks from msrv

davidhewitt:
This PR moves the `benches/` subdirectory inside a `pyo3-benches/` isolated workspace, similar to how we already made `pyo3-ffi-check/` its own workspace. To keep running benchmarks convenient, I've added `nox -s bench` as a wrapper around `cargo bench`.

The motivation is to remove benchmark dependencies from MSRV (aka criterion), I think to make CI pass in #3348 everything I can take out of the dependency graph will make life easier. This also means that dependabot upgrades for criterion will be trivially mergeable in future without MSRV concerns.


.github/workflows/gh-pages.yml
```diff
@@ -74,8 +74,10 @@ jobs:
 
       - name: Run benchmarks
         run: |
-          for bench in call dict gil list pyclass pyobject set tuple; do
-            cargo bench --features hashbrown --bench "bench_$bench" -- --output-format bencher | tee -a output.txt
+          python -m pip install --upgrade pip && pip install nox
```
adamreichold, line 77: This installs Nox but does not seem to use it?

davidhewitt: Yeah it should be `nox -s bench` in the command below, will fixup 🤦

adamreichold (approved):

PR #3356, opened 2023-07-29
remove all #py quoting

davidhewitt:
Split out from #3347, to review separately after the 0.19.2 patch release has been done.

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -141,13 +141,13 @@ pub enum ExtractErrorMode {
 }
 
 impl ExtractErrorMode {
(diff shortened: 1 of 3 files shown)


davidhewitt:
Ok, I've rebased. My justification for why I believe we can do this:
- We now call the symbol `py` everywhere in the macros. Previously we used a mix of `py` and `_py` which 
- We always define `py` in the top-level `quote!`, which has `call_site()` hygiene. All other `quote!` uses will also have `call_site()` hygiene. So it's only when we use `quote_spanned!` with a different hygiene that we need to place this ident in with the correct hygiene. I've put a comment in the location where that's necessary.

adamreichold (approved):

PR #3358, opened 2023-07-30
fixes #3325 -- mark `AsPyPointer` as `unsafe trait`

### LINKED ISSUE #3325: AsPyPointer should be unsafe trait
### Bug Description

Various code in pyo3 relies on `AsPyPointer::as_ptr` to return a valid pointer. But because it's not an unsafe trait, it has no invariants.

### Steps to Reproduce

```rust
struct Malice;

impl pyo3::AsPyPointer for Malice {
    fn as_ptr(&self) -> *mut pyo3::ffi::PyObject {
        0xDEAFBEEF as *mut pyo3::ffi::PyObject
    }
}

fn main() {
    pyo3::Python::with_gil(|py| {
        let obj = Malice;
        pyo3::types::PyIterator::from_object(py, &obj).expect("Failed to create iterator");
    });
}
```

### Backtrace

```shell
/t/t ❯❯❯ cargo run
   Compiling t v0.1.0 (/private/tmp/t)
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/t`
fish: Job 1, 'cargo run' terminated by signal SIGSEGV (Address boundary error)
/t/t ❯❯❯ gdb^C
/t/t ❯❯❯ lldb ./target/debug/t
(lldb) target create "./target/debug/t"
Current executable set to '/tmp/t/target/debug/t' (arm64).
(lldb) run
Process 46947 launched: '/tmp/t/target/debug/t' (arm64)
Process 46947 stopped
* thread #1, name = 'main', queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xdeafbef7)
    frame #0: 0x000000010062f120 Python`PyObject_GetIter + 16
Python`PyObject_GetIter:
->  0x10062f120 <+16>: ldr    x8, [x0, #0x8]
    0x10062f124 <+20>: ldr    x9, [x8, #0xd8]
    0x10062f128 <+24>: cbz    x9, 0x10062f160           ; <+80>
    0x10062f12c <+28>: blr    x9
Target 0: (t) stopped.
```
```


Impacts all

Impacts all

Impacts all

### Your PyO3 version

0.19

Impacts all

### Additional Info

_No response_

### PR DESCRIPTION
**Pull request by alex**

### CODE REVIEW

File: guide/src/migration.md [hunk 1]
```diff
@@ -62,6 +62,10 @@ fn add(a: u64, b: u64) -> u64 {
 }
 ```
 
+### `AsPyPointer` and `IntoPyPointer` now `unsafe` traits
```
davidhewitt, line 65:
```suggestion
### `AsPyPointer` and `IntoPyPointer` are now `unsafe` traits
```

src/conversion.rs
```diff
@@ -58,7 +58,9 @@ use std::ptr::NonNull;
 /// and the Python object is dropped immediately after the `0xabad1dea_u32.into_py(py).as_ptr()`
 /// expression is evaluated. To fix the problem, bind Python object to a local variable like earlier
 /// to keep the Python object alive until the end of its scope.
-pub trait AsPyPointer {
+///
+/// Implementors must ensure this returns a valid pointer to a Python object.
```
adamreichold, line 62: Does this suffice? Do we want to demand some relation to the `&self` parameter, e.g. that the resulting must be derived from and only from `&self` to avoid implementations returning completely unrelated pointers?

As a more general observation. Do we need this in generic context? As an alternative would an inherent `PyAny::as_ptr` combined with `Deref<Target=PyAny>` not serve the same purpose without inventing a new contract via this trait?

alex: Yes, this probably has a further requirement on the refcount (notably that this borrows a reference from `self`, whereas `into` shoudl be an owned reference). I'll update.

I'm not sure to what extent we could use the deref in place of this trait.

adamreichold: > I'm not sure to what extent we could use the deref in place of this trait.

My idea is to go from `&self` to `&PyAny` via `Deref<Target=PyAny>` which is unambiguously safe due to the validity requirements of `&PyAny`. From there we can go to `*mut ffi::PyObject` via `self.0.get()` without relying on extra contracts which must be upheld by implementors.

I the main question is whether we use this in any situation where `Deref<Target=PyAny>` does not apply. I would be surprised if we did tough, because returning a valid pointer to a Python object basically means I can construct a sufficiently short-lived `&PyAny` from that pointer.

davidhewitt: I think `PyAny::is` and also `__traverse__` are the only public APIs which immediately come to my mind, but we can find nice solutions to those I'm sure.

I would definitely like to start by seeing how far adding the inherent methods to `PyAny` and `Py<T>` gets us, and see if there's much work after that to replace and deprecate the traits entirely.

alex: This makes sense to me. I'm not going to have time to work on it today. My suggestion would be: I think this makes sense to merge as is, and then looking to minimize and deprecate the usage of these traits is really orthagonal.

davidhewitt: See #3359 as a starting point.

My only reason to wait on merging this PR here is that if we can get to a point where the traits are deprecated in 0.20 then I'd prefer not to change them at all before that release.

alex:
At this point here are the functions that take `impl AsPyPointer` as an argument:

- [ ] `Py::is`
- [ ] `PyAny::is`
- [ ] `PyIterator::from_object`
- [ ] `PyByteArray::from`
- [ ] `PyVisit::call`
- [ ] `impl From for PyObject`
- [ ] `impl IntoPy<PyObject> for &T`

I think the first four can be straightly switched to something else (e.g. `IntoPy<PyObject>`) and then we can call the inherent method.

The conversion `impls` can likely be removed or replaced with impls for each of the concrete types.

`PyVisit::call` is the one I'm least confident about. I don't think using `IntoPy<PyObject>` is quite right, since we don't really want to call back to random code there, which many implementors do. Any thoughts?

davidhewitt:
`PyAny::is` and `Py::is` don't dereference the pointer, so I believe they are safe without adding constraints to `AsPyPointer`. We could consider leaving the trait and use it just for them. To deprecate the trait they could be changed to take `&PyAny` (breaking change) or some trait implemented for `&PyAny` and `&Py<T>`.

`PyIterator::from_object` and `PyByteArray::from` should probably just take a single argument `&'py PyAny` rather than `py: Python<'py>` plus `AsPyPointer`. That's breaking but a consistency improvement anyway. I also suggest changing `PyDict::from_sequence` at the same time, which currently takes `py: Python<'py>` and `PyObject`.

Agreed re the conversion impls.

`PyVisit::call` will need to be usable without the GIL, so definitely can't use `IntoPy<PyObject>`. Because it doesn't have access to the GIL I think the only types which will be passed at that callsite will be `&Py<T>` and `&Option<Py<T>>`, so I suggest we deprecate `call` and add `PyVisit::visit(&Py<T>)` and `PyVisit::visit_option(&Option<T>)`. (That's breaking enough that I think it warrants a deprecation of at least one release.)

src/conversion.rs
```diff
@@ -56,13 +56,15 @@ use std::ptr::NonNull;
 /// and the Python object is dropped immediately after the `0xabad1dea_u32.into_py(py).as_ptr()`
 /// expression is evaluated. To fix the problem, bind Python object to a local variable like earlier
 /// to keep the Python object alive until the end of its scope.
-pub trait AsPyPointer {
+///
+/// Implementors must ensure this returns a valid pointer to a Python object, which borrows a reference count from `&self`.
```
mejrs, line 60: This pointer may also be null, like in the case of the implementation of `Option<T>`

alex:
Realizing that I missed one on my original list: `impl FromPyObject for Py<T>`, which I'm having trouble writing without `AsPyPointer`, would love some thoughts.

adamreichold:
> Realizing that I missed one on my original list: `impl FromPyObject for Py<T>`, which I'm having trouble writing without `AsPyPointer`, would love some thoughts.

This handles the different between going for `&T` and `&PyCell<T>` as the target, right? Maybe is #2956 helpful to resolve this?

alex:
It handles going from `&PyAny` to `Py<T>`, I'm not positive if that PR would help or not.

davidhewitt:
Hmm I had a go in #3406 which looks a possible route.

davidhewitt:
Given we're not sure about the decision in #3406, I'm beginning to warm to the idea to merging this now and releasing 0.20 with the remaining points addressed by the `unsafe` bound. We can then make further progress in removing this trait with the API reworks of 0.21?

alex:
That works for me.

davidhewitt:
Ok, let's do that. I'll create a follow up issue to finish the technical of this trait as we go.

PR #3359, opened 2023-07-30
add as_ptr and into_ptr inherent methods

davidhewitt:
Ref #3325.

Adds `as_ptr` and `into_ptr` methods to `Py`, `PyAny`, `PyRef`, and `PyRefMut`.

Between these four types, we can bin off a significant amount of PyO3's internal usage of `AsPyPointer` and `IntoPyPointer`, and this is probably also true for downstream.

Future PRs can figure out the remaining APIs which interact with these traits.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -356,6 +356,34 @@ where
     }
 }
 
+impl<T> Py<T> {
+    /// Returns the raw FFI pointer represented by self.
+    ///
+    /// # Safety
+    ///
+    /// Callers are responsible for ensuring that the pointer does not outlive self.
+    ///
+    /// The reference is borrowed; callers should not decrease the reference count
+    /// when they are finished with the pointer.
+    #[inline]
+    pub fn as_ptr(&self) -> *mut ffi::PyObject {
+        self.0.as_ptr()
+    }
+
+    /// Returns an owned raw FFI pointer represented by self.
+    ///
+    /// # Safety
+    ///
+    /// The reference is owned; when finished the caller should either transfer ownership
+    /// of the pointer or decrease the reference count (e.g. with [`pyo3::ffi::Py_DecRef`](crate::ffi::Py_DecRef)).
+    #[inline]
+    pub fn into_ptr(self) -> *mut ffi::PyObject {
+        let ptr = self.0.as_ptr();
+        std::mem::forget(self);
+        ptr
+    }
+}
+
 impl<T> Py<T>
 where
     T: PyClass,
@@ -897,7 +925,7 @@ impl<T> IntoPy<PyObject> for Py<T> {
     }
 }
 
-impl<T> AsPyPointer for Py<T> {
+impl<T> crate::AsPyPointer for Py<T> {
     /// Gets the underlying FFI pointer, returns a borrowed pointer.
     #[inline]
     fn as_ptr(&self) -> *mut ffi::PyObject {
--- a/src/pycell.rs
+++ b/src/pycell.rs
@@ -603,7 +603,6 @@ impl<T: PyClass + fmt::Debug> fmt::Debug for PyCell<T> {
 ///
 ///     fn format(slf: PyRef<'_, Self>) -> String {
 ///         // We can get *mut ffi::PyObject from PyRef
-///         use pyo3::AsPyPointer;
 ///         let refcnt = unsafe { pyo3::ffi::Py_REFCNT(slf.as_ptr()) };
 ///         // We can get &Self::BaseType by as_ref
 ///         let basename = slf.as_ref().basename;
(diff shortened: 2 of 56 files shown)


src/instance.rs
```diff
@@ -356,6 +356,36 @@ where
...
+    #[inline]
+    pub fn as_ptr(&self) -> *mut ffi::PyObject {
+        self.0.as_ptr()
+    }
+
+    /// Returns an owned raw FFI pointer represented by self.
+    ///
+    /// # Safety
+    ///
+    /// Callers are responsible for ensuring that the pointer does not outlive self.
```
alex, line 377: I don't think this is correct?

src/types/any.rs
```diff
@@ -1065,6 +1063,33 @@ impl PyAny {
...
+    pub fn as_ptr(&self) -> *mut ffi::PyObject {
+        // Safety: &PyAny has the same layout as *mut ffi::PyObject
+        unsafe { std::mem::transmute(self) }
+    }
+
+    /// Returns an owned raw FFI pointer represented by self.
+    ///
+    /// # Safety
+    ///
+    /// Callers are responsible for ensuring that the pointer does not outlive self.
```
alex, line 1084: Not right :-)

src/instance.rs
davidhewitt: Thanks yes that's wrong based on copypasta from `as_ptr` 🙈

src/instance.rs
```diff
@@ -1,13 +1,13 @@
...
 use crate::pycell::{PyBorrowError, PyBorrowMutError, PyCell};
 use crate::pyclass::boolean_struct::{False, True};
 use crate::types::{PyDict, PyString, PyTuple};
 use crate::{
-    ffi, AsPyPointer, FromPyObject, IntoPy, IntoPyPointer, PyAny, PyClass, PyClassInitializer,
-    PyRef, PyRefMut, PyTypeInfo, Python, ToPyObject,
+    ffi, FromPyObject, IntoPy, IntoPyPointer, PyAny, PyClass, PyClassInitializer, PyRef, PyRefMut,
+    PyTypeInfo, Python, ToPyObject,
 };
+use crate::{gil, AsPyPointer};
```
adamreichold, line 10: This looks unintentional?

src/pycell.rs
```diff
@@ -638,6 +637,34 @@ where
...
+    /// Returns the raw FFI pointer represented by self.
+    ///
+    /// # Safety
+    ///
+    /// Callers are responsible for ensuring that the pointer does not outlive self.
+    ///
+    /// The reference is borrowed; callers should not decrease the reference count
+    /// when they are finished with the pointer.
+    #[inline]
+    pub fn as_ptr(&self) -> *mut ffi::PyObject {
```
adamreichold, line 650: I wonder if `PyRef` should `Deref` into `PyAny` via forwarding to its inner `PyCell` field instead as that more generally useful?

src/instance.rs
davidhewitt: Yes, I removed `AsPyPointer` from line 7 but then found there are several implementations in this file which use it, so left it here for now. I'll revert this diff to just leave `AsPyPointer` imported on line 7.

src/pycell.rs
davidhewitt: Unfortunately `PyRef<T>` already implements `Derek<Target = T>`, which is extremely useful 😉

adamreichold: Indeed. How about `AsRef<PyAny>` as alternative? (Personally, I aim for `AsRef<T>` as well if I have `Deref<Target=T>` which is usually not problematic as one can have multiple `AsRef` impls.)

(My reasoning here is that I would like to centralize the pointer conversion as much as possible at `Py` and `PyAny`.)

davidhewitt: I'll give that a try and see what breaks 😅

davidhewitt: So I started down this road; here's a reason not to add `AsRef<PyAny>` for now:

The implementation would be

```
impl<T: PyClass> AsRef<PyAny> for PyRef<'_, T> {
    fn as_ref(&self) -> &PyAny {
        unsafe { self.py().from_borrowed_ptr(self.as_ptr()) }
    }
}
```

and the problem is that the `&PyAny` there has a much shortened lifetime, it's the duration of the `.as_ref()` borrow rather than the GIL lifetime `'p` attached to the `PyRef`.

Given this isn't an ideal UX, I would like to take the position that we don't implement `AsRef<PyAny>` for `PyRef` or `PyRefMut` for now. Can I request we move ahead here by just keeping the inherent methods and then perhaps reconsider once we have migrated to `Py<'py, PyAny>`?

Presumably then we can have the following implementation which would have a better lifetime

```
impl<'p, T: PyClass> AsRef<Py<'p, PyAny>> for PyRef<'p, T> {
    fn as_ref(&self) -> &Py<'p, PyAny> {
        // self.inner would be &'p Py<'p, PyCell<T>> or similar, I imagine
        self.inner.as_ref()
    }
}
```

(In the long run we might want to have `PyRef<'py, 'cell, T>` or maybe `Py<'py, PyRef<'cell, T>>` to capture that there are two distinct lifetimes, that of the cell borrow and that of the GIL lifetime. Though that's an an API design question for another time.)

adamreichold (approved):

PR #3363, opened 2023-08-03
PyBuffer/Py<T>: Data race allowed on T

kuzeyardabulut:
Hi,
I found a memory-safety/soundness issue in this crate while scanning Rust code for potential vulnerabilities. This PR contains a fix for the issue.

# Issue Description
`PyBuffer<T>` unconditionally implements Sync. This allows users to create data races on `T: !Sync`. Such data races can lead to undefined behavior.

https://github.com/PyO3/pyo3/blob/ae982b8ad0fff71d4d2fdbc2f7055500a6fb450f/src/buffer.rs#L37-L38

same
https://github.com/PyO3/pyo3/blob/ae982b8ad0fff71d4d2fdbc2f7055500a6fb450f/src/instance.rs#L229-L230

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -34,8 +34,8 @@ pub struct PyBuffer<T>(Pin<Box<ffi::Py_buffer>>, PhantomData<T>);
 
 // PyBuffer is thread-safe: the shape of the buffer is immutable while a Py_buffer exists.
 // Accessing the buffer contents is protected using the GIL.
-unsafe impl<T> Send for PyBuffer<T> {}
-unsafe impl<T> Sync for PyBuffer<T> {}
+unsafe impl<T: Send> Send for PyBuffer<T> {}
+unsafe impl<T: Sync> Sync for PyBuffer<T> {}
 
 impl<T> Debug for PyBuffer<T> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -226,8 +226,8 @@ pub struct Py<T>(NonNull<ffi::PyObject>, PhantomData<T>);
 // The inner value is only accessed through ways that require proving the gil is held
 #[cfg(feature = "nightly")]
 unsafe impl<T> crate::marker::Ungil for Py<T> {}
-unsafe impl<T> Send for Py<T> {}
-unsafe impl<T> Sync for Py<T> {}
+unsafe impl<T: Send> Send for Py<T> {}
+unsafe impl<T: Sync> Sync for Py<T> {}
 
 impl<T> Py<T>
 where


alex:
Can you provide a reproducer that demonstrates the issue for each of these types?

adamreichold:
I think the comments mention what the reasoning behind those impls is

```rust
// PyBuffer is thread-safe: the shape of the buffer is immutable while a Py_buffer exists.
// Accessing the buffer contents is protected using the GIL.
```

i.e. they are unconditionally `Send + Sync` because one needs to hold the GIL to access them.

Dr-Emann:
That said, at least the PyBuffer ones are only meaningful for `T: Element`, which is implemented only for a bunch of builtin integer types, all of which are Send+Sync, I don't know if we'd really lose anything by limiting PyBuffer to only be send/sync if T is send/sync.

adamreichold:
> That said, at least the PyBuffer ones are only meaningful for `T: Element`, which is implemented only for a bunch of builtin integer types, all of which are Send+Sync, I don't know if we'd really lose anything by limiting PyBuffer to only be send/sync if T is send/sync.

This is for now, there is work underway to support record, struct and array types, c.f. https://github.com/PyO3/rust-numpy/discussions/321 and https://github.com/aldanor/pyo3-type-desc, meaning that this will support user-defined types on which we should not impose unnecessary bounds.

On a more general note, I am not happy with the OP: This is based on static analysis with apparently little actual understanding of the code. Issues with similar quality problems where opened in quick succession in multiple Rust projects, including even one invalid Rustsec PR. And here, no proof of concept is forthcoming which would show how this would actually be exploitable. Hence, I suspect that this security circus aimed at gaining profile, not a genuine effort to improve the soundness properties of PyO3.

(If no proof of concept is posted, I will close this. I will probably wait until the at the end of next week to avoid rushing this as I cannot rule out that there are still soundness holes and I want to do our due diligence as maintainers.)

kuzeyardabulut:
> This is for now, there is work underway to support record, struct and array types, c.f. [PyO3/rust-numpy#321](https://github.com/PyO3/rust-numpy/discussions/321) and https://github.com/aldanor/pyo3-type-desc, meaning that this will support user-defined types on which we should not impose unnecessary bounds.
> 
> On a more general note, I am not happy with the OP: This is based on static analysis with apparently little actual understanding of the code. Issues with similar quality problems where opened in quick succession in multiple Rust projects, including even one invalid Rustsec PR. And here, no proof of concept is forthcoming which would show how this would actually be exploitable. Hence, I suspect that this security circus aimed at gaining profile, not a genuine effort to improve the soundness properties of PyO3.
> 
> (If no proof of concept is posted, I will close this. I will probably wait until the at the end of next week to avoid rushing this as I cannot rule out that there are still soundness holes and I want to do our due diligence as maintainers.)

Here we are just trying to make this code more secure. If you think there is a problem with this fix, you can report it by replying. In this way, I will know if I have made a mistake.

What you said is correct; I'm trying to find and report potential vulnerabilities in many projects using static analysis. If you think that there is no vulnerability and that a report has been made in error, you comment below it. You have done this in many reports.

My goal here is to make the code more secure and to apply best practices. Not all insecure codes can have a PoC (Proof of Concept) made. If I find a vulnerability in a piece of code where a PoC can be made, I will report it with the PoC.

Moreover, many vulnerabilities in Rust cannot be exploited. This is what makes C and Rust different.

adamreichold:
> Here we are just trying to make this code more secure. If you think there is a problem with this fix, you can report it by replying. In this way, I will know if I have made a mistake.
>
> What you said is correct; I'm trying to find and report potential vulnerabilities in many projects using static analysis. If you think that there is no vulnerability and that a report has been made in error, you comment below it. You have done this in many reports.

There have already been two replies: A request for a proof of concept in https://github.com/PyO3/pyo3/pull/3363#issuecomment-1664294558 with which I wholeheartedly agree and my point that the comment already indicates with these bounds are sound in https://github.com/PyO3/pyo3/pull/3363#issuecomment-1664294558. Neither of which has gotten a response so far.

> My goal here is to make the code more secure and to apply best practices. Not all insecure codes can have a PoC (Proof of Concept) made. If I find a vulnerability in a piece of code where a PoC can be made, I will report it with the PoC.
> 
> Moreover, many vulnerabilities in Rust cannot be exploited. This is what makes C and Rust different.
> 
> Frankly, I don't understand what's bothering you here. I am submitting a PR to both improve myself and contribute to the project. If you assert and explain that there is no weakness here, we all learn something valuable, ensuring that no aspect of the project is overlooked.

Because this PR and the other issues in other projects suggest that you did not bother to understand the code at all. Hence to me, this looks like creating a lot of work for the involved maintainers with little gain for the projects involved, hence the moniker "security circus."

PyO3 interfaces two programming languages and this might be a hint that things are more complicated than general guidelines for Rust code. For example, the docs for [`Py`](https://docs.rs/pyo3/latest/pyo3/struct.Py.html) already indicate that access to the inner value is tied to the GIL which enforces synchronization where pure Rust code would rely on `Send` and `Sync`.

What you should have done in my opinion, is reach out to us, for example via the discussions feature and started with a question: "I have done static analysis and it flagged the unconditional impls. I don't know the code base and I would like to understand whether this might be a problem or not. Could you explain the reasoning behind them?"

adamreichold:
> Moreover, many vulnerabilities in Rust cannot be exploited. This is what makes C and Rust different.

This is wrong: Vulnerabilities are exploitable by definition. Not every bug, not every instance of undefined behaviour is exploitable/is a vulnerability though. This is the same for Rust and C.

PR #3366, opened 2023-08-03
Implement DoubleEndedIterator for PyTupleIterator

adamreichold:


src/types/mod.rs
```diff
@@ -75,6 +75,7 @@ pub mod iter {
     pub use super::dict::PyDictIterator;
     pub use super::frozenset::PyFrozenSetIterator;
     pub use super::set::PySetIterator;
+    pub use super::tuple::PyTupleIterator;
```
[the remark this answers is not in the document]
davidhewitt: Potentially worth calling out in a `3366.fixed.md`?

src/buffer.rs
```diff
@@ -495,7 +495,7 @@ impl<T: Element> PyBuffer<T> {
 
         err::error_on_minusone(py, unsafe {
             ffi::PyBuffer_ToContiguous(
-                target.as_ptr() as *mut raw::c_void,
+                target.as_mut_ptr() as *mut raw::c_void,
```
davidhewitt, line 498: I think `.cast()` is generally recommended these days to avoid `const` -> `mut` conversions.

```suggestion
                target.as_mut_ptr().cast(),
```

src/types/tuple.rs
```diff
@@ -513,6 +532,27 @@ mod tests {
...
+            assert_eq!(iter.size_hint(), (3, Some(3)));
+
+            assert_eq!(3_i32, iter.next().unwrap().extract::<'_, i32>().unwrap());
+            assert_eq!(iter.size_hint(), (2, Some(2)));
+
+            assert_eq!(2_i32, iter.next().unwrap().extract::<'_, i32>().unwrap());
+            assert_eq!(iter.size_hint(), (1, Some(1)));
+
+            assert_eq!(1_i32, iter.next().unwrap().extract::<'_, i32>().unwrap());
+            assert_eq!(iter.size_hint(), (0, Some(0)));
```
davidhewitt, line 552: Maybe confirm the iterator is empty?

```suggestion
            assert_eq!(iter.size_hint(), (0, Some(0)));
            
            assert!(iter.next().is_none());
```

davidhewitt (commented):
Thanks, I agree with the functionality! I guess the same for `list` is more tricky due to possible mutation during iteration? (Though we guard against that IIRC.)

Just a few review comments...

src/buffer.rs
adamreichold: (I'll change this to use `cast`, but the fix was the function taking `target: &mut T` but not using it mutably.)

davidhewitt (approved):
Great! Looks good to me 👍

PR #3040, opened 2023-03-11
mangle exported functions for PyPy

mattip:
Continuation of #3031. This [turned up](https://github.com/pypy/binary-testing/actions/runs/4390086363/jobs/7688256133#step:6:179) in the run of PyPy 3.10 HEAD against PyO3 HEAD. Should I add a new news fragment or is the one from #3031 sufficient?

PR #3374, opened 2023-08-08
Makes PathBuf FromPyObject implementation work on all os.PathLike

Tpt:
PyOS_FSPath is in abi3-py36


guide/src/conversions/tables.md
```diff
@@ -37,6 +37,7 @@ The table below contains the Python type and the corresponding function argument
 | `decimal.Decimal` | `rust_decimal::Decimal`[^5] | -                    |
 | `ipaddress.IPv4Address` | `std::net::IpAddr`, `std::net::IpV4Addr` | - |
 | `ipaddress.IPv6Address` | `std::net::IpAddr`, `std::net::IpV6Addr` | - |
+| `os.PathLike ` | `PathBuf`, `Path`              | `&PyString`, `&PyUnicode` |
 | `pathlib.Path` | `PathBuf`, `Path`              | `&PyString`, `&PyUnicode` |
```
adamreichold, line 41: Isn't `pathlib.Path` also an `os.PathLike`, i.e. doesn't the first line basically include the second one?

Tpt: Indeed. I left the two lines for explictness because I assumed a lot of people are not used to `os.PathLike`. If you prefer I can remove the `pathlib.Path` line.

src/conversions/std/path.rs
```diff
@@ -14,21 +16,9 @@ impl ToPyObject for Path {
...
-                    let path_str = ob.call_method0(intern!(py, "__str__"))?;
-                    OsString::extract(path_str)?
-                } else {
-                    return Err(err);
-                }
-            }
-        };
-        Ok(PathBuf::from(os_str))
+        // We use os.fspath to get the underlying path as bytes or str
+        let path = unsafe { PyAny::from_owned_ptr(ob.py(), ffi::PyOS_FSPath(ob.as_ptr())) };
```
adamreichold, line 20: I think this is missing error handling as `PyOS_FSPath` might return a null point and raise a `TypeError`, i.e. I think this should at least use `from_owned_ptr_or_err`.

pytests/tests/test_path.py
```diff
@@ -16,3 +16,15 @@ def test_take_pathbuf():
 def test_take_pathlib():
     p = pathlib.Path("/root")
     assert rpath.take_pathbuf(p) == str(p)
+
+
+def test_take_pathlike():
```
adamreichold, line 21: I think it would also be good to test the case where the argument is not path-like and where it is path-like but `__fspath__` does not return `str` or `bytes`.

guide/src/conversions/tables.md
adamreichold: Personally, I would merge them and maybe use "`pathlib.Path`, `os.PathLike`" for the first column to stay explicit. But let's see what the other maintainers think.

src/conversions/std/path.rs
Tpt: Indeed. My bad. Fixed and tested.

pytests/tests/test_path.py
Tpt: Done.

adamreichold (approved):

guide/src/conversions/tables.md
davidhewitt: I was thinking of upending this table anyway and describing it from the Rust types (what they accept and what they convert to), so I'm ok with leaving this as two rows for now.

davidhewitt (approved):
Thanks - please accept the new licensing change in #3108 before we can merge this.

Tpt:
Pypy 3.10 fails with "undefined reference to `PyOS_FSPath'". What is strange is that it's supposed to be supported by Pypy 3.7+: https://doc.pypy.org/en/latest/release-v7.3.0.html#python-3-6-c-api

adamreichold:
Is it possible that we need to link this under a different name using a `PyPy` prefix (we do this for many other symbols, c.f. e.g. #3040)? cc @mattip

davidhewitt:
Yes, on all supported PyPy versions it looks like it is exported under the `PyPy` prefix: https://github.com/PyO3/python3-dll-a/blob/main/src/libpypy3-c.def#L449

(I would love to catch cases where our definition is wrong in `pyo3-ffi-check`, but so far I didn't work out a way to do this, probably `rustdoc` metadata is insufficient to do so.)

Tpt:
Thank you! I have added the `link_name` attribute for Pypy.

> (I would love to catch cases where our definition is wrong in pyo3-ffi-check, but so far I didn't work out a way to do this, probably rustdoc metadata is insufficient to do so.)

Dumb idea: if Pyo3 is supposed to cover all/most symbols expose al `PyPy*` symbol one can just grep each of them to check if there are present in PyO3 source code. This might allow to cover link names.

PR #3378, opened 2023-08-11
add `__builtins__` to globals in `py.run()` if they're missing

### LINKED ISSUE #3370: Python 3.{8,9}: `NameError: __build_class__ not found` when globals are provided to `py.run()`
### Bug Description

Passing `Some(_)` as `globals` to `py.run()` breaks Python builtins under Python 3.8 and 3.9 (and maybe others, but not 3.10).

### Steps to Reproduce

Run the following code (linking with Python 3.8 or Python 3.9):
```rust
fn main() -> eyre::Result<()> {
    let script_text = r#"
def trivial(s):
    return s
class Foo:
    def __init__(self):
        print(trivial('Hello, world!'))
    "#;

    pyo3::Python::with_gil(|py| {
        let module = pyo3::types::PyDict::new(py);
        py.run(script_text, Some(module), Some(module))?;
        let foo_class = module.get_item("Foo").unwrap();
        let _foo = foo_class.call0()?;
        Ok(())
    })
}
```

Observe the following error:
```
NameError: __build_class__ not found
```

NixOS

Python 3.8.17, Python 3.9.17

rustc 1.73.0-nightly (8131b9774 2023-08-02)

### Your PyO3 version

0.19.2

Nix, no virtualenv

### Additional Info

Possibly related to #3371, #2891

GoldsteinE:
Python code doesn't like to run without `__builtins__`, so adding them if missing seems to be a good idea. Also that's what CPython >3.10 does.

See https://github.com/python/cpython/pull/24564
Resolves #3370

* [x] Agreed to dual-license (https://github.com/PyO3/pyo3/pull/3108#issuecomment-1674518411)
* [x] Entry in newsfragments
* [x] Tests for changed functions

### DIFF
--- a/newsfragments/3378.changed.md
+++ b/newsfragments/3378.changed.md
@@ -1,0 +1,1 @@
+Add `__builtins__` to globals in `py.run()` and `py.eval()` if they're missing.
--- a/src/marker.rs
+++ b/src/marker.rs
@@ -544,6 +544,9 @@ impl<'py> Python<'py> {
     /// If `globals` is `None`, it defaults to Python module `__main__`.
     /// If `locals` is `None`, it defaults to the value of `globals`.
     ///
+    /// If `globals` doesn't contain `__builtins__`, default `__builtins__`
+    /// will be added automatically.
+    ///
     /// # Examples
     ///
     /// ```
@@ -567,6 +570,9 @@ impl<'py> Python<'py> {
     ///
     /// If `globals` is `None`, it defaults to Python module `__main__`.
     /// If `locals` is `None`, it defaults to the value of `globals`.
+    ///
+    /// If `globals` doesn't contain `__builtins__`, default `__builtins__`
+    /// will be added automatically.
     ///
     /// # Examples
     /// ```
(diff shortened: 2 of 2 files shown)


src/marker.rs
```diff
@@ -632,6 +632,29 @@ impl<'py> Python<'py> {
...
+            let has_builtins = ffi::PyDict_Contains(globals, builtins_s);
+            if has_builtins == -1 {
+                return Err(PyErr::fetch(self));
+            }
+            if has_builtins == 0 {
+                // Inherit current builtins.
+                let builtins = ffi::PyEval_GetBuiltins();
+
+                // `PyDict_SetItem` doesn't take ownership of `builtins`, but `PyEval_GetBuiltins`
+                // seems to return a borrowed reference, so no leak here.
```
davidhewitt, line 652: 👍

davidhewitt (commented):
Many thanks, implementation looks great. My only thought: after evaluation should we attempt to remove `__builtins__` from the `globals` dictionary if we added it?

GoldsteinE:
> Many thanks, implementation looks great. My only thought: after evaluation should we attempt to remove __builtins__ from the globals dictionary if we added it?

I think that’s more confusing than the current option (especially given possibility of code like `__builtins__ = new_value`). CPython uses some private APIs to specify `__builtins__` for the new frame explicitly instead of adding them to `globals`, so we can’t easily mimic its behaviour here (in particular, we’d need `_PyFunction_FromConstructor`: I don’t see a way to manually specify `__builtins__` otherwise).

newsfragments/3378.changed.md
```diff
@@ -0,0 +1 @@
+Add `__builtins__` to globals in `py.run()` if they're missing.
```
davidhewitt, line 1:
```suggestion
Add `__builtins__` to globals in `py.run()` and `py.eval()` if they're missing.
```

davidhewitt (commented):
Agreed, in which case I think let's just document that `globals` will be modified in this way (on both `Python::run` and `Python::eval`) and call it a day.

newsfragments/3378.changed.md
GoldsteinE: Applied manually.

GoldsteinE:
Fixed the docs.

davidhewitt (approved):
Thanks again!

PR #3379, opened 2023-08-11
BigInt speedup

iliya-malecki:
My contributions to pyo3 are available under the terms of either the Apache 2.0 OR the MIT license :)
This PR is a continuation of #3365 
I am very much unsure about some parts, feel free to comment on how it might be improved or ditched. The balance i was trying to strike is between speed and maintainability - however, whether i was successful isnt very clear-cut. I succeeded in not using pointer offsets and other ridiculous methods of reaching into pyobjects memory, though

### DIFF
--- a/src/conversions/num_bigint.rs
+++ b/src/conversions/num_bigint.rs
@@ -51,25 +51,37 @@ use crate::{
     ffi, types::*, FromPyObject, IntoPy, Py, PyAny, PyObject, PyResult, Python, ToPyObject,
 };
 
-use num_bigint::{BigInt, BigUint};
-use std::os::raw::{c_int, c_uchar};
+use num_bigint::{BigInt, BigUint, Sign};
+use std::os::raw::c_int;
 
 #[cfg(not(Py_LIMITED_API))]
-unsafe fn extract(ob: &PyLong, buffer: &mut [c_uchar], is_signed: c_int) -> PyResult<()> {
+use std::os::raw::c_uchar;
+
+#[cfg(Py_LIMITED_API)]
+use std::slice;
+
+#[cfg(not(Py_LIMITED_API))]
+#[inline]
+unsafe fn extract(ob: &PyLong, length: usize, is_signed: c_int) -> PyResult<Vec<u32>> {
+    let mut buffer = Vec::<u32>::with_capacity(length);
     crate::err::error_on_minusone(
         ob.py(),
         ffi::_PyLong_AsByteArray(
             ob.as_ptr() as *mut ffi::PyLongObject,
-            buffer.as_mut_ptr(),
-            buffer.len(),
+            buffer.as_mut_ptr() as *mut u8,
+            length * 4,
             1,
             is_signed,
         ),
-    )
+    )?;
+    buffer.set_len(length);
+
+    Ok(buffer)
 }
 
 #[cfg(Py_LIMITED_API)]
-unsafe fn extract(ob: &PyLong, buffer: &mut [c_uchar], is_signed: c_int) -> PyResult<()> {
+#[inline]
+unsafe fn extract(ob: &PyLong, length: usize, is_signed: c_int) -> PyResult<Vec<u32>> {
     use crate::intern;
     let py = ob.py();
     let kwargs = if is_signed != 0 {
@@ -81,14 +93,16 @@ unsafe fn extract(ob: &PyLong, buffer: &mut [c_uchar], is_signed: c_int) -> PyRe
     };
     let bytes_obj = ob
         .getattr(intern!(py, "to_bytes"))?
-        .call((buffer.len(), "little"), kwargs)?;
+        .call((length * 4, intern!(py, "little")), kwargs)?;
     let bytes: &PyBytes = bytes_obj.downcast_unchecked();
-    buffer.copy_from_slice(bytes.as_bytes());
-    Ok(())
+    let bytes_u32 = slice::from_raw_parts(bytes.as_bytes().as_ptr().cast(), length);
+
+    Ok(bytes_u32.to_vec())
 }
 
+// for identical functionality between BigInt and BigUint
 macro_rules! bigint_conversion {
-    ($rust_ty: ty, $is_signed: expr, $to_bytes: path, $from_bytes: path) => {
+    ($rust_ty: ty, $is_signed: expr, $to_bytes: path) => {
         #[cfg_attr(docsrs, doc(cfg(feature = "num-bigint")))]
         impl ToPyObject for $rust_ty {
             #[cfg(not(Py_LIMITED_API))]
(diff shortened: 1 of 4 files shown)


iliya-malecki:
This is the speedup i achieved (note the slowdown for small numbers)
image

src/conversions/num_bigint.rs
```diff
@@ -138,53 +140,69 @@ macro_rules! bigint_conversion {
...
         }
     };
 }
-bigint_conversion!(BigUint, 0, BigUint::to_bytes_le, BigUint::from_bytes_le);
+bigint_conversion!(
+    BigUint,
+    0,
+    BigUint::to_bytes_le,
+    //from digits conversion
+    |digits: Vec<u32>| { BigUint::new(digits) }
```
davidhewitt, line 191: I think you don't need the closure here

```suggestion
    BigUint::new
```

davidhewitt (commented):
Thanks for the contribution!

So I've been reading through this and trying to get a sense of what might be the cause of the slowdown.

I noticed that internally `BigInt` seems to use either `u32` or `u64` according to the platform pointer size.

https://github.com/rust-num/num-bigint/blob/65f62a8b1484448bfb9789ef4123b50556254905/build.rs#L9-L12

https://github.com/rust-num/num-bigint/blob/65f62a8b1484448bfb9789ef4123b50556254905/src/biguint.rs#L527-L538

It looks to me that on 64-bit systems using `Vec<u32>` as an intermediary isn't beneficial, because this gets copied into a new `u64` representation. So the `Vec` gets dropped on the floor.

Given that, I'm a little surprised that we've got a speedup here. Have you got an intuition for which bits of the change yielded the performance boost? I see we now skip zero-initialization, I don't see any other clear wins...

src/conversions/num_bigint.rs
```diff
@@ -138,53 +140,69 @@ macro_rules! bigint_conversion {
...
-                        extract(num.as_ref(py), &mut buffer, $is_signed)?;
-                        Ok($from_bytes(&buffer))
+                    if n_bits == 0 {
+                        return Ok(0.try_into().unwrap())
                     }
+
+                    let n_digits = (n_bits + 31 + $is_signed) / 32;
+
+                    let mut buffer = Vec::<u32>::with_capacity(n_digits);
+                    buffer.set_len(n_digits);
```
adamreichold, line 168: I think this basically instant UB because it assumes initialization for data that is clear uninitialized.

Either we need to use raw pointers everywhere or something like [`spare_capacity_mut`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.spare_capacity_mut) if our MSRV allows it. In any case, we cannot call `set_len` until the buffer has actually been initialized.

davidhewitt: Ah, that's a very good catch 👍

davidhewitt: We could probably call `set_len` _after_ extract, and just pass a raw pointer to `extract`

iliya-malecki: im beyond puzzled. I Changed `buffer.len()` into a `length` parameter to the function, and when i use the function
```
let mut buffer = Vec::<u32>::with_capacity(n_digits);
                    extract(
                        num.as_ref(py),
                        &mut buffer,
                        n_digits,
                        $is_signed
                    )?;
                    buffer.set_len(n_digits);
```
it just produces wrong results. If i `set_len` before the call to extract in the abi3 mode, it works fine. If i zero-init the vector, it works fine. But if i leave the data uninit and dont set the length (bot of those things!), it produces noncence. Am i missing something here?

davidhewitt: In `abi3` mode there's a call to `buffer.copy_from_slice` so unless you changed that I would expect that to do nothing to a zero-length vector.

You could just sidestep the problem and call `BigInt::from_signed_bytes_le` (or similar for `BigUint`) on the bytes slice from the Python `to_bytes` call, though that will require a bit more refactoring.

adamreichold: > Am i missing something here?

Difficult to tell without looking at the code, but I suspect you still take `&mut [u32]`, but this will be a zero-length slice if you do not zero-initialize the buffer. As suggested, you should take a raw pointer instead. (A slice `&mut [u32]` also _requires_ the corresponding memory to be initialized to a valid value.) 

(Note that in https://github.com/PyO3/pyo3/discussions/3365#discussioncomment-6630845 I mentioned that skipping zero-initialization would make this messy. This is what I was talking about.)

iliya-malecki: thank you all, your crystall ball omniscience is impressive!

src/conversions/num_bigint.rs
iliya-malecki: i just assumed BigUint::new doesnt copy data, but num_bigint::u32_to_u64 function is clearly doing it, and in a funny way as well. So now im just confused about the speedup, mostly...

iliya-malecki: however, now im dreaming of the day num_bigint gets fixed and the speedup is massive! i think i know my next PR

src/conversions/num_bigint.rs
```diff
@@ -82,14 +88,15 @@ unsafe fn extract(ob: &PyLong, buffer: &mut [c_uchar], is_signed: c_int) -> PyRe
     };
     let bytes_obj = ob
         .getattr(intern!(py, "to_bytes"))?
-        .call((buffer.len(), "little"), kwargs)?;
+        .call((length * 4, "little"), kwargs)?;
     let bytes: &PyBytes = bytes_obj.downcast_unchecked();
-    buffer.copy_from_slice(bytes.as_bytes());
-    Ok(())
+    let bytes_u32 = std::slice::from_raw_parts(bytes.as_bytes().as_ptr() as *const u32, length);
```
adamreichold, line 93: Nits: Could you `use` either `slice` or `from_raw_parts` and use `.cast()` to change the pointer type?

src/conversions/num_bigint.rs
```diff
@@ -82,14 +88,15 @@ unsafe fn extract(ob: &PyLong, buffer: &mut [c_uchar], is_signed: c_int) -> PyRe
     };
     let bytes_obj = ob
         .getattr(intern!(py, "to_bytes"))?
-        .call((buffer.len(), "little"), kwargs)?;
+        .call((length * 4, "little"), kwargs)?;
```
adamreichold, line 91: Shouldn't "little" be interned as well?

src/conversions/num_bigint.rs
```diff
@@ -138,53 +145,66 @@ macro_rules! bigint_conversion {
...
+    BigUint::new
+);
 bigint_conversion!(
     BigInt,
     1,
     BigInt::to_signed_bytes_le,
-    BigInt::from_signed_bytes_le
+    //from digits conversion
+    |mut digits: Vec<u32>| {
+        if digits.last().unwrap() >> 31 != 0 {
```
adamreichold, line 201: I wonder if this loop could be merged with the `u32::from_le` loop based on the value of `$is_signed` to avoid stashing code down here?

iliya-malecki: I don't quite know how, since this loop is for negative to BigUint conversion

adamreichold: But it is applied only if `$is_signed` isn't? So `$from_digits` could always be `BigUint::new`/`BigInt::new` and the conversion could happen inside the code using e.g.

```rust
if $is_signed && digits.last().unwrap() >> 31 != 0 {
  buffer.iter().for_each(|chunk| *chunk = !u32::from_le(*chunk));
} else {
  buffer.iter().for_each(|chunk| *chunk = u32::from_le(*chunk));
}
```

giving us a single loop and the option to pipeline the negation and byte swap.

iliya-malecki: I guess it could be
```
if digits.last().unwrap() >> 31 != 0 {
  if $is_signed {
    buffer.iter().for_each(|chunk| *chunk = !u32::from_le(*chunk));
  } else {
    buffer.iter().for_each(|chunk| *chunk = u32::from_le(*chunk));
  }
}
```
But then how would I add/subtract one to the BigInt to finish the conversion from two's complement?

adamreichold: > But then how would I add/subtract one to the BigInt to finish the conversion from two's complement?

This is complicated to do on the digits because it might carry from the least up to the most significant digit, right?

iliya-malecki: Yup, exactly!

adamreichold: I know it is a bad habit when maintainers ask for unrelated clean-ups but since this is an optimization to begin, maybe can we avoid the macro entirely and use the same approach `num-bigint` internally, i.e. we directly implement `impl FromPyObject for BigUint` and then implement `impl FromPyObject for BigInt` in terms of that?

That would still leave us with two loops, but hopefully cleaner code that does not have to work in "pseudo-generic context."

iliya-malecki: Are you okay with some noticeable code duplication then? I would love to get rid of the macros and have a language server help me write code (it's pretty limited for macros) but that would lead to copypasting some functions for BigInt and BigUInt separately.

src/conversions/num_bigint.rs
iliya-malecki: sure!

src/conversions/num_bigint.rs
iliya-malecki: oh wait you meant only `impl FromPyObject` and not the whole thing... oops!

adamreichold: > oh wait you meant only impl FromPyObject and not the whole thing... oops!

That said, I would not mind a bit of code duplication to get completely rid of the macro either, as long as the meaty part which does the conversion to a Python object is factored out into a helper function similar to the existing `extract`. I think writing a bit of mindless boiler plate is a reasonable price to pay for avoiding the macro.

iliya-malecki: > I would not mind a bit of code duplication to get completely rid of the macro either
i couldnt come up with a sensible way to refactor it, all options smelled worse than the macro (to me personally, im a data scientist who mostly writes python so take it with a grain of salt), so i decided to leave it to people that are actually familiar with the code style and conventions. I did refactor the `impl FromPyObject` and left it at that

iliya-malecki:
im again puzzled, the patch should now have 100% coverage shouldnt it?

davidhewitt:
Looking at the [PR on codecov](https://app.codecov.io/gh/PyO3/pyo3/pull/3379) it's the error branches which have no coverage, which is often the case so not a huge surprise.

@iliya-malecki are you ok if I push some additional changes to this PR? I'm tempted to do a little further refactoring and see if I can get the small bigint performance back on par with the original implementation.

(And if you're interested in doing the upstream PR to `num-bigint` so that we could directly build the integer from the "native" u32 or u64 representation and avoid another Vec allocation then that would be a win for the whole ecosystem!)

iliya-malecki:
I would absolutely love that, I did this pr mainly to learn Rust so any input is very welcome

iliya-malecki:
Also, I'm looking at the codecov page but I'm not sure what I'm seeing, where does the value 15% (of uncovered lines) come from? From two lines of returning errors? Or is it a quirk that arises from those errors themselves not being covered anywhere?

src/conversions/num_bigint.rs
```diff
@@ -129,145 +99,215 @@ macro_rules! bigint_conversion {
...
+                long.as_ptr().cast(),
+                buffer.as_mut_ptr() as *mut u8,
+                n_digits * 4,
+                1,
+                is_signed.into(),
+            ),
+        )?;
+        buffer.set_len(n_digits)
+    };
+    buffer
```
adamreichold, line 202: One thing I wonder considering that the majority of users are probably on 64-bit architectures and even if we fix `num-bigint`, being able to rely on that fix will take a while, is whether using `Vec<u8>` with `_PyLong_AsByteArray` and `BigUint::from_bytes_le` might actually be faster in the 64-bit case?

adamreichold: (And might significantly simplify the signed case.)

iliya-malecki: I think the original solution was that, wasn't it?

adamreichold: > I think the original solution was that, wasn't it?

Well, with additional zero-initialization and one additional copy in the limited API case, right? But yes, I would be interesting to set if there is any speed up left in the 64-bit case when `Vec::with_capacity(n_digits)` is replaced by `vec![0; n_digits]`.

davidhewitt: main:

```
extract_bigint_huge_negative
                        time:   [2.0350 µs 2.0353 µs 2.0356 µs]
extract_bigint_huge_positive
                        time:   [1.5174 µs 1.5177 µs 1.5181 µs]
```

this branch + zero initialization

```
extract_bigint_huge_negative
                        time:   [1.1691 µs 1.1694 µs 1.1697 µs]
extract_bigint_huge_positive
                        time:   [1.1229 µs 1.1233 µs 1.1239 µs]
```

So it's not just skipping zero-initialisation giving the speedup. My intuition for what's going on (having peeked in the `num_bigint` code) is that the implementation we've written here to convert from u8 to u32 and also to do two's complement is better optimised than the upstream versions in these cases too.

For example, `BigInt::from_signed_bytes_le` takes a copy of the bytes to do two's complement on that, before then throwing that away to do a loop which uses a lot of shifting to merge u8 to u64 (where we effectively just transmute a Vec modulo endianness to achieve that conversion).

https://github.com/rust-num/num-bigint/blob/65f62a8b1484448bfb9789ef4123b50556254905/src/bigint/convert.rs#L408C44-L408C44
https://github.com/rust-num/num-bigint/blob/65f62a8b1484448bfb9789ef4123b50556254905/src/biguint/convert.rs#L43

davidhewitt (approved):
As far as I'm concerned, this is as good as we can do here. There looks like there's still more to be won with improvements upstream, however I think that's out of scope for me to worry about for now.

PR #3384, opened 2023-08-13
Add  `rename_all` attribute to `#[pyclass]`

### LINKED ISSUE #3383: Support renaming enum variants
It would be really convenient if the `#[pyclass]` attribute offered a way to rename all variants of an enum.

## The problem

By default the generated Python enums expose variants named after Rust's naming convention, which is not idiomatic for Python where SCREAMING_SNAKE_CASE is the norm. One can either create a new enum specifically tailored for PyO3, or apply `#[pyo3(name = "...")]` to every variants. Neither of these options are practical for big enums, and I can't make the second one work if I need to conditionally support PyO3: `#[cfg_attr(feature = "pyo3", pyo3(name = "...")]` won't compile.

## Solution

Add a new argument to `#[pyclass]` that would only apply to enums and that would rename all their variants: `rename_all` for instance.

It could be a good idea to also support `#[serde(rename_all = "...")]` if it is already applied to an enum, although I don't think it would happen often that people use screaming snake case when (de)serializing their types.

DataTriny:
Closes #3383

### DIFF
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -3,7 +3,7 @@ use std::borrow::Cow;
 use crate::attributes::kw::frozen;
 use crate::attributes::{
     self, kw, take_pyo3_options, CrateAttribute, ExtendsAttribute, FreelistAttribute,
-    ModuleAttribute, NameAttribute, NameLitStr, TextSignatureAttribute,
+    ModuleAttribute, NameAttribute, NameLitStr, RenameAllAttribute, TextSignatureAttribute,
     TextSignatureAttributeValue,
 };
 use crate::deprecations::{Deprecation, Deprecations};
@@ -14,9 +14,8 @@ use crate::pymethod::{
     impl_py_getter_def, impl_py_setter_def, MethodAndMethodDef, MethodAndSlotDef, PropertyType,
     SlotDef, __INT__, __REPR__, __RICHCMP__,
 };
-use crate::utils::{self, get_pyo3_crate, PythonDoc};
+use crate::utils::{self, apply_renaming_rule, get_pyo3_crate, PythonDoc};
 use crate::PyFunctionOptions;
-use heck::ToShoutySnakeCase;
 use proc_macro2::{Ident, Span, TokenStream};
 use quote::quote;
 use syn::ext::IdentExt;
@@ -67,7 +66,7 @@ pub struct PyClassPyO3Options {
     pub mapping: Option<kw::mapping>,
     pub module: Option<ModuleAttribute>,
     pub name: Option<NameAttribute>,
-    pub rename_all: Option<kw::rename_all>,
+    pub rename_all: Option<RenameAllAttribute>,
     pub sequence: Option<kw::sequence>,
     pub set_all: Option<kw::set_all>,
     pub subclass: Option<kw::subclass>,
@@ -88,7 +87,7 @@ enum PyClassPyO3Option {
     Mapping(kw::mapping),
     Module(ModuleAttribute),
     Name(NameAttribute),
-    RenameAll(kw::rename_all),
+    RenameAll(RenameAllAttribute),
     Sequence(kw::sequence),
     SetAll(kw::set_all),
     Subclass(kw::subclass),
@@ -118,7 +117,7 @@ impl Parse for PyClassPyO3Option {
             input.parse().map(PyClassPyO3Option::Module)
         } else if lookahead.peek(kw::name) {
             input.parse().map(PyClassPyO3Option::Name)
-        } else if lookahead.peek(attributes::kw::rename_all) {
+        } else if lookahead.peek(kw::rename_all) {
             input.parse().map(PyClassPyO3Option::RenameAll)
         } else if lookahead.peek(attributes::kw::sequence) {
             input.parse().map(PyClassPyO3Option::Sequence)
@@ -208,11 +207,6 @@ pub fn build_py_class(
             lt.span() =>
             "#[pyclass] cannot have lifetime parameters. \
             For an explanation, see https://pyo3.rs/latest/class.html#no-lifetime-parameters"
-        );
-    } else if let Some(rename_all) = args.options.rename_all {
-        bail_spanned!(
-            rename_all.span() =>
-            "rename_all should only be applied to enums."
         );
     }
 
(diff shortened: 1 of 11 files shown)


pyo3-macros-backend/src/pyclass.rs
```diff
@@ -203,6 +209,11 @@ pub fn build_py_class(
             "#[pyclass] cannot have lifetime parameters. \
             For an explanation, see https://pyo3.rs/latest/class.html#no-lifetime-parameters"
         );
+    } else if let Some(rename_all) = args.options.rename_all {
+        bail_spanned!(
+            rename_all.span() =>
+            "rename_all should only be applied to enums."
```
adamreichold, line 215: Couldn't this just as well be applied to generated getters and setters?

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -379,12 +390,21 @@ struct PyClassEnumVariant<'a> {
...
+        let name = self
+            .options
             .name
             .as_ref()
             .map(|name_attr| Cow::Borrowed(&name_attr.value.0))
-            .unwrap_or_else(|| Cow::Owned(self.ident.unraw()))
+            .unwrap_or_else(|| Cow::Owned(self.ident.unraw()));
+        if args.options.rename_all.is_some() {
+            Cow::Owned(Ident::new(
+                &format!("{}", name).to_shouty_snake_case(),
```
adamreichold, line 402: Considering [the analogy with serde](https://serde.rs/container-attrs.html#rename_all), I would have expected to specify the convention as an argument to `rename_all`.

I understand that this mainly targets making enum variants matching the Python conventions, but I wonder whether we should give this a more specific name then, e.g. `rename_variants`?

tests/ui/invalid_pyclass_args.stderr
```diff
@@ -1,4 +1,4 @@
-error: expected one of: `crate`, `dict`, `extends`, `freelist`, `frozen`, `get_all`, `mapping`, `module`, `name`, `sequence`, `set_all`, `subclass`, `text_signature`, `unsendable`, `weakref`
+error: expected one of: `crate`, `dict`, `extends`, `freelist`, `frozen`, `get_all`, `mapping`, `module`, `name`, `rename_all`, `sequence`, `set_all`, `subclass`, `text_signature`, `unsendable`, `weakref`
```
adamreichold, line 1: I think we also need a Python test which applies this and checks the resulting variant names.

DataTriny:
I'll try to update the `rename_all` attribute so that it accepts a case as argument and applies to struct fields as well, if there is interest in having this. Thanks for your feedback.

pyo3-macros-backend/src/pyclass.rs
DataTriny: `rename_all` now behaves exactly like serde's one. Keeping the name as is.

DataTriny:
@adamreichold Ready for another review.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -379,12 +389,20 @@ struct PyClassEnumVariant<'a> {
...
+    fn python_name(&self, args: &PyClassArgs) -> Cow<'_, syn::Ident> {
         self.options
             .name
             .as_ref()
             .map(|name_attr| Cow::Borrowed(&name_attr.value.0))
-            .unwrap_or_else(|| Cow::Owned(self.ident.unraw()))
+            .unwrap_or_else(|| {
+                let name = self.ident.unraw();
+                if let Some(attr) = &args.options.rename_all {
+                    let new_name = apply_renaming_rule(attr.value.rule, &format!("{}", name));
```
adamreichold, line 400: Isn't `&format("{}", name)` equivalent to `&name.to_string()`?

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -736,11 +737,21 @@ impl PropertyType<'_> {
...
+                renaming_rule,
+                ..
             } => {
                 let name = match (python_name, &field.ident) {
                     (Some(name), _) => name.value.0.to_string(),
-                    (None, Some(field_name)) => format!("{}\0", field_name.unraw()),
+                    (None, Some(field_name)) => {
+                        let name = format!("{}\0", field_name.unraw());
+                        if let Some(rule) = renaming_rule {
+                            utils::apply_renaming_rule(*rule, &name)
```
adamreichold, line 750: I am not sure anything bad will ever happen here, but since we rely on zero-termination for memory safety, I would prefer if we did the renaming first and then added the zero termination. (Which could maybe use `.push('\0')` then?)

adamreichold (approved):
Thank you for going the extra mile and making this fully generic! I think this is now a powerful helper that can save quite a bit of boiler plate for many people.

pyo3-macros-backend/src/pymethod.rs
DataTriny: I wasn't sure either so I tested and nothing bad happened. I even tried with reserved Rust keywords. Changing nonetheless so it's clearer.

DataTriny:
codecov is failing again... I think I added enough tests and it can be ignored right?

adamreichold:
> codecov is failing again... I think I added enough tests and it can be ignored right?

Yes, I think so too. The only reason I did not merge this yet is to give the other maintainers some time to chime in on this version of the code.

tests/test_class_attributes.rs
```diff
@@ -160,3 +160,42 @@ RuntimeError: An error occurred while initializing class BrokenClass"
...
+        assert!(struct_obj
+            .setattr("firstField", PyBool::new(py, false))
+            .is_ok());
+        py_assert!(py, struct_obj, "struct_obj.firstField == False");
+        py_assert!(py, struct_obj, "struct_obj.secondField == 5");
+        assert!(struct_obj
+            .setattr("third_field", PyBool::new(py, true))
+            .is_ok());
+        py_assert!(py, struct_obj, "struct_obj.third_field == True");
+    });
```
davidhewitt, line 200: It would be quite nice to have some kind of macro here which can test each type of rename_all in turn, so that we test they all work as expected.

```rust
test_case!(CamelCase, "camelCase", "ab_cd-ef", "abCdEf");
test_case!(SnakeCase, "snake_case", "ab_cd-ef", "ab_cd_ef");
```

etc

davidhewitt (approved):
Looks great to me, thanks for the solid implementation.

I just have one suggestion on an addition to the test suite (which will drive the coverage just a teeny bit higher too), otherwise I'm happy to see this merged whenever.

tests/test_class_attributes.rs
DataTriny: I added a macro which I think does the job. At least codecov likes it.

PR #3385, opened 2023-08-14
Remove IntoPyPointer

alex:
it wasn't used in any public APIs following #3359

refs #3358


adamreichold (approved):

davidhewitt (commented):
I am happy to see this go, especially given #3325, however I think we should add a note in the migration guide given we're not giving this a deprecation cycle.

alex:
Good call, done!

davidhewitt (approved):
Thanks!

Issue #2339, opened 2022-04-26
Dual licensing under Apache and MIT


mwcampbell:
How difficult would it be to dual-license PyO3 under the Apache and MIT licenses, like Rust itself? This would work around the incompatibility between the Apache license and GPL 2. I ask about this because I'm thinking about using PyO3 in an add-on to a GPL 2 program.

davidhewitt:
A good question. I am by no means an expert in licensing law. 

https://opensource.stackexchange.com/a/2032 suggests it could be very difficult. If all previous PyO3 contributors need to grant approval then in practice this may be impossible.

I'm sure that this must be done from time to time with open source projects. If you can find examples of how projects have done this in practice, I would be happy to cast my vote for dual licensing if a process was agreed.

alex:
We did this on pyca/cryptography many years ago: https://github.com/pyca/cryptography/issues/1209 is how we did it.

kngwyu:
Yeah I think in theory we can do it, though it might take time since we now have 182 contributors.

davidhewitt:
@alex thanks for sharing that. How did you identify the list of contributors to contact? Was it just everyone who had ever submitted to the repo?

alex:
Yeah, `git log`

mwcampbell:
I'm willing to personally donate some money to help pay someone to take care of this. There's a GPL2-licensed Python-based project I care very much about that, I think, could benefit from gradually adopting Rust in some extension modules. It would be unfortunate if the technical implementation of this had to be compromised due to a license incompatibility.

davidhewitt:
Sorry for running slow on this. I'm not sure a financial incentive is necessary here. It should be straightforward enough to generate a list of contributors and ask for approval.

Presumably that list would include all contributors from `rust-cpython` before this project was forked? A good first step may be to seek opinions from @dgrunwald and @fafhrd91 on what they think of changing PyO3 to be dual licensed.

kngwyu:
I just note that rust-cpython is distributed under MIT license. I hope it won't make the problem more complex.

DataTriny:
I'm also interested in this.

@dgrunwald and @fafhrd91: would you agree to such a change?

If yes, I'm willing to handle the task of contacting everyone and updating necessary files.

dgrunwald:
I am OK with dual-licensing rust-cpython under the Apache and MIT licenses.

fafhrd91:
it is fine with me

[...]

PR #3387, opened 2023-08-15
Remove non-MIT contributions

davidhewitt:
As per https://github.com/PyO3/pyo3/issues/2339#issuecomment-1678555706

This PR reverts commits by @cecini, @potocpav, and @thanatos, who have been the non-respondents to our intent to relicense.

Mostly these lines have been reworked or replaced already, so the majority of these reverts are empty and purely present for ceremonial purpose.

The one possible change which was meaningful was correct handling of `Py_TRACE_REFS` flag which was part of `Py_DEBUG` up to and including 3.7. Therefore to move forward I propose to drop support for *debug* builds on Python 3.7. @alex I assume that dropping support of 3.7 debug interpreters will not impact cryptography, given debug wheels aren't shipped to PyPI.

### DIFF
--- a/guide/src/exception.md
+++ b/guide/src/exception.md
@@ -2,7 +2,7 @@
 
 ## Defining a new exception
 
-You can use the [`create_exception!`] macro to define a new exception type:
+Use the [`create_exception!`] macro:
 
 ```rust
 use pyo3::create_exception;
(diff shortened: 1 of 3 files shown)


DataTriny:
I just double-checked we weren't missing anyone. Every boxes are now ticked except for the people mentioned here!

alex:
Yeah, I'm not sure I've ever done anything with a debug interpreter + pyo3, so no impact for us.

adamreichold (approved):

PR #3389, opened 2023-08-15
Migrate `PyIterator::from_object` and `PyByteArray::from` from `AsPyPointer` to `&PyAny`

alex:


davidhewitt (approved):
Thanks, as discussed on the issue, while this is direct breakage it should be relatively minor for most users and given the need to remove / minimize `AsPyPointer` I think it's prudent to accept this fallout without a deprecation dance.

PR #3391, opened 2023-08-16
Remove usage of `AsPyPointer` in traits for convergint to PyObject

alex:
Refs #3358

### DIFF
--- a/newsfragments/3391.changed.md
+++ b/newsfragments/3391.changed.md
@@ -1,0 +1,1 @@
+Replace `AsPyPointer` with `AsRef<PyAny>` as a bound in the blanket implementation of `From<&T> for PyObject`
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -933,12 +933,18 @@ impl<T> crate::AsPyPointer for Py<T> {
     }
 }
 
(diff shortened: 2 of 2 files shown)


alex:
I'm not sure if a changelog is actually required here.

alex:
Pretty convinced it does not: Unless a user manually implements `PyNativeType`, all implementations will have both `AsPyPointer` and also `AsRef<PyAny>` (or _be_ `PyAny`). Therefore, excluding that (unlikely) situation, this will not have any user visible impact.

davidhewitt (commented):
Personally I'd still prefer to see this documented. I think "Replace `AsPyPointer` with `AsRef<PyAny>` as a bound in the blanket implementation of `From<&T> for PyObject`" is enough.

davidhewitt:
Aside from that, in pretty certain this is the right solution to this one, nicely done 👍

I had wondered if we'd need to remove the blanket implementation and add concrete implementations in the macros in `types/mod.rs`, which would have been more breaking.

alex:
Okey doke. Done.

davidhewitt:
Thanks!

PR #3392, opened 2023-08-16
tidy up some file locations after dual licensing

davidhewitt:
Some tidy ups from #3108, which I thought better to push as my own PR rather than review comments on the extremely busy thread.

- Moves the newsfragment from the `changed` category to `packaging` category.
- Fixes the `LICENSE` symlinks for all sub-crates to be two `LICENSE-MIT` and `LICENSE-APACHE` symlinks.

cc @DataTriny


DataTriny (approved):
Looks good to me! Sorry for missing these.

PR #3393, opened 2023-08-16
Remove usage of `AsPyPointer` in `IntoPy<PyObject>` trait implementation

alex:


alex:
FWIW I believe the set of types here is comprehensive, looking at https://docs.rs/pyo3/latest/pyo3/conversion/trait.AsPyPointer.html#implementors:

- The first gazillion all implement `AsRef<PyAny>`
- `PyRef` and `PyMutRef` implement `IntoPy<Py<PyAny>>` themselves. Do we need an `impl` for `&PyRef` and `&PyMutRef`?
- `Py` is handled here
- `PyCell` implements `AsRef<PyAny>`

davidhewitt (commented):
While I think `&PyRef` and `&PyRefMut` are extremely rare cases, given the amount of churn we might create for users with all this change to `AsPyPointer` and `IntoPyPointer`, it seems fairer to users if we add those impls too. They did exist under the old blanket impl, after all.

alex:
Easy enough.

alex:
Done

src/instance.rs
```diff
@@ -925,6 +925,13 @@ impl<T> IntoPy<PyObject> for Py<T> {
     }
 }
 
+impl<T> IntoPy<PyObject> for &'_ Py<T> {
+    #[inline]
+    fn into_py(self, py: Python<'_>) -> PyObject {
+        unsafe { PyObject::from_borrowed_ptr(py, self.as_ptr()) }
```
adamreichold, line 931: Maybe just forward to `.to_object()` or the other way around?

davidhewitt (approved):
Thanks! One more log on the `AsPyPointer` bonfire. 😂

PR #3397, opened 2023-08-17
Pick a relevant ErrorKind when building an io::Error from a OSError subclass

Tpt:


adamreichold (approved):

davidhewitt (approved):
Thanks, this makes sense to roundtrip.

PR #3402, opened 2023-08-18
Unwrap the underlying PyErr when converting an io::Error back to a PyErr

Tpt:
Exposes it directly instead of loosing all information outside the message

This is useful for use cases like "wrap Python I/O in a `io::Read` Rust trait" where the python exception are converted to `io::Error` then converted again to exceptions


davidhewitt (approved):
👍 from me; we do similar stuff with `anyhow` errors.

adamreichold (approved):

PR #3403, opened 2023-08-18
update definitions for objimpl.h

davidhewitt:
While looking into #3064 I noticed that our definitions for `objimpl.h` were extremely out of date. (I wanted to use the added symbols for debugging purposes.)

Mostly this is just reshuffling, the API changes are in the newsfragments.


davidhewitt (commented):
As this is a mechanical update to these FFI definitions with nothing I'd expect to be contentious, I'm going to save on reviewer bandwidth and just merge this.

PR #3404, opened 2023-08-18
call PyObject_GC_Untrack before deallocating

### LINKED ISSUE #3064: ResourceWarning: Object of type XXX is not untracked before destruction

### LINKED ISSUE #3416: Run CI with Python 3.11 debug build
It struck me to test #3404 properly we'd need to run against a python debug build in CI. I think it'd be sufficient to run for 3.11 on Ubuntu, hopefully that would catch most things.

I don't know if there's a prebuilt debug build we can install; we might have to build it ourselves as part of CI.

davidhewitt:
Fixes #3064 
Closes #3416 

~~I verified by hand that this fixes the warning on a debug build of Python. As we don't run debug builds in CI I'm unsure if there's any way we can test this.~~ EDIT: now testing debug build in CI as part of this PR.

As a bit of refactoring here I removed `Layout` from `PyClassImpl` as it's no longer necessary.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -376,8 +376,11 @@ jobs:
           components: rust-src
       - name: Install python3 standalone debug build with nox
         run: |
-          wget https://github.com/indygreg/python-build-standalone/releases/download/20230826/cpython-3.11.5+20230826-x86_64-unknown-linux-gnu-debug-full.tar.zst
-          tar -I zstd -xf cpython-3.11.5+20230826-x86_64-unknown-linux-gnu-debug-full.tar.zst
+          PBS_RELEASE="20230826"
+          PBS_PYTHON_VERSION="3.11.5"
+          PBS_ARCHIVE="cpython-${PBS_PYTHON_VERSION}+${PBS_RELEASE}-x86_64-unknown-linux-gnu-debug-full.tar.zst"
+          wget "https://github.com/indygreg/python-build-standalone/releases/download/${PBS_RELEASE}/${PBS_ARCHIVE}"
+          tar -I zstd -xf "${PBS_ARCHIVE}"
           ls -l $(pwd)/python/install/bin
           ls -l $(pwd)/python/install/lib
           echo PATH=$(pwd)/python/install/bin:$PATH >> $GITHUB_ENV
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -363,6 +363,46 @@ jobs:
       - name: Test
         run: nox -s test-emscripten
 
+  test-debug:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: cargo-test-debug
+        continue-on-error: true
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rust-src
+      - name: Install python3 standalone debug build with nox
+        run: |
+          wget https://github.com/indygreg/python-build-standalone/releases/download/20230826/cpython-3.11.5+20230826-x86_64-unknown-linux-gnu-debug-full.tar.zst
+          tar -I zstd -xf cpython-3.11.5+20230826-x86_64-unknown-linux-gnu-debug-full.tar.zst
+          ls -l $(pwd)/python/install/bin
+          ls -l $(pwd)/python/install/lib
+          echo PATH=$(pwd)/python/install/bin:$PATH >> $GITHUB_ENV
+          echo LD_LIBRARY_PATH=$(pwd)/python/install/lib:$LD_LIBRARY_PATH >> $GITHUB_ENV
+          echo PYTHONHOME=$(pwd)/python/install >> $GITHUB_ENV
+          echo PYO3_PYTHON=$(pwd)/python/install/bin/python3 >> $GITHUB_ENV
+      - run: python3 -m sysconfig
+      - run: python3 -m pip install --upgrade pip && pip install nox
+      - run: |
+          PYO3_CONFIG_FILE=$(mktemp)
+          cat > $PYO3_CONFIG_FILE << EOF
+          implementation=CPython
+          version=3.11
+          shared=true
+          abi3=false
+          lib_name=python3.11d
+          lib_dir=${{ github.workspace }}/python/install/lib
+          executable=${{ github.workspace }}/python/install/bin/python3
+          pointer_width=64
+          build_flags=Py_DEBUG,Py_REF_DEBUG
+          suppress_build_script_link_lines=false
+          EOF
+          echo PYO3_CONFIG_FILE=$PYO3_CONFIG_FILE >> $GITHUB_ENV
+      - run: python3 -m nox -s test
+
   conclusion:
     needs:
       - fmt
--- a/noxfile.py
+++ b/noxfile.py
@@ -475,6 +475,7 @@ def set_minimal_package_versions(session: nox.Session):
     min_pkg_versions = {
         "rust_decimal": "1.26.1",
         "csv": "1.1.6",
+        "indexmap": "1.9.3",
         "hashbrown": "0.12.3",
         "log": "0.4.17",
         "once_cell": "1.17.2",
@@ -482,7 +483,6 @@ def set_minimal_package_versions(session: nox.Session):
         "rayon-core": "1.10.2",
         "regex": "1.7.3",
         "proptest": "1.0.0",
-        "indexmap": "1.9.3",
         "chrono": "0.4.25",
     }
 
--- a/src/impl_/freelist.rs
+++ b/src/impl_/freelist.rs
@@ -55,7 +55,7 @@ impl<T> FreeList<T> {
         }
     }
 
-    /// Inserts a value into the list. Returns `None` if the `FreeList` is full.
+    /// Inserts a value into the list. Returns `Some(val)` if the `FreeList` is full.
     pub fn insert(&mut self, val: T) -> Option<T> {
         let next = self.split + 1;
         if next < self.capacity {
--- a/src/pyclass/create_type_object.rs
+++ b/src/pyclass/create_type_object.rs
@@ -1,16 +1,18 @@
+use pyo3_ffi::PyType_IS_GC;
+
 use crate::{
     exceptions::PyTypeError,
     ffi,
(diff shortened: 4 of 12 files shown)


noxfile.py
```diff
@@ -475,14 +475,14 @@ def set_minimal_package_versions(session: nox.Session):
     min_pkg_versions = {
         "rust_decimal": "1.26.1",
         "csv": "1.1.6",
+        "indexmap": "1.9.3",
```
davidhewitt, line 478: `hashbrown` depends on `indexmap`, I was having trouble locally setting minimal package versions on 1.56 without putting `indexmap` above it.

src/impl_/freelist.rs
```diff
@@ -55,7 +55,7 @@ impl<T> FreeList<T> {
         }
     }
 
-    /// Inserts a value into the list. Returns `None` if the `FreeList` is full.
+    /// Inserts a value into the list. Returns `Some(val)` if the `FreeList` is full.
```
davidhewitt, line 58: Drive-by typo fix 🙈

src/pyclass/create_type_object.rs
```diff
@@ -32,22 +32,37 @@ where
...
-            .class_items(T::items_iter())
-            .build(py, T::NAME, T::MODULE, std::mem::size_of::<T::Layout>())
+        PyTypeBuilder {
+            slots: Vec::new(),
+            method_defs: Vec::new(),
+            getset_builders: HashMap::new(),
+            cleanup: Vec::new(),
+            tp_base: T::BaseType::type_object_raw(py),
+            tp_dealloc: tp_dealloc::<T>,
+            tp_dealloc_with_gc: tp_dealloc_with_gc::<T>,
```
davidhewitt, line 42: The implementation here compiles both `tp_dealloc` and `tp_dealloc_with_gc` for all `T`, and then selects which to actually use based on whether `has_traverse` gets set later. 

Given that there's a one-line difference between the two functions I didn't think it was worth trying to reduce the minimal code bloat by doing something fancy (probably with autoref specialization) to work out which one to use at compile time.

messense (approved):

src/pyclass/create_type_object.rs
```diff
@@ -32,22 +32,37 @@ where
...
+            is_mapping: T::IS_MAPPING,
+            is_sequence: T::IS_SEQUENCE,
+            has_new: false,
+            has_dealloc: false,
+            has_getitem: false,
+            has_setitem: false,
+            has_traverse: false,
```
adamreichold, line 49: Maybe we could simplify this using splat syntax `..Default::default()` or is spelling this out intentional?

adamreichold: (Ah sorry, you remove the `Default` impl...)

src/pyclass/create_type_object.rs
adamreichold: I also think having both fields in `PyTypeBuilder` is not an issue. I do wonder about inlining though, i.e. whether forwarding from `tp_dealloc_with_gc` to `tp_dealloc` is good.

I suspect both functions are relatively unlikely being inlined as they are called indirectly. So maybe should call `crate::impl_::trampoline::dealloc(obj, PyCell::<T>::tp_dealloc)` directyl from within `tp_dealloc_with_gc` as well?

davidhewitt:
With the `test-debug` CI run in #3430 we now can reproduce the resourcewarning lines from #3064, e.g. in https://github.com/PyO3/pyo3/actions/runs/6093238722/job/16532577116#step:9:1196

I've rebased on #3430 here and it's great to see that those lines are gone and even better, `test-debug` is now fixed.

src/pyclass/create_type_object.rs
davidhewitt: Sorry for the long delay, just force-pushed to adjust `tp_alloc_with_gc` as suggested.

.github/workflows/ci.yml
```diff
@@ -363,6 +363,46 @@ jobs:
...
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: cargo-test-debug
+        continue-on-error: true
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rust-src
+      - name: Install python3 standalone debug build with nox
+        run: |
+          wget https://github.com/indygreg/python-build-standalone/releases/download/20230826/cpython-3.11.5+20230826-x86_64-unknown-linux-gnu-debug-full.tar.zst
```
adamreichold, line 379: Could we factor out the date/version to make it easier to change this script?

adamreichold (approved):

.github/workflows/ci.yml
davidhewitt: Good idea, pushed something which I think will do the trick.

davidhewitt:
Looks like everything except 3.11 and pypy is broken 😅

davidhewitt:
Ok, think I've fixed the issues. The 3.12 failures were #3444, and the 3.10 (and older) cpython failures were due to an annoying assumption in `BaseException_dealloc` that the object was still tracked, see the last commit pushed. Will proceed to merge if this really does go green...

PR #3406, opened 2023-08-19
remove AsPyPointer from `FromPyObject` impl for Py<T>`

davidhewitt:
Related to #3358

This drops the `AsPyPointer` bound from the `Py<T>` conversion. I think the resulting implementation is more intuitive and almost certainly more performant.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -1017,14 +1017,12 @@ impl<T> Drop for Py<T> {
 impl<'a, T> FromPyObject<'a> for Py<T>
 where
     T: PyTypeInfo,
-    &'a T::AsRefTarget: FromPyObject<'a>,
-    T::AsRefTarget: 'a + AsPyPointer,
 {
-    /// Extracts `Self` from the source `PyObject`.
     fn extract(ob: &'a PyAny) -> PyResult<Self> {
-        unsafe {
-            ob.extract::<&T::AsRefTarget>()
-                .map(|val| Py::from_borrowed_ptr(ob.py(), val.as_ptr()))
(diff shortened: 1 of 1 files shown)


alex (approved):

davidhewitt:
I can't decide whether this one is CHANGELOG worthy. By removing bounds here I can't imagine it would affect user code?

src/instance.rs
```diff
@@ -1017,14 +1017,12 @@ impl<T> Drop for Py<T> {
...
-    &'a T::AsRefTarget: FromPyObject<'a>,
-    T::AsRefTarget: 'a + AsPyPointer,
 {
-    /// Extracts `Self` from the source `PyObject`.
     fn extract(ob: &'a PyAny) -> PyResult<Self> {
-        unsafe {
-            ob.extract::<&T::AsRefTarget>()
-                .map(|val| Py::from_borrowed_ptr(ob.py(), val.as_ptr()))
+        if ob.is_instance_of::<T>() {
+            Ok(unsafe { Py::from_non_null(NonNull::new_unchecked(ob.into_ptr())) })
```
adamreichold, line 1023: So this can assume that `val` will always point to the same object (i.e. not a sub-allocation) as `ob`?

davidhewitt: Ah, that's a good observation. I think I've always assumed that should be the case, and I'd quite like this to be true. But it is breaking. So if we like this property we could document this as a breaking change?

davidhewitt:
Closing as I think we'll release 0.20 without this change and then hopefully a rework in 0.21 will remove the need for this.

PR #3408, opened 2023-08-19
Add types for `None`, `Ellipsis`, and `NotImplemented`

### LINKED ISSUE #3162: Add types for `None`, `Ellipsis` and `NotImplemented`
Rather than returning `PyObject` here it would probably be nicer if we returned `&PyEllipsis`, and created a new struct `types::PyEllipsis`. (Would avoid the need to increase a reference count here.)

That would also mean that `x.is_instance::<PyEllipsis>()` would just work instead of needing `is_ellipsis`, which does have the upside of avoiding a special method. Not necessarily saying that it's better, perhaps `is_ellipsis` is still warranted.

Also, adding `types::PyEllipsis` would pave the way for instead just having `PyEllipsis::get(py)` to retrieve the singleton, rather than adding this to `Python` which I don't think really needs this method.

_Originally posted by @davidhewitt in https://github.com/PyO3/pyo3/pull/2911#discussion_r1087252466_

rytheo:
Fixes #3162

### DIFF
--- a/newsfragments/3408.added.md
+++ b/newsfragments/3408.added.md
@@ -1,0 +1,1 @@
+Add types for `None`, `Ellipsis`, and `NotImplemented`
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -262,19 +262,6 @@ where
     }
 }
 
-/// `()` is converted to Python `None`.
-impl ToPyObject for () {
-    fn to_object(&self, py: Python<'_>) -> PyObject {
-        py.None()
-    }
-}
-
-impl IntoPy<PyObject> for () {
-    fn into_py(self, py: Python<'_>) -> PyObject {
-        py.None()
-    }
-}
-
 impl IntoPy<PyObject> for &'_ PyAny {
     #[inline]
     fn into_py(self, py: Python<'_>) -> PyObject {
--- a/src/marker.rs
+++ b/src/marker.rs
@@ -118,7 +118,9 @@
 use crate::err::{self, PyDowncastError, PyErr, PyResult};
 use crate::gil::{GILGuard, GILPool, SuspendGIL};
 use crate::impl_::not_send::NotSend;
-use crate::types::{PyAny, PyDict, PyModule, PyString, PyType};
+use crate::types::{
+    PyAny, PyDict, PyEllipsis, PyModule, PyNone, PyNotImplemented, PyString, PyType,
+};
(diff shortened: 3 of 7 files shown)


davidhewitt (commented):
Thank you! Looks like a merge conflict and also needs a newsfragment `3408.added.md` please

rytheo:
The singletons don't expose their type objects in the C API. Is it necessary for these types to implement `PyTypeInfo`?

davidhewitt:
Ah yes, looks like they are not part of the stable API and will even been removed in Python 3.13 https://github.com/python/cpython/pull/107030

So the way to implement for `None` and `NotImplemented` should be like we implement `PySequence` and `PyMapping`. (You're right, we can't implement `PyTypeInfo` for them.)

davidhewitt (commented):
Thanks, this implementation looks correct to me. Would you be willing to add some tests please?

I think it should be sufficient to check that some arbitrary value like empty dict doesn't downcast to these types (e.g. `PyDict::new(py).downcast::<PyEllipsis>().is_err()`) and also that downcasting the singletons succeeds (e.g. `py.Ellipsis().as_ref(py).downcast().unwrap().is(&py.Ellipsis())`).

mejrs:
Should we deprecate/remove `Py::None` in favour of this?

davidhewitt:
That's a good question; I'd prefer to leave the `py.None()` etc APIs unchanged for 0.20 because they'll probably all change anyway in 0.21. Probably we do want to remove `py.None()` as it's untyped.

So for now, this PR looks good to me, thanks @rytheo!

davidhewitt:
Ah that's unfortunate, it looks like PyPy doesn't expose `PyEllipsis_Type`. Probably the most practical solution is to change the `PyEllipsis` implementation here to expose `PyTryFrom` instead of `PyTypeInfo` as well, then the three implementations work the same. (An alternative would be to use `#[cfg]` to have different CPython and PyPy implementations, but I don't think that's as nice for either maintainers or users.)

rytheo:
@davidhewitt the `coverage-ubuntu` CI failure was caused by some `unterminated character literal` error. It doesn't seem to be related to my changes. Any ideas?

davidhewitt:
Looks like a possible bug in `cargo llvm-cov`, let's see if rerunning fixes

davidhewitt:
Please rebase and CI should be working again.

PR #3412, opened 2023-08-25
don't reinitialize the interpreter when it is shutting down

davidhewitt:
Related to #3386.

If `auto-intialize` is enabled in an extension-module, we can hit a nasty C crash on interpreter shutdown where `auto-initialize` attempts to re-initialize it. I added an internal `mark_python_initialized_from_import` to disable `auto-initalize` when a `#[pymodule]` is imported (as the interpreter must be initialized at that point).

Second I add an assertion that the interpreter is initialized before attempting to acquire the GIL, to avoid odd crashes after the interpreter has been shut down (and just panic).

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -94,6 +94,11 @@ pub fn prepare_freethreaded_python() {
     });
 }
 
+/// When a PyO3 module is imported we know that the Python interpreter has been initialized
+pub(crate) fn mark_python_initialized_from_import() {
+    START.call_once(|| {})
+}
+
 /// Executes the provided closure with an embedded Python interpreter.
 ///
 /// This function initializes the Python interpreter, executes the provided closure, and then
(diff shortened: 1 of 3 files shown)


src/gil.rs
```diff
@@ -208,6 +213,14 @@ impl GILGuard {
             }
         }
 
+        // This is necessary to protect against Python::with_gil being called after the Python
+        // interpreter has started shutting down.
+        assert_eq!(
```
adamreichold, line 218: Could we somehow move is behind the `if gil_is_acquired()` check to avoid paying this cost in the no-op case of nested calls to `with_gil`?

adamreichold: Ah sorry, it already is behind the check because the check itself is duplicated. Sorry for noise.

adamreichold:
> I added an internal mark_python_initialized_from_import to disable auto-initalize when a #[pymodule] is imported (as the interpreter must be initialized at that point).

Since our own `START` can only protect us from multiple threads calling `prepare_freethreaded_python`, I thought it would be best effort in any case? Meaning that shouldn't `prepare_freethreaded_python` itself somehow be able to detect if we in shutdown on its own? Put differently, I would not want us to rely on the state of `START` beyond its original intention of curbing multiple threads racing initialization.

davidhewitt:
Agreed, I will scrap this. `START` will be consumed anyway if `with_gil` is called at any other point during program execution. The example from the discussion was an edge case that we don't need to cater for.

PR #3413, opened 2023-08-25
fix escaping of llvm-cov output

davidhewitt:
Looks like the output of `cargo llvm-cov show-env` changed from escaped by `"` to `'`. I think that's the cause of our CI issues.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -606,7 +606,11 @@ def _get_coverage_env() -> Dict[str, str]:
 
     for line in output.strip().splitlines():
         (key, value) = line.split("=", maxsplit=1)
-        env[key] = value.strip('"')
+        # Strip single or double quotes from the variable value
+        # - quote used by llvm-cov differs between Windows and Linux
+        if value and value[0] in ("'", '"'):
+            value = value[1:-1]
+        env[key] = value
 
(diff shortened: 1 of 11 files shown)


davidhewitt:
I'll also add updates for Rust 1.72 to this PR...

noxfile.py
```diff
@@ -606,7 +606,13 @@ def _get_coverage_env() -> Dict[str, str]:
 
     for line in output.strip().splitlines():
         (key, value) = line.split("=", maxsplit=1)
-        env[key] = value.strip('"')
+        # Strip single or double quotes from the variable value
+        # - quote used by llvm-cov differs between Windows and Linux
+        if value and value[0] == "'":
```
adamreichold, line 611: This looks overly complex because the two branches are the same?

adamreichold: Meaning

```python
if value and value[0] in ["'", '"']:
  value = value[1:-1]
```

should also work?

adamreichold (approved):
LGTM modulo `clippy::if_same_then_else` on the Python code.

noxfile.py
davidhewitt: Good point, fixed in force-push.

PR #3414, opened 2023-08-25
simplify thread checker implementation

davidhewitt:
I noticed that we don't need `ThreadCheckerInherited` because `PyCellLayout`'s `ensure_threadsafe` method already checks the whole inheritance tree.

At the same time I adjusted the implementation a little bit to reduce generic code and tidy up the error messages.

### DIFF
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -947,13 +947,9 @@ impl<'a> PyClassImplsBuilder<'a> {
         };
 
         let thread_checker = if self.attr.options.unsendable.is_some() {
-            quote! { _pyo3::impl_::pyclass::ThreadCheckerImpl<#cls> }
-        } else if self.attr.options.extends.is_some() {
-            quote! {
-                _pyo3::impl_::pyclass::ThreadCheckerInherited<#cls, <#cls as _pyo3::impl_::pyclass::PyClassImpl>::BaseType>
-            }
+            quote! { _pyo3::impl_::pyclass::ThreadCheckerImpl }
         } else {
-            quote! { _pyo3::impl_::pyclass::ThreadCheckerStub<#cls> }
+            quote! { _pyo3::impl_::pyclass::SendablePyClass<#cls> }
         };
 
         let (pymethods_items, inventory, inventory_class) = match self.methods_type {
--- a/src/impl_/pyclass.rs
+++ b/src/impl_/pyclass.rs
@@ -1018,53 +1018,46 @@ pub trait PyClassThreadChecker<T>: Sized {
     private_decl! {}
 }
 
-/// Stub checker for `Send` types.
+/// Default thread checker for `#[pyclass]`.
+///
+/// Keeping the T: Send bound here slightly improves the compile
+/// error message to hint to users to figure out what's wrong
+/// when `#[pyclass]` types do not implement `Send`.
 #[doc(hidden)]
-pub struct ThreadCheckerStub<T: Send>(PhantomData<T>);
+pub struct SendablePyClass<T: Send>(PhantomData<T>);
 
-impl<T: Send> PyClassThreadChecker<T> for ThreadCheckerStub<T> {
+impl<T: Send> PyClassThreadChecker<T> for SendablePyClass<T> {
     fn ensure(&self) {}
     fn can_drop(&self, _py: Python<'_>) -> bool {
         true
     }
     #[inline]
     fn new() -> Self {
-        ThreadCheckerStub(PhantomData)
+        SendablePyClass(PhantomData)
     }
     private_impl! {}
 }
 
-impl<T: PyNativeType> PyClassThreadChecker<T> for ThreadCheckerStub<crate::PyObject> {
-    fn ensure(&self) {}
-    fn can_drop(&self, _py: Python<'_>) -> bool {
-        true
-    }
-    #[inline]
-    fn new() -> Self {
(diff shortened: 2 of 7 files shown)


pyo3-macros-backend/src/pyclass.rs
```diff
@@ -947,13 +947,9 @@ impl<'a> PyClassImplsBuilder<'a> {
         };
 
         let thread_checker = if self.attr.options.unsendable.is_some() {
-            quote! { _pyo3::impl_::pyclass::ThreadCheckerImpl<#cls> }
-        } else if self.attr.options.extends.is_some() {
-            quote! {
-                _pyo3::impl_::pyclass::ThreadCheckerInherited<#cls, <#cls as _pyo3::impl_::pyclass::PyClassImpl>::BaseType>
-            }
+            quote! { _pyo3::impl_::pyclass::ThreadCheckerImpl }
```
adamreichold, line 950: Wouldn't a name change like `UnsendablePyClass` be more consistent then?

adamreichold: Yes, I meant purely to help someone reading the implementation.

src/impl_/pyclass.rs
```diff
@@ -1013,85 +1013,64 @@ pub trait PyClassThreadChecker<T>: Sized {
     private_decl! {}
 }
 
-/// Stub checker for `Send` types.
+/// Default thread checker for `#[pyclass]`.
+///
+/// Keeping the T: Send bound here slightly improves the compile
```
adamreichold, line 1018: But does that also imply keeping the generic parameter on the struct itself?

src/impl_/pyclass.rs
```diff
@@ -1013,85 +1013,64 @@ pub trait PyClassThreadChecker<T>: Sized {
...
+            "{} is unsendable, but sent to another thread",
+            type_name
         );
     }
-    fn can_drop(&self, py: Python<'_>) -> bool {
+
+    fn can_drop(&self, py: Python<'_>, type_name: &'static str) -> bool {
         if thread::current().id() != self.0 {
             PyRuntimeError::new_err(format!(
-                "{} is unsendbale, but is dropped on another thread!",
```
adamreichold, line 1061: I am pretty sure I am meant to reference unsent Cotton bales here... 🤦🏽‍♀️

adamreichold (approved):
Maybe one less type parameter and one naming inconsistency, otherwise LGTM.

src/impl_/pyclass.rs
davidhewitt: Yes, here's the error message with `SendablePyClass<T: Send>`:

```
error[E0277]: `Rc<i32>` cannot be sent between threads safely
 --> tests/ui/pyclass_send.rs:4:1
  |
4 | #[pyclass]
  | ^^^^^^^^^^ `Rc<i32>` cannot be sent between threads safely
  |
  = help: within `NotThreadSafe`, the trait `Send` is not implemented for `Rc<i32>`
  = note: use `std::sync::Arc` instead of `std::rc::Rc`
note: required because it appears within the type `NotThreadSafe`
 --> tests/ui/pyclass_send.rs:5:8
  |
5 | struct NotThreadSafe {
  |        ^^^^^^^^^^^^^
note: required by a bound in `SendablePyClass`
 --> src/impl_/pyclass.rs
  |
  | pub struct SendablePyClass<T: Send>(PhantomData<T>);
  |                               ^^^^ required by this bound in `SendablePyClass`
  = note: this error originates in the attribute macro `pyclass` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Here's the error message once the generic parameter is removed:

```
error[E0277]: `Rc<i32>` cannot be sent between threads safely
 --> tests/ui/pyclass_send.rs:4:1
  |
4 | #[pyclass]
  | ^^^^^^^^^^ `Rc<i32>` cannot be sent between threads safely
  |
  = help: within `NotThreadSafe`, the trait `Send` is not implemented for `Rc<i32>`
  = help: the trait `pyo3::impl_::pyclass::PyClassThreadChecker<T>` is implemented for `SendablePyClass`
note: required because it appears within the type `NotThreadSafe`
 --> tests/ui/pyclass_send.rs:5:8
  |
5 | struct NotThreadSafe {
  |        ^^^^^^^^^^^^^
  = note: required for `SendablePyClass` to implement `pyo3::impl_::pyclass::PyClassThreadChecker<NotThreadSafe>`
note: required by a bound in `PyClassImpl::ThreadChecker`
 --> src/impl_/pyclass.rs
  |
  |     type ThreadChecker: PyClassThreadChecker<Self>;
  |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `PyClassImpl::ThreadChecker`
  = note: this error originates in the attribute macro `pyclass` (in Nightly builds, run with -Z macro-backtrace for more info)
```

There's not a huge amount in it. The one with the generic parameter gets `Send` requirement directly underlined, after removing the generic parameter the error spills onto the trait implementation and doesn't give such a clear hint that `Send` should be implemented IMO.

So I'm going to keep the generic parameter for now and we can decide whether to remove it another time.

adamreichold: Fair enough.

davidhewitt:
The CI failure is because `abi3_nativetype_inheritance.stderr` no longer spits out any error message related to thread checkers 🎉

davidhewitt:
... and now it's caused by a `chrono` update 🙃

PR #3418, opened 2023-08-28
Mention inability to subclass Python classes in docs

### LINKED ISSUE #3415: Docs should mention that it's not possible to derive Rust classes from Python ones
From the answers to #296 I gather that it's not currently possible to have Rust-implemented pyclasses be subclasses of regular Python classes. As far as I can tell, the docs don't mention this explicitly and the [relevant section](https://pyo3.rs/v0.19.2/class.html) only talks about classes that are available as Rust types (like user-defined pyclasses and the ones that come with PyO3).

In my opinion, it wouldn't hurt to mention this explicitly in the docs, both in the text and in the reference table entry for `extends`. Someone else also requested this in https://github.com/PyO3/pyo3/issues/296#issuecomment-1078791626.

I thought that maybe the reason it isn't mentioned is that this is a limitation of Python extension types in general rather than one of PyO3, but [Python's docs on the issue](https://docs.python.org/3/extending/newtypes_tutorial.html#subclassing-other-types) say only that deriving extension types from built-in rather than custom Python types is the easiest option, not that it's the only one.

smheidrich:
Fixes #3415

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -267,6 +267,10 @@ Consult the table below to determine which type your constructor should return:
 
 By default, `object`, i.e. `PyAny` is used as the base class. To override this default,
 use the `extends` parameter for `pyclass` with the full path to the base class.
+Currently, only classes defined in Rust and builtins provided by PyO3 can be inherited
+from; inheriting from other classes defined in Python is not yet supported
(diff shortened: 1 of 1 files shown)


davidhewitt (approved):
Thanks!

PR #3421, opened 2023-08-29
try adding codspeed benchmarks

### LINKED ISSUE #3407: Continuous benchmarking, try 2
We previously tried running benchmarks in github actions and had a lot of noise as well as not great experience with charts not maintenance of the upstream action.

Recently I have been using https://codspeed.io/ with `pydantic-core` and been quite impressed by how it works. Also I have been in touch with the founder as part of our usage in pydantic so should be easy for us to give feedback if we have any.

If no objections, I'd be up for configuring codspeed in our CI and see what we think. It may be helpful given the work we're doing in performance recently and also potentially for 0.21

davidhewitt:
Closes #3407 

I'm curious to see if this works / how it looks like. If it's annoying or noisy we can turn it off.

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -1,0 +1,40 @@
+name: benches
+
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+  # `workflow_dispatch` allows CodSpeed to trigger backtest
+  # performance analysis in order to generate initial data.
+  workflow_dispatch:
+
+jobs:
+  benchmarks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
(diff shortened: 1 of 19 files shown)


davidhewitt:
Looks like the basics are working, I need to add codspeed criterion compatibility as per https://docs.codspeed.io/benchmarks/rust#from-criterionrs

davidhewitt:
https://github.com/CodSpeedHQ/codspeed-rust/issues/13

Also, it seems _extremely_ slow to build the benchmarks 🤔

davidhewitt:
Well, it built and ran! Took 63 minutes but I think that a lot of that is due to release builds, nothing being cached, and GHA runners being pretty slow. I'm going to merge this as an experiment and then we can revisit later whether this was a good idea, if it loads or blocks CI a lot we can remove.

PR #3427, opened 2023-09-01
fix chrono deprecation warnings

davidhewitt:
`DateTime::from_utc` and `DateTime::from_local` are deprecated in `chrono` 0.4.27. MSRV was also bumped...

### DIFF
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -250,7 +250,7 @@ impl FromPyObject<'_> for DateTime<FixedOffset> {
         let h = dt.get_hour().into();
         let m = dt.get_minute().into();
         let s = dt.get_second().into();
-        let tz = if let Some(tzinfo) = dt.get_tzinfo() {
+        let tz: FixedOffset = if let Some(tzinfo) = dt.get_tzinfo() {
             tzinfo.extract()?
         } else {
             return Err(PyTypeError::new_err("Not datetime.tzinfo"));
@@ -262,7 +262,7 @@ impl FromPyObject<'_> for DateTime<FixedOffset> {
                 .ok_or_else(|| PyValueError::new_err("invalid or out-of-range time"))?,
         );
         // `FixedOffset` cannot have ambiguities so we don't have to worry about DST folds and such
-        Ok(DateTime::from_local(dt, tz))
+        Ok(dt.and_local_timezone(tz).unwrap())
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/conversions/chrono.rs
```diff
@@ -262,7 +262,7 @@ impl FromPyObject<'_> for DateTime<FixedOffset> {
                 .ok_or_else(|| PyValueError::new_err("invalid or out-of-range time"))?,
         );
         // `FixedOffset` cannot have ambiguities so we don't have to worry about DST folds and such
-        Ok(DateTime::from_local(dt, tz))
+        Ok(dt.and_local_timezone(tz).unwrap())
```
adamreichold, line 265: I think we would need to bump our dependency on `chrono` to 0.4.20 if we use `and_local_timezone`.

src/conversions/chrono.rs
```diff
@@ -568,7 +568,7 @@ mod tests {
                         .unwrap()
                         .and_hms_micro_opt(hour, minute, second, ms)
                         .unwrap();
-                    let datetime = DateTime::<Utc>::from_utc(datetime, Utc).to_object(py);
+                    let datetime = Utc.from_utc_datetime(&datetime).to_object(py);
```
adamreichold, line 571: Is this equivalent to `datetime.and_utc()`? Why use one here and another in the implementation above?

newsfragments/3427.fixed.md
```diff
@@ -0,0 +1 @@
+Fix deprecation warnings when compiling with `chrono` 0.4.27 or newer.
```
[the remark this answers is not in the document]
davidhewitt: I thought compile warnings for dependencies does show on a clean build but maybe I'm misremembering.

src/conversions/chrono.rs
davidhewitt: Yeah I'm just really tired recently, will make it consistent.

src/conversions/chrono.rs
davidhewitt: Good catch, I'll adjust the newsfragment accordingly.

I wonder if we should put this minimum version into CI (maybe for the MSRV job). E.g. https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

newsfragments/3427.fixed.md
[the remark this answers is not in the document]
adamreichold: I think this applies only to path dependencies and a quick test indicates that indeed, creating a new project, doing `cargo add pyo3 --features chrono` and building it does not show any warnings. (`cargo check -vv` does, but that level even shows build script outputs.)

src/conversions/chrono.rs
adamreichold: I think minimum versions do not require any special handling and in any case our lockfile does not affect downstream users, so I think a simple `chrono = { version = "0.4.20", optional = true }` is correct and sufficient.

davidhewitt: So we actually need 0.4.25 for `and_utc`, I'll increase to that.

adamreichold (approved):

PR #3430, opened 2023-09-04
test debug build in CI

### LINKED ISSUE #3416: Run CI with Python 3.11 debug build

davidhewitt:
Attempts to resolve #3416 

Uses https://github.com/indygreg/python-build-standalone - at the moment I've hard coded the download link, there could be some smarter logic to update this automatically (and also verify it).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -363,6 +363,46 @@ jobs:
       - name: Test
         run: nox -s test-emscripten
 
+  test-debug:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


davidhewitt:
Closing as I've pulled this commit into #3404, where the full `test-debug` CI run passes after the fix in that PR.

PR #3433, opened 2023-09-05
Add `noplayground` annotation to guide snippets

Hofer-Julian:
None of them can be run in the playground

### DIFF
--- a/guide/book.toml
+++ b/guide/book.toml
@@ -9,3 +9,4 @@ command = "python3 guide/pyo3_version.py"
 [output.html]
 git-repository-url = "https://github.com/PyO3/pyo3/tree/main/guide"
 edit-url-template = "https://github.com/PyO3/pyo3/edit/main/guide/{path}"
+playground.runnable = false


davidhewitt (commented):
Thanks for this!

Two quick thoughts:
- From https://rust-lang.github.io/mdBook/format/mdbook.html#rust-playground it looks like we may be able to do this globally in the `book.toml` configuration, I wonder if that's preferable?
- No need for a CHANGELOG entry for a guide-only change; I've added the `CI-skip-changelog` label, please remove the newsfragment.

Hofer-Julian:
> From https://rust-lang.github.io/mdBook/format/mdbook.html#rust-playground it looks like we may be able to do this globally in the book.toml configuration, I wonder if that's preferable?

Yes, I think so. Added it and it works as expected.

> No need for a CHANGELOG entry for a guide-only change; I've added the CI-skip-changelog label, please remove the newsfragment.

Done

davidhewitt (approved):
Thanks 👍

PR #3434, opened 2023-09-06
add --no-fail-fast to cargo test in CI

davidhewitt:
This might be helpful when changes break multiple tests, or e.g. in #3430 when adding a new test job I'd like to see everything that is broken.

Upside is that all test failures are reported, so fewer pushes & CI runs may be needed overall to fix them.

Downside is that the failures are higher up the log, so scrolling needed to see the test break.

On balance I think this is worth it.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -643,7 +643,7 @@ def _run_cargo_test(
     command = ["cargo"]
     if "careful" in session.posargs:
         command.append("careful")
-    command.append("test")
+    command.extend(("test", "--no-fail-fast"))
     if "release" in session.posargs:
         command.append("--release")
     if package:


noxfile.py
```diff
@@ -643,7 +643,7 @@ def _run_cargo_test(
     command = ["cargo"]
     if "careful" in session.posargs:
         command.append("careful")
-    command.append("test")
+    command += ("test", "--no-fail-fast")
```
adamreichold, line 646: I know this is a super nit, but I think `.extend` would be more consistent with the usage of `.append` everywhere else in this function.

adamreichold (approved):
The rationale for the change seems convincing.

PR #3443, opened 2023-09-09
propagate exception context from pyfunctions

### LINKED ISSUE #3432: "TypeError: No constructor defined" discards exception context
### Bug Description

PyO3 raises `TypeError` when you try to construct a `#[pyclass]` without a constructor. The raised exception doesn't have a value for the `__context__` attribute, which breaks [exception chaining](https://docs.python.org/3/library/exceptions.html#exception-context). This is especially problematic when implementing a custom exception from Rust.

https://github.com/PyO3/pyo3/blob/bf093b3903214784433e3d58771f2979bcebcb8b/src/pyclass/create_type_object.rs#L438-L449

### Steps to Reproduce

PyO3 module:
```rs
use pyo3::exceptions::PyException;
use pyo3::prelude::*;

#[pyclass(extends=PyException)]
pub struct RustException {}

#[pymodule]
fn pyo3_init_bug(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_class::<RustException>()?;
    Ok(())
}
```

Test cases:
```py
from pyo3_init_bug import RustException
import pytest
from traceback import format_exception


class PythonException(Exception):
    def __new__(cls):
        raise TypeError("No constructor defined")


@pytest.mark.parametrize("use_rust_exception", [True, False])
def test_broken_exception(use_rust_exception: bool):
    with pytest.raises(Exception) as exc_info:
        try:
            raise ValueError("Original message")
        except Exception:
            if use_rust_exception:
                raise RustException
            else:
                raise PythonException

    print("".join(format_exception(exc_info.value)))
    assert exc_info.value.__context__ is not None
```

Only the `RustException` test case fails.

### Backtrace

```shell
(venv) PS C:\Users\redacted\pyo3_init_bug> pytest -rP                
=================================================================== test session starts ====================================================================
platform win32 -- Python 3.11.0, pytest-7.4.1, pluggy-1.3.0
rootdir: C:\Users\redacted\pyo3_init_bug
collected 2 items

test_broken_exception.py F.                                                                                                                           [100%]

========================================================================= FAILURES ========================================================================= 
_______________________________________________________________ test_broken_exception[True] ________________________________________________________________ 

use_rust_exception = True

    @pytest.mark.parametrize("use_rust_exception", [True, False])
    def test_broken_exception(use_rust_exception: bool):
        with pytest.raises(Exception) as exc_info:
            try:
                raise ValueError("Original message")
            except Exception:
                if use_rust_exception:
                    raise RustException
                else:
                    raise PythonException

        print("".join(format_exception(exc_info.value)))
>       assert exc_info.value.__context__ is not None
E       AssertionError: assert None is not None
E        +  where None = TypeError('No constructor defined').__context__
E        +    where TypeError('No constructor defined') = <ExceptionInfo TypeError('No constructor defined') tblen=1>.value

test_broken_exception.py:23: AssertionError
------------------------------------------------------------------- Captured stdout call ------------------------------------------------------------------- 
  File "C:\Users\redacted\pyo3_init_bug\test_broken_exception.py", line 18, in test_broken_exception
    raise RustException
TypeError: No constructor defined

========================================================================== PASSES ========================================================================== 
_______________________________________________________________ test_broken_exception[False] _______________________________________________________________ 
------------------------------------------------------------------- Captured stdout call ------------------------------------------------------------------- 
  File "C:\Users\redacted\pyo3_init_bug\test_broken_exception.py", line 15, in test_broken_exception
    raise ValueError("Original message")
ValueError: Original message

During handling of the above exception, another exception occurred:

  File "C:\Users\redacted\pyo3_init_bug\test_broken_exception.py", line 20, in test_broken_exception
    raise PythonException
  File "C:\Users\redacted\pyo3_init_bug\test_broken_exception.py", line 8, in __new__
    raise TypeError("No constructor defined")
TypeError: No constructor defined

=============================================================== 1 failed, 1 passed in 0.16s ================================================================
```

Windows 11

Python 3.11.0

rustc 1.70.0 (90c541806 2023-05-31)

### Your PyO3 version

0.19.0

Standard Windows installer. I used a venv.

### Additional Info

I ran into this issue with Pydantic's [ValidationError](https://github.com/pydantic/pydantic-core/blob/9ec1036b4583d5ada15f7ac3e32fcc04a981746a/src/errors/validation_exception.rs#L30) type, which doesn't have a constructor. This issue has more details on that particular use case: https://github.com/pypa/hatch/issues/959

davidhewitt:
Fixes #3432 

Our current implementation of `PyErr::restore` (before 3.12) doesn't propagate exception context when inside of an `except` block. This appears to be the root cause of #3432.

This PR attempts to resolve that and make the code paths between 3.12 and older versions more similar. I think there's more tidying up to do and might even be worth splitting into separate PRs to make motivations clear. Pushing this here so the code doesn't get forgotten while I multitask. Don't think it's necessarily worth reviewing yet.

### DIFF
--- a/pytests/src/pyclasses.rs
+++ b/pytests/src/pyclasses.rs
@@ -58,10 +58,14 @@ impl AssertingBaseClass {
     }
 }
 
+#[pyclass]
+struct ClassWithoutConstructor;
+
 #[pymodule]
 pub fn pyclasses(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
     m.add_class::<EmptyClass>()?;
     m.add_class::<PyClassIter>()?;
     m.add_class::<AssertingBaseClass>()?;
+    m.add_class::<ClassWithoutConstructor>()?;
     Ok(())
 }
--- a/pytests/tests/test_pyclasses.py
+++ b/pytests/tests/test_pyclasses.py
@@ -1,3 +1,4 @@
+from typing import Type
 import pytest
 from pyo3_pytests import pyclasses
 
@@ -32,7 +33,29 @@ class AssertingSubClass(pyclasses.AssertingBaseClass):
 
 
 def test_new_classmethod():
-    # The `AssertingBaseClass` constructor errors if it is not passed the relevant subclass.
+    # The `AssertingBaseClass` constructor errors if it is not passed the
+    # relevant subclass.
     _ = AssertingSubClass(expected_type=AssertingSubClass)
     with pytest.raises(ValueError):
         _ = AssertingSubClass(expected_type=str)
+
+
+class ClassWithoutConstructorPy:
+    def __new__(cls):
+        raise TypeError("No constructor defined")
+
+
+@pytest.mark.parametrize(
+    "cls", [pyclasses.ClassWithoutConstructor, ClassWithoutConstructorPy]
+)
+def test_no_constructor_defined_propagates_cause(cls: Type):
+    original_error = ValueError("Original message")
+    with pytest.raises(Exception) as exc_info:
+        try:
+            raise original_error
+        except Exception:
+            cls()  # should raise TypeError("No constructor defined")
+
+    assert exc_info.type is TypeError
+    assert exc_info.value.args == ("No constructor defined",)
+    assert exc_info.value.__context__ is original_error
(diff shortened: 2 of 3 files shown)


davidhewitt:
Replaced by #3455

Issue #576, opened 2019-08-14
Multiple Python Interpreters


acm0055:
Does pyo3 allow the use case of starting multiple separate interpreters? This would be similar to Python's multiprocessing.

thedrow:
It should be possible with the upcoming improvements in 3.8 and above.

reem:
Specifically would be amazing if it was possible to create multiple separate python interpreters that could be run on different threads in parallel, but which share the same memory space (with the type system used to ensure this is only observable from rust).

davidhewitt:
The complexity with sub-interpreters is that no Python objects should ever be shared between them. The pyO3 API isn't set up to prevent this at the moment, so it's going to take a fair whack of experimentation before we get anything stable in place.

clouds56:
I'd like to have multiple thread, each one has an interpreter. No PyObject would be send between thread.
IIRC, we couldn't hold a PyObject in rust, everything is ref to `UnsafeCell` and by default not send.

mohitreddy1996:
@davidhewitt with https://peps.python.org/pep-0684/, using sub interpreters or multiple interpreters to unlock true multi-core parallelism becomes possible.

Is adding support for this in pyo3 timelines or consideration?

AnatolyBuga:
> @davidhewitt with https://peps.python.org/pep-0684/, using sub interpreters or multiple interpreters to unlock true multi-core parallelism becomes possible.
> 
> Is adding support for this in pyo3 timelines or consideration?

Found [this](https://martinheinz.dev/blog/97) interesting article on current usage of sub interpreters in python (no rust there)

davidhewitt:
We are very aware of the per-intepreter parallelism landing in Python 3.12. There are significant changes which need to happen to PyO3's current implementation to support this correctly. We have been discussing some of these challenges in multiple discussions across this repo, such as #2885 which looks at the possible nogil option.

There are several main issues which are prominent in my mind, although others may exist:

- I understand interpreters cannot share Python objects. This implies that `Py<T>` needs to be removed or reworked significantly, maybe by removing `Send` and `Sync` from that type, probably also by somehow making the operation to attach `Py<T>` to a Python thread to be `unsafe` or runtime-checked in some way.
- We need to fully transition PyO3 to PEP 630 compatibility, which requires elimination of all `static` data which contains Python state. This is probably linked to the first bullet.
[...]

PR #3446, opened 2023-09-11
relax multiple-import check to only prevent subinterpreters

davidhewitt:
Some issues observed in the wild from the current implementation of the multiple-import check:
- https://github.com/sphinx-doc/sphinx/issues/11662
- https://github.com/pydantic/pydantic/issues/6584

The motivation for the check is from https://github.com/PyO3/pyo3/discussions/2346#discussioncomment-2913833 - it's to prevent use of PyO3 modules in sub-interpreters, which we are definitely not ready for yet.

Given the issues above don't actually use sub-interpreters, it seemed worth trying to fix them. This PR does so by storing the current interpreter ID when initializing a `#[pymodule]`, and checking that does not change on subsequent imports.

I've also updated the error message to point at #576 so that discussion on the community can at least be centralized here.

### DIFF
--- a/pyo3-build-config/src/lib.rs
+++ b/pyo3-build-config/src/lib.rs
@@ -146,6 +146,11 @@ pub fn print_feature_cfgs() {
     if rustc_minor_version >= 59 {
         println!("cargo:rustc-cfg=thread_local_const_init");
     }
+
+    // Enable use of OnceLock on Rust 1.70 and greater
+    if rustc_minor_version >= 70 {
+        println!("cargo:rustc-cfg=once_lock");
+    }
 }
 
 /// Private exports used in PyO3's build.rs
--- a/pyo3-build-config/src/lib.rs
+++ b/pyo3-build-config/src/lib.rs
@@ -146,11 +146,6 @@ pub fn print_feature_cfgs() {
     if rustc_minor_version >= 59 {
         println!("cargo:rustc-cfg=thread_local_const_init");
     }
-
-    // Enable use of OnceLock on Rust 1.70 and greater
-    if rustc_minor_version >= 70 {
-        println!("cargo:rustc-cfg=once_lock");
-    }
 }
 
 /// Private exports used in PyO3's build.rs
--- a/src/impl_/pymodule.rs
+++ b/src/impl_/pymodule.rs
@@ -2,7 +2,7 @@
 
 use std::cell::UnsafeCell;
 
-#[cfg(all(not(PyPy), Py_3_9))]
+#[cfg(all(not(PyPy), Py_3_9, not(all(windows, Py_LIMITED_API, not(Py_3_10)))))]
 use std::sync::atomic::{AtomicI64, Ordering};
 
 #[cfg(not(PyPy))]
@@ -15,7 +15,7 @@ pub struct ModuleDef {
     ffi_def: UnsafeCell<ffi::PyModuleDef>,
     initializer: ModuleInitializer,
     /// Interpreter ID where module was initialized (not applicable on PyPy).
-    #[cfg(all(not(PyPy), Py_3_9))]
+    #[cfg(all(not(PyPy), Py_3_9, not(all(windows, Py_LIMITED_API, not(Py_3_10)))))]
     interpreter: AtomicI64,
     /// Initialized module object, cached to avoid reinitialization.
     module: GILOnceCell<Py<PyModule>>,
@@ -58,7 +58,7 @@ impl ModuleDef {
             ffi_def,
             initializer,
             // -1 is never expected to be a valid interpreter ID
-            #[cfg(all(not(PyPy), Py_3_9))]
+            #[cfg(all(not(PyPy), Py_3_9, not(all(windows, Py_LIMITED_API, not(Py_3_10)))))]
             interpreter: AtomicI64::new(-1),
             module: GILOnceCell::new(),
         }
@@ -86,7 +86,9 @@ impl ModuleDef {
         // PyPy does not have subinterpreters, so no need to check interpreter ID.
         #[cfg(not(PyPy))]
         {
-            #[cfg(Py_3_9)]
+            // PyInterpreterState_Get is only available on 3.9 and later, but is missing
+            // from python3.dll for Windows stable API on 3.9
+            #[cfg(all(Py_3_9, not(all(windows, Py_LIMITED_API, not(Py_3_10)))))]
             {
                 let current_interpreter =
                     unsafe { ffi::PyInterpreterState_GetID(ffi::PyInterpreterState_Get()) };
@@ -104,7 +106,7 @@ impl ModuleDef {
                     }
                 }
             }
-            #[cfg(not(Py_3_9))]
+            #[cfg(not(all(Py_3_9, not(all(windows, Py_LIMITED_API, not(Py_3_10))))))]
             {
                 // CPython before 3.9 does not have APIs to check the interpreter ID, so best that can be
                 // done to guard against subinterpreters is fail if the module is initialized twice
--- a/src/impl_/pymodule.rs
+++ b/src/impl_/pymodule.rs
@@ -1,23 +1,24 @@
 //! Implementation details of `#[pymodule]` which need to be accessible from proc-macro generated code.
 
 use std::cell::UnsafeCell;
-#[cfg(once_lock)]
-use std::sync::OnceLock;
 
-#[cfg(not(once_lock))]
-use parking_lot::Mutex;
+#[cfg(all(not(PyPy), Py_3_9))]
+use std::sync::atomic::{AtomicI64, Ordering};
 
-use crate::{exceptions::PyImportError, ffi, types::PyModule, Py, PyResult, Python};
+#[cfg(not(PyPy))]
+use crate::exceptions::PyImportError;
+use crate::{ffi, sync::GILOnceCell, types::PyModule, Py, PyResult, Python};
 
 /// `Sync` wrapper of `ffi::PyModuleDef`.
 pub struct ModuleDef {
     // wrapped in UnsafeCell so that Rust compiler treats this as interior mutability
     ffi_def: UnsafeCell<ffi::PyModuleDef>,
     initializer: ModuleInitializer,
-    #[cfg(once_lock)]
-    interpreter: OnceLock<i64>,
-    #[cfg(not(once_lock))]
-    interpreter: Mutex<Option<i64>>,
+    /// Interpreter ID where module was initialized (not applicable on PyPy).
+    #[cfg(all(not(PyPy), Py_3_9))]
+    interpreter: AtomicI64,
+    /// Initialized module object, cached to avoid reinitialization.
+    module: GILOnceCell<Py<PyModule>>,
 }
 
 /// Wrapper to enable initializer to be used in const fns.
@@ -56,10 +57,10 @@ impl ModuleDef {
         ModuleDef {
             ffi_def,
             initializer,
-            #[cfg(once_lock)]
-            interpreter: OnceLock::new(),
-            #[cfg(not(once_lock))]
-            interpreter: Mutex::new(None),
+            // -1 is never expected to be a valid interpreter ID
+            #[cfg(all(not(PyPy), Py_3_9))]
+            interpreter: AtomicI64::new(-1),
+            module: GILOnceCell::new(),
         }
     }
     /// Builds a module using user given initializer. Used for [`#[pymodule]`][crate::pymodule].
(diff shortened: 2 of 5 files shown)


src/impl_/pymodule.rs
```diff
@@ -12,7 +14,10 @@ pub struct ModuleDef {
     // wrapped in UnsafeCell so that Rust compiler treats this as interior mutability
     ffi_def: UnsafeCell<ffi::PyModuleDef>,
     initializer: ModuleInitializer,
-    initialized: AtomicBool,
+    #[cfg(once_lock)]
+    interpreter: OnceLock<i64>,
```
adamreichold, line 18: From https://docs.python.org/3/c-api/init.html#c.PyInterpreterState_GetID, I think we can use `-1` as a manual niche, i.e. just turn this into a `AtomicI64` initialized to `-1` and considered uninitialized as long as it has this value. This should be faster and simpler than both `OnceLock<i64>` and `Mutex<Option<i64>>`.

src/impl_/pymodule.rs
```diff
@@ -74,9 +82,22 @@ impl ModuleDef {
...
+                *self.interpreter.lock().get_or_insert(current_interpreter)
+            }
+        });
+        if current_interpreter != initialized_interpreter {
             return Err(PyImportError::new_err(
-                "PyO3 modules may only be initialized once per interpreter process",
+                "PyO3 modules do not yet support subinterpreters, see https://github.com/PyO3/pyo3/issues/576",
             ));
         }
         (self.initializer.0)(py, module.as_ref(py))?;
```
adamreichold, line 103: Do we really want to call the initializer multiple times or should the second import become a no-op?

davidhewitt: I think the spirit of PEP 489 (which we don't support yet, but hopefully are heading towards) is that modules can be torn down and re-imported, so I think it's correct to call the initializer multiple times.

src/impl_/pymodule.rs
```diff
@@ -74,9 +82,22 @@ impl ModuleDef {
         let module = unsafe {
             Py::<PyModule>::from_owned_ptr_or_err(py, ffi::PyModule_Create(self.ffi_def.get()))?
         };
-        if self.initialized.swap(true, atomic::Ordering::SeqCst) {
+        let current_interpreter =
+            unsafe { ffi::PyInterpreterState_GetID(ffi::PyInterpreterState_Get()) };
```
davidhewitt, line 86: Looks like `PyInterpreterState_Get` was only added on Python 3.9. On 3.8 this was private `_PyInterpreterState_Get`.

So on 3.7 builds (or 3.8 abi3) we can't improve this check, I will adjust this PR to use the old one in those cases. The old check still works, it's just not as nice.

src/impl_/pymodule.rs
adamreichold: > which we don't support yet, but hopefully are heading towards

I think this is the main issue here, isn't it? Does this change mean we opt into allowing tear-down and reinitialization of PyO3-based modules? What invariants would this break? Do we test this somewhere?

(I am not trying to say that it doesn't work. Just that I myself do understand the consequences of this change and little confidence about its effects on existing code.)

src/impl_/pymodule.rs
adamreichold: In https://github.com/pydantic/pydantic/issues/6584#issuecomment-1673260886, you mention opt-in. Might this be advisable even for this limited relaxation? What if the module is torn down but `static` data still contains pointers into the old module (e.g. capsules) which are now invalid? For example, is the code in https://github.com/PyO3/rust-numpy/blob/c16fbb1e630b0538fdf17cbc53043bf0467459f9/src/npyffi/mod.rs#L30-L32 valid after this change?

src/impl_/pymodule.rs
adamreichold: In https://github.com/pydantic/pydantic/issues/6584#issuecomment-1673260886, you mention opt-in. Might this be advisable even for this limited relaxation? What if the module is torn down but static data still contains pointers into the old module (e.g. capsules) which are now invalid? For example, is the code in https://github.com/PyO3/rust-numpy/blob/c16fbb1e630b0538fdf17cbc53043bf0467459f9/src/npyffi/mod.rs#L30-L32 valid after this change?

letalboy:
Hey, did you guys see what I put in the issue #576 ? I didn't know that have this issue here to get around use multiple sub interpreters, I see that you guys also starting some activity in the multi-interpreters issue, now that I'm kind understanding what is going on I will help with the changes that you are proposing to made sub-interprers viable, however to the code I have done don't be waste for nothing you guys can use here some parts if made any sense, I will put MIT license in the repo RustPyNets that I uploaded into my Git, maybe can get some ideas, I have done a code that do something like what this issue proposes, hope you guys like, and sorry to send things that is more related to this topic on the other one... I didn't know that had this one to this, has a lot of topics I wasn't able to see all rs 😅

src/impl_/pymodule.rs
davidhewitt: There is `PyErr_Occurred` (or other APIs which read the error indicator) which can be used to disambiguate -1 return values between success and error cases. For example `PyLong_AsLong`.

I think in this case you're right that interpreter IDs are implemented by a counter starting at 0 and then increment upwards. There is always the worry that this changes in the future, but hey, maybe by then we'll have an MSRV with `OnceLock` stable 😝 

I note also that I haven't actually handled the -1 error case here, so I need to correct for that too.

src/impl_/pymodule.rs
davidhewitt: So I found the Cython implementation here: 

https://github.com/cython/cython/blob/ce1aa59ab6300e18e50c752b5c9e28a04b665060/Cython/Utility/ModuleSetupCode.c#L1648C18-L1648C18

That code:
 - Has a similar check that the interpreter hasn't changed ID
 - Caches a module instance in a C static and returns the single module instance always.

So I guess let's implement this strategy too and defer the question of PEP 489 compatibility to a future PR. My goal for this PR is primarily to fix the use cases in the OP, which I believe the Cython strategy would cover.

pyo3-build-config/src/lib.rs
```diff
@@ -146,6 +146,11 @@ pub fn print_feature_cfgs() {
     if rustc_minor_version >= 59 {
         println!("cargo:rustc-cfg=thread_local_const_init");
     }
+
+    // Enable use of OnceLock on Rust 1.70 and greater
+    if rustc_minor_version >= 70 {
+        println!("cargo:rustc-cfg=once_lock");
```
mejrs, line 152: Can we please not do this? This code will be significantly simpler without conditional compilation, and I don't see the benefit.

adamreichold: Hence, the suggestion to just use `AtomicI64` in https://github.com/PyO3/pyo3/pull/3446#discussion_r1321606949

davidhewitt: Yep I'm expecting to remove this later 😁

davidhewitt:
Ok, I've pushed a new commit to this PR which:
- Removes the conditional code with `OnceLock` to just use `AtomicI64`
- Keeps the existing behaviour on CPython 3.7 and 3.8, where we cannot detect subinterpreter changes
- Makes 3.9 and newer cache the module object in a `GILOnceCell` and repeatedly return it _as long as the interpreter is the same one which created the module object_.

The use of `GILOnceCell` may at first seem at odds with the discussion in #576 to remove it, but hey, I figure this is explicitly aimed at preventing subinterpreters so it seems fine to use it for now 😄

davidhewitt:
(I'm not a huge fan of the divergence of behaviour between Python 3.7/3.8 and 3.9+, but I think in this case it is justified as the 3.9 behaviour is now a relaxation of the existing constraint to improve ergonomics where we can safely do so.)

src/impl_/pymodule.rs
```diff
@@ -71,16 +80,47 @@ impl ModuleDef {
...
+                if initialized_interpreter != current_interpreter {
+                    return Err(PyImportError::new_err(
+                        "PyO3 modules do not yet support subinterpreters, see https://github.com/PyO3/pyo3/issues/576",
+                    ));
+                }
+            }
         }
-        (self.initializer.0)(py, module.as_ref(py))?;
-        Ok(module)
+        #[cfg(all(not(PyPy), not(Py_3_9)))]
```
adamreichold, line 102: I am a bit confused by the PyPy handling. I would have expected `cfg(any(PyPy, not(Py_3_9)))` here.

As it is written importing on PyPy always succeeds and returns a reference to the existing object. Is this intentional? If so, I would be glad if we could add a code comment to make it explicit.

src/impl_/pymodule.rs
adamreichold: While I think this strategy is significantly safer, I am still hesitant. While this should resolve keeping static data pointing into the modules valid, what about the other way around, i.e. the module containing data bound to the interpreter? Is it guaranteed that finalizing and re-initializing a single interpreter in the same process will yield a new interpreter ID? (I guess we lack the API to actually test ATM?)

src/impl_/pymodule.rs
davidhewitt: Will do. The idea is that on PyPy there's no such thing as a subinterpreter yet, so no interpreter check is necessary (it would always succeed).

src/impl_/pymodule.rs
davidhewitt: That's a great question. I might be able to craft a test using `with_embedded_interpreter`, will try tomorrow.

davidhewitt: TLDR: finalizing and re-initializing yields the same ID, Cython is similarly bugged here, and [`Py_Finalize`]https://docs.python.org/3/c-api/init.html#c.Py_FinalizeEx specifically calls out that things aren't generally expected to work more than once. (I made `with_embedded_python_interpreter` unsafe for this reason.)

It's clearly awkward to get this right. I think on balance it's desirable if we match Cython so that more of the ecosystem fails in the same way and it seems better that we only ever run the module initializer once for now. We can make progress towards per-module state to eventually remove this worry.

---

The interpreter ID is not incremented. For example, this code doesn't hit our safety check (the interpreter ID is 0 both times):

```rust
use pyo3::{prelude::*, wrap_pymodule};

#[pymodule]
fn my_module(_py: Python, _m: &PyModule) -> PyResult<()> {
    Ok(())
}

#[test]
fn test_reloading_interpreter_doesnt_allow_module() {
    unsafe {
        pyo3::with_embedded_python_interpreter(|py| wrap_pymodule!(my_module)(py));

        pyo3::with_embedded_python_interpreter(|py| wrap_pymodule!(my_module)(py));
    }
}
```

I tried a variant of this with a Cython module, and I get a crash which looks like it's related to extension types being incorrectly reused across interpreters:

```
#[test]
fn test_reloading_interpreter_doesnt_allow_module() {
    unsafe {
        pyo3::with_embedded_python_interpreter(|py| {
            py.import("yaml._yaml").unwrap();
        });

        pyo3::with_embedded_python_interpreter(|py| {
            py.import("yaml._yaml").map_err(|e| e.print(py)).unwrap();
        });
    }
}
```

adamreichold: > It's clearly awkward to get this right. I think on balance it's desirable if we match Cython so that more of the ecosystem fails in the same way and it seems better that we only ever run the module initializer once for now. We can make progress towards per-module state to eventually remove this worry.

I agree with that. Do you know the upstream position on this, i.e. would they consider fixing this or does it work as intended?

> (I made with_embedded_python_interpreter unsafe for this reason.)

I am somewhat confused as to what the consequences of this discovery are. Especially with the Cython module immediately showing the failure mode I was concerned about. Can we still relax this? Is the `unsafe` on `with_embedded_python_interpreter` sufficient? (I guess another other way to re-initialize Python in the current process would call unsafe Rust code or code in inherently unsafe languages like C, i.e. is trusted?)

src/impl_/pymodule.rs
davidhewitt: Reworked to put the whole check in a `#[cfg(not(PyPy))]` with a comment.

src/impl_/pymodule.rs
davidhewitt: I have just posted https://github.com/python/cpython/issues/109785 - let's see what the outcome of that discussion is.

davidhewitt: No movement yet on the upstream issue. @adamreichold are you happy if I proceed to merge this? As above at least we'll then be consistent with Cython and I think the generally accepted position is that things are expected to be wonky after finalizing and re-initializing the interpreter.

adamreichold: Yes, I think the language on `with_embedded_python_interpreter` is strong enough. This is the only API we provide to finalize and then re-initialize the interpreter?

adamreichold (approved):

src/impl_/pymodule.rs
davidhewitt: Yes (apart from the raw FFI bindings). `auto-initialize` never bothers finalizing the interpreter.

linas:
Continues to be an issue in Dec 2024. Users of python crypto modules are affected, including Ceph, which is trying to create X509 certificates, are hit by this bug. I've collected links to some of the affected projects here. https://github.com/pyca/cryptography/issues/12080

It appears that PyO3 is always downstream of all of these issues. Is there some way of refocusing attention to get this fixed?

PR #3449, opened 2023-09-12
Add __pyo3__ attribute to native modules

### LINKED ISSUE #3426: Add '__pyo3_version__` to native modules compiled with PyO3
From https://github.com/sphinx-doc/sphinx/issues/11662#issuecomment-1702462095

> To be cheeky, could I ask if there is there a 'blessed' way to introspect if a module is PyO3 generated? It seems the variables mod.build_profile and mod.build_info might be heuristics, but a __PYO3__ = True flag or similar (a la '__pypy__' in sys.modules) might be useful for future cases?

I can't think of one which exists right now, it seems like a practical improvement to have `__pyo3_version__` as a standard attribute in compiled PyO3 modules?

ongchi:
Closes #3426

### DIFF
--- a/newsfragments/3449.added.md
+++ b/newsfragments/3449.added.md
@@ -1,0 +1,1 @@
+Add '__pyo3__' attribue to native modules
--- a/src/impl_/pymodule.rs
+++ b/src/impl_/pymodule.rs
@@ -5,7 +5,12 @@ use std::{
     sync::atomic::{self, AtomicBool},
 };
 
-use crate::{exceptions::PyImportError, ffi, types::PyModule, Py, PyResult, Python};
+use crate::{
+    exceptions::PyImportError,
+    ffi,
+    types::{PyModule, PyO3Attr},
+    Py, PyResult, Python,
+};
 
 /// `Sync` wrapper of `ffi::PyModuleDef`.
 pub struct ModuleDef {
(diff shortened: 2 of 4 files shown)


mejrs:
Thanks!

What is your motivation for using a (handwritten?) pyclass for this? We've so far avoided this because pyo3 is linked statically; when two users write their own pyo3 modules, they'll have two different `PyO3Attr` types. (See the discussion around `PanicException` for why this can be problematic).

adamreichold:
> We've so far avoided this because pyo3 is linked statically; when two users write their own pyo3 modules, they'll have two different PyO3Attr types. (See the discussion around PanicException for why this can be problematic).

I think the consensus reached so far is that we should ship a separate (pure Python for now) package called something like `pyo3_runtime` on PyPI which would contain shared types like `PanicException` and `PyO3Attr`, c.f. the discussion in the (eventually rejected) approach proposed in #3073 .

davidhewitt:
Agreed, I've been meaning to do that for aaaaages, and finally did it! 😂 

#3450 
https://pypi.org/project/pyo3-runtime

ongchi:
> What is your motivation for using a (handwritten?) pyclass for this?

All because I'm noob.🤡

BTW, I think it will be nice if pyo3 modules include some build information (e.g. rust version, build target, features...) automatically with a general interface to fetch this info. This shall benefit both developers and users who install packages from pypi, could providing useful context for debugging/issue tracking.

adamreichold:
> BTW, I think it will be nice if pyo3 modules include some build information (e.g. rust version, build target, features...) automatically with a general interface to fetch this info. This shall benefit both developers and users who install packages from pypi, could providing useful context for debugging/issue tracking.

I think https://github.com/PyO3/pyo3-built might be what you are looking for?

ongchi:
Cool! I'm not aware of that. Thanks for pointing this out.

PR #3450, opened 2023-09-12
add pyo3-runtime stub package

davidhewitt:
As per discussion in #3073 

I just created a new project using hatch, stripped it down a bit and pushed it to PyPI. Now we've got the name reserved, let's merge this blank slate here and add a `PanicException` and maybe `PyO3Attr` from #3449 as follow up PRs...

https://pypi.org/project/pyo3-runtime


davidhewitt:
For now I will mark this as CI-skip-changelog, we can add a note to `Packaging` later when we actually change the Rust code to use this.

adamreichold (approved):

adamreichold:
> For now I will mark this as CI-skip-changelog, we can add a note to `Packaging` later when we actually change the Rust code to use this.

We should probably also start adding QA for this to our CI or is this already covered by the existing Nox sessions?

davidhewitt:
Indeed it's not covered by CI, or anything, that's all to come. This is literally just a dump of the published files so that we're working from the same base ;)

Issue #3424, opened 2023-08-31
Handling of leap seconds in `chrono` feature is probably wrong


davidhewitt:
While investigating our current CI break and #3266...

At the moment there is some [code in our `chrono` into-python conversion](https://github.com/PyO3/pyo3/blob/cc1aa598cd3fc11aead82e5b164a519ce405b7e7/src/conversions/chrono.rs#L137-L142) which checks for leap-seconds, and in the case of a leap-second, sets `fold=1` for the resultant Python datetime object.

Similarly on the `FromPyObject` leg, if `fold=1` we convert the result back into a leap-second. Although it makes our conversion above round-trip, *this is almost certainly wrong in the general case of #3266, because `fold` is meant to be used for DST, not leap seconds.*

NB that `datetime` does not even support leap seconds! https://github.com/python/cpython/issues/67762

I think we should remove this attempt to round-trip leap-seconds. Given that the only option to fail on the to-python leg is to panic (pending #1813), I'm slightly leaning towards clamping the leap-second and raising a warning. Even with fallible `IntoPy` I wonder if it might be better to do that than fail for a very select edge-case which (if we ever get another leap-second) will probably be at midnight UTC.

Opinions welcome, I will probably try to push an implementation / fix tomorrow.

PR #3456, opened 2023-09-16
Add conversion support for `either::Either`

aldanor:
Added conversion support for [`either`](http://docs.rs/either)'s `Either<L, R>` type, equivalent to `Union[L, R]` on the Python side.

It's a tiny but [pretty popular crate](https://lib.rs/crates/either), with 24k dependent crates and 6-7M downloads/m.

It also simplifies a very common use case of accepting or returning one of the two types, without having to define a helper enum for each particular case.

### DIFF
--- a/src/conversions/either.rs
+++ b/src/conversions/either.rs
@@ -1,0 +1,146 @@
+#![cfg(feature = "either")]
+
+//! Conversion to/from
+//! [either](https://docs.rs/either/ "A library for easy idiomatic error handling and reporting in Rust applications")’s
+//! [`Either`] type to a union of two Python types.
+//!
+//! Use of a generic sum type like [either] is common when you want to either accept one of two possible
+//! types as an argument or return one of two possible types from a function, without having to define
+//! a helper type manually yourself.
+//!
+//! # Setup
+//!
+//! To use this feature, add this to your **`Cargo.toml`**:
+//!
+//! ```toml
+//! [dependencies]
+//! ## change * to the version you want to use, ideally the latest.
+//! either = "*"
+#![doc = concat!("pyo3 = { version = \"", env!("CARGO_PKG_VERSION"),  "\", features = [\"either\"] }")]
+//! ```
+//!
+//! Note that you must use compatible versions of either and PyO3.
+//! The required either version may vary based on the version of PyO3.
+//!
+//! # Example: Convert a `int | str` to `Either<i32, String>`.
+//!
+//! ```rust
+//! use either::Either;
+//! use pyo3::{Python, ToPyObject};
+//!
+//! fn main() {
+//!     pyo3::prepare_freethreaded_python();
+//!     Python::with_gil(|py| {
+//!         // Create a string and an int in Python.
+//!         let py_str = "crab".to_object(py);
+//!         let py_int = 42.to_object(py);
+//!         // Now convert it to an Either<i32, String>.
+//!         let either_str: Either<i32, String> = py_str.extract(py).unwrap();
+//!         let either_int: Either<i32, String> = py_int.extract(py).unwrap();
+//!     });
+//! }
+//! ```
+//!
+//! [either](https://docs.rs/either/ "A library for easy idiomatic error handling and reporting in Rust applications")’s
+
+use crate::{
+    exceptions::PyTypeError, inspect::types::TypeInfo, FromPyObject, IntoPy, PyAny, PyObject,
+    PyResult, Python, ToPyObject,
+};
+use either::Either;
+
+#[cfg_attr(docsrs, doc(cfg(feature = "either")))]
+impl<L, R> IntoPy<PyObject> for Either<L, R>
+where
+    L: IntoPy<PyObject>,
+    R: IntoPy<PyObject>,
+{
+    #[inline]
+    fn into_py(self, py: Python<'_>) -> PyObject {
+        match self {
+            Either::Left(l) => l.into_py(py),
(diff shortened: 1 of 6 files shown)


aldanor:
Btw the errors that appear on CI:

```
---- conversions::chrono::tests::test_pyo3_time_frompyobject stdout ----
thread 'conversions::chrono::tests::test_pyo3_time_frompyobject' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'ValueError'>, value: ValueError('invalid or out-of-range time'), traceback: None }', src/conversions/chrono.rs:833:60
```

I was also seeing locally (on M1), was assuming it's some local problem, but apparently not...

adamreichold:
> I was also seeing locally (on M1), was assuming it's some local problem, but apparently not...

You might seeing fallout from #3424.

src/conversions/either.rs
```diff
@@ -0,0 +1,146 @@
+#![cfg(feature = "either")]
+
+//! Conversion to/from
+//! [either](https://docs.rs/either/ "A library for easy idiomatic error handling and reporting in Rust applications")’s
+//! [`Either`] type to a union of two Python types.
+//!
+//! Use of a generic sum type like [either] is common when you want to either accept one of two possible
```
adamreichold, line 7: Personally, I am not particularly fond of the `Either` type as I think it encourages type blindness, i.e. the two types usually have more specific roles than just `Left` and `Right` and defining an `enum` with `#[derive(FromPyObject)]` slapped onto it is usually the better approach IMHO.

Of course, the API surface of PyO3 might not be a suitable place to have this discussion. On the other hand, we to carefully consider the set of crates which we integrate directly and the list of potential candidates is unbounded.

src/conversions/either.rs
```diff
@@ -0,0 +1,146 @@
...
+#![doc = concat!("pyo3 = { version = \"", env!("CARGO_PKG_VERSION"),  "\", features = [\"either\"] }")]
+//! ```
+//!
+//! Note that you must use compatible versions of either and PyO3.
+//! The required either version may vary based on the version of PyO3.
+//!
+//! # Example: Convert a `int | str` to `Either<i32, String>`.
+//!
+//! ```rust
+//! use either::{Either};
```
adamreichold, line 28: These braces seem unnecessary.

src/conversions/either.rs
aldanor: Thanks for the comment!

> Personally, I am not particularly fond of the Either type as I think it encourages type blindness, i.e. the two types usually have more specific roles than just Left and Right and defining an enum with #[derive(FromPyObject)] slapped onto it is usually the better approach IMHO.

I would generally advise that as well in the general case, except when your code gets littered by `IntOrString`, `SeriesOrFrame`, `ThisOrThat` types that are only used on a single line of code.  Sometimes you just want a once-off sum-type which doesn't exist in stdlib (`Result` aside, which has a specific use-case), either for taking it as an argument, or for returning it, or just for manual extraction. The sum-type of two variants would be perhaps the most common of all in type signatures and would cover lots of simple cases; for anything more complex you could always reach out for a hand-written object with a FromPyObject-derive.

Another difference from the general case is that when using pyo3, you're (normally) not writing a crate to be used and shared within Rust, with a proper Rust API surface and proper enums crafted for each particular case; your actual API surface lives in Python land which doesn't really care if it's `IntOrString` or `Either<i32, String>` after it's all compiled. (Anyways, that gets pretty subjective so it's my personal take on this.)

> Of course, the API surface of PyO3 might not be a suitable place to have this discussion. On the other hand, we to carefully consider the set of crates which we integrate directly and the list of potential candidates is unbounded.

A few points re: `either`, it's been 1.0 for around 7 years, is a tiny crate, has a single exported type, zero dependencies and tons of dependent crates (mostly used as a internal type rather than part of the public API, but still). I wouldn't expect it to affect the API surface that much or increase maintenance burden.

adamreichold: > A few points re: either, it's been 1.0 for around 7 years, is a tiny crate, has a single exported type, zero dependencies and tons of dependent crates (mostly used as a internal type rather than part of the public API, but still). I wouldn't expect it to affect the API surface that much or increase maintenance burden.

Indeed, the maintenance effort seems to be bearable. I'd just like to avoid suggesting this approach to people, but we also do not put this on the frontend but it will be part of an ever growing list of integration features.

So let's put it like this: If this has any traction with the other maintainers, I consider my concerns overwritten as they are not severe enough to argue over for any amount of time.

davidhewitt:
Yes I'm working on a pr to fix the chrono problem, will try to push later

src/conversions/either.rs
davidhewitt: I have to agree with @adamreichold here, I've never wanted to use `either` myself as it's quite straightforward to derive most traits I want on a simple enum these days.

Similarly `either` is, as you say, very stable and it's unlikely we'll ever need to change this code other than for our own rearranging.

I'm probably a bit more lax than @adamreichold; I'd say that I myself don't want this but if there's multiple community members who want this then I can support it being merged.

adamreichold: So what do we do here? There has been no more feedback (neither for or against) and both me and David are reserved but not completely opposed.

I'd say let's honour putting in the work and merge this.

davidhewitt: To that end, I've just rebased this PR.

adamreichold (approved):

PR #3458, opened 2023-09-16
truncate leap seconds on `chrono` to `datetime` conversions

### LINKED ISSUE #3424: Handling of leap seconds in `chrono` feature is probably wrong

davidhewitt:
Fixes #3424 

As discussed in that issue, I decided to proceed to truncate leap-seconds and emit a warning when doing so.

Most the diff is adjustment to test code to adjust for the logic, check the warning is raised and also deal with increased strictness of when leap-seconds are allowed in `chrono` 0.4.31.

### DIFF
--- a/newsfragments/3458.changed.md
+++ b/newsfragments/3458.changed.md
@@ -1,0 +1,1 @@
+Truncate leap-seconds and warn when converting `chrono` types to Python `datetime` types (`datetime` cannot represent leap-seconds).
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -40,12 +40,12 @@
 //!     });
 //! }
 //! ```
-use crate::exceptions::{PyTypeError, PyValueError};
+use crate::exceptions::{PyTypeError, PyUserWarning, PyValueError};
 use crate::types::{
     timezone_utc, PyDate, PyDateAccess, PyDateTime, PyDelta, PyDeltaAccess, PyTime, PyTimeAccess,
     PyTzInfo, PyTzInfoAccess, PyUnicode,
 };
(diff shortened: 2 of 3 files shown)


davidhewitt:
I split out #3469, please review that first.

davidhewitt:
Also moved the warnings testing logic out into #3470.

davidhewitt:
Rebased; this now contains purely the behavioural changes plus updates to accompanying tests.

davidhewitt:
If I don't hear reasons not to, I'll merge this on Friday so that I can work on a fix for #3266 without having masses of conflicts.

PR #3466, opened 2023-09-21
docs: skip changelog check for docs-only PRs

davidhewitt:
Recently we've had a few contributors submit docs-only PRs and go through the trouble of submitting a newsfragment.

This PR adds the ability for PRs with `docs:` as the start of the title to skip the changelog check. IMO this is a nice ergonomics improvement for non-maintainers because they cannot apply the `CI-skip-changelog` label.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -416,6 +416,12 @@ def address_sanitizer(session: nox.Session):
     )
 
 
+_IGNORE_CHANGELOG_PR_CATEGORIES = (
+    "release",
+    "docs",
+)
+
+
 @nox.session(name="check-changelog")
 def check_changelog(session: nox.Session):
     event_path = os.environ.get("GITHUB_EVENT_PATH")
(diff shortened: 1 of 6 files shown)


noxfile.py
```diff
@@ -448,7 +455,9 @@ def check_changelog(session: nox.Session):
 
     if not fragments:
         session.error(
-            "Changelog entry not found, please add one (or more) to `newsfragments` directory. For more information see https://github.com/PyO3/pyo3/blob/main/Contributing.md#documenting-changes"
+            "Changelog entry not found, please add one (or more) to the `newsfragments` directory.\n"
+            "Alternatively, start the PR title with `docs:` if this PR is a docs-only PR.\n"
+            "See https://github.com/PyO3/pyo3/blob/main/Contributing.md#documenting-changes for more information."
```
davidhewitt, line 460: I reordered the last sentence because nox was appending a final full stop to the URL which was breaking the link in Github Actions (would send to the wrong anchor).

adamreichold (approved):

adamreichold:
Seems like this needs an `#[allow(invalid_from_utf8)]` on the `unicode_decode_error` test case.

PR #3469, opened 2023-09-23
refactor chrono implementation to be more DRY

davidhewitt:
Split out from #3458 

This reworks the `chrono` implementation to reuse all the logic to convert from / to naive dates and times as helper functions to help ensure consistency. There's no behaviour changes nor test changes in this PR, so it should be easy to review.

I've added FIXMES where we need to adjust behaviour for #3266 and #3424, which we can then close in separate PRs focussing on those cases.


davidhewitt:
I pushed a second commit to drive up test coverage.

adamreichold (approved):

PR #3470, opened 2023-09-23
add `assert_warnings` test helper

davidhewitt:
Split out from #3458.

This adds `CaptureWarnings` and `assert_warnings` helpers to `tests/common.rs`, and makes it possible to use those from unit tests inside `src` under the path `tests::common`.


src/tests/mod.rs
```diff
@@ -0,0 +1,10 @@
+#[macro_use]
+pub(crate) mod common {
+    use crate as pyo3;
+    include!("../../tests/common.rs");
```
adamreichold, line 4: I understand that it is additional churn, but I think having the common file placed at `src/tests/common.rs` and `include!`ing it using relative paths in the `tests` directory would be easier follow.

adamreichold (approved):

src/tests/mod.rs
davidhewitt: Sure thing, I was wondering similarly 👍

PR #3471, opened 2023-09-23
always normalize exceptions before raising

davidhewitt:
Split from #3455 

It seems like the part of the above PR which needs to be better understood is the use of `PyErr_SetObject`, so I've split out this PR which just ensures the exception is normalized when raising a "lazy" exception. Given that 3.12 already does this (by virtue of exceptions always being normalized) this is a consistency change for the older Pythons. It fixes the issue observed in #3439 (a test added).

To try to make things a bit simpler I've adjusted code paths a little bit so that `PyErrState::normalize` has just one implementation with different match arms for pre/post 3.12.

### DIFF
--- a/newsfragments/3471.fixed.md
+++ b/newsfragments/3471.fixed.md
@@ -1,0 +1,1 @@
+Fix `IterNextOutput::Return` not returning a value on PyPy.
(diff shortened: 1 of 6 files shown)


davidhewitt:
Rebased; this now has a very minimal diff in `src/err/err_state.rs` and everything else is test code. Hopefully this makes it easier to decide on the change of behaviour.

mejrs (approved):
LGTM

Issue #3452, opened 2023-09-13
Allow creating an Exception hierarchy


stinodego:
I'd like to be able to create an exception hierarchy in PyO3. In Python I would write something like:

```python
class PolarsError(Exception):
    ...

class OutOfBoundsError(PolarsError):
    ...
```

Then users could catch the specific error thrown by the code (`OutOfBoundsError`), or use catch more general exception type (`PolarsError`).

Right now, I'm using the `create_exception!` macro to define the exceptions, as mentioned in the user guide. However, this does not allow for specifying a hierarchy.

I've found some discussion on subclassing exceptions:
* https://github.com/PyO3/pyo3/pull/1591
* https://github.com/PyO3/pyo3/issues/295#issuecomment-852358088
* https://github.com/PyO3/pyo3/issues/732

...from this, it seems like my wish is already possible. However, I cannot find any code examples of how this is supposed to work. If it is already possible, could someone point me in the right direction? If not, consider this a feature request!

davidhewitt:
Hello, sorry for the slow reply.

This should definitely be possible and really this just needs proper documenting and testing, after which point I'd like to deprecate the `create_exception!` macro so we just have one blessed way of doing things.

I've opened #3475 which begins building such a test for an exception hierarchy, let's collaborate there to make sure that all the functionality you want is supported.

kylebarron:
For anyone curious, [this works now](https://github.com/developmentseed/obstore/blob/eb5fb933ac41aacdbb4d4ee82f848ad705667f06/pyo3-object_store/src/error.rs#L9-L25), and this issue can probably be closed.

```rs
// Base exception
create_exception!(
    pyo3_object_store,
    BaseError,
    pyo3::exceptions::PyException,
    "The base Python-facing exception from which all other errors subclass."
);
// Subclasses from base exception
create_exception!(
    pyo3_object_store,
    GenericError,
    BaseError,
    "A Python-facing exception wrapping [object_store::Error::Generic]."
);
```

It looks like polars figured this out as well: https://github.com/pola-rs/polars/pull/13615/files#diff-70f7a7b2eaa2591ee077ec6d5093f4beaccbb64e2ab07b7a069d5ac62eedee05R78-R79

PR #3475, opened 2023-09-27
add exception hierarchy pytests

davidhewitt:
WIP solution for #3452 

The goal here is to get examples and documentation working for how to create a Python exception hierarchy and interact with it from Rust. This will probably go through a few rounds of rebases as features / bugfixes inspired by it will be opened as separate PRs.


pytests/src/exceptions.rs
```diff
@@ -0,0 +1,72 @@
+use pyo3::exceptions::PyException;
+use pyo3::prelude::*;
+use pyo3::types::{PyDict, PyTuple};
+
+#[pyclass(extends=PyException, subclass)]
+struct CustomException {}
+
+#[pymethods]
+impl CustomException {
+    #[new(signature = (*_args, **_kwargs))]
```
davidhewitt, line 10: I wrote a bug here which confused me for a bit; `#[new]` ignores `signature`.

Either this should be a hard error or it should be equivalent to `#[new] #[pyo3(signature = (*_args, **_kwargs))]`.

davidhewitt:
@stinodego did you have any luck with an exception hierarchy on the polars side? I think this is a baseline which you might be able to start from, I think the most practical way forward is that if you try implementing what you want starting from this, please report back on your experience and we can fix any issues or papercuts.

stinodego:
Apologies for getting back to this so late. When I try to adapt this for Polars, I get an error that `PyClass` isn't implement for `PyException`. Any idea how to solve this? I must be doing something wrong.

davidhewitt:
Is polars using the `abi3` feature by any chance? That currently doesn't support subclassing native types (although I intend to fix that in the mid term).

stinodego:
Ah that's right - we are using the `abi3-py38` feature.

I have been able to get a basic version of this working though the `create_exception` macros - I didn't realize it would allow me to do basically what I had intended:

```rust
create_exception!(polars.exceptions, PolarsBaseError, PyException);
create_exception!(polars.exceptions, ComputeError, PolarsBaseError);
create_exception!(polars.exceptions, ColumnNotFoundError, PolarsBaseError);
```

davidhewitt:
In the long run I would like to deprecate the `create_exception` macro and only support the `#[pyclass]` route as proposed here, but you highlight a good point here that I probably need to solve #1344 first.

Issue #3482, opened 2023-09-29
Add a PyModule::from_pyc function


Fluffyalien1422:
@mejrs 
> If there is a C api function for that I wouldn't mind, but I don't see any.

This code works!
```rs
Python::with_gil(|_py| {
    let data = &include_bytes!("t.pyc")[16..];
    unsafe {
        let c_data_ptr = data.as_ptr() as *const i8;
        let obj = ffi::PyMarshal_ReadObjectFromString(c_data_ptr, data.len() as isize);

        let _mptr = ffi::PyImport_ExecCodeModuleEx(
            "t".as_ptr() as *const i8,
            obj,
            "t.pyc".as_ptr() as *const i8,
        );

        ffi::Py_DECREF(obj);
    }
});
```

I'll try to implement this and make a PR

You can see how this works by looking at the CPython repo
At [run_pyc_file in Python/pythonrun.c](https://github.com/python/cpython/blob/main/Python/pythonrun.c#L1742)
Specifically these lines:
```c
    magic = PyMarshal_ReadLongFromFile(fp);
    if (magic != PyImport_GetMagicNumber()) {
        if (!PyErr_Occurred())
            PyErr_SetString(PyExc_RuntimeError,
                       "Bad magic number in .pyc file");
        goto error;
    }
    /* Skip the rest of the header. */
    (void) PyMarshal_ReadLongFromFile(fp);
    (void) PyMarshal_ReadLongFromFile(fp);
    (void) PyMarshal_ReadLongFromFile(fp);
```

_Originally posted by @Fluffyalien1422 in https://github.com/PyO3/pyo3/discussions/3480#discussioncomment-7148961_

davidhewitt:
As we decided to close the PR, also closing here.

PR #3483, opened 2023-09-29
add PyModule::from_pyc (#3482)

Fluffyalien1422:
Adds a `PyModule::from_pyc` function. See #3482 and #3480.


Fluffyalien1422:
I have no idea how to fix the codecov/patch test. Help would be appreciated

assets/hello_world.pyc
davidhewitt, line 1: Personally I'm not very comfortable with shipping an opaque binary file of executable code in the git repository and running it as part of the test suite.

Also, doesn't the `.pyc` format have no stability guarantee across Python versions? 

I think a more robust solution would be to use APIs to compile a string sample into bytecode, and write that to a temporary `.pyc` file.

src/types/module.rs
```diff
@@ -149,6 +150,90 @@ impl PyModule {
...
+                return Err(PyErr::fetch(py));
+            }
+
+            let mptr = ffi::PyImport_ExecCodeModuleEx(module.as_ptr(), obj, filename.as_ptr());
+            ffi::Py_DecRef(obj);
+            if mptr.is_null() {
+                return Err(PyErr::fetch(py));
+            }
+
+            <&PyModule as crate::FromPyObject>::extract(py.from_owned_ptr_or_err(mptr)?)
```
davidhewitt, line 233: The error handling can be simplified:

```suggestion
            py.from_owned_ptr_or_err(
                ffi::PyImport_ExecCodeModuleEx(module.as_ptr(), obj, filename.as_ptr())
            )
```

mejrs:
I realize I never asked *why* you want this. What is your use case for embedding and executing `.pyc` files? The documentation should cover why someone would want to to do this (rather than use a different PyModule api or bundling their files with PyOxidizer).

I'm not familiar with `.pyc` files other than knowing it's some kind of bytecode format. My concerns are:
- I couldn't find guarantees about the `.pyc` format (this is probably why there is no C api function for running one). This means that we can't rely on it not changing, and this also means we can't just do what the (private) internals of the python interpreter do.
- The format depends on the Python version being used, this means you can't use this when you are running a different Python than the one the bytecode is compiled for. In other words - this is a gotcha for people using this and distributing their modules.

src/types/module.rs
```diff
@@ -149,6 +150,90 @@ impl PyModule {
...
+        // remove the header (see https://github.com/python/cpython/blob/14098b78f7453adbd40c53e32c29588611b7c87b/Python/pythonrun.c#L1758C39-L1758C39)
+        let data = &data[16..];
+
+        unsafe {
+            let c_data_ptr = data.as_ptr() as *const c_char;
+
+            let obj = ffi::PyMarshal_ReadObjectFromString(c_data_ptr, data.len() as isize);
+            if obj.is_null() {
+                return Err(PyErr::fetch(py));
+            }
```
davidhewitt, line 225: Maybe use the wrapper API `pyo3::marshal::loads` for this?

src/types/module.rs
```diff
@@ -149,6 +150,90 @@ impl PyModule {
...
+    /// let bytes = std::fs::read("assets/hello_world.pyc")?;
+    ///
+    /// Python::with_gil(|py| -> PyResult<()> {
+    ///     PyModule::from_pyc(py, &bytes, "example.py", "example")?;
+    ///     Ok(())
+    /// })?;
+    /// Ok(())
+    /// # }
+    /// ```
+    pub fn from_pyc<'p>(
```
davidhewitt, line 207: Should this function be `unsafe`? I think despite it being a security hazard I don't think it falls under Rust's typical definition of `unsafe`, so my argument is that it does not need to me.

davidhewitt (changes requested):
Thanks for the PR!

The patch coverage is failing because the error branches are not covered. I've posted some suggestions to make the implementation shorter, it may also not hurt to add a test for invalid `.pyc` data.

I do feel quite strongly that we should not ship a `.pyc` snippet in the repo. As you rightly point out in the test suite, it's dangerous to pass untrusted code to this function. We may be comfortable with our own asset but downstream packagers may not be.

mejrs:
>  I do feel quite strongly that we should not ship a .pyc snippet in the repo. As you rightly point out in the test suite, it's dangerous to pass untrusted code to this function. We may be comfortable with our own asset but downstream packagers may not be.

If we release this `from_pyc` function we are effectively encouraging users to ship pyc blobs with their modules - I'm not comfortable with that either.

src/types/module.rs
adamreichold: > I don't think it falls under Rust's typical definition of unsafe

Are you sure it cannot trigger undefined behaviour by passing invalid bytecode?

davidhewitt:
Given that all of us are feeling uncomfortable about security connotations of this one way or another, perhaps that is suggesting this is better to not merge and leave for downstream users to implement themselves if desired? Seems like we already expose the APIs to do so in pyo3-ffi.

mejrs:
I'm closing this per the above. Thank you for your efforts.

PR #3484, opened 2023-09-30
error on passing arguments to `#[new]` and similar attributes

davidhewitt:
This PR reworks `#[new]` to reject additional arguments rather than ignore them, as noted in https://github.com/PyO3/pyo3/pull/3475#discussion_r1338193900

We could change in the future to allow `#[new(signature = ())]` to be equivalent to `#[new] #[pyo3(signature = ())]` if desired, as a backwards-compatible change.

The bulk of the diff is because I threw out the old implementation of `parse_method_attributes` to make it compose better for the error messages I wanted to emit.

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -1,3 +1,5 @@
+use std::fmt::Display;
+
 use crate::attributes::{TextSignatureAttribute, TextSignatureAttributeValue};
 use crate::params::impl_arg_params;
 use crate::pyfunction::{FunctionSignature, PyFunctionArgPyO3Attributes};
@@ -9,7 +11,7 @@ use quote::ToTokens;
 use quote::{quote, quote_spanned};
 use syn::ext::IdentExt;
 use syn::spanned::Spanned;
-use syn::Result;
+use syn::{Ident, Result};
 
 #[derive(Clone, Debug)]
 pub struct FnArg<'a> {
@@ -67,24 +69,6 @@ fn handle_argument_error(pat: &syn::Pat) -> syn::Error {
         _ => "unsupported argument",
     };
     syn::Error::new(span, msg)
-}
-
-#[derive(Clone, PartialEq, Debug, Copy, Eq)]
-pub enum MethodTypeAttribute {
-    /// `#[new]`
-    New,
-    /// `#[new]` && `#[classmethod]`
-    NewClassMethod,
-    /// `#[classmethod]`
-    ClassMethod,
-    /// `#[classattr]`
-    ClassAttribute,
-    /// `#[staticmethod]`
-    StaticMethod,
-    /// `#[getter]`
-    Getter,
-    /// `#[setter]`
-    Setter,
 }
 
 #[derive(Clone, Debug)]
@@ -278,13 +262,10 @@ impl<'a> FnSpec<'a> {
             ..
         } = options;
 
-        let MethodAttributes {
-            ty: fn_type_attr,
-            mut python_name,
-        } = parse_method_attributes(meth_attrs, name.map(|name| name.value.0))?;
+        let mut python_name = name.map(|name| name.value.0);
 
         let (fn_type, skip_first_arg, fixed_convention) =
-            Self::parse_fn_type(sig, fn_type_attr, &mut python_name)?;
+            Self::parse_fn_type(sig, meth_attrs, &mut python_name)?;
         ensure_signatures_on_valid_method(&fn_type, signature.as_ref(), text_signature.as_ref())?;
 
         let name = &sig.ident;
@@ -331,9 +312,11 @@ impl<'a> FnSpec<'a> {
 
     fn parse_fn_type(
         sig: &syn::Signature,
-        fn_type_attr: Option<MethodTypeAttribute>,
+        meth_attrs: &mut Vec<syn::Attribute>,
         python_name: &mut Option<syn::Ident>,
     ) -> Result<(FnType, bool, Option<CallingConvention>)> {
+        let mut method_attributes = parse_method_attributes(meth_attrs)?;
+
         let name = &sig.ident;
         let parse_receiver = |msg: &'static str| {
             let first_arg = sig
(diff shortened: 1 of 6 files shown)


pyo3-macros-backend/src/method.rs
```diff
@@ -604,107 +603,123 @@ impl<'a> FnSpec<'a> {
...
+            Ok(Some(MethodTypeAttribute::ClassAttribute(path.span())))
+        } else if path.is_ident("getter") {
+            let name = extract_name(meta, "getter")?;
+            Ok(Some(MethodTypeAttribute::Getter(path.span(), name)))
+        } else if path.is_ident("setter") {
+            let name = extract_name(meta, "setter")?;
+            Ok(Some(MethodTypeAttribute::Setter(path.span(), name)))
+        } else {
+            Ok(None)
+        }
```
davidhewitt, line 705: This is essentially the new logic for these `#[pymethods]` attributes.

pyo3-macros-backend/src/method.rs
```diff
@@ -604,107 +603,123 @@ impl<'a> FnSpec<'a> {
...
+        let meta = &attr.meta;
+        let path = meta.path();
+
+        if path.is_ident("new") {
+            ensure_no_arguments(meta, "new")?;
+            Ok(Some(MethodTypeAttribute::New(path.span())))
+        } else if path.is_ident("__new__") {
+            // TODO deprecate this form?
+            ensure_no_arguments(meta, "__new__")?;
+            Ok(Some(MethodTypeAttribute::New(path.span())))
```
davidhewitt, line 687: I discovered `#[__new__]` is a valid attribute in our code, I think as this is undocumented and redundant with `#[new]` we should deprecate it.

tests/ui/invalid_pymethod_names.stderr
```diff
@@ -1,8 +1,8 @@
 error: `name` may only be specified once
-  --> tests/ui/invalid_pymethod_names.rs:10:19
+  --> tests/ui/invalid_pymethod_names.rs:11:14
    |
-10 |     #[pyo3(name = "num")]
-   |                   ^^^^^
+11 |     #[getter(number)]
+   |              ^^^^^^
```
davidhewitt, line 5: This test changed from erroring on the first of two attributes which specify name (`#[pyo3(name = "num")]` to erroring on the second site `(`#[getter(number)]`). Functionally no different as the code is rejected either way.

pyo3-macros-backend/src/method.rs
```diff
@@ -604,107 +603,123 @@ impl<'a> FnSpec<'a> {
...
-    ty: Option<MethodTypeAttribute>,
-    python_name: Option<syn::Ident>,
+enum MethodTypeAttribute {
+    New(Span),
+    ClassMethod(Span),
+    StaticMethod(Span),
+    Getter(Span, Option<Ident>),
+    Setter(Span, Option<Ident>),
+    ClassAttribute(Span),
 }
```
mejrs, line 613: I think most of our structs and enums carry around an `Ident`, not just the span, like...
```suggestion
enum MethodTypeAttribute {
    New(Ident),
    ...
}
```

Also ...
```rust
fn ident(&self) -> &'static str {
```

...it seems weird to not return an `Ident` here.

pyo3-macros-backend/src/method.rs
```diff
@@ -351,52 +332,70 @@ impl<'a> FnSpec<'a> {
...
-                )?),
-                true,
-                None,
-            ),
+            [first, second, ..] => {
+                bail_spanned!(
+                    second.span() =>
+                    format!("`#[{}]` may not be combined with `#[{}]`", first.ident(), second.ident())
+                )
+            }
```
mejrs, line 398: nit: this should list all the attributes to avoid a confusing error for
```rust
#[new]
#[classmethod] 
#[whatever]
fn method() {}
```

pyo3-macros-backend/src/method.rs
```diff
@@ -69,24 +69,6 @@ fn handle_argument_error(pat: &syn::Pat) -> syn::Error {
     syn::Error::new(span, msg)
 }
 
-#[derive(Clone, PartialEq, Debug, Copy, Eq)]
-pub enum MethodTypeAttribute {
-    /// `#[new]`
-    New,
-    /// `#[new]` && `#[classmethod]`
-    NewClassMethod,
```
mejrs, line 77: What is the thought behind yeeting the `NewClassMethod` variant? I have no particular opinion - just curious.

pyo3-macros-backend/src/method.rs
```diff
@@ -351,52 +332,70 @@ impl<'a> FnSpec<'a> {
                 .map(|stripped| syn::Ident::new(stripped, name.span()))
         };
 
-        let (fn_type, skip_first_arg, fixed_convention) = match fn_type_attr {
-            Some(MethodTypeAttribute::StaticMethod) => (FnType::FnStatic, false, None),
-            Some(MethodTypeAttribute::ClassAttribute) => (FnType::ClassAttribute, false, None),
-            Some(MethodTypeAttribute::New) | Some(MethodTypeAttribute::NewClassMethod) => {
+        let (fn_type, skip_first_arg, fixed_convention) = match method_attributes.as_mut_slice() {
```
mejrs, line 335: IMO `skip_first_arg` should really be a boolean-like enum, there are too many trues and falses in here.

tests/ui/invalid_pymethods.stderr
```diff
@@ -88,55 +88,89 @@ error: `signature` not allowed with `classattr`
...
     |       ^^^^^^^^^^^^
 
+error: `#[new]` does not take any arguments
+       = help: did you mean `#[new] #[pyo3(signature = ())]`?
+   --> tests/ui/invalid_pymethods.rs:118:7
+    |
+118 |     #[new(signature = ())]
+    |       ^^^
+
+error: `#[new]` does not take any arguments
```
mejrs, line 104: We should add some additional clarification, something like:
```suggestion
error: `#[new]` does not take any arguments
note = this was previously accepted and ignored
```

mejrs (commented):
Thanks!. I also like this implementation better.

tests/ui/invalid_pymethods.stderr
davidhewitt: Good idea, I like that, will add.

pyo3-macros-backend/src/method.rs
davidhewitt: Agreed, will adjust.

pyo3-macros-backend/src/method.rs
davidhewitt: A lot of the complexity from the old implementation came from the need to work out when this variant was being built out of the two attributes. By removing it and instead pattern matching this combination I think the code reads better.

pyo3-macros-backend/src/method.rs
davidhewitt: I agree but I'll deal with that in a follow-up PR.

pyo3-macros-backend/src/method.rs
davidhewitt: We don't really need the idents again, just useful in error messages. I've replaced `fn ident` with a `Display` implementation, which is hopefully a bit less weird.

davidhewitt:
Thanks for the review, force-pushed to apply all suggestions.

pyo3-macros-backend/src/method.rs
davidhewitt: Actually I pushed a second commit to address this, which moves this to be a method on the `FnType` enum.

davidhewitt:
@mejrs anything further you would like to see in this PR?

davidhewitt:
I'm going to assume this is now ok to merge; if there is more you wanted please forgive me and I'll happily follow up 🙏

PR #3488, opened 2023-10-02
support python 3.12

davidhewitt:
Switches to using 3.12.0 stable in CI!

This also adds a CHANGELOG entry, but it's mostly ceremonial given that we'd already been testing 3.12 betas in CI.


davidhewitt:
blocked on https://github.com/actions/setup-python/issues/736

PR #3493, opened 2023-10-03
release: 0.20.0

### LINKED ISSUE #3246: 0.20 release
Debian bookworm was released a week ago now, so we're able to release 0.20 whenever suits us and send the MSRV bump downstream. I don't think there's pressure to release right away, so here's a list of proposal of items I'd hope to ship in this release.

Given the new support for `__eq__` and the discussions around `__next__` / `__anext__`, I think it would be nice thematically to push a group of improvements to protocol methods in 0.20. Namely, I'm aware of possible changes to:

- [ ] `__del__` - #2484 and also a GC bug #3064 which I think both require a rework to our `tp_dealloc`
- [ ] `__next__` and `__anext__` - possible redesign as discussed in #3190
- [x] `#[new]` (i.e. `__new__`) - #2384 
- [ ] `__buffer__` and `__release_buffer__` - #3148

I've ordered these by my own opinion of their priority (which will differ for others). I'd like to aim to implement these in the same order over the next few weeks, and then prepare the release.

If anyone would like to see other items in the release, or to release sooner without these, please comment. Similarly if anyone wants to own the implementation for anything listed here, feel free to stake your claim so that we don't accidentally push over each other 😄 

Other possible items (from discussion below):
- [ ] #3061 
- [x] #3325

davidhewitt:
With Python 3.12 out the door and so many improvements ready to ship on our end, I think it's time to push 0.20 live.

There are still a few PRs which I would like to see merged before we release:
- #3490 - trivial bugfix
- #3488 
- #3484 - small UX improvement which is technically breaking
- #3455 - a bugfix which makes exceptions chain so I'd prefer ship in a feature release rather than a follow-up patch

Once we have either merged or closed those PRs I will rebase this and prepare to ship.

Closes #3246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,72 @@ To see unreleased changes, please see the [CHANGELOG on the main branch guide](h
 
 <!-- towncrier release notes start -->
 
+## [0.20.0] - 2023-10-11
+
+### Packaging
+
+- Dual-license PyO3 under either the Apache 2.0 OR the MIT license. This makes the project GPLv2 compatible. [#3108](https://github.com/PyO3/pyo3/pull/3108)
+- Update MSRV to Rust 1.56. [#3208](https://github.com/PyO3/pyo3/pull/3208)
+- Bump `indoc` dependency to 2.0 and `unindent` dependency to 0.2. [#3237](https://github.com/PyO3/pyo3/pull/3237)
+- Bump `syn` dependency to 2.0. [#3239](https://github.com/PyO3/pyo3/pull/3239)
+- Drop support for debug builds of Python 3.7. [#3387](https://github.com/PyO3/pyo3/pull/3387)
+- Bump `chrono` optional dependency to require 0.4.25 or newer. [#3427](https://github.com/PyO3/pyo3/pull/3427)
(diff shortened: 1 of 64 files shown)


mejrs:
Speaking of Python 3.12, it'd be nice to render the docs with Python 3.12 installed. We can't do much about the docs.rs docs (that is stuck on 3.10) but we should be able to have it on https://pyo3.rs

davidhewitt:
Reading the netlify documentation it sounds like the Python version available on their builder is very limited, however we could probably move to a Github action where we build the site using 3.12 and then push it with the netlify CLI.

davidhewitt:
Now that all of the PRs above have been merged (plus a couple of extra fixes) I have rebased this.

Given there's already been a bit of cooling off here and it's a good day for me to release (I'll be at EuroRust the next couple of days so will be able to find time to support any bugfixes needed), I plan to push this live this evening unless I hear reason to delay.

davidhewitt:
Ok I found a reason to wait in #3506 🙈

davidhewitt:
... and I've rebased the fix for that in now, thanks @mejrs. I'm still intending to push this evening :)

davidhewitt:
Release is live! 🚀

PR #3498, opened 2023-10-06
Pin byteorder on msrv

mejrs:
Fixes https://github.com/PyO3/pyo3/actions/runs/6435273809/job/17476207748?pr=3487

PR #3500, opened 2023-10-07
Fix `_PyFrameEvalFunction` receives an `_PyInterpreterFrame` since Python 3.11

ecarrara:
In Python 3.11 the `_PyFrameEvalFunction` receives a `_PyInterpreterFrame` opaque struct.

Python 3.11.0 (alpha 7) changelog:
> ...
> [bpo-46850](https://bugs.python.org/issue?@action=redirect&bpo=46850): Move the private _PyFrameEvalFunction type, and private _PyInterpreterState_GetEvalFrameFunc() and _PyInterpreterState_SetEvalFrameFunc() functions to the internal C API. The _PyFrameEvalFunction callback function type now uses the _PyInterpreterFrame type which is part of the internal C API. Patch by Victor Stinner.
> [bpo-46850](https://bugs.python.org/issue?@action=redirect&bpo=46850): Move the private undocumented _PyEval_EvalFrameDefault() function to the internal C API. The function now uses the _PyInterpreterFrame type which is part of the internal C API. Patch by Victor Stinner.
> ...

https://docs.python.org/3/whatsnew/changelog.html#id143

### DIFF
--- a/pyo3-ffi/src/cpython/ceval.rs
+++ b/pyo3-ffi/src/cpython/ceval.rs
@@ -5,7 +5,16 @@ use std::os::raw::c_int;
 extern "C" {
     // skipped non-limited _PyEval_CallTracing
 
+    #[cfg(not(Py_3_11))]
     pub fn _PyEval_EvalFrameDefault(arg1: *mut crate::PyFrameObject, exc: c_int) -> *mut PyObject;
+
+    #[cfg(Py_3_11)]
+    pub fn _PyEval_EvalFrameDefault(
+        tstate: *mut crate::PyThreadState,
+        frame: *mut crate::_PyInterpreterFrame,
+        exc: c_int,
+    ) -> *mut crate::PyObject;
+
     pub fn _PyEval_RequestCodeExtraIndex(func: freefunc) -> c_int;
     pub fn PyEval_SetProfile(trace_func: Option<Py_tracefunc>, arg1: *mut PyObject);
     pub fn PyEval_SetTrace(trace_func: Option<Py_tracefunc>, arg1: *mut PyObject);
--- a/pyo3-ffi/src/cpython/mod.rs
+++ b/pyo3-ffi/src/cpython/mod.rs
@@ -31,6 +31,8 @@ pub(crate) mod pystate;
 pub(crate) mod pythonrun;
 // skipped sysmodule.h
 pub(crate) mod floatobject;
+#[cfg(not(PyPy))]
+pub(crate) mod pyframe;
 pub(crate) mod tupleobject;
 pub(crate) mod unicodeobject;
 pub(crate) mod weakrefobject;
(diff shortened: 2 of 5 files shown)


mejrs (changes requested):
Thanks! Don't worry about the CI failures; you'll need to rebase once https://github.com/PyO3/pyo3/pull/3498 lands.

pyo3-ffi/src/cpython/ceval.rs
```diff
@@ -2,10 +2,22 @@ use crate::cpython::pystate::Py_tracefunc;
 use crate::object::{freefunc, PyObject};
 use std::os::raw::c_int;
 
+#[cfg(any(PyPy, Py_3_11))]
+opaque_struct!(_PyInterpreterFrame);
```
davidhewitt, line 6: It looks like (at least on CPython main) that this symbol resides in `cpython/pyframe.h`. Can you please create the corresponding `.rs` file and place this there?

pyo3-ffi/src/cpython/ceval.rs
```diff
@@ -2,10 +2,22 @@ use crate::cpython::pystate::Py_tracefunc;
...
     // skipped non-limited _PyEval_CallTracing
 
+    #[cfg(not(Py_3_11))]
     pub fn _PyEval_EvalFrameDefault(arg1: *mut crate::PyFrameObject, exc: c_int) -> *mut PyObject;
+
+    #[cfg(Py_3_11)]
+    pub fn _PyEval_EvalFrameDefault(
+        tstate: *mut crate::PyThreadState,
+        frame: *mut crate::_PyInterpreterFrame,
+        throwval: c_int,
```
davidhewitt, line 18: It looks like the header gives this argument a different name?

```suggestion
        exc: c_int,
```

davidhewitt (commented):
`main` should now be good again if you want to rebase.

pyo3-ffi/src/cpython/pyframe.rs
```diff
@@ -0,0 +1,2 @@
+#[cfg(any(PyPy, Py_3_11))]
```
mejrs, line 1: Is this available in PyPy? I couldn't find anything about it.

ecarrara: I couldn't find it either :sweat_smile: 

I've added `#[cfg(not(PyPy))]` to the `pub(crate) mod pyframe;` line on `src/cpython/mod.rs` because PyPy does not have something like "pyframe.h" in their source code.

Commit: https://github.com/PyO3/pyo3/pull/3500/commits/4931b00cbc8491f4562b76f13266ec127a017529

pyo3-ffi/src/cpython/mod.rs
```diff
@@ -58,6 +60,7 @@ pub use self::object::*;
 pub use self::objimpl::*;
 pub use self::pydebug::*;
 pub use self::pyerrors::*;
+pub use self::pyframe::*;
```
mejrs, line 63:
```suggestion
#[cfg(not(PyPy))]
pub use self::pyframe::*;
```

You need it on the re-export as well, since the module is not available on PyPy

mejrs (approved):
One last thing, can you squash the commits please? 🙂

PR #3502, opened 2023-10-08
add PyInterpreterConfig api

mejrs:


pyo3-ffi/src/cpython/pylifecycle.rs
```diff
@@ -44,5 +44,56 @@ extern "C" {
...
+#[cfg(Py_3_12)]
+#[repr(C)]
+pub struct PyInterpreterConfig {
+    use_main_obmalloc: c_int,
+    allow_fork: c_int,
+    allow_exec: c_int,
+    allow_threads: c_int,
+    allow_daemon_threads: c_int,
+    check_multi_interp_extensions: c_int,
+    gil: c_int,
```
davidhewitt, line 65: These fields need to be `pub` I think:

```suggestion
    pub use_main_obmalloc: c_int,
    pub allow_fork: c_int,
    pub allow_exec: c_int,
    pub allow_threads: c_int,
    pub allow_daemon_threads: c_int,
    pub check_multi_interp_extensions: c_int,
    pub gil: c_int,
```

davidhewitt (approved):
Thanks! I think just one tiny correction needed, afterwards please merge! :)

pyo3-ffi/src/cpython/pylifecycle.rs
mejrs: Oops

PR #3504, opened 2023-10-10
emit helpful error hint for classmethod with receiver

davidhewitt:
As per the UI test, this adds a better error message when using `#[classmethod]` on a function with a receiver.

Will skip changelog as this doesn't affect what code will compile.


davidhewitt:
I will merge this now so it can go in 0.20. I don't think it's controversial.

PR #3505, opened 2023-10-10
deprecate undocumented `#[__new__]` form of `#[new]`

davidhewitt:
This PR deprecates `#[__new__]` (which is never mentioned in documentation) as we always use it as its documented form `#[new]`.


davidhewitt:
I will merge this now so it can go in 0.20 release. I think it's not controversial.

PR #3506, opened 2023-10-11
Fix bug in default implementation of `__ne__`

### LINKED ISSUE #3417: Update docs for new __eq__ support
https://pyo3.rs/main/class/object#comparisons needs updating since we merged #2089 (to release in 0.20).

davidhewitt:
Closes #3417 

While updating the documentation for `__eq__` I fortunately stumbled across an error in the testing for `__eq__` and `__ne__` which caused us to miss the fact that `__ne__` was not defaulting to `__eq__` the same way it does for Python classes.

If anyone is available to give this a quick review (hopefully it's not a difficult review) then I'd be very grateful, as I think this blocks the release.

### DIFF
--- a/pytests/tests/test_comparisons.py
+++ b/pytests/tests/test_comparisons.py
@@ -23,10 +23,14 @@ def test_eq(ty: Type[Union[Eq, PyEq]]):
     c = ty(1)
 
     assert a == b
+    assert not (a != b)
     assert a != c
+    assert not (a == c)
 
     assert b == a
+    assert not (a != b)
     assert b != c
+    assert not (b == c)
 
     with pytest.raises(TypeError):
         assert a <= b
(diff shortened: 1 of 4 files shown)


pytests/tests/test_comparisons.py
```diff
@@ -49,17 +53,21 @@ def __eq__(self, other: Self) -> bool:
         return self.x == other.x
 
 
-@pytest.mark.parametrize("ty", (Eq, PyEq), ids=("rust", "python"))
+@pytest.mark.parametrize("ty", (EqDefaultNe, PyEqDefaultNe), ids=("rust", "python"))
 def test_eq_default_ne(ty: Type[Union[EqDefaultNe, PyEqDefaultNe]]):
```
davidhewitt, line 57: This was my original bug in the test suite.

src/impl_/pyclass.rs
```diff
@@ -805,11 +805,18 @@ slot_fragment_trait! {
...
+        // By default `__ne__` will try `__eq__` and invert the result
+        let slf: &PyAny = py.from_borrowed_ptr(slf);
+        let other: &PyAny = py.from_borrowed_ptr(other);
+        if !slf.eq(other)? {
+            Ok(ffi::Py_NewRef(ffi::Py_True()))
+        } else {
+            Ok(ffi::Py_NewRef(ffi::Py_False()))
+        }
     }
 }
```
davidhewitt, line 821: ... and this is the bugfix. :)

davidhewitt:
Skip changelog as `__ne__` is a new feature shipping in 0.20.

src/impl_/pyclass.rs
davidhewitt: (Adjusted to use a simpler implementation in a force-push)

mejrs (approved):
I just woke up but this seems good to me :)

davidhewitt:
Thanks!

PR #3507, opened 2023-10-11
Add support for `SmallVec` in conversion traits (#3440)

### LINKED ISSUE #3440: Support for `SmallVec` in conversion traits
Would the PyO3 project be interested in a PR that adds a `smallvec` feature (akin to `hashbrown` and `indexmap`) to add conversion traits for [the smallvec crate](https://docs.rs/smallvec/latest/smallvec/)?  In some of the higher-performance places I'm using PyO3 I'm moving some structures to smallvec for better cache locality, and at the moment I need to manually convert whenever one of those objects appears at the Python API boundary.

I'd be happy to contribute the PR, if it's a feature you'd like to merge.  I don't know how you feel about expanding the optional trait implementations to every crate under the sun, though!

orhun:
closes #3440

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ num-bigint = { version = "0.4", optional = true }
 num-complex = { version = ">= 0.2, < 0.5", optional = true }
 rust_decimal = { version = "1.0.0", default-features = false, optional = true }
 serde = { version = "1.0", optional = true }
-smallvec = { version = "1.11.1", optional = true }
+smallvec = { version = "1.0", optional = true }
 
 [dev-dependencies]
 assert_approx_eq = "1.1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ num-bigint = { version = "0.4", optional = true }
 num-complex = { version = ">= 0.2, < 0.5", optional = true }
 rust_decimal = { version = "1.0.0", default-features = false, optional = true }
 serde = { version = "1.0", optional = true }
+smallvec = { version = "1.11.1", optional = true }
 
 [dev-dependencies]
 assert_approx_eq = "1.1.0"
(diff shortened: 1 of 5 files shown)


davidhewitt:
Thank you! FYI other maintainers this was built with my collaboration in the EuroRust impl room, so it's best if I don't review (but can be responsible for follow up tidying).

Cargo.toml
```diff
@@ -41,6 +41,7 @@ num-bigint = { version = "0.4", optional = true }
 num-complex = { version = ">= 0.2, < 0.5", optional = true }
 rust_decimal = { version = "1.0.0", default-features = false, optional = true }
 serde = { version = "1.0", optional = true }
+smallvec = { version = "1.11.1", optional = true }
```
mejrs, line 44: Can this not use version 1.0?

mejrs (changes requested):
Thanks!

Please add an entry to the features table in lib.rs, like [here](https://docs.rs/pyo3/latest/pyo3/#optional-feature-flags)

Cargo.toml
orhun: What do you mean? `"smallvec = 1.0"`

orhun:
Added in 31acf0dc6f96ae1be9b60d4aee7359dfaad5d56e

Cargo.toml
davidhewitt: Yes, in general we try to accept as broad ranges of optional dependency versions as possible so that we don't have to force users to use specific versions.

orhun: Got it, made the change in 87cbd38229134e068df5c025a8843dbd7c51ebe1

davidhewitt (approved):
Thanks again @orhun and good to meet you at EuroRust - I think this looks good now, so I'll merge it 👍

davidhewitt:
CI failure looks caused by latest psutil release; I can try to fix separately later.

orhun:
Thanks @davidhewitt, it was nice to meet you as well, wish to see you again in events/open source world!

PR #3508, opened 2023-10-12
Create subinterpreter example

mejrs:
I wrote up most of this when I was thinking about how we can support these in pyo3. I figured I might as well share it.

### DIFF
--- a/examples/sequential/pyproject.toml
+++ b/examples/sequential/pyproject.toml
@@ -1,0 +1,20 @@
+[build-system]
+requires = ["maturin>=1,<2"]
+build-backend = "maturin"
+
+[project]
+name = "sequential"
+version = "0.1.0"
+classifiers = [
+    "License :: OSI Approved :: MIT License",
+    "Development Status :: 3 - Alpha",
+    "Intended Audience :: Developers",
+    "Programming Language :: Python",
+    "Programming Language :: Rust",
+    "Operating System :: POSIX",
+    "Operating System :: MacOS :: MacOS X",
+]
+requires-python = ">=3.12"
+
+[project.optional-dependencies]
+dev = ["pytest"]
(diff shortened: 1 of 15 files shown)


examples/sequential/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[build-system]
+requires = ["maturin>=1,<2"]
+build-backend = "maturin"
+
+[project]
+name = "string sum"
```
adamreichold, line 6: Copy pasta.

examples/sequential/requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+pytest>=3.5.0
+pip>=21.3
+maturin>=0.12,<0.13
```
adamreichold, line 3: Seems redundant to the Nox session and inconsistent with pyproject.toml? Maybe drop the requirements file outright and rely solely on the Nox session?

davidhewitt: Probably there's a broader tidy up we should consider for all of these examples, I suspect this was copypasta too.

davidhewitt (commented):
Thanks, this is cool to see! When it's laid out like this, it doesn't seem so hard right? Just need to figure out how to get our nice safe high-level wrappers to be sound under these conditions 😂

examples/sequential/requirements-dev.txt
davidhewitt: Resolved separately in #3536, I'll rebase this.

davidhewitt (approved):
I rebased this and fixed CI, looks good to me, will merge!

davidhewitt:
Fixed up for MSRV.

mejrs:
Thanks :)

PR #3514, opened 2023-10-15
Add `PyMemoryView` type

messense:
Provides a safe way to return a memoryview object to Python.


src/types/memoryview.rs
```diff
@@ -0,0 +1,19 @@
...
+/// Represents a Python `memoryview`.
+#[repr(transparent)]
+pub struct PyMemoryView(PyAny);
+
+pyobject_native_type_core!(PyMemoryView, pyobject_native_static_type_object!(ffi::PyMemoryView_Type), #checkfunction=ffi::PyMemoryView_Check);
+
+impl PyMemoryView {
+    /// Creates a new Python `memoryview` object from another Python object that
+    /// implements the buffer protocol.
+    pub fn from(src: &PyAny) -> PyResult<&PyMemoryView> {
```
adamreichold, line 13: I think this could be a `TryFrom` impl to make this available in generic context?

messense: This is modeled after [PyByteArray::from](https://docs.rs/pyo3/latest/pyo3/types/struct.PyByteArray.html#method.from).

adamreichold:
What is the relation of this to #2711 and #2792?

messense:
> What is the relation of this to #2711 and #2792?

Ah, I was searching for `memoryview` and didn't find the existing issue. 😂

messense:
IMO this PR is a very limited version of #2792 that just opens door to return memoryview object, the existing `PyBuffer` already handles the need for accessing objects that implements buffer protocol in Rust (at least for my use case).

src/types/memoryview.rs
adamreichold: Could we then add additional `TryFrom` impls for both cases?

messense: Added in https://github.com/PyO3/pyo3/pull/3514/commits/d4ed66fff0acbcf115a6511b6258cf1c6b1fc706

adamreichold (approved):

adamreichold:
> this PR is a very limited version of https://github.com/PyO3/pyo3/pull/2792

Ok, so we leave the original issue open.

davidhewitt (approved):
Thanks, looks good to me! Perhaps consider adding unit tests for the `TryFrom` implementations before merging? Otherwise please feel free to merge :)

messense:
> Perhaps consider adding unit tests for the `TryFrom` implementations before merging?

Added in https://github.com/PyO3/pyo3/pull/3514/commits/f4e64aadef251e1198a9b5884d5802bfa66b44ac, but it looks like the ergonomics of the new `TryFrom` impls are not great because they collides with `PyTryFrom::try_from` so users need to disambiguate them manually.

davidhewitt:
Ah yes, we should probably remove `PyTryFrom`. #3516

PR #3515, opened 2023-10-15
ci: drop psutil dependency

davidhewitt:
Not sure why the dependency is needed; it just broke PyPy builds, let's see what breaks if we remove it :)


davidhewitt:
Docs build... looks unrelated flakiness?

Issue #3382, opened 2023-08-11
Upcoming changes to PyO3's API in 0.21


davidhewitt:
**Update**: to avoid churn we have decided to keep `Py<T>` as-is and introduce the new API as `Bound<'py, PyAny>`. This naming can be revisited in future, but we have enough breaking changes with this 0.21 release that we do not wish to ship a wholesale rename at this point. (See https://github.com/PyO3/pyo3/issues/3674#issuecomment-1864891336)

The tracking issue of the remaining work to implement and make `Bound<'py, PyAny>` a public API is in #3684

We are also building documentation at https://pyo3.rs/main/migration#from-020-to-021

----

This is a ticket to pin to give visibility to discussion and feedback of an overhaul we're considering for PyO3's API, projected for 0.21 release later this year.

TLDR; we plan to move types like `&'py PyAny` to be stored instead in a smart pointer type `Py<'py, PyAny>`. The existing `Py<T>` type will continue to exist, but renamed, maybe to `PySend<T>` or `PyStatic<T>` (name ideas welcome).

The reason for doing this is because we believe this will offer a faster and lower memory usage API. The `&'py PyAny` types are reliant on an internal datastructure we call the "reference pool", and it's the cause of memory frustrations such as #1056. It's a thread-local datastructure, so it also has a performance overhead whenever it is accessed. From benchmarks in prototypes I think as much as 30% overhead can be removed when doing lots of operations with Python objects.

There will be some work needed to migrate to this new API, mostly due to changes in ownership semantics. I think it may be possible to expose the existing `&'py PyAny` API in a separate crate (say `pyo3_pool`, name to be bikeshed) to make migration easier.

There has been a lot of thought put into this API design over a large span of time. We want to preserve as many existing semantics as possible as well as provide an as simple a migration path as possible from current PyO3 code. We want PyO3 to be as fast as possible while also being accessible to Python users coming to Rust for the first time.

For the discussion of the latest prototype of this new API, see https://github.com/PyO3/pyo3/pull/3361

All feedback and questions welcome.

adamreichold:
> but renamed, maybe to PySend<T> or PyStatic<T> (name ideas welcome).

I was somewhat enamoured with `PyUngil<T>` matching the `Ungil` trait.

davidhewitt:
I could live with that, even if we have to rename it again in a nogil future.

adamreichold:
[...]

PR #3517, opened 2023-10-15
remove comparison to rust-cpython

davidhewitt:
I noticed recently that `rust-cpython` now displays a deprecation message instructing to use PyO3, and that it's not planning to add support for Python 3.12.

As such, I think there's not much point in keeping around a comparison page to rust-cpython. This page is also beginning to show its age, and would need a rewrite after #3382 anyway.


davidhewitt:
What might be a good replacement is a document with comparisons to other approaches like uniFFI and cffi, and maybe even Cython and pybind11.

messense (approved):

PR #3518, opened 2023-10-15
ci: install prebuilt cargo-careful

davidhewitt:
- use `taiki-e/install-action@cargo-careful` to avoid compiling `cargo-careful`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -287,7 +287,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@nightly
         with:
           components: rust-src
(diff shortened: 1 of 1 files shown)


messense (commented):
According to https://github.com/dtolnay/rust-toolchain#inputs

> When passing an explicit toolchain as an input instead of `@rev`, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action.

I think `master` is the correct usage.

davidhewitt:
Right you are, my bad. Rolled this back to just do the `cargo-careful` change.

messense (approved):

PR #3521, opened 2023-10-16
feat: Support for mappingproxy objects

michaelhly:
Adds support for mappingproxy objects. Continuation of #2435. Resolves https://github.com/PyO3/pyo3/issues/2108.


src/types/mappingproxy.rs
```diff
@@ -0,0 +1,597 @@
...
+    /// This is equivalent to the Python expression `self.copy()`.
+    pub fn copy(&self) -> PyResult<&PyMapping> {
+        self.call_method0("copy")
+            .and_then(|object| object.downcast::<PyMapping>().map_err(PyErr::from))
+    }
+
+    /// Checks if the mappingproxy is empty, i.e. `len(self) == 0`.
+    pub fn is_empty(&self) -> bool {
+        self.len().unwrap_or_default() == 0
+    }
```
davidhewitt, line 39: I think we can't justify this default for users and should propagate the error:

```suggestion
    pub fn is_empty(&self) -> PyResult<bool> {
        self.len().map(|len| len == 0)
    }
```

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,597 @@
...
+    ///    
+    /// Returns `None` if the item is not present, or if an error occurs.
+    ///
+    /// To get a `KeyError` for non-existing keys, use `PyAny::get_item`.
+    pub fn get_item<K>(&self, key: K) -> Option<&PyAny>
+    where
+        K: ToPyObject,
+    {
+        PyAny::get_item(self, key).ok()
+    }
```
davidhewitt, line 51: Given the recent changes around `PyDict::get_item`, similarly we should not suppress the error here. As there's no clear advantage of this over `PyAny::get_item` I guess we just remove this?

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,597 @@
...
+    /// Returns a list of mappingproxy keys.
+    ///
+    /// This is equivalent to the Python expression `list(mappingproxy.keys())`.
+    pub fn keys(&self) -> &PyList {
+        unsafe {
+            PySequence::try_from_unchecked(self.call_method0("keys").unwrap())
+                .to_list()
+                .unwrap()
+        }
+    }
```
davidhewitt, line 62: The unwraps are not ok here, let's have `PyResult<&PyList>` as the return type please!

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,597 @@
...
+
+    /// Returns a list of mappingproxy values.
+    ///
+    /// This is equivalent to the Python expression `list(mappingproxy.values())`.
+    pub fn values(&self) -> &PyList {
```
davidhewitt, line 67: Same here `PyResult<&PyList>`

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,597 @@
...
+            PySequence::try_from_unchecked(self.call_method0("values").unwrap())
+                .to_list()
+                .unwrap()
+        }
+    }
+
+    /// Returns a list of mappingproxy items.
+    ///
+    /// This is equivalent to the Python expression `list(mappingproxy.items())`.
+    pub fn items(&self) -> &PyList {
```
davidhewitt, line 78: Same here `PyResult<&PyList>`

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,597 @@
...
+impl<'py, T, I> IntoPyMappingProxy<'py> for I
+where
+    T: PyDictItem,
+    I: IntoIterator<Item = T>,
+{
+    fn into_py_mappingproxy(self, py: Python<'py>) -> PyResult<&'py PyMappingProxy> {
+        let dict = self.into_py_dict(py);
+        PyMappingProxy::new(py, dict.as_mapping())
+    }
+}
```
davidhewitt, line 144: This is a cute trait, but I'm not really convinced this is necessary given the user can just do `PyMappingProxy::new(py, v.into_py_dict(py).as_mapping())`. Have you got a use case in mind?

src/types/mappingproxy.rs
```diff
@@ -0,0 +1,597 @@
...
+
+pub struct PyMappingProxyIterator<'py> {
+    iterator: &'py PyIterator,
+    mappingproxy: &'py PyMappingProxy,
+}
+
+impl<'py> Iterator for PyMappingProxyIterator<'py> {
+    type Item = (&'py PyAny, &'py PyAny);
```
davidhewitt, line 104: I guess iteration here should probably be fallible? Overall I see why you added this and I think it makes sense to iterate pairs given the `PyDictIterator` works the same way.

```suggestion
    type Item = PyResult<(&'py PyAny, &'py PyAny)>;
```

davidhewitt (changes requested):
Thank you for this implementation, and sorry it's taken me a few days to reply!

Overall looks very good, but I would like to see some changes to propagate errors rather than introduce panics. The only case where I'm not 100% sure is in the iterator, but there I still think it's better to err on the side of fallibility. Would be interesting to see what other maintainers think.

Also, would you be willing to help us prepare this new API in a forwards-compatible way for #3382? I have started some other PRs like #3535 and #3531 which hopefully show you what I'm doing at the moment, we could add `PyMappingProxyMethods` here in the same way. It would be really interesting to get your opinion as a user how this new API feels, though I confess that it's still early days and unrefined so you might be in for a bumpy ride 🙏

davidhewitt:
Continued in #4644

PR #3522, opened 2023-10-16
Add `PyByteArray::from_ptr` unsafe function

messense:
This mimics [`PyBytes::from_ptr`](https://docs.rs/pyo3/latest/pyo3/types/struct.PyBytes.html#method.from_ptr).


src/types/bytearray.rs
```diff
@@ -71,6 +71,24 @@ impl PyByteArray {
...
+    /// Panics if out of memory.
+    ///
+    /// # Safety
+    ///
+    /// This function dereferences the raw pointer `ptr` as the
+    /// leading pointer of a slice of length `len`. [As with
+    /// `std::slice::from_raw_parts`, this is
+    /// unsafe](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety).
+    /// If `ptr` is null, the contents of the bytearray object are uninitialized.
+    pub unsafe fn from_ptr(py: Python<'_>, ptr: *const u8, len: usize) -> &PyByteArray {
```
adamreichold, line 85: I understand why the C API needs to be unsafe, but why don't we wrap this like

```rust
pub fn from_slice(py: Python<'_>, slice: &[u8]) -> &PyByteArray { ... }
```

if the pointer is referenced that way in any case?

As it is written it seems to offer very little that the FFI entry point does not.

adamreichold: (Or `slice: Option<&[u8]>` if we want to support the nullable case.)

mejrs: This would also better communicate what the function does. Does it copy the contents of the slice or does it just reference the pointer? I'm guessing the former but I'm not sure.

messense: https://github.com/python/cpython/blob/37bd8726b87c1c41f58cef031f5b6d5f900d265b/Objects/bytearrayobject.c#L141

It calls `memcpy` with the pointer and size.

messense: > As it is written it seems to offer very little that the FFI entry point does not.

Yes, it's just more convenient.

> ```rust
> pub fn from_slice(py: Python<'_>, slice: &[u8]) -> &PyByteArray { ... }
> ```

I'm fine with also adding this, the nullable case is unsafe so I think it's better for `from_slice` to just take `&[u8]` and use `from_ptr` for the nullable case.

davidhewitt: Is it problematic to have a way to read uninitialised memory from the nullable case? It seems a little scary to me, but I think because of the FFI boundary in practice this won't trigger UB and instead just gets treatedas arbitrary bytes slice.

What does it look like to print the uninitialised bytearray in Python? Might be a bit too easy to accidentally leak sensitive data if the buffer just gets printed.

We could have a `PyByteArray::zeroed` safe constructor, maybe?

src/types/bytes.rs
```diff
@@ -78,6 +78,7 @@ impl PyBytes {
     /// leading pointer of a slice of length `len`. [As with
     /// `std::slice::from_raw_parts`, this is
     /// unsafe](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety).
+    /// If `ptr` is null, the contents of the bytes object are uninitialized.
```
davidhewitt, line 81: Hmm. Bytes objects aren't really supposed to be mutated, so having uninitialised data here seems not ideal. Should we deprecate this and go for a slice here too?

PR #3526, opened 2023-10-18
ci: include beta as an allowed-to-fail job

davidhewitt:
I heard at EuroRust that rust's distribution team would love to see more testers of beta before the stable release. I figured it wouldn't be that hard for us to add beta as an additional allowed-to-fail job.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,16 +16,13 @@ on:
       rust-target:
         required: true
         type: string
-      msrv:
-        required: false
-        type: string
       extra-features:
         required: true
         type: string
 
 jobs:
   build:
-    continue-on-error: ${{ endsWith(inputs.python-version, '-dev') || contains(fromJSON('["3.7", "pypy3.7"]'), inputs.python-version) }}
+    continue-on-error: ${{ endsWith(inputs.python-version, '-dev') || contains(fromJSON('["3.7", "pypy3.7"]'), inputs.python-version) || inputs.rust == 'beta' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -49,18 +49,18 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
         with:
-          key: cargo-${{ inputs.python-architecture }}-${{ inputs.os }}-${{ inputs.msrv }}
+          key: cargo-${{ inputs.python-architecture }}
```
davidhewitt, line 49: Apparently the cache automatically includes host and rustc release, so I removed these from the key.

https://github.com/Swatinem/rust-cache/tree/v2/#cache-details

.github/workflows/build.yml
```diff
@@ -49,18 +49,18 @@ jobs:
...
         continue-on-error: true
 
       - if: inputs.os == 'ubuntu-latest'
         name: Prepare LD_LIBRARY_PATH (Ubuntu only)
         run: echo LD_LIBRARY_PATH=${pythonLocation}/lib >> $GITHUB_ENV
 
-      - if: inputs.msrv == 'MSRV'
+      - if: inputs.rust == '1.56.0'
```
davidhewitt, line 56: I removed the "msrv" input as a bit of simplification. It meant this one line gets coupled to our msrv, but I think that's trivial to replace when future msrv bumps come along.

messense (approved):

PR #3534, opened 2023-10-20
ci: use install-action to download cargo-codspeed binary

davidhewitt:
Should cut the benchmarks job time by ~5 minutes (as well as reduce the build cache size).


davidhewitt:
Looks like the prebuilt download failed, will retry after https://github.com/CodSpeedHQ/codspeed-rust/pull/24

davidhewitt:
Still nope; https://github.com/CodSpeedHQ/codspeed-rust/issues/27

Issue #1809, opened 2021-08-18
RFC: use a wrapper for C-API return values to avoid missing DECREFs


birkenfeld:
Since we're using Rust, couldn't we make the raw ffi API functions a little safer by having a transparent wrapper type for new reference return values of type `*mut PyObject` that can't be dropped implicitly, but has methods to be:

* converted to a PyO3 owned object (Py<X> or &X)
* dropped with a `(x)decref` method
* or explicitly unwrapped with a "trust me" method for special cases

Admittedly, most return values are already directly converted to PyO3 objects, so the gain is small, especially since the work to convert everything now is tedious, and it should to be done all in one release since it's a breaking change.

(I don't know off the top of my head how to make a type not-implicitly-droppable, but there must be a way - in the worst case there will be a dtolnay crate...)

davidhewitt:
This is an interesting idea. TBH I quite like that our ffi declaration is as close to a dumb wrapper around the C interface as possible, however I agree it can easily lead to errors in reference counting.

I think `#[must_use] #[repr(transparent)]` type would get us most of the way.

One could even imagine going further and encode any possible error condition in the return value (e.g. integer types, is the return value -1, or object types typically have `NULL` as an error return), or whether the return value is guaranteed to be a certain subtype of `PyObject` (e.g. `PyList_New` presumably always returns a `PyList` if successful).

PR #3535, opened 2023-10-20
implement `PySequenceMethods`

davidhewitt:
For #3382 

This moves the API for `PySequence` to be implemented by a pool-free implementation. As per the rest of the PRs, the new trait `PySequenceMethods` is restricted to `pub(crate)` for now while we're iterating on this.

### DIFF
--- a/src/instance/ffi_call.rs
+++ b/src/instance/ffi_call.rs
@@ -1,0 +1,59 @@
+use std::os::raw::c_int;
+
+use crate::err::{error_on_minusone, SignedInteger};
+use crate::types::any::PyAnyMethods;
+use crate::{ffi, Py2, PyErr, PyResult, Python};
+
+/// Internal helper to convert raw ffi call results such as pointers
+/// or integers into safe wrappers.
+///
+/// `unsafe` to implement because it is highly likely this trait is
+/// passed a pointer, and is free to do interpret it as it sees fit.
+pub(crate) unsafe trait FromFfiCallResult<'py, RawResult>: Sized {
+    fn from_ffi_call_result(py: Python<'py>, raw: RawResult) -> PyResult<Self>;
+}
+
+/// For Py2<T>, perform an unchecked downcast to the target type T.
+unsafe impl<'py, T> FromFfiCallResult<'py, *mut ffi::PyObject> for Py2<'py, T> {
+    fn from_ffi_call_result(py: Python<'py>, raw: *mut ffi::PyObject) -> PyResult<Self> {
+        unsafe { Py2::from_owned_ptr_or_err(py, raw).map(|any| any.downcast_into_unchecked()) }
+    }
+}
+
+unsafe impl<'py, T> FromFfiCallResult<'py, T> for ()
+where
+    T: SignedInteger,
+{
+    fn from_ffi_call_result(py: Python<'py>, raw: T) -> PyResult<Self> {
+        error_on_minusone(py, raw)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


src/types/sequence.rs
```diff
@@ -265,6 +208,312 @@ impl PySequence {
...
+                self.py(),
+                ffi::PySequence_InPlaceRepeat(self.as_ptr(), get_ssize_index(count)),
+            )
+            .map(|py2| py2.downcast_into_unchecked())
+        }
+    }
+
+    #[inline]
+    fn get_item(&self, index: usize) -> PyResult<Py2<'py, PyAny>> {
+        unsafe {
```
adamreichold, line 387: Would it make sense to abstract out the repeating

```rust
Py2::from_owned_ptr_or_err(
    self.py(),
    f(self.as_ptr()),
)
.map(|py2| py2.downcast_into_unchecked())
```

using a for example a closure-based API with `f: FnOnce(*mut PyObject) -> R`?

src/types/sequence.rs
```diff
@@ -3,10 +3,11 @@ use crate::exceptions::PyTypeError;
 #[cfg(feature = "experimental-inspect")]
 use crate::inspect::types::TypeInfo;
 use crate::internal_tricks::get_ssize_index;
+use crate::prelude::*;
```
adamreichold, line 6: I am not keen on using our own prelude inside our fundamental implementations. I think it is too easy to have circular dependencies this way.

src/types/sequence.rs
davidhewitt: Yes, I think there could definitely be a `pub(crate)` helper to do this. I'll have a play 👍

davidhewitt: I've pushed a commit which adds `Py2::do_ffi_call` which has a generic return type with a bit of trait magic to map raw ffi return values into various safe wrappers. What do you think?

src/instance/ffi_call.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::types::any::PyAnyMethods;
+use crate::{ffi, Py2, PyErr, PyResult, Python};
+
+/// Internal helper to convert raw ffi call results such as pointers
+/// or integers into safe wrappers.
+///
+/// `unsafe` to implement because it is highly likely this trait is
+/// passed a pointer, and is free to do interpret it as it sees fit.
+pub(crate) unsafe trait FromFfiCallResult<'py, RawResult>: Sized {
+    fn from_ffi_call_result(py: Python<'py>, raw: RawResult) -> PyResult<Self>;
```
davidhewitt, line 13: Staring at this again I wonder if it should have been a safe trait with the method being `unsafe fn from_ffi_call_result`. Both are a bit awkward because only the implementation for `*mut ffi::PyObject` actually needs to be `unsafe`. Anyway, this is only an internal API so it doesn't need to be perfect...

birkenfeld: Reading the comment above, why does it have to be unsafe at all? "May get a pointer and do unsafe stuff with it" doesn't sound like a strong argument :)

davidhewitt: I'm generally of the understanding that the pointer alone can't communicate that's it's safe to read.

Probably the right approach is to combine this with some wrapper along the lines of #1809 so that the trait doesn't need to be unsafe at all. I'll keep playing.

davidhewitt:
Closing in favour of #3572

Issue #3052, opened 2023-03-16
Support for GraalPy


timfel:
Hi,

We would like PyO3 to work on GraalPy. Would contributions to that end be welcome?

[GraalPy](https://www.graalvm.org/python/) is getting more compatible (we can run large packages like Pytorch, scikit-learn, Flask, ...). Like PyPy, we cannot be ABI compatible to CPython, but instead extensions need to be compiled against GraalPy's implementation of the CPython API. Currently, this means we are blocked from running extensions such as [cryptography](https://github.com/pyca/cryptography), because the build will bail when the pyo3-build-config script declares `error: unknown interpreter: GraalVM`.

We are happy to contribute the code to make this work. Are there concerns you have or issues you foresee that we should be aware of?

messense:
I'd be happy to accept contributions that add GraalPy support provided that it's well tested on our CI.

kngwyu:
The project looks interesting, and I also think that supporting it is fine if it doesn't need much maintenance effort. How different is the ABI?

konstin:
that would be really cool! do you by chance have any docs explaining how you deal with differences e.g. between graalpython's GC and the reference counting exposed in the C API as [pypy explains here](https://www.pypy.org/posts/2018/09/inside-cpyext-why-emulating-cpython-c-8083064623681286567.html)? Could you also comment on the state of hpy and graalpython, since my understanding was that hpy is supposed to become the dedicated new, easier to implement C API at some point.

davidhewitt:
I would be happy to see GraalPy support contributed. If you support HPy and would be willing to help us bind via that, would be the most reusable for sure. If you want to bind directly to your API that's fine too, we might need to think a little to make sure `pyo3-ffi` doesn't get too hard to navigate.

timfel:
> do you by chance have any docs explaining how you deal with differences

We don't have docs like that. We have different C API backends, for our "native" backend our implementation strategy is basically identical to PyPy's. For our "mnaged" backend, we interpret LLVM bitcode instead of running anything natively, so we can just ignore refcounts because all allocations even for native extensions are handled by the JVM.

> comment on the state of hpy and graalpython

[...]

PR #3536, opened 2023-10-20
examples: remove requirements-dev.txt files

davidhewitt:
While reviewing #3508 and #3052 we've noticed that we have a bit of duplication of maturin requirements as well as some inconsistency in how we install / pin maturin versions in the examples.

This PR attempts to fix that by removing the `requirements-dev.txt` files, and moving any "dev" requirements as optional dependencies in the `pyproject.toml` files. Finally, the `noxfile`s are changed to use `--config-settings` to pass arguments to maturin in a PEP517 install rather than by calling `maturin develop`, meaning those calls _also_ go through the maturin version configured in the `pyproject.toml`.


examples/word-count/.template/setup.cfg
davidhewitt, line 1: Dead file no longer needed.

examples/word-count/.template/pre-script.rhai
```diff
@@ -1,3 +1,4 @@
 variable::set("PYO3_VERSION", "0.20.0");
 file::rename(".template/Cargo.toml", "Cargo.toml");
+file::rename(".template/pyproject.toml", "pyproject.toml");
```
davidhewitt, line 3: This was missing from the template, thought I would fix at the same time while touching up these examples.

examples/decorator/noxfile.py
```diff
@@ -3,7 +3,5 @@
 
 @nox.session
 def python(session):
-    session.install("-rrequirements-dev.txt")
-    session.install("maturin")
-    session.run_always("maturin", "develop")
+    session.install(".[dev]", "--config-settings=build-args=--profile=dev")
```
adamreichold, line 6: This thrice nested argument passing looks quite unreadable. While putting everything into `pyproject.toml` is nice, is there no way around this incantation?

examples/word-count/.template/pyproject.toml
```diff
@@ -1,9 +1,13 @@
 [build-system]
-requires = ["setuptools>=41.0.0", "wheel", "setuptools_rust>=1.0.0"]
+requires = ["maturin>=1,<2"]
```
adamreichold, line 2: This does bring our `setuptools_rust` based examples down to one (the starter), right? Do we want this because Maturin is the preferred approach? (It would make sense to me, I would just like it being called out as a deliberate decision.)

davidhewitt: Actually this was already building with maturin in the noxfile, so I just fixed the inconsistency. But yes I think in general maturin is easier so it's been nicer to use it for examples.

examples/decorator/noxfile.py
davidhewitt: @messense any suggestions? I think we deprecated `MATURIN_PEP517_ARGS` so unless pip has an environment variable for this I think command line is the only way.

messense: `--config-settings` is the recommended way to pass arguments, but I'm thinking about un-deprecate `MATURIN_PEP517_ARGS` because Poetry doesn't seem to support `--config-settings`.

davidhewitt: In that case @adamreichold would you prefer I switch to the env var here?

@messense see also https://github.com/PyO3/setuptools-rust/issues/327#issuecomment-1525560005 which was another reason to want an environment variable.

adamreichold: > In that case @adamreichold would you prefer I switch to the env var here?

Yes, I think it would be preferable if it removes at least one level of `=` nesting.

davidhewitt: Done 👍

davidhewitt:
@adamreichold I'm going to assume this is good to merge now that I switched to use the `MATURIN_PEP517_ARGS` environment variable. If not, please ping and I can push follow ups / revert!

adamreichold (approved):
Sorry, just lost this from my to-do list.

PR #3539, opened 2023-10-22
add coverage for `emit_pyo3_cfgs`

davidhewitt:
While experimenting with https://github.com/PyO3/pyo3/pull/3247#discussion_r1367511468 I improved the test coverage for `pyo3-build-config` slightly. Might as well push this!

### DIFF
--- a/pyo3-build-config/src/errors.rs
+++ b/pyo3-build-config/src/errors.rs
@@ -12,15 +12,21 @@ macro_rules! ensure {
     ($condition:expr, $($args: tt)+) => { if !($condition) { bail!($($args)+) } };
 }
 
-/// Show warning. If needed, please extend this macro to support arguments.
+/// Show warning.
 #[macro_export]
 #[doc(hidden)]
 macro_rules! warn {
-    ($msg: literal) => {
-        println!(concat!("cargo:warning=", $msg))
+    ($($args: tt)+) => {
+        println!("{}", $crate::format_warn!($($args)+))
     };
-    ($fmt: expr, $($args: tt)+) => {
-        println!("cargo:warning={}", format_args!($fmt, $($args)+))
+}
+
+/// Format warning into string.
+#[macro_export]
+#[doc(hidden)]
+macro_rules! format_warn {
+    ($($args: tt)+) => {
+        format!("cargo:warning={}", format_args!($($args)+))
     };
 }
 
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -27,7 +27,7 @@ use target_lexicon::{Environment, OperatingSystem};
 use crate::{
     bail, ensure,
     errors::{Context, Error, Result},
-    warn,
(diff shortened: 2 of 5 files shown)


pyo3-build-config/src/errors.rs
```diff
@@ -12,15 +12,21 @@ macro_rules! ensure {
...
 
-/// Show warning. If needed, please extend this macro to support arguments.
+/// Show warning.
 #[macro_export]
 #[doc(hidden)]
 macro_rules! warn {
-    ($msg: literal) => {
-        println!(concat!("cargo:warning=", $msg))
+    ($($args: tt)+) => {
+        $crate::write_warn!(std::io::stdout(), "cargo:warning={}", format_args!($($args)+)).unwrap()
```
adamreichold, line 20: Is the additional layer of `format_args!` necessary? Couldn't they just be forwarded as they are?

pyo3-build-config/src/impl_.rs
```diff
@@ -154,30 +154,31 @@ pub struct InterpreterConfig {
 
 impl InterpreterConfig {
     #[doc(hidden)]
-    pub fn emit_pyo3_cfgs(&self) {
+    pub fn emit_pyo3_cfgs(&self, mut writer: impl Write) {
```
adamreichold, line 157: Clippy appears to be unhappy about passing `stdot` manually. 

I wonder we should try to simplify this more aggressively and e.g. have this method return `String` or even `Vec<String>` which the outside code then has to be print.

pyo3-build-config/src/errors.rs
davidhewitt: This was an honest oversight :)

pyo3-build-config/src/impl_.rs
davidhewitt: Done and force-pushed.

pyo3-build-config/src/errors.rs
```diff
@@ -12,15 +12,21 @@ macro_rules! ensure {
...
-    ($fmt: expr, $($args: tt)+) => {
-        println!("cargo:warning={}", format_args!($fmt, $($args)+))
+}
+
+/// Show warning on a specific writer.
+#[macro_export]
+#[doc(hidden)]
+macro_rules! format_warn {
+    ($($args: tt)+) => {
+        format!("cargo:warning={}", format_args!($($args)+))
```
adamreichold, line 29: The doc comment does not seem to apply any more as this yields a `String`. Also isn't the `format_args!` layer unnecessary again?

adamreichold: Ah sorry, the `format_args!` is necessary because this includes another format string at the call site. I force pushed a fix to the comment.

adamreichold (approved):

davidhewitt:
Thanks for tidying this up 🙏

davidhewitt:
Looks like I broke clippy, I'll fixup in a moment.

Issue #2582, opened 2022-08-22
wasm32-emscripten CI additions


davidhewitt:
Future additions for the emscripten CI job:

---

Well there's a couple more things I'm interested in:
1. I'd like to run the tests that mix Rust and C/C++. This is quite common in real packages.
2. I'm working on a way to have Pyodide virtual environments https://github.com/pyodide/pyodide/pull/2976 this will lead to rather more graceful testing.

_Originally posted by @hoodmane in https://github.com/PyO3/pyo3/issues/2412#issuecomment-1221315962_

davidhewitt:
@hoodmane given there's been some time I wonder if you can share some thoughts on how relevant you think these ideas still are, and maybe lay out what resources I should look at to make the changes you had in mind? (Or if there's other things you'd recommend the PyO3 CI start to cover?)

davidhewitt:
I noticed that in Python 3.12 there is a tool to build wasm in tree. I wonder if we want to use this. [Tools/wasm/wasm_build.py](https://github.com/python/cpython/tree/3.12/Tools/wasm/wasm_build.py)

Also we potentially want to test pyodide, maybe as a replacement for building in-tree cpython? (Or in addition to.)

hoodmane:
> given there's been some time I wonder if you can share some thoughts on how relevant you think these ideas still are

Well they are still relevant but the good news is that in practice it seems to work fine. My highest priority for Rust is to fix wasm-eh:
https://github.com/rust-lang/rust/issues/112195
https://github.com/emscripten-core/emscripten/issues/19612
Rust packages are the only things that can't build with wasm exception handling, but it seems to require fixing the `try` rustc intrinsic, so it won't work with current versions of rust.

Testing against Pyodide is pretty reasonable. The build settings that `wasm_build.py` produce are closely modeled on Pyodide's build settings so it should be similar. But IIRC the pyo3 tests want to statically link libpython into the test binaries so it would require some extra work to use Pyodide there.

davidhewitt:
Updated the OP to add a note for async tests after #3540 merges

PR #3540, opened 2023-10-23
feat: support `async fn` in macros with coroutine implementation

### LINKED ISSUE #1632: `async fn` tracking issue

wyfo:
~~Fixes #1632~~

Here is a draft for the implementation of asyncio coroutine and async fn support. It lacks exhaustive documentation and test, the goal is at first to review the implementation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,9 @@ unindent = { version = "0.2.1", optional = true }
 # support crate for multiple-pymethods feature
 inventory = { version = "0.3.0", optional = true }
 
+# coroutine implementation
+futures-util = "0.3"
+
 # crate integrations that can be added using the eponymous features
 anyhow = { version = "1.0", optional = true }
 chrono = { version = "0.4.25", default-features = false, optional = true }
@@ -54,6 +57,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.61"
 rayon = "1.6.1"
 widestring = "0.5.1"
+futures = "0.3.28"
 
 [build-dependencies]
 pyo3-build-config = { path = "pyo3-build-config", version = "0.21.0-dev", features = ["resolve-config"] }
--- a/guide/src/SUMMARY.md
+++ b/guide/src/SUMMARY.md
@@ -19,6 +19,7 @@
   - [Conversion traits](conversions/traits.md)]
 - [Python exceptions](exception.md)
 - [Calling Python from Rust](python_from_rust.md)
+- [Using `async` and `await`](async-await.md)
 - [GIL, mutability and object types](types.md)
 - [Parallelism](parallelism.md)
 - [Debugging](debugging.md)
--- a/guide/src/async-await.md
+++ b/guide/src/async-await.md
@@ -1,0 +1,78 @@
+# Using `async` and `await`
+
+*This feature is still in active development. See [the related issue](https://github.com/PyO3/pyo3/issues/1632).*
+
+`#[pyfunction]` and `#[pymethods]` attributes also support `async fn`.
+
+```rust
+# #![allow(dead_code)]
+use std::{thread, time::Duration};
+use futures::channel::oneshot;
+use pyo3::prelude::*;
+
+#[pyfunction]
+async fn sleep(seconds: f64, result: Option<PyObject>) -> Option<PyObject> {
+    let (tx, rx) = oneshot::channel();
+    thread::spawn(move || {
+        thread::sleep(Duration::from_secs_f64(seconds));
+        tx.send(()).unwrap();
+    });
+    rx.await.unwrap();
+    result
+}
+```
+
+*Python awaitables instantiated with this method can only be awaited in *asyncio* context. Other Python async runtime may be supported in the future.*
+
+## `Send + 'static` constraint
+
+Resulting future of an `async fn` decorated by `#[pyfunction]` must be `Send + 'static` to be embedded in a Python object.
+
+As a consequence, `async fn` parameters and return types must also be `Send + 'static`, so it is not possible to have a signature like `async fn does_not_compile(arg: &PyAny, py: Python<'_>) -> &PyAny`.
+
+It also means that methods cannot use `&self`/`&mut self`, *but this restriction should be dropped in the future.*
+
+
+## Implicit GIL holding
+
+Even if it is not possible to pass a `py: Python<'_>` parameter to `async fn`, the GIL is still held during the execution of the future – it's also the case for regular `fn` without `Python<'_>`/`&PyAny` parameter, yet the GIL is held.
+
+It is still possible to get a `Python` marker using [`Python::with_gil`]({{#PYO3_DOCS_URL}}/pyo3/struct.Python.html#method.with_gil); because `with_gil` is reentrant and optimized, the cost will be negligible.
+
+## Release the GIL across `.await`
+
+There is currently no simple way to release the GIL when awaiting a future, *but solutions are currently in development*.
+
+Here is the advised workaround for now:
+
+```rust,ignore
+use std::{future::Future, pin::{Pin, pin}, task::{Context, Poll}};
+use pyo3::prelude::*;
+
+struct AllowThreads<F>(F);
+
+impl<F> Future for AllowThreads<F>
+where
+    F: Future + Unpin + Send,
+    F::Output: Send,
+{
+    type Output = F::Output;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let waker = cx.waker();
+        Python::with_gil(|gil| {
+            gil.allow_threads(|| pin!(&mut self.0).poll(&mut Context::from_waker(waker)))
+        })
+    }
+}
+```
+
+## Cancellation
+
+*To be implemented*
+
+## The `Coroutine` type
+
+To make a Rust future awaitable in Python, PyO3 defines a [`Coroutine`]({{#PYO3_DOCS_URL}}/pyo3/coroutine/struct.Coroutine.html) type, which implements the Python [coroutine protocol](https://docs.python.org/3/library/collections.abc.html#collections.abc.Coroutine). Each `coroutine.send` call is translated to `Future::poll` call, while `coroutine.throw` call reraise the exception *(this behavior will be configurable with cancellation support)*.
+
+*The type does not yet have a public constructor until the design is finalized.*
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -228,6 +228,7 @@ pub struct FnSpec<'a> {
     pub output: syn::Type,
     pub convention: CallingConvention,
     pub text_signature: Option<TextSignatureAttribute>,
+    pub asyncness: Option<syn::Token![async]>,
     pub unsafety: Option<syn::Token![unsafe]>,
     pub deprecations: Deprecations,
 }
@@ -317,6 +318,7 @@ impl<'a> FnSpec<'a> {
             signature,
             output: ty,
             text_signature,
+            asyncness: sig.asyncness,
             unsafety: sig.unsafety,
             deprecations,
         })
@@ -445,7 +447,11 @@ impl<'a> FnSpec<'a> {
         let func_name = &self.name;
 
         let rust_call = |args: Vec<TokenStream>| {
-            quotes::map_result_into_ptr(quotes::ok_wrap(quote! { function(#self_arg #(#args),*) }))
+            let mut call = quote! { function(#self_arg #(#args),*) };
+            if self.asyncness.is_some() {
+                call = quote! { _pyo3::impl_::coroutine::wrap_future(#call) };
+            }
+            quotes::map_result_into_ptr(quotes::ok_wrap(call))
         };
 
         let rust_name = if let Some(cls) = cls {
--- a/pyo3-macros-backend/src/utils.rs
+++ b/pyo3-macros-backend/src/utils.rs
@@ -1,6 +1,6 @@
 use proc_macro2::{Span, TokenStream};
 use quote::ToTokens;
-use syn::{punctuated::Punctuated, spanned::Spanned, Token};
+use syn::{punctuated::Punctuated, Token};
 
 use crate::attributes::{CrateAttribute, RenamingRule};
 
@@ -135,17 +135,6 @@ impl quote::ToTokens for PythonDoc {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         self.0.to_tokens(tokens)
     }
-}
-
-pub fn ensure_not_async_fn(sig: &syn::Signature) -> syn::Result<()> {
-    if let Some(asyncness) = &sig.asyncness {
-        bail_spanned!(
-            asyncness.span() => "`async fn` is not yet supported for Python functions.\n\n\
-            Additional crates such as `pyo3-asyncio` can be used to integrate async Rust and \
-            Python. For more information, see https://github.com/PyO3/pyo3/issues/1632"
-        );
-    };
-    Ok(())
 }
 
 pub fn unwrap_ty_group(mut ty: &syn::Type) -> &syn::Type {
--- a/src/coroutine.rs
+++ b/src/coroutine.rs
@@ -1,0 +1,137 @@
+//! Python coroutine implementation, used notably when wrapping `async fn`
+//! with `#[pyfunction]`/`#[pymethods]`.
+use std::{
+    any::Any,
+    future::Future,
+    panic,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+use futures_util::FutureExt;
+use pyo3_macros::{pyclass, pymethods};
+
+use crate::{
+    coroutine::waker::AsyncioWaker,
+    exceptions::{PyRuntimeError, PyStopIteration},
+    panic::PanicException,
+    pyclass::IterNextOutput,
+    types::PyIterator,
+    IntoPy, Py, PyAny, PyErr, PyObject, PyResult, Python,
+};
+
+mod waker;
+
+const COROUTINE_REUSED_ERROR: &str = "cannot reuse already awaited coroutine";
+
+type FutureOutput = Result<PyResult<PyObject>, Box<dyn Any + Send>>;
+
+/// Python coroutine wrapping a [`Future`].
+#[pyclass(crate = "crate")]
+pub struct Coroutine {
+    future: Option<Pin<Box<dyn Future<Output = FutureOutput> + Send>>>,
+    waker: Option<Arc<AsyncioWaker>>,
+}
+
+impl Coroutine {
+    ///  Wrap a future into a Python coroutine.
+    ///
+    /// Coroutine `send` polls the wrapped future, ignoring the value passed
+    /// (should always be `None` anyway).
+    ///
+    /// `Coroutine `throw` drop the wrapped future and reraise the exception passed
+    pub(crate) fn from_future<F, T, E>(future: F) -> Self
+    where
+        F: Future<Output = Result<T, E>> + Send + 'static,
+        T: IntoPy<PyObject>,
+        PyErr: From<E>,
+    {
+        let wrap = async move {
+            let obj = future.await?;
+            // SAFETY: GIL is acquired when future is polled (see `Coroutine::poll`)
+            Ok(obj.into_py(unsafe { Python::assume_gil_acquired() }))
+        };
+        Self {
+            future: Some(Box::pin(panic::AssertUnwindSafe(wrap).catch_unwind())),
+            waker: None,
+        }
+    }
+
+    fn poll(
+        &mut self,
+        py: Python<'_>,
+        throw: Option<PyObject>,
+    ) -> PyResult<IterNextOutput<PyObject, PyObject>> {
+        // raise if the coroutine has already been run to completion
+        let future_rs = match self.future {
+            Some(ref mut fut) => fut,
+            None => return Err(PyRuntimeError::new_err(COROUTINE_REUSED_ERROR)),
+        };
+        // reraise thrown exception it
+        if let Some(exc) = throw {
+            self.close();
+            return Err(PyErr::from_value(exc.as_ref(py)));
+        }
+        // create a new waker, or try to reset it in place
+        if let Some(waker) = self.waker.as_mut().and_then(Arc::get_mut) {
+            waker.reset();
+        } else {
+            self.waker = Some(Arc::new(AsyncioWaker::new()));
+        }
+        let waker = futures_util::task::waker(self.waker.clone().unwrap());
+        // poll the Rust future and forward its results if ready
+        if let Poll::Ready(res) = future_rs.as_mut().poll(&mut Context::from_waker(&waker)) {
+            self.close();
+            return match res {
+                Ok(res) => Ok(IterNextOutput::Return(res?)),
+                Err(err) => Err(PanicException::from_panic_payload(err)),
+            };
+        }
+        // otherwise, initialize the waker `asyncio.Future`
+        if let Some(future) = self.waker.as_ref().unwrap().initialize_future(py)? {
+            // `asyncio.Future` must be awaited; fortunately, it implements `__iter__ = __await__`
+            // and will yield itself if its result has not been set in polling above
+            if let Some(future) = PyIterator::from_object(future).unwrap().next() {
+                // future has not been leaked into Python for now, and Rust code can only call
+                // `set_result(None)` in `ArcWake` implementation, so it's safe to unwrap
+                return Ok(IterNextOutput::Yield(future.unwrap().into()));
+            }
+        }
+        // if waker has been waken during future polling, this is roughly equivalent to
+        // `await asyncio.sleep(0)`, so just yield `None`.
+        Ok(IterNextOutput::Yield(py.None().into()))
+    }
+}
+
+pub(crate) fn iter_result(result: IterNextOutput<PyObject, PyObject>) -> PyResult<PyObject> {
+    match result {
+        IterNextOutput::Yield(ob) => Ok(ob),
+        IterNextOutput::Return(ob) => Err(PyStopIteration::new_err(ob)),
+    }
+}
+
+#[pymethods(crate = "crate")]
+impl Coroutine {
+    fn send(&mut self, py: Python<'_>, _value: &PyAny) -> PyResult<PyObject> {
+        iter_result(self.poll(py, None)?)
+    }
+
+    fn throw(&mut self, py: Python<'_>, exc: PyObject) -> PyResult<PyObject> {
+        iter_result(self.poll(py, Some(exc))?)
+    }
+
+    fn close(&mut self) {
+        // the Rust future is dropped, and the field set to `None`
+        // to indicate the coroutine has been run to completion
+        drop(self.future.take());
+    }
+
+    fn __await__(self_: Py<Self>) -> Py<Self> {
+        self_
+    }
+
+    fn __next__(&mut self, py: Python<'_>) -> PyResult<IterNextOutput<PyObject, PyObject>> {
+        self.poll(py, None)
+    }
+}
--- a/src/coroutine/waker.rs
+++ b/src/coroutine/waker.rs
@@ -1,0 +1,97 @@
+use crate::sync::GILOnceCell;
+use crate::types::PyCFunction;
+use crate::{intern, wrap_pyfunction, Py, PyAny, PyObject, PyResult, Python};
+use futures_util::task::ArcWake;
+use pyo3_macros::pyfunction;
+use std::sync::Arc;
+
+/// Lazy `asyncio.Future` wrapper, implementing [`ArcWake`] by calling `Future.set_result`.
+///
+/// asyncio future is let uninitialized until [`initialize_future`][1] is called.
+/// If [`wake`][2] is called before future initialization (during Rust future polling),
+/// [`initialize_future`][1] will return `None` (it is roughly equivalent to `asyncio.sleep(0)`)
+///
+/// [1]: AsyncioWaker::initialize_future
+/// [2]: AsyncioWaker::wake
+pub struct AsyncioWaker(GILOnceCell<Option<LoopAndFuture>>);
+
+impl AsyncioWaker {
+    pub(super) fn new() -> Self {
+        Self(GILOnceCell::new())
+    }
+
+    pub(super) fn reset(&mut self) {
+        self.0.take();
+    }
+
+    pub(super) fn initialize_future<'a>(&'a self, py: Python<'a>) -> PyResult<Option<&'a PyAny>> {
+        let init = || LoopAndFuture::new(py).map(Some);
+        let loop_and_future = self.0.get_or_try_init(py, init)?.as_ref();
+        Ok(loop_and_future.map(|LoopAndFuture { future, .. }| future.as_ref(py)))
+    }
+}
+
+impl ArcWake for AsyncioWaker {
+    fn wake_by_ref(arc_self: &Arc<Self>) {
+        Python::with_gil(|gil| {
+            if let Some(loop_and_future) = arc_self.0.get_or_init(gil, || None) {
+                loop_and_future
+                    .set_result(gil)
+                    .expect("unexpected error in coroutine waker");
+            }
+        });
+    }
+}
+
+struct LoopAndFuture {
+    event_loop: PyObject,
+    future: PyObject,
+}
+
+impl LoopAndFuture {
+    fn new(py: Python<'_>) -> PyResult<Self> {
+        static GET_RUNNING_LOOP: GILOnceCell<PyObject> = GILOnceCell::new();
+        let import = || -> PyResult<_> {
+            let module = py.import("asyncio")?;
+            Ok(module.getattr("get_running_loop")?.into())
+        };
+        let event_loop = GET_RUNNING_LOOP.get_or_try_init(py, import)?.call0(py)?;
+        let future = event_loop.call_method0(py, "create_future")?;
+        Ok(Self { event_loop, future })
+    }
+
+    fn set_result(&self, py: Python<'_>) -> PyResult<()> {
+        static RELEASE_WAITER: GILOnceCell<Py<PyCFunction>> = GILOnceCell::new();
+        let release_waiter = RELEASE_WAITER
+            .get_or_try_init(py, || wrap_pyfunction!(release_waiter, py).map(Into::into))?;
+        // `Future.set_result` must be called in event loop thread,
+        // so it requires `call_soon_threadsafe`
+        let call_soon_threadsafe = self.event_loop.call_method1(
+            py,
+            intern!(py, "call_soon_threadsafe"),
+            (release_waiter, self.future.as_ref(py)),
+        );
+        if let Err(err) = call_soon_threadsafe {
+            // `call_soon_threadsafe` will raise if the event loop is closed;
+            // instead of catching an unspecific `RuntimeError`, check directly if it's closed.
+            let is_closed = self.event_loop.call_method0(py, "is_closed")?;
+            if !is_closed.extract(py)? {
+                return Err(err);
+            }
+        }
+        Ok(())
+    }
+}
+
+/// Call `future.set_result` if the future is not done.
+///
+/// Future can be cancelled by the event loop before being waken.
+/// See <https://github.com/python/cpython/blob/main/Lib/asyncio/tasks.py#L452C5-L452C5>
+#[pyfunction(crate = "crate")]
+fn release_waiter(future: &PyAny) -> PyResult<()> {
+    let done = future.call_method0(intern!(future.py(), "done"))?;
+    if !done.extract::<bool>()? {
+        future.call_method1(intern!(future.py(), "set_result"), (future.py().None(),))?;
+    }
+    Ok(())
+}
--- a/src/impl_/coroutine.rs
+++ b/src/impl_/coroutine.rs
@@ -1,0 +1,19 @@
+use crate::coroutine::Coroutine;
+use crate::impl_::wrap::OkWrap;
+use crate::{IntoPy, PyErr, PyObject, Python};
+use std::future::Future;
+
+/// Used to wrap the result of async `#[pyfunction]` and `#[pymethods]`.
+pub fn wrap_future<F, R, T>(future: F) -> Coroutine
+where
+    F: Future<Output = R> + Send + 'static,
+    R: OkWrap<T>,
+    T: IntoPy<PyObject>,
+    PyErr: From<R::Error>,
+{
+    let future = async move {
+        // SAFETY: GIL is acquired when future is polled (see `Coroutine::poll`)
+        future.await.wrap(unsafe { Python::assume_gil_acquired() })
+    };
+    Coroutine::from_future(future)
+}
--- a/tests/test_coroutine.rs
+++ b/tests/test_coroutine.rs
@@ -1,0 +1,98 @@
+#![cfg(feature = "macros")]
+#![cfg(not(target_arch = "wasm32"))]
+use std::{task::Poll, thread, time::Duration};
+
+use futures::{channel::oneshot, future::poll_fn};
+use pyo3::{prelude::*, py_run};
+
+#[path = "../src/tests/common.rs"]
+mod common;
+
+fn handle_windows(test: &str) -> String {
+    let set_event_loop_policy = r#"
+    import asyncio, sys
+    if sys.platform == "win32":
+        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+    "#;
+    pyo3::unindent::unindent(set_event_loop_policy) + &pyo3::unindent::unindent(test)
+}
+
+#[test]
+fn noop_coroutine() {
+    #[pyfunction]
+    async fn noop() -> usize {
+        42
+    }
+    Python::with_gil(|gil| {
+        let noop = wrap_pyfunction!(noop, gil).unwrap();
+        let test = "import asyncio; assert asyncio.run(noop()) == 42";
+        py_run!(gil, noop, &handle_windows(test));
+    })
+}
+
+#[test]
+fn sleep_0_like_coroutine() {
+    #[pyfunction]
+    async fn sleep_0() -> usize {
+        let mut waken = false;
+        poll_fn(|cx| {
+            if !waken {
+                cx.waker().wake_by_ref();
+                waken = true;
+                return Poll::Pending;
+            }
+            Poll::Ready(42)
+        })
+        .await
+    }
+    Python::with_gil(|gil| {
+        let sleep_0 = wrap_pyfunction!(sleep_0, gil).unwrap();
+        let test = "import asyncio; assert asyncio.run(sleep_0()) == 42";
+        py_run!(gil, sleep_0, &handle_windows(test));
+    })
+}
+
+#[pyfunction]
+async fn sleep(seconds: f64) -> usize {
+    let (tx, rx) = oneshot::channel();
+    thread::spawn(move || {
+        thread::sleep(Duration::from_secs_f64(seconds));
+        tx.send(42).unwrap();
+    });
+    rx.await.unwrap()
+}
+
+#[test]
+fn sleep_coroutine() {
+    Python::with_gil(|gil| {
+        let sleep = wrap_pyfunction!(sleep, gil).unwrap();
+        let test = r#"import asyncio; assert asyncio.run(sleep(0.1)) == 42"#;
+        py_run!(gil, sleep, &handle_windows(test));
+    })
+}
+
+#[test]
+fn cancelled_coroutine() {
+    Python::with_gil(|gil| {
+        let sleep = wrap_pyfunction!(sleep, gil).unwrap();
+        let test = r#"
+        import asyncio
+        async def main():
+            task = asyncio.create_task(sleep(1))
+            await asyncio.sleep(0)
+            task.cancel()
+            await task
+        asyncio.run(main())
+        "#;
+        let globals = gil.import("__main__").unwrap().dict();
+        globals.set_item("sleep", sleep).unwrap();
+        let err = gil
+            .run(
+                &pyo3::unindent::unindent(&handle_windows(test)),
+                Some(globals),
+                None,
+            )
+            .unwrap_err();
+        assert_eq!(err.value(gil).get_type().name().unwrap(), "CancelledError");
+    })
+}
--- a/guide/src/ecosystem/async-await.md
+++ b/guide/src/ecosystem/async-await.md
@@ -1,5 +1,7 @@
 # Using `async` and `await`
 
+*`async`/`await` support is currently being integrated in PyO3. See the [dedicated documentation](../async-await.md)*
+
 If you are working with a Python library that makes use of async functions or wish to provide
 Python bindings for an async Rust library, [`pyo3-asyncio`](https://github.com/awestlake87/pyo3-asyncio)
 likely has the tools you need. It provides conversions between async functions in both Python and
--- a/newsfragments/3540.added.md
+++ b/newsfragments/3540.added.md
@@ -1,0 +1,1 @@
+Support `async fn` in macros with coroutine implementation
--- a/pyo3-macros-backend/src/pyfunction.rs
+++ b/pyo3-macros-backend/src/pyfunction.rs
@@ -6,7 +6,7 @@ use crate::{
     deprecations::Deprecations,
     method::{self, CallingConvention, FnArg},
     pymethod::check_generic,
-    utils::{ensure_not_async_fn, get_pyo3_crate},
+    utils::get_pyo3_crate,
 };
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
@@ -179,8 +179,6 @@ pub fn impl_wrap_pyfunction(
     options: PyFunctionOptions,
 ) -> syn::Result<TokenStream> {
     check_generic(&func.sig)?;
-    ensure_not_async_fn(&func.sig)?;
-
     let PyFunctionOptions {
         pass_module,
         name,
@@ -230,6 +228,7 @@ pub fn impl_wrap_pyfunction(
         signature,
         output: ty,
         text_signature,
+        asyncness: func.sig.asyncness,
         unsafety: func.sig.unsafety,
         deprecations: Deprecations::new(),
     };
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 
 use crate::attributes::{NameAttribute, RenamingRule};
 use crate::method::{CallingConvention, ExtractErrorMode};
-use crate::utils::{ensure_not_async_fn, PythonDoc};
+use crate::utils::PythonDoc;
 use crate::{
     method::{FnArg, FnSpec, FnType, SelfType},
     pyfunction::PyFunctionOptions,
@@ -188,7 +188,6 @@ pub fn gen_py_method(
     options: PyFunctionOptions,
 ) -> Result<GeneratedPyMethod> {
     check_generic(sig)?;
-    ensure_not_async_fn(sig)?;
     ensure_function_options_valid(&options)?;
     let method = PyMethod::parse(sig, meth_attrs, options)?;
     let spec = &method.spec;
--- a/src/impl_.rs
+++ b/src/impl_.rs
@@ -6,6 +6,8 @@
 //! APIs may may change at any time without documentation in the CHANGELOG and without
 //! breaking semver guarantees.
 
+#[cfg(feature = "macros")]
+pub mod coroutine;
 pub mod deprecations;
 pub mod extract_argument;
 pub mod freelist;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -397,6 +397,8 @@ pub mod buffer;
 pub mod callback;
 pub mod conversion;
 mod conversions;
+#[cfg(feature = "macros")]
+pub mod coroutine;
 #[macro_use]
 #[doc(hidden)]
 pub mod derive_utils;
@@ -469,6 +471,7 @@ pub mod doc_test {
     doctests! {
         "README.md" => readme_md,
         "guide/src/advanced.md" => guide_advanced_md,
+        "guide/src/async-await.md" => guide_async_await_md,
         "guide/src/building_and_distribution.md" => guide_building_and_distribution_md,
         "guide/src/building_and_distribution/multiple_python_versions.md" => guide_bnd_multiple_python_versions_md,
         "guide/src/class.md" => guide_class_md,
--- a/tests/ui/abi3_nativetype_inheritance.stderr
+++ b/tests/ui/abi3_nativetype_inheritance.stderr
@@ -4,6 +4,8 @@ error[E0277]: the trait bound `PyDict: PyClass` is not satisfied
 5 | #[pyclass(extends=PyDict)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `PyClass` is not implemented for `PyDict`
   |
-  = help: the trait `PyClass` is implemented for `TestClass`
+  = help: the following other types implement trait `PyClass`:
+            TestClass
+            Coroutine
   = note: required for `PyDict` to implement `PyClassBaseType`
   = note: this error originates in the attribute macro `pyclass` (in Nightly builds, run with -Z macro-backtrace for more info)
--- a/tests/ui/invalid_pyfunctions.rs
+++ b/tests/ui/invalid_pyfunctions.rs
@@ -7,9 +7,6 @@ fn generic_function<T>(value: T) {}
 fn impl_trait_function(impl_trait: impl AsRef<PyAny>) {}
 
 #[pyfunction]
-async fn async_function() {}
-
-#[pyfunction]
 fn wildcard_argument(_: i32) {}
 
 #[pyfunction]
--- a/tests/ui/invalid_pyfunctions.stderr
+++ b/tests/ui/invalid_pyfunctions.stderr
@@ -10,29 +10,21 @@ error: Python functions cannot have `impl Trait` arguments
 7 | fn impl_trait_function(impl_trait: impl AsRef<PyAny>) {}
   |                                    ^^^^
 
-error: `async fn` is not yet supported for Python functions.
-
-       Additional crates such as `pyo3-asyncio` can be used to integrate async Rust and Python. For more information, see https://github.com/PyO3/pyo3/issues/1632
-  --> tests/ui/invalid_pyfunctions.rs:10:1
-   |
-10 | async fn async_function() {}
-   | ^^^^^
-
 error: wildcard argument names are not supported
-  --> tests/ui/invalid_pyfunctions.rs:13:22
+  --> tests/ui/invalid_pyfunctions.rs:10:22
    |
-13 | fn wildcard_argument(_: i32) {}
+10 | fn wildcard_argument(_: i32) {}
    |                      ^
 
 error: destructuring in arguments is not supported
-  --> tests/ui/invalid_pyfunctions.rs:16:26
+  --> tests/ui/invalid_pyfunctions.rs:13:26
    |
-16 | fn destructured_argument((a, b): (i32, i32)) {}
+13 | fn destructured_argument((a, b): (i32, i32)) {}
    |                          ^^^^^^
 
 error: required arguments after an `Option<_>` argument are ambiguous
        = help: add a `#[pyo3(signature)]` annotation on this function to unambiguously specify the default values for all optional parameters
-  --> tests/ui/invalid_pyfunctions.rs:19:63
+  --> tests/ui/invalid_pyfunctions.rs:16:63
    |
-19 | fn function_with_required_after_option(_opt: Option<i32>, _x: i32) {}
+16 | fn function_with_required_after_option(_opt: Option<i32>, _x: i32) {}
    |                                                               ^^^
--- a/tests/ui/invalid_pymethods.rs
+++ b/tests/ui/invalid_pymethods.rs
@@ -163,11 +163,6 @@ impl MyClass {
 
 #[pymethods]
 impl MyClass {
-    async fn async_method(&self) {}
-}
-
-#[pymethods]
-impl MyClass {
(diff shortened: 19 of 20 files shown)


tests/test_coroutine.rs
```diff
@@ -0,0 +1,9 @@
+#![cfg(feature = "async")]
+
+use pyo3::prelude::*;
+
+#[path = "../src/tests/common.rs"]
+mod common;
+
+#[pyfunction]
+async fn _async_fn_is_supported() {}
```
davidhewitt, line 9: So I suppose that the challenge now is that even if we've gotten this far, we can't await on other Rust stuff that requires an async runtime because there isn't a Rust async runtime running?

I wonder, should we consider adding `impl Future for PyAny`, with the item being `PyResult<PyAny>`? That way, I guess we can start awaiting other Python objects, assuming a Python runtime is running...

(I haven't thought that second part through to completion.)

src/impl_/coroutine.rs
```diff
@@ -0,0 +1,19 @@
...
+where
+    F: Future<Output = R> + Send + 'static,
+    R: OkWrap<T> + Send,
+    T: IntoPy<PyObject> + Send,
+    R::Error: Send,
+    PyErr: From<R::Error>,
+{
+    Coroutine::from_future(async move {
+        // SAFETY: GIL is acquired when future is polled (see `Coroutine::poll`)
+        future.await.wrap(unsafe { Python::assume_gil_acquired() })
```
davidhewitt, line 17: Can this transformation be moved inside `from_future`, to put the safety invariant closer to `Coroutine::poll` ?

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+    #[allow(clippy::type_complexity)]
+    throw_callback: Option<Box<dyn FnMut(&PyAny) + Send>>,
+    waker: Option<Arc<PyFutureWaker>>,
+}
+
+impl Coroutine {
+    /// Wrap a boxed future in to a Python coroutine.
+    ///
+    /// Coroutine `send` polls the wrapped future, ignoring the value passed
+    /// (should always be `None` anyway).
```
davidhewitt, line 31: I suppose in a future Rust with generators, then the `send` value might be interesting?

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+    match result {
+        IterNextOutput::Yield(ob) => Ok(ob),
+        IterNextOutput::Return(ob) => Err(PyStopIteration::new_err(ob)),
+    }
+}
+
+#[pymethods(crate = "crate")]
+impl Coroutine {
+    fn send(&mut self, py: Python<'_>, _value: &PyAny) -> PyResult<PyObject> {
+        iter_result(self.poll(py, None)?)
```
davidhewitt, line 123: Should we throw if `_value` is not `None`?

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// Coroutine `send` polls the wrapped future, ignoring the value passed
+    /// (should always be `None` anyway).
+    ///
+    /// If `throw_callback` has been provided, coroutine `throw` calls it
+    /// with the exception passed, and poll the wrapped future.
+    /// Otherwise, the exception is reraised and the future dropped.
+    #[allow(clippy::type_complexity)]
+    pub fn new(
+        future: Pin<Box<dyn Future<Output = PyResult<PyObject>> + Send>>,
+        throw_callback: Option<Box<dyn FnMut(&PyAny) + Send>>,
```
davidhewitt, line 39: If I read correctly this is a more advanced API to do with cancellation rather than just `async fn` syntax?

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+                cb(exc)
+            } else {
+                self.future.take();
+                return Err(PyErr::from_value(exc));
+            };
+        }
+        loop {
+            // create a waker from an `asyncio.Future`, or try to modify it in place
+            if let Some(waker) = self.waker.as_mut().and_then(Arc::get_mut) {
+                waker.future = asyncio_future(py)?;
```
davidhewitt, line 86: So if I read it correctly, with this PR the way that we support additional async runtimes on the Python side would be by making the `waker` able to support various async interfaces?

davidhewitt (commented):
Thank you! I think this will need to go through several cycles of review as I get my head around it, so please bear with me 🙏

src/impl_/coroutine.rs
wyfo: I wanted to do so, but it requires to expose `OkWrap` trait. If you are ok with that move, I can update `from_future` signature.

src/coroutine.rs
wyfo: If PyO3 provide a `Generator` type, maybe, but python coroutine seems to always ignore the value passed.

src/coroutine.rs
wyfo: As written above, Python coroutines seems to always ignore the value, so prefer to keep this behavior.
Example:
```python
import asyncio
async def main():
    async def f():
        await asyncio.sleep(1)
    coro = f()
    fut = coro.send(None)
    fut.set_result(None)
    try:
        coro.send(42)  # doesn't raise because of 42
    except StopIteration:
        print("ok")
```

src/coroutine.rs
wyfo: Exactly. That's what I've done in *pyo3-async*, with a [`CoroutineWaker`](https://github.com/wyfo/pyo3-async/blob/e55eb3b5d8743693ceee1ffa99689fb3e64b7ba0/src/coroutine.rs#L15) trait and its implementations for *asyncio*, *trio* and *anyio*.

tests/test_coroutine.rs
wyfo: > we can't await on other Rust stuff that requires an async runtime because there isn't a Rust async runtime wrunning

Actually, you don't need a running Rust runtime to have something to await. `poll_fn`, `AtomicWaker` `thread::spawn`, etc. can do the job. I haven't write more than an empty test simply because I haven't taken the time yet.

> I wonder, should we consider adding impl Future for PyAny, with the item being PyResult<PyAny>?

Things are quite more complex than this when it comes to await Python awaitable. Contrary to Rust, Python coroutines should always be run to completion (no RAII-style future that can be dropped in the middle of the execution), that's why they should always be executed in a spawned task. So you should only await `asyncio.Future` or `asyncio.Task`, see pyo3-async [`FutureWrapper`](https://docs.rs/pyo3-async/latest/pyo3_async/asyncio/struct.FutureWrapper.html) for example.
(pyo3-async also provides [`AwaitableWrapper`](https://docs.rs/pyo3-async/latest/pyo3_async/asyncio/struct.AwaitableWrapper.html) to run manually an awaitable, but as mentioned above, this is not a good thing to do and I planned to remove it).

But that get event more complicated if you want to directly spawn the asyncio task in Rust. pyo3-asyncio does that, but it requires to keep the event loop reference all along, and to acquire the GIL at every spawn, and it doest support task group and so on. I've thought about a way to improve this a bit, with a "spawner" (which could use a task group) task running in asyncio background and fed by an mpsc; writing to the mpsc would require to acquire the GIL only when the receiver is awaiting, so not always, and being in the same event loop thread make possible to avoid `call_soon_threadsafe` call for callback registration. You can even go further with some `await asyncio.sleep(0)`-based spin loop on mpsc receiver side in order to avoid awaiting, and thus to avoid GIL acquiring on mpsc writer thread for the wakeup.
But that seems to be a little too much complex to be integrated in PyO3, even more considering exception handling in tasks if you wanna play with task groups.

I think that awaiting Python objects should be a completely seperate topic for the moment.

adamreichold:
Just as a process note, I would be glad if we could aim for at least two maintainer reviews here as this appears to be fundamental expansion of the scope of PyO3. Of course, not necessarily mine, just more than one to increase the bus factor just a little bit.

pyo3-macros-backend/src/method.rs
```diff
@@ -218,6 +218,7 @@ impl CallingConvention {
 }
 
 pub struct FnSpec<'a> {
+    pub asyncness: bool,
```
adamreichold, line 221: I think it would improve readability if we put this next to `unsafety` and handle it similarly w.r.t. the type.

pyo3-macros-backend/src/utils.rs
```diff
@@ -137,17 +137,6 @@ impl quote::ToTokens for PythonDoc {
     }
 }
 
-pub fn ensure_not_async_fn(sig: &syn::Signature) -> syn::Result<()> {
```
adamreichold, line 140: What is the error message if the optional dependency is disabled and an `async` function or method is used?

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+use std::sync::Arc;
+use std::task::{Context, Poll};
+
+const COROUTINE_REUSED_ERROR: &str = "cannot reuse already awaited coroutine";
+
+/// Python coroutine wrapping a [`Future`].
+#[pyclass(crate = "crate")]
+pub struct Coroutine {
+    future: Option<Pin<Box<dyn Future<Output = PyResult<PyObject>> + Send>>>,
+    #[allow(clippy::type_complexity)]
```
adamreichold, line 22: I think type aliases would be nice, e.g. `PyFuture` and `PyThrowCallback`.

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+        let future_rs = match self.future {
+            Some(ref mut fut) => fut,
+            None => return Err(PyRuntimeError::new_err(COROUTINE_REUSED_ERROR)),
+        };
+        // pass thrown exception to throw callback if there is one, otherwise reraise it
+        if let Some(exc) = throw {
+            if let Some(cb) = &mut self.throw_callback {
+                cb(exc)
+            } else {
+                self.future.take();
```
adamreichold, line 79: Please make dropping the future on the floor more explicit, e.g. using `let _ = ..` or `drop(..)` here and below.

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+                // `set_result(None)` in `ArcWake` implementation, so it's safe to unwrap
+                return Ok(IterNextOutput::Yield(future.unwrap().into()));
+            }
+            // Python future was done, i.e. `FutureWaker::wake_by_ref` has been called in
+            // `future_rs.as_mut().poll(...)`, so continue looping
+        }
+    }
+}
+
+pub(crate) fn iter_result(result: IterNextOutput<PyObject, PyObject>) -> PyResult<PyObject> {
```
adamreichold, line 113: This will our handling of iterators even more of a mess, won't it? So this is basically what we want to do with all iterators but it would apply only to coroutines? Maybe we should align all of this if we go ahead?

Cargo.toml
```diff
@@ -31,6 +31,9 @@ unindent = { version = "0.2.1", optional = true }
 # support crate for multiple-pymethods feature
 inventory = { version = "0.3.0", optional = true }
 
+# coroutine implementation
+futures = { version = "0.3", optional = true }
```
adamreichold, line 35: Do we need all of `futures` or just `futures-task`?

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    ///  Wrap a generic future into a Python coroutine.
+    pub fn from_future<F, T, E>(future: F) -> Self
+    where
+        F: Future<Output = Result<T, E>> + Send + 'static,
+        T: IntoPy<PyObject> + Send,
+        E: Send,
+        PyErr: From<E>,
+    {
+        let wrap = async move {
```
adamreichold, line 56: I think I would prefer using `FutureExt::map` here to make it more obvious what is happening.

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+    static FUTURE: GILOnceCell<PyObject> = GILOnceCell::new();
+    let import_future = || -> PyResult<_> {
+        let module = py.import("asyncio")?;
+        Ok(module.getattr("Future")?.into())
+    };
+    FUTURE.get_or_try_init(py, import_future)?.call0(py)
+}
+
+/// Alternative to `std::thread::current().id()` without unnecessary Arc clone + drop.
+pub(crate) type ThreadId = usize;
```
adamreichold, line 211: I am very much not a fan of implementing our own thread ID. Even if we do decided that this is a worthwhile performance optimization, I think it should be done in a separate change to keep this as simple as possible.

pyo3-macros-backend/src/utils.rs
wyfo: You're right, I have not think about this use case. I will add a `compile_error`.

src/coroutine.rs
adamreichold: Or alternatively, I don't think we we need thread ID at all. We just want to know if the current thread hosts the event loop, not which one actually is, i.e. a thread-local flag should suffice, e.g.

```rust
thread_local! {
  static HOSTS_EVENT_LOOP: Cell<bool> = const { Cell::new(false) };
}
```

which we set to `true` when the waker is constructed.

(Note that this should also conditionally use `const` syntax as we already do in

https://github.com/PyO3/pyo3/blob/2b91fea94f4f6371b995f4c649c1ea80bd5f4dd2/src/gil.rs#L15

but this might also better be a optimization follow up.)

src/coroutine.rs
wyfo: `PyFuture` can be ambiguous, because others `PyXxx` (`PyDict`, `PyIterator`, etc.) are Python objects, and there is a `Future` type in Python (there are two in fact).
Strangely, clippy only warns about type complexity for `Option<Box<dyn FnMut(&PyAny) + Send>>`, not for the future.
In pyo3-async, I've named it `ThrowCallback`, I could reuse the same name here.

src/coroutine.rs
wyfo: Do you prefer `self.future = None` instead?

src/coroutine.rs
adamreichold: I am not keen on any of the names and will defer to your experience in building this, I just think any type aliases will be an improvement.

src/coroutine.rs
adamreichold: (Of course, if this would make the different on depending on `futures` versus just `futures-task`, we should keep open-coding it and reduce our dependency closure.)

src/coroutine.rs
adamreichold: No, I think this is more implicit than `self.future.take()` because the point is not only modifying `self.future` but also why this mean for the value previously stored there.

Hence, I think `let _ = self.future.take();`  or `drop(self.future.take());` would be best. 

(These implicit drops have been shown to be one of the most common sources of bugs in unsafe code, i.e. forgetting to use `ptr::write` to suppress them for invalid values.)

src/coroutine.rs
adamreichold: (Sorry if this not clear, but this is mainly directed at @davidhewitt, not so much at @wyfo.)

Cargo.toml
wyfo: Right, only `futures-task` is needed.

src/coroutine.rs
wyfo: > this would make the difference on depending on futures versus just futures-task

That's what I wanted to say :)

src/coroutine.rs
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+    }
+}
+
+impl ArcWake for PyFutureWaker {
+    fn wake_by_ref(arc_self: &Arc<Self>) {
+        // Choose appropriate wake method depending on wether it is called on the same thread
+        // than the event loop
+        if current_thread_id() == arc_self.thread_id {
```
adamreichold, line 192: Just to make sure because I do not know this and it seems to be assumed here: The asyncio event loop can nether be moved to a different thread? Not even stopped on one and restart on another?

wyfo: That's a good concern, and I think it's indeed possible to move the event loop to a different thread. So this "optimization" is wrong, I will remove it.

src/coroutine.rs
wyfo: I honestly think `__next__` should allow to return `PyResult<PyObject>` in addition to `PyResult<IterNextOutput<PyObject, PyObject>>` or `PyResult<Option<PyObject>>`.
It would remove the need for this small utilitary function.
But I did want to touch too much parts of PyO3 in this first draft.

src/coroutine.rs
wyfo: As written in reply to your other comment, this thread id "optimization" is wrong and will be removed.

src/coroutine.rs
wyfo: There is no `unsafe` code here, and I think the main point is not to drop the Rust future but to reset the field to `None`.
That's why I prefer in fact `self.future = None`.
But I will follow the maintainer directive here :)

pyo3-macros-backend/src/utils.rs
wyfo: Actually, I can't add a `compile_error!` simply because it's not possible to know at macro expansion if the feature is enabled or not.
So the error message will be that `_pyo3::impl_::coroutine` doesn't exists. I know that this error message is not very explicit, but because the feature activated by default, I don't think it really matter.

A solution could be to export a global declarative macro like this
```rust
#[cfg(feature = "async")]
macro_rules! check_async {
    ($($tokens:tt)*) => { 
        $($tokens)*
    };
}

#[cfg(not(feature = "async"))]
macro_rules! check_async {
    ($($tokens:tt)*) => { 
        compile_error!("async fn support requires pyo3 \"async\" feature")
    };
}
```
but do you think it's worth adding a global macro to check for a default feature?

wyfo:
I've updated the PR with the feedbacks.

I've also completely refactored the waker to lazily initialize the Python future (so there is no object allocated when the future is already ready), but I've used `unsafe` code.
I can also replace the combination `AtomicBool` + `UnsafeCell<MaybeUninit<PyObject>>` by a `Mutex` + an `enum`, using only safe code thus, but it will obviously impact performance.

Moreover, I've added a method to release the GIL while polling the future (it was the first request when I released pyo3-async).

Obviously, I will need to work on test now, especially with the unsafe code added (if you want to keep it).

src/coroutine.rs
adamreichold: > There is no unsafe code here

The remark was only to illustrate that these implicit drops are a common source of confusion for Rust programmers. I know they are for me which why I asked for making them more explicit.

> I think the main point is not to drop the Rust future but to reset the field to None.

Maybe I do misunderstand then: My understanding was that this marks the coroutine as complete (by raising an exception in this case) _and_ cancels the underlying Rust future to keep Python and Rust in sync?

Admittedly, the situation is different in line 129 below where the Rust completes normally and we just want to avoid polling it again (i.e. `self.future = None;`) as the drop there does not really "cancel" anything, but I'd argue the uniformity of doing this in the same everywhere is preferable.

src/coroutine.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+}
+
+/// Lazy `asyncio.Future` wrapper, implementing [`ArcWake`] by calling `Future.set_result`.
+///
+/// asyncio future is let uninitialized until [`AsyncioWaker::initialize_future`] is called.
+/// If the waker is waken before future initialization (during Rust future polling),
+/// its state is updated atomically, in order to prevent future initialization coming after.
+///
+/// The waker also embeds the event loop thread id to determine if it's necessary to use
```
adamreichold, line 186: There are few comments here and there still referencing the thread-ID-based optimization.

adamreichold:
> I've also completely refactored the waker to lazily initialize the Python future (so there is no object allocated when the future is already ready), but I've used unsafe code.
I can also replace the combination AtomicBool + UnsafeCell<MaybeUninit<PyObject>> by a Mutex + an enum, using only safe code thus, but it will obviously impact performance.

Since you acquire the GIL anyway, did you consider our own [`GILOnceCell`](https://docs.rs/pyo3/latest/pyo3/sync/struct.GILOnceCell.html) to handle the lazy initialization? I am not if it is up to the more involved lifecycle but maybe we could extend it slightly to cover this use case?

adamreichold:
(Otherwise, I would prefer an initial safe version using `Mutex` and an enum which can optimize in a follow-up PR.)

src/coroutine.rs
```diff
@@ -0,0 +1,314 @@
...
+/// Set the result of an `asyncio.Future` (thread safe)
+fn future_set_result_threadsafe(future: &PyAny) {
+    // `set_result` must be called in event loop thread, so it requires `call_soon_threadsafe`
+    let set_result = future
+        .getattr(intern!(future.py(), "set_result"))
+        .expect("error while calling Future.set_result");
+    let event_loop = future
+        .call_method0(intern!(future.py(), "get_loop"))
+        .expect("error while calling Future.get_loop");
+    let call_soon_threadsafe = event_loop.call_method1(
```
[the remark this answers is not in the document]
wyfo: A future is normally bound to one event loop (and its `_loop` attribute is private), so I think its safe to cache the method.

src/coroutine.rs
```diff
@@ -0,0 +1,314 @@
...
+        let wrap = async move {
+            let obj = future.await?;
+            // SAFETY: GIL is acquired when future is polled (see `Coroutine::poll`)
+            Ok(obj.into_py(unsafe { Python::assume_gil_acquired() }))
+        };
+        Self::new(Box::pin(wrap), None)
+    }
+
+    ///  Wrap a generic future into a Python coroutine, releasing GIL when future is polled.
+    pub fn from_future_allow_threads<F, T, E>(future: F) -> Self
```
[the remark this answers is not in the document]
wyfo: It's not essential, i will revert the commit.

wyfo:
> Since you acquire the GIL anyway, did you consider our own [GILOnceCell](https://docs.rs/pyo3/latest/pyo3/sync/struct.GILOnceCell.html) to handle the lazy initialization? I am not if it is up to the more involved lifecycle but maybe we could extend it slightly to cover this use case?

I've considered it, but it makes

src/coroutine.rs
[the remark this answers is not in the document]
wyfo: Actually, the coroutine may yield several futures, and if you want to make awful things, you could run the coroutine manually (calling `send`) in one event loop, and then in another.

However, I'm not sure that caching the method is a good idea in the end, because I think that most of the Rust futures wrapped should only be waken once; if you have a bigger future with several of suspension points, you may want to spawn a task in a Rust runtime to have the future polled on Rust side.
I don't really want to penalize small futures with an additional allocation for the cached method handle, so I prefer to keep the method call.

By the way, calling directly `asyncio.Future()` was a mistake, because `EventLoop.create_future()` is the recommended method. I will fix this.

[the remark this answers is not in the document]
adamreichold: Fair enough, the resulting simplicity is also a nice benefit of not trying to memoize anything.

pyo3-macros-backend/src/utils.rs
davidhewitt: I think one way to solve this is add an `async` feature to `pyo3-macros-backend`. Can then add a friendly error message instructing to enable the feature if it's not enabled.

wyfo: This solution would bring the strange edge case of having it enabled on macro side but not on pyo3 side.

pyo3 has only 2 default features: `async` and `macros`. What could be the reason to disable default features, and only use `macros`? (I can see the interest of disabling macros only, but not the opposite) IMO this edge case doesn't really matter.

Otherwise, I still prefer a `#[doc(hidden)]` exported declarative macro.

adamreichold: > pyo3 has only 2 default features: async and macros. What could be the reason to disable default features, and only use macros?

One thing that comes to mind is avoid the related optional dependencies. If we say that "futures-task just is not big enough to want to avoid it", then why have a feature to disable it in the first place. Otherwise, it should be possible to build PyO3 with the smallest possible dependency closure. (Especially for libraries which extend PyO3 itself and should not pollute the dependency management of their users.)

mejrs: My preference would be to just not have this async feature. I don't think futures is big enough to justify the added complexity.

src/coroutine.rs
wyfo: I've reused `Coroutine::close` in `Coroutine::poll`, I find it way better now.

pyo3-macros-backend/src/utils.rs
wyfo: The feature is dropped

wyfo:
New version with `GILOnceCell` in the waker!
I've added some tests (fixing a bug of `Future.set_result` called after cancellation thanks to them).

I've tried the `throw_callback` feature in a test, and it works fine. However, I'm not fully sure of the design yet.

src/coroutine.rs
davidhewitt: We have a long discussion in #3190 about reworking `__next__` and `__anext__`. Really just needs someone to write up a bunch of test cases / examples so we can commit to a design. I keep meaning to do it, but am sort of trying to avoid breaking too much at the same time as #3382

src/coroutine.rs
davidhewitt: Ok a good reason to just ignore it 👍

src/sync.rs
```diff
@@ -167,6 +167,20 @@ impl<T> GILOnceCell<T> {
...
+    pub fn take(&mut self) -> Option<T> {
+        self.0.get_mut().take()
+    }
+
+    /// Consumes the cell, returning the wrapped value.
+    ///
+    /// Returns None if the cell has not yet been written.
+    pub fn into_inner(self) -> Option<T> {
+        self.0.into_inner()
+    }
```
davidhewitt, line 183: Would you be willing to split these into a separate PR please? I think they are reasonable features which are worth documenting in their own changelog entry, maybe add tests at the same time etc.

src/coroutine.rs
```diff
@@ -0,0 +1,230 @@
...
+    ///
+    /// Coroutine `send` polls the wrapped future, ignoring the value passed
+    /// (should always be `None` anyway).
+    ///
+    /// If `throw_callback` has been provided, coroutine `throw` calls it
+    /// with the exception passed, and poll the wrapped future.
+    /// Otherwise, the exception is reraised and the future dropped.
+    pub fn new(
+        future: Pin<Box<dyn Future<Output = PyResult<PyObject>> + Send>>,
+        throw_callback: Option<ThrowCallback>,
```
davidhewitt, line 38: I think given that our first pass at the `async` syntax doesn't have a way to deal with `throw_callback`, does it make sense to break this out into a separate PR? I am not mandating it, just wondering if we want to keep this API private for now.

(And if we do keep this API private, do we even need to make the `Coroutine` type public? If it's private, that might make it easier for us to iterate on functionality in the short term.)

tests/test_coroutine.rs
davidhewitt: Thank you, that's a lot of very good reasons why we don't want to do that yet 👍

davidhewitt (commented):
Thank you for the continued iteration here, I'm feeling relatively optimistic that we can get to `async fn` support from this approach. That said, it's definitely an expansion in PyO3's complexity so I agree we'll need a few maintainer passes, and I'd prefer we merged functionality in baby steps where we can be confident in each piece. Sorry I didn't manage to keep up with the threads from the last few days.

I've been thinking a bit about how we make sure that this functionality is a solid foundation which the rest of the ecosystem can build on.

I _think_ that `pyo3-asyncio` might be somewhat compatible with its existing `TaskLocals` and other helpers, however I guess it might want to return a custom object instead of our `Coroutine`. (I haven't thought particularly hard about this.)

Similarly I guess the way that users could support other async runtimes is by supplying their own `Coroutine` implementation?

From that, it seems to me like we might want to have `pyo3(coroutine_type = X)` as an option on async functions to switch the returned concrete type. Plus maybe a global api `pyo3::set_coroutine_type::<X>()` to do this once for the whole extension module.

However I think really we don't need to replace the whole coroutine type, just the waker? (After all, all `async def` functions in Python return the same coroutine type...)

src/sync.rs
wyfo: Yes I will.

src/coroutine.rs
davidhewitt: E.g. do we want a `pyo3(on_cancel = ...)` option for this, or will it always require the user manually constructing the Coroutine? It feels like an additional feature over and above the basics, and more for me to get my head around 😵‍💫

wyfo:
> I think that pyo3-asyncio might be somewhat compatible with its existing TaskLocals and other helpers, however I guess it might want to return a custom object instead of our Coroutine. (I haven't thought particularly hard about this.)

pyo3-asyncio is completely different, as `future_into_py` simply returns an `asyncio.Future`. IMO, this PR is orthogonal to pyo3-asyncio, and poy3 `Coroutine` will simply replace pyo3-asyncio `future_into_py` in the future, as `Coroutine` can also be used to wrap tokio/async-std `JoinHandle`.

> Similarly I guess the way that users could support other async runtimes is by supplying their own Coroutine implementation?

Do you talk about other Python async runtime? Indeed, it's about coroutine implementation, and more particularly the waker (the current coroutine code would still require some adaptations to make it generic, but nothing complicated). 
I will update pyo3-async with the improvements from this PR to test.

davidhewitt:
>  IMO, this PR is orthogonal to `pyo3-asyncio`, and poy3 Coroutine will simply replace pyo3-asyncio future_into_py in the future, as Coroutine can also be used to wrap tokio/async-std JoinHandle.

Yes, I agree that mostly this PR will not change things for `pyo3-asyncio`. What would be nice is for us to understand here how existing users of `pyo3-asyncio` will use this syntax. E.g. how much `pyo3-asyncio` will need to change its docs, if it's just a case of adding `async fn` and removing `future_into_py` calls, or if there need to be new abstractions in that library to allow migration. cc @awestlake87 in case you have any immediate thoughts on this.

Having at least a fork or branch of pyo3-asyncio with some examples using this `async fn` syntax would be very instructive for me. I am happy to make time to try this myself, but it might take at least a couple of weeks to find that opportunity 😄

wyfo:
New version here, with a rework of the cancellation handling! Now I'm quite satisfied with it (but I had to replace `futures-task` by `futures-util`).

I've also updated macros to take in account `CoroutineCancel` parameter (without additional attribute) the same way `Python` parameter is. You can see an example in the test, it's quite easy to use now.

@davidhewitt I took the liberty of resolving some of your conversations, but one remains open, I would like to have your opinion about https://github.com/PyO3/pyo3/pull/3540#discussion_r1369720307

src/coroutine.rs
wyfo: You convinced me, and I've removed `throw_callback` (I can revert it back ofc), but I came with an new way of handling cancellation, see https://github.com/PyO3/pyo3/pull/3540#issuecomment-1784128394.

However, I think it's necessary to keep `Coroutine` public, especially for capture issue – you may not want to capture all the parameters of your function into your future, so you may want to have a sync function returning a `Coroutine` instead (the same way you would return `impl Future<Output=...>` instead of using an async function).
It also allows to have synchronous code executed before returning the coroutine.

pyo3-macros-backend/src/utils.rs
```diff
@@ -41,6 +41,19 @@ pub fn is_python(ty: &syn::Type) -> bool {
...
 }
 
+/// Check if the given type `ty` is `pyo3::coroutine::CoroutineCancel`.
+pub fn is_coroutine_cancel(ty: &syn::Type) -> bool {
+    match unwrap_ty_group(ty) {
+        syn::Type::Path(typath) => typath
+            .path
+            .segments
+            .last()
+            .map(|seg| seg.ident == "CoroutineCancel")
```
davidhewitt, line 51: From experience with these proc macros we've generally found that having to rely on text idents for making user-facing behavioural changes is not great. (e.g. what if the user happens to make their own type called `CoroutineCancel`? We can't distinguish that. Also this functionality breaks if the user chooses to rename the import.)`

I know we do it for `py: Python` and also `Option`; I'm hoping one day we remove the special-casing on `Option`. The `py: Python` is so common that it's a reasonably clear ergonomic win but I would remove the text-based ident detection in a heartbeat.

An alternative way to do this would be to have `#[pyo3(cancel_handle = "x")]` which then makes `x` get treated as this special argument.

tests/test_coroutine.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+#[test]
+fn coroutine_cancel() {
+    #[pyfunction]
+    async fn cancellable_sleep(seconds: f64, mut cancel: CoroutineCancel) -> usize {
+        futures::select! {
+            _ = sleep(seconds).fuse() => 42,
+            _ = cancel.cancelled().fuse() => 0,
+        }
```
davidhewitt, line 97: This works quite nicely, and it seems to me to be a fairly good mapping to Python's `await` keyword potentially raising a `CancelledError`. 👍

src/coroutine/cancel.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Poll to retrieve the exception thrown in the associated coroutine.
+    pub fn poll_cancelled(&mut self, cx: &mut Context<'_>) -> Poll<PyObject> {
+        if self.is_cancelled() {
+            return Poll::Ready(self.take_exception());
+        }
+        self.0.waker.register(cx.waker());
+        if self.is_cancelled() {
+            return Poll::Ready(self.take_exception());
+        }
```
davidhewitt, line 48: I wonder, is all this complexity needed? Is a `oneshot` channel carrying the exception able to achieve the same effect?

src/coroutine.rs
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// Coroutine `send` polls the wrapped future, ignoring the value passed
+    /// (should always be `None` anyway).
+    ///
+    /// `Coroutine `throw` drop the wrapped future and reraise the exception passed
+    pub fn from_future<F, T, E>(future: F) -> Self
+    where
+        F: Future<Output = Result<T, E>> + Send + 'static,
+        T: IntoPy<PyObject> + Send,
+        E: Send,
```
davidhewitt, line 40: Where does the send bound on `E` come from?

src/coroutine/waker.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+/// Call `future.set_result` if the future is not done.
+///
+/// Future can be cancelled by the event loop before being waken.
+/// See <https://github.com/python/cpython/blob/main/Lib/asyncio/tasks.py#L452C5-L452C5>
+#[pyfunction(crate = "crate")]
+fn release_waiter(future: &PyAny) -> PyResult<()> {
+    if !future.call_method0("done")?.extract::<bool>()? {
```
davidhewitt, line 92: Why is the check for `!done` necessary?

src/coroutine/waker.rs
```diff
@@ -0,0 +1,96 @@
...
+        let future = event_loop.call_method0(py, "create_future")?;
+        Ok(Self { event_loop, future })
+    }
+
+    fn set_result(&self, py: Python<'_>) -> PyResult<()> {
+        static RELEASE_WAITER: GILOnceCell<Py<PyCFunction>> = GILOnceCell::new();
+        let release_waiter = RELEASE_WAITER
+            .get_or_try_init(py, || wrap_pyfunction!(release_waiter, py).map(Into::into))?;
+        // `Future.set_result` must be called in event loop thread,
+        // so it requires `call_soon_threadsafe`
```
davidhewitt, line 68: This sounds plausible but I couldn't find it from documentation?

src/coroutine.rs
```diff
@@ -0,0 +1,157 @@
...
+        let waker = futures_util::task::waker(self.waker.clone().unwrap());
+        // poll the Rust future and forward its results if ready
+        if let Poll::Ready(res) = future_rs.as_mut().poll(&mut Context::from_waker(&waker)) {
+            self.close();
+            return Ok(IterNextOutput::Return(res?));
+        }
+        // otherwise, initialize the waker `asyncio.Future`
+        if let Some(future) = self.waker.as_ref().unwrap().initialize_future(py)? {
+            // `asyncio.Future` must be awaited; fortunately, it implements `__iter__ = __await__`
+            // and will yield itself if its result has not been set in polling above
```
davidhewitt, line 114: I can't help but wonder if somehow deep down in the Rust future, if it awaits a Python awaitable, whether we want to yield whatever the Python awaitable that blocked it here, rather than make a new `asyncio.Future`. That seems like a possible avenue to support all Python runtimes automatically.

However, it needs some way to lift whatever that object was up to here. Maybe that's done through the waker, or maybe that's done through our own mini thread-local "runtime"?

Of course, there's also the possibility that this Rust future never awaits on anything from Python, and if that's the case, doing this `asyncio` dance seems correct.

src/coroutine/cancel.rs
wyfo: `oneshot` is one shot, but cancellation is not. Indeed, you can ignore it (see [last paragraph](https://docs.python.org/3/library/asyncio-task.html#task-cancellation)) and [uncancel](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.uncancel), so you may have several cancellation to handle.
That's the reason of this complexity.

src/coroutine.rs
wyfo: Nowhere, I will remove it (`Send`bound on `T` is also useless)

pyo3-macros-backend/src/utils.rs
wyfo: I agree. I started in this direction, but I finally chose the fast and dirty way to show you the feature first.

I will add a an attribute.

src/coroutine/waker.rs
wyfo: Future result/exception can only be set once, it raises `InvalidStateError` otherwise. In our case, the future may be cancelled by the event loop when the waker is called, so this check avoid an unnecessary `InvalidStateError`.

I've seen this exception raised in my cancellation test, before adding this `!done` check. Anyway, asyncio implementation does the same.

src/coroutine/waker.rs
wyfo: `asyncio.Future` is [documented](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future) as "not thread-safe".
Lot of [StackOverflow posts](https://stackoverflow.com/a/57145512) also cover this subject. 

When you look in the code, the reason is simple, there is no synchronization mechanism to synchronize concurrent calls to `set_result`, `set_exception`, etc. (contrary to `concurrent.futures.Future`)

src/coroutine.rs
wyfo: Actually, I've an idea to make this possible, and in a quite elegant way, but it would require a nightly [feature](https://doc.rust-lang.org/std/task/struct.Waker.html#method.as_raw).

> That seems like a possible avenue to support all Python runtimes automatically.

This would indeed be a way to support arbitrary awaitables. You would still need to target a particular runtime for arbitrary Rust futures (but you could also have a runtime-agnostic coroutine which would only support Python awaitables). However, it would not be possible to do a `select` between two Python awaitables, (because you can't yield two objects), so you would have to `panic!` in that case, and I'm not sure about the ergonomy of this.

An additional side effect would be cancellation, because if you delegate to the Python awaitable, you should also delegate cancellation, and `CoroutineCancel` should not register the exception in that case.

I really like this idea though! I may implement a POC in the next days.

pyo3-macros-backend/src/utils.rs
adamreichold: To be honest, I feel like this PR is repeatedly derailing as new stuff is added on top after we tried to whittle down the scope to increase the chances of this getting a proper review.

Is the cancellation part really necessary for an MVP? Could it be a follow-up PR?

davidhewitt: I would also prefer cancellation to be split out 👍 (as implied but not insisted upon in a previous comment)

src/coroutine.rs
davidhewitt: It would be interesting to explore, and if we find that this design makes the most sense but requires nightly, we can put it behind our `nightly` feature and use it as evidence for upstream to explore for stabilisation!

src/coroutine/waker.rs
davidhewitt: Yep thanks 👍

pyo3-macros-backend/src/utils.rs
wyfo: Cancellation is not necessary in the MVP, I will revert it. Sorry for spreading myself too thin, I should use pyo3-async to make these kind of design attempts (that's its primary goal)

wyfo:
I've tested this implementation in a real project using [rdkafka](https://github.com/fede1024/rust-rdkafka). And I've encountered a deadlock.

I ran into this situation because I had two unsynchronized mutexes, one being the GIL, and the second one in rdkafka (the internal queue lock).
The pattern is pretty common:
- Thread T1 acquires mutex M1 (asyncio thread run with the GIL)
- Thread T2 acquires mutex M2 (kafka background thread lock its internal queue)
- Thread T2 tries to acquire mutex M1 and wait (kafka backgound thread try to acquire the GIL to wakeup a coroutine)
- Thread T1 tries to acquire mutex M2 and wait (asyncio thread try to pop from the internal queue)
- deadlock

The solution was to release the GIL in the coroutine future (I've put back the reverted `allow_thread` feature in a local branch of my fork). Otherwise, it works really fine.

Also, I've thought about other waker optimizations (avoid acquiring the GIL or calling `call_soon_threadsafe`), but they will wait this PR to be merged.

adamreichold:
> The solution was to release the GIL in the coroutine future (I've put back the reverted allow_thread feature in a local branch of my fork). Otherwise, it works really fine.

From your description, it sounds like the best user experience here might be to always release the GIL before polling, i.e. unconditionally instead of as a separate feature/entry point. Would this have any downsides besides having to move the calls to `.into_py(py)` elsewhere?

wyfo:
> From your description, it sounds like the best user experience here might be to always release the GIL before polling, i.e. unconditionally instead of as a separate feature/entry point.

Releasing the GIL could have performance impact if another thread prevent you to acquire it back immediately. For small future, e.g. `tokio::sync::mpsc::Sender::send`, I don't think it's worth.
Anyway, the issue is quite the same as with sync functions, if you do a lot of things and/or call arbitrary C code, you should release the GIL (I'm not an experienced binary extension programmer, it should have been a reflex).
So I prefer to be consistent with sync functions, and not release the GIL by default, but we have to provide a simple way to have it released for polling, like a `#[pyo3(allow_threads)]` (this attribute could also decorate sync functions).

> Would this have any downsides besides having to move the calls to `.into_py(py)` elsewhere?

There is no need to move `.into_py(py)`, because the GIL should be released in the wrapped future, not in the wrapper, see https://github.com/PyO3/pyo3/pull/3540/commits/790cd97edd2f4075a40911688f029851078c0dad. But again, I prefer to be consistent with sync functions.

src/coroutine/waker.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+struct LoopAndFuture {
+    event_loop: PyObject,
+    future: PyObject,
+}
+
+impl LoopAndFuture {
+    fn new(py: Python<'_>) -> PyResult<Self> {
+        static GET_RUNNING_LOOP: GILOnceCell<PyObject> = GILOnceCell::new();
```
adamreichold, line 53: I don't think caching the running loop and storing within each future is useful. If we assume that loop never changes, we can just access `GET_RUNNING_LOOP` from e.g. within `LoopAndFuture::set_result` instead of storing the loop within each instance.

Additionally I would suggest using a single API cache, e.g.

```rust
struct AsyncioApi {
  create_future: PyObject,
  call_soon_threadsafe: PyObject,
  is_closed: PyObject,
  release_waiter: PyCFunction,
}

impl AsyncioApi {
  fn init(py: Python) -> PyResult<Self> {
    let asyncio = py.import("asyncio")?;
    let loop = asyncio.getattr("get_running_loop")?.into();
    let create_future = loop.getattr("create_future")?.into();
    let call_soon_threadsafe = loop.getattr("call_soon_threadsafe")?.into();
    let is_closed = loop.getattr("is_closed")?.into();
    let release_waiter = wrap_pyfunction!(release_waiter, py);

    Ok(Self { .. })
  }
}

static ASYNCIO_API: GILOnceCell<AsyncioApi> = GILOnceCell::new();

impl AsyncioFuture {
  fn new(py: Python) -> PyResult<Self> {
     let api = ASYNCIO_API.get_or_try_init(py, AsyncioApi::init)?;
     let future = api.create_future.call0(py)?; 

     Ok(Self { future })
  }
}
```

davidhewitt: I'm not convinced we should make the assumption the loop never changes, at least there will be one event loop per thread and I'm sure there must be software out there which has multiple event loops per thread for some arcane reason.

wyfo: I don't assume the loop will never change, I assume it can change (thanks to you @adamreichold by the way). That's why I don't think we can use a cache with event loop method. Also, you can have different event loop, one thread with asyncio, another with uvloop, everything is possible.

`LoopAndFuture::set_result` is called in `Waker::wake`, so it can be outside of asyncio thread, that's why `asyncio.get_running_loop` cannot be called here, it must be called in coroutine context (that's the case for `LoopAndFuture::new`).

Storing the loop is not necessary, it can be retrieved by `asyncio.Future.get_loop`, it's just a shortcut to avoid this method call, as we already have the loop object as it's used to create the future. 
I let you decide whether we keep it or call `get_loop`, it's the classical performance vs memory dilemma, and I've no opinion about it.

adamreichold: Ah sorry, I just misunderstood the code: You cache the method handle for `get_running_loop`, not its return value. Sorry for the noise.

adamreichold: (I also think caching the loop object itself is fine.)

src/coroutine/waker.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/// Call `future.set_result` if the future is not done.
+///
+/// Future can be cancelled by the event loop before being waken.
+/// See <https://github.com/python/cpython/blob/main/Lib/asyncio/tasks.py#L452C5-L452C5>
+#[pyfunction(crate = "crate")]
+fn release_waiter(future: &PyAny) -> PyResult<()> {
+    if !future.call_method0("done")?.extract::<bool>()? {
+        future.call_method1("set_result", (future.py().None(),))?;
```
adamreichold, line 93: Should `"done"` and `"set_result"` be interned as well? Isn't this the hot path, i.e. in contrast to `"is_closed"`?

adamreichold (approved):
This looks good to me now. I think we should not block this on the iterator change.

src/coroutine/waker.rs
wyfo: Again, I don't think you can cache methods. The future is not guaranteed to be an instance of `asyncio.Future`, as stated in the [documentation](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_future)

adamreichold: I just meant to do

```rust
future.call_method0(intern!(py, "done"))
```

to intern the Python string for method name as you already do above.

wyfo: Sorry, I misundestood, and you're right, I will fix that.

davidhewitt:
👍 @wyfo what's your thoughts on the readiness of this to merge? I am reasonably sure this is the right form for a basic future, though I think there are future questions around how to support other async runtimes on the Python side. (Either by adding a user API to tell pyo3 how to create wakers for them, or by automatically propagating blocking waits from Python via our internal waker.)

I'd quite like to convince myself that this can fit with pyo3-asyncio before merge, which I hope to do either this coming Friday or the one after.

davidhewitt:
Thanks and no worries, unfortunately due to various personal constraints I didn't manage to go further in the last two Fridays, and I'm playing catch up with a lot of stuff now so I suspect I'm unlikely to have a play with this and `pyo3-asyncio` on the next few Fridays either.

Perhaps let's get this PR into a mergeable state and then add future PRs for the more complicated extensions. I think that there are just two key steps remaining for this PR:
- Let's add a basic hint to `async` support somewhere in the guide. I suspect this either in `guide/src/function.md` or a subpage of that section.
- Given your intention to modify `Coroutine::new`, can we keep the constructor `pub(crate)` just for now with a TODO to make it public (for lifetime capture reasons as understood, this will need to be public when we're happy with the design).

Looking at the changes from your other comment, here's my feelings about where they should go:

> * coroutine `__name__`/`__qualname__`
> * warning when non-awaited coroutine are dropped (same as Python `async def` coroutine)

These two sound like they make sense to add to this PR as relatively obvious small refinements? Some context on the need for `__name__` / `__qualname__` might also be helpful.

> * generic waker with multiple async backend support (asyncio, trio and anyio (with sniffio))

Maybe worth an individual follow-up PR?

> * support for `&self`/`&mut self` in `#[pymethods]`

Not sure how much code was added for this, what was the original problem with it?

> * `#[pyo3(allow_threads)]` for `async fn` (and regular `fn` for coherence) with a generic parameter in `Coroutine::new`

I'm wondering if we can lean into 3.13 upcoming nogil support and avoid the need for this complexity? Definitely I'd like this as a separate follow-up PR given it also adds a feature to `pyfunction`.

> * cancellation support with `CancelHandle` and `#[pyo3(cancel_handle = "param")]` (and a generic `throw_callback` parameter in `Coroutine::new` if needed)

👍 I think this would be the right way to do this and seems useful to land in this basic implementation.

> * Rust-awaitable `PyFuture` as a wrapper around the result of `__await__`, can only be awaited in the context of a future polled by a coroutine and don't support `select!` (panic otherwise) _no nightly required_
> * like Python coroutines, PyO3 coroutines delegate `send`/`throw`/`close` call to the currently awaited `PyFuture` if there is one

These two are fantastic, especially given that allows us to avoid overheads, I suppose it's a performance improvement so let's land in a follow-up PR?

wyfo:
> Let's add a basic hint to async support somewhere in the guide. I suspect this either in guide/src/function.md or a subpage of that section.

Here are the things to document about asynchronous function support:
- the `Send + Sync + 'static` constraint for parameters/return types.
- the exception to the `'static` constraint for `&self`/`&mut self`
- `#[pyo3(cancel_handle = "param")]`
- `#[pyo3(allow_threads)]`, but it also concerns regular function, so it should be documented with other function options
- `__name__`/`__qualname__`, but as it's not mentioned for functions and classes, I don't think it's worth to mention it for coroutines
- the `Coroutine` type with it's full constructor
- the different Python runtime support
- `PyFuture` and its limitations (only with coroutine + no `select!`)

Instead of dispatching them in the documentation (one subpackage in function section, one section for `PyFuture`, and one note in classes section for `&self/&mut self`), I prefer having one section dedicated to async support, between `7. Calling Python from Rust` and `8. GIL, mutability and object types`. Something like `8. async fn and coroutine`, with a subsection `await Python awaitable` for `PyFuture`.

I have two questions still:
- Should we keep *python-asyncio* section in the documentation? As I've already said, I honestly think this crate will be kind of deprecated with the release of `Coroutine`/`PyFuture`. Should we for example add a note at the top of the page to inform about the current development of async support and the future deprecation, and remove the section when the development will be finalized?
- Should we mention the `async_support` branch of my fork, where all the features (cancellation, `PyFuture`, etc.) are already implemented (and tested), in order to allow people to test it in advance, and give feedback about design and performance? (I already do this in a project that should normally be open-sourced soon) 
On my side, I plan to write a Reddit post about this branch to invite people to try it, of course if you don't mind; I had some interesting feedbacks when I published *pyo3-async* before, and I want to renew the experience, particularly to hear about `PyFuture` performances.

> Given your intention to modify Coroutine::new, can we keep the constructor pub(crate) just for now with a TODO to make it public (for lifetime capture reasons as understood, this will need to be public when we're happy with the design).

That's a good idea, and I agree with you. We can also think about the use of builder pattern, to give us more flexibility for future modification. I'm not a fan of this solution, but this a possibility that should be addressed.

> These two sound like they make sense to add to this PR as relatively obvious small refinements? Some context on the need for __name__ / __qualname__ might also be helpful.

I prefer to not touch this PR again. It can go in a small additional PR. I've added `__name__`/`__qualname__` to be used in the "not awaited" warning. But the names is also used in `asyncio.Task.__repr__` for example, see https://github.com/python/cpython/blob/607b5e30c67bad35b90240d9ac176131e51423a5/Lib/asyncio/coroutines.py#L51-L63 (as written in the comment, coroutine should always be named). Without it, you get `<Coroutine without __name__>'`, and that's not really user friendly (I've read it a lot in my first tests with `asyncio in debug mode).

> Not sure how much code was added for this, what was the original problem with it?

The issue with `&self`/`&mut self` is that the future wrapped by the coroutine must be `'static`, so you cannot normally use a by-reference receiver. That's not convenient for methods, because you have to pass `self_: Py<Self>` and use `Python::with_gil(|gil| self_.borrow().my_method(...))` every time you want to use `&self`, as `PyRef` is not `Send`, so you cannot store it in a variable when there is suspension point (i.e. `.await`).
To make it possible, I've hacked the lifetime to capture the borrowed ref as `'static`, but making it safe by releasing the borrow in a guard captured with it.  You can see the code here `https://github.com/wyfo/pyo3/blob/27b0a9e984c5cdaf6fc956b6753f40e6b070f3a8/src/impl_/coroutine.rs#L38`. (I had also to prevent using `&'static self` in methods, because with this modification, it would compile, but would be completely unsound)
Actually, there is a design point to be discussed here: should the borrow be done before the coroutine is instantiated, or the first time the coroutine is polled? In my implementation, I've chosen the first option, but I don't have a strong opinion on that question for now. Also, it could be implemented in a different way than this transmuted `'static` capture, for example by retrieving the reference directly from the `RefGuard`, but as I knew there would be some discussion about the design, I've chosen the simplest implementation (only `release_ref`/`release_mut` to add to `PyCell`).

> I'm wondering if we can lean into 3.13 upcoming nogil support and avoid the need for this complexity? Definitely I'd like this as a separate follow-up PR given it also adds a feature to pyfunction.

Currently, in the wrapped future of a coroutine, it's not possible to `await` something with the `GIL released`, because you cannot await inside the synchronous closure of `Python::allow_threads`. It could be possible to add an `async fn allow_thread_async<F, T>(self, f: F) -> T where F: Ungil + Future<Output=T>, T: Ungil` method to `Python`, but having the `allow_threads` decorator is more ergonomic – y the way, I think we could also add this `Python::allow_thread_async`.
Actually, `#[pyo3(allow_threads)] allows to not tie your code to the use of PyO3, and that can be very important. In fact, in my project (the one [mentioned above](https://github.com/PyO3/pyo3/pull/3540#issuecomment-1789623562) where I encountered a deadlock), this is necessary, because my algorithm works with a generic runtime trait (with `spawn` and `sleep` methods), and I have two runtime implementations, `Tokio` and `Asyncio`. It allows me to use the same code for the Rust library and the Python wrapper (I use the algorithm both in Rust and Python projects). So I can't use `allow_thread_async` around each `.await`.
I have not look yet at the nogil thing, but I need this feature now :smile: 

> 👍 I think this would be the right way to do this and seems useful to land in this basic implementation.

I prefer to close this PR as soon as possible and add it in the following one.

> These two are fantastic, especially given that allows us to avoid overheads, I suppose it's a performance improvement so let's land in a follow-up PR?

It's indeed a massive performance improvement over the current "spawn a task and register a callback" workflow. It's also a dirty hack of the Rust async implementation, but worthwile, and without `unsafe`. 
I had mentioned earlier this [nightly feature](https://doc.rust-lang.org/std/task/struct.Waker.html#method.as_raw), but because I use a generic waker depending on the Python runtime, I cannot really use that feature because I would have one vtable per waker  implementation. So it would require to box the waker implementation, but I think the performance cost would be greater than my thread-local hack. This is something that should be tested.


So here is how I see the next steps:
- Add the missing documentation, make `Coroutine::new` private and merge the PR. Also answer the question about `async_support` branch.
- I will have to retro-engineer my branch to split it in independent PR. It might not be easy for some parts, but that's too bad for me, I should have work in a more structured way instead of writing everything at the same time :sweat_smile: I will have to order them to avoid conflicts. This order could be as following
- `__name__`/`__qualname__` + not-awaited warning
- `#[pyo3(cancel_handle = "param")]`
- `&self`/`&mut self`
- `#[pyo3(allow_threads)]`
- `PyFutures`
- generic waker with multiple Python runtime support; for this one I have an issue about allowing the user to define a default runtime at the crate level, and I'm asking your help on it (I will open the discussion in the dedicated PR).
- Expose `Coroutine` constructor; that will mark the achievement of the design.

Tell me if you want to change the order. I will open all the PR at the same time, with each one based on the previous one in the defined order. This way, you will be able to review all the PR independently, as long as they are merged in order.

davidhewitt:
> Instead of dispatching them in the documentation (one subpackage in function section, one section for PyFuture, and one note in classes section for &self/&mut self), I prefer having one section dedicated to async support, between 7. Calling Python from Rust and 8. GIL, mutability and object types. Something like 8. async fn and coroutine, with a subsection await Python awaitable for PyFuture.

A complete section makes sense to me; we can link to it from various places where we find we need to. I think in this PR the documentation does not need to be complete for all features, we should just document what the user can currently do after this PR, and link to the tracking issue with the future extensions.

> Should we keep python-asyncio section in the documentation? 

For now I think it's good enough to add a note at the top that new async support is landing in PyO3 and that the `pyo3-asyncio` functionality may be outdated / change in future. I'm not yet convinced that no feature of `pyo3-asyncio` will be useful after we have a base integration in PyO3, and also until we release, `pyo3-asyncio` continues to be the most commonly used solution.

> Should we mention the async_support branch of my fork

If you would like it in the documentation I'm ok with it going there in the short term, although I think we should remove that link before we publish 0.21. One option is to just keep the link in the tracking issue, and in the places where we point readers to the tracking issue we can say that the issue also has details of a prototype implementation if they want to test it.

> So here is how I see the next steps:

I'm happy with that order, many thanks 👍

davidhewitt:
> Actually, there is a design point to be discussed here: should the borrow be done before the coroutine is instantiated, or the first time the coroutine is polled? 

If it's ok with you, I'm going to punt on this one for now. When you open the PR to make this change, let's come back to this one.

wyfo:
I've hidden the constructor and added some documentation. I'm not a native English speaker, so feel free to correct and suggest.

I have not put the link to the full POC branch (it's still on the issue); a Reddit post should be enough to attract some beta testers.

I will try to open the next PR soon!

wyfo:
I've still added a last [commit](https://github.com/PyO3/pyo3/pull/3540/commits/6ab305bf056becbb8b06e7d3ea3985acbf9f3cac) to this PR with a small improvement I had forgotten in my previous list: panic handling while polling the future, in order to properly close the coroutine (not closing it led to surprising side effect when the coroutine was dropped with the not-awaited warning feature)

wyfo:
I need to update the code because of #3578

guide/src/SUMMARY.md
```diff
@@ -19,6 +19,9 @@
   - [Conversion traits](conversions/traits.md)]
 - [Python exceptions](exception.md)
 - [Calling Python from Rust](python_from_rust.md)
+- [Using `async` and `await`](async-await.md)
+  - [Awaiting Python awaitables]()
+  - [Other Python async runtimes]()
```
davidhewitt, line 24: I understand why you added these but I'd rather not have the dead links in the guide until we've definitely added support for these.

```suggestion
```

guide/src/async-await.md
davidhewitt, line 1: Please add this page to the `doctests` in `lib.rs` (and the examples may need to have some adjustments to compile).

Other than that, this doc looks good to me 👍

davidhewitt (commented):
Thanks very much, new docs look good. Just two final points to finish off.

Also, there's a lot of commits in this PR history now. We like to try to keep the history tidy, but GH merge queue sadly doesn't give us the option to squash. Can you please squash manually into one commit when you fixup the docs? (Or several commits if desired.)

guide/src/SUMMARY.md
wyfo: They are not deadlinks, they are draft chapters as stated in the mdBook documentation. They are equivalent to the cancellation section with *To be implemented*. And because I've already implemented all of these features, I prefer to showcase the titles (especially `Awaiting Python awaitables`), because I know the PR will come.
I will still remove them as asked. Tell me if I've convinced you to put them back :)

davidhewitt (approved):
Yep I was aware about the draft chapters, but I think just slightly better to not have them at all until we're ready to have them.

Thanks for the many revisions here, I think many users will be very excited to have this!

wyfo:
I didn't know about *emscripten*, but it seems that *asyncio* is [not supported](https://github.com/python/cpython/blob/main/Tools/wasm/README.md#network-stack) on this target.

Should I add some `#[cfg(not(target = "wasm32-unknown-emscripten"))]`  in the code?

davidhewitt:
Hmm good question. I believe that pyodide might support asyncio, in which case we could potentially benefit from testing that.

For now how about marking the tests which fail with `#[cfg_attr(target = "wasm32-unknown-emscripten", ignore)]`, and let's make an issue to investigate / solve that in a follow-up?

tests/test_coroutine.rs
```diff
@@ -0,0 +1,98 @@
+#![cfg(feature = "macros")]
+#![cfg(not(target_arch = "wasm32"))]
```
wyfo, line 2: I've added this line to deal with *emscripten* for now. Is it ok for you? Can I let you open the issue? I don't really know anything about wasm.

davidhewitt:
It's not my area of expertise also, though I've had a couple of rounds of understanding it for PyO3's sake. I've added it to #2582

davidhewitt:
Hmm, guide entry isn't MSRV compatible. I'm sort of okay with that, however that does create potential distribution problems for users who follow the guide. Is there a futures alternative we can pull in as a dev dependency?

wyfo:
Actually, the failing example is the workaround for `allow_threads`, and I don't really want to use `futures::pin_mut!` in it because nobody is using it today since `std::pin::pin!` has been released.
I will mark the workaround as `ignore` for now. Anyway, it's a workaround waiting other features to be implemented, so it should disappear soon.

wyfo:
:champagne:

PR #3546, opened 2023-10-24
ci: use older hashbrown and indexmap for MSRV

davidhewitt:
Ok, I think this is the solution which will finally work to resolve the `ahash` break.

If it's good, I'll merge it and push the other PRs through which are blocked on merging by this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


davidhewitt:
I've added a commit to allow `nightly` jobs to fail, which should let this merge. It's happened a few times that `nightly` breaks with upstream changes, so I think this is justified. I'll fix the nightly break in a follow up PR.

PR #3551, opened 2023-10-26
ci: switch from black to ruff

davidhewitt:
Everyone is celebrating the new `ruff format` command, and also by adopting `ruff` we get some linting on our Python examples, which seems like a good thing.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -66,22 +66,17 @@ def coverage(session: nox.Session) -> None:
     )
 
 
-@nox.session
-def fmt(session: nox.Session):
-    fmt_rust(session)
-    fmt_py(session)
-
-
-@nox.session(name="fmt-rust", venv_backend="none")
-def fmt_rust(session: nox.Session):
+@nox.session(venv_backend="none")
+def rustfmt(session: nox.Session):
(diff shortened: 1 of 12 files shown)


noxfile.py
```diff
@@ -66,22 +66,17 @@ def coverage(session: nox.Session) -> None:
...
 
-@nox.session(name="fmt-py")
-def fmt_py(session: nox.Session):
-    session.install("black==22.3.0")
-    _run(session, "black", ".", "--check")
+@nox.session(name="ruff")
+def ruff(session: nox.Session):
+    session.install("ruff")
+    _run(session, "ruff", "format", ".", "--check")
+    session.run("ruff", ".")
```
adamreichold, line 79: I understand it is the default, but could we explicitly run `ruff check .` here to make the intent clearer?

Also why not use the `_run` help to put this into a collapsible group in the GHA logs?

davidhewitt: Sure thing, will add `check` later. Similarly `_run` is just missing because I copied this from setuptools-rust, will fix later too.

adamreichold (approved):

PR #3552, opened 2023-10-26
docs: Example of dynamic return type in the "Python classes" guide

daemontus:
This PR is based on the discussion for [this question](https://github.com/PyO3/pyo3/issues/1637). It just adds a single method into the inheritance section of the class guide. 

The purpose of this new example is to show how a sub-type can be instantiated dynamically within a method other than `#[new]`. In other words, how to "instantiate" `PyClassInitializer` into an actual Python object from Rust.

If you know a better way to do this, feel free to suggest an alternative example. But the reactions in the question discussion seem to indicate that the answer is not entirely obvious, hence I considered it worth including.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -337,10 +337,27 @@ impl SubSubClass {
         let super_ = self_.into_super(); // Get PyRef<'_, SubClass>
         SubClass::method2(super_).map(|x| x * v)
     }
+
+    #[staticmethod]
+    fn dynamic_type(py: Python<'_>, val: usize) -> PyResult<PyObject> {
+        let base = PyClassInitializer::from(BaseClass::new());
+        let sub = base.add_subclass(SubClass { val2: val });
+        if val % 2 == 0 {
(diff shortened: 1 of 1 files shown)


guide/src/class.md
```diff
@@ -337,10 +337,27 @@ impl SubSubClass {
         let super_ = self_.into_super(); // Get PyRef<'_, SubClass>
         SubClass::method2(super_).map(|x| x * v)
     }
+
+    #[staticmethod]
+    fn dynamic_type(py: Python<'_>, val: usize) -> PyResult<PyObject> {
```
mejrs, line 342: What about naming this something like...
```suggestion
    fn factory_method(py: Python<'_>, val: usize) -> PyResult<PyObject> {
```
...because I think that when people want something like this they'll search for the "factory" pattern.

daemontus:
Sure, why not :)

mejrs (approved):
Thanks!

PR #3556, opened 2023-10-29
feat: add `take` and `into_inner` methods to `GILOnceCell`

wyfo:
`take` is used in #3540

Now, `GILOnceCell` has the same methods as standard [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html), but some traits implementation are still missing, like `Default`.

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -167,6 +167,20 @@ impl<T> GILOnceCell<T> {
         *inner = Some(value);
         Ok(())
     }
+
+    /// Takes the value out of the cell, moving it back to an uninitialized state.
+    ///
+    /// Has no effect and returns None if the cell has not yet been written.
+    pub fn take(&mut self) -> Option<T> {
+        self.0.get_mut().take()
+    }
+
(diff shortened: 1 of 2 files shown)


src/sync.rs
```diff
@@ -167,6 +167,20 @@ impl<T> GILOnceCell<T> {
         *inner = Some(value);
         Ok(())
     }
+
+    /// Takes the value out of the cell, moving it back to an uninitialized state.
+    ///
+    /// Has no effect and returns None if the cell has not yet been written.
+    pub fn take(&mut self) -> Option<T> {
+        self.0.get_mut().take()
```
adamreichold, line 175: Could we avoid the unsafe block `GILOnceCell::get_mut` in the same manner, i.e. `self.0.get_mut().get_mut()`?

adamreichold (approved):
One minor comment whether the existing code could be improved.

src/sync.rs
wyfo: Done.

PR #3559, opened 2023-10-30
add link to user guide to cross compile error message

davidhewitt:
Just thought in passing that it might be a nice touch for users to have this.

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -1361,9 +1361,13 @@ fn default_cross_compile(cross_compile_config: &CrossCompileConfig) -> Result<In
     let version = cross_compile_config
         .version
         .or_else(get_abi3_version)
-        .ok_or(
-            "PYO3_CROSS_PYTHON_VERSION or an abi3-py3* feature must be specified \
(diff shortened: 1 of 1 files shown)


pyo3-build-config/src/impl_.rs
```diff
@@ -1361,9 +1361,13 @@ fn default_cross_compile(cross_compile_config: &CrossCompileConfig) -> Result<In
     let version = cross_compile_config
         .version
         .or_else(get_abi3_version)
-        .ok_or(
-            "PYO3_CROSS_PYTHON_VERSION or an abi3-py3* feature must be specified \
-            when cross-compiling and PYO3_CROSS_LIB_DIR is not set.",
+        .ok_or_else(||
+            format!(
```
mejrs, line 1365: you dont have to change this, but you could use `concat!` to avoid creating the string 🤔

mejrs (approved):

pyo3-build-config/src/impl_.rs
davidhewitt: I did consider `concat` but slightly preferred how the code read with `format`. I'll adjust tomorrow unless someone clicks merge before then 😁

Issue #3218, opened 2023-06-08
Cannot stop rust program while running python code


ktjd123:
### Bug Description

When I run python code in fast loop from rust and want to stop the program, python takes interrupt so rust program won't exit.

### Steps to Reproduce

Run python code in Rust

### Your operating system and version

tested in mac, linux

### Your Python version (`python --version`)

Python 3.11.3

### Your Rust version (`rustc --version`)

rustc 1.70.0 (90c541806 2023-05-31)

### Your PyO3 version

0.19.0

### How did you install python? Did you use a virtualenv?

mac with brew

and apt for linux

### Additional Info

Thank you.

adamreichold:
Could you share some code showing what exactly you are doing?

For example, doesn't Python raise an exception when the signal is received? Where is that exception handled?

ktjd123:
Sorry for few information.

I think the CTRL-C is captured at python side, and thought That shouldn't happen since It's blocking input to main process, But I'll investigate more and re-open issue.

Thank you for reply

birkenfeld:
You can disable Python's `SIGINT` handler with
```
signal.signal(signal.SIGINT, signal.SIG_DFL)
```

ktjd123:
@birkenfeld Thank you for your kind help :)

PR #3560, opened 2023-11-02
docs: Include section on how to disable signals in python

Jgfrausing:
Occasionally, ctrl_c would be caught by Python and the rust code would continue to run. We have not been able to create a working example, but it seems that pandas or numpy, could affect if interrupts are caught in Rust or in Python.

Inspired by https://github.com/PyO3/pyo3/issues/3218

### DIFF
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -487,5 +487,26 @@ class House(object):
 }
 ```
 
+## Handling system signals/interrupts (Ctrl-C)
+
+The best way to handle system signals when running Rust code is to periodically call `Python::check_signals` to handle any signals captured by Python's signal handler. See also [the FAQ entry](./faq.md#ctrl-c-doesnt-do-anything-while-my-rust-code-is-executing).
+
+Alternatively, set Python's `signal` module to take the default action for a signal:
+
+```rust
+use pyo3::prelude::*;
+
+# fn main() -> PyResult<()> {
+Python::with_gil(|py| -> PyResult<()> {
+    let signal = py.import("signal")?;
+    // Set SIGINT to have the default action
+    signal
+        .getattr("signal")?
+        .call1((signal.getattr("SIGINT")?, signal.getattr("SIG_DFL")?))?;
+    Ok(())
+})
+# }
+```
+
 
 [`PyModule::new`]: {{#PYO3_DOCS_URL}}/pyo3/types/struct.PyModule.html#method.new


guide/src/python_from_rust.md
```diff
@@ -479,5 +479,22 @@ class House(object):
 }
 ```
 
+## Handling system signals/interrupts (ctrl_c)
+To ensure, that system signals are caught in Rust, you should disable them in Python.
```
davidhewitt, line 483:
```suggestion
To handle system signals when running Rust code, you can set Python's `signal` module to take the default action for a signal:
```

guide/src/python_from_rust.md
```diff
@@ -479,5 +479,22 @@ class House(object):
...
+                r#"
+import signal
+print("[python] Ignoring system signals")
+signal.signal(signal.SIGINT, signal.SIG_DFL)
+                "#,
+                None,
+                None,
+            )
+            .unwrap();
+        });
```
davidhewitt, line 496: Maybe let's use PyO3's API to do this directly:

```suggestion
Python::with_gil(|py| {
    let signal_module = py.import("signal")?;
    // Set SIGINT to have the default action
    signal
        .getattr("signal")?
        .call1((signal.getattr("SIGINT")?, signal.getattr("SIG_DFL")?))
});
```

davidhewitt (commented):
Thanks for the PR, and sorry it took me a little while to look at it.

This seems like a reasonable addition, I've suggested some small nits. Might also be worth adding an extra link to the FAQ entry on `check_signals` for users that don't want to completely disable signal handling?

https://pyo3.rs/v0.20.0/faq#ctrl-c-doesnt-do-anything-while-my-rust-code-is-executing

guide/src/python_from_rust.md
Jgfrausing, line 496: Python::with_gil(|py| {
    let signal = py.import("signal")?;
    // Set SIGINT to have the default action
    signal
        .getattr("signal")?
        .call1((signal.getattr("SIGINT")?, signal.getattr("SIG_DFL")?))?;
  });

Jgfrausing, line 496:
```
Python::with_gil(|py| {
    let signal = py.import("signal")?;
    // Set SIGINT to have the default action
    signal
        .getattr("signal")?
        .call1((signal.getattr("SIGINT")?, signal.getattr("SIG_DFL")?))?;
  });
```

davidhewitt: Ah yes, you may need a trailing `Ok(())` inside that closure though. 👍

guide/src/python_from_rust.md
```diff
@@ -479,5 +479,17 @@ class House(object):
 }
 ```
 
+## Handling system signals/interrupts (ctrl_c)
+To handle system signals when running Rust code, you can set Python's `signal` module to take the default action for a signal:
+```rust
+Python::with_gil(|py| {
+    let signal_module = py.import("signal")?;
```
davidhewitt, line 486:
```suggestion
    let signal = py.import("signal")?;
```

davidhewitt (commented):
Think the below should fix it, also please force-push to rebase and squash and then CI should hopefully be green!

PR #3564, opened 2023-11-07
fixes #3561 -- silence new clippy warning

### LINKED ISSUE #3561: clippy nightly warns on some function argument conversions
Consider the following minimal method:

```rust
#[pyo3::prelude::pyclass]
struct Pool;

#[pyo3::pymethods]
impl Pool {
    fn acquire(slf: pyo3::Py<Self>) -> pyo3::Py<Self> {
        slf
    }
}
```

With the latest nightly, it produces the following `clippy` warning:

```
/t/t ❯❯❯ cargo +nightly clippy
warning: use of a fallible conversion when an infallible one could be used
 --> src/lib.rs:6:21
  |
6 |     fn acquire(slf: pyo3::Py<Self>) -> pyo3::Py<Self> {
  |                     ^^^^^^^^^^^^^^ help: use: `From::from`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_fallible_conversions
  = note: `#[warn(clippy::unnecessary_fallible_conversions)]` on by default

warning: `t` (lib) generated 1 warning
```

alex:


pyo3-macros-backend/src/method.rs
```diff
@@ -178,6 +178,7 @@ impl SelfType {
                             .map_err(::std::convert::Into::<_pyo3::PyErr>::into)
                             .and_then(
                                 #[allow(clippy::useless_conversion)]  // In case slf is PyCell<Self>
+                                #[allow(unknown_lints, clippy::unnecessary_fallible_conversions)]  // In case slf is Py<Self>
```
adamreichold, line 181: The `unknown_lints` is necessary for this to work on stable, right?

I am not sure we should commit to doing this until the lint hits stable as lints often change/are demoted before being released to stable.

alex: Also for any older release, I assume

davidhewitt: I wonder if we can also silence this lint by moving away from `TryFrom` to some other implementation, given that's what clippy is complaining about.

alex: I _assume_ (but didn't check in depth) that `TryFrom` is needed for other receiver types.

davidhewitt: It is, but we could replace it all with some other trait or function call

adamreichold:
What is the status now that 1.74 is released? Is stable Rust affected? If so, from my point of view, I would prefer silencing the now stable lint instead of trying to work around Clippy by using a different trait.

davidhewitt:
I think this is now in beta, so we have one more release cycle to solve it 😂

davidhewitt:
It struck me that we probably want to fix this before it hits stable, so that users don't have to temporarily silence clippy waiting for a PyO3 upgrade.

pyo3-macros-backend/src/method.rs
davidhewitt: In the end in #3587 we stuck with `From`, so I think it's correct to stick to `TryFrom`

alex:
Yes, I agree with that. Obviously those of us who test on nightly got this a bit early :-)

davidhewitt:
The one alternative which I'd think we could consider is to use feature detection to only emit this on 1.75 and up. I'd slightly prefer that given we already have the machinery in our `build.rs` for this and then over time we can delete the feature detection, rather than leave it in the `unknown_lints` allow for some indefinite time.

alex:
Hmm, started looking at doing that, but `pyo3-macros-backend` (where this code is), doesn't depend on `pyo3-build-config` ATM. Is it better to add a dependency, or just do it this way?

Could also split the baby: Add a `cfg` to it, and then put it in a comment so we don't forget about it, even though it's not actually used programatically.

davidhewitt:
Ah good point. Maybe just add a comment `// TODO: remove unknown_lints allow when MSRV above 1.75` or similar, in that case?

alex:
Done

pyo3-build-config/src/lib.rs
```diff
@@ -153,6 +153,10 @@ pub fn print_feature_cfgs() {
     if rustc_minor_version >= 74 {
         println!("cargo:rustc-cfg=invalid_from_utf8_lint");
     }
+
+    if rustc_minor_version >= 75 {
+        println!("cargo:rustc-cfg=clippy_unnecessary_fallible_conversions_lint");
+    }
```
davidhewitt, line 159: I think this bit can be reverted, otherwise please merge 👍

davidhewitt (approved):

pyo3-build-config/src/lib.rs
alex: Done

PR #3573, opened 2023-11-14
ci: use trybuild branch which strips rust-src lines

davidhewitt:
This PR adds a temporary patch to our `trybuild` dependency which strips out rust standard library src lines from error messages. This removes our need to install `rust-src` in CI, hopefully resolving our `rustup` issue which has blocked CI.

See https://github.com/dtolnay/trybuild/pull/247

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,3 +127,7 @@ members = [
 no-default-features = true
 features = ["macros", "num-bigint", "num-complex", "hashbrown", "serde", "multiple-pymethods", "indexmap", "eyre", "chrono", "rust_decimal"]
 rustdoc-args = ["--cfg", "docsrs"]
+
+# workaround for issue installing rust-src on GitHub actions, see https://github.com/dtolnay/trybuild/pull/247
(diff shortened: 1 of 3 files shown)


davidhewitt:
How do fellow contributors feel about depending on a patch of try build for the moment, to unblock CI? If the trybuild PR is rejected we can switch to a different solution, though I don't think any are as tidy as this one.

adamreichold:
> How do fellow contributors feel about depending on a patch of try build for the moment, to unblock CI? If the trybuild PR is rejected we can switch to a different solution, though I don't think any are as tidy as this one.

Not great but still better than being unable to merge anything.

As a wild guess, did you try to delete `rust-toolchain.toml` so that we only install the component upon explicit request? IIRC, I saw those errors earlier and resolved them back then by explicitly installing the component upfront instead when automatically request via that file.

davidhewitt:
I'll push a PR and see what happens...

davidhewitt:
Well, it looks like #3575 will go green, so let's start with that.

Issue #3516, opened 2023-10-15
PyTryFrom conflicts with TryFrom


davidhewitt:
Since edition 2021 and `TryFrom` being in the prelude, the ergonomics of `PyTryFrom` have gotten significantly worse due to need to disambiguate.

Personally I'd like us to remove `PyTryFrom` as I think we can make the conversions work without it. Maybe this is something to deprecate as part of 0.21.

davidhewitt:
I think how this works is we can add `PyTypeCheck` which has the same `is_type_of` and `is_exact_type_of` methods from `PyTypeInfo`, and a blanket `impl<T: PyTypeInfo> PyTypeCheck for T`.

Then we change `PyAny::downcast[_exact]` to use `PyTypeCheck` instead.

Possibly we do this only for the new `PyAnyMethods` trait and leave the `&PyAny` method unchanged as we're about to deprecate and/or remove it anyway.

The reason for `PyTypeCheck` needing to be separate is for types like `PyIterator` which don't have a single type info, but can still be checked for and cast.

----

There might be a better design, alternative suggestions welcome!

I'll type this in eventually, if anyone else is interested in doing this before I get to it please just ping me if you want more details.

PR #3577, opened 2023-11-15
Implement `PyTypeInfo` for `PyEllipsis`, `PyNone`, and `PyNotImplemented`

davidhewitt:
Playing around with options to solve #3516, I noticed that we can implement `PyTypeInfo` without complication for the three singletons `Ellipsis`, `None`, and `PyNotImplemented`. So, here's the PR 😄

### DIFF
--- a/src/types/ellipsis.rs
+++ b/src/types/ellipsis.rs
@@ -1,4 +1,4 @@
-use crate::{ffi, PyAny, PyDowncastError, PyTryFrom, Python};
+use crate::{ffi, PyAny, PyTypeInfo, Python};
 
 /// Represents the Python `Ellipsis` object.
 #[repr(transparent)]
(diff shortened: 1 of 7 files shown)


src/types/ellipsis.rs
```diff
@@ -15,24 +15,26 @@ impl PyEllipsis {
...
+        // ellipsis is not usable as a base type
+        Self::is_exact_type_of(object)
     }
 
-    unsafe fn try_from_unchecked<V: Into<&'v PyAny>>(value: V) -> &'v Self {
-        let ptr = value.into() as *const _ as *const PyEllipsis;
-        &*ptr
+    #[inline]
+    fn is_exact_type_of(object: &PyAny) -> bool {
+        object.is(PyEllipsis::get(object.py()))
```
adamreichold, line 37: Nit: `Self::get` instead of `PyEllipsis::get` for consistency.

adamreichold: But more importantly why not use `object.is_ellipsis()` similar to `object.is_none()` below?

src/types/ellipsis.rs
```diff
@@ -15,24 +15,26 @@ impl PyEllipsis {
...
-    fn try_from<V: Into<&'v PyAny>>(value: V) -> Result<&'v Self, crate::PyDowncastError<'v>> {
-        let value: &PyAny = value.into();
-        if unsafe { ffi::Py_Ellipsis() == value.as_ptr() } {
-            return unsafe { Ok(value.downcast_unchecked()) };
-        }
-        Err(PyDowncastError::new(value, "ellipsis"))
+unsafe impl PyTypeInfo for PyEllipsis {
+    const NAME: &'static str = "ellipsis";
+
+    const MODULE: Option<&'static str> = None;
```
adamreichold, line 21: `help(type(...))` reports "builtins" as the module. Should we do this as well here or is `None` effectively the same?

src/types/notimplemented.rs
```diff
@@ -15,24 +15,26 @@ impl PyNotImplemented {
...
+        // NoneType is not usable as a base type
+        Self::is_exact_type_of(object)
     }
 
-    unsafe fn try_from_unchecked<V: Into<&'v PyAny>>(value: V) -> &'v Self {
-        let ptr = value.into() as *const _ as *const PyNotImplemented;
-        &*ptr
+    #[inline]
+    fn is_exact_type_of(object: &PyAny) -> bool {
+        object.is(PyNotImplemented::get(object.py()))
```
adamreichold, line 37: I think we should add `PyAny::is_not_implemented` and either use these methods for all three implementations or but the three implementations here and rewrite the methods in terms of them.

davidhewitt: I quite like `is_none()` but personally I think `is_not_implemented()` and `is_ellipsis()` are needed rarely enough that they're not worth pulling their weight.

There are also the three getter methods `py.None()`, `py.NotImplemented()` and `py.Ellipsis()`. Really these are just a slightly nicer way to write `PyNone::get(py)`. (They return `PyObject` but really they should return `&PyNone` etc in my opinion, I'll open a separate PR for that in a moment.)

What do you think about deprecating `obj.is_ellipsis()` in favour of `obj.is(py.Ellipsis())`?

davidhewitt: I decided to open #3578 specifically to change the return types. I'll come back to this PR afterwards and do the `is_ellipsis()` deprecation here as well as tidy up these implementations.

src/types/ellipsis.rs
davidhewitt: To be honest `MODULE` is only used in creation of type objects to build the `__qualname__` in pyclass types. I kinda want to deprecate this and move it to `PyClassImpl` as I think `T::type_object(py).getattr("__module__")` is the more correct way to access this if it's needed, as that's guaranteed to be right at runtime.

src/types/notimplemented.rs
```diff
@@ -15,24 +15,26 @@ impl PyNotImplemented {
...
+        unsafe { ffi::Py_TYPE(ffi::Py_NotImplemented()) }
     }
 
-    fn try_from_exact<V: Into<&'v PyAny>>(
-        value: V,
-    ) -> Result<&'v Self, crate::PyDowncastError<'v>> {
-        value.into().downcast()
+    #[inline]
+    fn is_type_of(object: &PyAny) -> bool {
+        // NoneType is not usable as a base type
```
alex, line 31: bad copy-paste :-)

src/types/notimplemented.rs
davidhewitt: Rebased to implement all three by `object.is(Self::get(py))`, and also deprecate `is_ellipsis`.

adamreichold (approved):

PR #3582, opened 2023-11-17
Full ADT support with pyclass for complex enums

### LINKED ISSUE #417: Support exporting Rust enums to Python

mkovaxx:
## Design Overview

A complex enum is an enum with any non-unit variants. For now, tuple-like variants are out of scope.

For each variant of a complex enum, PyO3 generates a hidden Rust zero-sized type (empty struct). The PyClass for each variant is a subclass of the PyClass for the enum itself. Whenever a value of the enum is passed to the Python side, it is cast to the PyClass of the specific variant. Each variant PyClass exposes getters for its fields.

The PyClass of the enum type itself has no constructor, so it cannot be instantiated from Python. Instead of a constructor, it has a class attribute corresponding to each variant, which returns the PyType of the matching variant newtype. This allows instantiating each variant newtype from Python.

## Example & Details

To make this more concrete, here's an example.

```rust
#[pyclass]
pub enum ComplexEnum {
    Int { i: i32 },
    Float { f: f64 },
    Str { s: String },
}
```

Under the hood, PyO3 generates these hidden ZSTs:

```rust
struct ComplexEnum_Int;
struct ComplexEnum_Float;
struct ComplexEnum_Str;
```

Each of these has a constructor and getter methods for its fields.

E.g. `ComplexEnum_Int` will have the following generated `impl`:
```rust
impl ComplexEnum_Int {
    fn __pymethod_constructor__(
        py: _pyo3::Python<'_>,
        i: i32,
    ) -> _pyo3::PyClassInitializer<ComplexEnum_Int> {
        let base_value = ComplexEnum::Int { i };
        _pyo3::PyClassInitializer::from(base_value).add_subclass(ComplexEnum_Int)
    }

    fn i(slf: _pyo3::PyRef<Self>) -> _pyo3::PyResult<i32> {
        match &*slf.into_super() {
            ComplexEnum::Int { i, .. } => Ok(i.clone()),
            _ => unreachable!("Wrong complex enum variant found in variant wrapper PyClass"),
        }
    }
}
```

(Omitting the generated PyClass impls for brevity.)

This allows the following operations on the Python side:
```python
# instantiation
value = ComplexEnum.Int(42)

# instance checks against variants: this will return True
isinstance(ComplexEnum.Float(3.14), ComplexEnum.Float)

# instance checks against variants: this will return False
isinstance(ComplexEnum.Float(3.14), ComplexEnum.Int)

# instance checks against the enum: this will return True
isinstance(ComplexEnum.Float(3.14), ComplexEnum)

# extracting fields
print(value.i)

# the wrapper newtype only has getters for fields actually present on the variant
# this will fail with AttributeError: 'builtins.ComplexEnum_Int' object has no attribute 'f'
print(value.f)
# this will fail with AttributeError: 'builtins.ComplexEnum_Int' object has no attribute 's'
print(value.s)

# pattern matching
match value:
    case ComplexEnum.Int(i=x):
        print(f"Int({x})")
    case ComplexEnum.Float(f=x):
        print(f"Float({x})")
    case ComplexEnum.Str(s=x):
        print(f"Str({x})")
    case _:
        print("Invalid")
```

Passing variant values between Python and Rust is also fully supported, e.g.:

```rust
#[pyfunction]
pub fn do_stuff(thing: &ComplexEnum) -> ComplexEnum {
    match thing {
        ComplexEnum::Int { i } => ComplexEnum::Str { s: format!("{i}") },
        ComplexEnum::Float { f } => ComplexEnum::Float { f: f * f },
        ComplexEnum::Str { s } => ComplexEnum::Int { i: s.len() as i32 },
    }
}
```

Such a function can be called from Python and works as expected.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -2,7 +2,7 @@
 
 PyO3 exposes a group of attributes powered by Rust's proc macro system for defining Python classes as Rust structs.
 
-The main attribute is `#[pyclass]`, which is placed upon a Rust `struct` or a fieldless `enum` (a.k.a. C-like enum) to generate a Python type for it. They will usually also have *one* `#[pymethods]`-annotated `impl` block for the struct, which is used to define Python methods and constants for the generated Python type. (If the [`multiple-pymethods`] feature is enabled, each `#[pyclass]` is allowed to have multiple `#[pymethods]` blocks.) `#[pymethods]` may also have implementations for Python magic methods such as `__str__`.
+The main attribute is `#[pyclass]`, which is placed upon a Rust `struct` or `enum` to generate a Python type for it. They will usually also have *one* `#[pymethods]`-annotated `impl` block for the struct, which is used to define Python methods and constants for the generated Python type. (If the [`multiple-pymethods`] feature is enabled, each `#[pyclass]` is allowed to have multiple `#[pymethods]` blocks.) `#[pymethods]` may also have implementations for Python magic methods such as `__str__`.
 
 This chapter will discuss the functionality and configuration these attributes offer. Below is a list of links to the relevant section of this chapter for each:
 
@@ -21,13 +21,13 @@ This chapter will discuss the functionality and configuration these attributes o
 
 ## Defining a new class
 
-To define a custom Python class, add the `#[pyclass]` attribute to a Rust struct or a fieldless enum.
+To define a custom Python class, add the `#[pyclass]` attribute to a Rust struct or enum.
 ```rust
 # #![allow(dead_code)]
 use pyo3::prelude::*;
 
 #[pyclass]
-struct Integer {
+struct MyClass {
     inner: i32,
 }
 
@@ -35,7 +35,15 @@ struct Integer {
 #[pyclass]
 struct Number(i32);
 
-// PyO3 supports custom discriminants in enums
+// PyO3 supports unit-only enums (which contain only unit variants)
+// These simple enums behave similarly to Python's enumerations (enum.Enum)
+#[pyclass]
+enum MyEnum {
+    Variant,
+    OtherVariant = 30, // PyO3 supports custom discriminants.
+}
+
+// PyO3 supports custom discriminants in unit-only enums
 #[pyclass]
 enum HttpResponse {
     Ok = 200,
@@ -44,14 +52,19 @@ enum HttpResponse {
     // ...
 }
 
+// PyO3 also supports enums with non-unit variants
+// These complex enums have sligtly different behavior from the simple enums above
+// They are meant to work with instance checks and match statement patterns
 #[pyclass]
-enum MyEnum {
-    Variant,
-    OtherVariant = 30, // PyO3 supports custom discriminants.
+enum Shape {
+    Circle { radius: f64 },
+    Rectangle { width: f64, height: f64 },
+    RegularPolygon { side_count: u32, radius: f64 },
+    Nothing { },
 }
 ```
 
-The above example generates implementations for [`PyTypeInfo`] and [`PyClass`] for `MyClass` and `MyEnum`. To see these generated implementations, refer to the [implementation details](#implementation-details) at the end of this chapter.
+The above example generates implementations for [`PyTypeInfo`] and [`PyClass`] for `MyClass`, `Number`, `MyEnum`, `HttpResponse`, and `Shape`. To see these generated implementations, refer to the [implementation details](#implementation-details) at the end of this chapter.
 
 ### Restrictions
 
@@ -964,7 +977,13 @@ Note that `text_signature` on `#[new]` is not compatible with compilation in
 
 ## #[pyclass] enums
 
-Currently PyO3 only supports fieldless enums. PyO3 adds a class attribute for each variant, so you can access them in Python without defining `#[new]`. PyO3 also provides default implementations of `__richcmp__` and `__int__`, so they can be compared using `==`:
+Enum support in PyO3 comes in two flavors, depending on what kind of variants the enum has: simple and complex.
+
+### Simple enums
+
+A simple enum (a.k.a. C-like enum) has only unit variants.
+
+PyO3 adds a class attribute for each variant, so you can access them in Python without defining `#[new]`. PyO3 also provides default implementations of `__richcmp__` and `__int__`, so they can be compared using `==`:
 
 ```rust
 # use pyo3::prelude::*;
@@ -986,7 +1005,7 @@ Python::with_gil(|py| {
 })
 ```
 
-You can also convert your enums into `int`:
+You can also convert your simple enums into `int`:
 
 ```rust
 # use pyo3::prelude::*;
@@ -1093,6 +1112,90 @@ enum BadSubclass {
 ```
 
 `#[pyclass]` enums are currently not interoperable with `IntEnum` in Python.
+
+### Complex enums
+
+An enum is complex if it has any non-unit (struct or tuple) variants.
+
+Currently PyO3 supports only struct variants in a complex enum. Support for unit and tuple variants is planned.
+
+PyO3 adds a class attribute for each variant, which may be used to construct values and in match patterns. PyO3 also provides getter methods for all fields of each variant.
+
+```rust
+# use pyo3::prelude::*;
+#[pyclass]
+enum Shape {
+    Circle { radius: f64 },
+    Rectangle { width: f64, height: f64 },
+    RegularPolygon { side_count: u32, radius: f64 },
+    Nothing { },
+}
+
+Python::with_gil(|py| {
+    let def_count_vertices = if py.version_info() >= (3, 10) { r#"
+        def count_vertices(cls, shape):
+            match shape:
+                case cls.Circle():
+                    return 0
+                case cls.Rectangle():
+                    return 4
+                case cls.RegularPolygon(side_count=n):
+                    return n
+                case cls.Nothing():
+                    return 0
+    "# } else { r#"
+        def count_vertices(cls, shape):
+            if isinstance(shape, cls.Circle):
+                return 0
+            elif isinstance(shape, cls.Rectangle):
+                return 4
+            elif isinstance(shape, cls.RegularPolygon):
+                n = shape.side_count
+                return n
+            elif isinstance(shape, cls.Nothing):
+                return 0
+    "# };
+
+    let circle = Shape::Circle { radius: 10.0 }.into_py(py);
+    let square = Shape::RegularPolygon { side_count: 4, radius: 10.0 }.into_py(py);
+    let cls = py.get_type::<Shape>();
+
+    pyo3::py_run!(py, circle square cls, &format!(r#"
+        assert isinstance(circle, cls)
+        assert isinstance(circle, cls.Circle)
+        assert circle.radius == 10.0
+
+        assert isinstance(square, cls)
+        assert isinstance(square, cls.RegularPolygon)
+        assert square.side_count == 4
+        assert square.radius == 10.0
+
+        {}
+
+        assert count_vertices(cls, circle) == 0
+        assert count_vertices(cls, square) == 4
+    "#, def_count_vertices))
+})
+```
+
+WARNING: `Py::new` and `.into_py` are currently inconsistent. Note how the constructed value is _not_ an instance of the specific variant. For this reason, constructing values is only recommended using `.into_py`.
+
+```rust
+# use pyo3::prelude::*;
+#[pyclass]
+enum MyEnum {
+    Variant { i: i32 },
+}
+
+Python::with_gil(|py| {
+    let x = Py::new(py, MyEnum::Variant { i: 42 }).unwrap();
+    let cls = py.get_type::<MyEnum>();
+    pyo3::py_run!(py, x cls, r#"
+        assert isinstance(x, cls)
+        assert not isinstance(x, cls.Variant)
+    "#)
+})
+```
 
 ## Implementation details
 
--- a/pyo3-macros-backend/src/pyclass.rs
+++ b/pyo3-macros-backend/src/pyclass.rs
@@ -7,7 +7,7 @@ use crate::attributes::{
 };
 use crate::deprecations::Deprecations;
 use crate::konst::{ConstAttributes, ConstSpec};
-use crate::method::FnSpec;
+use crate::method::{FnArg, FnSpec};
 use crate::pyimpl::{gen_py_const, PyClassMethodsType};
 use crate::pymethod::{
     impl_py_getter_def, impl_py_setter_def, MethodAndMethodDef, MethodAndSlotDef, PropertyType,
@@ -16,7 +16,7 @@ use crate::pymethod::{
 use crate::utils::{self, apply_renaming_rule, get_pyo3_crate, PythonDoc};
 use crate::PyFunctionOptions;
 use proc_macro2::{Ident, Span, TokenStream};
-use quote::quote;
+use quote::{format_ident, quote};
 use syn::ext::IdentExt;
 use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
@@ -30,6 +30,7 @@ pub enum PyClassKind {
 }
 
 /// The parsed arguments of the pyclass macro
+#[derive(Clone)]
 pub struct PyClassArgs {
     pub class_kind: PyClassKind,
     pub options: PyClassPyO3Options,
@@ -52,7 +53,7 @@ impl PyClassArgs {
     }
 }
 
-#[derive(Default)]
+#[derive(Clone, Default)]
 pub struct PyClassPyO3Options {
     pub krate: Option<CrateAttribute>,
     pub dict: Option<kw::dict>,
@@ -128,7 +129,7 @@ impl Parse for PyClassPyO3Option {
     }
 }
 
-impl PyClassPyO3Options {
+impl Parse for PyClassPyO3Options {
     fn parse(input: ParseStream<'_>) -> syn::Result<Self> {
         let mut options: PyClassPyO3Options = Default::default();
 
@@ -138,7 +139,9 @@ impl PyClassPyO3Options {
 
         Ok(options)
     }
+}
 
+impl PyClassPyO3Options {
     pub fn take_pyo3_options(&mut self, attrs: &mut Vec<syn::Attribute>) -> syn::Result<()> {
         take_pyo3_options(attrs)?
             .into_iter()
@@ -369,73 +372,24 @@ fn impl_class(
     })
 }
 
-struct PyClassEnumVariant<'a> {
-    ident: &'a syn::Ident,
-    options: EnumVariantPyO3Options,
-}
-
-impl<'a> PyClassEnumVariant<'a> {
-    fn python_name(&self, args: &PyClassArgs) -> Cow<'_, syn::Ident> {
-        self.options
-            .name
-            .as_ref()
-            .map(|name_attr| Cow::Borrowed(&name_attr.value.0))
-            .unwrap_or_else(|| {
-                let name = self.ident.unraw();
-                if let Some(attr) = &args.options.rename_all {
-                    let new_name = apply_renaming_rule(attr.value.rule, &name.to_string());
-                    Cow::Owned(Ident::new(&new_name, Span::call_site()))
-                } else {
-                    Cow::Owned(name)
-                }
-            })
-    }
-}
-
-struct PyClassEnum<'a> {
-    ident: &'a syn::Ident,
-    // The underlying #[repr] of the enum, used to implement __int__ and __richcmp__.
-    // This matters when the underlying representation may not fit in `isize`.
-    repr_type: syn::Ident,
-    variants: Vec<PyClassEnumVariant<'a>>,
+enum PyClassEnum<'a> {
+    Simple(PyClassSimpleEnum<'a>),
+    Complex(PyClassComplexEnum<'a>),
 }
 
 impl<'a> PyClassEnum<'a> {
     fn new(enum_: &'a mut syn::ItemEnum) -> syn::Result<Self> {
-        fn is_numeric_type(t: &syn::Ident) -> bool {
-            [
-                "u8", "i8", "u16", "i16", "u32", "i32", "u64", "i64", "u128", "i128", "usize",
-                "isize",
-            ]
-            .iter()
-            .any(|&s| t == s)
-        }
-        let ident = &enum_.ident;
-        // According to the [reference](https://doc.rust-lang.org/reference/items/enumerations.html),
-        // "Under the default representation, the specified discriminant is interpreted as an isize
-        // value", so `isize` should be enough by default.
-        let mut repr_type = syn::Ident::new("isize", proc_macro2::Span::call_site());
-        if let Some(attr) = enum_.attrs.iter().find(|attr| attr.path().is_ident("repr")) {
-            let args =
-                attr.parse_args_with(Punctuated::<TokenStream, Token![!]>::parse_terminated)?;
-            if let Some(ident) = args
-                .into_iter()
-                .filter_map(|ts| syn::parse2::<syn::Ident>(ts).ok())
-                .find(is_numeric_type)
-            {
-                repr_type = ident;
-            }
-        }
-
-        let variants = enum_
+        let has_only_unit_variants = enum_
             .variants
-            .iter_mut()
-            .map(extract_variant_data)
-            .collect::<syn::Result<_>>()?;
-        Ok(Self {
-            ident,
-            repr_type,
-            variants,
+            .iter()
+            .all(|variant| matches!(variant.fields, syn::Fields::Unit));
+
+        Ok(if has_only_unit_variants {
+            let simple_enum = PyClassSimpleEnum::new(enum_)?;
+            Self::Simple(simple_enum)
+        } else {
+            let complex_enum = PyClassComplexEnum::new(enum_)?;
+            Self::Complex(complex_enum)
         })
     }
 }
@@ -458,6 +412,208 @@ pub fn build_py_enum(
     let doc = utils::get_doc(&enum_.attrs, None);
     let enum_ = PyClassEnum::new(enum_)?;
     impl_enum(enum_, &args, doc, method_type)
+}
+
+struct PyClassSimpleEnum<'a> {
+    ident: &'a syn::Ident,
+    // The underlying #[repr] of the enum, used to implement __int__ and __richcmp__.
+    // This matters when the underlying representation may not fit in `isize`.
+    repr_type: syn::Ident,
+    variants: Vec<PyClassEnumUnitVariant<'a>>,
+}
+
+impl<'a> PyClassSimpleEnum<'a> {
+    fn new(enum_: &'a mut syn::ItemEnum) -> syn::Result<Self> {
+        fn is_numeric_type(t: &syn::Ident) -> bool {
+            [
+                "u8", "i8", "u16", "i16", "u32", "i32", "u64", "i64", "u128", "i128", "usize",
+                "isize",
+            ]
+            .iter()
+            .any(|&s| t == s)
+        }
+
+        fn extract_unit_variant_data(
+            variant: &mut syn::Variant,
+        ) -> syn::Result<PyClassEnumUnitVariant<'_>> {
+            use syn::Fields;
+            let ident = match &variant.fields {
+                Fields::Unit => &variant.ident,
+                _ => bail_spanned!(variant.span() => "Must be a unit variant."),
+            };
+            let options = EnumVariantPyO3Options::take_pyo3_options(&mut variant.attrs)?;
+            Ok(PyClassEnumUnitVariant { ident, options })
+        }
+
+        let ident = &enum_.ident;
+
+        // According to the [reference](https://doc.rust-lang.org/reference/items/enumerations.html),
+        // "Under the default representation, the specified discriminant is interpreted as an isize
+        // value", so `isize` should be enough by default.
+        let mut repr_type = syn::Ident::new("isize", proc_macro2::Span::call_site());
+        if let Some(attr) = enum_.attrs.iter().find(|attr| attr.path().is_ident("repr")) {
+            let args =
+                attr.parse_args_with(Punctuated::<TokenStream, Token![!]>::parse_terminated)?;
+            if let Some(ident) = args
+                .into_iter()
+                .filter_map(|ts| syn::parse2::<syn::Ident>(ts).ok())
+                .find(is_numeric_type)
+            {
+                repr_type = ident;
+            }
+        }
+
+        let variants: Vec<_> = enum_
+            .variants
+            .iter_mut()
+            .map(extract_unit_variant_data)
+            .collect::<syn::Result<_>>()?;
+        Ok(Self {
+            ident,
+            repr_type,
+            variants,
+        })
+    }
+}
+
+struct PyClassComplexEnum<'a> {
+    ident: &'a syn::Ident,
+    variants: Vec<PyClassEnumVariant<'a>>,
+}
+
+impl<'a> PyClassComplexEnum<'a> {
+    fn new(enum_: &'a mut syn::ItemEnum) -> syn::Result<Self> {
+        let witness = enum_
+            .variants
+            .iter()
+            .find(|variant| !matches!(variant.fields, syn::Fields::Unit))
+            .expect("complex enum has a non-unit variant")
+            .ident
+            .to_owned();
+
+        let extract_variant_data =
+            |variant: &'a mut syn::Variant| -> syn::Result<PyClassEnumVariant<'a>> {
+                use syn::Fields;
+                let ident = &variant.ident;
+                let options = EnumVariantPyO3Options::take_pyo3_options(&mut variant.attrs)?;
+
+                let variant = match &variant.fields {
+                    Fields::Unit => {
+                        bail_spanned!(variant.span() => format!(
+                        "Unit variant `{ident}` is not yet supported in a complex enum\n\
+                        = help: change to a struct variant with no fields: `{ident} {{ }}`\n\
+                        = note: the enum is complex because of non-unit variant `{witness}`",
+                        ident=ident, witness=witness))
+                    }
+                    Fields::Named(fields) => {
+                        let fields = fields
+                            .named
+                            .iter()
+                            .map(|field| PyClassEnumVariantNamedField {
+                                ident: field.ident.as_ref().expect("named field has an identifier"),
+                                ty: &field.ty,
+                                span: field.span(),
+                            })
+                            .collect();
+
+                        PyClassEnumVariant::Struct(PyClassEnumStructVariant {
+                            ident,
+                            fields,
+                            options,
+                        })
+                    }
+                    Fields::Unnamed(_) => {
+                        bail_spanned!(variant.span() => format!(
+                        "Tuple variant `{ident}` is not yet supported in a complex enum\n\
+                        = help: change to a struct variant with named fields: `{ident} {{ /* fields */ }}`\n\
+                        = note: the enum is complex because of non-unit variant `{witness}`",
+                        ident=ident, witness=witness))
+                    }
+                };
+
+                Ok(variant)
+            };
+
+        let ident = &enum_.ident;
+
+        let variants: Vec<_> = enum_
+            .variants
+            .iter_mut()
+            .map(extract_variant_data)
+            .collect::<syn::Result<_>>()?;
+
+        Ok(Self { ident, variants })
+    }
+}
+
+enum PyClassEnumVariant<'a> {
+    // TODO(mkovaxx): Unit(PyClassEnumUnitVariant<'a>),
+    Struct(PyClassEnumStructVariant<'a>),
+    // TODO(mkovaxx): Tuple(PyClassEnumTupleVariant<'a>),
+}
+
+trait EnumVariant {
+    fn get_ident(&self) -> &syn::Ident;
+    fn get_options(&self) -> &EnumVariantPyO3Options;
+
+    fn get_python_name(&self, args: &PyClassArgs) -> Cow<'_, syn::Ident> {
+        self.get_options()
+            .name
+            .as_ref()
+            .map(|name_attr| Cow::Borrowed(&name_attr.value.0))
+            .unwrap_or_else(|| {
+                let name = self.get_ident().unraw();
+                if let Some(attr) = &args.options.rename_all {
+                    let new_name = apply_renaming_rule(attr.value.rule, &name.to_string());
+                    Cow::Owned(Ident::new(&new_name, Span::call_site()))
+                } else {
+                    Cow::Owned(name)
+                }
+            })
+    }
+}
+
+impl<'a> EnumVariant for PyClassEnumVariant<'a> {
+    fn get_ident(&self) -> &syn::Ident {
+        match self {
+            PyClassEnumVariant::Struct(struct_variant) => struct_variant.ident,
+        }
+    }
+
+    fn get_options(&self) -> &EnumVariantPyO3Options {
+        match self {
+            PyClassEnumVariant::Struct(struct_variant) => &struct_variant.options,
+        }
+    }
+}
+
+/// A unit variant has no fields
+struct PyClassEnumUnitVariant<'a> {
+    ident: &'a syn::Ident,
+    options: EnumVariantPyO3Options,
+}
+
+impl<'a> EnumVariant for PyClassEnumUnitVariant<'a> {
+    fn get_ident(&self) -> &syn::Ident {
+        self.ident
+    }
+
+    fn get_options(&self) -> &EnumVariantPyO3Options {
+        &self.options
+    }
+}
+
+/// A struct variant has named fields
+struct PyClassEnumStructVariant<'a> {
+    ident: &'a syn::Ident,
+    fields: Vec<PyClassEnumVariantNamedField<'a>>,
+    options: EnumVariantPyO3Options,
+}
+
+struct PyClassEnumVariantNamedField<'a> {
+    ident: &'a syn::Ident,
+    ty: &'a syn::Type,
+    span: Span,
 }
 
 /// `#[pyo3()]` options for pyclass enum variants
@@ -506,10 +662,24 @@ fn impl_enum(
     doc: PythonDoc,
     methods_type: PyClassMethodsType,
 ) -> Result<TokenStream> {
+    match enum_ {
+        PyClassEnum::Simple(simple_enum) => impl_simple_enum(simple_enum, args, doc, methods_type),
+        PyClassEnum::Complex(complex_enum) => {
+            impl_complex_enum(complex_enum, args, doc, methods_type)
+        }
+    }
+}
+
+fn impl_simple_enum(
+    simple_enum: PyClassSimpleEnum<'_>,
+    args: &PyClassArgs,
+    doc: PythonDoc,
+    methods_type: PyClassMethodsType,
+) -> Result<TokenStream> {
     let krate = get_pyo3_crate(&args.options.krate);
-    let cls = enum_.ident;
+    let cls = simple_enum.ident;
     let ty: syn::Type = syn::parse_quote!(#cls);
-    let variants = enum_.variants;
+    let variants = simple_enum.variants;
     let pytypeinfo = impl_pytypeinfo(cls, args, None);
 
     let (default_repr, default_repr_slot) = {
@@ -519,7 +689,7 @@ fn impl_enum(
             let repr = format!(
                 "{}.{}",
                 get_class_python_name(cls, args),
-                variant.python_name(args),
+                variant.get_python_name(args),
             );
             quote! { #cls::#variant_name => #repr, }
         });
@@ -534,7 +704,7 @@ fn impl_enum(
         (repr_impl, repr_slot)
     };
 
-    let repr_type = &enum_.repr_type;
+    let repr_type = &simple_enum.repr_type;
 
     let (default_int, default_int_slot) = {
         // This implementation allows us to convert &T to #repr_type without implementing `Copy`
@@ -601,7 +771,10 @@ fn impl_enum(
         cls,
         args,
         methods_type,
-        enum_default_methods(cls, variants.iter().map(|v| (v.ident, v.python_name(args)))),
+        simple_enum_default_methods(
+            cls,
+            variants.iter().map(|v| (v.ident, v.get_python_name(args))),
+        ),
         default_slots,
     )
     .doc(doc)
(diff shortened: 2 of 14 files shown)


.python-version
davidhewitt, line 1: The major changes in benchmark speeds are probably because you changed this file, I suggest reverting :)

davidhewitt (commented):
Thanks very much for working on this! 🙏 

I haven't reviewed the code in detail yet, ping me whenever you want that done.

Can you summarise the intended functionality you want to have for the full ADTs, maybe by pushing some tests (in the `pytests/` subdirectory may be easiest). It would be worth checking everyone understands the target design is before we go into too much detail on reviewing the implementation.

davidhewitt:
Ah, I see your examples posted above 👍. Overall I think that looks sensible, and the ability to autogenerate constructors is consistent with what happens on the Rust side.

What are the next steps you have for this PR? Any questions you would like my input on?

.python-version
mkovaxx: Thanks for flagging this! I was indeed scratching my head over it... 😅

mkovaxx:
@davidhewitt Thanks for taking a look! I've been using PyO3 since 2019 and am getting a ton of value out of it, so I'm really excited about the opportunity to contribute back! :)

## TL;DR / TODO

Plan to get this PR ready for review:
1. Reduce scope to handling only struct-like (as opposed to tuple-like) variants for now, just to make my life easier. I think extension to tuple-like variants would be a relatively straightforward add-on in a follow-up PR.
2. Modify the generated `impl IntoPy<PyObject>` to wrap the native Rust enum into the matching variant PyClass.
3. Modify the generated `impl FromPyObject` to unwrap the native Rust enum from the variant PyClass.
4. Add `pytests`.

Any guidance on how best to approach TODOs 2 and 3 would be greatly appreciated! :)

## Design Overview

Because I don't understand everything in PyO3, I try to reuse as much machinery as possible instead of rolling my own. My understanding is particularly spotty around `PyClassPyO3Options`, so any thoughts would be welcome on whether changes are needed in that area! :)

A complex enum is an enum with any non-unit variants. For now, tuple-like variants are out of scope.

For each variant of a complex enum, PyO3 generates a hidden Rust newtype. Each variant newtype wraps a value of the enum, and implements PyClass with an appropriate constructor and getter methods for each of its fields. A runtime invariant is that each variant newtype must contain a value of the matching variant.

The PyClass of the enum type itself has no constructor, so it cannot be instantiated from Python. Instead of a constructor, it has a class attribute corresponding to each variant, which returns the PyType of the matching variant newtype. This allows instantiating each variant newtype from Python.

## Example & Details

To make this more concrete, here's an example.

```rust
#[pyclass]
pub enum ComplexEnum {
    Int { i: i32 },
    Float { f: f64 },
    Str { s: String },
}
```

Under the hood, PyO3 generates these hidden variant newtypes:

```rust
struct ComplexEnum_Int(ComplexEnum);
struct ComplexEnum_Float(ComplexEnum);
struct ComplexEnum_Str(ComplexEnum);
```

Each of these has a constructor and getter methods for its fields.

E.g. `ComplexEnum_Int` will have the following generated `impl`:
```rust
impl ComplexEnum_Int {
    fn __pymethod_constructor__(i: i32) -> Self {
        Self(ComplexEnum::Int { i })
    }
    fn i(&self) -> _pyo3::PyResult<i32> {
        match &self.0 {
            ComplexEnum::Int { i, .. } => Ok(i.clone()),
            _ => unreachable!("Wrong complex enum variant found in wrapper newtype"),
        }
    }
}
```

(Omitting the generated PyClass impls for brevity.)

This allows the following operations on the Python side:
```python
# instantiation
value = ComplexEnum.Int(42)

# extracting fields
print(value.i)

# the wrapper newtype only has getters for fields actually present on the variant
# this will fail with AttributeError: 'builtins.ComplexEnum_Int' object has no attribute 'f'
print(value.f)
# this will fail with AttributeError: 'builtins.ComplexEnum_Int' object has no attribute 's'
print(value.s)

# pattern matching
match value:
    case ComplexEnum.Int(i=x):
        print(f"Int({x})")
    case ComplexEnum.Float(f=x):
        print(f"Float({x})")
    case ComplexEnum.Str(s=x):
        print(f"Str({x})")
    case _:
        print("Invalid")
```

One peculiarity, however, is that `isinstance(ComplexEnum.Float(3.14), ComplexEnum)` returns `False`. As far as I understand, this would be difficult to fix because of how the Python C API works. It also doesn't seem to be a huge usability problem.

It should also be possible to seamlessly pass values back-and-forth between Rust and Python. (This isn't done yet, see TODOs 2 and 3.)

In particular, something like this should work:

```rust
#[pyfunction]
pub fn do_stuff(thing: &ComplexEnum) -> ComplexEnum {
    match thing {
        ComplexEnum::Int { i } => ComplexEnum::Str { s: format!("{i}") },
        ComplexEnum::Float { f } => ComplexEnum::Float { f: f * f },
        ComplexEnum::Str { s } => ComplexEnum::Int { i: s.len() as i32 }
    }
}
```

As far as I understand, this requires `impl FromPyObject` for the variant wrapper newtypes, and `impl IntoPy<PyObject>` for the enum itself.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -640,6 +768,157 @@ fn impl_enum(
...
+    }
+
+    let cls_impl = quote! {
+        #[doc(hidden)]
+        #[allow(non_snake_case)]
+        impl #variant_cls {
+            fn __pymethod_constructor__(#(#fields_with_types)*) -> Self {
+                // TODO(mkovaxx): Fix this to work with tuple variants
+                Self(#enum_name::#variant_ident { #(#field_names)* })
+            }
```
davidhewitt, line 909: With the zero-sized variant types I think this will need to be something more like:

```suggestion
            fn __pymethod_constructor__(py: Python<'_>, #(#fields_with_types)*) -> PyObject {
                // TODO(mkovaxx): Fix this to work with tuple variants
                #enum_name::#variant_ident { #(#field_names)* }.into_py(py)
            }
```

... and we can lean on the `IntoPy` implementation working correctly.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -640,6 +768,157 @@ fn impl_enum(
...
+        )?;
+
+        let field_getter_impl = quote! {
+            fn #field_name(&self) -> _pyo3::PyResult<#field_type> {
+                match &self.0 {
+                    // TODO(mkovaxx): Fix this to work with tuple variants
+                    #enum_name::#variant_ident { #field_name, .. } => Ok(#field_name.clone()),
+                    _ => unreachable!("Wrong complex enum variant found in wrapper newtype"),
+                }
+            }
```
davidhewitt, line 893: With the zero-sized variants I think this will instead need to be:

```suggestion
            fn #field_name(slf: PyRef<Self>) -> _pyo3::PyResult<#field_type> {
                match &*slf.into_super() {
                    // TODO(mkovaxx): Fix this to work with tuple variants
                    #enum_name::#variant_ident { #field_name, .. } => Ok(#field_name.clone()),
                    _ => unreachable!("Wrong complex enum variant found in wrapper newtype"),
                }
            }
```

davidhewitt (commented):
> For now, tuple-like variants are out of scope.

Seems reasonable, I'm happy to have that solved in a follow-up PR, agreed.

> One peculiarity, however, is that `isinstance(ComplexEnum.Float(3.14), ComplexEnum)` returns `False`. As far as I understand, this would be difficult to fix because of how the Python C API works. It also doesn't seem to be a huge usability problem.

For 3., things probably just work if we get this to work. I think having this return `True` is also desirable for users. I think this will also be important for `#[pymethods] impl ComplexEnum` to work properly on the variant subclasses.

I've added some suggestions to the code which might solve this, however this requires changing the variants from being newtypes to being zero-sized markers to build Python subclasses.

> [ ] 2. Modify the generated `impl IntoPy<PyObject>` to wrap the native Rust enum into the matching variant PyClass.

For 2., I think a working approach with the zero-sized variant types would be to have the generated `IntoPy` implementation look like this:

(See `impl_into_py` in `pyo3-macros-backend/src/pyclass.rs`)

```rust
                impl _pyo3::IntoPy<_pyo3::PyObject> for #cls {
                    fn into_py(self, py: _pyo3::Python) -> _pyo3::PyObject {
                        let variant = match &self {
                            // deduce the newtype variant here
                        }
                        _pyo3::IntoPy::into_py(_pyo3::Py::new(py, PyClassInitializer::from(self).add_subclass(variant)).unwrap(), py)
                    }
                }
```

The downside of this is that `IntoPy` for the enum will diverge from `Py::new`, which I think is a bit of a footgun. I suggest we focus on getting the rest of the functionality agreed upon first, and then later can work on a refactoring to get these aligned again.

davidhewitt:
Just to add again: thank you for working on this, so many users will be joyful to see this land! I think this is a difficult blend of implementation and design, so this will probably have several rounds of review and maybe some painful challenges. With patience we can make a lot of people happy 🚀

mkovaxx:
@davidhewitt Thanks for adding a ton of suggestions! I understand that getting this PR landed might take a while, and I'm prepared to help make the necessary changes. 🚀 

Using ZSTs for the variant PyClasses seems like an ingenious approach and not something I'd have thought of on my own! 😄  I'll spend some time today to see where that takes us.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -640,6 +770,164 @@ fn impl_enum(
...
+            #[allow(non_camel_case_types)]
+            struct #variant_cls;
+        }
+    });
+
+    let variant_cls_pytypeinfos = variants.iter().map(|variant| {
+        let variant_cls = gen_complex_enum_variant_class_ident(cls, variant.ident);
+
+        // TODO(mkovaxx): propagate variant.options
+        let options: PyClassPyO3Options = parse_quote! {
```
mkovaxx, line 809: @davidhewitt This is how I'm trying to make the variant PyClasses frozen and subclass the enum PyClass, but it doesn't seem to be effective... I'm stuck, please help! 😅

davidhewitt: It looks like on line 818 below you call `impl_pytypeinfo` with the options, but you're building separate args with default options for `PyClassImplsBuilder` below. Maybe merge this loop with the one below and reuse the same options?

mkovaxx: Whoops, that was a silly mistake.
Thanks for pointing it out! :)

mkovaxx:
@davidhewitt I need your help again for more high-level guidance, because I feel lost in the traits of PyO3. :sweat_smile:

Here's my understanding of the error messages I'm getting.

For each variant PyClass, already existing machinery generates wrapper methods for both the constructor and the field getters, i.e. `__pymethod___new____` and `__pymethod_get_#field__`, along with corresponding `INTRINSIC_ITEMS` on the `PyClassImpl`.

For example, for `ComplexEnum_Int`'s field `i`, we get the following getter generated:
```rust
unsafe fn __pymethod_get_i__(
    py: _pyo3::Python<'_>,
    _slf: *mut _pyo3::ffi::PyObject,
) -> _pyo3::PyResult<*mut _pyo3::ffi::PyObject> {
    let arg = _pyo3::impl_::extract_argument::extract_pyclass_ref::<ComplexEnum_Int>(
        py.from_borrowed_ptr::<_pyo3::PyAny>(_slf),
        &mut { _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT },
    )?;
    _pyo3::callback::convert(py, ComplexEnum_Int::i(arg))
}
```

I've slightly modified the code here, creating the `arg` binding so we can talk about it. The problem here seems to be that the type of `arg` comes out to be `&ComplexEnum_Int`, but the Rust-side `ComplexEnum_Int::i` takes a `PyRef`, because it needs to get its super:
```rust
fn i(slf: _pyo3::PyRef<Self>) -> _pyo3::PyResult<i32> {
    match &*slf.into_super() {
        ComplexEnum::Int { i, .. } => Ok(i.clone()),
        _ => unreachable!("Wrong complex enum variant found in variant wrapper PyClass"),
    }
}
```

So it seems like inside `__pymethod_get_i__` the `&ComplexEnum_Int` needs to be converted into a `PyRef<ComplexEnum_Int>`, but I'm not sure how to do that. Is that even what is supposed to happen here?

There's another problem inside the constructor wrapper, e.g. for `ComplexEnum_Int`:
```rust
unsafe fn __pymethod___new____(
    py: _pyo3::Python<'_>,
    subtype: *mut _pyo3::ffi::PyTypeObject,
    _args: *mut _pyo3::ffi::PyObject,
    _kwargs: *mut _pyo3::ffi::PyObject,
) -> _pyo3::PyResult<*mut _pyo3::ffi::PyObject> {
    use _pyo3::callback::IntoPyCallbackOutput;
    let function = ComplexEnum_Int::__pymethod_constructor__;
    const DESCRIPTION: _pyo3::impl_::extract_argument::FunctionDescription =
        _pyo3::impl_::extract_argument::FunctionDescription {
            cls_name: ::std::option::Option::Some(
                <ComplexEnum_Int as _pyo3::type_object::PyTypeInfo>::NAME,
            ),
            func_name: "__new__",
            positional_parameter_names: &["i"],
            positional_only_parameters: 0usize,
            required_positional_parameters: 1usize,
            keyword_only_parameters: &[],
        };
    let mut output = [::std::option::Option::None; 1usize];
    let (_args, _kwargs) =
            DESCRIPTION.extract_arguments_tuple_dict::<_pyo3::impl_::extract_argument::NoVarargs,
                    _pyo3::impl_::extract_argument::NoVarkeywords>(py, _args,
                    _kwargs, &mut output)?;
    let result = ComplexEnum_Int::__pymethod_constructor__(
        py,
            _pyo3::impl_::extract_argument::unwrap_required_argument(output[0usize]),
            &mut { _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT },
            "i",
        )?,
    );
    let initializer: _pyo3::PyClassInitializer<ComplexEnum_Int> = result.convert(py)?;
    let cell = initializer.create_cell_from_subtype(py, subtype)?;
    ::std::result::Result::Ok(cell as *mut _pyo3::ffi::PyObject)
}
```

Here `result` comes out of `ComplexEnum_Int::__pymethod_constructor__`, which looks like this:
```rust
fn __pymethod_constructor__(py: _pyo3::Python<'_>, i: i32) -> _pyo3::PyObject {
    use _pyo3::IntoPy;
    ComplexEnum::Int { i }.into_py(py)
}
```

So `result` is a `PyObject` aka `Py<PyAny>`, and the surrounding code attempts to convert that into a `PyClassInitializer<ComplexEnum_Int>`, which causes an error about trait impls missing.

I can easily change the code generated for `ComplexEnum_Int::__pymethod_constructor__` and `ComplexEnum_Int::i`, so if there's a way to solve these problems doing that, that would be the easiest.

But I realize that generating some trait impls might also solve these issues without changing anything else, so I figured I'd ask for your insight here.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -681,14 +961,155 @@ fn enum_default_methods<'a>(
...
+    field_name: &'a syn::Ident,
+    field_type: &'a syn::Type,
+    field_span: Span,
+) -> Result<MethodAndMethodDef> {
+    let signature = crate::pyfunction::FunctionSignature::from_arguments(vec![])?;
+
+    let self_type = crate::method::SelfType::Receiver {
+        mutable: false,
+        span: field_span,
+    };
```
davidhewitt, line 1091: I think you need to change this to `SelfType::TryFromPyCell` to make the field getters compile. (This is the cause of your generated getter trying to take `&self`.)

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -640,6 +770,156 @@ fn impl_enum(
...
+    }
+
+    let cls_impl = quote! {
+        #[doc(hidden)]
+        #[allow(non_snake_case)]
+        impl #variant_cls {
+            fn __pymethod_constructor__(py: _pyo3::Python<'_>, #(#fields_with_types)*) -> _pyo3::PyObject {
+                use _pyo3::IntoPy;
+                #enum_name::#variant_ident { #(#field_names)* }.into_py(py)
+            }
```
davidhewitt, line 910: Hmm, thinking about this further, probably you should make this return `PyClassInitializer<#cls>`, based on something like the implementation of the `IntoPy` suggestion I wrote at the bottom of https://github.com/PyO3/pyo3/pull/3582#pullrequestreview-1738372364

davidhewitt (commented):
Two suggestions which should hopefully point you in the right direction.

Just a note - all of the macros code is fairly organically grown to suit the functionality we needed; a lot of it is not beautiful. If you feel a need to refactor any of it, that's welcome. (Although maybe best to isolate refactoring into separate PRs, for the sake of review.)

pyo3-macros-backend/src/pyclass.rs
mkovaxx: This helped, thanks! The fix is a little ugly, but it works now! :)

pyo3-macros-backend/src/pyclass.rs
mkovaxx: This also did the trick, thank you! :)

mkovaxx:
@davidhewitt I've added more `pytests`. The remaining failures are due to the variant PyClasses being cast into the enum PyClass when returned from a Rust function. I think your comment https://github.com/PyO3/pyo3/pull/3582#pullrequestreview-1738372364 describes the fix for this, so I'll try that.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -640,6 +771,203 @@ fn impl_enum(
...
+
+        let variant_new = complex_enum_variant_new(cls, variant)?;
+
+        let (variant_cls_impl, field_getters) = impl_complex_enum_variant_cls(cls, variant)?;
+        variant_cls_impls.push(variant_cls_impl);
+
+        let pyclass_impl = PyClassImplsBuilder::new(
+            &variant_cls,
+            &variant_args,
+            PyClassMethodsType::Specialization,
```
davidhewitt, line 880: I think this hard-coding is the source of the clippy failure, you'll have to make sure it matches the corresponding input to `build_py_enum`.

mkovaxx: Thanks for pointing this out preemptively! This was definitely giving me a hard time figuring out what's going on. Even went down the rabbit hole of reading about the `inventory` crate. 😅

davidhewitt:
Exciting to see a green CI run! I'm dealing with a sick family at the moment so I may struggle to review this in depth for a few days, please bear with me.

mkovaxx:
> Exciting to see a green CI run! I'm dealing with a sick family at the moment so I may struggle to review this in depth for a few days, please bear with me.

@davidhewitt Thanks for the heads-up, and no rush! I'll do some cleanup and other chores in the meantime. I hope your family gets well soon! :)

mejrs (commented):
I haven't looked at the implementation in detail but I have played around with your branch and I'd like to point out the following footgun:

```python
x = enums.ComplexEnum.Int(42)

match x:
    case enums.ComplexEnum.Int(n = 42):
        pass
    case _:
        raise AssertionError    
```

which raises the `AssertionError`.

Perhaps it is easier to only support single tuple enums like 
```rust
#[pyclass]
pub enum Value {
    Int(i32),
    Str(String),
}
```
for now, instead of enums with named fields?

davidhewitt:
If I read right, the problem in that example is the wrong name for the attribute was used, `n = 42` when `i = 42` would have been correct.

Isn't that just an inevitable weakness of python match statements though? I'm unsure we can ever do better than just documenting this sort of looseness of the Python type system.

mkovaxx:
@mejrs Thanks for taking the time to try it out!

As David says, this is just how Python pattern matching syntax works, unfortunately.

Even if we had written that entirely in Python, something like this:
```python
from dataclasses import dataclass

@dataclass
class IntVariant:
    i: int
```

It would have the same behavior.
```python
x = IntVariant(42)

match x:
    case IntVariant(n=42):
        print(f"Field n is 42")
    case IntVariant(i=42):
        print(f"Field i is 42")
    case _:
        raise AssertionError
```

I.e. this prints "Field i is 42".

The goal here is to match the behavior of Python, so I don't think this is a problem.

davidhewitt:
Hmm, here's another problem case though, which makes subclassing a little more awkward.

Imagine we have a method which mutates the enum to change the variant, like this:

```rust
#[pymethods]
impl ComplexEnum {
    fn set_to_string(&mut self, s: String) {
        *self = ComplexEnum::Str { s }
    }
}
```

By changing the variant, we completely break our current type assumptions:

```python
>>> from pyo3_pytests.enums import ComplexEnum
>>> i = ComplexEnum.Int(32)
>>> isinstance(i, ComplexEnum.Int)
True
>>> i.set_to_string("foo")
>>> isinstance(i, ComplexEnum.Int)
True
>>> isinstance(i, ComplexEnum.Str)
False
>>> i.i
thread '<unnamed>' panicked at pytests/src/enums.rs:36:1:
internal error: entered unreachable code: Wrong complex enum variant found in variant wrapper PyClass
pyo3_runtime.PanicException: internal error: entered unreachable code: Wrong complex enum variant found in variant wrapper PyClass
```

I'm not entirely sure what the fix is here. A simple solution would be to require all enums to be `#[pyclass(frozen)]`. That might be attractive for other reasons related to nogil anyway. It _might_ even be backwards-compatible to remove that constraint later.

----

But I wonder, could we change the design to support mutating the variant? I suppose that the subclasses don't work as they are in that case. What happens if we make the subclasses have a custom `__instancecheck__`?

e.g. something like this pure-python code:

```python
from typing import Any


class FooVariantMeta(type):
    def __instancecheck__(self, instance: Any) -> bool:
        if not isinstance(instance, Foo):
            return False

        match self:
            case Foo.Int:
                return isinstance(instance.value, int)
            case Foo.Str:
                return isinstance(instance.value, str)


class Foo:
    value: int | str

    def __init__(self, value) -> None:
        self.value = value

    class Int(metaclass=FooVariantMeta):
        def __new__(cls, value: int) -> "Foo":
            return Foo(value)

    class Str:
        def __new__(cls, value: str) -> "Foo":
            return Foo(value)


i = Foo.Int(1)

print(isinstance(i, Foo))  # True
print(isinstance(i, Foo.Int))  # True
print(isinstance(i, Foo.Str))  # False

match i:
    case Foo.Int(value=x):
        print("int", x)  # 1
    case Foo.Str(value=x):
        print("str", x)  # Never reached
```

NB metaclasses aren't something we've attempted to support yet in PyO3, I think they only work with `abi3` on 3.12, so this would limit the version support I think.

... opinions welcome on what seems like the better route to go down.

mkovaxx:
@davidhewitt That's indeed an issue I haven't considered, thanks for flagging!

Here are the different perspectives that I can think of here.

## (1) Python Semantics

I think the idea of an _in-place_ operation that changes the type/class of a value is rather startling from a Python perspective. If we go with the design choice that the variants correspond to different types/classes in Python, then the function in your example does exactly that. Because of that, the ability to create such functions and call them from Python seems like a footgun. On the other hand, to enable Python pattern matching for the variants, they need to behave like different types/classes, either by virtue of _being_ or by a custom `__instancecheck__`.

## (2) Customizing `__instancecheck__`

A quick search brings up an [open issue](https://github.com/PyO3/pyo3/issues/906) and a [discussion](https://github.com/PyO3/pyo3/discussions/1970) that suggest it might currently be infeasible.

## (3) Rust in Practice

The need for a `&mut self` method on an `enum` feels rather contrived to me. Alternatives seem to exist for all cases I can think of.

First the example your mentioned:
```rust
#[pymethods]
impl ComplexEnum {
    fn set_to_string(&mut self, s: String) {
        *self = ComplexEnum::Str { s }
    }
}
```

I imagine this would be used from Python like so:
```python
x: ComplexEnum = ComplexEnum.Int(42)
x.set_to_string("hello")
```

Because the result doesn't depend on `self`, the method could be replaced with a function like this:
```rust
#[pymethods]
impl ComplexEnum {
    #[staticmethod]
    fn from_string(s: String) -> Self {
        ComplexEnum::Str { s }
    }
}
```

And used like this from Python:
```python
x: ComplexEnum = ComplexEnum.Int(42)
x = ComplexEnum.from_string("hello")
```

I've also tried to imagine cases where the result does depend on `self`, like this:
```rust
#[pymethods]
impl ComplexEnum {
    fn do_stuff_in_place(&mut self) {
        *self = match self {
            ComplexEnum::Int { i } => ComplexEnum::Str { s: i.to_string() },
            ComplexEnum::Float { f } => ComplexEnum::Float { f: f * f },
            ComplexEnum::Str { s } => ComplexEnum::Int { i: s.len() as i32 },
        }
    }
}
```

Again, used like this from Python:
```python
x: ComplexEnum = ComplexEnum.Int(42)
x.do_stuff_in_place()
```

It seems like it is always a viable alternative to construct and return a new value instead of in-place mutation:
```rust
#[pymethods]
impl ComplexEnum {
    fn do_stuff(&self) -> Self {
        match self {
            ComplexEnum::Int { i } => ComplexEnum::Str { s: i.to_string() },
            ComplexEnum::Float { f } => ComplexEnum::Float { f: f * f },
            ComplexEnum::Str { s } => ComplexEnum::Int { i: s.len() as i32 },
        }
    }
}
```

Then use it from Python like this:
```python
x: ComplexEnum = ComplexEnum.Int(42)
x = x.do_stuff()
```

In a nutshell, replacing `x.do_stuff_in_place()` with `x = x.do_stuff()` works both in Python and Rust. The Haskeller part of my brain insists that it is a more readable style as well. ;)

Based on the considerations above, I'm in favor of requiring enums to be `#[pyclass(frozen)]` for now.

Do you see any glaring downsides with this approach?

davidhewitt:
1 - I somewhat agree; while the Python snippet I wrote in https://github.com/PyO3/pyo3/pull/3582#issuecomment-1825174194 would change variant "type" when assigning to `Foo.value`, I do think it's a bit strange

2 - I think we could make `__instancecheck__` work, it would just take a lot of effort. There are a couple appealing things about making the variants not actually be real subclasses:
  - We can then mark the enum type as final (e.g. not subclassable) - at the moment we force `#[pyclass(subclass)]`. But if someone creates another subclass of `ComplexEnum` (e.g. in pure Python) then it will never match any of the variants. The `__instancecheck__` variants wouldn't require `ComplexEnum` to be subclassable, and I think even if someone did subclass it the `__instancecheck__` tests could probably be compatible with a subclass.
  - In Rust the enum variants aren't real types (at least not until an RFC comes along to change that), I think these variants that are only really useful for pattern matching are potentially a closer reproduction of that semantic.

3 - I am mostly ok with requiring the enums to be `#[pyclass(frozen)]`, but here's a couple of caveats:
  - I think we want to understand how we can lift this restriction in the future backwards-compatibly. E.g. I think if we had to change to the `__instancecheck__` design to lift the restriction, then it's potentially not backwards-compatible.
  - I mostly agree that constructing a new value rather than mutating in-place is a better design pattern, but there might be performance drawbacks to this (creating Python objects is relatively expensive, plus it might force copying of other data too).

mkovaxx:
@davidhewitt Thanks for taking the time to think about this issue and write your thoughts down!

To ensure that we will be able to backwards-compatibly move from the current "base class" approach to the "final" approach, I tried to identify the key design constraints that must be satisfied regardless of the approach.

The following key design constraints derive from the semantics of Python:

1. Pattern matching in Python desugars to a sequence of if-elif branches with `isinstance` checks. Therefore, each Rust enum variant must have a corresponding PyType (which may or may not be a PyClass), so that it may be used in patterns. E.g. for `ComplexEnum::Int` on the Rust side, there must be a corresponding PyType, so that a `case ComplexEnum.Int` pattern may be written.
2. It must be possible to construct a value of a variant using the corresponding PyType. E.g. to construct a `ComplexEnum::Int`, whatever the `ComplexEnum.Int` PyType may be, it must have either `__init__` or `__new__`.
3. Given a value of a variant, `isinstance` must return `True` with (1) the PyType of the variant and (2) the PyType of the enum, and must return `False` with any other PyType.

Please try to think of more key constraints we should include here!

The main approaches we've identified so far, and how they each satisfy the constraints above, are as follows.

## The "Base Class" Approach

The approach this PR currently explores.

The PyType of the enum is a base class, and the PyType of each variant is a subclass of it. Each actual value of the enum is an instance of a variant's PyClass.

Constraint (1) is trivially satisfied.
Constraint (2) is trivially satisfied.
Constraint (3) is satisfied because of subclassing.

To protect runtime invariants, we need to disallow in-place mutation of enum values, the enum's PyClass is made frozen.

## The "Final" Approach

The PyType of the enum is a final (i.e. non-subclassable) PyClass, and the PyType of each variant is unrelated. Each value of the enum is an instance of the _enum's_ PyClass; the variant PyTypes are never instantiated.

Constraint (1) is trivially satisfied.
Constraint (2) is satisfied by each variant PyType implementing `__new__`, returning a value of the enum's PyClass. Constraint (3) is satisfied by setting the metaclass of the enum's PyClass with a custom `__instancecheck__`. 

This approach would allow for in-place mutation of enum values. Because we have no examples of using the metaclass machinery in PyO3, more time and effort will be needed to work out the details of this approach.

## Backwards Compatibility

It seems reasonable to start with the base class approach, and eventually move to the final approach, while ensuring that the constraints above are satisfied.

The main concern is that the base class approach would make it possible for a user to subclass the enum's PyClass. I agree that ideally we'd have a mechanism to prevent that. I think it would be a reasonable start to simply warn users in the documentation against doing that.

davidhewitt:
@mkovaxx thanks for writing that up. Here's some further thoughts:

> possible for a user to subclass the enum's PyClass

I don't think it's a problem when switching from the "base class" approach to the "final" approach, because it's possible for users to add `#[pyclass(subclass)]` to continue to allow this.

----

For backwards compatibility, I'm more concerned about what properties change between the two approaches which we cannot mitigate.

I can only think of one at the moment, which is that with the "base class" approach then `type(my_enum)` will be one of the variants, e.g. `ComplexEnum.Int`, but with the "final" approach then `type(my_enum)` will always just be `ComplexEnum`. 

Is this a problem? In a corner case somewhere, maybe someone depends on this property. If we think this (or any other similar properties matter) then it means we have to be more careful about changing the approach later.

mkovaxx:
@davidhewitt That's another corner case indeed, good point!

The general wisdom in Python circles is to rely on `isinstance` instead of `type`. I can't imagine a use case where `type` must be used over `isinstance`, apart from observing private internals of the enum representation. In fact, heavy use of `isinstance` is itself discouraged, but pattern matches desugar to it so that's a given of the design here.

To bring back an earlier concern, I think the larger picture that emerges here is that making `&mut self` methods accessible from Python is likely not worth the trouble. There are two separate aspects here: strange surface behavior that breaks Python semantics, and increased complexity not only under the hood, but also leaking to the surface.

### First, the semantics-breaking aspect.

The [Python documentation](https://docs.python.org/3/reference/datamodel.html) says:

> An object’s type determines the operations that the object supports (e.g., “does it have a length?”) and also defines the possible values for objects of that type. The [type()](https://docs.python.org/3/library/functions.html#type) function returns an object’s type (which is an object itself). Like its identity, **an object’s _type_ is also unchangeable**. [1](https://docs.python.org/3/reference/datamodel.html#id13)

Then the fine print attached says:

> It _is_ possible in some cases to change an object’s type, under certain controlled conditions. It **generally isn’t a good idea** though, since it can lead to some very strange behaviour if it is handled incorrectly.

(Emphasis mine.)

My interpretation is that having a value change its type due to an operation is strange behavior, and the documentation advises against it. I think if I were to write a Python library where that happens, community feedback on it would be "please don't do that". Exposing `&mut self` methods to Python enables creating such behavior, so I think we should not allow it.

### Second, added complexity that also leaks to the surface.

Subclassing satisfies a lot of the key design constraints without any additional machinery.

The approaches without subclassing, on the other hand, have additional complexity to satisfy key design constraints:

1. Customize `__instancecheck__` to make `isinstance` and pattern matching work. This causes unintuitive behavior in relation to `type`, which cannot be customized. (This is the problem you identified above.)
2. Customize `__new__` on each variant, so that instances may be constructed. Again, this causes unintuitive behavior, because the variants themselves are never actually instantiated; every value is of the enum type.

This seems like a lot of complexity and induced strange behavior to alleviate what AFAIU is a concern about performance. I think we should not put a performance concern above general usability.

### Let's get the data!

I think we don't really understand how much of a need there is to expose `&mut self` methods. So it seems reasonable to go with the approach that disallows them, and wait for issues to be filed. In other words, let's collect the data!

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -628,6 +759,201 @@ fn impl_enum(
...
+    complex_enum: PyClassComplexEnum<'_>,
+    args: &PyClassArgs,
+    doc: PythonDoc,
+    methods_type: PyClassMethodsType,
+) -> Result<TokenStream> {
+    // Need to rig the enum PyClass options
+    let args = {
+        let mut rigged_args = args.clone();
+        // Needs to be frozen to disallow `&mut self` methods, which could break a runtime invariant
+        rigged_args.options.frozen = parse_quote!(frozen);
```
mkovaxx, line 772: @davidhewitt 

I've put this in to see if it works.
It does, but the error message might be a little difficult for a user to understand.
```
error[E0271]: type mismatch resolving `<ComplexEnum as PyClass>::Frozen == False`
  --> lib.rs:48:30
   |
48 |     pub fn do_stuff_in_place(&mut self) {
   |                              ^ expected `False`, found `True`
   |
note: required by a bound in `extract_pyclass_ref_mut`
  --> /home/mate/prog/pyo3/src/impl_/extract_argument.rs:57:56
   |
57 | pub fn extract_pyclass_ref_mut<'a, 'py: 'a, T: PyClass<Frozen = False>>(
   |                                                        ^^^^^^^^^^^^^^ required by this bound in `extract_pyclass_ref_mut`
```

Do you see a way to improve this?

davidhewitt:
Sorry for the delay in review, I got a bit swamped by a few things going on in code and irl. Overall I'm thinking it makes sense to move ahead with this design, I'll try to give a full in depth review by the end of the weekend.

mkovaxx:
@davidhewitt Hey David, no worries, thanks for coming back to this PR! :)

Please take your time with the review!

mkovaxx:
@davidhewitt Ping! Just want to make sure this PR doesn't sink. 😅

davidhewitt:
Thanks for the ping! Yes, I will try to open a PR to switch the pyclass defaults to frozen ASAP and come back to this one with an aim to merge after.

davidhewitt:
Ok, so to come back here, I've decided not to move ahead with changing the pyclass settings in 0.21 as there's too much changing anyway.

Let's proceed with this PR as designed regardless, unless I hear concerns from other maintainers before we proceed to merge.

@mkovaxx if I give this a fill review I'm the next few days, are you still interested in working on this according to any feedback I have? Sorry for the long delay.

mkovaxx:
@davidhewitt Thanks for the update, and happy holidays! :)

Yes, I'm committed to making any necessary changes based on review feedback!

pyo3-macros-backend/src/pyclass.rs
davidhewitt: I think this is totally fine; it's the same as what we already have for the rest of frozen pyclass UX (regardless of enums). It's purely a documentation problem at this point I think, outside the scope of this PR.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -628,6 +759,201 @@ fn impl_enum(
...
+                        let variant_value = _pyo3::Py::new(py, pyclass_init).unwrap();
+                        _pyo3::IntoPy::into_py(variant_value, py)
+                    }
+                }
+            })
+            .collect();
+
+        quote! {
+            impl _pyo3::IntoPy<_pyo3::PyObject> for #cls {
+                fn into_py(self, py: _pyo3::Python) -> _pyo3::PyObject {
```
davidhewitt, line 816: So we will still have the papercut that `Py::new` doesn't match `IntoPy` for these enums. I think we should document in this PR as a known footgun and we can solve in a follow-up issue.

pytests/src/enums.rs
```diff
@@ -0,0 +1,50 @@
...
+        SimpleEnum::Wednesday => SimpleEnum::Thursday,
+        SimpleEnum::Thursday => SimpleEnum::Friday,
+        SimpleEnum::Friday => SimpleEnum::Saturday,
+        SimpleEnum::Saturday => SimpleEnum::Sunday,
+    }
+}
+
+#[pyclass]
+pub enum ComplexEnum {
+    Int { i: i32 },
```
davidhewitt, line 38: What happens if we add a unit variant like `Empty` here. There's three forms:

```rust
    Empty,
    Empty { },
    Empty(),
```

I think the tuple-like case will be similar to the struct-like, and we can solve in a follow-up issue with the rest of the tuple-like variants. For the `Empty {}` struct-like, I think this will already behave as I'd expect (create a new variant subclass with no fields).

But what about the `Empty` identifier? Should this create a new variant too, or should it create a constant like the `SimpleEnum` cases?

I think it has a nice consistency if it works like a `SimpleEnum` variant, but I can also see reasons why that might be confusing. If we are feeling conservative, one option is to disallow `Empty` in `ComplexEnum` for now, track this as a known issue, and invite users to propose the right design later.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -442,24 +478,105 @@ impl<'a> PyClassEnum<'a> {
...
+            .map(|name_attr| Cow::Borrowed(&name_attr.value.0))
+            .unwrap_or_else(|| {
+                let name = self.ident.unraw();
+                if let Some(attr) = &args.options.rename_all {
+                    let new_name = apply_renaming_rule(attr.value.rule, &name.to_string());
+                    Cow::Owned(Ident::new(&new_name, Span::call_site()))
+                } else {
+                    Cow::Owned(name)
+                }
+            })
```
davidhewitt, line 500: There's a little duplication here with the `python_name` for the complex variant, perhaps can be unified?

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -628,6 +759,201 @@ fn impl_enum(
...
+        rigged_args.options.subclass = parse_quote!(subclass);
+        rigged_args
+    };
+
+    let krate = get_pyo3_crate(&args.options.krate);
+    let cls = complex_enum.ident;
+    let variants = complex_enum.variants;
+    let pytypeinfo = impl_pytypeinfo(cls, &args, None);
+
+    let default_slots = vec![];
```
davidhewitt, line 783: For the simple enums we generate a default `__repr__`, I wonder if that's useful here?

We also generate a default `__richcmp__` ordering conversion, which I guess we cannot do for this complex enum without opt-in for the user. Makes me more sure we want to have `#[pyclass(ord, eq)]` as opt-ins which simple enums could then opt in to, to make things more consistent. But that's a problem for a separate PR I think.

davidhewitt (commented):
Thanks, I finally got around to a full review. Again, sorry for the delays, and Happy New Year!

Overall I'm very happy with your implementation, well done on figuring out the weeds of the proc-macro implementation!

I've mostly written a few design questions which I think we will want to settle before merge; they're small edge cases so I think easy for us to decide what action to take.

It would also be great to add UI tests for things which we have decided to disallow, such as tuple variants, both so that we can see what users will experience and also to remind us what's not yet available. (See `test_compile_error`.)

pyo3-macros-backend/src/pyclass.rs
mkovaxx: I've added a test case using `test_compile_error` to document that `&mut self` methods are not supported.

pytests/src/enums.rs
mkovaxx: I've added a test case using `test_compile_error` to document that tuple-like variants are not currently supported.

You're correct; the struct-like `Empty {}` variant corresponds to a PyClass with no fields. I've extended the `pytests` to document this.

My stance is that a unit-like `Empty` variant should behave identically to a struct-like `Empty {}` variant as far as the Python side is concerned. (And that a tuple-like `Empty()` variant should be identical to them as well, when support is added.) This is what is currently implemented, so I've extended the `pytests` to document this.

The alternative, as you suggest, is to make a unit-like `Empty` variant correspond to a constant value on the Python side. This would require special-casing in multiple places in the generator. Because it seems to have little benefit in terms of functionality, apart from removing the `()` in expressions and patterns, I'm against doing this. Please let me know if you think this assessment is incorrect/incomplete.

On the usability side, I think it's also better to _not_ make the unit-like variants in a complex enum behave like the unit-like variants in a simple enum, for the purpose of clearly distinguishing these two types of enums from the user's perspective. A "simple enum" is essentially an int with named values (like `enum` in C), whereas a "complex enum" is a tagged union.

mkovaxx:
Hi David, Happy New Year, and thank you for the review! :)

I'm going to work through your comments in the next few days and let you know when I'm done.

pytests/src/enums.rs
davidhewitt: My concern is based around silent behaviour changes. Consider the following enum:

```rust
enum Color {
    Red,
    Blue,
    Green,
}
```

In current PyO3, this is already supported, and creates a single type with three constants `Color.Red` etc. which are instances of that type.

Now, add a fourth non-unit variant:

```rust
enum Color {
    Red,
    Blue,
    Green,
    Custom { r: u8, g: u8, b: u8 },
}
```

If we follow the line that `Red` is equivalent to `Red { }`, then just by adding the non-unit variant all the other enums have also changed behaviour. Now to get an instance of the red variant, a Python user must write `Color.Red()`, as `Color.Red` now instead refers to the variant type.

I think it's very likely that users will trip up on this while refactoring code. This makes me feel that the right thing to do is either:
- make sure that `Color.Red` does not change behaviour (as I propose)
- fail to compile and emit an error stating that unit and non-unit variants may not be mixed (which allows us to support this case later)

mkovaxx: Thanks for describing your concern in more detail, I understand!

I'm in favor of disallowing the mixing of unit and non-unit variants for now, and opening an issue for the remaining work, same as for tuple variants.

To add to your concern about silent behavior changes. Even if we manage to keep the instantiation and pattern syntax consistent across simple and complex enums, there will remain major differences in semantics. Simple enums can be coerced into integers, which is not meaningful for complex enums.

I'm going to make some changes to both disallow unit variants and ensure the code is in a good shape for future extension work.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -440,24 +476,127 @@ impl<'a> PyClassEnum<'a> {
...
+            .collect::<syn::Result<_>>()?;
+
+        Ok(Self { ident, variants })
     }
+}
 
-    let doc = utils::get_doc(&enum_.attrs, None);
-    let enum_ = PyClassEnum::new(enum_)?;
-    impl_enum(enum_, &args, doc, method_type)
+enum PyClassEnumVariant<'a> {
```
mkovaxx, line 532: @davidhewitt I've introduced this type to make it easier to find where to start adding unit and tuple variant support.

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -440,24 +476,127 @@ impl<'a> PyClassEnum<'a> {
...
+    // TODO(mkovaxx): Unit(PyClassEnumUnitVariant<'a>),
+    Struct(PyClassEnumStructVariant<'a>),
+    // TODO(mkovaxx): Tuple(PyClassEnumTupleVariant<'a>),
+}
+
+trait EnumVariant {
+    fn get_ident(&self) -> &syn::Ident;
+    fn get_options(&self) -> &EnumVariantPyO3Options;
+
+    fn get_python_name(&self, args: &PyClassArgs) -> Cow<'_, syn::Ident> {
```
mkovaxx, line 542: Factored out the duplicated `python_name` methods into this.

pyo3-macros-backend/src/pyclass.rs
mkovaxx: I agree about `__richcmp__` being opt-in.

AFAICT generating a default `__repr__` for a complex enum would only be possible if all variant field types also have `__repr__`. I'm not sure if there are facilities to perform that check... So maybe this means `__repr__` should be opt-in as well?

pyo3-macros-backend/src/pyclass.rs
mkovaxx: Could you please explain this in more detail?
I think there should be no `Py::new` for the enum PyClass itself, but I probably misunderstand what you're saying...

pyo3-macros-backend/src/pyclass.rs
```diff
@@ -440,24 +476,127 @@ impl<'a> PyClassEnum<'a> {
...
+}
 
-    let doc = utils::get_doc(&enum_.attrs, None);
-    let enum_ = PyClassEnum::new(enum_)?;
-    impl_enum(enum_, &args, doc, method_type)
+enum PyClassEnumVariant<'a> {
+    // TODO(mkovaxx): Unit(PyClassEnumUnitVariant<'a>),
+    Struct(PyClassEnumStructVariant<'a>),
+    // TODO(mkovaxx): Tuple(PyClassEnumTupleVariant<'a>),
+}
```
mkovaxx, line 536: @davidhewitt I've introduced this type to make it easier to find where to start adding unit and tuple variant support.

mkovaxx:
@davidhewitt This is ready for the next round, please take another look! :)

mkovaxx:
I'm trying to reproduce the CI failure locally, but things seem to be working fine.

This is the command I'm running:
```
RUSTFLAGS="-D warnings" cargo build --lib --tests --no-default-features --features "full "
```

I'm at a loss here; not sure what's going wrong... Please help! 😅

mkovaxx:
Never mind, `rustup default nightly-x86_64-pc-windows-msvc` did the trick...

Now I'm getting the failure locally as well. 😅

pyo3-macros-backend/src/pyclass.rs
davidhewitt: e.g. `Py::new(py, ComplexEnum::Int { i: 42 })` I think will already compile but will not produce an instance of the variant subtype, it will just create an instance of the main `ComplexEnum` base type. This is a less-used feature, so IMO I think we will get away with documenting this as slightly wrong for the first version of this feature and we can work out how to adjust later.

pytests/src/enums.rs
davidhewitt: Yes, the conversion also is a problem, so I agree that disallowing unit variants in complex enums makes sense for now.

pyo3-macros-backend/src/pyclass.rs
davidhewitt: Yes, I think that `#[pyclass(repr)]` which relies on `Debug` impls would also make a lot of sense.

pyo3-macros-backend/src/deprecations.rs
```diff
@@ -1,6 +1,7 @@
 use proc_macro2::{Span, TokenStream};
 use quote::{quote_spanned, ToTokens};
 
+#[derive(Clone)]
```
davidhewitt, line 4: Any reason why the `Clone` impl was necessary here and below? I wonder if the changes in this file can be reverted...

tests/ui/invalid_pyclass_enum.stderr
```diff
@@ -15,3 +15,15 @@ error: #[pyclass] can't be used on enums without any variants
    |
 16 | enum NoEmptyEnum {}
    |                  ^^
+
+error: Unit variants are not yet supported in complex enums.
+  --> tests/ui/invalid_pyclass_enum.rs:21:5
+   |
+21 |     UnitVariant,
```
davidhewitt, line 22: For unit variants, it would be very nice if the error message can name the conflicting variants and also hint how to make this compile:

```suggestion
error: Unit variant `UnitVariant` cannot be mixed with complex variant `StructVariant` in the same enum.
  = help: change to `UnitVariant { }` to opt in to a full variant type without any fields
  --> tests/ui/invalid_pyclass_enum.rs:21:5
   |
21 |     UnitVariant,
```

davidhewitt (commented):
The design and implementation here is now looking great! I'm happy with what we have settled on here.

Regarding the nightly rust failures, it looks like this might affect code that's not modified here too so it's probably best if someone (maybe me later) opens a separate PR to fix those. Seems like there's some new dead code detection.

I think the last step for this PR, now that we're happy with the design, is to update `class.md` in the guide to also cover this new functionality.

pyo3-macros-backend/src/deprecations.rs
mkovaxx: Good catch! I reverted this change and things still seem to work.

tests/ui/invalid_pyclass_enum.stderr
mkovaxx: I've added more details to these error messages, along the lines of what you suggest here, although not verbatim.

src/tests/hygiene/misc.rs
```diff
@@ -2,6 +2,7 @@
 
 #[derive(crate::FromPyObject)]
 #[pyo3(crate = "crate")]
+#[allow(dead_code)]
```
mkovaxx, line 5: Should I revert this?

tests/test_frompyobject.rs
```diff
@@ -421,6 +421,7 @@ TypeError: failed to extract enum Foo ('TupleVar | StructVar | TransparentTuple
 #[derive(Debug, FromPyObject)]
 enum EnumWithCatchAll<'a> {
     #[pyo3(transparent)]
+    #[allow(dead_code)]
```
mkovaxx, line 424: Should I revert this?

mkovaxx:
@davidhewitt I think this is ready for another round of review! :)

davidhewitt (approved):
Just one tiny final suggestion to add to the guide, and then we're good to merge! Thank you for all the work put in here and sorry it took so long to review.

If you rebase on main, the nightly CI failures should be fixed. If you could also please squash the commits into a single commit at the same time, that'd be great; we can't squash-merge with GitHub merge queue.

guide/src/class.md
```diff
@@ -21,21 +21,29 @@ This chapter will discuss the functionality and configuration these attributes o
...
 
-To define a custom Python class, add the `#[pyclass]` attribute to a Rust struct or a fieldless enum.
+To define a custom Python class, add the `#[pyclass]` attribute to a Rust struct or enum.
 ```rust
 # #![allow(dead_code)]
 use pyo3::prelude::*;
 
 #[pyclass]
-struct Integer {
+struct MyClass {
```
mkovaxx, line 30: Updated this so the description below the code section makes sense.

guide/src/class.md
```diff
@@ -21,21 +21,29 @@ This chapter will discuss the functionality and configuration these attributes o
...
     inner: i32,
 }
 
 // A "tuple" struct
 #[pyclass]
 struct Number(i32);
 
-// PyO3 supports custom discriminants in enums
+// PyO3 supports unit-only enums (which contain only unit variants)
```
mkovaxx, line 38: I think it makes more sense to have this before `HttpResponse`.

guide/src/class.md
```diff
@@ -44,14 +52,19 @@ enum HttpResponse {
...
+enum Shape {
+    Circle { radius: f64 },
+    Rectangle { width: f64, height: f64 },
+    RegularPolygon { side_count: u32, radius: f64 },
+    Nothing { },
 }
 ```
 
-The above example generates implementations for [`PyTypeInfo`] and [`PyClass`] for `MyClass` and `MyEnum`. To see these generated implementations, refer to the [implementation details](#implementation-details) at the end of this chapter.
+The above example generates implementations for [`PyTypeInfo`] and [`PyClass`] for `MyClass`, `Number`, `MyEnum`, `HttpResponse`, and `Shape`. To see these generated implementations, refer to the [implementation details](#implementation-details) at the end of this chapter.
```
mkovaxx, line 67: Updated this sentence to mention all types defined in the code section.

guide/src/class.md
```diff
@@ -1090,6 +1096,74 @@ enum BadSubclass {
...
+                case cls.Nothing():
+                    return 0
+
+        assert count_vertices(cls, circle) == 0
+        assert count_vertices(cls, square) == 4
+    "#)
+})
+```
+
+WARNING: `Py::new` and `.into_py` are currently inconsistent. Note how the constructed value is _not_ an instance of the specific variant. For this reason, constructing values is only recommended using `.into_py`.
```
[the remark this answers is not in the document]
mkovaxx: Good point! I'll add a bunch of issues mentioned in the review here as soon as the PR is landed.

mkovaxx:
@davidhewitt Thank you for the ton of guidance and the thorough review here!
Also, I'm not sure why the performance regression is happening, please help with that! :)

davidhewitt (approved):
Looks great, thanks so much!

Don't worry about the performance report; there is some volatility and I recently lowered the noise threshold as an experiment, but I think I must have gone too far.

As a final request, could you please squash the commits into one? Then this is good to merge!

mkovaxx:
@davidhewitt Changes are now in a single commit!

After you land this PR, I'll add issues for the following:
- `Py::new` vs `.into_py` inconsistency
- support tuple variants in complex enums
- support unit variants in complex enums

Did I miss anything?

guide/src/class.md
```diff
@@ -1094,6 +1113,74 @@ enum BadSubclass {
...
+        def count_vertices(cls, shape):
+            match shape:
+                case cls.Circle(radius=r):
+                    return 0
+                case cls.Rectangle(width=w, height=h):
+                    return 4
+                case cls.RegularPolygon(side_count=n, radius=r):
+                    return n
+                case cls.Nothing():
+                    return 0
```
mkovaxx, line 1157: @davidhewitt I think the merge queue checks fail because this `match` syntax is only available in Python 3.10 and up. What would be a good way to fix this?

pytests/tests/test_enums.py
```diff
@@ -0,0 +1,105 @@
...
+        case enums.ComplexEnum.Str(s=x):
+            assert x == "hello"
+        case enums.ComplexEnum.EmptyStruct():
+            assert True
+        case enums.ComplexEnum.MultiFieldStruct(a=x, b=y, c=z):
+            assert x == 42
+            assert y == 3.14
+            assert z is True
+        case _:
+            assert False
```
mkovaxx, line 77: @davidhewitt I think the merge queue checks fail because this `match` syntax is only available in Python 3.10 and up. What would be a good way to fix this?

davidhewitt: Agreed, yes. I am not sure what the right solution should be. Are you aware of any way to make `pytest` ignore files based on a version pattern?

pytests/tests/test_enums.py
```diff
@@ -0,0 +1,179 @@
...
+        z = variant.c
+        assert x == 42
+        assert y == 3.14
+        assert z is True
+    else:
+        assert False
+
+
+# NOTE(mkovaxx): Match statements require at least Python 3.10
+@pytest.mark.skipif(sys.version_info < (3, 10))
```
mkovaxx, line 88: I put these version guards on test cases that use `match` syntax.

pytests/tests/test_enums.py
```diff
@@ -0,0 +1,179 @@
...
+    "variant",
+    [
+        enums.ComplexEnum.Int(42),
+        enums.ComplexEnum.Float(3.14),
+        enums.ComplexEnum.Str("hello"),
+        enums.ComplexEnum.EmptyStruct(),
+        enums.ComplexEnum.MultiFieldStruct(42, 3.14, True),
+    ],
+)
+def test_complex_enum_desugared_match(variant: enums.ComplexEnum):
```
mkovaxx, line 64: I've added variants of the test cases that "desugar" the match statement syntax for Python < 3.10.

pytests/tests/test_enums.py
mkovaxx: @davidhewitt I think I've managed to fix the problem. Just waiting for CI to finish... 🤞

guide/src/class.md
```diff
@@ -1094,6 +1113,90 @@ enum BadSubclass {
...
+            if isinstance(shape, cls.Circle):
+                return 0
+            elif isinstance(shape, cls.Rectangle):
+                return 4
+            elif isinstance(shape, cls.RegularPolygon):
+                n = shape.side_count
+                return n
+            elif isinstance(shape, cls.Nothing):
+                return 0
+    "# };
```
mkovaxx, line 1157: This solves the version issue, but it makes the example ugly...
A more readable but less correct alternative would be to simply skip this test on Python < 3.10.

pytests/noxfile.py
```diff
@@ -13,7 +14,12 @@ def test(session: nox.Session):
     except CommandFailed:
         # No binary wheel for numpy available on this platform
         pass
-    session.run("pytest", *session.posargs)
+    ignored_paths = []
+    if sys.version_info < (3, 10):
+        # Match syntax is only available in Python >= 3.10
+        ignored_paths.append("tests/test_enums_match.py")
```
mkovaxx, line 20: This is where the version-checking magic happens to only run `match`-based tests on Python >= 3.10.

mkovaxx:
@davidhewitt Some tests now seem to fail on crate downloads not coming through.
Maybe that's just network flakiness?

davidhewitt:
Rust nightly had some odd issues yesterday, see https://github.com/PyO3/setuptools-rust/issues/398

I'll rerun the failures. The rest does sound like network failures, yes.

pytests/noxfile.py
davidhewitt: Great solution, nicely done

davidhewitt (approved):
Given the full build looks good, this should succeed to merge this time...

guide/src/class.md
davidhewitt: You could `!#[cfg(Py_3_10)]` at the top of the example, it of sight of users even.

Given I've just set to merge, let's clean this up in a follow up PR.

mkovaxx:
@davidhewitt Thank you for landing this! :)

I've added these follow-up issues:
https://github.com/PyO3/pyo3/issues/3747
https://github.com/PyO3/pyo3/issues/3748
https://github.com/PyO3/pyo3/issues/3749
https://github.com/PyO3/pyo3/issues/3750

PR #3587, opened 2023-11-20
feat: allow classmethods to receive `Py<PyType>`

wyfo:
This is necessary for async classmethods

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -113,12 +113,14 @@ impl FnType {
             }
             FnType::FnClass | FnType::FnNewClass => {
                 quote! {
-                    _pyo3::types::PyType::from_type_ptr(py, _slf as *mut _pyo3::ffi::PyTypeObject),
+                    #[allow(clippy::useless_conversion)]
+                    ::std::convert::Into::into(_pyo3::types::PyType::from_type_ptr(py, _slf as *mut _pyo3::ffi::PyTypeObject)),
                 }
             }
             FnType::FnModule => {
                 quote! {
-                    py.from_borrowed_ptr::<_pyo3::types::PyModule>(_slf),
+                    #[allow(clippy::useless_conversion)]
+                    ::std::convert::Into::into(py.from_borrowed_ptr::<_pyo3::types::PyModule>(_slf)),
                 }
             }
         }
--- a/newsfragments/3587.added.md
+++ b/newsfragments/3587.added.md
@@ -1,0 +1,2 @@
+- Classmethods can now receive `Py<PyType>` as their first argument
+- Function annotated with `pass_module` can now receive `Py<PyModule>` as their first argument
--- a/pyo3-macros-backend/src/pyfunction.rs
+++ b/pyo3-macros-backend/src/pyfunction.rs
@@ -199,7 +199,8 @@ pub fn impl_wrap_pyfunction(
         .collect::<syn::Result<Vec<_>>>()?;
 
     let tp = if pass_module.is_some() {
-        const PASS_MODULE_ERR: &str = "expected &PyModule as first argument with `pass_module`";
+        const PASS_MODULE_ERR: &str =
+            "expected &PyModule or Py<PyModule> as first argument with `pass_module`";
         ensure_spanned!(
             !arguments.is_empty(),
             func.span() => PASS_MODULE_ERR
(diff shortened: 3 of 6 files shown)


adamreichold:
Could you add a test case and change log entry for this? Thanks!

wyfo:
> Also, if you're willing, it would be great to also support this for #[pyfunction(pass_module)] please. I think that will have the same problem.

Indeed, I've added it. I had to modify `type_is_pymodule` function, but I am wondering if this function is really useful, as there is no check like this for `#[classmethod]`.

pyo3-macros-backend/src/method.rs
```diff
@@ -113,12 +113,14 @@ impl FnType {
...
-                    _pyo3::types::PyType::from_type_ptr(py, _slf as *mut _pyo3::ffi::PyTypeObject),
+                    #[allow(clippy::useless_conversion)]
+                    ::std::convert::Into::into(_pyo3::types::PyType::from_type_ptr(py, _slf as *mut _pyo3::ffi::PyTypeObject)),
                 }
             }
             FnType::FnModule => {
                 quote! {
-                    py.from_borrowed_ptr::<_pyo3::types::PyModule>(_slf),
+                    #[allow(clippy::useless_conversion)]
+                    ::std::convert::Into::into(py.from_borrowed_ptr::<_pyo3::types::PyModule>(_slf)),
```
adamreichold, line 123: No for a semantic question: `Into` is quite generic here. Do we want and/or have a specific trait to handle this? `IntoPy`?

wyfo: We need a trait that can return either `&PyModule` or `Py<PyModule>`, so `IntoPy` cannot do the job here. `Into` is the simplest choice, but it's true that it makes possible to use `PyObject` instead of `Py<PyType>`. (There is a macro-level check for `pass_module`, so it's not possible to use `PyObject,` but as I said, I would like to remove it to rely on compilation error)

I will add a new `PyOrRef` trait.

adamreichold: > I will add a new ToPyOrRef trait.

Please wait until doing that to give others some time to chime in. Adding a new trait that becomes part of the `#[doc(hidden)]` macro API is a somewhat costly addition w.r.t. maintenance and I wouldn't want to rust it.

wyfo: Sorry, I have seen you comment after pushing the change. I'm rollbacking it for now.

adamreichold: I would if we can press our usual traits for function argument extract into service here, i.e. `FromPyObject`?

wyfo: `FromPyObject` returns a `Result`, so it requires to add `.unwrap()`, but you would have the same issue as with `Into`, i.e. you can use `PyObject` instead of `Py<PyType>`. I still prefer using `Into` instead of `FromPyObject`.

davidhewitt: There's a strong overlap here with #3564, where we can't decide if we want to stay with `TryFrom`. I think `From` is better than `FromPyObject`, because `FromPyObject` will involve a needless type check (and is fallible as @wyfo observed).

I can't really see any better options than sticking with `From`/`TryFrom` and adding clippy allows, or adding traits to solve clippy complaints with our current `pass_module`, `classmethod` and method receiver implementations.

Given that the existing code works, and there's not really any harm with users taking the module as a `PyObject`, I'm leaning towards sticking with `From`/`TryFrom`.

adamreichold: Let's stick to the appropriately infallible `From` then.

adamreichold (approved):

davidhewitt:
👍

@wyfo similar finishing request here - can you please squash it? Also, if you're willing to check the patch diff on codecov and add an additional test or two to hit the coverage target, that would be massively appreciated 🙏

wyfo:
Actually, the issue with coverage comes from  the error cases `pyfunction::type_is_pymodule`, because I've added 4 of them. But they cannot be covered, because test would not compile; actually, I thought that UI test would be counted in coverage, but it seems not, so I've no way to add coverage for these branches. Have I missed something here?
By the way, I've proposed to remove `pyfunction::type_is_pymodule`, because it's redundant with the type checking, and not consistent with `#[classmethod]` which doesn't have a macro-level check. It would solve the coverage issue.

PR #3588, opened 2023-11-20
feat: add coroutine `__name__`/`__qualname__`

wyfo:
Relates to #1632

### DIFF
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -455,13 +455,23 @@ impl<'a> FnSpec<'a> {
         let func_name = &self.name;
 
         let rust_call = |args: Vec<TokenStream>| {
-            let call = quote! { function(#self_arg #(#args),*) };
-            let wrapped_call = if self.asyncness.is_some() {
-                quote! { _pyo3::PyResult::Ok(_pyo3::impl_::wrap::wrap_future(#call)) }
-            } else {
-                quotes::ok_wrap(call)
-            };
-            quotes::map_result_into_ptr(wrapped_call)
+            let mut call = quote! { function(#self_arg #(#args),*) };
+            if self.asyncness.is_some() {
+                let python_name = &self.python_name;
+                let qualname_prefix = match cls {
+                    Some(cls) => quote!(Some(<#cls as _pyo3::PyTypeInfo>::NAME)),
+                    None => quote!(None),
+                };
+                call = quote! {{
+                    let future = #call;
+                    _pyo3::impl_::coroutine::new_coroutine(
+                        _pyo3::intern!(py, stringify!(#python_name)),
+                        #qualname_prefix,
+                        async move { _pyo3::impl_::wrap::OkWrap::wrap(future.await) }
+                    )
+                }};
+            }
+            quotes::map_result_into_ptr(quotes::ok_wrap(call))
         };
 
         let rust_name = if let Some(cls) = cls {
--- a/src/coroutine.rs
+++ b/src/coroutine.rs
@@ -14,10 +14,10 @@ use pyo3_macros::{pyclass, pymethods};
 
 use crate::{
     coroutine::waker::AsyncioWaker,
-    exceptions::{PyRuntimeError, PyStopIteration},
+    exceptions::{PyAttributeError, PyRuntimeError, PyStopIteration},
     panic::PanicException,
     pyclass::IterNextOutput,
-    types::PyIterator,
+    types::{PyIterator, PyString},
     IntoPy, Py, PyAny, PyErr, PyObject, PyResult, Python,
 };
 
@@ -30,6 +30,8 @@ type FutureOutput = Result<PyResult<PyObject>, Box<dyn Any + Send>>;
 /// Python coroutine wrapping a [`Future`].
 #[pyclass(crate = "crate")]
 pub struct Coroutine {
+    name: Option<Py<PyString>>,
+    qualname_prefix: Option<&'static str>,
     future: Option<Pin<Box<dyn Future<Output = FutureOutput> + Send>>>,
     waker: Option<Arc<AsyncioWaker>>,
 }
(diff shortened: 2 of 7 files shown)


davidhewitt:
@wyfo if you can rebase this please I'll get around to reviewing as soon as possible after 👍

src/impl_/coroutine.rs
```diff
@@ -1,19 +1,33 @@
...
-    F: Future<Output = R> + Send + 'static,
-    R: OkWrap<T>,
+    F: Future<Output = Result<T, E>> + Send + 'static,
     T: IntoPy<PyObject>,
-    PyErr: From<R::Error>,
+    PyErr: From<E>,
 {
-    let future = async move {
-        // SAFETY: GIL is acquired when future is polled (see `Coroutine::poll`)
-        future.await.wrap(unsafe { Python::assume_gil_acquired() })
```
wyfo, line 16: I've removed this `Python::assume_gil_acquired()` using `OkWrap::wrap_no_gil`. Actually, this modification was not required for this PR (it is for the following about `allow_threads`), but because I had to modified `wrap_future`, I've took the opportunity to do it.

src/impl_/wrap.rs
```diff
@@ -40,6 +44,9 @@ where
     T: IntoPy<PyObject>,
 {
     type Error = E;
+    fn wrap_no_gil(self) -> Result<T, Self::Error> {
+        self
+    }
```
davidhewitt, line 49: Hmm, I think this adds some complexity to the `OkWrap` trait which we didn't previously have, I'll push an alternative, let's see what we think of the alternative.

wyfo: I've rebased on main.

src/coroutine.rs
```diff
@@ -113,6 +122,20 @@ pub(crate) fn iter_result(result: IterNextOutput<PyObject, PyObject>) -> PyResul
 
 #[pymethods(crate = "crate")]
 impl Coroutine {
+    #[getter]
+    fn __name__(&self) -> PyResult<Py<PyString>> {
+        self.name
+            .clone()
```
adamreichold, line 128: Not sure if it is worth it, but using [`clone_ref`](https://docs.rs/pyo3/latest/pyo3/prelude/struct.Py.html#method.clone_ref) would avoid the implicit check for the GIL.

src/coroutine.rs
```diff
@@ -135,3 +158,21 @@ impl Coroutine {
...
+impl Drop for Coroutine {
+    fn drop(&mut self) {
+        if self.future.is_some() {
+            Python::with_gil(|gil| {
+                let qualname = self
+                    .qualname
+                    .as_ref()
+                    .map_or(Ok("<coroutine>"), |n| n.as_ref(gil).to_str())
+                    .unwrap();
+                let message = format!("coroutine {} was never awaited", qualname);
```
adamreichold, line 171: Does this only happen if it was never awaited or also if it was not polled to completion?

pyo3-macros-backend/src/method.rs
```diff
@@ -455,13 +455,27 @@ impl<'a> FnSpec<'a> {
...
+                        _pyo3::impl_::coroutine::method_coroutine_qualname::<#cls>(py, stringify!(#python_name))
+                    },
+                    None => quote! {
+                        _pyo3::impl_::coroutine::coroutine_qualname(py, py.from_borrowed_ptr_or_opt::<_pyo3::types::PyModule>(_slf), stringify!(#python_name))
+                    },
+                };
+                call = quote! {{
+                    let future = #call;
+                    _pyo3::impl_::coroutine::new_coroutine(
+                        _pyo3::types::PyString::new(py, stringify!(#python_name)).into(),
```
davidhewitt, line 472: It would probably be worth interning here using our `intern!` macro, and changing `new_coroutine` to take `N: IntoPy<Py<PyString>>`.

src/impl_/coroutine.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+pub fn method_coroutine_qualname<'py, T: PyClass>(py: Python<'py>, name: &str) -> &'py PyString {
+    let class = T::NAME;
+    let qualname = match T::MODULE {
+        Some(module) => format!("{}.{}.{}", module, class, name),
+        None => format!("{}.{}", class, name),
+    };
+    PyString::new(py, &qualname)
+}
```
davidhewitt, line 36: I wonder, to avoid Python string creation with every coroutine, can this be done lazily (by e.g. storing a reference to `T::type_object`, or taking a function pointer, or storing a reference to the module etc).

src/coroutine.rs
```diff
@@ -135,3 +158,21 @@ impl Coroutine {
         self.poll(py, None)
     }
 }
+
+impl Drop for Coroutine {
+    fn drop(&mut self) {
+        if self.future.is_some() {
+            Python::with_gil(|gil| {
```
davidhewitt, line 165: This structure reminds me of https://github.com/PyO3/pyo3/discussions/3386#discussioncomment-6732913.

I'm not entirely sure if it's a blocker to not have `try_with_gil` or some other guard to avoid potential nasty crashes here. We should have a little think about that before merge.

davidhewitt (commented):
Ok, this diff is now nicely focussed just on the async code so I've given it a full review 👍 

I agree with @adamreichold's points and also added a couple of brief thoughts of my own.

src/coroutine.rs
davidhewitt: Ahh that's a good suggestion, I should really pick up `__del__` support again soon. Are you suggesting split this PR in two; merge the names now and keep the warning as a draft pr for the moment?

wyfo: I am indeed.

src/coroutine.rs
wyfo: In fact, instead of testing `self.future.is_some()`, I should test `self.waker.is_none()`, as it's always `Some(_)` when the coroutine has been polled at least once.

src/impl_/coroutine.rs
wyfo: I've finally used a `qualname_prefix` instead of passing the `__qualname__` directly. 
The constructor parameter may be modified when it will be exposed (maybe using `Option<Cow<'static, str>>` instead of `Option<&'static str>` for example), but for now, it's adapted to the macro and that's enough.

src/coroutine.rs
wyfo: I've opened an issue for the warning.

davidhewitt (approved):
This looks good to me, thanks! Will follow up with discussion on the issue.

PR #3594, opened 2023-11-22
remove `type_is_pymodule`

davidhewitt:
Follow-up to #3587 

This removes the `type_is_pymodule` check. I had to propagate the span of the first argument through the macro code to get the compiler error to hint at the right place, overall I think this is a nice cleanup.


tests/ui/invalid_need_module_arg_position.rs
```diff
@@ -3,10 +3,10 @@ use pyo3::prelude::*;
 #[pymodule]
 fn module(_py: Python<'_>, m: &PyModule) -> PyResult<()> {
     #[pyfn(m, pass_module)]
-    fn fail(string: &str, module: &PyModule) -> PyResult<&str> {
+    fn fail<'py>(string: &str, module: &'py PyModule) -> PyResult<&'py str> {
```
davidhewitt, line 6: The original check was masking an ambiguous lifetime error; I needed to add lifetime hints to avoid having this error also included in the trybuild output.

davidhewitt:
cc @wyfo

davidhewitt:
Ah I'll push a test for the empty arguments branch in the morning.

wyfo:
> I had to propagate the span of the first argument through the macro code to get the compiler error to hint at the right place

Good point, but now it's my turn to tell you that I think you should do the same for `FnClass`/`FnNewClass` :smiley:

wyfo:
Thanks to the structure of your modification, I've realized that I forgot to update the line 380 in method.rs. Could you do that in this PR?

tests/ui/invalid_need_module_arg_position.rs
davidhewitt, line 1: Moved this test into `invalid_pyfunctions.rs`

pyo3-macros-backend/src/method.rs
```diff
@@ -111,10 +111,12 @@ impl FnType {
...
+            FnType::FnClass(span) | FnType::FnNewClass(span) => {
+                let py = syn::Ident::new("py", Span::call_site());
+                let slf: Ident = syn::Ident::new("_slf", Span::call_site());
+                quote_spanned! { *span =>
                     #[allow(clippy::useless_conversion)]
-                    ::std::convert::Into::into(_pyo3::types::PyType::from_type_ptr(py, _slf as *mut _pyo3::ffi::PyTypeObject)),
+                    ::std::convert::Into::into(_pyo3::types::PyType::from_type_ptr(#py, #slf.cast())),
                 }
             }
             FnType::FnModule(span) => {
```
wyfo, line 122: Should `_slf` also be declared as a `syn::Indent` like above?

wyfo:
I've no *Approve* button, so I will just write :+1:

davidhewitt:
👍 thanks for the review!

> I've no Approve button, so I will just write 👍

If you're reviewing regularly we should look at getting that set up 👀

PR #3595, opened 2023-11-24
refactor `OkWrap` to not call `.into_py(py)`

davidhewitt:
Motivated by https://github.com/PyO3/pyo3/pull/3588/files#r1403933922

The goal here is to keep the `OkWrap` trait as simple as it can possibly be.

@wyfo what do you think? It conflicts a little bit with #3588, so that'd need rebasing if this merged, but I like how this keeps `OkWrap` simple.


wyfo:
I didn't dare to do this refactor as part of the async work, but I fully agree with this simplification

src/impl_/wrap.rs
```diff
@@ -30,8 +30,9 @@ where
     T: IntoPy<PyObject>,
 {
     type Error = PyErr;
```
wyfo, line 32: I think the error here should be `Infallible`, as `PyErr` implements `From<Infallible>`

src/impl_/wrap.rs
```diff
@@ -40,11 +41,44 @@ where
...
+pub fn map_result_into_py<T: IntoPy<PyObject>>(
+    py: Python<'_>,
+    result: PyResult<T>,
+) -> PyResult<PyObject> {
+    result.map(|err| err.into_py(py))
+}
+
+/// Used to wrap the result of async `#[pyfunction]` and `#[pymethods]`.
+#[cfg(feature = "macros")]
+pub fn wrap_future<F, R, T>(future: F) -> crate::coroutine::Coroutine
```
wyfo, line 70: I would say it should not be moved, but it don't really matter because it will be reworked in the next PR

src/impl_/wrap.rs
```diff
@@ -40,11 +41,44 @@ where
...
+}
+
+/// This is a follow-up function to `OkWrap::wrap` that converts the result into
+/// a safe wrapper.
+pub fn map_result_into_py<T: IntoPy<PyObject>>(
+    py: Python<'_>,
+    result: PyResult<T>,
+) -> PyResult<PyObject> {
+    result.map(|err| err.into_py(py))
+}
```
wyfo, line 66:
```suggestion
pub fn map_result_into_ptr<T, E>(
    py: Python<'_>,
    result: Result<T, E>,
) -> PyResult<*mut ffi::PyObject>
where
    T: IntoPy<PyObject>,
    PyErr: From<E>,
{
    Ok(result?.into_py(py).into_ptr())
}

/// This is a follow-up function to `OkWrap::wrap` that converts the result into
/// a safe wrapper.
pub fn map_result_into_py<T, E>(
    py: Python<'_>,
    result: Result<T, E>,
) -> PyResult<PyObject>
where
    T: IntoPy<PyObject>,
    PyErr: From<E>,
{
    Ok(result?.into_py(py))
}
```

tests/test_compile_error.rs
```diff
@@ -33,7 +33,7 @@ fn test_compile_errors() {
     t.compile_fail("tests/ui/invalid_pymethod_receiver.rs");
     t.compile_fail("tests/ui/missing_intopy.rs");
     // adding extra error conversion impls changes the output
-    #[cfg(all(target_os = "linux", not(any(feature = "eyre", feature = "anyhow"))))]
```
wyfo, line 36: Why this modification?

davidhewitt: It's a bit of a drive-by but I was working on macOS today and I broke CI in the first attempt because this UI test regressed. Turns out the macos and linux outputs are the same.

src/impl_/wrap.rs
davidhewitt: Good idea 👍

src/impl_/wrap.rs
davidhewitt: I guess it's a bit of a style thing, I slightly prefer the combinators over `?` in these one-line implementations.

wyfo: The suggestion is not about the `?` but using `Result<T, E>` instead of `PyResult<T>`. If you use `map_err` instead of `?`, you can in fact relax the bound to `E: Into<PyErr>` instead of `PyErr: From<E>`, and that's maybe better (I should also relax the bound in async support implementation by the way)

davidhewitt: Ah I see. I take your point (and I did initially accept `Result<T, E>`) but having the `E: Into<PyErr>` bound introduced on this function and doing `.map_err` in here instead of directly in the macro generated code means this internal function `map_result_into_ptr` appears in the compiler error.

I didn't think that was useful (that was the UI test regression noted in the comment below) so I kept with just `PyResult` for now as all we need.

I nearly changed the async support but actually I think it's better if I don't touch that in this PR so that it conflicts less with what you've got stacked up.

wyfo: > [...] means this internal function `map_result_into_ptr` appears in the compiler error.

Makes sense indeed.

wyfo:
:+1: for me

davidhewitt:
Thanks again for review, as this touches more code which other contributors initially helped design I'll wait to merge a little bit in case of other opinions. But not too long, as you'll need to rebase off this 🙂

tests/test_compile_error.rs
mejrs: (I think) I added this because the output is different on Windows. The error message for this lists all the variants, and windows has additional error classes.

davidhewitt: Yep that's what I recall too, which is the why I left windows in the `not(any())` list

PR #3598, opened 2023-11-26
ci: run beta clippy as an allowed-to-fail job

davidhewitt:
Following #3561 I thought it might be a good idea to have beta clippy running in our CI as an allowed-to-fail job; this can function as a canary giving us 6 weeks to release a fix before users are impacted by any new clippy failures in macro code.


davidhewitt:
Looks like this is green now that we merged #3564. I'm leaning towards merging this so that we can catch future incoming lints without relying on user reports.

PR #3599, opened 2023-11-26
feat: add `coroutine::CancelHandle`

wyfo:
Relates to #1632 

See https://github.com/PyO3/pyo3/pull/3540#discussion_r1375988581 for the reason behind `CancelHandle` implementation.

### DIFF
--- a/pyo3-macros-backend/src/attributes.rs
+++ b/pyo3-macros-backend/src/attributes.rs
@@ -11,6 +11,7 @@ use syn::{
 pub mod kw {
     syn::custom_keyword!(annotation);
     syn::custom_keyword!(attribute);
+    syn::custom_keyword!(cancel_handle);
     syn::custom_keyword!(dict);
     syn::custom_keyword!(extends);
     syn::custom_keyword!(freelist);
--- a/pyo3-macros-backend/src/method.rs
+++ b/pyo3-macros-backend/src/method.rs
@@ -24,6 +24,7 @@ pub struct FnArg<'a> {
     pub attrs: PyFunctionArgPyO3Attributes,
     pub is_varargs: bool,
     pub is_kwargs: bool,
+    pub is_cancel_handle: bool,
 }
 
 impl<'a> FnArg<'a> {
@@ -44,6 +45,8 @@ impl<'a> FnArg<'a> {
                     other => return Err(handle_argument_error(other)),
                 };
 
+                let is_cancel_handle = arg_attrs.cancel_handle.is_some();
+
                 Ok(FnArg {
                     name: ident,
                     ty: &cap.ty,
@@ -53,6 +56,7 @@ impl<'a> FnArg<'a> {
                     attrs: arg_attrs,
                     is_varargs: false,
                     is_kwargs: false,
+                    is_cancel_handle,
                 })
             }
         }
@@ -455,9 +459,27 @@ impl<'a> FnSpec<'a> {
         let self_arg = self.tp.self_arg(cls, ExtractErrorMode::Raise);
         let func_name = &self.name;
 
+        let mut cancel_handle_iter = self
+            .signature
+            .arguments
+            .iter()
+            .filter(|arg| arg.is_cancel_handle);
+        let cancel_handle = cancel_handle_iter.next();
+        if let Some(arg) = cancel_handle {
+            ensure_spanned!(self.asyncness.is_some(), arg.name.span() => "`cancel_handle` attribute can only be used with `async fn`");
+            if let Some(arg2) = cancel_handle_iter.next() {
+                bail_spanned!(arg2.name.span() => "`cancel_handle` may only be specified once");
+            }
+        }
+
         let rust_call = |args: Vec<TokenStream>| {
             let mut call = quote! { function(#self_arg #(#args),*) };
             if self.asyncness.is_some() {
+                let throw_callback = if cancel_handle.is_some() {
+                    quote! { Some(__throw_callback) }
+                } else {
+                    quote! { None }
+                };
                 let python_name = &self.python_name;
                 let qualname_prefix = match cls {
                     Some(cls) => quote!(Some(<#cls as _pyo3::PyTypeInfo>::NAME)),
@@ -468,9 +490,17 @@ impl<'a> FnSpec<'a> {
                     _pyo3::impl_::coroutine::new_coroutine(
                         _pyo3::intern!(py, stringify!(#python_name)),
                         #qualname_prefix,
-                        async move { _pyo3::impl_::wrap::OkWrap::wrap(future.await) }
+                        #throw_callback,
+                        async move { _pyo3::impl_::wrap::OkWrap::wrap(future.await) },
                     )
                 }};
+                if cancel_handle.is_some() {
+                    call = quote! {{
+                        let __cancel_handle = _pyo3::coroutine::CancelHandle::new();
+                        let __throw_callback = __cancel_handle.throw_callback();
+                        #call
+                    }};
+                }
             }
             quotes::map_result_into_ptr(quotes::ok_wrap(call))
         };
@@ -483,12 +513,21 @@ impl<'a> FnSpec<'a> {
 
         Ok(match self.convention {
             CallingConvention::Noargs => {
-                let call = if !self.signature.arguments.is_empty() {
-                    // Only `py` arg can be here
-                    rust_call(vec![quote!(py)])
-                } else {
-                    rust_call(vec![])
-                };
+                let args = self
+                    .signature
+                    .arguments
+                    .iter()
+                    .map(|arg| {
+                        if arg.py {
+                            quote!(py)
+                        } else if arg.is_cancel_handle {
+                            quote!(__cancel_handle)
+                        } else {
+                            unreachable!()
+                        }
+                    })
+                    .collect();
+                let call = rust_call(args);
 
                 quote! {
                     unsafe fn #ident<'py>(
--- a/src/coroutine.rs
+++ b/src/coroutine.rs
@@ -21,7 +21,11 @@ use crate::{
     IntoPy, Py, PyAny, PyErr, PyObject, PyResult, Python,
 };
 
+pub(crate) mod cancel;
 mod waker;
+
+use crate::coroutine::cancel::ThrowCallback;
+pub use cancel::CancelHandle;
 
 const COROUTINE_REUSED_ERROR: &str = "cannot reuse already awaited coroutine";
 
@@ -32,6 +36,7 @@ type FutureOutput = Result<PyResult<PyObject>, Box<dyn Any + Send>>;
 pub struct Coroutine {
     name: Option<Py<PyString>>,
     qualname_prefix: Option<&'static str>,
+    throw_callback: Option<ThrowCallback>,
     future: Option<Pin<Box<dyn Future<Output = FutureOutput> + Send>>>,
     waker: Option<Arc<AsyncioWaker>>,
 }
@@ -46,6 +51,7 @@ impl Coroutine {
     pub(crate) fn new<F, T, E>(
         name: Option<Py<PyString>>,
         qualname_prefix: Option<&'static str>,
+        throw_callback: Option<ThrowCallback>,
         future: F,
     ) -> Self
     where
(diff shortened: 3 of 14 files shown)


pyo3-macros-backend/src/attributes.rs
```diff
@@ -9,9 +9,10 @@ use syn::{
 };
 
 pub mod kw {
-    syn::custom_keyword!(args);
     syn::custom_keyword!(annotation);
+    syn::custom_keyword!(args);
```
davidhewitt, line 13: #3602 🙈

davidhewitt (commented):
Thanks, I haven't reviewed the full implementation in detail yet because I think there's a little bit of discussion we could have around the design.

First off, there's two alternatives I can think of and I wonder if they're any good.

The more similar alternative would be to have the cancel handle annotation be on the argument: 

```rust
    #[pyfunction]
    async fn cancellable_sleep(seconds: f64, #[pyo3(cancel_handle)] mut cancel: CancelHandle) -> usize {
        futures::select! {
            _ = sleep(seconds).fuse() => 42,
            _ = cancel.cancelled().fuse() => 0,
        }
    }
```

and another alternative could be to have a function which returns a future which only resolves if the current coroutine is cancelled:

```rust
    async fn cancellable_sleep(seconds: f64) -> usize {
        futures::select! {
            _ = sleep(seconds).fuse() => 42,
            _ = pyo3::coroutine_is_cancelled().fuse() => 0,
        }
    }
```

I think this would be possible by putting the `throw` object into the waker and then this future could retrieve it from the waker.

I quite like this for not needing to pass the `cancel: CancelHandle` argument, but it does feel a bit "global state"-y, which may have unintended footguns. 

----

There's then another question I have in the back of my mind, which is how this interacts with `PyFuture`. What happens if that inner Python awaitable gets cancelled? It would be quite nice for this cancel handle to also deal with that case, but I don't think it will in this design. Is there a way we can unify these?

wyfo:
I like argument annotation, and I prefer it actually. 
I don't like free function for two reasons:
- It seems to require the same kind of hack I've already done for `PyFuture`, but I don't like to have too much of this. What should be the behavior if the function is called outside of coroutine context? panic (like I do for `PyFuture`)? With `CancelHandle` parameter, the check is already done.
- Ignoring cancellation is not really natural with that way, you have to do a `select`, but join `coroutine_is_cancelled` with a pending future in order to have it polled, but not selected. With `CancelHandle` parameter, you have just to not use it.

> There's then another question I have in the back of my mind, which is how this interacts with PyFuture. What happens if that inner Python awaitable gets cancelled? It would be quite nice for this cancel handle to also deal with that case, but I don't think it will in this design. Is there a way we can unify these?

`PyFuture` can't be selected, they must be awaited alone (it panics otherwise). In this case, the `send`/`throw`/`close` methods delegate to the `PyFuture`, which handle cancellation itself (and may return an error which can checked and handled).

wyfo:
> Would this change in the future with the functionality available on nightly?

Yes, normally.

> It might be ok for this future to always just be pending if not in a coroutine context.

I prefer panicking here, because I think it's less error prone.

> Can you please explain in a little bit more detail why this is the case?

The default behavior is to reraise the error when `throw` is called`. This behavior is changed if a `CancelHandle` is registered.
So to delay or ignore cancellation, you just have to not use the `CancelHandle` in the code section you want to "shield. 
How can you do that with a global `coroutine_is_cancelled`? Because you have to disable the default behavior before the shielded code section, and check for cancellation after.

pyo3-macros-backend/src/method.rs
```diff
@@ -44,6 +45,7 @@ impl<'a> FnArg<'a> {
                     other => return Err(handle_argument_error(other)),
                 };
 
+                let is_cancel_handle = arg_attrs.cancel_handle.is_some();
```
adamreichold, line 48: Please leave an empty line after this statement.

pyo3-macros-backend/src/method.rs
```diff
@@ -468,9 +489,17 @@ impl<'a> FnSpec<'a> {
...
                         _pyo3::intern!(py, stringify!(#python_name)),
                         #qualname_prefix,
-                        async move { _pyo3::impl_::wrap::OkWrap::wrap(future.await) }
+                        #throw_callback,
+                        async move { _pyo3::impl_::wrap::OkWrap::wrap(future.await) },
                     )
                 }};
+                if cancel_handle.is_some() {
+                    call = quote! {{
+                        let __cancel_handle = _pyo3::coroutine::CancelHandle::new();
```
adamreichold, line 498: I am not sure how we do this elsewhere, but maybe those names should include something like `__pyo3_` to have to some ad-hoc namespacing?

src/coroutine.rs
```diff
@@ -77,9 +84,13 @@ impl Coroutine {
...
-        if let Some(exc) = throw {
-            self.close();
-            return Err(PyErr::from_value(exc.as_ref(py)));
+        match (throw, &self.throw_callback) {
+            (Some(exc), Some(cb)) => cb.throw(exc.as_ref(py)),
+            (Some(exc), None) => {
+                self.close();
+                return Err(PyErr::from_value(exc.as_ref(py)));
+            }
+            _ => {}
```
adamreichold, line 93: Maybe write this out as `(None, _)` just to make it more obvious.

src/coroutine/cancel.rs
```diff
@@ -0,0 +1,74 @@
...
+        self.0.waker.register(cx.waker());
+        if self.is_cancelled() {
+            return Poll::Ready(take());
+        }
+        Poll::Pending
+    }
+
+    /// Retrieve the exception thrown in the associated coroutine.
+    pub async fn cancelled(&mut self) -> PyObject {
+        poll_fn(|cx| self.poll_cancelled(cx)).await
```
adamreichold, line 55: Since this probably quite likely to be used in `select!` statements, I wonder whether it is reasonable to use a custom type for the future here which actually implements `FusedFuture` directly (thereby avoiding the need to call `.fuse()` and turning it into a no-op in generic context)?

Would this require more state than just checking the pointer? Since it might just be a single bit, do we do pointer tagging else and could apply that here as well?

pyo3-macros-backend/src/method.rs
wyfo: There is no such kind of namespacing anywhere else. On the contrary, variables are named `function`, `output`, `_args`, `_kwnames`, etc. I was quite innovative already by using a double underscore :smile: 
Anyway, I can add this prefix if you want, I've no opinion about it.

src/coroutine/cancel.rs
wyfo: `FusedFuture` semantic is quite questionable, see https://github.com/rust-lang/futures-rs/issues/2779 (I encountered https://github.com/rust-lang/futures-rs/issues/2455 myself while working on pyo3-async...).
That's why I prefer to forget about `FusedFuture` here. But I'm a tokio user, I never use `FusedFuture` at all, so my opinion may be biased.

src/coroutine/cancel.rs
```diff
@@ -0,0 +1,74 @@
...
+        ThrowCallback(self.0.clone())
+    }
+}
+
+#[doc(hidden)]
+pub struct ThrowCallback(Arc<Inner>);
+
+impl ThrowCallback {
+    pub(super) fn throw(&self, exc: &PyAny) {
+        let ptr = self.0.exception.swap(exc.into_ptr(), Ordering::Relaxed);
```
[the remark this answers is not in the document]
wyfo: `thow` is not public, so it should not be called from multiple threads (it's only called by the coroutine). Also, I don't really understand your point. This is a swap, you could have as much concurrent swap as you want, each of them would retrieve a valid (or null) pointer. The memory models rules mainly concerns R/W synchronization between multiple memory locations, but there is only one here. And the mutex already has an acquire/release semantic, so you don't need more barrier here in my opinion. (And this part of the code is protected by the GIL)

pyo3-macros-backend/src/method.rs
adamreichold: No, let's stay consistent then.

src/coroutine/cancel.rs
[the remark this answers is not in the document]
adamreichold: > thow is not public, so it should not be called from multiple threads (it's only called by the coroutine). 

It should not, but there nothing enforcing and/or documenting this and things can change. Especially as concurrency in CPython is up for fundamental changes in the medium term future.

> And the mutex already has an acquire/release semantic, so you don't need more barrier here in my opinion. (And this part of the code is protected by the GIL)

Yes the mutex (and the GIL) imply a stronger memory ordering but they are not used everywhere which makes relying on their properties tricky. Personally, I think the potential performance gain from using the relaxed ordering is not worth it to rely on these complexities.

> Also, I don't really understand your point. This is a swap, you could have as much concurrent swap as you want, each of them would retrieve a valid (or null) pointer.

The point is that the pointer alone is not necessarily all the memory which is accessed drop the drop call. For example if the exception object contains pointers to other objects, their state in memory is not necessarily visible on the thread doing the `drop` under relaxed ordering and hence the drop could encounter invalid values. Acquire-release ordering would ensure that any modifications made to memory reachable via the pointer before swapping it in the atomic variable is also visible to the thread swapping it out.

So long story short, I don't think the concurrency situation here is simple enough so that the relaxed ordering is an easy and/or obvious optimization and most likely is not worth it anyway as it does happen next to stronger memory barriers which will inhibit the reordering potential implied by relaxed in any case. But how exactly is far from cleat, at least to me, and hence going for acquire/release seems to be the reasonable choice here IMHO.

src/coroutine/cancel.rs
adamreichold: It is an optimization without requiring API breakage, so we can investigate whenever anyone asks for it if there are an reservations towards the trait. Let's focus on getting the fundamentals right instead then.

src/coroutine/cancel.rs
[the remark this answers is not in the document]
wyfo: To make this code simpler, I will include the swap in the mutex scope in the next MR, so there will be no doubt about synchronization.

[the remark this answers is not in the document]
adamreichold: Actually, I would suggest to use only the mutex for now and do anything else as a separate optimization pass. There should be little chance for contention and then a mutex is just an atomic access anyway. This will be obviously correct and side-step any discussion of memory orderings.

Taking a step backwards. Couldn't we just store the exception as `Py<PyAny>` then to further avoid unsafe code?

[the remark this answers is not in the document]
wyfo: > The point is that the pointer alone is not necessarily all the memory which is accessed drop the drop call. For example if the exception object contains pointers to other objects, their state in memory is not necessarily visible on the thread doing the drop under relaxed ordering and hence the drop could encounter invalid values.

You're 100% right.

> Taking a step backwards. Couldn't we just store the exception as Py<PyAny> then to further avoid unsafe code?

You're 100% right.

adamreichold (approved):

src/coroutine/cancel.rs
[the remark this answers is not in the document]
adamreichold: :+1: Much simpler now.

Let's look into getting that last bit of performance in follow-up changes.

wyfo:
Thank you a lot for the review. I've added a last test for `CancelHandle::is_cancelled`.

davidhewitt (commented):
Sorry for not managing to review this until just now. All looks great in my eyes, thanks @adamreichold for being so thorough as always!

I just had one tiny suggestion, which may not be a great idea depending on compiler output...

tests/ui/invalid_argument_attributes.rs
```diff
@@ -12,4 +12,26 @@ fn from_py_with_string(#[pyo3("from_py_with")] param: String) {}
...
+#[pyfunction]
+async fn cancel_handle_repeated2(
+    #[pyo3(cancel_handle)] _param: String,
+    #[pyo3(cancel_handle)] _param2: String,
+) {
+}
+
+#[pyfunction]
+fn cancel_handle_synchronous(#[pyo3(cancel_handle)] _param: String) {}
+
```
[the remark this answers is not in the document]
wyfo: I've added your test cases. I let you judge about the compiler output.

PR #3531, opened 2023-10-20
implement PyDictMethods

davidhewitt:
As part of #3382, this implements the `PyDictMethods` trait as `pub(crate)` and re-implements the existing `PyDict` API on top of it.

PR #3600, opened 2023-11-27
replace `PyTryFrom` by splitting `PyTypeInfo`

davidhewitt:
This is a proposed resolution to #3516 and overall I think a positive incremental cleanup to our traits setup.

This PR splits `PyTypeInfo` into three traits:
 - `HasPyGilRef`, which can be removed in the near future when we get rid of the GIL refs, but will be useful in the short term. It defines the `AsRefTarget` associated type.
 - `PyTypeCheck`, which is used for types like `PyIterator`, `PySequence`, and `PyMapping`, as these can't implement `PyTypeInfo` as they don't have concrete type objects.
 - `PyTypeInfo`, which retains its existing API.

This then removes PyO3's internal dependency on `PyTryFrom`:
- `PyAny::downcast()` and `PyAnyMethods::downcast` use `PyTypeCheck`.
- `PyAny::downcast_exact()` and `PyAnyMethods::downcast_exact` to use `PyTypeInfo`.
- `PyAny::downcast_unchecked()` uses `HasPyGilRef`. (`PyAnyMethods::downcast_unchecked` needs no helper trait.)

From my experience playing around with #3382, e.g. in #3531, having this adjustment makes it much easier to work with `PyIterator`, `PySequence`, and `PyMapping` in code which is migrating between the two APIs.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -304,7 +304,7 @@ where
     T: PyClass,
 {
     fn extract(obj: &'a PyAny) -> PyResult<Self> {
-        PyTryFrom::try_from(obj).map_err(Into::into)
+        obj.downcast().map_err(Into::into)
     }
 }
 
@@ -313,7 +313,7 @@ where
     T: PyClass + Clone,
 {
     fn extract(obj: &'a PyAny) -> PyResult<Self> {
-        let cell: &PyCell<Self> = PyTryFrom::try_from(obj)?;
+        let cell: &PyCell<Self> = obj.downcast()?;
         Ok(unsafe { cell.try_borrow_unguarded()?.clone() })
     }
 }
@@ -323,7 +323,7 @@ where
     T: PyClass,
 {
     fn extract(obj: &'a PyAny) -> PyResult<Self> {
-        let cell: &PyCell<T> = PyTryFrom::try_from(obj)?;
+        let cell: &PyCell<T> = obj.downcast()?;
         cell.try_borrow().map_err(Into::into)
     }
 }
@@ -333,7 +333,7 @@ where
     T: PyClass<Frozen = False>,
 {
     fn extract(obj: &'a PyAny) -> PyResult<Self> {
-        let cell: &PyCell<T> = PyTryFrom::try_from(obj)?;
+        let cell: &PyCell<T> = obj.downcast()?;
         cell.try_borrow_mut().map_err(Into::into)
     }
 }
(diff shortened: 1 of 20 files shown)


src/types/iterator.rs
```diff
@@ -97,22 +101,6 @@ impl<'v> PyTryFrom<'v> for PyIterator {
     }
 }
 
-impl Py<PyIterator> {
```
davidhewitt, line 100: These special cases for `PyIterator` etc are no longer needed, because `as_ref()` and `into_ref()` are made available by the `HasPyGilRef` implementation for all `T: HasPyGilRef`.

src/types/iterator.rs
```diff
@@ -33,11 +34,7 @@ impl PyIterator {
     ///
     /// Equivalent to Python's built-in `iter` function.
     pub fn from_object(obj: &PyAny) -> PyResult<&PyIterator> {
-        Self::from_object2(Py2::borrowed_from_gil_ref(&obj)).map(|py2| {
-            // Can't use into_gil_ref here because T: PyTypeInfo bound is not satisfied
-            // Safety: into_ptr produces a valid pointer to PyIterator object
-            unsafe { obj.py().from_owned_ptr(py2.into_ptr()) }
-        })
+        Self::from_object2(Py2::borrowed_from_gil_ref(&obj)).map(Py2::into_gil_ref)
```
davidhewitt, line 37: This is an example of where this adjustment simplifies interchange between the gil-ref and `Py2` APIs; there are more cases in #3572 which can be simplified also.

src/conversion.rs
```diff
@@ -381,78 +381,61 @@ pub trait PyTryInto<T>: Sized {
...
-    U: for<'v> PyTryFrom<'v>,
-{
-    fn try_into(&self) -> Result<&U, PyDowncastError<'_>> {
-        <U as PyTryFrom<'_>>::try_from(self)
-    }
-    fn try_into_exact(&self) -> Result<&U, PyDowncastError<'_>> {
-        U::try_from_exact(self)
-    }
-}
+mod implementations {
```
adamreichold, line 384: Could you explain why you moved these into the module? Does this make the impls non-public?

src/err/mod.rs
```diff
@@ -53,6 +53,7 @@ pub struct PyDowncastError<'a> {
 impl<'a> PyDowncastError<'a> {
     /// Create a new `PyDowncastError` representing a failure to convert the object
     /// `from` into the type named in `to`.
+    #[cold]
```
adamreichold, line 56: This seems unrelated am I am not sure it is a universally good idea. For example, what about `downcast` chains to identify the correct type? We steer users towards downcasting instead of extracting primarily since constructing `PyDowncastError` is cheaper than `PyErr` and this could affect this.

I am not saying this should never happen, just that I would prefer this a separate change with some measurements and discussion on its impact on reasonable code patterns.

adamreichold (approved):
Some remarks but the change appears fundamentally right to me.

I do wonder whether this needs a migration guide entry but implementors of `PyTypeInfo` are probably not that common in user code?

src/conversion.rs
davidhewitt: Has no functional change, it was just a precursor to #3601 where it makes it easy to slap `#[allow(deprecated)]` on it. I can revert if preferred.

src/err/mod.rs
davidhewitt: Agreed this was probably unnecessary, I'll revert and confirm that `codspeed` still thinks the diff is fine. I think the `inline` hints had a much greater effect than this.

I noticed recently that `#[cold]` doesn't prevent inlining and PGO is likely to do better than manual branch hints, so I suspect we should rarely if ever use `#[cold]` to be honest 👍

davidhewitt:
Thanks for the review; I will fixup all points hopefully tomorrow. I think also there is no harm in a small migration note, we can state in it that most users will not care about the difference, though it may affect imports occasionally I guess.

src/conversion.rs
adamreichold: No need. I am just making sure I am not missing something here.

davidhewitt:
Codspeed seems fine after removing `#[cold]`, and the only other changes since review are docs changes so I'll proceed to merge this and rebase other stuff.

PR #3601, opened 2023-11-27
deprecate `PyTryFrom` and `PyTryInto`

### LINKED ISSUE #3516: PyTryFrom conflicts with TryFrom

davidhewitt:
This is a follow-up to #3600 (first commit is the same) which goes further and deprecates `PyTryFrom` and `PyTryInto`.

Closes #3516


guide/src/migration.md
```diff
@@ -15,7 +15,7 @@ To migrate, switch all type casts to use `obj.downcast()` instead of `try_from(o
 
 Before:
 
-```rust
+```rust,compile_fail
```
adamreichold, line 18: I think adding `# #[allow(deprecated)]` would be nicer.

newsfragments/3601.changed.md
```diff
@@ -0,0 +1 @@
+Deprecate `PyTryFrom` and `PyTryInto` traits.
```
adamreichold, line 1: This should mention their replacement by `PyTypeCheck` and `PyTypeInfo`.

guide/src/migration.md
davidhewitt: Sure thing, done 👍

src/instance.rs
```diff
@@ -1642,7 +1642,9 @@ a = A()
         }
 
         #[test]
+        #[allow(deprecated)]
```
adamreichold, line 1645: Maybe we want to adjust this test to use form we actually using now?

newsfragments/3601.changed.md
davidhewitt: Good idea, I also explicitly suggested `any.downcast()` as that's how users will typically engage with them.

adamreichold (approved):

src/instance.rs
adamreichold: I guess the question is whether this meant to cover (the now deprecated) trait impl, or the use case of getting a `PyCell<T>` from a `PyAny`.

davidhewitt: Yep, it's for coverage of the `PyTryFrom for PyCell` implementation, so I'll keep it in this form.

PR #3604, opened 2023-11-29
Refactor create_type_object so that most of the code is monomorphic

alex:
In pyca/cryptography this function is the #1 source of lines of generated LLVM IR, because it is duplicated 42x (and growing!). By rewriting it so most of the logic is monomorphic, we reduce the generated LLVM IR for this function by 4x.


davidhewitt (approved):
Nice, I had been concerned when it switched to the builder style undoing some of my previous code size reduction efforts so happy to have this re-optimised 👍

PR #3606, opened 2023-11-29
"full" implementation of new `Py2` API

### LINKED ISSUE #3382: Upcoming changes to PyO3's API in 0.21

davidhewitt:
Closes #3382 

This branch has been used to implement https://github.com/pydantic/pydantic-core/pull/1085 

It's based on top of #3572, and also includes #3600 as well as updated forms of most of the other PRs introducing Py2 types.

I feel like this strikes a good balance of introducing the new `Py2` API without breaking much of the existing gil refs API. The only thing which really needed to change was the `FromPyObject` trait, to take `&Py2<'py, PyAny>` instead of `&'py PyAny`. I don't think I'd want to land this amongst other breaking changes. I do think this might be releasable as an 0.22 release with both APIs still present (maybe gate the GIL refs API with a `backcompat` feature), with a view to dropping the gil refs API in 0.24 maybe.

Next steps:

- We want to release `pydantic-core` against a form of this branch in the near future. Any opinions if I should release this as a fork of PyO3, or we release this as an 0.22 alpha?

- I plan to start splitting this out into reviewable PRs and keep this one rebased as we move forwards.

- This also needs further cleanup / documentation etc.

### DIFF
--- a/guide/src/class.md
+++ b/guide/src/class.md
@@ -1253,7 +1253,7 @@ impl<'a, 'py> pyo3::impl_::extract_argument::PyFunctionArgument<'a, 'py> for &'a
     type Holder = ::std::option::Option<pyo3::PyRef<'py, MyClass>>;
 
     #[inline]
-    fn extract(obj: &'py pyo3::PyAny, holder: &'a mut Self::Holder) -> pyo3::PyResult<Self> {
+    fn extract(obj: pyo3::impl_::extract_argument::PyArg<'py>, holder: &'a mut Self::Holder) -> pyo3::PyResult<Self> {
         pyo3::impl_::extract_argument::extract_pyclass_ref(obj, holder)
     }
 }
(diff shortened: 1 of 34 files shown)


davidhewitt:
This branch has now outlived its usefulness. We've split out basically everything into other PRs, and I see no further need to keep trying to rebase this here for the few remaining odds and ends, so I'll close this. Great work everyone!

PR #3608, opened 2023-11-29
refactor: drop futures_util dependency

wyfo:
See https://github.com/PyO3/pyo3/issues/1632#issuecomment-1831365338

Draft based on #3599 

@davidhewitt Skip changelog?

### DIFF
--- a/src/coroutine.rs
+++ b/src/coroutine.rs
@@ -1,19 +1,17 @@
 //! Python coroutine implementation, used notably when wrapping `async fn`
 //! with `#[pyfunction]`/`#[pymethods]`.
 use std::{
-    any::Any,
     future::Future,
     panic,
     pin::Pin,
     sync::Arc,
-    task::{Context, Poll},
+    task::{Context, Poll, Waker},
 };
 
-use futures_util::FutureExt;
 use pyo3_macros::{pyclass, pymethods};
 
 use crate::{
-    coroutine::waker::AsyncioWaker,
+    coroutine::{cancel::ThrowCallback, waker::AsyncioWaker},
     exceptions::{PyAttributeError, PyRuntimeError, PyStopIteration},
     panic::PanicException,
     pyclass::IterNextOutput,
@@ -24,12 +22,9 @@ use crate::{
 pub(crate) mod cancel;
 mod waker;
 
-use crate::coroutine::cancel::ThrowCallback;
 pub use cancel::CancelHandle;
 
 const COROUTINE_REUSED_ERROR: &str = "cannot reuse already awaited coroutine";
-
-type FutureOutput = Result<PyResult<PyObject>, Box<dyn Any + Send>>;
 
 /// Python coroutine wrapping a [`Future`].
 #[pyclass(crate = "crate")]
@@ -37,7 +32,7 @@ pub struct Coroutine {
     name: Option<Py<PyString>>,
     qualname_prefix: Option<&'static str>,
     throw_callback: Option<ThrowCallback>,
-    future: Option<Pin<Box<dyn Future<Output = FutureOutput> + Send>>>,
+    future: Option<Pin<Box<dyn Future<Output = PyResult<PyObject>> + Send>>>,
     waker: Option<Arc<AsyncioWaker>>,
 }
 
@@ -68,7 +63,7 @@ impl Coroutine {
             name,
             qualname_prefix,
             throw_callback,
-            future: Some(Box::pin(panic::AssertUnwindSafe(wrap).catch_unwind())),
+            future: Some(Box::pin(wrap)),
             waker: None,
         }
     }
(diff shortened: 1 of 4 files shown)


alex:
thank you!

src/coroutine/cancel.rs
```diff
@@ -1,16 +1,16 @@
...
-use futures_util::future::poll_fn;
-use futures_util::task::AtomicWaker;
+use std::future::Future;
+use std::pin::Pin;
 use std::ptr;
 use std::ptr::NonNull;
 use std::sync::atomic::{AtomicPtr, Ordering};
-use std::sync::Arc;
-use std::task::{Context, Poll};
+use std::sync::{Arc, Mutex};
```
adamreichold, line 7: I think we use `parking_lot::Mutex` for our other internal mutex (in the `gil` module). I think it would improve consistency to use it here as well.

alex: FWIW, I've been sitting on a patch that drops parking_lot for our existing uses, by raising the MSRV to 1.60.

I think it's probably either practical now or will be very soon. (For pyca/cryptography, our next release will go to an MSRV of 1.63)

src/coroutine/cancel.rs
```diff
@@ -32,27 +34,25 @@ impl CancelHandle {
...
+            let ptr = self.0.exception.swap(ptr::null_mut(), Ordering::Relaxed);
+            // pointer cannot be null because it is checked the line before,
+            // and the swap is protected by `&mut self`
+            // SAFETY: only valid owned pointer are set in `ThrowCallback::throw`
+            return Poll::Ready(unsafe { Py::from_non_null(NonNull::new(ptr).unwrap()) });
         }
-        self.0.waker.register(cx.waker());
-        if self.is_cancelled() {
-            return Poll::Ready(take());
+        if !matches!(guard.deref(), Some(waker) if cx.waker().will_wake(waker)) {
```
adamreichold, line 47: I think this is too clever, i.e. checking whether the guard contains `Some` and whether the waker will be woken should be two nested if expressions.

wyfo: Do you want something like this?
```rust
if let Some(waker) = guard.deref() {
    if cx.waker().will_wake(waker) {
        return Poll::Pending;
    }
}
*guard = Some(cx.waker().clone());
Poll::Pending
```

adamreichold: Yes, exactly this. I understand it is more text, but parsing the guard expression inside a negated `matches!` clauses is just making this unnecessarily difficult to read IMHO.

One other alternative I see would be something like

```rust
if guard.deref().map_or(false, |waker| cx.waker().will_wake(waker)) {
  return Poll::Pending;
}
```

but even this seems less readable than the two-clause variant.

wyfo: done

wyfo:
I've rebased on the last version of #3599, it should be mergeable right after #3599 merge

adamreichold (approved):
Second commit looks good to me. The Clippy issues seem to stem from #3599

wyfo:
It seems to me that it's not a clippy issue but a benchmark issue (as written in the email I received). Clippy jobs are just cancelled because of benchmark failure (same in #3599).
However, I don't understand the error:
```
  INTERNALERROR>     from .pyo3_pytests import *
  INTERNALERROR> ModuleNotFoundError: No module named 'pyo3_pytests.pyo3_pytests'
```
I cannot reproduce it on my computer.

wyfo:
Should be better now, but there was a Codecov issue in the CI:
```
[2023-12-04T16:14:41.932Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```
@davidhewitt Do you want to relaunch coverage job?

src/coroutine.rs
```diff
@@ -1,21 +1,19 @@
 //! Python coroutine implementation, used notably when wrapping `async fn`
 //! with `#[pyfunction]`/`#[pymethods]`.
+use std::task::Waker;
```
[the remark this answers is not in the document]
wyfo: I will do the modification.
However, maybe we should consider using nightly *rustfmt* someday. For example, in all my projects, I add the following `.rustfmt.toml`:
```toml
group_imports = "StdExternalCrate"
imports_granularity = "Crate"
```

wyfo:
Previous pipeline showed that some tests was not deterministic, so I replace 1s sleeps by 999s.

src/coroutine.rs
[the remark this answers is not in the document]
davidhewitt: My concern would be that this makes it harder for contributors to use a stable toolchain if it's trying to conflict with the nightly rustfmt settings, but I haven't explored this. We could certainly consider it.

PR #3609, opened 2023-11-30
feat: allow async fn to accept receiver

wyfo:
Relates to #1632 

Draft based on #3608


wyfo:
As previously mentioned in https://github.com/PyO3/pyo3/pull/3540#issuecomment-1817700248, there is a design decision to be discussed here.
`&self`/`&mut self` are not available yet for async methods because of the `Send + 'static' requirement. In fact, even if you capture `Py<Self>` in the future, you cannot hold `PyCell<Self>`/`PyRef<Self>`/`PyRefMut` across future suspension points, because they are not `Send`. It's still possible to hack the requirement by actually capturing `Py<Self>`, but playing with borrowing manually. But we need to define the borrowing semantic for async method receiver, and there are three possibilities:
- borrow the cell at the coroutine instantiation, and release it when it's dropped
- borrow the cell the first time the coroutine is polled, and release it when it's dropped
- borrow the cell each time the coroutine is polled, and release it at the end of the poll.

In this PR, I've implemented the first one, as illustrated in the related test, because I think it's the simpler to reason about. 

Another solution could be to expose `Send` equivalent to `PyRef`/`PyRefMut`, to give more flexibility to users.

wyfo:
It seems my current implementation has a big MSRV issue, because it only works in 1.74 :thinking: 

I don't find anything in 1.74 changelog that would explain the error observed in the CI [job](https://github.com/PyO3/pyo3/actions/runs/7092188891/job/19302954781?pr=3609). I will investigate more.

davidhewitt:
@wyfo I got a bit lost with the recent opening of several PRs, this is the only one not in draft, are you waiting for reviews on any of these (or this one)?

wyfo:
Indeed, sorry, I should have made it more explicit. 
In #3540, I wrote:
> So here is how I see the next steps:
> - Add the missing documentation, make Coroutine::new private and merge the PR. Also answer the question about async_support branch.
> - I will have to retro-engineer my branch to split it in independent PR. It might not be easy for some parts, but that's too bad for me, I should have work in a more structured way instead of writing everything at the same time 😅 I will have to order them to avoid conflicts. This order could be as following
> -__name__/__qualname__ + not-awaited warning
>- #[pyo3(cancel_handle = "param")]
> - &self/&mut self
> -#[pyo3(allow_threads)]
> - PyFutures
> - generic waker with multiple Python runtime support; for this one I have an issue about allowing the user to define a default runtime at the crate level, and I'm asking your help on it (I will open the discussion in the dedicated PR).
> - Expose Coroutine constructor; that will mark the achievement of the design.
>
> Tell me if you want to change the order. I will open all the PR at the same time, with each one based on the previous one in the defined order. This way, you will be able to review all the PR independently, as long as they are merged in order.

That's what I've done, but PR based on a previous one as draft, because they can't be merged before their base PR.
So as long the final #3613 is not merged, you should always have at least one PR to review.
 In fact, having multiple PR allows parallel reviews of distinct and quite unrelated features – of course, if you have the time :sweat_smile: – as it was the case for #3599 and #3608.

src/impl_/coroutine.rs
```diff
@@ -16,3 +24,67 @@ where
 {
     Coroutine::new(Some(name.into()), qualname_prefix, throw_callback, future)
 }
+
+fn get_ptr<T: PyClass>(obj: &Py<T>) -> *mut T {
+    // SAFETY: Py<T> can be casted as *const PyCell<T>
+    unsafe { &*(obj.as_ptr() as *const PyCell<T>) }.get_ptr()
+}
+
+pub struct RefGuard<T: PyClass>(Py<T>);
```
davidhewitt, line 33: To simplify the implementation, I wonder if this should be

```suggestion
pub struct RefGuard<T: PyClass>(ManuallyDrop<PyRef<'static, T>>);
```

guide/src/async-await.md
```diff
@@ -30,8 +30,7 @@ Resulting future of an `async fn` decorated by `#[pyfunction]` must be `Send + '
 
 As a consequence, `async fn` parameters and return types must also be `Send + 'static`, so it is not possible to have a signature like `async fn does_not_compile(arg: &PyAny, py: Python<'_>) -> &PyAny`.
 
-It also means that methods cannot use `&self`/`&mut self`, *but this restriction should be dropped in the future.*
-
+However, there is an exception for method receiver, so async methods can accept `&self`/`&mut self`
```
davidhewitt, line 33: Is it worth pointing out here that there's a good case for being extra careful with `&mut self` as a method receiver in a future? I think it's very likely users will accidentally hit borrow errors due to overlapping futures.

We could potentially restrict this to `&self` on frozen classes? That would help prevent users creating aliasing borrow errors from async code. What do you think about that restriction?

davidhewitt (commented):
Great! Review posted, sorry for being generally slow :)

src/impl_/coroutine.rs
wyfo: If you want to use `PyRef` directly, you should not use `ManuallyDrop`, because you need to release the borrow with `PyRef` destructor.
Also, `PyRef` doesn't own the object, so you have to also capture `Py<T>` in the future, to prevent the object dealloacation. And you have to be careful to the drop order, because if `Py<T>` is dropped before `PyRef`, the associated object could be deallocated before running `PyRef` destructor, and you could crash.
Last but not least, `PyRef` is not `Sent`, so you have to implement `Send` manually for `RefGuard`.

All in all, I'm not sure that's simpler, especially taking in account the crash potential with the destructor order.
I've just rewritten `RefGuard`/`RefMutGuard` constructors which was overly complex.

guide/src/async-await.md
wyfo: > We could potentially restrict this to &self on frozen classes?

Honestly, I prefer to just add a warning in the documentation. I don't want to prevent careful users to use a `mpsc::Receiver` in an efficient way. 
Maybe you could ping another maintainer to discuss this decision?

src/impl_/coroutine.rs
davidhewitt: All great points why my suggestion is no good 👍

guide/src/async-await.md
davidhewitt: Ok let's keep this. I think what I'd prefer to do is make frozen the default and then document what the risks of `&mut` are if users choose to enable this.

adamreichold:
> there are three possibilities:
> 
>     * borrow the cell at the coroutine instantiation, and release it when it's dropped
> 
>     * borrow the cell the first time the coroutine is polled, and release it when it's dropped
> 
>     * borrow the cell each time the coroutine is polled, and release it at the end of the poll.
> 
> 
> In this PR, I've implemented the first one, as illustrated in the related test, because I think it's the simpler to reason about.

I am obviously late to the party, but I think there is a reason besides simplicity for choosing the first option: I think it also matches the behaviour of plain Rust async methods, i.e. writing `async fn incr(&mut self)` will desugar into something like `fn incr<'a>(&'a mut self) -> impl Future<Output=()> + 'a` where the borrow lives as long as the future does.

In any case, if I had fast enough to join the fray, I would have asked for the guide entry to expand on this, i.e. to explicitly call out how long the method receiver borrow lasts and that this is natural transfer of the existing semantics of async methods.

jopemachine:
@wyfo Thanks for your great work.

May I have a question?
I got the below compile error when I tried to specify some argument (`num`) with the async method.
Is there a way to avoid this error?

```
future cannot be sent between threads safely
within `pyo3::PyAny`, the trait `Sync` is not implemented for `UnsafeCell<pyo3::ffi::PyObject>`
```


```rust
    #[pymethods]
    impl Counter {
        #[new]
        fn new() -> Self {
            Self(0)
        }
        async fn get(&self) -> usize {
            self.0
        }
        async fn incr(&mut self) -> usize {
            self.0 += 1;
            self.0
        }

        // Error occurs here because of `num`
        async fn incr_num(&mut self, num: usize) -> Result<usize, Error> {
            self.0 += num;
            Ok(self.0)
        }
    }
```

davidhewitt:
@jopemachine let's split that out into a new discussion. Probably the "argument extraction" code for `num` is being included in the future, but that's blocking code so we can potentially rearrange the macro output.

PR #3610, opened 2023-11-30
feat: add #[pyo3(allow_threads)] to release the GIL in (async) functions

wyfo:
Relates to #1632 

Draft based on #3609


wyfo:
@davidhewitt Here is the next one! I'm more explicit now :smile:

src/coroutine.rs
```diff
@@ -96,7 +99,13 @@ impl Coroutine {
         let waker = Waker::from(self.waker.clone().unwrap());
         // poll the Rust future and forward its results if ready
         // polling is UnwindSafe because the future is dropped in case of panic
-        let poll = || future_rs.as_mut().poll(&mut Context::from_waker(&waker));
+        let poll = || {
+            if self.allow_threads {
+                py.allow_threads(|| future_rs.as_mut().poll(&mut Context::from_waker(&waker)))
```
davidhewitt, line 104: Doesn't this now violate the safety promise on line 61?

wyfo: Indeed, and that's a big mistake of mine, I've fixed it.

wyfo: Actually, this is were the PR start to diverge from my last POC, in which I had boxed the entire coroutine instead of only the future, so my retro-engineering may have some issues. Fortunately, there is a review :sweat_smile:

src/coroutine.rs
```diff
@@ -26,13 +26,44 @@ pub use cancel::CancelHandle;
...
+        allow_threads: bool,
+    ) -> Poll<PyResult<PyObject>> {
+        let result = if allow_threads {
+            py.allow_threads(|| self.poll(&mut Context::from_waker(waker)))
+        } else {
+            self.poll(&mut Context::from_waker(waker))
+        };
+        result.map_ok(|obj| obj.into_py(py)).map_err(Into::into)
+    }
+}
```
davidhewitt, line 57: 👍 lovely solution!

davidhewitt (approved):
Excellent!

adamreichold:
@davidhewitt I have a few minutes now and the merge queue failed. If we could hold off on merging I could have a look as well?

tests/ui/invalid_pyfunction_signatures.stderr
```diff
@@ -16,7 +16,7 @@ error: expected argument from function definition `y` but got argument `x`
 13 | #[pyo3(signature = (x))]
    |                     ^
 
-error: expected one of: `name`, `pass_module`, `signature`, `text_signature`, `crate`
+error: expected one of: `allow_threads`, `name`, `pass_module`, `signature`, `text_signature`, `crate`
```
adamreichold, line 19: What I would really want to have here is a test that checks that I cannot get `py: Python<'_>` as my argument when using `allow_threads`. This should not be possible to do as the GIL is released, right?

davidhewitt:
Sure thing, I'll leave this for you to review and merge 👍

tests/ui/invalid_pyfunction_signatures.stderr
adamreichold: (Ideally, it would also have a nice error message explaining why these two contradict each other.)

wyfo: You would have a compilation error, because `Python` is not `Ungil`, but you're right, I will add that case in the UI tests.

src/coroutine.rs
```diff
@@ -26,13 +26,44 @@ pub use cancel::CancelHandle;
 
 const COROUTINE_REUSED_ERROR: &str = "cannot reuse already awaited coroutine";
 
+trait CoroutineFuture {
```
adamreichold, line 29: I am not sure which role the trait plays here? Why not just have a free generic `poll` function?

src/coroutine.rs
```diff
@@ -26,13 +26,44 @@ pub use cancel::CancelHandle;
...
+    E: Into<PyErr> + Send,
+{
+    fn poll(
+        self: Pin<&mut Self>,
+        py: Python<'_>,
+        waker: &Waker,
+        allow_threads: bool,
+    ) -> Poll<PyResult<PyObject>> {
+        let result = if allow_threads {
+            py.allow_threads(|| self.poll(&mut Context::from_waker(waker)))
```
adamreichold, line 51: Is it important the the context is constructed without the GIL being held? Could this factored out into a binding to make it more obvious that this is the same between both cases, e.g. `let cx = &mut Context::from_waker(waker);`?

tests/ui/invalid_pyfunction_signatures.stderr
wyfo: As expected, there is a compilation error, but the output will not make @davidhewitt happy :sweat_smile:

src/coroutine.rs
wyfo: I need a trait to box it.

src/coroutine.rs
wyfo: `Context` is explicitely not `Send`

src/coroutine.rs
```diff
@@ -26,13 +26,44 @@ pub use cancel::CancelHandle;
...
+where
+    F: Future<Output = Result<T, E>> + Send,
+    T: IntoPy<PyObject> + Send,
+    E: Into<PyErr> + Send,
+{
+    fn poll(
+        self: Pin<&mut Self>,
+        py: Python<'_>,
+        waker: &Waker,
+        allow_threads: bool,
```
adamreichold, line 48: This might be me going off into the weeds, but we turning a compile time information into a runtime flag here and generating code to handle both cases even only one will ever be executed.

Hence, I wonder if we should use a const generic `const ALLOW_THREADS: bool` on `Coroutine` to ensure that only the required code is generated?

src/coroutine.rs
wyfo: To be more precise, I could not use `Future` anymore because I had to apply `IntoPy` to the `Future` result, see https://github.com/PyO3/pyo3/pull/3610#discussion_r1419234657

src/coroutine.rs
wyfo: In fact, that's what I have in my POC. I've made the whole `Coroutine` generic and boxed it in the `#[pyclass]`.
But I decide to go for simplicity here. I can change that if you think the performance impact will be significant.

tests/ui/invalid_pyfunction_signatures.stderr
wyfo: You can add an error message for `Python`, but not really for things like `&PyAny`

src/coroutine.rs
adamreichold: Indeed. Thinking about the alternatives, this does appear to be a very nice solution. I would suggest having `: Send` as a bound for the trait itself though to make it more obvious why the blanket impl includes that bound as well.

tests/ui/invalid_pyfunction_signatures.stderr
adamreichold: > You can add an error message for Python, but not really for things like &PyAny

Indeed, but I think we should do what we can. Sorry for putting more work on your table.

wyfo:
The issue with the merge queue was surely the `async fn` test added in `gil.rs`. I've annotated it with `#[cfg(not(target_arch = "wasm32"))]`

src/coroutine.rs
wyfo: done

tests/ui/invalid_pyfunction_signatures.stderr
wyfo: Not an issue, it will just wait a few hour but I will do the modification :smiley:

src/coroutine.rs
adamreichold: I think simplicity is certainly good for the first iteration and this could definitely be a follow-up PR.

That said, I was wondering if we could hide this behind the boxed `CoroutineFuture` trait to avoid generics all the way up, e.g. a marker wrapper like

```rust
struct AllowThreads<F>(F);

impl<F, T, E> CoroutineFuture for AllowThreads<F>
where
    F: Future<Output = Result<T, E>> + Send,
    T: IntoPy<PyObject> + Send,
    E: Into<PyErr> + Send;
```

but I am not sure how well this can be tucked away to avoid complicating the instantiation too much.

tests/ui/invalid_pyfunction_signatures.stderr
adamreichold: (Just for coverage, we should also duplicate the compiler error test for plain and async functions.)

adamreichold:
One thing I only thought about now: How does this interact with async methods taking `&self/&mut self`. That should also be incompatible with `allow_threads`, shouldn't it?

adamreichold:
> One thing I only thought about now: How does this interact with async methods taking `&self/&mut self`. That should also be incompatible with `allow_threads`, shouldn't it?

To add insult to injury, we should also consider how this would interact with #3646.

adamreichold:
> (My code might be off but hopefully the point is communicated.)

Ah, that makes sense! The interior of the pyclass is not bound to the GIL, just getting access to it is. (Similar to how getting an `ArrayView` out of a NumPy array requires the GIL, but the view can then be process within `allow_threads`.)

> I wonder if this would imply that the future needs to be polled on the worker thread? Hmm.

I think this would be the default outcome if both MR were merged as is. This should be fine, i.e. the `Send` bound already exists and hence `allow_threads` can soundly do this. I mainly wonder whether performance is show stopper for these short-lived closure which only want to poll the future once and then move on. Maybe this would require some deeper integration like having the future always live on the worker thread and just send a request to poll it or something like that.

tests/ui/invalid_pyfunction_signatures.stderr
wyfo: I've added a check for `Python<'_>` field for both `allow_threads` and async functions. The error raised by `Bound<'_, PyAny>` for these cases is still ugly, but as said earlier, I don't think it's possible to add a pretty error message, and this is already the case for `Python::allow_threads` tests in *ui/not_send.rs* and *ui/not_send2.rs*.

src/coroutine.rs
wyfo: Done.

wyfo:
As said in #1632, i'm back! @adamreichold @davidhewitt

wyfo:
I don't know what I can do for the nightly pipeline error 🤷‍♂️ 
```
error: non-local `impl` definition, they should be avoided as they go against expectation
```

mejrs:
> I don't know what I can do for the nightly pipeline error 🤷‍♂️
> 
> ```
> error: non-local `impl` definition, they should be avoided as they go against expectation
> ```

Feel free to allow the lint for now. You'll have to chase down where our code triggers this and apply `#[allow(non_local_definitions]` (and a TODO) there.

It's a new nightly lint, it's not something you did. We have done work earlier to adapt our code to this lint, but this part is new. You can fix it if you want but the last time it was a substantial amount of work so it should be fixed in its own PR.

wyfo:
> You'll have to chase down where our code triggers this

That's the issue, it's not my code which is triggering the lint. My changes have nothing to do with the following error
```
error: non-local `impl` definition, they should be avoided as they go against expectation
   --> tests/test_proto_methods.rs:643:1
    |
643 | #[pymethods]
    | ^^^^^^^^^^^^
    |
    = help: move this `impl` block outside the of the current function `trampoline` and up 3 bodies
```

It's an issue of generated code (parts I haven't touched). My explaination is that a bug has been very recently introduced in the nightly build. 
By the way #4052 is failing too for the same reasons, while it's just a typo correction.  There is definitely an issue with the nightly CI.

mejrs:
You can also wait for someone to send a PR (probably me tomorrow) and rebase after it is merged.

wyfo:
#4033 broke some code here, I will update the PR.
Could I ask then for a new review? @adamreichold? I would like to unblock the async PRs to close the async chapter for the next release.

Icxolu (commented):
Generally LGTM, but this is definitely a new area for me, so someone with more experience here should have a look as well.

We might additionally want a test to verify that for (`async`) `allow_threads` _methods_ only `&self`/`&mut self` receivers are allowed and `Bound`/`PyRef`/`PyRefMut` are rejected.

pyo3-macros-backend/src/method.rs
```diff
@@ -676,20 +695,21 @@ impl<'a> FnSpec<'a> {
...
-                            #(#args),*
-                        )
-                    }
-                }
+                    (quote!(__self,), quote! { let (__self,) = (#self_arg); })
+                };
+                quote! {{
+                    #self_arg_decl
+                    #(let #arg_names = #args;)*
+                    py.allow_threads(|| function(#self_arg_name #(#arg_names),*))
```
davidhewitt, line 708: There's a question in my head here about after #3646 which of the `allow_threads` variants we will use here (cc @adamreichold). I think it would be possible to create a scoped TLS issue by setting up a TLS scope and then calling into Python. Similar thinking applies to the async case below. I guess we have to stick to the principle that the soundness is important and use the "safe but slow" form for this attribute. If users want to go `unsafe` they can release the GIL themselves. This feels the right way to do it, I think.

adamreichold: > This feels the right way to do it, I think.

Yes, I think so too. Especially since we have already accumulated a significant amount of high risk changes for 0.22 and we can always optimize things in following releases.

davidhewitt:
@wyfo I may try to rebase and merge this tomorrow at the PyCon sprints. Any objections?

itamarst:
If anyone has time to work on this, it could make a PR I need to write have rather smaller diffs :grin:

Icxolu:
@davidhewitt If we're good with including this in 0.23 I could go rebase this and fixup the last final remarks from above.

davidhewitt:
I would love to have this, but I am worried about interactions with #3646. I think merging this before that PR is potentially a problem (might have breakage), and I am worried that PR is necessary but also makes async extremely complicated (because that PR basically disables all TLS with `allow_threads`, but the `tokio` runtime is frequently accessed by TLS, for example). I haven't had a chance to sit down and reason through all of this myself, however I would be happy to hand over for someone else to evaluate the design space and understand how bad the interaction between #3646 and async is.

itamarst:
(My _personal_ use case is non-async functions, so I'd be happy with this feature to land with initially just supporting normal functions. But I do see why this would be much more of a useful thing for async functions.)

mejrs:
The interaction between async runtimes, threadlocals and #3646 remains whether or not this PR is merged. Is your concern that we'll have more fallout if we merge this and then that?

wyfo:
@davidhewitt  Why does the PR disable all TLS?

mejrs:
> @davidhewitt Why does the PR disable all TLS?

Merging that PR means that allow_threads runs its closures on a threadpool, so it's executed on a different thread. It'd be surprising to people who are accessing threadlocals from it.

davidhewitt:
> The interaction between async runtimes, threadlocals and #3646 remains whether or not this PR is merged. Is your concern that we'll have more fallout if we merge this and then that?

Yes, I still feel like we should aim to do something like #3646 but the more I look at it the more extreme the fallout seems to be. Merging this seems to encourage use of `allow_threads` by making it really cheap to slap as an attribute on basically everything. Which I do think is probably a good thing, but if #3646 then goes and breaks all those, I'm worried about getting the order of these additions right.

itamarst:
> Merging that PR means that allow_threads runs its closures on a threadpool, so it's executed on a different thread. It'd be surprising to people who are accessing threadlocals from it.

As a user I would find the use of a _thread pool_ very surprising and it would make `#[pyo3(allow_threads)]` into something I probably can't use. As context I'm looking at using this in Polars Python APIs, that all need to release the GIL before calling into Polars' internal thread pool (otherwise you get deadlocks). `#[pyo3(allow_threads)]` resulting in dispatching to a completely different thread pool first would I suspect be undesirable to Polars core devs.

(From my personal perspective the thread pool option seems very intrusive and in my own projects I would likely default to the unsound-if-you-put-Python-objects-into-TLS version that runs in current thread.)

More practically for this PR: there could be two different attributes, one for each proposed allow_threads strategy?

wyfo:
Could https://doc.rust-lang.org/reference/unsafe-keyword.html#unsafe-attributes-unsafeattr be a thing?

davidhewitt:
Yes, though it's still in RFC and we might want to see how that works out: https://github.com/rust-lang/rfcs/pull/3715

wyfo:
@davidhewitt I've just realized that this feature shouldn't be an issue regarding TLS smuggling. Indeed, functions annotated with `#[pyfunction(allow_threads)]` are called from Python, and the `allow_threads()` call is done in the macro generated code, with no possibility to set a TLS before.
Better, annotate a function with `#[pyfunction(allow_threads)]` would be a safe way to release the GIL without needing to use a different thread.
Did I miss something?

itamarst:
You have a stack that goes back and forth between Rust and Python, it isn't necessarily just Python → Rust:

Python function → Rust function that stores something in TLS → Python function → Rust function annotated with `#[pyfunction(allow_threads)]` which reads from TLS

davidhewitt:
Yes, it is exactly that kind of mixed language call stack which is my worry.

I really want this feature, and I _think_ the practical way forward is probably to use the `unsafe` attribute as you suggest, though the RFC is in limbo.

I will be speaking at RustNation in a few weeks and will be bringing up this TLS pain as part of my talk, so maybe it will spur some bright minds in the Rust ecosystem to help us figure out new ways forward we haven't yet seen.

wyfo:
Indeed, sorry about the noise.

Anyway, let's recap for once why this PR exists. We don't care about synchronous function here, as I supported them only for the symmetry with asynchronous functions. The last one needed a special support, because to release the GIL at each poll, you need to use the unsafe but guaranteed assumption that the GIL is acquired when `poll` is called. So it was to avoid the user to type the `unsafe` keyword.
However, now that `allow_threads` become unsafe, we no longer care about saving this `Python::assume_gil_acquired()` to the user, has `unsafe` become inevitable. So it's up to the user to wrap his future in a newtype doing `unsafe { Python::assume_gil_acquired().allow_threads_unsafe(/* call poll */) }`. If the user wants to avoid unsafe with the `allow_threads_safe`, then the cost of the safe `Python::with_gil` becomes negligible anyway compared to using a second thread.

Therefore, I will simply close this PR.
I've a bit of time now, so I will try get back to the rest of the PR series, to finally have the full async support merged!

davidhewitt:
Ok, makes sense and awesome to hear you have some capacity!

I must apologise in advance as I am going extremely slowly recently and seeing be back to full capacity until March. But I will do my best to review anything you push, async support got stuck for too long.

Since we last collaborated on this I think I saw new `Waker` APIs got stabilised for data, so maybe new optimisations are possible. We also had to bring `pyo3-asyncio` into the PyO3 org as `pyo3-async-runtimes`.

wyfo:
> Since we last collaborated on this I think I saw new Waker APIs got stabilised for data, so maybe new optimisations are possible.

It has indeed, so I could use that to save me my dirty [TLS trick](https://github.com/PyO3/pyo3/pull/3611/files#diff-47d11c3256f338267bc0c43e46e150d0b10530514f5354b7cfb328258004cdb2R105) in the next PR. However, it would imply to radically bump the MSRV to 1.83, so I don't think we want to do that for now. However, I will let a TODO as a reminder for the next bump.

PR #3616, opened 2023-11-30
Enable `GILProtected` access via `PyVisit`

### LINKED ISSUE #3615: Access `GILProtected` via `PyVisit`
Since [you can't access a `Python<'_>` from within `__traverse__`][1], it becomes impossible to access a `GILProtected` object while traversing, which causes problems if there is an object that needs to be traversed somewhere inside the `GILProtected`. Seeing as [`GILProtected::get` doesn't actually do anything][2], this should be a valid operation despite the restrictions imposed during GC.

Rather than rehashing the discussions from previous issues regarding marker access during GC, it seems relatively straight-forward to simply allow access to the inner object using a `PyVisit` object.

[1]: https://github.com/PyO3/pyo3/pull/3168
[2]: https://docs.rs/pyo3/latest/src/pyo3/sync.rs.html#37-39

neachdainn:
Closes #3615

This simply adds a new method which uses the existence of a `PyVisit` object as proof that the GIL is held instead of a `Python` object. This allows `GILProtected` to be used in instances where contained Python objects need to participate in garbage collection. Usage in this situation should be valid since no Python calls are made and this does not provide any additional mechanism for accessing a `Python` object.

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -1,5 +1,5 @@
 //! Synchronization mechanisms based on the Python GIL.
-use crate::{types::PyString, types::PyType, Py, PyErr, Python};
+use crate::{types::PyString, types::PyType, Py, PyErr, PyVisit, Python};
 use std::cell::UnsafeCell;
 
 /// Value with concurrent access protected by the GIL.
@@ -35,6 +35,11 @@ impl<T> GILProtected<T> {
 
     /// Gain access to the inner value by giving proof of having acquired the GIL.
     pub fn get<'py>(&'py self, _py: Python<'py>) -> &'py T {
+        &self.value
+    }
+
+    /// Gain access to the inner value by giving proof that garbage collection is happening.
+    pub fn traverse<'py>(&'py self, _visit: PyVisit<'py>) -> &'py T {
         &self.value
     }
 }
--- a/newsfragments/3616.added.md
+++ b/newsfragments/3616.added.md
@@ -1,0 +1,1 @@
+Add `traverse` method to `GILProtected`


neachdainn:
I am 100% willing to change the name of the method. I just wanted the MR to go with the issue ASAP so both can be closed quickly (assuming there is little to no need to discuss the feature).

davidhewitt:
I'm feeling split about this feature. I understand why in practice this seems ok at the moment, but can we really rely on the fact that the GC will never traverse the graph in parallel?

Can you give some use cases of what you need this for? With the advent of nogil, we sort of need to be killing `GILProtected` so perhaps we can work out alternative patterns...

neachdainn:
>  can we really rely on the fact that the GC will never traverse the graph in parallel?

If we can't, then PyO3 already has a major issue around marking types as `unsendable`. That being said, I think we can based on the fact that making the garbage collector parallel would be a major happening in Python and would probably break a fair number of existing extensions.

> Can you give some use cases of what you need this for? 

Any time you have any Python object behind a `GILProtected`, you would need this in order to enroll in GC correctly. My specific use case is I am implementing something similar to a MPMC channel and I want to leverage the fact that the GIL is present in order to not have to pull in another mutex or deal with atomics. The vastly oversimplified version looks something like this:

```rust
struct Inner {
    items: VecDeque<PyObject>,
}

#[pyclass]
struct Sender {
    inner: Arc<GILProtected<RefCell<Inner>>>,
}

#[pymethods]
impl Sender {
    fn send(&self, py: Python<'_>, obj: PyObject) {
        self.inner.get(py).borrow_mut().items.push_back(obj);
    }
    
    fn __traverse__(&self, visit: PyVisit<'_>) -> Result<(), PyTraverseError> {
        // unable to traverse objects in the queue
    }
    
    fn __clear__(&mut self, py: Python<'_>) {
        self.inner.get(py).borrow_mut().items.clear();
    }
}
```

> With the advent of nogil, we sort of need to be killing `GILProtected` so perhaps we can work out alternative patterns...

I'll need to read up on the nogil but I'm not opposed to alternative patterns. I mostly want these changes so I can remove instances of `self.inner.get(Python::assume_gil_acquired())` from my code.

src/sync.rs
```diff
@@ -37,6 +37,11 @@ impl<T> GILProtected<T> {
     pub fn get<'py>(&'py self, _py: Python<'py>) -> &'py T {
         &self.value
     }
+
+    /// Gain access to the inner value by giving proof that garbage collection is happening.
+    pub fn get_visit<'py>(&'py self, _visit: PyVisit<'py>) -> &'py T {
```
adamreichold, line 42: I am on board with the concept but would suggest `traverse` as the method name.

adamreichold (approved):

src/sync.rs
davidhewitt: I think `traverse` is an excellent name here 👍

davidhewitt:
Fair points! Good thought about `unsendable` too, I vaguely recall hearing that the GC runs in a dedicated thread on nogil so we might indeed have a problem there, but that's a different discussion.

I think for practicality's sake I'm happy to move forward here with the name `traverse`, and when we start dealing with nogil (hopefully on my target for early in the new year) we can figure out the replacements. I suspect that `GILProtected<RefCell<T>>` will be best served by something like `RwLock<T>`.

adamreichold:
Please also fix the reference to the original method name in the changelog entry. Thanks!

neachdainn:
> Please also fix the reference to the original method name in the changelog entry. Thanks!

Good catch. Done!

PR #3622, opened 2023-12-03
Add additional definitions for `_PyImport_Frozen*`

messense:
These were added in https://github.com/python/cpython/commit/074fa5750640a067d9894c69378a00ceecc3b948, useful in embedding Python interpreter.


davidhewitt (approved):
Looks fine to me to have these 👍

PR #3626, opened 2023-12-04
ci: don't isolate venv for benches

davidhewitt:
Hopefully this can help fix the build errors we are observing in the benchmarks.


davidhewitt:
Clearly not the solution.

PR #3627, opened 2023-12-04
try to debug codspeed issue

davidhewitt:
...

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -36,7 +36,7 @@ jobs:
         run: pip install nox
 
       - name: Run the benchmarks
-        uses: CodSpeedHQ/action@v1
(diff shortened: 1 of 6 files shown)


davidhewitt:
Ok so I think I see what was going on, it looks like probably what was happening is that `pip install` was installing into the virtualenv and hitting https://github.com/PyO3/maturin/issues/490

I'm not sure why this wasn't a problem initially. I wonder if it was a cache issue related to the same problems we were having recently with rustup conflicts.

I removed the `pyo3_pytests/__init__.py` to use a pure-rust maturin project, which should fix the problem. I also took the opportunity to refactor as per #3536

davidhewitt:
🎉 looks like that's fixed it. I'll merge this as I think there's nothing much to review here and it improves CI for the rest of the reviews.

Issue #3226, opened 2023-06-11
Review conversions and ensure consistency


davidhewitt:
Continued from https://github.com/PyO3/pyo3/pull/3197#discussion_r1225910628

We would like to ensure that we're consistent in how conversions from Python objects to Rust objects select for what's allowed.

I think the right approach would be to do something like https://github.com/PyO3/pyo3/blob/main/guide/src/conversions/tables.md where for each Rust type we state what Python types will convert to it (i.e. `FromPyObject`), and what Python type(s) it will convert back into (i.e. `IntoPy<PyObject>`).

That information does live on that page, but perhaps it can benefit from being rearranged. 

In general we seem to be preferring conversions which are allow the user to be a bit loose with typing (e.g. string for IP address). So I think we should ensure we stick to this philosophy everywhere (e.g. maybe we should accept `int` for `bool`, or even all truthy types)?

We may also wish to consider a "strict" mode where only the exact Python type can be input (e.g. Python bool <-> Rust bool).

yusharth:
Can I take it?

davidhewitt:
Sure thing, please do! Is there any extra information you need to get started?

mikeleppane:
Hey! If this is still open and something that requires improvement, I would be interested to a take look. Hopefully, I could contribute more in the future.

davidhewitt:
Hello, thank you for your interest! Yes, this is absolutely still open to improvement.

First, I'd suggest starting with updating the conversions documentation in `tables.md` as I wrote in the original post above.

Second, if you're interested in going beyond collecting what's currently implemented to make things better, I think we would like to go down the path of having "default" conversions which allow a range of inputs and also a user-facing attribute to control when they want exactly a specified type.

I wrote about this topic most recently in https://github.com/PyO3/pyo3/pull/3638#issuecomment-1858583166

We called the option `#[pyo3(noconvert)]` in that PR.

PR #3638, opened 2023-12-10
Try harder by looking for a __bool__ magic method when extracing bool values from Python objects.

### LINKED ISSUE #3637: Error on extracting numpy boolean
In most circumstances numpy scalar types can be used as drop in replacements to Python's scalar types. PyO3 is almost no exception to this rule as it can successfully extract Rust types from many numpy types such as np.int32, np.float32, etc. However, for some reason it is not possible to extract a Rust boolean from a numpy boolean. Consider this example:

```rust
// lib.rs
use pyo3::prelude::*;

#[pyfunction]
fn bool_as_string(a: bool) -> PyResult<String> {
    Ok(a.to_string())
}

#[pyfunction]
fn int_as_string(a: i32) -> PyResult<String> {
    Ok(a.to_string())
}
#[pymodule]
fn my_module(_py: Python, m: &PyModule) -> PyResult<()> {
    m.add_function(wrap_pyfunction!(bool_as_string, m)?)?;
    m.add_function(wrap_pyfunction!(int_as_string, m)?)?;
    Ok(())
}
```

```python3
# test.py
import numpy as np
from my_module import bool_as_string, int_as_string

a = np.int32(42)
print(int_as_string(a))
print(bool_as_string(a > 0))
```

When running `python test.py`, the first print statement prints `42` as expected. The second print statement however raises an exception:

```
  File "/home/****/****/test.py", line 6, in <module>
    print(bool_as_string(a > 0))
TypeError: argument 'a': 'bool_' object cannot be converted to 'PyBool'
```

There is a simple workaround, as the numpy boolean can be easily converted to a Python boolean before calling the Rust function.

```python3
print(bool_as_string(bool(a > 0))) # works
```

However, it would still be nice if this issue could be fixed within PyO3 as it is hard to differentiate between Python booleans and numpy booleans as a developer in a project which heavily relies on numpy.

This was tested with the latest versions of PyO3 (0.20.0) and numpy (1.26.2).

adamreichold:
I decided to not implement the full protocol for [truth value testing](https://docs.python.org/3/library/stdtypes.html#truth) as it seems confusing in the context of function arguments if basically any instance of custom classes or non-empty collections turns into `true`.

Closes #3637

### DIFF
--- a/newsfragments/3638.changed.md
+++ b/newsfragments/3638.changed.md
@@ -1,1 +1,1 @@
-Values of type `bool` can now be extracted from all Python values defining a `__bool__` magic method.
+Values of type `bool` can now be extracted from NumPy's `bool_`.
--- a/newsfragments/3638.changed.md
+++ b/newsfragments/3638.changed.md
@@ -1,0 +1,1 @@
+Values of type `bool` can now be extracted from all Python values defining a `__bool__` magic method.
--- a/src/types/boolobject.rs
+++ b/src/types/boolobject.rs
@@ -81,6 +81,13 @@ impl<'source> FromPyObject<'source> for bool {
             return Ok(obj.is_true());
         }
 
+        let missing_conversion = |obj: &PyAny| {
+            PyTypeError::new_err(format!(
+                "object of type '{}' does not define a '__bool__' conversion",
+                obj.get_type()
+            ))
+        };
+
         #[cfg(not(any(Py_LIMITED_API, PyPy)))]
         unsafe {
             let ptr = obj.as_ptr();
@@ -95,14 +102,14 @@ impl<'source> FromPyObject<'source> for bool {
                 }
             }
 
-            Err(PyTypeError::new_err("object has no __bool__ magic method"))
+            Err(missing_conversion(obj))
         }
 
         #[cfg(any(Py_LIMITED_API, PyPy))]
         {
             let meth = obj
                 .lookup_special(crate::intern!(obj.py(), "__bool__"))?
-                .ok_or_else(|| PyTypeError::new_err("object has no __bool__ magic method"))?;
+                .ok_or_else(|| missing_conversion(obj))?;
 
             let obj = meth.call0()?.downcast::<PyBool>()?;
             Ok(obj.is_true())
@@ -174,13 +181,13 @@ class D:
             let c = module.getattr("C").unwrap().call0().unwrap();
             assert_eq!(
                 c.extract::<bool>().unwrap_err().to_string(),
-                "TypeError: object has no __bool__ magic method",
+                "TypeError: object of type '<class 'test.C'>' does not define a '__bool__' conversion",
             );
 
             let d = module.getattr("D").unwrap().call0().unwrap();
             assert_eq!(
                 d.extract::<bool>().unwrap_err().to_string(),
-                "TypeError: object has no __bool__ magic method",
+                "TypeError: object of type '<class 'test.D'>' does not define a '__bool__' conversion",
             );
         });
     }
--- a/src/types/boolobject.rs
+++ b/src/types/boolobject.rs
@@ -77,43 +77,52 @@ impl IntoPy<PyObject> for bool {
 /// Fails with `TypeError` if the input is not a Python `bool`.
 impl<'source> FromPyObject<'source> for bool {
     fn extract(obj: &'source PyAny) -> PyResult<Self> {
-        if let Ok(obj) = obj.downcast::<PyBool>() {
-            return Ok(obj.is_true());
-        }
-
-        let missing_conversion = |obj: &PyAny| {
-            PyTypeError::new_err(format!(
-                "object of type '{}' does not define a '__bool__' conversion",
-                obj.get_type()
-            ))
+        let err = match obj.downcast::<PyBool>() {
+            Ok(obj) => return Ok(obj.is_true()),
+            Err(err) => err,
         };
 
-        #[cfg(not(any(Py_LIMITED_API, PyPy)))]
-        unsafe {
-            let ptr = obj.as_ptr();
+        if obj
+            .get_type()
+            .name()
+            .map_or(false, |name| name == "numpy.bool_")
+        {
+            let missing_conversion = |obj: &PyAny| {
+                PyTypeError::new_err(format!(
+                    "object of type '{}' does not define a '__bool__' conversion",
+                    obj.get_type()
+                ))
+            };
 
-            if let Some(tp_as_number) = (*(*ptr).ob_type).tp_as_number.as_ref() {
-                if let Some(nb_bool) = tp_as_number.nb_bool {
-                    match (nb_bool)(ptr) {
-                        0 => return Ok(false),
-                        1 => return Ok(true),
-                        _ => return Err(crate::PyErr::fetch(obj.py())),
+            #[cfg(not(any(Py_LIMITED_API, PyPy)))]
+            unsafe {
+                let ptr = obj.as_ptr();
+
+                if let Some(tp_as_number) = (*(*ptr).ob_type).tp_as_number.as_ref() {
+                    if let Some(nb_bool) = tp_as_number.nb_bool {
+                        match (nb_bool)(ptr) {
+                            0 => return Ok(false),
+                            1 => return Ok(true),
+                            _ => return Err(crate::PyErr::fetch(obj.py())),
+                        }
                     }
                 }
+
+                return Err(missing_conversion(obj));
             }
 
-            Err(missing_conversion(obj))
+            #[cfg(any(Py_LIMITED_API, PyPy))]
+            {
+                let meth = obj
+                    .lookup_special(crate::intern!(obj.py(), "__bool__"))?
+                    .ok_or_else(|| missing_conversion(obj))?;
+
+                let obj = meth.call0()?.downcast::<PyBool>()?;
+                return Ok(obj.is_true());
+            }
         }
 
-        #[cfg(any(Py_LIMITED_API, PyPy))]
-        {
-            let meth = obj
-                .lookup_special(crate::intern!(obj.py(), "__bool__"))?
-                .ok_or_else(|| missing_conversion(obj))?;
-
-            let obj = meth.call0()?.downcast::<PyBool>()?;
-            Ok(obj.is_true())
-        }
+        Err(err.into())
     }
 
     #[cfg(feature = "experimental-inspect")]
@@ -124,7 +133,7 @@ impl<'source> FromPyObject<'source> for bool {
 
 #[cfg(test)]
 mod tests {
-    use crate::types::{PyAny, PyBool, PyModule};
+    use crate::types::{PyAny, PyBool};
     use crate::Python;
     use crate::ToPyObject;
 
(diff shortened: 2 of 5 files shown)


birkenfeld:
I don't think it is good to single out objects that define `__bool__`. It's one of several equally valid ways to define truth conversion, on the Python side it absolutely doesn't matter which one you use. For example, when implementing a collection class I might give it an explicit `__bool__` or not, depending on if I can be bothered to look up how the behavior is otherwise.

Contrary to something like `__int__`, `__bool__` isn't reserved for object that "feel like a bool" (e.g. numpy bools).

I.e. my recommendation is, either keep being strict here, or just use `PyAny::is_true()`.

adamreichold:
While I can see the argument for being consistent in one way or the other, i.e. either strict insofar a function argument is sort of type specification for the Python API or loose where we try as hard as possible to convert any incoming value into a the target type, I think that we cannot solve this issue as long as the single `FromPyObject` trait is doing double duty for both approaches.

There is also precedent for this middle ground approach in other implementations, e.g. `Vec` accepts only implementors of the `Sequence` ABC even the implementation to convert the Python iterable into a `Vec` does not require it at all.

So while I would prefer a principled approach, the practicalities of ergonomic usage of PyO3 based on a single trait prevent this for now as far I can see.

birkenfeld:
I can see that, but my argument is that the approach in this PR is worse than either. It happens to work for numpy bools, fixing the original issue, but also allows conversion for a whole host of other types, none of which have to be "bool-like", and arbitrarily excludes others which are no less "bool-like".

adamreichold:
> but also allows conversion for a whole host of other types, none of which have to be "bool-like", and arbitrarily excludes others which are no less "bool-like".

But does this have any concrete usability consequences or is it merely an uncomfortably inconsistent design?

adamreichold:
Maybe as a data point that this can work, pybind11 seems have run into the same problem, i.e. https://github.com/pybind/pybind11/pull/925 [^1] and solved it in similar manner albeit optimized to avoid the cost of `lookup_special` (which could do here as well).

Admittedly, they are in the advantageous position of having the `pyarg().noconvert()` modifier so that the author can explicit request stricter or looser conversions. But then again, they even special-cased the loose behaviour to always apply to `numpy.bool_` (which is also something we could do here).

[^1]: I was not aware of that PR went starting to implement this one.

adamreichold:
But they certainly also had a long discussion before settling on the final approach. 😅

davidhewitt:
This reminds me a lot of https://github.com/python/cpython/pull/15609 - in Python 3.12 they changed all `C` functions to accept any object and check using equivalent of `PyAny::is_true`.

Slight aside, what do you think of renaming `PyAny::is_true` to `PyAny::is_truthy`? I keep intuitively thinking `x.is_true()` would map to the python `x is True` :)

adamreichold:
> Slight aside, what do you think of renaming `PyAny::is_true` to `PyAny::is_truthy`? I keep intuitively thinking `x.is_true()` would map to the python `x is True` :)

This I like.

> > This reminds me a lot of [python/cpython#15609](https://github.com/python/cpython/pull/15609) - in Python 3.12 they changed all `C` functions to accept any object and check using equivalent of `PyAny::is_true`.

This, not so much.

davidhewitt:
👍 I will open an `is_truthy` PR hopefully later in the week.

Regarding the Python 3.12 change, I agree it's not my favourite behaviour either (and PyO3 being principled about the difference between `1` and `True` helped me isolate a bug in the past). Despite that I sort of think if we're making this conversion more lax it might be best if we match CPython behaviour to minimise surprises to users. But I'll try to review this PR in detail as soon as I get a moment.

davidhewitt:
(Sorry I was a little slow on review here, ended up getting excited by the Py2 work...)

Hmmm, it's hard to make a judgement on the right solution for this. I am definitely in favour of us trying to fix the numpy case.

Overall, we've definitely been trending slowly towards being more lax on these Python -> Rust conversions, e.g. #3374, #3197, as this seems to be what users want.

>  they are in the advantageous position of having the `pyarg().noconvert()` modifier

Yes, I definitely think there is room for something like this; I called it "strict" mode in https://github.com/PyO3/pyo3/issues/3226#issue-1751584990, we could have `#[pyo3(noconvert)]` as an annotation on function arguments, for example.

----

Overall, I think that pybind11 prior art is a reasonable one to follow here, and I like consistency in the ecosystem (if we think the CPython choice is too lax). I also think it's less breaking for future users if we chose to go more lax again to match CPython than it would be to get stricter.

I definitely think that where we're at is that `FromPyObject` is at the moment a lax conversion. We allowed `str` to ip address in #3197, after all.

So perhaps the road ahead is to proceed with the implementation of this as is currently written in this PR, and then if users want to opt-out, we should add a `#[pyo3(noconvert)]` attribute? I think that could be supported by a separate trait, so we can implement it on-demand starting with bool. The hardest thing might be naming that trait :)

All this said, would `let x: bool = any.extract()?;` being roughly equivalent to Python `bool(any)` be a good thing? My Rust brain says no but my Python brain says yes 😖 

I think I need to sleep on this and see how I feel in the morning...

davidhewitt (approved):

davidhewitt:
With fresh eyes I've come around to this solution, I think the shaker behaviour change is preferable.

If users want to have the stricter method again I suggest we explore the `#[pyo3(noconvert)]` option.

src/types/boolobject.rs
```diff
@@ -56,7 +59,36 @@ impl IntoPy<PyObject> for bool {
...
+                if let Some(nb_bool) = tp_as_number.nb_bool {
+                    match (nb_bool)(ptr) {
+                        0 => return Ok(false),
+                        1 => return Ok(true),
+                        _ => return Err(crate::PyErr::fetch(obj.py())),
+                    }
+                }
+            }
+
+            Err(PyTypeError::new_err("object has no __bool__ magic method"))
```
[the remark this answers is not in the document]
adamreichold: I changed the error message into the suggested form.

adamreichold:
@birkenfeld I would be quite unhappy to merge this over you direct objection. I wonder if you can be placated by the suggestion of adding a `#[pyo3(noconvert)]` attribute to put these decisions into our user's hands eventually? Also maybe the pybind11 precedent suggests that this is tenable? I think we all agree that this is messy and inconsistent and from that perspective it can be said that "the approach in this PR is worse than either", but I also think that this is an remedy for an actual pain point felt by our users working in a messy and inconsistent ecosystem.

birkenfeld:
I'm all in favor of the `noconvert` annotation. However, it doesn't touch my core objection: that considering `__bool__` but not `__len__` is simply inconsistent. The availability of a way to opt out of broad boolean conversion makes this inconsistency even more glaring since there is no justification for limiting the scope in search of a compromise.

Just think about the user's perspective. If I can pass in `None` and objects like `winreg.HKEY`s or `gdbm.gdbm`s, but not others that Python considers bool-convertible just fine in other circumstances, does that really make sense?

adamreichold:
> The availability of a way to opt out of broad boolean conversion makes this inconsistency even more glaring since there is no justification for limiting the scope in search of a compromise.

Yes, if we had `noconvert` today, using the full truth value testing procedure might be a better choice for the "broad/loose" setting. But we do not have that and hence the compromise.

> Just think about the user's perspective. If I can pass in None and objects like winreg.HKEYs or gdbm.gdbms, but not others that Python considers bool-convertible just fine in other circumstances, does that really make sense?

I think there is a difference "bool-convertible" and "truthy". The truth value testing which includes `__len__` is about what is "truthy" and does not say anything about being "bool-convertible" which I would argue is something that duck-typed Python does not care about at all, but statically notionally typed Rust and C++ do. And while I think considering a non-empty collection or a not-`None` class instance true-like for an `if` statement works, I don't think it works for a boolean function argument because it just looses too much information.

davidhewitt:
> I think there is a difference between "bool-convertible" and "truthy". The truth value testing which includes __len__ is about what is "truthy" and does not say anything about being "bool-convertible" which I would argue is something that duck-typed Python does not care about at all, but statically notionally typed Rust and C++ do. 

This is my feeling too when I was reflecting on `let x: bool = any.extract()?;` above, I think our intention there has always been for that to be book conversion and not truth testing.

birkenfeld:
My point is that `__bool__` is not a method for "bool conversion". It is specifically for defining truthiness, as done by `bool(x)` or `if x`.

Just check out https://github.com/search?type=code&auto_enroll=true&q=%22def+__bool__%22 - many of the objects defining `__bool__` are far from "objects similar to a bool that should be convertible to it". (Just like my example of HKEY and gdbm, which are from the stdlib.)

davidhewitt:
I see what you mean, that "bool conversion" isn't separable from truth testing on the Python side.

Given the strength of @birkenfeld's conviction, maybe we should slow down for now and just take the route `pybind11` did for their strict pathway, i.e. special case for numpy bools?

adamreichold:
> Given the strength of @birkenfeld's conviction, maybe we should slow down for now and just take the route pybind11 did for their strict pathway, i.e. special case for numpy bools?

Yes, if only to resolve the issue at hand. I do see some types in the above list would be good candidates as well, like pytorch's `BoxedBool`, but I guess it is better to solve some issues than none at all. Will push the change as a separate commit...

newsfragments/3638.changed.md
```diff
@@ -0,0 +1 @@
+Values of type `bool` can now be extracted from all Python values defining a `__bool__` magic method.
```
davidhewitt, line 1: CHANGELOG entry is now outdated

davidhewitt (approved):
Looks good to me, save for the CHANGELOG entry. I think if we want to go further here the next steps will be to figure out how the `#[pyo3(noconvert)]` annotation might work.

newsfragments/3638.changed.md
adamreichold: Thanks for spotting this!

adamreichold:
Uff, it seems `__qualname__` on PyPy is `bool_` instead of `numpy.bool_`? Do would just skip the test on PyPy or loosen the name check to `numpy.bool_` or `bool_`? Should I use something other than `get_type().name()`?

davidhewitt:
I wonder if this changes based on version, e.g. maybe 3.9 or 3.10 change?

It could be worth adding PyPy 3.10 on Ubuntu to the pr ci to get a hint what might work before the full merge queue runs?

adamreichold:
I tested this locally using `pyenv` and it does seem independent of the PyPy version, i.e. the behaviour is the same for 3.8, 3.9 and 3.10 and `name()` always yields "bool_" without the module.

adamreichold:
But reading https://peps.python.org/pep-3155/ and testing this in the CPython REPL, `__qualname__` is not intended to include the module name. I wonder if it is really a full replacement for `tp_name` as used by us?

adamreichold:
From reading https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_name, I think we should use `__module__.__name__` instead of `__qualname__` as the abi3 fallback?

adamreichold:
So, I think ideally, this would be rebased on #3660 and made use of `PyType::full_name` if that one is accepted

adamreichold:
Rebased this onto main now that the full name is available on `PyType` and used this to limit the scope of the looser conversion.

PR #3646, opened 2023-12-12
Close TLS-based soundness holes by running closure on a newly created thread.

### LINKED ISSUE #2141: `Python::allow_threads` is unsound in the presence of `send_wrapper`.

### LINKED ISSUE #3640: `Python::allow_threads` is unsound in the presence of `scoped-tls`.
In analogy to

* #2141

we can smuggle `Ungil` data across `Python::allow_threads` using `scoped-tls`, too.

```rs
use pyo3::prelude::*;
use pyo3::types::PyString;
use scoped_tls::scoped_thread_local;

fn main() {
    Python::with_gil(|py| {
        let string = PyString::new(py, "foo");

        scoped_thread_local!(static WRAPPED: PyString);

        WRAPPED.set(string, || {
            py.allow_threads(|| {
                WRAPPED.with(|smuggled: &PyString| {
                    println!("{:?}", smuggled);
                });
            });
        });
    });
}
```
(results in segfault in my test)

***Unlike #2141, this issue is virtually unsolvable***, i.e. even the auto trait approach with the `feature="nightly"` enabled cannot catch this at all. There’s no property in the callback to `allow_threads` that can catch this. Really, the callback doesn’t even capture _anything at all_:

```rs
use pyo3::prelude::*;
use pyo3::types::PyString;
use scoped_tls::scoped_thread_local;

scoped_thread_local!(static WRAPPED: PyString);

fn callback() {
    WRAPPED.with(|smuggled: &PyString| {
        println!("{:?}", smuggled);
    });
}

fn main() {
    Python::with_gil(|py| {
        let string = PyString::new(py, "foo");

        WRAPPED.set(string, || {
            py.allow_threads(callback); // callback is an ordinary `fn() -> ()` item.
        });
    });
}
```

Again, there’s a “whose fault” question to be asked, whether the fact that *the standard library*’s thread-locals require `T: 'static` means there’s any guarantees that non-`'static` data isn’t allowed to be “smuggled” through thread-local storage. In my view, if you look at what kind of things it offers, `scoped-tls` seems _even less_ unreasonable to be called “sound”, compared to `sync_wrapper`. Yet the consequences for `Ungil` are more detrimental.

### LINKED ISSUE #3658: `F: Ungil` on nightly insufficient for soundness of `allow_threads`, even without `scoped-tls` or `send_wrapper`
```rs
use std::cell::Cell;
use std::thread;

use pyo3::prelude::*;

#[pyclass]
struct Class(Cell<u8>);

fn main() -> PyResult<()> {
    Python::with_gil(|py| {
        let class_1 = Py::new(py, Class(Cell::new(0)))?;
        let class_2 = class_1.clone_ref(py);
        let cell: &Cell<u8> = &class_1.borrow(py).0;
        py.allow_threads(|| {
            let t = thread::spawn(move || {
                Python::with_gil(|py| {
                    let cell: &Cell<u8> = &class_2.borrow(py).0;
                    let init = cell.get();
                    loop {
                        let gotten = cell.get();
                        if init != gotten {
                            println!("concurrently modified! - {init} vs {gotten}");
                            break;
                        }
                    }
                });
            });
            loop {
                cell.set(cell.get().wrapping_add(1));
                if t.is_finished() {
                    break;
                }
            }
            t.join().unwrap();
        });
        PyResult::Ok(())
    })?;
    PyResult::Ok(())
}
```
```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/…`
concurrently modified! - 63 vs 66
```
(concurrent modification of a `Cell` is UB; I skipped the additional step where you could abuse this data race to cause more obvious UB such as reading memory out of bounds)

This is issue is possibly irrelevant in light of existing issues like #3640 (and fixed by #3646 which abandons `Ungil`). Also, it might in principle also be fixable by making `.borrow` do a dynamic check that we’re in the same thread as previous `.borrow`; or by demanding `Sync` for `pyclass`; or letting the user choose between either of the two options (dynamic check or `Sync`).

I’m raising this issue as another point of argument that nightly `Ungil` isn’t as good as it claims to be, and this would need to be addressed in case the argument would win that we should keep `Ungil` (with documented unsoundness warnings) for its benefits of not conservatively requiring `Send`.

adamreichold:
To be done before this is a candidate for merging:
- [x] add unsafe escape hatch to avoid threading overhead
- [x] update inline documentation
- [x] determine if Python thread state should be moved to runtime thread
- [x] switch `unsafe_allow_threads` to unsafe token approach
- [x] decide which method takes the existing `allow_threads` name
- [x] switch to builder-like API for choosing strategy
- [ ] add changelog and migration guide entries
- [x] actually benchmark the performance impact?
- [ ] enable allowing threads even if the GIL is not actually held as discussed [here](https://github.com/PyO3/pyo3/discussions/4097)

Closes #2141
Closes #3640
Closes #3658


src/marker.rs
```diff
@@ -528,21 +286,123 @@ impl<'py> Python<'py> {
...
+
+            result.write(catch_unwind(AssertUnwindSafe(f)));
+
+            result_sender.send(()).unwrap();
+        };
+        // SAFETY: the current thread will block until the closure has returned
+        let task = Task(unsafe { transmute(&mut task as &mut dyn FnMut()) });
+
+        // Enqueue task and spawn thread if necessary
+        static THREADS: Mutex<Vec<SyncSender<Task>>> = const_mutex(Vec::new());
```
adamreichold, line 356: @davidhewitt @steffahn I added a second commit to re-use threads and as is to be expected by manual implementation of a scoped thread pool, there is quite a lot of unsafe involved about which I am somewhat unhappy. (I am especially nervous about unwinding as we need to keep the stack-bound data alive until the threads are done with it.)

Generally speaking, it does not seem nice to manually re-implement this but our requirements are somewhat special if I understand things correctly. For example, we never want to block waiting to re-use an existing thread, but rather spawn a new one if none are currently ready.

Finally, adding a tunable for the duration for which the threads wait for new tasks while idling is also not a welcome addition.

steffahn: I haven’t looked at the code yet, but one thing that came to mind is that the moment the helper thread is started, and the moment it ends, we are holding the GIL (which is a kind of mutex-like thing itself, if I understand correctly), so maybe that can serve to simplify synchronization concerns? Not that that would cut down on the amount of unsafe code…

For panics, one should probably best run `f()` with `catch_unwind`… then send the `Result` back to the thread where `allow_threads` was called, and [re-throw](https://doc.rust-lang.org/std/panic/fn.resume_unwind.html) the panic at the end, after the worker thread was put back into the queue.

steffahn: To fully and conservatively address concerns of accidentally leaving in place something in the main thread logic that can panic, during the section where the task was (or is about to be) sent, and the working thread not joined yet, it’s possible to insert a defensive abort bomb, i.e. a type that aborts the program (and informs the user that there’s a bug in `pyo3`) when dropped, and that gets defused after the worker thread’s response was received.

adamreichold: > I hadn’t looked at the code yet before writing this first paragraph, but one thing that came to mind is that the moment the helper thread is started, and the moment it ends, we are holding the GIL (which is a kind of mutex-like thing itself, if I understand correctly), so maybe that can serve to simplify synchronization concerns? Not that that would cut down on the amount of unsafe code… so feel free to ignore this thought.

I considered this (i.e. `GILProtected<RefCell<..>>` instead of `Mutex<..>`) but decided against it as the was the possibility that the GIL would not yet be unlocked when the closure starts (on another thread) because it would be quite easy to accidentally still have it acquired when I send the task to a waiting thread. Hence, to keep things simple I chose to first release it and then do my own looking using a standard mutex.

> For panics, one should probably best run f() with catch_unwind… then send the Result back to the thread where allow_threads was called, and [re-throw](https://doc.rust-lang.org/std/panic/fn.resume_unwind.html) the panic at the end, after the worker thread was put back into the queue.

This is basically what I did, even though I did not use `resume_unwind` but rather just `unwrap`ed on a `std::thread::Result`. Will change that...

> To fully and conservatively address concerns of accidentally leaving in place something in the main thread logic that can panic, during the section where the task was (or is about to be) sent, and the working thread not joined yet, it’s possible to insert a defensive abort bomb, i.e. a type that aborts the program (and informs the user that there’s a bug in pyo3) when dropped, and that gets defused after the worker thread’s response was received.

Yes, I though about this and we can probably reuse our existing `PanicTrap` for that (which we currently use to avoid unwinding across the FFI boundary). Will add this...

steffahn: I don’t know how I overlooked the `catch_unwind`; probably because it’s so far away (in code) from the `spawn`.

adamreichold: Seems like `ManuallyDrop` and `MaybeUninit` and the accompanying `unsafe` blocks are not actually necessary.

src/marker.rs
```diff
@@ -528,21 +286,133 @@ impl<'py> Python<'py> {
...
+        let (result_sender, result_receiver) = sync_channel::<Result<T>>(0);
+
+        let mut f = Some(f);
+
+        let mut task = || {
+            let f = f.take().unwrap();
+
+            let result = catch_unwind(AssertUnwindSafe(f));
+
+            result_sender.send(result).unwrap();
```
davidhewitt, line 351: Does the `.unwrap()` here give us any benefit? The thread dies either way, and presumably the unwind here isn't interesting.

```suggestion
            let _ = result_sender.send(result);
```

src/marker.rs
```diff
@@ -528,21 +286,133 @@ impl<'py> Python<'py> {
...
+        };
+
+        // SAFETY: the current thread will block until the closure has returned
+        let task = Task(unsafe { transmute(&mut task as &mut dyn FnMut()) });
+
+        // 2. Dispatch task to waiting thread, spawn new thread if necessary
+        let trap = PanicTrap::new(
+            "allow_threads panicked while stack data was accessed by another thread which is a bug",
+        );
```
davidhewitt, line 360: Feels good when our existing pattern works for a new use case.

src/marker.rs
davidhewitt: It looks like we got to the point where the only unsafe code is the sharing of the function pointer to the stack data between the threads, nice! I suppose if we wanted to eliminate that we could use an allocation, but I personally like this way.

adamreichold: > I suppose if we wanted to eliminate that we could use an allocation, but I personally like this way.

An extra allocation would remove the unsafe block when calling the tasks, but there would still be the unsafe block when transmuting the closure to lie to the compiler that it has a `+ 'static` bound (the trait object lifetimes are defaulted here and not immediately visible in the types) which cannot have because neither `F` nor `T` have a `+ 'static` bound.

src/marker.rs
```diff
@@ -528,21 +286,133 @@ impl<'py> Python<'py> {
...
+
+                task_sender.send(task).unwrap();
+
+                task_sender
+            }
+            State::Dispatched(task_sender) => task_sender,
+        };
+
+        // 3. Wait for completion and check result
+        let result = result_receiver.recv().unwrap();
```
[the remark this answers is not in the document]
adamreichold: Yes, if the worker thread died and disconnected the channel without a value which should not happen because we catch panics from the closure.

And indeed, this will trigger the panic trap and abort with the corresponding error message.

Generally, I went for more rather than fewer `unwrap`s to increase the chances that we actually hit our panic trap if anything unexpected happens rather than silently take the error and hit some unexpected implied condition. Mainly the because some downstream inconsistencies often leads to debugging sessions beginning with a question "But this state should not be possible?!".

src/marker.rs
adamreichold: As written above, I wanted the make the possible failure points more explicit and a panicking thread with a detached join handle does at least trigger a log message, c.f. <https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=de70d0560a08094a5527502334e014ac>

(I did consider naming the threads to generally improve usability here by indicating that these are runtime threads used by PyO3.)

src/marker.rs
```diff
@@ -528,21 +286,133 @@ impl<'py> Python<'py> {
...
+                        unsafe { (*task.0)() };
+                    }
+                });
+
+                task_sender.send(task).unwrap();
+
+                task_sender
+            }
+            State::Dispatched(task_sender) => task_sender,
+        };
```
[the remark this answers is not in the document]
adamreichold: Will drop `enum State` even though I have to say that I did the like explicitness of turning that state into types instead of handling it via control flow, but the named block is indeed more succinct.

I would however prefer to keep moving `task` instead of deriving `Copy` or doing the basically equivalent `Task(task.0)` dance because having only one handle by which the task is called is an important safety invariant here.

src/marker.rs
davidhewitt: 😅 great reference in the playground!

Sure thing, makes sense to keep this then 👍

src/marker.rs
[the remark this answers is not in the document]
davidhewitt: Agreed that this should panic, the only question here imo is whether this should be `.panic()` or `.expect("worker thread died unexpectedly")`

src/marker.rs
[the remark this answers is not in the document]
davidhewitt: Yes I didn't love the `Task(task.0)` dance but missed earlier a way to avoid that, looks like this can work:

```rust
        let mut task = task;
        // Try to get an idle worker thread
        while let Some(task_sender) = THREADS.pop() {
            task = match task_sender.send(task) {
                Ok(()) => break 'sender task_sender,
                // the worker thread timed out and shutdown
                Err(SendError(task)) => task,
            }
        };
```

src/marker.rs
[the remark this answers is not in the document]
adamreichold: I think `.expect` is basically always better than `.unwrap` and it is just laziness to not do that. The only downside I see is that the code could become cluttered eventually, but since this part is so delicate, I think the additional documentation and debuggability are definitely worth at little clutter. Pushed a version which documents all `unwrap`s.

src/marker.rs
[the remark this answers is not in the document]
adamreichold: Hhhmm, I just bit the bullet and used a new name for `SendError(same_task) => task = same_task,` since that seem more readable to me even if I hate having to name things just for the syntactical sake of it.

adamreichold:
> The only downside I can see about this is that it brings a whole new set of thread locals into play, for the worker thread. I think there are interactions here which we'd need to at least document:
> 
>     * Callers of `allow_threads` will have to treat all `thread_local` values as in an undefined state at the beginning of the execution of the closure. I say "undefined" and not "uninitialized" because with the thread pool, reuse of the threads means that it's possible to observe effects from one closure in another. I wonder if `tokio` has had prior discussion of this with their `spawn_blocking` API, which I think works similarly.

Indeed. I think we should certainly document this and of course, what I would really want to have from the Rust standard library here would be a function like

```rust
with_isolated_tls<F,T>(f: F) where F: Send + FnOnce() -> T, T: Send;
```

which would just wipe the slate clean w.r.t. the system-level TLS implementation if that is actually possible? Do you have an educated guess on that @steffahn?

It did not find any directly addressing this in the `spawn_blocking` documentation. I guess this is because it is more obvious that this is using a thread pool and leaves the reader to figure out the implications. I will try to come up with some useful wording myself and then we can knead into shape together.

>     * Inside the closure, if a user then called `Python::with_gil`, this will create a new Python thread state for the worker thread. Is this expected?
>       
>       * Do we make `with_gil` run back on the original thread which is blocked waiting for the result? (e.g. by using a similar mechanism to `Task` in reverse?).
>       * Or do we just document that this will happen, i.e. put it in the same bucket as all the rest of the "thread locals are undefined" statement of above?

I would prefer to not trying to call back and forth, mainly to keep `with_gil` simple, performant and more explainable. Hence, I would indeed work this into the general caveat on thread identity as discussed above. 

> I understand the desire to keep the safe `allow_threads` name, but I still have a feeling that the semantic change here might mean that it's better to rename and break existing code.

Still torn because I think we should keep the safe way the obvious and effortless way. Otherwise, people like me will start looking into the alternatives and see that the unsafe one is more efficient and wonder "Maybe that little bit of unsafe is worth avoid the HUGE COST of spawning a thread?!" and be tempted by the dark side of Rust.

If the more efficient but unsafe way must be looked for and opted in explicit, this will more likely happen only if there is an actually diagnosed/measured performance issue. At least for scientific code for example, I find that unlikely because the closure in allow threads will usually start/call upon a whole bunch of threads doing something using e.g. Rayon.

It is not something I would like to fight about, but I would be glad if we could get some more input from other maintainers and/or users. (And this stays a draft until we decided one way or the other. I am not trying to decided this having this PR which hijacks name and is still missing the escape hatch version in any case.)

adamreichold:
(Will take care of the MSRV build separately... Uff.)

davidhewitt:
> (Will take care of the MSRV build separately... Uff.)

Time to bump to 1.62? 😈

davidhewitt:
> Still torn because I think we should keep the safe way the obvious and effortless way. Otherwise, people like me will start looking into the alternatives and see that the unsafe one is more efficient and wonder "Maybe that little bit of unsafe is worth avoid the HUGE COST of spawning a thread?!" and be tempted by the dark side of Rust.

I take your point here, my concern can probably be mitigated by a clear entry in the migration guide.

davidhewitt:
> I would prefer to not trying to call back and forth, mainly to keep with_gil simple, performant and more explainable. Hence, I would indeed work this into the general caveat on thread identity as discussed above.

One option which _might_ work is to move the Python thread state into the worker thread using some kind of `PyEval_RestoreThread` technique. Actually I'm not sure if `with_gil` is called inside `allow_threads` already whether a new thread state is created, maybe my concern here is moot. We should check.

adamreichold:
> Indeed. I think we should certainly document this and of course, what I would really want to have from the Rust standard library here would be a function like
> 
> ```rust
> fn with_isolated_tls<F,T>(f: F) where F: Send + FnOnce() -> T, T: Send;
> ```
> 
> which would just wipe the slate clean w.r.t. the system-level TLS implementation if that is actually possible? Do you have an educated guess on that @steffahn?

Just noticed that this would only close the `scope-tls` loophole, not the `send-wrapper` one, so never mind...

src/marker.rs
```diff
@@ -528,20 +338,174 @@ impl<'py> Python<'py> {
...
+                })
+                .expect("failed to create allow_threads runtime thread");
+
+            task_sender
+        };
+
+        // 3. Wait for completion and check result
+        let result = result_receiver
+            .recv()
+            .expect("allow_threads runtime thread died unexpectedly");
```
bjorn3, line 450: This returns before the TLS destructors get a chance to run, right? Is that sound?

davidhewitt: Can you elaborate on a particular example of what you're concerned might run in a TLS destructor to create a potential soundness issue?

davidhewitt: Ah I see your point. The soundness concern with TLS has so far relied on `scoped-tls`, so I think that any TLS destructor there is not a concern because the scope of any TLS must by construction be smaller than the task which we execute on the spawned thread.

If we can find a way to demonstrate a concern without using `scoped-tls`, then I think this is a problem

steffahn: The python objects also carry a lifetime, so they cannot be put into `'static` thread-local storage. TLS destructors should only be relevant for sort-of “ordinary” `thread_local!`, which can never contain or reference the python objects. The `scoped-tls` API only ever leaves values in TLS for the duration of a callback, which could only be contained (and thus also ending) _within_ the callback to `f`, never keep state across multiple `while let Ok(task) = next_task` iterations.

bjorn3: Right, because of the non-`'static` lifetime of python objects, it wouldn't be possible to put them in a tls variable without `scoped-tls` and with `scoped-tls` it would be remove again when exiting the scoped-tls scope before the tls destructor gets a chance to run and before we receive the result here. As such I don't think it is exploitable.

adamreichold:
This is not really a surprise, but the overhead of `allow_threads` itself does increase by two orders of magnitude due to this change:

```console
allow_threads           time:   [5.3486 µs 5.3774 µs 5.4055 µs]
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe

unsafe_allow_threads    time:   [28.566 ns 28.577 ns 28.591 ns]
Found 16 outliers among 100 measurements (16.00%)
  7 (7.00%) high mild
  9 (9.00%) high severe
```

I'd say it is still small for the expected use cases of performing I/O or dispatching work into thread pools, but the cost of this mitigation is certainly not negligible.

adamreichold:
> > I would prefer to not trying to call back and forth, mainly to keep with_gil simple, performant and more explainable. Hence, I would indeed work this into the general caveat on thread identity as discussed above.
> 
> One option which _might_ work is to move the Python thread state into the worker thread using some kind of `PyEval_RestoreThread` technique. Actually I'm not sure if `with_gil` is called inside `allow_threads` already whether a new thread state is created, maybe my concern here is moot. We should check.

I checked there is indeed a difference: Without moving to another thread, re-acquiring the GIL will yield the same thread state which was original present. After this change, a different thread state is found however.

So if I understand your idea correctly: I move the thread state into the task, put it into a thread-local variable there which is picked up by our `GILGuard` and installed using `PyEval_RestoreThread` instead of just calling `PyGILState_Ensure`?

This would add another TLS look-up to every call to `Python::with_gil` though. Could you re-iterate what observable state is bound to this interpreter thread state, i.e. what exactly do we gain by potentially installing it in the worker thread?

adamreichold:
@cjrh @GodTamIt @wallies `tantivy-py` has a call to `allow_threads` in its [`search` method](https://github.com/quickwit-oss/tantivy-py/blob/539129154160e850d3e7d9a8ec5bc358b9b121d7/src/searcher.rs#L166). Does one of you have a use case and the spare time to test `tantivy-py`'s main branch against this branch here to determine the performance impact of the mitigation for a realistic scenario? I think that would be rather helpful to inform the discussion and I think search using Tantivy is a bit of a worst case as its usually a purely computational task (assuming everything is cached). Thank you!

davidhewitt:
> > > I would prefer to not trying to call back and forth, mainly to keep with_gil simple, performant and more explainable. Hence, I would indeed work this into the general caveat on thread identity as discussed above.
> > 
> > 
> > One option which _might_ work is to move the Python thread state into the worker thread using some kind of `PyEval_RestoreThread` technique. Actually I'm not sure if `with_gil` is called inside `allow_threads` already whether a new thread state is created, maybe my concern here is moot. We should check.
> 
> I checked there is indeed a difference: Without moving to another thread, re-acquiring the GIL will yield the same thread state which was original present. After this change, a different thread state is found however.
> 
> So if I understand your idea correctly: I move the thread state into the task, put it into a thread-local variable there which is picked up by our `GILGuard` and installed using `PyEval_RestoreThread` instead of just calling `PyGILState_Ensure`?
> 
> This would add another TLS look-up to every call to `Python::with_gil` though. Could you re-iterate what observable state is bound to this interpreter thread state, i.e. what exactly do we gain by potentially installing it in the worker thread?

I think the main reason would be python thread locals and contextvars, which I think are tied to a python thread state. I think the Python stack frames might be also. If I am right about any of these, then calling a method on a Python object with that new thread state might have quite a few observable differences compared to calling it back on the original thread.

I haven't actually checked the CPython source to confirm the details here, nor am I sure which of those semantics might be considered implementation detail. Nor am I sure that moving the python thread state to a different OS thread would actually be legal.

One option to play it safe could be to ban using `with_gil` inside `allow_threads` by using the same solution we used for preventing `with_gil` during GC traversal? This is at least something we could relax in the future following more research.

adamreichold:
> I think the main reason would be python thread locals and contextvars, which I think are tied to a python thread state.

If that is the case, then I think we should _not_ move the thread state because this would bring native TLS and Python's version of it out of sync. I think this would basically also make it illegal to move the thread state, e.g. if the Python TLS contains reference to native TLS which would be invalidated this way.

> One option to play it safe could be to ban using with_gil inside allow_threads by using the same solution we used for preventing with_gil during GC traversal? This is at least something we could relax in the future following more research.

Personally, I don't think we need to do anything and would prefer to leave `with_gil` working as it is because at least my tests did not show up anything untoward happening after re-acquiring the GIL on the worker thread. I also personally use patterns where the GIL is released but sometimes taken again inside to e.g. report results via Python callbacks which would break due to this but would not care at all about running on another thread (they already do in any case).

davidhewitt:
> Personally, I don't think we need to do anything and would prefer to leave `with_gil` working as it is because at least my tests did not show up anything untoward happening after re-acquiring the GIL on the worker thread. I also personally use patterns where the GIL is released but sometimes taken again inside to e.g. report results via Python callbacks which would break due to this but would not care at all about running on another thread (they already do in any case).

I think I agree with you on this, I think in practice being able to nest is very likely and also hard to prevent.

Between the performance and subtle behavioural changes, I'm getting more sure that we need to introduce this under a new name to at least force users to think when they upgrade. Whether we also introduce an `unsafe` variant under a new name, I think the`allow_threads` symbol needs to go.

adamreichold:
> I think I agree with you on this, I think in practice being able to nest is very likely and also hard to prevent.

Ok, let's leave `with_gil` alone for now then.
 
> Between the performance and subtle behavioural changes, I'm getting more sure that we need to introduce this under a new name to at least force users to think when they upgrade. Whether we also introduce an `unsafe` variant under a new name, I think the`allow_threads` symbol needs to go.

I added commit where I renamed `allow_threads` to `safe_allow_threads` which would force everyone to make a choice upon upgrading, i.e. either switch to the safe but slow `safe_allow_threads` or switch to the unsafe but fast `unsafe_allow_threads`. What do you think?

I am not completely confident I found all occurrences but I guess we will get there. `unsafe_allow_threads` sill has the API problem that it infects the closure. Should we keep this for now as a detriment to using it or should I immediately switch to the "unsafe token" variant?

I also added a separate commit which replaces the two rendezvous channels which have an unintended serializing effect by single hand-written mailbox. This does yield a measurable if small improvement to the benchmark at the cost of additional code complexity. Not sure how we want to play this? (The new code does make the shared state machine explicit though which actually could be considered an improvement. Then again, these could also all be follow-up MR which can be reviewed separately.)

Also noticed that we already ship an `allow_threads`-sensitive benchmark in the word count example and interestingly enough, its performance is completely unaffected (all changes below the noise threshold going in both directions) by the change. This probably means that using `ThreadPoolExecutor` dwarfs the overhead of `safe_allow_threads`.

adamreichold:
(Alternatively, I could squash commits 2 to 6 to simplify the history a bit for easier review.)

adamreichold:
@davidhewitt Does splitting into `safe_allow_threads` and `unsafe_allow_threads` alleviate your naming concern? If so, then there is only documentation updates left to be done...

davidhewitt:
Yes, with two thoughts:
- I'm trying to think of alternative names to `safe_allow_threads` and `unsafe_allow_threads`, but there's nothing obviously better.
- I think it's worth switching `unsafe_allow_threads` to use the unsafe-token API now, as I think we agree that's got a smaller blast radius for the `unsafe`.

adamreichold:
So I think from implementation point of view this is ready for review. It is just missing the migration guide entry but I would like to have a project consensus before the gruesome task of actually writing honest to goodness documentation.

Which is of course the main thing still missing: So do we actually want to do this or not? It does close the known soundness loopholes w.r.t. `allow_threads` which I think is a nice place to be and I also what I would want personally. But the cost is high, either the problematic performance profile of `safe_allow_threads` (the micro benchmark is bad, but the example usage is completely unaffected, we are still missing measurements from more realistic scenarios) or the unsafe and weird API of `unsafe_allow_threads` with safety invariants which are basically uncheckable if unlikely to be broken unintentionally.

Personally, I also have hopes that in the future, there might be a more efficient implementation of `safe_allow_threads` by gaining control over TLS without starting threads but that is admittedly not ensured. So this, the peace of mind provided by a sound API and the existence of the unsafe escape hatch lead me to want to commit to this change.

src/marker.rs
```diff
@@ -522,26 +337,282 @@ impl<'py> Python<'py> {
...
+
+            result = Some(catch_unwind(AssertUnwindSafe(f)));
+        };
+
+        // SAFETY: the current thread will block until the closure has returned
+        let mut task = Task(unsafe { transmute(&mut task as &mut (dyn FnMut() + '_)) });
+
+        // 2. Dispatch task to waiting thread, spawn new thread if necessary
+        let trap = PanicTrap::new(
+            "allow_threads panicked while stack data was accessed by another thread which is a bug",
```
davidhewitt, line 415: Is it worth adding our issue tracker here?

```suggestion
            "`safe_allow_threads` panicked while stack data was accessed by another thread, please report this as a bug at https://github.com/PyO3/pyo3/issues",
```

src/marker.rs
```diff
@@ -522,26 +337,282 @@ impl<'py> Python<'py> {
...
+        F: Send + FnOnce() -> T,
+        T: Send,
+    {
+        use std::mem::{replace, transmute};
+        use std::panic::{catch_unwind, resume_unwind, AssertUnwindSafe};
+        use std::sync::Arc;
+        use std::thread::Builder;
+        use std::time::Duration;
+
+        use parking_lot::{const_mutex, Condvar, Mutex};
```
davidhewitt, line 381: As much as I'd love to use std's primitives for const-construction, I think 0.21 contains enough churn that we don't need to push an MRSV bump on users too. We can tidy up later on 😄

src/marker.rs
```diff
@@ -522,26 +337,282 @@ impl<'py> Python<'py> {
...
+
+        // SAFETY: the current thread will block until the closure has returned
+        let mut task = Task(unsafe { transmute(&mut task as &mut (dyn FnMut() + '_)) });
+
+        // 2. Dispatch task to waiting thread, spawn new thread if necessary
+        let trap = PanicTrap::new(
+            "allow_threads panicked while stack data was accessed by another thread which is a bug",
+        );
+
+        enum MailboxInner {
```
davidhewitt, line 418: This implementation looks really solid. I kept wondering about small micro-optimisations like replacing the `Arc` with a `Box` and a raw pointer, but I felt they were likely all irrelevant compared the the initial cost of creating the thread. The warm path with an idle thread already ready to go looks about as fast as it can possibly be.

src/marker.rs
```diff
@@ -522,26 +337,282 @@ impl<'py> Python<'py> {
...
+                        Task(_) | Working => self.flag.wait(&mut inner),
+                        Empty | Abandoned => {
+                            unreachable!("awaited completion from inactive worker")
+                        }
+                    }
+                }
+            }
+        }
+
+        static THREADS: Mutex<Vec<Arc<Mailbox>>> = const_mutex(Vec::new());
```
davidhewitt, line 513: There might be a possible simplification / optimisation here to use TLS (😂) to have a 1:1 relationship between a regular thread and a worker thread. That would remove the complexity around the mutex and the loop to find a non-dead mailbox.

The downside I guess is that there might potentially be more cold starts and higher peak thread count than if we shared workers between all threads. My gut tells me that if a thread is using `safe_allow_threads` it's likely to do so repeatedly in quick succession, so maybe there _is_ merit to this proposed simplification.

It might also come down to whether switching from a mutex to TLS has any meaningful impact on the microbenchmark?

davidhewitt (commented):
You may not be surprised to hear I am also in favour of moving forward with this. Fully agree with your weighing up of the tradeoffs.

One other thing to note is that by pushing the slight clunkiness of this onto users we might spur one into designing an even better future solution (potentially with new yet unwritten language features). I'll remain hopeful 😂

steffahn:
On naming, I don’t like the `unsafe` or `safe` in the name. All safe API is safe, no need to spell it out, and unsafe API doesn’t need to put `unsafe` into its name. Typically, unsafe API at most has descriptive things like `unchecked` in it when it fits and differentiates from a checked alternative; quite often there’s no indication of the unsafety *at all* in the name, and the name rather describes just _what_ the function does.

For something like `allow_threads`, one could call the unsafe version that executes the closure locally for example `allow_threads_local`. The desire to rename it either way to make users check which they want means we cannot just use `allow_threads` and `allow_threads_local`. I don’t know the best suffix, I’ll re-think to see if I have more ideas.

On that note, I think a suffix is a lot better also because the methods list in documentation is alphabetical, so it’s good to have both versions start the same.

---

Also, I just came up with an alternative contender for to the safety marker argument construction: The function can also be of the kind
```rs
unsafe fn allow_threads_local<F: FnOnce() -> T, T>(&self) -> impl FnOnce(F) -> T
```
so it’s called like
```rs
unsafe { py.allow_threads_local() }(|| {
    println!("effects!");
    "return value"
});
```

[hold on…](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b1871ce6079469da56a7985219dd14b1) that doesn’t quite work, it’d need some extra parentheses
```rs
(unsafe { py.allow_threads_local() })(|| {
    println!("effects!");
    "return value"
});
```

Alternatively… well, I suppose if we want a suffix anyways that might even be quite fitting, we can return some struct with a method instead of `FnOnce`, so that the syntax works out without the extra bracketing, but with additional method-call syntax:

So if safe usage was called, say, `allow_threads_within`
```rs
py.allow_threads_within(|| {
    println!("effects!");
    "return value"
});
```
then the unsafe version could be looking like this:
```rs
unsafe { py.allow_threads_locally() }.within(|| {
    println!("effects!");
    "return value"
});
```
implemented with something like the following mockup / demo
```rs
use core::marker::PhantomData;

#[derive(Copy, Clone)]
pub struct Python<'a>(PhantomData<(&'a (), *const ())>);
impl<'a> Python<'a> {
    pub fn allow_threads_within<F: FnOnce() -> T + Send, T: Send>(self, f: F) -> T {
        // …do the thing…, then
        f()
    }
    // More efficient than normal `allow_threads_within`,
    // calls look like `unsafe { py.allow_threads_locally() }.within(|| { … })`
    // Safety, etc…
    pub unsafe fn allow_threads_locally(self) -> AllowThreadsLocallyWithin<'a> {
        AllowThreadsLocallyWithin(self)
    }
}
/// A marker struct to allow more limited `unsafe {}` block scoping for `Python::allow_threads_locally`
pub struct AllowThreadsLocallyWithin<'a>(Python<'a>);
impl AllowThreadsLocallyWithin<'_> {
    pub fn within<F: FnOnce() -> T + Send, T: Send>(self, f: F) -> T {
        // …do the thing…, then
        f()
    }
}

pub fn usage() {
    let py = Python(PhantomData);

    py.allow_threads_within(|| {
        println!("effects!");
        "return value"
    });

    unsafe { py.allow_threads_locally() }.within(|| {
        println!("effects!");
        "return value"
    });
}
```
([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=60a7d9c04bde94132ff0e57d672583dc))

----

Also, I’ve quickly wondered whether for the unsafe version, the `Send` bounds should be further relaxed (aka removed); or perhaps that could be yet-another, even-more-unsafe variant, to facilitate users that *want* to share stuff like an `Rc` over the closure boundary.

src/marker.rs
davidhewitt: Also I guess an even more radical option would be to not have any global state in PyO3 and just expose this as an `AllowThreadsQueue` channel which users could use to manage their own worker threads. Personally I think I'm ok with this global state being here, but felt it worth tabling the option.

davidhewitt:
@steffahn I do agree with your point on naming. Based on your suggestion of returning the `impl FnOnce`, maybe we can go further and turn it into a builder-like API with `.remote()` and `.local()` methods:

```rust
// safe form
py.allow_threads().remote(|| {
    println!("effects!");
    "return value"
});

// unsafe form
py.allow_threads().local(unsafe { py.allow_threads_locally() }, || {
    println!("effects!");
    "return value"
});
```

src/marker.rs
adamreichold: > My gut tells me that if a thread is using safe_allow_threads it's likely to do so repeatedly in quick succession, so maybe there is merit to this proposed simplification.

I think so too as most Python still have a main thread that only branches out to schedule limited work onto thread pools and such. Will push that as a separate commit...

> Personally I think I'm ok with this global state being here, but felt it worth tabling the option.

I would like to avoid exposing this to user control to keep all our options to open to resolve this using a different implementation in the future.

davidhewitt:
> @steffahn I do agree with your point on naming. Based on your suggestion of returning the `impl FnOnce`, maybe we can go further and turn it into a builder-like API with `.remote()` and `.local()` methods:
> 
> ```rust
> // safe form
> py.allow_threads().remote(|| {
>     println!("effects!");
>     "return value"
> });
> 
> // unsafe form
> py.allow_threads().local(unsafe { py.allow_threads_locally() }, || {
>     println!("effects!");
>     "return value"
> });
> ```

Also, I’ve quickly wondered whether for the unsafe version, the Send bounds should be further relaxed (aka removed); or perhaps that could be yet-another, even-more-unsafe variant, to facilitate users that want to share stuff like an Rc over the closure boundary.

I think that having `Send` is a very useful guard rail against making the basic mistakes, so I think it's worth keeping. With the API I suggest above we could also have a `.unchecked()` form with no guard rails, if we so wanted.

src/marker.rs
```diff
@@ -14,105 +14,81 @@
 //! awaiting a future
 //! - Once that is done, reacquire the GIL
 //!
-//! That API is provided by [`Python::allow_threads`] and enforced via the [`Ungil`] bound on the
-//! closure and the return type. This is done by relying on the [`Send`] auto trait. `Ungil` is
-//! defined as the following:
```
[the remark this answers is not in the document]
adamreichold: > As per discussion on https://github.com/PyO3/pyo3/issues/3658 it seems to me that in the long-term it might still be the better auto trait than Send (once GILProtected is removed and nogil is the only build mode).

I think when that day actually comes, this API itself might be on the chopping block. Until then, I actually feel like `Send` is somewhat appropriate as we do talk about state (holding the GIL) attached to the current thread.

I certainly like the reduction of accidental complexity (auto trait that actually is `Send` on stable, features that only work on nightly, etc.).

adamreichold:
Ok, so instead of the name split, I will go for the builder-like interface and support three variants: "default/remote/safe", "local/unsafe" and "unchecked/unsafe".

adamreichold:
Ok, this now uses a builder-like API with one worker thread per user thread. Please have a look before I tried to wrangle the documentation into shape again.

steffahn:
It’s not really a “check” we’re skipping but a trait bound / constraint. Maybe `unconstrained`? It’s *also* still local, and longer names for more unsafe API should be a good thing, so maybe `local_unconstrained()`? Maybe even `.local().unconstrained()`?

adamreichold:
Agreed, when for the longish "local_unconstrained" name for the really bad one.

src/sync.rs
```diff
@@ -266,6 +281,247 @@ impl Interned {
...
+    pub unsafe fn local(self) -> LocalAllowThreads<'py> {
+        LocalAllowThreads(self.0)
+    }
+
+    /// TODO
+    ///
+    /// # Safety
+    ///
+    /// TODO
+    pub unsafe fn local_unconstrained(self) -> LocalUnconstrainedAllowThreads<'py> {
```
davidhewitt, line 302: I guess that the caller now writes

```rust
unsafe { py.allow_threads().local_unconstrained() }.with(|| ...)
```

Which looks good to me 👍. I guess we would need to document this style to help users avoid accidentally infecting the closure with unsafe.

I like the consistency of calling `.with` in all cases, and the proposed names make sense to me 🎉

davidhewitt (commented):
I'm really happy with this now 🎉

Just curious if switching from the vec to TLS made any observable change in the benchmark?

src/sync.rs
adamreichold: Yes, this is exactly the style I ended up using in the benchmarks. I will document all thread options in the doc comment for `allow_threads` and emphasize the "infected closure" problem...

adamreichold:
> Just curious if switching from the vec to TLS made any observable change in the benchmark?

Note that I think the main difference here (no chance for contention, better locality, but higher chance of cold starts and more threads overall) will not be reflected by the micro benchmark which measures only the uncontended and warmed-up case and is unable to quantify system-level effects like the cost of more threads overall.

I did rerun the micro benchmark and the word-count benchmarks and in the first case, any changes were lost within the noise of switching to another thread (i.e. basically still 5µs remote versus 30ns local) and in the second case, the difference between local and remote was dwarf by the cost of Python's `ThreadPoolExecutor`.

So while I have little numbers to back this, I do think the improved locality and code simplicity of the TLS variant is the better choice.

mejrs:
First of all I very much dislike having a custom threadpool in pyo3, let alone one written with unsafe. Personally I don't feel qualified to review it and convince myself there are no soundness holes or gotchas in it - let alone make any changes or adjustments to it myself.

> Which is of course the main thing still missing: So do we actually want to do this or not? It does close the known soundness loopholes w.r.t. `allow_threads` which I think is a nice place to be and I also what I would want personally. But the cost is high, either the problematic performance profile of `safe_allow_threads` (the micro benchmark is bad, but the example usage is completely unaffected, we are still missing measurements from more realistic scenarios)

I think we shouldn't do this, see https://github.com/PyO3/pyo3/issues/3640#issuecomment-1858840518 for my reasoning. tl;dr: these issues are so niche, hard to accidentally encounter and unlikely to actually "work" that it's not worth the overhead and complexity of this solution. Also, what is more likely: someone hitting the current soundness issue, or an issue with that custom written threadpool? 

> or the unsafe and weird API of `unsafe_allow_threads` with safety invariants which are basically uncheckable if unlikely to be broken unintentionally.

`unsafe_allow_threads` is just bad api design. Its unsafety pollutes the entire program and all of its dependencies. It's hard to check that it is actually sound, and that it remains sound as the user's program grows. To me, it looks like it's just there in the case that when it *does* go wrong we can just say "well, it's unsafe and you used the unsafe keyword so it's not our fault" which is technically true but doesn't help anyone.

mejrs:
Finally I do want to thank you for all the effort you and the reviewers have put into this. I really appreciate the commitment whether or not we end up merging this 👍

adamreichold:
> First of all I very much dislike having a custom threadpool in pyo3, let alone one written with unsafe. 

Note that we are not using a global thread pool any more but rather a sort of "companion thread", i.e. each user thread will have at most one worker thread associated and there is no cross communication between these pairs of user and worker thread which simplifies the concurrency structure significantly.

The unsafety is also limited to implementing "scoped execution" for the closures, i.e. the actual communication with the worker thread is entirely safe including passing the closure and its result between user and worker thread. Hence, I would argue that this is actually somewhat simpler than say the GIL pool which we are currently maintaining.

That said, ...

> Also, what is more likely: someone hitting the current soundness issue, or an issue with that custom written threadpool?

... I don't think that us making a mistake in the implementation of dispatching those closures is the main problem here (We will probably make mistakes as elsewhere but we can fix those centrally here.), but rather the change in observable behaviour, e.g. user code which actually relies on thread-local storage to do something useful.

This why we need to have the three options on the builder for remote/safe or local/unsafe or even unconstrained/unsafe operation, so that code which does rely on TLS can continue to work (of course with the punted responsibility of only using the right kind of TLS).

Furthermore, I'd say that the status quo involving the stable-this-nightly-that `Ungil` trait with its similarity to core's rightfully unloved `Unpin` is by no means simple. Finally, we can have a very simple implementation with the same soundness properties, i.e. the first commit which basically changes a single line from `f()` to `thread::scope(|s| s.spawn(f).join().unwrap())` which is completely safe if less efficient.

But in the end, the crux of the issue IMHO is

> tl;dr: these issues are so niche, hard to accidentally encounter and unlikely to actually "work" that it's not worth the overhead and complexity of this solution.

i.e. how much do we value soundness in the sense that safe API cannot be abused. Personally, I'd say the whole story of Rust is investing additional engineering effort to allow safe code to be written without having to worry what some other code in the code base or its dependency closure does. So from my perspective, us taking on additional complexity and maintenance effort is yet another step in the basic direction that Rust embodies in the programming language domain.

davidhewitt:
As per https://github.com/PyO3/pyo3/issues/3640#issuecomment-1859007536 I feel quite strongly that we cannot do nothing here, and in particular I don't think we can leave the existing `allow_threads` as a safe API despite its flaws.

> > tl;dr: these issues are so niche, hard to accidentally encounter and unlikely to actually "work" that it's not worth the overhead and complexity of this solution.
> 
> i.e. how much do we value soundness in the sense that safe API cannot be abused. Personally, I'd say the whole story of Rust is investing additional engineering effort to allow safe code to be written without having to worry what some other code in the code base or its dependency closure does. So from my perspective, us taking on additional complexity and maintenance effort is yet another step in the basic direction that Rust embodies in the programming language domain.

I agree with this statement, I think a key part of the point of our work here is to build a watertight solution for Rust/Python interop, so that downstream code can worry about their application choices without worrying about whether our implementation is sound.

> `unsafe_allow_threads` is just bad api design. Its unsafety pollutes the entire program and all of its dependencies. It's hard to check that it is actually sound, and that it remains sound as the user's program grows.

While I agree with this too, I also think that the existing `allow_threads` is even worse. It's got carries the same unsoundness properties. If we make the argument that `unsafe_allow_threads` is difficult to verify then I think we would be straight up misleading users if we left that functionality present without making it unsafe, as then they are not even aware that they should be thinking about the soundness of their program. Just changing the exact same API to be `unsafe` greatly clarifies the functionality available to users, even if the fundamental design is no different.

> To me, it looks like it's just there in the case that when it does go wrong we can just say "well, it's unsafe and you used the unsafe keyword so it's not our fault" which is technically true but doesn't help anyone.

I think this argument can be made about any `unsafe` API. I think a more helpful framing of the issue is that here in this PR we have tried our best to build a safe construction, and we give the user the choice whether the concerns we highlight are relevant to them. Again, I think the key merit of `unsafe` here is about communication to users, not about who we can blame when there's a problem.

----

@mejrs - I think we all hope to have maintainer consensus before moving forward, so it's doubly painful given that I feel doing nothing isn't an option. Is there a reduced form of this PR or an alternative proposal that you would find more palatable?

davidhewitt:
So, following up on this a month later, my feeling is that we proceed to merge this for 0.21. It doesn't seem to me that we have any real option other than to try to make this API sound. While I don't love having three forms of what's more or less the same API this seems like the lesser of the evils we need to pick from.

mejrs:
My preferred alternative proposal is the debug mode check that the gil is held.

But, given that there are no real life cases of people running into these issues (as far as I am aware) I think just doing nothing is also acceptable.

davidhewitt:
cc @alex; as you are a security expert, any opinion you have on what we should be trying to do here would be valuable.

I definitely want to be cautious but if the consensus is a less drastic option is a better choice then I can be overridden.

alex:
(Adding to my TODO list to review tonight)

adamreichold:
> I think just doing nothing is also acceptable.

I don't think doing nothing is a good approach. We could decide to not lose the soundness loophole, but I think in that case we should remove `Ungil` as it just raises expectations that it will never be able to fulfil. And we should update our documentation to explicitly state that we do not consider this loophole to be an actual problem and will hence not fix it.

I also think that we should involve the community pro-actively if we do not close it as it does run against the general consensus established for the Rust ecosystem. Personally, I would add that I think this should be done by those advocating for not closing it as I see it as part of the necessary work to implement that decision.

alex:
On performance (and, honestly, aesthetic) grounds I hate having to run closures on a different thread. I think it takes what should be a simple operation and makes it very complex and gives unexpected performance implications.

But soundness is kind of important, it's why we like Rust :-)

So my big question is: Are there any alternative designs that might provide soundness without the performance cost? @mejrs mentioned a gil held check, but I don't think I saw the description of that.

Assuming there's not a better solution to providing soundness (and I guess we're talking here because we don't have one), I think we should do this. Yes, it's ugly. But it provides safety. Note: I haven't reviewed the PR itself, I'm not vouching for the implementation, just the concept :-)

A major motivator in my thinking is that once there's a nogil python, it obviates this question entirely: if there's no GIL, there's no need to call `allow_threads` :-) So if you think this is ugly, the best thing to do is put our energy into getting pyo3 compatible with py3.13's nogil.

davidhewitt:
> The proposal was to use a `debug_assert!`-like check which would surface the problem in dev builds but not provide soundness for release builds.

It's worth noting here that on the surface this seemed to me a little bit like how integer overflow panics only in dev builds, but rather than being UB in release integer overflow is well-defined as two's complement. So I think the choice to have soundness only in dev would be somewhat more extreme.

(We also haven't measured how much this performance would impact dev builds, nor how many places in the code we'd have to introduce this check.)

adamreichold:
An unexpected but IMHO salient data point on how theoretical this problem: https://github.com/PyO3/pyo3/discussions/3788#discussioncomment-8326779

davidhewitt:
So, I think we're quickly approaching the 0.21 release, and we have yet to make a decision here.

I still strongly believe that despite no option being appealing, moving forward with this is the most practical way to close the soundness hole and that we should do this.

I am tempted to just rebase this and see it over the line, but I'm also mindful about the amount of churn already going into 0.21 with the Bound API.

So, how about as a compromise, we let this slip one release until 0.22? That gives folks with other opinions time to gather support, and if there hasn't been a compelling alternative presented by the time we're ready to push 0.22 then we default to this option.

alex:
I'm personally ok with that.

davidhewitt:
@adamreichold are you ok if I take this PR over and push the finishing work tomorrow while I'm at the PyCon sprints?

adamreichold:
> @adamreichold are you ok if I take this PR over and push the finishing work tomorrow while I'm at the PyCon sprints?

I would be rather grateful if find the time to do this! I have been dreading rebasing this for some time now...

davidhewitt:
Well in the end time shot past me but I'll aim to rebase this in the very near future so we can ship 0.22

davidhewitt:
Ok, first conflict resolution done, I think there will be some finishing work which I'll try later.

davidhewitt:
Ok, this now has all conflicts / tests fixed. Just needs some docs, which I'll aim to do in the very near future.

PR #3647, opened 2023-12-14
Fix the Crossbeam ecosystem to point releases before it required Rust 1.61.

adamreichold:


davidhewitt (approved):
Assuming CI passes, LGTM, thanks👍

Issue #3633, opened 2023-12-08
Provide fallbacks for datetime-related API when Py_LIMITED_API is enabled


Tpt:
Currently `datetime`-related APIs are only available when the `Py_LIMITED_API` is not enabled.

This includes:
* [PyDate](https://docs.rs/pyo3/latest/pyo3/types/struct.PyDate.html)
* [PyDateTime](https://docs.rs/pyo3/latest/pyo3/types/struct.PyDateTime.html)
* [PyDelta](https://docs.rs/pyo3/latest/pyo3/types/struct.PyDelta.html)
* [PyTime](https://docs.rs/pyo3/latest/pyo3/types/struct.PyTime.html)
* [PyTzInfo](https://docs.rs/pyo3/latest/pyo3/types/struct.PyTzInfo.html)
* [PyDateAccess](https://docs.rs/pyo3/latest/pyo3/types/trait.PyDateAccess.html)
* [PyDeltaAccess](https://docs.rs/pyo3/latest/pyo3/types/trait.PyDeltaAccess.html)
* [PyTimeAccess](https://docs.rs/pyo3/latest/pyo3/types/trait.PyTimeAccess.html)
* [PyTzInfoAccess](https://docs.rs/pyo3/latest/pyo3/types/trait.PyTzInfoAccess.html)
* [timezone_utc](https://docs.rs/pyo3/latest/pyo3/types/fn.timezone_utc.html)

Would a fallback using Python interpreter calls like `py.import('datetime')?.getattr('timezone')?.getattr('utc')` be welcome into PyO3 when `Py_LIMITED_API` is enabled? I think it might make easier for some PyO3 users to use abi3. I am willing to contribute it.

davidhewitt:
Yes I think that would be a useful thing to support 👍

adamreichold:
The `chrono` conversions now work for abi3 builds which is what we decided to pursue here in the end. Hence closing this.

PR #3648, opened 2023-12-14
Adds internal timezone_from_offset function

Tpt:
Adds internal timezone_from_offset function

It allows to build conversions from chrono without direct access to the C API

### DIFF
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -21,8 +21,10 @@ use crate::ffi::{
 };
 use crate::instance::PyNativeType;
 use crate::types::PyTuple;
-use crate::{IntoPy, Py, PyAny, Python};
+use crate::{AsPyPointer, IntoPy, Py, PyAny, Python};
 use std::os::raw::c_int;
+#[cfg(feature = "chrono")]
+use std::ptr;
 
 fn ensure_datetime_api(_py: Python<'_>) -> &'static PyDateTime_CAPI {
     unsafe {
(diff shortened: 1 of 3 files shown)


Tpt:
This change is a first step for #3633. It allows to remove all C API calls from the `chrono` conversion.

I am not sure adding two functions is the best way to go. An other option might be to convert the three functions constructing `datetime.timezone` to associated functions of the `PyTzInfo` object.

src/types/datetime.rs
```diff
@@ -487,6 +488,32 @@ pub fn timezone_utc(py: Python<'_>) -> &PyTzInfo {
     unsafe { &*(ensure_datetime_api(py).TimeZone_UTC as *const PyTzInfo) }
 }
 
+/// Equivalent to `datetime.timezone` constructor
+///
+/// Use [`timezone_from_offset_and_name`] for a named timezones.
+pub fn timezone_from_offset<'a>(py: Python<'a>, offset: &PyDelta) -> PyResult<&'a PyTzInfo> {
```
adamreichold, line 494: My main question here is why this has to become new public API if it is mainly meant as extension point to differentiate between accessing unlimited and limited API? Wouldn't `pub(crate)` suffice for that purpose?

davidhewitt:
Note that there is a long history of why these functions don't exist, see https://github.com/PyO3/pyo3/pull/1588#issuecomment-1174614451

I'm not totally against adding them, but also do understand why others did not want them.

Tpt:
> Note that there is a long history of why these functions don't exist, see [#1588 (comment)](https://github.com/PyO3/pyo3/pull/1588#issuecomment-1174614451)
> 
> I'm not totally against adding them, but also do understand why others did not want them.

Thank you! I was not aware of this discussion. The concerns make sense. I have rewritten this MR to make `timezone_from_offset` private and I dropped `timezone_from_offset_and_name`.

src/types/datetime.rs
Tpt: Yes, indeed. Done.

newsfragments/3648.changed.md
davidhewitt, line 1: No need to document an internal refactor; I've added the skip-changelog label. Please remove this fragment 😊

davidhewitt (approved):
Overall this now looks like it just relocated the function, but I've no problem with that if it's a helpful first step for the follow up feature work 👍

newsfragments/3648.changed.md
Tpt: Done.

davidhewitt (approved):

PR #3649, opened 2023-12-14
hold onto module name properly in `PyCFunction::internal_new`

davidhewitt:
While going through the rewrites in #3606, I spotted that there's a slight "oops" in `PyCFunction::internal_new` where we do basically `m.name()?.into_py(py).as_ptr()` to create a dangling pointer. We get away with this at present because `.into_py(py)` uses `PyString::new` and puts the string into the GIL pool, thus despite the dangling pointer the reference count is still one for the duration we need it.

One the pool bandaid comes off, this bit of code segfaults.

Opened as a review separately because I think this is trivially reviewable and mergeable.

### DIFF
--- a/src/types/function.rs
+++ b/src/types/function.rs
@@ -4,7 +4,7 @@ use crate::prelude::*;
 use crate::{
     ffi,
     impl_::pymethods::{self, PyMethodDef},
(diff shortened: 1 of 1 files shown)


adamreichold (approved):
Great catch!

PR #3651, opened 2023-12-14
implement `PyBytesMethods` and `PyByteArrayMethods`

davidhewitt:
Implements `PyBytesMethods` and `PyByteArrayMethods` traits as per #3382 

The main interesting thing here is that to be able to forward the `.as_bytes()` and `.as_bytes_mut()` methods from the gil refs to the underlying types, I had to extend the lifetime of the returned slices. Unclear to me if there's a way to change things to not need this.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -217,6 +217,96 @@ unsafe impl<T> AsPyPointer for Py2<'_, T> {
     }
 }
 
+/// A borrowed equivalent to `Py2`.
+///
+/// The advantage of this over `&Py2` is that it avoids the need to have a pointer-to-pointer, as Py2
+/// is already a pointer to an `ffi::PyObject``.
+#[repr(transparent)]
+pub(crate) struct Py2Borrowed<'a, 'py, T>(
+    NonNull<ffi::PyObject>,
+    PhantomData<&'a Py<T>>,
+    Python<'py>,
+);
+
+impl<'a, 'py> Py2Borrowed<'a, 'py, PyAny> {
+    /// # Safety
+    /// This is similar to `std::slice::from_raw_parts`, the lifetime `'a` is completely defined by
+    /// the caller and it's the caller's responsibility to ensure that the reference this is
+    /// derived from is valid for the lifetime `'a`.
+    pub(crate) unsafe fn from_ptr_or_err(
+        py: Python<'py>,
+        ptr: *mut ffi::PyObject,
+    ) -> PyResult<Self> {
+        NonNull::new(ptr).map_or_else(
+            || Err(PyErr::fetch(py)),
+            |ptr| Ok(Self(ptr, PhantomData, py)),
+        )
+    }
+
+    /// # Safety
+    /// This is similar to `std::slice::from_raw_parts`, the lifetime `'a` is completely defined by
+    /// the caller and it's the caller's responsibility to ensure that the reference this is
+    /// derived from is valid for the lifetime `'a`.
+    pub(crate) unsafe fn from_ptr_or_opt(py: Python<'py>, ptr: *mut ffi::PyObject) -> Option<Self> {
+        NonNull::new(ptr).map(|ptr| Self(ptr, PhantomData, py))
+    }
+
+    /// # Safety
+    /// This is similar to `std::slice::from_raw_parts`, the lifetime `'a` is completely defined by
+    /// the caller and it's the caller's responsibility to ensure that the reference this is
+    /// derived from is valid for the lifetime `'a`.
+    pub(crate) unsafe fn from_ptr(py: Python<'py>, ptr: *mut ffi::PyObject) -> Self {
+        Self(
+            NonNull::new(ptr).unwrap_or_else(|| crate::err::panic_after_error(py)),
+            PhantomData,
+            py,
+        )
+    }
+}
+
+impl<'a, 'py, T> From<&'a Py2<'py, T>> for Py2Borrowed<'a, 'py, T> {
+    /// Create borrow on a Py2
+    fn from(instance: &'a Py2<'py, T>) -> Self {
+        Self(
+            unsafe { NonNull::new_unchecked(instance.as_ptr()) },
+            PhantomData,
+            instance.py(),
+        )
+    }
+}
+
+impl<'py, T> Py2Borrowed<'py, 'py, T>
+where
+    T: HasPyGilRef,
+{
+    pub(crate) fn from_gil_ref(gil_ref: &'py T::AsRefTarget) -> Self {
+        // Safety: &'py T::AsRefTarget is expected to be a Python pointer,
+        // so &'py T::AsRefTarget has the same layout as Self.
+        unsafe { std::mem::transmute(gil_ref) }
+    }
+
+    // pub(crate) fn into_gil_ref(self) -> &'py T::AsRefTarget {
+    //     // Safety: self is a borrow over `'py`.
+    //     unsafe { self.py().from_borrowed_ptr(self.0.as_ptr()) }
+    // }
+}
+
+impl<T> std::fmt::Debug for Py2Borrowed<'_, '_, T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Py2::fmt(self, f)
+    }
+}
+
+impl<'py, T> Deref for Py2Borrowed<'_, 'py, T> {
+    type Target = Py2<'py, T>;
+
+    #[inline]
+    fn deref(&self) -> &Py2<'py, T> {
+        // safety: Py2 has the same layout as NonNull<ffi::PyObject>
+        unsafe { &*(&self.0 as *const _ as *const Py2<'py, T>) }
+    }
+}
+
 /// A GIL-independent reference to an object allocated on the Python heap.
 ///
 /// This type does not auto-dereference to the inner object because you must prove you hold the GIL to access it.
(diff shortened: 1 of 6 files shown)


src/types/bytearray.rs
```diff
@@ -200,7 +202,9 @@ impl PyByteArray {
     /// apply to this function as well.
     #[allow(clippy::mut_from_ref)]
     pub unsafe fn as_bytes_mut(&self) -> &mut [u8] {
-        slice::from_raw_parts_mut(self.data(), self.len())
+        let slice = Py2::borrowed_from_gil_ref(&self).as_bytes_mut();
+        // SAFETY: &self guarantees the reference is alive long enough
+        unsafe { slice::from_raw_parts_mut(slice.as_mut_ptr(), slice.len()) }
```
adamreichold, line 207: This is unfortunate. Can we maybe generalize the reasoning? For example into something like

```rust
pub fn borrowed_from_gil_ref(gil_ref: &'py T::AsRefTarget) -> &'py Self;
```

to at least produce `&'py Py2<'py, T::AsRefTarget>` to avoid having to do this in each implementation? Is that even valid? I guess the GIL pool should keep the reference alive for the whole GIL lifetime, shouldn't it?

davidhewitt: Not quite like that because `&'py gil_ref` is a pointer, but `&'py Self` is a pointer to a pointer. I think we might be able to do that with a closure?

```rust
pub fn borrow_from_gil_ref<R>(gil_ref: &'py T::AsRefTarget, f: impl FnOnce(&'py Self) -> R) -> R;
```

I also have a type `Py2Borrowed<'a, 'py, T>(NonNull<ffi::PyObject>, PhantomData<&'a Py2<'py, T>>)` which _is_ pointer layout and necessary to avoid introducing reference count increases on e.g. `PyTuple::get_item`. That can implement `from_gil_ref(gil_ref: &'py T::AsRefTarget) -> Py2Borrowed<'py, 'py, T>;` soundly. Let me also push that into a separate PR so you can browse it.

davidhewitt: Ah no, the closure doesn't work, because of `'py` lifetime on the reference it can legally escape the closure:

```rust
let mut borrow = None;
Py2::<PyBytes>::borrow_from_gil_ref(&self, |b| {
    borrow = Some(b);
});
dbg!(borrow);  // this blows up
```

davidhewitt: Ah, I think I see a way to make this work using `Py2Borrowed` as a base implementation, I'll try to push that later this evening after my son's bedtime...

src/types/bytes.rs
```diff
@@ -102,16 +128,9 @@ impl Py<PyBytes> {
...
-        // isn't really a `PyBytes`, but better safe than sorry.
-        unsafe {
-            let buffer = ffi::PyBytes_AsString(self.as_ptr()) as *const u8;
-            let length = ffi::PyBytes_Size(self.as_ptr()) as usize;
-            debug_assert!(!buffer.is_null());
-            std::slice::from_raw_parts(buffer, length)
-        }
+        // Safety: `Py2Borrowed` has the same layout as `Py`
+        let borrowed: Py2Borrowed<'a, '_, PyBytes> = unsafe { std::mem::transmute_copy(self) };
+        borrowed.as_bytes()
```
davidhewitt, line 133: Ok, so this didn't work out as nicely as I hoped. I _thought_ I could have the following implementation:

```rust
    pub fn as_bytes<'a>(&'a self, py: Python<'_>) -> &'a [u8] {
        Py2Borrowed::from(self.attach(py)).as_bytes()
    }
```

... but the compiler wants to have an explicit lifetime `Python<'py>` with a constraint `'py: 'a`. I feel like that might be a technical weakness in the current borrow checker which could be overcome with time, but it could be that I've misunderstood some fine detail about lifetimes. I'll try to ask language folks about this on Zulip.

davidhewitt: Ok, so I asked in https://users.rust-lang.org/t/is-this-a-limitation-of-the-current-borrow-checker/103979/3 and I think I have some ideas now, so I'll come back and push a solution later today.

davidhewitt: Ok, this now works; I had to adjust the definition of the ``Py2Borrowed` struct slightly, no unsafe needed any more :)

src/instance.rs
```diff
@@ -724,6 +858,10 @@ impl<T> Py<T> {
         Py2(py, ManuallyDrop::new(self))
     }
 
+    pub(crate) fn attach_borrow<'a, 'py>(&'a self, py: Python<'py>) -> Py2Borrowed<'a, 'py, T> {
```
adamreichold, line 861: Reposted from #3654: But doesn't attach_borrow sort of require the immutability guarantee of Py<PyBytes>::as_bytes now to avoid unsafely extending references into e.g. mutable pyclasses beyond the duration for which the GIL is held? Wouldn't this imply that attach_borrow needs to be unsafe?

davidhewitt: Can you elaborate a bit on a the safety violation you see? I might be missing something here.

I think you're right that `attach_borrow` isn't practically useful except in the `Py<PyBytes>` kind of situations, but I don't see how creating a `Py2Borrowed` with lifetime `'a` longer than `'py` is in itself problematic.

adamreichold: I think this would mainly be relevant for Python implemented using interior mutability: Let's say we have `Py2<'py, T>` with `fn borrow<'a>(&'a self) -> RefWrapper<'a>` which set an internal borrow flag protected by the GIL. This is fine as `&'a Py2<'py, T>` requires that `'py` outlives `'a`, i.e. the GIL is held while the returned reference wrapper is alive.

But now, I can use `attach_borrow` to add `fn borrow<'a>(&'a self) -> RefWrapper<'a>` to `Py<T>` giving me a reference wrapper which will reset the internal borrow flag after the GIL was already released, racing with other threads having the GIL and trying to get a borrow themselves.

Of course, this is not public so we could just declare "don't do that", but I think we should try to help ourselves avoid mistakes as well. Having a reference into an object on the Python heap that outlives holding the GIL is so unexpected/fragile, that I would prefer making the constructor of such a type unsafe.

davidhewitt: If I follow your wording correctly, then the construct which concerns you is this:

```rust
impl Py<T> {
    fn borrow<'a>(&'a self, py: Python<'_>) -> RefWrapper<'a> {
         self.attach_borrow(py).borrow()
    }
}
```

Based on the lessons I've learned wrangling with these lifetimes trying to get this all right, I'm reasonably sure we're ok and that this won't compile.

The reason is that `self.attach_borrow(py).borrow()` would create a temporary `Py2Borrowed`, which we then need to deref to access the `borrow` method of `Py2`. Therefore the lifetime of the returned `RefWrapper` is not long enough, as it would be limited to the deref of the temporary.

To make the above `fn borrow` compile, I think we'd need the following method on `Py2Borrowed` to consume the temporary and spit out an appropriately long `RefWrapper`:

```rust
impl<'a, T> Py2Borrowed<'a, '_, T> {
    fn borrow(self) -> RefWrapper<'a> {
        /* ... */
    }
}
```

And I think we should be able to catch in review that that method can't be sound, because it spits out a `RefWrapper` ignoring the GIL lifetime.

adamreichold: Yes, you are right, I was missing the `as_bytes` method directly on `Py2Borrow` which is required to make this possible. And indeed without it, `error[E0515]: cannot return value referencing temporary value` prevents anything bad from happening.

src/instance.rs
```diff
@@ -214,6 +225,129 @@ unsafe impl<T> AsPyPointer for Py2<'_, T> {
...
+/// A borrowed equivalent to `Py2`.
+///
+/// The advantage of this over `&Py2` is that it avoids the need to have a pointer-to-pointer, as Py2
+/// is already a pointer to a PyObject.
+#[repr(transparent)]
+pub(crate) struct Py2Borrowed<'a, 'py, T>(
+    NonNull<ffi::PyObject>,
+    PhantomData<&'a Py<T>>,
+    Python<'py>,
+); // TODO is it useful to have the generic form?
```
adamreichold, line 237: This `TODO` should go away because we are already using it for `PyBytes`.

adamreichold (approved):
LGTM except for the stray `TODO` comment

src/instance.rs
davidhewitt: Agreed, I may also delete / comment away some of the other `Py2Borrowed` functionality not needed strictly for this PR. Hopefully can find time later tonight.

davidhewitt:
Thanks again for all the reviews this week, I'm really excited to have some momentum on the new API!

src/types/bytearray.rs
```diff
@@ -188,7 +188,9 @@ impl PyByteArray {
     /// }
     /// ```
     pub unsafe fn as_bytes(&self) -> &[u8] {
-        slice::from_raw_parts(self.data(), self.len())
+        let slice = Py2::borrowed_from_gil_ref(&self).as_bytes();
+        // SAFETY: &self guarantees the reference is alive long enough
+        unsafe { slice::from_raw_parts(slice.as_ptr(), slice.len()) }
```
adamreichold, line 193: Couldn't these two use some form of `Py2Borrowed` as well to extend the lifetime of the slice?

src/types/bytes.rs
```diff
@@ -88,6 +89,31 @@ impl PyBytes {
...
+impl<'py> PyBytesMethods<'py> for Py2<'py, PyBytes> {
+    #[inline]
+    fn as_bytes(&self) -> &[u8] {
+        Py2Borrowed::from(self).as_bytes()
+    }
+}
+
+impl<'a> Py2Borrowed<'a, '_, PyBytes> {
+    /// Gets the Python string as a byte slice.
+    fn as_bytes(self) -> &'a [u8] {
```
adamreichold, line 116: And Clippy is being a bit of a nag here. Maybe just `#[allow(clippy::wrong_self_convention)]`?

adamreichold:
@davidhewitt Won't this fail due to the Clippy warning for the self convention on `Py2Borrowed<PyBytes>::as_bytes`?

davidhewitt:
Oh yikes, probably, yes. I've cancelled the merge and will fixup. Good catch!

src/types/bytearray.rs
davidhewitt: Yes, will fixup with clippy. 👍

PR #3652, opened 2023-12-14
implement `PyBoolMethods`

davidhewitt:
Implements `PyBoolMethods` trait as per #3382 

This one will remain in draft, because `Py2::<PyBool>::borrowed_from_gil_ref(&self).is_true()` requires the `PyBool` in the turbofish there to avoid ambiguity. I'm going to open a PR which might offer a way to avoid this.

### DIFF
--- a/src/types/boolobject.rs
+++ b/src/types/boolobject.rs
@@ -1,6 +1,8 @@
 #[cfg(feature = "experimental-inspect")]
 use crate::inspect::types::TypeInfo;
-use crate::{ffi, FromPyObject, IntoPy, PyAny, PyObject, PyResult, Python, ToPyObject};
+use crate::{
+    ffi, instance::Py2, FromPyObject, IntoPy, PyAny, PyObject, PyResult, Python, ToPyObject,
+};
 
 /// Represents a Python `bool`.
 #[repr(transparent)]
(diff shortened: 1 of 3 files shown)


src/types/boolobject.rs
```diff
@@ -18,6 +20,24 @@ impl PyBool {
...
     pub fn is_true(&self) -> bool {
+        Py2::<PyBool>::borrowed_from_gil_ref(&self).is_true()
+    }
+}
+
+/// Implementation of functionality for [`PyBool`].
+///
+/// These methods are defined for the `Py2<'py, PyBool>` smart pointer, so to use method call
+/// syntax these methods are separated into a trait, because stable Rust does not yet support
+/// `arbitrary_self_types`.
```
birkenfeld, line 31: I know I'm late to this party, but why does
```
impl<'py> Py2<'py, PyBool> {
    ...
}
```
not work?

adamreichold: IIRC, the main problem with that approach was documentation as this would make all methods for all types end up documentation on single page for the `Py2` type.

davidhewitt: That, plus we end up leaving `rust_numpy` and other downstream crates out in the cold, as they cannot add methods to `Py2<T>`. This approach works consistently for everyone.

My long-term hope is that arbitrary self types land and we can remove the traits more or less without anyone noticing.

birkenfeld: :+1:

davidhewitt:
Rebased this now that #3653 is in; this should be a fairly straightforward review hopefully :)

adamreichold (approved):

PR #3653, opened 2023-12-14
Add `AsRefSource` to `PyNativeType`.

davidhewitt:
This adds an associated type `AsRefSource` to `PyNativeType`.

This allows type inference in the reverse direction `&T -> Py<T>`, which is particularly useful for `&PyCell<T> -> Py2<T>` . It also helps for forwarding our gil-ref methods to the new traits, e.g. by solving the inference problem in #3652 

Really this is most helpful when we're in a transition phase where users might be using `Py2::borrowed_from_gil_ref` themselves. So I'm also happy to decide that we don't merge this until we're ready to release `Py2` as a whole, and for now we just continue to add type annotations while we're landing the groundwork internally.

### DIFF
--- a/newsfragments/3653.changed.md
+++ b/newsfragments/3653.changed.md
@@ -1,0 +1,1 @@
+Add `AsRefSource` to `PyNativeType`.
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -24,6 +24,9 @@ use std::ptr::NonNull;
 ///
 /// This trait must only be implemented for types which cannot be accessed without the GIL.
 pub unsafe trait PyNativeType: Sized {
+    /// The form of this which is stored inside a `Py<T>` smart pointer.
+    type AsRefSource: HasPyGilRef<AsRefTarget = Self>;
+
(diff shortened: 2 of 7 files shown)


adamreichold (approved):
I don't think we should hold off on the quality-of-life improvements for internal work as it makes this faster.

I am currently not seeing a downside besides a general unease with more coupling within our already complex trait landscape. Is there anything in particular that one should be aware of?

davidhewitt:
Thanks; yes this makes migration work nicer for sure. I think the only downside is that manual implementors of `PyNativeType` will need to add this associated type for very little immediate benefit, depending on whether the `Py2` stuff makes it into 0.21 (still unsure how fast we're going to manage to go).

In the longer term, I think we can throw away the whole `PyNativeType` trait when we get rid of the pool and gil refs. So we have to ride over a peak of complexity in the short term and can then simplify again.

davidhewitt:
(it should be noted that I don't expect there to be downstream manual implementors of `PyNativeType`, but you never know...)

PR #3655, opened 2023-12-15
Rename `Py<T>` to `PyDetached<T>`

davidhewitt:
To make space for my preferred name of `Py<'py, T>` for the new API in #3382, this PR proposes renaming `Py<T>` to `PyDetached<T>`. (Other ideas are `PySend<T>` and `PyUngil<T>`.)

Benefit of opening a PR is so we can see how painful it looks to do this rename. (And maybe so that internally we can stop writing `Py2<...>` everywhere.)

I'm split on whether it makes sense to land this rename in 0.21 and then the new `Py<'py, T>` in 0.22, or land both at the same time.

### DIFF
--- a/examples/decorator/src/lib.rs
+++ b/examples/decorator/src/lib.rs
@@ -13,7 +13,7 @@ pub struct PyCounter {
     count: Cell<u64>,
 
     // This is the actual function being wrapped.
-    wraps: Py<PyAny>,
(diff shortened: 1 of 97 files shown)


davidhewitt:
I think from staring at the diff, this is enough of an ergonomics hit that we should not do this until we're ready to make the new API public.

PR #3656, opened 2023-12-15
Draft: Makes PyDelta compatible with abi3

Tpt:
This is a draft to get feedback before writing to much code

Open questions:
- is the way the basic conversion traits are implemented good? The challenge is that `PyDateTime_Delta` and `PyDelta_Check` are not in ABI3. I plan to make it a macro to duplicate for the other types.
- what is the best thing to do with the delta `normalize` constructor parameter that is always `true` in the constructor exposed to Python. Ignore the value? Print a Python warning? A panic? I would tend to just document the limitation and don't do anything at runtime.

### DIFF
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -3,29 +3,30 @@
 //! For more details about these types, see the [Python
 //! documentation](https://docs.python.org/3/library/datetime.html)
 
+#[cfg(not(Py_LIMITED_API))]
 use crate::err::PyResult;
+#[cfg(not(Py_LIMITED_API))]
 use crate::ffi::{
-    self, PyDateTime_CAPI, PyDateTime_FromTimestamp, PyDateTime_IMPORT, PyDate_FromTimestamp,
-};
-use crate::ffi::{
-    PyDateTime_DATE_GET_FOLD, PyDateTime_DATE_GET_HOUR, PyDateTime_DATE_GET_MICROSECOND,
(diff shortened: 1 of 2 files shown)


src/types/datetime.rs
```diff
@@ -504,13 +529,54 @@ pub fn timezone_from_offset<'a>(py: Python<'a>, offset: &PyDelta) -> PyResult<&'
...
 pub struct PyDelta(PyAny);
+#[cfg(not(Py_LIMITED_API))]
 pyobject_native_type!(
     PyDelta,
     crate::ffi::PyDateTime_Delta,
     |py| ensure_datetime_api(py).DeltaType,
     #module=Some("datetime"),
     #checkfunction=PyDelta_Check
 );
+// TODO: convert to a macro
```
adamreichold, line 540: I am not sure whether we should try to have a wrapper `PyDelta` if only the limited API is available, i.e. does add little beyond `PyAny`. Wasn't the aim here to support the chrono conversions? If so, I think it would be easier (and more consistent with other conversions, c.f. e.g. `mod num_complex`) to have separate code paths to do the conversion, but not trying to provide a `mod datetime` that internally just calls the Python API.

Tpt: Yes, that's a good point. Let's close this PR and just implement fallback in `chrono` conversions

Tpt: Thank you!

davidhewitt: (This is probably a result of me not being very specific in https://github.com/PyO3/pyo3/issues/3633#issuecomment-1846844225)

PR #3657, opened 2023-12-16
Transition is_true to is_truthy to clarify that the test is not based on identity with or equality to the True singleton.

adamreichold:
Based on the discussion in https://github.com/PyO3/pyo3/pull/3638#issuecomment-1852134420 and following.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -787,7 +787,15 @@ impl<T> Py<T> {
     /// Returns whether the object is considered to be true.
     ///
     /// This is equivalent to the Python expression `bool(self)`.
+    #[deprecated(since = "0.21.0", note = "use `.is_truthy()` instead")]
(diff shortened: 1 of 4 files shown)


src/instance.rs
```diff
@@ -787,7 +787,15 @@ impl<T> Py<T> {
     /// Returns whether the object is considered to be true.
     ///
     /// This is equivalent to the Python expression `bool(self)`.
+    #[deprecated(since = "0.20.0", note = "use `.is_truthy()` instead")]
```
davidhewitt, line 790:
```suggestion
    #[deprecated(since = "0.21.0", note = "use `.is_truthy()` instead")]
```

davidhewitt (approved):
Thanks! Just one tiny typo!

PR #3660, opened 2023-12-18
 Rename name to qualname and add name to better match Python and ensure abi3 fallbacks. 

adamreichold:


src/types/typeobject.rs
```diff
@@ -35,6 +38,29 @@ impl PyType {
         self.getattr(intern!(self.py(), "__qualname__"))?.extract()
     }
 
+    /// Gets the full name including the module of the `PyType`.
+    pub fn full_name(&self) -> PyResult<Cow<'_, str>> {
```
adamreichold, line 42: I can see this replacing `name`, but on the other hand, this appears to be needless churn.

davidhewitt:
Possibly relevant here: we should check out `PyType_GetName` and `PyType_GetQualName` added in 3.11. Maybe they solve the abi3 case (for 3.11 and up)?

src/types/typeobject.rs
davidhewitt: I'd be quite happy to have `.name()` and `.qualname()` if we didn't mind a bit of churn to match CPython better

adamreichold:
> Possibly relevant here: we should check out `PyType_GetName` and `PyType_GetQualName` added in 3.11. Maybe they solve the abi3 case (for 3.11 and up)?

For `name`, it would work by using `GetQualName`. But for `full_name` we are still missing the `__module__` part which is also derived from `tp_name`. I am not sure I can get that from elsewhere? Will certainly push a commit to use those two instead of attribute look-ups where applicable.

src/types/typeobject.rs
davidhewitt: I suppose we could call them `get_name` and `get_qualname` to avoid silent changes?

adamreichold: Would work, but would also be unidiomatic. I did change the return type form `&str` to `String` which might already induce some breakage (to be compatible with a `Py2`-based implementation). Let me know if this (and the migration guide entry pushed here) seems sufficient for you. Otherwise, I'll amend the commit to use the `get_` prefixes.

src/types/typeobject.rs
```diff
@@ -30,9 +33,52 @@ impl PyType {
...
+        };
+
+        name
+    }
+
+    /// Gets the full name, which includes the module, of the `PyType`.
+    pub fn name(&self) -> PyResult<Cow<'_, str>> {
+        #[cfg(not(any(Py_LIMITED_API, PyPy)))]
+        {
+            let name = unsafe { CStr::from_ptr((*self.as_type_ptr()).tp_name) }.to_str()?;
```
davidhewitt, line 60: Now I stare at this, I wonder if we need to be defensive against reassignments to `tp_name`? Looking at CPython source, it looks like this is allowed if the type doesn't have flag `Py_TPFLAGS_IMMUTABLETYPE`.

src/types/typeobject.rs
```diff
@@ -30,9 +33,52 @@ impl PyType {
...
+
+            Ok(Cow::Borrowed(name))
+        }
+
+        #[cfg(any(Py_LIMITED_API, PyPy))]
+        {
```
davidhewitt, line 66: It would have been nice for this branch to use `PyType_GetSlot()`, but it looks like there's no slot identifier for `tp_name`. Maybe we can ask for this in 3.13 :)

guide/src/migration.md
```diff
@@ -74,6 +74,9 @@ Python::with_gil(|py| {
 });
 ```
 
+### `PyType::name` is now `PyType::qualname`
+
+`PyType::name` has been renamed to `PyType::qualname` to indicate that it does indeed return the [qualified name](), matching the `__qualname__` attribute. The newly added `PyType::name` yields the full name including the module name now which corresponds to `__module__.__name__` on the level of attributes.
```
davidhewitt, line 79: Looks like a link is missing here.

src/types/typeobject.rs
davidhewitt: Yep that looks okay to me 👍

davidhewitt (commented):
Yep, I think all of this makes sense, thanks 👍 

Just a few short comments, and one slight concern about lifetime of the borrow.

src/types/typeobject.rs
adamreichold: So copy out the value if the flag is not set?

adamreichold: This flags seems to exist only for 3.10+, so always copy out before that version?

davidhewitt: Yeah, I suspect that's probably the correct approach 👍

src/types/typeobject.rs
davidhewitt: Ah I see, that's unfortunate. Good find👍

davidhewitt (approved):
Looks good to me! Thanks for the revisions here; this worked out more fiddly than I think either of us expected 👀

Issue #2935, opened 2023-02-03
RFC: remove default of None for trailing optional arguments


davidhewitt:
In #2934 I've documented a longstanding PyO3 behaviour to automatically add a default of `None` to all trailing `Option<T>` arguments.

I think that behaviour is a small footgun, as I don't think first-time users necessarily expect this. I see two downsides:
- Users accidentally allow default of `None` for these arguments when they wanted them to be required
- Users reordering function arguments can easily remove the default without noticing.

We currently have the deprecation from #2703 which we can upgrade to an error in 0.20. This helps the situation, as it makes the refactoring in the second bullet fail nosily.

I'd be keen to remove the trailing default, with a deprecation warning for functions using this behaviour which don't currently have a `#[pyo3(signature = (...))]` annotation. I think it's a simpler API for users if PyO3 never adds an implicit default, and removes the need for the error on the refactoring case (which is purely a guard rail to avoid the current footgun).

What do others think? I'd be keen to hear from folks who think the conveniences of the current behaviour (i.e. fewer macro annotations) outweigh the drawbacks listed above. (@mejrs - in #2193 I got the impression that you might hold this view.)

birkenfeld:
IMO the current mode (with the warning elevated to error) is fine. With arguments that can be `None`, it's pretty typical that they also have a default of `None`.

davidhewitt:
@ritchie46 - after reading #2925 I'd be interested in your comment as a user - you were either unaware of this default, or would prefer the change proposed here?

davidhewitt:
#2932 sounds like another case of a user confused by this (maybe just because it wasn't documented).

birkenfeld:
> I'm generally against doing implicit things, so this sounds good.

Although "implicit" can mean a lot of different things.

In this case, you could argue that any case of not using `pyo3(signature)` is implicit, and `pyo3(signature)` should be mandatory.

ritchie46:
> @ritchie46 - after reading #2925 I'd be interested in your comment as a user - you were either unaware of this default, or would prefer the change proposed here?

I was unaware of the defaults. I think having a well documented default makes sense. It can remove some boilerplate for the default case.

davidhewitt:
> In this case, you could argue that any case of not using `pyo3(signature)` is implicit, and `pyo3(signature)` should be mandatory.

[...]

PR #3661, opened 2023-12-18
Replace (A)IterNextOutput by autoref-based specialization to allow returning arbitrary value

### LINKED ISSUE #3190: `__anext__` should be able to return `&PyAny`

adamreichold:
To be done:
- [x] changelog entry
- [x] migration guide entry

Closes #3190


pyo3-macros-backend/src/pymethod.rs
```diff
@@ -987,17 +989,23 @@ fn extract_object(
...
             ReturnMode::Conversion(conversion) => quote! {
-                let _result: _pyo3::PyResult<#conversion> = #call;
+                let _result: _pyo3::PyResult<#conversion> = _pyo3::callback::convert(py, #call);
                 _pyo3::callback::convert(py, _result)
             },
+            ReturnMode::SpecializedConversion(traits, tag) => quote! {
+                let _result = #call;
+                use _pyo3::callback::{#traits};
+                (&_result).#tag().convert(py, _result)
+            },
```
adamreichold, line 1006: @davidhewitt This is by no means complete but I wanted to run the design by you before I go off into the weeds.

So as per the discussion in #3190, I tried to remove any special handling from `__next__` so that if user code wants to stop iteration, I can just return a `PyStopIteration` error. `Option` is still handled separately though using autoref-based specialization. Here, I opted to deprecate the format but I am not actually sure whether we want to fully deprecate this or just keep the specialized behaviour?

Furthermore, I am not sure whether I messed up the return value conversion by delaying the call to `convert` but this was the only way I was able to make type inference work out. I also wonder whether `PyResult<Option<T>>` is another form we should consider supporting now?

davidhewitt:
Fantastic! I am a little tight for time today, I'll do my best to give this a full read tomorrow 👍

davidhewitt:
I've had a read and makes sense, I'll hold off from too many comments on the diff while we're figuring out the design.

I would like to suggest we don't deprecate the `Option<T>`  special case. As per https://github.com/PyO3/pyo3/issues/3190#issuecomment-1721269643 I think we can use the `None` variant for efficiency of returning `NULL`. I also think the fact that Rust iterators return `Option` will make this specialization very attractive in user code.

And I guess if we did keep that, probably `PyResult<Option<T>>` also needs to specialise in the same way? Rust fallible iterators of `Option<Result<...>>` can probably then just `.transpose()` away the error. Or would we want to specialise `Option<Result<..>>` directly? 🤔

pyo3-macros-backend/src/pymethod.rs
davidhewitt: I think delaying the call to `convert` is fine in all cases except when we use temporary `PyRef` and `PyRefMut` in the expression to extract `&T` and `&mut T` class types.

I wonder whether we could abuse `FnOnce::Output` to inspect the type for tag inference before we actually create the value. I can try to express that further when I'm back at my computer.

davidhewitt:
If you are not fond of specialization introducing subtle behavioural difference based on return type, maybe rather than specialize on `Option<T>` we could have our own enum to specialize on, with handy traits to convert options and options-of-results to it? (And deprecate the `Option<T>` pathway as you currently have done here)

davidhewitt:
> If I understand this correctly, I would need to remove the generic `Target` and enforce `*mut ffi::PyObject` for that to work which should work since the specialization is used only for this single slot type?

I think so, yes. Given the tags only work for this slot, I can't see a potential downside right now, and as a performance optimization I guess we can back out again if it's problematic.

davidhewitt:
> I already added `PyResult<Option<T>>`. I think adding `Option<Result<..>>` would require a second layer of specialization and be even harder to explain, so I would like to avoid adding it and defer Rust users to `transpose` (which suitable guide entries). Especially since even the conversion trait bounds themselves will be a bit messy due to the need to convert the error type as most iterators will probably not yield `PyErr` directly.

👍 Makes sense to me!

adamreichold:
> If you are not fond of specialization introducing subtle behavioural difference based on return type, maybe rather than specialize on `Option<T>` we could have our own enum to specialize on, with handy traits to convert options and options-of-results to it? (And deprecate the `Option<T>` pathway as you currently have done here)

Yes, I am not fond of that. But the resulting usability and flexibility is a strong argument for supporting `Option` directly in this manner. And with really one special case, I think it is mainly a question of documentation which I will need to overhaul in any case after we settled design and implementation.

pyo3-macros-backend/src/pymethod.rs
adamreichold: > I think delaying the call to convert is fine in all cases except when we use temporary PyRef and PyRefMut in the expression to extract &T and &mut T class types.

So this is with regards to extending the lifetime of temporaries? So the test test case would be a `__next__` implementation taking `&T` where `T: PyClass` as a parameter?

> I wonder whether we could abuse FnOnce::Output to inspect the type for tag inference before we actually create the value. I can try to express that further when I'm back at my computer.

I am not sure as autoref-based specialization cannot work on types alone, it needs references to values. We could probably try to create these references out of thin air but I fear we would run into safety/validity issues, i.e. insta-UB.

davidhewitt: > So the test test case would be a __next__ implementation taking &T where T: PyClass as a parameter?

I think it's needs to take `&T` and also return data borrowed from that `T` back out of the function. For example, try replacing `__next__` in `test_proto_methods.rs` line 661 with this implementation and you'll get a compile error which #3666 fixes:

```rust
    fn __next__<'py>(&'py mut self, py: Python<'py>) -> Option<&'py PyAny> {
        if !self.polled {
            self.polled = true;
            Some(self.future.as_ref(py))
        } else {
            None
        }
    }
```

src/lib.rs
```diff
@@ -343,28 +343,6 @@ pub mod class {
...
-    /// Old module which contained some implementation details of the `#[pyproto]` module.
-    ///
-    /// Prefer using the same content from `pyo3::pyclass`, e.g. `use pyo3::pyclass::IterNextOutput` instead
-    /// of `use pyo3::class::pyasync::IterNextOutput`.
-    ///
-    /// For compatibility reasons this has not yet been removed, however will be done so
-    /// once <https://github.com/rust-lang/rust/issues/30827> is resolved.
-    pub mod iter {
-        pub use crate::pyclass::{IterNextOutput, PyIterNextOutput};
-    }
```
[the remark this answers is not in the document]
adamreichold: Yeah, I guess we should. I just like deleting code...

src/pyclass/mod.rs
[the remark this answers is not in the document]
adamreichold: Ah, that makes sense. I think the mod version is less clutter because that directory contains only three files. I did not want to make a statement of which style of paths we should generally use. I also did it as a separate commit so I can easily drop it.

So maybe I just do that to avoid derailing this PR with an unrelated change?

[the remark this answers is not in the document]
adamreichold: Dropped that change.

pyo3-macros-backend/src/pymethod.rs
adamreichold: Modified the test case and with the holders from #3667 it now compiles.

src/lib.rs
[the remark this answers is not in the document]
adamreichold: Ok, they are deprecated now and the `IntoPyCallbackOutput` impl is reduced to the one case of directly going to `*mut ffi::PyObject` which means there does not need to be a specialization for this type at all. (Which makes sense, this is a dedicated type after all and we can just implement the trait directly. Only because `__(a)next__` need special handling for `Option` that is distinct from the already existing trait implementation do we need to go for specialization.)

adamreichold:
Turns out I did not fully wire the holders change through. Added this as a separate commit here and now the modified test case does indeed compile.

tests/ui/invalid_frozen_pyclass_borrow.stderr
```diff
@@ -5,17 +5,33 @@ error: cannot use `#[pyo3(set)]` on a `frozen` class
...
 error[E0271]: type mismatch resolving `<Foo as PyClass>::Frozen == False`
-  --> tests/ui/invalid_frozen_pyclass_borrow.rs:11:19
+  --> tests/ui/invalid_frozen_pyclass_borrow.rs:9:1
    |
+9  | #[pymethods]
+   | ^^^^^^^^^^^^ expected `False`, found `True`
+10 | impl Foo {
 11 |     fn mut_method(&mut self) {}
-   |                   ^ expected `False`, found `True`
+   |                   - required by a bound introduced by this call
```
adamreichold, line 14: I'd say this is a regression, but I am also not sure where I introduced this...

adamreichold: Sorry for the noise. Found as there was only one place with a span...

pyo3-macros-backend/src/method.rs
```diff
@@ -173,10 +184,15 @@ impl SelfType {
                 } else {
                     syn::Ident::new("extract_pyclass_ref", *span)
                 };
+                let holder = syn::Ident::new(&format!("holder_{}", holders.len()), *span);
+                holders.push(quote! {
```
davidhewitt, line 188: I think this fixes the UI test "regression", I assume by causing the new error to be deduplicated away based on the span:

```suggestion
                holders.push(quote_spanned! { *span =>
```

src/callback.rs
```diff
@@ -176,3 +177,169 @@ where
 {
     value.convert(py)
 }
+
+// Autoref-based specialization for handling `__next__` returning `Option`
```
davidhewitt, line 181: For what it's worth I was sort of hoping that in `src/callback.rs` we'll either have only public API or nothing at all. 

Would you be amenable to moving these types to `src/impl_/pymethods.rs`, maybe?

davidhewitt (commented):
Looking good!

src/callback.rs
adamreichold: Yes, no problem. Will do.

pyo3-macros-backend/src/method.rs
davidhewitt: Ok it sounds like you solved it 👀

adamreichold: Indeed this fixed it. Thanks!

adamreichold: I solved the one part which was definitely a regression but this also solves the extra error messages hence is definitely better.

guide/src/migration.md
```diff
@@ -74,7 +74,124 @@ Python::with_gil(|py| {
 });
 ```
 
-### `PyType::name` is now `PyType::qualname`
+### `Iter(A)NextOutput` are deprecated
+
+The `IterNextOutput` used by `__next__` and `IterANextOutput` used by `__anext__` are deprecated. These slots can now return any type convertible into Python objects direclty using the same mechanism as other functions and methods implemented using PyO3. Most importantly, this allows returning an awaitable from `__anext__` without non-sensically wrapping it into `Yield` or `Some`. Only the return types `Option<T>` and `PyResult<Option<T>>` are still handled in special manner where `Some(val)` yields `val` and `None` stops iteration.
```
davidhewitt, line 79:
```suggestion
The `__next__` and `__anext__` magic methods can now return any type convertible into Python objects directly just like all other `#[pymethods]`. The `IterNextOutput` used by `__next__` and `IterANextOutput` used by `__anext__` are subsequently deprecated. Most importantly, this change allows returning an awaitable from `__anext__` without non-sensically wrapping it into `Yield` or `Some`. Only the return types `Option<T>` and `PyResult<Option<T>>` are still handled in a special manner where `Some(val)` yields `val` and `None` stops iteration.
```

src/impl_/pymethods.rs
```diff
@@ -299,3 +302,163 @@ pub(crate) fn get_name(name: &'static str) -> PyResult<Cow<'static, CStr>> {
...
+
+pub struct AsyncIterResultOptionTag;
+
+impl AsyncIterResultOptionTag {
+    #[inline]
+    pub fn convert<Value>(
+        self,
+        py: Python<'_>,
+        value: PyResult<Option<Value>>,
+    ) -> PyResult<*mut ffi::PyObject>
```
davidhewitt, line 445: I'm trying to think if this can potentially have a generic parameter `E: Into<PyErr>` for the result (and same for the non-async case)?

davidhewitt (commented):
Looks great! I have just one suggestion on wording in the guide, and also a thought on whether we can/should generalise the error type for the result specialisations.

src/impl_/pymethods.rs
adamreichold: Will try whether this breaks inference...

adamreichold: Appears to not overwhelm type inference, so added as a separate commit.

adamreichold:
(I guess at some point in the future, we'd want to do something similar for function arguments to avoid matching on the "Option" name?)

davidhewitt:
Quite possibly, yes. I played around with something to that regard in the past, but I think the patch is now lost/dead. Also I believe that to completely stop matching on `"Option"` as a text string, we would need to proceed with #2935

davidhewitt (approved):
Looks fantastic, thanks! It's very pleasing to see that this makes user code read nicer :)

gi0baro:
@davidhewitt Just noticed this and I have a question in regards to the deprecation: is there a way to yield `None` (also known as _bare yielding_) without using `IterNextOutput`?

For instance, how this code could be translated to avoid `IterNextOutput` usage?

```rust
#[pyclass]
pub(crate) struct PyIterAwaitable {
    result: Option<PyResult<PyObject>>,
}

impl PyIterAwaitable {
    pub(crate) fn new() -> Self {
        Self { result: None }
    }

    pub(crate) fn set_result(&mut self, result: PyResult<PyObject>) {
        self.result = Some(result);
    }
}

#[pymethods]
impl PyIterAwaitable {
    fn __await__(pyself: PyRef<'_, Self>) -> PyRef<'_, Self> {
        pyself
    }

    fn __iter__(pyself: PyRef<'_, Self>) -> PyRef<'_, Self> {
        pyself
    }

    fn __next__(&self, py: Python) -> PyResult<IterNextOutput<PyObject, PyObject>> {
        match &self.result {
            Some(res) => match res {
                Ok(v) => Ok(IterNextOutput::Return(v.clone_ref(py))),
                Err(err) => Err(err.clone_ref(py)),
            },
            _ => Ok(IterNextOutput::Yield(py.None())),
        }
    }
}
```

davidhewitt:
@gi0baro great question!

To update, you'd want to directly return `py.None()` from `__next__`, and when you're ready to `Return`, raise a `StopIteration` error like how any other exception is raised by PyO3:

```rust
    fn __next__(&self, py: Python) -> PyResult<PyObject> {
        match &self.result {
            Some(res) => match res {
                Ok(v) => Ok(PyStopIteration::new_err(v.clone_ref(py)).into_py(py)),
                Err(err) => Err(err.clone_ref(py)),
            },
            _ => Ok(py.None().into_py(py)),
        }
    }
```

PR #3662, opened 2023-12-18
Chrono: makes test independent of datetime C-API

Tpt:
This is a preliminary work to enable abi3 support of chrono conversions in order to make abi3 and non-abi3 implementations use the same tests


adamreichold (approved):
Nicely systematic approach to the issue. Thanks!

PR #3663, opened 2023-12-18
Chrono: makes FromPyObject impl on DateTime generic

Tpt:
ToPyObject was already generic

### DIFF
--- a/src/conversions/chrono.rs
+++ b/src/conversions/chrono.rs
@@ -44,7 +44,7 @@ use crate::exceptions::{PyTypeError, PyUserWarning, PyValueError};
 use crate::types::datetime::timezone_from_offset;
 use crate::types::{
     timezone_utc, PyDate, PyDateAccess, PyDateTime, PyDelta, PyDeltaAccess, PyTime, PyTimeAccess,
-    PyTzInfo, PyTzInfoAccess, PyUnicode,
+    PyTzInfo, PyTzInfoAccess,
 };
 use crate::{FromPyObject, IntoPy, PyAny, PyErr, PyObject, PyResult, Python, ToPyObject};
 use chrono::offset::{FixedOffset, Utc};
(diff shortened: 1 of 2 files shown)


src/conversions/chrono.rs
```diff
@@ -200,34 +200,24 @@ impl<Tz: TimeZone> IntoPy<PyObject> for DateTime<Tz> {
...
-            return Err(PyTypeError::new_err(
-                "expected a datetime with non-None tzinfo",
-            ));
-        };
-        let dt = NaiveDateTime::new(py_date_to_naive_date(dt)?, py_time_to_naive_time(dt)?);
-        Ok(dt.and_utc())
+        dt.and_local_timezone(tz).single().ok_or_else(|| {
+            PyValueError::new_err(format!(
+                "The datetime {} contains an incompatible or ambiguous timezone",
+                ob.repr()
```
adamreichold, line 217: You could use the built-in `Debug/repr` or `Display/str` implementations for `PyAny` which already handle fallbacks.

adamreichold: They are defined in

https://github.com/PyO3/pyo3/blob/ff50285d1f47327d8cd549730e24dd62944caf6c/src/types/mod.rs#L94-117

if you want to check whether they work for you.

Tpt: Thank you! Done.

Tpt: I allowed myself to update an other `.repr()` call in the file

adamreichold (approved):
LGTM. Will not merge immediately so others can have a look as well.

davidhewitt (approved):
Thanks @Tpt, I wanted to do similar in the past too!

By the way there's a related issue here #3266 which you might find interesting / relevant while you look into all of this. 😅

PR #3664, opened 2023-12-19
Chrono: compatibility with abi3

Tpt:


src/conversions/chrono.rs
```diff
@@ -41,141 +39,217 @@
...
+                .expect("failed to construct datetime.timedelta")
+                .into()
+        }
     }
 }
 
 impl IntoPy<PyObject> for Duration {
     fn into_py(self, py: Python<'_>) -> PyObject {
-        ToPyObject::to_object(&self, py)
+        self.to_object(py)
```
adamreichold, line 104: Why this change?

Tpt: To use a consistent style in the file.  Similary I allowed myself to put all the conversion code in `ToPyObject`  implementation and make `IntoPy<PyObject>` wrappers on top of it. If you want I can revert these changes.

src/conversions/chrono.rs
```diff
@@ -463,7 +658,7 @@ mod tests {
             );
             assert_eq!(
                 none.extract::<Utc>().unwrap_err().to_string(),
-                "TypeError: 'NoneType' object cannot be converted to 'PyTzInfo'"
+                "ValueError: expected datetime.timezone.utc"
```
adamreichold, line 661: Is this error message stable against using the limited and unlimited API? (We run our unit tests against both.)

Tpt: Yes! I simplified slightly the unlimited API implementation to not downcast to `tzinfo` before checking equality with `timezone.utc`

src/conversions/chrono.rs
adamreichold: No need. I am just making sure I did not miss some semantic change here.

src/conversions/chrono.rs
```diff
@@ -39,143 +38,220 @@
...
+        check_type(dt, datetime_class(dt.py()), "PyDateTime")?;
+
         // If the user tries to convert a timezone aware datetime into a naive one,
         // we return a hard error. We could silently remove tzinfo, or assume local timezone
         // and do a conversion, but better leave this decision to the user of the library.
+        #[cfg(not(Py_LIMITED_API))]
         if dt.get_tzinfo().is_some() {
             return Err(PyTypeError::new_err("expected a datetime without tzinfo"));
         }
+        #[cfg(Py_LIMITED_API)]
```
adamreichold, line 251: Maybe factor this out into a binding, e.g. `let has_tzinfo = ...;` to avoid duplicating the error message construction.

src/conversions/chrono.rs
```diff
@@ -223,17 +306,32 @@ impl<Tz: TimeZone + for<'a> FromPyObject<'a>> FromPyObject<'_> for DateTime<Tz>
...
+            let td = PyDelta::new(py, 0, seconds_offset, 0, true)
+                .expect("failed to construct timedelta");
+            timezone_from_offset(py, td)
+                .expect("Failed to construct PyTimezone")
+                .into()
+        }
+        #[cfg(Py_LIMITED_API)]
+        {
+            let td = Duration::seconds(seconds_offset.into()).into_py(py);
+            static TIMEZONE: GILOnceCell<Py<PyType>> = GILOnceCell::new();
```
adamreichold, line 321: I think I would prefer to handle this uniformly using a helper function as for the other type objects.

I also wonder whether it would make sense to load all of these types from `datetime` in a single pass, i.e. put a struct containing multiple `Py<PyType>` fields into a single `GILOnceCell` to reduce the overhead and collect these into a single place?

src/conversions/chrono.rs
```diff
@@ -243,21 +341,30 @@ impl FromPyObject<'_> for FixedOffset {
     /// Note that the conversion will result in precision lost in microseconds as chrono offset
     /// does not supports microseconds.
     fn extract(ob: &PyAny) -> PyResult<FixedOffset> {
-        let py_tzinfo: &PyTzInfo = ob.downcast()?;
-        // Passing `ob.py().None()` (so Python's None) to the `utcoffset` function will only
+        #[cfg(not(Py_LIMITED_API))]
+        let ob: &PyTzInfo = ob.extract()?;
+        #[cfg(Py_LIMITED_API)]
+        {
+            static TZINFO: GILOnceCell<Py<PyType>> = GILOnceCell::new();
```
adamreichold, line 348: Similar to the above, we should have a helper function for the type object.

src/conversions/chrono.rs
```diff
@@ -243,21 +341,30 @@ impl FromPyObject<'_> for FixedOffset {
...
-        let py_timedelta = py_tzinfo.call_method1("utcoffset", (ob.py().None(),))?;
-        let py_timedelta: &PyDelta = py_timedelta.downcast().map_err(|_| {
-            PyTypeError::new_err(format!("{:?} is not a fixed offset timezone", py_tzinfo))
-        })?;
-        let days = py_timedelta.get_days() as i64;
-        let seconds = py_timedelta.get_seconds() as i64;
-        // Here we won't get microseconds as noted before
-        // let microseconds = py_timedelta.get_microseconds() as i64;
-        let total_seconds = Duration::days(days) + Duration::seconds(seconds);
+        let py_timedelta = ob.call_method1("utcoffset", ((),))?;
```
adamreichold, line 360: If we are touch this, wouldn't passing `()` instead of `((),)` work as well or does `utcoffset` really check for `None` instead of a missing parameter?

Tpt: Sadly not. `utcoffset` indeed checks for `None`

src/conversions/chrono.rs
Tpt: Done.

Tpt:
@adamreichold Thank you for your review!

adamreichold (approved):

davidhewitt (approved):
This looks great, thanks @Tpt and @adamreichold!

PR #3666, opened 2023-12-19
use tag builder to avoid materializing result

davidhewitt:
@adamreichold possible solution for https://github.com/PyO3/pyo3/pull/3661/files#r1431111113

I'm unsure what I think of the horrors here, but it solves the problem 😂

### DIFF
--- a/pyo3-macros-backend/src/pymethod.rs
+++ b/pyo3-macros-backend/src/pymethod.rs
@@ -994,17 +994,35 @@ enum ReturnMode {
 }
 
 impl ReturnMode {
-    fn return_call_output(&self, call: TokenStream) -> TokenStream {
+    fn return_call_output(
+        &self,
+        call: TokenStream,
+        cls: &syn::Type,
+        spec: &FnSpec<'_>,
+    ) -> TokenStream {
         match self {
             ReturnMode::Conversion(conversion) => quote! {
                 let _result: _pyo3::PyResult<#conversion> = _pyo3::callback::convert(py, #call);
                 _pyo3::callback::convert(py, _result)
             },
-            ReturnMode::SpecializedConversion(traits, tag) => quote! {
-                let _result = #call;
-                use _pyo3::callback::{#traits};
-                (&_result).#tag().convert(py, _result)
-            },
+            ReturnMode::SpecializedConversion(traits, tag) => {
+                let unreachable_args = std::iter::repeat(quote! { ::core::panic!() }).take(
+                    spec.signature.arguments.len()
+                        + if spec.tp.skip_first_rust_argument_in_python_signature() {
+                            1
+                        } else {
+                            0
+                        },
+                );
+                let fn_ident = spec.name;
+                quote! {
+                    use _pyo3::callback::{#traits};
+                    #[allow(unreachable_code)]
+                    let _type_hinter = || #cls::#fn_ident(#(#unreachable_args),*);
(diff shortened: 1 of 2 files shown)


adamreichold:
> possible solution for https://github.com/PyO3/pyo3/pull/3661/files#r1431111113

To be honest, I am not sure what the specific problem there is, hence I am not yet confident to decided whether this additional complexity is worth it or not.

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -994,17 +994,35 @@ enum ReturnMode {
...
+                            0
+                        },
+                );
+                let fn_ident = spec.name;
+                quote! {
+                    use _pyo3::callback::{#traits};
+                    #[allow(unreachable_code)]
+                    let _type_hinter = || #cls::#fn_ident(#(#unreachable_args),*);
+                    let tag = (&_pyo3::callback::TagBuilder::new(&_type_hinter)).#tag();
+                    tag.convert(py, #call)
```
adamreichold, line 1023: Wouldn't it be simpler if we tried to wrap `call` directly, e.g.

```rust
let call = move || #call;
let tag = (&_pyo3::callback::TagBuilder::new(&call)).#tag();
tag.convert(py, call());
```

or does that break the lifetime extension as well?

davidhewitt: This is actually exactly what I tried first! 😂 

The problem is that `#call` might contain diverging branches to return errors (e.g. by `?`), and the return type of the function call is generally incompatible with this (e.g. because we `OkWrap` non-results).

Given that we know the full function signature, another option is to just carry the return type through the macro as tokens rather than rely on inference. The only complication here is that any lifetimes in the return type need to be stripped (we used to have code to do that; it wasn't much fun and quite complex, though this hack with `_type_hinter` is on a similar level of evil).

adamreichold: This complexity is why I would like to take one step back to actually look at a breaking test case. I mean, `__next__` and `__anext__` will not take any arguments besides `self` and the GIL token, right? At least the existing code from our repo seems to compile without the single-expression approach to the conversion.

davidhewitt: Absolutely, sorry, I should have started there - see https://github.com/PyO3/pyo3/pull/3661#discussion_r1431514517, which hopefully you also find gets broken without a single expression.

adamreichold: Thinking back when we introduced the requirement for single expression form, I would even say that using the alternative from back then (the holder bindings) might actually be preferable as it would avoid complicating the already complicated specialization and would more generally allow us to inspect the return value using arbitrary code, not just specialization.

davidhewitt: Yes, I think you might be right. At the time I was hopeful that was a good simplification, but it's caused us pain a few times now.

adamreichold: Ok, I will try to resurrect that PR if I can find it and try to get that into mergeable form before continuing in #3661.

adamreichold: Seems it was #3140...

davidhewitt:
I think that this isn't needed after #3667, so will close.

PR #3667, opened 2023-12-19
Bring back holder bindings as prerequiste to specialization on return value types

adamreichold:
Prerequisite for #3661


davidhewitt (approved):
LGTM if CI is happy 🚀

Issue #3668, opened 2023-12-19
`GILPool` is a broken abstraction in presence of `gevent`


davidhewitt:
From investigation in https://github.com/pydantic/pydantic/issues/7211

I am able to produce a small minimal case which shows that in the presence of `gevent` task switching, the `GILPool` is not sound. 

The fundamentals of the interaction can be summarised as follows:

1.  `gevent` task A calls into a PyO3 function, creating a `GILPool` we will call pool `A`.
2. The PyO3 function does some work, populating pool `A`, before running a pure-python callback which allows `gevent` to switch task
3. `gevent` task B calls into the PyO3 function, creating another new pool `B`. _Due to the current `GILPool` scoping design, this pool is necessarily freed when pool `A` is freed._
4. The PyO3 function again does work, filling `GilPool` B, and then eventually also switches task.
5. Task `A` resumes and finishes its PyO3 function call, freeing pool `A` (which also frees the contents of pool `B`).
6. Switch back to task `B`, which continues PyO3 function execution, reading from the freed pool `B` and thus leading to UB.

----

Some quick thoughts:

- This is pretty painful; I need to have a good think whether there's any mitigation which we can install in the `GILPool` implementation to avoid this.
- The good news is that one concrete mitigation is to continue the new `Py2` API work and rip out the `GILPool` completely.

It's unclear to me whether the root cause is that `GILPool` relies on an invalid assumption regarding nature of Rust scopes in the presence of FFI calls, or that `gevent` is not sound by assuming that it's always safe to switch task. Given the long existence of `gevent`, probably this lesson is for us.

adamreichold:
> This is pretty painful; I need to have a good think whether there's any mitigation which we can install in the GILPool implementation to avoid this.

As a short-term mitigation, I think we could switch the pool to design to not store the objects in TLS, but directly in the pool itself and only put the currently active pool into TLS. This should avoid pool A freeing objects allocated for pool B in the above.

davidhewitt:
Possibly, but I think with just that change the gevent task switching will still confuse the active pool in the TLS, and instead when task A resumes it will start writing new objects into task B's pool. I think that must still necessarily be unsound if task B completes before task A.

davidhewitt:
One option is to just `panic` if `gevent` is present in `sys.modules`, but that's pretty disruptive.

adamreichold:
Ah, of course, gevent switching tasks does not switch TLS which is basically again what this is about...

[...]

PR #3669, opened 2023-12-19
add test which is broken with gevent

davidhewitt:
Minimal reproduction of the case described in #3668 

Here is a pretty typical output from the test:

```
start 0 0
gevent sleep 0 0
start 1 0
gevent sleep 1 0
after gevent sleep 0 0
dropping: ['key', ArbitraryClass(0, 0), 'ArbitraryClass(0, 0)', 'key', ArbitraryClass(1, 0), 'ArbitraryClass(1, 0)', None, 'ArbitraryClass(0, 0)']
del 0 0
del 1 0
end 0 0
start 0 1
gevent sleep 0 1
after gevent sleep 1 0
Fatal Python error: Segmentation fault
```

... i.e. after task 0 completes, task 1's pool of references is also freed. The crash when task 1 resumes is plain and simple UB; I have seen both segfaults and assertion failures.

### DIFF
--- a/pytests/src/misc.rs
+++ b/pytests/src/misc.rs
@@ -18,7 +18,11 @@ fn accepts_bool(val: bool) -> bool {
 }
 
 #[pyfunction]
-fn get_item_and_run_callback(dict: &PyDict, callback: &PyAny) -> PyResult<()> {
+fn get_item_and_run_callback(dict: Bound<'_, PyDict>, callback: Bound<'_, PyAny>) -> PyResult<()> {
+    // This function gives the opportunity to run a pure-Python callback so that
+    // gevent can instigate a context switch. This had problematic interactions
+    // with PyO3's removed "GIL Pool".
+    // For context, see https://github.com/PyO3/pyo3/issues/3668
     let item = dict.get_item("key")?.expect("key not found in dict");
     let string = item.to_string();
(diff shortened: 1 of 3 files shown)


pytests/src/misc.rs
```diff
@@ -1,13 +1,23 @@
...
+use pyo3::{prelude::*, types::PyDict};
 
 #[pyfunction]
 fn issue_219() {
     // issue 219: acquiring GIL inside #[pyfunction] deadlocks.
     Python::with_gil(|_| {});
 }
 
+#[pyfunction]
+fn get_item_and_run_callback(dict: &PyDict, callback: &PyAny) -> PyResult<()> {
```
adamreichold, line 10: I wonder what exactly we need to provide so that this could pass. The argument extraction code for supporting `Bound` argument and this could pass?

davidhewitt: Good question, it might already be possible to make this pass, I'll have a try...

davidhewitt: Indeed it turns out that just taking the `dict` and `callback` as `Bound<'_, PyDict>` and `Bound<'_, PyAny>` respectively fixes the crash (by changing the whole function body to use the new API).

adamreichold (approved):

PR #3670, opened 2023-12-19
Adds std::duration::Duration from/to Python conversions  

Tpt:


Tpt:
We already support `chrono::Duration`, it might be nice to support `std::duration::Duration` for consistency.

Things to notice:
- `std::duration::Duration` only supports positive durations, an exception is raised when converting a negative `timedelta`
- a warning is created if precision is lost when converting from `Duration` to `timedelta`, opposite to `chrono::Duration` (is it a good idea?)

src/tests/common.rs
```diff
@@ -129,9 +129,9 @@ mod inner {
     #[macro_export]
     macro_rules! assert_warnings {
         ($py:expr, $body:expr, [$(($category:ty, $message:literal)),+] $(,)? ) => {{
-            CatchWarnings::enter($py, |w| {
+            $crate::tests::common::CatchWarnings::enter($py, |w| {
                 $body;
-                let expected_warnings = [$((<$category>::type_object($py), $message)),+];
+                let expected_warnings = [$((<$category as $crate::type_object::PyTypeInfo>::type_object($py), $message)),+];
```
Tpt, line 134: I allowed myself to make this macro more portable

adamreichold (approved):
LGTM

davidhewitt (approved):
👍 likewise, thanks!

It's a little sad that Rust stdlib duration is strictly positive, but that's not really our design problem to worry about here.

davidhewitt:
Hmm, there are some test failures?

> * a warning is created if precision is lost when converting from `Duration` to `timedelta`, opposite to `chrono::Duration`. Is it a good idea?

Yeah I think that's fine to warn 👍

adamreichold:
@Tpt Argh, unused import on `wasm32-wasi` derailed the CI...

adamreichold:
Pushed a fixup.

PR #3671, opened 2023-12-19
Add `Py2` variants of `PyDowncastError`

davidhewitt:
Part of #3382 

To keep `.downcast()` APIs performant in the error case, and also off the pool, this PR introduces two types used by `PyAnyMethods`.

`PyDowncastError2` (name to change later when we make public) is used for `.downcast()` and `.downcast_exact()`.

`PyDowncastIntoError` is used for `.downcast_into()` and `.downcast_into_exact()`.

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -1,5 +1,7 @@
+use crate::instance::Py2;
 use crate::panic::PanicException;
 use crate::type_object::PyTypeInfo;
+use crate::types::any::PyAnyMethods;
 use crate::types::{PyTraceback, PyType};
 use crate::{
     exceptions::{self, PyBaseException},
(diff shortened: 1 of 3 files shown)


src/err/mod.rs
```diff
@@ -61,6 +63,42 @@ impl<'a> PyDowncastError<'a> {
     }
 }
 
+/// Error that indicates a failure to convert a PyAny to a more specific Python type.
+#[derive(Debug)]
+pub(crate) struct PyDowncastError2<'a, 'py> {
+    from: &'a Py2<'py, PyAny>,
```
adamreichold, line 69: Two vague questions:
* Is it reasonable to make the three downcast errors type aliases for a common base type with generic parameter for the `from` part? Is there a reasonable trait bound for that? Does it really make sense if at least one of the version is going away soon?
* Should this particular version use `Py2Borrowed<'a, 'py, PyAny>` instead of `&'a Py2<'py, Any>` to remove one indirection?

davidhewitt: I'm 50/50 of the idea of a generic parameter for these and a common implementation, so if you'd prefer that, I can switch. Otherwise, there's not _that_ much duplication so I'm tempted to leave as-is. I don't think there's a suitable trait bound right now but it would probably be `AsRef<Py2<PyAny>>` if we implemented that for all the types?

Regarding `Py2Borrowed` - actually if we are happy for users to add one lifetime parameter to `PyDowncastError` then we can merge `PyDowncastError` and `PyDowncastError2` already, and store `Py2Borrowed`?

The conversion `&Py2<PyAny> -> Py2Borrowed<PyAny>` does involve one pointer traversal, so I think if we care about every nanosecond the `PyDowncastError2` as in this PR maybe the tiniest fraction faster, but maybe it's better for simplicity to just go with `Py2Borrowed` for now? Once the gil refs API is gone we could then reassess if we wanted to micro-optimize.

adamreichold: Nothing really sticks out as definite improvements, especially since the main smell here is a bit of duplicate boiler plate. Maybe just leave it as it is and use a helper function to construct the error message because that seems to be the most involved part?

davidhewitt: Done 👍

adamreichold (approved):

PR #3675, opened 2023-12-20
implement `PyDictMethods`

davidhewitt:
Split from #3606 

Fairly mechanical. Most interesting things to note here:
- I was able to implement the existing gil-ref `PyDictIterator` as a thin wrapper around the new `PyDictIter` 
- I added `PyBorrowed::from_ptr_unchecked` to skip the null check completely in cases where we can make the assumption that a null would be a Python interpreter bug

### DIFF
--- a/src/types/dict.rs
+++ b/src/types/dict.rs
@@ -1,8 +1,12 @@
 use super::PyMapping;
 use crate::err::{self, PyErr, PyResult};
 use crate::ffi::Py_ssize_t;
+use crate::ffi_ptr_ext::FfiPtrExt;
+use crate::instance::Py2;
+use crate::py_result_ext::PyResultExt;
+use crate::types::any::PyAnyMethods;
 use crate::types::{PyAny, PyList};
-use crate::{ffi, PyObject, Python, ToPyObject};
+use crate::{ffi, Python, ToPyObject};
 
 /// Represents a Python `dict`.
 #[repr(transparent)]
@@ -54,7 +58,7 @@ pyobject_native_type_core!(
 impl PyDict {
     /// Creates a new empty dictionary.
     pub fn new(py: Python<'_>) -> &PyDict {
-        unsafe { py.from_owned_ptr::<PyDict>(ffi::PyDict_New()) }
+        unsafe { py.from_owned_ptr(ffi::PyDict_New()) }
     }
 
     /// Creates a new dictionary from the sequence given.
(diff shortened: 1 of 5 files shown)


src/types/dict.rs
```diff
@@ -308,27 +255,309 @@ impl PyDict {
...
+
+    fn into_iter(self) -> Self::IntoIter {
+        PyDictIterator(PyDictIter::new(Py2::borrowed_from_gil_ref(&self).clone()))
+    }
+}
+
 /// PyO3 implementation of an iterator for a Python `dict` object.
-pub struct PyDictIterator<'py> {
-    dict: &'py PyDict,
+pub(crate) struct PyDictIter<'py> {
```
adamreichold, line 548: For consistency, shouldn't this be `Py2DictIterator` or `PyDictIterator2`? Replacing `Iterator` by `Iterator` does not really explain the difference...

src/types/dict.rs
```diff
@@ -308,27 +255,309 @@ impl PyDict {
...
+        self.0.len()
+    }
+}
+
+impl<'a> std::iter::IntoIterator for &'a PyDict {
+    type Item = (&'a PyAny, &'a PyAny);
+    type IntoIter = PyDictIterator<'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        PyDictIterator(PyDictIter::new(Py2::borrowed_from_gil_ref(&self).clone()))
```
adamreichold, line 543: Couldn't this call the `iter` method directly?

src/types/dict.rs
```diff
@@ -308,27 +255,309 @@ impl PyDict {
...
+    }
+}
+
+impl<'py> ExactSizeIterator for PyDictIterator<'py> {
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+}
+
+impl<'a> std::iter::IntoIterator for &'a PyDict {
```
adamreichold, line 538: I think `IntoIterator` is part of std's prelude?

src/types/dict.rs
davidhewitt: Yes, and I think also for `&Py2`'s `into_iter()`.

src/types/dict.rs
davidhewitt: Yep, nice!

adamreichold (approved):

Issue #2275, opened 2022-04-05
GIL-free references to Python data


davidhewitt:
I think there's a few other immutable types like `str` and `tuple` which I guess can also have treatment like this where `py` is necessary for reading, but once read, the GIL can be released. Might be interesting to follow up this another time.

_Originally posted by @davidhewitt in https://github.com/PyO3/pyo3/pull/2235#pullrequestreview-915041031_

PR #3677, opened 2023-12-20
implement `PyStringMethods`

davidhewitt:
For #3382 

There are some interesting bits going on in this PR:
- `to_str` needs the GILPool to work for abi3 before Python 3.10, so I have added a `#[cfg]` on the corresponding `PyStringMethods` API.
- I've introduced `to_cow` as a new API in `PyStringMethods` which is an alternative for users on abi3 before 3.10 (cc @alex)
- I've added `Py::to_str`, `Py::to_cow` and `Py::to_string_lossy` as reads which can outlast the GIL lifetime, like we did for bytes (cf #2275)

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -1,7 +1,11 @@
 #[cfg(not(Py_LIMITED_API))]
 use crate::exceptions::PyUnicodeDecodeError;
+use crate::ffi_ptr_ext::FfiPtrExt;
+use crate::instance::Py2Borrowed;
+use crate::types::any::PyAnyMethods;
+use crate::types::bytes::PyBytesMethods;
 use crate::types::PyBytes;
-use crate::{ffi, PyAny, PyResult, Python};
+use crate::{ffi, IntoPy, Py, Py2, PyAny, PyResult, Python};
 use std::borrow::Cow;
 use std::os::raw::c_char;
 use std::str;
@@ -165,7 +169,7 @@ impl PyString {
     pub fn from_object<'p>(src: &'p PyAny, encoding: &str, errors: &str) -> PyResult<&'p PyString> {
         unsafe {
             src.py()
-                .from_owned_ptr_or_err::<PyString>(ffi::PyUnicode_FromEncodedObject(
+                .from_owned_ptr_or_err(ffi::PyUnicode_FromEncodedObject(
                     src.as_ptr(),
                     encoding.as_ptr() as *const c_char,
                     errors.as_ptr() as *const c_char,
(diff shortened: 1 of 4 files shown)


alex:
Makes sense to me.

src/types/string.rs
```diff
@@ -165,61 +169,40 @@ impl PyString {
...
+            }?;
+            Ok(unsafe { std::str::from_utf8_unchecked(bytes.as_bytes()) })
+        }
     }
 
     /// Converts the `PyString` into a Rust string.
     ///
     /// Unpaired surrogates invalid UTF-8 sequences are
     /// replaced with `U+FFFD REPLACEMENT CHARACTER`.
     pub fn to_string_lossy(&self) -> Cow<'_, str> {
```
adamreichold, line 204: Maybe it would help migrating downstream code if `PyString` also had the `to_cow` method, so that using that method and changing the pointer type could be separate steps.

src/types/string.rs
```diff
@@ -238,6 +221,138 @@ impl PyString {
...
+            return Ok(Cow::Borrowed(s));
+        }
+
+        let bytes = unsafe {
+            ffi::PyUnicode_AsUTF8String(ptr)
+                .assume_owned_or_err(py)?
+                .downcast_into_unchecked::<PyBytes>()
+        };
+        Ok(Cow::Owned(
+            unsafe { str::from_utf8_unchecked(bytes.as_bytes()) }.into(),
```
adamreichold, line 328: I think `.to_owned()` would be nicer here to make the allocation more obvious.

src/types/string.rs
```diff
@@ -238,6 +221,138 @@ impl PyString {
...
+    #[allow(clippy::wrong_self_convention)]
+    fn to_cow(self) -> PyResult<Cow<'a, str>> {
+        // TODO: this method can probably be deprecated once Python 3.9 support is dropped,
+        // because all versions then support the more efficient `to_str`.
+
+        let ptr = self.as_ptr();
+        let py = self.py();
+
+        #[cfg(any(Py_3_10, not(Py_LIMITED_API)))]
+        if let Ok(s) = self.to_str() {
```
adamreichold, line 318: Couldn't this error be forwarded (in contrast to `to_string_lossy`) because `PyUnicode_AsUTF8String` cannot succeed if the UTF-8 check already failed?

davidhewitt: Yes, I think you are right, and that would also improve coverage.

src/types/string.rs
davidhewitt: Good idea, added.

adamreichold (approved):

davidhewitt:
We've had a little bit of flakiness with PyPy 3.8 previously so will retry in the hope it's not caused by this PR.

davidhewitt:
Hmm I'll try to investigate this later 🤔

davidhewitt:
I noticed that PyPy 3.8 succeeded in the second run (but failed in the first), so I think this is just the usual PyPy flakiness :(

PR #3679, opened 2023-12-20
implement datetime traits for `Bound`

davidhewitt:
Split from #3606 

This works a bit differently from the other `PyXMethods` traits, because we already have traits for the datetime types, so we can just implement them for `Bound` types.

The only slight snag is that `PyTzInfoAccess` returns a gil-ref, so ~~for now I've added `PyTzInfoAccess2` (and we can clean that up when we make the new API public)~~ I've had to modify that to return a `Bound` tzinfo instead.

### DIFF
--- a/src/types/datetime.rs
+++ b/src/types/datetime.rs
@@ -19,9 +19,11 @@ use crate::ffi::{
     PyDateTime_TIME_GET_FOLD, PyDateTime_TIME_GET_HOUR, PyDateTime_TIME_GET_MICROSECOND,
     PyDateTime_TIME_GET_MINUTE, PyDateTime_TIME_GET_SECOND,
 };
+use crate::ffi_ptr_ext::FfiPtrExt;
 use crate::instance::PyNativeType;
+use crate::types::any::PyAnyMethods;
 use crate::types::PyTuple;
-use crate::{AsPyPointer, IntoPy, Py, PyAny, Python};
+use crate::{Bound, IntoPy, Py, PyAny, Python};
 use std::os::raw::c_int;
 #[cfg(feature = "chrono")]
 use std::ptr;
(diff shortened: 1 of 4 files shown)


src/types/datetime.rs
```diff
@@ -160,13 +162,13 @@ pub trait PyTimeAccess {
...
-pub trait PyTzInfoAccess {
+pub trait PyTzInfoAccess<'py> {
     /// Returns the tzinfo (which may be None).
     ///
     /// Implementations should conform to the upstream documentation:
     /// <https://docs.python.org/3/c-api/datetime.html#c.PyDateTime_DATE_GET_TZINFO>
     /// <https://docs.python.org/3/c-api/datetime.html#c.PyDateTime_TIME_GET_TZINFO>
-    fn get_tzinfo(&self) -> Option<&PyTzInfo>;
+    fn get_tzinfo(&self) -> Option<Bound<'py, PyTzInfo>>;
 }
```
davidhewitt, line 172: Following the rebase onto the now-public `Bound` API I've decided to just make (small) breaking changes to the `PyTzInfoAccess` trait, particularly in changing its return type.

davidhewitt: Naturally I forgot to update our own code for this change...

adamreichold: Maybe the docs should explain how to get from the new return type to a GIL ref? Or will this happen elsewhere and we can just explain it once in the migration guide?

davidhewitt: I would prefer not add it to this API doc, as I think we'll repeat it in several places. Instead yes I'll start working on the documentation which will explain how to migrate.

adamreichold (approved):
LGTM. The CI says the pytests still need to be adjusted though.

davidhewitt:
Thanks for fixing those up! I ended up getting a bit deep into rebasing #3606 for the new `Bound` / `Borrowed` names; that branch is looking _a lot_ tidier now!

PR #3680, opened 2023-12-20
implement `PyListMethods`

davidhewitt:
Split from #3606 

Follows the same patterns we've seen in #3675 for the gil-ref iterator to wrap the new iterator. I think this is again extremely mechanical.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -228,6 +228,17 @@ pub(crate) struct Py2Borrowed<'a, 'py, T>(
     Python<'py>,
 );
 
+impl<'py, T> Py2Borrowed<'_, 'py, T> {
+    /// Creates a new owned `Py2` from this borrowed reference by increasing the reference count.
+    pub(crate) fn to_owned(self) -> Py2<'py, T> {
+        unsafe { ffi::Py_INCREF(self.as_ptr()) };
+        Py2(
+            self.py(),
+            ManuallyDrop::new(unsafe { Py::from_non_null(self.0) }),
+        )
+    }
+}
+
 impl<'a, 'py> Py2Borrowed<'a, 'py, PyAny> {
     /// # Safety
     /// This is similar to `std::slice::from_raw_parts`, the lifetime `'a` is completely defined by
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -221,6 +221,8 @@ unsafe impl<T> AsPyPointer for Py2<'_, T> {
 ///
 /// The advantage of this over `&Py2` is that it avoids the need to have a pointer-to-pointer, as Py2
 /// is already a pointer to an `ffi::PyObject``.
+///
+/// Similarly, this type is `Copy` and `Clone`, like a shared reference (`&T`).
 #[repr(transparent)]
 pub(crate) struct Py2Borrowed<'a, 'py, T>(
     NonNull<ffi::PyObject>,
@@ -325,6 +327,14 @@ impl<'py, T> Deref for Py2Borrowed<'_, 'py, T> {
         unsafe { &*(&self.0 as *const _ as *const Py2<'py, T>) }
     }
 }
+
+impl<T> Clone for Py2Borrowed<'_, '_, T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T> Copy for Py2Borrowed<'_, '_, T> {}
 
 /// A GIL-independent reference to an object allocated on the Python heap.
 ///
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -4,6 +4,7 @@ use std::iter::FusedIterator;
 use crate::err::{self, PyResult};
 use crate::ffi::{self, Py_ssize_t};
 use crate::ffi_ptr_ext::FfiPtrExt;
+use crate::instance::Py2Borrowed;
 use crate::internal_tricks::get_ssize_index;
 use crate::types::{PySequence, PyTuple};
 use crate::{Py2, PyAny, PyObject, Python, ToPyObject};
@@ -391,7 +392,7 @@ impl<'py> PyListMethods<'py> for Py2<'py, PyList> {
             // PyList_GetItem return borrowed ptr; must make owned for safety (see #890).
             ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t)
                 .assume_borrowed_or_err(self.py())
-                .map(|borrowed_any| borrowed_any.clone())
+                .map(Py2Borrowed::to_owned)
         }
     }
 
@@ -405,7 +406,7 @@ impl<'py> PyListMethods<'py> for Py2<'py, PyList> {
         // PyList_GET_ITEM return borrowed ptr; must make owned for safety (see #890).
         ffi::PyList_GET_ITEM(self.as_ptr(), index as Py_ssize_t)
             .assume_borrowed(self.py())
-            .clone()
+            .to_owned()
     }
 
     /// Takes the slice `self[low:high]` and returns it as a new list.
--- a/src/types/list.rs
+++ b/src/types/list.rs
@@ -3,9 +3,13 @@ use std::iter::FusedIterator;
 
 use crate::err::{self, PyResult};
 use crate::ffi::{self, Py_ssize_t};
+use crate::ffi_ptr_ext::FfiPtrExt;
 use crate::internal_tricks::get_ssize_index;
 use crate::types::{PySequence, PyTuple};
-use crate::{Py, PyAny, PyObject, Python, ToPyObject};
+use crate::{Py2, PyAny, PyObject, Python, ToPyObject};
+
+use crate::types::any::PyAnyMethods;
+use crate::types::sequence::PySequenceMethods;
 
 /// Represents a Python `list`.
 #[repr(transparent)]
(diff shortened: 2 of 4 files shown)


src/types/list.rs
```diff
@@ -125,12 +386,12 @@ impl PyList {
...
-    pub fn get_item(&self, index: usize) -> PyResult<&PyAny> {
+    fn get_item(&self, index: usize) -> PyResult<Py2<'py, PyAny>> {
         unsafe {
-            let item = ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t);
             // PyList_GetItem return borrowed ptr; must make owned for safety (see #890).
-            ffi::Py_XINCREF(item);
-            self.py().from_owned_ptr_or_err(item)
+            ffi::PyList_GetItem(self.as_ptr(), index as Py_ssize_t)
+                .assume_borrowed_or_err(self.py())
+                .map(|borrowed_any| borrowed_any.clone())
```
adamreichold, line 394: I wonder if we should have a `ToOwned` impl connecting `Py2` and `Py` so that we can call `.to_owned` which better convey the intent IMHO.

adamreichold: (While `deref` then `clone` will do the same thing, it is somewhat implicit.)

src/types/list.rs
```diff
@@ -253,51 +515,99 @@ impl PyList {
...
+    fn to_tuple(&self) -> Py2<'py, PyTuple> {
+        unsafe {
+            ffi::PyList_AsTuple(self.as_ptr())
+                .assume_owned(self.py())
+                .downcast_into_unchecked()
+        }
     }
 }
 
-index_impls!(PyList, "list", PyList::len, PyList::get_slice);
```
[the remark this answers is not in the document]
davidhewitt: Ah, yes, this is one of the sadder parts about the `Py2` APIs. Because index operators expect references, we just flat-out can't have this with `Py2`. I'd forgotten to point this out, but I also don't see a way around this.

(Given that we are more or less decided that we're going to move ahead with Py2 for 0.21, I think I will start building a migration guide entry now.)

src/types/list.rs
davidhewitt: Unfortunately it can't be [`std::borrow::ToOwned`](https://doc.rust-lang.org/stable/std/borrow/trait.ToOwned.html#) because the companion requirement of `std::borrow::Borrow` requires the "borrowed" thing to be a reference.

But we could have `.to_owned()` as an associated method.

Interestingly I ran into this same idea in https://github.com/PyO3/pyo3/pull/3681#discussion_r1432899731, where I call it `clone_ref`.

src/types/list.rs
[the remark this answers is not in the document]
adamreichold: Yeah, that doesn't work. Most likely because those traits predate generic associated types. Do we have any idea how common it is to access `PyList` via `Index`?

src/types/list.rs
adamreichold: Wouldn't the impl

```rust
impl<'a, 'py, T> std::borrow::ToOwned for Py2Borrowed<'a, 'py, T> {
    type Owned = Py2<'py, T>;

    fn to_owned(&self) -> Self::Owned {
        todo!()
    }
}

impl<'a, 'py, T> std::borrow::Borrow<Py2Borrowed<'a, 'py, T>> for Py2<'py, T> {
    fn borrow(&self) -> &Py2Borrowed<'a, 'py, T> {
        todo!()
    }
}
```

work? The `Borrow` is admittedly useless, but it doesn't hurt either, does it?

davidhewitt: Hmm, interesting. That seems slightly dangerous because I think the `'a` lifetime is unconstrained so can be used to accidentally extend the inferred lifetime of the reference.

src/types/list.rs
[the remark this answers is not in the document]
davidhewitt: I suspect it's a popular convenience but I still don't see a way to let users keep it 😭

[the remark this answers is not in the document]
adamreichold: Fair enough.

src/types/list.rs
adamreichold: Isn't this just what `attach_borrow` already does?

davidhewitt: Not quite, with `attach_borrow` the `'a` is constrained by `&'a self`.

adamreichold: No, it binds `&'a self`. And if both impls get a `'py: 'a` bound?

davidhewitt: 👍 I'll try to play with this later on

davidhewitt: So here's an example of what makes me uncomfortable. The following compiles:

```rust
impl<'a, 'py, T> std::borrow::ToOwned for Py2Borrowed<'a, 'py, T>
where
    'py: 'a,
{
    type Owned = Py2<'py, T>;

    fn to_owned(&self) -> Self::Owned {
        todo!()
    }
}

impl<'a, 'py, T> std::borrow::Borrow<Py2Borrowed<'a, 'py, T>> for Py2<'py, T>
where
    'py: 'a,
{
    fn borrow(&self) -> &Py2Borrowed<'a, 'py, T> {
        todo!()
    }
}

fn dangerous<'py>(any: &Py2<'py, PyAny>) {
    let value = any.get_item("foo").unwrap();
    let borrowed: &Py2Borrowed<'py, 'py, PyAny> = value.borrow();
    // let borrowed_unsound = *borrowed;  // if Py2Borrowed implemented `Copy`
}
```

This `borrowed` on the last line supposedly claims the lifetime of the reference owned by `'value` is `'py`. At the moment this can't do any damage because we're constrained by `&Py2Borrowed` where that reference is tied to `value`. In combination with the `Clone + Copy` implementations which I was proposing in #3681 then this would be able to copy to a new `Py2Borrowed<'py, 'py, PyAny>`, which must be unsound because `value` does not live for `'py`.

My intuition tells me that `Py2Borrowed` is reference-like, so the `Copy` implementation seems plausibly sound, which is why I infer that this proposal is not sound.

adamreichold: Yes, I agree. This is a use-after-free in waiting. And the `'py: 'a` does not change that. It is "pool think" as I assumed that the Python reference would be alive for at least `'py` but that just isn't the case with `Py2<'py, T>` which is bound for `'py` but might be much more short-lived.

So, I'd say we go with the associated `to_owned` with proper bounds. But I also don't mind `clone_ref` if you think it better fits into the overall framework. Personally, I'd say `to_owned` does make the intent a bit clearer though.

davidhewitt: Given our newly found names, what do you think of `.to_bound()`?

adamreichold: To be honest, I would prefer both `clone_ref` (for the structural consistency) and `to_owned` (for signalling intent) as I think of both `Bound` and `Borrowed` as bound (both carry the GIL lifetime), just that `BoundBorrowed` is too verbose, so the change is not from "not bound" to "bound", it is from "bound and borrowed" to "bound and owned".

davidhewitt: Ah yes, good reason why `to_bound` is not ideal👍 

Let's start with `to_owned` as an associated method for now then, as the `clone_ref()` changing type of `Borrowed` -> `Bound` is just a little different to the `Py::clone_ref` which just does `Py` -> `Py`. While it's still got a nice analogy, it will make users think just as much as an associated `to_owned` might.

src/instance.rs
```diff
@@ -228,6 +228,17 @@ pub(crate) struct Py2Borrowed<'a, 'py, T>(
     Python<'py>,
 );
 
+impl<'py, T> Py2Borrowed<'_, 'py, T> {
+    /// Creates a new owned `Py2` from this borrowed reference by increasing the reference count.
+    pub(crate) fn to_owned(self) -> Py2<'py, T> {
```
davidhewitt, line 233: I took `self` by ownership here because it avoids the need for a closure in `.assume_borrowed(py).map(Py2Borrowed::to_owned)`.

Unsurprisingly clippy doesn't like this, and wants it to be called `into_owned`. What do you think of that?

adamreichold: If `ToOwned/Borrow` were to support GAT, Clippy would also complain about this impl, wouldn't it? So I'd say Clippy is wrong and the lint should generally not fire for `Copy` types. Or maybe it does and we should just also add the `Clone` and `Copy` impls for `Py2Borrowed` here as well?

davidhewitt: Ah indeed for `Copy` types, clippy even wants `to_` methods to take `self`: https://rust-lang.github.io/rust-clippy/master/index.html#/wrong_self_convention

So I've cherry-picked the `Copy` commit from #3681 here.

src/instance.rs
```diff
@@ -221,13 +221,26 @@ unsafe impl<T> AsPyPointer for Py2<'_, T> {
 ///
 /// The advantage of this over `&Py2` is that it avoids the need to have a pointer-to-pointer, as Py2
 /// is already a pointer to an `ffi::PyObject``.
+///
+/// Similarly, this type is `Copy` and `Clone`, like any pointer.
```
adamreichold, line 225: I think this should be "like shared references". Because it does not act as a raw pointer and mutable references are not `Copy`.

adamreichold (approved):
Tiny nit w.r.t. documentation, otherwise LGTM.

src/instance.rs
davidhewitt: Yep agreed 👍

davidhewitt:
Thanks for all these reviews 🙏 

I think my next step today is to write a high-level summary of the tasks remaining before I think we can release this API, so that we all have a sense of progress (and can coordinate picking up items in parallel if desired).

PR #3681, opened 2023-12-20
implement `PyTupleMethods`

davidhewitt:
Split off from #3606 

There are a few interesting pieces going on here, because it is safe to borrow items from a `tuple` due to its immutability:
- Iterating tuples with the gil-ref API yields borrowed objects. This however isn't possible when calling `.into_iter()` on a `Py2<PyTuple>` because the reference is then owned by the iterator, and we don't have lending iterators in Rust yet.
  - For this reason, I introduced `PyTupleIterator2` as yielding owned objects.
  - Users can opt-in to borrowed iteration with an `.iter_borrowed()` method, which creates an iterator `PyTupleIterBorrowed` which yields `Py2Borrowed` items. The gil-ref iterator wraps this borrowing iterator.
- Similarly the gil-ref API `get_item` borrows objects. To keep types predictable on the migration from `&PyAny` -> `Py2<'py, PyAny>` the new `get_item` returns owned objects, and I have added `get_borrowed_item` which returns borrowed objects and is what the gil-ref API wraps.

This way the more efficient "borrowed" methods are still available, but users have to opt-into them. I'd argue this creates a simpler migration pathway while also leaving the more efficient operations available for us to use.

Finally, it turns out that to implement this it starts getting useful if `Py2Borrowed` implements `Copy` (which seems reasonable as it's basically a pointer), so I've pushed that as a precursor commit.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -325,10 +325,10 @@ impl<'py, T> Borrowed<'py, 'py, T>
 where
     T: HasPyGilRef,
 {
-    // pub(crate) fn into_gil_ref(self) -> &'py T::AsRefTarget {
-    //     // Safety: self is a borrow over `'py`.
-    //     unsafe { self.py().from_borrowed_ptr(self.0.as_ptr()) }
-    // }
+    pub(crate) fn into_gil_ref(self) -> &'py T::AsRefTarget {
+        // Safety: self is a borrow over `'py`.
+        unsafe { self.py().from_borrowed_ptr(self.0.as_ptr()) }
+    }
 }
 
 impl<T> std::fmt::Debug for Borrowed<'_, '_, T> {
--- a/src/types/tuple.rs
+++ b/src/types/tuple.rs
@@ -60,9 +60,12 @@ pyobject_native_type_core!(PyTuple, pyobject_native_static_type_object!(ffi::PyT
 impl PyTuple {
     /// Deprecated form of `PyTuple::new_bound`.
     #[track_caller]
-    #[deprecated(
-        since = "0.21.0",
-        note = "`PyTuple::new` will be replaced by `PyTuple::new_bound` in a future PyO3 version"
+    #[cfg_attr(
+        not(feature = "gil-refs"),
+        deprecated(
+            since = "0.21.0",
+            note = "`PyTuple::new` will be replaced by `PyTuple::new_bound` in a future PyO3 version"
+        )
     )]
     pub fn new<T, U>(
         py: Python<'_>,
@@ -115,9 +118,12 @@ impl PyTuple {
     }
 
     /// Deprecated form of `PyTuple::empty_bound`.
-    #[deprecated(
-        since = "0.21.0",
-        note = "`PyTuple::empty` will be replaced by `PyTuple::empty_bound` in a future PyO3 version"
+    #[cfg_attr(
+        not(feature = "gil-refs"),
+        deprecated(
+            since = "0.21.0",
+            note = "`PyTuple::empty` will be replaced by `PyTuple::empty_bound` in a future PyO3 version"
+        )
     )]
     pub fn empty(py: Python<'_>) -> &PyTuple {
         Self::empty_bound(py).into_gil_ref()
@@ -361,7 +367,7 @@ impl<'py> PyTupleMethods<'py> for Bound<'py, PyTuple> {
     }
 
     fn get_borrowed_item<'a>(&'a self, index: usize) -> PyResult<Borrowed<'a, 'py, PyAny>> {
-        Borrowed::from(self).get_borrowed_item(index)
+        self.as_borrowed().get_borrowed_item(index)
     }
 
     #[cfg(not(any(Py_LIMITED_API, PyPy)))]
@@ -371,7 +377,7 @@ impl<'py> PyTupleMethods<'py> for Bound<'py, PyTuple> {
 
     #[cfg(not(any(Py_LIMITED_API, PyPy)))]
     unsafe fn get_borrowed_item_unchecked<'a>(&'a self, index: usize) -> Borrowed<'a, 'py, PyAny> {
-        Borrowed::from(self).get_borrowed_item_unchecked(index)
+        self.as_borrowed().get_borrowed_item_unchecked(index)
     }
 
     #[cfg(not(Py_LIMITED_API))]
@@ -406,7 +412,7 @@ impl<'py> PyTupleMethods<'py> for Bound<'py, PyTuple> {
     }
 
     fn iter_borrowed<'a>(&'a self) -> BorrowedTupleIterator<'a, 'py> {
-        BorrowedTupleIterator::new(Borrowed::from(self))
+        BorrowedTupleIterator::new(self.as_borrowed())
     }
 
     fn to_list(&self) -> Bound<'py, PyList> {
@@ -496,7 +502,7 @@ impl<'py> Iterator for BoundTupleIterator<'py> {
     fn next(&mut self) -> Option<Self::Item> {
         if self.index < self.length {
             let item = unsafe {
-                BorrowedTupleIterator::get_item(Borrowed::from(&self.tuple), self.index).to_owned()
+                BorrowedTupleIterator::get_item(self.tuple.as_borrowed(), self.index).to_owned()
             };
             self.index += 1;
             Some(item)
@@ -517,7 +523,7 @@ impl<'py> DoubleEndedIterator for BoundTupleIterator<'py> {
     fn next_back(&mut self) -> Option<Self::Item> {
         if self.index < self.length {
             let item = unsafe {
-                BorrowedTupleIterator::get_item(Borrowed::from(&self.tuple), self.length - 1)
+                BorrowedTupleIterator::get_item(self.tuple.as_borrowed(), self.length - 1)
                     .to_owned()
             };
             self.length -= 1;
@@ -797,7 +803,7 @@ tuple_conversion!(
 #[cfg(test)]
 #[allow(deprecated)] // TODO: remove allow when GIL Pool is removed
 mod tests {
-    use crate::types::{PyAny, PyList, PyTuple};
+    use crate::types::{any::PyAnyMethods, tuple::PyTupleMethods, PyAny, PyList, PyTuple};
     use crate::{Python, ToPyObject};
     use std::collections::HashSet;
 
@@ -822,8 +828,18 @@ mod tests {
             let ob = (1, 2, 3).to_object(py);
             let tuple: &PyTuple = ob.downcast(py).unwrap();
             assert_eq!(3, tuple.len());
+            assert!(!tuple.is_empty());
             let ob: &PyAny = tuple.into();
             assert_eq!((1, 2, 3), ob.extract().unwrap());
+        });
+    }
+
+    #[test]
+    fn test_empty() {
+        Python::with_gil(|py| {
+            let tuple = PyTuple::empty(py);
+            assert!(tuple.is_empty());
+            assert_eq!(0, tuple.len());
         });
     }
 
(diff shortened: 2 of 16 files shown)


src/instance.rs
```diff
@@ -221,13 +221,25 @@ unsafe impl<T> AsPyPointer for Py2<'_, T> {
...
 
+impl<'py, T> Py2Borrowed<'_, 'py, T> {
+    pub(crate) fn clone_ref(self) -> Py2<'py, T> {
+        unsafe { ffi::Py_INCREF(self.as_ptr()) };
+        Py2(
+            self.py(),
+            ManuallyDrop::new(unsafe { Py::from_non_null(self.0) }),
+        )
+    }
+}
```
davidhewitt, line 241: `.clone_ref()` becomes necessary because implementing `Clone` and `Copy` means that the previous `.clone()` calls I'd been using to convert `Py2Borrowed` -> `Py2` no longer compile as they instead will now just clone `Py2Borrowed`. (Previously they worked because they would first deref to `Py2` before cloning.)

I think this API probably makes sense, it's also consistent in naming with `Py::clone_ref`.

src/types/tuple.rs
```diff
@@ -276,23 +509,103 @@ impl<'a> DoubleEndedIterator for PyTupleIterator<'a> {
...
+    }
+}
+
+impl<'a, 'py> Iterator for PyTupleIterBorrowed<'a, 'py> {
+    type Item = Py2Borrowed<'a, 'py, PyAny>;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.index < self.length {
+            let item = unsafe { Self::get_item(self.tuple, self.index) };
```
davidhewitt, line 573: This is where the `Py2Borrowed` copying occurs, but I can rework this if adding copying now seems too soon.

src/types/tuple.rs
```diff
@@ -276,23 +509,103 @@ impl<'a> DoubleEndedIterator for PyTupleIterator<'a> {
...
+            tuple,
+            index: 0,
+            length,
+        }
+    }
+
+    unsafe fn get_item(
+        tuple: Py2Borrowed<'a, 'py, PyTuple>,
+        index: usize,
+    ) -> Py2Borrowed<'a, 'py, PyAny> {
```
davidhewitt, line 558: Associated method so that `PyTupleIterator2` can also use this.

src/instance.rs
davidhewitt: This is now just `to_owned` as per #3680

davidhewitt:
@adamreichold I've rebased this PR on all the changes we've merged recently.

I also pushed two additional commits to make this PR go further than the other `PyXMethods` PRs.

1. I introduce the `PyTuple::new_bound` and `PyTuple::empty_bound` methods
2. I introduce a `no-pool` feature which is similar to #3685

The interesting idea I had is that with the `no-pool` feature enabled, we can also replace entirely the old names `PyTuple::new` and `PyTuple::empty` with their new forms with return `Bound<'_, PyTuple>`.

There's two nice upsides to this:
- Users who are prepared to put in the work to completely drop the pool can also then immediately opt-in to the future API names without waiting for us to do a great remove-and-replace release. This makes that future release less breaking.
- It makes it easy for us to re-use the existing unit tests for both variants of the API depending on whether the `no-pool` feature is active.

adamreichold:
> The interesting idea I had is that with the no-pool feature enabled, we can also replace entirely the old names PyTuple::new and PyTuple::empty with their new forms with return Bound<'_, PyTuple>.

I have not yet looked at the code in question, but from the description I have to admit that I am not enthusiastic:
* Feature should be additive, i.e. add API and not remove them. Consider a third party crate still depending on the pool API but the top-level removing them by enabling a `no-pool` feature. This makes for a bad developer experience and will likely lead to unnecessary bug reports.
* I also dislike a feature changing methods instead of adding them, i.e. dependencies which do want the pool will access the same functionality under different names if they also interact with the new API. The PyO3 code base already feels a bit like a Rust version of `#ifdef` hell due to the need to support different Python versions, variants and API levels. I would prefer if we did not add to that if we can avoid it.
* I think the reuse is in the unit tests is somewhat accidental, i.e. it works only if the tests happen to never break types when `::new` or `::empty` change their type which give us some coverage for free, but it also raises the bar for understanding and extending these tests.

To me, it feels like we are continuously oscillating between "this will already be a hefty migration, let's avoid any unnecessary breakage" and "that pool-free future is glorious and I want to take everybody there with us, now". Personally, I am strongly in the first camp, I'd like us to use the next release to introduce the new API and their significant benefits to downstream projects but leave it as a completely optional thing that projects can look into at their own pace. Even deprecating the non-pool API is a pretty strong approach IMHO, but at least that can be easily silenced using `#![allow(deprecated)]`.

davidhewitt:
Hmm, yes I take your point. I think the thing that pains me is that it feels like we will be publishing 0.21 (and probably 0.22) with an API story that is very much "this is in migration, and we'll be asking you to join us in finishing off migration in a future release". And maybe that's just the inevitable reality of it.

I agree that this particular proposal of a `no-pool` feature is a bit of an abuse of how features work, and probably that's a good enough reason that we shouldn't do this.

I'll drop that last commit and let's continue to focus on getting the API feature complete first.

tests/ui/invalid_result_conversion.stderr
```diff
@@ -6,8 +6,8 @@ error[E0277]: the trait bound `PyErr: From<MyError>` is not satisfied
    |
    = help: the following other types implement trait `From<T>`:
              <PyErr as From<PyBorrowError>>
-             <PyErr as From<PyBorrowMutError>>
              <PyErr as From<std::io::Error>>
+             <PyErr as From<PyBorrowMutError>>
```
davidhewitt, line 10: From some quick bisecting this diff appears to have been caused by adding `PyTupleMethods` to the prelude, which doesn't make any sense to me. Still, I don't think there's any particular problem with the order here changing (it does between Rust versions anyway).

davidhewitt:
Looking at how these deprecations are affecting even just our own code, I'm beginning to think maybe we should go even slower and not even deprecate stuff? https://github.com/PyO3/pyo3/pull/3703#issuecomment-1869817794

As keen as I am to get everyone onto a safer and more performant API, I think patience and good documentation of how and why the changes are occurring might be a better route than trying to encourage a rapid migration.

adamreichold:
> Looking at how these deprecations are affecting even just our own code, I'm beginning to think maybe we should go even slower and not even deprecate stuff? [#3703 (comment)](https://github.com/PyO3/pyo3/pull/3703#issuecomment-1869817794)
> 
> As keen as I am to get everyone onto a safer and more performant API, I think patience and good documentation of how and why the changes are occurring might be a better route than trying to encourage a rapid migration.

Yes, I think shipped without any deprecations of the GIL ref API would be a better experience for our downstream projects. The main reason, I would like to add them is that this is usually part of our removal story, i.e. deprecate in version N and remove in version N+2. So if do not do this, we need to wait until N+3 for removal? Our maybe we use something like the pinned issue to tell people that we did not deprecate but still plan to remove in N+2 after deprecating in N+1?

(If we add a `pool` feature, downstream projects which want to make sure to migrate could still make sure they fully migrate. But of course, a warning is nicer than an error for the last three usages of GIL refs out of 243 original ones. Maybe a feature that will enable deprecations of the GIL ref API? But that feels like an abuse of features as well?)

davidhewitt:
Going off the `pool` feature idea, maybe this: in 0.21 the feature just turns off the deprecation warnings, i.e. makes these APIs fully available.

In 0.22 we go further and gate the APIs behind the `pool` feature, and even with `pool` feature these APIs are deprecated.

And then in 0.23 (or maybe 0.24 if we feel conservative at the time) we remove the APIs and the feature entirely?

I think that should both create an additive feature and also hopefully a realistic vision for how we can ship 0.21?

tests/ui/invalid_result_conversion.stderr
adamreichold: Yes, I encountered this as well I think the churn is to be expected if we change the input set of types for the compiler. (Using an order that is stable w.r.t. this like alphabetic seems unergonomic for the developer.)

src/types/tuple.rs
adamreichold: I certainly agree that `Borrowed` should `Copy` like shared references which I think we already merged in an earlier PR.

src/types/tuple.rs
```diff
@@ -276,20 +610,17 @@ impl<'a> DoubleEndedIterator for PyTupleIterator<'a> {
...
-impl FusedIterator for PyTupleIterator<'_> {}
-
-impl<'a> IntoIterator for &'a PyTuple {
-    type Item = &'a PyAny;
-    type IntoIter = PyTupleIterator<'a>;
+impl FusedIterator for BorrowedTupleIterator<'_, '_> {}
 
-    fn into_iter(self) -> Self::IntoIter {
-        self.iter()
+impl IntoPy<Py<PyTuple>> for Bound<'_, PyTuple> {
```
adamreichold, line 621: Do we need/have this for the other native types?

src/types/tuple.rs
```diff
@@ -201,54 +219,370 @@ impl PyTuple {
...
+}
+
+impl<'py> Iterator for BoundTupleIterator<'py> {
+    type Item = Bound<'py, PyAny>;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.index < self.length {
+            let item = unsafe {
+                BorrowedTupleIterator::get_item(Borrowed::from(&self.tuple), self.index).to_owned()
```
adamreichold, line 501: I wonder if the amount of reuse could be increased by providing a method to go from `&'a mut BoundTupleIterator<'py>` to `&'a mut BorrowedTupleIterator<'a, 'py`>'?

I am torn on this though: While they have the same memory layout (after slapping on `repr(C)`), unsafe transmuting between feels a bit heavy-handed and the amount code duplication w.r.t. the indexing seems bearable.

src/types/tuple.rs
davidhewitt: Yep we merged this already in #3680 as the `Copy` property kept showing up as useful.

src/types/tuple.rs
davidhewitt: This came up as useful because `PyAny::call_method` has `IntoPy<Py<PyTuple>>` as a bound on the `args` parameter.

I think this implementation could be solved generically but it's a little fiddly because `IntoPy<PyObject> for Bound<'_, T>` can easily conflict with `IntoPy<PyObject> for Bound<'_, PyAny>`. So I'd prefer to just special-case this one for now and punt on the more complete formulation (which might be easier to reason about once the gil refs are removed and there's a lower complexity level in these types again).

src/types/tuple.rs
davidhewitt: Yes I wondered about a similar thing; another option was to have `BoundTupleIterator` be 

```
struct BoundTupleIterator<'py> {
    tup: Bound<'py, PyTuple>,
    borrowed: BorrowedTupleIterator<'static, 'py, PyTuple>
```

where this type is sort of self-referencing; `tup` would own the reference used in `borrowed`, hence why `'static` lifetime on borrowed. It'd need some `unsafe` trickery again though, and in the end the argument of slightly more elegant code reuse didn't seem to justify unsafe here when I was able to formulate a safe solution that was only slightly clunky.

src/types/tuple.rs
```diff
@@ -201,54 +219,370 @@ impl PyTuple {
...
+    ///     let tuple: &PyTuple = ob.downcast(py).unwrap();
+    ///     let obj = tuple.get_item(0);
+    ///     assert_eq!(obj.unwrap().extract::<i32>().unwrap(), 1);
+    ///     Ok(())
+    /// })
+    /// # }
+    /// ```
+    fn get_item(&self, index: usize) -> PyResult<Bound<'py, PyAny>>;
+
+    /// Like `get_item`, but returns a borrowed object, which is a slight performance optimization
```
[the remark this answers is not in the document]
davidhewitt: Good idea, done.

src/types/tuple.rs
```diff
@@ -56,6 +58,23 @@ pub struct PyTuple(PyAny);
 pyobject_native_type_core!(PyTuple, pyobject_native_static_type_object!(ffi::PyTuple_Type), #checkfunction=ffi::PyTuple_Check);
 
 impl PyTuple {
+    /// Deprecated form of `PyTuple::new_bound`.
+    #[track_caller]
+    #[deprecated(
```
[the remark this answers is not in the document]
davidhewitt: I've pushed a feature to action this here so we can see what it looks like, however I've also pushed #3707 to formally discuss and add the feature, as well as give it a proper changelog entry. I'd prefer we merged that PR and then I can rebase here.

src/types/tuple.rs
adamreichold: I agree on the reasoning for the specific impl instead of a generic one. But are tracking these missing specific impls somewhere because I think we should add them before release.

src/types/tuple.rs
adamreichold: Agreed. Let's life with this lack of reuse but the increase in safety it allows.

src/types/tuple.rs
davidhewitt: I'll start an issue tracking bits we might want to polish. 👍

davidhewitt: For now I just added this as extra bullet to #3684

src/types/tuple.rs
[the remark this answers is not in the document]
davidhewitt: Rebased.

adamreichold (approved):

PR #3683, opened 2023-12-21
Some boy scouting w.r.t. usage of our internal helper functions when handling collection ABC

adamreichold:


src/types/mapping.rs
```diff
@@ -130,8 +126,10 @@ impl PyTypeCheck for PyMapping {
         PyDict::is_type_of(object)
             || get_mapping_abc(object.py())
                 .and_then(|abc| object.is_instance(abc))
-                // TODO: surface errors in this chain to the user
-                .unwrap_or(false)
+                .unwrap_or_else(|err| {
+                    err.write_unraisable(object.py(), Some(object));
```
davidhewitt, line 130: Using `write_unraisable` here is a very good idea 👍

davidhewitt (approved):
Thanks, love to see negative-diff tidy ups which also improve QoL for users!

PR #3685, opened 2023-12-21
Hide GILPool behind a feature to test ability to fully avoid it

adamreichold:
Builds even if it obviously doesn't work due to `unimplemented!()` in `register_owned`. cfg'ing this one out and seeing whether we can get the result to compile would be the next after completing the `Py2`-based API.

### DIFF
--- a/src/gil.rs
+++ b/src/gil.rs
@@ -1,12 +1,13 @@
 //! Interaction with Python's global interpreter lock
 
+#[cfg(feature = "pool")]
 use crate::impl_::not_send::{NotSend, NOT_SEND};
 use crate::{ffi, Python};
 use parking_lot::{const_mutex, Mutex, Once};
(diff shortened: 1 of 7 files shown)


src/gil.rs
```diff
@@ -139,17 +140,27 @@ where
...
 
     // Safety: the GIL is already held because of the Py_IntializeEx call.
+    #[cfg(feature = "pool")]
     let pool = GILPool::new();
+    #[cfg(feature = "pool")]
+    let py = pool.python();
+    #[cfg(not(feature = "pool"))]
+    let gil = WithGIL::during_call();
+    #[cfg(not(feature = "pool"))]
+    let py = gil.python();
```
davidhewitt, line 150: Yep this looks the right way to keep `py` bounded correctly 👍

Issue #3684, opened 2023-12-21
New `Bound` API implementation tracking issue


davidhewitt:
This is a tracking issue for #3382 

----
PRs currently waiting for review (all input is greatly appreciated, doesn't need to be an existing maintainer):

#3708 

----

TODO (this list is a fairly good coverage of all PRs which occurred for the Bound implementation, but it is probably not complete):

- [x] Make all of the types and method traits public API
  - [x] ~~What to do with constructors e.g. `PyTuple::new` which return gil refs?~~ We have decided we will introduce `new_bound` variants of these constructors in 0.21 and let migration be incremental over a few release cycles, eventually changing `PyTuple::new` etc. in maybe 0.23.
- [x] Headline entry in migration guide
- [x] Rename `Py2` -> `Bound` and `Py2Borrowed` -> `Borrowed` - #3686
- [x] Finish implementation of all `PyXMethods` traits for types which have `&self` methods
  - [x] `PyCapsuleMethods` - #3770 
  - [x] `PyComplexMethods` - #3767 
  - [x] `datetime` types - #3679 
  - [x] `PyFrozenSetMethods` - #3694 
  - [x] `PyModuleMethods` - #3703 
  - [x] `PySetMethods` - #3694
  - [x] `PySliceMethods` - #3763
  - [x] `PyTracebackMethods` - #3763
  - [x] `PyTupleMethods` - #3681 
  - [x] `PyTypeMethods` - #3705 
- Add `Bound` constructor variants to all Python types, and also `_bound` method variants where applicable (this list is not exhaustive, will add as things are found).
  - [x] `PyComplex` - #3866 
  - [x]  `datetime` types - #3778 
  - [x] `Python` (lots of methods) 
    - #3806 
    - #3816 
    - #3832
    - #3846 
  - [x] `dict` constructors - #3823  
  - [x] `PyEllipsis` - #3797 
  - [x] `PyCFunction` / `PyFunction` - #3800 
  - [x] `PyNotImplemented` - #3797 
  - [x] `PyModule` constructors - #3775 
  - [x] `PyType` - #3705 
  - [x] `PyBuffer::get` - #3836
  - [x] `pyo3::marshal` dumps / loads - #3796 
  - [x] `PyTypeInfo` methods - #3782 
  - [x] `PyErr::from_type`, `PyErr::value` etc. (lots of methods of `PyErr` need updating)
    - #3819 
    - #3820 
    - #3826 
    - #3885  
[...]

PR #3686, opened 2023-12-21
make Bound and Borrowed types public API

davidhewitt:
This completes renaming of our temporary names `Py2` and `Py2Borrowed` to `Bound` and `Borrowed`, and adds these plus the methods traits to the public API (and the prelude).

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -1,4 +1,4 @@
-use crate::instance::Py2;
+use crate::instance::Bound;
 use crate::panic::PanicException;
 use crate::type_object::PyTypeInfo;
 use crate::types::any::PyAnyMethods;
@@ -66,14 +66,14 @@ impl<'a> PyDowncastError<'a> {
 /// Error that indicates a failure to convert a PyAny to a more specific Python type.
 #[derive(Debug)]
 pub(crate) struct PyDowncastError2<'a, 'py> {
-    from: &'a Py2<'py, PyAny>,
+    from: &'a Bound<'py, PyAny>,
     to: Cow<'static, str>,
 }
 
 impl<'a, 'py> PyDowncastError2<'a, 'py> {
     /// Create a new `PyDowncastError` representing a failure to convert the object
     /// `from` into the type named in `to`.
-    pub fn new(from: &'a Py2<'py, PyAny>, to: impl Into<Cow<'static, str>>) -> Self {
+    pub fn new(from: &'a Bound<'py, PyAny>, to: impl Into<Cow<'static, str>>) -> Self {
         PyDowncastError2 {
             from,
             to: to.into(),
@@ -84,14 +84,14 @@ impl<'a, 'py> PyDowncastError2<'a, 'py> {
 /// Error that indicates a failure to convert a PyAny to a more specific Python type.
 #[derive(Debug)]
 pub(crate) struct PyDowncastIntoError<'py> {
-    from: Py2<'py, PyAny>,
+    from: Bound<'py, PyAny>,
     to: Cow<'static, str>,
 }
 
 impl<'py> PyDowncastIntoError<'py> {
     /// Create a new `PyDowncastIntoError` representing a failure to convert the object
     /// `from` into the type named in `to`.
-    pub fn new(from: Py2<'py, PyAny>, to: impl Into<Cow<'static, str>>) -> Self {
+    pub fn new(from: Bound<'py, PyAny>, to: impl Into<Cow<'static, str>>) -> Self {
         PyDowncastIntoError {
             from,
             to: to.into(),
@@ -811,7 +811,7 @@ impl<'a> std::error::Error for PyDowncastError<'a> {}
 
 impl<'a> std::fmt::Display for PyDowncastError<'a> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
-        display_downcast_error(f, Py2::borrowed_from_gil_ref(&self.from), &self.to)
+        display_downcast_error(f, Bound::borrowed_from_gil_ref(&self.from), &self.to)
     }
 }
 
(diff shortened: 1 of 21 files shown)


src/instance.rs
```diff
@@ -836,18 +832,18 @@ where
...
-    pub(crate) fn attach_into(self, py: Python<'_>) -> Py2<'_, T> {
-        Py2(py, ManuallyDrop::new(self))
+    pub(crate) fn attach_into(self, py: Python<'_>) -> Bound<'_, T> {
+        Bound(py, ManuallyDrop::new(self))
     }
 
-    pub(crate) fn attach_borrow<'a, 'py>(&'a self, py: Python<'py>) -> Py2Borrowed<'a, 'py, T> {
-        Py2Borrowed(self.0, PhantomData, py)
+    pub(crate) fn attach_borrow<'a, 'py>(&'a self, py: Python<'py>) -> Borrowed<'a, 'py, T> {
+        Borrowed(self.0, PhantomData, py)
```
davidhewitt, line 846: I think these three methods should probably rename, maybe `attach` -> `bind`, `attach_into` -> `into_bound`, and `attach_borrow` -> `bind_borrowed`?

src/types/boolobject.rs
```diff
@@ -21,13 +21,13 @@ impl PyBool {
     /// Gets whether this boolean is `true`.
     #[inline]
     pub fn is_true(&self) -> bool {
-        Py2::borrowed_from_gil_ref(&self).is_true()
+        Bound::borrowed_from_gil_ref(&self).is_true()
```
davidhewitt, line 24: `Bound::borrowed_from_gil_ref` feels like quite a mouthful now, I wonder if it would be better to add a new trait method `as_bound` to `PyNativeType` so that this can instead be:

```suggestion
        self.as_bound().is_true()
```

We could similarly also have `PyNativeType::as_borrowed`.

Probably these are better done in a follow-up PR to discuss separately in a smaller diff?

src/instance.rs
adamreichold: Yes, that sounds consistent with the chosen names.

src/types/boolobject.rs
adamreichold: Yep, better as a follow-up.

src/err/mod.rs
```diff
@@ -65,15 +65,15 @@ impl<'a> PyDowncastError<'a> {
 
 /// Error that indicates a failure to convert a PyAny to a more specific Python type.
 #[derive(Debug)]
-pub(crate) struct PyDowncastError2<'a, 'py> {
-    from: &'a Py2<'py, PyAny>,
+pub struct PyDowncastError2<'a, 'py> {
```
adamreichold, line 68: Oh, this one needs a matching name as well. `PyBoundDowncastError`?

src/instance.rs
```diff
@@ -172,15 +172,15 @@ impl<'py, T> Py2<'py, T> {
         self.into_non_null().as_ptr()
     }
 
-    /// Internal helper to convert e.g. &'a &'py PyDict to &'a Py2<'py, PyDict> for
+    /// Internal helper to convert e.g. &'a &'py PyDict to &'a Bound<'py, PyDict> for
     /// backwards-compatibility during migration to removal of pool.
     #[doc(hidden)] // public and doc(hidden) to use in examples and tests for now
```
adamreichold, line 177: Should this stay hidden if `Bound` becomes part of the public API? Shouldn't users also be able to switch between GIL references and `Bound` smart pointers?

src/err/mod.rs
davidhewitt: I've added an item for #3684, to remember this (and also the iterator types), if we want to save this for a follow-up? Maybe just `DowncastError` and `DowncastIntoError`? Could also rename the gil-ref one to `PyAnyDowncastError` and make this `PyDowncastError`?

src/instance.rs
davidhewitt: Are you willing to accept this to stay as-is just for the moment, and then in the follow up adding `PyNativeType::as_bound` maybe we can delete this completely?

src/types/boolobject.rs
adamreichold: (The follow-up PR will also give the opportunity to consider what the public API story for the compatibility methods transitioning between old and new API is.)

src/types/list.rs
```diff
@@ -592,14 +594,14 @@ impl<'a> IntoIterator for &'a PyList {
 }
 
 /// Used by `PyList::iter()`.
-pub(crate) struct PyListIterator2<'py> {
-    list: Py2<'py, PyList>,
+pub struct PyListIterator2<'py> {
```
adamreichold, line 597: This one should also loose it `2` in exchange for a `Bound` somewhere.

src/types/dict.rs
```diff
@@ -617,7 +617,7 @@ impl<'py> ExactSizeIterator for PyDictIterator2<'py> {
 }
 
 impl<'py> PyDictIterator2<'py> {
```
adamreichold, line 619: Similarly still referencing `2`

src/instance.rs
adamreichold: Alright, let's consider this together with `as_bound`.

src/err/mod.rs
adamreichold: I think it would like concentrate the renaming here because I mainly just want to be done with it.

> Maybe just DowncastError and DowncastIntoError?

Are my favourites from the list. I would not rename the existing type as it appears as gratuitous churn to me.

src/types/list.rs
adamreichold: Too me, `Bound/BorrowFooIterator` seem the most reasonable of the choices.

davidhewitt: `BoundListIterator` it is 👍

src/types/dict.rs
davidhewitt: `BoundDictIterator` now pushed👍

adamreichold:
`downcast_into_unchecked` is now public yet missing a `# Safety` section.

davidhewitt:
Added, sorry for the slight delay, had a few bits and pieces going on this afternoon.

PR #3689, opened 2023-12-22
Turn calls of __traverse__ into no-ops for unsendable pyclass if on the wrong thread

### LINKED ISSUE #3688: Program aborts when Python's garbage collector gets called from another thread and attempts to traverse an unsendable pyclass instance.
I have created a repository providing a full breakdown and minimal reproducible example of the error 
at [https://github.com/JRRudy1/pyo3_gc_error](https://github.com/JRRudy1/pyo3_gc_error). I will provide a summary below, but please check out 
the repository instead as I put a lot of effort into clearly presenting and investigating the issue.

In summary, I have discovered an error, or perhaps an undocumented limitation, in the way 
PyO3 handles thread-checking for "unsendable" `pyclass` instances as they are being traversed 
by Python's garbage collector (GC). In particular, this occurs when garbage collection is triggered 
from a separate thread, and the pyclasses integrate with the GC by implementing the `__traverse__` 
magic method. The error (or limitation) results in a hard abort, and is particularly problematic 
since it cannot be caught from Python using a `try`/`except` block.

The conditions and sequence of events leading to the error can be summarized as:
1. Two (or more) instances of an "unsendable" pyclass are created from Python
2. The objects are in a reference cycle, and the pyclass defines `__traverse__`/`__clear__` to break it
4. All references to them outside the cycle are dropped, so the next GC cycle should clean them up
5. Before the GC runs automatically, it gets explicitly called from another thread (`gc.collect` from 
    Python or `GcCollect` from C)
6. When the GC calls back into Rust to traverse the objects, PyO3 detects that the calling thread is
    not the original thread and incorrectly deduces that the object was sent between threads
7. PyO3 triggers a panic and the program aborts with a misleading error message

I have gotten reasonably familiar with PyO3's internals and may be interested in working on this, 
but I would need some guidance from an "expert" with a more nuanced understanding of the 
possible implications. It is possible that the limitation cannot be safely fixed, and the only solution 
is to improve the error message and add a warning to the documentation.

As mentioned above, please visit [https://github.com/JRRudy1/pyo3_gc_error](https://github.com/JRRudy1/pyo3_gc_error) for more information.

adamreichold:
Adds a "threadsafe" variant of `PyCell::try_borrow` which will fail instead of panicking if called on the wrong thread and use it in `call_traverse` to turn GC traversals of unsendable pyclasses into no-ops if on the wrong thread.

This can imply leaking the underlying resource if the originator thread has already exited so that the GC will never run there again, but it does avoid hard aborts as we cannot raise an exception from within `call_traverse`.

Closes #3688

### DIFF
--- a/src/pycell.rs
+++ b/src/pycell.rs
@@ -351,6 +351,14 @@ impl<T: PyClass> PyCell<T> {
             .map(|_| PyRef { inner: self })
     }
 
+    /// Variant of [`try_borrow`][Self::try_borrow] which fails instead of panicking if called from the wrong thread
+    pub(crate) fn try_borrow_threadsafe(&self) -> Result<PyRef<'_, T>, PyBorrowError> {
+        self.check_threadsafe()?;
+        self.borrow_checker()
+            .try_borrow()
+            .map(|_| PyRef { inner: self })
+    }
+
     /// Mutably borrows the value `T`, returning an error if the value is currently borrowed.
     /// This borrow lasts as long as the returned `PyRefMut` exists.
     ///
@@ -975,6 +983,7 @@ impl From<PyBorrowMutError> for PyErr {
 #[doc(hidden)]
 pub trait PyCellLayout<T>: PyLayout<T> {
     fn ensure_threadsafe(&self);
+    fn check_threadsafe(&self) -> Result<(), PyBorrowError>;
     /// Implementation of tp_dealloc.
     /// # Safety
     /// - slf must be a valid pointer to an instance of a T or a subclass.
@@ -988,6 +997,9 @@ where
     T: PyTypeInfo,
 {
     fn ensure_threadsafe(&self) {}
+    fn check_threadsafe(&self) -> Result<(), PyBorrowError> {
+        Ok(())
+    }
     unsafe fn tp_dealloc(py: Python<'_>, slf: *mut ffi::PyObject) {
         let type_obj = T::type_object_raw(py);
         // For `#[pyclass]` types which inherit from PyAny, we can just call tp_free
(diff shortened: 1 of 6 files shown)


adamreichold:
@JRRudy1 Could you use a dependency specification like

```toml
pyo3 = { git = "https://github.com/PyO3/pyo3.git", branch = "unsendable-threadsafe-traverse" }
```

and verify that this would work for your use case? Thanks.

src/pycell.rs
```diff
@@ -975,6 +983,7 @@ impl From<PyBorrowMutError> for PyErr {
 #[doc(hidden)]
 pub trait PyCellLayout<T>: PyLayout<T> {
     fn ensure_threadsafe(&self);
+    fn check_threadsafe(&self) -> Result<(), PyBorrowError>;
```
davidhewitt, line 986: Potentially this should return `bool` to match the `PyClassThreadChecker` trait implementation? It would make `try_borrow_threadsafe` implementation a bit messier but might be nice for consistency.

davidhewitt (commented):
Thanks, nice! Looks straightforward enough.

I wonder if there's a way to add a test for this. I suspect the most reasonable option, if we felt it worth it, would be to add a test which manually calls the traverse slot?

src/pycell.rs
adamreichold: I had written down both and ended up with this version because the calling (i.e. `try_borrow_threadsafe`) was more consistent this way without the `PyCellLayout` impl becoming more complicated. The other way around, both impls looked messier. But to be honest, the differences were small/purely aesthetic either way, so If you'd like me to, I can certainly change this back.

adamreichold:
> I wonder if there's a way to add a test for this. I suspect the most reasonable option, if we felt it worth it, would be to add a test which manually calls the traverse slot?

Will give it a try.

I also noticed that this should be mentioned in the guide...

src/pycell.rs
davidhewitt: I trust your taste for good design, so let's keep it like this then 👍

davidhewitt (approved):
Looks great to me, thanks, great test 👍

PR #3690, opened 2023-12-22
Copy note on using check_signals on non-main thread/interpreter from Python docs.

### LINKED ISSUE #3682: check_signals should mention its restrictions on being called from threads
[`check_signals`](https://docs.rs/pyo3/latest/pyo3/marker/struct.Python.html#method.check_signals) calls [`PyErr_CheckSignals`](https://docs.python.org/3/c-api/exceptions.html?highlight=pyerr_checksignals#c.PyErr_CheckSignals) under the hood (which it does mention). But I feel this caveat from `PyErr_CheckSignals` is rather important:

> If the function is called from a non-main thread, or under a non-main Python interpreter, it does nothing and returns 0.

This warning should probably be replicated on the PyO3 function, because not everyone clicks through to read the documentation of `PyErr_CheckSignals`.

adamreichold:
Closes #3682

### DIFF
--- a/src/marker.rs
+++ b/src/marker.rs
@@ -906,6 +906,9 @@ impl<'py> Python<'py> {
     /// As Python's [`signal`][2] API allows users to define custom signal handlers, calling this
     /// function allows arbitrary Python code inside signal handlers to run.
     ///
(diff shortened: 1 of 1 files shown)


davidhewitt (approved):
👍

PR #3692, opened 2023-12-23
Add `as_borrowed` conversion from gil-refs to `Bound<T>`

davidhewitt:
Update: we changed to add `.as_borrowed()` instead of `.as_bound()` as it gave additional flexibility.

----

This is the promised follow-up to #3686 which adds a `.as_bound()` method to the GIL refs API types to convert into the new `Bound<T>` smart pointer.

I've switched all internal code over to use this API to give a feel of how it reads; I think it's much nicer to type and read `self.as_bound()` over `Bound::borrowed_from_gil_ref(&self)`. Hopefully this helps users to migrate.

I think the next PR I plan to open will be a beginning of the migration guide entry; I'd just like this conversion API agreed upon first as this will form a key part of migration instructions.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -197,6 +197,15 @@ impl<'py, T> Bound<'py, T> {
         self.into_non_null().as_ptr()
     }
 
+    /// Casts this `Bound<T>` to a `Borrowed<T>` smart pointer.
+    pub fn as_borrowed<'a>(&'a self) -> Borrowed<'a, 'py, T> {
+        Borrowed(
+            unsafe { NonNull::new_unchecked(self.as_ptr()) },
+            PhantomData,
+            self.py(),
+        )
+    }
+
     /// Casts this `Bound<T>` as the corresponding "GIL Ref" type.
     ///
     /// This is a helper to be used for migration from the deprecated "GIL Refs" API.
@@ -300,11 +309,7 @@ impl<'a, 'py> Borrowed<'a, 'py, PyAny> {
 impl<'a, 'py, T> From<&'a Bound<'py, T>> for Borrowed<'a, 'py, T> {
     /// Create borrow on a Bound
     fn from(instance: &'a Bound<'py, T>) -> Self {
-        Self(
-            unsafe { NonNull::new_unchecked(instance.as_ptr()) },
-            PhantomData,
-            instance.py(),
-        )
+        instance.as_borrowed()
     }
 }
 
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -28,6 +28,19 @@ pub unsafe trait PyNativeType: Sized {
     /// The form of this which is stored inside a `Py<T>` smart pointer.
     type AsRefSource: HasPyGilRef<AsRefTarget = Self>;
 
+    /// Cast `&self` to a `Borrowed` smart pointer.
+    ///
+    /// `Borrowed<T>` implements `Deref<Target=Bound<T>>`, so can also be used in locations
+    /// where `Bound<T>` is expected.
+    ///
+    /// This is available as a migration tool to adjust code from the deprecated "GIL Refs"
+    /// API to the `Bound` smart pointer API.
+    fn as_borrowed(&self) -> Borrowed<'_, '_, Self::AsRefSource> {
+        // Safety: &'py Self is expected to be a Python pointer,
+        // so has the same layout as Borrowed<'py, 'py, T>
+        unsafe { std::mem::transmute(self) }
+    }
+
     /// Returns a GIL marker constrained to the lifetime of this type.
     #[inline]
     fn py(&self) -> Python<'_> {
@@ -184,18 +197,6 @@ impl<'py, T> Bound<'py, T> {
         self.into_non_null().as_ptr()
     }
 
-    /// Internal helper to convert e.g. &'a &'py PyDict to &'a Bound<'py, PyDict> for
-    /// backwards-compatibility during migration to removal of pool.
-    #[doc(hidden)] // public and doc(hidden) to use in examples and tests for now
-    pub fn borrowed_from_gil_ref<'a, U>(gil_ref: &'a &'py U) -> &'a Self
-    where
-        U: PyNativeType<AsRefSource = T>,
-    {
-        // Safety: &'py T::AsRefTarget is expected to be a Python pointer,
-        // so &'a &'py T::AsRefTarget has the same layout as &'a Bound<'py, T>
-        unsafe { std::mem::transmute(gil_ref) }
-    }
-
     /// Casts this `Bound<T>` as the corresponding "GIL Ref" type.
     ///
     /// This is a helper to be used for migration from the deprecated "GIL Refs" API.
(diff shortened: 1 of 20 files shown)


src/instance.rs
```diff
@@ -27,6 +27,16 @@ pub unsafe trait PyNativeType: Sized {
     /// The form of this which is stored inside a `Py<T>` smart pointer.
     type AsRefSource: HasPyGilRef<AsRefTarget = Self>;
 
+    /// Cast `&self` to a `Bound` smart pointer.
+    ///
+    /// This is available as a migration tool to adjust code from the deprecated "GIL Refs"
+    /// API to the `Bound` smart pointer API.
+    fn as_bound<'a, 'py>(self: &'a &'py Self) -> &'a Bound<'py, Self::AsRefSource> {
```
adamreichold, line 34: Wouldn't the conversion to `Borrowed` be more natural here? Or should we have both?

src/instance.rs
```diff
@@ -172,18 +182,6 @@ impl<'py, T> Bound<'py, T> {
...
-    where
-        U: PyNativeType<AsRefSource = T>,
-    {
-        // Safety: &'py T::AsRefTarget is expected to be a Python pointer,
-        // so &'a &'py T::AsRefTarget has the same layout as &'a Bound<'py, T>
-        unsafe { std::mem::transmute(gil_ref) }
-    }
-
     /// Internal helper to get to pool references for backwards compatibility
     #[doc(hidden)] // public and doc(hidden) to use in examples and tests for now
```
adamreichold, line 186: I think we also need to make these into proper public API for the reverse direction?

src/types/any.rs
```diff
@@ -2193,7 +2173,7 @@ impl<'py> PyAnyMethods<'py> for Bound<'py, PyAny> {
 
     #[cfg(not(PyPy))]
     fn py_super(&self) -> PyResult<Bound<'py, PySuper>> {
-        PySuper::new2(Bound::borrowed_from_gil_ref(&self.get_type()), self)
+        PySuper::new2(self.get_type().as_bound(), self)
```
adamreichold, line 2176: `new_bound` instead of `new2`?

src/types/iterator.rs
```diff
@@ -34,7 +34,7 @@ impl PyIterator {
     ///
     /// Equivalent to Python's built-in `iter` function.
     pub fn from_object(obj: &PyAny) -> PyResult<&PyIterator> {
-        Self::from_object2(Bound::borrowed_from_gil_ref(&obj)).map(Bound::into_gil_ref)
+        Self::from_object2(obj.as_bound()).map(Bound::into_gil_ref)
```
adamreichold, line 37: `from_bound(_object)` instead of `from_object2`?

src/types/pysuper.rs
```diff
@@ -57,18 +57,15 @@ impl PySuper {
...
     pub fn new<'py>(ty: &'py PyType, obj: &'py PyAny) -> PyResult<&'py PySuper> {
-        Self::new2(
-            Bound::borrowed_from_gil_ref(&ty),
-            Bound::borrowed_from_gil_ref(&obj),
-        )
-        .map(Bound::into_gil_ref)
+        Self::new2(ty.as_bound(), obj.as_bound()).map(Bound::into_gil_ref)
     }
 
     pub(crate) fn new2<'py>(
```
adamreichold, line 63: `new_bound`

src/instance.rs
davidhewitt: Yes I'll add both. I want to mostly encourage users to use `Bound` as much as possible so that they don't get overwhelmed by choice, and only use `Borrowed` when it's really needed.

src/instance.rs
davidhewitt: Sure thing, it is definitely reasonable to have them even even if they have clunky names and we encourage users to use `.as_bound()` / `.as_borrowed()`.

src/types/any.rs
davidhewitt: How would you feel if I just changed `PySuper::new` to return `Bound<T>` and same for `PyIterator::from_object`? I think that churn is probably in the category of acceptable. (And I was thinking we would do this for `PyTuple::new` and constructors in general.)

I guess we could also introduce these as `_bound` variants for 0.21 and deprecate the original names, and then change the return types of the original names later in 0.22 or 0.23?

davidhewitt: Actually now I think a little more it's probably indeed better to add `_bound` variants for 0.21 and just deprecate the old names? I'd rather break as little user code as possible given the size of the API change already.

src/instance.rs
adamreichold: > we encourage users to use .as_bound() / .as_borrowed().

I am possibly missing something. I though these are not redundant to `as_bound` (GIL -> bound) as they provide the reverse direction (bound -> GIL)?

src/instance.rs
adamreichold: I was thinking about reducing choice as one can always go from `Borrowed` to `Bound` automatically via `Deref`, i.e. we actually only need `Borrowed` and producing a full `Bound` will just be a call to `.to_owned()` in either approaches.

(I would even consider having `as_bound` return `Borrowed` but that does seem a bit confusing.)

src/types/any.rs
adamreichold: > Actually now I think a little more it's probably indeed better to add _bound variants for 0.21 and just deprecate the old names? I'd rather break as little user code as possible given the size of the API change already.

Yes, I think I would prefer that too. I guess there will eventually be a great reckoning release where we remove all the deprecated functionality and start transitioning to use the shorter names for the long-term approach.

src/instance.rs
davidhewitt: Sorry my mistake, from the location of the comment I thought earlier this was suggesting to make `Bound::borrowed_from_gil_ref` public API instead of deleting it. I see now that this is suggesting to make `.as_gil_ref()` public; we _definitely_ want to do that!

src/instance.rs
davidhewitt: Actually I like that idea of reducing choice, as it makes the general pattern much simpler to just have `.as_borrowed()` everywhere. I'll pushed a commit to switch to that.

davidhewitt:
Well, I added commits for all the suggestions here, but this PR has gotten a bit bulky as a result.

I've split some pieces and related changes into #3694,  #3695, #3697 and #3698.

Let's review those separately and then I'll come back to this one to rebase it and we can review again.

davidhewitt:
Given that this PR is one of those PRs which conflicts with a lot of the remaining PRs which I'd like to open wrt `PyXMethods` trait implementations, I've come back here and dropped all the commits which are addressed in other PRs.

That means that this should now be reviewable / mergeable again and the only feedback not addressed here is the `PyIterator::from_object2` change which is open to merge separately in #3702.

If this one does merge I'll rebase all the other PRs which conflict with it 🧑‍💻

adamreichold (approved):

PR #3693, opened 2023-12-24
begin drafting Bound<T> migration guide

davidhewitt:
This PR is a start of the migration guide to `Bound<T>`. I think there's still a lot more to be added here, but I wanted to push this anyway so that we can have a central place to start collecting migration notes as well as agree on the way we want to communicate this to users.

### DIFF
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -5,6 +5,40 @@ For a detailed list of all changes, see the [CHANGELOG](changelog.md).
 
 ## from 0.20.* to 0.21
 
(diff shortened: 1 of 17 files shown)


guide/src/migration.md
```diff
@@ -5,6 +5,40 @@ For a detailed list of all changes, see the [CHANGELOG](changelog.md).
...
+
+The "GIL Ref" `&'py PyAny` and similar types such as `&'py PyDict` continue to be available as a deprecated API. Due to the advantages of the new API it is advised that all users make the effort to upgrade as soon as possible.
+
+In addition to the major API type overhaul, PyO3 has needed to make a few small breaking adjustments to other APIs to close correctness and soundness gaps.
+
+The recommended steps to update to PyO3 0.21 is as follows:
+  1. Add compatibility code to convert new `Bound<T>` smart pointers back to GIL Refs
```
davidhewitt, line 15: Ok I think this is _not_ the approach we are taking any more, I'll come back to rewrite this section later.

davidhewitt:
Superseded by #3707

PR #3695, opened 2023-12-24
implement iterator for `Bound` iterator

davidhewitt:
Implements the ability to iterate `Bound<'_, PyIterator>` and `&Bound<'_, PyIterator>` in the same way that `&'py PyIterator` can already be iterated.

### DIFF
--- a/src/types/iterator.rs
+++ b/src/types/iterator.rs
@@ -1,4 +1,5 @@
 use crate::ffi_ptr_ext::FfiPtrExt;
+use crate::instance::Borrowed;
 use crate::py_result_ext::PyResultExt;
 use crate::{
     ffi, AsPyPointer, Bound, PyAny, PyDowncastError, PyErr, PyNativeType, PyResult, PyTypeCheck,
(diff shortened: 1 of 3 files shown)


src/types/iterator.rs
```diff
@@ -46,6 +46,28 @@ impl PyIterator {
...
+/// Implementation of functionality for [`PyIterator`].
+///
+/// These methods are defined for the `Bound<'py, PyIterator>` smart pointer, so to use method call
+/// syntax these methods are separated into a trait, because stable Rust does not yet support
+/// `arbitrary_self_types`.
+#[doc(alias = "PyIterator")]
+pub trait PyIteratorMethods<'py> {
+    /// Equivalent of `next(self)`.
+    fn next(&self) -> Option<PyResult<Bound<'py, PyAny>>>;
+}
```
davidhewitt, line 58: Actually it's probably not a good idea to have this `next()` as it'll create a name clash, will rework.

davidhewitt: Ok, this PR should be reviewable now.

src/types/iterator.rs
```diff
@@ -56,18 +56,60 @@ impl<'p> Iterator for &'p PyIterator {
...
 
-        match unsafe { py.from_owned_ptr_or_opt(ffi::PyIter_Next(self.0.as_ptr())) } {
+    #[cfg(not(Py_LIMITED_API))]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let hint = unsafe { ffi::PyObject_LengthHint(self.as_ptr(), 0) };
+        (hint.max(0) as usize, None)
+    }
+}
+
+impl<'py> Iterator for &'_ Bound<'py, PyIterator> {
```
adamreichold, line 91: Could you explain why this impl is needed? I would have assumed that std's [`by_ref`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.by_ref) method would be sufficient? (I understand a mutable reference is strange in Python land, it is a mutable reference to pointer after all which should straight forward to produce even in a PyO3 program (at worst by cloning an existing `&Bound`).

src/types/iterator.rs
```diff
@@ -56,18 +56,60 @@ impl<'p> Iterator for &'p PyIterator {
...
             Some(obj) => Some(Ok(obj)),
             None => PyErr::take(py).map(Err),
         }
     }
 
     #[cfg(not(Py_LIMITED_API))]
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let hint = unsafe { ffi::PyObject_LengthHint(self.0.as_ptr(), 0) };
-        (hint.max(0) as usize, None)
+        (*self).size_hint()
```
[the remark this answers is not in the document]
davidhewitt: Good catch, yes will tidy up. Originally both were on `Bound` but that didn't work for `next` as `&Bound` can't be cast to `&mut Bound`.

src/types/iterator.rs
davidhewitt: I feel like it was convenient in some context when I was experimenting with this API in the past (which is why it's ended up here), but I can't recall the concrete example right now so let's delete it and can add it back later if we find motivation.

davidhewitt:
Uff, that second commit depends on #3694 so to avoid a deadlock I've split that out into #3702. This PR is the first of the three that is mergeable 😮‍💨

adamreichold (approved):

PR #3697, opened 2023-12-24
expose `Bound::as_gil_ref` and `Bound::into_gil_ref`

davidhewitt:
This PR exposes `Bound::as_gil_ref` and `Bound::into_gil_ref` as deprecated public APIs.

As a general strategy for making the deprecation warnings manageable internally, for the blocks like `impl PyMapping` which implements the GIL ref API for `PyMapping`, I've put `#[allow(deprecated)]` on the whole block. IMO I think this is better than marking `#[allow(deprecated)]` all uses inside those methods, but I can be persuaded otherwise.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -183,8 +183,9 @@ impl<'py, T> Bound<'py, T> {
         unsafe { std::mem::transmute(gil_ref) }
     }
 
-    /// Internal helper to get to pool references for backwards compatibility
-    #[doc(hidden)] // public and doc(hidden) to use in examples and tests for now
+    /// Casts this `Bound<T>` as the corresponding "GIL Ref" type.
+    ///
+    /// This is a helper to be used for migration from the deprecated "GIL Refs" API.
     pub fn as_gil_ref(&'py self) -> &'py T::AsRefTarget
     where
         T: HasPyGilRef,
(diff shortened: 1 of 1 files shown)


src/instance.rs
```diff
@@ -184,17 +187,28 @@ impl<'py, T> Bound<'py, T> {
         unsafe { std::mem::transmute(gil_ref) }
     }
 
-    /// Internal helper to get to pool references for backwards compatibility
-    #[doc(hidden)] // public and doc(hidden) to use in examples and tests for now
+    /// Casts this `Bound<T>` as the corresponding "GIL Ref" type.
+    ///
+    /// This is a helper to be used for migration from the deprecated "GIL Refs" API.
+    #[deprecated(
```
adamreichold, line 193: Do we really want to introduce a newly introduced method? Especially since downstream might need these methods to e.g. interface with third-party code which might not be fast to change (cough rust-numpy cough).

I would rather have expected that we e.g. deprecate the whole `impl PyAny` block that is replaced by `PyAnyMethods`?

davidhewitt: Hmm good point, so we deprecate instead all the APIs which we plan to remove. Yes, I think that's better, even if it's more typing on our side. I'll update this PR to see what that looks like later. 👍

davidhewitt: To help keep PRs small and mergeable I've just updated this PR purely to make these APIs public, we can return to the question of deprecation later on once more of the APIs we need are in place (mostly the `PyXMethods` traits, I think).

adamreichold (approved):

PR #3698, opened 2023-12-24
implement `PyErr::write_unraisable_bound`

davidhewitt:
In the spirit of the suggestion to go for the `deprecate-and-add-bound-variant` approach, here's a PR which implements this change for `PyErr::write_unraisable`, as I bumped into this when implementing #3697


src/instance.rs
```diff
@@ -98,9 +98,7 @@ fn python_format(
 ) -> Result<(), std::fmt::Error> {
     match format_result {
         Result::Ok(s) => return f.write_str(&s.as_gil_ref().to_string_lossy()),
-        Result::Err(err) => {
-            err.write_unraisable(any.py(), std::option::Option::Some(any.as_gil_ref()))
-        }
+        Result::Err(err) => err.write_unraisable_bound(any.py(), std::option::Option::Some(any)),
```
adamreichold, line 101: Why is `Some` fully qualified here?

adamreichold (approved):
LGTM. One nit in `python_format`.

src/instance.rs
davidhewitt: Probably a copy-and-paste, will fixup 👍

davidhewitt:
Rebased to fix the compile errors.

FYI I added a commit to export `Bound` and `Borrowed` as `pyo3::Bound` and `pyo3::Borrowed` so that the doctests and macros can work outside of the crate. I don't think that's controversial so I'll proceed with the merge, we can adjust this if there are other opinions about what their public path should be.

PR #3701, opened 2023-12-24
Don't try to run the gevent tests on py37

alex:
The version of gevent we require is 3.8+ only


davidhewitt:
Ah, whoops, thanks for picking this up! I'd slightly prefer we downgrade the gevent required to be 3.7 compatible, I don't think that will materially change the test.

alex:
That works.

davidhewitt (approved):
Thanks 👍

PR #3702, opened 2023-12-24
introduce PyIterator::from_bound_object

davidhewitt:
Follows the standard pattern we're now going for to introduce a new bound constructor while deprecating the old one.

This isn't mergeable until after #3694

### DIFF
--- a/src/types/iterator.rs
+++ b/src/types/iterator.rs
@@ -31,14 +31,23 @@ pyobject_native_type_named!(PyIterator);
 pyobject_native_type_extract!(PyIterator);
 
 impl PyIterator {
-    /// Constructs a `PyIterator` from a Python iterable object.
-    ///
-    /// Equivalent to Python's built-in `iter` function.
+    /// Deprecated form of `PyIterator::from_bound_object`.
+    #[cfg_attr(
+        not(feature = "gil-refs"),
+        deprecated(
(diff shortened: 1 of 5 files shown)


src/types/iterator.rs
```diff
@@ -31,14 +31,22 @@ pyobject_native_type_named!(PyIterator);
...
+    )]
     pub fn from_object(obj: &PyAny) -> PyResult<&PyIterator> {
-        Self::from_object2(&obj.as_borrowed()).map(Bound::into_gil_ref)
+        Self::from_bound_object(&obj.as_borrowed()).map(Bound::into_gil_ref)
     }
 
-    pub(crate) fn from_object2<'py>(obj: &Bound<'py, PyAny>) -> PyResult<Bound<'py, PyIterator>> {
+    /// Constructs a `PyIterator` from a Python iterable object.
+    ///
+    /// Equivalent to Python's built-in `iter` function.
```
adamreichold, line 48: I think an intra doc link to `PyAnyMethods::iter` would be nice as this is usually the more convenient form compared to calling this directly.

adamreichold (approved):

src/types/iterator.rs
davidhewitt: Agreed 👍

PR #3703, opened 2023-12-26
implement `PyModuleMethods`

davidhewitt:
Split from #3606

### DIFF
--- a/src/types/module.rs
+++ b/src/types/module.rs
@@ -1,11 +1,12 @@
 use crate::callback::IntoPyCallbackOutput;
 use crate::err::{PyErr, PyResult};
-use crate::exceptions;
-use crate::ffi;
+use crate::ffi_ptr_ext::FfiPtrExt;
 use crate::pyclass::PyClass;
-use crate::types::{PyAny, PyCFunction, PyDict, PyList, PyString};
-use crate::{IntoPy, Py, PyObject, Python};
-use std::ffi::{CStr, CString};
+use crate::types::{
+    any::PyAnyMethods, list::PyListMethods, PyAny, PyCFunction, PyDict, PyList, PyString,
+};
+use crate::{exceptions, ffi, Bound, IntoPy, Py, PyNativeType, PyObject, Python};
+use std::ffi::CString;
 use std::str;
 
 /// Represents a Python [`module`][1] object.
@@ -151,11 +152,7 @@ impl PyModule {
 
     /// Returns the module's `__dict__` attribute, which contains the module's symbol table.
     pub fn dict(&self) -> &PyDict {
-        unsafe {
-            // PyModule_GetDict returns borrowed ptr; must make owned for safety (see #890).
-            let ptr = ffi::PyModule_GetDict(self.as_ptr());
-            self.py().from_owned_ptr(ffi::_Py_NewRef(ptr))
-        }
+        self.as_borrowed().dict().into_gil_ref()
     }
 
     /// Returns the index (the `__all__` attribute) of the module,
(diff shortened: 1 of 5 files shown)


davidhewitt:
I might be able to drive up coverage by adding more tests & also following the same strategy of `no-pool` testing as per https://github.com/PyO3/pyo3/pull/3681#issuecomment-1869612858

davidhewitt:
Well, I started down the road of adding the new module constructors, and deprecating the existing ones.

It creates _a lot_ of documentation churn to deal with the deprecation warnings, which is pretty painful to deal with.

In the spirit of https://github.com/PyO3/pyo3/pull/3681#issuecomment-1869743978, I'm now wondering if we should slow things down even more and start purely by introducing the `_bound` variants of the APIs without deprecating anything in PyO3 0.21.

Once we have a complete implementation released in 0.21 we can start thinking for what the best way to start encouraging users to migrate should be. This also gives more time for `rust-numpy` and other downstream crates to catch up.

davidhewitt:
If we agree that not deprecating anything for now might be better, I'll open a PR to undo the deprecations already merged.

alex:
FWIW, from a user perspective, I'd find the warnings valuable. Basically every use of the old APIs is a performance opportunity, so having those clearly marked is helpful.

davidhewitt:
I do agree with that too, and adding these warnings was helping me find all the locations internally in PyO3 where these are used.

I suppose we are happy with documentation being updated to encourage users to use the new APIs? It's more churn in our code here, but maybe we just accept that it's going to be period of a lot of churn in PyO3 and we just have to look forward to the "great reckoning release".

That said, introducing and deprecating the module constructors here produced a lot of busywork so I definitely don't want to do the deprecations as part of this diff.

adamreichold:
Sorry for distributing the discussion over multiple places. (Maybe we should use #3882?) See also https://github.com/PyO3/pyo3/pull/3681#issuecomment-1870994678 with the TL;DR that I think shipping without deprecations by default would be nicer. Downstream projects who want to commit to a complete migration can disable the `pool` feature.

src/types/module.rs
```diff
@@ -388,8 +349,294 @@ impl PyModule {
...
+    }
+
+    fn add<V>(&self, name: &str, value: V) -> PyResult<()>
+    where
+        V: IntoPy<PyObject>,
+    {
+        self.index()?
+            .append(name)
+            .expect("could not append __name__ to __all__");
+        self.setattr(name, value.into_py(self.py()))
```
adamreichold, line 594: If `add_wrapped` has a debloating `inner` method, shouldn't `add` do so as well?

src/types/module.rs
```diff
@@ -388,8 +349,294 @@ impl PyModule {
...
+    }
+
+    fn add_function(&self, fun: &Bound<'_, PyCFunction>) -> PyResult<()> {
+        let name = fun.getattr(__name__(self.py()))?;
+        self.add(name.extract()?, fun)
+    }
+}
+
+impl<'a, 'py> Borrowed<'a, 'py, PyModule> {
+    fn name(self) -> PyResult<&'a str> {
```
adamreichold, line 630: This works because the name of a module is immutable? If so, a comment might be nice.

davidhewitt: Great catch, it looks to me like this is not sound, the `__name__` is just stored in the module dictionary and this could easily produce a dangling reference. Will change this.

https://github.com/python/cpython/blob/3f84a19e6291db682fc9a570e7612e80e2ffbbb5/Objects/moduleobject.c#L540-L550

davidhewitt:
Thanks as always for the review, pushed an updated form of the commit here. I'll still defer adding the constructors / feature-toggled deprecations in a follow up PR.

Making these changes I also wondered whether we want to split the part which is for building new modules out into a `PyModuleBuilder` at some point in the future. The current API allows to import and modify `builtins`, for example. I think that's an unrelated complication which might make more sense to consider along with #3294 in a future simpler PyO3.

adamreichold (approved):

PR #3705, opened 2023-12-26
implement `PyTypeMethods`

davidhewitt:
Split from #3606 

This PR implements the `PyTypeMethods` trait and the bound variants of the constructors for `PyType`.

The most interesting thing in this PR is the bound variant of `PyType::from_type_ptr`, which I decided to introduce as `PyType::from_type_ptr_borrowed`, because that pointer doesn't carry ownership.

There's possibly arguments that one (or both) of the `PyType` constructors should not have bound variants introduced and we should rework the APIs, e.g.: 
- Deprecate `PyType::new` in favour of `T::type_object` (or `Python::get_type`)
- Find a way to make `PyType::from_type_ptr_borrowed` less like `std::slice::from_raw_parts`, e.g. should it force a reference count increase?


davidhewitt:
Let's make a decision about https://github.com/PyO3/pyo3/pull/3703#issuecomment-1869817794 before reviewing this; if we agree to slow things down I'll remove the deprecations from here.

davidhewitt:
Okay, rebased. To make the deprecation in the macro output conditional I added a `gil-refs` feature to `pyo3-macros`. I think that's fine as we'll eventually want that feature to enable / disable creating a `GILPool` in `#[pyfunction]` calls.

Also, I'm really not feeling keen on `from_type_ptr_borrowed` so I might ponder better replacements for that. Suggestions welcome.

adamreichold:
> Okay, rebased. To make the deprecation in the macro output conditional I added a gil-refs feature to pyo3-macros. I think that's fine as we'll eventually want that feature to enable / disable creating a GILPool in #[pyfunction] calls.

IIRC, all `GILPool` usages are already moved into the `impl_` module. Nevertheless, I don't see a reason to avoid forwarding that feature into our macro helper crate.

pyo3-macros-backend/src/method.rs
```diff
@@ -125,8 +125,14 @@ impl FnType {
...
                 let py = syn::Ident::new("py", Span::call_site());
                 let slf: Ident = syn::Ident::new("_slf", Span::call_site());
+                let allow_deprecated = if cfg!(feature = "gil-refs") {
+                    quote!()
+                } else {
+                    quote!(#[allow(deprecated)])
+                };
                 quote_spanned! { *span =>
                     #[allow(clippy::useless_conversion)]
+                    #allow_deprecated
```
adamreichold, line 135: I am not sure if I follow things correctly: If the feature is enabled, the GIL ref methods are not deprecated and we do need to suppress the warning. If the feature is not disabled, then we do suppress it. But I thought we actually want to either warn about the usage of the deprecated API or use the new API in that case? Is it impossible to avoid the GIL ref API here?

davidhewitt: I stuck with the GIL-Ref API here because I feel like the unconstrained `'a` lifetime in `from_type_ptr_borrowed` is waiting to introduce accidental unsoundness, and I didn't feel comfortable putting it in the macro output here. I want to come up with a better way to solve this one than straight-up switch to `from_type_ptr_borrowed` (and I'm also ok if we block merging this PR until we figure out a solution).

davidhewitt: The simplest I can think of right now is to put a `where 'py: 'a` bound on `from_type_ptr_borrowed`, which at least works for this case and helps add a guard rail against arbitrarily long "borrowed references" on the type ptr. I'd still love to have something better if possible.

adamreichold: I am wondering again if should have this bound on the `Borrowed` type itself just as we it is implicitly present for `&'a T`, `Borrowed<'a, 'py, T>` standing in for `&'a Bound<'py, T>` should have a `Bound<'py, T>: 'a` bound, meaning at least `'py: 'a`.

Of course, we do need to extend the lifetime at some place, but I suspect having to use unsafe there might be preferable to having this foot gun present in an `Borrowed`-based API.

davidhewitt: Yes, this might be a tidy solution for now. It's probably non breaking to relax that constraint again later.

I'll try to push a commit later today which implements this adjustment so we can see how it looks.

davidhewitt: Pushed as a third commit; I definitely don't love the constraint, I had to add 4 transmutes to extend the lifetimes back for the `Py<PyString>` and `Py<PyBytes>` access.

I think the main thing which might inform a better design is what type would we prefer users to use for `#[classmethod]`? It's a borrowed reference, so should they take `cls: Borrowed<PyType>`, or `cls: &Bound<PyType>`?

For now as implemented they would take `Borrowed<PyType>` but `&Bound<PyType>` might be more consistent with what we're doing elsewhere, and the reference would necessarily be bounded by the duration the pointer is on the stack, which would provide an alternative route to making this safe.

adamreichold: > Pushed as a third commit; I definitely don't love the constraint, I had to add 4 transmutes to extend the lifetimes back for the Py<PyString> and Py<PyBytes> access.

I agree that the `transmute`s are bad, but we could also add a more misuse resistant helper method which would only ever extend the first lifetime. That said...

> I think the main thing which might inform a better design is what type would we prefer users to use for #[classmethod]? It's a borrowed reference, so should they take cls: Borrowed<PyType>, or cls: &Bound<PyType>?

> For now as implemented they would take Borrowed<PyType> but &Bound<PyType> might be more consistent with what we're doing elsewhere, and the reference would necessarily be bounded by the duration the pointer is on the stack, which would provide an alternative route to making this safe.

...I agree that using `&Bound<'py, PyType>` might be the preferable approach here. The main issue I see is that we need to somehow document and enforce our stance towards `Borrowed`, e.g. "it should never be used in a public API if `&Bound` works as well" or "used only to prevent otherwise necessary but generally superfluous reference increments". Could we maybe even make `Borrowed` unnameable for downstream crates so that it can only be accessed as a return value or via type interference on trait impls?

davidhewitt: > Could we maybe even make `Borrowed` unnameable for downstream crates so that it can only be accessed as a return value or via type interference on trait impls?

I'm certainly up for trying that, and even if we decide to make `Borrowed` more public later, it does help keep down the number of new things we're introducing to users in a single release.

I'll push one or two more commits later tonight to switch to `&Bound` here, and take `Borrowed` out of reachable paths for now. Let's see how that works out.

davidhewitt: I haven't forgotten about this; just had illness and subsequent exhaustion in the family. Might be a couple of days before I have energy to get to this still 😭

davidhewitt: Finally completed the `&Bound<PyType>` bit in #3831, so I'll rebase here.

davidhewitt:
Ok, this is finally reviewable! :tada:

LilyFoote:
I would like to review this, but it would be helpful if the conflicts were fixed first.

davidhewitt:
Ahh aha as soon as I'd got it ready the new changes to `PyErr` conflicted! I'll aim to update tonight 👍

davidhewitt:
Fixed!

src/types/typeobject.rs
```diff
@@ -11,38 +13,141 @@ pub struct PyType(PyAny);
...
+    /// - The pointer must be non-null.
+    /// - The pointer must be valid for the entire of the lifetime 'a for which the reference is used,
+    ///   as with `std::slice::from_raw_parts`.
+    #[inline]
+    pub unsafe fn from_type_ptr_bound<'a, 'py>(
+        py: Python<'py>,
+        p: &'a *mut ffi::PyTypeObject,
+    ) -> &'a Bound<'py, PyType> {
+        let object_ptr = &*(p as *const *mut ffi::PyTypeObject as *const *mut ffi::PyObject);
+        Bound::ref_from_ptr(py, object_ptr).downcast_unchecked()
```
Icxolu, line 70: Is this usage common enough, that we don't simply want to take ownership of the pointer? `& *mut _` just looks a bit odd to me, but I'm fine with it, if we want to provide a borrowing API.

src/types/typeobject.rs
```diff
@@ -11,38 +13,141 @@ pub struct PyType(PyAny);
...
 
 impl PyType {
-    /// Creates a new type object.
+    /// Deprecated form of [`PyType::new_bound`].
     #[inline]
+    #[cfg_attr(
+        not(feature = "gil-refs"),
+        deprecated(
+            since = "0.21.0",
+            note = "`new` will be replaced by `new_bound` in a future PyO3 version"
```
Icxolu, line 22: For the other types we have included the type name in the note, but I wouldn't consider that super important.

Icxolu (commented):
I added one remark, otherwise this looks good to me 🚀

src/types/typeobject.rs
davidhewitt: Ah good catch, I'll fixup, this is because this code is now so old it misses out a lot of our current conventions 👍

src/types/typeobject.rs
davidhewitt: At one point this was used in the macro implementation, but that has now been removed with the changes in #3831. This is now only used in `PyAnyMethods::get_type`, which immediately makes it owned.

So I think this is a fair concern, let's just make this owned and keep things simple until users show us that borrowing here is performance critical (and their borrow is safe).

LilyFoote (approved):

src/types/typeobject.rs
```diff
@@ -11,38 +13,144 @@ pub struct PyType(PyAny);
...
+            since = "0.21.0",
+            note = "`from_type_ptr` will be replaced by `from_type_ptr_bound` in a future PyO3 version"
+        )
+    )]
     pub unsafe fn from_type_ptr(py: Python<'_>, p: *mut ffi::PyTypeObject) -> &PyType {
-        py.from_borrowed_ptr(p as *mut ffi::PyObject)
+        Self::from_type_ptr_bound(py, p).into_gil_ref()
+    }
+
+    /// Converts the given FFI pointer into `&Bound<PyType>`, to use in safe code.
```
Icxolu, line 58:
```suggestion
    /// Converts the given FFI pointer into `Bound<PyType>`, to use in safe code.
```

src/types/typeobject.rs
```diff
@@ -11,38 +13,144 @@ pub struct PyType(PyAny);
...
+    ///
+    /// This does not take ownership of the FFI pointer's "reference". The target type
+    /// object will have its reference count increased by 1, which will be released when
+    /// the `Bound` is dropped.
```
Icxolu, line 62: This _does_ now take ownership, or did I misunderstand what you meant here?

src/types/typeobject.rs
```diff
@@ -11,38 +13,144 @@ pub struct PyType(PyAny);
...
-    /// - The pointer must be non-null.
-    /// - The pointer must be valid for the entire of the lifetime for which the reference is used.
+    ///
+    /// - The pointer must a valid non-null reference to a `PyTypeObject`.
     #[inline]
+    #[cfg_attr(
+        not(feature = "gil-refs"),
+        deprecated(
+            since = "0.21.0",
+            note = "`from_type_ptr` will be replaced by `from_type_ptr_bound` in a future PyO3 version"
```
Icxolu, line 51:
```suggestion
            note = "`PyType::from_type_ptr` will be replaced by `PyType::from_type_ptr_bound` in a future PyO3 version"
```

Here the type is also missing 🙈 (sorry, i could have mentioned that in the first round)

src/types/typeobject.rs
davidhewitt: Uff, it's assuming that the reference is a borrowed pointer and creates a new reference. I agree my wording here was not the best.

Maybe it's better if we go a little bit inconsistent here and introduce this method with the new clearer name `from_borrowed_type_ptr` (forget the bound bit).

That'd then match the `from_borrowed` methods which I think I'm about to introduce in a new PR...

src/types/typeobject.rs
davidhewitt: 🙈

src/types/typeobject.rs
davidhewitt: I pushed that as a new commit.

Icxolu (approved):

src/types/typeobject.rs
Icxolu: This reads clearer to me 👍

src/types/typeobject.rs
```diff
@@ -11,38 +13,143 @@ pub struct PyType(PyAny);
...
+
     /// Retrieves the underlying FFI pointer associated with this Python object.
     #[inline]
     pub fn as_type_ptr(&self) -> *mut ffi::PyTypeObject {
-        self.as_ptr() as *mut ffi::PyTypeObject
+        self.as_borrowed().as_type_ptr()
     }
 
-    /// Retrieves the `PyType` instance for the given FFI pointer.
+    /// Deprecated form of [`PyType::from_type_ptr_bound`].
```
LilyFoote, line 41:
```suggestion
    /// Deprecated form of [`PyType::from_borrowed_type_ptr`].
```

davidhewitt:
🎉 thanks @adamreichold @Icxolu @LilyFoote - this was a thorny one and it feels like quite the milestone to finally get this one rounded out!

src/types/typeobject.rs
```diff
@@ -11,38 +13,143 @@ pub struct PyType(PyAny);
...
+    /// The function creates a new reference from the given pointer, and returns
+    /// it as a `Bound<PyType>`.
+    ///
+    /// # Safety
+    /// - The pointer must be a valid non-null reference to a `PyTypeObject`
+    #[inline]
+    pub unsafe fn from_borrowed_type_ptr<'py>(
+        py: Python<'py>,
+        p: *mut ffi::PyTypeObject,
+    ) -> Bound<'py, PyType> {
```
LilyFoote, line 69:
```suggestion
    pub unsafe fn from_borrowed_type_ptr(
        py: Python<'_>,
        p: *mut ffi::PyTypeObject,
    ) -> Bound<'_, PyType> {
```

PR #3706, opened 2023-12-27
Add `extract_bound` method to FromPyObject

davidhewitt:
This PR proposes a way that we can support migration of `FromPyObject` off the GIL Refs API in a backwards-compatible fashion.

It does so by adding a new `extract_bound` method to `FromPyObject<'py>`, which takes `&Bound<'py, PyAny>`. Note that the lifetime `'py` in the trait becomes only the `'py` lifetime and not the lifetime for which the `Bound` smart pointer is alive. This is important for backwards-compatibility; maybe in the future we could have `FromPyObject<'a, 'py>` and the argument could be `&'a Bound<'py, PyAny>`, but let's cross that bridge another time.

Also for backwards-compatibility's sake, both `extract` and `extract_bound` have default implementations in terms of each other. This allows for users to migrate from using one set of implementations to the other at their own pace. The documentation encourages to implement just `extract_bound`, as the default implementation requires a `.into_gil_ref()` call to insert into the pool.

To avoid a huge diff here, I've pushed a second commit which just includes a couple of randomly-selected implementations which I migrate from `extract` to `extract_bound`, to give a feeling for how this works.

If we like this and merge it, I can work on one or more follow-ups which migrate the rest of our internals to `extract_bound` (which will go a long way towards getting PyO3's internals off the pool).

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -6,7 +6,7 @@ use crate::pyclass::boolean_struct::False;
 use crate::type_object::PyTypeInfo;
 use crate::types::PyTuple;
 use crate::{
-    ffi, gil, Py, PyAny, PyCell, PyClass, PyNativeType, PyObject, PyRef, PyRefMut, Python,
+    ffi, gil, Bound, Py, PyAny, PyCell, PyClass, PyNativeType, PyObject, PyRef, PyRefMut, Python,
 };
 use std::cell::Cell;
 use std::ptr::NonNull;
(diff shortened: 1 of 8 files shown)


src/conversion.rs
```diff
@@ -215,11 +215,26 @@ pub trait IntoPy<T>: Sized {
...
+        Self::extract_bound(&ob.as_borrowed())
+    }
+
+    /// Extracts `Self` from the bound smart pointer `obj`.
+    ///
+    /// Implementors are encouraged to implement this method and leave `extract` defaulted, as
+    /// this will be most compatible with PyO3's future API.
+    fn extract_bound(ob: &Bound<'source, PyAny>) -> PyResult<Self> {
+        Self::extract(ob.clone().into_gil_ref())
+    }
```
davidhewitt, line 237: The default implementation has infinite recursion if a user implements neither method. This is a slightly clunky footgun but I'm fine with it as a temporary measure; I think in 0.22 once we start pushing harder then we can just default `extract` and require an implementation of `extract_bound`.

davidhewitt:
Unless we can think of alternatives here, I'd quite like to merge this PR, as merging this unblocks us to migrate the rest of PyO3's implementations of `FromPyObject` off the GIL Refs API. I think that also is a significant chunk of where PyO3 will be using the old API. (And a lot of that I've got stacked in #3606, so I can split that out into PRs once this lands.)

src/conversion.rs
adamreichold: Does implementing this using only the defaults trigger a warning due to unconditional mutual recursion?

adamreichold (approved):
I agree that the general approach is the best we can do for 0.21. I think this should be highlighted in the migration guide though, either as part of this PR or as a task item on the tracking issue.

src/conversion.rs
davidhewitt: Sadly no, it just leads to a stack overflow at runtime. I've tried to repeat the expectation in documentation that implementors should implement `extract_bound`.

Hopefully it's a fairly obvious programming error, and at least we can get rid of the `.extract_bound()` default in 0.22 and remove this footgun again.

davidhewitt:
Thank you for all the reviews today! I'll have plenty of PRs to push once this one goes in ;)

Completely agree that this is worthy of going in the migration guide. I'll add it to this PR; I'll rebase once #3755 goes in and add some detail before merging this.

PR #3708, opened 2023-12-28
update `extract_argument` to use Bound APIs

davidhewitt:
Builds upon #3681 to move our internal argument extraction mechanisms off the GIL Pool.

Only the last commit is new, and I don't think this is worth reviewing right until #3681 is in, after which I'll rebase and tidy this a bit further. Pushing now purely because I think this should hopefully demonstrate a tantalising performance speedup which we're so close to merging! 😂


davidhewitt:
Hmm I'll investigate later why this hasn't yielded the speedup I've seen in #3606, there's likely some other part I need to merge first...

davidhewitt:
Ah, I think it's `PyDict::new_bound` which is missing, I'll open a PR for that once we've merged the feature proposed in #3707

davidhewitt:
The performance regressions with the `ordered_X` benchmarks are genuine and look like they're caused because part of the proc macros go through `&PyCell<T>` GIL Ref machinery, and the compatibility implementation of `FromPyObject::extract_bound` is hurting this.

Probably the right approach is to get this PR merged (well, needs #3784 reviewed and merged first) and also #3792. Then  I think there will be enough pieces in place to deal with updating the proc macro internals in a follow-up.

davidhewitt:
This is getting closer to being reviewable, but I think first it was worth splitting out #3801 and #3802.

src/instance.rs
```diff
@@ -522,6 +522,19 @@ impl<'a, 'py> Borrowed<'a, 'py, PyAny> {
         Self(NonNull::new_unchecked(ptr), PhantomData, py)
     }
 
+    #[inline]
+    pub(crate) fn downcast<T>(self) -> Result<Borrowed<'a, 'py, T>, DowncastError<'a, 'py>>
```
davidhewitt, line 526: Given that the argument extraction works with `Borrowed` as the primary type, this turned out to be useful to get to `Borrowed<PyString>` for the keyword extraction.

src/err/mod.rs
davidhewitt, line 1: The changes in this file were to make `Borrowed::downcast` possible.

pytests/src/pyfunctions.rs
```diff
@@ -4,62 +4,66 @@ use pyo3::types::{PyDict, PyTuple};
 #[pyfunction(signature = ())]
 fn none() {}
 
+type Any<'py> = Bound<'py, PyAny>;
+type Dict<'py> = Bound<'py, PyDict>;
+type Tuple<'py> = Bound<'py, PyTuple>;
```
davidhewitt, line 9: Clippy was warning about type complexity in the return value tuples without defining these aliases.

src/types/string.rs
```diff
@@ -357,7 +357,7 @@ impl<'py> PyStringMethods<'py> for Bound<'py, PyString> {
 impl<'a> Borrowed<'a, '_, PyString> {
     #[cfg(any(Py_3_10, not(Py_LIMITED_API)))]
     #[allow(clippy::wrong_self_convention)]
-    fn to_str(self) -> PyResult<&'a str> {
+    pub(crate) fn to_str(self) -> PyResult<&'a str> {
```
davidhewitt, line 360: Used in keyword argument extraction from borrowed keyword names.

src/types/dict.rs
```diff
@@ -517,6 +517,16 @@ impl<'py> PyDictMethods<'py> for Bound<'py, PyDict> {
...
 
+impl<'py> Bound<'py, PyDict> {
+    /// Iterates over the contents of this dictionary without incrementing reference counts.
+    ///
+    /// # Safety
+    /// It must be known that this dictionary will not be modified during iteration.
+    pub(crate) unsafe fn iter_borrowed<'a>(&'a self) -> BorrowedDictIter<'a, 'py> {
+        BorrowedDictIter::new(self)
+    }
+}
```
davidhewitt, line 528: The purpose of adding this crate-private API was that I wanted to keep keyword argument processing using `Borrowed` for both the "fastcall" and default calling conventions. As far as I am aware, argument processing can safely process a `kwargs` dict in this way because it's the only referee of the kwargs dict.

davidhewitt (commented):
Whew, I think this is finally ready to review and I'll deal with the `PyRef`/`PyCell` changes separately.

pyo3-macros-backend/src/method.rs
```diff
@@ -583,7 +583,8 @@ impl<'a> FnSpec<'a> {
                     ) -> _pyo3::PyResult<*mut _pyo3::ffi::PyObject> {
                         let function = #rust_name; // Shadow the function name to avoid #3017
                         #( #holders )*
-                        #call
+                        let result = #call;
+                        result
```
davidhewitt, line 587: These changes to return `result` on a separate line were caused by the fact that `#call` creates a borrow on `_args` and `_kwargs`, which now have `Drop` implementations due to being `Bound<PyTuple>` and `Option<Bound<PyDict>>` (if `*args`, or `**kwargs` are accepted).

So to satisfy the borrow checker we had to split the expression like this.

I think given our decision in macro code is to trend away from a single giant expression anyway, and it'll help me in #3847 where I need to do this too, this seemed a reasonable adjustment to me.

src/instance.rs
davidhewitt: Correction: this also enabled `Borrowed<T>` to be used as the holder in `impl PyFunctionArgument for &Bound`.

src/impl_/extract_argument.rs
```diff
@@ -1,10 +1,51 @@
...
+    #[inline]
+    pub fn from_borrowed<T>(borrowed: Borrowed<'py, 'py, T>) -> Self {
+        Self(borrowed.into_any())
+    }
+
+    #[inline]
+    pub fn as_borrowed(self) -> Borrowed<'py, 'py, PyAny> {
+        self.0
+    }
+}
```
davidhewitt, line 41: To be more specific to answer the docstring's future possibility of just using `Borrowed` everywhere, I think that the only step needed is to expose `Borrowed::from_ptr` family of functions as public API. Maybe I should break that into a separate PR and then I could rework this PR one more time to avoid this newtype?

LilyFoote (commented):
I spotted one nit.

src/impl_/extract_argument.rs
```diff
@@ -268,13 +308,21 @@ impl FunctionDescription {
 
         // Handle keyword arguments
         let mut varkeywords = K::Varkeywords::default();
-        if let Some(kwnames) = py.from_borrowed_ptr_or_opt::<PyTuple>(kwnames) {
+
+        // Safety: kwnames is known to be a pointer to a tuple, or null
+        let kwnames: &'py Option<Bound<'py, PyTuple>> = std::mem::transmute(&kwnames);
```
Icxolu, line 313: Maybe we want to create a `Bound::ref_from_ptr_or_opt` to avoid `std::mem::transmute`?

src/impl_/extract_argument.rs
```diff
@@ -268,13 +308,21 @@ impl FunctionDescription {
...
+        if let Some(kwnames) = kwnames.as_ref() {
             // Safety: &PyAny has the same memory layout as `*mut ffi::PyObject`
```
Icxolu, line 315: This safety comment needs updating

src/impl_/extract_argument.rs
```diff
@@ -303,17 +351,20 @@ impl FunctionDescription {
...
     where
         V: VarargsHandler<'py>,
         K: VarkeywordsHandler<'py>,
     {
-        let args = py.from_borrowed_ptr::<PyTuple>(args);
-        let kwargs: ::std::option::Option<&PyDict> = py.from_borrowed_ptr_or_opt(kwargs);
+        // Safety: Bound has the same layout as a raw pointer, and reference is known to be
+        // borrowed for 'py.
+        let args: &'py Bound<'py, PyTuple> = std::mem::transmute(&args);
+        let kwargs: &'py Option<Bound<'py, PyDict>> = std::mem::transmute(&kwargs);
```
Icxolu, line 366: Same as above

src/impl_/extract_argument.rs
Icxolu: I think we would need to expose them anyway, if we want to provide any replacement for the `Python::from_borrowed_ptr` family, right?

davidhewitt: I think the alternative would be to force users to increase reference counts and take ownership with `Bound::from_borrowed_ptr` methods instead, like we have on `Py`. There's a potentially good argument to do that for simplicity and consistency, and keep `Borrowed` deliberately minimalist.

Yet another option is to add both sets of methods and just encourage users to use `Bound` instead of `Borrowed` as much as possible.

davidhewitt: Ok, I've removed this type.

src/err/mod.rs
davidhewitt: (Removed these again.)

src/instance.rs
davidhewitt: Removed this and used `&Bound` in those places.

davidhewitt:
Thanks for the reviews! I've removed the `PyArg` newtype following #3858, which has made this implementation a fair bit nicer on the macro side. I'll try to find time to tidy the rest up later.

src/impl_/extract_argument.rs
davidhewitt: In the end I adjusted the `iter_borrowed()` implementations to allow calling them from a `Borrowed` instance, and then I was able to use `Borrowed::from_ptr` methods in this code here. I think it's fine to use `Borrowed` instead of `ref_from_ptr` here because we can control that pesky borrowed lifetime `'a` without worrying about user code, unlike the macros.

src/impl_/extract_argument.rs
```diff
@@ -374,11 +406,21 @@ impl FunctionDescription {
...
-            // error if this function doesn't accept **kwargs). Rust source is always UTF-8
-            // and so all argument names in `#[pyfunction]` signature must be UTF-8.
-            if let Ok(kwarg_name) = kwarg_name_py.downcast::<PyString>()?.to_str() {
+            // Safety: All keyword arguments should be UTF-8 strings, but if it's not, `.to_str()`
+            // will return an error anyway.
+            #[cfg(any(Py_3_10, not(Py_LIMITED_API)))]
+            let kwarg_name =
+                unsafe { kwarg_name_py.downcast_unchecked::<crate::types::PyString>() }.to_str();
+
+            #[cfg(not(any(Py_3_10, not(Py_LIMITED_API))))]
```
LilyFoote, line 415: I think we can avoid the double-negative:

```suggestion
            #[cfg(all(not(Py_3_10), Py_LIMITED_API))]
```

src/impl_/extract_argument.rs
```diff
@@ -374,11 +406,21 @@ impl FunctionDescription {
...
+            let kwarg_name = kwarg_name_py.extract::<crate::pybacked::PyBackedStr>();
+
+            if let Ok(kwarg_name_owned) = kwarg_name {
+                #[cfg(any(Py_3_10, not(Py_LIMITED_API)))]
+                let kwarg_name = kwarg_name_owned;
+                #[cfg(not(any(Py_3_10, not(Py_LIMITED_API))))]
```
LilyFoote, line 421:
```suggestion
                #[cfg(all(not(Py_3_10), Py_LIMITED_API))]
```

src/impl_/extract_argument.rs
```diff
@@ -410,6 +452,11 @@ impl FunctionDescription {
         }
 
         if !positional_only_keyword_arguments.is_empty() {
+            #[cfg(not(any(Py_3_10, not(Py_LIMITED_API))))]
```
LilyFoote, line 455:
```suggestion
            #[cfg(all(not(Py_3_10), Py_LIMITED_API))]
```

src/types/dict.rs
```diff
@@ -662,6 +672,60 @@ impl<'py> IntoIterator for Bound<'py, PyDict> {
...
+        fn next(&mut self) -> Option<Self::Item> {
+            let mut key: *mut ffi::PyObject = std::ptr::null_mut();
+            let mut value: *mut ffi::PyObject = std::ptr::null_mut();
+
+            if unsafe { ffi::PyDict_Next(self.dict.as_ptr(), &mut self.ppos, &mut key, &mut value) }
+                != 0
+            {
+                let py = self.dict.py();
+                self.len -= 1;
+                Some(unsafe { (key.assume_borrowed(py), value.assume_borrowed(py)) })
```
LilyFoote, line 700: Should we add safety commends for these `unsafe` blocks?

LilyFoote (approved):
Looking good! Just a few nitpicks.

pyo3-macros-backend/src/method.rs
```diff
@@ -196,10 +196,11 @@ impl SelfType {
                 holders.push(quote_spanned! { *span =>
                     #[allow(clippy::let_unit_value)]
                     let mut #holder = _pyo3::impl_::extract_argument::FunctionArgumentHolder::INIT;
+                    let mut #slf = _pyo3::Borrowed::from_ptr(#py, #slf);
```
Icxolu, line 199: Would `BoundRef` work here? Just to be a extra cautious about the `Borrowed` lifetime.

pyo3-macros-backend/src/params.rs
```diff
@@ -44,8 +44,8 @@ pub fn impl_arg_params(
             .collect::<Result<_>>()?;
         return Ok((
             quote! {
-                let _args = py.from_borrowed_ptr::<_pyo3::types::PyTuple>(_args);
-                let _kwargs: ::std::option::Option<&_pyo3::types::PyDict> = py.from_borrowed_ptr_or_opt(_kwargs);
+                let _args = _pyo3::Borrowed::from_ptr(py, _args);
+                let _kwargs = _pyo3::Borrowed::from_ptr_or_opt(py, _kwargs);
```
Icxolu, line 48: Same here maybe?

pyo3-macros-backend/src/pymethod.rs
```diff
@@ -997,7 +990,7 @@ fn extract_object(
     });
     extract_error_mode.handle_error(quote! {
         _pyo3::impl_::extract_argument::extract_argument(
-            #source,
+            &_pyo3::Borrowed::from_ptr(py, #source_ptr),
```
Icxolu, line 993: And here maybe?

Icxolu (approved):
I spotted a few `Borrowed::from_ptr` calls in proc macro generated code. It might be nice to replace them with `BoundRef` if possible, for a bit of extra safety. Otherwise LGTM!

src/types/dict.rs
davidhewitt: Good idea, done 👍

pyo3-macros-backend/src/method.rs
davidhewitt: Thanks, yes using `BoundRef` instead of `Borrowed` is a much safer idea for macro code. I think I need to add `Deref<Target = Bound<T>> for BoundRef<T>` to make this work, but I see no problem with that impl.

LilyFoote (approved):

Icxolu (approved):

pyo3-macros-backend/src/method.rs
Icxolu: Nice, I don't see a problem with implementing `Deref` either, since `BoundRef` is really just `&Bound` 👍

davidhewitt:
Thanks both, think we're finally there on this one!

MarshalX:
> Hmm I'll investigate later why this hasn't yielded the speedup I've seen in #3606, there's likely some other part I need to merge first...

What is the latest result btw?

davidhewitt:
https://github.com/PyO3/pyo3/pull/3708#issuecomment-1871521345, but I think it's quite likely that parts of the speedup already got merged now, given how many times this got split up.

I will probably analyze again some time and see if there's any further tricks to apply here, but for now I think this is good.

PR #3710, opened 2023-12-29
ci: updates for Rust 1.75

davidhewitt:
Updates UI test output for Rust 1.75. There's a slight regression in one of the compiler messages related to `#[pyclass]` requiring Send; I think for now I'm inclined to leave it if we're revisiting that code in the near future to require `Sync` on classes for nogil anyway.

While I was at it, I pushed a tiny second commit which deletes a bit of unused code and guide text which was left behind after deletion of some old features.

Assuming green CI, I'll proceed to merge this as there's nothing particularly worth discussing here, and then I can rebase everything else.


adamreichold (approved):

davidhewitt:
Turns out the `abi3` inheritance UI test also now emits slightly more diagnostics; again that's something I'd like to improve in the future so not too worried about that either.

PR #3711, opened 2023-12-29
add `call_bound` and `call_method_bound`

davidhewitt:
Adds `Py::call_bound` and `Py::call_method_bound`, which take `kwargs: Option<&Bound<'_, PyDict>>`.

For `&PyAny`, I decided not to add these new variants as instead the correct change should be to switch to `Bound<'_, PyAny>`.

Similarly I changed the argument in `PyAnyMethods::call` and `PyAnyMethods::call_method` without introducing new variants.

The other functions like `call0` don't need changing as they don't take an `Option<&PyDict>` argument.

### DIFF
--- a/guide/src/python_from_rust.md
+++ b/guide/src/python_from_rust.md
@@ -10,13 +10,13 @@ Any Python-native object reference (such as `&PyAny`, `&PyList`, or `&PyCell<MyC
 
 PyO3 offers two APIs to make function calls:
 
-* [`call`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.call) - call any callable Python object.
-* [`call_method`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.call_method) - call a method on the Python object.
+* [`call`]({{#PYO3_DOCS_URL}}/pyo3/prelude/trait.PyAnyMethods#tymethod.call) - call any callable Python object.
+* [`call_method`]({{#PYO3_DOCS_URL}}/pyo3/prelude/trait.PyAnyMethods#tymethod.call_method) - call a method on the Python object.
 
 Both of these APIs take `args` and `kwargs` arguments (for positional and keyword arguments respectively). There are variants for less complex calls:
 
-* [`call1`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.call1) and [`call_method1`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.call_method1) to call only with positional `args`.
-* [`call0`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.call0) and [`call_method0`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.call_method0) to call with no arguments.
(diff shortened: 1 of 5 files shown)


src/types/any.rs
```diff
@@ -2184,6 +2184,7 @@ impl<'py> PyAnyMethods<'py> for Bound<'py, PyAny> {
 }
 
 #[cfg(test)]
+#[cfg_attr(not(feature = "gil-refs"), allow(deprecated))]
```
davidhewitt, line 2187: I felt like the most practical option w.r.t keeping coverage up and test churn down during the initial implementation phase was to just add this conditional allow to tests at quite a wide level for now.

Then on the way from 0.21 to 0.22 when we're adding the full feature gate we can tidy up tests and make these allows narrower.

davidhewitt:
Uff will fixup doctests later, ideally this evening.

davidhewitt:
(This one is ready for review as far as I'm concerned; the doctests look to be passing now :tada:)

guide/src/python_from_rust.md
```diff
@@ -95,22 +95,28 @@ fn main() -> PyResult<()> {
...
+        fun.call_bound(py, (), Some(&kwargs.into_py_dict(py).as_borrowed()))?;
 
         Ok(())
     })
 }
 ```
 
+<div class="warning">
+
+During PyO3's [migration from "GIL Refs" to the `Bound<T>` smart pointer](./migration.md#migrating-from-the-gil-refs-api-to-boundt), `call` is temporarily named `call_bound`.
```
adamreichold, line 116: Maybe this could be extend to stress that only the inherent methods on `PyAny` have these names whereas those accessible via `PyAnyMethods` start with the short names and will stay that way, i.e. I can already use the short names in a fully bound world, `call_bound` et al are necessary only for mixed code bases.

src/instance.rs
```diff
@@ -1016,14 +1016,30 @@ impl<T> Py<T> {
...
     /// Calls the object.
     ///
     /// This is equivalent to the Python expression `self(*args, **kwargs)`.
-    pub fn call(
+    pub fn call_bound(
         &self,
         py: Python<'_>,
         args: impl IntoPy<Py<PyTuple>>,
-        kwargs: Option<&PyDict>,
+        kwargs: Option<&Bound<'_, PyDict>>,
```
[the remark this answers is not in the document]
davidhewitt: It's a reasonable question. In other APIs e.g. `PyIterator::from_bound_object` we just added in #3702 I've been choosing to use `&Bound<'py, PyAny>` because there's been no need to transfer ownership, and `Option<&Bound<'_, PyDict>>` seemed to be the logical parallel. Here we also don't need to transfer ownership of a python reference, so taking `&Bound<'_, PyDict>` avoids a needless reference count increase / decrease cycle.

If we're concerned by the double pointer indirection we _could_ take `Option<Borrowed<'_, '_, PyDict>>` instead, though so far I've tried to avoid making the user to deal with `Borrowed` as anything other than a return value from specific APIs where returning a `Bound` would force the reference count increase.

guide/src/python_from_rust.md
davidhewitt: So to be clear in this PR I don't add `call_bound` as a `PyAny` inherent method, I felt that there it's good enough for the user to do `any.as_borrowed().call(...)`. I just added `_bound` variants to `Py<T>`.

I'm open to adding `PyAny::call_bound` as a consistency piece, if you think it's desirable.

And I can extend the doc either way 👍

davidhewitt: Ok great, updated the doc 👍

adamreichold (approved):

PR #3712, opened 2023-12-29
add PyAnyMethods for binary operators

### LINKED ISSUE #3709: Expose binary operations (*, &, etc.)
These are available in the CPython API via `PyNumber_And`, `PyNumber_Multiply`, etc.

I'm happy to send PRs for these. Though a) it probably makes sense to do this after #3684, b) is there a better way to do this than dozens of new methods on `PyAny`? (https://docs.python.org/3/c-api/number.html)

alex:
also pow

fixes #3709

### DIFF
--- a/src/types/any.rs
+++ b/src/types/any.rs
@@ -1208,6 +1208,58 @@ pub trait PyAnyMethods<'py> {
     where
         O: ToPyObject;
 
+    /// Computes `self + other`.
+    fn add<O>(&self, other: O) -> PyResult<Bound<'py, PyAny>>
+    where
+        O: ToPyObject;
+
+    /// Computes `self - other`.
+    fn sub<O>(&self, other: O) -> PyResult<Bound<'py, PyAny>>
+    where
+        O: ToPyObject;
+
+    /// Computes `self * other`.
+    fn mul<O>(&self, other: O) -> PyResult<Bound<'py, PyAny>>
+    where
+        O: ToPyObject;
+
+    /// Computes `self / other`.
+    fn div<O>(&self, other: O) -> PyResult<Bound<'py, PyAny>>
+    where
+        O: ToPyObject;
+
+    /// Computes `self << other`.
+    fn lshift<O>(&self, other: O) -> PyResult<Bound<'py, PyAny>>
+    where
+        O: ToPyObject;
+
+    /// Computes `self >> other`.
+    fn rshift<O>(&self, other: O) -> PyResult<Bound<'py, PyAny>>
+    where
+        O: ToPyObject;
+
+    /// Computes `self ** other % modulus` (`pow(self, other, modulus)`).
+    /// `py.None()` may be passed for the `modulus`.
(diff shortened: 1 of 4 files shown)


tests/test_arithmetics.rs
```diff
@@ -233,6 +233,13 @@ fn binary_arithmetic() {
         py_expect_exception!(py, c, "1 ** c", PyTypeError);
 
         py_run!(py, c, "assert pow(c, 1, 100) == 'BA ** 1 (mod: Some(100))'");
+
+        let c: Bound<'_, PyAny> = c.extract().unwrap();
+        assert_eq!(c.add(&c).unwrap().extract::<&str>().unwrap(), "BA + BA");
+        assert_eq!(
+            c.pow(&c, py.None()).unwrap().extract::<&str>().unwrap(),
+            "BA ** BA (mod: None)"
+        );
```
davidhewitt, line 242: I wonder if we can write a helper to make this sort of check easier, e.g.

```suggestion
        assert_py_eq!(c.add(&c).unwrap(), "BA + BA");
```

... which I guess calls `.eq("BA + BA").unwrap()`.

It might also be nice to list out all these methods again explicitly, even if it's a bit tedious.

src/types/any.rs
```diff
@@ -1855,6 +1907,38 @@ impl<'py> PyAnyMethods<'py> for Bound<'py, PyAny> {
             .and_then(|any| any.is_truthy())
     }
 
+    implement_binop!(add, PyNumber_Add, "+");
+    implement_binop!(sub, PyNumber_Subtract, "-");
+    implement_binop!(mul, PyNumber_Multiply, "*");
+    implement_binop!(div, PyNumber_TrueDivide, "/");
+    implement_binop!(bitand, PyNumber_And, "&");
```
davidhewitt, line 1914: We seem to be missing bitor, xor, modulo, matrix multiplication etc. Was there a reason you limited to just these ones rather than the full list?

davidhewitt (commented):
Thanks! Some thoughts...

tests/test_arithmetics.rs
alex: Good idea.

src/types/any.rs
alex: There was a reason (unclear if it was a _good_ reason): I did the basic arithmetic ops + the set of things I needed :-)

davidhewitt: Heh, fair enough. Maybe let's at least in this PR list them all down and agree what makes sense to include to have a consistent set if we don't do everything? For example, I think at least bitor makes sense if we have bitand 😂

davidhewitt (approved):
This looks solid to me! 🚀

alex:
Is that test known to be flakey? It doesn't look obviously related to this PR.

adamreichold:
> Is that test known to be flakey? It doesn't look obviously related to this PR.

Not really and from looking at the source and running it locally, I cannot reproduce the failure on `main`. I did try to restart the CI job to see whether the failure is deterministic on this branch.

alex:
This failure was definitely a network flake :-(

PR #3714, opened 2023-12-30
docs: Document how to represent custom Python-native classes

clayote:
I had trouble figuring out how to make my own structs representing my own Python classes that weren't implemented in Rust, so I documented how to do it.


clayote:
I would recommend squashing this...

clayote:
My documented code isn't very reliable and I think this may not be intended as part of the public API anyhow.

davidhewitt:
Thanks for sharing this, and sorry that it caused you some pain. I think at the moment `PyTypeInfo` is in this blurry space where we don't really expect users to implement it, even if it is possible to. I'd really like to improve this in the future, but definitely not until we've got 0.21.0 complete and are ready to think about simplifying PyO3 a bit.

PR #3721, opened 2024-01-02
Use a definite version specification when depending on pyo3-build-config.

### LINKED ISSUE #3719: Possible breaking change in `pyo3-build-config` from `0.20.0` `0.20.1`?
I haven't looked into this deeply, and possibly dependabot has done something screwy with the lockfile — but this patch upgrade seems to raise an error in https://github.com/PRQL/prql/pull/4033

The error is at https://github.com/PRQL/prql/actions/runs/7378972354/job/20074705518?pr=4033

adamreichold:
We already do this for other internal `pyo3-*` dependencies and it seems prudent to apply this to `pyo3-build-config` as well.

Closes #3719

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ widestring = "0.5.1"
 futures = "0.3.28"
 
 [build-dependencies]
-pyo3-build-config = { path = "pyo3-build-config", version = "0.21.0-dev", features = ["resolve-config"] }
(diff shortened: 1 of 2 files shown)


davidhewitt (approved):
Agreed, shame it's too late to apply this to 0.20.0. I'm tempted to backport for 0.20.2 anyway.

PR #3722, opened 2024-01-02
Fix missing feature flags in implementation of Either conversion.

adamreichold:


adamreichold:
Since we'd need to do a 0.20.2 release to fix our doc builds, we might want to backport #3721 as well.

adamreichold:
Another thing we want to add here is a CI step (at least for full builds) matching docs.rs, i.e. using a nightly rustdoc and the features and cfgs defined via `package.metadata.docs.rs`.

.github/workflows/ci.yml
```diff
@@ -313,6 +313,22 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: cargo-careful
+        continue-on-error: true
+      - uses: dtolnay/rust-toolchain@nightly
+        with:
+          components: rust-src
+      - run: cargo rustdoc --lib --no-default-features --features "macros num-bigint num-complex hashbrown serde multiple-pymethods indexmap eyre either chrono rust_decimal" -Zunstable-options --config "build.rustdocflags=[\"--cfg\", \"docsrs\"]"
```
davidhewitt, line 330: It looks like the difference between this and `full` is that it doesn't include `experimental-inspect` but does include `multiple-pymethods`.

To help keep things consistent should we take `experimental-inspect` out of `full` (and manually add it back in to relevant CI jobs) and then make this 

```suggestion
      - run: cargo rustdoc --lib --no-default-features --features full -Zunstable-options --config "build.rustdocflags=[\"--cfg\", \"docsrs\"]"
```

(Or `--features "full multiple-pymethods"`, but I think `multiple-pymethods` has no documentation differences, just internal machinery.)

davidhewitt (commented):
Thanks for catching this, I agree we should do a patch release to fixup docs, and a CI job is a very good idea.

.github/workflows/ci.yml
davidhewitt: (Or maybe we should be documenting `experimental-inspect` also; it might help people notice the feature and motivate someone to come along and help finish it off.)

adamreichold: I think documenting `experimental-inspect` would be preferable. Or rather, I think the only reason to not document a feature is if it breaks the docs.rs build or we do not want people to use it, neither of which appears to be the case here.

adamreichold: Added a commit unifying the docs.rs and full builds.

davidhewitt (approved):
Perfect, thanks!

utkarshgupta137:
Still broken on 0.20.2: https://github.com/PyO3/pyo3/blob/main/src/conversions/either.rs#L97

PR #3724, opened 2024-01-02
re-add emit_pyo3_cfgs for pyo3 0.20.0 compatibility

### LINKED ISSUE #3719: Possible breaking change in `pyo3-build-config` from `0.20.0` `0.20.1`?

davidhewitt:
I tested locally that this makes pyo3 0.20.0 build against a newer pyo3-build-config. So I think if we include this in 0.20.2 then we should resolve #3719 for users doing stuff with `-Z direct-minimal-versions` etc.

### DIFF
--- a/pyo3-build-config/src/impl_.rs
+++ b/pyo3-build-config/src/impl_.rs
@@ -154,6 +154,13 @@ pub struct InterpreterConfig {
 
 impl InterpreterConfig {
     #[doc(hidden)]
+    pub fn emit_pyo3_cfgs(&self) {
+        for cfg in self.build_script_outputs() {
+            println!("{}", cfg);
+        }
+    }
+
+    #[doc(hidden)]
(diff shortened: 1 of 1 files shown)


adamreichold:
How do we get rid of this again eventually? Bumping to 0.21.0?

davidhewitt:
Yes, I think this is only needed to backport a fix, because 0.20.0 `pyo3` should never try to build against a 0.21 `pyo3-build-config` (and similar for pyo3 0.21 vs older pyo3-build-config), unless the user is deliberately patching deps in an a typical way (and they should not be surprised by breakage then).

adamreichold (approved):

PR #3725, opened 2024-01-03
release: 0.20.2

davidhewitt:
This cherry-picks #3721 onto the `release-0.20` branch and then also bumps the version number to 0.20.2 so we can push this as a patch release.

If I hear no concerns I'll put this live sometime tomorrow so we can get the docs build fixed asap.


davidhewitt:
Pushed!

PR #3730, opened 2024-01-08
Conversion between chrono_tz::Tz and zoneinfo.ZoneInfo

Tpt:


Tpt:
CI fails because of a more aggressive `dead_code` lint

Cargo.toml
```diff
@@ -35,6 +35,7 @@ inventory = { version = "0.3.0", optional = true }
 # crate integrations that can be added using the eponymous features
 anyhow = { version = "1.0", optional = true }
 chrono = { version = "0.4.25", default-features = false, optional = true }
+chrono-tz = { version = ">= 0.6, < 0.9", default-features = false, optional = true }
```
Tpt, line 38: This is to make sure we are compatible with MSRV

PR #3734, opened 2024-01-10
ffi: Add definition for PyType_GetModuleByDef

jadedpasta:
This API is available starting in 3.11. It is not part of the Stable ABI.

PyType_GetModuleByDef searches the MRO of the given type for a module matching the given module spec. It can be useful for users use that `pyo3_ffi` directly and want to support multiple interpreters/per interpreter GIL. It is useful to obtain access to the module state from special methods like `tp_new` that can't use the METH_METHOD calling convention.

I am not sure how to check if this API is available on PyPy. If someone could give me a pointer to where to check, that would be great!

### DIFF
--- a/newsfragments/3734.added.md
+++ b/newsfragments/3734.added.md
@@ -1,0 +1,1 @@
+Add definition for `PyType_GetModuleByDef` to `pyo3_ffi`.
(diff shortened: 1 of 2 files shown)


davidhewitt (commented):
Thanks for the PR! One tiny adjustment regarding PyPy, and then this should be good to merge. Also please squash the commits, sadly with GitHub Merge Queue we can't do that automatically.

davidhewitt (approved):
Thanks! 👍

PR #3736, opened 2024-01-11
Adds conversion between SystemTime and datetime

Tpt:


src/conversions/std/time.rs
```diff
@@ -0,0 +1,189 @@
+//! Conversions here do not rely on the floating point timestamp of the timestamp/fromtimestamp APIs
+//! to avoid loosing precision but goes through the timedelta/std::time::Duration types by taking for
```
davidhewitt, line 2:
```suggestion
//! to avoid losing precision but goes through the timedelta/std::time::Duration types by taking for
```

src/conversions/std/time.rs
davidhewitt, line 1: It might be nice to move the `duration` conversions from `std/duration.rs` to this file, given it's `std::time::Duration` and this is `std::time::SystemTime`.

src/conversions/std/time.rs
```diff
@@ -0,0 +1,189 @@
+//! Conversions here do not rely on the floating point timestamp of the timestamp/fromtimestamp APIs
```
davidhewitt, line 1: I wonder, is it possible there exists a fast-path for cases where the timestamp doesn't fall into a range where the loss of precision is a potential problem? I think Python datetimes only carry microsecond accuracy, I'd expect there is a range where `f64` can express this without error.

davidhewitt (commented):
Sorry for the slow review on my side here. Overall this makes sense and is well implemented, thanks!

My main thing I wonder is about the timestamp accuracy. If there's a range where the floating-point APIs are accurate enough, we might want to consider using that, as it's probably faster than going through an intermediate `timedelta`.

Tpt:
> My main thing I wonder is about the timestamp accuracy. If there's a range where the floating-point APIs are accurate enough, we might want to consider using that, as it's probably faster than going through an intermediate `timedelta`.

That's a great point. Indeed for Python -> Rust getting a flowing point timestamp (with a Python function call there is no function in the C API for that) might work better. For the Rust -> Python code the `fromtimestamp` implementation in C Python relies on the C `localtime` and `gmtime` functions so, they might get restricted to up to 2038 on some platforms. I did not felt like trying to know what the bound is and preferred to use a safer fallback. Would you be ok if I leave the implementation as it is with a TODO to investigate using timestamps directly for a possible speedup?

src/conversions/std/time.rs
Tpt: Done

src/conversions/std/time.rs
Tpt: Replied in the main discussion.

src/conversions/std/time.rs
Tpt: Done

davidhewitt (approved):
Looks great to me, sorry for the slow review cycles on this one.

> Would you be ok if I leave the implementation as it is with a TODO to investigate using timestamps directly for a possible speedup?

Yes, I think that makes sense; given the edge cases around the use of timestamps I think it makes sense that we get this base known-good implementation in, and can review the performance enhancement in its own diff.

PR #3738, opened 2024-01-12
fix some nightly lints 2024-01-12

davidhewitt:
Some lints are firing on latest nightly versions; hopefully this fixes them!

### DIFF
--- a/pyo3-macros-backend/src/utils.rs
+++ b/pyo3-macros-backend/src/utils.rs
@@ -146,9 +146,10 @@ pub fn unwrap_ty_group(mut ty: &syn::Type) -> &syn::Type {
 
 /// Extract the path to the pyo3 crate, or use the default (`::pyo3`).
 pub(crate) fn get_pyo3_crate(attr: &Option<CrateAttribute>) -> syn::Path {
-    attr.as_ref()
-        .map(|p| p.value.0.clone())
-        .unwrap_or_else(|| syn::parse_str("::pyo3").unwrap())
+    match attr {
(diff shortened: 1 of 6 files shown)


pyo3-macros-backend/src/utils.rs
```diff
@@ -146,9 +146,10 @@ pub fn unwrap_ty_group(mut ty: &syn::Type) -> &syn::Type {
 
 /// Extract the path to the pyo3 crate, or use the default (`::pyo3`).
 pub(crate) fn get_pyo3_crate(attr: &Option<CrateAttribute>) -> syn::Path {
-    attr.as_ref()
-        .map(|p| p.value.0.clone())
-        .unwrap_or_else(|| syn::parse_str("::pyo3").unwrap())
+    match attr {
+        Some(attr) => attr.value.0.clone(),
```
davidhewitt, line 150: Here nightly clippy didn't like the `attr.as_ref().map(|attr| attr.value.0.clone())` and wanted `attr.clone()`, but I don't agree with that, because it's cloning more data than necessary.

davidhewitt:
Will just set this to merge and hope it goes in.

pyo3-macros-backend/src/utils.rs
alex: The new lint is way overly aggressive: https://github.com/rust-lang/rust-clippy/issues/12135

davidhewitt: Agreed, it seems a bit broken. Thanks for filing that upstream 👍

PR #3739, opened 2024-01-12
fix size of pypy private fields in Py_buffer definition

davidhewitt:
Fixes the latest PyPy ffi-check failures; it looks like the size of some private fields changed on PyPy 7.3.14. As this is technically an ABI break in the PyPy layer, I think there's not much point worrying about it (anyone building wheels against this PyPy version will be broken with older versions). I've just updated this and will skip a changelog entry.


davidhewitt:
Wow GitHub seems flaky today...

Issue #108, opened 2018-02-09
Generated extension functions are weakly typed


Vlad-Shcherbina:
Consider the README example:
```rust
[pyfn(m, "sum_as_string")]
fn sum_as_string_py(a:i64, b:i64) -> PyResult<String>
```

Here is what happens when this function is called with non-int arguments:

```python
>>> from rust2py import sum_as_string
>>> sum_as_string(1, 2)  # good
'3'

>>> sum_as_string(10**100, 2)  # good
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OverflowError: int too big to convert

>>> sum_as_string([], [])  # good
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: int() argument must be a string, a bytes-like object or a number, not 'list'

>>> sum_as_string('1', 2)  # bad
'3'

>>> sum_as_string(1.9, 2.9)  # bad
'3'
```

Generally Python functions that expect int argument raise TypeError when fed strings or floats. This helps to catch some programming errors.
```python
>>> range('42')  # good
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'str' object cannot be interpreted as an integer

>>> range(42.9)  # good
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'float' object cannot be interpreted as an integer
```

I think it would be more consistent if generated extension functions behaved the same way.

fafhrd91:
thanks! this is definitely bug.

Vlad-Shcherbina:
So it seems relevant that I'm running this example on Windows (target_pointer_width="64", target_os="windows"), and that the argument type is `i64` and not, say, `i32`.

The problem appears to be here: https://github.com/PyO3/pyo3/blob/63a2066a39a6389f24ed1d5acda322425ade0937/src/objects/num3.rs#L121-L130

`PyLong_Check(x)` is roughly equivalent to `isinstance(x, int)`, and [`PyNumber_Long(x)`](https://github.com/python/cpython/blob/8ff53564730f7ba503f5ad94418c309c48e8516d/Include/abstract.h#L728) is analagous to `int(x)`, hence the frivolous conversion.

This code was there since it was written in https://github.com/dgrunwald/rust-cpython/commit/55ea6cc85b53566494ae550020b3dcea03de1fc0#diff-2eab662fc2799e2bdab71b4844e11090R155

[...]

PR #3742, opened 2024-01-14
improve performance of successful int extract by ~30% by avoiding calls to `__index__` where redundant

samuelcolvin:
Results from benchmarks:

```
extract_int_extract_success
                        time:   [5.0137 ns 5.0166 ns 5.0195 ns]
                        change: [-33.466% -32.879% -32.348%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  6 (6.00%) low mild
  6 (6.00%) high mild
  1 (1.00%) high severe

extract_int_extract_fail
                        time:   [190.90 ns 191.15 ns 191.41 ns]
                        change: [+0.4500% +0.9483% +1.3921%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

🚀

### DIFF
--- a/src/conversions/std/num.rs
+++ b/src/conversions/std/num.rs
@@ -44,8 +44,39 @@ macro_rules! int_fits_larger_int {
     };
 }
 
+macro_rules! extract_int {
+    ($obj:ident, $error_val:expr, $pylong_as:expr) => {
+        extract_int!($obj, $error_val, $pylong_as, false)
+    };
+
+    ($obj:ident, $error_val:expr, $pylong_as:expr, $force_index_call: literal) => {
+        // In python 3.8+ `PyLong_AsLong` and friends takes care of calling `PyNumber_Index`,
+        // however 3.8 & 3.9 do lossy conversion of floats, hence we only use the
+        // simplest logic for 3.10+ where that was fixed - python/cpython#82180.
+        // `PyLong_AsUnsignedLongLong` does not call `PyNumber_Index`, hence the `force_index_call` argument
+        // See https://github.com/PyO3/pyo3/pull/3742 for detials
+        if cfg!(Py_3_10) && !$force_index_call {
+            err_if_invalid_value($obj.py(), $error_val, unsafe { $pylong_as($obj.as_ptr()) })
+        } else if let Ok(long) = $obj.downcast::<crate::types::PyLong>() {
+            // fast path - checking for subclass of `int` just checks a bit in the type $object
+            err_if_invalid_value($obj.py(), $error_val, unsafe { $pylong_as(long.as_ptr()) })
+        } else {
+            unsafe {
+                let num = ffi::PyNumber_Index($obj.as_ptr());
+                if num.is_null() {
+                    Err(PyErr::fetch($obj.py()))
+                } else {
+                    let result = err_if_invalid_value($obj.py(), $error_val, $pylong_as(num));
+                    ffi::Py_DECREF(num);
+                    result
+                }
+            }
+        }
+    };
+}
+
 macro_rules! int_convert_u64_or_i64 {
-    ($rust_type:ty, $pylong_from_ll_or_ull:expr, $pylong_as_ll_or_ull:expr) => {
+    ($rust_type:ty, $pylong_from_ll_or_ull:expr, $pylong_as_ll_or_ull:expr, $force_index_call:literal) => {
         impl ToPyObject for $rust_type {
             #[inline]
             fn to_object(&self, py: Python<'_>) -> PyObject {
@@ -64,18 +95,8 @@ macro_rules! int_convert_u64_or_i64 {
             }
         }
         impl<'source> FromPyObject<'source> for $rust_type {
-            fn extract(ob: &'source PyAny) -> PyResult<$rust_type> {
-                let ptr = ob.as_ptr();
-                unsafe {
-                    let num = ffi::PyNumber_Index(ptr);
-                    if num.is_null() {
-                        Err(PyErr::fetch(ob.py()))
-                    } else {
-                        let result = err_if_invalid_value(ob.py(), !0, $pylong_as_ll_or_ull(num));
-                        ffi::Py_DECREF(num);
-                        result
-                    }
-                }
+            fn extract(obj: &'source PyAny) -> PyResult<$rust_type> {
+                extract_int!(obj, !0, $pylong_as_ll_or_ull, $force_index_call)
             }
 
             #[cfg(feature = "experimental-inspect")]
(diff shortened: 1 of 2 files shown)


src/conversions/std/num.rs
```diff
@@ -106,15 +107,19 @@ macro_rules! int_fits_c_long {
...
-                let val = unsafe {
-                    let num = ffi::PyNumber_Index(ptr);
-                    if num.is_null() {
-                        Err(PyErr::fetch(obj.py()))
-                    } else {
-                        let val = err_if_invalid_value(obj.py(), -1, ffi::PyLong_AsLong(num));
-                        ffi::Py_DECREF(num);
-                        val
+                // fast path - checking for subclass of `int` just checks a bit in the type object
+                let val = if let Ok(long) = obj.downcast::<PyLong>() {
```
alex, line 111: Should this be `downcast_exact`? Can an `int` subclass override the behavior?

samuelcolvin: From [the docs](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong)

> Return a `C long` representation of `obj`. If `obj` is not an instance of [`PyLongObject`](https://docs.python.org/3/c-api/long.html#c.PyLongObject), first call its [`__index__()`](https://docs.python.org/3/reference/datamodel.html#object.__index__) method (if present) to convert it to a [`PyLongObject`](https://docs.python.org/3/c-api/long.html#c.PyLongObject).

So I would think any subclass of `int` is okay.

I could also use `is_exact_instance_of` which I tried and has the exact same performance as this.

adamreichold: To me this reads like we could always avoid calling `PyNumber_Index` because `PyLong_AsLong` will do that anyway if necessary?

adamreichold: Which does indeed seem to be the case, i.e.

https://github.com/python/cpython/blob/77b45fa6d0b8c0c14657b5117b21a3f3f2ce97d8/Objects/longobject.c#L480

So we really only need the fast path?

src/conversions/std/num.rs
```diff
@@ -106,15 +107,19 @@ macro_rules! int_fits_c_long {
...
+                    err_if_invalid_value(obj.py(), -1, unsafe { ffi::PyLong_AsLong(long.as_ptr()) })
```
samuelcolvin, line 112: I admit that I don't really understand why we're using `PyLong_AsLong`, not `PyLong_AsLongLong` here? But I kept it as-is.

adamreichold: Note the name of the macro this is part of: `int_fits_c_long`. Types which could benefit from a larger representation use the other macros defined below.

src/conversions/std/num.rs
samuelcolvin: That matches my tests - e.g. nothing in pydantic-core's tests fails if you remove the slow path.

I assume the reason the code currently has the`PyNumber_Index` path was that the docs sentence is ambiguous - either meaning:
* "If obj is not an instance of `PyLongObject`, **you should** first call its `__index__()`..."
* or, "If obj is not an instance of `PyLongObject`, **this method will** first call its `__index__()`..." - which is I think what they meant

Should I remove the slow path from here, `BigInt` and anywhere else I find it?

src/conversions/std/num.rs
samuelcolvin: Makes sense, I got confused as wikipedia says `c long` is 32bits, but I think that's mostly not the case.

alex: On most platforms `long` is the "native" width (so 64-bit on 64-bit); on Windows it's always 32-bit.

src/conversions/std/num.rs
adamreichold: Personally, I'd say yes. We probably want to do a full CI build though to see if the `PyLong_AsLong` was implemented differently in older versions of CPython, i.e. we should as the `CI-build-full` label after the reduce PR is green.

samuelcolvin: makes sense, it looks like even [in 3.8](https://github.com/python/cpython/blob/v3.8.18/Objects/longobject.c#L496) there was similar logic albeit with a different name.

davidhewitt:
Nice! It looks like this code might have originated from discussion in #108. I think that logic is likely extremely dated and we're now right to simplify here.

davidhewitt:
https://github.com/python/cpython/issues/80229 looks relevant,  which was in 3.8 so we may need to keep a different path for 3.7.

samuelcolvin:
Logic changed for 3.8 to just call `PyLong_AsLong`, with this the performance increased to 37% improvement.

I haven't changed the extraction logic for `BigInt` since that is ultimately calling `_PyLong_NumBits` which doesn't internally call `PyNumber_Index`, so I think it's right - I guess it's also used much less.

src/conversions/std/num.rs
```diff
@@ -106,17 +106,37 @@ macro_rules! int_fits_c_long {
...
+                // with python 3.8+ PyLong_AsLong takes care of calling PyNumber_Index itself
+                // hence no need for the slow `PyNumber_Index` path
+                // see https://github.com/PyO3/pyo3/pull/3742 for detials.
+                #[cfg(Py_3_8)]
+                {
+                    val = err_if_invalid_value(obj.py(), -1, unsafe {
+                        ffi::PyLong_AsLong(obj.as_ptr())
+                    })?;
+                }
+                #[cfg(not(Py_3_8))]
```
adamreichold, line 119: Since both paths compile on 3.8, I think a `if cfg!(Py_3_8) { .. } else { .. }` would be more readable.

samuelcolvin: changed.

davidhewitt (changes requested):
The current patch looks good, though I think the same thing needs to be added also around line 70 in the `int_convert_u64_or_i64` macro?

Also once that's added, please force-push to squash; GitHub merge queue doesn't let us choose to squash-merge.

src/conversions/std/num.rs
```diff
@@ -738,4 +748,14 @@ mod tests {
...
+
+    #[test]
+    fn test_i64_bool() {
+        Python::with_gil(|py| {
+            let obj = true.to_object(py);
+            assert_eq!(1, obj.extract::<i64>(py).unwrap());
+            let obj = false.to_object(py);
+            assert_eq!(0, obj.extract::<i64>(py).unwrap());
+        })
+    }
```
davidhewitt, line 760: It might be worth adding a test that a float does not covert to i32 or i64, which I believe was the motivation of calling `PyNumber_Index` directly.

samuelcolvin: added `i64` tests, for float, is that enough?

davidhewitt:
Uff unfortunately the test failure looks legitimate, `PyLong_AsUnsignedLong` and `PyLong_AsUnsignedLongLong` do _not_ call `index`, so we will have to manually wrap those to call `__index__` again. 

https://github.com/python/cpython/blob/41a94c9e7be94760baab1dcb33427d8781bea64a/Objects/longobject.c#L630C68-L630C68

samuelcolvin:
turns out the benchmarks extract a lot of `usize`s...

davidhewitt (approved):
I guess both `usize` and `u64` are the standard type in the benchmarks, yes 😂 

This looks great to me, perhaps just can you squash it please? Then let's merge 🚀

samuelcolvin:
squashed

davidhewitt:
The full build suggests that 3.8 logic isn't quite right. I've added the label now so if you push further here it'll run without going to the merge queue...

samuelcolvin:
Looks like this will fail, I think the lossy float conversion wasn't fixed until 3.10 in https://github.com/python/cpython/issues/82180.

davidhewitt:
Ok great; so now we can do the fast-path on 3.10+ instead. Makes sense.

samuelcolvin:
I'm assuming the segmentation fault with numpy on the merge queue is not something for me to fix?

davidhewitt:
Nope; rerun succeeded. There is some known flakiness on PyPy that I believe to be between them and pytest, we're just an unhappy casualty of it.

PR #3744, opened 2024-01-19
support `&Bound<PyType>` for `classmethod`

davidhewitt:
Build on stream this morning; there's a little bit of tidy up to do later.

This solves https://github.com/PyO3/pyo3/pull/3705#discussion_r1443763530 by introducing some helper functions in `impl_/pymethods.rs`

PR #3752, opened 2024-01-21
[WIP] Fix complex enum Py::new vs .into_py inconsistency

### LINKED ISSUE #3747: Py::new and .into_py are inconsistent on complex enums
`Py::new` and `.into_py` are currently inconsistent.
Note how the constructed value in the example below is _not_ an instance of the specific variant.
This is intended to be fixed before 0.21, see review comment here:
https://github.com/PyO3/pyo3/pull/3582/files#r1451402473

```rust
use pyo3::prelude::*;

#[pyclass]
enum MyEnum {
    Variant { i: i32 },
}

Python::with_gil(|py| {
    let x = Py::new(py, MyEnum::Variant { i: 42 }).unwrap();
    let cls = py.get_type::<MyEnum>();
    pyo3::py_run!(py, x cls, r#"
        assert isinstance(x, cls)
        assert not isinstance(x, cls.Variant)
    "#)
})
```

mkovaxx:
Closes #3747


pyo3-macros-backend/src/pyclass.rs
```diff
@@ -1399,6 +1400,28 @@ impl<'a> PyClassImplsBuilder<'a> {
         }
     }
 
+    fn impl_into_pyclassinit(&self) -> TokenStream {
```
mkovaxx, line 1403: @davidhewitt I've tried removing the blanket `impl From<T> for PyClassInitializer<T>` stuff and generating it for each `#[pyclass]` like this.

This has problems because these impls rely on internal details of the module in which they were originally defined.

Maybe it's better to instead generate `impl Into<PyClassInitializer<T>> for T` instead?

davidhewitt: I don't immediately see what the advantage of implementing `Into` is, but if you see a clear path, we can certainly explore it.

In general, what I'd recommend when you need to rely on crate internals like this, is to put one or more hidden public APIs inside `src/impl_/pyclass.rs`, and call that from the macro generated code.

I actually did this on stream on Friday to build #3744 :)

PR #3755, opened 2024-01-23
add list bound constructors

davidhewitt:
This adds `PyList::new_bound` and `PyList::empty_bound`, with all the usual design patterns. 

At the same time I noticed that tuple benchmarks had not yet updated to the new API, so I updated these in an accompanying commit.

I had to implement `ToPyObject` and `IntoPy` for `Borrowed` to make the benchmarks compile, I don't think there's any question that these impls should exist.

### DIFF
--- a/guide/src/migration.md
+++ b/guide/src/migration.md
@@ -45,7 +45,7 @@ pyo3 = { version = "0.21", features = ["gil-refs"] }
 
 The `PyTryFrom` trait has aged poorly, its [`try_from`] method now conflicts with `try_from` in the 2021 edition prelude. A lot of its functionality was also duplicated with `PyTypeInfo`.
 
-To tighten up the PyO3 traits ahead of [a proposed upcoming API change](https://github.com/PyO3/pyo3/issues/3382) the `PyTypeInfo` trait has had a simpler companion `PyTypeCheck`. The methods [`PyAny::downcast`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.downcast) and [`PyAny::downcast_exact`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.downcast_exact) no longer use `PyTryFrom` as a bound, instead using `PyTypeCheck` and `PyTypeInfo` respectively.
+To tighten up the PyO3 traits as part of the deprecation of the GIL Refs API the `PyTypeInfo` trait has had a simpler companion `PyTypeCheck`. The methods [`PyAny::downcast`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.downcast) and [`PyAny::downcast_exact`]({{#PYO3_DOCS_URL}}/pyo3/types/struct.PyAny.html#method.downcast_exact) no longer use `PyTryFrom` as a bound, instead using `PyTypeCheck` and `PyTypeInfo` respectively.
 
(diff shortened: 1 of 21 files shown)


pyo3-benches/benches/bench_frompyobject.rs
```diff
@@ -23,20 +23,22 @@ fn enum_from_pyobject(b: &mut Bencher<'_>) {
...
     Python::with_gil(|py| {
-        let any: &PyAny = PyList::empty(py).into();
+        let any: &Bound<'_, PyAny> = &PyList::empty_bound(py);
 
         b.iter(|| {
-            let _list: &PyList = black_box(any).extract().unwrap();
+            let list: Bound<'_, PyList> = black_box(any).extract().unwrap();
+            black_box(list)
         });
     })
```
davidhewitt, line 43: I'm hoping that these two benchmarks (which are often volatile and a bit noisy) are more stable after this change, we'll have to see...

adamreichold: Returning the list is certainly worthwhile but I don't think the black box around the return value is necessary IIRC.

guide/src/migration.md
```diff
@@ -234,7 +237,15 @@ impl PyClassAsyncIter {
...
 ### Migrating from the GIL-Refs API to `Bound<T>`
 
-TODO
+To minimise breakage of code using the GIL-Refs API, the `Bound<T>` smart pointer has been introduced by adding complements to all functions which accept or return GIL Refs. This allows code to migrate by replacing the deprecated APIs with the new ones.
+
+For example, the following APIs have gained updated variants:
+- `PyList::new`, `PyTyple::new` and similar constructors have replacements `PyList::new_bound`, `PyTuple::new_bound` etc.
+
+Because the new `Bound<T>` API brings ownership out of the PyO3 framework and into user code, there are a few places where user code is expected to need to adjust while switching to the new API:
+- Code will need to add the occasional `&` to borrow the new smart pointer as `&Bound<T>` to pass these types around (or use `.clone()` at the cost of increasing the Python reference count)
```
adamreichold, line 246: I think we should qualify "cost" here. Maybe "small cost"? I'd say that generally there is little reason to avoid these `clone()` calls in application unless profiling indicates a bottleneck. (Of course, this reasoning does not apply PyO3 itself where we do not know the usage patterns upfront.)

I would even go so far as explicitly stating that cloning `Bound` is cheap because it just bumps the Python reference count instead of actually cloning the value. (Probably in the inline docs, not in the migration guide.)

adamreichold (approved):
A suggestion on wording the documentation, otherwise LGTM.

pyo3-benches/benches/bench_frompyobject.rs
davidhewitt: You're correct, the return value from the closure is already black-boxed inside `criterion`.

My goal here was to ensure that the `.extract().unwrap()` isn't optimised out. I'll give this another pass to tidy up.

guide/src/migration.md
davidhewitt: That's a great point! I don't want to deter folks from using `clone` as lots of reference count changes will be inevitable anyway. I think I'll use "very small" even.

> I would even go so far as explicitly stating that cloning `Bound` is cheap because it just bumps the Python reference count instead of actually cloning the value. (Probably in the inline docs, not in the migration guide.)

Agreed, I think there's a new bit of documentation for `Bound` needed in `types.md`, I've got an item to do that in the tracking issue, and I've added a sub-bullet to remember to mention cloning.

PR #3757, opened 2024-01-23
update ffi structures for PyPy 7.3.15

davidhewitt:
The ABI break in #3739 got reverted in PyPy 7.3.15: https://github.com/pypy/pypy/pull/4825

... however, it looks like the constant also got renamed to match CPython, so I've updated accordingly.


davidhewitt:
Will proceed to merge to unblock CI as this is a mechanical ffi correction.

PR #3758, opened 2024-01-23
benches: include debuginfo for codspeed / flamegraphs

davidhewitt:
Changes the `bench` profile so that debug info is included, which should allow codspeed benchmarks to gather execution traces...


davidhewitt:
Well, this didn't seem to make a difference to the traces codspeed captured, so I'll close for now. Sorry for the noise.

Issue #3756, opened 2024-01-23
Change the type of length from c_long to isize (in PySlice::index)


cmpute:
https://github.com/PyO3/pyo3/blob/3af73fabcfda5ecd1357819bb8f475bff2f4f467/src/types/slice.rs#L69

The underlying function `PySlice_GetIndicesEx` has an input `length` of type `Py_ssize_t`, which should be mapped to `isize`, instead of `c_long`. The conversion from `c_long` to `Py_ssize_t` seems unnecessary.

davidhewitt:
Agreed, a PR to fix this would be greatly appreciated 👍

davidhewitt:
That said, I assume the length should be non-negative? Maybe `usize` is better, and we can overflow if the `usize` is greater than `isize::MAX`?

cmpute:
Agree!. And if more changes are allowed, I propose further change the type of start/stop/length of `PySliceIndices` all to `usize`, and implement use the recommended new APIs `PySlice_Unpack` and `PySlice_AdjustIndices`. (according to [official C-API docs](https://docs.python.org/3/c-api/slice.html#c.PySlice_GetIndicesEx))

davidhewitt:
I'm certainly open to making those changes to fit better with what Python offers, but we should consider the compatibility story. There's a lot changing in 0.21 already and I'm cautious of adding too much more.

cmpute:
> I'm certainly open to making those changes to fit better with what Python offers, but we should consider the compatibility story. There's a lot changing in 0.21 already and I'm cautious of adding too much more.

This change is already a break change, but I will try to make the changes small, when I create the PR probably this or next week.

PR #3761, opened 2024-01-26
Change the types of `PySliceIndices` and `PySlice::indices

cmpute:
This fixes https://github.com/PyO3/pyo3/issues/3756. I decided not to change the type of `length` to `usize`, to leave the user with the option to skip the check, because `isize::MAX` is already very large as a length of sequences.


davidhewitt:
Thanks! Overall I think this is a straightforward thing for users to update for, and makes things a bit tidier, so overall I'm warm to this change.

My only hesitation as I mentioned before is [migrating from 0.20 to 0.21](https://pyo3.rs/main/migration#from-020-to-021) is already set to be a big job for users so I'm a bit wary of adding more breaking changes right at this minute, especially as this doesn't seem strictly necessary.

I'd be happy to hear other maintainer's views here; otherwise if you can forgive me @cmpute, I'd prefer to take a step of caution and save this for 0.22.

cmpute:
I'm open for the merge timepoint :) Waiting for 0.22 sounds good to me!

cmpute:
I have rebased the code, not sure why the coverage check failed tho.

jessekrubin:
I am using slices, and this is causing me some trouble in upgrading to 0.22. do I **HAVE** to enable `gil-refs` just need those slice indices?

davidhewitt:
No, you need to use the new `PySliceMethods` trait.

PR #3763, opened 2024-01-26
implement `PyTracebackMethods` and `PySliceMethods`

Icxolu:
I took a stab at porting `PyTraceback` and `PySlice` over to the new `Bound` API from #3684.

Mostly mechanical changes and fixes to make the compiler happy again.

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -272,6 +272,18 @@ impl PyErr {
         exc
     }
 
+    /// Deprecated form of [`PyErr::traceback_bound`].
+    #[cfg_attr(
+        not(feature = "gil-refs"),
+        deprecated(
+            since = "0.21.0",
+            note = "`PyErr::traceback` will be replaced by `PyErr::traceback_bound` in a future PyO3 version"
+        )
+    )]
+    pub fn traceback<'py>(&'py self, py: Python<'py>) -> Option<&'py PyTraceback> {
+        self.normalized(py).ptraceback(py).map(|b| b.into_gil_ref())
+    }
+
     /// Returns the traceback of this exception object.
     ///
     /// # Examples
@@ -280,10 +292,10 @@ impl PyErr {
     ///
     /// Python::with_gil(|py| {
     ///     let err = PyTypeError::new_err(("some type error",));
-    ///     assert!(err.traceback(py).is_none());
+    ///     assert!(err.traceback_bound(py).is_none());
     /// });
     /// ```
-    pub fn traceback<'py>(&'py self, py: Python<'py>) -> Option<&'py PyTraceback> {
+    pub fn traceback_bound<'py>(&self, py: Python<'py>) -> Option<Bound<'py, PyTraceback>> {
         self.normalized(py).ptraceback(py)
     }
 
@@ -476,10 +488,16 @@ impl PyErr {
 
         #[cfg(not(Py_3_12))]
         unsafe {
+            // keep the bound `traceback` alive for entire duration of
+            // PyErr_Display. if we inline this, the `Bound` will be dropped
+            // after the argument got evaluated, leading to call with a dangling
+            // pointer.
+            let traceback = self.traceback_bound(py);
             ffi::PyErr_Display(
                 self.get_type(py).as_ptr(),
                 self.value(py).as_ptr(),
-                self.traceback(py)
+                traceback
+                    .as_ref()
                     .map_or(std::ptr::null_mut(), |traceback| traceback.as_ptr()),
             )
         }
@@ -658,15 +676,15 @@ impl PyErr {
     ///
     /// # Examples
     /// ```rust
-    /// use pyo3::{exceptions::PyTypeError, PyErr, Python};
+    /// use pyo3::{exceptions::PyTypeError, PyErr, Python, prelude::PyAnyMethods};
     /// Python::with_gil(|py| {
     ///     let err: PyErr = PyTypeError::new_err(("some type error",));
     ///     let err_clone = err.clone_ref(py);
     ///     assert!(err.get_type(py).is(err_clone.get_type(py)));
     ///     assert!(err.value(py).is(err_clone.value(py)));
-    ///     match err.traceback(py) {
-    ///         None => assert!(err_clone.traceback(py).is_none()),
-    ///         Some(tb) => assert!(err_clone.traceback(py).unwrap().is(tb)),
+    ///     match err.traceback_bound(py) {
+    ///         None => assert!(err_clone.traceback_bound(py).is_none()),
+    ///         Some(tb) => assert!(err_clone.traceback_bound(py).unwrap().is(&tb)),
     ///     }
     /// });
     /// ```
@@ -747,7 +765,7 @@ impl std::fmt::Debug for PyErr {
             f.debug_struct("PyErr")
                 .field("type", self.get_type(py))
                 .field("value", self.value(py))
-                .field("traceback", &self.traceback(py))
+                .field("traceback", &self.traceback_bound(py))
                 .finish()
         })
     }
--- a/src/types/slice.rs
+++ b/src/types/slice.rs
@@ -1,6 +1,8 @@
 use crate::err::{PyErr, PyResult};
 use crate::ffi::{self, Py_ssize_t};
-use crate::{PyAny, PyObject, Python, ToPyObject};
+use crate::ffi_ptr_ext::FfiPtrExt;
+use crate::types::any::PyAnyMethods;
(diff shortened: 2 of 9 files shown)


src/err/mod.rs
```diff
@@ -283,7 +283,7 @@ impl PyErr {
     ///     assert!(err.traceback(py).is_none());
     /// });
     /// ```
-    pub fn traceback<'py>(&'py self, py: Python<'py>) -> Option<&'py PyTraceback> {
+    pub fn traceback<'py>(&self, py: Python<'py>) -> Option<Bound<'py, PyTraceback>> {
```
davidhewitt, line 286: I think rather than changing this API we should keep it unchanged and add a new `traceback_bound` for now. Gosh, I'm going to be pleased when we can clean up again in 0.22 / 0.23 ;)

src/types/slice.rs
```diff
@@ -42,31 +44,81 @@ impl PySliceIndices {
...
-            py.from_owned_ptr(ptr)
+            )
+            .assume_owned(py)
+            .downcast_into_unchecked()
         }
     }
 
-    /// Constructs a new full slice that is equivalent to `::`.
+    /// Deprecated form of `PySlice::full_bound`.
+    #[track_caller]
```
davidhewitt, line 74: I think we don't generally expect this API to panic so we probably don't need `track_caller` here.

```suggestion
```

src/types/slice.rs
```diff
@@ -42,31 +44,81 @@ impl PySliceIndices {
 }
 
 impl PySlice {
-    /// Constructs a new slice with the given elements.
+    /// Deprecated form of `PySlice::new_bound`.
+    #[track_caller]
```
davidhewitt, line 48: Same here re no panic expected.

```suggestion
```

davidhewitt (commented):
Brilliant, thank you for helping us get this API over the finish line!

There's a few minor comments, other than those this looks great.

Are you interested in helping out with more of these? (And if so, have you got a next item that you plan to pick up?)

Also, I'd love to hear any thoughts you have about this new `Bound` API. The documentation on it is still quite thin but you've done a good job following the patterns we've established so far with it. If you had any thoughts about the name, or documentation which is missing, or anything else, now before release is a great time for us to make stuff better 😂

davidhewitt:
Also, don't worry about the codspeed "regressions" here, there's a bit of noisy volatility on a couple of particular benchmarks that probably just needs tidying up a bit.

src/err/mod.rs
Icxolu: sure, i guess I'll add add a deprecation warning to thin one than

src/types/slice.rs
Icxolu: I guess this comes from where ever I took the deprecation warning from, haven looked close enough 🙃

src/instance.rs
```diff
@@ -217,7 +217,7 @@ impl<'py, T> Bound<'py, T> {
     /// Casts this `Bound<T>` as the corresponding "GIL Ref" type.
     ///
     /// This is a helper to be used for migration from the deprecated "GIL Refs" API.
-    pub fn as_gil_ref(&'py self) -> &'py T::AsRefTarget
+    pub fn as_gil_ref(&self) -> &'py T::AsRefTarget
```
Icxolu, line 220: I do think this does not have to be tied to the lifetime 'self of the `Bound`, but only the GIL lifetime 'py. Am i correct here? Otherwise the created reference can not be returned from functions.

Icxolu:
Thanks for the warm words!

> Are you interested in helping out with more of these? (And if so, have you got a next item that you plan to pick up?)
I am, `PyComplex` seems to look like a simple one i might try next

>Also, I'd love to hear any thoughts you have about this new Bound API. The documentation on it is still quite thin but you've done a good job following the patterns we've established so far with it. If you had any thoughts about the name, or documentation which is missing, or anything else, now before release is a great time for us to make stuff better

I do like this new API, I think it makes a lot of sense to decouple the GIL lifetime, from the Rust reference lifetime. For me the naming makes sense. At least `Bound` does right now, for `Borrowed` i can see the intent, but I haven quite got the difference and overlap of `Borrowed<'a, 'py, _>` vs `&'a Bound<'py, _>` and when one would use one over the other. This is something i would need to look into in more detail, and this will probably deserve a section in the guide. I guess this has to do with that it describes something that could not be expressed with the current GIL-refs.

src/err/mod.rs
```diff
@@ -479,7 +491,7 @@ impl PyErr {
             ffi::PyErr_Display(
                 self.get_type(py).as_ptr(),
                 self.value(py).as_ptr(),
-                self.traceback(py)
+                self.traceback_bound(py)
```
davidhewitt, line 494: Interestingly here I think we now end up introducing a dangling pointer, because the `.map_or` will throw away the original bound.

So here we need to do something like

```rust
            let traceback = self.traceback_bound(py);
            ffi::PyErr_Display(
                 self.get_type(py).as_ptr(),
                 self.value(py).as_ptr(),
                 traceback.as_ref().map_or(std::ptr::null_mut(), Bound::as_ptr)
           )
```

It'll be hard to observe a crash from the dangling pointer because `traceback` will continue to be owned on the error object, but in a free threaded Python future there is just the tiniest possibility that some other thread assigns to `__traceback__` on this error at the same time as this is running, and that could then cause a use-after-free.

src/types/traceback.rs
```diff
@@ -119,10 +169,13 @@ def f():
...
             let err = f.call0().unwrap_err();
-            let traceback = err.traceback(py).unwrap();
+            let traceback = err.traceback_bound(py).unwrap();
             let err_object = err.clone_ref(py).into_py(py).into_ref(py);
 
-            assert!(err_object.getattr("__traceback__").unwrap().is(traceback));
+            assert!(err_object
+                .getattr("__traceback__")
+                .unwrap()
+                .is(traceback.as_gil_ref()));
```
davidhewitt, line 178: Does it work if we remove the `.as_gil_ref()`, and insert `&` instead? It'd be nice to avoid going back to the old API here. If it doesn't compile, it would be great for us to think how it can work (maybe as a future PR).

```suggestion
                .is(&traceback));
```

davidhewitt (commented):
Overall looking great! Fortunately renaming the method to `traceback_bound` helped catch a case where the change in ownership means we need to make an adjustment to avoid a dangling pointer. I think after that's addressed, this will be good to merge.

> I am, `PyComplex` seems to look like a simple one i might try next

Fantastic, thank you so much!

> I do like this new API, I think it makes a lot of sense to decouple the GIL lifetime, from the Rust reference lifetime. For me the naming makes sense. At least Bound does right now, for Borrowed i can see the intent, but I haven quite got the difference and overlap of `Borrowed<'a, 'py, _>` vs `&'a Bound<'py, _>` and when one would use one over the other. This is something i would need to look into in more detail, and this will probably deserve a section in the guide. I guess this has to do with that it describes something that could not be expressed with the current GIL-refs.

Yes `Borrowed` is definitely the only remaining rough edge. It's mostly forced upon us because some CPython APIs return object references that do not carry ownership, e.g. `PyTuple_GetItem`. To make those sound, we either have to increase the reference count and make them into a full `Bound`, or in some cases where we know it's safe to pass around the borrowed reference, we can make `Borrowed`. 

As much as possible I've been trying to keep `Borrowed` in the user-facing API isolated to higher-performance alternatives, e.g. `PyTupleMethods::get_borrowed_item`. Because I agree with you that most people should be thinking in `Bound` and `&Bound`, especially while these are new ideas.

I think I need to send a PR to `types.md` in the guide to try to explain these new types ASAP.

src/types/traceback.rs
Icxolu: This works! Even better.

src/err/mod.rs
Icxolu: Oh wow, very subtle. But I agree, the bound `traceback` gets dropped after the argument got evaluated, rendering the pointer invalid. I changed it to your suggestion and added a small comment for future reference.

davidhewitt (approved):
This looks great, thanks!

The `test_compile_error` failure is an annoying one; that particular UI test has been unstable lately, though it seems that for a given set of sources it is at least deterministic. It's been trending towards the output which we're getting now in this PR, so hopefully once this is merged we'll have a new stable output...

If you run `TRYBUILD=overwrite cargo test --test test_compile_error` then you should find it updates the output locally for you. Otherwise, I can push a commit for it.

davidhewitt:
Ah, maybe I should have said, the UI tests are sensitive to Rust version so you'll need to be running exactly current stable (1.75) - older or newer and the output will break CI 🙈

davidhewitt:
Looks like GitHub actions flakiness...

PR #3767, opened 2024-01-27
implement `PyComplexMethods`

Icxolu:
This ports `PyComplex` to the new Bound API from #3684.

Binary operator are implemented for different combinations of `Bound`, `&Bound`, as well as for `Borrowed` and GIL-refs. The inherent `pow` method is left with GIL-refs for backwards compatibility (I think its only callable from GIL-refs anyways?), the one on the trait uses the new API.

I think the magic `__complex__` method can be changed together with other magic methods that currently return GIL-refs.

### DIFF
--- a/guide/src/class/numeric.md
+++ b/guide/src/class/numeric.md
@@ -170,8 +170,8 @@ impl Number {
         self.0 as f64
     }
 
-    fn __complex__<'py>(&self, py: Python<'py>) -> &'py PyComplex {
-        PyComplex::from_doubles(py, self.0 as f64, 0.0)
+    fn __complex__<'py>(&self, py: Python<'py>) -> Bound<'py, PyComplex> {
+        PyComplex::from_doubles_bound(py, self.0 as f64, 0.0)
     }
 }
 ```
@@ -321,8 +321,8 @@ impl Number {
         self.0 as f64
     }
 
-    fn __complex__<'py>(&self, py: Python<'py>) -> &'py PyComplex {
-        PyComplex::from_doubles(py, self.0 as f64, 0.0)
+    fn __complex__<'py>(&self, py: Python<'py>) -> Bound<'py, PyComplex> {
+        PyComplex::from_doubles_bound(py, self.0 as f64, 0.0)
     }
 }
 
--- a/src/types/complex.rs
+++ b/src/types/complex.rs
@@ -1,4 +1,4 @@
-use crate::{ffi, PyAny, Python};
+use crate::{ffi, types::any::PyAnyMethods, Bound, PyAny, PyNativeType, Python};
 use std::os::raw::c_double;
 
 /// Represents a Python [`complex`](https://docs.python.org/3/library/functions.html#complex) object.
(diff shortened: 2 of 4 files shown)


guide/src/class/numeric.md
```diff
@@ -171,7 +171,7 @@ impl Number {
     }
 
     fn __complex__<'py>(&self, py: Python<'py>) -> &'py PyComplex {
-        PyComplex::from_doubles(py, self.0 as f64, 0.0)
+        PyComplex::from_doubles_bound(py, self.0 as f64, 0.0).into_gil_ref()
     }
```
davidhewitt, line 175: Does it work already to have this implementation?

```rust
    fn __complex__<'py>(&self, py: Python<'py>) -> Bound<'py, PyComplex> {
         PyComplex::from_doubles_bound(py, self.0 as f64, 0.0)
     }
```

src/types/complex.rs
```diff
@@ -19,118 +19,224 @@ pyobject_native_type!(
...
+                ffi::_Py_c_pow,
+            )
+            .assume_owned(self.py())
+            .downcast_into_unchecked()
+        }
+    }
 }
 
 #[cfg(not(any(Py_LIMITED_API, PyPy)))]
 mod not_limited_impls {
```
davidhewitt, line 108: I think I have a slight preference to have this `mod not_limited_impls` above the `PyComplexMethods` trait, but it's not a strong opinion if you had a reason for this ordering.

src/types/complex.rs
```diff
@@ -19,118 +19,224 @@ pyobject_native_type!(
...
         ffi::PyComplex_FromCComplex(operation(l_val, r_val))
     }
 
-    impl<'py> Add for &'py PyComplex {
-        type Output = &'py PyComplex;
-        fn add(self, other: &'py PyComplex) -> &'py PyComplex {
-            unsafe {
-                self.py()
-                    .from_owned_ptr(complex_operation(self, other, ffi::_Py_c_sum))
+    macro_rules! bin_ops {
```
davidhewitt, line 137: Nice use of a macro here 👍 

(There are sooo many implementations of these traits for the various combinations of borrowed forms 😂)

src/types/complex.rs
```diff
@@ -19,118 +19,224 @@ pyobject_native_type!(
...
+    bin_ops!(Sub, sub, -, ffi::_Py_c_diff);
+    bin_ops!(Mul, mul, *, ffi::_Py_c_prod);
+    bin_ops!(Div, div, /, ffi::_Py_c_quot);
+
+    impl<'py> Neg for &'py PyComplex {
         type Output = &'py PyComplex;
-        fn sub(self, other: &'py PyComplex) -> &'py PyComplex {
+        fn neg(self) -> &'py PyComplex {
             unsafe {
+                let val = (*(self.as_ptr() as *mut ffi::PyComplexObject)).cval;
```
davidhewitt, line 196: While you're touching up this code, in new code I have a slight preference to use `.cast` instead of `as`:

```suggestion
                let val = (*(self.as_ptr().cast::<ffi::PyComplexObject>())).cval;
```

... it might even be possible to omit some parentheses:

```suggestion
                let val = (*self.as_ptr().cast::<ffi::PyComplexObject>()).cval;
```

davidhewitt (commented):
Thank you very much for help with these implementations!

I think regarding `__complex__`, switching the return type to `Bound` hopefully already works with the traits we've got.

guide/src/class/numeric.md
Icxolu: Jup, this works already, nice 👍

src/types/complex.rs
Icxolu: I just put it there because this look roughly like the location it was added in the other modules, but I'm happy to move it down if you prefer that.

src/types/complex.rs
Icxolu: Yeah, i started with handwritten impls, but that got out of hand quickly 🙃

davidhewitt (approved):
Great, this looks good to me, thanks very much again for the help here 👍 

If you're interested in doing more and want ideas what to look into next, please do feel welcome to ping me on #3684!

Issue #3762, opened 2024-01-26
Deprecate / remove `_Py` internal APIs from `pyo3-ffi`


davidhewitt:
The main outcome I got from [my stream today](https://www.youtube.com/watch?v=oqUGdfsGPvs) was that @encukou and @vstinner (and presumably the rest of the Core Devs) want to regard CPython C API symbols prefixed with `_Py` as internal APIs which don't carry any version stability (not even necessarily in the same Python minor version patch series).

I think this means that we should consider removing all such definitions with this `_Py` prefix from `pyo3-ffi`, or at the very least making users think *very hard* before they choose to use these, if there's a good reason for any of these to be exposed in PyO3.

I took some notes on stream about where we use these definitions in PyO3 itself, they're on my desktop PC along with a small patch which removes use of a couple, I'll aim to push both tomorrow.

davidhewitt:
So, excluding `_Py_NewRef`, which we now removed, these are the symbols in use inside PyO3 which we found while I was streaming:

- [ ] `_PyCFunctionFast` / `_PyCFunctionFastWithKeywords`
   These are typedefs for function pointers to use with `METH_FASTCALL`. I think it's expected that these are stable API now, so I've opened https://github.com/python/cpython/pull/114627 to update these names (hopefully backwards-compatibly) in CPython.

- [ ] `_Py_InitializeMain`
   This is used inside only inside a PyO3 test for PEP 587. I think to be honest the test predates `pyo3-ffi-check`, so it might be that we can just delete this test now.

- [ ] `_PyLong_AsByteArray`, `_PyLong_FromByteArray`, `_PyLong_NumBits`
   These are used to implement fast paths for conversion of Python integers to Rust "big" integers which are outside the range of C `long long`. While the measurement is dependent on the size of the integer, we found on-stream that there's as much as a 2x - 9x performance slowdown in PyO3's conversion routine for switching off this fast path. So if there's room to propose these APIs as candidates for stability upstream, it would be nice to keep using these.

- [ ] `_PyUnicode_COMPACT_DATA`
   This is only used inside a PyO3 test, so I can clean that up trivially in a PR. However there's a related problem here that we rely on reading a C bitfield to make `PyUnicode_DATA` work. Nearly a year ago I opened https://github.com/python/cpython/pull/102589 to propose a (wrong) solution upstream to fix this, however @encukou already stated the correct solution on that PR and it's just waiting on me (or someone else) to complete that.

[...]

PR #3769, opened 2024-01-27
remove `test_pep_587`

davidhewitt:
Related to #3762 

This test uses `_Py_InitializeMain` as part of its machinery for testing PEP 587 embedding. Also, it was originally built as a crude way to check that the FFI structure definitions were correct.

It was originally introduced in #2374 and `pyo3-ffi-check` didn't come until later in #2423.

So, I think that for now it's best if we just remove this test, to get rid of the CPython-private `_Py_InitializeMain` usage.

I suspect that once we drop Python 3.7 we'll be implementing some new replacement for `prepare_freethreaded_python` which uses the PEP 587 APIs, so at that point we'll be adding more coverage again too.


alex (approved):
Tiven that this really tests only the ffi layer, I agree that's better covered by the ffi-check

PR #3774, opened 2024-01-29
add `PyString::new_bound`

davidhewitt:
Part of #3684

This implements `PyString::new_bound` and deprecates `PyString::new` as per the established pattern, updating internal code.

There are other constructors on `PyString`, particularly `PyString::intern`, which I will cover in a separate PR to keep the diff smaller.

### DIFF
--- a/src/err/mod.rs
+++ b/src/err/mod.rs
@@ -206,7 +206,7 @@ impl PyErr {
     ///     assert_eq!(err.to_string(), "TypeError: ");
     ///
     ///     // Case #3: Invalid exception value
-    ///     let err = PyErr::from_value(PyString::new(py, "foo").into());
+    ///     let err = PyErr::from_value(PyString::new_bound(py, "foo").as_gil_ref());
     ///     assert_eq!(
     ///         err.to_string(),
     ///         "TypeError: exceptions must derive from BaseException"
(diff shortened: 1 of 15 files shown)


davidhewitt:
Sorry for the force-pushes here; I'll try to make the next one I push a bit tidier on first push... 🤦

src/err/mod.rs
```diff
@@ -206,7 +206,7 @@ impl PyErr {
     ///     assert_eq!(err.to_string(), "TypeError: ");
     ///
     ///     // Case #3: Invalid exception value
-    ///     let err = PyErr::from_value(PyString::new(py, "foo").into());
+    ///     let err = PyErr::from_value(PyString::new_bound(py, "foo").as_gil_ref());
```
adamreichold, line 209: This will have to wait for the corresponding `PyErr` change, right?

adamreichold (approved):

src/err/mod.rs
davidhewitt: Yes, we need to add `PyErr::from_value_bound` I think.

(or add a trait which abstracts over `&PyAny` and `&Bound<PyAny>` just temporarily to ease migration, thought that'd be a new design question and less mechanical.)

PR #3777, opened 2024-01-29
add `PyBytes::new_bound`

davidhewitt:
For #3684 

This adds `PyBytes::new_bound` and `PyBytes::new_bound_with`, deprecating the old constructors and migrating internal code.

### DIFF
--- a/pytests/src/buf_and_str.rs
+++ b/pytests/src/buf_and_str.rs
@@ -42,10 +42,9 @@ impl BytesExtractor {
 }
 
 #[pyfunction]
-fn return_memoryview(py: Python<'_>) -> PyResult<&PyMemoryView> {
-    let bytes: &PyAny = PyBytes::new(py, b"hello world").into();
-    let memoryview = TryInto::try_into(bytes)?;
-    Ok(memoryview)
+fn return_memoryview(py: Python<'_>) -> PyResult<Bound<'_, PyMemoryView>> {
+    let bytes = PyBytes::new_bound(py, b"hello world");
+    PyMemoryView::from_bound(&bytes)
 }
 
 #[pymodule]
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -1,4 +1,6 @@
+use crate::ffi_ptr_ext::FfiPtrExt;
 use crate::instance::{Borrowed, Bound};
+use crate::types::any::PyAnyMethods;
 use crate::{ffi, FromPyObject, IntoPy, Py, PyAny, PyNativeType, PyResult, Python, ToPyObject};
 use std::borrow::Cow;
 use std::ops::Index;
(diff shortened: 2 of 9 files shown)


pytests/src/buf_and_str.rs
```diff
@@ -43,7 +43,7 @@ impl BytesExtractor {
 
 #[pyfunction]
 fn return_memoryview(py: Python<'_>) -> PyResult<&PyMemoryView> {
-    let bytes: &PyAny = PyBytes::new(py, b"hello world").into();
+    let bytes: &PyAny = PyBytes::new_bound(py, b"hello world").into_gil_ref();
```
Icxolu, line 46: after #3786 this could return a `Bound<PyMemoryView>` now 👼

src/types/bytes.rs
```diff
@@ -80,10 +113,23 @@ impl PyBytes {
     /// `std::slice::from_raw_parts`, this is
     /// unsafe](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety).
     pub unsafe fn from_ptr(py: Python<'_>, ptr: *const u8, len: usize) -> &PyBytes {
```
Icxolu, line 115: Since this returns a gil-ref, should we deprecate it as well, in favor of `bound_from_ptr()`?

Icxolu (commented):
Left some small suggestions above.

Also, do we want implement to `Index` for `Bound`/`&Bound` as well, or do we want to require calling `as_bytes()`  before indexing?

Otherwise LGTM

davidhewitt:
Thanks for the review!

> Also, do we want implement to Index for Bound/&Bound as well, or do we want to require calling as_bytes() before indexing?

Ah great spot - unfortunately `Index for Bound` isn't possible due to ownership requirements, there's some discussion at https://github.com/PyO3/pyo3/pull/3680#discussion_r1432944233, I'll add to the migration notes here.

pytests/src/buf_and_str.rs
davidhewitt: Ah, lovely! 🚀

src/types/bytes.rs
davidhewitt: Yes, my mistake to miss this; that's the downside of the commits I've been cherry-picking from #3606 - they were mostly trying to get a complete compile with a good range of the types rather than a complete implementation 🙈 

I'll add here 👍

davidhewitt:
> Ah great spot - unfortunately `Index` for `Bound` isn't possible due to ownership requirements, there's some discussion at https://github.com/PyO3/pyo3/pull/3680#discussion_r1432944233, I'll add to the migration notes here.

Oh _actually_ looking again here it _is_ possible in this case, because we can still return `&u8`! Even better that you made me check 👍

guide/src/migration.md
```diff
@@ -246,7 +246,7 @@ For example, the following APIs have gained updated variants:
 
 Because the new `Bound<T>` API brings ownership out of the PyO3 framework and into user code, there are a few places where user code is expected to need to adjust while switching to the new API:
 - Code will need to add the occasional `&` to borrow the new smart pointer as `&Bound<T>` to pass these types around (or use `.clone()` at the very small cost of increasing the Python reference count)
-- `Bound<PyList>` and `Bound<PyTuple>` cannot support indexing with `list[0]`, you should use `list.get_item(0)` instead.
+- `Bound<PyList>`, `Bound<PyTuple>`, and `Bound<PyBytes>` cannot support indexing with `list[i]` or `tuple[i]`, because `std::ops::Index` requires reference return values. Use `list.get_item(i)` and `tuple.get_item(i)` instead. For bytes, use `bytes.as_bytes()[i]` to index the byte slice.
```
Icxolu, line 249: Now that we have the `Index` impl for `Bound<PyBytes>` it can be removed from here again

src/types/bytes.rs
```diff
@@ -142,6 +186,15 @@ impl<I: SliceIndex<[u8]>> Index<I> for PyBytes {
     }
 }
 
+/// This is the same way [Vec] is indexed.
+impl<I: SliceIndex<[u8]>> Index<I> for Bound<'_, PyBytes> {
```
Icxolu, line 190: we could additionally implement it on `&Bound` as well, so it also works an a borrowed bound

Icxolu (commented):
> Oh actually looking again here it is possible in this case, because we can still return &u8! Even better that you made me check 👍

Great, I could not see the reason why this would not work here. Glad I'm not blind 😆

guide/src/migration.md
davidhewitt: Ufff I started writing this when I noticed it could be implemented; I didn't mean to commit this 🤦

src/types/bytes.rs
davidhewitt: I think it's not needed due to the way `bytes[I]` gets desugared into the `index` call.

I've pushed an additional part to `test_bound_bytes_index` which indexes on `&Bound<PyBytes>`, see if you agree with me that the one implementation is sufficient?

Icxolu: Perfect, that even better! 👍

adamreichold (approved):

davidhewitt:
Thanks both! 🙏

PR #3778, opened 2024-01-29
add `_bound` constructors for datetime types

davidhewitt:
This PR adds `new_bound` and variants for the `PyDate`, `PyDateTime`, `PyTime` and `PyDelta` types.

Also adds `timezone_utc_bound`.

Updates internal code to use these (basically the `chrono` conversions).

WIP - the `pytests` changes as written don't compile until #3776 allows `&Bound<T>` as a function argument.


davidhewitt:
Ok, I think this one should now be good to review! 🎉

pytests/src/datetime.rs
```diff
@@ -17,34 +17,34 @@ fn get_date_tuple<'p>(py: Python<'p>, d: &PyDate) -> Bound<'p, PyTuple> {
...
+fn make_time<'py>(
+    py: Python<'py>,
     hour: u8,
     minute: u8,
     second: u8,
     microsecond: u32,
-    tzinfo: Option<&PyTzInfo>,
-) -> PyResult<&'p PyTime> {
-    PyTime::new(py, hour, minute, second, microsecond, tzinfo)
+    tzinfo: Option<&Bound<'_, PyTzInfo>>,
```
Icxolu, line 31: I assume the compiler is smart enough to figure out that the inner lifetime here has to be `'py`, but maybe its nicer for consistency to explicitly name it if we have it named anyway.

pytests/src/datetime.rs
```diff
@@ -179,15 +187,19 @@ impl TzClass {
...
+        PyDelta::new_bound(py, 0, 3600, 0, true)
     }
 
-    fn tzname(&self, _py: Python<'_>, _dt: &PyDateTime) -> String {
+    fn tzname(&self, _dt: &Bound<'_, PyDateTime>) -> String {
         String::from("+01:00")
     }
 
-    fn dst(&self, _py: Python<'_>, _dt: &PyDateTime) -> Option<&PyDelta> {
+    fn dst(&self, _dt: &Bound<'_, PyDateTime>) -> Option<&PyDelta> {
```
Icxolu, line 202: I think this can/should return an `Option<Bound>` now

src/ffi/tests.rs
```diff
@@ -79,7 +79,7 @@ fn test_timezone_from_offset() {
     use crate::types::PyDelta;
 
     Python::with_gil(|py| {
-        let delta = PyDelta::new(py, 0, 100, 0, false).unwrap();
+        let delta = PyDelta::new_bound(py, 0, 100, 0, false).unwrap();
         let tz: &PyAny = unsafe { py.from_borrowed_ptr(PyTimeZone_FromOffset(delta.as_ptr())) };
```
Icxolu, line 83: Maybe we can remove this gil-ref (and the one below) as well, while we're here.

src/types/datetime.rs
```diff
@@ -289,11 +348,41 @@ impl PyDateTime {
...
-            py.from_owned_ptr_or_err(ptr)
+            )
+            .assume_owned_or_err(py)
+            .downcast_into_unchecked()
         }
     }
 
+    /// Deprecated form of [`PyDateTime::new_bound_with_fold`].
+    #[allow(clippy::too_many_arguments)]
+    pub fn new_with_fold<'p>(
```
Icxolu, line 359: This misses the deprecation (or was it intentional?)

src/types/datetime.rs
```diff
@@ -599,15 +765,33 @@ pub fn timezone_utc(py: Python<'_>) -> &PyTzInfo {
     unsafe { &*(ensure_datetime_api(py).TimeZone_UTC as *const PyTzInfo) }
 }
 
+/// Equivalent to `datetime.timezone.utc`
+pub fn timezone_utc_bound(py: Python<'_>) -> Bound<'_, PyTzInfo> {
```
Icxolu, line 769: We should deprecate `timezone_utc` and implement it in terms of this new one

src/types/datetime.rs
```diff
@@ -277,9 +309,36 @@ impl PyDateTime {
...
+            second,
+            microsecond,
+            tzinfo.map(PyTzInfo::as_borrowed).as_deref(),
+        )
+        .map(Bound::into_gil_ref)
+    }
+
+    /// Creates a new `datetime.datetime` object.
+    #[allow(clippy::too_many_arguments)]
+    pub fn new_bound<'p>(
```
Icxolu, line 328: While where here, lets also name the lifetimes `'py` everywhere

pytests/src/datetime.rs
davidhewitt: Agreed, I had been tempted to leave these out as it didn't feel strictly necessary. On refection just having `'py` consistently added to all `Bound` (if the lifetime is named) is a simple rule which makes readable code 👍 

I'll update the whole file.

src/types/datetime.rs
davidhewitt: Just a typical case of my error-prone typing, good catch & will fix!

src/ffi/tests.rs
davidhewitt: Good idea. Reading this code I see they're also bugged. [`PyTimeZone_FromOffset`](https://docs.python.org/3/c-api/datetime.html#c.PyTimeZone_FromOffset) is documented to return a new reference, so I'll change this to `.assume_owned(py)`.

davidhewitt:
Thanks for the thorough review! In the spirit of #3811 I've chosen to push the review feedback as new commits to make it easier to see what I've updated.

Icxolu (approved):
LGTM

src/ffi/tests.rs
Icxolu: Perfekt 👍

PR #3779, opened 2024-01-29
expose `Bound::from_owned_ptr` etc

davidhewitt:
Split from #3606 

This makes `Bound::from_owned_ptr` and variants part of the public API, instead of `pub(crate)`.

This is useful for the occasional use case outside of PyO3 which deals with raw FFI. These APIs also have equivalents with the same names which already exist on `Py<T>`, so it seems to me that we might as well have these.

### DIFF
--- a/src/instance.rs
+++ b/src/instance.rs
@@ -105,33 +105,34 @@ where
 }
 
 impl<'py> Bound<'py, PyAny> {
-    /// Constructs a new Bound from a pointer. Panics if ptr is null.
+    /// Constructs a new `Bound<'py, PyAny>` from a pointer. Panics if `ptr` is null.
     ///
     /// # Safety
     ///
-    /// `ptr` must be a valid pointer to a Python object.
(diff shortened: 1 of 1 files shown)


davidhewitt:
I suppose the one design question here is whether we think the simpler `Bound::from_ptr`, `Bound::from_ptr_or_opt` etc would be better(i.e. dropping the `_owned` bit).

Nicer names maybe, but they then lose the consistency with `Py`.

Icxolu:
I think it makes sense to have these as a complement to the ones on `Py`. I would suggest to keep the names the same. It's not that much longer to keep `_owned` and I think its easier to reason about for users if the names are consistent.

I would also suggest to deprecate the corresponding functions on `Python`, which look like the predecessors of these.

davidhewitt:
> Agreed, let me see if I can throw together something for those quickly now...

Actually there's still quite a lot of work to migrate off these internally, for now I've just added bullets to #3684 under a deprecations section.

